Issue #13127, opened 2016-03-19
[Pls send halp!] The Ultimate Glimmer 2 Test Porting Guide


chancancode:
:recycle: :recycle: :recycle: **Please consider the environment before printing this Github issue.** :recycle: :recycle: :recycle:

## The Backstory

@wycats and I (and many other people who have helped along the way) have been working on a rebuild of the rendering engine for the last six months, codenamed "Glimmer 2".

It started as a fork of [htmlbars](https://github.com/tildeio/htmlbars) but almost every line of code has been rewritten (sometimes several times) by now. We distilled all the learnings from building the previous-generation rendering pipelines (handlebars, htmlbars, the [original glimmer project](https://github.com/emberjs/ember.js/pull/10501), etc), resulting in an architecture that's simultaneously more fitting for Ember's use cases but is also more flexible and well-positioned for future extensions and other non-Ember use cases.

You can find the codez at https://github.com/tildeio/glimmer. It's (re)written in TypeScript, and I think it's pretty cool. Anyway, more on that at [EmberConf](http://emberconf.com/speakers.html#godfrey-chan).

## The Integration

While there are still a lot of work (mostly optimizations) we would like to do on the engine, we believe we have implemented enough to cover all the basic Ember use cases. So [about two months ago](https://github.com/emberjs/ember.js/pull/12772), we have started integrating the new engine into Ember proper. You can follow [this meta-issue](https://github.com/emberjs/ember.js/issues/12907) for our progress so-far.

While it is not yet possible to use the new engine in real apps just yet, we expect that work to be completed relatively soon. The expectation is that once we have finished the implementation, it will be a relatively painless, drop-in upgrade for apps, just like the original Handlebars to HTMLBars migration.

(It should be noted that the ability to toggle the feature flag will probably land *before* all the existing features are implemented, so it will probably *not* work seamless with your app from the get-go.)

## Pls send halp

So, you might be wondering, "How can I help?"

I am glad you asked! :boom: :sparkles: :fireworks: :tada: 

At this point, the highest value you can do to help is to help port the existing tests (and help reviewing these PRs). You see, Ember has a pretty extensive test suite that tests the "view layer" behavior. The problem is that a lot of these tests are written in ways that are quite coupled to the existing implementation or otherwise uses legacy semantics (such as `{{view.foo}}`) that are no longer supported.

[...]

PR #13136, opened 2016-03-19
[Glimmer2] Port {{#each-in}} tests

mmun:
Part of #13127.
- Adds a simple `assertStableRerender` helper because it was a pattern I kept reusing.
- Ports the each-in tests and uses BasicConditionalsTest to test the inverse behaviour.

cc @chancancode.


packages/ember-glimmer/tests/integration/syntax/each-in-test.js
```diff
@@ -0,0 +1,134 @@
...
+      categories: null
+    });
+
+    // The empty comment is a placeholder for empty content.
+    this.assertHTML(`<ul><!----></ul>`);
+
+    this.assertStableRerender();
+  }
+
+  [`@test it renders the inverse block when given a falsy value`]() {
```
chancancode: Can we test this by subclassing `BasicConditionalsTest` like https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/tests/integration/syntax/each-test.js#L14 ?

mmun: Fiiiiine. I'll try.

packages/ember-glimmer/tests/integration/syntax/each-in-test.js
```diff
@@ -0,0 +1,134 @@
...
+
+    this.assertHTML(strip`
+      <ul>
+        <li>Televisions: 183</li>
+        <li>Alarm Clocks: 999</li>
+      </ul>
+    `);
+
+    this.assertStableRerender();
+  }
```
chancancode: Can we merge this test into the previous test using the "I-N-U-R" pattern? (I agree that in this case the existing test already covered the same things we want to cover with that pattern, but a lot of existing tests doesn't do it. It seems good to have one pattern that we can bring everywhere and we are guaranteed to get good coverage every time.)

packages/ember-glimmer/tests/utils/abstract-test-case.js
```diff
@@ -164,6 +164,12 @@ export class TestCase {
       this.assertSameNode(newSnapshot[i], oldSnapshot[i]);
     }
   }
+
+  assertStableRerender() {
+    this.takeSnapshot();
+    this.runTask(() => this.rerender());
+    this.assertInvariants();
+  }
```
chancancode: :+1:

mmun:
Updated.

packages/ember-glimmer/tests/integration/syntax/each-in-test.js
```diff
@@ -0,0 +1,142 @@
...
+        {{/each-in}}
+      </ul>
+    `, {
+      categories: {
+        'Smartphones': 8203,
+        'JavaScript Frameworks': Infinity
+      }
+    });
+
+    this.assertStableRerender();
```
chancancode: Can you move `assertStableRerender` *after* `assertHTML`? (After a brief pause, I realize they are isomorphic because `assertStableRerender` takes a snapshot and assert that they are the same, but the "initial render -> assert result -> assertStableRerender" flow matches other test cases and make the intention more clear.)

packages/ember-glimmer/tests/integration/syntax/each-in-test.js
```diff
@@ -0,0 +1,142 @@
...
+    this.assertHTML(strip`
+      <ul>
+        <li>Smartphones: 8203</li>
+        <li>JavaScript Frameworks: Infinity</li>
+      </ul>
+    `);
+
+    this.runTask(() => set(this.context, 'categories', {
+      'Smartphones': 100
+    }));
```
chancancode: This should be tested via mutation:

```javascript
this.runTask(() => set(this.context, 'categories.Smartphones', 100));
```

It *might* need to be this for HTMLBars to work?

```javascript
this.runTask(() => {
  set(this.context, 'categories.Smartphones', 100)
  this.rerender(); // TODO: this is not needed in Glimmer 2
});

this.assertHTML(strip`
  <ul>
    <li>Smartphones: 100</li>
    <li>JavaScript Frameworks: Infinity</li>
  </ul>
`);
```

I think we can do away with the explicit re-render in Glimmer 2, but if it's required to make HTMLBars pass, then it seems fine to have it.

packages/ember-glimmer/tests/integration/syntax/each-in-test.js
```diff
@@ -0,0 +1,142 @@
...
+      'Bad Programming Languages': 456
+    }));
+
+    this.assertHTML(strip`
+      <ul>
+        <li>Programming Languages: 199303</li>
+        <li>Good Programming Languages: 123</li>
+        <li>Bad Programming Languages: 456</li>
+      </ul>
+    `);
```
chancancode: This is testing replacement, which is usually covered implicitly with the "reset via replacement" step. It seems fine to also test this as an extra step in here if you want. However, can you still add the "reset via replacement" step?

```javascript
this.runTask(() => {
  set(this.context, 'categories', {
    'Smartphones': 8203,
    'JavaScript Frameworks': Infinity
  }
});

this.assertHTML(strip`
  <ul>
    <li>Smartphones: 100</li>
    <li>JavaScript Frameworks: Infinity</li>
  </ul>
`);
```

This step has helped us catch some subtle caching bugs in the past. Since it's hard to reason about when those kind of bugs are possible (we were definitely surprised a few times), we decided to make this mandatory. (And we need to test replacement anyway, which is why we usually fold it into a single "reset with replacement" step.)

packages/ember-glimmer/tests/integration/syntax/each-in-test.js
```diff
@@ -0,0 +1,142 @@
...
+    `, { categories });
+
+    this.assertStableRerender();
+
+    this.assertHTML(strip`
+      <ul>
+        <li>Televisions: 183</li>
+        <li>Alarm Clocks: 999</li>
+      </ul>
+    `);
```
chancancode: ditto (re-order)

packages/ember-glimmer/tests/integration/syntax/each-in-test.js
```diff
@@ -0,0 +1,142 @@
...
+
+    this.runTask(() => set(this.context, 'categories', {
```
chancancode: Add mutation step

packages/ember-glimmer/tests/integration/syntax/each-in-test.js
```diff
@@ -0,0 +1,142 @@
...
+
+    this.assertHTML(strip`
+      <ul>
+        <li>Programming Languages: 199303</li>
+        <li>Good Programming Languages: 123</li>
+        <li>Bad Programming Languages: 456</li>
+      </ul>
+    `);
+
+    this.runTask(() => set(this.context, 'categories', categories));
```
chancancode: Can you make a new `categories` object here for the replacement step?

Something like:

```javascript
categories = Object.create(protoCategories);
categories['Televisions'] = 183;
categories['Alarm Clocks'] = 999;
```

packages/ember-glimmer/tests/integration/syntax/each-in-test.js
```diff
@@ -0,0 +1,142 @@
...
+    // TODO: figure out what the rest of the cases are
+    {},
+    Object.create({ 'Not Empty': 1 }),
+    undefined,
+    null
+  ])
+
+  // TODO(mmun): Add support for object proxies and
+  // include the ObjectTestCases mixin.
+);
```
chancancode: :clap: :tada:

chancancode:
Added some comments about the INUR test pattern, otherwise looks great :eyes:

chancancode:
(I added a few FIXMEs in `packages/ember-glimmer/tests/integration/content-test.js` in #13141. When you rebase, can you also swap them with your new helper? :grimacing:)

packages/ember-glimmer/tests/integration/syntax/each-in-test.js
mmun: Yes, my bad. I should have left it how it was!

packages/ember-glimmer/tests/integration/syntax/each-in-test.js
mmun: This is not possible in HTMLBars because

```js
set(this.context, 'categories.Smartphones', 100)
```

does not dirty anything because there is no stream observing `categories.Smartphones`. This would require a "star" observer or hooking Ember.set, but either way it would be a change from the current semantics.

mmun: Apparently `this.rerender()` does not care if things are dirty or not because it does indeed update the DOM after calling it...

mmun:
@chancancode Ready for another round of review!

PR #13143, opened 2016-03-20
[Glimmer2] Move escape tests to `ember-glimmer`

code0100fun:
Moved to content tests.
- [X] Convert capitalized path deprecation test
- [X] Convert null object properties test
- [X] Add trusted curlies `moduleFor` (works with TextNode ignore in `assertInvariants`)
- [X] HTML content in non-escaped curlies
- [X] HTML content in escaped curlies
- [X] Test all escapes of matrix values

### DIFF
--- a/packages/ember-glimmer/tests/integration/content-test.js
+++ b/packages/ember-glimmer/tests/integration/content-test.js
@@ -59,6 +59,54 @@ moduleFor('Static content tests', class extends RenderingTest {
 
 });
 
+moduleFor('Escaped content tests', class extends RenderingTest {
+
+  ['@test it renders HTML as escaped text in normal curlies']() {
+    this.render('{{htmlContent}}{{nested.htmlContent}}', {
+      htmlContent: '<b>Max</b>',
+      nested: { htmlContent: '<b>James</b>' }
+    });
+    this.assert.equal(this.$('b').length, 0);
+    this.assertText('<b>Max</b><b>James</b>');
+
+    this.runTask(() => this.rerender());
+    this.assert.equal(this.$('b').length, 0);
+    this.assertText('<b>Max</b><b>James</b>');
+
+    this.runTask(() => set(this.context, 'htmlContent', '<i>Max</i>'));
+    this.assert.equal(this.$('i,b').length, 0);
+    this.assertText('<i>Max</i><b>James</b>');
+
+    this.runTask(() => set(this.context, 'htmlContent', '<b>Max</b>'));
+    this.assert.equal(this.$('b').length, 0);
+    this.assertText('<b>Max</b><b>James</b>');
+
+    this.runTask(() => set(this.context, 'nested', { htmlContent: '<i>James</i>' }));
+    this.assert.equal(this.$('b,i').length, 0);
+    this.assertText('<b>Max</b><i>James</i>');
+  }
+
+  ['@test it renders HTML in "trusted" curlies']() {
+    this.render('{{{htmlContent}}}{{{nested.htmlContent}}}', {
+      htmlContent: '<b>Max</b>',
+      nested: { htmlContent: '<b>James</b>' }
+    });
+    this.assertHTML('<b>Max</b><b>James</b>');
+
+    this.runTask(() => this.rerender());
+    this.assertHTML('<b>Max</b><b>James</b>');
+
+    this.runTask(() => set(this.context, 'htmlContent', '<i>Max</i>'));
+    this.assertHTML('<i>Max</i><b>James</b>');
+
+    this.runTask(() => set(this.context, 'htmlContent', '<b>Max</b>'));
+    this.assertHTML('<b>Max</b><b>James</b>');
+
+    this.runTask(() => set(this.context, 'nested', { htmlContent: '<i>James</i>' }));
+    this.assertHTML('<b>Max</b><i>James</i>');
+  }
+});
+
 class DynamicContentTest extends RenderingTest {
 
   /* abstract */
@@ -85,6 +133,30 @@ class DynamicContentTest extends RenderingTest {
     this.runTask(() => set(this.context, 'message', 'hello'));
 
     this.assertText('hello');
+    this.assertInvariants();
+  }
+
+  ['@test it can render a capitalized path with no deprication']() {
+    this.renderPath('CaptializedPath', { CaptializedPath: 'noDeprication' });
+
+    expectNoDeprecation();
+
+    this.assertText('noDeprication');
+
+    this.assertStableRerender();
+
+    this.runTask(() => set(this.context, 'CaptializedPath', 'stillNoDeprication'));
+
+    expectNoDeprecation();
+
+    this.assertText('stillNoDeprication');
+    this.assertInvariants();
+
+    this.runTask(() => set(this.context, 'CaptializedPath', 'noDeprication'));
+
+    expectNoDeprecation();
+
+    this.assertText('noDeprication');
     this.assertInvariants();
   }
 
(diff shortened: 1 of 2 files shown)


packages/ember-glimmer/tests/integration/content-test.js
```diff
@@ -73,6 +73,8 @@ class DynamicContentTest extends RenderingTest {
   ['@test it can render a dynamic path']() {
     this.renderPath('message', { message: 'hello' });
 
+    expectNoDeprecation();
```
code0100fun: To cover [this test](https://github.com/emberjs/ember.js/blob/caa99153e6abb26b186f3c2beb19b20694e6858c/packages/ember-htmlbars/tests/integration/escape_integration_test.js#L24). Not sure if this is necessary anymore. Looks like its the last piece related to deprecated global lookup (86fa0ab)?

packages/ember-htmlbars/tests/integration/escape_integration_test.js
```diff
@@ -1,117 +0,0 @@
...
-  equal(view.$().text(), '1');
-
-  run(function() {
-    set(context, 'aNumber', 2);
-  });
-
-  equal(view.$().text(), '2');
-});
-
-QUnit.test('should read from an Object.create(null)', function() {
```
code0100fun: Not sure if this is now covered by [these changes](https://github.com/emberjs/ember.js/blob/c6556381a56f0c7fd72ed9c2ab8cf1bea08823fd/packages/ember-glimmer/tests/integration/content-test.js#L241) from c655638, @chancancode?

packages/ember-glimmer/tests/integration/content-test.js
chancancode: :+1: :scissors:

packages/ember-htmlbars/tests/integration/escape_integration_test.js
chancancode: I think that test is slightly different: the test I added was for `{{nullObject}}` (which apparently does not work in htmlbars :P), whereas this is testing `{{nullObject.foo}}` works. So it seems fine to keep both!

packages/ember-glimmer/tests/integration/content-test.js
chancancode: Actually, the that test is testing that when you do `{{CaptializedPath}}`, *and* that you have a `CaptializedPath` property on the object, it does not emit a deprecation warning. Can you add this test back? (And remove this `expectNoDeprecation`)

packages/ember-htmlbars/tests/integration/escape_integration_test.js
```diff
@@ -1,117 +0,0 @@
...
-
-QUnit.module('ember-htmlbars: Escaped Integration', {
-  teardown() {
-    runDestroy(view);
-
-    view = null;
-  }
-});
-
-QUnit.test('should read from a global-ish simple local path without deprecation', function() {
```
chancancode: https://github.com/emberjs/ember.js/pull/13143/files#r57370979

chancancode:
I think most of these tests can be "matrix tested" by adding a new module:

```
moduleFor('Dynamic content tests ("trusted" curlies)', class extends DynamicContentTest {

  renderPath(path, context = {}) {
    this.render(`{{{${path}}}`, context);
  }

});
```

This would run all the shared tests cases against `{{{foo}}}`, following the I-N-U-R pattern as appropriate. You can then add more methods in this class to test extra stuff not covered by the shared cases.

packages/ember-glimmer/tests/integration/content-test.js
```diff
@@ -59,6 +59,102 @@ moduleFor('Static content tests', class extends RenderingTest {
 
 });
 
+moduleFor('Null object tests', class extends RenderingTest {
+  ['@test it should read from an Object.create(null)']() {
```
chancancode: This could probably go into the shared `DynamicContentTest` to test against all the possible positions. Something like:

```javascript
  ['@test it can read from a null object']() {
    let nullObject = Object.create(null);
    nullObject['message'] = 'hello';

    this.renderPath('nullObject.message', { nullObject });

    this.assertText('hello');

    this.assertStableRerender();

    this.runTask(() => set(nullObject, 'message', 'goodbye'));

    this.assertText('goodbye');
    this.assertInvariants();

    nullObject = Object.create(null);
    nullObject['message'] = 'hello';

    this.runTask(() => set(this.context, 'nullObject', nullObject));

    this.assertText('hello');
    this.assertInvariants();
  }

```

packages/ember-glimmer/tests/integration/content-test.js
```diff
@@ -59,6 +59,102 @@ moduleFor('Static content tests', class extends RenderingTest {
...
+
+    this.runTask(() => set(this.context, 'aNumber', 2));
+    this.assertText('2,3,2');
+
+    this.runTask(() => set(this.context, 'aNumber', 1));
+    this.assertText('1,3,1');
+
+    this.runTask(() => set(this.context, 'nested', { aNumber: 4 }));
+    this.assertText('1,4,1');
+  }
```
chancancode: If we make the "escaped" position part of the matrix, it should be covered by the shared test cases

packages/ember-glimmer/tests/utils/abstract-test-case.js
```diff
@@ -179,7 +179,9 @@ export class TestCase {
     this.assert.strictEqual(newSnapshot.length, oldSnapshot.length, 'Same number of nodes');
 
     for (let i = 0; i < oldSnapshot.length; i++) {
-      this.assertSameNode(newSnapshot[i], oldSnapshot[i]);
+      if (!(newSnapshot[i] instanceof TextNode && oldSnapshot[i] instanceof TextNode)) {
+        this.assertSameNode(newSnapshot[i], oldSnapshot[i]);
+      }
```
code0100fun: It looks like escaped content creates a new node instead of updating the node value. This happens in htmlbars and glimmer. Not sure if this is intended or expected behavior or a bug. For now this is how I have made the `assertInvariants` pass for escaped text content.

chancancode: I think this is fine for now as long as we clean it up before merge. I'll try to come up a suggestion. I have a few ideas, but I need to look into the details to know if they would work.

As for your question, this is expected. The content inside a triple curlies are parsed and populated by the browser. If the source string changes, we have to wipe out the existing nodes and have the browser re-create them. So this is currently just a definciency in the test abstraction, not a bug in the implementation.

packages/ember-glimmer/tests/integration/content-test.js
```diff
@@ -244,6 +340,14 @@ moduleFor('Dynamic content tests (content position)', class extends DynamicConte
 
 });
 
+moduleFor('@htmlbars Dynamic content tests ("trusted" curlies)', class extends DynamicContentTest {
```
code0100fun: These tests are currently [failing](https://travis-ci.org/emberjs/ember.js/jobs/120383033) on IE9/10/11. Tagged them for HTMLBars for now.

code0100fun:
I couldn't figure out a clean way to integrate the escaped HTML tests into the test matrix. 

All the existing values in the matrix are tested in `{{{triple curlies}}}` but there are some glimmer failures on IE. 

I left explicit tests of HTML content mainly because `assertInvariants` was causing issues with HTML content and I don't have a clear understanding of how to refactor the test runner to account for these changes. 

@chancancode, hopefully this is close to acceptable. Sorry you've had to sink so much time on this 😅

chancancode:
Thank you for your work on this! I'll take a look tomorrow!

chancancode:
@code0100fun Thank you again for your work, I merged this in #13259. This turned out to be a bit more difficult to abstract than I expected, because it's testing something pretty low-level. HTMLBars tend to have more "marker" elements than Glimmer (empty text nodes, etc), but I think I have found an okay path. It's still not as nice as I wanted it to be, but it's good enough for now! (You can see the result in #13259.)

If you or other people have better ideas for those abstractions, feel free to open subsequent PRs to improve it.

PR #13144, opened 2016-03-21
[CLEANUP canary] Remove Ember.Select

HeroicEric:
As requested in https://github.com/emberjs/ember.js/issues/13127:

This removes `Ember.Select`.

Views were removed from the 2.0 API unless using the [ember-legacy-views](https://github.com/emberjs/ember-legacy-views) addon, which is intended to end support after Ember v2.4. This includes `Ember.select`.

http://emberjs.com/deprecations/v1.x/#toc_ember-view


rwjblue:
@HeroicEric next round is on me :heart:

mixonic:
Can we make this Canary? I don't see much of a reason for dropping legacy helpers in 2.5-beta. It is already a beta release, we missed the boat, no need to rush this.

rwjblue:
Agreed. Canary seems good.

jonnii:
good night sweet prince

HeroicEric:
Updated the commit message for canary

mixonic:
Thanks @HeroicEric :fire: :fire: :fire: :fire: :fire: :fire: :fire:

mixonic:
I'm merging this after some discussion by a subset of core. We continue to make progress on landing the glimmer refactor into Ember master ([see epic burndown](https://github.com/emberjs/ember.js/issues/13127)). It is not expected that this work makes into into canary in time for Ember 2.6, however 2.7 is looking likely.

In several recent blog posts and issues core has raised that **Ember 2.4 will likely be the final release supporting the legacy view and controller APIs**. It looks like that may actually be 2.5, but as 2.4 is an LTS support will live on in 2.4 for far longer than 2.5 regardless.

So that 2.7 is not a complete rush, we're opening the current canary (2.6) to removals of APIs incompatible with the glimmer refactor. Additionally this will be an opportunity to cull a good amount of code from Ember that has only existed to support the addons via intimate feature flags. These are already deprecated in 2.4 per the LTS guidelines.

Please see the [deprecation guide](http://emberjs.com/deprecations/v1.x/#toc_ember-select) for further details about migrating away from Ember's select solution, or try out the fantastic [ember-power-select](https://github.com/cibernox/ember-power-select).

wycats:
@mixonic I would probably be more circumspect about exact versions :wink:

The high order bit is we want to ship Glimmer 2 as soon as we can, but no sooner. In the meantime, we'll have the six-week release cycle to keep us warm at night.

PR #13145, opened 2016-03-21
[Glimmer 2] Add support for dynamic `tagName`

chancancode:


packages/ember-glimmer/lib/components/curly-component.js
```diff
@@ -77,6 +77,16 @@ const MANAGER = new CurlyComponentManager();
...
 import Component from '../ember-views/component';
 
+function tagName(vm) {
+  let { tagName } = vm.dynamicScope().view;
+
+  if (tagName === '') {
+    throw new Error('Not implemented: fragments (`tagName: ""`)');
+  }
+
+  return new ValueReference(tagName || 'div');
```
rwjblue: Technically, we need to support falsey tag name (tagName: '') so this will need to be tweaked as we move forward right?

rwjblue: I think we also support tagName as a CP (though that should be deprecated and unlikely)

chancancode: Confirm. We are just throwing (on line 84) for now so that whoever running into this will not be confused

rwjblue: Kk, gotcha. For future travelers the current implementation is https://github.com/emberjs/ember.js/blob/master/packages/ember-views/lib/system/build-component-template.js#L136-L153

PR #13146, opened 2016-03-21
[Glimmer2] Remove tagless-view-tests

chadhietala:
Should be covered by
https://github.com/emberjs/ember.js/blob/bab44470c733b0ab285739dd971473b5ff2b4fed/packages/ember-glimmer/tests/utils/shared-conditional-tests.js#L709.

Part of #13127

/cc @chancancode


chancancode:
:scissors: :scissors: :scissors:

rwjblue:
We should ensure we have tests for these descriptions. The implementations all tested {{if}} and {{each}}, but the important thing was that given a tagless view it can add remove children. AFAIK, this is not supported in glimmer2 yet...

PR #13147, opened 2016-03-21
[Glimmer2] Move Helper Lookup Test

chadhietala:
While this is indirectly tested in multiple places probably the best to
have an actually test for it.

Part of #13127

/cc @chancancode


rwjblue:
:+1:

PR #13148, opened 2016-03-21
[Glimmer2] Removing append-templated-view-test

chadhietala:
This test most definitly testing dynamic lookup characteristics of the
`{{#view}}` helper.

Part of #13127

/cc @chancancode


chancancode:
:fire:

PR #13151, opened 2016-03-21
[Glimmer2] Bring back "adding and removing children" tests

chadhietala:
This adds a shared test that follows I-N-U-R for adding and removing
children. These types of tests were removed in #13146.

Part of #13127

/cc @rwjblue @chancancode

### DIFF
--- a/packages/ember-glimmer/tests/utils/shared-conditional-tests.js
+++ b/packages/ember-glimmer/tests/utils/shared-conditional-tests.js
@@ -745,6 +745,51 @@ export class TogglingSyntaxConditionalsTest extends TogglingConditionalsTest {
     this.assertText('F-outer');
   }
 
+  ['@test can add and remove children']() {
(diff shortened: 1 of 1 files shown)


chancancode:
I might be misunderstanding – I think the adding and removing is not the important thing here: going from `["inner","-","before"]` to `["inner-after"]` is the same as removing all three then adding one back (I think it's good to test mutation, so free free to change/add more "U" step in the previous test, instead of adding another one).

I might be wrong, but I think @rwjblue's concern is we lost a file that's supposed to (although not anymore) testing "tagless" views, aka fragments (`Component.extend(tagName: '', ...)`. The conditionals used to be implemented like that, and that's why the file was named that way.

I think there are already some tests for that in components invocation, but I could be wrong.

rwjblue:
@chancancode is correct, my concern was for the exact things being tested in the test removed, but for the conceptual testing of tagless components.

chadhietala:
Closing as this does not address the concern.

PR #13156, opened 2016-03-21
[Glimmer2] Move willDestroyElement tests.

krisselden:
This is more a direct port, I'd rather test all the lifecycle hooks, (at least do didInsertElement along with willDestroyElement) for built-ins that can add/remove components.


packages/ember-glimmer/tests/integration/components/will-destroy-element-hook-test.js
```diff
@@ -0,0 +1,30 @@
+import { set } from 'ember-metal/property_set';
+import Component from 'ember-views/components/component';
+import { moduleFor, RenderingTest } from '../../utils/test-case';
+
+moduleFor('Component willDestroyElement hook', class extends RenderingTest {
+  ['@htmlbars it calls willDestroyElement when removed by if'](assert) {
+    let willDestroyElementCount = 0;
+    let FooBarComponent = Component.extend({
+      didInsertElement() {
+        assert.notEqual(this.element.parentNode, null, 'precond component is in DOM');
```
chancancode: Is there a reason you moved the assertion into here (as opposed to have the hooks set an outer var and assert as part of the testing flow)? If the hook did not fire, it won't fail the test. The `willDestroyElement` hook happens to not have this problem because of the counter, but it seems easy for future contributors to accidentally mess that up. The previous "set an outer var" approach seemed better to me

chancancode:
This was merged into our branch in #13168 :+1:

PR #13161, opened 2016-03-22
[Glimmer2] [CLEANUP canary] Remove deprecated legacy-each support including {{collection}}

HeroicEric:
I spent some time trying to figure out how to remove the `collection` helper/keyword/view for https://github.com/emberjs/ember.js/issues/13127. As far as I can tell, a bunch of other things depend on it and also need to be removed in order to do so.
- [x] `{{each itemView=}}`
- [x] `{{each itemViewClass=}}`
- [x] `{{each itemController=}}`
- [x] `{{each tagName=}}`
- [x] `{{each emptyView=}}`
- [x] `{{each emptyViewClass=}}`
- [x] `{{collection}}`

Much of my thoughts here are based on my understanding of what's going on in https://github.com/emberjs/ember.js/blob/master/packages/ember-template-compiler/lib/plugins/transform-each-into-collection.js#L43-L57

There is still cleanup to do here but I tried to organize the commits so that they could possibly be brought in separately. That worked fine until I got to actually removing the 
### Questions
1. Am I correct in thinking that these things also need to be removed?
2. Should I break out these commits into smaller PRs?

Guidance welcomed :smile:


packages/ember-htmlbars/lib/templates/legacy-each.hbs
```diff
@@ -1,21 +1,9 @@
 {{~#each view._arrangedContent -legacy-keyword=view.keyword as |item|~}}
```
rwjblue: I think this whole file should :fire:

packages/ember-template-compiler/lib/plugins/transform-each-into-collection.js
```diff
@@ -44,12 +44,7 @@ function validate(node) {
   if ((node.type === 'BlockStatement' || node.type === 'MustacheStatement') && node.path.original === 'each') {
     return any(node.hash.pairs, pair => {
       let key = pair.key;
-      return key === 'itemController' ||
-             key === 'itemView' ||
-             key === 'itemViewClass' ||
-             key === 'tagName' ||
-             key === 'emptyView' ||
-             key === 'emptyViewClass';
+      return key === 'emptyViewClass';
```
rwjblue: This whole file should :fire:

packages/ember-views/lib/mixins/empty_view_support.js
```diff
@@ -23,42 +20,5 @@ export default Mixin.create({
...
-      } else {
-        set(actualEmpty, 'template', inverse);
-      }
-    } else if (inverse && emptyViewClass) {
-      return emptyViewClass.extend({ template: inverse });
-    }
-
-    return actualEmpty;
-  })
+  emptyViewClass: View
```
rwjblue: Whole file :skull:

packages/ember-views/lib/views/legacy_each_view.js
```diff
@@ -3,15 +3,12 @@
 //behavior such as itemController.
 
 import legacyEachTemplate from 'ember-htmlbars/templates/legacy-each';
-import { get } from 'ember-metal/property_get';
 import { computed } from 'ember-metal/computed';
```
rwjblue: :hocho:

rwjblue:
> Am I correct in thinking that these things also need to be removed?

Yes

> Should I break out these commits into smaller PRs?

No

HeroicEric:
@rwjblue I still need to update the two references to `CollectionView` in the examples I tagged in the last commit but I think the rest is ready to go.

Should squash when done?

packages/ember-template-compiler/lib/index.js
```diff
@@ -31,13 +30,8 @@ registerPlugin('ast', TransformInputOnToOnEvent);
...
 registerPlugin('ast', TransformUnescapedInlineLinkTo);
 registerPlugin('ast', AssertNoEachIn);
-
-if (_Ember.ENV._ENABLE_LEGACY_VIEW_SUPPORT) {
-  registerPlugin('ast', TransformEachIntoCollection);
-} else {
-  registerPlugin('ast', AssertNoViewAndControllerPaths);
-  registerPlugin('ast', AssertNoViewHelper);
-}
+registerPlugin('ast', AssertNoViewAndControllerPaths);
```
rwjblue: These should only be registered if the legacy view flag is off. We definitely should have a follow up PR to remove the flag and all usages of it, but it should be a separate change from this one IMO.

rwjblue:
I left one more tweak, but this is looking good. Once you have those docs updated, go ahead and squash and we can land this...

HeroicEric:
I think this is all set :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire:

packages/ember-runtime/lib/system/native_array.js
```diff
@@ -114,13 +114,16 @@ NativeArray = NativeArray.without.apply(NativeArray, ignore);
 
   Example
 
+  TODO: Update example to not use CollectionView
```
rwjblue: Need to kill this TODO comment, looks like the docs below already took care of it...

rwjblue:
Sorry, just one more nit to pick...

packages/ember-runtime/lib/system/native_array.js
rwjblue: Fixed in b295edb.

rwjblue:
Thanks for this @HeroicEric!

PR #12877, opened 2016-01-31
[BUGFIX release] Upgrade htmlbars to 0.14.14

mmun:
For more information see https://github.com/tildeio/htmlbars/pull/447.

PR #13162, opened 2016-03-22
[BUGFIX release-2-0] Upgrade htmlbars to 0.14.14

Dhaulagiri:
Brings the input fix from https://github.com/emberjs/ember.js/pull/12877 to those still on 2.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "express": "^4.5.0",
     "github": "^0.2.3",
     "glob": "~4.3.2",
-    "htmlbars": "0.14.2-2",
+    "htmlbars": "0.14.14",
     "qunit-extras": "^1.3.0",
     "qunitjs": "^1.16.0",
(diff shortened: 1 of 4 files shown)


rwjblue:
@Dhaulagiri - The failures seem to be JSHint related (most likely improvements in JSHint parsing engine now catches things that were previously skipped).  Can you add a second commit fixing them?

Dhaulagiri:
@rwjblue I fixed the jshint errors but there are still quite a few other failures.  It's probably not worth spinning tires on this too much since we are just trying to back port this fix.

rwjblue:
Hmm, I thought when I looked that there were only the JSHint errors.  I'll double check again after Travis is done with this run...

Dhaulagiri:
This was not as easy as I hoped so I think our effort is better spent getting our app off 2.0

PR #13163, opened 2016-03-22
Controller#replaceRoute should be public

alexspeller:
This method should be public, incorrectly marked as private in the big update last year


rwjblue:
Discussed in core team call today, this was intended to be public and is somewhat required as a compliment to `transitionToRoute`.

PR #13173, opened 2016-03-23
[Glimmer2] Implement get helper

ro0gr:
Left to do:
- [x] Figure out how to setup `{{input}}` helper 
- [x] Port tests using `input` and `mut`
- [x] Figure our what is the purpose of `get()` for the `GetHelperReference`
- [x] tests for `get()` 
- [x] handle `null` for `sourceReference` and `propertyPathReference` (marked tests as `[htmlbars]` only)

### DIFF
--- a/packages/ember-glimmer/lib/utils/references.js
+++ b/packages/ember-glimmer/lib/utils/references.js
@@ -53,6 +53,7 @@ export class UpdatableReference extends RootReference {
   }
 }
 
+// @implements PathReference
 export class GetHelperReference {
   constructor(sourceReference, pathReference) {
     this.sourceReference = sourceReference;
--- a/packages/ember-glimmer/tests/integration/helpers/get-test.js
+++ b/packages/ember-glimmer/tests/integration/helpers/get-test.js
@@ -1,14 +1,14 @@
 import { RenderingTest, moduleFor } from '../../utils/test-case';
 import { set } from 'ember-metal/property_set';
 import { get } from 'ember-metal/property_get';
-import EmberObject from 'ember-runtime/system/object';
 import TextField from 'ember-views/views/text_field';
 import Component from 'ember-views/components/component';
 
 moduleFor('Helpers test: {{get}}', class extends RenderingTest {
+
   ['@test should be able to get an object value with a static key']() {
     this.render(`[{{get colors 'apple'}}] [{{if true (get colors 'apple')}}]`, {
-      colors: { apple: 'red', banana: 'yellow' }
+      colors: { apple: 'red' }
     });
 
     this.assertText('[red] [red]');
@@ -17,16 +17,24 @@ moduleFor('Helpers test: {{get}}', class extends RenderingTest {
 
     this.assertText('[red] [red]');
 
-    this.runTask(() => set(this.context, 'colors', {
-      apple: 'green', banana: 'purple'
-    }));
+    this.runTask(() => set(this.context, 'colors.apple', 'green'));
 
     this.assertText('[green] [green]');
+
+    this.runTask(() => set(this.context, 'colors', {
+      apple: 'red'
+    }));
+
+    this.assertText('[red] [red]');
   }
 
   ['@test should be able to get an object value with nested static key']() {
     this.render(`[{{get colors "apple.gala"}}] [{{if true (get colors "apple.gala")}}]`, {
-      colors: { apple: { gala: 'red and yellow' }, banana: 'yellow' }
+      colors: {
+        apple: {
+          gala: 'red and yellow'
+        }
+      }
     });
 
     this.assertText('[red and yellow] [red and yellow]');
@@ -35,15 +43,13 @@ moduleFor('Helpers test: {{get}}', class extends RenderingTest {
 
     this.assertText('[red and yellow] [red and yellow]');
 
-    this.runTask(() => set(this.context, 'colors', {
-      apple: { gala: 'yellow and red striped' }, banana: 'purple'
-    }));
+    this.runTask(() => set(this.context, 'colors.apple.gala', 'yellow and red striped'));
 
     this.assertText('[yellow and red striped] [yellow and red striped]');
 
-    this.runTask(() => set(this.context, 'colors.apple.gala', 'yellow-redish'));
+    this.runTask(() => set(this.context, 'colors', { apple: { gala: 'red and yellow' } }));
 
-    this.assertText('[yellow-redish] [yellow-redish]');
+    this.assertText('[red and yellow] [red and yellow]');
   }
 
   ['@test should be able to get an object value with a bound/dynamic key']() {
@@ -51,21 +57,30 @@ moduleFor('Helpers test: {{get}}', class extends RenderingTest {
       colors: { apple: 'red', banana: 'yellow' },
       key: 'apple'
     });
+
+    this.assertText('[red] [red]');
+
+    this.runTask(() => this.rerender());
+
     this.assertText('[red] [red]');
 
     this.runTask(() => set(this.context, 'key', 'banana'));
+
     this.assertText('[yellow] [yellow]');
 
-    this.runTask(() => set(this.context, 'colors', {
-      apple: 'green',
-      banana: 'purple'
-    }));
+    this.runTask(() => {
+      set(this.context, 'colors.apple', 'green');
+      set(this.context, 'colors.banana', 'purple');
+    });
+
     this.assertText('[purple] [purple]');
 
     this.runTask(() => set(this.context, 'key', 'apple'));
+
     this.assertText('[green] [green]');
 
-    this.runTask(() => set(this.context, 'colors.apple', 'red'));
+    this.runTask(() => set(this.context, 'colors', { apple: 'red' }));
+
     this.assertText('[red] [red]');
   }
 
@@ -80,13 +95,24 @@ moduleFor('Helpers test: {{get}}', class extends RenderingTest {
       },
       key: 'apple.gala'
     });
+
+    this.assertText('[red and yellow] [red and yellow]');
+
+    this.runTask(() => this.rerender());
+
     this.assertText('[red and yellow] [red and yellow]');
 
     this.runTask(() => set(this.context, 'key', 'apple.mcintosh'));
+
     this.assertText('[red] [red]');
 
     this.runTask(() => set(this.context, 'key', 'banana'));
+
     this.assertText('[yellow] [yellow]');
+
+    this.runTask(() => set(this.context, 'key', 'apple.gala'));
+
+    this.assertText('[red and yellow] [red and yellow]');
   }
 
   ['@test should be able to get an object value with subexpression returning nested key']() {
@@ -95,42 +121,70 @@ moduleFor('Helpers test: {{get}}', class extends RenderingTest {
         apple: {
           gala: 'red and yellow',
           mcintosh: 'red'
-        },
-        banana: 'yellow'
+        }
       },
       key: 'apple.gala'
     });
+
     this.assertText('[red and yellow] [red and yellow]');
 
-    this.runTask(() => set(this.context, 'colors', { apple: { gala: 'yellow and red striped' }, banana: 'purple' }));
+    this.runTask(() => this.rerender());
+
+    this.assertText('[red and yellow] [red and yellow]');
+
+    this.runTask(() => set(this.context, 'colors.apple.gala', 'yellow and red striped'));
+
     this.assertText('[yellow and red striped] [yellow and red striped]');
 
     this.runTask(() => set(this.context, 'colors.apple.gala', 'yellow-redish'));
+
     this.assertText('[yellow-redish] [yellow-redish]');
+
+    this.runTask(() => set(this.context, 'colors', {
+        apple: {
+          gala: 'red and yellow',
+          mcintosh: 'red'
+        }
+      })
+    );
+
+    this.assertText('[red and yellow] [red and yellow]');
   }
 
-  ['@test should be able to get an object value with a GetStream key']() {
+  ['@test should be able to get an object value with a get helper as the key']() {
     this.render(`[{{get colors (get possibleKeys key)}}] [{{if true (get colors (get possibleKeys key))}}]`, {
       colors: { apple: 'red', banana: 'yellow' },
       key: 'key1',
       possibleKeys: { key1: 'apple', key2: 'banana' }
     });
+
+    this.assertText('[red] [red]');
+
+    this.runTask(() => this.rerender());
+
     this.assertText('[red] [red]');
 
     this.runTask(() => set(this.context, 'key', 'key2'));
+
     this.assertText('[yellow] [yellow]');
 
-    this.runTask(() => set(this.context, 'colors', { apple: 'green', banana: 'purple' }));
+    this.runTask(() => {
+      set(this.context, 'colors.apple', 'green');
+      set(this.context, 'colors.banana', 'purple');
+    });
+
     this.assertText('[purple] [purple]');
 
     this.runTask(() => set(this.context, 'key', 'key1'));
+
     this.assertText('[green] [green]');
 
-    this.runTask(() => set(this.context, 'colors.apple', 'red'));
+    this.runTask(() => set(this.context, 'colors', { apple: 'red', banana: 'yellow' }));
+
     this.assertText('[red] [red]');
   }
 
-  ['@test should be able to get an object value with a GetStream value and bound/dynamic key']() {
+  ['@test should be able to get an object value with a get helper value as a bound/dynamic key']() {
     this.render(`[{{get (get possibleValues objectKey) key}}] [{{if true (get (get possibleValues objectKey) key)}}]`, {
       possibleValues: {
         colors1: { apple: 'red', banana: 'yellow' },
@@ -139,25 +193,37 @@ moduleFor('Helpers test: {{get}}', class extends RenderingTest {
       objectKey: 'colors1',
       key: 'apple'
     });
+
+    this.assertText('[red] [red]');
+
+    this.runTask(() => this.rerender());
+
     this.assertText('[red] [red]');
 
     this.runTask(() => set(this.context, 'objectKey', 'colors2'));
+
     this.assertText('[green] [green]');
 
     this.runTask(() => set(this.context, 'objectKey', 'colors1'));
+
     this.assertText('[red] [red]');
 
     this.runTask(() => set(this.context, 'key', 'banana'));
+
     this.assertText('[yellow] [yellow]');
 
     this.runTask(() => set(this.context, 'objectKey', 'colors2'));
+
     this.assertText('[purple] [purple]');
 
     this.runTask(() => set(this.context, 'objectKey', 'colors1'));
+
     this.assertText('[yellow] [yellow]');
+
+    this.runTask(() => set(this.context, 'key', 'apple'));
   }
 
-  ['@test should be able to get an object value with a GetStream value and GetStream key']() {
+  ['@test should be able to get an object value with a get helper as the value and a get helper as the key']() {
     this.render(`[{{get (get possibleValues objectKey) (get possibleKeys key)}}] [{{if true (get (get possibleValues objectKey) (get possibleKeys key))}}]`, {
       possibleValues: {
         colors1: { apple: 'red', banana: 'yellow' },
@@ -170,25 +236,38 @@ moduleFor('Helpers test: {{get}}', class extends RenderingTest {
       },
       key: 'key1'
     });
+
+    this.assertText('[red] [red]');
+
+    this.runTask(() => this.rerender());
+
     this.assertText('[red] [red]');
 
     this.runTask(() => set(this.context, 'objectKey', 'colors2'));
+
     this.assertText('[green] [green]');
 
     this.runTask(() => set(this.context, 'objectKey', 'colors1'));
+
     this.assertText('[red] [red]');
 
     this.runTask(() => set(this.context, 'key', 'key2'));
+
     this.assertText('[yellow] [yellow]');
 
     this.runTask(() => set(this.context, 'objectKey', 'colors2'));
+
     this.assertText('[purple] [purple]');
 
-    this.runTask(() => set(this.context, 'objectKey', 'colors1'));
-    this.assertText('[yellow] [yellow]');
+    this.runTask(() => {
+      set(this.context, 'objectKey', 'colors1');
+      set(this.context, 'key', 'key1');
+    });
+
+    this.assertText('[red] [red]');
   }
 
-  ['@test should yield `get`']() {
+  ['@test the result of a get helper can be yielded']() {
     let fooBarInstance;
     let FooBarComponent = Component.extend({
       init() {
@@ -212,18 +291,21 @@ moduleFor('Helpers test: {{get}}', class extends RenderingTest {
     this.assertText('banana');
 
     this.runTask(() => this.rerender());
+
     this.assertText('banana');
 
     this.runTask(() => {
       set(fooBarInstance, 'mcintosh', 'yellow');
       set(this.context, 'colors', { yellow: 'bus' });
     });
+
     this.assertText('bus');
 
     this.runTask(() => {
       set(fooBarInstance, 'mcintosh', 'red');
       set(this.context, 'colors', { red: 'banana' });
     });
+
     this.assertText('banana');
   }
 
@@ -231,12 +313,43 @@ moduleFor('Helpers test: {{get}}', class extends RenderingTest {
     this.render(`[{{get colors 'apple'}}] [{{if true (get colors 'apple')}}]`, {
       colors: null
     });
+
+    this.assertText('[] []');
+
+    this.runTask(() => this.rerender());
+
     this.assertText('[] []');
 
     this.runTask(() => set(this.context, 'colors', { apple: 'green', banana: 'purple' }));
+
     this.assertText('[green] [green]');
 
     this.runTask(() => set(this.context, 'colors', null));
+
+    this.assertText('[] []');
+  }
+
+  ['@htmlbars should handle object keys as nulls']() {
+    this.render(`[{{get colors key}}] [{{if true (get colors key)}}]`, {
+      colors: {
+        apple: 'red',
+        banana: 'yellow'
+      },
+      key: null
+    });
+
+    this.assertText('[] []');
+
+    this.runTask(() => this.rerender());
+
+    this.assertText('[] []');
+
+    this.runTask(() => set(this.context, 'key', 'banana'));
+
+    this.assertText('[yellow] [yellow]');
+
+    this.runTask(() => set(this.context, 'key', null));
+
     this.assertText('[] []');
   }
 
@@ -245,23 +358,29 @@ moduleFor('Helpers test: {{get}}', class extends RenderingTest {
       colors: null,
       key: null
     });
+
     this.assertText('[] []');
   }
 
-  ['@htmlbars get helper value should be updatable using {{input}} and (mut) - dynamic key']() {
+  ['@htmlbars get helper value should be updatable using {{input}} and (mut) - dynamic key'](assert) {
     this.registerComponent('-text-field', { ComponentClass: TextField });
 
     this.render(`{{input type='text' value=(mut (get source key)) id='get-input'}}`, {
-      source: EmberObject.create({
+      source: {
         banana: 'banana'
-      }),
+      },
       key: 'banana'
     });
 
-    this.assert.strictEqual(this.$('#get-input').val(), 'banana');
+    assert.strictEqual(this.$('#get-input').val(), 'banana');
+
+    this.runTask(() => this.rerender());
+
+    assert.strictEqual(this.$('#get-input').val(), 'banana');
 
     this.runTask(() => set(this.context, 'source.banana', 'yellow'));
-    this.assert.strictEqual(this.$('#get-input').val(), 'yellow');
+
+    assert.strictEqual(this.$('#get-input').val(), 'yellow');
 
     this.runTask(() => {
       this.$('#get-input').val('some value');
@@ -277,60 +396,79 @@ moduleFor('Helpers test: {{get}}', class extends RenderingTest {
       this.component.childViews[0]._elementValueDidChange();
     });
 
-    this.assert.strictEqual(this.$('#get-input').val(), 'some value');
-    this.assert.strictEqual(get(this.context, 'source.banana'), 'some value');
+    assert.strictEqual(this.$('#get-input').val(), 'some value');
+    assert.strictEqual(get(this.context, 'source.banana'), 'some value');
+
+    this.runTask(() => set(this.context, 'source', { banana: 'banana' }));
+
+    assert.strictEqual(this.$('#get-input').val(), 'banana');
   }
 
-  ['@htmlbars get helper value should be updatable using {{input}} and (mut) - dynamic nested key']() {
+  ['@htmlbars get helper value should be updatable using {{input}} and (mut) - dynamic nested key'](assert) {
     this.registerComponent('-text-field', { ComponentClass: TextField });
 
     this.render(`{{input type='text' value=(mut (get source key)) id='get-input'}}`, {
-      source: EmberObject.create({
+      source: {
         apple: {
           mcintosh: 'mcintosh'
         }
-      }),
+      },
       key: 'apple.mcintosh'
     });
 
+    assert.strictEqual(this.$('#get-input').val(), 'mcintosh');
 
-    this.assert.strictEqual(this.$('#get-input').val(), 'mcintosh');
+    this.runTask(() => this.rerender());
+
+    assert.strictEqual(this.$('#get-input').val(), 'mcintosh');
 
     this.runTask(() => set(this.context, 'source.apple.mcintosh', 'red'));
 
-    this.assert.strictEqual(this.$('#get-input').val(), 'red');
+    assert.strictEqual(this.$('#get-input').val(), 'red');
 
     this.runTask(() => {
       this.$('#get-input').val('some value');
       this.component.childViews[0]._elementValueDidChange();
     });
 
-    this.assert.strictEqual(this.$('#get-input').val(), 'some value');
-    this.assert.strictEqual(get(this.context, 'source.apple.mcintosh'), 'some value');
+    assert.strictEqual(this.$('#get-input').val(), 'some value');
+    assert.strictEqual(get(this.context, 'source.apple.mcintosh'), 'some value');
+
+    this.runTask(() => set(this.context, 'source', { apple: { mcintosh: 'mcintosh' } }));
+
+    assert.strictEqual(this.$('#get-input').val(), 'mcintosh');
   }
 
-  ['@htmlbars get helper value should be updatable using {{input}} and (mut) - static key']() {
+  ['@htmlbars get helper value should be updatable using {{input}} and (mut) - static key'](assert) {
     this.registerComponent('-text-field', { ComponentClass: TextField });
 
     this.render(`{{input type='text' value=(mut (get source 'banana')) id='get-input'}}`, {
-      source: EmberObject.create({
+      source: {
         banana: 'banana'
-      }),
+      },
       key: 'banana'
     });
 
-    this.assert.strictEqual(this.$('#get-input').val(), 'banana');
+    assert.strictEqual(this.$('#get-input').val(), 'banana');
+
+    this.runTask(() => this.rerender());
+
+    assert.strictEqual(this.$('#get-input').val(), 'banana');
 
     this.runTask(() => set(this.context, 'context.source.banana', 'yellow'));
 
-    this.assert.strictEqual(this.$('#get-input').val(), 'yellow');
+    assert.strictEqual(this.$('#get-input').val(), 'yellow');
 
     this.runTask(() => {
       this.$('#get-input').val('some value');
       this.component.childViews[0]._elementValueDidChange();
     });
 
-    this.assert.strictEqual(this.$('#get-input').val(), 'some value');
-    this.assert.strictEqual(get(this.context, 'source.banana'), 'some value');
+    assert.strictEqual(this.$('#get-input').val(), 'some value');
+    assert.strictEqual(get(this.context, 'source.banana'), 'some value');
+
+    this.runTask(() => set(this.context, 'source', { banana: 'banana' }));
+
+    assert.strictEqual(this.$('#get-input').val(), 'banana');
   }
 });


packages/ember-glimmer/lib/utils/references.js
```diff
@@ -52,6 +52,27 @@ export class UpdatableReference extends RootReference {
...
+
+  value() {
+    let objRef = this.positionalArgs.at(0).value(),
+      pathRef = this.positionalArgs.at(1).value();
+
+    return objRef ? get(objRef, pathRef) : objRef;
+  }
+
+  get(propertyKey) {
+    throw new Error("get() is not implemented");
```
ro0gr: Can't find when `get()` is  supposed to be called. It's not invoked with tests.
I believe I'll find the answer but any guidance appreciated.

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,224 @@
...
+      key: null
+    });
+    this.assertText('[] []');
+  };
+
+  // Looks like some misconfiguration here.
+  //
+  // Htmlbars: Error: Assertion Failed: HTMLBars error: Could not find component named "-text-field" (no component or template with that name was found)
+  // Glimmer: Compile Error: input is not a helper
+  ['@skip get helper value should be updatable using {{input}} and (mut) - dynamic key']() {
```
ro0gr: I get exceptions dumped in comments above for both `[htmlbars]` and `[glimmer]` modes.
`[glimmer]` fault is clear enough. Currently can't figure out why does `[htmlbars]` mode fails while the old-fashioned htmlbars tests work. Looks like some kind of `RenderingTest` test misconfiguration. Haven't investigated yet.

packages/ember-glimmer/lib/utils/references.js
```diff
@@ -52,6 +52,27 @@ export class UpdatableReference extends RootReference {
...
+    this.positionalArgs = args.positional;
+  }
+
+  isDirty() { return true; }
+
+  value() {
+    let objRef = this.positionalArgs.at(0).value(),
+      pathRef = this.positionalArgs.at(1).value();
+
+    return objRef ? get(objRef, pathRef) : objRef;
```
ro0gr: Fullfils [should handle object values as nulls](https://github.com/emberjs/ember.js/pull/13173/files#diff-350c933360e807598f74de6573a7fcaaR189) test.

Should it return `null` for any falsy `objRef` value?

packages/ember-glimmer/lib/utils/references.js
chadhietala: I believe get is used for something that is opaque from the consumer.

```
{{!-- application.hbs --}}
{{foo-bar as |value|}}
  {{value.foo}}
{{/foo-bar}}
```

```
{{!-- templates/components/foo-bar.hbs --}}
{{yield (get someObj somePathThatPointsToAnotherObj)}}
```

chadhietala: Doesn't seem to be a test for this. I would suggest adding a test that does this type of yielding.

packages/ember-glimmer/lib/utils/references.js
chadhietala: I think it's a requirement of this helper to have 2 arguments passed https://github.com/emberjs/ember.js/blob/master/packages/ember-htmlbars/lib/keywords/get.js#L89-L90. This assertion needs to be brought back. It will also simplify this code and actually becomes pretty simple I believe.

```
export class GetHelperReference {
  constructor(args) {
    this.positionalArgs = args.positional;
  }

  isDirty() { return true; }

  value() {
    let objRef = this.positionalArgs.at(0).value();
    let pathRef = this.positionalArgs.at(1).value();
    return  get(objRef, pathRef);
  },
  get(key) {
   return get(this.value(), key);
  }
}
```

packages/ember-glimmer/lib/utils/references.js
chancancode: In general, a `Reference` is an abstract data type with the following interface:

```
interface Reference<T> {
  value(): T;
}
```

A reference represents a "computation" (similar to a "stream"), and calling `.value()` will return the current value of that computation. For example, a reference representing `{{foo}}` will return the current value of the `foo` property in the template's context. This gives you a stable value/wrapper (hence the name "reference") to hold on to that you can query in the future to re-render the template with the latest value as needed.

A `PathReference` is an extension to the system:

```
interface PathReference<T> extends Reference<T> {
  get(key: string): PathReference<any>;
}
```

Implementing the `PathReference` interface means that the `value()`s of the computation supports a conceptual `get` operation. For the purpose of understanding this, you can apply the `Ember.get` semantics here – i.e. the `value()`s returned by this computation makes sense as input to the first argument of `Ember.get`.

Alternatively, you can also think about the `get` semantics in terms of the templates. Anything `foo` that can be used in a `{{foo.bar}}` position must implement `PathReference`. In this case, `{{foo.bar}}` will take the `{{foo}}` reference as input and call `.get("bar")` on it to produce another stable reference (another `PathReference, actually) representing `{{foo.bar}}`.

In practice, almost anything that can be used in a template needs to implement `PathReference`, because in handlebars almost anything can be used in that position. In addition to @chadhietala's example, there are a few other ways that a `get` helper can get into that position: `{{(get foo bar).baz}}`, `{{#with (get foo bar) as |gotBar|}} {{gotBar.baz}} {{/with}}`, etc.

In this case, the reference you are implementing is representing the result of the `(get foo bar)` computation over time, where foo and bar themselves are also `PathReferences`. The `get(path)` method (confusingly named in this context) should produce a reference representing the `(get foo bar).path` computation.

Putting everything together, I think something like this would work:

```
...
toReference(args) {
  let sourceReference = args.positional.at(0); // foo in (get foo bar)
  let propertyPathReference = args.positional.at(1); // bar in (get foo bar)

  if (isConst(propertyPathReference)) {
    // (get foo "bar"), so `propertyPathReference` is a reference whose value cannot change ever
    // this is pretty pointless, you could have just typed `{{foo.bar}}`
    return sourceReference.get(propertyPathReference.value());
  } else {
    return GetHelperReference(sourceReference, propertyPathReference);
  }
}
...

class GetHelperReference {
  // ...mostly following your implementation

  get(path) {
    new PropertyReference(this, path);
  }
}
```

wycats: Another thing to understand is why you need `PathReference` in the first place. Certainly, you can just call `.value()`  on the parent and use a regular property lookup on the reified value.

The reason for `PathReference` is that the underlying root reference might have additional information about when you need to reify the value for paths derived off of it:

1. in Ember, if nobody called `.set()` on the parent object, you don't need to recompute the value.
2. in Ember, if you're looking at a computed property, you can rely on the computed property cache.
3. when working with immutable data structures, you only need to recompute the value if the root of the data structure has been replaced.
4. when working with primitive values, the `get()` virtually always produces a null value.
5. when working with non-configurable, non-writable values, you only need to compute the value once.

In other words, there's all kinds of information in the JavaScript **data** layer about how to most-efficiently derive a path off of a value, and `PathReference` allows the owner of the data to encode that strategy.

In Ember, this results in significant efficiency gains when Ember objects are used. Those gains would be portable to other data management strategies like immutable.js without forcing every data structure in your entire app to use Ember objects or immutable.js.

packages/ember-glimmer/lib/utils/references.js
ro0gr: @chadhietala yes, sounds reasonable. Anyway currently we have [this](https://github.com/emberjs/ember.js/blob/master/packages/ember-htmlbars/tests/helpers/get_test.js#L325-L337) `ember-htmlbars` test. 
It ensures that we support `nulls`.
So my concern here is should we support `nulls` in the `ember-glimmer` or to just mark this test as `[htmlbars]`.

packages/ember-glimmer/lib/utils/references.js
ro0gr: Thank you guys for such comprehensive explanations. It's definitely more than I could expect!

packages/ember-glimmer/lib/utils/references.js
chadhietala: Does it not work in glimmer?

ro0gr: No, at least with my imlementation

ro0gr:
Sorry for a long progress. Some irrelevant tasks take much time.

I've added `get` implementation in `GetHelperReference` as @chancancode proposed and an appropriate `yield` test according to @chadhietala request.

Currently when you try to
```
this.sourceReference.get(this.pathReference.value()).value();
```
when the `sourceReference` has a falsy underlying value it produces the following error:
```
Cannot call get with 'apple' on an undefined object.
``` 
That's why [null check tests](https://github.com/emberjs/ember.js/pull/13173/files#diff-350c933360e807598f74de6573a7fcaaR229) are marked as `htmlbars` only for now.

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+    this.assertText('[red and yellow] [red and yellow]');
+
+    this.runTask(() => set(this.context, 'colors', { apple: { gala: 'yellow and red striped' }, banana: 'purple' }));
+    this.assertText('[yellow and red striped] [yellow and red striped]');
+
+    this.runTask(() => set(this.context, 'colors.apple.gala', 'yellow-redish'));
+    this.assertText('[yellow-redish] [yellow-redish]');
+  }
+
+  ['@test should be able to get an object value with a GetStream key']() {
```
ro0gr: Any suggestions for a better name here? `GetStream` looks like an `htmlbars` term

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+    this.assertText('[purple] [purple]');
+
+    this.runTask(() => set(this.context, 'key', 'key1'));
+    this.assertText('[green] [green]');
+
+    this.runTask(() => set(this.context, 'colors.apple', 'red'));
+    this.assertText('[red] [red]');
+  }
+
+  ['@test should be able to get an object value with a GetStream value and bound/dynamic key']() {
```
ro0gr: Any suggestions for a better name here? `GetStream` looks like an `htmlbars` term

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+    this.assertText('[yellow] [yellow]');
+
+    this.runTask(() => set(this.context, 'objectKey', 'colors2'));
+    this.assertText('[purple] [purple]');
+
+    this.runTask(() => set(this.context, 'objectKey', 'colors1'));
+    this.assertText('[yellow] [yellow]');
+  }
+
+  ['@test should be able to get an object value with a GetStream value and GetStream key']() {
```
ro0gr: Any suggestions for a better name here? `GetStream` looks like an `htmlbars` term

chancancode:
@ro0gr I probably need to circle back to this after EmberConf (at least after my talk :smile:). Meanwhile, I wrote up the stuff we talked about in more details here: https://github.com/tildeio/glimmer/compare/master...chancancode:docs

ro0gr:
@chancancode no problem. I'm not in hurry. 
Thank you for the link.

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
+import { RenderingTest, moduleFor } from '../../utils/test-case';
+import { set } from 'ember-metal/property_set';
+import { get } from 'ember-metal/property_get';
+import EmberObject from 'ember-runtime/system/object';
+import TextField from 'ember-views/views/text_field';
+import Component from 'ember-views/components/component';
+
+moduleFor('Helpers test: {{get}}', class extends RenderingTest {
```
chancancode: nit: we usually have a new blank line here for the new tests

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+  ['@test should be able to get an object value with a static key']() {
+    this.render(`[{{get colors 'apple'}}] [{{if true (get colors 'apple')}}]`, {
+      colors: { apple: 'red', banana: 'yellow' }
```
chancancode: :scissors: banana

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+
+    this.runTask(() => this.rerender());
+
+    this.assertText('[red] [red]');
+
+    this.runTask(() => set(this.context, 'colors', {
+      apple: 'green', banana: 'purple'
+    }));
+
+    this.assertText('[green] [green]');
```
chancancode: Should do the "R(eset)" step by changing it back to `apple` back to red (and an assert)

packages/ember-htmlbars/tests/helpers/get_test.js
```diff
@@ -1,495 +0,0 @@
...
-    run(function() {
-      Ember.TEMPLATES = {};
-    });
-    runDestroy(view);
-    runDestroy(owner);
-    owner = view = null;
-  }
-});
-
-QUnit.test('should be able to get an object value with a static key', function() {
```
chancancode: :heavy_check_mark: should be able to get an object value with a static key

packages/ember-htmlbars/tests/helpers/get_test.js
```diff
@@ -1,495 +0,0 @@
...
-  equal(view.$().text(), '[green] [green]', 'should return \'green\' for {{get colors \'apple\'}}');
-
-  run(function() {
-    view.set('context.colors.apple', 'red');
-  });
-
-  equal(view.$().text(), '[red] [red]', 'should return \'red\' for {{get colors \'apple\'}}');
-});
-
-QUnit.test('should be able to get an object value with nested static key', function() {
```
chancancode: :heavy_check_mark: should be able to get an object value with nested static key

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+  }
+
+  ['@test should be able to get an object value with nested static key']() {
+    this.render(`[{{get colors "apple.gala"}}] [{{if true (get colors "apple.gala")}}]`, {
+      colors: { apple: { gala: 'red and yellow' }, banana: 'yellow' }
```
chancancode: :scissors: banana

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+
+    this.runTask(() => set(this.context, 'colors', {
+      apple: { gala: 'yellow and red striped' }, banana: 'purple'
+    }));
+
+    this.assertText('[yellow and red striped] [yellow and red striped]');
+
+    this.runTask(() => set(this.context, 'colors.apple.gala', 'yellow-redish'));
+
+    this.assertText('[yellow-redish] [yellow-redish]');
```
chancancode: Test reset after this

packages/ember-htmlbars/tests/helpers/get_test.js
```diff
@@ -1,495 +0,0 @@
...
-  equal(view.$().text(), '[yellow and red striped] [yellow and red striped]', 'should return \'yellow and red striped\' for {{get colors \'apple.gala\'}}');
-
-  run(function() {
-    view.set('context.colors.apple.gala', 'yellow-redish');
-  });
-
-  equal(view.$().text(), '[yellow-redish] [yellow-redish]', 'should return \'yellow-redish\' for {{get colors \'apple.gala\'}}');
-});
-
-QUnit.test('should be able to get an object value with a bound/dynamic key', function() {
```
chancancode: :heavy_check_mark: should be able to get an object value with a bound/dynamic key

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+
+    this.assertText('[red] [red]');
+
+    this.runTask(() => this.rerender());
+
+    this.assertText('[red] [red]');
+
+    this.runTask(() => set(this.context, 'colors', {
+      apple: 'green', banana: 'purple'
+    }));
```
chancancode: Should use mutation here: `set(this.context, 'colors.apple', 'green')`

packages/ember-glimmer/tests/integration/helpers/get-test.js
chancancode: (The reset should use "replacement" like you did above: `set(this.context, 'colors', { apple: 'green' })`)

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+
+    this.assertText('[red and yellow] [red and yellow]');
+
+    this.runTask(() => this.rerender());
+
+    this.assertText('[red and yellow] [red and yellow]');
+
+    this.runTask(() => set(this.context, 'colors', {
+      apple: { gala: 'yellow and red striped' }, banana: 'purple'
+    }));
```
chancancode: mutation (`set(this.context, 'colors.apple. gala', 'yellow and red striped')`)

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+    });
+    this.assertText('[red] [red]');
+
+    this.runTask(() => set(this.context, 'key', 'banana'));
+    this.assertText('[yellow] [yellow]');
+
+    this.runTask(() => set(this.context, 'colors', {
+      apple: 'green',
+      banana: 'purple'
+    }));
```
chancancode: mutation

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+    this.assertText('[purple] [purple]');
+
+    this.runTask(() => set(this.context, 'key', 'apple'));
+    this.assertText('[green] [green]');
+
+    this.runTask(() => set(this.context, 'colors.apple', 'red'));
```
chancancode: replacement (`set(this.context, 'colors', { apple: 'red' })`)

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+    this.assertText('[red] [red]');
```
chancancode: also – extremely nit: there are some inconsistency between blank lines `this.runTask` and `this.assertText`, sometimes they are bundled, sometimes there is a new line between them; I think we happen to prefer the moar space side of things in the new tests, but you can just pick one or the other for this file and stick to it.

packages/ember-htmlbars/tests/helpers/get_test.js
```diff
@@ -1,495 +0,0 @@
...
-  equal(view.$().text(), '[green] [green]', 'should return \'green\' for {{get colors key}} (key = \'apple\')');
-
-  run(function() {
-    view.set('context.colors.apple', 'red');
-  });
-
-  equal(view.$().text(), '[red] [red]', 'should return \'red\' for {{get colors key}}  (key = \'apple\')');
-});
-
-QUnit.test('should be able to get an object value with nested dynamic key', function() {
```
chancancode: :heavy_check_mark: should be able to get an object value with nested dynamic key

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+      },
+      key: 'apple.gala'
+    });
+    this.assertText('[red and yellow] [red and yellow]');
+
+    this.runTask(() => set(this.context, 'key', 'apple.mcintosh'));
+    this.assertText('[red] [red]');
+
+    this.runTask(() => set(this.context, 'key', 'banana'));
+    this.assertText('[yellow] [yellow]');
```
chancancode: add a reset step (set key back to 'apple.gala')

packages/ember-htmlbars/tests/helpers/get_test.js
```diff
@@ -1,495 +0,0 @@
...
-  equal(view.$().text(), '[red] [red]', 'should return \'red\' for {{get colors \'apple.mcintosh\'}}');
-
-  run(function() {
-    view.set('context.key', 'banana');
-  });
-
-  equal(view.$().text(), '[yellow] [yellow]', 'should return \'yellow\' for {{get colors \'banana\'}}');
-});
-
-QUnit.test('should be able to get an object value with subexpression returning nested key', function() {
```
chancancode: :heavy_check_mark: should be able to get an object value with subexpression returning nested key

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+  }
+
+  ['@test should be able to get an object value with subexpression returning nested key']() {
+    this.render(`[{{get colors (concat 'apple' '.' 'gala')}}] [{{if true (get colors (concat 'apple' '.' 'gala'))}}]`, {
+      colors: {
+        apple: {
+          gala: 'red and yellow',
+          mcintosh: 'red'
+        },
+        banana: 'yellow'
```
chancancode: :scissors: unused banana

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+
+    this.runTask(() => set(this.context, 'colors.apple', 'red'));
+    this.assertText('[red] [red]');
+  }
+
+  ['@test should be able to get an object value with nested dynamic key']() {
+    this.render(`[{{get colors key}}] [{{if true (get colors key)}}]`, {
+      colors: {
+        apple: {
+          gala: 'red and yellow',
```
chancancode: :scissors: unused gala (or add a step to use it?)

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+          mcintosh: 'red'
+        },
+        banana: 'yellow'
+      },
+      key: 'apple.gala'
+    });
+    this.assertText('[red and yellow] [red and yellow]');
+
+    this.runTask(() => set(this.context, 'colors', { apple: { gala: 'yellow and red striped' }, banana: 'purple' }));
```
chancancode: mutation

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+    this.assertText('[yellow and red striped] [yellow and red striped]');
+
+    this.runTask(() => set(this.context, 'colors.apple.gala', 'yellow-redish'));
+    this.assertText('[yellow-redish] [yellow-redish]');
```
chancancode: Either add a "reset with replacement" step after this or change this to test that

packages/ember-glimmer/tests/integration/helpers/get-test.js
chancancode: :+1: I agree. I think the `GetStream` is just referring to the nested get helper. Maybe just "should be able to use the result of a `get` helper as the key"?

packages/ember-htmlbars/tests/helpers/get_test.js
```diff
@@ -1,495 +0,0 @@
...
-  equal(view.$().text(), '[yellow and red striped] [yellow and red striped]', 'should return \'yellow and red striped\' for {{get colors \'apple.gala\'}}');
-
-  run(function() {
-    view.set('context.colors.apple.gala', 'yellow-redish');
-  });
-
-  equal(view.$().text(), '[yellow-redish] [yellow-redish]', 'should return \'yellow-redish\' for {{get colors \'apple.gala\'}}');
-});
-
-QUnit.test('should be able to get an object value with a GetStream key', function() {
```
chancancode: :heavy_check_mark: should be able to get an object value with a GetStream key

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+      colors: { apple: 'red', banana: 'yellow' },
+      key: 'key1',
+      possibleKeys: { key1: 'apple', key2: 'banana' }
+    });
+    this.assertText('[red] [red]');
+
+    this.runTask(() => set(this.context, 'key', 'key2'));
+    this.assertText('[yellow] [yellow]');
+
+    this.runTask(() => set(this.context, 'colors', { apple: 'green', banana: 'purple' }));
```
chancancode: mutation(s)

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+    this.assertText('[purple] [purple]');
+
+    this.runTask(() => set(this.context, 'key', 'key1'));
+    this.assertText('[green] [green]');
+
+    this.runTask(() => set(this.context, 'colors.apple', 'red'));
```
chancancode: replacement

packages/ember-glimmer/tests/integration/helpers/get-test.js
chancancode: :+1: see above

packages/ember-htmlbars/tests/helpers/get_test.js
```diff
@@ -1,495 +0,0 @@
...
-  equal(view.$().text(), '[green] [green]', 'should return \'green\'');
-
-  run(function() {
-    view.set('context.colors.apple', 'red');
-  });
-
-  equal(view.$().text(), '[red] [red]', 'should return \'red\'');
-});
-
-QUnit.test('should be able to get an object value with a GetStream value and bound/dynamic key', function() {
```
chancancode: :heavy_check_mark: should be able to get an object value with a GetStream value and bound/dynamic key

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+    this.assertText('[red] [red]');
+
+    this.runTask(() => set(this.context, 'key', 'banana'));
+    this.assertText('[yellow] [yellow]');
+
+    this.runTask(() => set(this.context, 'objectKey', 'colors2'));
+    this.assertText('[purple] [purple]');
+
+    this.runTask(() => set(this.context, 'objectKey', 'colors1'));
+    this.assertText('[yellow] [yellow]');
```
chancancode: Missing reset (set `key` back to `apple`)

packages/ember-glimmer/tests/integration/helpers/get-test.js
chancancode: :+1: see above

packages/ember-htmlbars/tests/helpers/get_test.js
```diff
@@ -1,495 +0,0 @@
...
-  equal(view.$().text(), '[purple] [purple]', 'should return \'purple\'');
-
-  run(function() {
-    view.set('context.objectKey', 'colors1');
-  });
-
-  equal(view.$().text(), '[yellow] [yellow]', 'should return \'yellow\'');
-});
-
-QUnit.test('should be able to get an object value with a GetStream value and GetStream key', function() {
```
chancancode: :heavy_check_mark: should be able to get an object value with a GetStream value and GetStream key

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+    this.assertText('[red] [red]');
+
+    this.runTask(() => set(this.context, 'key', 'key2'));
+    this.assertText('[yellow] [yellow]');
+
+    this.runTask(() => set(this.context, 'objectKey', 'colors2'));
+    this.assertText('[purple] [purple]');
+
+    this.runTask(() => set(this.context, 'objectKey', 'colors1'));
+    this.assertText('[yellow] [yellow]');
```
chancancode: add reset

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+  }
+
+  ['@test should yield `get`']() {
```
chancancode: "@test the result of a get helper can be yielded" ?

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+  }
+
+  ['@test should be able to get an object value with a bound/dynamic key']() {
+    this.render(`[{{get colors key}}] [{{if true (get colors key)}}]`, {
+      colors: { apple: 'red', banana: 'yellow' },
+      key: 'apple'
+    });
+    this.assertText('[red] [red]');
+
+    this.runTask(() => set(this.context, 'key', 'banana'));
```
chancancode: Add a test for no-op re-render before this

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+          gala: 'red and yellow',
+          mcintosh: 'red'
+        },
+        banana: 'yellow'
+      },
+      key: 'apple.gala'
+    });
+    this.assertText('[red and yellow] [red and yellow]');
+
+    this.runTask(() => set(this.context, 'key', 'apple.mcintosh'));
```
chancancode: Add a test for no-op re-render before this

packages/ember-glimmer/tests/integration/helpers/get-test.js
chancancode: Add a test for no-op re-render before this

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+
+  ['@test should be able to get an object value with a GetStream key']() {
+    this.render(`[{{get colors (get possibleKeys key)}}] [{{if true (get colors (get possibleKeys key))}}]`, {
+      colors: { apple: 'red', banana: 'yellow' },
+      key: 'key1',
+      possibleKeys: { key1: 'apple', key2: 'banana' }
+    });
+    this.assertText('[red] [red]');
+
+    this.runTask(() => set(this.context, 'key', 'key2'));
```
chancancode: Add a test for no-op re-render before this

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+      possibleValues: {
+        colors1: { apple: 'red', banana: 'yellow' },
+        colors2: { apple: 'green', banana: 'purple' }
+      },
+      objectKey: 'colors1',
+      key: 'apple'
+    });
+    this.assertText('[red] [red]');
+
+    this.runTask(() => set(this.context, 'objectKey', 'colors2'));
```
chancancode: Add a test for no-op re-render before this

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+      objectKey: 'colors1',
+      possibleKeys: {
+        key1: 'apple',
+        key2: 'banana'
+      },
+      key: 'key1'
+    });
+    this.assertText('[red] [red]');
+
+    this.runTask(() => set(this.context, 'objectKey', 'colors2'));
```
chancancode: Add a test for no-op re-render before this

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+        this.mcintosh = 'red';
+      }
+    });
+
+    this.registerComponent('foo-bar', {
+      ComponentClass: FooBarComponent,
+      template: `{{yield (get colors mcintosh)}}`
+    });
+
+    this.render(`{{#foo-bar colors=colors as |value|}}{{value}}{{/foo-bar}}`, {
```
chancancode: To test the `get()` method on the reference, you would need to do `{{value.foo}}`

Actually, if that's the goal, there is a simpler test that doesn't require a component:

```
{{#with (get foo bar) as |baz|}}
  {{baz.bat}}
{{/with}}
```

The baz in `{{baz.bat}}` would be a `GetReference`, and `{{baz.bat}}` would call `bazGetReference.get('bat')`.

packages/ember-glimmer/lib/utils/references.js
```diff
@@ -53,6 +53,25 @@ export class UpdatableReference extends RootReference {
   }
 }
 
+export class GetHelperReference {
```
chancancode: should probably say `// @implements PathReference` above here

packages/ember-glimmer/lib/utils/references.js
```diff
@@ -53,6 +53,25 @@ export class UpdatableReference extends RootReference {
...
+  constructor(sourceReference, pathReference) {
+    this.sourceReference = sourceReference;
+    this.pathReference = pathReference;
+  }
+
+  isDirty() { return true; }
+
+  value() {
+    return this.sourceReference.get(this.pathReference.value()).value();
```
chancancode: Since we are returning a value immediately, there is no need to go through `get` on the source reference. Something like your original implementation...

```
let object = this.sourceReference.value();
let path = this.pathReference.value();

return get(object, path);
```

...should be enough (might need some null guards here and there). In general, it is only worth it to produce a reference (`.get` returns a new `PathReference`) if you are going to hold onto it for and use it again in the future. The point of a reference is that you can save it off and query it over and over again to get the latest value.

chancancode: Probably need to be a `object && get(object, path)` to pass the null test.

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+    this.assertText('bus');
+
+    this.runTask(() => {
+      set(fooBarInstance, 'mcintosh', 'red');
+      set(this.context, 'colors', { red: 'banana' });
+    });
+    this.assertText('banana');
+  }
+
+  ['@htmlbars should handle object values as nulls']() {
```
chancancode: I *think* https://github.com/emberjs/ember.js/pull/13173/files#r58126138 should fix it?

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+    this.render(`[{{get colors 'apple'}}] [{{if true (get colors 'apple')}}]`, {
+      colors: null
+    });
+    this.assertText('[] []');
+
+    this.runTask(() => set(this.context, 'colors', { apple: 'green', banana: 'purple' }));
```
chancancode: add rerender before this

chancancode: also, prefer mutation here

packages/ember-htmlbars/tests/helpers/get_test.js
```diff
@@ -1,495 +0,0 @@
...
-  equal(view.$().text(), '[purple] [purple]', 'should return \'purple\'');
-
-  run(function() {
-    view.set('context.objectKey', 'colors1');
-  });
-
-  equal(view.$().text(), '[yellow] [yellow]', 'should return \'yellow\'');
-});
-
-QUnit.test('should handle object values as nulls', function() {
```
chancancode: :heavy_check_mark: should handle object values as nulls

packages/ember-htmlbars/tests/helpers/get_test.js
```diff
@@ -1,495 +0,0 @@
...
-  equal(view.$().text(), '[green] [green]', 'should return \'green\' for {{get colors \'apple\'}} (colors = { apple: \'green\', banana: \'purple\' })');
-
-  run(function() {
-    view.set('context.colors', null);
-  });
-
-  equal(view.$().text(), '[] []', 'should return \'\' for {{get colors \'apple\'}} (colors = null)');
-});
-
-QUnit.test('should handle object keys as nulls', function() {
```
chancancode: This test appears to be missing :money_with_wings:

packages/ember-htmlbars/tests/helpers/get_test.js
```diff
@@ -1,495 +0,0 @@
...
-  equal(view.$().text(), '[yellow] [yellow]', 'should return \'yellow\' for {{get colors key}} (key = \'banana\')');
-
-  run(function() {
-    view.set('context.key', null);
-  });
-
-  equal(view.$().text(), '[] []', 'should return \'\' for {{get colors key}}  (key = null)');
-});
-
-QUnit.test('should handle object values and keys as nulls', function() {
```
chancancode: :heavy_check_mark: should handle object values and keys as nulls

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+    this.runTask(() => set(this.context, 'colors', null));
+    this.assertText('[] []');
+  }
+
+  ['@htmlbars should handle object values and keys as nulls']() {
+    this.render(`[{{get colors 'apple'}}] [{{if true (get colors key)}}]`, {
+      colors: null,
+      key: null
+    });
+    this.assertText('[] []');
```
chancancode: should probably do the I-N-U-R steps here to test that it can go *from* null to something else (and back)

packages/ember-htmlbars/tests/helpers/get_test.js
```diff
@@ -1,495 +0,0 @@
...
-    context: context,
-    template: compile('[{{get colors key}}] [{{if true (get colors key)}}]')
-  });
-
-  runAppend(view);
-
-  equal(view.$().text(), '[] []', 'should return \'\' for {{get colors key}}  (colors=null, key = null)');
-});
-
-QUnit.test('get helper value should be updatable using {{input}} and (mut) - dynamic key', function() {
```
chancancode: :heavy_check_mark: get helper value should be updatable using {{input}} and (mut) - dynamic key

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+  }
+
+  ['@htmlbars get helper value should be updatable using {{input}} and (mut) - dynamic key']() {
+    this.registerComponent('-text-field', { ComponentClass: TextField });
```
chancancode: :scissors: this should not be necessary, it's probably broken on Glimmer at the moment, but you already marked it as `@htmlbars`

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+
+    this.render(`{{input type='text' value=(mut (get source key)) id='get-input'}}`, {
+      source: EmberObject.create({
+        banana: 'banana'
+      }),
```
chancancode: Don't need `EmberObject` here, pojo is :ok_hand:

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+      key: 'banana'
+    });
+
+    this.assert.strictEqual(this.$('#get-input').val(), 'banana');
```
chancancode: I think we general prefer

```
['@test ...'](assert) {
  ...
  assert.strictEqual(...);
  ...
}
```

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+
+    this.runTask(() => set(this.context, 'source.banana', 'yellow'));
```
chancancode: Add re-render before this

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+
+      // have tried to fire _elementValueDidChange via events but no luck
+      // so looks like the only option is to call _elementValueDidChange manually
+      // which I suspect doesn't seem to work for glimmer once it's implemented
+      //
+      // this.$('#get-input').trigger('input');
+      // this.$('#get-input').trigger('paste');
+      // this.$('#get-input').trigger('change');
+      // this.$('#get-input').triggerHandler('input');
+      this.component.childViews[0]._elementValueDidChange();
```
chancancode: It's fine :+1: we can deal with it when we enable this test

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+    });
+
+    this.assert.strictEqual(this.$('#get-input').val(), 'some value');
+    this.assert.strictEqual(get(this.context, 'source.banana'), 'some value');
+  }
+
+  ['@htmlbars get helper value should be updatable using {{input}} and (mut) - dynamic nested key']() {
+    this.registerComponent('-text-field', { ComponentClass: TextField });
```
chancancode: :scissors:

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+
+    this.render(`{{input type='text' value=(mut (get source key)) id='get-input'}}`, {
+      source: EmberObject.create({
```
chancancode: :scissors: `EmberObject`

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+      this.$('#get-input').val('some value');
+      this.component.childViews[0]._elementValueDidChange();
+    });
+
+    this.assert.strictEqual(this.$('#get-input').val(), 'some value');
+    this.assert.strictEqual(get(this.context, 'source.apple.mcintosh'), 'some value');
+  }
+
+  ['@htmlbars get helper value should be updatable using {{input}} and (mut) - static key']() {
+    this.registerComponent('-text-field', { ComponentClass: TextField });
```
chancancode: :scissors:

packages/ember-glimmer/tests/integration/helpers/get-test.js
```diff
@@ -0,0 +1,337 @@
...
+
+    this.render(`{{input type='text' value=(mut (get source 'banana')) id='get-input'}}`, {
+      source: EmberObject.create({
```
chancancode: :scissors: `EmberObject`

packages/ember-htmlbars/tests/helpers/get_test.js
```diff
@@ -1,495 +0,0 @@
...
-  run(function() {
-    view.$('#get-input').val('some value');
-    view.childViews[0]._elementValueDidChange();
-  });
-
-  equal(view.$('#get-input').val(), 'some value');
-  equal(view.get('context.source.banana'), 'some value');
-});
-
-QUnit.test('get helper value should be updatable using {{input}} and (mut) - dynamic nested key', function() {
```
chancancode: :heavy_check_mark: get helper value should be updatable using {{input}} and (mut) - dynamic nested key

packages/ember-htmlbars/tests/helpers/get_test.js
```diff
@@ -1,495 +0,0 @@
...
-  run(function() {
-    view.$('#get-input').val('some value');
-    view.childViews[0]._elementValueDidChange();
-  });
-
-  equal(view.$('#get-input').val(), 'some value');
-  equal(view.get('context.source.apple.mcintosh'), 'some value');
-});
-
-QUnit.test('get helper value should be updatable using {{input}} and (mut) - static key', function() {
```
chancancode: :heavy_check_mark: get helper value should be updatable using {{input}} and (mut) - static key

chancancode:
@ro0gr sorry for the delay! I left some comments. Looks good to me otherwise!

chancancode:
Ah, one more thing – can you rebase and squash?

packages/ember-glimmer/lib/utils/references.js
wycats: Just chatted with @chancancode. I don't think we can use `Ember.get`, because `Ember.get` is not **always** 1:1 with `PathReference#get`. The most notable example of this today is unbound references, which cache a constant reference.

For now, going through the proper interface seems fine.

ro0gr:
Thank you @wycats for merging this work and sorry again for such a long progress on this.

Anyway I've walked through your comments @chancancode. There were so much small issues. Hope I've fixed all of them in the latest commit.

krisselden:
Sorry we dropped the ball linking that this PR was superseded by #13264 which addressed the above comments and was merged yesterday.

ro0gr:
@krisselden I've seen this @wycats pr. 
Actually there are some unaddressed issues wich are fixed here:
- add bunch of missing I-N-U-R steps according to @chancancode comments
- add missing [@htmlbars should handle object keys as nulls](https://github.com/emberjs/ember.js/pull/13173/commits/bd03f06aac2363acaea9995280c1b8101a99bcd8#diff-350c933360e807598f74de6573a7fcaaR332)

ro0gr:
Should it be delivered via a new pr or left as it is?

krisselden:
@ro0gr thank you

ro0gr:
@krisselden and thank you :smile:

Issue #13061, opened 2016-03-07
yield & component helper => two way data binding not working ?


plcarmel:
DEBUG: -------------------------------
DEBUG: Ember      : 2.4.1
DEBUG: Ember Data : 2.4.0
DEBUG: jQuery     : 2.2.1
DEBUG: -------------------------------

I don't know if this is a bug but it seems to contradict the documentation at
[http://emberjs.com/api/classes/Ember.Templates.helpers.html#toc_nested-usage]

This is the example that is given:
`{{yield (hash nameInput=(component "my-input-component" value=model.name placeholder="First Name"))}}`

I tried it but without being able to achieve two way data binding.

## To replicate the problem ##

templates/components/test-binding.hbs:
`{{yield (component 'my-input' value=value)}}`

templates/components/my-input.hbs:
`{{input value=value}}`

templates/application.hbs:
```
{{#test-binding value=value as |component|}}
  {{component component}}
{{/test-binding}}

value: {{value}}
```

## A work-around ##

It works it the value is encapsulated inside a hash.  Please note that the way I do it differs from the example in the documentation, which is not working.

templates/components/test-binding-2.hbs:
`{{yield (component 'my-input-2' model=model)}}`

templates/components/my-input-2.hbs:
`{{input value=model.value}}`

templates/application.hbs:
```
{{#test-binding-2 model=model as |component|}}
  {{component component}}
{{/test-binding-2}}

model.value: {{model.value}}
```

controllers/application.js:
```
import Ember from 'ember';
export default Ember.Controller.extend({
  model: { value: 'Hello world' }
});
```

GCorbel:
Can you try simply : 

```
{{#test-binding value=value as |component|}}
  {{component}}
{{/test-binding}}

value: {{value}}
```

I find `{{component component}}` a bit weird.

plcarmel:
From the same documentation:

"Contextual components are created using the nested form of the component helper, and may be passed attrs and positional params. Contextual components **must be invoked with a . in their path, unless they are being passed to the invoking {{component helper.**"

pixelhandler:
@plcarmel can you create an ember-twiddle example that demonstrate the issue? Also I'm curious if 2-way binding is supposed to function, the docs mention binding, but not 2-way.

plcarmel:
In the example given, a (custom) input component is used. So, it is implicit that two way binding should work in that example. There is no action registered on the component, no one-way data binding would render the component useless.

[...]

PR #13177, opened 2016-03-24
[BUGFIX release] Add mut to attributes in contextual components

Serabe:
As explained in issue #13061, given a component `change-button` with template:

``` hbs
<button {{action (action (mut val) 10)}}>Change to 10</button>
```

If it is invoked as in the following template:

``` hbs
{{change-button val=value}}
<span class="value">
  {{value}}
</span>
```

Clicking on the button will cause `.value` to contain `10`.

On the other hand, invoking it via a closure component:

``` hbs
{{component (component "change-button" val=value)}}
<span class="value">
  {{value}}
</span>
```

Clicking on the button will not change the text in `.value`.

This PR adds a transform that adds `mut` to params and hash values.


rwjblue:
I restarted the sauce labs tests, but can you also update the prefix to `[BUGFIX release]` (I think we will want this to land into the LTS)?

Serabe:
Done!

rwjblue:
OK, this looks good to me.  I'm going to land into canary and beta for some testing, but will backport into the LTS before shipping (assuming all is good).

rwjblue:
Thank you @Serabe!

Serabe:
Thanks. If you see any report that might be related to this, just mention
me, please.

rwjblue:
Included in v2.4.4.

PR #13180, opened 2016-03-25
[DOC release] remove misleading statement

GavinJoyce:
The `Ember build tools will remove any calls to Ember.assert() when doing a production build` statement is pretty misleading IMO. I expected that ember-cli would strip them from my app, which [isn't the case](https://github.com/ember-cli/ember-cli/issues/5680#issuecomment-201142739).


mixonic:
@GavinJoyce can you just edit this to clarify assertion lines are stripped from Ember.js production builds, and make no-op for assertions in application code?

GavinJoyce:
@mixonic updated, thanks

mixonic:
:clap:

PR #13181, opened 2016-03-25
[BUGFIX] Context not set when only queryParams

### LINKED ISSUE #13150: QueryParams refreshModel does not work on route without dynamic segments
I am working on a index route with queryParams that control paging and sorting of table data. When changing the queryParams with the back button the queryParams on the controller are changed but the model is not refreshed.
On a similar page with detail data the refreshModel property on the queryParams are working as expected.
After some testing I found that adding a dynamic seqment (/:id) to my index route solves my problems with the refreshModel. But adding unused dynamic segments to my routes is not a solution that will be accepted in production.
I will try to find the root cause of the problem, but if someone with more knowledge on how the router works can help me that will be great.

Wimmen:
sawParams is not set to true when all params are queryParams.
Fixes #13150

### DIFF
--- a/packages/ember-routing/lib/system/route.js
+++ b/packages/ember-routing/lib/system/route.js
@@ -1466,6 +1466,7 @@ var Route = EmberObject.extend(ActionHandler, Evented, {
     var queryParams = get(this, '_qp.map');
 
     for (var prop in params) {
+      sawParams = true;
       if (prop === 'queryParams' || (queryParams && prop in queryParams)) {
         continue;
       }
@@ -1474,7 +1475,6 @@ var Route = EmberObject.extend(ActionHandler, Evented, {
         name = match[1];
         value = params[prop];
       }
-      sawParams = true;
     }
 
     if (!name && sawParams) {
--- a/packages/ember-routing/lib/system/route.js
+++ b/packages/ember-routing/lib/system/route.js
@@ -1462,11 +1462,11 @@ var Route = EmberObject.extend(ActionHandler, Evented, {
     @public
   */
   model(params, transition) {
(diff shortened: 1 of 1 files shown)


rwjblue:
This has a failing test:

> Routing with Query Params: A child of a resource route still defaults to parent route's model even if the child route has a query param

---

Also, this will need a new test to be added for us to land the fix.

WimAtIHomer:
Changed the pull request to pass al existing tests, also tried to make a failing test on the original code but failed to do so. The call to the function model comes from the function deserialize, which is used somewhere deep in ember.js. To deep for me to understand at this moment.

rwjblue:
I think you should be able to model a test demonstrating the original issue.  I think something like [this one](https://github.com/emberjs/ember.js/blob/49cabd9ad2d539e91ba90a5b96f397c19a3e561c/packages/ember/tests/routing/query_params_test.js#L2417-L2464) is close.  It is using `refreshModel` on an index route already, but I'm unclear of the circumstances that cause the error you described in https://github.com/emberjs/ember.js/issues/13150.

WimAtIHomer:
As I mentioned in #13150 I have made a small project demonstration the issue at https://github.com/WimAtIHomer/queryParams . The issue occurs using the history back function in combination with queryParams that control the paging of a list. Not sure if this can be done in a test case.

WimAtIHomer:
As mentioned in #13150 I solved my problems by improving my code.

mmun:
@WimAtIHomer Thanks for the work that you put into the issue and for following up.

@rwjblue Does this patch look correct to you? Are we just trying to figure out a reproduction case?

Issue #13182, opened 2016-03-25
{{render}} helper no longer able to bubble actions without legacy controller addon.


rwjblue:
Given the following example:

```hbs
{{! index template}}
{{render 'some-name' someModel}}
```

```hbs
<p>{{this}} - {{this.target}} - {{this.parentController}}</p>

<button {{action 'thingInApplicationRoute'}}>
  Should hit route!
</button>
```

In Ember 2.0 and 2.1, that action would bubble properly to the router (and then on to the application route).  In Ember 2.2 and above this bubbling behavior is no longer present, unless the `ember-legacy-controller` addon is in use.

The root cause is that we no longer create the `controller` template local unless that addon is present (this was done in https://github.com/emberjs/ember.js/pull/12289 which landed in 2.2), however the `{{render` helper still grabs the `controller` local to specify `target` for the newly created controller instance (the `controller` local is looked up [here](https://github.com/emberjs/ember.js/blob/9c64558177e02b5932f73c90f46e36c16362c78d/packages/ember-routing-htmlbars/lib/keywords/render.js#L188) and passed as the target [here](https://github.com/emberjs/ember.js/blob/9c64558177e02b5932f73c90f46e36c16362c78d/packages/ember-routing-htmlbars/lib/keywords/render.js#L199) and [here](https://github.com/emberjs/ember.js/blob/9c64558177e02b5932f73c90f46e36c16362c78d/packages/ember-routing-htmlbars/lib/keywords/render.js#L208)).  

[2.1 demo](http://rwjblue.jsbin.com/mihirak/edit?html,js,output)
[2.2 demo](http://rwjblue.jsbin.com/begegi/edit?html,js,output)
[2.3 demo](http://rwjblue.jsbin.com/hisuxo/edit?html,js,output)
[2.4 demo](http://rwjblue.jsbin.com/wovasa/edit?html,js,output)

Working [2.4 demo](http://rwjblue.jsbin.com/xuyeko/edit?html,js,output) with the legacy controller flag enabled.

---

The fix is to fallback to the [previously looked up](https://github.com/emberjs/ember.js/blob/9c64558177e02b5932f73c90f46e36c16362c78d/packages/ember-routing-htmlbars/lib/keywords/render.js#L131) router instance if the `parentController` variable is `undefined`.  PR will be incoming shortly....

rwjblue:
Updated with a bunch of demos.

PR #13185, opened 2016-03-25
[BUGFIX release] Ensure `{{render}}` sets up target properly.

rwjblue:
When the _ENABLE_LEGACY_CONTROLLER_SUPPORT flag is disabled, the `controller` local is `undefined`. This causes the controller created by `{{render}}` to have an undefined `target` (so it swallows any actions triggered).

This properly sets the `target` to `router:main` if there is no `parentController` present.

Fixes https://github.com/emberjs/ember.js/issues/13182.


rwjblue:
Thank you!

rwjblue:
Included in v2.4.4.

PR #13186, opened 2016-03-25
Remove `latest` from `chrome` sauce labs tests.

rwjblue:
This is erroring for all PR's saying that `latest` is not a valid
version. We need to dig in and/or report an issue to the Sauce Labs
folks, but this should gets us back to a passing build.


rwjblue:
@johanneswuerbach - FYI - I'm not sure what the right steps are here to report this...

PR #13188, opened 2016-03-26
[WIP] Remove Ember.ContainerView.

rwjblue:
I started working on this a few days ago, but haven't had a chance to circle around to it.


rwjblue:
@HeroicEric - Want another challenge?

HeroicEric:
heh, I actually tried this one already the other day but got stuck. I can try though :smile:

packages/ember-views/tests/views/view/remove_test.js
```diff
@@ -10,44 +9,8 @@ import viewKeyword from 'ember-htmlbars/keywords/view';
 import { objectAt } from 'ember-runtime/mixins/array';
 
 
-// .......................................................
-// removeChild()
```
HeroicEric: @rwjblue was this meant to be removed? If so, should `View#removeAllChildren` and friends have the same fate?

HeroicEric: @rwjblue was this meant to be removed? If so, should `View#removeAllChildren` and friends have the same fate?

rwjblue: I was uncertain if removeAllChildren is a "normal" view method or just for container view.

HeroicEric: @rwjblue `removeAllChildren` is from the `LegacyViewSupport` mixin but it uses `removeChild` https://github.com/rwjblue/ember.js/blob/remove-container-view/packages/ember-views/lib/mixins/legacy_view_support.js#L35-L39

rwjblue: kk, seems good lets kill it for now and we can see where that puts us

PR #13189, opened 2016-03-26
[Glimmer2] Migrate and Remove block params test

Joelkang:
Remove `should raise error if helper not available`: [custom-helpers-test.js:287](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/tests/integration/helpers/custom-helper-test.js#L287) already tests this
Remove `basic block params usage`: [syntax/with-test.js:21](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/tests/integration/syntax/with-test.js#L21) already tests this

Keeping `components can yield values`: Even though [helpers/hash-test.js](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/tests/integration/helpers/hash-test.js#L76) already tests yielding hashes, I added a test to components/curly-components.js to test yielding non-hash properties

Part of  #13127

cc @chancancode


packages/ember-glimmer/tests/integration/syntax/with-test.js
```diff
@@ -292,4 +292,43 @@ moduleFor('Syntax test: Multiple {{#with as}} helpers', class extends RenderingT
     this.assertText('Los Pivots');
   }
 
+  ['@test nested {{#with}} blocks should have access to root context']() {
+    this.render(`{{name}}{{#with committer1.name as |name|}}[{{name}}{{#with committer2.name as |name|}}[{{name}}]{{/with}}{{name}}]{{/with}}{{name}}{{#with committer2.name as |name|}}[{{name}}{{#with committer1.name as |name|}}[{{name}}]{{/with}}{{name}}]{{/with}}{{name}}`,
```
mmun: We've been using the `strip` helper to keep things readable. It removes all leading whitespace and newlines. For example, check out https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/tests/integration/syntax/each-in-test.js#L110-L116.

chadhietala:
Thanks @Joelkang! Looks good to me.

Joelkang:
Rebased, squashed and added the an R in INUR for the infinite loop test

chancancode:
@Joelkang looks like we lost the removals from the rebase :smile:

Joelkang:
herp derp. fixed

PR #13190, opened 2016-03-27
Begin making it possible to boot apps with ember-glimmer.

rwjblue:


rwjblue:
Obviously, there is a lot more work to be done (all of these disabled tests need to pass), but once this PR lands it is possible to boot an ember-cli app.

Issue #12360, opened 2015-09-17
Docs for .@each and .[] are unclear about use-cases


jamesarosen:
The [docs for `.@each` and `.[]` properties](http://emberjs.com/api/classes/Ember.Array.html#property__each) aren't clear that `.@each` should *never* be used as a terminal and `.[]` should *only* be used as a terminal.

See https://github.com/emberjs/ember.js/pull/11994#issuecomment-131879238

stefanpenner:
:+1:

kellyselden:
I have this bookmarked for whenever I need a reminder.

https://github.com/emberjs/ember.js/issues/2016#issuecomment-14402033

pixelhandler:
@jamesarosen true that will be helpful to clarify :)

rwjblue:
Closed by https://github.com/emberjs/ember.js/commit/e0e8f7feec456587ecfb13cce81fa3c99c6d8706, and a fix by @krisselden (which I can't find the link for ATM) that makes terminal @each expand to .[].

PR #13198, opened 2016-03-28
[DOC] Add more guidance on using @each rather than []

abepetrillo:
Updated docs in response to https://github.com/emberjs/ember.js/issues/12360

@each should not be used as a leaf node.

### DIFF
--- a/packages/ember-runtime/lib/mixins/array.js
+++ b/packages/ember-runtime/lib/mixins/array.js
@@ -540,8 +540,7 @@ export default Mixin.create(Enumerable, {
     return an enumerable that maps automatically to the named key on the
     member objects.
 
-    @each should only be used in a non-terminal context, and is deprecated when
-    used as a leaf node. Example:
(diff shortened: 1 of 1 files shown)


packages/ember-runtime/lib/mixins/array.js
```diff
@@ -540,8 +540,23 @@ export default Mixin.create(Enumerable, {
     return an enumerable that maps automatically to the named key on the
     member objects.
 
-    If you merely want to watch for any items being added or removed to the array,
-    use the `[]` property instead of `@each`.
+    @each should only be used in a non-terminal context, and is deprecated when
```
rwjblue: This isn't deprecated any more :crying_cat_face:, can you remove mention of the deprecation? (all of other changes look great!)

abepetrillo:
@rwjblue updated

rwjblue:
Thank you!

PR #13200, opened 2016-03-28
Newer Chrome versions support `new Error().stack`. Prefer it when available.

### LINKED ISSUE #12266: deprecation code doesn't need to throw Errors on all browsers
new chrome works fine with `new Error().stack`

something like the following should work:

```js
var captureStack = new Error().stack ? function() {
  return new Error().stack;
} : function() {
  try { throw new Error(); } catch(e) { return e; }
}
```

cc @teddyzeenny

yaymukund:
Closes #12266

### DIFF
--- a/packages/ember-debug/lib/deprecate.js
+++ b/packages/ember-debug/lib/deprecate.js
@@ -30,13 +30,23 @@ registerHandler(function logDeprecationToConsole(message, options) {
   Logger.warn('DEPRECATION: ' + updatedMessage);
 });
 
+let captureErrorForStack;
+
+if (new Error().stack) {
+  captureErrorForStack = function() {
+    return new Error();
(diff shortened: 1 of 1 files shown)


yaymukund:
:crying_cat_face: ie9

yaymukund:
I removed my test because it seems as if IE9 doesn't currently print a stack trace. This is inconsistent with all other browsers. See the screenshots in saucelabs for an example of IE9's output:

https://saucelabs.com/jobs/b33e64ff69ae403880bdadbbf8f3437f
(e.g. https://assets.saucelabs.com/jobs/b33e64ff69ae403880bdadbbf8f3437f/0033screenshot.png )

rwjblue:
Thank you for keeping at it!

Issue #11992, opened 2015-08-05
`didReceiveAttrs` requires `this._super` call.


mmelvin0:
When didReceiveAttrs fires on first render, this.get('value') === this.getAttr('value').
When didReceiveAttrs fires on re-renders, this.get('value') === (previous value) and this.getAttr('value') === (current value).

rwjblue:
* For which `Ember.VERSION`?
* Can you provide a JSBin / ember-twiddle reproduction?

mmelvin0:
1.13.5 and 1.13.6
http://ember-twiddle.com/7642c2bfe8a8868a3b5f

pixelhandler:
This page seems to be the only place that `didReceiveAttrs` is documented http://emberjs.com/blog/2015/06/12/ember-1-13-0-released.html#toc_component-lifecycle-hooks

I don't see `getAttr` listed in the API docs at http://emberjs.com/api/classes/Ember.Component.html

Does seem like a bug for `this.get('value')` though. However since `getAttr` and `didReceiveAttrs` don't seem to be documented in the api docs for 1.13.x I'm confused as to what I should expect. Seems that both the guides and api docs lack explanation of the lifecycle hooks. I'm curious to how these landed in the release without any docs (aside from a blog post announcing them). 

Maybe we're only supposed to use `getAttr` in `didRecieveAttrs` I see that the blog post mentions `this.get` can be expected to be accurate after `didInitAttrs` So in my opinion this does seem to be a BUG.

rwjblue:
@pixelhandler:

> Seems that both the guides and api docs lack explanation of the lifecycle hooks. I'm curious to how these landed in the release without any docs (aside from a blog post announcing them).

* We did miss adding some docs for features as they were added during the 1.13/glimmer work, but we are working on updated guides and API docs for 2.0/2.1 that will correct that mistake.  We have also discussed and decided that we will not be landing new features/API's without docs in the future (this was not a rule that existed during any of the 1.x series but we feel that it is very important for 2.x+).
* Usage of `getAttr`/`setAttr` and the various hooks isn't really "promoted" yet (but it is definitely important that they work), because significant portions of functionality are still not landed (likely some time in 2.1 and 2.2).

---

This is indeed a bug, `didReceiveAttrs` is firing (and the new attributes are set in `this.attrs`) prior to the values being proxied back to the root of the component.  Most likely this is an ordering problem, but I am not 100% sure just yet...

rwjblue:
[...]

Issue #12035, opened 2015-08-10
didInitAttrs cannot see bound values unless they are explicitly defined on the component class.


jakesjews:
I just noticed this in 1.13.7. If a property isn't explicitly defined on a component's class then its value is never seen in the didInitAttrs method even though it is bound in the template. Defining the property with a value of null does work but this wasn't required in 1.13.6. I'm not 100% sure if this is a bug or if the previous behavior was incorrect and it's been fixed now.

Ember-twiddle at http://ember-twiddle.com/62d39c8e4f57e0eb1fbb

aortbals:
I can confirm this bug. Immediately after upgrading, I am seeing `undefined` for an attribute passed to a component in `init` of the component:

```handlebars
{{foo-component bar='something'}}
```

```js
init() {
  Ember.assert("The {{foo-component}} attribute 'bar' was not configured", this.get('bar'));
},
```

The following assert throws in 1.13.7 where it would be correctly bound to `'something'` in 1.13.6. I can also confirm that defining the attribute on the component as `null` fixes the issue. In my experience, it is pretty common to see attributes passed to the component that aren't explicitly defined (regardless of what people feel about the practice of non-explicit attributes).

martndemus:
I bumped into this too. This was during `didReceiveAttrs`.
If you declare `foo: null` on the object things work again.

EricSchank:
I think I have a fix for this, but I'm writing a test to verify -- unless one of you has a breaking test already?

aortbals:
I do not. That's great to hear though!

noslouch:
don't have a breaking test, just a breaking app :-/

gerry3:
+1

rwjblue:
Fixed by #12071.

rwjblue:
Will be in a 1.13.8 soonish.

Issue #11571, opened 2015-06-27
Component attributes shadow properties


jamesarosen:
### JSBin

http://emberjs.jsbin.com/fuvozo/1/edit?html,js,output

### Summary

If a component takes in a property, `foo` and changes it (say, in `didReceiveAttrs`), then the component's template will see the original value.

OrKoN:
@jamesarosen is it the same/similar issue https://github.com/emberjs/ember.js/issues/11559 ?

rwjblue:
@OrKoN - Yes, they seem related.

rwjblue:
Submitted https://github.com/emberjs/ember.js/pull/11572 to address this.

Issue #11559, opened 2015-06-26
[REGRESSION] In Ember 1.13.0+ initial literal values of component attributes cannot be updated


OrKoN:
See jsbin http://emberjs.jsbin.com/likiqofaxa/1/edit?html,js,console,output

If you pass a literal as an attribute value, the component cannot update it:

```handlebars
    {{my-component test="initial-value"}}
    {{my-component}}
```

where MyComponent is

```
var MyComponent = Ember.Component.extend({
  test: null,
  init: function() {
    this._super();
    this.set('test', 'updated');
  }
});
```
The first does not work(`test` is not updated in the component template) and the second works as expected.

I am migrating from 1.12. jsbin http://emberjs.jsbin.com/tuhisokijo/1/edit?html,js,console,output

stefanpenner:
i would assume:

```js
this.set('test', this.get('test') || 'updated');
```

would be the way to deal with defaults in `init`

OrKoN:
@stefanpenner the problem is that one cannot override the defaults

rwjblue:
Submitted https://github.com/emberjs/ember.js/pull/11572 to address this.

OrKoN:
@rwjblue Thanks a lot! Will the fix appear in the 1.13.3 release? When may I expect it?

rwjblue:
Yes, it will be in 1.13.3 (should be out in a day or two).

OrKoN:
@rwjblue @stefanpenner This is happening again in Ember 2.1.0 (regression compared to 2.0.2) http://emberjs.jsbin.com/yinuqe/edit?html,js,console,output

rwjblue:
@OrKoN - Can you open a new issue?

OrKoN:
@rwjblue sure, opened a new one

PR #13203, opened 2016-03-28
[Glimmer2] [WIP] Migrate attrs lookup tests

Joelkang:
More or less migrated verbatim as per #13127

cc @chancancode 

I'm leaving out `should not need to call _super in didReceiveAttrs (GH #11992)` because it's not clear to me what the test is actually doing (or how it's testing against the issue). It does look like INUR should take care of the internals of this particular test case, but please advise if otherwise. 

Imo, if we really want to test that `_super` is not required, `_super` should be spied on (or clobbered with assertions?) I haven't spent enough time digging into the glimmer2 code, so I'm not even sure what the `_super` of didReceiveAttrs would actually be.


Joelkang:
Eh, CI failed because I moved `attrs-lookup-test.js` to under `components` in the `ember-glimmer` package since it made sense to me there, but obviously this fails for `ember-htmlbars`. @chancancode thoughts?

packages/ember-glimmer/tests/integration/components/attrs-lookup-test.js
```diff
@@ -0,0 +1,107 @@
...
+    this.runTask(() => set(this.context, 'firstAttr', 'second attr'));
+
+    assert.equal(instance.get('first'), 'second attr');
+
+    this.runTask(() => set(this.context, 'firstAttr', 'first attr'));
+
+    this.assertText('first attr');
+  }
+
+  ['@html should be able to modify a provided attr into local state #11571 / #11559'](assert) {
```
chadhietala: Is `@html` meant to be `@htmlbars`?

Joelkang: fixed

packages/ember-glimmer/tests/integration/components/attrs-lookup-test.js
```diff
@@ -0,0 +1,107 @@
+import { RenderingTest, moduleFor } from '../../utils/test-case';
+import { Component } from '../../utils/helpers';
+import { set } from 'ember-metal/property_set';
+
+moduleFor('Components test: attrs lookup', class extends RenderingTest {
```
chancancode: nit: new line between here and the first test

packages/ember-htmlbars/tests/integration/attrs_lookup_test.js
```diff
@@ -1,154 +0,0 @@
...
-  },
-
-  teardown() {
-    runDestroy(owner);
-    runDestroy(view);
-    owner = view = null;
-  }
-});
-
-QUnit.test('should be able to lookup attrs without `attrs.` - template access', function() {
```
chancancode: :heavy_check_mark: it should be able to lookup attrs without `attrs.` - template access

packages/ember-htmlbars/tests/integration/attrs_lookup_test.js
```diff
@@ -1,154 +0,0 @@
...
-    [OWNER]: owner,
-    template: compile('{{foo-bar first="first attr"}}')
-  }).create();
-
-  runAppend(view);
-
-  equal(view.$().text(), 'first attr');
-});
-
-QUnit.test('should be able to lookup attrs without `attrs.` - component access', function() {
```
chancancode: :heavy_check_mark: it should be able to lookup attrs without `attrs.` - component access

packages/ember-htmlbars/tests/integration/attrs_lookup_test.js
```diff
@@ -1,154 +0,0 @@
...
-    [OWNER]: owner,
-    template: compile('{{foo-bar first="first attr"}}')
-  }).create();
-
-  runAppend(view);
-
-  equal(component.get('first'), 'first attr');
-});
-
-QUnit.test('should be able to modify a provided attr into local state #11571 / #11559', function() {
```
chancancode: should be able to modify a provided attr into local state #11571 / #11559

packages/ember-glimmer/tests/integration/components/attrs-lookup-test.js
```diff
@@ -0,0 +1,107 @@
...
+        instance = this;
+      },
+
+      didReceiveAttrs() {
+        this.set('first', this.attrs.first.toUpperCase());
+      }
+    });
+    this.registerComponent('foo-bar', { ComponentClass: FooBarComponent, template: '{{first}}' });
+
+    this.render(`{{foo-bar first="first attr"}}`);
```
chancancode: Can you make the argument dynamic (`this.render(`{{foo-bar first=...}}`, { ... });`) so we can do the I-N-U-R tests correctly here?

packages/ember-htmlbars/tests/integration/attrs_lookup_test.js
```diff
@@ -1,154 +0,0 @@
...
-    template: compile('{{foo-bar first="first attr"}}')
-  }).create();
-
-  runAppend(view);
-
-  equal(view.$().text(), 'FIRST ATTR', 'template lookup uses local state');
-  equal(component.get('first'), 'FIRST ATTR', 'component lookup uses local state');
-});
-
-QUnit.test('should be able to access unspecified attr #12035', function() {
```
chancancode: :heavy_check_mark: should be able to access unspecified attr #12035

packages/ember-glimmer/tests/integration/components/attrs-lookup-test.js
```diff
@@ -0,0 +1,107 @@
...
+      init() {
+        this._super(...arguments);
+        instance = this;
+      },
+
+      didReceiveAttrs() {
+        assert.equal(this.get('woot'), 'yes', 'found attr in didReceiveAttrs');
+      }
+    });
+    this.registerComponent('foo-bar', { ComponentClass: FooBarComponent});
```
chancancode: Missing a space after `FooBarComponent` which is causing a jscs error

packages/ember-htmlbars/tests/integration/attrs_lookup_test.js
```diff
@@ -1,154 +0,0 @@
...
-    template: compile('{{foo-bar woot="yes"}}')
-  }).create();
-
-  runAppend(view);
-
-  // equal(view.$().text(), 'FIRST ATTR', 'template lookup uses local state');
-  equal(component.get('woot'), 'yes', 'component found attr');
-});
-
-QUnit.test('should not need to call _super in `didReceiveAttrs` (GH #11992)', function() {
```
chancancode: @Joelkang this is testing `getAttr` have the correct value (i.e. getAttr === get) in both initial render and update time, so we can port it as-is.

The mention of `_super` in the test name is not relevant – calling `this._super()` happen to fix this bug in the original buggy implementation, but you shouldn't have to. So feel free to rename it to something like "getAttr should ..."

packages/ember-htmlbars/tests/integration/attrs_lookup_test.js
```diff
@@ -1,154 +0,0 @@
-import compile from 'ember-template-compiler/system/compile';
```
chancancode: @Joelkang I don't fully understand your question: what was failing without this?

chancancode:
@Joelkang thank you for working on this (and sorry about the delay!). I left some comments inline.

packages/ember-htmlbars/tests/integration/attrs_lookup_test.js
Joelkang: Ahh ok, I think I must have spaced out over the `get` vs `getAttr` comparison. Will port.

packages/ember-glimmer/tests/integration/components/attrs-lookup-test.js
Joelkang: @chancancode My understanding is that once you set the local state, it no longer is a CP (or it somehow breaks the shadowing link) so i can't do the U part of the INUR. I actually tried this and it failed to change the value; the existing test doesn't actually update the value of the dynamic arg so this was never tested.

packages/ember-htmlbars/tests/integration/attrs_lookup_test.js
Joelkang: no worries, i rebased and it looks like the components folder was symlinked to the ember-htmlbars package so this is no longer an issue. Essentially the original test file was in `test/integration` while the new one is in `test/integration/components` so the relative import paths was failing for one but passing for the other.

packages/ember-glimmer/tests/integration/components/attrs-lookup-test.js
```diff
@@ -57,32 +58,44 @@ moduleFor('Components test: attrs lookup', class extends RenderingTest {
...
 
     this.runTask(() => this.rerender());
 
-    assert.equal(instance.get('first'), 'FIRST ATTR');
+    assert.equal(instance.get('first'), 'FIRST ATTR', 'component lookup uses local state during rerender');
+    this.assertText('FIRST ATTR');
+
+    this.runTask(() => set(this.context, 'firstAttr', 'second attr'));
+
+    assert.equal(instance.get('first'), 'SECOND ATTR', 'component lookup uses modified local state');
```
Joelkang: this was failing for me, `didRecieveAttrs()` wasn't getting called after the `set()` in the `runTask()` above, but I can't for the life of me figure out why Am pushing to see if the test fails in CI too.

Joelkang: looks like it's failing in CI too; @chancancode does anything look wrong here? I've stared at it for so long and it looks logically correct to me.

chadhietala: I think there is a legit bug here. I pulled this locally and placed the a currly on the outside of the component to make sure things were sane. It appears `didReceiveAttrs` is not actually being re-triggered.
screen shot 2016-04-07 at 10 49 32 pm

I then decided to see if this was a shadowing issue with the `attr` and local property being the same. The componet code changed to setting a local property called `local` and using it in the template like `{{local}}`.

```
let FooBarComponent = Component.extend({
  init() {
     this._super(...arguments);
     instance = this;
     this.local = null;
  },
  didReceiveAttrs() {
    this.set('local', this.getAttr('first').toUpperCase());
  }
});
```
This resulted in an even more bizarre output.
screen shot 2016-04-07 at 11 14 56 pm

So it started updating but then the attr was not uppercase. Adding @rwjblue since he may have more context since he added the original tests around this.

chadhietala: The weird thing is that this appears to work just fine in an app. https://ember-twiddle.com/74fe5a71ddb8d5336bd61516a2439618?fileTreeShown=false&openFiles=components.foo-bar.js%2Ctemplates.components.foo-bar.hbs

chancancode: This is indeed pretty weird, especially when it works in apps. I don't have time to investigate this right now, but I don't want to block this PR on that. Since this is new stuff we added (i.e. we are not loosing coverage), can you comment out the failing steps for now with a TODO comment pointing to this thread? We can revisit this when we enable this test for Glimmer – re-implementing it in Glimmer will probably teach us why.

LGTM otherwise!

Joelkang:
Commented out the failing test and squashed; thanks @chancancode!

PR #13209, opened 2016-03-28
Raise assertion when user overrides layout property on Component.

### LINKED ISSUE #13020: Ember 2.2 - Components without layouts not working (including link-to)
We're a bit behind, but I’m trying to update our app which is currently ember 2.1 and I’m running into an issue that is introduced in 2.2. 

I’m getting an error `Uncaught TypeError: Cannot read property 'fragmentReason' of undefined` thats triggered in the htmlbars `component-node-manager.js` file. 

I can get the app to render by adding empty templates for all my components that do not currently have a template but I know this cannot be the actual solution. The worst part is that this error is getting thrown for any instance of `link-to`, `input` and any other built in ember component as well.

It looks like this block of code from `component-node-manager.js` is returning a layout object even when it shouldn’t

```
   // If the component specifies its layout via the `layout` property
   // instead of using the template looked up in the container, get it
   // now that we have the component instance.
   if (!layout) {
      layout = _emberMetalProperty_get.get(component, 'layout');
   }
```

Then as a result, a few lines later when it should just return, it tries to dig into the above layout object which only contains some undefined properties and errors out because layout.meta is undefined

```
  if (!layout) {
    return;
  }

  var fragmentReason = layout.meta.fragmentReason;
```

Feels like I'm missing something really simple like a feature flag or something, especially since this has been out for months and I can't find any other issue close to this. Let me know what other info I might provide to help clear this up. Thanks!

Oladele:
Fixes #13020

### DIFF
--- a/packages/ember-htmlbars/lib/node-managers/component-node-manager.js
+++ b/packages/ember-htmlbars/lib/node-managers/component-node-manager.js
@@ -86,7 +86,13 @@ ComponentNodeManager.create = function ComponentNodeManager_create(renderNode, e
 
     if (!layout) { return; }
 
-    let fragmentReason = layout.meta.fragmentReason;
+    let fragmentReason;
+    if (layout.meta === undefined) {
+      assert('Meta key missing from layout. This usually happens when there is no template for a Component, and the Component has overrided the layout key. Please rename the property to something other than "layout".');
+    } else {
+      fragmentReason = layout.meta.fragmentReason;
+    }
+
     if (isAngleBracket && fragmentReason) {
       switch (fragmentReason.name) {
         case 'missing-wrapper':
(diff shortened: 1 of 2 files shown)


packages/ember-htmlbars/lib/node-managers/component-node-manager.js
```diff
@@ -86,7 +86,13 @@ ComponentNodeManager.create = function ComponentNodeManager_create(renderNode, e
 
     if (!layout) { return; }
 
-    let fragmentReason = layout.meta.fragmentReason;
+    let fragmentReason;
+    if (layout.meta === undefined) {
+      assert('Meta key missing from layout. This usually happens when there is no template for a Component, and the Component has overrided the layout key. Please rename the property to something other than "layout".');
+    } else {
+      fragmentReason = layout.meta.fragmentReason;
+    }
```
mixonic: Since asserts are stripped from prod builds, they should contain the test:

```js
    assert('Meta key missing from layout. This usually happens when there is no template for a Component, and the Component has overrided the layout key. Please rename the property to something other than "layout".', layout.meta !== undefined);
    let fragmentReason = layout.meta.fragmentReason;
```

mixonic:
@Oladele can you look at a rebase? This would be great to merge in :-) Apologies it sat so long.

locks:
Hi @Oladele! Are you able to update your PR?

locks:
Closing as no longer relevant, since `ember-htmlbars` is no longer used.

PR #13218, opened 2016-03-30
[BUGFIX RELEASE] Do not refresh routes on initial transition

### LINKED ISSUE #10945: Router `transition.retry` with `refreshModel` queryParams causes unexpected transition sequence
Please bare with me. This is kind of a hard issue to explain, but I do have a jsbin to help demonstrate it.

We discovered an issue when a user refreshes a page that has queryParams configured with `refreshModel` and there is a `transition.retry` involved. The scenario is someone deep-linking into the app with queryParams, but first being redirected to the login page.

It turns out that after logging in and calling `retry`, all of the resources/routes with dynamic segments have their `serialize` hooks called with `undefined` model values instead of calling the `model` hook with the params. Here is the internal transition log for the router, it shows entirely different transition paths: https://gist.github.com/workmanw/20a5a2a667f2bb9d27a7 .

The best way to demonstrate this issue is clicking the jsbin link below. You'll see it first redirects you to a login page, then a timer "simulates" login after 1 second, then it triggers the `transition.retry` that causes the `serialize` hook to be called with an undefined model. Then if you repeat the process without a queryParam, it works just fine.

With queryParam: http://emberjs.jsbin.com/gonadewaki#/post/1/comments?search=awesome
Without queryParam: http://emberjs.jsbin.com/gonadewaki#/post/1/comments

In my debugging, I discovered that if set `refreshModel` back to false or if didn't redirect to login, it works as expected. So this bug is a bit obscure.

Hopefully that is a clear explanation.

======

**EDIT** Here is a link to the same issue on canary for quick update testing: http://output.jsbin.com/humibo/1#/post/1/comments?search=awesome

raytiley:
Refreshing the route heighercy do to query param change on the initial transition is pointless, so now we only call refresh if there is a `currentState`. This fixes #10945

### DIFF
--- a/packages/ember-routing/lib/system/route.js
+++ b/packages/ember-routing/lib/system/route.js
@@ -759,7 +759,7 @@ var Route = EmberObject.extend(ActionHandler, Evented, {
       var totalChanged = Object.keys(changed).concat(Object.keys(removed));
       for (var i = 0, len = totalChanged.length; i < len; ++i) {
         var qp = qpMap[totalChanged[i]];
-        if (qp && get(this._optionsForQueryParam(qp), 'refreshModel')) {
+        if (qp && get(this._optionsForQueryParam(qp), 'refreshModel') && this.router.currentState) {
           this.refresh();
         }
       }
--- a/packages/ember/tests/routing/query_params_test.js
+++ b/packages/ember/tests/routing/query_params_test.js
@@ -378,6 +378,33 @@ if (isEnabled('ember-routing-route-configured-query-params')) {
     equal(indexModelCount, 2);
   });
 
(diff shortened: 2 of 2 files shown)


rwjblue:
LGTM

rwjblue:
Thanks @raytiley!

stefanpenner:
@mike-north does this fix the bug you found?

rwjblue:
Included in v2.4.4.

mike-north:
Possibly obsoletes https://github.com/stefanpenner/ember-redirect-to/blob/master/addon/mixins/redirect-to.js @stefanpenner

PR #13219, opened 2016-03-30
WIP improve _lazyInjections performance and eachComputedProperty.

krisselden:


krisselden:
There is some circular ref or missing dep that is breaking the only the node tests, not sure what is up with that yet.

stefanpenner:
sweet looks good

rwjblue:
@krisselden - Is this still needed?  We guarded all of the lazyInjection stuff to only run in debug builds, and I refactored `eachComputedProperty` in https://github.com/emberjs/ember.js/pull/13787 (still pending ATM).

PR #12666, opened 2015-12-01
[BUGFIX release] Prevent triggering V8 memory leak bug through registry / resolver access.

dgeb:
This fix changes the expectations of Registry to accept a `resolver` 
that’s an object with a `resolve` method instead of a straight function.

This allows us to avoid closing over access to a resolver object inside 
a function. It also allows us to avoid setting functions which shadow 
prototype functions unnecessarily.

Setting `Registry#resolver` to a function is still allowed but 
deprecated, regardless of whether it’s through a constructor option or 
via an instance override. The `resolver` function will be converted into
an object with a `resolve` method and will present a single deprecation 
warning on first access.

This fix also eliminates the need for application instances to override
their registry’s `normalizeFullName` and `makeToString` methods. 
Instead, the fallback registry will be checked when evaluating these
methods before returning the defaults. Again, this avoids the need to
override instance functions that shadow prototype functions.

[Fixes #12618]

/cc @rwjblue @stefanpenner

Issue #12618, opened 2015-11-16
Memory on acceptance tests on ember 1.13.10+


calderas:
test repo: https://github.com/calderas/leaky-mcleakerson

After upgrading our app to ember 1.13 chrome crashes when running Acceptance Tests

```
V8 error: Allocation failed - process out of memory (CALL_AND_RETRY_LAST).  Current memory usage: 1949 MB
```

Following the same approach as https://github.com/emberjs/ember.js/issues/11748 I ran some tests on a simple app to compare different versions of ember

### Observations: 
* Memory drops when running unit tests on ember 1.12.
* Memory allocated for tests is  ~ twice on ember 1.13 & 2.1 for same app.
* Memory allocation tends to increase overtime on acceptance tests.

### Questions:
* Is this expected ?
* Thoughts on how to debug further ?

### Memory comparison

![screen shot 2015-11-16 at 12 59 31 pm](https://cloud.githubusercontent.com/assets/118375/11191170/d0f1f530-8c67-11e5-8248-3a14047bb550.png)
![screen shot 2015-11-16 at 12 58 20 pm](https://cloud.githubusercontent.com/assets/118375/11191169/d0ef5668-8c67-11e5-9b95-ceea14cfab28.png)
![screen shot 2015-11-16 at 12 58 49 pm](https://cloud.githubusercontent.com/assets/118375/11191171/d0f3e822-8c67-11e5-9e16-86639734783b.png)

Looking at the timeline seems like event listeners remain after tests have completed and detached elements show up on heap profiles.

<img width="1432" alt="ef067318-892c-11e5-8a4d-fe75297533cb" src="https://cloud.githubusercontent.com/assets/118375/11191592/0638b89e-8c6a-11e5-9cfd-4e4a9579ead2.png">
<img width="1267" alt="f09d0494-892c-11e5-8c0f-aba694f2b135" src="https://cloud.githubusercontent.com/assets/118375/11191591/06376eee-8c6a-11e5-8bb0-739752603ab6.png">

test repo: https://github.com/calderas/leaky-mcleakerson

jcope2013:
this one seems related also https://github.com/emberjs/ember.js/issues/12490 and I have been having similar issues with my app

lmcardle:
I've got the same thing going on. I've been trying to track the issue down for the past few weeks to no avail. I am also on 1.13.10, here is what my timeline and profile look like.
[...]

PR #12652, opened 2015-11-26
[WIP] work around v8 memory leak

stefanpenner:
i don't really want to merge this, I'm likely going to explore an approach that doesn't trip up this specific v8 bug tomorrow.

PR #13220, opened 2016-03-30
[BUGFIX release-1-13] backports V8 memory leak bug through registry / resolver access.

calderas:
- [x] Backports #12666 to ember 1.13
- [x] Fixes jshint tests on ember-1-13 branch

Related to #12618

### DIFF
--- a/packages/container/lib/registry.js
+++ b/packages/container/lib/registry.js
@@ -28,7 +28,13 @@ if (Ember.FEATURES.isEnabled('ember-application-instance-initializers')) {
 function Registry(options) {
   this.fallback = options && options.fallback ? options.fallback : null;
 
-  this.resolver = options && options.resolver ? options.resolver : function() {};
+  if (options && options.resolver) {
+    this.resolver = options.resolver;
+    if (typeof this.resolver === 'function') {
+      deprecateResolverFunction(this);
+    }
+
+  }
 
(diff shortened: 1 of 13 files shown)


rwjblue:
Hmm, I'm somewhat concerned about this change being backported. It is a fairly large set of changes, and due to that it poses a level of risk that I'm not sure we should take on...

rwjblue:
We discussed this in detail during the most recent core team meeting ([see notes here](https://github.com/emberjs/core-notes/blob/master/ember.js/2016-04/april-08.md#evaluatediscuss-risk-of-backporting)), and we generally agree that their is a bit more risk in backporting than we are comfortable with.  It is likely that other changes are required (there were when the original PR landed) in various parts of the ecosystem (like ember-test-helpers, ember-qunit, etc) to support this change.

We would absolutely consider a fix that is less risky (ala making these functions too large to inline by adding a large string)...

rwjblue:
See https://github.com/emberjs/ember.js/pull/12652/files for an example of a very low risk change that prevents the leak also.

rwjblue:
Closing based on response to last few comments.

micky2be:
I tried using this patch on Ember 1.13.13 and still get PhantomJS to crash.
I'm trying to update my app from 1.12.2 to 2.x and going through 1.13.x is kinda mandatory for us to make sure we are not breaking anything on the way (it's a big app).

PR #13228, opened 2016-04-01
[BUGFIX release] re-enable link-to when disabledWhen becomes false

sly7-7:
Unless I missed something, when `disableWhen` is set to true, it was impossible to re-enable the link.
When I read https://github.com/emberjs/ember.js/blob/v2.4.0/packages/ember-routing-views/lib/components/link-to.js#L68, I think this is what the `dynamic` behavior should be.

### DIFF
--- a/packages/ember-routing-views/lib/components/link-to.js
+++ b/packages/ember-routing-views/lib/components/link-to.js
@@ -772,7 +772,7 @@ let LinkComponent = EmberComponent.extend({
     assert('You must provide one or more parameters to the link-to component.', params.length);
 
     let disabledWhen = get(this, 'disabledWhen');
-    if (disabledWhen) {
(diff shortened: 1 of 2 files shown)


rwjblue:
Thanks @sly7-7!

sly7-7:
Glad to help :). As a little side note, I wonder how `disabledWhen` is used... It's the first time I use it, and this bug appears immediatly. So to be honest I was pretty sure I was missing something...

Mbehnam:
what is exactly your point?

sly7-7:
@Mbehnam I have no particular point, actually only very surprised this bug was not encountered earlier.

rwjblue:
Included in v2.4.4.

sly7-7:
Great, thanks @rwjblue

Issue #13217, opened 2016-03-30
CP dependency keys split on commas?


nickiaconis:
Is this a thing we (I?) can deprecate? It would be nice to remove the extra loop from that function, and it would also open up the ability to use key names with commas (this bit me recently). I don't think this functionality is documented in the guides/api, but I didn't look super closely.

nickiaconis:
"That function" being github.com/emberjs/ember.js/blob/v2.4.0/packages/ember-metal/lib/expand_properties.js#L48-L52 of course. :sweat:

rwjblue:
Using brace expansion was a feature added sometime in 1.5ish timeframe.

    name: computed('model.{last,first}', function() {
      let { last, first } = this.getProperties('last', 'first');
      return first + ' ' + last;
    })

It seems to me that that comma detection/handling is what is supporting this feature, am I missing something?

nathanhammond:
I had no idea that was a feature.

1. We should at least do this without `forEach` for such a common code path.
2. If we can show significant perf wins removing this would that be grounds for deprecation of the sugar?
3. Alternatively we could probably do a transform at compile time to drop the runtime code and still have the feature.

nickiaconis:
Interesting. I too was not aware of this feature. Do you know if/where it's documented? Might we want to only split on commas that are inside curly braces? That is `'model.{first,last}'` gets split, but `'ember-cli,-the-next-generation'` does not.

Also, I can put together a PR for 1). A question that comes to mind is, would `'person.{name.{first,last},twitterHandle}'` be expected expand to expand to `'person.name.first', 'person.name.last', 'person.twitterHandle'` or does this only go one level deep?

acorncom:
If we do remove this, it'd be helpful to know over on the guides as well, as we just merged a PR to put it into the guides ;-)  See https://github.com/emberjs/guides/pull/1348 for more detail (and a link back to the original blog post where it was introduced)

stefanpenner:
* This isn't going to be removed.
* Moving to for loops seem fine
* this is done at design time, and is quite unlikely to pose much of a performance hit especially relative to things like extend and merge mixins. If someone can demonstrate this to be incorrect, a Babel plugin could be created. That being said, I would believe the overhead (although it wouldn't surprise me if some small TLC is required) to be negligible

krisselden:
@stefanpenner that was the original idea, but if you define a lot of classes design time starts being an issue for cold boot.

[...]

PR #13231, opened 2016-04-02
Correctly handle commas in CP property keys

nickiaconis:
This addresses the issue of splitting on all commas in dependency key names, which prompted the opening of #13217.

Currently:

``` javascript
expandProperties('a,b,c.d.e,f', console.log); //=> 'a', 'b', 'c.d.e', 'f'
```

With this PR applied:

``` javascript
expandProperties('a,b,c.d.e,f', console.log); //=> 'a,b,c.d.e,f'
```

**Edited** to remove reference of nested brace expansion, since we don't intend to support it.

### DIFF
--- a/packages/ember-metal/lib/expand_properties.js
+++ b/packages/ember-metal/lib/expand_properties.js
@@ -40,28 +40,8 @@ export default function expandProperties(pattern, callback) {
     'Brace expanded properties cannot contain spaces, e.g. "user.{firstName, lastName}" should be "user.{firstName,lastName}"',
     pattern.indexOf(' ') === -1
   );
-  assert(
-    `Brace expanded properties have to be balanced and cannot be nested, pattern: ${pattern}`,
-    ((str) => {
-      let inBrace = 0;
-      let char;
-      for (let i = 0; i < str.length; i++) {
-        char = str.charAt(i);
 
-        if (char === '{') {
-          inBrace++;
-        } else if (char === '}') {
-          inBrace--;
-        }
-
-        if (inBrace > 1 || inBrace < 0) {
-          return false;
-        }
-      }
-
-      return true;
-    })(pattern));
-
+  let unbalancedNestedError = `Brace expanded properties have to be balanced and cannot be nested, pattern: ${pattern}`;
   let properties = [pattern];
 
   // Iterating backward over the pattern makes dealing with indices easier.
@@ -77,6 +57,8 @@ export default function expandProperties(pattern, callback) {
         if (!inside) {
           bookmark = i - 1;
           inside = true;
+        } else {
+          assert(unbalancedNestedError, false);
         }
         break;
       // Opening curly brace will be the last character of the brace expansion we encounter.
@@ -96,9 +78,14 @@ export default function expandProperties(pattern, callback) {
             }
           }
           inside = false;
+        } else {
+          assert(unbalancedNestedError, false);
         }
         break;
     }
+  }
+  if (inside) {
+    assert(unbalancedNestedError, false);
   }
 
   for (let i = 0; i < properties.length; i++) {
--- a/packages/ember-metal/lib/expand_properties.js
+++ b/packages/ember-metal/lib/expand_properties.js
@@ -5,8 +5,7 @@ import { assert } from './debug';
 @submodule ember-metal
 */
 
-const SPLIT_REGEX = /\{|\}/;
-const END_WITH_EACH_REGEX = /\.@each$/;
+var END_WITH_EACH_REGEX = /\.@each$/;
 
 /**
   Expands `pattern`, invoking `callback` for each expansion.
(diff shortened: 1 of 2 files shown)


packages/ember-metal/lib/expand_properties.js
```diff
@@ -42,31 +42,103 @@ export default function expandProperties(pattern, callback) {
...
   var all = [];
+  var parts = splitOutsideBraces(match[1], ',');
 
-  properties.forEach((property) => {
-    currentParts.forEach((part) => {
-      var current = property.slice(0);
-      current[index] = part;
-      all.push(current);
-    });
+  parts.forEach((part) => {
```
stefanpenner: Map?  Or should this be unrolled

rwjblue: Seems like it should be a straight for loop here.

rwjblue:
I was under the impression that we did not intend to support nested brace expansion, so I'm not sure if we should add that.

@stefanpenner/@krisselden - Thoughts?

I'm also somewhat curious to see if we have made this better or worse performance wise, would it be possible to make a small benchmark before and after?

nickiaconis:
@rwjblue I did not remember if we talked about the nesting in person or, if we did, what the conclusion was, and I didn't get a response about it over on #13217. It's possible to remove the second commit, which is the one that introduces nested handling, without affecting comma handling.

I'll put together a benchmark to compare before/comma/nested.

packages/ember-metal/lib/expand_properties.js
nickiaconis: Good point about the map, @stefanpenner.

Since there's an overall desire to reduce function calls, I'll rewrite this as a straight for loop.

rwjblue:
Awesome, that would be great, thank you.

rwjblue:
@nickiaconis - Any updates here?

nickiaconis:
Here are the [results of the benchmarking](https://gist.github.com/nickiaconis/85692900b4675b3d262e5d17d933a37d#file-results-md). The code used to run the benchmark is also included in the gist.

Since the "comma" and "nested" versions were so much slower than the "before" version, I wrote a few other implementations. The "iteration" version is a little slower than the "before" version, but I think it's close enough to be acceptable.

rwjblue:
Thanks for putting in the time on that benchmark! The iteration version looks good to me (implementation and perf wise).

@krisselden - What do you think?

nickiaconis:
On April 12, @krisselden mentioned he was reviewing my benchmarks by trying them in a browser:

> JIT stuff can make what you are testing not actually what you think
> I just need to confirm them
> I’m just not sure why they are so different

I wonder at what conclusion he arrived.

stefanpenner:
I am nervous about supporting this feature at all, Is it really common for people to valid properties with `,` in them?

nickiaconis:
@stefanpenner I don't know if it's common, but it can definitely happen, especially when the property names are generated dynamically. For example, the place where I bumped into this was Erik's app-cache-demo from EmberConf 2016. The (static) data is set in the application route, and it has a model whose [ID contains a comma](https://github.com/nickiaconis/ember-appcache-demo/blob/master/app/routes/application.js#L314). Later, that ID is used to [create a computed property](https://github.com/ebryn/ember-appcache-demo/blob/93cb108a7c84a05c6abc83c0ffbb6d67de7f4ef2/app/components/is-interested-button.js#L13-L15). Of course, this doesn't work because of the bug this PR fixes, hence the need to [remove all commas from IDs](https://github.com/ebryn/ember-appcache-demo/commit/edfc6e2edabb9b44fc8d7e0a6956dac8c165676b#diff-a35ace277496b5cc0ee78e3c7fb7e2a4).

stefanpenner:
The perf improvement in the iteration fix variant is very nice. I appreciate the due diligence here.

We (the ember core team) needs to decide if this is something we intend to support or not. As demonstrated it doesn't appear to make the status quo worse.

The only technical concern I can think of is, can we make the status quo better, and would enabling this feature prevent that?

@tomdale would love your guidance here.

packages/ember-metal/lib/expand_properties.js
```diff
@@ -42,31 +41,46 @@ export default function expandProperties(pattern, callback) {
...
 
-  var parts = pattern.split(SPLIT_REGEX);
-  var properties = [parts];
+  let properties = [pattern];
 
-  for (let i = 0; i < parts.length; i++) {
-    let part = parts[i];
-    if (part.indexOf(',') >= 0) {
-      properties = duplicateAndReplace(properties, part.split(','), i);
+  // Iterating backward over the pattern makes dealing with indeces easier.
```
tomdale: Nitpick: s/indeces/indices/

tomdale:
This seems like a great improvement to me. While it's easy to consider the previous `,`-splitting behavior an edge case, it's infuriating to be able to express something valid in the language that isn't supported by your tools—and becomes the kind of thing that makes people think they don't like "magic."

My only comment is that I wouldn't mind having a few more tests to verify correct behavior in other, weird cases like nested `{`s, missing `}`s, etc. And I wonder if there is any performance benefit to having a fast path that checks for `indexOf('}')` and bails early if < 0? (It may be not, given the nature of the implementation.)

(And let me just say, @nickiaconis, I am very impressed that you implemented a stateful mini-parser here instead of trying to gin up an overly complicated regular expression, which seems like what most people try to do in these scenarios.)

nickiaconis:
Thanks for the review, @tomdale!

Upstream changes have added tests for nested `{`s and missing `}`s. Now that I've rebased, those cases are covered. Were there any other tests you had in mind?

Speaking of upstream changes, an assertion to catch unbalanced and nested braces was added, which involves iterating over `pattern`, just as the mini-parser does. I've rolled that assertion into the mini-parser's iteration for a slight performance win. Do you agree with this choice? I've made it a separate commit for now, so it can be either squashed or pruned before merging.

The performance of `pattern.indexOf` is O(n), and the performance of the mini-parser when `pattern.indexOf('{') === -1` is O(n), so there would be no performance benefit to bailing early. It might boost code clarity, but at the cost of a minor performance degradation when `pattern.indexOf('{') !== -1`, so I'd recommend against.

P.S. What's the rational behind the no-const-outside-module-scope ESLint rule?

rwjblue:
> P.S. What's the rational behind the no-const-outside-module-scope ESLint rule?

See "liberal let" in http://madhatted.com/2016/1/25/let-it-be.

nickiaconis:
Thanks for the info @rwjblue! I'm of the "constantly const" opinion as articulated in https://ponyfoo.com/articles/var-let-const, but it's good to read a differing opinion.

nickiaconis:
@rwjblue I've rebased and resolved merge conflicts, so this is ready for review and/or merge.

rwjblue:
LETS DO IT!

PR #13233, opened 2016-04-02
[DOC] Update NativeArray documentation

elwayman02:
`NativeArray.activate` was removed several months ago, but the documentation for the mixin still recommends that method for activating it. This change updates the docs to suggest the same code that is used internally by the `NativeArray` mixin to apply itself to `Array.prototype`.

/cc @rwjblue


rwjblue:
Thanks!

PR #13235, opened 2016-04-02
[Glimmer2] Remove `{{#view}}` helper tests

chadhietala:
This currently removes just the tests related to the `view` helper. I have left some tests in the file because I have a feeling that these tests are cross cutting and were used as coverage for components. I could be 100% incorrect about this but would like to hear from @rwjblue and @chancancode before they are removed :bomb:.

Furthermore, it looks like we will need to dereference all the `view` helpers in all of the tests before we can finally remove the helper. Luckily majority of the tests in #13127 should be addressing this.

Part of #13127

PR #13236, opened 2016-04-02
[Glimmer2] Add more curly component tests

chadhietala:
This adds more tests around custom ids and bindings. These cases seem to be
only tested by the `{{#view}}` helper and that is going away. To ensure
we are at least tracking this functionality I added some tests to the
curly-component integration tests.

### DIFF
--- a/packages/ember-glimmer/tests/integration/components/curly-components-test.js
+++ b/packages/ember-glimmer/tests/integration/components/curly-components-test.js
@@ -128,6 +128,18 @@ moduleFor('Components test: curly components', class extends RenderingTest {
     this.assertComponentElement(this.firstChild, { tagName: 'div', attrs: { 'class': classes('ember-view foo bar') }, content: 'hello' });
   }
 
+  ['@htmlbars it should apply classBinding without condition always']() {
+    this.registerComponent('foo-bar', { template: 'hello' });
+
+    this.render('{{foo-bar classBinding=":foo"}}');
+
+    this.assertComponentElement(this.firstChild, { tagName: 'div', content: 'hello', attrs: {'class': classes('foo  ember-view') } });
+
+    this.runTask(() => this.rerender());
+
+    this.assertComponentElement(this.firstChild, { tagName: 'div', content: 'hello', attrs: {'class': classes('foo  ember-view') } });
+  }
+
   ['@htmlbars should not apply falsy class name']() {
     this.registerComponent('foo-bar', { template: 'hello' });
 
--- a/packages/ember-glimmer/tests/integration/components/curly-components-test.js
+++ b/packages/ember-glimmer/tests/integration/components/curly-components-test.js
@@ -700,25 +700,25 @@ moduleFor('Components test: curly components', class extends RenderingTest {
 
     assert.strictEqual(this.firstChild.childNodes.length, 4);
 
-    this.assertElement(this.firstChild.childNodes[2], { tagName: 'b', content: 'bold'});
+    this.assertElement(this.firstChild.childNodes[2], { tagName: 'b', content: 'bold' });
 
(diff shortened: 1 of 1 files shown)


chadhietala:
Related to #13235. Part of #13127.

packages/ember-glimmer/tests/integration/components/curly-components-test.js
```diff
@@ -18,6 +19,29 @@ moduleFor('Components test: curly components', class extends RenderingTest {
     this.assertComponentElement(this.firstChild, { content: 'hello' });
   }
 
+  // Note this functionality seems weird
```
rwjblue: In `{{foo-bar id=blah}}`, we evaluate the value of `blah` and "freeze" it as `elementId` in the component instance.

`elementId` can't be changed, if it did change then any actions/events would stop working. The view registry is keyed on the `elementId`, and is used to look up action/event targets. So it is definitely possible to make changing the `id`/`elementId` work, but it would be a new feature.

We should ensure that a test also exists in this file for:

```
let fooBarInstance;
this.registerComponent('foo-bar', {
    elementId: 'blahzorz',
    init() {
      this._super(...arguments);
      fooBarInstance = this;
    }
  }),

  template: `{{elementId}}`
});

this.render('{{foo-bar}}');

this.assertComponentElement(this.firstChild, { tagName: 'div', attrs: { id: 'blahzorz' }, content: 'blahzorz'});

this.runtask(() => {
  expectAssertion(() => {
    fooBarInstance.set('elementId', 'herpyderpy');
  }, /can not change elementId/);
});

this.assertComponentElement(this.firstChild, { tagName: 'div', attrs: { id: 'blahzorz' }, content: 'blahzorz'});
```

rwjblue:
This looks good, thanks for moving these tests.  I left a comment explaining the behavior of `id=foo`, and suggested adding another test for changing `elementId` after render.

packages/ember-glimmer/tests/integration/components/curly-components-test.js
chadhietala: Makes sense. Added this test. The only thing is that an `EmberError` gets thrown so `expectAssertion` is not the correct assertion as it won't be stripped in prod builds.

rwjblue: Ahh, gotcha. I wasn't sure if it was an assertion or an error.

chadhietala:
@rwjblue I think CI just needs to be kicked on this one. I had to do some interesting stuff to get the prod tests to pass. https://github.com/emberjs/ember.js/pull/13236/files#diff-c47395a5aaf92aa0a7edbfa812d4e0afR61

packages/ember-glimmer/tests/integration/components/curly-components-test.js
```diff
@@ -18,6 +20,53 @@ moduleFor('Components test: curly components', class extends RenderingTest {
...
+      }
+    });
+
+    this.registerComponent('foo-bar', { ComponentClass: FooBarComponent, template: '{{elementId}}' });
+
+    this.render('{{foo-bar}}');
+
+    this.assertComponentElement(this.firstChild, { tagName: 'div', attrs: { id: 'blahzorz' }, content: 'blahzorz' });
+
+    if (EmberDev && !EmberDev.runningProdBuild) {
```
rwjblue: If only https://github.com/emberjs/ember.js/blob/master/packages/ember-views/lib/views/states/in_dom.js#L24 used `assert` this wouldn't be needed (you could use `expectAssertion` which does these checks internally).

This is good though, I'm just grumbling :smiling_imp:

rwjblue:
Kicked sauce labs job...

PR #13239, opened 2016-04-03
[BUGFIX release] Prevent warnings for unknown feature flags.

rwjblue:
Only warn when features are enabled that we know about. This prevents a warning when users enable Ember Data features when running stable versions of Ember with canary versions of Ember Data.

/cc @bmac


rlivsey:
I'm still getting this warning when running Ember 2.7.2 and Ember Data canary.

screen shot 2016-09-01 at 10 03 47

PR #13251, opened 2016-04-04
[Glimmer2] Glimmer compiler options

chadhietala:
I think this all that is missing for AST plugins and other options to be used in the glimmer compiler.

### DIFF
--- a/packages/ember-template-compiler/lib/system/compile.js
+++ b/packages/ember-template-compiler/lib/system/compile.js
@@ -30,7 +30,7 @@ export default function(templateString, options) {
       throw new Error('Cannot call `compile` without the template compiler loaded. Please load `ember-template-compiler.js` prior to calling `compile`.');
     }
 
-    return template(compile(templateString, options));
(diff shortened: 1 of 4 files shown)


chadhietala:
I think I might have to symlink here to make sure both compilers are running the test but would like some :eyes: /cc @chancancode @rwjblue @wycats

chadhietala:
Tested this in an app and appears to work. When I mean work, I mean the AST transforms get executed.

wycats:
I'm going to merge this but I'm pretty uneasy about making this API feel even more public. Just to be clear, it's emphatically not intended to be a public API, and it will likely break more over time.

We do need a public API for solving the exact problem you have in mind, which we should chat about (so we can make sure we hit your requirements when we write the RFC). In the meantime, caveat emptor.

rwjblue:
@wycats this enables the Ember specific AST transforms that enable a number of things via rewrites. This doesn't infer in any way that AST transforms are considered public API (though I do think we should find a way to provide some of this functionality via public API at some point).

PR #13252, opened 2016-04-04
[DOCS] Remove get/set documentation from ComputedProperty.prototype

krisselden:
These were documented as though it is documenting Ember.get/set usage of computed properties but these particular methods were very much internal.

Also, remove @constructor from the class description since this is not documenting the constructor it is describing the general usage of the descriptor instance.


rwjblue:
Thanks @krisselden!

PR #13261, opened 2016-04-05
Support descending columns in enumerable sortBy

straydogstudio:
It would be useful to have intermediate colums be descending without creating computed properties.


straydogstudio:
My rebase didn't quite work. Sigh. Closing and will reopen.

mmun:
We want to avoid introduce microsyntaxes into more places (yes, even if it would be consistent with other APIs). We need a strategy to unify all 3 sorting methods with a new, backwards-compatible API. Check out this RFC comment: https://github.com/emberjs/rfcs/pull/87#issuecomment-187432850

BTW, this is a new feature so it would need to be feature flagged as well. You can grep around for `isEnabled(` to see some examples.

mmun:
I was thinking something like

```js
let { desc } = Ember.sort;
array.sortBy(desc('createdAt'), 'name')
```

straydogstudio:
Thanks @mmun. I was definitely trying to be consistent with the computable sortBy implementation, and I didn't notice the RFC. I'll try implementing the new backwards compatible api without microsyntaxes for computable and enumerable.

So the general consensus is to keep computable.sort and enumerable.sortBy, without changing either interface, but providing a better syntax for a descending column? While computable.sort is backwards compatible with the microsyntax?

mmun:
@straydogstudio You got it. I think it's important to add `computed.sortBy` as part of the effort as well.

straydogstudio:
@mmun: What would be the difference between computable.sort and computable.sortBy?

straydogstudio:
@mmun: If we want equal interfaces, should enumerable.sortBy accept a function?

Also, this will create common code. Should I create it in enumerable and import that into computed? Create another mixin?

PR #13169, opened 2016-03-23
[BUGFIX beta] Fixes {{link-to}} classes when transition is aborted

GavinJoyce:
fixes https://github.com/emberjs/ember.js/issues/13164

Issue #13164, opened 2016-03-22
`{{link-to}}` transitioning classes are incorrect after `transition.abort()`


GavinJoyce:
twiddle: https://ember-twiddle.com/8aa9da800a7883b6c59a?openFiles=routes.page2.js%2C

The `ember-transitioning-in` and `.ember-transitioning-out` classes don't seem to be updated correctly when linking to a route that aborts the transition:

``` js
export default Ember.Route.extend({
  beforeModel: function(transition) {
    return delay(1000).then(function() {
      transition.abort();   
    });
  }
});
```

I'll try to create a failing test now

GavinJoyce:
possibly related: https://github.com/emberjs/ember.js/issues/10195

GavinJoyce:
Failing test: https://github.com/emberjs/ember.js/pull/13165

GavinJoyce:
@rwjblue perhaps you could reopen this issue due to https://github.com/emberjs/ember.js/pull/13262?

rwjblue:
Good point.  Reopened.

pixelhandler:
@GavinJoyce I updated the twiddle example using v2.5.1 https://ember-twiddle.com/5e91f27f1df42b6e41231f53a12af3da?openFiles=application.template.hbs%2C&route=%2Fpage1 and labeled "BUG"

kjhangiani:
Just got bit by this issue, found via #15518.  Is there a workaround to force ember to recompute link-to active states?  The workaround mentioned in that ticket does not seem to work for me.

kjhangiani:
So, this was somewhat of a rabbithole.  I was able to reproduce the specific issue I was having on a twiddle here:

https://ember-twiddle.com/128d462ce1fa0acdebd696338ca3dc14?openFiles=templates.application.hbs%2C

In particular, my issue was due to a `transition.abort()` followed by an `Ember.run.later(transition, transition.retry, 50)` in the `willTransition` hook of a route.  The problem was specifically when changing between the same route, with just a positional param model_id change.

I found a workaround here: https://github.com/emberjs/ember.js/issues/10557 which thankfully works (essentially replaces `transition.retry` with a `this.transitionTo`)

In my digging, I found a number of related tickets, linking them all here - these all seem related to me:
https://github.com/emberjs/ember.js/issues/10557 (solution found here)
https://github.com/emberjs/ember.js/issues/10195
https://github.com/emberjs/ember.js/issues/13164
[...]

PR #13262, opened 2016-04-05
[BUGFIX beta] revert #13169 and add test to cover {{link-to}} active class regression

### LINKED ISSUE #13256: Active class missing after transition to child route, since latest beta
since v2.5.0-beta.4

#13169 seems to break something. Active class is missing after transition to child route.
Twiddle example

https://ember-twiddle.com/8e38c8ca6ba43b6331a3b0ff4c5a4cf7?openFiles=templates.application.hbs%2C

GavinJoyce:
fixes https://github.com/emberjs/ember.js/issues/13256
reopens https://github.com/emberjs/ember.js/issues/13164

TIL doing a `transitionTo` in a route model hook results in a `TransitionAborted` error. I was [using](https://github.com/emberjs/ember.js/pull/13169) that to reset the router target state to address https://github.com/emberjs/ember.js/issues/13164. This won't work so I'm reverting https://github.com/emberjs/ember.js/pull/13169 and adding a test to cover the regression.

### DIFF
--- a/packages/ember-routing/lib/system/router.js
+++ b/packages/ember-routing/lib/system/router.js
@@ -731,11 +731,6 @@ var EmberRouter = EmberObject.extend(Evented, {
   currentState: null,
   targetState: null,
 
-  _resetTargetState() {
-    let currentState = this.get('currentState');
-    this.set('targetState', currentState);
-  },
-
   _handleSlowTransition(transition, originRoute) {
     if (!this.router.activeTransition) {
       // Don't fire an event if we've since moved on from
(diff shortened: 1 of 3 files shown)


rwjblue:
LGTM, I need to figure out how far back I cherry-picked the other commit.  I presume it is only `beta` based on the prefix here, is that correct (can you double check commits on `release` branch)?

GavinJoyce:
I don't see it in the [release](https://github.com/emberjs/ember.js/commits/release) commits. The [previous commit](https://github.com/emberjs/ember.js/pull/13169) was marked `[BUGFIX beta]` and was merged into master 14 days ago

rwjblue:
Thanks for double checking @GavinJoyce!

Issue #13224, opened 2016-03-31
Ember core components should support other WAI-ARIA attributes


kboucher:
Currently only the `role` (`ariaRole`) attribute is supported in Ember core components. Various WAI-ARIA attributes are essential to providing a usable experience for impaired users.

For example using `link-to`:
```
    {{#link-to 'my-route' aria-label="This link will take you to some other route." title="Go to other route name"}}
        My Link Text
    {{/link-to}}
```
renders:
```
    <a id="ember1012" href="#/my-route" title="Go to other route name" class="ember-view">
        My Link Text
    </a>
```
Due to the number of WAI-ARIA attributes and their validity in various tags makes it difficult to suggest an exhaustive list. Perhaps any attributes with the `aria-` prefix added to a helper could just be rendered?

pixelhandler:
@kboucher Nice work forming a solution #13265. I did see some discussion on Aria- use in this RFC, https://github.com/emberjs/rfcs/blob/angle-bracket-components/active/0000-component-unification.md Perhaps comment on that RFC as well. 

Since this ia a feature/enhancement request. See [CONTRIBUTING.md#requesting-a-feature](https://github.com/emberjs/ember.js/blob/master/CONTRIBUTING.md#requesting-a-feature) which recommends creating an RFC Issue to request a feature/enhancement (no need for a PR with RFC, only an issue in the repo)

It may be a good idea to get the discussion going and address the concerns raised regarding performance too.

sundered:
Because of the performance concern, I suggested a new API for this in the PR. It would be great if you guys could take a look and give your 2 cents.

rwjblue:
Features are generally discussed in the https://github.com/emberjs/rfcs/ repository. You can create an issue there for feature requests or a pull request to propose a new feature.  I'm closing this issue (this isn't really a bug in the framework), but please follow up with an RFC issue/PR if you are interested in taking this further.

As mentioned in the PR, the RFC will need to wrestle with adding additional API surface here in the short-term (that likely has a negative performance impact) when we have a much better long term plan (with angle bracket components the invocation site can supply any attributes that it wants, and things will "just work").

I'm going to close this (and the PR) for now, please do open that RFC to flesh out further discussion....

PR #13265, opened 2016-04-06
[FEATURE wai-aria-attribute-support] Modifies the aria_role_support Mixin to support any WAI-ARIA property that starts with 'aria-'

hleimig:
Implements the feature https://github.com/emberjs/ember.js/issues/13224

With this modification, any property defined in Views or Components that starts with 'aria-' will be rendered into the View/Component's DOM, improving accessibility in Ember.js apps.

### DIFF
--- a/packages/ember-glimmer/lib/ember-views/component.js
+++ b/packages/ember-glimmer/lib/ember-views/component.js
@@ -3,7 +3,7 @@ import ChildViewsSupport from './child-views-support';
 import ClassNamesSupport from './class-names-support';
 import ViewStateSupport from 'ember-views/mixins/view_state_support';
 import InstrumentationSupport from 'ember-views/mixins/instrumentation_support';
-import AriaRoleSupport from 'ember-views/mixins/aria_role_support';
+import AriaSupport from 'ember-views/mixins/aria_support';
 import ViewMixin from 'ember-views/mixins/view_support';
 import EmberView from 'ember-views/views/view';
 
@@ -12,7 +12,7 @@ export default CoreView.extend(
   ViewStateSupport,
   ClassNamesSupport,
   InstrumentationSupport,
-  AriaRoleSupport,
+  AriaSupport,
   ViewMixin, {
     isComponent: true,
     template: null,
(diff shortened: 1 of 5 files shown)


sundered:
I'm not sure if `Object.keys` is the best way to access the properties of a View/Component. I tried to use the `attrs` property, but it is `null` while the Mixin is being created.

mixonic:
@sundered we would want to see benchmarks that ensure this does not regress performance. Iterating the keys of every component during creation is likely going to give us a hit. Creating an observer for each binding will likely hit performance.

Generally I believe we're designing glimmer component with the accessibility use-cases in mind and will have a nicer solution there.

However there needs to be more investigation before landing this.

sundered:
Ok, I'll try to improve this and get some benchmarks.

sundered:
So, I had an ideia. What if the user could set a property called `aria` with an object that contained the aria- values and we used the `computed` set method to bind the variables. Something like this

```javascript
ariaProperties: {
    label: "A brief description",
    aria-disabled: false
}
```
```handlebar
{{foo-bar aria=ariaProperties}}
```

rwjblue:
I'm not sure how I feel about adding a new API here since we intend for angle bracket components to automatically place any passed in attributes on the top level element of the component.  When that lands, no custom/special support will be needed to allow specifying of these properties.  

Example without needing to define `app/components/foo-bar.js`:

```hbs
{{! app/templates/components/foo-bar.hbs }}
<div class="some-special-thing"></div>
```

```hbs
{{! Invocation syntax }}
<foo-bar aria-label="foo" data-bar="blammo">
</foo-bar>
```

```html
<!-- rendered HTML -->
<div class="some-special-thing" aria-label="foo" data-bar="blammo"></div>
```

---

Adding any additional logic to curly component instantiation at this point is fairly risky due to performance concerns....

rwjblue:
Closing...

See detailed comment in https://github.com/emberjs/ember.js/issues/13224#issuecomment-213861097.

rwjblue:
And thank you very much for tackling this! I know it seems like a somewhat arbitrary thing, but we definitely need to put these sorts of conversations about new public API's through the RFC process. We have learned quite a bit from our users since starting the process, and I suspect this will be another example of that.

PR #13266, opened 2016-04-06
[DOC release] Remove `Em` shorthand sentence.

locks:
Creates some confusion in the Ember CLI modules world.


rwjblue:
Thanks @locks!

PR #13267, opened 2016-04-06
[DOC release] Fixes `defineProperty` computed property code sample

locks:
Was using function prototype extension instead of passing arguments to `Ember.computed`.


rwjblue:
Thanks @locks!

Issue #13183, opened 2016-03-25
Deprecate usage of `{{render` helper with a model param.


rwjblue:
This is always better written today as a component invocation, we should deprecate this.

Bad:

```hbs
{{render 'foo-bar' someModelThingy}}
```

```hbs
{{foo-bar model=someModelThingy}}
```

---

- [ ] AST transform to deprecate while compiling templates (to provide line number information)
- [ ] Add deprecation guide entry in emberjs/website

cesardariogarcia:
@rwjblue I tried to solve this issue but I'm new to ember so I'm not pretty sure that i did it well,
could you check this changes to see if they are good to make a PR.
Or in defect I am open to any suggestion to achieve this, thanks :). 

https://github.com/cesardariogarcia/ember.js/commit/1f987180a7d4b8f16fa67ab6489b104f7ea28d40

p.s. I located the deprecated call in the validation because there is where the `{{render + model` comes in i think.
```
if (params.length > 1)
+      deprecate('Usage of `{{render` with a model param is deprecated, use `{{render "foo"} ` instead.', false, { id: 'ember-metal-views.renderer', until: '3.0.0' });
```

sweatypitts:
I'll take a stab at this.

rwjblue:
Awesome, thanks!

sweatypitts:
@rwjblue  How's this look for a failing test? https://github.com/emberjs/ember.js/compare/master...sweatypitts:render-model-param-deprecation?expand=1

PR #13268, opened 2016-04-06
Deprecate usage of `{{render` helper with a model param

pittst3r:
Addresses first part of #13183.

This is my first time diving into Ember internals like this, so I'm pretty sure it won't be 100% off the bat. Also not sure about how to tag this.

### DIFF
--- a/packages/ember-template-compiler/lib/plugins/deprecate-render-model.js
+++ b/packages/ember-template-compiler/lib/plugins/deprecate-render-model.js
@@ -1,0 +1,52 @@
+import { deprecate } from 'ember-metal/debug';
+import calculateLocationDisplay from
+  'ember-template-compiler/system/calculate-location-display';
+
+export default function DeprecateRenderModel(options) {
+  this.syntax = null;
+  this.options = options;
+}
+
+DeprecateRenderModel.prototype.transform =
+  function DeprecateRenderModel_transform(ast) {
+  let moduleName = this.options.moduleName;
+  let walker = new this.syntax.Walker();
+
+  walker.visit(ast, function(node) {
+    if (!validate(node)) { return; }
+
+    each(node.params, (param) => {
+      if (param.type !== 'PathExpression') { return; }
+
+      deprecate(deprecationMessage(moduleName, node, param), false, {
+        id: 'ember-template-compiler.deprecate-render-model',
+        until: '3.0.0'
+      });
+    });
+  });
+
+  return ast;
+};
+
+function validate(node) {
+  return (node.type === 'MustacheStatement') &&
+    (node.path.original === 'render') &&
+    (node.params.length > 1);
+}
+
+function each(list, callback) {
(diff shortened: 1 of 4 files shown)


packages/ember-template-compiler/lib/plugins/transform-render-model.js
```diff
@@ -0,0 +1,54 @@
+import { deprecate } from 'ember-metal/debug';
```
rwjblue: Small nit-pick, but this file doesn't transform anything.  Can you rename to `deprecate-render-model.js` instead?

packages/ember-template-compiler/lib/plugins/transform-render-model.js
```diff
@@ -0,0 +1,54 @@
...
+    if (!validate(node)) { return; }
+
+    each(node.params, (param) => {
+      if (param.type !== 'PathExpression') { return; }
+
+      let sourceInformation = calculateLocationDisplay(moduleName, node.loc);
+      let preferred = preferredForm(node, param);
+      let deprecationMessage =
+        `Using the '{{render' helper with a model param is deprecated.` +
+        ` Please use '${preferred}' instead. ${sourceInformation}`;
```
rwjblue: You might tweak this to say:

```
Please refactor `{{render 'whatever-name-they-typed' modelParamName}}` to a component and invoke via `{{whatever-name-they-typed model=modelParamName}}`.
```

rwjblue:
This is looking great, thanks for tackling!  I left a few comments inline for tweaks/suggestions, and when you are done can you squash commits?

We will also need a PR to the website repo adding this deprecation to the list for Ember 2.6.

packages/ember-template-compiler/lib/plugins/transform-render-model.js
sweatypitts: Yeah, I was wondering about that (the best name for the file).

sweatypitts:
@rwjblue No problem, it was fun. Made the requested changes. I'll work on the website update later/tomorrow.

rob-e-pitts:
[Website PR](https://github.com/emberjs/website/pull/2535)

packages/ember-template-compiler/lib/plugins/deprecate-render-model.js
```diff
@@ -0,0 +1,52 @@
...
+
+  walker.visit(ast, function(node) {
+    if (!validate(node)) { return; }
+
+    each(node.params, (param) => {
+      if (param.type !== 'PathExpression') { return; }
+
+      deprecate(deprecationMessage(moduleName, node, param), false, {
+        id: 'ember-template-compiler.deprecate-render-model',
+        until: '3.0.0'
```
rwjblue: Whoops! I just realized that this doesn't include the URL from the website PR. Can you submit a follow up to add that in?

rob-e-pitts: Done!

PR #13270, opened 2016-04-06
Bump glimmer

chadhietala:


rwjblue:
I was under the impression that there were some changes that had to be made here to bump?

krisselden:
@rwjblue to support components with slash in the name.

rwjblue:
Ya, I'm totally in favor of updating, but what I'm saying is that there are broken tests when we update. So I assumed that some changes upstream in glimmer need corresponding updates here.

krisselden:
@rwjblue confirm there is a branch dealing with the reference/validator changes.

krisselden:
this was superseded by #13278

PR #13272, opened 2016-04-06
[DOC release] updated docs regarding ember debug methods

webark:
This brings them all in line with #13180

Was brought up during conversation in https://github.com/ember-cli/rfcs/pull/50#issuecomment-206509353

### DIFF
--- a/packages/ember-debug/lib/deprecate.js
+++ b/packages/ember-debug/lib/deprecate.js
@@ -95,8 +95,10 @@ export let missingOptionsUntilDeprecation = 'When calling `Ember.deprecate` you 
 
 /**
   Display a deprecation warning with the provided message and a stack trace
-  (Chrome and Firefox only). Ember build tools will remove any calls to
-  `Ember.deprecate()` when doing a production build.
+  (Chrome and Firefox only).
+
+  * In a production build, this method is defined as an empty function (NOP).
+  Uses of this method in Ember itself are stripped from the ember.prod.js build.
 
   @method deprecate
   @param {String} message A description of the deprecation.
--- a/packages/ember-debug/lib/index.js
+++ b/packages/ember-debug/lib/index.js
@@ -30,10 +30,10 @@ import { generateTestAsFunctionDeprecation } from 'ember-debug/handlers';
 
 
 /**
-  Define an assertion that will throw an exception if the condition is not
-  met. Ember build tools will remove any calls to `Ember.assert()` when
-  doing an Ember.js framework production build and will make the assertion a
-  no-op for an application production build. Example:
+  Define an assertion that will throw an exception if the condition is not met.
+
+  * In a production build, this method is defined as an empty function (NOP).
+  Uses of this method in Ember itself are stripped from the ember.prod.js build.
 
   ```javascript
   // Test for truthiness
@@ -71,8 +71,10 @@ setDebugFunction('assert', function assert(desc, test) {
 });
 
 /**
-  Display a debug notice. Ember build tools will remove any calls to
-  `Ember.debug()` when doing a production build.
+  Display a debug notice.
+
+  * In a production build, this method is defined as an empty function (NOP).
+  Uses of this method in Ember itself are stripped from the ember.prod.js build.
 
   ```javascript
   Ember.debug('I\'m a debug notice!');
@@ -89,6 +91,9 @@ setDebugFunction('debug', function debug(message) {
 /**
   Display an info notice.
 
+  * In a production build, this method is defined as an empty function (NOP).
+  Uses of this method in Ember itself are stripped from the ember.prod.js build.
+
   @method info
   @private
 */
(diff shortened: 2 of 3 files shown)


packages/ember-debug/lib/deprecate.js
```diff
@@ -95,8 +95,9 @@ export let missingOptionsUntilDeprecation = 'When calling `Ember.deprecate` you
 
 /**
   Display a deprecation warning with the provided message and a stack trace
-  (Chrome and Firefox only). Ember build tools will remove any calls to
-  `Ember.deprecate()` when doing a production build.
+  (Chrome and Firefox only). Ember build tools will remove any calls
+  to `Ember.deprecate()` when doing an Ember.js framework production build
+  and will make the assertion a no-op for an application production build.
```
justinaray: Is `make the assertion` a copy/paste mistake from the assert jsdoc?

mixonic: Seems like it should say `deprecation`. Is that really the app-space behavior for deprecate in prod?

packages/ember-debug/lib/index.js
```diff
@@ -102,8 +105,9 @@ setDebugFunction('info', function info() {
   Display a deprecation warning with the provided message and a stack trace
   (Chrome and Firefox only) when the assigned method is called.
 
-  Ember build tools will not remove calls to `Ember.deprecateFunc()`, though
-  no warnings will be shown in production.
+  Ember build tools will remove any calls to `Ember.deprecateFunc()` when
+  doing an Ember.js framework production build and will make the assertion a
+  no-op for an application production build.
```
mixonic: This is not true. The original message seems correct?

packages/ember-debug/lib/warn.js
```diff
@@ -24,8 +24,9 @@ export let missingOptionsIdDeprecation = 'When calling `Ember.warn` you must pro
 */
 
 /**
-  Display a warning with the provided message. Ember build tools will
-  remove any calls to `Ember.warn()` when doing a production build.
+  Display a warning with the provided message. Ember build tools will remove
+  any calls to `Ember.warn()` when doing an Ember.js framework production build
+  and will make the assertion a no-op for an application production build.
```
mixonic: Not an assertion.

packages/ember-debug/lib/deprecate.js
justinaray: I cannot speak for @webark, but I think this was to eventually match the behavior to be implemented by https://github.com/ember-cli/rfcs/pull/50.

We recently came across this when we removed all custom deprecation handlers and the default Ember console.log implementation started to flood our prod consoles.

I took from my slack conversation that this RFC was to land quickly.  Which might be the motivation not to updated the docs to match current behavior only to change them back after impl?

webark:
Ya, I was simply copying and pasting from what was decided in #13180 and seems like some better wording should be done. The motivation for this PR was two things. First, based off of the current documentation, I thought that `Ember.info` would behave the same in production, cause it never said elsewise in the documentation. Second, it does seem based off of the wording that it will actually remove the calls, not just turn them into a empty function.

webark:
I'll update the wording for them though so that it's a little better then a :scissors: :honey_pot: (didn't have a brush or glue. figure honey could be used as glue in a pinch)

webark:
was just basing which methods where 'no-op'ed from this. https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/lib/debug.js I left out the debug seal cause There was no documentation on that as is.

packages/ember-debug/lib/deprecate.js
mixonic: We should definitely not add documentation that is incorrect, but would be if/when some other work lands. Right?

mixonic:
Seems totally great to move forward with! We should just be sure the docs are accurate :-)

webark:
@mixonic @justinaray Updated to hopefully be more strait forward and less wordy. Once ember-cli/rfcs#50 is complete, the documentation could be updated to reflect the eventual implementation.

packages/ember-debug/lib/deprecate.js
```diff
@@ -95,8 +95,9 @@ export let missingOptionsUntilDeprecation = 'When calling `Ember.deprecate` you
 
 /**
   Display a deprecation warning with the provided message and a stack trace
-  (Chrome and Firefox only). Ember build tools will remove any calls to
-  `Ember.deprecate()` when doing a production build.
+  (Chrome and Firefox only).
+
+  * In a production build, this method is defined as an empty function (NOP).
```
mixonic: Can we note both behaviors?

> * In a production build, this method is defined as an empty function (NOP). Uses of this method in Ember itself are stripped from the `ember.prod.js` build.

packages/ember-debug/lib/index.js
```diff
@@ -30,10 +30,9 @@ import { generateTestAsFunctionDeprecation } from 'ember-debug/handlers';
 
 
 /**
-  Define an assertion that will throw an exception if the condition is not
-  met. Ember build tools will remove any calls to `Ember.assert()` when
-  doing an Ember.js framework production build and will make the assertion a
-  no-op for an application production build. Example:
+  Define an assertion that will throw an exception if the condition is not met.
+
+  * In a production build, this method is defined as an empty function (NOP).
```
mixonic: Can we note both behaviors?

> * In a production build, this method is defined as an empty function (NOP). Uses of this method in Ember itself are stripped from the `ember.prod.js` build.

mixonic:
I've noted two spots where we should mention the calls being stripped from production, but you should look to the actual list for confirmation:

* https://github.com/emberjs/ember.js/blob/master/ember-cli-build.js#L38

Thanks! :clap:

webark:
@mixonic updated! And no worries! 😄

mixonic:
Thanks @webark!

Issue #13258, opened 2016-04-05
Assertion Failed: You're not allowed to have more than one controller property map to the same query param key


Arkham:
Hi there,

I just updated one of my apps to v2.4.4 and I saw this error popping up:

```
Error while processing route: other Assertion Failed: You're not allowed to have more
than one controller property map to the same query param key, but both
`application:perPage` and `application:perPage` map to `per_page`. You can fix this
by mapping one of the controller properties to a different query param key via the `as`
config option, e.g. `perPage: { as: 'other-perPage' }`
```

The weird thing is that it appears that the error is referring to the same queryParam. If I revert this commit (https://github.com/emberjs/ember.js/pull/13202) the code works as intended.

Here is an ember twiddle reproducing the error: https://ember-twiddle.com/23c9bc41197b73fc8874a35b04c1a635

The setup I've been running is:
```
Ember      : 2.4.4
Ember Data : 2.4.3
jQuery     : 1.11.3
```

rwjblue:
Thank you for the issue and reproduction! Would you mind submitting a failing test? That way it will be easier to confirm that we squash it properly...

raytiley:
This is because your remapping `perPage` to `per_page`. It seems like the `activeTransition` has it as the form it shows up in the url, but when the query params are finalized it puts them as they appear on the controller. 

Still need to work through this a bit more to see what the correct solution is. I'll at least get a failing test written later tonight.

raytiley:
This fixes the issue for me: https://github.com/emberjs/ember.js/pull/13273

Still need to write tests / make sure I didn't break anything else.

pixelhandler:
This past week I started working on a new app, was on Ember 2.4.3 and also tried LTS release 2.4.5. 

The error I saw is like… "You're not allowed to have more than one controller property map to the same query param key, but both `list.index:sortParam` and `list.index:sortParam` map to `sort`."

pixelhandler:
@rwjblue will the fix for this issue be applied to the LTS release ?

pixelhandler:
@rwjblue @chancancode any update on resolution for LTS version yet?

jcbvm:
Seems like 2.4.3 is also affected by this.

pixelhandler:
@jcbvm I've only seen the issue affecting versions 2.4.4 and 2.4.5, rolling back to 2.4.3 was my workaround.

pixelhandler:
[...]

PR #13273, opened 2016-04-07
[BUGFIX] - Merge query params from previous transition into active transition

raytiley:
This merges in queryParams from an active transition in such a way that `prepareQueryParams` won't duplicate them.

It also won't merge in the qp if it is already int he qp changed list, for example if it's already been set during route setup by a call to `controller.set('qpProp', 'someValue')`.

Fixing these things with my current test jsbin also recreated another issue that has been biting people and that is that a call to `transitionTo` creates an aborted transition that calls `didTransition` with an empty set of handlerInfos. This would cause the `updatePaths` method to fail since it couldn't calculate the paths without the handler infos. We now bail early from `updatePaths` it is called with an empty array of handlerInfos.

I think this is fine because the aborted transition is followed by a successful transition that will update the paths appropriatly.

Still todo - TESTS

### DIFF
--- a/packages/ember-routing/lib/system/router.js
+++ b/packages/ember-routing/lib/system/router.js
@@ -717,11 +717,6 @@ const EmberRouter = EmberObject.extend(Evented, {
     assert(`The route ${targetRouteName} was not found`, targetRouteName && this.router.hasRoute(targetRouteName));
 
     let queryParams = {};
-    // merge in any queryParams from the active transition which could include
-    // queryparams from the url on initial load.
-    if (this.router.activeTransition) {
-      assign(queryParams, this.router.activeTransition.queryParams);
-    }
 
     this._processActiveTransitionQueryParams(targetRouteName, models, queryParams, _queryParams);
 
@@ -738,7 +733,7 @@ const EmberRouter = EmberObject.extend(Evented, {
 
   _processActiveTransitionQueryParams(targetRouteName, models, queryParams, _queryParams) {
     // merge in any queryParams from the active transition which could include
-    // queryparams from the url on initial load.
+    // queryParams from the url on initial load.
     if (!this.router.activeTransition) { return; }
 
     var unchangedQPs = {};
@@ -749,9 +744,9 @@ const EmberRouter = EmberObject.extend(Evented, {
       }
     }
 
-    // We need to fully scope query params so that we can create one object
-    // that represetns both pased in query params and ones that arent' changed
-    // from the actice transition
+    // We need to fully scope queryParams so that we can create one object
+    // that represents both pased in queryParams and ones that aren't changed
+    // from the active transition.
     this._fullyScopeQueryParams(targetRouteName, models, _queryParams);
     this._fullyScopeQueryParams(targetRouteName, models, unchangedQPs);
     assign(queryParams, unchangedQPs);
--- a/packages/ember-routing/lib/system/router.js
+++ b/packages/ember-routing/lib/system/router.js
@@ -723,6 +723,8 @@ const EmberRouter = EmberObject.extend(Evented, {
       assign(queryParams, this.router.activeTransition.queryParams);
     }
 
+    this._processActiveTransitionQueryParams(targetRouteName, models, queryParams, _queryParams);
+
     assign(queryParams, _queryParams);
     this._prepareQueryParams(targetRouteName, models, queryParams);
 
@@ -732,6 +734,27 @@ const EmberRouter = EmberObject.extend(Evented, {
     didBeginTransition(transition, this);
 
     return transition;
+  },
+
+  _processActiveTransitionQueryParams(targetRouteName, models, queryParams, _queryParams) {
+    // merge in any queryParams from the active transition which could include
+    // queryparams from the url on initial load.
+    if (!this.router.activeTransition) { return; }
+
+    var unchangedQPs = {};
+    var qpUpdates = this._qpUpdates || {};
+    for (var key in this.router.activeTransition.queryParams) {
+      if (!qpUpdates[key]) {
+        unchangedQPs[key] = this.router.activeTransition.queryParams[key];
+      }
+    }
+
+    // We need to fully scope query params so that we can create one object
+    // that represetns both pased in query params and ones that arent' changed
+    // from the actice transition
+    this._fullyScopeQueryParams(targetRouteName, models, _queryParams);
+    this._fullyScopeQueryParams(targetRouteName, models, unchangedQPs);
+    assign(queryParams, unchangedQPs);
   },
 
   _prepareQueryParams(targetRouteName, models, queryParams) {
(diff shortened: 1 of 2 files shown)


packages/ember-routing/lib/system/router.js
```diff
@@ -636,7 +636,17 @@ var EmberRouter = EmberObject.extend(Evented, {
     // merge in any queryParams from the active transition which could include
     // queryparams from the url on initial load.
     if (this.router.activeTransition) {
-      assign(queryParams, this.router.activeTransition.queryParams);
+      var unchangedQPs = {};
```
rwjblue: Can this bit of code live in a separate method (`_processActiveTransitionQueryParams` or something)?

rwjblue:
Thanks for working on this @raytiley, let me know if I can help...

raytiley:
Come to Maine and babysit for a few hours?  I should be able to wrap this
up tonight.

rwjblue:
> Come to Maine and babysit for a few hours? I should be able to wrap this up tonight.

Sounds fun! Might have to make a pit stop at the Bissell Brothers though....

knownasilya:
@raytiley I'm up to help any way I can, although visiting to babysit might be out of the question.

packages/ember-routing/lib/system/router.js
```diff
@@ -694,6 +704,33 @@ var EmberRouter = EmberObject.extend(Evented, {
...
+    stashParamNames(this, handlerInfos);
+
+    for (var i = 0, len = handlerInfos.length; i < len; ++i) {
+      var route = handlerInfos[i].handler;
+      var qpMeta = get(route, '_qp');
+
+      for (var j = 0, qpLen = qpMeta.qps.length; j < qpLen; ++j) {
+        var qp = qpMeta.qps[j];
+
+        var presentProp = qp.prop in queryParams  && qp.prop ||
```
igorT: Can you please break this down and add more local vars to make this more readable

raytiley:
Hey Everyone... sorry. I've been really busy at work. I promise I will clean this up / get tests this week.

igorT:
I've been testing this PR in our app and it seems to fix certain query param issues that broke in 2.4.4 but introduces another bug where initial load of a page with a query 404s. Investigating

packages/ember-routing/lib/system/router.js
```diff
@@ -694,6 +704,33 @@ var EmberRouter = EmberObject.extend(Evented, {
     };
   },
 
+  _normalizeQueryParams(leafRouteName, contexts, queryParams) {
+    var state = calculatePostTransitionState(this, leafRouteName, contexts);
+    var handlerInfos = state.handlerInfos;
+    var appCache = this._bucketCache;
```
igorT: This is unused

raytiley:
First steps... I rebased the code and fixed the jshint error. Tomorrow I'll write some tests based on some jsbin cases I used to write this code. Sorry for the delay.

igorT:
@raytiley I added some tests here for things that worked in 2.4.3 and broke in 2.4.4 I think your branch fixed them as of yesterday. 
https://github.com/emberjs/ember.js/commit/53fcdbc6376981d8a219cde00d04bd18d59cac9d

packages/ember/tests/routing/query_params_test.js
```diff
@@ -2490,6 +2490,43 @@ if (isEnabled('ember-routing-route-configured-query-params')) {
     bootApplication();
   });
 
+  QUnit.test('queryParams are updated when a controller property is set and the route is refreshed. Issue #13263  ', function() {
```
pixelhandler: This test looks good!

Seems to me that this test will fail with "Uncaught Error, Assertion Failed: You're not allowed to have more than one controller property map to the same query param key…"  without the code to fix.

lolmaus:
This issue needs to be properly tagged. At least with `Query Params`.

lolmaus:
@Sinled has discovered a regression with an array query param.

Previously, you could have an array query param like this:

```js
  queryParams: ['fields',],
  fields: ['field1', 'field2']
})
```

Ember would properly (and transparently!) serialize them into strings and then deserialze back.

After the regression, array query params are not deserialized and the Controller's array property gets rewritten with a string.

Me and @Sinled managed to narrow down the regression to this commit by @raytiley: `master` https://github.com/emberjs/ember.js/commit/2b42d5659498df2bf725f09e65632bfe5c3161c0, `lts-2-4` https://github.com/emberjs/ember.js/commit/7460b686d67eb349068482ddad131cd3c76c3dc1.

Please tell us whether:

* this is a bug or works as designed;
* this is a known or new issue;
* it should be reported as a new GitHub issue ticket.

rwjblue:
> This issue needs to be properly tagged. 

This comment is not helpful, please keep things constructive.

> Please tell us whether:
> 
> * this is a bug or works as designed;

Bug.

> * this is a known or new issue;

If you didn't find it via a GitHub search, it is a new issue.  I personally do not recall an issue opened about 2.4.4 regarding array query params.

> * it should be reported as a new GitHub issue ticket.

If you didn't find it via a GitHub search, it is a new issue.  Please test this PR to see if it fixes.

packages/ember-routing/lib/system/router.js
```diff
@@ -987,6 +1032,8 @@ function calculatePostTransitionState(emberRouter, leafRouteName, contexts) {
 
 function updatePaths(router) {
   let infos = router.router.currentHandlerInfos;
+  if (infos.length === 0) { return; }
```
mattmarcum: I'm running into a similar problem but in my case `infos` isn't even an array.  I did something like:

```
import { isArray } from 'ember-runtime/utils';
...
if (!isArray(infos) || infos.length === 0) { return; }
```

knownasilya:
I'm totally getting bit by the QPs not being deserialized properly (arrays) in the `queryParamsDidChange` hook. How's this PR going, and would it solve that?

pixelhandler:
@stefanpenner @chancancode question for you, do we need a "regression" label on this?

pixelhandler:
@raytiley thanks for putting this PR into play. This will help people that would like to adopt the LTS release. I spoke to a few people at our meetup Wed. And noticed those interested in LTS are running 2.4.3 instead of 2.4.5 due to #13258

lolmaus:
Related: https://github.com/emberjs/ember.js/issues/13591

Sinled:
2.4.6 fixed my [problem](https://github.com/emberjs/ember.js/pull/13273#issuecomment-214739521l) with query params serialisation, but it still present in 2.6.0

rwjblue:
Confirm. This PR is still needed. 

@igorT - Were you going to pick this up?

rwjblue:
🎉

nathanhammond:
Hey everybody who paid attention to this thread! We've a few open bugs likely triggered by this change. Anybody have time to dive in?

PR #13274, opened 2016-04-07
Make computed.or and computed.and support property expansion

ming-codes:
`computed.or` and `computed.and` already supports more than two properties. I added some test cases for this use case.

Added code to pass `getProperties` through `Ember.expandProperties` to further enhance the above case.

### DIFF
--- a/packages/ember-metal/lib/computed_macros.js
+++ b/packages/ember-metal/lib/computed_macros.js
@@ -14,11 +14,15 @@ import expandProperties from 'ember-metal/expand_properties';
 
 function getProperties(self, propertyNames) {
   var ret = {};
-  for (var i = 0; i < propertyNames.length; i++) {
-    expandProperties(propertyNames[i], function (name) {
-      ret[name] = get(self, name);
-    });
+
+  function extractProperty(name) {
+    ret[name] = get(self, name);
   }
+
+  for (var i = 0; i < propertyNames.length; i++) {
+    expandProperties(propertyNames[i], extractProperty);
+  }
+
   return ret;
 }
 
--- a/packages/ember-metal/lib/computed_macros.js
+++ b/packages/ember-metal/lib/computed_macros.js
@@ -413,22 +413,30 @@ export function lte(dependentKey, value) {
   A computed property that performs a logical `and` on the
   original values for the provided dependent properties.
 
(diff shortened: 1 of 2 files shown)


packages/ember-metal/lib/computed_macros.js
```diff
@@ -14,7 +15,9 @@ import alias from 'ember-metal/alias';
 function getProperties(self, propertyNames) {
   var ret = {};
   for (var i = 0; i < propertyNames.length; i++) {
-    ret[propertyNames[i]] = get(self, propertyNames[i]);
+    expandProperties(propertyNames[i], function (name) {
```
rwjblue: This is triggering a JSHint failure for "don't make functions in a loop"

ming-codes:
@rwjblue All tests passed

rwjblue:
Looks good, can you update the docs for `computed.or` and `computed.and` to add an example using brace expansion and also squash commits?

rwjblue:
Hmm, actually, I just had a thought.  It seems like we should call `expandProperties` upon creation of the macro (inside `generateComputedWithProperties` [around here](https://github.com/ming-codes/ember.js/blob/computed-property-expansion/packages/ember-metal/lib/computed_macros.js#L31)).  The current implementation seems like it would be calling `expandProperties` on every dependent key invalidation (which seems like a bit of wasted work).

ming-codes:
`computed.and` doesn't seem to follow the JavaScript-like behavior of `computed.or`. As in, `computed.and` always coerce to `false` rather than the original DK's falsy value. I wonder if this is a bug so I can fix it.

Test case below

```js
  var obj = { one: true, two: true };
  defineProperty(obj, 'oneAndTwo', and('one', 'two'));

  set(obj, 'one', null);
  set(obj, 'two', 'Yes');

  equal(get(obj, 'oneAndTwo'), null, 'returns falsy value as in &&'); // expects `null` as JavaScript expression `null && 'Yes'` eval to `null`
```

rwjblue:
Yes, I believe that it should behave the same as `computed.or` (and normal JS)

ming-codes:
@rwjblue Done. I think github let you squash and merge now. Or I can do it. Let me know.

rwjblue:
@ming-codes thank you!

PR #13017, opened 2016-02-26
[FEATURE ember-improved-instrumentation] Rebase + Tests

chadhietala:
Rebase and Adds Test For #12550. However this just introduces **action** instrumentation. There are still issues around instrumenting the router. I have saved off the diff for the time being. https://gist.github.com/chadhietala/2b3065f3da3d8ae6e2ff

PR #13279, opened 2016-04-08
Add action and dom event instrumentation

asakusuma:
This is just the action and dom instrumentation part copied and merged over from @chadhietala's PR here: https://github.com/emberjs/ember.js/pull/13017

This is the highest priority part for getting us off of private APIs for tracking.

I noticed that some of the tests seem to be disabled for glimmer2, so I'm wondering what I need to do to get this working and tested with glimmer2.

cc @krisselden

### DIFF
--- a/features.json
+++ b/features.json
@@ -9,6 +9,7 @@
     "ember-htmlbars-local-lookup": true,
     "ember-application-engines": null,
     "ember-glimmer": null,
-    "ember-runtime-computed-uniq-by": null
+    "ember-runtime-computed-uniq-by": null,
+    "ember-improved-instrumentation": null
   }
 }
--- a/packages/ember-metal/lib/instrumentation.js
+++ b/packages/ember-metal/lib/instrumentation.js
@@ -1,4 +1,5 @@
 import Ember from 'ember-metal/core';
+import isEnabled from 'ember-metal/features';
 
 /**
   The purpose of the Ember Instrumentation module is
@@ -105,14 +106,26 @@ export function instrument(name, _payload, callback, binding) {
   }
 }
 
+var flaggedInstrument;
+if (isEnabled('ember-improved-instrumentation')) {
+  flaggedInstrument = instrument;
+} else {
+  flaggedInstrument = function(name, payload, callback) {
+    return callback();
+  };
+}
+export { flaggedInstrument };
+
 function withFinalizer(callback, finalizer, payload, binding) {
+  let result;
   try {
-    return callback.call(binding);
+    result = callback.call(binding);
   } catch(e) {
     payload.exception = e;
-    return payload;
+    result = payload;
   } finally {
-    return finalizer();
+    finalizer();
+    return result;
   }
 }
 
@@ -151,7 +164,9 @@ export function _instrumentStart(name, _payload) {
     var timestamp = time();
     for (i = 0, l = listeners.length; i < l; i++) {
       listener = listeners[i];
-      listener.after(name, timestamp, payload, beforeValues[i]);
+      if (typeof listener.after === 'function') {
+        listener.after(name, timestamp, payload, beforeValues[i]);
+      }
     }
 
     if (STRUCTURED_PROFILE) {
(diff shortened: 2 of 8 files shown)


features.json
```diff
@@ -8,6 +8,7 @@
     "ember-metal-ember-assign": true,
     "ember-htmlbars-local-lookup": true,
     "ember-application-engines": null,
-    "ember-glimmer": null
+    "ember-glimmer": null,
+    "ember-improved-instrumentation": true
```
mixonic: This should be `null` in the PR.

packages/ember-metal/lib/instrumentation.js
```diff
@@ -105,6 +106,38 @@ export function instrument(name, _payload, callback, binding) {
...
+  to call the finalizer that this function returns at some point
+  in the future.
+  @public
+  @method asyncInstrument
+  @param {String} [name] Namespaced event name
+  @param {Function} _payload
+  @return {Function} a finalizer function to call when the instrumented
+    work has finished.
+*/
+export function asyncInstrument(name, _payload) {
```
mixonic: Where is this used?

packages/ember-routing-htmlbars/tests/helpers/closure_action_test.js
```diff
@@ -583,3 +584,38 @@ QUnit.test('objects that define INVOKE can be casted to actions', function(asser
...
 
 }
+
+if (isEnabled('ember-improved-instrumentation')) {
+  QUnit.test('action should fire interaction event', function(assert) {
+    assert.expect(2);
+
+    subscribe('interaction.ember-action', {
+      before() {
+        console.log(arguments);
```
mixonic: `console.log`

packages/ember-routing-htmlbars/tests/helpers/closure_action_test.js
```diff
@@ -583,3 +584,38 @@ QUnit.test('objects that define INVOKE can be casted to actions', function(asser
...
+    });
+
+    innerComponent = EmberComponent.extend({
+      fireAction() {
+        this.attrs.submit();
+      }
+    }).create();
+
+    outerComponent = EmberComponent.extend({
+      layout: compile('{{view innerComponent submit=(action outerSubmit)}}'),
```
mixonic: Would be nice to not use the `view` helper in new tests.

packages/ember-routing-htmlbars/tests/helpers/closure_action_test.js
```diff
@@ -583,3 +584,38 @@ QUnit.test('objects that define INVOKE can be casted to actions', function(asser
...
+  QUnit.test('action should fire interaction event', function(assert) {
+    assert.expect(2);
+
+    subscribe('interaction.ember-action', {
+      before() {
+        console.log(arguments);
+        assert.ok(true, 'instrumentation subscriber was called');
+      },
+      after() {
+      }
```
mixonic: Empty `after() {` can probably be dropped.

asakusuma: Removed

packages/ember-routing-htmlbars/tests/helpers/closure_action_test.js
asakusuma: I'm assuming the right thing to use would be a component helper, though I'm unsure the best way to do a component setup in this test. Should I try to use [RenderingTest](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/tests/integration/components/dynamic-components-test.js#L7)? Or perhaps do some custom setup like in [local-lookup-test.js](https://github.com/emberjs/ember.js/blob/master/packages/ember-htmlbars/tests/integration/local-lookup-test.js)?

chadhietala: I think you would need to introduce the `owner` stuff that is in the local-look-test. Basically you need a means of registering.

PR #13283, opened 2016-04-08
[Glimmer 2] Fix undefined path lookups

chadhietala:
This fixes issues where `get` would error if the context side was
undefined.

### DIFF
--- a/packages/ember-glimmer/lib/utils/references.js
+++ b/packages/ember-glimmer/lib/utils/references.js
@@ -75,9 +75,12 @@ class PropertyReference extends CachedReference { // jshint ignore:line
     let { _parentReference, _parentObjectTag, _propertyKey } = this;
 
     let parent = _parentReference.value();
-    _parentObjectTag.update(tagFor(parent));
 
-    return get(parent, _propertyKey);
+    if (parent) {
+      _parentObjectTag.update(tagFor(parent));
+
+      return get(parent, _propertyKey);
+    }
   }
 
   get(propertyKey) {
--- a/packages/ember-glimmer/lib/utils/references.js
+++ b/packages/ember-glimmer/lib/utils/references.js
@@ -74,12 +74,14 @@ class PropertyReference extends CachedReference { // jshint ignore:line
   compute() {
     let { _parentReference, _parentObjectTag, _propertyKey } = this;
 
-    let parent = _parentReference.value();
+    let parentValue = _parentReference.value();
 
-    if (parent) {
-      _parentObjectTag.update(tagFor(parent));
+    _parentObjectTag.update(tagFor(parentValue));
 
-      return get(parent, _propertyKey);
+    if (parentValue && typeof parentValue === 'object') {
+      return get(parentValue, _propertyKey);
+    } else {
+      return null;
     }
   }
 
--- a/packages/ember-glimmer/tests/integration/syntax/each-in-test.js
+++ b/packages/ember-glimmer/tests/integration/syntax/each-in-test.js
@@ -1,7 +1,7 @@
 import { set } from 'ember-metal/property_set';
 import { strip } from '../../utils/abstract-test-case';
 import { applyMixins } from '../../utils/abstract-test-case';
-import { moduleFor } from '../../utils/test-case';
+import { moduleFor, RenderingTest } from '../../utils/test-case';
 import {
   BasicConditionalsTest,
   SyntaxCondtionalTestHelpers,
@@ -155,4 +155,27 @@ moduleFor('@htmlbars Syntax test: {{#each-in}}', class extends EachInTest {
     `);
   }
 
+});
+
+moduleFor('@htmlbars Syntax test: {{#each-in}} undefined path', class extends RenderingTest {
+  ['@test keying off of `undefined` does not render'](assert) {
+    this.render(strip`
+      {{#each-in foo.bar.baz as |thing|}}
+        {{thing}}
+      {{/each-in}}`, { foo: {} });
+
+    this.assertText('');
+
+    this.runTask(() => this.rerender());
+
+    this.assertText('');
+
+    this.runTask(() => set(this.context, 'foo', { bar: { baz: { 'Here!': 1 } } }));
+
+    this.assertText('Here!');
+
+    this.runTask(() => set(this.context, 'foo', {}));
+
+    this.assertText('');
+  }
 });
--- a/packages/ember-glimmer/tests/integration/content-test.js
+++ b/packages/ember-glimmer/tests/integration/content-test.js
@@ -483,4 +483,89 @@ moduleFor('Dynamic content tests (integration)', class extends RenderingTest {
     this.assertHTML(ember);
   }
 
+  ['@test it should evaluate to nothing if part of the path is `undefined`']() {
+    this.render('{{foo.bar.baz.bizz}}', {
+      foo: {}
+    });
+
+    this.assertText('');
+
+    this.runTask(() => this.rerender());
+
+    this.assertText('');
+
+    this.runTask(() => set(this.context, 'foo', {
+      bar: { baz: { bizz: 'Hey!' } }
+    }));
+
+    this.assertText('Hey!');
+
+    this.runTask(() => set(this.context, 'foo', {}));
+
+    this.assertText('');
+
+    this.runTask(() => set(this.context, 'foo', {
+      bar: { baz: { bizz: 'Hello!' } }
+    }));
+
+    this.assertText('Hello!');
+
+    this.runTask(() => set(this.context, 'foo', {}));
+
+    this.assertText('');
+  }
+
+  ['@test it should evaluate to nothing if part of the path is a primative']() {
+    this.render('{{foo.bar.baz.bizz}}', {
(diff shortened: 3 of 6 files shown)


chadhietala:
This is ready for review /cc @chancancode @krisselden.

packages/ember-glimmer/lib/utils/references.js
```diff
@@ -75,9 +75,12 @@ class PropertyReference extends CachedReference { // jshint ignore:line
...
 
     let parent = _parentReference.value();
-    _parentObjectTag.update(tagFor(parent));
 
-    return get(parent, _propertyKey);
+    if (parent) {
+      _parentObjectTag.update(tagFor(parent));
+
+      return get(parent, _propertyKey);
+    }
```
krisselden: do you not have to do anything with the tag in the else case?

packages/ember-glimmer/tests/integration/syntax/each-in-test.js
```diff
@@ -156,3 +156,26 @@ moduleFor('@htmlbars Syntax test: {{#each-in}}', class extends EachInTest {
...
+
+    this.assertText('');
+
+    this.runTask(() => set(this.context, 'foo', { bar: { baz: { 'Here!': 1 } } }));
+
+    this.assertText('Here!');
+
+    this.runTask(() => set(this.context, 'foo', {}));
+
+    this.assertText('');
```
krisselden: can you check that it can become something again? related to my other comment

packages/ember-glimmer/lib/utils/references.js
chadhietala: I confirmed locally that updating again works. I'm not sure if you would need to update the tag because compute would be idempotent operation until you have a truthy value. That brings up another point, can the path ref here ever evaluate to something that is not getable e.g. `true`?

chadhietala:
@krisselden Not sure what the steps forward are here?

chancancode:
I can answer your questions when I get home! :D

chancancode:
I can explain how the reference chain works. I'll assume you have already read the [references](https://github.com/tildeio/glimmer/blob/master/guides/04-references.md) and [validators](https://github.com/tildeio/glimmer/blob/master/guides/05-validators.md) guides.

For `{{foo.bar.baz}}`, we will produce...

```javascript
let selfRef = new RootReference(component);
let fooRef = selfRef.get("foo"); // == new PropertyReference(selfRef, "foo");
let barRef = fooRef.get("bar"); // == new PropertyReference(selfRef, "bar");
let bazRef = fooRef.get("baz"); // == new PropertyReference(selfRef, "baz");

// pseudo code
let textNode = append(bazRef); 
update(textNode, bazRef);
```

(For conditionals and loops, we further wrap `bazRef` with a conditional reference or an iterator. However, even a simple content curly `{{something.is.undefined.along.the.way}}` should exhibit the same problem, and we should have a test for that. For the rest of this I'm going to focus on that scenario since conditionals and loops are just a simple extension.)

So the tag on a reference are guaranteeing that calling `.value()` again will give you an equivalent value so long as the tag's `value()` didn't change. This doesn't work the other way around – if the tag's `value()` has changed, you might still get back an equivalent value. Also, an "equivalent" value doesn't necessarily mean `===`. For example, a helper reference (say the `hash` helper) might choose to recompute (joining the string parts) every time you call its `value()`, in which case they might return two different objects, but they are still semantically "equivalent".

So how do we know if `foo.bar.baz` has changed?

Let's start with `selfRef`.

In Ember, the `self` (`this`) in a template cannot change between initial render and updates – it's always going to be the current component instance). Therefore, the answer is easy – `selfRef` will always return the same `value()` (the component instance). So we can give a `CONSTANT_TAG` to `selfRef`.

Moving on to `fooRef`.

From the perspective of the `PropertyReference`, we are just given some `parentReference`. It just so happens that it is a `RootReference`, but we don't know that. Fortunately, the entity tag system is designed to be compassable, so we can just compose the parent reference's tag.

We can start with this:

```typescript
class PropertyReference implements PathReference {
  private parentRef: PathReference;
  private propertyPath: string;

  constructor(parentRef: PathReference, propertyPath: string) {
    this.tag = parentRef.tag;
    this.parentRef = parentRef;
    this.propertyPath = propertyPath;
  }

  value(): any {
    let { parentRef, propertyPath } = this;
    let parentValue = parentRef.value();

    if (parentValue && typeof parentValue === 'object') { 
      return Ember.get(parentValue, propertyPath);
    } else {
      return null;
    }
  }
}
```

Unfortunately, this does not work. Even when `parentValue` is the same object (`parentRef.tag.value()` and `parentRef.value()` didn't change – and in our specific case here, it never will), `Ember.get(parentValue, propertyPath)` might still return a different value if someone has done `Ember.set(parentValue, propertyPath, newValue)` at some point.

Fortunately, this is also possible to capture this in the property reference's tag. In Ember, each object has its own tag that guarantees (sometimes I like to think of it as "underwrites") its interior mutations – if the object's tag has not changed, then `Ember.get(taggedObject, "somePathWithoutADot")` also cannot change. (The implementation is that each time you call `Ember.set(taggedObject, "somePathWithoutADot", ...)`, the tag will be incremented to the current revision.)

So we can try this:

```typescript
class PropertyReference implements PathReference {
  private parentRef: PathReference;
  private propertyPath: string;

  constructor(parentRef: PathReference, propertyPath: string) {
    let parentValue = parentRef.value();

    this.tag = combine(parentRef.tag, tagFor(parentValue));
    this.parentRef = parentRef;
    this.propertyPath = propertyPath;
  }

  value(): any {
    let { parentRef, propertyPath } = this;
    let parentValue = parentRef.value();

    if (parentValue && typeof parentValue === 'object') { 
      return Ember.get(parentValue, propertyPath);
    } else {
      return null;
    }
  }
}
```

This almost works, but it has two problems:

First, it eagerly pulls the `value()` from the parent reference in the constructor. This is actually not a big deal most of the time – when you construct a reference you are probably about to put it into a DOM, so you would have done the same thing in the following `value()` call anyway. (It does the work twice, but the parent reference probably have a cache anyway.) However, this is undesirable for things like helper references with lazy semantics for one reason or another.

But more importantly, the bigger problem here is `parentValue` can of course change over time, that's the whole point of passing in a parent reference. (We happen to know in our very specific case it can't change, but in general it is expected to change over time.) Therefore, saving off the tag for the object representing a particular value in a particular "time slice" is not really appropriate – you will start to notice incorrect tag changes that are completely unrelated to the new value, or if the old value gets GC'ed your consumers might mistakenly believe the reference is not changing.

As an aside, this is actually a nice thing about the entity tag system – because we are holding onto a stale object's tag instead of the stale object itself, and the tag does not have a pointer to the object, the worst we damage we can do (besides getting incorrect change semantics as I mentioned above) is leaking a stale tag, as opposed to leaking an entire object tree.

To solve both of these problems, we can use an updatable tag instead:

```typescript
class PropertyReference implements PathReference {
  private parentValueTag: UpdatableTag;
  private parentRef: PathReference;
  private propertyPath: string;

  constructor(parentRef: PathReference, propertyPath: string) {
    this.parentValueTag = new UpdatableTag(CURRENT_TAG);

    this.tag = combine(parentRef.tag, parentValueTag);
    this.parentRef = parentRef;
    this.propertyPath = propertyPath;
  }

  value(): any {
    let { parentValueTag, parentRef, propertyPath } = this;
    let parentValue = parentRef.value();

    parentValueTag.update(tagFor(parentValue));

    if (parentValue && typeof parentValue === 'object') {
      return Ember.get(parentValue, propertyPath);
    } else {
      return null;
    }
  }
}
```

The idea is that in the constructor, we don't know what the current `parentValue` is yet, and we don't want to pull it eagerly if we don't end up using it. However, we know `parentValue` is "within the boundaries" of the Ember object model, so we can give it the `CURRENT_TAG`. It basically means "if *anything* in the system changed, assume `parentValue` might have changed", which is a fair guess without actually knowing what the value is.

When we call `value()` on the property reference, we will have a chance to look at the current `parentValue`. We will take advantage of that and update the `parentValueTag` to the actual tag on the object – essentially "refining" our estimation.

And this happens recursively for the rest of `{{foo.bar.baz}}`.

* * *

To answer your original question, what happens when `parentValue` becomes "un`get`table"? i.e. `null`, `undefined`, `true`, `"zomg"`, etc. The answer is that since primitive values does not have "members" (maybe with the exception of `{{someString.length}}`), so will never go through an "interior mutation". Therefore, we can update the `parentValue` tag with a `CONSTANT_TAG`. (I believe `tagFor` is already smart enough to return that for primitive values, so we can just unconditionally update the tag with `tagFor(parentValue)`.)

In other words, the property reference will always return `null` so long as the parent reference didn't change.

The current implementation in this PR doesn't update the tag. This means that we might incorrectly believe the `PropertyReference` has changes even when the parent reference didn't change (we saved off a stale tag from an old object – that object might still be alive somewhere else and going through interior mutations).

* * * 

TODO items here:

* [ ] Always update the tag for the parent value
* [ ] Add a test case for `{{something.is.undefined.along.the.way}}`
* [ ] Add a test case for `{{something.is.a.primitive.value.along.the.way}}` (i.e. a string, a number, `true`, etc)

Issue #13284, opened 2016-04-09
`Ember.A(null)` now returns `null` instead of `[]`


workmanw:
Here is a pretty clear snippet that explains the issue:

```javascript
// Ember 2.4.x
Ember.A(null);   // returns []

// Ember Beta
Ember.A(null);   // returns null
```

So what happened? It looks like this commit broke it: https://github.com/emberjs/ember.js/commit/586d4c6ab0a82abee0df1f331121e9c99433770e#diff-7 :(. It seems very harmless, but that code is transpiled as:

```javascript
A = A = function () {
  var arr = arguments.length <= 0 || arguments[0] === undefined ? [] : arguments[0];
  return arr;
};
```

The problem is that because `Ember.A(null)` has an argument length greater than 1, it's value [`null`] is ultimately returned. 

Unfortunately this did break a few tests in our application. Specifically with the fixture adapter, see here: https://github.com/emberjs/ember-data-fixture-adapter/blob/master/addon/index.js#L298-L300 . I believe this to be part of the public API and should be fixed. If so, I'd be happy to add a test and take care of it. 

-------

EDIT: As a side note, I feel like I should voice some my concern about the aforementioned commit ( https://github.com/emberjs/ember.js/commit/586d4c6ab0a82abee0df1f331121e9c99433770e ). There are several other places where `argValue || []` was replaced with default arguments, potentially leading to the same issue.

mmun:
Yep. This should totally be fixed and a tested added.

Out of curiosity, which version of Ember.A(null) do you think makes more sense?

@thoov can you do a quick audit of what else might have broken?

workmanw:
> Out of curiosity, which version of Ember.A(null) do you think makes more sense?

@mmun I was just asking myself that. I read the API docs to see if they indicated one way or another, but it was still kind of ambiguous. I could probably make an argument either way.

I sometimes use `Ember.A` as a safety guard for functions expecting a list argument. Eg:

```javascript

mapListValues(list) {
  return Ember.A(list).map(item => { /* ... */ });
})
```

But I could be convinced that this isn't the purpose of `Ember.A`.

workmanw:
This issue was fixed with #13287 being merged in.

rwjblue:
Thanks for catching this @workmanw!

workmanw:
@rwjblue My pleasure! 

[...]

PR #13287, opened 2016-04-09
[BUGFIX beta] Fix a recent regression with `Ember.A`

workmanw:
Due to a recent regression, `Ember.A(null)` would return `null` instead of `[]`. See issue #13284.

### DIFF
--- a/packages/ember-runtime/lib/system/native_array.js
+++ b/packages/ember-runtime/lib/system/native_array.js
@@ -138,9 +138,10 @@ var A;
 
 if (Ember.EXTEND_PROTOTYPES === true || Ember.EXTEND_PROTOTYPES.Array) {
   NativeArray.apply(Array.prototype);
-  A = function (arr = []) { return arr; };
+  A = function (arr) { return arr || []; };
 } else {
-  A = function (arr = []) {
+  A = function (arr) {
+    if (!arr) { arr = []; }
     return EmberArray.detect(arr) ? arr : NativeArray.apply(arr);
   };
 }
--- a/packages/ember-runtime/tests/system/native_array/a_test.js
+++ b/packages/ember-runtime/tests/system/native_array/a_test.js
@@ -1,0 +1,12 @@
+import EmberArray from 'ember-runtime/mixins/array';
+import { A } from 'ember-runtime/system/native_array';
+
+QUnit.module('Ember.A');
+
+QUnit.test('Ember.A', function() {
+  deepEqual(A([1, 2]), [1, 2], 'array values were not be modified');
+  deepEqual(A(), [], 'returned an array with no arguments');
+  deepEqual(A(null), [], 'returned an array with a null argument');
+  ok(EmberArray.detect(A()), 'returned an ember array');
+  ok(EmberArray.detect(A([1, 2])), 'returned an ember array');
+});


packages/ember-runtime/lib/system/native_array.js
```diff
@@ -138,9 +138,10 @@ var A;
 
 if (Ember.EXTEND_PROTOTYPES === true || Ember.EXTEND_PROTOTYPES.Array) {
   NativeArray.apply(Array.prototype);
-  A = function (arr = []) { return arr; };
+  A = function (arr = []) { return arr || []; };
 } else {
   A = function (arr = []) {
+    if (arr === null) { arr = []; }
```
mixonic: In the native version, you rely on truthiness as a check for `arr`, but in this version you look explicitly for `null` instead of `!arr`. Is there a reason for the difference?

workmanw: @mixonic Yea that's a good point. I think I was just trying to be more explicit about what was happening. Someone reading the following might feel that it's redundant:

```javascript
A = function (arr = []) {
  if (!arr) { arr = []; }
  return EmberArray.detect(arr) ? arr : NativeArray.apply(arr);
};
```

But it's definitely inconsistent between the two. So let's fix this, which do you think is prefered?

mixonic: Dunno- I would defer to the old behavior since this is a regression fix. Can you perhaps try it out on 2.4 or and older version and simply match that? Thanks man.

workmanw: @mixonic Ha. It turns out the old behavior was inconsistent too. https://github.com/emberjs/ember.js/commit/586d4c6ab0a82abee0df1f331121e9c99433770e#diff-7 I think I'll just use the truthiness check and add a code comment. Then see what others think.

mixonic: :sob: truthiness was the mode with prototype extensions enabled (which is the most common), so I agree normalizing around that as a BUGFIX makes sense :+1:

packages/ember-runtime/lib/system/native_array.js
```diff
@@ -138,9 +138,11 @@ var A;
 
 if (Ember.EXTEND_PROTOTYPES === true || Ember.EXTEND_PROTOTYPES.Array) {
   NativeArray.apply(Array.prototype);
-  A = function (arr = []) { return arr; };
+  A = function (arr = []) { return arr || []; };
```
krisselden: why bother then with the default arg?

packages/ember-runtime/lib/system/native_array.js
```diff
@@ -138,9 +138,11 @@ var A;
...
 } else {
   A = function (arr = []) {
+    // This may seem redudant, but `A(null)` should return an array.
+    if (!arr) { arr = []; }
```
krisselden: same here

krisselden:
the default arg is redundant to handling an explicitly undefined or null

krisselden:
also, the comment is unneeded with the regression test.

krisselden:
This regression is unreleased?

workmanw:
Chatted with @krisselden briefly on Slack. We agreed that keeping the default args was unnecessary in this case. As a result of that I removed the code comment.

This affects beta, it does not affect any release.

mixonic:
Thanks @workmanw

workmanw:
Anytime! 😄

PR #13290, opened 2016-04-10
[BUGFIX beta] Use a symbol to detect if an object is a stream.

### LINKED ISSUE #13271: `isStream` should be flagged as a reserved property on Components
[See this twiddle for a reduced test case.](https://ember-twiddle.com/02b921550f59393fbd6ce997253ce424?fileTreeShown=false&numColumns=2&openFiles=components.my-component.js%2Ctemplates.components.my-component.hbs) Open the console to see the error message:
```
Uncaught TypeError: object.subscribe is not a function
```

If you set `isStream` as a property on a Component (at least, maybe other classes too), rendering borks because `isStream` is used internally. I can put together a failing test and possibly a  PR, just wanted to note it here in case anyone else wants to pick it up.

I'm imagining a helpful error message at run time that just states that `isStream` is a reserved property name.

rwjblue:
Previously, having an `isStream` property on a component would cause us to
assume that component was itself a stream. Using an internal symbol allows
us to avoid reserving the property `isStream`.

Fixes #13271.

### DIFF
--- a/packages/ember-glimmer/tests/integration/components/curly-components-test.js
+++ b/packages/ember-glimmer/tests/integration/components/curly-components-test.js
@@ -917,4 +917,43 @@ moduleFor('Components test: curly components', class extends RenderingTest {
     this.assertElement(this.firstChild.childNodes[2], { tagName: 'b', content: 'bold' });
   }
 
+  ['@test can use isStream property without conflict (#13271)']() {
+    let component;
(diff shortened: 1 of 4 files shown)


rwjblue:
@mmun - r?

mmun:
You rule @rwjblue!

noslouch:
🎆

PR #13292, opened 2016-04-10
enhance helper documentation

BryanCrotaz:
Fix issue #13291

### DIFF
--- a/packages/ember-htmlbars/lib/helper.js
+++ b/packages/ember-htmlbars/lib/helper.js
@@ -41,8 +41,8 @@ import Object from 'ember-runtime/system/object';
 
   Additionally, class helpers can call `recompute` to force a new computation.
 
-  If you only require a simple functional style helper (where the output is only
-  dependent on the current input) then you can use the `Helper.helper` function.
+  If the output of your helper is only dependent on the current input, then you
+  can use the `Helper.helper` function.
 
   In this form the example above becomes:
   
@@ -54,9 +54,9 @@ import Object from 'ember-runtime/system/object';
   });
   ```
 
-  This latter form is very slightly more efficient at run time and uses fewer characters
-  in the compiled js. It is also easier to test by using the following structure
-  and importing the `formatCurrency` function into a test:
+  This latter form is more efficient at run time and results in smaller compiled js.
+  It is also easier to test by using the following structure and importing the 
+  `formatCurrency` function into a test:
 
   ```js
   export function formatCurrency (params, hash) {
@@ -66,6 +66,10 @@ import Object from 'ember-runtime/system/object';
   }
 
   export default Ember.Helper.helper(formatCurrency);
+
+  // tests/myhelper.js
+  import {formatCurrency} from ..../helpers/myhelper
+  // add some tests
   ```
 
   @class Ember.Helper
--- a/packages/ember-htmlbars/lib/helper.js
+++ b/packages/ember-htmlbars/lib/helper.js
@@ -41,6 +41,33 @@ import Object from 'ember-runtime/system/object';
 
   Additionally, class helpers can call `recompute` to force a new computation.
 
+  If you only require a simple functional style helper (where the output is only
+  dependent on the current input) then you can use the `Helper.helper` function.
+
+  In this form the example above becomes:
+  
+  ```js
+  export default Ember.Helper.helper((params, hash) => {
+    let cents = params[0];
+    let currency = hash.currency;
+    return `${currency}${cents * 0.01}`;
+  });
+  ```
+
+  This latter form is very slightly more efficient at run time and uses fewer characters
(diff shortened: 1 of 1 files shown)


mixonic:
@BryanCrotaz can you update the `Ember.Helper.helper` does instead? There are indeed two forms of helper, but the `Ember.Helper` docs should likely be for the class style and `Ember.Helper.helper` docs for the function style.

BryanCrotaz:
The reason I've put it there is it explains how to use the Helper class.
Most users on their first outing into helpers will read the example and
assume this is how it is supposed to be used. However they will get
confused (as I did) when the ember-cli generator creates a completely
difference usage.

On 10 April 2016 at 18:58, Matthew Beale <notifications@github.com> wrote:

> @BryanCrotaz <https://github.com/BryanCrotaz> can you update the
> Ember.Helper.helper does instead? There are indeed two forms of helper,
> but the Ember.Helper docs should likely be for the class style and
> Ember.Helper.helper docs for the function style.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/emberjs/ember.js/pull/13292#issuecomment-208032514>
>



-- 
Bryan Crotaz
Managing Director
Silver Curve

mixonic:
You could perhaps link from the class form docs to the function form docs and the inverse?

packages/ember-htmlbars/lib/helper.js
```diff
@@ -41,6 +41,33 @@ import Object from 'ember-runtime/system/object';
...
+  ```js
+  export default Ember.Helper.helper((params, hash) => {
+    let cents = params[0];
+    let currency = hash.currency;
+    return `${currency}${cents * 0.01}`;
+  });
+  ```
+
+  This latter form is very slightly more efficient at run time and uses fewer characters
+  in the compiled js. It is also easier to test by using the following structure
```
pixelhandler: Good point on ability to test the plain function :+1:

packages/ember-htmlbars/lib/helper.js
```diff
@@ -41,6 +41,33 @@ import Object from 'ember-runtime/system/object';
...
+  
+  ```js
+  export default Ember.Helper.helper((params, hash) => {
+    let cents = params[0];
+    let currency = hash.currency;
+    return `${currency}${cents * 0.01}`;
+  });
+  ```
+
+  This latter form is very slightly more efficient at run time and uses fewer characters
```
pixelhandler: @BryanCrotaz I'm curious if you can revisit this sentance. I get the point that it can be more efficient at run time. I'm wondering if the phrase `is very slightly more efficient` can be shorter. Maybe just a brief statement like "This latter form can be more efficient at run time, and exporting the function makes it easy to test."

rwjblue: I'd prefer to not mention the performance aspect at all, and just rely on the fact that it is shorter/less boilerplatey...

packages/ember-htmlbars/lib/helper.js
```diff
@@ -41,6 +41,33 @@ import Object from 'ember-runtime/system/object';
 
   Additionally, class helpers can call `recompute` to force a new computation.
 
+  If you only require a simple functional style helper (where the output is only
```
locks: This could be shortened to "If you require a helper where the output is only dependent on the current input, then you can use the `Helper.helper` function".

locks:
When @BryanCrotaz addresses the comments left inline, I am ok with merging this from a documentation standpoint.

mixonic:
Again, I'd really like to just see a cross-link here. `helper` is a function with its own documentation. Documenting it on the class without at *minimum* a cross link seems misleading. IMO documenting it at all on the class is unfortunate.

locks:
@mixonic how does the cross-linking work?

BryanCrotaz:
What's the format for a cross link?

mixonic:
Just linking to another section of the documentation, ala https://github.com/emberjs/ember.js/blob/57cf48cfce2acd895572f6f4c0b9478490259cdc/packages/ember-htmlbars/lib/index.js#L68

packages/ember-htmlbars/lib/helper.js
```diff
@@ -41,6 +41,37 @@ import Object from 'ember-runtime/system/object';
 
   Additionally, class helpers can call `recompute` to force a new computation.
 
+  If the output of your helper is only dependent on the current input, then you
+  can use the `Helper.helper` function.
```
pixelhandler: @BryanCrotaz Perhaps this is where we need the crosslink that @mixonic mentioned

> See [Ember.Helper.helper](/api/classes/Ember.Helper.html#method_helper)

rwjblue:
Closed by a slightly reworded version in #13291, thank you again for bringing it up and working with us!

PR #11225, opened 2015-05-19
[BUGFIX beta] Failing test for two-way binding through CP

mixonic:
Two way binding when a CP is present on the inner view does not propagate changes to the outer view.

PR #13293, opened 2016-04-10
A two way binding flows upstream through a CP.

rwjblue:
This is a rewritten version of #11225 to work with the new testing structure. 

Closes #11225.

### DIFF
--- a/packages/ember-glimmer/tests/integration/components/curly-components-test.js
+++ b/packages/ember-glimmer/tests/integration/components/curly-components-test.js
@@ -7,6 +7,7 @@ import { strip } from '../../utils/abstract-test-case';
 import { moduleFor, RenderingTest } from '../../utils/test-case';
 import { classes } from '../../utils/test-helpers';
 import { htmlSafe } from 'ember-htmlbars/utils/string';
+import { computed } from 'ember-metal/computed';
 
 moduleFor('Components test: curly components', class extends RenderingTest {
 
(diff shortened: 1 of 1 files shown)


rwjblue:
I'd like @mixonic to confirm that this properly replaces https://github.com/emberjs/ember.js/pull/11225.

packages/ember-glimmer/tests/integration/components/curly-components-test.js
```diff
@@ -956,4 +957,49 @@ moduleFor('Components test: curly components', class extends RenderingTest {
...
+          this._bar = value;
+          return this._bar;
+        }
+      })
+    });
+
+    this.registerComponent('foo-bar', {
+      ComponentClass: FooBarComponent,
+
+      template: '{{bar}}'
```
mixonic: That the CP is consumed by the template is different than in https://github.com/emberjs/ember.js/pull/11225

rwjblue: Gotcha, I will add another test without that consumption.

rwjblue: Added another test for that (see below).

rwjblue:
Added a second test that must be skipped (along with a small comment explaining when the regression was introduced). This test should be made to pass prior to enabling `ember-glimmer` (it should have been fixed in HTMLBars in 1.13, but that :ship: has sailed).

PR #13294, opened 2016-04-11
[DEPRECATE release] Deprecate usage of `{{render}}` with a block.

rwjblue:
This was not intended to be supported and is rarely used. Adding support for this in tildeio/glimmer will be pretty hard (due to very bizarre `controller` and target values for the block).

This deprecation is intended to help guide anyone actually using this functionality before upgrading to 2.5.0 or later.


rwjblue:
Added deprecation guide entry in https://github.com/emberjs/website/pull/2540.

PR #12575, opened 2015-11-09
Make acceptance helpers fire native events instead of jQuery ones.

cibernox:
Following issue #12569

Goal: To fire events in relatively accurate and cross-browser way.

Simulate real events in pure javascript accurately is nearly impossible, but
this is a best effort. This fires 3 different kind of events:
- MouseEvents: click, dbclick, mousedown, mouseup, mouseenter, mouseleave, ...
- KeyEvents: keydown, keypress and keyup
- Events: Anything other than mouse or keyboard events falls in this generic kind of event.

This should probably cover 99% of use cases.

Caveats found:

I've only encountered 2 failing tests after this refactor, both related to the same problem:

```
triggerEvent('.input', 'keydown', { keyCode: 13 });
// ...
equal(event.keyCode, 13, 'options were passed');
```

It turns that `keyCode` is a reserved property that only works with keyboardEvents, so although the event has this property mixed, the wrapped jquery event does not copy that property if the original event is not a keyboard event. 
Seems very reasonable and probably that test was a bad example.

PR #13310, opened 2016-04-12
[BUGFIX release] Fix mouseenter typo in ember-testing helpers

jakesjews:
It looks like there was a typo in https://github.com/emberjs/ember.js/pull/12575 `mousenter` should be `mouseenter`


rwjblue:
Great catch!  This was just released in v2.5.0 (cut earlier today), and we are in the process of rolling canary -> beta and beta -> release.  Can you update commit prefix to be `[BUGFIX release]` so we can backport into a future v2.5.1?

jakesjews:
No problem!
It looks like there might be another issue I have to fix though. When I run the test in the browser or with ember-cli it passes but it looks like its failing with `npm test`. Does anything jump out that I'm doing wrong?

PR #13311, opened 2016-04-12
[BUGFIX beta] Allow numeric keys for the `get` keyword

### LINKED ISSUE #13296: `get` keyword does not allow numeric keys to be used
Due to [this line](https://github.com/emberjs/ember.js/blob/6b78bcb17f575df1c61fe63d896e28a4463db3c4/packages/ember-htmlbars/lib/keywords/get.js#L38), using `get` with a numeric key does not work properly.

``` js
  model() {
    return {
      indexes: [2,3,4],
      items: {
        2: 'blah',
        3: 'zorz',
        4: 'derp'        
      }
    }
  },
```

``` hbs
    <ul>
      {{#each model.indexes as |index|}}
        <li>{{get model.items index}}</li>
      {{/each}}
    </ul>
```

---

Demo: http://rwjblue.jsbin.com/loneva/edit?html,js,output

duggiefresh:
Fixes #13296

### DIFF
--- a/packages/ember-glimmer/lib/utils/references.js
+++ b/packages/ember-glimmer/lib/utils/references.js
@@ -127,6 +127,11 @@ export class PropertyReference extends CachedReference { // jshint ignore:line
   }
 
   get(propertyKey) {
+    if (typeof propertyKey === 'number') {
+      // Allows us to get numeric keys
+      propertyKey = '' + propertyKey;
+    }
+
     return new PropertyReference(this, propertyKey);
   }
 }
--- a/packages/ember-htmlbars/lib/keywords/get.js
+++ b/packages/ember-htmlbars/lib/keywords/get.js
@@ -37,6 +37,8 @@ const DynamicKeyStream = BasicStream.extend({
     let key = this.keyDep.getValue();
     if (typeof key === 'string') {
       return key;
+    } else if (typeof key === 'number') {
+      return '' + key;
     }
   },
 
--- a/packages/ember-glimmer/tests/integration/helpers/get-test.js
+++ b/packages/ember-glimmer/tests/integration/helpers/get-test.js
@@ -51,6 +51,31 @@ moduleFor('Helpers test: {{get}}', class extends RenderingTest {
     this.assertText('[red and yellow] [red and yellow]');
   }
 
+  ['@test should be able to get an object value with numeric keys']() {
+    this.render(`{{#each indexes as |index|}}[{{get items index}}]{{/each}}`, {
+      indexes: [1, 2, 3],
+      items: {
+        1: 'First',
+        2: 'Second',
+        3: 'Third'
+      }
+    });
+
+    this.assertText('[First][Second][Third]');
+
+    this.runTask(() => this.rerender());
+
+    this.assertText('[First][Second][Third]');
+
+    this.runTask(() => set(this.context, 'items.1', 'Qux'));
+
+    this.assertText('[Qux][Second][Third]');
+
+    this.runTask(() => set(this.context, 'items', { 1: 'First', 2: 'Second', 3: 'Third' }));
+
+    this.assertText('[First][Second][Third]');
+  }
+
   ['@test should be able to get an object value with a bound/dynamic key']() {
     this.render(`[{{get colors key}}] [{{if true (get colors key)}}]`, {
       colors: { apple: 'red', banana: 'yellow' },


packages/ember-htmlbars/lib/keywords/get.js
```diff
@@ -35,8 +35,8 @@ let DynamicKeyStream = BasicStream.extend({
 
   key() {
     let key = this.keyDep.getValue();
-    if (typeof key === 'string') {
-      return key;
+    if (typeof key === 'string' || typeof key === 'number') {
```
mmun: We should align with how the ES2015 spec treats `foo[bar]`. So basically

```js
return typeof key === 'symbol' ? key : String(key);
```

rwjblue: Awesome, thank you @mmun!

duggiefresh:
Updated per review. I made an update to [Glimmer util/references](https://github.com/emberjs/ember.js/pull/13311/files#diff-68fe34e1d224b37871364fe5d84cc481R88) that will "stringify" numeric keys.

Thanks!

packages/ember-htmlbars/lib/keywords/get.js
```diff
@@ -35,8 +35,10 @@ let DynamicKeyStream = BasicStream.extend({
 
   key() {
     let key = this.keyDep.getValue();
-    if (typeof key === 'string') {
+    if (typeof key === 'symbol') {
```
chancancode: Can you add a test for this? (If the browser supports it?)

chancancode:
```javascript
if (supportsNativeSymbols) {
  moduleFor('Helpers test: {{get}} with symbols', class extends RenderingTest {
    // ...
  });
}
```

duggiefresh: Can do! Is there an example somewhere of how to check for browser support within tests? Thanks

chancancode: I think something like what I mentioned above should work:

```javascript
moduleFor('Helpers test: {{get}}', class extends RenderingTest {
  // current tests
}

if (typeof Symbol === 'function') { // I think?
  moduleFor('Helpers test: {{get}} with symbols', class extends RenderingTest {
    // moar tests
  });
}
```

duggiefresh: Thanks for the help. However, it seems there isn't support for symbols in templates. Do we want to add symbol support for the `{{get}}` keyword?

For example, something like this doesn't work either:
```js
['@test should be able to get an object value with numeric keys']() {
  let first = Symbol('first');
  let second = Symbol('second');
  let third = Symbol('third');

  this.render(`{{#each indexes as |index|}}[{{get items index}}]{{/each}}`, {
    indexes: [first, second, third],
    items: {
      [first]: 'First',
      [second]: 'Second',
      [third]: 'Third'
    }
  });

°°°
```

duggiefresh:
Sorry to bother, but just a ping to move this along.

Please see: https://github.com/emberjs/ember.js/pull/13311#discussion_r59550341

Thanks, @chancancode @rwjblue

packages/ember-htmlbars/lib/keywords/get.js
rwjblue: No, if we don't support symbols already in template-land, then we don't need to support it explicitly here.

duggiefresh: Great! Thanks will update the PR as necessary

krisselden:
I don't understand why we don't check for number and coerce it?  Can we start conservatively?  Does it make sense to toString anything?  This will toString `null` to `'null'` for example.

krisselden:
Especially if this is targeted as a bugfix to beta.

mixonic:
@duggiefresh I think there was interest in removing the `symbol` specific code here? /cc @rwjblue 

And @krisselden's suggestion re: swapping to a conservative approach, only coercing numbers, seems reasonable.

Would love to see this land :-)

duggiefresh:
@mixonic Thanks for pinging me on this. I believe I'll be attending your Contributors Workshop at WGE. Perhaps, we can land this before the conference. :)

See you then

duggiefresh:
Updated per comments! I removed the `symbol` references and am now coercing `number` keys. Thanks :ice_cream:

duggiefresh:
ping @mixonic @krisselden Thank you!

Turbo87:
does this work for `0` as a key too? I know I can use `firstObject` too, but in a index-loop it would be easier if `0` would work too.

mixonic:
@Turbo87 I'm not quite sure what you're asking. Can you provide an example?

packages/ember-glimmer/lib/utils/references.js
```diff
@@ -127,6 +127,11 @@ export class PropertyReference extends CachedReference { // jshint ignore:line
   }
 
   get(propertyKey) {
+    if (typeof propertyKey === 'number') {
+      // Allows us to get numeric keys
+      propertyKey = '' + propertyKey;
```
mixonic: I kind of expect you would only need the logic in the keyword. What other use-cases does this cover?

Turbo87:
@mixonic I'm talking about something like 

```hbs
{{#each someArray as |value index|}}
  value: {{get some.other.array index}}
{{/each}}
```

packages/ember-htmlbars/lib/keywords/get.js
```diff
@@ -37,6 +37,8 @@ const DynamicKeyStream = BasicStream.extend({
     let key = this.keyDep.getValue();
     if (typeof key === 'string') {
       return key;
+    } else if (typeof key === 'number') {
+      return '' + key;
```
krisselden: can you explain why it is needed to be check in both places?

packages/ember-glimmer/lib/utils/references.js
duggiefresh: I added the coercion here too, because of the assertion here: https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/lib/property_get.js#L49

packages/ember-htmlbars/lib/keywords/get.js
duggiefresh: I added a coercion in both `ember-glimmer` and `ember-htmlbars` because of the assertion here: https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/lib/property_get.js#L49

Perhaps there's a better place to add this logic?

locks:
@duggiefresh are you able to rebase and solve the conflicts? Thank you!

tsteuwer:
Pease let this happen. I have to manually change keys to strings to use get with them 😊

rwjblue:
I rebased and updated to work on master, but this PR predates github's feature that allows project maintainers to push to PR branches. I resubmitted (with correct commit attribution) in #15366.

PR #13312, opened 2016-04-12
[BUGFIX canary] Fix default args for calculateLocationDisplay.

dgeb:
Ensure that undefined or null args are allowed for 
`calculateLocationDisplay`.

Fixes an error transforming an inline `{{link-to}}` without legacy
view support.


rwjblue:
:+1: - LGTM (waiting for Travis to catch up on its queue)

rwjblue:
I think it will end up being `[BUGFIX beta]` though, since we just rolled what was canary into beta (for 2.6.0-beta.1).

PR #13314, opened 2016-04-12
[CLEANUP beta] Remove Metamorph view and mixin.

rwjblue:


githubSathishkumar:
@rwjblue  can you please add a discussion thread ID here to know why this cleanup

rwjblue:
It had been deprecated for quite a long time.  The deprecation made it clear that it would only be around until 2.4.0. I noticed that it had still not been removed, so I removed it...

PR #13316, opened 2016-04-12
[DO NOT MERGE] Glimmer 2 Final Countdown

chadhietala:
### Update

Markdown file did not pan out. There was much confusion on what tasks were taken. So what is the plan? I have migrated all of the tasks and relevant information to a [Trello board](https://trello.com/b/6C6i3eGV/glimmer-2) 😱 . No worries you do not need to signup for Trello or any of that nonsense. Please continue 🔒 ing things by just commenting on this PR along with the link to the trello board. **Example:**

> Hey everyone I'm going to take https://trello.com/c/pR3KtDCH/20-ensure-falsy-values-do-not-become-class-names 🔒 .

I will then personally go move the card to "In Progress". Hopefully this is not too much burden and will actually result in clear direction of "what needs to be done".

Thanks everyone!

**Board**
https://trello.com/b/6C6i3eGV/glimmer-2

---

This is the 🔥 🔥 🔥 burn down 🔥 🔥 🔥  issue to ship Glimmer 2. While there are still in flight PRs, this is known state of the world at SHA b418b95ea3d31e3a3656906e6ad784422f648c31.

Awesome 80s hair band

We are experimenting on using a temporary Markdown instead of a regular checklist so that we can add line comments for each item. _Do not merge this pull request._

PR #13317, opened 2016-04-12
[Glimmer 2] Allow for null lookups on `get` to fail gracefully

chadhietala:
Fixes 3 failing tests under `get` helper in #13316

### DIFF
--- a/packages/ember-glimmer/lib/utils/references.js
+++ b/packages/ember-glimmer/lib/utils/references.js
@@ -119,7 +119,7 @@ export class GetHelperReference extends CachedReference {
 
   isDirty() { return true; }
 
-  value() {
+  compute() {
     let key = this.pathReference.value();
     let keyType = typeof key;
 
--- a/packages/ember-glimmer/lib/utils/references.js
+++ b/packages/ember-glimmer/lib/utils/references.js
@@ -120,7 +120,14 @@ export class GetHelperReference extends CachedReference {
   isDirty() { return true; }
 
   value() {
-    return this.sourceReference.get(this.pathReference.value()).value();
+    let key = this.pathReference.value();
+    let keyType = typeof key;
+
+    if (key && (keyType === 'string' || keyType === 'number')) {
+      return this.sourceReference.get(key).value();
+    }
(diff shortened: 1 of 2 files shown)


packages/ember-glimmer/lib/utils/references.js
```diff
@@ -78,7 +78,7 @@ class PropertyReference extends CachedReference { // jshint ignore:line
 
     _parentObjectTag.update(tagFor(parentValue));
 
-    if (parentValue && typeof parentValue === 'object') {
```
rwjblue: See https://github.com/emberjs/ember.js/issues/13296 and https://github.com/emberjs/ember.js/pull/13311.  We should support `{{get foo 1}}` (iow, not just string keys).

packages/ember-glimmer/lib/utils/references.js
```diff
@@ -78,7 +78,7 @@ class PropertyReference extends CachedReference { // jshint ignore:line
...
+    if (parentValue && typeof parentValue === 'object' && typeof _propertyKey === 'string') {
```
rwjblue: See https://github.com/emberjs/ember.js/issues/13296 and https://github.com/emberjs/ember.js/pull/13311.  We should support `{{get foo 1}}` (iow, not just string keys).

packages/ember-glimmer/lib/utils/references.js
```diff
@@ -120,7 +120,14 @@ export class GetHelperReference extends CachedReference {
...
   value() {
-    return this.sourceReference.get(this.pathReference.value()).value();
+    let key = this.pathReference.value();
+    let keyType = typeof key;
+
+    if (key && (keyType === 'string' || keyType === 'number')) {
+      return this.sourceReference.get(key).value();
+    }
+
+    return NULL_REFERENCE.value();
```
chancancode: I didn't realize we are at the "leaf" of a chain here. In that case, we can just return `null` here. `NULL_REFERENCE` is just a POJO whose `value` and `get` is hardcoded to always return `null`.

PR #13320, opened 2016-04-13
[Glimmer2] Update htmlbars only tests in curly components

Serabe:
There are three tests now passing in Glimmer2 too.


rwjblue:
Thanks @Serabe!

PR #13328, opened 2016-04-13
[FEATURE ember-improved-instrumentation] Add link-to instrumentation

asakusuma:
Adding basic instrumenting to link-to.

### DIFF
--- a/packages/ember/tests/helpers/link_to_test.js
+++ b/packages/ember/tests/helpers/link_to_test.js
@@ -4,6 +4,7 @@ import Controller from 'ember-runtime/controllers/controller';
 import { set } from 'ember-metal/property_set';
 import Route from 'ember-routing/system/route';
 import run from 'ember-metal/run_loop';
+import { subscribe, reset } from 'ember-metal/instrumentation';
 import isEnabled from 'ember-metal/features';
 import alias from 'ember-metal/alias';
 import Application from 'ember-application/system/application';
@@ -76,6 +77,7 @@ function sharedSetup() {
 function sharedTeardown() {
   run(function() { App.destroy(); });
   setTemplates({});
+  reset();
 }
 
 import { test } from 'ember-glimmer/tests/utils/skip-if-glimmer';
@@ -166,6 +168,78 @@ QUnit.test('The {{link-to}} helper moves into the named route', function() {
   equal(jQuery('#self-link.active', '#qunit-fixture').length, 1, 'The self-link was rendered with active class');
   equal(jQuery('#home-link:not(.active)', '#qunit-fixture').length, 1, 'The other link was rendered without active class');
 });
+
+if (isEnabled('ember-improved-instrumentation')) {
+  QUnit.test('The {{link-to}} helper fires an interaction event', function(assert) {
+    assert.expect(2);
+    Router.map(function(match) {
+      this.route('about');
+    });
+
+    bootApplication();
+
+    run(function() {
+      router.handleURL('/');
+    });
+
+    subscribe('interaction.link-to', {
+      before() {
+        assert.ok(true, 'instrumentation subscriber was called');
+      },
+      after() {
+        assert.ok(true, 'instrumentation subscriber was called');
+      }
+    });
+
+    jQuery('#about-link', '#qunit-fixture').click();
+  });
+
+  QUnit.test('The {{link-to}} helper interaction event includes the route name', function(assert) {
+    assert.expect(2);
+    Router.map(function(match) {
+      this.route('about');
+    });
+
+    bootApplication();
+
+    run(function() {
+      router.handleURL('/');
+    });
+
+    subscribe('interaction.link-to', {
+      before(name, timestamp, { routeName }) {
+        assert.equal(routeName, 'about', 'instrumentation subscriber was passed route name');
+      },
+      after(name, timestamp, { routeName }) {
+        assert.equal(routeName, 'about', 'instrumentation subscriber was passed route name');
+      }
+    });
+
+    jQuery('#about-link', '#qunit-fixture').click();
+  });
+
+  QUnit.test('The {{link-to}} helper interaction event includes the transition in the after hook', function(assert) {
+    assert.expect(1);
+    Router.map(function(match) {
+      this.route('about');
+    });
+
+    bootApplication();
+
+    run(function() {
+      router.handleURL('/');
+    });
+
+    subscribe('interaction.link-to', {
+      before() {},
+      after(name, timestamp, { transition }) {
+        assert.equal(transition.targetName, 'about', 'instrumentation subscriber was passed route name');
+      }
+    });
+
+    jQuery('#about-link', '#qunit-fixture').click();
+  });
+}
 
 QUnit.test('The {{link-to}} helper supports URL replacement', function() {
   setTemplate('index', compile(`<h3>Home</h3>{{#link-to 'about' id='about-link' replace=true}}About{{/link-to}}`));
@@ -1757,4 +1831,3 @@ QUnit.test('GJ: {{link-to}} to a parent root model hook which performs a `transi
 
   shouldBeActive('#parent-link');
 });
-
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -56,3 +56,4 @@ for a detailed explanation.
 
   - `interaction.<event-name>` for events handled by a component.
   - `interaction.ember-action` for closure actions.
+  - `interaction.link-to` for link-to execution.
--- a/packages/ember-htmlbars/lib/components/link-to.js
+++ b/packages/ember-htmlbars/lib/components/link-to.js
@@ -323,6 +323,8 @@ import 'ember-runtime/system/service'; // creates inject.service
 import ControllerMixin from 'ember-runtime/mixins/controller';
 import layout from '../templates/link-to';
 import EmberComponent, { HAS_BLOCK } from '../component';
+import { flaggedInstrument } from 'ember-metal/instrumentation';
+
 
 /**
   `Ember.LinkComponent` renders an element whose `click` event triggers a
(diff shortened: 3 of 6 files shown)


rwjblue:
Restarted sauce labs job.

asakusuma:
For some reason [this test](https://github.com/emberjs/ember.js/blob/57cf48cfce2acd895572f6f4c0b9478490259cdc/packages/ember-htmlbars/tests/helpers/closure_component_test.js#L708) is failing in Microsoft Edge. Not sure why, the test does not seem to use link-to...

asakusuma:
Here is the error:
RangeError: Array length must be a finite positive integer

Video: https://saucelabs.com/jobs/7266308c4dc94740acc91265e645c45f

I'm going to just manually step through the test to try to figure out why this is happening. May also try to get a hold of a windows machine.

Any history with Microsoft Edge causing problems for ember test runs?

asakusuma:
@rwjblue I'm trying to figure out what actual lines of code map to each entry in the ember.debug.js stack trace in the saucelab video. However, the stack has changed between saucelab runs. I would assume this means that either the failures are not deterministic or that saucelab rebases before every run.

Neither of the stack traces line up with what I'm seeing in my ember.debug.js locallly. I have not rebased since pushing to the PR branch. Any tips on debugging a saucelab output?

stacktrace

asakusuma:
Some weird stuff is going on in MS Edge. I tried running the tests on a windows machine with Edge and could not reproduce the issue. Eventually I figured out that that backburner.run was somehow getting passed no arguments, causing [this line](https://github.com/ebryn/backburner.js/blob/master/lib/backburner.js#L160) to throw the `RangeError: Array length must be a finite positive integer`. The source of the errors is [run.join(target, action, ...args)](https://github.com/asakusuma/ember.js/blob/master/packages/ember-routing-htmlbars/lib/keywords/closure-action.js#L90).

When I change the syntax for [run.join(target, action, ...args)](https://github.com/asakusuma/ember.js/blob/master/packages/ember-routing-htmlbars/lib/keywords/closure-action.js#L90) to [something that should be equivalent](https://github.com/asakusuma/ember.js/blob/link-to-int/packages/ember-routing-htmlbars/lib/keywords/closure-action.js#L90-L94), the tests pass. Feels like a browser bug to me. I don't see anyway in which [run.join(target, action, ...args)](https://github.com/asakusuma/ember.js/blob/master/packages/ember-routing-htmlbars/lib/keywords/closure-action.js#L90) can result in an array of length zero.

packages/ember/tests/helpers/link_to_test.js
```diff
@@ -168,6 +170,83 @@ QUnit.test('The {{link-to}} helper moves into the named route', function() {
...
+      router.handleURL('/');
+    });
+
+    subscribe('interaction.link-to', {
+      after(name, timestamp, { transition }) {
+        assert.equal(transition.targetName, 'about', 'instrumentation subscriber was passed route name');
+      }
+    });
+
+    run(function() {
```
mixonic: Should not be a runloop here. The runloop is inside the `click` handler.

packages/ember-metal/lib/instrumentation.js
```diff
@@ -155,7 +155,7 @@ export function _instrumentStart(name, _payload) {
   var timestamp = time();
   for (i = 0; i < listeners.length; i++) {
     listener = listeners[i];
-    beforeValues[i] = listener.before(name, timestamp, payload);
+    beforeValues[i] = typeof listener.before === 'function' ? listener.before(name, timestamp, payload) : null;
```
mixonic: is the value every anything but a function? Can this simply check truthiness?

Making `before` optional would call for a documentation update as well.

packages/ember/tests/helpers/link_to_test.js
mixonic: Also a few places above.

krisselden:
Asa, can you rebase?

asakusuma:
Did some more digging today and I believe this is a bug in MS Edge. Looks like saucelabs is on an unsupported version of MS Edge. I've tweeted at them to see if they've got some plans to upgrade.

asakusuma:
@rwjblue any insight on how we can get Saucelabs to upgrade their version of Microsoft Edge? Version 13 is pretty out of date and no longer supported.

https://en.wikipedia.org/wiki/Microsoft_Edge#Release_history

rwjblue:
They are using 13.10586	 of `EdgeHTML`, which correlates to version 25 in that chart.

asakusuma:
It says in the [saucelabs metadata](https://saucelabs.com/jobs/35a3991395be4d20bdf04e892e077521) that 13.10586 is the "browser version." How do we know that the version is referring to `EdgeHTML` and not the generic browser version?

rwjblue:
I suppose we don't know for 100% certainty, but the version number matching exactly what is used by Edge 25 (and the only time that number exists on the wiki page you referenced) would be a heck of a coincidence.

screenshot

In addition, a forum comment by Sauce Labs support staff mentioned that Edge 25 was available as of March of this year: https://support.saucelabs.com/customer/portal/questions/16186272-edge-25-availability.

asakusuma:
You're right @rwjblue https://twitter.com/saucelabs/status/727232878380208128

My windows machine is on Edge 20, I'll upgrade and see if I can reproduce this.

asakusuma:
I believe this is an Edge bug, but would love for someone else to take a look and confirm or deny my suspicions.

Somehow, [run.join(target, action, ...args)](https://github.com/emberjs/ember.js/blob/74937c7631c5166cdfbd3c0a34bc2101fa47e4a0/packages/ember-routing-htmlbars/lib/keywords/closure-action.js#L95) is creating a function call that has no arguments, which I believe should be impossible. That is causing [this line](https://github.com/ebryn/backburner.js/blob/master/lib/backburner.js#L237) in backburner to throw the `RangeError: Array length must be a finite positive integer` error.

I was able to reproduce this on a local machine. It's not reproducible when:

- Test is run alone
- All tests are run, but with try catch turned off
- Try catch is added around the [offending line of code](https://github.com/emberjs/ember.js/blob/74937c7631c5166cdfbd3c0a34bc2101fa47e4a0/packages/ember-routing-htmlbars/lib/keywords/closure-action.js#L95)
- Conditional breakpoint is set on offending line of code
- Guard is added [here](https://github.com/emberjs/ember.js/blob/74937c7631c5166cdfbd3c0a34bc2101fa47e4a0/packages/ember-metal/lib/run_loop.js#L121) to throw an error if `arguments` length is zero

Here is the error strack, along with the compiled JS code which map to the line numbers in the error stack:

error stack trace
join
closure action

cc @krisselden @chadhietala @nathanhammond @pete-the-pete

nathanhammond:
@asakusuma It sounds to me like maybe Edge/Chakra is optimizing an object out of existence while setting up that closure.

Things that might make that more clear:
- `try { args; } catch(e) {}` ... might prevent optimization of the function.
- `var temp = args[0];` ... adding a simple read beforehand might realize that it needs to be in scope.

That's all I've got.

krisselden:
Have you tried rebasing? I thought I merged a PR that remove arguments.length caching

asakusuma:
Thanks @nathanhammond, I'll give the read hint a shot.

@krisselden I'll try that too.

This PR, which migrates the failing test, might also help: https://github.com/emberjs/ember.js/pull/13491

asakusuma:
So adding [a guard with a throw](https://github.com/emberjs/ember.js/pull/13328/commits/8e8f576f9ced6b94c01b1438938b7b92d9b8116b) fixes the issue. As @nathanhammond said, perhaps this removes a bad optimization in Edge/Chakra. How should I proceed @krisselden @rwjblue?

asakusuma:
Essentially I'm asking if it's acceptable to add a guard that should not be needed but is necessary to work around what appears to be a browser bug. Also wondering if it should be an ember assert instead. I'll also need to add some sort of comment explaining the issue.

nathanhammond:
@asakusuma `Ember.assert` would be stripped in a production build, so I do not recommend. 😛 I believe we still have a few guards in places where optimizations in the underlying JS engines are broken, so this wouldn't be the first.

Who do we know on the Chakra team?

/cc @stefanpenner

packages/ember-routing-htmlbars/lib/keywords/closure-action.js
```diff
@@ -92,7 +92,11 @@ function createClosureAction(stream, target, action, valuePath, actionArguments)
 
       let payload = { target, args, label: labelFor(stream) };
       return flaggedInstrument('interaction.ember-action', payload, () => {
-        return run.join(target, action, ...args);
+        let params = [target, action].concat(args);
```
stefanpenner: shouldn't we error before line 94 ? That way it fails fast

stefanpenner:
@asakusuma / @nathanhammond can one of your outline the issue in edge concisely? I'll make sure we get someone from other there to take a look, but I myself am having trouble following.

asakusuma:
@stefanpenner I've described the complete issue in [this comment](https://github.com/emberjs/ember.js/pull/13328#issuecomment-218809667), but an even simpler description would be:

Function `foo` gets called via `apply` with the following arguments array: `[a, b].concat(args)`, where `a` and `b` are any variables, and `args` is an array of any length. Inside `foo`, `arguments.length` is `0`. This should not be possible since the arguments array is hardcoded to be at least length `2`.

asakusuma:
And to connect my general description to the actual issue at hand, `foo` in my example is run loop `join` in this issue.

packages/ember-routing-htmlbars/lib/keywords/closure-action.js
asakusuma: In theory yes, not sure if that will still pacify Chakra, but I'll give it a shot.

stefanpenner:
@asakusuma so the arity changes on edge?

stefanpenner:
cc @jdalton / @bterlson seems like we may be running into an issue with IE Edge.  Specifically outlined [here](https://github.com/emberjs/ember.js/pull/13328#issuecomment-219551154)

asakusuma:
@stefanpenner it would appear so. Not sure if the actual arity changes or if the representation of arity changes. In other words, `arguments.length` is zero, but I'm not sure if arguments[0] would actually have the intended value.

Issue is not reproducible when using a debugger.

asakusuma:
@stefanpenner as I suspected, if the guard is moved, it doesn't work [see the commit pass/fail status](https://github.com/emberjs/ember.js/pull/13328/commits)

stefanpenner:
> @stefanpenner as I suspected, if the guard is moved, it doesn't work. See the commit pass/fail status.

thanks for trying

asakusuma:
@stefanpenner @nathanhammond @krisselden looks like the [closure action test migration](https://github.com/emberjs/ember.js/pull/13491) has "fixed" the issue. Tests are passing now without the guard hack.

jdalton:
> cc @jdalton / @bterlson seems like we may be running into an issue with IE Edge. Specifically outlined here any thoughts/ideas?

This sounds familiar. It would help if you all have a small repro. Have you tried on insider builds of Windows 10?

asakusuma:
I've had no luck re-creating in a small repro. Unfortunately, the issue has only ever surfaced when running the entire test suite, not just the single test.

I have not tried on an insider build. I'll see if I can get IT to arrange that. I've created [a branch](https://github.com/asakusuma/ember.js/tree/edge-repro) that can reproduce the issue.

Are there any current known issues that might inform a method for reproducing this issue?

asakusuma:
@krisselden @stefanpenner what needs to happen to get this merged?

stefanpenner:
@asakusuma it looks good to me, but im not terribly familiar with this. @krisselden should likely sign-off.

chadhietala:
@asakusuma I think it should be fine now that the tests will actually run when the feature flag is on. 👍

PR #13329, opened 2016-04-13
[GLIMMER2] Make safe string tests more robust

zackthehuman:
This fixes up some of the tests related to `htmlSafe` and `{{{thing}}}`. The actual DOM output between HTMLBars and Glimmer2 is slightly different and these changes take those differences into account by making them not matter.


rwjblue:
LGTM

chadhietala:
@zackthehuman Can you please move https://github.com/emberjs/ember.js/blob/master/packages/ember-htmlbars/tests/compat/safe_string_test.js can you migrate this test and then nuke safe_string_test.js?

chadhietala:
Its up for ✂️  in #13127

PR #13332, opened 2016-04-14
Misc Glimmer 2 fixes

chancancode:
- Glimmer compiler needs wire-format and references
- [Const reference optimization](https://github.com/emberjs/ember.js/pull/13332) for dynamic component
  
  This optimizes `{{component "foo-bar"}}` into the same thing as `{{foo-bar}}` in the updating program. It also works for passed-in attrs like `{{component @stuff}}` where `@stuff` is static on the invocation side `{{x-outside stuff="foo-bar"}}`.
- Correctly dirty `Ember.set(foo, 'bar.baz.bat', ...)`
  
  Previously, doing a `Ember.set(foo, 'bar.baz.bat', ...)` incorrectly marks both the "foo", "bar" and "baz" objects as dirty. This fixes the problem and only mark "bat" as dirty, i.e. making this the same as doing `Ember.set(Ember.get(foo, 'bar.baz'), 'bat', ...)`.
- Disable AST transforms on Glimmer for now
  
  This is causing issues because we have not implemented the `mut` helper in `ember-glimmer` (among other things). We can selectively re-enable stuff we still need (and works) as we go.
- Make conditional helpers/references not volatile
  
  We previously marked them as `volatile` with a FIXME because we need to support proxies. This reduces the scope and only mark conditionals with proxies as volatile (we can still do better than that by making proxy objects have a special tag that combines its own `DirtyableTag` with the tag from its `content`, but it's pretty low priority at the moment).
  
  The refactor also added a new optimizations: by reusing the same code across the syntax (`{{#if}}`) and the helper (`{{if}}`) versions, the inline conditional helpers now also participate in the same const optimization, i.e. `{{if true foo bar}}` will be optimized into `{{foo}}` in the updating program, so as `{{if @stuff foo bar}}` if `@stuff` is static.
  
  TODO: fix tests that uses fake proxies, see comment in `references.js` (cc @chadhietala)

cc @krisselden

### DIFF
--- a/packages/ember-template-compiler/lib/index.js
+++ b/packages/ember-template-compiler/lib/index.js
@@ -1,4 +1,5 @@
 export { default as _Ember } from 'ember-metal'; // Is this still needed
+import isEnabled from 'ember-metal/features';
 import { ENV } from 'ember-environment';
 import precompile from 'ember-template-compiler/system/precompile';
 import compile from 'ember-template-compiler/system/compile';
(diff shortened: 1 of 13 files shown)


packages/ember-template-compiler/lib/index.js
```diff
@@ -23,19 +24,21 @@ import AssertNoEachIn from 'ember-template-compiler/plugins/assert-no-each-in';
...
-registerPlugin('ast', TransformTopLevelComponents);
-registerPlugin('ast', DeprecateRenderModel);
-registerPlugin('ast', PreventRenderBlock);
-registerPlugin('ast', AssertNoEachIn);
-registerPlugin('ast', TransformInlineLinkTo);
+if (!isEnabled('ember-glimmer')) {
+  registerPlugin('ast', TransformOldBindingSyntax);
+  registerPlugin('ast', TransformOldClassBindingSyntax);
+  registerPlugin('ast', TransformItemClass);
+  registerPlugin('ast', TransformClosureComponentAttrsIntoMut);
```
krisselden: why not just guard this one? some of the others are needed to pass tests

PR #13333, opened 2016-04-14
[BUGFIX RELEASE] Fix toString to not add the property on the class

kratiahuja:
When calling `toString()` method on an Ember Object, it attaches the `toString` as a class property for the first time. This does not seem right since this should be on the prototype of the class. We could memoize the value on the class meta data.

In v2.5.0 and above, when calling `.set()` on an Ember Object, you end up also attaching the `toString()` property on the class. This [PR](https://github.com/emberjs/ember.js/commit/3a2486eb911ad77d2ab1a4324c1aacef08c156da) calls `toString()` on the object during setting any property. Therefore, when you walk through the object properties you end up getting an additional property on the object.

Thanks to Kris & Chad for guiding on this!

cc: @krisselden @chadhietala

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -15,6 +15,7 @@
         "throws",
         "deepEqual",
         "ok",
+        "notOk",
         "strictEqual",
         "expect",
         "minispade",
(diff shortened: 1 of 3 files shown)


chadhietala:
This looks good to me, CI just needs to be bounced once #13334 lands. Currently master is  🔴 .

chadhietala:
@jheth Was your push intentional? Seems like it was a mistake.

krisselden:
@rwjblue just to explain this, toString() mutates the shape of object and in the 2.5.0 release an assertion called object toString() in Ember.set() and while in production, asserts are stripped this failed tests that were asserting keys() because Ember.set() was adding a toString() to this.

rwjblue:
I have no clue why we would want to write `toString` to the instance anyways (maybe some work around for another issue).  This seems good.

PR #13335, opened 2016-04-14
[BUGFIX release] Ensure injected property assertion checks `container`.

rwjblue:
During the Ember 2.3.0 cycle, we introduced the `getOwner` / `setOwner` concepts and deprecated using the injected `.container` property on objects. Unfortunately, if you are using an older library to create objects that hasn't been updated to use `setOwner` you would only have the `container` specified which causes this injection assertion to be thrown (since `getOwner(this)` returns undefined).

Specifically, using Ember Data 1.13.<latest> with Ember 2.4.5 and having a model with an injected property will trigger an assertion.


dgeb:
Nice little fix @rwjblue :shipit:

PR #13337, opened 2016-04-14
[Glimmer2] Fix class bindings

asakusuma:
Not sure this is correct way to get the class name out of the binding object.

### DIFF
--- a/packages/ember-glimmer/lib/environment.js
+++ b/packages/ember-glimmer/lib/environment.js
@@ -62,6 +62,39 @@ function wrapClassAttribute(args) {
   return args;
 }
 
+function wrapClassBindingAttribute(args) {
+  let hasClassBinding = args.named.has('classBinding');
+
+  if (hasClassBinding) {
+    let { value , type } = args.named.at('classBinding');
+    if (type === 'value') {
+      let [ prop, truthy, falsy ] = value.split(':');
+      let spec;
+      if (prop === '') {
+        spec = ['helper', ['-class'], [truthy], null];
+      } else if (falsy) {
+        let parts = prop.split('.');
+        spec = ['helper', ['-class'], [['get', parts], truthy, falsy], null];
+      } else if (truthy) {
+        let parts = prop.split('.');
+        spec = ['helper', ['-class'], [['get', parts], truthy], null];
+      }
+
+      if (spec) {
+        let syntax;
+        if (args.named.has('class')) {
+          // If class already exists, merge
+          let classValue = args.named.at('class').value;
+          syntax = HelperSyntax.fromSpec(['helper', ['concat'], [classValue, ' ', spec]]);
+        } else {
+          syntax = HelperSyntax.fromSpec(spec);
+        }
+        args.named.add('class', syntax);
+      }
+    }
+  }
+}
+
 export default class Environment extends GlimmerEnvironment {
   static create(options) {
     return new Environment(options);
@@ -94,6 +127,7 @@ export default class Environment extends GlimmerEnvironment {
         let definition = this.getComponentDefinition(path);
 
         if (definition) {
+          wrapClassBindingAttribute(args);
           wrapClassAttribute(args);
           return new CurlyComponentSyntax({ args, definition, templates });
         }
--- a/packages/ember-glimmer/lib/helpers/-class.js
+++ b/packages/ember-glimmer/lib/helpers/-class.js
@@ -3,14 +3,21 @@ import { dasherize } from 'ember-runtime/system/string';
 
 function classHelper({ positional }) {
   let path = positional.at(0);
-  let propName = positional.at(1);
+  let truthyPropName = positional.at(1);
+  let falsyPropName = positional.at(2);
   let value = path.value();
 
   if (value === true) {
-    return dasherize(propName.value());
+    if (truthyPropName) {
+      return dasherize(truthyPropName.value());
+    }
+    return null;
   }
 
   if (value === false) {
+    if (falsyPropName) {
+      return dasherize(falsyPropName.value());
+    }
     return null;
   }
 
--- a/packages/ember-glimmer/tests/integration/components/curly-components-test.js
+++ b/packages/ember-glimmer/tests/integration/components/curly-components-test.js
@@ -9,6 +9,9 @@ import { classes, equalTokens, equalsElement } from '../../utils/test-helpers';
 import { htmlSafe } from 'ember-htmlbars/utils/string';
 import { computed } from 'ember-metal/computed';
 
+const bindingDeprecationMessage = '`Ember.Binding` is deprecated. Consider' +
+  ' using an `alias` computed property instead.';
+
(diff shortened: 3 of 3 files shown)


packages/ember-glimmer/tests/utils/test-helpers.js
```diff
@@ -1,5 +1,7 @@
 import { tokenize } from 'simple-html-tokenizer';
 
+let whitespace = new RegExp('( )+', 'g');
```
krisselden: `const WHITESPACE = /\s+/;`

stefanpenner: isn't this the same as `split(' ' )` ?

chadhietala:
I'm not sure if this the approach we want to take. After talking @wycats it seems like glimmer needs to expose a `refineExpression` hook in the environment to compliment `refineStatement`. This is so that statements don't need to "peek" inside to handle functionality like this.

asakusuma:
Looks like saucelabs firefox is crapping out

```
Error: [init({"browserName":"firefox","version":"latest","platform":"","deviceName":"","device-orientation":"","tags":["Saucie","test"],"name":"Saucie tests","public":"public","build":"18176","tunnel-identifier":"18176.6","record-video":true})] The environment you requested was unavailable.
```

asakusuma:
Build is passing now

packages/ember-glimmer/lib/environment.js
```diff
@@ -62,6 +62,29 @@ function wrapClassAttribute(args) {
...
+
+  if (hasClassBinding) {
+    let { value , type } = args.named.at('classBinding');
+    if (type === 'value') {
+      let [ prop, truthy, falsy ] = value.split(':');
+      if (prop === '') {
+        let syntax = HelperSyntax.fromSpec(['helper', ['-class'], [truthy], null]);
+        args.named.add('class', syntax);
+      } else if (falsy) {
+        let parts = prop.split('.');
```
chadhietala: Can't you read the `parts` of the `Ref` syntax?

packages/ember-glimmer/lib/environment.js
```diff
@@ -62,6 +62,29 @@ function wrapClassAttribute(args) {
...
+function wrapClassBindingAttribute(args) {
+  let hasClassBinding = args.named.has('classBinding');
+
+  if (hasClassBinding) {
+    let { value , type } = args.named.at('classBinding');
+    if (type === 'value') {
+      let [ prop, truthy, falsy ] = value.split(':');
+      if (prop === '') {
+        let syntax = HelperSyntax.fromSpec(['helper', ['-class'], [truthy], null]);
+        args.named.add('class', syntax);
```
chadhietala: I think the issue with this is that you are causing some extra work being done. Consider that you are taking the component call site and going from:

```handlebars
{{foo-bar classBinding=foo}}
```
to
```handlebars
{{foo-bar classBinding=foo class=(-class 'truthy') }}
```

What I think you should do is probably replacing at the callsite:

```handlebars
{{foo-bar classBinding=(-class 'truthy')}}
```

In the end it may not matter but from a debugging point of view it may be weird.

packages/ember-glimmer/lib/helpers/-class.js
```diff
@@ -3,14 +3,21 @@ import { dasherize } from 'ember-runtime/system/string';
...
 function classHelper({ positional }) {
   let path = positional.at(0);
-  let propName = positional.at(1);
+  let truthyPropName = positional.at(1);
+  let falsyPropName = positional.at(2);
   let value = path.value();
 
   if (value === true) {
-    return dasherize(propName.value());
+    if (truthyPropName) {
```
chadhietala: When is this not the case?

asakusuma: This case: https://github.com/emberjs/ember.js/pull/13337/files#diff-c47395a5aaf92aa0a7edbfa812d4e0afR218

asakusuma:
@rwjblue I just rebased and now the tests are failing because of a deprecation being thrown as an error.

```
Uncaught Error: `Ember.Binding` is deprecated. Consider using an `alias` computed property instead.
```

It's unclear to me if this is expected, and if so, does that mean that classBinding should not be implemented for glimmer?

Looks like this PR introduced the deprecation: https://github.com/emberjs/ember.js/pull/13424

chadhietala:
@asakusuma It needs to be implemented as `Ember.Binding` won't be removed till 3.0.0. https://github.com/emberjs/ember.js/pull/13424/files#diff-d50d2452ca8c5f57019c1974e6a3fb52R298

rwjblue:
All tests for `classBinding` will have to `expectDeprecation`.

packages/ember-glimmer/lib/environment.js
asakusuma: @chadhietala in this case the object returned from `args.named.at('classBinding')` only has `type` and `value` properties. There's no `ref` property.

asakusuma:
@chadhietala I think we were overcomplicating this. Update PR is now a lot simpler. Let me know if this is not a public API, but seems public to me.

krisselden:
This is creating the binding in the wrong scope.  It is supposed to be the template scope not the component. Unfortunately the test does not actually cover bindings, just static classes, but doing a quick search of github, this does seemed to be a used feature though undocumented and not tested.

packages/ember-glimmer/lib/environment.js
asakusuma: So if we get rid of the classBinding param, the deprecation doesn't happen. We can obviously add a deprecation here, but not sure it's worth it for a feature that's on it's way out anyways.

packages/ember-glimmer/lib/environment.js
```diff
@@ -62,6 +62,39 @@ function wrapClassAttribute(args) {
...
+      } else if (truthy) {
+        let parts = prop.split('.');
+        spec = ['helper', ['-class'], [['get', parts], truthy], null];
+      }
+
+      if (spec) {
+        let syntax;
+        if (args.named.has('class')) {
+          // If class already exists, merge
+          let classValue = args.named.at('class').value;
```
asakusuma: This currently won't work with helpers. If we need to support this, we'll need a way to generate a spec from a helper.

packages/ember-glimmer/lib/environment.js
```diff
@@ -62,6 +62,39 @@ function wrapClassAttribute(args) {
...
+function wrapClassBindingAttribute(args) {
+  let hasClassBinding = args.named.has('classBinding');
+
+  if (hasClassBinding) {
+    let { value , type } = args.named.at('classBinding');
+    if (type === 'value') {
+      let [ prop, truthy, falsy ] = value.split(':');
+      let spec;
+      if (prop === '') {
+        spec = ['helper', ['-class'], [truthy], null];
```
krisselden: doesn't need the helper

rwjblue:
Thank you @asakusuma!

PR #13343, opened 2016-04-15
[DOC beta] Update Route#loading example

### LINKED ISSUE #13341: [DOC] Need better example for Route#loading
https://github.com/emberjs/ember.js/blob/v2.5.0/packages/ember-routing/lib/system/route.js#L615-L622
Uses deprecated Ember.View API in the example. I would bet this is confusing to new users coming to Ember and seeing this hook.

duggiefresh:
Fixes #13341

### DIFF
--- a/packages/ember-routing/lib/system/route.js
+++ b/packages/ember-routing/lib/system/route.js
@@ -612,16 +612,12 @@ var Route = EmberObject.extend(ActionHandler, Evented, {
     App.ApplicationRoute = Ember.Route.extend({
       actions: {
         loading: function(transition, route) {
-          var view = Ember.View.create({
-            classNames: ['app-loading']
-          })
-          .append();
+          let controller = this.controllerFor('foo');
+          controller.set('currentlyLoading', true);
 
-          this.router.one('didTransition', function() {
-            view.destroy();
+          transition.finally(function() {
+            controller.set('currentlyLoading', false);
           });
-
-          return true; // Bubble the loading event
         }
       }
     });


packages/ember-routing/lib/system/route.js
```diff
@@ -605,23 +605,19 @@ var Route = EmberObject.extend(ActionHandler, Evented, {
...
-          var view = Ember.View.create({
-            classNames: ['app-loading']
-          })
-          .append();
+          let controller = this.controllerFor('foo');
+          controller.set('currentlyLoading', true);
 
-          this.router.one('didTransition', function() {
-            view.destroy();
+          transition.promise.finally(function() {
```
stefanpenner: is the promise public?

duggiefresh: [Looking at the router, it is marked as `public`](https://github.com/tildeio/router.js/blob/master/lib/router/transition.js#L104-L115) :smile:

Serabe: It is, but as DOC says, it is there if you want to pass around a promise without passing the full transition. In this case,  I think [`transition.finally`](https://github.com/tildeio/router.js/blob/master/lib/router/transition.js#L167-L182) would do too.

duggiefresh: Good call, will update :+1:

Serabe: Thank you.

rwjblue:
Thanks @duggiefresh!

PR #13344, opened 2016-04-15
[BUGFIX release] Remove Ember.merge deprecation.

rwjblue:
The deprecation was too noisy, and not enough libraries updated prior to release (to avoid deprecation). At this point the deprecation is just a nuisance to our users.


chadhietala:
LGTM

workmanw:
Just to clarify, the plan is still to eventually deprecate and remove `Ember.merge`? This is just a removal of the warning for now until the rest of us addon maintainers get caught up?

stefanpenner:
@workmanw i believe the plan is to re-evaluate at a later date.

rwjblue:
Confirm. We do not currently have plans to reintroduce, but I would be somewhat surprised if both methods make it into 3.0.0.  But obviously that is quite a ways away, and we don't really know what that will look like.

workmanw:
@stefanpenner @rwjblue Thanks for the clarifications!

PR #13345, opened 2016-04-15
[Glimmer2] Migrate local lookup tests

asakusuma:
Not sure if this is the right way to simulate the resolver going forward.


rwjblue:
Assigning to myself for review, will try to review tonight.

rwjblue:
I'm unsure what `packages/container/tests/test-helpers/.Rhistory` is (maybe an editor artifact?).

packages/container/tests/test-helpers/build-owner.js
```diff
@@ -3,11 +3,39 @@ import Registry from 'container/registry';
...
+  return resolver;
+}
+
 export default function buildOwner(props) {
   let Owner = EmberObject.extend(RegistryProxy, ContainerProxy, {
     init() {
       this._super(...arguments);
-      let registry = new Registry(this._registryOptions);
+      let registry = new Registry(this._registryOptions || {
+        resolver: buildResolver()
```
rwjblue: I don't think that all tests should get a local lookup enabled resolver.  Especially since the actual patterns for how local lookup will work are still up in the air (see https://github.com/emberjs/rfcs/pull/124).   It seems better (to me) to inline this change, and the resolver above in the local-lookup-test file until those things are settled...

rwjblue:
We should also update `ApplicationTest`'s `registerTemplate` method in `abstract-test-case.js` to pass in the source.

asakusuma:
Got rid of `packages/container/tests/test-helpers/.Rhistory`

When you say pass in the source, are you talking about `moduleName`? I added that to registerTemplate

rwjblue:
> When you say pass in the source, are you talking about moduleName? I added that to registerTemplate

Yes, I meant `moduleName`, sorry about the confusion.

rwjblue:
This is looking good, I see you still have it labeled as `WIP` though.  Ping me when you are ready for another round of review...

asakusuma:
@rwjblue ok I removed the WIP as I have attempted to migrate all the local lookup tests now. Except for the one test that was for when `ember-htmlbars-local-lookup=false`, since the flag is true in `features.json`.

rwjblue:
@asakusuma - I believe [this file](https://github.com/emberjs/ember.js/blob/master/packages/ember-htmlbars/tests/integration/local-lookup-test.js) should be deleted and a symlink setup (to match the other sets of ember-glimmer / ember-htmlbars integration tests.  Also, the commit still has `[WIP]` prefixed, can you remove that as well?

Let me know when ready for the next round of review...

asakusuma:
@rwjblue ready for review. I deleted the old test, but the parent folder of the new test is already symlinked. `[WIP]` prefixed has been removed from commit.

rwjblue:
Ahh, I see, I didn't realize the parent folder was symlinked.  Thank you for pointing that out.

rwjblue:
Thank you @asakusuma!

Issue #11023, opened 2015-05-05
[Glimmer] - Data binding issue when using Ember.computed.alias


raytiley:
Here is a jsbin reproducing an issue in my app. The `action-checkbox` fires an action when clicked and the `x-item` component  then adds / removes it from an array to manage selection. 

My current implementation uses an alias inside the `action-checkbox` to alias `isChecked` to `checked` which is an attribute binding. After repeated checks the `selected` CP on the `x-item` component stops being calculated and is out of sync. You can see from the console logs that the `toggle` action on the `x-item` component is continuing to fire.

Broken: http://emberjs.jsbin.com/cisitoluho/1/edit?html,js,output

Working w/ out alias:http://emberjs.jsbin.com/docosi/1/edit?html,js,output

Worked with @rwjblue a bit on this and he asked me to open the issue.

mixonic:
@raytiley this is different than https://github.com/emberjs/ember.js/issues/10992? Seems similar.

rwjblue:
Thanks for opening this issue.  Will be working on this after #11013.

raytiley:
@mixonic not sure... I didn't fully grok that other jsbin when I played around with it. In my case though its definitely a regression in glimmer... Works fine in canary.

rwjblue:
I did a bit more work on this, and have a further reduction JSBin: http://emberjs.jsbin.com/rwjblue/461/edit?html,js,output.

rwjblue:
Trying to work on a test that demonstrates the bug.

rwjblue:
Even simpler repro: http://emberjs.jsbin.com/rwjblue/463/edit?html,js

---

This bug only exists when:

* using a `mut` binding backed by a CP
* the bound attribute exists in the component
* the upstream value is changed a second time :trollface:

rwjblue:
Test added in https://github.com/emberjs/ember.js/compare/computed-attribute-updates.

PR #13351, opened 2016-04-17
[BUGFIX release] Setting a property during `init` flows upstream.

### LINKED ISSUE #11192: Can't pass component to controller
To get access a component in a controller, I set the component as a property on itself which exposes it to the controller([link](http://www.samselikoff.com/blog/getting-ember-components-to-respond-to-actions/)).

- Works in Ember 1.12: http://emberjs.jsbin.com/wokelikice/1/edit
- Fails in 1.13.0-beta.1: http://jsbin.com/gohovifedi/2/edit
- Also tried using ember-cli: http://vast-plate.surge.sh

Not sure what I need to change to make it work again?

rwjblue:
This `legacyPropertyDidChange` method is used by the `AttrsProxy` to properly set the upstream values, it is included in the `default` state [here](https://github.com/emberjs/ember.js/blob/master/packages/ember-views/lib/views/states/default.js#L23-L34) (so it is available in all other states), but `preRender` state (which is where `init` is executed in) was being overridden with a noop.  This overrride was introduced in https://github.com/emberjs/ember.js/pull/11081, however the implementation of the `AttrsProxy` completely changed from what was causing the original reported regression (https://github.com/emberjs/ember.js/issues/11023) and the test added for that regression still passes when removing this change.

Fixes #11192.

### DIFF
--- a/packages/ember-glimmer/tests/integration/components/curly-components-test.js
+++ b/packages/ember-glimmer/tests/integration/components/curly-components-test.js
@@ -2101,4 +2101,40 @@ moduleFor('Components test: curly components', class extends RenderingTest {
 
     this.assertText('initial value');
   }
+
+  ['@htmlbars ensure attributes which have been passed in can be mutated, when they are mutated the external source is updated (#11192)']() {
+    let component;
+    let FooBarComponent = Component.extend({
+      init() {
+        this._super(...arguments);
+        component = this;
+        this.set('bar', 'hellllloooooo');
+      }
+    });
+
+    this.registerComponent('foo-bar', {
+      ComponentClass: FooBarComponent,
+      template: ''
+    });
+
+    expectDeprecation(() => {
+      this.render('{{localBar}}{{foo-bar bar=localBar}}', {
+        localBar: undefined
+      });
+    }, /You modified localBar twice in a single render/);
+
+    this.assertText('hellllloooooo');
+
+    this.runTask(() => this.rerender());
+
+    this.assertText('hellllloooooo');
+
+    this.runTask(() => { component.set('bar', 'updated value'); });
+
+    this.assertText('updated value');
+
+    this.runTask(() => { component.set('bar', 'hellllloooooo'); });
+
+    this.assertText('hellllloooooo');
+  }
 });
--- a/packages/ember-views/lib/views/states/pre_render.js
+++ b/packages/ember-views/lib/views/states/pre_render.js
@@ -1,5 +1,4 @@
 import _default from 'ember-views/views/states/default';
-import assign from 'ember-metal/assign';
 
 /**
 @module ember
@@ -7,9 +6,5 @@ import assign from 'ember-metal/assign';
 */
 
 let preRender = Object.create(_default);
-
-assign(preRender, {
-  legacyPropertyDidChange(view, key) {}
-});
 
 export default preRender;


packages/ember-glimmer/tests/integration/components/curly-components-test.js
```diff
@@ -2101,4 +2101,40 @@ moduleFor('Components test: curly components', class extends RenderingTest {
 
     this.assertText('initial value');
   }
+
+  ['@htmlbars a two way binding flows upstream when set during init (#11192)']() {
```
stefanpenner: this description implies change propagation happens during init, i believe (if i understand correctly, but correct me if im wrong) it means that properties set during init, can later change and propogate change events.

rwjblue: Hmm. I think you are correct, but in this case the change events are not on the object that is running `init`, but on the thing that has passed in a two way bound property.  This was definitely one of the regressions introduced in 1.13 (same issue that caused `viewName` to not work properly).

FWIW - Doing this, does (properly) trigger a deprecation about causing a rerender.

rwjblue:
@stefanpenner - I wasn't sure from your comment above if you were opposed to this change, or just clarifying what was happening.  r?

stefanpenner:
@rwjblue i am not apposed, but was merely asking that the test description be improved.

packages/ember-glimmer/tests/integration/components/curly-components-test.js
stefanpenner: maybe a better nomenclature is external attributes vs internal attributes. Rather then describing this as upstream or downstream binding propagation.

stefanpenner: maybe something like: `ensure attributes which have been passed in can be mutated, when they are mutated the external source is notified`

rwjblue: No, no, I didn't take it as pedantic, but I wasn't sure (until you clarified) if the issue was with my horrible prose, or the concept being tested.

rwjblue: No, no, I didn't take it as pedantic, but I wasn't sure (until you clarified) if the issue was with my horrible prose, or the concept being tested.

rwjblue:
@stefanpenner - OK, updated using your suggestion. Should be ready for another review...

stefanpenner:
👍  lets get @krisselden to double check though.

packages/ember-glimmer/tests/integration/components/curly-components-test.js
```diff
@@ -2101,4 +2101,40 @@ moduleFor('Components test: curly components', class extends RenderingTest {
...
+      template: ''
+    });
+
+    expectDeprecation(() => {
+      this.render('{{localBar}}{{foo-bar bar=localBar}}', {
+        localBar: undefined
+      });
+    }, /You modified localBar twice in a single render/);
+
+    this.assertText('hellllloooooo');
```
krisselden: I would not have expected this to work, since I would not expect that set() to cause a change event that would flow back to previous sibling view, by what magic does this work? and people really do this, I hope not...

krisselden:
I'm scared of this change to be quite honest.  If it has been broken this long and is for deprecated behavior (causing render to invalidate during rendering was deprecated before 2.0) why are we bringing it back now?

The idea of sets in a component flowing back to already rendered things is about the worst of 2 way bindings as you can get.  The initial value should flow down, the back flow should be from something like an input change event.

I'm sorry we didn't fix this as part of 1.13 it clearly was a regression then but I've many doubts about introducing it now.

/cc @wycats

rwjblue:
The common usage of this was similar to what folks did for `viewName`:

```hbs
<label>Title: {{input value=audioPlayer.title}}</label>
  
{{audio-player title="Michael Jackson" register-as=audioPlayer}}
```

Where `AudioPlayer` component does:

```js
init() {
  this._super(...arguments);
  this.set('register-as', this);
}
```

---

A few points here:

* I agree that this is deprecated behavior and it absolutely triggers a deprecation, but if you look at that deprecation is shows `until: '3.0.0'` which means it is still be supported.
* In the PR description I linked to the change that introduced the no-op. The no op was not introduced for philosophical reasons, but for practical ones: the attrs proxy that we were using at the time was broken and caused a cycle when things were set outside of `init` (https://github.com/emberjs/ember.js/pull/11081). This is no longer an issue (as shown by this PR), and the original hack (overriding `legacyPropertyDidChange` with a noop) is not needed.
* The current state is extremely arbitrary. We allow this sort of behavior in other hooks (i.e. `didInsertElement`, `didUpdateElement`, and `didRender`), but not from `init`. Having to know which hooks this will and will not work from is pretty surprising.

krisselden:
To others reading this thread, the difference is the old viewName stuff or parentView, the set was onto something that was initialized and then passed to `init()`.  The attrs proxy thing is done via a back door in Ember.set() if we remove this noop for init so that a set() on `this` can propagate changes, that is inconsistent about partially initialized objects propagating change events.

PR #13168, opened 2016-03-23
[Glimmer 2] Implement component lifecycle hooks

chancancode:

PR #13353, opened 2016-04-18
[BUGFIX beta] Fix revalidation hook messages.

### LINKED ISSUE #11493: "A property ... was modified inside the didInsertElement hook" deprecation warnings (v1.12.0 -> v1.13.0)
After upgrading from 1.12.0 to 1.13.0 (and Ember Data 1.0.0-beta.19.1) I am seeing a lot of these warnings:

`DEPRECATION: A property of <Ember.OutletView:ember1021> was modified inside the didInsertElement hook. You should never change properties on components, services or models during didInsertElement because it causes significant performance degradation.`

However, I have very few `didInsertElement` hooks, they only read properties and set up jQuery stuff, and I see this message even when none of them are being called.

What else might cause this? Also, why isn't this deprecation in [the guide](http://emberjs.com/deprecations/v1.x/)?

### LINKED ISSUE #11505: Wrong view and hook in deprecation of render revalidation during render
If you change state during render, the deprecation message output tells you the wrong view that triggered it and has the hook hard coded to didInsertElement.

http://emberjs.jsbin.com/boyoxihija/1/edit?html,js,console,output

rwjblue:
The source of the revalidation was being lost, and the message indicated the root `ownerView` (which was not the thing that triggered a rerender). Also, the `_dispatching` flag was not being changed after calling `didInsertElement` / `didUpdateElement` and before calling `didRender`.

Fixes #11505.
Fixes #11493.

### DIFF
--- a/packages/ember-htmlbars/lib/renderer.js
+++ b/packages/ember-htmlbars/lib/renderer.js
@@ -85,6 +85,7 @@ Renderer.prototype.dispatchLifecycleHooks =
         case 'didUpdate': this.didUpdate(hook.view); break;
       }
 
+      ownerView._dispatching = 'didRender';
       this.didRender(hook.view);
     }
 
--- a/packages/ember-htmlbars/lib/utils/subscribe.js
+++ b/packages/ember-htmlbars/lib/utils/subscribe.js
@@ -21,6 +21,6 @@ export default function subscribe(node, env, scope, stream) {
       node.shouldReceiveAttrs = true;
     }
 
-    node.ownerNode.emberView.scheduleRevalidate(node, labelFor(stream));
+    node.ownerNode.emberView.scheduleRevalidate(node, labelFor(stream), undefined, component);
   }));
 }
--- a/packages/ember-views/lib/mixins/view_support.js
+++ b/packages/ember-views/lib/mixins/view_support.js
@@ -508,7 +508,7 @@ export default Mixin.create({
     this.scheduledRevalidation = false;
   },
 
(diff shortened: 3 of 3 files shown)


rwjblue:
I need to add tests, but the lifecycle tests do not exist yet for ember-glimmer in the new structure (that is being done in https://github.com/emberjs/ember.js/pull/13168).  I'd like to just add tests to that PR (separately from this) in an attempt to avoid writing old style tests, that then just have to be manually ported again to the new structure.

Any objections?

rwjblue:
https://github.com/emberjs/ember.js/pull/13168 has landed, I'll try to get a test added here so we can land this.

packages/ember-metal-views/lib/htmlbars-renderer.js
```diff
@@ -84,6 +84,7 @@ Renderer.prototype.dispatchLifecycleHooks =
         case 'didUpdate': this.didUpdate(hook.view); break;
       }
 
+      ownerView._dispatching = 'didRender';
```
stefanpenner: just double checking, `_dispatching` is already defined in the constructor of ownerView ?

rwjblue: Yes, it is initialized to `null` by `CoreView.prototype.init` [here](https://github.com/emberjs/ember.js/blob/e81233a6aa8cb29b72b92e799be510ad66046db7/packages/ember-views/lib/views/core_view.js#L47).

PR #13355, opened 2016-04-18
[BUGFIX beta] Fix issue with `Ember.trySet` on destroyed objects.

### LINKED ISSUE #12356: `Ember.trySet()` throws 'Assertion Failed: calling set on destroyed object'
From the documentation:

> Error-tolerant form of Ember.set. Will not blow up if any part of the chain is undefined, null, or **destroyed**.

1.13.10:
http://emberjs.jsbin.com/sufaji/edit?html,js,console

2.0.2:
http://emberjs.jsbin.com/yafacu/edit?html,js,console

Open up the console to see the assertion error.

rwjblue:
`Ember.trySet` is designed to be an error-tolerant form of `Ember.set` (per its public API documentation). However, in some circumstances an error is still thrown when setting on a destroyed object.

Prior to this change, the following would properly prevent an error:

``` js
let obj = { some: { deep : 'path', isDestroyed: true }};

Ember.trySet(obj, 'some.deep', 'stuff');
```

But the following would throw an error incorrectly:

``` js
let obj = { some: 'path', isDestroyed: true };

Ember.trySet(obj, 'some', 'stuff');
```

This fixes the latter case...

Fixes #12356.

### DIFF
--- a/packages/ember-metal/lib/property_set.js
+++ b/packages/ember-metal/lib/property_set.js
@@ -47,7 +47,11 @@ export function set(obj, keyName, value, tolerant) {
   assert(`Cannot call set with '${keyName}' on an undefined object.`, obj && typeof obj === 'object' || typeof obj === 'function');
   assert(`The key provided to set must be a string or number, you passed ${keyName}`, typeof keyName === 'string' || (typeof keyName === 'number' && !isNaN(keyName)));
   assert(`'this' in paths is not supported`, typeof keyName !== 'string' || keyName.lastIndexOf('this.', 0) !== 0);
-  assert(`calling set on destroyed object: ${toString(obj)}.${keyName} = ${toString(value)}`, !obj.isDestroyed);
+
+  if (obj.isDestroyed) {
+    assert(`calling set on destroyed object: ${toString(obj)}.${keyName} = ${toString(value)}`, tolerant);
+    return;
+  }
 
   if (isPath(keyName)) {
     return setPath(obj, keyName, value, tolerant);
--- a/packages/ember-metal/tests/accessors/set_test.js
+++ b/packages/ember-metal/tests/accessors/set_test.js
@@ -1,6 +1,7 @@
 import {
   get,
   set,
+  trySet,
   setHasViews
 } from '../..';
 import { moduleFor, AbstractTestCase } from 'internal-test-helpers';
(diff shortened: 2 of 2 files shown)


rwjblue:
@krisselden / @stefanpenner - r?

stefanpenner:
Looks ok, but now that if conditional won't be stripped

rwjblue:
@stefanpenner - Yep, (a similar `if` is also present in `setPath` below this). If I embed the check into the assertion itself (which is possible) then we would actually do a `set` on a destroyed object when in production (because there is nothing left in the code to do a short-circuit).  Is that better?

stefanpenner:
Oops you are correct. This looks good to me

rwjblue:
@krisselden - Any objections?

rwjblue:
I'm going to assign to @krisselden for review.  I want to make sure that we are all happy/ok with this...

krisselden:
I'm ok with this, though, isDestroyed isn't really an ember-metal concept, we should have a value we set meta to, to mean it is destroyed.

Maybe then we could fix the long standing issue of recreating meta() on a destroyed object.
/cc @stefanpenner

rwjblue:
@krisselden - Would you prefer that to happen in this PR or can it happen subsequently?

krisselden:
It can be after, I'm sure enforcing not recreating meta will expose some bugs

krisselden:
Please though open a tracking issue, I keep forgetting about. I think our teardown sometimes recreates the mets itself

packages/ember-metal/lib/property_set.js
```diff
@@ -62,8 +62,11 @@ export function set(obj, keyName, value, tolerant) {
     return setPath(obj, keyName, value, tolerant);
   }
 
-  assert(`calling set on destroyed object: ${toString(obj)}.${keyName} = ${toString(value)}`,
-         !obj.isDestroyed);
+  if (obj.isDestroyed) {
```
stefanpenner: @krisselden should we actually allow setting through destroyed objects or to the descs? My intuition leads me to believe this should actually occur before line 53.

stefanpenner: basically 3 scenarios exist.

1. set on plain object property
2. set on descriptor
3. set on deep object

We currently guard/assert for 1, and I believe we must absolutely also guard/assert for 2. I also believe we should consider the same for 3, if others disagree I would like to be convinced.

vvohra:
I was wondering about the status of this PR.

Looking at this [twiddle](https://ember-twiddle.com/cfe9901bb9d15917bde3b7b3375af1f1?openFiles=components.my-component.js%2Ctemplates.components.my-component.hbs) and checking the console, it seems the issue is still valid on Ember 2.12.

mmun:
@rwjblue I've rebased this PR onto master.

PR #13356, opened 2016-04-18
[BUGFIX beta] Update Registry#has to always return true/false.

### LINKED ISSUE #12082: container.has() behavior has been changed without deprecations warning
`container.has('route:http://www.google.com')` used to return false (pre-Ember 1.13.x), it now throws an error. No deprecation warnings were previously logged.

rwjblue:
Fixes #12082.

### DIFF
--- a/packages/container/lib/registry.js
+++ b/packages/container/lib/registry.js
@@ -5,7 +5,7 @@ import EmptyObject from 'ember-metal/empty_object';
 import assign from 'ember-metal/assign';
 import Container from './container';
 
-var VALID_FULL_NAME_REGEXP = /^[^:]+.+:[^:]+$/;
+var VALID_FULL_NAME_REGEXP = /^[^:]+:[^:]+$/;
 
 /**
  A registry used to store factory and option information keyed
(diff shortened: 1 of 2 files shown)


packages/container/lib/registry.js
```diff
@@ -692,9 +694,13 @@ Registry.prototype = {
     return assign({}, fallbackKnown, localKnown, resolverKnown);
   },
 
-  validateFullName(fullName) {
+  validateFullName(fullName, tolerant) {
```
dgeb: How about this alternative: 

* Create a new method `isValidFullName` that returns a boolean and never throws. 
* `validateFullName` could internally call `isValidFullName`, as could `has`.
* `validateFullName` could continue to throw.

I think this works because `validateFullName` sounds imperative and I would expect it to throw, plus the `tolerant` param seems a bit iffy.

rwjblue: 👍 will update

rwjblue: OK, just updated.  Thanks again, this is a much better pattern.

dgeb:
Looks a lot better to me. Thanks @rwjblue 👍

PR #13016, opened 2016-02-26
[FEATURE ember-route-serializers] Introduce serialize method to Router DSL options and deprecate Route#serialize

trentmwillis:
Implementation of emberjs/rfcs#120.

TODO:
- [x] Add entry to `FEATURES.md`
- [x] Add entry to emberjs/website deprecation guide for `ember-routing.serialize-function` (and update `deprecate` calls here to include the `url`) [emberjs/website#2557]
- [x] Review documentation, and determine update needs for guides and API docs (making an issue/PR in emberjs/guides for changes needed there, and an issue in emberjs/ember.js with a list of updates needed here when the feature is enabled). [emberjs/guides#1380, emberjs/ember.js#13360]

PR #13360, opened 2016-04-18
Update documentation for route-serializers

trentmwillis:
For merging once emberjs/ember.js#13016 has been enabled.


trentmwillis:
Discarding since this is no longer going to be enabled globally.

PR #13363, opened 2016-04-18
[BUGFIX beta] Enumerable#any bug - not always returning a boolean

### LINKED ISSUE #12533: Enumerable#any vs Array#some
In the docs, it says `Enumerable#any` should return a boolean, but this isn't necessarily true. It returns the first truthy value returned or the last falsey value returned.

This is inconsistent with `Array#some` as well, which always returns `true` or `false`.

Is this correct functionality? Either way, I can make either a PR to the docs or this repo to correct the behavior/explanation.

bryanaka:
closes #12533 

`Enumerable#any` is supposed to return only a boolean, but is currently returning whatever truthy value is returned in the callback. This changed functionality is now aligned with `Array#some`. 

Potentially a breaking change if someone was relying on the buggy functionality. It was introduced here - https://github.com/emberjs/ember.js/pull/4144/

### DIFF
--- a/packages/ember-runtime/lib/mixins/enumerable.js
+++ b/packages/ember-runtime/lib/mixins/enumerable.js
@@ -667,7 +667,7 @@ var Enumerable = Mixin.create({
 
     next = last = null;
     context = pushCtx(context);
-    return found;
+    return !!found;
   },
 
   /**
--- a/packages/ember-runtime/tests/suites/enumerable/any.js
+++ b/packages/ember-runtime/tests/suites/enumerable/any.js
@@ -19,7 +19,7 @@ suite.test('any should should invoke callback on each item as long as you return
     found.push(i);
     return false;
   });
-  equal(result, false, 'return value of obj.any');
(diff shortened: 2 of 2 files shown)


bryanaka:
hate to be a bother, but was wondering if this PR is valid or not? 

I know it has some possibly far-reaching consequences (people relying on the buggy behavior), and the fact that it may be removed with iterators and more of the newer Array functions like Array#some being so well supported.

I'm cc'ing original commenters on the #12533 @pixelhandler @Serabe

locks:
Thank you for this effort and sorry for having let it sit for so long 🙇.

I have to side with your argument about the possible breaking changes this bugfix will introduce.
Regardless of documentation or intent, the current behaviour of `any` is the defacto one, being in effect for nearly four years.
What I believe we have to do in this case is to address the documentation of `any` to mention this quick and move from `any` to `some`. There is already [an RFC](https://github.com/emberjs/rfcs/pull/233) that intends to do this. Let's figure out the path forward there :)

PR #12737, opened 2015-12-21
LocationApi:: Added notes about how to extend

danDanV1:
Put some notes about how ember scans /app/locations/\* and an example of how to extend to make a custom location.

PR #13365, opened 2016-04-18
enhance custom location API documentation 

H1D:
didn't  find anything on how to register cutom location implementation for router.

Not sure should i use `ember-cli` code examples with `export default {` instead of `Em.Application.initializer({`


rwjblue:
This seems similar to https://github.com/emberjs/ember.js/pull/12737 (i.e. how to have a custom location implementation).  Can you update to avoid the use of an initializer?

H1D:
@rwjblue done

rwjblue:
There is some extra trailing whitespace that is causing the build to fail here.

H1D:
@rwjblue looks good now

rwjblue:
Thank you @H1D!

PR #13366, opened 2016-04-19
[Glimmer2] Add test for component displaying a property

Serabe:
While taking a look at implementing positional params @chancancode realised there is no test for components where it just displays a parameter.

This fixes it.


rwjblue:
Thanks @serabe!

PR #13367, opened 2016-04-19
Remove extra whitespace from FEATURES.md

seanjohnson08:
Just noticed that after rebasing, my changes to FEATURES.md were indented one space too much. This causes the bullet point to be nested in the point above it, rather than at the top level.

See: https://github.com/emberjs/ember.js/blob/master/FEATURES.md


locks:
lgtm

rwjblue:
Thanks!

PR #13369, opened 2016-04-19
More `Ember` global cleanup

krisselden:


tests/index.html
```diff
@@ -40,9 +40,7 @@
         window.Ember = {
           testing: true
         };
-        window.ENV = window.ENV || {
-          _ENABLE_LEGACY_VIEW_SUPPORT: true
```
rwjblue: The tests still use this flag. A number of tests still use the {{view helper. A great many of the usages of {{view have been replaced during the migration to ember-glimmer, but many still exist (in other packages than ember-htmlbars).

rwjblue: This change is what caused the build to fail.

rwjblue:
Thanks @krisselden!

PR #13371, opened 2016-04-19
[FEATURE] Create helper integration tests

Gaurav0:
This is for: ember-cli/ember-cli#4517 and is paired with https://github.com/ember-cli/ember-cli-legacy-blueprints/pull/22

Note: I have not yet made integration tests default. I think it might be a good idea to do that in a followup PR once we know all is well and working?

### DIFF
--- a/blueprints/helper-test/index.js
+++ b/blueprints/helper-test/index.js
@@ -1,15 +1,52 @@
 /*jshint node:true*/
 
-var testInfo = require('ember-cli-test-info');
-var stringUtils = require('ember-cli-string-utils');
+var testInfo                 = require('ember-cli-test-info');
+var stringUtils              = require('ember-cli-string-utils');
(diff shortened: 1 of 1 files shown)


rwjblue:
What output does this result in?

Gaurav0:
@rwjblue Sample output from `ember g helper-test bar-baz --integration` resulting in `tests/integration/helpers/bar-baz-test.js`

```
import { moduleForComponent, test } from 'ember-qunit';

moduleForComponent('bar-baz', 'helper:bar-baz', {
  integration: true
});

test('it renders', function(assert) {
  this.set('inputValue', '1234');

  this.render(hbs`{{bar-baz inputValue}}`);

  assert.equal(this.$().text().trim(), '1234');
});
```

rwjblue:
Thanks @Gaurav0!

Gaurav0:
Thank you for reviewing, @rwjblue.

PR #13372, opened 2016-04-19
[Glimmer 2] Differentiate between bound `class` and `classNames`

chadhietala:
This fixes a behavior where a bound `class` can mutate, where as bound `classNames` are const'd post creation. Also migrating some of the attr_node tests that were testing `class` behavior.


packages/ember-glimmer/lib/components/curly-component.js
```diff
@@ -1,6 +1,7 @@
 import { StatementSyntax, ValueReference } from 'glimmer-runtime';
 import { AttributeBindingReference, RootReference, applyClassNameBinding } from '../utils/references';
 import EmptyObject from 'ember-metal/empty_object';
+import Dict from 'ember-metal/dictionary';
```
rwjblue: I think this should be an `EmptyObject` (not a dictionary) since we do not need delete from it.

rwjblue: See the comments in these two files for a relatively good explanation:

* https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/lib/empty_object.js
* https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/lib/dictionary.js

`ember-metal/dictionary` (when used with `null` parent) creates an `EmptyObject` then adds an item and deletes it to force the runtime into keeping it in dictionary mode. Since we don't need this object to be a dictionary, we can just use `new EmptyObject()` directly.

chancancode: We probably don't need it to be an actual dict since we know all the fields (like meta). We can just make an actual class for it

chancancode: The object is not exposed to users either, so it's probably not very important for it to be an empty object either

chancancode: We do want the object to have a fixed shape though even when some of those meta info are not present.

rwjblue: :+1: to using a class for this...

chadhietala: Done

packages/ember-glimmer/tests/integration/content-test.js
```diff
@@ -584,4 +585,157 @@ moduleFor('Dynamic content tests (integration)', class extends RenderingTest {
 
     this.assertText('');
   }
+
+  ['@test quoteless class attributes update correctly']() {
```
chadhietala: [Migrated](https://github.com/emberjs/ember.js/pull/13372/files#diff-b90868f410f10bc636abb45a8c6eb215L39)

packages/ember-glimmer/tests/integration/content-test.js
```diff
@@ -584,4 +585,157 @@ moduleFor('Dynamic content tests (integration)', class extends RenderingTest {
...
+    this.runTask(() => set(this.context, 'model.classes', 'fizz bizz'));
+
+    this.assertElement(this.firstChild, { tagName: 'div', content: 'hello', attrs: { 'class': classes('fizz bizz') } });
+
+    this.runTask(() => set(this.context, 'model', { classes: 'foo bar baz' }));
+
+    this.assertElement(this.firstChild, { tagName: 'div', content: 'hello', attrs: { 'class': classes('foo bar baz') } });
+  }
+
+  ['@test quoted class property can contain multiple classes']() {
```
chadhietala: [Migrated](https://github.com/emberjs/ember.js/pull/13372/files#diff-b90868f410f10bc636abb45a8c6eb215L96)

chadhietala:
This is going to fail until we bump glimmer but figured I would post it.

packages/ember-htmlbars/tests/attr_nodes/class_test.js
```diff
@@ -35,112 +35,3 @@ QUnit.test('class renders before didInsertElement', function() {
...
-  appendView(view);
-
-  strictEqual(view.element.firstChild.className, 'btn-large pre-pre pre-post    whoop', 'classes are set');
-
-  run(view, view.set, 'context.prefix', '');
-
-  strictEqual(view.element.firstChild.className, 'btn-large -post    whoop', 'classes are updated');
-});
-
-QUnit.test('class attribute stays in order', function() {
```
chadhietala: [Consolidated](https://github.com/emberjs/ember.js/pull/13372/files#diff-e493f9c2deceaafd80b881cfabc92186R675)

packages/ember-htmlbars/tests/attr_nodes/class_test.js
```diff
@@ -35,112 +35,3 @@ QUnit.test('class renders before didInsertElement', function() {
...
-  appendView(view);
-
-  strictEqual(view.element.firstChild.className, 'large small', 'classes are set');
-
-  run(view, view.set, 'context.size', 'medium');
-
-  strictEqual(view.element.firstChild.className, 'medium', 'classes are updated');
-});
-
-QUnit.test('class attribute can grok concatted classes, and update', function() {
```
chadhietala: [Consolidated](https://github.com/emberjs/ember.js/pull/13372/files#diff-e493f9c2deceaafd80b881cfabc92186R675)

packages/ember-htmlbars/tests/attr_nodes/class_test.js
```diff
@@ -35,112 +35,3 @@ QUnit.test('class renders before didInsertElement', function() {
...
-
-  strictEqual(view.element.firstChild.className, 'large blue no-shape', 'classes are set');
-
-  run(view, view.set, 'context.hasColor', false);
-  run(view, view.set, 'context.hasShape', true);
-
-  strictEqual(view.element.firstChild.className, 'large  round', 'classes are updated');
-});
-
-QUnit.test('class attribute can accept multiple classes from a single value, and update', function() {
```
chadhietala: [Migrated](https://github.com/emberjs/ember.js/pull/13372/files#diff-e493f9c2deceaafd80b881cfabc92186R653)

packages/ember-htmlbars/tests/attr_nodes/class_test.js
```diff
@@ -35,112 +35,3 @@ QUnit.test('class renders before didInsertElement', function() {
...
-  view = EmberView.create({
-    context: { size: 'large', color: 'blue' },
-    template: compile('<div class=\'{{size}} {{color}} round\'></div>')
-  });
-  appendView(view);
-
-  strictEqual(view.element.firstChild.className, 'large blue round', 'classes are set');
-});
-
-QUnit.test('class attribute accepts nested helpers, and updates', function() {
```
chadhietala: [Migrated](https://github.com/emberjs/ember.js/pull/13372/files#diff-e493f9c2deceaafd80b881cfabc92186R708)

packages/ember-htmlbars/tests/attr_nodes/class_test.js
```diff
@@ -35,112 +35,3 @@ QUnit.test('class renders before didInsertElement', function() {
...
-  appendView(view);
-
-  equal(view.element.firstChild.className, 'large', 'attribute is output');
-
-  run(view, view.set, 'context.class', null);
-
-  equal(view.element.firstChild.className, '', 'attribute is removed');
-});
-
-QUnit.test('class attribute concats bound values', function() {
```
chadhietala: [Migrated](https://github.com/emberjs/ember.js/pull/13372/files#diff-e493f9c2deceaafd80b881cfabc92186R675)

packages/ember-htmlbars/tests/attr_nodes/class_test.js
```diff
@@ -35,112 +35,3 @@ QUnit.test('class renders before didInsertElement', function() {
   equal(view.element.firstChild.className, 'blue', 'attribute is output');
   equal(matchingElement.length, 1, 'element is in the DOM when didInsertElement');
 });
-
-QUnit.test('class property can contain multiple classes', function() {
```
chadhietala: [Migrated](https://github.com/emberjs/ember.js/pull/13372/files#diff-e493f9c2deceaafd80b881cfabc92186R631)

packages/ember-htmlbars/tests/attr_nodes/class_test.js
```diff
@@ -35,112 +35,3 @@ QUnit.test('class renders before didInsertElement', function() {
...
-  });
-  appendView(view);
-
-  equalInnerHTML(view.element, '<div class="large blue"></div>',
-                 'attribute is output');
-  ok(view.$('.large')[0], 'first class found');
-  ok(view.$('.blue')[0], 'second class found');
-});
-
-QUnit.test('class property is removed when updated with a null value', function() {
```
chadhietala: [Migrated](https://github.com/emberjs/ember.js/pull/13372/files#diff-e493f9c2deceaafd80b881cfabc92186R653)

chadhietala:
I would like to 💣 that whole test file as I don't believe the first test is valid any more as it's testing `View` context stuff.

packages/ember-htmlbars/tests/attr_nodes/class_test.js
```diff
@@ -1,146 +0,0 @@
...
-
-QUnit.module('ember-htmlbars: class attribute', {
-  teardown() {
-    if (view) {
-      run(view, view.destroy);
-    }
-  }
-});
-
-QUnit.test('class renders before didInsertElement', function() {
```
chadhietala: [Migrated](https://github.com/emberjs/ember.js/pull/13372/files#diff-c47395a5aaf92aa0a7edbfa812d4e0afR160)

packages/ember-htmlbars/tests/attr_nodes/class_test.js
```diff
@@ -1,146 +0,0 @@
...
-    context: { color: 'blue' },
-    template: compile('<div class={{color}}>Hi!</div>')
-  });
-  appendView(view);
-
-  equal(view.element.firstChild.className, 'blue', 'attribute is output');
-  equal(matchingElement.length, 1, 'element is in the DOM when didInsertElement');
-});
-
-QUnit.test('class property can contain multiple classes', function() {
```
chadhietala: [Migrated](https://github.com/emberjs/ember.js/pull/13372/files#diff-e493f9c2deceaafd80b881cfabc92186R631)

packages/ember-glimmer/lib/components/curly-component.js
```diff
@@ -25,17 +25,19 @@ function attrsToProps(keys, attrs) {
...
     merged[name] = value;
   }
 
   return merged;
 }
 
+class ComponentStateBucket {
+  constructor(component) {
+    this.component = component;
+    this.classRef = undefined;
```
chancancode: We usually use `null` for "there is explicitly nothing here"

packages/ember-glimmer/tests/integration/components/curly-components-test.js
```diff
@@ -199,6 +215,24 @@ moduleFor('Components test: curly components', class extends RenderingTest {
...
+
+    this.render('{{foo-bar class=(if fooBar "foo-bar")}}', {
+      fooBar: true
+    });
+
+    this.assertComponentElement(this.firstChild, { content: 'hello', attrs: { 'class': classes('ember-view foo-bar') } });
+
+    this.runTask(() => this.rerender());
+
+    this.assertComponentElement(this.firstChild, { content: 'hello', attrs: { 'class': classes('foo-bar ember-view') } });
```
chancancode: The difference in order is not significant right? (`foo-bar ember-view` vs `ember-view foo-bar`) Would prefer them to be consistent in the test code.

packages/ember-glimmer/tests/integration/components/curly-components-test.js
```diff
@@ -199,6 +215,24 @@ moduleFor('Components test: curly components', class extends RenderingTest {
...
+
+    this.runTask(() => set(this.context, 'fooBar', false));
+
+    this.assertComponentElement(this.firstChild, { content: 'hello', attrs: { 'class': classes('ember-view') } });
```
chancancode: Add reset step (set `fooBar` to `true`)

packages/ember-glimmer/tests/integration/content-test.js
```diff
@@ -584,4 +585,236 @@ moduleFor('Dynamic content tests (integration)', class extends RenderingTest {
...
+
+    this.runTask(() => this.rerender());
+
+    this.assertElement(this.firstChild, { tagName: 'div', content: 'hello', attrs: { 'class': classes('foo-bar') } });
+
+    this.runTask(() => set(this.context, 'fooBar', false));
+
+    // HTMLBars differs in behavior here as it leaves the empty
+    // class attribute.
+    assert.equal(this.firstChild.className, '');
```
chancancode: Eh, I think this is actually "correct". I guess we should have checked for only `!== null && !== undefined` in Glimmer after all.

packages/ember-glimmer/tests/integration/content-test.js
```diff
@@ -584,4 +585,236 @@ moduleFor('Dynamic content tests (integration)', class extends RenderingTest {
...
+
+    this.runTask(() => set(this.context, 'fooBar', true));
+
+    this.assertElement(this.firstChild, { tagName: 'div', content: 'hello', attrs: { 'class': classes('foo-bar') } });
+  }
+
+    ['@test unquoted class property can contain multiple classes']() {
```
chancancode: indentation

packages/ember-glimmer/tests/integration/content-test.js
chancancode: (I don't think we need to block this PR for that, but we should probably go back to glimmer and fix it)

packages/ember-glimmer/tests/integration/content-test.js
```diff
@@ -584,4 +585,236 @@ moduleFor('Dynamic content tests (integration)', class extends RenderingTest {
...
+    this.runTask(() => set(this.context, 'model.classes', 'fizz bizz'));
+
+    this.assertElement(this.firstChild, { tagName: 'div', content: 'hello', attrs: { 'class': classes('fizz bizz') } });
+
+    this.runTask(() => set(this.context, 'model', { classes: 'foo bar baz' }));
+
+    this.assertElement(this.firstChild, { tagName: 'div', content: 'hello', attrs: { 'class': classes('foo bar baz') } });
+  }
+
+  ['@test unquoted class property']() {
```
chancancode: property -> attribute

packages/ember-glimmer/tests/integration/content-test.js
```diff
@@ -584,4 +585,236 @@ moduleFor('Dynamic content tests (integration)', class extends RenderingTest {
...
+    this.runTask(() => set(this.context, 'model.foo', 'fizz'));
+
+    this.assertElement(this.firstChild, { tagName: 'div', content: 'hello', attrs: { 'class': classes('fizz') } });
+
+    this.runTask(() => set(this.context, 'model', { foo: 'foo' }));
+
+    this.assertElement(this.firstChild, { tagName: 'div', content: 'hello', attrs: { 'class': classes('foo') } });
+  }
+
+  ['@test quoted class property']() {
```
chancancode: property -> attribute

packages/ember-glimmer/tests/integration/content-test.js
```diff
@@ -584,4 +585,236 @@ moduleFor('Dynamic content tests (integration)', class extends RenderingTest {
...
+    this.runTask(() => set(this.context, 'model.foo', 'fizz'));
+
+    this.assertElement(this.firstChild, { tagName: 'div', content: 'hello', attrs: { 'class': classes('fizz') } });
+
+    this.runTask(() => set(this.context, 'model', { foo: 'foo' }));
+
+    this.assertElement(this.firstChild, { tagName: 'div', content: 'hello', attrs: { 'class': classes('foo') } });
+  }
+
+  ['@test quoted class property can contain multiple classes']() {
```
chancancode: property -> attribute
 Add a line note

packages/ember-glimmer/tests/integration/components/curly-components-test.js
rwjblue: I thought order did matter for CSS rules?  I could be completely wrong though, feel free to disregard if I am...

rwjblue: Although, IIRC, we are sorting in the test helper to do a stable comparison.  Maybe this is the result of that?  If so, we should probably test the real element order too somehow...

chancancode: The order doesn't matter: http://stackoverflow.com/questions/1321692/how-to-specifiy-the-order-of-css-classes

Also, yep, the `classes` helper sorts them anyway; which is why I suggested changing it to be consistent to not mislead other people in thinking that there is any significance between the two lines

packages/ember-glimmer/tests/integration/content-test.js
```diff
@@ -584,4 +585,236 @@ moduleFor('Dynamic content tests (integration)', class extends RenderingTest {
...
+    this.assertElement(this.firstChild, { tagName: 'div', content: 'hello', attrs: { 'class': classes('foo bar bizz') } });
+
+
+    this.runTask(() => set(this.context, 'model.foo', 'fizz'));
+
+    this.assertElement(this.firstChild, { tagName: 'div', content: 'hello', attrs: { 'class': classes('fizz bar bizz') } });
+
+    this.runTask(() => set(this.context, 'model.bar', null));
+
+    this.assertElement(this.firstChild, { tagName: 'div', content: 'hello', attrs: { 'class': classes('fizz  bizz') } });
```
chancancode: extra space between `fizz  buzz` is not significat right? ✂️

chancancode: I see what's going on now. The spirit of the `classes` helper is to ignore spaces and ordering – but we currently have a bug that sometimes give you extra `""`. I think we should fix the helper to split with a whitespace regex (instead of a single space character) and not worry about it here. If for some reason we do want to test the actual output (I don't think we want to), character by character, we can always just pass a string here to do an exact match instead of using the `classes` helper.

packages/ember-glimmer/tests/integration/content-test.js
```diff
@@ -584,4 +585,236 @@ moduleFor('Dynamic content tests (integration)', class extends RenderingTest {
...
+    this.render(`<div class="{{if model.hasSize model.size}} {{if model.hasShape model.shape}}">hello</div>`, {
+      model: {
+        size: 'large',
+        hasSize: true,
+        hasShape: false,
+        shape: 'round'
+      }
+    });
+
+    this.assertElement(this.firstChild, { tagName: 'div', content: 'hello', attrs: { 'class': classes('large ') } });
```
chancancode: see https://github.com/emberjs/ember.js/pull/13372/files?diff=unified#r60633975

packages/ember-glimmer/tests/integration/content-test.js
```diff
@@ -584,4 +585,236 @@ moduleFor('Dynamic content tests (integration)', class extends RenderingTest {
...
+
+    this.runTask(() => this.rerender());
+
+    this.assertElement(this.firstChild, { tagName: 'div', content: 'hello', attrs: { 'class': classes('large ') } });
```
chancancode: see https://github.com/emberjs/ember.js/pull/13372/files?diff=unified#r60633975

packages/ember-glimmer/tests/integration/content-test.js
```diff
@@ -584,4 +585,236 @@ moduleFor('Dynamic content tests (integration)', class extends RenderingTest {
...
+
+    this.runTask(() => set(this.context, 'model.hasShape', true));
+
+    this.assertElement(this.firstChild, { tagName: 'div', content: 'hello', attrs: { 'class': classes('large round') } });
+
+    this.runTask(() => set(this.context, 'model.hasSize', false));
+
+    this.assertElement(this.firstChild, { tagName: 'div', content: 'hello', attrs: { 'class': classes(' round') } });
```
chancancode: see https://github.com/emberjs/ember.js/pull/13372/files?diff=unified#r60633975

packages/ember-glimmer/tests/integration/content-test.js
```diff
@@ -584,4 +585,236 @@ moduleFor('Dynamic content tests (integration)', class extends RenderingTest {
...
+
+    this.runTask(() => set(this.context, 'model', {
+      size: 'large',
+      hasSize: true,
+      hasShape: false,
+      shape: 'round'
+    }));
+
+    this.assertElement(this.firstChild, { tagName: 'div', content: 'hello', attrs: { 'class': classes('large ') } });
```
chancancode: see https://github.com/emberjs/ember.js/pull/13372/files?diff=unified#r60633975

packages/ember-glimmer/tests/integration/content-test.js
```diff
@@ -584,4 +585,236 @@ moduleFor('Dynamic content tests (integration)', class extends RenderingTest {
...
+    this.runTask(() => this.rerender());
+
+    this.assertElement(this.firstChild, { tagName: 'div', content: 'hello', attrs: { 'class': classes('foo bar fizz baz') } });
+
+    this.runTask(() => {
+      set(this.context, 'model.foo', null);
+      set(this.context, 'model.fizz', null);
+    });
+
+    this.assertElement(this.firstChild, { tagName: 'div', content: 'hello', attrs: { 'class': classes(' bar  baz') } });
```
chancancode: see https://github.com/emberjs/ember.js/pull/13372/files?diff=unified#r60633975

packages/ember-htmlbars/tests/attr_nodes/class_test.js
chancancode: ✔️

packages/ember-htmlbars/tests/attr_nodes/class_test.js
chancancode: ✔️

packages/ember-htmlbars/tests/attr_nodes/class_test.js
chancancode: ✔️

packages/ember-htmlbars/tests/attr_nodes/class_test.js
chancancode: ✔️

packages/ember-htmlbars/tests/attr_nodes/class_test.js
chancancode: ✔️

packages/ember-htmlbars/tests/attr_nodes/class_test.js
chancancode: ✔️

packages/ember-htmlbars/tests/attr_nodes/class_test.js
chancancode: ✔️

packages/ember-htmlbars/tests/attr_nodes/class_test.js
chancancode: ✔️

packages/ember-htmlbars/tests/attr_nodes/class_test.js
chancancode: ✔️

packages/ember-htmlbars/tests/attr_nodes/class_test.js
chancancode: Hm, the original test explicitly claim to care about ordering – and the new test does not (via the `classes` helper which sorts them). Like I mentioned above, I don't think we have a reason to care, but if for some reason we have I think the test will still pass (you would use a string instead of `classes` which does an exact match). @rwjblue can we drop this?

chadhietala: [Added a class for this](https://github.com/emberjs/ember.js/pull/13372/files#diff-e493f9c2deceaafd80b881cfabc92186R821)

PR #13374, opened 2016-04-19
[FEATURE] Add support for positional parameters in the component blueprint

orf:
Hello,
This is a feature I've been feeling is lacking. I often make components that take positional parameters, and I think a flag on the generator would be a good addition. This PR adds a `--positional-parameters` (aliased to `-p`) flag that makes a component like so:

``` javascript
import ...;

const Component = Ember.Component.extend({...});

Component.reopenClass(({
  positionalParams: []
});

export default Component;
```

I'm not sure where the tests are for the generators, I will write one if someone points me in the right direction. I've made this PR to kick off the CI as my local tests seem to have a lot of (hopefully) unrelated failures and to get feedback - this is my first PR so I'm sure there are problems!

Also I'd like to enable people to specify a list of positional parameters through the command line, like `-p arg1,arg2,arg3` but I'm not sure how to add this.

### DIFF
--- a/blueprints/component/files/__root__/__path__/__name__.js
+++ b/blueprints/component/files/__root__/__path__/__name__.js
@@ -1,4 +1,12 @@
 import Ember from 'ember';
 <%= importTemplate %>
-export default Ember.Component.extend({<%= contents %>
+<% if (posParams) { %>const Component = <% } else { %> export default <% } %>Ember.Component.extend({<%= contents %>
 });
+
+<% if (posParams) { %>
+Component.reopenClass({
+  positionalParams: []
+});
+
+export default Component;
+<% } %>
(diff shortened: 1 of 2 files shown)


rwjblue:
In general, I think most API's should not require `positionalParams`. When we re-implement angle bracket components they will not be able to use postional params at all, so it is likely best for us to keep them more as a "power feature" when you really need to take an ordered argument...

I'd prefer to keep the blueprint as is...

orf:
Understandable. Perhaps it's a personal style choice but I find I often add positionalParams to smaller components, mostly to point at the model attribute.

Feel free to close this then if you'd rather keep the blueprint as-is :)

rwjblue:
Thank you very much for putting the time into this. I am going to go ahead and close this though...

PR #13376, opened 2016-04-19
[Glimmer2] Add compile-time assert for block usage of custom helper

Joelkang:
This throws an error during compile time when custom helpers are used in a block form

### DIFF
--- a/packages/ember-glimmer/lib/environment.js
+++ b/packages/ember-glimmer/lib/environment.js
@@ -74,11 +74,7 @@ export default class Environment extends GlimmerEnvironment {
     }
 
     let nativeSyntax = super.refineStatement(statement);
-
-    if (!nativeSyntax && key && this.hasHelper(key)) {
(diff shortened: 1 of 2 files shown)


packages/ember-glimmer/lib/environment.js
```diff
@@ -72,7 +73,13 @@ export default class Environment extends GlimmerEnvironment {
       }
     }
 
-    return super.refineStatement(statement);
+    let nativeSyntax = super.refineStatement(statement);
+
+    if (!nativeSyntax && key && this.hasHelper(key)) {
```
chadhietala: So I think we want to actually pass all of this as the second argument to the `assert` if possible. The way that assert works is that it is only there in dev build. The production build strips the `assert`, so in the prod build we actually would have something like this:

```
if (!nativeSyntax && key && this.hasHelper(key)) {}
```

You may want to do an IIFE in the assertion to make it a little bit easer to read.

Joelkang: yea I thought about that, but textStatements (errthing false/undefined) were also being passed into `refineStatement()`, which was problematic (especially when `key` is undefined). Anyway I updated the logic and tests pass

chadhietala:
Thanks! This seems correct to me. Paging @chancancode for 👍 and merge.

PR #13378, opened 2016-04-19
[Glimmer2] Implement glimmer local lookup for components

asakusuma:
This PR requires this [glimmer template meta PR](https://github.com/tildeio/glimmer/pull/141) before the tests will pass.

cc @rwjblue

### DIFF
--- a/packages/ember-glimmer/lib/environment.js
+++ b/packages/ember-glimmer/lib/environment.js
@@ -158,7 +158,7 @@ export default class Environment extends GlimmerEnvironment {
    // isn't going to return any syntax and the Glimmer engine knows how to handle
    // this case.
 
-  refineStatement(statement) {
+  refineStatement(statement, symbolTable) {
     // 1. resolve any native syntax – if, unless, with, each, and partial
     let nativeSyntax = super.refineStatement(statement);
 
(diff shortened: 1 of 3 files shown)


packages/ember-glimmer/lib/environment.js
```diff
@@ -79,14 +79,20 @@ export default class Environment extends GlimmerEnvironment {
...
+      let { component: ComponentClass, layout } = lookupComponent(this.owner, name);
 
       if (ComponentClass || layout) {
-        definition = this._components[name] = new CurlyComponentDefinition(name, ComponentClass, layout);
+        definition = this._components[name] = new CurlyComponentDefinition(nameSegments, ComponentClass, layout);
+      } else if (meta && meta.moduleName) {
+        let segments = meta.moduleName.split('/')
+        segments.shift();
+        segments.push(name);
+        return this.getComponentDefinition(segments);
```
rwjblue: This is slightly confusing to me, can you explain the idea here?

packages/ember-glimmer/lib/environment.js
```diff
@@ -79,14 +79,20 @@ export default class Environment extends GlimmerEnvironment {
...
   }
 
-  getComponentDefinition(name) {
+  getComponentDefinition(nameSegments, meta) {
+    let name = nameSegments.join('/');
     let definition = this._components[name];
 
     if (!definition) {
-      let { component: ComponentClass, layout } = lookupComponent(this.owner, name[0]);
+      let { component: ComponentClass, layout } = lookupComponent(this.owner, name);
```
rwjblue: To support local lookup we need to pass a third arg to `lookupComponent` here ([see here](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/utils/lookup-component.js#L27)). I think that arg would be `{ source: meta && meta.moduleName}`.  Then that helper function looks for a local result first, and falls back to global.  I'm not sure why we need `else if` below if we pass the third arg to `lookupComponent`.

packages/ember-glimmer/lib/environment.js
asakusuma: moduleName will be somethings like `compoment/parent-foo` and `nameSegments` will be `['my-foo']`. This code creates `segments`, which ends up being `['parent-foo', 'my-foo']` and becomes the `nameSegments` of the recursive call.

packages/ember-glimmer/lib/environment.js
asakusuma: Ahh, I think I need to get rid of the code you commented on below and instead use this third parameter.

rwjblue: 👍 - The idea is that `lookupComponent` is the only thing that has to care about the fallback stuff.

asakusuma:
Closed in favor of https://github.com/emberjs/ember.js/pull/13378

stefanpenner:
@asakusuma such :trollface: that link links to this PR, i clicked it 3 times :P

asakusuma:
@stefanpenner you know how I do. Closed in favor of https://github.com/emberjs/ember.js/pull/13778

PR #13385, opened 2016-04-20
[DOC] Update component hook documentation for FastBoot

danmcclain:


stefanpenner:
Should we introduce fastboot mentions in ember API docs? Doesn't seem quite correct. Is there something more "neutral" that can be said?

rwjblue:
Ember supports the primary feature that drives fastboot (the visit API) directly.  I think the key isn't about fastboot or visit though, I believe it is all based on a `hasDOM` check.  If we can confirm that, then we can mention that instead of fastboot.

stefanpenner:
@rwjblue ya, something not coupled to fast-boot. maybe some common vernacular they could both share?

cc @tomdale

tomdale:
This is a little tricky because it's weird to mention an addon in the Ember docs, but in practice people may end up going to the Ember docs to learn about which hooks are invoked or not in FastBoot, and I don't want to put ideology ahead of pragmatism.

I think `hasDOM` is the right way to think about it, but it's not something end users will likely be thinking of. Maybe we can say something as simple "Not invoked in Node.js environments"?

pixelhandler:
@tomdale cc @danmcclain I think it would be better to document how FastBoot utilizes the component hooks in the FastBoot docs instead of in the Ember docs.

tomdale:
@pixelhandler Keep in mind that the ability to run Ember apps in Node comes from the Ember framework itself, not FastBoot, via the `visit()` API. There will likely be other ways to run Ember apps in the future where these hooks don't get invoked.

locks:
We'll likely move these efforts to https://github.com/emberjs/guides/pull/1982.

PR #13390, opened 2016-04-21
computed.and/or: Short-circuit behavior

### LINKED ISSUE #13384: computed.and/or: Short-circuit behavior
Example: 

```js 
{
  a: Ember.computed(function() {
    console.log('a');
    return false;
  }),

  b: Ember.computed(function() {
    console.log('b');
    return 'b';
  }),

  c: Ember.computed(function() {
    console.log('c');
    return 'c';
  }),

  foo: Ember.computed.or('a', 'b', 'c'),
}
```

When I use `{{foo}}` in the corresponding template I correctly see `b` as a value. I did however see `a`, `b` and `c` get printed on the console, which indicates that the current implementation of the `Ember.computed.and/or` helpers are not using short-circuit behavior. In other words: since `b` returns something trueish in this example `c` shouldn't have been evaluated. The performance in this example is not that relevant but once the computed properties become more complex it might pay off to only evaluate the properties if they are really needed.

It looks like the current implementation is just calling `.getProperties()` on the object which leads to the current behavior. If that was replaced with a loop calling `.get()` for each property until one fulfills the "and/or" predicate the performance might be improved for complex properties.

/cc @rwjblue

Turbo87:
This PR resolves #13384 

/cc @rwjblue @stefanpenner

### DIFF
--- a/packages/ember-metal/lib/computed_macros.js
+++ b/packages/ember-metal/lib/computed_macros.js
@@ -20,20 +20,27 @@ function getProperties(self, propertyNames) {
   return ret;
 }
 
+function expandPropertiesToArray(properties) {
+  var expandedProperties = [];
+
+  function extractProperty(entry) {
+    expandedProperties.push(entry);
+  }
+
+  for (let i = 0; i < properties.length; i++) {
+    expandProperties(properties[i], extractProperty);
+  }
+
+  return expandedProperties;
+}
+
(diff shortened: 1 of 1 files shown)


stefanpenner:
lgtm, lets let this ride canary and see if people rely on the old (wrong) behavior.

rwjblue:
Awesome, thanks @Turbo87!

PR #13222, opened 2016-03-31
[Glimmer2] Port <input> tests to ember-glimmer

paddyobrien:
Relates to https://github.com/emberjs/ember.js/issues/13127

Port over input element tests. There are two cases that fail in glimmer which I've marked as `@htmlbars`.

Both failures look to be caused by missing camelization. i.e. `tabindex` is the html attribute, but the `HTMLInputElement` property is `tabIndex`

PR #13392, opened 2016-04-21
[Glimmer2] Cleanup input tests

chadhietala:
This will supersede #13222


packages/ember-glimmer/tests/integration/input-test.js
```diff
@@ -0,0 +1,159 @@
...
+    this.assertAttributeHasValue(attributeName, values[0], `${attributeName} is set on noop rerender`);
+
+    this.setComponentValue(values[1]);
+    this.assertAttributeHasValue(attributeName, values[1], `${attributeName} is set on rerender`);
+
+    this.setComponentValue(values[0]);
+    this.assertAttributeHasValue(attributeName, values[0], `${attributeName} can be set back to the initial value`);
+  }
+
+  ['@test input disabled attribute']() {
```
stefanpenner: why do we wrap these in brackets? leaving them off is valid es6 syntax i think.

packages/ember-glimmer/tests/integration/input-test.js
```diff
@@ -0,0 +1,155 @@
+import { RenderingTest, moduleFor } from '../utils/test-case';
+import { set } from 'ember-metal/property_set';
+
+moduleFor('Input element tests', class extends RenderingTest {
+  runAttributeTest(attributeName, values) {
+    let template = `<input ${attributeName}={{value}}>`;
```
mmun: Did you intentionally not quote or could it also have been `="{{value}}"` instead?

packages/ember-glimmer/tests/integration/input-test.js
```diff
@@ -0,0 +1,155 @@
...
+
+  ['@test input disabled attribute']() {
+    this.runPropertyTest('disabled', [false, true]);
+  }
+
+  ['@test input value attribute']() {
+    this.runPropertyTest('value', ['foo', 'bar']);
+  }
+
+  ['@htmlbars input placeholder attribute']() {
```
mmun: What is preventing this from working in glimmer?

packages/ember-glimmer/tests/integration/input-test.js
```diff
@@ -0,0 +1,155 @@
...
+
+  ['@htmlbars input maxlength attribute']() {
+    this.runAttributeTest('maxlength', [2, 3]);
+  }
+
+  ['@test input size attribute']() {
+    this.runAttributeTest('size', [2, 3]);
+  }
+
+  ['@htmlbars input tabindex attribute']() {
```
mmun: Same as https://github.com/emberjs/ember.js/pull/13392/files#r60971131.

packages/ember-glimmer/tests/integration/input-test.js
```diff
@@ -0,0 +1,155 @@
...
+
+  ['@htmlbars input placeholder attribute']() {
+    this.runAttributeTest('placeholder', ['foo', 'bar']);
+  }
+
+  ['@test input name attribute']() {
+    this.runAttributeTest('name', ['nam', 'name']);
+  }
+
+  ['@htmlbars input maxlength attribute']() {
```
mmun: Same as https://github.com/emberjs/ember.js/pull/13392/files#r60971131.

packages/ember-glimmer/tests/integration/input-test.js
```diff
@@ -0,0 +1,155 @@
...
+    this.setDOMValue('hello');
+    this.setSelectionRange(1, 3);
+
+    this.setComponentValue('hello');
+
+    this.assertSelectionRange(1, 3);
+
+    // Note: We should eventually get around to testing reseting, however
+    // browsers handle `selectionStart` and `selectionEnd` differently
+    // when are synthetically testing movement of the cursor.
```
mmun: :+1:

packages/ember-glimmer/tests/integration/input-test.js
chadhietala: I'm actually not sure, in IE9 this fails and I think it has to do with attributes vs. properties.

packages/ember-glimmer/tests/integration/input-test.js
```diff
@@ -0,0 +1,155 @@
...
+
+    this.assertValue('hello', 'Value is initialised');
+
+    this.setDOMValue('hola');
+
+    this.assertValue('hola', 'DOM is used');
+
+    this.setComponentValue('hello');
+
+    this.assertValue('hola', 'DOM is used');
```
mmun: This is not correct. It should be `hello`. While it is true that the DOM is the SSOT, this doesn't mean that mutation via `setComponentValue` cannot mutate the SSOT.

packages/ember-glimmer/tests/integration/input-test.js
chadhietala: We could do both 😄 . Just trying to get this carried over the line.

packages/ember-glimmer/tests/integration/input-test.js
mmun: This should pass if you make it @htmlbars for now.

packages/ember-glimmer/tests/integration/input-test.js
mmun: I think it's fine as is.

packages/ember-glimmer/tests/integration/input-test.js
chadhietala: `@skip` for now.

Issue #13158, opened 2016-03-21
Component positional params don't update when used with the component helper


GavinJoyce:
I came across this when [porting component helper tests to Glimmer2](https://github.com/emberjs/ember.js/pull/13140).

Twiddle: https://ember-twiddle.com/8bd17b5f64fdf1f400d8

When this is fixed, please uncomment https://github.com/emberjs/ember.js/pull/13140/files#r56916691

rwjblue:
This works properly in the glimmer implementation but not HTMLBars or it doesn't work in either case?

GavinJoyce:
I don't believe positional params have been implemented in glimmer yet

GavinJoyce:
From https://github.com/emberjs/ember.js/issues/13127:

<img width="656" alt="screen shot 2016-03-21 at 19 42 23" src="https://cloud.githubusercontent.com/assets/2526/13940706/2aa348fc-ef9d-11e5-821a-c6c193775cd8.png">

pixelhandler:
@GavinJoyce I'm curious in the twiddle example regarding using `positionalParams` in two separate component that both use `Ember.Component.reopenClass` to do the same thing. Is that how you would used positional params in an app? or would you extend the component then reopenClass ?

GavinJoyce:
@pixelhandler thanks, my mistake. I've updated the twiddle, the bug is still present. I'll update the tests in ember.js too, the currently `Ember.Component.reopenClass`.

locks:
@Serabe can you review?

Serabe:
Yes. A [quick test](https://ember-twiddle.com/6752a23b8899ad87c2e2106be0b26f5d?numColumns=2&openFiles=components.my-component.js%2Ctemplates.components.my-component.hbs) shows that for rest pos params it only updates for the first one (😲). I'm working on other parts of positional parameters, so I'll take a look while at it.

Serabe:
#13435 fixes this problem in HTMLBars. Glimmer 2 support for positional parameters is on its way (#13393) and won't need a fix.

PR #13393, opened 2016-04-21
[Glimmer2] Implement positional parameter

Serabe:
This PR implements positional parameters in Glimmer 2.
- [x] Basic processing of positional parameters
- [x] Query parameters in dynamic components (`{{component "x-component" "a" "b"`}}`)
- [x] Improve processing based on Glimmer 2 positional args API.
- [x] Pass all tests that currently pass in `@htmlbars`.
- [x] If `id` arg is implemented before closing this PR, change tests back to `id="x"`.

### DIFF
--- a/packages/ember-glimmer/lib/components/curly-component.js
+++ b/packages/ember-glimmer/lib/components/curly-component.js
@@ -1,8 +1,8 @@
 import { StatementSyntax, ValueReference } from 'glimmer-runtime';
 import { AttributeBindingReference, RootReference, applyClassNameBinding } from '../utils/references';
 import { DIRTY_TAG } from '../ember-views/component';
-import EmptyObject from 'ember-metal/empty_object';
 import { assert } from 'ember-metal/debug';
+import processArgs from '../utils/process-args';
 
 export class CurlyComponentSyntax extends StatementSyntax {
   constructor({ args, definition, templates }) {
@@ -18,21 +18,6 @@ export class CurlyComponentSyntax extends StatementSyntax {
   }
 }
 
-function attrsToProps(keys, attrs) {
-  let merged = new EmptyObject();
-
-  merged.attrs = attrs;
-
-  for (let i = 0, l = keys.length; i < l; i++) {
-    let name = keys[i];
-    let value = attrs[name];
-
-    merged[name] = value;
-  }
-
-  return merged;
-}
-
 class ComponentStateBucket {
   constructor(component, args) {
     this.component = component;
@@ -46,8 +31,7 @@ class CurlyComponentManager {
     let parentView = dynamicScope.view;
 
     let klass = definition.ComponentClass;
-    let attrs = args.named.value();
-    let props = attrsToProps(args.named.keys, attrs);
+    let { attrs, props } = processArgs(args, klass.positionalParams);
 
     props.renderer = parentView.renderer;
 
@@ -135,8 +119,7 @@ class CurlyComponentManager {
     if (!args.tag.validate(argsRevision)) {
       bucket.argsRevision = args.tag.value();
 
-      let attrs = args.named.value();
-      let props = attrsToProps(args.named.keys, attrs);
+      let { attrs, props } = processArgs(args, component.constructor.positionalParams);
 
       let oldAttrs = component.attrs;
       let newAttrs = attrs;
--- a/packages/ember-glimmer/lib/utils/process-args.js
+++ b/packages/ember-glimmer/lib/utils/process-args.js
@@ -1,0 +1,95 @@
+import { assert } from 'ember-metal/debug';
+import assign from 'ember-metal/assign';
+import EmptyObject from 'ember-metal/empty_object';
+
+export default function processArgs(args, positionalParamsDefinition) {
+  let attrs = assign(new EmptyObject(), args.named.value());
+  processPositionalParams(positionalParamsDefinition || [], args.positional, attrs);
+
+  let props = attrsToProps(keysForAttrs(args, positionalParamsDefinition), attrs);
+
+  return { attrs, props };
+}
+
+// Transforms an object of external attributes to internal properties.
+// This means that each key-value pair needs to be accessible both as key and
+// attrs.key
+function attrsToProps(keys, attrs) {
+  let merged = new EmptyObject();
+
+  merged.attrs = attrs;
+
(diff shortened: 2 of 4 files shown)


chancancode:
@Serabe I pushed a change in Glimmer for the component helper case in https://github.com/tildeio/glimmer/commit/2b660889f3fc76eeb4f473c9309d0eb3a2501f14

The corresponding Ember change would be something like this (basically what I did in the Glimmer test):

```diff
diff --git a/packages/ember-glimmer/lib/components/dynamic-component.js b/packages/ember-glimmer/lib/components/dynamic-component.js
index 58dd886..e8a8fa0 100644
--- a/packages/ember-glimmer/lib/components/dynamic-component.js
+++ b/packages/ember-glimmer/lib/components/dynamic-component.js
@@ -1,10 +1,13 @@
-import { StatementSyntax } from 'glimmer-runtime';
+import { ArgsSyntax, StatementSyntax } from 'glimmer-runtime';

 export class DynamicComponentSyntax extends StatementSyntax {
   constructor({ args, templates }) {
     super();
-    this.args = args;
-    this.definition = dynamicComponentFor;
+    this.definition = {
+      args: ArgsSyntax.fromPositionalArgs(args.positional.slice(0, 1)),
+      factory: dynamicComponentFor
+    };
+    this.args = ArgsSyntax.build(args.positional.slice(1), args.named);
     this.templates = templates;
     this.shadow = null;
   }
diff --git a/packages/ember-glimmer/lib/components/outlet.js b/packages/ember-glimmer/lib/components/outlet.js
index 784237a..a07020f 100644
--- a/packages/ember-glimmer/lib/components/outlet.js
+++ b/packages/ember-glimmer/lib/components/outlet.js
@@ -1,4 +1,4 @@
-import { StatementSyntax } from 'glimmer-runtime';
+import { ArgsSyntax, StatementSyntax } from 'glimmer-runtime';
 import { ConstReference } from 'glimmer-reference';
 import { generateGuid, guidFor } from 'ember-metal/utils';
 import { RootReference, NULL_REFERENCE } from '../utils/references';
@@ -6,8 +6,8 @@ import { RootReference, NULL_REFERENCE } from '../utils/references';
 export class OutletSyntax extends StatementSyntax {
   constructor({ args }) {
     super();
-    this.args = args;
-    this.definition = outletComponentFor;
+    this.definition = { args, factory: outletComponentFor };
+    this.args = ArgsSyntax.empty();
     this.templates = null;
     this.shadow = null;
   }
```

chancancode:
re: above – should just work now with https://github.com/emberjs/ember.js/pull/13405 merged

Serabe:
There are still tests not passing for #13158. I'm trying to look at what the problem might be. Since I'll need it for the contextual parameters part as well, what is the equivalent point to `ember-htmlbars/lib/keywors/component.js` in Glimmer?

chancancode:
It'd be in the dynamic component file.

chadhietala:
@Serabe Do you have time to take this over line? Otherwise I can take it on.

Serabe:
Yep. I'll try to do so this weekend.

Serabe:
I had to change a couple of things. I had to do the operation in two steps, since passing `props` to an action makes `JSON.stringify` fail because it detects a circular reference, not so sure why.

I also separated a couple of dynamic component test that were mixed in other tests in curly components.

packages/ember-glimmer/lib/components/curly-component.js
```diff
@@ -32,6 +41,30 @@ function attrsToProps(keys, attrs) {
...
+  let posKeys = keysForPositionalParameters(args, posParamsDefinition);
+
+  return namedKeys.concat(posKeys);
+}
+
+function keysForPositionalParameters(args, posParamsDefinition) {
+  let numberOfPosParams = args.positional.length;
+  if (numberOfPosParams === 0) {
+    // If there are no positional parameters passed, returned an empty array
+    return [];
```
chadhietala: Maybe just return `args.positional` here instead of allocating the new array?

chadhietala:
Just the small suggestion. Otherwise this LGTM.

packages/ember-glimmer/lib/components/curly-component.js
Serabe: `args.positional` are the values of the positional parameters but we are looking for an array of keys to pass to `attrsToProps`.

packages/ember-glimmer/lib/components/curly-component.js
```diff
@@ -45,8 +78,8 @@ class CurlyComponentManager {
     let parentView = dynamicScope.view;
 
     let klass = definition.ComponentClass;
-    let attrs = args.named.value();
-    let props = attrsToProps(args.named.keys, attrs);
+    let attrs = processAttrs(args, klass.positionalParams);
+    let props = attrsToProps(keysForAttrs(args, klass.positionalParams), attrs);
```
chancancode: `processAttrs` and `attrsToProps` have basically lost their meaning at this point 😨

Can you extract them into your new file in utils and make it something like `let { attrs, props } = processArgs(args, klass.positionalParams);`?

`attrs` is the thing we pass to the `did*Attrs` hooks, and `props` is what we set on the component instance.

I don't think there is necessarily value of keeping them as separate functions anymore, and passing keys etc. The reason it was factored that way was largely an evolution of the requirements so far, and the new requirements you have introduced here have change the way they work significantly.

I think it could have been much better factored if you weren't subject to these random historical constraints. :smile:

packages/ember-glimmer/lib/components/curly-component.js
```diff
@@ -119,8 +152,8 @@ class CurlyComponentManager {
     if (!args.tag.validate(argsRevision)) {
       bucket.argsRevision = args.tag.value();
 
-      let attrs = args.named.value();
-      let props = attrsToProps(args.named.keys, attrs);
+      let attrs = processAttrs(args, component.constructor.positionalParams);
+      let props = attrsToProps(keysForAttrs(args, component.constructor.positionalParams), attrs);
```
chancancode: ditto

packages/ember-glimmer/lib/components/curly-component.js
Serabe: Thank you. Updated!

chancancode:
Merged in #13940, thanks @serabe 👍👍👍

PR #13394, opened 2016-04-22
[Glimmer2] hasBlock support

Joelkang:
Added tests for the variety of ways `has-block` can be used in glimmer

### DIFF
--- a/packages/ember-glimmer/lib/ember-template-compiler/plugins/transform-has-block-syntax.js
+++ b/packages/ember-glimmer/lib/ember-template-compiler/plugins/transform-has-block-syntax.js
@@ -27,6 +27,8 @@ export default function TransformHasBlockSyntax() {
 
 const OLD_HAS_BLOCK = 'hasBlock';
 const NEW_HAS_BLOCK = 'has-block';
+const OLD_HAS_BLOCK_PARAMS = 'hasBlockParams';
+const NEW_HAS_BLOCK_PARAMS = 'has-block-params';
 
 /**
   @private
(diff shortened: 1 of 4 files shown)


packages/ember-glimmer/lib/ember-template-compiler/plugins/transform-has-block-syntax.js
```diff
@@ -0,0 +1,58 @@
...
+  @method transform
+  @param {AST} ast The AST to be transformed.
+*/
+TransformHasBlockSyntax.prototype.transform = function TransformHasBlockSyntax_transform(ast) {
+  let { traverse, builders: b } = this.syntax;
+
+  traverse(ast, {
+    PathExpression(node) {
+      if (node.original === OLD_HAS_BLOCK) {
+        return b.sexpr(b.path(NEW_HAS_BLOCK));
```
mmun: We should pass through the `loc` of the PathExpression into the SubExpression. Even though it is kind of hacky, it still represents the source code that ultimately generated the subexpression. Same applies below in the SubExpression visitor.

mmun:
@Joelkang Excellent!

packages/ember-glimmer/lib/ember-template-compiler/plugins/transform-has-block-syntax.js
Joelkang: Looks like buildSexpr and buildPath [don't actually support](https://github.com/tildeio/glimmer/blob/master/packages/glimmer-syntax/lib/builders.ts#L93) a `loc` parameter. I could add it to glimmer and then update this PR?

Joelkang:
Much thanks @mmun, replied to the comment; also let me know if hasBlockParams should go into a separate PR.

Incidentally, I'm not entirely sure what `{{hasBlockParams "inverse"}}` actually does (but I added tests for it and it transforms correctly)

packages/ember-glimmer/lib/ember-template-compiler/plugins/transform-has-block-syntax.js
mmun: Ah! Let's merge this as is and fix that separately then.

mmun:
@Joelkang `{{hasBlockParams "inverse"}}` is always false because it is currently not syntactically possible to use block params in an inverse block. Keeping the test around seems fine though.

Joelkang:
Fascinating, looks like htmlbars will return `1` or `0` when `{{hasBlockParams}}` is used in an attribute without a named block. `<button name={{hasBlockParams "default"}}` actually returns `<button name="true">` Marking as `@glimmer` for now

mmun:
Interesting. It must be that it returns the `arity` of the block rather than `!!arity`?

Joelkang:
[Looks to be the case](https://github.com/emberjs/ember.js/blob/master/packages/ember-htmlbars/lib/hooks/get-root.js#L13), but I'm not familiar enough with the HTMLBars code to say if we should fix this.

It'll probably be a regression but then again I'm hoping no one is actually relying on the output being 1 or 0 rather than a boolean; if it's being used in purely logical manners, `!!arity` probably won't break anyone. Not sure what the guidance is on something like that though, so I'll defer to you on whether we should fix it or leave it.

rwjblue:
It could be used to do different things based on the number of block params provided. I am not sure that it is terribly important to preserve (AFAIK this was not documented or publicized).

mmun:
:+1: `!!arity`.

chadhietala:
This LGTM.

PR #13338, opened 2016-04-14
[Glimmer2] Ensure falsy values do not become class names

asakusuma:

PR #13401, opened 2016-04-23
[Glimmer 2] Dasherize `someClass` to `some-class`

chadhietala:
This will dasherize `someClass` to `some-class` when a binding is set on
a component invocation for the class attribute.

### DIFF
--- a/packages/ember-glimmer/lib/environment.js
+++ b/packages/ember-glimmer/lib/environment.js
@@ -1,4 +1,7 @@
-import { Environment as GlimmerEnvironment } from 'glimmer-runtime';
+import {
+  Environment as GlimmerEnvironment,
+  HelperSyntax
+} from 'glimmer-runtime';
 import Dict from 'ember-metal/empty_object';
 import { assert } from 'ember-metal/debug';
 import { CurlyComponentSyntax, CurlyComponentDefinition } from './components/curly-component';
@@ -23,6 +26,7 @@ import { default as get } from './helpers/get';
 import { default as hash } from './helpers/hash';
 import { default as loc } from './helpers/loc';
 import { default as log } from './helpers/log';
+import { default as classHelper } from './helpers/-class';
 import { default as unbound } from './helpers/unbound';
 import { OWNER } from 'container/owner';
 
@@ -34,8 +38,24 @@ const builtInHelpers = {
   hash,
   loc,
   log,
-  unbound
+  unbound,
+  '-class': classHelper
 };
+
+function wrapClassAttribute(args) {
+  let hasClass = args.named.has('class');
+
+  if (hasClass) {
+    let { ref, type } = args.named.at('class');
+
+    if (type === 'get') {
+      let propName = ref.parts[ref.parts.length - 1];
+      let syntax = HelperSyntax.fromSpec(['helper', ['-class'], [['get', ref.parts], propName], null]);
+      args.named.add('class', syntax);
+    }
+  }
+  return args;
+}
 
 export default class Environment extends GlimmerEnvironment {
   static create(options) {
@@ -68,6 +88,7 @@ export default class Environment extends GlimmerEnvironment {
         let definition = this.getComponentDefinition(path);
 
         if (definition) {
+          wrapClassAttribute(args);
           return new CurlyComponentSyntax({ args, definition, templates });
         }
       }
--- a/packages/ember-glimmer/lib/helpers/-class.js
+++ b/packages/ember-glimmer/lib/helpers/-class.js
@@ -1,0 +1,25 @@
+import { InternalHelperReference } from '../utils/references';
+import { dasherize } from 'ember-runtime/system/string';
+
+function classHelper({ positional }) {
+  let path = positional.at(0);
+  let propName = positional.at(1);
+  let value = path.value();
+
+  if (value === true) {
+    return dasherize(propName.value());
+  }
+
+  if (value === false) {
+    return null;
+  }
+
+  return value;
+}
+
+export default {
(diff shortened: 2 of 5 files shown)


packages/ember-glimmer/lib/helpers/-class.js
```diff
@@ -0,0 +1,21 @@
+import { InternalHelperReference } from '../utils/references';
+import { dasherize } from 'ember-runtime/system/string';
+
+function classHelper({ positional }) {
```
chadhietala: Note: going to refactor this to not use the private `_propertyKey`

packages/ember-glimmer/lib/helpers/-class.js
```diff
@@ -0,0 +1,25 @@
...
+  let path = positional.at(0);
+  let className = positional.at(1);
+  let value = path.value();
+
+  if (value === true) {
+    return dasherize(className.value());
```
antonshmakov: Wouldn't it be better to use className as it is when explicitly passed? (Ideally to use dasherized attribute name by default for helpers as well, but feels like it is not possible?)

packages/ember-glimmer/lib/environment.js
```diff
@@ -68,7 +88,8 @@ export default class Environment extends GlimmerEnvironment {
         let definition = this.getComponentDefinition(path);
 
         if (definition) {
-          return new CurlyComponentSyntax({ args, definition, templates });
+          wrapClassAttribute(args);
```
antonshmakov: is there a way to achieve this in `args` setters/getters for `'class'`? It would be more scalable and, potentially, slightly more performant (won't call wrapper against every component)

packages/ember-glimmer/lib/environment.js
```diff
@@ -68,7 +88,8 @@ export default class Environment extends GlimmerEnvironment {
...
+          return new CurlyComponentSyntax({ args: args, definition, templates });
```
antonshmakov: ES6 shortcut for `args`?

packages/ember-glimmer/lib/environment.js
chadhietala: `args` is [EvaluatedArgs](https://github.com/tildeio/glimmer/blob/078009646b6838474aa2268f332721570153db8e/packages/glimmer-runtime/lib/compiled/expressions/args.ts#L32-L49) at this point and you do not want to special case any of this stuff in the rendering engine.

chadhietala: It also only wraps cases were the binding is used directly. For example `{{foo-bar class=(if fooBar "lol")}}` will not get wrapped. We are changing the syntax at compile time.

packages/ember-glimmer/lib/helpers/-class.js
chadhietala: The behavior today in Ember is if `{{foo-bar class=fooBar}}` is executed and `fooBar` evaluates to `true` we have to reflect on the binding e.g. `fooBar` and turn the binding name to `foo-bar`.

chadhietala:
This will supersede #13338

packages/ember-glimmer/lib/helpers/-class.js
antonshmakov: Yeah, I meant that for case `class=fooBar` it should transform into `foo-bar`, but in case `(-class fooBar "fooBar")`, when class passed explicitly by consumer, it should stay `fooBar` (to solve this you could pass dasherized version from the component and do not dasherize in the helper)

packages/ember-glimmer/lib/environment.js
```diff
@@ -34,9 +38,25 @@ const builtInHelpers = {
...
 };
 
+function wrapClassAttribute(args) {
+  let hasClass = args.named.has('class');
+
+  if (hasClass) {
+    let { ref, type } = args.named.at('class');
+
+    if (type === 'get') {
+      let className = ref.parts[ref.parts.length - 1];
```
krisselden: this should be propName since it isn't actually the class name at this point

packages/ember-glimmer/lib/helpers/-class.js
```diff
@@ -0,0 +1,25 @@
+import { InternalHelperReference } from '../utils/references';
+import { dasherize } from 'ember-runtime/system/string';
+
+function classHelper({ positional }) {
+  let path = positional.at(0);
+  let className = positional.at(1);
```
krisselden: again this is propName which is leading the the confusion below, the thing above is the className unless it is a boolean

packages/ember-glimmer/lib/helpers/-class.js
krisselden: the last arg is the name of the ref passed in, the ref is the className unless it is a boolean.

krisselden: another way to think of it is the last arg is metadata, essential reflection of the path that was bound.  the first argument is the className binding, this is an internal helper refined from the `class={{boundProp}}` to add this additional metadata to handle the case of the boundProp not being a string but a boolean

packages/ember-glimmer/lib/helpers/-class.js
```diff
@@ -0,0 +1,25 @@
...
+function classHelper({ positional }) {
+  let path = positional.at(0);
+  let propName = positional.at(1);
+  let value = path.value();
+
+  if (value === true) {
+    return dasherize(propName.value());
+  }
+
+  if (value === false) {
```
chadhietala: Once https://github.com/tildeio/glimmer/pull/150 lands this can go away.

Issue #13387, opened 2016-04-20
Ember.computed.max/min: Date object turns back in timestamp 


tennisonchan:
When we use `Ember.computed.min` or `Ember.computed.max` for javascript date object, the returned value was in timestamp instead of a date object.

Demo:
https://ember-twiddle.com/fbb355b2a22bc92bdb3713ec59f866f3?fileTreeShown=false&openFiles=controllers.application.js%2C

```javascript
{
  dates: Ember.computed(function() {
    return [
      new Date("Apr 19 2016"),
      new Date("Apr 20 2016")
    ];
  }),
  earliestDate: Ember.computed.min('dates'),
}

this.get('earliestDate'); // return 1461095445000, the timestamp on 19th April 2016
```

rwjblue:
Internally, we are using `Math.min` to determine the lowest ([see here](https://github.com/emberjs/ember.js/blob/6739196db628b26b403d66c4271715ec14747dfe/packages/ember-runtime/lib/computed/reduce_computed_macros.js#L151)), which [per the spec](http://www.ecma-international.org/ecma-262/6.0/#sec-math.min) does:

> Given zero or more arguments, calls ToNumber on each of the arguments and returns the smallest of the resulting values.

> * If no arguments are given, the result is +∞.
> * If any value is NaN, the result is NaN.
> * The comparison of values to determine the smallest value is done using the Abstract Relational Comparison algorithm (7.2.11) except that +0 is considered to be larger than −0.

`ToNumber` defers to `ToPrimitive` (defined [here](http://www.ecma-international.org/ecma-262/6.0/#sec-toprimitive)) and in this case calls `valueOf` on the `Date` instance.

```js
let date = new Date();
console.log(date.valueOf());
// => 1461208340910
```

---

That was a very roundabout way of saying that I think the current behavior is correct (though somewhat unexpected).

tennisonchan:
Hey @rwjblue, thanks for the details explanation. Yes, it is somewhat unexpected that it changed the datatype of the input values. 

Here is lodash min function returns the date object. 
```javascript
_.min([new Date(), new Date()])
// => Fri Apr 22 2016 08:44:30 GMT-0400 (EDT)
```

rwjblue:
Hmm, did it change the input values or just return a value that isn't the same?

tennisonchan:
It changed the datatype of the value, from a js date object to a number of timestamp.

rwjblue:
[...]

PR #13406, opened 2016-04-24
[Bugfix] Using simple comparison for Ember.computed.max, min #13387

eshirazi:
Using simple comparison instead of Math.max, Math.min for the implementation of Ember.computed.max, Ember.computed.min. issue #13387
This approach is the one used for _.min, _.max.
When computing the min / max of dates, the result will be a Date instance instead of a Number.

### DIFF
--- a/packages/ember-runtime/lib/computed/reduce_computed_macros.js
+++ b/packages/ember-runtime/lib/computed/reduce_computed_macros.js
@@ -110,7 +110,7 @@ export function sum(dependentKey) {
   @public
 */
 export function max(dependentKey) {
-  return reduceMacro(dependentKey, (max, item) => Math.max(max, item), -Infinity);
+  return reduceMacro(dependentKey, (max, item) => item > max ? item : max, -Infinity);
 }
 
 /**
@@ -148,7 +148,7 @@ export function max(dependentKey) {
   @public
 */
 export function min(dependentKey) {
-  return reduceMacro(dependentKey, (min, item) => Math.min(min, item), Infinity);
+  return reduceMacro(dependentKey, (min, item) => item < min ? item : min, Infinity);
 }
 
 /**
--- a/packages/ember-runtime/tests/computed/reduce_computed_macros_test.js
+++ b/packages/ember-runtime/tests/computed/reduce_computed_macros_test.js
@@ -1466,6 +1466,34 @@ QUnit.test('max recomputes when the current max is removed', function() {
   equal(obj.get('max'), 1, 'max is recomputed when the current max is removed');
 });
 
+QUnit.test('max returns a Date type when all inputs are dates', function() {
+  var items = obj.get('items');
+  items.removeObject(1);
+  items.removeObject(2);
+  items.removeObject(3);
+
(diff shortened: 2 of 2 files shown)


rwjblue:
@stefanpenner - Thoughts?

rwjblue:
I'm somewhat concerned that this changes semantics a bit, for example:

    somestring < otherstring

Checks of one is a substring of the other (I believe). Whereas Math.min approach would return NaN when inputs are non-numeric.

oriSomething:
for me it doesn't seems like a concern. it's extends the the computed property helper to a more useful functionality in a very minimal solution

stefanpenner:
> for me it doesn't seems like a concern

Yes, but it changes the existing contract to a public API.

stefanpenner:
> @stefanpenner - Thoughts?

Im unconvinced this is a Bugfix, as it will change the behavior in existing apps. I strongly suspect many are working around the admittedly awkward current functionality, and if we introduce this change we may break them.

Now the docs are pretty vague on this: https://github.com/emberjs/ember.js/pull/13406/files#diff-be5c035bb6df695f80dbf5d940188c68R79 as are the tests.

> Checks of one is a substring of the other (I believe). Whereas Math.min approach would return NaN when inputs are non-numeric.

Ya, currently this guarantees numeric output, which has some assurances. This will need to be taken into account.

--- 

This seems like a new feature, and should most likely go through that process. I would recommend and RFC outlining the change, the potential breaks, rollout plan etc

eshirazi:
Thanks for your quick review! 
I'll cool with creating an RFC if you think that's the recommended solution.
Personally I think the new implementation is highly unlikely to break existing usages because:
- In the current implementation there is no use for passing strings that aren't numeric, because the result is guaranteed to be NaN, so no one would do it anyway.
- If the input array contains numbers or numeric strings, the current functionality will be preserved, because string to number comparisons works like number to number comparisons in Javascript:
```
>>> '12' < 13
true
>>> '13' < 13
false
>>> '14' < 13
false
>>> 13 > '12'
true
```
- If the two compared items are strings, than the new implementation will compare them lexicographically, which is luckily (:smile:) compatible with number comparisons. Even if some of the compared items are negative, because minus has a lower ASCII code than all decimal digits.

The new implementation also allows for comparisons of Dates, and strings that aren't numeric, and follows the guidelines of other common libraries like underscore / lodash.

Anyway, let me know what's the verdict and I'll take it from here :)

stefanpenner:
Just to be clear. The new implementation feels much better, we should just be diligent about not breaking existing code.

Thank you for the write up, the extra thought helps. It does seem like the vector for issues is fairly small.

rwjblue:
Sorry for leaving this open for so long.  We discussed this again at the last weekly meeting, and we decided to stick with our current implementation. It is quite likely that merging this PR would result in breakage for users (folks that are relying on the current semantics for example), and using `Math.max` / `Math.min` as a way to back `Ember.computed.max` / `Ember.computed.min` seems to be the least surprising implementation to us.

I definitely agree that the CP you propose here is useful (especially if you may have to be working on dates), and this would make a wonderful addon.

PR #13407, opened 2016-04-24
[BUGFIX beta] Produce a nice message when undefined is passed to closure action helper

### LINKED ISSUE #12718: Handling undefined closure actions passed down as attributes
Sometimes users may only want/need to handle a subset of the actions sent up by a component. For example, imagine a component that sends up actions `click-a` and `click-b` like so:

```javascript
// controller.js
  actions: {
    handleClickOnA() {}
  }
});
```
```handlebars
<!-- template.bhs -->
{{my-component click-a=(action "handleClickOnA")}}
```
```handlebars
<!-- my-component.hbs -->
<button onclick={{action (action attrs.click-a)}}>Button A</button>
<button onclick={{action (action attrs.click-b)}}>Button B</button>
```

In this case, I only care about handling a click on button A, so I'm only passing the `click-a` attribute down into my component. However, Ember throws a vague error (`Cannot read property 'INVOKE [id=ember12345...]' of null`) when it encounters the action handler on button B in the template for the component, because `attrs.click-b` is undefined. (I've made an Ember Twiddle if you want to see this in action: https://ember-twiddle.com/b36807b311324b1be59a)

The two workarounds to this are both undesirable in my opinion. (1) I can specifiy `click-b` when instantiating the component in template.hbs and map it to an empty dummy action in my controller. This is less than ideal, because it's a bunch of useless and thus potentially misleading code. (2) I can handle the click on button B via an action in the component.js file itself. There, I'd check to see if `attrs.click-b` was specified, and call it if so. So I'm merely implementing a middle-man to forward the action up the chain, which gets tedious if I have many nested components, and is exactly what I thought closure actions were meant to prevent.

I guess the downside to allowing action handlers to be optional would be that users are now at risk of simply forgetting to define these action handlers. But could we just produce a warning in the console instead of throwing an error?

Serabe:
If the action helper is passed an undefined value from `attrs` (as in `(action
attrs.some-undefined-value)`) the output message was a bit cryptic. This
creates a nicer error message for the user.

Take into account that the main problem is `attrs` containing the values and
not Streams, therefore the error cannot show a good label.

Fixes #12718

### DIFF
--- a/packages/ember-routing-htmlbars/tests/helpers/closure_action_test.js
+++ b/packages/ember-routing-htmlbars/tests/helpers/closure_action_test.js
@@ -1,7 +1,6 @@
 import run from 'ember-metal/run_loop';
 import compile from 'ember-template-compiler/system/compile';
 import EmberComponent from 'ember-views/components/component';
-import EmberView from 'ember-views/views/view';
 import { computed } from 'ember-metal/computed';
 import { INVOKE } from 'ember-routing-htmlbars/keywords/closure-action';
 import { subscribe, unsubscribe } from 'ember-metal/instrumentation';
@@ -55,8 +54,8 @@ function registerComponent(name, factory) {
 }
 
 function appendViewFor(template, moduleName='', hash={}) {
-  let view = EmberView.extend({
-    template: compile(template, { moduleName }),
+  let view = EmberComponent.extend({
+    layout: compile(template, { moduleName }),
     [OWNER]: owner
   }).create(hash);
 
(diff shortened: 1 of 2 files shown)


Serabe:
Closure action tests need some cleanup. If I have any spare time before it is time to migrate them to Glimmer2, I will clean them.

packages/ember-routing-htmlbars/tests/helpers/closure_action_test.js
```diff
@@ -268,6 +267,22 @@ QUnit.test('an error is triggered when bound action function is undefined', func
...
+  registerTemplate('components/inner-component', '<button id="inner-button" {{action (action attrs.external-action)}}>Click me</button>');
+
+  view = EmberComponent.extend({
+    layout: compile('{{inner-component}}'),
+    [OWNER]: owner
+  }).create();
+
+  throws(function() {
+    runAppend(view);
+  }, /Action passed is null or undefined in \(action [^)]*\)./);
```
stefanpenner: is it possible to say which template this occurred in?

Serabe: I can show the default target, would that be enough?

rwjblue: That is definitely better. I believe we *could* pass through the source template from the original action invocation, but it would be tricky.

stefanpenner: ya, that would be sweet. But this already is a big win, I was just thinking what would make it even better.

Serabe:
@stefanpenner, @rwjblue I think this is the best we can do now. Thank you for reviewing!

stefanpenner:
we can merge when green

packages/ember-routing-htmlbars/tests/helpers/closure_action_test.js
```diff
@@ -55,7 +54,7 @@ function registerComponent(name, factory) {
 }
 
 function appendViewFor(template, moduleName='', hash={}) {
-  let view = EmberView.extend({
+  let view = EmberComponent.extend({
     template: compile(template, { moduleName }),
```
rwjblue: This needs to change to `layout` (a `template` property does nothing in a component).

rwjblue:
One tweak, but looks good

Serabe:
@rwjblue: fixed.

rwjblue:
Thanks @Serabe!

PR #13408, opened 2016-04-25
Move computed property tests to "ember-metal" package

Turbo87:
It seems that these were forgotten when the actual code files were moved from "ember-runtime" to "ember-metal"


krisselden:
I was trying to understand this PR, because the computed macros should be in runtime, now I'm trying to figure out when they were moved to metal.  /cc @stefanpenner

Turbo87:
@krisselden looks like I was wrong, they apparently have always been in "metal" (see https://github.com/emberjs/ember.js/commit/c7b11e70ce0a33e6fc264fafa6f4b80765bf7b6c)

also:

> tbieniek [20:31] 
> am I blind or are there currently no tests for `Ember.computed.and/or`?
> 
> [20:32] 
> nevermind, found them... although I don't understand why the code is in `ember-metal` and the tests are in `ember-runtime`
> 
> rwjblue [22:05] 
> @tbieniek: computed used to live in runtime, then was moved
> 
> [22:05] 
> must have missed the tests

krisselden:
@Turbo87 I would accept a PR that moved them alongside reduce_computed_macros.js

They shouldn't have ever been added to ember-metal.

stefanpenner:
> I was trying to understand this PR, because the computed macros should be in runtime, now I'm trying to figure out when they were moved to metal. /cc @stefanpenner

I seem to remember we moved them for a reason, but this would have been years ago. I don't recall why.

Turbo87:
@stefanpenner this must have been before the computed macros were split from the "computed" concept itself so I guess that the reason might not have been valid for the macros.

stefanpenner:
I believe in metal we actually use computed (eachComputed etc), but ya the macros could come in runtime

krisselden:
closed in favor of #13409

PR #13410, opened 2016-04-25
Add node test for protected bound attrs with no value

ronco:
See https://github.com/tildeio/htmlbars/pull/455

This test demonstrates that with the current version of htmlbars bound
view attributes requiring sanitization will throw an error in node if
their value is undefined.

### DIFF
--- a/tests/node/component-rendering-test.js
+++ b/tests/node/component-rendering-test.js
@@ -44,6 +44,17 @@ QUnit.test("Component with dynamic value", function(assert) {
   assert.ok(html.match(/<h1>Hello World<\/h1>/));
 });
 
+QUnit.test("Ensure undefined attributes requiring protocol sanitization do not error", function(assert) {
+  var component = buildComponent('', {
+    tagName: 'link',
+    attributeBindings: ['href', 'rel'],
+    rel: 'canonical'
+  });
+
+  var html = renderComponent(component);
+  assert.ok(html.match(/rel="canonical"/));
+});
+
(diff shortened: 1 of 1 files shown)


tests/node/component-rendering-test.js
```diff
@@ -44,6 +44,17 @@ QUnit.test("Component with dynamic value", function(assert) {
   assert.ok(html.match(/<h1>Hello World<\/h1>/));
 });
 
+QUnit.test("Component with dynamic safe attribute undefined", function(assert) {
```
rwjblue: I understand what you are doing here (because I read through the HTMLBars PR), but this test title doesn't seem very descriptive.  Maybe something like:

> ensure undefine attributes requiring protocol sanitization do not error

ronco:
@rwjblue updated the test description as per your recommendation.  Thanks!

rwjblue:
I believe if you rebase, this test should pass.

package.json
```diff
@@ -40,7 +40,7 @@
     "github": "^0.2.3",
     "glimmer-engine": "tildeio/glimmer#2c22999",
     "glob": "^5.0.13",
-    "htmlbars": "0.14.16",
+    "htmlbars": "0.14.18",
```
rwjblue: I think master has updated to 0.14.20, can you drop this change?

ronco:
@homu @rwjblue Rebased and tests appear to be running again.  Thanks!

rwjblue:
Thanks @ronco$

PR #13411, opened 2016-04-25
Expose Ember.lookup for compatibility with add-ons using it.

### LINKED ISSUE #13352: Issue with Ember.lookup in canary
While testing an addon, I came across the following issue: When using the current Ember canary with Ember-Data canary (via bower, not the ember-data addon version in package.json), Ember-Data throws an error when trying to set `Ember.lookup.DS`

```
_ember.default.lookup.DS = _emberDataPrivateCore.default;

Stacktrace:
ember-data.js:12881 Uncaught TypeError: Cannot set property 'DS' of undefined
    (anonymous function) @ ember-data.js:12881
    Module.exports @ ember-data.js:94
    Module.build @ ember-data.js:144
    findModule @ ember-data.js:192
    requireModule @ ember-data.js:179
    (anonymous function) @ ember-data.js:16707
    (anonymous function) @ ember-data.js:16708
```

Steps to reproduce: 

1. Create a new ember app
2. Add dependency for ember-data#canary in `bower.json`

krisselden:
We will add deprecation in a followup PR, this is just to stop breaking add-ons using it.

Fixes #13352.

### DIFF
--- a/packages/ember-metal/lib/index.js
+++ b/packages/ember-metal/lib/index.js
@@ -379,6 +379,17 @@ Object.defineProperty(Ember, 'ENV', {
 });
 
 /**
+  The context that Ember searches for namespace instances on.
+
+  @private
+ */
+Object.defineProperty(Ember, 'lookup', {
+  get: () => context.lookup,
+  set: value => context.lookup = value,
+  enumerable: false
+});
+
+/**
   A function may be assigned to `Ember.onerror` to be called when Ember
   internals encounter an error. This is useful for specialized error handling
   and reporting code.


rwjblue:
Thanks @krisselden, this LGTM (and fixes #13352).

packages/ember-metal/lib/index.js
```diff
@@ -379,6 +379,16 @@ Object.defineProperty(Ember, 'ENV', {
 });
 
 /**
+ * The context that Ember searches for namespace instances on.
+ * @private
```
rwjblue: Looks like the JSCS rule that requires `@private` / `@public` doesn't like it when the comments include a leading `* ` like this.

rwjblue:
Thanks @krisselden!

PR #13413, opened 2016-04-25
[FEATURE beta] Refactor auto mut in closure components to use traverse

Serabe:
When adding the plugin for adding mut to contextual components I did not know
about `this.syntax.traverse` so I had to rely on a hack on `walker.visit`
creating a fairly complicated code.

This PR simplifies that code while keeping the tests green.


rwjblue:
:heart:

PR #13415, opened 2016-04-26
Introduce ember-templates package

krisselden:
TODO
- [x] move LinkComponent reexport and fork it
- [x] move bootstrap tests
- [ ] createElement() booting a hard coded HTMLBars renderer.

Going to do the other in a separate PR


rwjblue:
It looks like a number of the modules being moved into `ember-templates` (i.e. `ember-templates/component`) are just shims around other modules.  Why create additional shims, instead of moving the implementation itself?

rwjblue:
Looks like `Ember._Renderer` is not being exported after this (it is used in the node tests for example).

krisselden:
@rwjblue Ember._Renderer needs to be different depending on glimmer flag too, I'll update to export it.

krisselden:
@rwjblue They are lazy for tests that run ENABLE_ALL_FEATURES so they can be imported without relying on the reexports.

krisselden:
@chadhietala green. I'll try to do the TODO, there are still issues in that this does not attempt to deal with createElement() booting a hard coded HTMLBars renderer but I think maybe I should merge then continue working on the TODO.

chadhietala:
Yea lets land it as is.

PR #13416, opened 2016-04-26
[BUGFIX beta] Change error message when Ember.computed.or contains a space

### LINKED ISSUE #13176: Confusing error when using `Ember.computed.or` incorrectly
I wasn't thinking very hard, and tried to do:
`Ember.computed.or('someKey', 'tried to put a default string here')`
The error that popped up didn't really inform me what was going on:
> Error: Brace expanded properties cannot contain spaces, e.g. 'user.{firstName, lastName}' should be 'user.{firstName,lastName}'

More helpful would have been an assertion error, something like
> Assertion Failed: Dependent keys passed to Ember.computed.or() can't have spaces

tyleriguchi:
Fixes #13176 by returning a less confusing error message and example.

### DIFF
--- a/packages/ember-metal/lib/expand_properties.js
+++ b/packages/ember-metal/lib/expand_properties.js
@@ -38,7 +38,8 @@ var END_WITH_EACH_REGEX = /\.@each$/;
 export default function expandProperties(pattern, callback) {
   assert('A computed property key must be a string', typeof pattern === 'string');
   assert(
-    'Brace expanded properties cannot contain spaces, e.g. "user.{firstName, lastName}" should be "user.{firstName,lastName}"',
+    'Dependent keys passed to Ember.computed.or() cannot contain spaces,' +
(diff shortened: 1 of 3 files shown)


packages/ember-metal/lib/expand_properties.js
```diff
@@ -38,7 +38,8 @@ var END_WITH_EACH_REGEX = /\.@each$/;
 export default function expandProperties(pattern, callback) {
   assert('A computed property key must be a string', typeof pattern === 'string');
   assert(
-    'Brace expanded properties cannot contain spaces, e.g. "user.{firstName, lastName}" should be "user.{firstName,lastName}"',
+    'Dependent keys passed to Ember.computed.or() cannot contain spaces,' +
```
rwjblue: This method is used by all computed properties (and observers), saying computed.or here isn't necessarily correct

pixelhandler: @rwjblue any tip on where to add the not on `computed.or` ?

PR #13418, opened 2016-04-26
[BUGFIX release] run.later(cb, NaN) should not break all the things

stefanpenner:


rwjblue:
Thanks @stefanpenner!

Issue #13243, opened 2016-04-04
Deprecate Ember.Binding


mmun:
This issue tracks the deprecation of Ember.Binding (and *Binding properties). They've not been needed for a very long time. The existing use cases are covered by computeds and service injections.

- [ ] Add an entry in the deprecation guides
- [ ] Add deprecations to core
  - Trigger a deprecation when the binding connects so that we can provide better upgrade suggestions. If its oneWay, suggest using a readOnly computed. Otherwise, suggest an alias. If the binding is to a global, suggest using a service.

rwjblue:
@mmun - Sounds like a good plan, thanks for working on it...

krisselden:
@mmun this alternatives are extend time, bindings are allowed at create time, I don't think we need an alternative, I'm just pointing it out for the guide.

mmun:
Good point

mmun:
If someone wants to give this a shot in the meantime, please do!

robbiepitts:
I'll take a stab!

mmun:
@robbiepitts Thanks so much! Hit me up in the #dev-ember channel of the community Slack if you have any problems.

Serabe:
@mmun I see both PRs have been merged. Is there something left to do or can this be closed?

Thank you!

robbiepitts:
@Serabe This can be closed, thanks!

mmun:
Thanks all!

PR #13424, opened 2016-04-28
[DEPRECATE] Deprecate Ember.Binding

pittst3r:
Addresses issue #13243.

This is ready for review but not merge. Still to be done is the deprecation guide update. Once I finish that I will update the links in this PR and squash.

cc @mmun

### DIFF
--- a/packages/ember-glimmer/tests/integration/helpers/unbound-test.js
+++ b/packages/ember-glimmer/tests/integration/helpers/unbound-test.js
@@ -252,9 +252,9 @@ moduleFor('Helpers test: {{unbound}}', class extends RenderingTest {
 
     this.render(`{{unbound (repeat foo count=bar)}} {{repeat foo count=bar}} {{unbound (repeat foo count=2)}} {{repeat foo count=4}}`, {
       foo: 'X',
-      numRepeatsBinding: 'bar',
       bar: 5
     });
+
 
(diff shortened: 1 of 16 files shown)


robbiepitts:
Looks like I missed some tests. Update forthcoming.

robbiepitts:
Links to guide added and commits squashed. PR for deprecation guide referenced above. 😀

packages/ember-glimmer/tests/integration/helpers/unbound-test.js
```diff
@@ -250,11 +250,17 @@ moduleFor('Helpers test: {{unbound}}', class extends RenderingTest {
       return a.join('');
     });
 
-    this.render(`{{unbound (repeat foo count=bar)}} {{repeat foo count=bar}} {{unbound (repeat foo count=2)}} {{repeat foo count=4}}`, {
-      foo: 'X',
-      numRepeatsBinding: 'bar',
```
rwjblue: Is this even used in this test?  I don't see any usage of `numRepeats`?

robbiepitts: Good find, doesn't look like it. Also doesn't appear to me to be relevant to the test. Tests remain green after removal of that and the deprecation expectation.

I will amend my commit with these changes.

PR #13425, opened 2016-04-28
[GLIMMER2] Update glimmer version and un-gate helper element form tests

zackthehuman:
This bumps `glimmer-engine` to the version containing basic support for element modifiers. As a result, we can now detect if helpers are used in element form and throw an error. This also enables the associated test for Glimmer.

### DIFF
--- a/packages/ember-glimmer/lib/environment.js
+++ b/packages/ember-glimmer/lib/environment.js
@@ -42,6 +42,9 @@ const builtInHelpers = {
   '-class': classHelper
 };
 
+const builtInModifiers = {
+};
+
 function wrapClassAttribute(args) {
   let hasClass = args.named.has('class');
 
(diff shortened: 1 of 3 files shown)


packages/ember-glimmer/lib/environment.js
```diff
@@ -96,6 +100,7 @@ export default class Environment extends GlimmerEnvironment {
 
     let nativeSyntax = super.refineStatement(statement);
     assert(`Helpers may not be used in the block form, for example {{#${key}}}{{/${key}}}. Please use a component, or alternatively use the helper in combination with a built-in Ember helper, for example {{#if (${key})}}{{/if}}.`, !nativeSyntax && key && this.hasHelper(key) ? !isBlock : true);
+    assert(`Helpers may not be used in the element form.`, !nativeSyntax && key && this.hasHelper(key) ? !isModifier : true);
```
chancancode: Can you include the example in here too? i.e. https://github.com/emberjs/ember.js/blob/master/packages/ember-htmlbars/lib/system/invoke-helper.js#L13

Issue #11563, opened 2015-06-26
QP only transition in beforeModel of application route throws error


raytiley:
Reproduction: http://emberjs.jsbin.com/verocu/2/edit?html,js,output

Our app does some `beforeModel` checking of data to see what "location" the user is operating. If know query param for this is provided call `transitionTo` in order to force the user is at a location.

Because of some other logic I just added a `catch` to the promise chain that the `transitionTo` is a part of.  Now I'm noticing the error shown in the above jsbin.

Interesting thing when trying to recreate is that if you have no `catch` on the promise chain and the QP is set `refreshTrue` then a second transition will fire, will have the correct query param and so the user won't notice the failure.

mattmarcum:
:+1: I created two failing tests for this  

https://github.com/emberjs/ember.js/compare/master...mattmarcum:transition-to-in-before-model-with-query-params-test?expand=1

I went ahead and created failing tests in both the default code path and in the `ember-routing-route-configured-query-params` code path (which you can run in the tests by turning that feature flag on).

This also fail in the model hook, I've only tried those two hooks.

Anyone know whats up with that feature-flag?  Its not in the FEATURES.md yet...

mattmarcum:
This bug is pretty tricky because its caused by interaction between tildeo/router.js and ember.js/route.

When router.js is doing a queryParamTransition it generates an empty Transition because it should be a no-op transition.  But ember.js/route needs to know the leaf node so that it can lookup the query params on the controller and do all its magic.  When ember.js/route#finalizeQueryParamChange finally gets called, the transition passed to it contains no information about the current route, `transition.state.handlerInfos` is an empty array.

So it really seems like the fix should go in tildeo/router.js, probably around this line: https://github.com/tildeio/router.js/blob/master/lib/router/router.js#L126

If we did something like:
```
newTransition.state.handlerInfos = newState.handlerInfos;
```

I think that might fix it, but I'm not sure if that would break anything else;

mattmarcum:
Actually that doesn't fix it.  Even if you get the `handlerInfos`  the controller for the route isn't created yet...

mattmarcum:
This is definitely caused by the fact that `route.setup` has not been called yet, it's not called until after the `beforeModel` hook.  `finalizeQueryParams` expects that the route has already been setup.  

Got to love the comments here: https://github.com/emberjs/ember.js/blob/v2.5.0/packages/ember-routing/lib/system/route.js#L1182-L1183

[...]

PR #13426, opened 2016-04-28
[BUGFIX beta] fixes transitionTo failing when called in beforeModel hook

mattmarcum:
This is a fix for [11563](https://github.com/emberjs/ember.js/issues/11563).

This bug is caused when a `queryParamOnly` `transitionTo` is called in the `beforeModel` hook.  

There's a couple of problems here.  One is that `route.finalizeQueryParams` expects the route's controller to already be setup, but it isn't until the `route.setup` hook is called.  The problem is that `route.setup` is not called until after the `beforeModel` hook is run.

The solution I came up with is to extract out the relevant portions of the controller setup from `route.setup` and then check during `route.finalizeQueryParams` to see if the controller has been added, and if not, add it to the route.

There was another problem with `updatePaths` on the router not exiting out quick enough.  @raytiley has a similar fix in https://github.com/emberjs/ember.js/pull/13273 but in my case the `infos` param was not even an array.


mattmarcum:
All the tests pass locally I promise!  I think the build is failing because of a deprecated package in npm.

rwjblue:
I think its fixed (we cleared the Travis cache), going to restart the build...

packages/ember-routing/lib/system/route.js
```diff
@@ -811,18 +811,26 @@ var Route = EmberObject.extend(ActionHandler, Evented, {
 
       var handlerInfos = transition.state.handlerInfos;
       var router = this.router;
-      var qpMeta = router._queryParamsFor(handlerInfos[handlerInfos.length - 1].name);
+      var qpMeta;
       var changes = router._qpUpdates;
       var replaceUrl;
 
-      stashParamNames(router, handlerInfos);
+      if (!handlerInfos.length) {
```
rwjblue: So we don't want to `stashParamNames` if there are no `handlerInfos`?  This is to ensure that `#setup` has been called first, right?  If so, is there a better way to check this (or add a comment that explains why)?

packages/ember-routing/lib/system/router.js
```diff
@@ -987,6 +988,9 @@ function calculatePostTransitionState(emberRouter, leafRouteName, contexts) {
 
 function updatePaths(router) {
   let infos = router.router.currentHandlerInfos;
+
+  if (!isArray(infos) || infos.length === 0) { return; }
```
rwjblue: Can we just use `Array.isArray` here? Since `infos` is always going to be a native array, we don't need `Ember.isArray`'s support for `Enumerable` and whatnot.

raytiley:
This looks good to me with the caveat that my eyes start to bleed when staring at at the router code for too long.

This will be great cause this issue is what I was blocked on for #13273. Once this is in I'll finish up some of the other tests and then query params not bug people so much for a while :)

rwjblue:
I think that this augments #13273 pretty well. I had a couple of minor nitpicks that I'd like to see addressed, but other than that I'm pretty happy with this.  Thank you for taking the time to work on it!

mattmarcum:
@rwjblue fixed those.  I used `transition.queryParamsOnly` when deciding whether to call `stashParamNames`.  The `queryParamsOnly` flag gets set when the no-op transition gets created (hence the reason the handlerInfos are empty).  I think it makes it more clear.

rwjblue:
Awesome, thank you!  Will merge once Travis agrees....

mattmarcum:
Oops, looks like that last change broke some other tests...I'm going to revert back and just add a comment instead.

mattmarcum:
After looking at this code again, I'm thinking that a better solution would need to touch the `router.js` code too.  I'm going to close this pr and think about a better solution.  Any one else, feel free to try and fix this.

PR #13427, opened 2016-04-28
[DOC release] Fix computed property syntax with prototype extensions.

eventualbuddha:
The previous code was a syntax error since you can't use a shorthand object method as part of a member expression.


rwjblue:
Awesome, thank you!

PR #13429, opened 2016-04-28
router.js: Fixed YUIDoc tag

lolmaus:
The `forEachRouteAbove` function is a closure. No need to define it as `@private`: it's not part of the class, cannot be overridden and cannot even be invoked outside the file.

Having a `@private` tag without a `@method`/`@property`/`@event` results in an unnamed, unclassified item appearing in YUIDoc's resulting JSON.

### DIFF
--- a/packages/ember-routing/lib/system/router.js
+++ b/packages/ember-routing/lib/system/router.js
@@ -792,15 +792,13 @@ var EmberRouter = EmberObject.extend(Evented, {
   }
 });
 
-/*
-  Helper function for iterating root-ward, starting
-  from (but not including) the provided `originRoute`.
 
-  Returns true if the last callback fired requested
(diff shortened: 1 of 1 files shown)


rwjblue:
YUIDoc only looks for documentation blocks that start with `/**`, and the one that you replaced started with `/*` so it should have been ignored.

From the intro paragraph of https://yui.github.io/yuidoc/syntax/index.html:

> YUIDoc relies on tags such as @param or @return embedded in comment blocks that start with /** and end with */.

lolmaus:
You're right.

I've ported YUIDoc.DocParser to ES6/Ember. It has the same comment start regex: `/^\s*\/\*\*/`, but for some reason it does include this comment.

PR #13430, opened 2016-04-29
[Glimmer2] Rerender on run loop if data has changed

nickiaconis:
Marking an object as dirty kicks off a run loop if one has not already been started. At the start of every run loop, a task is scheduled on the render queue to rerender the main view if any changes have occurred.

Tested by disabling the automatic invocation of manual rerendering in the glimmer tests.

Will require ember-dev to be updated since this invalidates the assumption that a run loop should never be automatically started inside a test.

### DIFF
--- a/packages/ember-glimmer/lib/ember-metal-views/index.js
+++ b/packages/ember-glimmer/lib/ember-metal-views/index.js
@@ -1,6 +1,6 @@
 import { RootReference } from '../utils/references';
 import run from 'ember-metal/run_loop';
-// import { CURRENT_TAG } from 'glimmer-reference';
+import { CURRENT_TAG } from 'glimmer-reference';
 
 class DynamicScope {
   constructor({ view, controller, outletState, isTopLevel }) {
@@ -16,15 +16,15 @@ class DynamicScope {
 }
 
 const RENDERED_ROOTS = [];
-// let LAST_TAG_VALUE;
+let LAST_TAG_VALUE;
 
 function maybeUpdate() {
-  // if (CURRENT_TAG.validate(LAST_TAG_VALUE)) { return; }
+  if (CURRENT_TAG.validate(LAST_TAG_VALUE)) { return; }
   for (let i = 0; i < RENDERED_ROOTS.length; ++i) {
     let view = RENDERED_ROOTS[i];
     view.renderer.rerender(view);
   }
-  // LAST_TAG_VALUE = CURRENT_TAG.value();
+  LAST_TAG_VALUE = CURRENT_TAG.value();
 }
 
 function scheduleMaybeUpdate() {
@@ -32,7 +32,7 @@ function scheduleMaybeUpdate() {
 }
 
 function registerView(view) {
-  // LAST_TAG_VALUE = LAST_TAG_VALUE || CURRENT_TAG.value();
+  LAST_TAG_VALUE = LAST_TAG_VALUE || CURRENT_TAG.value();
   if (!RENDERED_ROOTS.length) {
     run.backburner.on('begin', scheduleMaybeUpdate);
   }
--- a/packages/ember-glimmer/lib/ember-metal-views/index.js
+++ b/packages/ember-glimmer/lib/ember-metal-views/index.js
@@ -28,11 +28,11 @@ function maybeUpdate() {
 }
 
 function scheduleMaybeUpdate() {
+  LAST_TAG_VALUE = CURRENT_TAG.value();
   run.backburner.schedule('render', maybeUpdate);
 }
 
 function registerView(view) {
-  LAST_TAG_VALUE = LAST_TAG_VALUE || CURRENT_TAG.value();
   if (!RENDERED_ROOTS.length) {
     run.backburner.on('begin', scheduleMaybeUpdate);
   }
--- a/packages/ember-glimmer/lib/ember-metal-views/index.js
+++ b/packages/ember-glimmer/lib/ember-metal-views/index.js
@@ -15,36 +15,36 @@ class DynamicScope {
   }
 }
 
-const RENDERED_ROOTS = [];
-let LAST_TAG_VALUE;
-
-function maybeUpdate() {
-  if (CURRENT_TAG.validate(LAST_TAG_VALUE)) { return; }
-  for (let i = 0; i < RENDERED_ROOTS.length; ++i) {
-    let view = RENDERED_ROOTS[i];
-    view.renderer.rerender(view);
+class Scheduler {
+  constructor() {
+    this._roots = [];
+    this._scheduleMaybeUpdate = () => {
+      run.backburner.schedule('render', this, this._maybeUpdate, CURRENT_TAG.value());
+    };
   }
-  LAST_TAG_VALUE = CURRENT_TAG.value();
-}
 
-function scheduleMaybeUpdate() {
-  LAST_TAG_VALUE = CURRENT_TAG.value();
-  run.backburner.schedule('render', maybeUpdate);
-}
-
-function registerView(view) {
-  if (!RENDERED_ROOTS.length) {
-    run.backburner.on('begin', scheduleMaybeUpdate);
+  registerView(view) {
+    if (!this._roots.length) {
+      run.backburner.on('begin', this._scheduleMaybeUpdate);
+    }
+    this._roots.push(view);
   }
-  RENDERED_ROOTS.push(view);
-}
 
-function deregisterView(view) {
-  let viewIndex = RENDERED_ROOTS.indexOf(view);
-  if (~viewIndex) {
-    RENDERED_ROOTS.splice(viewIndex, 1);
-    if (!RENDERED_ROOTS.length) {
-      run.backburner.off('begin', scheduleMaybeUpdate);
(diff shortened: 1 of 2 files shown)


packages/ember-glimmer/lib/ember-metal-views/index.js
```diff
@@ -13,6 +15,17 @@ class DynamicScope {
   }
 }
 
+function setupRerenderLoop(renderer, view) {
+  // let LAST_TAG_VALUE = CURRENT_TAG.value();
+  run.backburner.on('begin', () => {
+    run.backburner.schedule('render', () => {
```
krisselden: we don't need to recreate this anon function every time
it also needs to go away when the root view is gone, maybe the renderer should track root views.

packages/ember-metal/lib/tags.js
```diff
@@ -1,5 +1,7 @@
 import { meta as metaFor } from './meta';
 import require, { has } from 'require';
+import run from './run_loop';
+import { K } from './core';
```
krisselden: just define `function K() {}`

packages/ember-glimmer/lib/ember-metal-views/index.js
nickiaconis: So the Renderer's appendOutletView and appendTo methods would add a view to the list, and its remove method would remove that view? Then, when a view is removed, if the list is empty the begin handler is unregistered from the run loop?

packages/ember-glimmer/lib/ember-metal-views/index.js
```diff
@@ -61,6 +99,7 @@ class Renderer {
   }
 
   remove(view) {
+    deregisterViewAutorender(view);
```
krisselden: does this get called while looping over RENDERED_VIEWS or out of band?

packages/ember-glimmer/lib/ember-metal-views/index.js
```diff
@@ -13,6 +15,38 @@ class DynamicScope {
...
 }
 
+const RENDERED_VIEWS = [];
+// let LAST_TAG_VALUE;
+
+function maybeRerender() {
+  // if (CURRENT_TAG.validate(LAST_TAG_VALUE)) { return; }
+  for (let i = 0; i < RENDERED_VIEWS.length; i += 2) {
+    let view = RENDERED_VIEWS[i];
+    let renderer = RENDERED_VIEWS[i + 1];
```
krisselden: renderer is injected onto the root view

packages/ember-glimmer/lib/ember-metal-views/index.js
```diff
@@ -37,6 +71,8 @@ class Renderer {
     let result = view.template.asEntryPoint().render(self, env, { appendTo: target, dynamicScope });
     env.commit();
 
+    registerViewAutorender(view, this);
```
krisselden: just make it registerView

packages/ember-glimmer/lib/ember-metal-views/index.js
```diff
@@ -13,6 +15,38 @@ class DynamicScope {
...
+  // if (CURRENT_TAG.validate(LAST_TAG_VALUE)) { return; }
+  for (let i = 0; i < RENDERED_VIEWS.length; i += 2) {
+    let view = RENDERED_VIEWS[i];
+    let renderer = RENDERED_VIEWS[i + 1];
+    if (!view.isDestroyed) { renderer.rerender(view); }
+  }
+  // LAST_TAG_VALUE = CURRENT_TAG.value();
+}
+
+function scheduleRerender() {
```
krisselden: rerender and renderer  visually looks similar, maybe scheduleMaybeUpdate and maybeUpdate?

packages/ember-metal/lib/tags.js
```diff
@@ -19,11 +20,19 @@ export function tagFor(object, _meta) {
   }
 }
 
+function K() {}
+function ensureRunloop() {
+  if (!run.backburner.currentInstance) {
```
krisselden: `const { backburner } = run;` up top?

packages/ember-metal/lib/tags.js
```diff
@@ -19,11 +20,19 @@ export function tagFor(object, _meta) {
...
+    run.backburner.schedule('actions', K);
+  }
+}
+
 if (hasGlimmer) {
   ({ DirtyableTag, CONSTANT_TAG, CURRENT_TAG } = require('glimmer-reference'));
   makeTag = function() { return new DirtyableTag(); };
 
   markObjectAsDirty = function(meta) {
+    ensureRunloop();
```
krisselden: since the current ember-glimmer test harness always starts a run loop with runTask() we could do this in a followup PR, just focus on getting the other part in place and validating the tag.

packages/ember-glimmer/lib/ember-metal-views/index.js
```diff
@@ -61,6 +98,7 @@ class Renderer {
   }
 
   remove(view) {
```
krisselden: is this only called on a top level view?

packages/ember-glimmer/lib/ember-metal-views/index.js
```diff
@@ -13,6 +15,37 @@ class DynamicScope {
   }
 }
 
+const RENDERED_VIEWS = [];
```
krisselden: should name this clearly, like RENDERED_ROOTS; want to ensure people realize this are roots and not any view.

packages/ember-glimmer/tests/utils/test-case.js
```diff
@@ -31,10 +31,10 @@ export class RenderingTest extends AbstractRenderingTest {
     this.renderer._root = this.component;
   }
 
-  runTask(callback) {
-    super.runTask(() => {
-      callback();
-      this.component.rerender();
-    });
-  }
+  // runTask(callback) {
```
krisselden: just remove this.

packages/ember-metal/lib/tags.js
```diff
@@ -1,6 +1,8 @@
 import { meta as metaFor } from './meta';
 import require, { has } from 'require';
+// import run from './run_loop';
```
krisselden: better to pull this up into its own commit based on this PR

packages/ember-glimmer/lib/ember-metal-views/index.js
```diff
@@ -13,12 +15,41 @@ class DynamicScope {
...
 class Renderer {
   constructor({ dom, env, destinedForDOM = false }) {
     this._root = null;
     this._dom = dom;
     this._env = env;
     this._destinedForDOM = destinedForDOM;
+    this._renderedRoots = [];
+    this._scheduler = function scheduleMaybeUpdate() {
+      run.backburner.schedule('render', null, maybeUpdate, CURRENT_TAG.value(), this._renderedRoots);
+    }.bind(this);
```
krisselden: Just use () => {} instead of bind

krisselden: Could also just make this the second arg too and put maybeUpdate on the renderer and not pass the args but just use this

packages/ember-glimmer/lib/ember-metal-views/index.js
```diff
@@ -13,12 +15,41 @@ class DynamicScope {
...
+  }
+}
+
 class Renderer {
   constructor({ dom, env, destinedForDOM = false }) {
     this._root = null;
     this._dom = dom;
     this._env = env;
     this._destinedForDOM = destinedForDOM;
+    this._renderedRoots = [];
```
krisselden: Make this an object with register and unregister methods

packages/ember-glimmer/lib/ember-metal-views/index.js
stefanpenner: could we make the caller instead support the following (that way we don't need to retain the context, or rebind later?

```js
invoke(schedulable)

function invoke(x) {
  // later invoke x on the method in schedulable
  x.scheduleMaybeUpdate()
}
```

nickiaconis: @stefanpenner I don't understand how this helps with the fact that the callback's context is lost once it is passed to `backburner.on`. Could you explain?

stefanpenner: sorry, i was mostly thinking outloud, I think it would be great if someday we merely have some nice protocol to avoid that.

nickiaconis:
Kicking Travis...

packages/ember-glimmer/lib/ember-metal-views/index.js
nickiaconis: Something like how you can pass a context to schedule? That would be nice.

chadhietala:
LGTM. Are there any tests that we can either flip to `@test` or ones we could write to validate? Maybe `this.runTask` stuff can now be cleaned up?

nickiaconis:
I think `this.runTask` needs to stay for now. The changes here rerender on
the run loop, but we'll need to ensure a run loop is running whenever
something is set, which is not included here, before we can strip the
`this.runTask` stuff.

PR #13436, opened 2016-05-02
Bump glimmer-engine version and re-export UNDEFINED and NULL references

zackthehuman:
This bumps up the version of glimmer-engine so that the exported `UNDEFINED_REFERENCE` and `NULL_REFERENCE` can be used.


chadhietala:
Restarting CI as Safari just bailed out.

PR #13437, opened 2016-05-02
Add test for parameterless helper invocation.

### LINKED ISSUE #12506: Parameter-less helper calls don't work inside of attributes
Example: http://ember-twiddle.com/098c793577e88f993e1d

Given a `foo-bar` helper, you aren't able to call it without parameters inside of an attribute: 
```
<a href={{foo-bar}}>look at my href</a>
```

Quick workaround is to specify a parameter that won't be used:
```
<a href={{foo-bar helper=true}}>look at my href</a>
```

rwjblue:
Was enabled by work done in https://github.com/tildeio/glimmer/pull/148.

Closes #12506.


mmun:
r+

krisselden:
needs rebase, probably just the package.json change.

Issue #13432, opened 2016-04-29
Simple extension of {{link-to}} component fails 


mike-north:
In ember 2.5, one can extend the `{{link-to}}` component as follows

**my-component.js**
```js
import Ember from 'ember';

export default Ember.LinkComponent.extend({
});
```

**my-component.hbs**
(copied from the [framework template for link-to](https://github.com/emberjs/ember.js/blob/release/packages/ember-htmlbars/lib/templates/link-to.hbs), and very slightly tweaked)
```hbs
[{{#if linkTitle}}{{linkTitle}}{{else}}{{yield}}{{/if}}]
```

**Example:** https://ember-twiddle.com/488364295bf020f03f62307c7f9e0349?fileTreeShown=false&numColumns=2&openFiles=components.my-link.js%2Ctemplates.components.my-link.hbs

If you switch the framework version to the beta channel, this example fails, and the error that's thrown seems to indicate a change in the ordering of parameters

```
Uncaught Error: There is no route named Index
```

Serabe:
They thing is this is working properly because there is no longer an inline version of `link-to` since it is now a trick: it is done in a plugin for the compiler (as seen [here](https://github.com/emberjs/ember.js/blob/2affd3b929bbd51ea7075947158ba90555a763b6/packages/ember-template-compiler/lib/plugins/transform-inline-link-to.js). This was introduced in #12229 and if you follow the thread there, you see that the opinion is that the inline form is only for `{{link-to}}` and not for the component that extends from it.

mike-north:
This is going to cause a breaking change in a bunch of addons, in a way that makes it quite difficult to write tests around (i.e., existing behavior should work, but only for Ember < 2.5) and to consume as a developer.

I would argue that it violates SemVer, since we advise people to customize {{link-to}} by reopening it or extending it.

rwjblue:
> This is going to cause a breaking change in a bunch of addons

Can you list a few for context?

> I would argue that this change violates SemVer, since we advise people to customize {{link-to}} by reopening it or extending it.  Just like with a public API surface, common usage patterns that were valid (and officially supported) in Ember 2.5 should continue in 2.6+.

I am not sure that non-block usage of something extending from `Ember.LinkComponent` is something that we have suggested (though perhaps it is).

Either way, I agree that if this is commonly in use in the wild that we either need to fix or revert.

mike-north:
> Can you list a few for context?

* ember-cli-materialize
[...]

PR #12229, opened 2015-08-28
Reimplement inline link-to as an AST transform

mmun:
I also had to fix broken tests that didn't pass enough arguments to link-to because they were triggering an assertion that `params.length > 0`.

PR #13438, opened 2016-05-02
[BUGFIX beta] Partially revert #12229 and add a test for #13432

Serabe:
This PR partially reverts #12229 given that it made the inline form of
components extending from `LinkTo` impossible.

Beware that matching the exact behaviour is not there yet (as it wasn't
before #12229) given that `{{{link-to title route}}}` would unescape `title`
while `{{{my-link-to title route}}}` would not. This behaviour was not working
before and therefore is not working after this revert.

This PR does not address whether the inline form for `link-to` should be
deprecated or not.

### DIFF
--- a/packages/ember-routing-htmlbars/tests/helpers/link-to_test.js
+++ b/packages/ember-routing-htmlbars/tests/helpers/link-to_test.js
@@ -161,4 +161,16 @@ QUnit.test('able to safely extend the built-in component and use the normal path
   equal(view.$().text(), 'my custom link-to component', 'rendered a custom-link-to component');
 });
 
+QUnit.test('[GH#13432] able to safely extend the built-in component and invoke it inline', function() {
+  view = EmberView.create({
+    [OWNER]: owner,
+    title: 'my custom link-to component',
+    template: compile('{{custom-link-to view.title \'index\'}}')
+  });
+
+  runAppend(view);
+
+  equal(view.$().text(), 'my custom link-to component', 'rendered a custom-link-to component');
+});
+
 }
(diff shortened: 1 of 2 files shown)


Serabe:
OTOH, I think the behaviour for triple curlies in `link-to` should be deprecated given:

* It is a hack not easily replicable in the component itself.
* Makes `link-to` behave differently from other components.
* That behaviour cannot be ported easily to all components extending from `link-to`.
* It can be easily transformed from `{{{link-to title route}}}` to `{{#link-to route}} {{{title}}} {{/link-to}}`.

rwjblue:
Ultimately, I'd like to deprecate this inline form for custom `Ember.LinkComponent` classes (and possibly even for all inline `link-to`), so leaving the AST transform for `{{link-to 'thing' 'thing'}}` but bringing back support in `packages/ember-routing-views/lib/components/link-to.js` allows us to support the old behavior and still deprecate custom `LinkComponent` inline form.

*tldr;* Leave the AST transform added in https://github.com/emberjs/ember.js/pull/12229 in, add the test you have added here, and revert the changes https://github.com/emberjs/ember.js/pull/12229 made to `packages/ember-routing-views/lib/components/link-to.js`.

Serabe:
Updated! Thank you for the quick response!

rwjblue:
Awesome, this LGTM.  We are having an issue with Sauce Labs ATM (they aren't recognizing one of the Firefox builds).  I'll try to dig into that later today.

rwjblue:
https://github.com/emberjs/ember.js/pull/13439 should have fixed the build, thanks for rebasing so quickly

rwjblue:
Thanks @Serabe!

PR #13440, opened 2016-05-02
Cleanup Ember.Test, Ember.testing, Ember.onerror

krisselden:


kiwiupover:
Looks great @krisselden one more step towards tree shaking.

packages/ember-metal/tests/run_loop/onerror_test.js
```diff
@@ -20,11 +20,11 @@ QUnit.test('With Ember.onerror set, errors in Ember.run are caught', function ()
...
-  Ember.onerror = function(error) { caught = error; };
+  setOnerror(error => { caught = error; });
 
-  run(function() { throw thrown; });
+  run(() => { throw thrown; });
 
   deepEqual(caught, thrown);
 
-  Ember.onerror = undefined;
+  setOnerror(undefined);
```
rwjblue: I realize that this was already like this, but we should probably reset in a tear down in case the test dies.

packages/ember-runtime/lib/ext/rsvp.js
```diff
@@ -54,32 +31,39 @@ export function onerrorDefault(reason) {
...
-        throw error;
-      }
-    } else if (Ember.onerror) {
-      Ember.onerror(error);
-    } else {
-      Logger.error(error.stack);
-    }
+    dispatchError(error);
   }
+  // let onerror = getOnerror();
```
rwjblue: We should kill this commented code.

packages/ember-runtime/lib/ext/rsvp.js
```diff
@@ -54,32 +31,39 @@ export function onerrorDefault(reason) {
...
-export function after (cb) {
-  run.schedule(run.queues[run.queues.length - 1], cb);
+export function after(cb) {
+  run.backburner.schedule('rsvpAfter', null, cb);
 }
 
 RSVP.on('error', onerrorDefault);
 RSVP.configure('after', after);
 
+run._addQueue('rsvpAfter', 'destroy');
```
rwjblue: Can we use one of our fake symbols here instead of `rsvpAfter`?  I would hate to have people hook into this themselves, we should make it super clear that it is private...

packages/ember-runtime/lib/ext/rsvp.js
```diff
@@ -54,32 +31,39 @@ export function onerrorDefault(reason) {
...
-        Logger.error(error.stack);
-      } else {
-        throw error;
-      }
-    } else if (Ember.onerror) {
-      Ember.onerror(error);
-    } else {
-      Logger.error(error.stack);
-    }
+    dispatchError(error);
```
stefanpenner: I''ll try to review this code path, is it functionally equivalent just refactored or should I look out for stuff.

packages/ember-runtime/lib/ext/rsvp.js
stefanpenner: we could also just do this in after, which happens after all the queues flush.

packages/ember-runtime/lib/ext/rsvp.js
rwjblue: @stefanpenner - It is largely the same path, but instead of guarding all over for `Ember.Test` (for debug versions) we just register an `dispatchErrorOverride` method when `ember-testing` package loads.

packages/ember-runtime/lib/ext/rsvp.js
```diff
@@ -1,85 +1,57 @@
...
-var asyncEnd = function() {
-  if (Ember.Test && Ember.Test.adapter) {
-    Ember.Test.adapter.asyncEnd();
-  }
-};
-
-RSVP.configure('async', function(callback, promise) {
-  var async = !run.currentRunLoop;
+const backburner = run.backburner;
+run._addQueue('rsvpAfter', 'destroy');
```
rwjblue: Should we use a private symbol for this? I'd hate for folks to try and schedule into our special queue...

rwjblue:
@krisselden - This looks very good to me.  I would like to test ember-qunit with these changes to make sure things work properly.  I should be able to finish that tonight or first thing tomorrow morning...

krisselden:
@rwjblue anything I can do to help verify ember-qunit? Would it just be running its test suite against this?  I also will update if issues come to light.

PR #13441, opened 2016-05-02
[Glimmer2] Mark glimmer2 tests on context.isTruthy to be @htmlbars only

Joelkang:
Fixes the FIXME introduced in https://github.com/emberjs/ember.js/pull/13332 

cc @chancancode

### DIFF
--- a/packages/ember-glimmer/tests/integration/syntax/with-test.js
+++ b/packages/ember-glimmer/tests/integration/syntax/with-test.js
@@ -167,13 +167,9 @@ moduleFor('Syntax test: {{#with as}}', class extends TogglingSyntaxConditionalsT
 
     this.assertText('Yehuda Katz');
 
-    this.runTask(() => set(this.context, 'proxyThing.isTruthy', false));
+    this.runTask(() => set(this.context, 'proxyThing.content', { name: 'Godfrey Chan' }));
 
-    this.assertText('');
-
-    this.runTask(() => set(this.context, 'proxyThing.name', 'Godfrey Chan'));
-
-    this.assertText('');
+    this.assertText('Godfrey Chan');
 
     this.runTask(() => set(this.context, 'proxyThing', ObjectProxy.create({ content: { name: 'Tom Dale' } })));
 
--- a/packages/ember-glimmer/tests/integration/syntax/with-test.js
+++ b/packages/ember-glimmer/tests/integration/syntax/with-test.js
@@ -4,6 +4,7 @@ import { A as emberA } from 'ember-runtime/system/native_array';
 import { moduleFor, RenderingTest } from '../../utils/test-case';
 import { TogglingSyntaxConditionalsTest } from '../../utils/shared-conditional-tests';
 import { strip } from '../../utils/abstract-test-case';
+import ObjectProxy from 'ember-runtime/system/object_proxy';
 
 moduleFor('Syntax test: {{#with}}', class extends TogglingSyntaxConditionalsTest {
 
@@ -153,7 +154,7 @@ moduleFor('Syntax test: {{#with as}}', class extends TogglingSyntaxConditionalsT
 
   ['@test can access alias of a proxy']() {
     this.render(`{{#with proxyThing as |person|}}{{person.name}}{{/with}}`, {
-      proxyThing: { isTruthy: true, name: 'Tom Dale' }
+      proxyThing: ObjectProxy.create({ content: { name: 'Tom Dale' } })
     });
 
     this.assertText('Tom Dale');
@@ -174,7 +175,7 @@ moduleFor('Syntax test: {{#with as}}', class extends TogglingSyntaxConditionalsT
 
     this.assertText('');
 
-    this.runTask(() => set(this.context, 'proxyThing', { isTruthy: true, name: 'Tom Dale' }));
+    this.runTask(() => set(this.context, 'proxyThing', ObjectProxy.create({ content: { name: 'Tom Dale' } })));
 
     this.assertText('Tom Dale');
   }
--- a/packages/ember-glimmer/tests/utils/shared-conditional-tests.js
+++ b/packages/ember-glimmer/tests/utils/shared-conditional-tests.js
@@ -351,13 +351,13 @@ export const ObjectTestCases = {
     });
 
     this.assertText('T1T2F3');
-  }
+  },
 
   ['@test setting the isTruthy property of an object proxy overrides its truthiness regardless of its content']() {
     this.renderValues(
       ObjectProxy.create({ content: {} }),
       EmberObject.create({ content: false, isTruthy: true }),
-      ObjectProxy.create()
+      ObjectProxy.create({ content: null })
     );
 
     this.assertText('T1T2F3');
@@ -390,9 +390,9 @@ export const ObjectTestCases = {
     this.assertText('T1T2T3');
 
     this.runTask(() => {
-      set(this.context, 'cond1.content', undefined);
-      set(this.context, 'cond2.content', undefined);
-      set(this.context, 'cond3.content', undefined);
+      set(this.context, 'cond1.content', null);
+      set(this.context, 'cond2.content', null);
+      set(this.context, 'cond3.content', null);
     });
 
     this.assertText('T1T2F3');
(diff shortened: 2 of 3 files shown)


chadhietala:
I actually think we need to update the tests to use a real proxy. This is still just testing a pojo. https://github.com/emberjs/ember.js/pull/13441/files#diff-f070d41054fd84b7bc5a98e949ed81dbR170

There are other places where we are testing pojos with `isTruthy`.

chadhietala:
This should also be a proxy https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/tests/utils/shared-conditional-tests.js#L477

Joelkang:
Fixed the 2nd thing, there actually are already tests for proxies themselves https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/tests/utils/shared-conditional-tests.js#L317 https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/tests/utils/shared-conditional-tests.js#L424.

I'm checking the other files for isTruthy looks like i missed out a couple

packages/ember-glimmer/tests/integration/syntax/with-test.js
```diff
@@ -174,7 +175,7 @@ moduleFor('Syntax test: {{#with as}}', class extends TogglingSyntaxConditionalsT
```
chancancode: can you change line 170 to set `proxyThing.content`? We should not be setting `isTruthy` directly (and I guess line 174 doesn't make sense anymore)

chancancode: can you change line 170 to set `proxyThing.content`? We should not be setting `isTruthy` directly (and I guess line 174 doesn't make sense anymore)

packages/ember-glimmer/tests/utils/shared-conditional-tests.js
```diff
@@ -349,6 +351,59 @@ export const ObjectTestCases = {
     });
 
     this.assertText('T1T2F3');
+  },
+
+  ['@test setting the isTruthy property of an object proxy overrides its truthiness regardless of its content']() {
+    this.renderValues(
+      ObjectProxy.create({ content: {} }),
+      EmberObject.create({ content: false, isTruthy: true }),
```
chancancode: wow, EmberObject is a proxy ?________?

packages/ember-glimmer/tests/utils/shared-conditional-tests.js
```diff
@@ -349,6 +351,59 @@ export const ObjectTestCases = {
...
+
+    this.assertText('T1T2F3');
+
+    this.runTask(() => this.rerender());
+
+    this.assertText('T1T2F3');
+
+    this.runTask(() => {
+      set(this.context, 'cond1.isTruthy', false);
+      set(this.context, 'cond2.isTruthy', false);
```
chancancode: ditto – we should not be setting `isTruthy` on the proxy, just change/set the content

packages/ember-glimmer/tests/utils/shared-conditional-tests.js
```diff
@@ -585,7 +642,8 @@ export class TogglingHelperConditionalsTest extends TogglingConditionalsTest {
     this.assertText('T1F2');
   }
 
-  ['@test it tests for `isTruthy` on the context if available']() {
+  // This no longer makes sense for glimmer since `this` refers to the component, which cannot be a proxy
+  ['@htmlbars it tests for `isTruthy` on the context if available']() {
```
chancancode: Do we have another shared/generic test that tests changing the content of the proxy?

packages/ember-glimmer/tests/utils/shared-conditional-tests.js
```diff
@@ -662,7 +720,8 @@ export class TogglingSyntaxConditionalsTest extends TogglingConditionalsTest {
     this.assertText('T1F2');
   }
 
-  ['@test it tests for `isTruthy` on the context if available']() {
+  // Marking as @htmlbars since this no longer relevant for glimmer as `this` refers to the component, which is never a proxy
+  ['@htmlbars it tests for `isTruthy` on the context if available']() {
```
chancancode: ditto, do we have another shared/generic test that tests changing the content of the proxy?

chancancode:
Thank you for picking this up and sorry for the delay! Left some comments on the diffs. Looking good!

packages/ember-glimmer/tests/utils/shared-conditional-tests.js
```diff
@@ -349,6 +351,59 @@ export const ObjectTestCases = {
...
+      set(this.context, 'cond1.content', 1);
+      set(this.context, 'cond2.content', 1);
+      set(this.context, 'cond3.content', 1);
+    });
+
+    this.assertText('F1F2T3');
+
+
+    this.runTask(() => {
+      set(this.context, 'cond1.isTruthy', true);
```
krisselden: why are we testing setting isTruthy directly?

packages/ember-glimmer/tests/utils/shared-conditional-tests.js
```diff
@@ -349,6 +351,59 @@ export const ObjectTestCases = {
     });
 
     this.assertText('T1T2F3');
+  },
+
+  ['@test setting the isTruthy property of an object proxy overrides its truthiness regardless of its content']() {
```
krisselden: Is this a test we actually want? /cc @wycats

This test seems terrible. At best this is intimate API (private API that has become popular to abuse) and should be clearly marked as such.

Joelkang: yea so i too went back and forth on this internally. it seems like a very specious test, but this specific test case functioned as a port of an existing test (at least in spirit, if not verbatim). I'm happy to remove it altogether.

packages/ember-glimmer/tests/utils/shared-conditional-tests.js
Joelkang: See my comment below to @krisselden's point about this particular test case. 

The htmlbars test that this was replacing seemed to be testing the truthiness of a non-proxy object that masqueraded as a proxy via the isTruthy property. I didn't know if we wanted to just remove it completely since we would theoretically be removing test coverage (although for what exactly I cannot say) from htmlbars, so I ported the spirit of the test over with these strange cases.

packages/ember-glimmer/tests/utils/shared-conditional-tests.js
Joelkang: Same as above

packages/ember-glimmer/tests/utils/shared-conditional-tests.js
Joelkang: The ArrayTestCases and ObjectTestCases do provide coverage for proxy content changing, and are mixed into the `TogglingConditionalsTest`.

`with-tests.js` extends `TogglingSyntaxConditionalsTest` which in turn is an extension of `TogglingConditionalsTest`. `each-tests.js` explicitly mixes in the ArrayTestCases so i'd say we're covered

Joelkang:
Rebased to fix CI issue. @krisselden @chancancode let me know if my comments make sense and if we should remove that last test. I also think that the array/objectTestCases should cover the necessary proxy conditional logic, but let me know if im missing something

packages/ember-glimmer/tests/utils/shared-conditional-tests.js
krisselden: This is likely testing observable behavior that was unintentional.  Likely it would have been better to have made this a symbol or meta.  I don't think we should block on this since it is just a port of an existing test.

packages/ember-glimmer/tests/utils/shared-conditional-tests.js
```diff
@@ -283,7 +284,8 @@ export const ObjectTestCases = {
     this.assertText('T1F2');
   },
 
-  ['@test it tests for `isTruthy` on Ember objects if available']() {
+  // Marking as @htmlbars since POJOs can no longer masquerade as Proxies in glimmer2. See proxy tests below.
+  ['@htmlbars it tests for `isTruthy` on Ember objects if available']() {
```
rwjblue: If this is not supported behavior (and that is basically what we are saying), we should just remove this test.

Joelkang: Is it not problematic to remove code coverage for legacy behaviour though?

packages/ember-glimmer/tests/utils/shared-conditional-tests.js
Joelkang: I do believe though, @rwjblue, that this whole test should just be removed.

chancancode:
Merged in #13674, thanks @Joelkang!

Issue #13318, opened 2016-04-12
Deprecation of Handlebars.SafeString prevents clear checks if something is a SafeString


dustyjewett:
With the deprecation of Handlebars.SafeString, the constructor chain of SafeString doesn't include Handlebars.SafeString, making instanceof checks fail.

See Twiddle: https://ember-twiddle.com/32e5e3818cd66aa224f5a422fdafd83a?openFiles=controllers.application.js%2C

For context, the latest change in the chain, fixing the deprecation:
https://github.com/emberjs/ember.js/pull/13226/files

rwjblue:
Can you help me understand why you need to check if a given string is a SafeString?

dustyjewett:
We were using this in a test, verifying that our helper returned a SafeString.

topaxi:
I have some library code that uses this to emit a warning when an unsafe string was passed.

https://github.com/topaxi/ember-custom-css-properties/blob/master/addon/mixins/custom-css-properties.js#L67

rwjblue:
@dustyjewett - Wouldn't it be better to just use the helper, and ensure its return value was displayed as HTML?

```
moduleForComponent('asdf', { integration: true });

test('renders html', function(assert) {
  this.render(hbs`{{some-helper}}`);

  assert.equal(this.$('.some-selector-for-html-values-from-helper').length, 1, 'found html from helper');
});
```

rwjblue:
@topaxi - We generally use a `toHTML` function in the HTMLBars internals to determine if a thing is a safe string.  This *may* change with Glimmer (I believe a symbol is used in Glimmer), but that is still TBD.

dustyjewett:
@rwjblue We certainly test that our end result is as-expected, but because SafeString has an explicit impact in how helpers are rendered into the DOM, we want to check that as well. 

To get around this issue, we verify that the return object is NOT a string, but that the .toString() is as-expected. But that is just testing known ways that things might go wrong, we can't check the actual, expected behavior. 

(also, many of our objects have toHTML methods, so checking that won't ensure that we didn't accidentally just return one of our other classes)

pixelhandler:
@dustyjewett sounds like the deprecation has a side-effect if you expect to check the instance, `new Ember.Handlebars.SafeString('string') instanceof Ember.Handlebars.SafeString;`

I'm curious how hard would it be in your app to make the suggested change to `Ember.String.htmlSafe`. Also would using that allow you to use `instanceof` again?

[...]

PR #13442, opened 2016-05-03
[BUGFIX beta] Remove SafeString deprecation.

rwjblue:
See #13318 for details.

To bring this back, we need to have implemented https://github.com/emberjs/rfcs/issues/139.


workmanw:
@rwjblue Thanks!

PR #13446, opened 2016-05-04
Show a better error message for invalid deep nested @each computed properties

cesarizu:
Currently when using deep nested properties the error presented to the developer is generic and doesn't give any clue on what computed property is causing it like:

`WARNING: Dependent keys containing @each only work one level deep. You cannot use nested forms like todos.@each.owner.name or todos.@each.owner.@each.name. Please create an intermediary computed property.`

This PR adds the property to the message so it appears like:

`WARNING: Dependent keys containing @each only work one level deep. You cannot use nested forms like todos.@each.owner.name or todos.@each.owner.@each.name. You used: some.@each.nested.property . Please create an intermediary computed property.`

### DIFF
--- a/packages/ember-metal/lib/computed.js
+++ b/packages/ember-metal/lib/computed.js
@@ -250,7 +250,8 @@ ComputedPropertyPrototype.property = function() {
     warn(
       `Dependent keys containing @each only work one level deep. ` +
         `You used the key "${property}" which is invalid. ` +
-          `Please create an intermediary computed property.`,
+          `You used: ${property} . ` +
+            `Please create an intermediary computed property.`,
       DEEP_EACH_REGEX.test(property) === false,
       { id: 'ember-metal.computed-deep-each' }
     );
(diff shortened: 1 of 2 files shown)


Serabe:
Test are not passing. Can you fix them, please?

cesarizu:
@Serabe I've rebased the branch and fixed the tests.

locks:
Do we need the examples if we're telling the user what path they entered wrong?

cesarizu:
I think that the examples are useful the first time you see that message and they are not relevant afterwards. The important part is what's after the first `@each`. So, if you have `a.b.@each.c.d` the relevant part is only `c.d`. The `a.b` part is ok but that might not be obvious for a developer that encounters this problem for the first time.

A better error message would say something like:

```WARNING: Dependent keys containing @each only work one level deep. You used: some.@each.nested.property . Please replace "nested.property" with an intermediary computed property```

I haven't checked what are the exact cases where this error is presented and/or if the actual failure is available when printing the error.

cesarizu:
@homu this PR is obsolete after those changes. I'm closing it.

Issue #13444, opened 2016-05-04
getWithDefault broken with chained properties


jamesarosen:
Let's explore how `get(...) || y` and `getWithDefault(x, y)` compare.

```js
> o = Ember.Object.create()
> o.get('foo') || 'x'
"x"
> o.getWithDefault('foo', 'x')
"x"
```
👍 OK, that sounds totally right.

```js
> o.set('foo', 'bar')
> o.get('foo') || 'x'
"bar"
> o.getWithDefault('foo', 'x')
"bar"
```
👍 Cool. Makes sense.

```js
> o = Ember.Object.extend({ foo: null }).create()
> o.get('foo') || 'x'
"x"
> o.getWithDefault('foo', 'x')
null
```
😕 Hmm. That obeys the [API docs](http://emberjs.com/api/#method_getWithDefault) but it doesn't seem super intuitive. [Others agree with me, but there's also good reason to keep it as-is](https://github.com/emberjs/ember.js/pull/13056). In our app, we use `foo: null` as an indication that the client should pass `foo` in; maybe we should change that to `foo: undefined`. Oh well, let's press on.
```js
> o = Ember.Object.create()
> o.get('foo.bar')
undefined
> o.getWithDefault('foo.bar', 'x')
"x"
```
👍 Diggin' it.
```js
> o = Ember.Object.extend({ foo: null }).create()
> o.get('foo.bar') || 'x'
"x"
> o.getWithDefault('foo.bar', 'x')
null
```
WHAT!?!?!?!??!!? :trollface:

mmun:
This is a bug in Ember.get:

```js
Ember.get({ foo: null }, 'bar') // null
```

It should be `undefined`.

jamesarosen:
@mmun on Ember 1.13, I get
```js
> Ember.get({ foo: null }, 'bar')
undefined
```
which is correct as far as I can tell.

mmun:
Oops. I was checking using the console on http://discuss.emberjs.com/ but it seems to be running 1.12 :P

mmun:
You want `foo.bar` there, btw.

mmun:
I just checked it against canary and `Ember.get({ foo: null }, 'foo.bar') === null` still.

jamesarosen:
> You want foo.bar there, btw.

Ah!

robbiepitts:
I have a failing test for this. I'll start a WIP PR.

jamesarosen:
[...]

PR #13449, opened 2016-05-04
[BUGFIX beta] `Ember.get` with first part of path chain resolving to `null` returns `null` rather than `undefined`

pittst3r:
Addresses issue #13444.

As is:

``` js
Ember.get({ foo: null }, 'foo.bar') === null
```

Expected:

``` js
Ember.get({ foo: null }, 'foo.bar') === undefined
```

This bug results in:

``` js
let o = Ember.Object.extend({ foo: null }).create();
o.get('foo.bar') || 'x' === 'x' 
o.getWithDefault('foo.bar', 'x') === null // wat
```

### DIFF
--- a/packages/ember-metal/lib/property_get.js
+++ b/packages/ember-metal/lib/property_get.js
@@ -76,7 +76,7 @@ export function _getPath(root, path) {
 
   for (let i = 0; i < parts.length; i++) {
     if (obj == null) {
-      return obj;
+      return undefined;
     }
 
     obj = get(obj, parts[i]);
--- a/packages/ember-runtime/tests/legacy_1x/mixins/observable/observable_test.js
+++ b/packages/ember-runtime/tests/legacy_1x/mixins/observable/observable_test.js
@@ -129,8 +129,8 @@ QUnit.test('should return null when property value is null on Ember.Observable',
 });
 
 QUnit.test('should call unknownProperty when value is undefined on Ember.Observable', function() {
-  equal(get(object, 'unknown'), 'unknown');
-  equal(object.lastUnknownProperty, 'unknown');
+  equal(get(objectA, 'unknown'), 'unknown');
+  equal(objectA.lastUnknownProperty, 'unknown');
 });
 
 QUnit.test('should get normal properties on standard objects', function() {
--- a/packages/ember-metal/tests/accessors/get_path_test.js
+++ b/packages/ember-metal/tests/accessors/get_path_test.js
@@ -18,7 +18,8 @@ var moduleOpts = {
       emptyString: '',
       Wuz: {
         nar: 'foo'
-      }
+      },
+      nullValue: null
     };
   },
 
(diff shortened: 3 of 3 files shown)


krisselden:
Good catch

packages/ember-runtime/tests/legacy_1x/mixins/observable/observable_test.js
```diff
@@ -129,8 +129,8 @@ QUnit.test('should return null when property value is null on Ember.Observable',
 });
 
 QUnit.test('should call unknownProperty when value is undefined on Ember.Observable', function() {
-  equal(get(object, 'unknown'), 'unknown');
```
robbiepitts: This would fail if only the `Ember.get()` module was run and relied on the `object.get()` module being run first to pass—for obvious reasons.

robbiepitts:
K, ready for review!

mmun:
Looks good!

@rwjblue @krisselden Does this need a canary cycle or can we put it into beta?

packages/ember-metal/lib/property_get.js
```diff
@@ -76,7 +76,7 @@ export function _getPath(root, path) {
 
   for (let i = 0; i < parts.length; i++) {
     if (obj == null) {
```
krisselden: we only want to call get on object like things, this should do if (!isObjectLike(obj)) { return undefined; }

krisselden:
This change is low risk.  I think it is bad that we let random truthy values get to get() that aren't gettable too, not just that we return null sometimes.

robbiepitts:
@krisselden Agreed. Would you like that addressed in this PR? Alternatively I can create an issue and work on it as a separate thing.

packages/ember-metal/lib/property_get.js
krisselden: To be specific, I would check falsy, and that the typeof was 'string', 'object', 'function' but at a minimum this can simply check falsy.  I'm sure the `obj == null` check was a bad fix to this sometimes returning false vs undefined.

krisselden:
@robbiepitts at least in this PR reduce it to just a falsy check.

packages/ember-runtime/tests/legacy_1x/mixins/observable/observable_test.js
rwjblue: Ugh.  Good catch!

rwjblue:
Seems like `[BUGFIX beta]` to me.

packages/ember-metal/lib/property_get.js
mmun: Falsey alone won't work because of empty strings, e.g.

`Ember.get({ foo: "" }, 'foo.length')` should be 0 not undefined. (Can you add a test for this one @robbiepitts?)

krisselden: Confirm, using get for string.length seems super lame.

mixonic:
👍

PR #13451, opened 2016-05-04
Element Actions for Glimmer

zackthehuman:
Port existing element actions code into Glimmer.

### DIFF
--- a/packages/ember-glimmer/lib/modifiers/action.js
+++ b/packages/ember-glimmer/lib/modifiers/action.js
@@ -1,0 +1,193 @@
+import { assert } from 'ember-metal/debug';
+import run from 'ember-metal/run_loop';
+import { uuid } from 'ember-metal/utils';
+import { isSimpleClick } from 'ember-views/system/utils';
+import ActionManager from 'ember-views/system/action_manager';
+
+const MODIFIERS = ['alt', 'shift', 'meta', 'ctrl'];
+const POINTER_EVENT_TYPE_REGEX = /^click|mouse|touch/;
+
+function isAllowedEvent(event, allowedKeys) {
+  if (allowedKeys === null || typeof allowedKeys === 'undefined') {
+    if (POINTER_EVENT_TYPE_REGEX.test(event.type)) {
+      return isSimpleClick(event);
+    } else {
+      allowedKeys = '';
+    }
+  }
+
+  if (allowedKeys.indexOf('any') >= 0) {
+    return true;
+  }
+
+  for (let i = 0; i < MODIFIERS.length; i++) {
+    if (event[MODIFIERS[i] + 'Key'] && allowedKeys.indexOf(MODIFIERS[i]) === -1) {
+      return false;
+    }
+  }
+
+  return true;
+}
+
+export let ActionHelper = {
+  // registeredActions is re-exported for compatibility with older plugins
+  // that were using this undocumented API.
+  registeredActions: ActionManager.registeredActions,
+
+  registerAction(actionState) {
+    let { actionId } = actionState;
+    let actions = ActionManager.registeredActions[actionId];
+
+    if (!actions) {
+      actions = ActionManager.registeredActions[actionId] = [];
+    }
+
+    actions.push(actionState);
+
+    return actionId;
+  },
+
+  unregisterAction(actionState) {
+    let { actionId } = actionState;
+    let actions = ActionManager.registeredActions[actionId];
+
+    let index = actions.indexOf(actionState);
+
+    if (index !== -1) {
+      actions.splice(index, 1);
+    }
+
+    if (actions.length === 0) {
+      delete ActionManager.registeredActions[actionId];
+    }
+  }
+};
+
+export class ActionState {
+  constructor(actionId, actionName, actionArgs, namedArgs, dynamicScope) {
+    this.actionId = actionId;
+    this.actionName = actionName;
+    this.actionArgs = actionArgs;
+    this.namedArgs = namedArgs;
+    this.dynamicScope = dynamicScope;
+    this.eventName = this.getEventName();
+  }
+
+  getEventName() {
+    return this.namedArgs.get('on').value() || 'click';
+  }
+
+  getActionArgs() {
+    let result = new Array(this.actionArgs.length);
+
+    for (let i = 0; i < this.actionArgs.length; i++) {
+      result[i] = this.actionArgs[i].value();
+    }
+
+    return result;
+  }
+
+  getTarget() {
+    let { dynamicScope, namedArgs } = this;
+    let target;
+
+    if (namedArgs.has('target')) {
+      target = namedArgs.get('target').value();
+    } else {
+      target = dynamicScope.controller || dynamicScope.view;
+    }
+
+    return target;
+  }
+
+  handler(event) {
+    let { actionName, namedArgs } = this;
+    let bubbles = namedArgs.get('bubbles');
+    let preventDefault = namedArgs.get('preventDefault');
+    let allowedKeys = namedArgs.get('allowedKeys');
+    let target = this.getTarget();
+
+    if (!isAllowedEvent(event, allowedKeys.value())) {
+      return true;
+    }
+
+    if (preventDefault.value() !== false) {
+      event.preventDefault();
+    }
+
+    if (bubbles.value() === false) {
+      event.stopPropagation();
+    }
+
+    run(() => {
+      if (typeof actionName === 'function') {
+        actionName.apply(target, this.getActionArgs());
+        return;
+      }
+      if (target.send) {
+        target.send.apply(target, [actionName, ...this.getActionArgs()]);
+      } else {
+        assert(
+          `The action '${actionName}' did not exist on ${target}`,
+          typeof target[actionName] === 'function'
+        );
+
+        target[actionName].apply(target, this.getActionArgs());
+      }
+    });
+  }
+
+  destroy() {
+    ActionHelper.unregisterAction(this);
+  }
+}
+
+// implements ModifierManager<Action>
+export default class ActionModifierManager {
+  install(element, args, dom, dynamicScope) {
+    let { named, positional } = args;
+    let actionName;
+
+    if (positional.length > 0) {
+      actionName = positional.at(0).value();
+    }
+
+    assert(
+      'You specified a quoteless path to the {{action}} helper ' +
+      'which did not resolve to an action name (a string). ' +
+      'Perhaps you meant to use a quoted actionName? (e.g. {{action \'save\'}}).',
+      typeof actionName === 'string' || typeof actionName === 'function'
+    );
+
+    let actionArgs = [];
+    for (let i = 1; i < positional.length; i++) {
+      actionArgs.push(positional.at(i));
+    }
+
+    let actionId = uuid();
+    let actionState = new ActionState(actionId, actionName, actionArgs, named, dynamicScope);
+
+    ActionHelper.registerAction(actionState);
+
+    dom.setAttribute(element, 'data-ember-action', '');
+    dom.setAttribute(element, `data-ember-action-${actionId}`, actionId);
+
+    return actionState;
+  }
+
+  update(modifier, element, args, dom, dynamicScope) {
+    let { positional } = args;
+
+    modifier.actionName = positional.at(0).value();
+    modifier.dynamicScope = dynamicScope;
+    modifier.eventName = modifier.getEventName();
+
+    // Not sure if this is needed? If we mutate the actionState is that good enough?
+    ActionHelper.unregisterAction(modifier);
+    ActionHelper.registerAction(modifier);
+  }
+
+  getDestructor(modifier) {
+    return modifier;
+  }
+}
--- a/packages/ember-glimmer/lib/modifiers/action.js
+++ b/packages/ember-glimmer/lib/modifiers/action.js
@@ -64,12 +64,12 @@ export let ActionHelper = {
 };
 
 export class ActionState {
-  constructor(actionId, actionName, actionArgs, namedArgs, dynamicScope) {
+  constructor(actionId, actionName, actionArgs, namedArgs, implicitTarget) {
     this.actionId = actionId;
     this.actionName = actionName;
     this.actionArgs = actionArgs;
     this.namedArgs = namedArgs;
-    this.dynamicScope = dynamicScope;
+    this.implicitTarget = implicitTarget;
     this.eventName = this.getEventName();
   }
 
@@ -88,13 +88,13 @@ export class ActionState {
   }
 
   getTarget() {
-    let { dynamicScope, namedArgs } = this;
+    let { implicitTarget, namedArgs } = this;
     let target;
 
     if (namedArgs.has('target')) {
       target = namedArgs.get('target').value();
     } else {
-      target = dynamicScope.controller || dynamicScope.view;
+      target = implicitTarget.value();
     }
 
     return target;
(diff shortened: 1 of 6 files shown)


chadhietala:
Are there some tests that can now be ran in both Glimmer and HTMLBars?

zackthehuman:
@chadhietala Currently there aren't any in `ember-glimmer` (so no INRU tests, etc.). I have some uncommitted work where I started porting over existing tests from `ember.js/packages/ember-routing-htmlbars/tests/helpers/element_action_test.js`. I can add them to this PR once they're ported.

krisselden:
Looks good, will merge when we have a test in place

packages/ember-glimmer/lib/modifiers/action.js
```diff
@@ -0,0 +1,193 @@
+import { assert } from 'ember-metal/debug';
+import run from 'ember-metal/run_loop';
+import { uuid } from 'ember-metal/utils';
+import { isSimpleClick } from 'ember-views/system/utils';
+import ActionManager from 'ember-views/system/action_manager';
+
+let MODIFIERS = ['alt', 'shift', 'meta', 'ctrl'];
```
stefanpenner: const

packages/ember-glimmer/lib/modifiers/action.js
```diff
@@ -0,0 +1,193 @@
...
+let POINTER_EVENT_TYPE_REGEX = /^click|mouse|touch/;
```
stefanpenner: const

zackthehuman:
I ported most of the existing element action tests. There are a few tests that I stubbed out and marked with `@skip` because I'm either not sure how they fit into the ember-glimmer test harness or think they should be cut because they test internals. Would like to get feedback on that aspect.

packages/ember-glimmer/tests/integration/helpers/element-action-test.js
```diff
@@ -0,0 +1,1340 @@
...
+    this.assert.ok(button.attr('data-ember-action').match(''), 'An empty data-ember-action attribute was added');
+    this.assert.ok(attributes[0].match(/data-ember-action-\d+/), 'A data-ember-action-xyz attribute with a guid was added');
+  }
+
+  ['@skip it should by default register a click event']() {
+    // TODO
+    // Should we be testing implementation details? Why?
+  }
+
+  ['@skip it should allow alternative events to be handled']() {
```
rwjblue: I don't think testing of `{{action 'foo' on='change'}}` is an implementation detail. Changing the event that is listened to is public API that should be tested.

packages/ember-glimmer/tests/integration/helpers/element-action-test.js
```diff
@@ -0,0 +1,1340 @@
...
+    this.render('<button {{action "show"}}>me the money</button>');
+
+    let button = this.$('button');
+    let attributes = getActionAttributes(button.get(0));
+
+    this.assert.ok(button.attr('data-ember-action').match(''), 'An empty data-ember-action attribute was added');
+    this.assert.ok(attributes[0].match(/data-ember-action-\d+/), 'A data-ember-action-xyz attribute with a guid was added');
+  }
+
+  ['@skip it should by default register a click event']() {
```
rwjblue: This is already implicitly tested by both `'@test it can call an action on its enclosing component'` and `@test it can call an action with parameters`.

packages/ember-glimmer/tests/integration/helpers/element-action-test.js
```diff
@@ -0,0 +1,1340 @@
...
+    // TODO
+    // Should we be testing implementation details? Why?
+  }
+
+  ['@skip it should allow alternative events to be handled']() {
+    // TODO
+    // Should we be testing implementation details? Why?
+  }
+
+  ['@skip it should by default target the view\'s controller']() {
```
rwjblue: 🔥

packages/ember-glimmer/tests/integration/helpers/element-action-test.js
```diff
@@ -0,0 +1,1340 @@
...
+    // TODO
+    // Should we be testing implementation details? Why?
+  }
+
+  ['@skip inside a yield, the target points at the original target']() {
```
rwjblue: This is important to cover.  Given the following:

```hbs
{{! app/components/some-foo.hbs }}
{{#some-other}}
  <button {{action 'foo'}}>Click me</button>
{{/some-other}}
```

We should definitely assert that when clicking that button, it should target `some-foo`.

packages/ember-glimmer/tests/integration/helpers/element-action-test.js
```diff
@@ -0,0 +1,1340 @@
...
+
+    let TargetComponent = Component.extend({
+      actions: {
+        wat() {
+          targetWatted = true;
+        }
+      }
+    });
+
+    let target = TargetComponent.create();
```
rwjblue: I would prefer to avoid manually calling `.create` on components.  One possible alternative to this would be:

```js
this.registerComponent('parent-target', {
  ComponentClass: TargetComponent,
  template: '{{yield this}}'
});

// ...snip...
this.render(`{{#parent-target as |parent|}}{{other-component anotherTarget=parent}}{{/parent-target}}`);
```

packages/ember-glimmer/tests/integration/helpers/element-action-test.js
```diff
@@ -0,0 +1,1340 @@
...
+
+    this.assert.deepEqual(previousAttributes, rerenderedAttributes, 'the same action id is used across rerenders');
+    this.assert.notDeepEqual(previousAttributes, newAttributes, 'the action id is regenerated when the action is torn down');
+  }
+
+  ['@skip it should unregister event handlers on rerender']() {
+    // TODO: Port this test for HTMLBars only.
+  }
+
+  ['@skip it should unregister event handlers on inside virtual views']() {
```
rwjblue: While the concept of virtual views is definitely not a thing any more, we should probably assert that actions rendered within an `{{#each` should be cleaned up.

packages/ember-glimmer/tests/integration/helpers/element-action-test.js
```diff
@@ -0,0 +1,1340 @@
...
+    // TODO
+  }
+
+  ['@test it should properly capture events on child elements of a container with an action']() {
```
rwjblue: I don't fully understand this test description. Can you explain and/or tweak?

packages/ember-glimmer/tests/integration/helpers/element-action-test.js
```diff
@@ -0,0 +1,1340 @@
...
+    let passedContext;
+
+    let aTarget = Component.extend({
+      actions: {
+        edit(context) {
+          passedTarget = this;
+          passedContext = context;
+        }
+      }
+    }).create();
```
rwjblue: Can we avoid manually creating components like this?

packages/ember-glimmer/tests/integration/helpers/element-action-test.js
```diff
@@ -0,0 +1,1340 @@
...
+    editHandlerWasCalled = false;
+
+    this.runTask(() => {
+      this.$('a').trigger('mouseover');
+    });
+
+    this.assert.notOk(editHandlerWasCalled, 'the event handler was not called on mouseover');
+  }
+
+  ['@skip should unwrap controllers passed as a context']() {
```
rwjblue: I think that we don't support this any more (because proxying controllers are removed).

packages/ember-glimmer/tests/integration/helpers/element-action-test.js
```diff
@@ -0,0 +1,1340 @@
...
+  ['@test it should allow multiple contexts to be specified']() {
+    let passedContexts;
+    let models = [EmberObject.create(), EmberObject.create()];
+
+    let ExampleComponent = Component.extend({
+      modelA: models[0],
+      modelB: models[1],
+      actions: {
+        edit() {
+          passedContexts = [].slice.call(arguments);
```
rwjblue: Would using something like this be better?

```
edit(...args) {
  passedContexts = args;
}
```

packages/ember-glimmer/tests/integration/helpers/element-action-test.js
```diff
@@ -0,0 +1,1340 @@
...
+
+  ['@test it should allow multiple contexts to be specified mixed with string args']() {
+    let passedContexts;
+    let model = EmberObject.create();
+
+    let ExampleComponent = Component.extend({
+      model: model,
+      actions: {
+        edit() {
+          passedContexts = [].slice.call(arguments);
```
rwjblue: Maybe use rest args instead of slice?

packages/ember-glimmer/tests/integration/helpers/element-action-test.js
```diff
@@ -0,0 +1,1340 @@
...
+    }
+
+    testBoundAction('rock');
+    testBoundAction('paper');
+    testBoundAction('scissors');
+
+    this.assert.deepEqual(actionOrder, ['rock', 'paper', 'scissors'], 'action name was looked up properly');
+  }
+
+  ['@test a quoteless parameter should lookup actionName in context [DEPRECATED]']() {
```
rwjblue: I don't believe that this is deprecated...

packages/ember-glimmer/tests/integration/helpers/element-action-test.js
```diff
@@ -0,0 +1,1340 @@
...
+
+    testBoundAction('rock');
+    testBoundAction('paper');
+    testBoundAction('scissors');
+
+    this.assert.deepEqual(actionOrder, ['rock', 'paper', 'scissors'], 'action name was looked up properly');
+  }
+
+  ['@skip a quoteless string parameter should resolve actionName, including path']() {
+    // TODO: This seems to be covered by the test above. Seems like it should be cut.
```
rwjblue: Agreed, but use this test name ;P

packages/ember-glimmer/tests/integration/helpers/element-action-test.js
```diff
@@ -0,0 +1,1340 @@
...
+    this.assert.ok(clickActionWasCalled, 'the clicked action was called');
+
+    this.runTask(() => {
+      this.$('a').trigger('dblclick');
+    });
+
+    this.assert.ok(doubleClickActionWasCalled, 'the doubleClicked action was called');
+  }
+
+  ['@htmlbars allows multiple actions on a single element']() {
```
rwjblue: Seems like this, and the test above are roughly the same but have a small amount of differences.  It would be nice to either merge them, or somehow mark this `@htmlbars` as something that doesn't need to be "fixed" before enabling glimmer feature flag...

packages/ember-glimmer/tests/integration/helpers/element-action-test.js
```diff
@@ -0,0 +1,1340 @@
...
+    let event = jQuery.Event('click');
+
+    this.runTask(() => {
+      this.$('a').trigger(event);
+    });
+
+    this.assert.equal(event.isDefaultPrevented(), false, 'should not preventDefault');
+  }
+
+  ['@test it should respect preventDefault option if provided']() {
```
rwjblue: This test seems duplicated with the one above.

rwjblue:
I left a bunch of comments, but this is looking really good.  I a bit confused about the way multiple actions are stored, why do we need multiple attributes on the element?

zackthehuman:
@rwjblue The reason action data attributes are stored across multiple attributes in Glimmer is because they are assigned at render and we want to avoid reading attributes at render time. In HTMLBars the data attribute is read via `getAttribute` in order to be reused.

rwjblue:
As far as I understand it (haven't dug in too much):

* HTMLBars will read the attribute
  * if present it will reuse the same id. This results in one `getAttribute` for the second action.
  * if not present it will create a new id and `setAttribute`. This would be one `getAttribute` and one `setAttribute` (2 total operations) for first action on an element.
* This PR for Glimmer2 will always `setAttribute` twice.

In the existing HTMLBars case, we have the potential of using a single DOM operation whereas with this implementation for Glimmer we will always have two DOM operations.  In the worst case with HTMLBars today, there are still only 2 operations.  The primary difference between the two being that there is a `getAttribute` then a `setAttribute` with HTMLBars, and  two `setAttributes` with Glimmer.

Are we saying that two sets is better than a get and a set (and ignoring the somewhat unusual case of multiple action handlers being more optimized)?

zackthehuman:
@rwjblue My understanding from the reading of this code (https://github.com/emberjs/ember.js/blob/783a656e105a7121fc0dc38374931880d14b6bc3/packages/ember-routing-htmlbars/lib/keywords/element-action.js#L47) is that a `getAttribute` and a `setAttribute` will always take place, per action.

The approach I took for the Glimmer implementation is make the operation write-only -- it never performs a `getAttribute` at render time (or any other time, but the important aspect is that it doesn't take place during first render). The goal here is not necessarily to reduce DOM operations, but to eliminate reading from the DOM. @krisselden may have more input on this.

rwjblue:
> @rwjblue My understanding from the reading of this code (https://github.com/emberjs/ember.js/blob/783a656e105a7121fc0dc38374931880d14b6bc3/packages/ember-routing-htmlbars/lib/keywords/element-action.js#L47) is that a getAttribute and a setAttribute will always take place, per action.

You are correct, thank you for clarifying that!

> The goal here is not necessarily to reduce DOM operations, but to eliminate reading from the DOM.

Right, this is the heart of my question.  If the operations are the same, why jump through extra hoops just to avoid reading from the DOM.  I kinda assume that there is something fundamental that I am missing...

packages/ember-glimmer/tests/integration/helpers/element-action-test.js
zackthehuman: I guess what I'm getting at here is the original test for this intercepts/inspects the slots inside of `ActionManager.registeredActions`. I think a better way to test this would be to set up an action for several events, test each one, and assert that it fires for the respective event.

rwjblue: yep, agreed

packages/ember-glimmer/tests/integration/helpers/element-action-test.js
zackthehuman: Agreed. The existing test does this by probing the `ActionManager` which feels wrong. I'm not sure any other way to observe the cleanup without inspecting internals.

packages/ember-glimmer/tests/integration/helpers/element-action-test.js
zackthehuman: It doesn't seem like it. The original test name included `[DEPRECATED]` so I carried it along.

packages/ember-glimmer/tests/integration/helpers/element-action-test.js
zackthehuman: I'm gonna restore the existing test and mark as `@htmlbars` for now.

zackthehuman:
@rwjblue I believe I have addressed all of your test feedback. Take another look when you have a chance. As respects your `getAttribute` question -- it's something I was asked to avoid doing at render time. You may have to get more details from @wycats about why this is desirable.

rwjblue:
Yep, no problem. My comments RE avoiding the getAttribute first were most curiousity. 

I'll try to do another round of review tonight or tomorrow morning....

krisselden:
@zackthehuman can you rebase? also the just use master's package.json, when you rebase master is 'ours'

krisselden:
@rwjblue did you have any outstanding concerns or can I merge this?

PR #13457, opened 2016-05-05
[Glimmer 2] Skip tests instead of bailing

chadhietala:


krisselden:
sorry about that, it is just because of the deleted test can you rebase?

PR #13459, opened 2016-05-06
Engine registrations

dgeb:
Splits registrations between Application and Engine classes. Registrations that have proven to be necessary for all engines, based on work in the [ember-engines](https://github.com/dgeb/ember-engines) addon, are moved from the Application class to the Engine class.

This eliminates the need for overrides in ember-engines, while tests still ensure that applications are initialized with the same entries in their registries as always. This does make the registry verification tests rather verbose, but I wanted to make sure nothing was dropped during this transition.

There is still work TBD re: synchronization between engine instances and their parents which could be handled either in this same PR or separately.

/cc @rwjblue

### DIFF
--- a/packages/ember-application/lib/system/application.js
+++ b/packages/ember-application/lib/system/application.js
@@ -12,11 +12,8 @@ import Namespace, {
 } from 'ember-runtime/system/namespace';
 import { runLoadHooks } from 'ember-runtime/system/lazy_load';
 import run from 'ember-metal/run_loop';
-import Controller from 'ember-runtime/controllers/controller';
 import HTMLBarsDOMHelper from 'ember-htmlbars/system/dom-helper';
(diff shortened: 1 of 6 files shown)


rwjblue:
@dgeb - This looks good to me.  

> There is still work TBD re: synchronization between engine instances and their parents which could be handled either in this same PR or separately.

I'm in favor of landing things in smaller pull requests (like this one) if possible.  Can we button this one up, and land it then submit subsequent PRs?

dgeb:
> I'm in favor of landing things in smaller pull requests (like this one) if possible. Can we button this one up, and land it then submit subsequent PRs?

@rwjblue Sounds good. I've rebased and removed the "WIP" tag.

rwjblue:
Thanks @dgeb!

PR #13461, opened 2016-05-06
Prevent ungettable objects from getting gotten during path lookup

pittst3r:
This is a follow-up to #13449.

It was observed by @krisselden that doing something like the following…

``` js
Ember.get({ foo: 42 }, 'foo.bar');
```

… would eventually call `get` internally like…

``` js
get(42, 'bar');
```

… which is no bueno.

To generalize, we should only call `get` from `_getPath` with objects (except `null`), functions, and strings. I'm also wondering if we should add a deprecation to warn people when they do something like…

``` js
Ember.get(42, 'bar');
```

… so we can eventually add an assertion. Right now that will just return `undefined`.

### DIFF
--- a/packages/ember-metal/lib/property_get.js
+++ b/packages/ember-metal/lib/property_get.js
@@ -5,6 +5,12 @@
 import { assert } from 'ember-metal/debug';
 import { isPath, hasThis } from 'ember-metal/path_cache';
 
+const ALLOWABLE_TYPES = {
+  object: true,
+  function: true,
+  string: true
+};
+
 // ..........................................................
 // GET AND SET
 //
@@ -75,7 +81,7 @@ export function _getPath(root, path) {
   let parts = path.split('.');
 
   for (let i = 0; i < parts.length; i++) {
-    if (obj == null) {
+    if (!isGettable(obj)) {
       return undefined;
     }
 
@@ -87,6 +93,14 @@ export function _getPath(root, path) {
   }
 
   return obj;
+}
+
+function isGettable(obj) {
+  if (obj == null) {
+    return false;
+  }
+
+  return ALLOWABLE_TYPES[typeof obj];
 }
 
 /**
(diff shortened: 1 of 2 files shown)


packages/ember-metal/lib/property_get.js
```diff
@@ -89,6 +89,22 @@ export function _getPath(root, path) {
   return obj;
 }
 
+function isGettable(obj) {
+  if (obj == null) {
+    return false;
+  }
+
+  let allowableTypes = ['object', 'function', 'string'];
```
krisselden: move to const outside the function

krisselden: also can just be an object literal with 1s

rwjblue: like:

```
const ALLOWABLE_TYPES = {
  object: true,
  function: true,
  string: true
}
```

robbiepitts: Will do. Thanks, guys.

robbiepitts: Just pushed the update. In the update I am casting `undefined` into `false`, so let me know if you believe that to be superfluous and I will remove.

krisselden:
I think @mmun wanted a test for `obj = { str: '' }; get(obj, 'str.length');`

robbiepitts:
There's already one [here](https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/tests/accessors/get_path_test.js#L53), but now that I'm looking at it again it's only `equal` and not `strictEqual`. I can update that.

robbiepitts:
K, `strictEqual` added. 👍

packages/ember-metal/lib/property_get.js
```diff
@@ -89,6 +95,14 @@ export function _getPath(root, path) {
   return obj;
 }
 
+function isGettable(obj) {
+  if (obj == null) {
+    return false;
+  }
+
+  return !!ALLOWABLE_TYPES[typeof obj];
```
rwjblue: I don't think the explicit coercion is needed here.  It is probably fine either way though...

robbiepitts: Just pushed the update for this.

krisselden:
This looks good, though, this is cleanup, not really a BUGFIX.

PR #13152, opened 2016-03-21
[Glimmer2] Fragment Tests WIP

chadhietala:
This seems to not be implemented in Glimmer2.
- [x] Basic tests to ensure fragment templates work properly
- [ ] Coalesce other fragment tests

**Existing Fragment Tests**
- [ ] [if_unless_test.js](https://github.com/emberjs/ember.js/blob/fd03495063ea7862bdb67a6c96dd57159d074a52/packages/ember-htmlbars/tests/helpers/if_unless_test.js#L192)
- [ ]  [Fragment with event handler part 1](https://github.com/emberjs/ember.js/blob/fd03495063ea7862bdb67a6c96dd57159d074a52/packages/ember-views/tests/views/component_test.js#L309)
- [ ] [Fragment with event handler part 2](https://github.com/emberjs/ember.js/blob/fd03495063ea7862bdb67a6c96dd57159d074a52/packages/ember-views/tests/views/component_test.js#L334)
- [ ]  [Fragment with event handler part 3](https://github.com/emberjs/ember.js/blob/fd03495063ea7862bdb67a6c96dd57159d074a52/packages/ember-views/tests/views/component_test.js#L358)
- [ ]  [Fragment with classNameBindings ](https://github.com/emberjs/ember.js/blob/fd03495063ea7862bdb67a6c96dd57159d074a52/packages/ember-views/tests/views/view/create_element_test.js#L41)
- [ ]  [DEPRECATED  renders the child view templates in the right context](https://github.com/emberjs/ember.js/blob/fd03495063ea7862bdb67a6c96dd57159d074a52/packages/ember-views/tests/views/view/create_element_test.js#L81)
- [ ] [DEPRECATED does not wrap many tr children in tbody elements](https://github.com/emberjs/ember.js/blob/fd03495063ea7862bdb67a6c96dd57159d074a52/packages/ember-views/tests/views/view/create_element_test.js#L106)
- [ ] [jQuery lookups](https://github.com/emberjs/ember.js/blob/fd03495063ea7862bdb67a6c96dd57159d074a52/packages/ember-views/tests/views/view/jquery_test.js#L55)
- [ ]  [Fragment parent](https://github.com/emberjs/ember.js/blob/fd03495063ea7862bdb67a6c96dd57159d074a52/packages/ember-views/tests/views/view_test.js#L93)
[...]

PR #13463, opened 2016-05-06
[Glimmer2] Migrate fragment tests

asakusuma:
Picking up what @chadhietala started here: https://github.com/emberjs/ember.js/pull/13152

I omitted the "virtual view" (not quite sure what that is) tests and the [Fastboot](https://github.com/emberjs/ember.js/blob/fd03495063ea7862bdb67a6c96dd57159d074a52/tests/node/visit-test.js#L307) test that @chadhietala originally listed. Not sure the fastboot test belongs here since it seems that the focus is fastboot and not templating features.

@krisselden @rwjblue can either of you guys speak to whether the "virtual view" tests belong here?

### DIFF
--- a/packages/ember-glimmer/tests/integration/components/fragment-components-test.js
+++ b/packages/ember-glimmer/tests/integration/components/fragment-components-test.js
@@ -3,8 +3,13 @@ import { strip } from '../../utils/abstract-test-case';
 import { set } from 'ember-metal/property_set';
 import Component from 'ember-views/components/component';
 
-
 moduleFor('Components test: fragment components', class extends RenderingTest {
+  getCustomDispatcherEvents() {
+    return {
+      hitDem: 'folks'
+    };
+  }
+
   ['@htmlbars fragments do not render an outer tag']() {
     let instance;
     let FooBarComponent = Component.extend({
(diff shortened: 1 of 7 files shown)


krisselden:
virtual view had to do with parentView/childViews stuff, back when each, if and helpers were views.

packages/ember-glimmer/tests/utils/test-helpers.js
```diff
@@ -83,7 +83,7 @@ export function equalsElement(element, tagName, attributes, content) {
     QUnit.push(element instanceof HTMLElement, null, null, 'Element must be an HTML Element, not an SVG Element');
   } else {
     QUnit.push(
-      element.attributes.length === expectedCount,
+      element.attributes.length === expectedCount || !attributes,
```
asakusuma: Not sure if this is kosher, but seemed strange to me that you couldn't use `assertElement` without explicitly expecting certain attributes. For some assertions, attributes are irrelevant (see `renders a contained view with omitted start tag and tagless parent view context`).

packages/ember-glimmer/tests/integration/components/fragment-components-test.js
```diff
@@ -0,0 +1,156 @@
...
+
+    this.render(`{{root-table}}`);
+
+    this.assertElement(this.firstChild, { tagName: 'table' });
+    this.assertElement(this.firstChild.firstElementChild, { tagName: 'tr' });
+
+    this.runTask(() => this.rerender());
+
+    this.assertElement(this.firstChild, { tagName: 'table' });
+    this.assertElement(this.firstChild.firstElementChild, { tagName: 'tr' });
```
krisselden: Can we use a different tag combo? this html wouldn't SSR correctly and rehydrate because of the missing tbody, I feel weird about codifying it in a test. I know it was ported from an existing test, but I think that was also originally testing if the old morph tags survived the auto tbody insertion.  This no longer seems relevant.

PR #13466, opened 2016-05-09
[Glimmer2] Fix a few dynamic component tests for Glimmer

mitchlloyd:
I was perusing some of the skipped tests and saw some low hanging fruit:
1. "component helper properly invalidates hash params inside an {{each}} invocation" This seems to be fixed now that Glimmer component lifecycle hooks are working. 
2. "component with dynamic/static name argument resolving to non-existent component". It appears that these were only missing an assertion.


rwjblue:
Awesome, thank you @mitchlloyd!

PR #13475, opened 2016-05-11
[Glimmer2] Support fragments

asakusuma:
MJ says go tagless
MJ go tagless

cc @krisselden @chancancode @chadhietala


packages/ember-glimmer/lib/components/curly-component.js
```diff
@@ -66,6 +67,11 @@ class CurlyComponentManager {
       bucket.classRef = args.named.get('class');
     }
 
+    assert('You cannot use `classNameBindings` on a tag-less component: ' + component.toString(), () => {
```
chancancode: What about `attributeBindings`, `elementId` etc?

asakusuma: It doesn't appear that there was ever a test for that. I will add the assertions and tests. Any other assertions that need to happen? Events, `classNameBindings`, and `$()` are the only things asserted right now.

asakusuma: I'm guessing [`element`](http://emberjs.com/api/classes/Ember.Component.html#property_element) should not be allowed either

PR #13477, opened 2016-05-11
[Glimmer2] Add fragment assertions

asakusuma:
As requested by @chancancode in https://github.com/emberjs/ember.js/pull/13475.

Doesn't seem like this stuff was being tested in htmlbars

### DIFF
--- a/packages/ember-views/lib/mixins/view_support.js
+++ b/packages/ember-views/lib/mixins/view_support.js
@@ -619,7 +619,7 @@ export default Mixin.create({
   init() {
     this._super(...arguments);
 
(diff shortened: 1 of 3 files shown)


packages/ember-views/lib/mixins/view_support.js
```diff
@@ -619,7 +619,7 @@ export default Mixin.create({
   init() {
     this._super(...arguments);
 
-    if (!this.elementId) {
+    if (!this.elementId && this.tagName !== '') {
```
krisselden: looks good, this was at one point guarded.

PR #13478, opened 2016-05-11
[Glimmer2] Support id as alias to elementId

asakusuma:
cc @chadhietala

### DIFF
--- a/packages/ember-glimmer/tests/integration/components/curly-components-test.js
+++ b/packages/ember-glimmer/tests/integration/components/curly-components-test.js
@@ -23,7 +23,7 @@ moduleFor('Components test: curly components', class extends RenderingTest {
     this.assertComponentElement(this.firstChild, { content: 'hello' });
   }
 
-  ['@htmlbars it can have a custom id and it is not bound']() {
+  ['@test it can have a custom id and it is not bound']() {
     this.registerComponent('foo-bar', { template: '{{id}} {{elementId}}' });
 
     this.render('{{foo-bar id=customId}}', {
@@ -90,7 +90,7 @@ moduleFor('Components test: curly components', class extends RenderingTest {
     assert.equal(foundId, newFoundId);
   }
 
-  ['@htmlbars passing literal id is ignored and results in a default elementId'](assert) {
+  ['@test id is an alias for elementId'](assert) {
     let FooBarComponent = Component.extend({
       tagName: 'h1'
     });
--- a/packages/ember-glimmer/lib/components/curly-component.js
+++ b/packages/ember-glimmer/lib/components/curly-component.js
@@ -18,6 +18,13 @@ export class CurlyComponentSyntax extends StatementSyntax {
   }
 }
 
+function handleAliases(attrs) {
+  if (attrs.id && !attrs.elementId) {
+    attrs.elementId = attrs.id;
+  }
+  return attrs;
+}
+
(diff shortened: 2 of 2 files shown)


chadhietala:
LGTM /cc @krisselden @chancancode

packages/ember-views/lib/mixins/view_support.js
```diff
@@ -620,7 +620,7 @@ export default Mixin.create({
     this._super(...arguments);
 
     if (!this.elementId && this.tagName !== '') {
-      this.elementId = guidFor(this);
```
chancancode: This is the wrong place to do it IMO. I don't think we want to actually allow setting `this.id`  on the instance – we have to rename the `id` arg to `elementId` [on the way in](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/components/curly-component.js#L50).

packages/ember-glimmer/tests/integration/components/curly-components-test.js
```diff
@@ -93,7 +107,7 @@ moduleFor('Components test: curly components', class extends RenderingTest {
     assert.equal(foundId, newFoundId);
   }
 
-  ['@htmlbars passing literal id is ignored and results in a default elementId'](assert) {
```
chancancode: what is (was) this test title saying ?_?

asakusuma: Not sure. I just noticed that it seemed to be testing the `id` feature, so I tried running the test in glimmer and it worked.

packages/ember-views/lib/mixins/view_support.js
asakusuma: Renaming it in `CurlyComponentMananger.create` throws a `Changing a view's elementId after creation is not allowed` error. I don't quite understand the difference between doing it in `CurlyComponentMananger.create` vs where it is now in `view_support`. To me, both seem to be allowing `id` setting per instance. Which seems to be [what's expected anyways](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/tests/integration/components/curly-components-test.js#L1480).

krisselden: There is a big difference, in that we want to isolate all of normalizing of component props to one place.  If this didn't work there there is something odd happening in the curly component manager that needs to be addressed.

chancancode: Sorry for the confusion – github attached the comment to one line above. I meant to say that we shouldn't have to do `this.id || guidFor(this);` at this point – because we should already have normalized it to `elementId` by now.

chancancode: We convert the args to a "creation props" pojo that we pass to `Component.create(...)`. That pojo should contain `elementId`, not `id`.

asakusuma: Figured out the problem. I didn't realize my alias code was getting run on update as well as creation. Seems to work fine now.

packages/ember-glimmer/tests/integration/components/curly-components-test.js
krisselden: test should just be titled that id is an alias for elementId, the title is massively confusing

chancancode:
👍 LGTM. I'll merge this into #13490 since I refactored the params processing code

PR #13479, opened 2016-05-11
A more friendly assert for incorrect mut value

grapho:
I believe the term "binding" makes more sense to most users as opposed to "path", because there are helpers which _do_ accept "paths as strings"

This comes from a user I talked to, whom encountered the assert after doing `{{mut 'isActive'}}` by mistake and had trouble deciphering the meaning of the current assert error.

@wycats as the original author, do you have any thoughts/review for this?

### DIFF
--- a/packages/ember-htmlbars/lib/keywords/mut.js
+++ b/packages/ember-htmlbars/lib/keywords/mut.js
@@ -140,7 +140,7 @@ function mutParam(read, stream, internal) {
       stream = new LiteralStream(literal);
     }
   } else {
-    assert('You can only pass a path to mut', isStream(stream));
+    assert('You can only pass a binding to mut (e.g. `mut foo`)', isStream(stream));
   }
 
   if (stream[MUTABLE_REFERENCE]) {


wycats:
@grapho I think "binding" might be a little jargon'y here. What did you find confusing about "path"?

grapho:
@wycats  path is not a bad term either, however i guess the format of the path is the potentially confusing part.  I encountered a user that was attempting to do `(mut 'foo.bar')` which is a path represented as a sting.. much like how a link-to helper works or another similar helper that accept paths as strings.

taras:
There are several places in the code base where these are described as [quoteless path](https://github.com/emberjs/ember.js/blob/029968601abcd0f2d15e3d0ad420969c0e4653c3/packages/ember-htmlbars/lib/keywords/element-action.js#L19). You can see all [instances here](https://github.com/emberjs/ember.js/search?utf8=✓&q=quoteless). 

It doesn't seem like a strong name because it describes them as the absence of something rather than describing the thing itself. LISP seems to use [binding](https://www.emacswiki.org/emacs/DynamicBindingVsLexicalBinding) to describe these variable bits. Binding is not a very easy thing for beginners to visualize. 

It would be great if we could come up with a name that specifically identifies these dynamic bits.

pixelhandler:
I tend to like "path" one as it describes the object `thing.prop`, one reference that seems to support the concept of "path" is the JSON Patch spec describing what prop on an object is affected by an operation, see http://jsonpatch.com. (my vote is to stick with "path")

grapho:
Oh wow.. i forgot I actually made this a PR and not just an issue :p 

If we are all lukewarm about this change i have no issue closing.  I was considering newer users whom had questions on #slack not understanding what it meant

locks:
Thanks for the discussion everyone, I'm closing as current art seems adequate.

PR #13481, opened 2016-05-11
[Glimmer 2] Migrate attributeBindings tests

chadhietala:


chadhietala:
This is now dependent on https://github.com/tildeio/htmlbars/pull/460. /cc @krisselden

chadhietala:
Merged as part of #13493

PR #13483, opened 2016-05-12
[Glimmer] Ensure run loop when dirtying

nickiaconis:
Ensure there is a run loop running when an object is dirtied. This complements #13430, so that re-rendering happens automatically when properties are changed.

What still needs to be done?
- [x] Don't use an array for re-render scheduling unless necessary.
- [x] If possible, move `schedulerRegistrar` out of `ember-metal/tags`.
- [x] Resolve test failures

### DIFF
--- a/packages/ember-metal/lib/tags.js
+++ b/packages/ember-metal/lib/tags.js
@@ -1,5 +1,7 @@
 import { meta as metaFor } from './meta';
 import require, { has } from 'require';
+import run from './run_loop';
+import { schedulerRegistrar } from 'ember-glimmer/renderer';
 
 let hasGlimmer = has('glimmer-reference');
 let CONSTANT_TAG, CURRENT_TAG, DirtyableTag, makeTag;
@@ -19,11 +21,19 @@ export function tagFor(object, _meta) {
   }
 }
 
+function K() {}
+function ensureRunloop() {
+  if (schedulerRegistrar.hasRegistrations() && !run.backburner.currentInstance) {
+    run.schedule('actions', K);
+  }
+}
+
 if (hasGlimmer) {
   ({ DirtyableTag, CONSTANT_TAG, CURRENT_TAG } = require('glimmer-reference'));
   makeTag = function() { return new DirtyableTag(); };
 
   markObjectAsDirty = function(meta) {
+    ensureRunloop();
     let tag = (meta && meta.readableTag()) || CURRENT_TAG;
     tag.dirty();
   };
--- a/packages/ember-metal/lib/tags.js
+++ b/packages/ember-metal/lib/tags.js
@@ -2,7 +2,12 @@ import { meta as metaFor } from './meta';
 import require, { has } from 'require';
 
 let hasGlimmer = has('glimmer-reference');
-let CONSTANT_TAG, CURRENT_TAG, DirtyableTag, makeTag, rendererHasViews, run;
+let CONSTANT_TAG, CURRENT_TAG, DirtyableTag, makeTag, run;
+
+let hasViews = () => false;
+export function setHasViews(fn) {
+  hasViews = fn;
+}
 
 export let markObjectAsDirty;
 
@@ -25,11 +30,7 @@ function ensureRunloop() {
     run = require('ember-metal/run_loop').default;
   }
 
-  if (!rendererHasViews) {
-    rendererHasViews = require('ember-glimmer/renderer').rendererHasViews;
-  }
-
-  if (rendererHasViews() && !run.backburner.currentInstance) {
+  if (hasViews() && !run.backburner.currentInstance) {
     run.schedule('actions', K);
   }
 }
--- a/packages/ember-metal/lib/tags.js
+++ b/packages/ember-metal/lib/tags.js
@@ -2,7 +2,7 @@ import { meta as metaFor } from './meta';
 import require, { has } from 'require';
 
 let hasGlimmer = has('glimmer-reference');
-let CONSTANT_TAG, CURRENT_TAG, DirtyableTag, makeTag, schedulerRegistrar, run;
+let CONSTANT_TAG, CURRENT_TAG, DirtyableTag, makeTag, rendererHasViews, run;
 
 export let markObjectAsDirty;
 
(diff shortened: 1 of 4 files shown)


packages/ember-routing-htmlbars/tests/helpers/closure_action_test.js
```diff
@@ -287,20 +287,20 @@ test('action value is returned', function(assert) {
...
+  outerComponent = run(EmberComponent.extend({
     layout: compile('{{view innerComponent submit=(action outerSubmit)}}'),
     innerComponent,
     outerSubmit() {
       return returnedValue;
     }
-  }).create();
+  }), 'create');
 
   runAppend(outerComponent);
```
krisselden: I would have moved the view creation into the runAppend helper

packages/ember-routing-htmlbars/tests/helpers/closure_action_test.js
```diff
@@ -312,20 +312,20 @@ test('action value is returned', function(assert) {
...
-  outerComponent = EmberComponent.extend({
+  outerComponent = run(EmberComponent.extend({
     layout: compile('{{view innerComponent submit=(action outerSubmit)}}'),
     innerComponent,
     isOuterComponent: true,
     outerSubmit() {
       assert.ok(this.isOuterComponent, 'action has the correct context');
     }
-  }).create();
+  }), 'create');
```
krisselden: extend should not need the run loop, only create sets

packages/ember/tests/helpers/link_to_test.js
```diff
@@ -131,10 +131,10 @@ test('Using {{link-to}} does not cause an exception if it is rendered before the
 test('Using {{link-to}} does not cause an exception if it is rendered without a router.js instance', function(assert) {
   appInstance.register('component-lookup:main', ComponentLookup);
 
-  let component = Component.extend({
+  let component = run(Component.extend({
```
krisselden: should just move this into the other loop below

rwjblue:
The number of testing changes required to support this change is a bit concerning to me.  I worry that this would require run loops in app tests where they were not required before (which I would consider a somewhat breaking change).  Are my concerns unfounded?

krisselden:
@rwjblue with old style component testing yes, not with the new style and not in practice in your actual app.

packages/ember-routing-htmlbars/tests/helpers/closure_action_test.js
nickiaconis: Although I'm undoing this, `extend` is not using the run loop here. The result of `extend` is being passed as the context of the run loop on which `'create'` is looked up/invoked.

packages/ember-metal/lib/run_loop.js
```diff
@@ -683,3 +683,55 @@ run._addQueue = function(name, after) {
...
+    let callbacks = run._eventCallbacks[eventName];
+    if (callbacks) {
+      for (let i = 0; i < callbacks.length; ++i) {
+        callbacks[i](arg1, arg2);
+      }
+    }
+  };
+}
+
+for (let eventName in backburner._eventCallbacks) {
```
krisselden: since this is internal, and we are only using begin, it could have just been begin.

packages/ember-metal/lib/tags.js
```diff
@@ -1,5 +1,7 @@
 import { meta as metaFor } from './meta';
 import require, { has } from 'require';
+import run from './run_loop';
+import { schedulerRegistrar } from 'ember-glimmer/renderer';
```
nickiaconis: This causes a circular dependency.

packages/ember-metal/lib/tags.js
```diff
@@ -19,11 +21,49 @@ export function tagFor(object, _meta) {
...
+        backburner.on(eventName, bindEventName(eventName));
+      }
+    }
+  }
+
+  hasRegistrations() {
+    return !!this._eventCallbacks.length;
+  }
+}
+export const schedulerRegistrar = new SchedulerRegistrar();
```
nickiaconis: Defining this in `ember-metal/tags` bothers me. Is there a better way to do this?

chadhietala: Can't you move this stuff to it's own file in `ember-metal` e.g. `ember-metal/schedule-registrar` and then export `schedulerRegistrar` and `ensureRunloop`?

packages/ember-metal/lib/tags.js
krisselden: can we put this part into metal and have renderer pull it from there?

packages/ember-metal/lib/tags.js
nickiaconis: `schedulerRegistrar` is closely related to `ember-glimmer/renderer`. Hence, why it feels wrong to include it in `ember-metal` at all.

nickiaconis: But including _anything_ from `ember-glimmer` in `ember-metal` causes a circular dependency, which breaks things.

packages/ember-metal/lib/tags.js
nickiaconis: I did that at first, but it doesn't feel right b/c `schedulerRegistrar` is so highly related to the glimmer renderer.

packages/ember-metal/lib/tags.js
```diff
@@ -19,11 +24,23 @@ export function tagFor(object, _meta) {
   }
 }
 
+function K() {}
+function ensureRunloop() {
+  if (!run) {
+    run = require('ember-metal/run_loop').default;
```
krisselden: This should be able to move to the top now that this isn't importing ember-glimmer

nickiaconis: It cannot. Within ember-metal, run_loop imports property_events, property_events imports tags, and tags needs to import run_loop. However, this creates a circular dependency. It had worked at one point because of a beneficial, coincidental ordering of evaluation. Using the imperative require lets us explicitly choose where the circle unwinds, instead of leaving it up to fate.

nickiaconis: @krisselden ^^^See previous note.^^^ (I don't know if you get a notification unless you're tagged.)

PR #13487, opened 2016-05-12
Helper references reorg

chancancode:
- Formalize the reference creation pattern. `SomeReference.create` is responsible for determining "const-ness" and apply the right optimization if applicable.
- Move specialized reference out of the `/utils/references.js` file. This file should be reserved for common "util" references like `RootReference` and `CachedReference`. The specialized ones could just live in where they are used.
- New optimization: simple helpers with const args produces const references – that means things like `{{concat "foo" "bar"}}` (or more likely – `{{concat foo bar}}` where `foo` and `bar` are known to come from const sources) does not incur any updating cost.
  
  This implements the optimization discussed https://github.com/tildeio/glimmer/pull/133#discussion_r60147755

cc @zackthehuman @krisselden

### DIFF
--- a/packages/ember-glimmer/lib/utils/references.js
+++ b/packages/ember-glimmer/lib/utils/references.js
@@ -1,7 +1,7 @@
 import { get } from 'ember-metal/property_get';
 import { tagFor } from 'ember-metal/tags';
-import { CURRENT_TAG, CONSTANT_TAG, VOLATILE_TAG, ConstReference, DirtyableTag, UpdatableTag, combine, isConst, referenceFromParts } from 'glimmer-reference';
-import { ConditionalReference as GlimmerConditionalReference, UNDEFINED_REFERENCE } from 'glimmer-runtime';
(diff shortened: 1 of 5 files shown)


packages/ember-glimmer/lib/utils/references.js
```diff
@@ -211,42 +151,26 @@ export class ConditionalReference extends GlimmerConditionalReference {
...
-  static create(_condRef, _truthyRef, _falsyRef) {
-    let condRef = ConditionalReference.create(_condRef);
-    let truthyRef = _truthyRef || UNDEFINED_REFERENCE;
-    let falsyRef = _falsyRef || UNDEFINED_REFERENCE;
-
-    if (isConst(condRef)) {
-      return condRef.value() ? truthyRef : falsyRef;
+export class SimpleHelperReference extends CachedReference {
+  static create(helper, args) {
+    if (isConst(args)) {
```
chancancode: This is where the new const args optimization happens

PR #13488, opened 2016-05-12
[Glimmer2] Add support for textarea

asakusuma:
~~The remaining `@htmlbars` tests are waiting for [this](https://github.com/emberjs/ember.js/pull/13481/files#diff-68fe34e1d224b37871364fe5d84cc481R338) to land.~~

### DIFF
--- a/packages/ember-glimmer/lib/environment.js
+++ b/packages/ember-glimmer/lib/environment.js
@@ -31,6 +31,10 @@ import { default as unbound } from './helpers/unbound';
 import { default as classHelper } from './helpers/-class';
 import { OWNER } from 'container/owner';
 
+const builtInComponents = {
+  textarea: '-text-area'
+};
+
 const builtInHelpers = {
   concat,
   if: inlineIf,
@@ -129,6 +133,18 @@ export default class Environment extends GlimmerEnvironment {
         let definition = this.getComponentDefinition(path);
 
         if (definition) {
+          wrapClassBindingAttribute(args);
+          wrapClassAttribute(args);
+          return new CurlyComponentSyntax({ args, definition, templates });
+        }
+      } else {
+        // Check if it's a keyword
+        let mappedKey = builtInComponents[key];
+        if (mappedKey) {
+          if (mappedKey !== key) {
+            path = path.map((segment) => segment === key ? mappedKey : segment);
+          }
+          let definition = this.getComponentDefinition(path);
           wrapClassBindingAttribute(args);
           wrapClassAttribute(args);
           return new CurlyComponentSyntax({ args, definition, templates });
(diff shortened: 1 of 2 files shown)


packages/ember-glimmer/lib/environment.js
```diff
@@ -133,6 +137,18 @@ export default class Environment extends GlimmerEnvironment {
           wrapClassAttribute(args);
           return new CurlyComponentSyntax({ args, definition, templates });
         }
+      } else {
+        // Check if it's a keyword
+        let mappedKey = keywords[key];
+        if (mappedKey) {
+          if (mappedKey !== key) {
+            path = path.map((segment) => segment === key ? mappedKey : segment);
```
chadhietala: Is this going to from `textarea` to `-textarea`? Is this because of the `-` requirement for components? If so we should extract this logic out to a `toKeyword` function that does the conversion and does the arg wrapping.

asakusuma: It's going from `textarea` to `-text-area`. It's because `-text-area` is the [registered name of the textarea component](https://github.com/emberjs/ember.js/blob/master/packages/ember-application/lib/system/application.js#L1103).

htmlbars is does the same [name re-mapping](https://github.com/emberjs/ember.js/blob/master/packages/ember-htmlbars/lib/keywords/textarea.js#L193).

chadhietala: Sounds good, but lets extract it out and make it generic if possible. I have a feeling we will need this for `{{input}}` as well.

asakusuma: How should this be more generic? Should be able to just add the `input` mapping to the [`keyword` POJO](https://github.com/emberjs/ember.js/pull/13488/files#diff-c8938eb5cb17b733982744e1621cb8e4R34) to get input working.

asakusuma: I took a look at input, and to get input to work, we'll need to add the ability to choose the component based on the keyword, since input can be a checkbox or a textfield. I think we should just merge what I have now and we can add the dynamic component lookup when we get to input.

chadhietala: Sounds good.

asakusuma:
I believe this PR may need to wait for https://github.com/emberjs/ember.js/pull/13415 cc @krisselden

asakusuma:
@krisselden does [your PR](https://github.com/emberjs/ember.js/pull/13500/files#diff-54b8f1605a8b0cf557eddab267f2db40R21) handle resolving the glimmer component when in glimmer mode?

packages/ember-glimmer/lib/environment.js
```diff
@@ -31,6 +31,10 @@ import { default as unbound } from './helpers/unbound';
 import { default as classHelper } from './helpers/-class';
 import { OWNER } from 'container/owner';
 
+const keywords = {
```
krisselden: keywords was a specific thing in htmlbars that doesn't existing in glimmer 2.  This is introducing a new concept that non of the other glimmer 2 builtins have used.  I would say builtin component is more what you are describing here.  /cc @wycats

asakusuma: Roger, changed `keywords` to `builtInComponents`

krisselden:
lgtm, will merge when travis is caught up to master

PR #13491, opened 2016-05-12
[GLIMMER2] Closure actions + tests

zackthehuman:
Implements closure actions for Glimmer2. There are a few things that are unresolved:
- <strike>Generate a label for the action (streams had labels, references don't?)</strike>
- <strike>Support `mut` (there is no `mut` in Glimmer yet)</strike>
- <strike>Support `INVOKE`ables. Not sure what this is or how it's used, so I left it out.</strike>


packages/ember-glimmer/lib/ember-template-compiler/plugins/transform-action-syntax.js
```diff
@@ -39,17 +39,29 @@ TransformActionSyntax.prototype.transform = function TransformActionSyntax_trans
...
+        insertThisAsFirstParam(node, b);
       }
     }
   });
 
   return ast;
 };
+
+function isAction(node) {
+  return (node.type === 'MustacheStatement' || node.type === 'SubExpression' || node.type === 'ElementModifierStatement') && (node.path.original === 'action');
```
Serabe: I don't think you need to check the `node.type` anymore. `traverse` does that for you already.

PR #13493, opened 2016-05-13
[Glimmer 2] Migrate attributeBindings tests

krisselden:


krisselden:
Includes #13481

krisselden:
@chadhietala this is in purgatory, because simple-html-tokenizer in htmlbars is used also for glimmer, and the current version breaks glimmer tests which are for some reason run in ember too, and I can't update glimmer because of other changes.  So this is currently blocked.

PR #13503, opened 2016-05-16
Application / Engine boot refactor

dgeb:
Common boot concerns for engine instances and application instances have been moved to the `EngineInstance` class. These concerns include running engine initializers and engine instance initializers. 

Engine initializers are run once before the first engine instance is built, following the [approach taken in ember-engines](https://github.com/dgeb/ember-engines/commit/a7f3a891a4f243e97cc5c1b2d94b2d36b45505bb).

Setup of an engine instance's registry based upon its parent will be handled in a separate PR.

### DIFF
--- a/packages/ember-application/tests/system/engine_instance_initializers_test.js
+++ b/packages/ember-application/tests/system/engine_instance_initializers_test.js
@@ -1,0 +1,366 @@
+import run from 'ember-metal/run_loop';
+import Engine from 'ember-application/system/engine';
+import EngineInstance from 'ember-application/system/engine-instance';
+
+let MyEngine,
+    myEngine,
+    myEngineInstance;
+
+QUnit.module('Ember.Engine instance initializers', {
+  setup() {
+  },
+
+  teardown() {
+    run(function() {
+      if (myEngineInstance) {
+        myEngineInstance.destroy();
+      }
+
+      if (myEngine) {
+        myEngine.destroy();
+      }
+    });
+  }
+});
+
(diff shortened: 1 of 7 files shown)


packages/ember-application/tests/system/engine_instance_initializers_test.js
```diff
@@ -0,0 +1,366 @@
...
+        secondEngineInstance.destroy();
+
+        firstEngine.destroy();
+        secondEngine.destroy();
+      });
+    });
+});
+
+QUnit.test('initializers are per-engine', function() {
+  expect(0);
```
chadhietala: There might be a better assertion for this.

dgeb: I based that on the analogous [application instance initializer test](https://github.com/emberjs/ember.js/blob/master/packages/ember-application/tests/system/instance_initializers_test.js#L320-L333).

I think using `expect(0)` to verify no error is raised is used pretty often, but I'm open to another alternative. I suppose we should change both tests if we change this one.

dgeb: I'll add some verification that initializers raise if duplicated, which would make both tests more rigorous.

chadhietala: Sounds good

dgeb:
@chadhietala Thanks for the review. Just updated with some more rigorous testing for uniqueness of initializer names.

chadhietala:
@dgeb LGTM. Great stuff! ✨

rwjblue:
Thanks @dgeb!

PR #13504, opened 2016-05-16
Remove unreachable code in transformation

Serabe:
In `transform-top-level-components` there was an unused callback and the node's
type was checked twice. This PR fixes them.


krisselden:
@Serabe this looks good, was does it target beta though?

rwjblue:
Ya, lets target canary on this one.

PR #13505, opened 2016-05-16
[Glimmer2] Components should have access to the view registry

chadhietala:
Not sure if this is something we want to continue with or not
/cc @wycats @chancancode

### DIFF
--- a/packages/ember-glimmer/lib/ember-views/child-views-support.js
+++ b/packages/ember-glimmer/lib/ember-views/child-views-support.js
@@ -65,6 +65,7 @@ export default Mixin.create({
 
     instance.parentView = this;
     instance.ownerView = this.ownerView;
+    instance._viewRegistry = this._viewRegistry;
   },
 
   unlinkChild(instance) {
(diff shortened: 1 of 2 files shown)


krisselden:
This is fine for now but this test is testing something internal more HOW instead of WHAT, I would like to move the view registry to the renderer or event dispatcher (which in reality should be per document rather than per app).

chadhietala:
@krisselden Gotcha. I think I was more or less asking if this is the right place. In the current implementation this gets set in a hook that I don't think exists in glimmer https://github.com/emberjs/ember.js/blob/master/packages/ember-views/lib/mixins/view_child_views_support.js#L83-L126

krisselden:
I'm surprised that is even used by anything, that has more to do with the old view layer than anything else

landongn:
(▀̿Ĺ̯▀̿ ̿) :shipit:

mixonic:
@chadhietala should it also be set onto the instance in `init` for shaping perhaps?

krisselden:
@mixonic seems like this could just be injected onto the event dispatcher, I don't understand why every component needs this field?

rwjblue:
Today, the views register themselves upon init. This could easily be done by the renderer or other higher level environment thing though.

Ultimately, I don't think we *have* to inject this into each component. The key bit is that the registry must exist and been kept up to date.

krisselden:
superseded by #13852 which handles event dispatching without needing the behavior this test was testing.

PR #13519, opened 2016-05-17
[WIP] [FEATURE canary] Pass owner to internal helpers in Glimmer 2

Serabe:
Closure components need access to the owner to get the `positionalParams`
property in the component's class. This is needed to both ease processing the
args and to error if a positional parameter clashes with a named argument.


Serabe:
Changed to WIP because we might not be able to use a helper at all.

krisselden:
@Serabe can this be closed for now?  Seems like name lookup should go through the environment

PR #13520, opened 2016-05-18
[Bugfix] fixes bugs around baseURL and rootURL in HistoryLocation and introduces rootURL to NoneLocation

### LINKED ISSUE #13464: FastBoot Bug - Location none doesn't respect rootURL
Location none doesn't respect the router's `rootURL`.  This causes issues while [generating urls](https://github.com/emberjs/ember.js/blob/45248e2fb2c71ec60bb879c7588beeaeffe6dc6c/packages/ember-routing/lib/location/none_location.js#L85-L90) (link-to's for example will omit the `rootURL`).  Compare this to [history's formatURL](https://github.com/emberjs/ember.js/blob/45248e2fb2c71ec60bb879c7588beeaeffe6dc6c/packages/ember-routing/lib/location/history_location.js#L216) and you'll see the difference between implementations.

This is causing issues within the non-browser path of the `visit` API since it [always defaults](https://github.com/emberjs/ember.js/blob/master/packages/ember-application/lib/system/application-instance.js#L384-L388) to the `none` location.

I would be happy to help patch, but first would like to hear feedback though before venturing off.  For example, do we want to stub out the window.location API when in Node and stick with using the history location, do we patch the none location to respect rootURL in formatURL, or something I'm not thinking of?

Part of this is related to:
https://github.com/ember-fastboot/ember-fastboot-server/issues/53

jasonmit:
Currently `rootURL` isn't respected within `Ember.NoneLocation` and the visit API which results in the route not to be found.

The second bug, `formatURL` also does not return the url with the `rootURL` prefixed.  This breaks `linkto` from building the proper string used in hrefs.

This is blocking projects that rely on `rootURL` to migrate to FastBoot.
- Fixes `baseURL` and `rootURL` from being string replaced at any point in the url for `Ember.HistoryLocation`
- Fixes #13464 https://github.com/ember-fastboot/fastboot/issues/53

### DIFF
--- a/packages/ember-application/lib/system/application-instance.js
+++ b/packages/ember-application/lib/system/application-instance.js
@@ -278,10 +278,13 @@ const ApplicationInstance = EngineInstance.extend({
       }
     };
 
-    // Keeps the location adapter's internal URL in-sync
-    get(router, 'location').setURL(url);
+    let location = get(router, 'location');
 
-    return router.handleURL(url).then(handleResolve, handleReject);
+    // Keeps the location adapter's internal URL in-sync
+    location.setURL(url);
+
+    // getURL returns the set url with the rootURL stripped off
+    return router.handleURL(location.getURL()).then(handleResolve, handleReject);
   }
 });
 
--- a/packages/ember-routing/lib/location/none_location.js
+++ b/packages/ember-routing/lib/location/none_location.js
@@ -1,3 +1,4 @@
+import { assert } from 'ember-metal/debug';
 import { get } from 'ember-metal/property_get';
 import { set } from 'ember-metal/property_set';
 import EmberObject from 'ember-runtime/system/object';
@@ -22,15 +23,38 @@ export default EmberObject.extend({
   implementation: 'none',
   path: '',
 
+  detect() {
+    let rootURL = this.rootURL;
+
+    assert('rootURL must end with a trailing forward slash e.g. "/app/"',
+                 rootURL.charAt(rootURL.length - 1) === '/');
+  },
+
   /**
-    Returns the current path.
+    Will be pre-pended to path.
+
+    @private
+    @property rootURL
+    @default '/'
+  */
+  rootURL: '/',
+
+  /**
+    Returns the current path without `rootURL`.
 
     @private
     @method getURL
     @return {String} path
   */
   getURL() {
-    return get(this, 'path');
+    let path = get(this, 'path');
+    let rootURL = get(this, 'rootURL');
+
+    // remove trailing slashes if they exists
+    rootURL = rootURL.replace(/\/$/, '');
+
+    // remove rootURL from url
+    return path.replace(new RegExp('^' + rootURL), '');
   },
 
   /**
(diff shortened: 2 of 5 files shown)


packages/ember-application/lib/system/application-instance.js
```diff
@@ -281,7 +282,8 @@ const ApplicationInstance = EngineInstance.extend({
     // Keeps the location adapter's internal URL in-sync
     get(router, 'location').setURL(url);
 
-    return router.handleURL(url).then(handleResolve, handleReject);
+    // getURL returns the set url with the rootURL stripped off
```
jasonmit: and base URL if used in the context of the HistoryLocation

packages/ember-routing/lib/location/none_location.js
```diff
@@ -23,14 +23,30 @@ export default EmberObject.extend({
...
   getURL() {
-    return get(this, 'path');
+    let path = get(this, 'path');
+    let rootURL = get(this, 'rootURL');
+
+    // remove trailing slashes if they exists
+    rootURL = rootURL.replace(/\/$/, '');
+
+    // remove rootURL from url
+    return path.replace(rootURL, '');
```
rwjblue: Given a `rootURL` of `/foo` we want to strip it from `/foo/bar/baz` but we do not want `/foo/baz/foo/bar` to become `/baz/bar`.

packages/ember-routing/lib/location/none_location.js
```diff
@@ -23,14 +23,30 @@ export default EmberObject.extend({
...
     @method getURL
     @return {String} path
   */
   getURL() {
-    return get(this, 'path');
+    let path = get(this, 'path');
+    let rootURL = get(this, 'rootURL');
+
+    // remove trailing slashes if they exists
+    rootURL = rootURL.replace(/\/$/, '');
```
rwjblue: Can we store the normalized `rootURL` (perhaps in `init`), instead of removing the trailing slash on every `getURL` / `formatURL`?

packages/ember-routing/lib/location/none_location.js
jasonmit: Yup I thought the same.  This was taken from history location.  Want me to fix both?

packages/ember-routing/lib/location/none_location.js
jasonmit: 👍

packages/ember-routing/lib/location/none_location.js
rwjblue: Yes, absolutely!

packages/ember-routing/lib/location/none_location.js
jasonmit: The caveat this introduces is you wouldn't be able to change the rootURL after the router setup.  Unsure if that's an actual use case or not though.  History location is also doing this as well.

jasonmit:
@rwjblue ready for another review -- thanks again

packages/ember-routing/lib/location/none_location.js
rwjblue: Yeah, that makes sense. We could make it a CP with a setter maybe, but I'm not sure it's worth the effort.

rwjblue:
Thanks!

jasonmit:
@rwjblue we should likely force the rootURL to / when in testing.  People likely have relied on this bug, I.e, a computed property as the rootURL or an initializer without checking the environment first.

Thoughts?

PR #13523, opened 2016-05-18
[CLEANUP] Split template specific registries

chadhietala:


packages/ember-application/lib/system/engine.js
```diff
@@ -16,18 +16,13 @@ import DefaultResolver from 'ember-application/system/resolver';
 import EngineInstance from './engine-instance';
 import isEnabled from 'ember-metal/features';
 import symbol from 'ember-metal/symbol';
-
 import Controller from 'ember-runtime/controllers/controller';
 import TextField from 'ember-templates/components/text_field';
```
krisselden: these should be done the same way, not use the ember-templates pseudo package.  Ideally all internal stuff would be done with injections and pseudo package just for the reexports

packages/ember-htmlbars/lib/setup-registry.js
```diff
@@ -0,0 +1,35 @@
...
+import topLevelViewTemplate from 'ember-htmlbars/templates/top-level-view';
+import { OutletView as HTMLBarsOutletView } from 'ember-htmlbars/views/outlet';
+import EmberView from 'ember-views/views/view';
+import TextField from 'ember-htmlbars/components/text_field';
+import TextArea from 'ember-htmlbars/components/text_area';
+import Checkbox from 'ember-htmlbars/components/checkbox';
+import LinkToComponent from 'ember-htmlbars/components/link-to';
+
+export function setupApplicationRegistry(commonSetupRegistry, registry) {
+  commonSetupRegistry(registry);
```
krisselden: this should not be passed in, should just call them sequentially in the caller:

```
function setupApplicationRegistry(registry, options) {
   commonSetupRegistry(registry);
   if (options[GLIMMER)) {
      let glimmerSetupRegistry = require('ember-glimmer/setup-registry').setupEngineRegistry;
      glimmerSetupRegistry(registry);
   } else { ... }
}
```

dgeb:
@chadhietala thanks for doing this cleanup 💯

PR #13525, opened 2016-05-19
[Glimmer2] Support partials

asakusuma:
~~Need some guidance @krisselden @mmun. Why are the test partials [registered with a `_` in the name](https://github.com/emberjs/ember.js/compare/master...asakusuma:partials?expand=1#diff-6200fd37f2358aea2699a69836cc0f2bR9)?~~

Right now, I have ember's `refineStatement` passthrough to glimmer's `refineStatement` to do [the PartialSyntax creation](https://github.com/tildeio/glimmer/blob/master/packages/glimmer-runtime/lib/environment.ts#L147-L151).

### DIFF
--- a/packages/ember-glimmer/lib/environment.js
+++ b/packages/ember-glimmer/lib/environment.js
@@ -1,3 +1,4 @@
+import lookupPartial, { hasPartial } from 'ember-views/system/lookup_partial';
 import {
   Environment as GlimmerEnvironment,
   HelperSyntax
@@ -128,7 +129,7 @@ export default class Environment extends GlimmerEnvironment {
       templates
     } = statement;
 
-    if (isSimple && (isInline || isBlock)) {
+    if (key !== 'partial' && isSimple && (isInline || isBlock)) {
       if (key === 'component') {
         return new DynamicComponentSyntax({ args, templates, isBlock });
       } else if (key === 'outlet') {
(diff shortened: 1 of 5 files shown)


asakusuma:
I believe the reason partials still don't work is because [`registerPartial`](https://github.com/emberjs/ember.js/blob/e38b48ee204d692f15ba097a2dd2077448ccd618/packages/ember-glimmer/tests/utils/abstract-test-case.js#L367-L374) needs to actually create a [`PartialDefinition`](https://github.com/tildeio/glimmer/blob/master/packages/glimmer-runtime/index.ts#L144). Not sure how to only do this when running in glimmer mode

packages/ember-glimmer/lib/environment.js
```diff
@@ -162,6 +163,22 @@ export default class Environment extends GlimmerEnvironment {
     return definition;
   }
 
+  hasPartial(name) {
+    return hasPartial(this, name[0]);
+  }
+
+  lookupPartial(name) {
+    let partial = {
+      template: lookupPartial(this, name[0]).spec
```
asakusuma: This doesn't seem right to me

asakusuma:
`this.assertStableRerender();` is the only thing failing right now

asakusuma:
@rwjblue can you shed some light on this travis issue or restart? Everything passed except for the `blueprints` test suite. Glimmer was the only changed dependency.

```
Could not require 'ember-cli-build.js': Cannot find module 'esprima'
Error: Could not require 'ember-cli-build.js': Cannot find module 'esprima'
```

packages/ember-glimmer/lib/environment.js
mmun: Seems ok to me. What is the first argument of `lookupPartial` for?

packages/ember-glimmer/lib/environment.js
```diff
@@ -124,7 +125,7 @@ export default class Environment extends GlimmerEnvironment {
       templates
     } = statement;
 
-    if (isSimple && (isInline || isBlock)) {
+    if (key !== 'partial' && isSimple && (isInline || isBlock)) {
```
krisselden: with short circuiting you want the common cases first, check the booleans before doing the key compare

krisselden:
looks good, will merge after travis catches up

PR #13526, opened 2016-05-19
Clone shared dependencies between an engine instance and its parent.

dgeb:
Continues the move upstream of functionality from [ember-engines](https://github.com/dgeb/ember-engines) into ember core:
- Certain registrations must be global to an application and all its engines. These are cloned from the parent registry to the engine’s registry when the engine instance is booted.

Notes: 
- These changes are only enabled via the `ember-application-engines` feature flag.
- The concept of [custom dependencies](https://github.com/dgeb/ember-engines/blob/master/addon/-private/engine-instance-ext.js#L253-L278) will remain wholly in the addon for now.

### DIFF
--- a/packages/ember-application/lib/system/engine-instance.js
+++ b/packages/ember-application/lib/system/engine-instance.js
@@ -8,6 +8,7 @@ import EmberError from 'ember-metal/error';
 import Registry from 'container/registry';
 import ContainerProxy from 'ember-runtime/mixins/container_proxy';
 import RegistryProxy from 'ember-runtime/mixins/registry_proxy';
+import { privatize as P } from 'container/registry';
 import { getEngineParent, setEngineParent } from 'ember-application/system/engine-parent';
 import { assert } from 'ember-metal/debug';
 import run from 'ember-metal/run_loop';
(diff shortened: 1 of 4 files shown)


packages/ember-application/lib/system/engine-instance.js
```diff
@@ -154,6 +159,32 @@ if (isEnabled('ember-application-engines')) {
...
+    cloneParentDependencies() {
+      let parent = getEngineParent(this);
+
+      [
+        'route:basic',
+        'event_dispatcher:main',
+        P`-bucket-cache:main`,
+        'service:-routing'
+      ].forEach((key) => {
+        this.register(key, parent.resolveRegistration(key));
```
rwjblue: I haven't dug in much, but the Travis failures indicate that either parent is undefined in the tests or it doesn't have resolveRegistration...

dgeb: @rwjblue I wasn't properly mocking the engine's parent in some instance initializer tests. Should be fixed now.

packages/ember-application/tests/system/engine_instance_initializers_test.js
```diff
@@ -9,7 +9,10 @@ let MyEngine,
 
 function buildEngineInstance(EngineClass) {
   let engineInstance = EngineClass.buildInstance();
-  setEngineParent(engineInstance, {});
+  setEngineParent(engineInstance, {
+    lookup() { return {}; },
+    resolveRegistration() { return {}; }
```
dgeb: I'm undecided whether the parent should be mocked this way, but it seemed preferable to booting a separate full parent app for each test (since the tests are for _engine_ instance initializers).

dgeb:
@rwjblue r?

rwjblue:
Thanks @dgeb!

PR #13532, opened 2016-05-19
[Glimmer2] Add support for target actions and move related ember-view tests to new harness

Joelkang:
I didn't delete the tests in the ember-view package since it's not clear to me that we can symlink them the way we did for the ember-htmlbars tests.

Some open questions:
1. [Is actionContext a legacy thing?](https://github.com/emberjs/ember.js/compare/master...Joelkang:target-actions?expand=1#diff-f538d2ecb81fe3f5868ac7d827d5aea8R86)
2. [Is there a better way of checking if something is a reference?](https://github.com/emberjs/ember.js/compare/master...Joelkang:target-actions?expand=1#diff-f538d2ecb81fe3f5868ac7d827d5aea8R207)
3. [This seems to be the only way to obtain the controller of the current outlet](https://github.com/emberjs/ember.js/compare/master...Joelkang:target-actions?expand=1#diff-f538d2ecb81fe3f5868ac7d827d5aea8R206). Let me know if I'm way off base.
4. For the [privatization of the `targetObject` attr](https://github.com/emberjs/ember.js/compare/master...Joelkang:target-actions?expand=1#diff-7c87fffed88a1c7be941a3ec6989c16fR19), I followed the `aliasIdToElement()` pattern, hopefully it's right

### DIFF
--- a/packages/ember-glimmer/lib/ember-views/target-action-support.js
+++ b/packages/ember-glimmer/lib/ember-views/target-action-support.js
@@ -1,0 +1,223 @@
+import { Mixin } from 'ember-metal/mixin';
+import TargetActionSupport from 'ember-runtime/mixins/target_action_support';
+import alias from 'ember-metal/alias';
+import { computed } from 'ember-metal/computed';
+import { assert } from 'ember-metal/debug';
+import { inspect } from 'ember-metal/utils';
+import isNone from 'ember-metal/is_none';
+import { get } from 'ember-metal/property_get';
+import { ARGS } from '../component';
+
+/**
+`Ember.ComponentTargetActionSupport` is a mixin that can be included in a
+component class to add a `triggerAction` method with semantics similar to
+the Handlebars `{{action}}` helper. By default the action's target is the
+component itself, or the controller of the component if it is an outlet.
+
+Note: In normal Ember usage, the `{{action}}` helper is usually the best
+choice. This mixin is most often useful when you are doing more complex
+event handling in custom component subclasses.
+
+For example:
+
+```javascript
+App.SaveButtonComponent = Ember.Component.extend(Ember.ComponentTargetActionSupport, {
+  action: 'save',
+  click: function() {
+    this.triggerAction(); // Sends the `save` action, along with the current context
+                          // to the current controller
+  }
+});
+```
+
+The `action` can be provided as properties of an optional object argument
+to `triggerAction` as well.
+
+```javascript
+App.SaveButtonComponent = Ember.Component.extend(Ember.ComponentTargetActionSupport, {
+  click: function() {
+    this.triggerAction({
+      action: 'save'
+    }); // Sends the `save` action, along with the current context
+        // to the current controller
+  }
+});
+```
+
+@class ComponentTargetActionSupport
+@namespace Ember
+@extends Ember.TargetActionSupport
+@private
+*/
+export default Mixin.create(TargetActionSupport, {
+  /**
+   @property target
+   @private
+  */
+  target: computed('outletState', function () {
+    let outletState = get(this, 'outletState');
+    return outletState ? getCurrentOutletState(outletState).render.controller : undefined;
+  }),
+
+  // @override
+  send(actionName, ...args) {
+    var target;
+    var action = this.actions && this.actions[actionName];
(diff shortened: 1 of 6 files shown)


packages/ember-glimmer/lib/ember-views/target-action-support.js
```diff
@@ -0,0 +1,218 @@
...
+    return outletState ? getCurrentOutletState(outletState).render.controller : undefined;
+  }),
+
+  // @override
+  send(actionName, ...args) {
+    let target;
+    let action = this.actions && this.actions[actionName];
+
+    if (action) {
+      var shouldBubble = action.apply(this, args) === true;
```
chadhietala: Does this just bubble on it's own?

packages/ember-glimmer/lib/syntax/curly-component.js
```diff
@@ -14,6 +14,15 @@ function aliasIdToElementId(args, props) {
   }
 }
 
+// Use `_targetObject` to avoid stomping on a CP
+// that exists in the component
+function privatizeTargetObject(props) {
+  if (props.targetObject) {
+    props._targetObject = props.targetObject;
+    delete props.targetObject;
```
chadhietala: Not sure if want to delete off of props as I think it's going to cause `props` to go into dictionary mode and make it's lookups much slower. However if it's already a `dict` it should be fine. Can you look `processedArgs.value()` actually evaluates to in glimmer?

packages/ember-glimmer/tests/integration/components/target-action-test.js
```diff
@@ -0,0 +1,537 @@
+import { moduleFor, RenderingTest, ApplicationTest } from '../../utils/test-case';
```
chadhietala: Are there tests that you can delete that were ported?

chadhietala: Sorry didn't see you commit message.

packages/ember-glimmer/lib/syntax/curly-component.js
Joelkang: Looks like it's an EmptyObject, so should be fine. But I've just gone ahead and set this to undefined.

packages/ember-glimmer/lib/ember-views/target-action-support.js
```diff
@@ -0,0 +1,217 @@
+import { Mixin } from 'ember-metal/mixin';
```
rwjblue: This duplication seems odd.  What is actually changed from the mixin in ember-views?

Joelkang: It is indeed, the [mixin in ember views](https://github.com/emberjs/ember.js/blob/master/packages/ember-views/lib/mixins/view_target_action_support.js) actually only provided the TargetActionSupport mixin and a couple of CPs. Most of the actual methods for targetable actions in views was implemented in [ember-htmlbars/lib/component.js](https://github.com/emberjs/ember.js/blob/master/packages/ember-htmlbars/lib/component.js#L281). It made sense to me to actually unify these implementations into its own mixin and not pollute the component.js file for glimmer.

rwjblue: I agree with the intent, but I would rather see the implementation moved into the "real" `target-action-support.js`.  Every file we add to `packages/ember-glimmer/lib/<some other package name>/lib/` **must** be moved before we ship, and assuming the implementation here is the same as `ember-htmlbars/component` there is no reason to make our lives harder...

Joelkang: the only key difference is in the way the action is resolved either from `attrs.` or the `ARGS` symbol on the component.

That said, I get what you're saying, but I'm not sure how do make the change without modifying the API of the old `view_target_action_support.js` mixin. If I move the `send()` and `sendAction()` into `target-action-support.js` then `view_target_action_support` will also have those methods, even though in glimmer1 they're only added at the `component.js` level, where `view_target_action_support is mixed-in.

This means that anyone manually mixing-in `view_target_action_support.js` will get more than they bargained for. If this is fine though, I'll go ahead and do that

PR #10899, opened 2015-04-18
Internals should prefer generic array methods over prototype_extensions

stefanpenner:
- [x] addArrayObserver
- [x] removeArrayObserver
- [x] `objectAt`
- [x] `replace`
- [x] `remoteAt`
- [x] `insertAt`
- [ ] pushObject
- [ ] pushObjects
- [ ] re-org public api
- [ ] etc ...
- [ ] docs
- [ ] all the tests

The goal here is merely to decouple the internals from the the observable based prototype extensions.
This will hopefully give us the flexibility to make a future call re: prototype extensions in general. This also eases the pain of using ember with prototype extensions disabled.

All this is accomplished without breaking semver in anyway.

More todo:
- [ ] Merge Enumerable + NativeArray
- [ ] Merge MutableEnumerable + MutableArray
- [ ] Deprecate Enumerable as [bugfix release]
- [ ] Deprecate MutableEnumerable as [bugfix release]

Issue #9269, opened 2014-10-03
EXTEND_PROTOTYPES false by default.


jdalton:
Currently Ember extends built-in prototypes of `Array`, `Function`, & `String` by default. Although Ember doesn't require these extensions and they [can be disabled](http://emberjs.com/guides/configuring-ember/disabling-prototype-extensions/), because they are enabled by default they tend to stay that way as evident by sites like vine, twitch, nbcnews, & discourse using the default settings.

The problem with adding methods like `Array#compact`, `Array#without`, `Array#uniq`, `Function#on`, & `String#capitalize` is that it increases the likelihood that they'll be used and cause problems/roadblocks for future built-in language extensions. For example, if ES7 tried to add `Array#without` that behaved in a different way than Ember's `Array#without` newer environments would get the built-in version while older would get Ember's causing an inconsistency and increased possibility of breaking thing. We've seen this recently with MooTools and `Array#contains`, but it's happened before with `Function#bind`, `String#contains`, and even DOM4 `Element#remove`.

totaltrek:
+1

Extending native prototypes is always dangerous and usually avoidable. I haven't yet used Ember, and this is the main reason why. I want my apps to gracefully *upgrade* as well as they gracefully *degrade*.

It seems to me that the reason prototypes are extended inside Ember is for syntactic sugar, which, let's be honest, is a fantastic goal. I just worry that it's following such a bad practice that the project itself will suffer in the long term.

As an example, let's look at `Function#property` -- the sugar that lets us "allow you to treat a function like a property" (quoting the docs). If ECMA version 7 defines a radically different `#property` on the native `Function` prototype, **all** ember apps risk immediate and permanent failure without crisis-mode maintenance. Instead of attaching `#property` why is the sugar not defined as follows?

```
App.model = Ember.Object.extend{
    oneAttr: 'a string',
    anotherAttr: [ 'an', 'array' ],
    someAttr: Ember.Property( function () {
        return this.anotherAttr.reverse().concat( this.oneAttr ).reverse()
    } )
}
```

This solves at least three problems: one, interfering with the native prototype; two, replacing sugar with safety and consistency; three, I'm not using Ember yet. The only problem remaining is the dusty example code on stackoverflow, random blogs, codeschool, and the like; most of which is already behind if not broken since it was published.

[...]

PR #13534, opened 2016-05-20
[BUGFIX beta] Don't use prototype extensions (arrayContent{Did,Will}Change, removeAt)

btecu:
Do not rely on prototype extensions for `removeAt`.
Related to #9269 and #10899.

### DIFF
--- a/packages/ember-runtime/lib/mixins/mutable_array.js
+++ b/packages/ember-runtime/lib/mixins/mutable_array.js
@@ -26,6 +26,23 @@ import MutableEnumerable from 'ember-runtime/mixins/mutable_enumerable';
 import Enumerable from 'ember-runtime/mixins/enumerable';
 import isEnabled from 'ember-metal/features';
 
+export function removeAt(array, start, len) {
+  if ('number' === typeof start) {
+    if ((start < 0) || (start >= get(array, 'length'))) {
+      throw new EmberError(OUT_OF_RANGE_EXCEPTION);
+    }
+
+    // fast case
+    if (len === undefined) {
+      len = 1;
+    }
+
+    array.replace(start, len, EMPTY);
+  }
+
+  return array;
+}
+
 /**
   This mixin defines the API for modifying array-like objects. These methods
   can be applied only to a collection that keeps its items in an ordered set.
@@ -129,9 +146,9 @@ export default Mixin.create(EmberArray, MutableEnumerable, {
     ```javascript
     let colors = ['red', 'green', 'blue', 'yellow', 'orange'];
 
-    colors.removeAt(0);     // ['green', 'blue', 'yellow', 'orange']
-    colors.removeAt(2, 2);  // ['green', 'blue']
-    colors.removeAt(4, 2);  // Error: Index out of range
+    removeAt(colors, 0);     // ['green', 'blue', 'yellow', 'orange']
+    removeAt(colors, 2, 2);  // ['green', 'blue']
+    removeAt(colors, 4, 2);  // Error: Index out of range
     ```
 
     @method removeAt
@@ -141,20 +158,7 @@ export default Mixin.create(EmberArray, MutableEnumerable, {
     @public
   */
   removeAt(start, len) {
-    if ('number' === typeof start) {
-      if ((start < 0) || (start >= get(this, 'length'))) {
-        throw new EmberError(OUT_OF_RANGE_EXCEPTION);
-      }
-
-      // fast case
-      if (len === undefined) {
-        len = 1;
-      }
-
-      this.replace(start, len, EMPTY);
-    }
-
-    return this;
+    return removeAt(this, start, len);
   },
 
   /**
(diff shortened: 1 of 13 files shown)


packages/ember-runtime/lib/mixins/mutable_array.js
```diff
@@ -128,9 +145,9 @@ export default Mixin.create(EmberArray, MutableEnumerable, {
     ```javascript
     var colors = ['red', 'green', 'blue', 'yellow', 'orange'];
 
-    colors.removeAt(0);     // ['green', 'blue', 'yellow', 'orange']
-    colors.removeAt(2, 2);  // ['green', 'blue']
-    colors.removeAt(4, 2);  // Error: Index out of range
+    removeAt(colors, 0);     // ['green', 'blue', 'yellow', 'orange']
```
rwjblue: These examples are showing how to use the mutable array mixin itself, I would think that we want to leave them as `colors.removeAt(2, 2)`.

rwjblue:
I'm not 100% sure on this.  It seems like the right direction, but I wouldn't expect so many changes in tests that were already using things extended by `NativeArray` (i.e. `Ember.A()`).  We still support `Ember.A().removeAt(...)` and friends, so we should continue to test them.

krisselden:
This is much like moving away from global API internally.  In general, the tests migrated to the modules API but we should have tests that test we export the API onto `Ember`.

stefanpenner:
nice!

krisselden:
we should probably get this in ASAP as it is likely to bit quickly with the churn in glimmer 2 PRs

krisselden:
@rwjblue I think NativeArray could have a test but that any other tests not testing NativeArray directly should be using the standalone methods.

btecu:
I've added one test and update those comments.

btecu:
Rebased

packages/ember-runtime/lib/system/array_proxy.js
```diff
@@ -33,6 +35,40 @@ var EMPTY = [];
 
 function K() { return this; }
 
+export function removeAt(array, start, len) {
+  if ('number' === typeof start) {
```
krisselden: why extract this? it doesn't let you fire the events for this operation without using Ember.A().

krisselden: the first argument is still required to be an ember array.

This isn't the equivalent of Ember.set(obj, key, val) where the obj can be a pojo.

btecu: @krisselden to be able to call `removeAt(array, 1, 2);` in some other files such as `arranged_content_test.js`, although there might not be any advantage over `array.removeAt(1, 2);`.

krisselden: it if requires it to be an array, there isn't a point to the standalone, there is only value migrating to extend prototypes being false by default if Ember isn't doing Ember.A() all over the place, which is very slow.

btecu: That makes sense. I've reverted that chage.

btecu:
Rebased.

rwjblue:
Thanks @btecu, sorry for letting this linger so long.

workmanw:
I'm not sure where or how to direct this feedback, so I'm going to do so here. If there is a better place for this discussion please let me know.

These commits are problematic for [ember-data-model-fragments](https://github.com/lytics/ember-data-model-fragments). The root of the problem is that because `replace` (and others) no longer call `this.arrayContentDidChange(...)`, instead calling `(arrayContentDidChange(this, ...)`, we are unable to override this function in our subclasses array. See here: [array/stateful.js#L178-L190](https://github.com/lytics/ember-data-model-fragments/blob/master/addon/array/stateful.js#L178-L190).

Ultimately I acknowledge that this an model-fragments issue. We were overriding a public API and now the way it's invoked internally from ember has change. So I don't think this is exactly API breakage, but I can definitely see this as problematic for anyone else in a similar situation.

stefanpenner:
@workmanw that sounds like a breaking change to me

workmanw:
@stefanpenner If that's the case, I'd like to get this fixed up before ember 2.8 ships so it doesn't become an upgrade blocker for us, or anyone else. I can submit a PR to revert part of this commit so that internally ember will call `this.arrayContentDidChange(...)` instead of `arrayContentDidChange(this, ...)` if you think that's the right direction.

btecu:
Seems unlikely that breaking the overriding of an api can be considered a breaking change.

stefanpenner:
@btecu it is a public API, although maybe this is a gray area?

@workmanw can you open an issue so we can track and decide if it is an issue or not?

Serabe:
@stefanpenner it seems so. Though a public API, it is a method to call to notify about the change, not to be notified about it. Interesting issue!

workmanw:
@stefanpenner Yea, absolutely.

PR #13535, opened 2016-05-20
[FEATURE ember-route-serializers] Use getSerializer method on router.js

trentmwillis:
Next step in the [Route Serializer RFC](https://github.com/emberjs/rfcs/blob/master/text/0120-route-serializers.md). Here's the [relevant commit from router.js](https://github.com/tildeio/router.js/pull/182).

Unsure if I should go ahead and remove the default `serialize` method from `Route`, or if that should occur after the deprecation period ends.

I think the existing tests should cover the changes here, as it is primarily an internal refactor.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -12,7 +12,7 @@
   "devDependencies": {
     "backburner": "https://github.com/ebryn/backburner.js.git#b893c979cb38f9819f65561725907add1aed7b21",
     "rsvp": "https://github.com/tildeio/rsvp.js.git#3.0.20",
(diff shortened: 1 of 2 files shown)


rwjblue:
I believe we must wait to remove the default implementation until we are removing support. Specifically because of anyone calling _super in a custom serialize.

trentmwillis:
> Specifically because of anyone calling _super in a custom serialize.

Ah, hadn't thought about the `super` use case. Sounds good, I think this is good for review then. Once `Route#serialize` is actually removed, the logic here will greatly simplify.

mixonic:
lgtm

PR #13539, opened 2016-05-21
[CLEANUP] Split template compiler

chadhietala:
This splits the template compiler so that tests and applications have the guarantee of using the correct template compiler.

**Things this solves**
- Up to this point booting an app required that you built a version of Ember with certain AST plugins commented out.
- The compilers were split for tests but not applications meaning Glimmer AST plugins were not applied
- Builds single template compiler file with both compilers inside that respect the feature flag
- Provides a path forward in test migration for more integration type of tests

**Things To Note:**
- `registerPlugin` is essentially legacy at this point as you can now pass AST plugins in the options and they will be added in. Kept for backwards compatiablity.
- AST Plugins are no longer global
- Compile options are now immutable
- Removes legacy AST plugins
  - AssertNoEachIn: Was deprecated in pre 2.0
  - AssertNoViewAndControllerPaths: Was deprecated pre 2.0
  - AssertNoViewHelper: Was deprecated pre 2.0
  - PreventRenderBlock: Was set for removal in 2.4.0


chadhietala:
Dependent on https://github.com/emberjs/emberjs-build/pull/156

rwjblue:
Why?

chadhietala:
@rwjblue See above. Also @krisselden may be able to add some more color.

krisselden:
It has to do with the way the glimmer test harness works (ember-glimmer tests run with or without the flag enabled via setting up its own container and application) and unifying the setup stuff between the harness.  This also cleans up how plugins are passed into the compilers.

@rwjblue Do you have a specific complaint? I'm assuming Chad added the detail above after your comment, I didn't see it before.

rwjblue:
No complaint at all, I just didn't understand the strategy. I read (and replied) via email and the notification didn't have any description.

PR #13541, opened 2016-05-21
[Glimmer2] Add mut helper and update readonly helper

Joelkang:
Implemented based on comments in https://github.com/emberjs/ember.js/pull/13502
- [x] readonly helper tests forthcoming
- [x] Support for invoking closure actions with the `mut` helper
- [x] {{mut}} of {{readonly}} support
- [x] `{{action attrs.somethingUndefined}}` returns PropertyReference as opposed to UNDEFINED_REFERENCE, causing a call-time error rather than a create time error


packages/ember-glimmer/tests/integration/mutable-binding-test.js
```diff
@@ -76,7 +76,7 @@ moduleFor('@htmlbars Mutable bindings integration tests', class extends Renderin
   }
 
   // See https://github.com/emberjs/ember.js/commit/807a0cd for an explanation of this test
-  ['@test using a string value through middle tier does not trigger assertion'](assert) {
+  ['@htmlbars using a string value through middle tier does not trigger assertion'](assert) {
```
chadhietala: Is this because the `mut` wrapping isn't happening?

Joelkang: you're right; updated

packages/ember-glimmer/tests/integration/helpers/readonly-test.js
```diff
@@ -0,0 +1,163 @@
...
+      expectAssertion(() => set(component, 'value', 13), /cannot update the value of a readonly/);
+    } else {
+      this.assert.notOk(component.attrs.value.update);
+    }
+
+    this.assertText('12', 'local property is not updated');
+    this.assert.equal(get(this.context, 'val'), 12, 'upstream attribute is not updated');
+  }
+
+  ['@htmlbars {{mut}} of a {{readonly}} mutates only the middle and bottom tiers']() {
```
Joelkang: @krisselden @chancancode so this is still failing in glimmer.

To make it backwards compatible for curly components, what ember-htmlbars seems to be doing is disconnecting the readonly if it is updated from a child, such that `get(child, 'parentAttr')` is no longer the same as `child.attr.parentAttr`. But when a change comes from the upstream property, the value correctly updates.

I've tried a variety of ways of disconnecting the value from the reference and reconnecting it, but nothing I've tried really works .It also seems like there could be multiple places this logic can go into (should it be in `process-args` or `mut`?) Any direction here would be helpful.

packages/ember-glimmer/lib/utils/process-args.js
```diff
@@ -48,15 +50,32 @@ class SimpleArgs {
...
 
+class MutableCell {
+  constructor(ref, value) {
+    this._ref = ref;
+    this[MUTABLE_CELL] = true;
+    this.value = value;
+  }
+
+  update(val) {
+    this._ref[UPDATE](val);
```
chadhietala: Doesn't the value need to be reflected on both sides?

Joelkang: So I added tests to verify the value of `component.attrs.myAttr.value` and `component.set('myAttr')`and they all seem to be passing without any code changes. Is that what you meant?

packages/ember-glimmer/lib/utils/references.js
```diff
@@ -7,9 +7,13 @@ import { ConditionalReference as GlimmerConditionalReference, NULL_REFERENCE, UN
...
 import { dasherize } from 'ember-runtime/system/string';
+import { ARGS } from '../component';
 
 export const UPDATE = symbol('UPDATE');
-
+export const READONLY = symbol('READONLY');
+export const MUT = symbol('MUT');
+export const ACTION = symbol('ACTION');
+export const INVOKE = symbol('INVOKE');
 // FIXME: fix tests that uses a "fake" proxy (i.e. a POJOs that "happen" to
```
Joelkang: Decoupling from htmlbars instead of importing the existing symbols. Let me know if this makes sense

packages/ember-glimmer/lib/helpers/mut.js
```diff
@@ -0,0 +1,37 @@
...
+    super(...arguments);
+    this[MUT] = true;
+  }
+
+  [UPDATE](val) {
+    let argRef = this.args.positional.at(0);
+    argRef[UPDATE](val);
+  }
+
+  [INVOKE](val) {
```
Joelkang: This looks like an ember-internal thing and as far as I can tell no internal package really calls INVOKE, should we remove this? It's a symbol anyway so it can't be called externally.

packages/ember-glimmer/tests/integration/helpers/closure-action-test.js
```diff
@@ -246,40 +247,27 @@ moduleFor('Helpers test: closure {{action}}', class extends RenderingTest {
...
     this.registerComponent('outer-component', {
-      ComponentClass: OuterComponent,
       template: '{{inner-component submit=(action somethingThatIsUndefined)}}'
     });
 
-    // TODO: Improve this to test that the name of the parameter is contained
-    // within the error message.
     this.assert.throws(() => {
       this.render('{{outer-component}}');
     }, /An action could not be made for `.*` in .*\. Please confirm that you are using either a quoted action name \(i\.e\. `\(action '.*'\)`\) or a function available in .*\./);
```
Joelkang: I actually think this should assert for the same error as the test below it: looks like it should throw a compile-time error than a run-time error. (which actually means we don't really have a test for compile-time check for UNDEFINED_REFERNECE or NULL_REFERENCE.

packages/ember-glimmer/tests/integration/helpers/closure-action-test.js
```diff
@@ -2,7 +2,8 @@ import run from 'ember-metal/run_loop';
 import { computed } from 'ember-metal/computed';
 import isEnabled from 'ember-metal/features';
 import { subscribe, unsubscribe } from 'ember-metal/instrumentation';
-import { INVOKE } from 'ember-htmlbars/keywords/closure-action';
+import { INVOKE as HTMLBARS_INVOKE } from 'ember-htmlbars/keywords/closure-action';
+import { INVOKE as GLIMMER_INVOKE } from 'ember-glimmer/utils/references';
```
zackthehuman: Would be nice to put this forking into one of the helper files so that the test can just `import { INVOKE } from '../../utils/helpers';` and it would get either the HTMLBars or Glimmer symbol.

packages/ember-glimmer/tests/integration/helpers/closure-action-test.js
```diff
@@ -246,46 +247,53 @@ moduleFor('Helpers test: closure {{action}}', class extends RenderingTest {
...
-    // TODO: Improve this to test that the name of the parameter is contained
-    // within the error message.
+    // The assertion is different because in the HTMLBars case, the value is always a stream,
+    // whether or not the path read actually has value.
+    // In the Glimmer case, we're checking the value and not the reference, which could be undefined
+    // and thus throws the correct error.
     this.assert.throws(() => {
       this.render('{{outer-component}}');
-    }, /An action could not be made for `.*` in .*\. Please confirm that you are using either a quoted action name \(i\.e\. `\(action '.*'\)`\) or a function available in .*\./);
+    }, this.isGlimmer ?
```
zackthehuman: In cases like this I have just created a separate test that only applies to Glimmer since Glimmer has extended functionality.

Joelkang: This is just a change in the assert though, is that worth the cost of one more test to maintain though

packages/ember-glimmer/lib/helpers/mut.js
rwjblue: In general, it is used (as private API) by ember-concurrency and ember-route-action-helper addons.

packages/ember-glimmer/lib/utils/references.js
rwjblue: I would prefer to reuse the same symbols, but understand if that is not possible...

packages/ember-glimmer/lib/component.js
```diff
@@ -71,18 +71,16 @@ const Component = CoreView.extend(
       let args, reference;
 
       if ((args = this[ARGS]) && (reference = args[key])) {
-        if (reference[UPDATE]) {
-          reference[UPDATE](get(this, key));
-        } else {
-          let name = this._debugContainerKey.split(':')[1];
-          let value = get(this, key);
-          throw new Error(strip`
+        let name = this._debugContainerKey.split(':')[1];
```
rwjblue: This is only used for the assertion, but as written it will not get stripped.  Can you refactor to avoid splitting in prod builds where the assertion is not needed?  Something like:

```js
let value = get(this, key);
runInDebug(function() {
  let name = this._debugContainerKey.split(':')[1];
  assert(strip`
    Cannot set the \`${key}\` property (on component ${name}) to
    \`${value}\`. The \`${key}\` property came from an immutable
    binding in the template, such as {{${name} ${key}="string"}}
    or {{${name} ${key}=(if theTruth "truth" "false")}}.
  `, reference[UPDATE]);
});

reference[UPDATE](value);
```

packages/ember-glimmer/lib/helpers/action.js
```diff
@@ -31,15 +33,21 @@ export class ClosureActionReference extends CachedReference {
...
     // look to "target".
     let actionType = typeof rawAction;
     let action = rawAction;
 
-    if (actionType === 'string') {
+    if (isNone(rawAction)) {
+      throw new EmberError(`Action passed is null or undefined in (action) from ${target}.`);
+    } else if (rawActionRef[INVOKE]) {
+      // TODO: can we just replace this with isMut/UPDATE altogether
+      // since INVOKE is not really public API anyway?
```
rwjblue: This exposes some experimentation (via private API) in addons like ember-concurrency and ember-route-action-helper.  If possible, I'd prefer to avoid breaking that...

packages/ember-glimmer/lib/helpers/mut.js
```diff
@@ -0,0 +1,37 @@
...
+};
+
+function mut({ positional }) {
+  return positional.at(0).value();
+}
+
+class MutHelperReference extends InternalHelperReference {
+   constructor() {
+    super(...arguments);
+    this[MUT] = true;
```
rwjblue: I vaguely recall @krisselden mentioning we should use meta for flags like this, maybe he can c/d?

rwjblue:
This is looking really good. I am no expert in the ember-glimmer techniques (I'll leave that to the rest of the team) but I am curious what you see as the roadmap for this PR.  What is still outstanding, what is blocking, etc...

One thing I was noticing/thinking about was the auto-mut AST transform, should that be reenabled in this PR?

Joelkang:
Rebased and addressed your concerns, @rwjblue . As far as a I can tell there's nothing really blocking this PR, once these issues are resolved

- [ ] [This line](https://github.com/emberjs/ember.js/pull/13541/files#diff-056a1b1ec3418c7dc38c91fb15fb05f4R70)
- [ ] Meta tag for MUT symbol? <-- @krisselden 
- [ ] Re-enable AST transform to make everything mut? <-- @chancancode

packages/ember-glimmer/lib/helpers/action.js
```diff
@@ -57,14 +63,15 @@ export class ClosureActionReference extends CachedReference {
       if (!action) {
         throw new EmberError(`An action named '${actionName}' was not found in ${target}`);
       }
+    } else if (action && typeof action[INVOKE] === 'function') {
+      target = action;
+      action = action[INVOKE];
     } else if (actionType !== 'function') {
-      throw new EmberError(`An action could not be made for \`${rawAction}\` in ${target}. Please confirm that you are using either a quoted action name (i.e. \`(action '${rawAction}')\`) or a function available in ${target}.`);
+      // FIXME: Is there a better way of doing this?
```
rwjblue: Can you change this to a `TODO:`?  I discussed with @chancancode and this seems okish for now, and we can circle back to it later...

rwjblue:
I was chatting with @chancancode about this abit today, he is going to try to give it more review sometime next week...

PR #13547, opened 2016-05-23
Fix failing htmlbars lifecyle tests

chadhietala:
@krisselden


mixonic:
@chadhietala I presume this has something to do with migrating to the `ember-glimmer` namespace? LGTM, the helper utils are available in both `ember-glimmer` and `ember-htmlbars`

PR #13549, opened 2016-05-23
ESLint

### LINKED ISSUE #13006: Replace JSHint/JSCS with ESLint
`ember-cli` has internally switched from using JSHint/JSCS to ESLint lately (see https://github.com/ember-cli/ember-cli/pull/5455) because it was considered unlikely that JSHint will include async/await support any time soon. I would like to propose switching to ESLint for `ember` too for similar reasons.

Note that this is not talking about forcing `ember` users to use ESLint in their project, this is only about linting the `ember` codebase itself.

Turbo87:
This PR adds ESLint code checking. JSHint/JSCS are not removed yet and ESLint is only part of the mocha test suite so far until `emberjs-build` supports ESLint too. Most of the JSHint/JSCS rule equivalents have been enabled and only two of the custom JSCS rules are missing at the moment and need to be reimplemented on the ESLint API.

partly resolves #13006 
- [x] port `require-comments-to-include-access` rule
- [x] port `disallow-const-outside-module-scope` rule

/cc @stefanpenner @rwjblue

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,3 @@
+blueprints/*/*files/**/*.js
+dist/
+tmp/
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,36 @@
+module.exports = {
+  root: true,
+  parserOptions: {
+    ecmaVersion: 6,
+    sourceType: 'module',
+  },
+  extends: 'eslint:recommended',
+  env: {
+    qunit: true,
+  },
+  globals: {
+    'expectAssertion': true,
+    'expectDeprecation': true,
+    'expectNoDeprecation': true,
+    'expectWarning': true,
+    'expectNoWarning': true,
+    'ignoreAssertion': true,
+    'ignoreDeprecation': true,
+
+    // A safe subset of "browser:true":
+    'window': true,
+    'document': true,
+    'setTimeout': true,
+    'clearTimeout': true,
+    'setInterval': true,
+    'clearInterval': true,
+
+    'Symbol': true,
+    'WeakMap': true,
+  },
+  rules: {
+    // temporarily disabled
+    'no-unused-vars': 'off',
+    'comma-dangle': 'off',
+  },
+};
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -29,6 +29,9 @@ module.exports = {
     'WeakMap': true,
   },
   rules: {
+    'require-yuidoc-access': 'error',
+    'no-const-outside-module-scope': 'error',
+
     // temporarily disabled
     'no-unused-vars': 'off',
     'comma-dangle': 'off',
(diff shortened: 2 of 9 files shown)


Turbo87:
All JSHint/JSCS equivalents have been ported and/or enabled now! 🎉

Turbo87:
looks like I'm being bitten by the `node_modules` caching on TravisCI... :-/

Turbo87:
I've extracted the custom ESLint rules into https://github.com/Turbo87/eslint-plugin-ember-internal to avoid the subproject issues that caused the previous CI failures

rwjblue:
Why are there so many file changes here?  I believe that we should be applying the existing rules only...

Turbo87:
@rwjblue except for the last commit: because JSHint apparently isn't that great at catching these things. It's been the same in all the other repos I've converted so far. ESLint usually found a lot more issues than JSHint.

packages/ember-glimmer/tests/integration/helpers/closure-action-test.js
```diff
@@ -689,10 +689,10 @@ moduleFor('Helpers test: closure {{action}}', class extends RenderingTest {
     });
 
     let OuterComponent = Component.extend({
-      init() {
-        this._super(...arguments);
-        outerComponent = this;
-      },
+      // init() {
```
stefanpenner: is this still WIP?

stefanpenner:
I'm 👍  on this.

packages/ember-glimmer/tests/integration/helpers/closure-action-test.js
Turbo87: The issue here was that `outerComponent` was unused. I wasn't sure if it was actually supposed to be used in a test somewhere so I commented it out for now. Let me know if I should remove is completely.

rwjblue: If the tests pass without it, just delete it.  We can always get it back if needed 😸

rwjblue:
Please squash

rwjblue:
> ESLint is only part of the mocha test suite so far until emberjs-build supports ESLint too

This must be fixed before landing.  The linting failures need to be included in the normal `ember test --server` output (just like in an ember-cli app).

Turbo87:
@rwjblue @stefanpenner would love to get some feedback on https://github.com/emberjs/emberjs-build/issues/161, then I'll start working on integrating ESLint into `ember test --server`.

Turbo87:
@rwjblue rebased and updated `emberjs-build` to v0.19.0

rwjblue:
Restarted travis build (was failing bizarrely)

Turbo87:
@rwjblue failure might be related to the fact that I didn't update the yarn lockfile

(and I still believe that it's stupid that it's failing because of that...)

Turbo87:
🍏

rwjblue:
Thanks for working so hard on this!

PR #13553, opened 2016-05-24
[FEATURE ember-runtime-enumerable-includes] Implements Array.includes and deprecates Array.contains

bmeurant:
Implementation of emberjs/rfcs#136

TODO : 
- [x]  Implement `Enumerable.includes` 
- [x]  Implement `Array.includes`
- [x]  Deprecate `contains` under feature flag
- [x]  Replace usages of `contains` by `includes`
- [x]  Update docs, guides (emberjs/guides#1439) & API. To be merged once feature enabled.
- [x] Write deprecation guide (emberjs/website#2600) & add url in deprecate call
- [x] Rebase / squash

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -57,3 +57,8 @@ for a detailed explanation.
   - `interaction.<event-name>` for events handled by a component.
   - `interaction.ember-action` for closure actions.
   - `interaction.link-to` for link-to execution.
+
+* `ember-runtime-enumerable-includes`
+
+Deprecates `Enumerable#contains` and `Array#contains` in favor of `Enumerable#includes` and `Array#includes` 
+to stay in line with ES standards (see [RFC](https://github.com/emberjs/rfcs/blob/master/text/0136-contains-to-includes.md)).
--- a/features.json
+++ b/features.json
@@ -10,6 +10,7 @@
     "ember-route-serializers": null,
     "ember-glimmer": null,
     "ember-runtime-computed-uniq-by": null,
-    "ember-improved-instrumentation": null
+    "ember-improved-instrumentation": null,
+    "ember-runtime-enumerable-includes": null
   }
 }
--- a/packages/ember-runtime/lib/mixins/array.js
+++ b/packages/ember-runtime/lib/mixins/array.js
@@ -30,6 +30,8 @@ import {
 import { meta as metaFor } from 'ember-metal/meta';
 import { markObjectAsDirty } from 'ember-metal/tags';
 import EachProxy from 'ember-runtime/system/each_proxy';
+import { deprecate } from 'ember-metal/debug';
+import isEnabled from 'ember-metal/features';
 
 function arrayObserversHelper(obj, target, opts, operation, notify) {
   var willChange = (opts && opts.willChange) || 'arrayWillChange';
@@ -112,7 +114,7 @@ export function isEmberArray(obj) {
   @since Ember 0.9.0
   @public
 */
-export default Mixin.create(Enumerable, {
+var ArrayMixin = Mixin.create(Enumerable, {
 
   [EMBER_ARRAY]: true,
 
@@ -214,6 +216,14 @@ export default Mixin.create(Enumerable, {
 
   // optimized version from Enumerable
   contains(obj) {
+    if (isEnabled('ember-runtime-enumerable-includes')) {
+      deprecate(
+        '`Enumerable#contains` is deprecated, use `Enumerable#includes` instead.',
+        false,
+        { id: 'ember-runtime.enumerable-contains', until: '3.0.0', url: 'http://emberjs.com/deprecations/v2.x#toc_enumerable-contains' }
+      );
+    }
+
     return this.indexOf(obj) >= 0;
   },
 
@@ -573,3 +583,55 @@ export default Mixin.create(Enumerable, {
     return this.__each;
   }).volatile()
 });
+
+if (isEnabled('ember-runtime-enumerable-includes')) {
+  ArrayMixin.reopen({
+    /**
+      Returns `true` if the passed object can be found in the array.
+      This method is a Polyfill for ES 2016 Array.includes.
+      If no `startAt` argument is given, the starting location to
+      search is 0. If it's negative, searches from the index of
+      `this.length + startAt` by asc.
+      ```javascript
+      [1, 2, 3].includes(2);     // true
+      [1, 2, 3].includes(4);     // false
+      [1, 2, 3].includes(3, 2);  // true
+      [1, 2, 3].includes(3, 3);  // false
+      [1, 2, 3].includes(3, -1); // true
+      [1, 2, 3].includes(1, -1); // false
+      [1, 2, 3].includes(1, -4); // true
+      [1, 2, NaN].includes(NaN); // true
+      ```
+      @method includes
+      @param {Object} obj The object to search for.
+      @param {Number} startAt optional starting location to search, default 0
+      @return {Boolean} `true` if object is found in the array.
+      @public
+    */
+    includes(obj, startAt) {
+      var len = get(this, 'length');
+      var idx, currentObj;
+
+      if (startAt === undefined) {
+        startAt = 0;
+      }
+
+      if (startAt < 0) {
+        startAt += len;
+      }
+
+      for (idx = startAt; idx < len; idx++) {
+        currentObj = objectAt(this, idx);
+
+        // SameValueZero comparison (NaN !== NaN)
+        if (obj === currentObj || (obj !== obj && currentObj !== currentObj)) {
+          return true;
+        }
+      }
+
+      return false;
+    }
+  });
+}
+
+export default ArrayMixin;
--- a/packages/ember-runtime/lib/mixins/enumerable.js
+++ b/packages/ember-runtime/lib/mixins/enumerable.js
@@ -29,6 +29,7 @@ import {
 } from 'ember-metal/events';
 import compare from 'ember-runtime/compare';
 import require from 'require';
+import { assert, deprecate } from 'ember-metal/debug';
 
 let _emberA;
 
@@ -231,6 +232,14 @@ var Enumerable = Mixin.create({
     @public
   */
   contains(obj) {
+    if (isEnabled('ember-runtime-enumerable-includes')) {
+      deprecate(
+        '`Enumerable#contains` is deprecated, use `Enumerable#includes` instead.',
+        false,
+        { id: 'ember-runtime.enumerable-contains', until: '3.0.0', url: 'http://emberjs.com/deprecations/v2.x#toc_enumerable-contains' }
+      );
+    }
+
     var found = this.find(function(item) {
       return item === obj;
     });
@@ -797,8 +806,8 @@ var Enumerable = Mixin.create({
 
   /**
     Returns a new enumerable that excludes the passed value. The default
-    implementation returns an array regardless of the receiver type unless
-    the receiver does not contain the value.
+    implementation returns an array regardless of the receiver type.
+    If the receiver does not contain the value it returns the original enumerable.
 
     ```javascript
     var arr = ['a', 'b', 'a', 'c'];
@@ -1110,6 +1119,65 @@ if (isEnabled('ember-runtime-computed-uniq-by')) {
         if (!(guid in seen)) {
           seen[guid] = true;
           ret.push(item);
+        }
+      });
+
+      return ret;
+    }
+  });
+}
+
+if (isEnabled('ember-runtime-enumerable-includes')) {
+  Enumerable.reopen({
+    /**
+      Returns `true` if the passed object can be found in the enumerable.
+      ```javascript
+      [1, 2, 3].includes(2);                     // true
+      [1, 2, 3].includes(4);                     // false
+      [1, 2, undefined].includes(undefined);     // true
+      [1, 2, null].includes(null);               // true
+      [1, 2, NaN].includes(NaN);                 // true
+      ```
+      @method includes
+      @param {Object} obj The object to search for.
+      @return {Boolean} `true` if object is found in the enumerable.
+      @public
+    */
+    includes(obj) {
+      assert('Enumerable#includes cannot accept a second argument "startAt" as enumerable items are unordered.', arguments.length === 1);
+
+      var len = get(this, 'length');
+      var idx, next;
+      var last = null;
+      var found = false;
+
+      var context = popCtx();
+
+      for (idx = 0; idx < len && !found; idx++) {
+        next = this.nextObject(idx, last, context);
+
+        found = obj === next || (obj !== obj && next !== next);
+
+        last = next;
+      }
+
+      next = last = null;
+      context = pushCtx(context);
+
+      return found;
+    },
+
+    without(value) {
+      if (!this.includes(value)) {
+        return this; // nothing to do
+      }
+
+      var ret = emberA();
+
+      this.forEach(function(k) {
+        // SameValueZero comparison (NaN !== NaN)
+        if (!(k === value || k !== k && value !== value)) {
+          ret[ret.length] = k;
         }
       });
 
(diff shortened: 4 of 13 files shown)


alexspeller:
Oh awesome, thanks for getting this implementation done!

bmeurant:
Here is a first proposal for emberjs/rfcs#136. Work is in progress and feedbacks expected.

I made the asumption that we have to keep `Enumerable` & `Array` APIs consistent and then have to provide a consistent `includes` method in `Enumerable` (with no index). Let me know if this is not expected.

Aliasing `contains` with the new `includes` method can break some existing apps because, according to the spec (https://tc39.github.io/ecma262/2016/#sec-array.prototype.includes), `includes` relies on the `SameValueZero` algorithm and actual `contains` relies on strict equality leading to differences in `NaN` detections (I will document it). Can we still make this ?

mixonic:
@bmeurant Thanks :-)

`contains` should not be aliased to `includes`. Contains should keep its current implementation, and we can deprecate it though we will likely do that in a separate PR. In this PR it should simply retain its current behavior.

bmeurant:
@mixonic thx for your feedback.

I updated this PR to keep current implementation for `Enumerable.contains` & `Array.contains`.

> we can deprecate it though we will likely do that in a separate PR

Does it means that I should also remove `@deprecated` and deprecation warning for now ?

rwjblue:
I believe that the deprecation should be included here when the feature flag is enabled.

mixonic:
@rwjblue I am easily convinced, was just erring on the conservative side. @bmeurant please include the deprecation behind the feature flag! :-)

bmeurant:
@mixonic I added feature flag and move deprecation behind it. Let me know if I forgotten something.

I also removed `@deprecated` from API because `contains` is not actuelly deprecated if feature is not enabled, right ?

But I was wondering: doing that, users will loose the information that `contains` will be deprecated soon and that they should use `includes` instead. Is it exists a way to provide this kind of information in APIs ?

mixonic:
@bmeurant Yeah the documentation may require some followup tweaks. Feature flagging things like the `@deprecated` flag is tricky :-/. I think you should feel free to create a followup issue of things to do once the feature is "go'd" for beta.

FEATURES.md
```diff
@@ -56,3 +56,8 @@ for a detailed explanation.
 
   - `interaction.<event-name>` for events handled by a component.
   - `interaction.ember-action` for closure actions.
+
+* `ember-runtime-enumerable-contains`
+
+Deprecates `Enumerable.contains` and `Array.contains` in favor of `Enumerable.includes` and `Array.includes`
```
mixonic: Again here I lean toward `Enumerable#contains`, per https://github.com/emberjs/website/pull/2600#discussion_r64891064. It is a Rubyish approach, curious if there are other thoughts.

packages/ember-runtime/lib/mixins/enumerable.js
```diff
@@ -239,6 +248,27 @@ var Enumerable = Mixin.create({
...
+    [1, 2, 3].includes(2);     // true
+    [1, 2, 3].includes(4);     // false
+    [1, 2, NaN].includes(NaN); // true
+    ```
+    @method includes
+    @param {Object} obj The object to search for.
+    @return {Boolean} `true` if object is found in the enumerable.
+    @public
+  */
+  includes(obj) {
```
mixonic: Can you add an assertion that there is no second argument? Something like:

```js
  includes(obj) {
    assert('Enumerable#includes cannot accept a second argument "startAt" as enumerable items are unordered.', arguments.length === 1);
```

FEATURES.md
alexspeller: I think `Enumerable#contains` is the convention in Ember even though it was originally a rubyism - at least it has the advantage that it's not valid syntax for something that is completely different

packages/ember-runtime/lib/mixins/mutable_array.js
```diff
@@ -388,7 +388,7 @@ export default Mixin.create(EmberArray, MutableEnumerable, {
     @public
   */
   addObject(obj) {
-    if (!this.contains(obj)) {
+    if (!this.includes(obj)) {
```
mixonic: This is a small breaking change. I've left a note that it be called out in the deprecation guide https://github.com/emberjs/website/pull/2600#issuecomment-222124560. Here and `without`.

packages/ember-runtime/tests/suites/array/includes.js
```diff
@@ -0,0 +1,35 @@
+import {SuiteModuleBuilder} from 'ember-runtime/tests/suites/suite';
+
+var suite = SuiteModuleBuilder.create();
+
+suite.module('includes');
```
mixonic: Should include a test/tests without a starting position

mixonic:
@bmeurant heh, there are no tests for `Array#contains`?! Can you add one asserting basic behavior and the deprecation please?

This is looking pretty good after some tweaks!

packages/ember-runtime/lib/mixins/array.js
```diff
@@ -214,9 +216,63 @@ export default Mixin.create(Enumerable, {
...
+    [1, 2, 3].includes(1, -4); // true
+    [1, 2, NaN].includes(NaN); // true
+    ```
+    @method includes
+    @param {Object} obj The object to search for.
+    @param {Number} startAt optional starting location to search, default 0
+    @return {Boolean} `true` if object is found in the array.
+    @public
+  */
+  includes(obj, startAt) {
```
mixonic: This function (and the enumerable one) must be added only when the feature flag is enabled. For example:

```js
let ArrayMixin = Mixin.create(Enumerable, {
 /* snip */
});
if (isEnabled('ember-runtime-enumerable-includes')) {
  ArrayMixin.reopen({
    includes(obj, startAt) {
      /* ... */
    }
  });
}
export default ArrayMixin;
```

packages/ember-runtime/tests/suites/enumerable/includes.js
```diff
@@ -0,0 +1,28 @@
+import {SuiteModuleBuilder} from 'ember-runtime/tests/suites/suite';
+
+var suite = SuiteModuleBuilder.create();
+
+suite.module('includes');
+
+suite.test('includes returns true if items is in enumerable', function() {
```
mixonic: Once the method itself is behind a feature flag, you will also need to wrap these tests inside an `isEnabled` check.

features.json
```diff
@@ -10,6 +10,7 @@
     "ember-route-serializers": null,
     "ember-glimmer": null,
     "ember-runtime-computed-uniq-by": null,
-    "ember-improved-instrumentation": null
+    "ember-improved-instrumentation": null,
+    "ember-runtime-enumerable-contains": null
```
mixonic: Lastly the feature flag here seems poorly named. Perhaps `ember-runtime-enumerable-includes`

packages/ember-runtime/lib/mixins/enumerable.js
```diff
@@ -239,6 +248,27 @@ var Enumerable = Mixin.create({
...
+    var found = this.find(function(item) {
+      // SameValueZero comparison (NaN !== NaN)
+      return item === obj || (item !== item && obj !== obj);
+    });
+
+    return found !== undefined;
```
mixonic: Wouldn't this break for `includes(undefined)`? Can you add a test?

bmeurant:
Thx for feedbacks, I'll work on it ASAP

packages/ember-runtime/tests/suites/array/includes.js
bmeurant: Tests without starting position are imported from Enumerable test suite, no ?

bmeurant:
It seemed to me that Array#contains tests are also imported and run from Enumerable#contains tests. Did I missed something ?

mixonic:
Ah I see @bmeurant :-) right there is no `contains` on arrays, only on enumerable 👍

packages/ember-runtime/tests/suites/array/includes.js
mixonic: I'm unsure what you mean, maybe I am missing something. This code tests the implementation of `Array#includes`. How are the tests for `Enumerable#includes` applied to an array?

packages/ember-runtime/lib/mixins/mutable_array.js
bmeurant: Sure. Moreover, I think I have to use includes instead of contains only behind feature flag, right?

mixonic: 👍 yup

packages/ember-runtime/lib/mixins/enumerable.js
bmeurant: Right ! I missed it, sorry :-(

packages/ember-runtime/lib/mixins/mutable_array.js
```diff
@@ -388,7 +389,9 @@ export default Mixin.create(EmberArray, MutableEnumerable, {
     @public
   */
   addObject(obj) {
-    if (!this.contains(obj)) {
+    var included = isEnabled('ember-runtime-enumerable-includes') ? this.includes(obj) : this.contains(obj);
```
bmeurant: Impacts on `addObject` are not so small because using `includes` instead of  `contains` here will allow to add `null` and `undefined` values into an Enumerable ... Is it expected ?

bmeurant:
I wanted to add a test to verify assertion calling `Enumerable#includes` with a second argument but, because tests defined on Enumerable are also run on Array suites, I did not find a good way to do it. Any advice ?

packages/ember-runtime/lib/mixins/enumerable.js
```diff
@@ -811,7 +820,9 @@ var Enumerable = Mixin.create({
     @public
   */
   without(value) {
```
bmeurant: I think we have a problem line 832 during `k !== value`. If the feature is enabled, `without` become inconsistent because `NaN !== NaN` is `true`

packages/ember-runtime/lib/mixins/enumerable.js
```diff
@@ -1118,4 +1127,63 @@ if (isEnabled('ember-runtime-computed-uniq-by')) {
...
+        last = next;
+      }
+
+      next = last = null;
+      context = pushCtx(context);
+
+      return found;
+    },
+
+    without(value) {
```
bmeurant: I prefered replacing `without` behind feature flag to not add multiple `isEnabled` calls in original method

packages/ember-runtime/lib/mixins/array.js
```diff
@@ -214,6 +216,14 @@ export default Mixin.create(Enumerable, {
 
   // optimized version from Enumerable
   contains(obj) {
+    if (isEnabled('ember-runtime-enumerable-includes')) {
+      deprecate(
+        '`Enumerable#contains` is deprecated, use `Enumerable#includes` instead.',
+        false,
+        { id: 'ember-runtime.enumerable-contains', until: '3.0.0', url: 'http://emberjs.com/deprecations/v2.x#toc_enumerable-contains' }
```
bmeurant: I did not change deprecation id because feature is about `includes` but deprecation is about  `contains`.

But I don't know if it is accepted. Let me know

mixonic: the deprecation id and feature id are totally distinct. Seems fine to use different names 👍

packages/ember-runtime/lib/mixins/enumerable.js
mixonic: 👍

packages/ember-runtime/lib/mixins/enumerable.js
```diff
@@ -1118,4 +1127,63 @@ if (isEnabled('ember-runtime-computed-uniq-by')) {
...
+      if (!this.includes(value)) {
```
mixonic: Calling `includes` before the `forEach` means we walk the contents of the array twice. Additionally it seems a bit unexpected for `without()` to sometimes return a new array, and sometimes return the array passed in.

I'd prefer to drop the `includes` check and just run the array builder.

packages/ember/tests/component_registration_test.js
```diff
@@ -39,9 +40,13 @@ function cleanup() {
 }
 
 function cleanupHelpers() {
+  var included;
+
   keys(helpers).
     forEach((name) => {
-      if (!originalHelpers.contains(name)) {
+      included = isEnabled('ember-runtime-enumerable-includes') ? originalHelpers.includes(name) : originalHelpers.contains(name);
```
mixonic: Unfortunately the ternary syntax is not supported by feature flag stripping. Please rewrite this to use an `if` statement.

packages/ember-runtime/lib/mixins/mutable_array.js
mixonic: Unfortunately the ternary syntax is not supported by feature flag stripping. Please rewrite this to use an `if` statement.

mixonic:
@bmeurant it seems ok to add a test for `contains` without an offset in `packages/ember-runtime/tests/mixins/enumerable_test.js`, since it only impacts enumerables.

mixonic:
Thanks for your continued work on this 😄 it looks really close, and quite nice!

packages/ember-runtime/lib/mixins/enumerable.js
bmeurant: I think I'd also prefer but it seems that returning the same instance if not included (or not contained), is part of the API since it is tested here : https://github.com/emberjs/ember.js/blob/master/packages/ember-runtime/tests/suites/enumerable/without.js#L19

Does this PR should also change this behaviour?

I could make it, of course, and update the deprecation guide. But in any case, I propose to update `without` API doc to explicitely mention it.

mixonic: 👍  to updating the docs and keeping the current behavior. Great catch.

The `without` method on enumerables and on mixins just should just 🔥 🔥 🔥. On enumerables we can just use `reject` which seems more sane, and on mixins the functionality is only used for native array prototype application. I attempted killing it once and ran out of steam.

mixonic:
Looking nice. A rebase/squash to some number is semantically meaningful commits (one or more) would be great, each prefixed with `[FEATURE ember-runtime-enumerable-includes]` would be great.

bmeurant:
Done :ok_hand:

mixonic:
Looks pretty great to me @bmeurant! Thank you! I'll get someone else on core to review and we can :shipit: with their 👍

packages/ember-runtime/lib/mixins/array.js
```diff
@@ -573,3 +583,55 @@ export default Mixin.create(Enumerable, {
...
+      @param {Object} obj The object to search for.
+      @param {Number} startAt optional starting location to search, default 0
+      @return {Boolean} `true` if object is found in the array.
+      @public
+    */
+    includes(obj, startAt) {
+      var len = get(this, 'length');
+      var idx, currentObj;
+
+      if (Ember.isNone(startAt)) {
```
rwjblue: Should not use the global here (`Ember.isNone`), if we need `isNone` it should be imported.  Can we just do the `null` / `undefined` check ourselves?

rwjblue:
This looks good, only one small tweak and this is good to go...

packages/ember-runtime/lib/mixins/array.js
bmeurant: @rwjblue using the global here was surely a mistake, sorry. `isNone` is already imported and used in `slice`. I'll update. But why do you prefer to not use imported `isNone` and doing `null` / `undefined` check ourselves in this case?

bmeurant: Anyway, explicit null check is unnecessary as we only iterating on `idx = startAt`. I removed it.

mixonic:
This is great! Thank you @bmeurant!

rwjblue:
FYI - I created https://github.com/rwjblue/ember-runtime-enumerable-includes-polyfill to make this a bit easier for addons to avoid deprecations, use the newer syntax, and continue to support older versions of Ember.

PR #13556, opened 2016-05-25
[DOC RELEASE] Clarify that types are not preserved by computed.min and max

Gaurav0:
This clarification is suggested in issue #13387


acorncom:
@rwjblue r+?

mixonic:
Thanks @Gaurav0!

PR #13560, opened 2016-05-25
[Glimmer] Unskip component registration tests

nickiaconis:
Title says it all.


packages/ember-views/tests/utils/helpers.js
```diff
@@ -0,0 +1,10 @@
+import isEnabled from 'ember-metal/features';
+import require from 'require';
+
+export let OutletView;
+
+if (isEnabled('ember-glimmer')) {
+  OutletView = require('ember-glimmer/views/outlet').default;
+} else {
+  OutletView = require('ember-htmlbars/views/outlet').OutletView;
```
krisselden: ember-templates already does this branch

krisselden: nevermind, I thought it did, but I was wrong. I'm trying to get away from ember-views requiring ember-glimmer or ember-htmlbars, I'm trying to have it just be where stuff that is shared not branched is.

krisselden: if this is used from a test in packages/ember/tests only can you move it there?

nickiaconis: @krisselden Makes sense. I've moved the file into the ember package.

chadhietala:
LGTM

packages/ember-glimmer/lib/utils/lookup-component.js
```diff
@@ -7,14 +8,40 @@ function lookupComponentPair(componentLookup, owner, name, options) {
...
 
-  if (!layout && component) {
-    let layoutProp = component.proto().layout;
-    if (layoutProp) {
-      let templateFullName = 'template:components/' + name;
-      owner.register(templateFullName, layoutProp);
+  if (component) {
+    let templateFullName = 'template:components/' + name;
+    let {
+      templateForName,
```
krisselden: why do we mixin this into component and pull it off? can we just drop it

packages/ember-glimmer/lib/utils/lookup-component.js
```diff
@@ -7,14 +8,40 @@ function lookupComponentPair(componentLookup, owner, name, options) {
...
+      owner.register(templateFullName, defaultLayoutProp);
       layout = owner.lookup(templateFullName, options);
-    } else {
-      layout = owner.lookup(DEFAULT_LAYOUT);
+    } else if (!layout) {
+      if (layoutProp) {
+        owner.register(templateFullName, layoutProp);
+        layout = owner.lookup(templateFullName, options);
+      } else if (layoutNameProp) {
+        layout = templateForName(layoutNameProp, null, owner);
```
krisselden: this is weird, lets just move the templateForName logic to the mixin that is actually calling it.

packages/ember-glimmer/lib/utils/lookup-component.js
nickiaconis: @krisselden Htmlbars expects it to be on the view/component, but component init runs too late for the same logic to apply in glimmer. I was trying to keep things DRY by pulling templateForName off the class, but I can WET-ify that function if it's preferred.

krisselden: I think we should emit a different opcode or the dynamic opcode when a layout is late bound and assert on the layout property like the tag expression is asserted.

krisselden: maybe just use the partial opcode for the layout

chadhietala:
Superseded by #13588.

PR #13562, opened 2016-05-26
Remove fragmentReason meta that was for the removed angleBrackets

krisselden:
This should have bee removed with the angle bracket feature was removed


chadhietala:
Seems good. Pinging @zackthehuman just to make sure.

zackthehuman:
LGTM. I don't know of any usage of fragmentReason in my addons, so I think this is good :)

PR #13563, opened 2016-05-26
[BUGFIX beta] Change error message when Ember.computed.and/or contain…

### LINKED ISSUE #13176: Confusing error when using `Ember.computed.or` incorrectly

chena:
Fixes #13176

### DIFF
--- a/packages/ember-runtime/lib/computed/computed_macros.js
+++ b/packages/ember-runtime/lib/computed/computed_macros.js
@@ -12,7 +12,7 @@ import expandProperties from 'ember-metal/expand_properties';
 @submodule ember-metal
 */
 
-function expandPropertiesToArray(properties) {
+function expandPropertiesToArray(predicateName, properties) {
   var expandedProperties = [];
 
   function extractProperty(entry) {
@@ -21,7 +21,7 @@ function expandPropertiesToArray(properties) {
 
   for (let i = 0; i < properties.length; i++) {
     let property = properties[i];
-    assert('Dependent keys passed to Ember.computed.and() or Ember.computed.or() can\'t have spaces.', property.indexOf(' ') < 0);
+    assert(`Dependent keys passed to Ember.computed.${predicateName}() can\'t have spaces.`, property.indexOf(' ') < 0);
 
     expandProperties(property, extractProperty);
   }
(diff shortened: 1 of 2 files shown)


courajs:
Hey @rwjblue, how's this look?

packages/ember-runtime/lib/computed/computed_macros.js
```diff
@@ -20,7 +20,10 @@ function expandPropertiesToArray(properties) {
   }
 
   for (let i = 0; i < properties.length; i++) {
-    expandProperties(properties[i], extractProperty);
+    let property = properties[i];
+    assert('Dependent keys passed to Ember.computed.and() or Ember.computed.or() can\'t have spaces.', property.indexOf(' ') < 0);
```
rwjblue: Looks good, but I'd much prefer the message to be tailored to the specific type of computer macro that was being used. Is this same message fired for ither macros? If it is, can we somehow pass in the macro being used and use that in the message directly?

chena: @rwjblue @courajs I wonder if the only way to  know which macro is being processed is to pass it along explicitly (ie. "and" or "or"), since the predicate from the caller `generateComputedWithPredicate` is a callback that can be arbitrary?

rwjblue: Ya, that was what I was worried about. How hard would it be do do that?

courajs: Adding a second argument `label` to `generateComputedWithPredicate` and `expandPropertiesToArray` should work out well. It seemed a little awkward to me at first, but it leads to better messages for the user. If they've used both `and` and `or` in their app, this will help them narrow it down.

courajs:
LGTM!

stefanpenner:
👍

courajs:
Congrats @chena on first PR! Thanks :)

PR #13568, opened 2016-05-26
WIP [CLEANUP] Removing tests dependent on  on views

chadhietala:
@rwjblue While this is still a WIP it would be great to get feedback about these removals. I believe these are tests that are testing `templateName` which actually doesn't give you the ability to swap templates in Ember.Component.


packages/ember/tests/routing/basic_test.js
```diff
@@ -2127,46 +2107,6 @@ test('The rootURL is passed properly to the location implementation', function()
   bootApplication();
 });
 
-
-test('Only use route rendered into main outlet for default into property on child', function() {
```
rwjblue: This one doesn't seem related to templateName, you can likely remove the templateName usage below and allow menu/view template be resolved.

rwjblue:
One minor comment, but looks good.

rwjblue:
Rebased in #13638.

PR #13570, opened 2016-05-27
Remove obsolete "bin/git_wrapper" script and "Rakefile"

Turbo87:
This script was once used by the `Rakefile` but isn't referenced there anymore. I didn't find any other references to it in this repo either so I assume it is no longer used.


locks:
@Turbo87 could you remove Rakefile while you're at it? I thought I had sent a PR, but looks otherwise.

Turbo87:
@locks with pleasure!

rwjblue:
Thanks!

PR #13571, opened 2016-05-27
[BUGFIX] Update helper integration test blueprints

Gaurav0:
So since I created the helper integration test blueprints for https://github.com/ember-cli/ember-cli/issues/4517 , blueprints and node tests for blueprints have been added to the ember.js repo. This changes the appropriate blueprints for the ember addon and adds appropriate tests.


blueprints/helper-test/mocha-files/tests/__testType__/helpers/__name__-test.js
```diff
@@ -0,0 +1,40 @@
+/* jshint expr:true */
+import { expect } from 'chai';
+<% if (testType == 'integration') { %>
+import { describeComponent, it } from 'ember-mocha';
+import hbs from 'htmlbars-inline-precompile';
+
+  describeComponent('<%= dasherizedModuleName %>', 'helper:<%= dasherizedModuleName %>',
```
Turbo87: any reason for the additional indentation here?

blueprints/helper-test/qunit-files/tests/__testType__/helpers/__name__-test.js
```diff
@@ -0,0 +1,28 @@
...
+  });
+
+// Replace this with your real tests.
+  test('it renders', function(assert) {
+    this.set('inputValue', '1234');
+
+    this.render(hbs`{{<%= dasherizedModuleName %> inputValue}}`);
+
+    assert.equal(this.$().text().trim(), '1234');
+  });
```
Turbo87: indentation here seems wrong too

package.json
```diff
@@ -30,16 +30,21 @@
     "ember-cli": "^2.4.2",
     "ember-cli-blueprint-test-helpers": "^0.12.0",
     "ember-cli-dependency-checker": "^1.2.0",
+    "ember-cli-internal-test-helpers": "^0.7.0",
+    "ember-cli-is-package-missing": "^1.0.0",
+    "ember-router-generator": "^1.2.1",
```
Turbo87: where are those "new" dependencies used?

Gaurav0: I got they were missing when running `npm run test:blueprints`

Gaurav0: I assumed therefore tests would fail without them.

blueprints/helper-test/mocha-files/tests/__testType__/helpers/__name__-test.js
Gaurav0: I think my IDE trolled me. Fixing.

package.json
Turbo87: hmm... that's surprising since it seems to run fine without them on Travis and locally here. you haven't modified any code or required them anywhere so you shouldn't need to add them. is it possible that some other dependencies on your machine were not up-to-date?

Gaurav0: You're right.

I can see that is-package-missing is required from https://github.com/emberjs/ember.js/blob/master/blueprints/helper-test/index.js

I'm wondering how blueprint tests can run without these. Is there a place where test only dependencies are specified? Or could it be that I'm running npm v2?

Gaurav0: Yes, ember-cli dependencies (not devDependencies) should be able to be looked up directly in node-modules. My install is borked.

Turbo87: > I'm wondering how blueprint tests can run without these. Is there a place where test only dependencies are specified? Or could it be that I'm running npm v2 and thus not picking up ember-cli dependencies in root of node_modules?

npm@2 could indeed be the reason why it's failing for you. good catch anyway, would be cool though if you could extract that dependency fix into another PR or at least commit.

Gaurav0:
@Turbo87 Addressed comments. I hope this is good to go.

If I can reproduce npm2 issue on another machine, I will open it separately.

Turbo87:
LGTM 👍

rwjblue:
Thank you @Gaurav0!

PR #13573, opened 2016-05-27
Unskip some basic routing tests

asakusuma:
What should we do about tests like [this](https://github.com/emberjs/ember.js/blob/master/packages/ember/tests/routing/basic_test.js#L368-L386) that explicitly test `Ember.View` features?

``` JavaScript
App.HomeView = EmberView.extend({
  templateName: 'the_real_home_template'
});
```

### DIFF
--- a/packages/ember/tests/routing/basic_test.js
+++ b/packages/ember/tests/routing/basic_test.js
@@ -321,8 +321,8 @@ QUnit.test('Model passed via renderTemplate model is set as controller\'s model'
   equal(jQuery('p:contains(emberjs)', '#qunit-fixture').length, 1, 'Passed model was set as controllers model');
 });
 
-test('Renders correct view with slash notation', function() {
-  setTemplate('home/page', compile('<p>{{view.name}}</p>'));
+QUnit.test('Renders correct view with slash notation', function() {
+  setTemplate('home/page', compile('<p>put some respeck on it</p>'));
 
   Router.map(function() {
     this.route('home', { path: '/' });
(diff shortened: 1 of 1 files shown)


chadhietala:
What is this testing?

asakusuma:
@chadhietala I believe this is testing that `route.render` renders the right template.

packages/ember/tests/routing/basic_test.js
```diff
@@ -321,8 +321,8 @@ QUnit.test('Model passed via renderTemplate model is set as controller\'s model'
   equal(jQuery('p:contains(emberjs)', '#qunit-fixture').length, 1, 'Passed model was set as controllers model');
 });
 
-test('Renders correct view with slash notation', function() {
-  setTemplate('home/page', compile('<p>{{view.name}}</p>'));
+QUnit.test('Renders correct view with slash notation', function() {
```
mixonic: This entire test should be removed or renamed.

This test explicitly tests the "correct view" is rendered. Since views are gone in Canary, we no longer need to test that they render. Obviously `App.HomePageView` is no longer used here, so that variable can do away. The title of this test is also wrong, as your update drops the use of a view.

We should keep a test that `Renders correct template with slash notation` and perhaps even `Renders correct controller with slash notation`. Perhaps those are elsewhere in this file, perhaps we should add them.


/cc @krisselden @rwjblue

packages/ember/tests/routing/basic_test.js
```diff
@@ -340,11 +340,11 @@ test('Renders correct view with slash notation', function() {
 
   bootApplication();
 
-  equal(jQuery('p:contains(Home/Page)', '#qunit-fixture').length, 1, 'The homepage template was rendered');
+  equal(jQuery('p:contains(put some respeck on it)', '#qunit-fixture').length, 1, 'The homepage template was rendered');
 });
 
-test('Renders the view given in the view option', function() {
-  setTemplate('home', compile('<p>{{view.name}}</p>'));
+QUnit.test('Renders the view given in the view option', function() {
```
mixonic: Canary is dropping support for views, and thus also the `view` option to `render`. This test can be removed.

asakusuma:
I've opened up [another PR](https://github.com/emberjs/ember.js/pull/13579) for removing tests.

I'll keep this one open for the tests in this file that need modification to keep the original intent but without using views.

PR #13574, opened 2016-05-27
[Glimmer] Make helpers amenable to testing

nickiaconis:


packages/ember-views/lib/views/core_view.js
```diff
@@ -7,9 +7,9 @@ import ActionHandler, { deprecateUnderscoreActions } from 'ember-runtime/mixins/
 import { typeOf } from 'ember-runtime/utils';
 
 import { InteractiveRenderer } from 'ember-htmlbars/renderer';
+import DOMHelper from 'ember-htmlbars/system/dom-helper';
```
krisselden: why did you change this?  this was for a path only used by tests in ember-views we are working to get rid of.

krisselden:
I'm not sure why we are testing helpers in ember-views, are these not covered by tests in ember-glimmer/ember-htmlbars?

packages/ember-application/tests/system/dependency_injection/default_resolver_test.js
```diff
@@ -91,22 +88,6 @@ QUnit.test('the default resolver resolves *:main on the namespace', function() {
   detectEqual(application.FooBar, locator.lookupFactory('foo-bar:main'), 'looks up FooBar type without name on application');
 });
 
-QUnit.test('the default resolver resolves helpers', function() {
```
mixonic: lol

mixonic:
looks pretty good pending some followup on @krisselden's comment.

packages/ember-views/lib/views/core_view.js
nickiaconis: @krisselden I was trying to make it so this could use `ember-htmlbars/system/dom-helper` or `ember-glimmer/dom` as necessary in order to make tests pass, but then I talked to @chadhietala regarding these. He said "Don’t port [views] They will be deleted", so I need to revise this.

Should the tests (outside of the ember-views package) that create a view in order to test something about helpers be updated to create a component instead?

nickiaconis:
Updated to address krisselden's feedback and rebased with upstream changes.

packages/ember-glimmer/tests/utils/global-helpers-hash.js
```diff
@@ -0,0 +1 @@
+../../../ember-htmlbars/lib/helpers.js
```
rwjblue: This is slightly confusing to me. Why would ember-glimmer point at ember-htmlbars helpers?

Can you help me understand?

rwjblue:
I think this has been superseded by other PR's that have landed since this was submitted.

PR #13577, opened 2016-05-30
Ember.assign: Prevent collision of index variable (i)

frederikbosch:
Current code in Ember.assign uses the variable `i` twice. This is error prone.


stefanpenner:
With `let` it is not error prone, as let is correctly scoped

frederikbosch:
Oops, I should work on my ES6 skills. Sorry for the interruption.

stefanpenner:
@frederikbosch not a problem, pre ES6 this would be very much something we would have concerned ourselves with. Luckily, ES6 brings along some nice things.

PR #13579, opened 2016-05-31
Remove view tests

asakusuma:
These tests either test a view specific feature, or already have test
coverage in the analogous controller/template feature.

cc @chadhietala @mixonic


packages/ember/tests/routing/basic_test.js
```diff
@@ -321,90 +321,6 @@ QUnit.test('Model passed via renderTemplate model is set as controller\'s model'
   equal(jQuery('p:contains(emberjs)', '#qunit-fixture').length, 1, 'Passed model was set as controllers model');
 });
 
-test('Renders correct view with slash notation', function() {
```
asakusuma: [Here](https://github.com/asakusuma/ember.js/blob/45d0a7957e1b22233ab3f94bba8fc652306c7bd3/packages/ember/tests/routing/basic_test.js#L1876) is the analogous template test

packages/ember/tests/routing/basic_test.js
```diff
@@ -321,90 +321,6 @@ QUnit.test('Model passed via renderTemplate model is set as controller\'s model'
...
-  App.HomePageView = EmberView.extend({
-    name: 'Home/Page'
-  });
-
-  bootApplication();
-
-  equal(jQuery('p:contains(Home/Page)', '#qunit-fixture').length, 1, 'The homepage template was rendered');
-});
-
-test('Renders the view given in the view option', function() {
```
chadhietala: I don't know if this can be removed or not. This is talking about things registered as `view:${someView}` in the registry. If it's no longer supported we can't just delete the test the feature needs to be removed as well. This means you have to clean up all of this code. https://github.com/emberjs/ember.js/blob/master/packages/ember-routing/lib/system/route.js#L1962-L1978

asakusuma: Gotcha. I'll work on cleaning that up.

asakusuma:
A couple questions.

1. Should [`disconnectOutlet`](https://github.com/emberjs/ember.js/blob/master/packages/ember-routing/lib/system/route.js#L2025-L2049) be removed since we are removing Views? In other words, does it have any use in a world without views?
2. Was view support technically removed in `2.x`? Since removing this view stuff is backwards incompatible with views.

packages/ember/tests/routing/basic_test.js
chadhietala: What I'm saying is the API was never deprecated for 2.4 so I'm not sure if can be removed. This is a question for someone on core.

chadhietala:
@asakusuma `View` and `{{view}}` as a public API is no longer supported as of 2.4. It will continue to be used internally for a while. So just because you see `View` doesn't mean you can delete it. You need to assess what is actually being tested. Since Components currently share a subset of functionality that `View` had we need to make sure we are not loosing coverage. It is highly likely we were piggy backing off of `View` tests for coverage on `Component`s.

packages/ember/tests/routing/basic_test.js
asakusuma: @mixonic gave the 👍 for removal: https://github.com/emberjs/ember.js/pull/13573#discussion_r65086409

krisselden:
http://emberjs.com/api/classes/Ember.Route.html#method_disconnectOutlet

krisselden:
disconnectOutlet is the opposite of manually renderTemplate() calls in route.

PR #13581, opened 2016-06-01
Bump glimmer; allow dynamic components to be optional

zackthehuman:
This un-gates a test for Glimmer to allow dynamic components to "go missing" when their path is falsy. This means that a component will not be rendered instead of throwing an exception (which is what it does now). This PR is complimentary to https://github.com/tildeio/glimmer/pull/177.


chadhietala:
LGTM. Just waiting on CI.

locks:
I kicked the build and it's passing now 😁

krisselden:
glimmer has been bumped pass this, can you remove the package.json change?  This PR may have been superseded already.

zackthehuman:
rebased and updated the component code

PR #13585, opened 2016-06-02
[GLIMMER 2] Throw if {{link-to}} syntax is wrong

chadhietala:


packages/ember-htmlbars/lib/components/link-to.js
```diff
@@ -758,10 +758,20 @@ let LinkComponent = EmberComponent.extend({
   willRender() {
     let queryParams;
 
-    // Do not mutate params in place
-    let params = get(this, 'params').slice();
+    let params = get(this, 'params');
```
chadhietala: Based on how `processedArgs` for `positionalParams` in glimmer we must guard here.

PR #13587, opened 2016-06-02
[Pre-Glimmer2] Remove references to HTMLBars and keywords

nickiaconis:
Remove all references to HTMLBars and keywords in test file that covers Ember's test helpers.

Q: Should I also remove all references to View? Or is okay to do that as part of un-skipping tests for Glimmer?

### DIFF
--- a/packages/ember-testing/tests/helpers_test.js
+++ b/packages/ember-testing/tests/helpers_test.js
@@ -3,8 +3,8 @@ import Controller from 'ember-runtime/controllers/controller';
 import run from 'ember-metal/run_loop';
 import EmberObject from 'ember-runtime/system/object';
 import RSVP from 'ember-runtime/ext/rsvp';
+import Component from 'ember-templates/component';
 import EmberView from 'ember-views/views/view';
-import Checkbox from 'ember-htmlbars/components/checkbox';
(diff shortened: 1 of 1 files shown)


mixonic:
@nickiaconis open to feedback from others, but I don't believe there is any specific timing around the process of cleaning up tests (View based or otherwise). This is a good step forward!

packages/ember-testing/tests/helpers_test.js
```diff
@@ -355,27 +350,29 @@ test('`click` triggers appropriate events in order', function() {
...
     didInsertElement() {
       this.$().on('mousedown focusin mouseup click', function(e) {
         events.push(e.type);
       });
     },
 
-    Checkbox: Checkbox.extend({
-      click() {
+    actions: {
+      checkboxClicked() {
```
mixonic: @nickiaconis this seemed to be asserting click behaviors. Per the description:

```js
'`click` triggers appropriate events in order'
```

However by using actions instead of a component with a `click` and `change` handler, I think the intended timing test was lost (now it tests action timing). Can this be tweaked to use a component with `click` and `change` where `{{view view.Checkbox}}` was before?

chadhietala:
Superseded by #13617

PR #13588, opened 2016-06-02
[Glimmer] Late bound layouts

chadhietala:
In Ember we typically leverage convention over configuration for figuring out what component class goes with which template. That being said, today it's totally valid to opt-out of those conventions by importing the template into the class and setting it as the layout for the component. This PR re-introduces the support for this functionality. Depends on https://github.com/tildeio/glimmer/pull/179


packages/ember-views/tests/views/view/template_test.js
```diff
@@ -1,162 +0,0 @@
-import { get } from 'ember-metal/property_get';
```
chadhietala: These don't makes any more sense with components.

packages/ember-views/tests/views/view/layout_test.js
```diff
@@ -1,118 +0,0 @@
...
-  });
-
-  expectAssertion(function() {
-    get(view, 'layout');
-  }, /cantBeFound/);
-});
-
-import { test } from 'ember-glimmer/tests/utils/skip-if-glimmer';
-
-test('should use the template of the associated layout', function() {
```
chadhietala: Doesn't seem legit use case anymore.

packages/ember-views/tests/views/view/layout_test.js
```diff
@@ -1,118 +0,0 @@
...
-
-  run(function() {
-    view.createElement();
-  });
-
-  equal(templateCalled, 0, 'template is not called when layout is present');
-  equal(layoutCalled, 1, 'layout is called when layout is present');
-});
-
-test('should use the associated template with itself as the context', function() {
```
chadhietala: [Migrarted](https://github.com/emberjs/ember.js/pull/13588/files#diff-c47395a5aaf92aa0a7edbfa812d4e0afR74)

packages/ember-views/tests/views/view/layout_test.js
```diff
@@ -1,118 +0,0 @@
...
-
-  run(function() {
-    view.createElement();
-  });
-
-  equal('template was called for Tom DAAAALE', view.$('#twas-called').text(),
-        'the named template was called with the view as the data source');
-});
-
-test('should fall back to defaultLayout if neither template nor templateName are provided', function() {
```
chadhietala: [Migrated](https://github.com/emberjs/ember.js/pull/13588/files#diff-c47395a5aaf92aa0a7edbfa812d4e0afR93)

packages/ember-views/tests/views/view/layout_test.js
```diff
@@ -1,118 +0,0 @@
...
-
-  run(function() {
-    view.createElement();
-  });
-
-  equal('used default layout', view.$().text(),
-        'the named template was called with the view as the data source');
-});
-
-test('should not use defaultLayout if layout is provided', function() {
```
chadhietala: [Migrated](https://github.com/emberjs/ember.js/pull/13588/files#diff-c47395a5aaf92aa0a7edbfa812d4e0afR111)

packages/ember-glimmer/lib/syntax/curly-component.js
```diff
@@ -105,6 +97,38 @@ class CurlyComponentManager {
...
+    // the runtime compiler phase :(
+    if (TemplateFactory) {
+      template = new TemplateFactory(env);
+    } else {
+      let layoutName = component.layoutName && get(component, 'layoutName');
+      if (layoutName) {
+        template = owner.lookup('template:' + layoutName);
+      }
+      if (!template) {
+        template = component.defaultLayout;
```
krisselden: is this a template factory too?

packages/ember-glimmer/lib/syntax/curly-component.js
```diff
@@ -105,6 +97,38 @@ class CurlyComponentManager {
...
+    let { component } = bucket;
+    let TemplateFactory = component.layout;
+    let template;
+    let owner = getOwner(component);
+
+    // If the user has supplied their own layout property we must
+    // create an instance of the wrapper class with the env to go back into
+    // the runtime compiler phase :(
+    if (TemplateFactory) {
+      template = new TemplateFactory(env);
```
krisselden: when we create the factory, if this component is rendered again, we won't find the definition and recreate and recompile. I was caching in lateBound() but I'm realizing we need a way to identify the template itself.

krisselden: We should move the id to the template() function we compile the modules with and give the id to the factory, so we can cache the factory instance on the late bound component definition.  We also should assume that the template wasn't looked up via the container.

packages/ember-glimmer/lib/syntax/curly-component.js
```diff
@@ -211,4 +231,18 @@ export class CurlyComponentDefinition extends ComponentDefinition {
     builder.attrs.dynamic('id', elementId);
     builder.attrs.static('class', 'ember-view');
   }
+
+  lateBound(template) {
+    let definition;
+    if (this._cache) {
+      definition = this._cache[template.id];
```
krisselden: given we create new instances from directly imported modules we likely should cache the instance we created for our env from the factory, otherwise the component will recreate and recompile everything if you directly import it onto layout.

packages/ember-glimmer-template-compiler/tests/precompile-test.js
```diff
@@ -0,0 +1,16 @@
...
+
+QUnit.test('returns a string', (assert) => {
+  let str = precompile('Hello');
+  assert.equal(typeof str, 'string');
+});
+
+QUnit.test('when wrapped in a template, precompile is the same as compile', (assert) => {
+  // Simulating what happens in a broccoli plugin
+  // when it is creating an AMD module
+  let precompiled = template(precompile('Hello'));
```
krisselden: in an amd module, `eval("require('template')("+precompile('Hello')+")")`

krisselden: we should assert that precompile returns a string, and that when it is evaluated, it returns what template() expects.

mixonic: I think something around here was biting me w/ ember-performance

https://github.com/eviltrout/ember-performance/blob/master/compile-templates/index.js#L96

which is then eval'd?

https://github.com/eviltrout/ember-performance/blob/eb3f94aaefd37272a8f2a62d7b66145b81cf39bd/test-client/test-client.js#L168

Glimmer doesn't seem to expect the same treatment.

Issue #13589, opened 2016-06-02
inline `if`s shouldn't eagerly eval their inverse arg


machty:
`{{if falsy (some-helper)}}` should not eval `some-helper`, but it does, which can be annoying when you want to check the presence of some property before passing it to a helper that'll barf if the prop is falsy.

https://ember-twiddle.com/b06224282690fe6e61c4d16f6fb05aa1?openFiles=controllers.application.js%2C

machty:
Updated the twiddle to also include an example of a CP that also shouldn't be eagerly read.

chancancode:
I believe the glimmer version does not have this problem, I can write a
test to verify

pixelhandler:
@chancancode should this issue be labelled "Glimmer 2" if the resolution is for that to land?

chancancode:
Hmm, this probably needs to have the opposite of that tag (htmlbars/glimmer
1).

There is no reason why this cannot be fixed on the current implementation -
if this is urgent and someone has the cycles and know-how to fix it, I
think we would accept that patch. On the other hand, if this is still open
by the time we land glimmer 2, we can just close it since that issue
doesn't exist there.

rwjblue:
> There is no reason why this cannot be fixed on the current implementation - if this is urgent and someone has the cycles and know-how to fix it, I think we would accept that patch.

Agreed.  Definitely welcome a patch, but that seems unlike at this point (though I'm happy to be wrong here).

> if this is still open by the time we land glimmer 2, we can just close it since that issue doesn't exist there.

I'm inclined to close now....

PR #13596, opened 2016-06-03
Add lazy evaluation tests

chancancode:
In both `{{#if truthy}}...{{else}}{{some-helper}}{{/if}}` and `{{if truthy ... (some-helper)}}`, `some-helper` should NOT be evaluated. This is especially important when you are trying to use the `if` helper to guard for `null`s etc.

They already works in Glimmer. The block form works in HTMLBars, but not the inline form.

See #13589


chancancode:
Interesting fact: in `{{if cond (x-truthy) (x-falsy)}}` when changing `cond` from `true` to `false` and then back to `true`, Glimmer 2 does NOT evaluate `(x-truthy)` a second time because its inputs (in this case, no inputs) has not changed since the last time it was evaluated. 😬

PR #13600, opened 2016-06-04
Revert QP changes from LTS 2.4.

rwjblue:
We discussed this in depth in a recent core team call, and decided that we should revert these changes from the LTS branch. They fix a number of issues that existed since 2.0.0, but they also introduced some new failures.  We will continue to iterate and address the new failures in 2.7+, but we should strive to ensure the latest LTS patch release is usable for folks.

Sadly, reverting these commits means that anyone using 2.4.5 and relying on the bugs that were originally being fixed will now have an issue on 2.4.6 (when released). I agree that this is annoying, but I think it is much better to keep the issues that are present on 2.4 branch consistent with the earlier point releases.


rwjblue:
**Note:** this targets the lts-2-4 branch.

PR #13602, opened 2016-06-04
Ensure compile options are functioning properly.

### LINKED ISSUE #13594: Registered template compilation plugins not used.
Source:

* https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer-template-compiler/lib/system/compile-options.js#L15-L16
* https://github.com/emberjs/ember.js/blob/master/packages/ember-htmlbars-template-compiler/lib/system/compile-options.js#L27-L28

In general, when calling `Ember.HTMLBars.compile` a `plugin` option is not required to be passed in `options`.  If user supplied plugins have been registered, but `compile` is not called with a `plugins: { ast: [] }` property in `options` the user supplied plugins are never used.

/cc @krisselden

rwjblue:
Fixes more issues with `Ember.Handlebars.compile` / `Ember.HTMLBars.compile`.
- The AST plugins were not being run (due to incorrect defaulting of the `plugins` property to an array instead of an object with an `ast` prop).
- Updated ember-glimmer test infrastructure to use `ember-glimmer-template-compiler` to compile.
- Ensure `ember-template-compiler/system/compile-options` uses the engine specific compile options.
- Fix assertions in compile options tests to ensure they stay in sync as more plugins are added.

Fixes #13594.

### DIFF
--- a/packages/ember-glimmer/tests/utils/abstract-test-case.js
+++ b/packages/ember-glimmer/tests/utils/abstract-test-case.js
@@ -14,7 +14,6 @@ import isEnabled from 'ember-metal/features';
 import { privatize as P } from 'container/registry';
 import DefaultComponentTemplate from 'ember-glimmer/templates/component';
 import EventDispatcher from 'ember-views/system/event_dispatcher';
-import { defaultCompileOptions } from 'ember-template-compiler';
 import { PartialDefinition } from 'glimmer-runtime';
 
 const packageTag = `@${packageName} `;
@@ -277,7 +276,7 @@ export class ApplicationTest extends TestCase {
   }
 
   compile(string, options) {
-    return compile(string, assign({}, defaultCompileOptions(), options));
+    return compile(...arguments);
   }
 
   registerRoute(name, route) {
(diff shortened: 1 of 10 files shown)


packages/ember-glimmer/tests/utils/abstract-test-case.js
```diff
@@ -277,7 +276,7 @@ export class ApplicationTest extends TestCase {
   }
 
   compile(string, options) {
-    return compile(string, assign({}, defaultCompileOptions(), options));
```
krisselden: yay!

packages/ember-htmlbars-template-compiler/lib/index.js
```diff
@@ -1,4 +1,5 @@
 export { default as compile } from './system/compile';
 export { default as precompile } from './system/precompile';
 export { default as template } from './system/template';
+export { default as defaultCompileOptions } from './system/compile-options';
```
krisselden: why do we export this? defaults should just be applied internally

krisselden:
Big improvement over what is there

packages/ember-htmlbars-template-compiler/lib/index.js
rwjblue: So that `ember-template-compiler/system/compile-options` can use the existing `../compiler` module to grab all that it needs.  Otherwise, we would have to encode the feature flag in another location...

PR #13604, opened 2016-06-04
Mention async cache in doc

mixonic:
Without flushing the async cache, you may end up building Ember with a cached set of feature flags.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,13 @@ Ember.js also provides access to the most advanced features of Javascript, HTML 
 6. Run `bower install` to ensure required web dependencies are installed.
 7. Run `npm run build` to build Ember.js. The builds will be placed in the `dist/` directory.
 
+If you edit the `feature.json` file, you may also need to clear the async disk
+cache to get a proper build:
+
+```js
+rm -rf tmp $TMPDIR/async-disk-cache/
+```
+
 ## npm install troubleshooting
 
 If you encounter a problem with downloading dependencies like:


mixonic:
Will consolidate this section of the `README.md` with the `CONTRIBUTING.md` file and update `README.md` to simply link to `CONTRIBUTING.md`.

krisselden:
Features need to be added to the cache key, along with anything else that makes plugins act differently of the babel broccoli plugin.  It is weird we would document this, do we have an issue tracking the babel broccoli plugin issue?

krisselden:
https://github.com/babel/broccoli-babel-transpiler/issues/67

rwjblue:
Not documenting as a work around just further trolls folks here. No one is happy with this state of affairs, but at least this gives folks a fighting chance at a properly functioning build after changing a features.json value.

krisselden:
seems like we could at least over https://github.com/babel/broccoli-babel-transpiler/blob/master/index.js#L103 and add features.

rwjblue:
Yep, we need to iterate plugins and allow them to augment the cache keys used.

stefanpenner:
lets just fix this bug please

rwjblue:
Feel free. For me, it hasn't been a large enough issue to spend much time on...

mixonic:
@rwjblue this bug is fixed and this can be closed?

rwjblue:
YES!

PR #13605, opened 2016-06-05
[BUGFIX beta] Ensure pauseTest runs after other async helpers

trentmwillis:
While looking into implementing `resumeTest`, I noticed that `pauseTest` didn't seem to be behaving properly (in that it wasn't properly waiting for prior async helpers). I'm unsure why this wasn't registered as an `asyncHelper` to begin with, but after looking at the repo in the state of the original commit, it looks like some internals of the helpers have changed which probably caused the regression.

### DIFF
--- a/packages/ember-testing/lib/helpers.js
+++ b/packages/ember-testing/lib/helpers.js
@@ -242,7 +242,7 @@ helper('currentURL', currentURL);
  @return {Object} A promise that will never resolve
  @public
 */
-helper('pauseTest', pauseTest);
+asyncHelper('pauseTest', pauseTest);
 
 /**
   Triggers the given DOM event on the element identified by the provided selector.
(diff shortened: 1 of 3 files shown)


trentmwillis:
Looks like Sauce had a problem on one of the boxes, seems like a fluke.

mixonic:
@trentmwillis I kicked sauce

mixonic:
@rwjblue @krisselden does this violate semver? If there an issue with changing this from a helper to an async helper?

rwjblue:
@mixonic - It seems like a clear bugfix to me. The behavior before this change was pretty trolling.  

Consider:

```js
test('asdf', function(assert) {
  visit('/');
  click('.blah');
  return pauseTest();
});
```

The `pauseTest` executes and returns before `visit` and `click` here...

mixonic:
👍 seems good

mixonic:
@trentmwillis can you flag this `[BUGFIX beta]` please? Lets roll it as a bugfix and see what happens in beta :-)

trentmwillis:
@mixonic sounds good; I've update the commit flag.

mixonic:
Thanks @trentmwillis!

PR #13608, opened 2016-06-05
PERF: Guard QP generation

chadhietala:
`generateURL` in general needs a lot of love and currently accounts for a decent amount of slowness for each different type of URL. This is more of a common sense fix that makes sure we do not normalize query params when we do not have any.


rwjblue:
Seems fine to me.  Does this result in any change performance wise?

chadhietala:
@rwjblue Basically prevents you from calling into something like this.
screen shot 2016-06-05 at 3 26 17 pm

This is calling into `{{link-to}}` from Glimmer.

chadhietala:
Ugh this is much more intwined than I thought.

Issue #13607, opened 2016-06-05
Ember router should only handle URLs that're only a nested under rootURL


batjaa:
We're embedding our ember app in a legacy application, but when you transition from an ember route to a non-ember URL, Ember route throws `UnrecognizedUrlError`. Looks like when override the `location.onUpdateURL` event we should check if it's nested under rootURL.

https://github.com/emberjs/ember.js/blob/v2.5.0/packages/ember-routing/lib/system/router.js#L176

rwjblue:
I'm not really following how this is triggered here.  Is the legacy application another single page app that you are transitioning between?  If not, then the page refresh should void any `onUpdateURL` that we setup anyways...

batjaa:
Yes, the legacy application is SPA. When it pushes a new state to history I think `popstate` triggers the `onUpdateURL` callback.

batjaa:
Bump

acorncom:
@batjaa Could you make up a reproduction of your issue on Twiddle (or possibly a Github repo if it's hard to trigger on Twiddle)?  This is going to be hard for people to help with without it ...

batjaa:
https://github.com/batjaa/ember-root-url-test

Created this repo for reproduction. It's a little bit vague but I hope you get the idea.

The main app (index.html) is a SPA that talks to the browser history API. Assuming some action has triggered the `popstate` event while transition to a URL from the parent app when you have the ember app embedded and it's running. You can it throws an exception that the URL is not defined.

les2:
@batjaa Can you serve the ember app from a different path? E.g., serve the legacy SPA off of `/#/legacy/routes/here` and the Ember SPA off `/foo/*` - I know this works because I am using this setup to migrate from a legacy application.

But it just seems like asking for trouble if you have both the legacy SPA and Ember managing history api events at the same time.

pixelhandler:
@batjaa @acorncom @rwjblue is this still an issue, perhaps we should close or create a new reproduction of this, what do you think?

pixelhandler:
Per our triage policy I'll close this out for now, feel free to reopen if the issue persists on the current release.

- https://github.com/emberjs/ember.js/blob/master/CONTRIBUTING.md#triaging-policy

PR #13609, opened 2016-06-06
Call handleURL() only if url is nested under rootURL

batjaa:
https://github.com/emberjs/ember.js/issues/13607

### DIFF
--- a/packages/ember-routing/lib/location/history_location.js
+++ b/packages/ember-routing/lib/location/history_location.js
@@ -190,11 +190,13 @@ export default EmberObject.extend({
 
     this._popstateHandler = () => {
       var url = this.getURL();
+      var shouldHandleURL = this._shouldHandleURL(url);
       // Ignore initial page load popstate event in Chrome
       if (!popstateFired) {
         popstateFired = true;
-        if (!this._shouldHandleURL(url)) { return; }
+        if (!shouldHandleURL) { return; }
       }
+      if (!shouldHandleURL) { return; }
       callback(url);
     };
 
--- a/packages/ember-routing/lib/location/history_location.js
+++ b/packages/ember-routing/lib/location/history_location.js
@@ -189,15 +189,21 @@ export default EmberObject.extend({
     this._removeEventListener();
 
     this._popstateHandler = () => {
+      var url = this.getURL();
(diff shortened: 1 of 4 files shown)


packages/ember-routing/lib/system/router.js
```diff
@@ -200,7 +200,9 @@ var EmberRouter = EmberObject.extend(Evented, {
     this._setupRouter(router, location);
 
     location.onUpdateURL((url) => {
-      this.handleURL(url);
+      if (this._shouldHandleURL(url)) {
```
rwjblue: To me this feels like a location responsibility.  Can you move this into the locations and only call the `onUpdateURL` callback when we are handling the rootURL?

batjaa:
I see that the `hash` location works differently.

batjaa:
Note sure  what's breaking. Reverting my changes back and I still get the same errors.

Serabe:
Hello, @batjaa,

do you have time to keep working on this?

Thank you for your work!

batjaa:
Sure, will update my PR and fix tests.

batjaa:
Anything here I can do to improve?

packages/ember-routing/lib/location/history_location.js
```diff
@@ -189,17 +189,25 @@ export default EmberObject.extend({
...
+      var shouldHandleURL = this._shouldHandleURL(url);
       // Ignore initial page load popstate event in Chrome
       if (!popstateFired) {
         popstateFired = true;
-        if (this.getURL() === this._previousURL) { return; }
+        if (!shouldHandleURL) { return; }
       }
-      callback(this.getURL());
+      if (!shouldHandleURL) { return; }
+      callback(url);
```
pixelhandler: I'm curious if there can just be one early return instead of two.

```js
    this._popstateHandler = () => {
      var url = this.getURL();
      if (!this._shouldHandleURL(url)) {
        return;
      }
      // Ignore initial page load popstate event in Chrome
      if (!popstateFired) {
        popstateFired = true;
      }
      callback(url);
    };
```

auvipy:
is this still needed or can be closed?

MelSumner:
Since the related issue was closed, I'm going to close this PR as well, primarily due to age.

PR #13613, opened 2016-06-06
Add getOwner/setOwner shims

cibernox:
Added shims for `getOwner` / `setOwner` as done in https://github.com/ember-cli/ember-cli-shims/pull/65


stefanpenner:
👍 @rwjblue r?

stefanpenner:
actually this was the result of @rwjblue's suggestion so this should be good

stefanpenner:
ah, a wild @rwjblue has merged. 👍

PR #13614, opened 2016-06-07
[CLEANUP] Fix direct access of meta

krisselden:
Accessing ember meta directly makes refactoring one of the most perf sensitive areas very difficult.


rwjblue:
👍

PR #13615, opened 2016-06-07
[Glimmer2] Implement input

asakusuma:
I don't think the late bound stuff can help with implementing `input` since [`lateBound`](https://github.com/emberjs/ember.js/blob/1a61ed01d8e1bf1113b69e0698a7647d2487aec7/packages/ember-glimmer/lib/syntax/curly-component.js#L238-L250) gets called after [`CurlyComponentManager.create`](https://github.com/emberjs/ember.js/blob/1a61ed01d8e1bf1113b69e0698a7647d2487aec7/packages/ember-glimmer/lib/syntax/curly-component.js#L46).

cc @krisselden @chadhietala @GavinJoyce

### DIFF
--- a/packages/ember-glimmer-template-compiler/lib/plugins/transform-input-type-syntax.js
+++ b/packages/ember-glimmer-template-compiler/lib/plugins/transform-input-type-syntax.js
@@ -1,0 +1,67 @@
+/**
+ @module ember
+ @submodule ember-glimmer
+*/
+
+/**
+  A Glimmer2 AST transformation that replaces all instances of
+
+  ```handlebars
+ {{input type=boundType}}
+  ```
+
+  with
+
+  ```handlebars
+ {{input (-input-type boundType) type=boundType}}
+  ```
+
+  Note that the type parameters is not removed as the -input-type helpers
+  is only used to select the component class. The component still needs
+  the type parameter to function.
+
+  @private
+  @class TransformInputTypeSyntax
+*/
+
+export default function TransformInputTypeSyntax() {
+  // set later within Glimmer2 to the syntax package
+  this.syntax = null;
+}
+
+/**
+  @private
+  @method transform
+  @param {AST} ast The AST to be transformed.
+*/
+TransformInputTypeSyntax.prototype.transform = function TransformInputTypeSyntax_transform(ast) {
+  let { traverse, builders: b } = this.syntax;
+
+  traverse(ast, {
+    MustacheStatement(node) {
+      if (isInput(node)) {
+        insertTypeHelperParameter(node, b);
+      }
+    }
+  });
+
+  return ast;
+};
+
+function isInput(node) {
+  return node.path.original === 'input';
+}
+
+function insertTypeHelperParameter(node, builders) {
+  let pairs = node.hash.pairs;
+  let pair = null;
+  for (let i = 0; i < pairs.length; i++) {
+    if (pairs[i].key === 'type') {
+      pair = pairs[i];
+      break;
+    }
+  }
+  if (pair && pair.value.type !== 'StringLiteral') {
+    node.params.unshift(builders.sexpr('-input-type', [builders.path(pair.value.original, pair.loc)], null, pair.loc));
+  }
+}
--- a/packages/ember-glimmer/lib/environment.js
+++ b/packages/ember-glimmer/lib/environment.js
@@ -31,6 +31,7 @@ import { default as mut } from './helpers/mut';
 import { default as readonly } from './helpers/readonly';
 import { default as unbound } from './helpers/unbound';
 import { default as classHelper } from './helpers/-class';
+import { default as inputTypeHelper } from './helpers/-input-type';
 import { default as queryParams } from './helpers/query-param';
 import { default as eachIn } from './helpers/each-in';
 import { default as normalizeClassHelper } from './helpers/-normalize-class';
@@ -38,6 +39,42 @@ import { OWNER } from 'container/owner';
 
 const builtInComponents = {
   textarea: '-text-area'
+};
+
+function createCurly(args, templates, definition) {
+  wrapClassAttribute(args);
+  return new CurlyComponentSyntax({ args, definition, templates });
+}
+
+function buildTextFieldSyntax({ args, templates }, getDefinition) {
+  let definition = getDefinition('-text-field');
+  wrapClassAttribute(args);
+  return new CurlyComponentSyntax({ args, definition, templates });
+}
+
+const builtInDynamicComponents = {
+  input({ key, args, templates }, getDefinition) {
+    if (args.named.has('type')) {
+      let typeArg = args.named.at('type');
+      if (typeArg.type === 'value') {
+        if (typeArg.value === 'checkbox') {
+          assert(
+            '{{input type=\'checkbox\'}} does not support setting `value=someBooleanValue`; ' +
+            'you must use `checked=someBooleanValue` instead.',
+            !args.named.has('value')
+          );
+
+          return createCurly(args, templates, getDefinition('-checkbox'));
+        } else {
+          return buildTextFieldSyntax({ args, templates }, getDefinition);
+        }
+      }
+    } else {
+      return buildTextFieldSyntax({ args, templates }, getDefinition);
+    }
+
+    return new DynamicComponentSyntax({ args, templates });
+  }
 };
 
 const builtInHelpers = {
@@ -55,6 +92,7 @@ const builtInHelpers = {
   unless: inlineUnless,
   '-class': classHelper,
   '-each-in': eachIn,
+  '-input-type': inputTypeHelper,
   '-normalize-class': normalizeClassHelper
 };
 
@@ -143,6 +181,11 @@ export default class Environment extends GlimmerEnvironment {
 
     if (isSimple && (isInline || isBlock)) {
       // 2. built-in syntax
+
+      let generateBuiltInSyntax = builtInDynamicComponents[key];
+      // Check if it's a keyword
+      let mappedKey = builtInComponents[key];
+
       if (key === 'component') {
         return new DynamicComponentSyntax({ args, templates });
       } else if (key === 'outlet') {
@@ -156,11 +199,14 @@ export default class Environment extends GlimmerEnvironment {
         definition = this.getComponentDefinition([internalKey]);
       } else if (key.indexOf('-') >= 0) {
         definition = this.getComponentDefinition(path);
+      } else if (mappedKey) {
+        definition = this.getComponentDefinition([mappedKey]);
       }
 
       if (definition) {
-        wrapClassAttribute(args);
-        return new CurlyComponentSyntax({ args, definition, templates });
+        return createCurly(args, templates, definition);
+      } else if (generateBuiltInSyntax) {
+        return generateBuiltInSyntax(statement, (path) => this.getComponentDefinition([path]));
       }
 
       assert(`Could not find component named "${key}" (no component or template with that name was found)`, !isBlock || this.hasHelper(key));
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "express": "^4.5.0",
     "finalhandler": "^0.4.0",
     "github": "^0.2.3",
-    "glimmer-engine": "tildeio/glimmer#bf8213",
+    "glimmer-engine": "tildeio/glimmer#d4d1e3a",
     "glob": "^5.0.13",
     "htmlbars": "0.14.23",
     "mocha": "^2.4.5",
--- a/packages/ember-glimmer-template-compiler/lib/system/compile-options.js
+++ b/packages/ember-glimmer-template-compiler/lib/system/compile-options.js
@@ -1,5 +1,6 @@
 import defaultPlugins from 'ember-template-compiler/plugins';
 import TransformActionSyntax from '../plugins/transform-action-syntax';
+import TransformInputTypeSyntax from '../plugins/transform-input-type-syntax';
 import TransformAttrsIntoProps from '../plugins/transform-attrs-into-props';
 import TransformEachInIntoEach from '../plugins/transform-each-in-into-each';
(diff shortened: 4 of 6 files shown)


packages/ember-glimmer/lib/syntax/curly-component.js
```diff
@@ -43,7 +43,7 @@ class CurlyComponentManager {
   create(definition, args, dynamicScope, hasBlock) {
     let parentView = dynamicScope.view;
 
-    let klass = definition.ComponentClass;
+    let klass = definition.getComponentClass(args);
```
chadhietala: Why?

packages/ember-glimmer/lib/syntax/curly-component.js
```diff
@@ -235,6 +235,10 @@ export class CurlyComponentDefinition extends ComponentDefinition {
     builder.attrs.static('class', 'ember-view');
   }
 
+  getComponentClass() {
```
chadhietala: This seems unnecessary.

packages/ember-glimmer/lib/syntax/curly-component.js
asakusuma: Because for input, class selection is dynamic: https://github.com/emberjs/ember.js/pull/13615/files#diff-f9ead5f5a3d079b17b0d5b74759b00b0R9

packages/ember-glimmer/lib/syntax/curly-component.js
asakusuma: See comment above

packages/ember-glimmer/lib/environment.js
```diff
@@ -146,6 +157,11 @@ export default class Environment extends GlimmerEnvironment {
           assert(`A helper named '${path[0]}' could not be found`, false);
         }
       } else {
+        let generateSyntax = builtInDynamicComponents[key];
```
asakusuma: Sure. It would just mean always doing the lookup up-front which seems fine.

asakusuma:
For whatever reason, one of the input tests is failing in Safari and Microsoft browsers. I'm looking into it.

asakusuma:
I think saucelabs is having problems

```
Error: 07 Jun 05:09:02 - All jobs using tunnel have finished.
07 Jun 05:09:02 - Sauce Connect could not establish a connection.
07 Jun 05:09:02 - Please check your firewall and proxy settings.
```

packages/ember-glimmer/lib/syntax/curly-component.js
krisselden: Lets just rewrite all inputs that don't have a static type to be dynamic

packages/ember-glimmer/lib/syntax/curly-component.js
krisselden: Definitions cache compiled layout, they can't change class, this is why late bound layouts clone the definition and cache based on template id

krisselden: We don't have enough tests to cover mixing up stuff for the caching we are doing because these issues are not being caught

krisselden: The idea is that static lookup will be the norm or use the dynamic helper, these inbetween dynamic stuff has to be done carefully and is not intended to be long term.

packages/ember-glimmer/lib/syntax/curly-component.js
asakusuma: @krisselden are you talking about using `DynamicComponentSyntax`? If so, I'll need to modify it to support using named params, not just assuming first positional.

packages/ember-glimmer/lib/syntax/curly-component.js
asakusuma: 1. I don't understand the relationship between caching the layout (I'm not explicitly setting the layout here, just the class) and changing the class
2. Does "static lookup" just mean using a plain `CurlyComponentSyntax`?
3. What is the "dynamic helper"?
4. What's the difference between `DynamicComponentSyntax` and late bound `CurlyComponentSyntax`?

asakusuma:
I've changed the approach to re-write the AST. Any ideas why travis doesn't seem to be running?

rwjblue:
wat

packages/ember-glimmer-template-compiler/lib/plugins/transform-input-type-syntax.js
```diff
@@ -0,0 +1,67 @@
...
+  let pairs = node.hash.pairs;
+  let pair = null;
+  for (let i = 0; i < pairs.length; i++) {
+    if (pairs[i].key === 'type') {
+      pair = pairs[i];
+      break;
+    }
+  }
+  if (pair && pair.value.type !== 'StringLiteral') {
+    node.params.unshift(builders.sexpr('-input-type', [builders.path(pair.value.original)], null));
```
rwjblue: Please add `loc` information to these (`builders.sexpr` and `builders.path`).

rwjblue: What happens if `-input-type` is already specified?

packages/ember-glimmer/lib/environment.js
```diff
@@ -31,13 +31,49 @@ import { default as log } from './helpers/log';
...
 
 const builtInComponents = {
   textarea: '-text-area'
 };
 
+function buildTextFieldSyntax({ args, templates }, getDefinition) {
+  let definition = getDefinition('-text-field');
+  wrapClassBindingAttribute(args);
+  wrapClassAttribute(args);
+  return new CurlyComponentSyntax({ args, definition, templates });
```
rwjblue: This combo is in this file at least three times:

```js
wrapClassBindingAttribute(args);
wrapClassAttribute(args);
return new CurlyComponentSyntax({ args, definition, templates });
```

Is this something we could extract and share?

packages/ember-glimmer-template-compiler/lib/plugins/transform-input-type-syntax.js
asakusuma: @rwjblue how do I add `loc` information here? Neither [`buildSexpr` nor `buildPath`](https://github.com/tildeio/glimmer/blob/master/packages/glimmer-syntax/lib/builders.ts#L93-L111) accept a `loc` arguments like some of the other builders. Does that need to be added?

I don't think end-users can add a `-input-type` helper since it's registered as an internal helper. But if one already existed, I believe `DynamicComponentSyntax` [only looks at the first argument](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/syntax/dynamic-component.js#L7), so if `-input-type` was already specified, it would be ignored.

packages/ember-glimmer/lib/environment.js
asakusuma: Yep, will improve here

asakusuma:
Working on figuring out why tests fail in Safari and MS browsers. Safari is failing for a different reason than MS.

asakusuma:
Modifying input.selectionStart, which is utilized in the cursor tests, causes an event in Safari (does not happen in Chrome). This event kicks off a chain of events that results in an error because the [view does not have `sendAction` implemented](https://github.com/emberjs/ember.js/blob/8d948623e5fba84551d856c80be701bc4aee04cf/packages/ember-views/lib/mixins/text_support.js#L348). Not sure if this is the right approach, but I solved this by turning off the event dispatcher for the input tests, since we are not interested in the event dispatcher for these tests.

packages/ember-glimmer-template-compiler/lib/plugins/transform-input-type-syntax.js
rwjblue: > Neither buildSexpr nor buildPath accept a loc arguments like some of the other builders. Does that need to be added?

Yes.  Each AST node should be able to be built with a loc.

> I don't think end-users can add a -input-type helper since it's registered as an internal helper. 

There is nothing stopping a user from having already defined a `-input-type` helper in their application, but I'm not sure we should care about that too much...

> I believe DynamicComponentSyntax only looks at the first argument, so if -input-type was already specified, it would be ignored.

If we expect this to go into exactly `node.params[0]` we should do that, and error if there was already something there.

For example today this "works" (doesn't blow up):

```js
{{input "blah blah blah" value=someThing}}
```

If we are going to add (internally) significance to the first ordered argument to `{{input`, then we should have a helpful assertion if the user puts something there...

asakusuma: @rwjblue I meant that the end-user cannot use `-input-helper` unless they themselves define `-input-helper`. I confirmed this locally. My understanding is that a leading dash is reserved for internal helpers.

I can certainly add a check to blow up if there's any pre-existing positional params. In other words, we are now saying that you cannot use positional parameters with `{{input}}`.

chadhietala:
@asakusuma Yea we need to land #13532.

krisselden:
@asakusuma Should you skip the failing test for now and we can recheck after sendAction?

asakusuma:
I'm fine doing that, though I'm not sure `sendAction` is what's causing the Microsoft browser issues.

All the tests pass in Chrome, Safari, and Firefox. In Microsoft browsers, the [null values test](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/tests/integration/helpers/input-test.js#L426-L435) fails. What's failing is that when you set the value of the input to `null`, the value does not become an empty string in MS browsers for whatever reason.

asakusuma:
Discovered the source of the problem in MS Edge. `removeAttribute('value');` doesn't seem to work in Edge. When the `value` needs to be `null`, glimmer uses `removeAttribute('value');` to update the input element.

Here's a simple repro: https://gist.github.com/asakusuma/aad85f430f465903e56feee2b53a414c

cc @jdalton

rwjblue:
@asakusuma - Have you reported an issue against MS Edge?

jdalton:
@asakusuma What does MS Edge return?

asakusuma:
@jdalton in MS Edge, `value` is still "I have the best inputs" even after `removeAttribute` is run. On chrome, value attribute is removed and the input is empty.

I have not yet filed a ticket.

jdalton:
@asakusuma I filed a ticket internally.

rwjblue:
@jdalton - Thank you!

jdalton:
I've gotten `removeAttribute` to work for other attributes on other elements. Do you all know of any other problem attributes?

asakusuma:
@jdalton thanks! I'm not aware of any other problem attributes.

jdalton:
While making a feature detection for this I noticed that in Chrome:
```js
var input = document.createElement('input');
var buggyRemoveAttributeValue = !!(input.value=1,input.removeAttribute('value'),input.value);
console.log(buggyRemoveAttributeValue);
// => true
```
but
```js
var div = document.createElement('div');
div.innerHTML = '<input value=1>';
var input = div.firstChild;

var buggyRemoveAttrValue = !!(input.value=1,input.removeAttribute('value'),input.getAttribute('value'));
console.log(buggyRemoveAttrValue)
// => false (but true in Edge)
```

jdalton:
It's the old attribute vs property thing.

asakusuma:
Did some bisecting, and it would appear that https://github.com/tildeio/glimmer/commit/951cbe5693400ac5cea703f1226b709ffcb28a2d introduced an incompatibility with the ember build.

```
Could not require 'ember-cli-build.js': Cannot find module 'glimmer-engine/ember-cli-build'
Error: Could not require 'ember-cli-build.js': Cannot find module 'glimmer-engine/ember-cli-build'
    at Function.Module._resolveFilename (module.js:339:15)
    at Function.Module._load (module.js:290:25)
    at Module.require (module.js:367:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/home/travis/build/emberjs/ember.js/ember-cli-build.js:45:21)
    at Module._compile (module.js:413:34)
    at Object.Module._extensions..js (module.js:422:10)
    at Module.load (module.js:357:32)
    at Function.Module._load (module.js:314:12)
    at Module.require (module.js:367:17)
    at require (internal/module.js:20:19)
    at module.exports (/home/travis/build/emberjs/ember.js/node_modules/ember-cli/lib/utilities/find-build-file.js:19:17)
    at CoreObject.module.exports.Task.extend.setupBroccoliBuilder (/home/travis/build/emberjs/ember.js/node_modules/ember-cli/lib/models/builder.js:48:21)
    at CoreObject.module.exports.Task.extend.init (/home/travis/build/emberjs/ember.js/node_modules/ember-cli/lib/models/builder.js:89:10)
    at CoreObject.superWrapper [as init] (/home/travis/build/emberjs/ember.js/node_modules/core-object/lib/assign-properties.js:32:18)
    at CoreObject.Class (/home/travis/build/emberjs/ember.js/node_modules/core-object/core-object.js:32:33)
```
cc @tomdale @rwjblue

tomdale:
@asakusuma I think this is related to https://github.com/tildeio/glimmer/pull/190, we need to ship `packages/` and `ember-cli-build.js` with the npm package for use by Ember.

asakusuma:
After including the glimmer fix, Travis is now failing with a new problem

```
The Broccoli Plugin: [Funnel] failed with:
Error: ENOENT: no such file or directory, lstat '/home/travis/build/emberjs/ember.js/node_modules/glimmer-engine/node_modules/simple-html-tokenizer/lib'
````

**Comment by asakusuma** [comment]
@rwjblue do we need to increase the timeout?

Blueprints is also failing but I'm not seeing a useful error.

```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

asakusuma:
npm is still borking on the `blueprints` job. it keeps doing this which leads me to believe there is something wrong, but there's no useful error, just `npm ERR! code 1`

rwjblue:
Cleared cache and restarted builds.

rwjblue:
Thank you @asakusuma for sticking with this so long!

PR #13616, opened 2016-06-07
Add 'ember-testing' export path, containing `isTesting` and `setTesting` named exports

cibernox:
Following https://github.com/ember-cli/ember-cli-shims/pull/66

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,42 @@
 /* jshint node: true */
 'use strict';
+var stew = require('broccoli-stew');
+
+var paths = {};
+var absolutePaths = {};
+
+function add(paths, name, path) {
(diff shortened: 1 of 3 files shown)


krisselden:
This particular flag does not actually have to do with the package ember-testing.   It disables autoboot of Ember.Application on dom ready and errors on autorun.  It has nothing to do with the testing adapter stuff  /cc @rwjblue

rwjblue:
Yes, I/we understand that it is not related to the internal `ember-testing` package.  The existing modules for things inside of the internal `ember-testing` package are from `ember-test` (not `ember-testing`).

Where would you suggest exporting these helpers from instead?

PR #13617, opened 2016-06-07
[GLIMMER] Cleanup Glimmer View Support Mixin

chadhietala:
In glimmer we have a better strategy around setting layout and thus
private APIS such as `templateNameFor` are no longer needed. This also
removes the following from the view support mixin.
- CP for `template` (does nothing today)
- CP for `layout` (cannot change post init, slotted in component class)
- Slotted `layoutName` (in Component class)
- Slotted `templateName` (doesn't work in components and was deprecated)
- Slotted `isView` (not used in glimmer)

This also consolidates the component lookup between both engines and removes the `{{view}}` keyword from HTMLBars.

/cc @krisselden


rwjblue:
Why is `packages/ember-views/lib/mixins/view_support.js` completely duplicated into `packages/ember-glimmer/lib/ember-views/view-support.js`?

chadhietala:
It's not a duplicate it removes [all of this stuff](https://github.com/emberjs/ember.js/blob/master/packages/ember-views/lib/mixins/view_support.js#L21-L128) or the properties are already present in the component class in Glimmer. `ember-htmlbars` internals still make assumptions about this stuff being there.

rwjblue:
Views are dead, what breaks when we just delete that code in the main file?

chadhietala:
HTMLBars `{{outlet}}`, `{{render}}` and `{{view}}`. `{{view}}` does not count. I believe here is where the outlet stuff breaks. https://github.com/emberjs/ember.js/blob/master/packages/ember-htmlbars/lib/renderer.js#L34-L41

rwjblue:
I guess my thought is that we should go the other way. Use a more specific one in htmlbars. In the long run, we will just delete ember-htmlbars but moving the special version into ember-glimmer will require more work (to move it back out).

chadhietala:
@rwjblue Will do. Going to just create a specific mixin for the removals.

rwjblue:
This looks much better to me, but I'm curious how ember-glimmer gets a different version without the mixin?

chadhietala:
Ah good question, so the HTMLBars component [subclasses `View`](https://github.com/emberjs/ember.js/blob/master/packages/ember-htmlbars/lib/component.js#L125) which will come with the [`ViewSupport` mixin and the new `TemplateSupport` mixin](https://github.com/emberjs/ember.js/pull/13617/files#diff-a9aeff1ccd7f8ef823c5b1c347b9deb8R657). In Glimmer we extend [`CoreView` and re-build up the mixins](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/component.js#L22-L28). So we are removing the functionality because we don't mix in `TemplateSupport` to the Glimmer component.

rwjblue:
We do not need to mix `TemplateSupport` into the HTMLBars version of `Component` either, these concepts are only a thing for non-component views (the examples you linked are related to non-component views).

For example, if you review [ember-htmlbars/lib/component.js](https://github.com/emberjs/ember.js/blob/49730043e687a1d789d90246e375b0712199a8fe/packages/ember-htmlbars/lib/component.js#L187-L189) we explicitly set `template`, `templateName`, `layout` to `null` (clobbering the computed properties being extracted here).

chadhietala:
@rwjblue Ok I think have sorted through this. There are places in HTMLBars (specifically the outlet) where you need the `TemplateSupport` mixin. Everywhere else I removed the need for it. This included refactoring tests to not to use views. I also nuked the `{{view}}` keyword as I would have had to update all of the tests for something that we know is going away.

packages/ember-htmlbars/lib/component.js
```diff
@@ -138,6 +139,21 @@ const Component = View.extend(TargetActionSupport, {
     }
   }),
 
+  templateForName(name, type) {
```
chadhietala: This is still used for component registration and is exercised in this [test](https://github.com/emberjs/ember.js/blob/master/packages/ember/tests/component_registration_test.js#L154-L172).

packages/ember-htmlbars/lib/keywords/view.js
```diff
@@ -1,289 +0,0 @@
-/**
-@module ember
-@submodule ember-templates
-*/
-
-import { readViewFactory } from '../streams/utils';
```
chadhietala: No longer supported

packages/ember-htmlbars/lib/utils/lookup-component.js
```diff
@@ -1,22 +0,0 @@
-function lookupComponentPair(componentLookup, owner, name, options) {
```
chadhietala: There is now a shared version of this in `ember-templates`

packages/ember-htmlbars/tests/helpers/render_test.js
```diff
@@ -594,31 +594,6 @@ QUnit.test('throws an assertion if {{render}} is called with a literal for a mod
   }, 'The second argument of {{render}} must be a path, e.g. {{render "post" post}}.');
 });
 
-QUnit.test('{{render}} helper should let view provide its own template', function() {
```
chadhietala: Seems to be testing view specifics.

packages/ember-htmlbars/tests/helpers/view_test.js
```diff
@@ -1,1243 +0,0 @@
-/*globals EmberDev */
```
chadhietala: No longer supported.

packages/ember-testing/tests/acceptance_test.js
```diff
@@ -364,7 +360,7 @@ QUnit.test('test must not finish while asyncHelpers are pending', function () {
   }
 });
 
-QUnit.test('visiting a URL that causes another transition should yield the correct URL', function () {
+test('visiting a URL that causes another transition should yield the correct URL', function () {
```
chadhietala: These need to be skipped now because `{{input}}` is not finished on the glimmer side.

packages/ember-testing/tests/integration_test.js
```diff
@@ -67,14 +62,17 @@ QUnit.module('ember-testing Integration', {
...
     run(App, App.destroy);
     App = null;
     Test.adapter = originalAdapter;
   }
 });
 
-QUnit.test('template is bound to empty array of people', function() {
+import { test } from 'ember-glimmer/tests/utils/skip-if-glimmer';
+
+test('template is bound to empty array of people', function() {
```
chadhietala: These need to be skipped because the test setup appears to be wrong.

packages/ember/tests/routing/basic_test.js
```diff
@@ -2258,82 +2247,6 @@ test('The template is not re-rendered when the route\'s context changes', functi
   equal(insertionCount, 1, 'view should still have inserted only once');
 });
 
-
-test('The template is not re-rendered when two routes present the exact same template, view, & controller', function() {
```
chadhietala: This is specifically testing view insertion.

packages/ember/tests/routing/basic_test.js
```diff
@@ -3827,19 +3700,6 @@ test('Can disconnect from nested render helpers', function() {
   equal(jQuery('#qunit-fixture .bar').text(), '');
 });
 
-test('Can render with layout', function() {
```
chadhietala: This is testing `layoutName` on a view. This is now tested by [these tests](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/tests/integration/components/curly-components-test.js#L94-L128)

packages/ember-testing/tests/helpers_test.js
```diff
@@ -414,10 +398,10 @@ test('`click` triggers appropriate events in order', function() {
     events = [];
     return click('.index-view input[type=checkbox]');
   }).then(function() {
-    // i.e. mousedown, mouseup, change:true, click, click:true
+    // i.e. mousedown, mouseup, click
     // Firefox differs so we can't assert the exact ordering here.
     // See https://bugzilla.mozilla.org/show_bug.cgi?id=843554.
-    equal(events.length, 5, 'fires click and change on checkboxes');
+    equal(events.length, 3, 'fires click and change on checkboxes');
```
chadhietala: Not sure if `change:true` `click:true` were view events or not.

packages/ember-views/tests/views/view/child_views_test.js
```diff
@@ -1,172 +0,0 @@
...
-import { registerKeyword, resetKeyword } from 'ember-htmlbars/tests/utils';
-import viewKeyword from 'ember-htmlbars/keywords/view';
-import { setOwner } from 'container/owner';
-
-var originalViewKeyword;
-var parentView, childView;
-
-import { test, testModule } from 'ember-glimmer/tests/utils/skip-if-glimmer';
-
-testModule('tests/views/view/child_views_tests.js', {
```
chadhietala: These are now covered by [these child component tests](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/tests/integration/components/curly-components-test.js#L672).

packages/ember-views/tests/views/view_test.js
```diff
@@ -86,53 +79,3 @@ test('should re-render if the context is changed', function() {
 
   equal(jQuery('#qunit-fixture #template-context-test').text(), 'bang baz', 're-renders the view with the updated context');
 });
-
-test('propagates dependent-key invalidated sets upstream', function() {
```
chadhietala: This now tested [here](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/tests/integration/components/curly-components-test.js#L2184-L2318)

packages/ember-views/tests/views/view_test.js
```diff
@@ -86,53 +79,3 @@ test('should re-render if the context is changed', function() {
...
-  var childView = view.get('childView');
-
-  run(function() {
-    childView.set('childProp', 'new-value');
-  });
-
-  equal(view.get('parentProp'), 'new-value', 'new value is propagated across template');
-});
-
-test('propagates dependent-key invalidated bindings upstream', function() {
```
chadhietala: Tested [here](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/tests/integration/components/curly-components-test.js#L2184-L2318)

rwjblue:
> ember-views/tests/views/view/view_lifecycle_test.js: line 3, col 8, 'EmberObject' is defined but never used.

packages/ember-testing/tests/helpers_test.js
rwjblue: They were just above in the setup: https://github.com/emberjs/ember.js/pull/13617/files#diff-f834632dddeade3002a271a3e9fb68a3L369. Can you tweak this one to register a custom `{{x-checkbox` helper so that we can still confirm these events?

packages/ember/tests/routing/basic_test.js
rwjblue: Not really.  It is confirming that `setOutletState` is stable if the controller/template/view are the same.  This test should be salvaged, and just remove usage of views from it (specify `templateName` in the routes instead).

rwjblue:
This looks great, only two things to fix (other than the JSHint issue).

packages/ember-testing/tests/integration_test.js
```diff
@@ -67,14 +62,17 @@ QUnit.module('ember-testing Integration', {
...
+    setTemplates({});
     jQuery('#ember-testing-container, #ember-testing').remove();
     run(App, App.destroy);
     App = null;
     Test.adapter = originalAdapter;
   }
 });
 
-QUnit.test('template is bound to empty array of people', function() {
+import { test } from 'ember-glimmer/tests/utils/skip-if-glimmer';
```
rwjblue: The location of this changed (to fix beta builds).

packages/ember-htmlbars/lib/component.js
krisselden: outlet in htmlbars can use owner

chadhietala: Can we address this at a later time?

packages/ember/tests/routing/basic_test.js
chadhietala: Brought this back

packages/ember-htmlbars/lib/component.js
rwjblue: It isn't just RE: outlet anyways. This method is how we support layoutName (in component's init). I'm happy to refactor in the future, but IMO this is definitely better already...

rwjblue:
I'm :+1: on this. 

@krisselden - You see any blockers?

packages/ember-glimmer/lib/environment.js
```diff
@@ -8,7 +8,7 @@ import { assert } from 'ember-metal/debug';
 import { CurlyComponentSyntax, CurlyComponentDefinition } from './syntax/curly-component';
 import { DynamicComponentSyntax } from './syntax/dynamic-component';
 import { OutletSyntax } from './syntax/outlet';
-import lookupComponent from './utils/lookup-component';
+import lookupComponent from 'ember-templates/utils/lookup-component';
```
krisselden: this is a circular package ref to the pseudo package that is just about reexports, shared stuff should go into ember-views or we can symlink for now.  I want to try to avoid imports from ember-glimmer and ember-htmlbars to ember-templates

packages/ember-htmlbars/lib/component.js
krisselden: @rwjblue yes, but there isn't any reason they couldn't have both used the same technique.  The same glimmer pattern could have been easily moved to where component manager in htmlbars does get('layout') that happens in ensureCompilable.

krisselden: The issue is I want to get these private APIs cleaned up early, I'm hoping no one complains and tries to argue templateForName is intimate but it is best I think to give people some time with canary.

rwjblue: @krisselden - I definitely don't think `templateForName` is an API that we need to support.  I'd be fine if this were moved to a simple util that we call...

rwjblue:
We just need to clarify what needs to change for https://github.com/emberjs/ember.js/pull/13617/files#r66305411, everything else looks good here.

krisselden:
@chadhietala lets not branch where not needed, I'm going to merge, but we should try to get the paths the same as much as possible, so when we improve a path for the glimmer side that isn't actually tied to how glimmer works, we should bring both sides along.  I reduces complexity and confusion.  This forked just simply because glimmer wasn't ready for layout as a prop so it was hacked into the static resolution, if late bound layout had been supported, we wouldn't have forked this.

chadhietala:
Yup I'll take a stab at this once i get the rest of the cleanup PRs in shippable states.

PR #13618, opened 2016-06-07
[DOC helper] fix `mut` helper docs in connection with `action` helper

deepflame:
You not get the `mut` helper to work until I removed the parenthesis around the variable.


rwjblue:
Thank you!

PR #13624, opened 2016-06-08
Added a blueprint util-addon to fix addon generation with pod style

morhook:
This should fix a problem I've reported on https://github.com/ember-cli/ember-cli/issues/5942


morhook:
Should this modification also go to https://github.com/ember-cli/ember-cli-legacy-blueprints ? I'm a little bit confused by that one

rwjblue:
Looks good to me.

@Turbo87 / @trabus - Would one of y'all mind reviewing also?

Turbo87:
I'm wondering if this should be fixed in https://github.com/emberjs/ember.js/blob/master/blueprints/-addon-import.js instead. Is `util` the only blueprint that is having issues with addon+pod?

morhook:
I've tried to fix https://github.com/emberjs/ember.js/blob/master/blueprints/-addon-import.js with no-luck

There are other blueprints failing pretty similar. After some testing I've found out:
* `ember generate helper help-someone --pod`
* `ember generate initializer start-me-up --pod`

Looks like this cases should be ignored on `-add-on-import.js`? Or maybe we can patch the three of them with helper-addon ?

morhook:
Found another generator that is failing on addon + pod. `ember generate instance-initializer start-me-down --pod`

node-tests/blueprints/helper-addon-test.js
```diff
@@ -20,4 +20,14 @@ describe('Acceptance: ember generate and destroy helper-addon', function() {
...
+  
+  it('in-addon helper-addon foo-bar --pod', function() {
+    var args = ['helper-addon', 'foo-bar', '--pod'];
+
+    return emberNew({ target: 'addon' })
+      .then(() => emberGenerateDestroy(args, _file => {
+        expect(_file('app/helpers/foo-bar.js'))
+          .to.contain("export { default, fooBar } from 'my-addon/helpers/foo-bar';");
+      }));
+  });
```
Turbo87: it's certainly useful to have these tests here. I'm just not sure what the expected path layout for Pods in these cases is.

/cc @rwjblue @dgeb

Turbo87:
@morhook I might have put you on the wrong path. I actually meant https://github.com/ember-cli/ember-cli/blob/master/blueprints/addon-import/index.js in my first comment 🙈

morhook:
@Turbo87 I think that file should be patched the same way as on this PR to https://github.com/emberjs/ember.js/blob/master/blueprints/-addon-import.js . Seems pretty similar code :laughing:

locks:
Hi @morhook, your contribution is appreciated but as mentioned in https://github.com/ember-cli/ember-cli/issues/5942#issuecomment-242951458, we are not supporting pods, concentrating efforts on the module unification rfc instead. Thanks for the work and hope to see you around.

PR #13580, opened 2016-06-01
[WIP] [Glimmer2] Unskip some test-helper tests

nickiaconis:
The fix for unskipping `'fillIn' focuses on the element` is currently breaking the HTMLBars version of the test, but I haven't been able to figure out why yet.

PR #13625, opened 2016-06-09
[CLEANUP] Remove view options in routes

chadhietala:
This removes the `viewName` and `view` option in `Route`. This is dependent on #13580  or I can make them part of this PR.

### DIFF
--- a/packages/ember-routing/lib/system/route.js
+++ b/packages/ember-routing/lib/system/route.js
@@ -2117,8 +2117,6 @@ function buildRenderOptions(route, namePassed, isDefaultRender, _name, options) 
     controller.set('model', options.model);
   }
 
-  let viewName = namePassed && name || name;
-  let ViewClass = owner._lookupFactory(`view:${viewName}`);
   let template = owner.lookup(`template:${templateName}`);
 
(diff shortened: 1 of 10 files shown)


packages/ember-routing/lib/system/route.js
```diff
@@ -2150,7 +2117,7 @@ function buildRenderOptions(route, namePassed, isDefaultRender, _name, options)
     controller.set('model', options.model);
   }
 
-  let viewName = options && options.view || namePassed && name || route.viewName || name;
+  let viewName = namePassed && name || name;
   let ViewClass = owner._lookupFactory(`view:${viewName}`);
```
chadhietala: I'm assuming this cannot be removed because Outlets are implemented and registered as `View`?

rwjblue: I believe the whole thing here can be removed.  We shouldn't need any `view:*` lookups here...

rwjblue:
This looks good, can you squash and rebase now that #13626 has landed?

rwjblue:
Rebased and submitted #13642.

PR #13631, opened 2016-06-09
[DOC] Add documentation to second parameter in Ember.computed.deprecatingAlias

### LINKED ISSUE #13630: Using `deprecatingAlias` triggers additional deprecation warnings when used as instructed
Using `deprecatingAlias` [as instructed by the docs](http://emberjs.com/api/classes/Ember.computed.html#method_deprecatingAlias) triggers additional deprecation warnings. In order to use it without triggering additional deprecation warnings, you have to pass in an object with an id and an until property:

`oldAndBusted: computed.deprecatingAlias('newHotness', { id: 'foo', until: '1.42' });`

Documentation should be updated to express that.

Serabe:
Given that this second parameter is passed as is to `Ember.deprecate`, the
documentation just references this other public method.

Fixes #13630

### DIFF
--- a/packages/ember-runtime/lib/computed/computed_macros.js
+++ b/packages/ember-runtime/lib/computed/computed_macros.js
@@ -625,6 +625,7 @@ export function readOnly(dependentKey) {
   @method deprecatingAlias
   @for Ember.computed
   @param {String} dependentKey
+  @param {Object} options Options for `Ember.deprecate`.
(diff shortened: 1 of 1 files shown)


luketheobscure:
I opened a near identical PR here: https://github.com/emberjs/ember.js/pull/13633/files

Serabe:
Closing then. Thank you!

PR #13632, opened 2016-06-09
Add failing test for tagless blockless component regression

amk221:
Related issue #13494

Issue #13494, opened 2016-05-13
[BETA and CANARY] `block` is undefined on tagless and blockless components


miguelcobain:
I'm getting a `TypeError: Cannot read property 'template' of undefined` because `block` is undefined in this line: https://github.com/emberjs/ember.js/blob/master/packages%2Fember-htmlbars%2Flib%2Fhooks%2Fcomponent.js#L57

The tests are ok in ember-release and fail on beta and canary channels.

The tests that fail involve tagless blockless components, so I assume the bug is related to that.

Build log: https://travis-ci.org/miguelcobain/ember-leaflet/jobs/129892214

Serabe:
Seems related to local lookup. // @rwjblue

rwjblue:
This was introduced by https://github.com/emberjs/ember.js/commit/869ed3084eb10ba32c8148fbc08c222b544eead1#diff-7d2b815a163f3a57292f65c26e98c051R56.  When no block is passed, there is no `state.manager.block`.  

The first step would be to submit a failing test PR (and also adding a test for a tagless blockless component using local lookup).  The fix *might* be as simple as guarding there, but I suspect to support local lookup in tagless/blockless we will need to tweak the logic.

asakusuma:
Now that we have a simple repro I'll work on fixing this

rwjblue:
Thanks @asakusuma

rwjblue:
I believe this applies to 2.6.0, so it will likely warrant a point release once fixed.

ForSpareParts:
If anybody else is having this problem, a temporary workaround is to add a tag to your tagless component until the point release lands.

mixonic:
@rwjblue @asakusuma this can be closed?

asakusuma:
@mixonic yes this can be closed

mixonic:
Thanks @asakusuma :-)

miguelcobain:
Looks like this problem isn't solved yet @asakusuma @rwjblue 

I get a `null is not an object (evaluating 'sm.scope._view.template.meta')` error.

Log: https://travis-ci.org/miguelcobain/ember-leaflet/jobs/146079848

Any idea?

asakusuma:
@miguelcobain can you provide a failing test PR?

[...]

PR #13634, opened 2016-06-10
[BUGFIX] Fix blockless component meta resolution

asakusuma:
Addresses https://github.com/emberjs/ember.js/issues/13494
Closes #13632.


asakusuma:
cc @rwjblue

rwjblue:
Can you add another local lookup test that ensures a tagless and block less component can use local lookup (and is able to rerender)?

asakusuma:
@rwjblue How do you write a blockless component that uses local lookup? My understanding is that the component has to define a template where a local lookup occurs, and if there's a template, there's a block.

rwjblue:
Thanks for the added test, I was just concerned about this case (but its nice to have more tests).

This LGTM when green...

PR #13641, opened 2016-06-10
[DOC canary] Adding second arg to findWithAssert and find methods in docs

peculiarism:
And examples too.

### DIFF
--- a/packages/ember-testing/lib/helpers.js
+++ b/packages/ember-testing/lib/helpers.js
@@ -111,8 +111,16 @@ asyncHelper('fillIn', fillIn);
   var $el = find('.my-selector');
   ```
 
+  With the `context` param:
+
+  ```javascript
+  var $el = find('.my-selector', 'parent-element');
+  ```
+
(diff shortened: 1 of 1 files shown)


packages/ember-testing/lib/helpers.js
```diff
@@ -111,8 +111,16 @@ asyncHelper('fillIn', fillIn);
   var $el = find('.my-selector');
   ```
 
+  With the `context` param:
+
+  ```javascript
+  var $el = find('.my-selector', 'parent-element');
```
mixonic: Here and below: Perhaps `'.parent-element-class'`? Using a custom element in an example like this seems a bit cutting edge.

EGlenfield22: Ah, my bad, quick PR and didn't check it through, thank you for pointing that out! :)

mixonic:
Closed in https://github.com/emberjs/ember.js/pull/14107 - Thanks!

PR #13647, opened 2016-06-10
[CLEANUP] Remove the need for templateForName in htmlbars components

chadhietala:
/cc @krisselden @rwjblue


rwjblue:
Thanks @chadhietala!

PR #13648, opened 2016-06-11
[CLEANUP] Remove unnecessary branch in teardown

chadhietala:
@krisselden is there more to this? None of the tests failed when I removed it so wasn't sure if there was more cleanup.


krisselden:
Yes, stuff in the cleanup, that also has stuff around make the view go back to a prerender state for inserting back into the container view. In the template layer all Dom stuff should be done by this point. Only at the top level should you destroy a component and that is done only by tests

Issue #13644, opened 2016-06-10
Glimmer Pre-flight Checklist


chadhietala:
![You're closer than you think](http://www.nasa.gov/sites/default/files/styles/ubernode_alt_horiz/public/11572307115_1194957269_o_0.jpg "You're closing in")

Below is the pre-flight checklist to launch Glimmer. If you are going to take a test to fix please leave a comment on this thread. Please mention this issue in your PR so we can track more effectively. 

## Features
- [x] `{{render}}`
- [x] `sendAction` #13532 🔒 @Joelkang
- [x] `{{mut}}` and `{{readonly}}` #13541 🔒 @Joelkang
- [x] `{{input}}` #13615 🔒 @asakusuma
- [x] Closure components 🔒 @zackthehuman
- [x] Local Lookup 🔒 @asakusuma
- [x] `{{each-in}}` 🔒 @Joelkang #13492

## Testing

__Failing Tests__

- [x] [attrs property of a component without a template should not collide with internal structures](https://github.com/emberjs/ember.js/blob/d638d524b55ebea5b77e662f9be84d3a4be18648/packages/ember/tests/component_registration_test.js#L339)
- [x] [Components receive the top-level view as their ownerView](https://github.com/emberjs/ember.js/blob/d638d524b55ebea5b77e662f9be84d3a4be18648/packages/ember/tests/component_registration_test.js#L404)
- [x] [Using {{link-to}} does not cause an exception if it is rendered without a router.js instance](https://github.com/emberjs/ember.js/blob/60461a144d68835ada96ab3db25cca56c25d02c8/packages/ember/tests/helpers/link_to_test.js#L126)
- [x] [The {{link-to}} helper supports 'classNameBindings' with custom values [GH #11699]](https://github.com/emberjs/ember.js/blob/60461a144d68835ada96ab3db25cca56c25d02c8/packages/ember/tests/helpers/link_to_test.js#L512)
- [x] [The Special page returning an error invokes SpecialRoute's error handler](https://github.com/emberjs/ember.js/blob/93a9aef89f1929dee5b07e38053b35bd420fbf08/packages/ember/tests/routing/basic_test.js#L769)
- [x] [ApplicationRoute's default error handler can be overridden](https://github.com/emberjs/ember.js/blob/93a9aef89f1929dee5b07e38053b35bd420fbf08/packages/ember/tests/routing/basic_test.js#L855)
[...]

PR #13649, opened 2016-06-11
[GLIMMER] Dashed property names

chadhietala:
Sort of sad that this is even a feature. @krisselden @wycats or @chancancode do you have any suggestions on this, I didn't want to go weaving `isBlock` all over the place.

Part of #13644

### DIFF
--- a/packages/ember-glimmer/lib/environment.js
+++ b/packages/ember-glimmer/lib/environment.js
@@ -121,7 +121,43 @@ export default class Environment extends GlimmerEnvironment {
     };
   }
 
+   // Hello future traveler, welcome to the world of syntax refinement.
+   // The method below is called by Glimmer's runtime compiler to allow
+   // us to take generic statement syntax and refine it to more meaniful
+   // syntax for Ember's use case. This on the fly switch-a-roo sounds fine
+   // and dandy, however Ember has precedence on statement refinement that you
+   // need to be aware of. The presendence for language constructs is as follows:
+   //
+   // ------------------------
+   // Native & Built-in Syntax
+   // ------------------------
+   //   User-land components
+   // ------------------------
+   //     User-land helpers
+   // ------------------------
+   //
+   // The one caveat here is that Ember also allows for dashed references that are
+   // not a component or helper:
+   //
+   // export default Component.extend({
+   //   'foo-bar': 'LAME'
+   // });
+   //
+   // {{foo-bar}}
+   //
+   // The heuristic for the above situation is a dashed "key" in inline form
+   // that does not resolve to a defintion. In this case refine statement simply
+   // isn't going to return any syntax and the Glimmer engine knows how to handle
+   // this case.
+
   refineStatement(statement) {
+    // 1. resolve any native syntax – if, unless, with, each, and partial
+    let nativeSyntax = super.refineStatement(statement);
+
+    if (nativeSyntax) {
+      return nativeSyntax;
+    }
+
     let {
       isSimple,
       isInline,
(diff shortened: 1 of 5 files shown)


packages/ember/tests/component_registration_test.js
```diff
@@ -139,18 +139,6 @@ QUnit.test('Late-registered components can be rendered with ONLY the template re
   ok(!helpers['borf-snorlax'], 'Component wasn\'t saved to global helpers hash');
 });
 
-test('Component-like invocations are treated as bound paths if neither template nor component are registered on the container', function() {
```
chadhietala: Duplicated test.

rwjblue:
This seems good to me...

chancancode:
The syntax refinement in ember is getting pretty confusing, what about something like this?

```
diff --git a/packages/ember-glimmer/lib/environment.js b/packages/ember-glimmer/lib/environment.js
index ebe1df4..2c769f9 100644
--- a/packages/ember-glimmer/lib/environment.js
+++ b/packages/ember-glimmer/lib/environment.js
@@ -122,6 +122,13 @@ export default class Environment extends GlimmerEnvironment {
   }
 
   refineStatement(statement) {
+    // 1. resolve any native syntax – if, unless, with, each, and partial
+    let nativeSyntax = super.refineStatement(statement);
+
+    if (nativeSyntax) {
+      return nativeSyntax;
+    }
+
     let {
       isSimple,
       isInline,
@@ -133,37 +140,34 @@ export default class Environment extends GlimmerEnvironment {
       templates
     } = statement;
 
-    if (key !== 'partial' && isSimple && (isInline || isBlock)) {
+    if (isSimple && (isInline || isBlock)) {
+      // 2. built-in syntaxs
       if (key === 'component') {
         return new DynamicComponentSyntax({ args, templates });
       } else if (key === 'outlet') {
         return new OutletSyntax({ args });
+      }
+
+      // 3. resolve components
+      let internalKey = builtInComponents[key];
+      let definition = null;
+
+      if (internalKey) {
+        definition = this.getComponentDefinition([internalKey]);
       } else if (key.indexOf('-') >= 0) {
-        let definition = this.getComponentDefinition(path);
-
-        if (definition) {
-          wrapClassBindingAttribute(args);
-          wrapClassAttribute(args);
-          return new CurlyComponentSyntax({ args, definition, templates });
-        } else if (isBlock && !this.hasHelper(key)) {
-          assert(`A helper named '${path[0]}' could not be found`, false);
-        }
-      } else {
-        // Check if it's a keyword
-        let mappedKey = builtInComponents[key];
-        if (mappedKey) {
-          let definition = this.getComponentDefinition([mappedKey]);
-          wrapClassBindingAttribute(args);
-          wrapClassAttribute(args);
-          return new CurlyComponentSyntax({ args, definition, templates });
-        }
+        definition = this.getComponentDefinition(path);
+      }
+
+      if (definition) {
+        wrapClassBindingAttribute(args);
+        wrapClassAttribute(args);
+        return new CurlyComponentSyntax({ args, definition, templates });
       }
     }
 
-    let nativeSyntax = super.refineStatement(statement);
-    assert(`Helpers may not be used in the block form, for example {{#${key}}}{{/${key}}}. Please use a component, or alternatively use the helper in combination with a built-in Ember helper, for example {{#if (${key})}}{{/if}}.`, !nativeSyntax && key && this.hasHelper(key) ? !isBlock : true);
-    assert(`Helpers may not be used in the element form.`, !nativeSyntax && key && this.hasHelper(key) ? !isModifier : true);
-    return nativeSyntax;
+    assert(`Helpers may not be used in the block form, for example {{#${key}}}{{/${key}}}. Please use a component, or alternatively use the helper in combination with a built-in Ember helper, for example {{#if (${key})}}{{/if}}.`, !isBlock || !this.hasHelper(key));
+    assert(`Helpers may not be used in the element form.`, !isModifier || !this.hasHelper(key));
+    assert(`Could not find component named "${key}" (no component or template with that name was found)`, isBlock);
   }
 
   hasComponentDefinition() {
@@ -179,8 +183,6 @@ export default class Environment extends GlimmerEnvironment {
 
       if (ComponentClass || layout) {
         definition = this._components[name] = new CurlyComponentDefinition(name, ComponentClass, layout);
-      } else if (!this.hasHelper(name)) {
-        assert(`Glimmer error: Could not find component named "${name}" (no component or template with that name was found)`, !!(ComponentClass || layout));
       }
     }
 ```

**Comment by chadhietala** [comment]
@chancancode I tried this locally and all of the late bound layout stuff is going to fail.

File: packages/ember-glimmer/lib/environment.js [hunk 2]
```diff
@@ -133,37 +170,37 @@ export default class Environment extends GlimmerEnvironment {
       templates
     } = statement;
 
-    if (key !== 'partial' && isSimple && (isInline || isBlock)) {
+    assert(`You attempted to overwrite the built-in syntax "${key}" which is not allowed. Please rename the helper.`, !(builtInHelpers[key] && this.owner.hasRegistration(`helper:${key}`)));
```
chancancode: s/syntax/helper

packages/ember-glimmer/lib/syntax/dynamic-component.js
```diff
@@ -10,6 +11,8 @@ function dynamicComponentFor(vm) {
     let name = nameRef.value();
     let definition = env.getComponentDefinition([name]);
 
+    assert(`Glimmer error: Could not find component named "${name}" (no component or template with that name was found)`, definition);
```
chancancode: I think we can drop the "Glimmer error" prefix here too

packages/ember-glimmer/lib/syntax/dynamic-component.js
```diff
@@ -44,6 +47,9 @@ class DynamicComponentReference {
 
     if (typeof name === 'string') {
       let definition = env.getComponentDefinition([name]);
+
+      assert(`Glimmer error: Could not find component named "${name}" (no component or template with that name was found)`, definition);
```
chancancode: ditto

chancancode:
LGTM 👍

krisselden:
There was a problem with installing node earlier, I refreshed Travis. The only failure now is linting. You added a block comment and didn't specify @private or @public

chadhietala:
Should be good to go now.

PR #13653, opened 2016-06-12
Run chrome/firefox in Travis (not sauce).

rwjblue:


rwjblue:
@krisselden / @stefanpenner - Ready for review...

bin/run-travis-browser-tests.js
```diff
@@ -0,0 +1,68 @@
...
+    child.stdout.on('data', function(data) {
+      console.log(data.toString());
+    });
+
+    child.stderr.on('data', function(data) {
+      console.error(data.toString());
+    });
+
+    child.on('error', function(err) {
+      reject(err);
```
stefanpenner: we may need to force kill the child here: https://nodejs.org/api/child_process.html#child_process_child_kill_signal

bin/run-travis-browser-tests.js
```diff
@@ -0,0 +1,68 @@
...
+  })
+  .then(function() {
+    return run('./node_modules/.bin/testem', ['launchers']);
+  })
+  .then(function() {
+    return run('./node_modules/.bin/testem', ['ci', '--file', 'testem.travis-browsers.js', '--port', '7000']);
+  })
+  .catch(function(error) {
+    console.log('error:');
+    console.log(error);
```
stefanpenner: should we also print the stack?

stefanpenner:
some comments but largely looks great.

---

Future work: this may make a really good add-on, that our travis users could install and get great FF/chrome coverage right from there CLI apps.

PR #13655, opened 2016-06-12
[BUGFIX release] Fix inlining of superWrapper.

krisselden:
This fixes the inlining of superWrapper.  The bug that copying the arguments was addressing has been fixed for long enough that more users are affected by fix than the fix benefits.  This also fixes a deopt in the megamorphic load of toString, since we store metadata on functions we have many shapes of functions.

I also retested call vs apply branching in modern browsers and did not see any difference between the two.

### DIFF
--- a/packages/ember-metal/lib/events.js
+++ b/packages/ember-metal/lib/events.js
@@ -7,10 +7,7 @@
 @submodule ember-metal
 */
 import { assert } from 'ember-metal/debug';
-import {
-  apply,
-  applyStr
-} from 'ember-metal/utils';
+import { applyStr } from 'ember-metal/utils';
 import { meta as metaFor, peekMeta } from 'ember-metal/meta';
 import { deprecate } from 'ember-metal/debug';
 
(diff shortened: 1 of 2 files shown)


stefanpenner:
bugfix release or beta?

stefanpenner:
This is functionally equivalent, and @krisselden has done research across a large number of mobile devices. It appears the devices this affected the most, are now in a minority, and the others are  adversely affected by this.

@rwjblue I think this could be `[Bugfix release]` but I would settle for `[Bugfix beta]` if you feel it is appropriate.

rwjblue:
I am happy with either here.  We already have to do a 2.6.1 so `[BUGFIX release]` seems fine...

rwjblue:
@krisselden - Can you remove the `apply` export from `ember-metal/utils`?

rwjblue:
FWIW - I think that `apply` here was possible the single largest issue with stepping through Ember app code...

stefanpenner:
> FWIW - I think that apply here was possible the single largest issue with stepping through Ember app code...

have you met the run-loop ? :trollface:  🏃

rwjblue:
Thanks @krisselden!

knownasilya:
So beautiful! Thanks @krisselden and company!

PR #13659, opened 2016-06-13
[WIP] Refactor Meta's deps structure to an array of arrays.

rwjblue:
/cc @krisselden @ef4

### DIFF
--- a/packages/ember-metal/lib/meta.js
+++ b/packages/ember-metal/lib/meta.js
@@ -40,7 +40,7 @@ let members = {
   mixins: inheritedMap,
   bindings: inheritedMap,
   values: inheritedMap,
-  deps: inheritedMapOfMaps,
+  deps: inheritedMapOfArrays,
   chainWatchers: ownCustomObject,
   chains: inheritedCustomObject,
   tag: ownCustomObject
@@ -101,6 +101,14 @@ Meta.prototype._getOrCreateOwnMap = function(key) {
   let ret = this[key];
   if (!ret) {
     ret = this[key] = new EmptyObject();
+  }
+  return ret;
+};
+
+Meta.prototype._getOrCreateOwnArray = function(key) {
+  let ret = this[key];
+  if (!ret) {
+    ret = this[key] = [];
   }
   return ret;
 };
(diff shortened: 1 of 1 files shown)


rwjblue:
@krisselden - I updated this to use an array of arrays (instead of a map of arrays).  I left it in two commits so that it was easy to compare the difference (in case you end up liking the map of arrays approach better).

The best summary of the new structure is from `writeDep` where we setup the structure:

```js
    // add the subkey to the outerArray
    keyArray.push([
      subkey,
      // array of itemKeys for the subkey
      [
        [itemKey, value]
      ]
    ]);
```

machty:
this used to be 5.4 http://numberofpeoplewhounderstandembermetal.com/

packages/ember-metal/lib/meta.js
```diff
@@ -176,45 +176,80 @@ Meta.prototype._findInherited = function(key, subkey) {
...
+function inheritedArrayOfArrays(name, Meta) {
   let key = memberProperty(name);
   let capitalized = capitalize(name);
 
-  Meta.prototype['write' + capitalized] = function(subkey, itemkey, value) {
-    let outerMap = this._getOrCreateOwnMap(key);
-    let innerMap = outerMap[subkey];
-    if (!innerMap) {
-      innerMap = outerMap[subkey] = new EmptyObject();
+  Meta.prototype['write' + capitalized] = function(subkey, itemKey, value) {
```
chadhietala: Nit can you do ``write${capitalized}``

packages/ember-metal/lib/meta.js
```diff
@@ -176,45 +176,80 @@ Meta.prototype._findInherited = function(key, subkey) {
 
 export const UNDEFINED = symbol('undefined');
 
-// Implements a member that provides a lazily created map of maps,
+// Implements a member that provides a lazily created array of arrays,
 // with inheritance at both levels.
-function inheritedMapOfMaps(name, Meta) {
+function inheritedArrayOfArrays(name, Meta) {
   let key = memberProperty(name);
   let capitalized = capitalize(name);
```
chadhietala: Is it possible to just pass the capitalized name at the call site?

packages/ember-metal/lib/meta.js
```diff
@@ -228,20 +263,26 @@ Meta.prototype._forEachIn = function(key, subkey, fn) {
...
+    let keyArray = pointer[key];
+    if (keyArray) {
+      for (let i = 0; i < keyArray.length; i++) {
+        if (keyArray[i][0] === subkey) {
+          let subkeyArray = keyArray[i][1];
+
+          for (let j = 0; j < subkeyArray.length; j++) {
+            let innerKey = subkeyArray[j][0];
+
+            if (!seen[innerKey]) {
```
chadhietala: Should `seen` be a `Dict`? What is the upper bound of `seen`?

rwjblue:
Updated #13660 to migrate to using the lodash-like object interface for this functionality.  Closing this in favor of that.

@chadhietala - I'll make sure to incorporate your comments there though...

PR #13660, opened 2016-06-13
Use a map to back meta's cache.

rwjblue:
This is a first step to making the baking store of `Meta`'s underlying data structures swappable. The first implementation is using a simple array of key value pairs, but since we are using the `set`, `get`, `has`, `delete` interface we can change the backing data store at will (or even "upgrade" at runtime).

The underlying implementation of the "lodash-stack" is mostly copied from lodash itself (with inline attribution), but in the longer term we should simply use this from lodash itself (and embed these modules from lodash-es in our build process).

/cc @krisselden @ef4

### DIFF
--- a/packages/ember-metal/lib/chains.js
+++ b/packages/ember-metal/lib/chains.js
@@ -182,8 +182,8 @@ function lazyGet(obj, key) {
   // Otherwise attempt to get the cached value of the computed property
   } else {
     let cache = meta.readableCache();
-    if (cache && key in cache) {
-      return cache[key];
+    if (cache && cache.has(key)) {
+      return cache.get(key);
     }
   }
 }
--- a/packages/ember-metal/lib/computed.js
+++ b/packages/ember-metal/lib/computed.js
@@ -316,8 +316,8 @@ ComputedPropertyPrototype.didChange = function(obj, keyName) {
   }
 
   let cache = meta.readableCache();
-  if (cache && cache[keyName] !== undefined) {
-    cache[keyName] = undefined;
+  if (cache && cache.get(keyName) !== undefined) {
+    cache.set(keyName, undefined);
     removeDependentKeys(this, obj, keyName, meta);
   }
 };
@@ -329,8 +329,8 @@ ComputedPropertyPrototype.get = function(obj, keyName) {
 
   let meta = metaFor(obj);
   let cache = meta.writableCache();
+  let result = cache.get(keyName);
 
-  let result = cache[keyName];
   if (result === UNDEFINED) {
     return undefined;
   } else if (result !== undefined) {
@@ -339,9 +339,9 @@ ComputedPropertyPrototype.get = function(obj, keyName) {
 
   let ret = this._getter.call(obj, keyName);
   if (ret === undefined) {
-    cache[keyName] = UNDEFINED;
+    cache.set(keyName, UNDEFINED);
   } else {
-    cache[keyName] = ret;
+    cache.set(keyName, ret);
   }
 
   let chainWatchers = meta.readableChainWatchers();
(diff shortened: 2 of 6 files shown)


ef4:
Looks good to me.

For those following along at home, the motivation here is that it turns out to be bad for performance to use a lot of plain old Javascript objects as if they were `Map`s because you rapidly use up V8's inline cache. @krisselden should correct me if I'm not summarizing the problem correctly, the analysis was all his work.

krisselden:
@rwjblue I'm still making my way back from the face to face it is going to be a bit before I can benchmark this

rwjblue:
Updated:

* incorporate changes from #13659
* Add `forEach` to the interface used from the lodash-like map
* Refactored various formatting related things

>  it is going to be a bit before I can benchmark this

@krisselden - No giant rush from my end. Please let me know if I can do anything to help...

rwjblue:
The following storage strategies exist and are used from the new `Stack` implementation:

* `ListCache`
* `Hash`
* `MapCache`

##### `ListCache`

This is the initial storage mechanism. The underlying data format is as follows:

```js
[
  [key, value]
  [key, value]
  [key, value]
  [key, value]
]
```

##### `Hash`

The underlying data format is as follows:

```js
{
  [key]: value
}
```

The base object is an `EmptyObject` (roughly `Object.create(null)`).

##### `MapCache`

`MapCache` defers to different underlying storage mechanisms based on
the key being stored:

* `string` -- Utilizes a `Hash` dedicated to string keys.
* `number` / `symbol` / `boolean` -- Utilizes a dedicated `Hash` for
  these types (all are stored in the same `Hash`).
* Everything else falls back to using `ListCache`.

---

The overall strategy is to store items in the `ListCache` until it has 200 items in it, then we translate from `ListCache` to `MapCache` (which actually uses 3 different structures internally depending on which object type is being used as a key).

All credit to this mechanism goes to Lodash, they did incredible work here and we are just inlining here while we test things out. Once the performance of this within Ember is validated, we will migrate away from this locally vendored version of this file and simply use a stripped version of `lodash` itself from the build.

rwjblue:
@chadhietala - I believe I have incorporated all of your suggestions from #13659.

rwjblue:
@krisselden - I backported this into 2.6 for easier performance testing.  The built assets are [here](https://github.com/components/ember/compare/master...rwjblue:meta-cache-object-tests), and you can install by changing `bower.json` in an ember-cli app to use `rwjblue/ember#meta-cache-object-tests` for `ember`.

jonnii:
@rwjblue are you guys going to publish the performance results? i would find that super interesting.

stefanpenner:
Also interested in performance metrics (across devices and in some non-trivial apps). This stuff is always tricky, but I do think this path is good to explore, and I hope it works out.

ef4:
We may not see immediate performance benefits if the inline cache budget is still being blown by other parts of the codebase. @krisselden, is there a good way to measure progress on a more inline-cache-specific metric? It would be awesome to prevent regressions in CI, so we can keep driving the usage monotonically downward.

stefanpenner:
> We may not see immediate performance benefits if the inline cache budget is still being blown by other parts of the codebase

We need to find some criteria for merging speculative fixes (that require more steps to fully realize). I think your idea of seeing if we can see if we are reducing the pressure or not, coupled with ensuring minimal regressions will do the trick.

rwjblue:
I agree, we should definitely confirm that there are fewer IC misses on this branch are than on master, and also confirm that this isn't slower than the current implementation.

However, I also think that this interface (`has`, `get`, `set`) is generally an improvement because it allows us to easily change how we store in the future.

rwjblue:
I did some benchmarking of this using [travis-web](https://github.com/travis-ci/travis-web).  Detailed [gist is here](https://gist.github.com/rwjblue/02419014a63c1d31ae1647b8e3bceac4).

boxplot
histogram

@krisselden - Have you had a chance to test this?

nathanhammond:
Those five outlier tests are concerning. Worst case is worse, median case is better. This is what Safari optimizes for on consistency inside of JSC and Intel on their SSDs. Best-case/worst-case boundaries to prevent jank. Done again I'd be curious to see if it had the same number of outliers, and why those outliers were that much slower since everything else was in a tight cluster.

stefanpenner:
> Those five outlier tests are concerning.

Sorta by definition, we should ignore the outliers, unless test isn't evenly distributing the work.

nathanhammond:
@stefanpenner It's 5 "outliers" in 30 runs; so 1/6th of the total number of runs are between 2-3x slower than the median case.

mmun:
Superceded by https://github.com/emberjs/ember.js/pull/16222.

PR #13667, opened 2016-06-14
Support jQuery 3.0

### LINKED ISSUE #13661: EmberJS does not work with jQuery 3.0
`jQuery.event.fixHooks` was removed in jQuery 3.0

https://jquery.com/upgrade-guide/3.0/#breaking-change-jquery-event-props-and-jquery-event-fixhooks-removed

https://github.com/emberjs/ember.js/blob/f6bad3dad476f952d79c3386bf9f7be12b075380/packages/ember-views/lib/system/jquery.js#L23

tricknotes:
I use `jQuery.event.addProp` instead of `jQuery.event. fixHooks` to work with jQuery 3.
It seems that `addProp` is introduced by https://github.com/jquery/jquery/commit/e61fccb9d736235b4b011f89cba6866bc0b8997d instead of `fixHooks`.

It has no ability to extend individual event.
However, what extending `jQuery.Event` in globally has no bad effect as far as I know.

Close https://github.com/emberjs/ember.js/issues/13661

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -5,7 +5,7 @@
     "url": "https://github.com/emberjs/ember.js.git"
   },
   "dependencies": {
-    "jquery": ">=1.11.1 <3.0.0",
+    "jquery": ">=1.11.1 <4.0.0",
     "qunit": "^1.20.0",
     "qunit-phantom-runner": "jonkemp/qunit-phantomjs-runner#1.2.0"
   },
--- a/bin/run-tests.js
+++ b/bin/run-tests.js
@@ -126,6 +126,9 @@ function generateOldJQueryTests() {
   testFunctions.push(function() {
     return run('jquery=1.10.2&nojshint=true');
   });
+  testFunctions.push(function() {
+    return run('jquery=2.2.4&nojshint=true');
+  });
 }
 
 function generateExtendPrototypeTests() {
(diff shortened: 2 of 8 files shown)


bower.json
```diff
@@ -5,7 +5,7 @@
     "url": "https://github.com/emberjs/ember.js.git"
   },
   "dependencies": {
-    "jquery": ">=1.11.1 <3.0.0",
```
rwjblue: I do not think we are using private API's after this change, perhaps this should just be `>= 1.11.1 < 4.0.0`?

rwjblue:
For context https://github.com/emberjs/ember.js/issues/1073 was the original issue that caused us to add `fixHooks` usage for `dataTransfer`.

rwjblue:
This fix looks good.  

@emberjs/core - r?

bower.json
tricknotes: That's right. I rebased it as https://github.com/tricknotes/ember.js/commit/437726f63476e3ffd1ceabfeeb37b434b9fbdbb8 .

tricknotes:
Thanks for your in-depth explanation about this change's background.
It should have been described from me :<

rwjblue:
@tricknotes - Can you rebase?

tricknotes:
Rebased.

rwjblue:
Thanks again @tricknotes!

PR #13671, opened 2016-06-14
[Glimmer 2] Remove tests related to legacy ember-test-helpers

chadhietala:
These tests are specifically testing patterns that used to be used in
ember-test-helpers. Since these tests have been introduced
ember-test-helpers has moved to the container pattern with
`lookupFactory`. In general we need to start migrating tests away from
things that did not come out of the container with injections on them.

This is because we need to remove this hardcoded `InteractiveRenderer`
[here](https://github.com/emberjs/ember.js/blob/master/packages/ember-views/lib/views/core_view.js#L46-L52)

Part of #13644

### DIFF
--- a/packages/ember/tests/helpers/link_to_test.js
+++ b/packages/ember/tests/helpers/link_to_test.js
@@ -8,14 +8,11 @@ import { subscribe, reset } from 'ember-metal/instrumentation';
 import isEnabled from 'ember-metal/features';
 import alias from 'ember-metal/alias';
 import Application from 'ember-application/system/application';
-import Component from 'ember-templates/component';
-import ComponentLookup from 'ember-views/component_lookup';
 import jQuery from 'ember-views/system/jquery';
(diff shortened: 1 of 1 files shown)


chadhietala:
@rwjblue Can you confirm this is ok?

rwjblue:
I'd like to keep the tests, but we need to build the registry properly.  The issue here is that these tests do not use `Ember.Application.buildRegistry` to create the registry instance. They passed before because there were no services that were used that weren't setup, but they fail in glimmer because `service:-document` isn't present and the Environment requires it.

chadhietala:
Yea I walked through the code to figure out why it was failing. Do you know of other tests that go through `buildRegistry`? I would just move these to the new build infra, but these tests are sort of in an in between state of what the new infra supports.

rwjblue:
I think moving them to ember package may be a better bet.

rwjblue:
Rebased in https://github.com/emberjs/ember.js/pull/13713

PR #13680, opened 2016-06-15
[BUGFIX] Remove code for IE <9

tricknotes:
`document.createEventObject()` is an API for legacy IE.
It is deprecated since IE 9 and `document.createEvent()` (that is a DOM Level 3 standard API) is supported.

ref: https://msdn.microsoft.com/en-us//library/ff986080(v=vs.85).aspx


rwjblue:
Thanks!

PR #13682, opened 2016-06-15
[BUGFIX] normalize query params when merging in from active transition

fivetanley:
rebased version of https://github.com/emberjs/ember.js/pull/13273, plus added test from @igorT.

From the original PR/commit message (i lazily squashed them into one, @raytiley may want to help edit):

Normalize query params when merging in from active transition

This merges in queryParams from an active transition in such a way that
`prepareQueryParams` won't duplicate them. It also won't merge in the
qp if it is already int he qp changed list, for example if it's already
been set during route setup by a call to `controller.set('qpProp',
'someValue')`. Fixing these things with my current test jsbin also
recreated another issue that has been biting people and that is that a
call to `transitionTo` creates an aborted transition that calls
`didTransition` with an empty set of handlerInfos. This would cause the
`updatePaths` method to fail since it couldn't calculate the paths
without the handler infos. We now bail early from `updatePaths` it is
called with an empty array of handlerInfos. I think this is fine
because the aborted transition is followed by a successful transition
that will update the paths appropriatly.

This gets the active queryParams and the provided using the same key
`controller:prop` so that the calls to `assign` override as expected.
This fixes issues where the QP doesn't update properly on refresh
because the merged value conflicts with the provided value.

### DIFF
--- a/packages/ember-routing/lib/system/router.js
+++ b/packages/ember-routing/lib/system/router.js
@@ -674,6 +674,8 @@ const EmberRouter = EmberObject.extend(Evented, {
       assign(queryParams, this.router.activeTransition.queryParams);
     }
 
+    this._processActiveTransitionQueryParams(targetRouteName, models, queryParams, _queryParams);
+
     assign(queryParams, _queryParams);
     this._prepareQueryParams(targetRouteName, models, queryParams);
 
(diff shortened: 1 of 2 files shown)


packages/ember/tests/routing/query_params_test.js
```diff
@@ -2490,6 +2491,43 @@ if (isEnabled('ember-routing-route-configured-query-params')) {
     bootApplication();
   });
 
+  QUnit.test('queryParams are updated when a controller property is set and the route is refreshed. Issue #13263  ', function() {
+    Ember.TEMPLATES.application = compile(
```
rwjblue: Should be using `setTemplate` here.

fivetanley:
@rwjblue @raytiley Rebased and fixed tests. I have a clean build locally, I think Travis will say the same.

packages/ember-routing/lib/system/router.js
```diff
@@ -999,6 +1044,8 @@ function calculatePostTransitionState(emberRouter, leafRouteName, contexts) {
 
 function updatePaths(router) {
   let infos = router.router.currentHandlerInfos;
+  if (infos.length === 0) { return; }
```
raytiley: I added this in to try and fix the same issue that this was trying to resolve:

raytiley:
I commented on the one line that I'm not sure should be there any longer since I was never able to create a test for it. Other than that this looks like what I intended. Thanks everyone for picking up where I left off.

rwjblue:
@fivetanley - Failing with optional tests enabled.  You can test locally via: `http://localhost:7200/tests/index.html?hidepassed&hideskipped&enableoptionalfeatures`...

raytiley:
@fivetanley thanks for pushing to finally fix this. I took your rebase and fixed the tests: in #13273

PR #13683, opened 2016-06-15
[DOC release] fixed broken url in docs

mrosata:
Replaced broken url with a "current" url as discussed with @pixelhandler. I just removed the trailing whitespace pointed out by @rwjblue, thanks.


rwjblue:
Thank you @mrosata!

PR #13684, opened 2016-06-15
[Glimmer 2] Implement mut and readonly helpers

chancancode:
Also, implement support for the [INVOKE] symbol in closure actions.

This also enables (partial) "reference pooling" on components – when doing rending a `{{foo}}` in the `{{foo-bar}}` layout, it first checks if `{{foo}}` is an arg passed from the outside (`{{foo-bar foo=...}}`), and if so, it tries to reuse the same reference instead of making a newone. This should reduce the total number of references created and also reduce the number of `value()` computation as the same reference can be reused across multiple curlies.

This also allows Glimmer to better optimize things – for example, if the component is invoked with literals (`{{foo-bar foo=true}}`) and `foo` is then used in a conditional in its layout (`{{#if foo}}...`), Glimmer will be now able to see that `foo` is a const reference and optimize out the updating step. This works recursively as well – if `foo-bar` then passes the same property down (`{{bar-baz bar=foo}}`), we now propagate the same reference all the way down.

Thanks @Joelkang for doing most of the work on this one in #13541!

Closes #13541

### DIFF
--- a/packages/ember-glimmer/lib/helpers/action.js
+++ b/packages/ember-glimmer/lib/helpers/action.js
@@ -1,9 +1,12 @@
 import { CachedReference } from '../utils/references';
-import { NULL_REFERENCE, UNDEFINED_REFERENCE } from 'glimmer-runtime';
 import EmberError from 'ember-metal/error';
+import symbol from 'ember-metal/symbol';
 import run from 'ember-metal/run_loop';
 import { get } from 'ember-metal/property_get';
 import { flaggedInstrument } from 'ember-metal/instrumentation';
+import isNone from 'ember-metal/is_none';
+
+export const INVOKE = symbol('INVOKE');
 
 export class ClosureActionReference extends CachedReference {
   static create(args) {
@@ -23,6 +26,7 @@ export class ClosureActionReference extends CachedReference {
     let positionalValues = positional.value();
 
     let target = positionalValues[0];
+    let rawActionRef = positional.at(1);
     let rawAction = positionalValues[1];
 
     // The first two argument slots are reserved.
@@ -31,15 +35,18 @@ export class ClosureActionReference extends CachedReference {
     // Anything else is an action argument.
     let actionArgs = positionalValues.slice(2);
 
-    // TODO: Check if rawAction is INVOKE-able or (mut)
-
     // on-change={{action setName}}
     // element-space actions look to "controller" then target. Here we only
     // look to "target".
     let actionType = typeof rawAction;
     let action = rawAction;
 
-    if (actionType === 'string') {
+    if (isNone(rawAction)) {
+      throw new EmberError(`Action passed is null or undefined in (action) from ${target}.`);
+    } else if (rawActionRef[INVOKE]) {
+      target = rawActionRef;
+      action = rawActionRef[INVOKE];
+    } else if (actionType === 'string') {
       // on-change={{action 'setName'}}
       let actionName = rawAction;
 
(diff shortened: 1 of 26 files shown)


packages/ember-glimmer/lib/syntax/curly-component.js
```diff
@@ -102,17 +102,17 @@ class CurlyComponentManager {
...
     assert('You cannot use `classNameBindings` on a tag-less component: ' + component.toString(), () => {
       let { classNameBindings, tagName } = component;
-      return tagName || !classNameBindings || classNameBindings.length === 0;
+      return tagName !== '' || !classNameBindings || classNameBindings.length === 0;
     });
 
     assert('You cannot use `elementId` on a tag-less component: ' + component.toString(), () => {
       let { elementId, tagName } = component;
-      return tagName || (!elementId && elementId !== '');
+      return tagName !== '' || (!elementId && elementId !== '');
```
chancancode: This is causing some tests to randomly fail with this assertion. The default `tagName` for all components is intentionally `null`, and that gets converted into a 'div' later down the road. I don't know how this is not causing a lot more tests to fail – it might indicate a deeper problem, e.g. we are accidentally leaking `tagName: '...'` onto the prototype, etc.

chancancode: Fixed in https://github.com/emberjs/ember.js/pull/13684/commits/e0672cde9aa33e0134075f5a1fb26b06a0b40ef2

packages/ember-glimmer/lib/helpers/action.js
```diff
@@ -57,14 +64,15 @@ export class ClosureActionReference extends CachedReference {
       if (!action) {
         throw new EmberError(`An action named '${actionName}' was not found in ${target}`);
       }
+    } else if (action && typeof action[INVOKE] === 'function') {
+      target = action;
+      action = action[INVOKE];
     } else if (actionType !== 'function') {
-      throw new EmberError(`An action could not be made for \`${rawAction}\` in ${target}. Please confirm that you are using either a quoted action name (i.e. \`(action '${rawAction}')\`) or a function available in ${target}.`);
+      // TODO: Is there a better way of doing this?
```
rwjblue: Did you leave this TODO in on purpose?  Is there a better way to do this? :trollface:

packages/ember-glimmer/lib/syntax/curly-component.js
```diff
@@ -89,7 +89,7 @@ class CurlyComponentManager {
       bucket.classRef = args.named.get('class');
     }
 
-    assert('classNameBindings must not have spaces in them', () => {
+    assert('classNameBindings must not have spaces in them', (() => {
```
rwjblue: Can you add `component.toString()` in this message? I submitted https://github.com/emberjs/ember.js/pull/13685 to do this, but no need to land that and cause a rebase conflict...

packages/ember-glimmer/lib/utils/process-args.js
```diff
@@ -47,16 +50,39 @@ class SimpleArgs {
...
+const MUTABLE_CELL = symbol('MUTABLE_CELL');
+
+export function isCell(val) {
+  return val && val[MUTABLE_CELL];
+}
+
+class MutableCell {
+  constructor(ref, value) {
+    this[MUTABLE_CELL] = true;
+    this._ref = ref;
```
rwjblue: Can you make this a symbol too?

rwjblue:
Left a few nit-picks, but this LGTM once those are addressed...

packages/ember-glimmer/lib/helpers/action.js
Joelkang: This actually was in reference to the line below the todo: 
`let rawActionLabel = rawActionRef._propertyKey || rawAction;`

There's no guarantee, I think, that rawActionRef is a propertyReference, and accessing a private property on the reference anyway is problematic.

PR #13685, opened 2016-06-15
[GLIMMER] Add component name to classNameBindings assertion.

rwjblue:


rwjblue:
Should be fixed in #13684

PR #13688, opened 2016-06-16
[FEATURE ember-metal-weakmap] Add a WeakMap polyfill

mmun:
This patch exposes the Ember's private WeakMap polyfill under a feature flag. This implementation has a small but important caveat. It assumes that the weak map will live longer (in the sense of garbage collection) than all of its keys, otherwise it is possible to leak the values stored in the weak map.

This patch also includes a few improvements:
- implement WeakMap constructor
- implement WeakMap toString
- guard against non-Object keys

### DIFF
--- a/packages/ember-metal/lib/weak_map.js
+++ b/packages/ember-metal/lib/weak_map.js
@@ -1,4 +1,3 @@
-import { assert } from 'ember-metal/debug';
 import { GUID_KEY } from 'ember-metal/utils';
 import {
   peekMeta,
@@ -8,9 +7,15 @@ import {
 let id = 0;
 function UNDEFINED() {}
 
+// Returns whether Type(value) is Object according to the terminology in the spec
+function isObject(value) {
+  return (typeof value === 'object' && value !== null) || typeof value === 'function';
+}
+
 /*
- * @private
  * @class Ember.WeakMap
+ * @public
+ * @category ember-metal-weakmap
  *
  * A partial polyfill for [WeakMap](http://www.ecma-international.org/ecma-262/6.0/#sec-weakmap-objects).
  *
@@ -20,13 +25,23 @@ function UNDEFINED() {}
  * practice, most use cases satisfy this limitation which is why it is included
  * in ember-metal.
  */
-export default function WeakMap() {
-  assert(
-    'Invoking the WeakMap constructor with arguments is not supported at this time',
-    arguments.length === 0
-  );
+export default function WeakMap(iterable) {
+  if (!(this instanceof WeakMap)) {
+    throw new TypeError(`Constructor WeakMap requires 'new'`);
+  }
 
   this._id = GUID_KEY + (id++);
+
+  if (iterable === null || iterable === undefined) {
+    return;
+  } else if (Array.isArray(iterable)) {
+    for (let i = 0; i < iterable.length; i++) {
+      let [key, value] = iterable[i];
+      this.set(key, value);
+    }
+  } else {
+    throw new TypeError('The weak map constructor polyfill only supports an array argument');
+  }
 }
 
 /*
@@ -35,6 +50,10 @@ export default function WeakMap() {
  * @return {Any} stored value
  */
 WeakMap.prototype.get = function(obj) {
+  if (!isObject(obj)) {
+    return undefined;
+  }
+
   let meta = peekMeta(obj);
   if (meta) {
     let map = meta.readableWeak();
@@ -55,10 +74,9 @@ WeakMap.prototype.get = function(obj) {
  * @return {WeakMap} the weak map
  */
 WeakMap.prototype.set = function(obj, value) {
-  assert(
-    'Uncaught TypeError: Invalid value used as weak map key',
-    obj && (typeof obj === 'object' || typeof obj === 'function')
-  );
+  if (!isObject(obj)) {
+    throw new TypeError('Invalid value used as weak map key');
+  }
 
   if (value === undefined) {
     value = UNDEFINED;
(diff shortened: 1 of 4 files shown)


rwjblue:
The feature flagging looks good. I have a few questions:

* Can we warn (perhaps only in dev mode) if the weakmap is destroyed while it still contains keys? Seems like this would help address the main caveat above...
* Can you add class documentation (with @public) to the class, and use '@category ember-metal-weakmap'? This will be stripped from the built API docs until the feature has a 'true' value (done via a YUIDoc preparser).

packages/ember-metal/tests/weak_map_test.js
```diff
@@ -47,20 +47,33 @@ QUnit.test('has weakMap like qualities', function(assert) {
...
-  expectAssertion(function() {
-    new WeakMap([[{}, 1]]);
-  }, /Invoking the WeakMap constructor with arguments is not supported at this time/);
+QUnit.test('constructing a WeakMap with an invalid iterator throws an error', function(assert) {
+  let expectedError = new TypeError('Invalid iterable used to initialize weak map');
+
+  assert.throws(() => { new WeakMap({ a: 1 }); }, expectedError);
+});
+
+QUnit.test('constructing a WeakMap with a valid invalid iterator inserts the entries', function(assert) {
```
mmun: typo: valid invalid

packages/ember-metal/lib/weak_map.js
```diff
@@ -20,13 +24,19 @@ function UNDEFINED() {}
...
-export default function WeakMap() {
-  assert(
-    'Invoking the WeakMap constructor with arguments is not supported at this time',
-    arguments.length === 0
-  );
-
+export default function WeakMap(iterable) {
   this._id = GUID_KEY + (id++);
+
+  if (iterable === null || iterable === undefined) {
```
stefanpenner: `arguments.length < 1` maybe?

packages/ember-metal/lib/weak_map.js
```diff
@@ -20,13 +24,19 @@ function UNDEFINED() {}
...
+    return;
+  } else if (Array.isArray(iterable)) {
```
stefanpenner: @thoov ^

packages/ember-metal/lib/weak_map.js
```diff
@@ -20,13 +24,19 @@ function UNDEFINED() {}
...
+    for (let i = 0; i < iterable.length; i++) {
+      let entry = iterable[i];
+      this.set(entry[0], entry[1]);
+    }
+  } else {
+    throw new TypeError('Invalid iterable used to initialize weak map');
```
stefanpenner: Well, es6 Iterables are valid but not supported. I am unsure if this messaging is quite correct.

packages/ember-metal/lib/weak_map.js
mmun: was just following the spec here. `new WeakMap(null | undefined) === new WeakMap()`

packages/ember-metal/lib/weak_map.js
mmun: Noted, I'll change it.

packages/ember-metal/lib/weak_map.js
stefanpenner: ok

mmun:
@rwjblue Thanks for the `@category` tip! I don't think we can implement any reasonable form of warnings against memory leaks. The weakmap itself does not know what keys are "inside" of it (because they are stored remotely in the keys' meta) so even if we added a `destroy()` method (which would also violate the polyfill interface) it wouldn't be able to release the references to the weakmap values stored in the keys' meta.

packages/ember-metal/lib/weak_map.js
thoov: This should be fine and will cover the majority use case for iterables. A minor readability improvement could be:

```js
for (let i = 0; i < iterable.length; i++) {
  const [key, value] = iterable[i];
  this.set(key, value);
}
```

thoov: I will also update https://github.com/thoov/ember-weakmap to match this once it gets merged

mmun: Ok!

mmun:
I didn't bother supporting iterables in the polyfill, because a) they're pretty rare still and b) if you're in an environment that has Symbols it probably has a native WeakMap too.

mmun:
@stefanpenner Do you think it's time yet to attempt to use the native WeakMap when it is available?

packages/ember-metal/lib/weak_map.js
```diff
@@ -8,8 +7,14 @@ import {
...
 function UNDEFINED() {}
 
+// Returns whether Type(value) is Object according to the terminology in the spec
+function isObject(value) {
+  return (typeof value === 'object' && value !== null) || typeof value === 'function';
+}
+
 /*
- * @private
+ * @public
```
rwjblue: I think the YUIDoc commands have to be after the description, can you move these down a bit?

mmun: :+1:

stefanpenner:
> @stefanpenner Do you think it's time yet to attempt to use the native WeakMap when it is available?

some thoughts:

* WeakMap will work with frozen objects, meta wont
* WeakMap has full iterable support (we don't yet)
* WeakMap performance should be evaluated across sufficient number of devcies

rwjblue:
TODO note for @mmun from https://github.com/emberjs/ember.js/pull/13709#issuecomment-226966429:

* Update to require `new`

rwjblue:
@mmun - Needs a rebase, and the tweaks mentioned in https://github.com/emberjs/ember.js/pull/13688#issuecomment-226967430.

mmun:
@rwjblue Done and done.

(cc @thoov)

rwjblue:
Awesome, thank you!

PR #13694, opened 2016-06-16
[Infra] Refactor test infra to use injections

chadhietala:
We should move away from hardcoded the setup of tests and follow the DI pattern. /cc @krisselden


rwjblue:
LGTM

PR #13699, opened 2016-06-17
Revert "[BUGFIX beta] Don't use prototype extensions (arrayContent{Did,Will}Change, removeAt)"

rwjblue:
Reverts emberjs/ember.js#13534.

After discussing @krisselden I realized that I merged #13534 prematurely (I am sorry about that).  There are still a few things that need to be addressed there to allow us to drop `Ember.A` usage and those need to be addressed before landing again.  

/cc @btecu

### DIFF
--- a/packages/ember-glimmer/tests/integration/helpers/element-action-test.js
+++ b/packages/ember-glimmer/tests/integration/helpers/element-action-test.js
@@ -5,7 +5,6 @@ import { set } from 'ember-metal/property_set';
 
 import EmberObject from 'ember-runtime/system/object';
 import { A as emberA } from 'ember-runtime/system/native_array';
-import { removeAt } from 'ember-runtime/mixins/mutable_array';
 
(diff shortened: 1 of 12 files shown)


rwjblue:
@krisselden was going to take a second look, and confirm before the revert...

krisselden:
We can leave it, this doesn't expose the API as public, there is a lot that needs to be done before standalone API is made public if the end goal is to migrate off of prototype extensions, it would be a big mistake to expose the standalone methods public until they are more the equivalent to `Ember.set`

We need an actual plan, my guess is that we'd deprecate Ember.A without the ENV.EXTEND_PROTOTYPES.Array (you wouldn't need it for observability and you should use standalone methods for mutating), and with ENV.EXTEND_PROTOTYPES.Array on we'd deprecate it being true, but we can't do this until the standalone methods trigger a '[]'. I would like to remove the array and enumerable observer altogether.

PR #13702, opened 2016-06-17
Update QUnit test adapter to be 2.0 compatible

### LINKED ISSUE #13696: Update Test adapter to support QUnit 2.0
QUnit 2.0 was released today, yay! But, unfortunately, Ember's [test adapter for QUnit currently relies on `QUnit.stop`](https://github.com/emberjs/ember.js/blob/master/packages/ember-testing/lib/adapters/qunit.js#L15) which has been removed. The [recommended upgrade path is to use `assert.async()`](http://qunitjs.com/upgrade-guide-2.x/#replace-stop-and-start-with-assert-async) which works for most cases, but since the `asyncStart` of the test adapter is invoked behind an abstraction, passing through the `assert` object might be a bit difficult.

~~Opening this up for some discussion on thoughts to get around this. I have a few thoughts but they all seem too hacky right now.~~ I believe I have a solid solution, will open a PR soonish.

trentmwillis:
An attempt to fix #13696.

The good: it works! There are some other things to be cleaned up in the ecosystem (such as the use of `setup`/`teardown`), but this makes all the code in Ember itself QUnit 2.0 compatible.

The bad: it relies on some private QUnit knowledge. Without some revamping of how tests are run though I don't see much way around that.

### DIFF
--- a/packages/ember-testing/lib/adapters/qunit.js
+++ b/packages/ember-testing/lib/adapters/qunit.js
@@ -11,11 +11,19 @@ import { inspect } from 'ember-metal/utils';
   @public
 */
 export default Adapter.extend({
+  done: [],
   asyncStart() {
-    QUnit.stop();
+    this.done.push(QUnit.config.current.assert.async());
   },
   asyncEnd() {
-    QUnit.start();
+    let current = QUnit.config.current;
+    this.done.pop()();
+
+    // If we're done with async and the user hasn't invoked async manually,
+    // then set usedAsync to false so that assertions can still happen
+    if (this.done.length === 0 && current.semaphore === 0) {
+      current.usedAsync = false;
+    }
   },
   exception(error) {
     ok(false, inspect(error));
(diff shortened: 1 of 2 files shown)


rwjblue:
Can you confirm what versions of QUnit are supported by this change?

rwjblue:
> The bad: it relies on some private QUnit knowledge. 

How likely is this to break in the future?  Seems somewhat like QUnit private API...

> Without some revamping of how tests are run though I don't see much way around that.

I'd love to set some time aside and ensure that the proposed testing revamp properly accounts for this. Lets circle around...

trentmwillis:
> Can you confirm what versions of QUnit are supported by this change?

Will look into this tomorrow or Monday probably.

> How likely is this to break in the future?

I don't know of anything upcoming that would break it and the existing code has been around awhile, so I would say unlikely.

> I'd love to set some time aside and ensure that the proposed testing revamp properly accounts for this. 

Definitely. As a bit of perspective, I think the crux of the matter is that QUnit assumes you know when your test is async and that you shouldn't be `done` with async until everything has been finished. Whereas Ember seems to treat a test as having async "phases" (that start and stop) rather than the whole thing being async.

trentmwillis:
After discussing in person, going to close this in favor of defining a solution in ember-qunit.

PR #13703, opened 2016-06-17
Fix build for 2.6.0 release branch

kratiahuja:
The release versions of `babel-plugin-feature-flags` seem to be in a messed up state. 0.2.1 release contains a [fix](https://github.com/ember-cli/babel-plugin-feature-flags/commit/b120f9f905c42372cf661f8a2e5932b103b7a9d5) which ports to Babel 6 and should have gone in 0.3.0 release. This causes the build for v2.6.x to fail.

Per the readme of `babel-plugin-feature-flags`: "This plugin is for Babel 6. If you need to support Babel 5 use the 0.2.x releases.".

This PR locks the version of the plugin to 0.2.0.

PS: I did not find a `release-2-6` branch and figured 2.6 was released from this branch. So I did a diff against this branch. Build on master seems to work with Babel 6 so probably this needs to be ported to previous release branches too.

cc: @krisselden @chadhietala @rwjblue

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   },
   "devDependencies": {
     "aws-sdk": "~2.2.43",
-    "babel-plugin-feature-flags": "~0.2.0",
+    "babel-plugin-feature-flags": "0.2.0",
     "babel-plugin-filter-imports": "~0.2.0",
     "bower": "~1.7.7",
     "broccoli-stew": "^1.2.0",


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "ember",
   "license": "MIT",
-  "version": "2.6.0",
```
krisselden: we don't want to change this.

package.json
```diff
@@ -22,31 +22,28 @@
   },
   "devDependencies": {
     "aws-sdk": "~2.2.43",
-    "babel-plugin-feature-flags": "~0.2.0",
+    "babel-plugin-feature-flags": "0.2.0",
```
krisselden: isn't this the only thing you wanted to change?  The rest of the changes seem problematic for a release branch

package.json
```diff
@@ -22,31 +22,28 @@
...
-    "ember-cli-blueprint-test-helpers": "^0.11.0",
+    "ember-cli": "^2.6.1",
+    "ember-cli-blueprint-test-helpers": "^0.12.0",
     "ember-cli-dependency-checker": "^1.2.0",
-    "ember-cli-internal-test-helpers": "^0.7.0",
     "ember-cli-sauce": "^1.4.2",
     "ember-cli-yuidoc": "0.8.3",
     "ember-publisher": "0.0.7",
-    "emberjs-build": "0.6.4",
+    "emberjs-build": "0.10.0",
```
krisselden: this was pinned, not `^`

package.json
kratiahuja: Yes that is the only change I intended to do. But I did not find a `release-2-6` branch and looking at the logs 2.6.0 release was done from this branch it seems.

package.json
kratiahuja: I honestly didn't make this change. Not sure how it landed up with all of that.

rwjblue:
I believe the only change needed to fix usage of 0.2.1 is https://github.com/emberjs/ember.js/commit/29a1fc2114b28b8fa25ba4dd88e807ead1f581d9.

rwjblue:
But I think locking down makes more sense for an already released branch.

I'm 👍 on these changes.

kratiahuja:
@krisselden Fixed the mess. I reverted my previous commit and just pinned `babel-plugin-feature-flag`

kratiahuja:
@rwjblue : Yes that is correct. In order to pick 0.2.1, we need to remove those tests (that's the reason master build is passing).

package.json
rwjblue: Correct, we only push a `release-2-x` branch when it is no longer the current stable version.  `release` is the right target for this change.  Though I suspect that we should probably also fix `lts-2-4`...

kratiahuja: Aah ok good to know. Confirmed this broken on `lts-*` and `release-*` branches.

krisselden:
@rwjblue you disabled squashing, @kratiahuja can you squash this so it is one commit?

kratiahuja:
@krisselden Squashed.

rwjblue:
Yes, I had to disable it because the automation I have setup for the CHANGELOG doesn't work with the newer format squashed merges. I defintely want to fix it though.

rwjblue:
Thank you @kratiahuja!

kratiahuja:
No prob! Happy to help! :)

PR #13707, opened 2016-06-18
Remove JSHint directives from Mocha test blueprints

Turbo87:
The `expr` rule should instead be deactivated in `tests/.jshintrc`.

see also https://github.com/ember-cli/ember-cli-legacy-blueprints/pull/33


rwjblue:
Thank you @Turbo87!

PR #13709, opened 2016-06-18
[BUGFIX release] Fix to throw error for `Ember.Map()` without `new`

tricknotes:
The `this` keyword refers global context (such as `window`) inside function call.

``` js
this instanceof this.constructor // === window instanceof window.constructor
//=> true
```

This means that `Ember.Map()`(without `new`) contaminates global variables unexpectedly.

### DIFF
--- a/packages/ember-metal/lib/map.js
+++ b/packages/ember-metal/lib/map.js
@@ -240,13 +240,13 @@ OrderedSet.prototype = {
   @constructor
 */
 function Map() {
-  if (this instanceof this.constructor) {
+  if (this instanceof Map) {
     this._keys = OrderedSet.create();
     this._keys._silenceRemoveDeprecation = true;
     this._values = new EmptyObject();
     this.size = 0;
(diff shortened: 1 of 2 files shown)


packages/ember-metal/lib/map.js
```diff
@@ -240,13 +240,13 @@ OrderedSet.prototype = {
   @constructor
 */
 function Map() {
-  if (this instanceof this.constructor) {
+  if (this instanceof Map) {
```
rwjblue: Hmm. Doesn't use of this.constructor allow folks to subclass?

rwjblue: Never mind. I believe subclassing would also be fine with these changes.

rwjblue:
This seems good to me. 

@stefanpenner - r?

mmun:
I'll add this check to the weak map PR as well.

mmun:
r+

PR #13716, opened 2016-06-19
[BUGFIX beta] Bring back Ember.Test.waiters.

### LINKED ISSUE #13603: Expose Ember.Test.checkWaiters as public API
https://github.com/emberjs/ember.js/pull/13440 landed a big refactor of `Ember.Test`. Among the changes was to remove `Ember.Test.waiters` 💁  from the exposed API. `waiters` leaked the internal implementation of the waiter queue, and was not explicitly flagged as `@public`.

In updating `ember-test-helpers` @rwjblue and I found that `Ember.Test.waiters` is used by that library to support `wait()` ✋  in integration tests. See: https://github.com/switchfly/ember-test-helpers/blob/6207a7b6af9214ef47a79f52d9b240bb95e6a728/lib/ember-test-helpers/wait.js#L46

Though we're going to ship a hacky intimate-API workaround for the 2.7 beta (master at this writing), it seems like `checkWaiters` should be made public API to support this use case. Just something that was missed when `Ember.Test.waiters` was removed. Thoughts @krisselden?

And as a point of order, no, I didn't write an RFC as I do not believe this change would represent "substantial" change to Ember. IMO we're simply recovering from the mistake of dropping `waiters`, which was not documented as public but was actually used. ✌️

rwjblue:
Brings back `Ember.Test.waiters` with a deprecation (in favor of `Ember.Test.checkWaiters`).

Closes #13603.

/cc @krisselden @mixonic

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -58,3 +58,8 @@ to stay in line with ES standards (see [RFC](https://github.com/emberjs/rfcs/blo
   Ember.String.isHtmlSafe(plainString); // false
   Ember.String.isHtmlSafe(safeString);  // true
   ```
+
+* `ember-testing-check-waiters`
+
+Expose a simple mechanism for test tooling to determine if all foreign async has been
+handled before continueing the test. Replaces the intimate API `Ember.Test.waiters` (with a deprecation).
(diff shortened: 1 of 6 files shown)


rwjblue:
I updated this to feature flag the new public API, so that we can target fixing `Ember.Test.waiters` (via  getter) for 2.7.0-beta.2.

mixonic:
👍  this seems great @rwjblue

PR #13717, opened 2016-06-19
[glimmer2] fix unbound helper test

fivetanley:
Before, the test rendered the following:

``` handlebars
<div data-foo="{{if cond1 "T1" "F1"}}"></div>
<div data-foo="{{if cond2 "T2" "F2"}}"></div>
```

The test was only asserting against the value of the `data-foo` of the
first div, causing it fo fail because it was missing the value of the
`data-foo` in the second div.

This is because the test was using `wrappedTemplateFor`, which
is a convenience method which only expects _one_ template to be
rendered at a time.

Instead, we use `wrapperFor` and two `templateFor` calls to instead
create only _one_ div and assert against it. What's rendered in the test
now looks like:

``` handlebars
<div data-foo="{{if cond1 "T1" "F1"}}{{if cond2 "T2" "F2"}}"></div>
```

refs #13644

### DIFF
--- a/packages/ember-glimmer/tests/utils/shared-conditional-tests.js
+++ b/packages/ember-glimmer/tests/utils/shared-conditional-tests.js
@@ -494,12 +494,11 @@ export class TogglingHelperConditionalsTest extends TogglingConditionalsTest {
     this.render(wrappedTemplate, context);
   }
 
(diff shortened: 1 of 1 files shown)


rwjblue:
Thanks @fivetanley!

PR #13719, opened 2016-06-19
[CLEANUP] Compile/precompile should use same code path

chadhietala:


packages/ember-glimmer/tests/utils/helpers.js
```diff
@@ -1,13 +1,10 @@
...
-  defaultCompileOptions,
-  template
-} from 'ember-glimmer-template-compiler';
 import { setupApplicationRegistry, setupEngineRegistry } from 'ember-glimmer/setup-registry';
 import { default as _buildOwner } from 'container/tests/test-helpers/build-owner';
 
+export {
+  compile,
+  precompile
+} from 'ember-glimmer-template-compiler';
```
rwjblue: Should this be using `ember-glimmer-template-compiler/tests/utils/helpers` (seems like the `ember-htmlbars/tests/utils/helpers` version does use the `ember-htmlbars-template-compiler/tests/utils/helpers`).

rwjblue:
Thanks!

PR #13720, opened 2016-06-19
[FEATURE ember-string-ishtmlsafe] Prevent deprecation when not enabled.

rwjblue:
I missed this in review, but it seems that accessing `Ember.Handlebars.SafeString` was issuing a deprecation even with the `ember-string-ishtmlsafe` flag was not enabled.

/cc @workmanw for review


workmanw:
D'oh! My mistake. Yeap this looks good to me.

rwjblue:
Thanks for reviewing!

PR #13721, opened 2016-06-19
[glimmer2] fix #each with key option test

fivetanley:
In Ember, you can remove duplicates in an array by providing the `key`
option to the `{{#each}}` helper.

For example, if given the following `model` property:

``` javascript
[
  {
    "id": "1",
    "name": "Robert Jackson"
  },
  {
    "id": "1",
    "name": "Robert Jackson with diff name but same ID"
  },
  {
    "id": "2",
    "name": "Katie Gengler"
  }
]
```

And the following template:

``` handlebars
{{#each model key="id" as |person|
  {{person.name}}
{{/each}}
```

It would render the following:

``` html
Robert Jackson
Katie Gengler
```

Even though we have two objects with "Robert Jackson" as the name, they
share the same "id", so Ember will only render the first object it
encounters with that "id" while iterating. That's why "Robert Jackson"
shows up but not "Robert Jackson with diff name but same ID".

This test was adding in a new object (in our case `{text: ' '}`), but an
object with `{text: ' '}` already existed in the setup line. Because it
was cached by the same `'text'` value, it did not show up in the
subsequent rerender.

We change it to a different text value so we can assert it was added
correctly in the list.

refs #13644

### DIFF
--- a/packages/ember-glimmer/tests/integration/syntax/each-test.js
+++ b/packages/ember-glimmer/tests/integration/syntax/each-test.js
@@ -303,7 +303,7 @@ moduleFor('Syntax test: {{#each as}}', class extends EachTest {
     this.assertInvariants();
   }
 
-  [`@htmlbars it maintains DOM stability for stable keys when list is updated`]() {
+  [`@test it maintains DOM stability for stable keys when list is updated`]() {
     this.render(`{{#each list key="text" as |item|}}{{item.text}}{{/each}}`, {
       list: emberA([{ text: 'Hello' }, { text: ' ' }, { text: 'world' }])
     });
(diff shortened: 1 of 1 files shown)


rwjblue:
This seems like a good fix for this test, but we definitely need to fix the duplicate key case in general.  There are a few "duplicate keys with each" tests, can you confirm?

chancancode:
:+1: seems fine to fix this particular test, but also :+1: to fixing the collision case in general (and make sure we have tests for those).

The plan is to track collision inside the iterator on the ember side, and fallback to using a guid (or whatever we were doing before). Do we have tests for not specifying a key? I believe that should use the guid as well.

fivetanley:
> to fixing the collision case in general (and make sure we have tests for those).

tests: https://github.com/emberjs/ember.js/pull/13722

> Do we have tests for not specifying a key?

Yeah. 

[Objects](https://github.com/emberjs/ember.js/blob/da56d7a0b75caad63c8326fb070f5a32f5c7fd52/packages/ember-glimmer/tests/integration/syntax/each-test.js#L254)
[Primitives](https://github.com/emberjs/ember.js/blob/da56d7a0b75caad63c8326fb070f5a32f5c7fd52/packages/ember-glimmer/tests/integration/syntax/each-test.js#L232)

rwjblue:
@chancancode - Using the item being iterateds guid isn't enough, unless you are saying that we can never iterate over an array that contains the same object.  The same assumption was made during initial Ember 1.13 work, and was fixed by @krisselden in https://github.com/tildeio/htmlbars/pull/393.  For example:

```js
let obj = { foo: 'bar' };
let array = [ obj, obj, obj ];
```

It should definitely be possible to `{{#each ` over that array...

chancancode:
Yeah, Kris told me about it but I forgot the details. s/guid/the Kris algorithm/

PR #13723, opened 2016-06-20
Remove obsolete gemspec file and other ruby files

Turbo87:
I'm assuming these are now longer used since the `Rakefile` is gone.

/cc @locks


rwjblue:
We still use these files to publish to rubygems https://rubygems.org/gems/ember-source

Turbo87:
ah, okay. wasn't aware of that...

PR #13724, opened 2016-06-20
[Glimmer 2] Support classNameBindings from the call site

chadhietala:
This brings back the build time transforms for `classNameBindings` and `classBindings`.

### DIFF
--- a/packages/ember/tests/helpers/link_to_test.js
+++ b/packages/ember/tests/helpers/link_to_test.js
@@ -76,8 +76,6 @@ function sharedTeardown() {
   reset();
 }
 
-import { test } from 'internal-test-helpers/tests/skip-if-glimmer';
-
 QUnit.module('The {{link-to}} helper', {
   setup() {
     run(() => {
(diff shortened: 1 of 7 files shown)


packages/ember-glimmer/lib/helpers/-ctx.js
```diff
@@ -0,0 +1,14 @@
+import { InternalHelperReference } from '../utils/references';
+
+export const CTX = 'j0850347-25h2-47q8-bb63-c405402239';
+
+function ctx({ positional }) {
+  return positional.at(0).value();
```
chadhietala: I'm wondering if I should just build new `ArgSyntax`.

packages/ember/tests/helpers/link_to_test.js
```diff
@@ -76,7 +76,7 @@ function sharedTeardown() {
   reset();
 }
 
-import { test } from 'internal-test-helpers/tests/skip-if-glimmer';
+// import { test } from 'internal-test-helpers/tests/skip-if-glimmer';
```
rwjblue: Should likely delete this

PR #13725, opened 2016-06-20
Bind the watchModelTypes's release function to `this` so it correctly…

bmac:
… removes the array observers

Not sure what the best way to test this is since it's tied to the cleanup of the app.

### DIFF
--- a/packages/ember-extension-support/lib/data_adapter.js
+++ b/packages/ember-extension-support/lib/data_adapter.js
@@ -302,9 +302,7 @@ export default EmberObject.extend({
 
     addArrayObserver(records, this, observer);
 
-    function release() {
-      removeArrayObserver(records, this, observer);
-    }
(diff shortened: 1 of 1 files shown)


btecu:
Could it be just `let release = () => removeArrayObserver(records, this, observer);`?

rwjblue:
@btecu - Yes, that would work properly, but we do not care about the return value.

packages/ember-extension-support/lib/data_adapter.js
```diff
@@ -302,7 +302,7 @@ export default EmberObject.extend({
 
     addArrayObserver(records, this, observer);
 
-    function release() {
+    let release = () => {
       removeArrayObserver(records, this, observer);
     }
```
rwjblue: Missing semi-colon here

rwjblue:
LGTM

PR #13726, opened 2016-06-21
[GLIMMER2] Click action links instead of touching action manager internals

zackthehuman:
This should fix the following tests listed in #13644:
-  Events are triggered on the controller if a matching action name is implemented
-  Events are triggered on the current state when defined in actions object
-  Events defined in actions object are triggered on the current state when routes are nested
-  Actions are not triggered on the controller if a matching action name is implemented as a method
-  actions can be triggered with multiple arguments

The `ActionManager` internals differ between HTMLBars and Glimmer2, and the tests were written in a way that relied on the implementation details. This change simplifies things by doing what a user would do: _click on the thing_.


chadhietala:
LGTM and actually makes the tests less about the "how".

rwjblue:
Thanks @zackthehuman!

PR #13728, opened 2016-06-21
Improved warning message for unescaped style binding

IgorKvasn:
Sometimes it is really difficult to track down the cause of "unescaped style binding" warning (especially in some legacy application). I guess it would be nice to give people some hints where to find the problem itself. So I changed the warning message to

```
WARNING: Binding style attributes may introduce cross-site scripting vulnerabilities;
please ensure that values being bound are properly escaped. 
For more information,including how to disable this warning, 
seehttp://emberjs.com/deprecations/v1.x/#toc_binding-style-attributes. 
Style affected: "background-color: red; width:100px; height: 100px;"
```

btw. I am really not sure how to prefix the issue tittle - my guess is [Feature] ...


rwjblue:
This is great!  I have wanted a better message here also, thanks for working on it.  

One thing that struck me, is that it may be nice to have some more context than even just the `style` value being set.  We could tweak your changes here to pass in `this.element` (`createStyleWarning(this.element, value)`) and potentially print more info from the element itself...

btecu:
Also needs space after comma `information,including` and before the url `seehttp://emberjs.com`.

IgorKvasn:
@rwjblue yes, you are right - `style` is not enough - I tried to print `element` itself, but the problem is that `element.innerHtml()` returns `<!----->` (to be specific, I call `morph.element.innerHtml()`) and `element.toString()` returns a unhelpful `[object HTMLDivElement]`.
It would be nice to print at least element ID, but not every element has ID specified

@btecu thanks

packages/ember-htmlbars/lib/morphs/attr-morph.js
```diff
@@ -37,6 +31,13 @@ function deprecateEscapedStyle(morph, value) {
   );
 }
 
+export function createStyleWarning(value) {
+  return 'Binding style attributes may introduce cross-site scripting vulnerabilities;' +
```
btecu: You need a space at the end of the lines (35, 36 & 37).

rwjblue:
@IgorKvasn - Maybe we can use `elem.outerHTML` or even something like:

```js
`<${elem.tagName} id=${elem.id}>` 
```

locks:
Hi @IgorKvasn, sorry for the delay :\ Are you still able to push this to completion? From the discussion above we seem to be close.

IgorKvasn:
Hi @locks , I'll have a look at it during the weekend - I have to adapt it to the new Glimmer

IgorKvasn:
hmm looks like I messed up merging from upstream which resulted in 904 files changed - @locks, I will close this PR and create a new one

PR #13729, opened 2016-06-21
[BUGFIX beta] Remove unreferenced function from `ember-metal/replace`

tricknotes:
This function became unnecessary with the followings:
- Ember.ArrayController: 9899dffe
- Ember.Select: 4b5b9729

I think we can remove this function due to un-exported.
If are there some fears, should it be shown deprecation warning?

### DIFF
--- a/packages/ember-metal/lib/replace.js
+++ b/packages/ember-metal/lib/replace.js
@@ -1,6 +1,6 @@
 let { splice } = Array.prototype;
 
-export function _replace(array, idx, amt, objects) {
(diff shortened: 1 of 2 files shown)


rwjblue:
I wondered if we needed `_replace` also, but it looks like the `_replace` function is still being used in `ember-runtime/system/native_array`. 

Maybe we should just drop the underscore, and make it the default export in this file?

tricknotes:
Completely agreed 👍 
I updated my commit.

rwjblue:
Thanks again!

PR #13731, opened 2016-06-21
[CLEANUP beta] Remove unreferenced module 'ember-runtime/core'

tricknotes:
It is unnecessary since 149c4f41.


rwjblue:
Awesome, thank you!

PR #13733, opened 2016-06-21
[FEATURE ember-application-engines] Enable routable engines.

dgeb:
Moves routing-related code upstream from [ember-engines](https://github.com/dgeb/ember-engines).

Extends `Route`, `Router`, `Outlet`, and `DSL` classes to support routable engines.

Changes are all behind the `ember-application-engines` feature flag.


dgeb:
@rwjblue r?

rwjblue:
Thanks @dgeb!

PR #13736, opened 2016-06-21
[Glimmer2] Cleanup teardown

chadhietala:
@wycats or @chancancode this probably requires your 👀 . I may be wrong but it appears nothing special needs to be done for `destroy` in glimmer, so moved the `destroy` from the `view_support` mixin into the HTMLBars component.


packages/ember-glimmer/lib/component.js
```diff
@@ -117,6 +117,30 @@ const Component = CoreView.extend(
...
+    destroy() {
+      // get parentView before calling super because it'll be destroyed
+      let parentView = this.parentView;
+      let viewName = this.viewName;
+
+      if (!this._super(...arguments)) { return; }
+
+      // remove from non-virtual parent view if viewName was specified
+      if (viewName && parentView) {
+        parentView.set(viewName, null);
```
rwjblue: I am not sure we actually support this anymore.

@krisselden - What do you think?

PR #13739, opened 2016-06-22
[WIP][Glimmer2] Move renderNode refs in view_support

chadhietala:
/cc @rwjblue

### DIFF
--- a/packages/ember-glimmer/lib/component.js
+++ b/packages/ember-glimmer/lib/component.js
@@ -117,6 +117,11 @@ const Component = CoreView.extend(
       }
     },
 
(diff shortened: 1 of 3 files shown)


rwjblue:
Seems good, but we should include at least an empty function in the glimmer component case (because `Ember.TextSupport` calls `this.readDOMAttr()` and it is a public API).

chadhietala:
@rwjblue This doesn't fix a bunch of tests because a lot of the tests depend on `sendAction` at this point which we need figure out in #13532.

rwjblue:
Gotcha

PR #13741, opened 2016-06-22
Fix "super() use outside of constructor" issues

Turbo87:
Babel 6 flags these as SyntaxError

/cc @rwjblue @stefanpenner


rwjblue:
:+1:

PR #13744, opened 2016-06-23
[Docs] Document how a waiter works

knownasilya:
A simple explanation of how a waiter works would help anyone looking to write one.

### DIFF
--- a/packages/ember-testing/lib/test/waiters.js
+++ b/packages/ember-testing/lib/test/waiters.js
@@ -8,7 +8,9 @@ const callbacks = [];
    This allows ember-testing to play nicely with other asynchronous
    events, such as an application that is waiting for a CSS3
    transition or an IndexDB transaction. The waiter runs periodically
-   until the returning result is truthy.
+   after all async helpers (i.e. `click`, `andThen`, `visit`, etc) have executed,
+   until the returning result is truthy. After all waiters finish, the next batch of async helpers
+   are executed and the process repeats.
 
    For example:
 
--- a/packages/ember-testing/lib/test/waiters.js
+++ b/packages/ember-testing/lib/test/waiters.js
@@ -7,7 +7,8 @@ const callbacks = [];
 /**
    This allows ember-testing to play nicely with other asynchronous
    events, such as an application that is waiting for a CSS3
-   transition or an IndexDB transaction.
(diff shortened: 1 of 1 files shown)


packages/ember-testing/lib/test/waiters.js
```diff
@@ -7,7 +7,8 @@ const callbacks = [];
 /**
    This allows ember-testing to play nicely with other asynchronous
    events, such as an application that is waiting for a CSS3
-   transition or an IndexDB transaction.
+   transition or an IndexDB transaction. The waiter runs periodically
```
rwjblue: I would slightly clarify that it is ran periodically after async helpers (i.e. `click`, `andThen`, `visit`, etc) have executed prior to executing the next async step.

knownasilya: Got it.

Do all waiters just run or do you use `app.testHelpers.wait()` in your custom asyncHelper to do a wait check?

rwjblue: The work is done by way of `wait()`, but all async helpers run `app.testHelpers.wait()` at the end by default (you shouldn't have to `return app.testHelpers.wait()` manually in an async test helper).

packages/ember-testing/lib/test/waiters.js
```diff
@@ -8,7 +8,9 @@ const callbacks = [];
    This allows ember-testing to play nicely with other asynchronous
    events, such as an application that is waiting for a CSS3
    transition or an IndexDB transaction. The waiter runs periodically
-   until the returning result is truthy.
+   after all async helpers (i.e. `click`, `andThen`, `visit`, etc) have executed,
```
rwjblue: (sorry) `after all` -> `after each` 😺

packages/ember-testing/lib/test/waiters.js
knownasilya: Updated

locks:
@rwjblue is this good to go?

rwjblue:
Yes, this is good, just needs a squashing now...

locks:
@rwjblue is there a problem with using GitHub interface's squashing mechanism? It makes something weird to the git history, right?

rwjblue:
It breaks my CHANGELOG generation script so I disabled it until we can get it fixed.

PR #13745, opened 2016-06-23
[Glimmer2] sendAction

chadhietala:
This supersedes #13532 and sets up the action "scoping" setting similar to how HTMLBars does it. There are couple tasks that will quickly follow once this has landed:
- Unskip all of the glimmer tests that depend on this functionality
- Delete some of the old action tests
- Potentially consolidate some the tests in this PR as the implementation is roughly the same
- More cleanup around target action (currently there are like 3 things that handle actions)

Part of #13644

### DIFF
--- a/packages/ember-glimmer/lib/component.js
+++ b/packages/ember-glimmer/lib/component.js
@@ -5,6 +5,8 @@ import ViewStateSupport from 'ember-views/mixins/view_state_support';
 import InstrumentationSupport from 'ember-views/mixins/instrumentation_support';
 import AriaRoleSupport from 'ember-views/mixins/aria_role_support';
 import ViewMixin from 'ember-views/mixins/view_support';
+import ActionSupport from 'ember-views/mixins/action_support';
+import TargetActionSupport from 'ember-runtime/mixins/target_action_support';
 import EmberView from 'ember-views/views/view';
 import symbol from 'ember-metal/symbol';
 import EmptyObject from 'ember-metal/empty_object';
@@ -36,10 +38,14 @@ const Component = CoreView.extend(
   ClassNamesSupport,
   InstrumentationSupport,
   AriaRoleSupport,
+  TargetActionSupport,
+  ActionSupport,
   ViewMixin, {
     isComponent: true,
     layoutName: null,
     layout: null,
+    controller: null,
+    _controller: null,
 
     init() {
       this._super(...arguments);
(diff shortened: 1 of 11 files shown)


packages/ember-glimmer/lib/renderer.js
```diff
@@ -132,7 +132,7 @@ class Renderer {
     let self = new RootReference(view);
     let dynamicScope = new DynamicScope({
       view,
-      controller: view.outletState.render.controller,
+      controller: view.outletState.outlets.main.render.controller,
```
chadhietala: This seems correct based on this? https://github.com/emberjs/ember.js/blob/master/packages/ember-htmlbars/lib/keywords/outlet.js#L95

packages/ember-glimmer/lib/syntax/curly-component.js
```diff
@@ -37,6 +37,15 @@ function applyAttributeBindings(attributeBindings, component, operations) {
   }
 }
 
+// Use `_targetObject` to avoid stomping on a CP
+// that exists in the component
+function privatizeTargetObject(props) {
```
chadhietala: Need to remove this.

packages/ember-glimmer/lib/syntax/curly-component.js
```diff
@@ -78,6 +87,13 @@ class CurlyComponentManager {
     dynamicScope.view = component;
     parentView.appendChild(component);
 
+    if (dynamicScope.controller) {
```
chadhietala: Is this correct?

packages/ember-glimmer/lib/component.js
```diff
@@ -29,17 +31,67 @@ export const ROOT_REF = symbol('ROOT_REF');
 export const REFS = symbol('REFS');
 export const IS_DISPATCHING_ATTRS = symbol('IS_DISPATCHING_ATTRS');
 export const HAS_BLOCK = symbol('HAS_BLOCK');
+import isNone from 'ember-metal/is_none';
+
+
+import { MUTABLE_CELL } from './utils/process-args';
+
+function validateAction(component, actionName) {
```
chadhietala: Need to share this from htmlbars.

packages/ember-glimmer/lib/component.js
```diff
@@ -29,17 +31,67 @@ export const ROOT_REF = symbol('ROOT_REF');
...
   AriaRoleSupport,
+  TargetActionSupport,
   ViewMixin, {
     isComponent: true,
     layoutName: null,
     layout: null,
+    controller: null,
+    _controller: null,
+
+    sendAction(action, ...contexts) {
```
chadhietala: Need to share this from HTMLBars.

rwjblue:
Awesome work @Joelkang and @chadhietala!

PR #13749, opened 2016-06-24
[GLIMMER] Enable passing tests.

rwjblue:
These tests were unlocked by fixing `{{input}}` and `sendAction`, re-enable them.

### DIFF
--- a/packages/ember-testing/tests/acceptance_test.js
+++ b/packages/ember-testing/tests/acceptance_test.js
@@ -13,8 +13,6 @@ import { setTemplates, set as setTemplate } from 'ember-templates/template_regis
 
 var App, find, click, fillIn, currentRoute, currentURL, visit, originalAdapter, andThen, indexHitCount;
 
-import { test } from 'internal-test-helpers/tests/skip-if-glimmer';
-
 QUnit.module('ember-testing Acceptance', {
   setup() {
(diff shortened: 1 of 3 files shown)


rwjblue:
Hmm.  This is annoying. These are passing in Chrome, but not phantom it seems.  I will have to dig into it later.

If someone else wants to take a crack at it, feel free....

chadhietala:
I can't repro this locally so I dumped the cache and restarted the failing CI jobs.

chadhietala:
So after some investigation this appears to be runloop related. To fix the all the failing tests related to `fillIn` I modified like so

```diff
diff --git a/packages/ember-testing/lib/helpers/fill_in.js b/packages/ember-testing/lib/helpers/fill_in.js
index 6d7fe72..13e7f69 100644
--- a/packages/ember-testing/lib/helpers/fill_in.js
+++ b/packages/ember-testing/lib/helpers/fill_in.js
@@ -16,5 +16,8 @@ export default function fillIn(app, selector, contextOrText, text) {
     fireEvent(el, 'input');
     fireEvent(el, 'change');
   });
+  run(() => {
+    $el.val(text);
+  });
   return app.testHelpers.wait();
 }
```

Once I did that I had 2 other test failures and I believe they fail because DOM from other tests are not being torn down. When the tests are ran on their own they pass.

chadhietala:
This been superseded by #13837 #13835

PR #13751, opened 2016-06-26
Ensure top level destroys DOM

rwjblue:
Prior to this change, the render nodes are cleared (replaced with `<!---->`) but not removed from the DOM

/cc @krisselden


rwjblue:
Discussed with @krisselden, and we need to ensure that this isn't running for things other than top level view.

krisselden:
This should be fixed by #13775

PR #13752, opened 2016-06-26
Fixes #13750 update documentation to cast `getJSON` to compliant promise

### LINKED ISSUE #13750: Issues with PromiseProxyMixin: `Uncaught [object Object]` and `.catch is not a function`
Using this example from the guides, I would expect the error to be caught as `reason` (since "/some/remote/data.json" is obviously invalid):

```
var ObjectPromiseProxy = Ember.ObjectProxy.extend(Ember.PromiseProxyMixin);

var proxy = ObjectPromiseProxy.create({
  promise: Ember.$.getJSON('/some/remote/data.json')
});

proxy.then(function(json) {
  // the json
}, function(reason) {
  // the reason why you have no json
});
```

Instead, the full error message is printed to the console, as well as `Uncaught [object Object]`. Is this documentation out of date, or is there an issue with `PromiseProxyMixin`?

Another related issue is that using the following code instead yields the error `Uncaught TypeError: proxy.then(...).catch is not a function`:
```
proxy.then(function(json) {
  // the json
}).catch(function (reason) {
  // the reason why you have no json
});
```

Guide for reference:
http://emberjs.com/api/classes/Ember.PromiseProxyMixin.html

I'm using Ember 2.6.0, and encountered the same issues at https://ember-twiddle.com/

habdelra:
This is from Wicked Good Ember contributing workshop in row #2 https://docs.google.com/spreadsheets/d/12lAjN5ro2UO4j30FbVtfCQgMgJaywSow8mA2OEbIIV0/edit#gid=0

### DIFF
--- a/packages/ember-runtime/lib/mixins/promise_proxy.js
+++ b/packages/ember-runtime/lib/mixins/promise_proxy.js
@@ -38,7 +38,7 @@ function tap(proxy, promise) {
   let ObjectPromiseProxy = Ember.ObjectProxy.extend(Ember.PromiseProxyMixin);
 
   let proxy = ObjectPromiseProxy.create({
-    promise: $.getJSON('/some/remote/data.json')
+    promise: Ember.RSVP.cast($.getJSON('/some/remote/data.json'))
   });
 
   proxy.then(function(json){
@@ -60,6 +60,9 @@ function tap(proxy, promise) {
 
   When the $.getJSON completes, and the promise is fulfilled
   with json, the life cycle attributes will update accordingly.
+  Note that $.getJSON doesn't return an ECMA specified promise,
+  it is useful to wrap this with an `RSVP.cast` so that it behaves
+  as a spec compliant promise.
 
   ```javascript
   proxy.get('isPending')   //=> false


habdelra:
Note: here's a twiddle showing that this approach works https://ember-twiddle.com/55ae031da0bfe8647440410ce9016d67

packages/ember-runtime/lib/mixins/promise_proxy.js
```diff
@@ -60,6 +60,9 @@ function tap(proxy, promise) {
 
   When the $.getJSON completes, and the promise is fulfilled
   with json, the life cycle attributes will update accordingly.
+  Note that $.getJSON doesn't return an ECMA spcified promise,
```
rwjblue: Small typo here: s/spcified/specified/

habdelra:
@rwjblue needs your eyes

rwjblue:
@habdelra - 👓

rwjblue:
Thank you!

packages/ember-runtime/lib/mixins/promise_proxy.js
```diff
@@ -38,7 +38,7 @@ function tap(proxy, promise) {
   let ObjectPromiseProxy = Ember.ObjectProxy.extend(Ember.PromiseProxyMixin);
 
   let proxy = ObjectPromiseProxy.create({
-    promise: $.getJSON('/some/remote/data.json')
+    promise: Ember.RSVP.cast($.getJSON('/some/remote/data.json'))
```
sukima: `Ember.RSVP.cast` is actually undocumented and [in the source](https://github.com/tildeio/rsvp.js/blob/5727073c53c207bb687b82a1683eb3c1b332960b/lib/rsvp.js#L26) seems to be an alias to `resolve` Maybe _resolve_ is a better choice since it is documented and used in the guides.

Issue #13397, opened 2016-04-22
Should the click helper's 2nd argument be documented?


sethbrasile:
Looks like the `click` helper

http://emberjs.com/api/classes/Ember.Test.html#method_click
https://guides.emberjs.com/v2.5.0/testing/acceptance/#toc_asynchronous-helpers

Accepts `context` as a second arg.

The folks over on the guides repo are curious if the folks over here believe that it should be documented.

https://github.com/emberjs/guides/issues/1386

mixonic:
Seems like the API documentation should definitely be updated. Regarding the guides we can defer to @locks @acorncom about if they think the content is appropriate.

pixelhandler:
@mixonic @sethbrasile I belive the `context` (2nd) argument is due to using jQuery selectors, e.g. http://api.jquery.com/jQuery/#jQuery1 So aside from the client helper wouldn't the other test helpers also use this `context` too?

pixelhandler:
@locks the guides show a context arguement for the `find` helper but the API docs do not, http://emberjs.com/api/classes/Ember.Test.html#method_find and https://guides.emberjs.com/v2.5.0/testing/acceptance/#toc_synchronous-helpers see `find(selector, context)` so perhaps we should have `click(selector, context)` as well and the API docs + guides should agree.

PR #13753, opened 2016-06-27
[DOC] Issue #13397, update documentation for click helper

faizaanshamsi:
Update documentation to mention the `click` test helper takes a `context` argument.

Fixes https://github.com/emberjs/ember.js/issues/13397

### DIFF
--- a/packages/ember-testing/lib/helpers.js
+++ b/packages/ember-testing/lib/helpers.js
@@ -268,18 +268,19 @@ asyncHelper('visit', visit);
 
 /**
   Clicks an element and triggers any actions triggered by the element's `click`
-  event.
+  event. Accepts `context` as an optional second argument.
 
   Example:
 
   ```javascript
-  click('.some-jQuery-selector').then(function() {
+  click('.some-jQuery-selector', 'body').then(function() {
     // assert something
   });
   ```
 
   @method click
   @param {String} selector jQuery selector for finding element on the DOM
+  @param {String} context for finding element on the DOM
   @return {RSVP.Promise}
   @public
 */
--- a/packages/ember-testing/lib/helpers.js
+++ b/packages/ember-testing/lib/helpers.js
@@ -268,19 +268,27 @@ asyncHelper('visit', visit);
 
 /**
   Clicks an element and triggers any actions triggered by the element's `click`
-  event. Accepts `context` as an optional second argument.
+  event.
 
   Example:
 
   ```javascript
-  click('.some-jQuery-selector', 'body').then(function() {
+  click('.some-jQuery-selector').then(function() {
+    // assert something
+  });
+  ```
+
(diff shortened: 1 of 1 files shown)


packages/ember-testing/lib/helpers.js
```diff
@@ -268,18 +268,19 @@ asyncHelper('visit', visit);
...
   ```javascript
-  click('.some-jQuery-selector').then(function() {
+  click('.some-jQuery-selector', 'body').then(function() {
     // assert something
   });
   ```
 
   @method click
   @param {String} selector jQuery selector for finding element on the DOM
+  @param {String} context for finding element on the DOM
```
tomdale: Since `context` is an optional parameter, this should be `[context]` to indicate an optional parameter.

tomdale: This is more than just a string, this can be anything accepted by `jQuery()` as an argument, such as an element.

packages/ember-testing/lib/helpers.js
```diff
@@ -268,18 +268,19 @@ asyncHelper('visit', visit);
...
 /**
   Clicks an element and triggers any actions triggered by the element's `click`
-  event.
+  event. Accepts `context` as an optional second argument.
 
   Example:
 
   ```javascript
-  click('.some-jQuery-selector').then(function() {
+  click('.some-jQuery-selector', 'body').then(function() {
```
tomdale: Since `body` is the default scope, it seems like we could find a better example to motivate the optional second argument.

packages/ember-testing/lib/helpers.js
```diff
@@ -278,8 +278,17 @@ asyncHelper('visit', visit);
   });
   ```
 
+  Accepts `context` as an optional second argument. Useful for scoping the element lookup.
+
+  ```javascript
+  click('.some-jQuery-selector', '.modal-dialog').then(function() {
```
tomdale: Maybe the second argument should be an element instead of a selector string? This could be rewritten as:

```
click('.modal-dialog .some-jQuery-selector').then(function() {
```

Which is shorter and clearer. I actually don't know the use of the second argument so it's hard for me to suggest an atlernative. @rwjblue @locks Any ideas?

sukima:
What's the status on this? It is blocking the issue and with only three _Good for New Contributors_ issues and two (including this one) being blocked there isn't much left for new contributors to tackle.

Also why did changes to code commenting break the build? The YUIDoc syntax is correct; Travis should not have failed. Maybe it is an npm problem?

tomdale:
Looks like a transient error; I restarted the build on Travis and it's passing now.

locks:
Thank you for your contribution @faizaanshamsi. I hadn't realized there were two PRs for the same thing, and the other one was merged. See you on the next PR!

PR #13754, opened 2016-06-27
Fixes #13291 Improved helper function documentation

### LINKED ISSUE #13291: Helper documentation doesn't match cli helper generator

habdelra:
This is building on the solution from this PR: https://github.com/emberjs/ember.js/pull/13292 which includes a compromise in how much documentation we put at the class level vs the method level, as well as to include cross linking to the function level.


rwjblue:
I'm not sure exactly why, but it doesn't look like a Travis job was kicked off for this.  Can you rebase or amend to force a new build?

habdelra:
looks like its running now

PR #13755, opened 2016-06-27
Remove usage of `targetObject` computed property.

rwjblue:
This PR removes the `targetObject` CP in favor of simply looking up the correct `targetObject` when needed. This simplifies the code path for `sendAction` and removes some overloading of the term "controller" (before this change it could mean an actual outlet's controller or the parentView.controller which may be a component and not a "real" controller).

Fixes comments from @krisselden on the the `sendAction` PR (#13745 / #13532).

/cc @krisselden @chadhietala

### DIFF
--- a/packages/ember-glimmer/lib/syntax/curly-component.js
+++ b/packages/ember-glimmer/lib/syntax/curly-component.js
@@ -74,18 +74,14 @@ class CurlyComponentManager {
 
     props.renderer = parentView.renderer;
     props[HAS_BLOCK] = hasBlock;
+    // parentView.controller represents any parent components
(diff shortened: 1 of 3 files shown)


chadhietala:
👍

chadhietala:
Do you think we can consolidate these?
- `packages/ember-runtime/lib/mixins/target_action_support.js`
- `packages/ember-views/lib/mixins/view_target_action_support.js`
- `packages/ember-views/lib/mixins/action_support.js`

rwjblue:
@chadhietala - Possibly, I will review, might end up being a followup though...

chadhietala:
Seems good.

PR #13759, opened 2016-06-27
Update glimmer-engine to avoid pulling Typescript from github.

rwjblue:
It now uses typescript from npm nightlies.


rwjblue:
YOLO

PR #13761, opened 2016-06-27
[Glimmer 2] Impliment getAttr

chadhietala:
@rwjblue can we deprecate this? It is un-documented but appears to be an
intimate API. /cc @wycats @chancancode


krisselden:
We need to document this as private

PR #13763, opened 2016-06-27
[GLIMMER2] Fix ember-glimmer iterators when handling dupes in ({{#each}})

zackthehuman:
The purpose of this PR is to fix array iteration via `{{#each}}` in Glimmer2 when the iteratee contains duplicate items. The following tests from `ember-glimmer/tests/integration/syntax/each-test.js` are now passing for Glimmer2:
- it can render duplicate primitive items
- it can render duplicate objects

Key changes:
- The order that we check objects for array-like-ness changed such that we look for Emberish arrays before plain JS arrays. Emberish arrays seemed to be falling into the plain JS array check -- which seems wrong.
- `ArrayIterator` and `EmberArrayIterator` now keep track of the keys they've seen, and when a duplicate key is seen it creates a composite key using a GUID of the array as a separator.

### DIFF
--- a/packages/ember-glimmer/lib/utils/iterable.js
+++ b/packages/ember-glimmer/lib/utils/iterable.js
@@ -1,8 +1,11 @@
 import { get } from 'ember-metal/property_get';
 import { guidFor } from 'ember-metal/utils';
+import Dict from 'ember-metal/empty_object';
 import { objectAt, isEmberArray } from 'ember-runtime/mixins/array';
 import { UpdatableReference, UpdatablePrimitiveReference } from './references';
 import { isEachIn } from '../helpers/each-in';
+
+const ITERATOR_KEY_GUID = 'be277757-bbbe-4620-9fcb-213ef433cca2';
 
 export default function iterableFor(ref, keyPath) {
   return new Iterable(ref, keyFor(keyPath));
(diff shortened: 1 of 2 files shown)


packages/ember-glimmer/lib/utils/iterable.js
```diff
@@ -41,20 +41,34 @@ class ArrayIterator {
     this.length = array.length;
     this.keyFor = keyFor;
     this.position = 0;
+    this.guid = null;
+    this.seen = {};
```
chadhietala: This should be a `dict`.

packages/ember-glimmer/lib/utils/iterable.js
```diff
@@ -41,20 +41,34 @@ class ArrayIterator {
...
   next() {
-    let { array, length, keyFor, position } = this;
+    let { array, length, keyFor, position, guid, seen } = this;
 
     if (position >= length) { return null; }
 
     let value = array[position];
     let key = keyFor(value, position);
     let memo = position;
+    let hasSeen = seen[key];
```
chadhietala: Seems like this logic can be extracted into a function for both of these.

packages/ember-glimmer/lib/utils/iterable.js
```diff
@@ -68,19 +82,34 @@ class EmberArrayIterator {
...
 
     if (position >= length) { return null; }
 
     let value = objectAt(array, position);
     let key = keyFor(value, position);
+    let hasSeen = seen[key];
+
+    if (hasSeen) {
+      if (!guid) {
+        this.guid = guidFor(array);
```
krisselden: This can be const and don't use guidFor

krisselden: Basically you just need something unlikely to collide and only if you don't composite the non colliding keys

chadhietala: Is the idea to brand it like how things are branded in Glimmer e.g. [this](https://github.com/tildeio/glimmer/blob/master/packages/glimmer-reference/lib/const.ts#L5).

zackthehuman: Using a pre-generated GUID now.

chadhietala:
LGTM

PR #13764, opened 2016-06-27
[BUGFIX beta] Keep rest positional parameters when nesting contextual components if needed.

### LINKED ISSUE #13742: Positional rest params don't work as expected for nested closure components
Consider the following example:

```handlebars
{{component (component 'link-to' 'index')}}
```

This results in an error:

> You must provide one or more parameters to the link-to component.

If you re-write the declaration so that the outer component passes in positional args, then it works:

```handlebars
{{component (component 'link-to') 'index'}}
```

Which differs in behavior from using curried named parameters. I think the issue is here: https://github.com/emberjs/ember.js/blob/master/packages/ember-htmlbars/lib/utils/extract-positional-params.js#L40

```javascript
  // when no params are used, do not override the specified `attrs.stringParamName` value
  if (params.length === 0 && nameInAttrs) {
    return;
  }
```

I think that what we really want is to bail out early if `params.length` is 0, otherwise the condition could be false, which will result in creating a new stream that ultimately returns an empty array. Later the components see this empty array as "no parameters" and they don't work.

Serabe:
Steps to reproduce the bug before this patch:
1. Create a component with rest positional param (like `link-to`).
2. Create contextual component (`(component "link-to" "index")`).
3. Render it (`{{component (component "link-to" "index")}}`).
4. Realize the component did not receive `"index"` at all.
## What was causing it?

Because of how positional parameters work (both in components and contextual
components), it is needed to treat them at each level and then merge them.
Sadly, this was causing to overwrite the rest positional parameters because
when no parameters are passed, the attribute get an empty array nonetheless.

In the example above, assuming `LinkToComponent.positionalParams === 'params'`:
1. Contextual component in `2` gets its positional parameters processed and
   receives the following named parameters `{ params: ['index'] }`.
2. When rendering, the `{{component}}` processes its own parameters getting
   `{ params: [] }`. Then it merges in a `new EmptyObject()` these attributes
   on top of the ones in the contextual component. (See what I did there?
   I used the JS keyword `new` as an adjective! He he. Don't judge me, it is
   pretty late in here)
## Solving the issue

This PR changes the implementation of `mergeInNewHash` to keep the original
positional parameters if the new ones are empty and the positional parameter is
of type rest.

Fixes #13742

### DIFF
--- a/packages/ember-glimmer/tests/integration/components/closure-components-test.js
+++ b/packages/ember-glimmer/tests/integration/components/closure-components-test.js
@@ -40,6 +40,142 @@ moduleFor('@htmlbars Components test: closure components', class extends Renderi
     this.assertText('Hodi Hodari');
   }
 
+  ['@test GH#13742 keeps nested rest positional parameters if rendered with no positional parameters']() {
+    this.registerComponent('-looked-up', {
+      ComponentClass: Component.extend().reopenClass({
+        positionalParams: 'params'
+      }),
+      template: '{{#each params as |p|}}{{p}}{{/each}}'
+    });
+
+    this.render('{{component (component "-looked-up" model.greeting model.name)}}', {
+      model: {
+        greeting: 'Gabon ',
+        name: 'Zack'
+      }
+    });
+
+    this.assertText('Gabon Zack');
+
+    this.runTask(() => this.rerender());
+
+    this.assertText('Gabon Zack');
+
+    this.runTask(() => this.context.set('model.greeting', 'Good morning '));
+
+    this.assertText('Good morning Zack');
+
+    this.runTask(() => this.context.set('model.name', 'Matthew'));
+
+    this.assertText('Good morning Matthew');
+
+    this.runTask(() => this.context.set('model',  { greeting: 'Gabon ', name: 'Zack' }));
+
+    this.assertText('Gabon Zack');
+  }
+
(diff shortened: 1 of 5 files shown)


Serabe:
If you are wondering why so many tests, it is because the merging is done in several places. :)

packages/ember-glimmer/tests/integration/components/closure-components-test.js
```diff
@@ -40,6 +40,142 @@ moduleFor('@htmlbars Components test: closure components', class extends Renderi
     this.assertText('Hodi Hodari');
   }
 
+  ['@test keeps nested rest positional parameters if rendered with no positional parameters']() {
```
rwjblue: can you add `GH#13742` in the test title?

rwjblue:
Chatted in slack with @Serabe about this, he is going to rebase to trigger a fresh CI run then we can land for 2.7.

packages/ember-glimmer/tests/integration/components/closure-components-test.js
Serabe: To all of them or just this one?

Serabe: By all of them I mean only those added in this PR.

Serabe: I add that to the two that made sense to me. Let me know if you want it added to the rest!

PR #13766, opened 2016-06-28
[Glimmer] Basic `{{render}}` helper support

chancancode:
This does not deal with the router side of things yet


chancancode:
cc @chadhietala @mmun

mmun:
Ah, thanks for the ping. Feel free to 🔓  me in the other thread.

PR #13770, opened 2016-06-28
[glimmer2] fix each key collision detection

fivetanley:
When using `{{#each things key="key" as |thing|}}`, glimmer2 integration
did not update the DOM correctly when `things` was an array with objects
that shared the same `key` value. For example:

``` javascript
this.set('data', [
  {
    name: 'Yehuda' }
  },
  {
    name: 'Jenn'
  },
  {
    name: 'Yehuda'
  }
]);
```

and the following template:

``` handlebars
{{#each data key="name" as |person|}}
  {{person.name}}
{{/each}}
```

In the above example, we want each item in the list to render, even though two objects have the same
value for the key (Notice there are two "Yehuda"'s above.

As the iterator goes through the array, it will cache value of the `key` `name`.
In order to make sure the cache updates correctly, we assign it a unique key.

### DIFF
--- a/packages/ember-glimmer/lib/utils/iterable.js
+++ b/packages/ember-glimmer/lib/utils/iterable.js
@@ -17,7 +17,9 @@ function keyFor(keyPath) {
     case null:
       return identity;
     default:
-      return (item) => get(item, keyPath);
+      return function (item)  {
+        return this.generateUniqueKey(get(item, keyPath));
+      };
   }
 }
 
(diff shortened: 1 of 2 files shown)


fivetanley:
closing in favor of https://github.com/emberjs/ember.js/pull/13763/, which i thought i had claimed but i guess not :p I think @zackthehuman's PR fixes this and has the blessing of @krisselden. Maybe we should add the tests from this PR to zack's though.

PR #13773, opened 2016-06-29
[CLEANUP] reducing private API surface of views.

krisselden:
- remove unused internal `_willDestroyElement` hook
- cleanup `childViews` documentation not to refer to legacy methods
- remove undocumented `forEachChildView` used only by visibility support
  and refactor visibility support not to use it.
- remove `remove` and `removeFromParent` methods, refactor legacy style
  tests to avoid these.  Removing a view should be done via the template
  layer. Still leaves destroyElement which is equivalent and used in more
  legacy style tests.


rwjblue:
LGTM

packages/ember-views/lib/mixins/view_support.js
```diff
@@ -47,14 +47,16 @@ export default Mixin.create({
...
+    @deprecated use `yield` and contextual components for composition instead.
     @private
   */
   nearestWithProperty(property) {
-    let view = get(this, 'parentView');
+    let view = this.parentView;
 
     while (view) {
       if (property in view) { return view; }
       view = get(view, 'parentView');
```
rwjblue: This should be deleted (it is causing a JSHInt error).

PR #13774, opened 2016-06-29
[Glimmer2] Fix top level templates with loading/error states

chadhietala:
Top Level templates cannot be `ConstReferences` due to the fact they must account from top-level loading and error substates. That being said `TopLevelOutletComponentReference` is basically const because it only returns a new `TopLevelOutletComponentDefinition` when the new and last are not equal. This case only appears to occur in the substate case.

/cc @krisselden @wycats @chancancode @rwjblue

### DIFF
--- a/packages/ember-glimmer/lib/syntax/outlet.js
+++ b/packages/ember-glimmer/lib/syntax/outlet.js
@@ -1,5 +1,4 @@
 import { ArgsSyntax, StatementSyntax } from 'glimmer-runtime';
-import { ConstReference } from 'glimmer-reference';
 import { generateGuid, guidFor } from 'ember-metal/utils';
 import { RootReference } from '../utils/references';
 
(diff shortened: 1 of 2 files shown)


packages/ember-glimmer/lib/syntax/outlet.js
```diff
@@ -30,12 +29,24 @@ export class OutletSyntax extends StatementSyntax {
...
 }
 
-class TopLevelOutletComponentReference extends ConstReference {
+class TopLevelOutletComponentReference {
   constructor(reference) {
-    let outletState = reference.value();
-    let definition = new TopLevelOutletComponentDefinition(outletState.render.template);
+    this.outletReference = reference;
+    this.lastState = reference.value();
+    this.definition = new TopLevelOutletComponentDefinition(this.outletState.render.template);
```
rwjblue: Looks like this is failing when `this.outletState` is undefined.

chadhietala: Doh 😭

chadhietala:
Eh I need to do more understanding of the outlet state it appears.

chadhietala:
I think this ready for review now

Issue #13028, opened 2016-02-29
willDestroyElement is called after the component is destroyed when using the component helper


chrisseto:
When a component is rendered with the component helper it's `willDestroyElement` hook is called after the element has been removed from the DOM rather than before.

The twiddle contains additional details.

[Ember Twiddle](https://ember-twiddle.com/a97a8db3529bb5899929)

krisselden:
Seems like the teardown path is very different for `{{component "my-component"}}` than it is for `{{my-component}}` stepping through in a debugger.

Issue #12716, opened 2015-12-15
willDestroyElement not firing in {{each}} else


amk221:
```handlebars
A.                            |     B.
{{#each items as |item|}}     |     {{#each items as |item|}}
  {{!...}}                    |        {{!...}}
{{else}}                      |     {{/each}}
  {{my-component}}            |     {{#unless items.length}}
{{/each}}                     |       {{my-component}}
                              |     {{/unless}}
```

In example A. the `willDestroyElement` hook is not firing for `{{my-component}}` when it is torn down.
It works for example B.
[jsfiddle](http://jsfiddle.net/amk221/k2y2tc0j/)

amk221:
Thank you everyone!

PR #13775, opened 2016-06-30
[BUGFIX beta] Cleanup view teardown.

### LINKED ISSUE #12080: parentView undefined during willDestroyElement
This happens at least since v1.13.5. When changing to an older v1.11.1, assert never fails.

jsbin: http://emberjs.jsbin.com/lireho/edit?html,js,output

It would make sense that the parentView is still present since the event is called before the element gets destroyed.
So does this work as intended?

krisselden:
- [x] Add tests to ensure `.parentView` is available in all hooks.
- [x] Add tests to ensure `.element` is available in appropriate hooks.
- [x] add tests that assure counts of hooks in all dynamic scenarios
- [x] add regression tests
  - [x] `willDestroyElement` is called for components invoked from the `{{else}}` hook of an `{{#each` https://github.com/emberjs/ember.js/issues/12716.
  - [x] `this.parentView` is available in `willDestroyElement` https://github.com/emberjs/ember.js/issues/12080.
  - [x] `willDestroyElement` is called before actual destruction for components rendered with `{{component` helper.

This should fix #12080 #12716 and #13028

### DIFF
--- a/packages/ember-glimmer/lib/renderer.js
+++ b/packages/ember-glimmer/lib/renderer.js
@@ -192,7 +192,9 @@ class Renderer {
       _renderResult.destroy();
     }
 
-    view.destroy();
+    if (!view.isDestroying) {
+      view.destroy();
+    }
   }
 
   componentInitAttrs() {
--- a/packages/ember-htmlbars/lib/component.js
+++ b/packages/ember-htmlbars/lib/component.js
@@ -193,7 +193,7 @@ const Component = View.extend(TargetActionSupport, ActionSupport, {
     let attr = this._renderNode.childNodes.filter(node => node.attrName === name)[0];
     if (!attr) { return null; }
     return attr.getContent();
-  }
+  },
 
   /**
     Returns true when the component was invoked with a block template.
@@ -359,6 +359,7 @@ const Component = View.extend(TargetActionSupport, ActionSupport, {
     @public
     @since 1.13.0
   */
+  didReceiveAttrs() {},
 
   /**
     Called when the attributes passed into the component have been updated.
@@ -379,6 +380,7 @@ const Component = View.extend(TargetActionSupport, ActionSupport, {
     @public
     @since 1.13.0
   */
+  didRender() {},
 
   /**
     Called after a component has been rendered, both on initial render and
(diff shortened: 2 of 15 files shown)


packages/ember-glimmer/lib/renderer.js
```diff
@@ -192,7 +192,9 @@ class Renderer {
       _renderResult.destroy();
     }
 
-    view.destroy();
+    if (!view.isDestroying) {
```
stefanpenner: should we guard against both `isDestroying` and `isDestroyed`

packages/ember-htmlbars/lib/component.js
```diff
@@ -465,7 +471,8 @@ const Component = View.extend(TargetActionSupport, ActionSupport, {
 Component[NAME_KEY] = 'Ember.Component';
 
 Component.reopenClass({
-  isComponentFactory: true
+  isComponentFactory: true,
+  positionalParams: []
```
stefanpenner: tell me more

packages/ember-views/lib/views/states/in_dom.js
```diff
@@ -27,7 +27,9 @@ assign(inDOM, {
   },
 
   exit(view) {
-    view._unregister();
+    if (view.tagName !== '') {
```
stefanpenner: this may need a test

packages/ember-glimmer/lib/renderer.js
krisselden: we don't unset isDestroying when we set isDestroyed but I can add that

packages/ember-htmlbars/lib/component.js
krisselden: so I mean to move the doc here too, but the issue is `if (positionalParams) {` will optimize to false because this is uncommon but not exceptional and the branch deopts, I was doing a lot of clean as I go, I can remove that stuff and make it a separate PR

krisselden: this is the same with all the hooks

stefanpenner: ah, and this is already `prototype` only type thing. Seems good thanks

packages/ember-glimmer/lib/renderer.js
stefanpenner: ah, then its not needed. Seems good

rwjblue:
OK, I added regression tests for the various scenarios that were tweaked here (including the specific issues linked in the description).

krisselden:
@rwjblue thank you so much for doing these tests.

PR #13778, opened 2016-06-30
[Glimmer2] Implement glimmer local lookup for components

asakusuma:
Replaces https://github.com/emberjs/ember.js/pull/13378

### DIFF
--- a/packages/ember-glimmer/lib/environment.js
+++ b/packages/ember-glimmer/lib/environment.js
@@ -159,7 +159,7 @@ export default class Environment extends GlimmerEnvironment {
    // isn't going to return any syntax and the Glimmer engine knows how to handle
    // this case.
 
-  refineStatement(statement) {
+  refineStatement(statement, parentMeta) {
     // 1. resolve any native syntax – if, unless, with, each, and partial
     let nativeSyntax = super.refineStatement(statement);
 
@@ -188,7 +188,7 @@ export default class Environment extends GlimmerEnvironment {
       let mappedKey = builtInComponents[key];
 
       if (key === 'component') {
-        return new DynamicComponentSyntax({ args, templates });
+        return new DynamicComponentSyntax({ args, templates, parentMeta });
       } else if (key === 'render') {
         return new RenderSyntax({ args });
       } else if (key === 'outlet') {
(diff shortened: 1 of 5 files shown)


packages/ember-glimmer/lib/environment.js
```diff
@@ -224,12 +227,15 @@ export default class Environment extends GlimmerEnvironment {
     return false;
   }
 
-  getComponentDefinition(path) {
+  getComponentDefinition(path, parentMeta = {}) {
```
krisselden: Lets not default it here, lets either ensure it is set once on the template and always there or check presence

asakusuma: @krisselden what would be the right way to get the parent meta info in the dynamic case?

So [here](https://github.com/emberjs/ember.js/blob/c8824b01ccc1105fc1e9191b23ac905efd760cd4/packages/ember-glimmer/lib/syntax/dynamic-component.js#L49) and [here](https://github.com/emberjs/ember.js/blob/c8824b01ccc1105fc1e9191b23ac905efd760cd4/packages/ember-glimmer/lib/syntax/dynamic-component.js#L12).

asakusuma:
@krisselden I added [some tests](https://github.com/emberjs/ember.js/pull/13778/files#diff-60d0781473fafba06d2d55675ee1286cR81), which are failing, that illustrate the dynamic component local lookup case that I need guidance on.

packages/ember-glimmer/lib/environment.js
```diff
@@ -198,18 +198,21 @@ export default class Environment extends GlimmerEnvironment {
       let internalKey = builtInComponents[key];
       let definition = null;
 
+      let template = templates && templates['default'];
```
krisselden: this doesn't work for `inverse`

```hbs
{{#my-component}}  default {{else}} inverse {{/my-component}}
```

packages/ember-glimmer/lib/environment.js
```diff
@@ -224,12 +227,15 @@ export default class Environment extends GlimmerEnvironment {
     return false;
   }
 
-  getComponentDefinition(path) {
+  getComponentDefinition(path, parentMeta = {}) {
     let name = path[0];
     let definition = this._components[name];
```
krisselden: The name needs to be the resolved name (not the local name) before looking up the cached compiled definition. We need to refactor to address this, maybe use the _debugContainerKey or less debug friendly, use guidFor(ComponentClass) as the definition key.

asakusuma:
@krisselden blueprints is failing for some reason. everything else passes

`npm ERR! code 1`

rwjblue:
This looks good to me.

@asakusuma - Thanks for working on this, it has been a looooonnnnggggg road... 😸 

@krisselden - r?

PR #13781, opened 2016-07-02
[BUGFIX release] Fix NoneLocation#getURL

wagenet:
Prior fixes to NoneLocation#getURL broke the case where the rootURL
matched the start of the main URL. Resolve that case by checking for
a subsequent forward slash or end of string.


rwjblue:
This looks good, but based on the changes added in https://github.com/emberjs/ember.js/pull/13520, it seems that history location would have the same issue.

@wagenet - Would you mind updating history also?

wagenet:
@rwjblue should be good to go!

rwjblue:
Thank you!

PR #13782, opened 2016-07-02
[WIP] Ensure correct shaping for `isDestroy[ing|ed]` on CoreObject.

rwjblue:
Shaping the prototype is not enough, we should ensure the instance properties are set in the constructor.

### DIFF
--- a/packages/ember-runtime/lib/system/core_object.js
+++ b/packages/ember-runtime/lib/system/core_object.js
@@ -50,6 +50,13 @@ var finishPartial = Mixin.finishPartial;
 var reopen = Mixin.prototype.reopen;
 var hasCachedComputedProperties = false;
 
+const FALSE_DESCRIPTOR = {
+  configurable: true,
+  writable: true,
+  enumerable: false,
+  value: false
+};
+
 function makeCtor() {
   // Note: avoid accessing any properties on the object since it makes the
   // method a lot faster. This is glue code so we want it to be as fast as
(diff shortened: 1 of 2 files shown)


mixonic:
@rwjblue is this something we should be benchmarking to show improvement? Or are you otherwise just waiting on review from a perf team member (I expect @krisselden or @stefanpenner )?

rwjblue:
Yeah, this is still a bit WIP (I'll update it as such).

stefanpenner:
> @rwjblue is this something we should be benchmarking to show improvement?

This is most likely to small a change to notice much.

* today i do not believe those properties are nonEnumerable once set
* setting them via the nonEnumerable path is most likely going to be slower during construction. Is it possible for us to just set them as a property or ideally, move them to meta?

krisselden:
@stefanpenner the states should be moved to meta, and we should have accessors on the prototype that defer to meta for backwards compatibility.

I agree that this PR is a step backwards and that it is dubious to test non enumerability since they are enumerable once set currently.

stefanpenner:
> @stefanpenner the states should be moved to meta, and we should have accessors on the prototype that defer to meta for backwards compatibility.
> 
> I agree that this PR is a step backwards and that it is dubious to test non enumerability since they are enumerable once set currently.

yes

rwjblue:
Completely replaced by https://github.com/emberjs/ember.js/pull/14359

PR #13784, opened 2016-07-02
[BUGFIX beta] Various Chrome deopt fixes.

rwjblue:
- Prevent indexing into primitive types in `guidFor`. Doing `obj[GUID_KEY]` on primitive objects (strings, booleans, numbers, etc) will not work, and triggers a deopt.
- Prevent deopting `Mixin.detect` for each type found. With the truthiness check that existed before, this method is deopted everytime it sees a new type of value. (See screenshots from irhydra in commit message).
- Check `typeof` before testing for truthiness. Using the `typeof` check before the truthiness check, prevents deopts in Chrome for each new object type that is seen (after it initially optimizes).

### DIFF
--- a/packages/ember-glimmer/lib/utils/references.js
+++ b/packages/ember-glimmer/lib/utils/references.js
@@ -110,7 +110,7 @@ export class PropertyReference extends CachedReference { // jshint ignore:line
       _parentObjectTag.update(tagFor(parentValue));
     }
 
-    if (parentValue && typeof parentValue === 'object') {
+    if (typeof parentValue === 'object' && parentValue) {
       if (isEnabled('mandatory-setter')) {
         let meta = metaFor(parentValue);
         watchKey(parentValue, _propertyKey, meta);
--- a/packages/ember-metal/lib/utils.js
+++ b/packages/ember-metal/lib/utils.js
@@ -179,7 +179,11 @@ export function generateGuid(obj, prefix) {
   @return {String} the unique guid for this instance.
 */
 export function guidFor(obj) {
-  if (obj && obj[GUID_KEY]) {
+  let type = typeof obj;
+  let isObject = type === 'object' && obj !== null;
+  let isFunction = type === 'function';
+
+  if ((isObject || isFunction) && obj[GUID_KEY]) {
     return obj[GUID_KEY];
   }
 
@@ -193,7 +197,6 @@ export function guidFor(obj) {
   }
 
   let ret;
-  let type = typeof obj;
 
   // Don't allow prototype changes to String etc. to change the guidFor
   switch (type) {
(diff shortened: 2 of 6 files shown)


packages/ember-metal/lib/utils.js
```diff
@@ -179,7 +179,8 @@ export function generateGuid(obj, prefix) {
   @return {String} the unique guid for this instance.
 */
 export function guidFor(obj) {
-  if (obj && obj[GUID_KEY]) {
+  let type = typeof obj;
+  if ((type === 'object' || type === 'function') && obj !== null && obj[GUID_KEY]) {
```
krisselden: (type === 'object' && obj !== null) || type === 'function'

krisselden:
@rwjblue you should do the isAngle out of bounds one for the LTS

rwjblue:
Good point @krisselden, submitted https://github.com/emberjs/ember.js/pull/13785 for that.

packages/ember-metal/lib/utils.js
rwjblue: good point, I refactored that a bit to avoid the `!== null` check for functions.

rwjblue:
Confirmed that the `guidFor` and `Mixin.detect` deopts addressed here are fixed (via `initial-render-perf` repo, and irhyrda).  There is still a "wrong map" deopt for `guidFor` (which will continue to happen until we move the GUID into a WeakMap).

rwjblue:
### Before

screenshot

### After

screenshot

krisselden:
@rwjblue that complex list benchmark is pretty bogus, we need to be benchmarking real apps.

rwjblue:
KK, will test with chrome-tracing and report back...

krisselden:
@rwjblue these seem fine, you won't see benefit likely until the downstream deopts are fixed, guidFor needs the define non enumerable property path warmed since it optimizes before hitting it in views.

packages/ember-glimmer/lib/utils/references.js
```diff
@@ -110,7 +110,7 @@ export class PropertyReference extends CachedReference { // jshint ignore:line
       _parentObjectTag.update(tagFor(parentValue));
     }
 
-    if (parentValue && typeof parentValue === 'object') {
+    if (typeof parentValue === 'object' && parentValue) {
```
stefanpenner: rather then coercion we can be explicit, the only case we care about here is `parentValue !== null`

stefanpenner:
This looks good to me.

stefanpenner:
@homu r+

stefanpenner:
@homu retry

stefanpenner:
lets assume homu is dead..

PR #13785, opened 2016-07-02
[CLEANUP/BUGFIX lts] Remove remnants of angle bracket components.

rwjblue:
The original techniques/hooks were removed when the feature was removed, but this code remained. Unfortunately, when tagName is '' we were still attempting to `''.charCodeAt(0)` which deopts the native `charCodeAt`.


rwjblue:
@krisselden was concerned that removing this would remove the ability to render web components.  I have added tests to ensure that the web component fallback continues to work.

chadhietala:
LGTM

rwjblue:
I pulled this into `beta`, but `release` and `lts-2-4` will be pretty hard to pull it into (this code has been massively refactored in 2.7 betas making cherry-picking tricky).

PR #13787, opened 2016-07-03
Use meta to build `eachComputedProperty`.

rwjblue:
This avoids iterating all properties on a given class when `.eachComputedProperty` is used by storing each descriptor in the prototypes meta.

This also fixes an issue where, `.eachComputedProperty` would _not_ update properly when `Ember.defineProperty` is used to add a CP (instead of `.extend` or `.reopen`).

/cc @krisselden

### DIFF
--- a/packages/ember-metal/lib/mixin.js
+++ b/packages/ember-metal/lib/mixin.js
@@ -345,6 +345,9 @@ function applyMixin(obj, mixins, partial) {
 
   obj._super = ROOT;
 
+  let knownDescriptors = [];
+  m.forEachDescriptor(name => knownDescriptors.push(name));
+
   // Go through all mixins and hashes passed in, and:
   //
   // * Handle concatenated properties
@@ -367,6 +370,12 @@ function applyMixin(obj, mixins, partial) {
       let followed = followAlias(obj, desc, m, descs, values);
       desc = followed.desc;
       value = followed.value;
+    }
+
+    if (desc && desc.isDescriptor) {
+      m.writeDescriptor(key, desc);
+    } else if (knownDescriptors.indexOf(key) > -1) {
+      m.resetDescriptor(key);
     }
 
     if (desc === undefined && value === undefined) { continue; }
--- a/packages/ember-runtime/lib/system/core_object.js
+++ b/packages/ember-runtime/lib/system/core_object.js
@@ -7,8 +7,6 @@
   @submodule ember-runtime
 */
 
-// using ember-metal/lib/main here to ensure that ember-debug is setup
-// if present
 import {
   assign,
   guidFor,
@@ -21,7 +19,6 @@ import {
   assert,
   runInDebug,
   isFeatureEnabled,
-  get,
   meta,
   finishChains,
   sendEvent,
@@ -32,7 +29,6 @@ import {
   defineProperty,
   Binding,
   ComputedProperty,
-  computed,
   InjectedProperty,
   run,
   destroy
@@ -45,7 +41,6 @@ var schedule = run.schedule;
 var applyMixin = Mixin._apply;
 var finishPartial = Mixin.finishPartial;
 var reopen = Mixin.prototype.reopen;
-var hasCachedComputedProperties = false;
 
 function makeCtor() {
   // Note: avoid accessing any properties on the object since it makes the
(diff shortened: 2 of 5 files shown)


rwjblue:
If this looks good, we could expand this throughout the codebase wherever we are looking for descriptors.  A few examples:

* [calling descriptor `.set` method in Ember.set](https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/lib/property_set.js#L51-L61)
* [calling descriptor `.get` method in Ember.get](https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/lib/property_get.js#L58)
* [calling descriptor `willWatch` in watchKey](https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/lib/watch_key.js#L21-L25)
* [calling descriptor `willChange` in propertyWillChange](https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/lib/property_events.js#L52)

rwjblue:
Found one issue for a previously untested scenario (subclass overriding a parent's descriptor with `null` / `undefined` to reset it) while running Ember Data's test suite against this change. I addressed that by adding a `resetDescriptor` method, and marking the location with a special value to stop the upwards search of parent metas.

As of these changes, the Ember Data test suite is green with these changes.

packages/ember-runtime/lib/system/core_object.js
```diff
@@ -930,17 +907,4 @@ CoreObject.ClassMixin = ClassMixin;
 
 ClassMixin.apply(CoreObject);
 
-CoreObject.reopen({
-  didDefineProperty(proto, key, value) {
```
rwjblue: I can leave this method, but given the rest of the changes here I do not see that it is needed...

packages/ember-metal/lib/mixin.js
```diff
@@ -364,6 +364,12 @@ function applyMixin(obj, mixins, partial) {
       value = followed.value;
     }
 
+    if (desc && desc.isDescriptor) {
+      m.writeDescriptor(key, desc);
+    } else if (m.peekDescriptor(key)) {
```
rwjblue: If we wanted to avoid traversing the inherited meta's for each key that is merged, we could use `m.forEachDescriptor` once (outside of this loop) and change this guard to check if the key being added is in that listing.

Something like:

```js
let knownDescriptors = [];
m.forEachDescriptor(name => knownDescriptors.push(name));

// ...snip...
  mergeMixins(mixins, m, descs, values, obj, keys);

  for (let i = 0; i < keys.length; i++) {
    // ...snip...
    if (desc && desc.isDescriptor) {
      m.writeDescriptor(key, desc);
    } else if (knownDescriptors.indexOf(key) > -1) {
      m.resetDescriptor(key);
    }
```

Thoughts?

rwjblue: Hehe, I liked that better so I went ahead and updated.

packages/ember-runtime/lib/system/core_object.js
krisselden: It is used by ember data

krisselden: The hook not this method but until we make hooks in meta we probably want to keep an empty hook here

packages/ember-runtime/lib/system/core_object.js
```diff
@@ -873,15 +852,13 @@ var ClassMixinProps = {
...
-    var properties = get(this, '_computedProperties');
+    this.proto(); // ensure prototype is initialized
+    let empty = {};
 
-    for (var i = 0; i < properties.length; i++) {
-      property = properties[i];
-      callback.call(binding || this, property.name, property.meta || empty);
-    }
+    let m = meta(this.prototype);
+    m.forEachDescriptor((name, descriptor) => {
```
krisselden: We need to ensure this inlines

krisselden: I have doubts because of the binding

krisselden: IR hydra shows this look at the bindings in finishPartial

krisselden:
Looks good we need to test with ember data before and after with realistic data

rwjblue:
I tested https://github.com/rust-lang/crates.io using chrome-tracing, and recorded the details in [this gist](https://gist.github.com/rwjblue/2feb28161adb3ce096b0eaa321ac911b).

packages/ember-runtime/lib/system/core_object.js
rwjblue: @krisselden - I tested this branch against crates.io (locally checked out and linked), and I confirmed that `meta.forEachDescriptor` was inlined in `applyMixins`, but I do not see `eachComputedProperty` itself listed in the irhydra output.  (I uploaded to our shared profiles folder if you want to see the results)

krisselden:
Using the data from that gist https://gist.github.com/krisselden/55ddfa5b081fec16d88c8881768ec472

PR #13789, opened 2016-07-04
[Glimmer2] Implement local lookup for helpers

asakusuma:
cc @krisselden @rwjblue

### DIFF
--- a/packages/ember-glimmer/lib/environment.js
+++ b/packages/ember-glimmer/lib/environment.js
@@ -212,12 +212,12 @@ export default class Environment extends GlimmerEnvironment {
         return generateBuiltInSyntax(statement, (path) => this.getComponentDefinition([path], parentMeta));
       }
 
-      assert(`Could not find component named "${key}" (no component or template with that name was found)`, !isBlock || this.hasHelper(key));
+      assert(`Could not find component named "${key}" (no component or template with that name was found)`, !isBlock || this.hasHelper(key, parentMeta));
     }
 
-    assert(`Helpers may not be used in the block form, for example {{#${key}}}{{/${key}}}. Please use a component, or alternatively use the helper in combination with a built-in Ember helper, for example {{#if (${key})}}{{/if}}.`, !isBlock || !this.hasHelper(key));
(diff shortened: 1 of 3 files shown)


krisselden:
@asakusuma tests aren't passing, the helper tests in "Application Lifecycle" module

packages/ember-glimmer/lib/environment.js
```diff
@@ -257,12 +257,16 @@ export default class Environment extends GlimmerEnvironment {
...
+      this.owner.hasRegistration(`helper:${name}`) ||
+      (parentMeta && this.owner.hasRegistration(`helper:${name}`, { source: `template:${parentMeta.moduleName}` }));
   }
 
-  lookupHelper(name) {
-    let helper = builtInHelpers[name[0]] || this.owner.lookup(`helper:${name}`);
+  lookupHelper(name, parentMeta) {
+    let helper = builtInHelpers[name[0]] ||
+      this.owner.lookup(`helper:${name}`) ||
+      (parentMeta && this.owner.lookup(`helper:${name}`, { source: `template:${parentMeta.moduleName}` }));
```
asakusuma: @krisselden @rwjblue for some reason, doing a `lookup` on a helper with `source='-top-level'` doesn't work. Same for `hasRegistration`. This works for components, but not helpers. To demonstrate, I've added a `lookup` without the `source` info, and then a fallback with `source` info, and things work. Any idea why this is?

asakusuma:
I think I may need to add the moduleName somewhere in regusterHelper in abstract-test-case

asakusuma:
looks like the [htmlbars implementation](https://github.com/emberjs/ember.js/blob/28cedad1418bedd2dad44eb56668fa67b9769d86/packages/ember-htmlbars/lib/system/lookup-helper.js#L57-L65) attempts to lookup the helper using the `source`, but then falls back to a lookup without the `source` info. Is it acceptable to use the same method in glimmer? Doesn't seem ideal but I feel like this is something we can come back to and improve later. @krisselden @rwjblue

rwjblue:
Confirm, we must continue to do two lookups.

PR #13790, opened 2016-07-05
Update api doc for Enumerable#contains to flag it as deprecated

bmeurant:
Doc update related to #13553.

For merging only once `ember-runtime-enumerable-includes` has been enabled without feature flag.


rwjblue:
Thanks again for all of your hard work on this @bmeurant!

Issue #13227, opened 2016-04-01
JSCS tests fail when built in Windows


AmazingJaze:
@rwjblue quick question about unit test results. When I follow steps in the readme to build Ember and run unit tests in the browser I get 160+ jscs test failures. 
![ember_2 3 0_testresults](https://cloud.githubusercontent.com/assets/6732159/14199235/71b5bb4c-f797-11e5-8365-fc96b065fe9e.PNG)

So far I've verified that I see these failures in both Chrome and Edge for the v2.3.0, v2.4.3 and master branches of Ember. Any idea what's going on here? Ember-ci indicates all tests [should be passing](https://saucelabs.com/tests/d7c5d5c161044d099719c17bbcab173d)

// CC @jdalton

AmazingJaze:
To be clear, this is through running tests using `npm start` and http://localhost:4200/

stefanpenner:
Typically questions should be asked on discourse or stack overflow. This tracker is specifically for tracking bugs

jdalton:
Doc bug?

rwjblue:
The tests do generally pass. I suspect there is some sort of environment issue causing these failures. Based on the failure you pasted, my complete guess would be that the exception we have [here](https://github.com/emberjs/ember.js/blob/master/lib/jscs-rules/require-comments-to-include-access.js#L8) is not properly handling \r\n...

stefanpenner:
Hmm. Smells like a windows and its line endings then.

AmazingJaze:
I've edited the title for clarity.

Thanks @rwjblue and @stefanpenner  for the tips. I went and modified the [regex's here](https://github.com/AmazingJaze/ember.js/commit/317cae766471b6e0255846d95bad087e1cc26850) in my fork but it didn't seem to address the issue. 

Curiously [this function](https://github.com/AmazingJaze/ember.js/blob/317cae766471b6e0255846d95bad087e1cc26850/lib/jscs-rules/require-comments-to-include-access.js#L7-L9) doesn't appear to be running in Windows at all. For example, I added a console.log to it and re-ran the build commands but that console log didn't ever fire. Is it expected that this code would not run during the build process?

I also noticed the EmberJS repo doesn't have a .gitattributes file to automatically normalize line endings for collaborators. I've issued PR #13232  to add this behavior to the repo. Feel free to reject, if its previous exclusion was a conscious decision.

pixelhandler:
[...]

PR #13791, opened 2016-07-05
Fix Windows builds

bmeurant:
This PR is related to bug #13227 and should fix it.

Moreover, I updated `package.json` to change `test` and `test:sauce` script to add explicit `node` command behind. Without that, these command were not working on Windows.

I successfully tested both JSCS update and scripts on Windows and Ubuntu. I guess it should be also OK on OSX.

Let me know if I missed something.


rwjblue:
Awesome, thank you!

PR #13797, opened 2016-07-07
[BUGFIX beta] Ensure didInitAttrs deprecation is stripped in prod.

rwjblue:
In production builds, the `didInitAttrs` deprecation was not being fully stripped. This refactors to remove the guarding `if` (in favor `Ember.deprecate` test parameter).


chadhietala:
LGTM

PR #13799, opened 2016-07-07
Rename "Object" imports to "EmberObject"

Turbo87:
Importing this as "Object" hides the native "Object" API which can lead to unexpected results.

This is also necessary to enable https://github.com/emberjs/emberjs-build/pull/165


rwjblue:
Thank you!

PR #13800, opened 2016-07-07
Create `FunctionMeta` to store metadata for functions.

rwjblue:
This allows us to stop holding state information on functions themselves (aka stop treating functions as a "bag of properties").

Once this PR lands we will still be storing these function meta objects on the function itself (in either a `WeakMap` when supported or on a `__ember_function_meta__` property when `WeakMap` is not present) just like we store normal object meta.

/cc @krisselden

### DIFF
--- a/packages/ember-metal/lib/events.js
+++ b/packages/ember-metal/lib/events.js
@@ -6,13 +6,12 @@
 @module ember
 @submodule ember-metal
 */
-import { applyStr } from 'ember-utils';
+import { applyStr, functionMetaFor } from 'ember-utils';
 import { assert } from './debug';
 import { meta as metaFor, peekMeta } from './meta';
 import { deprecate } from './debug';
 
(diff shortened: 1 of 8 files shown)


rwjblue:
Just rebased this. It seems likely that we will land https://github.com/emberjs/ember.js/pull/13783 before landing this. If that is what we do, I'll be able to refactor things around to share the `setMeta` / `peekMeta` / `deleteMeta` WeakMap system with fallback for unsupported platforms.

rwjblue:
This PR has been updated to account for changes in the codebase (putting the `function-meta` implementation into `ember-utils` package to be used by `ember-utils/super`), and use native `WeakMap` when present (along the same lines as https://github.com/emberjs/ember.js/pull/13783 which just landed).  

@chancancode - Can you do a run of this against current master in ember-bench so we can confirm no performance regressions?

@krisselden - This is ready for another round of review...

rwjblue:
replaced by #16595

PR #13801, opened 2016-07-07
Avoid using a RegExp to check for `*Binding` properties.

rwjblue:


packages/ember-metal/lib/mixin.js
```diff
@@ -257,8 +257,12 @@ function mergeMixins(mixins, m, descs, values, base, keys) {
 
 let IS_BINDING = /^.+Binding$/;
 
-function detectBinding(obj, key, value, m) {
-  if (IS_BINDING.test(key)) {
+export function detectBinding(obj, key, value, m) {
+  if (key.length < 8) { return false; }
+
+  let lastIndex = key.indexOf('Binding', key.length - 7);
```
cibernox: is this better/faster than `String.prototype.lastIndexOf`

rwjblue: @cibernox - Well, its a bit different.  If we used `lastIndexOf`, it would continue to search the string (in reverse) until it got to the first character.  In our case, we only want to check if the last 7 chars match exactly `Binding`, so we use `.indexOf('Binding', length - 7)` so it effectively only has to check a single possible match.

rwjblue:
@krisselden - Do you want me to benchmark this, or are you comfortable with it as an improvement anyways?

packages/ember-metal/lib/mixin.js
stefanpenner: some other alternatives would be:

```
    var l = a.length;
    a.charAt(l - 1) !== 'g' &&
    a.charAt(l - 2) !== 'n' &&
    a.charAt(l - 3) !== 'i' &&
    a.charAt(l - 4) !== 'd' &&
    a.charAt(l - 5) !== 'n' &&
    a.charAt(l - 6) !== 'i' &&
    a.charAt(l - 7) !== **'';**
```

```
    var l = a.length;
    a.charAt(l - 1) !== 'g' && a.charAt(l = 7) !== 'B' && a.substring(l -7, l) !== 'Binding'
```

```js
   'Binding' !==  a.substring(a.length - 7, a.length);
```

rwjblue: Sure, but are they needed? Is there an issue with the implementation in this PR?

stefanpenner: my guess is the substring and charAt ones will perform similarly, and most likely fairly well.

stefanpenner: a quick benchmark yields:

```
  charAt ....... 26,712,091.19 op/s
  indexOf ...... 15,829,075.57 op/s
  RegExp ....... 7,994,272.77 op/s
  substring .... 27,049,018.51 op/s
```

stefanpenner: my test is always testing the same string though, i would be curious to see across a wide distributing of strings but interned and not interned.... hmmm

stefanpenner: ah if anyone cares to explore further (i was nerd sniped) 

see: https://github.com/stefanpenner/perf-scratch/commit/a57d6f9dcd9332928b813951c5a5774f3af0fe05

stefanpenner: I didn't do any dives to see what the underlying implementation is emitting so my test is unproven..

krisselden:
with real keys from our app that pass through detectBinding()

```
  indexOf ................ 24,744,886.53 op/s
  charCodeAt + indexOf .. 115,503,607.03 op/s
  charAt ................. 82,576,764.78 op/s
  charCodeAt ............. 94,268,708.09 op/s
  RegExp ................. 17,971,357.65 op/s
  substring .............. 41,617,095.26 op/s
```

Fastest being:
```
function detectBinding(key) {
  var l = key.length;
  return l > 7 && key.charCodeAt(l - 7) === 66 && key.indexOf('inding', l - 6) !== -1;
}
```

krisselden:
https://gist.github.com/krisselden/452507ecaa2e86f188023566f8b30a83

stefanpenner:
@krisselden cool, can you also share the json file?

packages/ember-metal/lib/mixin.js
stefanpenner: based on @krisselden work it seems like the following is out best bet:

````
 var l = key.length;
  return l > 7 && key.charCodeAt(l - 7) === 66 && key.indexOf('inding', l - 6) !== -1;
```

krisselden:
Samsung Galaxy III Chrome 51
```
indexOf ................. 2,969,546.15 op/s
charCodeAt + indexOf ... 18,193,348.89 op/s
charAt ................. 14,873,922.51 op/s
charCodeAt ............. 16,139,829.31 op/s
RegExp .................. 1,590,693.48 op/s
substring ............... 6,005,415.44 op/s
```

packages/ember-metal/lib/mixin.js
```diff
@@ -257,8 +257,12 @@ function mergeMixins(mixins, m, descs, values, base, keys) {
 
 let IS_BINDING = /^.+Binding$/;
```
krisselden: let's remove the RegExp

rwjblue: Its publicly exported as `Ember.IS_BINDING`. I'm fine with killing that, but I don't know if folks are using it.  I guess I can add a deprecation in `ember-metal/index` for it when accessed.

rwjblue:
Updated based on feedback. 

**Note:** I removed `Ember.IS_BINDING` global export, as I do not think this qualifies as an intimate private API ([very few results from a github search](https://github.com/search?l=JavaScript&q=IS_BINDING&type=Code&utf8=%E2%9C%93)).

packages/ember-metal/lib/mixin.js
rwjblue: I ended up removing the regexp and the export. There are only 208 results on GH search, and I reviewed about 50% of them are forks and gists of ember-metal/mixin itself.

packages/ember-runtime/lib/system/core_object.js
```diff
@@ -100,9 +100,7 @@ function makeCtor() {
           var keyName = keyNames[j];
           var value = properties[keyName];
 
-          if (IS_BINDING.test(keyName)) {
-            m.writeBindings(keyName, value);
```
stefanpenner: is there a reason we a testing if bindings exist now, but no longer writing them? I suspect this deletion was in error, or I am mis understanding something.

rwjblue: Nope, you are correct. This was a rebase error.

rwjblue: Fixed now, thanks for spotting.

stefanpenner: i wonder how few apps would have broken.. :P

stefanpenner:
LGTM :shipit:

PR #13807, opened 2016-07-11
[glimmer2] enable tests fixed by #13615

fivetanley:
refs #13644

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "express": "^4.5.0",
     "finalhandler": "^0.4.0",
     "github": "^0.2.3",
-    "glimmer-engine": "tildeio/glimmer#256f044",
+    "glimmer-engine": "tildeio/glimmer#b618846",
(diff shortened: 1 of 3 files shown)


chadhietala:
These are going to fail until the [two-flush branch](https://github.com/emberjs/ember.js/tree/two-way-flush) is merged I believe.

fivetanley:
@chadhietala They seem to pass individually. Can you help me understand what the two-flush branch does and why it might fix these issues?

chancancode:
@chadhietala I thought they were related, but I debugged this with @fivetanley and I think we just made a mistake (forgetting to clear the DOM when destroying the `RenderResult`). I am pretty confused how it went unnoticed for so long. We much have some other paths that "incidentally" clears the DOM most of the time so the other tests are not affected by it?

chancancode:
Merged in #13832

PR #13811, opened 2016-07-12
[Glimmer 2] Unskip tests related to inline styles and unsafe hrefs

chadhietala:
Part of #13644

### DIFF
--- a/packages/ember-glimmer/tests/integration/content-test.js
+++ b/packages/ember-glimmer/tests/integration/content-test.js
@@ -7,7 +7,7 @@ import EmberObject from 'ember-runtime/system/object';
 import ObjectProxy from 'ember-runtime/system/object_proxy';
 import { classes } from '../utils/test-helpers';
 import { getDebugFunction, setDebugFunction } from 'ember-metal/debug';
-import { styleWarning } from 'ember-htmlbars/morphs/attr-morph';
+import { STYLE_WARNING } from 'ember-views/system/utils';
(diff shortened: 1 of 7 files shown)


packages/ember-glimmer/tests/integration/content-test.js
```diff
@@ -975,10 +975,10 @@ if (!EmberDev.runningProdBuild) {
         userValue: 'width: 42px'
       });
 
-      assert.deepEqual(warnings, [styleWarning]);
+      assert.deepEqual(warnings, [STYLE_WARNING]);
     }
 
-    ['@test specifying `<div style={{{userValue}}}></div>` works properly without a warning'](assert) {
+    ['@htmlbars specifying `<div style={{{userValue}}}></div>` works properly without a warning'](assert) {
```
chadhietala: @krisselden Not sure how to handle this case because we have a `TrustingAppendOpcode` here which blindly appends and the context is lost in the compile.

chadhietala:
@rwjblue Does this test indicator mean that typescript needs to be a `devDependency` on Ember now? 

```
npm WARN broccoli-typescript-compiler@0.6.1 requires a peer of typescript@^1.6.2 || ^1.7.0 || ^1.8.0 || ^1.9.0-dev but none was installed.
npm ERR! code 1
```

rwjblue:
I don't think so, we are using 2.0.0-dev, but that isn't one of the options allowed.

PR #13816, opened 2016-07-13
[BUGFIX beta] Ability to set a query param to a string of "null" when default value is null

webark:
If a default query parameter is null, we should leave it as `null` and not switch it to a string value of `"null"`.

This came up when in a controller we had set a default value for a query parameter of `search` to be `null`. Then if we tried to do a search for `"null"` it would not perform the search since it was casting the `null` value to a string of `"null"` for the default, and failed the difference check between the new query param and the default.

Didn't see any pertinent tests in the local package tests, so I added a test to the general ember routing query param test.

### DIFF
--- a/packages/ember-routing/lib/system/route.js
+++ b/packages/ember-routing/lib/system/route.js
@@ -411,6 +411,8 @@ let Route = EmberObject.extend(ActionHandler, Evented, {
     // to a certain query param.
     if (defaultValueType === 'array') {
       return JSON.stringify(value);
+    } else if (defaultValueType === 'null' && value === null) {
+      return null;
     }
     return `${value}`;
   },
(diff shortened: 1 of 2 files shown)


webark:
Ahh I see where the tests are now! 😛

webark:
Wondering if anything could be done to get this merged in?

heat:
Is it the desirable behavior ? I wondering, if you really want a null value you need to just remove the parameter, right ?

webark:
@heat we came across it when someone named something in the system as "null" and ran a search for "null". Ember ended up removing that query paramater due to the casting that it does.

webark:
I think due to nathan's future work with serializing parameters, and the shifting around of this issue, and due to the fact that it's sad dormant for 6 months, I'm going to go ahead and close it as "unmerged".

PR #13817, opened 2016-07-13
[FEATURE ember-application-engines] Merge route serializers into engines work

trentmwillis:
Reference: https://github.com/dgeb/ember-engines/issues/156.

This PR scopes the route-serializer work to Engines only and merges them into the same feature flag. Also has a small fix to get tests working with Glimmer.

cc @dgeb @rwjblue @nathanhammond

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -15,10 +15,6 @@ for a detailed explanation.
   Implements RFC https://github.com/emberjs/rfcs/pull/38, adding support for
   routable components.
 
-* `ember-route-serializers`
-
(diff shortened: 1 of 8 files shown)


rwjblue:
I believe the `service:-glimmer-environment` stuff should be removed from the `setupApplicationRegistry` function, and moved into the `setupEngineRegistry` one (instead of duplicating).

trentmwillis:
@rwjblue seems you were correct. PR has been updated.

rwjblue:
Thank you @trentmwillis!

PR #13818, opened 2016-07-14
[FEATURE ember-application-engines] Enable loading and error substates for mounts

trentmwillis:
Reference: https://github.com/dgeb/ember-engines/issues/103

One open question, how shold substates work when used in conjunction with `resetNamespace` (with either `mount` or `route`)? In other words if I have:

``` js
Router.map(function() {
  this.route('news', function() {
    this.{mount|route}('blog', { resetNamespace: true });
  });
});
```

Should the path to the substate then be: `blog_loading` or `news.blog_loading`? I feel like it should be `blog_loading` but in the current checks the namespace is always applied ([see here](https://github.com/emberjs/ember.js/blob/master/packages/ember-routing/lib/system/router.js#L1021)), so it will look for `news.blog_loading` even if `resetNamespace` is applied. Seems like maybe this a bug? If so, I can fix that while I'm in here.

cc @rwjblue @dgeb @nathanhammond

### DIFF
--- a/packages/ember-routing/lib/system/dsl.js
+++ b/packages/ember-routing/lib/system/dsl.js
@@ -190,6 +190,12 @@ if (isEnabled('ember-application-engines')) {
       }
     }
 
+    if (this.enableLoadingSubstates) {
+      let dummyErrorRoute = `/_unused_dummy_error_path_route_${name}/:error`;
+      createRoute(this, `${name}_loading`, { resetNamespace: options.resetNamespace });
+      createRoute(this, `${name}_error`, { path: dummyErrorRoute });
+    }
+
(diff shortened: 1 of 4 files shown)


rwjblue:
@trentmwillis - Yes, I agree that this is a bug.

dgeb:
Yep, I agree that it's a bug as well. It's not specific to engines, so let's address this in a separate PR.

@trentmwillis given that, are you comfortable removing the "WIP" label so we can land this?

trentmwillis:
I need to make some minor adjustments, but will do that this evening.

trentmwillis:
@dgeb updated, should be good to merge once tests finish.

PR #13820, opened 2016-07-14
[FEATURE ember-application-engines] Introduce mount keyword.

dgeb:
The `{{mount}}` keyword lets you embed a routeless engine in a template.

Mounting an engine will cause an instance to be booted and its 
`application` template to be rendered.

### DIFF
--- a/packages/ember-htmlbars/lib/keywords/mount.js
+++ b/packages/ember-htmlbars/lib/keywords/mount.js
@@ -1,0 +1,152 @@
+/**
+@module ember
+@submodule ember-templates
+*/
+
+import ViewNodeManager from 'ember-htmlbars/node-managers/view-node-manager';
+import RenderEnv from 'ember-htmlbars/system/render-env';
+import { assert } from 'ember-metal/debug';
+import { getOwner, setOwner } from 'container/owner';
+import { isOutletStable } from './outlet';
+import { childOutletState } from './render';
+
+/**
+  The `{{mount}}` helper lets you embed a routeless engine in a template.
+
+  Mounting an engine will cause an instance to be booted and its `application`
+  template to be rendered.
+
+  For example, the following template mounts the `ember-chat` engine:
+
+  ```handlebars
+  {{! application.hbs }}
+  {{mount "ember-chat"}}
+  ```
+
+  Currently, the engine name is the only argument that can be passed to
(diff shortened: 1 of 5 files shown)


packages/ember-htmlbars/lib/keywords/mount.js
```diff
@@ -0,0 +1,207 @@
...
+  `{{mount}}`.
+
+  @method mount
+  @for Ember.Templates.helpers
+  @category ember-application-engines
+  @public
+*/
+export default {
+  willRender(renderNode, env) {
+    if (env.view.ownerView._outlets) {
```
rwjblue: When is this not present?  `{{outlet` doesn't have a guard here...

packages/ember-htmlbars/tests/integration/mount_test.js
```diff
@@ -0,0 +1,129 @@
...
+    setup() {
+      commonSetup();
+    },
+
+    teardown() {
+      commonTeardown();
+    }
+  });
+
+  QUnit.test('asserts that only a single param is passed', function(assert) {
```
rwjblue: I think you will need to change these to use `import test from 'skip-if-glimmer';` for it to pass on Travis.

packages/ember-htmlbars/lib/keywords/mount.js
dgeb: This guard seems to be necessary for the `mount` tests in this PR to pass. Curiously, it is no longer required in the addon's tests (on `master`, anyway).

packages/ember-htmlbars/lib/keywords/mount.js
```diff
@@ -0,0 +1,179 @@
...
+  }
+  return true;
+}
+
+function lookupEngineController(engineInstance) {
+  return engineInstance.lookup('controller:application');
+}
+
+function lookupEngineView(engineInstance, ownerView) {
+  let engineView = engineInstance.lookup('view:application') || engineInstance.lookup('view:toplevel');
```
rwjblue: We should no longer look for `view:application` (support for using/looking up a `view` has been removed in canary).

dgeb: Ok, just removed the `lookup` of `view:application` 👍

PR #13825, opened 2016-07-14
[GLIMMER2] Fix spooky {{render}}/route functionality

chadhietala:
This is carry over from things that were not removed in 2.0. We really should make it a point to deprecate this spooky functionality post-glimmer release. This will to make sure we do not carry this functionality for another JavaScript half-life.

### DIFF
--- a/packages/ember-glimmer/lib/renderer.js
+++ b/packages/ember-glimmer/lib/renderer.js
@@ -6,10 +6,11 @@ import { CURRENT_TAG, UNDEFINED_REFERENCE } from 'glimmer-reference';
 const { backburner } = run;
 
 class DynamicScope {
-  constructor({ view, controller, outletState, isTopLevel, targetObject }) {
+  constructor({ view, controller, outletState, rootOutletState, isTopLevel, targetObject }) {
     this.view = view;
     this.controller = controller;
     this.outletState = outletState;
+    this.rootOutletState = rootOutletState;
     this.isTopLevel = isTopLevel;
     this.targetObject = targetObject;
   }
@@ -132,11 +133,13 @@ class Renderer {
     let env = this._env;
     let self = new RootReference(view);
     let controller = view.outletState.render.controller;
+    let ref = view.toReference();
     let dynamicScope = new DynamicScope({
       view,
       controller,
       targetObject: controller,
-      outletState: view.toReference(),
+      outletState: ref,
+      rootOutletState: ref,
       isTopLevel: true
     });
 
@@ -161,6 +164,7 @@ class Renderer {
       // instance
       targetObject: view,
       outletState: UNDEFINED_REFERENCE,
+      rootOutletState: UNDEFINED_REFERENCE,
       isTopLevel: true
     });
 
--- a/packages/ember-glimmer/lib/syntax/render.js
+++ b/packages/ember-glimmer/lib/syntax/render.js
@@ -2,6 +2,7 @@ import { ArgsSyntax, StatementSyntax } from 'glimmer-runtime';
 import { ConstReference, isConst } from 'glimmer-reference';
 import { assert } from 'ember-metal/debug';
 import { RootReference } from '../utils/references';
+import { generateControllerFactory } from 'ember-routing/system/generate_controller';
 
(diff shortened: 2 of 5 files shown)


packages/ember-glimmer/lib/renderer.js
```diff
@@ -137,6 +138,7 @@ class Renderer {
       controller,
       targetObject: controller,
       outletState: view.toReference(),
+      rootOutletState: view.toReference(),
```
chancancode: Could do

```
let thing = view.toReference();

... 


{
  outletState: thing,
  rootOutletState: thing
}
```

packages/ember-glimmer/lib/syntax/render.js
```diff
@@ -81,7 +83,18 @@ class AbstractRenderManager {
 class SingletonRenderManager extends AbstractRenderManager {
   create(definition, args, dynamicScope) {
     let { name, env } = definition;
-    let controller = env.owner.lookup(`controller:${name}`);
+    let controller = env.owner.lookup(`controller:${name}`) || generateControllerFactory(env.owner, name);
+
+    if (dynamicScope.outletState) {
```
chancancode: Does the non-singleton render (`{{render "foo" someModel}}`) not support this feature?

chancancode: While you are here – @krisselden noticed a bug: we never torn down the controller. The fix is easy, you just implement getDestructor in the `AbstractRenderManager` to destroy the controller. Can you add a test for it?

packages/ember-glimmer/lib/syntax/render.js
```diff
@@ -81,7 +83,18 @@ class AbstractRenderManager {
...
+      let orphan = dynamicScope.outletState.get('__ember_orphans__');
```
chancancode: Pretty sure this needs to be `dynamicScope.rootOutletState` to work correctly, since only the top-level has `__ember_orphans__`. Also, instead of using the generic code and making an intermediate reference, I think I would rather add a specialized `getOrphan` to the root outlet state reference (and only to the root).

packages/ember-glimmer/lib/views/outlet.js
```diff
@@ -25,6 +25,39 @@ class OutletStateReference {
...
+
+    return currentOutletState;
+  }
+
+  get(name) {
+    let currentOutletState = this.inner.value();
+    currentOutletState.outlets[name] = this.orphanState;
+    this.orphanState.wasUsed = true;
+    return new ChildOutletStateReference(this, name);
+  }
```
chancancode: Do we actually support nested outlets inside an orphan outlet? Do we have a test for it?

packages/ember-glimmer/lib/views/outlet.js
```diff
@@ -25,6 +25,39 @@ class OutletStateReference {
...
+    this.name = name;
+  }
+
+  value() {
+    let currentOutletState = this.inner.value();
+    let orphanState = this.orphan.get(this.name).value();
+
+    if (!orphanState.wasUsed) {
+      currentOutletState.outlets[this.name] = orphanState;
+      orphanState.wasUsed = true;
```
chancancode: We don't want to do these kind of mutations on a reference – can we move this mutation into the caller? (the code in Manager#create)

packages/ember-glimmer/lib/renderer.js
chadhietala: I actually don't know if `rootOutletState` is needed. It's not used anywhere in this PR.

chancancode: I **think** that means the tests are not testing deep enough nesting – if the __orphan__ thing is only available at the root, wouldn't it mean you need to got through that?

packages/ember-glimmer/lib/views/outlet.js
chancancode: I **think** the child outlet of an orphan is not "special", it should be just doing the normal `get` (should probably share that code with a common super class). If it's important that we set the `wasUsed` in all the places, we should move that to `OutletManager.create`

chancancode:
👏

PR #13831, opened 2016-07-17
[FEATURE ember-computed-includes] Creates includes macro for Ember.computed

cursedquail:
Several times I've needed a computed property to find out if an array contains a string, or a number. 

``` javascript
  let Hamster = Ember.Object.extend({
    hasABanana: Ember.computed.contains('possessions', 'banana')
  });

  let hamster = Hamster.create();

  hamster.get('hasABanana'); // false
  hamster.set('possessions', ['orange']);
  hamster.get('hasABanana'); // false
  hamster.set('possessions', ['banana']);
  hamster.get('hasABanana'); // true
```

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -59,3 +59,21 @@ to stay in line with ES standards (see [RFC](https://github.com/emberjs/rfcs/blo
 
 Expose a simple mechanism for test tooling to determine if all foreign async has been
 handled before continueing the test. Replaces the intimate API `Ember.Test.waiters` (with a deprecation).
+
+* `ember-computed-includes`
+
+  Introduces a new macro for Ember.computed to detect if an array includes a given element. Example:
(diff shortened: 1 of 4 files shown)


sandstrom:
@NLincoln Is there a reason this PR was closed? I think it's an awesome addition, though I would change it from `Ember.computed.contains` to `Ember.computed.includes` to better align with ES6/7.

I'd say out of all computed properties (that aren't using the macros) probably ~1/5 are simple includes-tests. So this would be very welcome!

```js
isAvailable: function() {
  return this.get('someList').includes(this.get('someKey'));
}.property('someList', 'someKey'),
```

locks:
@sandstrom this can easily be provided by an addon, and new APIs are required to go through the RFC process. So I recommend that path! :)

PR #13835, opened 2016-07-18
[Glimmer2] Fix some tests related input

chadhietala:


packages/ember-views/lib/mixins/text_support.js
```diff
@@ -196,9 +197,12 @@ export default Mixin.create(TargetActionSupport, {
   },
 
   _elementValueDidChange() {
-    // Using readDOMAttr will ensure that HTMLBars knows the last
-    // value.
-    set(this, 'value', this.readDOMAttr('value'));
+    if (isEnabled('ember-glimmer')) {
```
krisselden: curious, why did you need to guard this? what case does this break in htmlbars?

packages/ember-testing/lib/helpers/fill_in.js
```diff
@@ -11,10 +10,10 @@ export default function fillIn(app, selector, contextOrText, text) {
   $el = app.testHelpers.findWithAssert(selector, context);
   el = $el[0];
   focus(el);
-  run(() => {
```
krisselden: we should cleanup any other test helpers that are nesting run loops in the wait one.

chadhietala: I'll follow up with that.

chadhietala: See #13836

PR #13839, opened 2016-07-19
[FEATURE ember-improved-instrumentation] Basic actions

asakusuma:
Ensure basic actions, not just closure actions, are instrumented

### DIFF
--- a/packages/ember-glimmer/tests/integration/helpers/element-action-test.js
+++ b/packages/ember-glimmer/tests/integration/helpers/element-action-test.js
@@ -9,6 +9,8 @@ import { removeAt } from 'ember-runtime/mixins/mutable_array';
 
 import ActionManager from 'ember-views/system/action_manager';
 import jQuery from 'ember-views/system/jquery';
+import isEnabled from 'ember-metal/features';
+import { subscribe, reset } from 'ember-metal/instrumentation';
 
 function getActionAttributes(element) {
   let attributes = element.attributes;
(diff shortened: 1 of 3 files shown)


packages/ember-glimmer/tests/integration/helpers/element-action-test.js
```diff
@@ -25,6 +27,54 @@ function getActionAttributes(element) {
...
+          foo() {}
+        }
+      });
+
+      this.registerComponent('example-component', {
+        ComponentClass: ExampleComponent,
+        template: '<button {{action "foo" "bar"}}>Click me</button>'
+      });
+
+      let subscriber = subscribe('interaction.ember-action', {
```
rwjblue: I think this and the teardown should be done in module setup/teardown to ensure we don't leak subscribers if this errors.

packages/ember-glimmer/tests/integration/helpers/element-action-test.js
```diff
@@ -25,6 +27,54 @@ function getActionAttributes(element) {
...
+
+      this.runTask(() => {
+        this.$('button').click();
+      });
+
+      this.assert.equal(subscriberCallCount, 1, 'subscriber has been called 1 time');
+      this.assert.equal(subscriberPayload.name, 'foo', 'subscriber called with correct name');
+      this.assert.equal(subscriberPayload.args[0], 'bar', 'subscriber called with correct args');
+
+      unsubscribe(subscriber);
```
rwjblue: Can you move to modules `afterEach` / `teardown`?

asakusuma:
@rwjblue moved to `reset` instead of `unsubscribe`, and moved it to `teardown`. Can you start travis? For whatever reason it's not starting on it's own.

rwjblue:
Ya, Travis has been taking a few minutes to schedule a job (and set the "pending" commit status).  Looks like it is running now.

asakusuma:
Ah I see. Did not realize it could be that async. Looks like things are green.

PR #13842, opened 2016-07-19
Register `component-lookup:main` for all engines.

dgeb:
This registration is required for applications _and_ engines.

Follow up to https://github.com/emberjs/ember.js/pull/13841


rwjblue:
Thanks @dgeb!

PR #13843, opened 2016-07-19
Make `ember-template-compiler` unneeded in ember.prod.js.

rwjblue:
- Move `template` method to `ember-templates`. This is currently exposed
  as `Ember.HTMLBars.template` and `Ember.Handlebars.template` (we should probably figure out a more generic
  place for this in both globals and modules land), and is the only piece of
  the `ember-template-compiler` that is actually needed at runtime when the
  `dist/ember-template-compiler.js` file is not present.
- Move globals based template boostrapping code into
  `ember-template-compiler` (it is only used/needed when the template
  compiler is present and we would prefer not to run this code in
  production builds where the real template compiler is not present anyways).

This paves the way to removing `ember-template-compiler` package from `ember.debug.js` and `ember.prod.js` (without changing any public API's or expectations).

/cc @krisselden


rwjblue:
Pending updates in emberjs-build: https://github.com/emberjs/emberjs-build/pull/169

rwjblue:
Also, discussed with @krisselden and we are going to move `Ember.HTMLBars.template` into the respective `ember-htmlbars` and `ember-glimmer` packages then have `ember-templates` reexport the appropriate one.

chadhietala:
LGTM

rwjblue:
@krisselden - Updated to keep the engine specific implementations of `template` in the engine packages themselves, and grab them from the `index.js` of the engine package.

PR #13845, opened 2016-07-19
[Glimmer2] tests for html-safe and render

chadhietala:
/cc @krisselden @rwjblue 

Pretty sure we want to just use the SafeString from HandleBars, however it is unclear how we should import it sense it was a proxy import in HTMLBars.


rwjblue:
@chadhietala - This is good to me.  Can you make https://github.com/emberjs/ember.js/blob/master/packages/ember-htmlbars/lib/utils/string.js a symlink to the `ember-glimmer` one, and expose it to `Ember.Handlebars` and `Ember.HTMLBars` in `ember-templates`?

packages/ember-glimmer/lib/utils/string.js
```diff
@@ -0,0 +1 @@
+../../../ember-htmlbars/lib/utils/string.js
```
rwjblue: I think this should go the other way since in the long term we are going to `rm -rf packages/ember-htmlbars`.

rwjblue:
Overall this look good (one comment/tweak), can you confirm that `escapeExpression` is exposed by `glimmer`'s build?

chadhietala:
What I ended up doing is moving the string module into `ember-glimmer` and also inlining the `SafeString` class and `escapExpression` functions as they came from HTMLBars. `SafeString` in HTMLBars is just a proxy to Handlebars.

packages/ember-templates/lib/index.js
```diff
@@ -21,6 +29,39 @@ Ember.TextField = TextField;
 Ember.TextArea = TextArea;
 Ember.LinkComponent = LinkComponent;
 
+if (Ember.HTMLBars) {
+  Ember.HTMLBars.SafeString = SafeString;
+} else {
+  Ember.HTMLBars = { SafeString };
+}
+
+if (Ember.HandlerBars) {
```
rwjblue: Should be `Ember.Handlebars` (a few places below are typo'ed too).

packages/ember-templates/lib/index.js
```diff
@@ -21,6 +29,39 @@ Ember.TextField = TextField;
 Ember.TextArea = TextArea;
 Ember.LinkComponent = LinkComponent;
 
+if (Ember.HTMLBars) {
+  Ember.HTMLBars.SafeString = SafeString;
```
rwjblue: Can you move these `Ember.HTMLBars` exports to `ember-templates/compat` (we already normalize and ensure `Ember.HTMLBars` is present there)?

packages/ember-templates/lib/index.js
rwjblue: Can you also move these `Ember.Handlebars` exports to `ember-templates/compat`?

packages/ember-templates/lib/index.js
```diff
@@ -21,6 +29,39 @@ Ember.TextField = TextField;
...
+    Ember.HandlerBars.Utils.escapeExpression = escapeExpression;
+  } else {
+    Ember.HandlerBars.Utils = { escapeExpression };
+  }
+} else {
+  Ember.HandlerBars = { SafeString, Utils: { escapeExpression } };
+}
+
+if (Ember.String) {
+  Ember.String.Utils = { htmlSafe };
```
rwjblue: This should be `Ember.String.htmlSafe` (no `Utils`).  Can you also import `String` from `ember-runtime`?

packages/ember-templates/lib/index.js
```diff
@@ -21,6 +29,39 @@ Ember.TextField = TextField;
...
+} else {
```
rwjblue: No need to guard here, we should import `String`

packages/ember-views/lib/mixins/action_support.js
```diff
@@ -132,7 +132,9 @@ export default Mixin.create({
       if (!shouldBubble) { return; }
     }
 
-    if (target = get(this, 'target')) {
+    target = get(this, 'target') || get(this, '_targetObject');
```
rwjblue: Is this change related to the others?

chadhietala: Not really I can PR separately. Just trying to reduce CI cycles 😁

chadhietala:
Had to do some reorg and deletion of some similar work. Also moved the action test to #13850.

packages/ember-templates/lib/compat.js
```diff
@@ -1,7 +1,46 @@
...
+      return SafeString;
+    }
+  });
+} else {
+  EmberHandlebars.SafeString = SafeString;
+}
+
+EmberHTMLBars.SafeString = SafeString;
 EmberHTMLBars.template = EmberHandlebars.template = template;
+EmberHTMLBarsUtils.escapeExpression = EmberHandleBarsUtils.escapeExpression = escapeExpression;
```
rwjblue: I do not believe that we ever exposed this previously on `Ember.HTMLBars`, can you confirm and if we haven't fixup?

chadhietala: Fixed.

rwjblue:
LGTM

PR #13847, opened 2016-07-20
[FEATURE ember-improved-instrumentation] Pass link-to component

asakusuma:
Include the link-to component in the `interaction.link-to` event payload. Useful including metadata in tracking/analytics associated with clicking a link.


krisselden:
We shouldn't create the payload outside of the feature flag

krisselden:
Also, I would rather have a general interaction event for tracking from the dispatcher and/or a way to get Ember context from an event target element.

asakusuma:
@krisselden looking into that

krisselden:
we should do an RFC

locks:
Closed for RFC.

PR #13849, opened 2016-07-20
Pave the way for shipping only glimmer engine.

rwjblue:
Update build system to allow excluding `ember-htmlbars` and its dependencies if `ember-glimmer` feature flag is `true`.

### DIFF
--- a/lib/packages.js
+++ b/lib/packages.js
@@ -15,23 +15,9 @@ var packages = {
   'ember-template-compiler': {
     trees: null,
     templateCompilerOnly: true,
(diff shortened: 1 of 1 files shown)


rwjblue:
OK, I have removed all the extraneous changes from this PR.  Now this only contains the packaging changes required to make this work.  

It is still not possible to remove `ember-htmlbars` (by changing `ember-glimmer` flag to `true` in `features.json`), but we will continue chipping away at the failures in separate more focused PR's.

PR #13851, opened 2016-07-21
[GLIMMER] Migrate remaining `style` attribute tests from HTMLBars.

rwjblue:


rwjblue:
Seems like a legit failure for the glimmer side on IE9:

screenshot

rwjblue:
Changed the test failing in Glimmer to be run only in HTMLBars for now.

@chadhietala - That test should probably be added to the burn down issue of still pending tests.

chadhietala:
Added to "The List"

PR #13853, opened 2016-07-21
[DOC beta] Mark `LinkComponent#activeClass` as public.

Serabe:
`activeClass` is mark as private in `Ember.LinkComponent` though in the
documentation suggests to use it for modifying the class added when link is
active.


Serabe:
Fixed!

mixonic:
Thanks @Serabe!

PR #13858, opened 2016-07-22
[Glimmer2] Backtracking Rerender

chadhietala:
Depends on https://github.com/tildeio/glimmer/pull/223.

Backtracking re-render refers to a scenario where, in the middle of the rendering process, you have modified something that has already been rendered. This behavior has always been unreliable, but was partially supported with a deprecation (since Ember 1.13). In Glimmer 2, while the extra re-render is relatively cheap, the extra book-keeping to detect a backtracking `set` is not. One of the wins of the Glimmer 2 system is that it does not need to eagerly setup observers to track changes. If you re-render a value in a single transaction we now throw an error.

The feature previously known as two-waying flushing. 🚽

### DIFF
--- a/features.json
+++ b/features.json
@@ -10,6 +10,7 @@
     "ember-runtime-enumerable-includes": true,
     "ember-string-ishtmlsafe": true,
     "ember-testing-check-waiters": true,
-    "ember-metal-weakmap": null
+    "ember-metal-weakmap": null,
+    "ember-glimmer-allow-backtracking-rerender": null
   }
 }
--- a/packages/ember-glimmer-template-compiler/lib/plugins/transform-attrs-into-args.js
+++ b/packages/ember-glimmer-template-compiler/lib/plugins/transform-attrs-into-args.js
@@ -13,7 +13,7 @@
   to
 
   ```handlebars
- {{foo.bar}}
+ {{@foo.bar}}
   ```
 
   as well as `{{#if attrs.foo}}`, `{{deeply (nested attrs.foobar.baz)}}` etc
@@ -38,7 +38,10 @@ TransformAttrsToProps.prototype.transform = function TransformAttrsToProps_trans
   traverse(ast, {
     PathExpression(node) {
       if (node.parts[0] === 'attrs') {
-        return b.path(node.original.substr(6));
+        let path = b.path(node.original.substr(6));
+        path.original = `@${path.original}`;
+        path.data = true;
+        return path;
       }
     }
   });
--- a/packages/ember-glimmer/lib/renderer.js
+++ b/packages/ember-glimmer/lib/renderer.js
@@ -4,6 +4,20 @@ import { setHasViews } from 'ember-metal/tags';
 import { CURRENT_TAG, UNDEFINED_REFERENCE } from 'glimmer-reference';
 import fallbackViewRegistry from 'ember-views/compat/fallback-view-registry';
 import { assert } from 'ember-metal/debug';
+import _runInTransaction from 'ember-metal/transaction';
+import isEnabled from 'ember-metal/features';
+
+let runInTransaction;
+
+if (isEnabled('ember-glimmer-detect-backtracking-rerender') ||
+    isEnabled('ember-glimmer-allow-backtracking-rerender')) {
+  runInTransaction = _runInTransaction;
+} else {
+  runInTransaction = callback => {
+    callback();
+    return false;
+  };
+}
 
 const { backburner } = run;
 
@@ -146,9 +160,22 @@ class Renderer {
       isTopLevel: true
     });
 
-    env.begin();
-    let result = view.template.asEntryPoint().render(self, env, { appendTo: target, dynamicScope });
-    env.commit();
+    let result, shouldReflush, afterRender, didChangeAfterRender;
+
+    let callback = () => {
+      result = view.template.asEntryPoint().render(self, env, { appendTo: target, dynamicScope });
+    };
+
+    let rerender = () => result.rerender({ alwaysRevalidate: shouldReflush });
+
+    do {
+      env.begin();
+      shouldReflush = runInTransaction(callback);
+      afterRender = CURRENT_TAG.value();
+      env.commit();
+      didChangeAfterRender = !CURRENT_TAG.validate(afterRender);
+      callback = rerender;
+    } while (shouldReflush || didChangeAfterRender);
 
     this._scheduler.registerView(view);
 
@@ -171,9 +198,22 @@ class Renderer {
       isTopLevel: true
     });
 
-    env.begin();
-    let result = view.template.asEntryPoint().render(self, env, { appendTo: target, dynamicScope });
-    env.commit();
+    let result, shouldReflush, afterRender, didChangeAfterRender;
+
+    let callback = () => {
+      result = view.template.asEntryPoint().render(self, env, { appendTo: target, dynamicScope });
+    };
+
+    let rerender = () => result.rerender({ alwaysRevalidate: shouldReflush });
+
+    do {
+      env.begin();
+      shouldReflush = runInTransaction(callback);
+      afterRender = CURRENT_TAG.value();
+      env.commit();
+      didChangeAfterRender = !CURRENT_TAG.validate(afterRender);
+      callback = rerender;
+    } while (shouldReflush || didChangeAfterRender);
 
     this._scheduler.registerView(view);
 
@@ -185,11 +225,21 @@ class Renderer {
   }
 
   rerender(view) {
-    let env = this._env;
+    let { _env: env } = this;
 
-    env.begin();
-    (view['_renderResult'] || this._root['_renderResult']).rerender();
-    env.commit();
+    let renderResult = view['_renderResult'] || this._root['_renderResult'];
+    let shouldReflush = false;
+    let afterRender, didChangeAfterRender;
+
+    let callback = () => renderResult.rerender({ alwaysRevalidate: shouldReflush });
+
+    do {
+      env.begin();
+      shouldReflush = runInTransaction(callback);
+      afterRender = CURRENT_TAG.value();
+      env.commit();
+      didChangeAfterRender = !CURRENT_TAG.validate(afterRender);
+    } while (shouldReflush || didChangeAfterRender);
   }
 
   remove(view) {
--- a/packages/ember-glimmer/lib/utils/references.js
+++ b/packages/ember-glimmer/lib/utils/references.js
@@ -1,6 +1,7 @@
 import { get } from 'ember-metal/property_get';
 import { set } from 'ember-metal/property_set';
 import { tagFor } from 'ember-metal/tags';
+import { didRender } from 'ember-metal/transaction';
 import symbol from 'ember-metal/symbol';
 import { CURRENT_TAG, CONSTANT_TAG, VOLATILE_TAG, ConstReference, DirtyableTag, UpdatableTag, combine, isConst } from 'glimmer-reference';
 import { ConditionalReference as GlimmerConditionalReference, NULL_REFERENCE, UNDEFINED_REFERENCE } from 'glimmer-runtime';
@@ -86,6 +87,41 @@ export class RootReference extends ConstReference {
   }
 }
 
+let TwoWayFlushDetectionTag;
+
+if (isEnabled('ember-glimmer-detect-backtracking-rerender') ||
+    isEnabled('ember-glimmer-allow-backtracking-rerender')) {
+  TwoWayFlushDetectionTag = class {
+    constructor(tag, key, ref) {
+      this.tag = tag;
+      this.parent = null;
+      this.key = key;
+      this.ref = ref;
+    }
+
+    value() {
+      return this.tag.value();
+    }
+
+    validate(ticket) {
+      let { parent, key } = this;
+
+      let isValid = this.tag.validate(ticket);
+
+      if (isValid && parent) {
+        didRender(parent, key, this.ref);
+      }
+
+      return isValid;
+    }
+
+    didCompute(parent) {
+      this.parent = parent;
+      didRender(parent, this.key, this.ref);
+    }
+  };
+}
+
 export class PropertyReference extends CachedReference { // jshint ignore:line
   constructor(parentReference, propertyKey) {
     super();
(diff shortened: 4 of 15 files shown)


packages/ember-metal/lib/meta.js
```diff
@@ -2,11 +2,17 @@
 // Remove "use strict"; from transpiled module until
 // https://bugs.webkit.org/show_bug.cgi?id=138038 is fixed
 
+// This is less than ideal, but ensures that `runInDebug` is
+// populated.
+import require from 'require';
+require('ember-debug');
```
chadhietala: This sucks.

packages/ember-metal/lib/meta.js
```diff
@@ -46,6 +47,24 @@ let members = {
   tag: ownCustomObject
 };
 
+// This is less than ideal, but `runInDebug` is
+// not populated by the time we need it here. So
+// we create a function that has the same signature
+// as the real `runInDebug` so that it is stripped
+// from the production build. Unfornately, this
+// function will remain in production build.
+let runInDebug = (fn) => fn();
```
chadhietala: @chancancode I would like to bring this to your attention 😛

rwjblue: This confuses me.  When you say, it isn't populated, do you mean that it doesn't run the callback or that the import is undefined?

chadhietala: The function never calls the pass function. There is some race when `setDebugFunction` is [called](https://github.com/emberjs/ember.js/blob/28cedad1418bedd2dad44eb56668fa67b9769d86/packages/ember-debug/lib/index.js#L148) and when this module is required.

krisselden: @rwjblue this is the circular race that makes part of ember metal things not able to use debug stuff. What I was working toward moving it so metal imports ember debug.

krisselden: It's another strongly connected component with a top level side effect

chadhietala: @krisselden Is there a better path that I can take right now?

rwjblue: @krisselden - Yes, I recall discussing this with you. 

I think we decided the right path was to just ensure ember-debug was always present and swap the underlying implementation based on debug vs production build. I will try to work on this today...

packages/ember-glimmer-template-compiler/lib/plugins/transform-attrs-into-args.js
```diff
@@ -38,7 +38,10 @@ TransformAttrsToProps.prototype.transform = function TransformAttrsToProps_trans
   traverse(ast, {
     PathExpression(node) {
       if (node.parts[0] === 'attrs') {
-        return b.path(node.original.substr(6));
+        let path = b.path(node.original.substr(6));
+        path.original = `@${path.original}`;
```
rwjblue: I am concerned about the implications here, that users can start using the `@foo` syntax. We definitely should be preventing this until we have an RFC proposing it.

packages/ember-glimmer/tests/integration/components/curly-components-test.js
```diff
@@ -2001,6 +2002,7 @@ moduleFor('Components test: curly components', class extends RenderingTest {
       ComponentClass: Component.extend({
         value: null,
         pushDataUp: observer('value', function() {
+          window.billy = true;
```
rwjblue: O_O

packages/ember-metal/lib/transaction.js
```diff
@@ -0,0 +1,75 @@
...
+import { has } from 'require';
+import { deprecate, runInDebug } from 'ember-metal/debug';
+import isEnabled from 'ember-metal/features';
+
+let runInTransaction, didRender, assertNotRendered;
+
+if (isEnabled('ember-glimmer-allow-two-way-reflush')) {
+  let hasGlimmer = has('glimmer-reference');
+  let counter = 0;
+  let inTransaction = false;
```
rwjblue: Should this be tied to some other state? In general, it should be possible to have two applications sharing the same ember internal modules on the page at once. The way I read the use of this module level flag, it seems that it might be an issue?

packages/ember-glimmer-template-compiler/lib/plugins/transform-attrs-into-args.js
rwjblue: Discussed with @chancancode, we are going to do an AST transform that errors on the `@` syntax.  We'll just need to ensure it runs before the one you are modifying here, but that should be fine. 

Carry on... 😎

features.json
```diff
@@ -5,6 +5,7 @@
     "ember-libraries-isregistered": null,
     "ember-application-engines": true,
     "ember-glimmer": null,
+    "ember-glimmer-allow-two-way-reflush": true,
```
rwjblue: New features generally land as `false` initially, so we can confirm that both sides of the feature flagging works.  I suspect that this feature should actually be a development only feature (ala `mandatory-setter`), which is setup in `ember-cli-build.js` ([here](https://github.com/emberjs/ember.js/blob/master/ember-cli-build.js#L27)).

packages/ember-glimmer/tests/integration/components/curly-components-test.js
chancancode: This is very important for correctness 😂

features.json
chancancode: This is the thing I wanted to ask you 😄

1. The detailed reporting should be disabled on prod no matter what (`lastRenderedFrom` and friends – what @chadhietala wrapped in `runInDebug` currently)
2. By default, what is currently a deprecation in the code "You modified ${parts.join('.')} twice in a single render..." should be *an error* in development mode, with a more detailed message: it should describe the situation more clearly, that it is previously unreliable and very slow, and probably link to somewhere that tells you more including how to fix it. It should also mention that you can enable the feature with an addon, see (4) and (5).
3. By default, the `lastRendered` tracking should be off (what is currently in the "off" branches of "ember-glimmer-allow-two-way-reflush"
4. If the compatibility mode is on, in development mode, it should still emit the message as a deprecation.
5. If the compatibility mode is on, in production mode, it should work without a deprecation (which is the natural consequence of a deprecation I believe).

chancancode:
To be clear, this patch fixes two scenarios, one of which is completely harmless, one is very slow/difficult to support:

1. After rendering, the user did one or more `Ember.set` in the "post" (async) hooks (`didInsertElement`, `didRender`, `afterRender` queue, etc) – this is fine (as long as it doesn't cause an infinite loop).
2. During rendering, the user did one or more `Ember.set` in a synchronous hook (`init`, `didReceiveAttr`, `willRender`, etc) *that affected one of the curlies that has already been rendered on the screen somewhere, directly or indirectly. This is very difficult and slow to support (and has been previously unreliable).

Example:

```hbs
{{foo}}
{{foo-bar}}
```

```js
App.FooBarComponent = Component.extend({
  willRender() {
    this.parentView.set('foo', 'lol'); // {{foo}} has already been rendered by now, and this changes its value
  }
});
```

To support this, we need to carefully track all the already rendered curlies, which almost defeats the entire purpose of the project (to lower book-keeping cost during initial render by not setting up observers, etc).

Further, we have optimizations that skips entire sections of the screen if we are confident that nothing in there has changed. However, even that it would indeed produce the right result, we also loose the opportunity to do the "curlies tracking" in there to support this "backtracking" feature, which means in order to support it, we have to purposefully disable all of those optimizations for very good reason.

Hence, given that the feature is already deprecated *and previously unreliable anyway*, we agreed that we will try to make the slow-mode opt-in instead, and monitor how things goes during alpha testing and reconsider.

Also – the deprecation message on HTMLBars is not very accurate, and fires sometimes even when it is not strictly needed (i.e. not situation (2) here), and sometimes just plain wrong. It also includes the wrong information sometimes. We should try to fix those and back port the fixes if we can.

packages/ember-metal/lib/transaction.js
chadhietala: I think this is ok because we reset this after the transaction. You currently cannot have concurrent renders so this should just work.

chadhietala:
I would like some feedback on this second commit. The TL;DR is:

| Flag       | assert           | warn  |
| ------------- |:-------------:|:-----:|
| ember-glimmer-detect-two-way-reflush | X | |
| ember-glimmer-allow-two-way-reflush |  |X|

However, this is a bit tricky because it requires both build time and runtime flagging. This basically means we need 2 builds to support a "strict" and "loose" mode. This way you can get a build that will throw in dev and what that just warns in dev. I think we need to beef up the error message and say that there is a "loose" mode build and a link to some documentation that has more details.

bin/run-tests.js
```diff
@@ -104,7 +104,7 @@ function generateEachPackageTests() {
 function generateBuiltTests() {
   // Container isn't publicly available.
   // ember-testing/ember-debug are stripped from prod/min.
-  var common = 'skipPackage=container,ember-testing,ember-debug';
+  var common = 'skipPackage=container,ember-testing,ember-debug,internal-test-helpers';
```
chadhietala: Will remove.

packages/ember-glimmer/lib/renderer.js
```diff
@@ -4,6 +4,30 @@ import { setHasViews } from 'ember-metal/tags';
...
+
+if (isEnabled('ember-glimmer-detect-two-way-reflush') ||
+    isEnabled('ember-glimmer-allow-two-way-reflush')) {
+  runInTransaction = _runInTransaction;
+} else {
+  runInTransaction = callback => {
+    callback();
+    return false;
+  };
+}
```
chancancode: duplicated code o_0

chadhietala: Eh relic before I found out that we could have multiple `isEnabled` calls.

packages/ember-glimmer/lib/renderer.js
```diff
@@ -111,7 +135,14 @@ class Scheduler {
...
     if (view) {
-      view.renderer.rerender(view);
+      if (isEnabled('ember-glimmer-detect-two-way-reflush') ||
+          isEnabled('ember-glimmer-allow-two-way-reflush')) {
+        let shouldReflush;
+        let callback = () => view.renderer.rerender(view, { alwaysRevalidate: true });
+        do { shouldReflush = runInTransaction(callback); } while (shouldReflush);
+      } else {
+        view.renderer.rerender(view);
+      }
```
chancancode: I *think* this change is not needed/wrong, need to check my branch to be sure. What's making me suspicious is that `view.renderer.rerender` doesn't actually take the `alwaysRevalidate` option, and it already does the `runInTransaction` stuff there, so this is probably a copy/paste/merge error

packages/ember-metal/lib/meta.js
```diff
@@ -46,6 +46,12 @@ let members = {
   tag: ownCustomObject
 };
 
+if (isEnabled('ember-glimmer-detect-two-way-reflush') ||
+    isEnabled('ember-glimmer-allow-two-way-reflush')) {
+  members.lastRendered = ownMap;
+  members.lastRenderedFrom = ownMap; // Not used in production
```
chancancode: `// FIXME: not used in production, remove me from prod builds`

packages/ember-metal/lib/property_events.js
```diff
@@ -108,6 +111,13 @@ function propertyDidChange(obj, keyName) {
   }
 
   markObjectAsDirty(m);
+
+  if (isEnabled('ember-glimmer-detect-two-way-reflush') ||
+      isEnabled('ember-glimmer-allow-two-way-reflush')) {
+    if (has('glimmer-runtime')) {
```
chancancode: seems bad to do this as a runtime check since this is very hot path; should probably just make sure we only enable the flags with glimmer, (or if that's not possible, at least make the functions no-op without glimmer and call it here unconditionally)

packages/ember-metal/lib/transaction.js
```diff
@@ -0,0 +1,82 @@
...
+            let ref = meta.readableLastRenderedFrom();
+            let parts = [];
+            let lastRef = ref[key];
+
+            while (lastRef && lastRef._propertyKey && lastRef._parentReference) {
+              parts.unshift(lastRef._propertyKey);
+              lastRef = lastRef._parentReference;
+            }
+
+            return `You modified ${parts.join('.')} twice in a single render. This was unreliable and slow in Ember 1.x and will be removed in Ember 3.0.`;
```
chancancode: TODO: fix this message (can you open an issue an assign it to me once merged?)

packages/ember-metal/lib/property_events.js
chancancode: If we made them no-ops, then we can also remove the stub of `runInTransacation` in the renderer

chancancode: I think the solution is either

1. in emberjs-build, make sure we hard code these to false unless glimmer is enabled, or
2. in application boot, if these flags is enabled, do an assert for `has('glimmer-runtime')`

Then the functions can be no-ops

rwjblue: Umm.  Why not check `isEnabled('ember-glimmer')` also?

packages/ember-glimmer/lib/renderer.js
chancancode: yeah, this change is a mistake, ✂️

packages/ember-glimmer-template-compiler/lib/plugins/transform-attrs-into-args.js
chancancode: @chadhietala can you follow up on that once this lands? (I think we want to do that before doing an alpha)

packages/ember-glimmer/lib/utils/references.js
```diff
@@ -86,17 +87,54 @@ export class RootReference extends ConstReference {
   }
 }
 
+class TwoWayFlushDetectionTag {
```
chancancode: This class is only used when the flags are enabled, so it can probably be flagged

chadhietala:
@chancancode 👍 'd in Slack

packages/ember-metal/lib/transaction.js
```diff
@@ -0,0 +1,84 @@
+import { meta as metaFor } from './meta';
+import { assert, runInDebug, deprecate } from 'ember-metal/debug';
+import isEnabled from 'ember-metal/features';
+
+let runInTransaction, didRender, assertNotRendered;
+
+if (isEnabled('ember-glimmer-detect-backtracking-rerender') ||
+    isEnabled('ember-glimmer-allow-backtracking-rerender')) {
+  assert('It appears you are trying to use the backtracking rerender without the "ember-glimmer" flag turned on. Please make sure that "ember-glimmer" is turned on.');
```
chancancode: , isEnabled('ember-glimmer'));

PR #13863, opened 2016-07-22
[BUGFIX] remove top level side effect from circular module reference

krisselden:
Removes top level side effect from circular module reference from stream to utils that causes race condition dependent on ordering of module load.


stefanpenner:
LGTM, is this bugfix beta?

rwjblue:
No, I don't think it fixes any bugs. But as we move around the exports the cycles and up breaking tests.

PR #13865, opened 2016-07-23
Fix imports to avoid ember-htmlbars.

rwjblue:


krisselden:
This is circular

krisselden:
Nevermind it's only the tests,  really should move the tests since they require the template layer

rwjblue:
Ya, in a subsequent PR the ember-views test for component is removed.

PR #13881, opened 2016-07-25
Migrate more ember-views tests to ember-glimmer structure.

rwjblue:


rwjblue:
OK, tracked down the issues with the failures on IE9 - IE11.

It came down to this "oddity" in IE9 - IE11:

```js
let element = document.createElement('div');
element.setAttribute('style', 'display: none;');
element.getAttribute('style'); // => "display: none;"
element.setAttribute('style', '');
element.getAttribute('style'); // => null
```

On other platforms (Chrome, FF, Safari, MS Edge):

```js
let element = document.createElement('div');
element.setAttribute('style', 'display: none;');
element.getAttribute('style'); // => "display: none;"
element.setAttribute('style', '');
element.getAttribute('style'); // => ''
```

The solution was to create a detect to see what the result of `element.getAttribute('style')` is after having it populated then removed and use that value in the assertion.

chadhietala:
These fail in Glimmer? We already special case style attributes by having a `StyleAttributeChangeList`. I suspect we can something special here to normalize this [here](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/environment.js#L61-L75).

rwjblue:
@chadhietala - Ya, they currently fail in glimmer.  I suspect we aren't doing anything special for `isVisible` property?

rwjblue:
The last round of issues is:

```js
// IE11+, Chrome, FireFox, and Safari
let elem = document.createElement('div');
elem.setAttribute('style', 'display: none;');
elem.attributes.length; // => 1
elem.setAttribute('style', '');
elem.attributes.length; // => 0
```

```js
// IE9 and IE10
let elem = document.createElement('div');
elem.setAttribute('style', 'display: none;');
elem.attributes.length; // => 1
elem.setAttribute('style', '');
elem.attributes.length; // => 1
```

---

At this point I am somewhat giving up on the best assertion mechanism of using `this.assertComponentElement`, and falling back to simply asserting `style` attribute is set to the proper value.

PR #13883, opened 2016-07-25
[GLIMMER2] Implement closure components

zackthehuman:
This PR implements the `{{component}}` helper (when used as a subexpression) in Glimmer 2. This enables "closure components".

The basic idea is that the `{{component}}` helper produces a `ComponentDefinition` that composes a component and a set of args to use when constructing it. If the incoming component is itself a closure component, then the args are merged according to closure component rules. The resulting definition will have a populated `args` property, which can be used in a component manager's `prepareArgs` hook.

Since parameters are being merged according to closure component rules, some of the rest/position parameter validation logic had to be moved around.

Some of the code here can be cleaned up to reduce duplication, but I wanted to get eyes on this before anything fancy happens.

### DIFF
--- a/packages/ember-glimmer/lib/helpers/component.js
+++ b/packages/ember-glimmer/lib/helpers/component.js
@@ -1,7 +1,143 @@
+import { CachedReference } from '../utils/references';
+import { CurlyComponentDefinition } from '../syntax/curly-component';
+import { EvaluatedArgs, EvaluatedNamedArgs, EvaluatedPositionalArgs } from 'glimmer-runtime';
+import { assert } from 'ember-metal/debug';
+import assign from 'ember-metal/assign';
+
+const CLOSURE_COMPONENT = 'ba564e81-ceda-4475-84a7-1c44f1c42c0e';
+
+export function isClosureComponent(object) {
+  return object && !!object[CLOSURE_COMPONENT];
+}
+
+class ClosureComponentDefinition extends CurlyComponentDefinition {
+  constructor(name, ComponentClass, template, args) {
+    super(...arguments);
+    this[CLOSURE_COMPONENT] = true;
+  }
+}
+
+export class ClosureComponentReference extends CachedReference {
+  static create(args, parentMeta, env) {
+    return new ClosureComponentReference(args, parentMeta, env);
+  }
+
+  constructor(args, parentMeta, env) {
+    super();
+    this.defRef = args.positional.at(0);
+    this.env = env;
+    this.tag = args.tag;
+    this.parentMeta = parentMeta;
+    this.args = args;
+  }
+
+  compute() {
+    // TODO: Figure out how to extract this because it's nearly identical to
+    // DynamicComponentReference::compute(). The only differences besides
+    // currying are in the assertion messages.
+    let { args, defRef, env, parentMeta } = this;
+    let nameOrDef = defRef.value();
+    let definition = null;
+
+    if (typeof nameOrDef === 'string') {
+      definition = env.getComponentDefinition([nameOrDef], parentMeta);
+      assert(`The component helper cannot be used without a valid component name. You used "${nameOrDef}" via (component "${nameOrDef}")`, definition);
+    } else if (isClosureComponent(nameOrDef)) {
+      definition = nameOrDef;
+    } else {
+      assert(
+        `You cannot create a component from ${nameOrDef} using the {{component}} helper`,
+        nameOrDef
+      );
+      return null;
+    }
+
+    let newDef = createCurriedDefinition(definition, args);
+
+    return newDef;
+  }
+}
+
+function createCurriedDefinition(definition, args) {
+  let curriedArgs = curryArgs(definition, args);
+
+  return new ClosureComponentDefinition(
+    definition.name,
+    definition.ComponentClass,
+    definition.template,
+    curriedArgs
+  );
+}
+
+function curryArgs(definition, newArgs) {
+  let { args, ComponentClass } = definition;
+  let { positionalParams } = ComponentClass;
+  
+  // The args being passed in are from the (component ...) invocation,
+  // so the first positional argument is actually the name or component
+  // definition. It needs to be dropped in order for any actual positional
+  // args to coincide with the ComponentClass's positionParams.
+
+  // For "normal" curly components this slicing is done at the syntax layer,
+  // but we don't have that luxury here.
+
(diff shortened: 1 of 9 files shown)


packages/ember-glimmer/lib/helpers/component.js
```diff
@@ -1,7 +1,143 @@
+import { CachedReference } from '../utils/references';
+import { CurlyComponentDefinition } from '../syntax/curly-component';
+import { EvaluatedArgs, EvaluatedNamedArgs, EvaluatedPositionalArgs } from 'glimmer-runtime';
+import { assert } from 'ember-metal/debug';
+import assign from 'ember-metal/assign';
+
+const CLOSURE_COMPONENT = 'ba564e81-ceda-4475-84a7-1c44f1c42c0e';
+
+export function isClosureComponent(object) {
+  return object && !!object[CLOSURE_COMPONENT];
```
krisselden: please check typeof object before keyed field load if this is called on any opaque value.

packages/ember-glimmer/lib/helpers/component.js
```diff
@@ -1,7 +1,143 @@
+import { CachedReference } from '../utils/references';
+import { CurlyComponentDefinition } from '../syntax/curly-component';
+import { EvaluatedArgs, EvaluatedNamedArgs, EvaluatedPositionalArgs } from 'glimmer-runtime';
+import { assert } from 'ember-metal/debug';
+import assign from 'ember-metal/assign';
+
+const CLOSURE_COMPONENT = 'ba564e81-ceda-4475-84a7-1c44f1c42c0e';
```
krisselden: is this guid expected on the typescript interface?

zackthehuman: @krisselden No, but it is used to stamp the definition in the constructor of `ClosureComponentDefinition` and also by `isClosureComponent`.

packages/ember-glimmer/lib/syntax/curly-component.js
```diff
@@ -1,15 +1,41 @@
...
 import processArgs from '../utils/process-args';
 import { privatize as P } from 'container/registry';
+import assign from 'ember-metal/assign';
 import get from 'ember-metal/property_get';
 import { ComponentDefinition } from 'glimmer-runtime';
 import Component from '../component';
 
 const DEFAULT_LAYOUT = P`template:components/-default`;
 
+export function validatePositionalParameters(named, positional, positionalParamsDefinition) {
```
chancancode: If all we are doing is `assert`ing (which gets removed from prod builds, I think?), should we just wrap the entire thing in `runInDebug`?

chancancode: @krisselden ?

packages/ember-glimmer/lib/syntax/curly-component.js
```diff
@@ -62,6 +88,46 @@ class ComponentStateBucket {
...
+      // THOUGHT: It might be nice to have a static method on EvaluatedArgs that
+      // can merge two sets of args for us.
+      let mergedArgs = EvaluatedArgs.create({
+        named: EvaluatedNamedArgs.create({
+          map: mergedNamed
+        }),
+        positional: EvaluatedPositionalArgs.create({
+          values: mergedPositional
+        })
+      }).withInternal();
```
chancancode: If you are making a new one, I *think* you don't need to call `withInternal()`? /cc @krisselden

chadhietala:
Should be able to go to `tilde-io/glimmer#be998e6` now and unskip the tests.

packages/ember-glimmer/lib/syntax/curly-component.js
zackthehuman: @chancancode The component manager expects the resulting `EvaluatedArgs` to have `internal` storage. Is there another way to do this?

zackthehuman:
@chadhietala I just tried to bump to `tildeio/glimmer#be998e6` and unskip but unfortunately it seems like there are still bugs to work out. The definitions returned by the `component` helper are branded the same as other dynamic component definitions, but it just renders `[object Object]`. cc @chancancode

packages/ember-glimmer/lib/helpers/component.js
```diff
@@ -0,0 +1,140 @@
...
+  });
+
+  return mergedArgs;
+}
+
+export default {
+  isInternalHelper: true,
+
+  toReference(args, env) {
+    // TODO: Need to figure out what to do about parentMeta here.
```
krisselden: isn't it needed to resolve the component?  you have access when the syntax is refined and it can be added it as a const ref to the args.  The component is supposed to be looked up locally to the template that contained the root `(component path)`.

packages/ember-glimmer/lib/syntax/curly-component.js
krisselden: you can't wrap an export, the asserts should be inlined, you don't want asserts in an util function.

krisselden: but yes, this whole loop should be in a runInDebug or use (Array.prototype.reduce) to get it to a boolean.

packages/ember-glimmer/lib/syntax/curly-component.js
krisselden: Internal is setup by the opcode, you shouldn't be setting it up externally.  Is internal populated before prepareArgs?  If so we should fix that.  We also may want to make a clone method on args?

packages/ember-glimmer/lib/helpers/component.js
zackthehuman: Yes, but I chatted with @wycats and @chancancode and we decided to punt on it and figure out how it will work after we got the feature working.

packages/ember-glimmer/lib/syntax/curly-component.js
zackthehuman: It looks like it's set up in `PutComponentDefinitionOpcode` before it gets stuck back in the args register.

chancancode: Yeah, that whole thing needs rework. What I mean is that you only need to call `withInternal` if you are 1) possibly dealing with a shared object 2) you are about to put stuff into the dictionary.

For you case, you can just do:

```js
 let mergedArgs = EvaluatedArgs.create({ ... });
mergedArgs.internal = originalArgs.internal;
```

chadhietala:
Close in favor of #13923. Thanks @zackthehuman!

PR #13886, opened 2016-07-25
[CLEANUP] Remove unused controller as context cruft

krisselden:
Remove unused module `append-templated-view`
Remove `_context` setup which is no longer used by anything.


rwjblue:
LGTM

PR #13887, opened 2016-07-26
[CLEANUP beta] Unify glimmer behavior with htmlbars and remove unsed code

krisselden:
Normalize attribute assertions and behavior across htmlbars and glimmer

Remove _defaultTagName which was for support of legacy each AST plugin
which was removed

Some general cleanup


chadhietala:
LGTM

rwjblue:
Going to pull this into the beta branch, as it seems like a good change to have in 2.8.

PR #13890, opened 2016-07-26
[Bug] Canary bower-auto-build descriptions contain broken URLs

mike-north:
Currently the [git commit descriptions for the canary channel](https://github.com/components/ember/commits/canary) are building an incorrect url:

Example: 

```
Ember Bower Auto build for https://github.com/emberjs/ember.js/commits/f4a09d6eb346fd2314afacb692f89ef7fd890b77
```

it should be `.../commit/<sha>`

### DIFF
--- a/bin/bower_ember_build
+++ b/bin/bower_ember_build
@@ -59,7 +59,7 @@ git remote rm origin
 # Sending output to /dev/null to prevent GH_TOKEN leak on error.
 git remote add origin https://${USER}:${GH_TOKEN}@github.com/${COMPONENTS_EMBER_REPO_SLUG}.git &> /dev/null
 git add -A
-git commit -m "Ember Bower Auto build for https://github.com/emberjs/ember.js/commits/${TRAVIS_COMMIT}."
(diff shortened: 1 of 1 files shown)


rwjblue:
The urls that are being used are not invalid, they show the list of commits with the one in question as the top one. That allows you to see the recent history. 

I'm not opposed to changing, but the current URL was chosen on purpose and it does work as intended.

mike-north:
Aha, I see. I expected it to be clearly an individual SHA, indicating a snapshot of when the build was generated from the original codebase.

The fact that this list was the commit in question followed by recent commits wasn't at all obvious to me, because I'd have had to actually compare the two SHAs (something humans are not incredibly good at).

rwjblue:
I don't feel super strongly about it.  I'm not opposed to changing at all. Was just pointing out that it was intentional (and valid).

krisselden:
@mike-north do you still think it should be changed after hearing the reasoning?  @rwjblue said he was open to changing it.

mike-north:
Even taking into account that the URLs aren't broken, as @rwjblue pointed out, my opinion is that a URL for the individual commit that was cut is more clear than a list of commits.

PR #13892, opened 2016-07-26
Remove `View#createElement` / `View#destroyElement`.

rwjblue:
These have been private for quite a long time (perhaps always?), and are
used extremely infrequently outside of Ember's own test suite (which has
been cleaned up to avoid these methods over the last few weeks).


jackmatt2:
@rwjblue I have an ember addon that is using the createElement() method.  

```javascript
              const item = Component.create({
                model: model,
                layout: self.get('suggestionTemplate') // This is a compiled ember template
              }).createElement();
```

What is the correct way to create a component in code?

initram:
@rwjblue, I am trying to do something similar to @jackmatt2. But I have not been able to find any way to do that using 2.8 LTS. Did you ever find a solution to this?

Any help is appreciated.

jackmatt2:
@initram This is the way I solved it.

```
            const el = document.createElement('div');
            Component.create({
              model,
              layout: self.get('suggestionTemplate'), // 
              renderer: self.renderer // Required for Gimmer
            }).appendTo(el);
            return el;
```

PR #13904, opened 2016-07-26
[GLIMMER] Ensure objects with `.forEach` can be iterated.

rwjblue:
This was supported (though untested) in HTMLBars.  Added some tests and implemented in Glimmer.

### DIFF
--- a/packages/ember-glimmer/lib/utils/iterable.js
+++ b/packages/ember-glimmer/lib/utils/iterable.js
@@ -175,6 +175,12 @@ class Iterable {
       return new EmberArrayIterator(iterable, keyFor);
     } else if (Array.isArray(iterable)) {
       return iterable.length > 0 ? new ArrayIterator(iterable, keyFor) : EMPTY_ITERATOR;
+    } else if (typeof iterable.forEach === 'function') {
+      let array = [];
+      iterable.forEach(function(item) {
+        array.push(item);
+      });
+      return array.length > 0 ? new ArrayIterator(array, keyFor) : EMPTY_ITERATOR;
     } else if (isEachIn(ref)) {
       let keys = Object.keys(iterable);
       let values = keys.map(key => iterable[key]);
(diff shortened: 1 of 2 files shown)


packages/ember-glimmer/lib/utils/iterable.js
```diff
@@ -175,6 +175,12 @@ class Iterable {
       return new EmberArrayIterator(iterable, keyFor);
     } else if (Array.isArray(iterable)) {
       return iterable.length > 0 ? new ArrayIterator(iterable, keyFor) : EMPTY_ITERATOR;
+    } else if (iterable.forEach !== undefined) {
```
chancancode: Is it better to test for `typeof iterable.forEach === "function"`?

rwjblue: Unsure, but that seems fine too.  I will change it.

packages/ember-glimmer/lib/utils/iterable.js
```diff
@@ -175,6 +175,12 @@ class Iterable {
       return new EmberArrayIterator(iterable, keyFor);
     } else if (Array.isArray(iterable)) {
       return iterable.length > 0 ? new ArrayIterator(iterable, keyFor) : EMPTY_ITERATOR;
+    } else if (typeof iterable.forEach === 'function') {
+      let array = [];
+      iterable.forEach(function(item) {
+        array.push(item);
+      });
+      return array.length > 0 ? new ArrayIterator(array, keyFor) : EMPTY_ITERATOR;
     } else if (isEachIn(ref)) {
```
chancancode: isEachIn probably needs to be moved up – if you do `{{#each-in someArray ...}}` for some reason, it should do what you asked for instead of looping through the array. It's largely unrelated to what you are doing here because even if you move it above your branch it will still be a problem, but we should remember to fix it and add a test.

rwjblue: Sounds good, I'll follow up with that in a separate PR.

vvscode:
@rwjblue @chancancode https://github.com/emberjs/ember.js/pull/13915

packages/ember-glimmer/lib/utils/iterable.js
vvscode: https://github.com/emberjs/ember.js/pull/13915

PR #13906, opened 2016-07-26
[Glimmer2] Various skipped tests

chadhietala:
@krisselden can you verify that the input SSOT test is correct. The TL;DR of the test is that we do not know of an out of band DOM property set thus setting the value on the backing component was idempotent operation and never caused a render.

### DIFF
--- a/packages/ember-glimmer/tests/integration/input-test.js
+++ b/packages/ember-glimmer/tests/integration/input-test.js
@@ -94,7 +94,7 @@ moduleFor('Input element tests', class extends RenderingTest {
     this.assertValue('', 'Value is updated the second time');
   }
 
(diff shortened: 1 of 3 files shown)


packages/ember-glimmer/tests/integration/input-test.js
```diff
@@ -108,9 +108,18 @@ moduleFor('Input element tests', class extends RenderingTest {
...
+    this.setComponentValue('bye');
+
+    this.assertValue('bye', 'Value is used');
+
+    this.setDOMValue('hola');
+
+    this.assertValue('hola', 'DOM is used');
+
+    // idempotent
+    this.setComponentValue('hola');
```
krisselden: I think the idea is this would not fire the change action.

PR #13911, opened 2016-07-26
[BUGFIX beta] Avoid using `.env` property on components.

### LINKED ISSUE #13907: Passing `env` attr to component in template results in `null` value (on beta, canary)
Rendering a template and passing in an attr named `env` will result in its value being `null`, e.g. with the following code:

```
// application.hbs
{{my-component env="foo"}}

// my-component.hbs
env: {{env}}

// my-component.js
...
didReceiveAttrs() {
  console.log('this.env:', this.get('env')); // logs "this.env: null"
}
```

The browser logs "this.env: null" and the component's template renders: "env: " (should be: "env: foo").

Reproducible on beta (2.8.0-beta.1+2414d715 ) and canary (2.9.0-canary+7ee7fcf5 ).
With Ember v2.7.0, the expected behavior occurs (logs "this.env: foo" and renders "env: foo").

[Example ember-twiddle](https://ember-twiddle.com/57464a9f696b4fb003a41fcc1289e045?openFiles=templates.application.hbs%2Ctemplates.components.my-component.hbs)

rwjblue:
Fixes #13907

### DIFF
--- a/packages/ember-htmlbars/lib/renderer.js
+++ b/packages/ember-htmlbars/lib/renderer.js
@@ -46,7 +46,7 @@ Renderer.prototype.prerenderTopLevelView =
 
     renderHTMLBarsBlock(view, block, renderNode);
     view.lastResult = renderNode.lastResult;
-    this.clearRenderedViews(view.env);
+    this.clearRenderedViews(view._env);
   };
 
 Renderer.prototype.renderTopLevelView =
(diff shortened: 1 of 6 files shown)


rwjblue:
@krisselden - Would you mind reviewing?

krisselden:
LGTM, I'd rather put all the view stuff into one thing, but this is a step in the right direction.

rwjblue:
@krisselden - Ya, I will try to find time next week to work on the view meta stuff we have been chatting about (I think that is better than stashing more things on the views themselves like this).

Thanks for reviewing!

PR #13914, opened 2016-07-27
[Glimmer] Implement `isVisible`

chancancode:
This PR probably looks bigger than it is because I fixed a bunch of tests and moved a big chunk of code into its own file, but otherwise it is conceptually quite simple.


chancancode:
Looks like a legit failure on IE9, will investigate

chancancode:
IE9 is failing because we are returning a `SafeString`, Glimmer pass it to `setAttribute` directly. Other browsers seem to handle that alright, but IE9 doesn't like it.

We should change the signature on Glimmer's AttributeChangeList stuff to take an `Opaque` which would force us to do something like this: https://github.com/tildeio/glimmer/blob/master/packages/glimmer-runtime/lib/compiled/opcodes/content.ts#L31-L69

packages/ember-glimmer/lib/environment.js
```diff
@@ -374,3 +350,35 @@ export default class Environment extends GlimmerEnvironment {
     destroyable.destroy();
   }
 }
+
+runInDebug(() => {
```
chadhietala: Nice!

chadhietala:
This looks good to me. As a follow up I can [unskip this test](https://github.com/emberjs/ember.js/blob/e6f49f7474a0f2219b627f1c9f434bc71305ddaa/packages/ember-glimmer/tests/integration/content-test.js#L987) or you can do it in this PR. Up to you.

chancancode:
@chadhietala you can do the honor!

PR #13915, opened 2016-07-27
each-in: Iterate over array by keys, not by elements

vvscode:


chancancode:
Can you add the INUR steps? (Search for the "Pls send halp" issue)

Serabe:
@vvscode [here is the issue](https://github.com/emberjs/ember.js/issues/13127)

packages/ember-glimmer/tests/integration/syntax/each-in-test.js
```diff
@@ -305,3 +305,29 @@ moduleFor('Syntax test: {{#each-in}} undefined path', class extends RenderingTes
...
+
+moduleFor('Syntax test: {{#each-in}}  for array', class extends RenderingTest {
+  ['@test it iterate over array with `in` instead of walking over elements'](assert) {
+    let arr = [1, 2, 3];
+    arr.someKey = 'someKeyValue';
+
+    this.render(strip`
+      {{#each-in arr as |key value|}}
+        [{{key}}:{{value}}]
+      {{/each-in}}`, { arr: [] });
```
chancancode: Normally, you want to put the target test case first, so I would start with `{ arr }` instead.

packages/ember-glimmer/tests/integration/syntax/each-in-test.js
```diff
@@ -305,3 +305,29 @@ moduleFor('Syntax test: {{#each-in}} undefined path', class extends RenderingTes
...
+
+    this.assertText('');
+
+    this.runTask(() => this.rerender());
+
+    this.assertText('');
+
+    this.runTask(() => set(this.context, 'arr', arr));
```
chancancode: For the "U" (update with mutation) step, you want to mutate the object. So for example, here you could `set(arr, 'anotherKey', ...)` and assert the result.

chancancode: Optionally, you can add another step after this to `set(this.context, 'arr', [])`. It is indeed a good thing to test.

packages/ember-glimmer/tests/integration/syntax/each-in-test.js
```diff
@@ -305,3 +305,29 @@ moduleFor('Syntax test: {{#each-in}} undefined path', class extends RenderingTes
...
+
+    this.assertText('[0:1][1:2][2:3][someKey:someKeyValue]');
+
+    this.runTask(() => set(this.context, 'arr', []));
```
chancancode: For the "R" (reset with replacement) step, you want to revert the output to the initial state, but with a new/fresh object. So if you did what I suggested in the beginning and started with `arr`, you would want to make another instance of an array with the same objects/keys (don't reuse the instance you mutated in the "U" step).

packages/ember-glimmer/tests/integration/syntax/each-in-test.js
vvscode: @chancancode  should it rerender after `set(arr, 'someKey', 'someKeyValue')` ? Or i should run `this.rerender()` manually ( cause object didn't changed, as fact )

packages/ember-glimmer/tests/integration/syntax/each-in-test.js
```diff
@@ -305,3 +310,34 @@ moduleFor('Syntax test: {{#each-in}} undefined path', class extends RenderingTes
...
+    this.runTask(() => {
+      delete arr.someKey;
+      this.rerender();
+    });
+
+    this.assertText('[0:1][1:2][2:3]');
+
+    this.runTask(() => set(this.context, 'arr', []));
+
+    this.assertText('');
```
chancancode: Missing the "R" step after this:

`this.runTask(() => set(this.context, 'arr', [1, 2, 3]));`

(which reset the text output to the original value, except using a fresh array)

packages/ember-glimmer/tests/integration/syntax/each-in-test.js
```diff
@@ -1,8 +1,8 @@
-import { set } from 'ember-metal/property_set';
-import { get } from 'ember-metal/property_get';
-import { strip } from '../../utils/abstract-test-case';
-import { applyMixins } from '../../utils/abstract-test-case';
-import { moduleFor, RenderingTest } from '../../utils/test-case';
+import {set} from 'ember-metal/property_set';
+import {get} from 'ember-metal/property_get';
+import {strip} from '../../utils/abstract-test-case';
+import {applyMixins} from '../../utils/abstract-test-case';
+import {moduleFor, RenderingTest} from '../../utils/test-case';
```
chancancode: Don't remove the spaces (the linter will fail the build otherwise)

vvscode: Fixed, sorry

packages/ember-glimmer/tests/integration/syntax/each-in-test.js
vvscode: Thx

rwjblue:
This looks good to me.  I'll let @chancancode do a last review pass to make sure the various things he mentioned are good to go...

Issue #13912, opened 2016-07-27
Deprecation guide addition for Ember.Binding(s) that use connect


toranb:
Today I noticed the ember 2.7 deprecation for Ember.Binding and was curious what I should do in this situation.

```js
initialize: Ember.on("init", function() {
  Ember.Binding.from("model." + field).to("fieldName").connect(this);
})
```

I assumed it was a simple defineProperty inside the init but I'm not having the same luck in ember 1.13

```js
initialize: Ember.on("init", function() {
  Ember.defineProperty(this, "fieldName", Ember.computed("model." + field, function() {
    return this.get("model." + field);
  }));
})
```

If I can find a good solution I'll gladly submit a PR to the guides for others who have this same situation. Thank you in advance!

rwjblue:
Can you whip up an ember-twiddle or JSBin to demo the issue?

mike-north:
We **really** need some in-depth material here to ease deprecation-fixing pain. Here's the error message users will get
```
Error: `Ember.Binding` is deprecated. Consider using an `alias` computed property instead.
```
It offers no clue as to what property on what component was the offender. The "advanced ember" solution to this is 

1. Set a conditional breakpoint at this assert, with the breakpoint condition matching the assert condition
2. When the breakpoint is hit, jump up a couple frames in the stack until you find the offending property key
3. `git grep` for the property key

pixelhandler:
@mike-north great idea regarding giving an indication of what property the binding is associated with in the deprecation warning message.

pixelhandler:
@toranb regarding 1.13 I don't expect any change but understand your app may be using it so leaning toward a upgradable solution makes sense. As Robert mentioned an example using an ember-twiddle.com or a git repo would be great, and of source help with working on what could be added to the guides.

toranb:
@rwjblue @pixelhandler agreed :) I'm planning to carve out some time this week to create a solid twiddle showing this in action. Thanks again!

rwjblue:
I submitted https://github.com/emberjs/ember.js/pull/13920 to make the deprecation message more helpful in tracking down where a binding is being used.

toranb:
Well the good news - I've got a twiddle to prove those should be identical in behavior. The bad news is I think I've found a bug in my code :)

I can't reproduce this issue w/ twiddle for reasons unrelated to this Ember.Binding deprecation. For now I'll leave you with this twiddle and if I do find something broken I'll be sure to report it

[...]

PR #13920, opened 2016-07-27
[BUGFIX release] Add more info to the `Ember.Binding` deprecation.

rwjblue:
The deprecation before:

```
`Ember.Binding` is deprecated. Consider using an `alias` computed property instead.
```

The deprecation after:

```
The `twoWayTest` property of `<(subclass of Ember.Component):ember483>`
is an `Ember.Binding` connected to `direction`, but `Ember.Binding` is
deprecated. Consider using an `alias` computed property instead.
```

Addresses https://github.com/emberjs/ember.js/issues/13912#issuecomment-235525679.


mike-north:
This is a massive improvement. Thanks @rwjblue!

mmun:
@rwjblue Looks great!

PR #13923, opened 2016-07-27
[Glimmer2] Closure Components

chadhietala:
Incorporates and supersedes #13883.

### DIFF
--- a/packages/ember-glimmer/lib/environment.js
+++ b/packages/ember-glimmer/lib/environment.js
@@ -27,6 +27,7 @@ import {
 } from './helpers/if-unless';
 
 import { default as action } from './helpers/action';
+import { default as componentHelper } from './helpers/component';
 import { default as concat } from './helpers/concat';
 import { default as get } from './helpers/get';
 import { default as hash } from './helpers/hash';
@@ -94,14 +95,14 @@ const builtInDynamicComponents = {
     } else {
       return buildTextFieldSyntax({ args, templates }, getDefinition);
     }
-
-    return new DynamicComponentSyntax({ args, templates });
+    return DynamicComponentSyntax.create({ args, templates });
   }
 };
 
 const builtInHelpers = {
   if: inlineIf,
   action,
+  component: componentHelper,
   concat,
   get,
   hash,
@@ -209,6 +210,7 @@ export default class Environment extends GlimmerEnvironment {
     }
 
     let {
+      appendType,
       isSimple,
       isInline,
       isBlock,
@@ -229,7 +231,7 @@ export default class Environment extends GlimmerEnvironment {
       let mappedKey = builtInComponents[key];
 
       if (key === 'component') {
-        return new DynamicComponentSyntax({ args, templates, parentMeta });
+        return DynamicComponentSyntax.create({ args, templates, parentMeta });
       } else if (key === 'render') {
         return new RenderSyntax({ args });
       } else if (key === 'outlet') {
@@ -254,6 +256,14 @@ export default class Environment extends GlimmerEnvironment {
       }
 
       assert(`A helper named "${key}" could not be found`, !isBlock || this.hasHelper(key, parentMeta));
+    }
+
+    if ((!isSimple && appendType === 'unknown') || appendType === 'self-get') {
+      return statement.original.deopt();
+    }
+
+    if (!isSimple && path) {
+      return DynamicComponentSyntax.fromPath({ path, args, templates, parentMeta });
     }
 
     assert(`Helpers may not be used in the block form, for example {{#${key}}}{{/${key}}}. Please use a component, or alternatively use the helper in combination with a built-in Ember helper, for example {{#if (${key})}}{{/if}}.`, !isBlock || !this.hasHelper(key, parentMeta));
@@ -322,7 +332,7 @@ export default class Environment extends GlimmerEnvironment {
       this.owner.lookup(`helper:${name}`);
     // TODO: try to unify this into a consistent protocol to avoid wasteful closure allocations
     if (helper.isInternalHelper) {
-      return (vm, args) => helper.toReference(args);
+      return (vm, args) => helper.toReference(args, this);
     } else if (helper.isHelperInstance) {
       return (vm, args) => SimpleHelperReference.create(helper.compute, args);
     } else if (helper.isHelperFactory) {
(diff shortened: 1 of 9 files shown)


packages/ember-glimmer/lib/environment.js
```diff
@@ -256,6 +258,14 @@ export default class Environment extends GlimmerEnvironment {
       assert(`A helper named "${key}" could not be found`, !isBlock || this.hasHelper(key, parentMeta));
     }
 
+    if ((!isSimple && appendType === 'unknown') || appendType === 'self-get') {
+      return statement.original.deopt();
+    }
+
+    if (!isSimple && path) {
+      return DynamicComponentSyntax.fromPathReference({ path, args, templates, parentMeta });
```
chancancode: pathReference is a bit confusing, would call it `fromPath`, `fromPathAndInvocationArgs`, ... (just not "path reference" because that refers to another thing elsewhere

chancancode: curious – when is path falsy?

packages/ember-glimmer/lib/syntax/dynamic-component.js
```diff
@@ -21,11 +27,27 @@ function dynamicComponentFor(vm) {
...
 export class DynamicComponentSyntax extends StatementSyntax {
-  constructor({ args, templates, parentMeta }) {
+  // for {{component componentName}}
+  static create({ args, templates, parentMeta }) {
+    let definitionArgs = ArgsSyntax.fromPositionalArgs(args.positional.slice(0, 1));
+    let invocationArgs = ArgsSyntax.build(args.positional.slice(1), args.named);
+    return new this({ definitionArgs, args: invocationArgs, templates, parentMeta });
+  }
+
+  // Transforms {{foo.bar}} or {{#foo.bar}}{{/foo.bar}}
```
chancancode: this is only used for `{{foo.bar with="args"}}`

packages/ember-glimmer/lib/syntax/dynamic-component.js
```diff
@@ -21,11 +27,27 @@ function dynamicComponentFor(vm) {
...
+  // into {{component foo.bar}} or {{#component foo.bar}}{{/component}}
+  // with all of it's arguments
+  static fromPathReference({ path, args, templates, parentMeta }) {
+    let positional = ArgsSyntax.fromPositionalArgs(PositionalArgsSyntax.build([GetSyntax.build(path.join('.'))]));
```
chancancode: Once this lands, we should add a `GetSyntax.fromParts` to avoid this join and split dance

packages/ember-glimmer/lib/syntax/curly-component.js
```diff
@@ -62,6 +90,46 @@ class ComponentStateBucket {
...
+      // args (aka 'oldArgs') may be undefined or simply be empty args, so
+      // we need to fall back to an empty array or object.
+      let newNamed = args && args.named && args.named.map || {};
+      let newPositional = args && args.positional && args.positional.values || [];
+
+      let oldNamed = definition.args.named.map;
+      let oldPositional = definition.args.positional.values;
+
+      // Merge positional arrays
+      let mergedPositional = new Array(Math.max(oldPositional.length, newPositional.length));
```
chancancode: @krisselden is `new Array` okay here?

chancancode:
👏 thanks @zackthehuman and @chadhietala!!!

packages/ember-glimmer/lib/syntax/curly-component.js
chancancode: chatted with @krisselden quickly, we should not use `new Array` here. Can you change it to `[]`?

packages/ember-glimmer/tests/integration/components/closure-components-test.js
```diff
@@ -330,7 +331,7 @@ moduleFor('@htmlbars Components test: closure components', class extends Renderi
     this.assertText('Hodi Sigmundur 33');
   }
 
-  ['@glimmer bound outer named parameters get updated in the right scope']() {
+  ['@skip bound outer named parameters get updated in the right scope']() {
```
chancancode: Let's investigate what's up after merge

packages/ember-glimmer/lib/syntax/curly-component.js
```diff
@@ -62,6 +90,46 @@ class ComponentStateBucket {
...
+      // we need to fall back to an empty array or object.
+      let newNamed = args && args.named && args.named.map || {};
+      let newPositional = args && args.positional && args.positional.values || [];
+
+      let oldNamed = definition.args.named.map;
+      let oldPositional = definition.args.positional.values;
+
+      // Merge positional arrays
+      let mergedPositional = [];
+      mergedPositional.splice(0, oldPositional.length, ...oldPositional);
```
chancancode: this is now `push(...oldPositional)`

packages/ember-glimmer/lib/syntax/curly-component.js
```diff
@@ -62,6 +90,46 @@ class ComponentStateBucket {
...
+      mergedPositional.splice(0, newPositional.length, ...newPositional);
```
chancancode: `push(...newPositional)`

packages/ember-glimmer/tests/integration/components/closure-components-test.js
chadhietala: I made a mistake in #13906, didn't realize the entire module was being skipped for `@glimmer`. We should see if we can fix this issue, but currently it does not work in either engine from my understanding.

packages/ember-glimmer/lib/syntax/curly-component.js
```diff
@@ -62,6 +90,46 @@ class ComponentStateBucket {
...
 
 class CurlyComponentManager {
+  prepareArgs(definition, args) {
+    validatePositionalParameters(args.named, args.positional.values, definition.ComponentClass.positionalParams);
+
+    if (definition.args) {
+      // args (aka 'oldArgs') may be undefined or simply be empty args, so
+      // we need to fall back to an empty array or object.
+      let newNamed = args && args.named && args.named.map || {};
+      let newPositional = args && args.positional && args.positional.values || [];
```
chancancode: `args` cannot be `null`/`undefined` here, so you could just do

```
let newNamed = (args.named || EvaluatedNamedArgs.empty()).map;
let newPositional = (args.positional || EvaluatedPositionalArgs.empty()).values;
```

chancancode: (I don't even think `args.named` and `args.positional` can be null, but the code seems to suggest that they could be)

chancancode:
@chadhietala can you follow up with the comments?

chadhietala:
Yup

packages/ember-glimmer/lib/syntax/curly-component.js
chadhietala: You can't do this here because what you are doing is merging over top of the `oldPositional`

chancancode: You are right, just the first one

PR #13924, opened 2016-07-27
Merge #13915 and cleanup each-in tests

chancancode:
Also fixed Glimmer's `each-in` docs: Glimmer's `each-in` is not unbound. (The tag system is essentially a "star observer".)

### DIFF
--- a/packages/ember-glimmer/tests/integration/syntax/each-in-test.js
+++ b/packages/ember-glimmer/tests/integration/syntax/each-in-test.js
@@ -2,7 +2,7 @@ import { set } from 'ember-metal/property_set';
 import { get } from 'ember-metal/property_get';
 import { strip } from '../../utils/abstract-test-case';
 import { applyMixins } from '../../utils/abstract-test-case';
-import { moduleFor, RenderingTest } from '../../utils/test-case';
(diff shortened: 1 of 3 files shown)


packages/ember-glimmer/tests/integration/syntax/each-in-test.js
```diff
@@ -270,7 +279,6 @@ moduleFor('Syntax test: {{#each-in}}', class extends EachInTest {
         'Smartphones': 8203,
         'JavaScript Frameworks': Infinity
       });
-      this.rerender();
```
chancancode: This `rerender` is not necessary. Both engines correctly "observe" the whole object changing.

packages/ember-glimmer/tests/integration/syntax/each-in-test.js
```diff
@@ -281,9 +289,6 @@ moduleFor('Syntax test: {{#each-in}}', class extends EachInTest {
     `);
   }
 
-});
-
-moduleFor('Syntax test: {{#each-in}} undefined path', class extends RenderingTest {
```
chancancode: Not sure why this was a different class

rwjblue:
LGTM

PR #13925, opened 2016-07-27
[Glimmer2] Cleanup positional argument merging

chadhietala:
/cc @chancancode

### DIFF
--- a/packages/ember-glimmer/lib/syntax/curly-component.js
+++ b/packages/ember-glimmer/lib/syntax/curly-component.js
@@ -94,17 +94,16 @@ class CurlyComponentManager {
     validatePositionalParameters(args.named, args.positional.values, definition.ComponentClass.positionalParams);
 
     if (definition.args) {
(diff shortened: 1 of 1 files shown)


packages/ember-glimmer/lib/syntax/curly-component.js
```diff
@@ -94,17 +94,16 @@ class CurlyComponentManager {
     validatePositionalParameters(args.named, args.positional.values, definition.ComponentClass.positionalParams);
 
     if (definition.args) {
-      // args (aka 'oldArgs') may be undefined or simply be empty args, so
-      // we need to fall back to an empty array or object.
-      let newNamed = args && args.named && args.named.map || {};
-      let newPositional = args && args.positional && args.positional.values || [];
+      let newNamed = (args.named || EvaluatedNamedArgs.empty()).map;
+      let newPositional = (args.positional || EvaluatedPositionalArgs.empty()).values;
```
chancancode: Does anything fail (locally) if you remove the `|| X.empty()` guard? I am curious when that could be null

PR #13935, opened 2016-07-28
Fix glimmer after render

krisselden:


packages/ember-glimmer/lib/renderer.js
```diff
@@ -36,118 +36,65 @@ class DynamicScope {
...
-    if (!foundCallback) {
-      throw new TypeError('Cannot deregister a callback that has not been registered.');
-    }
-  }
-
-  hasRegistrations() {
-    return !!this.callback || !!this.callbacks && !!this.callbacks.length;
-  }
+function register(renderer) {
+  assert('renderer is not registered', renderers.indexOf(renderer) === -1);
```
chancancode: prefer "Cannot register the same renderer twice"

packages/ember-glimmer/lib/renderer.js
```diff
@@ -36,118 +36,65 @@ class DynamicScope {
...
-  }
-
-  destroy() {
-    if (this._root) {
-      this.deregisterView(this._root);
-    }
-  }
+function deregister(renderer) {
+  let index = renderers.indexOf(renderer);
+  assert('renderer is registered', index !== -1);
```
chancancode: prefer "Cannot deregister unknown unregistered renderer"

packages/ember-glimmer/lib/renderer.js
```diff
@@ -36,118 +36,65 @@ class DynamicScope {
...
+}
 
-  _maybeUpdate(lastTagValue) {
-    if (CURRENT_TAG.validate(lastTagValue)) { return; }
-    let view = this._root;
-    if (view) {
-      view.renderer.rerender(view);
+function K() {}
+
+let loop = 0;
```
chancancode: prefer "loops"

packages/ember-glimmer/lib/renderer.js
```diff
@@ -36,118 +36,65 @@ class DynamicScope {
...
+function loopEnd(current, next) {
+  for (let i = 0; i < renderers.length; i++) {
```
chancancode: reset `loops` at the beginning, in case something went wrong last time and we didn't reach the end?

packages/ember-glimmer/lib/renderer.js
```diff
@@ -36,118 +36,65 @@ class DynamicScope {
...
+    if (!renderers[i]._isValid()) {
+      if (loop > 10) {
+        loop = 0;
+        // TODO: do something better
+        renderers[i].destroy();
+        throw new Error('infinite rendering invalidation detected');
```
chancancode: infinite loop! 😗

packages/ember-glimmer/lib/renderer.js
```diff
@@ -197,83 +123,131 @@ class Renderer {
...
+    }
+    this._destroyed = true;
+    this._clearRoot();
   }
 
-  ensureViewNotRendering() {
-    // TODO: Implement this
-    // throw new Error('Something you did caused a view to re-render after it rendered but before it was inserted into the DOM.');
+  _renderRoot(root, template, self, target, dynamicScope) {
+    assert(!this._root, 'root is not in DOM');
```
chancancode: The arguments are flipped (also, I prefer "cannot append multiple root views" or something like that)

packages/ember-glimmer/lib/renderer.js
krisselden: I thought assert messages were supposed to be what you were asserting.

packages/ember-glimmer/lib/renderer.js
```diff
@@ -159,31 +106,10 @@ class Renderer {
...
-      callback = rerender;
-    } while (shouldReflush || didChangeAfterRender);
-
-    this._scheduler.registerView(view);
-
-    return result;
+    this._renderRoot(view, view.template, self, target, dynamicScope);
   }
 
   appendTo(view, target) {
```
chancancode: Maybe we should rename target to parentElement since we have `targetObject` below (I was confused way further down and had to scroll back up)

packages/ember-glimmer/lib/renderer.js
```diff
@@ -197,83 +123,131 @@ class Renderer {
...
+      return;
+    }
+    this._destroyed = true;
+    this._clearRoot();
   }
 
-  ensureViewNotRendering() {
-    // TODO: Implement this
-    // throw new Error('Something you did caused a view to re-render after it rendered but before it was inserted into the DOM.');
+  _renderRoot(root, template, self, target, dynamicScope) {
```
chancancode: rename target to parentElement ?

packages/ember-glimmer/lib/renderer.js
krisselden: good catch, that gets me often

packages/ember-glimmer/tests/utils/helpers.js
```diff
@@ -21,8 +22,16 @@ export function buildOwner(options) {
...
+
   owner.register('service:-document', document, { instantiate: false });
+  owner.register('-environment:main', {
+    isInteractive: true,
+    options: { jQuery }
+  }, { instantiate: false });
+  owner.inject('view', '_environment', '-environment:main');
+  owner.inject('component', '_environment', '-environment:main');
   owner.inject('service:-dom-helper', 'document', 'service:-document');
+  owner.inject('view', 'renderer', 'renderer:-dom');
```
chancancode: :+1:

chancancode:
This is great!! 👏👏👏 Should we add a new test for the `afterRender` case?

packages/ember-glimmer/lib/renderer.js
krisselden: what is going on here?

PR #13941, opened 2016-07-29
Disable publishing backtracking build.

rwjblue:


bin/publish_builds
```diff
@@ -3,7 +3,7 @@
 echo -e "CURRENT_BRANCH: ${TRAVIS_BRANCH}\n"
 echo -e "PULL_REQUEST: ${TRAVIS_PULL_REQUEST}\n"
 
-if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then
+if [ "$TRAVIS_PULL_REQUEST" == "false" && "$DISABLE_PUBLISHING" != "true"]; then
```
chancancode: I think this should be

`if [ "$TRAVIS_PULL_REQUEST" == "false" ] && [ "$DISABLE_PUBLISHING" != "true" ]; then`

chancancode: Retrying in #13944

PR #13952, opened 2016-07-29
Use native DOM methods instead of jQuery.

rwjblue:
This allows Ember itself to be evaluated without jQuery and mostly boot
an application without jQuery present.  The only caveat is with
`EventDispatcher`.

There are still future changes needed to make the event dispatcher use
native DOM API's, but that is a bit more tricky because we still need to
support `$('#foo').trigger('click')` which will not work if you only use
`document.addEventListener`.  There is a path forward here, but I'd like
to work on it separately.

### DIFF
--- a/packages/ember-routing/lib/location/hash_location.js
+++ b/packages/ember-routing/lib/location/hash_location.js
@@ -1,11 +1,9 @@
 import { get } from 'ember-metal/property_get';
 import { set } from 'ember-metal/property_set';
 import run from 'ember-metal/run_loop';
-import { guidFor } from 'ember-metal/utils';
 
 import EmberObject from 'ember-runtime/system/object';
 import EmberLocation from 'ember-routing/location/api';
-import jQuery from 'ember-views/system/jquery';
 
 /**
 @module ember
@@ -27,6 +25,8 @@ export default EmberObject.extend({
 
   init() {
     set(this, 'location', get(this, '_location') || window.location);
+
+    this._hashchangeHandler = undefined;
   },
 
   /**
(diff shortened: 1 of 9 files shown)


packages/ember-routing/lib/location/hash_location.js
```diff
@@ -117,7 +113,9 @@ export default EmberObject.extend({
 
         callback(path);
       });
-    });
+    };
+
+    window.addEventListener('hashchange', this._hashchangeHandler);
```
stefanpenner: looks good: http://caniuse.com/#search=addEventListener

packages/ember-routing/lib/location/hash_location.js
```diff
@@ -106,9 +104,7 @@ export default EmberObject.extend({
     @param callback {Function}
   */
   onUpdateURL(callback) {
-    let guid = guidFor(this);
-
-    jQuery(window).on(`hashchange.ember-location-${guid}`, () => {
+    this._hashchangeHandler = () => {
```
stefanpenner: If this function is run again, do we need to remove the existing handler?

rwjblue: it is only called during setup, but I am happy to add code to remove before adding

stefanpenner: @rwjblue actually, it might be predunt to an an assert that yells if the function is invoked if `this._hashchangeHandler` has already been set, or if its been invoked more then once.

packages/ember-views/lib/system/utils.js
```diff
@@ -59,3 +59,15 @@ export function getViewBoundingClientRect(view) {
...
+
+/**
+  Determines if the element matches the specified selector.
+
+  @private
+  @method matches
+  @param {DOMElement} el
+  @param {String} selector
+*/
+export function matches(el, selector) {
```
krisselden:
```
export const elMatches = Element.prototype.matches ||
  Element.prototype.matchesSelector || 
  Element.prototype.mozMatchesSelector ||
  Element.prototype.msMatchesSelector || 
  Element.prototype.oMatchesSelector || 
  Element.prototype.webkitMatchesSelector;

export function matches(el, selector) {
  return elMatches.call(el, selector);
}
```

rwjblue: That's way better. Thank you.

PR #13964, opened 2016-07-30
[Glimmer 2] Step 1 of reducing `CURRENT_TAG`s

chancancode:
Pulling the value of a reference could be a side-effectful operation from the perspective of the revision timeline (e.g. updating an `UpdatableTag` is a change on the timeline). Therefore we should always compute the value first, then pull from the validation tag.

If we always follow that protocol, there is no actual reason to initialize an `UpdatableTag` with a `CURRENT_TAG`. While this is always "correct", it causes values that doesn't otherwise matter to "poison" an tag combinator all the way up. (And also causes caches to miss – because the tag value might have changed after computing the value, as it is usually the case when an `UpdatableTag` is involved.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "git-repo-info": "^1.1.4",
     "git-repo-version": "^0.3.1",
     "github": "^0.2.3",
-    "glimmer-engine": "tildeio/glimmer#b7f441fd",
+    "glimmer-engine": "tildeio/glimmer#2b342e2",
     "glob": "^5.0.13",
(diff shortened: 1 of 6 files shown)


krisselden:
@chancancode we should be testing this.

chancancode:
@krisselden added tests in https://github.com/tildeio/glimmer/commit/48dac287c16b886d6f3b6f7760d64ef5b2257e38. On the Ember side, the semantics should be very well-covered by the test suite. For example, the helpers tests were failing before bumping Glimmer.

PR #13967, opened 2016-07-30
[glimmer2] fix cache of directly imported templates

krisselden:
- Simplify template factory.
- Add tests for template factory.
- Fix caching of template factory instances that are directly imported.


packages/ember-glimmer/lib/template.js
```diff
@@ -1,17 +1,12 @@
 import { Template } from 'glimmer-runtime';
 
-let templateId = 0;
-
 class Wrapper {
-  static create(options) {
```
chancancode: Isn't this needed for container-based instantiation? (However, I do see that the tests are passing, and I'm not sure why)

chancancode: I see, you moved it into below

packages/ember-glimmer/lib/template.js
```diff
@@ -33,15 +28,22 @@ class Wrapper {
...
   let Factory = class extends Wrapper {
     constructor(options) {
-      super(options, id);
-      this.spec = JSON.parse(json);
+      let { env } = options;
+      super(id, env, JSON.parse(json));
+    }
+
+    static create(options) {
+      return new this(options);
```
chancancode: I _think_ you could have just written `new this(id, options.env, JSON.parse(json))` here and left out the constructor there

chancancode: Which makes me realize, the whole thing could just be:

```
export default function template(json) {
  let id = ++templateId;

  return {
    id,
    create({ env }) {
      new Wrapper(id, env, JSON.parse(json));
    }
};
```

krisselden: I had that,  stuff failed because something else was using the constructor and not create which made me undo that refactor

krisselden: Without typescript changing exported signatures is dangerous

chancancode: 👌 I'm pretty sure this detail is internal to ember-glimmer, so we could refactor it, but we can come back to it once this is fixed

krisselden: * foo bar

PR #13969, opened 2016-07-30
[BUGFIX beta] Fix issue 12967. Using render without a name losing templateName property.

andrewmp1:
Using render without a name misses using templateName property on route.


rwjblue:
Awesome, thank you!

dguettler:
Haha, you bet me to it. I was struggling getting the test setup right.

PR #13971, opened 2016-07-30
[GLIMMER] Fix issues with actions with the `(mut` helper.

### LINKED ISSUE #13959: [Glimmer 2] (action (mut nonDefinedValue)) fails
Error:
> Action passed is null or undefined in (action)

Reproduction code:

```hbs
<input value={{nonDefinedProp}} oninput={{action (mut nonDefinedProp) value="target.value"}}>
```
_(note: the `value="target.value" is irrelevant`)_

Twiddle so you don't have to type the code above: 

https://ember-twiddle.com/56f80b7916bde51fb7d3ec47c7b39946?fileTreeShown=false&openFiles=templates.application.hbs%2C

rwjblue:
Fixes #13959.

### DIFF
--- a/packages/ember-glimmer/lib/modifiers/action.js
+++ b/packages/ember-glimmer/lib/modifiers/action.js
@@ -4,6 +4,7 @@ import { uuid } from 'ember-metal/utils';
 import { isSimpleClick } from 'ember-views/system/utils';
 import ActionManager from 'ember-views/system/action_manager';
 import { flaggedInstrument } from 'ember-metal/instrumentation';
+import { INVOKE } from '../helpers/action';
 
 const MODIFIERS = ['alt', 'shift', 'meta', 'ctrl'];
 const POINTER_EVENT_TYPE_REGEX = /^click|mouse|touch/;
@@ -126,6 +127,12 @@ export class ActionState {
         args,
         target
       };
+      if (typeof actionName[INVOKE] === 'function') {
+        flaggedInstrument('interaction.ember-action', payload, () => {
+          actionName[INVOKE].apply(actionName, args);
+        });
+        return;
+      }
       if (typeof actionName === 'function') {
         flaggedInstrument('interaction.ember-action', payload, () => {
           actionName.apply(target, args);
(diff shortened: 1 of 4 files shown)


zackthehuman:
LGTM. :+1:

packages/ember-glimmer/lib/modifiers/action.js
```diff
@@ -194,7 +209,13 @@ export default class ActionModifierManager {
...
     let { positional } = args;
 
-    modifier.actionName = positional.at(1).value();
+    let actionNameRef = positional.at(1);
+
+    if (actionNameRef[INVOKE]) {
+      modifier.actionName = actionNameRef;
+    } else {
+      modifier.actionName = actionNameRef.value();
+    }
```
chancancode: I don't fully understand this code yet, but the reference from `positional.at(1)` is a stable object (i.e. calling `position.at(1)` from here will give a `===` value from the same call in `install`). Like I said I need to keep reading to know what this is doing, but given the reference object itself cannot change, it's likely that you don't kneed to do any work here at all if `actionNameRef[INVOKE]` is truthy.

packages/ember-glimmer/lib/modifiers/action.js
```diff
@@ -194,7 +209,13 @@ export default class ActionModifierManager {
...
     modifier.eventName = modifier.getEventName();
 
     // Not sure if this is needed? If we mutate the actionState is that good enough?
```
chancancode: Would also like to know the answer to this

packages/ember-glimmer/tests/integration/helpers/element-action-test.js
```diff
@@ -1489,4 +1489,47 @@ moduleFor('Helpers test: element action', class extends RenderingTest {
...
+
+    this.runTask(() => {
+      this.$('button').click();
+    });
+
+    this.assert.equal(component.get('member'), 'foo', 'value is updated');
+
+    this.runTask(() => {
+      component.set('member', undefined);
+    });
```
chancancode: Is there supposed to be another assertion after here?

packages/ember-glimmer/lib/modifiers/action.js
rwjblue: Thank you, that makes sense.  Will update.

packages/ember-glimmer/tests/integration/helpers/element-action-test.js
rwjblue: No, the failure scenario to start with was just having `{{action (mut someUndefinedValue)}}`. Setting back to `undefined` with a rerender is the "reset" step.

packages/ember-glimmer/lib/modifiers/action.js
rwjblue: I do not see the point of the lines below.  Seems a but odd, I think we can delete them.

packages/ember-glimmer/tests/integration/helpers/element-action-test.js
chancancode: I see. It seemed a bit odd to have that followed by nothing. Would it make sense to rewrite the tests as something like:

`+      template: '<button {{action (mut label) "foo"}}>{{if label label "Click me"}}</button>'`

So we could assert the value by checking the rendered values?

chancancode:
:+1:

packages/ember-glimmer/lib/modifiers/action.js
zackthehuman: What if you bind the event name to a variable? How will the old handler be cleaned up unless you un/re register?

packages/ember-glimmer/tests/integration/helpers/element-action-test.js
rwjblue: Ya, good point.  Updated.

packages/ember-glimmer/lib/modifiers/action.js
rwjblue: I'm not sure what there is to clean up?

chancancode: Merged the fix, but we should still figure out the answer to this question

PR #13975, opened 2016-07-30
[WIP] [Glimmer 2] link-to classNameBindings assertion

### LINKED ISSUE #13955: [Glimmer 2] invalid inline`classNameBindings` in`link-to` yields an exception
The invalid `classNameBindings` below behaves inconsistently in htmlbars and glimmer (see https://github.com/GavinJoyce/glammer/blob/master/app/templates/issue1.hbs). 

```js
{{#link-to 'index' classNameBindings="someProperty"}}
```

**htmlbars:**

no warning, the follow is rendered:

```html
<a id="ember369" href="/" class="  ember-view">Click me to go to the index</a>
```

**glimmer:**

an exception is thrown when rendering:

`Cannot read property '0' of null (in hasHelper)`

screen shot 2016-07-29 at 14 21 46

Some questions:

 * Should glimmer 2 behave the same as htmlbars?
 * Should we issue a warning in development mode?

GavinJoyce:
fixes https://github.com/emberjs/ember.js/issues/13955

TODO:
- [ ] add similar assertion to class definition? **update:** maybe not - see https://github.com/emberjs/ember.js/pull/13975#issuecomment-236424314

---

``` hbs
{{#link-to "index" classNameBindings=":valid isBig:big isOpen:open:closed invalid"}}
  click me
{{/link-to}}
```

will now produce:

`'invalid' is not a valid classNameBinding. ('baz/foo-bar' @ L1:C19)`

### DIFF
--- a/packages/ember-template-compiler/lib/plugins/transform-old-class-binding-syntax.js
+++ b/packages/ember-template-compiler/lib/plugins/transform-old-class-binding-syntax.js
@@ -1,3 +1,6 @@
+import { assert } from 'ember-metal/debug';
+import calculateLocationDisplay from 'ember-template-compiler/system/calculate-location-display';
+
 export default function TransformOldClassBindingSyntax(options) {
   this.syntax = null;
   this.options = options;
@@ -6,6 +9,7 @@ export default function TransformOldClassBindingSyntax(options) {
 TransformOldClassBindingSyntax.prototype.transform = function TransformOldClassBindingSyntax_transform(ast) {
   let b = this.syntax.builders;
   let walker = new this.syntax.Walker();
+  let moduleName = this.options.moduleName;
 
   walker.visit(ast, node => {
     if (!validate(node)) { return; }
@@ -41,13 +45,17 @@ TransformOldClassBindingSyntax.prototype.transform = function TransformOldClassB
       node.hash.pairs.splice(index, 1);
     });
 
-    each(allOfTheMicrosyntaxes, ({ value, loc }) => {
+    each(allOfTheMicrosyntaxes, ({ value, loc, key }) => {
       let sexprs = [];
       // TODO: add helpful deprecation when both `classNames` and `classNameBindings` can
       // be removed.
 
       if (value.type === 'StringLiteral') {
         let microsyntax = parseMicrosyntax(value.original);
+
+        if (key === 'classNameBindings') {
+          assertValidClassNameBindings(microsyntax, moduleName, loc);
+        }
 
         buildSexprs(microsyntax, sexprs, b);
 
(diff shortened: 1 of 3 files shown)


chancancode:
Is it better to do this at runtime (bingings.js in ember-glimmer somewhere) so that we could catch the same thing if you supply the same thing on the class? Or is this only disallowed in this position in the template?

rwjblue:
We should do both ways. Template invocation is more likely and in that case we can provide much better feedback (exact line/column info), but we should still check in init or something to ensure up didn't do it in the class definition itself.

GavinJoyce:
I'll follow up tomorrow and add a similar assertion to the class definition.

GavinJoyce:
There seems to be a discrepancy between `classNameBindings` in a template and in a class definition.  ([twiddle](https://ember-twiddle.com/6c2537edb1c0445ed6ae5e94d5082757?openFiles=templates.application.hbs%2Ctemplates.components.my-link-to.hbs))

This is valid:

```js
export default Ember.LinkComponent.extend({
  classNameBindings: ['isUrgent'],
  isUrgent: true
});
```

it renders the following in htmlbars:

```html
<a id="ember355" href="/" class="ember-view active is-urgent">
  Index
</a>
```

The `isUrgent` property can be set at run time, so this is also a valid definition:

```js
export default Ember.LinkComponent.extend({
  classNameBindings: ['isUrgent']
});
```

A template seems to behave differently:

```hbs
{{#link-to 'index' classNameBindings='isUrgent' isUrgent=true}}
  Index
{{/link-to}}
```

```html
<a id="ember366" href="/" class="  ember-view active">
  Index
</a>
```

Should they behave the same? If so, adding an assertion isn't correct - we should fix the bug in htmlbars and glimmer (and possibly remove the extra rendered class spaces in glimmer for aesthetic reasons). If the behaviour shouldn't be the same, perhaps asserting in the template is all that we need?

scottmessinger:
FWIW, I think the existing functionality is essential and I have it at a couple points in my app. Without `classNameBindings` in the link-to component, there's no way (that I know of) to add a dynamic class. If `classNameBindings` can't be made to work in Glimmer2, what work around should we use to add a dynamic class to a link-to component?

rwjblue:
We are not discussing removing `classNameBindings`, just what to do when you provide an invalid value (and then we are trying to determine what is valid/invalid).

FWIW, I would likely just use `class="(if isFoo 'is-foo' 'not-foo')"` just like you would with "normal" HTML elements.

scottmessinger:
@rwjblue Ahh--I misunderstood. My app also broke with Glimmer2 due to classNameBindings (and I got the same error message as Gavin). It worked fine in 2.7 and, since there was nothing that *seemed* invalid about the value I was passing to classNameBindings, I just figured the conversation was about taking it out. Sorry about that!

chancancode:
@GavinJoyce sorry for misleading you! as it turns out (and as you have found out), these are two different features with the same name.

On the JS class side, if you have a `classNameBinding` of `fooBarBaz` and it resolves to true on the component instance, we add `foo-bar-baz` to the class. [Demo](https://ember-twiddle.com/50f418bb6ccc32f0f6fe517b0682e4a0)

On the invocation side, it's essentially the same – if you have a `classNameBinding` of `fooBarBaz` and it resolves to true on the **caller**, we add `foo-bar-baz` to the class. [Demo](https://ember-twiddle.com/fa785b1da57e3b2cd99c24186cd5c962)

So it seems like this is not invalid, after all. I guess we need to remove that warning from Glimmer, add the test for both, and implement that missing feature on Glimmer.

GavinJoyce:
@chancancode thanks, I'll continue working on this

PR #13977, opened 2016-07-30
Fix ported outlet test

krisselden:
Symlink integration tests to htmlbars.

LOG_VIEW_LOOKUPS is done in the router

Ensure outlet state matches router.

TODO: extract building outlet states, also, I think ViewClass is no longer used by anything.


packages/ember-glimmer/tests/integration/outlet-test.js
```diff
@@ -1,46 +1,50 @@
...
 
-  teardown() {
-    runDestroy(this.top);
-
-    super.teardown(...arguments);
-  }
-
-  withTemplate(string) {
-    return {
+  ['@htmlbars should render the outlet when set after DOM insertion']() {
```
rwjblue: This test title is wrong now, since when it appends to the DOM it has a fully populated  outletState.

rwjblue:
LGTM, thank you for cleaning this up.

PR #13978, opened 2016-07-31
[Glimmer2] More coverage around falsy input values

chadhietala:


chadhietala:
@krisselden Does this look good?

PR #13980, opened 2016-07-31
Migrate remainder of outlet tests to new testing setup.

rwjblue:
/cc @krisselden


rwjblue:
Ported tests and updated to ensure templates used in outlet state are all looked up from the owner properly.

rwjblue:
Migrating ember-test-helpers to look templates up from the container in https://github.com/switchfly/ember-test-helpers/pull/168.

rwjblue:
I discussed this a bit with @krisselden, and he wanted me to migrate these `this.assertText` calls into `assertHTML`.  I tried that but ran into some issues due to the fact that we cannot get access to the `id` of the top level wrapping component (which makes the assertions fail).

PR #13987, opened 2016-08-01
[GLIMMER] Fix `Component#readDOMAttr`.

rwjblue:
There are a number of specific attributes that must be converted to properties after initial values. Using `this.readDOMAttr` was designed to assist with the `<input value="adf">` case (to prevent cursor position loss).  This migrates from the temporary `this.element.getAttribute` to the more fully featured `readDOMAttr` that is exported by `glimmer-runtime`.


wycats:
r+

PR #13988, opened 2016-08-01
Refactor `{{#each}}` tests

chancancode:
- Refactor `TogglingConditionalsSyntaxTest`, so that `{{#each}}` and `{{#each-in}}` can share the same test cases with `{{#if}}`, `{{#unless}}` and `{{#with}}`
- Refactor most of the `{{#each}}` tests so that we can run them against `emberA`, array-like objects (objects with a `forEach` method) and array proxies

### DIFF
--- a/packages/ember-glimmer/lib/utils/iterable.js
+++ b/packages/ember-glimmer/lib/utils/iterable.js
@@ -165,20 +165,31 @@ class Iterable {
 
     if (isProxy(iterable)) {
       valueTag.update(CURRENT_TAG);
+      iterable = get(iterable, 'content');
     } else {
       valueTag.update(tagFor(iterable));
     }
 
     if (iterable === undefined || iterable === null) {
       return EMPTY_ITERATOR;
-    } else if (isEachIn(ref)) {
+    }
+
+    let typeofIterable = typeof iterable;
+
+    if (typeofIterable !== 'object' && typeofIterable !== 'function') {
+      return EMPTY_ITERATOR;
+    }
+
+    if (isEachIn(ref)) {
       let keys = Object.keys(iterable);
       let values = keys.map(key => iterable[key]);
       return keys.length > 0 ? new ObjectKeysIterator(keys, values, keyFor) : EMPTY_ITERATOR;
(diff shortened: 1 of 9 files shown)


chancancode:
Fix is in tildeio/glimmer#249

chancancode:
Tests should pass for Glimmer now.

HTMLBars is failing because:

1. Things like `{{#each null}}`, `{{#each true}}` are neither truthy or falsy (it renders nothing)
2. `{{#each-in}}` does not support `ObjectProxy` – it tries to iterate the proxy/wrapper instead of the content
3. `{{#each-in}}` supports strings – `{{#each-in "hello" as |c|}}[{{c}}]{{/each}}` would give you `[h][e][l][l][o]`. Admittedly, this is something I just disallowed for Glimmer in one of the commits, and I could easily make it work. However, `Object.keys("hello")` is illegal in ES5, so it throws in IE <= 11.

Personally, I consider all of them bugs in HTMLBars (especially number 1). I wouldn't mind seeing them fixed, but I don't know enough to make it happen.

@krisselden @rwjblue thoughts?

rwjblue:
I think focusing on glimmer makes sense. I don't see any compelling reason to try and backport these fixes into HTMLBars.

chancancode:
@rwjblue can haz review?

packages/ember-glimmer/lib/utils/iterable.js
```diff
@@ -165,28 +165,39 @@ class Iterable {
...
     }
 
     if (iterable === undefined || iterable === null) {
       return EMPTY_ITERATOR;
-    } else if (isEachIn(ref)) {
+    }
+
+    let typeofIterable = typeof iterable;
+
+    if (typeofIterable !== 'object' && typeofIterable !== 'function') {
```
chancancode: @krisselden I meant to ask about this. I saw this pattern (saving the typeof result into a variable) in some other places. Is this generally necessary, or could it end up doing more harm than good? (I wrote it because in the original version I did the check in several branches, but now it's just this one line I wonder if it's still needed.)

PR #13996, opened 2016-08-02
[Bugfix Beta] Fix for auto-mut wrapping of closure actions.

rwjblue:
Given the following:

``` hbs
{{x-foo derp=(action 'herk')}}
```

``` hbs
{{! app/templates/components/x-foo.hbs }}

{{x-bar derp=derp}}
```

``` js
  didInsertElement() {
    this.get('derp')();
  }
});
```

An error is triggered with the following output:

```
TypeError: this.attrs.derp is not a function
    at Class.fireAction (http://localhost:7200/ember-tests.js:21306:46)
    at http://localhost:7200/ember-tests.js:21344:24
    at Object.run (http://localhost:7200/ember.debug.js:294:25)
    at Object.run [as default] (http://localhost:7200/ember.debug.js:19515:27)
    at _class2.runTask (http://localhost:7200/ember-tests.js:32041:34)
    at _class2.testActionClosureDoesNotGetAutoMutWrapped [as @test action closure does not get auto-mut wrapped] (http://localhost:7200/ember-tests.js:21343:12)
    at Object.<anonymous> (http://localhost:7200/ember-tests.js:31691:31)
    at runTest (http://localhost:7200/qunit/qunit.js:843:28)
    at Object.run (http://localhost:7200/qunit/qunit.js:828:4)
    at http://localhost:7200/qunit/qunit.js:970:11
```

### DIFF
--- a/packages/ember-glimmer/lib/utils/process-args.js
+++ b/packages/ember-glimmer/lib/utils/process-args.js
@@ -3,6 +3,7 @@ import { CONSTANT_TAG } from 'glimmer-reference';
 import { ARGS } from '../component';
 import { UPDATE } from './references';
 import { MUTABLE_CELL } from 'ember-views';
+import { ACTION } from '../helpers/action';
 
 export default function processArgs(args, positionalParamsDefinition) {
   if (!positionalParamsDefinition || positionalParamsDefinition.length === 0 || args.positional.length === 0) {
@@ -50,7 +51,9 @@ class SimpleArgs {
       let ref = namedArgs.get(name);
       let value = attrs[name];
 
-      if (ref[UPDATE]) {
+      if (typeof value === 'function' && value[ACTION]) {
+        attrs[name] = value;
+      } else if (ref[UPDATE]) {
         attrs[name] = new MutableCell(ref, value);
       }
 
--- a/packages/ember-glimmer/lib/helpers/action.js
+++ b/packages/ember-glimmer/lib/helpers/action.js
@@ -13,6 +13,7 @@ import {
 } from 'ember-metal';
 
 export const INVOKE = symbol('INVOKE');
+export const ACTION = symbol('ACTION');
 
 /**
   The `{{action}}` helper provides a way to pass triggers for behavior (usually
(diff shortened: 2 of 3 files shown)


chancancode:
Something seems fishy here, why is any new code accessing `this.attrs.*` o_0?

chancancode:
Ah it's in the test, carry on

packages/ember-glimmer/lib/utils/process-args.js
```diff
@@ -51,7 +51,9 @@ class SimpleArgs {
       let ref = namedArgs.get(name);
       let value = attrs[name];
 
-      if (ref[UPDATE]) {
+      if (typeof value === 'function') {
+        attrs[name] = value;
```
rwjblue: This would prevent passing a class down to a child component.  Given the following setup (we should add a test for this):

```hbs
{{foo-bar klass=ModelClass}}
```

```js
{{! parent component }}
  init() {
    this._super();
    this.klass = Ember.Object.extend();
  }
});
```

```js
// foo-bar component

  actions: {
    updateKlass(newKlass) {
      this.attrs.klass.update(newKlass);
    }
  }
});
```

packages/ember-glimmer/lib/utils/process-args.js
```diff
@@ -51,7 +51,9 @@ class SimpleArgs {
       let ref = namedArgs.get(name);
       let value = attrs[name];
 
-      if (ref[UPDATE]) {
+      if (typeof value === 'function') {
```
rwjblue: In prior versions (1.13 - 2.8) we use a symbol stamped on the generated action function that we used as an identifier that this is an action.  

Assuming we do the same here (this would alleviate the concern about intentionally passing down a class / function that is intended to be two way bound), this would look something like:

```js
if (typeof value === 'function' && value[ACTION]) {
  attrs[name] = value;
} else if (ref[UPDATE]) {
  // ...snip...
}
```

chadhietala: Fixed.

Issue #13982, opened 2016-07-31
[Glimmer 2] Updating a bound attr of a contextual component creates a new component


cibernox:
If a contextual component has some data bound, and that data is updated, any instances of that component gets complete reinitialized like if it was an entirely new component.

```hbs
<button onclick={{action "toggleIsOpen"}}>Toggle</button> {{!-- Clicking here causes the component's `init` hook to be called again. Creates a new component --}}
{{#with (hash ctxCmp=(component "my-comp" isOpen=isOpen)) as |thing|}}
  {{#thing.ctxCmp}}This is a contextual component{{/thing.ctxCmp}}
{{/with}}
```
This is a regression, since in 2.7 the same component is reused and the state just gets updated.

Live twiddle (works in stable, fails in alpha):

https://ember-twiddle.com/5b6ea116582202584db37a667a21b056?openFiles=templates.application.hbs%2Ctemplates.components.my-comp.hbs

chancancode:
Thanks for reporting! This is definitely a bug we can/should fix 👍

chancancode:
We currently rely on the ComponentDefinition (an internal object) being stable/`===`, but looks like the ones produced by the component helper is not. So we either have to fix the component helper to produce stable values, or to beef up the stability check to not rely on `===`. Currently I believe the first approach would be simplier.

cibernox:
@chancancode In my mind that's how this worked already. I thought that the reference was stable unless the first argument (the component's name) changes.

zackthehuman:
I'm looking in to this.

rwjblue:
Should be closed by #13997, @cibernox can you confirm and close?

cibernox:
@rwjblue It does work on one case, but not in other I believe. I'll check if I can narrow it down to another twiddle.

cibernox:
@zackhall There is one case that still fails. If the name of the component is not a constant reference, the component keeps being initialized.

https://ember-twiddle.com/5b6ea116582202584db37a667a21b056?openFiles=templates.application.hbs%2Ctemplates.components.my-comp.hbs

rwjblue:
Hehe, I think you meant @zackthehuman here.

cibernox:
Indeed. Github's username suggestion is a bit drunk actually.

rwjblue:
Just landed #14005 which should fix.  

@cibernox - Can you confirm and close (will be ~ 15-20 minutes until alpha build is updated).

cibernox:
[...]

PR #13997, opened 2016-08-02
Make component helper reference stable + tests

zackthehuman:
This addresses #13982 by making the `component` helper's tag use only the component name reference's `tag` instead of `args.tag`.

Added two tests, one for the case where the reference is stable, and another for where it should be invalidated. Interestingly enough the latter case doesn't work in HTMLBars. Would love some eyes on the tests since I'm confused why it doesn't work in HTMLBars.

/cc @chancancode @cibernox

### DIFF
--- a/packages/ember-glimmer/lib/helpers/component.js
+++ b/packages/ember-glimmer/lib/helpers/component.js
@@ -26,7 +26,7 @@ export class ClosureComponentReference extends CachedReference {
     super();
     this.defRef = args.positional.at(0);
     this.env = env;
-    this.tag = args.tag;
+    this.tag = args.positional.at(0).tag;
     this.parentMeta = parentMeta;
     this.args = args;
   }
--- a/packages/ember-glimmer/tests/integration/components/closure-components-test.js
+++ b/packages/ember-glimmer/tests/integration/components/closure-components-test.js
@@ -806,6 +806,140 @@ moduleFor('Components test: closure components', class extends RenderingTest {
     assert.equal(this.$().text(), 'message: goodbye');
   }
 
+  ['@test GH#13982 contextual component ref is stable even when bound params change'](assert) {
+    let instance, previousInstance;
+    let initCount = 0;
+
+    this.registerComponent('my-comp', {
+      ComponentClass: Component.extend({
+        init() {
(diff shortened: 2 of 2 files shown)


packages/ember-glimmer/lib/helpers/component.js
```diff
@@ -26,7 +26,7 @@ export class ClosureComponentReference extends CachedReference {
     super();
     this.defRef = args.positional.at(0);
     this.env = env;
-    this.tag = args.tag;
+    this.tag = args.positional.at(0).tag;
```
chancancode: 💯

chancancode:
The fix looks good to me. What is failing on HTMLBars?

zackthehuman:
@chancancode After the component name is changed from `my-comp` to `your-comp` it appears that the components' `init` is not running, so all of the assertions that look for changes are failing:

```
10. a previous instance now exists
Expected: 	
true
Result: 	
false

12. the component was constructed exactly 2 times
Expected: 	
2
Result: 	
1

13. failed (text changed)
Expected: 	
"your-comp: open"
Result: 	
"my-comp: open"
Diff: 	
"myour-comp: open"
```

I'm not sure if the test I wrote is funky or what, but the behavior seems odd.

cibernox:
I'd say that on real apps that works. It **has** to work, otherwise the main usage of the component keyword is just busted.

rwjblue:
@chancancode - Apparently, the closure component helper in HTMLBars does not actually support changing component name when yielded as a contextual component?  

I made a JSBin against 2.7 just to confirm: http://jsbin.com/tuxiquv/edit?html,js,output

rwjblue:
@cibernox and I are both fairly surprised by this.

We made another JSBin that uses the non-closure `{{component` helper to render something passed in as a closure component and that does work. http://jsbin.com/yewiye/edit?html,js,output

I guess changing the name never actually worked in HTMLBars when using the "dot path" invocation syntax.

@Serabe - Can you confirm?

rwjblue:
I'm going to land this though, because it seems like an obvious improvement.  We can try to confirm things RE: HTMLBars implementation separately.

PR #14005, opened 2016-08-02
[GLIMMER2] Cache component def until name changes

zackthehuman:
Addresses an issue discovered in #13982. The `component` helper will now keep track of the value of its `nameRef` and when it changes it will construct a new definition, otherwise it returns the last definition it created. This is needed to avoid recomputing the definition when there is a false positive in the tag/reference system in Glimmer.

/cc @rwjblue @chancancode


chadhietala:
LGTM

PR #14009, opened 2016-08-03
[BUGFIX beta] attributeBindings not working with `role`

### LINKED ISSUE #14007: [BUG beta] attributeBindings not working for "role"
This works in 2.7 and fails in 2.8. Funny enough, it seems to work in general, it just fails (that I've know of) with the `role` attribute.

Twiddle with a demo: https://ember-twiddle.com/ba0c926a18c53ee7155dc23059c70caa?openFiles=components.my-component.js%2Ctemplates.components.my-component.hbs

cibernox:
Closes #14007

### DIFF
--- a/packages/ember-htmlbars/lib/system/build-component-template.js
+++ b/packages/ember-htmlbars/lib/system/build-component-template.js
@@ -186,7 +186,7 @@ function normalizeComponentAttributes(component, attrs) {
     }
   }
 
-  normalized.role = buildStatement('get', 'ariaRole');
+  normalized.role = normalized.role || buildStatement('get', 'ariaRole');
 
   if (attrs.tagName) {
     component.tagName = attrs.tagName;
--- a/packages/ember-glimmer/tests/integration/components/attribute-bindings-test.js
+++ b/packages/ember-glimmer/tests/integration/components/attribute-bindings-test.js
@@ -517,4 +517,26 @@ moduleFor('Attribute bindings integration', class extends RenderingTest {
 
     this.assertComponentElement(this.firstChild, { tagName: 'a', attrs: { href: 'unsafe:javascript:alert(\'foo\')' } });
   }
+
+  ['@test it can bind the role attribute (issue #14007)']() {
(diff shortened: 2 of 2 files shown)


packages/ember-htmlbars/lib/system/build-component-template.js
```diff
@@ -186,7 +186,7 @@ function normalizeComponentAttributes(component, attrs) {
     }
   }
 
-  normalized.role = buildStatement('get', 'ariaRole');
+  normalized.role = normalized.role || buildStatement('get', 'ariaRole');
```
rwjblue: I think you can remove the `||` by moving the original line up above line 160. That way we always set `normalized.role` and only if you happen to add an attributeBinding for it would it override (on line 185).  

Can you test to see if that fixes the failing test?

rwjblue:
Just wanted to confirm that this is working properly in 2.7, but not 2.8. If 2.7 were also broken we would need `BUGFIX release`...

packages/ember-htmlbars/lib/system/build-component-template.js
rwjblue: Hmm. I see that this is what @krisselden suggested in #14007. 

Seems good to me.

cibernox:
2.7 is fine, this bug is only in beta and canary

packages/ember-htmlbars/lib/system/build-component-template.js
cibernox: However it seems a bit confusing that there is two ways of achieving the same thing. `ariaRole="foo"` or `role="foo"` + attributeBindings.
I think that it should be only one blessed approach and the other one should be deprecated.

amk221: I'm getting a failure of the ability to set `ariaRole` in v 3 beta 1.
See here, all tests are passing on Travis apart from one
https://travis-ci.org/zestia/ember-select-box/jobs/352537510

Should I be using `ariaRole`, or `role`

PR #14011, opened 2016-08-03
Track unique history location states

bcardarella:
This PR will allow the HistoryLocation API to keep track of unique
state UUIDs for each state that is pushed/replaced. This change should not
break backwards compatibility and can allow for other libraries to use
this information to set proper scroll position.

The path alone does not provide enough information. For example, if you
visit page A, scroll down, then click on a link to page B, then click on
a link back to page A. Your actual browser history stack is [A, B, A].
Each of those nodes in the history should have their own unique scroll
position. In order to record this position we need an ID that is unique
for each node in the history.

### DIFF
--- a/packages/ember-routing/lib/location/history_location.js
+++ b/packages/ember-routing/lib/location/history_location.js
@@ -1,6 +1,7 @@
 import {
   get,
-  set
+  set,
+  isFeatureEnabled
 } from 'ember-metal';
 
 import { Object as EmberObject } from 'ember-runtime';
@@ -12,6 +13,19 @@ import EmberLocation from './api';
 */
 
 let popstateFired = false;
+
+let _uuid;
+
+if (isFeatureEnabled('ember-unique-location-history-state')) {
+  _uuid = function _uuid() {
+    return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function(c) {
+      var r, v;
+      r = Math.random() * 16 | 0;
+      v = c === 'x' ? r : r & 3 | 8;
+      return v.toString(16);
+    });
+  }
+}
 
 /**
   Ember.HistoryLocation implements the location API using the browser's
@@ -134,6 +148,10 @@ export default EmberObject.extend({
     from getState may be null if an iframe has changed a window's
     history.
 
+    The object returned will contain a `path` for the given state as well
+    as a unique state `id`. The state index will allow the app to distinguish
+    between two states with similar paths but should be unique from one another.
+
     @private
     @method getState
     @return state {Object}
@@ -155,6 +173,9 @@ export default EmberObject.extend({
   */
   pushState(path) {
     let state = { path };
+    if (isFeatureEnabled('ember-unique-location-history-state')) {
+      state.uuid = _uuid();
+    }
 
     get(this, 'history').pushState(state, null, path);
 
@@ -173,6 +194,10 @@ export default EmberObject.extend({
   */
   replaceState(path) {
     let state = { path };
+    if (isFeatureEnabled('ember-unique-location-history-state')) {
+      state.uuid = _uuid();
+    }
+
     get(this, 'history').replaceState(state, null, path);
 
     this._historyState = state;
--- a/features.json
+++ b/features.json
@@ -8,6 +8,7 @@
     "ember-testing-resume-test": null,
     "ember-factory-for": true,
     "ember-no-double-extend": null,
-    "ember-routing-router-service": null
+    "ember-routing-router-service": null,
+    "ember-unique-location-history-state": null
   }
 }
--- a/packages/ember-routing/tests/location/history_location_test.js
+++ b/packages/ember-routing/tests/location/history_location_test.js
@@ -1,4 +1,8 @@
-import { set, run } from 'ember-metal';
+import {
+  isFeatureEnabled,
+  set,
+  run
+} from 'ember-metal';
 import HistoryLocation from '../../location/history_location';
 
 let FakeHistory, HistoryTestLocation, location;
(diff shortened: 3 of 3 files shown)


packages/ember-routing/lib/location/history_location.js
```diff
@@ -150,7 +151,8 @@ export default EmberObject.extend({
    @param path {String}
   */
   pushState(path) {
-    let state = { path: path };
+    let id = guidFor(`${path}-${new Date()}`);
```
rwjblue: I don't think using `guidFor` on a string is very useful, if the idea is that all new states have a unique `id` I think this may be easier/better:

```js
let state = { path, id: ++stateCounter };
```

bcardarella: yeah I figured that was hairy. I'll go about the stateCounter

bcardarella:
@rwjblue updated

packages/ember-routing/lib/location/history_location.js
```diff
@@ -33,6 +33,7 @@ export default EmberObject.extend({
     set(this, 'location', get(this, 'location') || window.location);
 
     this._popstateHandler = undefined;
+    this.stateCounter = 0;
```
mixonic: This seems private, `this._stateCounter` would be preferable.

mixonic:
@bcardarella this also adds public API to `getState` in the form of `id`, correct? Can you add some documentation to `getState` about the shape of the object being returned?

Generally seems reasonable.

packages/ember-routing/lib/location/history_location.js
```diff
@@ -150,7 +151,8 @@ export default EmberObject.extend({
    @param path {String}
   */
   pushState(path) {
-    let state = { path: path };
+    let id = `${this.stateCounter++}`;
```
mixonic: Can this just be the integer?

bcardarella: I feel like I had a good reason to force it to a string. Let me go back and see why I did that. It might not be necessary now.

packages/ember-routing/lib/location/history_location.js
bcardarella: :+1:

packages/ember-routing/lib/location/history_location.js
bcardarella: This is something that I can handle upstream. Moving to just int

bcardarella:
@mixonic updated per your suggestions

rwjblue:
There is some question here if feature flagging and/or an RFC is needed here.  I think the quick summary is that If this is introducing public API, then we should flag it (and potentially create a small and quick RFC for it)...

bcardarella:
The counter itself is private as @mixonic suggested. The unique value added to the history states is not an API

bcardarella:
That's what I believe, but of course not my call ;)

mixonic:
@bcardarella yeah the return value from a public API is of course at best public API itself, or at the worst intimate. For modern things that are properties but private API we've started using symbols and non-enumerable getters, but that obviously hasn't propagated through the whole codebase.

The use-case you described seems reasonable, but a socialized RFC would help gather other use cases that may be related. It would be good to hear from some other community members.

If you can draft something I'm happy to help edit and champion this with you 🙇

bcardarella:
@mixonic finally got around to this: https://github.com/emberjs/rfcs/pull/186

stefanpenner:
thanks @bcardarella !

stefanpenner:
After reading through the RFC and then commenting on it, i realized I believe this can already be implemented via [map or WeakMap, but maybe we should talk about it on the RFC itself](https://github.com/emberjs/rfcs/pull/186#discussion_r91010352)

krisselden:
@stefanpenner  session state is serialized for browser crash recovery including the history state, so if you don't map the id to session storage or the sequence your app will be broken if someone chooses to recover their session after a crash.

In fact this PR should really store the id sequence start in session storage.

bcardarella:
@stefanpenner @krisselden should this conversation take place in the RFC?

krisselden:
@bcardarella I'll copy & paste it

stefanpenner:
> @stefanpenner session state is serialized for browser crash recovery including the history state, so if you don't map the id to session storage or the sequence your app will be broken if someone chooses to recover their session after a crash.
> 
> In fact this PR should really store the id sequence start in session storage.

@krisselden ya, i came to this conclusion in the linked discussion.

packages/ember-routing/lib/location/history_location.js
```diff
@@ -150,7 +155,8 @@ export default EmberObject.extend({
    @param path {String}
   */
   pushState(path) {
-    let state = { path: path };
+    let id = this._stateCounter++;
```
stefanpenner: We should update this to use a UUID more details [here](https://github.com/emberjs/rfcs/pull/186#issuecomment-272524976)

stefanpenner (dismissed):

bcardarella:
@stefanpenner updated

rwjblue (changes requested):
The RFC for this was just merged (:tada:), so now this is ready for some more work.

Basic steps to land are:

* add new feature to `feature.json` with `null` value
* feature flag API changes (`if (isFeatureEnabled('new-feature-name') { }` etc)
* feature flag tests

bcardarella:
@rwjblue I wrapped the private scope `_uuid` function like so:

```js
if (isFeatureEnabled('ember-unique-location-history-state')) {
  function _uuid() {
    return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function(c) {
      var r, v;
      r = Math.random() * 16 | 0;
      v = c === 'x' ? r : r & 3 | 8;
      return v.toString(16);
    });
  }
}
```

but am getting the following error:

```bash
18:3   error  Move function declaration to program root  no-inner-declarations
```

How should I resolve?

rwjblue:
I think generally we do something like this:

```js
let _uuid;
if (isFeatureEnabled('ember-unique-location-history-state')) {
  _uuid = function _uuid() { /*  implementation here */ };
}
```

bcardarella:
All set

rwjblue (approved):

knownasilya:
Can this be used in 2.14 or is the addon by DollarShaveClub still necessary?

bcardarella:
The addon is still necessary as it makes use of the token for the scroll positioning. If you have a use outside of the addon this functionality is in the HistoryLocation class in 2.14

PR #14012, opened 2016-08-03
[GLIMMER2] Remove remaining `CURRENT_TAG`s

chancancode:
some context

cc @krisselden @wycats

### DIFF
--- a/packages/ember-runtime/lib/mixins/-proxy.js
+++ b/packages/ember-runtime/lib/mixins/-proxy.js
@@ -25,7 +25,7 @@ import symbol from 'ember-metal/symbol';
 const IS_PROXY = symbol('IS_PROXY');
 
 export function isProxy(value) {
-  return value && value[IS_PROXY];
+  return typeof value === 'object' && value && value[IS_PROXY];
 }
 
 function contentPropertyWillChange(content, contentKey) {
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "git-repo-info": "^1.1.4",
     "git-repo-version": "^0.3.1",
     "github": "^0.2.3",
-    "glimmer-engine": "tildeio/glimmer#5a11172",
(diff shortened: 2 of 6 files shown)


packages/ember-runtime/lib/mixins/-proxy.js
```diff
@@ -25,7 +25,7 @@ import symbol from 'ember-metal/symbol';
 const IS_PROXY = symbol('IS_PROXY');
 
 export function isProxy(value) {
-  return value && value[IS_PROXY];
+  return typeof value === 'object' && value && value[IS_PROXY];
```
rwjblue: I think it may be better to do `typeof value === 'object' && value !== null` instead of coercing `value` to a boolean here.

chancancode: I just memorized the dance @krisselden told me :P

rwjblue: HAHA! Sounds good!

krisselden:
@rwjblue yes that is the only boolean coercion left,  it won't compile any others and the typeof prevents deopt on a different falsy coercion. The explicit check or the remaining coercion after the typeof check are the same at that point I lean toward the shorter one?

krisselden:
Likely though if isProxy is inlined into a hot path it will deopt on the first proxy it sees since they aren't very common anymore outside of ED promise proxy

chancancode:
The last commit removed all usage of `CURRENT_TAG`s 🎉

chancancode:
This took care of the "big page re-rendering regression" on skylight (the problem was that we have a big list with ED promise proxies in each row)

krisselden:
If you bump a git npm dep like glimmer you need to clear caches on Travis or the blueprints test will fail.  Not sure why but it consistently fails with the cache.

Issue #14001, opened 2016-08-02
[Glimmer 2] textarea renders "true" instead of ""


GavinJoyce:
The rendered value of `textarea` is `"true"` instead of `""` when the context value is `""`.

TODO:

 * [x] Create a failing test @gavinjoyce: https://github.com/tildeio/glimmer/pull/250
 * [x] Fix: https://github.com/tildeio/glimmer/pull/250

Given the following controller:

```js
export default Ember.Controller.extend({
  textareaContent1: '',
  textareaContent2: 'some default text',
});
```

And template:

```hbs
textareaContent1: {{textarea value=textareaContent1}} <br />
textareaContent2: {{textarea value=textareaContent2}} <br />
```

**htmlbars renders:**

<img width="626" alt="screen shot 2016-08-02 at 12 01 05" src="https://cloud.githubusercontent.com/assets/2526/17326459/753ebd86-58a9-11e6-8c02-d36919275173.png">

**glimmer renders:**

<img width="597" alt="screen shot 2016-08-02 at 12 00 36" src="https://cloud.githubusercontent.com/assets/2526/17326460/7906c12a-58a9-11e6-8a88-b009f22f8820.png">

Sample App: https://github.com/GavinJoyce/glammer (see http://localhost:4200/issue3)

This doesn't happen for `{{input}}`.

GavinJoyce:
We're calling `normalizePropertyValue` for `value` on `textarea` which sets a `""` value to `true`:

https://github.com/tildeio/glimmer/blob/9ecc9ade7efa41cf7068729ddeea4b09c1ce51de/packages/glimmer-runtime/lib/dom/props.ts#L33-L39

rwjblue:
Reopening here until we bump glimmer versions.

cibernox:
This is the only bit preventing EPS from work in glimmer2. I've even reduced the reproduction even more: https://ember-twiddle.com/64618480b3b1507f8d73475007f9b2a0?openFiles=templates.application.hbs%2C

PR #14021, opened 2016-08-04
[Glimmer 2] failing test for #14001

GavinJoyce:
Failing test for https://github.com/emberjs/ember.js/issues/14001

screen shot 2016-08-04 at 15 48 55

### DIFF
--- a/packages/ember-glimmer/tests/integration/helpers/text-area-test.js
+++ b/packages/ember-glimmer/tests/integration/helpers/text-area-test.js
@@ -120,6 +120,22 @@ moduleFor('Helpers test: {{textarea}}', class extends TextAreaRenderingTest {
     this.assertTextArea({ value: 'A beautiful day in Seattle' });
   }
 
+  ['@test GH#14001 Should correctly handle an empty string bound value']() {
+    this.render('{{textarea value=message}}', { message: '' });
+
+    this.assert.strictEqual(this.firstChild.value, '');
+
+    this.assertStableRerender();
+
+    this.runTask(() => set(this.context, 'message', 'hello'));
+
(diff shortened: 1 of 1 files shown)


rwjblue:
@GavinJoyce - I wonder if we can:

*  remove `value` from `attributeBinding` [here](https://github.com/emberjs/ember.js/blob/3d25f9805cf1eb37a3f5ae6e05e3e2646b4f6175/packages/ember-htmlbars/lib/components/text_area.js#L41)
* add a template for it ([in this folder](https://github.com/emberjs/ember.js/tree/3d25f9805cf1eb37a3f5ae6e05e3e2646b4f6175/packages/ember-glimmer/lib/templates)) that has a single `{{value}}` in it

GavinJoyce:
trying...

GavinJoyce:
@rwjblue spike here : https://github.com/emberjs/ember.js/compare/master...GavinJoyce:gj/spike-textarea-template

There are at least two failing tests:

screen shot 2016-08-04 at 16 09 40

rwjblue:
Based on that failing test, I would say that [this method](https://github.com/emberjs/ember.js/blob/master/packages/ember-views/lib/mixins/text_support.js#L199) isn't firing when those events are triggered.

GavinJoyce:
@rwjblue I've [pushed a "fix"](https://github.com/emberjs/ember.js/commit/0509c8903177ddfa25c0c8e0e61da171216c4089) for that failing test, the htmlbars one has three `childNodes` (two for wrapping the dynamic node I suspect) while the glimmer one has just one.

I'm not sure how this strategy would work with user input though, especially in htmlbars. I'd like to write some tests that actually send keys to the text area. Do you know of anything like this elsewhere in ember tests?

Issue #14015, opened 2016-08-03
Engines + visit + shouldRender: false incompatible


stefanpenner:
It appears (debugging with @rwjblue) that  the follwing in an engine's route, will currently always assume `renderTemplate` should be invoked.

```js
if (!this._environment || this._environment.options.shouldRender) {
```

source: https://github.com/emberjs/ember.js/blob/13ea22bd54a5e967fd4638704bb55b348fb07d32/packages/ember-routing/lib/system/route.js#L1235

it may be that: https://github.com/emberjs/ember.js/blob/master/packages/ember-application/lib/system/engine-instance.js#L170-L184 should also clone '-environment:main' forward.

----

### The failure scenarios:

* For one, we render (or attempt to render) even though we explicitly ask not to
* A hard failure is also possible:

```
Error in Ember run loop: TypeError: Cannot read property 'outlets' of undefined
```

That we are attempting to render (when are asked not ot), but for other reasons (not 100% yet) after the above branch fails, we attempt a render, then during `setupOutlets` we take [this branch](https://github.com/emberjs/ember.js/blob/9ebda97047be7f8ce2c5b3f4e4568cdc25844dea/packages/ember-routing/lib/system/router.js#L290) before `defaultParentState` has been [assigned](https://github.com/emberjs/ember.js/blob/9ebda97047be7f8ce2c5b3f4e4568cdc25844dea/packages/ember-routing/lib/system/router.js#L292), in this case `findLiveRoute` returns no live route [here](https://github.com/emberjs/ember.js/blob/9ebda97047be7f8ce2c5b3f4e4568cdc25844dea/packages/ember-routing/lib/system/router.js#L1325) which ends up \w us taking the [else](https://github.com/emberjs/ember.js/blob/9ebda97047be7f8ce2c5b3f4e4568cdc25844dea/packages/ember-routing/lib/system/router.js#L1335) and throwing because `defaultParentState` is still undefined

stefanpenner:
action items:

- [ ] investigate if
- [ ] add assertion/throw to https://github.com/emberjs/ember.js/blob/9ebda97047be7f8ce2c5b3f4e4568cdc25844dea/packages/ember-routing/lib/system/router.js#L1332 when we end up in this case.
- [ ] refactor https://github.com/emberjs/ember.js/blob/9ebda97047be7f8ce2c5b3f4e4568cdc25844dea/packages/ember-routing/lib/system/router.js#L278-L293 to be a-bit easier to follow
[...]

PR #14025, opened 2016-08-04
[BUGFIX] Pass environment options forward to Engines

trentmwillis:
Addressing #14015.

There are two commits here. The first simply adds two tests to verify the behavior of `shouldRender: false` in the `visit` API. The latter of those tests failed with an error due to attempting to render when it shouldn't (as described in the issue).

The second commit fixes the failing test by passing `-environment:main` from the parent into an EngineInstance. This also resulted in needing to update some tests to setup the registry in an ApplicationInstance that constructs the EngineInstance. I assume this will always happen in a real application, so I figured it made sense to do in the tests.

### DIFF
--- a/packages/ember-application/lib/system/engine-instance.js
+++ b/packages/ember-application/lib/system/engine-instance.js
@@ -179,8 +179,12 @@ if (isEnabled('ember-application-engines')) {
       [
         'router:main',
         P`-bucket-cache:main`,
-        '-view-registry:main'
+        '-view-registry:main',
+        '-environment:main'
       ].forEach(key => this.register(key, parent.lookup(key), { instantiate: false }));
+
+      this.inject('view', '_environment', '-environment:main');
+      this.inject('route', '_environment', '-environment:main');
     }
   });
 }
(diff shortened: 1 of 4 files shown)


packages/ember-application/lib/system/engine.js
```diff
@@ -499,6 +499,9 @@ function commonSetupRegistry(registry) {
 
   registry.injection('view:-outlet', 'namespace', 'application:main');
 
+  registry.injection('view', '_environment', '-environment:main');
+  registry.injection('route', '_environment', '-environment:main');
```
rwjblue: These injections should be done in the `engine-instance` (because the environment can change per `.visit` call).

trentmwillis: Good point. Updated.

stefanpenner:
@rwjblue / @trentmwillis is this correct to target master?

rwjblue:
The PR should target master with the commits prefixed with `[BUGFIX beta]` (since engines feature flag is enabled in beta).

trentmwillis:
Updated the commit tags.

rwjblue:
Thanks @trentmwillis!

rwjblue:
Pulled into beta branch, next build should fix for folks...

PR #14026, opened 2016-08-04
[DOC release] Use dash for partial naming

wagenet:


locks:
@wagenet you're sure this works? I was under the impression a coworker had to use an underscore to make it work.

teddyzeenny:
I can confirm this works.

PR #14028, opened 2016-08-04
Correct -normalize-class and AST plugin

chadhietala:
Codegen in HTMLBars seemed ok with PathExpressions being StringLiterals, however sense we have an interpreter in Glimmer these semantics actually matter. Also in Glimmer we appear to be actually expanding the AST into an sexpr with only positional args for `-normalize-class`.

### DIFF
--- a/packages/ember-glimmer/lib/helpers/-normalize-class.js
+++ b/packages/ember-glimmer/lib/helpers/-normalize-class.js
@@ -2,40 +2,16 @@ import { InternalHelperReference } from '../utils/references';
 import { dasherize } from 'ember-runtime/system/string';
 
 function normalizeClass({ positional, named }) {
+  let classNameParts = positional.at(0).value().split('.');
+  let className = classNameParts[classNameParts.length - 1];
   let value = positional.at(1).value();
-  let activeClass = named.at('activeClass').value();
-  let inactiveClass = named.at('inactiveClass').value();
 
-  // When using the colon syntax, evaluate the truthiness or falsiness
-  // of the value to determine which className to return.
-  if (activeClass || inactiveClass) {
-    if (!!value) {
-      return activeClass;
-    } else {
-      return inactiveClass;
-    }
-
-  // If value is a Boolean and true, return the dasherized property
-  // name.
-  } else if (value === true) {
-    let propName = positional.at(0);
-    // Only apply to last segment in the path.
-    if (propName) {
-      let segments = propName.split('.');
-      propName = segments[segments.length - 1];
-    }
-
-    return dasherize(propName);
-
-  // If the value is not false, undefined, or null, return the current
-  // value of the property.
-  } else if (value !== false && value != null) {
-    return value;
-
-  // Nothing to display. Return null so that the old class is removed
-  // but no new class is added.
+  if (value === true) {
+    return dasherize(className);
+  } else if (!value && value !== 0) {
+    return '';
   } else {
-    return null;
+    return String(value);
   }
 }
 
--- a/packages/ember-glimmer/tests/integration/components/class-bindings-test.js
+++ b/packages/ember-glimmer/tests/integration/components/class-bindings-test.js
@@ -48,51 +48,58 @@ moduleFor('ClassNameBindings integration', class extends RenderingTest {
   ['@test it can have class name bindings in the template']() {
     this.registerComponent('foo-bar', { template: 'hello' });
 
-    this.render('{{foo-bar classNameBindings="someInitiallyTrueProperty someInitiallyFalseProperty someInitiallyUndefinedProperty :static isBig:big isOpen:open:closed isUp::down"}}', {
-      someInitiallyTrueProperty: true,
-      someInitiallyFalseProperty: false,
-      isBig: true,
-      isOpen: false,
-      isUp: true
+    this.render('{{foo-bar classNameBindings="model.someInitiallyTrueProperty model.someInitiallyFalseProperty model.someInitiallyUndefinedProperty :static model.isBig:big model.isOpen:open:closed model.isUp::down model.bar:isTruthy:isFalsy"}}', {
+      model: {
+        someInitiallyTrueProperty: true,
+        someInitiallyFalseProperty: false,
+        isBig: true,
+        isOpen: false,
+        isUp: true,
+        bar: true
+      }
     });
 
     this.assertComponentElement(this.firstChild, {
-      attrs: { 'class': classes('ember-view some-initially-true-property static big closed') },
+      attrs: { 'class': classes('ember-view some-initially-true-property static big closed isTruthy') },
       content: 'hello'
     });
 
     this.runTask(() => this.rerender());
 
     this.assertComponentElement(this.firstChild, {
-      attrs: { 'class': classes('ember-view some-initially-true-property static big closed') },
+      attrs: { 'class': classes('ember-view some-initially-true-property static big closed isTruthy') },
       content: 'hello'
     });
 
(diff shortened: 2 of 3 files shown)


GavinJoyce:
👍 thanks. The build failure looks to be sauce labs flakiness

packages/ember-glimmer/lib/helpers/-normalize-class.js
```diff
@@ -2,17 +2,21 @@ import { InternalHelperReference } from '../utils/references';
...
   // When using the colon syntax, evaluate the truthiness or falsiness
   // of the value to determine which className to return.
   if (activeClass || inactiveClass) {
     if (!!value) {
-      return activeClass;
+      return dasherize(activeClass);
     } else {
-      return inactiveClass;
+      return dasherize(inactiveClass);
     }
```
chancancode: I don't think we need to dasherize here – we only do that when you use the "no colon" syntax and the value evaluates to a Boolean. Here (when using the colon syntax)  we expect these to be string literals

chancancode: We do need to return `null` if the value doesn't exists though.

chancancode: We do need to return `null` if the value doesn't exists though.

chancancode:
I wonder if we could just do most of the work in the AST transform.

1. `:foo` -> `"foo"`
2. `foo:bar:baz` -> `(if foo "bar" "bas")`
3. `foo:bar` -> `(if foo "bar" null)`
4. `foo::bar` -> `(if foo null "baz")`
5. `foo` -> `(-normalize-class foo name="foo")`

Then 1-4 doesn't need any special runtime stuff, and `-normalize-class` becomes (more or less) a simple helper `([value], {name}) => value  === true ? dasherize(name) : (value || null)`.

chadhietala:
@chancancode Gavin's failing test transforms into this:

```handlebars
{{foo-bar class=(concat 
  (if someInitiallyTrueProperty
    (-normalize-class "someInitiallyTrueProperty" someInitiallyTrueProperty)) " "
  (if someInitiallyFalseProperty
    (-normalize-class "someInitiallyFalseProperty" someInitiallyFalseProperty)) " "
  (if someInitiallyUndefinedProperty
    (-normalize-class "someInitiallyUndefinedProperty" someInitiallyUndefinedProperty)) " " "static" " "
  (if isBig "big") " "
  (if isOpen "open" "closed") " "
  (if isUp "" "down") " ") }}
```

chancancode:
@chadhietala in that case most of the code in https://github.com/chadhietala/ember.js/blob/8e49c29d78af1b9f00a8e9528ed225191bf003fb/packages/ember-glimmer/lib/helpers/-normalize-class.js can be deleted – the `activeClass` and `inactiveClass` stuff is for the colon syntax. Also, since `-normalize-class` already handles the truthiness check, these is no need to wrap it around an extra `if` (i.e. `(if someInitiallyTrueProperty
    (-normalize-class "someInitiallyTrueProperty" someInitiallyTrueProperty))` should just be `(-normalize-class "someInitiallyTrueProperty" someInitiallyTrueProperty)`.

chancancode:
We should also do the "extract the last segment of the path" thing in the AST plugin. With that, the `-normalize-class.js` should just be:

```javascript
import { helper } from '../helper';
import { dasherize } from 'ember-runtime/system/string';

function normalizeClass([name, value]) {
  // If value is `true`, return the dasherized property name.
  if (value === true) {
    return dasherize(name);
  } else if (!value) {
    return "";
  } else {
    return String(value);
  }
}

export default helper(normalizeClass);
```

chadhietala:
So I ended up splitting out the AST transforms here because we don't need to do nearly as much work in Glimmer and can expand into something that is smaller. In Glimmer `-normalize-class` now looks like this `(-normalize-class predicateRef consequentRefOrString alternativeRefOrString)`.

chancancode:
For `foo.bar`, I assume you need to emit `(-normalize-class foo.bar "bar")`, is there code/tests for that?

chancancode:
In what scenario would `consequentRefOrString` `alternativeRefOrString` be not strings? I understand that they would be represented as a (const) reference, but it seems like, if supplied, they are always const strings

packages/ember-glimmer-template-compiler/lib/plugins/transform-old-class-binding-syntax.js
```diff
@@ -0,0 +1,111 @@
...
+    // :my-class-name microsyntax for static values
+    if (propName === '') {
+      sexpr = b.string(activeClass);
+    } else {
+      let params = [b.path(propName)];
+
+      if (activeClass || activeClass === '') {
+        params.push(b.string(activeClass));
+      } else {
+        params.push(b.string(propName));
```
chancancode: I think this needs to be split over `.`, and if you want to go this route, you would have to do the dasherize here, or otherwise distinguish that from the `foo:barBaz` case where `barBaz` should not be dasherized (I believe?)

packages/ember-glimmer/lib/helpers/-normalize-class.js
```diff
@@ -2,40 +2,22 @@ import { InternalHelperReference } from '../utils/references';
...
+  if (positional.at(2)) {
+    inactiveClass = positional.at(2).value();
+  }
 
-  // Nothing to display. Return null so that the old class is removed
-  // but no new class is added.
+  if (value === true) {
+    return dasherize(activeClass);
+  } else if (value === false && inactiveClass) {
+    return dasherize(inactiveClass);
```
chancancode: I don't think we `activeClass` and `inactiveClass` should be `dasherize`d

packages/ember-glimmer/lib/helpers/-normalize-class.js
```diff
@@ -2,40 +2,22 @@ import { InternalHelperReference } from '../utils/references';
...
-  // value of the property.
-  } else if (value !== false && value != null) {
-    return value;
+  if (positional.at(2)) {
+    inactiveClass = positional.at(2).value();
+  }
 
-  // Nothing to display. Return null so that the old class is removed
-  // but no new class is added.
+  if (value === true) {
```
chancancode: should just be a truthiness check now

chancancode:
Hm actually, I think I don't fully understand this strategy now, after looking at the code.

It doesn't seem to (have enough information to) handle some cases correctly

packages/ember-glimmer/tests/integration/components/class-bindings-test.js
```diff
@@ -45,6 +45,58 @@ moduleFor('ClassNameBindings integration', class extends RenderingTest {
     this.assertComponentElement(this.firstChild, { tagName: 'div', attrs: { 'class': classes('ember-view foo enabled sad') }, content: 'hello' });
   }
 
+  ['@test it can have class name bindings in the template']() {
+    this.registerComponent('foo-bar', { template: 'hello' });
+
+    this.render('{{foo-bar classNameBindings="someInitiallyTrueProperty someInitiallyFalseProperty someInitiallyUndefinedProperty :static isBig:big isOpen:open:closed isUp::down"}}', {
```
krisselden: in order for this to follow the INUR pattern, you need have one more nested level that you can interior update then replace

chancancode:
Looks good to me!

@chadhietala follow up:

1. refactor the args ordering?
2. split the path in the transform

chadhietala:
Following up on these, but there is some HTMLBars WAT stuff I'm working through.

PR #14030, opened 2016-08-05
[GLIMMER] Fix OutletView#appendTo when jQuery is not present.

rwjblue:


krisselden:
I don't get the story between importing environment and the bootoptions for visit plus injection.  The story seems very inconsistent. Also, same for the import of jquery

rwjblue:
@krisselden - I can remove from the glimmer side because we don't support instantiating without DI yet, but in HTMLBars without the fallback things break when doing `View.create().appendTo('#ember-testing')`.

mixonic:
👍

PR #14033, opened 2016-08-05
[BUGFIX beta] `fillIn` only sets value to first match.

### LINKED ISSUE #14018: FillIn acceptance helper works only on the first matched input
I tried to `fillIn` multiple inputs in an acceptance test with one selector. So let's say my template was:

```
{{input value=someVal}}
{{input value=otherVal}}
```

And I was using `fillIn` like this: `fillIn('input', 'someval')`

The result would be that visually it would seem as if both inputs are filled in with the same value (html was updated), but the data binding would only be updated on `someVal`.

Replicated here: https://ember-twiddle.com/63d30b6cde4d82491c0def9bcc36ef28

(sidenote: I couldn't use Ember 2.7 due to a twiddle bug, but the same thing is happening in an app using Ember 2.7)

Serabe:
Currently, `fillIn` sets the new value to all matches, though the events are
only fired for the first one, making Ember only aware of that one.

This can cause a problem while trying to _debug_ a test.

Fixes #14018

### DIFF
--- a/packages/ember-testing/lib/helpers/fill_in.js
+++ b/packages/ember-testing/lib/helpers/fill_in.js
@@ -11,7 +11,7 @@ export default function fillIn(app, selector, contextOrText, text) {
   el = $el[0];
   focus(el);
 
-  $el.val(text);
+  $el.eq(0).val(text);
   fireEvent(el, 'input');
   fireEvent(el, 'change');
 
--- a/packages/ember-testing/tests/helpers_test.js
+++ b/packages/ember-testing/tests/helpers_test.js
@@ -722,6 +722,28 @@ QUnit.test('`fillIn` fires `input` and `change` events in the proper order', fun
   return wait();
 });
 
+QUnit.test('`fillIn` only sets the value in the first matched element', function() {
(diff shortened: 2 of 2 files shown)


Serabe:
IMHO, `fillIn` should raise an error if more than one element matches (like the `Capybara::Ambiguous` error). I understand that that would be backwards incompatible and would need to wait until 3.0 or to be added to the [Grand Testing Unification RFC](https://github.com/emberjs/rfcs/pull/119)

rwjblue:
@Serabe - Good point RE: ambiguous match. Can you comment on the RFC about that so I remember to incorporate it?

Serabe:
Thank you. [Just did!](https://github.com/emberjs/rfcs/pull/119#issuecomment-237914054)

rwjblue:
Also, this is good to go once CI passes.  Though at some point (hopefully soon) I need to come back and remove the remaining jQuery usage in the acceptance test helper stuff. We switched to native events (which is awesome), but still have some jQueryisms in here to fix.

rwjblue:
Thanks again @Serabe!

PR #14037, opened 2016-08-06
[DOC release] Boolean `attributeBindings` documentation updated

### LINKED ISSUE #14024: attributeBinding for booleans broken in 2.7
In earlier versions, view attribute binding for boolean properties would have generated a tag similar to below:

 `<div id="ember444" bindingval="true" class="ember-view">`

Now it is not displaying the attribute at all if false and doing the following if true

 `<div id="ember444" bindingval class="ember-view">`

This can be verified by creating a brand new app and rendering a simple component setup like below.  

```
import Ember from 'ember';

    bindingVal: true,
    attributeBindings: ['bindingVal']
});

```

Serabe:
From [HTML5, A vocabulary and associated APIs for HTML and
XHTML](https://www.w3.org/TR/html5/infrastructure.html#boolean-attributes):

```
A number of attributes are boolean attributes. The presence of a boolean
attribute on an element represents the true value, and the absence of the
attribute represents the false value.

**Note:** The values "true" and "false" are not allowed on boolean
attributes. To represent a false value, the attribute has to be omitted
altogether.
```

Fixes #14024

### DIFF
--- a/packages/ember-views/lib/views/view.js
+++ b/packages/ember-views/lib/views/view.js
@@ -255,7 +255,7 @@ import ViewMixin from 'ember-views/mixins/view_support';
   ```
 
   If the return value of an `attributeBindings` monitored property is a boolean
-  the property's value will be set as a coerced string:
+  the attribute will be present or absent depending on the value:
 
   ```javascript
   MyTextInput = Ember.View.extend({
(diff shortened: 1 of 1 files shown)


rwjblue:
Thank you!

PR #14038, opened 2016-08-06
[DOC release] Fix `setEach` documentation

### LINKED ISSUE #14022: Why use setAll is the most efficient way to do a setEach?
Hi, according to ember documentation Ember.enumerable setEach 'is more
    efficient than using other methods defined on this helper':

If we drop down to the ember code we can see a function that uses a vanila JS foreach and set method extracted from ember-metal.

```javascript
/**
    Sets the value on the named property for each member. This is more
    efficient than using other methods defined on this helper. If the object
    implements Ember.Observable, the value will be changed to `set(),` otherwise
    it will be set directly. `null` objects are skipped.
    @method setEach
    @param {String} key The key to set
    @param {Object} value The object to set
    @return {Object} receiver
    @public
  */
  setEach(key, value) {
    return this.forEach(function(item) {
      set(item, key, value);
    });
  },

```

Why this is the most 'efficient' way to do that? Its because we use set as a method instead of do item.set?

Thank you

Serabe:
As suggested in #14022, `setEach` is a more ergonomic, not more efficient
method.

Fixes #14022

### DIFF
--- a/packages/ember-runtime/lib/mixins/enumerable.js
+++ b/packages/ember-runtime/lib/mixins/enumerable.js
@@ -309,7 +309,7 @@ const Enumerable = Mixin.create({
 
   /**
     Sets the value on the named property for each member. This is more
-    efficient than using other methods defined on this helper. If the object
+    ergonomic than using other methods defined on this helper. If the object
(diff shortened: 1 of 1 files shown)


rwjblue:
Thanks!

PR #14040, opened 2016-08-06
More robustly fully scope query params

### LINKED ISSUE #14010: Exception thrown from _fullyScopeQueryParams
Since upgrading to 2.7, our app has been throwing errors in production on some transitions. We have around 600 errors recorded in the past week.

I haven't been able to reproduce these in a development environment, so all I have to go on is a stack trace. But it looks to me like there's something going askew on the path from `transitionTo` to the new `_fullyScopeQueryParams` introduced in #13273 by @raytiley.

The error we're seeing is:

> Cannot use 'in' operator to search for 'type' in undefined

(The exact query param searched for varies.)

Looking at [the definition of this method](https://github.com/emberjs/ember.js/blob/67081fa/packages/ember-routing/lib/system/router.js#L789-L813), the only use of the `in` operator is against its `queryParams` argument, which therefore must somehow be `undefined`.

Backing up the stack trace, [this parameter is set in `transitionTo`](https://github.com/emberjs/ember.js/blob/67081fa/packages/ember-routing/lib/system/router.js#L353-L368), and the only path I can see where it could possibly be undefined is this code:

```js
if (possibleQueryParams && possibleQueryParams.hasOwnProperty('queryParams')) {
  queryParams = args.pop().queryParams;
} else {
  queryParams = {};
}
```

I can't find anything in our app or dependencies that is doing strange things with queryParams in links or transitions, but my knowledge of the router guts ends here.

So while I'm stuck about the root cause, all I can suggest is that we could add a fallback. If the queryParams property of the popped arg isn't an object, default to `{}`, as in the other branch of the if statement.

Would that approach be acceptable? Or can anyone suggest further steps I could take to track down the root cause of the bug?

alisdair:
Fixes #14010.

Unfortunately I can't add a test for this change, as I don't have a reliable reproduction for the bug. But in production I'm seeing exceptions indicating that `queryParams` is undefined in some cases, and it seems reasonable for this function to be robust to that case by shortcut-return.

### DIFF
--- a/packages/ember/tests/routing/query_params_test.js
+++ b/packages/ember/tests/routing/query_params_test.js
@@ -3389,3 +3389,37 @@ QUnit.test('handle routes names that clash with Object.prototype properties', fu
   let controller = container.lookup('controller:constructor');
   equal(get(controller, 'foo'), '999');
 });
+
+QUnit.test('undefined queryParams is treated as empty object', function() {
+  expect(1);
+
+  App.Router.map(function() {
+    this.route('foo');
+    this.route('bar');
+  });
+
+  App.FooRoute = Route.extend({
+    beforeModel() {
+      this.transitionTo('bar', { queryParams: undefined });
+    }
+  });
+
+  App.BarController = Controller.extend({
+    queryParams: ['a']
+  });
+
+  App.BarRoute = Route.extend({
+    queryParams: {
+      a: {
+        defaultValue: 'xyz'
+      }
+    }
+  });
+
+  bootApplication();
+
+  run(router, 'transitionTo', 'foo', { queryParams: { a: '123' } });
+
+  let barController = container.lookup('controller:bar');
+  equal(get(barController, 'a'), '123');
+});
--- a/packages/ember/tests/routing/query_params_test.js
+++ b/packages/ember/tests/routing/query_params_test.js
@@ -3351,6 +3351,40 @@ if (isEnabled('ember-routing-route-configured-query-params')) {
       promiseResolve();
     });
   });
+
+  QUnit.test('undefined queryParams is treated as empty object', function() {
+    expect(1);
+
+    App.Router.map(function() {
+      this.route('foo');
+      this.route('bar');
+    });
+
+    App.FooRoute = Route.extend({
+      beforeModel() {
+        this.transitionTo('bar', { queryParams: undefined });
+      }
+    });
+
+    App.BarController = Controller.extend({
+      queryParams: ['a']
+    });
+
+    App.BarRoute = Route.extend({
+      queryParams: {
+        a: {
+          defaultValue: 'xyz'
+        }
+      }
+    });
+
+    bootApplication();
+
+    run(router, 'transitionTo', 'foo', { queryParams: { a: '123' } });
+
+    let barController = container.lookup('controller:bar');
+    equal(get(barController, 'a'), '123');
+  });
 }
 
 QUnit.test('warn user that routes query params configuration must be an Object, not an Array', function() {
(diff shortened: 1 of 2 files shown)


rwjblue:
This may be correct (it is somewhat hard to reason about), but I'm hesitant to land this without a regression test.

alisdair:
@rwjblue Thanks for the poke. I've added a test that I hope makes sense.

supersabillon:
Tried this fix on our project using 2.7 and it worked!

packages/ember/tests/routing/query_params_test.js
```diff
@@ -3351,6 +3351,40 @@ if (isEnabled('ember-routing-route-configured-query-params')) {
...
+    expect(1);
+
+    App.Router.map(function() {
+      this.route('foo');
+      this.route('bar');
+    });
+
+    App.FooRoute = Route.extend({
+      beforeModel() {
+        this.transitionTo('bar', { queryParams: undefined });
```
rwjblue: Does this error only occur when explicitly passing `undefined`?  Would the following also fail:

```js
this.transitionTo('bar');

// or
this.transitionTo('bar',  {});
```

?

alisdair: Yes, the error only seems to occur when passing `undefined`. I still don't know what elsewhere in the system is causing something to call `transitionTo('…', { queryParams: undefined })`, but that's the cause of all the app errors.

[Here's the code which is trying to guess if we have queryParams or not](https://github.com/emberjs/ember.js/blob/67081fa/packages/ember-routing/lib/system/router.js#L359-L364):

```js
    let possibleQueryParams = args[args.length - 1];
    if (possibleQueryParams && possibleQueryParams.hasOwnProperty('queryParams')) {
      queryParams = args.pop().queryParams;
    } else {
      queryParams = {};
    }
```

So:

- `this.transitionTo('bar')` works fine without my change, because `possibleQueryParams` is falsy and so we hit the second block of the `if`

- `this.transitionTo('bar', {})` also hits the second block of the `if`, because `{}` fails the second condition, but then (for unrelated reasons) then results in "Error: More context objects were passed than there are dynamic segments for the route: bar".

rwjblue: Hmm. I would really like to understand what is actually causing the `queryParams` object to be undefined....

alisdair: Me too, but I have no idea how I can possibly find that out. I've looked through the router code as closely as I can, and haven't found any possible explanation. Nor have I found any reproduction in development. Any suggestions?

Is understanding the root cause a blocker to getting this merged? I know it's a band-aid for the underlying problem, but the code-path causing this crash still seems like incorrect behaviour to me, even if we don't know what's triggering it.

alisdair:
Digging through a few stack traces a little further, it looks like the invalid `queryParams` values are coming from `link-to` component instances.

[In `_invoke`, we pull `values` out of `queryParams`](https://github.com/emberjs/ember.js/blob/v2.7.0/packages/ember-htmlbars/lib/components/link-to.js#L647) before passing the result to `transitionTo`:

```js
let queryParams = get(this, 'queryParams.values');
```

But [in `didReceiveAttrs` our fallback value of `queryParams` is `{}`](https://github.com/emberjs/ember.js/blob/v2.7.0/packages/ember-htmlbars/lib/components/link-to.js#L802-L810). I think that means that `get(this, 'queryParams.values')` would return `undefined` in this case.

I have no idea what is happening in the internals here. I don't know what `queryParams.values` refers to, and I don't know how to find out. But is that possibly the underlying source of the problem?

rwjblue:
replaced by #14068

PR #14041, opened 2016-08-07
[GLIMMER2] Optimize {{property}} PropertyReference

chancancode:
A quick survey in our app indicates ~77% of `PropertyReference` have a const parent. This is unsurprising, because most path lookup curlies are single-level (no dots in the path) lookup on the component directly.

This means we can do less work and book-keeping in those cases (checking if the parent is a proxy on every `compute`, etc).

### DIFF
--- a/packages/ember-glimmer/lib/component.js
+++ b/packages/ember-glimmer/lib/component.js
@@ -47,6 +47,7 @@ const Component = CoreView.extend(
 
     init() {
       this._super(...arguments);
+      this[IS_DISPATCHING_ATTRS] = false;
       this[DIRTY_TAG] = new DirtyableTag();
       this[ROOT_REF] = null;
       this[REFS] = new EmptyObject();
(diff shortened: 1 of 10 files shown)


chancancode:
In my <not completely scientific> experiments, this accounts for over 10% improvements in initial render time.

krisselden:
@mmun I thought we had tests that allowed {{str.length}}

chancancode:
@krisselden that crossed my mind too, but I am just mirroring the current behavior

chancancode:
@krisselden while you are here, can you review https://github.com/emberjs/ember.js/pull/14041/commits/4ce09546103ae4dc1fee97320b90add4f6ff7c37 ?

chancancode:
There seems to be additional wins associated with the latest commits, but my setup is too noisy to test that at the moment

krisselden:
@chancancode this is supported with the glimmer flag off https://ember-twiddle.com/cb89209f43d67d81fb74eaa59d442200?openFiles=templates.application.hbs%2C

PR #14044, opened 2016-08-08
[BUGFIX Beta] Bump router.js so that getHandlers is invoked lazily

trentmwillis:
This picks up recent changes from router.js. In particular, it makes `getHandler` lazy so that the work is not done upfront if it can be deferred until later (as is needed for lazy Engines). [See the PR for the changes here](https://github.com/tildeio/router.js/pull/187).

It also picks up [the shaping fixes](https://github.com/tildeio/router.js/pull/181) that @chadhietala made.

cc/ @rwjblue


rwjblue:

trentmwillis:
Sure thing, was unsure what prefix to use.

trentmwillis:
@rwjblue looks like this should be good to go now.

PR #14047, opened 2016-08-09
[DOC beta] Improve `Route#render` documentation

### LINKED ISSUE #14046: Documentation for render.into incorrect?
From the documentation here http://emberjs.com/api/classes/Ember.Route.html#method_render it seems like I should be able to render content into any arbitrary template, for example, given the (admittedly contrived) file structure:
```
templates/
  application.hbs
  error.hbs
  errorpage.hbs
```

it seems I should be able to do the following from any route:

```
renderTemplate() {
  this.render('error', {
    into: 'errorpage',
    outlet: 'error-message'
  });
}
```

However, this results in the error `Assertion Failed: You attempted to render into 'errorpage' but it was not found`

Looking into the code, I find that in `appendLiveRoute()`, the `liveRoutes` parameters only contains the parents of the current route, and therefore any template outside of that path will not be found.

I think the documentation needs to be updated to specify that you can only render into templates that exist in the path between the default root template and the route you are attempting to render into from.

Serabe:
This PR makes explicit that render only accepts a live template
as its `into` option.

Fix #14046

### DIFF
--- a/packages/ember-routing/lib/system/route.js
+++ b/packages/ember-routing/lib/system/route.js
@@ -1878,7 +1878,8 @@ let Route = EmberObject.extend(ActionHandler, Evented, {
     The string values provided for the template name, and controller
     will eventually pass through to the resolver for lookup. See
     Ember.Resolver for how these are mapped to JavaScript objects in your
-    application.
+    application. The template to render into needs to be related to  either the
+    current route or one of its ancestors.
 
     Not all options need to be passed to `render`. Default values will be used
     based on the name of the route specified in the router or the Route's


Serabe:
Renaming to `[DOC beta]` given that depends on https://github.com/emberjs/ember.js/commit/744e44cab326e038f7cd3e4a9c74ca70efa5c8c0

packages/ember-routing/lib/system/route.js
```diff
@@ -1878,7 +1878,8 @@ let Route = EmberObject.extend(ActionHandler, Evented, {
     The string values provided for the template name, and controller
     will eventually pass through to the resolver for lookup. See
     Ember.Resolver for how these are mapped to JavaScript objects in your
-    application.
+    application. The template needs to be either the one of the current route or one
```
locks: I think there's some extra words in here

Serabe: `needs to be related to either the current route or one of its ancestors`?

rwjblue: @Serabe - Yeah, that sounds better.

teddyzeenny: Should this be "The template to render into needs to be" (so that it's clear that these rules only apply to the `into` template and not to the template we want to render)?

mixonic:
👍 thanks!

PR #14048, opened 2016-08-10
Remove dead code

mitchlloyd:
Seeing this comment regarding "IE 6/7" made me wonder where this function was used. A quick search showed that it isn't used inside of this repo. I used GitHub's search to confirm that this function wasn't used in the `emberjs` or `ember-fastboot` organizations.


rwjblue:
:+1: awesome, thank you!

PR #14051, opened 2016-08-10
[DOC] Fix code blocks for includes

bmeurant:
As shown below, code block seems broken for `Enumerable#includes` and `Array#includes`

includes

I suspect missing line breaks before and after code block and I added it in this PR.

I cannot confirm because I was not able to reproduce locally. yuidoc generation is working well, even without line breaks.

If someone gives me instructions to reproduce exact steps for api docs generations, I can try to reproduce.


locks:
I haven't done it in a while, but it should be a matter of running [`npm run docs`](https://github.com/emberjs/ember.js/blob/master/package.json#L16). Let me know if it works!

bmeurant:
@locks yes I saw that and tested it already but generated yuidoc (without libe breaks fix) is fine as shown below.

includes2

... parsing seems ok there, I tested it against v2.7.0 tag and master

locks:
Thanks for your contribution!

PR #14053, opened 2016-08-11
chore: remove submodules

MylesBorins:
[According to rwjblue](https://github.com/nodejs/citgm/pull/159#issuecomment-239156530) the submodules are not used anymore


rwjblue:
Thank you!

PR #14054, opened 2016-08-11
[BUGFIX beta] Ensure substates properly work with resetNamespace

trentmwillis:
This fixes two apparent issues with substates. The first was discovered when [enabling substates for the `mount` keyword](https://github.com/emberjs/ember.js/pull/13818), but while fixing that a second seems to have been uncovered.

When looking up substates, the `findChildRouteName` method was always applying the parent route's namespace, even if the current route had `resetNamespace` applied. Some of the tests even seemed to be explicitly working around this issue (such as [here](https://github.com/emberjs/ember.js/blob/6558db677105a15efda5f35dcc7d9a70beff1984/packages/ember/tests/routing/substates_test.js#L797-L801) and [here](https://github.com/emberjs/ember.js/blob/6558db677105a15efda5f35dcc7d9a70beff1984/packages/ember/tests/routing/substates_test.js#L840-L845)), so maybe this was intended, but per the discussion in the other thread, it seems like no one currently expects it.

Additionally, `_loading` substates were having `resetNamespace` applied, but `_error` ones were not. This leads to inconsistent and unintuitive behavior. Again, maybe intended, but seems wrong.

### DIFF
--- a/packages/ember-routing/lib/system/dsl.js
+++ b/packages/ember-routing/lib/system/dsl.js
@@ -40,7 +40,7 @@ DSL.prototype = {
 
     if (this.enableLoadingSubstates) {
       createRoute(this, `${name}_loading`, { resetNamespace: options.resetNamespace });
-      createRoute(this, `${name}_error`, { path: dummyErrorRoute });
+      createRoute(this, `${name}_error`, { resetNamespace: options.resetNamespace, path: dummyErrorRoute });
     }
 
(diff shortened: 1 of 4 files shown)


rwjblue:
This should target beta branch, can you prefix with `[BUGFIX beta]`?

trentmwillis:
@rwjblue updated.

PR #14055, opened 2016-08-11
Fix basic component and helper resolution in engines

asakusuma:
Ensured that tests fail before fix

cc @krisselden @chancancode

### DIFF
--- a/packages/ember-glimmer/lib/template.js
+++ b/packages/ember-glimmer/lib/template.js
@@ -2,6 +2,14 @@ import { Template } from 'glimmer-runtime';
 
 class Wrapper {
   constructor(id, env, spec) {
+    let { owner } = env;
+    if (spec.meta) {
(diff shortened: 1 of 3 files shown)


packages/ember-glimmer/lib/template.js
```diff
@@ -2,6 +2,14 @@ import { Template } from 'glimmer-runtime';
 
 class Wrapper {
   constructor(id, env, spec) {
```
krisselden: The factory should be passed the owner, we should use the owner of the template, the env really should come from the app and be passed into the engine with instantiate false.  /cc @dgeb

krisselden:
@dgeb please review

dgeb:
@asakusuma sorry to delay following up. @rwjblue and I talked with @krisselden and have some ideas about how to resolve this, which we'll try to do in the next few days.

PR #14056, opened 2016-08-12
features: remove ember-routable-components

maletor:
It looks like someone forgot to remove this feature since it does not exist in `features.json`.


rwjblue:
:+1:

maletor:
Err, should ember-glimmer be in FEATURES.md?

rwjblue:
Yes, generally speaking every entry in `features.json` should be in `FEATURES.md`.  However, in this case `ember-glimmer` doesn't really add any new public API's here.  

If you don't mind adding (and mentioning it doesn't add new public API) that would be great though!

PR #14059, opened 2016-08-13
Unify most remaining @htmlbars tests

chancancode:
Got a lot of them working in this session, here are the remaining ones on the glimmer side:

skipped


chancancode:
cc @krisselden

rwjblue:
The test changes look good to me.  I'll let someone else speak to the implementation changes....

krisselden:
Looks like there are linting failures

rwjblue:
I believe there are some non-linting failures too:

> Test failed: [glimmer] Helpers test: element action:  it should unregister event handlers when an element action is removed

> Test failed: [htmlbars] Helpers test: element action:  it should unregister event handlers when an element action is removed

PR #14065, opened 2016-08-14
[DOC relase] Improve `mut` documentation

### LINKED ISSUE #14057: mut helper does not bind in action parameter
When a `mut` helper is passed as a parameter to an action, the action receives the original value, not the mutable getter/setter.

https://ember-twiddle.com/88a012a865086742ab2d2601565f6f57?openFiles=controllers.application.js%2C

From my understanding, it should log a `{ update, value}` object.

Serabe:
`mut` documentation contains code that does not work. It also references a blog
post with outdated information.

This PR fixes the code with the right behaviour and remove the link to avoid
confussion. It also adds a section about _Legacy behaviour_ of this helper.

Fixes #14057

### DIFF
--- a/packages/ember-glimmer/lib/helpers/mut.js
+++ b/packages/ember-glimmer/lib/helpers/mut.js
@@ -7,22 +7,20 @@ import { INVOKE } from './action';
   The `mut` helper lets you __clearly specify__ that a child `Component` can update the
   (mutable) value passed to it, which will __change the value of the parent component__.
 
-  This is very helpful for passing mutable values to a `Component` of any size, but
-  critical to understanding the logic of a large/complex `Component`.
-
   To specify that a parameter is mutable, when invoking the child `Component`:
 
   ```handlebars
   {{my-child childClickCount=(mut totalClicks)}}
   ```
 
-  The child `Component` can then modify the parent's value as needed:
+  The child `Component` can then modify the parent's value just by modifying its own
+  property:
 
   ```javascript
   // my-child.js
   export default Component.extend({
     click() {
-      this.get('childClickCount').update(this.get('childClickCount').value + 1);
+      this.incrementProperty('childClickCount');
     }
   });
   ```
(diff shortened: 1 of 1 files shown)


packages/ember-glimmer/lib/helpers/mut.js
```diff
@@ -47,8 +45,13 @@ import { INVOKE } from './action';
 
   The `mut` helper changes the `totalClicks` value to what was provided as the action argument.
 
-  See a [2.0 blog post](http://emberjs.com/blog/2015/05/10/run-up-to-two-oh.html#toc_the-code-mut-code-helper) for
-  additional information on using `{{mut}}`.
+  ### Legacy behaviour
+
+  In a Component's `attrs`, a `mut`ted property is represented as an object with a `value`
```
locks: oh god :X

packages/ember-glimmer/lib/helpers/mut.js
```diff
@@ -47,8 +45,13 @@ import { INVOKE } from './action';
 
   The `mut` helper changes the `totalClicks` value to what was provided as the action argument.
 
-  See a [2.0 blog post](http://emberjs.com/blog/2015/05/10/run-up-to-two-oh.html#toc_the-code-mut-code-helper) for
-  additional information on using `{{mut}}`.
+  ### Legacy behaviour
```
rwjblue: Can we avoid this whole section? Do we have to document what is in `this.attrs`?

packages/ember-glimmer/lib/helpers/mut.js
krisselden: Let's please keep attrs out of the documentation

Serabe:
Removed. The only interest of the legacy section was that that behaviour was leaked in both the documentation and [The transition to Ember 2.0 in detail post](http://emberjs.com/blog/2015/05/10/run-up-to-two-oh.html#toc_the-code-mut-code-helper).

PR #14066, opened 2016-08-15
[BUGFIX beta] Update router.js to fix Phantom 1.9 issues.

simonihmig:
Follow up to #14062 that missed the changes added to router.js in https://github.com/tildeio/router.js/pull/190


rwjblue:
Thank you!

PR #14068, opened 2016-08-15
Default queryParams for link-to must have values

### LINKED ISSUE #14010: Exception thrown from _fullyScopeQueryParams

alisdair:
When transitioning between two routes with query params, if the user interrupts the active transition due to a slow model by visiting a third route with no query params, this causes an exception to be raised.

The root cause is that `link-to` has a default/fallback `queryParams` of `{}`, but it needs to have a `values` field. Otherwise, we pass `undefined` to the router, which blows up as a result (see #14040).

This PR fixes #14010 at the correct level, and adds a better regression test than #14040.

Thanks to @rwjblue and @GavinJoyce for all their help!

### DIFF
--- a/packages/ember-glimmer/lib/components/link-to.js
+++ b/packages/ember-glimmer/lib/components/link-to.js
@@ -805,7 +805,7 @@ const LinkComponent = EmberComponent.extend({
     if (lastParam && lastParam.isQueryParams) {
       queryParams = params.pop();
     } else {
-      queryParams = {};
+      queryParams = { values: {} };
     }
     this.set('queryParams', queryParams);
 
--- a/packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
+++ b/packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
@@ -8,6 +8,7 @@ import Application from 'ember-application/system/application';
 import jQuery from 'ember-views/system/jquery';
 import NoneLocation from 'ember-routing/location/none_location';
 import { setTemplates, set as setTemplate } from 'ember-templates/template_registry';
+import RSVP from 'ember-runtime/ext/rsvp';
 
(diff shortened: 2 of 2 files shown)


rwjblue:
screenshot of failing

rwjblue:
Thanks for working so hard at this @alisdair!

PR #14072, opened 2016-08-16
Glimmer cleanup

chancancode:


rwjblue:
Many of the failures are suggesting that `ember-view` isn't added to the list of classes.

> Failed assertion: Element's class attribute should match ember-view class-one class-two, expected: ember-view class-one class-two, but was: class-one class-two

PR #14078, opened 2016-08-17
Adds external helpers for babel

chadhietala:
Depends on https://github.com/emberjs/emberjs-build/pull/177

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -61,6 +61,7 @@ function getFeatures(environment) {
 function babelConfigFor(environment) {
   var plugins = [];
   var features = getFeatures(environment);
+  var includeDevHelpers = true;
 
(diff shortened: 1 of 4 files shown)


rwjblue:
Can you release and update emberjs-build version so that this works?

chadhietala:
I don't have release on the build 😄 .

rwjblue:
released v0.15.0

chadhietala:
Since Glimmer uses the emberjs-build this is simply going to strip the helpers from Ember's code.This brings up an interesting point and that is we need to pass config from Ember to Glimmer during build to say "you're not standalone" and make some decisions based on that. Just don't want to break the world for people working on standalone Glimmer.

I'm ok with landing this now and follow up with a plan to rev the Glimmer's dependency on emberjs-build.

PR #14085, opened 2016-08-18
Repro engines + visit test issue

asakusuma:
Repro https://github.com/emberjs/ember.js/issues/14086

### DIFF
--- a/packages/ember-application/tests/system/visit_test.js
+++ b/packages/ember-application/tests/system/visit_test.js
@@ -380,9 +380,9 @@ QUnit.test('visit() returns a promise that resolves without rendering when shoul
 
   assert.strictEqual(jQuery('#qunit-fixture').children().length, 0, 'there are no elements in the fixture element');
 
-  return run(App, 'visit', '/blog', { shouldRender: false }).then(instance => {
+  return run(App, 'visit', '/blog', { shouldRender: true }).then(instance => {
     assert.ok(instance instanceof ApplicationInstance, 'promise is resolved with an ApplicationInstance');
-    assert.strictEqual(jQuery('#qunit-fixture').children().length, 0, 'there are still no elements in the fixture element after visit');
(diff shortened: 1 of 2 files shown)


packages/ember-application/tests/system/visit_test.js
```diff
@@ -380,9 +380,9 @@ QUnit.test('visit() returns a promise that resolves without rendering when shoul
 
   assert.strictEqual(jQuery('#qunit-fixture').children().length, 0, 'there are no elements in the fixture element');
 
-  return run(App, 'visit', '/blog', { shouldRender: false }).then(instance => {
+  return run(App, 'visit', '/blog', { shouldRender: true }).then(instance => {
     assert.ok(instance instanceof ApplicationInstance, 'promise is resolved with an ApplicationInstance');
-    assert.strictEqual(jQuery('#qunit-fixture').children().length, 0, 'there are still no elements in the fixture element after visit');
+    assert.strictEqual(0, 0, 'there are still no elements in the fixture element after visit');
```
chancancode: This doesn't seem correct?

asakusuma: I'm just reproducing a problem, not actually expecting this to get merged. This set of changes should not fail the test suite, but it does.

chancancode: I see 👀

rwjblue:
I chatted with @krisselden and @dgeb earlier today about this, and we think we may have come up with an explanation here.  Basically, the renderer is not being shared between the host and the engine. This ends up causing both app renderer and engine renderer to setup the normal run looping things, and causes it to end up being out of sync.

Right now [this is the list](https://github.com/emberjs/ember.js/blob/master/packages/ember-application/lib/system/engine-instance.js#L179-L184) of instances shared between the host and the engine.  We think that the renderer and environment need to be added to that list.

rwjblue:
I believe this issue is resolved. Closing for now, but I can reopen if you see it crop up again.

PR #14087, opened 2016-08-18
[BUGFIX beta] Check that handler exists before triggering event

trentmwillis:
This is necessary for lazily-loaded routes where the handler is not
available synchronously. This includes events like loading and
queryParamsDidChange which trigger synchronously after starting a
transition, in those cases we should by-pass handlers that are not
yet loaded.

cc/ @rwjblue @nathanhammond @dgeb 

Note: I feel like I should add a test for this, but I feel like it makes more
sense to just test this in the ember-engines repo, since this code path is
impossible to hit normally.

### DIFF
--- a/packages/ember-routing/lib/system/router.js
+++ b/packages/ember-routing/lib/system/router.js
@@ -1059,7 +1059,7 @@ export function triggerEvent(handlerInfos, ignoreFailure, args) {
     handlerInfo = handlerInfos[i];
     handler = handlerInfo.handler;
 
-    if (handler.actions && handler.actions[name]) {
+    if (handler && handler.actions && handler.actions[name]) {
       if (handler.actions[name].apply(handler, args) === true) {
         eventWasHandled = true;
       } else {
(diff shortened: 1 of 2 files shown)


dgeb:
LGTM. I think I'm fine merging without an additional test.

dgeb:
@trentmwillis could you please prefix the commit message with `[BUGFIX beta]` ?

nathanhammond:
The other approach we considered was a promise chain to guarantee that the handler was already loaded, but that was breaking everything in Ember and doesn't actually change the behavior here.

trentmwillis:
@dgeb updated. I'm always unsure of what to prefix my commits with.

dgeb:
@krisselden That's fair. Let's hold off on merging until we can figure out the best way to test this.

trentmwillis:
Added two unit tests, let me know if you think more is needed. Opted for unit tests since the `triggerEvent` function is just an implementation of a hook from Router.js.

rwjblue:
These tests seem good for the changes being made here.

I would like to make sure that we get at least one higher level acceptance style test that emulates what we are actually doing in ember-engines WRT to `getHandler` returning a promise.  That doesn't need to block this PR per-se, but it seems fairly important to help avoid accidental regressions here...

trentmwillis:
@rwjblue I agree. Unfortunately, having `getHandler` return a Promise doesn't work without also modifying the default query params behavior, which is something I don't think we want to upstream yet since we haven't quite tested it out in the real world.

The only reason this PR is happening right now is because it seemed to make more sense than trying to monkeypatch it into the addon.

rwjblue:
@trentmwillis - Would you mind making a followup issue here to add an acceptance test for `getHandler` returning a promise?

trentmwillis:
Opened an issue to follow up.

PR #14091, opened 2016-08-19
[GLIMMER] Prevent auto-run assertion for objects not rendered.

rwjblue:
Prior to this change any `Ember.set` done after **any** rendering has been done during tests will trigger the auto-run assertion.

This change prevents the auto-run assertion when a given object has not been used in rendering process.

### DIFF
--- a/packages/ember-metal/lib/tags.js
+++ b/packages/ember-metal/lib/tags.js
@@ -41,8 +41,14 @@ if (hasGlimmer) {
   makeTag = () => new DirtyableTag();
 
   markObjectAsDirty = function(meta) {
-    ensureRunloop();
-    let tag = (meta && meta.readableTag()) || CURRENT_TAG;
+    let tag = meta && meta.readableTag();
+
+    if (tag) {
+      ensureRunloop();
+    } else {
+      tag = CURRENT_TAG;
+    }
+
     tag.dirty();
   };
 } else {
--- a/packages/ember-metal/tests/accessors/set_test.js
+++ b/packages/ember-metal/tests/accessors/set_test.js
@@ -1,7 +1,12 @@
 import { get } from 'ember-metal/property_get';
 import { set } from 'ember-metal/property_set';
+import { setHasViews } from 'ember-metal/tags';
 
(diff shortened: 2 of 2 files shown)


rwjblue:
@wycats / @krisselden - Would you mind reviewing this? If this is not correct, can you explain what I am forgetting/missing?

rwjblue:
FWIW, this fixes the errors under ember-alpha for ember-data (identified in https://github.com/emberjs/data/pull/4503).

chancancode:
I am 90% sure that this is fine, in fact we perhaps don't even need to bump `CURRENT_TAG` anymore (since nothing in the system uses it); however, I would appreciate more eyes and brains on this

krisselden:
I have doubts,  i'd like to know the motivation

rwjblue:
There are any number of things that you may be using `Ember.set` for that have nothing to do with rendering.  

One example is a simply creating an array proxy (`new Ember.ArrayProxy({ content: []});`), before this change this would trigger the auto-run assertion if any views had been rendered ([demo here](https://jsbin.com/himiru/edit?html,js,output)).  That seems quite wild (and potentially breaking) to me.

krisselden:
"There are any number of things that you may be doing that trigger a propertyDidChange event that have nothing to do with rendering."

This is a given, that is how it works now.

rwjblue:
I'm not sure I follow. 

On current alpha builds (before the change in this PR), anything that calls propertyDidChange (regardless of if it were used in the templating system) is going to trigger the auto-run assertion in test mode. 

This is definitely a change from current beta. I am unsure if it will affect folks trying to test glimmer or not...

krisselden:
After thinking about it, if the view layer puts the tag onto anything it renders and CURRENT_TAG is unused by the view system, and observers then are used behind the tagged object going from view system to object layer, I think this change is good, but it should return early, not dirty CURRENT_TAG and we have to make sure we have no more CURRENT_TAG usage in the view layer.

rwjblue:
Replaced by https://github.com/emberjs/ember.js/pull/14093.

PR #14095, opened 2016-08-19
Add helpful assertion when template is missing.

rwjblue:
Prior to this change, calling `component.appendTo()` when there was no `template` would trigger an unhelpful error:

```
Cannot read property 'asEntryPoint' of undefined
```

After this change, it will at least point you in the direction of the offending component:

```
You cannot render `<dummy@component:foo-bar::ember168>` without a template.
```


chadhietala:
LGTM

PR #14104, opened 2016-08-21
Call IsVisibleBinding only if no style binding

les2:
Previously, the `seen` variable was used as both an `Array` and a
hashmap (plain JavaScript object) when it was really an `Array`. The
`Array` would never return `true` for `seen['style']` in this code block
meaning that the `IsVisibleBinding` call would always be invoked,
regardless of whether there was an attribute binding for `style`. This
behavior is most likely a bug. The author intends to investigate unit /
integration tests for this case and add them in a separate commit.
# Unit Tests

The author is investigating tests now but wanted to submit this sooner rather than later in case it is important.

### DIFF
--- a/packages/ember-glimmer/tests/integration/components/curly-components-test.js
+++ b/packages/ember-glimmer/tests/integration/components/curly-components-test.js
@@ -2464,6 +2464,51 @@ moduleFor('Components test: curly components', class extends RenderingTest {
     });
   }
 
+  ['@test adds isVisible binding when style binding is missing and other bindings exist'](assert) {
+    let assertStyle = (expected) => {
+      let matcher = styles(expected);
+      let actual = this.firstChild.getAttribute('style');
+
+      assert.pushResult({
+        result: matcher.match(actual),
+        message: matcher.message(),
+        actual,
+        expected
+      });
+    };
+
+    this.registerComponent('foo-bar', {
+      ComponentClass: Component.extend({
+        attributeBindings: ['foo'],
+        foo: 'bar'
+      }),
+      template: `<p>foo</p>`
+    });
+
+    this.render(`{{foo-bar id="foo-bar" foo=foo isVisible=visible}}`, {
+      visible: false,
+      foo: 'baz'
+    });
+
+    assertStyle('display: none;');
+
+    this.assertStableRerender();
+
+    this.runTask(() => {
+      set(this.context, 'visible', true);
+    });
+
+    assertStyle('');
+
+    this.runTask(() => {
+      set(this.context, 'visible', false);
+      set(this.context, 'foo', 'woo');
+    });
+
+    assertStyle('display: none;');
+    assert.equal(this.firstChild.getAttribute('foo'), 'woo');
+  }
+
   ['@test it can use readDOMAttr to read input value']() {
     let component;
     let assertElement = (expectedValue) => {
--- a/packages/ember-glimmer/lib/syntax/curly-component.js
+++ b/packages/ember-glimmer/lib/syntax/curly-component.js
@@ -65,7 +65,7 @@ function applyAttributeBindings(attributeBindings, component, operations) {
     i--;
   }
 
-  if (!seen['style']) {
+  if (seen.indexOf('style') === -1) {
     IsVisibleBinding.apply(component, operations);
   }
 }


chancancode:
:+1: thanks! We should keep the array because the linear search on small arrays is generally much faster than object lookup in JS engines.

les2:
@chancancode will update; i figured that constant factors might have influenced the decision to use array

les2:
Okay. Added unit test (rebased and put the unit test first).

The unit test covers the specific case where there are other attribute bindings except for style. If the `IsVisibleBinding.apply(component, operations);` is commented out within the `if` statement, then the only test that fails is the test I've just added (at least, it's the only one that failed when I ran `npm test:testem` locally).

Before this "fix", the code worked because the `if` statement always evaluated to `true` and the `IsVisibleBinding` always got `apply`ed and that redundant application was apparently harmless.

I have not been able to demonstrate that invoking the `IsVisibleBinding` function *when there is also a `style` binding* caused a problem. I suppose it was just redundant.

Test Cleanliness
-----------------

I copied that `assertStyle` function from another test. Given my newbie status, I'm not sure the best place to refactor something like that.

`assertComponentElement` was acting very bizarre and I could not use it for this test for various frustrating reasons. I tried many times. Why?

- Sometimes the `style` attribute is present and `''` and sometimes it's simply not there *depending on the version of jQuery the suite runs with* (I think). That causes the attribute length assertion to fail. I thought about tinkering with that but then thought better of it.

packages/ember-glimmer/tests/integration/components/curly-components-test.js
```diff
@@ -2464,6 +2464,51 @@ moduleFor('Components test: curly components', class extends RenderingTest {
     });
   }
 
+  ['@test adds isVisible binding when style binding is missing and other bindings exist'](assert) {
+    let assertStyle = (expected) => {
```
les2: Copied from https://github.com/emberjs/ember.js/pull/14104/files#diff-c47395a5aaf92aa0a7edbfa812d4e0afR2400

`assertComponentElement` is problematic for the `style` attribute which is prob. why this function was invented by its author for the other test.

les2:
Build failure appears to be sauce related:

```
Error: 21 Aug 04:27:30 - Sauce Connect could not establish a connection.
21 Aug 04:27:30 - Please check your firewall and proxy settings.
21 Aug 04:27:30 - You can also use sc --doctor to launch Sauce Connect in diagnostic mode.
21 Aug 04:27:30 - Goodbye.
    at handleError (/home/travis/build/emberjs/ember.js/node_modules/sauce-connect-launcher/lib/sauce-connect-launcher.js:235:17)
    at /home/travis/build/emberjs/ember.js/node_modules/sauce-connect-launcher/lib/sauce-connect-launcher.js:301:9
    at /home/travis/build/emberjs/ember.js/node_modules/sauce-connect-launcher/node_modules/lodash/index.js:3073:15
    at baseForOwn (/home/travis/build/emberjs/ember.js/node_modules/sauce-connect-launcher/node_modules/lodash/index.js:2046:14)
    at /home/travis/build/emberjs/ember.js/node_modules/sauce-connect-launcher/node_modules/lodash/index.js:3043:18
    at Function.<anonymous> (/home/travis/build/emberjs/ember.js/node_modules/sauce-connect-launcher/node_modules/lodash/index.js:3346:13)
    at Socket.<anonymous> (/home/travis/build/emberjs/ember.js/node_modules/sauce-connect-launcher/lib/sauce-connect-launcher.js:299:7)
    at emitOne (events.js:90:13)
    at Socket.emit (events.js:182:7)
    at readableAddChunk (_stream_readable.js:153:18)
    at Socket.Readable.push (_stream_readable.js:111:10)
    at Pipe.onread (net.js:534:20)
Running: ./node_modules/.bin/ember sauce:disconnect
Could not start watchman; falling back to NodeWatcher for file system events.
```

https://cdn.meme.am/instances/53159640.jpg

rwjblue:
@les2 - Restarted the Sauce Labs build...

rwjblue:
Thanks @les2!

chancancode:
@les2 have you tried using https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/tests/utils/test-helpers.js#L185, like this https://github.com/emberjs/ember.js/blob/3ba7ae2b30bee93bf98023618ebe3d37b0206873/packages/ember-glimmer/tests/integration/components/curly-components-test.js#L2445-L2448?

rwjblue:
@chancancode - It will almost certainly lead to pain.  See my series of comments on https://github.com/emberjs/ember.js/pull/13881 (where the `isVisible` tests were migrated).  Basically, there is no single thing that can be used across all platforms.  We could theoretically use a platform detection, but I'm not certain that is better.

Issue #14105, opened 2016-08-21
Assertion when passing `id` hash argument to tagless components.


rwjblue:
Invoking a tagless component and providing an `id` triggers an assertion with the following text:

```
Error: Assertion Failed: You cannot use `elementId` on a tag-less component: <App.FooBarComponent:ember275>
```

I completely agree that passing an `elementId` to a tagless component is obviously wrong, however passing `id` is not the same as passing `elementId`.  However, passing `id` is not "obviously wrong".

[Alpha Demo JSBin](https://jsbin.com/motokek/edit?html,js,output)
[Canary Demo JSBin](https://jsbin.com/laqevi/1/edit?html,js,output)
[Beta Demo JSBin](https://jsbin.com/cujoki/edit?html,js,output)

PR #14106, opened 2016-08-21
[BUGFIX beta] Avoid assertion when `id=` is provided to tagless components.

rwjblue:
Fixes https://github.com/emberjs/ember.js/issues/14105.


mixonic:
👍 :shipit: 🙇

krisselden:
Can we at least say people should do this?

krisselden:
Should not I mean,  warn that it won't do anything

rwjblue:
@krisselden - It does not do anything from Ember's perspective, but `id` is a perfectly valid hash argument (though ambiguous in many cases also) to use for many scenarios where it may be unrelated to `id` in the DOM.

I guess we could add a deprecation when using `id`, and folks would just need to come up with better hash argument names...

Issue #14023, opened 2016-08-04
Error in 2.7.0 but not 2.6.0: Uncaught Error: Assertion Failed: Attempted to register a view with an id already in use: null


notmessenger:
### Updated

The bug is that since elementId is now conditionally initialized based on tagName (was this way in 1.12 and below), if the tagName is inherited as '' and you set it again post `this._super(...arguments);`, then `elementId === null` but since tagName is now set to not be `""` then when the view is registered, it will be registered as `"null"` causing problems if you render more than one.

---

**When using Ember 2.7.0**

When there are two instances of the same component used in a template the error of `Uncaught Error: Assertion Failed: Attempted to register a view with an id already in use: null` is thrown.

If the component is used in a parent route's template as well as the child's, when navigating away these deprecations are thrown:

```
DEPRECATION: A property of <regressionDemo@view:-outlet::ember413> was modified inside the didInsertElement hook. You should never change properties on components, services or models during didInsertElement because it causes significant performance degradation. [deprecation id: ember-views.dispatching-modify-property]
        at logDeprecationStackTrace (http://localhost:4200/assets/vendor.js:16676:19)
        at HANDLERS.(anonymous function) (http://localhost:4200/assets/vendor.js:16783:7)
        at raiseOnDeprecation (http://localhost:4200/assets/vendor.js:16706:12)
        at HANDLERS.(anonymous function) (http://localhost:4200/assets/vendor.js:16783:7)
        at invoke (http://localhost:4200/assets/vendor.js:16799:7)
        at deprecate (http://localhost:4200/assets/vendor.js:16767:32)
        at Object.deprecate (http://localhost:4200/assets/vendor.js:29639:37)
        at Class.exports.default._emberMetalMixin.Mixin.create._Mixin$create.scheduleRevalidate (http://localhost:4200/assets/vendor.js:54123:22)
        at http://localhost:4200/assets/vendor.js:27891:32
7ember.debug.js:6568 DEPRECATION: A property of <regressionDemo@view:-outlet::ember413> was modified inside the didInsertElement hook. You should never change properties on components, services or models during didInsertElement because it causes significant performance degradation. [deprecation id: ember-views.dispatching-modify-property]
        at logDeprecationStackTrace (http://localhost:4200/assets/vendor.js:16676:19)
        at HANDLERS.(anonymous function) (http://localhost:4200/assets/vendor.js:16783:7)
[...]

Issue #13627, opened 2016-06-09
[Beta regression] `this.elementId` is null in tagless component


cibernox:
In 2.6, `this.elementId` called from within a tagless component returned something like `ember123`.
In 2.7 the same call in 2.7.0-beta.0 returns `null`.

Intended or a regression?

rwjblue:
Well, it seems unintentional, but the new behavior does seem logical. Why would something without an element have an elementId? Why do extra work to calculate an elementId when it isn't going to be used?

Is this a large issue for your addons?

cibernox:
Not a big deal, it is an easy fix. 
I just needed a unique identifier from which derive ids for the trigger and the content so I can relate them (since the content may be rendered in the root of the page, A11y requires some hint). 
I could just as easy generate an autoincremental number myself, but since I had `this.elementId` I was using it, assuming that it was an identifier for the component, regardless of if the component had a DOM element or not.

I will stop relying on it, but I wanted to raise this change.

rwjblue:
Thank you for reporting. 

I will check with others to see if we should bring it back or try and do a deprecation. I'm unsure how big of an issue it is for others though...

rwjblue:
Thank you again for reporting the issue. If we have significant reports of this in future releases we can revisit, but for now I'm going to close...

jmacqueen:
Add me to list as missing this ability. Building in accessibility relationships in tagless components was much easier with it!

bigtone1284:
was this fixed?

alias-mac:
Just for the record (for future reference), we can use `Ember.guidFor(this)` to generate the id and use it in the template.

PR #14108, opened 2016-08-21
[BUGFIX release] Ensure `elementId` is initialized properly.

rwjblue:
Prior to this change setting `tagName: ''` then updating `tagName` in your components `init` method (but after calling `this._super(...arguments)`) left the `elementId` being used as `null`.

This updates to set `elementId` on the component after it has completed instantiation but before it is inserted into DOM.

Fixes https://github.com/emberjs/ember.js/issues/14023.

### DIFF
--- a/packages/ember-glimmer/lib/syntax/curly-component.js
+++ b/packages/ember-glimmer/lib/syntax/curly-component.js
@@ -8,6 +8,7 @@ import assign from 'ember-metal/assign';
 import get from 'ember-metal/property_get';
 import { ComponentDefinition } from 'glimmer-runtime';
 import Component from '../component';
+import { guidFor } from 'ember-metal/utils';
 
 const DEFAULT_LAYOUT = P`template:components/-default`;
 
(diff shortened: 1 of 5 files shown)


rwjblue:
@krisselden - r?

krisselden:
Looks good to me,  hopefully no one expects it in the constructor

krisselden:
Another would be to try before and after

krisselden:
Sorry for closing,  phone UI

juwara0:
Will this fix be propagated to ember-beta as well?

krisselden:
@rwjblue sounds like @chancancode suggestion we register every view again.  I think we should make it global (since we want it unique across the document anyway) and we put it into dictionary mode straight away, also use get/set or some has, get, register, unregister.  I really want to be able to benchmark A/B against different backing stores.  I think for large caches it is likely better just to use a map or Object in 'slow' dictionary mode.  If a Map object gets too many keys it will transition into a dictionary anyway.

sly7-7:
@rwjblue Any news on this ? I was relying on the fact that a tagless component has an elementId (same use case as in https://github.com/emberjs/ember.js/issues/13627)

pixelhandler:
@rwjblue maybe time for a rebase?

PR #14109, opened 2016-08-21
[DOC canary] Document how a waiter works

mixonic:
A simple explanation of how a waiter works would help anyone looking to write one.

Closes https://github.com/emberjs/ember.js/pull/13744


rwjblue:
👍

Issue #13968, opened 2016-07-30
view._renderNode.lastResult giving "Uncaught TypeError: Cannot read property 'lastResult' of null"


kellyselden:
I have unsuccessful in tracking this down so far. A few days ago when canary ticked over to beta, I noticed that canary the whole time and now beta are failing for my addon:
```
Uncaught TypeError: Cannot read property 'lastResult' of null
Renderer_revalidateTopLevelView @ ember.debug.js:14397
exports.default._emberMetalMixin.Mixin.create._Mixin$create.revalidate @ ember.debug.js:42831
invoke @ ember.debug.js:990
flush @ ember.debug.js:1054
flush @ ember.debug.js:862
end @ ember.debug.js:176
(anonymous function) @ ember.debug.js:742
```

Here is one of the travis build logs https://travis-ci.org/kellyselden/ember-tree/builds/148237768. I tried removing dependencies to narrow it down, but then tests start failing for other reasons.

When debugging, `if (view._renderNode.lastResult) {` passes through the first time, then fails on the second. I inspected `view` and although it doesn't have a `_renderNode` property, it does have a `lastResult` property. I don't know if that helps.

krisselden:
_renderNode is private and not part of glimmer2

kellyselden:
OK, but I don't know how I'm causing this failure. I'm not using any private code in my addon code, maybe one of the addons I'm using is having issues with glimmer 2. My ember debugging skills are apparently lacking in this case.

kellyselden:
This is still happening. Would you say it's probably an addon I'm using that's using an old htmlbars version? Any debugging tips are appreciated! cc @krisselden

rwjblue:
It should be fixed by https://github.com/emberjs/ember.js/pull/14110.

kellyselden:
@rwjblue Using that branch, it didn't seem to fix it: https://travis-ci.org/kellyselden/ember-tree/jobs/154205905

rwjblue:
@kellyselden - Sorry, I forgot to update the bower tree after my last commit. Can you blow cache and test again?

kellyselden:
@rwjblue Works now!

Issue #13846, opened 2016-07-19
scheduleRevalidate in canary is causing test failures


backspace:
Hello! `travis-web` has recently started failing to build on canary. One of the [errors](https://travis-ci.org/travis-ci/travis-web/jobs/145931925#L595) is this:

`Promise rejected after visiting /builds/branches-tab: null is not an object (evaluating 'node.ownerNode.emberView.scheduleRevalidate')`

Which looks to be from [this line](https://github.com/emberjs/ember.js/blob/28cedad1418bedd2dad44eb56668fa67b9769d86/packages/ember-htmlbars/lib/utils/subscribe.js#L24) in `ember-htmlbars/lib/utils/subscribe.js`:

`node.ownerNode.emberView.scheduleRevalidate(node, labelFor(stream));`

Looking at the history of that file, I don’t see any reason why recent changes would have produced this failure, but I’m wholly unfamiliar with this code.

Hopefully this is a helpful error report and not just a red herring. Let me know if you need any more information or if maybe it’s a problem in `travis-web`.

Serabe:
Thank you for reporting this. I can confirm that this is failing and can be found in emberx-select (though canary is allowed to fail and therefore it was not noticed before). In that line:

* `node` is a `PropertyAttrMorph`. To be exact, a `selected` from an `x-option`.
* `node.ownerNode` is an `EmberMorph` whose `contextualElement` is `div#ember-testing`.

backspace:
Thanks for the confirmation, @Serabe. I forked the repository for the built version of Ember and made a [commit](https://github.com/backspace/ember/commit/952149936ef942f225f70fab9c3d509a5d134422) that adds guards against the absence of `node.ownerNode.emberView`. I [changed `travis-web`’s `ember-try` configuration](https://github.com/travis-ci/travis-web/commit/b4fc37228c52b7958eace3317c5c933e330b2651) to use that fork as the Bower dependency and [the tests now pass](https://travis-ci.org/travis-ci/travis-web/jobs/146166038) with this modified version of canary.

I don’t know enough about this to know whether that type of guard is truly desirable or perhaps a deeper fix where `node.ownerNode.emberView` should _always_ be present is more the issue.

rwjblue:
Can someone try to bisect and narrow down which change caused the issue?

backspace:
[...]

PR #14110, opened 2016-08-22
[BUGFIX beta] Fix issues with revalidation during teardown.

rwjblue:
When the toplevelView (aka `ownerView`) is being destroyed (during test teardown generally) an error was ocurring under certain "special" circumstances. Specifically, if removal of a sibling node triggered a revalidation of one still attached.

Consider the following example:

``` hbs
{{#x-select as |select|}}
 {{#select.option value="1"}}{/select.option}}
 {{#select.option value="2"}}{/select.option}}
{{/x-select}}
```

The components in question are using the registration pattern so that the children notify/register with the parent upon `didInsertElement` (and `willDestroyElement`).

When a new option is added or removed from the parent, it updates its `value` property which is bound to each options `selected` attribute (to toggle the selection state in the DOM).

The specific mechanism that causes the DOM to get updated when the various computed properties change is that a Stream object calls `ownerNode.emberView.scheduleRevalidate`.  This tells the rest of the system to start walking the tree to flush out any updates to the DOM.

When a view is being removed, we set the `emberView` property of its `renderNode` to `null` before traversing the subtree of children. This means, that as the children are removed (and therefore cause the `value` of the parent to change) the `selected` attribute binding attempts to call `ownerNode.emberView.scheduleRevalidation()`.

Unfortunately, when we are actually tearing down the `ownerNode.emberView` itself that invocation results in an error ( cannot call `scheduleRevalidation` on `null`).

---

This change adds a simple guard to avoid calling `scheduleRevalidation` when the `ownerNode.emberView` is being torn down.

Fixes https://github.com/emberjs/ember.js/issues/13846.
Fixes https://github.com/emberjs/ember.js/issues/13968.

### DIFF
--- a/packages/ember-glimmer/tests/integration/components/curly-components-test.js
+++ b/packages/ember-glimmer/tests/integration/components/curly-components-test.js
@@ -2570,4 +2570,74 @@ moduleFor('Components test: curly components', class extends RenderingTest {
 
     assertElement('foo');
   }
+
+  ['@test child triggers revalidate during parent destruction (GH#13846)']() {
+    let select;
+
+    this.registerComponent('x-select', {
+      ComponentClass: Component.extend({
+        tagName: 'select',
+
+        init() {
+          this._super();
+          this.options = emberA([]);
+          this.value = null;
+
+          select = this;
+        },
+
+        updateValue() {
+          var newValue = this.get('options.lastObject.value');
+
+          this.set('value', newValue);
+        },
+
+        registerOption(option) {
+          this.get('options').addObject(option);
+        },
+
+        unregisterOption(option) {
+          this.get('options').removeObject(option);
+
+          this.updateValue();
+        }
+      }),
+
+      template: '{{yield this}}'
+    });
+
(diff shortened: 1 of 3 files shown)


rwjblue:
I have uploaded the assets to make it a bit easier to test this in an ember-cli app.  All you should need to do is change your `bower.json` from:

```json
    "ember": "~2.8.0-beta.1",
```

To:

```json
    "ember": "rwjblue/ember#pr-14110",
```

rwjblue:
Since @krisselden worked on https://github.com/emberjs/ember.js/pull/13775, I'd love his review here...

cibernox:
I confirm that this branch fixes the test suite for Ember Power Select: https://travis-ci.org/cibernox/ember-power-select/jobs/154144524#L332

backspace:
Thanks so much for this fix!

cibernox:
@rwjblue Could you port this commit to the beta branch so ember-try pulls the fix and addons got pretty green travis badges again 😄

PR #14112, opened 2016-08-22
[BUGFix beta] Fix basic component and helper resolution in engines

dgeb:
A rebase and extension of #14055, this contains the following fixes:
- Clone additional parent dependencies into engine instances. Specifically `service:-glimmer-environment` (only for glimmer) and `renderer:-dom` or `renderer:-inert`
- Ensure that Cache lookups are performed using the correct owner.
- Ensure that engines are destroyed _before_ top-level views so that top-level
  views are not re-instantiated during engine teardown.

### DIFF
--- a/packages/ember-application/lib/system/engine-instance.js
+++ b/packages/ember-application/lib/system/engine-instance.js
@@ -170,18 +170,29 @@ if (isEnabled('ember-application-engines')) {
     cloneParentDependencies() {
       let parent = getEngineParent(this);
 
-      [
+      let registrations = [
         'route:basic',
         'event_dispatcher:main',
         'service:-routing'
-      ].forEach(key => this.register(key, parent.resolveRegistration(key)));
+      ];
 
-      [
+      if (isEnabled('ember-glimmer')) {
+        registrations.push('service:-glimmer-environment');
+      }
+
+      registrations.forEach(key => this.register(key, parent.resolveRegistration(key)));
+
+      let env = parent.lookup('-environment:main');
+      this.register('-environment:main', env, { instantiate: false });
+
+      let singletons = [
         'router:main',
         P`-bucket-cache:main`,
         '-view-registry:main',
-        '-environment:main'
-      ].forEach(key => this.register(key, parent.lookup(key), { instantiate: false }));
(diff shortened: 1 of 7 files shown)


packages/ember-glimmer/lib/environment.js
```diff
@@ -268,8 +268,9 @@ export default class Environment extends GlimmerEnvironment {
 
   getComponentDefinition(path, parentMeta) {
     let name = path[0];
+    let owner = !!(parentMeta && parentMeta.owner) ? parentMeta.owner : this.owner;
```
rwjblue: I believe that this may be a bug in glimmer.  The `parentMeta` should always be present.  

I believe that [these lines](https://github.com/tildeio/glimmer/blob/c0a92795b2df68bfed674683c2ae9c3d2b633835/packages/glimmer-runtime/lib/environment.ts#L235-L239) should be changed to include `blockMeta`.

I'm happy to iterate on this in a follow-up though...

packages/ember-application/lib/system/engine-instance.js
```diff
@@ -170,18 +170,32 @@ if (isEnabled('ember-application-engines')) {
...
         'router:main',
         P`-bucket-cache:main`,
-        '-view-registry:main',
-        '-environment:main'
-      ].forEach(key => this.register(key, parent.lookup(key), { instantiate: false }));
+        '-view-registry:main'
+      ];
+
+      if (isEnabled('ember-glimmer')) {
+        singletons.push(env.isInteractive ? 'renderer:-dom' : 'renderer:-inert');
```
rwjblue: I think this should be done in both sides of the feature flag.

packages/ember-glimmer/lib/environment.js
asakusuma: Went to look into this more since I wrote this code, then noticed that `hasComponentDefinition` returns false in ember. Why is this?

https://github.com/emberjs/ember.js/blob/5f5b9ae5f8c117c3dc45409b7f0d2272f042cb68/packages/ember-glimmer/lib/environment.js#L266

asakusuma: PR: https://github.com/tildeio/glimmer/pull/290

PR #14115, opened 2016-08-23
[GLIMMER2] Implement `getViewBounds` for components

chancancode:
`getViewBounds` is the primitive that is used to implement `getViewRange`, `getViewClientRects`, etc

### DIFF
--- a/packages/ember-glimmer/lib/syntax/outlet.js
+++ b/packages/ember-glimmer/lib/syntax/outlet.js
@@ -116,6 +116,7 @@ class AbstractOutletComponentManager {
   }
 
   didCreateElement() {}
+  didRenderLayout() {}
   didCreate(state) {}
   update(state, args, dynamicScope) {}
   didUpdate(state) {}
--- a/packages/ember-glimmer/tests/integration/components/utils-test.js
+++ b/packages/ember-glimmer/tests/integration/components/utils-test.js
@@ -1,6 +1,7 @@
 import { moduleFor, RenderingTest } from '../../utils/test-case';
 import { Component } from '../../utils/helpers';
 import {
+  getViewBounds,
   getViewClientRects,
   getViewBoundingClientRect
 } from 'ember-views/system/utils';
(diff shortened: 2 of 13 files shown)


chancancode:
cc @teddyzeenny @ef4

packages/ember-views/lib/system/utils.js
```diff
@@ -18,15 +22,53 @@ export const STYLE_WARNING = '' +
...
+
+/**
+  @private
+  @method getViewBounds
+  @param {Ember.View} view
+*/
+export function getViewBounds(view) {
+  if (view[BOUNDS]) {
+    return view[BOUNDS];
+  } else if (view._renderNode) {
```
krisselden: if you delegated to the _renderer, this logic wouldn't need to branch, htmlbars could do one thing/glimmer the other.

packages/ember-glimmer/lib/syntax/outlet.js
```diff
@@ -116,6 +116,7 @@ class AbstractOutletComponentManager {
   }
 
   didCreateElement() {}
+  didRenderLayout() {}
```
chancancode: Note: I am dropping the bounds on `{{outlet}}` and `{{render}}` currently because there are no components to attach them to, but we'll likely have to put it somewhere (e.g. the controller) for the inspector to work

packages/ember-views/lib/system/utils.js
chancancode: Sounds good

packages/ember-glimmer/tests/integration/components/utils-test.js
```diff
@@ -49,7 +91,7 @@ moduleFor('ember-views/system/utils', class extends RenderingTest {
     assert.ok(getViewClientRects(component) instanceof ClientRectListCtor);
   }
 
-  ['@htmlbars getViewBoudningClientRect'](assert) {
+  ['@test getViewBoudningClientRect'](assert) {
```
rwjblue: s/getViewBoudningClientRect/getViewBoundingClientRect/ 😝

Issue #14114, opened 2016-08-22
arrayContentDidChange/arrayContentWillChange can no longer be overriden


workmanw:
Carry over from this PR: https://github.com/emberjs/ember.js/pull/13534

In that PR, `arrayContentDidChange` and `arrayContentWillChange` were modified to make them implemented and invoked as pure functions (as opposed to object-oriented methods). That PR is problematic for [ember-data-model-fragments](https://github.com/lytics/ember-data-model-fragments). The root of the problem is that because `replace` (and others) no longer call `this.arrayContentDidChange(...)`, instead calling `arrayContentDidChange(this, ...)`, we are unable to override this function in our subclasses array. See here: [array/stateful.js#L178-L190](https://github.com/lytics/ember-data-model-fragments/blob/master/addon/array/stateful.js#L178-L190).

There was some discussion about about whether or not this was a breaking change because the API is not changing, just the behavior of a function. 

If possible, we should try to sort this out before Ember 2.8 ships, otherwise we'll have an upgrade blocker to resolve in ember-data-model-fragments.

Serabe:
Thank you for taking the time to open this!

mmun:
In general, one should assume that framework provided methods are not overridable unless explicitly documented so. In other words, borrowing from Java, they are `final`. We should improve our documentation in this regard.

Nevertheless, as 2.8 is an LTS we should fix this. It may be sufficient to restore the old behaviour only for classes that implement the Array mixin, like ArrayProxy in this case.

workmanw:
> In general, one should assume that framework provided methods are not overridable unless explicitly documented so. In other words, borrowing from Java, they are final. We should improve our documentation in this regard.

Yea I can certainly understand this. I'm definitely not asserting that the right solution is to restore old functionality (though admittedly it would make my life easier). More than anything else I wanted to bring this up and kick off a discussion.

------------

I think the reason I brought this up to begin with was because it impacts a data structure / interfact that is often extended. 

[...]

PR #14117, opened 2016-08-23
[BUGFIX beta] Call ArrayProxy's content change hooks

mmun:
Provides some temporary relief for #14114.

### DIFF
--- a/packages/ember-runtime/lib/system/array_proxy.js
+++ b/packages/ember-runtime/lib/system/array_proxy.js
@@ -20,8 +20,6 @@ import alias from 'ember-metal/alias';
 import {
   addArrayObserver,
   removeArrayObserver,
-  arrayContentDidChange,
-  arrayContentWillChange,
   objectAt
 } from 'ember-runtime/mixins/array';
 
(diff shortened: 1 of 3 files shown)


rwjblue:
@workmanw / @mmun - Can you confirm this fixes the addon?

workmanw:
You need to also update the `replace` implementation on `Ember.NativeArray` to use the member methods. See: [system/native_array.js#L55-L74](https://github.com/mmun/ember.js/blob/0c2a3f0e2a42ac151236221dce3b673d7caa7120/packages/ember-runtime/lib/system/native_array.js#L55-L74)

If you ping me after that, I'd be happy to verify the fix using the addon. Thank you.

workmanw:
@mmun @rwjblue Actually I retract my previous statements. This actually does fix the issue because the `array-mixins` implementation invokes `sendEvent('@array:change')` and `Ember.ArrayProxy` ultimately circles back around and calls the member function via it's array observer. So :+1:.

Thank you guys. Post Ember 2.8 LTS I'm happy to work to patch ember-data-model-fragments so it doesn't depend on this behavior. It seems like setting up another array observer in their arrays could be a solution.

PR #14121, opened 2016-08-23
Provide better error for non-resolving quoteless action

HeroicEric:
This makes a slight improvement to the assertion that is raised when a
quoteless path is used to specify an action by including additional
information to help track down which `{{action}}` invocation is
triggering the assertion.

### DIFF
--- a/packages/ember-glimmer/lib/modifiers/action.js
+++ b/packages/ember-glimmer/lib/modifiers/action.js
@@ -176,12 +176,14 @@ export default class ActionModifierManager {
       if (actionNameRef[INVOKE]) {
         actionName = actionNameRef;
       } else {
+        let actionLabel = actionNameRef._propertyKey;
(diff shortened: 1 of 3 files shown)


packages/ember-glimmer/lib/modifiers/action.js
```diff
@@ -176,12 +176,14 @@ export default class ActionModifierManager {
       if (actionNameRef[INVOKE]) {
         actionName = actionNameRef;
       } else {
+        let actionLabel = actionNameRef._propertyKey;
```
HeroicEric: I'm hoping somebody might point out a better way to do this 😄 

There is another area that is doing something similar https://github.com/emberjs/ember.js/pull/13684#discussion-diff-67275895

rwjblue:
Thanks @HeroicEric!

PR #14123, opened 2016-08-24
[BUGFIX beta] Avoid rerendering outlet state during router destruction.

rwjblue:
During each `Ember.Route`'s `willDestroy` we trigger a `run.once(this.router, '_setOutlets');`. This is so that the routes views are destroyed properly (by removing them from the outlet state).

During `Ember.Router`'s `willDestroy` we clear `this._toplevelView` (along with a bunch of other cleanup).

These two things combined can mean that `this._toplevelView` is `null` when `_setOutlets` is called again (during the `Router`'s destruction). In that scenario we are actually creating another `this._toplevelView` and setting up another rendered root (since one doesn't exist). When this happens the newly created root is never cleaned up, since the Router's `willDestroy` has already ran and can no longer clean up after itself.

---

This fixes a number of test failures that crop up when doing fuzz testing (via `seed` URL param with QUnit).

### DIFF
--- a/packages/ember-routing/lib/system/router.js
+++ b/packages/ember-routing/lib/system/router.js
@@ -140,6 +140,10 @@ const EmberRouter = EmberObject.extend(Evented, {
       this._engineInstances = new EmptyObject();
       this._engineInfoByRoute = new EmptyObject();
     }
+
+    // avoid shaping issues with checks during `_setOutlets`
+    this.isDestroyed = false;
+    this.isDestroying = false;
   },
 
   /*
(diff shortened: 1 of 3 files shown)


stefanpenner:
@rwjblue this may be tricky, but can we test this?

rwjblue:
@stefanpenner - I don't know.  This was the result of a gnarly debugging session when running Ember's own tests in randomized order (where there were a large number of failures since leaking a root view in Glimmer is worse). 

I am happy to try to add a specific test, but I think the only way to do it would be to force the router to be destroyed before a route instance.

rwjblue:
@stefanpenner - Added two tests (one with an engine and one without).

stefanpenner:
tests look legit

PR #14125, opened 2016-08-24
[WIP - DO NOT MERGE] Initial work on ember-glimmer node tests

scalvert:
**Description**

Initial work on ember-glimmer integration. WIP on ensuring tests pass within Node environment.

@chadhietala

### DIFF
--- a/bin/run-node-tests.js
+++ b/bin/run-node-tests.js
@@ -9,6 +9,18 @@ qe.runInContext(global);
 var glob = require('glob');
 var root = 'tests/';
 
+var OVERRIDE_FEATURES = process.env.OVERRIDE_FEATURES;
+if (OVERRIDE_FEATURES) {
+  var features = OVERRIDE_FEATURES.split(',');
+  var EmberENV = { FEATURES: { } };
+  for (var i = 0; i < features.length; i++) {
+    EmberENV.FEATURES[features[i]] = true;
+  }
+
+  console.log('setting EmberENV: ', EmberENV);
+  global.EmberENV = EmberENV;
+}
+
 function addFiles(files) {
   glob.sync(root + files)
     .map(function(name) {
--- a/features.json
+++ b/features.json
@@ -4,7 +4,7 @@
     "ember-routing-route-configured-query-params": null,
     "ember-libraries-isregistered": null,
     "ember-application-engines": true,
-    "ember-glimmer": null,
+    "ember-glimmer": true,
     "ember-runtime-computed-uniq-by": true,
     "ember-improved-instrumentation": null,
     "ember-runtime-enumerable-includes": true,
--- a/packages/ember-glimmer/lib/environment.js
+++ b/packages/ember-glimmer/lib/environment.js
@@ -45,6 +45,8 @@ import { default as normalizeClassHelper } from './helpers/-normalize-class';
 import { default as htmlSafeHelper } from './helpers/-html-safe';
 import { OWNER } from 'container';
 
+import installPlatformSpecificProtocolForURL from './protocol-for-url';
+
 const builtInComponents = {
   textarea: '-text-area'
 };
@@ -133,7 +135,7 @@ export default class Environment extends GlimmerEnvironment {
     super(...arguments);
     this.owner = owner;
 
-    this.uselessAnchor = this.getAppendOperations().createElement('a');
+    installPlatformSpecificProtocolForURL(this);
 
     this._definitionCache = new Cache(2000, ({ name, source, owner }) => {
       let { component: ComponentClass, layout } = lookupComponent(owner, name, { source });
@@ -161,11 +163,6 @@ export default class Environment extends GlimmerEnvironment {
     this.builtInModifiers = {
       action: new ActionModifierManager()
     };
-  }
-
-  protocolForURL(url) {
-    this.uselessAnchor.href = url;
-    return this.uselessAnchor.protocol;
   }
 
   // Hello future traveler, welcome to the world of syntax refinement.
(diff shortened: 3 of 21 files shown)


features.json
```diff
@@ -4,7 +4,7 @@
     "ember-routing-route-configured-query-params": null,
     "ember-libraries-isregistered": null,
     "ember-application-engines": true,
-    "ember-glimmer": null,
+    "ember-glimmer": true,
```
rwjblue: Note: this will need to be changed back before landing

packages/ember-glimmer/lib/setup-registry.js
```diff
@@ -14,15 +14,24 @@ export function setupApplicationRegistry(registry) {
...
+
 
   registry.register('service:-dom-changes', {
     create({ document }) { return new DOMChanges(document); }
   });
 
-  registry.register('service:-dom-tree-construction', {
-    create({ document }) { return new DOMTreeConstruction(document); }
-  });
+  if (typeof document !== 'undefined') {
```
rwjblue: Should do:

```js
import environment from 'ember-environment';

// ...snip...

if (environment.hasDOM) {

}
```

packages/ember-glimmer/lib/setup-registry.js
```diff
@@ -14,15 +14,24 @@ export function setupApplicationRegistry(registry) {
...
+    registry.register('service:-dom-tree-construction', {
+      create({ document }) { return new DOMTreeConstruction(document); }
+    });
+
+  } else {
+    registry.register('service:-dom-tree-construction', {
+      create({ document: simpleDocument }) { return new NodeDOMTreeConstruction(simpleDocument); }
```
rwjblue: This param is the injection setup ([from here](https://github.com/emberjs/ember.js/blob/master/packages/ember-application/lib/system/application-instance.js#L289)), and shouldn't need to be overridden like this.

I suspect something like this would work (once the `service:-document` injection is fixed):

```js
registry.register('service:-dom-tree-construction', {
  create({ document }) {
    let Implementation = environment.hasDOM ? DOMTreeConstruction : NodeDOMTreeConstruction;
    return new Implementation(document);
  }
});
```

packages/ember-templates/lib/compat.js
```diff
@@ -19,6 +20,7 @@ Object.defineProperty(EmberHandlebars, 'SafeString', {
 });
 
 EmberHTMLBars.template = EmberHandlebars.template = template;
+EmberHTMLBars.DOMHelper = DOMHelper;
```
rwjblue: I vaguely believe this is underscored?

packages/ember-glimmer/lib/environment.js
```diff
@@ -132,7 +132,7 @@ export default class Environment extends GlimmerEnvironment {
     super(...arguments);
     this.owner = owner;
 
-    this.uselessAnchor = document.createElement('a');
+    this.uselessAnchor = this.getAppendOperations().createElement('a');
```
chancancode: This is only used for `protocolForURL`, which needs to be overridden in Node anyway

features.json
scalvert: For sure. We're just hard coding it for now, but will need a way to switch this for testing. 

Would we want to run the tests 2 times, once for HTMLBars and once for Glimmer?

packages/ember-glimmer/lib/environment.js
scalvert: So what do you suggest?

packages/ember-glimmer/lib/setup-registry.js
scalvert: 👍

packages/ember-glimmer/lib/setup-registry.js
scalvert: OK, that's much cleaner. I'll fix this and the -document injection.

packages/ember-glimmer/lib/setup-registry.js
scalvert: Importing from 'ember-environment' resulted in a null environment. Is that the correct path?

chancancode: It's not the default export for some reason: https://github.com/emberjs/ember.js/blob/master/packages/ember-environment/lib/index.js#L106

So you'll need `import { environment } ...`

packages/ember-glimmer/lib/environment.js
chancancode: we can probably initialize it lazily in `protocolForURL`

packages/ember-glimmer/lib/setup-registry.js
scalvert: Ah, perfect. I'll try that. Thanks!

bin/run-node-tests.js
```diff
@@ -17,6 +17,6 @@ function addFiles(files) {
     .forEach(require);
 }
 
-addFiles('/**/*-test.js');
+addFiles('/**/app-boot-test.js');
```
scalvert: Just looking at fixing these for now. Will enable the rest as I move forward.

packages/ember-glimmer/lib/environment.js
scalvert: Is see. So we can pull it out of here and move it closer to usage. Makes sense.

packages/ember-templates/lib/compat.js
scalvert: I'll check. It didn't look as though it was.

packages/ember-glimmer/lib/syntax/curly-component.js
```diff
@@ -290,9 +290,11 @@ class CurlyComponentManager {
   }
 
   didCreate({ component }) {
-    component.trigger('didInsertElement');
-    component.trigger('didRender');
-    component._transitionTo('inDOM');
+    if (false) {
```
rwjblue: Hehe, this was just for testing on the node side.  Will remove before landing...

PR #14135, opened 2016-08-26
[BUGFIX beta] Fix basic component and helper resolution in engines.

rwjblue:
Rebase and update of https://github.com/emberjs/ember.js/pull/14112 (to use `symbolTable` throughout) which is a rebase of https://github.com/emberjs/ember.js/pull/14055.

Contains the following fixes:
- Clone additional parent dependencies into engine instances. Specifically service:-glimmer-environment (only for glimmer) and renderer:-dom or renderer:-inert
- Ensure that Cache lookups are performed using the correct owner.
- Ensure that engines are destroyed before top-level views so that top-level
  views are not re-instantiated during engine teardown.

### DIFF
--- a/packages/ember-application/lib/system/application-instance.js
+++ b/packages/ember-application/lib/system/application-instance.js
@@ -484,6 +484,7 @@ BootOptions.prototype.toEnvironment = function() {
   let env = assign({}, environment);
   // For compatibility with existing code
   env.hasDOM = this.isBrowser;
+  env.isInteractive = this.isInteractive;
   env.options = this;
   return env;
 };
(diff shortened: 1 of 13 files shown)


asakusuma:
Looks like [`Applications with autoboot set to false do not autoboot`](https://github.com/emberjs/ember.js/blob/eecc9f99b58c02c2dc076a9abd816e53671933a0/packages/ember-application/tests/system/visit_test.js#L80) is failing but only in Firefox and with feature flags off.

Serabe:
@asakusuma it seems to be failing only sometimes for some merges :$ 
image

rwjblue:
Yes, we have an intermittent issue with FF_Current.

rwjblue:
I updated the owner specific stuff to ensure that we do not use the environments owner (which in the case of engines would be the consuming apps not the engines).  Still need to add a couple tests around this (where the engine and the app both have a given component/helper but the "right" one is used in each context).

PR #14137, opened 2016-08-26
[DOC release] Make Map and MapWithDefault public

### LINKED ISSUE #13815: Map and MapWithDefault marked as private
The `Map` and `MapWithDefault` classes are currently marked as `@private`, yet they are exported through the [shims](https://github.com/ember-cli/ember-cli-shims/blob/master/app-shims.js#L94-L97). It looks like the `@private` tag was added in https://github.com/emberjs/ember.js/pull/11362.

It seems inconsistent to me to export the classes but mark them as private so I'd like to propose marking them as public.

/cc @pangratz

pixelhandler:
- Change `Map` and `MapWithDefault` to `public`
- See [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)

Fixes #13815

### DIFF
--- a/packages/ember-metal/lib/map.js
+++ b/packages/ember-metal/lib/map.js
@@ -235,7 +235,7 @@ OrderedSet.prototype = {
 
   @class Map
   @namespace Ember
-  @private
+  @public
   @constructor
 */
 function Map() {
@@ -252,7 +252,7 @@ function Map() {
 /**
   @method create
   @static
-  @private
+  @public
 */
 Map.create = function() {
   let Constructor = this;
@@ -269,7 +269,7 @@ Map.prototype = {
     @property size
     @type number
     @default 0
-    @private
+    @public
   */
   size: 0,
 
@@ -279,7 +279,7 @@ Map.prototype = {
     @method get
     @param {*} key
     @return {*} the value associated with the key, or `undefined`
-    @private
+    @public
   */
   get(key) {
     if (this.size === 0) { return; }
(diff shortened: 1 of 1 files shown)


krisselden:
I would like to remove these instead if they've been private since before < 2.0, we can move them to an add-on and ask people to install them.  Ember Data can stop using them or include them as a library.

pixelhandler:
@stefanpenner cc/ @krisselden I rebased off master. Per discussion in #13815 I think we want to make public vs. addon-ize.

stefanpenner:
> I would like to remove these instead if they've been private since before < 2.0

I think our path to removing these, is as part of a transition to ES6 Map. These are a strict subset of that map, so someday when we can cut over we should be able to just delete them.

pixelhandler:
@stefanpenner so making public is part of that path?

> I think our path to removing these, is as part of a transition to ES6 Map. These are a strict subset of that map, so someday when we can cut over we should be able to just delete them.

pixelhandler:
@stefanpenner @krisselden just curious any update on if this is a go or no-go ?

tomdale:
We discussed this (finally!) at the core team meeting today. @rwjblue is going to write down the resolution.

rwjblue:
We should extract two addons:

*  a map-polyfill (that does nothing if the supported platforms already have native map)
* an add on that provides ember-map-with-default that extends from native map if present or the polyfill if not 

Then ember-data should be updated to use the map-with-default add on directly. Once these changes are made, we can deprecate these two in Ember itself.

PR #14138, opened 2016-08-27
Refactor Descriptor

chancancode:
- Duck-type on `isDescriptor === true` instead of using `instanceof`
- Require all descriptors to have `setup` and `teardown`
- Refactored most descriptors to use ES6 classes except CP descriptors because it’s a big file and I ran out of time for it

cc @rwjblue @krisselden @stefanpenner

### DIFF
--- a/packages/ember-metal/lib/properties.js
+++ b/packages/ember-metal/lib/properties.js
@@ -17,8 +17,29 @@ import { overrideChains } from './property_events';
   @class Descriptor
   @private
 */
-export function Descriptor() {
-  this.isDescriptor = true;
+export class Descriptor {
+  constructor() {
+    this.isDescriptor = true;
+  }
+
+  setup(obj, key, isWatching) {
+    if (isEnabled('mandatory-setter')) {
+      if (isWatching) {
+        Object.defineProperty(obj, key, {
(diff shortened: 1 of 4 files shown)


packages/ember-metal/lib/properties.js
```diff
@@ -139,58 +160,41 @@ export function defineProperty(obj, keyName, desc, data, meta) {
   watching = watchEntry !== undefined && watchEntry > 0;
 
   if (existingDesc) {
-    existingDesc.teardown(obj, keyName);
+    existingDesc.teardown(obj, keyName, watching, meta);
   }
 
-  if (desc instanceof Descriptor) {
-    value = desc;
+  if (desc === null || desc === undefined) {
```
chancancode: This branch was just moved up, mostly unchanged. The condition was previously `if (desc == null)` – note the _double_ equal.

stefanpenner:
does ember correctly strip out the `classCheck` helper in babeliefed ES6 Modules?

chadhietala:
@stefanpenner it should.

pixelhandler:
@chancancode just curious do any tests need to be added / changed for this refactor?

PR #14139, opened 2016-08-27
😪 Make `view.childViews` lazy

chancancode:
:new: `getRootViews` returns an array of “top-level” components
:new: `getChildViews` to replace `view.childViews` (to be deprecated :soon:)
:new: “tagless” components are registered in the view-registry with a GUID
:cool: `view.childViews` is now a getter that delegates to `getChildViews`
:free: `view.childViews` no longer leak destroyed components (bug in beta)
:back: `view.childViews` now tracks updates (components added or removed during updates) correctly, which was unreliable since 1.13

Instead of tracking `childView` instances, we now track a list of IDs that can be used to look them up in the view registry. This has the benefit of not holding a reference to the view instance – thus do not require cleanup when the child views are destroyed.

When requested, we lazily query the view registry to reify the (still) live children and cleanup and stale IDs while we are at it.

This approach does leak the string IDs of destroyed components, however the effect is believed to be minimal. If this ended up causing problems, we can be smarter about this, such as tracking the number of removals and schedule a collection (outside of the render loop) when a view has accumulated enough removals.

### DIFF
--- a/packages/ember-glimmer/lib/syntax/curly-component.js
+++ b/packages/ember-glimmer/lib/syntax/curly-component.js
@@ -200,7 +200,9 @@ class CurlyComponentManager {
     dynamicScope.view = component;
     dynamicScope.targetObject = component;
 
-    parentView.appendChild(component);
+    if (parentView !== null) {
+      parentView.appendChild(component);
+    }
 
     component.trigger('didInitAttrs', { attrs });
     component.trigger('didReceiveAttrs', { newAttrs: attrs });
--- a/packages/ember-glimmer/tests/utils/helpers.js
+++ b/packages/ember-glimmer/tests/utils/helpers.js
@@ -18,6 +18,7 @@ export { DOMChanges } from 'glimmer-runtime';
 export { InteractiveRenderer, InertRenderer } from 'ember-glimmer/renderer';
 export { default as makeBoundHelper } from 'ember-glimmer/make-bound-helper';
 export { htmlSafe, SafeString } from 'ember-glimmer/utils/string';
+import dictionary from 'ember-metal/dictionary';
 
 export function buildOwner(options) {
   let owner = _buildOwner(options);
@@ -37,7 +38,7 @@ export function buildOwner(options) {
   owner.inject('component', 'renderer', 'renderer:-dom');
   owner.inject('template', 'env', 'service:-glimmer-environment');
 
-  owner.register('-view-registry:main', { create() { return {}; } });
+  owner.register('-view-registry:main', { create() { return dictionary(null); } });
   owner.inject('renderer', '_viewRegistry', '-view-registry:main');
 
   owner.register('template:-root', RootTemplate);
--- a/packages/ember-metal/tests/descriptor_test.js
+++ b/packages/ember-metal/tests/descriptor_test.js
@@ -3,6 +3,36 @@ import { Mixin } from 'ember-metal/mixin';
 import { defineProperty } from 'ember-metal/properties';
 import descriptor from 'ember-metal/descriptor';
 
+// IE9 soft-fails when trying to delete a non-configurable property
+const hasCompliantDelete = (function() {
+  let obj = {};
+
+  Object.defineProperty(obj, 'zomg', { configurable: false, value: 'zomg' });
+
+  try {
+    delete obj.zomg;
+  } catch(e) {
+    return true;
+  }
+
+  return false;
+})();
+
+// IE9 soft-fails when trying to assign to a non-writable property
+const hasCompliantAssign = (function() {
+  let obj = {};
+
+  Object.defineProperty(obj, 'zomg', { writable: false, value: 'zomg' });
+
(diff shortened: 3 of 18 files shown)


chancancode:
cc @teddyzeenny

packages/ember-metal/tests/descriptor_test.js
```diff
@@ -0,0 +1,352 @@
...
+      this.klass = EmberObject.extend(this.props);
+      return this.klass.create();
+    }
+
+    source() {
+      return this.klass.prototype;
+    }
+  },
+
+class extends DescriptorTest {
```
praxxis: indentation seems off here

packages/ember-glimmer/lib/syntax/curly-component.js
```diff
@@ -188,7 +188,6 @@ class CurlyComponentManager {
     aliasIdToElementId(args, props);
 
     props.parentView = parentView;
-    props.renderer = parentView.renderer;
```
chancancode: The last failure is because I removed this code. It seems like engine components are not getting these injections setup: https://github.com/emberjs/ember.js/blob/master/packages/ember-application/lib/system/application-instance.js#L291-L297 (maybe the rest of things in that block too?)

I can add this back, but it seems better to just fix the injection. @rwjblue @dgeb can you halp?

packages/ember-glimmer/tests/utils/helpers.js
```diff
@@ -37,7 +38,7 @@ export function buildOwner(options) {
   owner.inject('component', 'renderer', 'renderer:-dom');
   owner.inject('template', 'env', 'service:-glimmer-environment');
 
-  owner.register('-view-registry:main', { create() { return {}; } });
+  owner.register('-view-registry:main', { create() { return dictionary(null); } });
```
chancancode: cc @krisselden

packages/ember-glimmer/lib/syntax/curly-component.js
rwjblue: Ya, I can fix. Might be easier to put it back for now and I'll fix with a follow up PR later this afternoon...

chancancode: :+1:

rwjblue: Tackled this in https://github.com/emberjs/ember.js/pull/14140 (which has caused you yet another rebase conflict).  Sorry...

chancancode:
Fixed the IE9 errors – they are due to IE9 not supporting strict mode

chancancode:
Confirmed no regression: https://cdn.rawgit.com/chancancode/15c2b63ea158bc704b88758bd401a7be/raw/f387a5ef50aa9eed5e0a34bf39e2b9c9fddb9eb3/report.html

chancancode:
To be clear – the "confirmed no regression" means normal initial render performance, as opposed to `.childView` performance. It _is_ expected that `.childView` _will_ be slower, but since it's used for debugging we believe that is an acceptable tradeoff. This is better than the alternative where we have to pay the bookkeeping cost whenever we teardown a component during normal rendering.

packages/ember-views/lib/mixins/child_views_support.js
```diff
@@ -3,69 +3,40 @@
...
+  appendChild(view) {
+    this.linkChild(view);
+    addChildView(this, view);
   },
 
   linkChild(instance) {
-    if (!instance[OWNER]) {
+    if (!getOwner(instance)) {
       setOwner(instance, getOwner(this));
     }
```
chancancode: @rwjblue @dgeb as a follow-up, do you think we still need this?

packages/ember-views/lib/system/utils.js
```diff
@@ -20,6 +24,76 @@ export const STYLE_WARNING = '' +
 
 /**
   @private
+  @method getChildViews
+  @param {Ember.View} view
```
teddyzeenny: The should be `owner` instead of `view`.

packages/ember-views/lib/system/utils.js
```diff
@@ -20,6 +24,76 @@ export const STYLE_WARNING = '' +
 
 /**
   @private
+  @method getChildViews
```
teddyzeenny: `getRootViews`

teddyzeenny:
Shouldn't `getRootViews` and `getChildViews` be exposed on the `ViewUtils` object [here](https://github.com/emberjs/ember.js/blob/fa4360a008d69beb006e9c76290b1b7dd2853a07/packages/ember-views/lib/index.js#L36-L39)?

teddyzeenny:
Updated in https://github.com/emberjs/ember.js/pull/14143

packages/ember-views/lib/mixins/child_views_support.js
rwjblue: No, likely not needed.

PR #14140, opened 2016-08-27
[BUGFIX beta] Ensure injections happen in engine instances.

rwjblue:
Prior to this `_environment` could not be relied upon in routes or views (which is used when `shouldRender: false` is used with `visit`).

The refactor uses the environment to determine what injections to setup in the engine-instance, and has the `application-instance` defer to the `engine-instance` to avoid duplication.

Also, removes manual threading of `props.renderer` (as requested in https://github.com/emberjs/ember.js/pull/14139).

---

Ensure that all components receive type injections. 

Prior to this change we were defaulting to `Ember.Component` which does not receive injections (which meant it was missing the `renderer`). This worked "ok" in HTMLBars since we created a default global renderer to handle components/views without a container/owner.

This change ensures that in the template only component case we fall back to using the default component looked up from the owner. This ensures that injections are present.

### DIFF
--- a/packages/ember-application/lib/system/application-instance.js
+++ b/packages/ember-application/lib/system/application-instance.js
@@ -283,18 +283,9 @@ ApplicationInstance.reopenClass({
   */
   setupRegistry(registry, options = new BootOptions()) {
     registry.register('-environment:main', options.toEnvironment(), { instantiate: false });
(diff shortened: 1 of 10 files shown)


rwjblue:
Added another commit to ensure that all components receive type injections. See description for more details.

PR #14143, opened 2016-08-28
Expose `getRootViews and `getClientViews` on `ViewUtils`

teddyzeenny:


packages/ember-views/lib/index.js
```diff
@@ -37,6 +39,8 @@ const ViewUtils = Ember.ViewUtils = {};
 ViewUtils.isSimpleClick = isSimpleClick;
 ViewUtils.getViewClientRects = getViewClientRects;
 ViewUtils.getViewBoundingClientRect = getViewBoundingClientRect;
+ViewUtils.getRootViews = getRootViews;
+ViewUtils.getChildViews = getChildViews;
```
chancancode: @rwjblue does this seem good to you?

I think `getViewBounds` is in the same ballpark in terms of stability, so if we are adding these, maybe we should add that here too

teddyzeenny:
Added `getViewBounds`

packages/ember-views/lib/index.js
```diff
@@ -37,6 +40,9 @@ const ViewUtils = Ember.ViewUtils = {};
...
+ViewUtils.getViewBounds = getViewBounds;
```
chancancode: nit: can you bring this above getViewClientRects? 😄

teddyzeenny: done 😄

PR #14144, opened 2016-08-28
Migrate renderToElement tests

chancancode:
cc @rwjblue


packages/ember-glimmer/tests/integration/components/append-test.js
```diff
@@ -0,0 +1,316 @@
...
+    this.runTask(() => wrapper1 = this.append(this.first));
+    this.runTask(() => wrapper2 = this.append(this.second));
+
+    let componentElement1 = this.first.element;
+    let componentElement2 = this.second.element;
+
+    this.assertComponentElement(componentElement1, { content: 'x-first foo' });
+    this.assertComponentElement(componentElement2, { content: 'x-second bar' });
+
+    assert.equal(componentElement1.parentElement, wrapper1, 'The first component should be attached to the target');
```
rwjblue: Need to add `assert` argument to test function declaration.

packages/ember-glimmer/tests/integration/components/append-test.js
```diff
@@ -0,0 +1,316 @@
...
+    this.runTask(() => this.component.destroy());
+
+    assert.ok(!this.component.element, 'It should not have an element');
+
+    assert.ok(!componentElement.parentElement, 'The component element should be detached');
+
+    this.assert.equal(willDestroyCalled, 1);
+  }
+
+  ['@htmlbars appending, updating and destroying multiple components']() {
```
rwjblue: Need to add assert argument to test function declaration since it is used below.

packages/ember-glimmer/tests/integration/components/append-test.js
```diff
@@ -0,0 +1,316 @@
...
+
+    return wrapper;
+  }
+
+});
+
+/*
+
+
+  ['@test calling appendTo should append to the specified element']() {
```
rwjblue: Should probably just delete these I think...

PR #14148, opened 2016-08-29
[GLIMMER] Introduce mount syntax.

dgeb:
Introduces `{{mount}}` syntax for Glimmer.

### DIFF
--- a/packages/ember-glimmer/lib/syntax/mount.js
+++ b/packages/ember-glimmer/lib/syntax/mount.js
@@ -1,0 +1,97 @@
+import { ArgsSyntax, StatementSyntax } from 'glimmer-runtime';
+import { ConstReference, isConst, UNDEFINED_REFERENCE } from 'glimmer-reference';
+import { assert } from 'ember-metal/debug';
+import { RootReference } from '../utils/references';
+import { generateControllerFactory } from 'ember-routing/system/generate_controller';
+import { OutletLayoutCompiler } from './outlet';
+
+function makeComponentDefinition(vm) {
+  let env     = vm.env;
+  let args    = vm.getArgs();
+  let nameRef = args.positional.at(0);
+
+  assert(`The first argument of {{mount}} must be quoted, e.g. {{mount "chat-engine"}}.`, isConst(nameRef));
+
+  let name = nameRef.value();
+
+  assert(`You used \`{{mount '${name}'}}\`, but '${name}' can not be found as an engine.`, env.owner.hasRegistration(`engine:${name}`));
+
+  let engine = env.owner.buildChildEngineInstance(name);
+
+  let template = engine.lookup(`template:application`);
+
+  engine.boot();
+
+  return new ConstReference(new MountDefinition(name, engine, template, env, MOUNT_MANAGER));
+}
+
+export class MountSyntax extends StatementSyntax {
+  constructor({ args, symbolTable }) {
+    super();
+    this.definitionArgs = args;
+    this.definition = makeComponentDefinition;
+    this.args = ArgsSyntax.fromPositionalArgs(args.positional.slice(1, 2));
+    this.templates = null;
+    this.symbolTable = symbolTable;
+    this.shadow = null;
+  }
+
+  compile(builder) {
+    builder.component.dynamic(this.definitionArgs, this.definition, this.args, this.templates, this.symbolTable, this.shadow);
+  }
+}
+
+class MountManager {
+  prepareArgs(definition, args) {
+    return args;
+  }
+
+  create(definition, args, dynamicScope) {
+    let { engine } = definition;
+
+    let factory = engine._lookupFactory(`controller:application`) || generateControllerFactory(engine, 'application');
+    let controller = factory.create();
+
+    dynamicScope.outletState = UNDEFINED_REFERENCE;
+
+    return { controller };
+  }
+
+  layoutFor(definition, bucket, env) {
+    return env.getCompiledBlock(OutletLayoutCompiler, definition.template);
+  }
+
+  getSelf({ controller }) {
+    return new RootReference(controller);
+  }
+
+  getTag(state) {
+    return null;
+  }
+
+  getDestructor(state) {
+    return state.controller;
+  }
+
+  didCreateElement() {}
+  didRenderLayout() {}
+  didCreate(state) {}
+  update(state, args, dynamicScope) {}
+  didUpdateLayout() {}
+  didUpdate(state) {}
+}
+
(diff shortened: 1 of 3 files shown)


packages/ember-glimmer/lib/syntax/mount.js
```diff
@@ -0,0 +1,101 @@
...
+
+  create(definition, args, dynamicScope) {
+    let { name, env } = definition;
+    let modelRef = args.positional.at(0);
+
+    let factory = env.owner._lookupFactory(`controller:${name}`) || generateControllerFactory(env.owner, name);
+    let controller = factory.create({ model: modelRef.value() });
+
+    if (dynamicScope.rootOutletState) {
+      dynamicScope.outletState = dynamicScope.rootOutletState.getOrphan(name);
```
rwjblue: I am unsure if we need this. Seems like we generally do not support {{outlet inside of a {{mount....

packages/ember-glimmer/lib/syntax/mount.js
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    return { controller };
+  }
+
+  update({ controller }, args, dynamicScope) {
+    controller.set('model', args.positional.at(0).value());
```
rwjblue: We do not support passing > 1 argument (basically only the engine name). We should be able to make this a noop.

packages/ember-glimmer/lib/syntax/mount.js
```diff
@@ -0,0 +1,97 @@
...
+
+  return new ConstReference(new MountDefinition(name, engine, template, env, MOUNT_MANAGER));
+}
+
+export class MountSyntax extends StatementSyntax {
+  constructor({ args, symbolTable }) {
+    super();
+    this.definitionArgs = args;
+    this.definition = makeComponentDefinition;
+    this.args = ArgsSyntax.fromPositionalArgs(args.positional.slice(1, 2));
```
chancancode: I think this could just be `ArgsSyntax.empty()` – `{{mount}}` doesn't take more than one argument right?

rwjblue: Confirm, it accepts a single argument.  We should assert if more are passed (or if hash arguments are passed).  We want to reserve `{{mount 'foo' some=args passed=in}}` to allow passing data into the engine (will likely be future RFC work though).

packages/ember-glimmer/lib/syntax/mount.js
```diff
@@ -0,0 +1,97 @@
...
+    this.templates = null;
+    this.symbolTable = symbolTable;
+    this.shadow = null;
+  }
+
+  compile(builder) {
+    builder.component.dynamic(this.definitionArgs, this.definition, this.args, this.templates, this.symbolTable, this.shadow);
```
chancancode: Actually, since nothing actually changes, I think we can just use `builder.component.static` (https://github.com/tildeio/glimmer/blob/master/packages/glimmer-runtime/lib/opcode-builder.ts#L24) like https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/syntax/curly-component.js#L106-L119

chancancode: AH, but the name is still coming from the args. This is a bit silly since we actually require the args to be static. But okay for now.

packages/ember-glimmer/lib/syntax/mount.js
```diff
@@ -0,0 +1,97 @@
...
+
+  let name = nameRef.value();
+
+  assert(`You used \`{{mount '${name}'}}\`, but '${name}' can not be found as an engine.`, env.owner.hasRegistration(`engine:${name}`));
+
+  let engine = env.owner.buildChildEngineInstance(name);
+
+  let template = engine.lookup(`template:application`);
+
+  engine.boot();
```
chancancode: In general we should defer any expensive work (and work that requires cleanup) till we actually need to render (this could be wrapped inside a conditional, etc), so I'd move this to `Manager.create`

packages/ember-glimmer/lib/syntax/mount.js
```diff
@@ -0,0 +1,97 @@
...
+}
+
+class MountManager {
+  prepareArgs(definition, args) {
+    return args;
+  }
+
+  create(definition, args, dynamicScope) {
+    let { engine } = definition;
+
```
chancancode: I would move `engine.boot()` here

packages/ember-glimmer/lib/syntax/mount.js
```diff
@@ -0,0 +1,97 @@
...
+  getSelf({ controller }) {
+    return new RootReference(controller);
+  }
+
+  getTag(state) {
+    return null;
+  }
+
+  getDestructor(state) {
+    return state.controller;
```
chancancode: I _think_ we want to teardown the entire engine instance here, is that correct?

For something like `{{#if showFoo}}{{mount "foo"}}{{/if}}` do we expect to create a new engine instance every time we toggle it?

packages/ember-glimmer/lib/syntax/mount.js
```diff
@@ -0,0 +1,97 @@
...
+    let factory = engine._lookupFactory(`controller:application`) || generateControllerFactory(engine, 'application');
+    let controller = factory.create();
+
+    dynamicScope.outletState = UNDEFINED_REFERENCE;
+
+    return { controller };
+  }
+
+  layoutFor(definition, bucket, env) {
+    return env.getCompiledBlock(OutletLayoutCompiler, definition.template);
```
chancancode: @rwjblue what did we end up doing about caching? Does this cache it with the right owner?

packages/ember-glimmer/lib/syntax/mount.js
rwjblue: Yes, that is what we intend to do.

chancancode: I think that probably implies we should do all of the throw-away work in manager.create, instead of saving it onto the definition object.

Each time `{{mount "foo"}}` is rendered, it gets the same definition object passed into `.create`.

packages/ember-glimmer/lib/syntax/mount.js
chancancode: seems like this should pass engine as the third argument

rwjblue: Yep, also seems like we should add an assertion there to make sure this isn't missed again 😝

rwjblue:
Was closed by #14154.

PR #14149, opened 2016-08-29
Outlet reference protocol change for liquid-fire

ef4:
This makes the outlet reference protocol compatible with normal NestedPropertyReferences.

The glimmer2-compatible version of liquid-fire uses dynamically-scoped variable accessors as implemented in https://github.com/tildeio/glimmer/pull/294 in order to read and write the outlet state. (This is the value that is named `routeInfo` in https://github.com/emberjs/rfcs/pull/95, which exists to eventually make public API for it.)

That all works great except for one tiny annoyance in Ember itself: we can't bring an OutletReference into "userspace" and back again. Once we compute a value, we have an untyped POJO that results in a NestedPropertyReference when we pass it back into glimmer.

This change is the simplest thing that works. Alternatively, we should implement a real `RouteInfo` type (as described in https://github.com/emberjs/rfcs/pull/95) so that the values can round-trip through userspace and still get converted back into typed references.

### DIFF
--- a/packages/ember-glimmer/lib/environment.js
+++ b/packages/ember-glimmer/lib/environment.js
@@ -4,7 +4,8 @@ import {
   Environment as GlimmerEnvironment,
   AttributeChangeList,
   isSafeString,
-  compileLayout
+  compileLayout,
+  getDynamicVar
 } from 'glimmer-runtime';
 import Cache from 'ember-metal/cache';
 import { assert, warn, runInDebug } from 'ember-metal/debug';
@@ -108,7 +109,8 @@ export default class Environment extends GlimmerEnvironment {
       '-each-in': eachIn,
       '-input-type': inputTypeHelper,
       '-normalize-class': normalizeClassHelper,
-      '-html-safe': htmlSafeHelper
+      '-html-safe': htmlSafeHelper,
+      '-get-dynamic-var': { glimmerNativeHelper: getDynamicVar }
     };
   }
 
@@ -290,6 +292,8 @@ export default class Environment extends GlimmerEnvironment {
       return (vm, args) => SimpleHelperReference.create(helper.compute, args);
     } else if (helper.isHelperFactory) {
       return (vm, args) => ClassBasedHelperReference.create(helper, vm, args);
+    } else if (helper.glimmerNativeHelper) {
+      return helper.glimmerNativeHelper;
     } else {
       throw new Error(`${nameParts} is not a helper`);
     }
--- a/packages/ember-glimmer/lib/renderer.js
+++ b/packages/ember-glimmer/lib/renderer.js
@@ -38,6 +38,15 @@ class DynamicScope {
       this.view, this.outletState, this.rootOutletState, this.isTopLevel, this.targetObject
     );
   }
+
+  get(key) {
+    return this[key];
+  }
+
+  set(key, value) {
+    this[key] = value;
+    return value;
+  }
 }
 
 const renderers = [];
--- a/packages/ember-glimmer/lib/syntax.js
+++ b/packages/ember-glimmer/lib/syntax.js
@@ -3,6 +3,8 @@ import { OutletSyntax } from './syntax/outlet';
 import { MountSyntax } from './syntax/mount';
 import { DynamicComponentSyntax } from './syntax/dynamic-component';
 import { InputSyntax } from './syntax/input';
+import { WithDynamicVarsSyntax } from 'glimmer-runtime';
+
 
 let syntaxKeys = [];
 let syntaxes = [];
@@ -25,3 +27,8 @@ registerSyntax('outlet', OutletSyntax);
 registerSyntax('mount', MountSyntax);
 registerSyntax('component', DynamicComponentSyntax);
 registerSyntax('input', InputSyntax);
+registerSyntax('-with-dynamic-vars', class {
+  static create(environment, args, templates, symbolTable) {
+    return new WithDynamicVarsSyntax({ args, templates });
+  }
+});
--- a/packages/ember-glimmer/lib/syntax/outlet.js
+++ b/packages/ember-glimmer/lib/syntax/outlet.js
@@ -88,7 +88,7 @@ class OutletComponentReference {
 
 
     let outletName = outletNameRef.value();
-    let outletStateRef = parentOutletStateRef.get(outletName);
+    let outletStateRef = parentOutletStateRef.get('outlets').get(outletName);
     let newState = this.lastState = outletStateRef.value();
 
     this.outletStateTag.update(outletStateRef.tag);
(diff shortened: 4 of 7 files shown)


rwjblue:
Due to working on #14177 I actually half way understand this. I think you need to update to incorporate the issues fixed in #14177 (that the outletName itself can be bound and that reference should affect the resulting tag).  With these changes, it seems that the `outlets` key is also dynamic, so it should likely be part of the combined tag for the outlet component definition as well.

Can you add a test to https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/tests/integration/outlet-test.js to help us ensure no regressions in this after landing?

ef4:
Yes, and this PR also needs updates to reflect https://github.com/tildeio/glimmer/pull/297. I will work on it.

ef4:
Updated, tests green, new test added, and I have the liquid-fire test suite fully passing against this.
🎉 

> With these changes, it seems that the outlets key is also dynamic, 

The `outletState` objects generated by the router in normal usage are never mutated. That is a deliberate design decision. Whenever something about the route changes, we generate a completely fresh new outletState object and send it downward. So I don't think the outlets key really needs to be tracked dynamically.

I see that some tests are mutating and reusing outletState objects as a shortcut. That is not really testing things the same way they work in apps. The tests in https://github.com/ember-animation/liquid-fire/blob/glimmer2/tests/integration/helpers/liquid-outlet-test.js deliberately copy the outletState each time they set it (look inside the `asTop` method) to avoid this issue while still keeping the tests from being annoying to write.

rwjblue:
> I see that some tests are mutating and reusing outletState objects as a shortcut. That is not really testing things the same way they work in apps. 

Aha! I see.  I will try to take a stab at refactoring in a future PR.

packages/ember-glimmer/lib/environment.js
```diff
@@ -290,6 +292,8 @@ export default class Environment extends GlimmerEnvironment {
       return (vm, args) => SimpleHelperReference.create(helper.compute, args);
     } else if (helper.isHelperFactory) {
       return (vm, args) => ClassBasedHelperReference.create(helper, vm, args);
+    } else if (helper.glimmerNativeHelper) {
+      return helper.glimmerNativeHelper;
```
chancancode: @ef4 I think we can simplify this by making internal helpers just a function with this interface: https://github.com/tildeio/glimmer/blob/master/packages/glimmer-runtime/lib/environment.ts#L239-L241

Then you wouldn't need the `isInternalHelper` or `glimmerNativeHelper` flags – which are problematic because a userland helper can pretend to be one of those just by assigning that property on them.

Instead:

```
let builtInHelper = this.builtInHelpers[name];

if (builtInHelper) {
  return builtInHelper;
}

// do userland lookups...
```

packages/ember-glimmer/lib/renderer.js
```diff
@@ -38,6 +38,15 @@ class DynamicScope {
...
   }
+
+  get(key) {
+    return this[key];
+  }
+
+  set(key, value) {
+    this[key] = value;
+    return value;
+  }
```
chancancode: This changes the shape for dynamic scope, and also most of the other things on the dynamic scope are private (and more importantly – not a `PathRefence<Opaque>` as required by the `get` method on the interface).

Until we have the registration story figured out, we should just throw here (for both `get` and `set`) if the `key` is not `=== "outletState"`

packages/ember-glimmer/lib/syntax.js
```diff
@@ -25,3 +27,8 @@ registerSyntax('outlet', OutletSyntax);
 registerSyntax('mount', MountSyntax);
 registerSyntax('component', DynamicComponentSyntax);
 registerSyntax('input', InputSyntax);
+registerSyntax('-with-dynamic-vars', class {
+  static create(environment, args, templates, symbolTable) {
+    return new WithDynamicVarsSyntax({ args, templates });
+  }
```
chancancode: (This could have been a POJO instead of a class)

packages/ember-glimmer/lib/syntax/outlet.js
```diff
@@ -201,7 +201,7 @@ const TOP_LEVEL_MANAGER = new TopLevelOutletComponentManager();
 
 class OutletComponentManager extends AbstractOutletComponentManager {
   create(definition, args, dynamicScope) {
-    let outletStateReference = dynamicScope.outletState = dynamicScope.outletState.get(definition.outletName);
+    let outletStateReference = dynamicScope.outletState = dynamicScope.outletState.get('outlets').get(definition.outletName);
```
chancancode: This change feels a bit dubious to me – the reason for this change is that you could swap in the outlet state reference for a regular `EmberPathReference` in your code. However, we are still using the `ChildOutletStateReference` by default, which doesn't fully behave or duck-type like a regular path reference. For example, it has a `get isTopLevel` on it, which your `EmberPathReference` is not going to have. (I _think_ it happened to be unused, but we should just remove it if true.)

If we need to support replacing this with a `EmberPathReference`, I think we need to actually use that ourselves to guarantee compatibility. At minimum, we need the top-level `OutletStateReference` to return a `EmberPathReference` for its `get`, which would go through the regular paths from there onwards.

packages/ember-glimmer/tests/integration/outlet-test.js
```diff
@@ -278,4 +278,50 @@ moduleFor('outlet view', class extends RenderingTest {
...
+
+  ['@test outletState can pass through user code (liquid-fire initimate API) ']() {
+    this.registerTemplate('outer', 'A{{#-with-dynamic-vars outletState=(identity (-get-dynamic-var "outletState"))}}B{{outlet}}D{{/-with-dynamic-vars}}E');
+    this.registerTemplate('inner', 'C');
+
+    // This looks like it doesn't do anything, but its presence
+    // guarantees that the outletState gets converted from a reference
+    // to a value and then back to a reference. That is what we're
+    // testing here.
+    this.registerHelper('identity', ([a]) => a);
```
chancancode: Can you make this do _something_ instead of pass-through? This allows you to test updating behavior (see below)

packages/ember-glimmer/tests/integration/outlet-test.js
```diff
@@ -278,4 +278,50 @@ moduleFor('outlet view', class extends RenderingTest {
...
+      }
+    };
+
+    this.runTask(() => this.component.setOutletState({ render: {}, outlets: { main: outletState } }) );
+
+    runAppend(this.component);
+
+    this.assertText('ABCDE');
+
+    this.assertStableRerender();
```
chancancode: In general, we want to make sure the updating paths work and so we follow the [I-N-U-R pattern](https://github.com/emberjs/ember.js/issues/13127) for new tests.

Maybe you can get your helper to override the template based on a dynamic property, which would allow you to change it to a new template then replace it with the original. (Or if you could get it to do something similar to what LF is ultimately doing, then even better.)

packages/ember-glimmer/lib/syntax/outlet.js
chancancode: I guess we don't _have_ to use an `EmberPathReference` if we think the `ChildOutletStateReference` is simpler and faster (and looking at your test reminds me that it could have been _any_ PathReference, in your test it would be a HelperReference). I think the key is that we cannot rely on anything that is not strictly part of the usual `PathReference` interface – so things like `isTopLevel` raises a red-flag.

ef4: I think we don't really need a separate reference type for outlet states at all. I suspect we would be better off solving this with a better type at the value level, not the reference level.

PR #14150, opened 2016-08-29
[DOC release] Update `didInsertElement` documentation.

locks:
cc @rwjblue @wycats


rwjblue:
Seems good. Can you prefix with `[DOC release]` and squash?

PR #14153, opened 2016-08-29
add license to bower.json

anotheredward:


rwjblue:
This is not the `bower.json` you are looking for 😉.

Our official published [build repo](https://github.com/components/ember/) includes [this bower.json](https://github.com/emberjs/ember.js/blob/master/config/package_manager_files/bower.json#L4) which has the proper license attribution.

PR #14159, opened 2016-08-30
[BUGFIX beta] Fix cleanup

krisselden:


rwjblue:
Pushed a fix for the failures on IE (`document.contains` doesn't exist in IE).

PR #14171, opened 2016-08-31
[BUGFIX release] Fix array QP serialization

### LINKED ISSUE #13591: _array_ queryParam converted to _string_ after transitionTo
Hi,

a demo: https://ember-twiddle.com/650b6fb3657bf037a40170f3b1e74b84

Steps to reproduce (here's a quick video: https://v.usetapes.com/V0vlZepqbw):

1. Use "My Route!" link.
2. Click either "Set array" or "Set all" to update the `array` parameter.
3. The `array` parameter is now a string (and should remain an array).

Notes:
* "my-route" route uses application controller `controllerName: 'application'` to access query parameters from the parent route/controller.
* When you comment out `this.transitionTo('application')` setting parameters will work as expected (`array` parameter remains an array).
* "reset params" sets an empty array and that doesn't trigger the error.
* I've noticed the error after updating from Ember **2.4.1** to **2.4.5**. I can confirm that *2.4.1* doesn't have this error.

Thanks,

Artur

lan0:
A twiddle reproducing the issue: https://ember-twiddle.com/c0f6b0efd71731ecb7e474ddedc7a7a4 - after a click on set array and then on refresh the array QP becomes a string.

`"[1]"` gets stringified in the second call to `serializeQueryParam`, resulting in the serialized param being `""[1]""` instead of `"[1]"`.

This fixes #13591 and is similar to #13816.

### DIFF
--- a/packages/ember-routing/lib/system/route.js
+++ b/packages/ember-routing/lib/system/route.js
@@ -431,7 +431,7 @@ let Route = EmberObject.extend(ActionHandler, Evented, {
     // urlKey isn't used here, but anyone overriding
     // can use it to provide serialization specific
     // to a certain query param.
-    if (defaultValueType === 'array') {
+    if (defaultValueType === 'array' && Array.isArray(value)) {
       return JSON.stringify(value);
     }
     return `${value}`;
--- a/packages/ember/tests/routing/query_params_test.js
+++ b/packages/ember/tests/routing/query_params_test.js
@@ -2530,6 +2530,25 @@ if (isEnabled('ember-routing-route-configured-query-params')) {
     equal(router.get('location.path'), '/?foo=3', 'url is correct');
   });
 
+  QUnit.test('previously set array queryParam is preserved when a controller property is set and the route is refreshed.', function() {
+   App.ApplicationController = Controller.extend({
+     queryParams: ['array', 'foo'],
+     array: [],
+     foo: null,
+   });
+
+   startingURL = '/';
+   bootApplication();
+   let applicationRoute = container.lookup('route:application');
+   let applicationController = container.lookup('controller:application');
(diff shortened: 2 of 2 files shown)


nathanhammond:
@lan0 I believe that both this approach and the original behavior are wrong. Per our adoption of `route-recognizer` and the way it handles arrays the result should be:

```javascript
QS.stringify({
  array: [ 1 ]
});
// => ?array[] = 1
```

nathanhammond:
So these are the lines which implement the behavior:
https://github.com/emberjs/ember.js/blob/master/packages/ember-routing/lib/system/route.js#L421-L464

This is a completely custom (and broken) QP serialization strategy which doesn't match our approach for `route-recognizer`. The behavior for non-string and non-numeric values should be considered broken since _who knows when_ and therefore should not be relied upon.

nathanhammond:
Last note: this should be compared to `route-recognizer`'s current QP implementation:
- https://github.com/tildeio/route-recognizer/blob/master/lib/route-recognizer.js#L455-L483
- https://github.com/tildeio/route-recognizer/blob/master/lib/route-recognizer.js#L423-L453

The real bug here is actually that these are inconsistent. As a result we should not merge this PR (which will create yet another new failure mode) and instead solve it at the more-fundamental level.

Blog post coming on query strings this week.

lan0:
Agreed that the array serialization should be `?array[]=`, but it worked for us so we rolled with it long ago.

Closing this since we fixed it in our app by overwriting `serializeQueryParam` in routes where there are array QP:
```
  serializeQueryParam(value, urlKey, defaultValueType) {
    if (defaultValueType === 'array' && Array.isArray(value)) {
      return JSON.stringify(value);
    }
    return `${value}`;
  },
```

PR #14172, opened 2016-08-31
Refactor {{mount}} to error early and port {{mount}} tests

chancancode:
Starting point for adding more {{mount}} tests


dgeb:
Looks good. Just a couple linting errors:

> ember-glimmer/tests/utils/test-case.js: line 6, col 8, 'assign' is defined but never used.
> ember-glimmer/tests/utils/test-case.js: line 7, col 10, 'GLIMMER' is defined but never used.

PR #14173, opened 2016-08-31
[CLEANUP] Remove conditional testing infra

chancancode:
r? @rwjblue


packages/ember-views/lib/mixins/class_names_support.js
```diff
@@ -4,9 +4,8 @@
 */
 import { assert } from 'ember-metal/debug';
 import { Mixin } from 'ember-metal/mixin';
-import { A as emberA } from 'ember-runtime/system/native_array';
 
-const EMPTY_ARRAY = [];
+var EMPTY_ARRAY = [];
```
rwjblue: O_o

Also, we should freeze this array to ensure we don't pwn ourselves...

chancancode: Oops, I just copied and pasted it from the other file without checking

Issue #14170, opened 2016-08-31
Fix remaining skipped tests.


rwjblue:
The following tests are currently skipped:

- [ ] `Ember.Application - visit() Integration Tests: SKIPPED: Test setup (0)`
- [ ] `Ember.Application - visit() Integration Tests: SKIPPED: iframe setup (0)`
- [x] `[glimmer] append: no arguments (attaching to document.body): SKIPPED: appending, updating and destroying multiple components (0)`
- [x] `[glimmer] appendTo: a selector: SKIPPED: appending, updating and destroying multiple components (0)`
- [x] `[glimmer] appendTo: an element: SKIPPED: appending, updating and destroying multiple components (0)`
- [x] `[glimmer] renderToElement: no arguments (defaults to a body context): SKIPPED: appending, updating and destroying multiple components (0)`
- [x] `[glimmer] renderToElement: a div: SKIPPED: appending, updating and destroying multiple components (0)`
- [x] `[glimmer] Helpers test: closure {{action}} improved instrumentation: SKIPPED: action should fire interaction event (0)`
- [x] `[glimmer] Helpers test: closure {{action}} improved instrumentation: SKIPPED: interaction event subscriber should be passed parameters (0)`
- [x] `[glimmer] Helpers test: closure {{action}} improved instrumentation: SKIPPED: interaction event subscriber should be passed target (0)`
- [x] `[glimmer] Helpers test: {{render}}: SKIPPED: should render a template without a model only once (0)`
- [ ] `[glimmer] Helpers test: {{unbound}}: SKIPPED: helper form updates on parent re-render (0)`
  * Being tracked in https://github.com/emberjs/ember.js/issues/13946.  I believe that the likely outcome is that we remove the test...
- [x] `[glimmer] outlet view: SKIPPED: should render the outlet when set after DOM insertion (0)Rerun`
- [x] `[glimmer] outlet view: SKIPPED: should support bound outlet name (0)`
- [x] `system/props/events_test: SKIPPED: DEPRECATED: adding didInitAttrs as a listener is deprecated (0)`
- [ ] `Routing with Query Params: SKIPPED: can opt into full transition by setting refreshModel in route queryParams when transitioning from child to parent when all configuration is on route (0)`

chancancode:
FWIW, the two `visit` tests are legitimate pending features unrelated to Glimmer – they are not yet supported, but probably something we want to do at some point

rwjblue:
[...]

PR #14177, opened 2016-09-01
Fix skipped tests related to outlets.

rwjblue:
- Fix issue with setting outlet state after initial render.
  - When the initially rendered outlet state was `undefined` an error
    occurred (due to not handling the missing template case).
  - When a completely new outlet state object is setup, the `.value`
    calculation for `TopLevelOutletComponentReference` was completely
    incorrect (it was looking for `newState.outlets.main.render.template`
    instead of `newState.render.template`).
- Fix usage of bound outlet name. Previously, we were eagerly evaluating the bound reference and passing the final value around instead of the reference to the final outlet name value. This means that given `{{outlet foo}}` when the property `foo` changes then no updates would be made. The fix is to avoid eagerly evaluating the reference, band and ensure that the `ComponentDefition` includes a tag the represents both the outlet name and the outlet state. While debugging this, I also came across an untested scenario where we coerced any falsy values for outlet name into `'main'`. However, we should only be defaulting to `'main'` if no positional argument is present.

Addresses 2 more tests in https://github.com/emberjs/ember.js/issues/14170.

### DIFF
--- a/packages/ember-glimmer/lib/syntax/outlet.js
+++ b/packages/ember-glimmer/lib/syntax/outlet.js
@@ -58,11 +58,15 @@ class TopLevelOutletComponentReference {
   }
 
   value() {
-    let lastState = this.lastState;
-    let newState = this.outletReference.value();
+    let { lastState, outletReference, definition } = this;
+    let newState = outletReference.value();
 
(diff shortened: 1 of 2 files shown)


chenyingyin:
default

PR #14179, opened 2016-09-01
Add support for multiple roots.

rwjblue:
To support things like `ember-islands` we need to refactor the renderer to allow for multiple roots (in other words multiple components can call `appendTo`).
- [x] Refactor single root system
- [x] Add support for multiple roots
- [x] @krisselden r?
- [x] @chancancode r?

### DIFF
--- a/packages/ember-glimmer/lib/renderer.js
+++ b/packages/ember-glimmer/lib/renderer.js
@@ -49,6 +49,66 @@ class DynamicScope {
   }
 }
 
+let nextRootId = 0;
+class RootState {
+  constructor(env, root, template, self, parentElement, dynamicScope) {
+    this.id = ++nextRootId;
+    this.env = env;
+    this.root = root;
+    this.template = template;
+    this.self = self;
+    this.parentElement = parentElement;
+    this.dynamicScope = dynamicScope;
+
+    this.options = {
+      alwaysRevalidate: false
+    };
+    this.render = this.initialRender;
+    this.lastRevision = undefined;
+    this.result = undefined;
+  }
+
+  isFor(possibleRoot) {
+    return this.root === possibleRoot;
+  }
+
+  initialRender() {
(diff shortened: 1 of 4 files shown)


packages/ember-glimmer/lib/renderer.js
```diff
@@ -40,6 +40,63 @@ class DynamicScope {
   }
 }
 
+let nextRootId = 0;
+class RootState {
+  constructor(env, root, template, self, parentElement, dynamicScope) {
+    assert(`You cannot render \`${self.value()}\` without a template.`, template);
+
+    this.id = ++nextRootId;
```
krisselden: Don't we already have ids on views and a registry that we could use to weakly associate meta?

rwjblue: Ya, probably.  I am not really using this `id` at all, I was mostly just using it during debugging to make sure I was looking at the same `RootState` instance.

packages/ember-glimmer/lib/renderer.js
```diff
@@ -40,6 +40,49 @@ class DynamicScope {
...
+      let result = this.result = template.asEntryPoint().render(self, env, {
+        appendTo: parentElement,
+        dynamicScope
+      });
+
+      // override .render function after initial render
+      this.render = () => {
+        result.rerender(options);
+      };
+    };
```
chancancode: It seems like we can make all of this a prototype method by storing all these closed over states

```
render() {
  if (this.result) {
    // ...
  } else {
    // ...
  }
}
```

rwjblue: Indeed, but then we have to keep ahold of all of the other values needed for initial render. Also, it would require a much larger function to be optimized when 99.99999% of the time we only want the smaller version (since we only initial render once).

PR #14184, opened 2016-09-01
run(rsvp, "reject") should not throw trying to dispatchError

### LINKED ISSUE #14182: RSVP.reject where the arg is not of type Error breaks error dispatching
The following will throw with `Cannot read property 'indexOf' of undefined`

`run(RSVP, 'reject', 'foo');`

(https://github.com/emberjs/ember.js/blob/39b606fd40b1ca1a5b8fb69d81d99c446f3a86e7/packages/ember-metal/lib/error_handler.js#L5-L14) always assumes what it received is it an instance of an Error and has a stack.

This is currently happening on ember-alpha, passes on current stable release.

Fail test: https://github.com/jasonmit/ember.js/commit/54d16657ef617e9c3e29e14a1b06e0fa3501a686

jasonmit:
Fixes #14182

### DIFF
--- a/packages/ember-metal/lib/error_handler.js
+++ b/packages/ember-metal/lib/error_handler.js
@@ -6,7 +6,7 @@ let getStack = function(error) {
   var stack = error.stack;
   var message = error.message;
 
-  if (stack.indexOf(message) === -1) {
+  if (stack && stack.indexOf(message) === -1) {
     stack = message + '\n' + stack;
   }
 
--- a/packages/ember-runtime/tests/ext/rsvp_test.js
+++ b/packages/ember-runtime/tests/ext/rsvp_test.js
@@ -35,6 +35,38 @@ QUnit.test('TransitionAborted errors are not re-thrown', function() {
   ok(true, 'did not throw an error when dealing with TransitionAborted');
 });
 
+QUnit.test('Can reject with non-Error object', function(assert) {
+  let wasEmberTesting = isTesting();
+  setTesting(false);
+  expect(1);
+
+  try {
+    run(RSVP, 'reject', 'foo');
+  } catch(e) {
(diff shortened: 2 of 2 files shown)


packages/ember-metal/lib/error_handler.js
```diff
@@ -6,7 +6,7 @@ let getStack = function(error) {
   var stack = error.stack;
   var message = error.message;
 
-  if (stack.indexOf(message) === -1) {
+  if (stack && stack.indexOf(message) === -1) {
```
jasonmit: synonymous with the previous behavior where it will Log/dispatch undefined if the error did not have a stack

rwjblue:
Seems odd that the test is in runtime, but implementation is in metal. I guess the way your are excercising the bug is via RSVP (which is only in runtime). Is it possible to also add a test in metal?

Also, can you determine what published major/minor  this issue was introduced in? We should likely target that version for this BUGFIX.

jasonmit:
> Seems odd that the test is in runtime, but implementation is in metal. I guess the way your are excercising the bug is via RSVP (which is only in runtime). Is it possible to also add a test in metal?

I put it where I thought it related with the other tests, I can move to metal though since you're likely right.

> Also, can you determine what published major/minor this issue was introduced in?

I don't believe it's in a stable release yet but I will verify.

jasonmit:
> Also, can you determine what published major/minor this issue was introduced in?

2.8.0-beta.1 is where it was first introduced.

rwjblue:
Nah, this test is good here.

jasonmit:
@rwjblue done

jasonmit:
@rwjblue ready and the one selenium job failing is failing against everyones PR so I'm assuming it's unrelated.

Issue #14166, opened 2016-08-31
Detailed API Documentation Review: ember-htmlbars -> ember-glimmer


rwjblue:
All of the functionality of our templating system has been migrated from ember-htmlbars to ember-glimmer package.  We need to ensure that we have not lost significant documentation coverage.

At a minimum we need to review the following files in ember-htmlbars to confirm that all documentation contained within them has been migrated to the ember-glimmer package.

- [x] `packages/ember-htmlbars/lib/component.js`
- [x] `packages/ember-htmlbars/lib/helper.js`
- [x] `packages/ember-htmlbars/lib/helpers/-html-safe.js`
- [x] `packages/ember-htmlbars/lib/helpers/concat.js`
- [x] `packages/ember-htmlbars/lib/helpers/each-in.js`
- [x] `packages/ember-htmlbars/lib/helpers/each.js`
- [x] `packages/ember-htmlbars/lib/helpers/hash.js`
- [x] `packages/ember-htmlbars/lib/helpers/if_unless.js`
- [x] `packages/ember-htmlbars/lib/helpers/log.js`
- [x] `packages/ember-htmlbars/lib/helpers/query-params.js`
- [x] `packages/ember-htmlbars/lib/helpers/with.js`
- [x] `packages/ember-htmlbars/lib/helpers.js`
- [x] `packages/ember-htmlbars/lib/hooks/bind-local.js`
- [x] `packages/ember-htmlbars/lib/hooks/bind-self.js`
- [x] `packages/ember-htmlbars/lib/hooks/bind-shadow-scope.js`
- [x] `packages/ember-htmlbars/lib/hooks/classify.js`
- [x] `packages/ember-htmlbars/lib/hooks/cleanup-render-node.js`
- [x] `packages/ember-htmlbars/lib/hooks/concat.js`
- [x] `packages/ember-htmlbars/lib/hooks/destroy-render-node.js`
- [x] `packages/ember-htmlbars/lib/hooks/element.js`
- [x] `packages/ember-htmlbars/lib/hooks/get-child.js`
- [x] `packages/ember-htmlbars/lib/hooks/get-root.js`
- [x] `packages/ember-htmlbars/lib/hooks/get-value.js`
- [x] `packages/ember-htmlbars/lib/hooks/link-render-node.js`
- [x] `packages/ember-htmlbars/lib/hooks/subexpr.js`
- [x] `packages/ember-htmlbars/lib/hooks/update-self.js`
[...]

PR #14193, opened 2016-09-02
[DOC Beta] add API docs for component.js

toddjordan:
part of #14166 

moving documentation back into component.js

Couple of things of note: 
- lifecycle rule functions no longer defined on the Component prototype, but we still will call them if a child component implements them, so I went ahead and kept the documentation for them.  Lmk if I don't need to.  
- new component has more mixins.  I added these to `@uses` and made visible, this allows it to process inherited functions.
- there are a few new/changed public methods: `nearestChildOf` and `nearestInstanceOf` gone, but there now is a `nearestTypeOf`.  Also `isVisible` and `instrumentDisplay` gone as properties.


toddjordan:
cc @rwjblue

packages/ember-glimmer/lib/component.js
```diff
@@ -104,9 +199,282 @@ const Component = CoreView.extend(
...
+      {{yield}}
+      {{#if (hasBlock "inverse")}}
+        {{yield to="inverse"}}
+      {{else}}
+        How are you?
+      {{/if}}
+
+      ```
+      @public
+      @property hasBlock
```
rwjblue: We should kill this from the `Ember.Component` documentation, it is very confusing (as folks have reported before).  We need to document `hasBlock` / `has-block` like it was a template helper (like `{{each`, `{{with`, etc).

packages/ember-glimmer/lib/component.js
```diff
@@ -24,6 +24,101 @@ export const IS_DISPATCHING_ATTRS = symbol('IS_DISPATCHING_ATTRS');
 export const HAS_BLOCK = symbol('HAS_BLOCK');
 export const BOUNDS = symbol('BOUNDS');
 
+
+/**
+@module ember
+@submodule ember-views
```
rwjblue: Lets change this to `ember-glimmer`

packages/ember-glimmer/lib/component.js
```diff
@@ -104,9 +199,282 @@ const Component = CoreView.extend(
...
+      {{/foo-bar}}
+      {{! templates/components/foo-bar.hbs }}
+      {{#if hasBlockParams}}
+        This will be printed because a block was provided
+        {{yield this}}
+      {{/if}}
+      ```
+
+      @public
+      @property hasBlockParams
```
rwjblue: Same as `hasBlock`

toddjordan:
Should I remove the mixins that are private from `@uses`?  Upside is it would remove the private mixins from the api doc, downside it would remove some documented private methods in `Ember.Component`.

packages/ember-views/lib/mixins/action_support.js
```diff
@@ -1,3 +1,7 @@
+/**
+ @module ember
+ @submodule ember-views
```
rwjblue: ember-glimmer

rwjblue:
> Should I remove the mixins that are private from @uses?

Their class level docs should be marked as `@private` so that they don't show up in the left hand sidebar on emberjs.com/api/.

toddjordan:
well, I remember doing that feature and then removing it (b/c of the other classes that reference it).  It was too come back when we were more fine grained with it, but at the time when this drops the classes will show up...

rwjblue:
Hmm. I am definitely not in favor of these mixins showing up in the sidebar.  Lets leave them off for now then...

packages/ember-glimmer/lib/component.js
```diff
@@ -104,9 +200,282 @@ const Component = CoreView.extend(
...
+
+    The parameters can then be referred to by enumerating over the list:
+
+    ```hbs
+    {{#each attrs.names as |name|}}{{name}}{{/each}}
+    ```
+
+    @static
+    @public
+    @property positionalParams
```
toddjordan: Should we remove `positionalParams` for the same reason as hasBlock/hasBlockParams?

rwjblue: No, because `positionalParams` is something you have to use in the JS side.

toddjordan:
ok, addressed comments.  Did you see the third bullet above about missing public methods?  Is this ok?

rwjblue:
Yes, thank you.  I'm 👍 on this.

rwjblue:
Can you move https://github.com/emberjs/ember.js/blob/6c8abeb991a2355ff25ee7abe493aa68c4ced9eb/packages/ember-views/lib/mixins/aria_role_support.js#L14-L27 here and delete that file?  Also, we do not need the property to be initialized to `null` like that, we just need the docs here...

toddjordan:
done

rwjblue:
Looks like `ember-glimmer/component` is still importing `ember-views/mixins/aria_role_support` (causing errors).

toddjordan:
ah shoot.  fixing

rwjblue:
Thank you!

PR #14197, opened 2016-09-02
Rollup router, rsvp, route-recognizer

krisselden:


packages/ember-routing/lib/system/router.js
```diff
@@ -31,7 +31,6 @@ import dictionary from 'ember-metal/dictionary';
 */
 
 import Router from 'router';
-import 'router/transition';
```
rwjblue: We were just doing this for side effects?

rwjblue:
👍 - LGTM

PR #14200, opened 2016-09-03
[DOC Beta]add docs to helper files for glimmer 2

toddjordan:
Part of #14166 

Stab a converting htmlbar api documentation for glimmer 2

### DIFF
--- a/packages/ember-glimmer/lib/helpers/component.js
+++ b/packages/ember-glimmer/lib/helpers/component.js
@@ -1,3 +1,7 @@
+/**
+  @module ember
+  @submodule ember-glimmer
+*/
 import { assign } from 'ember-utils';
 import { CachedReference } from '../utils/references';
 import {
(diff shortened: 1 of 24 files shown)


packages/ember-glimmer/lib/helpers/component.js
```diff
@@ -1,9 +1,83 @@
+/**
+  @module ember
+  @submodule ember-glimmer
+  @public
```
rwjblue: I don't think `@public` belongs in this one

toddjordan:
Unsure where to put docs from the following files (no equivalent in the ember-glimmer package).  Could use some guidance on where to put the docs for these.

* lib/helpers/with.js
* lib/helpers/debugger.js
* lib/keywords/partial.js
* lib/streams/streams.js (subscribe function is public)

packages/ember-glimmer/lib/index.js
```diff
@@ -1,3 +1,85 @@
+/**
+
+ [Glimmer](https://github.com/tildeio/glimmer) is a [Handlebars](http://handlebarsjs.com/)
```
toddjordan: Updated the documentation to point at glimmer instead of htmlbars.  Combined the ember-htmlbars and ember-templates documentation into just ember-glimmer.  Need some confirmation that this is what is desired

rwjblue:
@toddjordan - I guess we can add the docs inline in `ember-glimmer/index`?

toddjordan:
ok, can do that

packages/ember-glimmer/lib/index.js
```diff
@@ -1,3 +1,216 @@
...
+  name.
+
+  @method partial
+  @for Ember.Templates.helpers
+  @param {String} partialName The name of the template to render minus the leading underscore.
+  @public
+*/
+
+/**
+ A method of subscribing to a stream which is safe for use with a non-stream
```
toddjordan: not sure that Ember.stream.subscribe even exists, but it was public. Any thoughts?

rwjblue: No, it was only public behind a feature flag (was never exposed without flag enabled, and the flag was never enabled).  ✂️ all references to "streams"

toddjordan: ✂️ 'd

toddjordan:
@rwjblue I believe a 2.8 docs release is happening soon (today?)  Do you think this is ready?

rwjblue:
Sorry about this needing a rebase, I missed the last round of fixes before I merged #14231.

toddjordan:
that's ok.  I'll do that now

rwjblue:
Thank you!

PR #14201, opened 2016-09-03
[CLEANUP Canary] Drop ViewTargetActionSupport

mixonic:
Drop this private and no longer useful mixin. Views are deprecated and removed from Ember.

The only usage I could find in the wild (besides with `View`) is here: https://github.com/crossroads/shared.goodcity/blob/637a3e1eb7dd5d712c97411c0abccc1445254521/app/components/online-button.js#L11 I'd appreciate eyes on that but I'm pretty sure it isn't doing anything.


rwjblue:
👍

PR #14204, opened 2016-09-03
[WIP] Expose meta on static glimmer templates

asakusuma:
This change requires the companion glimmer PR: https://github.com/tildeio/glimmer/pull/298. Getting this up now so folks can visualize the need in glimmer.

cc @krisselden @rwjblue

### DIFF
--- a/packages/ember-glimmer/lib/template.js
+++ b/packages/ember-glimmer/lib/template.js
@@ -36,14 +36,16 @@ class Wrapper {
 
 let templateId = 0;
 export default function template(json) {
+  let { meta, template } = JSON.parse(json);
   let id = ++templateId;
   return {
     id,
+    meta,
     create(options) {
       let env = options.env;
       let owner = options[OWNER];
 
-      return new Wrapper(id, env, owner, JSON.parse(json));
+      return new Wrapper(id, env, owner, JSON.parse(template));
     }
   };
 }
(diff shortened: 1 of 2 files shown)


packages/ember-glimmer/lib/template.js
```diff
@@ -36,14 +36,16 @@ class Wrapper {
 
 let templateId = 0;
 export default function template(json) {
+  let { meta, template } = JSON.parse(json);
```
rwjblue: We wanted to avoid parsing the JSON at eval time, which this undoes...

Why do we want the meta on the template factories in addition to the instances?

rwjblue:
One concern I have here, is that we are conflating the factory and instance information in `template.meta`.  The changes here to add a `.meta` property to the "factory" seem ok (other than forcing us to eagerly parse the JSON string), but I worry that later when we call `.create` and instantiate a `Template` instance that the object stashed on the template factory will be shared on the instances.  If that happens, then we would be adding the instances `owner` to our factories `meta` (which seems very bad 😝).  

Some solutions:

* Use `JSON.stringify`/`JSON.parse` in the `Wrapper` class.
* Use `let meta = Object.create(factoryMeta);` in the `Wrapper` class.
* Use `let meta = assign({}, factoryMeta)` (assuming the meta will be shallow).

rwjblue:
After re-reading this and the Glimmer PR (like 5 times), I realized that the `meta` on the factory is completely different than the meta in the `Wrapper#spec.meta` because we are duplicating it in the output.

asakusuma:
@rwjblue I agree, but I think that test should go in glimmer, not ember?

chadhietala:
I believe this was addressed in #14232

PR #14213, opened 2016-09-04
[DOC canary] Drop "View" references from docs

mixonic:
Across the codebase we still reference `Ember.View` in docs, even in places unrelated to views like the object model documentation. Clean this stuff up.

### DIFF
--- a/packages/ember-runtime/lib/system/core_object.js
+++ b/packages/ember-runtime/lib/system/core_object.js
@@ -216,13 +216,13 @@ CoreObject.PrototypeMixin = Mixin.create({
     Example:
 
     ```javascript
-    App.Person = Ember.Object.extend({
-      init: function() {
-        alert('Name is ' + this.get('name'));
+    const Person = Ember.Object.extend({
+      init() {
+        alert(`Name is ${this.get('name')}`);
       }
     });
 
-    var steve = App.Person.create({
+    let steve = Person.create({
       name: "Steve"
     });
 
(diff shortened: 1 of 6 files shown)


packages/ember-runtime/lib/system/core_object.js
```diff
@@ -523,20 +523,20 @@ var ClassMixinProps = {
...
-    This defines a new subclass of Ember.Object: `App.Person`. It contains one method: `say()`.
+    This defines a new subclass of Ember.Object: `Person`. It contains one method: `say()`.
 
     You can also create a subclass from any existing class by calling its `extend()` method.
-    For example, you might want to create a subclass of Ember's built-in `Ember.View` class:
+    For example, you might want to create a subclass of Ember's built-in `Ember.Component` class:
 
     ```javascript
-    App.PersonView = Ember.View.extend({
+    App.PersonComponent = Ember.Component.extend({
```
rwjblue: I'd prefer to use `export default` here, but I assume you thought of that and discarded it when updating this.  Can you share your thoughts here?

rwjblue:
I had one general question, but this looks good to me.

packages/ember-runtime/lib/system/core_object.js
mixonic: I think `const PersonComponent =` seems good as a compromise. I agree getting away from `App.` is good.

I don't think we've embraced the filename-comment-plus-default-export style of code blocks in the API docs thus far.

rwjblue: Sounds good, thanks for walking me through it.

mixonic: yop, doing some other cleanup in the core object file along the same lines

mixonic:
Most controversial is that I've changed the concatenatedProperties and mergedProperties docs to be general, not based on component and route base classes.

PR #14196, opened 2016-09-02
Add failing test around moduleName mutation

asakusuma:
Failing test that addresses https://github.com/emberjs/ember.js/issues/14192

Beta version of https://github.com/emberjs/ember.js/pull/14194

cc @krisselden @rwjblue @trentmwillis

PR #14219, opened 2016-09-06
Fix meta mutation #14192

### LINKED ISSUE #14192: Component rendering mutates module meta
We're attempting to upgrade our application to 2.7.2 and ran into an odd issue. Brief backstory, we rely on the `meta.moduleName` property from the compiled HTMLBars template to do some of our lookups for localization.

While attempting the upgrade, there are some tests that we have that are now failing after being run in a certain order. After a lot of debugging, the root issue is that the `meta.moduleName` property is being mutated which causes the value to be incorrect when we do our lookup for the translation.

The offending line of code seems to be this: https://github.com/emberjs/ember.js/blob/v2.7.2/packages/ember-htmlbars/lib/hooks/component.js#L62

I haven't narrowed down the exact circumstances under which this occurs, but in some cases it appears `env.meta` is the same reference as the `meta` property from a templates AMD module. Thus, if `templateMeta.moduleName` is used, the module is then mutated permanently.

cc @asakusuma @zackthehuman

krisselden:
Fixes #14192.
Closes https://github.com/emberjs/ember.js/pull/14196.

### DIFF
--- a/packages/ember-glimmer/tests/integration/components/local-lookup-test.js
+++ b/packages/ember-glimmer/tests/integration/components/local-lookup-test.js
@@ -49,6 +49,20 @@ moduleFor('Components test: local lookup', class extends RenderingTest {
     this.assertText('Nested template says: Hi!', 'Re-render works');
   }
 
+  ['@test moduleName remains unchanged']() {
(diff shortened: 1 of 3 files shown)


rwjblue:
Pulled into release branch for release

PR #14221, opened 2016-09-07
Avoid using `let` in `ember-cli-build.js`

chancancode:
"SyntaxError: Could not require 'ember-cli-build.js': Block-scoped declarations (let, const, function, class) not yet supported outside strict mode"

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -50,7 +50,7 @@ function dag() {
 }
 
 function backburner() {
-  let dist = path.dirname(require.resolve('backburner.js'));
+  var dist = path.dirname(require.resolve('backburner.js'));
   dist = path.join(dist, 'es6');
   return new Rollup(new Funnel(dist, {
     files: ['backburner.js']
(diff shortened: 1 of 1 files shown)


rwjblue:
Umm.  Just update your Node version. :trollface:

rwjblue:
But adding `'use strict';` to the top (as the message indicates) will fix this for Node 4.x also.

rwjblue:
Also, we plan on using `class` in the upcoming build refactor so....

chancancode:
I am on Node 5 :P But all of those other options sound fine to me too!

stefanpenner:
> Also, we plan on using class in the upcoming build refactor so....

Lets not do this, we plan to allow users to build canary via the build pipeline that requires node LTS compat with how we build ember.

stefanpenner:
this LGTM 👍

rwjblue:
> Lets not do this, we plan to allow users to build canary via the build pipeline that requires node LTS compat with how we build ember.

Meh.  It seems fine to me that using canary requires more current node. It is also not obvious to me that users will actually be building manually even for `beta` / `canary` channels.

PR #14223, opened 2016-09-07
[BUGFIX beta] Check state of router during transition

yawboakye:
Currently, during transitions, the state of the router is not taken into
consideration. This change makes sure that both `transitionTo()` and
`replaceWith()` examine the state of the router and uses that information
to choose the appropriate method.

If there has never been a transition since the app loaded, `replaceWith()`
will be used. If there's been a transition already `transitionTo()` will
be used instead. Regardless of which method the  developer used.

Special thanks to @alexspeller. In so many ways this work belongs to him.

### DIFF
--- a/packages/ember-routing/lib/system/route.js
+++ b/packages/ember-routing/lib/system/route.js
@@ -1067,7 +1067,10 @@ let Route = EmberObject.extend(ActionHandler, Evented, {
   */
   transitionTo(name, context) {
     let router = this.router;
-    return router.transitionTo(...prefixRouteNameArg(this, arguments));
+    if (router.router.oldState) {
+      return router.transitionTo(...prefixRouteNameArg(this, arguments));
+    }
+    return router.replaceWith(...prefixRouteNameArg(this, arguments));
   },
 
   /**
@@ -1151,6 +1154,9 @@ let Route = EmberObject.extend(ActionHandler, Evented, {
   */
   replaceWith() {
     let router = this.router;
+    if (router.router.oldState) {
+      return router.transitionTo(...prefixRouteNameArg(this, arguments));
+    }
     return router.replaceWith(...prefixRouteNameArg(this, arguments));
   },
 


packages/ember-routing/lib/system/route.js
```diff
@@ -1067,7 +1067,10 @@ let Route = EmberObject.extend(ActionHandler, Evented, {
   */
   transitionTo(name, context) {
     let router = this.router;
-    return router.transitionTo(...prefixRouteNameArg(this, arguments));
+    if (router.router.oldState) {
+      return router.transitionTo(...prefixRouteNameArg(this, arguments));
+    }
+    return router.replaceWith(...prefixRouteNameArg(this, arguments));
```
rwjblue: I would prefer the logic be swapped around here so that the "default" thing we do is `transitionTo` from `Route#transitionTo`.

rwjblue:
Thanks for working on this!

A few things that we need to do:

* Tests for the various states:
  * `transitionTo` during initial boot (likely from `route:index`)
  * `replaceWith` during initial boot (likely from `route:index`)
  * `transitionTo` outside of initial render, during another transition (from say `route:index` -> `route:posts` do a redirect to `route:posts/comments`)
  * `replaceWith` outside of initial render, during another transition (from say `route:index` -> `route:posts` do a redirect to `route:posts/comments`)
  * `transitionTo` outside of initial render, outside of an existing transition (from an action)
  * `replaceWith` outside of initial render, outside of an existing transition (from an action)
* Fix inline comment RE: `transitionTo`'s "default"

alexspeller:
There are a couple of issues with this:

1. You are having the same check in both methods – you are checking for `oldState` both times and calling `transitionTo` - the checks should be inversed.
  a. If you are calling transitionTo, and oldState is not present, then replaceWith must be called.
  b. If you are calling replaceWith, and oldState is present, then transitionTo must be called.

2. These checks must only apply in the validation phase of a transition. Currently they apply all the time, which means that you'll break replaceWith and transitionTo in use cases outside of the validation phase of a transition

alexspeller:
as an architectural issue, I wonder if these fixes shouldn't actually be in router.js rather than just for ember - AFAIK the bug is actually at the lower level in the router library

alexspeller:
PR filed to router https://github.com/tildeio/router.js/pull/197

rwjblue:
Will be fixed by https://github.com/tildeio/router.js/pull/197, closing this one...

PR #14227, opened 2016-09-07
Implement correctly ordered destruction hooks.

rwjblue:
/cc @krisselden

### DIFF
--- a/packages/ember-glimmer/lib/environment.js
+++ b/packages/ember-glimmer/lib/environment.js
@@ -69,6 +69,9 @@ export default class Environment extends GlimmerEnvironment {
     super(...arguments);
     this.owner = owner;
 
+    // can be removed once https://github.com/tildeio/glimmer/pull/305 lands
+    this.destroyedComponents = undefined;
+
     installPlatformSpecificProtocolForURL(this);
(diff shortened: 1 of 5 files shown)


rwjblue:
Discussed with @krisselden and while there is still work to be done in Glimmer to remove the need to override some of these things this is still a step forward and we can iterate on it...

packages/ember-glimmer/lib/environment.js
```diff
@@ -69,6 +69,9 @@ export default class Environment extends GlimmerEnvironment {
     super(...arguments);
     this.owner = owner;
 
+    // can be removed once https://github.com/tildeio/glimmer/pull/305 lands
+    this.destroyedComponents = undefined;
```
chancancode: I thought you indented to just use this.destructors since you merged that PR?

chancancode: ah we haven't merged yet. ignore me

rwjblue: Right, I'd rather have a small kludge here in Ember for beta.1 and fix upstream in that other PR for beta.2.

PR #14229, opened 2016-09-08
[BUGFIX beta] Fix issue with IE9 compatibility.

### LINKED ISSUE #14006: Internet Explorer IE9: _emberEnvironment.environment.global returns `undefined`
After upgrading one of our apps to 2.7, it failed to load in IE9. I reproduced the issue by installing a new sandbox app with v2.7.  The error is thrown in `supportsHashChange` with `global` being `undefined`.

```
function supportsHashChange(documentMode, global) {
    return 'onhashchange' in global && (documentMode === undefined || documentMode > 7);
  }
```

Reverting to 2.6 resolves the issue.

rwjblue:
A past refactor left `AutoLocation` in a broken state under IE9. This fixes that error and adds a (somewhat silly) test.

Fixes #14006.

### DIFF
--- a/packages/ember-routing/lib/location/auto_location.js
+++ b/packages/ember-routing/lib/location/auto_location.js
@@ -70,9 +70,9 @@ export default EmberObject.extend({
 
    @since 1.11
    @property global
-   @default environment.global
+   @default window
   */
-  global: environment.global,
+  global: environment.window,
 
   /**
     @private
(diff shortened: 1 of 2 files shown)


rwjblue:
I tested a simple `ember new foo` app with Ember version just before this change under IE9 the following error occurs:

```
SCRIPT5007: Invalid operand to 'in': Object expected 
```

After this change, the error does not occur...

packages/ember-routing/lib/location/auto_location.js
```diff
@@ -72,7 +72,7 @@ export default EmberObject.extend({
    @property global
    @default environment.global
   */
-  global: environment.global,
+  global: environment.window,
```
Turbo87: the default above should probably be updated too, right?

PR #14230, opened 2016-09-08
Add failing test for appendTo while rendering

mitchlloyd:
@rwjblue Here is a failing test for the behavior I'm seeing in https://github.com/mitchlloyd/ember-islands/pull/29 where `Renderer.commit()` is being called twice in a row and throwing an error.


rwjblue:
Thank you for the failing test! I will try to dig in tonight...

mitchlloyd:
Closed in favor of #14241.

PR #14232, opened 2016-09-08
Fix ssr action modifiers and template pre-compilation

chadhietala:


rwjblue:
Thanks for cleaning up that change @chadhietala.

LGTM

PR #14236, opened 2016-09-08
Don't destroy singleton controllers in {{render}}

chancancode:
If you didn't instantiate it, don't destroy it.

Thanks @fivetanley for reporting it!


rwjblue:
LGTM

Issue #14234, opened 2016-09-08
Engine routes deactivated after destruction


trentmwillis:
In 2.8.0, there was a change introduced to [the order of teardown when Engines are involved](https://github.com/emberjs/ember.js/commit/447df33c4db30c475a08415196dbb7012f08cc07). The change in question seems to have fixed another issue, but causes routes in Engines to be destroyed prior to their `deactivate` hook being run (during the Router's `reset`).

This means that if you do something such as `set` a property on the Route during `deactivate`, then errors will propagate in Acceptance tests.

Unsure what the proper fix is here. Seems like maybe only the views from the Engines should be torn down prior to resetting the Router.

(Here's [the prior implementation](https://github.com/dgeb/ember-engines/blob/v0.2/addon/-private/router-ext.js#L129) from the ember-engines addon.)

cc @asakusuma @krisselden

rwjblue:
@trentmwillis - Can you work up a failing test (best harness ATM for this is [here](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/tests/integration/application/engine-test.js))?

trentmwillis:
@rwjblue reproduction added in #14237.

trentmwillis:
@rwjblue / @asakusuma do you know if the original issue:

> Ensure that engines are destroyed _before_ top-level views so that top-level views are not re-instantiated during engine teardown.

Had tests to catch a regression? Because I looked into fixing this and just switching the order back seems to pass without failing any tests.

rwjblue:
@trentmwillis - Yeah, I think the `isDestroying` guard added in https://github.com/emberjs/ember.js/pull/14123 actually fixed the regression (just before the PR changing the order landed).

trentmwillis:
@rwjblue okay, so do you think switching the order back seems like the proper solution? If so, I'll update my PR with the failing test.

rwjblue:
@trentmwillis - Yeah, lets switch it back and confirm both your regression tests pass.

trentmwillis:
Fixed via #14237. Tested against the latest release build.

PR #14237, opened 2016-09-08
[BUGFIX release] Ensure Engine Routes are deactivated before destruction

trentmwillis:
Makes tests fail to reproduce #14234.

Causes some existing tests to fail, though I can add a totally separate tests for this if needed.

### DIFF
--- a/packages/ember-glimmer/tests/integration/application/engine-test.js
+++ b/packages/ember-glimmer/tests/integration/application/engine-test.js
@@ -227,4 +227,27 @@ moduleFor('Application test: engine rendering', class extends ApplicationTest {
       ], 'the expected hooks were fired');
     });
   }
+
+  ['@test deactivate should be called on Engine Routes before destruction'](assert) {
+    assert.expect(3);
+
+    this.setupAppAndRoutableEngine();
+
+    this.registerEngine('blog', Engine.extend({
+      init() {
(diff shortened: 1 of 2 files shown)


rwjblue:
Awesome, thank you!

>  I can add a totally separate tests for this if needed

That would be great if you don't mind. Iff nothing more it helps us ensure that our tests document the regression, and help ensure we don't accidentally "remove that errant set in deactivate" in the future.

trentmwillis:
@rwjblue updated. A bit more clear as to what is being tested now 😄

rwjblue:
Awesome! We should probably add some general purpose lifecycle ordering tests too at some point, but this is perfect to help track down the regression (and prevent it happening again in the future).

trentmwillis:
@rwjblue updated with a fix. Tests passing except Sauce, which seems to have failed to connect properly.

What should I tag this as? `BUGFIX release`?

rwjblue:
@trentmwillis - Yep, `[BUGFIX release]` sounds correct to me.

trentmwillis:
Tests are passing. Pushing the amended commit message.

rwjblue:
This has been cherry picked into beta and release branches.  Please test those channels (once Travis has pushed the builds up) to confirm before final release...

PR #14238, opened 2016-09-08
Add non-interactive {{action}} test

chadhietala:
We need to make sure `indexOf` doesn't blow up on `destroy` if actions were never registered.

### DIFF
--- a/packages/ember-glimmer/lib/modifiers/action.js
+++ b/packages/ember-glimmer/lib/modifiers/action.js
@@ -57,6 +57,10 @@ export let ActionHelper = {
     let { actionId } = actionState;
     let actions = ActionManager.registeredActions[actionId];
 
+    if (!actions) {
+      return;
+    }
+
     let index = actions.indexOf(actionState);
 
     if (index !== -1) {
--- a/packages/ember-glimmer/tests/integration/application/actions-test.js
+++ b/packages/ember-glimmer/tests/integration/application/actions-test.js
@@ -1,5 +1,6 @@
 import { Controller } from 'ember-runtime';
-import { moduleFor, ApplicationTest } from '../../utils/test-case';
+import { moduleFor, ApplicationTest, RenderingTest } from '../../utils/test-case';
+import { Component } from '../../utils/helpers';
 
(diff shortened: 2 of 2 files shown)


packages/ember-glimmer/lib/modifiers/action.js
```diff
@@ -55,7 +55,7 @@ export let ActionHelper = {
 
   unregisterAction(actionState) {
     let { actionId } = actionState;
-    let actions = ActionManager.registeredActions[actionId];
+    let actions = ActionManager.registeredActions[actionId] || [];
```
rwjblue: I think it may be better to exit early here since there is no point checking for an index in an empty array.

packages/ember-glimmer/tests/integration/application/actions-test.js
```diff
@@ -48,3 +49,25 @@ moduleFor('Application test: actions', class extends ApplicationTest {
...
+          fire() {}
+        }
+      }),
+      template: `<button {{action 'fire'}}>Fire!</button>`
+    });
+
+    this.render('{{foo-bar}}');
+    let btnHTML = this.firstChild.innerHTML;
+
+    assert.equal(btnHTML, '<button>Fire!</button>');
```
rwjblue: Might need to use one of the assert HTML helper methods to avoid issues across various browsers? (Just guessing though).

packages/ember-glimmer/tests/integration/application/actions-test.js
```diff
@@ -48,3 +49,25 @@ moduleFor('Application test: actions', class extends ApplicationTest {
...
+moduleFor('Rendering test: non-interactive actions', class extends RenderingTest {
+  getBootOptions() {
+    return { isInteractive: false };
+  }
+
+  [`@test doesn't attatch actions`](assert) {
+    this.registerComponent('foo-bar', {
+      ComponentClass: Component.extend({
+        actions: {
+          fire() {}
```
rwjblue: Can you add an assert.ok(false) and try to trigger a click on the button? Just to doubly ensure that nothing happens...

PR #14241, opened 2016-09-09
[BUGFIX beta] Ensure roots added during rendering work properly.

rwjblue:
When a new root is added during the rendering process (i.e. from `didInsertElement` of one component you `.appendTo` another one), we need to:

1) Ensure that we do not initiate a new transaction (`env.begin()` / `env.commit()`)
2) Ensure that we trigger _renderRoots again if roots were added. This ensures the new roots are properly rendered.

Addresses issue reported in https://github.com/mitchlloyd/ember-islands/pull/29#issuecomment-245484407.
Closes #14230
Fixes https://github.com/emberjs/ember.js/issues/14287

### DIFF
--- a/packages/ember-glimmer/lib/renderer.js
+++ b/packages/ember-glimmer/lib/renderer.js
@@ -61,6 +61,8 @@ class RootState {
     this.root = root;
     this.result = undefined;
     this.shouldReflush = false;
+    this.destroyed = false;
+    this._removing = false;
 
     let options = this.options = {
       alwaysRevalidate: false
@@ -82,6 +84,8 @@ class RootState {
 
   destroy() {
     let { result, env } = this;
+
+    this.destroyed = true;
 
     this.env = null;
     this.root = null;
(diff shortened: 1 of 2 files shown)


chancancode:
Could we get away with deferring it to the next runloop?

rwjblue:
@chancancode - Seems reasonable to me, I'll play with that and see if it ends up nicer...

mitchlloyd:
Confirmed that this fixes the issue I was seeing.

rwjblue:
@chancancode - I did a quick test to see if I could `run.schedule` in `renderRoot`, but the various ways I tried still ended with the reported error.  I'm happy to keep digging if this seems unsatisfactory though...

krisselden:
@rwjblue tomorrow could you walk me through this? flags like this generally don't age well, and I'd like to understand how it is being reentered.

rwjblue:
@krisselden - Yes, absolutely. I'm happy to work out a better way also, so it would be great to talk it through with you.

rwjblue:
Updates:

* Address poor factoring concerns (thanks to @krisselden for pairing with me on a better solution)
* Added tests to confirm `renderToElement`, `append`, and `appendTo` all allow appending and deleting roots during a root iteration
* Update inline comments to give future @rwjblue a fighting chance

PR #14244, opened 2016-09-09
[BUGFIX beta] Ensure params and hash are frozen in debug builds.

### LINKED ISSUE #14189: Helpers mutating `hash` may error if no hash was provided.
One optimization that the Glimmer engine does is to avoid creating objects when it does not need to. An example of this can be seen when evaluating helper and/or component invocations where there is no hash arguments.  Such as this:

```hbs
{{page-title "My App"}}
```

In order to avoid creating a wasted hash argument here, Glimmer will use a custom `EVALUATED_EMPTY_NAMED_ARGS` which always returns a shared constant named `EMPTY_DICT`.  In order to prevent that shared constant from being mutated, the engine uses `Object.freeze(EMPTY_DICT)`.

All of these assumptions/steps are completely reasonable.  Unfortunately, where this falls down is when a helper (such as [ember-page-title](https://github.com/tim-evans/ember-page-title/blob/3d9584736f3c04f3d327faf6ae07121b5fd8ae3b/addon/helpers/page-title.js#L21)) mutate that provided hash argument.  If you do this, the following error is triggered:

```
TypeError: Can't add property foo, object is not extensible
    at https://output.jsbin.com/solokun:55:12
    at Function.create (https://s3.amazonaws.com/builds.emberjs.com/canary/ember.debug.js:12837:22)
    at https://s3.amazonaws.com/builds.emberjs.com/canary/ember.debug.js:8301:69
    at CompiledHelper.evaluate (https://s3.amazonaws.com/builds.emberjs.com/canary/ember.debug.js:42855:20)
    at VM.evaluateOperand (https://s3.amazonaws.com/builds.emberjs.com/canary/ember.debug.js:50094:40)
    at PutValueOpcode.evaluate (https://s3.amazonaws.com/builds.emberjs.com/canary/ember.debug.js:45494:16)
    at VM.execute (https://s3.amazonaws.com/builds.emberjs.com/canary/ember.debug.js:50066:28)
    at Template.render (https://s3.amazonaws.com/builds.emberjs.com/canary/ember.debug.js:49616:23)
    at RootState.render (https://s3.amazonaws.com/builds.emberjs.com/canary/ember.debug.js:9969:61)
    at runInTransaction (https://s3.amazonaws.com/builds.emberjs.com/canary/ember.debug.js:20692:28)
```

[JSBin Demo Here](https://jsbin.com/solokun)

---

We need to decide if we must "fix" this by providing a custom hash object to each helper invocation (less than ideal because we would much prefer to benefit from doing less work).

rwjblue:
This ensures that `params` and `hash` are both ran through `Object.freeze` in debug builds, but tries to leave the codepaths as simple as possible so that when the `runInDebug`'s are removed the production code can be optimized the same way as prior to this change.

Closes #14189.

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -186,8 +186,8 @@ function babelConfigFor(environment) {
   if (isProduction) {
     includeDevHelpers = false;
     plugins.push(filterImports({
-      'ember-metal/debug': ['assert', 'debug', 'deprecate', 'info', 'runInDebug', 'warn', 'debugSeal'],
-      'ember-metal': ['assert', 'debug', 'deprecate', 'info', 'runInDebug', 'warn', 'debugSeal']
+      'ember-metal/debug': ['assert', 'debug', 'deprecate', 'info', 'runInDebug', 'warn', 'debugSeal', 'debugFreeze'],
+      'ember-metal': ['assert', 'debug', 'deprecate', 'info', 'runInDebug', 'warn', 'debugSeal', 'debugFreeze']
     }));
   }
 
--- a/packages/ember-debug/lib/index.js
+++ b/packages/ember-debug/lib/index.js
@@ -153,6 +153,10 @@ setDebugFunction('debugSeal', function debugSeal(obj) {
   Object.seal(obj);
 });
 
+setDebugFunction('debugFreeze', function debugFreeze(obj) {
+  Object.freeze(obj);
+});
+
 setDebugFunction('deprecate', _deprecate);
 
 setDebugFunction('warn', _warn);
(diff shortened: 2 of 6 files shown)


mmun:
This is nice! Do you have any idea if there's a performance penalty?

rwjblue:
Yes, adding `Object.freeze` will definitely have some cost. There should be basically no cost when this is in production builds, since the entire `runInDebug` statement is removed. I'm hoping the added "correctness" and the fact that prod builds will be the same will make up from any performance impact.

I'm also open to reverting if it turns out that this makes even debug builds too slow, but I generally think that will be unlikely...

PR #14245, opened 2016-09-09
[BUGFIX beta] Lookup partials from the correct owner

chadhietala:
Fixes issues where `{{partial}}`s were not looked up correctly from the owner. This commit fixes that by looking up the owner based on what owner initially invoked it.


rwjblue:
LGTM

rwjblue:
Actually, can you update commit message prefix to be [BUGFIX beta]? This should get into 2.9...

Issue #14097, opened 2016-08-19
Deprecate Outlet Orphaning


chadhietala:
This is a relic that did get cut in 2.0.0 and we should part ways with. It is used when you have a `{{render}}` helper in the route template and you render into it from the route. As you render into it you will come across an `{{outlet}}` that is orphaned. You now need to reason about what should go into that outlet.

The [path that will be deprecated](https://github.com/emberjs/ember.js/blob/205e5acfaf421c600769dfc428e40cb857ecacd5/packages/ember-routing/lib/system/router.js#L1296-L1303).

/cc @ef4

chancancode:
We can probably just deprecate `{{render}}`?

ef4:
Yes, deprecating {{render}} solves this.

There's a related behavior that makes sense to deprecate with it, which is
calling `Route#render` any time other than within the `renderTemplate`
hook. Last time this came up the objection was that we need to show a
better alternative for sidebar and modal rendering, which now exists in
ember-elsewhere.

With that, we can cut a bunch of complexity out.

PR #14246, opened 2016-09-09
[WIP] Deprecate `{{render}}` and outlet orphaning

chadhietala:
Want to get the ball rolling on #14097 so it's not forgotten.


josemarluedke:
This can be closed since #14441 has been merged.

PR #14252, opened 2016-09-09
[Bugfix Beta] Don't delete moduleName from options

chadhietala:
This is used by some AST plugins and was previously passed.

### DIFF
--- a/packages/ember-template-compiler/lib/system/compile-options.js
+++ b/packages/ember-template-compiler/lib/system/compile-options.js
@@ -10,8 +10,6 @@ export default function compileOptions(_options) {
   if (options.moduleName) {
     let meta = options.meta;
     meta.moduleName = options.moduleName;
-
-    delete options.moduleName;
   }
 
   if (!options.plugins) {


rwjblue:
Seems ok. We should probably make it a getter with a deprecation I guess.

chadhietala:
We use this to take ` {{t "i18n_hello_world"}}` and transform it into `{{t "i18n_hello_world" "modules/foo/bar/baz-bar.hbs"}}`. So we need this to be passed to AST transforms to carry forward the module name which is used as a key in a translations map.

rwjblue:
Sure, it was pass through to the transform before this PR. It's just moved to options.meta, I'm saying we should deprecate using options.moduleName in an AST and instead use options.meta.moduleName.

chadhietala:
Ah I see.

asakusuma:
@rwjblue so another issue is that deleting makes this operation not idempotent. If your build is dumb and processes the same file twice, moduleName won't be there the 2nd time.

rwjblue:
@asakusuma - Sure, but it would be in the new location and not need to be moved.

rwjblue:
Actually, we do an `assign({}, passedInOptions)` so we are not mutating the object passed in at all.

PR #14253, opened 2016-09-10
Prevent duplicated `Ember.meta` invocations.

rwjblue:
`mergeMixins` was passed in the meta for this object, there is no need to look it up again for _every_ mixin.


chadhietala:
LGTM. Curious as to what the win is here.

rwjblue:
@chadhietala - Well, its just doing less work since we already have the info around. 

This change results in around 100,000 fewer calls to `Ember.meta` during the Ember test suite (goes from 744k to ~ 623k), but I haven't really tested with the counters in an app. No code is faster than no code though right? 😝

PR #14259, opened 2016-09-10
[BUGFIX beta] Do not call set on PromiseProxy if destroyed.

### LINKED ISSUE #14250: Prevent PromiseProxyMixin from calling set on destroyed objects
I am submitting this issue just to get feedback on if this issue is worth fixing. I already have a solution working in my production application and am happy to submit a pull-request.

Here's an ember-twiddle showing the error I get: https://ember-twiddle.com/cdc0cff705e96ca7d136b55e50ee9f95?openFiles=routes.application.js%2C

The example in ember-twiddle is pretty contrived. I generally get this error in my tests.

My solution involves simply checking `proxy.isDestroyed`, as that [seems](http://discuss.emberjs.com/t/how-to-do-you-check-for-isdestroyed/9621) to be the [recommended](http://stackoverflow.com/questions/25141822/uncaught-error-assertion-failed-calling-set-on-destroyed-object) [solution](https://github.com/emberjs/ember.js/issues/3780#issuecomment-63470006).

Is there a reason I should not approach this direction?

Serabe:
Given an object including `PromiseProxy`, if such object is destroyed
before de promise is fullfill, an error is raised. This error is not
obvious from a user point of view.

This PR avoids setting anything if the proxy has been destroyed while
keeping the promise chain intact.

Fixes #14250

### DIFF
--- a/packages/ember-runtime/lib/mixins/promise_proxy.js
+++ b/packages/ember-runtime/lib/mixins/promise_proxy.js
@@ -19,16 +19,20 @@ function tap(proxy, promise) {
   });
 
   return promise.then(value => {
-    setProperties(proxy, {
-      content: value,
-      isFulfilled: true
-    });
+    if (!proxy.isDestroyed && !proxy.isDestroying) {
+      setProperties(proxy, {
+        content: value,
+        isFulfilled: true
+      });
+    }
     return value;
   }, reason => {
-    setProperties(proxy, {
-      reason: reason,
-      isRejected: true
-    });
+    if (!proxy.isDestroyed && !proxy.isDestroying) {
+      setProperties(proxy, {
+        reason: reason,
+        isRejected: true
+      });
+    }
     throw reason;
   }, 'Ember: PromiseProxy');
 }
--- a/packages/ember-runtime/tests/mixins/promise_proxy_test.js
+++ b/packages/ember-runtime/tests/mixins/promise_proxy_test.js
@@ -256,3 +256,83 @@ QUnit.test('should have reason when isRejected is set', function() {
     equal(e, error);
   }
 });
+
+QUnit.test('should not error if promise is resolved after proxy has been destroyed', function() {
+  let deferred = EmberRSVP.defer();
+
+  let proxy = ObjectPromiseProxy.create({
+    promise: deferred.promise
+  });
+
+  proxy.then(() => {}, () => {});
+
+  run(proxy, 'destroy');
+
+  run(deferred, 'resolve', true);
+
+  ok(true, 'resolving the promise after the proxy has been destroyed does not raise an error');
+});
+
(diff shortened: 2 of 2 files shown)


Serabe:
Given that `release` (2.8) is going to be the new LTS, it might be interesting to be a `[BUGFIX release]` though I did not want to do so without checking first.

packages/ember-runtime/lib/mixins/promise_proxy.js
```diff
@@ -19,16 +19,20 @@ function tap(proxy, promise) {
   });
 
   return promise.then(value => {
-    setProperties(proxy, {
-      content: value,
-      isFulfilled: true
-    });
+    if (!get(proxy, 'isDestroyed') && !get(proxy, 'isDestroying')) {
```
rwjblue: No need to use get for these, and maybe we can just use an early return?

```
if (this.isDestroyed || this.isDestroying) {
  return;
}
```

rwjblue:
I am not really sure this is a regression (AFAICT all versions of PromiseProxyMixin have had this bug). If it isn't a regression from a prior release, I'm inclined to leave it `BUGFIX beta`.

packages/ember-runtime/lib/mixins/promise_proxy.js
Serabe: Fixing the `get` issue. Avoid the early return, not a big fan since that would mean repeating code (`return value` or `throw reason`). As an example, the second parameter for `then` would be:

```js
  (reason) => {
    if (!proxy.isDestroyed && !proxy.isDestroying) {
      setProperties(proxy, {
        reason: reason,
        isRejected: true
      });
    }
    throw reason;
  }
```

only addressing the `get` (to be pushed as soon as I check the tests) and using an early return would be:

```js
  (value) => {
    if (proxy.isDestroyed || proxy.isDestroying) {
      throw reason;
    }
    setProperties(proxy, {
      reason: reason,
      isRejected: true
    });
    throw reason;
  }
```

IMHO, it is much harder to read having a method with two paths throwing the same error but having a conditional about one of them. I can change it too, just wanted to make sure is not a simple early return.

Serabe:
Leaving as `[BUGFIX beta]` and removing the use of `get` for the `isDestroying`/`isDestroyed` properties.

Can you please check [my inline comment](https://github.com/emberjs/ember.js/pull/14259#discussion_r78285059) about the early return and confirm it is the way to go?

Thank you for reviewing this!

Serabe:
@rwjblue let me know when you have some time to check my previous comment to know if you still want an early return. My comment is hidden here:

image

Thank you!

Serabe:
Rebased and pushed to check let CI do its work.

rwjblue:
Thanks @Serabe!

Issue #14264, opened 2016-09-11
Unexpected "Attempting to define property on object that is not extensible." on 2.9 beta


kellyselden:
I have a very simple [addon](https://github.com/kellyselden/ember-array-helper). It's latest master commit is from March. I deleted Travis cache and reran the master build. Here are the results:
https://travis-ci.org/kellyselden/ember-array-helper/builds/115857118
release working:
https://travis-ci.org/kellyselden/ember-array-helper/jobs/115857121
beta failing:
https://travis-ci.org/kellyselden/ember-array-helper/jobs/115857123

Here it is reproduced on twiddle:
working on release:
https://ember-twiddle.com/eb15180109e6467f27f88743245cf788/ab9ce64a8625d645b144795f930e10432e0c1816
broken on beta:
https://ember-twiddle.com/eb15180109e6467f27f88743245cf788/5d79be64c7da893b3fbce85e229fbed200b04963

Here is the relevant stack trace when testing locally:
![capture](https://cloud.githubusercontent.com/assets/602423/18414577/f89feea4-7783-11e6-87c9-f2cb4bb05c78.PNG)

It fails on this line https://github.com/emberjs/ember.js/blob/f97598ea6916dfa9094569715933a5e9c060a34d/packages/ember-metal/lib/meta.js#L365

Here is how I got it working:

``` js
import Ember from 'ember';

export function array(params/*, hash*/) {
  // this is the original failure
  // return params;

  // this also fails for the same reason (unexpectedly according to docs)
  // return Ember.A(params);

  // this makes it work
  let array = Ember.A();
  array.pushObjects(params);
  return array;
}

export default Ember.Helper.helper(array);
```

But there may be an unexpected change in beta that is causing the prototype extensions to act differently. This also only seems to be an issue with helpers.

rwjblue:
This is a consequence of https://github.com/emberjs/ember.js/pull/14244.  When the params / hash are frozen, we cannot put a meta on the object (which is done if it is used in the template).

krisselden:
You can't mutate the params array and we seal it in the debug build now.  You can just use slice. 

We made it fail fast when we realized people were doing this,  it prevents us from having to allocate tons of empty arrays.

rwjblue:
@krisselden - The specific thing being done here is Ember mutating the array (by trying to put meta on it).

krisselden:
Robert we should not try to observe a frozen object it isn't going to change

rwjblue:
[...]

PR #14265, opened 2016-09-11
[Bugfix Beta] Freeze attrs in debug

chadhietala:
@rwjblue @krisselden I believe there were talks about freezing these as well. I think we want to make `attrs` pay as you go as well.

### DIFF
--- a/packages/ember-glimmer/lib/component.js
+++ b/packages/ember-glimmer/lib/component.js
@@ -192,7 +192,11 @@ const Component = CoreView.extend(
     },
 
     getAttr(key) {
-      // TODO Intimate API should be deprecated
+      deprecate(`getAttr is deprecated please use "this.get(${key})"`, false, {
+        id: 'ember-views.component.getAttr',
+        until: '3.0.0',
+        url: 'http://emberjs.com/deprecations/v2.x/#toc_ember-component-get-attr'
+      });
       return this.get(key);
     },
 
--- a/packages/ember-glimmer/lib/syntax/curly-component.js
+++ b/packages/ember-glimmer/lib/syntax/curly-component.js
@@ -214,6 +214,10 @@ class CurlyComponentManager {
 
     props._targetObject = callerSelfRef.value();
 
+    runInDebug(() => {
+      attrs = Object.freeze(attrs);
+    });
+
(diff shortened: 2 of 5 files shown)


chadhietala:
I actually don't know if we can do this. If you use `attrs` in the backing class for something like dependent keys on a CP we're going to install `meta` and thats going to result in 

```
Cannot define property:__ember_meta__, object is not extensible
```

Specifically I ran into this with [Link-to's use of `attrs`](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/components/link-to.js#L588-L593) :trollface:.

rwjblue:
Yep, that is exactly what https://github.com/emberjs/ember.js/issues/14264 is about.

rwjblue:
Discussed in Slack, I think changing these lines:

* https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/utils/references.js#L145
* https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/utils/references.js#L195

To use this as the guard:

```js
if (isFeatureEnabled('mandatory-setter') && !Object.isFrozen(parentValue)) {
}
```

Should fix?

krisselden:
It should fix it,  to be honest it is annoying we haven't done this for frozen objects already

rwjblue:
@krisselden - Agreed.  I think we also need to do something in the CP system (since `{{link-to}}` is using `attrs.foo` as a dependent key).

krisselden:
That still goes through Ember.watch

chadhietala:
So guarding [here](https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/lib/watching.js#L37-L43) causes a bunch of other things to break, about 1000 tests.

krisselden:
@chadhietala that is because there are assumptions about state that it breaks that would need to be updated.

addObserver is addListener + watch.   cps use watch directly.  there maybe some other cases with isWatching.

Also maybe assumptions about corresponding state in mandatory setter.

I didn't say it would be simple, I just think it is worth it.

You should be able to render a frozen object in a template.

mmun:
Watching frozen objects works now.

PR #14274, opened 2016-09-13
Click helper enhancement

BryanCrotaz:
Improve click helper

Note that tests did not pass before the change, so could not verify behaviour.

### DIFF
--- a/packages/ember-testing/lib/events.js
+++ b/packages/ember-testing/lib/events.js
@@ -26,6 +26,10 @@ export function focus(el) {
   }
 }
 
+function isNullOrUndefined(obj) {
+  return obj === undefined || obj === null;
+}
+
 export function fireEvent(element, type, options = {}) {
   if (!element) {
     return;
@@ -38,10 +42,10 @@ export function fireEvent(element, type, options = {}) {
     let x = rect.left + 1;
     let y = rect.top + 1;
     let simulatedCoordinates = {
-      screenX: options.screenX || x + 5,
-      screenY: options.screenY || y + 95,
-      clientX: options.clientX || x,
-      clientY: options.clientY || y
+      screenX: isNullOrUndefined(options.screenX) ? x + 5 : options.screenX,
+      screenY: isNullOrUndefined(options.screenY) ? y + 95 : options.screenY,
+      clientX: isNullOrUndefined(options.clientX) ? x : options.clientX,
+      clientY: isNullOrUndefined(options.clientY) ? y : options.clientY
     };
     if (!options.button) {
       // default to left button
--- a/packages/ember-testing/lib/events.js
+++ b/packages/ember-testing/lib/events.js
@@ -47,7 +47,7 @@ export function fireEvent(element, type, options = {}) {
       // default to left button
       options.button = 0;
     }
-    options.which = options.button;
+    options.which = options.button + 1; // jquery normalises to 1,2,3 for left, middle, right
     event = buildMouseEvent(type, jQuery.extend(simulatedCoordinates, options));
   } else {
     event = buildBasicEvent(type, options);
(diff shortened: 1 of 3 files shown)


packages/ember-testing/lib/events.js
```diff
@@ -38,11 +38,16 @@ export function fireEvent(element, type, options = {}) {
     let x = rect.left + 1;
     let y = rect.top + 1;
     let simulatedCoordinates = {
-      screenX: x + 5,
-      screenY: y + 95,
-      clientX: x,
-      clientY: y
+      screenX: options.screenX || x + 5,
```
stefanpenner: what if the option provided is the coordinate of `0` it will be falsey, and the `||` will fall to the right hand side, is that intentional?

BryanCrotaz:
Whoops!!

locks (changes requested):
Please address https://github.com/emberjs/ember.js/pull/14274/files#r82728021 :)

BryanCrotaz:
@locks addressed

packages/ember-testing/lib/helpers/click.js
```diff
@@ -16,22 +16,34 @@ import { focus, fireEvent } from '../events';
...
+  });
+  ```
+
   @method click
-  @param {String} selector jQuery selector for finding element on the DOM
-  @param {Object} context A DOM Element, Document, or jQuery to use as context
+  @param {String} selector jQuery selector for finding an element in the DOM. Note that only the first match will be clicked.
+  @param {Object} context A DOM Element, Document, or jQuery to use as context.
+    If null or undefined, default is the root application DOM element.
+  @param {Object} event An optional hash of event parameters to mock the JQuery [Event]{@link http://api.jquery.com/category/events/event-object/} object.
```
stefanpenner: this may be considered a new feature, although I could be convinced this is a bugfix

@rwjblue what do you think good sir.

packages/ember-testing/lib/events.js
```diff
@@ -38,11 +42,16 @@ export function fireEvent(element, type, options = {}) {
...
-      clientX: x,
-      clientY: y
+      screenX: isNullOrUndefined(options.screenX) ? x + 5 : options.screenX,
+      screenY: isNullOrUndefined(options.screenY) ? y + 95 : options.screenY,
+      clientX: isNullOrUndefined(options.clientX) ? x : options.clientX,
+      clientY: isNullOrUndefined(options.clientY) ? y : options.clientY
     };
+    if (!options.button) {
+      // default to left button
+      options.button = 0;
```
stefanpenner: what was the original behavior?

packages/ember-testing/lib/events.js
```diff
@@ -38,11 +42,16 @@ export function fireEvent(element, type, options = {}) {
...
+    }
+    options.which = options.button + 1; // jquery normalises to 1,2,3 for left, middle, right
```
stefanpenner: what was the original behavior?

stefanpenner: what was the original behavior? (making sure this isn't a breaking change)

BryanCrotaz: Always used left button, like it or not. There was no way to test right click

packages/ember-testing/lib/helpers/click.js
rwjblue: Yes, likely is adding new functionality / API surface here. I think a quick / short RFC would be good. It does seem like a no brainer to me, but the RFC process helps make sure that we aren't missing obvious things...

pixelhandler:
@locks did you see the response from @BryanCrotaz yet?

kevinleeleeedds (approved):

kevinleeleeedds:
If th th nothing yt? U ight you yt? Your ugh occurring of cm kickoff unblock

locks (approved):

mmun:
These enhancements are available in https://github.com/emberjs/ember-test-helpers.

See https://github.com/emberjs/ember-test-helpers/blob/30b135f9f946185918ca6087edc1478ee3ce06ff/addon-test-support/@ember/test-helpers/dom/fire-event.js#L96-L104.

Issue #14273, opened 2016-09-12
Ember 2.8.0 - Setting properties in `willDestroyElement` results in an Error that breaks tests on teardown


fivetanley:
You used to be able to set properties in `willDestroyElement` and
have them work before Ember 2.8. This mostly works, except when you
have an `{{#if}}` helper in your template. On Ember 2.8.0, this
results in an error:

```
TypeError: Cannot read property 'push' of null
    at cleanupRenderNode (http://localhost:7357/assets/vendor.js:20703:47)
    at destroyNode (http://localhost:7357/assets/vendor.js:57340:9)
    at Object.visitChildren (http://localhost:7357/assets/vendor.js:56988:7)
    at Object.clearMorph (http://localhost:7357/assets/vendor.js:57357:29)
    at RenderResult.destroy (http://localhost:7357/assets/vendor.js:56666:32)
    at Renderer.remove (http://localhost:7357/assets/vendor.js:25025:18)
    at Object.destroy (http://localhost:7357/assets/vendor.js:54237:21)
    at Class.destroy (http://localhost:7357/assets/vendor.js:53203:26)
    at Class.superWrapper [as destroy] (http://localhost:7357/assets/vendor.js:35186:22)
    at Object.run (http://localhost:7357/assets/vendor.js:10681:25)
```

This error mainly seems to occur in tests when components are being
torn down.

Component definition:

```javascript
import Ember from 'ember';

export default Ember.Component.extend({
  showWelcome: true,

  willDestroyElement() {
    Ember.run(() => {
      this.set('showWelcome', false);
    });
    this._super(...arguments);
  }
});
```

Component Template:

```handlebars
{{#if showWelcome}}
  Welcome!
{{/if}}
```

Interestingly enough, this error does *not* occur if you have an
`{{else}}` case, like so:

```handlebars
{{#if showWelcome}}
  Welcome!
{{else}}
  Goodbye
{{/if}}
```

I made a reproduction repo [here](https://github.com/fivetanley/will-destroy-element-reproduction).

You can clone the repo, run npm install and bower install, and then start the test server (`ember t --s`) as usual to get the error state.

rwjblue:
@fivetanley - I cloned and tested that repo, but all tests are passing. I noticed that they are only JSHint tests and I don't see where you added custom components and/or tests after the initial commit.  Maybe you forgot to push your changes?

fivetanley:
[...]

Issue #14266, opened 2016-09-12
Ember 2.8.0 error while hitting `view.parentView = null;` in `cleanupRenderNode` in ember-htmlbars


josemariaruiz:
Hi,

I upgraded my app to 2.8.0 and now I get this error in the browser debugger.

```ember.debug.js:6778 Uncaught Error: Assertion Failed: You must use Ember.set() to set the `parentView` property (of <spa@component:md-card-content::ember940>) to `null`.```

This is interesting, as it's EmberJS telling me that something cannot be done (setting `parentView`  directly) and yet `ember-htmlbars` does it.

Check 

```
enifed('ember-htmlbars/hooks/cleanup-render-node', ['exports'], function (exports) {
  /**
  @module ember
  @submodule ember-htmlbars
  */

  'use strict';

  exports.default = cleanupRenderNode;

  function cleanupRenderNode(renderNode) {
    var view = renderNode.emberView;
    if (view) {
      view.renderer.willDestroyElement(view);
      view.ownerView._destroyingSubtreeForView.push(function (env) {
        view._transitionTo('destroying'); // unregisters view
        // prevents rerender and scheduling
        view._renderNode = null;
        renderNode.emberView = null;

        view.renderer.didDestroyElement(view);

        if (view.parentView && view.parentView === env.view) {
          view.parentView.removeChild(view);
        }
        view.parentView = null;

        view._transitionTo('preRender');
      });
    }

    if (renderNode.cleanup) {
      renderNode.cleanup();
    }
  }
});
```

josemariaruiz:
I traced the problem to the bower module (I don't know if that's the name) that you install with `bower install`. 

The issue is described in this issue report: `https://github.com/components/ember/issues/77`

rwjblue:
This is a duplicate of https://github.com/emberjs/ember.js/issues/14260. As I mentioned in that issue, you should generally not be observing `parentView`. Can you provide a reproduction / example of what you are doing?

josemariaruiz:
A `git grep` in my source code shows no use of `parentView` so I assume that the problem in in a third party.

I'm not doing anything strange in particular, app relays on `ember-cli-materialize` and it has a component called `md-card-content`. 

Running the same code with EmberJS 2.7.0 (that version works flawlessly) I've got this warning:

[...]

PR #14281, opened 2016-09-13
Fix release regressions.

rwjblue:
Fixes https://github.com/emberjs/ember.js/issues/14266.
Fixes https://github.com/emberjs/ember.js/issues/14273.

_Note_ This targets `release` directly (as HTMLBars is no longer part of master/beta branches).

### DIFF
--- a/packages/ember-views/lib/mixins/view_support.js
+++ b/packages/ember-views/lib/mixins/view_support.js
@@ -454,6 +454,10 @@ export default Mixin.create({
   },
 
   scheduleRevalidate(node, label, manualRerender) {
+    if (this.isDestroying) {
+      return;
+    }
+
     if (node && !this._dispatching && this._env.renderedNodes.has(node)) {
       if (manualRerender) {
         deprecate(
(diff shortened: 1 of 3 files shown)


krisselden:
@rwjblue looks good

krisselden:
@rwjblue I'm not sure about the `||` in the condition, why would you ever want to schedule if isDestroying?

packages/ember-views/lib/mixins/view_support.js
```diff
@@ -478,7 +478,7 @@ export default Mixin.create({
       { id: 'ember-views.dispatching-modify-property', until: '3.0.0' }
     );
 
-    if (!this.scheduledRevalidation || this._dispatching) {
+    if ((!this.scheduledRevalidation && !this.isDestroying) || this._dispatching) {
```
krisselden: this boolean doesn't seem correct, why would we schedule if isDestroying but we are this._dispatching? seems like isDestroying should just exit early altogether.

rwjblue:
Good point, updated.

PR #14295, opened 2016-09-15
Router.map variable declared twice with same value

jasonmit:
Minor, just noticed this while getting ready to send another PR to improve `Router.map`


rwjblue:
Thanks!

PR #14315, opened 2016-09-17
[Bugfix beta] Fix DOM output for props and attrs

chadhietala:
Going to require https://github.com/tildeio/glimmer/pull/319


packages/ember-glimmer/lib/utils/bindings.js
```diff
@@ -70,34 +70,9 @@ export const AttributeBinding = {
...
     if (attribute === 'style') {
       reference = new StyleBindingReference(reference, referenceForKey(component, 'isVisible'));
-    } else {
-      reference = map(reference, this.mapAttributeValue);
     }
 
     operations.addDynamicAttribute(element, attribute, reference);
-  },
-
-  mapAttributeValue(value) {
```
chadhietala: From what I can tell these were undergoing this normalization twice and would cause issues downstream.

PR #14319, opened 2016-09-19
[BUGFIX beta] Fixes #14298 rendering aliased paths

### LINKED ISSUE #14298: [Ember 2.9-beta2] Missing render revalidation in production build
I tried to update my app to v2.9.0-beta2 in which I encountered a rendering regression, where some of my UI failed to update after an initial render, deferring it until another part of the UI needed revalidation.

While I did expect some corner cases to have issues, I did not expect it to only appear when using the production builds, as this does!

I have spent some time to isolate the cause, and I ended up with this code on top of a brand new `ember init` project:

*app/templates/application.hbs*
```hbs
Value: {{value}}
```

*app/controllers/application.js*
```js
import Ember from 'ember';

    object: Ember.Object.create({ value: 'fail' }),
    value: Ember.computed.readOnly('object.value'),

    init() {
        this._super(...arguments);
        Ember.run.later(() => {
            this.set('object.value', 'ok');
        }, 200);
    },
});
```

* Rendered value using debug build: `ok`.
* Rendered value using production build: `fail`.

Notes:

* If I render `{{object.value}}` instead, it renders correctly in both cases.
* `this.get('value')` always gets the correct value.
* I have tried to replicate this in a jsbin, but it always renders correctly, even using the minified builds.
* This also fails on the latest `canary` build.

krisselden:
When dirtying an object, other objects with aliases that path through this object should dirty as well.

### DIFF
--- a/packages/ember-glimmer/tests/integration/content-test.js
+++ b/packages/ember-glimmer/tests/integration/content-test.js
@@ -403,6 +403,45 @@ class DynamicContentTest extends RenderingTest {
     this.assertContent('hello');
     this.assertInvariants();
   }
+
+  ['@test it can render a readOnly property of a path']() {
+    let Messenger = EmberObject.extend({
+      message: computed.readOnly('a.b.c')
+    });
+
+    let messenger = Messenger.create({
+      a: {
+        b: {
+          c: 'hello'
+        }
+      }
+    });
+
+    this.renderPath('messenger.message', { messenger });
+
+    this.assertContent('hello');
+
+    this.assertStableRerender();
+
+    this.runTask(() => set(messenger, 'a.b.c', 'hi'));
+
+    this.assertContent('hi');
+    this.assertInvariants();
+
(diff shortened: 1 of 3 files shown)


rwjblue (approved):

mmun (approved):

chancancode:
Benchmarks:

* [Skylight medium dataset](https://cdn.rawgit.com/chancancode/47e93954dcc9b2b13af6a5ba16481445/raw/f98a62f35197f1effc16da987c1ec9a3af7fc96c/medium.html)
* [Skylight large dataset](https://cdn.rawgit.com/chancancode/47e93954dcc9b2b13af6a5ba16481445/raw/f98a62f35197f1effc16da987c1ec9a3af7fc96c/large.html)

It does introduce a small performance cost for initialized but not observed aliases, but overall the impact is pretty small.

chancancode (approved):

PR #14320, opened 2016-09-20
Refactor descriptor computed

locks:
Complements https://github.com/emberjs/ember.js/pull/14138.


chadhietala:
I suspect we want to measure this before merging. While I don't think there is going to be perf regression, we need to make sure that the transpiled output in these hot paths have an effect.

locks:
@chadhietala 👍  how do I do that?

chadhietala:
Perhaps @chancancode can help / has an opinion.

PR #14330, opened 2016-09-21
Remove duplicated comments

san650:


packages/ember-glimmer/lib/component.js
```diff
@@ -308,24 +308,6 @@ const Component = CoreView.extend(
      */
 
     /**
-     Called when the attributes passed into the component have been updated.
-     Called both during the initial render of a container and during a rerender.
-     Can be used in place of an observer; code placed here will be executed
-     every time any attribute updates.
-     @event didReceiveAttrs
```
locks: where did `didReceiveAttrs` go?

locks (changes requested):

packages/ember-glimmer/lib/component.js
locks: nevermind, hidden in the fold above.

locks (approved):

locks:
Thanks for your contribution!

@homu r+

rwjblue:
These were not duplicated (as far as I can tell), the documentation is for them as both an event and a method.

locks:
Ah, damnit :( I triple checked and missed it. Reverting it, apologies @san650.

san650:
@rwjblue @locks You're right, I didn't notice they were different.

Sorry for all the trouble!

PR #14331, opened 2016-09-22
[DOC] objectAtContent should be public

ssured:
See lines 52-63 of this file, which document how to override this method.
https://github.com/ssured/ember.js/blob/95daad4d066cdd12ab961176b7e5f8ba92de2e26/packages/ember-runtime/lib/system/array_proxy.js#L52-L63


rwjblue:
This seems correct to me, but I'd like another core 👍 before landing.

@stefanpenner - Can you confirm this one?

stefanpenner:
cc @tomdale  / @wycats

tomdale:
Yes it should be public

PR #14339, opened 2016-09-23
[DOC] Make public two-week policy for closing issues

Serabe:
When triaging a team, we use the _Needs Submitter Response_ to mark
issues that cannot be solved without some extra information from the
original submitter. If the submitter does not provide any extra
information after seven days, the team gently reminds the original
submitter. If another week passes, the issue might be closed.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,6 +45,18 @@ a stab at fixing the bug yourself if you can!
 The more information you provide, the easier it is for us to validate that
 there is a bug and the faster we'll be able to take action.
 
+### Triaging policy
+
+* You might be requested to provide a reproduction or extra information. In that
+case, the issue will be labeled as _Needs Submitter Response_. If we did not
+get any response after seven days, we will ping you to remind you about it. We
+might close the issue if we do not hear from you after two weeks since the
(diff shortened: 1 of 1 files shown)


Serabe:
@locks, @pixelhandler: a draft of what we talked about today!

locks:
I think you should introduce a triaging policy header so we can list our policies, starting with this one, and the lifecycle of an issue. The copy is 👍 , so it's a good start.

rwjblue:
It seems great to document and codify the policies we follow here. I'll try to get confirmation from the rest of the core team on this during the F2F meeting this weekend.

stefanpenner:
👍

Serabe:
Thank you for your comments and so sorry for the long delay. I've added the header, as @locks suggested. I've also added what we currently do when the issue submitted is actually a feature request.

I think these two are the main policies we are following right now and are a good starting point for this public policy.

Thank you again for taking your time to review this!

tomdale:
LGTM

rwjblue:
Discussed with core team today (sorry for prolonged delay), we are 👍. Might tweak the process/verbiage a bit over time, but this is good to go...

PR #14341, opened 2016-09-24
Add @since doc tags to everything in Ember.Route that didn't already have it

ziggythehamster:
I keep finding myself finding stuff in the docs that I have to double-check to see if the version of Ember I'm using actually supports. Today, the thing I was looking at was in Ember.Route, and the docs were missing `@since` tags, so I had to check the source of Ember 1.10 to see if I could actually use it. I can.

This PR adds `@since` to all public APIs in Ember.Route, even when it's been there since 1.0.0, so that users like myself that aren't on the latest version can be sure that the thing that might solve our problem is available in our version of Ember.

I checked 0.9 out, and it seems like there wasn't a concept of a route in 0.9 (oh how far we've come!), so I didn't bother checking any earlier than 1.0.0. If I should have, let me know and I'll give it another whirl.

### DIFF
--- a/packages/ember-routing/lib/system/route.js
+++ b/packages/ember-routing/lib/system/route.js
@@ -80,6 +80,7 @@ export function hasDefaultSerialize(route) {
   @extends Ember.Object
   @uses Ember.ActionHandler
   @uses Ember.Evented
+  @since 1.0.0
   @public
 */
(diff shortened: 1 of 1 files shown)


locks (approved):
Great work @ziggythehamster! Thank you for the thoroughness and going back all the way to 1.0!

PR #14344, opened 2016-09-24
Component’s `element` should be present in `willInsertElement` hook

### LINKED ISSUE #14343: Element not present during `willInsertElement`
Inside `willInsertElement` the `element` property should be set (but not inserted into the DOM).

In <= 2.8 this is working properly ([see example](http://jsbin.com/xogijug/1/edit?html,js,output)).

In >= 2.9+ `element` is not set in `willInsertElement` ([see example](http://jsbin.com/pexarix/edit?html,js,output)).

ebryn:
@rwjblue 

Fixes #14343


rwjblue:
Thank you @ebryn 😽

PR #14346, opened 2016-09-25
[BUGFIX beta] Remove `isTopLevel` flag in `DynamicScope`.

rwjblue:
Remove the need for `DynamicScope#isTopLevel` to be threaded throughout the full system, and rely on the same setup that we recently added for `Component#appendTo`.

This sets the stage for sharing a single root template with an `{{#each` and removing the root iteration in the `Renderer`.

/cc @krisselden

### DIFF
--- a/packages/ember-glimmer/lib/environment.js
+++ b/packages/ember-glimmer/lib/environment.js
@@ -96,10 +96,12 @@ export default class Environment extends GlimmerEnvironment {
     }, ({ Template, owner }) => guidFor(owner) + '|' + Template.id);
 
     this._compilerCache = new Cache(10, Compiler => {
-      return new Cache(2000, template => {
+      return new Cache(2000, ({ template }) => {
         let compilable = new Compiler(template);
         return compileLayout(compilable, this);
-      }, template => template.id);
+      }, ({ template, owner })=> {
+        return guidFor(owner) + '|' + template.id;
+      });
     }, Compiler => Compiler.id);
 
     this.builtInModifiers = {
@@ -252,7 +254,7 @@ export default class Environment extends GlimmerEnvironment {
   // a Compiler can wrap the template so it needs its own cache
   getCompiledBlock(Compiler, template, owner) {
     let compilerCache = this._compilerCache.get(Compiler);
-    return compilerCache.get(template, owner);
+    return compilerCache.get({ template, owner });
   }
 
   hasPartial(name, symbolTable) {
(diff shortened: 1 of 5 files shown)


packages/ember-glimmer/lib/environment.js
```diff
@@ -252,7 +254,7 @@ export default class Environment extends GlimmerEnvironment {
   // a Compiler can wrap the template so it needs its own cache
   getCompiledBlock(Compiler, template, owner) {
     let compilerCache = this._compilerCache.get(Compiler);
-    return compilerCache.get(template, owner);
+    return compilerCache.get({ template, owner });
```
krisselden: as a follow up we may want to add a context object we reuse so we don't have to keep allocating this object for the lookup but we can follow up on that

rwjblue: kk, I also thought about making the cache pass through arbitrary numbers of positional arguments

packages/ember-glimmer/lib/syntax/outlet.js
```diff
@@ -299,9 +264,12 @@ class TopLevelOutletLayoutCompiler {
...
 TopLevelOutletLayoutCompiler.id = 'top-level-outlet';
 
-class OutletComponentDefinition extends AbstractOutletComponentDefinition {
+class OutletComponentDefinition extends ComponentDefinition {
   constructor(outletName, template) {
-    super(MANAGER, outletName, template);
+    super('outlet', MANAGER, null);
+    this.outletName = outletName;
+    this.template = template;
+    generateGuid(this);
```
krisselden: what is generateGuid(this) for?

rwjblue: It was previously done in the `AbstractOutletComponentDefinition` which is removed (since it does very little for us). I just kept it from there when I refactored out the abstract parent.

krisselden (commented):
Looks good, my one comment is I expected TopOutletManager to extend OutletManager and just do the wrapLayout compiler.

packages/ember-glimmer/lib/syntax/outlet.js
rwjblue: It was previously done in the `AbstractOutletComponentDefinition` which is removed (since it does very little for us). I just kept it from there when I refactored out the abstract parent.

rwjblue:
Updated to make `TopLevelOutletComponentManager` to extend from `OutletComponentManager`.

PR #14349, opened 2016-09-25
Collapse multiple arguments in `CoreObject.create` rather than inside of `makeCtor`

ebryn:
@rwjblue @krisselden 

TODOs/open questions:
- [ ] Benchmark
- [x] Should we avoid splat args? @krisselden says yes
- [ ] Should we try to reduce the code duplication? Yes, but let's benchmark this first
- [ ] Try using `this` instead of `C` in `CoreObject.create`
- [ ] Try to refactor out `initProperties`

### DIFF
--- a/packages/ember-runtime/lib/system/core_object.js
+++ b/packages/ember-runtime/lib/system/core_object.js
@@ -62,7 +62,7 @@ function makeCtor() {
     }
 
     if (arguments.length > 0) {
-      initProperties = [arguments[0]];
+      initProperties = arguments[0];
     }
 
     this.__defineNonEnumerable(GUID_KEY_PROPERTY);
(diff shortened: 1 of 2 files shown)


rwjblue:
Should we warn in the case that `Ember.Object.create` is called with multiple arguments?  I'm unsure how common that case actually is...

packages/ember-runtime/tests/mixins/array_test.js
```diff
@@ -24,8 +24,8 @@ import { A as emberA } from '../../system/native_array';
 const TestArray = EmberObject.extend(EmberArray, {
   _content: null,
 
-  init(ary = []) {
-    this._content = ary;
+  init() {
```
rwjblue: Should call `_super` before touching `this`

krisselden:
Are you doing the benchmarking?

krisselden:
@ebryn ping, do you need help?

packages/ember-runtime/tests/mixins/array_test.js
```diff
@@ -60,7 +60,7 @@ ArrayTests.extend({
 
   newObject(ary) {
     ary = ary ? ary.slice() : this.newFixture(3);
-    return new TestArray(ary);
+    return TestArray.create({ _content: ary });
```
stefanpenner: I think this would be a regression no?

krisselden: This is testing low level building blocks, this stuff isn't documented, I don't even remember which thing needs this set up this way.

packages/ember-runtime/tests/mixins/array_test.js
krisselden: that content array stuff is whack. I should take another look at it I can't remember why it even was like that.

krisselden:
We can still use the closure and remove changing the test, it would be still good to get the arg flattening, also @rwjblue you benched this at one point, can you include the data?

stefanpenner:
> @ebryn ping, do you need help?

^^

packages/ember-runtime/lib/system/core_object.js
```diff
@@ -706,12 +700,55 @@ var ClassMixinProps = {
...
+  create() {
     var C = this;
-    if (args.length > 0) {
-      this._initProperties(args);
+    if (arguments.length === 0) {
+      return new C();
+    } else if (arguments.length === 1) {
+      return new C(arguments[0]);
+    } else if (arguments.length > 1) {
+      let concatenatedProperties = this.concatenatedProperties;
```
stefanpenner: `this.concatentatedProperties` ` this.mergedProperties` are instance state, not class state. Im not sure how this works?

ebryn:
I could use help here for sure, don't have the benchmarking system setup on any of my machines.

packages/ember-runtime/lib/system/core_object.js
rwjblue: Actually, these are referring to the list of which properties should be merged/concattenated and definitely should be class state. I believe the code is correct here.

btecu:
This was superseded by #15594.

mmun:
Thanks @btecu.

PR #14355, opened 2016-09-26
Refactoring curly component argument processing

ef4:
This cleans up and unifies the handling of curly component arguments. It splits the work into two distinct phases
- gatherArgs assembles an EvaluatedArgs that combines all variations of positional and named arguments
- ComponentArgs.create takes the EvaluatedArgs and produces the curly-component specific attrs & props structures with mutable cells, etc.


rwjblue:
LGTM, @chancancode r?

chancancode:
I want to try to unify this with closure component args, I'll give it a
shit today

ef4:
This is rebased against master. @chancancode, can we break your lock and merge this? It doesn't preclude a further refactor that also includes closure component args.

PR #14359, opened 2016-09-26
Refactor Meta and object destruction tracking.

rwjblue:
- Moves the `isDestroying` and `isDestroyed` flags into meta
- Prevents recreating meta _while_ destroying the object (was happening
  during chain node removal)
- Allows reading values/caches after object destruction, but triggers an
  assertion if non-readable meta methods are called after the object is
  destroyed
- Avoids doing extra work when unwatching keys/paths for an object that
  is also destroyed

### DIFF
--- a/packages/ember-metal/lib/meta.js
+++ b/packages/ember-metal/lib/meta.js
@@ -5,7 +5,10 @@
 import { EmptyObject, lookupDescriptor, symbol } from 'ember-utils';
 import isEnabled from './features';
 import { protoMethods as listenerMethods } from './meta_listeners';
-import { runInDebug } from './debug';
+import { runInDebug, assert } from './debug';
+import {
+  removeChainWatcher
+} from './chains';
 
 let counters = {
   peekCalls: 0,
@@ -55,6 +58,10 @@ let members = {
   tag: ownCustomObject
 };
 
+const SOURCE_DESTROYING = 1 << 1;
+const SOURCE_DESTROYED = 1 << 2;
+const META_DESTROYED = 1 << 3;
+
 if (isEnabled('ember-glimmer-detect-backtracking-rerender') ||
     isEnabled('ember-glimmer-allow-backtracking-rerender')) {
   members.lastRendered = ownMap;
@@ -77,6 +84,10 @@ export function Meta(obj, parentMeta) {
   this._chainWatchers = undefined;
   this._chains = undefined;
   this._tag = undefined;
+
+  // initial value for all flags right now is false
+  // see FLAGS const for detailed list of flags used
+  this._flags = 0;
 
   // used only internally
   this.source = obj;
@@ -104,10 +115,78 @@ Meta.prototype.isInitialized = function(obj) {
   return this.proto !== obj;
 };
 
+const NODE_STACK = [];
+
+Meta.prototype.destroy = function() {
+  if (this.isMetaDestroyed()) { return; }
+
+  // remove chainWatchers to remove circular references that would prevent GC
+  let node, nodes, key, nodeObject;
+  node = this.readableChains();
+  if (node) {
+    NODE_STACK.push(node);
+    // process tree
+    while (NODE_STACK.length > 0) {
+      node = NODE_STACK.pop();
+      // push children
+      nodes = node._chains;
+      if (nodes) {
+        for (key in nodes) {
+          if (nodes[key] !== undefined) {
+            NODE_STACK.push(nodes[key]);
+          }
+        }
+      }
+
+      // remove chainWatcher in node object
+      if (node._watching) {
+        nodeObject = node._object;
+        if (nodeObject) {
+          let foreignMeta = peekMeta(nodeObject);
+          // avoid cleaning up chain watchers when both current and
+          // foreign objects are being destroyed
+          // if both are being destroyed manual cleanup is not needed
+          // as they will be GC'ed and no non-destroyed references will
+          // be remaining
+          if (foreignMeta && !foreignMeta.isSourceDestroying()) {
+            removeChainWatcher(nodeObject, node._key, node, foreignMeta);
+          }
+        }
+      }
+    }
+  }
+
+  this.setMetaDestroyed();
+};
+
 for (let name in listenerMethods) {
   Meta.prototype[name] = listenerMethods[name];
 }
 memberNames.forEach(name => members[name](name, Meta));
+
+Meta.prototype.isSourceDestroying = function isSourceDestroying() {
+  return (this._flags & SOURCE_DESTROYING) !== 0;
+};
+
+Meta.prototype.setSourceDestroying = function setSourceDestroying() {
+  this._flags |= SOURCE_DESTROYING;
+};
+
+Meta.prototype.isSourceDestroyed = function isSourceDestroyed() {
+  return (this._flags & SOURCE_DESTROYED) !== 0;
+};
+
+Meta.prototype.setSourceDestroyed = function setSourceDestroyed() {
+  this._flags |= SOURCE_DESTROYED;
+};
+
+Meta.prototype.isMetaDestroyed = function isMetaDestroyed() {
+  return (this._flags & META_DESTROYED) !== 0;
+};
+
+Meta.prototype.setMetaDestroyed = function setMetaDestroyed() {
+  this._flags |= META_DESTROYED;
+};
 
 // Implements a member that is a lazily created, non-inheritable
 // POJO.
@@ -135,6 +214,8 @@ function inheritedMap(name, Meta) {
   let capitalized = capitalize(name);
 
   Meta.prototype['write' + capitalized] = function(subkey, value) {
+    assert(`Cannot call write${capitalized} after the object is destroyed.`, !this.isMetaDestroyed());
+
     let map = this._getOrCreateOwnMap(key);
     map[subkey] = value;
   };
@@ -161,6 +242,8 @@ function inheritedMap(name, Meta) {
   };
 
   Meta.prototype['clear' + capitalized] = function() {
+    assert(`Cannot call clear${capitalized} after the object is destroyed.`, !this.isMetaDestroyed());
+
     this[key] = undefined;
   };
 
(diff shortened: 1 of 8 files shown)


packages/ember-metal/lib/property_events.js
```diff
@@ -99,15 +99,15 @@ function propertyDidChange(obj, keyName) {
...
+    notifyObservers(obj, keyName, m);
   }
 
 
   if (obj[PROPERTY_DID_CHANGE]) {
     obj[PROPERTY_DID_CHANGE](keyName);
   }
 
-  if (obj.isDestroying) { return; }
+  if (m && m.sourceDestroying) { return; }
```
stefanpenner: i believe we may also bail of `m` is explicity `null`?

packages/ember-metal/lib/meta.js
```diff
@@ -78,6 +78,12 @@ export function Meta(obj, parentMeta) {
   this._chains = undefined;
   this._tag = undefined;
 
+  this._isDestroyed = false;
+
+  // keep the objects destruction state
+  this.sourceDestroying = false;
```
rwjblue: I can do that.

packages/ember-metal/lib/meta.js
```diff
@@ -104,6 +110,11 @@ Meta.prototype.isInitialized = function(obj) {
   return this.proto !== obj;
 };
 
+Meta.prototype.destroy = function() {
+  // clean up chain watchers
```
rwjblue: It is being done when this is called in `ember-metal/watching`'s `destroy` function (which is the only place this is being called). I can move it here though if you'd prefer.

packages/ember-runtime/lib/system/core_object.js
```diff
@@ -418,7 +436,22 @@ CoreObject.PrototypeMixin = Mixin.create({
...
     @public
   */
-  isDestroying: false,
+  isDestroying: descriptor({
+    get() {
+      return meta(this).sourceDestroying;
+    },
+
+    set(value) {
+      // prevent setting while applying mixins
```
rwjblue: The guard here is to prevent an issue in our mixin apply code (we set the value before calling the `descriptor#setup` so the value being guarded against is this descriptor itself).

In general, I don't care about allowing this to be set outside of `destroy` and I can change this to throw if not being set by the quirk/bug in `defineProperty` mentioned above.

packages/ember-metal/lib/property_events.js
rwjblue: Ya we can, but we don't ever set it to `null` manually anymore (used to be nulled out in `deleteMeta`).

stefanpenner:
@krisselden thoughts on us getting clever \w a bitmask for these flags?

rwjblue:
@stefanpenner - Ya, that is what he was meaning in https://github.com/emberjs/ember.js/pull/14359#discussion_r80569246 (and he DM'ed to explain).

stefanpenner:
> @stefanpenner - Ya, that is what he was meaning in #14359 (comment) (and he DM'ed to explain).

Ah, yup that makes sense.

krisselden:
@stefanpenner it isn't just getting clever, it will be only one IC stub per read/write and places that check both isDestroying and isDestroyed can do it with one test.

stefanpenner:
> @stefanpenner it isn't just getting clever, it will be only one IC stub per read/write and places that check both isDestroying and isDestroyed can do it with one test.

I wasn't implying "clever" negatively, I think in this situation it also makes sense.

rwjblue:
Updated:

* Uses bitwise logic to track multiple flags in a single field
* Moves chain node watcher cleanup into `Meta#destroy`
* Throw an error when someone manually sets `isDestroyed` / `isDestroying`

Ready for another round of reviews...

packages/ember-runtime/lib/system/core_object.js
```diff
@@ -418,7 +435,21 @@ CoreObject.PrototypeMixin = Mixin.create({
     @default false
     @public
   */
-  isDestroying: false,
+  isDestroying: descriptor({
+    get() {
+      let m = meta(this);
+      return (m._flags & SOURCE_DESTROYING) !== 0;
```
stefanpenner: can we hide the bitmasks behind functions?`m.isSourceDestroying();` I believe it should come at little or no overhead, but be a-bit more maintainable. 

Petka of bluebird fame took this approach, because those methods tend to inline quite nicely. Nice related blog post -> https://reaktor.com/blog/javascript-performance-fundamentals-make-bluebird-fast/

rwjblue: @stefanpenner - Yeah, I started with that but then switched to using the flags directly in the source after discussing with @krisselden.  I have swapped back to using methods for now (it is much easier to grok for those of us that don't understand the bitwise operations at a glance), but we need to confirm with @krisselden that this is not going to negatively impact the IC hit/miss stuff he has been investigating lately.

rwjblue:
@stefanpenner / @krisselden - Updated again.  Exposes helper methods for the flags, and uses them throughout the source.  Ready for another review...

packages/ember-metal/lib/meta.js
```diff
@@ -78,6 +87,9 @@ export function Meta(obj, parentMeta) {
   this._chains = undefined;
   this._tag = undefined;
 
+  // initial value for all flags right now is false
```
stefanpenner: it help future travelers could we also make a comment here describing the layout of this bitfield?

packages/ember-runtime/lib/system/core_object.js
krisselden: Maybe I'm just old so it looks familiar but I don't find `if (m._flags & SOURCE_DESTROYING) {` unreadable or confusing.

packages/ember-metal/lib/meta.js
```diff
@@ -104,11 +116,65 @@ Meta.prototype.isInitialized = function(obj) {
...
+            NODE_STACK.push(nodes[key]);
+          }
+        }
+      }
+
+      // remove chainWatcher in node object
+      if (node._watching) {
+        nodeObject = node._object;
+        if (nodeObject) {
+          removeChainWatcher(nodeObject, node._key, node);
```
krisselden: nevermind I realize this is the foreign object pointing at us

krisselden: also, didn't we talk about not doing this if the object point at us was destroying or destroyed? the cycle doesn't matter the other object isn't going to be long lived.

krisselden:
@rwjblue / @stefanpenner should we black hole sets after destroy?  It can hide bugs and still fail anything but it also causes a lot of guards.

packages/ember-metal/lib/property_events.js
```diff
@@ -254,8 +255,8 @@ function changeProperties(callback, binding) {
   }
 }
 
-function notifyBeforeObservers(obj, keyName) {
-  if (obj.isDestroying) { return; }
+function notifyBeforeObservers(obj, keyName, m) {
```
dgeb: how about `meta` (instead of `m`) for consistency with `dependentKeysDidChange` and `dependentKeysWillChange`?

packages/ember-metal/lib/property_events.js
```diff
@@ -268,8 +269,8 @@ function notifyBeforeObservers(obj, keyName) {
   }
 }
 
-function notifyObservers(obj, keyName) {
-  if (obj.isDestroying) { return; }
+function notifyObservers(obj, keyName, m) {
```
dgeb: again `m` -> `meta`

packages/ember-runtime/lib/system/core_object.js
dgeb: Same, same ... not sure it's worth it to take a hard line against bitmasks that aren't even publicly exposed.

rwjblue:
OK, updated again:

* Rename single char variables to better names
* Prevent cleaning up chain nodes when both current object and foreign object are being destroyed
* Add comment describing flags

packages/ember-metal/lib/meta.js
```diff
@@ -104,11 +117,70 @@ Meta.prototype.isInitialized = function(obj) {
   return this.proto !== obj;
 };
 
+const NODE_STACK = [];
+
+Meta.prototype.destroy = function() {
+  // remove chainWatchers to remove circular references that would prevent GC
```
krisselden: destroy should be idempotent, should we not check the meta destroyed flag here and set it?

packages/ember-metal/lib/meta.js
```diff
@@ -104,11 +117,70 @@ Meta.prototype.isInitialized = function(obj) {
...
+        }
+      }
+
+      // remove chainWatcher in node object
+      if (node._watching) {
+        nodeObject = node._object;
+        if (nodeObject) {
+          let foreignMeta = peekMeta(nodeObject);
+          // avoid cleaning up chain watchers when both current and
+          // foreign objects are being destroyed
```
krisselden: just comment here that marking a source as destroying means you are expected to not be holding onto a reference to the source from anything reachable.

packages/ember-metal/lib/meta.js
```diff
@@ -104,11 +117,70 @@ Meta.prototype.isInitialized = function(obj) {
...
 }
 memberNames.forEach(name => members[name](name, Meta));
 
+for (let name in FLAGS) {
+  buildFlagAccessors(name, FLAGS[name], Meta);
+}
+
+function buildFlagAccessors(name, flag, Meta) {
+  let capitalized = capitalize(name);
+  Meta.prototype['is' + capitalized] = function() {
```
krisselden: lets just do normal methods here please

packages/ember-metal/lib/meta.js
```diff
@@ -104,11 +117,70 @@ Meta.prototype.isInitialized = function(obj) {
...
+    return (this._flags & flag) !== 0;
+  };
+
+  Meta.prototype['set' + capitalized] = function() {
+    this._flags |= flag;
+  };
+
+  Meta.prototype['unset' + capitalized] = function() {
```
krisselden: is this just for completeness? do we currently unset any flags?

rwjblue: Yes, I just did it for completeness.  Will remove.

packages/ember-metal/lib/meta.js
rwjblue: Updated!

packages/ember-metal/lib/meta.js
rwjblue: Great point! Updated...

packages/ember-metal/lib/meta.js
rwjblue: Great point, updated to noop if `isMetaDestroying` is already set.

PR #14360, opened 2016-09-26
[FEATURE factory-for] Implement factoryFor

chadhietala:
Implements `lookup.factoryFor` to avoid double extending all objects.

See [RFC](https://github.com/emberjs/rfcs/blob/master/text/0150-factory-for.md)

### DIFF
--- a/packages/container/lib/container.js
+++ b/packages/container/lib/container.js
@@ -5,13 +5,18 @@ import {
   setOwner,
   OWNER,
   assign,
-  NAME_KEY
+  NAME_KEY,
+  HAS_NATIVE_PROXY
 } from 'ember-utils';
 import { ENV } from 'ember-environment';
-import { assert, deprecate, runInDebug, isFeatureEnabled } from 'ember-metal';
+import {
+  assert,
+  deprecate,
+  runInDebug,
+  isFeatureEnabled
+} from 'ember-metal';
 
 const CONTAINER_OVERRIDE = symbol('CONTAINER_OVERRIDE');
-const HAS_PROXY = typeof Proxy === 'function';
 export const FACTORY_FOR = symbol('FACTORY_FOR');
 export const LOOKUP_FACTORY = symbol('LOOKUP_FACTORY');
 
@@ -38,9 +43,6 @@ export default function Container(registry, options) {
   this[CONTAINER_OVERRIDE] = undefined;
   this.isDestroyed = false;
 }
-
-Container.__FACTORY_FOR__ = FACTORY_FOR;
-Container.__LOOKUP_FACTORY__ = LOOKUP_FACTORY;
 
 Container.prototype = {
   /**
@@ -232,7 +234,7 @@ Container.prototype = {
  * set on the manager.
  */
 function wrapManagerInDeprecationProxy(manager) {
-  if (HAS_PROXY) {
+  if (HAS_NATIVE_PROXY) {
     let validator = {
       get(obj, prop) {
         if (prop !== 'class' && prop !== 'create') {
--- a/packages/container/lib/container.js
+++ b/packages/container/lib/container.js
@@ -142,9 +142,11 @@ Container.prototype = {
   lookupFactory(fullName, options) {
     assert('fullName must be a proper full name', this.registry.validateFullName(fullName));
 
-    if (isFeatureEnabled('ember-no-double-extend')) {
-      deprecate('Using "_lookupFactory" is deprecated. Please use container.factoryFor instead.', false, { id: 'container-lookupFactory', until: '2.13.0', url: 'TODO' });
-    }
+    deprecate(
+      'Using "_lookupFactory" is deprecated. Please use container.factoryFor instead.',
+      !isFeatureEnabled('ember-factory-for'),
+      { id: 'container-lookupFactory', until: '2.13.0', url: 'TODO' }
+    );
 
     return deprecatedFactoryFor(this, this.registry.normalize(fullName), options);
   },
@@ -154,57 +156,27 @@ Container.prototype = {
     return deprecatedFactoryFor(this, this.registry.normalize(fullName), options);
   },
 
+  /*
+   * This internal version of factoryFor swaps between the public API for
+   * factoryFor (class is the registered class) and a transition implementation
+   * (class is the double-extended class). It is *not* the public API version
+   * of factoryFor, which always returns the registered class.
+   */
   [FACTORY_FOR](fullName, options = {}) {
-    let manager;
     if (isFeatureEnabled('ember-no-double-extend')) {
-      let normalizedName = this.registry.normalize(fullName);
-      assert('fullName must be a proper full name', this.registry.validateFullName(normalizedName));
-
-      if (options.source) {
-        normalizedName = this.registry.expandLocalLookup(fullName, options);
-        // if expandLocalLookup returns falsey, we do not support local lookup
-        if (!normalizedName) { return; }
+      if (isFeatureEnabled('ember-factory-for')) {
+        return this.factoryFor(fullName, options);
+      } else {
+        /* This throws in case of a poorly designed build */
+        throw new Error('If ember-no-double-extend is enabled, ember-factory-for must also be enabled');
       }
-
-      let factory = this.registry.resolve(normalizedName);
-
-      if (factory === undefined) { return; }
-
-      manager = new FactoryManager(this, factory, fullName, normalizedName);
-    } else {
-      let factory = this.lookupFactory(fullName, options);
-      if (factory === undefined) { return; }
-      manager = new DeprecatedFactoryManager(this, factory, fullName);
     }
+    let factory = this.lookupFactory(fullName, options);
+    if (factory === undefined) { return; }
+    let manager = new DeprecatedFactoryManager(this, factory, fullName);
 
     runInDebug(() => {
-      if (HAS_PROXY) {
-        let validator = {
-          get(obj, prop) {
-            if (prop !== 'class' && prop !== 'create') {
-              throw new Error(`You attempted to access "${prop}" on a factory manager created by container#factoryFor. "${prop}" is not a member of a factory manager."`);
-            }
-
-            return obj[prop];
-          },
-          set(obj, prop, value) {
-            throw new Error(`You attempted to set "${prop}" on a factory manager created by container#factoryFor. A factory manager is a read-only construct.`);
-          }
-        };
-
-        // Note:
-        // We have to proxy access to the manager here so that private property
-        // access doesn't cause the above errors to occur.
-        let m = manager;
-        let proxiedManager = {
-          class: m.class,
-          create(props) {
-            return m.create(props);
-          }
-        };
-
-        manager = new Proxy(proxiedManager, validator);
-      }
+      manager = wrapManagerInDeprecationProxy(manager);
     });
 
     return manager;
@@ -255,10 +227,48 @@ Container.prototype = {
   }
 };
 
+/*
+ * Wrap a factory manager in a proxy which will not permit properties to be
+ * set on the manager.
+ */
+function wrapManagerInDeprecationProxy(manager) {
+  if (HAS_PROXY) {
+    let validator = {
+      get(obj, prop) {
+        if (prop !== 'class' && prop !== 'create') {
+          throw new Error(`You attempted to access "${prop}" on a factory manager created by container#factoryFor. "${prop}" is not a member of a factory manager."`);
+        }
+
+        return obj[prop];
+      },
+      set(obj, prop, value) {
+        throw new Error(`You attempted to set "${prop}" on a factory manager created by container#factoryFor. A factory manager is a read-only construct.`);
+      }
+    };
+
+    // Note:
+    // We have to proxy access to the manager here so that private property
+    // access doesn't cause the above errors to occur.
+    let m = manager;
+    let proxiedManager = {
+      class: m.class,
+      create(props) {
+        return m.create(props);
+      }
+    };
+
+    return new Proxy(proxiedManager, validator);
+  }
+
+  return manager;
+}
+
 if (isFeatureEnabled('ember-factory-for')) {
   /**
-   Given a fullName, return the corresponding factory. The consumer of factory is repsonsible
-   for the destruction of the factory as there is no way to understand the objects lifecycle.
+   Given a fullName, return the corresponding factory. The consumer of the factory
+   is responsible for the destruction of any factory instances, as there is no
+   way for the container to ensure instances are destroyed when it itself is
+   destroyed.
 
    @public
    @method factoryFor
@@ -267,8 +277,27 @@ if (isFeatureEnabled('ember-factory-for')) {
    @param {String} [options.source] The fullname of the request source (used for local lookup)
    @return {any}
    */
-  Container.prototype.factoryFor = function _factoryFor() {
-    return this[FACTORY_FOR](...arguments);
+  Container.prototype.factoryFor = function _factoryFor(fullName, options = {}) {
+    let normalizedName = this.registry.normalize(fullName);
+    assert('fullName must be a proper full name', this.registry.validateFullName(normalizedName));
+
+    if (options.source) {
+      normalizedName = this.registry.expandLocalLookup(fullName, options);
+      // if expandLocalLookup returns falsey, we do not support local lookup
+      if (!normalizedName) { return; }
+    }
+
+    let factory = this.registry.resolve(normalizedName);
+
+    if (factory === undefined) { return; }
+
+    let manager = new FactoryManager(this, factory, fullName, normalizedName);
+
+    runInDebug(() => {
+      manager = wrapManagerInDeprecationProxy(manager);
+    });
+
+    return manager;
   };
 }
 
--- a/packages/container/lib/container.js
+++ b/packages/container/lib/container.js
@@ -12,6 +12,8 @@ import { assert, deprecate, runInDebug, isFeatureEnabled } from 'ember-metal';
 
 const CONTAINER_OVERRIDE = symbol('CONTAINER_OVERRIDE');
 const HAS_PROXY = typeof Proxy === 'function';
+export const FACTORY_FOR = symbol('FACTORY_FOR');
+export const LOOKUP_FACTORY = symbol('LOOKUP_FACTORY');
 
 /**
  A container used to instantiate and cache objects.
@@ -36,6 +38,9 @@ export default function Container(registry, options) {
   this[CONTAINER_OVERRIDE] = undefined;
   this.isDestroyed = false;
 }
+
+Container.__FACTORY_FOR__ = FACTORY_FOR;
+Container.__LOOKUP_FACTORY__ = LOOKUP_FACTORY;
 
 Container.prototype = {
   /**
@@ -137,17 +142,21 @@ Container.prototype = {
   lookupFactory(fullName, options) {
     assert('fullName must be a proper full name', this.registry.validateFullName(fullName));
 
-    if (isFeatureEnabled('container-factoryFor')) {
+    if (isFeatureEnabled('ember-no-double-extend')) {
       deprecate('Using "_lookupFactory" is deprecated. Please use container.factoryFor instead.', false, { id: 'container-lookupFactory', until: '2.13.0', url: 'TODO' });
     }
 
     return deprecatedFactoryFor(this, this.registry.normalize(fullName), options);
   },
 
-  factoryFor(fullName, options = {}) {
-    let container = this;
+  [LOOKUP_FACTORY](fullName, options) {
+    assert('fullName must be a proper full name', this.registry.validateFullName(fullName));
+    return deprecatedFactoryFor(this, this.registry.normalize(fullName), options);
+  },
+
+  [FACTORY_FOR](fullName, options = {}) {
     let manager;
-    if (isFeatureEnabled('container-factoryFor')) {
+    if (isFeatureEnabled('ember-no-double-extend')) {
       let normalizedName = this.registry.normalize(fullName);
       assert('fullName must be a proper full name', this.registry.validateFullName(normalizedName));
 
@@ -161,50 +170,11 @@ Container.prototype = {
 
       if (factory === undefined) { return; }
 
-      manager = {
-        class: factory,
-        create(options = {}) {
-          let injections = injectionsFor(container, normalizedName);
-          let props = assign({}, injections, options);
-
-          runInDebug(() => {
-            let lazyInjections;
-            let validationCache = container.validationCache;
-            // Ensure that all lazy injections are valid at instantiation time
-            if (!validationCache[fullName] && factory && typeof factory._lazyInjections === 'function') {
-              lazyInjections = factory._lazyInjections();
-              lazyInjections = container.registry.normalizeInjectionsHash(lazyInjections);
-
-              container.registry.validateInjections(lazyInjections);
-            }
-
-            validationCache[fullName] = true;
-          });
-
-          this.class[NAME_KEY] = container.registry.makeToString(factory, fullName);
-
-          if (!this.class.create) {
-            throw new Error(`Failed to create an instance of '${normalizedName}'. Most likely an improperly defined class or` +
-                        ` an invalid module export.`);
-          }
-
-          if (this.class.prototype) {
-            injectDeprecatedContainer(this.class.prototype, container);
-          }
-
-          return this.class.create(props);
-        }
-      };
+      manager = new FactoryManager(this, factory, fullName, normalizedName);
     } else {
       let factory = this.lookupFactory(fullName, options);
       if (factory === undefined) { return; }
-
-      manager = {
-        class: factory,
-        create(props = {}) {
-          return instantiate(factory, props, container, fullName);
-        }
-      };
+      manager = new DeprecatedFactoryManager(this, factory, fullName);
     }
 
     runInDebug(() => {
@@ -221,7 +191,19 @@ Container.prototype = {
             throw new Error(`You attempted to set "${prop}" on a factory manager created by container#factoryFor. A factory manager is a read-only construct.`);
           }
         };
-        manager = new Proxy(manager, validator);
+
+        // Note:
+        // We have to proxy access to the manager here so that private property
+        // access doesn't cause the above errors to occur.
+        let m = manager;
+        let proxiedManager = {
+          class: m.class,
+          create(props) {
+            return m.create(props);
+          }
+        };
+
+        manager = new Proxy(proxiedManager, validator);
       }
     });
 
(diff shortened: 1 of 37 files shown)


rwjblue:
Haven't had a chance to review in detail, but I did notice that this still needs feature flagging.

chadhietala:
@stefanpenner I think this good for first round of review. Currently no feature flagging has been done because these are effectively internal changes with no user-land implications.

rwjblue:
> Currently no feature flagging has been done because these are effectively internal changes with no user-land implications.

O_o, isn't the whole point of the RFC to expose this as public API?

mixonic:
@chadhietala 

> inject the deprecated container in this new public API

I'm not quite sure what this means, the container is not deprecated. Will review here.

packages/container/lib/container.js
```diff
@@ -127,9 +135,86 @@ Container.prototype = {
...
+      if (!normalizedName) { return; }
+    }
+
+    let factory = this.registry.resolve(normalizedName);
+
+    if (factory === undefined) { return; }
+
+    let manager = {
+      class: factory,
+      create(options = {}) {
```
mixonic: I expected the implementation here to be

```js
let factory = this._lookupFactory(name);
return factory.create(options);
```

for the initial pass of `factoryFor` `create` per https://github.com/emberjs/rfcs/blob/master/text/0000-factoryFor.md#detailed-design. Additionally https://github.com/emberjs/rfcs/blob/master/text/0000-factoryFor.md#development-mode-proxy. In practice literally implementing as `_lookupFactory(name).create(props)` might not work, but it seems like it should and would preserve the double-extend behavior during instance creation. We should decide if this is really needed, will raise it in some conversations.

packages/container/lib/container.js
```diff
@@ -191,15 +280,47 @@ function lookup(container, fullName, options = {}) {
...
 
-  if (value === undefined) { return; }
+function isSingletonInstance(container, fullName, { instantiate, singleton }) {
+  return (isSingleton(container, fullName) && singleton !== false) && (shouldInstantiate(container, fullName) && instantiate !== false);
+}
 
-  if (isSingleton(container, fullName) && options.singleton !== false) {
-    container.cache[fullName] = value;
+function isFactoryClass(container, fullname, { instantiate, singleton }) {
+  return (!isSingleton(container, fullname) || singleton === false) && (!shouldInstantiate(container, fullname) && instantiate === false);
```
mixonic: putting the faster checks first seems ideal:

```js
function isFactoryClass(container, fullname, { instantiate, singleton }) {
  return (
    (singleton === false || !isSingleton(container, fullname)) &&
    (instantiate === false || !shouldInstantiate(container, fullname))
  );
}
```

thus avoiding the function calls in many cases.

packages/container/lib/container.js
```diff
@@ -127,9 +135,86 @@ Container.prototype = {
    */
   lookupFactory(fullName, options) {
     assert('fullName must be a proper full name', this.registry.validateFullName(fullName));
+
+    if (isFeatureEnabled('container-factoryFor')) {
+      deprecate('Using "_lookupFactory" is deprecated. Please use container.factoryFor instead.', false, { id: 'container-lookupFactory', until: '2.12.0', url: 'TODO' });
+    }
+
     return factoryFor(this, this.registry.normalize(fullName), options);
```
mixonic: can this function be renamed now, since it is not the `factoryFor` one would normally be looking for here.

packages/container/lib/container.js
```diff
@@ -191,15 +280,47 @@ function lookup(container, fullName, options = {}) {
     return container.cache[fullName];
   }
 
-  let value = instantiate(container, fullName);
+  return instantiateFactory(container, fullName, options);
```
mixonic: So this means`container.lookup` is now based on the the non-double-extended factory? This seems like the behavior we didn't intend to change until `_lookupFactory` is removed.

packages/container/lib/container.js
```diff
@@ -127,9 +135,86 @@ Container.prototype = {
    */
   lookupFactory(fullName, options) {
     assert('fullName must be a proper full name', this.registry.validateFullName(fullName));
+
+    if (isFeatureEnabled('container-factoryFor')) {
+      deprecate('Using "_lookupFactory" is deprecated. Please use container.factoryFor instead.', false, { id: 'container-lookupFactory', until: '2.12.0', url: 'TODO' });
```
mixonic: Must be `until: '2.13'`. 2.12 is an LTS release and must contain this deprecation.

packages/container/tests/container_test.js
```diff
@@ -15,6 +15,19 @@ QUnit.module('Container', {
   }
 });
 
+function lookupFactory(name, container, options) {
+  let factory;
+  if (isFeatureEnabled('container-factoryFor')) {
+    expectDeprecation(() => {
```
mixonic: There is also an `ignoreDeprecation` which might be appropriate/useful here. Example:

https://github.com/emberjs/ember.js/blob/aeeddf7c00b42a340494465252481b5414d87009/packages/ember-runtime/tests/controllers/controller_test.js#L118

packages/container/tests/container_test.js
```diff
@@ -561,7 +581,8 @@ QUnit.test('A deprecated `container` property is appended to every object instan
   }, 'Using the injected `container` is deprecated. Please use the `getOwner` helper instead to access the owner of this object.');
 });
 
-QUnit.test('A deprecated `container` property is appended to every object instantiated from a non-extendable factory, and a fake container is available during instantiation.', function() {
+// This is testing that container was passed as an option
+QUnit.skip('A deprecated `container` property is appended to every object instantiated from a non-extendable factory, and a fake container is available during instantiation.', function() {
```
mixonic: This test shows why we need to keep the "double extend" the default path for `lookup` until at least 2.13.

packages/container/lib/container.js
```diff
@@ -127,9 +135,86 @@ Container.prototype = {
...
+              throw new Error(`You attempted to access "${prop}" on a factory manager created by container#factoryFor. "${prop}" is not a member of a factory manager."`);
+            }
+
+            return obj[prop];
+          },
+          set(obj, prop, value) {
+            throw new Error(`You attempted to set "${prop}" on a factory manager created by container#factoryFor. A factory manager is a read-only construct.`);
+          }
+        };
+        manager = new Proxy(manager, validator);
```
mixonic: Would like to have tests included that:

* getting a property that is not `class` or `create` throws an error
* setting a property throws an error
* calling `instanceof` with the return value from `factoryFor` throws an error

Per: https://github.com/emberjs/rfcs/blob/master/text/0000-factoryFor.md#development-mode-proxy

mixonic (changes requested):
Thanks so much for getting this going @chadhietala. The big item is that `lookup` should still be using double extend. If you want to add a second feature flag (ideally in a followup) that experimentally switches `lookup` to the non-double-extended factory, that would definitely be welcome as a distinct feature flag.

packages/container/lib/container.js
chadhietala: Not sure if I follow. Can we not just transition all the internal stuff over to `factoryFor`?

packages/ember-extension-support/lib/data_adapter.js
```diff
@@ -163,7 +163,7 @@ export default EmberObject.extend({
 
   _nameToClass(type) {
     if (typeof type === 'string') {
-      type = getOwner(this)._lookupFactory(`model:${type}`);
+      type = getOwner(this).factoryFor(`model:${type}`).class;
```
mixonic: It seems there is some use of the types in Ember-Data. For example [here attributes are read off the class](https://github.com/emberjs/data/blob/c9d8212c857ca78218ad98d11621819b38dba98f/addon/-private/system/debug/debug-adapter.js#L32). Removing the double-extend means the `attributes` CP would be called and memoized on the base class, and thus shared between test runs etc.

This is definitely a change, described [in the design section](https://github.com/emberjs/rfcs/blob/master/text/0000-factoryFor.md#design-of-applicationinstancefactoryfor).

Some other examples of libraries this would impact are listed [in the drawbacks section](https://github.com/emberjs/rfcs/blob/master/text/0000-factoryFor.md#drawbacks).

packages/container/lib/container.js
mixonic: The intent was to encourage libs/apps to use the new API and drop their current reliance on the double-extended class, then change over internals and any other small breaking changes that will happen after most apps have refactored toward `factoryFor` and stopped expecting a double-extend.

chadhietala:
I made some changes to this, but likely needs further guidance.

stefanpenner:
reviewing now

packages/container/lib/container.js
```diff
@@ -127,7 +136,96 @@ Container.prototype = {
...
+              lazyInjections = factory._lazyInjections();
+              lazyInjections = container.registry.normalizeInjectionsHash(lazyInjections);
+
+              container.registry.validateInjections(lazyInjections);
+            }
+
+            validationCache[fullName] = true;
+          });
+
+          this.class[NAME_KEY] = container.registry.makeToString(factory, fullName);
```
stefanpenner: I don't believe we want to mutate the `class`, note: it is perfectly legal to register a class multiple times with different injection rules/similar for each. When this occurs, nothing should collide and currently NAME_KEY would collide. I believe NAME_KEY can be provided merely as a member of `props`. I believe in most cases this should work IFF

https://github.com/emberjs/ember.js/blob/master/packages/ember-runtime/lib/system/core_object.js#L546

changed to:

```js
var ret = '<' + (this[NAME_KEY] || this.constructor.toString()) + ':' + guidFor(this) + extension + '>';
```

This would mean, `class.toString()` wouldn't be terribly useful, but I don't believe it can be accomplished

packages/container/lib/container.js
```diff
@@ -127,7 +136,96 @@ Container.prototype = {
...
+
+          if (!this.class.create) {
+            throw new Error(`Failed to create an instance of '${normalizedName}'. Most likely an improperly defined class or` +
+                        ` an invalid module export.`);
+          }
+
+          if (this.class.prototype) {
+            injectDeprecatedContainer(this.class.prototype, container);
```
stefanpenner: I don't believe we want to mutate the class

packages/container/lib/container.js
stefanpenner: yes, we should only have 1 behavior in the system at any point in time, once we cut over it should be all no double extend.

packages/container/lib/container.js
```diff
@@ -350,7 +474,7 @@ function instantiate(container, fullName) {
 
     if (typeof factory.extend === 'function') {
```
stefanpenner: does this refer to the `factory.class` or our new factory object, if it is our new factory object this may no longer make sense?

packages/container/lib/container.js
```diff
@@ -374,6 +498,19 @@ function instantiate(container, fullName) {
...
 
+function factoryInjectionsFor(container, fullName) {
+  let registry = container.registry;
+  let splitName = fullName.split(':');
+  let type = splitName[0];
+
+  let factoryInjections = buildInjections(container,
+                                          registry.getFactoryTypeInjections(type),
+                                          registry.getFactoryInjections(fullName));
+  factoryInjections._debugContainerKey = fullName;
```
stefanpenner: @mixonic  / @rwjblue should this become a symbol or something in the future?

mixonic: In the future perhaps. I don't think it should be changed in this effort.

packages/ember-routing/lib/system/generate_controller.js
```diff
@@ -17,7 +17,7 @@ import {
 */
 
 export function generateControllerFactory(owner, controllerName, context) {
-  let Factory = owner._lookupFactory('controller:basic').extend({
+  let Factory = owner.factoryFor('controller:basic').class.extend({
```
stefanpenner: if previously, injections where set on `controller:basic` they would now be lost (as we bypass DI if we do `.class`

packages/ember-routing/lib/system/route.js
```diff
@@ -1631,16 +1632,15 @@ let Route = EmberObject.extend(ActionHandler, Evented, {
...
         assert(
           `You used the dynamic segment ${name}_id in your route ${routeName}, but ${namespace}.${StringUtils.classify(name)} did not exist and you did not override your route's \`model\` hook.`, !!modelClass);
 
         if (!modelClass) { return; }
 
-        assert(`${StringUtils.classify(name)} has no method \`find\`.`, typeof modelClass.find === 'function');
+        assert(`${StringUtils.classify(name)} has no method \`find\`.`, typeof modelClass.class.find === 'function');
 
-        return modelClass.find(value);
+        return modelClass.class.find(value);
```
stefanpenner: In a future iteration, we should consider deprecating this feature. it is... dubious at best (and was my fault...)

packages/ember-routing/lib/system/router.js
```diff
@@ -585,7 +585,7 @@ const EmberRouter = EmberObject.extend(Evented, {
       seen[name] = true;
 
       if (!handler) {
-        let DefaultRoute = routeOwner._lookupFactory('route:basic');
+        let DefaultRoute = routeOwner.factoryFor('route:basic').class;
```
stefanpenner: similar to the above `controller:basic` example, if injections existed on `route:basic` at this point they would be lost, which I believe is a breaking change.

stefanpenner (dismissed):

packages/ember-routing/lib/system/router.js
chadhietala: Same question as above.

packages/ember-routing/lib/system/generate_controller.js
mixonic: @chadhietala and I went over this. The change as it stands would indeed squash DI in the case where there was an injection on `controller:basic` and you expected it to show up on `controller:generated-controller`. Of course that whole concept, that your DI onto `controller:basic` will show up on `controller:generated-controller`, is pretty WAT. But we should preserve it if possible.

packages/ember-routing/lib/system/generate_controller.js
```diff
@@ -17,7 +17,7 @@ import {
...
     isGenerated: true,
```
mixonic: `isGenerated` can be killed off.

packages/ember-routing/lib/system/router.js
mixonic: If we care about injections on `route:basic` being available on all generated routes, we can take the current implementation:

```js
let DefaultRoute = routeOwner._lookupFactory('route:basic');
routeOwner.register(fullRouteName, DefaultRoute.extend());
handler = routeOwner.lookup(fullRouteName);
// Later on may lookup this name again
```

And hack a version that should do the double-DI (DI for `route:basic` and `route:generated-thing` with the current implementation:

```js
let factory = routeOwner.factoryFor('route:basic');
let DefaultRouteProxy = {
  extend(...diArgs) {
    return {
      create(...createArgs) {
         return factory.create(assign({}, ...diArg, ...createArgs));
      }
    };
  },
  create(...createArgs) {
    /* Not sure this is ever going to be used in the current container double-extend implementation */
    return factory.create(assign({}, ...createArgs));
  }
}
```

However in 99.99999% of cases I don't expect someone is doing an injection on `route:basic` (or `controller:basic`), instead they are probably doing injections on the type. To keep these cases fast, we can check the injections list and if possible simple re-register the class:

```js
let DefaultRoute = routeOwner.factoryFor('route:basic').class;
routeOwner.register(fullRouteName, DefaultRoute);
handler = routeOwner.lookup(fullRouteName);
```

In this case DI on the type still works.

packages/container/lib/container.js
```diff
@@ -466,3 +575,56 @@ function buildFakeContainerFunction(container, containerProperty, ownerProperty)
...
+        lazyInjections = this.class._lazyInjections();
+        lazyInjections = this.container.registry.normalizeInjectionsHash(lazyInjections);
+
+        this.container.registry.validateInjections(lazyInjections);
+      }
+
+      validationCache[this.fullName] = true;
+    });
+
+    this.class[NAME_KEY] = this.container.registry.makeToString(this.class, this.fullName);
```
mixonic: It does looks like this should be added to the injections [per master](https://github.com/emberjs/ember.js/blob/e601285bed0e2e108d41a68a61dc2098aac26dd7/packages/container/lib/container.js#L273) instead of set on the class.

packages/container/lib/container.js
mixonic: This refers to `extend` on the registered class, not `extend` on the factory object. The idea being `extend` is supported, injections have been made.

`instantiate` is only called by `lookup` though, and in that case we want the double extend to continue for now so this test would hold true.

mixonic:
Summary comment of our planning:

* [ ] There should be two feature flags. One adding `factoryFor` with the legacy double-extend implementation, one enabling no-more-double-extend mode.
* [ ] `isGenerated` is no longer needed as a flag on generated classes. (@rwjblue sanity check me?)
* [ ] In the `factoryFor` documentation we should note that by using `factoryFor` the consuming code takes responsibility for destroying the instance.
* [ ] It seems we can solve the injections-on-generated-singletons (routes, controllers) issue by continuing to use the register hack already in master, but we should also implement an above noted meta-hack to keep injections on `route:basic` and `controller:basic` showing up on all generated instances. I actually think this behavior is a) wrong and b) not relied upon and would prefer to just nuke it, but it warrants discussion.
* [ ] `NAME_KEY` needs fixin'
* [ ] `_nameToClass` in the data_adapter seems like it should return either the double extended class or the registered class depending on if the second no-more-double-extend feature flag is enabled.

rwjblue:
> isGenerated is no longer needed as a flag on generated classes. (@rwjblue sanity check me?)

Confirm.

packages/ember-routing/lib/system/router.js
chadhietala: So this doesn't seem to work for generated controllers due to the fact `.proto()` called right after we generate the controller. https://github.com/emberjs/ember.js/blob/192017b71070d5a32f24ec215ad75dd702d0c0bc/packages/ember-routing/lib/system/route.js#L193

chadhietala:
@mixonic so I think there is some problems with the last 2 items. The `_nameToClass` item above can't actually be tested sense flags are turned on in binary fashion. I can put the guard in there but we are basically flying blind on the `no-double-extend` path. The generated controllers and routes almost work however in the controller case we try to call `proto()` on the generated class so doing the proxying class hack doesn't really work.

packages/container/lib/container.js
```diff
@@ -28,6 +39,9 @@ export default function Container(registry, options) {
   this.isDestroyed = false;
 }
 
+Container.__FACTORY_FOR__ = FACTORY_FOR;
```
chadhietala: This seems bad but don't know how to make the Node tests pass without doing this.

packages/container/lib/container.js
```diff
@@ -127,7 +141,73 @@ Container.prototype = {
    */
   lookupFactory(fullName, options) {
     assert('fullName must be a proper full name', this.registry.validateFullName(fullName));
-    return factoryFor(this, this.registry.normalize(fullName), options);
+
+    if (isFeatureEnabled('ember-no-double-extend')) {
```
mixonic: I think this check should be `ember-factory-for`?

Additionally it seems like you would pass the boolean as an argument to `deprecate`, doesn't really matter though :-p

packages/container/lib/container.js
```diff
@@ -175,10 +255,31 @@ Container.prototype = {
   }
 };
 
+if (isFeatureEnabled('ember-factory-for')) {
+  /**
+   Given a fullName, return the corresponding factory. The consumer of factory is repsonsible
```
mixonic: "of factory" -> "of the factory"

"repsonsible" -> "responsible"

packages/container/lib/container.js
```diff
@@ -175,10 +255,31 @@ Container.prototype = {
...
+   for the destruction of the factory as there is no way to understand the objects lifecycle.
```
mixonic: "for the destruction of the factory" -> "for the destruction of any factory instances, as there is no way for the container to ensure instances are destroyed when it itself is destroyed"

packages/container/tests/container_test.js
```diff
@@ -15,6 +15,19 @@ QUnit.module('Container', {
   }
 });
 
+function lookupFactory(name, container, options) {
+  let factory;
+  if (isFeatureEnabled('ember-no-double-extend')) {
+    expectDeprecation(() => {
```
mixonic: `ignoreDeprecation` may be more appropriate for a test utility method.

packages/container/tests/container_test.js
```diff
@@ -465,15 +481,19 @@ QUnit.test('factory for non extendables resolves are cached', function() {
...
 
-  container.lookupFactory('foo:post');
+  lookupFactory('foo:post', container);
   deepEqual(resolveWasCalled, ['foo:post']);
 });
 
 QUnit.test('The `_onLookup` hook is called on factories when looked up the first time', function() {
-  expect(2);
+  if (isFeatureEnabled('ember-factory-for') && isFeatureEnabled('ember-no-double-extend')) {
+    expect(4);
```
mixonic: yeah `ignoreDeprecation` would allow you to not have this change in expect count for an unrelated test.

packages/ember-application/tests/system/dependency_injection/default_resolver_test.js
```diff
@@ -64,9 +65,15 @@ QUnit.test('the default resolver looks up templates in Ember.TEMPLATES', functio
...
 
-  equal(locator.lookupFactory('template:foo'), fooTemplate, 'resolves template:foo');
-  equal(locator.lookupFactory('template:fooBar'), fooBarTemplate, 'resolves template:foo_bar');
-  equal(locator.lookupFactory('template:fooBar.baz'), fooBarBazTemplate, 'resolves template:foo_bar.baz');
+  if (isFeatureEnabled('ember-factory-for')) {
+    equal(locator.factoryFor('template:foo').class, fooTemplate, 'resolves template:foo');
+    equal(locator.factoryFor('template:fooBar').class, fooBarTemplate, 'resolves template:foo_bar');
+    equal(locator.factoryFor('template:fooBar.baz').class, fooBarBazTemplate, 'resolves template:foo_bar.baz');
+  } else {
+    equal(locator.lookupFactory('template:foo'), fooTemplate, 'resolves template:foo');
```
mixonic: It is concerning that this test coverage is lost when the feature flag is enabled?

packages/ember-glimmer/lib/syntax/render.js
```diff
@@ -186,8 +187,9 @@ class NonSingletonRenderManager extends AbstractRenderManager {
   create(environment, definition, args, dynamicScope) {
     let { name, env } = definition;
     let modelRef = args.positional.at(0);
+    let controllerFactory = env.owner[FACTORY_FOR](`controller:${name}`);
```
mixonic: Doesn't using factoryFor here mean the double extended class is never used for `render`, regardless of the feature flag setting?

mixonic (changes requested):

mixonic (approved):

mixonic:
OMG this is good to go. To merge:

* One of us will rebase this down to a more meaningful commit history with `[FEATURE factory-for]` in the commits
* I would like to boot the build with various flags enabled/disabled in a real world app before merge. Just a quick sanity check.

Excited :-D

chadhietala:
Quick benchmarks from Travis-Web. This is consistent across runs and just looking at the homepage that is likely pretty static.

__Before: (Prod / Feature Flags Off)__
screen shot 2016-12-17 at 7 31 51 pm

__After: (Prod / Feature Flags On)__
screen shot 2016-12-17 at 7 32 51 pm

rwjblue:
@chadhietala - I assume the `After: (Prod / Feature Flags On)` screenshots above are with both `ember-factory-for` and `ember-no-double-extend` flags on, is that right?  Can you test with only `ember-factory-for` (I'm mostly looking to try to figure out if we will have temporary regressions in 2.12)?

packages/ember-application/lib/system/engine-instance.js
```diff
@@ -194,9 +194,25 @@ const EngineInstance = EmberObject.extend(RegistryProxyMixin, ContainerProxyMixi
 
     this.inject('view', '_environment', '-environment:main');
     this.inject('route', '_environment', '-environment:main');
+  },
+
+  [FACTORY_FOR](fullName, options) {
```
rwjblue: Lets make this part of the `ContainerProxyMixin`, this would simplify issues in `ember-test-helpers` land and actually keep the logic in the correct place (currently `.lookup` and `._lookupFactory` are done in `ContainerProxyMixin`).

packages/ember-application/lib/system/engine-instance.js
```diff
@@ -194,9 +194,25 @@ const EngineInstance = EmberObject.extend(RegistryProxyMixin, ContainerProxyMixi
...
+  },
+
+  [LOOKUP_FACTORY](fullName, options) {
+    return this.__container__[LOOKUP_FACTORY](fullName, options);
   }
 });
 
+if (isFeatureEnabled('ember-factory-for')) {
+  EngineInstance.reopen({
+    factoryFor(fullName, options) {
```
rwjblue: This can be moved to the mixin also...

rwjblue (dismissed):

packages/container/lib/container.js
```diff
@@ -1,8 +1,19 @@
...
+  OWNER,
+  assign,
+  NAME_KEY
+} from 'ember-utils';
 import { ENV } from 'ember-environment';
-import { assert, deprecate, runInDebug } from 'ember-metal';
+import { assert, deprecate, runInDebug, isFeatureEnabled } from 'ember-metal';
 
 const CONTAINER_OVERRIDE = symbol('CONTAINER_OVERRIDE');
+const HAS_PROXY = typeof Proxy === 'function';
```
rwjblue: Can we move this into `ember-utils` and export as `HAS_NATIVE_PROXY`?  Very similar to what was done in https://github.com/emberjs/ember.js/pull/14649 for `packages/ember-utils/lib/weak-map-utils.js`.

tests/node/helpers/build-owner.js
```diff
@@ -1,5 +1,14 @@
 module.exports = function buildOwner(Ember, resolver) {
-  var Owner = Ember.Object.extend(Ember._RegistryProxyMixin, Ember._ContainerProxyMixin);
+  var FACTORY_FOR = Ember.Container.__FACTORY_FOR__;
```
rwjblue: See comments in https://github.com/emberjs/ember.js/pull/14360#pullrequestreview-13465262 and https://github.com/emberjs/ember-test-helpers/pull/191, we should be able to remove this.

mixonic:
I've done some benchmarks against todomvc. I would expect the impact on a larger app to be much more significant (as @chadhietala's profiling shows).

This PR may have a minor performance regression:

screen shot 2016-12-18 at 2 19 48 pm

The `ember-factory-for` feature flag does not seem to have a significant impact on performance (which is expected).

However the `ember-no-double-extend` feature flag makes a significant improvement.

screen shot 2016-12-18 at 2 22 14 pm

I think we can continue to explore issues around backwards compatibility, performance, and the migration path to `factoryFor` on master. Merging this.

@chadhietala Thanks much for your patience and push on this gnarly, complex, cross-cutting part of Ember's API. The payoff (especially in big apps) is going to be really notable.

stefanpenner:
👍

sandstrom:
Awesome! Nice to be able to move into a public API for this + great performance improvements! ⛵️

PR #14361, opened 2016-09-27
Cleanup fastboot hooks and misc view cleanup

### LINKED ISSUE #14353: Fix fastboot hooks
* only call attr hooks in non-interactive mode
* throw when `element` is accessed
* add a private symbol to return the `element` regardless of the mode
* (for future) add a linting rule (see core team notes for details)

chancancode:
Fixes #14353


rwjblue:
Thank you @chancancode!

Issue #14358, opened 2016-09-26
[glimmer2] Attempted to register a view with an id already in use


HeroicEric:
Given a tagless component that accepts an `id` attribute that is passed down to a child component:

```js
// app/components/my-link-component.js
import Ember from 'ember';

export default Ember.Component.extend({
  tagName: ''
});
```

```hbs
{{! app/templates/components/my-link-component.hbs}}
{{#link-to "index" id=id}}
  {{yield}}
{{/link-to}}
```

When that component is used in a template and given an `id` attribute:

```hbs
{{#my-link-component id="foo"}}
  My link
{{/my-link-component}}
```

The following assertion fails:

```
Uncaught Error: Assertion Failed: Attempted to register a view with an id already in use: foo
```

This worked in previous versions but fails in ember 2.9. Same thing happens with other child components, not just the `{{link-to}}` component.

[Working Ember 2.8 twiddle](https://ember-twiddle.com/17f6da8ff187bf1988c3f60b760d37b8/39361a4d4226ffe9283e553755e02cac3fddb0a1)
[Failing Ember 2.9 twiddle](https://ember-twiddle.com/17f6da8ff187bf1988c3f60b760d37b8)

Note: The assertion in ember-twiddle shows up if you open the dev tools console

runspired:
@HeroicEric I believe this was intentional, I know @ef4 refactored liquid-fire due to it.

HeroicEric:
hmm, ok. I see this refactoring you're talking about for liquid-fire in https://github.com/ember-animation/liquid-fire/pull/481/commits/fd7ffa64fb00831f257369b2e31dc3db5fba3f3f

It feels awkward to have to come up with a different name for the `id` of the element, especially with `elementId` also not allowed 🐼 

I'll leave this open until somebody confirms that this was intentional.

rwjblue:
The solution here would be to change how getViewId works for tagless views (instead of using elementId we would always use guidFor). 

@heroiceric - I'm not sure I'll be able to get to this before 2.9 (or if it is a blocker), but I'd love a PR with a failing test (and the fix if it's as simple as what I said above) if you have time...

PR #14363, opened 2016-09-27
[BUGFIX beta] Always use `guidFor` for tagless component id

HeroicEric:
Fixes https://github.com/emberjs/ember.js/issues/14358


HeroicEric:
@rwjblue not sure what the deal is with the sauce failure but here you go 😄

rwjblue:
Restarted sauce job

rwjblue:
Thanks for fixing @HeroicEric!

PR #14364, opened 2016-09-27
Remove unnecessary import of `metaFor`.

dgeb:
`watchKey` internally calls `metaFor` if no `meta` has been provided.

### DIFF
--- a/packages/ember-glimmer/lib/utils/references.js
+++ b/packages/ember-glimmer/lib/utils/references.js
@@ -4,7 +4,6 @@ import {
   set,
   tagFor,
   didRender,
-  meta as metaFor,
   watchKey,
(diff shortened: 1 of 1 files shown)


stefanpenner:
> watchKey internally calls metaFor if no meta has been provided.

I believe the idea is to avoid calls to metaFor if possible. This seems to force us to always call metaFor.

dgeb:
@stefanpenner I don't believe this PR will affect the number of calls to `metaFor`. The intent of this PR is to just avoid the extra import and duplicate calls, [which will always be made from `watchKey` if no `meta` has been provided](https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/lib/watch_key.js#L15).

PR #14377, opened 2016-09-28
[DOC release] Publicly document the `readonly` helper

### LINKED ISSUE #14376: Htmlbars readonly helper not documented
The `readonly` helper as described in http://emberup.co/bindings-with-htmlbars-helpers/ is not documented in http://emberjs.com/api/classes/Ember.Templates.helpers.html. Is it private?

locks:
Fixed #14376.

This serves to document the `readonly` helper with the goal of surfacing it in the API documentation.

### DIFF
--- a/packages/ember-glimmer/lib/helpers/readonly.js
+++ b/packages/ember-glimmer/lib/helpers/readonly.js
@@ -5,6 +5,101 @@
 import { UPDATE } from '../utils/references';
 import { unMut } from './mut';
 
+
+/**
+  The `readonly` helper let's you specify that a binding is one-way only,
+  instead of two-way.
+  When you pass a `readonly` binding from an outer context (e.g. parent component),
+  to to an inner context (e.g. child component), you are saying that changing that
+  property in the inner context does not change the value in the outer context.
+
+  To specify that a binding is read-only, when invoking the child `Component`:
+
+  ```app/components/my-parent.js
+  export default Component.extend({
+    totalClicks: 3
+  });
+  ```
+
+  ```app/templates/components/my-parent.hbs
+  {{log totalClicks}} // -> 3
+  {{my-child childClickCount=(readonly totalClicks)}}
+  ```
+
+  Now, when you update `childClickCount`:
+
+  ```app/components/my-child.js
+  export default Component.extend({
+    click() {
+      this.incrementProperty('childClickCount');
+    }
+  });
+  ```
+
+  The value updates in the child component, but not the parent component:
+
+  ```app/templates/components/my-child.hbs
+  {{log childClickCount}} //-> 4
+  ```
+
+  ```app/templates/components/my-parent.hbs
+  {{log totalClicks}} //-> 3
+  {{my-child childClickCount=(readonly totalClicks)}}
+  ```
+
+  ### Objects and Arrays
+
+  When passing a property that is a complex object (e.g. object, array) instead of a primitive object (e.g. number, string),
+  only the reference to the object is protected using the readonly helper.
+  This means that you can change properties of the object both on the parent component, as well as the child component.
+  The `readonly` binding behaves similar to the `const` keyword in JavaScript.
+
+  Let's look at an example:
+
+  First let's set up the parent component:
+
+  ```app/components/my-parent.js
+  export default Ember.Component.extend({
+    clicks: null,
+
+    init() {
+      this._super(...arguments);
+      this.set('clicks', { total: 3 });
+    }
+  });
+  ```
+
+  ```app/templates/components/my-parent.hbs
+  {{log clicks.total}} //-> 3
+  {{my-child childClicks=(readonly clicks)}}
+  ```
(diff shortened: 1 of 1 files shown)


packages/ember-glimmer/lib/helpers/readonly.js
```diff
@@ -5,6 +5,101 @@
...
+  {{my-child childClickCount=(readonly totalClicks)}}
+  ```
+  
+  ### Objects and Arrays
+  
+  Note that while properties passed in as `readonly` parameters won't update
+  the property itself, if it is an Object or an Array, you can still modify
+  properties of it and it'll update in the parent. You can think of this as
+  working similarly to the `const` keyword introduced in ES6. Let's look at
+  an example:
```
pixelhandler: What do you think about:

When passing property that is a complex object vs. a primitive object, only the reference to the object is protected using the `readonly` helper. The contents of a bound property that is an `Object` or `Array` can be changed in both the `outerProperty` and the `componentProperty` simultaneously. This read only (or one-way) binding behaves similar to the `const` keyword in ES2015.

See the following example:

*Suggestion: use "ES2105" instead of "ES6" (ECMAScript 6 (ES6)/ ECMAScript 2015 (ES2015): The 6th edition of ECMAScript, standardized in 2015)*

pixelhandler:
@locks I think this also fixes #14377

locks:
@pixelhandler that's the number of the PR, did you mean #14378? ;)

Serabe:
@locks: can you fix the trailling whitespaces so CI would not complain? Thank you for your work!

packages/ember-glimmer/lib/helpers/readonly.js
```diff
@@ -5,6 +5,101 @@
 import { UPDATE } from '../utils/references';
 import { unMut } from './mut';
 
+
+/**
+  The `readonly` helper let's you specify that when you updated a value passed in
+  to a child component, it does not update in the parent component.
```
pixelhandler: `this` scope is indeed scary

rwjblue:
@locks - What else do we need here?

locks:
@rwjblue I want to wordsmith a bit more. I'm reviewing my documentation PRs this weekend, and let you know.

pixelhandler:
👍 @locks

packages/ember-glimmer/lib/helpers/readonly.js
```diff
@@ -5,6 +5,101 @@
...
+  ```javascript
+  // my-child.js
+  export default Component.extend({
+    click() {
+      this.incrementProperty('childClickCount');
+    }
+  });
+  ```
+  
+  The value updates in the child component, but not the parent component:
```
mixonic: I think the current text is good, or at the least if something like the suggested change was used it should be specific to the property names used in the example.

packages/ember-glimmer/lib/helpers/readonly.js
```diff
@@ -5,6 +5,96 @@
...
+  {{log totalClicks}} //-> 3
+  {{my-child childClickCount=(readonly totalClicks)}}
+  ```
+  
+  ### Objects and Arrays
+  
+  When passing a property that is a complex object (e.g. object, array) instead of a primitive object (e.g. number, string),
+  only the reference to the object is protected using the readonly helper.
+  This means that you can change properties of the object both on the parent component, as well as the child component.
+  This read-only binding behaves similar to the const keyword in ES2015.
```
mixonic: `This read-only binding` makes it sound like object bindings are different than normal read-only bindings. There is only one read-only binding definition. I think more accurate, at the least, would be:

> Read-only binding behaves similar to the `const` keyword in JavaScript

ES2015 is probably overly technical. It is simply a JS feature.

packages/ember-glimmer/lib/helpers/readonly.js
```diff
@@ -5,6 +5,96 @@
...
+  
+  Let's look at an example:
+  
+  First let's set up the parent component:
+  
+  ```app/components/my-parent.js
+  export default Ember.Component.extend({
+    clicks: { total: 3 }
```
mixonic: It would be ideal to use an example that didn't store data on the prototype. Doing so is an anti-pattern, as two of these components would share their `total` value.

packages/ember-glimmer/lib/helpers/readonly.js
mixonic: Additionally Glimmer has a concept of `const`-ness, and that may imply the addition of a `const` helper later on. This would be different than `readonly` though, it would enforce immutability in the template layer and presume a value is static after initial render.

locks:
@rwjblue @mixonic updated

rwjblue:
Per core team meeting we are landing these docs as private for now. It is unclear if this helper will be needed for a migration / interop story with angle bracket components. When we have an RFC for that we can re-evaluate this helper to see if it should be public.

Issue #14369, opened 2016-09-28
tag-less component with bound id attribute throwing an error


ctusch:
I ran into this issue while updating our application from Ember 2.4.2 to 2.8.0. Suddenly our subroutes wouldn't render anymore. Turns out that while you can assign an id to a tag-less component, the id's value must be a constant. Unfortunately in our applicaton the id is bound to a property.

There is a related issue (and PR) by @rwjblue that fixed this behavior for constant values: #14105 

Here is a [twiddle](https://ember-twiddle.com/7d531ee89c0b39c6aeaa0d71b843c682?openFiles=templates.application.hbs%2Ctemplates.components.my-component.hbs) for easy reproduction. The error text reads:
> ember.debug.js:6778 Uncaught Error: Assertion Failed: You cannot use `elementId` on a tag-less component: <twiddle@component:my-component::ember354>

The assertion can be found [here](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/syntax/curly-component.js#L53).

HeroicEric:
just wanted to add that this seems to be fixed in beta and canary

ctusch:
Interesting, looking at the callstack and searching the sources the whole infrastructure changed. Does this have something to do with Glimmer 2?

chancancode:
@HeroicEric as I mentioned yesterday, we should probably have a separate test for this scenario (id is used as a regular bound property, but not passed down to a child's elementId). Do you have time to tackle that? :smile:

HeroicEric:
@chancancode I think I missed the conversation you're referring to but I can look into it 😄

pixelhandler:
@HeroicEric should this be closed as #14382 was merged?

rwjblue:
Yeah, closing this (as fixed in release branch).

PR #14382, opened 2016-09-29
[BUGFIX release] Avoid assertion when bound `id` provided to tagless component

HeroicEric:
Fixes https://github.com/emberjs/ember.js/issues/14369

I'm not totally sure about the protocol for submitting this change. The code fix is for code that is totally gone in 2.9+ but the test might still be useful to have around in later versions.

@chancancode is this what you were thinking?

### DIFF
--- a/packages/ember-glimmer/tests/integration/components/fragment-components-test.js
+++ b/packages/ember-glimmer/tests/integration/components/fragment-components-test.js
@@ -165,6 +165,19 @@ moduleFor('Components test: fragment components', class extends RenderingTest {
     this.assertText('baz');
   }
 
+  ['@test does not throw an error if `tagName` is an empty string and `id` is bound property specified via template']() {
+    let template = `{{id}}`;
+    let FooBarComponent = Component.extend({
+      tagName: ''
+    });
+
+    this.registerComponent('foo-bar', { ComponentClass: FooBarComponent, template });
+
+    this.render(`{{#foo-bar id=fooBarId}}{{/foo-bar}}`, { fooBarId: 'baz' });
+
+    this.assertText('baz');
+  }
+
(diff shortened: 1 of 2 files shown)


rwjblue:
Eeck, good catch. Yes, can you PR this test against master too?

packages/ember-glimmer/tests/integration/components/fragment-components-test.js
```diff
@@ -165,6 +165,19 @@ moduleFor('Components test: fragment components', class extends RenderingTest {
...
+    let template = `{{id}}`;
+    let FooBarComponent = Component.extend({
+      tagName: ''
+    });
+
+    this.registerComponent('foo-bar', { ComponentClass: FooBarComponent, template });
+
+    this.render(`{{#foo-bar id=fooBarId}}{{/foo-bar}}`, { fooBarId: 'baz' });
+
+    this.assertText('baz');
```
chancancode: Can you test updating the id and resetting it back to the original value? (The point of this test is to make sure we don't throw when you update the id value on a tag less component). :smile:

HeroicEric: Yes. Should the other tests in this file be doing the same thing?

ksnyde:
Just ran into this ... quite desperate to get this into 2.8.x. By being in "release" does that mean it'll make the next release in a few days? Is there any workaround until then?

For me at least, this is a big deal because I am developing an addon and the default config for addons is broken in Inspector for 2.9.0 so I can't develop there either [issue](https://github.com/emberjs/ember-inspector/issues/605).

rwjblue:
@ksnyde - Yes, this PR targeted the release branch (and you can use this channels build today if you'd like), and it will be included in 2.8.2 when we release it (likely Monday).

PR #14383, opened 2016-09-29
Move objects guid into Meta.

rwjblue:
- Refactor `Ember.guidFor` to always return a number (avoid extraneous string allocations, easier to compare, etc)
- Move guid storage from a property on the object (using `obj.__defineNonEmumerable` or `Object.defineProperty`) into the meta instance for the object. Where supported this will use a native `WeakMap`, where not supported this results in only a single property on the object instead of two (`GUID_KEY` and `__ember_meta__`).
- Maintain support for `Ember.generateGuid` (even though private)
- Maintain support for `Ember.guidFor`

### DIFF
--- a/packages/ember-metal/lib/guid.js
+++ b/packages/ember-metal/lib/guid.js
@@ -1,4 +1,4 @@
-import { intern } from 'ember-utils';
+import { meta as metaFor } from './meta';
 
 /**
  Previously we used `Ember.$.uuid`, however `$.uuid` has been removed from
@@ -31,41 +31,6 @@ const OBJECT_UUID = uuid();
 const ARRAY_UUID = uuid();
 const NULL_UUID = uuid();
 const UNDEFINED_UUID = uuid();
-
-/**
-  A unique key used to assign guids and other private metadata to objects.
-  If you inspect an object in your browser debugger you will often see these.
-  They can be safely ignored.
-
-  On browsers that support it, these properties are added with enumeration
-  disabled so they won't show up when you iterate over your properties.
-
-  @private
-  @property GUID_KEY
-  @for Ember
-  @type String
-  @final
-*/
-export const GUID_KEY = intern('__ember' + (+ new Date()));
-
-export let GUID_DESC = {
-  writable:     true,
-  configurable: true,
-  enumerable:   false,
-  value: null
-};
-
-let nullDescriptor = {
-  configurable: true,
-  writable: true,
-  enumerable: false,
-  value: null
-};
-
-export let GUID_KEY_PROPERTY = {
-  name: GUID_KEY,
-  descriptor: nullDescriptor
-};
 
 export function buildGuidFor(obj) {
   let type = typeof obj;
@@ -134,6 +99,17 @@ export function buildGuidFor(obj) {
   @return {String} the guid
 */
 export function generateGuid(obj, prefix) {
+  if (obj) {
+    let meta = metaFor(obj);
+
+    if (prefix) {
+      // TODO: this is kinda weird, we should deprecate and remove
+      meta._guid = prefix + meta._guid;
+    }
+
+    return meta.sourceGuid();
+  }
+
   let ret;
   if (prefix) {
     ret = prefix + uuid();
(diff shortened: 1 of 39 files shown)


runspired:
@rwjblue does this mean that addons (including ember-data) that currently access things via `obj[Ember.GUID_KEY]` need to refactor to `Ember.guidFor(obj)` ?

Furthermore, ember-data uses this to set the guid onto `InternalModel` itself (uses a completely custom ID).  Refactoring to `Ember.generateGuid(internalModeInstance, '-internalModel')` shouldn't be breaking anything but it would lose the nicety of knowing how many total internal models we've created based on the current guid increment there.  I think it's doable without ;)

runspired:
ember-data also pre-heats `__ember_meta__` for `InternalModel`, so a link to the new pattern to use there would also be helpful.

rwjblue:
Great questions!

> does this mean that addons (including ember-data) that currently access things via obj[Ember.GUID_KEY] need to refactor to Ember.guidFor(obj) ?

Getting an objects guid should always have been done through `Ember.guidFor`.  That works (and is public API) going back to Ember 1.0.  Being locked to supporting `obj[Ember.GUID_KEY]` makes future refactors pretty hard/impossible...

> Furthermore, ember-data uses this to set the guid onto InternalModel itself (uses a completely custom ID). Refactoring to Ember.generateGuid(internalModeInstance, '-internalModel') shouldn't be breaking anything but it would lose the nicety of knowing how many total internal models we've created based on the current guid increment there.

What Ember Data does [around here](https://github.com/emberjs/data/blob/1f7e2a24a4f5758ce0ec288db29e814d8d637124/addon/-private/system/model/internal-model.js#L84) seems totally fine in the general case.  There is no real reason to use `Ember.GUID_KEY` there as those objects are private and `Ember.guidFor` isn't even called on them in the codebase.  I can keep exporting `GUID_KEY` so that this code doesn't break (as of this PR it would be doing `this['undefined'] = ++uuid + '-internal-model';`).

> ember-data also pre-heats __ember_meta__ for InternalModel, so a link to the new pattern to use there would also be helpful.

This is still good for the non-native WeakMap case (where we still use `__ember_meta__`).  When native WeakMap's are available, neither meta nor guid require a property on the object.

runspired:
> and Ember.guidFor isn't even called on them in the codebase

Probably a topic digression but I believe the reasoning here isn't that it's called directly but that `InternalModel` is proxied through computed properties into a snapshot and relationships and they wanted it there for inevitable meta and property lookups because of that.

rwjblue:
> Probably a topic digression but I believe the reasoning here isn't that it's called directly but that InternalModel is proxied through computed properties into a snapshot and relationships and they wanted it there for inevitable meta and property lookups because of that

Gotcha.  That makes sense. Even given that, this PR doesn't fully break anything in Ember Data AFAICT (the `Ember.guidFor` results wouldn't match the `InternalModel[Ember.GUID_KEY]` value but they will still be unique and function in templates / CP's).  

That being said, it probably makes sense for ED to change that line to use `Ember.guidFor(this)` there instead...

stefanpenner:
ember-data no longer uses `GUID_KEY`, and i believe it was only done as a slight perf optimization. So it likely wouldn't break even when using older versions.

packages/ember-metal/lib/guid.js
```diff
@@ -0,0 +1,150 @@
...
+    separate the guid into separate namespaces.
+  @return {String} the guid
+*/
+export function generateGuid(obj, prefix) {
+  if (obj) {
+    let meta = metaFor(obj);
+
+    if (prefix) {
+      // TODO: this is kinda weird, we should deprecate and remove
+      meta._guid = prefix + meta._guid;
```
stefanpenner: droping the prefix would be nice. It seems somewhat un-needed. Although maybe for compat?

runspired: We dropped it in ED 2.10 or 2.11 (one of). For compat I don't think we can drop it just yet?

rwjblue:
Closing in favor of #15143

Issue #14381, opened 2016-09-29
[Beta Bug] attrs in bindings are always truthy


chadhietala:
*Repro*
https://ember-twiddle.com/7ed79a745f34ce66d44f6b23d5465758?openFiles=templates.application.hbs%2Ctemplates.components.my-component.hbs

We are likely not reading `value` from the `MutableCell` and thus the attribute bindings are always true.

HeroicEric:
> We are likely not reading value from the MutableCell and thus the attribute bindings are always true.

Yep. It looks like https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/utils/bindings.js#L175, `inner` is a `NestedPropertyReference` and `inner.value()` is a `MutableCell`, like you said.

chadhietala:
See #14384

HeroicEric:
I believe this was fixed by ^^ https://github.com/emberjs/ember.js/pull/14384

PR #14384, opened 2016-09-29
[Bugfix Beta] Fix attrs in className and attribute bindings

chadhietala:


chadhietala:
Fixes #14381

packages/ember-glimmer/lib/utils/bindings.js
```diff
@@ -14,6 +14,17 @@ function referenceForKey(component, key) {
...
 function referenceForParts(component, parts) {
+  let isAttrs = parts[0] === 'attrs';
+
+  // TODO deprecate this
+  if (isAttrs) {
+    parts.shift();
+
+    if (parts.length === 1) {
+      return referenceForKey(component, parts[0]);
+    }
```
krisselden: what is the else here? is it supposed to fall through or assert?

packages/ember-glimmer/lib/utils/bindings.js
```diff
@@ -14,6 +14,17 @@ function referenceForKey(component, key) {
 }
 
 function referenceForParts(component, parts) {
+  let isAttrs = parts[0] === 'attrs';
```
krisselden: is parts guaranteed to be length 1? is it very important that we never index outside of the array on get.

chadhietala: Upstream from this [the check is done](https://github.com/emberjs/ember.js/blob/473a4f5f05423837c70376d8c12a5206aa2f8629/packages/ember-glimmer/lib/utils/bindings.js#L66-L67) to make sure we have a path reference.

packages/ember-glimmer/lib/utils/bindings.js
chadhietala: We want this to fall through. So given `attrs.foo` we want to delegate to `referenceForKey`, however if we shift off the value and we still have a path e.g. `parts.length !== 1` (guranteed to never be 0) then we want to still create a path ref.

Issue #14380, opened 2016-09-29
[Regression]: Layout supports computed property (glimmer 2)


toranb:
In ember 2.8.1 if my layout prop for a component was a computed property like so ... all was good

```js
export default Ember.Component.extend({
  tagName: '',
  layout: Ember.computed(function () {
    //get some dynamic vars here ...
    return Ember.HTMLBars.compile(`
      {{component "ui-button"}}
    `);
  }).volatile()
});
```

Here is a screenshot of this working as you see above

<img width="789" alt="screen shot 2016-09-28 at 8 00 16 pm" src="https://cloud.githubusercontent.com/assets/147411/18937574/1a374594-85b7-11e6-999f-f24146857945.png">

Once I upgraded to ember 2.9.0-beta.4 I got this error `Uncaught TypeError: this.template.asLayout is not a function` (note: no deprecation in ember 2.8.1 alerted me that using a CP (for layout) would not be supported)

<img width="998" alt="screen shot 2016-09-28 at 8 02 00 pm" src="https://cloud.githubusercontent.com/assets/147411/18937576/1e774f00-85b7-11e6-9937-0929c1281c8d.png">

Here is the example repo to show the regression in action

https://github.com/toranb/broken-glimmer2/commits/master

I'd like to see if we can keep support for computed properties in glimmer 2 (for layout) because this has allowed @MiguelMadero and myself to build a solid hot reloader that is very useful (and until glimmer 2 this worked from 1.13+)

runspired:
@toranb theoretically, the ability to do this was never documented. I've bumped into fun things here repeatedly: 
 
https://github.com/html-next/flexi/blob/master/addon/lib/sustain.js#L223-L230

As I dive deeper into getting flexi to work on glimmer2 tonight, I'm betting I also bump into this.

rwjblue:
To fix this, we need to do:

1. Change [this](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/syntax/curly-component.js#L265) to be `let Template = get(component, 'layout');`
2. Add a test [like this one](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/tests/integration/components/curly-components-test.js#L137-L153) that uses a CP.

@toranb - Wanna give it a shot?

toranb:
@rwjblue you know it! thanks for the opportunity :)

PR #14390, opened 2016-09-30
[BUGFIX beta] layout now supports computed property

toranb:
In the 2.9.0-beta.4 release layout breaks if you are using a computed property. This pull request will close out issue #14380


rwjblue:
Thanks @toranb!

PR #14392, opened 2016-09-30
[DOC release] Use `Ember.computed`

locks:
Replace function prototype extension with  `Ember.computed`


packages/ember-runtime/lib/mixins/observable.js
```diff
@@ -163,6 +163,10 @@ export default Mixin.create({
 
   /**
     Sets the provided key or path to the value.
+
```
rwjblue: Trailing whitespace ✂️

rwjblue (changes requested):

rwjblue (approved):

PR #14395, opened 2016-09-30
[DOC release] Add code sample to Ember.get

locks:


packages/ember-metal/lib/property_get.js
```diff
@@ -22,6 +22,10 @@ const ALLOWABLE_TYPES = {
   Gets the value of a property on an object. If the property is computed,
   the function will be invoked. If the property is not defined but the
   object implements the `unknownProperty` method then that will be invoked.
+
```
rwjblue: looks like there is some trailing whitespace here that failed the tests

rwjblue (changes requested):

rwjblue (approved):

PR #14403, opened 2016-10-03
[BUGFIX beta] Fix tagless `willInsertElement`

### LINKED ISSUE #14398: Glimmer2 willInsertElement does not fire for tagless components.
Replication: https://ember-twiddle.com/0196406cd88aac368ba8b82bd88393b4?openFiles=components.tagless-component.js%2Ctemplates.components.tagless-component.hbs

While true that `didInsertElement` and `willInsertElement` make less semantic sense for tagless components, they function as standings for `didRenderInitially` and `willRenderInitially`

chancancode:
- Transition into `hasElement` state for tagless components
- Fire `willInsertElement` on tagless components
- Fire `willInsertElement` before firing `willRender` (thus ensuring `willRender` gets the element as well)
- Add lifecycle tests for tagless components
- Refactor the lifecycle tests to remove some duplications

Fixes #14398

### DIFF
--- a/packages/ember-glimmer/lib/syntax/curly-component.js
+++ b/packages/ember-glimmer/lib/syntax/curly-component.js
@@ -211,8 +211,14 @@ class CurlyComponentManager {
     component.trigger('didInitAttrs', { attrs });
     component.trigger('didReceiveAttrs', { newAttrs: attrs });
 
-    if (environment.isInteractive) {
-      component.trigger('willRender');
+    // We usually do this in the `didCreateElement`, but that hook doesn't fire for tagless components
+    if (component.tagName === '') {
+      component._transitionTo('hasElement');
+
+      if (environment.isInteractive) {
+        component.trigger('willInsertElement');
+        component.trigger('willRender');
+      }
     }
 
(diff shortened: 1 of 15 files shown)


packages/ember-glimmer/lib/syntax/curly-component.js
```diff
@@ -211,8 +211,14 @@ class CurlyComponentManager {
     component.trigger('didInitAttrs', { attrs });
     component.trigger('didReceiveAttrs', { newAttrs: attrs });
 
-    if (environment.isInteractive) {
-      component.trigger('willRender');
+    // We usually do this in the `didCreateElement`, but that hook doesn't fire for tagless components
```
rwjblue: I'd like to move this logic to a central location (to avoid duplicating the exact order and conceptual logic both here and in `didCreateElement`).  In a similar situation, I added logic to the `ComponentStateBucket` object but it could likely live in a few different places based on your preferences.

rwjblue (approved):
LGTM, the centralization of the logic can be done as a follow-up.

rwjblue:
@chancancode - How is this going?

PR #14407, opened 2016-10-04
Local lookup component closure test

sglanzer-fire:
Adds a local lookup test case for a component closure passed as a named argument to a component.

e.g. Given a local component:

`my-route/-components/local-component`

Usage in my-route.hbs:

``` hbs
{{global-component
  foo=(component 'local-component')
}}
```


sglanzer:
@rwjblue as discussed - passes on master, fails on 2.8.1

rwjblue:
Merging into master + beta, and we can target 2.8 (current release branch) with a second PR since it will need to fix implementation.

Issue #14335, opened 2016-09-22
Closure action causes unnecessary rerender


scottmessinger:
Hi! Thanks for all the hard work on Glimmer 2! I know it's taken a ton of work and I really appreciate it.

In using Glimmer 2 in my app, I think I've found a bug. If a component passes a closure action to its children and we change a property of the component, all the children are forced to rerender even  though we didn't pass that property to them (and thus none of their attrs (or internal state) changed).

For example, if I have a component (A) that has a `counter` property and I increment it, all of A's children will rerender even if I didn't pass them the counter property as long as I did pass them a closure action.

Here's the test case: open the console to see `didRender` being logged
https://ember-twiddle.com/5e495e8b7e6cb9ec8d56400f7b2d72ff

Notably, if you stop passing the closure action to `expensive-outline`, it will still rerender when `counter` is changed. This seems like a different bug but possibly related.

rwjblue:
Thank you for the awesome writeup and reproduction, @chancancode and I were able to track this down and fix in #14345.  Can you test the latest canary/alpha builds (in ~ 15 minutes when CI finishes) to confirm?

rwjblue:
I reviewed this after that CI build finished, and we still have the same issue.

To save myself time when I pick this back up again, let me try to explain the issue:

The way Glimmer 2 tracks changes and triggers rerenders, is by having a "tag" on each object that is used in a template. Think of these tags as a counter that just gets `counter++` and if the value is different then we actually do a rerender.

In the case of the example, the `{{expensive-outline`'s tag increments because it receives `counter`.  We are then passing `this.onAction` down to the `{{expensive-outline-item` component, and since the `expensive-outline` component's tag increments (and therefore the args for `{{expensive-outline-item`'s tag changes) which causes a rerender of the `expensive-outline-item`.

kanongil:
I would just like to chime in that this effects my code as well in canary, causing significant re-rendering, and is likely to block updating my app to 2.9+, if not resolved.

rwjblue:
I mentioned some of this in https://github.com/emberjs/ember.js/pull/14410#issuecomment-251687821, but I think having a way to absorb the difference (between per object and per property invalidation) would definitely be useful. The bet that we are making with Glimmer 2 is that tracking all individual properties for changes is actually more costly than recomputing.  We just need an escape hatch if that isn't the case (for targeted scenarios).

stefanpenner:
[...]

PR #14410, opened 2016-10-04
[BUGFIX beta] Return stable object from hash helper

### LINKED ISSUE #14362: Extraneous change events with Glimmer hash helper (performance regression)
When I observe a component property provided by a `hash` helper in Ember 2.9-beta.3 and canary, the observer fires any time the parent component is revalidated. It used to behave correctly in 2.8.

Failing pattern: ```{{my-component hash=(hash a=a)}}```

I have create a simple twiddle to demonstrate:

https://ember-twiddle.com/3d806dacb9ec04aa12afcdce0d6bc742?openFiles=components.my-component.js%2Ctemplates.components.my-component.hbs

In 2.8 the console output is:
```
```

In 2.9-beta.3 and canary it is:
```
aDidChange – "a"
```

kanongil:
The hash helper needs to return a stable object.

This is done by wrapping the previous EvaluatedNamedArgs in a new HashReference class, updating the values as they are invalidated.

This also fixes #14362.

### DIFF
--- a/packages/ember-glimmer/lib/helpers/hash.js
+++ b/packages/ember-glimmer/lib/helpers/hash.js
@@ -2,6 +2,54 @@
 @module ember
 @submodule ember-glimmer
 */
+import { set } from 'ember-metal';
+import { PrimitiveReference } from '../utils/references';
+
+class HashReference extends PrimitiveReference {
+
+  constructor(args) {
+    super();
+
+    this.tag = args.tag;
+    this.args = args;
+
+    this.hash = null;
+  }
+
+  value() {
+    let { args, hash } = this;
+    let { keys, values } = args;
+
+    if (!hash) {
+      this.hash = hash = args.value();
+
+      let lastRevisions = this._lastRevisions = new Array(keys.length);
+      for (let i = 0; i < keys.length; i++) {
+        let ref = values[i];
+        lastRevisions[i] = ref.tag.value();
+      }
+    } else {
+      let { _lastRevisions:lastRevisions } = this;
+
+      for (let i = 0; i < keys.length; i++) {
+        let ref = values[i];
+        let rev = lastRevisions[i];
+        if (!ref.tag.validate(rev)) {
+          let key = keys[i];
+          lastRevisions[i] = ref.tag.value();
+          set(hash, key, ref.value());
+        }
+      }
+    }
+
+    return hash;
+  }
+
+  get(path) {
+    let { args } = this;
+    return args.get(path);
+  }
+}
 
 /**
    Use the `{{hash}}` helper to create a hash to pass as an option to your
@@ -31,5 +79,5 @@
  */
 
 export default function(vm, args) {
-  return args.named;
+  return new HashReference(args.named);
 }
--- a/packages/ember-glimmer/tests/integration/helpers/hash-test.js
+++ b/packages/ember-glimmer/tests/integration/helpers/hash-test.js
@@ -1,6 +1,6 @@
 import { RenderingTest, moduleFor } from '../../utils/test-case';
 import { Component } from '../../utils/helpers';
-import { set } from 'ember-metal';
+import { get, set, observer } from 'ember-metal';
 
 moduleFor('Helpers test: {{hash}}', class extends RenderingTest {
 
@@ -166,5 +166,87 @@ moduleFor('Helpers test: {{hash}}', class extends RenderingTest {
     });
 
     this.assertText('Chad Hietala');
+  }
+
+  ['@test returns stable object when updated']() {
+    let fooBarInstance;
+    let FooBarComponent = Component.extend({
+      init() {
+        this._super();
+        fooBarInstance = this;
+      }
+    });
+
+    this.registerComponent('foo-bar', {
+      ComponentClass: FooBarComponent,
+      template: `{{model.firstName}} {{model.lastName}}`
+    });
+
+    this.render(`{{foo-bar model=(hash firstName=firstName lastName="Arbeo")}}`, {
+      firstName: 'Sergio'
+    });
+
+    let hashInstance = get(fooBarInstance, 'model');
+
+    strictEqual(hashInstance, get(fooBarInstance, 'model'));
+    this.assertText('Sergio Arbeo');
+
+    this.runTask(() => this.rerender());
+
+    strictEqual(hashInstance, get(fooBarInstance, 'model'));
+    this.assertText('Sergio Arbeo');
+
+    this.runTask(() => set(this.context, 'firstName', 'Godfrey'));
+
+    strictEqual(hashInstance, get(fooBarInstance, 'model'));
+    this.assertText('Godfrey Arbeo');
+  }
+
+  ['@test correctly triggers observers']() {
+    let modelTriggerCnt = 0;
+    let firstNameTriggerCnt = 0;
+    let lastNameTriggerCnt = 0;
+
+    let FooBarComponent = Component.extend({
+      modelObserver: observer('model', () => {
+        modelTriggerCnt++;
+      }),
+
+      firstNameObserver: observer('model.firstName', () => {
+        firstNameTriggerCnt++;
+      }),
+
+      lastNameObserver: observer('model.lastName', () => {
+        lastNameTriggerCnt++;
+      })
+    });
+
+    this.registerComponent('foo-bar', {
+      ComponentClass: FooBarComponent,
+      template: `{{model.firstName}} {{model.lastName}}`
+    });
+
+    this.render(`{{foo-bar model=(hash firstName=firstName lastName="Arbeo")}}`, {
+      firstName: 'Sergio'
+    });
+
+    equal(modelTriggerCnt, 0);
+    equal(firstNameTriggerCnt, 0);
+    equal(lastNameTriggerCnt, 0);
+    this.assertText('Sergio Arbeo');
+
+    this.runTask(() => this.rerender());
+
+    equal(modelTriggerCnt, 0);
+    equal(firstNameTriggerCnt, 0);
+    equal(lastNameTriggerCnt, 0);
+    this.assertText('Sergio Arbeo');
+
+    this.runTask(() => set(this.context, 'firstName', 'Godfrey'));
+
+    equal(modelTriggerCnt, 0);
+    equal(firstNameTriggerCnt, 1);
+    equal(lastNameTriggerCnt, 0);
+    this.assertText('Godfrey Arbeo');
   }
 });


krisselden (changes requested):
We should implement as a non user helper because the hash was already a ref and it could be returned

kanongil:
@krisselden I'm not sure what you are suggesting? as far as I know, the hash helper is already a non user helper, though I am unsure about the terminology.

The existing implementation returns an `EvaluatedNamedArgs` which is a reference.

Ember itself uses the hash helpers `.value()` when it is invalidated in `CurlyComponentManager.update()`. Here it is treated as a regular object, which is `set()` on the component. Since the current implementation re-calculates the object on each `.value()` invocation, this results in triggering observes on itself and all child properties, even constant ones.

Note that while the component usage is merely a performance regression, it results in an actual bug if the component implementation adds a property to the object provided by a hash helper, which will vanish on the next update.

Come to think of it, I am basing the stable object off the one returned from `EvaluatedNamedArgs.value()`, and rely on this object being mutable. I'm not quite sure if this is a safe assumption about the Glimmer value interface. Maybe someone can enlighten me?

rwjblue:
The `hash` helper in prior versions was not stable either ([see demo here](https://ember-twiddle.com/8d59d36de9671dad0b979d6991ccc95f)). When the bound properties changed, a new object was returned (they were not `===` the same object). 

The main difference in 2.9+ is that the helper always returns a new object if _any_ property on the current context changes. We are no longer tracking per property changes and instead we are relying on much coarser grained "dirtiness".  Our position is that the majority of the time this is perfectly fine, as it is very cheap do this work, but when you mix this with observers (and possibly CP's) you end up with your observers firing for changes due to something in the path changing (and not the leaf property that you are using/depending on).  FWIW, you can avoid the issue of your observers firing too many times, by checking/guarding for the values that you care about changing.

I think there definitely is a place for the thing that has been built here, but I don't think we should make the default `hash` helper do more work.  Perhaps a different name (maybe something like `stable-hash`?) and a feature flag (but could be provided to 2.9 via an addon until the feature flag is enabled).

kanongil:
@rwjblue Thanks for the detailed reply.

Did you do any investigation into whether this "bet" was worth pursuing? While, I agree that it is usually dirt cheap to recompute, especially with internal caches, you seem to forget that this will generate a lot of extra `set()` calls which needs to be factored in, especially with these kind of unstable `.value()` responses. I would almost go as far to say that this bet _requires_ stable responses whenever possible.

In my app the unstable hash response causes a kind of cascading effect, which means that the 2.9 upgrade re-validates 200+ extra components due to an unrelated property change in a parent.

kanongil:
Btw, the implementation can be simplified to better align with this bet by dropping the revision tracking.

kanongil:
So I tried to profiling my heavily optimised app, and during a performance critical scrolling operation, I am observing a 5x reduction in performance, due to extra validations.

While I realise that performance is not an explicit target for the Glimmer upgrade, this is un-acceptable to me.

ef4:
> Did you do any investigation into whether this "bet" was worth pursuing? 

@kanongil you are basically asking if anybody thought through the whole point of the glimmer2 architecture and did any measurements. And the answer is: yes, multiple entire engineer years have gone into measurement and validation of the architecture. We are very confident it lets people build dramatically faster applications.

Your heavily optimized app was optimized under pre-glimmer2 conditions. The same optimizations do not apply here. The fact that those optimizations are no longer optimizations is part of the point. 

You mention excessive observer firing as being a problem. But relying on _any_ observer firing is precisely the kind of pattern we are deliberately pushing people away from, because it's extremely difficult for users to do correctly and for the runtime to optimize.

kanongil:
@ef4 It seems you mistook my comment as being in bad spirit, and I'm sorry for that. I was simply trying to inquire if you had any such investigations to share.

I am not questioning the validity of your design, and I have deep respect for all the work that has gone into this feature. I am questioning some of the specifics in the Ember-Glimmer2 integration, which are causing performance degradations, and trying to provide potential solutions.

As it is now, using certain features in specific ways can cause this. If you think that this is acceptable, you really need to document the most common patterns, and how to fix it. I'm sure I'm not the only one eagerly awaiting this update, and people that have similar experiences are likely to be dis-illusioned without a clear migration path. I'm still heavily in favour of rooting out these issues, and fixing them like I'm trying to do with this PR, as I'm sure people will continue to repeat any bad patterns, ultimately labelling Ember as slow.

I don't believe I am using Ember in a wrong way, and the extra observer firing just seems to be a symptom of an underlying issue.

ef4:
It sounds like we should take a look at the specifics of your app and why it is so sensitive to this issue. If you can share more specifics I would be happy to review.

scottmessinger:
@ef4 It doesn't appear this bug only affects apps with unique designs or only apps using observers or CP. If I'm understanding the root cause, the lack of "a way to absorb the difference (between per object and per property invalidation)" (according to @rwjblue) is causing Ember to rerender quite unexpectedly which is how this issue is related to #14335.

In 2.9, if I pass a hash to a component towards the top of my component hierarchy and then do a `Ember.set(this, 'theHash', aNewHash)` *all* the components that are children of that component will rerender even though they aren't using `theHash` or any properties inside it. Here's a [twidde demonstrating the issue](https://ember-twiddle.com/ed778b37d75d558108599af80a80f5a3?openFiles=templates.components.a-child.hbs%2Ctemplates.components.a-child.hbs). Having worked with Ember since 0.9.6, that cascading rerender really surprises me and it's not how 2.8 works.

raytiley:
@ef4 @scottmessinger I've been surprised even before glimmer2 some of the times ember decides to re-render. I brought this up a while ago with an example in my app, but I always thought that the point of CPs and chaining is that you could stop expensive computations if the input values haven't changed.

Seems like any value in the chain changing will trigger recomputes even if the the actual dependent properties don't change: https://ember-twiddle.com/ed7b9db652dd5202baa3b3124731946d?openFiles=controllers.application.js%2C

scottmessinger:
@raytiley That's a **great** example. Of all the twiddles, I think that gets at the problem most succinctly.

raytiley:
@scottmessinger  :)

FWIW @mixonic thought there was a way that this could be improved, but it's probably a future optimization.

ef4:
@raytiley's example is informative for explaining the semantics, but it's also not new in glimmer2. 

@scottmessinger I'm not clear on how the hash helper relates to your example. The twiddle doesn't use it. 

>  it's not how 2.8 works.

Agreed, but it is also part of the point of glimmer2. Your example shows `didReceiveAttrs` hooks firing, but that by itself is not bad. As long as any expensive computations in those components are contained within CPs, and those CPs own dependent keys have not been touched, nothing expensive will happen and the full re-render of the component hierarchy is extremely fast. And the fact that those hooks fire greatly simplifies the problem of handling interior mutability. Just because a component's inputs are all `===` doesn't mean nothing down deeper needs to rerender (that can only be guaranteed with fully immutable objects).

scottmessinger:
@ef4 My understanding is glimmer2's concept of dirtiness is the connection between this issue and #14335.

rwjblue:
@scottmessinger - Confirm, they are linked in the conceptual "this is how the world works now" way. And its convenient to have the conversation in one place 😺 .

kanongil:
The @raytiley issue is indeed not new, and can be handled with something like [ember-computed-change-gate](https://github.com/GavinJoyce/ember-computed-change-gate) on properties that are likely to produce similar results from changing inputs.

I'm guess @scottmessinger's twiddle just demonstrates the new rendering mechanics, based on the "bet" @rwjblue mentioned.

@ef4 I would like to question this "extremely fast" assertion that you are making. I have created a jsbin that renders 500 logic less components. When I change an unrelated value on the main component, it renders orders of magnitude faster in 2.8 vs 2.9. See http://jsbin.com/miroso/21/edit?html,js,output

ef4:
@kanongil seems bad, thanks for the reproduction.

ef4:
I spent some time running the scenario from @kanogil's twiddle through ember-bench. 

First, I will caution that this kind of microbenchmark has often led people astray. It's too easy to accidentally move work around, making one scenario faster while hurting other scenarios, or even just moving work from phase to phase so that it escapes getting captured by the benchmark.

Nevertheless, here is the stable re-render performance of the 500 components, which does indeed show that glimmer2 is on the order of 20ms slower:

screen shot 2016-10-10 at 12 06 02 pm

However, here is the _initial_ render performance of the same scenario:

screen shot 2016-10-10 at 11 26 06 am

We save 80ms at initial render, at the cost of 20ms during re-render. This is a huge win for the majority of Ember apps, because Ember's re-renders have already been blazing fast since 1.13 -- it's initial render that causes practical problems. We were able to completely eliminate 60ms of work, and move another 20ms to a time that is less critical.

Now, that doesn't mean we shouldn't *also* attempt to push that re-render cost downward. But I hope this illustrates why coarser observation is a winning strategy.

kanongil:
Yes, initial render performance is indeed important, and I'm sure that this will benefit many, including my app. It seems that Glimmer2 optimizes for this, along with massively changing UI's.

However, I dare say that most normal apps have many operations that only change the UI a tiny bit, like my example. Previously, this worked fine even with complex UI's, if you accepted the initial render hit. It worries me that this is no longer the case. As it is, I am unable to update to 2.9 due to this change, and it seems to me you are saying I just need to suck it up and somehow find a way to optimize my app around this.

Note that ~40ms to render a simple change in the UI is means that I can't run it at more than 25 fps, and probably 5 fps on Android devices. It doesn't really help me that Glimmer2 enables me to also change all 500 items in that same time.

ef4:
> it seems to me you are saying I just need to suck it up

This issue is still open and nobody has told you to suck it up.  The point of my comment was not to dismiss the problem, it was to quantify it and put it into context.

kanongil:
@ef4 Sorry, I was putting things a bit on the edge, but thanks for the clarification.

It would be nice if someone would acknowledge that, while Glimmer2 is awesome for initial rendering and full UI updates, it can fall short for simple changes in complex UI's. Something I fear will put off a lot of existing users.

mixonic:
A bit of a sidebar, but I'd like to note that jsbin's make a fairly poor platform for performance evaluation with Ember. @ef4 porting it to an ember-cli app seems to have been a good place to start for analysis.

For example it seems that the `ember-template-compiler` version of the object model was polluting the runtime object model. If I swap the order of the `ember.js` asset and `ember-template-compiler.js` asset the measured timings change (times on Chrome 54, i7 OSX):

* [original jsbin with `ember-template-compiler.js` first](http://jsbin.com/miroso/21/edit?html,js,output), usually under 12.5ms rerender
* [jsbin with `ember.prod.js` first](http://jsbin.com/meyajilixe/1/edit?html,js,output), usually under 9.5ms rerender

Of course `ember-template-compiler` doesn't ship to the client in Ember-CLI apps so this isn't an issue for normal usage. The asset order change doesn't close the gap to 2.8 or anything, just want to be sure everyone is looking at the same data.

scottmessinger:
Thanks @mixonic and @ef4. Great insights. Regarding the example, I'm a bit concerned it's drawing attention to the wrong problem. To me, the point isn't that it takes 20ms or 9ms to rerender -- the point is that the rerender is happening at all. In the simple example @kanongil put together, speeding the re-render obviously helps solve the problem, but what if those 500 components resembled components from actual apps and had some level of complexity to them? What if they had sub-components? What if they took 1ms a piece to re-render? At 500 components, even if we took the re-render down 50%, we're still looking at a 250ms frame. If the solution to the problem is "we'll make the re-render faster", I fear we'd still bound to run into performance problems in a real apps, especially when we make change a property towards the top of the component tree. Coarser observation might indeed be the winning strategy, but I'm struggling to see how unnecessary re-rendering can ever be the right choice. Whether coarse or fine observation, it seems the goal needs to be preventing unnecessary re-renders. Am I misunderstanding your point, @ef4? You understand this pretty well and I might be misinterpreting the implication of the comment you made about re-rendering.

ef4:
@scottmessinger you are quite correct, the goal is not to go from one extreme (fine-grained observation everywhere) to another (rerender the world every time) but to find a happy balance, and we can tune the validation system to get there.

ef4:
If anybody can provide me access to additional real apps that are exhibiting this expensive re-render behavior, that would help us figure out the appropriate tuning sooner. (Isolated examples are illustrative, but they are harder to trust when it comes to balancing holistic app performance.)

scottmessinger:
@ef4 I'll work on cleaning up the code in my app and ping you later in the week. Do you need to see the code or just a demo of the app?

For my app, we have an outline with 300-500 things in it. When you press the up/down arrow, we set a property at the top of the outline called `focusedItemId`. The outline items set the focus on their input when their id is the same as the `focusedItemId` via the `didReceiveAttrs` hook -- no rerender happens. In 2.9, the entire outline rerenders. Here's what it looks like:

```hbs
{{! application.hbs}}
{{outline arrowDown="arrowDown" focusedItemId=focusedItemId}}

{{! outline.hbs }}
{{#each items as |item| }}
   {{outline-item
     arrowDown=(action arrowDown item.id)
     hasFocus=(eq item.id focusedItemId) }}
{{/each}}
```

```js
// application.js
  actions: {
     arrowDown(id){ Ember.set(this, 'focusedItemId', id) }
  } 

// components/outline-item.js
  didReceiveAttrs(){
    this._super(...arguments)
    if (get(this, 'hasFocus')){
      Ember.run.scheduleOnce('afterRender', () => this.$('.js-outline-item-input')[0].focus() )
    }
  }
```

kanongil:
Now that #14461 has been merged, my revalidation jsbin is no longer an issue, and much of the comments in this issue are irrelevant regarding the original PR. As far as I can tell it is significantly faster than 2.8! 👍

Note that a stable hash helper is still interesting for performance, since modifying a property in the hash currently (2.8 & beta) triggers observers against all keys in the hash. As such, I would still like to see this implemented somehow, though as future enhancement, rather than a bugfix. I have filed a separate issue to track this in #14468.

Thanks for all the interest and hard work. Looking forward to the next release!

PR #14414, opened 2016-10-04
[CLEANUP] Remove ember-routing-route-configured-query-params feature flag

trentmwillis:
Per-discussion in the community Slack, this feature appears dead. Part of cleaning up QPs.

### DIFF
--- a/features.json
+++ b/features.json
@@ -1,7 +1,6 @@
 {
   "features": {
     "features-stripped-test": null,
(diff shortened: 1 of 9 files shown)


trek:
Heyo @trentmwillis, I was also tackling this (it felt right since I added it). Were you going to keep working on moving the behavior back to controllers as well? I don't want to duplicate further effort.

PR #14421, opened 2016-10-05
Fix for local lookup of closure components

sglanzer-fire:
The source/module in the env when creating a closure component is correct, but wasn't accessible/used when rendering the closure component in a child context.  This fix retains the source/module and sets the env appropriately when a closure component is rendered.

### DIFF
--- a/packages/ember-glimmer/tests/integration/components/local-lookup-test.js
+++ b/packages/ember-glimmer/tests/integration/components/local-lookup-test.js
@@ -118,6 +118,68 @@ moduleFor('Components test: local lookup', class extends RenderingTest {
     this.assertText('yall finished or yall done?');
   }
 
+  ['@test it can local lookup a dynamic component from a passed named argument']() {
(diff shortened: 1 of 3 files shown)


sglanzer:
@rwjblue

rwjblue:
This looks good to me!  

@Serabe - Would you mind reviewing also?

sglanzer:
Missed a semicolon, should pass CI now

Serabe:
Should we add a test for rerendering with a different component with different scope?

In any case, LGTM.

rwjblue:
Good idea!

@sglanzer - Would you have time to add that test?

sglanzer:
Yeah, I can add that test - so e.g. swap from local to global scope on the rerender?

rwjblue:
@sglanzer - Yes, exactly.

sglanzer:
Added https://github.com/emberjs/ember.js/pull/14421/files#diff-60d0781473fafba06d2d55675ee1286cR166

rwjblue:
Thank you again @sglanzer!

PR #14422, opened 2016-10-05
Porting additional local-lookup tests

sglanzer-fire:
Ported from #14421 against release (2.8)


rwjblue:
Thanks again @sglanzer!

Issue #14423, opened 2016-10-05
[Beta Bug] `{{#each this.attrs.* as |item|}}{{/each}}` no longer works


chadhietala:
Repro: https://ember-twiddle.com/e745d78ac6b44a6cfc1c891eaff04a1a?openFiles=templates.application.hbs%2Ctemplates.components.my-component.hbs

One use to be able to do this in a component:

```hbs
{{#each this.attrs.items as |item|}}
  {{item}}
{{/each}}
```

rwjblue:
Fixed by #14425

PR #14425, opened 2016-10-05
[BUGFIX beta] Transform `this.attrs` in templates

chadhietala:


chadhietala:
Fixes #14423

packages/ember-template-compiler/lib/plugins/transform-attrs-into-args.js
```diff
@@ -27,6 +27,20 @@ export default function TransformAttrsToProps() {
   this.syntax = null;
 }
 
+function isAttrs(node) {
+  if (node.parts[0] === 'attrs') {
+    return true;
+  }
+
+  let [_this, attrs] = node.parts;
```
stefanpenner: we may have already worked around this generally but just in-case please double check as this destructuring is not what I suspect you think it is:

https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-2&code=let%20%5Bz%5D%20%3D%20node.parts

packages/ember-template-compiler/lib/plugins/transform-attrs-into-args.js
```diff
@@ -27,6 +27,20 @@ export default function TransformAttrsToProps() {
...
+
+  if (_this === null && attrs === 'attrs') {
+    node.parts.shift();
+    node.original = node.original.slice(5);
+    return true;
+  }
```
runspired: should we have the ability to warn when we do this so that people can fix their code bases?

packages/ember-template-compiler/lib/plugins/transform-attrs-into-args.js
chadhietala: I removed this just in case. I think I was _safe_ sense I was just reading the values.

packages/ember-template-compiler/lib/plugins/transform-attrs-into-args.js
chadhietala: We said we weren't going to introduce any deprecations as part of 2.9.0. We should deprecate a lot of this stuff in 2.10.0.

runspired: heh, so basically we just silently deprecated it :P

PR #14427, opened 2016-10-06
[BUGFIX beta] Setting in update should render

chadhietala:


packages/ember-glimmer/lib/renderer.js
```diff
@@ -367,8 +369,10 @@ class Renderer {
 
     try {
       this._renderRoots();
-    } finally {
+    } catch (e) {
```
rwjblue: :( doesn't this bring back the pain while debugging?

Could we do this in the finally with a local variable?

rwjblue: Specifically:

```
var completedWithoutError = false;
try {
  this._renderRoots();
  completedWithoutError = true;
} finally {
  if (!completedWithoutError) {
    this._lastRevision = CURRENT_TAG.value();
  }

  this._isRenderingRoots = false;
}
```

rwjblue: The issue is with debugging the rethrow. It completely masks errors and makes "stop on uncaught" exceptions in chrome unusable. 

I was suggesting a way to avoid the catch, but still have the behavior you changed here (only update last revision in the error case).

krisselden: I'm ok with that, it looks terrible, we could also make it a try/finally between begin/commit

krisselden: we'd need to extract the loop inside so that the try/finally is minimized

chadhietala: Fixed this.

PR #14429, opened 2016-10-06
[CLEANUP] Move TestCase helpers to internal-test-helpers

trentmwillis:
Follow up to https://github.com/emberjs/ember.js/pull/14428. Moves the `TestCase` and associated helpers into `internal-test-helpers`.

A couple notes:
1. To avoid needing to re-write all the imports in the glimmer test suite, I opted to re-export any moved helpers in the same files. If we want to clean that up, I think it should be done in a follow up PR.
2. I standardized all the "test" classes to have the suffix `TestCase`; including renaming the base class to `AbstractTestCase`. If this is undesirable, I can switch it back (mainly did it so that everything is consistent).

cc/ @rwjblue


rwjblue:
Thanks @trentmwillis!

PR #14431, opened 2016-10-07
Add tests around hook counts

chadhietala:
This adds assertions for when and also how many times a components hooks
are called. Unlike the other tests in this file these tests are meant to
be used to test the interaction between built-in syntax and the component life-cycle.


stefanpenner:
@wycats / @chancancode could use your 👀 here

wycats:
In principle this looks good. What aspects require the most 👀 ?

mmun:
@chadhietala Should we attempt to salvage any of the tests here or is this safe to close?

PR #14433, opened 2016-10-07
[CLEANUP+BUGFIX] Refactor query params tests

trentmwillis:
This PR refactors the Query Param tests to use a setup similar to the `ember-glimmer` tests. It also adds a few tests that were identified as gaps in the original coverage. The primary goal is to provide a set of abstractions to make writing/maintaining these tests easier.

The follow-up to this PR will be to better organize these tests so that we can more easily identify potential gaps in coverage.

### DIFF
--- a/packages/ember-routing/lib/system/route.js
+++ b/packages/ember-routing/lib/system/route.js
@@ -375,14 +375,15 @@ let Route = EmberObject.extend(ActionHandler, Evented, {
     let transition = this.router.router.activeTransition;
     let state = transition ? transition.state : this.router.router.state;
 
-    let params = {};
     let fullName = getEngineRouteName(getOwner(this), name);
+    let params = assign({}, state.params[fullName]);
+    let queryParams = getQueryParamsFor(route, state);
 
-    assign(params, state.params[fullName]);
-
-    assign(params, getQueryParamsFor(route, state));
-
-    return params;
+    return Object.keys(queryParams).reduce((params, key) => {
+      assert(`The route '${this.routeName}' has both a dynamic segment and query param with name '${key}'. Please rename one to avoid collisions.`, !params[key]);
+      params[key] = queryParams[key];
+      return params;
+    }, params);
   },
 
   /**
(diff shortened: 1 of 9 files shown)


trentmwillis:
This PR is ready to review now. There are a couple skipped tests / fixmes which I am unsure how to address. They've been brought up in #dev-router on Slack and will try to get resolution before moving forward on this.

trentmwillis:
This should be ready to go now. I fixed the above issues based on discussions in Slack (documented as separate commits).

packages/ember-routing/lib/system/route.js
```diff
@@ -375,14 +375,15 @@ let Route = EmberObject.extend(ActionHandler, Evented, {
...
-    assign(params, state.params[fullName]);
-
-    assign(params, getQueryParamsFor(route, state));
-
-    return params;
+    let params = assign({}, state.params[fullName]);
+    let queryParams = getQueryParamsFor(route, state);
+
+    return Object.keys(queryParams).reduce((params, key) => {
+      assert(`The route '${this.routeName}' has both a dynamic segment and query param with name '${key}'. Please rename one to avoid collisions.`, !params[key]);
```
ef4: Could people have gotten away with this before? New assertions smell like new public API that may need to get documented.

trentmwillis: Sort of. This assertion used to be done during `_serializeQueryParams`, which means that it caught the issue the majority of the time, but would miss it if landing on a route with colliding params first (since we didn't serialize the params, they were just given).

Additionally, this should be a slight performance win since this will cache after the first attempt to look up the QPs for a given route. Previously, the collision check happened for every serialization of every link.

trentmwillis: Oh nevermind, thought this was on a different part of the code. Yeah this is a new assertion, we can drop this if needed, but the behavior in this case has never really been defined

rwjblue:
Thanks @trentmwillis!

packages/internal-test-helpers/lib/test-cases/query-param.js
```diff
@@ -0,0 +1,77 @@
...
+
+  get appRouter() {
+    return this.applicationInstance.lookup('router:main');
+  }
+
+  get routerOptions() {
+    return {
+      location: 'test'
+    };
+  }
```
trentmwillis: Will address this in my next PR.

PR #14435, opened 2016-10-07
Ensure constness of this property does not change

stefanpenner:
As it turns out, V8 stores the following information about properties:
- kind: (raw data, or accessor)
- constness: (const or mutable)
- value: Smi, Double, HeapObject, Tagged

"Currently, V8 tracks constant values only for properties containing JS functions."
source: https://docs.google.com/document/d/1VEeXn6BfKmVkYpfMuxNs3otLLNPIxIVA1cLoSwGMIxI/edit#

So rather then making our UNDEFINED sentinal a function, we make it a very
unique string (which we already do in other parts of the system.

---

Note: I am not actually sure this makes much of a difference. I will do more research and see, regardless it does seem like a good cleanup.

### DIFF
--- a/packages/ember-metal/lib/computed.js
+++ b/packages/ember-metal/lib/computed.js
@@ -1,7 +1,7 @@
 import { inspect } from 'ember-utils';
 import { assert, warn } from './debug';
 import { set } from './property_set';
-import { meta as metaFor, peekMeta } from './meta';
+import { meta as metaFor, peekMeta, UNDEFINED } from './meta';
 import expandProperties from './expand_properties';
 import EmberError from './error';
 import {
@@ -21,9 +21,6 @@ import {
 @module ember
 @submodule ember-metal
 */
-
-
-function UNDEFINED() { }
 
 const DEEP_EACH_REGEX = /\.@each\.[^.]+\./;
 
--- a/packages/ember-metal/lib/cache.js
+++ b/packages/ember-metal/lib/cache.js
@@ -1,4 +1,5 @@
 import { EmptyObject } from 'ember-utils';
+import { UNDEFINED } from './meta';
 
 export default class Cache {
   constructor(limit, func, key, store) {
@@ -53,8 +54,6 @@ export default class Cache {
     this.misses = 0;
   }
 }
-
-function UNDEFINED() {}
 
 class DefaultStore {
   constructor() {
--- a/packages/ember-metal/lib/weak_map.js
+++ b/packages/ember-metal/lib/weak_map.js
@@ -1,11 +1,11 @@
 import { GUID_KEY } from 'ember-utils';
 import {
   peekMeta,
-  meta as metaFor
+  meta as metaFor,
+  UNDEFINED
 } from './meta';
 
 let id = 0;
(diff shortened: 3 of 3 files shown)


packages/ember-metal/lib/computed.js
```diff
@@ -23,7 +23,7 @@ import {
 */
 
 
-function UNDEFINED() { }
+export const UNDEFINED = symbol('undefined');
```
stefanpenner: should i used the shared existing UNDEFINED from https://github.com/emberjs/ember.js/blob/27d67841783004538ce0096083b37c6d4e96e824/packages/ember-metal/lib/meta.js#L282 ?

stefanpenner:
I have confirmed this does indeed cause a one-way transition:

```js
const pojo = { };

pojo.apple = 1;
pojo.apple = function( ) { } // [generalizing]apple:s{Any}->t{Any} (+2 maps) [~+317 at index.js:4]
```

This also happens in the other direction


```js
const pojo = { };
pojo.apple = function( ) { }  //[generalizing]apple:c->t{Any} (+1 maps) [~+278 at index.js:6]
pojo.apple = 1
```

tested with: `--trace-generalization`

bmeurer:
Some background on that: The first example also introduces a new map (currently, we're going to change that), but doesn't require instance migration. The second example is way worse, it tracks the function on the map (i.e. doesn't have a field in the object for that), and then has to allocate a field in the object and migrate old instances (this is what we will fix as part of https://docs.google.com/document/d/1VEeXn6BfKmVkYpfMuxNs3otLLNPIxIVA1cLoSwGMIxI).

stefanpenner:
@bmeurer 

> The second example is way worse, it tracks the function on the map (i.e. doesn't have a field in the object for that), and then has to allocate a field in the object and migrate old instances (this is what we will fix as part of

Ya we do this all over the place.. (both transitions)

We may want to `[BUGFIX release]` or `[BUGFIX beta]`this one...

@krisselden  / @rwjblue thoughts?

chancancode:
Probably BUGFIX beta then in a few weeks consider BUGFIX lts (which is the current release)

stefanpenner:
I am going to try and find some time to confirm benefit here. That being said, it really does feel like good cleanup regardless.

bmeurer:
There's no non const -> const transition (usually, setting the object as prototype is one possible exception here). But note that just because you do the const -> non const transitions all over the code base does not necessarily mean that this is a major problem, although it's fairly likely that this will cause some (maybe noticable) overhead.

stefanpenner:
@bmeurer thanks. It would seem that if a member, is most likely to never store a const, we should avoid making it a const, when a non const value also suffices?

bmeurer:
Indeed

stefanpenner:
updated to bugfix beta

krisselden:
👍

bmeurer:
@stefanpenner Yes. Constant tracking only makes sense if you always have exactly only one closure there. It was added for tracking closures for methods on prototypes. In all other cases it will just hurt you (note that we are about to change the machinery here, as outlined in the document).

PR #14442, opened 2016-10-11
[BUGFIX beta] restore didInitAttrs + didReceiveAttrs to POST_INIT

stefanpenner:
- [x] test POST_INIT
- [X] restore hooks to POST_INIT
- [x] tests for `didInitAttrs + didReceiveAttrs` during POST_INIT
- [x] cleanup
- [x] feedback / review / 🚲 🏚 
- [x] test in offending app, going to delegate this to -> @chadhietala  / @krisselden

### DIFF
--- a/packages/ember-glimmer/lib/syntax/curly-component.js
+++ b/packages/ember-glimmer/lib/syntax/curly-component.js
@@ -189,7 +189,7 @@ class CurlyComponentManager {
 
     let klass = definition.ComponentClass;
     let processedArgs = ComponentArgs.create(args);
-    let { attrs, props } = processedArgs.value();
+    let { props } = processedArgs.value();
 
     aliasIdToElementId(args, props);
 
@@ -207,9 +207,6 @@ class CurlyComponentManager {
     if (parentView !== null) {
       parentView.appendChild(component);
     }
-
-    component.trigger('didInitAttrs', { attrs });
-    component.trigger('didReceiveAttrs', { newAttrs: attrs });
 
     // We usually do this in the `didCreateElement`, but that hook doesn't fire for tagless components
     if (component.tagName === '') {
--- a/packages/ember-glimmer/tests/integration/components/curly-components-test.js
+++ b/packages/ember-glimmer/tests/integration/components/curly-components-test.js
@@ -2782,6 +2782,50 @@ moduleFor('Components test: curly components', class extends RenderingTest {
     }, /didInitAttrs called/);
   }
 
+  ['@test did{Init,Receive}Attrs fires after .init() but before observers become active'](assert) {
+    expectDeprecation(/didInitAttrs called/);
+
+    let fooDidChangeCount = 0;
+    let barDidChangeCount = 0;
+
+    this.registerComponent('foo-bar', {
+      ComponentClass: Component.extend({
+        init() {
+          this._super(...arguments);
+          this.didInit = true;
+        },
+        didInitAttrs(attrs) {
+          assert.ok(this.didInit, 'expected init to have run before didInitAttrs');
+          this.set('foo', attrs.foo);
+        },
+
+        willReceieveAttrs(attrs) {
+          assert.ok(this.didInit, 'expected init to have run before willReceieveAttrs');
+          this.set('bar', attrs.bar);
+        },
+
+        fooDidChange: observer('foo', () => { fooDidChangeCount++; }),
+        barDidChange: observer('bar', () => { barDidChangeCount++; })
+      }),
+      template: ''
+    });
+
+    this.render(`{{foo-bar foo=foo bar=bar}}`, { foo: 1, bar: 1 });
+
+    assert.equal(fooDidChangeCount, 0, 'expected NO observer firing for: foo');
+    assert.equal(barDidChangeCount, 0, 'expected NO observer firing for: bar');
+
+    this.runTask(() => set(this.context, 'foo', 2));
+
+    assert.equal(fooDidChangeCount, 1, 'expected observer firing for: foo');
+    assert.equal(barDidChangeCount, 0, 'expected NO observer firing for: bar');
+
+    this.runTask(() => set(this.context, 'bar', 2));
+
+    assert.equal(fooDidChangeCount, 1, 'expected observer firing for: foo');
+    assert.equal(barDidChangeCount, 1, 'expected observer firing for: bar');
+  }
+
   ['@test returning `true` from an action does not bubble if `target` is not specified (GH#14275)'](assert) {
     this.registerComponent('display-toggle', {
       ComponentClass: Component.extend({
--- a/packages/ember-runtime/lib/system/core_object.js
+++ b/packages/ember-runtime/lib/system/core_object.js
@@ -14,7 +14,8 @@ import {
   guidFor,
   generateGuid,
   makeArray,
-  GUID_KEY_PROPERTY
+  GUID_KEY_PROPERTY,
+  symbol
 } from 'ember-utils';
 import {
   assert,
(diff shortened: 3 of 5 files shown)


packages/ember-glimmer/lib/syntax/curly-component.js
```diff
@@ -191,6 +191,7 @@ class CurlyComponentManager {
     let processedArgs = ComponentArgs.create(args);
     let { attrs, props } = processedArgs.value();
 
+    props.__attrs = attrs;
```
stefanpenner: such, WAT

rwjblue: WAT indeed, why are you doing this?

stefanpenner: quick hack, turns out `attrs` is on `this`

packages/ember-glimmer/tests/integration/components/curly-components-test.js
```diff
@@ -2782,6 +2782,60 @@ moduleFor('Components test: curly components', class extends RenderingTest {
     }, /didInitAttrs called/);
   }
 
+  ['@test didInitAttrs is invoked before finishChains'](assert) {
+    expectDeprecation(/didInitAttrs called/);
+
+    let component;
```
chancancode: what do you think if we change the test to...

```js
this.registerComponent('foo-bar', {
  template: '{{foo}}-{{bar}}',

    didInitAttrs({ attrs }) {
      this.set('bar', attrs.foo + 1);
    },

    barDidChange: ...
  })
});

assert.equal(fooDidChangeCount, 0);

this.render('{{foo-bar foo=foo}}', { foo: 1 });

assert.equal(fooDidChangeCount, 0);
this.assertText('1-2');

...

this.runTask(() => set(this.context, 'foo', 2);

...

```

Benefits (imo):

1. avoids the saving off `this` init dance
2. uses the attrs arg
3. (subjective) more meaningful/realistic

packages/ember-glimmer/tests/integration/components/curly-components-test.js
```diff
@@ -2782,6 +2782,60 @@ moduleFor('Components test: curly components', class extends RenderingTest {
...
+
+    assert.equal(fooDidChangeCount, 0);
+    this.render('{{foo-bar}}');
+    assert.equal(fooDidChangeCount, 0);
+    run(() => component.set('foo', 2));
+    assert.equal(fooDidChangeCount, 1);
+  }
+
+  ['@test first willReceieveAttrs is invoked before finishChains'](assert) {
+    let component;
```
chancancode: ditto

packages/ember-glimmer/tests/integration/components/curly-components-test.js
```diff
@@ -2782,6 +2782,60 @@ moduleFor('Components test: curly components', class extends RenderingTest {
...
+      })
+    });
+
+    assert.equal(fooDidChangeCount, 0);
+    this.render('{{foo-bar}}');
+    assert.equal(fooDidChangeCount, 0);
+    run(() => component.set('foo', 2));
+    assert.equal(fooDidChangeCount, 1);
+  }
+
```
chancancode: Can we add a test ~ like this?

```js
// This is not fully supported, but some component unit tests might rely on it?
['@test did{Init,Receive}Attrs fires on init without rendering/appending to DOM']() {
  expectDeprecation(/didInitAttrs called/);

  let didInitAttrsFired = 0;
  let didReceiveAttrsFired = 0;

  let MyComponent = Component.extend({
    didInitAttrs() {
      didInitAttrsFired++;
    },
    didReceiveAttrs() {
      didReceiveAttrsFired++;
    }
  });

  this.component = Component.create();

  // assert  === 1 for both
}
```

chancancode: Just to keep track of the implicit requirements, to make things easier when revisiting in the future

chancancode (changes requested):
👍

packages/ember-glimmer/tests/integration/components/curly-components-test.js
chancancode: To do it properly, we will need to `this.registerComponent('foo-bar', { ComponentClass: ... });` + this.owner._lookupFactory. I don't think we support creating components w/o the injections – but that likely doesn't matter for this test as those restrictions are about ensuring the renderer (etc) are present, and we are explicitly not rendering here. The whole scenario is pretty wat and we should deprecate component unit tests and remove this test ¯\_(ツ)_/¯

packages/ember-glimmer/tests/integration/components/curly-components-test.js
stefanpenner: I would prefer to avoid rendering entirely, the timings this is testing occurs well before rendering.

If there is a better file to put more focused tests, please direct me to it.

stefanpenner: I have removed rendering entirely from the mix.

stefanpenner:
@chancancode tests changed.

packages/ember-glimmer/tests/integration/components/curly-components-test.js
```diff
@@ -2782,6 +2782,54 @@ moduleFor('Components test: curly components', class extends RenderingTest {
...
+      },
+
+      fooDidChange: observer('foo', () => { fooDidChangeCount++; }),
+      barDidChange: observer('bar', () => { barDidChangeCount++; })
+    });
+
+    assert.equal(fooDidChangeCount, 0);
+    assert.equal(barDidChangeCount, 0);
+
+    let component = ComponentClass.create({ renderer: {} });
```
rwjblue: Manually creating components like this is a special case scenario (also tested, but not in this file), we should continue to use `render` to test this in the actual way that we care about this problem.  

Specifically, the thing observed by your work app had nothing to do with manually calling `.create()` (AFAIK), it was specifically that during initial render we trigger events/observers/etc.  That is the thing to test here...

packages/ember-glimmer/tests/integration/components/curly-components-test.js
```diff
@@ -2782,6 +2782,54 @@ moduleFor('Components test: curly components', class extends RenderingTest {
     }, /didInitAttrs called/);
   }
 
+  ['@test did{Init,Receive}Attrs fires after .init() but before finishChains'](assert) {
```
rwjblue: Can you tweak this to avoid the "inside baseball" of `finishChains`?  Would you be ok with something like:

`@test did{Init,Receive}Attrs fires after .init() but before events are triggered and observers are setup`

I fear using `finishChains` will lead to confusion about what this test is actually intended to test over time.

rwjblue (changes requested):

packages/ember-glimmer/tests/integration/components/curly-components-test.js
stefanpenner: the test should be tested via manual creation, which file is better suited for it. I will move this test there.

stefanpenner: I can do both, but according to our own documenting this is public: http://emberjs.com/api/classes/Ember.Component.html#method_create

stefanpenner: @rwjblue I can add another test that tests rendering, but given that this "in the wild" i would prefer to also keep around the very focused test.

rwjblue (approved):

stefanpenner:
got +1 from @chancancode in chat

stefanpenner:
testing the cherry-pick to beta now

stefanpenner:
https://github.com/emberjs/ember.js/commit/c5ac34b5d52cfaabb7a43214f4a9709a33ed90e1

waiting on: https://travis-ci.org/emberjs/ember.js/builds/166819104 for the beta build of this

PR #14446, opened 2016-10-11
Fix attrs test

chancancode:
Turns out the new attrs tests added in #14442 were not working correctly – I noticed this because it was testing the wrong hook. They were only passing because of the attrs -> props alias, so I fixed to test to make sure it is actually testing the thing we intended.

Also added back the unit test replication which got lost in confusion.

cc @stefanpenner @rwjblue


stefanpenner:
:shipit: 

I'll CP to beta once this is green.

PR #14450, opened 2016-10-11
[BUGFIX beta] Fix willRender and willInsertElement ordering and element availability

chadhietala:
This fixes the updating semantics for the `willRender` and `willInsterElement` hooks on a component. Prior to this commit this were not getting called at the correct time.

### DIFF
--- a/packages/ember-glimmer/lib/syntax/curly-component.js
+++ b/packages/ember-glimmer/lib/syntax/curly-component.js
@@ -210,11 +210,14 @@ class CurlyComponentManager {
 
     // We usually do this in the `didCreateElement`, but that hook doesn't fire for tagless components
     if (component.tagName === '') {
+      if (environment.isInteractive) {
+        component.trigger('willRender');
+      }
+
       component._transitionTo('hasElement');
 
       if (environment.isInteractive) {
         component.trigger('willInsertElement');
-        component.trigger('willRender');
       }
     }
 
@@ -225,6 +228,10 @@ class CurlyComponentManager {
     }
 
     processComponentInitializationAssertions(component, props);
+
+    if (environment.isInteractive && component.tagName !== '') {
+      component.trigger('willRender');
+    }
 
     return bucket;
   }
@@ -290,7 +297,6 @@ class CurlyComponentManager {
 
     if (environment.isInteractive) {
       component.trigger('willInsertElement');
-      component.trigger('willRender');
     }
   }
 
@@ -366,11 +372,14 @@ class TopComponentManager extends CurlyComponentManager {
 
     // We usually do this in the `didCreateElement`, but that hook doesn't fire for tagless components
     if (component.tagName === '') {
+      if (environment.isInteractive) {
+        component.trigger('willRender');
+      }
+
       component._transitionTo('hasElement');
 
       if (environment.isInteractive) {
         component.trigger('willInsertElement');
-        component.trigger('willRender');
       }
     }
 
--- a/packages/ember-glimmer/tests/integration/components/life-cycle-test.js
+++ b/packages/ember-glimmer/tests/integration/components/life-cycle-test.js
@@ -3,6 +3,7 @@ import { Component } from '../../utils/helpers';
 import { strip } from '../../utils/abstract-test-case';
 import { moduleFor, RenderingTest } from '../../utils/test-case';
 import { getViewId, getViewElement } from 'ember-views';
+import { classes } from '../../utils/test-helpers';
 
 class LifeCycleHooksTest extends RenderingTest {
   constructor() {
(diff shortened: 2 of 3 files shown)


packages/ember-glimmer/lib/syntax/curly-component.js
```diff
@@ -210,11 +210,14 @@ class CurlyComponentManager {
 
     // We usually do this in the `didCreateElement`, but that hook doesn't fire for tagless components
     if (component.tagName === '') {
+      if (environment.isInteractive) {
+        component.trigger('willRender');
+      }
```
chancancode: I think you need to move this into `create`? I thought the point is you want to `set` on the component before we apply the bindings

packages/ember-glimmer/lib/syntax/curly-component.js
```diff
@@ -259,6 +262,10 @@ class CurlyComponentManager {
   }
 
   didCreateElement({ component, classRef, environment }, element, operations) {
+    if (environment.isInteractive) {
+      component.trigger('willRender');
+    }
```
chancancode: Move this up to `create` and remove the `tagName` guard

packages/ember-glimmer/tests/integration/components/life-cycle-test.js
```diff
@@ -1378,8 +1381,64 @@ moduleFor('Run loop and lifecycle hooks', class extends RenderingTest {
...
   }
+
+  ['@test element availability in `willInsertElement`'](assert) {
+    let ComponentClass = Component.extend({
+      tagName: 'a',
+      customHref: 'http://google.com',
+      attributeBindings: ['customHref:href'],
+      willInsertElement() {
+        assert.ok(this.element, 'should have element');
+        assert.notOk(this.element.parentNode, 'is detached');
```
chancancode: Isn't this already covered in the lifecycle tests?

packages/ember-glimmer/tests/integration/components/life-cycle-test.js
```diff
@@ -1378,8 +1381,64 @@ moduleFor('Run loop and lifecycle hooks', class extends RenderingTest {
...
+        id: 'foo',
+        href: 'http://willRender.com',
+        class: classes('ember-view')
+      }
+    });
+  }
 });
 
+
+
```
chancancode: ✂️

packages/ember-glimmer/tests/integration/components/life-cycle-test.js
```diff
@@ -1378,8 +1381,64 @@ moduleFor('Run loop and lifecycle hooks', class extends RenderingTest {
...
+    });
+  }
+
+  ['@test `willRender` can set before render'](assert) {
+    let ComponentClass = Component.extend({
+      tagName: 'a',
+      customHref: 'http://google.com',
+      attributeBindings: ['customHref:href'],
+      willRender() {
+        assert.notOk(this.element, 'should not have element');
```
chancancode: Probably ✂️ this assertion ? It's already covered by the general lifecycle tests, so this just distracts from the point of the test imo

packages/ember-glimmer/tests/integration/components/life-cycle-test.js
```diff
@@ -1378,8 +1381,64 @@ moduleFor('Run loop and lifecycle hooks', class extends RenderingTest {
...
+      tagName: 'a',
+      attrs: {
+        id: 'foo',
+        href: 'http://google.com',
+        class: classes('ember-view')
+      }
+    });
+  }
+
+  ['@test `willRender` can set before render'](assert) {
```
chancancode: Could use a comment, or link to github issue

packages/ember-glimmer/lib/syntax/curly-component.js
chadhietala: I actually don't know if we want to move this.

packages/ember-glimmer/tests/integration/components/life-cycle-test.js
chadhietala: Removed.

packages/ember-glimmer/tests/integration/components/life-cycle-test.js
chadhietala: Removed

packages/ember-glimmer/lib/syntax/curly-component.js
chancancode: I think the difference is probably not observable. All that would have happened between the end of create and here is that we would have compiled the layout, entered it, created an empty element for in the VM (but haven't appended it into DOM). So it should be okay to align this particular book between tagless and regular components.

chancancode: Do you have other concerns? I could totally have missed something

chadhietala:
Fixes https://github.com/emberjs/ember.js/issues/14458

packages/ember-glimmer/lib/syntax/curly-component.js
chadhietala: Actually did some more investigation. Moved this and updated the tests accordingly.

packages/ember-glimmer/tests/integration/components/life-cycle-test.js
chadhietala: Done

PR #14453, opened 2016-10-12
[BUGFIX beta] [FIXES #14447] fix didDestroyElement

### LINKED ISSUE #14447: [BUG] Components via appendTo still affected by #14439
https://github.com/emberjs/ember.js/pull/14442 addressed #14439 but the issue is also in the other mode.

I'm working to:

- [x] didInitAttr / didRetrieveAttrs only once: #14448
- [ ] destroy hooks (didDestroyElement first before willDestroyElement for TopComponent managed components

* enhancing the existing lifecycle checks to ensure this is covered
* explore missing tests (we may be missing proper coverage for these types of components in the lifecycle tests)
* fix by removing the duplicate hooks

stefanpenner:
Components via appendTo, must behave the same when torn down via destroy, and the regular rendering system tearing them down. This code-path isn’t as first class as it likely should be, as it only exists for supporting the more legacy style of unit testing. Using the component integration tests, the typical code path is taken.

---

_Note: this appears to work, but I don't really like it. Open to suggestion_

---
- [x] restore tests
- [x] make it work
- [x] get feedback 
- [x] explore making it suck less?

---

Marvin is not amused by this humans lolcode.

photo on 10-11-16 at 7 29 pm 2

### DIFF
--- a/packages/ember-glimmer/lib/renderer.js
+++ b/packages/ember-glimmer/lib/renderer.js
@@ -92,6 +92,7 @@ class RootState {
     this.destroyed = true;
 
     this.env = null;
+    let root = this.root;
     this.root = null;
     this.result = null;
     this.render = null;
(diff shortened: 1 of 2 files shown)


chadhietala:
LGTM.

krisselden:
I think it is important to get this fixed, I'm not crazy about it being called in 2 paths, but I'm not sure that is the highest priority right now so 👍

PR #14454, opened 2016-10-12
initial d8-runner ready to go

stefanpenner:


krisselden:
I don't understand why the node dom helper isn't being chosen.  You shouldn't need the dom patches, and they will not work if something actually uses `{{{html}}}`

krisselden:
Otherwise, awesome work.

stefanpenner:
> I don't understand why the node dom helper isn't being chosen. You shouldn't need the dom patches, and they will not work if something actually uses {{{html}}}

I was having a hard time debugging in D8. Maybe we can sit down later today or tomorrow and address what you mentioned

rwjblue:
@stefanpenner  - Any reason not to land this and iterate further while on master?

stefanpenner:
@rwjblue no reason, other then. we should then likely open an issue for

> We obviously want to make sure this uses the Node dom helper stuff, i don't think it is yet.

stefanpenner:
https://github.com/emberjs/ember.js/issues/14551

PR #14455, opened 2016-10-12
[BUGIFX beta] Ensure constness of this property does not change

stefanpenner:
As it turns out, V8 stores the following information about properties:
- kind: (raw data, or accessor)
- constness: (const or mutable)
- value: Smi, Double, HeapObject, Tagged

"Currently, V8 tracks constant values only for properties containing JS functions."
source: https://docs.google.com/document/d/1VEeXn6BfKmVkYpfMuxNs3otLLNPIxIVA1cLoSwGMIxI/edit#

So rather then making our UNDEFINED sentinal a function, we make it a very
unique string (which we already do in other parts of the system).

### DIFF
--- a/packages/ember-metal/lib/cache.js
+++ b/packages/ember-metal/lib/cache.js
@@ -1,4 +1,5 @@
 import { EmptyObject } from 'ember-utils';
+import { UNDEFINED } from './meta';
 
 export default class Cache {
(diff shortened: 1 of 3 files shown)


stefanpenner:
waiting for tests to run against Beta

cc @chancancode

PR #14457, opened 2016-10-12
ember-source related cleanup

### LINKED ISSUE #14434: ember-source related cleanup
- [x] Fix `prepublish` script in ember to do `ember build -prod`
- [x] Fix `.npmignore` (we are including waaaaaayyyyy to much in the tar.gz right now)
- [x] Fix `ember-cli-htmlbars` for the ember-source name change https://github.com/ember-cli/ember-cli-htmlbars/pull/98

locks:
Fixes #14434. See issue for checklist.
- [x] Fix `prepublish` script in ember to do `ember build -prod`
- [x] Fix `.npmignore` (we are including waaaaaayyyyy to much in the tar.gz right now)
- [x] Fix `ember-cli-htmlbars` for the ember-source name change https://github.com/ember-cli/ember-cli-htmlbars/pull/98
- [x] use node + `execSync` for release script

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,1 +1,33 @@
 node-tests/
+CODE_OF_CONDUCT.md
+CONTRIBUTING.md
+FEATURES.md
+STYLEGUIDE.md
+VERSION
+bin
+blueprints
+bower.json
+bower_components
+config
+ember-cli-build.js
+ember-source.gemspec
+features.json
+generators
+lib
+node-tests
+node_modules
+packages
+scripts
+server
+testem.dist.json
+testem.json
+testem.travis-browsers.js
+tests
+tmp
+vendor
+!vendor/ember/
+yarn.lock
+yuidoc.json
+
+# In the future we will likely want to restrict exactly
+# which parts of `/dist` we want to include.
--- a/scripts/release.js
+++ b/scripts/release.js
@@ -1,0 +1,4 @@
+var execSync = require('child_process').execSync;
+
+execSync("ember build --environment production");
+execSync("npm publish");
--- a/package.json
+++ b/package.json
@@ -14,7 +14,8 @@
     "test:blueprints": "node node-tests/nodetest-runner.js",
     "start": "ember serve",
     "docs": "ember ember-cli-yuidoc",
-    "sauce:launch": "ember sauce:launch"
+    "sauce:launch": "ember sauce:launch",
+    "release": "node scripts/release.js"
   },
   "repository": {
     "type": "git",


.npmignore
```diff
@@ -1 +1,32 @@
...
+STYLEGUIDE.md
+VERSION
+bin
+blueprints
+bower.json
+bower_components
+config
+# dist
+ember-cli-build.js
+ember-source-2.10.0-alpha.1.tgz
```
btecu: Should this be something like `ember-source-*.tgz` or will be necessary to update after each version?

.npmignore
```diff
@@ -1 +1,32 @@
...
+CONTRIBUTING.md
+FEATURES.md
+STYLEGUIDE.md
+VERSION
+bin
+blueprints
+bower.json
+bower_components
+config
+# dist
```
locks: It's just a comment. In the future we might want to only ship certain things inside `/dist`

.npmignore
locks: Removed, was a left over from copying the `ls` output

mitchlloyd:
If we're omitting most of the files from npm, we may want to consider using the package.json "files" configuration over npmignore.

https://docs.npmjs.com/files/package.json#files

nathanhammond:
@mitchlloyd We did investigate and consider that. @locks probably remembers the reasoning for that not being a thing.

locks:
From what I understand, `files` has priority over `.npmignore`. That coupled with `files` only allowing file inclusion and not exclusion led to my understanding that if (when? 😂 ) we get trolled by `.npmignore`, it should be in a way that doesn't break things for the user.

scripts/release.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+ember build --production && npm publish
```
nathanhammond: Turn this into a JavaScript file and `execSync` and this will be portable. I'd prefer not to make the release process non-portable.

.npmignore
```diff
@@ -1 +1,34 @@
...
+VERSION
+bin
+blueprints
+bower.json
+bower_components
+config
+# In the future we will likely want to restrict exactly
+# which parts of `/dist` we want to include.
+#
+# dist
```
nathanhammond: I feel like we should split this out to the very bottom of the file. Especially since it's commented to begin with. It can just be presented as a documentation of:

> Currently we include the entire `dist` folder, but at some point we may wish to be more selective about what we include.

nathanhammond (changes requested):
I feel like there should be an (Ember CLI?) test which runs `npm pack ember.js && npm install ember-source*.tgz && npm test` to protect us from possibly doing the wrong things here.

scripts/release.sh
locks: Do you have an `execSync` example handy?

nathanhammond: Nope. https://nodejs.org/api/child_process.html#child_process_child_process

josemarluedke: `--production` is not a valid option. It should be `ember build --environment production` or `ember build -prod`.

nathanhammond:
@rwjblue I believe this is ready to land. The only thing it doesn't currently have is a test for the assets which get published to ensure that all of the pieces we need get published. I feel like that test likely belongs on the Ember CLI side and not part of this repo so I'm in favor of landing this.

@locks Thank you for getting this to work in a portable manner and helping to push this over the line. <3

scripts/release.js
```diff
@@ -0,0 +1,4 @@
+var execSync = require('child_process').execSync;
+
+execSync("ember build --environment production");
+execSync("npm publish");
```
nathanhammond: This actually seems a bit scary to auto-trigger. I feel like we just need:
`execSync("ember build --environment production");`

And that it should run as part of the `prepublish` hook.

PR #14459, opened 2016-10-12
[BUGFIX beta] [FIXES #14447] fix didDestroyElement

### LINKED ISSUE #14447: [BUG] Components via appendTo still affected by #14439

stefanpenner:
Components via appendTo, must behave the same when torn down via destroy, and the regular rendering system tearing them down. This code-path isn’t as first class as it likely should be, as it only exists for supporting the more legacy style of unit testing. Using the component integration tests, the typical code path is taken.

### DIFF
--- a/packages/ember-glimmer/lib/renderer.js
+++ b/packages/ember-glimmer/lib/renderer.js
@@ -92,6 +92,7 @@ class RootState {
     this.destroyed = true;
 
     this.env = null;
+    let root = this.root;
     this.root = null;
     this.result = null;
     this.render = null;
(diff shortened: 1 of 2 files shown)


stefanpenner:
getting CI to run on beta branch before this cherry-pick lands

PR #14461, opened 2016-10-12
[BUGFIX beta] Per-property tagging

chadhietala:
/cc @wycats @chancancode @krisselden 

Note this is targeting the beta branch


kanongil:
Awesome. This seems to fix the major performance regressions I have experienced with Glimmer2. Though, it still seems the `hash` and `action` helpers can cause redundant re-validations, see #14410 & #14335.

rwjblue:
This seems good to me.

@krisselden - Any objections?

rwjblue:
Also, just to note, this targets beta branch (2.9) and https://github.com/emberjs/ember.js/pull/14464 has been submitted to target master.

PR #14465, opened 2016-10-13
Move IS_PROXY flag into meta

chadhietala:
/cc @chancancode @krisselden @wycats @stefanpenner

### DIFF
--- a/packages/ember-metal/lib/is_proxy.js
+++ b/packages/ember-metal/lib/is_proxy.js
@@ -1,7 +1,10 @@
-import { symbol } from 'ember-utils';
-
-export const IS_PROXY = symbol('IS_PROXY');
+import { peekMeta } from './meta';
 
 export function isProxy(value) {
-  return typeof value === 'object' && value && value[IS_PROXY];
+  if (typeof value === 'object' && value) {
+    let meta = peekMeta(value);
+    return meta && meta.isProxy();
+  }
+
+  return false;
 }
--- a/packages/ember-metal/lib/meta.js
+++ b/packages/ember-metal/lib/meta.js
@@ -58,9 +58,11 @@ let members = {
   tags: ownMap
 };
 
+// FLAGS
 const SOURCE_DESTROYING = 1 << 1;
 const SOURCE_DESTROYED = 1 << 2;
 const META_DESTROYED = 1 << 3;
+const IS_PROXY = 1 << 4;
 
 if (isEnabled('ember-glimmer-detect-backtracking-rerender') ||
     isEnabled('ember-glimmer-allow-backtracking-rerender')) {
@@ -187,6 +189,14 @@ Meta.prototype.isMetaDestroyed = function isMetaDestroyed() {
 
 Meta.prototype.setMetaDestroyed = function setMetaDestroyed() {
   this._flags |= META_DESTROYED;
+};
+
+Meta.prototype.isProxy = function isProxy() {
+  return (this._flags & IS_PROXY) !== 0;
+};
+
+Meta.prototype.setProxy = function setProxy() {
+  this._flags |= IS_PROXY;
 };
 
 // Implements a member that is a lazily created, non-inheritable
--- a/packages/ember-runtime/lib/mixins/-proxy.js
+++ b/packages/ember-runtime/lib/mixins/-proxy.js
@@ -21,7 +21,6 @@ import {
   Mixin,
   observer,
   tagFor,
-  IS_PROXY
 } from 'ember-metal';
 import { bool } from '../computed/computed_macros';
 
@@ -71,8 +70,6 @@ class ProxyTag extends CachedTag {
   @private
 */
 export default Mixin.create({
-  [IS_PROXY]: true,
-
   /**
     The object whose properties will be forwarded.
 
@@ -82,6 +79,11 @@ export default Mixin.create({
     @private
   */
   content: null,
+
+  init() {
+    this._super(...arguments);
+    meta(this).setProxy();
+  },
 
   _initializeTag: on('init', function() {
     meta(this)._tag = new ProxyTag(this);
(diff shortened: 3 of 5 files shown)


packages/ember-metal/lib/is_proxy.js
```diff
@@ -1,7 +1,10 @@
-import { symbol } from 'ember-utils';
-
-export const IS_PROXY = symbol('IS_PROXY');
+import { meta as metaFor } from './meta';
 
 export function isProxy(value) {
-  return typeof value === 'object' && value && value[IS_PROXY];
+  if (typeof value === 'object' && value) {
+    let meta = metaFor(value);
```
rwjblue: I think we should use `peekMeta` (since there is no need to create a meta if one isn't already found).

Something like:

```js
let meta = peekMeta(value);
return typeof meta === 'object' && meta !== null && meta.isProxy()
```

packages/ember-runtime/lib/mixins/-proxy.js
```diff
@@ -83,6 +80,10 @@ export default Mixin.create({
   */
   content: null,
 
+  _initializeProxy: on('init', function() {
+    meta(this).setProxy();
```
rwjblue: Any reason not to put this in `init` method directly?

rwjblue (changes requested):

packages/ember-runtime/lib/mixins/-proxy.js
chadhietala: Yea, I was work or less following the pattern with the tag initialization. I will move both of them to `init` so it will be easier to move away (one day) from the eventing stuff.

rwjblue: Ahh, actually the tag initialization needs to stay as an on('init') still, this is because content can be set in init (after _super is called by folks extending). It's "special" here (maybe it warrants a comment to make it clearer for future travelers)?

packages/ember-metal/lib/is_proxy.js
chadhietala: Fixed.

packages/ember-runtime/lib/mixins/-proxy.js
chadhietala: Fixed now

packages/ember-metal/lib/is_proxy.js
chadhietala: I have to see why this is failing now in browsers.

packages/ember-runtime/lib/mixins/-proxy.js
```diff
@@ -83,9 +79,11 @@ export default Mixin.create({
   */
   content: null,
 
-  _initializeTag: on('init', function() {
+  init() {
+    this._super(...arguments);
+    meta(this).setProxy();
     meta(this)._tag = new ProxyTag(this);
```
chancancode: This has to happen after `init` otherwise you could miss `content` being set inside the user `init`

packages/ember-metal/lib/meta.js
```diff
@@ -189,6 +191,14 @@ Meta.prototype.setMetaDestroyed = function setMetaDestroyed() {
   this._flags |= META_DESTROYED;
 };
 
+Meta.prototype.isProxy = function isProxy() {
+  return (this._flags & IS_PROXY) !== 0;
+};
+
+Meta.prototype.setProxy = function isProxy() {
```
chancancode: `function setProxy`

packages/ember-runtime/lib/mixins/-proxy.js
```diff
@@ -83,9 +79,11 @@ export default Mixin.create({
   */
   content: null,
 
-  _initializeTag: on('init', function() {
+  init() {
+    this._super(...arguments);
+    meta(this).setProxy();
```
rwjblue: Let's cache the meta here:

```
let m = meta(this);
m.setProxy();
m._tag = new ProxyTag(this)
```

It isn't a huge deal, just saves a little extra work.

packages/ember-runtime/lib/mixins/-proxy.js
rwjblue: This still needs to be addressed.

packages/ember-runtime/lib/mixins/-proxy.js
rwjblue: Hmm. Actually, I just noticed that we still need to fix the issue that @chancancode mentioned below.

The tag needs to be set on meta in an on('init').

rwjblue (changes requested):

chadhietala:
I think I've addressed everything here.

packages/ember-metal/lib/is_proxy.js
```diff
@@ -1,7 +1,10 @@
-import { symbol } from 'ember-utils';
-
-export const IS_PROXY = symbol('IS_PROXY');
+import { peekMeta } from './meta';
 
 export function isProxy(value) {
-  return typeof value === 'object' && value && value[IS_PROXY];
+  if (typeof value === 'object' && value) {
+    let meta = peekMeta(value);
+    return typeof meta === 'object' && meta !== null && meta.isProxy();
```
chancancode: does peekMeta ever return anything other than null or the meta object?

chancancode (approved):
Looks good otherwise!

packages/ember-metal/lib/is_proxy.js
chancancode: Right, I was asking if we could just check for either === null or undefined (whatever it actually returns when there is no meta) instead of he generic check we usually do.

rwjblue: Correct, we could simplify to just be `return meta !== undefined && meta.isProxy()`  or something like that.  

@chadhietala - Sorry for the wild-goose chase (I realize this snippet came from my suggestion in a prior comment).

rwjblue (approved):

PR #14466, opened 2016-10-13
[CLEANUP] Remove qp.prefix and use qp.controllerName

trentmwillis:
This PR is being made purely to help clarify/simplify some internal logic for QPs.

`prefix` was only ever set to the `ctrl` property, so removing that helps
simplify the code. Renaming `ctrl` to `controllerName` for added clarity.


stefanpenner:
👍

rwjblue (approved):
LGTM, will merge when green...

PR #14471, opened 2016-10-13
Support query params in asynchronously loaded routes

trentmwillis:
This PR enables support for query params in routes that are lazily loaded while not altering existing behavior. There are lots of details about the changes in the individual commit messages.

The bulk of the changes center on two key themes:
1. Not using the `handler` (route) before it is absolutely needed
2. Making it possible to decouple the Router's logic from the presence of Routes

### DIFF
--- a/packages/ember-routing/lib/system/route.js
+++ b/packages/ember-routing/lib/system/route.js
@@ -421,10 +421,7 @@ let Route = EmberObject.extend(ActionHandler, Evented, {
     // urlKey isn't used here, but anyone overriding
     // can use it to provide serialization specific
     // to a certain query param.
-    if (defaultValueType === 'array') {
-      return JSON.stringify(value);
-    }
-    return `${value}`;
+    return this.router._serializeQueryParam(value, defaultValueType);
   },
 
   /**
@@ -440,17 +437,7 @@ let Route = EmberObject.extend(ActionHandler, Evented, {
     // urlKey isn't used here, but anyone overriding
     // can use it to provide deserialization specific
     // to a certain query param.
-
-    // Use the defaultValueType of the default value (the initial value assigned to a
-    // controller query param property), to intelligently deserialize and cast.
-    if (defaultValueType === 'boolean') {
-      return (value === 'true') ? true : false;
-    } else if (defaultValueType === 'number') {
-      return (Number(value)).valueOf();
-    } else if (defaultValueType === 'array') {
-      return emberA(JSON.parse(value));
-    }
-    return value;
+    return this.router._deserializeQueryParam(value, defaultValueType);
   },
 
   /**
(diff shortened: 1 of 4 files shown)


packages/ember/tests/routing/query_params_test/query_param_async_get_handler_test.js
```diff
@@ -0,0 +1,80 @@
...
+
+import { QueryParamTestCase, moduleFor } from 'internal-test-helpers';
+
+moduleFor('Query Params - async get handler', class extends QueryParamTestCase {
+  get routerOptions() {
+    return {
+      location: 'test',
+      _getHandlerFunction() {
+        let getHandler = this._super(...arguments);
+        return (routeName) => Promise.resolve().then(() => getHandler(routeName));
```
rwjblue: Likely should use off of `RSVP.Promise` (since we support platforms without native Promise).  Also, when you make that change you should use "real" async and make a helper method to resolve in a `setTimeout` or something...

rwjblue (changes requested):
Changes mostly look good, only one small nit-pick

packages/ember/tests/routing/query_params_test/query_param_async_get_handler_test.js
trentmwillis: The problem with always deving in Chrome...will fix. I'll be sure to add a small setTimeout. Still planning to add several more tests.

trentmwillis:
@rwjblue made the requested changes. I have at least one more test to write and want to verify this works in the ember-engines repo, but the functional code is pretty much done. Also added a lot of docs around what is going on.

packages/ember-routing/lib/system/router.js
```diff
@@ -653,22 +654,94 @@ const EmberRouter = EmberObject.extend(Evented, {
...
+    @method _serializeQueryParam
+    @param {Object} value
+    @param {String} type
+  */
+  _serializeQueryParam(value, type) {
+    if (type === 'array') {
+      return JSON.stringify(value);
+    }
+
+    return `${value}`;
```
trentmwillis: This was copied form the original function in route.js. That said, I actually like this as a way of coercing a string value.

packages/ember-routing/lib/system/route.js
```diff
@@ -142,14 +142,37 @@ let Route = EmberObject.extend(ActionHandler, Evented, {
...
 
   /**
+    Sets the name for this route, including a fully resolved name for routes
+    inside engines.
+
+    @private
+    @method setRouteName
+    @param {String} name
+  */
+  setRouteName(name) {
```
rwjblue: We should use a symbol or underscore any newly added private methods to make it super obvious that they are private (and avoid naming collisions with userland methods/properties).

rwjblue (changes requested):

rwjblue:
Awesome, thanks @trentmwillis!

Issue #14469, opened 2016-10-13
Support `target` option for closure actions


Panman8201:
It seems when using a closure`(action)` the `target` option is not respected in the same way as with the "traditional" helper `{{action}}`. My main need is for the called action/function to have the `this` context of object itself, not the caller. The most common use case is when you're trying to pass in an action into an ember component. Ex:

``` hbs
{{#my-button click=(action someObj.funcName target=someObj)}}
  Do Action
{{/my-button}}
```

However, at the same time if you're able to use the "traditional" helper, then it works. Ex:

``` hbs
<button {{action someObj.funcName target=someObj}}>
  Do Action
</button>
```

I have a [working example in an Ember Twiddle](https://ember-twiddle.com/acc1be18eb9213778e1e2632a4a10f27?fileTreeShown=false&numColumns=2&openFiles=controllers.application.js%2Ctemplates.application.hbs). But for reference here, the called function would look something like:

``` js
let someObj = Ember.Object.create({
  propName: 'foobar',
  funcName() {
    // Do something with `this`
    this.get('propName');
  }
});
```

In chatting with @rwjblue and company in the Slack community, apparently this can be handled in the [action AST file](https://github.com/emberjs/ember.js/blob/v2.9.0-beta.5/packages/ember-template-compiler/lib/plugins/transform-action-syntax.js), which should then be able to eliminate some ["ugly code" in the action modifier](https://github.com/emberjs/ember.js/blob/v2.9.0-beta.5/packages/ember-glimmer/lib/modifiers/action.js#L95-L106). I'm still trying to wrap my head around the internals and AST stuff, but at least there is direction in where to start.

Panman8201:
Looking back at more Slack chat, Robert provided even more direction which is great!

https://embercommunity.slack.com/archives/-help/p1476381597028760

> https://github.com/emberjs/ember.js/blob/v2.9.0-beta.5/packages/ember-template-compiler/lib/plugins/transform-action-syntax.js#L65-L67 needs to be rewritten to search the `hash` arguments for `target` and use that if present

> then we can completely remove https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/modifiers/action.js#L95-L106

/CC @sukima since you were involved with that particular convo.

zackthehuman:
This makes sense to me. If I understand the suggestion -- basically we would change the AST plugin to check for a `target` hash arg, and if there isn't one we just create one that does `target=this`.

[...]

PR #14479, opened 2016-10-16
Make actions use `target=this` implicitly

zackthehuman:
This addresses the bug in #14469. Might need a new test. Please review @rwjblue @chancancode.

### DIFF
--- a/packages/ember-template-compiler/lib/plugins/transform-action-syntax.js
+++ b/packages/ember-template-compiler/lib/plugins/transform-action-syntax.js
@@ -1,3 +1,5 @@
+import hashPairForKey from '../system/hash-pair-for-key';
+
 /**
  @module ember
  @submodule ember-glimmer
@@ -15,9 +17,9 @@
   with
 
   ```handlebars
- <button {{action target=this 'foo'}}>
- <button onblur={{action target=this 'foo'}}>
- <button onblur={{action target=this (action target=this 'foo') 'bar'}}>
+ <button {{action 'foo' target=this}}>
+ <button onblur={{action 'foo' target=this}}>
+ <button onblur={{action (action 'foo' target=this) 'bar' target=this}}>
   ```
 
   If an action already has a target it is left unmodified.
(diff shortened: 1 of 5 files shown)


zackthehuman:
I added a test.

packages/ember-template-compiler/lib/plugins/transform-action-syntax.js
```diff
@@ -15,11 +15,13 @@
   with
 
   ```handlebars
- <button {{action this 'foo'}}>
- <button onblur={{action this 'foo'}}>
- <button onblur={{action this (action this 'foo') 'bar'}}>
+ <button {{action target=this 'foo'}}>
```
rwjblue: This isn't quite right, the hash has to be after positional args...

rwjblue (changes requested):

packages/ember-template-compiler/lib/plugins/transform-action-syntax.js
```diff
@@ -15,11 +15,13 @@
...
+ <button onblur={{action target=this 'foo'}}>
+ <button onblur={{action target=this (action target=this 'foo') 'bar'}}>
```
sukima: Isn't this also backwards.

packages/ember-template-compiler/lib/plugins/transform-action-syntax.js
```diff
@@ -62,6 +67,24 @@ function isAction(node) {
...
+function hashPairForKey(hash, key) {
+  for (let i = 0; i < hash.pairs.length; i++) {
+    let pair = hash.pairs[i];
+    if (pair.key === key) {
+      return pair;
+    }
+  }
+
+  return false;
+}
```
sukima: This function is unfortunately repeated. @rwjblue should the util function be DRY somewhere? And if so where?

packages/ember-template-compiler/lib/plugins/transform-action-syntax.js
```diff
@@ -62,6 +64,20 @@ function isAction(node) {
...
+  }
+}
+
+function hashPairForKey(hash, key) {
+  for (let i = 0; i < hash.pairs.length; i++) {
+    let pair = hash.pairs[i];
+    if (pair.key === key) {
+      return pair;
+    }
+  }
```
sukima: This function is unfortunately repeated. @rwjblue should the util function be DRY somewhere? And if so where?

zackthehuman: I will extract it and then make another PR to update other refs.

packages/ember-template-compiler/lib/plugins/transform-action-syntax.js
zackthehuman: Yes; it has been corrected.

Serabe:
@zackthehuman is this ready for a new review? Thank you!

zackthehuman:
@Serabe I think we may punt on this due to what was discussed in #14469. Closing for now.

Serabe:
@zackthehuman thank you!

Issue #14158, opened 2016-08-30
Deprecate `renderToElement`


chancancode:

runspired:
@chancancode so long as there is still a way to manually instantiate a component and render it :D

miguelcobain:
@runspired there is still `appendTo` as I understand it.

Gaurav0:
appendTo is private. Please do not deprecate this unless there is another public way to render a component programmatically.

miguelcobain:
`renderToElement` is public? I'd be surprised if it was.

rwjblue:
@Gaurav0 - As mentioned in https://github.com/emberjs/ember.js/pull/14482#issuecomment-254623792, neither is public today.  Would love an RFC to make a public mechanism for programmatic creation of components.

Also, lets centralize conversation on the topic instead of spreading it all over the place (making it hard for others to follow)?

rwjblue:
Closed by #14482

PR #14482, opened 2016-10-16
Deprecate renderToElement

josemarluedke:
This implements #14158. 

Deprecation guide: emberjs/website#2702.

### DIFF
--- a/packages/ember-views/lib/mixins/view_support.js
+++ b/packages/ember-views/lib/mixins/view_support.js
@@ -215,10 +215,21 @@ export default Mixin.create({
     @method renderToElement
     @param {String} tagName The tag of the element to create and render into. Defaults to "body".
     @return {HTMLBodyElement} element
+    @deprecated Use appendTo instead.
     @private
   */
   renderToElement(tagName) {
     tagName = tagName || 'body';
+
+    deprecate(
+      `Using the \`renderToElement\` is deprecated in favor of \`appendTo\`. Called in ${this.toString()}`,
+      false,
+      {
+        id: 'ember-views.render-to-element',
+        until: '2.12.0',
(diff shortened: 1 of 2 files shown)


packages/ember-views/lib/mixins/view_support.js
```diff
@@ -215,11 +215,22 @@ export default Mixin.create({
...
   */
   renderToElement(tagName) {
     tagName = tagName || 'body';
 
+    deprecate(
+      `Using the \`renderToElement\` is deprecated in favor of \`appendTo\`. Called in ${this.toString()}`,
+      false,
+      {
+        id: 'ember-views.render-to-element',
+        until: '3.0.0',
```
josemarluedke: Not sure if 3.0.0 is the intended version to drop this, can someone confirm?

Gaurav0:
`appendTo` is private. Please do not merge this unless there is another public way to render a component programmatically.

rwjblue:
> appendTo is private. Please do not merge this unless there is another public way to render a component programmatically

@Gaurav0 - That really isn't a compelling motivation. Both `renderToElement` and `appendTo` are private, this is just deprecating one private mechanism for another.

I totally agree that we should have a public API for this though, maybe you can try your hand at writing the RFC if this is something you need?

packages/ember-views/lib/mixins/view_support.js
rwjblue: I believe this would be `until: '2.12.0'`

chancancode: 👍

josemarluedke: Will update this then.

josemarluedke: Updated!

rwjblue:
Thank you @josemarluedke!

PR #14485, opened 2016-10-17
remove _toString, always use NAME_KEY

stefanpenner:
no longer:
- have two concepts
- processAllNamespaces + find NAME_KEY if the name was pre-determined by the container (previously _toString was set, but not NAME_KEY, which made this code assume it must calculate the name again)
- change shape of `this` with a constant descriptor of toString
- have _toString (NAME_KEY is now used)

---

I noticed ember-data's test suite was spending an unexpected amount of time in processAllNamespaces, which lead me to realize `_toString` is not consider  in classToString, but it really should have been.


stefanpenner:
@krisselden r?

packages/container/lib/container.js
```diff
@@ -1,6 +1,6 @@
 import { dictionary, symbol, setOwner, OWNER } from 'ember-utils';
 import { ENV } from 'ember-environment';
-import { assert, deprecate, runInDebug } from 'ember-metal';
+import { assert, deprecate, runInDebug, NAME_KEY } from 'ember-metal';
```
rwjblue: Can you move `NAME_KEY` (which I think is just a symbol right?) into the `ember-utils` package instead?  We are ultimately trying to make `container` package not rely on ember-metal (the `assert`, `deprecate`, `runInDebug` functionality was the only remaining bit).

stefanpenner: sure

stefanpenner: i made the change, but something after moving that stuff around has broken things ... need to figure out what

rwjblue:
This looks good to me...

PR #14486, opened 2016-10-17
Support `target` option for closure actions

sukima:
This is a spike to support the `target` hash argument for closure actions.

**This is a Work In Progress** I am putting this WIP as a PR so I can get some comments on it. Specifically I need help with the following:
1.  The helper functions were copied from `packages/ember-template-compiler/lib/plugins/transform-input-on-to-onEvent.js`. Since now two files need these the same functions should be DRY. But I don't know the ember code well enough to know where such utility functions should go.
2.  How and where to test. The tests that touch this are in the glimmer package. There appears to be no tests for `transform-action-syntax.js` in the `ember-template-compiler` so I don't have a good ground to work from. Any help would be appreciated.

Related to issue #14469

### DIFF
--- a/packages/ember-template-compiler/lib/plugins/transform-action-syntax.js
+++ b/packages/ember-template-compiler/lib/plugins/transform-action-syntax.js
@@ -63,5 +63,35 @@ function isAction(node) {
 }
 
 function insertThisAsFirstParam(node, builders) {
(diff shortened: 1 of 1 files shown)


sukima:
Seems #14479 supersedes this. Closing.

PR #14487, opened 2016-10-17
[BUGFIX] Use routeName instead of fullRouteName in QP meta

trentmwillis:
Accidentally broke the Controller lookup for QPs in Engines in the last round of changes. I mistakenly thought we should use `fullRouteName` inside of `_qp` to avoid potential collisions in the BucketCache, but that value is also used to look up the Route's Controller so it shouldn't be fully resolved.

Additionally, the BucketCache seems to be unique per engine-instance, so collisions shouldn't be an issue.

Edit: Added a bunch of tests in the ember-engines repo to verify everything works as expected with this change. See https://github.com/dgeb/ember-engines/pull/227 for more info.


rwjblue:
Awesome, thank you @trentmwillis!

PR #14488, opened 2016-10-18
[BUGFIX beta] Fix cache key for getCompiledBlock

krisselden:
Just use the owner on the template since at this point it already has to have one.


chadhietala:
LGTM

PR #14492, opened 2016-10-19
[BUGFIX beta] Fix unknownProperty key

### LINKED ISSUE #14491: unknownProperty not being called with original key
Original function https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/lib/property_get.js#L50

`Ember.get(obj, 'foo.bar');`

I expected `foo.bar` to get passed to `unknownProperty()` but the actual key which gets passed is just `foo`

https://ember-twiddle.com/51768db4dcb497ea1b4e8d6a6ad7be9b?openFiles=controllers.application.js%2C

krishandley:
Fixes #14491

When calling `get(obj, key)` allow deep keys to be passed
to unknownProperty

e.g.

```
let obj = {
  unknownProperty(key) {
    assert(key, 'foo.bar');
  }
}

get(obj, 'foo.bar');
```

Before only `foo` would have been passed.

### DIFF
--- a/packages/ember-metal/lib/property_get.js
+++ b/packages/ember-metal/lib/property_get.js
@@ -11,6 +11,12 @@ const ALLOWABLE_TYPES = {
   string: true
 };
 
+function canCallUnknownProperty(obj, key) {
+  return 'object' === typeof obj &&
+         !(key in obj) &&
+         'function' === typeof obj.unknownProperty;
+}
+
 // ..........................................................
 // GET AND SET
 //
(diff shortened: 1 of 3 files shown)


krisselden:
This is not a bugfix, it is a feature request.

This change does not behave cohesively with other things in the system.

stefanpenner:
Thank you for the submission, new feature need to go through our RFC process. Let me recommendation that is a first step: https://github.com/emberjs/rfcs#ember-rfcs

PR #5591, opened 2014-09-16
Get firstObject/lastObject notification only when cached

opichals:
Created Em.Array#arrayFirstLastObjectChange abstraction in order to be able to override the Em.Array#firstObject/lastObject notification implementation which by default calls #objectAt(0) and #objectAt(length-1).

These calls are not optimal in case of e.g. sparse RecordArrays which would trigger unnecessary content[index] requests.

Consider an example of using Ember.ListView which only pulls the rendered items out of the content array. In case the content is an ArrayController which dynamically loads items as rendered the Em.Array would actually request the index of 0 and length-1 even in cases where those index content items are not to be rendered at all and therefore trigger unnecessary store layer roundtrips.

PR #14493, opened 2016-10-19
Get firstObject/lastObject notifications only when cached

stefanpenner:
Basically the contents of https://gist.github.com/mmun/877bad33158578a8609b

[fixes #5591]

---

changed:

``` js
cacheFor(this, 'firstObject');
cacheFor(this, 'lastObject');
```

to:

``` js
cacheFor.get(cache, 'firstObject');
cacheFor.get(cache, 'lastObject');
```

this lets us skip needing to re-fetch the cache (we just did above), but not leak the sentinel `UNDEFINED` value (which is why we have the methods in the first place.

---
- Also switched to `objectAt(array, index)` and `this` to `array`
- Also made `length` lazy unless `lastObject` needs accessing

### DIFF
--- a/packages/ember-runtime/lib/mixins/array.js
+++ b/packages/ember-runtime/lib/mixins/array.js
@@ -7,6 +7,8 @@
 // HELPERS
 //
 import { symbol } from 'ember-utils';
+
+import { peekMeta } from 'ember-metal';
 import Ember, { // ES6TODO: Ember.A
   get,
   computed,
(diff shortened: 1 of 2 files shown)


stefanpenner:
@krisselden  / @rwjblue should this ride canary or be a `[BUGFIX]` of some sort?

rwjblue:
I think we can go with `[BUGFIX beta]`, we are still super early in the beta cycle...

stefanpenner:
going to sleep, leave me feedback I can address when i wakeup.

stefanpenner:
Actually this appears to have some issues that need some investigation, specifically `firstObject` and `lastObject` are not always computed properties in our tests. So using cache as a signal isn't quite correct.

stefanpenner:
@krisselden should we go all the way (like the PR currently does) and compare firstObject to the last firstObject we new about before broadcasting the changes(more aligned with our setter strategy), or should we merely broadcast if the cache has been set before (more aligned with our existing CP DK strategy)?

krisselden:
@stefanpenner it is currently already more like the set() strategy, it was just doing it before firstObject and lastObject were ever consumed which is what this was fixing.  I lean toward landing it as is and proposing considering the other changes after

stefanpenner:
@krisselden sounds good, we can always switch to the more CP style by checking affected array offsets in the future.

PR #14504, opened 2016-10-22
Remove space from generated unit-tests

bryanhickerson:
When generating a unit-test for a component, a blank line with a space is generated, which causes linting to fail. Related PR in ember-cli-legacy-blueprints: https://github.com/ember-cli/ember-cli-legacy-blueprints/pull/67.


rwjblue:
Thank you!

PR #14506, opened 2016-10-22
[Doc release] Fix nested contextual component example

### LINKED ISSUE #14503: Possible documentation error for component helper
The API documentation for the `component` helper includes a section on [Nested Usage](http://emberjs.com/api/classes/Ember.Templates.helpers.html#toc_nested-usage) ([source](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/helpers/component.js)) with an example that I suspect is incorrect.

A `person-form` component template defined as follows
```
{{yield (hash
    nameInput=(component "my-input-component" value=model.name placeholder="First Name"))}}
```
is shown being used like this:
```
{{#person-form as |form|}}
  {{component form.nameInput placeholder="Username"}}
{{/person-form}}
```
If my understanding is correct, the use of the `component` keyword in the block is an error, since `form.nameInput` is itself a component, not the name of a component.

The usage above also appears to be at odds with what is shown in the 2.4 beta [release note](http://emberjs.com/blog/2016/01/15/ember-2-3-released.html#toc_contextual-components) and in the [guides](https://guides.emberjs.com/v2.8.0/components/wrapping-content-in-a-component/#toc_sharing-component-data-with-its-wrapped-content).

I'm happy to provide the fix, but would like someone to confirm my interpretation, since I'm only just learning about contextual components for the first time.

pixelhandler:
Change example docs for contextual component…

A `person-form` component template defined as follows

```
{{yield (hash
    nameInput=(component "my-input-component" value=model.name placeholder="First Name"))}}
```

The docs were incorrect…

```
{{#person-form as |form|}}
  {{component form.nameInput placeholder="Username"}}
{{/person-form}}
```

there is no need to use `component` as `form.nameInput` is a contextual component…

```
{{#person-form as |form|}}
  {{form.nameInput placeholder="Username"}}
{{/person-form}}
```
- Fixes #14503

### DIFF
--- a/packages/ember-glimmer/lib/helpers/component.js
+++ b/packages/ember-glimmer/lib/helpers/component.js
@@ -66,19 +66,34 @@ import { assert } from 'ember-metal';
 
   ```handlebars
   {{yield (hash
-      nameInput=(component "my-input-component" value=model.name placeholder="First Name"))}}
+    nameInput=(component "my-input-component" value=model.name placeholder="First Name")
+  )}}
   ```
 
-  The following snippet:
+  When yielding the component via the `hash` helper, the component is invocked directly.
+  See the following snippet:
 
   ```
   {{#person-form as |form|}}
-    {{component form.nameInput placeholder="Username"}}
+    {{form.nameInput placeholder="Username"}}
   {{/person-form}}
   ```
 
-  would output an input whose value is already bound to `model.name` and `placeholder`
+  Which outputs an input whose value is already bound to `model.name` and `placeholder`
   is "Username".
+
+  When yielding the component without the hash helper use the `component` helper.
+  For example, below is a `full-name` component template:
+
+  ```handlebars
+  {{yield (component "my-input-component" value=model.name placeholder="Name")}}
+  ```
+
+  ```
+  {{#full-name as |field|}}
+    {{component field placeholder="Full name"}}
+  {{/full-name}}
+  ```
 
   @method component
   @since 1.11.0


locks (approved):

rwjblue (changes requested):
Both versions are valid (the direct invocation only works when a hash is yielded, but he component helper is needed when the component itself is yielded).

Could you incorporate this info?

pixelhandler:
@rwjblue I added another section like so, does the following look correct?

 When yielding the component without the hash helper use the `component` helper:
  ```handlebars
  {{yield
    nameInput=(component "my-input-component" value=model.name placeholder="First Name")
  }}
  ```
  ```
  {{#person-form as |form|}}
    {{component form.nameInput placeholder="Username"}}
  {{/person-form}}
  ```

packages/ember-glimmer/lib/helpers/component.js
```diff
@@ -66,20 +66,36 @@ import { assert } from 'ember-metal';
...
 
-  would output an input whose value is already bound to `model.name` and `placeholder`
+  Which outputs an input whose value is already bound to `model.name` and `placeholder`
   is "Username".
 
+  When yielding the component without the hash helper use the `component` helper:
+
+  ```handlebars
+  {{yield
+    nameInput=(component "my-input-component" value=model.name placeholder="First Name")
```
rwjblue: I think this is invalid syntax? Should just be {{yield (component ....)}} I think

rwjblue (changes requested):
One small typo, but other than that it looks great

packages/ember-glimmer/lib/helpers/component.js
pixelhandler: Yes, you are correct, I made a twiddle to confirm, https://ember-twiddle.com/84a2ce1300640be9022d87bfa58e580b?openFiles=templates.components.z-comp.hbs%2C

pixelhandler:
@rwjblue Thanks for reviewing. I updated the non-hash example (after making a twiddle to confirm use). I squashed the commits as well.

rwjblue (approved):
Awesome, thank you!

rwjblue:
Thank you!

pixelhandler:
@rwjblue my pleasure, and likewise - thanks for your dedication and service.

PR #14507, opened 2016-10-22
[Doc] Fix example for nested usage of component helper

MarkMT:
Removes erroneous use of `component` keyword, addressing https://github.com/emberjs/ember.js/issues/14503


MarkMT:
Duplicates https://github.com/emberjs/ember.js/pull/14506

PR #14267, opened 2016-09-12
Add ember-utils package.

rwjblue:
- Removes a number of circular reference issues (i.e. container -> ember-metal
  -> container for `getOwner` / `setOwner`).
- Paves the way for removing ember-metal from ember-template-compiler
  package.

Related to build refactoring being worked on by @krisselden.

PR #14509, opened 2016-10-23
[BUGFIX release] Fix overwriting rest positional parameters when passed as named  parameters.

### LINKED ISSUE #14508: LinkComponent cannot be yielded as a contextual component
First of all, I did look at StackOverflow before creating this issue. As a matter of fact, I could not find anything about this, and I am pretty sure that I am not the only one who tried this. 

```handlebars
{{!-- components/a-component/template.hbs --}}
{{yield (hash
  contextual-link=(component "link-to" class="a-component__contextual-link")
}}
{{!-- END --}}

{{!-- somewhere else --}}
{{#a-component as |component|}}
  {{#component.contextual-link "some.route"}}
    Unfortunately, this does not work.
  {{/component.contextual-link}}
{{/a-component}}
{{!-- END --}}
```
When used like that, LinkComponent will complain the it needs at least one parameter. When one bogus parameter is supplied in the template of `a-component`, such as "application", it won't be replaced by the time the `contextual-link` is invoked.

Am I missing something?

Serabe:
If rest positional parameters are passed as named arguments, it might
not be passed to the component when called with a contextual component.

Example:

``` hbs
{{component (component "link-to") params=someParams}}
```

Fix #14508

### DIFF
--- a/packages/ember-glimmer/tests/integration/components/closure-components-test.js
+++ b/packages/ember-glimmer/tests/integration/components/closure-components-test.js
@@ -3,6 +3,7 @@ import { applyMixins, strip } from '../../utils/abstract-test-case';
 import { moduleFor, RenderingTest } from '../../utils/test-case';
 import assign from 'ember-metal/assign';
 import isEmpty from 'ember-metal/is_empty';
+import { A as emberA } from 'ember-runtime/system/native_array';
 
 moduleFor('@htmlbars Components test: closure components', class extends RenderingTest {
   ['@test renders with component helper']() {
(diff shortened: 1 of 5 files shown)


Serabe:
There is a test failing because jshint does not like the line:

```
let Symbol = Symbol || null;
```

And cannot find a way around it :$

I guess this needs to be backported to LTS since it is a regression introduced partially in 2.7 (no transition) and 2.8 (breaks). Checking this next.

Thank you for taking the time to look at this PR!

Serabe:
I checked and this commit can be cherry-picked on top of lts-2-8 without problem. I have also a branch with that done [in my fork](https://github.com/Serabe/ember.js/tree/fix/14508-lts).

Serabe:
Uhm, the test is not failing in CI but in local does :$

packages/ember-metal/tests/utils_test.js
```diff
@@ -10,6 +10,7 @@ QUnit.module('Ember Metal Utils');
 QUnit.test('inspect outputs the toString() representation of Symbols', function() {
   // Symbol is not defined on pre-ES2015 runtimes, so this let's us safely test
   // for it's existence (where a simple `if (Symbol)` would ReferenceError)
+  /* jshint latedef: false */
   let Symbol = Symbol || null;
```
rwjblue: This is pretty wild.  Can we just change it (future PR is fine)?

I think this is a better guard (below):

```js
if (typeof Symbol !== 'undefined') {

}
```

Then we don't have to do any global clobbering variables...

rwjblue (approved):
Changes here look good, can you PR the new tests against master so we keep them?

Also, does this need to be pulled back into lts-2-8 (now that 2.9 is `release`)?  If it does, can you update commit message to be `[BUGFIX lts-2-8]`?

packages/ember-metal/tests/utils_test.js
stefanpenner: the following very much fails:

```js
(function() { let Symbol = Symbol || null;}());
```

with 
```
VM257:1 Uncaught ReferenceError: Symbol is not defined
```

`let Symbol` ensures that before that declaration `Symbol` is in TDZ. basically, it is non-addressable until after the `let`.  As it turns out, the right hand side of `let Symbol = <right hand side>` is technically before the `let`, and since `let` has reserved `Symbol` for that scope, even if the a global `Symbol` exists, this will error.

stefanpenner: opened a comment on a closed babel issue: https://github.com/babel/babel/issues/527#issuecomment-256353308

hzoo: left a comment (can add a tdz option to the plugin)

stefanpenner: @hzoo thanks!

rwjblue: Submitted https://github.com/emberjs/ember.js/pull/14531 to fix this specific test in release/lts-2-8 branches, and they were already fixed in https://github.com/emberjs/ember.js/pull/14267 (see [here](https://github.com/emberjs/ember.js/blob/master/packages/ember-utils/tests/inspect_test.js#L5)).

packages/ember-metal/tests/utils_test.js
```diff
@@ -10,6 +10,7 @@ QUnit.module('Ember Metal Utils');
 QUnit.test('inspect outputs the toString() representation of Symbols', function() {
   // Symbol is not defined on pre-ES2015 runtimes, so this let's us safely test
   // for it's existence (where a simple `if (Symbol)` would ReferenceError)
+  /* jshint latedef: false */
```
rwjblue: Can you remove this and rebase now that https://github.com/emberjs/ember.js/pull/14531 is landed?

rwjblue (changes requested):

packages/ember-metal/tests/utils_test.js
Serabe: On it!

rwjblue (approved):

rwjblue:
Pulled into lts-2-8 branch

PR #14528, opened 2016-10-27
[BUGFIX release] unregister views from viewRegistry (lts-2-8)

### LINKED ISSUE #14526: ember 2.9 2.8 not unregistering views correctly
2.9.0
* https://github.com/emberjs/ember.js/blob/v2.9.0/packages/ember-htmlbars/lib/renderer.js#L263
* https://github.com/emberjs/ember.js/blob/v2.9.0/packages/ember-glimmer/lib/renderer.js#L226

2.8.1
* https://github.com/emberjs/ember.js/blob/v2.8.1/packages/ember-htmlbars/lib/renderer.js#L263
* https://github.com/emberjs/ember.js/blob/v2.8.1/packages/ember-glimmer/lib/renderer.js#L226

2.8.0
* https://github.com/emberjs/ember.js/blob/v2.8.0/packages/ember-htmlbars/lib/renderer.js#L263
* https://github.com/emberjs/ember.js/blob/v2.8.1/packages/ember-glimmer/lib/renderer.js#L226

introduced `v2.8.0-beta.1` https://github.com/emberjs/ember.js/commit/2a233df8c6aa387898299d79dd4b2965b4266015

--

Root cause really that the original code appeared to not be covered by tests, so reasonable refactoring broke it, but without test failure.

--

issue to be sure we don't forget about a test on master: https://github.com/emberjs/ember.js/issues/14530

stefanpenner:
[FIXES #14526]
- [x] ~~needs test~~ (tests added upstream in https://github.com/emberjs/ember.js/pull/14100)

### DIFF
--- a/packages/ember-glimmer/lib/renderer.js
+++ b/packages/ember-glimmer/lib/renderer.js
@@ -223,7 +223,7 @@ class Renderer {
   }
 
   _unregister(view) {
-    delete this._viewRegistry[this.elementId];
+    delete this._viewRegistry[view.elementId];
   }
 }
 
--- a/packages/ember-htmlbars/lib/renderer.js
+++ b/packages/ember-htmlbars/lib/renderer.js
@@ -255,12 +255,12 @@ Renderer.prototype.didDestroyElement = function (view) {
 
 
 Renderer.prototype._register = function Renderer_register(view) {
-  assert('Attempted to register a view with an id already in use: ' + view.elementId, !this._viewRegistry[this.elementId]);
+  assert('Attempted to register a view with an id already in use: ' + view.elementId, !this._viewRegistry[view.elementId]);
   this._viewRegistry[view.elementId] = view;
 };
 
 Renderer.prototype._unregister = function Renderer_unregister(view) {
-  delete this._viewRegistry[this.elementId];
+  delete this._viewRegistry[view.elementId];
 };
 
 export const InertRenderer = {


packages/ember-glimmer/lib/renderer.js
```diff
@@ -218,12 +218,12 @@ class Renderer {
   }
 
   _register(view) {
-    assert('Attempted to register a view with an id already in use: ' + view.elementId, !this._viewRegistry[this.elementId]);
+    assert('Attempted to register a view with an id already in use: ' + view.elementId, !view._viewRegistry[this.elementId]);
```
sly7-7: this should be `this._viewRegistry`, but `view.elementId` right ?

stefanpenner: yup

rwjblue:
This needs to target `lts-2-8` (not `release-2-8`), can you update?

packages/ember-glimmer/lib/renderer.js
```diff
@@ -223,7 +223,7 @@ class Renderer {
   }
 
   _unregister(view) {
-    delete this._viewRegistry[this.elementId];
+    delete this._viewRegistry[view.elementId];
```
rwjblue: Seems fine to change here, but this file isn't used at all on 2.8/2.9 (since the ember-glimmer flag is off).

rwjblue:
Damn.  I'm sorry about this @stefanpenner, I fixed this in https://github.com/emberjs/ember.js/pull/14100 but didn't properly mark it as `[BUGFIX beta` or `[BUGFIX release` 😞 .  Thank you for fixing properly.

rwjblue:
> This needs to target lts-2-8 (not release-2-8), can you update?

I had forgotten that we can do this via normal Github UI now, I updated.

rwjblue:
Rebased on top of #14531 (to fix the JSHint errors this was having).

PR #14537, opened 2016-10-27
[BUGFIX beta] Improve behavior for QPs with undefined values

trentmwillis:
For the case of lazy loading Engines, all QPs get serialized into the URL initially. `undefined` shouldn't get serialized to a string, so that when it gets pulled out the value is not a string.

cc @nathanhammond

### DIFF
--- a/packages/ember-routing/lib/system/router.js
+++ b/packages/ember-routing/lib/system/router.js
@@ -669,6 +669,8 @@ const EmberRouter = EmberObject.extend(Evented, {
       if (qp) {
         delete queryParams[key];
         queryParams[qp.urlKey] = qp.route.serializeQueryParam(value, qp.urlKey, qp.type);
+      } else if (value === undefined) {
+        return; // We don't serialize undefined values
       } else {
         queryParams[key] = this._serializeQueryParam(value, typeOf(value));
       }
--- a/packages/ember/tests/routing/query_params_test.js
+++ b/packages/ember/tests/routing/query_params_test.js
@@ -1231,14 +1231,14 @@ moduleFor('Query Params - main', class extends QueryParamTestCase {
     });
   }
 
-  ['@test Undefined isn\'t deserialized into a string'](assert) {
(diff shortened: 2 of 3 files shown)


nathanhammond:
People willing to make hard decisions: can we take this a step further? I'd like to force all values passed into `queryParams` to be JSON serializable.

```javascript
 _serializeQueryParams(handlerInfos, queryParams) {
  queryParams = JSON.parse(JSON.serialize(queryParams));
  // ...
```

This makes it possible for people to either pass arbitrary objects which implement `toJSON` as values to keys and also sets us up well for a world in which we do structure query params.

It also neatly sidesteps this issue:

```javascript
JSON.parse(JSON.stringify({foo:undefined}));
// => {}
JSON.parse(JSON.stringify({foo:null}));
// => { foo: null }
```

Thoughts?

packages/ember-routing/lib/system/router.js
```diff
@@ -669,7 +669,7 @@ const EmberRouter = EmberObject.extend(Evented, {
       if (qp) {
         delete queryParams[key];
         queryParams[qp.urlKey] = qp.route.serializeQueryParam(value, qp.urlKey, qp.type);
-      } else {
+      } else if (value !== undefined) {
```
rwjblue: Super not-picky but I'd rather leave this as an unconditional else clause. Could you swap this to check if it's undefined and just noop (as an else if before the final else)?

rwjblue (changes requested):
Looks good to me, with one small nit-pick...

rwjblue:
@nathanhammond - Let's keep that separate from this PR, but generally I think it could work (generally speaking). Happy to review a PR (or general sketch of the logic).

trentmwillis:
@rwjblue updated, but getting a linting error. Should I just return from the block?

rwjblue (approved):

rwjblue:
Thanks @trentmwillis!

PR #14540, opened 2016-10-28
[DOC beta] Add common patterns for `{{action (mut...`

### LINKED ISSUE #14525: [DOC] - Additional pattern for {{action (mut ... ) }}
(Moved from https://github.com/emberjs/guides/issues/1713)

Looking at the guides & published API docs, a very handy pattern for `{{mut}}` do not seem to be covered.

#### setting the value of a property to some **literal** value
```hbs
{{#if myToggle}}
   {{!-- Clicking this will set myToggle to false --}}   
   <button onClick={{action (mut myToggle) false}}>Hide</button>
{{else}}
   {{!-- Clicking this will set myToggle to true --}}
   <button onClick={{action (mut myToggle) true}}>Show</button>
{{/if}}
```

#### Variant - accessing a particular property off of the event
```hbs
<input type='range'
   onInput={{action (mut size) value="target.value"}} />
```
While the `value` option is part of the action helper, the way it works in conjunction with `mut` may make it worth mentioning explicitly as a use case 

For clarity, here examples of both: https://ember-twiddle.com/fc4179799bcf33c58f67d9cdd84111d0?openFiles=templates.application.hbs%2C

Serabe:
Add two common patters for the `mut` helper used as parameter of
`action`.

Fixes #14525

### DIFF
--- a/packages/ember-glimmer/lib/helpers/mut.js
+++ b/packages/ember-glimmer/lib/helpers/mut.js
@@ -52,6 +52,25 @@ import { INVOKE } from './action';
 
   The `mut` helper changes the `totalClicks` value to what was provided as the action argument.
 
+  The `mut` helper, when used with `action`, will return a function that
+  sets the value passed to `mut` to its first argument. This works like any other
+  closure action and interacts with the other features `action` provides.
+  As an example, we can create a button that increments a value passing the value
+  directly to the `action`:
+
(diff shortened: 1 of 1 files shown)


mike-north:
This is great! I'm sure a lot of people will appreciate this

rwjblue:
Thank you!

PR #14543, opened 2016-10-28
[DOC] Add an example for block syntax to component helper docs

kaeufl:
Although the docs for the component helper shortly mention block syntax in one sentence ("The provided block will be applied as the template for the component.") this can easily be overlooked and no example is provided. While this may seem trivial if you know what you're looking for, it took me quite some time to figure out that it is possible to use the component helper in combination with block syntax.


locks:
Thank you for your contribution @kaeufl! I agree that this is a useful addition, and hopefully helps people make the connection to component's block form :)
I have [sent a PR](https://github.com/kaeufl/ember.js/pull/1/files) to your repository with some additional suggestions, and a little bit of house cleaning that we've been doing across API documentation. Ping us for any question!

locks:
@kaeufl can you squash the commits into a single one as well?

locks:
Thank you!

@homu r+

PR #14546, opened 2016-10-28
[BUGFIX beta] Update route-recognizer to v0.2.8.

nathanhammond:
Changes in this version account for edge cases which result in `undefined` behavior in route-recognizer:
- https://github.com/tildeio/route-recognizer/pull/116
- https://github.com/tildeio/route-recognizer/pull/118

Should be backward compatible but making sure that we get most of a canary cycle with eyes on this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "mocha": "^2.4.5",
     "qunit-extras": "^1.5.0",
     "qunitjs": "^1.22.0",
-    "route-recognizer": "^0.2.7",
+    "route-recognizer": "^0.2.8",
     "rsvp": "~3.2.1",
     "serve-static": "^1.10.0",
     "simple-dom": "^0.3.0",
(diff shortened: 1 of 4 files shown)


nathanhammond:
There is one actual Ember bug: we appear to recalculate `link-to`'s `href` during teardown if a bound param into the `link-to` component is torn down prior to the `link-to` being torn down. The upstream changes throw an error if you attempt to generate a URL without all of the specified params, so in combination it now throws an error.

I haven't finished tracing that issue. As this is actually a separate and different issue it will come separate from this PR. Depending upon how complicated and invasive that fix is will determine if I tag it as a `[BUGFIX release|beta]`.

nathanhammond:
After building the test `the {{link-to}} helper does not throw an error if its route has exited` as an application I have discovered that the test itself is invalid. I've modified the test to do what it was originally intending to accomplish (confirm that we don't tear down `controller`s) and made the test pass.

I could not produce the failure of my original hypothesis (which is why I was delayed in addressing this).

rwjblue:
Thanks @nathanhammond!

PR #14550, opened 2016-10-30
[BUGFIX lts-2-8] Allow canceling items queued by `run.schedule`.

rwjblue:
All of the scheduling related `Ember.run.*` methods return the backburner timer cancelation token with the exception of `Ember.run.schedule`. (this includes `run.next`, `run.later`, `run.debounce`, `run.throttle`, and `run.scheduleOnce`).

This seems like a fairly glaring mistake (and makes it impossible to avoid `if (this.isDestroying) { return; }` guards when using things like `run.schedule('afterRender'`), so I have marked it as `BUGFIX lts-2-8`. If others disagree with the severity here, I can change to a simple `[BUGFIX beta]`.

### DIFF
--- a/packages/ember-metal/lib/run_loop.js
+++ b/packages/ember-metal/lib/run_loop.js
@@ -260,7 +260,7 @@ run.end = function() {
     will be resolved on the target object at the time the scheduled item is
     invoked allowing you to change the target function.
   @param {Object} [arguments*] Optional arguments to be passed to the queued method.
-  @return {void}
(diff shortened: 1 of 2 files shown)


stefanpenner:
apparently `once` https://github.com/ebryn/backburner.js/issues/161 may not be cancellable, but `schedules` should be. But this PR doesn't talk about `once` or `scheduleOnce`

stefanpenner:
if we are encouraging people to cancel, we may want improve how we cancel https://github.com/ebryn/backburner.js/issues/177

PR #14554, opened 2016-10-30
[BUGFIX beta] Improve error for improper invocations of link-to.

nathanhammond:
Unfortunately, to get decent error messages, we need to do something like this. In some future state we should be able to use a "feature flag" which allows us to strip this without needing to call it twice.

``` javascript
if (isProdBuild()) {
  // Do the useful debug thing, probably including try/catch.
} else {
  // Do the performant thing.
}
```

### DIFF
--- a/packages/ember-glimmer/lib/components/link-to.js
+++ b/packages/ember-glimmer/lib/components/link-to.js
@@ -313,7 +313,8 @@ import {
   deprecate,
   get,
   computed,
-  flaggedInstrument
+  flaggedInstrument,
+  runInDebug
 } from 'ember-metal';
 import {
   deprecatingAlias,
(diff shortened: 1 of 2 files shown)


packages/ember-glimmer/lib/components/link-to.js
```diff
@@ -711,6 +712,26 @@ const LinkComponent = EmberComponent.extend({
...
     let routing = get(this, '_routing');
     let queryParams = get(this, 'queryParams.values');
+
+    runInDebug(() => {
+      /*
+       * Unfortunately, to get decent error messages, we need to do this.
+       * In some future state we should be able to use a "feature flag"
+       * which allows us to strip this without needing to call it twice.
+       *
+       * if (isProdBuild()) {
```
rwjblue: I believe either the truthy/falsey blocks should be flip/flopped or this conditional should be changed to `isDebugBuild()` (since we want to do the most performant thing in prod mode)...

rwjblue (changes requested):

rwjblue:
This looks great, I have visions of threading through the callsite information so we can show template module name and line number info as well, but this is a great improvement already.

packages/ember-glimmer/lib/components/link-to.js
nathanhammond: Definitely wrote the wrong thing. Adjusting.

rwjblue (approved):
LGTM, definitely a better error message than the default one...

nathanhammond:
@rwjblue Note that the error, by including `e.message`, also includes the name of the segment that you didn't pass in via https://github.com/tildeio/route-recognizer/pull/116. 😄

nathanhammond:
K, we're good here.

rwjblue:
Awesome, thank you @nathanhammond!

Issue #14561, opened 2016-10-31
Helper is resolved instead of property


ro0gr:
When I use a property with the name equal to a helper name then helper wins and gets rendered.
It happens even if the property name is camelCased and helper name is dash-erized(surprise for me). 

Reproduction: https://ember-twiddle.com/3ad62a907c9de66dd0678e99fab6ae49?openFiles=templates.application.hbs%2C

This reproduction is based on ember@2.8. If you switch app to ember@1.13 it works as expected(property wins).

I couldn't find related issue in this issue tracker but this pull request  https://github.com/tildeio/glimmer/pull/335 looks related.

rwjblue:
https://github.com/tildeio/glimmer/pull/335 is a bit different than this twiddle, it is talking about block params (not properties on the template context).  

I believe that helpers winning has always been true actually (see your demo [twiddle with dasherized name in 1.13](https://ember-twiddle.com/9e146f4bffa8b3e36b6a28f6ecbc1300?openFiles=twiddle.json%2C)), and was an intentional design decision.

The fact that `{{appName}}` looks up a helper named `app-name` is a separate issue, and I suspect the ember-cli resolver's `normalize` function is doing this...

ro0gr:
Thank you for clarifications and explanatory example @rwjblue.

At this point it's still not very clear to me why does property try to be resolved via resolver while it's just a plain javascript value in the parent scope? 

I mean: 
> ember-cli resolver's normalize function is doing this

this line makes me a little bit unsure... Is this also an intentional decision?

I understand my question is out of current issue scope. Anyway I would like to understand if I need to fill another issue and which one :smirk_cat:

ro0gr:
Looks like I finally found something relates https://github.com/emberjs/ember.js/issues/13404 but not about dasherizing of camelCased props.

lukemelia:
I propose that we reopen this. The following twiddle shows how camelCase expressions resolve to dasherized helper names. This is surprising when you have a local camelCase property you are trying to render. 

https://ember-twiddle.com/54ad94b74246529c57286369273f410c?openFiles=templates.application.hbs%2Ctemplates.components.component-that-updates.hbs

You can switch this Twiddle to Ember 1.13.x to see that the property name was resolved as expected under 1.13.

lukemelia:
Somewhat related: https://github.com/ember-cli/ember-resolver/issues/108

lukemelia:
[...]

PR #14563, opened 2016-10-31
Add failing test for resolving props over helpers

ro0gr:
Provided an integration test for #14561


ro0gr:
Closing according to @rwjblue explanation: precedence of helpers over local properties is an intentional design decision.

PR #14564, opened 2016-10-31
Sets _qpDelegate before running code which could trigger _qpChanged

rauhryan:
In scenarios where you are sharing query params across routes, let's say in a service. 

Transitioning between routes, if you've made a state change that would trigger _qpChanged to fire, the route will throw an exception 

```
TypeError: delegate is not a function
```

<details>
```
ember.debug.js:30007 TypeError: delegate is not a function
    at Class._qpChanged (https://cdnjs.cloudflare.com/ajax/libs/ember.js/2.1.2/ember.debug.js:23829:7)
    at Object.apply (https://cdnjs.cloudflare.com/ajax/libs/ember.js/2.1.2/ember.debug.js:21790:18)
    at Object.sendEvent (https://cdnjs.cloudflare.com/ajax/libs/ember.js/2.1.2/ember.debug.js:15616:28)
    at notifyObservers (https://cdnjs.cloudflare.com/ajax/libs/ember.js/2.1.2/ember.debug.js:19247:25)
    at propertyDidChange (https://cdnjs.cloudflare.com/ajax/libs/ember.js/2.1.2/ember.debug.js:19057:5)
    at Object.notify (https://cdnjs.cloudflare.com/ajax/libs/ember.js/2.1.2/ember.debug.js:13310:9)
    at chainsDidChange (https://cdnjs.cloudflare.com/ajax/libs/ember.js/2.1.2/ember.debug.js:19163:9)
    at Object.propertyDidChange (https://cdnjs.cloudflare.com/ajax/libs/ember.js/2.1.2/ember.debug.js:19056:5)
    at set (https://cdnjs.cloudflare.com/ajax/libs/ember.js/2.1.2/ember.debug.js:19495:38)
    at setPath (https://cdnjs.cloudflare.com/ajax/libs/ember.js/2.1.2/ember.debug.js:19534:12)
```
</details>

I've created a scaled down [twiddle](https://ember-twiddle.com/6eba729783b5b955ac13b9ce91b5b998) that you can reproduce the error in. 

An example of usage in the "real world", in HuBoard a side panel with filters in it. Those filters are synced to query params.  Our `index` route shows a task board, but if you transition over to the `milestones` route, we show a similar drag and drop board but we preserve your current filtered down subset of issues between the two routes. 

Kind of like in Trello does, if you add filters on your board and then your transition over to the "calendar" view your filters are still active.

### DIFF
--- a/packages/ember/tests/routing/query_params_test/shared_state_test.js
+++ b/packages/ember/tests/routing/query_params_test/shared_state_test.js
@@ -1,0 +1,76 @@
+import {
+  Controller,
+  Service
+} from 'ember-runtime';
+import Ember from 'ember';
+import { run } from 'ember-metal';
+import { jQuery } from 'ember-views';
+import { QueryParamTestCase, moduleFor } from 'internal-test-helpers';
+
+moduleFor('Query Params - shared service state', class extends QueryParamTestCase {
+
+  boot() {
+    this.setupApplication();
+    return this.visitApplication();
+  }
+
+  setupApplication() {
+    this.router.map(function() {
+      this.route('home', { path: '/' });
+      this.route('dashboard');
+    });
+
+    this.application.register('service:filters', Service.extend({
+      shared: true
+    }));
+
+    this.registerController('home', Controller.extend({
+      filters: Ember.inject.service()
+    }));
+
+    this.registerController('dashboard', Controller.extend({
+      filters: Ember.inject.service(),
+      queryParams: [
+        { 'filters.shared': 'shared' }
+      ]
+    }));
+
+    this.registerTemplate('application', `{{link-to 'Home' 'home' }} <div> {{outlet}} </div>`);
+    this.registerTemplate('home', `{{link-to 'Dashboard' 'dashboard' }}{{input type="checkbox" id='filters-checkbox' checked=(mut filters.shared) }}`);
+    this.registerTemplate('dashboard', `{{link-to 'Home' 'home' }}`);
+  }
+  visitApplication() {
+    return this.visit('/');
+  }
+
+  ['@test can modify shared state before transition'](assert) {
+    assert.expect(1);
+
+    return this.boot().then(() => {
+      this.$input = jQuery('#filters-checkbox');
+
+      // click the checkbox once to set filters.shared to false
+      run(this.$input, 'click');
+
+      return this.visit('/dashboard').then(() => {
+        assert.ok(true, 'expecting navigating to dashboard to succeed');
+      });
+    });
+  }
+
+  ['@test can modify shared state back to the default value before transition'](assert) {
+    assert.expect(1);
+
+    return this.boot().then(() => {
+      this.$input = jQuery('#filters-checkbox');
+
+      // click the checkbox twice to set filters.shared to false and back to true
+      run(this.$input, 'click');
+      run(this.$input, 'click');
+
+      return this.visit('/dashboard').then(() => {
+        assert.ok(true, 'expecting navigating to dashboard to succeed');
+      });
+    });
+  }
+});
--- a/packages/ember-routing/lib/system/route.js
+++ b/packages/ember-routing/lib/system/route.js
@@ -1316,6 +1316,9 @@ let Route = EmberObject.extend(ActionHandler, Evented, {
     let queryParams = get(this, '_qp');
 
     let states = queryParams.states;
+
+    controller._qpDelegate = states.allowOverrides;
+
     if (transition) {
       // Update the model dep values used to calculate cache keys.
       stashParamNames(this.router, transition.state.handlerInfos);
@@ -1336,8 +1339,6 @@ let Route = EmberObject.extend(ActionHandler, Evented, {
         }
       });
     }
-
-    controller._qpDelegate = states.allowOverrides;
 
     if (transition) {
       let qpValues = getQueryParamsFor(this, transition.state);


rauhryan:
For what it's worth, I have a dirty workaround for this that reopens the Route class to set the _qpDelegate and then runs `_super`

```
import Ember from 'ember';

function addQueryParamsObservers(controller, propNames) {
  propNames.forEach(function(prop) {
    controller.addObserver(prop + '.[]', controller, controller._qpChanged);
  });
}

let get = Ember.get;

export function initialize(container, application){
  Ember.Route.reopen({
    setup(context, transition) {
      var controller;

      var controllerName = this.controllerName || this.routeName;
      var definedController = this.controllerFor(controllerName, true);

      if (!definedController) {
        controller =  this.generateController(controllerName, context);
      } else {
        controller = definedController;
      }


      // Assign the route's controller so that it can more easily be
      // referenced in action handlers. Side effects. Side effects everywhere.
      if (!this.controller) {
        var propNames = get(this, '_qp.propertyNames');
        addQueryParamsObservers(controller, propNames);
        this.controller = controller;
      }

      let queryParams = get(this, '_qp');

      let states = queryParams.states;

      controller._qpDelegate = states.allowOverrides;

      return this._super(...arguments);
    }
  });
}

export default {
  name: 'punch-router',
  initialize: initialize
};
```

See for the full file here: https://github.com/huboard/huboard-web/blob/4c355ed530d1b8fd298037110bb922ff893ec83d/ember-app/app/initializers/punch-router.js

packages/ember/tests/routing/query_params_test/shared_state_test.js
```diff
@@ -0,0 +1,88 @@
...
+    });
+
+    this.application.register('service:filters', Service.extend({
+      shared: true
+    }));
+
+    this.registerController('home', Controller.extend({
+      filters: Ember.inject.service(),
+      queryParams: [
+        { 'filters.shared': 'shared' }
```
trentmwillis: Based on messing with the Twiddle, this shouldn't be needed. The requirements to reproduce seem to just be that the dependent property of a QP be modified prior to visiting the route of that QP.

So, I think we can simplify this reproduction by not having these query params

packages/ember/tests/routing/query_params_test/shared_state_test.js
```diff
@@ -0,0 +1,88 @@
...
+
+    this.registerTemplate('application', `{{link-to 'Home' 'home' }} <div> {{outlet}} </div>`);
+    this.registerTemplate('home', `{{link-to 'Dashboard' 'dashboard' }}{{input type="checkbox" id='filters-checkbox' checked=(mut filters.shared) }}`);
+    this.registerTemplate('dashboard', `{{link-to 'Home' 'home' }}`);
+  }
+  visitApplication() {
+    return this.visit('/');
+  }
+
+  ['@test should not fail if shared state is modified from default before transition']() {
```
trentmwillis: You can remove the `should not fail if` part of these test descriptions, it's implied.

packages/ember/tests/routing/query_params_test/shared_state_test.js
```diff
@@ -0,0 +1,88 @@
...
+    let assert = this.assert;
```
trentmwillis: You can/should just use the assert logic passed into the function (that still exists here even though we deviate from QUnit a bit).

packages/ember/tests/routing/query_params_test/shared_state_test.js
```diff
@@ -0,0 +1,88 @@
...
+    let done = assert.async();
```
trentmwillis: No need to use `done` when returning a Promise. QUnit will wait from the Promise to finish automatically.

packages/ember/tests/routing/query_params_test/shared_state_test.js
```diff
@@ -0,0 +1,88 @@
+import {
+  Controller,
+  Service
+} from 'ember-runtime';
+import Ember from 'ember';
+import { run } from 'ember-metal';
+import { jQuery } from 'ember-views';
+import { QueryParamTestCase, moduleFor } from 'internal-test-helpers';
+
+class SharedStateQPTestCase extends QueryParamTestCase {
```
trentmwillis: You can just define one class, instead of defining this one here only to extend it below.

trentmwillis (changes requested):
This seems reasonable to me. In messing with the Twiddle, I think the reproduction can be reduced further which is called out in one of my comments.

rwjblue:
@rauhryan - Ping! I'd love to get this landed into the beta branch (for an Ember 2.10 release), but we'll need the various cleanups mentioned by @trentmwillis before we can land it...

rauhryan:
I'm hoping to get this knocked out tomorrow or Wednesday. 

What's the deadline for the 2.10 beta? 

 - Ryan

rwjblue:
> What's the deadline for the 2.10 beta?

There will be another couple beta's. We will release one today/tomorrow then another one early next week.

rauhryan:
ping @rwjblue & @trentmwillis 

thanks for the feedback, I've just pushed changes to address your requested changes

packages/ember/tests/routing/query_params_test/shared_state_test.js
```diff
@@ -61,15 +54,11 @@ moduleFor('Query Params - shared service state', class extends SharedStateQPTest
...
-        done();
       });
     });
   }
 
-  ['@test should not fail if shared state is equal to the default value before transition']() {
-    let assert = this.assert;
-    let done = assert.async();
-
+  ['@test can modify shared state back to it the default value before transition'](assert) {
```
rwjblue: small typo here, I think it should be:

> can modify shared state back to the default value before transition

rwjblue (changes requested):
Looks great, thank you for working on this!

I left one small typo related comment that needs fixing, and needs to be sqaushed down into a single commit then we can land...

rauhryan:
Ok,

  - typo fixed
  - commit squashed

rwjblue (approved):

PR #14570, opened 2016-11-01
Update router to npm

krisselden:


ember-cli-build.js
```diff
@@ -103,23 +103,9 @@ function routeRecognizer() {
 }
 
 function router() {
-  // TODO upstream this to router.js and publish on npm
-  return new Rollup('bower_components/router.js/lib', {
+  return new Rollup(path.resolve(path.dirname(require.resolve('router_js')), '../lib'), {
```
nathanhammond: Can we base this off information in `router_js`' `package.json`?

Something like:
`path.resolve(path.dirName(require.resolve('router_js')), fs.readFileSync(require.resolve('router_js/package.json')).main)`?

nathanhammond (commented):
This currently hardcodes a bit of information about how router.js is set up into Ember. Can we make that not the case?

ember-cli-build.js
krisselden: also, that is what `require.resolve('router_js')` returns

nathanhammond: Confirm we don't want the `main`, but maybe `'jsnext:main'` or [`directories.lib`](https://docs.npmjs.com/files/package.json#directorieslib).

krisselden: Yes, it can just be `require('router_js/package')['jsnext:main']` at that point.  I thought though they were pushing 'module' now instead of 'jsnext:main'?

Issue #14557, opened 2016-10-31
TypeError: Invalid value used as weak map key, in get() with canary


backspace:
Hey, I’ve been trying out the Ember beta and canary versions with [`travis-web`](https://github.com/travis-ci/travis-web) and getting a WeakMap error. When I run with `canary` in `bower.js` in Chrome 56, I get this stacktrace:

```
WeakMap_setMeta (ember.debug.js:19145)
meta (ember.debug.js:19246)
watchKey (ember.debug.js:22349)
RootPropertyReference (ember.debug.js:14278)
get (ember.debug.js:14184)
referenceFromParts (ember.debug.js:44196)
evaluate (ember.debug.js:46140)
evaluate (ember.debug.js:45613)
evaluateArgs (ember.debug.js:53312)
evaluate (ember.debug.js:48694)
execute (ember.debug.js:53279)
RootState.render (ember.debug.js:11301)
runInTransaction (ember.debug.js:22272)
_renderRoots (ember.debug.js:11566)
_renderRootsTransaction (ember.debug.js:11604)
_renderRoot (ember.debug.js:11525)
_appendDefinition (ember.debug.js:11438)
appendOutletView (ember.debug.js:11421)
invoke (ember.debug.js:337)
flush (ember.debug.js:407)
flush (ember.debug.js:531)
end (ember.debug.js:601)
run (ember.debug.js:724)
fire (jquery.js:3187)
fireWith (jquery.js:3317)
ready (jquery.js:3536)
completed (jquery.js:3552)
```

Similar errors happen with Safari, Firefox, and Chrome 53.

This can be reproduced with the repository running locally with `beta` or `canary` as the Ember version in `bower.json`. When I enable pausing on exceptions, I see that the `get` is on a property key of `config`:

![image](https://cloud.githubusercontent.com/assets/43280/19846566/d87e71ec-9f40-11e6-8d65-6e644136a248.png)

I’ve been working on permanent [beta](https://ember-beta.travis-ci.org/) and [canary](https://ember-canary.travis-ci.org/) production environment deployments of Ember, which are failing differently with this error:

```
ember.prod.js:20058 Uncaught TypeError: Cannot read property 'config' of undefined
```

The presence of `config` there makes me wonder if they’re somehow related though, but the stack trace is pretty different:

<details>
<summary>Production canary stacktrace</summary>
```
r (ember.prod.js:20058)
t.compute (ember.prod.js:13338)
t.value (ember.prod.js:13213)
[...]

PR #14571, opened 2016-11-01
[BUGFIX beta] Fix non object paths

krisselden:


rwjblue (approved):

krisselden:
I'm not sure how to test these.   There are issues with RootReference and NestedReference, there is also another old node/Chrome issue.  The fix in chains fixes that, I'm going to just push a fix to watchKey and unwatchKey for now.

krisselden:
This should fix #14557

PR #14574, opened 2016-11-03
Use absolute path in mocha acceptance test blueprints

Dhaulagiri:
This PR pulls in the [same change](https://github.com/ember-cli/ember-cli-mocha/pull/109) I made to ember-cli-mocha prior to the blueprints being extracted to this repo:

> Avoids issues with nested acceptance tests having invalid relative paths in them.
> 
> For instance, `ember g acceptance-test foo/bar` currently results in:
> 
> `import startApp from '../helpers/start-app'`
> 
> when it should be 
> 
> `import startApp from '../../helpers/start-app'`.
> 
> After this change it will be the below which should work regardless of how deeply nested (or not) the acceptance test is:
> 
> `import startApp from 'app-name/tests/helpers/start-app'`


rwjblue:
👍

Dhaulagiri:
The test failures don't seem related to my changes here 🌵

locks (approved):

Issue #14572, opened 2016-11-02
Bug: Ember.get with empty string as path no longer returns target object


jasonmit:
Previously, `Ember.get(obj, '')` would return `obj`.

This behavior was removed in #14345 and caused some breakage in ember-cp-validations which relied on this.  https://github.com/offirgolan/ember-cp-validations/pull/380

My opinion is this was likely a bug that people relied on that could instead be marked for deprecation.

Dhaulagiri:
@jasonmit see https://github.com/emberjs/ember.js/issues/14386 for discussion of a similar issue related to the change from #14345

jasonmit:
Thanks, I'll close this one

rwjblue:
I still think we can make an assertion that the key passed to Ember.get is not empty string. That would at least have made the other logic issue (in Ember-count-validations) easier to track down.

technomage:
Note that this is a breaking change, and should have been a deprecation (in my opinion), not cause working tests to fail when migrating from 2.7 to 2.10 and requiring patch to add ons relying on the prior behavior.  In my case an add-on ember-cli-select-picker.

bstro:
This is breaking a bunch of our tests – is there any particular trick to finding the call site where Ember.get is being called with an empty string? This error is occurring when an acceptance test tries to render a handlebars template and my stack-trace has no helpful information. Not sure how else I can debug this.

PR #14575, opened 2016-11-04
[FEATURE beta] Assert key in `Ember.get` is not empty

Serabe:
This adds an assertion to let the user know when they are using an empty
string as key for `Ember.get`.

Addresses [this comment](https://github.com/emberjs/ember.js/issues/14572#issuecomment-258202256) in #14572


rwjblue (approved):
Thank you for following up on that comment!

PR #14577, opened 2016-11-04
[FEATURE beta] Assert expand properties are balanced and not nested

### LINKED ISSUE #13693: undefined behavior for nested brace expansion
I've come across some code that _could_ use nested brace expansion. I tried it out manually first and got this:

```
Ember.expandProperties('first.{second1,second2.{third1,third2}}', echo);
first.second1third1
first.second1third2
first.second2.third1
first.second2.third2
```

I've read [this](https://github.com/emberjs/ember.js/pull/13231#issuecomment-204632269) which hints that it was discussed or decided to not support. If it is unsupported, perhaps it should error, like when you put a space in. Or maybe it just needs a regex update to fully support it? It could make computeds further collapsable.

Serabe:
In #13693, it is asked to error for an unsupported feature. Though we
don't say that this is not supported, it would be nice if we could let
the user know that what he is trying to do will not work instead of
silently fail.

Fixes #13693

### DIFF
--- a/packages/ember-metal/lib/expand_properties.js
+++ b/packages/ember-metal/lib/expand_properties.js
@@ -41,6 +41,27 @@ export default function expandProperties(pattern, callback) {
     'Brace expanded properties cannot contain spaces, e.g. "user.{firstName, lastName}" should be "user.{firstName,lastName}"',
     pattern.indexOf(' ') === -1
   );
+  assert(
+    `Brace expanded properties have to be balanced and cannot be nested, pattern: ${pattern}`,
+    ((str) => {
+      let inBrace = 0;
+      let char;
+      for (let i = 0; i < str.length; i++) {
+        char = str.charAt(i);
+
+        if (char === '{') {
+          inBrace++;
+        } else if (char === '}') {
+          inBrace--;
+        }
+
+        if (inBrace > 1 || inBrace < 0) {
+          return false;
+        }
+      }
+
+      return true;
+    })(pattern));
 
   let parts = pattern.split(SPLIT_REGEX);
   let properties = [parts];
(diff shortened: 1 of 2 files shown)


packages/ember-metal/lib/expand_properties.js
```diff
@@ -8,6 +8,27 @@ import { assert } from './debug';
 const SPLIT_REGEX = /\{|\}/;
 const END_WITH_EACH_REGEX = /\.@each$/;
 
+// Checks if the braces inside a properties path are not nested and balanced
+function checkPropertiesBalance(str) {
```
rwjblue: Can you move this down to the assertion so that it will also be stripped from production builds?

```js
assert(
  'Brace expanded properties have to be balanced and cannot be nested, e.g. "user.{name.{first,last},age}" and "user.{firstName},lastName}" are not allowed',
  ((str) => {
    let inBrace = 0;
    let char;
    for (let i = 0; i < str.length; i++) {
      char = str.charAt(i);

      if (char === '{') {
        inBrace++;
      } else if (char === '}') {
        inBrace--;
      }

      if (inBrace > 1 || inBrace < 0) {
        return false;
      }
    }

    return true;
  })(pattern)
);
```

packages/ember-metal/lib/expand_properties.js
```diff
@@ -41,6 +62,9 @@ export default function expandProperties(pattern, callback) {
     'Brace expanded properties cannot contain spaces, e.g. "user.{firstName, lastName}" should be "user.{firstName,lastName}"',
     pattern.indexOf(' ') === -1
   );
+  assert(
+    'Brace expanded properties have to be balanced and cannot be nested, e.g. "user.{name.{first,last},age}" and "user.{firstName},lastName}" are not allowed',
```
rwjblue: Can you tweak the message to show the actual brace expansion that the user used (instead of a static value)?  Possibly do the same for the assertion just above?

rwjblue (changes requested):

Serabe:
Done. Thank you for reviewing!

rwjblue (approved):

stefanpenner:
👍

Issue #14168, opened 2016-08-31
Deprecate usage of `targetObject`.


rwjblue:
We currently support having a `targetObject` property on components, but as of 2.9 we no longer use this property on components internally.

We should deprecate having a `targetObject` [here](https://github.com/emberjs/ember.js/blob/8112e7112ddabf871044fdce1a893d78e6852d3f/packages/ember-runtime/lib/mixins/target_action_support.js#L140-L141).

Cryrivers:
Thanks for the clarification, although I wasn't able to explain clearly on Slack. :)

Basically we used targetObject in a component to get the controller that it belongs to. (Say you need to call a function in the controller, from your components. I wouldn't suggest to do so since Ember 2.3 introduced closure actions.)

However, as of 2.9, targetObject no longer exists. I tried with this._targetObject but it has a different behavior compared with the original targetObject. In 2.9, _targetObject in a component points to its parent component, if any, instead of the controller it belongs to.

So I believe this deprecation is needed and hopefully can help you if you are running into targetObject problem in 2.9

rwjblue:
This issue represents deprecating having a `targetObject` specified (either via template invocation or in your component definition). The `targetObject` property on `Ember.Component` was intended as a "write only API" (as in you are supposed to define your own if you'd like, but not try to read from the default one). We certainly never intended for `this.get('targetObject')` to be used. 

> Basically we used targetObject in a component to get the controller that it belongs to.

It sounds to me like you were using `this.get('targetObject')`, is that right?

> I wouldn't suggest to do so since Ember 2.3 introduced closure actions.

FWIW, Ember 1.13 introduced closure actions.

@Cryrivers - Can you give me an example JSBin or ember-twiddle to show the behavior you are discussing so we can better understand if we need to deprecate *reading* from `targetObject` as well as specifying it?

Cryrivers:
@rwjblue https://ember-twiddle.com/e2f017a62fb16c2163d267d49e2d0ed9?openFiles=components.component-b.js%2C

You can open the console and switch between Ember `2.7.0` and `canary`.

In `2.7.0`, targetObject points to `ApplicationController` whereas it points to its parent component in `canary`.

rwjblue:
@Cryrivers - Thanks for working through this with me.  You have definitely identified a bug (action bubbling from within a component's block does not work correctly on canary), but I'm still not 100% sure if we should add a `targetObject` property back that issues a deprecation.

[...]

PR #14590, opened 2016-11-08
[BUGFIX beta] Adds a deprecation message when using targetObject

sduquej:
For https://github.com/emberjs/ember.js/issues/14168

@rwjblue is this enough or should we also assert when setting `targetObject`? If so, where might I add that check?

### DIFF
--- a/packages/ember-runtime/lib/mixins/target_action_support.js
+++ b/packages/ember-runtime/lib/mixins/target_action_support.js
@@ -7,9 +7,10 @@ import { context } from 'ember-environment';
 import {
   get,
   Mixin,
-  computed
+  computed,
+  descriptor
 } from 'ember-metal';
-import { assert } from 'ember-debug';
+import { assert, deprecate } from 'ember-debug';
 /**
 `Ember.TargetActionSupport` is a mixin that can be included in a class
 to add a `triggerAction` method with semantics similar to the Handlebars
@@ -24,6 +25,22 @@ doing more complex event handling in Components.
 */
 export default Mixin.create({
   target: null,
+  targetObject: descriptor({
+    configurable: true,
+    enumerable: false,
+    get() {
+      let message = `${this} Usage of \`targetObject\` is deprecated. Please use \`target\` instead.`;
+      let options = { id: 'ember-runtime.using-targetObject', until: '3.5.0' };
+      deprecate(message, false, options);
+      return this._targetObject;
+    },
+    set(value) {
+      let message = `${this} Usage of \`targetObject\` is deprecated. Please use \`target\` instead.`;
+      let options = { id: 'ember-runtime.using-targetObject', until: '3.5.0' };
+      deprecate(message, false, options);
+      this._targetObject = value;
+    }
+  }),
   action: null,
   actionContext: null,
 
@@ -121,16 +138,7 @@ export default Mixin.create({
 });
 
 function getTarget(instance) {
-  // TODO: Deprecate specifying `targetObject`
-  let target = get(instance, 'targetObject');
-
-  // if a `targetObject` CP was provided, use it
-  if (target) { return target; }
-
-  // if _targetObject use it
-  if (instance._targetObject) { return instance._targetObject; }
-
-  target = get(instance, 'target');
+  let target = get(instance, 'target');
   if (target) {
     if (typeof target === 'string') {
       let value = get(instance, target);
@@ -143,6 +151,12 @@ function getTarget(instance) {
       return target;
     }
   }
+
+  // if a `targetObject` CP was provided, use it
+  if (target) { return target; }
+
+  // if _targetObject use it
+  if (instance._targetObject) { return instance._targetObject; }
 
   return null;
 }
--- a/packages/ember-runtime/tests/mixins/target_action_support_test.js
+++ b/packages/ember-runtime/tests/mixins/target_action_support_test.js
@@ -89,6 +89,26 @@ QUnit.test('it should use an actionContext object specified as a property on the
   ok(true === obj.triggerAction(), 'a valid target and action were specified');
 });
 
+
+QUnit.test('it should raise a deprecation warning when targetObject is specified and used', function() {
+  expect(4);
+  let obj;
+  expectDeprecation(() => {
+    obj = EmberObject.extend(TargetActionSupport).create({
+      action: 'anEvent',
+      actionContext: {},
+      targetObject: EmberObject.create({
+        anEvent(ctx) {
+          ok(obj.actionContext === ctx, 'anEvent method was called with the expected context');
+        }
+      })
+    });
+  }, /Usage of `targetObject` is deprecated. Please use `target` instead./);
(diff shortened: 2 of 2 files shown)


rwjblue:
This is good for the specifying of a `targetObject`, but we also need to deprecate folks asking for it (i.e. `this.get('targetObject')`). 

The thought was that we would need to use an actual JS getter with a deprecation for that. Something like:

```js
import { descriptor }  from 'ember-metal';

// ...snip...

 targetObject: descriptor({
    get() {
      deprecate('good message', false, { until: '2.12.0', url: 'some-path', id: 'ember-views.targetObject' });
      return this._targetObject;
    },

    set(value) {
      deprecate('good message', false, { until: '2.12.0', url: 'some-path', id: 'ember-views.targetObject' });
      return this._targetObject = value;
    }
  })
```

sduquej:
Thanks, @rwjblue. That was really helpful.
Is this looking better?

pixelhandler:
@sduquej Can you add the code (provided by @rwjblue) to define `targetObject` as deprecated via set/set to the mixin as well ? 

- https://github.com/emberjs/ember.js/blob/master/packages/ember-runtime/lib/mixins/target_action_support.js ?

sduquej:
Tried that @pixelhandler, but then ran into tests failing as classes that include the `Ember.TargetActionSupport` mixin would now have a `targetObject` property (the descriptor). 

There's a few places where the framework iterates over all properties in an object so `targetObject` would come up raising the deprecation warning, for example here: https://github.com/emberjs/ember.js/blob/master/packages/ember-runtime/lib/inject.js#L56

It didn't feel right to add special-case handling in these places as they were a consequence of adding the code to the mixin, which is why we came up with the `init` approach.

rwjblue:
Ya, we have to make it non-enumerable (which will cause it to not show up in list of keys).

sduquej:
I totally missed that option 😅. Have rebased and udpated the PR accordingly.

sduquej:
What's the best way to silence deprecations within the framework itself? 
I read https://github.com/emberjs/rfcs/blob/master/text/0065-deprecation-warning-handlers.md
but found no usages outside of `ember-debug`. 

The build is failing in the production suite currently because I want to silence deprecations here:
https://github.com/emberjs/ember.js/pull/14590/files#diff-eba568646bdc7cb8b35def52b7e49e9cR173

MiguelMadero:
Is the idea to use `target` instead? 

In Ember 2.7 we used to be able to read and set targetObject. It's gone in 2.9, but using _targetObject worked. While both are private they worked. By reading this, if I understood correctly the idea is to use `target` instead, but that didn't work in the way I was using it on https://github.com/toranb/ember-cli-hot-loader/pull/34

We have inside a component wrapper something like:

```
{{#component wrappedComponentName _targetObject =_targetObject 
```

I understand based on the deprecation warning that we were supposed to use `target` instead, however that didn't work either

```
{{#component wrappedComponentName target =target 
```

What worked across versions was, but that's supposed to be deprecated is reading from `_targetObject` and setting `targetObject`: 

```
{{#component wrappedComponentName targetObject =_targetObject 
```

packages/ember-runtime/lib/mixins/target_action_support.js
```diff
@@ -159,5 +165,23 @@ function getTarget(instance) {
     }
   }
 
+  registerHandler(function(message, options, next) {
```
rwjblue: I think we should remove this line and the line below that does `target = get(instance, 'targetObject')` (so that we don't trigger the deprecation).

Things should fall through now that we have the getter/setter above.

sduquej: :+1: Updated

packages/ember-runtime/lib/mixins/target_action_support.js
```diff
@@ -144,5 +148,17 @@ function getTarget(instance) {
...
+
+  // if _targetObject use it
+  if (instance._targetObject) { return instance._targetObject; }
+
   return null;
 }
+
+function _deprecateTargetObject() {
+  let message = 'Usage of `targetObject` is deprecated. Please use `target` instead.';
+  let options = { id: 'ember-runtime.using-targetObject', until: '2.15.0' };
```
sduquej: Changed this to `2.15.0`. Perhaps we'd like it to be a later version, if so which?

rwjblue: Ya, it will need to be 2.17.0 at this point.

sduquej:
@rwjblue Thanks! Updated.

GavinJoyce:
@rwjblue it would be good to get this merged. As it's been ~6 weeks since the [deprecation target was set for `2.17.0`](https://github.com/emberjs/ember.js/pull/14590#discussion_r134092731), perhaps it now needs to be `2.18.0`?

packages/ember-runtime/lib/mixins/target_action_support.js
```diff
@@ -24,6 +25,18 @@ doing more complex event handling in Components.
 */
 export default Mixin.create({
   target: null,
+  targetObject: descriptor({
+    configurable: true,
+    enumerable: false,
+    get() {
+      _deprecateTargetObject();
```
rwjblue: Unfortunately, I think we actually want to duplicate the code for this helper method so that it can properly be stripped from production builds.

Can you move the code into both the `get` and `set` directly?

packages/ember-runtime/lib/mixins/target_action_support.js
```diff
@@ -144,5 +148,17 @@ function getTarget(instance) {
...
+  if (target) { return target; }
+
+  // if _targetObject use it
+  if (instance._targetObject) { return instance._targetObject; }
+
   return null;
 }
+
+function _deprecateTargetObject() {
+  let message = 'Usage of `targetObject` is deprecated. Please use `target` instead.';
```
rwjblue: I think we should pass a bit more debugging info here.  If I saw this message in the console, I'm not sure how I would track down where it was coming from.  Perhaps you can include `this` (which will display something like `<component:foo-bar#1234>` which might give someone a clue...).

packages/ember-runtime/lib/mixins/target_action_support.js
```diff
@@ -144,5 +148,17 @@ function getTarget(instance) {
...
+  let options = { id: 'ember-runtime.using-targetObject', until: '2.17.0' };
```
rwjblue: I've let this slip yet again, and I am very sorry.  Can you update this again to 3.5.0?

rwjblue (changes requested):

sduquej:
@rwjblue Updated! 🙇

rwjblue (approved):
Thank you very much for your patience and hard work here. Sorry for all the delays, great work! 👏👏🎉

sduquej:
🍻 @rwjblue No problem at all! Perhaps you'd like to close the issue (https://github.com/emberjs/ember.js/issues/14168) now?

PR #14602, opened 2016-11-11
[FEATURE ember-metal-meta-destructors]

rwjblue:
Introduces a mechanism to add additional work to be done at object / meta destruction time.

Example:

```js
// app/utils/run-later.js
export default function runLater(object, functionOrName, delay) {
  let cancelId = run.later(object, functionOrName, delay);
  Ember.meta(object).addDestructor(() => run.cancel(cancelId));
}
```

```js
// some component
  stuff() { },

  actions: {
    doStuff() {
      runLater(this, this._stuff, 1000);
    }
  }
})
```

When this component is destroyed, the scheduled task will be canceled. A number of popular addons do similar things by monkey patching the objects `willDestroy` at runtime, but that is less than ideal (due to shape changes and whatnot).

This provides a mechanism to easily entangle cleanup tasks for objects from utility functions...

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -58,3 +58,35 @@ for a detailed explanation.
 * `ember-testing-resume-test`
 
   Introduces the `resumeTest` testing helper to complement the `pauseTest` helper.
+
+* `ember-metal-meta-destructors`
+
+  Introduces a mechanism to add additional work to be done at object / meta destruction time.
+
+  Example:
+
+  ```js
+  // app/utils/run-later.js
+  export default function runLater(object, functionOrName, delay) {
+    let cancelId = run.later(object, functionOrName, delay);
+    Ember.meta(object).addDestructor(() => run.cancel(cancelId));
+  }
+  ```
+
+  ```js
+  // some component
+  export default Component.extend({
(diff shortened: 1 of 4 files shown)


rwjblue:
Would love review/feedback from folks I know have been working on and around similar issues to my example above.

/cc @machty @krisselden @stefanpenner @runspired

machty:
Alternatively, wouldn't it be possible to implement destructors using ember-metal's listener APIs and just broadcast a willDestroy event? Then anyone will be able to do `Ember.on('willDestroy', ()=>{})` and have all the destructors determined up front at mixin time. Not to mention people would expect there to be `willDestroy` events if there's an `init` event.

runspired:
I think we should go all in on cancellation tokens as the primitive for these things, and I'm hesitant to grow the API surface area instead of doubling down on building a great async story top to bottom.

But otherwise, looks good, seems forward compatible.

locks:
Not applicable anymore.

rwjblue:
For those that might stumble across this old PR, this was RFC'ed in https://github.com/emberjs/rfcs/issues/545

PR #12848, opened 2016-01-20
Make dependencies that end in `@each` expand to `[]`.

krisselden:
Maybe should also add an assertion about `/@each\..*\./`  to this, but likely should be a separate PR.

PR #14605, opened 2016-11-14
improve doc for filter CP

Leooo:
Was really not obvious to me before that the `dependentKey` for a `filter` CP could be anything different than the array used to filter.

### DIFF
--- a/packages/ember-runtime/lib/computed/reduce_computed_macros.js
+++ b/packages/ember-runtime/lib/computed/reduce_computed_macros.js
@@ -278,6 +278,28 @@ export function mapBy(dependentKey, propertyKey) {
   hamster.get('remainingChores'); // [{name: 'write more unit tests', done: false}]
   ```
 
+  You can also use `@each.property` in your dependent key, the callback will still use the underlying array:
+
+  ```javascript
+  let Hamster = Ember.Object.extend({
+    remainingChores: Ember.computed.filter('chores.@each.done', function(chore, index, array) {
+      return !chore.get('done');
+    })
+  });
+
+  let hamster = Hamster.create({
+    chores: Ember.A([
+      Ember.Object.create({ name: 'cook', done: true }),
+      Ember.Object.create({ name: 'clean', done: true }),
+      Ember.Object.create({ name: 'write more unit tests', done: false })
+    ])
+  });
+  hamster.get('remainingChores'); // [{name: 'write more unit tests', done: false}]
+  hamster.get('chores').objectAt(2).set('done', true);
+  hamster.get('remainingChores'); // []
+  ```
+
+
   @method filter
   @for Ember.computed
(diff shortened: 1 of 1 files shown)


rwjblue (changes requested):
Maybe we can give both examples? I think the example that exists now is probably the more common scenario so we should just give both...

Leooo:
@rwjblue amended

packages/ember-runtime/lib/computed/reduce_computed_macros.js
```diff
@@ -278,6 +278,28 @@ export function mapBy(dependentKey, propertyKey) {
   hamster.get('remainingChores'); // [{name: 'write more unit tests', done: false}]
   ```
 
+  You can also use `@each` in your dependent key, the callback will still use the underlying array:
```
locks: 🤔  I'm a bit concerned people will hurriedly read and start using `chores.@each` (for example) without the leaf property. Maybe `@each.property`? Maybe it's not a concern.

Leooo:
@locks right, amended

packages/ember-runtime/lib/computed/reduce_computed_macros.js
rwjblue: @locks - FYI `chores.@each` works properly without deprecation as of https://github.com/emberjs/ember.js/pull/12848 (landed in 2.5.0).

packages/ember-runtime/lib/computed/reduce_computed_macros.js
```diff
@@ -278,6 +278,28 @@ export function mapBy(dependentKey, propertyKey) {
...
+  });
+
+  let hamster = Hamster.create({
+    chores: [
+      { name: 'cook', done: true },
+      { name: 'clean', done: true },
+      { name: 'write more unit tests', done: true }
+    ]
+  });
+  hamster.set('done', false);
```
pixelhandler: What is this setting? should the example be to set the last item of `chores` as `done: false`? or perhaps just remove the `set` example and define one of the chores to have `done: false` in the call to `Hamster.create`

locks:
@Leooo hi, if you need help addressing @pixelhandler's comments let us know! Let's get this merged :D

Leooo:
didn't see @locks comment, on it thanks

Leooo:
@locks @pixelhandler better? I tested the last example, thanks for spotting the issue

locks:
@Leooo thank you for your work! :) Can you squash the commits for us, and I'll get rwjblue to re-review your PR?

Leooo:
@locks @rwjblue done thanks

locks (approved):

rwjblue (approved):

rwjblue:
Thank you!

Issue #5566, opened 2014-09-09
TransitionAborted when performing a full route transition with changed query params


skoryky:
Why does a TransitionAborted error get thrown when performing a full route transition when changing query params?

I am working off of the "Opt-in to full transition via refresh()" example provided at the bottom of the query parameters Ember guide (http://emberjs.com/guides/routing/query-params/).

The only change I made was to log errors:

```
Ember.RSVP.configure('onerror', function(error) {
  Ember.Logger.assert(false, error);
});
```

Here is an updated JS Bin: http://jsbin.com/kinajoginedo/1/edit?console,output

If you click the "Change it" button, you'll see a TransitionAborted error thrown in the console.

I've been trying to upgrade Ember in my app from 1.7.0-beta.1+canary for some time now, but many of my QUnit tests fail when I run them all together. Most of my tests are fine if I run them in isolation. I suspect the TransitionAborted errors may be causing the test failures. However, I'm not sure if TransitionAborted exceptions are normal when performing full transitions. If anyone can confirm one way or the other, that would be very helpful in getting to the root cause of my problems, whatever they might be.

btecu:
I'm running into this issue with Ember 1.11.0-beta.4. Updates?

gorandev:
Facing this issue too. It only bugs me because it gets logged as an error -- works fine otherwise (I'm doing pagination and the pages change and everything works fine).

tim-evans:
Seeing this on 1.11.0-beta.1+canary.26c1084c.

r4m:
I'm facing this issue to and as @gorandev it is only logged but it does not compromise the app flow.

tim-evans:
Hmm...

This is the behavior that I'm trying to implement (which I'm not able to do at the moment):

A user can type (normally) in a search box, which refreshes the model of a route. Currently, if I swallow the TransitionAborted errors, the typing UI doesn't respond well and will replace the text box with the finished transition. I would prefer to discard requests that pile up rather than wait for them to resolve.

```javascript
export default Ember.Route.extend({
  queryParams: {
    q: { replace: true, refreshModel: true },
    sort: { replace: true, refreshModel: true },
    filter: { replace: true, refreshModel: true }
  },

  model: function(params) {
     return this.store.find('user', params).then(null, (error) => {
        if (error.message === "TransitionAborted") {
          return [];
        }
        throw error;
     });
  }
}
```

gerry3:
[...]

PR #14607, opened 2016-11-14
[ISSUEFIX] Break loop after one refresh in queryParamsDidChange

### LINKED ISSUE #14597: Optimization for Ember.Route.actions.queryParamsDidChange
Hey, sorry if I'm seeing this wrong, but looking at this line:
https://github.com/emberjs/ember.js/blob/v2.9.1/packages/ember-routing/lib/system/route.js#L776
it seems that after finding one queryParam that needs to trigger a refresh the loop should break.
If 2 parameters changed there's no need to trigger 2 refreshes at the same time, right?
One refresh should be enough.

daibhin:
Fixes #14597

### DIFF
--- a/packages/ember-routing/lib/system/route.js
+++ b/packages/ember-routing/lib/system/route.js
@@ -826,6 +826,7 @@ let Route = EmberObject.extend(ActionHandler, Evented, {
         let qp = qpMap[totalChanged[i]];
         if (qp && get(this._optionsForQueryParam(qp), 'refreshModel') && this.router.currentState) {
           this.refresh();
+          break;
         }
       }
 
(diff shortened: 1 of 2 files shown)


rwjblue:
I'd suggest adding a test similar to [this one](https://github.com/emberjs/ember.js/blob/d37d38b8a21f859ecb8b62f74a9dea92b0289866/packages/ember/tests/routing/query_params_test.js#L527-L568), that overrides `refresh` and confirms it is only called once when two params change.

rwjblue:
Test looks good!  Can you confirm that it fails without your changes?

daibhin:
Did a manual confirmation and the test indeed fails without the `break;`. Is there anything I need to do to confirm this?

rwjblue:
Nope, looks great to me! Thank you!!

daibhin:
@wagenet this was to address https://github.com/emberjs/ember.js/issues/14597. Not fully sure of the context in https://github.com/emberjs/ember.js/issues/5566 but I think that's a different issue.

wagenet:
Ah I see. Thanks :)

Issue #14334, opened 2016-09-22
Second level of service are not injected when calling from template


cbou:
When calling a service function from a template e.g. `onClick={{action myFirstLevel.hello}}`, the services which should be injected into `myFirstLevel` stay undefined.

The call through a component action is working.

```hbs
<button onClick={{action myFirstLevel.hello}}>Hello Service</button>
<button onClick={{action 'hello'}}>Hello Action</button>
```

```javascript
App = Ember.Application.create();

App.MyFirstLevelService = Ember.Service.extend({
  mySecondLevel: Ember.inject.service(),
  hello: function() {
    console.log('Hello first level');
    this.get('mySecondLevel').hello();
  }
});

App.MySecondLevelService = Ember.Service.extend({
  hello: function() {
    console.log('Hello second level');
  }
});

App.MyButtonComponent = Ember.Component.extend({
  myFirstLevel: Ember.inject.service(),
  
  actions: {
    hello:  function() {
      this.get('myFirstLevel').hello();
    }
  }
});
```

http://emberjs.jsbin.com/conaxaheno/1/edit?html,js,console,output

locks:
If you log `this` from within the `hello` function of `MyFirstLevelService`, you'll notice that it's `MyButtonComponent`. So the problem isn't that the service isn't being injected, but that `{{action}}` binds the action to the current context, which is this case is `MyButtonComponent`.
Here's how to make it work:

    <button onClick={{action myFirstLevel.hello target=myFirstLevel}}>Hello Service</button>

Hope this helps!

cbou:
Thanks for your help. Unfortunately it didn't worked.

locks:
@cbou you can do `<button {{action myFirstLevel.hello}}>Hello Service</button>`. Doing the DOM event handler instead messes up the contexts.

cbou:
@locks Yes this work now. Is that a bug?

I'm often confused between both syntax and I'm sure I'm not alone. Te me both should act exactly the same way. Is there a reason why it's not the case?

locks:
`<button {{action 'actionName'}}>` defines an Ember action on the button. Ember is responsible for managing it.
`<button onClick={{action 'actionName'}}>` defines an event handler for the DOM `onClick` event. So it's basically the browser that manages it, and eventually triggers the event handler.

You can ask in our Slack channel for a more technical explanation, I don't know the nitty gritty.

PR #14619, opened 2016-11-19
[DOC release] Update `get` helper action code sample

samselikoff:
Got tripped up by this today, there was another outdated sample.

I think pushing people towards onevent= is better?

### DIFF
--- a/packages/ember-glimmer/lib/helpers/get.js
+++ b/packages/ember-glimmer/lib/helpers/get.js
@@ -36,16 +36,16 @@ import {
 
   ```handlebars
   {{get person factName}}
-  <button onclick={{action (action (mut factName)) "height"}}>Show height</button>
-  <button onclick={{action (action (mut factName)) "weight"}}>Show weight</button>
+  <button {{action (action (mut factName)) "height"}}>Show height</button>
+  <button {{action (action (mut factName)) "weight"}}>Show weight</button>
   ```
 
   The `{{get}}` helper can also respect mutable values itself. For example:
 
   ```handlebars
   {{input value=(mut (get person factName)) type="text"}}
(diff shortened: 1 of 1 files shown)


locks:
@samselikoff these are two different things though, the event handler will receive the event as an argument, while the action won't. The correct fix is the ugly double action helper: `{{action (action (mut factName) "height")}}`.

locks (changes requested):
Do `{{action (action (mut factName) "height")}` instead because of the differing semantics of DOM event handlers.

samselikoff:
Updated, should we `<button {{action` or `<button onclick={{action`?

locks:
@samselikoff sorry for the delay, IMO it should be `<button {{action` so that the event handler gets scheduled with the global dispatcher thingie.

samselikoff:
Should be all set

locks (approved):

locks:
Thank yoouu 👯‍♂️

nightire:
@locks May I ask what's the difference between `<button {{action` and `<button onclick={{action` behind the scene? I can't find concrete explanation through the api documentation, be more specific, I'm really eager to understand "...gets scheduled with the global dispatcher thingie", what does it mean?

kumkanillam:
In [this thread](https://github.com/emberjs/ember.js/issues/14334) @locks explained ,
<button {{action 'actionName'}}> defines an Ember action on the button. Ember is responsible for managing it.
<button onClick={{action 'actionName'}}> defines an event handler for the DOM onClick event. So it's basically the browser that manages it, and eventually, triggers the event handler.

nightire:
@kumkanillam I see, thanks for your explanation!

PR #14629, opened 2016-11-22
ember-testing/setup_for_testing: Use Ember.Test.Adapter if QUnit is unavailable

### LINKED ISSUE #14628: Deprecate Ember.Test.QUnitAdapter and Ember.Test.adapter defaulting to it
I would like to propose the deprecation of the `Ember.Test.QUnitAdapter` class and the `Ember.Test.adapter` property defaulting to it. Instead we should move the `QUnitAdapter` implementation into the `ember-qunit` project and default to using the empty `Ember.Test.Adapter` base implementation.

Since this would be a breaking change we should copy the existing stuff over to `ember-qunit`, deprecate it in `ember.js` and then remove it once Ember 3.0 is released. At that point the deprecation warning about `Ember.Test.adapter` not being explicitly set should likely be replaced by the default moving to `Ember.Test.Adapter`.

Turbo87:
Resolves #14628 

/cc @rwjblue @trentmwillis

### DIFF
--- a/packages/ember-testing/lib/setup_for_testing.js
+++ b/packages/ember-testing/lib/setup_for_testing.js
@@ -1,3 +1,5 @@
+/* global self */
+
 import { setTesting } from 'ember-metal';
 import { jQuery } from 'ember-views';
 import {
@@ -9,6 +11,7 @@ import {
   decrementPendingRequests,
   clearPendingRequests
 } from './test/pending_requests';
+import Adapter from './adapters/adapter';
 import QUnitAdapter from './adapters/qunit';
 
 /**
@@ -29,7 +32,7 @@ export default function setupForTesting() {
   let adapter = getAdapter();
   // if adapter is not manually set default to QUnit
   if (!adapter) {
-    setAdapter(new QUnitAdapter());
+    setAdapter((typeof self.QUnit === 'undefined') ? new Adapter() : new QUnitAdapter());
   }
 
   jQuery(document).off('ajaxSend', incrementPendingRequests);
--- a/packages/ember-testing/tests/adapters_test.js
+++ b/packages/ember-testing/tests/adapters_test.js
@@ -4,11 +4,12 @@ import Adapter from '../adapters/adapter';
 import QUnitAdapter from '../adapters/qunit';
 import { Application as EmberApplication } from 'ember-application';
 
-var App, originalAdapter;
+var App, originalAdapter, originalQUnit;
 
 QUnit.module('ember-testing Adapters', {
   setup() {
     originalAdapter = Test.adapter;
+    originalQUnit = window.QUnit;
   },
   teardown() {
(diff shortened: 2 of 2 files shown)


packages/ember-testing/lib/setup_for_testing.js
```diff
@@ -29,7 +30,7 @@ export default function setupForTesting() {
   let adapter = getAdapter();
   // if adapter is not manually set default to QUnit
   if (!adapter) {
-    setAdapter(new QUnitAdapter());
+    setAdapter((typeof self.QUnit === 'undefined') ? new Adapter() : new QUnitAdapter());
```
rwjblue: Can you move this logic into a new module (I'd suggest `./adapters/default`)? Then we can simply do:

```js
import DefaultAdapter from './adapter/default';
//..snip...
setAdapter(new DefaultAdapter());
```

packages/ember-testing/tests/adapters_test.js
```diff
@@ -50,3 +50,22 @@ QUnit.test('QUnitAdapter is used by default', function() {
 
   ok(Test.adapter instanceof QUnitAdapter);
 });
+
+QUnit.test('Adapter is used by default (if QUnit is not available)', function() {
+  expect(2);
+
+  var originalQUnit = window.QUnit;
```
rwjblue: Please stash this in a module scoped var in the module `setup` above.

packages/ember-testing/tests/adapters_test.js
```diff
@@ -50,3 +50,22 @@ QUnit.test('QUnitAdapter is used by default', function() {
...
+  delete window.QUnit;
+
+  Test.adapter = null;
+
+  run(function() {
+    App = EmberApplication.create();
+    App.setupForTesting();
+  });
+
+  window.QUnit = originalQUnit;
```
rwjblue: Please reset this in the modules `teardown` above.  We need to ensure that this test failing does not completely hose all the rest of the tests by not cleaning up after itself properly.

rwjblue (changes requested):

packages/ember-testing/lib/setup_for_testing.js
Turbo87: hmm, I'm not sure how that would work. In the current state it checks the `QUnit` global at the time of the `setAdapter()` call, but if we put it inside a separate module we would have to do something like `export default (hasQUnit ? QUnitAdapter : Adapter);` and move the check to boot-time, which makes it harder to test.

packages/ember-testing/tests/adapters_test.js
Turbo87: 👍 

not familiar with QUnit... 😉

Turbo87: @rwjblue done!

PR #14636, opened 2016-11-24
sync initializer and instance-initializer blueprints

kellyselden:
This unifies the older style in initializer-test with the newer style in [instance-initializer-test](https://github.com/emberjs/ember.js/blob/39b2707900a31f9de541f080817f2102e2c9323c/blueprints/instance-initializer-test/qunit-files/tests/unit/instance-initializers/__name__-test.js). The only logic change is that it now calls `destroyApp` like the instance-initializer-test does. That bit should be determined if it is needed.


rwjblue:
Thanks @kellyselden! We also need to update the legacy blueprints repo versions too.

kellyselden:
@rwjblue Will do. Was waiting to see if this was approved.

PR #14637, opened 2016-11-25
Avoid using `let` in `ember-cli-build.js`

patsy-issa:
"SyntaxError: Could not require 'ember-cli-build.js': Block-scoped declarations (let, const, function, class) not yet supported outside strict mode"


rwjblue:
Change seems good for now. We should also add use strict to this file (and you should upgrade your node version 😜)...

Issue #14514, opened 2016-10-24
Warning about usage of @each needs to show more details


adambuczynski:
The following warning

```
WARNING: Dependent keys containing @each only work one level deep. You cannot use nested
forms like todos.@each.owner.name or todos.@each.owner.@each.name. Please create an 
intermediary computed property.
```

doesn't really tell me _where_ in my code the problem lies. This makes it very hard to debug, especially since after searching for `@each` in our entire code base, we can't find a single place where we are using `@each` more than one level deep.

Can this error message be clarified with more information about which property in which component is the offending one? Thanks.

btecu:
Most likely you have a computed property based off an `@each` and then you have another computed property with an `@each` based on that first one.

adambuczynski:
> Most likely you have a computed property based off an @each and then you have another computed property with an @each based on that first computed property.

@btecu that doesn't make sense, because the message clearly states "Please create an 
intermediary computed property", which implies that that should be the correct way to handle this.

mmun:
@adambuczynski This could definitely be improved. Thanks for pointing it out!

adambuczynski:
@mmun no worries, thanks for looking into it.

Serabe:
@mmun: this would solve #13622 as well and if you find a general solution, then we can close #13629 as well.

GavinJoyce:
This should help: https://github.com/emberjs/ember.js/pull/14639

rwjblue:
Closed by https://github.com/emberjs/ember.js/pull/14639.

Issue #13622, opened 2016-06-08
Confusing warning message for computed.mapBy('foo', 'bar.baz')


Turbo87:
When using a computed property like `computed.mapBy('foo', 'bar.baz')` I'm getting the following warning:

> WARNING: Dependent keys containing `@each` only work one level deep. You cannot use nested forms like `todos.@each.owner.name` or `todos.@each.owner.@each.name`. Please create an intermediary computed property.

It took me quite a while to figure out that the `mapBy` is actually the problem and using `@each` under the hood. It might be worth it to throw a more descriptive warning if `mapBy` is used improperly (i.e. with a nested path).

rwjblue:
Agreed

swastik:
I took a shot at this [here](https://github.com/emberjs/ember.js/pull/13629).

I wasn't too sure what to tag the commit—it isn't a new feature, not a bugfix; it's likely enhanced documentation, but I wasn't sure—so I've left it untagged.

I've also added a test. Let me know if I should change anything! :smiley:

GavinJoyce:
@rwjblue I believe that https://github.com/emberjs/ember.js/pull/14639 allows us to close this too

rwjblue:
Sounds good, thanks!

PR #14639, opened 2016-11-26
[BUGFIX beta] improve nested @each property key warning

GavinJoyce:
I think this should be enough to help people easily locate the offending properties in their app. If so, we should be able to close:

 * https://github.com/emberjs/ember.js/issues/14514
 * https://github.com/emberjs/ember.js/issues/13622

When using a key such as `comments.@each.author.name`, we'll now get the warning:

```
WARNING: Dependent keys containing @each only work one level deep. 
You used the key "comments.@each.author.name" which is invalid. 
Please create an intermediary computed property.
```

instead of the previous generic message:

```
WARNING: Dependent keys containing @each only work one level deep. 
You cannot use nested forms like todos.@each.owner.name or todos.@each.owner.@each.name. 
Please create an intermediary computed property.
```

TODO:

 * [x] Verify that this gives a good warning with `computed.mapBy('todos', 'owner.name')`

```js
people: [
  { 
    nested: { name: 'Alex' } 
  }, 
  { 
    nested: { name: 'Ben' } 
  }
],
names: Em.computed.mapBy('people', 'nested.name')
```

gives:

screen shot 2016-11-26 at 08 49 04

### DIFF
--- a/packages/ember-metal/lib/computed.js
+++ b/packages/ember-metal/lib/computed.js
@@ -249,7 +249,7 @@ ComputedPropertyPrototype.property = function() {
   function addArg(property) {
     warn(
       `Dependent keys containing @each only work one level deep. ` +
-        `You cannot use nested forms like todos.@each.owner.name or todos.@each.owner.@each.name. ` +
+        `You used the key "${property}" which is invalid. ` +
           `Please create an intermediary computed property.`,
       DEEP_EACH_REGEX.test(property) === false,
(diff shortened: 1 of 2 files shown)


GavinJoyce:
Here's a before and after for our app:

**before:**

screen shot 2016-11-26 at 08 55 45

**after:**

screen shot 2016-11-26 at 08 53 07

locks (approved):
Great work :)

rwjblue:
Awesome!

Issue #14647, opened 2016-11-28
[CLEANUP] Remove enabled features


chancancode:
![bitmoji](https://render.bitstrips.com/v2/cpanel/10213084-50269186_6-s1-v1.png?transparent=1&width=246)

https://github.com/emberjs/ember.js/blob/master/features.json

Example: https://github.com/emberjs/ember.js/commit/b8bc459dbcd808dbc66d8811eb7116fa0a01831b

- [ ] ember-runtime-computed-uniq-by
- [ ] ember-runtime-enumerable-includes
- [ ] ember-string-ishtmlsafe
- [ ] ember-testing-check-waiters

GavinJoyce:
I'll cut a PR for `ember-runtime-computed-uniq-by` now

Patsy-issa:
Working on a pr for all of the above.

vvscode:
@chancancode  this issue can be closed?

locks:
Yup, thanks for pointing out!

PR #14648, opened 2016-11-28
Cleanup features

patsy-issa:
Removes enabled features check #14647

### DIFF
--- a/packages/ember-runtime/lib/mixins/array.js
+++ b/packages/ember-runtime/lib/mixins/array.js
@@ -21,8 +21,7 @@ import Ember, { // ES6TODO: Ember.A
   removeListener,
   sendEvent,
   hasListeners,
-  deprecate,
-  isFeatureEnabled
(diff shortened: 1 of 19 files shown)


packages/ember-runtime/lib/mixins/array.js
```diff
@@ -582,55 +579,53 @@ const ArrayMixin = Mixin.create(Enumerable, {
...
-      @return {Boolean} `true` if object is found in the array.
-      @public
-    */
-    includes(obj, startAt) {
-      let len = get(this, 'length');
-
-      if (startAt === undefined) {
-        startAt = 0;
-      }
+ArrayMixin.reopen({
```
bmeurant: I think `includes` can be now implented directly in the Mixin, no ? Since we do not have to check if the feature is enabled anymore, we do not have to use `reopen` here, I guess.

locks (approved):

PR #14658, opened 2016-11-30
[BUGFIX beta] Make the ember-source build work.

nathanhammond:
The build in 2.11.0-beta.1 is "broken" as `dist` doesn't include jQuery. It should do so dynamically so that it gets SemVer drift instead of being hard-set bundled per build.

It should also prefer a user-installed version of `jquery` to the version bundled with `ember-source`.

This PR:
- [X] Fixes the jQuery version lockdown bug.
- [X] Default uses the specified jQuery from the project directory.
- [X] Falls back to the version bundled with `ember-source` if not user-specified.

---
I think we broke this here: https://github.com/emberjs/ember.js/pull/14556/files

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,8 @@
 /* jshint node: true */
 'use strict';
 var stew = require('broccoli-stew');
+var path = require('path');
+var resolve = require('resolve');
 
 var paths = {};
 var absolutePaths = {};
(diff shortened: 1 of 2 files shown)


nathanhammond:
Also, how should we allow end-users to override and set their own `jquery` version?

index.js
```diff
@@ -32,6 +34,19 @@ module.exports = {
   paths: paths,
   absolutePaths: absolutePaths,
   treeForVendor: function() {
+
+    var jqueryPath;
+    try {
+      jqueryPath = path.dirname(resolve.sync('jquery/package.json', { basedir: this.project.root }));
+    } catch (error) {
+      jqueryPath = path.dirname(require.resolve('jquery/package.json'));
+    }
```
chadhietala: Do we want to do a `finally` here and to make sure we a `jqueryPath` and if not throw?

nathanhammond: If we don't have a `jqueryPath` at this point something else is going to have resulted in this breaking _long_ before we got this far. The one in the `catch` statement is depended upon by `ember-source` itself.

package.json
```diff
@@ -30,7 +30,6 @@
     "broccoli-funnel": "^1.0.6",
     "broccoli-merge-trees": "^1.1.4",
     "broccoli-rollup": "^1.0.3",
-    "broccoli-stew": "^1.2.0",
```
nathanhammond: This was already in `dependencies`.

index.js
chadhietala: SG

PR #14670, opened 2016-12-02
[BUGFIX beta] Add blueprints for "ember-cli-mocha >= 0.12.0"

### LINKED ISSUE #14667: Add blueprints for "ember-cli-mocha >= 0.12.0"
`ember-cli-mocha@0.12.0` introduced a [new testing API](https://github.com/emberjs/ember-mocha#upgrading) and deprecated the old `describeXXX()` helpers. The blueprints in this repository need to be updated to reflect those changes.

Instead of just updating we should make sure that uses with all versions of `ember-cli-mocha` get the right blueprints. We can accomplish that by checking the installed version of `ember-cli-mocha` using a [VersionChecker](https://github.com/rwjblue/ember-cli-version-checker) and then deciding whether to use the `mocha-files` or `mocha-0.12-files` blueprints folder.

The tracking issue for this is https://github.com/ember-cli/ember-cli-mocha/issues/142

Turbo87:
Resolves #14667

/cc @rwjblue @Patsy-issa

### DIFF
--- a/blueprints/test-framework-detector.js
+++ b/blueprints/test-framework-detector.js
@@ -1,4 +1,6 @@
+var fs = require('fs');
 var path = require('path');
+var VersionChecker = require('ember-cli-version-checker');
 
 module.exports = function(blueprint) {
   blueprint.supportsAddon = function() {
@@ -12,7 +14,12 @@ module.exports = function(blueprint) {
     if ('ember-cli-qunit' in dependencies) {
       type = 'qunit';
     } else if ('ember-cli-mocha' in dependencies) {
-      type = 'mocha';
+      var checker = new VersionChecker({ project: this.project });
+      if (fs.existsSync(this.path + '/mocha-0.12-files') && checker.for('ember-cli-mocha', 'npm').satisfies('>=0.12.0')) {
+        type = 'mocha-0.12';
+      } else {
+        type = 'mocha';
+      }
     } else {
       this.ui.writeLine('Couldn\'t determine test style - using QUnit');
       type = 'qunit';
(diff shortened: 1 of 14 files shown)


Turbo87:
discovered a small issue with this, do not merge for now... 😞

blueprints/test-framework-detector.js
```diff
@@ -12,7 +14,12 @@ module.exports = function(blueprint) {
     if ('ember-cli-qunit' in dependencies) {
       type = 'qunit';
     } else if ('ember-cli-mocha' in dependencies) {
-      type = 'mocha';
+      if (fs.existsSync(this.path + '/mocha-0.12-files') &&
+        semver.satisfies('0.12.0', dependencies['ember-cli-mocha'])) {
+        type = 'mocha-0.12';
+      } else {
+        type = 'mocha';
```
rwjblue: I think we need to flip-flop this logic, I (and I could be going out on a limb here) think that the new syntax will likely stay around even if we have to make a 0.13 / 1.0 / etc of ember-mocha.  We should update this conditional so the `else` is the new format.

I'd also suggest `mocha-current` and `mocha-lt-0.12` as the folder paths.

Thoughts?

rwjblue (changes requested):
Thank you for putting this together!

blueprints/test-framework-detector.js
Turbo87: @rwjblue that is precisely what I meant with "small issue" above. I've fixed it in https://github.com/emberjs/data/pull/4691 by using the `VersionChecker` class and generating fake package manifests for the mocha blueprint tests. will update this PR now too.

Turbo87: @rwjblue this should be resolved now

Turbo87: regarding the folder names: I wanted to keep the diff small, so I decided to leave the old blueprints at their original location

rwjblue (approved):

PR #14673, opened 2016-12-02
[CLEANUP] Remove reference to closure components

Serabe:
While contextual components were being developed, I referred to them to
closure components since initial implementation looked a lot like
closure actions (except for the invocation part). Also because English
is not my first language and I was too lazy to create a switch in my
head for this.

Sadly, this error is still in the codebase after more than a year though
I might admit that I felt guilt, shame and a bit of pain each time I had
to touch the file since then.

This PR fixes my conscience and, on top of that, some naming
inconsistency.


rwjblue:
Thanks!

PR #14675, opened 2016-12-02
Add d8-*.js to .npmignore.

rwjblue:


Turbo87:
shouldn't you also remove those files so that the rule works properly?

rwjblue:
I'm not sure I follow. .npmignore has nothing to do with files being present or not. Maybe you mistook this for the gitignore?

Turbo87:
you're right... time for bed!

rwjblue:
Hehe, fair enough ;P

Issue #14655, opened 2016-11-29
ember 2.10 + jquery 3 issue


kellyselden:
I tried tracking it down to somewhere in my code, but I think it leads me to an issue with 2.10 using jquery 3.

ember 2.9 + jquery 2: [success](https://travis-ci.org/kellyselden/ember-datetimepicker/jobs/179623233)
ember 2.9 + jquery 3: [success](https://travis-ci.org/kellyselden/ember-datetimepicker/jobs/179623234)
ember 2.10 + jquery 2: [success](https://travis-ci.org/kellyselden/ember-datetimepicker/jobs/179624749)
ember 2.10 + jquery 3: [fail](https://travis-ci.org/kellyselden/ember-datetimepicker/jobs/179624750)

When running the addon's dummy locally, there are no console errors. The addon, as you might guess, wraps a jquery datepicker plugin. I inspected, and I'm sending the right options hash to the plugin init, but the options are not taking effect in the datepicker.

Normally, I would point the finger at the datepicker, but it works fine with jquery 3. I just have to rollback to ember 2.9. It also works fine with ember 2.10 if I use jquery 2. So the combo of ember 2.10 and jquery 3 seems to be culprit.

I tried updating the jquery plugin to most recent version, no change.

Here is one of the success cases:
<img width="563" alt="screen shot 2016-11-29 at 10 05 10 am" src="https://cloud.githubusercontent.com/assets/602423/20722312/ec72f4ca-b61a-11e6-87b9-6078d5eb27d9.png">

Here is the failure case:
<img width="395" alt="screen shot 2016-11-29 at 10 06 36 am" src="https://cloud.githubusercontent.com/assets/602423/20722317/f38901c8-b61a-11e6-8062-54b5c8738bec.png">

Here is the branch to see it failing: https://github.com/kellyselden/ember-datetimepicker/tree/ember-bug-report

sandstrom:
A good place to start would be the jQuery changes between 2 -> 3.

This page has a list of breaking changes:
https://jquery.com/upgrade-guide/3.0/

Also, it's a bit unclear in what way it isn't working. I ran your branch and although the failure case looked like the image above it opens just fine when I click the input. I.e. the only thing I see is a small difference in the default state of the input. Perhaps you can explain in more detail?

![image](https://cloud.githubusercontent.com/assets/122287/20774015/c162cf68-b754-11e6-8903-d973de152ffe.png)

*image is from the ember-bug-report branch*

[...]

PR #14677, opened 2016-12-05
hash helper, failing test for #14655

kellyselden:
failing test for https://github.com/emberjs/ember.js/issues/14655
in ember 2.9, this assertion would be true.

### DIFF
--- a/packages/ember-glimmer/tests/integration/helpers/hash-test.js
+++ b/packages/ember-glimmer/tests/integration/helpers/hash-test.js
@@ -1,6 +1,6 @@
 import { RenderingTest, moduleFor } from '../../utils/test-case';
 import { Component } from '../../utils/helpers';
-import { set } from 'ember-metal';
+import { get, set } from 'ember-metal';
 
 moduleFor('Helpers test: {{hash}}', class extends RenderingTest {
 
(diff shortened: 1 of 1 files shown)


rwjblue:
Right, our hash objects are the result of `Object.create(null)`, and will not have a constructor.

rwjblue:
will comment further in #14655 though

stefanpenner:
this seems like expected behavior no?

rwjblue:
@stefanpenner - Yep, it is exactly what we expect. However, it would be nice if our objects could pass for a `isPlainObject` but due to the `new EmptyObject()` usage we don't (see comment in #14655 for links and stuff explaining why).

stefanpenner:
@rwjblue in this case, I believe using `Object.create(null)` is fine. And it sounds like it should address the issue.

My reasoning:

* some runtimes `Object.create(null)` is slow, but new v8 (and likely others) its getting better
* `Object.create(null)` several times, for each meta === was slow (10,000+ can get created for big renders) but for yielded blocks, it may not be not be an issue. As we do by definition create way way less of those, then we do metas.

chancancode:
@stefanpenner re: "occur much less", we make a few of these per each argument object (helper/component invocation and such) plus a couple of other places in Glimme, so it's probably still a non-trivial amount

mmun:
Closing, see https://github.com/emberjs/ember.js/issues/14655#issuecomment-368136353.

Issue #14672, opened 2016-12-02
Rendering aliases break when rendering a CP dependent on them


lukemelia:
Update: it doesn't have to do with yield, or attrs, it is just the order of rendering (ref consumption) and whether the CP is rendered (ref consumed). Updated fiddle: https://ember-twiddle.com/9f5a189ebe41a8191c1b280ce80592a1

Under Ember 2.10, when a component yields a property defined with computed.readOnly, under certain circumstances in *production builds of Ember only*, downstream usage are not updated when the property changes.
 
I worked with @krisselden to created an isolated reproduction of this issue:
 
https://ember-twiddle.com/868fbd21b0532d3dfa583ca43790dde7?openFiles=templates.application.hbs%2Ctemplates.components.consuming-comp.hbs

The reproduction demonstrates that the order that the attributes are passed into a consuming component determines which property fails to update. o_O

lukemelia:
Kris' hypothesis in our discussion is that there is a race in how AliasProperty (which readOnly uses) adds its dependent key and the consumption of the property, and that something in the development build papers over the bug.

krisselden:
It seems to be that consuming the CP breaks rendering of the dependent keys if they are aliases  https://ember-twiddle.com/9f5a189ebe41a8191c1b280ce80592a1

chancancode:
@lukemelia this was fixed in 2.10.1 😄

PR #14681, opened 2016-12-06
[BUGFIX release] Fix issue #14672

krisselden:
PROPERTY_DID_CHANGE is the mechanism by which Ember's Glimmer refs are dirtied.  Currently if something watches then unwatches an alias, the DK that is installed by the alias is removed.

### DIFF
--- a/packages/ember-metal/lib/alias.js
+++ b/packages/ember-metal/lib/alias.js
@@ -14,6 +14,8 @@ import {
   removeDependentKeys
 } from './dependent_keys';
 
+const CONSUMED = {};
+
 export default function alias(altKey) {
   return new AliasedProperty(altKey);
 }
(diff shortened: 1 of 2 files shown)


packages/ember-metal/lib/alias.js
```diff
@@ -28,33 +28,36 @@ AliasedProperty.prototype = Object.create(Descriptor.prototype);
...
+const CONSUMED = {};
 
+AliasedProperty.prototype.get = function AliasedProperty_get(obj, keyName) {
+  let meta = metaFor(obj);
+  let cache = meta.writableCache();
+  if (cache[keyName] !== CONSUMED) {
+    cache[keyName] = CONSUMED;
+    addDependentKeys(this, obj, keyName, meta);
+  }
   return get(obj, this.altKey);
```
stefanpenner: the above behavior is slightly different then how regular CP work: https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/lib/computed.js#L338-L349

* we add the DK first, then invoke the `get`. Should we follow the pattern of regular CP, add the DK after the first get?
* should `get` on alias be cached, until the DK changes?

krisselden: I'll change the order to be consistent, though getters really shouldn't have side effects.  As for caching, it is likely cached on the other side, I'm just trying to track unconsumed vs not.

stefanpenner: > though getters really shouldn't have side effects.

They shouldn't but they often do, especially in development mode where we `toString` objects, and users can provide there own `toString`

Not a big deal though, only doit if you are ok with it.

stefanpenner:
:shipit:

krisselden:
Trying to push an update to this but the push failed and but trying again says everything is up to date.

medokin:
@krisselden Will that come into the next 2.10 patch?

stefanpenner:
@medokin yes

chancancode:
@medokin released in 2.10.1 😄

PR #14685, opened 2016-12-06
[BUGFIX release] this.$() undefined in willDestroyElement hook

### LINKED ISSUE #14666: Ember 2.10.0 - Calling this.$() in willDestroyElement hooks during unit tests returns undefined
After upgrading to Ember 2.10.0, our component unit tests started breaking when `willDestroyElement` runs because `this.$()` is now returning `undefined`. This was not happening in Ember 2.9.0.

You can see this happening in [this ember-twiddle](https://ember-twiddle.com/fb2369455c41b581444f4b34d92f0cf6?openFiles=tests.unit.components.my-component-test.js%2C).

canufeel:
Trying to access `this.$()` in `willDestroyElement` now returns `undefined` as `willDestroyElement` hook is run after the component enters the `destroying` state since 3dcde0f was merged (2.9.0-beta.1). In versions before that `willDestroyElement` was called before transitioning to `destroying` state so the component was using `inDom` state object that would process `this.$()` properly.

Technically the element is still accessible but we have to use `$(this.element)` instead of `this.$()` which makes upgrading cumbersome as it is breaking apps that utilise `willDestroyElement` to deregister eventHandlers previously registered on component element. 

Fixes #14666

### DIFF
--- a/packages/ember-views/lib/views/states/destroying.js
+++ b/packages/ember-views/lib/views/states/destroying.js
@@ -1,7 +1,6 @@
 import { assign } from 'ember-utils';
 import { Error as EmberError } from 'ember-metal';
 import _default from './default';
-import jQuery from '../../system/jquery';
 /**
 @module ember
 @submodule ember-views
@@ -10,12 +9,6 @@ import jQuery from '../../system/jquery';
 const destroying = Object.create(_default);
 
 assign(destroying, {
-  $(view, sel) {
-    let elem = view.element;
-    if (elem) {
-      return sel ? jQuery(sel, elem) : jQuery(elem);
-    }
-  },
   appendChild() {
     throw new EmberError('You can\'t call appendChild on a view being destroyed');
   },
--- a/packages/ember-views/lib/views/states/destroying.js
+++ b/packages/ember-views/lib/views/states/destroying.js
@@ -1,6 +1,7 @@
 import { assign } from 'ember-utils';
 import { Error as EmberError } from 'ember-metal';
 import _default from './default';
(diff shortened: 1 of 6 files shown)


packages/ember-views/lib/views/states/destroying.js
```diff
@@ -9,6 +10,12 @@ import _default from './default';
 const destroying = Object.create(_default);
 
 assign(destroying, {
+  $(view, sel) {
```
krisselden: lets just move it out of states and just make this the method

krisselden (changes requested):
looks good to me otherwise

packages/ember-views/lib/views/states/destroying.js
canufeel: @krisselden Thanks for review! View mixin seems to be the most appropriate place for that method. PR updated.

chancancode:
@canufeel Thank you!

hemp:
With the context of #14666, thoughts on if this be worthy of cherry-picking back to 2.10.x since it introduced unexpected behavior during an upgrade from 2.9 to 2.10?  

Otherwise I think we might end up waiting to upgrade from 2.9 until 2.11 stable is cut (which is no problem).

chancancode:
Yes I'll do the cherry pick and release it next week!

hemp:
Ok, awesome! 🎉

wagenet:
@chancancode this is hitting our own app 😄

chancancode:
@hemp @canufeel @wagenet this is released in 2.10.2, thanks again for working on this!

PR #14687, opened 2016-12-07
[BUGFIX release] Fix issues with dist output

krisselden:
Fixes spurious sourceMappingURLs in ember.debug.js and ember.prod.js

Fix minified dist to not use negate iife.


rwjblue:
LGTM

PR #14695, opened 2016-12-08
[DOC] Simplify example for Ember.Route.modelFor

HeroicEric:
Some updates for the example for `Ember.Route.modelFor`

- Fixes indentation
- Removes unnecessary `resetNamespace`
- Updates example to avoid setting state on the route

### DIFF
--- a/packages/ember-routing/lib/system/route.js
+++ b/packages/ember-routing/lib/system/route.js
@@ -1855,9 +1855,9 @@ let Route = EmberObject.extend(ActionHandler, Evented, {
     // ...
 
     Router.map(function() {
-        this.route('post', { path: '/post/:post_id' }, function() {
-          this.route('comments', { resetNamespace: true });
-        });
+      this.route('post', { path: '/posts/:post_id' }, function() {
+        this.route('comments');
+      });
     });
 
     export default Router;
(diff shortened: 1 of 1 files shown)


packages/ember-routing/lib/system/route.js
```diff
@@ -1855,9 +1855,9 @@ let Route = EmberObject.extend(ActionHandler, Evented, {
     // ...
 
     Router.map(function() {
-        this.route('post', { path: '/post/:post_id' }, function() {
-          this.route('comments', { resetNamespace: true });
-        });
+      this.route('post', { path: '/post/:post_id' }, function() {
+        this.route('comments', { resetNamespace: true });
```
locks: Sorry to drop this on you, but do you know if we need `resetNamespace` here?

HeroicEric: Hmm, that's a good point. That isn't needed to demonstrate `modelFor` and would probably be useful to make the example as simple as possible.

Now that I'm looking at the example, setting `post` as a prop on the route also seems strange (`this.set('post', this.modelFor('post'))`). Let me see if I can make the example a little bit more straightforward.

HeroicEric:
@locks I updated the example to avoid setting state on the route and also removed the unneeded `resetNamespace` you mentioned.

locks:
👌

Issue #14692, opened 2016-12-08
API docs are often missing @since tags


Turbo87:
Our API docs are currently not versioned, meaning that we only publish the most recent ones to the website. That makes it hard for users to figure out what things are available in the Ember version they use and what is not available.

YUIDoc has a handy [`@since`](http://yui.github.io/yuidoc/syntax/index.html#since) tag for that purpose, that can be used to indicate in what release something was introduced.

https://github.com/emberjs/ember.js/search?q=since&type=Code shows that we are already using this in some places, but 88 occurrences is obviously not covering all of our API.

- should we use this more often? I assume the answer to this is: yes

- how can we figure out in what version something was introduced? git blame is one option, but it still takes quite a bit of effort to follow all the locations a method or property might have been moved. is there a better way?

- should we require future PRs that add functionality to include a `@since` tag?

/cc @locks

pixelhandler:
@Turbo87 It seems that the release process would have to support another step, for any feature flags that are removed someone would need to add the `@since` labels in a commit prior to release.

locks:
@pixelhandler I'm not quite following your argument, why would a feature flag have a @since?

pixelhandler:
@locks new features are behind feature flags so when those are removed we know that feature is landing in a specific release. That would be the time to add the since label. (going forward) As for existing API We only have the blog posts for the releases so those announcements may be the way to identify what version parts of the API have landed in the release channel.

locks:
Superseded by https://github.com/emberjs/ember.js/issues/15062.

PR #14700, opened 2016-12-09
[DOC release] Add @since tags to Ember.deprecate(), .warn() and .assert()

Turbo87:
Resolves another small amount of #14692

/cc @locks


locks:
@homu r+

PR #14701, opened 2016-12-09
[BUG beta] blueprints/acceptance-test/mocha: Return last async helper from test function

Turbo87:
This is recommended by `ember-mocha`, now that https://github.com/emberjs/ember-mocha/pull/121 is merged. Compared to the originally proposed solution using a wrapper function, this solution is backwards-compatible and will also work with older `ember-mocha` releases.

/cc @rwjblue


rwjblue (approved):
LGTM, I'll land once green

Turbo87:
@rwjblue looks like this was also affected by https://github.com/nodejs/build/issues/562

PR #14707, opened 2016-12-10
improved style deprecation message

IgorKvasn:
Second attempt to accomplish #13728

### DIFF
--- a/packages/ember-views/lib/system/utils.js
+++ b/packages/ember-views/lib/system/utils.js
@@ -13,11 +13,14 @@ export function isSimpleClick(event) {
   return !modifier && !secondaryClick;
 }
 
-export const STYLE_WARNING = '' +
-  'Binding style attributes may introduce cross-site scripting vulnerabilities; ' +
-  'please ensure that values being bound are properly escaped. For more information, ' +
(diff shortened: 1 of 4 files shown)


packages/ember-views/lib/system/utils.js
```diff
@@ -13,11 +13,14 @@ export function isSimpleClick(event) {
   return !modifier && !secondaryClick;
 }
 
-export const STYLE_WARNING = '' +
-  'Binding style attributes may introduce cross-site scripting vulnerabilities; ' +
-  'please ensure that values being bound are properly escaped. For more information, ' +
-  'including how to disable this warning, see ' +
-  'http://emberjs.com/deprecations/v1.x/#toc_binding-style-attributes.';
+export function constructStyleDeprecationMessage(affectedStyle) {
```
IgorKvasn: I wanted to use template string as well, but multiline template string produces multiline result - e.g.: 
```
`Hello
world`
```
produces:
```
Hello 
world
```

instead of just `Hello world`.
I guess this is not the "right" place to use template string - I would have to write the whole deprecation message as single-line which I guess would make the code a little bit messy :(

PR #14710, opened 2016-12-11
[FIXES #14405] deprecation of enumerable events:

### LINKED ISSUE #14405: explore/consider removal/deprecation of enumerable events (and hasArrayObservers/hasEnumerableObservers)
`@enumerable:change` et al

stefanpenner:
deprecate:

- [x] addEnumerableObserver
- [x] removeEnumerableObserver
- [x] hasEnumerableObserver
- [ ] hasArrayObserver

Remaining tasks:
- [ ] come up with an implementation to deprecate without causing lots of spew.
  this is tricky, because we both depend on this method, and invoke it internally (as sometimes our way of enabling the public API)

- [ ] deprecate all pieces
- [ ] make tests green
- [ ] ???

Questions:

- [ ] deprecate `until` when?
- [ ] we may also want to deprecate other mechanisms of subscribing to `enumerable:change`
- [ ] do we create a internal methods (via private symbol) that enables un-deprecated internal usage?

### DIFF
--- a/packages/ember-runtime/lib/mixins/enumerable.js
+++ b/packages/ember-runtime/lib/mixins/enumerable.js
@@ -92,7 +92,7 @@ function iter(key, value) {
   @since Ember 0.9
   @private
 */
-const Enumerable = Mixin.create({
+export default Mixin.create({
 
   /**
     __Required.__ You must implement this method to apply this mixin.
@@ -872,6 +872,12 @@ const Enumerable = Mixin.create({
     @private
   */
   addEnumerableObserver(target, opts) {
+    deprecate(
+      'Usage of `Enumerable.prototype.addEnumerableObserver` is deprecated.',
+      false,
+      { until: '2.12.0', id: 'ember-metal.enumerable.addEnumerableObserver' }
+    );
+
     let willChange = (opts && opts.willChange) || 'enumerableWillChange';
     let didChange  = (opts && opts.didChange) || 'enumerableDidChange';
     let hasObservers = get(this, 'hasEnumerableObservers');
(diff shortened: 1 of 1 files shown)


packages/ember-runtime/lib/mixins/enumerable.js
```diff
@@ -872,6 +872,12 @@ const Enumerable = Mixin.create({
     @private
   */
   addEnumerableObserver(target, opts) {
+    deprecate(
+      'Usage of `Enumerable.prototype.addEnumerableObserver` is deprecated.',
+      false,
+      { until: '2.12.0', id: 'ember-metal.enumerable.addEnumerableObserver' }
```
stefanpenner: @mixonic / @chancancode what is the appropriate `until` value here?

stefanpenner:
@rwjblue may want to chat with you about this one sometime this week. Maybe tue or wed?

packages/ember-runtime/lib/mixins/enumerable.js
chancancode: I can never remember; does 2.12 mean 2.12 is the last version where this continues to work? I'll have to update my patch accordingly.

chancancode:
Is this also deprecating the hasArrayObserver property, or do we need to track that separately?

stefanpenner:
@chancancode oops, i can add that.

chancancode:
>  come up with an implementation to deprecate without causing lots of spew. this is tricky, because we both depend on this method, and invoke it internally (as sometimes our way of enabling the public API)

@stefanpenner not sure if it applies here, but in the past we have done something like this:

```javascript
export actuallyDoFoo(instance, args) {
  // ... no deprecation... use this for internal uses
}

export default SomeClass.extends({
  doFoo(args) {
    deprecate(...);
    return actuallyDoFoo(this, args);
  }
});
```

stefanpenner:
@chancancode ya, that partially works. We will need to adapt some more as `actuallyDoFoo` if defined is a public API that we must invoke (and deprecate).

Basically we need to add another level of indirection to correctly deprecate only users use of this, but not our own internal use of it.

```
export actuallyDoFoo(instance, args) {
  // ... no deprecation... use this for internal uses
}

export default SomeClass.extends({
  doFoo(args) {
    deprecate(...);
    if (this.actuallyDoFoo) { // deprecate } else { actuallyDoFoo(this, args); }
  }
});
```

krisselden:
where are these used internally that would cause spew?

mmun:
Enumerable events were removed in https://github.com/emberjs/ember.js/pull/16116.

PR #14728, opened 2016-12-16
[DOC] Change Ember.computed.equal example to use number instead of string

schonstal:
@locks 

Twiddle: https://ember-twiddle.com/02e7c7529165da14a91378cc2da1eabf?openFiles=controllers.application.js%2C


schonstal:
The original documentation uses a string as the example value, so it looks like it should work for equality of two properties, since strings as the first arguments to computed macros usually means properties. It's clearer that it expects a value if you use a different type.

locks:
Thank you! I agree with the justification, and the change seems a net positive.

PR #14729, opened 2016-12-16
Templates cannot read a property from a function

kturney:
We ran into this when trying to upgrade our app that uses https://github.com/LocusEnergy/ember-d3-helpers to 2.10.

Specifically, we were trying to get the `bandwidth` function property off a band scale, which comes from D3 and is itself a function (`options.xScale.bandwidth`).

`Ember.get(options, 'xScale.bandwidth')` works just fine.

I've included a failing test case.

### DIFF
--- a/packages/ember-glimmer/tests/integration/content-test.js
+++ b/packages/ember-glimmer/tests/integration/content-test.js
@@ -482,6 +482,30 @@ class DynamicContentTest extends RenderingTest {
     this.assertContent('hello');
     this.assertInvariants();
   }
+
+  ['@test it can render a property on a function']() {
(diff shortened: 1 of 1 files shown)


krisselden:
@kturney while props on functions not rendering is a regression being able to update them via Ember.set or Ember.notifyPropertyChange is a new feature.  I checked 1.12, 1.13, 2.9 and function props are rendered as though they are constant `{{const fn.prop}}`.

kturney:
superseded by https://github.com/emberjs/ember.js/pull/15482 and  https://github.com/emberjs/ember.js/pull/15535

Issue #14704, opened 2016-12-10
Ember 2.10.0 - Component isVisible property broke inline styles


3AHAT0P:
After upgrading to Ember 2.10.0, behaviour of our components has been changed.
Now after of every change isVisible property (except null->true) all inline styles of component were cleared.
This was not happening in Ember 2.9.0.

**UP:** Also, the problem is clearly visible when using the standart helper - {{textarea}}. If resize it and change isVisible -> size of textarea will be back to default.

You can see this happening in this [ember-twiddle.](https://ember-twiddle.com/82071c2cb3c534a031030da0857476b2)

locks:
Can you confirm this isn't a duplicate of https://github.com/emberjs/ember.js/issues/14683 ?

3AHAT0P:
@locks This issue isn't a duplicate of #14683 .

workmanw:
@3AHAT0P I know this issue is probably far more complex than just changing a style property, but I figured I'd point out that if you use set `attributeBindings` and use a CP it does seem to be a viable workaround. I modified your twiddle: [ember-twiddle](https://ember-twiddle.com/8bb7a675bb43803be3d0671c4f634987?openFiles=components.test-two.js%2C). 

Anyways, just figured I'd point it out in the unlikely event that you weren't aware. This is definitely still an issue though.

3AHAT0P:
@workmanw 
Hmm...  
I simple improve this workaround, and it work. But I'm not sure that this solution will not lead to any new problems.
Also code has become more difficult.
[ember-twiddle.](https://ember-twiddle.com/82838dbf90847f89eebc2c62ba46dc8f?openFiles=components.test-two.js%2Ctemplates.components.test-three.hbs)

Also, the problem is clearly visible when using the standart helper - {{textarea}}. If you resize it and change isVisible -> size of textarea will be back to default.

GavinJoyce:
I've created a failing test here: https://github.com/emberjs/ember.js/pull/14731

It's failing as the [`StyleBindingReference`](https://github.com/emberjs/ember.js/blob/bb46126d07f79edcde330be45bade5ad17fd400c/packages/ember-glimmer/lib/utils/bindings.js#L94-L116) doesn't update when the style is changed directly using the DOM (or when resizing a textarea by dragging it). I'm not certain, but I don't believe the direct DOM manipulations like these are intended to be supported.

[...]

PR #14731, opened 2016-12-17
failing test for #14704

GavinJoyce:
I'm not sure that this is something that we intend to support, but here is a failing test for https://github.com/emberjs/ember.js/issues/14704

It's failing as the [`StyleBindingReference`](https://github.com/emberjs/ember.js/blob/bb46126d07f79edcde330be45bade5ad17fd400c/packages/ember-glimmer/lib/utils/bindings.js#L94-L116) doesn't update when the style is changed directly using the DOM (or when resizing a textarea by dragging it). I'm not certain, but I don't believe the direct DOM manipulations like these are intended to be supported.


GavinJoyce:
screen shot 2016-12-17 at 20 19 28


closing

PR #14733, opened 2016-12-18
[BUGFIX beta] Make `lazyGet` check cache correctly

### LINKED ISSUE #14686: lazyGet doesn't check cache correctly
```js
    let cache = meta.readableCache();
    if (cache && key in cache) {
      return cache[key];
    }
```

this should use `cacheFor.get`  which checks for the `UNDEFINED` value or `undefined` for not present.  The above will be true in cases where it isn't actually cached, since cache is cleared with = undefined.  Though returning undefined from here has the same effect in that it does not descend further in wiring the chain.

pittst3r:
Fixes #14686

### DIFF
--- a/packages/ember-metal/tests/chains_test.js
+++ b/packages/ember-metal/tests/chains_test.js
@@ -1,6 +1,6 @@
 import { addObserver } from '../observer';
 import { get } from '../property_get';
-import { finishChains } from '../chains';
+import { ChainNode, finishChains } from '../chains';
 import { defineProperty } from '../properties';
 import computed from '../computed';
 import { propertyDidChange } from '../property_events';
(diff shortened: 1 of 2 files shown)


packages/ember-metal/tests/chains_test.js
```diff
@@ -76,3 +76,14 @@ QUnit.test('observer and CP chains', function() {
   get(obj, 'qux'); // CP chain re-recreated
   ok(true, 'no crash');
 });
+
+QUnit.test('checks cache correctly', function(assert) {
+  let obj = { foo: {} };
+  let parentChainNode = new ChainNode(null, null, obj);
+  let chainNode = new ChainNode(parentChainNode, 'foo');
+
+  defineProperty(obj, 'foo', computed(function() { return UNDEFINED; }));
```
krisselden: This should return undefined, when it is `undefined` the cache will have `UNDEFINED` and the chain will descend when it should stop.

krisselden (changes requested):
The test needs to be updated so it reflects a realistic scenario.  `UNDEFINED` isn't going to be the direct value of a CP.

robbiepitts:
@krisselden Just pushed your requested changes.

krisselden (approved):

PR #14736, opened 2016-12-19
[FEATURE ember-factory-for] Remove duplicated code.

rwjblue:
This functionality was moved into `ContainerProxyMixin`, and doesn't need to be duplicated here.

### DIFF
--- a/packages/ember-application/lib/system/application-instance.js
+++ b/packages/ember-application/lib/system/application-instance.js
@@ -4,7 +4,7 @@
 */
 
 import { assign } from 'ember-utils';
-import { deprecate, get, set, run, computed, isFeatureEnabled } from 'ember-metal';
+import { deprecate, get, set, run, computed } from 'ember-metal';
 import {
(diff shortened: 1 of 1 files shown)


packages/ember-application/lib/system/application-instance.js
```diff
@@ -279,25 +279,9 @@ const ApplicationInstance = EngineInstance.extend({
...
-  [FACTORY_FOR](fullName, options) {
-    return this.__container__[FACTORY_FOR](fullName, options);
-  },
-
-  [LOOKUP_FACTORY](fullName, options) {
-    return this.__container__[LOOKUP_FACTORY](fullName, options);
   }
 });
 
-if (isFeatureEnabled('ember-factory-for')) {
```
chadhietala: The `isFeatureEnabled` import needs to be removed.

rwjblue: Ahh, thank you! Will update

Issue #14741, opened 2016-12-20
`makeArray` does not work as commented


pswai:
In `make-array`, it is commented as followed:

```javascript
 let controller = Ember.ArrayProxy.create({ content: [] });

 Ember.makeArray(controller) === controller;  // true
```

However it is not true. It seems `Ember.makeArray(controller)` is using `Array.isArray` now, which does not handle array-like object.

rwjblue:
The documentation should be updated to reflect that an array proxy has no special behavior here.

PR #14742, opened 2016-12-20
Add test to verify `makeArray` bug

pswai:
This adds tests to https://github.com/emberjs/ember.js/issues/14741

### DIFF
--- a/packages/ember-utils/tests/make-array_test.js
+++ b/packages/ember-utils/tests/make-array_test.js
@@ -1,0 +1,32 @@
+import { makeArray } from '../index';
+import { ArrayProxy } from 'ember-runtime';
+
+QUnit.module('ember-utils makeArray');
+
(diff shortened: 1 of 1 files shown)


mixonic:
@pswai thanks for looking into this. I believe this utility is private, and I expect we're simply not concerned with the array proxy case as they are rarely (never?) used in Ember internals.

Can you confirm you're making a comment about the private `makeArray` API here? Or perhaps I'm missing where this is impacting public API.

pswai:
@mixonic I was looking at this file ember.js/packages/ember-utils/lib/make-array.js. The block comment there gives example of `ArrayProxy` so I am not sure whether it is due to broken `makeArray` or the comment needs to be updated

rwjblue:
Sorry for the confusion, yes we should definitely fix the docs here...

rwjblue:
Closing this in favor of a future documentation specific PR...

PR #14750, opened 2016-12-22
assert that a component `tagName` can not be a computed property

### LINKED ISSUE #14749: Components with a computed tagName throw an error when trying to render
Having a computed property for `tagName` seems to break in recent Ember versions (2.8 and later).  Something like: 

```
  isButton: true,

  tagName: Ember.computed('isButton', function () {
    return this.get('isButton') ? 'button' : 'a';
  }),
});
```

throws:

image

Demonstrated by this ember-twiddle:

https://ember-twiddle.com/d54856b5f03e13bd2ccb66f07efc093b?openFiles=templates.application.hbs%2Ctemplates.components.my-component.hbs

Digging in, it looks like the `ComputedProperty` itself is being passed to the render instead of the computed string value for the tag name.

GavinJoyce:
fixes https://github.com/emberjs/ember.js/issues/14749

```js
Component.extend({
  tagName: computed(function() {
    return 'foo-bar';
  })
});
```

Now produces an assertion:

```You cannot use a computed property for the component's `tagName` (<(subclass of Ember.Component):ember445>)```

### DIFF
--- a/packages/ember-glimmer/lib/component.js
+++ b/packages/ember-glimmer/lib/component.js
@@ -167,6 +167,8 @@ const Component = CoreView.extend(
           }
         }
       )());
+
+      assert(`You cannot use a computed property for the component's \`tagName\` (${this}).`, !(this.tagName && this.tagName.isDescriptor));
     },
 
     rerender() {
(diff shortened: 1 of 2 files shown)


locks:
@GavinJoyce do you mean "for the component's `tagName`"? Otherwise it sounds like the component is named `tagName` to me.

GavinJoyce:
@locks yes, thanks

rwjblue:
Thanks @GavinJoyce!

Issue #14727, opened 2016-12-16
TextField component file type setting in Safari 10.0


mykehsd:
Ember version > 2.10
Safari version tested: 10.0.1
Reproducible:  Every time ->  https://ember-twiddle.com/aa37dcf277451f5954d6ca2d17e16dc6?openFiles=templates.application.hbs%2C
![Safari 10.0.1](https://cloud.githubusercontent.com/assets/195844/21268773/eef2c052-c364-11e6-905b-deb4406b7ffb.png)

Issue: When rendering multiple textfield components on a page, only the first type will be allowed on the page
.  The set of the test element .type property successfully returns however the type actually doesn't alter the value of the property.

Problem line:  https://github.com/emberjs/ember.js/blob/7819409b7b6559e9fc0300a29e66a16356fb67a7/packages/ember-glimmer/lib/components/text_field.js#L32

Possible solution:  Lowering the scope of inputTypeTestElement to inside the method resolves the problem, but I assume it was scoped for performance reasons.

mykehsd:
After more investigation, the problem is isolated to file types.  Safari does not allow creating an input type and changing it to "file". 

```
input = document.createElement('input');
input.type = 'text';    // === 'text'
input.type = 'file';     // === 'text'
```

pixelhandler:
Seems like a bug, second input for `type="file"` doesn't work in Safari, but does in Chrome, Firefox and Brave. Also odd that If I switch the order of inputs the file inputs works and so does the search input, https://ember-twiddle.com/529223e54e9ce39150c3366001a42742?openFiles=templates.application.hbs%2C seem that the issue is more of a problem with the file input type

rwjblue:
![screenshot](https://monosnap.com/file/CkapHSAbmegKMyVFZT6iBSqKbBqk23.png)

cibernox:
FWIW, I also found this problem yesterday in `emberx-file-input`. I PR'ed a [workaround](https://github.com/thefrontside/emberx-file-input/pull/48) by ...not using `{{input}}` and using a plain `<input>` (I think there is no good reason to use `{{input}}` ever again since ember 2.3.1)

iezer:
[...]

PR #14753, opened 2016-12-24
[BUGFIX beta] File element type

anon4562:
You cannot create an element and change it's type from * to "file".  

See: #14727 for more details


stefanpenner:
@anon4562 sorry for letting this sit, can you add a test?

Serabe:
Rebased and added a test. The problem cannot be reproduced in Safari 10.1.2 though.

Serabe:
Something went really wrong with the tests. Into it.

Serabe:
My test was wrong. I was testing that an input element can change from any type to file type while the issue was testing the type is supported. I changed this to render two inputs, being the file one the second one.

Sorry!

Issue #14754, opened 2016-12-24
Deprecate `canDispatchToEventManager`?


GavinJoyce:
From https://github.com/emberjs/ember.js/blob/e24e403656b9de62d55b1a3111e4eb309276d4b8/packages/ember-views/lib/system/event_dispatcher.js#L105-L133:

```
Note that most Ember applications do not use this feature. If your app also
does not use it, consider setting this property to false to gain some performance
improvement by allowing the EventDispatcher to skip the search for the
`eventManager` on the view tree.
```

<img width="985" alt="screen shot 2016-12-24 at 22 11 05" src="https://cloud.githubusercontent.com/assets/2526/21468909/e3031ed2-ca25-11e6-934c-2e11ffeddf36.png">

Some questions:

 * Do many apps use this? If not, perhaps we can deprecate (with a RFC)
 * If we can't deprecate, can we disable it by default and make it opt-in?
 * What are the use cases for custom `eventManager`s? Can we suggest alternative solutions?

rwjblue:
> Do many apps use this?

I checked with @tomdale and apparently this was originally added to allow views to differentiate between certain types of touch input, however the addon that originally used this ([ember-touch](https://github.com/emberjs-addons/ember-touch/)) has been deprecated in favor of [Hammer.js](http://hammerjs.github.io/).

> If not, perhaps we can deprecate (with a RFC)

Yes, we should absolutely deprecate it.

> If we can't deprecate, can we disable it by default and make it opt-in?

I think we can make it more "pay-go" either way, so that most apps don't have to pay the cost of the feature they never use...

GavinJoyce:
TODOs:

 * [x] Make it pay-as-you-go : https://github.com/emberjs/ember.js/pull/14756 🍻 @rwjblue 
 * [ ] Create RFC to deprecate `canDispatchToCustomEventManager` @gavinjoyce
 * [ ] Deprecate `canDispatchToEventManager`

PR #14756, opened 2016-12-25
Disable costly eventManager support when unused.

rwjblue:
For every event that we observe (which is basically all bubbling events in the DOM) we would iterate the entire `parentView` structure of the target elements view looking for an `eventManager` (and then we would dispatch the event to the `eventManager` instead of the view itself).

This support has existed for a *very* long time, and is generally unused in most applications. Unfortunately, the iteration upwards through the view heirarchy is much more costly than we would like. This is currently being done for events like `mouseenter` and  `mousemove` and that very very very few applications (if any) actually take advantage of this support.

This changes the `EventDispatcher` to (by default) disable support for `eventManager`'s (and avoids the costly `parentView` iteration) until we actually instantiate a component that has an `eventManager` property.

In the future, we should deprecate specifying an `eventManager`, but this makes the feature much more "pay as you go".

Related to https://github.com/emberjs/ember.js/issues/14754

### DIFF
--- a/packages/ember-views/lib/mixins/view_support.js
+++ b/packages/ember-views/lib/mixins/view_support.js
@@ -1,4 +1,4 @@
-import { guidFor } from 'ember-utils';
+import { guidFor, getOwner } from 'ember-utils';
 import { assert, deprecate, descriptor, Mixin } from 'ember-metal';
 import { environment } from 'ember-environment';
 import { matches } from '../system/utils';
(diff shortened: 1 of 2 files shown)


rwjblue:
@GavinJoyce - Would you like to work on the deprecation RFC (we can use the same conditional to add the deprecation)?

GavinJoyce:
Nice one! 

> Would you like to work on the deprecation RFC (we can use the same conditional to add the deprecation)

Sure, I'll work on the RFC and deprecation

rwjblue:
FYI - We have a couple of tests that confirm `.eventManager` support ([here](https://github.com/emberjs/ember.js/blob/aeeddf7c00b42a340494465252481b5414d87009/packages/ember-glimmer/tests/integration/event-dispatcher-test.js#L39-L62) and [here](https://github.com/emberjs/ember.js/blob/aeeddf7c00b42a340494465252481b5414d87009/packages/ember-glimmer/tests/integration/event-dispatcher-test.js#L64-L96)), so in general the original feature still works (without app changes required).

jonnii:
@rwjblue I'd have thought this would be a huge performance win, do you have any idea how big?

rwjblue:
@jonnii - I agree it should have a nice perf win, but haven't really done any testing. Most of the benchmarks I've been running locally are specifically testing initial render or rerender performance, and those types of tests generally wouldn't benefit from this (AFAICT).  I think the types of things that would benefit would be rerenders where many events are firing.  I'd definitely be interested in some numbers if you happened to have a perf testing harness setup in this way...

ef4 (commented):
This looks fine to me.

Issue #14725, opened 2016-12-16
changed each-behaviour for indexed arrays in 2.10


Suven:
Hey!

I just noticed that Ember 2.10 introduces a changed behaviour when looping over indexed arrays with unset indexes.

Let's assume this array:

```js
let a = [];
a[5] = 'foo';
a[6] = 'bar';
```

and this hbs:

```hbs
{{#each testArray as |v k|}}{{k}},{{/each}}
```

In 2.9 this produces `5,6` and in 2.10 it produces `0,1,2,3,4,5,6`.

I also made a [twiddle](https://ember-twiddle.com/5e7d30c69358b5094ae644ddeaf9f947) for that

pixelhandler:
@Suven that seems like a bug. Perhaps the implementation has changed from using `forEach` to a `for` loop. (just a guess).

⚡ cat changed-each.js

```js
a = [];
a[5] = 'foo';
a[6] = 'bar';

for (i = 0; i < a.length; i++) {
  console.log(i, a[i]);
}

a.forEach(function(v, i) {
  console.log(i, v);
});
```

⚡ node changed-each.js

```
0 undefined
1 undefined
2 undefined
3 undefined
4 undefined
5 'foo'
6 'bar'
5 'foo'
6 'bar'
```

Suven:
Yup, that might be. I guess for-loops were tempting with perfomance having in mind first.

Just a little remark (although I guess that's obvious): This also happens when only the value is accessed (i.e `each myArray as |val|`)

kamilogorek:
I can confirm that it's bugged. Also using `each-in` instead of `each` temporary solves this problem.

rwjblue:
The issue here is that (for performance reasons) we are using a standard `for` loop for `{{#each` support in 2.10+. Prior versions used `.forEach` and that natively support sparse arrays (and that is one of the reasons for some known performance issues).

We did it intend to support sparse arrays in prior versions and making changes to 2.10+ to support them would negatively impact performance of _all_ usages of `{{#each` which is not an acceptable trade off for us.

We should document the fact that we do not support sparse arrays with `{{#each` in the API docs and add a test (likely a tweak of #14760) that shows the current behavior is intentional. 

For folks that definitely need this support, I'd recommend using `{{#each-in` as someone else mentioned above.

pixelhandler:
@Suven I labelled this as "documentation" per @rwjblue comments above 

[...]

PR #14760, opened 2016-12-27
{{#each}} and {{#each-in}} helpers with sparse array tests

kamilogorek:
Related issue https://github.com/emberjs/ember.js/issues/14725

### DIFF
--- a/packages/ember-glimmer/tests/integration/syntax/each-in-test.js
+++ b/packages/ember-glimmer/tests/integration/syntax/each-in-test.js
@@ -443,6 +443,21 @@ moduleFor('Syntax test: {{#each-in}}', class extends BasicEachInTest {
     this.assertText('[0:1][1:2][2:3][foo:bar]');
   }
 
+  ['@test it skips holes in sparse arrays'](assert) {
+    let arr = [];
+    arr[5] = 'foo';
+    arr[6] = 'bar';
+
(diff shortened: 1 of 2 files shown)


rwjblue:
See associated comment on #14725, can you tweak this to show the current 2.10 behavior as the expected result? Also, it might be good to explicitly add a test showing the work around (using `{{#each-in`) supports those needing sparse array support.

kamilogorek:
Definitely. Is current test in `each-in` enough for

>  explicitly add a test showing the work around (using {{#each-in)

or do you want me to move it to `each` test suit?

rwjblue:
Oops, sorry, yes the existing test is good for each-in. Thank you!

rwjblue:
Yes, we intentially fall back to forEach when Array.isArray is false, this is intentional (e.g. To support things like linked lists or other objects that are iterable (in the future we will likely support `Symbol.iterBle`)).

kamilogorek:
@rwjblue could you point me to a solution for this difference? or should I just move the test underneath `// TODO: Refactor the following tests so we can run them against different kind of arrays` and specify what types of arrays we want to test?

kamilogorek:
Updated both tests. We're back to green now.

rwjblue (approved):

rwjblue:
Thank you!

PR #14763, opened 2016-12-28
Add unit tests for `ember-glimmer/utils/iterable`

pittst3r:
While investigating #14725 I noticed there were no unit tests for this file, so I added them. Looks like the issue has since been resolved though.


locks (approved):

rwjblue:
Awesome, thank you!

chancancode:
@robbiepitts we are in the process of moving Ember to Glimmer VM master and these tests are failing because they rely on internals that doesn't exists anymore (`EvaluatedPositionalArgs`, specifically). Can you investigate porting the relevant tests to integration style? Or can we just delete them?

robbiepitts:
@chancancode I don't like deleting tests without good reason (I don't see one in this case?). I will investigate porting them.

rwjblue:
For now I have removed the `iterableForObject` tests (in https://github.com/emberjs/ember.js/pull/15245). Happy to bring them back when we have time though...

PR #14764, opened 2016-12-29
Fixed string capitalize for accented characters

IgorKvasn:
String capitalize function did not work properly, when first character is  accented (e.g. ščťžý)

### DIFF
--- a/packages/ember-runtime/lib/system/string.js
+++ b/packages/ember-runtime/lib/system/string.js
@@ -60,7 +60,7 @@ const UNDERSCORE_CACHE = new Cache(1000, function(str) {
     replace(STRING_UNDERSCORE_REGEXP_2, '_').toLowerCase();
 });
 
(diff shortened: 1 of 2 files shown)


packages/ember-runtime/lib/system/string.js
```diff
@@ -60,7 +60,7 @@ const UNDERSCORE_CACHE = new Cache(1000, function(str) {
     replace(STRING_UNDERSCORE_REGEXP_2, '_').toLowerCase();
 });
 
-const STRING_CAPITALIZE_REGEXP = (/(^|\/)([a-z])/g);
+const STRING_CAPITALIZE_REGEXP = (/(^|\/)([a-zA-Z\u00C0-\u024F])/g);
```
rwjblue: I'm not sure I understand why `A-Z` was added here, can you explain?

IgorKvasn: you're right :) I'll remove A-Z from the regex

locks (approved):

Issue #14548, opened 2016-10-29
[DOC] Ember.Component#layout is not documented as public


mike-north:
Inspired by: https://github.com/ember-cli/ember-cli.github.io/issues/100

Currently the [default addon component blueprint](https://github.com/emberjs/ember.js/blob/v2.10.0-beta.2/blueprints/component/index.js#L59-L68) sets up the component template via the `layout` property. 

There is no evidence in the [API docs](http://emberjs.com/api/classes/Ember.Component.html) that it's public, or any mention or use of it in [the appropriate section of the guides](https://guides.emberjs.com/v2.9.0/components/defining-a-component/), leading to confusion as to whether it's ok to use. Add in [the deprecated `Ember.View` concept of the same name](http://emberjs.com/api/classes/Ember.View.html#toc_layouts), and it's understandable why people new to the framework are tripping up here.

rwjblue:
Looks like a bunch of docs from `Ember.View` haven't been moved over to `Ember.Component` yet.  

Here is a small list (there are likely more things though):

* https://github.com/emberjs/ember.js/blob/v1.13.0/packages/ember-views/lib/views/view.js#L753-L767
* https://github.com/emberjs/ember.js/blob/v1.13.0/packages/ember-views/lib/views/view.js#L714-L725
* https://github.com/emberjs/ember.js/blob/v1.13.0/packages/ember-views/lib/views/view.js#L909-L921
* https://github.com/emberjs/ember.js/blob/v1.13.0/packages/ember-views/lib/views/view.js#L1093-L1119
* https://github.com/emberjs/ember.js/blob/v1.13.0/packages/ember-views/lib/views/view.js#L60-L673

Also, in reviewing here it seems that `layout` was indeed private in Ember 1.13.  We should probably fix that (or change the blueprint)...

sclatter:
@rwjblue I'm going to take a crack at this. Just to clarify - do we want to document tagName API on Component also? Or, just classNames and classNameBindings?

sclatter:
PR submitted: https://github.com/emberjs/ember.js/pull/14765

PR #14765, opened 2016-12-29
[DOC release-2-9] Move API documentation to Ember.Component

sclatter:
Fixes https://github.com/emberjs/ember.js/issues/14548.

Reconciling documentation from Ember.View to Ember.Component, noting layout as public. All tests pass. 

```
QUnit 1.23.1; Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/55.0.2883.95 Safari/537.36
Tests completed in 271116 milliseconds.
30312 assertions of 30312 passed, 0 failed.
skippedEmber.Application - visit() Integration Tests: SKIPPED: Test setup (0)Rerun
skippedEmber.Application - visit() Integration Tests: SKIPPED: iframe setup (0)Rerun
skippedHelpers test: {{unbound}}: SKIPPED: helper form updates on parent re-render (0)Rerun
skippedEmberObject ES Compatibility: SKIPPED: calling extend on an ES subclass of EmberObject (0)Rerun
```


rwjblue:
This looks good to me (mostly moving things from the older `Ember.View` documentation locations to `Ember.Component`), but I suspect there will need to be some language tweaks (because these are using older globals style and may refer to some concepts that we don't really support/advise).

@locks - Can you review and advise on that kind of thing?

locks:
I'll review shortly! I agree that since we're touching this documentation we should update and review to make sure it's clear and current.
@sclatter how do you prefer going through this? I was thinking of doing a pass, commenting inline, and then we could discuss any particular point. You can also reach me or the learning team on our Slack, at `#-team-learning`.

sclatter:
@locks - that sounds like a good plan - doing one pass and then we can discuss any points. Sorry I dropped the PR and then disappeared on a very long bike ride. I will be more consistently online tomorrow. In addition to making sure examples follow latest conventions, I'm wondering if this is the right place to discuss any anti-patterns? Perhaps the guide might be a better place. I plan to hit that next to incorporate the layouts piece under the Component guides. I think we should discuss somewhere avoiding deeply nested components, aim for components as leaf nodes in the system, and preference for composition over inheritance, etc. I will take the time to make the doc great, but would love your guidance on where/what to discuss.

sclatter:
@locks, @rwjblue I went ahead and removed a few of the globals and made generic component naming more consistent in the examples.

packages/ember-glimmer/lib/component.js
```diff
@@ -85,7 +85,7 @@ export const BOUNDS = symbol('BOUNDS');
   `hello` for the `app-profile` component:
 
   ```javascript
-  App.AppProfileComponent = Ember.Component.extend({
```
locks: API docs support paths, so you can do something like the following:

    ```app/components/profile.hbs
    import Ember from 'ember';
  
    …
    });
    ```

The idea is to provide as full a file as possible to try to make it less confusing. The paths should use the classic structure, not pods, and imports should be added. This last point will hopefully help users if the JavaScript module API RFC gets greenlit.

packages/ember-glimmer/lib/component.js
```diff
@@ -109,13 +109,384 @@ export const BOUNDS = symbol('BOUNDS');
   elements. This is consistent with the same
   requirement in web components.
 
+    ## HTML Tag
```
locks: The indentation seems off here.

packages/ember-glimmer/lib/component.js
```diff
@@ -109,13 +109,384 @@ export const BOUNDS = symbol('BOUNDS');
...
+  The default HTML tag name used for a components's DOM representation is `div`. This
+  can be customized by setting the `tagName` property. The following component
+  class:
+  ```javascript
```
locks: The documentation seems to be missing newlines in general, but around code samples specifically.

packages/ember-glimmer/lib/component.js
```diff
@@ -109,13 +109,384 @@ export const BOUNDS = symbol('BOUNDS');
...
+    tagName: 'form',
+    attributeBindings: ['novalidate'],
+    novalidate: null
+  });
+  ```
+  Updates to the property of an attribute binding will result in automatic
+  update of the  HTML attribute in the component's rendered HTML representation.
+  `attributeBindings` is a concatenated property. See [Ember.Object](/api/classes/Ember.Object.html)
+  documentation for more information about concatenated properties.
+  ## Layouts
```
locks: Missing newlines for vertical spacing here.

locks (changes requested):
Some general changes

locks:
> @locks - that sounds like a good plan - doing one pass and then we can discuss any points. Sorry I dropped the PR and then disappeared on a very long bike ride. I will be more consistently online tomorrow.

No worries, getting a nice PR to review is already a treat 😁  I have left some comments inline!

> In addition to making sure examples follow latest conventions, I'm wondering if this is the right place to discuss any anti-patterns?

You are correct that the Guides are a better place to address anti-patterns than the API documentation, which I feel should be focused on the capabilities of the APIs. However, the Guides are also based on the "green path", and I'm not sure we currently have material in that direction, so it will need some additional discussing with the learning team. It's always a tricky balance being nuanced enough in that kind of material and not get too bogged down in the details. I'll put it in the learning team meeting agenda, you're welcome to attend!

packages/ember-glimmer/lib/component.js
```diff
@@ -109,13 +109,384 @@ export const BOUNDS = symbol('BOUNDS');
...
+  ```
+  Will result in component instances with an HTML representation of:
+  ```html
+  <div id="ember1" class="ember-view my-class my-other-class"></div>
+  ```
+  `class` attribute values can also be set by providing a `classNameBindings`
+  property set to an array of properties names for the component. The return value
+  of these properties will be added as part of the value for the components's `class`
+  attribute. These properties can be computed properties:
+  ```javascript
```
locks: For these "generic" components, I think we can keep the name and do:
   
    ```app/components/my-component.js

packages/ember-glimmer/lib/component.js
sclatter: fixed

packages/ember-glimmer/lib/component.js
sclatter: Added 1 line before and after code samples

packages/ember-glimmer/lib/component.js
sclatter: Updated to do consistent vertical spacing throughout - 2 lines before sections - 1 line everywhere else.

packages/ember-glimmer/lib/component.js
sclatter: Used this convention throughout when it made sense to reference a file name.

sclatter:
@locks - I just noticed your request to remove outdated docs. Also did that in addition to addressing your comments on the new docs. Commit no.5 is the one addressing comments and no.6 is deleting old docs. 

Looking forward to hopefully joining the next Learning meeting. Please send details when available. I'm also happy to pick up additional docs tasks in the future.

packages/ember-glimmer/lib/component.js
```diff
@@ -41,16 +41,16 @@ export const BOUNDS = symbol('BOUNDS');
 
   The easiest way to create an `Ember.Component` is via
   a template. If you name a template
-  `components/my-foo`, you will be able to use
+  `app/components/my-foo`, you will be able to use
```
locks: Missing file extension here.

packages/ember-glimmer/lib/component.js
```diff
@@ -41,16 +41,16 @@ export const BOUNDS = symbol('BOUNDS');
...
-  ```handlebars
-  {{app-profile person=currentUser}}
+  ```app/components/my-foo.hbs
+  {{profile-component person=currentUser}}
   ```
 
-  ```handlebars
-  <!-- app-profile template -->
+  ```app/components/profile-component.hbs
+  <!-- profile-component template -->
```
locks: Left over comment.

packages/ember-glimmer/lib/component.js
```diff
@@ -109,13 +110,512 @@ export const BOUNDS = symbol('BOUNDS');
   elements. This is consistent with the same
   requirement in web components.
 
+  ## HTML Tag
+  The default HTML tag name used for a components's DOM representation is `div`. This
+  can be customized by setting the `tagName` property. The following component
+  class:
+
+
```
locks: Extra newline.

packages/ember-glimmer/lib/component.js
```diff
@@ -109,13 +110,512 @@ export const BOUNDS = symbol('BOUNDS');
...
+  ```javascript
```
locks: This and next line should be updated with file path.

packages/ember-glimmer/lib/component.js
```diff
@@ -109,13 +110,512 @@ export const BOUNDS = symbol('BOUNDS');
...
+  //app/components/paragraph-component.js
+
```
locks: Extra newline.

packages/ember-glimmer/lib/component.js
```diff
@@ -109,13 +110,512 @@ export const BOUNDS = symbol('BOUNDS');
...
+  import Ember from 'ember';
+
+  ParagraphComponent = Ember.Component.extend({
+    tagName: 'em'
+  });
+  ```
+  Would result in instances with the following HTML:
```
locks: Missing newline.

packages/ember-glimmer/lib/component.js
```diff
@@ -109,13 +110,512 @@ export const BOUNDS = symbol('BOUNDS');
...
+  can be customized by setting the `tagName` property. The following component
+  class:
+
+
+  ```javascript
+  //app/components/paragraph-component.js
+
+  import Ember from 'ember';
+
+  ParagraphComponent = Ember.Component.extend({
```
locks: There is no export in this module.

locks (changes requested):
Closer 😁  Some nitpicks:
- Make sure to add exports to the code samples
- No newline between code fence (triple backtick) and code
- Code fences should have newlines _around_ them to avoid rendering problems

A matter that I also want to raise is naming the component `profile-component`. While global-style should be the most minimal of use cases, I personally think it might be confusing to have a component with a `-component` suffix because in global-style that would make it `ProfileComponentComponent`. I also don't want to induce people into thinking components need to have that suffix. What do you think?

packages/ember-glimmer/lib/component.js
```diff
@@ -79,13 +78,16 @@ export const BOUNDS = symbol('BOUNDS');
...
-  `hello` for the `app-profile` component:
+  `hello` for the `profile` component:
 
-  ```javascript
-  App.AppProfileComponent = Ember.Component.extend({
+  ```app/components/profile-component.js
+
+  import Ember from 'ember';
+
+  ProfileComponent = Ember.Component.extend({
```
locks: As an example of a missing export, this could be rewritten `export default Ember.Component.extend({`.

locks:
@sclatter learning team meetings are Thursdays at 19h GMT. We use PureCloud for the meetings, @toddjordan usually posts registering instructions ahead of time in the #-learning channel of the community Slack.

sclatter:
@locks - I just made a pass at the latest set of comments. Tests all passing still. 

1. I made it so most examples did not have "component" in the file name. The only place I kept it was outer-component and inner-component, since I think the naming is helpful there. profile-component became person-profile and my-component became my-widget. Happy to put either back if not a helpful change. 

2. I moved all examples to use export. I noticed that in the guides when you use export, you do not use global-style, so I followed that convention. Let me know if you want global-style back in any of the examples. 

3. I think I cleaned up all the spacing problems. Let me know if I've missed anything - looked at it a few times, but it's hard to read code in comments :/

packages/ember-glimmer/lib/component.js
```diff
@@ -79,13 +77,15 @@ export const BOUNDS = symbol('BOUNDS');
   handle events or actions, you implement a subclass
   of `Ember.Component` named after the name of the
   component. Note that `Component` needs to be appended to the name of
```
locks: "Note that …" should be removed, IMO.

packages/ember-glimmer/lib/component.js
```diff
@@ -79,13 +77,15 @@ export const BOUNDS = symbol('BOUNDS');
...
-  your subclass like `AppProfileComponent`.
+  your subclass like `PersonProfileComponent`.
 
   For example, you could implement the action
-  `hello` for the `app-profile` component:
+  `hello` for the `person profile` component:
```
locks: Missing hyphen in component name, or was that on purpose?

packages/ember-glimmer/lib/component.js
```diff
@@ -109,13 +108,489 @@ export const BOUNDS = symbol('BOUNDS');
...
+  <div id="ember1" class="ember-view empty"></div>
+  ```
+  If you want to add a class name for a property which evaluates to true and
+  and a different class name if it evaluates to false, you can pass a binding
+  like this:
+
+  ```app/components/my-widget.js
+  import Ember from 'ember';
+
+  // Applies 'enabled' class when isEnabled is true and 'disabled' when isEnabled is false
```
locks: Do we need this comment if the explain in long-form right below?

packages/ember-glimmer/lib/component.js
```diff
@@ -109,13 +108,489 @@ export const BOUNDS = symbol('BOUNDS');
...
+    novalidate: null
+  });
+  ```
+  Updates to the property of an attribute binding will result in automatic
+  update of the  HTML attribute in the component's rendered HTML representation.
+  `attributeBindings` is a concatenated property. See [Ember.Object](/api/classes/Ember.Object.html)
+  documentation for more information about concatenated properties.
+
+
+  ## Layouts
```
locks: This section needs some rewording due to codebase changes. Approach is being discussed in Slack.

packages/ember-glimmer/lib/component.js
```diff
@@ -109,13 +108,489 @@ export const BOUNDS = symbol('BOUNDS');
...
+  See [Ember.Templates.helpers.yield](/api/classes/Ember.Templates.helpers.html#method_yield)
+  for more information.
+
+
+  ## Responding to Browser Events
+
+  Components can respond to user-initiated events in one of three ways: method
+  implementation, through an event manager, and through `{{action}}` helper use
+  in their template or layout.
+  ### Method Implementation
```
locks: Missing empty newlines.

packages/ember-glimmer/lib/component.js
```diff
@@ -109,13 +108,489 @@ export const BOUNDS = symbol('BOUNDS');
...
+
+  export default Ember.Component.extend({
+    click: function(event) {
+      // will be called when an instance's
+      // rendered element is clicked
+    }
+  });
+  ```
+
+  ### Event Managers
```
locks: Can you remove this section, but keep it handy somehow? We'll address it in a follow-up PR. The sets will be something like adding an `eventManager` property definition in `Ember.Component` and moving this documentation there.

packages/ember-glimmer/lib/component.js
```diff
@@ -401,6 +876,68 @@ const Component = CoreView.extend(
...
+      ```
+      If not manually set a default value will be provided by the framework.
+      Once rendered an element's `elementId` is considered immutable and you
+      should never change it. If you need to compute a dynamic value for the
+      `elementId`, you should do this when the component or element is being
+      instantiated:
+
+      ```javascript
+        export default Ember.Component.extend({
+          setElementId: function() {
```
locks: Using `.on` event handlers is less desired nowadays. Something like the following would be more up to date:

```javascript
init() {
  this._super(...arguments);

  var index = this.get('index');
  this.set('elementId', `component-id${index}`);
}
```

locks (changes requested):
Couple more nitpicks.

packages/ember-glimmer/lib/component.js
sclatter: Just got this in my last commit.

packages/ember-glimmer/lib/component.js
```diff
@@ -96,26 +95,453 @@ export const BOUNDS = symbol('BOUNDS');
...
+  ```html
+  <div id="ember1" class="ember-view disabled"></div>
+  ```
+
+  Updates to the value of a class name binding will result in automatic
+  update of the  HTML `class` attribute in the component's rendered HTML
+  representation. If the value becomes `false` or `undefined` the class name
+  will be removed.
+  Both `classNames` and `classNameBindings` are concatenated properties. See
+  [Ember.Object](/api/classes/Ember.Object.html) documentation for more
```
acorncom: @locks this thought can be left to another PR, but do we want to be using absolute references here or should we start shifting to relative references to work better with versioned api docs?

packages/ember-glimmer/lib/component.js
```diff
@@ -96,26 +95,453 @@ export const BOUNDS = symbol('BOUNDS');
...
+  Updates to the property of an attribute binding will result in automatic
+  update of the  HTML attribute in the component's rendered HTML representation.
+  `attributeBindings` is a concatenated property. See [Ember.Object](/api/classes/Ember.Object.html)
+  documentation for more information about concatenated properties.
+
+
+  ## Layouts
+
+  Components can have a secondary template that wraps their main template. Like
+  primary templates, layouts can be any function that  accepts an optional
```
acorncom: Two spaces before accepts

locks (approved):

packages/ember-glimmer/lib/component.js
locks: @acorncom we should figure out a way to automate to whatever we choose, I think.

PR #14768, opened 2016-12-30
[DOC release] Fix registerOptionsForType example

amiel:
I could be wrong, but I believe that `optionsForType` is deprecated in favor of `registerOptionsForType`. The documented example for `registerOptionsForType` showed using `optionsForType`. This updates the example to use `registerOptionsForType`.


rwjblue:
Yep, we missed updating the method being called when we moved this to the registry proxy mixin.  Good catch, thank you for fixing!

PR #14772, opened 2017-01-02
Test that Ember.K triggers a deprecation.

locks:


rwjblue:
Discussed with @locks in slack, we are going to make this deprecate on access (instead of only deprecate when invoked).

packages/ember/lib/index.js
```diff
@@ -212,19 +212,21 @@ Object.defineProperty(Ember, 'onerror', {
...
+      {
+        id: 'ember-metal.ember-k',
+        until: '3.0.0',
+        url: 'http://emberjs.com/deprecations/v2.x#toc_code-ember-k-code'
+      }
+    );
+
+    return function() { return this; };
+  }
+})
```
rwjblue: Looks like this is failing JSHint (missing `;` here)

packages/ember/lib/index.js
```diff
@@ -212,19 +212,21 @@ Object.defineProperty(Ember, 'onerror', {
...
+      'Ember.K is deprecated in favor of defining a function inline.',
+      false,
+      {
+        id: 'ember-metal.ember-k',
+        until: '3.0.0',
+        url: 'http://emberjs.com/deprecations/v2.x#toc_code-ember-k-code'
+      }
+    );
+
+    return function() { return this; };
```
rwjblue: I think we should probably keep this as a stable function (in case folks are doing `this.something === Ember.K`).

Small tweak to support ^ would be:

```js
function deprecatedEmberK() { return this; }
Object.defineProperty(Ember, 'K', {
  get() {
    deprecate(/* current stuff here */);
    return deprecatedEmberK;
  }
});
```

rwjblue (changes requested):

packages/ember/lib/index.js
rwjblue: @locks - Looks like this still needs to be addressed...

packages/ember-metal/tests/main_test.js
```diff
@@ -49,3 +49,9 @@ QUnit.test('Ember.Backburner is deprecated', function() {
     new Ember.Backburner(['foo']);
   }, 'Usage of Ember.Backburner is deprecated.');
 });
+
+QUnit.test('Ember.K is deprecated', function() {
+  expectDeprecation(function() {
+    Ember.K;
```
rwjblue: Hmm, looks like this is causing a JSHint failure now too:

> ember-metal/tests/main_test.js: line 55, col 11, Expected an assignment or function call and instead saw an expression.

packages/ember/lib/index.js
locks: @rwjblue 👍  missed that comment before, on it

packages/ember-metal/tests/main_test.js
```diff
@@ -49,3 +49,13 @@ QUnit.test('Ember.Backburner is deprecated', function() {
...
   }, 'Usage of Ember.Backburner is deprecated.');
 });
+
+QUnit.test('Ember.K is deprecated', function(assert) {
+  expectDeprecation(function() {
+    let obj = {
+      noop: Ember.K
+    };
+
+    assert.equal(obj, obj.noop());
```
locks: this is be being tricksy and avoiding the jshint for non-used variable :X

rwjblue (approved):

rwjblue:
Thanks @locks

PR #14775, opened 2017-01-03
[DOC release] Mark LinkComponent as public

locks:
As per discussion related to https://github.com/emberjs/rfcs/pull/176#issuecomment-267383628, LinkComponent should be changed to be a public class.


tomdale (approved):
I don't love this, but it's de facto public API at this point. Hopefully we can push people towards using the router service.

rwjblue (approved):

mitchlloyd:
I'm glad to see this was updated to be a public API since several codebases I've worked on rely on `data-*` attributes on links for integration with click tracking libraries. 👍 

I look forward to seeing alternative solutions that the router service provides!

PR #14778, opened 2017-01-03
Update text_field.js

locks:


rwjblue:
Why?

locks:
@rwjblue wanted Travis to tell me what would break because my machine can't handle the tests and opened the PR in the wrong repo :$

locks:
Out of curiosity, do you know why the try catch is there? Seems like something we should cover with a test.

rwjblue:
Some browsers throw when changing `.type` on an `<input>` element.

PR #14781, opened 2017-01-04
Integrate Glimmer Macros For Binary VM

chadhietala:
This moves Ember over to the new Binary VM in glimmer (Requires [this branch](https://github.com/tildeio/glimmer/commits/runtime-null-checks)). Notable changes:

- `{{debugger}}` is now implemented as a built-in Glimmer statement syntax due to the fact it needs access to the `SymbolTable` and other information from the VM. Also we now hard error if the debugger is passed an expression, from what I can tell this just failed silently is we did nothing with the positional or named params.
- `refineStatement` on the `Environment` is no more in favor of `populateMacros` and `block|inlines.addMissing`.
- No more syntax objects but now byte arrays. 😎

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -89,7 +89,7 @@ function rsvp() {
 function routeRecognizer() {
   var packageJson = require('route-recognizer/package');
   var packageDir = path.dirname(require.resolve('route-recognizer/package'));
-  var entry = path.join(packageDir, packageJson['module'] || packageJson['js:next'] || packageJson['main'].replace(/dist\//, 'dist/es6/'));
+  var entry = path.join(packageDir, packageJson['module'] || packageJson['jsnext:main'] || packageJson['main'].replace(/dist\//, 'dist/es6/'));
   var basename = path.basename(entry);
   var es6 = new Funnel(path.dirname(entry), {
     files: [ basename ]
@@ -116,7 +116,7 @@ function buildPackage(name, options) {
     throw new Error('If resolving from a non-package.json entry point, you must supply the srcDirectory.');
   }
 
-  var entryModule = packageJson['module'] || packageJson['js:next'] || packageJson['main'].replace(/dist\//, 'dist/es6/');
+  var entryModule = packageJson['module'] || packageJson['jsnext:main'] || packageJson['main'].replace(/dist\//, 'dist/es6/');
   var funnelDir = path.join(packageDir, options.entry ? options.srcDir : path.dirname(entryModule));
   var sourceEntry = options.entry ? options.entry : path.basename(entryModule);
 
@@ -300,7 +300,7 @@ module.exports = function() {
     'router':                router(),
     'dag-map':               dag(),
     'route-recognizer':      routeRecognizer(),
-    'simple-html-tokenizer': htmlbarsPackage('simple-html-tokenizer', { libPath: 'node_modules/glimmer-engine/dist/es6' }),
+    'simple-html-tokenizer': buildPackage('simple-html-tokenizer'),
 
     '@glimmer/compiler':     buildPackage('@glimmer/compiler', {
                                external: ['@glimmer/syntax', '@glimmer/wire-format', '@glimmer/util']
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -107,6 +107,46 @@ function routeRecognizer() {
   });
 }
 
+function buildPackage(name, options) {
+  options = options ? options : {};
+  var packageJson = require(name + '/package');
+  var packageDir = path.dirname(require.resolve(name + '/package'));
+
+  if (options.entry && !options.srcDir) {
+    throw new Error('If resolving from a non-package.json entry point, you must supply the srcDirectory.');
+  }
+
+  var entryModule = packageJson['module'] || packageJson['js:next'] || packageJson['main'].replace(/dist\//, 'dist/es6/');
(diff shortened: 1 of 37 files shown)


packages/ember-glimmer/lib/syntax/dynamic-component.js
```diff
@@ -16,37 +12,33 @@ function dynamicComponentFor(vm, symbolTable) {
...
+  let definitionArgs = [params.slice(0, 1), null, null, null];
+  let args = [params.slice(1), hash, _default, inverse];
+  builder.component.dynamic(definitionArgs, dynamicComponentFor, args, builder.symbolTable);
+  return true;
+}
 
-  compile(builder) {
-    builder.component.dynamic(this.definitionArgs, this.definition, this.args, this.symbolTable, this.shadow);
-  }
+export function inlineComponentMacro(path, params, hash, builder) {
```
chadhietala: This likely can go away in favor of `dynamicComponentMacro`.

ember-cli-build.js
```diff
@@ -197,26 +221,26 @@ function babelConfigFor(environment) {
...
-  }), {
-    files: ['**/*.js'],
-    pattern: {
-      match: /\/\/#\s+sourceMappingURL.*/g,
-      replacement: ''
-    },
-    annotation: 'strip sourceMappingURL'
-  });
-}
+// var glimmerEngine = require('glimmer-engine/ember-cli-build')({
```
rwjblue: We can probably just delete it right?

packages/ember-glimmer/lib/helpers/component.js
```diff
@@ -206,6 +205,11 @@ function createCurriedDefinition(definition, args) {
   );
 }
 
+const EMPTY_BLOCKS = {
```
rwjblue: Should this be frozen?

packages/ember-glimmer/lib/syntax.js
```diff
@@ -1,44 +1,85 @@
...
+  if (path.length > 1) {
+    return closureComponentMacro(path, params, hash, null, null, builder);
+  }
 
-export function findSyntaxBuilder(key) {
-  let index = syntaxKeys.indexOf(key);
+  let { symbolTable } = builder;
+
+  let definition;
+  if (name.indexOf('-') > -1) {
```
rwjblue: Should we use the contains dash caching here?

packages/ember-glimmer/lib/syntax/-in-element.js
```diff
@@ -0,0 +1,23 @@
+import { BaselineSyntax, compileArgs } from '@glimmer/runtime';
+import { unwrap } from '@glimmer/util';
+
+const {
+  defaultBlock,
+  params,
+  hash
+} = BaselineSyntax.NestedBlock;
+
+export function _inElementMacro(sexp, builder) {
```
rwjblue: Why is this moved into Ember itself? I thought it was baked into Glimmer?

packages/ember-glimmer/lib/syntax.js
```diff
@@ -1,44 +1,85 @@
...
+import { assert } from 'ember-metal';
 
+function refineInlineSyntax(path, params, hash, builder) {
+  let [ name ] = path;
 
-let syntaxKeys = [];
-let syntaxes = [];
+  assert(`You attempted to overwrite the built-in helper "${name}" which is not allowed. Please rename the helper.`, !(builder.env.builtInHelpers[name] && builder.env.owner.hasRegistration(`helper:${name}`)));
 
-export function registerSyntax(key, syntax) {
```
rwjblue: I would still like to export a mechanism to register macros. Without one we loose the ability to experiment with different syntaxes (e.g. `ember-let`). 

It seems totally good to change it to a more generic thing like `registerMacro`, but we shouldn't remove the ability to customize/experiment.

package.json
```diff
@@ -27,6 +27,11 @@
     "test:testem": "testem -f testem.dist.json"
   },
   "dependencies": {
+    "@glimmer/compiler": "^0.21.1",
+    "@glimmer/node": "^0.21.1",
+    "@glimmer/reference": "^0.21.0",
```
chadhietala: No I don't believe so. The way `lerna publish` works is that it looks if packages were effected in between the `HEAD` and the last tag if its not effected it will not publish a new version.

packages/ember-glimmer/lib/syntax/-in-element.js
chadhietala: Seems like it was removed from Glimmer.

packages/ember-glimmer/lib/syntax.js
chadhietala: I added an API for this. Let me know what you think.

package.json
```diff
@@ -35,12 +40,14 @@
     "ember-cli-test-info": "^1.0.0",
     "ember-cli-valid-component-name": "^1.0.0",
     "ember-cli-version-checker": "^1.1.7",
+    "handlebars": "~3",
```
chadhietala: @krisselden might want to chime in here. I don't remember why we did this.

packages/ember-glimmer/lib/helpers/component.js
krisselden: at least in dev mode, freezing it does make it a different "shape" polymorphic with the unfrozen versions, not sure if it is already polymorphic and empty and non empty should be both seen by the time it is JITed so may not matter.

rwjblue:
Awesome job here @chadhietala!

PR #14784, opened 2017-01-04
[performance] avoid looking up viewRegistry n times

GavinJoyce:
This change avoids getting `owner` and looking up `viewRegistry` for each event at the slight complexity expense of an additional parameter to the `setupHandler` method. We setup [27 events by default](https://github.com/emberjs/ember.js/blob/master/packages/ember-views/lib/system/event_dispatcher.js#L59-L87) so this change is worth it IMO (provided that there are no other considerations that I'm missing).

Questions:

 * Is the perf win worth the more complicated interface?
 * Perhaps memoization be better? We'd get to keep the same method interface and with the slight expense of n function calls to the memoized function.
 * This function is marked private, but does anyone rely on it (internally or externally)
 * Are there any concerns with application testing?


chancancode (approved):

mixonic:
👍

Issue #14615, opened 2016-11-17
current-when not working with dynamic segment


amk221:
Here is an example:

https://ember-twiddle.com/93563986146cde7c19cb7e9e985f32c0?openFiles=templates.application.hbs%2C&route=%2Ffoo%2F1

locks:
@amk221 I don't see any problem, can you please describe your expectation and what's actually happening?

amk221:
Sure, the hyperlink to `foo` has `current-when='foo.index foo.bar'`
Therefore, I am expecting that hyperlink to be _bold_ (active) when I visit `foo.bar`
But it does not.

Remove the dynamic segment from the link-to, and it works as expected.

locks:
Ah, it's `current-when=['foo.index', 'foo.bar']`. Not that I recommend it :P

amk221:
Why do you not recommend it?
Why is it not documented?
Why does it work when removing the dynamic segment?

locks:
> Why do you not recommend it?
> Why is it not documented?
Because it should ideally be done via route nesting/architecture.

> Why does it work when removing the dynamic segment?
I don't know what you mean

amk221:
> Because it should ideally be done via route nesting/architecture.

If I am doing something ill-advised, can you link me to how I should be doing it?

> Why does it work when removing the dynamic segment?
> I don't know what you mean
I mean, if you remove the number "1" from the `link-to` it will highlight correctly.

locks:
Oh, you were passing two arguments to a route (`foo`) that only has one dynamic segment.

amk221:
:) nope, I'm not the arguments are correct. 
'foo' 'is the title text
'foo' is the route
'1' is the dynamic arg

psteininger:
@amk221 did you ever get around this issue? I am running into the same situation, on a link with 2 dynamic elements `/facilities/:facility_id/issues/:issue_id` on `/facilities/:facility_id/issues` it works, but on `/facilities/:facility_id/issues/:issue_id` it doesn't despite:
```hbs
{{#link-to 'facility.issues' model current-when='facility.issues facility.issue'}}
```
the routes are like so:
```js
...
 this.route('facilities', {resetNamespace: true});
    this.route('facility', {resetNamespace: true, path: '/facilities/:facility_id'}, function () {
      ...
      this.route('issues');
      this.route('issue', {path: '/issues/:issue_id'});
    });
...
```
[...]

PR #14787, opened 2017-01-04
Add failing test for current-when

amk221:
Failing test for a scenario where using `current-when` is not working as expected.

I've set up a simple Twiddle, showing that it works as expected _without_ route params, but if route params are introduced, it does not work.

[Original issue #14615](https://github.com/emberjs/ember.js/issues/14615)
[Example twiddle](https://ember-twiddle.com/93563986146cde7c19cb7e9e985f32c0?openFiles=templates.application.hbs%2C&route=%2Ffoo%2F1%2Fnested%2F2)


locks:
Can you migrate this to a QUnit todo?

rwjblue:
Sadly, I don't think we can use todo's just yet (testem doesn't support them).

locks:
Ah darn :/ Apologies for the suggestion.

Krinkle:
> Sadly, I don't think we can use todo's just yet (testem doesn't support them).

Fixed :) https://github.com/testem/testem/pull/1400

kategengler:
Is this still relevant?

johanrd:
> Is this still relevant?

yes, tried to run the test now on main, and the test fails with:
  - current-when with nested routes containing dynamic segments (:fooId, :barId) doesn't correctly mark links as active
  - The first assertion fails: the link with current-when='foo.index foo.bar' is not marked .active when visiting /foo/1/bar/2                                                                                       
  - seems to be a bug in link-to's isActiveForRoute() logic — it doesn't properly handle route matching when dynamic segments are involved in current-when routes

johanrd:
this can now be closed, see https://github.com/emberjs/ember.js/pull/21202

NullVoxPopuli:
Tyty

Issue #14560, opened 2016-10-31
Using `controllerName` with a parent controller that has a query-param causes assertion on Ember 2.10-beta.2. 


workmanw:
Our ember-try tests alerted us to a possible regression. After investigation I discovered that if a route declares a `controllerName` that matches a parent controller and that parent controller has query-params it will cause an assertion to be thrown when using a `link-to`; or `this.transitionTo` w/ query-params.

This assertion only occurs on Ember-2.10-beta, it does not occur on Ember-2.9.

Here is a twiddle: https://ember-twiddle.com/515fb727fe1e8b39b2520ec3fae23bff

In this scenario we have an "invite key" for a user. Redeeming an invite is a multi-step [multi-route] process similar to most SaaS apps. The namespaced controller (`controllers/invite`) houses the application state for all steps in this process, including the query-param (`key`). Each of the route classes declare `controllerName: 'invite'` so their templates can interface with the `invite` controller directly.

The problem is that now two handlers (`invite` and `invite.verify`) are both using the same controller. When `transitionTo` calls into the router's `_prepareQueryParams` function, a newly added check throws this exception because both handlers are using the same query-params (https://github.com/emberjs/ember.js/commit/d1124fa8424ee71125f801bf93e361de29df8a86). The exception is also a bit non-obvious:

> Assertion Failed: You're not allowed to have more than one controller property map to the same query param key, but both `invite:key` and `invite:key` map to `key`. You can fix this by mapping one of the controller properties to a different query param key via the `as` config option, e.g. `key: { as: 'other-key' }

So why use `controllerName` at all? Why not just use an invite service to house the state from each phase of the invite process? The answer is two parts: 1) It's old code and we haven't touched it in a long long time :stuck_out_tongue_winking_eye:, but 2) there is not a clean way for a service to access query-params on the router. 

--------

So yea, I know this is just another one of the crazy query-param issues that has come up. If you guys want to just ¯\\\_(ツ)_/¯, I'm happy with that. I can probably find my own work around. But as always I'm just reporting the issues I see.

rwjblue:
I haven't really dug into the demo yet, but this is related to a large scale router and query param cleanup effort. 

Most likely @trentmwillis will have a better understanding of exactly what is going on here, and can advise to it being a "bug" or a new "feature" ( 😺  ).

trentmwillis:
I'll have to dig in more later, but this assertion isn't really new, it was just moved to a different part of the QP flow.

[...]

PR #14791, opened 2017-01-05
[BUGFIX beta] exempt routes that share a controller from duplicate assertion

ghost:
start of fix for #14560

### DIFF
--- a/packages/ember-routing/lib/system/router.js
+++ b/packages/ember-routing/lib/system/router.js
@@ -858,8 +858,9 @@ const EmberRouter = EmberObject.extend(Evented, {
       for (let i = 0; i < qpMeta.qps.length; i++) {
         let qp = qpMeta.qps[i];
         let urlKey = qp.urlKey;
+        let qpOther = qpsByUrlKey[urlKey];
 
-        if (qpsByUrlKey[urlKey]) {
+        if (qpOther && qpOther.controllerName !== qp.controllerName) {
           let otherQP = qpsByUrlKey[urlKey];
           assert(`You're not allowed to have more than one controller property map to the same query param key, but both \`${otherQP.scopedPropertyName}\` and \`${qp.scopedPropertyName}\` map to \`${urlKey}\`. You can fix this by mapping one of the controller properties to a different query param key via the \`as\` config option, e.g. \`${otherQP.prop}: { as: \'other-${otherQP.prop}\' }\``, false);
         }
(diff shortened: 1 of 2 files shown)


efx:
@trentmwillis mind reviewing to ensure fix is in right direction? 
Ideally, a fix for this could be merged since this issue is blocking our upgrade. I figured a PR with the beginnings here is more sensible then writing some workaround specific to our application.

packages/ember-routing/lib/system/router.js
```diff
@@ -858,8 +858,9 @@ const EmberRouter = EmberObject.extend(Evented, {
       for (let i = 0; i < qpMeta.qps.length; i++) {
         let qp = qpMeta.qps[i];
         let urlKey = qp.urlKey;
+        let qpOther = qpsByUrlKey[urlKey];
 
-        if (qpsByUrlKey[urlKey]) {
+        if (qpOther && (qpOther.hasOwnProperty('controllerName') && qpOther.controllerName !== qp.controllerName)) {
```
trentmwillis: I'm unsure the `hasOwnProperty` check here is necessary. Ideally if a collision happens, and the property isn't there then we should do the assertion.

packages/ember-routing/tests/system/router_test.js
```diff
@@ -255,3 +255,57 @@ QUnit.test('Router#triggerEvent ignores handlers that have not loaded yet', func
 
   triggerEvent(handlerInfos, false, ['loading']);
 });
+
+QUnit.test('Router#_queryParamsFor returns the cached leaf handler when present', function(assert) {
```
trentmwillis: I would rather these tests be done as acceptance-type tests rather than unit tests. They should probably go in this test file which does the other collision checks: https://github.com/emberjs/ember.js/blob/master/packages/ember/tests/routing/query_params_test/overlapping_query_params_test.js

trentmwillis (changes requested):
Thanks for working on this!

efx:
thanks for the feedback; will be updating here within the morning.

packages/ember/tests/routing/query_params_test/overlapping_query_params_test.js
```diff
@@ -105,7 +106,20 @@ moduleFor('Query Params - overlapping query param property names', class extends
       this.assertCurrentPath('/parent/child?childPage=3&parentPage=4');
     });
   }
+  ['@test query params does not error when a query parameter exists for route instances that share a controller'](assert) {
```
trentmwillis: Nitpick: Need a newline before and after the test block

trentmwillis (changes requested):
One small thing, otherwise seems good to me. Will need a review from an actual maintainer though.

trentmwillis (approved):
Looks good to me!

rwjblue:
OK, I squashed, rebased, and updated commit message.  Will merge once Travis is green...

efx:
thanks @trentmwillis and @rwjblue !

PR #14793, opened 2017-01-06
Avoid raising _lookupFactory deprecation during lookup

mixonic:
The entire `container.lookupFactory` API is deprecated, however it was also being [used in `lookup`](https://github.com/mixonic/ember.js/blob/333f4b05c0504d2b3be9002c796bd2f11f7fefe7/packages/container/lib/container.js#L144) and thus caused deprecation messaged to be raise at times other than when we wanted.

### DIFF
--- a/packages/container/lib/container.js
+++ b/packages/container/lib/container.js
@@ -173,7 +173,7 @@ Container.prototype = {
         throw new Error('If ember-no-double-extend is enabled, ember-factory-for must also be enabled');
       }
     }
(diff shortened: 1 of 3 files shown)


chadhietala:
lgtm

mixonic:
screen shot 2017-01-06 at 10 53 37 am

^ One more failure when only `ember-factory-for` is enabled.

mixonic:
That last failure should be good now.

Issue #14788, opened 2017-01-04
Query param is sticky for different route model in routeable ember-engine


rajaalauddin:
I have reproduced this issue in this repo https://github.com/rajaalauddin/ember-engines-demo

Sample steps on how to reproduce the issue:
1) Create a route inside routable ember engine with dynamic segment. e.g /viewers/:id
2) Create link-to /viewers/2 for viewers.hbs template
3) Add query param to viewer controller e.g queryParams: ['active']
3) Go to /viewers/1?active=true
4) See how link-to generated on /viewers/2 also has the query param e.g /viewers/2?active=true

These are not the case when done outside engine.

Another developer and i have traced this issue down to how calculateCacheKey is called.

`var cacheKey = _emberRoutingUtils.calculateCacheKey(aQp.controllerName, aQp.parts, aQp.values);`

Changing the above to:

`var cacheKey = _emberRoutingUtils.calculateCacheKey(aQp.route.fullRouteName, aQp.parts, aQp.values);`

fixed the problem for us. Let me know if this is the correct solution so that i can create a pull request.

rwjblue:
Thanks for reporting!  

>  Let me know if this is the correct solution so that i can create a pull request

The solution seems reasonable in general (specifically, we need to ensure that the cache key is built up correctly inside the engine). I think the next step is getting an isolated failing test case.  

Please take a look at some of these [engine tests](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/tests/integration/application/engine-test.js)
tests for an idea of how to set it up.  I believe we should be able to add a test in that file that replicates the scenario you describe in the issue description.

pixelhandler:
@rajaalauddin perhaps make a PR for the failing test? And the fix too :)

rajaalauddin:
@pixelhandler I have created pull request here https://github.com/emberjs/ember.js/pull/14794

Serabe:
Closed via #14794

PR #14794, opened 2017-01-06
Fix query param stickiness between models in ember-engines

rajaalauddin:
See https://github.com/emberjs/ember.js/issues/14788

Summary:

In ember-engines, query param is sticky between different models (see generated link-to). This is not the behavior outside engine and as specified https://guides.emberjs.com/v2.10.0/routing/query-params/#toc_sticky-query-param-values

This is happening because they share the same cacheKey when controller is used as the prefix.

### DIFF
--- a/packages/ember-glimmer/tests/integration/application/engine-test.js
+++ b/packages/ember-glimmer/tests/integration/application/engine-test.js
@@ -20,6 +20,7 @@ moduleFor('Application test: engine rendering', class extends ApplicationTest {
         this.route('comments');
         this.route('likes');
       });
+      this.route('category', {path: 'category/:id'});
     });
     this.registerRoute('application', Route.extend({
       model() {
@@ -33,6 +34,9 @@ moduleFor('Application test: engine rendering', class extends ApplicationTest {
         this.register('controller:application', Controller.extend({
           queryParams: ['lang'],
           lang: ''
+        }));
+        this.register('controller:category', Controller.extend({
+          queryParams: ['type'],
         }));
         this.register('template:application', compile('Engine{{lang}}{{outlet}}'));
         this.register('route:application', Route.extend({
@@ -570,6 +574,20 @@ moduleFor('Application test: engine rendering', class extends ApplicationTest {
       return transition.then(() => {
         this.runTaskNext(() => this.assertText('ApplicationEngineLikes'));
       });
+    });
+  }
+
+  ['@test query params don\'t have stickiness by default between model'](assert) {
+    assert.expect(1);
+
+    this.setupAppAndRoutableEngine();
+    this.additionalEngineRegistrations(function() {
+      this.register('template:category', compile('{{#link-to "blog.category" 1337}}Category 1337{{/link-to}}'))
+    });
+
+    return this.visit('/blog/category/1?type=news').then(() => {
+      let href = this.element.querySelector('a').href;
+      assert.equal(href.match(/type=news/), null);
     });
   }
 });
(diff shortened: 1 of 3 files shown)


packages/ember-glimmer/tests/integration/application/engine-test.js
```diff
@@ -572,4 +576,18 @@ moduleFor('Application test: engine rendering', class extends ApplicationTest {
...
+    assert.expect(1);
+
+    this.setupAppAndRoutableEngine();
+    this.additionalEngineRegistrations(function() {
+      this.register('template:category', compile('{{#link-to "blog.category" 1337}}Category 1337{{/link-to}}'))
+    });
+
+    return this.visit('/blog/category/1?type=news').then(() => {
+      let href = this.element.querySelector('a').href;
+      assert.equal(href.match(/type=news/), null);
```
trentmwillis: Can we change this to just be `assert.equal(href, '/blog/category/1337');`? It's easier to follow and will give clearer feedback (IMO) if it breaks.

trentmwillis (changes requested):
One small thing, but overall this seems good to me. I think using the route name also is in the direction of moving away from controllers. @rwjblue can you review when you have a chance?

rwjblue:
This is looking good!  A few specific issues that we should chat through are:

* It is odd to have most of the QP system based on `controllerName` and another part based on `route.fullRouteName`.
* How do we handle scenarios where folks have customized `controllerName` to specify a controller other than the route's controller is handling their QP's?  With this change, I _think_ that use case might break?

packages/ember-glimmer/tests/integration/application/engine-test.js
rwjblue: agreed, I'd much prefer to assert against the value we actually expect...

rajaalauddin: I've updated this to use .endsWith since full href include port number. Let me know if you need further change.

rwjblue: Should be able to use something like:

```js
let { pathname } = this.element.querySelector('a');
assert.equal(pathname, '/blog/category/1337');
```

rajaalauddin: `endsWith` solution didn't work because of not being implemented in phantomjs. `pathname` solution also didn't work because of ie9 different implementation. I changed this to `indexOf` to simulate `endsWith`. Let me know if you guys have better preference.

packages/ember-glimmer/tests/integration/application/engine-test.js
```diff
@@ -572,4 +587,41 @@ moduleFor('Application test: engine rendering', class extends ApplicationTest {
...
+    return this.visit('/blog/category/1?type=news').then(() => {
+      let suffix = '/blog/category/1337';
+      let href = this.element.querySelector('a').href;
+
+      // check if link ends with the suffix
+      assert.ok(this.stringsEndWith(href, suffix));
+    });
+  }
+
+  ['@test query params in customized controllerName have stickiness by default between model'](assert) {
```
rajaalauddin: > How do we handle scenarios where folks have customized controllerName to specify a controller other than the route's controller is handling their QP's? With this change, I think that use case might break?

@rwjblue I added a test on this and you are correct this change will break queryParam in customized `controllerName`.

rajaalauddin: I stand corrected - this change doesn't break custom `controllerName` case. There was an error in the way i register the route causing the test to fail. I've updated the setup and now it's passing.

rajaalauddin:
> How do we handle scenarios where folks have customized controllerName to specify a controller other than the route's controller is handling their QP's? With this change, I think that use case might break?

@rwjblue I added test to cover this scenario and it's passing. Please verify if that is what you have in mind. thanks

PR #14804, opened 2017-01-09
[DOCS] Tweak addObserver function signature examples

knownasilya:
The semicolons would have been invalid JS. I also converted to ES2015 while I was here, let me know if that's fine.

### DIFF
--- a/packages/ember-runtime/lib/mixins/observable.js
+++ b/packages/ember-runtime/lib/mixins/observable.js
@@ -352,7 +352,16 @@ export default Mixin.create({
     you do not pass a `context` parameter:
 
     ```javascript
-    fooDidChange: function(sender, key, value, rev) { };
+    export default Ember.Component.extend({
+      init() {
+        this._super(...arguments);
+        this.addObserver('foo', this, 'fooDidChange');
+      },
+
+      fooDidChange(sender, key, value, rev) {
+        // your code
+      }
+    });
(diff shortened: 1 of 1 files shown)


packages/ember-runtime/lib/mixins/observable.js
```diff
@@ -352,7 +352,7 @@ export default Mixin.create({
     you do not pass a `context` parameter:
 
     ```javascript
-    fooDidChange: function(sender, key, value, rev) { };
```
locks: Can you wrap the examples in an object? I'm guessing this minimal code sample might be confusing.

locks (changes requested):
Thanks for spotting this.
Can you improve it a bit more since you're in there?

packages/ember-runtime/lib/mixins/observable.js
knownasilya: should i add an example like

```js
  init() {
    this._super(...arguments);
    this.addObserver('someProperty', this, 'aFunc');
  },

  aFunc(sender, key, value, rev) {
    // your code
  }
});
```

rwjblue: @knownasilya - Yes, that looks much better to me...

locks (approved):

locks:
Thank you very much!

PR #13868, opened 2016-07-23
[WIP] Router service

locks:

PR #14805, opened 2017-01-09
Public Router service MVP

scalvert:
This PR represents the MVP of the public router service described in [this RFC](https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md). It incorporates work done by @locks in [this PR](https://github.com/emberjs/ember.js/pull/13868).

It includes a minimal API:

- `currentRouteName`
- `currentURL`
- `location`
- `rootURL`
- `transitionTo`

**Status**: Work in progress

**Changes**:
- Added `ember-routing-router-service` feature flag
- Added new router service and minimal API as described above
- Added deprecation warning for use of private router service when feature enabled
- Tests in progress

**How to test drive this PR**:
- `yarn start`
- Open up 2 tabs pointing to `http://localhost:4200/tests/index.html`, one with `Enable Opt Features` checked, ensure tests pass 

**TODO**:
Write tests for
- [x] `currentRouteName`
- [x] `currentURL`
- [x]  `location`
- [x] `rootURL`
- [x] `transitionTo`

RFC Implementation Phases
----------------------------

1. MVP for public router service with minimal API
2. Extended API, building on the work in phase 1
3. Implementation of `urlFor`, ensuring correct performance profile
4. Implementation of `RouteInfo` and associated API

### DIFF
--- a/packages/ember-application/lib/system/application.js
+++ b/packages/ember-application/lib/system/application.js
@@ -34,6 +34,8 @@ import ApplicationInstance from './application-instance';
 import { privatize as P } from 'container';
 import Engine from './engine';
 import { setupApplicationRegistry } from 'ember-glimmer';
+import { RouterService } from 'ember-routing';
+import { isFeatureEnabled } from 'ember-metal';
 
 let librariesRegistered = false;
 
@@ -1037,6 +1039,11 @@ function commonSetupRegistry(registry) {
   registry.register('location:none', NoneLocation);
 
   registry.register(P`-bucket-cache:main`, BucketCache);
+
+  if (isFeatureEnabled('ember-routing-router-service')) {
+    registry.register('service:router', RouterService);
+    registry.injection('service:router', 'router', 'router:main');
+  }
 }
 
 function registerLibraries() {
--- a/packages/ember-routing/lib/services/router.js
+++ b/packages/ember-routing/lib/services/router.js
@@ -1,0 +1,49 @@
+/**
+@module ember
+@submodule ember-routing
+*/
+
+import {
+  Service,
+  readOnly
+} from 'ember-runtime';
+import { get } from 'ember-metal';
+import RouterDSL from '../system/dsl';
+
+/**
+   The Router service is the public API that provides component/view layer
+   access to the router.
+
+   @public
+   @class RouterService
+   @category ember-routing-router-service
+ */
+const RouterService = Service.extend({
+  currentRouteName: readOnly('router.currentRouteName'),
+  currentURL: readOnly('router.currentURL'),
+  location: readOnly('router.location'),
+  rootURL: readOnly('router.rootURL'),
+
+  /**
+     Transition the application into another route. The route may
+     be either a single route or route path:
+
+     See [Route.transitionTo](http://emberjs.com/api/classes/Ember.Route.html#method_transitionTo) for more info.
+
+     @method transitionTo
+     @category ember-routing-router-service
+     @param {String} name the name of the route or a URL
+     @param {...Object} models the model(s) or identifier(s) to be used while
+       transitioning to the route.
+     @param {Object} [options] optional hash with a queryParams property
+       containing a mapping of query parameters
+     @return {Transition} the transition object associated with this
+       attempted transition
+     @public
+   */
+  transitionTo() {
+    this.router.transitionTo(...arguments);
+  }
+});
+
+export default RouterService;
--- a/packages/ember-routing/lib/system/router.js
+++ b/packages/ember-routing/lib/system/router.js
@@ -138,6 +138,10 @@ const EmberRouter = EmberObject.extend(Evented, {
   init() {
     this._super(...arguments);
 
+    this.currentURL = null;
+    this.currentRouteName = null;
+    this.currentPath = null;
+
     this._qpCache = new EmptyObject();
     this._resetQueuedQueryParameterChanges();
     this._handledErrors = dictionary(null);
@@ -629,6 +633,7 @@ const EmberRouter = EmberObject.extend(Evented, {
 
     let doUpdateURL = function() {
       location.setURL(lastURL);
+      set(emberRouter, 'currentURL', lastURL);
     };
 
     router.updateURL = function(path) {
@@ -639,6 +644,7 @@ const EmberRouter = EmberObject.extend(Evented, {
     if (location.replaceURL) {
       let doReplaceURL = function() {
         location.replaceURL(lastURL);
+        set(emberRouter, 'currentURL', lastURL);
       };
 
       router.replaceURL = function(path) {
@@ -1291,9 +1297,11 @@ function updatePaths(router) {
 
   let path = EmberRouter._routePath(infos);
   let currentRouteName = infos[infos.length - 1].name;
+  let currentURL = router.get('location').getURL();
 
   set(router, 'currentPath', path);
   set(router, 'currentRouteName', currentRouteName);
+  set(router, 'currentURL', currentURL);
 
   let appController = getOwner(router).lookup('controller:application');
 
--- a/packages/ember/tests/routing/router_service_test/currenturl_lifecycle_test.js
+++ b/packages/ember/tests/routing/router_service_test/currenturl_lifecycle_test.js
@@ -1,0 +1,206 @@
+import Logger from 'ember-console';
+import {
+  Controller,
+  inject,
+  readOnly
+} from 'ember-runtime';
+import { Component } from 'ember-glimmer';
+import { Route, NoneLocation } from 'ember-routing';
+import {
+  run,
+  get,
+  set
+} from 'ember-metal';
+import { jQuery } from 'ember-views';
+import {
+  ApplicationTestCase,
+  moduleFor
+} from 'internal-test-helpers';
+
+import { isFeatureEnabled } from 'ember-metal';
+
+if (isFeatureEnabled('ember-routing-router-service')) {
+  let results = [];
+  let ROUTE_NAMES = ['index', 'child', 'sister', 'brother'];
+
+  let InstrumentedRoute = Route.extend({
+    routerService: inject.service('router'),
+
+    beforeModel() {
+      let service = get(this, 'routerService');
+      results.push([service.get('currentRouteName'), 'beforeModel', service.get('currentURL')]);
+    },
+
+    model() {
+      let service = get(this, 'routerService');
+      results.push([service.get('currentRouteName'), 'model', service.get('currentURL')]);
+    },
+
+    afterModel() {
+      let service = get(this, 'routerService');
+      results.push([service.get('currentRouteName'), 'afterModel', service.get('currentURL')]);
+    }
+  });
+
+  moduleFor('Router Service - currentURL lifecycle', class extends ApplicationTestCase {
+    constructor() {
+      super();
+
+      results = [];
+
+      this.router.map(function() {
+        this.route('parent', { path: '/' }, function() {
+          this.route('child');
+          this.route('sister');
+          this.route('brother');
+          this.route('stepsister');
+        });
+        this.route('dynamic', { path: '/dynamic/:post_id' });
+      });
+
+      ROUTE_NAMES.forEach((name) => {
+        let routeName = `parent.${name}`;
+        this.registerRoute(routeName, InstrumentedRoute.extend());
+        this.registerTemplate(routeName, '{{current-url}}');
+      });
+
+      this.registerComponent('current-url', {
+        ComponentClass: Component.extend({
+          routerService: inject.service('router'),
+          currentURL: readOnly('routerService.currentURL')
+        }),
+        template: '{{currentURL}}'
+      });
+    }
+
+    ['@test RouterService#currentURL is correctly set for top level route'](assert) {
+      assert.expect(1);
+
+      let routerService;
+
+      this.registerRoute('parent.index', Route.extend({
+        routerService: inject.service('router'),
+        init() {
+          this._super();
+          routerService = get(this, 'routerService');
+        }
+      }));
+
+      return this.visit('/').then(() => {
+        assert.equal(routerService.get('currentURL'), '/');
+      });
+    }
+
+    ['@test RouterService#currentURL is correctly set for child route'](assert) {
+      assert.expect(1);
+
+      let routerService;
+
+      this.registerRoute('parent.child', Route.extend({
+        routerService: inject.service('router'),
+        init() {
+          this._super();
+          routerService = get(this, 'routerService');
+        }
+      }));
+
+      return this.visit('/child').then(() => {
+        assert.equal(routerService.get('currentURL'), '/child');
+      });
+    }
+
+    ['@test RouterService#currentURL is correctly set after transition'](assert) {
+      assert.expect(1);
+
+      let routerService;
+
+      this.registerRoute('parent.child', Route.extend({
+        routerService: inject.service('router'),
+        init() {
+          this._super();
+          routerService = get(this, 'routerService');
+        },
+
+        afterModel() {
+          this.transitionTo('parent.sister');
+        }
+      }));
+
+      return this.visit('/child').then(() => {
+        assert.equal(routerService.get('currentURL'), '/sister');
+      });
+    }
+
+    ['@test RouterService#currentURL is correctly set on each transition'](assert) {
+      assert.expect(3);
+
+      let routerService;
+
+      this.registerRoute('parent.child', Route.extend({
+        routerService: inject.service('router'),
+        init() {
+          this._super();
+          routerService = get(this, 'routerService');
+        }
+      }));
+
+      return this.visit('/child')
+        .then(() => {
+          assert.equal(routerService.get('currentURL'), '/child');
+
+          return this.visit('/sister');
+        })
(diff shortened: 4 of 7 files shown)


packages/ember-routing/lib/services/router.js
```diff
@@ -0,0 +1,62 @@
...
+    @param {Object} [options] optional hash with a queryParams property
+      containing a mapping of query parameters
+    @return {Transition} the transition object associated with this
+      attempted transition
+    @public
+  */
+  urlFor() {
+    let router = get(this, 'router');
+
+    return router.generate(...arguments);
```
chadhietala: Doesn't this still take the slow path with the QPs?

scalvert: Good question. Will verify.

krisselden (changes requested):
Any API that doesn't meet the RFC here https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md#query-parameter-semantics shouldn't be included until it does.

packages/ember-routing/lib/services/router.js
scalvert: Removing until we verify the implementation details.

scalvert:
Added tests for `currentRouteName`.

scalvert:
@krisselden I've stripped the `urlFor` implementation out of this review. We can work out how to proceed in a subsequent review.

packages/ember-application/lib/system/application.js
```diff
@@ -34,6 +34,8 @@ import ApplicationInstance from './application-instance';
 import { privatize as P } from 'container';
 import Engine from './engine';
 import { setupApplicationRegistry } from 'ember-glimmer';
+import RouterService from 'ember-routing/services/router';
+import isEnabled from 'ember-metal/features';
```
rwjblue: `import { isFeatureEnabled } from 'ember-metal';`

packages/ember-application/lib/system/application.js
```diff
@@ -34,6 +34,8 @@ import ApplicationInstance from './application-instance';
 import { privatize as P } from 'container';
 import Engine from './engine';
 import { setupApplicationRegistry } from 'ember-glimmer';
+import RouterService from 'ember-routing/services/router';
```
rwjblue: `import { RouterService } from 'ember-routing';`

packages/ember-routing/lib/services/router.js
```diff
@@ -0,0 +1,49 @@
+/**
+@module ember
+@submodule ember-routing
+*/
+
+import Service from 'ember-runtime/system/service';
```
rwjblue: Imports should be from the entry points (e.g. `from 'ember-runtime';` instead of `from 'ember-runtime/somethign';`)

packages/ember-routing/lib/services/router.js
```diff
@@ -0,0 +1,49 @@
...
+
+    @method transitionTo
+    @param {String} name the name of the route or a URL
+    @param {...Object} models the model(s) or identifier(s) to be used while
+      transitioning to the route.
+    @param {Object} [options] optional hash with a queryParams property
+      containing a mapping of query parameters
+    @return {Transition} the transition object associated with this
+      attempted transition
+    @public
```
rwjblue: We should add `@category ember-routing-router-service` so that the YUIDoc will be able to remove these until the feature is enabled

packages/ember-routing/lib/services/router.js
```diff
@@ -0,0 +1,49 @@
...
+  */
+  transitionTo() {
+    let router = get(this, 'router');
```
rwjblue: Since this is injected we should be able to swap this to just `this.router` here.

packages/ember-routing/lib/services/router.js
```diff
@@ -0,0 +1,49 @@
...
+
+    router.transitionTo(...arguments);
+  }
+});
+
+RouterService.map = RouterDSL.map;
```
rwjblue: We may not want to do this `RouterDSL.map` part yet, lets remove?

packages/ember/tests/routing/router_service_test.js
```diff
@@ -0,0 +1,121 @@
+import Logger from 'ember-console';
+import { Controller } from 'ember-runtime';
+import { Route } from 'ember-routing';
+import {
+  run,
+  get
+} from 'ember-metal';
+import { jQuery } from 'ember-views';
+import { inject } from 'ember-runtime';
```
rwjblue: Group with the `ember-runtime` import above...

packages/ember/tests/routing/router_service_test.js
```diff
@@ -0,0 +1,121 @@
...
+import { ApplicationTestCase, moduleFor } from 'internal-test-helpers';
+
+import isEnabled from 'ember-metal/features';
```
rwjblue: `import { isFeatureEnabled } from 'ember-metal';`

packages/ember/tests/routing/router_service_test.js
```diff
@@ -0,0 +1,121 @@
...
+      let routerService;
+
+      this.router.map(function() {
+        this.route('parent', { path: '/' });
+      });
+
+      this.registerRoute('parent', Route.extend({
+        router: inject.service(),
+        init() {
+          routerService = get(this, 'router');
```
rwjblue: we should call `_super` before `this` here

packages/ember/tests/routing/router_service_test.js
```diff
@@ -0,0 +1,121 @@
...
+        }
+      }));
+
+      return this.visit('/child').then(() => {
+        assert.equal(routerService.get('currentRouteName'), 'parent.child');
+
+        return this.visit('/sister').then(() => {
```
rwjblue:
```js
return this.visit('/child').then(() => {
  return this.visit('/sister');
}).then(() => {
  assert.equal(routerService.get('currentRouteName'), 'parent.sister');

  return this.visit('/brother');
}).then(() => {
  assert.equal(routerService.get('currentRouteName'), 'parent.brother');
});
```

packages/ember-routing/lib/services/router.js
```diff
@@ -0,0 +1,49 @@
...
+  access to the router.
+
+  @public
+  @class RouterService
+*/
+const RouterService = Service.extend({
+  currentRouteName: readOnly('router.currentRouteName'),
+  currentURL: readOnly('router.location.path'),
+  location: readOnly('router.location'),
+  rootUrL: readOnly('router.rootURL'),
```
rwjblue: Correct the casing here (I think it should be `rootURL`)

packages/ember-application/lib/system/application.js
scalvert: Done

packages/ember-application/lib/system/application.js
scalvert: Done

packages/ember-routing/lib/services/router.js
scalvert: Done

packages/ember-routing/lib/services/router.js
scalvert: Done

packages/ember-routing/lib/services/router.js
scalvert: Done

packages/ember-routing/lib/services/router.js
scalvert: Done

packages/ember-routing/lib/services/router.js
scalvert: Removed.

packages/ember/tests/routing/router_service_test.js
scalvert: Done.

packages/ember/tests/routing/router_service_test.js
scalvert: Done

packages/ember/tests/routing/router_service_test.js
scalvert: Done for all inits.

packages/ember/tests/routing/router_service_test.js
scalvert: Done

packages/ember-routing/lib/services/router.js
```diff
@@ -0,0 +1,48 @@
...
+*/
+
+import {
+  Service,
+  readOnly
+} from 'ember-runtime';
+import { get } from 'ember-metal';
+import RouterDSL from '../system/dsl';
+
+/*
```
locks: The docstrings don't seem to match out usual style, see [Router](https://github.com/emberjs/ember.js/blob/master/packages/ember-routing/lib/system/router.js#L226-L253) for example.

locks (changes requested):

packages/ember-routing/lib/services/router.js
scalvert: I'll update.

krisselden (approved):

krisselden:
I'd love @ef4 review to see if there is any concern about the timing of when these properties update as compared to the RFC, I know that the RFC was a chance to improve the router not just give it a prettier API.

packages/ember-routing/lib/system/router.js
```diff
@@ -629,6 +633,7 @@ const EmberRouter = EmberObject.extend(Evented, {
 
     let doUpdateURL = function() {
       location.setURL(lastURL);
+      set(emberRouter, 'currentURL', lastURL);
```
rwjblue: I think we should actually do this in the `router.updateURL` and `router.replaceURL` methods directly above the `run.once(doUpdateURL)` bit...

packages/ember-routing/lib/system/router.js
```diff
@@ -1291,9 +1297,11 @@ function updatePaths(router) {
 
   let path = EmberRouter._routePath(infos);
   let currentRouteName = infos[infos.length - 1].name;
+  let currentURL = router.get('location').getURL();
 
   set(router, 'currentPath', path);
   set(router, 'currentRouteName', currentRouteName);
+  set(router, 'currentURL', currentURL);
```
rwjblue: I would think this change is not needed since we are setting `router.currentURL` above in the changes to `_setupRouter`.

rwjblue (commented):
This is looking really good!  I'm glad you got to the bottom of that `currentURL` async issue we chatted about. 👏 

I think the only real thing missing here at this point are some tests for `.transitionTo`, `.rootURL`, and `.location` properties.

packages/ember-routing/lib/services/router.js
```diff
@@ -0,0 +1,48 @@
...
+} from 'ember-runtime';
+import { get, descriptor } from 'ember-metal';
+import RouterDSL from '../system/dsl';
+
+/**
+   The Router service is the public API that provides component/view layer
+   access to the router.
+
+   @public
+   @class RouterService
```
rwjblue: Should add the `@category ember-routing-router-service` flag here too (so that while the feature is pending it isn't listed in emberjs.com/api).

packages/ember-routing/lib/services/router.js
```diff
@@ -0,0 +1,48 @@
+/**
+@module ember
+@submodule ember-routing
+*/
+
+import {
+  Service,
+  readOnly
+} from 'ember-runtime';
+import { get, descriptor } from 'ember-metal';
```
scalvert: I'll remove it. Thanks!

packages/ember-routing/lib/system/router.js
scalvert: I can do that. My thought was I wanted to keep both *set* calls in the same place, but these methods are so closely related that it's find either way.

packages/ember-routing/lib/services/router.js
scalvert: Done!

scalvert:
@rwjblue thanks for the additional comments. I'm adding tests for the outstanding properties/method now - mainly taking it slow so I make sure I cover high value test cases.

scalvert:
@krisselden I appreciate you wanting to make sure we're focussing on the right things with this PR, in terms of the opportunity to improve the router in general. This intent of this particular PR is to get a useful public API for the router in the hands of devs sooner rather than later. This will help provide immediate value, but does not preclude us circling back and refactoring parts of the internals later. In fact, that's our direct intention. We wanted to try to restrict these changes to simple, surface level changes before we get into the more meaty implementation, specifically the refactoring to use `RouteInfo`. Does this make sense to you?

I'll update the PR's description above to indicate the 4 phases, as discussed by @rwjblue and I.

packages/ember-routing/lib/system/router.js
scalvert: Ah, so it turns out that setting this after a scheduled call is what's fixing this issue. So, it looks like I have to leave it where it is.

packages/ember-routing/lib/system/router.js
scalvert: I thought this too, initially. It turns out that there's a different codepath whether you're visiting vs. transitioning, and we need this update to occur in both places. I think this surfaces part of @krisselden's concerns, where we want to take advantage, while we have the engine out on the floor, to refactor this to make it cleaner and more consistent.

workmanw:
This is super exciting. Thank you for taking this on and helping driving it home.

One additional thing I'd really like to see fall into one of the later phases is some mechanism for components to subscribe to `willTransition` and `didTransition`. We use that kind of thing so often in our app to prevent transitioning when data is dirty. I know the RFC talks about providing new hooks to do that and that might mean `willTransition` and `didTransition` have to wait until that part is flushed out. I just don't want to see these guys forgotten about.

locks (approved):

packages/ember/tests/routing/router_service_test.js
```diff
@@ -0,0 +1,432 @@
...
+        init() {
+          this._super();
+          routerService = get(this, 'routerService');
+        }
+      }));
+
+      return this.visit('/').then(() => {
+        let location = routerService.get('location');
+        assert.ok(location);
+        assert.ok(location instanceof Ember.NoneLocation);
```
rwjblue: This should be imported from `ember-routing/locations/none` (using global `Ember` within Ember or the test suite isn't allowed). This is currently causing the linting tests to fail...

packages/ember/tests/routing/router_service_test.js
```diff
@@ -0,0 +1,432 @@
...
+
+      let routerService;
+
+      this.buildRouterMap();
+
+      this.registerRoute('parent.index', Route.extend({
+        routerService: inject.service('router'),
+        init() {
+          this._super();
+          set(this.router, 'rootURL', '/test');
```
rwjblue: I have a slight preference to use something other than `/test` here (since when tests are running we are already in `/tests` path).  It definitely doesn't matter for this test as written, but it would just make this a tad bit easier for my brain to parse 😝

packages/ember/tests/routing/router_service_test.js
scalvert: Oh weird. I didn't see that failure. Will fix right now!

packages/ember/tests/routing/router_service_test.js
scalvert: Will change to `/brainparse` immediately! :)

packages/ember/tests/routing/router_service_test/currenturl_lifecycle_test.js
```diff
@@ -0,0 +1,209 @@
...
+          ]);
+        });
+    }
+
+    ['@test RouterService#currentURL is correctly set with component'](assert) {
+      assert.expect(1);
+
+      return this.visit('/').then(() => {
+        this.assertText('/');
+      });
```
rwjblue: Can you add an update and reset phase to this (kinda like the INUR stuff for the glimmer tests)?

I'm thinking something like:

```js
.then(() => routerService.transitionTo('child'))
.then(() => this.assertText('/child'))
.then(() => routerService.transitionTo('/'))
.then(() => this.assertText('/'))
```

packages/ember/tests/routing/router_service_test/currenturl_lifecycle_test.js
```diff
@@ -0,0 +1,209 @@
...
+
+      return this.visit('/')
+        .then(() => {
+          this.assertText('/');
+
+          return this.visit('/child');
+        })
+        .then(() => {
+          this.assertText('/child');
+        });
```
rwjblue: Can you add the update and reset steps (similar to comment for test above)?

rwjblue (approved):

nickiaconis:
@workmanw The RFC introduces the `routeWillChange` and `routeDidChange` events in place of the `willTransition` and `didTransition` hooks/events, but I agree. I too would like to see them on the roadmap.

janmisek:
What about rest of features covered by RFC (RouteInfo, transition events ...) ? is there any subsequent issue to track it ?

locks:
@janmisek There is not.

janmisek:
And are these features still planned for implementation ?

ef4:
Yes, the RFC is merged which means anyone can PR the features. I started on several of them but had to back off to clear some other blockers first.

PR #14809, opened 2017-01-10
Fix deopt for the field load of `.concat`

krisselden:
Removes polymorphic check of .concat and call that causes several deopts of mixin merge and apply.

This passed all meaningful tests, but there is one test that is asserting something I'm shocked about, that I'm going to submit this for review anyway.  If this were true we'd do things like call concat on a string with an array.

Which makes little sense to me.

I believe this test isn't testing a real thing.

### DIFF
--- a/packages/ember-metal/lib/mixin.js
+++ b/packages/ember-metal/lib/mixin.js
@@ -43,7 +43,9 @@ import {
 function ROOT() {}
 ROOT.__hasSuper = false;
 
-const a_slice = [].slice;
+const a_slice = Array.prototype.slice;
+const a_concat = Array.prototype.concat;
+const isArray = Array.isArray;
 
 function isMethod(obj) {
   return 'function' === typeof obj &&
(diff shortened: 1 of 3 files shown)


rwjblue (approved):

rwjblue:
That test struck me as quite odd when I recently worked on freezing concatenated properties in debug builds. I'm 👍 on changing it.

packages/ember-utils/lib/make-array.js
```diff
@@ -1,3 +1,5 @@
+const isArray = Array.isArray;
```
stefanpenner: should we use destructuring here? ` const { isArray } = Array`

packages/ember-metal/lib/mixin.js
```diff
@@ -43,7 +43,9 @@ import {
 function ROOT() {}
 ROOT.__hasSuper = false;
 
-const a_slice = [].slice;
+const a_slice = Array.prototype.slice;
```
stefanpenner: destructuring?

stefanpenner (approved):

PR #14816, opened 2017-01-11
Remove internal factoryInjection API

locks:
- [x] Container: A factory returned from lookupFactory has a debugkey
- [x] Ember.Application Dependency Injection – toString: instances
- [x] Ember.Application Dependency Injection – toString: with a custom resolver


packages/container/lib/container.js
```diff
@@ -447,16 +447,13 @@ function deprecatedFactoryFor(container, fullName, options = {}) {
...
-
-    factoryInjections[NAME_KEY] = registry.makeToString(factory, fullName);
+    let cacheable = !areInjectionsDynamic(injections);
 
     let injectedFactory = factory.extend(injections);
 
     // TODO - remove all `container` injections when Ember reaches v3.0.0
     injectDeprecatedContainer(injectedFactory.prototype, container);
-    injectedFactory.reopenClass(factoryInjections);
+    injectedFactory.reopenClass({ [NAME_KEY]: registry.makeToString(factory, fullName) });
```
rwjblue: Ideally we can get rid of this also (I'm not sure we have solved the "problem" of factory injections if we cannot remove this `reopenClass`).

rwjblue (commented):
I'm happy with all of this work, but I'd really like to see us remove that last `reopenClass` in `deprecatedFactoryFor`.  Though maybe it isn't worth a ton of effort since we plan to remove that whole chunk of code with the `ember-factory-for` feature flag...

stefanpenner:
@locks ping

stefanpenner:
@locks nice, I think this may be a bugfix beta?

locks:
@stefanpenner there was a problem with https://github.com/emberjs/ember.js/pull/14816/commits/5b2dc4d8137fe3634dad3af7db590062baaa0a28 that I don't know how to address. I think it particularly broke some of Ember Data's output.

stefanpenner:
> @stefanpenner there was a problem with 5b2dc4d that I don't know how to address. I think it particularly broke some of Ember Data's output.

@locks can you enumerate the problem, steps to reproduce?

rwjblue:
I believe this will be replaced by #15193.

PR #14817, opened 2017-01-12
From ES5 to ES6

artemgurzhii:
Code refactoring.


btecu:
Really nice!

You should discard the changes to `yarn.lock`.
Also, to be consistent, I would not add wrapping parenthesis to arrow function with a single argument:
`(prop) => ...` but `prop => ...`.

artemgurzhii:
Also, I'm getting an error when using `for of` loop and `String.includes`. Is there any way to fix this?

locks:
@artemgurzhii related: https://github.com/emberjs/emberjs-build/pull/186

locks:
@artemgurzhii the PR is getting big enough to be confusing to review. Could you send a PR per package instead? This would also speed up merging, since a problem in one file wouldn't block another file!

artemgurzhii:
Ok, sure.

locks:
Appreciated! Thanks for the work so far.

PR #14821, opened 2017-01-12
Code refactoring for ember-routing package

artemgurzhii:


packages/ember-routing/lib/system/dsl.js
```diff
@@ -133,7 +131,7 @@ function createRoute(dsl, name, options, callback) {
   dsl.push(options.path, fullName, callback, options.serialize);
 }
 
-DSL.map = function(callback) {
+DSL.map = callback => {
```
locks: I'm worried about this. cc @rwjblue

artemgurzhii:
Can I try to do some optimization for this package? After all, I will show benchmark tests for each method that I have optimized.

packages/ember-routing/lib/ext/controller.js
```diff
@@ -55,8 +55,8 @@ ControllerMixin.reopen({
     @method _qpChanged
     @private
   */
-  _qpChanged(controller, _prop) {
-    let prop = _prop.substr(0, _prop.length - 3);
+  _qpChanged(controller, prop) {
```
stefanpenner: we typically prefer to not overwrite argument bindings.

stefanpenner:
@artemgurzhii this looks awesome, thank you!

PR #14833, opened 2017-01-12
Code refactoring for ember-utils package

artemgurzhii:


bjornharrtell:
Are the extra () really needed for the function to arrow function replacements?

artemgurzhii:
Not sure that I understand what do you mean by that. Are you talking about [this line](https://github.com/emberjs/ember.js/pull/14833/commits/9261ac3d7a80967a1e929bbd276b4371c8286cb7#diff-522ab32c86869b9716d46d27c7d7f3b0L4)?

stefanpenner (approved):
🎉  👍

bjornharrtell:
This:
```js
export const HAS_NATIVE_WEAKMAP = ((() => {
  // detect if `WeakMap` is even present
  let hasWeakMap = typeof WeakMap === 'function';
  if (!hasWeakMap) { return false; }
}))();
```

Seems to me have superflous () i.e this is equivalent:

```js
export const HAS_NATIVE_WEAKMAP = (() => {
  // detect if `WeakMap` is even present
  let hasWeakMap = typeof WeakMap === 'function';
  if (!hasWeakMap) { return false; }
})();
```

But perhaps there is something I'm simply missing here.. :)

artemgurzhii:
I think I made mistake here, as I understand it should be an IIFE function, and this parenthesis can be omitted.

PR #14842, opened 2017-01-15
Code refactoring for ember-metal package

artemgurzhii:


packages/ember-metal/lib/alias.js
```diff
@@ -20,66 +20,67 @@ export default function alias(altKey) {
...
 }
 
-export function AliasedProperty(altKey) {
-  this.isDescriptor = true;
-  this.altKey = altKey;
-  this._dependentKeys = [altKey];
-}
+export class AliasedProperty extends Descriptor {
+  constructor(altKey) {
+    super();
```
stefanpenner: @krisselden / @chadhietala do we do the classCheck removal thing on the ember codebase? I don't recall, but you two may.

We should verify the associated compiled output for this is OK before we merge this one.

stefanpenner (changes requested):
We likely need to double check the output for the ES2015 `class`, make sure its good.

The rest of the PR likes fine, if we split the two we can land right away.

packages/ember-metal/lib/alias.js
stefanpenner: just checked, the output looks good

stefanpenner (approved):

stefanpenner:
🎉 👍

PR #12630, opened 2015-11-20
[BUGFIX release] allow current-when to be a boolean #12512

ssured:
Bring current-when in line with the docs

Sorry but I created this PR from within Github, as a start. My testing experience is not enough to create a test and run it locally on my machine. Hope this helps though...

Issue #12512, opened 2015-10-21
link-to current-when cannot be supplied as `true` as per docs


rlivsey:
The [docs specify][0] that:

```
  A link will be active if `current-when` is `true` or the current
  route is the route this link would transition to.
```

However, this currently does not appear to be the case, as `current-when` is always treated as a string and `split` [called on it][1].

Should the docs be updated to reflect that this is expected to just be a string separated list of route names, or should `true` work as documented? I shall PR either way!

[0]: https://github.com/emberjs/ember.js/blob/1743e317955f2957d07a7d749c2d02cdf7cad174/packages/ember-routing-views/lib/components/link-to.js#L158
[1]: https://github.com/emberjs/ember.js/blob/1743e317955f2957d07a7d749c2d02cdf7cad174/packages/ember-routing-views/lib/components/link-to.js#L561

rwjblue:
The fact that it doesn't work seems like a regression, and I think we should make the docs correct.

beerlington:
@rlivsey Did you find a work around for this?

rwjblue:
I think https://github.com/emberjs/ember.js/pull/12630 should fix this.

rlivsey:
@beerlington @rwjblue sorry I got sidetracked after opening this and then seem to have completely forgotten about it!

#12630 looks like it should do the trick though (could do with a test or two though to prevent future regressions)

beerlington:
I'll keep an eye on that, thanks!

locks:
Fixed by https://github.com/emberjs/ember.js/pull/12344.

PR #12344, opened 2015-09-14
[BUGFIX release-1-13] Allow current-when to be specified via a variable

Dhaulagiri:
Per guidance from @stefanpenner this fixes the issue where we are not able to set `current-when` to a variable property in 1.13.

Fixes #12296

Issue #12296, opened 2015-09-04
currentWhen.split is not a function


Dhaulagiri:
Upgrading from 1.12 to 1.13.9 I'm seeing a an issue using the `current-when` argument on a `link-to` helper resulting in an error `Uncaught TypeError: currentWhen.split is not a function`.  [This twiddle](http://ember-twiddle.com/c9cb299e54380e527fb6) demonstrates the problem which seems to occur regardless of if `currentWhen` is defined or not.

Note that the problem doesn't persist if I set `current-when` directly to a string in the template, but we need to be able to set this dynamically.

xtian:
Looks like this is fixed in canary:
http://ember-twiddle.com/50efad9f9a81c9b5613b

rwjblue:
Awesome, thanks for checking/confirming!

Dhaulagiri:
Will that fix be back ported to 1.13 @rwjblue?

rwjblue:
@Dhaulagiri - I'm not sure which commit fixes it.  I am not opposed to backporting if someone tracks it down, and it is safe enough for immediate release.

Dhaulagiri:
I can try to look, but I'm not super familiar with the Ember code base.  This is a blocking issue for our 1.13 upgrade though so I will try to make some time.

stefanpenner:
@rwjblue I would like for us to chat about this class of issue at the meeting today. We need a stance on upgrade blocking bugs, as they can undermine our ease of upgrade goals.

Obviously we cannot fix all. But I want to make sure everyone's understanding is aligned.

Dhaulagiri:
This is also fixed in 2.0.  The difference seems to be in this particular case that `current-when` is an object in 1.13 wheres in >= 2.0 (and presumably 1.12), it is a string.

stefanpenner:
A quick glance suggests we are leaking MutCells (or w/e they are called)

![screen shot 2015-09-11 at 2 46 09 pm](https://cloud.githubusercontent.com/assets/1377/9827113/fcf56228-5893-11e5-989a-d853792ee7d6.png)

stefanpenner:
ya we are leaking it and potentially others [here](https://github.com/emberjs/ember.js/blob/v1.13.9/packages/ember-routing-views/lib/views/link.js#L423-L432)

stefanpenner:
@wycats hinted at `get-value`

a quick check confirms `getValue` instead of direct attrs interaction will solve the issue.

```js
[...]

PR #14848, opened 2017-01-18
[BUGFIX beta] Allow boolean values for current-when

whatthewhat:
[Quote](https://github.com/emberjs/ember.js/blob/v2.12.0/packages/ember-glimmer/lib/components/link-to.js#L157-L158) from the docs for `current-when`:

```
A link will be active if `current-when` is `true` or the current
route is the route this link would transition to.
```

But currently using a boolean value produces a TypeError.

Looks like this might have been broken since 1.13 and https://github.com/emberjs/ember.js/pull/12344 did not seem to actually fix this particular issue.

Related issues: https://github.com/emberjs/ember.js/issues/12512, https://github.com/emberjs/ember.js/pull/12630 (fix was not merged), https://github.com/emberjs/ember.js/issues/12296

### DIFF
--- a/packages/ember-glimmer/lib/components/link-to.js
+++ b/packages/ember-glimmer/lib/components/link-to.js
@@ -549,8 +549,12 @@ const LinkComponent = EmberComponent.extend({
     let routing = get(this, '_routing');
     let models = get(this, 'models');
     let resolvedQueryParams = get(this, 'resolvedQueryParams');
-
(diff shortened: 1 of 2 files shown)


whatthewhat:
Rebased and fixed merge conflicts, bump 😉

rwjblue:
Thank you!! Sorry for the delay.

PR #14850, opened 2017-01-18
add test verifying that registering an unknown service raises an exception

GavinJoyce:
This now passes, but didn't in Ember 2.10. I git bisected the test and it passes since https://github.com/emberjs/ember.js/pull/14360

### DIFF
--- a/packages/ember-glimmer/tests/integration/components/curly-components-test.js
+++ b/packages/ember-glimmer/tests/integration/components/curly-components-test.js
@@ -2483,6 +2483,18 @@ moduleFor('Components test: curly components', class extends RenderingTest {
     this.assertText('Jackson');
   }
 
+  ['@test injecting an unknown service raises an exception'](assert) {
+    this.registerComponent('foo-bar', {
+      ComponentClass: Component.extend({
+        missingService: inject.service()
+      })
+    });
+
+    expectAssertion(() => {
+      this.render('{{foo-bar}}');
+    }, 'Attempting to inject an unknown injection: \'service:missingService\'');
+  }
+
   ['@test can access `actions` hash via `_actions` [DEPRECATED]']() {
     let component;
 
--- a/packages/container/lib/container.js
+++ b/packages/container/lib/container.js
@@ -405,7 +405,9 @@ function buildInjections(/* container, ...injections */) {
       }
     }
 
-    container.registry.validateInjections(injections);
+    runInDebug(() => {
+      container.registry.validateInjections(injections);
+    });
 
     for (let i = 0; i < injections.length; i++) {
       injection = injections[i];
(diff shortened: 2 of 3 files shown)


packages/ember-glimmer/tests/integration/components/curly-components-test.js
```diff
@@ -2483,6 +2483,18 @@ moduleFor('Components test: curly components', class extends RenderingTest {
...
   }
 
+  ['@test injecting an unknown service raises an exception']() {
+    this.registerComponent('foo-bar', {
+      ComponentClass: Component.extend({
+        missingService: inject.service()
+      })
+    });
+
+    throws(() => {
```
rwjblue: Should use `expectAssertion` here (otherwise it will fail for prod builds).

rwjblue (changes requested):

packages/ember-glimmer/tests/integration/components/curly-components-test.js
GavinJoyce: apparently not: https://github.com/emberjs/ember.js/blob/aeeddf7c00b42a340494465252481b5414d87009/packages/container/lib/registry.js#L712

Should this be an assertion?

rwjblue: That `validateInjections` method is only called from here: 
https://github.com/emberjs/ember.js/blob/9b3d581e9b706ae636691b1fff199a86c793b53d/packages/container/lib/container.js#L505-L513

Which is inside of a `runInDebug`...

GavinJoyce: ah, thanks

GavinJoyce: It's also called from here, which doesn't seem to be wrapped in a `runInDebug`:

https://github.com/emberjs/ember.js/blob/9b3d581e9b706ae636691b1fff199a86c793b53d/packages/container/lib/container.js#L408

GavinJoyce: I'm going to wrap the `validateInjections` in a `runInDebug` and change the `Error` to an assert

GavinJoyce:
@rwjblue I believe that this is ready to go (I think the failing CI is just sauce labs flakiness).

I've wrapped a `container.registry.validateInjections(injections)` in a `runInDebug` so we get a bonus performance improvement

rwjblue (approved):
LGTM, I restarted the sauce labs job.

GavinJoyce:
It's 🍏, thanks

spruce:
Hey,
I just ran into this problem with my addon ember-rapid-forms. Where it is possible to to use i18n but not a requirement. But with this change it fails the tests. Is there a way to walk around this issue. Maybe even a better way to support it. I have no better idea. Could you help? See here: https://travis-ci.org/piceaTech/ember-rapid-forms/jobs/207164170#L593

GavinJoyce:
@spruce this should help you: https://github.com/RSSchermer/ember-multiselect-checkboxes/pull/41

rwjblue:
https://github.com/piceaTech/ember-rapid-forms/pull/153

PR #14853, opened 2017-01-20
Update "handlebars" to v4.0.6

Turbo87:
related to https://github.com/tildeio/glimmer/pull/397

/cc @rwjblue

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -274,13 +274,16 @@ function qunit() {
 }
 
 function handlebarsFix() {
-  var HANDLEBARS_UTIL = /\/utils.js$/;
+  var HANDLEBARS_PARSER = /\/parser.js$/;
   return {
     load: function(id) {
-      if (HANDLEBARS_UTIL.test(id)) {
+      if (HANDLEBARS_PARSER.test(id)) {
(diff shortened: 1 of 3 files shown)


ember-cli-build.js
```diff
@@ -274,13 +274,16 @@ function qunit() {
...
   return {
     load: function(id) {
-      if (HANDLEBARS_UTIL.test(id)) {
+      if (HANDLEBARS_PARSER.test(id)) {
         var code = fs.readFileSync(id, 'utf8');
         return {
-          code: code.replace(/export var isFunction/, 'export { isFunction }'),
+          code: code
+            .replace('exports.__esModule = true;', '')
+            .replace('exports[\'default\'] = handlebars;', 'export default handlebars;'),
```
chadhietala: I'm wondering if we can just open up a PR on the Handlebars repo to distribute the ES6 source, so we can just use it wholesale instead of re-writing it.

locks: Did this get updated to use the ES6 distributed files?

Turbo87: Not yet, because there is no release yet that distributes it as ES6. It's only on master so far.

PR #14856, opened 2017-01-20
Run backtracking assertion only in debug mode

elwayman02:
This wraps the `assertNotRendered` check in `runInDebug` so that it won't execute in production, even when either of the flags are enabled.

### DIFF
--- a/packages/ember-metal/lib/property_events.js
+++ b/packages/ember-metal/lib/property_events.js
@@ -113,10 +113,12 @@ function propertyDidChange(obj, keyName, _meta) {
 
   markObjectAsDirty(meta, keyName);
 
-  if (isEnabled('ember-glimmer-detect-backtracking-rerender') ||
-    isEnabled('ember-glimmer-allow-backtracking-rerender')) {
-    runInDebug(() => { assertNotRendered(obj, keyName, meta) });
-  }
+  runInDebug(() => {
+    if (isEnabled('ember-glimmer-detect-backtracking-rerender') ||
+      isEnabled('ember-glimmer-allow-backtracking-rerender')) {
+      assertNotRendered(obj, keyName, meta)
+    }
+  });
 }
 
 let WILL_SEEN, DID_SEEN;
--- a/packages/ember-metal/lib/property_events.js
+++ b/packages/ember-metal/lib/property_events.js
@@ -1,4 +1,5 @@
 import { guidFor, symbol } from 'ember-utils';
+import { runInDebug } from 'ember-metal';
 import {
   peekMeta
 } from './meta';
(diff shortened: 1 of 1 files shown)


packages/ember-metal/lib/property_events.js
```diff
@@ -113,8 +114,8 @@ function propertyDidChange(obj, keyName, _meta) {
   markObjectAsDirty(meta, keyName);
 
   if (isEnabled('ember-glimmer-detect-backtracking-rerender') ||
-      isEnabled('ember-glimmer-allow-backtracking-rerender')) {
-    assertNotRendered(obj, keyName, meta);
+    isEnabled('ember-glimmer-allow-backtracking-rerender')) {
+    runInDebug(() => { assertNotRendered(obj, keyName, meta) });
```
GavinJoyce: perhaps move the `runInDebug` to wrap the `if` statement too?

elwayman02: Seems reasonable

locks (approved):

Turbo87 (approved):

packages/ember-metal/lib/property_events.js
```diff
@@ -112,10 +113,12 @@ function propertyDidChange(obj, keyName, _meta) {
 
   markObjectAsDirty(meta, keyName);
 
-  if (isEnabled('ember-glimmer-detect-backtracking-rerender') ||
+  runInDebug(() => {
+    if (isEnabled('ember-glimmer-detect-backtracking-rerender') ||
       isEnabled('ember-glimmer-allow-backtracking-rerender')) {
-    assertNotRendered(obj, keyName, meta);
-  }
+      assertNotRendered(obj, keyName, meta)
```
rwjblue: So I dug into this a bit, and it turns out that this isn't the right fix. Specifically, in the case where `ember-glimmer-allow-backtracking-rerender` is enabled we have to actually call `assertNotRendered` or we will not trigger a rerender when we hit this backtracking.  In other words, wrapping this in `runInDebug` does avoid an error (due to a bug in `assertNotRendered`) but it also breaks the ability for rerendering to happen.


I submitted https://github.com/emberjs/ember.js/pull/14876 to address.

rwjblue (changes requested):

rwjblue:
I'm going to close this in favor of the refactoring in https://github.com/emberjs/ember.js/pull/14876.  Thank you for pointing out the issue and taking a stab at fixing!

vvohra:
@rwjblue we're trying to fix the issue in this ticket for 2.10.2. I see that your PR is against master and includes some other refactorings. It's hard to figure out if there's a simpler fix to get this working in 2.10.2.

rwjblue:
Its likely that my commit in https://github.com/emberjs/ember.js/pull/14876 applies cleanly to release-2-10 branch.

vvohra:
You're probably right, let me try it once more and run through tests.

PR #14858, opened 2017-01-21
Fix Rubygems publishing script

locks:
cc @rwjblue


rwjblue:
Will test locally, but seems good to me.

chancancode:
:+1:

chancancode:
Does it need to trim the trailing newline?

locks:
@chancancode seems fine here. I think `cat` reads the trailing newline as the file terminator or somesuch?

```bash
bash-3.2$ echo ember-source-`cat VERSION`.gem
ember-source-2.12.0.alpha.1.gem
```

PR #14860, opened 2017-01-21
[BUGFIX beta] Add back mainContext to loader #14859

### LINKED ISSUE #14859: `mainContext` is undefined in ember-environment/lib/global.js
After the ESLint warning cleanup the `mainContext` is `undefined` and brake the `ember-template-compiler.js` compilation in grunt environment.

Previous conversation about this issue:
https://github.com/emberjs/ember.js/commit/e24e403656b9de62d55b1a3111e4eb309276d4b8#diff-5b2ea89606fd2097e9e17d66ae54ed51L2

Opened issue in `grunt-ember-templates` repo: https://github.com/dgeb/grunt-ember-templates/issues/94

cc @Turbo87 @andyhot

zoltan-nz:
Fixing #14859

After the ESLint warning cleanup the `mainContext` is undefined and brake the ember-template-compiler.js compilation in grunt environment.

This PR removes the rest of the `mainContext` reference from the `global.js`.

Previous conversation about this issue:
e24e403#diff-5b2ea89606fd2097e9e17d66ae54ed51L2

Opened issue in grunt-ember-templates repo: dgeb/grunt-ember-templates#94

cc @Turbo87 @andyhot

### DIFF
--- a/packages/loader/lib/index.js
+++ b/packages/loader/lib/index.js
@@ -1,4 +1,5 @@
 var enifed, requireModule, Ember;
+var mainContext = this; // Used in ember-environment/lib/global.js
 
 (function() {
   var isNode = typeof window === 'undefined' &&


packages/ember-environment/lib/global.js
```diff
@@ -14,5 +14,4 @@ function checkElementIdShadowing(value) {
 export default checkGlobal(checkElementIdShadowing(typeof global === 'object' && global)) ||
                checkGlobal(typeof self === 'object' && self) ||
                checkGlobal(typeof window === 'object' && window) ||
-               mainContext || // set before strict mode in Ember loader/wrapper
```
rwjblue: Can you bring this back, and add it back to the correct location in `loader.js`? The ESLint cleanup PR should not have removed it (and I missed that it did when I reviewed).

rwjblue (changes requested):
Thanks for working on this! We need to bring back the line that was removed in the loader. If you can update then I'm happy to land this...

packages/ember-environment/lib/global.js
Turbo87: Sorry about that. Can we add a comment to it this time describing what the unused variable does?

andyhot: And also add a `// eslint-disable-next-line no-unused-vars` line before the mainContext definition in packages/loader/lib/index.js (even though no-unused-vars is globally off for now)

Still, i'm a bit puzzled by the global.js code... If mainContext is defined nowhere, the code will still fail. And if it is, the next line `new Function('return this')()` will never run. 

Would it make sense to replace `mainContext` with `(typeof mainContext === 'object' && mainContext)` ?

zoltan-nz:
@rwjblue @Turbo87 `mainContext` added back to the loader with a tiny comment.

rwjblue:
Thank you!

PR #14861, opened 2017-01-21
Phase 2 of public Router service - adding replaceWith

scalvert:
This PR represents the second phase of the public router service described in [this RFC](https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md).

It extends the API with:

- `replaceWith`

**Status**: Work in progress

**Changes**:

- Added `replaceWith` method in public router service
- Added tests

How to test drive this PR:

- yarn start
- Open up 2 tabs pointing to http://localhost:4200/tests/index.html, one with Enable Opt Features checked, ensure tests pass

### DIFF
--- a/packages/ember-routing/lib/services/router.js
+++ b/packages/ember-routing/lib/services/router.js
@@ -42,7 +42,28 @@ const RouterService = Service.extend({
      @public
    */
   transitionTo() {
-    this.router.transitionTo(...arguments);
+    return this.router.transitionTo(...arguments);
+  },
+
+  /**
+     Transition into another route while replacing the current URL, if possible.
+     The route may be either a single route or route path:
+
+     See [Route.replaceWith](http://emberjs.com/api/classes/Ember.Route.html#method_replaceWith) for more info.
+
+     @method replaceWith
+     @category ember-routing-router-service
+     @param {String} name the name of the route or a URL
+     @param {...Object} models the model(s) or identifier(s) to be used while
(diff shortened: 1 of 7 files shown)


rwjblue:
Looks good to me, why is it still WIP (what things are you still looking to add)?

scalvert:
As per our previous discussions, I wasn't sure if we wanted to try to add more to this or whether this was good to go as is. I'd prefer to keep these additions as small as possible so we can iterate faster.

Thoughts?

rwjblue:
Yep, agreed.  Lets keep the PR's small and focused. 

Per our chat in slack I think you have a few changes to make (WRT to transitions in the `visit.then` callback instead of in various `afterModel` hooks), but this should be good to go once that is done.

packages/ember-routing/lib/services/router.js
```diff
@@ -43,6 +43,27 @@ const RouterService = Service.extend({
...
+     @param {...Object} models the model(s) or identifier(s) to be used while
+       transitioning to the route.
+     @param {Object} [options] optional hash with a queryParams property
+       containing a mapping of query parameters
+     @return {Transition} the transition object associated with this
+       attempted transition
+     @public
+   */
+  replaceWith() {
+    this.router.replaceWith(...arguments);
```
rwjblue: We should `return` here I believe...

rwjblue:
Awesome job, keep it up? :clap:

PR #14866, opened 2017-01-23
Use eslint plugin for custom eslint rules.

rwjblue:
When emberjs-build is linked, the eslint resolution rules do not work
properly so we must fallback to manually providing the rules paths (as
we did before).

When emberjs-build is not linked, this follows the standard/simple eslint
pattern (and avoids requiring editor specific hacks to get linting working).


.eslintrc.js
```diff
@@ -29,11 +42,23 @@ module.exports = {
...
     // temporarily disabled
     'no-unused-vars': 'off',
     'comma-dangle': 'off',
   },
 };
+
+if (isEmberJSBuildLinked()) {
+  delete options.plugins;
+
+  options.rules['require-yuidoc-access'] = options.rules['ember-internal/require-yuidoc-access'];
```
Turbo87: @rwjblue can we put that in a loop that just does that for all rules starting with `ember-internal`?

Turbo87:
@rwjblue seems fine to me. can you double check that https://github.com/Turbo87/eslint-plugin-ember-internal is identical to the rules we had in here?

also I'd like to make sure that we agree that this shouldn't be a permanent solution. I'm still hoping that ESLint will improve the situation on their side (see https://github.com/eslint/eslint/issues/3458)

.eslintrc.js
rwjblue: Sure

rwjblue:
> can you double check that https://github.com/Turbo87/eslint-plugin-ember-internal is identical to the rules we had in here?

Yep, just confirmed they are the same as what is being deleted here.

> also I'd like to make sure that we agree that this shouldn't be a permanent solution. I'm still hoping that ESLint will improve the situation on their side (see eslint/eslint#3458)

Confirm, I think it sucks.

PR #14868, opened 2017-01-23
Fix import for glimmer-runtime -> @glimmer/runtime.

rwjblue:


chadhietala:
Unclear how this even worked before.

rwjblue:
@chadhietala - It didn't, I landed https://github.com/emberjs/ember.js/pull/14763 (which was green from before https://github.com/emberjs/ember.js/pull/14781 landed), but when I merged https://github.com/emberjs/ember.js/pull/14763 after https://github.com/emberjs/ember.js/pull/14781 it failed the build.

PR #14870, opened 2017-01-24
Resolve no-unused-vars, no-undef eslint warnings

andyhot:


rwjblue:
Does this resolve all of the warnings for `no-unused-vars` / `no-undef`?  If it does, we should consider having those rules trigger errors so that we do not regress in future iterations...

andyhot:
I'll push an update to this PR that handles all of them (including the tests).

I'm setting `no-unused-vars` to ``["error", { "args": "none" }]` so that it doesn't report unused arguments but I plan to explore this in an upcoming commit (or PR)

andyhot:
I'd need some guidance on how to continue.

* After https://github.com/emberjs/ember.js/pull/14870/commits/44fe179903814c1877163ed7d21497d0d7793626 i see only 2 remaining eslint errors 
`
/ember.js/packages/loader/lib/index.js
  46:5  error  Move function declaration to function body root  no-inner-declarations
  54:5  error  Move function declaration to function body root  no-inner-declarations
`

andyhot:
I'd need some guidance on how to continue.

* After https://github.com/emberjs/ember.js/pull/14870/commits/aef5325edc565a04a493f46f100e1e5ddd0fb9c1 i see only 2 remaining eslint errors -  https://github.com/emberjs/ember.js/blob/master/packages/loader/lib/index.js#L43 and https://github.com/emberjs/ember.js/blob/master/packages/loader/lib/index.js#L51 (both `Move function declaration to function body root  no-inner-declarations`)
Perhaps we want that error suppressed there?
* I set `no-unused-vars': ["error", { "args": "after-used" }]`  to see the unused arguments issues. I now get 57 errors (and in fact 392 more in the tests). Most of those 57 errors are of the type
`  filterBy(key, value) {
    return this.filter(iter.apply(this, arguments));
  }`
and I not in favor of 'fixing' those. It looks like keeping `no-unused-vars to ["error", { "args": "none" }]` works best for ember.js

andyhot:
This has long become obsolete

PR #14871, opened 2017-01-24
Remove unneeded check as `data-ember-action` no longer contains a value

GavinJoyce:
The Glimmer 2 action helper [doesn't set a value for `data-ember-action`](https://github.com/emberjs/ember.js/blob/b0801b77dce8c546074322f3fed1c9f7ab4df2bb/packages/ember-glimmer/lib/modifiers/action.js#L223-L224), so we no longer need to compare it to `""`.

### DIFF
--- a/packages/ember-views/lib/system/event_dispatcher.js
+++ b/packages/ember-views/lib/system/event_dispatcher.js
@@ -218,33 +218,23 @@ export default EmberObject.extend({
     });
 
     rootElement.on(`${event}.ember`, '[data-ember-action]', evt => {
-      let actionId = jQuery(evt.currentTarget).attr('data-ember-action');
-      let actions = ActionManager.registeredActions[actionId];
+      let attributes = evt.currentTarget.attributes;
+      let attributeCount = attributes.length;
+      let actions = [];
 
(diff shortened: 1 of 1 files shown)


GavinJoyce:
this is 🍏 and ready for review

packages/ember-views/lib/system/event_dispatcher.js
```diff
@@ -218,33 +218,23 @@ export default EmberObject.extend({
...
-        for (let i = 0; i < attributeCount; i++) {
-          let attr = attributes.item(i);
-          let attrName = attr.name;
-
-          if (attrName.indexOf('data-ember-action-') === 0) {
-            actions = actions.concat(ActionManager.registeredActions[attr.value]);
-          }
+      let attributes = evt.currentTarget.attributes;
+      let attributeCount = attributes.length;
+      let actions = [];
```
stefanpenner: allocating this array even if we don't need too is a-bit strange. (not a big deal though)

stefanpenner: ya, i guess this really doesn't matter here :P

PR #14872, opened 2017-01-24
Simplify action event handler

GavinJoyce:
As [@rwjblue notes](https://github.com/emberjs/ember.js/pull/14872#discussion_r97623843), the pre-glimmer 2 implementation had a single `data-ember-action="someid"` and multiple actions would be registered on that id.

Glimmer 2 has a [unique action id per action registration](https://github.com/emberjs/ember.js/blob/b0801b77dce8c546074322f3fed1c9f7ab4df2bb/packages/ember-glimmer/lib/modifiers/action.js#L223-L224) so we no longer need to store the registered actions in an array.

TODO:

 * [x] [check if we're breaking any addons](https://emberobserver.com/code-search?codeQuery=registeredActions) (just one it seems: [ember-native-dom-event-dispatcher](https://github.com/rwjblue/ember-native-dom-event-dispatcher/blob/d970bece76583f8932fc92d664a38c7d002fb6df/addon/index.js#L76-L114))

### DIFF
--- a/packages/ember-views/lib/system/event_dispatcher.js
+++ b/packages/ember-views/lib/system/event_dispatcher.js
@@ -219,30 +219,17 @@ export default EmberObject.extend({
 
     rootElement.on(`${event}.ember`, '[data-ember-action]', evt => {
       let attributes = evt.currentTarget.attributes;
-      let attributeCount = attributes.length;
-      let actions = [];
 
-      for (let i = 0; i < attributeCount; i++) {
+      for (let i = 0; i < attributes.length; i++) {
         let attr = attributes.item(i);
         let attrName = attr.name;
 
-        if (attrName.indexOf('data-ember-action-') === 0) {
-          actions = actions.concat(ActionManager.registeredActions[attr.value]);
-        }
-      }
+        if (attrName.lastIndexOf('data-ember-action-', 0) !== -1) {
+          let action = ActionManager.registeredActions[attr.value];
 
-      // We have to check for actions here since in some cases, jQuery will trigger
-      // an event on `removeChild` (i.e. focusout) after we've already torn down the
-      // action handlers for the view.
-      if (actions.length === 0) {
-        return;
-      }
-
-      for (let index = 0; index < actions.length; index++) {
-        let action = actions[index];
-
-        if (action && action.eventName === eventName) {
-          return action.handler(evt);
+          if (action.eventName === eventName) {
+            action.handler(evt);
+          }
         }
       }
     });
--- a/packages/ember-glimmer/lib/modifiers/action.js
+++ b/packages/ember-glimmer/lib/modifiers/action.js
@@ -38,34 +38,16 @@ export let ActionHelper = {
 
   registerAction(actionState) {
     let { actionId } = actionState;
-    let actions = ActionManager.registeredActions[actionId];
 
-    if (!actions) {
-      actions = ActionManager.registeredActions[actionId] = [];
-    }
-
-    actions.push(actionState);
+    ActionManager.registeredActions[actionId] = actionState;
 
     return actionId;
   },
 
   unregisterAction(actionState) {
     let { actionId } = actionState;
-    let actions = ActionManager.registeredActions[actionId];
 
-    if (!actions) {
-      return;
-    }
-
-    let index = actions.indexOf(actionState);
-
-    if (index !== -1) {
-      actions.splice(index, 1);
-    }
-
-    if (actions.length === 0) {
-      delete ActionManager.registeredActions[actionId];
-    }
+    delete ActionManager.registeredActions[actionId];
   }
 };
 


packages/ember-views/lib/system/event_dispatcher.js
```diff
@@ -218,43 +218,22 @@ export default EmberObject.extend({
...
-      if (actionId === '') {
-        let attributes = evt.currentTarget.attributes;
-        let attributeCount = attributes.length;
+      for (let i = 0; i < attributeCount; i++) {
+        let attr = attributes.item(i);
+        let attrName = attr.name;
 
-        actions = [];
+        if (attrName.indexOf('data-ember-action-') === 0) {
+          //TODO: GJ: why does `registeredActions` contains arrays? are they always of length 1?
```
rwjblue: In the pre-glimmer2 implementation it was an array because we only tagged the element with `data-ember-action="someid"` and that element could have multiple actions:

```hbs
<button {{action 'foo' on="click"}} {{action 'bar' on="doubleClick"}}></button>
```

In the glimmer 2 implementation, we have maintained it as an array but it can only ever have one entry.  

Some references:

* https://github.com/emberjs/ember.js/blob/c998c34de5eec1e5df3453b98c4c056e64b0dff1/packages/ember-glimmer/lib/modifiers/action.js#L205
* https://github.com/emberjs/ember.js/blob/c998c34de5eec1e5df3453b98c4c056e64b0dff1/packages/ember-glimmer/lib/modifiers/action.js#L43-L47

I think that `registerAction` can be simplified to something like:

```js
  registerAction(actionState) {
    let { actionId } = actionState;

    ActionManager.registeredActions[actionId] = actionState;

    return actionId;
  },
```

packages/ember-views/lib/system/event_dispatcher.js
```diff
@@ -218,43 +218,22 @@ export default EmberObject.extend({
...
-      // be coalesced.
-      if (actionId === '') {
-        let attributes = evt.currentTarget.attributes;
-        let attributeCount = attributes.length;
+      for (let i = 0; i < attributeCount; i++) {
+        let attr = attributes.item(i);
+        let attrName = attr.name;
 
-        actions = [];
+        if (attrName.indexOf('data-ember-action-') === 0) {
```
rwjblue: Since we only care about the attribute starting with the given string, I think you could do:

```js
if (attrName.lastIndexOf('data-ember-action-', 0) !== -1) {

}
```

packages/ember-views/lib/system/event_dispatcher.js
```diff
@@ -218,43 +218,22 @@ export default EmberObject.extend({
...
+      let attributeCount = attributes.length;
 
-      // In Glimmer2 this attribute is set to an empty string and an additional
-      // attribute it set for each action on a given element. In this case, the
-      // attributes need to be read so that a proper set of action handlers can
-      // be coalesced.
-      if (actionId === '') {
-        let attributes = evt.currentTarget.attributes;
-        let attributeCount = attributes.length;
+      for (let i = 0; i < attributeCount; i++) {
```
rwjblue: I don't think we need to cache the length here (we are not mutating AFAICT).

rwjblue:
ready for a rebase also

packages/ember-views/lib/system/event_dispatcher.js
GavinJoyce: Thanks for the context, I'll update `registerAction` to remove the arrays and simplify the action handler

GavinJoyce:
@rwjblue thanks for the feedback, this is 🍏 now.

It appears that [only one addon](https://emberobserver.com/code-search?codeQuery=registeredActions) uses `ActionManager.registeredActions`:

 * [ember-native-dom-event-dispatcher](https://github.com/rwjblue/ember-native-dom-event-dispatcher/blob/d970bece76583f8932fc92d664a38c7d002fb6df/addon/index.js#L76-L114)

packages/ember-views/lib/system/event_dispatcher.js
```diff
@@ -219,30 +219,17 @@ export default EmberObject.extend({
...
-      }
-
-      for (let index = 0; index < actions.length; index++) {
-        let action = actions[index];
+        if (attrName.lastIndexOf('data-ember-action-', 0) !== -1) {
+          let action = ActionManager.registeredActions[attr.value];
 
-        if (action && action.eventName === eventName) {
-          return action.handler(evt);
+          if (action.eventName === eventName) {
```
GavinJoyce: is it ok to remove the `if(action &&` guard here? Do we care about supporting non existent actions?

rwjblue: I think this is probably fine.  I can't see how we would end up here with an `action` that is falsey without having already screwed something majorly up elsewhere.

rwjblue (approved):

PR #14873, opened 2017-01-24
Add @since tag to doc for the hash helper

ziggythehamster:
This is a simple change to the documentation for the hash helper.


rwjblue:
Awesome, thank you!

PR #14878, opened 2017-01-26
Fix yuidoc package paths

### LINKED ISSUE #14874: Ember.RSVP removed from emberjs.com/api documentation
I discovered `Ember.RSVP` was missing from all lists when trying to access http://emberjs.com/api.  This update seems to have happened here:

https://github.com/emberjs/emberjs.github.com/commit/0702657c9f4b4dd051ffb38877b235bad589ffe1

GavinJoyce:
Fixes https://github.com/emberjs/ember.js/issues/14874.

This fixes one of the issues that we're seeing with the API docs. The RSVP classes are now displayed in the left hand nav:

**before**

screen shot 2017-01-26 at 21 16 08

**after**

screen shot 2017-01-26 at 21 16 18

### DIFF
--- a/yuidoc.json
+++ b/yuidoc.json
@@ -17,10 +17,10 @@
       "packages/ember-testing/lib",
       "packages/ember-extension-support/lib",
       "packages/container/lib",
-      "bower_components/rsvp/lib",
-      "bower_components/router.js/lib/router"
+      "node_modules/rsvp/lib",
+      "node_modules/router_js/lib/router"
     ],
-    "exclude": "vendor,bower_components/router.js/lib/router/handler-info,bower_components/router.js/lib/router/transition-intent,bower_components/router.js/lib/router/router.js,bower_components/router.js/lib/router/transition-intent.js,bower_components/router.js/lib/router/transition-state.js,bower_components/router.js/lib/router/unrecognized-url-error.js,bower_components/router.js/lib/router/utils.js",
+    "exclude": "vendor,node_modules/router_js/lib/router/handler-info,node_modules/router_js/lib/router/transition-intent,node_modules/router_js/lib/router/router_js,node_modules/router_js/lib/router/transition-intent.js,node_modules/router_js/lib/router/transition-state.js,node_modules/router_js/lib/router/unrecognized-url-error.js,node_modules/router_js/lib/router/utils.js",
     "outdir":   "docs",
     "preprocessor": "bin/feature-flag-yuidoc-filter"
   }


GavinJoyce:
We get lots of warnings when running `npm run docs`:

```
info: Scanning for yuidoc.json file.
warn: Skipping node_modules directory while scanning for yuidoc.json
info: Loading package.json data from: /Users/gavin/dev/opensource/ember.js/package.json
info: Loading yuidoc.json data from: /Users/gavin/dev/opensource/ember.js/yuidoc.json
info: YUIDoc Starting from: node_modules/router_js/lib,node_modules/rsvp/lib,packages/container/lib,packages/ember-application/lib,packages/ember-debug/lib,packages/ember-extension-support/lib,packages/ember-glimmer/lib,packages/ember-metal/lib,packages/ember-routing/lib,packages/ember-runtime/lib,packages/ember-testing/lib,packages/ember-utils/lib,packages/ember-views/lib,packages/ember/lib
warn: replacing incorrect tag: returns with return:  packages/container/lib/container.js:219
warn: replacing incorrect tag: returns with return:  packages/ember-routing/lib/system/route.js:810
warn: Missing item type:  node_modules/router_js/lib/router/router.js:90
warn: 		The main entry point into the router. The API is essentially
the same as the `map` method in `route-recognizer`.

This method extracts the String handler at the last `.to()`
call and uses it as the name of the whole route.
warn: Missing item type:  node_modules/router_js/lib/router/router.js:163
warn: 		Clears the current and target route handlers and triggers exit
on each of them starting at the leaf and traversing up through
its ancestors.
warn: Missing item type:  node_modules/router_js/lib/router/router.js:183
warn: 		var handler = handlerInfo.handler;
The entry point for handling a change to the URL (usually
via the back and forward button).

Returns an Array of handlers and the parameters associated
with those parameters.
warn: Missing item type:  node_modules/router_js/lib/router/router.js:204
warn: 		Hook point for updating the URL.
warn: Missing item type:  node_modules/router_js/lib/router/router.js:213
warn: 		Hook point for replacing the current URL, i.e. with replaceState

By default this behaves the same as `updateURL`
warn: Missing item type:  node_modules/router_js/lib/router/router.js:224
warn: 		Transition into the specified named route.

If necessary, trigger the exit callback on any handlers
that are no longer represented by the target route.
warn: Missing item type:  node_modules/router_js/lib/router/router.js:260
warn: 		Identical to `transitionTo` except that the current URL will be replaced
if possible.

This method is intended primarily for use with `replaceState`.
warn: Missing item type:  node_modules/router_js/lib/router/router.js:272
warn: 		Take a named route and context objects and generate a
URL.
warn: Missing item type:  node_modules/router_js/lib/router/router.js:376
warn: 		Hook point for logging transition status updates.
warn: Missing item type:  node_modules/router_js/lib/router/router.js:384
warn: Missing item type:  node_modules/router_js/lib/router/router.js:402
warn: Missing item type:  node_modules/router_js/lib/router/router.js:482
warn: Missing item type:  node_modules/router_js/lib/router/router.js:523
warn: Missing item type:  node_modules/router_js/lib/router/router.js:658
warn: Missing item type:  node_modules/router_js/lib/router/router.js:709
warn: Missing item type:  node_modules/router_js/lib/router/transition.js:315
warn: Missing item type:  node_modules/router_js/lib/router/transition.js:326
warn: Missing item type:  node_modules/router_js/lib/router/utils.js:14
warn: 		Determines if an object is Promise by checking if it is "thenable".
warn: Missing item type:  node_modules/router_js/lib/router/utils.js:33
warn: Missing item type:  node_modules/router_js/lib/router/utils.js:50
warn: Missing item type:  node_modules/router_js/lib/router/utils.js:66
warn: Missing item type:  packages/ember-application/lib/system/application-instance.js:88
warn: 		Overrides the base `EngineInstance._bootSync` method with concerns relevant
to booting application (instead of engine) instances.

This method should only contain synchronous boot concerns. Asynchronous
boot concerns should eventually be moved to the `boot` method, which
returns a promise.

Until all boot code has been made asynchronous, we need to continue to
expose this method for use *internally* in places where we need to boot an
instance synchronously.
warn: Missing item type:  packages/ember-application/lib/system/application-instance.js:139
warn: 		This hook is called by the root-most Route (a.k.a. the ApplicationRoute)
when it has finished creating the root View. By default, we simply take the
view and append it to the `rootElement` specified on the Application.

In cases like FastBoot and testing, we can override this hook and implement
custom behavior, such as serializing to a string and sending over an HTTP
socket rather than appending to DOM.
warn: Missing item type:  packages/ember-application/lib/system/application-instance.js:155
warn: 		Tells the router to start routing. The router will ask the location for the
current URL of the page to determine the initial URL to start routing to.
To start the app at a specific URL, call `handleURL` instead.
warn: Missing item type:  packages/ember-application/lib/system/application-instance.js:168
warn: Missing item type:  packages/ember-application/lib/system/application-instance.js:185
warn: 		Directs the router to route to a particular URL. This is useful in tests,
for example, to tell the app to start at a particular URL.
warn: Missing item type:  packages/ember-application/lib/system/application-instance.js:199
warn: Missing item type:  packages/ember-application/lib/system/application-instance.js:213
warn: 		Returns the current URL of the app instance. This is useful when your
app does not update the browsers URL bar (i.e. it uses the `'none'`
location adapter).
warn: Missing item type:  packages/ember-application/lib/system/application-instance.js:229
warn: 		Navigate the instance to a particular URL. This is useful in tests, for
example, or to tell the app to start at a particular URL. This method
returns a promise that resolves with the app instance when the transition
is complete, or rejects if the transion was aborted due to an error.
warn: Missing item type:  packages/ember-application/lib/system/application.js:566
warn: 		Unfortunately, a lot of existing code assumes the booting process is
"synchronous". Specifically, a lot of tests assumes the last call to
`app.advanceReadiness()` or `app.reset()` will result in the app being
fully-booted when the current runloop completes.

We would like new code (like the `visit` API) to stop making this assumption,
so we created the asynchronous version above that returns a promise. But until
we have migrated all the code, we would have to expose this method for use
*internally* in places where we need to boot an app "synchronously".
warn: Missing item type:  packages/ember-application/lib/system/engine-instance.js:82
warn: 		Unfortunately, a lot of existing code assumes booting an instance is
synchronous – specifically, a lot of tests assume the last call to
`app.advanceReadiness()` or `app.reset()` will result in a new instance
being fully-booted when the current runloop completes.

We would like new code (like the `visit` API) to stop making this
assumption, so we created the asynchronous version above that returns a
promise. But until we have migrated all the code, we would have to expose
this method for use *internally* in places where we need to boot an instance
synchronously.
warn: Missing item type:  packages/ember-application/lib/system/engine-instance.js:131
warn: Missing item type:  packages/ember-routing/lib/location/auto_location.js:107
warn: 		Called by the router to instruct the location to do any feature detection
necessary. In the case of AutoLocation, we detect whether to use history
or hash concrete implementations.
warn: Missing item type:  packages/ember-routing/lib/location/auto_location.js:231
warn: Missing item type:  packages/ember-routing/lib/location/util.js:1
warn: Missing item type:  packages/ember-routing/lib/location/util.js:16
warn: Missing item type:  packages/ember-routing/lib/location/util.js:25
warn: Missing item type:  packages/ember-routing/lib/location/util.js:105
warn: 		Replaces the current location, making sure we explicitly include the origin
to prevent redirecting to a different origin.
warn: Missing item type:  packages/ember-routing/lib/system/router.js:1180
warn: 		Finds the name of the substate route if it exists for the given route. A
substate route is of the form `route_state`, such as `foo_loading`.
warn: Missing item type:  packages/ember-routing/lib/system/router.js:1204
warn: 		Finds the name of the state route if it exists for the given route. A state
route is of the form `route.state`, such as `foo.loading`. Properly Handles
`application` named routes.
warn: Missing item type:  packages/ember-routing/lib/system/router.js:1229
warn: 		Determines whether or not a route has been defined by checking that the route
is in the Router's map and the owner has a registration for that route.
warn: Missing item type:  packages/ember-runtime/lib/mixins/container_proxy.js:138
warn: Missing item type:  packages/ember-runtime/lib/mixins/controller.js:44
warn: Missing item type:  packages/ember-utils/lib/apply-str.js:1
warn: Missing item type:  packages/ember-utils/lib/guid.js:3
warn: 		Previously we used `Ember.$.uuid`, however `$.uuid` has been removed from
jQuery master. We'll just bootstrap our own uuid now.
warn: Missing item type:  packages/ember-utils/lib/guid.js:12
warn: 		Generates a universally unique identifier. This method
is used internally by Ember for assisting with
the generation of GUID's and other unique identifiers.
warn: Missing item type:  packages/ember-utils/lib/intern.js:1
warn: 		Strongly hint runtimes to intern the provided string.

When do I need to use this function?

For the most part, never. Pre-mature optimization is bad, and often the
runtime does exactly what you need it to, and more often the trade-off isn't
worth it.

Why?

Runtimes store strings in at least 2 different representations:
Ropes and Symbols (interned strings). The Rope provides a memory efficient
data-structure for strings created from concatenation or some other string
manipulation like splitting.

Unfortunately checking equality of different ropes can be quite costly as
runtimes must resort to clever string comparison algorithms. These
algorithms typically cost in proportion to the length of the string.
Luckily, this is where the Symbols (interned strings) shine. As Symbols are
unique by their string content, equality checks can be done by pointer
comparison.

How do I know if my string is a rope or symbol?

Typically (warning general sweeping statement, but truthy in runtimes at
present) static strings created as part of the JS source are interned.
Strings often used for comparisons can be interned at runtime if some
criteria are met.  One of these criteria can be the size of the entire rope.
For example, in chrome 38 a rope longer then 12 characters will not
intern, nor will segments of that rope.

Some numbers: http://jsperf.com/eval-vs-keys/8

Known Trick™
warn: Missing item type:  packages/ember/lib/index.js:143
warn: 		The context that Ember searches for namespace instances on.
warn: Found out dir, deleting: docs
info: Making out dir: docs
info: Parsed 291 files in 0.295 seconds
info: Building..
info: Compiling Templates
info: Making default directories: classes,elements,modules,files
(node:21706) DeprecationWarning: os.tmpDir() is deprecated. Use os.tmpdir() instead.
info: Copying Assets
info: Rendering and writing 10 modules pages.
info: Loading theme from /Users/gavin/dev/opensource/ember.js/node_modules/yuidocjs/themes/default/theme.json
info: Preparing index.html
info: Writing API Meta Data
info: Writing index.html
info: Finished writing module files
info: Rendering and writing 87 class pages.
info: Rendering and writing 0 class pages.
info: Finished writing element files
info: Rendering and writing 201 source files.
info: Finished writing class files
info: Rendering and writing 201 source files.
info: Finished writing source files
info: Finished writing source files
info: Finished writing 500 files in 5.544 seconds
```

PR #14881, opened 2017-01-27
Update index.js

locks:


Turbo87 (approved):

packages/ember-glimmer/lib/index.js
```diff
@@ -1,23 +1,19 @@
...
 
   ### Showing a property
 
   Templates manage the flow of an application's UI, and display state (through
   the DOM) to a user. For example, given a component with the property "name",
   that component's template can use the name in several ways:
 
-  ```javascript
-    // app/components/person.js
+  ```app/components/person.js
```
rwjblue: I do not believe this syntax work with our YUIDoc parser, does it?

locks: It should! http://emberjs.com/api/classes/Ember.Route.html#method_model https://github.com/emberjs/ember.js/blob/v2.11.0/packages/ember-routing/lib/system/route.js#L1481

PR #14883, opened 2017-01-27
[BUGFIX beta] simplify ActionModifierManager.update

GavinJoyce:
The `UpdateModifierOpcode.modifier` (`actionState`) is the same instance as the already registered `actionState` in `ActionHelper` so we don't need to register and unregister it, we can just mutate it directly.


rwjblue:
Seems good. I've seen and been annoyed by that TODO lately also. Thanks for cleaning this up!

PR #14897, opened 2017-02-01
[BUGFIX beta] Avoid reading from `this` in constructor.

rwjblue:


chancancode (approved):

GavinJoyce:
TIL, thanks

topaxi:
Curious, why? I want to TIL too :D

GavinJoyce:
@topaxi I believe by accessing `this.defRef` mid constructor, the runtime is forced to create two different shapes for this object. If we don't read from `this`, it only creates one

topaxi:
Oh, thanks! I always assumed this only happens when setting properties 😄

Issue #12791, opened 2016-01-09
Test.adapter.exception forces acceptance test to fail (RSVP error fails test)


pixelhandler:
My workaround is here: https://github.com/emberjs/guides/issues/1157 at first I thought it may be best to address with an update to testing guides, but not so sure that is the best way to fix.

**Steps to reproduce:**  

Here is an example repository with an Ember CLI app: https://github.com/pixelhandler/application-error-test
(Can't be done w/ a jsbin/fiddle/ember-twiddle - since this issue is with the testing env.)

Below are the steps I used to create the example app.

1. Create an app with one of the model hooks making an XHR request.
2. Create an application-error template, perhaps use `{{model.message}}` since the error will become the model via application-error route's setupController hook.
3. Create an acceptance test to test that the application-error template displays the error message.
4. In the acceptance test, mock the XHR - to return a 502 response. (I used Sinon.js with ember-sinon addon)
5. Write an assertion that checks the DOM, e.g. `visit('/'); andThen(function(){ assert.equal(find('.error-message').length(), 1); });`

**Expected Result:**

The test passes even though an error was thrown in the route's model hook by the mocked 502 response.

**Actual Result:**

The Assertion passes as the template is rendered, but the test fails since `Test.adapter.exception` is called. Which throws, and causes the test to fail - as a result of the "same" expected error that causes the application-error substate to render.

**Attempts to solve:**

- Tried 2.4.0-canary+10ab23d8 since a couple developer mentions this looks related: https://github.com/emberjs/ember.js/pull/12549 
- Setup a custom error for the XHR failure that has a `name` property of `TransitionAborted`. This is the code fails the test for the applicaiton-error.hbs template -> https://github.com/emberjs/ember.js/blob/master/packages/ember-runtime/lib/ext/rsvp.js#L66 (this works but is a hack)
- Reopened `Ember.Test.QUnitAdapter` and redefined `exception` as the example below:

Because an error is thrown and the env is `testing` - Ember RSVP calls `Test.adapter.exception`. Which fails the test; because there was an error (we expected an error to be handled by the application-error route/controller/template and it was). In the acceptance test, using `assert.throws` will not help since it's async.

I found that I had to reopen my test adapter in the acceptance test like so:

[*tests/acceptance/application-error-test.js*](https://github.com/pixelhandler/application-error-test/blob/master/tests/acceptance/application-error-test.js)

```js
import { test } from 'qunit';
import Ember from 'ember';
[...]

PR #14898, opened 2017-02-01
[BUGFIX] Changed backburner's error handler to use `dispatchError` instead of `onError`

### LINKED ISSUE #14864: Why does backburner call `Ember.onError` instead of `dispatchError`?
Recently we discovered a bug in our application that our tests should have caught. The error was actually a crash inside of the ember router that occurred when an action was sent to the router while it was transition from the `application_loading` route to another router. The details of this are not important as it was our mistake (I could probably provide a fiddle if desired). 

Despite this error, the tests continued on and actually passed. This left me with a bad feeling because our acceptance test use their own QUnitAdapter to ensure than errors cause a test failure.  So I decided to dig in to see what was going on. 

Here is a normal stack from an exception during the test:
```javascript
at Class.exception (https://localhost:4200/assets/tests.js:1042:27)
at Class.superWrapper [as exception] (https://localhost:4200/assets/vendor.js:54581:22)
at adapterDispatch (https://localhost:4200/assets/vendor.js:53410:13)
at Object.dispatchError (https://localhost:4200/assets/vendor.js:31831:7)
at onerrorDefault (https://localhost:4200/assets/vendor.js:45334:19)
at Object.trigger (https://localhost:4200/assets/vendor.js:72784:9)
at https://localhost:4200/assets/vendor.js:73668:26
at invokeWithOnError (https://localhost:4200/assets/vendor.js:15293:16)"
```

Here is the stack that failed silently:
```javascript
at representEmptyRoute (https://localhost:4200/assets/vendor.js:43088:7)
at Class._setOutlets (https://localhost:4200/assets/vendor.js:41851:22)
at invokeWithOnError (https://localhost:4200/assets/vendor.js:15293:16)
at Queue.flush (https://localhost:4200/assets/vendor.js:15352:9)
at DeferredActionQueues.flush (https://localhost:4200/assets/vendor.js:15476:15)
at Backburner.end (https://localhost:4200/assets/vendor.js:15546:23)
at Backburner.run (https://localhost:4200/assets/vendor.js:15660:16)
at Object.run (https://localhost:4200/assets/vendor.js:36459:27)
at Class.hash.success (https://localhost:4200/assets/buttercup.js:45341:31)"
```

The difference can be found by contrasting [RSVP's error handling](https://github.com/emberjs/ember.js/blob/b307387a7a3bad857498ed3ef6f11c2798ff81d6/packages/ember-runtime/lib/ext/rsvp.js#L19-L26) to in [Backburner's error handling](https://github.com/emberjs/ember.js/blob/b307387a7a3bad857498ed3ef6f11c2798ff81d6/packages/ember-metal/lib/run_loop.js#L31-L41).

You can see that RSVP's error handling logic will call [dispatchError](https://github.com/emberjs/ember.js/blob/b307387a7a3bad857498ed3ef6f11c2798ff81d6/packages/ember-metal/lib/error_handler.js#L28-L34) which will defer to a `TestAdapter` if set to do so. Unfortunately Backburner just calls `Ember.onError` directly. 

I don't really possess a whole lot of knowledge about ember-metal, but just looking at this I can't see any obvious reason why Backburner's error handling couldn't call `dispatchError` instead of directly calling `Ember.onError`. 

Any thoughts? I'd sleep much better knowing that these error were being routed to my test suite.

workmanw:
This is so that backburner errors can be caught by the `Test.adapter`. Fixes #14864 (more details about the reason for this change in the issue).

### DIFF
--- a/packages/ember-metal/lib/error_handler.js
+++ b/packages/ember-metal/lib/error_handler.js
@@ -34,6 +34,9 @@ export function dispatchError(error) {
 }
 
 // allows testing adapter to override dispatch
+export function getDispatchOverride() {
+  return dispatchOverride;
+}
 export function setDispatchOverride(handler) {
   dispatchOverride = handler;
 }
--- a/packages/ember-metal/lib/run_loop.js
+++ b/packages/ember-metal/lib/run_loop.js
@@ -2,7 +2,7 @@ import { GUID_KEY } from 'ember-utils';
 import { assert } from './debug';
 import { isTesting } from './testing';
 import {
-  getOnerror,
+  dispatchError,
   setOnerror
 } from './error_handler';
 import {
@@ -21,7 +21,7 @@ function onEnd(current, next) {
 
 const onErrorTarget = {
   get onerror() {
-    return getOnerror();
+    return dispatchError;
   },
   set onerror(handler) {
     return setOnerror(handler);
--- a/packages/ember-metal/tests/run_loop/onerror_test.js
+++ b/packages/ember-metal/tests/run_loop/onerror_test.js
@@ -1,5 +1,6 @@
 import run from '../../run_loop';
-import { setOnerror, getOnerror } from '../../error_handler';
+import { setOnerror, getOnerror, setDispatchOverride, getDispatchOverride } from '../../error_handler';
+import { isTesting, setTesting } from '../../testing';
 
 QUnit.module('system/run_loop/onerror_test');
 
@@ -23,13 +24,20 @@ QUnit.test('With Ember.onerror undefined, errors in Ember.run are thrown', funct
 QUnit.test('With Ember.onerror set, errors in Ember.run are caught', function () {
   let thrown = new Error('Boom!');
   let original = getOnerror();
+  let originalDispatchOverride = getDispatchOverride();
+  let originalIsTesting = isTesting();
 
   let caught;
   setOnerror(error => { caught = error; });
+  setDispatchOverride(null);
+  setTesting(false);
+
   try {
     run(() => { throw thrown; });
   } finally {
     setOnerror(original);
+    setDispatchOverride(originalDispatchOverride);
+    setTesting(originalIsTesting);
   }
 
   deepEqual(caught, thrown);
--- a/packages/ember-testing/tests/adapters_test.js
+++ b/packages/ember-testing/tests/adapters_test.js
@@ -12,9 +12,11 @@ QUnit.module('ember-testing Adapters', {
     originalQUnit = window.QUnit;
   },
   teardown() {
-    run(App, App.destroy);
-    App.removeTestHelpers();
-    App = null;
+    if (App) {
+      run(App, App.destroy);
+      App.removeTestHelpers();
+      App = null;
+    }
 
     Test.adapter = originalAdapter;
     window.QUnit = originalQUnit;
@@ -67,4 +69,19 @@ QUnit.test('Adapter is used by default (if QUnit is not available)', function() 
 
   ok(Test.adapter instanceof Adapter);
   ok(!(Test.adapter instanceof QUnitAdapter));
+});
+
+QUnit.test('With Ember.Test.adapter set, errors in Ember.run are caught', function () {
+  let thrown = new Error('Boom!');
+
+  let caught;
+  Test.adapter = QUnitAdapter.create({
+    exception(error) {
+      caught = error;
+    }
+  });
+
+  run(() => { throw thrown; });
+
+  deepEqual(caught, thrown);
 });


rwjblue:
Asking @krisselden for review, I believe he worked on the most recent refactors...

workmanw:
@krisselden Any chance you can squeeze in a review? Hopefully this is a very minor change and I'd be happy to do any additional leg work to get this pushed through. I feel pretty strongly about this change because I recently had a regression go to production that slide past our tests due to the exception not being in a promise handler.

stefanpenner:
Pinging kris again

rwjblue:
Thank you @krisselden!

workmanw:
Thanks!!! 👏

stefanpenner:
thanks @workmanw  sorry for letting this linger. (these PR slip through the cracks too easily)

workmanw:
@stefanpenner No worries at all, really. I know how these things go. And with any change to `ember-metal`, I'd much prefer to wait until the proper parties have reviewed it. :+1:

Btown6983:
image/webp,image/*,*/*;q=0.8

Thank You
Aaron Perkins

jasonmit:
Tests that were previously asserting errors were thrown are no longer working as expected:

```js
expect(() => {
  this.render(hbs`{{x-foo}}`);
}).to.throw('bar must be defined');
```

Now the throw is captured by mocha/chai and test fails before the assertion.  I'm also hearing this is happening with the qunit adapter as well.

Anyone else experiencing this?

workmanw:
@miguelcobain Also experienced this with ember-paper. I've been thinking all day about what the best solutions is.

I'm confident in saying that this change was the right thing, even though it broke some addons / apps. The reason I say that is 2 parts, 1) in your code snippet, you're inside of an integration test which has a runloop going and you're expecting an assertion/exception to bubble through the runloop all the way to the top. I think this expectation is problematic because the runloop has it's own error handling, it just was not working the same in all cases which brings me to my second point. 2) If your assertion/exception took place inside of a promise handler, it would have been swallowed in the very same way. Example:

```
 assert.throws(() => {
    Ember.run(() => {
      Ember.RSVP.resolve(() => { throw 'bar must be defined'; });
    });
}, /bar must be defined/);
```

That said, I'm definitely very sorry that this caused breakage. Believe me, I know how much it stinks to upgrade your ember version and have broken tests. I did expect maybe a few apps would have broken tests that were legitimate uncaught failures that were now erroring. However I did not expect  this side effect, nor know even know that people were testing assertions this way.

-----------

So how do we fix this? I would love to have some collaboration on this, but my belief is that the whole point of this change, and the test adapter, is to provide a mechanism to catch errors in RSVP and Backburner. 

The idea I've been kicking around is to provides a helper mechanism that leverages replacing the test adapter with a temporary one for a period of time. Here is an example using QUint:

```javascript
function expectEmberError(assert, callback, match) {
    let origTestAdapter = Ember.Test.adapter;
    let testError;
    let TestAdapter = QUnitAdapter.extend({
      exception(error) {
        error = testError;
      }
    });
 
   Ember.run(() => { Ember.Test.adapter = TestAdapter.create(); });
   callback();
   Ember.run(() => { Ember.Test.adapter.destroy(); });
   Ember.Test.adapter = origTestAdapter;

  assert.ok(error && error.message.match(matcher));
}

exceptEmberError(assert, () => {
    Ember.run(() => {
      Ember.RSVP.resolve(() => { throw 'bar must be defined'; });
    }, /bar must be defined/);
}, /bar must be defined/);
```

I also maintain https://github.com/workmanw/ember-qunit-assert-helpers and this feels perfect for that (and it would solve the issue with ember-paper, but doesn't help with mocha). 

Also, one other thought is I wonder if the TestAdapter were to just rethrow the error (instead of using `assert`), if that would restore this behavior.

Thoughts? Would you mind trying that last suggestion? You should be able to easily test it by just adding the following to your setup:

```javascript
import MochaTestAdapter from 'ember-mocha-adapter;

Ember.Test.adapter = MochaTestAdapter.extend({
  exception(error) {
    throw error;
  }
}).create();
 ```

**Comment by jasonmit** [comment]
@workmanw a variation of what you posted seems to do the trick.  I opened an issue against ember-mocha so it's also on their radar.

Also, no need to apologize this is the cost of moving things forward sometimes :)

**Comment by toranb** [comment]
@jasonmit this also broke my test suite for the following 

```js
test('each computed is truly readonly', function(assert) {
  assert.expect(1);
  this.render(hbs`{{count-list}}`);
  try {
    this.$('.btn-alter').trigger('click');
  } catch (e) {
    assert.ok(e.message.indexOf('Cannot set read-only property') > -1);
  }
});
```

toranb:
@workmanw I was able to update this to work with qunit with help from assert.throws

```js
test('each computed is truly readonly', function(assert) {
  assert.expect(1);
  this.render(hbs`{{count-list}}`);
  Ember.run(() => {
    assert.throws(() => {
      this.$('.btn-alter').trigger('click');
    }, (e) => {
      return e.message.indexOf('Cannot set read-only property') > -1;
    });
  });
});
```

The full diff shows the qunit adapter hacking required to make this work (not usually something ember core is proud to see me showing off but it does solve the immediate issue)

https://github.com/ember-redux/ember-redux/pull/102/files

toranb:
@workmanw I spoke 2 soon after all :) was able to dump the hacks I had previously and the diff is just a swap of the try/catch for assert throws (wrapped inside ember.run)

jasonmit:
@toranb that sounds much cleaner :)

jasonmit:
I unfortunately wasn't able to get this working, but in my example I was throwing "on render" and not an action.  Perhaps this can be handled by patching `this.render`?

cibernox:
I am on a similar situation. Tests in ember-power-select started failing with the update from 2.11.2 to 2.11.3.

The test itself is hard to fix because the error happens immediatly when the select is rendered without the mandatory `onchange` action.

Test: https://github.com/cibernox/ember-power-select/blob/master/tests/integration/components/power-select/assertions-and-deprecations-test.js#L9-L19

Assertion: https://github.com/cibernox/ember-power-select/blob/master/addon/components/power-select.js#L99-L109

I could use some guidance, because everything I've tried (wrap things un Ember.run, use try/catch...) didn't work.

workmanw:
Sorry for the delay on my part. So I was able to make the adapter helper I previously discussed work. Here is a complete QUnit test that pass with Ember-2.11.3.

```js
// tests/integration/component-assert-test.js
import { test, moduleForComponent } from 'ember-qunit';
import { QUnitAdapter } from 'ember-qunit';
import Ember from 'ember';

function expectEmberAssert(qAssert, callback, matcher) {
  let origTestAdapter = Ember.Test.adapter, testError;
  let TestAdapter = QUnitAdapter.extend({
    exception(error) {
      testError = error;
    }
  });

  Ember.run(() => { Ember.Test.adapter = TestAdapter.create(); });
  callback();
  Ember.run(() => {
    Ember.Test.adapter.destroy();
    Ember.Test.adapter = origTestAdapter;
  });

  qAssert.ok(testError && testError.message && testError.message.match(matcher));
}

moduleForComponent('x-assert-test', 'Ember.assert in integration test', {
  integration: true,
  beforeEach() {
    this.register('component:x-assert-test', Ember.Component.extend({
      init() {
        this._super();
        Ember.assert('x-assert-test will always assert');
      }
    }));
  }
});

test('Check for assert', function(assert) {
  expectEmberAssert(assert, () => {
    this.render(hbs`{{x-assert-test}}`);
  }, /x-assert-test will always assert/);
});
```

So the impact for the tests are minor, just have to use a new utility for asserting. 

As I previously mentioned I've been working on [ember-qunit-assert-helpers](https://github.com/workmanw/ember-qunit-assert-helpers) to provide qunit aware asserts. I will carve out some time **today** to add this helper over there and cut a new version.

workmanw:
Okay @miguelcobain @toranb @cibernox . As promised, I added this functionality to `ember-qunit-assert-helpers` and released a new version `0.1.3`. If you add that addon to your package, it "should be" as simple as changing `assert.throws` to `assert.expectAssertion`.

```javascript
assert.expectAssertion(() => {
  this.render(hbs`
    {{#power-select options=countries selected=selected as |opt|}}{{opt}}{{/power-select}}
  `);
}, /requires an `onchange` function/);
```

toranb:
@cibernox I'd take a look at the work @workmanw did w/ that qunit helper assuming you have the bandwidth. Without that it may be possible to get it working with something like the below (but this isn't anything I would advice people do generally as it requires monkey patching the TestAdapter/ and Logger)

```js
let application, originalLoggerError, originalTestAdapterException;

module('testing some error that is thrown up ... good times', {
    beforeEach() {
        application = startApp(); //if acceptance testing
        originalLoggerError = Ember.Logger.error;
        originalTestAdapterException = Ember.Test.adapter.exception;
        Ember.Logger.error = function() {};
        Ember.Test.adapter.exception = function() {};
    },
    afterEach() {
        Ember.Logger.error = originalLoggerError;
        Ember.Test.adapter.exception = originalTestAdapterException;
        Ember.run(application, 'destroy');
    }
});
```

workmanw:
Also, FWIW, @rwjblue proposed adding an additional API to ember-qunit that makes this process more sane, without replacing the adapter. I'm going to pursue that and submit a PR this week. See: https://github.com/workmanw/ember-qunit-assert-helpers/pull/6#issuecomment-285959516

cibernox:
@workmanw I'll then wait a couple days to see how this evolves. Thanks you all for the prompt response.

sandersky:
I don't want to argue whether or not this fix should go in, but I personally think Ember 2.11.4 should be released to revert this change. Adding breaking changes in patch releases isn't very kosher when it comes to [semver](http://semver.org/).

lolmaus:
We're using Ember 2.11.3 which includes this fix, yet we're still running into #12791. 

Please tell me if this PR is supposed to resolve #12791.

Use case: we have a `store.findRecord().catch()` in a route. The `catch` makes 404 errors silent and transparent for end user. But the test still fails due to a 404 from Mirage.

stefanpenner:
This PR appears to break the following invariant:

```
Ember.run(() => {
  throw new Error('error');
});

// we get here, but we should not. `Ember.run`ate the error
```

workmanw:
Going to follow up on #15600.

runspired:
So I know I'm late to this party and I'm about to stand on a soap-box but I really think this was the wrong idea all around.  While this change does help catch a class of errors that were otherwise silently ignored, it also breaks every expectation of "how the system works" and takes us very far from writing "just Javascript".

For example, I write this code:

```js
class CustomError extends Error {}
...
throw new CustomError('we are error prone!');
```

With this change, in order to test my error case, I now must

1) understand what the run-loop is
2) determine whether my error will be thrown within a run-loop in the test
3) understand that if I am in a run-loop, I will receive a totally different error thrown from an internal `Ember.assert` instead of the actual Error that I wrote and threw.
4) hack my way around carefully annotating what I did and why if I want to assert the actual error class in tests
5) give up in despair if I both need a run-loop and access to the error class

rwjblue:
For those interested, I submitted https://github.com/emberjs/ember.js/pull/15871 which is a partial revert of this PR.

Issue #14891, opened 2017-01-31
`computed.union()` doesn't do what people would naturally expect it to do


Herriau:

locks:
@Herriau can you please provide a description of what the behaviour you expect is, and what the current behaviour is?

Herriau:
I'd expect it to work similarly to `computed.intersect()` (http://emberjs.com/api/classes/Ember.computed.html#method_intersect), except obviously it'd be computing a union instead of an intersection, e.g.

```javascript
let obj = Ember.Object.extend({
  allFriends: Ember.computed.union('adaFriends', 'charlesFriends')
}).create({
  adaFriends: ['Charles Babbage', 'John Hobhouse', 'William King', 'Mary Somerville'],
  charlesFriends: ['William King', 'Mary Somerville', 'Ada Lovelace', 'George Peacock']
});

obj.get('allFriends'); // ['Charles Babbage', 'John Hobhouse', 'William King', 'Mary Somerville', 'Ada Lovelace', 'George Peacock']
```

The current behavior is to:

> computes a new array with all the unique elements from the dependent array

locks:
Your code snippet seems to be how it currently works? https://ember-twiddle.com/fb5f3d17bb84526b37b762b7472eee16

Are you saying that the output should have 'Mary Somerville' twice?

bwbuchanan:
Perhaps the problem is that the docs don't make it clear that `computed.union()` accepts multiple dependent keys, because it's an alias for `computed.uniq()`?

From a code clarity point of view, it makes sense for an app to use `uniq` if it wants the set of unique elements from a single array, and `union` if it wants the set union of several arrays, even though they're the same function. The docs could be updated to reflect this and to indicate that `union` accepts multiple dependent keys.

locks:
I want to note that `computed.intersect` also doesn't reflect you can pass multiple property keys and should be updated with the other macros.

Herriau:
@bwbuchanan Correct this seems to be a documentation issue rather than an implementation issue.

@locks The doc for `computed.intersect` is actually pretty clear about that:

> A computed property which returns a new array with all the duplicated elements from two or more dependent arrays.

PR #14904, opened 2017-02-03
[DOC release] Update reduce_computed_macros.js

locks:
Fixes https://github.com/emberjs/ember.js/issues/14891#issuecomment-277327006.

### DIFF
--- a/packages/ember-runtime/lib/computed/reduce_computed_macros.js
+++ b/packages/ember-runtime/lib/computed/reduce_computed_macros.js
@@ -427,7 +427,33 @@ export function uniqBy(dependentKey, propertyKey) {
 }
 
 /**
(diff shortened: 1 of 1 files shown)


bwbuchanan:
Should `@param {String} propertyKey*` instead be `@param {String} ...propertyKey` in these cases? (`union`, `intersect`, `or`, `and`, etc.)

https://yui.github.io/yuidoc/syntax/#param

perlun:
Hi @bwbuchanan 

> Should @param {String} propertyKey* instead be @param {String} ...propertyKey in these cases? (union, intersect, or, and, etc.)

I think you had a point here. The current YUIDoc doesn't really reflect the example; the examples are clear, but the YUIDoc is confusing. 😄 

Will you do a PR?

Issue #14626, opened 2016-11-22
Possible memory leak when switching routes


Haidy777:
I'm using Ember `2.8.3`. Every time I'm switching between routes there are some strange detached DOM trees inside the heap snapshop.

I've got a small repository (https://github.com/Haidy777/leaky-ember) which has two branches (master with `2.8.3`, ember-2-9-1 with `2.9.1`). Even when I'm switch between this "small" routes ("page1", "page2") I've got some of these detached elements.

Since the application which I'm working on is much more complex than this simple demo, there are even more detached elements. Which causes huge trouble on iOS-Devices (since they are very memory limited)

Haidy777:
@locks What else do you need to verify this bug?

stefanpenner:
> Even when I'm switch between this "small" routes ("page1", "page2") I've got some of these detached elements.

can you share some screenshots of what you see in the heap profiler and such? That helps makes sure all available information is shared to anyone working to  confirm/diagnose.

Haidy777:
Of course.
I also added one of the snapshots into the repository mentiond above.
After some switching between the two routes there are serval of this 'HTMLDivElement' and 'HTMLHeadingElement' visible.
On this small scale it doesn't cost much memory, but as more content and more complex DOM gets changed, the footage grows.

To give you an overview: on one route we have a huge table with dynamically loaded entries which use the same component with some more nested components. On the other we have a medium sized grid which also gets dynamically loaded.

I will try to get the demo more clear to reproduce this problem better.

![image](https://cloud.githubusercontent.com/assets/5011101/20528793/ddf021e0-b0cd-11e6-9faa-668f0dec2b84.png)

Haidy777:
I just update the demo, no there are more Detached DOM trees when changing the routes:
![image](https://cloud.githubusercontent.com/assets/5011101/20795712/d73713e8-b7d2-11e6-96bd-4d73a4b693f4.png)
Also there is another snapshot: https://github.com/Haidy777/leaky-ember/blob/master/Heap-20161201T143246.heapsnapshot

herom:
@locks @stefanpenner do you need a twiddle in order to verfiy this bug? if so, I could try to get this example running in a twiddle too.

stefanpenner:
Reproduction in app form as  I believe you have added should be enough. I just have not had a chance to loook.

If I don't comment by mid-week feel free to pig me so I do :)

herom:
@stefanpenner alles klar, danke :)

[...]

PR #14911, opened 2017-02-07
Clearing references resolve memory leaks

pieter-v:
Maintaining these references results in detached DOM trees

### DIFF
--- a/packages/ember-glimmer/lib/component.js
+++ b/packages/ember-glimmer/lib/component.js
@@ -575,6 +575,13 @@ const Component = CoreView.extend(
       assert(`You cannot use a computed property for the component's \`tagName\` (${this}).`, !(this.tagName && this.tagName.isDescriptor));
     },
 
+    destroy: function () {
+      this._super(...arguments);
+
+      this[ROOT_REF] = undefined;
+      this[BOUNDS] = undefined;
+    },
+
     rerender() {
       this[DIRTY_TAG].dirty();
       this._super();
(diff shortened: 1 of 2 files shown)


pieter-v:
This fixes for me alse the issue #14626

mmun:
@pieter-v Thanks for looking into this leak! Have you managed to find the root cause of the leak? By your analysis, it must be that we are accidentally retaining a destroyed object somewhere.

pieter-v:
With our app, we had  detached DOM trees, which should normally be garbage collected.
I looks that it has something to do with circular references (which I find strange, because I thought that the GC used a mark-sweep algorithm)

The references I cleared is just an old try and error approach ;-)

stefanpenner:
> I looks that it has something to do with circular references

It's most likely that there is a different dominating retainer (something connecting `window` or `self` or similar to the objects you are seeing retained). The circles are a red herring (that is unless there is a bug in the JS runtime you are using)

pieter-v:
Did some more investigation:
* some references were indeed held by our code
* some references were held by ember-data but emberjs/data#2f18405a4785486f3bd7c865ca1096ea44e2c184 solves this
* and some references were held by _lastRenderedReferenceMap but it seems that this one is only used in debug builds and not in production builds

So, I think this pull requeste can be aborted.

As a side remark: to track down the references held by our code, I changed the implementation of HAS_NATIVE_WEAKMAP, to return always false. Otherwise everything dissapears in the 'metaStore' which is difficult to explore. Should it be possible to specify this is a build option?

rwjblue:
@pieter-v - I think that is an interesting observation.  I'm not quite sure what we should do, but you may well be right that we should expose some sort of configuration.  Would you mind opening a new issue that is specifically for that?

PR #14916, opened 2017-02-08
Remove -in-element and -with-dynamic-args

chadhietala:
This should be in Glimmer. This PR fixes this https://github.com/tildeio/glimmer/pull/406


rwjblue:
LGTM, will need to bump glimmer version before landing this...

Issue #14884, opened 2017-01-28
isEmpty returns false negative if tested object has a falsey 'size' attribute


ballPointPenguin:
Given an object foo
 {
  name: 'foo',
  size: 0
}

Ember.isEmpty(foo) will return True.
I expected it to return False.

This is causing weird issues for me when passing records to ember-selectize, because the records have an arbitrary 'size' attribute which default to the number 0.

ballPointPenguin:
FYI my "solution" is to rename all 'size' attributes on my models.
Not sure if anything else should be done here. Maybe documentation of de facto reserved attribute names would be helpful.

pixelhandler:
@ballPointPenguin yeah it does seem odd not to know that `size` is reserved and to have to rename, e.g. `xSize` instead of `size` to avoid the invalid output from checking `isEmpty` on a non-empty object.

quick repo of the issue: https://ember-twiddle.com/f139797bbf4c82c70438f12565869e14?openFiles=controllers.application.js%2C (check the console to see the output).

Looks like the implementation checks for `size` as a `Number`. See https://github.com/emberjs/ember.js/blob/v2.11.0/packages/ember-metal/lib/is_empty.js#L36-L38

Not sure yet, if this is a `Documentation` issue or a `Bug`. I'm curious to find out the reason that the implementation checks for a size property.

pixelhandler:
@locks do you think this is a Documentation issue? The docblock has no mention of checking for a `size` property that is a number And that `size: 0` means "Empty" but that does kind of make sense.

ballPointPenguin:
I think there is a valid use case for a model that happens to have an attribute called 'size', and this size happens to equal the number 0.

Unexpected things happen as a result. For example, ember-cli-selectize makes wrong assumptions https://github.com/miguelcobain/ember-cli-selectize/blob/e5d2f5831f00ea293dceb4a994abd275074735a4/addon/components/ember-selectize.js#L146

Still, I don't want to call it a bug, and I don't think `isEmpty` should change (who knows how many other things depend on its exact behavior?). I think including 'reserved attributes' or something like that in the docs might be helpful.

PR #14928, opened 2017-02-13
[BUGFIX release] Fixes `isEmpty` for non-Ember.Map object with size property

locks:
Fixes https://github.com/emberjs/ember.js/issues/14884
Relevant context: https://github.com/emberjs/ember.js/pull/9414

### DIFF
--- a/packages/ember-metal/lib/is_empty.js
+++ b/packages/ember-metal/lib/is_empty.js
@@ -1,5 +1,6 @@
 import { get } from './property_get';
 import isNone from './is_none';
+import EmberMap from './map';
 
 /**
   Verifies that a value is `null` or an empty string, empty array,
@@ -39,7 +40,7 @@ export default function isEmpty(obj) {
 
   let objectType = typeof obj;
 
-  if (objectType === 'object') {
+  if (objectType instanceof EmberMap) {
     let size = get(obj, 'size');
     if (typeof size === 'number') {
       return !size;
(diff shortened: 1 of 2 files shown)


packages/ember-metal/lib/is_empty.js
```diff
@@ -39,7 +40,7 @@ export default function isEmpty(obj) {
 
   let objectType = typeof obj;
 
-  if (objectType === 'object') {
+  if (objectType instanceof EmberMap) {
```
stefanpenner: not sure if this matters, but this would not work if 2 embers where present on the same page. As there would be two different `EmberMap` and the `isEmpty` of one, could not work with the other.

rwjblue: The fix for this, would be to add some property to `EmberMap` that we can check (e.g. `obj.isEmberMap` or `obj[SOME_PRIVATE_EMBER_MAP_SYMBOL_THINGY]` depending on how paranoid we want to be)...

locks: Sorry for letting this slide. What's the call here? Will investigate CI failure.

mixonic: @lock I think the private, non-enumerable symbol would be the correct path.

locks:
Close due to https://github.com/emberjs/rfcs/pull/237.

PR #14942, opened 2017-02-20
Fix ember-utils export ROOT and use it in ember-metal

fsmanuel:
Somehow `export ROOT` got missing but `ember-utils/index.js` exports it from `./super` [Code](https://github.com/emberjs/ember.js/blob/master/packages/ember-utils/lib/index.js#L25).
With the export in place we can use it in `ember-metal/mixin.js`

### DIFF
--- a/packages/ember-metal/lib/mixin.js
+++ b/packages/ember-metal/lib/mixin.js
@@ -11,6 +11,7 @@ import {
   guidFor,
   GUID_KEY,
   NAME_KEY,
+  ROOT,
   wrap,
   makeArray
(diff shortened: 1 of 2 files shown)


rwjblue:
What made you go looking into this? Was there a failure scenario or some other issue?

rwjblue:
Yep, totally makes sense! I was just curious if there was a failure condition, and if we should create a regression test for it. Seems like probably not since its not really a breakage for "normal" consumers, but it is definitely something that we should fix.

Thanks for working on it!

Issue #14909, opened 2017-02-07
Make Location A Public API


chadhietala:
It's a mistake that the interface to the location is a private API. Creating an issue so we don't forget. /cc @rwjblue.

rwjblue:
Confirm, the interface [discussed here](https://github.com/emberjs/ember.js/blob/3bd80eef4182a8d5eddbc24eb93382fdc41bd338/packages/ember-routing/lib/location/api.js#L105-L120) is definitely intended to be public.

I suspect what we need to do is rework those guides a bit though:

* make it so that we can document the API and the `Custom implementation` section as public
* move the documentation of `AutoLocation`, `HistoryLocation`, and `NoneLocation` into their own actual implementation files
* list the various implementations and cross link them in the API docs

PR #14946, opened 2017-02-21
[DOC] Fixing documentation of Location API and various implementations

sclatter:
https://github.com/emberjs/ember.js/issues/14909

Also, doing a minor spiff of Component to accurately reflect current
state of the Layout API, which was previously marked as public.

@locks @chadhietala @rwjblue  - please review for intent of the requested documentation changes. 

Tests: 
QUnit 1.23.1; Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/56.0.2924.87 Safari/537.36
Tests completed in 347338 milliseconds.
29731 assertions of 29731 passed, 0 failed.
skippedEmber.Application - visit() Integration Tests: SKIPPED: Test setup (0)Rerun
skippedEmber.Application - visit() Integration Tests: SKIPPED: iframe setup (0)Rerun
skippedHelpers test: {{unbound}}: SKIPPED: helper form updates on parent re-render (0)Rerun
skippedEmberObject ES Compatibility: SKIPPED: calling extend on an ES subclass of EmberObject (0)Rerun

### DIFF
--- a/packages/ember-glimmer/lib/component.js
+++ b/packages/ember-glimmer/lib/component.js
@@ -430,6 +430,28 @@ export const BOUNDS = symbol('BOUNDS');
   See [Ember.Templates.helpers.yield](/api/classes/Ember.Templates.helpers.html#method_yield)
   for more information.
 
+  Layout can be used to wrap content in a component. In addition
+  to wrapping content in a Component's template, you can also use
+  the public layout API in your Component JavaScript.
+
+  ```app/components/person-profile.js
+    import Ember from 'ember';
+
+    export default Ember.Component.extend({
+      layout: Ember.HTMLBars.compile("<h1>Person's Title</h1><div class='details'>{{yield}}</div>")
+    });
+  ```
+
+  The above will result in the following HTML output:
+
+  ```html
+    <h1>Person's Title</h1>
+    <div class="details">
+      <h2>Chief Basket Weaver</h2>
+      <h3>Fisherman Industries</h3>
+    </div>
+  ```
+
 
   ## Responding to Browser Events
 
@@ -807,14 +829,7 @@ const Component = CoreView.extend(
      */
 
     /**
-      A component may contain a layout. A layout is a regular template but
-      supersedes the `template` property during rendering. It is the
-      responsibility of the layout template to retrieve the `template`
-      property from the component (or alternatively, call `Handlebars.helpers.yield`,
-      `{{yield}}`) to render it in the correct location.
-      This is useful for a component that has a shared wrapper, but which delegates
-      the rendering of the contents of the wrapper to the `template` property
-      on a subclass.
+      Layout can be used to wrap content in a component.
       @property layout
       @type Function
       @public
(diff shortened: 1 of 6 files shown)


packages/ember-routing/lib/location/api.js
```diff
@@ -13,94 +13,14 @@ import { getHash } from './util';
...
-  Keep in mind that since some of your users will use `HistoryLocation`, your
-  server must serve the Ember app at all the routes you define.
-
-  ### NoneLocation
-
-  Using `NoneLocation` causes Ember to not store the applications URL state
-  in the actual URL. This is generally used for testing purposes, and is one
-  of the changes made when calling `App.setupForTesting()`.
 
   ## Location API
```
chadhietala: Below there is an example of extending `Ember. HistoryLocation` and I'm not sure if we want to encourage that pattern or not. When @rwjblue originally talked we had mentioned that the Location implementations are "private", however this likely means that you will end up in a world with copy pasted code. I would like to propose that the Location implementations become "protected" so that one code extend or compose (see below) for a cleaner way of creating custom location objects.


```
import Ember from 'ember';
const { HistoryLocation } = Ember;

export default class MyHistory {
  implementation: 'my-custom-history',
  constructor() {
    this._history = HistoryLocation.create(...arguments);
  }
  create() {
    return new this(...arguments);
  }
  pushState(path) {
     console.log(path);
     this._history.pushState(path);
  }
}
```

packages/ember-glimmer/lib/component.js
```diff
@@ -430,6 +430,28 @@ export const BOUNDS = symbol('BOUNDS');
...
 
+  Layout can be used to wrap content in a component. In addition
+  to wrapping content in a Component's template, you can also use
+  the public layout API in your Component JavaScript.
+
+  ```app/components/person-profile.js
+    import Ember from 'ember';
+
+    export default Ember.Component.extend({
+      layout: Ember.HTMLBars.compile("<h1>Person's Title</h1><div class='details'>{{yield}}</div>")
```
rwjblue: This is definitely correct, but more commonly these days we would use ember-htmlbars-inline-precompile babel plugin or importing an actual template module as opposed to relying on compiling the template at runtime. 

I'd suggest demonstrating the latter:

```
import layout from '../templates/components/person-profile';

  layout
});
```

packages/ember-routing/lib/location/api.js
rwjblue: Ya, I think my point in discussion was that the main public "location API" methods were public API, and we should allow extending from HistoryLocation / NoneLocation / etc, but only the main location API methods should be considered public on them.

mixonic:
@sclatter is this still blocked on your edits?

sclatter:
@mixonic - yes, apologies. Been traveling and this dropped off my todo list. Will pick up again this week.

sclatter:
@rwjblue @chadhietala - see if this is more what you wanted now - I did put the example in - sounded like there was not opposition to it. Also, moved implementations to protected.

locks:
Superseded by https://github.com/emberjs/ember.js/pull/15523 because I couldn't figure out how to update this PR. Apologies ;x

Issue #14933, opened 2017-02-15
Setting checked property of input[type="checkbox"] after prevent default does not update the checkbox's checked state


smfoote:
[Ember Twiddle](https://ember-twiddle.com/5be44e8c89497b808b76293698193bda?openFiles=components.toggle-pill.js%2C)

I'm creating a component to wrap a checkbox, and I'm trying to follow the Data Down, Actions Up principle. The interface template interface for the component looks like:

```
{{toggle-pill checked=checked onToggle=(action "pillToggled")}}
```

The component stops (`preventDefault`, `stopPropagation`) click events on the input in order to avoid mutating the `checked` data internally (key events omitted for simplicity in the Twiddle), in order to allow the consumer to prevent the checkbox from checking or unchecking as they see fit (e.g. to wait for an AJAX call to complete before checking).

However, when the `pillToggled` event updates the `checked` property, the actual checkbox does not change state. In the Twiddle I am outputting the `checked` property as text, as well as setting `data-checked` on the input to show that the `checked` property is in fact changing.

Additionally, I have added a button external to the component, which also toggles the `checked` property through an action. Toggling the `checked` property through the button correctly updates the checked state of the checkbox.

Finally, I have observed that if the problematic call to set the `checked` property is wrapped in `Ember.run.later` (in application.js), the checkbox UI correctly updates. This is not a solution, but hopefully helpful in pinning down the source of the issue.

pixelhandler:
@smfoote Thanks for the detailed explanation and the twiddle example :)

This is not a bug in Ember, the input helpers were build back in a time when two way binding was a primary way of building with Ember. I'm pretty sure the input helpers are not components, I think that is why using `{{input}}` helper doesn't have a copy of the primitive boolean properly for `checked`. So it must be changed, preventing default blocks that change so the box never becomes checked.

The work around is to use a html `<input type="checkbox">` instead. See this example: https://ember-twiddle.com/d0c78546ca4c647dd06df8fb55b6dd87?openFiles=templates.components.toggle-pill.hbs%2C

I'll close this for now since there is a work around and the input template helpers are working as originally intended.

smfoote:
[...]

PR #14947, opened 2017-02-21
[BUGFIX beta] Fixes binding in input type=checkbox

tomasznapieralski:
This PR fixes bug described in #14933.

Demo that shows bug is [here](https://ember-twiddle.com/5a5574717bf6475c56170381f83ba029?openFiles=controllers.application.js%2C)

When input has bound `check` property it should be possible to check/uncheck checkbox by controlling this property, but right now it is not possible when checkbox is clicked, because this property is overwrote by jQuery value.


michaljach (approved):

stefanpenner:
From what I can tell, the issue is more that multiple `click` events are being fired, the `isSelected` toggles back and forth....

checkbox-fun

stefanpenner:
It appears this test passes without the above provided fix.

PR #14954, opened 2017-02-24
Update Glimmer

chadhietala:
This commit brings glimmer `master` and ember `master` in line.

### DIFF
--- a/packages/ember-glimmer/lib/environment.js
+++ b/packages/ember-glimmer/lib/environment.js
@@ -180,16 +180,7 @@ export default class Environment extends GlimmerEnvironment {
     }
   }
 
-  hasHelper(nameParts, symbolTable) {
-    assert('The first argument passed into `hasHelper` should be an array', Array.isArray(nameParts));
-
-    // helpers are not allowed to include a dot in their invocation
-    if (nameParts.length > 1) {
-      return false;
-    }
-
-    let name = nameParts[0];
-
+  hasHelper(name, symbolTable) {
     if (this.builtInHelpers[name]) {
       return true;
     }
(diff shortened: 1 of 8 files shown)


packages/ember-glimmer/lib/environment.js
```diff
@@ -180,16 +180,13 @@ export default class Environment extends GlimmerEnvironment {
     }
   }
 
-  hasHelper(nameParts, symbolTable) {
-    assert('The first argument passed into `hasHelper` should be an array', Array.isArray(nameParts));
+  hasHelper(name, symbolTable) {
 
     // helpers are not allowed to include a dot in their invocation
-    if (nameParts.length > 1) {
+    if (name.indexOf('.') > 1) {
```
chancancode: ✂️  I took care of this in the pre-compiler already

packages/ember-glimmer/lib/environment.js
```diff
@@ -239,31 +233,27 @@ export default class Environment extends GlimmerEnvironment {
...
     }
   }
 
-  hasModifier(nameParts) {
-    assert('The first argument passed into `hasModifier` should be an array', Array.isArray(nameParts));
-
+  hasModifier(name) {
     // modifiers are not allowed to include a dot in their invocation
-    if (nameParts.length > 1) {
+    if (name.indexOf('.') > -1) {
```
chancancode: ✂️  I took care of this in the pre-compiler already

chancancode (approved):

rwjblue:
Need some more cases for the AST transform:

```hbs 
{{foo.bar bar="something"}}
```

```hbs
{{foo.bar derp}}
```

```hbs 
{{#foo.bar}}stuff here{{/foo.bar}}
```

Specifically, rendering previously saved off closure components that are not derived from block params is totally valid and should be supported.

rwjblue:
👏

Issue #14914, opened 2017-02-08
Allow concatenated properties the {{component}} keyword


cibernox:
This feature is a nice to have. If it's computationally expensive it might not worth adding, but if added it would make a few composition patterns easier.

Say you are building a components that wraps another component that yields contextual components, and you want to bind some classes and allow users to bind their own.

```hbs
{{! imaginary datepicker component }}
{{#basic-dropdown as |dd|
  {{yield (hash
     trigger=(component dd.trigger class="datepicker-trigger")
     content=dd.content
  )}}
{{/basic-dropdown}}
```

You want the user to be able to add more classes to the trigger without removing the default one (`datepicker-trigger`):

```hbs
{{#my-datepicker as |dp|}}
  {{#dp.trigger class="extra-class"}}Foobar{{/dp.trigger}}
{{/my-datepicker}}
```

At the moment this code will generate
```hbs
<div class="basic-dropdown-trigger extra-class">Foobar</div>
{{! note that  "datepicker-trigger" is not present}}
```

One dirty-ish workaround is abuse the duality `class`/`classNames` un ember and do:

```hbs
{{! imaginary datepicker component }}
{{#basic-dropdown as |dd|
  {{yield (hash
     trigger=(component dd.trigger classNames="datepicker-trigger")
     content=dd.content
  )}}
{{/basic-dropdown}}
``` 

Since most people will use `class`, both will be present.

My feature request is to investigate if it's possible to consider special properties, like `class` concatenated properties in the `{{component}}` keyword, so the first example generates a div with both classes.

If this perf penalty it's significant disregard this request.

/cc @Serabe

locks:
Does this not need an RFC?

cibernox:
Unsure. @rwjblue said it could be considered a bug because it's not how components work, but it is indeed on the verge of being considered a breaking change.

rwjblue:
IMO, it is a bug. It should work, and I was surprised it didn't....

cibernox:
Concatenated properties is a feature to concatenate concatenate properties of a superclass with the ones defined in the class itself. 
This should work if se considering creating a creating a contextual component analogous to subclassing a component. If we consider that creating a contextual component (Technically that created a `ComponentDefinition` object) different this is working as expected.

pixelhandler:
@rwjblue so should this be labeled a bug?

mixonic:
[...]

PR #14956, opened 2017-02-24
[WIP] [BUGFIX beta] Support concatenatedProperties in contextual components

Serabe:
Given a component `my-comp` with a `concatenatedProperties: ['foo']` this:

```hbs
{{component (component (component "my-comp" foo=1) foo=2)}}
```

should produce `foo: [1, 2]` (unless there are other values in the class, then
those values would be concatenated as well).

This is not happening due to `curryArgs` does not take into account the
`concatenatedProperties` and `mergedProperties` properties. One of the problems
with current implementation is that these properties work at mixin level, thus
making them only accessible working out the `proto` (see emberjs/rfcs#209 for
more details.

This initial draft does not fully work. For this to continue I need to know
some details:

1. Is there an implemented way in glimmer to create an Reference that is
   a collection of reference? I need to concatenate two references/array of
   references into an array.
2. Is the `proto` method expensive enough not to continue?

### DIFF
--- a/packages/ember-glimmer/lib/helpers/component.js
+++ b/packages/ember-glimmer/lib/helpers/component.js
@@ -9,11 +9,13 @@ import {
   validatePositionalParameters
 } from '../syntax/curly-component';
 import {
+  combineTagged,
   EvaluatedArgs,
   EvaluatedNamedArgs,
   EvaluatedPositionalArgs,
   isComponentDefinition
 } from '@glimmer/runtime';
+import makeArray from 'ember-utils/make-array';
 import { assert, runInDebug } from 'ember-metal';
 
 /**
@@ -148,7 +150,7 @@ export class ClosureComponentReference extends CachedReference {
 
   constructor(args, symbolTable, env) {
     super();
-    
+
     let firstArg = args.positional.at(0);
     this.defRef = firstArg;
     this.tag = firstArg.tag;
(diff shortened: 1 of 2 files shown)


AnastasiyaYelets:
@Serabe hey, why did this pull close? Issue still not fixed in ember

Serabe:
I can try to give this another try, the thing was supporting this is (or at least was) pretty expensive and nobody had requested it in over two years. We can try and give it another go.

KirillSuhodolov:
@Serabe I founded similar issue https://github.com/emberjs/ember.js/issues/14914

I am trying to have merged properties that passing via component 

```
{{component (component (component "my-comp" foo=(hash a=2)) foo=(hash b=3))}}
```

foo defined as merged property at component

looks at codebase it not implemented at ember-glimmer component helpers

Serabe:
Yes, I know. The problem with `concatenatedProperties` and `mergedProperties` is that they are instance properties, and therefore would require creating an instance of the component to know its value.

KirillSuhodolov:
so, at such example will be create two extra waste components at runtime?

Serabe:
Yes.

KirillSuhodolov:
@Serabe anyway, is it possible add to ember?

Serabe:
@AnastasiyaYelets  @KirillSuhodolov this won't be fixed. This feature would be pretty expensive computationally and had some undesired side-effects from initialising the component each time.

Please, for anyone reading this, the behaviour that won't be fixed is not that concatenated properties won't work if a component is invoked as a contextual component, but that concatenated properties work in nested contextual components. Let me show this with an example:

```hbs
{{! this won't concatenate properties }}

{{#with (component "my-component" classNames="first-declaration") as |comp|}}
  {{#with (component comp classNames="second-declaration") as |nestedComp|}}
    {{component nestedComp}}
  {{/with}}
{{/with}}
```

`second-declaration` is overwriting `first-declaration` as value. This [ember twiddle](https://ember-twiddle.com/8e674001c73db34d206f107d99c40fd9?openFiles=templates.application.hbs%2C) shows the behaviour with a few classes.

Thank you!

PR #14961, opened 2017-02-25
[BUGIX release] [Fixes #14925] remove duplicate `/` in pathname

### LINKED ISSUE #14925: Ember 2.11 double forward slash causes application to break
accidently entered double forward slash in path breaks the application

```http://localhost:4200//```
or 
```http://localhost:4200//some-route```

route error handler never gets called to recover

### PR DESCRIPTION
**Pull request by stefanpenner**
[Fixes #14925]

This avoids a DOMException, which interprets duplicate `/` as a security violation and prevents the associated history state modifications.

error in question:
screen shot 2017-02-24 at 11 32 53 pm

---

I'm not totally sure this is the appropriate fix, so maybe someone else should sanity check this one.

### DIFF
--- a/packages/ember-routing/lib/location/history_location.js
+++ b/packages/ember-routing/lib/location/history_location.js
@@ -101,7 +101,8 @@ export default EmberObject.extend({
     // remove baseURL and rootURL from start of path
     let url = path
       .replace(new RegExp(`^${baseURL}(?=/|$)`), '')
-      .replace(new RegExp(`^${rootURL}(?=/|$)`), '');
+      .replace(new RegExp(`^${rootURL}(?=/|$)`), '')
+      .replace(/\/\/$/g,'/'); // remove extra slashes
 
     let search = location.search || '';
     url += search + this.getHash();
--- a/packages/ember-routing/tests/location/history_location_test.js
+++ b/packages/ember-routing/tests/location/history_location_test.js
@@ -333,3 +333,21 @@ QUnit.test('HistoryLocation.getURL() includes location.hash and location.search'
 
   equal(location.getURL(), '/foo/bar?time=morphin#pink-power-ranger');
 });
+
+
+QUnit.test('HistoryLocation.getURL() drops duplicate slashes', function() {
+  expect(1);
+
+  HistoryTestLocation.reopen({
+    init() {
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by krisselden** [comment]
@stefanpenner is this a regression? does this happen when there are multiple epsilon segments?

**Comment by stefanpenner** [comment]
> @stefanpenner is this a regression? 

I don't believe it is a regression, rather browsers hard-error (but in the past they did not).

> does this happen when there are multiple epsilon segments?

I'm not totally sure. Is your question if we should only strip some excess `/` not all?

**Comment by stefanpenner** [comment]
Chatting with @krisselden this is more likely an issue in RR not collapsing these "epsilon" segments.

**Comment by stefanpenner** [comment]
Actually. This appears to be unrelated to RR (route recognizer), and should be thought of as more of a "cleanup" of user input (from location.pathname)

**Comment by scalvert** [comment]
👍

File: packages/ember-routing/lib/location/history_location.js [hunk 1]
```diff
@@ -101,7 +101,8 @@ export default EmberObject.extend({
     // remove baseURL and rootURL from start of path
     let url = path
       .replace(new RegExp(`^${baseURL}(?=/|$)`), '')
-      .replace(new RegExp(`^${rootURL}(?=/|$)`), '');
+      .replace(new RegExp(`^${rootURL}(?=/|$)`), '')
+      .replace(/\/\//g,'/'); // remove extra slashes
```
rwjblue: Should this be anchored? If not, can you add a test showing that something like `/foo//bar/` is actually supported?

stefanpenner: I'm open to suggestions, I can glady add that test, or anchor it. thoughts?

rwjblue: lets just anchor at the end, we can always make it more permissive if we end up finding a need for it...

stefanpenner: sounds good

stefanpenner:
Updated as per @rwjblue's comment. We should be good to go if CI agrees.

rwjblue (approved):

stefanpenner:
@rwjblue should I backport or are you? I don't want to confuse the "process".

PR #14966, opened 2017-02-26
Add filter for undefined options.path during component-test blueprint/generator

mansona:
Howdy folks 👋 

I don't know the root cause of this issue but I thought I thought I would open the discussion with a PR that actually fixes it for me. 

The issue occurs when I use `ember g component-test super-component` from the command line. The render calls in the generated tests have the component name prepended with a `/` for some reason. Here is what it looks like: 

```
  this.render(hbs`{{/super-component}}`);

  this.render(hbs`
    {{#/super-component}}
      template block text
    {{//super-component}}
  `);
```

Doing a bit of debugging I found that `options.path` was undefined when coming into the generator, but like I said I don't know **why** that might be the case. All I know is that the fix in this PR works for me now 👍 

I didn't include any tests or anything because at first glance I didn't see any test harness that actually exercises the generators. Let me know if there is one and I'll create whatever tests are required. 

As ever if you have any questions just let me know 👍

### DIFF
--- a/blueprints/component-test/index.js
+++ b/blueprints/component-test/index.js
@@ -44,7 +44,7 @@ module.exports = useTestFrameworkDetector({
     var friendlyTestDescription = testInfo.description(options.entity.name, 'Integration', 'Component');
 
     if (options.pod && options.path !== 'components' && options.path !== '') {
-      componentPathName = [options.path, dasherizedModuleName].join('/');
+      componentPathName = [options.path, dasherizedModuleName].filter(Boolean).join('/');
     }
 
     if (options.testType === 'unit') {
--- a/node-tests/blueprints/component-test.js
+++ b/node-tests/blueprints/component-test.js
@@ -1,5 +1,7 @@
 'use strict';
 
+var fs = require('fs');
+
 var blueprintHelpers = require('ember-cli-blueprint-test-helpers/helpers');
 var setupTestHooks = blueprintHelpers.setupTestHooks;
 var emberNew = blueprintHelpers.emberNew;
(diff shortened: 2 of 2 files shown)


blueprints/component-test/index.js
```diff
@@ -44,7 +44,7 @@ module.exports = useTestFrameworkDetector({
     var friendlyTestDescription = testInfo.description(options.entity.name, 'Integration', 'Component');
 
     if (options.pod && options.path !== 'components' && options.path !== '') {
-      componentPathName = [options.path, dasherizedModuleName].join('/');
+      componentPathName = [options.path, dasherizedModuleName].filter(item => item).join('/');
```
rwjblue: You can simply the filter to `.filter(Boolean)`

rwjblue (commented):
I only had a small tweak suggested in implementation.

The tests for this is in https://github.com/emberjs/ember.js/blob/master/node-tests/blueprints/component-test.js, can you tweak to ensure it was failing and is now passing?

mansona:
Thanks for the feedback @rwjblue, much cleaner using the Boolean that way 👍 

So I noticed there was already a pod based test in the blueprint tests and figured it was a `.ember-cli` `usePods: true` issue. You can see the test failing here https://travis-ci.org/emberjs/ember.js/jobs/205521189

I know this probably raises questions about maybe running other tests with `usePods: true` and I know the structure of this test is slightly different to the rest of the file. How do you want me to proceed?

rwjblue:
Seems good to me as is, refactoring all the blueprints to test with and without `usePods` seems outside the scope here. 

Can you squash the commits and prefix with `[BUGFIX release]`?

mansona:
Like that? 

I tend to squash commits using github's "squash and merge" feature so I've never done it like this before 😂

rwjblue:
Yep, looks great!

PR #14972, opened 2017-02-28
[Deprecated] Add deprecation for partials

chadhietala:
This adds a deprecation for `{{partial}}`. Like `{{render}}` they go against the Ember's programming model.

### DIFF
--- a/packages/ember-glimmer/tests/integration/application/engine-test.js
+++ b/packages/ember-glimmer/tests/integration/application/engine-test.js
@@ -155,6 +155,7 @@ moduleFor('Application test: engine rendering', class extends ApplicationTest {
   }
 
   ['@test attrs in an engine']() {
+    expectDeprecation(/Partials are now deprecated in favor of using components. Please change "troll" to a component./)
     this.setupEngineWithAttrs([]);
 
     return this.visit('/').then(() => {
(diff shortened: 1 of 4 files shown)


workmanw:
I'm 👎  on this. I've often found partials a good way to break up a large page / component into more logical sections. Sure you could do that with components too, but the difference is scoping. Using components this way requires more boilerplate (passing data in and emitting actions up).

Obviously not the end of the world, but I would definitely be sad to see partials go.

rwjblue:
> I've often found partials a good way to break up a large page / component into more logical sections.

This is **exactly** why they should go 😜.  Making a file smaller does not actually reduce complexity, and instead just increases the mental overhead when you have to parse both the calling location (where you invoke `{{partial`) and the actual partial being invoked for access to all surrounding scope.  There is absolutely no API, no way to refactor the partial away from assuming "eval" style access to the calling scope, and no real benefit to splitting files this way.

> Using components this way requires more boilerplate (passing data in and emitting actions up).

Perhaps, but when you add an actual API to the thing you have many benefits. Including things like the ability to refactor, ability to test independently, etc.

---

Regardless of our disagreements here, this will need to end up becoming a deprecation RFC. Lets hash out the pros/cons over there....

mixonic:
This will absolutely go through an RFC process. 👍

locks:
Closing for now, until RFC is go.

Issue #11288, opened 2015-05-27
Unable to override serializeQueryParam within route


joshfarrant:
It seems like any overrides applied to the serializeQueryParam method in a route aren't being applied.

For example I could set `someQueryParam: 1` in the controller, then have something like this in the route (unrealistic but I'd have expected it to work):

```
serializeQueryParam: function(value, urlKey, defaultValueType) {
     return 20;
},
```

I'd expect the resulting param in the request to look something like this. 

?someQueryParam=20

However it remains like this

?someQueryParam=1

Is this expected?

exupero:
I'm seeing the same behavior using Ember 2.4. The method is called, but doesn't appear to have any effect.

I see several examples of using `serializeQueryParam` on GitHub, but all using Ember 1.

denzo:
Any updates on this?

denzo:
I see that `serializeQueryParam` is marked as `@private`. Is that intentional and means that I should avoid overriding it?

https://github.com/emberjs/ember.js/blob/v2.10.0/packages/ember-routing/lib/system/route.js#L430

ming-codes:
Any update on this? `serializeQueryParam` seems to work for me, but I'm also curious why it's marked as private.

acorncom:
@denzo @ming-codes Was asking about this myself this morning.  Per @rwjblue on Slack, this may not be public but it's at least "intimate API" (i.e. can't be broken).  So probably safe to use (see https://github.com/emberjs/ember.js/pull/14979 for work on the router service that includes tests for it).

@joshfarrant @exupero If this is still a problem for you, it'd be quite helpful to get a reproduction on Twiddle so we can check this more carefully.  Thanks!

acorncom:
@pixelhandler given the fact that `serializeQueryParam` is listed in Ed's PR tracking public API here (https://github.com/emberjs/ember.js/pull/16910/files#diff-d6b0c0d618513a72b8aec4c128e8b345R471), I think it remains an open question whether we're supporting this or not. As such, I don't think we should close this just yet.

PR #14979, opened 2017-03-03
Phase 3 of public Router Service - adding urlFor

scalvert:
This PR represents the second phase of the public router service described in [this RFC](https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md).

It extends the API with:

- `urlFor`

**Status**: Work in progress

Changes:

- Added `urlFor` method in public router service
- Added tests for `urlFor` 
- Added some extra tests for `transitionTo` and `replaceWith` that cover URLs as the first parameter

**How to test drive this PR**:

yarn start
Open up 2 tabs pointing to http://localhost:4200/tests/index.html, one with Enable Opt Features checked, ensure tests pass

### DIFF
--- a/packages/ember-routing/lib/services/router.js
+++ b/packages/ember-routing/lib/services/router.js
@@ -32,7 +32,7 @@ const RouterService = Service.extend({
 
      @method transitionTo
      @category ember-routing-router-service
-     @param {String} name the name of the route or a URL
+     @param {String} routeNameOrUrl the name of the route or a URL
      @param {...Object} models the model(s) or identifier(s) to be used while
        transitioning to the route.
      @param {Object} [options] optional hash with a queryParams property
@@ -41,7 +41,7 @@ const RouterService = Service.extend({
        attempted transition
      @public
    */
-  transitionTo() {
+  transitionTo(/* routeNameOrUrl, ...models, options */) {
     return this.router.transitionTo(...arguments);
   },
 
@@ -53,7 +53,7 @@ const RouterService = Service.extend({
 
      @method replaceWith
      @category ember-routing-router-service
-     @param {String} name the name of the route or a URL
+     @param {String} routeNameOrUrl the name of the route or a URL
      @param {...Object} models the model(s) or identifier(s) to be used while
        transitioning to the route.
      @param {Object} [options] optional hash with a queryParams property
@@ -62,8 +62,24 @@ const RouterService = Service.extend({
        attempted transition
      @public
    */
-  replaceWith() {
+  replaceWith(/* routeNameOrUrl, ...models, options */) {
     return this.router.replaceWith(...arguments);
+  },
+
+  /**
+     Generate a URL based on the supplied route name.
+
+     @method urlFor
+     @param {String} routeName the name of the route
+     @param {...Object} models the model(s) or identifier(s) to be used while
+       transitioning to the route.
+     @param {Object} [options] optional hash with a queryParams property
+       containing a mapping of query parameters
+     @return {String} the string representing the generated URL
+     @public
+   */
+  urlFor(/* routeName, ...models, options */) {
+    return this.router.generate(...arguments);
   }
 });
 
--- a/packages/ember-routing/lib/system/router.js
+++ b/packages/ember-routing/lib/system/router.js
@@ -411,8 +411,8 @@ const EmberRouter = EmberObject.extend(Evented, {
     return this.transitionTo(...arguments).method('replace');
   },
 
-  generate(...args) {
-    let url = this._routerMicrolib.generate(...args);
+  generate() {
+    let url = this._routerMicrolib.generate(...arguments);
     return this.location.formatURL(url);
   },
 
--- a/packages/ember-routing/lib/system/router.js
+++ b/packages/ember-routing/lib/system/router.js
@@ -411,8 +411,8 @@ const EmberRouter = EmberObject.extend(Evented, {
     return this.transitionTo(...arguments).method('replace');
   },
 
-  generate() {
-    let url = this._routerMicrolib.generate(...arguments);
+  generate(...args) {
+    let url = this._routerMicrolib.generate(...args);
     return this.location.formatURL(url);
   },
 
--- a/packages/ember/tests/routing/router_service_test/urlFor_test.js
+++ b/packages/ember/tests/routing/router_service_test/urlFor_test.js
@@ -1,0 +1,67 @@
+import { inject } from 'ember-runtime';
+import { Component } from 'ember-glimmer';
+import { Route, NoneLocation } from 'ember-routing';
+import {
+  get,
+  set
+} from 'ember-metal';
+import {
+  RouterTestCase,
+  moduleFor
+} from 'internal-test-helpers';
+
+import { isFeatureEnabled } from 'ember-metal';
+
+if (isFeatureEnabled('ember-routing-router-service')) {
+  moduleFor('Router Service - urlFor', class extends RouterTestCase {
+    constructor() {
+      super();
+    }
+
+    ['@test RouterService#urlFor returns URL for simple route'](assert) {
+      assert.expect(1);
+
+      return this.visit('/').then(() => {
+        let expectedURL = this.routerService.urlFor('parent.child');
+
+        assert.equal('/child', expectedURL);
+      });
+    }
+
+    ['@test RouterService#urlFor returns URL for simple route with dynamic segments'](assert) {
+      assert.expect(1);
+
+      let dynamicModel = { id: 1, contents: 'much dynamicism' };
+
+      return this.visit('/').then(() => {
+        let expectedURL = this.routerService.urlFor('dynamic', dynamicModel);
+
+        assert.equal('/dynamic/1', expectedURL);
+      });
+    }
+
+    ['@test RouterService#urlFor returns URL for simple route with query params'](assert) {
+      assert.expect(1);
+
+      let queryParams = { queryParams: { foo: 'bar' } };
+
+      return this.visit('/').then(() => {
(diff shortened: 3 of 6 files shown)


chadhietala:
Looking good. What is left? More tests? Really excited about this!

stefanpenner (approved):

packages/ember-routing/lib/system/router.js
```diff
@@ -411,8 +411,8 @@ const EmberRouter = EmberObject.extend(Evented, {
     return this.transitionTo(...arguments).method('replace');
   },
 
-  generate() {
-    let url = this._routerMicrolib.generate(...arguments);
+  generate(...args) {
+    let url = this._routerMicrolib.generate(...args);
```
rwjblue: Why do we need this change?  As you can see [here](http://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-2&targets=&browsers=&builtIns=false&code=let%20obj%20%3D%20%7B%0A%20%20generate(...args)%20%7B%0A%20%20%20%20let%20url%20%3D%20this._routerMicrolib.generate(...args)%3B%0A%20%20%20%20return%20this.location.formatURL(url)%3B%0A%20%20%7D%0A%7D%0A%0Alet%20obj2%20%3D%20%7B%0A%20%20generate()%20%7B%0A%20%20%20%20let%20url%20%3D%20this._routerMicrolib.generate(...arguments)%3B%0A%20%20%20%20return%20this.location.formatURL(url)%3B%0A%20%20%7D%0A%7D) the transpiled output is the same before and after.

Seems fine to make the change, but I was just curious what motivated it...

packages/ember/tests/routing/router_service_test/urlFor_test.js
```diff
@@ -0,0 +1,67 @@
...
+import {
+  RouterTestCase,
+  moduleFor
+} from 'internal-test-helpers';
+
+import { isFeatureEnabled } from 'ember-metal';
+
+if (isFeatureEnabled('ember-routing-router-service')) {
+  moduleFor('Router Service - urlFor', class extends RouterTestCase {
+    constructor() {
```
rwjblue: This `constructor` isn't really needed here, can remove..

packages/ember-routing/lib/services/router.js
```diff
@@ -62,8 +62,24 @@ const RouterService = Service.extend({
...
+  },
+
+  /**
+     Generate a URL based on the supplied route name.
+
+     @method urlFor
+     @param {String} routeName the name of the route
+     @param {...Object} models the model(s) or identifier(s) to be used while
+       transitioning to the route.
+     @param {Object} [options] optional hash with a queryParams property
```
rwjblue: The RFC suggests that the argument signature for `urlFor` is `routeName, ...models, queryParams`, however this is proposing `routeName, ...models, { queryParams }` (destructuring `queryParams` off of the last "options" argument).  See [RFC details here](https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md#new-method-url-generation).  

Why are we diverging?  Did you discover something during implementation that suggests we should have these diverge from the RFC (and no longer have the same API as `transitionTo` and `replaceWith`)?

rwjblue (changes requested):
Looking good!

I think we need test cases for the following scenarios:

* `router.urlFor('something', { selectedItems: ['a', 'b', 'c'] })` (a query param that is an array)
* `router.urlFor('something', { stuff: null })`
* `router.urlFor('something', { stuff: undefined })`
* `router.urlFor('something', { stuff: 'foo' })` where `something` route defines a custom `serializeQueryParam` method
* Tests defining a controller for the target, where the controller throws an error in `init`. The idea here is to ensure that we are not forcing controller instances to be created when generating the URL (which is a significant part of the performance issues surrounding `{{link-to}}`).

rwjblue:
> Tests defining a controller for the target, where the controller throws an error in init. The idea here is to ensure that we are not forcing controller instances to be created when generating the URL (which is a significant part of the performance issues surrounding {{link-to}}).

This [JSBin Demo](https://jsbin.com/tusekuc/edit?html,js,output) demonstrates that a bit more work will be needed to satisfy this constraint. 

Currently, when `routerMicrolib` invokes its `getHandler` function ([defined here](https://github.com/emberjs/ember.js/blob/a82158362cf4c4ce1447cefae3f8445ff37a9f46/packages/ember-routing/lib/system/router.js#L577-L619)), we look up the route, then eagerly invoke `route._populateQPMeta()` ([here](https://github.com/emberjs/ember.js/blob/a82158362cf4c4ce1447cefae3f8445ff37a9f46/packages/ember-routing/lib/system/router.js#L612)).

rwjblue:
> Currently, when `routerMicrolib` invokes its `getHandler` function ([defined here](https://github.com/emberjs/ember.js/blob/a82158362cf4c4ce1447cefae3f8445ff37a9f46/packages/ember-routing/lib/system/router.js#L577-L619)), we look up the route, then eagerly invoke `route._populateQPMeta()` ([here](https://github.com/emberjs/ember.js/blob/a82158362cf4c4ce1447cefae3f8445ff37a9f46/packages/ember-routing/lib/system/router.js#L612)). 

As I was digging into this, I was able to chart a path forward.  Please review https://github.com/emberjs/ember.js/pull/14980, which should resolve the specific entanglement I mentioned above...

scalvert:
@chadhietala more tests, working out specifics of performance characteristics (@rwjblue already found issues with eager instantiation of controllers), etc.

packages/ember-routing/lib/system/router.js
scalvert: I added this while I was fiddling with the params to ensure they were passed through to `_routerMicrolib.generate` in the expected format. I realized that this method handles extracting the params correctly already, so this is just a remnant of that. Will revert for sure!

packages/ember/tests/routing/router_service_test/urlFor_test.js
scalvert: I had been doing some custom work in this constructor, but removed it. The constructor was just left in place. Will remove!

packages/ember-routing/lib/services/router.js
scalvert: My interpretation was that we should mimic the signatures of `transitionTo` and `replaceWith`, as it's preferable to have a consistent API across all public methods. That said, I think what you said here is totally legit, in that the `queryParams` object should *be* the queryParams itself, not an object having a `queryParams` property that is the actual queryParams. 

I'm going to revisit all of them for consistency, but I know that `EmberRouter` expects an object that has a `queryParams` property that it unpacks. If we go this route we may want to revisit how `EmberRouter` is handling extracting the QPs.

Thoughts?

scalvert:
@rwjblue thanks for all the investigation! I totally appreciate it. 

The controller.init -> error test case was one that we'd discussed before, and is definitely one I want to add. I'm going to start listing all the test cases in a todo in the description of this PR, so we can ensure we're covering all the permutations.

I've reviewed https://github.com/emberjs/ember.js/pull/14980 and dropped a comment on it.

packages/ember-routing/lib/services/router.js
rwjblue: Yeah, I think it makes sense to use the same interface as `.transitionTo` / `.replaceWith` (which both have options with a `queryParams` property on them).  We can get @ef4 and others to confirm just before landing, but I think it makes sense to stay consistent for now.

If we do decide to change the API, they should all be changed, which is definitely unrelated to this PR...

scalvert: As per the RFC:

> This takes the same arguments as transitionTo

chadhietala:
Since this is one of the places where we need to improve on performance I think it might be useful to use a spy on methods we know we should not enter or entered a well known amount of time. In the long term we need to get Heimdall instrumentation in here, but I think asserting call counts in the tests can give us some confidence here.

packages/ember-routing/lib/services/router.js
rwjblue: HAHA! Derp.  I just was focusing on the code snippet:

```js
urlFor(routeName, ...models, queryParams)
```

Which is actually incorrect, it should have said:

```js
urlFor(routeName, ...models, { queryParams })
```

Or:

```js
urlFor(routeName, ...models, options)
```

(based on the prose just below)

https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md#new-method-url-generation

scalvert:
Looks like #14980 correctly resolved `urlFor` causing lookups on controllers. I verified this by adding the getters to the app as per your JSBin, ensuring the tests failed, and subsequently merging in the upstream changes. After the merge the tests now pass.

⭐️

scalvert:
@rwjblue added more tests for permutations of calls. What's left?

rwjblue (commented):
Additional tests needed:

* generating path via `routerService.urlFor(...stuff...)` then calling `routerService.transitionTo(pathGenerated)` ends up on the same path as `urlFor` generated
  * no dynamic segments (models) or query params
  * with dynamic segments (models) no query params
  * no dynamic segments (models) with query params
  * with dynamic segments (models) and query params
* Still need tests for paths generated for routes including a custom `serializeQueryParam`

scalvert:
Good ideas on the tests! Will work on adding those. 

The last one where we're testing paths generated for routes including a custom `serializeQueryParam` - shouldn't we wait until we fix the QP serialization, which we discussed would be a fast follow of this work?

rwjblue:
Yep, sounds good! Also, those `serializeQueryParam` tests and implementation work can close https://github.com/emberjs/ember.js/issues/11288.

rwjblue (approved):

rwjblue:
Thank you @scalvert!

PR #14980, opened 2017-03-03
Decouple controller lookup from URL generation.

rwjblue:
Currently, calling `emberRouter.routerMicrolib.generate('somethign');` will force both `something` route and `something` controller to be looked up. We do this so that we can properly handle any query params within that route/controller pair.

The `ember-routing-router-service` work is adding a `urlFor` method, that will ultimately allow users to generate urls without forcing the controllers to be looked up eagerly. Unfortunately, at the moment the eager controller lookup is completely entangled with the `getHandler` function that we provide to the router microlib.

This commit removes the forced eager evaluation of `_qp` on a route within the `getHandler` function, while still supporting the current implementation of `{{link-to}}` (which still utilizes the eager `_qp` evaluation to generate `href`'s).

Some work is needed in ember-engines to support this PR, which I am working on now...

I did emberobserver.com searches for the following:

* [route-meta](https://emberobserver.com/code-search?codeQuery=route-meta) - Only one real usage outside of Ember itself (ember-engines).
* [_populateQPMeta](https://emberobserver.com/code-search?codeQuery=_populateQPMeta) - Two usages (ember-engines and ember-cli-bundle-loader).

TODO: 

- [x] PR to make ember-engines compatible (https://github.com/ember-engines/ember-engines/pull/353)
- [x] PR to make ember-cli-bundle-loader compatible (https://github.com/MiguelMadero/ember-cli-bundle-loader/issues/30)

### DIFF
--- a/packages/ember-routing/lib/system/router.js
+++ b/packages/ember-routing/lib/system/router.js
@@ -609,7 +609,6 @@ const EmberRouter = EmberObject.extend(Evented, {
       }
 
       handler._setRouteName(routeName);
-      handler._populateQPMeta();
 
       if (engineInfo && !hasDefaultSerialize(handler)) {
         throw new Error('Defining a custom serialize method on an Engine route is not supported.');
--- a/packages/ember-routing/lib/system/route.js
+++ b/packages/ember-routing/lib/system/route.js
@@ -155,16 +155,6 @@ let Route = EmberObject.extend(ActionHandler, Evented, {
   },
 
   /**
-    Populates the QP meta information in the BucketCache.
-
-    @private
-    @method _populateQPMeta
-  */
-  _populateQPMeta() {
-    this._bucketCache.stash('route-meta', this.fullRouteName, this.get('_qp'));
-  },
-
-  /**
     @private
 
     @property _qp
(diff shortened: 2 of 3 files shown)


packages/ember-routing/lib/system/router.js
```diff
@@ -609,7 +609,6 @@ const EmberRouter = EmberObject.extend(Evented, {
       }
 
       handler._setRouteName(routeName);
-      handler._populateQPMeta();
```
scalvert: So you've removed this to enable lazy evaluating of qp meta, but where is this lazy evaluation happening? How is this backwards compatible?

rwjblue: @scalvert - Great question, general reasons below:

* if you look at `_populateQPMeta()`'s definition, all it does is eagerly `this.get('_qp')` and save it off into the bucket cache, however *nothing ever looks it up*!!!
* Things like `Ember.Router#_prepareQueryParams` (used by `{{link-to`, `transitionTo`, etc) are calling `route._getQPMeta()` which does `return route.get('_qp')` (I'm guessing that at some point it was returning from the bucket cache, but it does not do this any longer).  This still works (regardless of `_populateQPMeta()` being called first).

chadhietala:
Good investigative work on this @rwjblue!

PR #14982, opened 2017-03-03
[DOC] Document setupController hook behavior when model is undefined

Bestra:
The default hook implementation doesn't update the `model` property of the controller
if the resolved model is undefined, and I don't think that behavior is recorded anywhere except the source itself.

### DIFF
--- a/packages/ember-routing/lib/system/route.js
+++ b/packages/ember-routing/lib/system/route.js
@@ -1703,7 +1703,8 @@ let Route = EmberObject.extend(ActionHandler, Evented, {
     model supplied by the `model` hook.
 
     By default, the `setupController` hook sets the `model` property of
-    the controller to the `model`.
+    the controller to the `model`. If `model` is `undefined`, the default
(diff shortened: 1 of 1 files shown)


packages/ember-routing/lib/system/route.js
```diff
@@ -1703,7 +1703,8 @@ let Route = EmberObject.extend(ActionHandler, Evented, {
     model supplied by the `model` hook.
 
     By default, the `setupController` hook sets the `model` property of
-    the controller to the `model`.
+    the controller to the `model`. If `model` is `undefined`, the default
```
rwjblue: Instead of stating the default then correcting it with what happens when model is `undefined`, can we tweak how we state the default?

Perhaps something like:

> By default, the `setupController` hook sets the `model` property of the controller to the `model` specified when it is not `undefined`.

Bestra: For sure!

locks:
Thanks @Bestra :)

PR #14985, opened 2017-03-07
[BUGFIX release] Don't leak babel helpers globally

chadhietala:
These  should not be global variables and will break other external babel helpers.


krisselden:
I think you mean [BUGFIX release]

PR #14987, opened 2017-03-08
[BUGFIX release] don’t leak last destroyedComponents

stefanpenner:
Ensure we don’t retain the previous set of destroyed components.

---

Although `commit` did reset the `destroyedComponents` array, it does seem strange to retain the last set of `destroyedComponents` until the next render.

### DIFF
--- a/packages/ember-glimmer/lib/environment.js
+++ b/packages/ember-glimmer/lib/environment.js
@@ -64,7 +64,7 @@ export default class Environment extends GlimmerEnvironment {
     this.isInteractive = owner.lookup('-environment:main').isInteractive;
 
     // can be removed once https://github.com/tildeio/glimmer/pull/305 lands
-    this.destroyedComponents = undefined;
+    this.destroyedComponents = [];
 
     installPlatformSpecificProtocolForURL(this);
 
@@ -275,16 +275,16 @@ export default class Environment extends GlimmerEnvironment {
     this.inTransaction = true;
 
     super.begin();
-
-    this.destroyedComponents = [];
   }
 
   commit() {
+    let destroyedComponents = this.destroyedComponents;
+    this.destroyedComponents = [];
     // components queued for destruction must be destroyed before firing
     // `didCreate` to prevent errors when removing and adding a component
     // with the same name (would throw an error when added to view registry)
-    for (let i = 0; i < this.destroyedComponents.length; i++) {
-      this.destroyedComponents[i].destroy();
+    for (let i = 0; i < destroyedComponents.length; i++) {
+      destroyedComponents[i].destroy();
     }
 
     super.commit();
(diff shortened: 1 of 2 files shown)


packages/ember-glimmer/lib/environment.js
```diff
@@ -275,16 +275,16 @@ export default class Environment extends GlimmerEnvironment {
...
 
   commit() {
+    let destroyedComponents = this.destroyedComponents;
+    this.destroyedComponents = [];
     // components queued for destruction must be destroyed before firing
     // `didCreate` to prevent errors when removing and adding a component
     // with the same name (would throw an error when added to view registry)
-    for (let i = 0; i < this.destroyedComponents.length; i++) {
-      this.destroyedComponents[i].destroy();
+    for (let i = 0; i < destroyedComponents.length; i++) {
```
scalvert: Don't you want to cache length here?

packages/ember-glimmer/lib/environment.js
```diff
@@ -275,16 +275,16 @@ export default class Environment extends GlimmerEnvironment {
     this.inTransaction = true;
 
     super.begin();
-
-    this.destroyedComponents = [];
   }
 
   commit() {
+    let destroyedComponents = this.destroyedComponents;
+    this.destroyedComponents = [];
```
scalvert: Why do this assignment here. Why not just reassign to an empty array after the subsequent loop? More of a stylistic thing I think.

packages/ember-glimmer/tests/integration/components/destroy-test.js
```diff
@@ -0,0 +1,21 @@
...
+import { Component } from '../../utils/helpers';
+import { moduleFor, RenderingTest } from '../../utils/test-case';
+
+moduleFor('Component destroy', class extends RenderingTest {
+  ['@test it correctly releases the destroyed components'](assert) {
+    let FooBarComponent = Component.extend({});
+
+    this.registerComponent('foo-bar', { ComponentClass: FooBarComponent, template: 'hello' });
+
+    this.render('{{#if switch}}{{#foo-bar}}{{foo-bar}}{{/foo-bar}}{{/if}}', { switch: true });
```
scalvert: foobar inception...

packages/ember-glimmer/lib/environment.js
stefanpenner: if the loop fails `destroy` could throw, we still likely want to release the contents of the original `destroyedComponents`

packages/ember-glimmer/lib/environment.js
stefanpenner: tell me more

packages/ember-glimmer/tests/integration/components/destroy-test.js
stefanpenner: http://theinceptionbutton.com/#

packages/ember-glimmer/lib/environment.js
scalvert: lol stop messing with the old guy...

krisselden:
Can you set the length to 0? It will release it, is there a reason to make a new array?

stefanpenner:
> Can you set the length to 0? It will release it, is there a reason to make a new array?

I could, in this case I didn't believe the cost of a new array mattered so i left it closer to how it was originally implemented.

rwjblue:
Good catch @stefanpenner!

PR #14991, opened 2017-03-09
access active link state in the context

fly1ted:
I had a use-case where I needed a simple way to change an element inside {{#link-to}} so not to write additional css targeting that list. Exposing `this` helps to solve this.

```
{{#link-to 'foo' classNames="" as |l|}}
  <i class="Icon Icon--messages"></i>
  {{#unless l.active}}
    <span class="">
      bar
    </span>
  {{/unless}}
{{/link-to}}
```


bekzod:
I think https://github.com/emberjs/ember.js/pull/15835 is less obtrusive way which going for RFC too

locks:
Superseded by https://github.com/emberjs/ember.js/pull/15835, thanks!

PR #15000, opened 2017-03-11
Avoid a couple of redundant property loads.

bmeurer:
Due to the generally high level of polymorphism, these redundant property loads cannot be optimized appropriately by the JavaScript engines. So better just avoid them in the first place.


rwjblue:
Awesome, thank you!

scalvert (commented):
👍

bmeurer:
Nice. Thanks!

PR #15002, opened 2017-03-11
[Fixes #15001] Remove internal EmptyObject usage

### LINKED ISSUE #15001: Drop EmptyObject and use Object.create(null) again
It seems that `new EmptyObject` was added as a work-around for `Object.create(null)` performance. This has a couple of drawbacks for Ember itself, but also comes with a performance penalty in recent Chrome versions (i.e. in all current channels) that include [this CL](https://codereview.chromium.org/2430273007).

From what I can tell, `new EmptyObject` is often used to create objects that are supposed to be used as dictionaries. But at least in V8 function constructors always create fast mode objects, which means Ember probably wastes (metadata) memory in the browser plus time just to let the engine figure out that it should essentially transition these objects to dictionary mode objects. `Object.create(null)` always creates a dictionary mode object from the start (since the aforementioned CL landed), and it's compatible with `jQuery.isPlainObject` and other stuff. The allocation is roughly on par with the performance of `new EmptyObject` in all current Chrome channels.

So I'd like to trigger a discussion to get rid of `EmptyObject`.

/cc @krisselden @stefanpenner

stefanpenner:
- [x] confirm all are correctly removed
- [x] confirm tests pass
- [ ] do some tests

----

If this ends up good, we should likely consider deprecating `Ember.EmptyObject`.

--- 

cc @bmeurer


packages/ember-utils/lib/empty-object.js
```diff
@@ -1,8 +1,9 @@
-// This exists because `Object.create(null)` is absurdly slow compared
-// to `new EmptyObject()`. In either case, you want a null prototype
-// when you're treating the object instances as arbitrary dictionaries
-// and don't want your keys colliding with build-in methods on the
-// default object prototype.
+// This exists because `Object.create(null)` WAS slow compared to `new
```
rwjblue: We should likely remove this file completely. I did an ember-observer code search ([here](https://emberobserver.com/code-search?codeQuery=empty-object)) and there are no direct usages of our version any longer (though there are addons that have made their own copies of this code).

stefanpenner: i thought ember-data used it, but it ships its own: https://github.com/emberjs/data/search?utf8=%E2%9C%93&q=EmptyObject&type=Code

Which we should then also remove

rwjblue: Yep, I had already created an issue over there for removing it too 😸

stefanpenner: https://github.com/emberjs/data/issues/4853

stefanpenner:
ember-data PR: https://github.com/emberjs/data/pull/4854

stefanpenner:
@homu r+

PR #15006, opened 2017-03-13
allocate iteration structures lazily

stefanpenner:
Many times when we go through these functions, we never need to alloc the `seen` or `calls` arrays

@krisselden r?

### DIFF
--- a/packages/ember-metal/lib/meta.js
+++ b/packages/ember-metal/lib/meta.js
@@ -277,16 +277,18 @@ export class Meta {
 
   _forEachIn(key, subkey, fn) {
     let pointer = this;
-    let seen = new EmptyObject();
-    let calls = [];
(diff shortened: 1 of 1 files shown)


krisselden:
looks good, though this whole area could use some more testing.

stefanpenner:
Benchmarks for this guy:

```js
meta forEachIn
testing
- new (empty)
- old (empty)
- new (notEmpty)
- old (notEmpty)
  new (empty) ...... 57,116,740.98 op/s
  old (empty) ...... 13,495,844.67 op/s
  new (notEmpty) .... 2,384,559.17 op/s
  old (notEmpty) .... 2,346,057.37 op/s
```


For empty traversals (extremely common) we are now 4.3x faster and non empty traversals are about the same.

PR #15008, opened 2017-03-14
[BUGFIX beta] only defProp the deprecated container once per prototype

stefanpenner:
Currently we defProp once pre FactoryManager.prototype.create.

So if you have 3500 models of the same type, we defProp its class's prototype 3500 times instead of once.

### DIFF
--- a/packages/container/lib/container.js
+++ b/packages/container/lib/container.js
@@ -535,7 +535,7 @@ function instantiate(factory, props, container, fullName) {
       obj = factory.create(assign({}, injections, props));
 
       // TODO - remove when Ember reaches v3.0.0
-      if (!Object.isFrozen(obj) && 'container' in obj) {
+      if (!Object.isFrozen(obj)) {
         injectDeprecatedContainer(obj, container);
       }
     }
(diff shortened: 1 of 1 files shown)


stefanpenner:
Woah, this is pretty hefty:

emberaddons.com

before:

screen shot 2017-03-13 at 9 05 07 pm


After:

screen shot 2017-03-13 at 9 03 00 pm

chadhietala (approved):

stefanpenner:
cc @rwjblue just a note, we gotta make shit this lands in beta.

packages/container/lib/container.js
```diff
@@ -559,6 +559,7 @@ function factoryInjectionsFor(container, fullName) {
 
 // TODO - remove when Ember reaches v3.0.0
 function injectDeprecatedContainer(object, container) {
+  if ('container' in object) { return; }
```
rwjblue: This isn't quite right, but it may not matter at this point. In the case above (in `instantiate`) where we are creating an instance on a factory that does not have `.extend` we have to ensure that the `.container` property on the instance (which may be the `container._fakeContainerToInject` is replaced by a getter that issues a deprecation. 

Now, considering that this has been deprecated for quite some time (since ember 2.3), it's might be fine to drop support for the deprecation in this odd edge case.

I'll work up a failing test case so you can see what I mean.

PR #15009, opened 2017-03-14
[BUGFIX beta] [PERF] Cache FactoryManagers

### LINKED ISSUE #15007: cache FactoryManager instances
[FactoryManager](https://github.com/emberjs/ember.js/blob/ef8f46e15d920dfbe008594a0177a9a2103f90ed/packages/container/lib/container.js#L663) should likely be cached, as it seems strange to create new ones for each instance.

In addition, we could cache the result of `makeToString` on it, so we don't need to create so many of those either.

cc @chadhietala

stefanpenner:
* 1 FactoryManager per container/normalizeFullName
* cache madeToString per factoryManager

[fixes #15007]

- [x] add tests – @chadhietala can you do this?

### DIFF
--- a/packages/container/lib/container.js
+++ b/packages/container/lib/container.js
@@ -38,7 +38,7 @@ export default function Container(registry, options) {
   this.owner           = options && options.owner ? options.owner : null;
   this.cache           = dictionary(options && options.cache ? options.cache : null);
   this.factoryCache    = dictionary(options && options.factoryCache ? options.factoryCache : null);
-  this.factoryManagerCache = dictionary(options && options.factoryCache ? options.factoryCache : null);
+  this.factoryManagerCache = dictionary(options && options.factoryManagerCache ? options.factoryManagerCache : null);
   this.validationCache = dictionary(options && options.validationCache ? options.validationCache : null);
   this._fakeContainerToInject = buildFakeContainerWithDeprecations(this);
   this[CONTAINER_OVERRIDE] = undefined;
--- a/packages/container/lib/container.js
+++ b/packages/container/lib/container.js
@@ -38,6 +38,7 @@ export default function Container(registry, options) {
   this.owner           = options && options.owner ? options.owner : null;
   this.cache           = dictionary(options && options.cache ? options.cache : null);
   this.factoryCache    = dictionary(options && options.factoryCache ? options.factoryCache : null);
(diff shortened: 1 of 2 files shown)


packages/container/lib/container.js
```diff
@@ -289,16 +291,23 @@ if (isFeatureEnabled('ember-factory-for')) {
...
     }
 
+    let cached = this.factoryManagerCache[normalizedName];
+
+    if (cached) { return cached; }
+
     let factory = this.registry.resolve(normalizedName);
 
-    if (factory === undefined) { return; }
+    if (factory === undefined) {
```
krisselden: should misses not cache?

stefanpenner: i thought so, but there where test failures and I'm tired :P Maybe i can convince  @chadhietala to investigate?

stefanpenner: `resolve` does also cache so its not the end of the world either..

chadhietala (approved):

chadhietala:
Added tests to make sure the `FactoryManager` instances are the same on a cache hit and are not equal on a cache miss.

packages/container/lib/container.js
```diff
@@ -38,6 +38,7 @@ export default function Container(registry, options) {
   this.owner           = options && options.owner ? options.owner : null;
   this.cache           = dictionary(options && options.cache ? options.cache : null);
   this.factoryCache    = dictionary(options && options.factoryCache ? options.factoryCache : null);
+  this.factoryManagerCache = dictionary(options && options.factoryCache ? options.factoryCache : null);
```
rwjblue: This should not share the same cache as `this.factoryCache`.  I suspect this should be:

```js
this.factoryManagerCache = dictionary(options && options.factoryManagerCache ? options.factoryManagerCache : null);
```

rwjblue: Pushed a commit fixing this

stefanpenner:
Thanks @chadhietala

stefanpenner:
benchmarks (4.7x improvement):

```
node factory/factory-for.js                                                                                                 (1) (432ms)
testing
- factoryFor (new)
- factoryFor (old)
  factoryFor (new) .... 4,342,542.07 op/s
  factoryFor (old) ...... 915,862.86 op/s
```

PR #15012, opened 2017-03-14
[BUGFIX beta] avoid unneeded ToBoolean coercion in meta

stefanpenner:
cc @bmeurer 

TL;DR these functions are often hot, but are not always optimized, by being a little more verbose and explicit we can avoid costly coercions.

This sort of work is likely only appropriate in very hot functions

### DIFF
--- a/packages/ember-metal/lib/meta.js
+++ b/packages/ember-metal/lib/meta.js
@@ -224,7 +224,7 @@ export class Meta {
     let pointer = this;
     while (pointer !== undefined) {
       let map = pointer[key];
-      if (map) {
+      if (map !== undefined) {
         return map;
       }
       pointer = pointer.parent;
@@ -235,7 +235,7 @@ export class Meta {
     let pointer = this;
     while (pointer !== undefined) {
       let map = pointer[key];
-      if (map) {
+      if (map !== undefined) {
         let value = map[subkey];
         if (value !== undefined) {
           return value;
@@ -252,7 +252,7 @@ export class Meta {
 
     let outerMap = this._getOrCreateOwnMap('_deps');
     let innerMap = outerMap[subkey];
-    if (!innerMap) {
+    if (innerMap === undefined) {
       innerMap = outerMap[subkey] = Object.create(null);
     }
     innerMap[itemkey] = value;
@@ -262,9 +262,9 @@ export class Meta {
     let pointer = this;
     while (pointer !== undefined) {
       let map = pointer._deps;
-      if (map) {
+      if (map !== undefined) {
         let value = map[subkey];
-        if (value) {
+        if (value !== undefined) {
           let itemvalue = value[itemkey];
           if (itemvalue !== undefined) {
             return itemvalue;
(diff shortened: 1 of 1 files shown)


packages/ember-metal/lib/meta.js
```diff
@@ -262,7 +262,7 @@ export class Meta {
   hasDeps(subkey) {
     let pointer = this;
     while (pointer !== undefined) {
-      if (pointer._deps && pointer._deps[subkey]) {
+      if (pointer._deps !== undefined && pointer._deps[subkey] !== undefined) {
```
bmeurer: I'd suggest as follow-up cleanup to avoid the pointer._deps reload, i.e.

```javascript
let pointerDeps = pointer._deps;
if (pointerDepts !== undefined && pointerDeps[subkey] !== undefined) {
  return true;
}
```

packages/ember-metal/lib/meta.js
```diff
@@ -326,7 +327,7 @@ export class Meta {
     let descriptor = lookupDescriptor(obj, key);
     let isMandatorySetter = descriptor && descriptor.set && descriptor.set.isMandatorySetter;
```
bmeurer: Probably also use `descriptor !== undefined` etc. here.

packages/ember-metal/lib/meta.js
```diff
@@ -326,7 +327,7 @@ export class Meta {
...
 
-    if (isMandatorySetter) {
+    if (isMandatorySetter === true) {
```
bmeurer: This looks like overkill to me, because if you use `!==` in the individual `&&` cases above, then the compiler knows statically that `isMandatorySetter` is Boolean.

packages/ember-metal/lib/meta.js
```diff
@@ -493,7 +495,7 @@ if (isEnabled('mandatory-setter')) {
     let descriptor = lookupDescriptor(obj, key);
     let isMandatorySetter = descriptor && descriptor.set && descriptor.set.isMandatorySetter;
```
bmeurer: Same here.

bmeurer (changes requested):

packages/ember-metal/lib/meta.js
stefanpenner: this surprises me actually

bmeurer: For `r && l`, if both `r` and `l` are known to be Boolean (i.e. because they are `!==` comparisons), then the compiler knows that the result `r && l` will also be a Boolean.

stefanpenner: I'm surprised it knows `l` is a boolean in `let foo = r && l` As `l` being boolean or not shouldn't change the value of `foo`, only `r` can do that.

bmeurer: It doesn't, that's why you need to change the definition of `isMandatorySetter` above. :-)

krisselden:
It isn't immediately obvious which change is causing the failures.

stefanpenner:
I'll fix shortly

stefanpenner:
fixed

stefanpenner:
ah fixed for some of the matrix but not all, will look more closely tonight.

stefanpenner:
tests are green, but going to rename the commit message to `[BUGFIX beta]`

stefanpenner:
this should make señor @bmeurer happy :)

bmeurer:
Looks good, I like it 👍

PR #14967, opened 2017-02-27
Refactor {{mount}} syntax

mike183:
Supersedes https://github.com/ember-engines/ember-engines/pull/338

This PR adds no new functionality and is purely a refactor of the `{{mount}}` syntax in order to lay the ground work required for:

- ember-engines/ember-engines#117
- ember-engines/ember-engines#232

cc/ @rwjblue

PR #15015, opened 2017-03-14
Allow mounting routeless engines with a bound engine name

### LINKED ISSUE #117: Route less engine mounting with a variable
`{{mount someVar}}` does not work where `someVar` is a variable with the engine-name

It throws below error:
`Uncaught Error: Assertion Failed: The first argument of {{mount}} must be quoted, e.g. {{mount "chat-engine"}}.`

Not sure if this constraint is by design or an issue

mike183:
This PR builds upon the work in #14967, introducing the ability to pass bound properties to the `{{mount}}` syntax, enabling the dynamic mounting of "routeless" engines.

This also resolves ember-engines/ember-engines#117

cc/ @rwjblue

### DIFF
--- a/packages/ember-glimmer/lib/syntax/mount.js
+++ b/packages/ember-glimmer/lib/syntax/mount.js
@@ -47,11 +47,6 @@ export function mountMacro(path, params, hash, builder) {
     params.length === 1 && hash === null
   );
 
-  assert(
-    'The first argument of {{mount}} must be quoted, e.g. {{mount "chat-engine"}}.',
-    typeof params[0] === 'string'
-  );
-
   let definitionArgs = [params.slice(0, 1), null, null, null];
   let args = [null, null, null, null];
   builder.component.dynamic(definitionArgs, dynamicEngineFor, args, builder.symbolTable);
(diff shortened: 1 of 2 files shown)


rwjblue:
LGTM

@dgeb / @chancancode / @trentmwillis - Any objections?

packages/ember-glimmer/lib/syntax/mount.js
```diff
@@ -72,23 +67,27 @@ class DynamicEngineReference {
...
 
-    this._lastName = nameOrDef;
-    this._lastDef = new MountDefinition(nameOrDef);
+      this._lastName = nameOrDef;
+      this._lastDef = new MountDefinition(nameOrDef);
 
-    return this._lastDef;
+      return this._lastDef;
+    } else {
+      return null;
```
trentmwillis: IMO this should log if the argument isn't a `string/undefined/null`.

trentmwillis (changes requested):
Overall, I'm good with it. For bad input though, I'd like to see feedback given to the user.

packages/ember-glimmer/lib/syntax/mount.js
rwjblue: ya, lets have it assert

mike183:
@trentmwillis Thanks for the review, I've added an assertion to check that the provided engine name is valid and also added a test to go along with it.

cc/ @rwjblue

rwjblue (approved):

mike183:
🎉

PR #15016, opened 2017-03-15
small cleanup of `container`

bekzod:


packages/container/lib/container.js
```diff
@@ -421,15 +421,16 @@ function deprecatedFactoryFor(container, fullName, options = {}) {
...
+  if (cached) {
+    return cached;
   }
   let factory = registry.resolve(fullName);
   if (factory === undefined) {
     return;
   }
 
-  let type = fullName.split(':')[0];
+  let [type] = fullName.split(':');
```
stefanpenner: what does this transpile to? I think it cant infer  split returns an array.

bekzod: I think it infers from square brackets that it is array :P, https://cl.ly/0D1F3j103F42

stefanpenner: nah, it creates that silly helper:

```
var _slicedToArray = function () { function sliceIterator(arr, i) { var _arr = []; var _n = true; var _d = false; var _e = undefined; try { for (var _i = arr[Symbol.iterator](), _s; !(_n = (_s = _i.next()).done); _n = true) { _arr.push(_s.value); if (i && _arr.length === i) break; } } catch (err) { _d = true; _e = err; } finally { try { if (!_n && _i["return"]) _i["return"](); } finally { if (_d) throw _e; } } return _arr; } return function (arr, i) { if (Array.isArray(arr)) { return arr; } else if (Symbol.iterator in Object(arr)) { return sliceIterator(arr, i); } else { throw new TypeError("Invalid attempt to destructure non-iterable instance"); } }; }();

```

bekzod: probably it is created once for such cases and reused, https://cl.ly/2O0P2r0a3Z0Q, and such usage of spread exist in other places so that silly helper probably already exist :P

bekzod: such spread already exist here`ember.js/packages/ember-application/lib/system/resolver.js:205-208`

stefanpenner: yes, but it must run that code. If we leave it without destructuring it is not required.

object descturing `{ foo} = bar` works fine, but array destructuring is abit dubious unless the array is in scope`[first] = [1,2,3]`

bekzod: ok, free to close pr :P

locks:
Closing PR due to discussion in code review.

PR #15018, opened 2017-03-15
`container` `buildInjections` refactor

bekzod:


packages/container/lib/container.js
```diff
@@ -379,30 +379,24 @@ function areInjectionsDynamic(injections) {
...
   let hash = {};
+  let injections = Array.prototype.concat(...injectionsArray)
 
-  if (arguments.length > 1) {
-    let container = arguments[0];
-    let injections = [];
-    let injection;
-
-    for (let i = 1; i < arguments.length; i++) {
-      if (arguments[i]) {
```
bekzod: not sure here, I have removed this guard basically

packages/container/lib/container.js
```diff
@@ -379,31 +379,27 @@ function areInjectionsDynamic(injections) {
   return !!injections._dynamic;
 }
 
-function buildInjections() /* container, ...injections */{
+function buildInjections(container, ...injectionsArray) {
   let hash = {};
+  let injections = Array.prototype.concat(...injectionsArray);
```
artemgurzhii: Here you can use 
```js
let injections = [...injectionsArray];
```

stefanpenner: Isn't `...injectiosnArray` already the unrolled array?

stefanpenner: seems like it is, so I don think there is a point to this concat step.

https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=es2016%2Clatest%2Creact%2Cstage-2&targets=&browsers=&builtIns=false&code=function%20buildInjections(container%2C%20...injectionsArray)%20%7B%0A%20%20return%20injectionsArray%3B%0A%0A%7D%0A&experimental=true&loose=true&spec=false&playground=false

packages/container/lib/container.js
artemgurzhii: Shouldn't it be changed to push? [Benchmark for push vs concat.](http://jsbench.github.io/#1d1f32cc92a00df64e33d8b16aed4d82)

packages/container/lib/container.js
bekzod: since `injectionsArray` is array of arrays, you still need concat step https://cl.ly/1L3w103u3116, it either had to be `[].concat(...injectionsArray)` or `Array.prototype.concat(...injectionsArray)`, perf-wise my assumption it would be better to concat all arrays inside `injectionsArray` immediately rather than `for loop`ing one by one.

packages/container/lib/container.js
bekzod: @artemgurzhii I think you cannot use push because you are dealing with array elements

PR #15019, opened 2017-03-15
ember-views cleanup

artemgurzhii:


packages/ember-views/lib/system/event_dispatcher.js
```diff
@@ -196,12 +196,12 @@ export default EmberObject.extend({
...
   setupHandler(rootElement, event, eventName, viewRegistry) {
-    let self = this;
-
     if (eventName === null) {
       return;
     }
 
+    let self = this;
+
     rootElement.on(`${event}.ember`, '.ember-view', function(evt, triggeringManager) {
```
stefanpenner: should we instead use `=>` that way `this` is preserved behind the scenes.

stefanpenner (approved):

packages/ember-views/lib/views/core_view.js
```diff
@@ -72,10 +72,8 @@ const CoreView = FrameworkObject.extend(Evented, ActionHandler, {
     let name = arguments[0];
     let method = this[name];
     if (method) {
-      let args = new Array(arguments.length - 1);
-      for (let i = 1; i < arguments.length; i++) {
-        args[i - 1] = arguments[i];
-      }
+      let args = [...arguments];
```
stefanpenner: this seems to result in 

```
var args = [].concat(Array.prototype.slice.call(arguments));
```

The existing code is going to be more efficient

stefanpenner (changes requested):

packages/ember-views/lib/system/event_dispatcher.js
artemgurzhii: Wouldn't work because of [this](https://github.com/emberjs/ember.js/pull/15019/files#diff-c7b9de937661f47b233bb5e3101436e0R206).

packages/ember-views/lib/views/core_view.js
artemgurzhii: Done!

packages/ember-views/lib/mixins/action_support.js
```diff
@@ -131,8 +127,8 @@ export default Mixin.create({
     let action = this.actions && this.actions[actionName];
 
     if (action) {
-      let shouldBubble = action.apply(this, args) === true;
-      if (!shouldBubble) { return; }
+      let shouldBubble = action.apply(this, args);
+      if (shouldBubble === false) { return; }
```
chancancode: The before/after does not do the same thing.

artemgurzhii: Why?

chancancode: Before it was checking that the return value is exactly true, now it is checking that the return value is exactly false; if the function returns anything other than true or false, including returning nothing (undefined), the behavior is different

artemgurzhii: Done!

PR #15021, opened 2017-03-15
Avoid allocation for ChainNodes#_paths

stefanpenner:
* mostly this structure is empty (checked afew sites only 5%-7% of chain nodes have members in this property)
* don’t allocate it eagerly
* short-circuit accessing it if it is not present (except for additions of course)

### DIFF
--- a/packages/ember-metal/lib/chains.js
+++ b/packages/ember-metal/lib/chains.js
@@ -152,7 +152,7 @@ class ChainNode {
     this.count = 0;
 
     this._value = value;
-    this._paths = {};
+    this._paths = undefined;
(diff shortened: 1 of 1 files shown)


krisselden:
Good catch, paths is only a root node thing. Should this target beta? This area has low churn, easy to backport

stefanpenner:
updated to target beta

krisselden:
awesome!

stefanpenner:
```
- Original
- Updated (with the patch)
  Original ... 14,315,166.10 op/s
  Updated .... 19,221,458.12 op/s
```

PR #15022, opened 2017-03-15
`container` checking if injection is dynamic once

bekzod:
it allows not to call slightly expensive `isSingleton` repeatedly


stefanpenner (approved):

stefanpenner:
@bekzod good catch!!

bekzod:
@stefanpenner thanks ! learning :P

PR #15027, opened 2017-03-16
don’t allocate `[]` for no reason

stefanpenner:


krisselden:
looks great, this is also a very slow changing part of code, seems easy to cherry pick to beta.

stefanpenner:
@krisselden updated.

Issue #15020, opened 2017-03-15
Ember.A seems to modify array instance


marcoow:
We're seeing a bug in the Ember Simple Auth dummy app (simplabs/ember-simple-auth#1260) that seems to be caused by the fact that ~Enumerable.uniq~`Ember.A` tries to modify the provided array. This only seems to affect addon projects where prototype extension is disabled (this is mostly an assumption for now though).

Reproduction steps:

```js
var array = ['a', 'a'];
Object.freeze(array);
Ember.A(array);

// =>
ember.debug.js:21542 Uncaught TypeError: Cannot add property nextObject, object is not extensible
    at Object.defineProperty (ember.debug.js:21542)
    at applyMixin (ember.debug.js:20576)
    at MixinPrototype.apply (ember.debug.js:20770)
    at Object.A (ember.debug.js:37397)
    at eval (eval at init (query-string.js:35), <anonymous>:1:7)
    at Class.init (query-string.js:35)
    at new Class (ember.debug.js:36118)
    at Function._ClassMixinProps.create (ember.debug.js:36306)
    at Class.buildQueryString (oauth2-code.js:110)
    at Class.buildUrl (oauth2-code.js:120)
```

This only seems to fail in addon dummy apps but not in real apps.

/cc @Turbo87

stefanpenner:
I think its `Ember.A(array)` that is doing the modification

marcoow:
Hm, right - @Turbo87 and me just debugged this and we were sure it was the `uniq` but you're right,

```
var array = ['a', 'a'];
Object.freeze(array);
Ember.A(array);
```

actually results in the same exception. Should I update the issue above?

I assume this is still a bug though as `Ember.A` should also not modify the passed array?

Serabe:
[`Ember.A`'s definition when prototypes are not to be extended](https://github.com/emberjs/ember.js/blob/master/packages/ember-runtime/lib/system/native_array.js#L141-L149) applies `NativeArray` to the object and that defines properties on it, so I would say this is suggested behaviour.

Did this behaviour changed between versions? If you extend prototypes, is it still there?

Thank you!

stefanpenner:
`Ember.A()` has no choice but to modify the array prototype, even if frozen. That is the contract it has, if it cannot it will most likely produce downstream errors.

Turbo87:
[...]

PR #15032, opened 2017-03-17
Clarify Ember.NativeArray docs

marcoow:
This clarifies the docs for `Ember.NativeArray` a bit, making clear that `Ember.A` does not modify the original object's contents (it does actually modify the object itself, see #15020).


rwjblue:
Thank you @marcoow!

stefanpenner:
Lgtm

PR #15034, opened 2017-03-17
add options hash to replaceWith docs

kellyselden:
Since it sends to transitionTo with `...arguments`, the options hash goes with it.


locks (approved):

rwjblue:
Thanks!

PR #15037, opened 2017-03-18
Remove emberjs-build

chadhietala:
This removes emberjs-build and introduces babel @ 6.0.0. Also sets things up Svelte and [RFC#176](https://github.com/emberjs/rfcs/blob/master/text/0176-javascript-module-api.md).

- [x] Dev tests pass
- [x] Prod tests pass
- [x] Publish Backburner@0.4.0 /cc @lynchbomb @krisselden 
- [x] Node Tests

Also this makes rebuilds about twice as fast.

screen shot 2017-03-18 at 12 19 26 pm


package.json
```diff
@@ -55,15 +55,34 @@
...
+    "babel-plugin-check-es2015-constants": "^6.22.0",
+    "babel-plugin-feature-flags": "^0.3.1",
+    "babel-plugin-filter-imports": "^0.3.1",
+    "babel-plugin-transform-es2015-arrow-functions": "^6.22.0",
+    "babel-plugin-transform-es2015-block-scoping": "^6.23.0",
+    "babel-plugin-transform-es2015-classes": "^6.23.0",
+    "babel-plugin-transform-es2015-computed-properties": "^6.22.0",
+    "babel-plugin-transform-es2015-destructuring": "^6.23.0",
+    "babel-plugin-transform-es2015-modules-amd": "^6.24.0",
+    "babel-plugin-transform-es2015-modules-commonjs": "^6.24.0",
```
rwjblue: I don't think this is needed

broccoli/addon.js
```diff
@@ -0,0 +1,13 @@
...
+/* eslint-env node */
+
+const Funnel = require('broccoli-funnel');
+
+module.exports = function addon(name, project) {
+  var addon = project.findAddonByName(name);
+  var tree = addon.treeFor('addon');
+
+  return new Funnel(tree, {
+    srcDir: 'modules'
```
rwjblue: This changes in ember-cli@2.13.  We no longer emit `treeForAddon` into `modules/`.

broccoli/ember-tests-es.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+/* eslint-env node */
+
+const Funnel = require('broccoli-funnel');
+
+module.exports = function emberTestES(name) {
+  return new Funnel(`packages/${name}/tests`, {
+    getDestinationPath(relativePath) {
```
rwjblue: I think this can be changed to avoid requiring us to stat and symlink all the files:

```js
return new Funnel(`packages/${name}/tests`, {
  destDir: `${name}/tests/`
});
```

broccoli/find-lib.js
```diff
@@ -0,0 +1,19 @@
...
+
+  libPath = libPath || getLibPath(packagePath);
+
+  return path.resolve(packageRoot, libPath);
+};
+
+function getLibPath(packagePath) {
+  let packageJson = require(packagePath);
+
+  return path.dirname(packageJson['module'] || packageJson['main'] || '.');
```
rwjblue: The last `|| '.'` seems odd, do we need that?

broccoli/glimmer-template-compiler.js
```diff
@@ -0,0 +1,37 @@
...
+GlimmerTemplatePrecompiler.prototype.targetExtension = 'js';
+
+GlimmerTemplatePrecompiler.prototype.baseDir = function() {
+  return __dirname;
+};
+
+GlimmerTemplatePrecompiler.prototype.processString = function(content, relativePath) {
+  var compiled = this.precompile(content, { meta: { moduleName: relativePath } });
+  var template = 'import template from "../template";\n';
+  template += 'export default template(' + compiled + ');';
```
rwjblue: could probably use template strings nicely here:

```js
const stripIndent = require('common-tags').stripIndent;
//...snip...
return stripIndent`
  import template from '../template';
  export default template(${compiled});
`;
```

broccoli/minify.js
```diff
@@ -0,0 +1,32 @@
+const Funnel = require('broccoli-funnel');
+const Uglify = require('broccoli-uglify-js');
+const path = require('path');
+
+module.exports = function _minify(tree, name) {
+  return new Funnel(new Uglify(tree, {
```
rwjblue: Would you mind moving this `new Uglify` invocation to a separate variable declaration?  It makes it easier to parse below which options go to which thing...

broccoli/packages.js
```diff
@@ -0,0 +1,280 @@
...
+const VERSION_PLACEHOLDER = /VERSION_STRING_PLACEHOLDER/g;
+
+module.exports.routerES = function _routerES() {
+  return new Rollup(findLib('router_js', 'lib'), {
+    rollup: {
+      external: ['route-recognizer', 'rsvp'],
+      entry: 'router.js',
+      plugins: [{
+        transform(code, id) {
+          if (/[^t][^e][^r]\/router\.js$/.test(id)) {
```
rwjblue: Can you make an issue for us to follow up to remove this?  It seems to me that this should be part of router_js and we shouldn't have to monkey patch like this...

broccoli/packages.js
```diff
@@ -0,0 +1,280 @@
...
+      dest: '@glimmer/di.js',
+      external: ['@glimmer/util'],
+      format: 'es',
+      exports: 'named'
+    }
+  });
+}
+
+module.exports.emberGlimmerES = function _emberGlimmerES() {
+  return new GlimmerTemplatePrecompiler(new Funnel('packages/ember-glimmer/lib', {
```
rwjblue: Can you split the funnel to a separate variable here?

broccoli/packages.js
```diff
@@ -0,0 +1,280 @@
...
+    include: ['**/*.js', '**/*.hbs'],
```
rwjblue: What else is in this that we are filtering out?

broccoli/packages.js
```diff
@@ -0,0 +1,280 @@
...
+    getDestinationPath(relativePath) {
```
rwjblue: This can be replaced by: `destDir: 'ember-glimmer'`

broccoli/packages.js
```diff
@@ -0,0 +1,280 @@
...
+      dest: 'handlebars.js',
+      format: 'es',
+      exports: 'named',
+      plugins: [handlebarsFix()]
+    },
+    annotation: 'handlebars'
+  })
+}
+
+function handlebarsFix() {
```
rwjblue: Is there a path to removing this monkey patch? (not blocking, just curious)

broccoli/packages.js
```diff
@@ -0,0 +1,280 @@
...
+module.exports.dagES = function _dagES() {
+  return funnelLib('dag-map', {
+    files: ['dag-map.js'],
+    annotation: 'dag-map es'
+  });
+}
+
+module.exports.routeRecognizerES = function _routeRecognizerES() {
+  return funnelLib('route-recognizer', {
+    files: ['route-recognizer.es.js'],
```
rwjblue: We should make an issue and have this lib use the same structure as our other microlibs...

broccoli/packages.js
```diff
@@ -0,0 +1,280 @@
...
+        external: externs,
+        format: 'es',
+        exports: 'named'
+      }
+    });
+  }
+
+  return new Funnel(`packages/${name}/lib`, {
+    exclude: ['.gitkeep'],
+    getDestinationPath(relativePath) {
```
rwjblue: generally we should be able to `destDir: name` here

broccoli/packages.js
```diff
@@ -0,0 +1,280 @@
...
+      dest: `${name}.js`,
+      external: externs,
+      format: 'es',
+      exports: 'named'
+    },
+    annotation: `${name} es`
+  });
+}
+
+module.exports.emberTestsES = function _emberTestES(name) {
```
rwjblue: I'm confused, is this the same as `broccoli/ember-tests-es.js`, do we need both?

broccoli/transforms/inject-node-globals.js
```diff
@@ -0,0 +1,52 @@
...
+        if (path.node === importDecl) {
+          path.scope.registerDeclaration(path);
+        }
+      }
+    }
+  }
+}
+
+injectNodeGlobals.baseDir = function() {
+  return 'babel-plugin-transform-es2015-modules-amd';
```
rwjblue: This seems odd, it should return a path

ember-cli-build.js
```diff
@@ -6,428 +6,399 @@
...
+  emberPkgES,
+  glimmerPkgES,
+  emberTestsES,
+  nodeModuleUtils,
+  emberVersionES,
+  emberLicense,
+  emberFeaturesES,
+  packageManagerJSONs,
+  nodeTests
+} = require('./broccoli/packages');
```
rwjblue: Destructuring requires node 6 😢. I'm ok with requiring node@6 here, but when we ultimately have ember-cli apps building ember-source themselves (when not running a tagged version) we will need to downgrade to support node@4.

lib/packages.js
```diff
@@ -7,7 +7,7 @@ module.exports = function(features) {
     'ember-metal':                { trees: null,  requirements: ['ember-environment', 'ember-utils'], vendorRequirements: ['backburner'] },
     'ember-debug':                { trees: null,  requirements: [] },
     'ember-runtime':              { trees: null,  vendorRequirements: ['rsvp'], requirements: ['container', 'ember-environment', 'ember-console', 'ember-metal'] },
-    'ember-views':                { trees: null,  requirements: ['ember-runtime'] },
+    'ember-views':                { trees: null,  requirements: ['ember-runtime'], skipTests: true },
```
rwjblue: Are there no tests, or did you move them?

packages/ember-metal/lib/tags.js
```diff
@@ -64,7 +64,7 @@ function K() {}
 
 function ensureRunloop() {
   if (!run) {
-    run = require('ember-metal/run_loop').default;
+    run = require('ember-metal').run;
```
rwjblue: This is to bust a cycle?

packages/external-helpers/lib/external-helpers-prod.js
```diff
@@ -26,19 +26,19 @@ function defineProperties(target, props) {
...
 
-function createClass(Constructor, protoProps, staticProps) {
+export function createClass(Constructor, protoProps, staticProps) {
   if (protoProps) defineProperties(Constructor.prototype, protoProps);
   if (staticProps) defineProperties(Constructor, staticProps);
   return Constructor;
 }
 
-function interopExportWildcard(obj, defaults) {
+export function interopExportWildcard(obj, defaults) {
```
rwjblue: I think with `noInterop` and `strict` set for modules, this should not be needed?

broccoli/packages.js
chadhietala: Yes we need Handlebars to release. I think Tobias has fixed this.

lib/packages.js
chadhietala: There are no more tests in there.

packages/ember-metal/lib/tags.js
chadhietala: Likely

broccoli/transforms/inject-node-globals.js
chadhietala: So it's going to do hash-for-dep. So `__dirname` is no good, this thing doesn't have any dependencies and deriving the cachkey from Ember's deps seems wrong.

broccoli/packages.js
chadhietala: Done. https://github.com/tildeio/router.js/issues/208

broccoli/packages.js
chadhietala: Done https://github.com/tildeio/route-recognizer/issues/134

rwjblue (approved):

broccoli/transforms/inject-node-globals.js
krisselden: it does, it is dependent on babel and its own file, but it at least should be returning itself.

broccoli/packages.js
krisselden: it is fix in handlebars 4, didn't we bump in glimmer?

PR #15038, opened 2017-03-19
Ember application cleanup

artemgurzhii:


packages/ember-application/lib/system/engine-instance.js
```diff
@@ -40,12 +40,7 @@ const EngineInstance = EmberObject.extend(RegistryProxyMixin, ContainerProxyMixi
 
     guidFor(this);
 
-    let base = this.base;
-
-    if (!base) {
-      base = this.application;
-      this.base = base;
-    }
+    let base = this.base || (this.base = this.application);
```
rwjblue: I don't find this easier to read/understand, can you undo this bit?

packages/ember-application/lib/system/engine-instance.js
```diff
@@ -220,13 +215,12 @@ EngineInstance.reopenClass({
     registry.injection('view', '_environment', '-environment:main');
     registry.injection('route', '_environment', '-environment:main');
 
+    let inject = '-inert';
```
rwjblue: This is a good change, but lets give the variable a better name.  In context, I would suggest `rendererType`.

packages/ember-application/lib/system/resolver.js
```diff
@@ -197,7 +197,7 @@ export default EmberObject.extend({
 
   parseName(fullName) {
     return this._parseNameCache[fullName] || (
-      (this._parseNameCache[fullName] = this._parseName(fullName))
```
rwjblue: These extra parens are pretty common when doing an assignment inside of a boolean expression.

packages/ember-application/lib/system/resolver.js
```diff
@@ -269,10 +267,6 @@ export default EmberObject.extend({
     return description;
   },
 
-  makeToString(factory, fullName) {
```
rwjblue: This is just being deleted?  Where is it added back?

packages/ember-application/tests/system/dependency_injection/to_string_test.js
```diff
@@ -32,7 +32,7 @@ QUnit.module('Ember.Application Dependency Injection – toString', {
   }
 });
 
-QUnit.test('factories', function() {
+QUnit.test('factories', () => {
```
rwjblue: Tests have their own context, lets not use `() => {}` here. Please change these back to `function() { }`.

rwjblue (changes requested):

packages/ember-application/lib/system/engine-instance.js
artemgurzhii: How about this? 
```js
let base = this.base;

if (!base) {
  this.base = base = this.application;
}
```

packages/ember-application/lib/system/resolver.js
artemgurzhii: This function is not doing anything complicated + second argument is not being used. Can I replace all `makeToString` with `.toString()`?

rwjblue: Umm. No. 😸

packages/ember-application/lib/system/engine-instance.js
rwjblue: Seems good.

packages/ember-application/lib/system/engine-instance.js
```diff
@@ -220,13 +219,12 @@ EngineInstance.reopenClass({
...
     registry.injection('route', '_environment', '-environment:main');
 
+    let rendererType = 'inert';
     if (options.isInteractive) {
-      registry.injection('view', 'renderer', 'renderer:-dom');
-      registry.injection('component', 'renderer', 'renderer:-dom');
-    } else {
-      registry.injection('view', 'renderer', 'renderer:-inert');
-      registry.injection('component', 'renderer', 'renderer:-inert');
+      rendererType = 'dom';
```
locks: I much prefer the code on the left, no need for additional variable and interpolations, and is quite explicit.

packages/ember-application/lib/system/resolver.js
```diff
@@ -254,13 +253,12 @@ export default EmberObject.extend({
...
   lookupDescription(fullName) {
     let parsedName = this.parseName(fullName);
-    let description;
 
     if (parsedName.type === 'template') {
       return `template at ${parsedName.fullNameWithoutType.replace(/\./g, '/')}`;
     }
 
-    description = `${parsedName.root}.${StringUtils.classify(parsedName.name).replace(/\./g, '')}`;
+    let description = `${parsedName.root}.${StringUtils.classify(parsedName.name).replace(/\./g, '')}`;
```
locks: Isn't this hoisted to the top of the file, incurring the cost of transforming the description? I think that's why the original was split up.

packages/ember-application/lib/system/application.js
```diff
@@ -705,13 +706,11 @@ const Application = Engine.extend({
 
       // See documentation on `_autoboot()` for details
       if (this.autoboot) {
-        let instance;
+        // If we already have the __deprecatedInstance__ lying around, boot it to
```
locks: I'm suspicious of this change, I'd keep the original.

locks (changes requested):

packages/ember-application/lib/system/engine-instance.js
artemgurzhii: @locks Shouldn't it be more DRY? Those 4 lines are the same except renderer type.

locks: DRY doesn't mean reduce all duplication at all costs, it should be about duplication of units of knowledge in a codebase. Having a couple of simple method calls repeated is fine, it's very clear what it does.

packages/ember-application/lib/system/resolver.js
artemgurzhii: The reason why I moved this, is because there is no need to initialize a new variable before first `if` statement, as it's not being used in it.
> Isn't this hoisted to the top of the file, incurring the cost of transforming the description?

No, because `let` is block scoped.

locks: The nearest block for this `let` declaration is the function itself, though.

locks:
Closing due to git conflicts.

PR #15040, opened 2017-03-19
Remove cycle mixins/array -> system/each_proxy

chadhietala:
Seems ok to just inline this in the other file to remove the cycle.


rwjblue:
Seems fine to me.

PR #15042, opened 2017-03-20
`ember-debug/handlers` removed unnecessary check

bekzod:


packages/ember-debug/lib/handlers.js
```diff
@@ -13,8 +13,6 @@ export function invoke(type, message, test, options) {
 
   let handlerForType = HANDLERS[type];
 
-  if (!handlerForType) { return; }
-
   if (handlerForType) {
```
locks: should we `handlerForType !== undefined` here? cc @stefanpenner

rwjblue:
Thanks!

PR #15045, opened 2017-03-21
Register x-handlebars scripts into the container

mixonic:
Ember templates detected in the DOM have long been added to `Ember.TEMPLATES`. This made them available to applications via the `Ember.DefaultResolver`.

With the aim of eventually dropping the `DefaultResolver`, we need to add these templates elsewhere. This commit changes them to register directly into an application's registry. As a result, one could use a custom resolver (without `Ember.TEMPLATES` support) and still leverage templates stored in the DOM.

**Alternative:** Instead of ensuring DOM-loaded templates work regardless of resolver, there may be a way to only register them if the `DefaultResolver` is being used. For example the `Resolver` would be checked during initialization to confirm it is an `instanceof` `DefaultResolver`, and only then would templates be added to `TEMPLATES`. Apps without the default resolver would then *not* support DOM-loaded templates.

### DIFF
--- a/packages/ember-template-compiler/tests/system/bootstrap-test.js
+++ b/packages/ember-template-compiler/tests/system/bootstrap-test.js
@@ -1,12 +1,6 @@
 import { run } from 'ember-metal';
 import { jQuery } from 'ember-views';
-import {
-  Component,
-  getTemplate,
-  setTemplates,
-  hasTemplate,
-  setTemplate
-} from 'ember-glimmer';
+import { Component } from 'ember-glimmer';
 import bootstrap from '../../system/bootstrap';
 import {
   runAppend,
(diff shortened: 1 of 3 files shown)


packages/ember-template-compiler/tests/system/bootstrap-test.js
```diff
@@ -69,57 +62,42 @@ QUnit.test('template without data-template-name or id should default to applicat
   checkTemplate('application');
 });
 
-if (typeof Handlebars === 'object') {
```
mixonic: I'm pretty sure this is not supported any longer, as the string `x-raw-handlebars` appears nowhere in the codebase but here.

rwjblue: Definitely not supported. 🔥

rwjblue:
You have confirmed that things like JSBin still function properly with this change?

mixonic:
@rwjblue it lives: http://emberjs.jsbin.com/luwidaleya/1/edit?html,js,output

mixonic:
Upon further reflection and review, I think this approach is incorrect. This approach would allow an application to use DOM-loaded templates regardless of resolver, when in fact I believe we only want DOM-loaded templates to work when the resolver recognizes them (so, when one uses the default resolver or a subclass of it).

The change in approach here is basically zero-impact for now. In an upcoming PR I'll be adjusting tests based on the rational that `Ember.TEMPLATES` being present and used for template resolution is part of the default resolver use case.

PR #15048, opened 2017-03-21
Router not part of default resolver

mixonic:
`App.Router` is not part of the global-based default resolver API (e.g. `App.FooController`), but part of Ember's own API (e.g. `App.Resolver`). This change registers the `Router` class into the container instead of relying on resolution, removing a dependency on the default resolver
from applications that use `App.Router`.

Commits:

* The first commit in this PR passes tests, however the goal is that individual test can be migrated away from the default resolver.
* The second commit in this PR migrates the application tests away from the default resolver.
* Third commit changes the default resolver for `ApplicationTest` in ember-glimmer. Turns out most tests in this system use register instead of resolving templates, and that is also changed in this commit.
* Fourth commit migrates acceptance test tests to use the `TestResolver`

### DIFF
--- a/packages/ember-application/lib/system/application-instance.js
+++ b/packages/ember-application/lib/system/application-instance.js
@@ -103,6 +103,20 @@ const ApplicationInstance = EngineInstance.extend({
   _bootSync(options) {
     if (this._booted) { return this; }
 
+    /*
+     * router:main is always an instance of `App.Router`. In the past, this
+     * was because the container resolved the 'main' by looking to `App[type]`.
+     * By registering it explicitly, the dependency on the DefaultResolver is
+     * avoided.
+     *
+     * A copy is registered on the application instance since a new value for
+     * App.Router may have been set after the value registered on the
+     * application.
+     */
+    if (!this.resolveRegistration('router:main')) {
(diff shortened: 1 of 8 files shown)


mixonic:
@rwjblue promised me a comment, but he would like to treat `App.Router` the same as `App.FooController`, just part of the globals resolver. I'm amenable to this, but it is an even bigger change to our tests and to apps that opt for a completely custom (non default or default subclass) resolver.

rwjblue:
> @rwjblue promised me a comment, but he would like to treat App.Router the same as App.FooController, just part of the globals resolver

Sorry @mixonic. Yes, supporting `App.Router` is definitely a "globals mode" concern. You should definitely not be able to do this (and have it mean anything) in an ember-cli app:

```
// app/app.js snippet
const App = Ember.Application.extend({
  modulePrefix: config.modulePrefix,
  podModulePrefix: config.podModulePrefix,
  Resolver
});

App.Router = Ember.Router.extend({ /* custom stuff */ })

export default App;
```

mixonic:
Thanks, closing this in favor of `App.Router` being treated as a global lookup.

PR #15055, opened 2017-03-22
Decouple test from the default resolver

mixonic:
The Ember test suite is very coupled to the default globals-mode resolver. For example:

* Tests often make a controller available to an application as `App.FooController`
* Tests often use `setTemplates` to set a template onto `Ember.TEMPLATES`
* Even more tests use `App.Router` to set location or other configuration

All of these ways of passing things to Ember are considered part of the globals-mode resolver API.

At some far future date, we would like to drop the default resolver and its globals API from Ember. Nearly all (but not all) Ember apps are built using Ember-CLI. Those applications (with a few minor tweaks) do not need the default resolver. Already they use the [ember-resolver](https://github.com/ember-cli/ember-resolver), however this itself is a subclass of the default resolver and retains globals loading as a fallback behavior.

With this goal in mind, these commits start the process of migrating the Ember test suite away from the globals mode resolver and toward a `TestResolver` that provides the minimal interface for an Ember resolver. The `TestResolver` also provides a mechanism to add modules during test.

Combined with the newer class-based testing interface introduced to the codebase during the Glimmer2 merge cycle, we have a pretty decent test interface coming together here. For example:

```js
// These modules are also available via the internal-test-helpers package
import { moduleFor, ApplicationTest } from '../../utils/test-case';

moduleFor('Link-to component', class extends ApplicationTest {

  ['@test should be able to be inserted in DOM when the router is not present']() {
    this.addTemplate('application', `{{#link-to 'index'}}Go to Index{{/link-to}}`);

      this.assertText('Go to Index');
    });
  }

});
```
[source](https://github.com/mixonic/ember.js/blob/e536c5d4fb4c20ebe38dde36b80f1fc5dde62b8b/packages/ember-glimmer/tests/integration/components/link-to-test.js#L38-L44)

This PR:

* Changes `ApplicationTest` to use the `TestResolver` instead of the default resolver.
* The core team has consensus that we would like to test the *resolve* path over the *registry* path in our suite, so this PR also drops the test API of `this.registerController(Thing)` and family in favor of `this.add('type:name', Thing)`. `add` is backed by the `TestResolver`.
* Of course some test cases are not covered by `ApplicationTest`. This PR introduces a new test case `AutobootApplicationTestCase` for testing the timing of an autobooted application and uses it in `application_test`. This is a big diff, give it a look. *Some* tests here actually test behavior of the default resolver, and they are now flagged as such at the test module level.
* During `commonRegistrySetup` on the `Application`, `'router:main'` is now registered with the default value of `Router`. With the default resolver, `App.Router` is always found as a fallback. Without that fallback, we still require a default router at the `'router:main'` lookup. For default resolver apps this change should have no impact.

Once we have alignment on these API changes, I will create an epic to track further migration of the testing codebase away from the default resolver and registry.

### DIFF
--- a/packages/ember-application/tests/system/application_test.js
+++ b/packages/ember-application/tests/system/application_test.js
@@ -24,389 +24,404 @@ import {
   _loaded
 } from 'ember-runtime';
 import { compile } from 'ember-template-compiler';
-import { setTemplates, setTemplate } from 'ember-glimmer';
+import { setTemplates } from 'ember-glimmer';
 import { privatize as P } from 'container';
 import {
   verifyInjection,
   verifyRegistration
 } from '../test-helpers/registry-check';
+import { assign } from 'ember-utils';
+import {
+  moduleFor,
+  ApplicationTestCase,
+  AbstractTestCase,
+  AutobootApplicationTestCase
+} from 'internal-test-helpers';
 
 let { trim } = jQuery;
+let secondApp;
 
-let app, application, originalLookup, originalDebug, originalWarn;
+moduleFor('Ember.Application, autobooting multiple apps', class extends ApplicationTestCase {
+  constructor() {
+    jQuery('#qunit-fixture').html(`
+      <div id="one">
+        <div id="one-child">HI</div>
+      </div>
+      <div id="two">HI</div>
+    `);
+    super();
+  }
 
-QUnit.module('Ember.Application', {
-  setup() {
-    originalLookup = context.lookup;
-    originalDebug = getDebugFunction('debug');
-    originalWarn = getDebugFunction('warn');
+  get applicationOptions() {
+    return assign(super.applicationOptions, {
+      rootElement: '#one',
+      router: null,
+      autoboot: true
+    });
+  }
 
-    jQuery('#qunit-fixture').html('<div id=\'one\'><div id=\'one-child\'>HI</div></div><div id=\'two\'>HI</div>');
-    application = run(() => Application.create({ rootElement: '#one', router: null }));
-  },
+  buildSecondApplication(options) {
+    let myOptions = assign(this.applicationOptions, options);
+    return this.secondApp = Application.create(myOptions);
+  }
 
   teardown() {
-    jQuery('#qunit-fixture').empty();
-    setDebugFunction('debug', originalDebug);
-    setDebugFunction('warn', originalWarn);
+    super.teardown();
 
-    context.lookup = originalLookup;
-
-    if (application) {
-      run(application, 'destroy');
+    if (this.secondApp) {
+      run(this.secondApp, 'destroy');
     }
+  }
 
-    if (app) {
-      run(app, 'destroy');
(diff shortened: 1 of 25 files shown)


packages/ember-application/tests/system/application_test.js
```diff
@@ -24,389 +24,404 @@ import {
...
-        });
-      }
+  [`@test can specify custom router`](assert) {
+    let MyRouter = Router.extend();
+    run(() => {
+      let app = this.createApplication();
+      app.Router = MyRouter;
     });
+    assert.ok(
+      this.application.__container__.lookup('router:main') instanceof MyRouter,
```
rwjblue: Would be nice to remove this usage of `__container__` (you can use `__deprecatedInstance__` instead for now), as it should be issuing a deprecation...

packages/ember-application/tests/system/application_test.js
```diff
@@ -24,389 +24,404 @@ import {
...
-  equal(trim(jQuery('#qunit-fixture').text()), 'Hello World');
-});
+  [`@test initialize application with stateManager via initialize call from Router class`](assert) {
+    run(() => {
+      this.createApplication();
+      this.addTemplate('application', '<h1>Hello!</h1>');
+    });
+    // This is not a public way to access the container; we just
+    // need to make some assertions about the created router
+    let router = this.application.__container__.lookup('router:main');
```
rwjblue: as mentioned above, we should use `__deprecatedInstance__` instead, accessing `application.__container__` should have been deprecated long ago but was apparently not...

packages/internal-test-helpers/lib/test-resolver.js
```diff
@@ -0,0 +1,43 @@
+import { compile } from 'ember-template-compiler';
+
+class Resolver {
+  constructor() {
+    this._registered = {};
+    this.constructor.lastInstance = this;
```
rwjblue: this seems _slightly_ odd, why do we need this (as opposed to keeping track of the instance when we create it)?

rwjblue (approved):
👍 overall, I had a few minor questions (none of them are blocking though)

packages/internal-test-helpers/lib/test-resolver.js
mixonic: The instance is created by `Application`. We could access it via `application.__registry__.resolver` however that would couple this to the arrangement of "registry has a resolver", which isn't true for glimmer/di. Of course we will have a FF for that anyway so we could change the logic. It just feels ugly to be relying on `reaching.deep.into.__private__.stuff` to find the instance.

rwjblue: Ahh, right the resolver itself is not inside the container.

> It just feels ugly to be relying on reaching.deep.into.__private__.stuff to find the instance.

Agreed. This is one of the reasons I called out a few `__container__` things in this PR.

---

Seems good, thanks for explaining!

Issue #15058, opened 2017-03-23
[QUEST] Decouple Ember tests from the globals resolver


mixonic:
![](https://media.tenor.co/images/2b62869d23e1f9f407691b855cf8fd4a/tenor.gif)

(lifted some of this introduction from [#15055](https://github.com/emberjs/ember.js/pull/15055))

The Ember test suite is very coupled to the default globals-mode resolver. For example:

- Tests often make a controller available to an application as `App.FooController`
- Tests often use `setTemplates` to set a template onto `Ember.TEMPLATES`
- Even more tests use `App.Router` to set location or other configuration
 
All of these ways of passing things to Ember are considered part of the globals-mode resolver API.

At some far future date, we would like to drop the default resolver and its globals API from Ember. Nearly all (but not all) Ember apps are built using Ember-CLI. Those applications (with a few minor tweaks) do not need the default resolver. Already they use the [ember-resolver](https://github.com/ember-cli/ember-resolver), however this itself is a subclass of the default resolver and retains globals loading as a fallback behavior.

The first step (of many) toward removal of the globals-mode default resolver is to refactor Ember's test suite to not be reliant upon it. To this end [#15055](https://github.com/emberjs/ember.js/pull/15055) introduced a configurable resolver mock to be used in Ember tests, ideally through the class-based testing abstraction added during the [Glimmer2 test porting effort](https://github.com/emberjs/ember.js/issues/13127). For example:

``` js
/*
 * internal-test-helpers provides several testcase classes.
 */
import { moduleFor, ApplicationTestCase } from 'internal-test-helpers';

/*
 * Pass an anonymous class extending from the testcase to moduleFor.
 */
moduleFor('Link-to component', class extends ApplicationTestCase {

  /* 
   * As you are defining an ES class for your test module, you may
   * override APIs from the parent class and call super. If needed, you
   * can even create a whole new test case in internal-test-helpers
   */

  /*
   * In this format, tests are any property on the class that starts
   * with "@test".
   */
  ['@test should be able to be inserted in DOM when the router is not present']() {
    /*
     * Inside the test the class instance is `this`. For example
     * 
     *     this.addTemplate('application', '<p>{{prop}}</p>');
     *
     * Will add a template to the test resolver. To add a controller
     * you would use the specifier and `add`:
     *
     *     this.add('controller:application', Controller.extend());
     *
     * This replaces older test styles where App.ApplicationController
     * would be set.
     *
[...]

PR #15060, opened 2017-03-23
Decouple test from the default resolver for routing/toplevel_dom

josemarluedke:
This PR is to help remove the usage of the default resolver from the tests, which is described in  #15058. 

I think this covers the guideline from the [EPIC description](https://github.com/emberjs/ember.js/issues/15058), let me know if I have missed something.

### DIFF
--- a/packages/ember/tests/routing/toplevel_dom_test.js
+++ b/packages/ember/tests/routing/toplevel_dom_test.js
@@ -1,56 +1,11 @@
-import { run } from 'ember-metal';
-import { compile } from 'ember-template-compiler';
-import { Application } from 'ember-application';
-import { jQuery } from 'ember-views';
-import { NoneLocation } from 'ember-routing';
-import { setTemplates, setTemplate } from 'ember-glimmer';
+import { moduleFor, ApplicationTestCase } from 'internal-test-helpers';
 
-let App, templates, container;
+moduleFor('Top Level DOM Structure', class extends ApplicationTestCase {
+  ['@test Topmost template always get an element'](assert) {
+    this.addTemplate('application', 'hello world');
 
-function bootApplication() {
-  for (let name in templates) {
-    setTemplate(name, compile(templates[name]));
-  }
-  container.lookup('router:main');
-  run(App, 'advanceReadiness');
-}
-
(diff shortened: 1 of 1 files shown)


packages/ember/tests/routing/toplevel_dom_test.js
```diff
@@ -1,56 +1,11 @@
-import { run } from 'ember-metal';
-import { compile } from 'ember-template-compiler';
-import { Application } from 'ember-application';
-import { jQuery } from 'ember-views';
-import { NoneLocation } from 'ember-routing';
-import { setTemplates, setTemplate } from 'ember-glimmer';
+import { moduleFor, ApplicationTestCase } from 'internal-test-helpers';
 
-let App, templates, container;
+moduleFor('Top Level DOM Structure', class extends ApplicationTestCase {
```
mixonic: @josemarluedke It is ok for this test, but I want to mention there is a difference in how the application is booted here you may not have noticed.

In the original, options for `Application.create` are only `name, rootElement`. If you look at [`AbstractApplicationTestCase`](https://github.com/emberjs/ember.js/blob/2f686e30b65d52b2c48a8f6fd27df199cd143a62/packages/internal-test-helpers/lib/test-cases/abstract-application.js#L30) you'll see that it passes `autoboot: false`. The old tests were implicitly testing an auto-booted app, but using `deferReadiness` and `advanceReadiness` to provide a window during which they could set things up. Using `ApplicationTestCase`, you're instead using `autoboot: false` and [a call to the `visit` API](https://github.com/emberjs/ember.js/blob/2f686e30b65d52b2c48a8f6fd27df199cd143a62/packages/internal-test-helpers/lib/test-cases/abstract-application.js#L58) to do something similar.

As long as the semantics of autoboot aren't the subject of the test (I don't think they are here) this is good to go 👍

mixonic:
💯  Looks great to me :-) Thank you @josemarluedke!

packages/ember/tests/routing/toplevel_dom_test.js
josemarluedke: Thanks for letting me know. That is very helpful to work in the other tests. :D

Issue #15063, opened 2017-03-24
Allow running acceptance tests without jQuery


cibernox:
As part of the work of the work performed in https://github.com/cibernox/ember-native-dom-helpers to provide consistent jquery-free test helpers that work both in integration and acceptance I've found that acceptance tests don't work without jQuery.

The reason for that is (at least) this file: https://github.com/emberjs/ember.js/blob/master/packages/ember-testing/lib/setup_for_testing.js#L38-L44

The task is to feature-detect jquery and skip those lines if not present.

For those interested, you can see how to remove jquery from an ember-app in this commit: https://github.com/cibernox/ember-native-dom-helpers/pull/30/commits/5ebeeb61926240949b9a7a43bad9ebadde7dba95

rwjblue:
I submitted #15065 for the specific thing mentioned, but I suspect there are more things that we need to do after that lands.

pixelhandler:
@cibernox @rwjblue should this be in a RFC issue (not PR) instead of here? @Serabe are will still encouraging feature types of issues to go through the RFCs repo?

rwjblue:
@pixelhandler - This isn't really adding a new feature at all.  Simply removing an error condition in one usage scenario.  We are definitely not saying that folks should drop jQuery (only they want to), and we are not modifying the default blueprint.

mixonic:
Are there more items to cover here @cibernox, or can this issue be closed?

rwjblue:
@mixonic - The first known issue was addressed but we need to test again against canary builds. I haven't had a chance yet TBH.

cibernox:
I [just tried](https://github.com/cibernox/ember-native-dom-helpers/pull/48) and no, there is still more work to do.

samjjc:
I would like to work on this. It might take a little time, though. I am new here. any advice to help me get started is welcome

Serabe:
@cibernox ping!

cibernox:
Can u ping me next week? This one I'm on the verge of collapse

pixelhandler:
@cibernox does this issue need to stay open?

cibernox:
Actually, no. Acceptance tests are fine, provided people don't use the built-in global helper (`find`, `fillIn`) but the ones in https://github.com/cibernox/ember-native-dom-helpers

I'll close it.

PR #15065, opened 2017-03-24
[BUGFIX beta] Guard jQuery access in `setupForTesting`.

rwjblue:
Begins addressing #15063.

### DIFF
--- a/packages/ember-testing/lib/setup_for_testing.js
+++ b/packages/ember-testing/lib/setup_for_testing.js
@@ -35,11 +35,13 @@ export default function setupForTesting() {
     setAdapter((typeof self.QUnit === 'undefined') ? new Adapter() : new QUnitAdapter());
   }
 
-  jQuery(document).off('ajaxSend', incrementPendingRequests);
-  jQuery(document).off('ajaxComplete', decrementPendingRequests);
+  if (jQuery) {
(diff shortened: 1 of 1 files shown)


packages/ember-testing/lib/setup_for_testing.js
```diff
@@ -35,11 +35,13 @@ export default function setupForTesting() {
     setAdapter((typeof self.QUnit === 'undefined') ? new Adapter() : new QUnitAdapter());
   }
 
-  jQuery(document).off('ajaxSend', incrementPendingRequests);
-  jQuery(document).off('ajaxComplete', decrementPendingRequests);
+  if (jQuery) {
```
cibernox: can you do `import { jQuery } from 'ember-views';` if there is no jquery? What does that import?

cibernox (approved):

packages/ember-testing/lib/setup_for_testing.js
rwjblue: @cibernox - Yes, it exports `undefined` in that case you can see ([from here](https://github.com/emberjs/ember.js/blob/master/packages/ember-views/lib/system/jquery.js#L3-L30)) that it basically does:

```js
let jQuery

export default jQuery
```

Issue #15062, opened 2017-03-24
[Project Svelte] Deprecation Introduction Hitlist


chadhietala:
## Wanted
We need to know in which versions of Ember these deprecations were introduced. Please simply report the version a specific id was introduced. To ensure that we don't have overlapping efforts, please state what deprecation you are investigating before you do the work.

## Why

With the eventual introduction of Svelte builds, there will be a mechanism to say "strip out any code for deprecated features from 2.0 to 2.8". However, to enable this range-based decision, we need both the start of the range.

## How

Ways to find deprecations:

* `deprecate()` method in the codebase
* `[DEPRECATE]` and `[DEPRECATION]` tags in the [CHANGELOG](https://github.com/emberjs/ember.js/blob/master/CHANGELOG.md).

## Deprecation IDs

- container-lookup-factory 
- ember-application.injected-container
- ember-application.app-instance-registry
- ember-application.registry-resolver-as-function
- ember-application.app-instance-container
- ember-application.app-initializer-initialize-arguments
- ember-application.validate-type
- ember-debug.deprecate-options-missing
- ember-debug.deprecate-id-missing
- ember-debug.deprecate-until-missing
- ember-debug.warn-options-missing
- ember-debug.warn-id-missing
- ember-views.component.defaultLayout
- ember-routing-views.controller-wrapped-param
- ember-htmlbars.make-bound-helper
- ember-htmlbars.ember-handlebars-safestring
- ember-metal.binding
- ember-views.did-init-attrs
- ember-metal.required
- ember-metal.observer-argument-order
- ember-metal.immediate-observer
- ember-views.render-double-modify
- ember-routing.router-resource
- ember-routing.top-level-render-helper
- ember-runtime.controller-proxy
- ember-runtime.action-handler-_actions
- ember-runtime.enumerable-contains
- ember-runtime.will-merge-mixin
- ember-runtime.frozen-copy
- ember-runtime.enumerable-contains
- ember-runtime.freezable-init
- ember-string-utils.fmt
- ember-template-compiler.deprecate-render-model
- ember-template-compiler.deprecate-render
- ember-template-compiler.transform-input-on-to-onEvent.dynamic-value
- ember-template-compiler.transform-input-on-to-onEvent.no-action
- ember-template-compiler.transform-input-on-to-onEvent.normalized-on
- ember-template-compiler.transform-old-binding-syntax
- ember-testing.test-waiters
- ember-views.lifecycle-hook-arguments
[...]

PR #15066, opened 2017-03-24
[Project Svelte] Add deprecation flags

chadhietala:
Currently this is a WIP based on #15062. Need a Confirm/Deny on a couple versions and then we can start marking code paths that can be safely stripped.


jonnii:
@chadhietala there has never been an expectation for svelte builds, is the goal to produce svelte builds for pre-2.12/13 versions? If not would it be ok to have the deprecations only stripped from 2.13 onwards?

rwjblue:
@jonnii - I agree that it seems pretty unlikely that we would support consuming anything older than ember-source@2.14 and getting to "svelte out" deprecated features, but either way we have to instrument the code (for both new and old deprecations) and it seems better to put the "real" version numbers in there "since we are in there"...

jonnii:
@rwjblue makes sense, i suppose it's good to know when a deprecation went into affect outside of the svelte builds. that information could be used in the docs and in places like ember-try. :+1:

locks (approved):

rwjblue:
@chadhietala - This can land, right?

PR #15067, opened 2017-03-25
Decouple test from the default resolver for helpers/helper_registration

josemarluedke:
Part of #15058.

I wasn't sure if I should use `AutobootApplicationTestCase` or `ApplicationTestCase`, so I went with the simpler version: `ApplicationTestCase`.

### DIFF
--- a/packages/ember/tests/helpers/helper_registration_test.js
+++ b/packages/ember/tests/helpers/helper_registration_test.js
@@ -1,112 +1,67 @@
+import { moduleFor, ApplicationTestCase } from 'internal-test-helpers';
 import { Controller, Service, inject } from 'ember-runtime';
-import { run } from 'ember-metal';
-import { compile } from 'ember-template-compiler';
-import {
-  Helper,
-  helper,
-  setTemplates,
-  setTemplate
-} from 'ember-glimmer';
-import { Application } from 'ember-application';
-import { Router } from 'ember-routing';
-import { jQuery } from 'ember-views';
+import { Helper, helper } from 'ember-glimmer';
 
-let App, appInstance;
+moduleFor('Application Lifecycle - Helper Registration', class extends ApplicationTestCase {
+  ['@test Unbound dashed helpers registered on the container can be late-invoked'](assert) {
+    this.addTemplate('application', `<div id='wrapper'>{{x-borf}} {{x-borf 'YES'}}</div>`);
 
-QUnit.module('Application Lifecycle - Helper Registration', {
-  teardown() {
-    run(() => {
-      if (App) {
-        App.destroy();
-      }
+    let myHelper = helper(params => params[0] || 'BORF');
+    this.application.register('helper:x-borf', myHelper);
 
-      App = appInstance = null;
-      setTemplates({});
+    return this.visit('/').then(() => {
+      assert.equal(this.$('#wrapper').text(), 'BORF YES', 'The helper was invoked from the container');
     });
   }
-});
 
-function boot(callback) {
-  run(() => {
-    App = Application.create({
-      name: 'App',
-      rootElement: '#qunit-fixture'
-    });
+  ['@test Bound helpers registered on the container can be late-invoked'](assert) {
+    this.addTemplate('application', `<div id='wrapper'>{{x-reverse}} {{x-reverse foo}}</div>`);
 
-    App.deferReadiness();
-
-    App.Router = Router.extend({
-      location: 'none'
-    });
-
-    // We shouldn't be testing this
-    appInstance = App.__deprecatedInstance__;
-
-    if (callback) { callback(); }
-  });
-
-  let router = appInstance.lookup('router:main');
-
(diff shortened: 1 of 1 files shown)


packages/ember/tests/helpers/helper_registration_test.js
```diff
@@ -1,112 +1,67 @@
...
-      }
+import { Helper, helper } from 'ember-glimmer';
 
-      App = appInstance = null;
-      setTemplates({});
-    });
-  }
-});
+moduleFor('Application Lifecycle - Helper Registration', class extends ApplicationTestCase {
+  ['@test Unbound dashed helpers registered on the container can be late-invoked'](assert) {
```
josemarluedke: I'm not sure what `Unbound` means for this test.

packages/ember/tests/helpers/helper_registration_test.js
```diff
@@ -1,112 +1,67 @@
...
-  boot(() => {
-    App.register('helper:x-borf', myHelper);
-  });
-
-  equal(jQuery('#wrapper').text(), 'BORF YES', 'The helper was invoked from the container');
-});
-
-QUnit.test('Bound helpers registered on the container can be late-invoked', function() {
-  setTemplate('application', compile('<div id=\'wrapper\'>{{x-reverse}} {{x-reverse foo}}</div>'));
+  ['@test Bound helpers registered on the container can be late-invoked'](assert) {
```
josemarluedke: I'm also not sure what `Bound` means for this test. It seems very similar to the previous test.

packages/ember/tests/helpers/helper_registration_test.js
```diff
@@ -1,112 +1,67 @@
...
-function boot(callback) {
-  run(() => {
-    App = Application.create({
-      name: 'App',
-      rootElement: '#qunit-fixture'
-    });
-
-    App.deferReadiness();
+    let myHelper = helper(params => params[0] || 'BORF');
+    this.add('helper:x-borf', myHelper);
```
mixonic: As this test is looking to test helpers "registered on the container" `helper:x-borf` should be registered here instead of added to the resolver.

```
this.application.register('helper:x-borf', myHelper);
```

Instead of adding.

packages/ember/tests/helpers/helper_registration_test.js
```diff
@@ -1,112 +1,67 @@
...
+    this.addTemplate('application', `<div id='wrapper'>{{x-reverse}} {{x-reverse foo}}</div>`);
 
-  boot(() => {
-    appInstance.register('controller:application', Controller.extend({
+    this.add('controller:application', Controller.extend({
       foo: 'alex'
     }));
 
-    appInstance.register('helper:x-reverse', helper(function([ value ]) {
+    this.add('helper:x-reverse', helper(function([ value ]) {
```
mixonic: This should also be registered

packages/ember/tests/helpers/helper_registration_test.js
```diff
@@ -1,112 +1,67 @@
...
+  }
 
-QUnit.test('Undashed helpers registered on the container can be invoked', function() {
-  setTemplate('application', compile('<div id=\'wrapper\'>{{omg}}|{{yorp \'boo\'}}|{{yorp \'ya\'}}</div>'));
+  ['@test Undashed helpers registered on the container can be invoked'](assert) {
+    this.addTemplate('application', `<div id='wrapper'>{{omg}}|{{yorp 'boo'}}|{{yorp 'ya'}}</div>`);
 
-  boot(() => {
-    appInstance.register('helper:omg', helper(() => 'OMG'));
+    this.add('helper:omg', helper(() => 'OMG'));
```
mixonic: `helper:omg` and `helper:yorp` as well.

packages/ember/tests/helpers/helper_registration_test.js
```diff
@@ -1,112 +1,67 @@
...
-  boot(() => {
-    appInstance.register('service:name-builder', Service.extend({
+    this.add('service:name-builder', Service.extend({
       build() {
         serviceCalled = true;
       }
     }));
-    appInstance.register('helper:full-name', Helper.extend({
+
+    this.add('helper:full-name', Helper.extend({
```
mixonic: But leaving this as an `add` seems great.

packages/ember/tests/helpers/helper_registration_test.js
mixonic: bound-ness for this test and the one following refers to the arguments passed to the helper. In this test, a raw string in the template is passed and in the second test a variable is passed.

mixonic (changes requested):
Some small tweaks here to align with that the tests say they are doing, but with that cleanup this looks excellent. Thank you!

josemarluedke:
Updated per feedback and now is ready for another review.

mixonic (approved):

mixonic:
Thank you @josemarluedke! 😁

PR #15068, opened 2017-03-25
Clean router service

Exelord:
This PR will clean unused imports in `RouterService` and improve readability of computed properties.

### DIFF
--- a/packages/ember-routing/lib/services/router.js
+++ b/packages/ember-routing/lib/services/router.js
@@ -7,8 +7,6 @@ import {
   Service,
   readOnly
 } from 'ember-runtime';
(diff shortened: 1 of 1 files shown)


locks (changes requested):
Hi @Exelord, thanks for contributing! Could you leave the computed properties as they were and only remove the unused imports?

Exelord:
OK,
but what about `router: null`. I think this should be added to keep the interface of service clear and explicit.

rwjblue:
`router: null` seems good to me

Also, if you have the time, could you update our linting config to error for these sorts of unused imports (in a follow up PR)?

locks (approved):

PR #15069, opened 2017-03-25
Add queryParams to RouterService

Exelord:
Can we add router params and queryParams to the public interface of RouterService? It will help in many cases when you try to access queryParams in the components.

### DIFF
--- a/packages/ember-routing/lib/services/router.js
+++ b/packages/ember-routing/lib/services/router.js
@@ -7,7 +7,7 @@ import {
   Service,
   readOnly
 } from 'ember-runtime';
-import { get } from 'ember-metal';
+import { get, computed } from 'ember-metal';
 import RouterDSL from '../system/dsl';
 
 /**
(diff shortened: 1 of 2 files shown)


rwjblue:
This router service is still behind a feature flag, and not available to folks.  It is also aiming to implement the specific API that was proposed and accepted in [emberjs/rfcs#95](https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md), which I don't _think_ have these properties.

Adding them would require another RFC I believe...

Exelord:
I think this is a great feature. We should highly consider it even in the case of opening a new RFC.
but... it's your choice. I just wanted to solve a problem that I currently meet in many places on the web.

sdhull:
@Exelord this would be extremely helpful to me like ... right now 😁

lifeart:
up

locks:
I'm closing this for now as the router service RFC does not include these APIs. The next step forward would be to propose a new RFC, and go from there :)
Thanks for the contribution!

PR #15073, opened 2017-03-27
Decouple controller_test from globals resolver 

defraz:
#15058


mixonic:
Excellent. Thank you @defraz!

PR #15074, opened 2017-03-27
[DOC] Fix controllerFor typo

yogipatel:
`Ember.route#controllerFor` was documented as returning a model instead of a controller. This PR corrects that.

### DIFF
--- a/packages/ember-routing/lib/system/route.js
+++ b/packages/ember-routing/lib/system/route.js
@@ -1762,7 +1762,7 @@ let Route = EmberObject.extend(ActionHandler, Evented, {
   },
 
   /**
-    Returns the resolved controller of the current route, or a parent (or any ancestor)
+    Returns the controller of the current route, or a parent (or any ancestor)
     route in a route hierarchy.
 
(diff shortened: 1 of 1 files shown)


packages/ember-routing/lib/system/route.js
```diff
@@ -1762,7 +1762,7 @@ let Route = EmberObject.extend(ActionHandler, Evented, {
   },
 
   /**
-    Returns the resolved model of the current route, or a parent (or any ancestor)
+    Returns the resolved controller of the current route, or a parent (or any ancestor)
```
rwjblue: I realize this isn't really related to your change, but we should probably remove `resolved` from here.  We don't do any promise resolution, and it looks like this was just copy/pasta'ed from `modelFor` (which does resolve promisey things).

**tldr;** Would you mind removing `resolved` here?

yogipatel: 👍 I was wondering the same thing.

mmun:
Thank you.

PR #15075, opened 2017-03-27
Decouple router_map_test from globals resolver

josemarluedke:
Part of #15058.

### DIFF
--- a/packages/ember/tests/routing/router_map_test.js
+++ b/packages/ember/tests/routing/router_map_test.js
@@ -1,88 +1,40 @@
+import { moduleFor, ApplicationTestCase } from 'internal-test-helpers';
 import { run } from 'ember-metal';
-import { compile } from 'ember-template-compiler';
-import { Application } from 'ember-application';
 import { Router } from 'ember-routing';
-import { jQuery } from 'ember-views';
-import { setTemplates, setTemplate } from 'ember-glimmer';
 
-let router, App, container;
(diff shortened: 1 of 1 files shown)


packages/ember/tests/routing/router_map_test.js
```diff
@@ -1,88 +1,38 @@
...
+    this.router.map(function() {
+      this.route('goodbye');
     });
-  },
 
-  teardown() {
     run(() => {
-      App.destroy();
-      App = null;
+      this.visit('/hello').then(() => this.assertText('Hello!'));
```
mixonic: In theory, promises may be async. The `visit` in this runloop may not complete before the on in the following test completes.

Practically, these two loops will render synchronously because of our current promise implementation (promises resolve in the runloop) however it isn't really future safe.

Can you give something like the following a try?

```js
return run(() => {
  return this.visit('/hello').then(() => {
    this.assertText('Hello!');
  }).then(() => {
    return this.visit('/goodbye');
  }).then(() => {
    this.assertText('Goodbye!');
  });
});
```

mixonic (changes requested):
Asked for a small cleanup, but this looks like a good start. Thank @josemarluedke

josemarluedke:
@mixonic updated!

mixonic (approved):

mixonic:
Great :-) Thank you @josemarluedke!

PR #15078, opened 2017-03-27
[DEPRECATION] #14754 deprecate `canDispatchToEventManager`

habdelra:
This is to address the deprecation in #14754.  I'm not sure if the deprecation message provides enough detail.

### DIFF
--- a/packages/ember-glimmer/tests/integration/event-dispatcher-test.js
+++ b/packages/ember-glimmer/tests/integration/event-dispatcher-test.js
@@ -6,6 +6,7 @@ import {
   run
 } from 'ember-metal';
 import { EMBER_IMPROVED_INSTRUMENTATION } from 'ember/features';
+import { EventDispatcher } from 'ember-views';
 
 let canDataTransfer = !!document.createEvent('HTMLEvents').dataTransfer;
 
@@ -136,6 +137,34 @@ moduleFor('EventDispatcher#setup', class extends RenderingTest {
     this.render(`{{x-foo}}`);
 
     this.$('div').trigger('myevent');
+  }
+
+  ['@test canDispatchToEventManager is deprecated'](assert) {
+    this.dispatcher.canDispatchToEventManager = null;
+    this.registerComponent('x-foo', {
+      ComponentClass: Component.extend({
+        eventManager: {
+          myEvent() {}
+        }
+      }),
+      template: `<p>Hello!</p>`
+    });
+
+    expectDeprecation(() => {
+      this.render(`{{x-foo}}`);
+    }, '[DEPRECATED] `canDispatchToEventManager` has been deprecated.');
+
+    this.$('div').trigger('myevent');
+  }
+
+  ['@test canDispatchToEventManager is deprecated in EventDispatcher'](assert) {
+    let MyDispatcher = EventDispatcher.extend({
+      canDispatchToEventManager: null
+    });
+
+    expectDeprecation(() => {
+      MyDispatcher.create();
+    }, '[DEPRECATED] `canDispatchToEventManager` has been deprecated.');
   }
 
   ['@test a rootElement can be specified'](assert) {
--- a/packages/ember-glimmer/tests/integration/event-dispatcher-test.js
+++ b/packages/ember-glimmer/tests/integration/event-dispatcher-test.js
@@ -56,6 +56,8 @@ moduleFor('EventDispatcher', class extends RenderingTest {
       template: `<input id="is-done" type="checkbox">`
     });
 
+
+    expectDeprecation(/`eventManager` has been deprecated/);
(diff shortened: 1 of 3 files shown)


packages/ember-glimmer/tests/integration/event-dispatcher-test.js
```diff
@@ -138,6 +139,36 @@ moduleFor('EventDispatcher#setup', class extends RenderingTest {
...
+            assert.ok(true, 'custom event was triggered');
+          }
+        }
+      }),
+      template: `<p>Hello!</p>`
+    });
+
+    expectDeprecation(() => {
+      this.render(`{{x-foo}}`);
+    }, '[DEPERECATED] `canDispatchToEventManager` has been deprecated.');
```
jpadilla: Should be `[DEPRECATED]` instead of `[DEPERECATED]`.

packages/ember-glimmer/tests/integration/event-dispatcher-test.js
```diff
@@ -138,6 +139,36 @@ moduleFor('EventDispatcher#setup', class extends RenderingTest {
     this.$('div').trigger('myevent');
   }
 
+  ['@test canDispatchToEventManager is deprecated'](assert) {
+    this.dispatcher.canDispatchToEventManager = null;
+    this.registerComponent('x-foo', {
+      ComponentClass: Component.extend({
+        eventManager: {
+          myEvent() {
+            assert.ok(true, 'custom event was triggered');
```
mmun: It's not clear from how this test is written that this line of code is actually executing. Can you set a boolean and check it at the end, or use assert.expect?

mmun (commented):
Thank you for participating in the contributor's workshop!! :)

packages/ember-glimmer/tests/integration/event-dispatcher-test.js
habdelra: whoops--thats copy-pasta, that assert is completely unrelated to this test

habdelra:
@rwjblue could I get your

packages/ember-glimmer/tests/integration/event-dispatcher-test.js
```diff
@@ -138,6 +139,34 @@ moduleFor('EventDispatcher#setup', class extends RenderingTest {
     this.$('div').trigger('myevent');
   }
 
+  ['@test canDispatchToEventManager is deprecated'](assert) {
+    this.dispatcher.canDispatchToEventManager = null;
```
rwjblue: In theory, we should not need to force this to be `null` in this test, the default of `undefined` _should_ be enough

packages/ember-views/lib/mixins/view_support.js
```diff
@@ -538,7 +538,16 @@ export default Mixin.create({
       let owner = getOwner(this);
       let dispatcher = owner && owner.lookup('event_dispatcher:main');
 
-      if (dispatcher && dispatcher.canDispatchToEventManager === null) {
+      deprecate(
+        `[DEPRECATED] \`canDispatchToEventManager\` has been deprecated.`,
```
rwjblue: Let's have this message mention `eventManager` instead of `canDispatchToEventManager`, and also add the ${this} to the interpolation (so it's clear to someone seeing the deprecation which component caused it.

packages/ember-views/lib/mixins/view_support.js
```diff
@@ -538,7 +538,16 @@ export default Mixin.create({
...
+        !('canDispatchToEventManager' in dispatcher),
+        {
+          id: 'ember-views.event-dispatcher.canDispatchToEventManager',
+          until: '3.0.0'
```
rwjblue: Let's change this to 2.16.0

packages/ember-views/lib/system/event_dispatcher.js
```diff
@@ -129,13 +130,22 @@ export default EmberObject.extend({
...
     @private
   */
-  canDispatchToEventManager: null,
 
   init() {
     this._super();
     assert('EventDispatcher should never be instantiated in fastboot mode. Please report this as an Ember bug.', environment.hasDOM);
+
+    deprecate(
+      `[DEPRECATED] \`canDispatchToEventManager\` has been deprecated.`,
```
rwjblue: Let's mention which file this is in.

rwjblue: Also, you should not need to prefix the message with `[DEPRECATED]`

rwjblue (changes requested):

habdelra:
@rwjblue good morning sunshine, I need more

rwjblue (approved):

rwjblue:
Thank you @habdelra!

PR #15079, opened 2017-03-27
Decouple container_debug_adapter_test from globals resolver

josemarluedke:
Part of #15058.

### DIFF
--- a/packages/ember-extension-support/tests/container_debug_adapter_test.js
+++ b/packages/ember-extension-support/tests/container_debug_adapter_test.js
@@ -1,52 +1,47 @@
+import { moduleFor, ApplicationTestCase } from 'internal-test-helpers';
+import { assign } from 'ember-utils';
 import { run } from 'ember-metal';
 import { Controller as EmberController } from 'ember-runtime';
 import '../index'; // Must be required to export Ember.ContainerDebugAdapter.
-import { Application } from 'ember-application';
 
-let adapter, App, appInstance;
+let adapter;
 
-function boot() {
-  run(App, 'advanceReadiness');
-}
+moduleFor('Container Debug Adapter', class extends ApplicationTestCase {
+  constructor() {
+    super();
+    adapter = this.application.__deprecatedInstance__.lookup('container-debug-adapter:main');
+  }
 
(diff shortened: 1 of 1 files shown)


packages/ember-extension-support/tests/container_debug_adapter_test.js
```diff
@@ -1,52 +1,47 @@
...
+    assert.equal(adapter.canCatalogEntriesByType('controller'), true, 'canCatalogEntriesByType should return true for controller');
+    assert.equal(adapter.canCatalogEntriesByType('route'), true, 'canCatalogEntriesByType should return true for route');
+    assert.equal(adapter.canCatalogEntriesByType('view'), true, 'canCatalogEntriesByType should return true for view');
+  }
 
-QUnit.test('the default ContainerDebugAdapter catalogs controller entries', function() {
-  App.PostController = EmberController.extend();
-  let controllerClasses = adapter.catalogEntriesByType('controller');
+  ['@test default ContainerDebugAdapter catalogs controller entries'](assert) {
+    this.application.PostController = EmberController.extend()
```
josemarluedke: It seems that is not possible to use application registry for the controller since the actually implementation of `catalogEntriesByType` looks for properties in the app.

packages/ember-extension-support/tests/container_debug_adapter_test.js
```diff
@@ -1,52 +1,47 @@
...
-let adapter, App, appInstance;
+let adapter;
 
-function boot() {
-  run(App, 'advanceReadiness');
-}
+moduleFor('Container Debug Adapter', class extends ApplicationTestCase {
+  constructor() {
+    super();
+    adapter = this.application.__deprecatedInstance__.lookup('container-debug-adapter:main');
```
josemarluedke: Using the `autoboot: true` did not create an application instance, so I had to keep `__deprecatedInstance__ `.

mmun: Good point. We should make sure that `this.applicationInstance` is available if autoboot is specified.

mmun:
Thanks for the PR. @rwjblue has given his blessing.

PR #15081, opened 2017-03-27
Ember utils polyfill

### LINKED ISSUE #14855: [Performance] ember-utils assign should expose Object.assign or polyfill
Hello, 
Emberjs has a `assign` method that basicly polyfill's ES6 Object.assign:
https://github.com/emberjs/ember.js/blob/master/packages/ember-utils/lib/assign.js

`Ember.assign` is defined as `Object.assign`(if existent) or the `assign`  polyfill's(above).
https://github.com/emberjs/ember.js/blob/master/packages/ember/lib/index.js#L25

Many internal packages import `assign` function directly from ember-utils, the **native** implementation(when existent) is never used internally.

Ideia: `ember-utils/lib/assign.js` should exposes 2 functions
```js
export function assign(original) {
  // polyfill implementation
}

export default Object.assign || assign;
```
So we can import `assign` polyfill  and test it properly, `Object.assign` will be used internally through ember-utils `assign`.

Let me know what you guys think about it (y)

jkusa:
__Changes__

This PR resolves #14855 

 - Use native `Object.assign` instead of polyfill if available in `Ember.assign`
 - Expose `assignPolyfill` for direct testing
- Add test coverage to `Object.assign` polyfill

__TODO__
 - [ ] Run benchmarks to make sure that using native `Object.assign` doesn't cause regressions

### DIFF
--- a/packages/ember-utils/lib/assign.js
+++ b/packages/ember-utils/lib/assign.js
@@ -15,7 +15,7 @@
   @return {Object}
   @public
 */
-export default function assign(original) {
+export function assign(original) {
   for (let i = 1; i < arguments.length; i++) {
     let arg = arguments[i];
     if (!arg) { continue; }
@@ -30,3 +30,5 @@ export default function assign(original) {
 
   return original;
 }
+
+export default Object.assign || assign;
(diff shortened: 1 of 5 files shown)


rwjblue (commented):
Looks good, can you also uncomment https://github.com/emberjs/ember.js/blob/master/packages/ember/tests/reexports_test.js#L11 now (it should pass)?

mmun:
🐎 🐎 🐎 🐎 🚀

Issue #14780, opened 2017-01-03
Deprecate calling `this._super` when there is no method on parent class.


rwjblue:
We should not support calling _super when the parent class does not have a method defined (aka when there is no _super). This likely means adding a deprecation for now, and converting that to an error in 3.0.0.

Examples:

```js
let A = Ember.Object.extend({
  foo() { return 'A - Foo'; },

  qux: Ember.computed(function() {
    return 'A - qux';
  })
});

let B = A.extend({
  bar() {
    // should trigger deprecation
    return this._super(...arguments) + '| B - Bar';
  },

  foo() {
    // should not trigger deprecation
    return this._super(...arguments) + ' | B - Foo | ' + this.bar();
  },

  qux: Ember.computed(function() {
    // should not trigger deprecation
    return this._super(...arguments) + ' | B - qux';
  }),

  baz: Ember.computed(function() {
    // should trigger deprecation
    return this._super(...arguments) + ' | B - baz';
  })
});
```

See [original context](https://github.com/emberjs/ember.js/pull/13242#issuecomment-208076203)

workmanw:
One thing that came to mind was the component lifecycle hooks (`didReceiveAttrs`, `didUpdateAttrs`, `willRender`, etc). It appears that this hooks are not defined as no-op functions on the component class [ember-glimmer/lib/component.js](https://github.com/emberjs/ember.js/blob/f4d573efd1f375efac4a4c8e37d3d8e99824024b/packages/ember-glimmer/lib/component.js#L303-L401).

It seems like this would trigger a deprecation:
```javascript
export default Ember.Component.extend({
  didReceiveAttrs() {
    this._super();
    /* ... */
  }
});
```

That feels like it could be slightly problematic in that it requires each component class that extends another to know whether it's parent has implemented the lifecycle hook if it too wants to use that hook. And even then, it creates a bad practice in that you should always call `_super` on hooks. 

Seems like one native solution would be to implement no-op functions on the `Ember.Component` class for each of these hooks. Thoughts? I could submit a PR if this seems favorable.

rwjblue:
Yes, that is absolutely the same conclusion made in the comment that I linked to in the main issue description. PR's to ensure that all "framework hook" methods are properly defined are definitely welcome. :smiling_cat:

krisselden:
[...]

PR #12062, opened 2015-08-12
Remove this.__nextSuper

krisselden:
This just ensures that the superFunction is wrapped with a terminal _super instead of relying on this.__nextSuper = null while invoked to prevent recursion. Allows wrapped functions not to have to go through 2 argument applies, makes debugging a lot easier and performs better.

PR #15082, opened 2017-03-27
[DEPRECATION] (WIP) calling `this._super` when there is no method on parent class

givanse:
https://github.com/emberjs/ember.js/issues/14780

 - [ ] document  `didDestroyElement`, only available in interactive mode

### DIFF
--- a/packages/ember-metal/lib/mixin.js
+++ b/packages/ember-metal/lib/mixin.js
@@ -95,6 +95,15 @@ function giveDescriptorSuper(meta, key, property, values, descs, base) {
   }
 
   if (superProperty === undefined || !(superProperty instanceof ComputedProperty)) {
+    /*
+    deprecate(
+      'Calling `_super` when there is no computed on the parent class is deprecated.',
+      false,
+      { id: 'ember-metal.missing-super-computed', until: '3.0.0' }
+    );
+    */
+    console.warn('no super computed');
+
     return property;
   }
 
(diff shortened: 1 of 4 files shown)


packages/ember-metal/lib/mixin.js
```diff
@@ -129,6 +138,23 @@ function giveMethodSuper(obj, key, method, values, descs) {
...
+    /*
+    deprecate(
+      'Calling `_super` when there is no method on the parent class is deprecated.',
+      false,
+      { id: 'ember-metal.missing-super-computed', until: '3.0.0' }
+    );
+    *
+    */
+    /*
+    deprecate(
```
givanse: copy-pasta from within this file as a sanity check, it errors with:
```
Uncaught Error: Ember.required is deprecated as its behavior is inconsistent and unreliable.
Uncaught TypeError: Cannot read property 'registerDeprecationHandler' of undefined
Uncaught TypeError: Cannot read property 'extend' of undefined
```

packages/ember-metal/lib/mixin.js
```diff
@@ -129,6 +138,23 @@ function giveMethodSuper(obj, key, method, values, descs) {
 
   // Only wrap the new method if the original method was a function
   if (superMethod === undefined || 'function' !== typeof superMethod) {
+    /*
+    deprecate(
```
givanse:
```
Uncaught Error: Calling `_super` when there is no method on the parent class is deprecated.
```

:thinking:

givanse: `deprecate` is meant to be executed at app runtime

givanse:
The last broken test is:

> Ember.Mixin#reopen: using reopen() and calling _super where there is not a super function does not cause infinite recursion

Previous fixes are found in these PRs https://github.com/emberjs/ember.js/pull/9678 https://github.com/emberjs/ember.js/pull/12062 I'm still trying to piece together what is going on with that code, any hints are appreciated.

locks:
Apologies for letting this work linger all this time 😞  At this point I think rebasing this PR would be more work than starting anew so I'm closing it for now.

Issue #14797, opened 2017-01-07
Missing API documentation for Ember.Templates.helpers.html#method_yield


callovarne:
The docs for `{{yield}}` have gone missing. I'm not having much luck finding it in the source either. API change?

The page appears to have existed here at some point: http://emberjs.com/api/classes/Ember.Templates.helpers.html#method_yield

Dead links:
* http://emberjs.com/api/classes/Ember.View.html#toc_layouts

The guides are thin on information, too, and don't link to the API page. I was looking for specifics on `{{yield to="inverse"}}`.

locks:
Relatedly, we removed the "Layouts" section when moving the documentation to the Ember.Component class: https://github.com/emberjs/ember.js/pull/14765.

mixonic:
Oh gee, 😬 there appear to be several issues here:

* The docs for yield are not appearing
* There are dead links in the Ember.View ToC
* The guides have no information on `{{yield to="inverse"}}`. I think this API should 100% be mentioned in the API docs, I'll defer to the learning team on if they want to present it in the guides. Regardless that would be an issue in the guides repo, lets keep it focused to API docs here. So do the API docs document this?
* The "Layouts" section was dropped. I'm not at all sure what this has to do this this issue.

yogipatel:
@callovarne @mixonic I looked into this a bit, but I'm very much a beginner in Ember.js land. AFACIT, yield hasn't had documentation. I looked at 2.10, 2.7, even 2.0; the Ember.Templates.helpers.html#method_yield link has probably been dead for a while. Perhaps it lived somewhere in HTMLBars docs?

As @locks mentioned, the broken link is now in Ember.Component here: https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/component.js#L432

I'm probably not at the level where I should be doc'ing yield, but if you want a hand digging deeper, I'd be happy to help.

elucid:
My best guess is that the yield docs disappeared here: https://github.com/emberjs/ember.js/commit/816f252580e0e83d7453d6284a0741e99747fee3

yogipatel:
@elucid Nice find. Here it is: https://github.com/emberjs/ember.js/commit/816f252580e0e83d7453d6284a0741e99747fee3#diff-6de598048f2facdff459c26b8d8746da

Question is, where is it implemented / do the docs belong now?

mattmarcum:
[...]

PR #15083, opened 2017-03-27
[DOC] update missing yield docs in api #14797

mattmarcum:
For [14979](https://github.com/emberjs/ember.js/issues/14797)

Just getting the ball running on these missing docs.  These are copied over from the old docs that were removed, with the section on Ember.View removed.  Let me know if you want anymore changes.

### DIFF
--- a/packages/ember-glimmer/lib/index.js
+++ b/packages/ember-glimmer/lib/index.js
@@ -135,6 +135,31 @@
    First name: <input type="text" />
   </label>
   ```
+
+  Additionally you can `yield` properties into the context for use by the consumer:
+
+  ```handlebars
+  <!-- application.hbs -->
+  {{#labeled-textfield value=someProperty validator=(action 'firstNameValidator') as |validationError|}}
+  {{#if validationError}}
+    <p class="error">{{ValidationError}}</p>
+  {{/if}}
+  First name:
+  {{/labeled-textfield}}
+  ```
+  ```handlebars
+  <!-- components/labeled-textfield.hbs -->
+  <label>
+   {{yield validationError}} {{input value=value}}
+  </label>
+  ```
+  Result:
+  ```html
+  <label>
+   <p class="error">First Name must be at least 3 characters long.</p>
+   First name: <input type="text" />
+  </label>
+  ```
   @method yield
   @for Ember.Templates.helpers
   @param {Hash} options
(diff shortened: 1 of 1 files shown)


mmun:
This looks great. Would you be able change the example to yield an argument? or add an additional example?

Additionally, we should probably pull this into the release channel. Does that sound good @rwjblue?

mattmarcum:
Ok, I gave it a go

packages/ember-glimmer/lib/index.js
```diff
@@ -110,6 +110,62 @@
   @public
  */
 
+/**
+  `{{yield}}` denotes an area of a template that will be rendered inside
+  of another template.
+
+  ### Use with Ember.Component
+  When designing components `{{yield}}` is used to denote where, inside the component's
```
locks: add empty new line between the header and the text please

packages/ember-glimmer/lib/index.js
```diff
@@ -110,6 +110,62 @@
...
+  template, an optional block passed to the component should render:
+  ```handlebars
```
locks: empty lines around the code block please:

    should render:

    ```handlebars
    ...
    ```

    ```handlebars
    …
    ```

packages/ember-glimmer/lib/index.js
```diff
@@ -110,6 +110,62 @@
...
+  <!-- application.hbs -->
+  {{#labeled-textfield value=someProperty validator=(action 'firstNameValidator') as |validationError|}}
+  {{#if validationError}}
+    <p class="error">{{ValidationError}}</p>
+  {{/if}}
+  First name:
+  {{/labeled-textfield}}
+  ```
+  ```handlebars
```
locks: same whitespace here

packages/ember-glimmer/lib/index.js
locks: you can also do

    ```app/application.hbs

instead of the comment, and our system will render a nice header with the path

locks (changes requested):
Some whitespace and code block tweaks necessary.

mixonic:
@locks there is an update here if you can review, thanks 🙇

locks (approved):

locks:
We are good to gooooo, thanks @mattmarcum for sticking the landing :)

PR #15084, opened 2017-03-27
Decouple test from the default resolver for system/bootstrap-test

ballPointPenguin:
This was done at the EmberConf contributors workshop as part of https://github.com/emberjs/ember.js/issues/15058


mixonic:
Thank you @ballPointPenguin! 👏

PR #15087, opened 2017-03-27
Decouple ember-testing/helpers_test from default resolver

julienfitz:
#15058


mmun:
Thank you. This is a great start. We need to also migrate the body's of each test into using the ApplicationTestCase API's e.g. `this.add(...)` as opposed to relying on globals.

mixonic:
Thank you for the work on this @lbaillie. I'm going to close it as the patch is superseded by https://github.com/emberjs/ember.js/pull/15337.

PR #15088, opened 2017-03-27
Remove use of private method makeArray in examples

jkusa:
`Ember.makeArray` is [marked as private](https://github.com/emberjs/ember.js/blob/52d9cf213f802f5612e05a880480594e32fc64f8/packages/ember-utils/lib/make-array.js#L24). This PR removes the use of it in the `Ember.typeOf` documentation example.

### DIFF
--- a/packages/ember-runtime/lib/utils.js
+++ b/packages/ember-runtime/lib/utils.js
@@ -89,7 +89,7 @@ export function isArray(obj) {
   Ember.typeOf(new Number(101));        // 'number'
   Ember.typeOf(true);                   // 'boolean'
   Ember.typeOf(new Boolean(true));      // 'boolean'
-  Ember.typeOf(Ember.makeArray);        // 'function'
+  Ember.typeOf(Ember.A);                // 'function'
   Ember.typeOf([1, 2, 90]);             // 'array'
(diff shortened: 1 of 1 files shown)


packages/ember-runtime/lib/utils.js
```diff
@@ -89,7 +89,7 @@ export function isArray(obj) {
   Ember.typeOf(new Number(101));        // 'number'
   Ember.typeOf(true);                   // 'boolean'
   Ember.typeOf(new Boolean(true));      // 'boolean'
-  Ember.typeOf(Ember.makeArray);        // 'function'
+  Ember.typeOf(Ember.A);                // 'function'
```
rwjblue: I think this should invoke `Ember.A()` (the old version also was wrong)

jkusa: I thought this was showing the return type `'function'`. 

Do we need the examples:
 - `Ember.typeOf(Ember.A())   // 'array'`
 - `Ember.typeOf(Ember.get)   // 'function'`

rwjblue: DOH! You are correct

locks:
Seems we are good here!

cibernox:
Ember.makeArray is private? 😯

PR #15091, opened 2017-03-27
Decouple test from the default resolver for data_adapter

Gaurav0:
For #15058

### DIFF
--- a/packages/ember-extension-support/tests/data_adapter_test.js
+++ b/packages/ember-extension-support/tests/data_adapter_test.js
@@ -7,224 +7,231 @@ import {
 } from 'ember-metal';
 import { Object as EmberObject, A as emberA } from 'ember-runtime';
 import EmberDataAdapter from '../data_adapter';
-import { Application as EmberApplication, Resolver as DefaultResolver } from 'ember-application';
+import { moduleFor, ApplicationTestCase } from 'internal-test-helpers';
+
 
 let adapter, App;
 const Model = EmberObject.extend();
 
+const PostClass = Model.extend();
+
 const DataAdapter = EmberDataAdapter.extend({
   detect(klass) {
     return klass !== Model && Model.detect(klass);
-  }
-});
-
-QUnit.module('Data Adapter', {
-  setup() {
-    run(() => {
-      App = EmberApplication.create();
-      App.toString = function() { return 'App'; };
-      App.deferReadiness();
-      App.register('data-adapter:main', DataAdapter);
-    });
   },
-  teardown() {
-    run(() => {
-      adapter.destroy();
-      App.destroy();
-    });
-  }
-});
-
-QUnit.test('Model types added with DefaultResolver', function() {
-  App.Post = Model.extend();
-
-  adapter = App.__container__.lookup('data-adapter:main');
-  adapter.reopen({
-    getRecords() {
-      return emberA([1, 2, 3]);
-    },
-    columnsForType() {
-      return [{ name: 'title', desc: 'Title' }];
-    }
-  });
-
-  run(App, 'advanceReadiness');
-
-  function modelTypesAdded(types) {
-    equal(types.length, 1);
-    let postType = types[0];
-    equal(postType.name, 'post', 'Correctly sets the name');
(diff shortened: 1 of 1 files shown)


packages/ember-extension-support/tests/data_adapter_test.js
```diff
@@ -8,223 +8,233 @@ import {
...
-      App.register('data-adapter:main', DataAdapter);
-    });
   },
-  teardown() {
-    run(() => {
-      adapter.destroy();
-      App.destroy();
+  init() {
+    this._super(...arguments);
+    let self = this;
```
mixonic: I don't believe `self` is used here

packages/ember-extension-support/tests/data_adapter_test.js
```diff
@@ -8,223 +8,233 @@ import {
...
-    deepEqual(postType.columns, [{ name: 'title', desc: 'Title' }], 'Correctly sets the columns');
-  }
 
-  run(adapter, 'watchModelTypes', modelTypesAdded);
-});
+moduleFor('Data Adapter', class extends ApplicationTestCase {
 
-QUnit.test('getRecords gets a model name as second argument', function() {
-  App.Post = Model.extend();
+  ['@test Model types added with DefaultResolver']() {
```
mixonic: So this test is specifically mentioning `DefaultResolver`. By migrating to the `ApplicationTestCase` and using `add` etc, it is no longer using the default resolver. Instead it now uses the TestResolver.

Either the test description is wrong and the data adapter doesn't require the DefaultResolver, or this test should not be ported. I''m not sure which.

packages/ember-extension-support/tests/data_adapter_test.js
```diff
@@ -8,223 +8,233 @@ import {
...
-    });
   },
-  teardown() {
-    run(() => {
-      adapter.destroy();
-      App.destroy();
+  init() {
+    this._super(...arguments);
+    let self = this;
+    this.set('containerDebugAdapter', {
```
mixonic: I don't think this property is used?

packages/ember-extension-support/tests/data_adapter_test.js
```diff
@@ -8,223 +8,233 @@ import {
 import { Object as EmberObject, A as emberA } from 'ember-runtime';
 import EmberDataAdapter from '../data_adapter';
 import { Application as EmberApplication, Resolver as DefaultResolver } from 'ember-application';
+import { moduleFor, ApplicationTestCase } from 'internal-test-helpers';
+import { ModuleBasedResolver } from 'internal-test-helpers/test-resolver';
```
mixonic: This import is not used.

mixonic (changes requested):
I believe that first test can just have its description changed. A read of https://github.com/emberjs/ember.js/blob/master/packages/ember-extension-support/lib/data_adapter.js doesn't make me believe these tests are specific to the default resolver, nor do these tests.

Some small cleanup remains. Thanks @Gaurav0!

packages/ember-extension-support/tests/data_adapter_test.js
Gaurav0: Tests fail without a mock containerDebugAdapter.

packages/ember-extension-support/tests/data_adapter_test.js
Gaurav0: Fixed.

packages/ember-extension-support/tests/data_adapter_test.js
Gaurav0: Fixed.

packages/ember-extension-support/tests/data_adapter_test.js
Gaurav0: Removing DefaultResolver from description

packages/ember-extension-support/tests/data_adapter_test.js
mixonic: I misread this the first time, looks good 👍

mixonic (approved):

mixonic:
Thank you @Gaurav0!

PR #15100, opened 2017-04-01
Decouple controller_for test from globals resolver

stevenwu:
Thanks @mmun for the help.

#15058

### DIFF
--- a/packages/ember-routing/tests/system/controller_for_test.js
+++ b/packages/ember-routing/tests/system/controller_for_test.js
@@ -1,112 +1,59 @@
-import { get, run } from 'ember-metal'; // A
 import {
-  Namespace,
-  String as StringUtils,
   Controller
 } from 'ember-runtime';
 import controllerFor from '../../system/controller_for';
 import generateController from '../../system/generate_controller';
-import { buildOwner } from 'internal-test-helpers';
 import { EMBER_NO_DOUBLE_EXTEND } from 'ember/features';
+import { moduleFor, ApplicationTestCase } from 'internal-test-helpers';
 
-function buildInstance(namespace) {
-  let owner = buildOwner();
+moduleFor('Ember.controllerFor', class extends ApplicationTestCase {
+  ['@test controllerFor should lookup for registered controllers'](assert) {
+    this.add('controller:app', Controller.extend());
 
-  owner.__registry__.resolver = resolverFor(namespace);
-  owner.registerOptionsForType('view', { singleton: false });
+    return this.visit('/').then(() => {
+      let appInstance = this.applicationInstance;
+      let appController = appInstance.lookup('controller:app');
+      let controller = controllerFor(appInstance, 'app');
+      assert.equal(appController, controller, 'should find app controller');
+    });
+  }
+});
 
-  owner.register('application:main', namespace, { instantiate: false });
+moduleFor('Ember.generateController', class extends ApplicationTestCase {
+  ['@test generateController should return Ember.Controller'](assert) {
+    return this.visit('/').then(() => {
+      let controller = generateController(this.applicationInstance, 'home');
+      assert.ok(controller instanceof Controller, 'should return controller');
+    });
+  }
 
-  owner.register('controller:basic', Controller, { instantiate: false });
+  ['@test generateController should return App.Controller if provided'](assert) {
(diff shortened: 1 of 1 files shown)


packages/ember-routing/tests/system/controller_for_test.js
```diff
@@ -1,112 +1,59 @@
...
+    let MainController = Controller.extend();
+    this.add('controller:basic', MainController);
 
-  ok(controller instanceof Controller, 'should return controller');
-});
-
-
-QUnit.test('generateController should return App.Controller if provided', function() {
-  let controller;
-  namespace.Controller = Controller.extend();
```
mmun: This line became https://github.com/emberjs/ember.js/pull/15100/files#diff-5251a48efdb95856746f72151cc33bbbR32, but I would have expected the factory be registered as `controller:main` and not `controller:basic`.

mixonic: Seems like it is `basic` https://github.com/emberjs/ember.js/blob/ee8ab348484421f254fe995c6f04f0d4a0859cfa/packages/ember-routing/lib/system/generate_controller.js#L19 Works in the resolver b/c `useRouterNaming` is called for controllers https://github.com/emberjs/ember.js/blob/master/packages/ember-application/lib/system/resolver.js#L288 barf.

packages/ember-routing/tests/system/controller_for_test.js
```diff
@@ -1,112 +1,59 @@
...
+    return this.visit('/').then(() => {
+      let controller = generateController(this.applicationInstance, 'home');
+      assert.ok(controller instanceof Controller, 'should return controller');
     });
   }
-});
 
-QUnit.test('generateController should return Ember.Controller', function() {
-  let controller = generateController(appInstance, 'home');
+  ['@test generateController should return App.Controller if provided'](assert) {
```
mixonic: Can this test be changed to `should return controller:basic if resolved` instead? The global `App.Controller` has no relevance to the remaining test (and is less informative).

packages/ember-routing/tests/system/controller_for_test.js
```diff
@@ -1,112 +1,59 @@
...
+    });
+  }
 
-  if (EMBER_NO_DOUBLE_EXTEND) {
-    ok(controller instanceof BasicController, 'should return base class of controller');
-  } else {
-    let doubleExtendedFactory;
-    ignoreDeprecation(() => {
-      doubleExtendedFactory = appInstance._lookupFactory('controller:basic');
+  ['@test generateController should return controller:basic if provided'](assert) {
```
mixonic: This test was poorly named. It was testing the case of a registered factory. The test above it tests the resolved case.

As such, this test should not use `this.add('controller:basic'` but instead should register the factory:

* `this.application.register('controller:basic'`
* Even more like the old test would be `this.applicationInstance.register('controller:basic'` after the `visit` call.

Finally the test name should be `@test generateController should return controller:basic if registered` (note "registered" vs. "resolved" in the above test).

mixonic (changes requested):
I've requested some very small cleanup but this is pretty close!

packages/ember-routing/tests/system/controller_for_test.js
mmun: @mixonic Whoa. Didn't expect that!

packages/ember-routing/tests/system/controller_for_test.js
```diff
@@ -1,112 +1,58 @@
...
 
-  if (EMBER_NO_DOUBLE_EXTEND) {
-    ok(controller instanceof BasicController, 'should return base class of controller');
-  } else {
-    let doubleExtendedFactory;
-    ignoreDeprecation(() => {
-      doubleExtendedFactory = appInstance._lookupFactory('controller:basic');
+  ['@test generateController should return controller:basic if registered'](assert) {
+    let BasicController = Controller.extend();
+    this.application.register('controller:basic', BasicController);
```
stevenwu: @mixonic I tried `this.applicationInstance.register('controller:basic', BasicController)` after `visit` but the test fails.

mixonic (approved):

mixonic:
Excellent, thank you @stevenwu!

PR #15102, opened 2017-04-02
Remove sourcemaps

chadhietala:


rwjblue (approved):

chadhietala:
@rwjblue I think Travis may be doing something new and it's just exiting because the browser tests are taking too long.

Turbo87:
@chadhietala it would be awesome to know *why* sourcemaps are removed. I would very much like to have sourcemaps for the final build output that is distributed on npm and bower. Without sourcemaps it becomes much harder to figure out from what original file an error on Sentry came.

PR #15103, opened 2017-04-02
[BUGFIX beta] Allow calling `Ember.warn` without test.

### LINKED ISSUE #15096: Ember.warn throws deprecation if optional argument is not specified
Following docs `test` argument of `Ember.warn` is optional but if it's not specified deprecation _ember-debug.warn-options-missing_ is thrown.

Following docs this syntax should work: `Ember.warn('A warning to display', { id: 'ember-debug.feature-flag-with-features-stripped' });`

Serabe:
If `Ember.warn` is called without the `test` parameter, a deprecation message is triggered.

[RFC 65](https://github.com/emberjs/rfcs/blob/master/text/0065-deprecation-warning-handlers.md) makes the `test` parameter required for `Ember.deprecate` but not so for `Ember.warn`.

Fixes #15096

### DIFF
--- a/packages/ember-debug/lib/warn.js
+++ b/packages/ember-debug/lib/warn.js
@@ -70,6 +70,10 @@ export let missingOptionsIdDeprecation = 'When calling `Ember.warn` you must pro
   @since 1.0.0
 */
 export default function warn(message, test, options) {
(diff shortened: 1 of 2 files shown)


rwjblue:
Thanks!

PR #15104, opened 2017-04-03
Bump broccoli-rollup@1.2.0

chadhietala:
1.2.0 will do a noop if the file did not change.


stefanpenner:
@chadhietala can you double check that rollup related rebuilds still work correctly? I don't see why they wouldn't but better safe then sorry.

chadhietala:
Confirmed this with stef.

PR #15106, opened 2017-04-04
[Cleanup] Introduce a tiny bit more debugging data to warnings about CP dependencies

trek:
I ran across this updating an app from 2.4LTS to 2.5. 

We started with code a bit like this:

```
const foos = 'abc';
computed('model.@each', foos, function() {})
```

And mutated to:

```
const foos = ['abc', 'def'];
// !!! -> `foos` not properly tracking because it is now an array, not a string. 
computed('model.@each', foos, function() {})
```

Then we started the 2.4 -> 2.5 update in our app.

In 2.5, Ember started protecting from this error by checking `typeof` for dependencies and asserting if you passed other types. It was a little rough to track down with only "A computed property key must be a string" as a guide. 

This PR updates the assertion with a tiny bit more debugging data.

### DIFF
--- a/packages/ember-metal/lib/expand_properties.js
+++ b/packages/ember-metal/lib/expand_properties.js
@@ -35,7 +35,7 @@ var END_WITH_EACH_REGEX = /\.@each$/;
   expansion, and is passed the expansion.
 */
 export default function expandProperties(pattern, callback) {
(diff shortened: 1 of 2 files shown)


rwjblue:
Awesome, thank you @trek!

trek:
👋  long time no commit.

locks:
Such emeritus 😘

PR #15107, opened 2017-04-04
[BUGFIX beta] avoid toBoolean conversion when possible (chains)

stefanpenner:


stefanpenner:
cc @bmeurer / @krisselden

Issue #14843, opened 2017-01-16
arrayContentDidChange will often call objectAt(0) and objectAt(len - 1)


hjdivad:
When subclassing `MutableArray` it is surprising that any mutation can trigger `objectAt(0)` and `objectAt(len -1)`.  This happens because of the checks [here](https://github.com/emberjs/ember.js/blob/03b9dbc21e1eef88bc5a080dbb500547fe6d55f7/packages/ember-runtime/lib/mixins/array.js#L143-L156) for triggering change events for `firstObject` and `lastObject`.

This causes some difficulties when dematerializing records in Ember Data, because removing unloaded records from relationships, which is done asynchronously, would cause other records to be rematerialized because the materialization is lazy and the records are retrieved for the sake of this check.

This was reported in #5379 which was never fixed, but was closed in favour of a more limited problem in #5591 which *was* fixed in #14493.

One possible solution is to trigger changes for `firstObject` and `lastObject` only based on the indices of deletions and additions.  This will mean that change events will be triggered even in the case of `[a,a,b].removeAt(0);` which is not the case today.

cc @igorT

pixelhandler:
@hjdivad just curious do you know if anyone has a PR with a failing test on this issue?

hjdivad:
@pixelhandler #15110 

https://github.com/emberjs/ember.js/pull/15110/commits/8b851162583963832e6f6662415adbe9ee4341a2  updates a test to illustrate

https://github.com/emberjs/ember.js/pull/15110/commits/dabcbc29ca605490f74f220c470d1d339c71d7fb illustrates the suggested fix

mmun:
I like the proposed solution.

bekzod:
I think this is fixed in https://github.com/emberjs/ember.js/pull/15510 no ?

PR #15110, opened 2017-04-05
Update a test to fail #14843

hjdivad:
Update a test to fail #14843 and provide example code to illustrate the change suggested in that issue.

Before this can be merged we'd need to decide to deprecate the current semantics of `firstObject` and `lastObject`.  The trade-off is between having `objectAt` called in nearly every case for array mutations, or not detecting that `[A, B].insertAt(0, A)` does not change `firstObject`.

If that tradeoff is made, we should probably feature flag & deprecate the existing behaviour.

### DIFF
--- a/packages/ember-runtime/lib/mixins/array.js
+++ b/packages/ember-runtime/lib/mixins/array.js
@@ -142,17 +142,23 @@ export function arrayContentDidChange(array, startIdx, removeAmt, addAmt) {
   let cache = meta && meta.readableCache();
 
   if (cache) {
+    let length = get(array, 'length') + removeAmt - addAmt;
     if (cache.firstObject !== undefined &&
-        objectAt(array, 0) !== cacheFor.get(cache, 'firstObject')) {
+        ((startIdx === 0 && (addAmt > 0 || removeAmt > 0)) ||
+        (startIdx <= -length && (addAmt > 0 || removeAmt > 0)))) {
       propertyWillChange(array, 'firstObject');
       propertyDidChange(array, 'firstObject');
     }
+
     if (cache.lastObject !== undefined &&
-        objectAt(array, get(array, 'length') - 1) !== cacheFor.get(cache, 'lastObject')) {
+        ((startIdx >= length && addAmt > 0) ||
+        (startIdx >= 0 && (startIdx + removeAmt >= length)) ||
+        (startIdx < 0 && (removeAmt >= -startIdx)))) {
       propertyWillChange(array, 'lastObject');
       propertyDidChange(array, 'lastObject');
     }
   }
+
   return array;
 }
 
--- a/packages/ember-runtime/tests/suites/mutable_array/insertAt.js
+++ b/packages/ember-runtime/tests/suites/mutable_array/insertAt.js
@@ -135,10 +135,21 @@ suite.test('[A,B,C].insertAt(1,X) => [A,X,B,C] + notify', function() {
   let after  = [before[0], item, before[1], before[2]];
   let obj = this.newObject(before);
   let observer = this.newObserver(obj, '[]', '@each', 'length', 'firstObject', 'lastObject');
+  let objectAtCalls = [];
+
+  let objectAt = obj.objectAt;
+  obj.objectAt = (ix) => {
+    objectAtCalls.push(ix);
+    return objectAt.call(obj, ix);
+  }
 
   obj.getProperties('firstObject', 'lastObject'); /* Prime the cache */
 
+  objectAtCalls.splice(0, objectAtCalls.length);
+
   obj.insertAt(1, item);
+
+  deepEqual(objectAtCalls, [], 'objectAt is not called when only inserting items');
 
   deepEqual(this.toArray(obj), after, 'post item results');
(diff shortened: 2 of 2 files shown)


packages/ember-runtime/lib/mixins/array.js
```diff
@@ -142,17 +142,23 @@ export function arrayContentDidChange(array, startIdx, removeAmt, addAmt) {
   let cache = meta && meta.readableCache();
 
   if (cache) {
+    let length = get(array, 'length') + removeAmt - addAmt;
     if (cache.firstObject !== undefined &&
-        objectAt(array, 0) !== cacheFor.get(cache, 'firstObject')) {
+        ((startIdx === 0 && (addAmt > 0 || removeAmt > 0)) ||
```
stefanpenner: should we still include `objectAt(array, 0) !== cacheFor.get(cache, 'firstObject')` guard?

packages/ember-runtime/lib/mixins/array.js
```diff
@@ -142,17 +142,23 @@ export function arrayContentDidChange(array, startIdx, removeAmt, addAmt) {
...
+        (startIdx <= -length && (addAmt > 0 || removeAmt > 0)))) {
       propertyWillChange(array, 'firstObject');
       propertyDidChange(array, 'firstObject');
     }
+
     if (cache.lastObject !== undefined &&
-        objectAt(array, get(array, 'length') - 1) !== cacheFor.get(cache, 'lastObject')) {
+        ((startIdx >= length && addAmt > 0) ||
+        (startIdx >= 0 && (startIdx + removeAmt >= length)) ||
+        (startIdx < 0 && (removeAmt >= -startIdx)))) {
```
stefanpenner: should we still include `objectAt(array, get(array, 'length') - 1) !== cacheFor.get(cache, 'lastObject')) {` guard ?

packages/ember-runtime/lib/mixins/array.js
```diff
@@ -142,17 +142,23 @@ export function arrayContentDidChange(array, startIdx, removeAmt, addAmt) {
   let cache = meta && meta.readableCache();
 
   if (cache) {
+    let length = get(array, 'length') + removeAmt - addAmt;
```
stefanpenner: we should likely avoid this `get` if we are not needing it.

This is getting complicated, if we go down this path we should do a helper function that can be unit tested (and more readable).

packages/ember-runtime/tests/suites/mutable_array/insertAt.js
```diff
@@ -135,11 +135,22 @@ suite.test('[A,B,C].insertAt(1,X) => [A,X,B,C] + notify', function() {
   let after  = [before[0], item, before[1], before[2]];
   let obj = this.newObject(before);
   let observer = this.newObserver(obj, '[]', '@each', 'length', 'firstObject', 'lastObject');
+  let objectAtCalls = [];
+
+  let objectAt = obj.objectAt;
```
stefanpenner: we will require more test permutations, as the above code has many bounds

stefanpenner (changes requested):
Left some notes, if we decide to proceed.

Would love @krisselden's input

hjdivad:
Three possible courses of action:

1. do nothing
2. check the index ranges before calling `objectAt`
3. remove `objectAt` entirely and only check index ranges

stefanpenner:
@bekzod if you happen to have cycles. I think we would like to move forward with this PR, maybe you can lend @hjdivad a hand and help him land it?

bekzod:
@stefanpenner sure, will look at it

bekzod:
maybe following code is a bit clearer (tested with new test) ? @stefanpenner @hjdivad 
```
 if (cache !== undefined) {
    if (cache.firstObject !== undefined && startIdx === 0 && objectAt(array, 0) !== cacheFor.get(cache, 'firstObject')) {
      propertyWillChange(array, 'firstObject');
      propertyDidChange(array, 'firstObject');
    }

    if (cache.lastObject !== undefined) {
      let delta = (addAmt === -1 ? 0 : addAmt) - (removeAmt === -1 ? 0 : removeAmt);
      let length = get(array, 'length');

      let lastIndex = length - delta - 1;
      if ((lastIndex === -1 || delta === 0 || !(startIdx < lastIndex && lastIndex < length)) && objectAt(array, length - 1) !== cacheFor.get(cache, 'lastObject')) {
        propertyWillChange(array, 'lastObject');
        propertyDidChange(array, 'lastObject');
      }
   }
  }
```
also can `startIdx` be negative in `arrayContentDidChange`  ?

bekzod:
I think the approach suggested is good enough, cannot think of better way for now will improve this once merged if I get any better ideas

stefanpenner:
@bekzod nice, are you able to do the rebase + fix any failing tests? Feel free to open a new PR.

hjdivad:
Closed in favour of #15510

hjdivad:
> also can startIdx be negative in arrayContentDidChange ?

Yes, the semantics mirror those of [splice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice)

PR #15112, opened 2017-04-05
[BUGFIX release] Ensure Ember.Test globals setup when including ember-testing.js

trentmwillis:
Addressing one part of https://github.com/ember-cli/ember-cli/issues/6870.

After this, it looks like there are other errors that pop-up, but those will be handled in a separate PR.

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -6,6 +6,7 @@
 //
 // DISABLE_ES3=true DISABLE_JSCS=true DISABLE_JSHINT=true DISABLE_MIN=true DISABLE_DEREQUIRE=true ember serve --environment=production
 
+const stripIndent = require('common-tags').stripIndent;
 const MergeTrees = require('broccoli-merge-trees');
 const babelHelpers = require('./broccoli/babel-helpers');
 const bootstrapModule = require('./broccoli/bootstrap-modules');
@@ -275,8 +276,7 @@ module.exports = function(options) {
     license,
     loader,
     emberTestingAMD,
-    nodeModule,
-    bootstrapModule('ember-testing')
+    nodeModule
   ]);
 
   let emberTestsBundle = new MergeTrees([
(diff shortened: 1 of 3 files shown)


trentmwillis:
Looks like Safari bombed out on SauceLabs, due to a timeout.

rwjblue:
restarted

ember-cli-build.js
```diff
@@ -300,7 +300,15 @@ module.exports = function(options) {
   });
 
   emberTestingBundle = concat(emberTestingBundle, {
-    outputFile: 'ember-testing.js'
+    outputFile: 'ember-testing.js',
+    hasBootstrap: false,
+    footer: stripIndent`
+      var testing = requireModule('ember-testing');
+      Ember.Test = testing.Test;
```
rwjblue: Is `Ember` guaranteed to exist here? What happens when you eval `ember-testing.js` before `ember.debug.js` or `ember.prod.js`?

rwjblue (commented):
Can you share the source of `dist/ember-testing.js` with this PR?

trentmwillis:
Here's the output of [ember-testing.js](https://gist.github.com/trentmwillis/95afa3bcfda5f8d0657435346cd36444).

ember-cli-build.js
trentmwillis: Yes, due to [these lines](https://gist.github.com/trentmwillis/95afa3bcfda5f8d0657435346cd36444#file-ember-testing-js-L19-L23). I originally had `var Ember = requireModule('ember-metal/core');` but that was flagged as a redeclaration.

rwjblue: Confirmed that this works as intended via shared asset in https://github.com/emberjs/ember.js/pull/15112#issuecomment-292230312

rwjblue:
@trentmwillis - The entire build system has changed between release/beta and master, and df839ae cannot be cherry-picked.  Would you mind submitting a PR targetting beta with that commit?

trentmwillis:
@rwjblue, sure, target beta but still tag as `BUGFIX release`?

rwjblue:
Yes, beta and release are pretty close (the build side is basically the same), so it should still cherry-pick from beta to release.

trentmwillis:
So turns out this needs to be ported into emberjs-build and then we'll have to bump that package on the beta/release branch here.

PR #15115, opened 2017-04-06
Small refactor of tracking pending requests in acceptance tests

cibernox:
There is no need of putting requests in an array, we're only interested
in the amount of them.

I don't know how to label the PR since it's not a bugfix really.

/cc @rwjblue


cibernox:
Ignore this PR. It turns out it is necessary to keep track of the exact requests to avoid requests from the previous tests finishing in the current one incorrectly decrementing the current the current count. 
Perhaps a weakmap could do the same thing, but it doesn't worth it.

Issue #15113, opened 2017-04-06
TypeError: Cannot convert object to primitive value


shamcode:
There is the TypeError when you overriden computed property in component with array of `hash` elements. 
Exception:
```
TypeError: Cannot convert object to primitive value
    at Array.toString (native)
    at Object.toString (https://cdnjs.cloudflare.com/ajax/libs/ember.js/2.12.0/ember.debug.js:40911:18)
    at Object.set (https://cdnjs.cloudflare.com/ajax/libs/ember.js/2.12.0/ember.debug.js:22479:131)
    at ComputedProperty.computedPropertyClobberSet [as clobberSet] (https://cdnjs.cloudflare.com/ajax/libs/ember.js/2.12.0/ember.debug.js:17428:29)
    at ComputedProperty.computedPropertySetEntry [as set] (https://cdnjs.cloudflare.com/ajax/libs/ember.js/2.12.0/ember.debug.js:17411:19)
    at new Class (https://cdnjs.cloudflare.com/ajax/libs/ember.js/2.12.0/ember.debug.js:36565:22)
    at Function._ClassMixinProps.create (https://cdnjs.cloudflare.com/ajax/libs/ember.js/2.12.0/ember.debug.js:36775:12)
    at instantiate (https://cdnjs.cloudflare.com/ajax/libs/ember.js/2.12.0/ember.debug.js:1706:23)
    at DeprecatedFactoryManager.create (https://cdnjs.cloudflare.com/ajax/libs/ember.js/2.12.0/ember.debug.js:1839:14)
    at Proxy.create (https://cdnjs.cloudflare.com/ajax/libs/ember.js/2.12.0/ember.debug.js:1405:22)
```
[Twiddle with error](https://ember-twiddle.com/e18d6f7636b3815e762159ae82d9701c)
[Twiddle without error (replace hash helper to custom)](https://ember-twiddle.com/ea790324b004df88e185ec822cb14d85)
[Twiddle without error (replace computed to plain object)](https://ember-twiddle.com/448ca6401cbd56dc5c5fbabc5f8ade3f)

Exception in dev-tools:
![ember-type-error](https://cloud.githubusercontent.com/assets/2228977/24756038/93bf58d6-1b06-11e7-880c-e78fe04a4ee7.png)

Serabe:
Closing as duplicate. Thank you!

PR #15117, opened 2017-04-07

### LINKED ISSUE #14738: Cannot convert object to primitive value
This is similar to https://github.com/emberjs/ember.js/issues/14187, but I have a reproduction in Ember 2.10 with no addons:

https://ember-twiddle.com/4df2d62bf5f1c05bd74086dc79326020?openFiles=components.my-component.js%2C

The critical ingredient seems to be that I am taking an object created by the `hash` helper, putting it inside an `Array`, and passing the array into another component. When such an array hits `toString` it causes an exception.

(It's hitting `toString` in the first place to support an assertion.)

Serabe:
`ember-utils/to-string` is documented as a `toString` function that supports
objects without a `toString` method.

Arrays do have a string method that is based on their `join` method. This
method spec ([Array.prototype.join in ecma262 2017
draft](https://tc39.github.io/ecma262/#sec-array.prototype.join)) is based on
calling `toString` on the objects that form the collection. If an object were
not to have such method, `Array.prototype.toString` would throw and error.

This PR recreates `Array.prototype.join` as closer to the spec as I could,
changing the `ToString` part with the safe version of `toString`.

Fixes #14738 and #15113.

### DIFF
--- a/packages/ember-utils/lib/to-string.js
+++ b/packages/ember-utils/lib/to-string.js
@@ -1,11 +1,36 @@
 const objectToString = Object.prototype.toString;
 
+function isNone(obj) {
+  return obj === null || obj === undefined;
+}
+
 /*
  A `toString` util function that supports objects without a `toString`
  method, e.g. an object created with `Object.create(null)`.
 */
 export default function toString(obj) {
-  if (obj && typeof obj.toString === 'function') {
+  let type = typeof obj;
+  if (type === "string") { return obj; }
+
+  if (Array.isArray(obj)) {
+    // Reimplement Array.prototype.join according to spec (22.1.3.13)
+    // Changing ToString(element) with this safe version of ToString.
+    let len = obj.length;
+    let sep = ',';
+    let r = '';
+
+    for (let k = 0; k < len; k++) {
+      if (k > 0) {
+        r += ',';
+      }
+
+      if (!isNone(obj[k])) {
+        r += toString(obj[k]);
+      }
+    }
+
+    return r;
+  } else if (obj != null && typeof obj.toString === 'function') {
     return obj.toString();
   } else {
     return objectToString.call(obj);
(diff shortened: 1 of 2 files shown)


rwjblue (approved):
This is _slightly_ annoying to have to ship, but I think it is a good solution to a fairly gnarly problem.  Another solution would be to have the `(hash` helper return a "regular" object with a `.toString` method (which of course possibly pollutes).

Seems good to me...

packages/ember-utils/lib/to-string.js
```diff
@@ -1,11 +1,33 @@
...
+        r += ',';
+      }
+
+      if (!isNone(obj[k])) {
+        r += toString(obj[k]);
+      }
+    }
+
+    return r;
+  } else if (obj && typeof obj.toString === 'function') {
```
stefanpenner: can `obj` be some `truthy` non object?

Serabe: Would `!isNone(obj) && typeof obj.toString === 'function'` be ok?

stefanpenner: `typeof obj === 'object' && !== obj && typeof obj.toString === 'function'` should do the trick

Serabe: That would not work for strings.

stefanpenner: yes, strings should be a short circuit at the root of the function.

```js
if (typeof obj === 'string') { return obj; }
```

invoking `string.toString()` seems funky?

stefanpenner:
```js
let type = obj;

if (type === 'string') { return obj; }
if (Array.isArray(obj)) { /* do new stuff */}
else if (obj !== null && typeof obj.toString === 'function') {
   return obj.toString();
} else {
   return objectToString.call(obj); 
}
```

Serabe:
Restarting sauce.

Serabe:
Tests are passing now

notmessenger:
If you are looking for this fix it wasn't released until `2.13.0` on 27 Apr 2017 (https://github.com/emberjs/ember.js/blob/v2.13.0/packages/ember-utils/lib/to-string.js)

**HOWEVER...**

It **was** merged into the `#lts-2-12` branch on 27 Apr 2017 (https://github.com/emberjs/ember.js/blob/lts-2-12/packages/ember-utils/lib/to-string.js)

PR #15119, opened 2017-04-07
[DEPRECATION] Remove ember-metal/map

### LINKED ISSUE #15041: See about cleaning up OrderedSet
Source: https://github.com/emberjs/ember.js/blob/v2.11.0/packages/ember-metal/lib/map.js#L55

**Edited by @serabe with @locks consent**: See https://github.com/emberjs/ember.js/issues/15041#issuecomment-315065566 for latest information as of 2017-07-13.

Serabe:
`ember-metal/map` contains three private classes `OrderedSet`, `Map`,
`MapWithDefault`. They were not used anywhere in the codebase except for tests
or reexports, so they are being deprecated.

Ember Data team has been removing their usages.

Added emberjs/website#2884.

Fixes #15041
Fixes https://github.com/emberjs/ember.js/issues/13815

### DIFF
--- a/packages/ember-metal/lib/map.js
+++ b/packages/ember-metal/lib/map.js
@@ -20,6 +20,7 @@
   Map is mocked out to look like an Ember object, so you can do
   `Ember.Map.create()` for symmetry with other Ember classes.
 */
+import { deprecate } from 'ember-debug';
 import { guidFor } from 'ember-utils';
 
 function missingFunction(fn) {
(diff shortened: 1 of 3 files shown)


rwjblue (commented):
Seems like there is still some usage of this in the wild...

* https://emberobserver.com/code-search?codeQuery=Ember.OrderedSet
* https://emberobserver.com/code-search?codeQuery=Ember.Map
* https://emberobserver.com/code-search?codeQuery=Ember.MapWithDefault

Seems like we should confirm that these usages can be migrated to something else (the ember-data usage seems to be the highest priority), then add a deprecation for using these with an `until: 2.16`...

Serabe:
I'll change this to deprecations this weekend :)

Serabe:
Updated PR. Not sure I approached the deprecation the best way and I for sure have no idea on which the exact URL would be.

Feedback more than welcomed!

stefanpenner:
> Ember Data team has been removing their usages.

This isn't true, although some scenarios in ED incorrectly used a set, many legitimately require it.

> so they are being deprecated.

I'm alright with deprecating, but there needs to be a transition plan in place before that. As ember-data does require an efficient OrderedSet. Maybe via add-on, or as part of ember-data itself.

Serabe:
Rebased.

@stefanpenner we talked about this in #dev-ember and seemed so. In any case, seems more natural to provide an `OrderedSet` from an add-on or via EmberData itself, given it is not being used in any place in this repo. Happy to move the code and tests to a new repo as part of this PR.

Serabe:
Closing this.

PR #15125, opened 2017-04-10
Extract jquery into an ember addon

topaxi:
This also allows applications to switch to the `jquery.slim` build released with [jQuery 3](https://blog.jquery.com/2016/06/09/jquery-3-0-final-released/) which excludes ajax, effects, and currently deprecated code.

Example `ember-cli-build`:
```javascript
module.exports = function(defaults) {
  let app = new EmberAddon(defaults, {
    jquery: {
      slim: true
    }
  });
};
```


rwjblue:
Thanks for working on this!  I am not 100% sure on the path forward here...

topaxi:
Me neither :) Let me know if and how I can make some progress on this. I assume I'd need to patch / remove some things in ember-cli too?

There are some references to jQuery here and onwards: https://github.com/ember-cli/ember-cli/blob/f0ef350681f96e03e17d4352f20a63f114f99cf1/lib/broccoli/ember-app.js#L308

I guess this should all stay backwards compatible, and I'm not 100% sure what and how to do this as I don't have any knowledge of ember-cli internals (yet?) :)

stefanpenner:
This looks awesome.

cc @cibernox this may interest you.

cibernox:
> cc @cibernox this may interest you.

It does!

@topaxi I've been working myself towards removing jQuery, so this is nice. There is a few things about which we need to know what the core thinks.

1. Can this approach of moving jquery to an addon be acceptable from a semver perspective. I don't know wether or not extracting functionality out of the core into addons that then go into the default blueprint can be considered breaking. Take into account that without this addon, things will break.

2. The opposite approach could be easier, have an `ember-no-jquery` addon that does the opposite. 

Things that an app without jQuery must do include:
- Replace the event dispatcher with `ember-native-dom-event-dispatcher`
- Ensure apps don't use the built in test helpers but some jQuery free alternative (`ember-native-dom-helpers`)

If we want to follow approach 1, we would have to extract the jQuery event dispatcher, put the `ember-native-dom-event-dispatcher` inside ember instead, and make the `ember-jquery` addon use it. It is unclear to me how to deal with the fact that test helpers rely so much in jQuery.

locks:
> Take into account that without this addon, things will break.
That's why it's included by default :P

cibernox:
@locks Sure, I wanted to be sure that moving functionality out of Ember into an addon is OK from a SemVer perspective, even if that addon is installed by default.

rwjblue:
I think that we need to make ember-cli aware of this actually.  Basically the path forward here is for the ember-source addon itself to manage the `app.import`'s that are done for the various things in `vendorFiles` in ember-cli ([around here](https://github.com/ember-cli/ember-cli/blob/115e22e0e747b7dc618a9add297385e566ee15c6/lib/broccoli/ember-app.js#L307)). In the end `vendorFiles` should be used _only_ for supporting bower versions of Ember.

cibernox:
@rwjblue If we follow that path that means that we can't use it when installing ember from bower, and right now it's something we have to do to test in beta, canary and LTS. Shouldn't we be able to install those versions from npm first?

rwjblue:
I think we would create a ember-source-legacy addon or something that would be used when using with bower. This also allows us to have that addon being ember-cli-legacy-blueprints (removing yet another legacy dinosaur from ember-cli).

Serabe:
In first place, so sorry for letting this get stale. Our shame.

Sadly, this is being closed and not merged. While removing jQuery will happen there are a few steps that need to be taken before.

Thank you and sorry!

topaxi:
I'm aware that this need more work, if you can point me in directions how and where I can help, I'd gladly look into some issues :)

Serabe:
The approach being taken is different than this one, there are some edges to address before we can finally drop jQuery.

Thank you!

PR #15127, opened 2017-04-10
Fix misspellings

olleolleolle:
This PR fixes a few misspellings.

I used the [misspellings](https://pypi.python.org/pypi/misspellings) tool to locate these:

```
$ find . | misspellings -f -
```

### DIFF
--- a/packages/ember/tests/routing/basic_test.js
+++ b/packages/ember/tests/routing/basic_test.js
@@ -3697,7 +3697,7 @@ QUnit.test('Doesnt swallow exception thrown from willTransition', function() {
 
   throws(() => {
     run(() => router.handleURL('/other'));
-  }, /boom/, 'expected an exception that didn\'t happen');
+  }, /boom/, 'expected an exception but none was thrown');
 });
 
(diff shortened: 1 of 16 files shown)


packages/ember/tests/routing/basic_test.js
```diff
@@ -3697,7 +3697,7 @@ QUnit.test('Doesnt swallow exception thrown from willTransition', function() {
 
   throws(() => {
     run(() => router.handleURL('/other'));
-  }, /boom/, 'expected an exception that didnt happen');
+  }, /boom/, 'expected an exception that didn\'t happen');
```
locks: I _think_  we might want to use backticks here, need confirmation.

locks (changes requested):
Great work! I have a nitpick to confirm.

packages/ember/tests/routing/basic_test.js
olleolleolle: Rewording:

```
expected an exception but none was thrown
```

locks (approved):

locks:
Thank you!

PR #15129, opened 2017-04-11
Fix access to service:-document in ember-engines

dnachev:
Run into some problems trying to get some code to render server-side. I think the problem was related to `ember-wormhole`, although the addon currently doesn't have a good way to work in server-side at all, but it still required a reference to the document service to be able to initialize.

Although this can be fixed in ember-wormhole, I think it's good idea to allow components in engines to get access to the document if they need to and cover future problems.

Not sure whether enhancing the application instance test is enough, but not sure where to put a test, which says components in `ember-engines` must have access to the `service:-document` in applications, which uses `ember-cli-fastboot`.

### DIFF
--- a/packages/ember-application/lib/system/engine-instance.js
+++ b/packages/ember-application/lib/system/engine-instance.js
@@ -188,7 +188,8 @@ const EngineInstance = EmberObject.extend(RegistryProxyMixin, ContainerProxyMixi
       'router:main',
       P`-bucket-cache:main`,
       '-view-registry:main',
-      `renderer:-${env.isInteractive ? 'dom' : 'inert'}`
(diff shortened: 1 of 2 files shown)


dnachev:
Actually, on second thought, it may be worth to have a test in ember-engines to validate that a component has access to the document service in an engine instance. I don't think Fastboot plays a role in this fix. @trentmwillis what do you think?

trentmwillis:
@rwjblue does this seem like a reasonable fix? I'm not entirely sure what all the `-document` service is used for. Should Engines inherit it from the host or create their own?

rwjblue:
Yes, this is the correct fix.  Sadly, I missed this PR when originally submitted 😢.

PR #15134, opened 2017-04-12
remove old deprecation flag in Map/OrderedSet

stefanpenner:


rwjblue:
Restarted travis build

PR #15136, opened 2017-04-12
Fix documentation

marpo60:
This is the Router used in the documentation:
```
     Router.map(function() {
       this.route('post', { path: '/posts/:post_id' }, function() {
         this.route('comments');
       });
     });
```


locks:
Thank you!
@homu r+

locks:
Welp, homu seems to be on vacation.

PR #15141, opened 2017-04-14
[PERF/Correctness] Remove global didDefineProperty

stefanpenner:
* removes an extra mixin from all Object Hierarchies
* only runs when actually defining a CP if actually defining a new computed property.
* misc cleanup

### DIFF
--- a/packages/ember-metal/lib/index.js
+++ b/packages/ember-metal/lib/index.js
@@ -73,7 +73,8 @@ export {
 } from './property_events';
 export {
   defineProperty,
-  Descriptor
+  Descriptor,
+  _hasCachedComputedProperties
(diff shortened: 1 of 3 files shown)


stefanpenner:
test failure was just yarn, giving it another kick

```
 The command "yarn install
  " failed and exited with 1 during .
```

stefanpenner:
This may also be a `[BUGFIX beta]` as the behavior is the same, just without wasteful things.

rwjblue:
LGTM

@krisselden - r?

rwjblue (approved):

krisselden:
When things are labelled perf we should try to include some reference to benchmark for reference

PR #15142, opened 2017-04-14
[BUGFIX release] Don’t leak container while injection deprecated cont…

stefanpenner:
…ainer

### DIFF
--- a/packages/container/lib/container.js
+++ b/packages/container/lib/container.js
@@ -527,25 +527,28 @@ function factoryInjectionsFor(container, fullName) {
   return factoryInjections;
 }
 
+const INJECTED_DEPRECATED_CONTAINER_DESC = {
+  configurable: true,
+  enumerable: false,
+  get() {
+    deprecate('Using the injected `container` is deprecated. Please use the `getOwner` helper instead to access the owner of this object.', false, { id: 'ember-application.injected-container', until: '3.0.0', url: 'http://emberjs.com/deprecations/v2.x#toc_injected-container-access' });
+    return this[CONTAINER_OVERRIDE];
+  },
+
+  set(value) {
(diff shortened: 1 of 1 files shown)


rwjblue:
The approach looks good, but there is some sort of failure still...

packages/container/lib/container.js
```diff
@@ -527,25 +517,27 @@ function factoryInjectionsFor(container, fullName) {
...
-    }
-  });
+    return value;
+  }
+};
+
+// TODO - remove when Ember reaches v3.0.0
+function injectDeprecatedContainer(object) {
+  if ('container' in object) { return; }
+  Object.defineProperty(object, 'container', INJECTED_DEPRECATED_CONTAINER_DESC);
```
stefanpenner: I need to actually set the container again here `setOwner(object, container)`

stefanpenner: looks like `container.owner` is null for the offending tests. and this PR closing over `container` masked that.

@rwjblue do i expect `container.owner` to be null? Is that a test config error?

stefanpenner: I'll actually investigate more, just gotta head out now.

stefanpenner:
@rwjblue / @krisselden fixed, ready for review

rwjblue (approved):

krisselden (commented):
Why are we setting override instead of setOwner? Also, the owner is supposed to be the app instance and likely that test didn't have an app instance.  Object in the above is the prototype with no double extend so this would still leak

stefanpenner:
Because the original code was explicitly returning the container (Not whatever the owner happened to be) I didn't want to change behavior for a leak fix.

I think as a next step we should likely just remove this code entirely in canary, thoughts?

stefanpenner:
Or a better question, what would need to be done to remove this in canary? It may just be update the deprecation message in release/beta

PR #15143, opened 2017-04-14
Refactor guidFor to use WeakMap if present.

rwjblue:
- Refactor `Ember.guidFor` to always return a number (avoid extraneous string allocations, easier to compare, etc)
- Move guid storage from a property on the object (using `obj.__defineNonEmumerable` or `Object.defineProperty`) into a `WeakMap` when native WeakMap's are available, where not supported fall back to setting `GUID_KEY`.

TODO:

- [x] Update Backburner to allow passing a `getGuid` function (instead of only a `GUID_KEY`) (https://github.com/BackburnerJS/backburner.js/pull/211)

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "babel-plugin-transform-es2015-template-literals": "^6.22.0",
     "babel-plugin-transform-proto-to-assign": "^6.23.0",
     "babel-template": "^6.23.0",
-    "backburner.js": "^0.4.0",
+    "backburner.js": "github:rwjblue/backburner.js#peek-guid-check-in-dist",
     "broccoli-babel-transpiler": "next",
     "broccoli-concat": "^3.2.2",
     "broccoli-file-creator": "^1.1.1",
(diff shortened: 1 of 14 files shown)


runspired:
Cc @stefanpennet for OrderedSet compat

runspired:
Ugh GitHub mobile sux0rs

runspired:
@stefanpenner

rwjblue:
I ran these changes against ember-macro-benchmark (each commit individually). The results are a bit underwhelming, so I'd love some general review of the code changes to see if I've made a silly mistake somewhere.

samples

[results-refactor-guid.zip](https://github.com/emberjs/ember.js/files/923810/results-refactor-guid.zip)

stefanpenner:
@rwjblue  I would expect to see a performance improvement here, potentially a regression for common cases.

stefanpenner:
@runspired the new ordered set doesn't fall back to guidKey at all until the set grows > 50n (once it lands)... Which should help mitigate much of the overhead.

@rwjblue I would love to see the difference for InternalModel allocations (weakMap id vs guidKey property). Currently they pre-define `GUID_KEY` so that they can be quickly added/removed from todays ordered sets.

krisselden:
@rwjblue I need to add more apps, and the above might be different with Stefan's Set refactor

When we did this for meta, I think there was a better balance between the WeakMap perf and some of the shaping issues with meta, not sure, or maybe it was just the app we tried.

mmun:
Merged in #16289 (except for making the guide integers).

PR #15144, opened 2017-04-15
Deprecate ApplicationInstance.didCreateRootView

### LINKED ISSUE #14136: Deprecate `didCreateRootView`
This was added for fastboot, but is no longer used. It also leaks the "top-level view" which is not really a thing anymore.

AKST:
# Fix #14136 

## Meta / Questions

Disclaimer I'm a first time contributor 😅 So I have a few remaining questions regarding my PR

- Should I include the yarn.lock file? I didn't modify any dependencies but there appears to be some changes...
- Is there a url this PR deprecation should include?
- Is there an alternative API this message should suggest to use in this features place
- Is the chosen id for the deprecation appropriate?
- Is this PR pointed at the right branch? The Contributing.md mentions some stuff but not sure where adding deprecations fits in

## Changes

- Wrote a test that checks that the deprecation warning has been fired, however I'm personally not sure if this is the type of thing tests should check or not

- Added the deprecation warning in the designated location

- Refactor the remaining usage (a single function call, in the routes) by inlining the exact
   functionality of the method (which is a single line) to where it was being called.
  - I think this is okay because there is no accessing of private fields within the original method.
  - In the same breath I do feel this is the most sketchiest piece of this PR, so i'm seeking input on that part specifically

### DIFF
--- a/packages/ember-application/lib/system/application-instance.js
+++ b/packages/ember-application/lib/system/application-instance.js
@@ -153,7 +153,7 @@ const ApplicationInstance = EngineInstance.extend({
     deprecate(
       'Using `ApplicationInstance.didCreateRootView` is deprecated.',
       false, {
-        id: 'ember-application.app-instance-didCreateRootView',
+        id: 'ember-application.app-instance.did-create-root-view',
         until: '3.0.0',
       }
     );
--- a/packages/ember-application/lib/system/application-instance.js
+++ b/packages/ember-application/lib/system/application-instance.js
@@ -154,7 +154,7 @@ const ApplicationInstance = EngineInstance.extend({
       'Using `ApplicationInstance.didCreateRootView` is deprecated.',
       false, {
         id: 'ember-application.app-instance.did-create-root-view',
-        until: '3.0.0',
+        until: '2.16.0',
       }
     );
     view.appendTo(this.rootElement);
--- a/packages/ember-application/lib/system/application-instance.js
+++ b/packages/ember-application/lib/system/application-instance.js
@@ -150,6 +150,13 @@ const ApplicationInstance = EngineInstance.extend({
     @private
   */
   didCreateRootView(view) {
+    deprecate(
+      'Using `ApplicationInstance.didCreateRootView` is deprecated.',
+      false, {
+        id: 'ember-application.app-instance-didCreateRootView',
+        until: '3.0.0',
+      }
+    );
     view.appendTo(this.rootElement);
   },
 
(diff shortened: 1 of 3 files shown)


packages/ember-application/lib/system/application-instance.js
```diff
@@ -150,6 +150,13 @@ const ApplicationInstance = EngineInstance.extend({
     @private
   */
   didCreateRootView(view) {
+    deprecate(
+      'Using `ApplicationInstance.didCreateRootView` is deprecated.',
+      false, {
+        id: 'ember-application.app-instance.did-create-root-view',
+        until: '3.0.0',
```
rwjblue: Let's use 2.16.0 here since it was never public API

rwjblue (commented):
> Should I include the yarn.lock file? I didn't modify any dependencies but there appears to be some changes...

Feel free to submit another PR with only the `yarn.lock`, but I wouldn't include it in this PR.

> Is there a url this PR deprecation should include?

That will need a PR to the emberjs/website repo to add the deprecation message. 

> Is there an alternative API this message should suggest to use in this features place

Generally there would be, but I believe the point here is that we don't think there is a replacement. This was only added and used by FastBoot, so I believe the answer mostly is: "refactor to not need it".

> Is the chosen id for the deprecation appropriate?

Yes, seems good. 

> Is this PR pointed at the right branch? The Contributing.md mentions some stuff but not sure where adding deprecations fits in

Yes, targeting master branch is perfect. 

---

Overall, I think this is pretty solid, and is likely good to land on master. 

@kratiahuja / @tomdale - Can y'all confirm that the FastBoot usage is removed?

kratiahuja:
@rwjblue ember-cli-fastboot uses this currently to clear double boot. What is the alternative if this is getting deprecated?

packages/ember-application/lib/system/application-instance.js
AKST: Okay sweet, too easy! I'll change that tonight (Australian time)

AKST:
@rwjblue I've change the version, but does @kratiahuja 's comments about fast boot still using this void the necessity of this PR / issue in general? If so feel free to close this issue 😅 I'll see if I can find another deprecation issue as it was quite simple.

I'll probably leave the yarn lock file out of the PR then, I think it might be due to my version of node (I'm running on a unstable release), or something weird about system.

AKST:
This has been open for more than a year, and I have a feeling this has either been address or these changes are no longer applicable.

Not really sure, but I'm going to close this PR I made as I'm just trying tidy up my own outstanding PRs.

PR #12202, opened 2015-08-25
[BUGFIX release] Fix bug in utils.isArray wherein an object with a length property would be considered an array.

schreiaj:
Fixes issue #12094

PR #15146, opened 2017-04-16
[BUGFIX beta] Improve isArray detection

### LINKED ISSUE #12094: isArray returns true for objects with property of length
This includes models that contain a property called `length`. 

Simple fix involves changing https://github.com/emberjs/ember.js/blob/master/packages/ember-runtime/lib/utils.js#L51 

to be 

``` javascript
if ((!isNaN(obj.length) && 'object' === type) { return true; }
```

Serabe:
An object containing an arbitrary `length` property would be detected
as an array, even if said property was not a number.

Fixes #12094 and supersedes #12202

### DIFF
--- a/packages/ember-runtime/lib/utils.js
+++ b/packages/ember-runtime/lib/utils.js
@@ -49,7 +49,8 @@ export function isArray(obj) {
 
   let type = typeOf(obj);
   if ('array' === type) { return true; }
-  if ((obj.length !== undefined) && 'object' === type) { return true; }
+  let length = obj.length;
+  if (typeof length === 'number' && length === length && 'object' === type) { return true; }
   return false;
 }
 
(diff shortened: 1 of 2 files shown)


Serabe:
Trying to ease the pain of merging old PRs. I don't like this way cause original authorship is lost. Any other option?

packages/ember-runtime/lib/utils.js
```diff
@@ -49,7 +49,7 @@ export function isArray(obj) {
 
   let type = typeOf(obj);
   if ('array' === type) { return true; }
-  if ((obj.length !== undefined) && 'object' === type) { return true; }
+  if (!isNaN(obj.length) && 'object' === type) { return true; }
```
stefanpenner: does our browser support matrix support `isNan`? http://caniuse.com/#search=isNaN

stefanpenner:
> Trying to ease the pain of merging old PRs. I don't like this way cause original authorship is lost. Any other option?

you can cherry-pick there work, and amend it with the needed changes. They will be the author, you will be the commiter.

rwjblue:
You can also generally commit to their branches (after amending their commits) to avoid creating new PR's (this is a newish github feature)...

Serabe:
Cherry-picked! @rwjblue did not know about that feature. Will try in a more controlled environment before doing so in here though.

Thank you!

PR #15147, opened 2017-04-16
[BUGFIX beta] finishChains should operate on metas

stefanpenner:
Previously, finishChains (run in every EmberObject creation) would have to lookup its own meta, even though the caller already has access to this meta.

* also fix an invalid test (I introduced the test bug here: https://github.com/emberjs/ember.js/commit/e46e056f1d5ed3136efbdf2136febebacd0461c6#commitcomment-21779571)

### DIFF
--- a/packages/ember-metal/lib/chains.js
+++ b/packages/ember-metal/lib/chains.js
@@ -347,22 +347,16 @@ function lazyGet(obj, key) {
 
 import { makeChainNode } from './watch_path';
 
-export function finishChains(obj) {
-  // We only create meta if we really have to
-  let m = peekMeta(obj);
-  if (m !== undefined) {
-    m = metaFor(obj);
-
-    // finish any current chains node watchers that reference obj
-    let chainWatchers = m.readableChainWatchers();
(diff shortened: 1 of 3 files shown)


packages/ember-metal/lib/chains.js
```diff
@@ -347,22 +347,16 @@ function lazyGet(obj, key) {
 
 import { makeChainNode } from './watch_path';
 
-export function finishChains(obj) {
-  // We only create meta if we really have to
-  let m = peekMeta(obj);
```
stefanpenner: uneeded

packages/ember-metal/lib/chains.js
```diff
@@ -347,22 +347,16 @@ function lazyGet(obj, key) {
...
-  if (m !== undefined) {
-    m = metaFor(obj);
```
stefanpenner: uneeded

stefanpenner:
test failure appears to be random sauce labs failure, restarting that task.

rwjblue (approved):

Issue #11843, opened 2015-07-21
Route queryParams replace does not work if refreshModel is true 


novtor:
For a route with queryParams, replace does not work if refreshModel is set to true:
http://jsbin.com/jebuko/edit?html,js,output
Open jsbin as live preview, enter something in input field, submit the input, it will modify query parameter 'a' and refresh the template. Do it several times. Then 'back' button will pass trough all the submit history.
Is it expected or I do something wrong?

pixelhandler:
@novtor can you take a look at #10945 to see if this is a duplicate issue?

stefanpenner:
will reopen if it is demonstrated to not be a duplicate

novtor:
No it is not duplicate. I opened another one some time ago: #10972 which is a duplicate, but this one is independent.
In fact, #10945 and #10972 say that the transition retry does not work with refreshModel query params and dynamic segments : there is an exception thrown, stacktrace is present in console, there is a problem with the page visualisation.
The issue I describe here is that independently of dynamic segments, if you put in a route:
```javascript
App.ArtistSongsRoute = Ember.Route.extend({
queryParams: [
  qp: {
     refreshModel: true,
     replace: true
  }
]
....
```
This should result in, AFAIU, that if the model is refreshed several times with different 'qp', the browser history will only hold one reference on this route url. So, let's say the user searches for an artist:
artist/someartist
and then looks for artist songs:
artist/someartist/songs?start=1980&end=1981
and then 
artist/someartist/songs?start=1981&end=1982
and then, let's say
artist/someartist/songs?style=rock
this would only create one entry in the browser history for artist's songs and if the user clicks 'back' browser button, she should be routed to the "artist/someartist" route.

http://emberjs.jsbin.com/zawiwe

enter '1' in the input and submit, then '2' and submit, then '3' and submit. Now if you click 'back' browser button you will work through &qp=2 and &qp=1 history

novtor:
@stefanpenner, is my explanation clear? Do you think it is not duplicate issue?

richmolj:
I'm experiencing this on Ember 2.2 as well. `replace` is ignored if `refreshModel` is true.

@novtor what do you mean by "send with replace query params"? Looking for a workaround.

novtor:
hi @richmolj , see the jsbin example

Panman8201:
This is a duplicate of #10577

pixelhandler:
[...]

PR #15148, opened 2017-04-17
Add failing test for query params replace and refreshModel not working together

alexspeller:
Add failing test for query params not working with both replace and refreshModel

Ember issue: #11843
Resolved by: tildeio/router.js#210

### DIFF
--- a/packages/ember/tests/routing/query_params_test.js
+++ b/packages/ember/tests/routing/query_params_test.js
@@ -567,6 +567,51 @@ moduleFor('Query Params - main', class extends QueryParamTestCase {
     });
   }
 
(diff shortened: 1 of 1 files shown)


alexspeller:
Note - the tests are supposed to fail at the moment, this requires a new release of router.js and then bumping the version in package.json for the test to pass (checked locally that this resolves this issue, which it does and doesn't cause any other test failures)

rwjblue:
@alexspeller - Mind rebasing, so we can land this?  I think the recent bump to router_js@1.2.7 should unblock...

rwjblue:
Thanks again @alexspeller!

PR #15153, opened 2017-04-19
Experiment with removing string / number cache for guidFor

stefanpenner:
now:

```js
guidFor(1) // => '1'
guidFor('string') // => 'string'
```

before:

```js
guidFor(1) => 'nu1';
guidFor('string') => 'st-123123'
```

I think the only real API constraint is that `guidFor` returns a string, and that `guidFor(x) === guidFor(x)` which this still upholds.

### DIFF
--- a/packages/ember-utils/lib/guid.js
+++ b/packages/ember-utils/lib/guid.js
@@ -31,10 +31,6 @@ export function uuid() {
  */
 const GUID_PREFIX = 'ember';
 
-// Used for guid generation...
-const numberCache  = [];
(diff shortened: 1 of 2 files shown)


stefanpenner:
cc @runspired

krisselden:
It violates x !== y then guidFor(x) !== guidFor(y)

runspired:
@krisselden I believe you are saying this would be a violation (using the alterations):

```
guidFor('1') !== guidFor(1)
```

PR #15161, opened 2017-04-24
resolves #15154

### LINKED ISSUE #15154: Internet Explorer (v11): Object doesn't support property or method 'includes'
I was about to deploy my ember app to our testing environment where I am able to test in Internet Explorer and this "browser" comes up with an error. I tracked it down to 

ember.js/packages/ember-metal/lib/error_handler.js

```
// To maintain stacktrace consistency across browsers
let getStack = error => {
  let stack = error.stack;
  let message = error.message;

  if (stack && !stack.includes(message)) {
    stack = `${message}\n${stack}`;
  }

  return stack;
};
```
But I don't have no clue how to fix it - I don't know If this is something caused by my code... In the IE debugger the variable 'error' is undefined... In every other browser there is no issue... :-/

mk-conn:
Fixes #15154.
Implemented small fix as suggested by @rwjblue.

### DIFF
--- a/packages/ember-metal/lib/error_handler.js
+++ b/packages/ember-metal/lib/error_handler.js
@@ -6,7 +6,7 @@ let getStack = error => {
   let stack = error.stack;
   let message = error.message;
 
-  if (stack && !stack.includes(message)) {
+  if (stack && stack.indexOf(message) === -1) {
     stack = `${message}\n${stack}`;
   }
 


locks:
Hi @mk-conn, could you change the commit message to be a bit more descriptive? Something like `[BUGFIX release] Fixes  #15154. Replaces `.includes` with `.indexOf`.

stefanpenner:
> Hi @mk-conn, could you change the commit message to be a bit more descriptive? Something like "[BUGFIX release] Fixes #15154. Replaces .includes with .indexOf".

This would merely duplicate the `diff` in english in the commit message. The commit message should explain `why` not `what`. The `diff` tell us `what`.

locks:
@mk-conn sorry for tweaking your commit like that, usually we'd wait for your action but 2.13 is coming out soon and I wanted to be sure this would get in!

mk-conn:
No prob @locks - I was laying in bed sick yesterday and did not really want to hang at the computer so I could not respond...

PR #15163, opened 2017-04-24
[BUGFIX beta] Avoid storing container on the prototype.

rwjblue:
In order to support `this.container` (deprecated since 2.3) being available to objects for backwards compatibility, we define the `container` property on all classes that we instantiate via normal DI APIs.

The bulk of the work to support this was done in prior refactorings. This change is targetted at ensuring that we do not leak the container instance onto the object's prototype in the `CONTAINER_OVERRIDE` slot.

After these changes, we still support all of the APIs that we supported previously, but will fallback to the current objects owner `__container__` property (instead of storing it on the prototype directly).

Also, the test being updated here was not providing the `owner` to the container in the unit test. This test being changed was previously only doing part of the required setup for creating a `registry` / `container` / `owner` group.

### DIFF
--- a/packages/container/tests/container_test.js
+++ b/packages/container/tests/container_test.js
@@ -554,8 +554,9 @@ QUnit.test('An object with its owner pre-set should be returned from ownerInject
 });
 
 QUnit.test('A deprecated `container` property is appended to every object instantiated from an extendable factory', function() {
+  let owner = { };
   let registry = new Registry();
-  let container = registry.container();
+  let container = owner.__container__ = registry.container({ owner });
   let PostController = factory();
   registry.register('controller:post', PostController);
   let postController = container.lookup('controller:post');
@@ -741,4 +742,3 @@ QUnit.test('#factoryFor options passed to create clobber injections', (assert) =
 
   assert.equal(instrance.ajax, 'fetch');
 });
-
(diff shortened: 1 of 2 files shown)


packages/container/tests/container_test.js
```diff
@@ -554,8 +554,9 @@ QUnit.test('An object with its owner pre-set should be returned from ownerInject
 });
 
 QUnit.test('A deprecated `container` property is appended to every object instantiated from an extendable factory', function() {
+  let owner = { };
   let registry = new Registry();
-  let container = registry.container();
+  let container = owner.__container__ = registry.container({ owner });
```
rwjblue: FYI - This is exactly what is done outside of these unit tests.  See [EngineInstance](https://github.com/emberjs/ember.js/blob/f44f1d54ce3ffedc1c41309a9746dbfdad570591/packages/ember-application/lib/system/engine-instance.js#L50-L57) code for confirmation.

krisselden: App instance didn't already have a ref to the container?

stefanpenner (approved):

packages/container/tests/container_test.js
krisselden: never mind, I realize here owner is a stub for a unit test, ignore my comment.

PR #15164, opened 2017-04-24
[WIP] Ensure using no-double-extend does not modify an object's own keys.

rwjblue:
Given:

```js
owner.register('some:name', Ember.Object.extend());
let instance = owner.lookup('some:name');

assert.deepEqual(Object.keys(instance), []);
```

This assertion passes in Ember 2.12 and older, but not Ember 2.13.0 beta series. The cause of this regression was disabling `no-double-extend` feature and now some properties like `_debugContainerKey` and `NAME_KEY` are being set on new instances.

### DIFF
--- a/packages/container/lib/container.js
+++ b/packages/container/lib/container.js
@@ -16,6 +16,7 @@ import {
 } from 'ember/features';
 
 const CONTAINER_OVERRIDE = symbol('CONTAINER_OVERRIDE');
+export const FACTORY_MANAGER = symbol('FACTORY_MANAGER');
 export const FACTORY_FOR = symbol('FACTORY_FOR');
 export const LOOKUP_FACTORY = symbol('LOOKUP_FACTORY');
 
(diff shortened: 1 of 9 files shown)


packages/container/lib/container.js
```diff
@@ -618,6 +647,18 @@ class FactoryManager {
     this.normalizedName = normalizedName;
     this.madeToString = undefined;
     this.injections = undefined;
+
+    if (isObject(factory)) {
+      setFactoryManager(factory, this);
+    }
+  }
+
+  factoryToString() {
```
krisselden: why not just make this `toString() {}`

packages/ember-runtime/lib/system/core_object.js
```diff
@@ -540,7 +541,9 @@ CoreObject.PrototypeMixin = Mixin.create({
   toString() {
     let hasToStringExtension = typeof this.toStringExtension === 'function';
     let extension = hasToStringExtension ? `:${this.toStringExtension()}` : '';
-    let ret = `<${this[NAME_KEY] || this.constructor.toString()}:${guidFor(this)}${extension}>`;
+    let factoryManager = peekFactoryManager(Object.getPrototypeOf(this));
+    let factoryToString = this[NAME_KEY] || (factoryManager && factoryManager.factoryToString());
```
krisselden: if `factoryToString()` was just toString() you wouldn't need to call it you can just let the string template call it.

packages/container/lib/container.js
```diff
@@ -618,6 +647,18 @@ class FactoryManager {
     this.normalizedName = normalizedName;
     this.madeToString = undefined;
     this.injections = undefined;
+
+    if (isObject(factory)) {
+      setFactoryManager(factory, this);
```
krisselden: factory here doesn't not have a 1 to 1 with FactoryManager, FactoryManager has only one factory but a factory can have multiple managers.

krisselden (changes requested):

packages/container/lib/container.js
rwjblue: Yeah, this is a fairly major issue for us. I'll ponder it a bit...

Issue #15151, opened 2017-04-17
Ember.typeOf returns "object" for async function


RobIsHere:
https://github.com/emberjs/ember.js/blob/master/packages/ember-runtime/lib/utils.js#L114

Chrome Console:
```
toString.call(item)
"[object AsyncFunction]"
```
(Chrome is behaving right according to: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function)

Expected when checking for an async function with typeOf is "function". 

I'm turning babel transpiling off in dev to get better debugging experience:
```
        babel: {
            sourceMaps: 'inline',
            blacklist: [
                'es6.arrowFunctions',
                'regenerator'
            ]
        },
```
(looks like this is changing at the moment from blacklist to exclude with babel 6 support)

RobIsHere:
Summary: Ember.typeOf will be deprecated. Should use native typeof. PR for this is on the way in ember-route-action-helper.

PR #15165, opened 2017-04-25
add async functions and generator functions to Ember.typeof

anonrose:
https://github.com/emberjs/ember.js/issues/15151

### DIFF
--- a/packages/ember-runtime/lib/utils.js
+++ b/packages/ember-runtime/lib/utils.js
@@ -5,16 +5,17 @@ import EmberObject from './system/object';
 // TYPING & ARRAY MESSAGING
 //
 const TYPE_MAP = {
-  '[object Boolean]':       'boolean',
-  '[object Number]':        'number',
-  '[object String]':        'string',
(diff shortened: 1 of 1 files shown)


mmun:
Should consider generator functions as well.

anonrose:
@mmun thanks

rwjblue:
I think his is probably fine to land, but I do wonder if we should work towards deprecating `Ember.typeOf` completely, and instead suggest folks to "use the platform". All of these small custom "Embery" bits generally chafe the wrong way now that the platform is actually progressing...

anonrose:
@rwjblue I completely agree

stefanpenner:
we should bring this up at the meeting for discussion.

rwjblue:
We discussed this in the weekly Ember core team meeting, and came to the conclusion that this API is not something that we want to continue making changes to. It is likely that we will work towards creating a deprecation RFC.

rwjblue:
Thank you very much for taking the time to work on this! Given the decision to not change this, I am going to close this PR...

PR #15168, opened 2017-04-25
[BUGFIX] Retry transition created with `replaceWith` causes a history replacement

cibernox:
This issue was fixed upstream in router.js, I'm just repeating the information here
for context.
Before this change the next code:

```js
let transition = this.replaceWith('new.route')
transition.abort()
transition.retry();
```

caused the app to transition to `'new.route'` without replacing the previous history entry.
After this change, `Transition#retry` respects the intention of the original transition. In
the snippet above `'new.route'` will now replace the previous route in the history.

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
     "qunit-phantomjs-runner": "^2.2.0",
     "qunitjs": "^1.22.0",
     "route-recognizer": "^0.3.0",
-    "router_js": "^1.2.4",
+    "router_js": "^1.2.7",
     "semver": "^5.3.0",
     "serve-static": "^1.10.0",
     "simple-dom": "^0.3.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2299,14 +2299,14 @@ error@^7.0.0:
     string-template "~0.2.1"
     xtend "~4.0.0"
 
-es5-ext@^0.10.14, es5-ext@^0.10.7, es5-ext@^0.10.9, es5-ext@~0.10.14, es5-ext@~0.10.2, es5-ext@~0.10.7:
+es5-ext@^0.10.14, es5-ext@^0.10.9, es5-ext@~0.10.14:
   version "0.10.14"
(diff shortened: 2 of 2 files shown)


stefanpenner (approved):

workmanw:
Ember try alerted us to some test failures with ember-2.14.0-beta.1 and I bisected them down to this commit. I tried to create a twiddle, but it seems that I cannot do so with beta builds at the moment.

The test failure is simply redirecting an unauthenticated user to the login page, then calling `transition.retry` after login. The transition is successfully redirected to the right route, but it leaves the URL in the wrong state.

Should I open a new issue for this?

cibernox:
@workmanw I'm interested. Is the transition made with replaceWith? Can you share an example?

workmanw:
@cibernox I guess I could probably make an example ember app in lieu of a twiddle.

Here are some examples:

### Saving the retry transition and redirecting

```javascript
// app/routes/authenticated.js
// Everything inside this route requires the user to be authenticated
export default Ember.route.extend({
  afterModel(model, transition) {
    let authService = this.get('auth');
    if (!authService.get('isLoggedIn')) {
      authService.set('retryTransition', transition);
      this.transitionTo('login');
    }
  }
});
```

### User login

```javascript
// app/services/auth.js
export default Ember.Service.extend({
  login(user, password) {
    this.get('ajax').post('/login', { user, password }).then(() => {
      let retryTransition = this.get('retryTransition');
      if (retryTransition) {
        retryTransition.retry();
        this.set('retryTransition', null);
      } else {
        this.get('routerServiceStub').transitionTo('authenticated.home');
      }
    });
  }
});
```

----

So you can see nothing is too out of the ordinary as far as I can tell.

rwjblue:
@workmanw - Thanks for alerting us! Can you create a new issue once you have a repro?

workmanw:
@rwjblue Roger that. I just created a failing repo. I have to jump for a few, but you can expect a new issue later on. Thanks!

PR #15169, opened 2017-04-25
[BUGFIX Beta] [PERF] avoid extra `meta` lookup in constructor

stefanpenner:
Noticed this while profiling ember-data stuff.


rwjblue (commented):
This seems good, but I am pretty surprised to see that `Ember.sendEvent` is marked as public.

Do we need to be concerned with adding another param to this (since its public)? I think probably no, but wanted to make sure...

PR #15174, opened 2017-04-26
[FEATURE ember-engines-mount-params] Allow {{mount}} syntax to accept parameters

mike183:
Implements ember-engines/ember-engines#98 for routeless engines behind the `ember-engines-mount-params` feature flag.

Parameters passed into `{{mount}}` will be accessible within the engine as the model on the engines application controller/template.

RFC Link: https://github.com/emberjs/rfcs/pull/225

Example:

```hbs
{{!-- app/templates/application.hbs --}}
{{mount "foo" bar="baz"}}
```

```hbs
{{!-- foo/templates/application.hbs --}}
{{model.bar}}
```

### DIFF
--- a/packages/ember-glimmer/lib/syntax/mount.js
+++ b/packages/ember-glimmer/lib/syntax/mount.js
@@ -12,6 +12,7 @@ import { generateControllerFactory } from 'ember-routing';
 import { OutletLayoutCompiler } from './outlet';
 import AbstractManager from './abstract-manager';
 import { DEBUG } from 'ember-env-flags';
+import { EMBER_ENGINES_MOUNT_PARAMS } from 'ember/features';
 
 function dynamicEngineFor(vm, symbolTable) {
   let env     = vm.env;
@@ -42,13 +43,20 @@ function dynamicEngineFor(vm, symbolTable) {
   @public
 */
 export function mountMacro(path, params, hash, builder) {
-  assert(
-    'You can only pass a single argument to the {{mount}} helper, e.g. {{mount "chat-engine"}}.',
-    params.length === 1 && hash === null
-  );
+  if (EMBER_ENGINES_MOUNT_PARAMS) {
+    assert(
+      'You can only pass a single positional argument to the {{mount}} helper, e.g. {{mount "chat-engine"}}.',
+      params.length === 1
+    );
+  } else {
+    assert(
+      'You can only pass a single argument to the {{mount}} helper, e.g. {{mount "chat-engine"}}.',
+      params.length === 1 && hash === null
+    );
+  }
 
   let definitionArgs = [params.slice(0, 1), null, null, null];
-  let args = [null, null, null, null];
+  let args = [null, hash, null, null];
   builder.component.dynamic(definitionArgs, dynamicEngineFor, args, builder.symbolTable);
   return true;
 }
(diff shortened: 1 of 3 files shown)


rwjblue (commented):
Looks good to me, but I think we need to keep the assertion around for the positional params (since we do not expose a mechanism to use those in this PR).

rwjblue:
@trentmwillis - Can you take a look also (I couldn't add you as a reviewer 😖 )?

mike183:
FYI the RFC can be found here: https://github.com/emberjs/rfcs/pull/225

mike183:
> Looks good to me, but I think we need to keep the assertion around for the positional params (since we do not expose a mechanism to use those in this PR).

Added the assertion back in for positional params along with a test to ensure its working.

packages/ember-glimmer/lib/syntax/mount.js
```diff
@@ -43,13 +44,20 @@ function dynamicEngineFor(vm, symbolTable) {
...
+  } else {
+    assert(
+      'You can only pass a single argument to the {{mount}} helper, e.g. {{mount "chat-engine"}}.',
+      params.length === 1 && hash === null
+    );
+  }
 
   let definitionArgs = [params.slice(0, 1), null, null, null];
-  let args = [null, null, null, null];
+  let args = [params.slice(1), hash, null, null];
```
trentmwillis: Shouldn't the first arg continue being `null`? If `params.length === 1` then this `slice` seems a bit unnecessary.

packages/ember-glimmer/lib/syntax/mount.js
```diff
@@ -113,25 +121,35 @@ class MountManager extends AbstractManager {
...
 
-  getSelf(engine) {
+  getSelf(bucket) {
+    let { engine, args } = bucket;
+
     let applicationFactory = engine[FACTORY_FOR](`controller:application`);
-    let factory = applicationFactory || generateControllerFactory(engine, 'application');
-    return new RootReference(factory.create());
+    let controllerFactory = applicationFactory || generateControllerFactory(engine, 'application');
+    let controller = bucket.controller = controllerFactory.create();
```
trentmwillis: Pardon my lack of rendering knowledge, but is mutating the `bucket` input here kosher?

trentmwillis (commented):
Overall, seems fine to me, but I'm not super familiar with rendering internals. Some of the mutation/side-effecty things going on seem like a smell, but I'm unsure.

packages/ember-glimmer/lib/syntax/mount.js
rwjblue: Good point

packages/ember-glimmer/lib/syntax/mount.js
rwjblue: Ya, the bucket is a simple state bucket that we can use in these manager hooks.  A very similar thing happens in https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/syntax/curly-component.js (around L330).

packages/ember-glimmer/lib/syntax/mount.js
mike183: Good catch, fixed.

trentmwillis (approved):

rwjblue:
The associated RFC is in final comment period (and seems to be progressing nicely), but we need to address a few things here in the PR:

* Rebase + fix merge conflicts
* Add tests for the scenario described in https://github.com/emberjs/rfcs/pull/225#issuecomment-298084397.  Essentially confirming that `{{mount widget=(component 'some-thing')}}`:
  * the passed in closure component can be rendered inside the engine
  * `some-thing` has access to the host applications registry/owner/etc, and not the engines'. Specifically, the test should confirm that `some-thing'`s template can invoke helpers that only available in the host app.

mike183:
Thanks @rwjblue.

I'll take care of these when I get a moment, should be done within the next day or so.

mike183:
I've rebased, resolved all merge conflicts and also added a test case to cover passing contextual components into params.

rwjblue:
I rebased again, squashed, and updated the commit message.

The RFC upstream has been merged, this should be good to land once CI agrees.

knownasilya:
Will this model be accessible from the `application/route.js` `setupController`?

rwjblue:
@knownasilya - There is no route in `{{mount` based engines, so no.

PR #15175, opened 2017-04-26
[PERF] inlines _initializeListeners in meta

### LINKED ISSUE #15171: We should just inline _initializeListeners in meta
No point in keeping this separate we should just inline it, rather then waiting for the JIT to doit.

https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/lib/meta.js#L127

tylerturdenpants:
This fixes #15171, per @stefanpenner

### DIFF
--- a/packages/ember-metal/lib/meta.js
+++ b/packages/ember-metal/lib/meta.js
@@ -124,9 +124,9 @@ export class Meta {
       }
     }
 
-	   this._listeners = undefined;
-	   this._listenersFinalized = undefined;
-	   this._suspendedListeners = undefined;
+	this._listeners = undefined;
+	this._listenersFinalized = undefined;
+	this._suspendedListeners = undefined;
   }
 
   isInitialized(obj) {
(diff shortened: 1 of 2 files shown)


packages/ember-metal/lib/meta.js
```diff
@@ -124,7 +124,9 @@ export class Meta {
       }
     }
 
-    this._initializeListeners();
+	this._listeners = undefined;
```
stefanpenner: Mind fixing up the indentation?

stefanpenner (changes requested):
nice!!

Let one nit-pick though, but then we can land.

packages/ember-metal/lib/meta.js
tylerturdenpants: sorry, my editor settings!  will fix

tylerturdenpants:
@stefanpenner forgive me, i'm trying to figure out the indentation

tylerturdenpants:
fails saucelab tests, not sure why.

stefanpenner (approved):

stefanpenner:
@tylerturdenpants our sauce labs tests tend to be somewhat brittle, restarting them now. Otherwise this looks good!

PR #15176, opened 2017-04-26
[BUGFIX] controller `replaceRoute` considers engine's mount point

spencer516:
Seems like there was a small error and a missing test when using `replaceRoute` from a controller that is inside of an engine. This change mirrors the same for `transitionToRoute.


rwjblue:
This fix looks good, sorry for taking so long to review! Can you rebase so we can get a new CI run kicked off to make sure it all still passes?

spencer516:
@rwjblue — rebased and it's passed!

rwjblue:
Thank you!

PR #15177, opened 2017-04-27
[BUGFIX beta] Avoid creating enumerable properties on all objects created by DI.

rwjblue:
Prior to these changes, every object created via `owner.factoryFor(...).create()` was populated with `NAME_KEY`, `_debugContainerKey`, and `OWNER`.

After these changes:

* For `Ember.Object`'s, only actual injections are included in the create arguments.
* For non-`Ember.Object`'s, only `OWNER` is included in the create arguments.

`OWNER` is still provided for non-`Ember.Object`'s in order to support various classes that are used in the rendering layer that expect to be passed `options[OWNER]` so they can function properly. Examples are `ember-glimmer/environment` and `ember-glimmer/template`. These should be refactored away from `OWNER` being passed in, and this fallback should be removed.

A handful of tests were added / updated. In some cases `_debugContainerKey` assertions were removed, since this is no longer supported configuration (e.g. there is no way to inject `_debugContainerKey` onto the class itself without double extend).

### DIFF
--- a/packages/container/lib/container.js
+++ b/packages/container/lib/container.js
@@ -418,6 +418,8 @@ function deprecatedFactoryFor(container, fullName, options = {}) {
     let cacheable = !areInjectionsDynamic(injections) && !areInjectionsDynamic(factoryInjections);
 
     factoryInjections[NAME_KEY] = registry.makeToString(factory, fullName);
+    injections._debugContainerKey = fullName;
+    setOwner(injections, container.owner);
 
     let injectedFactory = factory.extend(injections);
 
@@ -443,9 +445,6 @@ function injectionsFor(container, fullName) {
   let type = splitName[0];
 
   let injections = buildInjections(container, registry.getTypeInjections(type), registry.getInjections(fullName));
-  injections._debugContainerKey = fullName;
-
-  setOwner(injections, container.owner);
 
   return injections;
 }
@@ -488,6 +487,7 @@ function instantiate(factory, props, container, fullName) {
       // to create time injections
       // TODO: support new'ing for instantiation and merge injections for pure JS Functions
       let injections = injectionsFor(container, fullName);
+      injections._debugContainerKey = fullName;
 
       // Ensure that a container is available to an object during instantiation.
       // TODO - remove when Ember reaches v3.0.0
@@ -613,11 +613,20 @@ class DeprecatedFactoryManager {
 class FactoryManager {
   constructor(container, factory, fullName, normalizedName) {
     this.container = container;
+    this.owner = container.owner;
     this.class = factory;
     this.fullName = fullName;
     this.normalizedName = normalizedName;
     this.madeToString = undefined;
     this.injections = undefined;
+  }
+
+  toString() {
+    if (!this.madeToString) {
+      this.madeToString = this.container.registry.makeToString(this.class, this.fullName);
+    }
+
+    return this.madeToString;
   }
 
   create(options = {}) {
@@ -631,7 +640,6 @@ class FactoryManager {
     }
     let props = assign({}, injections, options);
 
-    props[NAME_KEY] = this.madeToString || (this.madeToString = this.container.registry.makeToString(this.class, this.fullName));
 
     if (DEBUG) {
       let lazyInjections;
@@ -654,6 +662,21 @@ class FactoryManager {
     let prototype = this.class.prototype;
     if (prototype) {
       injectDeprecatedContainer(prototype, this.container);
+    }
+
+    // required to allow access to things like
+    // the customized toString, _debugContainerKey,
+    // owner, etc. without a double extend and without
+    // modifying the objects properties
+    if (typeof this.class._initFactory === 'function') {
+      this.class._initFactory(this);
+    } else {
+      // in the non-Ember.Object case we need to still setOwner
+      // this is required for supporting glimmer environment and
+      // template instantiation which rely heavily on
+      // `options[OWNER]` being passed into `create`
+      // TODO: clean this up, and remove in future versions
+      setOwner(props, this.owner);
     }
 
     return this.class.create(props);
--- a/packages/ember-metal/lib/meta.js
+++ b/packages/ember-metal/lib/meta.js
@@ -98,6 +98,7 @@ export class Meta {
     this._chains = undefined;
     this._tag = undefined;
     this._tags = undefined;
+    this._factory = undefined;
 
     // initial value for all flags right now is false
     // see FLAGS const for detailed list of flags used
@@ -344,6 +345,14 @@ export class Meta {
     } else {
       obj[key] = value;
     }
+  }
+
+  set factory(factory) {
+    this._factory = factory;
+  }
+
+  get factory() {
+    return this._factory;
   }
 }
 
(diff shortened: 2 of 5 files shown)


rwjblue:
Did a quick benchmark via ember-macro-benchmark to see if this has an effect on perf and it looks pretty good to me. 

@krisselden / @stefanpenner - Mind confirming my interpretation?

phases

Raw data: [results-pr-15177.zip](https://github.com/emberjs/ember.js/files/960118/results-pr-15177.zip)

rwjblue:
I chatted with @krisselden in slack about the results, we came to roughly the same conclusion. That the changes in this PR do not reject the null hypothesis (that the variance between control and experiment are random) and therefore these changes do not represent a change in the performance profile of the application under test (emberaddons.com in this case).

rwjblue:
Changes like this (changing the `makeCtor` functionality) should receive significant scrutiny, so I'd still like @stefanpenner and/or @krisselden to review this PR and approve....

packages/container/lib/container.js
```diff
@@ -656,6 +664,21 @@ class FactoryManager {
       injectDeprecatedContainer(prototype, this.container);
     }
 
+    // required to allow access to things like
+    // the customized toString, _debugContainerKey,
+    // owner, etc. without a double extend and without
+    // modifying the objects properties
+    if (this.class._initFactory) {
```
stefanpenner: `typeof this.class._initFactory === 'function'` (more correct, and makes the JIT happy)

packages/ember-runtime/lib/system/core_object.js
```diff
@@ -540,7 +546,11 @@ CoreObject.PrototypeMixin = Mixin.create({
   toString() {
     let hasToStringExtension = typeof this.toStringExtension === 'function';
     let extension = hasToStringExtension ? `:${this.toStringExtension()}` : '';
-    let ret = `<${this[NAME_KEY] || this.constructor.toString()}:${guidFor(this)}${extension}>`;
+
+    let m = meta(this);
```
stefanpenner: 550 and 551 are only required if `this[NAME_KEY]` isn't set, and it will be set most of the time. Can we avoid 550 and 551 unless we need to access it?

something like:

```js
this[NAME_KEY] || meta(this).getFactory() || this.cosntructor.toString()
```

packages/ember-metal/lib/meta.js
```diff
@@ -345,6 +346,14 @@ export class Meta {
       obj[key] = value;
     }
   }
+
+  setFactory(factory) {
```
stefanpenner: Any preference to function accessors vs normal accessors? I can be fine with either, was just surprised this isn't a accessor and am curious if there was a reason for it

packages/ember-runtime/lib/system/object.js
```diff
@@ -21,7 +24,48 @@ import { DEBUG } from 'ember-env-flags';
...
+const EmberObject = CoreObject.extend(Observable, {
+  _debugContainerKey: descriptor({
+    enumerable: false,
+    get() {
+      if (this[OVERRIDE_CONTAINER_KEY]) {
+        return this[OVERRIDE_CONTAINER_KEY];
+      }
+
+      let meta = peekMeta(this);
+      let factory  = meta.getFactory();
```
stefanpenner: if we `peekMeta` `meta` may be `undefined` so we must guard before accessing `getFactory`

something like:

```js
let meta = peekMeta(this)
if (meta === undefined) { return; }
let factory = meta.getFactory();
if (factory === undefined) { return ; }
return factory.fullName;
```

packages/ember-runtime/lib/system/object.js
```diff
@@ -21,7 +24,48 @@ import { DEBUG } from 'ember-env-flags';
...
+        return this[OVERRIDE_OWNER];
+      }
+
+      let meta = peekMeta(this);
+      let factory  = meta.getFactory();
+
+      return factory && factory.owner;
+    },
+
+    // we need a setter here largely to support the legacy
```
stefanpenner: ah, so this `[OVERRIDE_OWNER]` both in the setter and getter can be removed once the legacy support is dropped?

packages/ember-runtime/lib/system/object.js
```diff
@@ -21,7 +24,48 @@ import { DEBUG } from 'ember-env-flags';
...
+  }),
+
+  [OWNER]: descriptor({
+    enumerable: false,
+    get() {
+      if (this[OVERRIDE_OWNER]) {
+        return this[OVERRIDE_OWNER];
+      }
+
+      let meta = peekMeta(this);
```
stefanpenner: Is it possible to have a ember-object without a meta? As we create its meta during within its constructor.

packages/ember-runtime/lib/system/object.js
```diff
@@ -21,7 +24,48 @@ import { DEBUG } from 'ember-env-flags';
...
-const EmberObject = CoreObject.extend(Observable);
+const EmberObject = CoreObject.extend(Observable, {
+  _debugContainerKey: descriptor({
+    enumerable: false,
+    get() {
+      if (this[OVERRIDE_CONTAINER_KEY]) {
+        return this[OVERRIDE_CONTAINER_KEY];
+      }
+
+      let meta = peekMeta(this);
```
stefanpenner: Is it possible to have a ember-object without a meta? As we create its meta during within its constructor.

stefanpenner (commented):
looks good, left some nit-picks.

packages/ember-runtime/lib/system/object.js
rwjblue: Yes, confirm

packages/ember-runtime/lib/system/object.js
krisselden: it is not possible for it not to exists

packages/ember-runtime/lib/system/object.js
krisselden: confirm, OVERRIDE_OWNER is only for deprecated container setter

krisselden (approved):

packages/ember-runtime/lib/system/object.js
stefanpenner: isn't it a tad strange to use `peekMeta` instead of `meta` ?

packages/ember-metal/lib/meta.js
rwjblue: `get` / `set` seem fine to me, I just _slightly_ dislike the [babel transpilation](http://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=es2015-loose&targets=&browsers=&builtIns=false&debug=false&code=class%20Foo%20%7B%0A%20%20get%20derp()%20%7B%20%7D%0A%20%20set%20derp(value)%20%7B%20%7D%0A%7D):

```js
// input

class Foo {
  get derp() { }
  set derp(value) { }
}

// output
var Foo = function () {
  function Foo() {
    _classCallCheck(this, Foo);
  }

  _createClass(Foo, [{
    key: "derp",
    get: function get() {},
    set: function set(value) {}
  }]);

  return Foo;
}();
```

PR #15181, opened 2017-04-27
[FEATURE ember-no-double-extend] Remove feature flagging and cleanup.

rwjblue:
* Remove `FACTORY_FOR` symbol. This was created as a way to internally use the new `.factoryFor` system, and not worry if `no-double-extend` feature was enabled or disabled.
* Remove `LOOKUP_FACTORY` symbol. This was added as a "backdoor" to use `container.lookupFactory` (and its double extend semantics) without triggering a deprecation.


stefanpenner (approved):
LGTM

PR #15183, opened 2017-04-27
Update dependencies to the latest versions within range.

rwjblue:
Most of these were already allowed via our `package.json`, but I also wanted to update the `yarn.lock`.

This pulls in some goodies from `rsvp`, `route-recognizer`, `babel`, `dag-map`, `babel-plugin-debug-macros`, and more...


stefanpenner:
should we enable greenkeeper, or will that be too annoying?

rwjblue:
@stefanpenner - IMO, it will be somewhat annoying until they also update the `yarn.lock`.  The builds that we publish (both for Bower and S3) really need to be using the `yarn.lock` for generation, so when we bring in greenkeeper we would have to continually update `yarn.lock` manually (which is roughly where we are already 😝 ).

rwjblue:
Had to lock Testem down to 1.15.0 due to https://github.com/testem/testem/issues/1115.

PR #15184, opened 2017-04-27
[DOC release] Consider making `Ember.expandProperties` public

SaladFork:
Merging this PR would make `Ember.expandProperties` public.

Often when working on computed property macros, it is helpful to be able to parse dependent keys in a manner similar to how Ember does. Though special properties like `@each` and `[]` remain tricky, making `expandProperties` public would help greatly with consistent handling of brace expansion.

This private method has already been used in addons that provide computed property macros, such as [`ember-computed-decorators`](https://github.com/rwjblue/ember-computed-decorators), recently through [`ember-macro-helpers`](https://github.com/kellyselden/ember-macro-helpers).

### DIFF
--- a/packages/ember-metal/lib/expand_properties.js
+++ b/packages/ember-metal/lib/expand_properties.js
@@ -29,7 +29,7 @@ var END_WITH_EACH_REGEX = /\.@each$/;
 
   @method expandProperties
   @for Ember
-  @private
+  @public
   @param {String} pattern The property pattern to expand.
   @param {Function} callback The callback to invoke.  It is invoked once per
   expansion, and is passed the expansion.


rwjblue (approved):
IMO, this is already at least intimate API (e.g. would have to go through a deprecation + LTS cycle before removal even if we wanted to), and it makes sense to make it public (I do not _really_ see us removing it in the future anyways)...

Need a few other core team 👍 I think before we should land this...

locks:
Hey @SaladFork, this was discussed at the last core meeting and we decided to give the opposite of a thumbs up :trollface:
I kid, the actual decision was that this should be exposed, but in the new modules defined in [RFC #176](https://github.com/emberjs/rfcs/blob/master/text/0176-javascript-module-api.md).
The decision to not make it public right away may seem odd, but new modules are already going through the release cycle and are expected to be the default in 2.16. Since that's only one release away, we'd rather not add another thing to the `Ember` namespace.

Thank you for bringing this to our attention, seems like peeps agree that it should be public!
Could you make a PR to https://github.com/ember-cli/ember-rfc176-data adding `expandProperties` as a named export of `@ember/utils`?
It'd be used like so:

```javascript
import { expandProperties } from '@ember/utils';
```

Cheers!

rwjblue:
Unfortunately, I wasn't able to attend the last meeting where this was discussed. 

I'm pretty confused by the last comment by @locks. It seems like it was assumed that this PR was suggesting adding a new API to the `window.Ember` global, but that is completely untrue. `Ember.expandProperties` has been exposed for quite some time

If we are saying this API is public, this PR should be merged.

rwjblue:
Reopening this so we don't loose track of it while we discuss.

locks:
So… I confused myself and I still not quite sure what happened there :P But the decision was to indeed make this public.

PR #15185, opened 2017-04-28
correct casing of _routermicrolib deprecation url

Dhaulagiri:
The current link will not actually scroll down to the anchor specified because of the casing difference


rwjblue:
Ugh, thank you for correcting this!

Can you update the commit with a `[DOC release]` prefix?

Dhaulagiri:
@rwjblue done

stefanpenner (approved):

PR #15186, opened 2017-04-28
[Bugfix] Test for undefined in `path.parts`

dhilipsiva:
Attempting to fix this error: https://circleci.com/gh/appknox/irene/1372


rwjblue:
Can you share more about the underlying issue? What is this trying to solve?

rwjblue:
Perhaps you could add a test to https://github.com/emberjs/ember.js/blob/master/packages/ember-template-compiler/tests/plugins/transform-dot-component-invocation-test.js to help me understand the failure scenario better?

dhilipsiva:
@rwjblue Apologies. This does not seem to have solved my problem. Will close this for now and will investigate more.

PR #15188, opened 2017-04-28
[BUGFIX beta] don’t assume a controller supports queryParams

stefanpenner:
* use setProperties helper over controller.setProperties
* default queryParams to {} if not present

- [x] needs test


rwjblue (approved):

stefanpenner:
pushed fix, tests are now green locally.

PR #15189, opened 2017-04-28
added detection for backslash folder delimiter

SpikedKira:
`npm run build` currently fails here on platforms that don't use `/` to separate folders in paths.


rwjblue (approved):

rwjblue:
Thank you!!

PR #15192, opened 2017-04-30
Remove support for deprecated `.container` property on objects.

rwjblue:
This has been deprecated since Ember 2.3 (through multiple LTS releases).


rwjblue:
This is in a similar area but independent of #15193.

stefanpenner (approved):
:shipit:

PR #15193, opened 2017-04-30
Remove `_lookupFactory` support.

rwjblue:
This was deprecated in Ember 2.12 (which was an LTS version).

This removes `owner._lookupFactory` APIs completely. It also includes
factory injections, and factory type injections.

### DIFF
--- a/packages/container/lib/container.js
+++ b/packages/container/lib/container.js
@@ -166,6 +166,10 @@ Container.prototype = {
       return;
     }
 
+    if (DEBUG && factory && typeof factory._onLookup === 'function') {
+      factory._onLookup(fullName);
+    }
+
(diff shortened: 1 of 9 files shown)


rwjblue:
This is in a similar area but independent of https://github.com/emberjs/ember.js/pull/15192.

rwjblue:
Please note that there is one commit that is actually a `BUGFIX beta` (but may need to be `BUGFIX lts`). It turns out that we did not validate things like `Ember.inject.service` when `owner.factoryFor` was used (this is because we were only invoking `_onLookup` in the `lookupFactory` path.

stefanpenner (approved):

stefanpenner:
@rwjblue thanks for the cleanup!

Issue #15198, opened 2017-05-02
ember-template-compiler.js output is incorrect


rwjblue:
`ember-template-compiler.js` contains development builds of the various packages, but the production build of `ember-babel`. This causes errors because `_classCallCheck` is not present in the production version of `ember-babel` package, but the `_classCallCheck`'s have not been stripped from the various transpiled classes (due to them being transpiled as a debug build).

PR #15199, opened 2017-05-02
[BUGFIX beta] Always emit the same ember-template-compiler.js (as a debug build).

rwjblue:
Changing the output of `ember-template-compiler.js` based on `ember b -prod` means that the resulting asset can only be used with `ember.prod.js` (and not with `ember.debug.js`).

This removes the customization and forces all versions of `ember-template-compiler.js` to be in debug mode.

Addresses #15198

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -261,25 +261,6 @@ module.exports = function(options) {
       bootstrapModule('ember')
     ]);
 
-    // Note:
-    // We have to build custom production template compiler
-    // because we strip babel helpers in the prod build
-    let prodTemplateCompiler = new MergeTrees(templateCompiler(babelProdHelpersES5));
-
-    prodTemplateCompiler = stripForProd(prodTemplateCompiler)
-
-    prodTemplateCompiler = new MergeTrees([
-      nodeModule,
-      prodTemplateCompiler,
-      version,
-      license,
-      loader
-    ]);
-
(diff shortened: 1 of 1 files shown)


chadhietala (approved):

ember-cli-build.js
```diff
@@ -305,22 +286,22 @@ module.exports = function(options) {
...
-        nodeModule
-    ]);
+    trees.push(emberRuntimeBundle, emberProdBundle, emberMinBundle, emberProdTestsBundle);
+  }
 
-    emberTemplateCompilerBundle = concat(emberTemplateCompilerBundle, {
-      outputFile: 'ember-template-compiler.js'
-    });
+  let emberTemplateCompilerBundle = new MergeTrees([
+    ...templateCompiler(babelDebugHelpersES5),
```
cibernox: I believe that the spread operator is behind the harmony flag in node 4: http://node.green/#ES2015-syntax-spread-------operator

cibernox (changes requested):

ember-cli-build.js
rwjblue: Ya, confirm but we are using node 6 (and have a few other usages of node 6isms).

cibernox: I asked in case this code is run (or will eventually) but consuming apps. Then, no arguments.

cibernox (approved):

cibernox:
@rwjblue what is the status of this?

PR #15204, opened 2017-05-03
[BUGFIX release] MODEL_FACTORY_INJECTIONS is now always false.

stefanpenner:
Double extend means this flag no longer means what it once did. To ensure ember-data functions correctly, we should now force it to always be false.

Code affect in ember-data is: https://github.com/emberjs/data/blob/e6cb564bdb33da4b3c15c3c668dc4b1fdafdf190/addon/-debug/index.js#L37-L39

---

- [x] add some tests ensuring the behavior is as expected
- [x] remove from ember-cli blueprints: ember-cli/ember-cli#7025
- [X] add to website's deprecations: https://github.com/emberjs/website/pull/2895

### DIFF
--- a/packages/ember-environment/lib/index.js
+++ b/packages/ember-environment/lib/index.js
@@ -1,7 +1,6 @@
 /* globals module */
 import global from './global';
 import { defaultFalse, defaultTrue, normalizeExtendPrototypes } from './utils';
-
 /**
   The hash of environment variables used to control various configuration
   settings. To specify your own or override default settings, add the
@@ -65,9 +64,6 @@ ENV.LOG_STACKTRACE_ON_DEPRECATION = defaultTrue(ENV.LOG_STACKTRACE_ON_DEPRECATIO
   @public
 */
 ENV.LOG_VERSION = defaultTrue(ENV.LOG_VERSION);
-
-// default false
-ENV.MODEL_FACTORY_INJECTIONS = defaultFalse(ENV.MODEL_FACTORY_INJECTIONS);
 
 /**
   Debug parameter you can turn on. This will log all bindings that fire to
--- a/packages/ember/lib/index.js
+++ b/packages/ember/lib/index.js
@@ -1,4 +1,5 @@
 import require, { has } from 'require';
+import { DEBUG } from 'ember-env-flags';
 
 // ****ember-environment****
 import { ENV, context } from 'ember-environment';
@@ -174,11 +175,23 @@ Object.defineProperty(Ember, 'LOG_VERSION', {
   enumerable: false
 });
 
-Object.defineProperty(Ember, 'MODEL_FACTORY_INJECTIONS', {
-  get()      { return ENV.MODEL_FACTORY_INJECTIONS;    },
-  set(value) { ENV.MODEL_FACTORY_INJECTIONS = !!value;  },
-  enumerable: false
-});
+if (DEBUG) {
+  Object.defineProperty(Ember, 'MODEL_FACTORY_INJECTIONS', {
+    get()      { return false; },
+    set(value) {
+      deprecate(
+        'Ember.MODEL_FACTORY_INJECTIONS is no longer required',
+        false,
+        {
+          id: 'ember-metal.model_factory_injections',
+          until: '2.17.0',
+          url: 'http://emberjs.com/deprecations/v2.x#toc_code-ember-model-factory-injections'
+        }
+      );
+    },
+    enumerable: false
+  });
+}
 
 Object.defineProperty(Ember, 'LOG_BINDINGS', {
   get()      { return ENV.LOG_BINDINGS;    },
--- a/packages/container/tests/container_test.js
+++ b/packages/container/tests/container_test.js
@@ -4,16 +4,7 @@ import { get } from 'ember-metal';
 import { Registry } from '..';
 import { factory } from 'internal-test-helpers';
 
-let originalModelInjections;
-
-QUnit.module('Container', {
-  setup() {
-    originalModelInjections = ENV.MODEL_FACTORY_INJECTIONS;
-  },
-  teardown() {
-    ENV.MODEL_FACTORY_INJECTIONS = originalModelInjections;
-  }
-});
+QUnit.module('Container');
 
 QUnit.test('A registered factory returns the same instance each time', function() {
   let registry = new Registry();
@@ -166,8 +157,6 @@ QUnit.test('An invalid factory throws an error', function() {
 });
 
 QUnit.test('Injecting a failed lookup raises an error', function() {
-  ENV.MODEL_FACTORY_INJECTIONS = true;
-
   let registry = new Registry();
   let container = registry.container();
 
--- a/packages/ember-application/tests/system/dependency_injection/to_string_test.js
+++ b/packages/ember-application/tests/system/dependency_injection/to_string_test.js
@@ -5,13 +5,10 @@ import Application from '../../../system/application';
 import { Object as EmberObject } from 'ember-runtime';
 import DefaultResolver from '../../../system/resolver';
 
-let originalLookup, App, originalModelInjections;
+let originalLookup, App;
 
 QUnit.module('Ember.Application Dependency Injection – toString', {
   setup() {
-    originalModelInjections = ENV.MODEL_FACTORY_INJECTIONS;
-    ENV.MODEL_FACTORY_INJECTIONS = true;
-
     originalLookup = context.lookup;
 
     run(() => {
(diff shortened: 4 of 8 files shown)


stefanpenner:
This (if others agree) should likely join w/e release/beta/canary that disables double extend so that ember-data works correctly even if the user has the flag set to true.

packages/ember-environment/lib/index.js
```diff
@@ -66,8 +66,10 @@ ENV.LOG_STACKTRACE_ON_DEPRECATION = defaultTrue(ENV.LOG_STACKTRACE_ON_DEPRECATIO
 */
 ENV.LOG_VERSION = defaultTrue(ENV.LOG_VERSION);
 
-// default false
-ENV.MODEL_FACTORY_INJECTIONS = defaultFalse(ENV.MODEL_FACTORY_INJECTIONS);
+// always false
+Object.defineProperty(ENV, 'MODEL_FACTORY_INJECTIONS', {
```
rwjblue: We should issue a deprecation if `MODEL_FACTORY_INJECTIONS in ENV`, and update ember-cli to confirm it isn't specified in the blueprint. 

Then instead of setting it to false here, we just leave it undefined (which would also be seen as falsey by ED's assertion).

packages/ember/lib/index.js
```diff
@@ -181,8 +181,10 @@ Object.defineProperty(Ember, 'LOG_VERSION', {
 });
 
 Object.defineProperty(Ember, 'MODEL_FACTORY_INJECTIONS', {
-  get()      { return ENV.MODEL_FACTORY_INJECTIONS;    },
-  set(value) { ENV.MODEL_FACTORY_INJECTIONS = !!value;  },
+  get()      { return false; },
+  set(value) {
+    Ember.logger.warn('As of 2.13.x MODEL_FACTORY_INJECTIONS has no power, feel free to remove it from app/app.js');
```
rwjblue: Should be an assertion, and this whole Object.defineProperty block should be in an `if (DEBUG) {`.

rwjblue (changes requested):

rwjblue:
Also, I agree, we should backport to release branch for a 2.13.1 release.

stefanpenner:
> and update ember-cli to confirm it isn't specified in the blueprint.

This is a tad tricky, as it may be confusing for users upgrading ember + ember-cli but not ember-data ? Thoughts?

rwjblue:
The ember-cli blueprints assume the package versions specified in that blueprints package.json. If the user isn't updating ember-data, they can't remove the flag either. 

The main things I'm concerned with (and my comments would address) are:

* We have finally removed special meaning of this flag in Ember, we shouldn't bloat Embers file size permenantly just to set a flag to a falsey value (which is what it's value would be anyways 😜).
* Keeping the flag around, when it has no effect is pretty confusing. A properly worded deprecation (with a link to the deprecations page for more details) can provide the right context to provide folks with guidance.

packages/ember/lib/index.js
stefanpenner: Assertion sounds good but I don't think we can strip the flag as users who ignore the deprecation and set the flag to true will have a bad time.

rwjblue: @stefanpenner - Setting it to true has literally no effect in production builds.  The only usage that I can see in Ember Data is [here](https://github.com/emberjs/data/blob/e6cb564bdb33da4b3c15c3c668dc4b1fdafdf190/addon/-debug/index.js#L37) which is _also_ in an `if (DEBUG)` and therefore stripped in production builds.  

The only thing we need to make sure to do here, is blow up if someone attempts to `Ember.ENV.MODEL_FACTORY_INJECTIONS = <anything>` in debug builds, because in production builds there is no behavior changes anyways...

stefanpenner:
> Keeping the flag around, when it has no effect is pretty confusing.

Ya, it has an affect that effect is backward compact to ease upgrading. Let keep it around for an LTS and then drop it? (like other deprecations)

Sounds ok?

rwjblue:
> Ya, it has an affect that effect is backward compact to ease upgrading. Let keep it around for an LTS and then drop it? (like other deprecations)

It actually has no effect in production versions (as I mentioned and linked to code for in another comment), I'm super happy to keep both the deprecation (when `MODEL_FACTORY_INJECTIONS in ENV`) in `packages/ember-environment/lib/index.js` and either a deprecation or assertion in `packages/ember/lib/index.js` until 2.16 but both should be in an `if (DEBUG) {` flag so it is stripped in production.

rwjblue:
FWIW, I checked ember-data back to v2.2.0 and the only thing that it was using the `MODEL_FACTORY_INJECTIONS` flag for in `addon/` dir was this same assertion.

stefanpenner:
> It actually has no effect in production versions (

Ya good point. Ok we can strip that one out in prod builds!

packages/ember-environment/lib/index.js
```diff
@@ -66,8 +67,19 @@ ENV.LOG_STACKTRACE_ON_DEPRECATION = defaultTrue(ENV.LOG_STACKTRACE_ON_DEPRECATIO
...
-// default false
-ENV.MODEL_FACTORY_INJECTIONS = defaultFalse(ENV.MODEL_FACTORY_INJECTIONS);
+if (DEBUG) {
+  if ('MODEL_FACTORY_INJECTIONS' in ENV) {
+    deprecate(
+      'Ember.MODEL_FACTORY_INJECTIONS is no longer required',
+      false,
+      {
+        id: 'ember-metal.model_factory_injections',
+        until: '3.0.0',
```
stefanpenner: @rwjblue what should `until` be?

stefanpenner: @rwjblue should I make this page or is this deprecation self-explanatory?

packages/ember/lib/index.js
```diff
@@ -180,11 +180,23 @@ Object.defineProperty(Ember, 'LOG_VERSION', {
...
+if (DEBUG) {
+  Object.defineProperty(Ember, 'MODEL_FACTORY_INJECTIONS', {
+    get()      { return false; },
+    set(value) {
+      deprecate(
+        'Ember.MODEL_FACTORY_INJECTIONS is no longer required',
+        false,
+        {
+          id: 'ember-metal.model_factory_injections',
+          until: '3.0.0',
```
stefanpenner: same as above

packages/ember/lib/index.js
```diff
@@ -180,11 +180,23 @@ Object.defineProperty(Ember, 'LOG_VERSION', {
...
+          url: 'http://emberjs.com/deprecations/v2.x#toc_code-ember-model-factory-injections'
```
stefanpenner: same as above

packages/ember-environment/lib/index.js
```diff
@@ -66,8 +67,19 @@ ENV.LOG_STACKTRACE_ON_DEPRECATION = defaultTrue(ENV.LOG_STACKTRACE_ON_DEPRECATIO
 */
 ENV.LOG_VERSION = defaultTrue(ENV.LOG_VERSION);
 
-// default false
-ENV.MODEL_FACTORY_INJECTIONS = defaultFalse(ENV.MODEL_FACTORY_INJECTIONS);
+if (DEBUG) {
+  if ('MODEL_FACTORY_INJECTIONS' in ENV) {
+    deprecate(
+      'Ember.MODEL_FACTORY_INJECTIONS is no longer required',
```
rwjblue: I think this should be `EmberENV.MODEL_FACTORY_INJECTIONS`?

packages/ember-environment/lib/index.js
rwjblue: Hmm, good question. This isn't _really_ a deprecation per-se, just a handy warning that a setting that used to matter doesn't any more.

We can either swap this with a `warn` (instead of `deprecate`) or leave it as a deprecation and say `until: "2.17.0"`...

rwjblue (commented):
This is looking good, I'm happy with either `deprecate` or `warn` here.  If we leave it as `deprecate` we need to add a section to the website to link to...

stefanpenner:
PR to removing these from blueprints: https://github.com/ember-cli/ember-cli/pull/7025

stefanpenner:
PR with website: https://github.com/emberjs/website/pull/2895

stefanpenner:
@rwjblue r?

stefanpenner:
@rwjblue r?

stefanpenner:
@rwjblue let me know if this should be `[BUGFIX release]` or `[BUGFIX beta]`

stefanpenner:
updated to release.

mixonic (approved):
This seems great. 👍

rwjblue (approved):

Issue #15201, opened 2017-05-03
Ember appears indifferent to closing brace in computed property brace expansion


kennethdavidbuck:
This doesn't seem like a big deal, but I noticed that, when it comes to computed properties, Ember seems indifferent to the presence of a closing brace when doing brace expansion. Please see the following twiddle that demonstrates the behaviour (I would not expect model.bar or model.baz to be watched and updated, given that the end of the brace expansion is missing)

https://ember-twiddle.com/dc26de45263b444157865c7c7f6b7d28?openFiles=controllers.application.js%2C

Serabe:
Thank you, this is fixed in 2.13. I am adding a test for this case and checking if a fix should be backported to 2.12.

Thank you!

kennethdavidbuck:
Thats great, thank you!

Serabe:
@stefanpenner any reason to keep this open?

kennethdavidbuck:
Gonna assume it's safe to close this now :)

PR #15205, opened 2017-05-03
[CLEANUP] Add test for brace expansion assertion

Serabe:
Adding a case for a not allowed brace expansion. This case is failing in
2.12 as shown in #15201 but not in 2.13.

I'm adding this tests because, though the previous cases were added over
six months ago, they did not catch it.


rwjblue:
Awesome, thank you!

PR #15210, opened 2017-05-05
[BUGFIX beta] property_set should reuse reference to its meta

stefanpenner:


scalvert:
Good catch! No peeking :)

krisselden (approved):

packages/ember-runtime/lib/mixins/array.js
```diff
@@ -653,22 +653,26 @@ EachProxy.prototype = {
   arrayWillChange(content, idx, removedCnt, addedCnt) {
     let keys = this._keys;
     let lim = removedCnt > 0 ? idx + removedCnt : -1;
+    let meta;
     for (let key in keys) {
+      meta = meta || peakMeta(this);
```
runspired: is this a misspelling ?

runspired: It is, should be `peekMeta`

runspired: although I do like that we've hit "peak meta"

packages/ember-runtime/lib/mixins/array.js
```diff
@@ -653,22 +653,26 @@ EachProxy.prototype = {
...
+      propertyWillChange(this, key, meta);
     }
   },
 
   arrayDidChange(content, idx, removedCnt, addedCnt) {
     let keys = this._keys;
     let lim = addedCnt > 0 ? idx + addedCnt : -1;
+    let meta;
     for (let key in keys) {
+      meta = meta || peakMeta(this);
```
runspired: same as above, `peekMeta`

runspired (changes requested):
Fix da peak

runspired (approved):

stefanpenner:
@bmeurer this should eat into the 5% of time spent in weakmap lookup you guys where seeing.

stefanpenner:
@homu r+

stefanpenner:
@homu retry

stefanpenner:
@homu what's taking you so long :(

rwjblue:
I think homu is basically broken here :(

bmeurer:
I don't have the full context on this change.

stefanpenner:
Sorry. This pull request removes repeated and unnecessary `weakmap get`'s... I realize you guys are also working on improving weakmap perf so I figure I would loop you in.

TL;DR this pr reduces the number of WeakMapGets's to at least 1/3 in some cases.

bmeurer:
Ah I see. That makes total sense, no matter what V8 (or any other engine) is doing.

stefanpenner:
@bmeurer yup, should work in combination with any work engines do.

PR #15213, opened 2017-05-07
[BUGFIX LTS] Warn about invalid properties

Serabe:
There was a bug in the brace expansion implementation before 2.13 that
admitted expressions like `a.{b,c` and `a.b,c` in property expansion
without failing.

On February 2017, #13231 was merged improving the implementation of
`expandProperties` and fixing the previous issue.

Sadly, people upgrading from 2.12 are having problems with this. We have
two options here:

- Backporting #13231 fixes it and can be done, as of today, without any
  problem just by cherry-picking its two commits. Sadly, this would
  break people apps in a patch version, and this bug does not seem like
  a critical bugfix.

- Adding a warning if a user is using invalid properties. I went with
  this option since it looks like the less intrusive one while letting
  the users know that their apps will break in the next update.

### DIFF
--- a/packages/ember-metal/lib/expand_properties.js
+++ b/packages/ember-metal/lib/expand_properties.js
@@ -64,8 +64,7 @@ export default function expandProperties(pattern, callback) {
     })(pattern));
 
   warn(
-    `You used the key "${pattern}" which is invalid and is not longer expanding on Ember 2.13.`,
-    ((str) => {
+    ...((str, opts) => {
       let inBrace = 0;
       let char;
       for (let i = 0; i < str.length; i++) {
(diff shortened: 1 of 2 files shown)


Serabe:
I used `warn` because I did not feel like we were _deprecating_ a feature, but feedback is welcome as always!

rwjblue:
Seems good to me.  Choosing between `warn` and `deprecate` here is somewhat hard, but I agree that this isn't really a "deprecation" per-se.

Serabe:
Updated messages to depend on the reason it stopped working. Properties like `model.{baz` will not longer work but expressions like `a,b` will continue to work though they will not expand.

Thank you!

rwjblue (approved):

rwjblue:
Thank you @Serabe!

PR #15216, opened 2017-05-09
[DOC release] ember-glimmer/component: Remove duplicate API docs

Turbo87:
/cc @locks @rwjblue

### DIFF
--- a/packages/ember-glimmer/lib/component.js
+++ b/packages/ember-glimmer/lib/component.js
@@ -719,35 +719,9 @@ const Component = CoreView.extend(
      */
 
     /**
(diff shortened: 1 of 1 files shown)


rwjblue:
I don't understand. Where are they duplicated?

Turbo87:
@rwjblue right in the same file. the same API doc comments exist twice.

rwjblue:
Ahh, but if you look they are not actually the same 😉. One is describing the `@event` the other is describing the `@method`.  They are two different things...

Turbo87:
see, I'm blind...

PR #15219, opened 2017-05-09
expandProperties optimizations

SpikedKira:
benchmarks: https://jsperf.com/expandproperties/1

### DIFF
--- a/packages/ember-metal/lib/expand_properties.js
+++ b/packages/ember-metal/lib/expand_properties.js
@@ -40,55 +40,36 @@ export default function expandProperties(pattern, callback) {
     'Brace expanded properties cannot contain spaces, e.g. "user.{firstName, lastName}" should be "user.{firstName,lastName}"',
     pattern.indexOf(' ') === -1
   );
+  // regex to look for double open, double close, or unclosed braces
+  assert(
+    `Brace expanded properties have to be balanced and cannot be nested, pattern: ${pattern}`,
+    pattern.match( /\{[^}{]*\{|\}[^}{]*\}|\{[^}]*$/g ) === null
+  );
 
-  let unbalancedNestedError = `Brace expanded properties have to be balanced and cannot be nested, pattern: ${pattern}`;
-  let properties = [pattern];
+  let start = pattern.indexOf('{');
+  if (start < 0) {
+    callback( pattern.replace(END_WITH_EACH_REGEX, '.[]') );
+  } else {
+    dive('', pattern, start, callback);
+  }
+}
 
-  // Iterating backward over the pattern makes dealing with indices easier.
-  let bookmark;
-  let inside = false;
-  for (let i = pattern.length; i > 0; --i) {
-    let current = pattern[i - 1];
+function dive(prefix, pattern, start, callback) {
+  let end = pattern.indexOf('}'),
+      i = 0,
+      newStart,
+      arrayLength;
+  let tempArr = pattern.substring(start + 1, end).split(',');
+  let after = pattern.substring(end + 1);
+  prefix = prefix + pattern.substring(0, start);
 
-    switch (current) {
-      // Closing curly brace will be the first character of the brace expansion we encounter.
-      // Bookmark its index so long as we're not already inside a brace expansion.
-      case '}':
-        if (!inside) {
-          bookmark = i - 1;
-          inside = true;
-        } else {
-          assert(unbalancedNestedError, false);
-        }
-        break;
-      // Opening curly brace will be the last character of the brace expansion we encounter.
-      // Apply the brace expansion so long as we've already seen a closing curly brace.
-      case '{':
-        if (inside) {
-          let expansion = pattern.slice(i, bookmark).split(',');
-          // Iterating backward allows us to push new properties w/out affecting our "cursor".
-          for (let j = properties.length; j > 0; --j) {
-            // Extract the unexpanded property from the array.
-            let property = properties.splice(j - 1, 1)[0];
-            // Iterate over the expansion, pushing the newly formed properties onto the array.
-            for (let k = 0; k < expansion.length; ++k) {
-              properties.push(property.slice(0, i - 1) +
-                              expansion[k] +
-                              property.slice(bookmark + 1));
-            }
-          }
-          inside = false;
-        } else {
-          assert(unbalancedNestedError, false);
-        }
-        break;
+  arrayLength = tempArr.length;
+  while (i < arrayLength) {
+    newStart = after.indexOf('{');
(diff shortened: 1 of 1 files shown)


rwjblue:
Would love thoughts from @krisselden  / @stefanpenner.  

Also, pinging @nickiaconis to see if he minds reviewing as well (IIRC he worked on the current implementation).

stefanpenner:
for the `indexOf` I've noticed a simple loop + charCodeAt is often faster. May be worth trying out (but not a blocker).

```js
function charCodeIndexOf(string, charCode) {
  for (let i = 0; i < string.length; i++) {
     if (string.charCodeAt(i) === charCode) {
       return i;
     }
  }
}
```

packages/ember-metal/lib/expand_properties.js
```diff
@@ -5,7 +5,8 @@ import { assert } from 'ember-debug';
 @submodule ember-metal
 */
 
-var END_WITH_EACH_REGEX = /\.@each$/;
+var END_WITH_EACH_REGEX = /\.@each$/,
+    NESTED_BRACES = /\{[^}{]*\{|\}[^}{]*\}|\{[^}]*$/g;
```
stefanpenner: we should likely in-line this into 46, and leave a comment. That way when the assert is stripped the RegExp is aswell.

packages/ember-metal/lib/expand_properties.js
```diff
@@ -40,55 +41,34 @@ export default function expandProperties(pattern, callback) {
...
+  }
+}
 
-  // Iterating backward over the pattern makes dealing with indices easier.
-  let bookmark;
-  let inside = false;
-  for (let i = pattern.length; i > 0; --i) {
-    let current = pattern[i - 1];
+function dive(prefix, pattern, callback) {
+  let start = pattern.indexOf('{');
```
stefanpenner: At least the first time through, this `pattern.indexOf('{` has already been run on 49, could we re-use that result?

stefanpenner:
> Also, pinging @nickiaconis to see if he minds reviewing as well (IIRC he worked on the current implementation).

Ya it would be great to get @nickiaconis feedback, maybe he has some further ideas.

packages/ember-metal/lib/expand_properties.js
SpikedKira: agreed

SpikedKira:
In regards to the for/indexOf... I tried to walk the line between the optimizations of various browsers.  Chrome has overly optimized indexOf, and while for is faster in Firefox, indexOf in Firefox is still far faster than for-loop in Chrome.  I'm also seeing indexOf optimized over for-loop in Edge. (I'm missing data on Safari)

packages/ember-metal/lib/expand_properties.js
SpikedKira: yeah that optimization was for the most likely case of no-expansion.  I did look at re-using once already but I'll look again with some freshly caffeinated eyes :D

stefanpenner:
@SpikedKira interesting. In my (potentially flawed) tests `indexOf` and `for` + `charCodeAt` suggests the latter.

I would be curious if you could take a look, and ensure I am not trolling myself.

```js
var string = 'foo.{bar,baz}';

require('do-you-even-bench')([
  { name: 'charCodeAt',
    fn: function() {
      for (var i = 0; i < string.length; i ++) {
        if (string.charCodeAt(i) === 123) {
          return i;
        }
      }
      return -1;
    }
  },
  { name: 'indexOf', fn: function() { return string.indexOf('{'); } }
]);


// testing
// - charCodeAt
// - indexOf
// running first test, please wait...
//  charCodeAt ... 48,437,997.64 op/s
//  indexOf ...... 17,884,808.89 op/s

/*
npm version
{ 'perf-scratch': '1.0.0',
  npm: '4.0.5',
  ares: '1.10.1-DEV',
  cldr: '30.0.3',
  http_parser: '2.7.0',
  icu: '58.2',
  modules: '51',
  node: '7.4.0',
  openssl: '1.0.2j',
  tz: '2016j',
  unicode: '9.0',
  uv: '1.10.1',
  v8: '5.4.500.45',
  zlib: '1.2.8' }
*/
```

stefanpenner:
Ah, it may be turbofan vs crankshaft difference. As the version of node ^ this code-path his cranksaft and doesn't produce nearly as nice code as TF

stefanpenner:
TL;DR @SpikedKira I think using `indexOf` here is a good future proof thing for us to use. The only downside I could see is that ember currently does extend the string prototype which may cause some optimization sadness. cc @krisselden

SpikedKira:
@stefanpenner admittedly, had to look some of that up.  Crackshaft explains why I was seeing a much different picture when benchmarking in node vs the browsers.  Would very much welcome info on your benching setup.

stefanpenner:
@SpikedKira ya, I should write a blog post or something sometime.

> Crackshaft explains why I was seeing a much different picture when benchmarking in node vs the browsers.

I suspect you were seeing better performance in browsers?

SpikedKira:
@stefanpenner correct, Node was showing unique performance in several loops vs latest browsers

SpikedKira:
New changes based on feedback from @stefanpenner . Pretty sure I covered them all, I think it extracted a few extra ops/sec.

packages/ember-metal/lib/expand_properties.js
```diff
@@ -40,55 +40,36 @@ export default function expandProperties(pattern, callback) {
...
+    pattern.match( /\{[^}{]*\{|\}[^}{]*\}|\{[^}]*$/g ) === null
+  );
 
-  let unbalancedNestedError = `Brace expanded properties have to be balanced and cannot be nested, pattern: ${pattern}`;
-  let properties = [pattern];
+  let start = pattern.indexOf('{');
+  if (start < 0) {
+    callback( pattern.replace(END_WITH_EACH_REGEX, '.[]') );
+  } else {
+    dive("", pattern, start, callback);
```
stefanpenner: To address my ADD, let's make this single quotes instead of double quotes.

stefanpenner (changes requested):
Let one nit-pick, then I am good to :shipit:

stefanpenner (approved):

stefanpenner:
mind squashing, and prefixing the commit message with `[BUGFIX beta]` ? That way we can sneak it in sooner.

rwjblue:
Thanks @SpikedKira!

nickiaconis:
I'm late to the party, but this looks good to me. I added a test with my fix, and tests passed with these changes, so all's good. Thanks for the improvement @SpikedKira!

PR #15224, opened 2017-05-11
[BUGFIX beta] Refactor ember-debug to support better prod stripping.

rwjblue:
Prior to this change, things like `Ember.assert` and `Ember.runInDebug` were still firing in production builds. Lots of debug specific code was also left in the minified builds.

### DIFF
--- a/packages/ember-debug/lib/deprecate.js
+++ b/packages/ember-debug/lib/deprecate.js
@@ -1,4 +1,5 @@
 /*global __fail__*/
+import { DEBUG } from 'ember-env-flags';
 
 import EmberError from './error';
 import Logger from 'ember-console';
@@ -6,6 +7,11 @@ import Logger from 'ember-console';
 import { ENV } from 'ember-environment';
 
 import { registerHandler as genericRegisterHandler, invoke } from './handlers';
+
+/**
+@module ember
+@submodule ember-debug
+*/
 
 /**
   Allows for runtime registration of handler functions that override the default deprecation behavior.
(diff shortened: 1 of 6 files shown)


rwjblue:
Much easier to review this via: https://github.com/emberjs/ember.js/pull/15224/files?w=1

rwjblue:
Here is a zip file of the assets (debug + prod + min) if folks want to review those too:

[pr-15224-dist.zip](https://github.com/emberjs/ember.js/files/991631/pr-15224-dist.zip)

packages/ember-debug/lib/deprecate.js
```diff
@@ -42,150 +48,159 @@ import { registerHandler as genericRegisterHandler, invoke } from './handlers';
...
+      warning will be removed.
+    @param {String} [options.url] An optional url to the transition guide on the
+      emberjs.com website.
+    @for Ember
+    @public
+    @since 1.0.0
+  */
+  deprecate = function deprecate(message, test, options) {
+    if (!options || (!options.id && !options.until)) {
+      deprecate(
```
chadhietala: Doesn't this cause a stack overflow?

rwjblue: I had a similar thought, but no because we are passing `options` and `options.id` and `options.until`.  If we remove the guarding `if` (and move those conditions to the `deprecate`'s predicate) it would be a stack overflow.

lukemelia:
@rwjblue One of our internal addons just started failing in beta and canary because `Ember.deprecate` is undefined. Maybe related to this change?

rwjblue:
Confirm. Thank you for spotting and reporting so quickly.

Turbo87:
@rwjblue I've hit the same issue in `ember-cli-shims` just now

PR #15226, opened 2017-05-11
small cleanup in container

bekzod:


packages/container/lib/container.js
```diff
@@ -113,7 +113,7 @@ Container.prototype = {
    @param {String} fullName optional key to reset; if missing, resets everything
    */
   reset(fullName) {
-    if (arguments.length > 0) {
+    if (fullName !== undefined) {
```
stefanpenner: we prefer arity for such things, as accidentally passing undefined around is quite common.

stefanpenner (changes requested):

packages/container/lib/container.js
bekzod: gotcha, thank you

stefanpenner (approved):

stefanpenner:
LGTM

PR #15229, opened 2017-05-11
[BUGFIX beta] CoreView#trigger: type-check method before invoking it

### LINKED ISSUE #15228: CoreView#trigger blindly expects a property to be a function, if it is named after the event
[ember-views/lib/views/core_view.js#L62-L81](https://github.com/emberjs/ember.js/blob/v2.12.0/packages/ember-views/lib/views/core_view.js#L62-L81).

```js
  /**
    Override the default event firing from `Ember.Evented` to
    also call methods with the given name.
    @method trigger
    @param name {String}
    @private
  */
  trigger() {
    this._super(...arguments);
    let name = arguments[0];
    let method = this[name];
    if (method) {
      let args = new Array(arguments.length - 1);
      for (let i = 1; i < arguments.length; i++) {
        args[i - 1] = arguments[i];
      }
      return method.apply(this, args);
    }
  }
```

If there is a truthy property on a component that has the same name as an event that is triggered, this method throws the following error:

```
TypeError: method.apply is not a function
```

Proposed solution: type-check that `method` actually is a function.

```diff
  /**
    Override the default event firing from `Ember.Evented` to
    also call methods with the given name.
    @method trigger
    @param name {String}
    @private
  */
  trigger() {
    this._super(...arguments);
    let name = arguments[0];
    let method = this[name];
-   if (method) {
+   if (typeof method === 'function') {
      let args = new Array(arguments.length - 1);
      for (let i = 1; i < arguments.length; i++) {
        args[i - 1] = arguments[i];
      }
      return method.apply(this, args);
    }
  }
```

Would you like a PR?

buschtoens:
Fixes #15228 

I was gonna add a test, but there are no tests for the `ember-views` package so far. :sweat_smile:

### DIFF
--- a/packages/ember-glimmer/tests/integration/components/curly-components-test.js
+++ b/packages/ember-glimmer/tests/integration/components/curly-components-test.js
@@ -2995,6 +2995,38 @@ moduleFor('Components test: curly components', class extends RenderingTest {
     this.runTask(() => this.$('button').click());
   }
 
+  ['@test triggering an event only attempts to invoke an identically named method, if it actually is a function (GH#15228)'](assert) {
+    assert.expect(3);
+
+    let payload = ['arbitrary', 'event', 'data'];
+
+    this.registerComponent('evented-component', {
+      ComponentClass: Component.extend({
+        someTruthyProperty: true,
+
+        init() {
+          this._super(...arguments);
+          this.trigger('someMethod', ...payload);
+          this.trigger('someTruthyProperty', ...payload);
+        },
+
+        someMethod(...data) {
+          assert.deepEqual(data, payload, 'the method `someMethod` should be called, when `someMethod` is triggered');
+        },
+
+        listenerForSomeMethod: on('someMethod', function(...data) {
+          assert.deepEqual(data, payload, 'the listener `listenerForSomeMethod` should be called, when `someMethod` is triggered');
+        }),
+
+        listenerForSomeTruthyProperty: on('someTruthyProperty', function(...data) {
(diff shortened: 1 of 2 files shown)


rwjblue:
The tests for this are in ember-glimmer package.  You can add a test similar to this one https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/tests/integration/components/curly-components-test.js#L2970-L2996 (but modified for your scenario).

buschtoens:
Thanks!

Should I amend the test to the `[BUGFIX beta]` commit or should I make a separate commit?  
In the latter case, is there a preferred commit message format?

The contribution guide is not specific about that.

rwjblue:
I would amend. Basically, keep each commit as a "standalone unit" (which IMHO means the test goes with the functionality change).

buschtoens:
I added a test similar to the one you have linked. The test asserts, that

1. the method `someMethod` should be called, when `someMethod` is triggered
2. the listener `listenerForSomeMethod` should be called, when `someMethod` is triggered
3. the listener `listenerForSomeTruthyProperty` should be called, when `someTruthyProperty` is triggered

And, most importantly, it *implicitly* checks that there is no

```
TypeError: method.apply is not a function
```

as that error would bubble up and kill the test.

I could not think of a way to explicitly test this behavior.

buschtoens:
Sorry re-commiting, I  screwed up the commit message and overstepped the maximum commit message length. :see_no_evil:

stefanpenner:
noticed https://github.com/emberjs/ember.js/pull/15232 by reading this diff :)

PR #15231, opened 2017-05-11
use rsvp helper for resolved promise

bekzod:


packages/ember-application/lib/system/engine-instance.js
```diff
@@ -75,7 +75,7 @@ const EngineInstance = EmberObject.extend(RegistryProxyMixin, ContainerProxyMixi
   boot(options) {
     if (this._bootPromise) { return this._bootPromise; }
 
-    this._bootPromise = new RSVP.Promise(resolve => resolve(this._bootSync(options)));
+    this._bootPromise = RSVP.resolve(this._bootSync(options));
```
stefanpenner: this is different, as now if `bootSync` now throws, `boot` will throw. But previously the error would be correctly propagated via `this._bootPromise`

stefanpenner (changes requested):

packages/ember-application/lib/system/engine-instance.js
bekzod: ok, thanks

PR #15247, opened 2017-05-17
[BUGFIX release] Ensure nested custom elements render properly.

### LINKED ISSUE #15221: 2.13 custom elements regression
https://github.com/glimmerjs/glimmer-vm/issues/453

rwjblue:
A bug was introduced with the Glimmer update that was included in Ember 2.13 that caused an error to be thrown when using nested custom elements like:

```hbs
<foo-bar>
  <baz-qux></baz-qux>
</foo-bar>
```

The fix was done upstream in https://github.com/glimmerjs/glimmer-vm/pull/488.

Fixes https://github.com/emberjs/ember.js/issues/15221

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,11 +39,11 @@
     "link:glimmer": "node bin/yarn-link-glimmer.js"
   },
   "dependencies": {
-    "@glimmer/compiler": "^0.22.0",
-    "@glimmer/node": "^0.22.0",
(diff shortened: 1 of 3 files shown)


bjornharrtell:
Still having problems in what appears related to this. See https://github.com/miguelcobain/ember-paper/pull/709 and https://github.com/glimmerjs/glimmer-vm/issues/453.

PR #15252, opened 2017-05-18
[DOC release] Improve documentation of `if` template helper

locks:
Fixes https://github.com/emberjs/guides/issues/1737.

### DIFF
--- a/packages/ember-glimmer/lib/helpers/if-unless.js
+++ b/packages/ember-glimmer/lib/helpers/if-unless.js
@@ -16,64 +16,6 @@ import {
   isConst
 } from '@glimmer/reference';
 
-
-/**
-  Use the `if` block helper to conditionally render a block depending on a
-  property. If the property is "falsey", for example: `false`, `undefined`,
-  `null`, `""`, `0`, `NaN` or an empty array, the block will not be rendered.
-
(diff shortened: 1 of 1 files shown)


rwjblue (commented):
We need to figure out what happens here, because from my perspective it looks like there are two things defined the same way (inline if and normal block form).  I would think that only one would win and clobber the other?  This comment only applies to inline if...

locks:
@rwjblue can I haz re-review?

packages/ember-glimmer/lib/helpers/if-unless.js
```diff
@@ -16,11 +16,55 @@ import {
...
+  }
+}
 
 /**
-  Use the `if` block helper to conditionally render a block depending on a
-  property. If the property is "falsey", for example: `false`, `undefined`,
-  `null`, `""`, `0`, `NaN` or an empty array, the block will not be rendered.
+  The `if` helper allows you to conditionally render one of two branches,
+  depending on the "truthiness" of a property.
+  For example the following values are all also: `false`, `undefined`, `null`, `""`, `0`, `NaN` or an empty array.
```
kumkanillam: for me something is missing here

PR #15253, opened 2017-05-19
[CLEANUP] Output correct `console.log` for `bin/run-tests`

alexdiliberto:
Super small update here. This just applies for the 'travis-browsers' case.


rwjblue:
Awesome, thank you!

PR #15254, opened 2017-05-19
[FEATURE glimmer-custom-component-manager] Custom Component Manager

twokul:
This is the first step towards opening up new APIs for Ember.js community as per
[Custom Components RFC](https://github.com/emberjs/rfcs/blob/custom-components/text/0000-custom-components.md).

This will allow the following:

+ add-on authors can now start experimenting with Glimmer API while continue to support existing Ember.js components
+ developers will be able to _opt into_ using custom component managers

This change should be considered cutting-edge and introduced behind a feature flag `GLIMMER_CUSTOM_COMPONENT_MANAGER`.

To enable this feature flag in `config/environment.js` of your Ember application:

```js
module.exports = function(environment) {
  let ENV = {
    ...
    EmberENV: {
      FEATURES: {
        'glimmer-custom-component-manager': true
      }
    }
  };

  ...

  return ENV;
};
```

Or in Glimmer application, `config/environment.js`:

```js
module.exports = function(environment) {
  let ENV = {
    ...

    GlimmerENV: {
      FEATURES: {
        'glimmer-custom-component-manager': true
      }
    }
  };

  return ENV;
};
```

After that, you should be able to opt into registering your own component managers:

+ for applications

App with classic structure:

```
my-app/
  component-managers/
    nyan.js
  components/
  templates/
    components/
      nyan-cat.hbs
```

```hbs
{{!- tempates/components/nyan-cat.hbs -}}
{{use-component-manager "nyan"}}

<h1>🐱</h1>
```

App with Module Unification:

```
my-app/
  src/
    ui/
      component-managers/
        nyan.js
      components/
        nyan-cat/
          template.hbs
```

```hbs
{{!- src/ui/components/nyan-cat.hbs -}}
{{use-component-manager "nyan"}}

<h1>🐱</h1>
```

+ for add-ons

```
nyan-cat-addon/
  addon/
    component-managers/
      glimmer.js
  app/
    templates/
      components/
        nyan-cat.hbs
```

### DIFF
--- a/packages/ember-glimmer/lib/environment.js
+++ b/packages/ember-glimmer/lib/environment.js
@@ -53,6 +53,10 @@ import { default as htmlSafeHelper } from './helpers/-html-safe';
 import installPlatformSpecificProtocolForURL from './protocol-for-url';
 import { default as ActionModifierManager } from './modifiers/action';
 
+import {
+  GLIMMER_CUSTOM_COMPONENT_MANAGER
+} from 'ember/features';
+
 function instrumentationPayload(name) {
   return { object: `component:${name}` };
 }
@@ -74,9 +78,17 @@ export default class Environment extends GlimmerEnvironment {
 
     this._definitionCache = new Cache(2000, ({ name, source, owner }) => {
       let { component: componentFactory, layout } = lookupComponent(owner, name, { source });
+      let customManager = undefined;
 
       if (componentFactory || layout) {
-        return new CurlyComponentDefinition(name, componentFactory, layout);
+        if (GLIMMER_CUSTOM_COMPONENT_MANAGER) {
+          let managerId = layout && layout.meta.managerId;
+
+          if (managerId) {
+            customManager = owner.factoryFor(`component-manager:${managerId}`).class;
+          }
+        }
+        return new CurlyComponentDefinition(name, componentFactory, layout, undefined, customManager);
       }
     }, ({ name, source, owner }) => {
       let expandedName = source && owner._resolveLocalLookupName(name, source) || name;
--- a/packages/ember-glimmer/tests/integration/custom-component-manager-test.js
+++ b/packages/ember-glimmer/tests/integration/custom-component-manager-test.js
@@ -1,0 +1,83 @@
+import {
+  compileLayout,
+  PrimitiveReference
+} from '@glimmer/runtime';
+import { moduleFor, RenderingTest } from '../utils/test-case';
+import {
+  EMBER_GLIMMER_ALLOW_BACKTRACKING_RERENDER,
+  GLIMMER_CUSTOM_COMPONENT_MANAGER,
+  MANDATORY_SETTER
+} from 'ember/features';
+
+import ExperimentalComponentManager from 'ember-glimmer/component-managers/curly'
+
+if (GLIMMER_CUSTOM_COMPONENT_MANAGER) {
+  /*
+    Custom layout compiler. Exists mostly to inject
+    class and attributes for testing purposes.
+  */
+  class TestLayoutCompiler {
+    constructor(template) {
+      this.template = template;
+    }
+
+    compile(builder) {
+      builder.wrapLayout(this.template.asLayout());
+      builder.tag.dynamic(() => {
+        return PrimitiveReference.create('p');
+      });
+      builder.attrs.static('class', 'hey-oh-lets-go');
+      builder.attrs.static('manager-id', 'test');
+    }
+  }
+
+  /*
+    Implementation of custom component manager, `ComponentManager` interface
+  */
+  class TestComponentManager {
+    prepareArgs(definition, args) { return args; }
+
+    create(env, definition, args, dynamicScope, caller, hasBlock) {
+      return definition.ComponentClass.create();
+    }
+
+    layoutFor(definition, bucket, env) {
+      return env.getCompiledBlock(TestLayoutCompiler, definition.template);
+    }
+
+    getSelf({ component }) { return component; }
+
+    didCreateElement(component, element, operations) { }
+
+    didRenderLayout(component, bounds) { }
+
+    didCreate(component) { }
+
+    getTag({ component }) { return null; }
+
+    update(component, dynamicScope) { }
+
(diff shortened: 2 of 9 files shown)


packages/ember-glimmer/lib/environment.js
```diff
@@ -74,9 +78,17 @@ export default class Environment extends GlimmerEnvironment {
...
       let { component: componentFactory, layout } = lookupComponent(owner, name, { source });
+      let customManager = undefined;
 
       if (componentFactory || layout) {
-        return new CurlyComponentDefinition(name, componentFactory, layout);
+        if (GLIMMER_CUSTOM_COMPONENT_MANAGER) {
+          let managerId = layout && layout.meta.managerId;
+
+          if (managerId) {
+            customManager = owner.resolveRegistration(`component-manager:${managerId}`);
```
rwjblue: I think in general we want either `.lookup` or `.factoryFor().class` instead of `resolveRegistration`.

packages/ember-glimmer/lib/index.js
```diff
@@ -286,3 +286,4 @@ export {
 export { setupEngineRegistry, setupApplicationRegistry } from './setup-registry';
 export { DOMChanges, NodeDOMTreeConstruction, DOMTreeConstruction } from './dom';
 export { registerMacros as _registerMacros, experimentalMacros as _experimentalMacros } from './syntax';
+export { default as ExperimentalComponentManager } from './syntax/experimental-component-manager';
```
rwjblue: Not sure we need to export this, I don't think that we want folks extending from our curly component manager. We want to allow them to create their own (which is basically "just" implement XYZ hooks and go to town).

packages/ember-glimmer/lib/syntax/experimental-component-manager.js
```diff
@@ -0,0 +1,308 @@
+import { OWNER } from 'ember-utils';
```
rwjblue: Not sure about this file naming.  I'd prefer `packages/ember-glimmer/lib/component-managers/curly.js`.

packages/ember-glimmer/tests/integration/custom-component-manager-test.js
```diff
@@ -0,0 +1,27 @@
...
+  MANDATORY_SETTER
+} from 'ember/features';
+
+if (GLIMMER_CUSTOM_COMPONENT_MANAGER) {
+  moduleFor('Components test: curly components with custom manager', class extends RenderingTest {
+    ['@test it can render a basic component with custom component manager']() {
+      this.owner.register('component-manager:glimmer', new TestComponentManager());
+      this.registerComponent('foo-bar', {
+        template: '{{use-component-manager "glimmer"}}hello',
+        managerId: 'glimmer'
```
rwjblue: We probably shouldn't use `glimmer` here, because we actually intend to use that managerId in the future and it may confuse future readers of the test about what we are talking...

packages/ember-glimmer/tests/integration/custom-component-manager-test.js
```diff
@@ -0,0 +1,27 @@
+import { moduleFor, RenderingTest } from '../utils/test-case';
+import TestComponentManager from './helpers/test-component-manager';
```
rwjblue: I'd prefer to just implement a super simple manager in this file directly, for now we don't have a need for it across many files...

rwjblue: We should not attempt to extend from the curly manager, but create a new one that satisifies [this interface](https://github.com/glimmerjs/glimmer-vm/blob/9fb4cd315589186ce7179bed4015d56764001ee7/packages/%40glimmer/runtime/lib/component/interfaces.ts#L21-L86).

packages/ember-glimmer/tests/integration/custom-component-manager-test.js
```diff
@@ -0,0 +1,27 @@
...
+    ['@test it can render a basic component with custom component manager']() {
+      this.owner.register('component-manager:glimmer', new TestComponentManager());
+      this.registerComponent('foo-bar', {
+        template: '{{use-component-manager "glimmer"}}hello',
+        managerId: 'glimmer'
+      });
+
+      this.render('{{foo-bar}}');
+
+      this.assertComponentElement(this.firstChild, { content: 'hello' });
```
rwjblue: These assertions should somehow ensure that the correct manager was used.  Perhaps the manager could add a special attribute to the element (for example) that you could assert is present?

packages/ember-template-compiler/lib/plugins/extract-pragma-tag.js
```diff
@@ -0,0 +1,26 @@
+const PRAGMA_TAG = 'use-component-manager';
```
rwjblue: How does this play with https://github.com/ember-cli/ember-cli-htmlbars/pull/114?  It seems like only one or the other should be required, not both.

packages/ember-template-compiler/lib/system/compile-options.js
```diff
@@ -10,6 +10,7 @@ export default function compileOptions(_options) {
   if (options.moduleName) {
     let meta = options.meta;
     meta.moduleName = options.moduleName;
+    meta.managerId = options.managerId;
```
rwjblue: This block is only meant as a fallback for older versions of ember-cli-htmlbars which passed `{ moduleName: 'here' }` instead of `{ meta: { moduleName: 'here' } }`.  No need to put managerId here like this.

packages/ember-template-compiler/lib/plugins/extract-pragma-tag.js
twokul: I think it would make more sense to leave AST transform here.

rwjblue (approved):

rwjblue:
LGTM, we can land once green. Can you comment on the RFC PR mentioning the initial implementation here?

rwjblue:
Thanks for working on this @twokul!

twokul:
@rwjblue sure thing. thanks for reviewing!

PR #15262, opened 2017-05-20
Use https in references to emberjs website

ahmadsoe:
Now emberjs website is served via https. emberjs/website#2828


locks:
Appreciated :)

PR #15263, opened 2017-05-21
cleanup transitioningIn/transitioningOut

bekzod:


rwjblue (approved):

packages/ember-glimmer/lib/components/link-to.js
```diff
@@ -594,17 +594,19 @@ const LinkComponent = EmberComponent.extend({
   }),
 
   transitioningIn: computed('active', 'willBeActive', function computeLinkToComponentTransitioningIn() {
-    let willBeActive = get(this, 'willBeActive');
-    if (typeof willBeActive === 'undefined') { return false; }
-
-    return !get(this, 'active') && willBeActive && 'ember-transitioning-in';
+    if (get(this, 'willBeActive') === true && !get(this, 'active')) {
```
locks: I'm wondering why the mix of comparing to a boolean on the first clause, but using `!` on the second one?

bekzod: they might return
`get(this, 'active')` = > false, undefined, string
`get(this, 'willBeActive')` => true, false, undenfined (I am removing this case)

bekzod: I think I can check for explicit `false`, let me test

bekzod: didn't work `!get(this, 'active')` is correct way for now

Issue #10195, opened 2015-01-12
{{link-to}} classname `ember-transitioning-in` stuck


zeppelin:
When:

- ember-canary
- The link is pointing to the same route
- `active=false` is set

Repro:

http://emberjs.jsbin.com/guvugalopa/1/edit?html,css,js,output

- Click link `Foo`
- Page transitioned to `Foo route`
- Click link `Foo` again
- Link `Foo`'s background is red (see CSS tab)

`ember-transitioning-in` is removed after transitioning to another route, yet the correct behavior should be to not to show this classname at all, if already on the target route.

DerToti:
This seems stll to be a problem in ember 1.11 beta version.
Is there any workaround for this or is it possible to disable this feature?

machty:
@DerToti a fix is coming shortly, but since all this feature does is apply a CSS class name, you can "disable" the feature by not supplying CSS for `.ember-transitioning-X`.

rwjblue:
Closing in favor of https://github.com/emberjs/ember.js/issues/13164.

PR #15265, opened 2017-05-21
[BUGFIX] fixed issue when passing `false` to `activeClass` would break things

bekzod:
this might have caused issues when `activeClass` falsy value  would make `ember-transitioning-in` and `ember-transitioning-out` incorrectly get appended


bekzod:
possibly fixes #10195 if that still a issue

rwjblue:
Changes seem good, but I think this needs a rebase now.

locks:
@bekzod can you squash the commits, please?

bekzod:
@locks done :P

PR #15269, opened 2017-05-22
wrapped `_popstateHandler` with `run.bind`

bekzod:
I am not sure why it was not wrapped in run, as hashLocation same callback is wrapped already https://github.com/emberjs/ember.js/blob/master/packages/ember-routing/lib/location/hash_location.js#L114

### DIFF
--- a/packages/@ember/-internals/routing/lib/location/history_location.js
+++ b/packages/@ember/-internals/routing/lib/location/history_location.js
@@ -1,13 +1,11 @@
 import { get, set } from '@ember/-internals/metal';
-
+import { bind } from '@ember/runloop';
 import { Object as EmberObject } from '@ember/-internals/runtime';
 import EmberLocation from './api';
 
 /**
 @module @ember/routing
 */
-
-let popstateFired = false;
 
 function _uuid() {
   return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function(c) {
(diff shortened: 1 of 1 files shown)


packages/ember-routing/lib/location/history_location.js
```diff
@@ -208,14 +207,14 @@ export default EmberObject.extend({
...
-    this._popstateHandler = () => {
+    this._popstateHandler = run.bind(this, function() {
       // Ignore initial page load popstate event in Chrome
-      if (!popstateFired) {
-        popstateFired = true;
+      if (!this.__popstateFired) {
+        this.__popstateFired = true;
         if (this.getURL() === this._previousURL) { return; }
       }
       callback(this.getURL());
```
stefanpenner: is it possible `callback` contains the run? If so, we may need to cleanup that side.

bekzod: I don't think `callback` contains run it starts here https://github.com/emberjs/ember.js/blob/master/packages/ember-routing/lib/system/router.js#L224 and calls various methods in router

stefanpenner: I'm surprised this (prior to this PR) doesn't trigger the runloops auto-run assertion, causing peoples tests to fail.

We likely need to figure out why, something smells fishy.

stefanpenner (changes requested):

packages/ember-routing/lib/location/history_location.js
krisselden: tests generally don't navigate with the url right?

rwjblue: Confirm.

auvipy (changes requested):
can u plz fix the conflicts?

locks:
Closing due to no longer being relevant. Thanks for the work!

PR #15271, opened 2017-05-23
`render` method cleanup

bekzod:
made it less confusing


stefanpenner:
looks like sauce failed, re-running it

PR #15272, opened 2017-05-24
cleanup  `route/render` 

bekzod:
will cleanup it up further once #15271 merged


bekzod:
it is odd when outlet is not passed to `render` and `disconnectOutlet` it uses `main` but it throws error at same time, it either should throw error and not use `main` as default or it should use `main` as default but warn instead of throw

PR #15276, opened 2017-05-24
Test to explore reasonable ways to handle UnrecognizedURLError on boot

stefanpenner:
There may be a more expected solution to this, but I am unaware of that. I suspect we should decide on a reasonable "pit of success" way of handling this.

### What is this?

when visiting an unrecognized URL on boot, we correctly throw and error. But the regular route related error handling is not utilized.... Leaving the user to fend for themselves in pretty bespoke ways.

One way to potential handle this is (although not without issue) could be:

```js
Ember.onerror = function(error) {
  if (error !== null &&
     typeof error === 'object' &&
     error.name === 'UnrecognizedURLError') {

    let url = error.message;
    // display 404 page
  } else {
    throw error; // or w/e we normally do
  }
};


```

### DIFF
--- a/packages/ember/tests/routing/basic_test.js
+++ b/packages/ember/tests/routing/basic_test.js
@@ -1878,15 +1878,14 @@ QUnit.test('Parent route context change', function() {
 QUnit.test('Router accounts for rootURL on page load when using history location', function() {
   let rootURL = window.location.pathname + '/app';
   let postsTemplateRendered = false;
-  let setHistory, HistoryTestLocation;
 
-  setHistory = function(obj, path) {
-    obj.set('history', { state: { path: path } });
-  };
+  function setHistory(obj, path) {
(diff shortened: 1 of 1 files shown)


rwjblue:
Hmm, the most common way to do custom 404 pages is with a glob route.  At which point there isn't really a `UnrecognizedURLError` thrown, you can just handle the glob route being hit and do whatever is needed there.

stefanpenner:
> Hmm, the most common way to do custom 404 pages is with a glob route. At which point there isn't really a UnrecognizedURLError thrown, you can just handle the glob route being hit and do whatever is needed there.

Ah that's likely a much better approach, are there downsides you can think of?

stefanpenner:
@rwjblue I suspect we should just close this then. I totally brain-farted on globed routing.

PR #15292, opened 2017-05-26
removed `namePassed` flag

bekzod:
since `typeof _name === 'string' && !!_name ` is same as `!isEmpty(_name)` ( in the context it used at least) `namePassed`  could be replaced with `!isDefaultRender`


rwjblue:
Nice cleanup, thank you!

PR #15294, opened 2017-05-27
micro optimizations in core-object

bekzod:


rwjblue:
restarted sauce labs job

PR #15305, opened 2017-05-30
[BUGFIX beta] private registry entries should have names that don’t n…

stefanpenner:
…eed to be normalized

Previously they contained a `.`

### DIFF
--- a/packages/container/lib/registry.js
+++ b/packages/container/lib/registry.js
@@ -861,7 +861,7 @@ function has(registry, fullName, source) {
 }
 
 const privateNames = dictionary(null);
-const privateSuffix = `${Math.random()}${Date.now()}`;
+const privateSuffix = `${Math.random()}${Date.now()}`.replace('.', '');
 
 export function privatize([fullName]) {
   let name = privateNames[fullName];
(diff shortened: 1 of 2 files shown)


stefanpenner:
@rwjblue are you OK with this also getting the next release?

packages/container/lib/registry.js
```diff
@@ -861,7 +861,7 @@ function has(registry, fullName, source) {
 }
 
 const privateNames = dictionary(null);
-const privateSuffix = `${Math.random()}${Date.now()}`;
+const privateSuffix = `${Math.random()}${Date.now()}`.replace('.', '');
```
rwjblue: Seems fine, but I'm surprised that we don't consider `Date.now()` to be private enough....

rwjblue (commented):
No real objection (seems safe enough), it just doesn't seem like a bug fix.

packages/container/lib/registry.js
stefanpenner: super secure

stefanpenner:
> No real objection (seems safe enough), it just doesn't seem like a bug fix.

Ya, its just causing some annoyances with stricter resolving, since the `dot` shows up.

PR #15308, opened 2017-05-31
Add EventDispatcher tests for event bubbling

simonihmig:
This adds some missing tests for asserting that events bubble up the view hierarchy, and being able to prevent this.

@rwjblue These should be the tests (they were missing) you mentioned in https://github.com/rwjblue/ember-native-dom-event-dispatcher/pull/10#issuecomment-303771206


rwjblue:
Thank you!!!!

Issue #15290, opened 2017-05-26
runInDebug may be leaking when high-iteration each helpers used


DLiblik:
We're running in 2.13.1 and we have a component template that has a pair of nested each blocks, with volumes of 53 and about 30 respectively, resulting in about 1500 iterations.

There are quite a few helpers in the markup, so the helpers themselves are being called a lot (for example, the `if` helper is being invoked about  10 times per iteration, or 14,000 times).

When we run in development mode, leaving the route and returning takes longer and longer, linearly (about 50-100ms each time) slowly building up to 2s, 3s, etc.  It does not seem to matter what helpers are in use (and we are not using any custom helpers - btw it occurs even if we remove all helpers except for the two `each` helpers themselves, and even if we remove all HTML, so that only the `each` helpers are looping but doing nothing else within).

Using Chrome Profiler reveals that the culprit is a call stack that starts with `runInDebug` and then drills to `compute`, `value`, `compute`, `compute`, `value`, `concat`, `compute`, `value`... etc.

In production, there is no such issue (which makes sense - runInDebug is a NOP there).

Here is the profile (note that all code is in ember.debug.js or vendor.js but seems to be ember code).

![image](https://cloud.githubusercontent.com/assets/4047997/26510673/6bd330c0-422c-11e7-8d81-14d577d7f17c.png)

I tried to debug into this but it's above my pay-grade w.r.t. ember :blush:

rwjblue:
@DLiblik - Thanks for digging into this!  It would be really helpful if you could create a reproduction demo of this so we can help figure out what is going on.  I think either an ember-twiddle or full blown ember-cli app would work very well to demo the issue...

stefanpenner:
Yup great sleuthing. With a reproduction we can dive in and help figure out whatsup asap.

DLiblik:
Ok let me see what I can put together - thanks.

DLiblik:
Here is a test project, and an update: 

The issue occurs on any high volume of custom helpers being rendered while in debug mode. 

In the attached project, a simple loop repeats a helper that returns a fixed string, 2,500 times.  You can flip back and forth in and out of the route using a provided link, and the console log will show the ever-increasing render time each time you leave and return to the route:

![image](https://cloud.githubusercontent.com/assets/4047997/26646391/64671050-4609-11e7-9d5f-20ac557c87ab.png)

If you remove the `my-helper` helper from the `big-list` component template, then the problem goes away!

[...]

PR #15312, opened 2017-05-31
[BUGFIX release] Avoid re-freezing already frozen objects.

rwjblue:
When a helper is invoked we pass in the arguments, those arguments are frozen (via Object.freeze) so that helpers can't mutate them (and cause issues in the rendering engine itself). When a helper doesn't have hash arguments, we use a shared `EMPTY_HASH` empty object to avoid allocating a bunch of empty objects for no reason (we do roughly the same thing when no positional params are passed). Since these objects are shared they are being frozen over and over again (throughout the lifetime of the running application). Turns out that there is what we think is almost certainly a bug in Chrome, where re-freezing the same object many times starts taking significantly more time upon each freeze attempt.

This change introduces a guard to ensure we do not re-freeze repeatedly.

Thanks to @krisselden for identifying the root cause.

Fixes https://github.com/emberjs/ember.js/issues/15290

### DIFF
--- a/packages/ember-glimmer/lib/utils/references.js
+++ b/packages/ember-glimmer/lib/utils/references.js
@@ -38,6 +38,22 @@ export const UPDATE = symbol('UPDATE');
 
 export { NULL_REFERENCE, UNDEFINED_REFERENCE } from '@glimmer/runtime';
 
+let maybeFreeze;
+if (DEBUG) {
+  // gaurding this in a DEBUG gaurd (as well as all invocations)
+  // so that it is properly stripped during the minification's
+  // dead code elimination
+  maybeFreeze = (obj) => {
+    // re-freezing an already frozen object introduces a significant
+    // performance penalty on Chrome (tested through 59).
+    //
+    // See: https://bugs.chromium.org/p/v8/issues/detail?id=6450
+    if (!Object.isFrozen(obj) && HAS_NATIVE_WEAKMAP) {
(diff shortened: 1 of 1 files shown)


stefanpenner (approved):

krisselden:
https://bugs.chromium.org/p/v8/issues/detail?id=6450

packages/ember-glimmer/lib/utils/references.js
```diff
@@ -38,6 +38,19 @@ export const UPDATE = symbol('UPDATE');
 
 export { NULL_REFERENCE, UNDEFINED_REFERENCE } from '@glimmer/runtime';
 
+let maybeFreeze = () => {};
+if (DEBUG) {
+  maybeFreeze = (obj) => {
```
mixonic: Can you add a note here regarding intended usage? I notice you're left the `if (DEBUG) {` statements even though the function is a noop when `!DEBUG` and I presume it is for DCE to do its thing. A note here might help a future traveller from thinking they can rely on the noop in non-debug.

rwjblue: Sounds good. And yes, that is why I left the if guards.

DLiblik:
@krisselden - thanks for tracing down the root cause - when I posted #15290 I didn't specify which browser we were on - didn't occur to me that the browser itself was the issue!

PR #15314, opened 2017-06-01
[BUGFIX beta] Avoid error while parsing array QP

### LINKED ISSUE #15306: Ember crashes after encountering unexpected query param type
When a controller tries to deserialize an array based query param it can fail if the query param is not properly stringified, causing Ember to completely crash. This can occur when a user tries to edit the query param directly. It may be better to gracefully handle cases like this to keep the app alive since a simple refresh will not fix the issue. 

https://ember-twiddle.com/ea932831eba4f75f641f1a91f200712e?fileTreeShown=false&openFiles=controllers.index.js%2C

Change URL field to `/?arrayQueryParam=test`

Serabe:
While deserializing an array query parameter, `JSON.parse` is used. If
it throws an error, the application cannot be easily recovered.

This array catches that exception, returns an empty array and shows
information about the error.

Fixes #15306

### DIFF
--- a/packages/ember-routing/lib/system/router.js
+++ b/packages/ember-routing/lib/system/router.js
@@ -17,7 +17,8 @@ import {
   Error as EmberError,
   deprecate,
   assert,
-  info
+  info,
+  warn
 } from 'ember-debug';
 import {
   Object as EmberObject,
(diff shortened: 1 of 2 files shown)


Serabe:
Not sure info is the best way. Open to suggestions.

Serabe:
Problem with yarn. Restarting one of the jobs to see if it persists.

stefanpenner:
I'm nervous to hide errors:

* Random logging can be easily missed, and wont be captured by bug trackers
* If the QP is wrong, empty may cause further issues.

That being said, we likely need an appropriate way to surface the errors. Can we transition to an error route?

Serabe:
I can change this to error tomorrow :)

Serabe:
Closing this.

Issue #13216, opened 2016-03-30
Testing a Component using _ProxyMixin produces autorun error.


bdiz:
Trying to use a component that mixes in _ProxyMixin (ember v2.4.1). I get the autorun error during integration testing.

```
import Ember from 'ember';

export default Ember.Component.extend(Ember._ProxyMixin, {
});
```

Then in my test:
```
moduleForComponent('my-component', 'Integration | Component | my component', {
  integration: true
});

test('it responds to changes in content object', function(assert) {
  this.set('myContent', Ember.Object.create({myProperty: true}));
  this.render(hbs`{{my-component content=myContent }}`);
  this.get('myContent').set('myProperty', false); // Autorun error happens when setting a property on content
  assert.ok(...);
});
```

This results in the autorun error: `Error: Assertion Failed: You have turned on testing mode, which disabled the run-loop's autorun. You will need to wrap any code with asynchronous side-effects in a run`

I found when I add a run-loop to `lib/mixins/-proxy.js` the error goes away:

```
import run from 'ember-metal/run_loop';

function contentPropertyDidChange(content, contentKey) {
  var key = contentKey.slice(8); // remove "content."
  if (key in this) { return; } // if shadowed in proxy
  run(() => propertyDidChange(this, key)); // Added run-loop here.
}
```

I don't know enough about the run-loop and ember internals to know if this has any implications beyond being an annoyance in testing, but I thought I'd report it just in case.

I can wrap setting of `content` properties in a run-loop as a workaround.

```
Ember.run(() => this.get('myContent').set('myProperty', false));
```

Can close if this is small fry. I know it's a bit obscure. :) Thanks.

pixelhandler:
@bdiz I'm curious have you discussed this in the testing channel of the embercommunity.slack.com chatroom ?

Usually added an Ember.run solves the testing error.

However I'm a bit hesitant when relying on private APIs, see https://github.com/emberjs/ember.js/blob/v2.4.4/packages/ember-runtime/lib/mixins/-proxy.js#L36-L44 

Would using ArrayProxy or ObjectProxy solve the issue?

bdiz:
Hi @pixelhandler, no I didn't know about the slack channel, but thanks for letting me know about it. 

Unfortunately ArrayProxy and ObjectProxy won't help me as I need a "ComponentProxy" as it were.  

There doesn't seem to be an explanation as to why _ProxyMixin was made private when it was originally public.

rwjblue:
[...]

PR #15317, opened 2017-06-01
[BUGFIX beta] PromiseProxyMixin: wrap calls to setProperties in run

JoelWAnna:
Similar to https://github.com/emberjs/ember.js/issues/13216

Using the PromiseProxyMixin in a test will throw
"Assertion Failed: You have turned on testing mode, which disabled the run-loop's autorun. You will need to wrap any code with asynchronous side-effects in a run"

see https://ember-twiddle.com/e94c8ac10b1a5705c5a66eecd5e66fcc for example test, error in console

### DIFF
--- a/packages/ember-runtime/lib/mixins/promise_proxy.js
+++ b/packages/ember-runtime/lib/mixins/promise_proxy.js
@@ -1,5 +1,6 @@
 import {
   get,
+  run,
   setProperties,
   computed,
   Mixin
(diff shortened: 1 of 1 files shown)


packages/ember-runtime/lib/mixins/promise_proxy.js
```diff
@@ -20,17 +21,21 @@ function tap(proxy, promise) {
 
   return promise.then(value => {
     if (!proxy.isDestroyed && !proxy.isDestroying) {
-      setProperties(proxy, {
-        content: value,
-        isFulfilled: true
-      });
+      run(() => {
```
stefanpenner: A run here, will change ordering and cause nested renders. I don't believe this will work

stefanpenner (changes requested):

stefanpenner:
We recommend you use Ember.RSVP.Promise instead of the Promise global. We recommend this so that:

* cross browser support is met
* timing between embers rendering and promise resolution is ideal. When using native promises unfortunately this wont be, and will lead to a waterfall of renders and such. (Although we have plans to improve this, for the time being it is discouraged)

JoelWAnna:
In my application I are using Ember.RSVP.Promise and seeing the same assertion, looks like I'll need to do some more digging

PR #15093, opened 2017-03-28
(WIP) 15058: Decouple ember test from global resolver

ssriera:
This PR resolves issue [15058](https://github.com/emberjs/ember.js/issues/15058) using the global resolver in `reset_test.js`. 

### Problem 
Tests are failing, I'm getting this error- seems like it wants to use the global resolver?

screen shot 2017-03-27 at 7 14 06 pm

PR #15318, opened 2017-06-02
Decouple reset_test from globals resolver

mixonic:
Closes https://github.com/emberjs/ember.js/pull/15093
References https://github.com/emberjs/ember.js/issues/15058

### DIFF
--- a/packages/ember-application/tests/system/reset_test.js
+++ b/packages/ember-application/tests/system/reset_test.js
@@ -1,234 +1,142 @@
 import { run, get, set } from 'ember-metal';
-import EmberApplication from '../../system/application';
 import { Object as EmberObject, Controller } from 'ember-runtime';
 import { Router } from 'ember-routing';
-import { Registry } from 'container';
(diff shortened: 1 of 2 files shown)


packages/ember-application/tests/system/reset_test.js
```diff
@@ -1,234 +1,142 @@
...
+    this.visit('/one');
 
-  application.reset();
-
-  equal(readyCallCount, 2, 'ready was called twice');
-});
+    assert.equal(get(applicationController, 'currentPath'), 'one');
+  }
 
-QUnit.test('With ember-data like initializer and constant', function() {
```
mixonic: This test is simply removed. `DS.defaultStore` is no longer a thing. This test asserts that `init` is called on the registered factory, and that `willDestroy` is not called when `reset` is called without a runloop. This is not meaningful.

rwjblue: 🔪👍

PR #15319, opened 2017-06-02
Decouple ember link_to_test from globals resolver

mixonic:
References #15058

It is a big one. About 30% of the way through converting the file, I'd like to finish the conversion in a single commit.

### DIFF
--- a/packages/ember/tests/helpers/link_to_test.js
+++ b/packages/ember/tests/helpers/link_to_test.js
@@ -1,1459 +1,1557 @@
 import Logger from 'ember-console';
+import {
+  moduleFor,
+  ApplicationTestCase,
+  AutobootApplicationTestCase
+} from 'internal-test-helpers';
 
 import {
   Controller,
-  Object as EmberObject,
   inject,
   A as emberA
 } from 'ember-runtime';
 import {
-  set,
-  run,
   instrumentationSubscribe as subscribe,
-  instrumentationReset as reset,
   alias
 } from 'ember-metal';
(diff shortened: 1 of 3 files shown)


packages/ember/tests/helpers/link_to_test.js
```diff
@@ -18,7 +19,7 @@ import { Application } from 'ember-application';
 import { jQuery } from 'ember-views';
 import { compile } from 'ember-template-compiler';
 import { setTemplates, setTemplate } from 'ember-glimmer';
-import { EMBER_IMPROVED_INSTRUMENTATION } from 'ember-debug';
```
mixonic: Pretty sure this is just wrong, and the tests in this file controlled by this value were not running.

packages/internal-test-helpers/lib/test-cases/abstract-application.js
```diff
@@ -59,8 +59,10 @@ export default class AbstractApplicationTestCase extends AbstractTestCase {
     if (applicationInstance) {
       return run(applicationInstance, 'visit', url, options);
     } else {
-      return run(this.application, 'visit', url, options).then(instance => {
-        this.applicationInstance = instance;
```
mixonic: b/c this `.then` is not inside he runloop, `this.applicationInstance` was being assigned *after* `this.visit('/foo')`. The more predictable behavior is to wrap the `then` in a runloop so we can presume the instance is set on the next line of code after `this.visit`

PR #15325, opened 2017-06-04
[FEATURE ember-engines-mount-params] Constrain to `model` hash arg.

rwjblue:
After discussion during a recent core team meeting, the original implementation 
(and RFC) were determined to be a bit future hostile.

Specifically, with the implementation prior to this PR it would have been impossible to add any additional hash arguments that were targeting the `{{mount` syntax directly (e.g. namespacing for use with tree shaking) without breaking changes (since all hash arg names were allowed to be passed through before).

This change updates things a bit to limit the pass through to the `model` hash argument, which actually makes things a tad bit more teachable anyways. The ability to pass down arbitrary named values is absolutely still present through the usage of the `(hash` helper.


mike183:
👍

Thanks for the heads up.

knownasilya:
Sweet

PR #15337, opened 2017-06-08
Decouple acceptance_test, helpers_test frm globals

mixonic:
1) Decouple the `acceptance_test` and `helpers_test` files from the globals
resolver per #15058.
2) Update `setupForTesting` to reference the router factory in the registry
instead of the globally resolved one on the application instance.
3) Extend the Router during application registry setup to isolate
`reopen` calls.

### DIFF
--- a/packages/ember-testing/lib/ext/application.js
+++ b/packages/ember-testing/lib/ext/application.js
@@ -60,7 +60,8 @@ EmberApplication.reopen({
     the app when your tests are ready to run. It also sets the router's
     location to 'none', so that the window's location will not be modified
     (preventing both accidental leaking of state between tests and interference
-    with your testing framework).
+    with your testing framework). `setupForTesting` should only be called after
+    setting a custom `router` class (for example `App.Router = Router.extend(`).
 
     Example:
 
@@ -76,7 +77,7 @@ EmberApplication.reopen({
 
     this.testing = true;
 
-    this.Router.reopen({
+    this.resolveRegistration('router:main').reopen({
       location: 'none'
     });
   },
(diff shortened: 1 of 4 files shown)


mixonic:
This is close to complete, but I'm going to call it WIP. I split the changes to application code into a commit on their own- the change causes two failures. I'm not sure if the tests are poor or if the change regresses on an edge case. Will figure it out and make sure tests pass on that commit alone.

packages/ember-testing/lib/ext/application.js
```diff
@@ -76,7 +76,7 @@ EmberApplication.reopen({
 
     this.testing = true;
 
-    this.Router.reopen({
+    this.resolveRegistration('router:main').reopen({
```
rwjblue: I wonder if we can extend and re-register instead of reopen. Thoughts?

rwjblue: FWIW, this concern is definitely non-blocking. Feel free to land this when you are ready.  I just think we _can_ avoid the reopen, but I may be missing something (or just be wrong 😝).

mixonic: No, I don't think we can. I played with a few different ways to support re-registration and none of them were going very well.

This change, practically, means that you *must* call `App.setupForTesting()` *after* you set a custom router (usually via `App.Router = Router.extend(`). I updated the docs for `setupForTesting` to make this clear, it was already implied by the `reopen` call and the docs, but was hopelessly coupled to the globals resolver.

Also in practice, this `reopen` doesn't matter. In Ember-CLI apps (most apps) we set `location: 'none'` via the `env` anyway. So there is some dubious stuff going on here.

Perhaps instead of forcing `location: 'none'`, we should warn or assert if the location is something other than `none`.

I'm going to call this good for canary.

PR #15339, opened 2017-06-08
calling repeated methods in for loop in `metal/meta` 

bekzod:


rwjblue:
This looks good, but I think we'd (at least @krisselden and I have discussed) like to remove the meta-programming here and just define `Meta` like a normal class. Yes, there will be some amount of duplication, but I believe that it will be much easier to understand and still be essentially the same after compression...

Would you mind updating this to unrolling the meta-programmyness?

bekzod:
Sure, yeah I agree that it looks funky, can this be merged for now :P ?

stefanpenner (approved):

PR #15341, opened 2017-06-08
cleanup `metal/chain`

bekzod:


packages/ember-metal/lib/chains.js
```diff
@@ -176,10 +176,7 @@ class ChainNode {
 
   destroy() {
     if (this._watching === true) {
-      let obj = this._object;
-      if (obj) {
```
bekzod: this check not needed because `removeChainWatcher` has `isObject` check inside

packages/ember-metal/lib/chains.js
```diff
@@ -275,9 +272,7 @@ class ChainNode {
       let parentValue = this._parent.value();
 
       if (parentValue !== this._object) {
-        if (this._object !== undefined) {
```
bekzod: this check not needed because `removeChainWatcher` has `isObject` check inside

packages/ember-metal/lib/chains.js
stefanpenner: thanks for leaving that note, it wasn't clear from the PR

PR #15345, opened 2017-06-09
fix factory injection deprecation link

### LINKED ISSUE #2912: ember-metal.model_factory_injections anchor is broken?
The anchor for the `ember-metal.model_factory_injections` deprecation has an unexpected pair of `-em-`'s in it around `factory`:

> https://emberjs.com/deprecations/v2.x/#toc_ember-model-em-factory-em-injections-removed, 

which doesn't match what is being linked to from the [actual deprecation](https://github.com/emberjs/ember.js/blob/d338c608e5ce00eb98d2617f4b2e7d102035b6e4/packages/ember/lib/index.js#L188).

I was going to make a PR to ember to fix the link but this seems like it shouldn't be happening in the first place as I don't see any other anchors on https://emberjs.com/deprecations/v2.x with these `-em-`s in them.  Is there a reason this is happening?

Dhaulagiri:
The fact that this deprecation's id has underscores in it is causing some unexpected behavior when it is pulled in to the [docs](https://emberjs.com/deprecations/v2.x) in that it cause the word `factory` to be italicized and the anchor url to have unexpected characters around the word factory as well.  This PR at least fixes the link so that it takes you to the correct section of the deprecation guide.  Ideally we could change the deprecation id to be dasherized, but I'm not sure if it would be too late to do that?

Fixes https://github.com/emberjs/website/issues/2912

### DIFF
--- a/packages/ember/lib/index.js
+++ b/packages/ember/lib/index.js
@@ -185,7 +185,7 @@ if (DEBUG) {
         {
           id: 'ember-metal.model_factory_injections',
           until: '2.17.0',
-          url: 'https://emberjs.com/deprecations/v2.x#toc_code-ember-model-factory-injections'
+          url: 'https://emberjs.com/deprecations/v2.x/#toc_ember-model-em-factory-em-injections-removed'
         }
       );
     },


rwjblue:
Thank you @Dhaulagiri!

PR #15346, opened 2017-06-10
wrap `validationCache` and `.validateInjections` and  in DEBUG

bekzod:
since `registery.validateInjections` and `validationCache` is used only in DEBUG mode I think it is appropriate to wrap them in DEBUG


rwjblue:
Thank you!

PR #15347, opened 2017-06-10
[BUGFIX beta] Make better errors for meta updates after object destruction.

rwjblue:
Prior to this change the errors being thrown by `Meta` were extremely hard to track down since no "real" information was given.

After this change, all errors include the `object.toString` and property name in question (if applicable).

### DIFF
--- a/packages/ember-metal/lib/meta.js
+++ b/packages/ember-metal/lib/meta.js
@@ -1,7 +1,8 @@
 import {
   HAS_NATIVE_WEAKMAP,
   lookupDescriptor,
-  symbol
+  symbol,
+  toString
 } from 'ember-utils';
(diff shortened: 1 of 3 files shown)


packages/ember-metal/lib/meta.js
```diff
@@ -374,7 +375,7 @@ export class Meta {
   }
 
   writeWatching(subkey, value) {
-    assert(`Cannot call writeWatching after the object is destroyed.`, !this.isMetaDestroyed());
+    assert(`Cannot update watchers for \`hello\` on \`${toString(this.source)}\` after it has been destroyed.`, !this.isMetaDestroyed());
```
nag5000: hello?

rwjblue: Hi 👋

PR #15349, opened 2017-06-10
[PERF] Updated babel block-scoping config

### LINKED ISSUE #15233: Update transpilation to avoid closures when using `let`.
In some cases using `let` can force Babel transpilation to include extra closures. We should update our Babel options to throw an error when this occurs.

See https://github.com/emberjs/data/pull/4950 for an example of how to do this.

patsy-issa:
Fixes #15233


rwjblue (approved):

rwjblue:
Thank you!!

PR #15356, opened 2017-06-13
[DOC release] replace lingering NPM commands.

patricksrobertson:
Part of the contributing guide tells you to use yarn, part NPM. I
suspect we want to use yarn all the way down.


rwjblue:
Awesome, thank you!

Issue #15327, opened 2017-06-05
Error during initial render causes non-globals `Application` visit promise to hang


mixonic:
Preamble:

`Ember.Application` has a globals a non-globals mode. This has nothing to do with the resolver. Instead, it refers to applications which have been autobooted. Amongst other impacts of being globals, these applications set a `MyApp.__deprecatedInstance__` that implies only a single `applicationInstance` is running for a given app.

---

Given a test like the following:

```js
import {
  moduleFor,
  ApplicationTestCase,
  AutobootApplicationTestCase
} from 'internal-test-helpers';

moduleFor('The {{link-to}} helper', class extends ApplicationTestCase {
  [`@test the {{link-to}} helper throws a useful error if you invoke it wrong`](assert) {
    this.router.map(function() {
      this.route('post', { path: 'post/:post_id' });
    });
    this.addTemplate('application', `{{#link-to 'post'}}Post{{/link-to}}`);

    assert.throws(() => {
      this.visit('/');
    }, /(You attempted to define a `\{\{link-to "post"\}\}` but did not pass the parameters required for generating its dynamic segments.|You must provide param `post_id` to `generate`)/);
  }
});
```

All following tests will fail with errors like:

* `Assertion Failed: You cannot use the same root element (DIV) multiple times in an Ember.Application`
* `Error: a glimmer transaction was begun, but one already exists. You may have a nested transaction`

This is because the application instance created during the `visit` call is never destroyed. The way destruction is supposed to happen is simple: The `AbstractApplicationTestCase` [sets `this.applicationInstance` on the test context](https://github.com/emberjs/ember.js/blob/d338c608e5ce00eb98d2617f4b2e7d102035b6e4/packages/internal-test-helpers/lib/test-cases/abstract-application.js#L63) after a successful transition. At [teardown the instance is destroyed](https://github.com/emberjs/ember.js/blob/d338c608e5ce00eb98d2617f4b2e7d102035b6e4/packages/internal-test-helpers/lib/test-cases/abstract-application.js#L52). If the initial transition fails (say an exception raised from the model hook) then there is no `applicationInstance`.

In theory this is just fine, when a transition fails [`Application.visit` already takes care of destroying the instance](https://github.com/emberjs/ember.js/blob/d338c608e5ce00eb98d2617f4b2e7d102035b6e4/packages/ember-application/lib/system/application.js#L978). The test harness should only need to know about success cases.

[...]

PR #15358, opened 2017-06-13
Decouple link-to + query param tests from the global resolver.

patricksrobertson:
This should be good for review now!

### Old
There's a test skipped in the port over to the modern resolver. 50% of the time the test suite runs all the following tests fail with "Assertion Failed: You cannot use the same root element (DIV) multiple times in an Ember.Application". This leads me to believe the error is not allowing the teardown method to run, but I have no glaringly obvious means to force it.

### DIFF
--- a/packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
+++ b/packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
@@ -1,506 +1,578 @@
-import { set, run } from 'ember-metal';
 import { Controller, RSVP } from 'ember-runtime';
-import { Route, NoneLocation } from 'ember-routing';
-import { compile } from 'ember-template-compiler';
-import { Application } from 'ember-application';
-import { jQuery } from 'ember-views';
-import { setTemplates, setTemplate } from 'ember-glimmer';
+import { Route, Router } from 'ember-routing';
+import { moduleFor, ApplicationTestCase, AutobootApplicationTestCase } from 'internal-test-helpers';
 
-let Router, App, router, registry, container;
-
-function bootApplication() {
-  router = container.lookup('router:main');
-  run(App, 'advanceReadiness');
-}
-
-
-function shouldNotBeActive(selector) {
-  checkActive(selector, false);
-}
-
-function shouldBeActive(selector) {
-  checkActive(selector, true);
-}
-
-function checkActive(selector, active) {
-  let classList = jQuery(selector, '#qunit-fixture')[0].className;
-  equal(classList.indexOf('active') > -1, active, selector + ' active should be ' + active.toString());
-}
-
-let updateCount, replaceCount;
-
-function sharedSetup() {
-  App = Application.create({
-    name: 'App',
-    rootElement: '#qunit-fixture'
-  });
-
-  App.deferReadiness();
-
-  updateCount = replaceCount = 0;
-  App.Router.reopen({
-    location: NoneLocation.create({
-      setURL(path) {
-        updateCount++;
-        set(this, 'path', path);
-      },
-
-      replaceURL(path) {
-        replaceCount++;
-        set(this, 'path', path);
-      }
-    })
-  });
-
-  Router = App.Router;
-  registry = App.__registry__;
-  container = App.__container__;
-}
-
-function sharedTeardown() {
-  run(() => App.destroy());
-  setTemplates({});
-}
-
-QUnit.module('The {{link-to}} helper: invoking with query params', {
-  setup() {
-    run(() => {
-      sharedSetup();
-
-      App.IndexController = Controller.extend({
-        queryParams: ['foo', 'bar', 'abool'],
-        foo: '123',
-        bar: 'abc',
-        boundThing: 'OMG',
-        abool: true
-      });
-
-      App.AboutController = Controller.extend({
+moduleFor('The {{link-to}} helper: invoking with query params', class extends ApplicationTestCase {
+  constructor() {
+    super();
+    let indexProperties = {
+      foo: '123',
+      bar: 'abc'
+    }
+    this.add('controller:index', Controller.extend({
+      queryParams: ['foo', 'bar', 'abool'],
+      foo: indexProperties.foo,
+      bar: indexProperties.bar,
+      boundThing: 'OMG',
+      abool: true
+    }));
+    this.add('controller:about', Controller.extend({
         queryParams: ['baz', 'bat'],
         baz: 'alex',
         bat: 'borf'
+    }));
+    this.indexProperties = indexProperties;
+  }
+
+  shouldNotBeActive(assert, selector) {
+    this.checkActive(assert, selector, false);
+  }
+
+  shouldBeActive(assert, selector) {
+    this.checkActive(assert, selector, true);
+  }
+
+  getController(name) {
+    return this.applicationInstance.lookup(`controller:${name}`);
+  }
+
+  checkActive(assert, selector, active) {
+    let classList = this.$(selector)[0].className;
+    assert.equal(classList.indexOf('active') > -1, active, selector + ' active should be ' + active.toString());
+  }
+
+  [`@test doesn't update controller QP properties on current route when invoked`](assert) {
+    this.addTemplate('index', `
+      {{#link-to 'index' id='the-link'}}Index{{/link-to}}
+    `);
+
+    return this.visit('/').then(() => {
+      this.click('#the-link');
+      let indexController = this.getController('index');
+
+      assert.deepEqual(indexController.getProperties('foo', 'bar'),
+        this.indexProperties,
+        'controller QP properties do not update'
+      );
+    });
+  }
+
+  [`@test doesn't update controller QP properties on current route when invoked (empty query-params obj)`](assert) {
+    this.addTemplate('index', `
+      {{#link-to 'index' (query-params) id='the-link'}}Index{{/link-to}}
+    `);
+
+    return this.visit('/').then(() => {
+      this.click('#the-link');
+      let indexController = this.getController('index');
+
+      assert.deepEqual(indexController.getProperties('foo', 'bar'),
+        this.indexProperties,
+        'controller QP properties do not update'
+      );
+    });
+  }
+
+  [`@test doesn't update controller QP properties on current route when invoked (empty query-params obj, inferred route)`](assert) {
+    this.addTemplate('index', `
+      {{#link-to (query-params) id='the-link'}}Index{{/link-to}}
+    `);
+
+    return this.visit('/').then(() => {
+      this.click('#the-link');
+      let indexController = this.getController('index');
+
+      assert.deepEqual(indexController.getProperties('foo', 'bar'),
+        this.indexProperties,
+        'controller QP properties do not update'
+      );
+    });
+  }
+
+  ['@test updates controller QP properties on current route when invoked'](assert) {
+    this.addTemplate('index', `
+      {{#link-to 'index' (query-params foo='456') id="the-link"}}
+        Index
+      {{/link-to}}
+    `);
+
+    return this.visit('/').then(() => {
+      this.click('#the-link');
+      let indexController = this.getController('index');
+
+      assert.deepEqual(indexController.getProperties('foo', 'bar'),
+        { foo: '456', bar: 'abc' },
+        'controller QP properties updated'
+      );
+    });
+  }
+
+  ['@test updates controller QP properties on current route when invoked (inferred route)'](assert) {
+    this.addTemplate('index', `
+      {{#link-to (query-params foo='456') id="the-link"}}
+        Index
+      {{/link-to}}
+    `);
+
+    return this.visit('/').then(() => {
+      this.click('#the-link');
+      let indexController = this.getController('index');
+
+      assert.deepEqual(indexController.getProperties('foo', 'bar'),
+        { foo: '456', bar: 'abc' },
+        'controller QP properties updated'
+      );
+    });
+  }
+
+  ['@test updates controller QP properties on other route after transitioning to that route'](assert) {
+    this.router.map(function() {
+      this.route('about');
+    });
+
+    this.addTemplate('index',`
+      {{#link-to 'about' (query-params baz='lol') id='the-link'}}
+        About
+      {{/link-to}}
+    `);
+
+    return this.visit('/').then(() => {
+      let theLink = this.$('#the-link');
+      assert.equal(theLink.attr('href'), '/about?baz=lol');
+
+      this.click('#the-link');
+      let aboutController = this.getController('about');
+
+      assert.deepEqual(aboutController.getProperties('baz', 'bat'),
+        { baz: 'lol', bat: 'borf' },
+        'about controller QP properties updated'
+      );
+    });
+  }
+
+  ['@test supplied QP properties can be bound'](assert) {
+    this.addTemplate('index', `
+      {{#link-to (query-params foo=boundThing) id='the-link'}}Index{{/link-to}}
+    `);
+
+    return this.visit('/').then(() => {
+      let indexController = this.getController('index');
+      let theLink = this.$('#the-link')
+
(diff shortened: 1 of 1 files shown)


patricksrobertson:
@mixonic this should be ready for review at this point.

packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
```diff
@@ -1,506 +1,567 @@
...
+  shouldNotBeActive(selector) {
+    this.checkActive(selector, false);
+  }
+
+  shouldBeActive(selector) {
+    this.checkActive(selector, true);
+  }
+
+  checkActive(selector, active) {
+    let classList = this.$(selector, '#qunit-fixture')[0].className;
```
mixonic: `this.$(selector)` will already be scoped here, no need to pass a second argument.

packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
```diff
@@ -1,506 +1,567 @@
...
+    equal(classList.indexOf('active') > -1, active, selector + ' active should be ' + active.toString());
+  }
+
+  ['@test doesn\'t update controller QP properties on current route when invoked']() {
```
mixonic: Using backticks for this string would avoid the escaping of a quote.

packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
```diff
@@ -1,506 +1,567 @@
...
 
-  run(jQuery('#the-link'), 'click');
-  let indexController = container.lookup('controller:index');
-  deepEqual(indexController.getProperties('foo', 'bar'), { foo: '123', bar: 'abc' }, 'controller QP properties not');
-});
+    this.visit('/').then(() => {
+      this.$('#the-link').click();
+      let indexController = this.applicationInstance.lookup('controller:index');
 
-QUnit.test('link-to with no params throws', function() {
```
mixonic: I see this is moved to the end of the file with a new name 👍

packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
```diff
@@ -1,506 +1,567 @@
...
+
+  ['@test updates controller QP properties on current route when invoked']() {
+    this.addTemplate('index', `
+      {{#link-to 'index' (query-params foo='456') id="the-link"}}
+        Index
+      {{/link-to}}
+    `);
+
+    this.visit('/').then(() => {
+      run(() => this.$('#the-link').click());
```
mixonic: `run` here should not be needed.

packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
```diff
@@ -1,506 +1,567 @@
...
+
+  ['@test updates controller QP properties on current route when invoked (inferred route)']() {
+    this.addTemplate('index', `
+      {{#link-to (query-params foo='456') id="the-link"}}
+        Index
+      {{/link-to}}
+    `);
+
+    this.visit('/').then(() => {
+      run(() => this.$('#the-link').click());
```
mixonic: `run` here not needed.

packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
mixonic: Also, test declarations should have `assert` passed in as their argument. Then you'll need to pass that `assert` into other assertion helpers:

```js
  checkSomething(assert, arg) {
    assert.equal(arg, 'expected val', 'description of assertion');
  },
  [`@test foobar`](assert) {
    this.checkSomething(assert, 'other arg');
    assert.ok(true, 'yup, it is true');
  }
```

packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
```diff
@@ -1,506 +1,567 @@
...
+    this.addTemplate('index',`
+      {{#link-to 'about' (query-params baz='lol') id='the-link'}}
+        About
+      {{/link-to}}
+    `);
+
+    this.visit('/').then(() => {
+      let theLink = this.$('#the-link');
+      equal(theLink.attr('href'), '/about?baz=lol');
+      run(() => theLink.click());
```
mixonic: `run` is not needed.

packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
```diff
@@ -1,506 +1,567 @@
...
-QUnit.test('updates controller QP properties on current route when invoked (inferred route)', function() {
-  setTemplate('index', compile(`{{#link-to (query-params foo='456') id='the-link'}}Index{{/link-to}}`));
-  bootApplication();
+      equal(theLink.attr('href'), '/?foo=OMG');
 
-  run(jQuery('#the-link'), 'click');
-  let indexController = container.lookup('controller:index');
-  deepEqual(indexController.getProperties('foo', 'bar'), { foo: '456', bar: 'abc' }, 'controller QP properties updated');
-});
+      run(indexController, 'set', 'boundThing', 'ASL');
```
mixonic: Prefer:

```js
this.runTask(() => indexController.set('boundThing', 'ASL'));
```

packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
```diff
@@ -1,506 +1,567 @@
...
+      queryParams: ['page'],
+      page: 1
+    }));
 
-QUnit.test('The {{link-to}} with only query params always transitions to the current route with the query params applied', function() {
-  // Test harness for bug #12033
+    this.visit('/').then(()=> {
+      let router = this.appRouter;
+      let carsController = this.applicationInstance.lookup('controller:cars');
+      this.runTask(() => router.handleURL('/cars/create'));
```
mixonic: Here you should be able to chain `visit` calls.


```js
this.visit('/').then(() => {
  assert.ok(true, 'something');
  return this.visit('/otherRoute');
}).then(() => {
  assert.ok(true, 'after visiting a new route');
});
```

packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
```diff
@@ -1,506 +1,567 @@
...
 
-  setTemplate('cars', compile(`
-    {{#link-to 'cars.create' id='create-link'}}Create new car{{/link-to}}
-    {{#link-to (query-params page='2') id='page2-link'}}Page 2{{/link-to}}
-    {{outlet}}
-  `));
+      this.runTask(() => {
```
mixonic: These three `runTask` calls should not be needed.

packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
```diff
@@ -1,506 +1,567 @@
...
+      sort: 'title',
+      showDetails: true
+    }));
+
+    this.visit('/').then(() => {
+      let router = this.appRouter;
+      //Basic tests
+      this.shouldNotBeActive('#cat-link');
+      this.shouldNotBeActive('#dog-link');
+      run(router, 'handleURL', '/?foo=cat');
```
mixonic: Again this can chain `visit` calls :-) All these `handleURL` bits can be replaced with `visit`.

packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
```diff
@@ -1,506 +1,567 @@
...
 
-  bootApplication();
-  equal(jQuery('#foos-link').attr('href'), '/foos');
-  equal(jQuery('#baz-foos-link').attr('href'), '/foos?baz=true');
-  equal(jQuery('#bars-link').attr('href'), '/bars?quux=true');
+      run(() => foosLink.click());
+      this.shouldNotBeActive('#foos-link');
 
-  equal(router.get('location.path'), '');
+      run(foos, 'resolve');
```
mixonic: `this.runTask(() => resolve());`

packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
```diff
@@ -1,506 +1,567 @@
...
 
-  run(foos, 'resolve');
+        this.add('router:main', Router.extend({
+          location: 'none'
+        }));
 
-  equal(router.get('location.path'), '/foos');
-  shouldBeActive('#foos-link');
-});
+        this.router.map(function() {});
```
mixonic: Should not need to declare an empty route map.

packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
```diff
@@ -1,506 +1,567 @@
...
+      equal(router.get('location.path'), '/foos');
+      this.shouldBeActive('#foos-link');
+    });
+  }
+});
 
-  run(jQuery('#bars-link'), 'click');
-  shouldNotBeActive('#bars-link');
+moduleFor('The {{link-to}} helper + query params - globals mode app', class extends AutobootApplicationTestCase {
+  [`@test the {{link-to}} helper throws a useful error if you invoke it wrong`](assert) {
```
mixonic: Please add a link back to https://github.com/emberjs/ember.js/issues/15327 in a comment, thank you!

mixonic (changes requested):
Great start on this! A few items I've noted examples of, but I will ask you to clean up across the whole PR:

* Calls to `jqueryObj.click()` don't need to be wrapped in a runloop.
* In general if there is a runloop, use `this.runTask(() => { /* somthing */ });` instead.
* `router.handleURL` is just a legacy way to write `this.visit`, and we should move toward the latter.
* `assert` is the first argument to a test declaration, and assertion methods should be called on it.

mixonic:
Refs #15058

packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
patricksrobertson: What I found is that `run` was needed to get the tests to pass here. Why I think it is that way: we are waiting for the query params to change on the controller as a result of user interaction. Is there another way to handle this without waiting via run loop?

packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
```diff
@@ -1,506 +1,577 @@
...
+
+  ['@test updates controller QP properties on current route when invoked (inferred route)'](assert) {
+    this.addTemplate('index', `
+      {{#link-to (query-params foo='456') id="the-link"}}
+        Index
+      {{/link-to}}
+    `);
+
+    this.visit('/').then(() => {
+      this.runTask(() => this.$('#the-link').click());
```
karthiick: @patricksrobertson `click` is already wrapped within a runTask here. So there is no need for it, i guess.

https://github.com/emberjs/ember.js/blob/master/packages/internal-test-helpers/lib/test-cases/abstract.js#L91

packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
```diff
@@ -1,506 +1,577 @@
...
+  }
+
+  [`@test doesn't update controller QP properties on current route when invoked`](assert) {
+    this.addTemplate('index', `
+      {{#link-to 'index' id='the-link'}}Index{{/link-to}}
+    `);
+
+    this.visit('/').then(() => {
+      this.$('#the-link').click();
+      let indexController = this.applicationInstance.lookup('controller:index');
```
karthiick: Why not extend `QueryParamTestCase` and make use of getController method here.

https://github.com/emberjs/ember.js/blob/master/packages/internal-test-helpers/lib/test-cases/query-param.js#L7
https://github.com/emberjs/ember.js/blob/master/packages/internal-test-helpers/lib/test-cases/query-param.js#L49

Or we can at least define getController method like that in here and make use of it.

patricksrobertson: Alternate idea: add `getController` to `AbstractTestCase`.

packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
```diff
@@ -1,506 +1,577 @@
...
+      this.shouldBeActive(assert, '#app-link');
+      this.shouldBeActive(assert, '#parent-link');
 
-  run(jQuery('#bars-link'), 'click');
-  shouldNotBeActive('#bars-link');
+      this.click('#app-link');
 
-  run(jQuery('#foos-link'), 'click');
-  shouldNotBeActive('#foos-link');
+      equal(router.get('location.path'), '/parent');
```
karthiick: Missed `assert`

packages/internal-test-helpers/lib/test-cases/application.js
```diff
@@ -45,6 +45,10 @@ export default class ApplicationTestCase extends TestResolverApplicationTestCase
     }
   }
 
+  getController(name) {
```
mixonic: I'm not a great fan of adding these `getTypexxx` APIs. Seems like we will forever be playing whack-a-mole and guessing which ones are in the base class and which are not.

Can you just add it to the `moduleFor` using `getController`? I'd rather see it be a local helper than pushed to the base class.

packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
```diff
@@ -1,506 +1,577 @@
...
-    router.handleURL('/search/results?search=same&sort=title&showDetails=true');
-  });
-  //shouldBeActive('#same-sort-child-only');
-  shouldBeActive('#same-search-parent-only');
-  shouldNotBeActive('#change-search-parent-only');
-  shouldBeActive('#same-search-same-sort-child-and-parent');
-  shouldNotBeActive('#same-search-different-sort-child-and-parent');
-  shouldNotBeActive('#change-search-same-sort-child-and-parent');
-});
+    this.visit('/?page=5').then(() => {
```
mixonic: If these `visit` calls are to run in sequence, they should be chained:

```js
this.visit('/').then(() => {
  this.shouldNotBeActive(assert, '#page-link');
  return this.visit('/?page=5');
}).then(() => {
  this.shouldBeActive(assert, '#page-link');
});
```

packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
```diff
@@ -1,506 +1,577 @@
...
+      this.shouldBeActive(assert, '#dog-link');
+      this.shouldNotBeActive(assert, '#cat-link');
+      this.shouldBeActive(assert, '#change-nothing');
 
-  run(() => {
-    equal(router.currentRouteName, 'cars.create');
-    jQuery('#close-link').click();
-  });
+    });
+    this.visit('/search?search=same').then(() => {
```
mixonic: I mentioned some chaining of `visit` call stuff below. It applies here and maybe a few other spots as well. Basically because we currently have synchronous `visit` implementations, these test pass even though we haven't properly chained them. However in the future we may make this work async, and for that eventuality let's chain them. The fact that they are promises implies we should chain them to ensure sequential execution anyway.

mixonic (changes requested):
This is great work @patricksrobertson. I've asked for two pretty minor changes, although the tweak to chain `visit` will mean another quick read and edit of the file.

Thank for you this iteration!

packages/internal-test-helpers/lib/test-cases/application.js
patricksrobertson: I was avoiding using the `QueryParam` test case, because it has a lot of additional API space that would go unused. Adding this to the `moduleFor` sounds like a better plan to me.

mixonic (approved):

PR #15362, opened 2017-06-14
Decouple test from the default resolver for instance_initializers_test

karthiicksiva:
#15058

### DIFF
--- a/packages/ember-application/tests/system/instance_initializers_test.js
+++ b/packages/ember-application/tests/system/instance_initializers_test.js
@@ -1,410 +1,375 @@
-import { run } from 'ember-metal';
-import Application from '../../system/application';
-import ApplicationInstance from '../../system/application-instance';
+import { assign } from 'ember-utils';
+import { moduleFor, AutobootApplicationTestCase } from 'internal-test-helpers';
+import { Application, ApplicationInstance } from 'ember-application';
 import { jQuery } from 'ember-views';
 
-let app;
+moduleFor('Ember.Application instance initializers', class extends AutobootApplicationTestCase {
+  constructor() {
+    jQuery('#qunit-fixture').html(`
+      <div id="one">ONE</div>
+      <div id="two">TWO</div>
+    `);
+    super();
+  }
 
-QUnit.module('Ember.Application instance initializers', {
+  get applicationOptions() {
+    return assign(super.applicationOptions, {
+      rootElement: '#one'
+    });
+  }
+
+  createSecondApplication(options, MyApplication=Application) {
+    let myOptions = assign(this.applicationOptions, {
+      rootElement: '#two'
+    }, options);
+    let secondApp = this.secondApp = MyApplication.create(myOptions);
+    return secondApp;
+  }
+
   teardown() {
-    if (app) {
-      run(() => app.destroy());
+    super.teardown();
+
+    if (this.secondApp) {
+      this.runTask(() => this.secondApp.destroy());
     }
   }
-});
 
-QUnit.test('initializers require proper \'name\' and \'initialize\' properties', function() {
-  let MyApplication = Application.extend();
+  [`@test initializers require proper 'name' and 'initialize' properties`]() {
+    let MyApplication = Application.extend();
 
-  expectAssertion(() => {
-    run(() => {
+    expectAssertion(() => {
       MyApplication.instanceInitializer({ name: 'initializer' });
     });
-  });
 
-  expectAssertion(() => {
-    run(() => {
+    expectAssertion(() => {
       MyApplication.instanceInitializer({ initialize() {} });
     });
-  });
-});
 
-QUnit.test('initializers are passed an app instance', function() {
-  let MyApplication = Application.extend();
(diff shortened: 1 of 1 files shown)


packages/ember-application/tests/system/instance_initializers_test.js
```diff
@@ -1,410 +1,386 @@
...
+    let myOptions = assign({
+      rootElement: '#one'
+    }, options);
+    super.createApplication(myOptions, MyApplication);
+  }
+
+  createSecondApplication(options, MyApplication) {
+    let myOptions = assign({
+      rootElement: '#two',
+      Resolver: TestResolver
```
mixonic: This is a great start! Instead of importing the `TestResolver`, I suggest something more like [what happens here](https://github.com/emberjs/ember.js/blob/f4eef941cbb03e17f4e7eae1d52b0aa39de8605f/packages/ember-application/tests/system/application_test.js#L55-L66).

I provided a version of the code that I think works here, replacing everything from `createApplication` to `teardown`. Some notes:

* Uses `applicationOptions` instead of changing `createApplication`.
* Uses `applicationOptions` in `createSecondApplication` (avoids needing `TestResolver`.
* Default the value of `MyApplication` to `Application` in `createSecondApplication`.
* Setting `this.secondResolver` isn't needed as that property is not used.
* Prefer `runTask` over `run` (avoid the import).

```js
  get applicationOptions() {
      rootElement: '#one'
    });
  }

  createSecondApplication(options, MyApplication=Application) {
    let myOptions = assign(this.applicationOptions, {
      rootElement: '#two'
    }, options);
    let secondApp = this.secondApp = MyApplication.create(myOptions);
    return secondApp;
  }

  teardown() {
    super.teardown();

    if (this.secondApp) {
      this.runTask(() => this.secondApp.destroy());
    }
  }
```

packages/ember-application/tests/system/instance_initializers_test.js
```diff
@@ -1,410 +1,386 @@
...
+        MyApplication.instanceInitializer({ name: 'initializer' });
+      });
     });
-  });
 
-  expectAssertion(() => {
-    run(() => {
-      MyApplication.instanceInitializer({ initialize() {} });
+    expectAssertion(() => {
+      run(() => {
```
mixonic: As mentioned above, using `this.runTask` instead of `run` is preferred. This allows us to change the runloop abstraction later if we want to. This suggestion applies across the PR.

Here specifically, I don't expect we need a runloop at all. If these assertions are just triggered by `instanceInitializer` the runloop is spurious.

packages/ember-application/tests/system/instance_initializers_test.js
```diff
@@ -1,410 +1,386 @@
...
+      initialize(instance) {
+        assert.ok(instance instanceof ApplicationInstance, 'initialize is passed an application instance');
+      }
     });
-  });
-});
 
-QUnit.test('initializers are passed an app instance', function() {
-  let MyApplication = Application.extend();
+    this.runTask(() => this.createApplication({}, MyApplication));
```
mixonic: 👍 a good example of how much cleaner this is :-)

packages/ember-application/tests/system/instance_initializers_test.js
```diff
@@ -1,410 +1,386 @@
...
+      }
     });
-  });
-});
 
-QUnit.test('initializers are executed in their own context', function() {
-  expect(1);
+    this.runTask(() => {
+      this.createApplication({
+        ready() {
```
mixonic: Can this `ready() {` stay on the extended object instead of being moved to the `create` options?

packages/ember-application/tests/system/instance_initializers_test.js
```diff
@@ -1,410 +1,386 @@
...
+  }
 
-  MyApplication.instanceInitializer({
-    name: 'initializer',
-    initialize() {
-      ok(!readyWasCalled, 'ready is not yet called');
-    }
-  });
+  [`@test initializers are run before ready hook`](assert) {
+    expect(2);
```
mixonic: `assert.`

packages/ember-application/tests/system/instance_initializers_test.js
```diff
@@ -1,410 +1,386 @@
...
+      initialize(instance) {
+        assert.ok(!!instance, 'Initializer got an instance');
+      }
     });
-  });
 
-  run(app, 'reset');
+    this.runTask(() => this.createApplication({}, MyApplication));
+
+    run(this.application, 'reset');
```
mixonic: `this.runTask(() => this.application.reset());` please.

mixonic (changes requested):
Highlighted a few more changes, but this is getting pretty close to ready. Thanks for the continued iterations!

karthiick:
@mixonic i have made the changes. Let me know, if i have missed something.

mixonic (approved):

mixonic:
@karthiick this is great! Thank you for your time on these! 😄

PR #15364, opened 2017-06-15
Expose listenersFor from Evented

claudioc:
## Reason
I had the need to remove listeners from an Evented-mixed class. There is already the nice `has()` method but I didn't find a way to loop through all the listeners and remove them. I noticed that the `ember-metal/events` object has the `listenersFor` method so I just exposed it through the Evented interface.

## Notes on test
- I was not able to `yarn install` ember@master because `"https://registry.yarnpkg.com/@glimmer/runtime/-/runtime-0.24.0-beta.4.tgz: Request failed "401 Unauthorized""`. Interestingly, `npm install` worked fine though
- I am not able to run tests even on the master branch: Broccoli explodes with `TypeError: container.js: path.inShadow is not a function` (I am using node 6.x on OSX 10.11)
- I didn't find any test for the Evented mixin

So, sorry about tests (I will try again from a newer laptop) but I wasn't able to try the PR itself, so take it as it is (or trash it)

Thanks!

### DIFF
--- a/packages/ember-metal/lib/events.js
+++ b/packages/ember-metal/lib/events.js
@@ -255,11 +255,11 @@ export function hasListeners(obj, eventName) {
 }
 
 /**
-  @private
   @method listenersFor
   @for Ember
   @param obj
   @param {String} eventName
+  @public
 */
 export function listenersFor(obj, eventName) {
   let ret = [];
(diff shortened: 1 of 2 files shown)


claudioc:
If this PR makes sense I will try to go forward and implement also the `removeAllListeners` method

mmun:
@claudioc Sorry for the major delay. This seems like a reasonable feature. If you're still interested, it would need to go through the RFC process (https://github.com/emberjs/rfcs).

PR #15365, opened 2017-06-16
Decouple toString tests from the Global resolver.

patricksrobertson:
I'm not sure this should be using the Lookup context, but I wanted to avoid radically changing the assertions.

### DIFF
--- a/packages/ember-application/tests/system/dependency_injection/to_string_test.js
+++ b/packages/ember-application/tests/system/dependency_injection/to_string_test.js
@@ -1,62 +1,58 @@
-import { guidFor } from 'ember-utils';
-import { ENV, context } from 'ember-environment'; // lookup, etc
-import { run } from 'ember-metal';
-import Application from '../../../system/application';
+import { guidFor, assign } from 'ember-utils';
 import { Object as EmberObject } from 'ember-runtime';
-import DefaultResolver from '../../../system/resolver';
+import { Resolver as DefaultResolver } from 'ember-application';
+import { moduleFor, ApplicationTestCase, ModuleBasedTestResolver,
+  DefaultResolverApplicationTestCase } from 'internal-test-helpers';
 
-let originalLookup, App;
+moduleFor('Ember.Application Dependency Injection - DefaultResolver#toString', class extends
+  DefaultResolverApplicationTestCase {
+  constructor() {
+    super();
+    this.createApplication();
+    this.application.Post = EmberObject.extend();
+    this.visit('/');
+  }
 
-QUnit.module('Ember.Application Dependency Injection – toString', {
-  setup() {
-    originalLookup = context.lookup;
+  ['@test factories'](assert) {
+    let PostFactory = this.applicationInstance.factoryFor('model:post').class;
+    assert.equal(
+      PostFactory.toString(), '.Post',
+      'expecting the model to be post'
+    );
+  }
 
-    run(() => {
-      App = Application.create();
-      context.lookup = {
-        App: App
-      };
-    });
+  ['@test instances'](assert) {
+    let post = this.applicationInstance.lookup('model:post');
+    let guid = guidFor(post);
 
-    App.Post = EmberObject.extend();
-  },
-
-  teardown() {
-    context.lookup = originalLookup;
-    run(App, 'destroy');
+    assert.equal(post.toString(), '<.Post:' + guid + '>', 'expecting the model to be post');
   }
 });
 
-QUnit.test('factories', function() {
-  let PostFactory;
-  PostFactory = App.__container__.factoryFor('model:post').class;
-  equal(PostFactory.toString(), 'App.Post', 'expecting the model to be post');
-});
+moduleFor('Ember.Application Dependency Injection - Resolver#toString', class extends ApplicationTestCase {
+  constructor() {
+    super();
+    this.visit('/');
+  }
 
-QUnit.test('instances', function() {
-  let post = App.__container__.lookup('model:post');
-  let guid = guidFor(post);
-
-  equal(post.toString(), '<App.Post:' + guid + '>', 'expecting the model to be post');
-});
-
-QUnit.test('with a custom resolver', function() {
-  run(App, 'destroy');
-
-  run(() => {
-    App = Application.create({
-      Resolver: DefaultResolver.extend({
-        makeToString(factory, fullName) {
+  get applicationOptions() {
+    return assign(super.applicationOptions, {
+      Resolver: class extends ModuleBasedTestResolver {
+        makeToString(_, fullName) {
           return fullName;
         }
-      })
+      }
     });
-  });
+  }
 
-  App.register('model:peter', EmberObject.extend());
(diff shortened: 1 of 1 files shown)


packages/ember-application/tests/system/dependency_injection/to_string_test.js
```diff
@@ -1,62 +1,63 @@
...
 import { Object as EmberObject } from 'ember-runtime';
-import DefaultResolver from '../../../system/resolver';
+
+import { moduleFor, AutobootApplicationTestCase, ModuleBasedTestResolver } from 'internal-test-helpers';
 
 let originalLookup, App;
 
-QUnit.module('Ember.Application Dependency Injection – toString', {
-  setup() {
+moduleFor('Ember.Application Dependency Injection - toString', class extends AutobootApplicationTestCase {
```
mixonic: Can this test use the `ApplicationTestCase` instead of the `AutobootApplicationTestCase`? Something like:

```js
moduleFor('Ember.Application Dependency Injection - DefaultResolver#toString', class extends ApplicationTestCase {
  constructor() {
    this._originalLookup = context.lookup;
    super();
    this.application.Post = EmberObject.extend();
    context.lookup = { App: this.application };
    return this.visit('/');
  }
  get applicationOptions {
      Resolver: DefaultResolver
    });
  }

  teardown() {
    super.teardown();
    context.lookup = originalLookup;
  }

  ['@test factories'](assert) {
    let PostFactory = this.applicationInstance.factoryFor('model:post').class;
    assert.equal(
      PostFactory.toString(), 'App.Post',
      'expecting the model to be post'
    );
  }
```

I'm actually pretty sure you can take out the `context` stuff entirely. With the default resolver the application itself is already the namespace. But I would need to muck with it to be sure.

packages/ember-application/tests/system/dependency_injection/to_string_test.js
```diff
@@ -1,62 +1,63 @@
...
-      Resolver: DefaultResolver.extend({
-        makeToString(factory, fullName) {
-          return fullName;
-        }
-      })
-    });
-  });
 
-  App.register('model:peter', EmberObject.extend());
+  ['@test factories'](assert) {
```
mixonic: Above I put this test into a module with `DefaultResolver` in the module name. That's important, since this test is really testing `toString` on the default resolver.

packages/ember-application/tests/system/dependency_injection/to_string_test.js
```diff
@@ -1,62 +1,63 @@
...
+    let PostFactory = App.__container__.factoryFor('model:post').class;
+    assert.equal(PostFactory.toString(),
+      'App.Post',
+      'expecting the model to be post');
+  }
+
+  ['@test instances'](assert) {
```
mixonic: This also tests the default resolver and can share a module with the above.

packages/ember-application/tests/system/dependency_injection/to_string_test.js
```diff
@@ -1,62 +1,63 @@
...
+    let post = App.__container__.lookup('model:post');
+    let guid = guidFor(post);
 
-  let peter = App.__container__.lookup('model:peter');
-  let guid = guidFor(peter);
+    assert.equal(post.toString(), '<App.Post:' + guid + '>', 'expecting the model to be post');
+  }
+
+  ['@test with a custom resolver'](assert) {
```
mixonic: This test does not test the global resolver, so it should have its own module. Something like:

```js
moduleFor('Ember.Application Dependency Injection - Resolver#toString', class extends ApplicationTestCase {
  constructor() {
    super();
    return this.visit('/');
  }
  get applicationOptions {
      Resolver: class extends ModuleBasedTestResolver {
        makeToString(factory, fullName) {
          return fullName;
        }
      }
    });
  }

  ['@test toString called on a resolver'](assert) {
    this.add('model:peter', EmberObject.extend());
    
    let peter = this.applicationInstance.lookup('model:peter');
    let guid = guidFor(peter);
    assert.equal(
      peter.toString(), `<model:peter:${guid}>`,
      'expecting the supermodel to be peter'
    );
  }
```

mixonic (changes requested):
This is a gnarly one. The first two tests are testing the default resolver, and the last testing the resolver API itself. Left some copious notes, I think you may be able to stop using `context` entirely.

mixonic:
Refs #15058

packages/ember-application/tests/system/dependency_injection/to_string_test.js
```diff
@@ -1,62 +1,60 @@
...
-        App: App
-      };
+import { Resolver as DefaultResolver } from 'ember-application';
+import { moduleFor, ApplicationTestCase, ModuleBasedTestResolver } from 'internal-test-helpers';
+
+moduleFor('Ember.Application Dependency Injection - DefaultResolver#toString', class extends ApplicationTestCase {
+  constructor() {
+    super();
+    this.application.Post = EmberObject.extend();
+    this.runTask(() => this.visit('/'));
```
mixonic: This run loop shouldn't be needed, `visit` [brings its own](https://github.com/emberjs/ember.js/blob/master/packages/internal-test-helpers/lib/test-cases/application.js#L38).

packages/ember-application/tests/system/dependency_injection/to_string_test.js
```diff
@@ -1,62 +1,60 @@
...
+  }
+  get applicationOptions() {
+    return assign(super.applicationOptions, {
+      Resolver: DefaultResolver
```
mixonic: I added a `DefaultResolverApplicationTestCase` in a recent PR, [example here](https://github.com/emberjs/ember.js/blob/4ec05b8091a5e701ffc95f4c9f687831f7478262/packages/ember-application/tests/system/application_test.js#L189). It would allow you to avoid this setup.

mixonic (changes requested):
Looking for two small changes, dropping `runTask`s around `visit` and to try and use the new [`DefaultResolverApplicationTestCase`](https://github.com/emberjs/ember.js/blob/4ec05b8091a5e701ffc95f4c9f687831f7478262/packages/internal-test-helpers/lib/test-cases/default-resolver-application.js#L11). Don't get too hung up on the latter, I *think* is it drop-in though.

mixonic (approved):

mixonic:
Looks great @patricksrobertson! I kicked the build on sauce, which is likely just intermittent.

patricksrobertson:
@mixonic this appears to be a consistent failure with sauce but I'm not sure how to go about debugging it. Any tips?

mixonic:
@patricksrobertson I'll kick it off one more time, but if you load up [the build failure](https://travis-ci.org/emberjs/ember.js/jobs/247507740) you can see the failure is on `not ok 1 SL_MS_Edge - bailout`. The error is `Error response status: 13, , UnknownError - An unknown server-side error occurred while processing the command. Selenium error: 'JSTestingReporterSL' is undefined`.

I don't think that is really a failure, but running tests with MS Edge would be the first thing to try if you want to attempt a local recreation.

PR #15367, opened 2017-06-16
Decouple test from the default resolver for initializers_test

karthiicksiva:
#15058

### DIFF
--- a/packages/ember-application/tests/system/initializers_test.js
+++ b/packages/ember-application/tests/system/initializers_test.js
@@ -1,422 +1,381 @@
-import { run } from 'ember-metal';
-import Application from '../../system/application';
+import { assign } from 'ember-utils';
+import { moduleFor, AutobootApplicationTestCase } from 'internal-test-helpers';
+import { Application } from 'ember-application';
 import { jQuery } from 'ember-views';
 
-let app;
+moduleFor('Ember.Application initializers', class extends AutobootApplicationTestCase {
+  constructor() {
+    jQuery('#qunit-fixture').html(`
+      <div id="one">ONE</div>
+      <div id="two">TWO</div>
+    `);
+    super();
+  }
 
-QUnit.module('Ember.Application initializers', {
+  get applicationOptions() {
+    return assign(super.applicationOptions, {
+      rootElement: '#one'
+    });
+  }
+
+  createSecondApplication(options, MyApplication=Application) {
+    let myOptions = assign(this.applicationOptions, {
+      rootElement: '#two'
+    }, options);
+    let secondApp = this.secondApp = MyApplication.create(myOptions);
+    return secondApp;
+  }
+
   teardown() {
-    if (app) {
-      run(() => app.destroy());
+    super.teardown();
+
+    if (this.secondApp) {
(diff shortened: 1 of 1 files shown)


packages/ember-application/tests/system/initializers_test.js
```diff
@@ -1,422 +1,396 @@
...
-    name: 'first',
-    before: 'second',
-    initialize(registry) {
-      order.push('first');
+    let app = this.application;
+
+    try {
+      app.boot().then(
+        (app) => {
+          assert.ok(false, 'The boot promise should not resolve when there is a boot error');
```
karthiick: I understand that, we are pausing & resuming the test runner here. But why? Would like to know the reason behind this.

@mixonic

mixonic: `app.boot().then(() => {` is not inside a runloop, so the thennable callback will be called in an autorun making it async. I think you could avoid the async-ness if you wrapped the promise a runloop, it would then flush in the actions queue. Nice catch:

```js
try {
  this.runTask(() => {
    app.boot().then(app => {
      assert.ok(false, 'The boot promise should not resolve when there is a boot error');
    }, error => {
      assert.ok(error instanceof Error, 'The boot promise should reject with an error');
      assert.equal(error.message, 'boot failure');
    });
  });
} catch (error) {
  // ...
```

packages/ember-application/tests/system/initializers_test.js
```diff
@@ -1,422 +1,396 @@
...
+moduleFor('Ember.Application initializers', class extends AutobootApplicationTestCase {
+  constructor() {
+    jQuery('#qunit-fixture').html(`
+      <div id="one">ONE</div>
+      <div id="two">TWO</div>
+    `);
+    super();
+  }
+
+  createApplication(options, MyApplication) {
```
mixonic: Using `applicationOptions` is preferred here. Replace `createApplication` through `teardown` with:

```js
 get applicationOptions() {
      rootElement: '#one'
    });
  }

  createSecondApplication(options, MyApplication=Application) {
    let myOptions = assign(this.applicationOptions, {
      rootElement: '#two'
    }, options);
    let secondApp = this.secondApp = MyApplication.create(myOptions);
    return secondApp;
  }

  teardown() {
    super.teardown();

    if (this.secondApp) {
      this.runTask(() => this.secondApp.destroy());
    }
  }
```

More details at https://github.com/emberjs/ember.js/pull/15362/files#r122541946

mixonic (changes requested):
Many of the changed I'd like to see are already laid out in https://github.com/emberjs/ember.js/pull/15362, but I'll reiterate a few:

* `assert` is passed as the first argument to tests, should be used for assertions instead of the globals.
* `this.runTask` is preferred over `run`
* `run` shouldn't be needed around `initializer` calls

Excellent beginning though, thank you for the continued iteration!!

karthiick:
This PR is ready to go. // @mixonic

mixonic (approved):

mixonic:
This looks great to me, thanks @karthiick 👍 I kicked the sauce build, it had what I presume was a spurious failure.

PR #15368, opened 2017-06-16
[FEATURE ember-module-unification] Initial implementation

iezer:
To support module unification, we must modify the container and registry to pass the source as second parameter to the resolver. This PR introduces a new feature flag called `ember-module-unification` which does just that.

I'll follow up with some PRs in the `ember-resolver` and `glimmer-resolver` repos, and an end-to-end example of an Ember app using module unification and private components.

When the feature flag is on, this change breaks two tests related to `expandLocalLookup`. I'm looking at a way to support the new functionality and make those tests pass. While I'm doing that, I'd appreciate some feedback on the overall design here.

**EDIT**

Related PRs:
 - https://github.com/ember-cli/ember-resolver/pull/196
 - https://github.com/glimmerjs/glimmer-resolver/pull/16

@mixonic @rwjblue

### DIFF
--- a/packages/container/lib/container.js
+++ b/packages/container/lib/container.js
@@ -1,5 +1,6 @@
 /* globals Proxy */
 import { assert } from 'ember-debug';
+import { EMBER_MODULE_UNIFICATION } from 'ember/features';
 import { DEBUG } from 'ember-env-flags';
 import {
   dictionary,
@@ -135,6 +136,10 @@ Container.prototype = {
     return { [OWNER]: this.owner };
   },
 
+  _resolverCacheKey(name, options) {
+    return this.registry.resolverCacheKey(name, options);
+  },
+
   /**
    Given a fullName, return the corresponding factory. The consumer of the factory
    is responsible for the destruction of any factory instances, as there is no
@@ -153,18 +158,28 @@ Container.prototype = {
     assert('fullName must be a proper full name', this.registry.validateFullName(normalizedName));
 
     if (options.source) {
-      normalizedName = this.registry.expandLocalLookup(fullName, options);
+      let expandedFullName = this.registry.expandLocalLookup(fullName, options);
       // if expandLocalLookup returns falsey, we do not support local lookup
-      if (!normalizedName) {
-        return;
+      if (!EMBER_MODULE_UNIFICATION) {
+        if (!expandedFullName) {
+          return;
+        }
+
+        normalizedName = expandedFullName;
+      } else if (expandedFullName) {
+        // with ember-module-unification, if expandLocalLookup returns something,
+        // pass it to the resolve without the source
+        normalizedName = expandedFullName;
+        options = {};
       }
     }
 
-    let cached = this.factoryManagerCache[normalizedName];
+    let cacheKey = this._resolverCacheKey(normalizedName, options);
+    let cached = this.factoryManagerCache[cacheKey];
 
     if (cached !== undefined) { return cached; }
 
-    let factory = this.registry.resolve(normalizedName);
+    let factory = EMBER_MODULE_UNIFICATION ? this.registry.resolve(normalizedName, options) : this.registry.resolve(normalizedName);
 
     if (factory === undefined) {
       return;
@@ -180,7 +195,7 @@ Container.prototype = {
       manager = wrapManagerInDeprecationProxy(manager);
     }
 
-    this.factoryManagerCache[normalizedName] = manager;
+    this.factoryManagerCache[cacheKey] = manager;
     return manager;
   }
 };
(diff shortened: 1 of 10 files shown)


packages/container/tests/container_test.js
```diff
@@ -615,3 +616,44 @@ QUnit.skip('#factoryFor does not add properties to the object being instantiated
...
+    let registry = new Registry();
+    registry.resolve = function(fullName, { source }) {
+      if (fullName === 'component:my-input' && src === source) {
+        return PrivateComponent;
+      }
+    };
+
+    let container = registry.container();
+
+    strictEqual(container.factoryFor('component:my-input', { source: src }).class, PrivateComponent, 'The correct factory was provided');
```
mixonic: Tests are passed `assert` as an argument and should use that over the global assertion methods. For example:

```js
QUnit.test('The container can pass a source to factoryFor', function(assert) {
  /* ... */
  assert.strictEqual(
    container.factoryFor('component:my-input', {source: src}).class, PrivateComponent,
    'The correct factory was provided'
  );
  /* ... */
```

In fact re-using the string `'component:my-input'` instead of re-typing it would clean things up as well, as would using `source` instead of `src` as the source string variable name (then this can be `{source}`).

packages/container/tests/container_test.js
```diff
@@ -615,3 +616,44 @@ QUnit.skip('#factoryFor does not add properties to the object being instantiated
...
+    strictEqual(container.factoryManagerCache[`template:routes/application:component:my-input`].class,
```
mixonic: Instead of asserting the contents of the private cache, can the cache be make `_` private and here you can assert `resolve` or the method on the resolver isn't called a second time? or that it *is* called with a different `source`? Test the behavior, not the state of the cache?

mixonic (changes requested):
Left comments for a few minor items. The testing and feature flag stuff can be improved but the functionality here looks good at a glance. Please do cross-link to the PRs and issues elsewhere.

packages/container/lib/container.js
```diff
@@ -153,18 +159,29 @@ Container.prototype = {
...
+          return;
+        }
+
+        normalizedName = expandedFullName;
+      } else if (expandedFullName) {
+        // with ember-module-unification, if expandLocalLookup returns something,
+        // pass it to the resolve without the source
+        normalizedName = expandedFullName;
+        options = copy(options, true);
+        delete options.source;
```
mixonic: Copying all properties then deleting one is not ideal. It likely [causes the object to go into hash table mode](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#521-objects-that-are-in-hash-table-mode-aka-normalized-objects-dictionary-mode---objects-who-have-a-hash-table-as-a-backing-data-structure-are-not-simple-enumerables) which  is a slower data structure.

It would be ideal to create an object with only the data we want to pass forward. Kind of silly that we are doing all this work since I think there are no other properties on `options` :-p

But please just make an object.  Even `options = {}` would be better if there are no other options to copy.

mixonic (changes requested):

mixonic (changes requested):
Talked with @dgeb, and I think this psuedo code would be correct for us to ensure we get the higher precedence for local/private lookups over global when resolving a pair of things:

```
`
a:/my-app/widgets/x-mark
b:/my-app/widgets/top-level/x-mark
`

let source = 'b:/my-app/widgets/top-level';
let a = di.lookup('a:x-mark', {source})

let pairedB;
if (a) {
  pairedB = di.lookup('b', {source: a})
}

let b;
if (pairedB) {
  b = pairedB;
} else {
  b = di.lookup('b:x-mark', {source})
  if (b) {
    let pairedA = di.lookup('a', {source: b})
    a = pairedA;
  }
}

let paired = [a,b]
return paired;
```

packages/container/lib/container.js
iezer: This is scenario that will never get hit in practice because it would only be executed if `expandedLocalLookup` is implemented and returns something. Are you suggesting that I do `options = {};`? I'm concerned that some upstream caller might add an addition property to `options` and expect it to be passed to the resolver.

mixonic: @iezer Right, `options` as an argument to `lookup` is only a contract with the `lookup` method. We don't document the argument as `resolverOptions` where it is implied the object is passed transparently.

A patch that build the object up by iterating `Object.keys` but skipping `source` would be ok, but really IMO just building the new object (in this case passing an empty object) is fine. Building a whole new object by iterating would be slow, calling the `delete` after copying is slow, and this is fairly performance sensitive code. Doing the fast and API-conservative thing is ok.

iezer: Ok I'll make that change

iezer:
@mixonic I'm not sure I follow that example. When you call `di.lookup` you still have no way of knowing if the lookup found the factory with the source or without.

mixonic (approved):

mixonic:
I've added the commit prefix and an entry to `FEATURES.md` along with some other comments in the code. I think this is gtg as a feature flag.

dgeb (approved):
Thanks for all your hard work @iezer. I think you've walled everything off appropriately behind the feature flag, which should allow us to iterate on the implementation.

PR #15373, opened 2017-06-19
Clarify documentation for `.any` return value

jacobq:
The note about corresponding to the `.some` method is slightly misleading. Though the concept is the same the behavior is different in cases where the callback function does not return `true` or `false`. (Admittedly, the documentation regarding the callback functions specifies that they **should** return `true` or `false` but not that they **must** so people may lazily assume that a truthy or falsey result is good enough.) See the examples below:

```
var a = Ember.A();
a.pushObject(Ember.Object.create());
var anyResult = a.any(x => null)
var someResult = a.some(x => null)
console.log(`Results: some -> ${someResult}; any -> ${anyResult}`); // Results: some -> false; any -> null
```

```
var i = 0, j = 0;
var a = Ember.A();
a.pushObject(Ember.Object.create());
a.pushObject(Ember.Object.create());
var anyResult = a.any(x => i++)
var someResult = a.some(x => j++)
console.log(`Results: some -> ${someResult}; any -> ${anyResult}; i = ${j}; i = ${j}`);
// Results: some -> true; any -> 1; i = 2; i = 2
```

### DIFF
--- a/packages/ember-runtime/lib/mixins/enumerable.js
+++ b/packages/ember-runtime/lib/mixins/enumerable.js
@@ -603,7 +603,9 @@ const Enumerable = Mixin.create({
 
   /**
     Returns `true` if the passed function returns true for any item in the
-    enumeration. This corresponds with the `some()` method in JavaScript 1.6.
+    enumeration. This corresponds with the `some()` method in JavaScript 1.6
+    except that if the callback function returns a value other than `true` or
+    `false` then that value will be returned by this function as well.
 
     The callback method you provide should have the following signature (all
     parameters are optional):


mixonic:
@jacobq thanks for the detailed write up. I'm going to mark this as a duplicate and merge https://github.com/emberjs/ember.js/pull/15372 as it just avoids discussing `some` at all 👍 

Thank you.

jacobq:
@mixonic OK, but would you also then consider replacing **should** with **must** in the part about the evaluation function? If it returns null, for example, then the contract is broken and `.any` is not guaranteed to return a boolean value any more.

mixonic:
@jacobq from the code it actually looks like `true` is not required. It is any truthy value that marks inclusion. I created a PR, please take a look: https://github.com/emberjs/ember.js/pull/15391

jacobq:
@mixonic then could you please explain why I get different results here? Perhaps I'm doing something dumb without knowing it.

https://ember-twiddle.com/a716e4104aa3bf975df2285eb44831c9?openFiles=controllers.application.js%2C

PR #15374, opened 2017-06-20
Decouple test from the default resolver for component_registration_test

karthiicksiva:
Ref #15058

### DIFF
--- a/packages/ember/tests/component_registration_test.js
+++ b/packages/ember/tests/component_registration_test.js
@@ -1,200 +1,198 @@
 import { Controller } from 'ember-runtime';
-import { run } from 'ember-metal';
+import { Component } from 'ember-glimmer';
+import { moduleFor, AutobootApplicationTestCase } from 'internal-test-helpers';
 
-import { Application } from 'ember-application';
-import { Router } from 'ember-routing';
-import { compile } from 'ember-template-compiler';
-import {
-  Component,
-  setTemplates,
-  setTemplate
-} from 'ember-glimmer';
-import { jQuery } from 'ember-views';
+moduleFor('Application Lifecycle - Component Registration', class extends AutobootApplicationTestCase {
 
-let App, appInstance;
+  ['@test The helper becomes the body of the component'](assert) {
+    this.runTask(() => {
+      this.createApplication();
 
-function prepare() {
-  setTemplate('components/expand-it', compile('<p>hello {{yield}}</p>'));
-  setTemplate('application', compile('Hello world {{#expand-it}}world{{/expand-it}}'));
-}
-
-function cleanup() {
-  run(() => {
-    try {
-      if (App) {
-        App.destroy();
-      }
-      App = appInstance = null;
-    } finally {
-      setTemplates({});
-    }
-  });
-}
-
-QUnit.module('Application Lifecycle - Component Registration', {
-  setup: prepare,
-  teardown: cleanup
-});
-
-function boot(callback, startURL = '/') {
-  run(() => {
-    App = Application.create({
-      name: 'App',
-      rootElement: '#qunit-fixture'
+      this.addTemplate('components/expand-it', '<p>hello {{yield}}</p>');
+      this.addTemplate('application', 'Hello world {{#expand-it}}world{{/expand-it}}');
     });
 
-    App.deferReadiness();
+    let text = this.$('div.ember-view > div.ember-view').text().trim();
+    assert.equal(text, 'hello world', 'The component is composed correctly');
+  }
 
-    App.Router = Router.extend({
-      location: 'none'
-    });
+  ['@test If a component is registered, it is used'](assert) {
+    this.runTask(() => {
+      this.createApplication();
 
-    appInstance = App.__deprecatedInstance__;
+      this.addTemplate('components/expand-it', '<p>hello {{yield}}</p>');
+      this.addTemplate('application', `Hello world {{#expand-it}}world{{/expand-it}}`);
 
-    if (callback) { callback(); }
-  });
-
-  let router = appInstance.lookup('router:main');
-
-  run(App, 'advanceReadiness');
-  run(() => router.handleURL(startURL));
-}
-
-QUnit.test('The helper becomes the body of the component', function() {
-  boot();
-  equal(jQuery('div.ember-view > div.ember-view', '#qunit-fixture').text(), 'hello world', 'The component is composed correctly');
-});
-
-QUnit.test('If a component is registered, it is used', function() {
-  boot(() => {
-    appInstance.register('component:expand-it', Component.extend({
-      classNames: 'testing123'
-    }));
-  });
-
-  equal(jQuery('div.testing123', '#qunit-fixture').text(), 'hello world', 'The component is composed correctly');
-});
-
-QUnit.test('Late-registered components can be rendered with custom `layout` property', function() {
-  setTemplate('application', compile('<div id=\'wrapper\'>there goes {{my-hero}}</div>'));
-
-  boot(() => {
-    appInstance.register('component:my-hero', Component.extend({
-      classNames: 'testing123',
-      layout: compile('watch him as he GOES')
-    }));
-  });
-
-  equal(jQuery('#wrapper').text(), 'there goes watch him as he GOES', 'The component is composed correctly');
-});
-
-QUnit.test('Late-registered components can be rendered with template registered on the container', function() {
-  setTemplate('application', compile('<div id=\'wrapper\'>hello world {{sally-rutherford}}-{{#sally-rutherford}}!!!{{/sally-rutherford}}</div>'));
-
-  boot(() => {
-    appInstance.register('template:components/sally-rutherford', compile('funkytowny{{yield}}'));
-    appInstance.register('component:sally-rutherford', Component);
-  });
-
-  equal(jQuery('#wrapper').text(), 'hello world funkytowny-funkytowny!!!', 'The component is composed correctly');
-});
-
-QUnit.test('Late-registered components can be rendered with ONLY the template registered on the container', function() {
-  setTemplate('application', compile('<div id=\'wrapper\'>hello world {{borf-snorlax}}-{{#borf-snorlax}}!!!{{/borf-snorlax}}</div>'));
-
-  boot(() => {
-    appInstance.register('template:components/borf-snorlax', compile('goodfreakingTIMES{{yield}}'));
-  });
-
-  equal(jQuery('#wrapper').text(), 'hello world goodfreakingTIMES-goodfreakingTIMES!!!', 'The component is composed correctly');
-});
-
-QUnit.test('Assigning layoutName to a component should setup the template as a layout', function() {
-  expect(1);
-
-  setTemplate('application', compile('<div id=\'wrapper\'>{{#my-component}}{{text}}{{/my-component}}</div>'));
-  setTemplate('foo-bar-baz', compile('{{text}}-{{yield}}'));
-
-  boot(() => {
-    appInstance.register('controller:application', Controller.extend({
-      'text': 'outer'
-    }));
-
-    appInstance.register('component:my-component', Component.extend({
-      text: 'inner',
-      layoutName: 'foo-bar-baz'
-    }));
-  });
-
-  equal(jQuery('#wrapper').text(), 'inner-outer', 'The component is composed correctly');
-});
-
-QUnit.test('Assigning layoutName and layout to a component should use the `layout` value', function() {
-  expect(1);
-
-  setTemplate('application', compile('<div id=\'wrapper\'>{{#my-component}}{{text}}{{/my-component}}</div>'));
-  setTemplate('foo-bar-baz', compile('No way!'));
-
-  boot(() => {
-    appInstance.register('controller:application', Controller.extend({
-      'text': 'outer'
-    }));
-
-    appInstance.register('component:my-component', Component.extend({
-      text: 'inner',
-      layoutName: 'foo-bar-baz',
-      layout: compile('{{text}}-{{yield}}')
-    }));
-  });
-
-  equal(jQuery('#wrapper').text(), 'inner-outer', 'The component is composed correctly');
-});
-
-QUnit.test('Assigning defaultLayout to a component should set it up as a layout if no layout was found [DEPRECATED]', function() {
-  expect(2);
-
-  setTemplate('application', compile('<div id=\'wrapper\'>{{#my-component}}{{text}}{{/my-component}}</div>'));
-
-  expectDeprecation(() => {
-    boot(() => {
-      appInstance.register('controller:application', Controller.extend({
-        'text': 'outer'
+      this.applicationInstance.register('component:expand-it', Component.extend({
+        classNames: 'testing123'
       }));
+    });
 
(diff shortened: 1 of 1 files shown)


packages/ember/tests/component_registration_test.js
```diff
@@ -1,200 +1,172 @@
...
-      }));
+    expectDeprecation(() => {
+      this.addComponent('my-component', {
+        ComponentClass: Component.extend({
+          text: 'inner',
+          defaultLayout: this.compile('should not see this!')
+        }),
+        template: '{{text}}-{{yield}}'
+      });
+    }, /Specifying `defaultLayout` to .+ is deprecated\./);
```
karthiick:
```js
expectDeprecation(() => {
      this.application.register('component:my-component', Component.extend({
        text: 'inner',
        defaultLayout: this.compile('{{text}}-{{yield}}')
      }));
    }, /Specifying `defaultLayout` to .+ is deprecated\./);
```
The above code fails here. Moving the expectDeprecation to `this.visit` works.

```js
    expectDeprecation(() => {
      this.visit('/').then(() => {
        let text = this.$('#wrapper').text().trim();
        assert.equal(text, 'inner-outer', 'The component is composed correctly');
      });
    }, /Specifying `defaultLayout` to .+ is deprecated\./);
```

I can see, we are calling applicationInstance.visit here. From there, we have called boot(). But i would like to know, when does the component registration takes place or from where?

karthiick:
@mixonic This PR is ready for review.

packages/ember/tests/component_registration_test.js
```diff
@@ -1,200 +1,172 @@
...
 
-    App.deferReadiness();
+moduleFor('Application Lifecycle - Component Registration', class extends ApplicationTestCase {
 
-    App.Router = Router.extend({
-      location: 'none'
+  ['@test The helper becomes the body of the component'](assert) {
+    this.addTemplate('application', 'Hello world {{#expand-it}}world{{/expand-it}}');
+    this.addComponent('expand-it', {
+      template: '<p>hello {{yield}}</p>'
```
karthiick: @mixonic i have some basic questions,

1) whats the difference between addComponent & registerComponent.
2) whats the meaning of skipping the ComponentClass in the above method

Thanks.

mixonic: where is `registerComponent` defined? I'm not sure what you're looking at, but I expect `registerComponent` would register the factory into the registry. There are lots of tests in the Ember suite that tried to avoid the global default resolver by registering all their factories into the registry. The problem is that this doesn't really reflect reality, where the resolver almost always provides a user-level factory into the container.

Adding the `TestResolver` is our way around the default resolver while also avoiding the registry hack used in some Glimmer tests. Eventually it would be nice to update tests using the registry, but it isn't a priority.

`addComponent` here doesn't seem necessary. It is just a convenience method for adding both a template and a component class to the resolver at the same time. You could safely rewrite this as:

```js
this.addTeamplate('component:expand-it', `<p>hello {{yield}}</p>`);
```

mixonic: Ah I'm seeing some of these tests explicitly test registered component vs resolved items. I'll leave some notes.

packages/ember/tests/component_registration_test.js
```diff
@@ -1,200 +1,172 @@
...
-}
-
-QUnit.test('The helper becomes the body of the component', function() {
-  boot();
-  equal(jQuery('div.ember-view > div.ember-view', '#qunit-fixture').text(), 'hello world', 'The component is composed correctly');
-});
-
-QUnit.test('If a component is registered, it is used', function() {
-  boot(() => {
-    appInstance.register('component:expand-it', Component.extend({
```
mixonic: You'll need to use an autoboot application test case for this. The test here is that registering a component onto the application instance inside the runloop where that application is created still loads that component.

```js
moduleFor('Application Lifecycle - Component Registration w/ autoboot', class extends AutobootApplicationTestCase {

  [`@test If a component is registered, it is used'](assert) {
    this.runTask(() => {
      this.createApplication();
      this.addTemplate('application', `Hello world {{#expand-it}}world{{/expand-it}}`);
      this.applicationInstance.register('component:expand-it', Component.extend({
        classNames: 'testing123'
      });
    });
    /* ... assertion ... */
  }

});
```

In this test the application template is added to the resolver, but the component is registered on the application instance.

packages/ember/tests/component_registration_test.js
```diff
@@ -1,200 +1,172 @@
...
-});
+    this.visit('/').then(() => {
+      let text = this.$('div.testing123').text().trim();
+      assert.equal(text, 'hello world', 'The component is composed correctly');
+    });
+  }
+
+  ['@test Late-registered components can be rendered with custom `layout` property'](assert) {
+    this.addTemplate('application', '<div id=\'wrapper\'>there goes {{my-hero}}</div>');
+    this.addComponent('my-hero', {
```
mixonic: per the original test, this should also be registered onto an autobooted application instance

packages/ember/tests/component_registration_test.js
```diff
@@ -1,200 +1,172 @@
...
 
-  boot(() => {
-    appInstance.register('component:my-hero', Component.extend({
-      classNames: 'testing123',
-      layout: compile('watch him as he GOES')
-    }));
-  });
+  ['@test Late-registered components can be rendered with template registered on the container'](assert) {
+    this.addTemplate('application', '<div id=\'wrapper\'>hello world {{sally-rutherford}}-{{#sally-rutherford}}!!!{{/sally-rutherford}}</div>');
+    this.addComponent('sally-rutherford', {
```
mixonic: per the original test, both the template and component class should be registered onto an autobooted application instance

packages/ember/tests/component_registration_test.js
```diff
@@ -1,200 +1,172 @@
...
-QUnit.test('Late-registered components can be rendered with template registered on the container', function() {
-  setTemplate('application', compile('<div id=\'wrapper\'>hello world {{sally-rutherford}}-{{#sally-rutherford}}!!!{{/sally-rutherford}}</div>'));
+  ['@test Late-registered components can be rendered with ONLY the template registered on the container'](assert) {
+    this.addTemplate('application', '<div id=\'wrapper\'>hello world {{borf-snorlax}}-{{#borf-snorlax}}!!!{{/borf-snorlax}}</div>');
 
-  boot(() => {
-    appInstance.register('template:components/sally-rutherford', compile('funkytowny{{yield}}'));
-    appInstance.register('component:sally-rutherford', Component);
-  });
+    this.addComponent('borf-snorlax', {
```
mixonic: per the original test, this should be registered onto an autobooted application instance

packages/ember/tests/component_registration_test.js
```diff
@@ -1,200 +1,172 @@
...
-  setTemplate('application', compile('<div id=\'wrapper\'>hello world {{borf-snorlax}}-{{#borf-snorlax}}!!!{{/borf-snorlax}}</div>'));
+  ['@test Assigning layoutName to a component should setup the template as a layout'](assert) {
+    assert.expect(1);
 
-  boot(() => {
-    appInstance.register('template:components/borf-snorlax', compile('goodfreakingTIMES{{yield}}'));
-  });
+    this.addTemplate('application', '<div id=\'wrapper\'>{{#my-component}}{{text}}{{/my-component}}</div>');
+    this.addTemplate('foo-bar-baz', '{{text}}-{{yield}}');
+    this.add('controller:application', Controller.extend({
```
mixonic: per the original test, this should be registered onto an autobooted application instance. the component `my-component` should be as well.

packages/ember/tests/component_registration_test.js
```diff
@@ -1,200 +1,172 @@
...
-  setTemplate('foo-bar-baz', compile('{{text}}-{{yield}}'));
+  ['@test Assigning layoutName and layout to a component should use the `layout` value'](assert) {
+    assert.expect(1);
 
-  boot(() => {
-    appInstance.register('controller:application', Controller.extend({
-      'text': 'outer'
+    this.addTemplate('application', '<div id=\'wrapper\'>{{#my-component}}{{text}}{{/my-component}}</div>');
+    this.addTemplate('foo-bar-baz', 'No way!');
+    this.add('controller:application', Controller.extend({
```
mixonic: per the original test, this should be registered onto an autobooted application instance.

the `my-component` below should be as well.

packages/ember/tests/component_registration_test.js
```diff
@@ -1,200 +1,172 @@
...
+    assert.expect(2);
 
-  setTemplate('application', compile('<div id=\'wrapper\'>{{#my-component}}{{text}}{{/my-component}}</div>'));
-  setTemplate('foo-bar-baz', compile('No way!'));
+    this.addTemplate('application', '<div id=\'wrapper\'>{{#my-component}}{{text}}{{/my-component}}</div>');
 
-  boot(() => {
-    appInstance.register('controller:application', Controller.extend({
-      'text': 'outer'
+    this.add('controller:application', Controller.extend({
```
mixonic: per the original test, this should be registered onto an autobooted application instance

as should `my-component`. Not that you can register on both `application` and on `applicationInstance` and it does have a subtle difference. We should try to recreate the tests accurately here IMO.

packages/ember/tests/component_registration_test.js
```diff
@@ -1,200 +1,172 @@
...
-
-  setTemplate('application', compile('<div id=\'wrapper\'>{{#my-component}}{{text}}{{/my-component}}</div>'));
-  setTemplate('components/my-component', compile('{{text}}-{{yield}}'));
-
-  expectDeprecation(() => {
-    boot(() => {
-      appInstance.register('controller:application', Controller.extend({
-        'text': 'outer'
-      }));
+    this.add('controller:application', Controller.extend({
```
mixonic: per the original test, this should be registered onto an autobooted application instance. As should `my-component` below.

packages/ember/tests/component_registration_test.js
karthiick: > where is registerComponent defined?

https://github.com/emberjs/ember.js/blob/master/packages/internal-test-helpers/lib/test-cases/abstract-rendering.js#L102

> Ah I'm seeing some of these tests explicitly test registered component vs resolved items. I'll leave some notes.

Thanks.

> addComponent here doesn't seem necessary.
```js
this.addTemplate('component:expand-it', `<p>hello {{yield}}</p>`);
```
This line adds just a template right. i mean, generally without a `Component.extend()`, we cannot create a component right? Am i missing something here?

packages/ember/tests/component_registration_test.js
```diff
@@ -1,200 +1,172 @@
...
+    }, /Specifying `defaultLayout` to .+ is deprecated\./);
+  }
 
-QUnit.test('Using name of component that does not exist', function () {
-  setTemplate('application', compile('<div id=\'wrapper\'>{{#no-good}} {{/no-good}}</div>'));
+  ['@test Using name of component that does not exist'](assert) {
+    this.addTemplate('application', '<div id=\'wrapper\'>{{#no-good}} {{/no-good}}</div>');
 
-  expectAssertion(() => boot(), /.* named "no-good" .*/);
+    expectAssertion(() => this.visit('/'), /.* named "no-good" .*/);
```
mixonic: I expect this is falling prey to https://github.com/emberjs/ember.js/issues/15327 and failing. But if you write it as an autoboot test, it will work just fine. Since the rest of the file should probably be autoboot, making this as well seems good.

If you wanted to add a comment ala https://github.com/emberjs/ember.js/blob/1c720c8d4f5ec85036eff50289d7cd334c96ec37/packages/ember/tests/helpers/link_to_test.js#L1518-L1536 I think that would be good. At least mention the issue number.

mixonic (changes requested):
@karthiick let me know if anything here is unclear. Basically I'd like to hew closer to the original tests which may be testing specific timing or place-of-registration/resolution issues.

packages/ember/tests/component_registration_test.js
mixonic: That link to `abstract-rendering` is a good example of hacking around the resolver issue by using the registry instead. We'd like to avoid that in our work here, and in general in the future as it really doesn't reflect real-world apps. 👍 

Components only need a template *or* a component class. You can disregard the other if you have one of them.

karthiick:
@mixonic This is ready for another look.

packages/ember/tests/component_registration_test.js
```diff
@@ -1,200 +1,216 @@
...
-      location: 'none'
-    });
-
-    appInstance = App.__deprecatedInstance__;
-
-    if (callback) { callback(); }
-  });
-
-  let router = appInstance.lookup('router:main');
+    this.visit('/').then(() => {
```
mixonic: This `visit` is probably not needed, but is harmless

packages/ember/tests/component_registration_test.js
```diff
@@ -1,200 +1,216 @@
...
 
-  boot(() => {
-    appInstance.register('controller:application', Controller.extend({
-      'text': 'outer'
-    }));
+  ['@test Late-registered components can be rendered with custom `layout` property'](assert) {
+    this.runTask(() => {
+      this.createApplication();
+
+      this.addTemplate('application', `<div id=\'wrapper\'>there goes {{my-hero}}</div>`);
```
mixonic: the `'` quotes around `wrapper` should not be escaped.

packages/ember/tests/component_registration_test.js
mixonic: In fact I think many of the visit calls here are not needed. Because this is an "autoboot" test, the app will be booted and rendered immediate after the `runTask` wrapping `createApplication`

packages/ember/tests/component_registration_test.js
```diff
@@ -1,200 +1,216 @@
...
 
-  equal(jQuery('#wrapper').text(), 'inner-outer', 'The component is composed correctly');
-});
+  ['@test Late-registered components can be rendered with template registered on the container'](assert) {
+    this.runTask(() => {
+      this.createApplication();
 
-QUnit.test('Assigning layoutName and layout to a component should use the `layout` value', function() {
-  expect(1);
+      this.addTemplate('application', `<div id=\'wrapper\'>hello world {{sally-rutherford}}-{{#sally-rutherford}}!!!{{/sally-rutherford}}</div>`);
```
mixonic: Also there are quotes here should that should not be escaped.

packages/ember/tests/component_registration_test.js
```diff
@@ -1,200 +1,216 @@
...
-      layout: compile('{{text}}-{{yield}}')
-    }));
-  });
+  ['@test Late-registered components can be rendered with ONLY the template registered on the container'](assert) {
+    this.runTask(() => {
+      this.createApplication();
 
-  equal(jQuery('#wrapper').text(), 'inner-outer', 'The component is composed correctly');
-});
+      this.addTemplate('application', `<div id=\'wrapper\'>hello world {{borf-snorlax}}-{{#borf-snorlax}}!!!{{/borf-snorlax}}</div>`);
```
mixonic: quotes that should not be escaped

mixonic (changes requested):

mixonic (approved):

PR #15379, opened 2017-06-21
cleanup `core-object

bekzod:


packages/ember-runtime/lib/system/core_object.js
```diff
@@ -132,11 +132,7 @@ function makeCtor() {
               let baseValue = this[keyName];
 
               if (baseValue) {
-                if ('function' === typeof baseValue.concat) {
```
bekzod: removed brittle way of checking if array, `makeArray` already contains proper `isArray` check

PR #15387, opened 2017-06-22
use flat array in `dsl`

bekzod:


stefanpenner:
many failures, doing a restart just in-case. But they may be legit

bekzod:
It might yarn.lock issue it might need update

locks:
We are green!

PR #15389, opened 2017-06-22
[BUGFIX beta] cleanup `setPath`

bekzod:
cleaned up `setPath` using simpler logic, which is probably faster as test running a bit faster too

### DIFF
--- a/packages/ember-metal/lib/property_set.js
+++ b/packages/ember-metal/lib/property_set.js
@@ -94,21 +94,16 @@ if (MANDATORY_SETTER) {
 }
 
 function setPath(root, path, value, tolerant) {
-  // get the last part of the path
-  let keyName = path.slice(path.lastIndexOf('.') + 1);
-
-  // get the first part of the part
-  path = (path === keyName) ? keyName : path.slice(0, path.length - (keyName.length + 1));
-
-  // unless the path is this, look up the first part to
-  // get the root
-  if (path !== 'this') {
-    root = getPath(root, path);
-  }
+  let parts = path.split('.');
+  let keyName = parts.pop();
 
   if (!keyName || keyName.length === 0) {
     throw new EmberError('Property set failed: You passed an empty path');
   }
+
+  path = parts.length > 0 ? parts.join('.') : keyName;
+
+  root = getPath(root, path);
 
   if (!root) {
     if (tolerant) {
--- a/packages/ember-metal/lib/property_set.js
+++ b/packages/ember-metal/lib/property_set.js
@@ -97,23 +97,21 @@ function setPath(root, path, value, tolerant) {
   let parts = path.split('.');
   let keyName = parts.pop();
 
-  if (!keyName || keyName.length === 0) {
-    throw new EmberError('Property set failed: You passed an empty path');
-  }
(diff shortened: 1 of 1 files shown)


packages/ember-metal/lib/property_set.js
```diff
@@ -100,16 +100,13 @@ if (MANDATORY_SETTER) {
...
+  let parts = path.split('.');
+  let keyName = parts.pop();
 
   // get the first part of the part
-  path = (path === keyName) ? keyName : path.slice(0, path.length - (keyName.length + 1));
+  path = (path === keyName) ? keyName : parts.join('.');
 
-  // unless the path is this, look up the first part to
-  // get the root
-  if (path !== 'this') {
```
bekzod: there is already assertion to not allow that https://github.com/bekzod/ember.js/blob/57f47143694d0db1197bfec0a582a40defbb0ca9/packages/ember-metal/lib/property_set.js#L43

packages/ember-metal/lib/property_set.js
```diff
@@ -99,22 +99,17 @@ if (MANDATORY_SETTER) {
...
 function setPath(root, path, value, tolerant) {
-  // get the last part of the path
-  let keyName = path.slice(path.lastIndexOf('.') + 1);
-
-  // get the first part of the part
-  path = (path === keyName) ? keyName : path.slice(0, path.length - (keyName.length + 1));
-
-  // unless the path is this, look up the first part to
-  // get the root
-  if (path !== 'this') {
```
bekzod: there is already assertion to not allow that https://github.com/bekzod/ember.js/blob/57f47143694d0db1197bfec0a582a40defbb0ca9/packages/ember-metal/lib/property_set.js#L43

stefanpenner: @krisselden r?

packages/ember-metal/lib/property_set.js
```diff
@@ -99,22 +99,17 @@ if (MANDATORY_SETTER) {
...
-    root = getPath(root, path);
-  }
+  let parts = path.split('.');
+  let keyName = parts.pop();
 
   if (!keyName || keyName.length === 0) {
     throw new EmberError('Property set failed: You passed an empty path');
```
krisselden: we should probably just assert this.

packages/ember-metal/lib/property_set.js
```diff
@@ -99,22 +99,17 @@ if (MANDATORY_SETTER) {
...
   }
 
+  path = parts.length > 0 ? parts.join('.') : keyName;
+
+  root = getPath(root, path);
```
krisselden: we shouldn't mutate arguments, we should use a new var.

packages/ember-metal/lib/property_set.js
krisselden: This is a very good catch, we removed support for this.paths, it was a workaround for capitalized paths back when they meant global (which never fully worked anyway).

krisselden (approved):
thank you for this, I kind of what this to be marked for beta, @stefanpenner thoughts?

stefanpenner:
@bekzod we can likely sneak this into beta, if you could update the commit message to be prefixed with `[BUGFIX beta]` that would be great.

PR #15390, opened 2017-06-22
removed `hasThisCache`

bekzod:
since `hasThisCache` only used `DEBUG` mode

### DIFF
--- a/packages/ember-metal/lib/property_get.js
+++ b/packages/ember-metal/lib/property_get.js
@@ -3,7 +3,7 @@
 */
 
 import { assert } from 'ember-debug';
-import { isPath, hasThis } from './path_cache';
+import { isPath } from './path_cache';
 
 const ALLOWABLE_TYPES = {
   object: true,
@@ -51,7 +51,7 @@ export function get(obj, keyName) {
   assert(`Get must be called with two arguments; an object and a property key`, arguments.length === 2);
   assert(`Cannot call get with '${keyName}' on an undefined object.`, obj !== undefined && obj !== null);
   assert(`The key provided to get must be a string, you passed ${keyName}`, typeof keyName === 'string');
-  assert(`'this' in paths is not supported`, !hasThis(keyName));
+  assert(`'this' in paths is not supported`, keyName.lastIndexOf('this.', 0) !== 0);
   assert('Cannot call `Ember.get` with an empty string', keyName !== '');
 
   let value = obj[keyName];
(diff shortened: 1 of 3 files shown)


packages/ember-metal/lib/property_get.js
```diff
@@ -51,7 +51,7 @@ export function get(obj, keyName) {
   assert(`Get must be called with two arguments; an object and a property key`, arguments.length === 2);
   assert(`Cannot call get with '${keyName}' on an undefined object.`, obj !== undefined && obj !== null);
   assert(`The key provided to get must be a string, you passed ${keyName}`, typeof keyName === 'string');
-  assert(`'this' in paths is not supported`, !hasThis(keyName));
+  assert(`'this' in paths is not supported`, keyName.lastIndexOf('this.', 0) !== 0);
```
stefanpenner: I'm not sure we have `lastIndexOf` support on all the browsers we support:http://caniuse.com/#search=lastIndexOf

In theory we could use a helper function to choose the built-in or fallback to polyfill?

stefanpenner: @krisselden would love your input on this change (From a performance stand point)

bekzod: `lastIndexOf` already have been used here https://github.com/emberjs/ember.js/pull/15390/files#diff-05b465f364bc76dfb813d27102cec7b6L9

stefanpenner: hmmm..... seems funky

rwjblue: Why?  `lastIndexOf` was ES5, we only support ES5 browsers...

stefanpenner: @rwjblue ah have we dropped pre IE11 support?

bekzod: @rwjblue It turns out `String.lastIndexOf` has been around since ES3 and `Array.lastIndexOf ` since ES5

rwjblue: @stefanpenner - I don't understand what you mean. IE9 was a fully ES5 browser...

rwjblue: @stefanpenner - In the link you shared, if you click "show all" you will see IE9 and IE10 listed as supported.

stefanpenner: @rwjblue silly me 📦

PR #15391, opened 2017-06-22
[DOCS] Update enumerable.js

mixonic:
More doc tweaks re: https://github.com/emberjs/ember.js/pull/15373#issuecomment-310418518

### DIFF
--- a/packages/ember-runtime/lib/mixins/enumerable.js
+++ b/packages/ember-runtime/lib/mixins/enumerable.js
@@ -616,8 +616,9 @@ const Enumerable = Mixin.create({
     - `index` is the current index in the iteration.
     - `enumerable` is the enumerable object itself.
 
-    It should return `true` to include the item in the results, `false`
-    otherwise.
+    It must return a truthy value (i.e. `true`) to include an item in the
+    results. Any non-truthy return value will discard the item from the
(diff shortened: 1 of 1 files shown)


locks (approved):

jacobq:
I still don't think that's right. It seems that returning values that are not strictly `true` or `false` can cause the return value of `.any` to be neither `true` nor `false`. See:
https://ember-twiddle.com/a716e4104aa3bf975df2285eb44831c9?openFiles=controllers.application.js%2C

jacobq:
Looking at https://github.com/emberjs/ember.js/blob/b2f3f6ee6f5a0f055c2a84af2a263934f66899b5/packages/ember-runtime/lib/mixins/enumerable.js#L652-L656 it seems that the search ends when the callback returns a truthy value, but the value that is returned is `found` (the value of the callback), not `!!found` (the truthy-ness of said value). Hence it is mandatory that the callback return either `true` or `false` in order for the assumption that `.any` will always return `true` or `false` to hold. Alternatively, you could change the return value to be `!!found`. Performance impact should be negligible.

locks:
@jacobq I'm merging as in, but feel free to send an additional PR with your tweaks.

PR #15392, opened 2017-06-22
Decouple View Instrumentation tests from the global resolver.

patricksrobertson:
References #15058

### DIFF
--- a/packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
+++ b/packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
@@ -140,7 +140,8 @@ moduleFor('The {{link-to}} helper: invoking with query params', class extends Ap
       let theLink = this.$('#the-link');
       assert.equal(theLink.attr('href'), '/about?baz=lol');
 
-      this.click('#the-link');
+      this.runTask(() => this.click('#the-link'));
+
       let aboutController = this.getController('about');
 
       assert.deepEqual(aboutController.getProperties('baz', 'bat'),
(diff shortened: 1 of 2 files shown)


mixonic (approved):

packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
```diff
@@ -474,8 +475,9 @@ moduleFor('The {{link-to}} helper: invoking with query params', class extends Ap
       this.shouldBeActive(assert, '#app-link');
       this.shouldBeActive(assert, '#parent-link');
 
-      this.click('#app-link');
-
+      return this.runTask(() => this.click('#app-link'));
```
mixonic: Using a runloop seems legit.

Ember's event handlers use `run.join`. If there is a runloop the schedule work into it, if there is not they create one. Thus if you have a `this.click()` at the top level of a test (where there is no runloop) the rendering happens synchronously. However if you have a click inside a thennable block (where there is a runloop) the rendering won't happen until after the end of the sync promise chain.

Two ways of addressing this would be:

```js
this.visit('/').then(() => {
  this.click('#app-link');
  return new Ember.RSVP.Promise(resolve => Ember.run.later(resolve, 0));
}).then(() => {
  /* assert rendering after click */
});
```

Or by forcing the click handler to join a nested runloop, ala:

```js
this.visit('/').then(() => {
  this.runTask(() => this.click('#app-link'));
  /* assert rendering after click */
});
```

For now lets call either of these valid. Forcing the nested runloop here seems ok.

---

That said, the promise chaining used twice in this patch is not necessary. Tests pass with the following changed which should be made here:

```diff
diff --git a/packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js b/packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
index b3f5f4961..0fd783744 100644
--- a/packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
+++ b/packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
@@ -140,8 +140,8 @@ moduleFor('The {{link-to}} helper: invoking with query params', class extends Ap
       let theLink = this.$('#the-link');
       assert.equal(theLink.attr('href'), '/about?baz=lol');
 
-      return this.runTask(() => this.click('#the-link'));
-    }).then(() => {
+      this.runTask(() => this.click('#the-link'));
+
       let aboutController = this.getController('about');
 
       assert.deepEqual(aboutController.getProperties('baz', 'bat'),
@@ -475,9 +475,8 @@ moduleFor('The {{link-to}} helper: invoking with query params', class extends Ap
       this.shouldBeActive(assert, '#app-link');
       this.shouldBeActive(assert, '#parent-link');
 
-      return this.runTask(() => this.click('#app-link'));
-    }).then(() => {
-      let router = this.appRouter;
+      this.runTask(() => this.click('#app-link'));
+
       assert.equal(router.get('location.path'), '/parent');
     });
   }
```

mixonic (changes requested):

mixonic (approved):

PR #15398, opened 2017-06-24
Refactor blueprints to eliminate directly accessing Ember properties

bryanhickerson:
Eliminate directly accessing properties of the Ember object in blueprints. When using ember-suave (https://github.com/DockYard/eslint-plugin-ember-suave/blob/master/docs/rules/no-direct-property-access.md) for linting, this results in files that immediately fail linting. Connected to https://github.com/ember-cli/ember-cli-legacy-blueprints/pull/80.


locks (changes requested):
For me, the "noise" introduced by this change far outweighs the potential benefit if you happen to be using `ember-suave`. Mind you, all the incidental blueprint-related documentation (Guides, API) would have to be updated to match.
I would like to also point out that the canary channel of Ember CLI–via [`ember-cli-shims`](https://github.com/ember-cli/ember-cli-shims)– now contains an implementation of [RFC #176](https://github.com/emberjs/rfcs/blob/master/text/0176-javascript-module-api.md), with `@ember/component` style imports. Production-stopping bugs aside, it should follow the natural cycle of canary-beta-release and be available for everyone in one of the next releases.

Turbo87:
I agree with @locks, we should focus on the new module imports instead of forcing everyone to use the sometimes unreadable destructuring

bryanhickerson:
Ah okay, thanks for the info.

PR #15400, opened 2017-06-24
`UNDEFINED` as function not string in `meta`

bekzod:
isn't it better to have `UNDEFINED` as `function` or `object` so it is passed by reference (instead of copying string around) ?

### DIFF
--- a/packages/ember-metal/lib/meta.js
+++ b/packages/ember-metal/lib/meta.js
@@ -1,7 +1,6 @@
 import {
   HAS_NATIVE_WEAKMAP,
   lookupDescriptor,
-  symbol,
   toString
 } from 'ember-utils';
 import { protoMethods as listenerMethods } from './meta_listeners';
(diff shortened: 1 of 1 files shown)


packages/ember-metal/lib/meta.js
```diff
@@ -33,7 +32,7 @@ if (DEBUG) {
 @module ember-metal
 */
 
-export const UNDEFINED = symbol('undefined');
```
stefanpenner: I think we made it a non-function so the property doesn't transition...

bekzod: I assumed that, checking equality by reference is faster than checking equality between strings for example, it maybe consumes less memory because strings are not copied around

SpikedKira: It's not a string though, right? It's a unique object with a string description.  I think we are probably good to leave this as Symbol for future interpreter optimization.

Granted, the transpile may do string compare now, but we shouldn't write code based on how it's transpiled.

mixonic: Landed in https://github.com/emberjs/ember.js/commit/9471527454b30b9a892d6d40f21735365b5a2897. If we're going to change this it should be with some analysis of what v8 is doing with this.

bekzod:
I guess this does not make huge difference so closing it for now

PR #15404, opened 2017-06-24
cleanup `computed`

bekzod:


packages/ember-metal/lib/computed.js
```diff
@@ -531,17 +528,11 @@ ComputedPropertyPrototype.teardown = function(obj, keyName) {
   @return {Ember.ComputedProperty} property descriptor instance
   @public
 */
-export default function computed(func) {
-  let args;
-
-  if (arguments.length > 1) {
-    args = [].slice.call(arguments);
```
bekzod: babel `for loop` toArray is faster than `[].slice.call`

mixonic:
Thanks @bekzod!

PR #15407, opened 2017-06-25
cleanup `computed.teardown`

bekzod:


packages/ember-metal/lib/computed.js
```diff
@@ -426,11 +426,11 @@ ComputedPropertyPrototype._set = function computedPropertySet(obj, keyName, valu
 };
 
 /* called before property is overridden */
-ComputedPropertyPrototype.teardown = function(obj, keyName) {
+ComputedPropertyPrototype.teardown = function(obj, keyName, _meta) {
   if (this._volatile) {
     return;
   }
-  let meta = metaFor(obj);
```
bekzod: it is pure speculation, I think it should be `peekMeta`

stefanpenner: yup this is correct, but we will now need to guard against `meta === null`.

something like
```js
+  let meta = _meta || peekMeta(obj);
 
if ( meta === null) { return; }
```

stefanpenner (changes requested):

packages/ember-metal/lib/computed.js
stefanpenner: another question though is, if teardown isn't public and the only place we call it we already provide meta, then maybe we don't need to do a meta lookup at all? What are your thoughts?

bekzod: sure...doing that

packages/ember-metal/lib/computed.js
```diff
@@ -426,13 +426,12 @@ ComputedPropertyPrototype._set = function computedPropertySet(obj, keyName, valu
 };
 
 /* called before property is overridden */
-ComputedPropertyPrototype.teardown = function(obj, keyName) {
-  if (this._volatile) {
+ComputedPropertyPrototype.teardown = function(obj, keyName, meta) {
+  if (this._volatile || !meta) {
```
stefanpenner: i believe it may be impossible for `meta` to be null here. Or..?

stefanpenner (changes requested):

packages/ember-metal/lib/computed.js
bekzod: I believe too ... so no guard ?

stefanpenner (approved):

stefanpenner:
restarting sauce

locks:
image

PR #15410, opened 2017-06-25
converted `throw`  to `assertion` in `reduce_computed_macros`

bekzod:
since other  macros `sort`, `mapBy` do assertion  and `setDiff` was left behind :)


stefanpenner:
restarting the 1 failed build.

locks:
Seems to be a legit error:

        Failed assertion: setDiff requires two dependent arrays, expected: /requires exactly two dependent arrays/, but was: undefined

PR #15413, opened 2017-06-25
reuse empty func in `ember-debug`

bekzod:


packages/ember-debug/lib/index.js
```diff
@@ -16,19 +16,21 @@ export { default as Error } from './error';
...
 // These are the default production build versions:
-let assert = () => {};
-let info = () => {};
-let warn = () => {};
-let debug = () => {};
-let deprecate = () => {};
-let debugSeal = () => {};
-let debugFreeze = () => {};
-let runInDebug = () => {};
+const noop = () => {};
```
stefanpenner: thoughts on `const noop = assert = info = /*.... etc */`?

stefanpenner (changes requested):

packages/ember-debug/lib/index.js
bekzod: tried, its failed travis, (there was some weirdness to it I forgot), so reverted to initial per line approach

stefanpenner (approved):

PR #15416, opened 2017-06-26
converted bitwise flags into bool flags in `meta`

bekzod:
this is just idea for discussion, and I was motivated by following

using bool flags
* avoids doing extra operations while setting/getting a flag
* is more explicit, easier to debug

using bitwise flags
* probably avoids creating 3 extra props
* does extra work each time you set/get flag
* less explicit

maybe I am missing something 🤔

### DIFF
--- a/packages/ember-metal/lib/meta.js
+++ b/packages/ember-metal/lib/meta.js
@@ -35,12 +35,6 @@ if (DEBUG) {
 
 export const UNDEFINED = symbol('undefined');
 
-// FLAGS
-const SOURCE_DESTROYING = 1 << 1;
-const SOURCE_DESTROYED = 1 << 2;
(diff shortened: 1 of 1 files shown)


krisselden:
this is field loads are more expensive than bitwise ops, integers have different in object storage.  this is a hot area of code.  adding more fields to meta isn't without cost.

stefanpenner:
I think we should prefer bitwise:

* less memory
* less lookups
* more strange/complex (but it is kept isolated)

stefanpenner:
> is more explicit, easier to debug

Is a good point. It may be very useful to have a `meta.debugInfo()` or `meta.inspect()` something that exists in development mode that can be used to explore/inspect during debugging. Thoughts?

PR #15417, opened 2017-06-26
correct `undefined` checks

bekzod:


SpikedKira:
My tests show `typeof x === 'undefined'` are slightly faster than `x === undefined`.
Did you see some different performance numbers?

bekzod:
I think it is not https://jsperf.com/typeof-vs-undefined-check/12 also babel transpiles `x === undefined` to `x === void 0` which is fastest case

SpikedKira:
https://jsperf.com/spikedkira-undefined

bekzod:
@SpikedKira it is still faster :P

SpikedKira:
@bekzod interesting, and possible.  My phone shows faster for undefined but chrome and firefox both are faster at typeof.  I wonder if my phone doesn't have the new interpreter o.O

stefanpenner:
@bekzod / @SpikedKira in general, (as long as there isn't a major issue) the more concise the expression the more comfortable we should be taking that direction.

For example `typeof key === 'undefined'` vs `key === undefined` feels like they should eventually converge to something similar. But `key === undefined` feels more explicit, so I suspect our preference should be `key === undefined` (as long as there isn't a HUGE difference).

PR #15418, opened 2017-06-26
deprecate `targetObject` in `target_action_support`

bekzod:
not sure if this is still valid, and not sure about `id` and `until` values

### DIFF
--- a/packages/ember-runtime/lib/mixins/target_action_support.js
+++ b/packages/ember-runtime/lib/mixins/target_action_support.js
@@ -9,7 +9,19 @@ import {
   Mixin,
   computed
 } from 'ember-metal';
-import { assert } from 'ember-debug';
+import { assert, deprecate } from 'ember-debug';
+
+function deprecateTargetObject() {
+  deprecate(
+    'Using `targetObject` has been deprecated in TargetActionSupport Mixin',
+    false,
+    {
+      id: 'ember-runtime.mixins.target-action-support.targetObject',
+      until: '3.0.0'
+    }
+  );
+}
+
 /**
 `Ember.TargetActionSupport` is a mixin that can be included in a class
 to add a `triggerAction` method with semantics similar to the Handlebars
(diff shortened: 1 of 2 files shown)


rwjblue (commented):
We should also deprecate in the `get` path (changing the `getTarget` method to use `_targetObject` directly).

bekzod:
added depreciation to `get` too

packages/ember-runtime/lib/mixins/target_action_support.js
```diff
@@ -9,7 +9,19 @@ import {
   Mixin,
   computed
 } from 'ember-metal';
-import { assert } from 'ember-debug';
+import { assert, deprecate } from 'ember-debug';
+
+function depricateTargetObject() {
```
runspired: sp. deprecateTargetObject*

packages/ember-runtime/lib/mixins/target_action_support.js
```diff
@@ -9,7 +9,19 @@ import {
...
+  deprecate(
+    'Using `targetObject` has been deprecated in TargetActionSupport Mixin',
+    false,
+    {
+      id: 'ember-runtime.mixins.target-action-support.targetObject',
+      until: '3.0.0'
```
runspired: do we add `since` now for the svelte work, or is that information stored elsewhere?

bekzod: cannot find any usage of `since: ` anywhere

rwjblue: We add it to the features.json file

bekzod: should I add it ? what version it should be ?

krisselden:
We should deprecate the whole mixin.  This isn't used by anything in Ember, it is just exported.

locks:
@bekzod hey! We discussed this deprecation, and we decided that `TargetActionSupport` should be deprecated entirely, like @krisselden suggested. Since it's private API that's not even used by Ember, the deprecation should be until the next LTS, which is 2.16.

bekzod:
@locks hi, thanks for letting know, I think it is better to be handled by person who knows `TargetActionSupport` better, so closing this

PR #15423, opened 2017-06-27
Decouple transitioning CSS classes tests from global resolver.

patricksrobertson:
I did a fair amount of other refactorings here:
* split the class assertions in a 1:1 per selector. I wasn't
particurarly fond of the assertion function that relied on argument
length and found having explicit assertions per selector to be helpful.
* split the two cases into two modules because the test case setup was
different enough to merit it (imo).

refs #15058

### DIFF
--- a/packages/ember/tests/helpers/link_to_test/link_to_transitioning_classes_test.js
+++ b/packages/ember/tests/helpers/link_to_test/link_to_transitioning_classes_test.js
@@ -1,200 +1,244 @@
 import { RSVP } from 'ember-runtime';
-import { Route, NoneLocation } from 'ember-routing';
-import { run, set } from 'ember-metal';
-import { compile } from 'ember-template-compiler';
-import { Application } from 'ember-application';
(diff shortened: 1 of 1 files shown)


packages/ember/tests/helpers/link_to_test/link_to_transitioning_classes_test.js
```diff
@@ -1,200 +1,231 @@
...
+    super();
+
+    this.router.map(function() {
+      this.route('about');
+      this.route('other');
+    });
+
+    this.add('route:about', Route.extend({
+      model() {
+        aboutDefer = RSVP.defer();
```
mixonic: Can you swap these various variables to be properties on the test instead of closed over variables? Thank.

mixonic (changes requested):
Asking for a small cleanup here, but this looks great after that 🙇

mixonic (approved):

PR #15424, opened 2017-06-27
removed redundant check in `removeListener`

bekzod:


packages/ember-metal/lib/events.js
```diff
@@ -126,11 +126,7 @@ export function removeListener(obj, eventName, target, method) {
     target = null;
   }
 
-  metaFor(obj).removeFromListeners(eventName, target, method, (...args) => {
-    if ('function' === typeof obj.didRemoveListener) {
```
bekzod: callback is checked for being `function` inside `meta. removeFromListeners ` already

packages/ember-metal/lib/events.js
```diff
@@ -126,11 +126,7 @@ export function removeListener(obj, eventName, target, method) {
...
-      obj.didRemoveListener(...args);
-    }
-  });
+  metaFor(obj).removeFromListeners(eventName, target, method, obj.didRemoveListener);
```
stefanpenner: this will result in `didRemoveListener` being invoked without a `this`

stefanpenner (changes requested):

packages/ember-metal/lib/events.js
bekzod: oops..... I remade what I did.... it is better that before I think

packages/ember-metal/lib/events.js
```diff
@@ -126,11 +126,9 @@ export function removeListener(obj, eventName, target, method) {
     target = null;
   }
 
-  metaFor(obj).removeFromListeners(eventName, target, method, (...args) => {
-    if ('function' === typeof obj.didRemoveListener) {
-      obj.didRemoveListener(...args);
-    }
-  });
+  let func = ('function' === typeof obj.didRemoveListener) ?
+    obj.didRemoveListener.bind(obj) : ()=> {};
```
bekzod: * avoids creating wrapping function
* does not do `if check` each cycle
* does not convert arguments to array
* does not do work when `obj.didRemoveListener` missing

stefanpenner: bind is heavily optimized in modern, this may be a good idea.

stefanpenner (approved):

krisselden:
we should be using bind, not creating closures.  only reason in the past for not doing it was phantom, I don't think that is a good reason

PR #15434, opened 2017-06-29
Decouple default_resolver_test from defaults

mixonic:
The `default_resolver_test.js` was written under the assumption that the Ember `DefaultResolver` globals resolver would be an application default. This decouples away from that.

Refs: https://github.com/emberjs/ember.js/issues/15058

### DIFF
--- a/packages/internal-test-helpers/lib/module-for.js
+++ b/packages/internal-test-helpers/lib/module-for.js
@@ -1,3 +1,4 @@
+import { RSVP } from 'ember-runtime';
 import applyMixins from './apply-mixins';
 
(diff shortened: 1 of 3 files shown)


packages/internal-test-helpers/lib/module-for.js
```diff
@@ -6,6 +7,9 @@ export default function moduleFor(description, TestClass, ...mixins) {
   QUnit.module(description, {
     setup() {
       context = new TestClass();
+      if (context.beforeEach) {
+        return context.beforeEach();
```
mixonic: Permits tests to return a promise from their `beforeEach` method.

PR #15436, opened 2017-06-29
use `peekMeta` in `watchedEvents`

bekzod:


packages/ember-metal/lib/watch_key.js
```diff
@@ -87,7 +87,7 @@ export function unwatchKey(obj, keyName, _meta) {
   let meta = _meta || peekMeta(obj);
 
   // do nothing of this object has already been destroyed
-  if (meta === undefined || meta.isSourceDestroyed()) { return; }
```
bekzod: since meta can be `null` too

packages/ember-metal/lib/events.js
```diff
@@ -212,7 +211,7 @@ export function sendEvent(obj, eventName, params, actions, _meta) {
                      meta.matchingListeners(eventName);
   }
 
-  if (actions === undefined || actions.length === 0) { return; }
```
bekzod: to be consistent so `sendEvent` returns `true/false` only

stefanpenner (approved):

stefanpenner (approved):

stefanpenner:
restarting sauce. Man the Edge browser update recently is really trolling us with flakyness.

PR #15438, opened 2017-06-30
cleanup `import/export` in `watching` and `properties`

bekzod:


packages/ember-metal/lib/watching.js
```diff
@@ -31,16 +31,14 @@ import {
   @param obj
   @param {String} _keyPath
 */
-function watch(obj, _keyPath, m) {
-  if (!isPath(_keyPath)) {
-    watchKey(obj, _keyPath, m);
-  } else {
+export function watch(obj, _keyPath, m) {
+  if (isPath(_keyPath)) {
```
locks: 👌

packages/ember-metal/lib/property_events.js
```diff
@@ -110,8 +105,7 @@ function propertyDidChange(obj, keyName, _meta) {
     obj[PROPERTY_DID_CHANGE](keyName);
   }
 
-  if (hasMeta) {
-    if (meta.isSourceDestroying()) { return; }
+  if (hasMeta && !meta.isSourceDestroying()) {
```
locks: this will no longer eagerly return, does that not change the behaviour of the function?

bekzod: oops good catch thanks !

PR #15439, opened 2017-06-30
[DOC beta] Fix wording on private Routing service

locks:
I need confirmation on the channel this should go to.


locks:
Router service is now enabled by default on beta :shipit:

PR #15444, opened 2017-07-02
meta iterators refactor

bekzod:
refactored `_getInherited ` and `for each iterators` to accept list of keys

### DIFF
--- a/packages/ember-metal/lib/meta.js
+++ b/packages/ember-metal/lib/meta.js
@@ -175,6 +175,38 @@ export class Meta {
     return this[key] || (this[key] = Object.create(null));
   }
 
+  _forEachInherited() {
+    let pointer = this;
+    let keys = arguments;
+    let len = keys.length - 1;
+    let callback = arguments[len];
+    let value, key, i;
+    let seen = Object.create(null);
+
+    while (pointer !== undefined) {
+      value = pointer;
+      i = 0;
+      for (; i < len; i++) {
+        key = keys[i];
+        value = value[key];
+        if (value === undefined) {
+          break;
+        }
+      }
+
+      if (i === len) {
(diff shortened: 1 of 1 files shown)


packages/ember-metal/lib/meta.js
```diff
@@ -183,27 +183,55 @@ export class Meta {
     return this[key] || (this[key] = Object.create(null));
   }
 
-  _getInherited(key) {
+  _forEachInherited() {
```
bekzod: accepts list of keys, when `obj.prop1.prop2.prop3` `_forEachInherited('obj','prop1','prop2','prop3', fn)` will call fn with value of `prop3` with current meta and its parent

packages/ember-metal/lib/meta.js
```diff
@@ -183,27 +183,55 @@ export class Meta {
...
+          }
+        }
       }
+
       pointer = pointer.parent;
     }
   }
 
-  _findInherited(key, subkey) {
+  _findInherited() {
```
bekzod: accepts list of keys, when `obj.prop1.prop2.prop3` `_findInherited('obj','prop1','prop2','prop3')` will find first `prop3` property that has `not undefined` value and returns it

krisselden:
what is the motivation?

krisselden:
Also, this has to be benchmarked using a macro bench, this code though for sure can use some cleanup.  I'm surprised seen is using Object.create(null).

bekzod:
motivation was just consolidate repeating logic, `Object.create(null)` was used before so kept it, will benchmark it

bekzod:
benchmark: https://cl.ly/3T3X3a0V2T2O/results.pdf

bekzod:
closing since it performed bad in benchmarks

PR #15448, opened 2017-07-03
Decouple substate tests from the global resolver.

patricksrobertson:
I've moved this into three different test modules:

* A default state that relies on the visit API.
* A nested route module that has shared setup for roughly half the tests
* An AutobootTestCase module for dealing with the one test case that
cannot be worked around.

I've removed the step testing function. This seemed to be more important
for debugging the authoring of the original sets of tests instead of
actually asserting code.

The instance of 'BRO' has been replaced with 'TURTLE' because I hate fun
and value inclusivity.

Refs #15058

### DIFF
--- a/packages/ember/tests/routing/substates_test.js
+++ b/packages/ember/tests/routing/substates_test.js
@@ -1,1042 +1,1016 @@
-import { RSVP, Controller } from 'ember-runtime';
-import { Route, NoneLocation } from 'ember-routing';
-import { run } from 'ember-metal';
-import { compile } from 'ember-template-compiler';
-import { Application, Resolver } from 'ember-application';
-import { jQuery } from 'ember-views';
-import { setTemplates, setTemplate } from 'ember-glimmer';
+import { RSVP } from 'ember-runtime';
+import { Route } from 'ember-routing';
+import {
+  moduleFor,
+  ApplicationTestCase,
+  AutobootApplicationTestCase } from 'internal-test-helpers';
 
-let Router, App, templates, router, container, counter;
+  let counter;
 
-function step(expectedValue, description) {
-  equal(counter, expectedValue, 'Step ' + expectedValue + ': ' + description);
-  counter++;
-}
-
-function bootApplication(startingURL) {
-  for (let name in templates) {
-    setTemplate(name, compile(templates[name]));
+  function step(expectedValue, description) {
+    equal(counter, expectedValue, 'Step ' + expectedValue + ': ' + description);
+    counter++;
   }
 
-  if (startingURL) {
-    NoneLocation.reopen({
-      path: startingURL
-    });
-  }
-
-  startingURL = startingURL || '';
-  router = container.lookup('router:main');
-  run(App, 'advanceReadiness');
-}
-
-QUnit.module('Loading/Error Substates', {
-  setup() {
+moduleFor('Loading/Error Substates', class extends ApplicationTestCase {
+  constructor() {
+    super();
     counter = 1;
 
-    run(function() {
-      App = Application.create({
-        name: 'App',
-        rootElement: '#qunit-fixture',
-        // fake a modules resolver
-        Resolver: Resolver.extend({ moduleBasedResolver: true })
-      });
+    this.addTemplate('application', `<div id="app">{{outlet}}</div>`);
+    this.addTemplate('index', 'INDEX');
+  }
 
-      App.deferReadiness();
+  getController(name) {
+    return this.applicationInstance.lookup(`controller:${name}`);
+  }
 
-      App.Router.reopen({
-        location: 'none'
-      });
+  get currentPath() {
+    return this.getController('application').get('currentPath');
+  }
 
-      Router = App.Router;
+  ['@test Slow promise from a child route of application enters nested loading state'](assert) {
+    let turtleDeferred = RSVP.defer();
 
-      container = App.__container__;
-
-      templates = {
-        application: '<div id="app">{{outlet}}</div>',
-        index: 'INDEX',
-        loading: 'LOADING',
-        bro: 'BRO',
-        sis: 'SIS'
-      };
-    });
-  },
-
-  teardown() {
-    run(function() {
-      App.destroy();
-      App = null;
-
-      setTemplates({});
+    this.router.map(function() {
+      this.route('turtle');
     });
 
-    NoneLocation.reopen({
-      path: ''
+    this.add('route:application', Route.extend({
+      setupController() {
+        step(2, 'ApplicationRoute#setupController');
+      }
+    }));
+
+    this.add('route:turtle', Route.extend({
+      model() {
+        step(1, 'TurtleRoute#model');
+        return turtleDeferred.promise;
+      }
+    }));
+    this.addTemplate('turtle', 'TURTLE');
+    this.addTemplate('loading', 'LOADING');
+
+    let promise = this.visit('/turtle').then(() => {
+      text = this.$('#app').text();
+      assert.equal(
+        text,
+        'TURTLE',
+        `turtle template has loaded and replaced the loading template`
+      );
+    });
+
+    let text = this.$('#app').text();
+    assert.equal(
+      text,
+      'LOADING',
+      `The Loading template is nested in application template's outlet`
+    );
+
+    turtleDeferred.resolve();
+    return promise;
+  }
+
+  [`@test Slow promises returned from ApplicationRoute#model don't enter LoadingRoute`](assert) {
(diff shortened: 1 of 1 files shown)


packages/ember/tests/routing/substates_test.js
```diff
@@ -1,1042 +1,880 @@
...
-  let broModel = {};
-  let broDeferred = RSVP.defer();
-
-  Router.map(function() {
-    this.route('bro');
-  });
-
-  App.ApplicationRoute = Route.extend({
-    setupController() {
-      step(2, 'ApplicationRoute#setup');
```
mixonic: Isn't some coverage lost by dropping these `step` assertions for what hooks fire in what order?

packages/ember/tests/routing/substates_test.js
```diff
@@ -1,1042 +1,880 @@
...
+      }
+    }));
 
-  let deferred = RSVP.defer();
-  App.FooBarRoute = Route.extend({
-    model() {
-      return deferred.promise;
-    }
-  });
+    this.visit('/grandma/mom/sally');
```
mixonic: Please prefer the promise version of `visit`:

```js
return this.visit('/grandma/mom/sally').then(() => {
  let text = this.$('#app').text();
  /* ... */
});
```

packages/ember/tests/routing/substates_test.js
```diff
@@ -1,1042 +1,880 @@
...
+      assert.equal(text, 'DUMMY', `dummy template has been rendered`);
+    });
+  }
 
-  App.IndexRoute = Route.extend({
-    model() {
-      return indexDeferred.promise;
-    }
-  });
+  ['@test Slow promises returned from ApplicationRoute#model enter ApplicationLoadingRoute if present - NEW'](assert) {
```
mixonic: Isn't this the same test?

* https://github.com/emberjs/ember.js/pull/15448/files#diff-a6a3e15b340e41ccc770054658e38485L733

Why `NEW` in the name?

packages/ember/tests/routing/substates_test.js
```diff
@@ -1,1042 +1,880 @@
...
-  ok(appController.get('currentPath') !== 'loading', 'loading state not entered');
+    this.visit('/');
+    assert.ok(loadingRouteEntered, 'ApplicationLoadingRoute was entered');
 
-  run(indexDeferred, 'resolve', {});
-  equal(jQuery('#app', '#qunit-fixture').text(), 'INDEX');
-});
+    this.runTask(() => appDeferred.resolve());
 
-QUnit.test('Enter loading route if only LoadingRoute defined', function() {
```
mixonic: Is this test missing in the refactored file?

packages/ember/tests/routing/substates_test.js
```diff
@@ -1,1042 +1,880 @@
...
-  }, function(err) { return err.msg === 'did it broke?'; });
+    return this.visit('/grandma/mom/sally').then(() => {
+      let appController = this.getController('application');
+      assert.equal(appController.get('currentPath'), 'grandma.mom.sally',
+        'Initial route fully loaded'
+      );
 
-  equal(jQuery('#app', '#qunit-fixture').text(), 'FOOBAR ERROR: did it broke?', 'foo.bar_error was entered (as opposed to something like foo/foo/bar_error)');
-});
+      this.visit('/grandma/smells');
```
mixonic: Should return the promise here and chain.

```js
  return this.visit('/grandma/smells');
}).then(() => {
  let appController = this.getController('application');
  assert.equal(appController.get('currentPath'), 'grandma.loading', /* ... */
```

packages/ember/tests/routing/substates_test.js
```diff
@@ -1,1042 +1,880 @@
...
+        this.addTemplate('application_error', `
+          <p id="toplevel-error">TOPLEVEL ERROR: {{model.msg}}</p>
+        `);
+      })
+    }, (err) => err.msg === 'BAD NEWS BEARS', 'it went poorly');
+
+    let text = this.$('#toplevel-error').text();
+    assert.equal(text, 'TOPLEVEL ERROR: BAD NEWS BEARS',
+      'toplevel error rendered'
+    );
```
mixonic: Doesn't the original test attempt a `transitionTo` on the router after this?

* https://github.com/emberjs/ember.js/pull/15448/files#diff-a6a3e15b340e41ccc770054658e38485L1039

mixonic (changes requested):
One or two things I would appreciate your thoughts on. Great start on a large migration, thank you.

I'd push your edits even further and swap `smells` for `cats` ;-)

packages/ember/tests/routing/substates_test.js
```diff
@@ -1,1042 +1,880 @@
...
+  [`@test Loading actions bubble to root but don't enter substates above pivot `](assert) {
+    let sallyDeferred = RSVP.defer();
+    let smellsDeferred = RSVP.defer();
 
-QUnit.test('Slow promises returned from ApplicationRoute#model enter application_loading if template present', function() {
-  expect(3);
+    this.add('route:application', Route.extend({
+      actions: {
+        loading(transition, route) {
+          ok(true, 'loading action received on ApplicationRoute');
```
karthiick: missed assert

packages/ember/tests/routing/substates_test.js
patricksrobertson: The problem is that the visit promise will not resolve until the model hook promise is resolved.

Is there another pattern to follow ie:
```js
let visit = this.visit('grandma/mom/sally');
// assertion before model is resolved
model.resolve();
visit.then(() => {
  //assertion after model hook is resolved
});
```

packages/ember/tests/routing/substates_test.js
patricksrobertson: same thing here as above ^, this relies on assertions before and after promises are resolved on `model()` hooks in routes.

packages/ember/tests/routing/substates_test.js
patricksrobertson: I can add that back in, my $0.02 is that you're testing behavior well tested in the basic routing test with that second assertion at the cost of pretty gnarly testing setup.

packages/ember/tests/routing/substates_test.js
patricksrobertson: Fair point!

packages/ember/tests/routing/substates_test.js
patricksrobertson: Shoot. The `- NEW` was when I was migrating tests over I could tell which of the two copies were failing. Will clean it out.

packages/ember/tests/routing/substates_test.js
patricksrobertson: https://github.com/emberjs/ember.js/pull/15448/files#diff-a6a3e15b340e41ccc770054658e38485R106

packages/ember/tests/routing/substates_test.js
mixonic: Yeah that would works well. The thing I was reaching for was to chain the promises, and your example does that. Another alternative:

```js
let promise = this.visit('/grandma/mom/sally').then(() => {
  let text = this.$('#app').text();
});
model.resolve();
return promise;
```

packages/ember/tests/routing/substates_test.js
mixonic: Yeah but it also is asserting that you can transition after an error substate. I'd prefer to add it back in.

packages/ember/tests/routing/substates_test.js
```diff
@@ -1,1042 +1,951 @@
...
 
-  run(router, 'transitionTo', 'grandma.smells');
-  equal(appController.get('currentPath'), 'grandma.mom.sally', 'still in initial state because the only loading state is above the pivot route');
+        assert.equal(text, 'YAY', 'foo.bar.index fully loaded');
+      });
+      let text = this.$('#app').text();
 
-  run(smellsDeferred, 'resolve', {});
+      assert.equal(text, 'FOOBAR LOADING', `
+        foo.bar_loading was entered (as opposed to something like
```
mixonic: Seems odd for an assertion message to start with a newline. Would prefer:

```js
  assert.equal(
    text, 'FOOBAR LOADING',
    `foo.bar_loading was entered (as opposed to something like foo/foo/bar_loading`
  );
```

mixonic (changes requested):
Just some small cleanup and I think this is good to go.

patricksrobertson:
@mixonic addressed!

mixonic (approved):

PR #15449, opened 2017-07-04
rename `_watching` => `_isWatching` in `ChainNode`

bekzod:
to make it less confusing, since `_watching` in meta contains hash of watch counts and in `ChainNode ` its boolean

### DIFF
--- a/packages/ember-metal/lib/chains.js
+++ b/packages/ember-metal/lib/chains.js
@@ -137,20 +137,20 @@ class ChainNode {
     this._parent = parent;
     this._key    = key;
 
-    // _watching is true when calling get(this._parent, this._key) will
-    // return the value of this node.
-    //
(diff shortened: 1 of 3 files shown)


locks:
Can you rebase?

packages/ember-metal/lib/chains.js
```diff
@@ -137,20 +137,20 @@ class ChainNode {
...
-    let isWatching = this._watching = (value === undefined);
-
     this._chains = undefined;
     this._object = undefined;
     this.count = 0;
 
     this._value = value;
     this._paths = undefined;
+
+    // _watching is true when calling get(this._parent, this._key) will
```
mmun: should be _isWatching

mmun: nm, not a big deal.

PR #15451, opened 2017-07-04
remove `chainsFor` in `watch_path`

bekzod:


packages/ember-metal/lib/watch_path.js
```diff
@@ -22,7 +16,7 @@ export function watchPath(obj, keyPath, meta) {
 
   m.writeWatching(keyPath, counter + 1);
```
bekzod: since here we are incrementing meta

PR #15455, opened 2017-07-05
small cleanup in `metal/chains`

bekzod:


packages/ember-metal/lib/chains.js
```diff
@@ -185,23 +185,19 @@ class ChainNode {
...
-  }
+    }
     return ret;
   }
 
   // called on the root node of a chain to setup watchers on the specified
   // path.
   add(path) {
-    let paths = this._paths || (this._paths = {});
+    let paths = this._paths || (this._paths = Object.create(null));
```
krisselden: Needs to be measured. Object.create null puts object into dictionary (slow mode) which is fast if large but I think this tends to stay small.

bekzod: will revert this I think not super critical

krisselden (approved):

PR #15458, opened 2017-07-05
Refactor feature flag conditionals to appease babel-plugin-debug-macros.

rwjblue:
Due to issues with babel-plugin-debug-macros, these conditionals had to be rewritten.  

The fundamental issue upstream is that we [only check for very specific types of statements](https://github.com/chadhietala/babel-plugin-debug-macros/blob/11b5d056071bc2bff9062b65ff9c8db13ce8156c/src/lib/utils/macros.js#L58-L63) to replace with `true` or `false`.  I will create an issue over there to track fixing this properly, but for now this PR unblocks the release of 2.15 beta.

### DIFF
--- a/packages/container/lib/container.js
+++ b/packages/container/lib/container.js
@@ -160,17 +160,19 @@ Container.prototype = {
     if (options.source) {
       let expandedFullName = this.registry.expandLocalLookup(fullName, options);
       // if expandLocalLookup returns falsey, we do not support local lookup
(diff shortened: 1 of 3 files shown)


rwjblue:
Cross linking https://github.com/chadhietala/babel-plugin-debug-macros/issues/43 for future reference.

mixonic (approved):
LGTM. Thank you @rwjblue

mmun:
These flagging styles are supported now.

PR #15469, opened 2017-07-07
[BUGFIX lts] Ensure event dispatcher is a singleton

### LINKED ISSUE #440: Tagless components issue within routeless engines
It seems that there are some event binding issues when using tagless components within a routeless engine.

Repro: https://github.com/mike183/engines-tagless-component-issue

**Steps to reproduce**
1. Clone the above repo
2. `yarn`
3. `ember serve`
4. Navigate to `http://localhost:4200/`
5. Click `Route #1` link at the top of the page
6. Click the 'Click Me' button several times and observe the clicks counter increment
7. Click `Route #2` link at the top of the page
8. Click `Route #1` link at the top of the page
9. Click the 'Click Me' button again, you'll notice that it no longer increments the counter

When inspecting the 'Click Me' button in the Chrome inspector, it appears that events aren't being registered after the tagless component has been rendered, so I'm not exactly sure where the issue could be.

If you comment out [this line](https://github.com/mike183/engines-tagless-component-issue/blob/master/lib/routeless-engine/addon/components/tagless-component.js#L5), making the component no longer tagless, everything begins to work as expected.

mike183:
This PR fixes https://github.com/ember-engines/ember-engines/issues/440

I have added a test that follows the reproduction steps in the issue above, though the test isn't failing without the change, so I'm not sure if either I'm doing something wrong, or if there is something about the test environment which is masking the issue.

FWIW I have verified that this fix is working in the reproduction app via `yarn link`.

cc/ @rwjblue

### DIFF
--- a/packages/ember-glimmer/tests/integration/mount-test.js
+++ b/packages/ember-glimmer/tests/integration/mount-test.js
@@ -188,6 +188,48 @@ moduleFor('{{mount}} test', class extends ApplicationTest {
     });
   }
 
+  ['@test it declares the event dispatcher as a singleton']() {
+    this.router.map(function() {
+      this.route('engine-event-dispatcher-singleton');
+    });
+
+    let controller;
+    let component;
+
+    this.add('controller:engine-event-dispatcher-singleton', Controller.extend({
+      init() {
+        this._super(...arguments);
+        controller = this;
+      }
+    }));
+    this.addTemplate('engine-event-dispatcher-singleton', '{{mount "foo"}}');
+
+    this.add('engine:foo', Engine.extend({
+      router: null,
+      init() {
+        this._super(...arguments);
+        this.register('template:application', compile('<h2>Foo Engine: {{tagless-component}}</h2>', { moduleName: 'application' }));
+        this.register('component:tagless-component', Component.extend({
+          tagName: "",
+          init() {
+            this._super(...arguments);
+            component = this;
+          }
+        }));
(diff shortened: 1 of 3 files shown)


mike183:
Can somebody give CI a kick please 🙂

packages/ember-glimmer/tests/integration/mount-test.js
```diff
@@ -188,6 +188,56 @@ moduleFor('{{mount}} test', class extends ApplicationTest {
...
+    }));
+
+    return this.visit('/engine-teardown').then(() => {
+      this.assertComponentElement(this.firstChild, { content: 'Clicks: 0 <button id="increment-engine-total-clicks">Increment</button>' });
+
+      this.runTask(() => this.$("#increment-engine-total-clicks").click());
+
+      this.assertComponentElement(this.firstChild, { content: 'Clicks: 1 <button id="increment-engine-total-clicks">Increment</button>' });
+
+      return this.visit('/engine-teardown-engine').then(() => {
```
rwjblue: Can you swap this to using links that are clicked (instead of this.visit which I believe creates a new application instance)? Might help repro the original issue without the fix...

mike183:
Refactored the test to avoid multiple calls to `this.visit`, but I still can't seem to reproduce the issue in the test suite...

mike183:
So I managed to make a failing test for this 🎉

This should be ready for another review assuming CI is green once it finishes.

rwjblue:
Thank you!!

PR #15475, opened 2017-07-07
Move `accumulateListeners` to the only place where it's used.

cibernox:
I found that this private function was only used in one place within Ember and exactly 0 times in the wild, so I made it a local function in the file where it's used.

This avoids a few import/export for no reason. Also improved so we void peeking a
meta that we already have in the context.

### DIFF
--- a/packages/ember-metal/lib/property_events.js
+++ b/packages/ember-metal/lib/property_events.js
@@ -3,8 +3,7 @@ import {
   peekMeta
 } from './meta';
 import {
-  sendEvent,
-  accumulateListeners
+  sendEvent
 } from './events';
 import {
   markObjectAsDirty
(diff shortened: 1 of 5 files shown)


stefanpenner:
:shipit:

cibernox:
@stefanpenner can you restart sauce?

stefanpenner:
restarting sauce

cibernox:
@stefanpenner again?

stefanpenner:
restarting

cibernox:
Is it normal that it fails so much or should I be worried?

stefanpenner:
> Is it normal that it fails so much or should I be worried?

its been brittle for the last more or two, I haven't had a chance to investigate.

stefanpenner:
@rwjblue should we just merge, code looks good...

bekzod:
@stefanpenner as promised looked into IE instability, not sure what is the cause but to be safe just upgraded `web-driver` and `sauce-connect-launcher` in `saucie` (`ember-cli-sauce`) will push downstream once there is release in `saucie` https://github.com/johanneswuerbach/saucie/pull/71

packages/ember-metal/lib/property_events.js
```diff
@@ -252,14 +251,49 @@ function changeProperties(callback, binding) {
...
+    if (target === array[i] && method === array[i + 1]) {
+      index = i;
+      break;
+    }
+  }
+  return index;
+}
+
+function accumulateListeners(obj, eventName, otherActions, meta) {
+  if (meta === null || meta === undefined) { return; }
```
bekzod: you can even remove meta existence check here since its always passed and its private api

cibernox: but it can be nil, can't it?

bekzod: No it can't, in particular use case of 'accummulateListeners' meta tha you are passing always not nil

cibernox: Updated.

cibernox:
Oh, sauce is green!

stefanpenner:
@cibernox merge merge merge!

Issue #14739, opened 2016-12-19
Glimmer 2 does not read properties on functions


kturney:
We encountered this while upgrading a project that uses https://github.com/LocusEnergy/ember-d3-helpers to Ember 2.10 and trying to bind to properties on D3 scales.

Twiddle: https://ember-twiddle.com/8bd17be80e30f6149bcdf644f3d1f68d?openFiles=templates.application.hbs%2C

If you change the Ember version in the twiddle to 2.9, the property renders as expected.

pr with failing test: https://github.com/emberjs/ember.js/pull/14729

chancancode:
I think I know how to fix this, but I am on vacation. If anyone wants to look into it there are a few places where we do `if (typeof foo === "object")` that should include functions as well. I believe it probably needs both a glimmer fix and an ember fix to cover all the cases.

chancancode:
PrimitiveReference and such

Serabe:
@locks, any update on this?

mixonic:
afaict this is still an issue.

locks:
This should be fixed by https://github.com/emberjs/ember.js/pull/15535.

PR #15482, opened 2017-07-09
enable templates read a property from a function

bekzod:
issue: https://github.com/emberjs/ember.js/issues/14739
broken test: #14729

### DIFF
--- a/packages/ember-glimmer/lib/utils/references.js
+++ b/packages/ember-glimmer/lib/utils/references.js
@@ -398,7 +398,7 @@ export class InternalHelperReference extends CachedReference {
 // @implements PathReference
 export class UnboundReference extends ConstReference {
   static create(value) {
-    if (typeof value === 'object' && value !== null || typeof result === 'function') {
+    if (typeof value === 'object' && value !== null || typeof value === 'function') {
       return new UnboundReference(value);
     } else {
       return PrimitiveReference.create(value);
--- a/packages/ember-glimmer/tests/integration/content-test.js
+++ b/packages/ember-glimmer/tests/integration/content-test.js
@@ -495,17 +495,17 @@ class DynamicContentTest extends RenderingTest {
 
     this.assertStableRerender();
 
-    // this.runTask(() => set(func, 'aProp', 'still a property on a function'));
-    // this.assertContent('still a property on a function');
-    // this.assertInvariants();
+    this.runTask(() => set(func, 'aProp', 'still a property on a function'));
+    this.assertContent('still a property on a function');
+    this.assertInvariants();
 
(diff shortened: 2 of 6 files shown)


packages/ember-glimmer/lib/utils/references.js
```diff
@@ -220,7 +220,7 @@ export class NestedPropertyReference extends PropertyReference {
       return parentValue.length;
     }
 
-    if (typeof parentValue === 'object' && parentValue) {
+    if ((typeof parentValue === 'object' || typeof parentValue === 'function') && parentValue) {
```
krisselden: Falsy check should have been !== null and it needs to stay together with the object check. Same with the following. Don't put the function check in between it's confusing it's either an object that isn't null or a function

krisselden (changes requested):
Needs a test.  Also, was this a regression?

locks:
@krisselden according to the original issue, this was working on 2.9.

packages/ember-glimmer/tests/integration/content-test.js
```diff
@@ -484,6 +484,30 @@ class DynamicContentTest extends RenderingTest {
...
+    let func = () => {};
+    func.aProp = 'this is a property on a function';
+
+    this.renderPath('func.aProp', { func });
+
+    this.assertContent('this is a property on a function');
+
+    this.assertStableRerender();
+
+    this.runTask(() => set(func, 'aProp', 'still a property on a function'));
```
krisselden: this is a new feature, the regression was rendering a prop on a function not using Ember.set() to update a prop on a function. https://ember-twiddle.com/c35628b179d00989b7469f84f1d08558?openFiles=controllers.application.js%2C I tried several versions of ember, props on functions were rendered as constant.

krisselden: Also, I don't think the `watchKey` and `watchPath` is needed for glimmer, just `tagForProperty`

Since this is a regression, it has to be back ported eventually to LTS, can you please separate out the `set(func, 'aProp', ...)` updating and the `tagForProperty` into a separate commit?   I don't mind adding that capability, just don't want to do it casually in a bug fix for an LTS regression.

krisselden (changes requested):

packages/ember-glimmer/tests/integration/content-test.js
bekzod: sure will do

krisselden:
I updated this to be focused on just the functionality that regressed. https://github.com/emberjs/ember.js/compare/pr-15482

locks:
@bekzod can you resolve the conflict?

PR #15485, opened 2017-07-10
cleanup `if-unless`

bekzod:
positional arguments list already returns `UNDEFINED_REFERENCE` for
missing positional argument, so no need to replace it with null
then replace it with `UNDEFINED_REFERENCE`

https://github.com/glimmerjs/glimmer-vm/blob/master/packages/@glimmer/runtime/lib/vm/arguments.ts#L159

### DIFF
--- a/packages/ember-glimmer/lib/helpers/action.js
+++ b/packages/ember-glimmer/lib/helpers/action.js
@@ -272,7 +272,6 @@ export default function(vm, args) {
   let { named, positional } = args;
 
   let capturedArgs = positional.capture();
-  let { references } = capturedArgs;
 
   // The first two argument slots are reserved.
   // pos[0] is the context (or `this`)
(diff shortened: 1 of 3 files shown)


packages/ember-glimmer/lib/helpers/action.js
```diff
@@ -272,7 +272,6 @@ export default function(vm, args) {
   let { named, positional } = args;
 
   let capturedArgs = positional.capture();
-  let { references } = capturedArgs;
```
bekzod: unused var

packages/ember-metal/lib/property_events.js
```diff
@@ -101,7 +101,7 @@ function propertyDidChange(obj, keyName, _meta) {
       dependentKeysDidChange(obj, keyName, meta);
     }
 
-    chainsDidChange(obj, keyName, meta, false);
```
bekzod: redundant argument

krisselden (approved):

PR #15487, opened 2017-07-11
ArrayProxy.length YUNO DK?

stefanpenner:
context: https://github.com/emberjs/data/pull/5058#discussion_r126596023

- [ ] @krisselden  r?
- [ ] test


stefanpenner:
nm, ember-data is merely violating the API provided by MutableArray, and not invoking a `arrayContentDidChange` I believe...

PR #15491, opened 2017-07-11
remove logic that already handled by `PrimitiveReference.create`

bekzod:
https://github.com/glimmerjs/glimmer-vm/blob/master/packages/@glimmer/runtime/lib/references.ts#L7


krisselden (approved):

krisselden:
Restarting the sauce test, still not sure why Edge bails out sometimes

PR #15493, opened 2017-07-11
avoid expending properties twice in `computed_macros`

bekzod:


packages/ember-runtime/lib/computed/computed_macros.js
```diff
@@ -49,9 +50,9 @@ function generateComputedWithPredicate(name, predicate) {
       }
 
       return get(this, expandedProperties[lastIdx]);
-    });
+    }, { dependentKeys: expandedProperties });
 
-    return computedFunc.property(...expandedProperties);
```
bekzod: `.property(..)` would call `expandProperties` on each already expanded property

krisselden:
This looks good, not sure what is going on with the Sauce tests, I'm going to try to restart.  I don't have time to dig into why Sauce labs is failing right now, but will try to look into it.

PR #15496, opened 2017-07-13
reused and refactored `shallowEqual` 

bekzod:


packages/ember-routing/lib/system/router_state.js
```diff
@@ -11,11 +11,8 @@ export default EmberObject.extend({
     let state = this.routerJsState;
     if (!this.routerJs.isActiveIntent(routeName, models, null, state)) { return false; }
 
-    let emptyQueryParams = isEmpty(Object.keys(queryParams));
-
-    if (queryParamsMustMatch && !emptyQueryParams) {
-      let visibleQueryParams = {};
-      assign(visibleQueryParams, queryParams);
+    if (queryParamsMustMatch && Object.keys(queryParams).length > 0) {
```
krisselden: This still loops keys three times, once to check length, once to clone and once to shallow equal on clone, seems like this could use more cleanup

bekzod: cannot think of clean way doing it, at least doesn't check length until queryParamsMustMatch is true, PR intention was to just reuse `shallowEqual` this is just what I picked up on the way

krisselden: fair enough, you did make it better than it was

PR #15497, opened 2017-07-13
removed `applyStr`

bekzod:
since
* its not being consistently used everywhere (just used in 2 places)
* probably modern browsers already optimize enough to not use this


packages/ember-utils/lib/invoke.js
```diff
@@ -44,6 +42,7 @@ export function canInvoke(obj, methodName) {
 */
 export function tryInvoke(obj, methodName, args) {
   if (canInvoke(obj, methodName)) {
-    return args ? applyStr(obj, methodName, args) : applyStr(obj, methodName);
+    let method = obj[methodName];
+    return args ? method.apply(obj, args) : method.call(obj);
```
bekzod: maybe just use `method.apply(obj, args)` (without args check)

packages/ember-metal/lib/events.js
```diff
@@ -185,17 +184,13 @@ export function sendEvent(obj, eventName, params, actions, _meta) {
...
+
+    if (params) {
+      method.apply(target, params);
     } else {
-      if (params) {
-        method.apply(target, params);
-      } else {
-        method.call(target);
-      }
+      method.call(target);
```
bekzod: maybe just use `method.apply(target, args)` (without args check)

stefanpenner: 👍

packages/ember-utils/lib/invoke.js
stefanpenner: 👍

stefanpenner (commented):
looks good, are you able to do some small bench?

krisselden:
Last time I checked, about 1 year ago it still made a difference. We should recheck

bekzod:
is this good enough ? 
https://gist.github.com/bekzod/6f81325167e67e9545e3b06477053d5c
```
calling `obj.sumAll` only
applyStr: 3075
.call: 3017
.apply: 2987
direct call: 2873

calling `obj.sum3` only
applyStr: 42
.call: 18
.apply: 22
direct call: 11

calling altogether 
applyStr: 2844
.call: 2679
.apply: 2780
direct call: 2700
```
`run in v6.8.1`

bekzod:
@rwjblue isn't this also good for 3.0 cleanup :P

mmun:
Looks good :+1:

krisselden:
@bekzod thank you

PR #15498, opened 2017-07-13
Merge pull request #15117 from Serabe/fix/14738

svenpl:

We had this problem during the upgrade to 12.2lts, so it would be nice to have it in the next release.


Serabe:
Can you cherry-pick this commit instead? https://github.com/emberjs/ember.js/commit/2acf4186fad0285894fd0e5dfb8b68945ea32403

That way the change will include the whole description of it and I think that makes it easier to do a release then.

svenpl:
I will do that

PR #15499, opened 2017-07-13

### LINKED ISSUE #14738: Cannot convert object to primitive value

svenpl:
`ember-utils/to-string` is documented as a `toString` function that supports
objects without a `toString` method.

Arrays do have a string method that is based on their `join` method. This
method spec ([Array.prototype.join in exma262 2017
draft](https://tc39.github.io/ecma262/#sec-array.prototype.join)) is based on
calling `toString` on the objects that form the collection. If an object were
not to have such method, `Array.prototype.toString` would throw and error.

This PR recreates `Array.prototype.join` as closer to the spec as I could,
changing the `ToString` part with the safe version of `toString`.

Fixes #14738 and #15113.

### DIFF
--- a/packages/ember-utils/lib/to-string.js
+++ b/packages/ember-utils/lib/to-string.js
@@ -1,11 +1,36 @@
 const objectToString = Object.prototype.toString;
 
+function isNone(obj) {
+  return obj === null || obj === undefined;
+}
+
 /*
  A `toString` util function that supports objects without a `toString`
  method, e.g. an object created with `Object.create(null)`.
 */
 export default function toString(obj) {
-  if (obj && typeof obj.toString === 'function') {
+  let type = typeof obj;
+  if (type === "string") { return obj; }
+
+  if (Array.isArray(obj)) {
+    // Reimplement Array.prototype.join according to spec (22.1.3.13)
+    // Changing ToString(element) with this safe version of ToString.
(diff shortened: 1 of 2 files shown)


locks:
Did you send this by mistake @svenpl? It looks like spam, as it's a copy of https://github.com/emberjs/ember.js/pull/15117.

locks:
Sorry, noticed the LTS target branch!

svenpl:
No problem. Maybe i should have added any kind of explaination. We ran into this bug as we where upgrading to Ember 2.12 lts, so it would be nice to have this in the next release.

notmessenger:
This was merged into the `#lts-2-12` **HOWEVER** this change was actually not released until `2.13.0` (https://github.com/emberjs/ember.js/blob/v2.13.0/packages/ember-utils/lib/to-string.js)

PR #15502, opened 2017-07-14
[BUGFIX release] [Fixes #15492] WorkAround WebKit/JSC JIT WorkARound

### LINKED ISSUE #15492: getOwner(this) is sometimes returning undefined after ember-2.14.0-beta2
This is related to:

https://github.com/intercom/ember-href-to/issues/81
and
https://github.com/Vestorly/torii/issues/349#issuecomment-302141236
and perhaps 
https://github.com/emberjs/ember.js/issues/15468

It appears that sometimes `getOwner(this)` returns `undefined`, and we are having to work around it with `const owner = getOwner(this) || this.container`

stefanpenner:
Affected version: Version 10.0.3 (12602.4.8)
Unaffected versions: Safari 11.0, WebKit 12604.1.29

Which suggests the issue has been addressed upstream.

---

The following commit https://github.com/emberjs/ember.js/commit/986710f9c57a9ff4520034b600e7fea93b56b955#diff-7e13eecefe753df1d82ce67b32bc4366R566 remove an implicit toBoolean conversion in WeakMap_peekParentMeta which apparently trips up Webkit…

Simply reversing the expression addressed the issue.

failed: `if (meta === null || meta !== undefined) { … }`
passed: `if (meta !== undefined || meta === null) { … }`

We also, confirmed that reverting the one line in question to `meta === null || !meta) {` fixed Safari.

This PR removes the need for the an unused feature which avoids the expression entirely.

@krisselden / @stefanpenner

### DIFF
--- a/packages/ember-metal/lib/meta.js
+++ b/packages/ember-metal/lib/meta.js
@@ -515,10 +515,7 @@ if (HAS_NATIVE_WEAKMAP) {
       if (DEBUG) {
         counters.peekCalls++;
       }
-      // stop if we find a `null` value, since
-      // that means the meta was deleted
-      // any other truthy value is a "real" meta
-      if (meta === null || meta !== undefined) {
+      if (meta !== undefined) {
         return meta;
       }
 
@@ -530,14 +527,10 @@ if (HAS_NATIVE_WEAKMAP) {
   };
 } else {
   setMeta = function Fallback_setMeta(obj, meta) {
-    // if `null` already, just set it to the new value
-    // otherwise define property first
-    if (obj[META_FIELD] !== null) {
-      if (obj.__defineNonEnumerable) {
-        obj.__defineNonEnumerable(EMBER_META_PROPERTY);
-      } else {
-        Object.defineProperty(obj, META_FIELD, META_DESC);
-      }
+    if (obj.__defineNonEnumerable) {
+      obj.__defineNonEnumerable(EMBER_META_PROPERTY);
+    } else {
+      Object.defineProperty(obj, META_FIELD, META_DESC);
     }
 
     obj[META_FIELD] = meta;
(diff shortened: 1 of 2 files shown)


stefanpenner:
found we still set null explicitly here: https://github.com/emberjs/ember.js/blob/master/packages/ember-runtime/lib/mixins/array.js#L641

packages/ember-metal/lib/meta.js
```diff
@@ -518,7 +518,7 @@ if (HAS_NATIVE_WEAKMAP) {
       // stop if we find a `null` value, since
```
krisselden: this comment seems incorrect

packages/ember-runtime/lib/mixins/array.js
```diff
@@ -638,7 +638,7 @@ const ArrayMixin = Mixin.create(Enumerable, {
 function EachProxy(content) {
   this._content = content;
   this._keys = undefined;
-  this.__ember_meta__ = null;
+  this.__ember_meta__ = undefined;
```
krisselden: Is this worth keeping?

packages/ember-metal/lib/meta.js
```diff
@@ -509,7 +509,7 @@ if (HAS_NATIVE_WEAKMAP) {
   peekMeta = function WeakMap_peekParentMeta(obj) {
     let pointer = obj;
     let meta;
-    while (pointer !== undefined && pointer !== null) {
+    while (pointer !== undefined && null !== pointer) {
```
krisselden: what is this for? this seems unrelated

packages/ember-runtime/lib/mixins/array.js
stefanpenner: I think we can strip it once we drop support for old browsers entirely. For now, keeping the shape stable seems important?

packages/ember-metal/lib/meta.js
stefanpenner: interesting...

packages/ember-runtime/lib/mixins/array.js
krisselden: why not then call meta(this) instead?

stefanpenner: I wanted to avoid making any significant changes for bugfix we intend to backport without proper measurement.

But I can gladly make the change

krisselden (approved):

PR #15503, opened 2017-07-14
cleanup resolver

bekzod:


packages/ember-application/lib/system/resolver.js
```diff
@@ -117,11 +117,9 @@ export default EmberObject.extend({
   init() {
     this._parseNameCache = dictionary(null);
   },
+
   normalize(fullName) {
-    let [
-      type,
-      name
-    ] = fullName.split(':', 2);
```
bekzod: last time I check second arguments slows down split, (I think thats why its not widely used in code)

packages/ember-application/lib/system/resolver.js
```diff
@@ -117,11 +117,9 @@ export default EmberObject.extend({
...
+    let [ type, name ] = fullName.split(':');
```
stefanpenner: also split `[ type, name ]` (unless loose mode is enabled for babel) produced some gnarly output.

If we are not using loose, lets move away from array destructuring, if we are we can leave it.

stefanpenner (changes requested):

packages/ember-application/lib/system/resolver.js
bekzod: I think it generates alright code https://cl.ly/2X0E1Z0Z163s/Screen%20Shot%202017-07-14%20at%209.46.10%20AM.jp2

stefanpenner: looks great, looks like we are correctly using loose mode. Thanks for checking!

stefanpenner (approved):

stefanpenner:
restarting sauce..

PR #15510, opened 2017-07-16
[BUGFIX beta] fix for lastObject/firstObject update issue

bekzod:
enhanced #15110 PR, @stefanpenner

### DIFF
--- a/packages/ember-runtime/lib/mixins/array.js
+++ b/packages/ember-runtime/lib/mixins/array.js
@@ -141,19 +141,29 @@ export function arrayContentDidChange(array, startIdx, removeAmt, addAmt) {
 
   let meta = peekMeta(array);
   let cache = meta && meta.readableCache();
+  if (cache !== undefined) {
+    let length = get(array, 'length');
+    let addedAmount = (addAmt === -1 ? 0 : addAmt);
+    let removedAmount = (removeAmt === -1 ? 0 : removeAmt);
+    let delta = addedAmount - removedAmount;
+    let previousLength = length - delta;
 
-  if (cache) {
-    if (cache.firstObject !== undefined &&
-        objectAt(array, 0) !== cacheFor.get(cache, 'firstObject')) {
-      propertyWillChange(array, 'firstObject', meta);
-      propertyDidChange(array, 'firstObject', meta);
+    let normalStartIdx = startIdx < 0 ? previousLength + startIdx : startIdx;
+    if (cache.firstObject !== undefined && normalStartIdx === 0) {
+      propertyWillChange(array, 'firstObject');
+      propertyDidChange(array, 'firstObject');
     }
-    if (cache.lastObject !== undefined &&
-        objectAt(array, get(array, 'length') - 1) !== cacheFor.get(cache, 'lastObject')) {
-      propertyWillChange(array, 'lastObject', meta);
-      propertyDidChange(array, 'lastObject', meta);
-    }
+
+    if (cache.lastObject !== undefined) {
+      let previousLastIndex = previousLength - 1;
+      let lastAffectedIndex = normalStartIdx + removedAmount;
+      if (previousLastIndex < lastAffectedIndex) {
+        propertyWillChange(array, 'lastObject');
+        propertyDidChange(array, 'lastObject');
+      }
+   }
   }
+
   return array;
 }
 
--- a/packages/ember-runtime/tests/suites/mutable_array/pushObject.js
+++ b/packages/ember-runtime/tests/suites/mutable_array/pushObject.js
@@ -54,4 +54,26 @@ suite.test('[A,B,C].pushObject(X) => [A,B,C,X] + notify', function() {
   equal(observer.validate('firstObject'), false, 'should NOT have notified firstObject');
 });
 
+suite.test('[A,B,C,C].pushObject(A) => [A,B,C,C] + notify', function() {
+  let before = this.newFixture(3);
+  let item = before[2]; // note same object as current tail. should end up twice
+  let after = [before[0], before[1], before[2], item];
+  let obj = this.newObject(before);
+  let observer = this.newObserver(obj, '[]', '@each', 'length', 'firstObject', 'lastObject');
+
+  obj.getProperties('firstObject', 'lastObject'); /* Prime the cache */
+
+  obj.pushObject(item);
+
+  deepEqual(this.toArray(obj), after, 'post item results');
+  equal(get(obj, 'length'), after.length, 'length');
+
+  equal(observer.timesCalled('[]'), 1, 'should have notified [] once');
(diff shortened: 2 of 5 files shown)


packages/ember-runtime/lib/mixins/array.js
```diff
@@ -141,19 +141,24 @@ export function arrayContentDidChange(array, startIdx, removeAmt, addAmt) {
...
-        objectAt(array, 0) !== cacheFor.get(cache, 'firstObject')) {
-      propertyWillChange(array, 'firstObject', meta);
-      propertyDidChange(array, 'firstObject', meta);
-    }
-    if (cache.lastObject !== undefined &&
-        objectAt(array, get(array, 'length') - 1) !== cacheFor.get(cache, 'lastObject')) {
-      propertyWillChange(array, 'lastObject', meta);
-      propertyDidChange(array, 'lastObject', meta);
+  if (cache !== undefined) {
+    if (cache.firstObject !== undefined && startIdx === 0 && objectAt(array, 0) !== cacheFor.get(cache, 'firstObject')) {
```
stefanpenner: i think one of the goals @hjdivad had, was to avoid the `objectAt(x)` lookups, as they would force the CP to be accessed.

I'll ping @hjdivad   to chime in.

stefanpenner (changes requested):

packages/ember-runtime/lib/mixins/array.js
bekzod: yeah but you need to invalidate cp when it changes no ?

stefanpenner: If the index for firstObject / lastObject changed, then we should broadcast. Even if the object is equal. That means we wont need both the objectAt stuff and the range based detection.

bekzod:
I updated my PR, so it does not call `objectAt` but now behavior is changed of `firstObject/lastObject` change event, so now it notifies change even if same object as `firstObject/lastObject` is shifted/pushed to `array` , I think it is still better than how was before.
adjusted/added test accordingly @stefanpenner

stefanpenner:
@bekzod ya, that was the approach @hjdivad was taking. It is likely a trade-off we are good with.

stefanpenner (approved):

packages/ember-runtime/lib/mixins/array.js
```diff
@@ -141,19 +141,27 @@ export function arrayContentDidChange(array, startIdx, removeAmt, addAmt) {
...
+
+    if (cache.lastObject !== undefined) {
+      let addedAmount = (addAmt === -1 ? 0 : addAmt);
+      let removedAmount = (removeAmt === -1 ? 0 : removeAmt);
+      let delta = addedAmount - removedAmount;
+      let length = get(array, 'length');
+
+      let lastIndex = length - delta - 1;
+      let lastAffectedIndex = startIdx + removedAmount;
+      if (lastIndex < lastAffectedIndex) {
```
hjdivad: I think this will miss cases where `lastObject` changes from negative indices, eg

```js
[0, 1, 2, 3].replace(-2, 2);
// => [0, 1]
```

packages/ember-runtime/tests/suites/mutable_array/pushObject.js
```diff
@@ -54,4 +54,26 @@ suite.test('[A,B,C].pushObject(X) => [A,B,C,X] + notify', function() {
   equal(observer.validate('firstObject'), false, 'should NOT have notified firstObject');
 });
 
+suite.test('[A,B,C,C].pushObject(A) => [A,B,C,C] + notify', function() {
```
hjdivad: 👍 

We'll need a test for removing the tail from a negative index

hjdivad (changes requested):

packages/ember-runtime/lib/mixins/array.js
```diff
@@ -141,19 +141,27 @@ export function arrayContentDidChange(array, startIdx, removeAmt, addAmt) {
...
-      propertyDidChange(array, 'firstObject', meta);
-    }
-    if (cache.lastObject !== undefined &&
-        objectAt(array, get(array, 'length') - 1) !== cacheFor.get(cache, 'lastObject')) {
-      propertyWillChange(array, 'lastObject', meta);
-      propertyDidChange(array, 'lastObject', meta);
-    }
+  if (cache !== undefined) {
+    if (cache.firstObject !== undefined && startIdx === 0) {
+      propertyWillChange(array, 'firstObject');
```
hjdivad: (which is fine, just want to confirm)

packages/ember-runtime/lib/mixins/array.js
bekzod: thanks, was not sure if negative startIndex could be passed will handle that

hjdivad: awesome thanks @bekzod 👍

packages/ember-runtime/tests/suites/mutable_array/pushObject.js
bekzod: added that, you could take another shot when you have time, thanks

packages/ember-runtime/tests/suites/mutable_array/replace.js
```diff
@@ -126,6 +126,27 @@ suite.test('[A,B,C,D].replace(2,2) => [A,B] + notify', function() {
   equal(observer.validate('firstObject'), false, 'should NOT have notified firstObject once');
 });
 
+suite.test('[A,B,C,D].replace(-1,1) => [A,B,C] + notify', function() {
```
hjdivad: 👍

packages/ember-runtime/lib/mixins/array.js
```diff
@@ -141,19 +141,29 @@ export function arrayContentDidChange(array, startIdx, removeAmt, addAmt) {
...
-      propertyDidChange(array, 'lastObject', meta);
-    }
+  if (cache !== undefined) {
+    let length = get(array, 'length');
+    let addedAmount = (addAmt === -1 ? 0 : addAmt);
+    let removedAmount = (removeAmt === -1 ? 0 : removeAmt);
+    let delta = addedAmount - removedAmount;
+    let previousLength = length - delta;
+
+    let normalStartIdx = startIdx < 0 ? previousLength + startIdx : startIdx;
```
hjdivad: 👍

hjdivad (approved):
LGTM

krisselden (approved):

stefanpenner:
@krisselden bugfix beta or let it ride the canary train?

stefanpenner:
chatted with @krisselden hes ok with `beta`

stefanpenner:
@bekzod can you squash, and prefix with `[BUGFIX beta]` ?

bekzod:
@stefanpenner done

Issue #15490, opened 2017-07-11
2.15 beta - onFulfillment is not a function


kellyselden:
From my stepping through the stack traces on both 2.14 and 2.15, it seems the RSVP catch logic may have changed a little between versions.

https://github.com/kellyselden/ember-wait-for-test-helper/pull/50/files#diff-d913131b1f61f7940eaff58ff6996ec6R35
https://travis-ci.org/kellyselden/ember-wait-for-test-helper/jobs/252511064#L1969

bekzod:
yes it has been upgraded to 3.5.0 in 2.14 and to 3.6.1 in 2.15.beta, is there any issue with it ?

rwjblue:
@stefanpenner - Can you please review https://github.com/emberjs/ember.js/blob/master/packages/ember-testing/lib/test/promise.js#L12-L14?  Specifically, it seems that we cannot assume `onFulfillment` is actually passed to our custom Promise subclasses `.then`. Is this a bug in RSVP somehow, or just in `isolate` assuming that the first arg is always present?

bekzod:
@stefanpenner I don't think RSVP behaving incorrectly here is gist for reproducing same issue 
both native inherited and rsvp behaving same way rejecting when inherited `then` has null `onFulfillment` 
https://gist.github.com/bekzod/68e8fc1c739745b6a5bdb12e99badfd4 (run in node 6.8.1)

stefanpenner:
@bekzod I agree. I don't believe RSVP is the offender here, although I am surprised upgrading causes a downstream issue at all (but I dont have time to do root cause analysis on this one), the issue is in embers code, we should just fix that guy up and cut a new beta.

PR #15511, opened 2017-07-16
[BUGFIX beta] possible fix for #15490

bekzod:
possible fix for #15490


locks:
Hi, can you prefix the commit and the PR with `[BUGFIX beta]`?

PR #15514, opened 2017-07-17
remove unnecessary check in `setPath`

bekzod:


packages/ember-metal/lib/property_set.js
```diff
@@ -95,9 +95,9 @@ function setPath(root, path, value, tolerant) {
   let parts = path.split('.');
   let keyName = parts.pop();
 
-  assert('Property set failed: You passed an empty path', keyName && keyName.length > 0)
+  assert('Property set failed: You passed an empty path', keyName.trim().length > 0)
 
-  let newPath = parts.length > 0 ? parts.join('.') : keyName;
```
bekzod: since we already checking if it is path here https://github.com/bekzod/ember.js/blob/14243438756b908b4557f9f3e296d3af89bcda81/packages/ember-metal/lib/property_set.js#L45 parts is guaranteed to be more than 0

packages/ember-metal/lib/weak_map.js
```diff
@@ -115,7 +115,7 @@ WeakMap.prototype.has = function(obj) {
  */
 WeakMap.prototype.delete = function(obj) {
   if (this.has(obj)) {
-    delete metaFor(obj).writableWeak()[this._id];
```
bekzod: `this.has` checks with `peekMeta` already so should be peekMeta here too
https://github.com/bekzod/ember.js/blob/14243438756b908b4557f9f3e296d3af89bcda81/packages/ember-metal/lib/weak_map.js#L100

Issue #15516, opened 2017-07-17
Ember/V8: Slow EmberError constructor


schuay:
I heard through the grapevine that error handling is very slow when using current V8 versions (e.g. 6.0). 

Looking at the `EmberError` constructor [0], the main offender is probably the call to `Error.captureStackTrace`. `captureStackTrace` used to format the stack trace lazily on access to the `stack` property (like builtin error objects), but since about a year it formats traces eagerly. That could be fairly expensive, and would explain a sudden slowdown in error handling code. We're looking into re-enabling lazy formatting in `Error.captureStackTrace` [1] soon.

I also noticed that each call to the `EmberError` constructor collects the stack trace three times:
1. In the call to `super()`
2. In the explicit `Error.call(this, message)`
3. In `Error.captureStackTrace`.

Maybe it'd be possible to avoid some of this useless work? In current V8 versions, it should be enough to just call `super(message)` - this will also filter to relevant frames (as `Error.captureStackTrace` does).

[0] https://github.com/emberjs/ember.js/blob/094740737771895e154e8ed2c35ed09fa63b0baa/packages/ember-debug/lib/error.js#L32
[1] https://crbug.com/v8/5962

cc @hashseed, @bmeurer, @stefanpenner

stefanpenner:
@schuay I believe our usage of `Error.captureStackTrace` are purely for cosmetic reasons. I'll just remove it.

stefanpenner:
- [ ] remove captureStackTrace: https://github.com/emberjs/ember.js/pull/15519
- [ ] remove duplicate "super" call `super()` and `Error.call(this, message)` but this will require a quick test to ensure we don't foobar anything.

stefanpenner:
> In the call to super()
> In the explicit Error.call(this, message)

is kinda funny bit-rot. Time to clean that up.

stefanpenner:
We also do stack trace stuff herE: https://github.com/emberjs/ember.js/blob/master/packages/ember-debug/lib/deprecate.js#L81-L118

When we provide deprecations. These get hit quite abit, I wonder if we should log a deprecation sans stack, but provide via console message how to enable stacks...

Thoughts @rwjblue ?

schuay:
[...]

PR #15519, opened 2017-07-17
[BUGFIX beta] remove Error.captureStackTrace

stefanpenner:
We only do this for stack cosmetic reasons, it really just removes 1 or so frames, but apparently more recently at a hefty cost:

context: https://github.com/emberjs/ember.js/issues/15516


stefanpenner:
cc @schuay

PR #15521, opened 2017-07-17
[BUGFIX release] remove Error.captureStackTrace

stefanpenner:
We only do this for stack cosmetic reasons, it really just removes 1 or so frames, but apparently more recently at a hefty cost:

context: https://github.com/emberjs/ember.js/issues/15516

cc @schuay

---

may as well just port this to release as well, no harm...


musically-ut:
I was probably summoned by mistake on this PR, but for what it's worth, LGTM. 👍  :)

locks (approved):

PR #15522, opened 2017-07-17
[BUGFIX release] Fix Ember.Error inheritance

stefanpenner:
As pointed out by @schuay https://github.com/emberjs/ember.js/issues/15516 our error constructor essentially calls super twice. The reasons seems to have been an error in ExtendableBuiltin constructor, which did not return its `apply` call. I can’t seem to find the issue ExtendBuiltin is working around. I suspect we can simply remove it

cc @schuay


chadhietala (approved):
I feel like this didn't work for some reason at some point, but if CI passes I'm 👍 .

stefanpenner:
@chadhietala I tested some cases manually, which seemed to work. Could you give it a try as well?

packages/ember-debug/lib/error.js
```diff
@@ -18,15 +7,10 @@ function ExtendBuiltin(klass) {
...
-
-    if (!(this instanceof EmberError)) {
-      return new EmberError(message);
-    }
+export default class EmberError extends Error {
+  constructor(message, code) {
+    let error = super(message, code);
 
-    let error = Error.call(this, message);
     this.stack = error.stack;
```
krisselden: if super() is called is this still needed?

stefanpenner: I think so in transpiled land

packages/ember-debug/lib/error.js
```diff
@@ -18,15 +7,12 @@ function ExtendBuiltin(klass) {
...
 */
-export default class EmberError extends ExtendBuiltin(Error) {
-  constructor(message) {
-    super();
 
-    if (!(this instanceof EmberError)) {
-      return new EmberError(message);
-    }
+export default EmberError;
+function EmberError(message, code) {
```
jherdman: Dumb question, but why make this a function instead of using the fancy ES6 classes?

stefanpenner: EmberError unfortunately supports being invoked without using new. This ends up being a less hacky solution. We should deprecate that, so we can eventually switch over entirely though.

locks:
@stefanpenner IE is not happy in saucelabs :\

packages/ember-debug/lib/error.js
```diff
@@ -18,15 +7,12 @@ function ExtendBuiltin(klass) {
...
+  if (this instanceof EmberError) {
+    let error = Error.call(this, message, code);
```
schuay: FYI: In V8, using `super()` would give you the benefit of omitting all frames up to and including the `EmberError` constructor (which is what I believe `Error.captureStackTrace` was used for). ES6 `Error` subclasses are basically the better `Error.captureStackTrace`. Copying error properties below would also be unnecessary.

I'm not sure how other browsers behave though.

packages/ember-debug/lib/error.js
jherdman: Cool! Thank you for taking a moment to respond to my question.

packages/ember-debug/lib/error.js
stefanpenner: We aren't actually using real super (getting transpiled) also we still support an unfortunate scenario calling the constructor without new. Which ES6 classes don't support. We will need to deprecate that first, then remove in the future.

stefanpenner: I'll open another PR that deprecated the call constructor behavior so by the best LTS we can drop

stefanpenner:
test failure seems like one of the spurious sauce labs issues...

Serabe:
@stefanpenner this is failing for real (restarted to get the screenshots).

0005screenshot

Turbo87:
closing this, as https://github.com/emberjs/ember.js/pull/17216 made it obsolete

PR #15524, opened 2017-07-18
Fix Chrome stack trace stuff

stefanpenner:
* error[‘arguments’] sniff no longer works
* use prepareStackTrace rather then replace. 

- [ ] get @schuays +1
- [ ] perf test
- [ ] ??
- [ ] profit

some experimenting based on @schuary's recommendations. @schuay what do you think?

### DIFF
--- a/packages/ember-debug/lib/deprecate.js
+++ b/packages/ember-debug/lib/deprecate.js
@@ -91,14 +91,18 @@ if (DEBUG) {
     if (ENV.LOG_STACKTRACE_ON_DEPRECATION) {
       let stackStr = '';
       let error = captureErrorForStack();
-      let stack;
 
-      if (error.stack) {
-        if (error['arguments']) {
-          // Chrome
-          stack = error.stack.replace(/^\s+at\s+/gm, '').
-            replace(/^([^\(]+?)([\n$])/gm, '{anonymous}($1)$2').
-            replace(/^Object.<anonymous>\s*\(([^\)]+)\)/gm, '{anonymous}($1)').split('\n');
+      // only do this when we are going to actually process the stack
+      Error.prepareStackTrace = function(error, stack) {
+        return stack;
+      }
+      // grab the new structured stack
+      let stack = error.stack;
+      // stop capturing structured stacks
+      Error.prepareStackTrace = undefined;
+
+      if (stack) {
+        if (typeof Error.captureStackTrace === 'function') {
           stack.shift();
         } else {
           // Firefox


packages/ember-debug/lib/deprecate.js
```diff
@@ -91,14 +91,18 @@ if (DEBUG) {
...
-
-      if (error.stack) {
-        if (error['arguments']) {
-          // Chrome
-          stack = error.stack.replace(/^\s+at\s+/gm, '').
-            replace(/^([^\(]+?)([\n$])/gm, '{anonymous}($1)$2').
-            replace(/^Object.<anonymous>\s*\(([^\)]+)\)/gm, '{anonymous}($1)').split('\n');
+
+      // only do this when we are going to actually process the stack
+      Error.prepareStackTrace = function(error, stack) {
```
schuay: It's probably safest to store & restore `prepareStackTrace` in a try-finally. See https://chromium-review.googlesource.com/c/574533/2/test/mjsunit/mjsunit.js#32 for a pretty complete example.

packages/ember-debug/lib/deprecate.js
```diff
@@ -91,14 +91,18 @@ if (DEBUG) {
...
+        return stack;
+      }
+      // grab the new structured stack
+      let stack = error.stack;
+      // stop capturing structured stacks
+      Error.prepareStackTrace = undefined;
+
+      if (stack) {
+        if (typeof Error.captureStackTrace === 'function') {
```
schuay: I'm not sure what the previous code was intended to accomplish. 

* If it's just to omit the current stack frame, that'd be easier with `Error.captureStackTrace`. 
* If it's for custom stack trace formatting, then I'd probably use the `CallSite` / `prepareStackTrace` API (https://github.com/v8/v8/wiki/Stack-Trace-API#customizing-stack-traces). See also the example from the previous comment.

Note that when using `prepareStackTrace` in this way, `stack` is not a string but an array of `CallSites`, so `slice` below probably won't work as intended. Also, `CallSite.toString` formats differently than normal stack traces:

```
d8> function f() { return new Error(); }
d8> f().stack
"Error
    at f ((d8):1:23)
    at (d8):1:1"
d8> Error.prepareStackTrace = function(error, stack) { return stack; }
d8> f().stack
[f ((d8):1:23), (d8):1:1]
```

packages/ember-debug/lib/deprecate.js
stefanpenner: You are correct. I have slight PTSD to try/finally re: perf, but you guys have recently made that no longer an issue. Will update when I get to the office.

packages/ember-debug/lib/deprecate.js
stefanpenner: After thinking about it I believe 1. Gets us most of the way there. Will experimen

rwjblue:
@stefanpenner - Ping?

Serabe:
@stefanpenner ping

PR #15528, opened 2017-07-19
Deprecate `Controller#content` alias

locks:
- [x] Removes `ControllerContentModelAliasDeprecation`-related code.
- [x] Deprecates the `Controller#content` alias.
  - [x] Use `deprecatingAlias`.
  - [x] Write [deprecation guide](https://github.com/emberjs/website/pull/2969).


packages/ember-runtime/lib/mixins/controller.js
```diff
@@ -44,6 +44,6 @@ export default Mixin.create(ActionHandler, ControllerContentModelAliasDeprecatio
   /**
     @private
   */
-  content: alias('model')
+  content: deprecatingAlias('model')
```
rwjblue: This will need an options hash with id, until, description, etc

rwjblue (changes requested):

rwjblue:
Can you also remove the entry in features.json?

locks:
@rwjblue please re-review and 👀 the commit message for correctness 🙇

packages/ember-runtime/lib/mixins/controller.js
```diff
@@ -44,6 +44,5 @@ export default Mixin.create(ActionHandler, ControllerContentModelAliasDeprecatio
   /**
     @private
   */
-  content: alias('model')
-
+  content: deprecatingAlias('model', { id: 'ember-runtime.controller-content', until: '2.16.0' })
```
rwjblue: We also need to add a deprecation guide entry to the website and add a `url` property with the link.

rwjblue (commented):
Looks good. Happy to land when green, but we still need to follow up with the deprecation guide stuff for the newly added deprecation.

packages/ember-runtime/lib/mixins/controller.js
locks: deprecation guide PRed: https://github.com/emberjs/website/pull/2969

rwjblue (approved):

PR #15529, opened 2017-07-19
fixing a test for reading a property of a function

KrishMunot:
This might fix a template to let them read a property from a method


rwjblue:
Just wanted to confirm that you intended this to be a failing test PR (since there are only tests being added here)?

krisselden:
this is a duplicate of #14729

krisselden:
also, set() called on a function property didn't update the template prior to the LTS

PR #15530, opened 2017-07-19
Chrome and Firefox stack trace fix

KrishMunot:


packages/ember-debug/lib/deprecate.js
```diff
@@ -105,6 +106,9 @@ if (DEBUG) {
           stack = error.stack.replace(/(?:\n@:0)?\s+$/m, '').
             replace(/^\(/gm, '{anonymous}(').split('\n');
         }
+      } finally {
+          Error.prepareStackTrace = prevPrepareStackTrace;
```
rwjblue: Where is `prevPrepareStackTrace` being captured?

packages/ember-debug/lib/deprecate.js
```diff
@@ -91,9 +91,10 @@ if (DEBUG) {
     if (ENV.LOG_STACKTRACE_ON_DEPRECATION) {
       let stackStr = '';
       let error = captureErrorForStack();
-      let stack;
-
-      if (error.stack) {
+     
+      // do this when we are actually processing the stack
+      try {
+      Error.prepareStackTrace = function(error, stack) {
```
rwjblue: Indentation seems off here...

rwjblue (commented):
Seems like this is roughly the same area as @stefanpenner's https://github.com/emberjs/ember.js/pull/15524, are y'all working on this together? Should we land them in any particular order?

locks:
@KrishMunot if you could join efforts with @stefanpenner in https://github.com/emberjs/ember.js/pull/15524 it would be very helpful 🙏

PR #15532, opened 2017-07-20
refactor set/get path

bekzod:
* checks `isPath` before accessing property in `get`
* uses spliced path for `setPath/getPath` to avoid extra join/splits
* avoids checking `isPath` when setting through `setPath/getPath`
* avoids using `isPath` (since indexOf and split seem to perform equally https://jsperf.com/string-match-vs-string-split so better just split)

### DIFF
--- a/packages/ember-metal/lib/property_get.js
+++ b/packages/ember-metal/lib/property_get.js
@@ -55,7 +55,7 @@ export function get(obj, keyName) {
   assert(`'this' in paths is not supported`, keyName.lastIndexOf('this.', 0) !== 0);
   assert('Cannot call `Ember.get` with an empty string', keyName !== '');
 
-  return isPath(keyName) ? _getPath(obj, keyName) : _get(obj, keyName);
+  return isPath(keyName) ? _getPath(obj, keyName.split('.')) : _get(obj, keyName);
 }
 
 function _get(obj, keyName) {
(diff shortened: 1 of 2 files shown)


packages/ember-metal/lib/property_get.js
```diff
@@ -54,13 +53,16 @@ export function get(obj, keyName) {
   assert(`'this' in paths is not supported`, keyName.lastIndexOf('this.', 0) !== 0);
   assert('Cannot call `Ember.get` with an empty string', keyName !== '');
 
+  let pathParts = keyName.split(/\./);
```
krisselden: why not `'.'`?

krisselden:
that benchmark isn't a good test, it is only comparing the worse case for indexOf on a string that is much longer than most paths.  It would be better to use more realistic data and a more realistic distribution of hits/misses.

I would check https://github.com/krisselden/ember-macro-benchmark with this change.

bekzod:
benchmark https://cl.ly/3U152N3b1n1v/results.pdf

krisselden:
looks like good evidence it has no impact, question, did you happen to check the tested URL and confirm the code was different?

bekzod:
sorry what you mean by `tested URL` ?

bekzod:
@krisselden you might be right it was serving same versions on both urls will check tmrw why, so that benchmark might be wrong will check

bekzod:
so it happens if I use `emberobserver.com.har` only

bekzod:
@krisselden had to use 2.14.1 ember (because of wireformat incompatibilities with `emberobserver.com.har` I guess) https://cl.ly/0f2A3U3n303B/results.pdf there is not significant difference, does `ember-macro-benchmark` take into account memory usage ?

krisselden:
@bekzod any chance you can commit your benchmark for peer review?

bekzod:
because of a lot of conflicts making different PR with less radical changes

PR #15535, opened 2017-07-20
[BUGFIX release] Fix rendering a property of a function

krisselden:


krisselden:
Rebased and focused version of #15482 on just the regression.   Before 2.10 you could render a property of a function, but you couldn't notify it updated.  This should make its way into the LTS since this regressed in 2.10 /cc @rwjblue @locks 

It is simple to support updates too, I just don't think it should be back ported since it wasn't possible before.

rwjblue (approved):

packages/ember-glimmer/tests/integration/content-test.js
```diff
@@ -484,6 +484,29 @@ class DynamicContentTest extends RenderingTest {
...
+    let func = () => {};
+    func.aProp = 'this is a property on a function';
+
+    this.renderPath('func.aProp', { func });
+
+    this.assertContent('this is a property on a function');
+
+    this.assertStableRerender();
+
+    // this.runTask(() => set(func, 'aProp', 'still a property on a function'));
```
stefanpenner: why comments?

krisselden: because I missed that? I didn't comment it out.

rwjblue: The test was trying to do INUR (which is likely what prompted the fixing of the updating side also), so my guess is that the update/reset parts here are just commented out.

Issue #13984, opened 2016-08-01
[DOCS] Ember.RSVP.cast is not documented


chadhietala:

rwjblue:
`RSVP.cast` is just `Ember.RSVP.resolve` AFAIK ([see here](https://github.com/tildeio/rsvp.js/blob/master/lib/rsvp.js#L26)). I think `Ember.RSVP.Promise.resolve` would be preferred over either, but @stefanpenner should chime in to confirm/deny...

stefanpenner:
`cast` is dead, don't use it, as 100% of `cast` usages can be expressed as `Promise.resolve`. It exists purely as a deprecated compat thing.

PR #15536, opened 2017-07-21
[DOC release] Replace RSVP.cast with RSVP.resolve

NetForce1:
cast is dead (see: https://github.com/emberjs/ember.js/issues/13984), and should be replaced with resolve


bekzod:
@NetForce1 should be replaced with resolve here too https://github.com/NetForce1/ember.js/blob/d006a79056b129ffca8c80e7dfee1b7ad46dd8a7/packages/ember-runtime/tests/mixins/promise_proxy_test.js#L187

locks:
@bekzod can you make that change separately?

locks:
Thanks @NetForce1 :)

PR #15538, opened 2017-07-22
[BUGFIX canary] Deprecate TargetActionSupport

locks:
WIP


rwjblue (commented):
Looks pretty good. Needs:

* an entry in website deprecation guide
* URL in the deprecation options
* entry in features.json

bekzod:
does this mean that it will be called each time when you use `TextSupport` mixin or create input ?

mmun:
We've deprecated usage of `targetObject`. Are we still wanting to deprecate the whole mixin?

PR #15540, opened 2017-07-23
convert weakmap polyfill to es6 and use native weakmap when available

bekzod:
not sure it is desirable feature, throwing this for discussion


rwjblue:
Seems good to me, but would like @stefanpenner's thoughts...

rwjblue:
Decided to land it anyways, we can always revert if needed...

PR #15544, opened 2017-07-24
remove `isFrozen` check in `native_array`

bekzod:
since `Ember.Freezable` is deprecated and nothing inherits from `Ember.Freezable`

### DIFF
--- a/packages/ember-runtime/lib/system/native_array.js
+++ b/packages/ember-runtime/lib/system/native_array.js
@@ -15,6 +15,7 @@ import EmberArray, {
 import MutableArray from '../mixins/mutable_array';
 import Observable from '../mixins/observable';
 import Copyable from '../mixins/copyable';
+import { assert } from 'ember-debug';
 import { FROZEN_ERROR } from '../mixins/freezable';
 import copy from '../copy';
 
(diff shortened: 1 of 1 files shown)


packages/ember-runtime/lib/system/native_array.js
```diff
@@ -54,10 +53,6 @@ let NativeArray = Mixin.create(MutableArray, Observable, Copyable, {
 
   // primitive for array support.
   replace(idx, amt, objects) {
-    if (this.isFrozen) {
```
rwjblue: I don't think we can remove this (even though `Ember.Freezable` is deprecated, it is still fully supported until 3.0.0).

Can you swap this to a debug mode assertion though?

rwjblue (changes requested):

packages/ember-runtime/lib/system/native_array.js
bekzod: will do

rwjblue: @bekzod - Friendly ping...

bekzod: done

rwjblue:
Thank you!

Issue #15545, opened 2017-07-24
cacheFor returns the wrong value for aliased properties


dwickern:
Aliased properties (`computed.alias`, `computed.readOnly`, etc) return the [CONSUMED](https://github.com/emberjs/ember.js/blob/2232478367c36f911c7dc3425e3fba478e266358/packages/ember-metal/lib/alias.js#L16) object when you try to get their cached value.

[ember-twiddle](https://ember-twiddle.com/e1f0129649ef8be7d84379cbea776884?openFiles=controllers.application.js%2C)
```js
export default Ember.Controller.extend({
  appName: 'Ember Twiddle',
  aliased: Ember.computed.alias('appName'),
  
  init() {
    this._super(...arguments);
    console.log('before:', this.cacheFor('aliased'));
    console.log('get:', this.get('aliased'));
    console.log('after:', this.cacheFor('aliased'));
  }
});
```

Expected output:
```
before: undefined
get: Ember Twiddle
after: Ember Twiddle
```

Actual output:
```
before: undefined
get: Ember Twiddle
after: Object {}
```

krisselden:
I was surprised that `cacheFor` is marked a `@public` to be honest, but cacheFor is only for computed props not aliases, aliases are just using the cache meta for a flag, they don't actually cache values.

bekzod:
I think expected behavior should be `alias` should default to `aliased value cache` done so in pr #15548

dwickern:
Does public API ever distinguish between computed properties and aliases? They're both created by `Ember.computed`. I assumed aliases are a special-case implementation for performance, but they should outwardly behave the same as computed properties

krisselden:
`alias` just uses `Ember.computed` as a namespace, internally they are different.

pixelhandler:
@krisselden is this mainly a documentation issue? cc\ @locks

pixelhandler:
@locks should this be closed in favor of a documention issue elsewhere?

pixelhandler:
@bekzod @dwickern @krisselden @rwjblue is this still an issue, perhaps we should close or create a new reproduction of this, what do you think?

pixelhandler:
I updated the twiddle to a current version of Ember, https://ember-twiddle.com/791c87e01d7492046a5978e1aafae197?openFiles=controllers.application.js%2C

locks:
Closing as this issue is a weird intersection of APIs and it is not likely to get fixed as the impact could be wide ranging.

[...]

PR #15548, opened 2017-07-25
correct behaviour of cachefor for aliased values

bekzod:
possible fix for #15545, although doesn't fix it entirely, at least prevents exposing internal CONSUMED object

### DIFF
--- a/packages/ember-metal/tests/alias_test.js
+++ b/packages/ember-metal/tests/alias_test.js
@@ -6,7 +6,8 @@ import {
   isWatching,
   addObserver,
   removeObserver,
-  tagFor,
+  getCachedValueFor as cacheFor,
+  computed,
 } from '..';
 import { meta } from 'ember-meta';
 import { moduleFor, AbstractTestCase } from 'internal-test-helpers';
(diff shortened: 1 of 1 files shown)


packages/ember-metal/lib/computed.js
```diff
@@ -548,12 +551,19 @@ export default function computed(...args) {
 function cacheFor(obj, key) {
   let meta = peekMeta(obj);
   let cache = meta && meta.source === obj && meta.readableCache();
-  let ret = cache && cache[key];
 
-  if (ret === UNDEFINED) {
-    return undefined;
+  if (cache !== undefined) {
+    let prop = obj[key];
+    if (prop instanceof AliasedProperty) {
```
bekzod: could check for existence of `prop.altKey` instead of `instanceof` for performance reasons

packages/ember-metal/lib/computed.js
```diff
@@ -548,12 +551,19 @@ export default function computed(...args) {
...
+      return cacheFor(obj, prop.altKey);
```
dwickern: I suspect this won't work for `alias('some.path')` because `cacheFor` doesn't handle paths

bekzod:
I think this is not relevant anymore

PR #15549, opened 2017-07-26
use safe toString for array assetion

### LINKED ISSUE #15421: addObserverForContentKey performance in 2.13.3
Hi,

I ran into an issue with debug builds taking ages to run when using test data with somewhat large arrays. I tracked it down to `addObserverForContentKey`, which I believe was already an issue earlier #3156

https://github.com/emberjs/ember.js/blob/master/packages/ember-runtime/lib/mixins/array.js#L735
```
 assert(`When using @each to observe the array ${content}, the array must return an object`, typeof item === 'object');
```
This is not a problem if content is an ArrayProxy, but a native array will do Array.toString() on this and will blow things up.

I know this is fixed in release builds, but not being able to test my app with large datasets is not good at all.

bekzod:
fixes #15421

### DIFF
--- a/packages/ember-runtime/lib/mixins/array.js
+++ b/packages/ember-runtime/lib/mixins/array.js
@@ -6,8 +6,7 @@
 // ..........................................................
 // HELPERS
 //
-import { symbol } from 'ember-utils';
-
+import { symbol, toString } from 'ember-utils';
 import Ember, { // ES6TODO: Ember.A
   get,
   computed,
@@ -743,7 +742,7 @@ function addObserverForContentKey(content, keyName, proxy, idx, loc) {
   while (--loc >= idx) {
     let item = objectAt(content, loc);
     if (item) {
-      assert(`When using @each to observe the array ${content}, the array must return an object`, typeof item === 'object');
+      assert(`When using @each to observe the array \`${toString(content)}\`, the array must return an object`, typeof item === 'object');
       _addBeforeObserver(item, keyName, proxy, 'contentKeyWillChange');
       addObserver(item, keyName, proxy, 'contentKeyDidChange');
     }


rwjblue:
As I mentioned in #15421 the changes to the assertion in `addObserverForContentKey` are not required (as of 2.14 and higher).

The other changes (adding `meta` to `propertyDidChange`) seem quite unrelated (though perhaps good on their own?).  Mind splitting that out and submitting that separately?

bekzod:
@rwjblue won't `content` blow up if it is `[Object.create(null), {}]` because that is reason I used `ember-utils`'s `toString`

rwjblue:
@bekzod - Yes, but only if the assertion would already fail (e.g. not the original scenario from #15421).  I think moving to using the custom `toString` from `ember-utils` is totally fine for that part, but we definitely do not want to eagerly calculate it (which the PR does at the moment).

rwjblue:
I'm going to reopen so we can update that.

packages/ember-runtime/lib/mixins/array.js
```diff
@@ -150,16 +151,16 @@ export function arrayContentDidChange(array, startIdx, removeAmt, addAmt) {
 
     let normalStartIdx = startIdx < 0 ? previousLength + startIdx : startIdx;
     if (cache.firstObject !== undefined && normalStartIdx === 0) {
-      propertyWillChange(array, 'firstObject');
-      propertyDidChange(array, 'firstObject');
+      propertyWillChange(array, 'firstObject', meta);
```
rwjblue: Please migrate this change to a separate / targeted PR.

packages/ember-runtime/lib/mixins/array.js
```diff
@@ -150,16 +151,16 @@ export function arrayContentDidChange(array, startIdx, removeAmt, addAmt) {
...
+      propertyDidChange(array, 'firstObject', meta);
     }
 
     if (cache.lastObject !== undefined) {
       let previousLastIndex = previousLength - 1;
       let lastAffectedIndex = normalStartIdx + removedAmount;
       if (previousLastIndex < lastAffectedIndex) {
-        propertyWillChange(array, 'lastObject');
-        propertyDidChange(array, 'lastObject');
+        propertyWillChange(array, 'lastObject', meta);
```
rwjblue: Please migrate this change to a separate / targeted PR.

packages/ember-runtime/lib/mixins/array.js
```diff
@@ -740,10 +741,13 @@ EachProxy.prototype = {
 };
 
 function addObserverForContentKey(content, keyName, proxy, idx, loc) {
+  let arrayString;
+  if (DEBUG) { arrayString = toString(content); }
```
rwjblue: Please remove the `arrayString` variable, and update the assertion message to call `${toString(content)}` instead.

rwjblue (changes requested):

bekzod:
@rwjblue done

rwjblue (approved):

rwjblue:
Awesome, thank you!!

PR #15557, opened 2017-07-28
allow `readOnly` to passed as option in `ComputedProperty`

bekzod:
looked odd to use `.readOnly()` with  `new ComputedProperty`

### DIFF
--- a/packages/ember-metal/lib/computed.js
+++ b/packages/ember-metal/lib/computed.js
@@ -129,7 +129,8 @@ const DEEP_EACH_REGEX = /\.@each\.[^.]+\./;
 */
 function ComputedProperty(config, opts) {
   this.isDescriptor = true;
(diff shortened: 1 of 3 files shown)


packages/ember-metal/lib/computed.js
```diff
@@ -127,9 +127,10 @@ const DEEP_EACH_REGEX = /\.@each\.[^.]+\./;
   @namespace Ember
   @public
 */
-function ComputedProperty(config, opts) {
+function ComputedProperty(config, opts = {}) {
```
krisselden: this will compile to something that touches arguments.length. just initialize this._dependentKeys and this._readOnly with a `opts === undefined`

The miss for {} is a waste too.

krisselden (changes requested):

krisselden (approved):

PR #15563, opened 2017-08-01
micro optimization in `enumerable`

bekzod:


packages/ember-runtime/lib/mixins/enumerable.js
```diff
@@ -48,12 +48,9 @@ function pushCtx(ctx) {
 function iter(key, value) {
   let valueProvided = arguments.length === 2;
 
-  function i(item) {
-    let cur = get(item, key);
-    return valueProvided ? value === cur : !!cur;
```
bekzod: avoid checking `valueProvided` each iteration

packages/ember-runtime/lib/mixins/enumerable.js
```diff
@@ -753,7 +750,7 @@ const Enumerable = Mixin.create({
       let method = x && x[methodName];
 
       if ('function' === typeof method) {
-        ret[idx] = args ? method.apply(x, args) : x[methodName]();
```
bekzod: args is array here

PR #15564, opened 2017-08-01
[DOCS] Use shorthand for inject.service example

knownasilya:


locks:
Can you check all docs?

knownasilya:
@locks what do you mean? Like, go through all examples? Wish I could, but kind of crunched at the moment with a big project. Just happened to be checking the docs in this place and submitted before I forgot.

locks:
Alrighty ;)

PR #15552, opened 2017-07-27
Update blueprints and tests to RFC #176

locks:
[RFC #176](https://github.com/emberjs/rfcs/blob/master/text/0176-javascript-module-api.md).
[Conversion tables for imports](https://github.com/ember-cli/ember-rfc176-data).

- [x] Normalize `""`/`''` in imports
- [x] Address minor documentation issues, like typos and formatting
- [x] Update documentation code blocks
- [ ] Update documentation cross-linking (e.g. `Ember.ApplicationInstance` to `ApplicationInstance`)
- [x] Update blueprints and tests
- [ ] Update meta-documentation (`@module`, `@submodule`, etc)

PR #15567, opened 2017-08-03
Convert blueprints to using new module syntax

RobbieTheWagner:
This PR converts the blueprints over to using the new module syntax. I’m not sure if we want to throw in checks for ember-cli-babel version, and switch what they do based on that or not.

Additionally, I’m unsure if we want to merge these in before updating the rest of the codebase to use modules as well.

I just know I could really use some blueprints with the new syntax, so wanted to get the ball rolling :smiley:


locks:
Hey @rwwagner90, thanks for the effort but this is a duplicate of https://github.com/emberjs/ember.js/pull/15552, which also converted the tests and documentation over.

rwwagner90:
@locks I did a search for "module" in PRs and found nothing. Guess I should have searched RFC 🙁

PR #15577, opened 2017-08-10
[BUGFIX beta] Include missing sourcemaps in vendorTree

### LINKED ISSUE #15464: Missing sourcemaps
I get this error when upgrading to `ember-source@2.14.0` and building the app:

```
Warning: ignoring input sourcemap for vendor/ember/ember.debug.js because ENOENT: no such file or directory ...
Warning: ignoring input sourcemap for vendor/ember/ember-testing.js because ENOENT: no such file or directory, open ...
```

There are two problems with that:
* the supplied source maps are missing from the build (obviously)
* the warnings break our CI (but this is more related to https://github.com/ef4/fast-sourcemap-concat/issues/35)

I think the reason for this is that the `dist` files contain a reference to their sourcemap at the end of the file: `//# sourceMappingURL=ember.debug.map`. This seems to be new for 2.14, at least 2.13 did not have this.  While there is a `dist/ember.debug.map` in the published npm package, these files are not added to the `vendorTree` in [index.js](https://github.com/emberjs/ember.js/blob/master/index.js), so they are really missing in the tmp tree when `fast-sourcemap-concat` tries to read them.

I can put a PR together exporting those files in the `vendorTree` if that is the way to go? But given there is https://github.com/emberjs/ember.js/pull/15102, I am not so sure about it...

simonihmig:
Fixes #15464 

This adds the missing sourcemap files to `vendorTree`, so the references in `.js` files (`//# sourceMappingURL=ember.debug.map`) are not pointing to non-existing files anymore.

There were no existing tests for the addon-output, so not sure if I should add some? (maybe separate PR?). Tested this in a real app, seems to work fine so far:

Also not sure if this should rather be tagged for `release` rather than `beta`?

cc @rwjblue

### DIFF
--- a/index.js
+++ b/index.js
@@ -58,11 +58,18 @@ module.exports = {
       'ember.debug.js',
       'ember.min.js',
       'ember.prod.js'
-    ].filter(function(file) {
-      var fullPath = path.join(__dirname, 'dist', file);
+    ]
+      .map(function(file) {
+        return [file, file.replace('.js', '.map')];
+      })
+      .reduce(function(flat, jsAndMap) {
+        return flat.concat(jsAndMap);
+      }, [])
+      .filter(function(file) {
+        var fullPath = path.join(__dirname, 'dist', file);
 
-      return fs.existsSync(fullPath);
-    });
+        return fs.existsSync(fullPath);
+      });
 
     var ember = new Funnel(__dirname + '/dist', {
       destDir: 'ember',


rwjblue (approved):
LGTM, thanks for working on this!

Yes, I think `[BUGFIX release]` is technically more correct, but I do not know if we will do another 2.14.x patch release...

Turbo87:
@rwjblue it sounds like this was included in v2.15, but I'm pretty sure v2.15 caused https://github.com/ember-cli/ember-cli-uglify/issues/29#issuecomment-334586796 for me 🤔

PR #15582, opened 2017-08-13
[BUGFIX beta] Throw if Ember is in both `package.json` and `bower.json`.

rwjblue:
This warning was non-obvious (which one "wins"?), and allows for misconfiguration in `config/ember-try.js` to _seem_ like you are testing many Ember versions but in reality you are just testing whatever `ember-source` version you have :sob:.

See https://github.com/emberjs/data/pull/5123 for the full sob story...

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,6 +4,7 @@
 var fs = require('fs');
 var path = require('path');
 var resolve = require('resolve');
+var SilentError = require('silent-error');
 
 var paths = {};
 var absolutePaths = {};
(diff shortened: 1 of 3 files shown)


Turbo87:
this sounds like it has implications for `ember-try` with `versionCompatibility` because that will always use bower deps but does not set `ember-source: null` by default yet

rwjblue:
Ya, good point @Turbo87 we should fix that...

mmun:
@rwjblue Should this be closed or is there something we can salvage?

rwjblue:
No, now that bower is not a viable option we don’t need this any longer.

PR #15583, opened 2017-08-15
Update ember-cli-version-checker

ef4:
The presence of ember-cli-version-checker prior to 2.0.0 anywhere in an app's dependencies makes it impossible to use dependencies that are hoisted above the ember app's root directory. Now that yarn has workspaces support, that's not an unreasonable thing to want to do.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "ember-cli-string-utils": "^1.1.0",
     "ember-cli-test-info": "^1.0.0",
     "ember-cli-valid-component-name": "^1.0.0",
-    "ember-cli-version-checker": "^1.3.1",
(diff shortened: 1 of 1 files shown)


rwjblue:
The blueprint tests are failing, likely due to the fact that `ember-cli-version-checker` is being used to switch the mocha blueprints (and ember-mocha is not found in the actual `node_modules` graph).

Hopefully @Turbo87 has a suggestion?

Turbo87:
maybe generating a dummy `node_modules/ember-mocha/package.json` file works? I guess this issue is similar to https://github.com/simplabs/eslint-config-simplabs/pull/29

mmun:
This was completed in https://github.com/emberjs/ember.js/pull/15854.

PR #15585, opened 2017-08-16
Remove outdated {{with}} documentation

rlivsey:
`{{with}}` used to change the scope of `this`, so I expect this is what's meant by "Without the as operator, it would be impossible to reference user.name in the example above." but that's not the case now as far as I'm aware.

[Twiddle to demonstrate](https://ember-twiddle.com/77c748e8449e44c974f68c918910368d?openFiles=templates.application.hbs%2C).


locks:
Thanks :)

PR #15592, opened 2017-08-18
Support ES6 iterables on {{#each}} helper

cibernox:
The requirement is that they support `Symbol.iterator`, only for browsers that have iterators. 
The only supported browser that doesn't support is Internet Explorer 11.

One caveat of this feature is that iterables are not observables, so each
cannot track updates to them, but the code is structured on a way that
this branch is a fallback, so all special cases where KVO is possible are
handled as they where before.

### DIFF
--- a/packages/ember-glimmer/lib/utils/iterable.ts
+++ b/packages/ember-glimmer/lib/utils/iterable.ts
@@ -17,7 +17,7 @@ import {
   _contentFor,
   isEmberArray
 } from 'ember-runtime';
-import { guidFor } from 'ember-utils';
+import { guidFor, HAS_NATIVE_SYMBOL } from 'ember-utils';
 import { isEachIn } from '../helpers/each-in';
 import {
   UpdatablePrimitiveReference,
@@ -244,7 +244,6 @@ class EachInIterable {
     reference.update(item.value);
   }
 }
-
 class ArrayIterable {
   public ref: UpdatableReference;
   public keyFor: KeyFor;
@@ -279,6 +278,8 @@ class ArrayIterable {
       let array: any[] = [];
       iterable.forEach((item: any) => array.push(item));
       return ArrayIterator.from(array, keyFor);
+    } else if (HAS_NATIVE_SYMBOL && typeof iterable[Symbol.iterator] === 'function') {
+      return ArrayIterator.from(Array.from(iterable), keyFor);
     } else {
       return EMPTY_ITERATOR;
     }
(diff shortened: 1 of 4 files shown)


cibernox:
@rwjblue I've resucitated this PR from the dead. Some refactor happened in the code and not it was simpler to implement.

I took the naive approach of just calling `Array.from(iterator)` and then using a standard ArrayIterator. I didn't see a reason to do something fancier.

packages/ember-glimmer/lib/utils/iterable.ts
```diff
@@ -279,6 +278,8 @@ class ArrayIterable {
       let array: any[] = [];
       iterable.forEach((item: any) => array.push(item));
       return ArrayIterator.from(array, keyFor);
+    } else if (typeof iterable[Symbol.iterator] === 'function') {
```
rwjblue: We can't rely on `Symbol`

packages/ember-glimmer/lib/utils/iterable.ts
```diff
@@ -279,6 +278,8 @@ class ArrayIterable {
...
+      return ArrayIterator.from(Array.from(iterable), keyFor);
```
rwjblue: Don't think `Array.from` is available on IE11

rwjblue (changes requested):

rwjblue:
FWIW, I think @chancancode is working on something in this space also, we should coordinate to avoid extra work...

packages/ember-glimmer/lib/utils/iterable.ts
chancancode: I fixed it locally

packages/ember-glimmer/lib/utils/iterable.ts
chancancode: I fixed it locally

cibernox: It is true, but every browser that supports symbols also supports `Array.from`, so a single check it enough.

cibernox:
I added a condition to check if the browser has native symbols. That is enough to make it safe for IE11.

@chancancode is your approach different / lower-level?

chancancode:
@cibernox I did a bunch of refactoring (including switching to "lazily" pull the iterator) and it happen to naturally took care of the issues

cibernox:
I'd love to see that

chancancode:
@cibernox I forgot to ping – I opened https://github.com/emberjs/ember.js/pull/16399 which included your commit, my refactor, as well as my initial work to support `Map`/`Symbol.iterator` in `{{#each-in}}`. I probably should have marked it as WIP since I still wanted to write tests for the `{{#each-in}}` cases as well as update the docs and CHANGELOG. Would you like to take over from here? If not I can do it too.

PR #15593, opened 2017-08-18
[DOC release] document how to use pauseTest() with ember-mocha

caseywatts:
Anyone using `pauseTest()` probably needs to disable the timeout for the test in order for it to be useful. This PR documents that in exactly the place where it would help people discover this most easily.

I've tried quite a few approaches to reducing the timeout before coming across this one:
- Through an environment variable I used to toggle the mocha setup as described [here](https://github.com/emberjs/ember-mocha/issues/28).
- hoping the cli flag `--no-timeouts` would pass through to mocha (but it didn't).
- Finally, just doing it in the specific test (I didn't even realize this was an option right away!)

(for another PR/discussion:) It would be even more awesome if pauseTest() could set timeout to 0 for us, since that's what developers are likely to want/hope/expect when using pauseTest(). For discussion of that, see: https://github.com/emberjs/ember-mocha/issues/75

### DIFF
--- a/packages/ember-testing/lib/helpers/pause_test.js
+++ b/packages/ember-testing/lib/helpers/pause_test.js
@@ -31,6 +31,28 @@ export function resumeTest() {
  return pauseTest();
  click('.btn');
  ```
+
+ You may want to turn off the timeout before pausing.
+
+ qunit (as of 2.4.0):
+
+ ```
+ visit('/');
+ assert.timeout(0);
+ return pauseTest();
+ click('.btn');
+ ```
+
+ mocha:
+
+ ```
+ visit('/');
+ this.timeout(0);
+ return pauseTest();
+ click('.btn');
+ ```
+
+
  @since 1.9.0
  @method pauseTest
  @return {Object} A promise that will never resolve


dwickern:
I use this disable the timeout when tests are running interactively:
```js
// tests/test-helper.js
let mocha = window.mocha;

mocha.setup({
  timeout: 10000,
  slow: 1000
});

if (window.location.pathname === '/tests') {
  // no timeout in development
  mocha.setup({ timeout: 0 });
}
```

ro0gr:
>  I wonder how ember-qunit does this?

It doesn't - [twiddle](https://ember-twiddle.com/b9d380a0110f11ae1bebf2f0bdee93ad?openFiles=tests.acceptance.my-acceptance-test.js%2C).

I usually work-around this by switching to [`ember-cli-qunit`'s dev mode](https://github.com/ember-cli/ember-cli-qunit/pull/120) when using `pauseTest()`.

caseywatts:
updated with how to do this in the same sort of way for qunit :)

I like the other ways of making this work better! I'd love it if there was eventually some standard way to make this work more easily, by default. I like the idea of not having timeouts in interactive tests (with `--serve`) - but I don't see a way to get that to make that a default easily.

In the meantime, this documentation PR could go a long way toward users actually being able to use the `pauseTest` feature. I just presented the `this.timeout(0)` method to my team at Heroku and they were :mindblown: :)

caseywatts:
quote from a coworker 😆 
image

dwickern:
Don't forget `resumeTest` since ember 2.10~

caseywatts:
yeah!! I love that when you run `pauseTest()` it mentions `resumeTest()` to you now :)

rwjblue:
Thank you!

PR #15599, opened 2017-08-25
[BUGFIX release] ensure “pause on exception” pauses in the right place

stefanpenner:
It is quite common (like in ember-data’s tests) to not have a global onError handler in tests.
But because of the previous state of code, we would still end up paused in the default dispatchError rather then where the actual error was thrown.

This addresses the issue, but ensuring onErrorTarget.onerror returns undefined if no onError has been set

before:

screen shot 2017-08-25 at 6 15 44 am

After: 

screen shot 2017-08-25 at 6 16 01 am

### DIFF
--- a/packages/ember-metal/lib/error_handler.js
+++ b/packages/ember-metal/lib/error_handler.js
@@ -13,6 +13,12 @@ let getStack = error => {
   return stack;
 };
 
+export const onErrorTarget = {
+  get onerror() {
+    return dispatchOverride || getOnerror();
+  }
+};
+
 let onerror;
 // Ember.onerror getter
 export function getOnerror() {
--- a/packages/ember-metal/lib/run_loop.js
+++ b/packages/ember-metal/lib/run_loop.js
@@ -1,8 +1,7 @@
 import { GUID_KEY } from 'ember-utils';
 import { assert, isTesting } from 'ember-debug';
 import {
-  dispatchError,
-  setOnerror
+  onErrorTarget
 } from './error_handler';
 import {
   beginPropertyChanges,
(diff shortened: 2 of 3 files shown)


packages/ember-metal/lib/error_handler.js
```diff
@@ -13,6 +13,12 @@ let getStack = error => {
   return stack;
 };
 
+export const onErrorTarget = {
+  get onerror() {
+    return dispatchError || getOnerror();
```
rwjblue: Maybe I'm missing something, but I don't see how this works since `dispatchError` is always defined just below here?

I would think you want this to actually be:

```js
export const onErrorTarget = {
  get onerror() {
    // will need to be moved to below where these are defined (e.g. down to L41 below)
    return dispatchOverride || onerror;
  }
}
```

rwjblue (commented):
I think we also need a test for this.  Something like:

```js
import { 
  onErrorTarget,
  getDispatchOverride,
  setDispatchOverride,
  getOnerror,
  setOnerror
 } from '../error_handler';

const DISPATCH_OVERRIDE = getDispatchOverride();
const ONERROR = getOnerror();

module('error_handler', {
  afterEach() {
    setOnerror(ONERROR);
    setDispatchOverride(DISPATCH_OVERRIDE);
  }
});

test('by default there is no onerror', function(assert) {
  assert.equal(onErrorTarget.onerror, undefined);
});

test('when dispatch override is registered', function(assert) {
  assert.expect(2);
  let override = () => assert.ok(true, 'dispatch called');
  setDispatchOverride(overide);
  assert.equal(onErrorTarget.onerror, override);
});

test('when onerror is registered', function(assert) {
  assert.expect(2);
  let override = () => assert.ok(true, 'onerror called');
  setOnerror(overide);
  assert.equal(onErrorTarget.onerror, override);
});
```

stefanpenner:
yup, test looks good. Adding it

rwjblue:
@stefanpenner - We should consider backporting this to 2.12 LTS.  Thoughts?

stefanpenner:
There appears to be a BB issue preventing this from working correctly.

stefanpenner:
I think -> https://github.com/BackburnerJS/backburner.js/pull/262 fixes the BB issue

stefanpenner:
ya that BB fix address this.

stefanpenner:
oh GH you so silly... https://github.com/emberjs/ember.js/pull/15600

PR #15600, opened 2017-08-25
[BUGFIX LTS] ensure “pause on exception” pauses in the right place

stefanpenner:
It is quite common (like in ember-data’s tests) to not have a global onError handler in tests.
But because of the previous state of code, we would still end up paused in the default dispatchError rather then where the actual error was thrown.

This addresses the issue, but ensuring onErrorTarget.onerror returns undefined if no onError has been set

before:

screen shot 2017-08-25 at 6 15 44 am

After: 

screen shot 2017-08-25 at 6 16 01 am

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "babel-plugin-transform-es2015-template-literals": "^6.22.0",
     "babel-plugin-transform-proto-to-assign": "^6.23.0",
     "babel-template": "^6.24.1",
-    "backburner.js": "^1.2.0",
+    "backburner.js": "^1.2.1",
     "broccoli-babel-transpiler": "next",
     "broccoli-concat": "^3.2.2",
     "broccoli-file-creator": "^1.1.1",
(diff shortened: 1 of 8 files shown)


packages/ember-metal/tests/error_handler_test.js
```diff
@@ -0,0 +1,66 @@
+import Ember from 'ember';
```
rwjblue: This test should likely be moved to the `packages/ember/tests` folder, since it requires the global now.

krisselden:
```Test failed: Template scoping examples: global failure
    Failed assertion: TypeError: undefined is not an object (evaluating '_ember.default.Test.adapter')```

rwjblue:
Updated:

* rebase
* update backburner.js to 1.2.1 (to include other fixes required)
* fix tests that were previously throwing errors that were swallowed in production
* fix test that was clobbering `backburner._platform`

I tested both prod and debug builds before pushing, so now we just have to convince sauce labs 😈 ...

stefanpenner:
thanks @rwjblue

Issue #15586, opened 2017-08-16
 Ember should have a default toString for a Mixins that doesn't return the Mixin constructor source code


cibernox:
To track comment https://github.com/emberjs/ember-inspector/pull/704#pullrequestreview-56326413 
This saves a workaround in the ember inspector.

meslater1030:
@cibernox I'd love to take this one. I'm a newbie contributor and it looks reasonably straightforward. Thanks for all your recent work on the inspector!

cibernox:
Please, be my guest 😄 . 

If you need help ping me anytime on Slack.

meslater1030:
Fixed by #15601

PR #12636, opened 2015-11-21
[BUGFIX beta] Ensure Mixin.prototype.toString is defined.

rwjblue:
`Mixin.prototype` is sealed after its initial definition, but `Mixin.prototype.toString` is assigned later in `ember-runtime/namespace`.  This causes issues when `ember-template-compiler.js` is loaded before the rest of Ember (because the `Mixin.prototype` is sealed).  Unfortunately, the debug functions (for `Object.seal`) does not work in the `ember-metal` package at this time.

This fix is a temporary fix ensures that the slot for `toString` exists on `Mixin.prototype` to allow the loading of `ember-template-compiler.js` to be before or after `ember.debug.js`. The more permanent fix is to ensure that `debugSeal` works in ember-metal properly.

Fixes #12623.

PR #15601, opened 2017-08-25
[BUGFIX beta] Ensure Mixin.prototype.toString does not return constructor code

meslater1030:
Addresses issue #15586. As mentioned in PR #12636 this particular definition of Mixin.prototype.toString is only used when ember-template-compiler.js is loaded before the rest of Ember (or at least that was the case in late 2015). I'm not sure how to write a test for that.


rwjblue (approved):

meslater1030:
@rwjblue This has been rebased. Thanks for taking a look!

PR #15604, opened 2017-08-27
Data Adapter: Only trigger model type update if the record live array count actually changed

teddyzeenny:
Fixes https://github.com/emberjs/ember-inspector/issues/690

The issue:

In some cases, Ember Data's `peekAll` triggers `arrayContentDidChange` on the record array manager's live record array even if the records didn't change. One example case is when we call `unloadRecord` followed by `peekAll` within the same run loop.

The `peekAll` will trigger the array observer's `didChange` callback (with zero changes). This is generally harmless as long as we guard against that in our data adapter. Without the guard, we risk causing an infinite recursion because our `didChange` observer itself calls `peekAll`, which in the above scenario will re-trigger the observer and so on.

### DIFF
--- a/packages/ember-extension-support/lib/data_adapter.js
+++ b/packages/ember-extension-support/lib/data_adapter.js
@@ -294,8 +294,12 @@ export default EmberObject.extend({
     }
 
     let observer = {
-      didChange() {
-        run.scheduleOnce('actions', this, onChange);
+      didChange(array, idx, removedCount, addedCount) {
+        // Only re-fetch records if the record count changed
+        // (which is all we care about as far as model types are concerned).
+        if (removedCount > 0 || addedCount > 0) {
+          run.scheduleOnce('actions', this, onChange);
+        }
       },
       willChange() { return this; }
     };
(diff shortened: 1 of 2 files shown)


nkgm:
@rwjblue @stefanpenner could this make it to 2.14? Right now it has big negative impact on ember inspector.

samselikoff:
+1 on this, it's killing my dev workflow.

Is there a temporary workaround?

xomaczar:
Chrome freezes for me too - running ember@2.15 and ember-data@2.16.

rwjblue:
This was released in 2.15.2

DLiblik:
@rwjblue When will 2.15.2 release?  This is causing an issue in one of our applications as well.

locks:
@DLiblik https://libraries.io/npm/ember-source/2.15.2 was out on the 4th and https://libraries.io/npm/ember-source/2.15.3 on the 10th.

DLiblik:
@rwjblue My fault - poor question on my part: I meant through ember-cli.  Running npm to install ember-cli@2.15.2 fails on the version not being available.  The releases there also do not list 2.15.2 https://github.com/ember-cli/ember-cli/releases - it might just be that I'm not clear how the two versioning processes sync up.

Serabe:
Ember can be upgraded without having to upgrade Ember CLI.

PR #15606, opened 2017-08-28
[BUGFIX LTS] add fs-extra to deps

kellyselden:
Used [here](https://github.com/emberjs/ember.js/blob/98e01ceb9308433023d4e0bc371422526082b2f7/blueprints/route/index.js#L3) and [here](https://github.com/emberjs/ember.js/blob/d0de6ae3a77ab5a5156e62d3be7543302caf8c0f/node-tests/blueprints/route-test.js#L3) but missing from dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "ember-cli-valid-component-name": "^1.0.0",
     "ember-cli-version-checker": "^1.3.1",
     "ember-router-generator": "^1.2.3",
+    "fs-extra": "^4.0.1",
     "inflection": "^1.12.0",
     "jquery": "^3.2.1",
     "resolve": "^1.3.3",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2905,6 +2905,14 @@ fs-extra@^2.0.0:
     graceful-fs "^4.1.2"
     jsonfile "^2.1.0"
 
+fs-extra@^4.0.1:
+  version "4.0.1"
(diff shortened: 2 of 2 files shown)


stefanpenner:
LGTM, are we worried this might be brittle? (CI wont fail)

kellyselden:
One of my workflow scripts includes `npm prune` which is how I discovered this. I wonder how we could test for stuff like this.

stefanpenner:
> One of my workflow scripts includes npm prune which is how I discovered this. I wonder how we could test for stuff like this.

Maybe a form of linting, saying some dir tree is depDeps another is prod only or something? That would be a way to mitigate the concern (But someone can convince me the concern is not warrented)

kellyselden:
This should be good to go. It is better than it was previously, even though still brittle. @stefanpenner  or @rwjblue merge?

package.json
```diff
@@ -101,6 +101,7 @@
     "eslint-plugin-ember-internal": "^1.1.0",
     "express": "^4.15.2",
     "finalhandler": "^1.0.2",
+    "fs-extra": "^4.0.1",
```
rwjblue: I am confused. You said we are using this in blueprints, wouldn't this need to be in `dependencies` not `devDependencies`?

kellyselden: That's wierd. I don't know what I was thinking. Will fix.

kellyselden:
@rwjblue fixed but I can't get saucelabs to pass

rwjblue:
Ya, that’s not this PR’s issue.

PR #15612, opened 2017-08-30
Document Ember.js' release process

locks:


auvipy (commented):
that would be very good addition

acorncom (approved):
Would be great to have this in, any reason we shouldn't merge and iterate as needed?

PR #15613, opened 2017-08-30
[BUGFIX beta] Don't throw an error, when not all query params are passed to routerService.transitionTo

cvx:
This PR attempts to fix two issues:

1. A bug introduced in #15414.
  Using `transitionTo` via router service fail with an error, unless all query parameters of the target route are passed into the method. Here's the failing test: https://github.com/CvX/ember.js/blob/1c8c6ecad42228e94deffb7672fc6de905fca007/packages/ember/tests/routing/router_service_test/transitionTo_test.js#L259-L277.
  And for googling folks, the error message was ```Error: Assertion Failed: You passed the `false` query parameter during a transition into [route-name-here]}, please update to [query-param-here]```.
2. A problem reported in #4570 (and attempted to be fixed in #4571). This fix (originally by  @raytiley) is required to make the above use case work. Without it, transitioning to a route that has multiple query params ends up with all of them in the url with the value `null`, i.e.
  ```
    Ember.Controller.extend({
      queryParams: ['q', 'other', 'stuff']
    });

    // in some component
    this.get('router').transitionTo('articles', {
      queryParams: { q: 'test' },
    });

    // We end up at: /articles?q=test&other=null&stuff=null
    // Instead of (after the fix): /articles?q=test
  ```

cc: @rwjblue

### DIFF
--- a/packages/ember-routing/lib/system/route.js
+++ b/packages/ember-routing/lib/system/route.js
@@ -848,7 +848,10 @@ let Route = EmberObject.extend(ActionHandler, Evented, {
         } else {
           if (presentKey) {
             svalue = params[presentKey];
-            value = route.deserializeQueryParam(svalue, qp.urlKey, qp.type);
+
+            if (svalue !== undefined) {
+              value = route.deserializeQueryParam(svalue, qp.urlKey, qp.type);
+            }
           } else {
             // No QP provided; use default value.
             svalue = qp.serializedDefaultValue;
@@ -2127,7 +2130,7 @@ let Route = EmberObject.extend(ActionHandler, Evented, {
       }
     });
 	```
-	
+
     @method disconnectOutlet
     @param {Object|String} options the options hash or outlet name
     @since 1.0.0
--- a/packages/ember-routing/lib/system/router.js
+++ b/packages/ember-routing/lib/system/router.js
@@ -701,7 +701,9 @@ const EmberRouter = EmberObject.extend(Evented, {
     @param {String} type
   */
   _serializeQueryParam(value, type) {
-    if (type === 'array') {
+    if (value === null || value === undefined) {
+      return null;
+    } else if (type === 'array') {
       return JSON.stringify(value);
     }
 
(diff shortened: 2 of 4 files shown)


Serabe:
Restarted sauce labs

CvX:
Failed again, any idea what's wrong?
The message (`UnknownError - An unknown server-side error occurred while processing the command. Selenium error: Unable to get property 'bad' of undefined or null reference`) doesn't tell me much! 😁

DingoEatingFuzz:
Thank you @CvX! I ran into this while upgrading as well.

CvX:
Alright, dunno who did what, but sauce labs tests are now passing! 🎉

rwjblue:
Thanks for working on this! I'm not 100% sure that we want the null/undefined coercion as described in #4571 / #4570. Can you explain why that change was required in order to fix the router service test case you added?

CvX:
Null/undefined fix isn't *needed* for the router service exception fix, but it just feels wrong to write a test that expects the url to have "queryParams=null". 😅 (see snippet in the original comment of this PR)

rmachielse:
Is there anything blocking at least the fix for the first issue from being merged? It is keeping us from upgrading to 2.15

toranb:
ping @rwjblue ^^ any further debate or discussion required on this PR boss?

pichfl:
We're waiting desperately for this to be released, any chance this will show up in 2.16?

alexdiliberto:
Any updates here? What are the specifics for not allowing the type coercion?

This issue prevents our app from using the router service in several places. It just doing feel right that doing the same transition from a template `{{#link-to}}` and some random route `transitionTo()` don't behave the same the as a component's `get(this, 'router').transitionTo()`

toranb:
Any word from ember core team here? I know ember v2.16.0 shipped without this fix ...

CvX:
Alright, to make it easier to move this forward I'm splitting the PR into two:

* #15726 - Fixes the bug in routing service. Should be an easy, non-controversial merge 😉 
* #15727 - Brings back the query param type coercion PR. Open to discussion.

rwjblue:
Apologies for not getting this in for 2.16, I'll get the changes into the next 2.17 beta release and once some folks can confirm all is working well for us I'll pull down into 2.16 for the LTS.

rwjblue:
@CvX - Thank you very much for working so hard on this!

PR #15623, opened 2017-09-01
[INTERNAL] Use JSON.parse to fetch version number

locks:
The json gem has been part of the Ruby stdlib for a while now.
With this change, we mirror what is done in the Ember Data codebase.


ember-source.gemspec
```diff
@@ -11,7 +11,7 @@ Gem::Specification.new do |gem|
   gem.homepage      = "https://github.com/emberjs/ember.js"
   gem.license       = 'MIT'
 
-  gem.version       = Ember.rubygems_version_string
+  gem.version       = Ember::VERSION
 
   gem.files = %w(VERSION) + Dir['dist/*.js', 'lib/ember/*.rb']
```
rwjblue: I think this also needs updating?

locks: Added `package.json`.
Question, though, do we still need that `VERSION` file?

locks: It seems to be in use in the deploy script:

    gem push ember-source-`cat VERSION`.gem

Serabe:
Restarted sauce :(

locks:
It's green, can someone review?

stefanpenner:
:+1:

PR #15632, opened 2017-09-05
avoid boolean coercion for meta values

bekzod:


bekzod:
sauce failing here

rwjblue:
restarted

locks:
@rwjblue green after another restart

PR #15633, opened 2017-09-05
[BUGFIX beta] getParentEngine was not being exported under Ember.

### LINKED ISSUE #15630: Ember.{get,set}EngineParent not exported
According to the [documentation](https://www.emberjs.com/api/ember/2.14/classes/Ember/methods/getEngineParent?anchor=getEngineParent) functions `getEngineParent` and `setEngineParent` should be accessible on the `Ember` global, but it seems that they are not exported.

Serabe:
As reported in #15630, these were marked as public but were not being
exported.

Fixes #15630

### DIFF
--- a/packages/ember/lib/index.js
+++ b/packages/ember/lib/index.js
@@ -580,6 +580,7 @@ Ember.ApplicationInstance = application.ApplicationInstance;
 Ember.Engine = application.Engine;
 Ember.EngineInstance = application.EngineInstance;
 Ember.DefaultResolver = Ember.Resolver = application.Resolver;
+Ember.getEngineParent = application.getEngineParent;
 
 runLoadHooks('Ember.Application', application.Application);
 
--- a/packages/ember/tests/global-api-test.js
+++ b/packages/ember/tests/global-api-test.js
@@ -1,5 +1,6 @@
 import { get } from 'ember-metal';
 import { isArray } from 'ember-runtime';
+import { getEngineParent } from 'ember-application';
 
 QUnit.module('Global API Tests');
 
@@ -18,3 +19,4 @@ confirmExport('Ember.generateController');
 confirmExport('Ember.Helper');
 confirmExport('Ember.Helper.helper');
 confirmExport('Ember.isArray', isArray);
+confirmExport('Ember.getEngineParent', getEngineParent);


rwjblue:
@dgeb - Can you chime in here and confirm that these were intended to be public API?

Serabe:
@rwjblue @dgeb fix a problem with this PR. While `getEngineParent` was marked as both exported an public, `setEngineParent` was marked as private.

I opened #15636 in case you want to mark it private, to ease the fix.

Thank you!

dgeb:
It seems pretty clear to me that `setEngineParent` should remain private, since parent re/assignment is the domain of the framework. To a much lesser extent, exposing `getEngineParent` publicly also feels like a boundary violation. I'd prefer to keep it private (as per #15636) unless a convincing use case surfaces.

jeloba:
I want to use it in testing: There is some state that I want to pass to an engine during integration tests in its dummy app. It concerns translation and is not a relevant API of the engine, so I don't want to expose this stuff.

Currently it is not possible to get the engine instance from the application (in an initializer) or the application from the engine. I had to introduce global state (during testing) to pass my settings. Its may cause flaky tests if. Assigning state to the engine or application instance would be preferable. Best solution would be to make the instantiated engine available in tests.

dgeb:
> There is some state that I want to pass to an engine during integration tests in its dummy app. It concerns translation and is not a relevant API of the engine, so I don't want to expose this stuff.

@jeloba Do you want your engine to "reach up" into the application instance to grab the state it needs (but only in testing)? Doesn't your engine need similar translation data in production? If so, how is it passed? Can you use a service that's shared between the application and engine?

jeloba:
@dgeb "reaching up" was an idea, but not my preferred one. What I'm doing is reopening the the i18n service and changing how it translates (adding asserts, translating to "") inside a test helper. This special behavior should only be activated in tests that use my helpers. This is the state I'm talking about. The engine and the application should know nothing about this, since its just a test feature.

The best solution would be accessing the i18n service on the engine inside an instance initializer, and set enable the debug behavior during app startup in the test. But I could not find a way to do this.

Serabe:
@dgeb can you check @jeloba's last comment? Thank you!

@jeloba I fix the link, a link to a gist ending in `/edit` cannot be seen by other people. Hope you don't mind. Thank you!

Serabe:
@dgeb can you check @jeloba's last comment? Thank you!

dgeb:
@Serabe @jeloba I'm sorry to delay with a response.

> The best solution would be accessing the i18n service on the engine inside an instance initializer, and set enable the debug behavior during app startup in the test. But I could not find a way to do this.

@jeloba I think I agree with this assessment. Allowing test-only instance initializers would be a cleaner approach than either your current solution or the solution that involves exposing `getEngineParent`. And I'm not sure that enabling a slightly more elegant workaround, like the commented out approach in the gist, warrants making `getEngineParent` public.

Let's instead focus on enabling cleaner test setup for engines ASAP.

jeloba:
@dgeb, you are right lets go for the cleaner approach and fix the docs.

dgeb:
Going with #15636 instead. Thanks for providing both options @Serabe!

PR #15636, opened 2017-09-05
[DOC beta] Mark getEngineParent as private

### LINKED ISSUE #15630: Ember.{get,set}EngineParent not exported

Serabe:
Fixes #15630 

Alternative to #15633


dgeb:
Thanks @Serabe!

PR #15638, opened 2017-09-06
Stash test timeout on pauseTest

ro0gr:
### Reason
Currently `pauseTest` do nothing special with `QUnit.config.testTimeout`. It leads to pure dev experience out of the box - [Demo](https://ember-twiddle.com/b9d380a0110f11ae1bebf2f0bdee93ad?openFiles=tests.acceptance.my-acceptance-test.js%2C)

The same issue is also valid for mocha - #15593 

### Changes
This pr extends `Test.Adapter` with a new [`stashTimeout()`](https://github.com/emberjs/ember.js/compare/master...ro0gr:resume-timeout?expand=1#diff-d30d6f515bf6770d40cb24e9bc3e7664R48) method. This potentially allows `pauseTest` to behave more correct.

---

Does the change of this kind require a RFC? 🙉 
The reason why I'm asking is that technically `Test.Adapter` is [public](https://emberjs.com/api/ember/2.15/classes/Ember.Test.Adapter). But in [fact](https://emberobserver.com/code-search?codeQuery=Test.Adapter) seems like it's used only by `ember-source`.

### DIFF
--- a/packages/ember-testing/lib/adapters/adapter.js
+++ b/packages/ember-testing/lib/adapters/adapter.js
@@ -36,6 +36,18 @@ export default EmberObject.extend({
   asyncEnd: K,
 
   /**
+    Removes test timeout for current test.
+
+    This allows `pauseTest` test helper to wait forever.
+    It should also store an original test timeout value in order to restore it on `resumeTest()`.
+
+    @public
+    @method stashTimeout
+    @return {function} Function that restores timeout
+  */
+  stashTimeout: () => K,
+
+  /**
     Override this method with your testing framework's false assertion.
     This function is called whenever an exception occurs causing the testing
     promise to fail.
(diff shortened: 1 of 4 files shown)


ro0gr:
At this moment I'm interested if this change looks reasonable in general. That's why there is no implementation for `qunit` adapter.

ro0gr:
This can be solved in ember-qunit and ember-mocha lands [now](https://github.com/emberjs/ember-test-helpers/commit/f065e1fb5646d1cdd4c72e52dfe25496b8dbabea)

rwjblue:
FWIW, the ember-qunit 3.0.0 (published today) adds custom wrapping to ensure `assert.timeout(-1)` is used (preventing timeouts). But it only exists (at the moment) in the newer style nested module API...

PR #15643, opened 2017-09-07
run Travis on tagged builds

kellyselden:
This reverts https://github.com/emberjs/ember.js/commit/a3712c52692481dfdb368c48ec43b142d7488c18 and adds a tag regex. Tested [here](https://travis-ci.org/kellyselden/ember-awesome-macros/jobs/273034825) and is working as expected.

cc @rwjblue

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,14 @@ install:
 after_success:
   - "./bin/publish_builds"
 
+branches:
+  only:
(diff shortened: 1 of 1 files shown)


rwjblue:
This branch will age out, but it would be great to fix this in master so it will trickle down for future releases.

PR #15651, opened 2017-09-10
[BUGFIX] Track applicationInstances created by Ember.Application for teardown

### LINKED ISSUE #15327: Error during initial render causes non-globals `Application` visit promise to hang

kpfefferle:
Resolves #15327 

The fixes here occur within `packages/ember-application/lib/system/application.js`:
- Add a private/internal `_applicationInstances` array property to `Ember.Application`
- Push each `applicationInstance` onto the `_applicationInstances` array each time one is created with `buildInstance()`
- Splice an `applicationInstance` out of the `_applicationInstances` array if it is destroyed so that we're not leaking instances after they have been destroyed
- During `willDestroy()`, check for the length of `_applicationInstances` and call `destroy()` on each one to make sure we are tearing down all `applicationInstances`

I have also refactored a couple of tests that were using Ember globals mode to work around the bug described in #15327. There are still a couple of tests that need to be refactored (noted below).

TODO:
- [x] Refactor [component_registration_test](https://github.com/emberjs/ember.js/blob/98cc07f7efeeabed7524511966a510e0470b1000/packages/ember/tests/component_registration_test.js#L170-L197) to not use globals mode workaround ~~_(note that this *entire* test file is using Ember's globals mode, so I'm unsure if the entire file needs refactored or if the note simply needs removed from this one test)_~~
- [x] Refactor [substates_test](https://github.com/emberjs/ember.js/blob/7dd709422e1961dffe65a967fc48b0e5940a0200/packages/ember/tests/routing/substates_test.js#L441-L502) to not use globals mode workaround
- [x] Add unit tests to `Ember.Application`:
  - [x] Test that each applicationInstance created by `buildInstance()` gets added to `_applicationInstances`
  - [x] Test that a destroyed applicationInstance gets spliced out of `_applicationInstances`
  - [x] Test that `willDestroy()` calls `destroy()` on each applicationInstance that has been added to `_applicationInstances`

cc: @mixonic

### DIFF
--- a/packages/ember-application/lib/system/application.js
+++ b/packages/ember-application/lib/system/application.js
@@ -329,6 +329,17 @@ const Application = Engine.extend({
   */
   _globalsMode: true,
 
+  /**
+    An array of application instances created by `buildInstance()`. Used
+    internally to ensure that all instances get destroyed.
+
+    @property _applicationInstances
+    @type Array
+    @default null
+    @private
+  */
+  _applicationInstances: null,
+
   init(options) { // eslint-disable-line no-unused-vars
     this._super(...arguments);
 
@@ -346,6 +357,7 @@ const Application = Engine.extend({
     // the Application's own `boot` method.
     this._readinessDeferrals = 1;
     this._booted = false;
+    this._applicationInstances = [];
 
     this.autoboot = this._globalsMode = !!this.autoboot;
 
@@ -369,6 +381,31 @@ const Application = Engine.extend({
     options.base = this;
     options.application = this;
     return ApplicationInstance.create(options);
+  },
+
+  /**
+    Start tracking an ApplicationInstance for this application.
+    Used when the ApplicationInstance is created.
+
+    @private
+    @method _watchInstance
+  */
+  _watchInstance(instance) {
+    this._applicationInstances.push(instance);
+  },
+
+  /**
+    Stop tracking an ApplicationInstance for this application.
+    Used when the ApplicationInstance is about to be destroyed.
+
+    @private
+    @method _unwatchInstance
+  */
+  _unwatchInstance(instance) {
+    let index = this._applicationInstances.indexOf(instance);
+    if (index > -1) {
+      this._applicationInstances.splice(index, 1);
+    }
   },
 
   /**
@@ -761,8 +798,9 @@ const Application = Engine.extend({
       _loaded.application = undefined;
     }
 
-    if (this._globalsMode && this.__deprecatedInstance__) {
-      this.__deprecatedInstance__.destroy();
+    if (this._applicationInstances.length) {
+      this._applicationInstances.forEach(i => i.destroy());
+      this._applicationInstances.length = 0;
     }
   },
 
--- a/packages/ember/tests/component_registration_test.js
+++ b/packages/ember/tests/component_registration_test.js
@@ -1,162 +1,178 @@
+import { Application } from 'ember-application';
 import { Controller } from 'ember-runtime';
 import { Component } from 'ember-glimmer';
-import { moduleFor, AutobootApplicationTestCase } from 'internal-test-helpers';
+import { compile } from 'ember-template-compiler';
+import { moduleFor, ApplicationTestCase } from 'internal-test-helpers';
 
-moduleFor('Application Lifecycle - Component Registration', class extends AutobootApplicationTestCase {
+moduleFor('Application Lifecycle - Component Registration', class extends ApplicationTestCase {
+
+  // This is necessary for this.application.instanceInitializer to not leak between tests
+  createApplication(options) {
+    return super.createApplication(options, Application.extend());
+  }
 
   ['@feature(!ember-glimmer-template-only-components) The helper becomes the body of the component']() {
-    this.runTask(() => {
-      this.createApplication();
+    this.addTemplate('components/expand-it', '<p>hello {{yield}}</p>');
+    this.addTemplate('application', 'Hello world {{#expand-it}}world{{/expand-it}}');
 
-      this.addTemplate('components/expand-it', '<p>hello {{yield}}</p>');
-      this.addTemplate('application', 'Hello world {{#expand-it}}world{{/expand-it}}');
+    return this.visit('/').then(() => {
+      this.assertText('Hello world hello world');
+      this.assertComponentElement(this.element.firstElementChild, { tagName: 'div', content: '<p>hello world</p>' });
     });
-
-    this.assertText('Hello world hello world');
-    this.assertComponentElement(this.element.firstElementChild, { tagName: 'div', content: '<p>hello world</p>' });
   }
 
   ['@feature(ember-glimmer-template-only-components) The helper becomes the body of the component']() {
-    this.runTask(() => {
-      this.createApplication();
+    this.addTemplate('components/expand-it', '<p>hello {{yield}}</p>');
+    this.addTemplate('application', 'Hello world {{#expand-it}}world{{/expand-it}}');
 
-      this.addTemplate('components/expand-it', '<p>hello {{yield}}</p>');
-      this.addTemplate('application', 'Hello world {{#expand-it}}world{{/expand-it}}');
+    return this.visit('/').then(() => {
+      this.assertInnerHTML('Hello world <p>hello world</p>');
     });
-
-    this.assertInnerHTML('Hello world <p>hello world</p>');
   }
 
   ['@test If a component is registered, it is used'](assert) {
-    this.runTask(() => {
-      this.createApplication();
+    this.addTemplate('components/expand-it', '<p>hello {{yield}}</p>');
+    this.addTemplate('application', `Hello world {{#expand-it}}world{{/expand-it}}`);
 
-      this.addTemplate('components/expand-it', '<p>hello {{yield}}</p>');
-      this.addTemplate('application', `Hello world {{#expand-it}}world{{/expand-it}}`);
-
-      this.applicationInstance.register('component:expand-it', Component.extend({
-        classNames: 'testing123'
-      }));
+    this.application.instanceInitializer({
+      name: 'expand-it-component',
+      initialize(applicationInstance) {
+        applicationInstance.register('component:expand-it', Component.extend({
+          classNames: 'testing123'
+        }));
+      }
     });
 
-    let text = this.$('div.testing123').text().trim();
-    assert.equal(text, 'hello world', 'The component is composed correctly');
+    return this.visit('/').then(() => {
+      let text = this.$('div.testing123').text().trim();
+      assert.equal(text, 'hello world', 'The component is composed correctly');
+    });
   }
 
   ['@test Late-registered components can be rendered with custom `layout` property'](assert) {
-    this.runTask(() => {
-      this.createApplication();
+    this.addTemplate('application', `<div id='wrapper'>there goes {{my-hero}}</div>`);
 
-      this.addTemplate('application', `<div id='wrapper'>there goes {{my-hero}}</div>`);
-
-      this.applicationInstance.register('component:my-hero', Component.extend({
-        classNames: 'testing123',
-        layout: this.compile('watch him as he GOES')
-      }));
+    this.application.instanceInitializer({
+      name: 'my-hero-component',
+      initialize(applicationInstance) {
+        applicationInstance.register('component:my-hero', Component.extend({
+          classNames: 'testing123',
+          layout: compile('watch him as he GOES')
+        }));
+      }
     });
 
-    let text = this.$('#wrapper').text().trim();
-    assert.equal(text, 'there goes watch him as he GOES', 'The component is composed correctly');
+    return this.visit('/').then(() => {
+      let text = this.$('#wrapper').text().trim();
+      assert.equal(text, 'there goes watch him as he GOES', 'The component is composed correctly');
+    });
   }
 
   ['@test Late-registered components can be rendered with template registered on the container'](assert) {
-    this.runTask(() => {
-      this.createApplication();
+    this.addTemplate('application', `<div id='wrapper'>hello world {{sally-rutherford}}-{{#sally-rutherford}}!!!{{/sally-rutherford}}</div>`);
 
-      this.addTemplate('application', `<div id='wrapper'>hello world {{sally-rutherford}}-{{#sally-rutherford}}!!!{{/sally-rutherford}}</div>`);
-
-      this.applicationInstance.register('template:components/sally-rutherford', this.compile('funkytowny{{yield}}'));
-      this.applicationInstance.register('component:sally-rutherford', Component);
+    this.application.instanceInitializer({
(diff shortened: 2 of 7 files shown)


rwjblue (commented):
As it stands, this leaks `ApplicationInstance` instances in things like fastboot and non `packages/ember-testing` related tests. In those circumstances many `ApplicationInstance` instances are created and "thrown away" to be garbage collected.

I think an alternative way to solve the same general problem would be to have the `ApplicationInstance` push itself into the `Application`'s `_applicationInstances ` array, then splice itself out of that array in the `ApplicationInstance#willDestroy`.  This setup would avoid leaking the `ApplicationInstance` when things are handled properly (e.g. in non-error conditions) and still provide the cleanup that is needed when things go haywire in a given test...

packages/ember-application/lib/system/application.js
```diff
@@ -762,6 +776,10 @@ const Application = Engine.extend({
       _loaded.application = undefined;
     }
 
+    if (this._applicationInstances.length) {
+      this._applicationInstances.forEach(i => i.destroy());
```
rwjblue: Should reset the array after looping through and destroying, thinking `this._applicaitonInstances.length = 0` or something...

packages/ember-application/lib/system/application.js
```diff
@@ -762,6 +776,10 @@ const Application = Engine.extend({
...
+    }
+
     if (this._globalsMode && this.__deprecatedInstance__) {
```
rwjblue: Seems like this might be able to be removed now? `this.__deprecatedInstance__` is assigned to the result of calling `this.buildInstance()`, so the `this.__deprecatedInstance__` would already have been destroyed by the loop above...

rwjblue (commented):
The test cleanup here is looking really nice! Please keep it up!

Still have the same concerns as mentioned previously about leakage. As of https://github.com/emberjs/ember-test-helpers/pull/237 (which is included in `ember-cli@2.17.0` out of the box) we create a single application instance and call `application.buildInstance()` once for each test in the application (and subsequently destroy them during normal test teardown). The array created to track all created `Ember.ApplicationInstance` instances **will** leak the entire contents of the container for each test (the same leak will occur in fastboot per request).

The simple fix (as mentioned above) is to have the `ApplicationInstance` instances splice their own entry out of the array (possibly by adding a `Ember.Application.prototype.destroyInstance` which would be the inverse of `Ember.Application.prototype.buildInstance`).

rwjblue:
Ping?

kpfefferle:
@rwjblue I made some solid progress on this over the break. I should be able to push up a review-ready update by the end of the week.

krisselden:
@kpfefferle this is blocking me at the moment, I need to wrap up this PR, so if you want me to base it off this PR or your solid WIP please push your latest.

krisselden:
@rwjblue should this target [BUGFIX beta]?

kpfefferle:
@krisselden @rwjblue I believe that this is ready for final review now. If you need me to target a different branch (or anything else), let me know and I will take care of it tomorrow.

packages/ember/tests/component_registration_test.js
```diff
@@ -1,162 +1,177 @@
...
+      initialize(applicationInstance) {
+        applicationInstance.register('component:expand-it', Component.extend({
+          classNames: 'testing123'
+        }));
+      }
     });
 
-    let text = this.$('div.testing123').text().trim();
-    assert.equal(text, 'hello world', 'The component is composed correctly');
+    this.visit('/').then(() => {
```
rwjblue: Needs a return

packages/ember/tests/component_registration_test.js
```diff
@@ -1,162 +1,177 @@
...
+    this.application.instanceInitializer({
+      name: 'borf-snorlax-component-template',
+      initialize(applicationInstance) {
+        applicationInstance.register('template:components/borf-snorlax', compile('goodfreakingTIMES{{yield}}'));
+      }
     });
 
-    let text = this.$('#wrapper').text().trim();
-    assert.equal(text, 'hello world goodfreakingTIMES-goodfreakingTIMES!!!', 'The component is composed correctly');
+    this.visit('/').then(() => {
```
rwjblue: return

packages/ember/tests/component_registration_test.js
```diff
@@ -1,162 +1,177 @@
...
+        applicationInstance.register('component:my-component', Component.extend({
+          text: 'inner',
+          layoutName: 'foo-bar-baz'
+        }));
+      }
     });
 
-    let text = this.$('#wrapper').text().trim();
-    assert.equal(text, 'inner-outer', 'The component is composed correctly');
+    this.visit('/').then(() => {
```
rwjblue: return

packages/ember/tests/component_registration_test.js
```diff
@@ -1,162 +1,177 @@
...
+          text: 'inner',
+          layoutName: 'foo-bar-baz',
+          layout: compile('{{text}}-{{yield}}')
+        }));
+      }
     });
 
-    let text = this.$('#wrapper').text().trim();
-    assert.equal(text, 'inner-outer', 'The component is composed correctly');
+    this.visit('/').then(() => {
```
rwjblue: return

packages/ember/tests/component_registration_test.js
```diff
@@ -1,162 +1,177 @@
...
   ['@test Using name of component that does not exist']() {
-    expectAssertion(() => {
-      this.runTask(() => {
-        this.createApplication();
+    this.addTemplate('application', `<div id='wrapper'>{{#no-good}} {{/no-good}}</div>`);
 
-        this.addTemplate('application', `<div id='wrapper'>{{#no-good}} {{/no-good}}</div>`);
-      });
+    expectAssertion(() => {
+      this.visit('/');
```
rwjblue: I don't know exactly how to handle this case properly. Currently `expectAssertion` doesnt' deal with async, the prior `runTask` wrapping was forcing it to be "pseudo-sync"

packages/ember/tests/routing/substates_test.js
```diff
@@ -436,64 +433,41 @@ moduleFor('Loading/Error Substates', class extends ApplicationTestCase {
...
+      <p id="toplevel-error">TOPLEVEL ERROR: {{model.msg}}</p>
+    `);
 
-    let text = this.$('#toplevel-error').text();
-    assert.equal(
-      text,
-      'TOPLEVEL ERROR: BAD NEWS BEARS',
-      'toplevel error rendered'
-    );
+    this.visit('/').then(() => {
```
rwjblue: return

packages/ember/tests/routing/substates_test.js
```diff
@@ -436,64 +433,41 @@ moduleFor('Loading/Error Substates', class extends ApplicationTestCase {
...
+        text,
+        'TOPLEVEL ERROR: BAD NEWS BEARS',
+        'toplevel error rendered'
+      );
+    });
 
     reject = false;
 
-    return this.visit('/').then(() => {
+    this.visit('/').then(() => {
```
rwjblue: return

packages/ember/tests/component_registration_test.js
kpfefferle: Would something like this work?
```js
expectAssertion(() => {
  this.runTask(() => this.visit('/'));
}, /.* named "no-good" .*/);
```
The test passes either way, so I'm not sure what the proper refactor here is.

mixonic:
@kpfefferle can you squash into a single commit and prefix the commit message with `[BUGFIX beta]` please? It can still target `master`.

packages/ember/tests/component_registration_test.js
rwjblue: The test passes either way because `this.visit` is sync at the moment 😭 .

@mixonic reminded me that `expectAssertion(/some regexp/)` can be used to handle the async cases like this, maybe we could swap to that instead?

packages/ember-application/lib/system/application.js
```diff
@@ -368,7 +380,23 @@ const Application = Engine.extend({
   buildInstance(options = {}) {
     options.base = this;
     options.application = this;
-    return ApplicationInstance.create(options);
+    let instance = ApplicationInstance.create(options);
+    this._applicationInstances.push(instance);
```
rwjblue: I would vaguely prefer to have this done by the `ApplicationInstance`'s `init` method. In my head it pairs nicely with the fact that `ApplicationInstance` is now the one responsible for deregistering.

I don't feel super strongly... What do you think?

packages/ember-application/lib/system/application.js
```diff
@@ -368,7 +380,23 @@ const Application = Engine.extend({
...
+  },
+
+  /**
+    Stop tracking an ApplicationInstance for this application.
+    Used when the ApplicationInstance is about to be destroyed.
+
+    @private
+    @method unwatchInstance
+  */
+  unwatchInstance(instance) {
```
rwjblue: Should probably be underscored to make it being private more obvious...

packages/ember/tests/component_registration_test.js
mixonic: Looked into this and although `expectDeprecation` has an async-friendly form `expectAssertion` does not currently. I think a `TODO: Use the async form of expectAssertion here when it is available` would be sufficient.

packages/ember-application/lib/system/application.js
mixonic: Seems like we should make a small refactor to:

* `Application#_unwatchInstance` - does what `unwatchInstance` does now.
* `Application#_watchInstance` - is called by `ApplicationInstance#init` at the beginning of that method. Does the watching bit here.

rwjblue: Sounds good to me!

packages/ember/tests/routing/substates_test.js
```diff
@@ -436,64 +433,41 @@ moduleFor('Loading/Error Substates', class extends ApplicationTestCase {
...
+      assert.equal(
+        text,
+        'TOPLEVEL ERROR: BAD NEWS BEARS',
+        'toplevel error rendered'
+      );
+    });
 
     reject = false;
 
     return this.visit('/').then(() => {
```
rwjblue: I think this needs to be updated to chain off of the higher up `return this.visit(...)` or it is unreachable...

mixonic:
Ok, I pushed a little more cleanup in a force push to this branch.

* Refactor toward two private APIs on the application for watch/unwatch.
* Fix an early `return` in a test which @rwjblue identified.
* Refactored some tests that were testing internal state on the private-API tracking array to instead test public API behaviors. When an application is destroyed, all application instances it created should also be destroyed.

@krisselden has pointed out that there are several related issues here. The goal of this PR is specifically:

* When an application is destroyed, the application instances it created should be destroyed. Before this patch autoboot application would always correctly do this. After this patch non-autoboot applications also teardown any instances they created.
* Several tests in master use autoboot just so the teardown during testing works correctly. Since we're fixing non-autoboot application to also teardown instances, we can remove the autoboot test helpers (simplifying the tests).

I've changed this back to targeting canary as well, since we don't actually know of any test suites or real app cases impact buy this (though you could theorize them).

mixonic (approved):

kpfefferle:
I added the above-mentioned `TODO` and then force-pushed. Hopefully the browser tests won't "unexpectedly exit" this time.

rwjblue:
That CI job will fail until https://github.com/emberjs/ember.js/issues/16065 is resolved.

mixonic:
Great, calling this solid then!

PR #15653, opened 2017-09-11
assert when multiple args passed with same name to component

### LINKED ISSUE #15618: [v2.15.0] Bug declaring the same component attribute twice (creates a mutableCell)
Hi, I just updated from 2.14 to 2.15 and found a weird "bug". By mistake I was rendering a component passing the same attr twice, like this: `{{my-component componentVar=myStringVar componentVar=myStringVar}}`

I know this makes no sense, but it was working correctly in 2.14
In 2.15 it doesn't work. Suddenly `componentVar` is no longer the `myStringVar` but instead is a mutable cell object.

I created a twiddle with a reproduction, check the console also.
https://ember-twiddle.com/c1f4d778120fe303436b0e5df029fcf2

Not saying this should be supported, but maybe throw an error? It can be difficult to debug the way it is.

bekzod:
attempt to fix #15618, not sure if it is right way solving the issue

### DIFF
--- a/packages/ember-glimmer/lib/utils/process-args.ts
+++ b/packages/ember-glimmer/lib/utils/process-args.ts
@@ -1,3 +1,4 @@
+import { assert } from 'ember-debug';
 import { symbol } from 'ember-utils';
 import { MUTABLE_CELL } from 'ember-views';
 import { CapturedNamedArguments } from '@glimmer/runtime';
@@ -5,25 +6,32 @@ import { ARGS } from '../component';
 import { ACTION } from '../helpers/action';
 import { UPDATE } from './references';
 
-// ComponentArgs takes EvaluatedNamedArgs and converts them into the
+// processComponentArgs takes CapturedNamedArguments and converts them into the
 // inputs needed by CurlyComponents (attrs and props, with mutable
 // cells, etc).
-export function processComponentArgs(namedArgs: CapturedNamedArguments) {
+export function processComponentArgs(namedArgs: CapturedNamedArguments, componentName: string) {
   let keys = namedArgs.names;
   let attrs = namedArgs.value();
   let props = Object.create(null);
   let args = Object.create(null);
 
-  props[ARGS] = args;
+  let seen;
+  if (DEBUG) {
+    seen = Object.create(null);
+  }
 
   for (let i = 0; i < keys.length; i++) {
     let name = keys[i];
+
+    assert(
+      `Component attribute \`${name}\` is declared multiple times for \`${componentName.split(':').slice(-1)}\` component.`,
+      seen[name] ? false : seen[name] = true
+    );
+
     let ref = namedArgs.get(name);
     let value = attrs[name];
 
-    if (typeof value === 'function' && value[ACTION]) {
-      attrs[name] = value;
-    } else if (ref[UPDATE]) {
+    if (!(typeof value === 'function' && value[ACTION]) && ref[UPDATE]) {
       attrs[name] = new MutableCell(ref, value);
     }
 
@@ -31,6 +39,7 @@ export function processComponentArgs(namedArgs: CapturedNamedArguments) {
     props[name] = value;
   }
 
+  props[ARGS] = args;
   props.attrs = attrs;
 
   return props;
--- a/packages/ember-glimmer/lib/component-managers/curly.ts
+++ b/packages/ember-glimmer/lib/component-managers/curly.ts
@@ -199,7 +199,7 @@ export default class CurlyComponentManager extends AbstractManager<ComponentStat
     let factory = definition.ComponentClass;
 
     let capturedArgs = args.named.capture();
-    let props = processComponentArgs(capturedArgs);
+    let props = processComponentArgs(capturedArgs, definition.name);
 
     aliasIdToElementId(args, props);
 
@@ -337,13 +337,13 @@ export default class CurlyComponentManager extends AbstractManager<ComponentStat
     let { component, args, argsRevision, environment } = bucket;
 
     if (DEBUG) {
-       this._pushToDebugStack(component._debugContainerKey, environment);
+      this._pushToDebugStack(component._debugContainerKey, environment)
     }
 
     bucket.finalizer = _instrumentStart('render.component', rerenderInstrumentDetails, component);
 
     if (!args.tag.validate(argsRevision)) {
-      let props = processComponentArgs(args);
+      let props = processComponentArgs(args, component._debugContainerKey);
 
       bucket.argsRevision = args.tag.value();
 
--- a/packages/ember-glimmer/tests/integration/components/curly-components-test.js
+++ b/packages/ember-glimmer/tests/integration/components/curly-components-test.js
@@ -3172,4 +3172,15 @@ moduleFor('Components test: curly components', class extends RenderingTest {
 
     this.assertText('3');
   }
+
+  ['@test passing attributes with same name multiple times throws']() {
+    let FooBarComponent = Component.extend({});
+
+    this.registerComponent('foo-bar', { ComponentClass: FooBarComponent });
+
+    expectAssertion(() => {
+      this.render('{{foo-bar foo="bar" foo="fizz"}}');
+    }, /Component attribute `foo` is declared multiple times for `foo-bar` component./);
+  }
+
 });


packages/ember-glimmer/lib/utils/process-args.js
```diff
@@ -1,10 +1,11 @@
...
 import { symbol } from 'ember-utils';
 import { ARGS } from '../component';
 import { UPDATE } from './references';
 import { MUTABLE_CELL } from 'ember-views';
 import { ACTION } from '../helpers/action';
 
-// ComponentArgs takes EvaluatedNamedArgs and converts them into the
+// processComponentArgs takes CapturedNamedArguments and converts them into the
 // inputs needed by CurlyComponents (attrs and props, with mutable
 // cells, etc).
```
alvincrespo: Should this be a YUI doc comment?

packages/ember-glimmer/lib/utils/process-args.js
```diff
@@ -13,23 +14,26 @@ export function processComponentArgs(namedArgs) {
...
+    if (seen[name]) { continue; }
+    seen[name] = true;
+
     let ref = namedArgs.get(name);
     let value = attrs[name];
 
-    if (typeof value === 'function' && value[ACTION]) {
-      attrs[name] = value;
-    } else if (ref[UPDATE]) {
+    if (!(typeof value === 'function' && value[ACTION]) && ref[UPDATE]) {
```
alvincrespo: If seen this check `typeof value === 'function' in multiple parts of the codebase, curious if there is a "helper" we can import to use?

alvincrespo: I like this cleanup 👏

bekzod: @alvincrespo inline `typeof value === 'function` is more optimizable by v8 than imported `isFunction` thats why you see it everywhere in codebase, most of things done inline

alvincrespo: Oh interesting, I did not know that. Thanks for the details here!

Serabe:
Can you add a test, please?

Thank you!

bekzod:
test added

packages/ember-glimmer/lib/utils/process-args.js
rwjblue: Probably not, it is more of an "implementors comment" than something written for Ember user's...

packages/ember-glimmer/lib/utils/process-args.js
```diff
@@ -13,23 +14,26 @@ export function processComponentArgs(namedArgs) {
   let props = Object.create(null);
   let args = Object.create(null);
 
-  props[ARGS] = args;
-
+  let seen = {};
```
rwjblue: IMHO, this should be a `DEBUG` only feature (it is used only for an assertion).

rwjblue: Also, should probably be `Object.create(null)`

packages/ember-glimmer/lib/utils/process-args.js
```diff
@@ -13,23 +14,26 @@ export function processComponentArgs(namedArgs) {
...
   for (let i = 0; i < keys.length; i++) {
     let name = keys[i];
+
+    assert(`Component attribute \`${name}\` is declared multiple times.`, !seen[name]);
```
rwjblue: We also need to include information about the name of the component and source template if possible. 

Getting the source template will likely prove to be difficult in JS-land, but I believe an AST transform could catch the most common cases...

rwjblue (changes requested):

packages/ember-glimmer/lib/utils/process-args.js
bekzod: added component name, but unsure how to get source :P

packages/ember-glimmer/lib/component-managers/curly.js
```diff
@@ -307,13 +307,13 @@ export default class CurlyComponentManager extends AbstractManager {
...
     if (DEBUG) {
-       this._pushToDebugStack(component._debugContainerKey, environment)
+      this._pushToDebugStack(component._debugContainerKey, environment)
     }
 
     bucket.finalizer = _instrumentStart('render.component', rerenderInstrumentDetails, component);
 
     if (!args.tag.validate(argsRevision)) {
-      let props = processComponentArgs(args);
+      let props = processComponentArgs(args, component._debugContainerKey.split(':')[1]);
```
rwjblue: This `.split(':')` is only needed for debug builds, can you restructure things to avoid doing the extra work in production?

rwjblue (changes requested):

packages/ember-glimmer/lib/component-managers/curly.js
bekzod: does `DEBUG` work inline ? 
can I do this `DEBUG ? component._debugContainerKey.split(':')[1] : '' ` ?
or there is better way

rwjblue: It does work in line.  `DEBUG` is replaced at build time with the literal `true` or `false` (depending on build type).  If we use a ternary here, please extract to separate variable declaration though...

packages/ember-glimmer/lib/utils/process-args.js
```diff
@@ -1,35 +1,45 @@
...
+    seen = Object.create(null);
+  }
 
   for (let i = 0; i < keys.length; i++) {
     let name = keys[i];
+
+    assert(
+      `Component attribute \`${name}\` is declared multiple times for \`${componentName.split(':').slice(-1)}\` component.`,
+      seen[name] ? false : seen[name] = true
+    );
```
Serabe: Is there any reason this check cannot be done in one pass in L20?

Something like:

```js
if (DEBUG) {
  let repeatedKeys = Array.from(keys).filter((key, idx, coll) => coll.indexOf(key) !== idx);
  assert(
    `Component attributes ${repeatedKeys.join(', ')} are declared multiple times for \`${componentName.split(':').slice(-1)}\` component.`,
    repeatedKeys.length === 0
  )
}
```

bekzod: might not work `Array.from([1,2,3,4,4,4,4]).filter((key, idx, coll) => coll.indexOf(key) !== idx)`
will give `[4, 4, 4]`
```

  if (DEBUG) {
    let repeatedKeys = keys.reduce(function(sum, key) {
      sum[key] = 0 || sum[key];
      sum[key]++;
      return sum;
    }, {});

    repeatedKeys = Object.keys(repeatedKeys).filter((key)=> repeatedKeys[key] > 1);

    assert(
      `Component attribute \`${repeatedKeys.join(', ')}\` is declared multiple times for \`${componentName.split(':').slice(-1)}\` component.`,
      repeatedKeys.length === 0
    );
  }
```
is this better ?

alvincrespo: I like how this reads, much clearer to understand.

locks:
@bekzod you have conflicts 👀

packages/ember-glimmer/lib/utils/process-args.ts
```diff
@@ -1,36 +1,45 @@
...
+
+    assert(
+      `Component attribute \`${name}\` is declared multiple times for \`${componentName.split(':').slice(-1)}\` component.`,
+      seen[name] ? false : seen[name] = true
+    );
+
     let ref = namedArgs.get(name);
     let value = attrs[name];
 
-    if (typeof value === 'function' && value[ACTION]) {
```
rwjblue: Why is this branch removed?  Specifically, before these changes when `value === 'function' && value[ACTION]` was set we set the `attrs[name] = value`. After these changes the `attrs[name]` is going to be a `MutableCell`. This branch was intentionally supporting folks doing `this.attrs.someAction()`, and I _believe_ that is now broken after this change.  

We should confirm that tests exists for `this.attrs.foo()` and add them if they are not present...

rwjblue (changes requested):

packages/ember-glimmer/lib/utils/process-args.ts
bekzod: on [line above](https://github.com/emberjs/ember.js/pull/15653/files#diff-069c99d4ffd5ddba5e8f92d6f8fd040dL22)  `let value = attrs[name]; ` I just removed those lines because it is assigning to attrs what is already there :P `attrs[name] = value; `

packages/ember-glimmer/lib/utils/process-args.ts
```diff
@@ -1,36 +1,45 @@
...
+export function processComponentArgs(namedArgs: CapturedNamedArguments, componentName: string) {
   let keys = namedArgs.names;
   let attrs = namedArgs.value();
   let props = Object.create(null);
   let args = Object.create(null);
 
-  props[ARGS] = args;
+  let seen;
+  if (DEBUG) {
+    seen = Object.create(null);
```
bekzod: I can change this to `Set`

bekzod:
I am closing this, I think it is better use `ember-template-lint`

PR #15654, opened 2017-09-11
convert to es6 `OrderedSet` `Map` `MapWithDefault`

bekzod:


packages/ember-metal/lib/map.js
```diff
@@ -197,24 +183,23 @@ OrderedSet.prototype = {
...
 
   /**
     @method copy
     @return {Ember.OrderedSet}
     @private
   */
   copy() {
-    let Constructor = this.constructor;
-    let set = new Constructor();
+    let set = new OrderedSet();
```
rwjblue: This change breaks inheritance, any child classes will now start returning `OrderSet` instead of `ThingThatExtendsOrderedSet`.

packages/ember-metal/lib/map.js
```diff
@@ -414,64 +380,61 @@ Map.prototype = {
...
-    defaultValue: this.defaultValue
-  }));
-};
+  /**
+    @method copy
+    @return {Ember.MapWithDefault}
+    @private
+  */
+  copy() {
+    return copyMap(this, new MapWithDefault({
```
rwjblue: Same as above, we need the `this.constructor` stuff

packages/ember-metal/tests/map_test.js
```diff
@@ -468,13 +468,6 @@ QUnit.test('Map.prototype.constructor', function() {
   equal(map.constructor, Map);
 });
 
-QUnit.test('Map() without `new`', function() {
```
rwjblue: Why do we need to remove this test?  Seems like it should still pass in debug builds, no?

packages/ember-metal/tests/map_test.js
```diff
@@ -519,13 +512,6 @@ QUnit.module('OrderedSet', {
   }
 });
 
-QUnit.test('OrderedSet() without `new`', function() {
```
rwjblue: Same as above, not sure why this would have to be removed in debug builds.  I believe we only strip the class call check helper method in production builds...

packages/ember-metal/lib/map.js
bekzod: thanks good catch 👍 will fix it

packages/ember-metal/lib/map.js
bekzod: thanks good catch 👍 will fix it

rwjblue:
To fix the production build CI run you will need to update the various tests ([like this one](https://github.com/emberjs/ember.js/blob/b27344abd38b8bd4b0f0df5380152401cc0a828a/packages/ember-metal/tests/map_test.js#L221-L246)) to use `expectAssertion` instead of `assert.throws` / `QUnit.throws`.  This is due to the PR changing from `throw new Error(...)` in a few places to `assert(....)` (the asserts are stripped in prod builds, so the test assertions for `assert.throws` are not satisified when running prod).

packages/ember-metal/tests/map_test.js
bekzod: wouldn't it be just checking babel :P I think other es6 classes don't have such test, am I wrong ?

rwjblue: Ya, but in general the other classes that we have in this repo are not attempting to polyfill the spec...

locks (approved):
looks good, but I'd want a second opinion from @rwjblue

rwjblue (approved):

PR #15655, opened 2017-09-11
avoid boolean coercion when checking for `object`

bekzod:


packages/ember-metal/lib/set_properties.js
```diff
@@ -23,7 +23,7 @@ import { set } from './property_set';
   @public
 */
 export default function setProperties(obj, properties) {
-  if (!properties || typeof properties !== 'object') { return properties; }
+  if (properties === null || typeof properties !== 'object') { return properties; }
```
rwjblue: 👍 on the changes here, but it made me notice that we don't support:

```js
function derp() { }

Ember.setProperties(derp, { foo: 'bar', baz: 'qux' });
```

I vaguely believe that we do support setting on functions with `Ember.set` so we should probably fix this to allow functions also (basically it should allow whatever `Ember.set` allows).

rwjblue (approved):

PR #15656, opened 2017-09-11
cleanup `buildInjections`

bekzod:
to avoid touching `arguments` list and avoid creating duplicate argument variables

### DIFF
--- a/packages/container/lib/container.js
+++ b/packages/container/lib/container.js
@@ -306,25 +306,16 @@ function instantiateFactory(container, fullName, options) {
   throw new Error('Could not create factory');
 }
 
-function buildInjections() /* container, ...injections */{
+function buildInjections(container, injections) {
   let hash = {};
   let isDynamic = false;
 
-  if (arguments.length > 1) {
-    let container = arguments[0];
-    let injections = [];
-    let injection;
-
-    for (let i = 1; i < arguments.length; i++) {
-      if (arguments[i]) {
-        injections = injections.concat(arguments[i]);
-      }
-    }
-
+  if (injections.length > 0) {
     if (DEBUG) {
       container.registry.validateInjections(injections);
     }
 
+    let injection;
     for (let i = 0; i < injections.length; i++) {
       injection = injections[i];
       hash[injection.property] = lookup(container, injection.fullName);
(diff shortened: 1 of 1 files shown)


packages/container/lib/container.js
```diff
@@ -314,24 +314,15 @@ function areInjectionsNotDynamic(injections) {
   return injections._dynamic !== true;
 }
 
-function buildInjections() /* container, ...injections */{
+function buildInjections(container, injections) {
```
locks: Shouldn't it be `...injections` here?

bekzod: the point of PR was to avoid touching arguments list :) and passing array instead, `...injections` will compile to again argument touching code I think :P

Serabe: It will not if there is a parameter first.

Serabe: Example: http://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&code_lz=GYVwdgxgLglg9mABACzgGwIYAoB0eZgBWAptPGAM4CUiA3gFCKIQIXrE5pwDmu-RpWKyr0AvvXqhIQpKkxYMAGkR4cBEmWF1GzVu048-agZspUA3GKA&debug=false&circleciRepo=&evaluate=false&lineWrap=true&presets=es2015%2Creact%2Cstage-2&prettier=false&targets=&version=6.26.0

bekzod: in our case it will :P, because there is for loop, http://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&code_lz=GYVwdgxgLglg9mABACzgGwIYAoB0eZgBWAptPGAM4CUiA3gFCKIQIXrE5pwDmu-RpWKyr0AvvXqhIQpKkxYMAGkR4cBEmWF1GzVu048-agZspUA3GIlTTKdNiUr-GmdW1NgcAE5YAbhi9EGEQAXkQABnMggB5EAEZImABqJKoGJgBIFkp9Ll51QXIKAG0YAF0LDJ1xUSA&debug=false&circleciRepo=&evaluate=false&lineWrap=true&presets=es2015%2Creact%2Cstage-2&targets=&version=6.26.0

bekzod: it is much better to pass just array :P

bekzod:
@locks isn't this good to merge :P ?

locks:
I'd prefer if @rwjblue or something else confirmed 😅

rwjblue (approved):

rwjblue:
Looks good to me, thanks @bekzod!

PR #15657, opened 2017-09-11
define `_dynamic` as non-enum in `markInjectionsAsDynamic`

bekzod:
so `_dynamic` is not merged to class instance properties

### DIFF
--- a/packages/container/lib/container.js
+++ b/packages/container/lib/container.js
@@ -307,7 +307,7 @@ function instantiateFactory(container, fullName, options) {
 }
 
 function markInjectionsAsDynamic(injections) {
-  injections._dynamic = true;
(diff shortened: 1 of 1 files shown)


rwjblue:
I totally agree that `_dynamic` should not end up as an instance's own property in this path, but I had always assumed this would mean extracting some state holding class that tracked both the injections themselves and the various flags/status (like `_dynamic` here). 

I'm unsure how often this particular code-path is ran during a "normal" Ember app.  I suspect its not terribly often.

@krisselden / @stefanpenner - Thoughts?

krisselden:
I don't even know what this code path is but would should not leak it for sure

rwjblue:
Closing in favor of the approach in #15713.

PR #15660, opened 2017-09-12
small cleanup in `utils/iterable`

bekzod:


packages/ember-glimmer/lib/utils/iterable.js
```diff
@@ -129,16 +129,11 @@ class ObjectKeysIterator extends ArrayIterator {
   constructor(keys, values, keyFor) {
     super(values, keyFor);
     this.keys = keys;
-    this.length = keys.length;
```
rwjblue: 👍 this is set in the super class already

packages/ember-glimmer/lib/utils/iterable.js
```diff
@@ -129,16 +129,11 @@ class ObjectKeysIterator extends ArrayIterator {
...
   }
 
   getMemo(position) {
     return this.keys[position];
   }
-
-  getValue(position) {
```
rwjblue: 👍 no need to override, the code is the same as `ArrayIterator`

rwjblue:
Thank you @bekzod!

PR #15665, opened 2017-09-13
Fix error & loading (sub-)states with async handlers.

rwjblue:
When a given `HandlerInfo` has an `undefined` `.handler` property, it means that either the handler threw an error during its `init` or that we are traversing into an async resolved handler (e.g. engines).  When handlers are loaded sync, any errors during `init` are immediately surfaced however when they are loaded in an async manner these errors were previously being obfuscated by logic deep in the Ember routing internals that itself fails and throws other errors.  The refactoring in this commit, avoids the masking errors that were being thrown.

### DIFF
--- a/packages/ember-routing/lib/system/router.js
+++ b/packages/ember-routing/lib/system/router.js
@@ -94,7 +94,7 @@ const EmberRouter = EmberObject.extend(Evented, {
 
   _initRouterJs() {
     let routerMicrolib = this._routerMicrolib = new Router();
-    routerMicrolib.triggerEvent = triggerEvent;
+    routerMicrolib.triggerEvent = triggerEvent.bind(this);
 
     routerMicrolib._triggerWillChangeContext = K;
     routerMicrolib._triggerWillLeave = K;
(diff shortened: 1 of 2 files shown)


packages/ember-routing/lib/system/router.js
```diff
@@ -1109,20 +1109,21 @@ function forEachRouteAbove(originRoute, handlerInfos, callback) {
...
-    originRoute.router._scheduleLoadingEvent(transition, originRoute);
+  willResolveModel(handlerInfos, transition, originRoute) {
+    this._scheduleLoadingEvent(transition, originRoute);
   },
 
   // Attempt to find an appropriate error route or substate to enter.
-  error(error, transition, originRoute) {
-    let handlerInfos = transition.state.handlerInfos;
-    let router = originRoute.router;
+  error(handlerInfos, error, transition) {
```
rwjblue: FYI - The `handlerInfos` passed in here (ultimately from `router.js` microlib's `trigger` utility function) are _only_ the resolved ones, this means that `forEachRouteAbove` no longer has to iterate all of the `transition.state.handlerInfo`'s and figure out which was the culprit.

The same is true for `loading` just below...

packages/ember-routing/lib/system/router.js
```diff
@@ -1285,7 +1287,7 @@ export function triggerEvent(handlerInfos, ignoreFailure, args) {
 
   let defaultHandler = defaultActionHandlers[name]
   if (defaultHandler) {
-    defaultHandler.apply(null, args);
+    defaultHandler.apply(this, [handlerInfos, ...args]);
```
rwjblue: This makes the `defaultActionHandlers` above have the router as their `this` context (as opposed to `null`), and allows us to remove the reliance on `originRoute.router` to access the router for the implementations of `error` and `loading`.

packages/ember-routing/lib/system/router.js
```diff
@@ -1109,20 +1109,21 @@ function forEachRouteAbove(originRoute, handlerInfos, callback) {
...
-  willResolveModel(transition, originRoute) {
-    originRoute.router._scheduleLoadingEvent(transition, originRoute);
+  willResolveModel(handlerInfos, transition, originRoute) {
+    this._scheduleLoadingEvent(transition, originRoute);
   },
 
   // Attempt to find an appropriate error route or substate to enter.
-  error(error, transition, originRoute) {
-    let handlerInfos = transition.state.handlerInfos;
-    let router = originRoute.router;
```
rwjblue: Calling `originRoute.router` triggers the `cannot access 'router' of undefined` error I mentioned in one of the commit messages when a handler is loaded via async (or threw an error during its `init`).

krisselden (approved):

PR #15668, opened 2017-09-14
cleanup `get` helper

bekzod:


packages/ember-glimmer/lib/helpers/get.js
```diff
@@ -77,7 +81,7 @@ class GetHelperReference extends CachedReference {
     this.pathReference = pathReference;
 
     this.lastPath = null;
-    this.innerReference = null;
+    this.innerReference = NULL_REFERENCE;
```
bekzod: using reference to avoid this https://github.com/emberjs/ember.js/pull/15668/files#diff-72ba77337e5ed54641529bb08b228338L109

rwjblue (approved):

PR #15672, opened 2017-09-19
Update router_js to 2.0.0.

rwjblue:
The diff of the changes is pretty noisey (due to lots of general cleanup work that was done), but for the most part the changes are:

* Drop support for non ES5 browsers (use `Array.isArray`,
  `Object.create`, etc directly)
* Refactor the dist output to emit `module` entry point.
* Update to use classes throughout the repo.
* Ensure `Route.prototype.serialize` is called bound to the route
  instance.

### DIFF
--- a/broccoli/packages.js
+++ b/broccoli/packages.js
@@ -16,24 +16,10 @@ const { stripIndent } = require('common-tags');
 const toES5 = require('./to-es5');
 
 module.exports.routerES = function _routerES() {
-  return new Rollup(findLib('router_js', 'lib'), {
+  return new Rollup(findLib('router_js'), {
     rollup: {
       external: ['route-recognizer', 'rsvp'],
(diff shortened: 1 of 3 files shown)


broccoli/packages.js
```diff
@@ -16,24 +16,10 @@ const { stripIndent } = require('common-tags');
 const toES5 = require('./to-es5');
 
 module.exports.routerES = function _routerES() {
-  return new Rollup(findLib('router_js', 'lib'), {
+  return new Rollup(findLib('router_js'), {
     rollup: {
       external: ['route-recognizer', 'rsvp'],
-      entry: 'router.js',
-      plugins: [{
-        transform(code, id) {
```
rwjblue: These overrides are not needed any longer, the upstream repo has corrected its invalid internals...

PR #15673, opened 2017-09-20
[WIP] Add generator for class-based helper

balinterdi:


balinterdi:
@kellyselden I started working on it, following the fixture format you suggested.

Before starting this, I tried it in a separate add-on and it worked there.

However, the first test that verifies if the generated file has the right content fails because the files doesn't seem to be created at all:

AssertionError: expected "app/helpers/foo/bar-baz.js" to exist:

```
  1) Acceptance: ember generate and destroy class-based helper class-based helper foo/bar-baz:
     AssertionError: expected "app/helpers/foo/bar-baz.js" to exist

parent path "app/helpers/foo" does not exist.
parent path "app/helpers" exists and contains:
- .gitkeep
```

I copied the example from the helper generator (then modified it to suit my needs) and looked into ember-cli-blueprint-test-helpers to see if I'm doing anything obviously wrong, but couldn't find anything.

It'd be great if you could take a quick look to see if anything jumps out as wrong.

Thank you.

kellyselden:
@balinterdi I'm on vacation for the next two weeks, so I may be slow to get to this.

Turbo87:
closing due to inactivity

balinterdi:
@Turbo87 I didn't indeed worked more on this but it has always felt an omission to not have a generator for class-based helpers. Do you think it'd be useful in ember-cli?

Turbo87:
@balinterdi yeah, it's probably useful, and this repository is the right place for it. the PR was outdated though and no longer matching the other blueprints in this repo. it might also make sense to make this an option on the existing `helper` blueprint, instead of creating a new blueprint for it 🤔

PR #15676, opened 2017-09-20
[BUGFIX beta] Fix range input values

### LINKED ISSUE #15675: `{{input type="range" min=0 max=10 value=someVal}}` fails to correctly set initial value
The problem is that `<input type="range">` actually cares about the order in which attributes are set when using the DOM APIs. See https://github.com/whatwg/html/issues/2427.

The solution, after discussing with @rwjblue is probably just to re-order this list here:
https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/components/text_field.js#L73-L81

A temporary workaround is to use the `<input>` tag, though that involves some machinations for two-way bindings to work correctly:
```js
<input type="range" min="0" max="10"
    value={{someVal}} oninput={{action (mut someVal) value="target.value"}}>
```

Serabe:
If an input a type range is set after value but before min and max, it
produces different results than done in markup.

Closes #15675

### DIFF
--- a/packages/ember-glimmer/lib/components/text_field.js
+++ b/packages/ember-glimmer/lib/components/text_field.js
@@ -70,6 +70,7 @@ export default Component.extend(TextSupport, {
     'inputmode',
     'lang',
     'list',
+    'type', // needs to be before min and max. See #15675
     'max',
     'min',
     'multiple',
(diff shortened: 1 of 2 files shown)


rwjblue (approved):
Awesome, this looks great to me.

rwjblue:
Thank you @Serabe!

PR #15689, opened 2017-10-01
[BUGFIX lts] Mark error as handled before transition for error routes and substates

kratiahuja:
We currently were not marking the error as "handled" before transitioning to the error route (even though we were handling the error by transitioning to the error route/substate). Due to this, the error was being bubbled up and thrown. In the browser this is fine, but in fastboot this causes the `visit` promise to be rejected and the error template to not be serialized.

Super thank you and shout out to @rwjblue for helping figure out the fix!

cc: @rwjblue

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "pretest": "ember build",
     "test": "node bin/run-tests.js",
     "test:blueprints": "node node-tests/nodetest-runner.js",
+    "test:node": "node bin/run-node-tests.js",
     "test:sauce": "node bin/run-sauce-tests.js",
     "test:testem": "testem -f testem.dist.json",
     "link:glimmer": "node bin/yarn-link-glimmer.js"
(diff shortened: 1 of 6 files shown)


rwjblue:
Great work here! Can you prefix the commit with `[BUGFIX lts]`?

kratiahuja:
Sure will do.

kratiahuja:
Fixed more routing test that were wrapping around the error originally being thrown. Hopefully travis god is 🍏 now.

kratiahuja:
Sauce test failures seem unrelated. This should be good to review.

rwjblue:
Ya, that sauce labs job has been giving us loads of grief lately. I definitely do not think this is related to your changes here.

rwjblue (approved):

rwjblue:
Thank you @kratiahuja!

PR #15690, opened 2017-10-02
avoid boolean coercion in `property_set`

bekzod:
this also makes things consistent with docs https://www.emberjs.com/api/ember/2.15/namespaces/Ember/methods/trySet?anchor=trySet

### DIFF
--- a/packages/ember-metal/lib/property_set.ts
+++ b/packages/ember-metal/lib/property_set.ts
@@ -171,22 +171,19 @@ if (DEBUG) {
   };
 }
 
-function setPath(root: object, path: string, value: any, tolerant?: boolean) {
+function setPath(root: object, path: string, value: any, tolerant?: boolean): any {
   let parts = path.split('.');
-  let keyName = parts.pop();
+  let keyName = parts.pop() as string;
 
   assert('Property set failed: You passed an empty path', keyName!.trim().length > 0);
 
   let newPath = parts.join('.');
-
   let newRoot = getPath(root, newPath);
 
-  if (newRoot) {
-    return set(newRoot, keyName!, value);
+  if (newRoot !== null && newRoot !== undefined) {
+    return set(newRoot, keyName, value);
   } else if (!tolerant) {
-    throw new EmberError(
-      `Property set failed: object in path "${newPath}" could not be found or was destroyed.`
-    );
+    throw new EmberError(`Property set failed: object in path "${newPath}" could not be found.`);
   }
 }
 
--- a/packages/ember-metal/tests/accessors/set_path_test.js
+++ b/packages/ember-metal/tests/accessors/set_path_test.js
@@ -77,8 +77,7 @@ moduleFor(
     // DEPRECATED
     //
     ['@test [obj, bla.bla] gives a proper exception message'](assert) {
-      let exceptionMessage =
-        'Property set failed: object in path "bla" could not be found or was destroyed.';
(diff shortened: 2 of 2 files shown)


packages/ember-metal/lib/property_set.js
```diff
@@ -98,10 +97,9 @@ function setPath(root, path, value, tolerant) {
   assert('Property set failed: You passed an empty path', keyName.trim().length > 0)
 
   let newPath = parts.join('.');
-
   let newRoot = getPath(root, newPath);
 
-  if (newRoot) {
+  if (newRoot !== null && newRoot !== undefined) {
```
rwjblue: Should this check for `typeof newRoot === “function” || (typeof newRoot === ‘object’ && newRoot !== null)`? 

I don’t think we can call set with other types of objects as the first arg can we?

bekzod: those cases are handled by `set` itself

bekzod: also confirms what is in docs :P https://www.emberjs.com/api/ember/2.15/namespaces/Ember/methods/trySet?anchor=trySet

bekzod: maybe `trySet` could be deprecated because it is using `set` anyway with last parameter as `true` ?

bekzod: or it should be renamed to `trySetPath` because it only works with path :), `trySet(undefined, 'path', 'val')` will still blowup in production

bekzod: added `isDestroyed` check too

packages/ember-metal/lib/property_set.js
```diff
@@ -98,10 +97,9 @@ function setPath(root, path, value, tolerant) {
   assert('Property set failed: You passed an empty path', keyName.trim().length > 0)
 
   let newPath = parts.join('.');
-
   let newRoot = getPath(root, newPath);
 
-  if (newRoot) {
+  if (newRoot !== null && newRoot !== undefined && !newRoot.isDestroyed) {
```
rwjblue: We shouldn't be hard coding `.isDestroyed` check here.  We moved the flag into meta, we should use that...

bekzod: this is already handled in https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/lib/property_set.ts#L68

bekzod:
ping this can be merged I think

packages/ember-metal/lib/property_set.ts
```diff
@@ -171,21 +171,20 @@ if (DEBUG) {
...
   let newRoot = getPath(root, newPath);
 
-  if (newRoot) {
-    return set(newRoot, keyName!, value);
+  if (newRoot !== null && newRoot !== undefined) {
+    return set(newRoot, keyName, value);
   } else if (!tolerant) {
     throw new EmberError(
-      `Property set failed: object in path "${newPath}" could not be found or was destroyed.`
+      `Property set failed: object in path "${newPath}" could not be found.`
```
bekzod: updated message here since destroyed assertion is already in `set` https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/lib/property_set.ts#L68

mmun:
I think you need to update the "set with path - deprecated" test

rwjblue (approved):

PR #15691, opened 2017-10-02
reuse meta for `didUnwatch` and `willWatch` in `alias descriptor`

bekzod:


packages/ember-metal/lib/watch_key.js
```diff
@@ -13,35 +13,34 @@ import {
...
     }
   }
 }
 
 
 if (MANDATORY_SETTER) {
-  let hasOwnProperty = (obj, key) => Object.prototype.hasOwnProperty.call(obj, key);
-
-  let propertyIsEnumerable = (obj, key) => Object.prototype.propertyIsEnumerable.call(obj, key);
+  const hasOwnProperty = (obj, key) => Object.prototype.hasOwnProperty.call(obj, key);
```
rwjblue: changing these to `const` is breaking CI...

`const` is only allowed in module scope

PR #15692, opened 2017-10-02
cleanup `_getPath`

bekzod:


packages/ember-metal/lib/property_get.js
```diff
@@ -74,24 +68,20 @@ export function _getPath(root, path) {
   let parts = path.split('.');
 
   for (let i = 0; i < parts.length; i++) {
-    if (!isGettable(obj)) {
+    if (obj === undefined || obj === null || obj.isDestroyed) {
```
rwjblue: I don't think we want to hard code `.isDestroyed` check here. In theory it would be better to `let m = peekMeta(obj); m.isDestroyed()`.

packages/ember-metal/lib/property_get.js
```diff
@@ -5,12 +5,6 @@
 import { assert } from 'ember-debug';
 import { isPath } from './path_cache';
 
-const ALLOWABLE_TYPES = {
```
rwjblue: Why is this removed?

bekzod: why to have it :P, even if other types passed `get` will return `undefined`

packages/ember-metal/lib/property_get.js
bekzod: it is how done before https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/lib/property_get.js#L83 <- just moved this check to top
I think checking `isDestroyed` with `meta` worth another PR since hardcoded `isDestroyed` check is done in more places

packages/ember-metal/lib/property_get.js
rwjblue: Generally speaking, its best to review the git history to understand a given bit of code's purpose.

* https://github.com/emberjs/ember.js/issues/13444
* https://github.com/emberjs/ember.js/pull/13449
* https://github.com/emberjs/ember.js/pull/13461

Specifically answering your question though is https://github.com/emberjs/ember.js/pull/13449#r62256688.

bekzod: added test for those cases if something missing will add it, I think all cases handled correctly

packages/ember-metal/lib/property_get.js
bekzod: I will do follow up PR to use `meta.isDestroyed()` does it sound good ? or should I do it with this PR

rwjblue: Ya, sounds good to do separately

bekzod:
ping, any objections to this ?

packages/ember-metal/lib/property_get.ts
```diff
@@ -178,24 +172,16 @@ export function _getPath<T extends object>(root: T, path: string): any {
   let parts = path.split('.');
 
   for (let i = 0; i < parts.length; i++) {
-    if (!isGettable(obj)) {
+    if (obj === undefined || obj === null || (obj as MaybeHasIsDestroyed).isDestroyed) {
```
bekzod: maybe move `(obj as MaybeHasIsDestroyed).isDestroyed ` to `get` itself

PR #15698, opened 2017-10-04
use `peekMeta` where appropriate

bekzod:
since `CoreObject` constructor already inits `meta` https://github.com/emberjs/ember.js/blob/master/packages/ember-runtime/lib/system/core_object.js#L68

### DIFF
--- a/packages/ember-runtime/lib/system/core_object.js
+++ b/packages/ember-runtime/lib/system/core_object.js
@@ -17,6 +17,7 @@ import {
 import {
   get,
   meta,
+  peekMeta,
   finishChains,
   sendEvent,
   detectBinding,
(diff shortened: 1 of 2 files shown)


packages/ember-runtime/lib/system/core_object.js
```diff
@@ -430,7 +431,7 @@ CoreObject.PrototypeMixin = Mixin.create({
   */
   isDestroying: descriptor({
     get() {
-      return meta(this).isSourceDestroying();
+      return peekMeta(this).isSourceDestroying();
```
rwjblue: While I totally agree that the meta is pre-initialized it does seem odd to use `peekMeta` without handling the case where there is no meta found. Is there a benefit to using peekMeta here?

bekzod: since it is guaranteed to have meta why not to use `peekMeta`,  benefit is just less work :P done

rwjblue (approved):

PR #15699, opened 2017-10-04
cleanup assertion

bekzod:
since assertion message is calculated lazily since ember 2.14 https://github.com/emberjs/ember.js/issues/15421#issuecomment-318076057


rwjblue:
Thank you!

PR #15701, opened 2017-10-04
remove `__defineNonEnumerable`

bekzod:
looks odd to have `__defineNonEnumerable` when you can use `Object.defineProperty `

### DIFF
--- a/packages/ember-glimmer/lib/component.ts
+++ b/packages/ember-glimmer/lib/component.ts
@@ -597,13 +597,6 @@ const Component = CoreView.extend(
       this._super();
     },
 
-    __defineNonEnumerable(property: {
-      name: string;
-      descriptor: { value: any }
-    }) {
-      this[property.name] = property.descriptor.value;
-    },
-
     [PROPERTY_DID_CHANGE](key: string) {
       if (this[IS_DISPATCHING_ATTRS]) { return; }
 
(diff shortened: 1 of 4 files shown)


bekzod:
made bold changes :P at least will know how it is used

rwjblue (commented):
This exists specifically because `Object.defineProperty` is slower than `obj[key] = value`.  On many objects (e.g. `Ember.Component` and `EachProxy`) we do not actually care if the property being added is enumerable or not, so we override this `__defineNonEnumerable` with `obj[key] = val;`.

This one is probably a no-go...

bekzod:
thanks for making it clear closing it

bekzod:
@rwjblue even if `Object.defineProperty` is slow,  is `__defineNonEnumerable` constantly used method to affect performance if `Object.defineProperty` used instead, also `CoreObject` already uses `Object.defineProperty` inside `__defineNonEnumerable` which means most of objects already use `Object.defineProperty`

rwjblue:
Yes, totally agreed. It’s very much only an optimization for a small number of things (the largest is components). 

I’m happy to re-evaluate this technique with some benchmarks to see if it still makes sense though...

bekzod:
will benchmark this :)

krisselden:
I think this is worth retesting

bekzod:
benchmark: https://cl.ly/3k3W0t0u2v1E/results.pdf

mmun:
Benchmark looks good.

PR #15703, opened 2017-10-04
rename `destroy` to `deleteMeta`

bekzod:
IMHO `deleteMeta` looks more explicit/clear than `destroy`


packages/ember-metal/lib/watching.js
```diff
@@ -54,15 +53,3 @@ export function unwatch(obj, _keyPath, m) {
...
-  Tears down the meta on an object so that it can be garbage collected.
-  Multiple calls will have no effect.
-
-  @method destroy
-  @for Ember
-  @param {Object} obj  the object to destroy
-  @return {void}
-  @private
-*/
-export { deleteMeta as destroy };
```
bekzod: exporting `deleteMeta` from `lib/watching.js` looked odd

packages/ember/lib/index.js
```diff
@@ -115,7 +115,7 @@ Ember.unwatchPath = metal.unwatchPath;
 Ember.watch = metal.watch;
 Ember.isWatching = metal.isWatching;
 Ember.unwatch = metal.unwatch;
-Ember.destroy = metal.destroy;
+Ember.destroy = metal.deleteMeta;
```
bekzod: could rename here too not sure

rwjblue: No, we need to keep `Ember.destroy` for now.

packages/ember/tests/reexports_test.js
```diff
@@ -96,7 +96,7 @@ QUnit.module('ember reexports');
   ['watch', 'ember-metal'],
   ['isWatching', 'ember-metal'],
   ['unwatch', 'ember-metal'],
-  ['destroy', 'ember-metal'],
+  ['deleteMeta', 'ember-metal'],
```
rwjblue: I think this needs to be:

```js
['destroy', 'ember-metal', 'deleteMeta'],
```

The first arg is `Ember.${path}`.  The second arg is the module it comes from, and the third arg is the export from that module (if different than the first arg).

bekzod: done

rwjblue (approved):

PR #15704, opened 2017-10-05
simplify `applyConcatenatedProperties`

bekzod:


bekzod:
ping

packages/ember-metal/lib/mixin.js
```diff
@@ -135,19 +135,7 @@ function giveMethodSuper(obj, key, method, values, descs) {
...
-  } else if (isArray(baseValue)) {
-    if (value === null || value === undefined) {
-      ret = baseValue;
-    } else {
-      ret = a_concat.call(baseValue, value);
-    }
-  } else {
-    ret = a_concat.call(makeArray(baseValue), value);
-  }
+  let ret = makeArray(baseValue).concat(makeArray(value));
```
bekzod: `makeArray` already has `null/undefined` checks which was done in `applyConcatenatedProperties` before

stefanpenner:
cc @krisselden

packages/ember-metal/lib/mixin.js
krisselden: there was a performance gain from avoiding the concat when it wasn't needed, this would need to be retested

krisselden:
needs an https://github.com/krisselden/ember-macro-benchmark/

bekzod:
benchmark: https://cl.ly/2e1k0Y3q2X3o/results.pdf

mmun:
Benchmark looks good.

PR #15706, opened 2017-10-05
use `isValidFullName` for `assert`

bekzod:
since `registry.isValidFullName` returns boolean while `registry.validateFullName` might throw, I think it is better to use `registry.isValidFullName` for assertions :)


bekzod:
also removed `registry.validateFullName` since it looks like it exhausted its purpose as assertions used everywhere

locks (approved):

locks:
@bekzod can you handle the conflict?

bekzod:
@locks done

locks (approved):

PR #15711, opened 2017-10-06
make consistent `destroyDestroyables` with `resetMember`

bekzod:
when single `container object` destroyed there is no `isInstantiatable(container, key)` check 
but when all `container object`s destroyed there is, I think it is inconsistent


rwjblue:
Thank you!

PR #15713, opened 2017-10-06
mark dynamic injections without polluting instance properties

bekzod:
supersedes #15657 , I think either could be merged I think this one is a bit straightforward


rwjblue:
I absolutely prefer this approach, looks like some legit failures need to be fixed up first...

bekzod:
fixed now sauce failing only

rwjblue:
Thank you!

PR #15716, opened 2017-10-06
[DOC Canary] Yuidoc annotation changes for rfc 176

toddjordan:
API Doc changes for 2.16 to go along with the following other PRs
ember-api-doc: https://github.com/ember-learn/ember-api-docs/pull/347/files
ember-jsonapi-doc: https://github.com/ember-learn/ember-jsonapi-docs/pull/41/files

For 2.16, the docs are set up to show top level packages and classes.  Functions are shown within their package.
* To show packages, we will use the yuidoc `@module` annotation
* To show standalone functions, they must be labelled `@static` and use the `@for` annotation to indicate their package.
* Classes that aren't instantiable and are used as namespaces for standalone functions should also be labelled as `@static`

The docs processing code will process doc output show non-static classes in the ToC as well as in the page for each package.  Functions will be shown under each package, separated by sub-package.

For long term plans for the UI, see here: https://github.com/ember-learn/ember-api-docs/issues/344

### Example ToC:
image

### Example Package:

image


packages/ember-application/lib/system/resolver.js
```diff
@@ -79,32 +78,25 @@ export const Resolver = EmberObject.extend({
...
-  'view:basic'              //=> Ember.View
+  'route:basic'             //=> Route
   'foo:post'                //=> App.PostFoo
   'model:post'              //=> App.Post
   ```
 
-  @class DefaultResolver
-  @namespace Ember
-  @extends Ember.Object
+  @class GlobalsResolver
```
locks: 🤔 ?

toddjordan: https://github.com/ember-cli/ember-rfc176-data/blob/master/mappings.json

toddjordan: image

locks (approved):

toddjordan:
This PR is currently blocking any doc updates to the source code.

Ideally I'd like to get this into Canary and then merge back to beta and then release, so we can deploy with 2.16.1.  Otherwise the auto-docs deploy for 2.16 will overwrite our rfc 176 changes (unless we disable) it.

PR #15719, opened 2017-10-09
avoid creating cachekey for non-singleton objects

bekzod:


rwjblue:
Nice work, thank you!

PR #15722, opened 2017-10-10
[BUGFIX release] empty path in `get` helper should not throw assertion

bekzod:
just noticed this in my app were `get` helper is throwing assertion for empty path which didn't before


rwjblue:
Thank you!

PR #15726, opened 2017-10-10
[BUGFIX] Passing query params via router service

### LINKED ISSUE #15725: Ember 2.16 router service unable to transition without all query parameters present
I upgraded my app from 2.15 to 2.16 today and I think I've found a bug with the `transitionTo` method in the new routing service.

In 2.15 I was using the `ember-router-service-polyfill` addon that I have since removed. The following code was working fine but now appears to be broken:

```js
// controller

export default Controller.extend({
  queryParams: ['page', 'message'],
  page: 1,
  message: ''
});

// component
import { inject as service } from '@ember/service';

  router: service(),

  actions: {
    jump() {
      let queryParams = { message: 'hello world!' };
      this.get('router').transitionTo('index', { queryParams });
    }
  }
});
```

The error I get in the `jump` action is: "You passed the `false` query parameter during a transition into index, please update to page", which I've narrowed down to be coming from [this line in router.js](https://github.com/emberjs/ember.js/blob/ed381d7003ce87dde032d6edae6c19bcf4d37c79/packages/ember-routing/lib/system/router.js#L968).

It does end up working if I include both `message` and `page` in the query params hash. In short:

```js
// broken
this.get('router').transitionTo('index', { queryParams: { message: '' } });
this.get('router').transitionTo('index', { queryParams: { page: 1 } });
this.get('router').transitionTo('index');

// works
this.get('router').transitionTo('index', { queryParams: { message: '', page: 1 } });
```

I have created this app that shows the broken behavior: https://github.com/pgrippi/ember-216-routing-service-bug

cvx:
Continued from PR #15613. (Closes #15725)

Fixes an error that happens when passing fewer than all query params of a given route via router service (a bug introduced in #15414).

### DIFF
--- a/packages/ember-routing/lib/system/router.js
+++ b/packages/ember-routing/lib/system/router.js
@@ -972,7 +972,7 @@ const EmberRouter = EmberObject.extend(Evented, {
               return true;
             }
 
-            if (_fromRouterService) {
+            if (_fromRouterService && presentProp !== false) {
               return false;
             }
 
--- a/packages/ember/tests/routing/router_service_test/transitionTo_test.js
+++ b/packages/ember/tests/routing/router_service_test/transitionTo_test.js
@@ -256,6 +256,27 @@ if (EMBER_ROUTING_ROUTER_SERVICE) {
         });
     }
 
+    ['@test RouterService#transitionTo with unspecified query params'](assert) {
+      assert.expect(1);
+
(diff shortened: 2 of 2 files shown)


rwjblue:
Merged these changes in https://github.com/emberjs/ember.js/pull/15613

PR #15727, opened 2017-10-10
Add null and undefined type coercion for query params

cvx:
Continued from PR #15613.

Fixes the problem reported in #4570 (and attempted to be fixed in #4571). This fix (originally by @raytiley) is required to make the above use case work. Without it, transitioning to a route that has multiple query params ends up with all of them in the url with the value null, i.e.

```
  Ember.Controller.extend({
    queryParams: ['q', 'other', 'stuff']
  });

  // in some component
  this.get('router').transitionTo('articles', {
    queryParams: { q: 'test' },
  });

  // We end up at: /articles?q=test&other=null&stuff=null
  // Instead of (after the fix): /articles?q=test
```

### DIFF
--- a/packages/ember-routing/lib/system/route.js
+++ b/packages/ember-routing/lib/system/route.js
@@ -848,7 +848,10 @@ let Route = EmberObject.extend(ActionHandler, Evented, {
         } else {
           if (presentKey) {
             svalue = params[presentKey];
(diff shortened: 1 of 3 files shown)


rwjblue:
Merged these changes in https://github.com/emberjs/ember.js/pull/15613.

PR #15730, opened 2017-10-11
[DOC release] Update Evented example code for RFC 176

fullsailor:
The example code still mixed in `Ember.Evented`, but the import was updated to just `Evented`


locks:
Thanks!

Issue #15732, opened 2017-10-12
computed.sort is not updated if the value of the itemsKey is null


pieter-v:
If the first time you 'get' the sorted property while the property itself is null, then the sorted property is never recalculated if the property receives a value afterwards (see attached file for a reproduction scenario).

works on 2.15.3, but fails on 2.16.0

The reason why this fails in our app, is that property is a sub-property of model (model.users and model.sortedUsers) and that rendering starts between the beforeModel and model hook. So, the value controller.model is still null.
[sorting.zip](https://github.com/emberjs/ember.js/files/1378708/sorting.zip)

rwjblue:
I believe that this was fixed by https://github.com/emberjs/ember.js/pull/15733.

pieter-v:
Thanks

PR #15733, opened 2017-10-12
[BUGFIX] Fix computed sort regression when array prop initially null

ilucin:
There was a regression introduced ([commit](https://github.com/emberjs/ember.js/commit/ea1c8db2a31afedd2ea76215835f384cdbb0c088#diff-be5c035bb6df695f80dbf5d940188c68)) in a sort computed macro.

If an array property (that is being sorted) is initially null or other non-array value - when computed property is first evaluated observers wouldn't be set and computed property would never work again. Provided test pretty much says it all.

### DIFF
--- a/packages/ember-runtime/lib/computed/reduce_computed_macros.js
+++ b/packages/ember-runtime/lib/computed/reduce_computed_macros.js
@@ -740,10 +740,6 @@ function propertySort(itemsKey, sortPropertiesKey) {
       activeObservers.forEach(args => removeObserver(...args));
     }
 
-    let itemsKeyIsAtThis = (itemsKey === '@this');
-    let items = itemsKeyIsAtThis ? this : get(this, itemsKey);
-    if (!isArray(items)) { return emberA(); }
-
     function sortPropertyDidChange() {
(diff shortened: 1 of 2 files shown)


bekzod:
was about to submit PR for this, good job @ilucin 👍

bekzod:
fixes #15732

rwjblue:
Thank you!

ilucin:
@rwjblue No probs. Shouldn't we release 2.16.1 with this fix included?

Techn1x:
We're already up to 2.16.2, so I guess it would be 2.16.3? I'm waiting for this PR fix before upgrading, since the broken sort does break some functionality in my webapp :/

Techn1x:
According to this;
https://github.com/emberjs/ember.js/commit/8bfa2013de2c15108b123f41b6a407a56f64108e

It looks like the fix is in 2.17-beta.2. Seems like it's going to be a few weeks before the fix is in stable channel, if it's targeted at 2.17...

ilucin:
I've opened another PR #15746 (over release branch) for 2.16.1

Issue #15728, opened 2017-10-11
currentPath does not respect resetNamespace dsl option


janmisek:
Even when `resetNamespace` is used `currentPath` returns path including segments from parent routes above resetNamespace.

Same behaviour observed for:
- `Ember.getOwner(this).lookup('controller:application').get('currentPath')`
- `Ember.getOwner(this).lookup('service:-routing').get('currentPath')`
- `currentPath()` testing helper

janmisek:
```
    this.route('root', {path: '/root'}, function () {
      this.route('reset', {path: '/reset'});
    });
```

link to works properly, following works
```
{{link-to 'reset' 'href title'}}
```
But 
- `Ember.getOwner(this).lookup('controller:application').get('currentPath')`
- `Ember.getOwner(this).lookup('service:-routing').get('currentPath')`
- `currentPath()` testing helper

Returns `root.reset` instead of `reset`

Serabe:
Hello, @janmisek,

this is the expected behaviour of `currentPath`. You are looking for `currentRouteName`.

See [this twiddle showing the difference between currentPath and currentRouteName](https://ember-twiddle.com/f20d82ba73968ed3b81becea9001cf63?fileTreeShown=false&numColumns=0&route=%2Froot%2Freset).

Thanks!

PR #15738, opened 2017-10-13
[CLEANUP] Fix test for resetNamespace

Serabe:
While triaging #15728, locks and I saw that the `resetNamespace` part of
the helpers test was flawed, given that it is set in a top level route.

This commit adds a test for a resetNamespace in a nested route.


rwjblue:
Thanks @Serabe!

PR #15739, opened 2017-10-13
[DEPRECATION canary] Deprecate Ember.String and prototype extension

Serabe:
This PR deprecates the use of String prototype extension and
Ember.String namespace.

Methods coming from ember-runtime
---

A new `StringUtils` object is added, to be used internally by Ember
instead of using the same object being exposed as `Ember.String`. This
simplifies testing and the changes required in the rest of the code.
Also, dropping the deprecated code will be much simpler.

Methods coming from ember-glimmer
---

Deprecated versions are being added and reexported as the same global
the non-deprecated versions were. Besides, the non-deprecated version is
being exported under the new global `Ember.Template`.

The `Ember.Template` is just a placeholder until the final global is chosen.

### DIFF
--- a/packages/ember-glimmer/lib/utils/string.ts
+++ b/packages/ember-glimmer/lib/utils/string.ts
@@ -2,6 +2,8 @@
 @module @ember/string
 */
 
+import { deprecate } from 'ember-debug';
+
 export class SafeString {
   public string: string;
 
@@ -74,6 +76,7 @@ export function escapeExpression(string: any): string {
   @static
   @return {Handlebars.SafeString} A string that will not be HTML escaped by Handlebars.
   @public
+  @deprecated
 */
 export function htmlSafe(str: string) {
   if (str === null || str === undefined) {
@@ -82,6 +85,19 @@ export function htmlSafe(str: string) {
     str = '' + str;
   }
   return new SafeString(str);
+}
+
+export function deprecatedHTMLSafe(str: string) {
+  deprecate(
+    'Ember.String namespace is deprecated. Please, import `htmlSafe` from `@ember/template`.',
+    false,
+    {
+      id: 'ember-glimmer.ember-string-html-safe',
+      until: '3.5.0',
+      url: '',
+    }
+  );
+  return htmlSafe(str);
 }
 
 /**
@@ -102,7 +118,21 @@ export function htmlSafe(str: string) {
   @static
   @return {Boolean} `true` if the string was decorated with `htmlSafe`, `false` otherwise.
   @public
+  @deprecated
 */
 export function isHTMLSafe(str: any | null | undefined): str is SafeString {
   return str !== null && typeof str === 'object' && typeof str.toHTML === 'function';
+}
+
+export function deprecatedIsHTMLSafe(str: any | null | undefined): str is SafeString {
+  deprecate(
+    'Ember.String namespace is deprecated. Please, import `isHTMLSafe` from `@ember/template`.',
+    false,
+    {
+      id: 'ember-glimmer.ember-string-is-html-safe',
+      until: '3.5.0',
+      url: '',
+    }
+  );
+  return isHTMLSafe(str);
 }
--- a/packages/ember-runtime/lib/ext/string.js
+++ b/packages/ember-runtime/lib/ext/string.js
@@ -3,6 +3,7 @@
 */
 
 import { ENV } from 'ember-environment';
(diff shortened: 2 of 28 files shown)


Serabe:
@locks any update on this?

Serabe:
restarting job. Seemed a problem with yarn.

Serabe:
Passing.

packages/ember-glimmer/lib/utils/string.ts
```diff
@@ -98,6 +99,19 @@ export function htmlSafe(str: string) {
...
 }
 
+export function deprecatedHTMLSafe(str) {
+  deprecate(
+    'Ember.String namespace is deprecated. Please, import `htmlSafe` from `ember/template`.',
+    false,
+    {
+      id: 'ember-glimmer.ember-string-html-safe',
+      until: '3.5.0',
+      url: ''
```
rwjblue: Do we know the URL's for these deprecations? I vaguely recall that there was already an open website PR for them...

packages/ember-runtime/lib/ext/string.js
```diff
@@ -14,9 +14,27 @@ import {
   capitalize,
   classify
 } from '../system/string';
+import { deprecate } from 'ember-debug';
+import { assign } from 'ember-utils';
 
 const StringPrototype = String.prototype;
 
+function deprecateEmberStringExtension(fn, opts = {}) {
+  let { name } = fn;
```
rwjblue: I don't think we can rely on `Function.name` (AFAIK it isn't supported on all platforms, which would make this fail in production).

Checkout https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name for a decent compat table (it isn't supported on IE at all)...

packages/ember-runtime/lib/system/string.js
```diff
@@ -115,6 +115,22 @@ function capitalize(str) {
   return CAPITALIZE_CACHE.get(str);
 }
 
+function deprecateEmberStringUtil(fn, opts = {}) {
```
rwjblue: This seems roughly like the same function in the other file, can we share between the two and only have one implementation?

packages/ember-runtime/lib/system/string.js
```diff
@@ -115,6 +115,22 @@ function capitalize(str) {
...
+  let { name } = fn;
```
rwjblue: Same concern as above RE: `Function.name`

packages/ember/lib/index.js
```diff
@@ -516,6 +527,19 @@ Object.defineProperty(Ember, 'TEMPLATES', {
...
 
+/**
+ Global for safe usage of methods coming from module 'ember/template'.
+
+ @property Template
+ @for Ember
+ @type Object
+ @private
+ */
+Ember.Template = {
```
rwjblue: Lets underscore this, I'd hate for folks to start using it ....

rwjblue (commented):
Looking really good!  Only a few small questions/adjustments that I commented on inline...

packages/ember-glimmer/lib/utils/string.ts
Serabe: Adding the URL from https://github.com/emberjs/website/pull/2995/files

Serabe:
Addressed your suggestions. Thanks for the review!

rwjblue (approved):

Serabe:
Rebased and squashed.

Serabe:
Having lots of timeouts while doing `yarn install` in CI

Serabe:
Yay! I got tests passing after restarting some jobs several times.

mmun:
@rwjblue should we merge this?

locks:
@mmun negatory, sir. we need to publish the deprecation guides and release the addon before.
sorry for dwelling on this 😞

Serabe:
Rebased!

Serabe:
I will rebase this at some point, quite likely, next Friday. Would that be ok, @locks ?

Serabe:
I redid the whole PR because this rebase was going to be harder than redoing everything.

Serabe:
Green again.

Serabe:
Closing this. Sorry, I cannot keep rebasing this.

locks:
Sorry 😢I'll wait for https://github.com/emberjs/ember.js/pull/16530.

PR #15744, opened 2017-10-17
Update glimmer version to 0.29.8

### LINKED ISSUE #15621: [2.15.0] {{link-to}} inside a partial stopped working
After upgrading to 2.15.0 some of our tests started failing because a `{{#link-to}}` inside of a partial stopped working.

[This is the `{{link-to}}` in question](https://github.com/TryGhost/Ghost-Admin/blob/master/app/templates/-user-list-item.hbs#L2). When rendering there are no immediate errors but the generated href is just `#` and the `data-test-user-id` attribute is not output when testing. Clicking the link outputs this warning:

`This link-to is in an inactive loading state because at least one of its parameters presently has a null/undefined value, or the provided route name is invalid.`

Everything else inside of the partial still renders fine, it's just the link-to that fails.

[Changing the partial to a component](https://github.com/TryGhost/Ghost-Admin/commit/19d940291d5dc89d9f5318e52c8c03868846810e) fixed the problem.

dsst95:
Fixes #15621

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,10 +40,10 @@
     "link:glimmer": "node bin/yarn-link-glimmer.js"
   },
   "dependencies": {
-    "@glimmer/compiler": "^0.25.3",
-    "@glimmer/node": "^0.25.3",
-    "@glimmer/reference": "^0.25.3",
-    "@glimmer/runtime": "^0.25.3",
+    "@glimmer/compiler": "^0.29.8",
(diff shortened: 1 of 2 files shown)


rwjblue:
The upgrade to 0.29 is a large effort (and unlikely to happen before Ember 3.0.0), @chadhietala backported the fix to the 0.25 branch. We need to release that as a patch version and update to _that_ version instead.

PR #15751, opened 2017-10-19
Consider defaulting RouterDSL parent name to 'application'

kyleshay:
Default `name` to _"application"_ for the RouterDSL constructor in dsl.js
https://github.com/emberjs/ember.js/blob/6d9a9735eeb04182589ff9bd98569288667cb8d8/packages/ember-routing/lib/system/dsl.js#L12

The reasoning behind this is that it is possible for a user to pass in _null_ to the constructor. Because of that, the generated route names from the `getFullName` function might be, for example: _"null.index"_ instead of _"index"_
https://github.com/emberjs/ember.js/blob/6d9a9735eeb04182589ff9bd98569288667cb8d8/packages/ember-routing/lib/system/dsl.js#L178

If the default name was _"application"_, then `getFullName` will return simply _"index"_ without the _null_ prefix (this happens because of the in the `canNest` function.)

This issue was introduced in v2.15.x - previously, supplying _null_ would return _"index"_

For more information please review https://github.com/nathanhammond/ember-route-alias/pull/22#discussion_r145839398

### DIFF
--- a/packages/ember-routing/lib/system/dsl.js
+++ b/packages/ember-routing/lib/system/dsl.js
@@ -9,7 +9,7 @@ import { assert, deprecate } from 'ember-debug';
 let uuid = 0;
 
 class DSL {
-  constructor(name, options) {
+  constructor(name = 'application', options) {
     this.parent = name;
     this.enableLoadingSubstates = options && options.enableLoadingSubstates;
     this.matches = [];


kyleshay:
Seems one of the tests in the the sauce suite [failed](https://travis-ci.org/emberjs/ember.js/jobs/290224232#L633) for a (seemingly) unrelated reason. Closing & reopening PR to re-trigger

kyleshay:
Another solution (if this change wasn't intended) could be to just add that check back into the `canNest` function if the name is null.

rwjblue:
Defaulting like this doesn't actually fix the explicit passing of `null` (since default arguments only coerce `undefined` to the default).

Thanks for working on this @kyleshay, but I think I'm going to close this for now.  I believe that the underlying issue you were hitting has been resolved in the addon (and released). Happy to reopen if I'm missing something...

kyleshay:
Sounds good, thanks for the update!

PR #15755, opened 2017-10-22
remove fs-extra dependency

kellyselden:


node-tests/blueprints/route-test.js
```diff
@@ -395,7 +391,6 @@ describe('Acceptance: ember generate and destroy route', function() {
     var args = ['route', 'application', '--pod'];
 
     return emberNew()
-      .then(() => remove(path.resolve('app/templates/application.hbs')))
```
rwjblue: @kellyselden - Do we just not need to remove this file?

package.json
```diff
@@ -55,7 +55,6 @@
     "ember-cli-valid-component-name": "^1.0.0",
     "ember-cli-version-checker": "^1.3.1",
     "ember-router-generator": "^1.2.3",
-    "fs-extra": "^4.0.1",
```
rwjblue: Even if we kept it, I think it should become a devDep anyways

node-tests/blueprints/route-test.js
kellyselden: Yea. I tried converting to `fs.unlink` first, but the file is never there. Must be legacy, and can come back in the future if needed.

package.json
kellyselden: Agreed

rwjblue:
Thanks @kellyselden !

PR #15760, opened 2017-10-25
avoid extra boolean conversion

bekzod:


packages/ember-routing/lib/system/router.js
```diff
@@ -118,10 +118,8 @@ const EmberRouter = EmberObject.extend(Evented, {
   },
 
   _buildDSL() {
-    let moduleBasedResolver = this._hasModuleBasedResolver();
-    let options = {
-      enableLoadingSubstates: !!moduleBasedResolver
```
bekzod: `moduleBasedResolver` is already boolean

packages/ember-routing/lib/system/router.js
```diff
@@ -174,13 +172,12 @@ const EmberRouter = EmberObject.extend(Evented, {
   _hasModuleBasedResolver() {
     let owner = getOwner(this);
 
-    if (!owner) { return false; }
-
-    let resolver = owner.application && owner.application.__registry__ && owner.application.__registry__.resolver;
-
-    if (!resolver) { return false; }
+    let resolver = owner &&
```
rwjblue: Heh, I wonder if we should just use `get` to avoid this annoying chaining... 🤔

PR #15761, opened 2017-10-25
use strippable `warn` in `link-to`

bekzod:


rwjblue (approved):

rwjblue:
Looks like we have some tests around that assertion that are now failing. Most likely they are older style tests that are monkey patching Ember.warn (and needs to be updated to use setDebugFunction).

bekzod:
updated tests

PR #15762, opened 2017-10-25
make sure `_isActive` returns boolean in `link-to`

bekzod:


packages/ember-glimmer/lib/components/link-to.js
```diff
@@ -548,19 +548,18 @@ const LinkComponent = EmberComponent.extend({
   _isActive(routerState) {
     if (get(this, 'loading')) { return false; }
 
-    let routing = get(this, '_routing');
-    let models = get(this, 'models');
-    let resolvedQueryParams = get(this, 'resolvedQueryParams');
     let currentWhen = get(this, 'current-when');
 
-    if (typeof currentWhen === 'boolean') {
-      return currentWhen ? get(this, 'activeClass') : false;
```
bekzod: `get(this, 'activeClass') ` is not boolean

rwjblue: Hmm, this subtly changes the semantics.  Previously you could specify `activeClass: false` and still never have `_isActive` return true...

bekzod: isn't that covered by tests ? that is still handled here https://github.com/bekzod/ember.js/blob/fcdd954c1cbf10df15c1afbdb4a03e99f54499bc/packages/ember-glimmer/lib/components/link-to.js#L590

locks (approved):

PR #15763, opened 2017-10-25
remove `bind` fallback

bekzod:
since `function.bind` supported from IE9


rwjblue:
Seems good to me, thank you!

PR #15766, opened 2017-10-25
[DOC release] Remove Extra Space in Computed Macros

mwagz:
Removes an extra space in computed macros documentation to fix misalignment. [Documentation](https://github.com/emberjs/ember.js/blob/v2.15.0/packages/ember-runtime/lib/computed/computed_macros.js#L550) was vertically aligned in the code, but misaligned when viewing [the API](https://emberjs.com/api/ember/2.15/namespaces/Ember.computed/methods/notEmpty?anchor=oneWay).

screenshot 2017-10-25 15 09 17

### DIFF
--- a/packages/ember-runtime/lib/computed/computed_macros.js
+++ b/packages/ember-runtime/lib/computed/computed_macros.js
@@ -549,7 +549,7 @@ export const or = generateComputedWithPredicate('or', value => !value);
 
   let teddy = User.create({
     firstName: 'Teddy',
-    lastName:  'Zeenny'
+    lastName: 'Zeenny'
   });
 
   teddy.get('nickName');              // 'Teddy'


rwjblue:
That font in the API docs is _wild_ O_o

rwjblue:
@locks do you happen to know why are we not using a fixed width font?

mwagz:
I'm looking into this failing CI. 

> 2 SL_MS_Edge - bailout

Not totally sure why a single space would cause it to fail. Maybe the CI is checking if the code is vertically aligned? Because with this removal, the code is no longer vertically aligned. I wonder if Edge renders the font differently. I know IE used to be notorious for that.

MelSumner:
I recently dealt with IE and fonts on a different project- we had to remove cache-control and pragma:no-cache from the headers, and add cache exclusions for OTF, WOFF and TTF file types.

rwjblue:
Thanks @mwagz!

PR #15769, opened 2017-10-25
[DOC canary] Update Ember.A() docs to include optional param

alexmiddeleer:
Array-like (with a hyphen) seems to be the more popular wording in the ember.js repo, so I updated it here. Also updated the example to show usage with the optional param.


Serabe (approved):

rwjblue:
Thank you!

Turbo87:
@rwjblue seems like this could also be `[DOC release]`, right?

rwjblue:
Yep!

PR #15773, opened 2017-10-26
Start removing implicit any errors

smfoote:
Thusfar:

component-managers/abstract.ts (2 errors remain)
component-managers/curly.ts (some errors remain)
component-managers/mount.ts
utils/curly-component-state-bucket.ts

### DIFF
--- a/packages/container/lib/index.d.ts
+++ b/packages/container/lib/index.d.ts
@@ -1,0 +1,17 @@
+interface Container {
+  registry: any;
+  owner: any | null;
+  cache: any | null;
+  factoryManagerCache: any | null;
+  isDestroyed: boolean;
+  validationCache: any | null;
+  lookup(fullName: string, options: any): any;
+  destroy(): void;
+  reset(fullName: string): void;
+  ownerInjection(): any;
+  _resolveCacheKey(name: string, options: any): any;
+  factoryFor(fullName: string, options: any): any;
+}
+export const privatize: any;
+export const Registry: any;
+export const Container: (registry: any, options?: any) => Container;
--- a/packages/ember-glimmer/lib/component-managers/curly.ts
+++ b/packages/ember-glimmer/lib/component-managers/curly.ts
@@ -253,12 +253,12 @@ export default class CurlyComponentManager extends AbstractManager<ComponentStat
   layoutFor(definition: CurlyComponentDefinition, bucket: ComponentStateBucket, env: Environment): CompiledDynamicProgram {
     let template = definition.template;
     if (!template) {
-      template = this.templateFor(bucket, env);
+      template = this.templateFor(bucket.component, env);
     }
     return env.getCompiledBlock(CurlyComponentLayoutCompiler, template);
   }
 
-  templateFor(component: ComponentStateBucket, env: Environment): OwnedTemplate {
+  templateFor(component: Component, env: Environment): OwnedTemplate {
     let Template = get(component, 'layout');
     let owner = component[OWNER];
     if (Template) {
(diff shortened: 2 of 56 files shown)


packages/container/lib/index.d.ts
```diff
@@ -0,0 +1,17 @@
+interface Container {
```
krisselden: I stubbed out a more limited interface for this in ember-glimmer we should unify

packages/ember-glimmer/lib/component-managers/curly.ts
```diff
@@ -74,26 +91,26 @@ function applyAttributeBindings(element, attributeBindings, component, operation
...
+function ariaRole(vm: any) {
   return vm.getSelf().get('ariaRole');
 }
 
 class CurlyComponentLayoutCompiler {
   static id: string;
   public template: any;
 
-  constructor(template) {
+  constructor(template: any) {
```
krisselden: I did make a type for template

packages/ember-glimmer/lib/component-managers/curly.ts
```diff
@@ -74,26 +91,26 @@ function applyAttributeBindings(element, attributeBindings, component, operation
...
-function tagName(vm) {
+function tagName(vm: any) {
   // tslint:disable-next-line:no-shadowed-variable
   let { tagName } = vm.dynamicScope().view;
 
   return PrimitiveReference.create(tagName === '' ? null : tagName || 'div');
 }
 
-function ariaRole(vm) {
+function ariaRole(vm: any) {
```
krisselden: there is a type exported for this, I think it is called PublicVM

packages/ember-glimmer/lib/component-managers/curly.ts
```diff
@@ -107,22 +124,22 @@ export class PositionalArgumentReference {
...
     this._references = references;
   }
 
   value() {
-    return this._references.map((reference) => reference.value());
+    return this._references.map((reference: any) => reference.value());
   }
 
-  get(key) {
+  get(key: any) {
```
krisselden: string

packages/ember-glimmer/lib/component-managers/curly.ts
```diff
@@ -231,16 +248,15 @@ export default class CurlyComponentManager extends AbstractManager<ComponentStat
...
-      let { component } = bucket;
-      template = this.templateFor(component, env);
+      template = this.templateFor(bucket, env);
     }
     return env.getCompiledBlock(CurlyComponentLayoutCompiler, template);
   }
 
-  templateFor(component, env) {
+  templateFor(component: ComponentStateBucket, env: Environment): any {
     let Template = get(component, 'layout');
```
krisselden: we have a type for what templates return look at the template .d.ts I updated

packages/ember-routing/lib/index.d.ts
```diff
@@ -0,0 +1,16 @@
+export const Location: any;
+export const NoneLocation: any;
+export const HashLocation: any;
+export const HistoryLocation: any;
+export const AutoLocation: any;
+
```
krisselden: lets at least type the signature of the functions

packages/ember-glimmer/lib/component-managers/curly.ts
smfoote: Works for the `ariaRole` function, but not `tagName`: `Property 'view' does not exist on type 'DynamicScope'`

smfoote:
Despite the green check, I don't think this is ready yet. There are several TypeScript errors still.

rwjblue:
@smfoote - Sounds good, let us know when you think it is ready for another round of review and possible merging.

smfoote:
Since the checks are passing, we might merge this one as is then fix the remaining TypeScript errors in a future PR. I'll leave that up to @krisselden

PR #15779, opened 2017-10-28
cleanup `active` computed property

bekzod:


locks (approved):

packages/ember-glimmer/lib/components/link-to.ts
```diff
@@ -592,10 +592,7 @@ const LinkComponent = EmberComponent.extend({
     @property active
     @private
   */
-  active: computed('attrs.params', '_active', function computeLinkToComponentActiveClass(this: any) {
-    let currentState = get(this, '_routing.currentState');
-    if (!currentState) { return false; }
```
rwjblue: This existed to avoid computing `active` if routing was not started (e.g. from integration or unit tests where there is no routing).  Can you confirm that `_active` does not throw in that case?

bekzod: it won't I think there is `_routing.currentState` check in https://github.com/bekzod/ember.js/blob/29266e0977e638f00cb87850ec0856a44fb5bbdf/packages/ember-glimmer/lib/components/link-to.ts#L599 so it is just removing duplication

rwjblue:
Awesome, thank you!

packages/ember-glimmer/lib/components/link-to.ts
thoov: @rwjblue @bekzod Why was `'attrs.params'` removed? CC: @jasonmit (https://github.com/emberjs/ember.js/pull/16060)

PR #15628, opened 2017-09-04
[Refactor]: Align component manager's methods with RFC

twokul:
Before this change, when defining a custom component manager, you'd have to define `layoutFor` (or extend from `AbstractComponentManager`)  for components to render properly. After this change, you'd have to define only 3 methods (as per [RFC](https://github.com/emberjs/rfcs/blob/custom-components/text/0000-custom-components.md)): `create`, `getSelf` (`getContext` in RFC) and `update`. This commit does not change existing behaviour.

Changes:
+ introduce a new "wrapper" class for custom component managers. `layoutFor` mimics the behaviour of curly components and `create`, `getSelf` and `update` are left for developers to define
+ tests

Relevant links:
+ [Custom Components RFC](https://github.com/emberjs/rfcs/blob/custom-components/text/0000-custom-components.md)

PR #15783, opened 2017-10-29
[WIP]: Component Capabilities

twokul:
Implementation of `capabilities` feature described in Custom Components
RFC.

This change depends on three things:

+ https://github.com/emberjs/ember.js/pull/15628
+ feature flag `GLIMMER_CUSTOM_COMPONENT_MANAGER`
+ version of `glimmer-vm` that has `capabilities` API implemented
(0.28.0+)

Relevant links:

+ https://github.com/emberjs/rfcs/blob/custom-components/text/0000-custom-components.md

### DIFF
--- a/packages/ember-glimmer/lib/component-managers/capabilities.ts
+++ b/packages/ember-glimmer/lib/component-managers/capabilities.ts
@@ -1,0 +1,40 @@
+import { assert } from 'ember-debug';
+import { assign } from 'ember-utils';
+
+// because we're not yet compatible with latest Glimmer
+// we lack typings in Ember.js.
+// steal capabilities interface from http://bit.ly/2gMdhv5 for now
+export interface ComponentCapabilities {
+  dynamicLayout: boolean;
+  dynamicTag: boolean;
+  prepareArgs: boolean;
+  createArgs: boolean;
+  attributeHook: boolean;
+  elementHook: boolean;
+}
+
+// default capabilities values stolen from http://bit.ly/2y9gCvc
+export const DEFAULT_CAPABILITIES: ComponentCapabilities = {
+  dynamicLayout: true,
+  dynamicTag: true,
+  prepareArgs: true,
+  createArgs: true,
+  attributeHook: true,
+  elementHook: true
+};
+
+const SPECIFIER_HASH = {
+  '2.16': DEFAULT_CAPABILITIES
+};
+
+function getCapabilities(specifier: string): ComponentCapabilities {
+  assert(`No capabilities mask is defined for "${specifier}" specifier`, SPECIFIER_HASH[specifier] !== undefined);
+
+  return DEFAULT_CAPABILITIES;
+}
+
+export default function capabilities(specifier: string, features: Object): ComponentCapabilities {
+  assert('You must pass a specifier, e.g. `2.16`', specifier !== undefined);
+
+  return assign({}, getCapabilities(specifier), features);
+}
--- a/packages/ember-glimmer/lib/component-managers/custom.ts
+++ b/packages/ember-glimmer/lib/component-managers/custom.ts
@@ -1,14 +1,38 @@
+import { get } from 'ember-metal';
+import { OWNER } from 'ember-utils';
+import { DEBUG } from 'ember-env-flags';
 import { assert } from 'ember-debug';
 import AbstractManager from './abstract';
 import {
   CurlyComponentLayoutCompiler
 } from './curly';
-import { ComponentManager } from '@glimmer/runtime';
-import ComponentStateBucket from '../utils/curly-component-state-bucket';
+import {
+  Arguments,
+  CompiledDynamicProgram,
+  ComponentDefinition,
+  ComponentManager,
+  DynamicScope,
+  ElementOperations
+} from '@glimmer/runtime';
+import ComponentStateBucket, { Component } from '../utils/curly-component-state-bucket';
+import { ComponentCapabilities } from './capabilities';
+import { Destroyable, Opaque } from '@glimmer/util';
+import { Option } from '@glimmer/interfaces';
+import Environment from '../environment';
+import { VersionedPathReference } from '@glimmer/reference';
+import { privatize as P } from 'container';
+import { OwnedTemplate } from '../template';
+
+const DEFAULT_LAYOUT = P`template:components/-default`;
+
+interface CustomComponentDefinition extends ComponentDefinition<ComponentStateBucket> {
+  template: OwnedTemplate;
+  args: Arguments | undefined;
+}
 
 /**
   Wrapper class for custom component managers as per
-  Custom Components RFC.
+  Custom Components RFC, http://bit.ly/2AlScAb.
 
   When defining a custom component manager, the following
   methods are required:
@@ -17,15 +41,17 @@ import ComponentStateBucket from '../utils/curly-component-state-bucket';
   - `getSelf`
   - `update`
 
-  `layoutFor` method is not required to be implemented and
-  mirrors the behaviour of curly components.
+  `layoutFor`/`templateFor` mirror the behaviour of curly components.
+  `didCreate`, `didUpdate`, `getDestructor` are optional methods which
+  could be overriden but by default are no-ops.
+
   @private
   @class CustomManager
 */
 export default class CustomManager extends AbstractManager<ComponentStateBucket> {
   private _manager: ComponentManager<ComponentStateBucket>;
 
-  constructor(manager) {
+  constructor(manager: ComponentManager<ComponentStateBucket>) {
     super();
 
     this._manager = manager;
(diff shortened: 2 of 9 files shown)


packages/ember-glimmer/lib/component-managers/capabilities.ts
```diff
@@ -0,0 +1,30 @@
+import { assert } from 'ember-debug';
```
twokul: is this a correct location for this file?

packages/ember-glimmer/lib/component-managers/capabilities.ts
```diff
@@ -0,0 +1,30 @@
...
+import { assign } from 'ember-utils';
+
+// because we're not yet compatible with latest Glimmer
+// we lack typings in Ember.js.
+// steal capabilities interface from http://bit.ly/2gMdhv5 for now
+export interface ComponentCapabilities {
```
twokul: can we omit typings here even thought we use TS now?

packages/ember-glimmer/lib/component-managers/capabilities.ts
```diff
@@ -0,0 +1,30 @@
...
+  dynamicLayout: boolean;
+  dynamicTag: boolean;
+  prepareArgs: boolean;
+  createArgs: boolean;
+  attributeHook: boolean;
+  elementHook: boolean;
+}
+
+// default capabilities values stolen from http://bit.ly/2zfZQyz
+export const DEFAULT_CAPABILITIES: ComponentCapabilities = {
```
twokul: @wycats @chadhietala could you guys confirm that these values are correct default values?

packages/ember-glimmer/lib/component-managers/capabilities.ts
```diff
@@ -0,0 +1,30 @@
...
+  dynamicLayout: true,
+  dynamicTag: true,
+  prepareArgs: true,
+  createArgs: true,
+  attributeHook: false,
+  elementHook: false
+};
+
+export default function capabilities(specifier: string, features: Object): Object {
```
twokul: what should `specifier` be here? just ember versions like `2.*` and so on?

packages/ember-glimmer/lib/component-managers/custom.ts
```diff
@@ -0,0 +1,101 @@
...
+    assert('You must implement `getSelf` method.', this._manager.getSelf !== undefined);
+    assert('You must implement `update` method.', this._manager.update !== undefined);
+  }
+
+  create(environment, definition, args, dynamicScope, callerSelfRef, hasBlock) {
+    return this._manager.create(environment, definition, args, dynamicScope, callerSelfRef, hasBlock);
+  }
+
+  // state is ComponentDefinition
+  getCapabilities(state): ComponentCapabilities {
```
twokul: will there ever be a case when we want to allow developers to override `getCapabilities`?

packages/ember-glimmer/lib/component-managers/capabilities.ts
rwjblue: Yes, lets use Ember versions for this.

packages/ember-glimmer/lib/component-managers/capabilities.ts
```diff
@@ -0,0 +1,30 @@
...
+  assert('you need to pass in a specifier', specifier !== undefined);
+
+  return assign(DEFAULT_CAPABILITIES, features);
```
rwjblue: Hmm, this seems wrong.  The `DEFAULT_CAPABILITIES` object will be mutated for each invocation.  I think you probably want:

```js
return assign({}, DEFAULT_CAPABILITIES, features);
```

packages/ember-glimmer/lib/component-managers/custom.ts
```diff
@@ -0,0 +1,101 @@
...
+import {
+  CurlyComponentLayoutCompiler
+} from './curly';
+import { ComponentManager } from '@glimmer/runtime';
+import ComponentStateBucket from '../utils/curly-component-state-bucket';
+import { ComponentCapabilities } from './capabilities';
+
+/**
+  Wrapper class for custom component managers as per
+  Custom Components RFC.
```
rwjblue: Can you link to the RFC here?

packages/ember-glimmer/lib/component-managers/custom.ts
```diff
@@ -0,0 +1,101 @@
...
+  When defining a custom component manager, the following
+  methods are required:
+
+  - `create`
+  - `getSelf`
+  - `update`
+
+  `layoutFor` method is not required to be implemented and
+  mirrors the behaviour of curly components.
+  @private
```
rwjblue: Can you add a newline above this `@private`?

packages/ember-glimmer/lib/component-managers/custom.ts
```diff
@@ -0,0 +1,101 @@
...
+  Custom Components RFC.
+
+  When defining a custom component manager, the following
+  methods are required:
+
+  - `create`
+  - `getSelf`
+  - `update`
+
+  `layoutFor` method is not required to be implemented and
```
rwjblue: This comment infers that it isn't required to be implemented, but it doesn't explain if it _can_ be implemented?

packages/ember-glimmer/lib/component-managers/custom.ts
rwjblue: I don't think so, at least not in the near future...

packages/ember-glimmer/lib/component-managers/custom.ts
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  // state is ComponentDefinition
+  getCapabilities(state): ComponentCapabilities {
+    // should likely return default?
+    return state.capabilities;
+  }
+
+  getSelf({ component }) {
+    return this._manager.getSelf(component);
```
rwjblue: The manager hooks should always receive the full value that was returned from its own `create` method (commonly called the "state bucket"). This should be updated to pass through the first argument unmodified.

packages/ember-glimmer/lib/component-managers/custom.ts
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  update(bucket, dynamicScope) {
+    return this._manager.update(bucket, dynamicScope);
+  }
+
+  layoutFor(definition, bucket, env) {
+    if (this._manager.layoutFor) {
+      return this._manager.layoutFor(definition, bucket, env);
```
rwjblue: I'm not 100% of the args here.  I generally think that all of the manager hooks should receive the state bucket as their first argument, but the thing I am _really_ unsure about is if we should be exposing the `env` or `definition`'s here.  It _somewhat_ infers that they would become public API which I don't think is what we want.

packages/ember-glimmer/lib/component-managers/custom.ts
```diff
@@ -0,0 +1,101 @@
...
+    if (!template) {
+      let { component } = bucket;
+      template = this.templateFor(component, env);
+    }
+    return env.getCompiledBlock(CurlyComponentLayoutCompiler, template);
+  }
+
+  didCreate(component) {
+    if (this._manager.didCreate !== undefined) {
+      this._manager.didCreate(component);
```
rwjblue: `didCreate` isn't mentioned in the inline docs above, should it be added?

packages/ember-glimmer/lib/component-managers/custom.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    this._manager = manager;
+
+    assert('You must implement `create` method.', this._manager.create !== undefined);
+    assert('You must implement `getSelf` method.', this._manager.getSelf !== undefined);
+    assert('You must implement `update` method.', this._manager.update !== undefined);
+  }
+
+  create(environment, definition, args, dynamicScope, callerSelfRef, hasBlock) {
+    return this._manager.create(environment, definition, args, dynamicScope, callerSelfRef, hasBlock);
```
rwjblue: We should most likely add some debug stack management here:

```
    if (DEBUG) {
      this._pushToDebugStack(`component:${definition.name}`, environment);
    }
```

It is super important to be done, but I don't think this is a detail we want to force upon all custom managers.  Thoughts?

rwjblue: Doing ^ would also mean implementing `didRenderLayout` with:

```js
    if (DEBUG) {
      this.debugStack.pop();
    }
```

packages/ember-glimmer/lib/component-managers/custom.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  didCreate(component) {
+    if (this._manager.didCreate !== undefined) {
+      this._manager.didCreate(component);
+    } else {
+      super.didCreate(component);
+    }
+  }
+
+  didCreateElement(component, element, operations) {
```
rwjblue: The first arg here is actually not `component`, it is `bucket`.

packages/ember-glimmer/lib/component-managers/custom.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    let template = definition.template;
+    if (!template) {
+      let { component } = bucket;
+      template = this.templateFor(component, env);
+    }
+    return env.getCompiledBlock(CurlyComponentLayoutCompiler, template);
+  }
+
+  didCreate(component) {
```
rwjblue: This should be named `bucket` not `component`

packages/ember-glimmer/lib/component-managers/custom.ts
```diff
@@ -0,0 +1,101 @@
...
+    if (this._manager.didUpdate !== undefined) {
+      this._manager.didUpdate(state);
+    } else {
+      super.didUpdate(state);
+    }
+  }
+
+  getDestructor(state) {
+    if (this._manager.getDestructor !== undefined) {
+      this._manager.getDestructor(state);
```
rwjblue: This method isn't mentioned in the API docs, can you add it?

packages/ember-glimmer/lib/component-managers/custom.ts
```diff
@@ -0,0 +1,101 @@
...
+    if (this._manager.didCreateElement !== undefined) {
+      this._manager.didCreateElement(component, element, operations);
+    } else {
+      super.didCreateElement(component, element, operations);
+    }
+  }
+
+  didUpdate(state) {
+    if (this._manager.didUpdate !== undefined) {
+      this._manager.didUpdate(state);
```
rwjblue: Same RE: API docs

rwjblue (commented):
Awesome job here, this is looking really good!

packages/ember-glimmer/lib/component-managers/capabilities.ts
chadhietala: If these are for Ember Components then I believe you need `attributeHook`. If I remember correctly this is how we do attribute bindings.

rwjblue:
Closing for now, I believe we have to rethink things...

PR #15785, opened 2017-10-29
correct `Libraries` class definition

bekzod:


packages/ember-metal/lib/libraries.js
```diff
@@ -18,14 +18,6 @@ export class Libraries {
     this._coreLibIndex = 0;
   }
 
-  isRegistered(name) {
```
bekzod: removed this since it was being rewritten by https://github.com/emberjs/ember.js/pull/15785/files#diff-5fe4db1b38a417a2db38f66e62b90ce3L26

rwjblue:
Thanks!

PR #15793, opened 2017-10-30
added ts linting rules

bekzod:
just copied linting rules from `glimmer.js`


rwjblue (approved):

rwjblue:
AFAICT these are still not going to be run in CI (and therefore are unenforced). Might be nice to follow up with some linting in the build (using broccoli-tslint).

locks:
You have a conflict @bekzod.

rwjblue:
Thanks!

PR #15798, opened 2017-10-31
remove unused dependencies

bekzod:
as `babel-plugin-debug-macros` used I think there no need for `babel-plugin-feature-flags` and `babel-plugin-filter-imports`

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,8 +66,6 @@
     "aws-sdk": "^2.46.0",
     "babel-plugin-check-es2015-constants": "^6.22.0",
     "babel-plugin-debug-macros": "^0.1.10",
-    "babel-plugin-feature-flags": "^0.3.1",
-    "babel-plugin-filter-imports": "^0.3.1",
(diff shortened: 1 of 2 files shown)


bekzod:
also removed `ember-cli-is-package-missing` which does same thing as `ember-cli-dependency-checker` I think

bekzod:
I think it just needs restart

rwjblue:
I just made some conflicts in another PR :/

package.json
```diff
@@ -47,7 +47,6 @@
     "broccoli-funnel": "^1.2.0",
     "broccoli-merge-trees": "^2.0.0",
     "ember-cli-get-component-path-option": "^1.0.0",
-    "ember-cli-is-package-missing": "^1.0.0",
```
rwjblue: This is definitely being used...

bekzod: oops

rwjblue:
Thank you!

PR #15802, opened 2017-11-01
Remove duplicate dependency

topaxi:


rwjblue:
🙀 can't believe I did that, thank you for fixing!

PR #15808, opened 2017-11-03
Cleanup at end of transaction to prevent memory leaks

### LINKED ISSUE #15803: Double render assert meta leaks memory

krisselden:
Cleanup at end of transaction to prevent memory leaks when objects that are rendered are retained.

Fix #15803

### DIFF
--- a/packages/ember-metal/lib/meta.js
+++ b/packages/ember-metal/lib/meta.js
@@ -7,11 +7,7 @@ import {
 import { protoMethods as listenerMethods } from './meta_listeners';
 import { assert } from 'ember-debug';
 import { DEBUG } from 'ember-env-flags';
-import {
-  EMBER_GLIMMER_DETECT_BACKTRACKING_RERENDER,
-  EMBER_GLIMMER_ALLOW_BACKTRACKING_RERENDER,
-  MANDATORY_SETTER
-} from 'ember/features';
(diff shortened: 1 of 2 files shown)


rwjblue:
@GavinJoyce - Can you review also?

stefanpenner (approved):
LGTM: Reviewed, and spoke with @krisselden about this. :shipit:

rwjblue (commented):
This generally looks good to me, thanks for tackling it!  I do find this coding style slightly confusing though (totally a pre-existing issue, and not something new introduced in this PR).

It would go a very long way if we had a comment that explained the general idea of what is going on. Basically documenting the `setup`, `addKey`, `checkKey`, and `cleanup` functions and how they play into the general transaction system here...

krisselden:
@rwjblue I think I made the code clearer, it doesn't diff as well, but can you take a look to see if you understand it better now?

GavinJoyce:
This looks good to me (and I agree that the code is now much easier to understand)

locks (approved):

rwjblue (approved):

rwjblue:
Thank you @krisselden! This is very much easier for me to understand (and likely better factored too boot).

rwjblue:
I believe that this needs to make its way into a 2.16 release once we have vetted it on 2.17 betas...

PR #15811, opened 2017-11-03
[BUGFIX beta] reuse `extractQueryParamsArgument`

bekzod:


locks:
@bekzod query param specs are complaining about this change, they're all coming back `null`.

bekzod:
this also aligns `router-service / transitionTo` behave like `router / transitionTo`
which was regressed a bit with https://github.com/emberjs/ember.js/commit/7c7f9bc287606f79348f3498029471685edfbc82 
where `router-service / transitionTo` would not correctly work when only `queryParam hash` passed as first argument

bekzod:
also added test for regression

rwjblue:
Thanks @bekzod!

Issue #13362, opened 2016-04-18
Ember.assign when called with Array.prototype.reduce produces a different result


kratiahuja:
I am not sure if I have explained this in a best way. But with Ember 2.5.0, I believe Ember.merge and Ember.assign are supposed to be functionally same (since there is a deprecation to use Ember.assign).

Following is the example in the change in behavior between Ember.merge and Ember.assign:
Using `Ember.merge`:
`[{'a': 'a'}, {'b': 'b'}].reduce(Ember.merge)` produces `{'a': 'a', 'b': 'b'}`

Using `Ember.assign`:
`[{'a': 'a'}, {'b': 'b'}].reduce(Ember.assign)` produces `{'a': 'a', 'b': 'b', '0': {'a': 'a', 'b': 'b', '0': {..}, '1': {...}}, '1': {'a': 'a', 'b': 'b', '0': {..}, '1': {..}}`

It ends up creating a nested object with circular dependencies.

Also to note this works correctly (produces the exact output as `Ember.merge`:
`[{'a': 'a'}, {'b': 'b'}].reduce(function(prev, curr) {
   return Ember.assign(prev, curr);
 })`

I haven't looked into why this is happening yet. I will take a look at it later this week.

Filing an issue to know if this is the expected behavior with Ember.assign?

rwjblue:
Can you provide a demo JSBin or ember-twiddle showing this? We need to see the actual invocations and whatnot to understand what is going on...

kratiahuja:
Sorry, about not including the JSBin. I'll create that today

nickiaconis:
My guess is `assign` is taking the fourth parameter from the [`reduce` callback](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce#Parameters) (i.e. the original array) and treating it like an object with properties 0 and 1, thereby merging it into the target object. Since [`Object.assign`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) accepts multiple sources, and [`Ember.assign` mimics this functionality](https://github.com/emberjs/ember.js/blob/v2.5.0/packages/ember-metal/lib/assign.js#L14), I don't think there's a way around this.

`Ember.merge` did not have this issue because it only accepted two arguments. Would it be worthwhile to continue offering `Ember.merge` as a two-parameter wrapper around `Ember.assign`?

rwjblue:
[...]

PR #15815, opened 2017-11-04
use `assign` internally for merge

bekzod:
until `merge` is deprecated it is good idea to use `assign` which fallbacks to native `Object.assign`


rwjblue (commented):
They actually do not have the same semantics. See https://github.com/emberjs/ember.js/issues/13362 for some background.

We should likely deprecate `Ember.merge` officially (cc @locks).

rwjblue:
Closing for now...

PR #15816, opened 2017-11-04
Make use of JS type coercion for validation

sarathkumarbaskaran:


Serabe:
AFAIK, avoiding coercion in the framework is an optimization, so I find this unlikely to be merged.

sarathkumarbaskaran:
@Serabe 
I think, validating the same key against two values (here) is not the optimized way, if the validation is quite certain.
Instead of writing every time `a === b` and `a === c`, it'd be better to have `a == b`, if at all the required condition passes. Even, it doesn't change the behavior while minify/uglify.
Correct me, if I'm wrong.

rwjblue:
Generally speaking, folks in JS land avoid using == because it’s somewhat trolling in some circumstances (I’m actually surprised our eslint rules are not making this fail). 

TBH, I’m unsure what is “best” here. Perhaps @krisselden can chime in RE how various engines handle it...

mmun:
As a point of style, we tend to use triple equal to make the code easier to understand at the expense of  brevity.

PR #15817, opened 2017-11-05
Fix - send event trace for keypress

sarathkumarbaskaran:
1. Sends `jQuery Event info`  to the action as handled in other events.
2. Refactored sendAction method in `text_support`. It's long time, since we shift from 
``` handlebars
{{input on="key-press" action="doStuff"}}
````
to 
``` handlebars
{{input key-press="doStuff"}}
````


rwjblue:
I’m not sure this is the direction we want to go with this. If you’d like to operate on the native event you should use a closure style action (which already receives the event), but in either case this definitely changes public API and should probably go through an RFC first.

Closing this PR, but I’d be happy to continue discussion either here or in an RFC...

PR #15832, opened 2017-11-08
[BUGFIX release] Query params refreshModel

Boudewijn26:
This reproduces the https://github.com/emberjs/ember.js/issues/15801 in a test case. The problem actually lies in the query param change detection and I'm more than willing to include that. This does demonstrate the most user facing part of this issue.

Any pointers on the best approach to fix this are more than welcome.

While writing this I also noticed that transitioning from the parent route to the child route through the router service, actually triggered a transition abort. I'm convinced this is caused by the same underlying issue, but would be more than willing to include another failing test case to demonstrate that.

### DIFF
--- a/packages/ember/tests/routing/router_service_test/transitionTo_test.js
+++ b/packages/ember/tests/routing/router_service_test/transitionTo_test.js
@@ -302,6 +302,44 @@ if (EMBER_ROUTING_ROUTER_SERVICE) {
         });
     }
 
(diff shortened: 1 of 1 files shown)


kategengler:
Is this still relevant?

Boudewijn26:
Sorry for the late response. Not to me, I've left the employment of the organisation for which this was relevant at the time. @rmachielse would likely be able to provide a better answer

PR #15834, opened 2017-11-08
ember-glimmer/string: Fix `escapeExpression()` types

Turbo87:
`escapeExpression()` accepts anything, converts it into a string and always returns a `string`.

/cc @rwjblue

### DIFF
--- a/packages/ember-glimmer/lib/utils/string.ts
+++ b/packages/ember-glimmer/lib/utils/string.ts
@@ -53,7 +53,7 @@ function escapeChar(chr: keyof typeof escape) {
   return escape[chr];
 }
 
-export function escapeExpression(string: string | SafeString) {
+export function escapeExpression(string: any): string {
   if (typeof string !== 'string') {
     // don't escape SafeStrings, since they're already safe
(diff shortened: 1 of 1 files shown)


rwjblue (approved):
Seems good to change the type here, but I'd also prefer to refactor the function to avoid calling the non-string argument a string.  It confuses me 😝 

Something like:

```ts
export function escapeExpression(input: any): string {
  let string;
  if (typeof input !== 'string') {
    // don't escape SafeStrings, since they're already safe
    if (input && input.toHTML) {
      return input.toHTML();
    } else if (input === null || input === undefined) {
      return '';
    } else if (!input) {
      return input + '';
    }

    // Force a string conversion as this will be done by the append regardless and
    // the regex test will do this transparently behind the scenes, causing issues if
    // an object's to string has escaped characters in it.
    string = '' + input;
  }

  if (!possible.test(string)) { return string; }
  return string.replace(badChars, escapeChar);
}
```

Turbo87:
@rwjblue totally agreed

rwjblue:
Awesome, thanks for volunteering!

PR #15838, opened 2017-11-09
promote eslint rule "semi" to entire project

kellyselden:
Thought it might be nice. Not sure how crazy you want to get with linting.

Code was autofixed.


rwjblue:
Awesome, thank you!

Issue #15843, opened 2017-11-10
[2.16.0] Computed filter regression


ilucin:
There's a regression bug in computed.filter not observing property changes properly -> in 2.16 release. 

Paste this into the console:
```
Obj = Ember.Object.extend({
  filteredItems: Ember.computed.filter('items.@each.{prop}', function(item) {
    return item.get('prop') === false;
  })
});

obj = Obj.create();
obj.set('items', [Ember.Object.create({prop: false})]);
console.log('Expected: 1 Actual: ', obj.get('filteredItems.length'));
obj.get('items')[0].set('prop', true);
console.log('Expected: 0 Actual: ', obj.get('filteredItems.length'));
```

This is the third core regression bug (#15836, #15733) I found in 2.16 and this one is pretty major. What's happening here, are there any more? Shall I PR failing tests in core_object module? @rwjblue

rwjblue:
Yes, a failing test PR would be helpful. I'm not sure of the root cause without digging yet...

ilucin:
Here it is. I've scanned the 2-15.3 -> 2.16.0 changes and I can see that there has been some ember-core updating going on but I can't find the source of these regressions.

Serabe:
Fixed in #15855

PR #15844, opened 2017-11-10
Add failing test for computed.filter @each with {} regression

ilucin:
Adds failing test for #15843


mmun:
This was merged as part of https://github.com/emberjs/ember.js/pull/15855.

PR #15845, opened 2017-11-10
Add failing test for observer computed reads 2.16 regression

ilucin:
Adds failing test for #15836

Issue #15836, opened 2017-11-09
[2.16.0] CoreObject observer+reads regression


ilucin:
Pretty weird regression happened in 2.16.0. I failed tracing down the commit that caused this so maybe you guyz can help. 

Just paste this snippet in the console of your Ember app and see for yourself:
```
 const Obj = Ember.Object.extend({
    name: Ember.computed.reads('something.else'),
    nameDidChange: Ember.observer('name', function() {})
  });

  Obj.create({name: 'obj1'});
  Obj.create({name: 'obj2'});
```

This is what happens: [Devtools sshot](http://take.ms/Kqv3F)

... or to put it verbally: "If you create more than one object with overriden computed.reads property with chained dependency and an active observer that watches that property then the world will collapse".

It's an edge-case but I need exactly this kind of behaviour in my app.

The error **won't** show up if:
- we change computed.reads to non-chained property `name: Ember.computed.reads('something')`
- if we observe the property that computed.reads points to `nameDidChange: Ember.observer('something.else', function() {})`
- if you don't override computed.reads prop in creation `Obj.create({})`
- if you use Ember version < 2.16

bekzod:
this is fixed in #15849

ilucin:
Confirming it's fixed in 2.17

PR #15849, opened 2017-11-11
[BUGFIX release] fix issue with unchaining ChainNodes

bekzod:
fix for https://github.com/emberjs/ember.js/pull/15845 #15836

### DIFF
--- a/packages/ember-metal/lib/chains.js
+++ b/packages/ember-metal/lib/chains.js
@@ -209,8 +209,11 @@ class ChainNode {
   remove(path) {
     let paths = this._paths;
     if (paths === undefined) { return; }
+
     if (paths[path] > 0) {
       paths[path]--;
+    } else {
+      return;
     }
 
     let key = firstKey(path);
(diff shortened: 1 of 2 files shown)


rwjblue:
Can you cherry pick the failing test commits here so that we can confirm they are passing?

rwjblue:
Can you confirm that the regression this fixes is in 2.16? If so, can you prefix the fixing commit with `[BUGFIX release]` so we know to include it in the next patch release?

bekzod:
done

bekzod:
accidentally closed it :P

rwjblue (approved):

rwjblue:
Thank you!

krisselden:
This is poor root cause analysis, remove should never be called unbalanced, we still don't know the cause of the original regression this is just masking the symptom.

krisselden:
we need to find the commit that caused the regression, this fix is just hiding the issue, we must have changed something around chains being overriden during extend or finish chains

krisselden:
@bekzod can you git bisect to find the regressing commit? this fix is no good, it is just hiding the unbalanced watch/unwatch.  This is going to cause other more subtle bugs.

bekzod:
@krisselden roger that

bekzod:
found offending commit https://github.com/emberjs/ember.js/commit/0c72faaa4586da57283691ac1d8f4a3f1f88f081

bekzod:
will checkout why it is happening later today

PR #15850, opened 2017-11-11
Cleanup ChainNode

bekzod:
splitting `path` once and using `array` instead of `string` for `tail` , which should perform better and looks much cleaner

### DIFF
--- a/packages/ember-metal/lib/chains.js
+++ b/packages/ember-metal/lib/chains.js
@@ -4,12 +4,6 @@ import { watchKey, unwatchKey } from './watch_key';
 import { getCachedValueFor } from './computed';
 import { eachProxyFor } from './each_proxy';
 
-const FIRST_KEY = /^([^\.]+)/;
-
-function firstKey(path) {
-  return path.match(FIRST_KEY)[0];
-}
-
 function isObject(obj) {
   return typeof obj === 'object' && obj !== null;
 }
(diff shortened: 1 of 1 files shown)


krisselden:
Looks good, did you measure it?

bekzod:
will post measures soon

bekzod:
benchmark: https://cl.ly/1R033K462y0C/results.pdf looks like makes tiny bit of difference :P

mmun:
@bekzod The PDF only includes the ember version, but not the application code that was tested. What  were you benchmarking exactly?

mmun:
> looks like makes tiny bit of difference :P

You should be careful when you're interpreting the chart. The last page reads:
 
> If the true location shift were equal to 0, there is a %75.58 chance of observing these samples:

This means there is a 75% chance that your code had performance impact at all (neither slower or faster)

bekzod:
> You should be careful when you're interpreting the chart. The last page reads:
> This means there is a 75% chance that your code had performance impact at all (neither slower or faster)
ooh ok, even if it makes no difference it at least does not regress :P and chops few lines of code

mmun:
BTW, I like this cleanup. It inspires me to look around and see where we could be sharing `.split('.')`s and avoiding `.join('.')`s. For example,

https://github.com/emberjs/ember.js/blob/706166a3fa92707c40374a4614bb76aad0764810/packages/ember-metal/lib/property_set.js#L118

bekzod:
nice, I had similiar approach here https://github.com/emberjs/ember.js/pull/15532 need to rebase and benchmark again :)

mmun:
I chatted with bekzod in #ember-dev on Slack and confirmed he was doing all the benchmarking steps correctly. He also shared his benchmarks with me in a zip. Here they all are together:

https://cl.ly/0o001L310624/bench.zip

PR #15854, opened 2017-11-14
updating ember-cli-version-checker

ef4:
Updating this everywhere is a prerequisite for getting ember-cli to work with yarn workspaces.

### DIFF
--- a/blueprints/test-framework-detector.js
+++ b/blueprints/test-framework-detector.js
@@ -14,7 +14,7 @@ module.exports = function(blueprint) {
     if ('ember-cli-qunit' in dependencies) {
       type = 'qunit';
     } else if ('ember-cli-mocha' in dependencies) {
-      var checker = new VersionChecker({ project: this.project });
+      var checker = new VersionChecker({ root: this.project.root });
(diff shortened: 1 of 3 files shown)


locks (approved):
cc @rwjblue please double check

rwjblue (approved):
Should be at least `[BUGFIX beta]`.

blueprints/test-framework-detector.js
```diff
@@ -14,7 +14,7 @@ module.exports = function(blueprint) {
     if ('ember-cli-qunit' in dependencies) {
       type = 'qunit';
     } else if ('ember-cli-mocha' in dependencies) {
-      var checker = new VersionChecker({ project: this.project });
+      var checker = new VersionChecker({ root: this.project.root });
```
rwjblue: generally speaking, we should do `new VersionChecker(this.project)` but this should work just fine

ef4: I swear I tried that and had test failures, but now I went back to prove it and it works fine. 🤷‍♂️

PR #15855, opened 2017-11-14
[BUGFIX release] fix regression with computed `filter/map/sort`

bekzod:
`filter/map/sort` would not expand `items.@each.{prop1,prop2}` therefore would not properly work

### DIFF
--- a/packages/ember-runtime/lib/computed/reduce_computed_macros.js
+++ b/packages/ember-runtime/lib/computed/reduce_computed_macros.js
@@ -17,7 +17,12 @@ import compare from '../compare';
 import { isArray } from '../utils';
 import { A as emberA } from '../system/native_array';
 
-function reduceMacro(dependentKey, callback, initialValue) {
+function reduceMacro(dependentKey, callback, initialValue, name) {
+  assert(
+    `Dependent key passed to \`Ember.computed.${name}\` shouldn't contain brace expanding pattern.`,
+    !/[\[\]\{\}]/g.test(dependentKey)
+  );
+
   let cp = new ComputedProperty(function() {
     let arr = get(this, dependentKey);
     if (arr === null || typeof arr !== 'object') { return initialValue; }
@@ -51,7 +56,11 @@ function arrayMacro(dependentKey, callback) {
   return cp;
 }
 
-function multiArrayMacro(_dependentKeys, callback) {
+function multiArrayMacro(_dependentKeys, callback, name) {
+  assert(
+    `Dependent keys passed to \`Ember.computed.${name}\` shouldn't contain brace expanding pattern.`,
+    _dependentKeys.every((dependentKey)=> !/[\[\]\{\}]/g.test(dependentKey))
+  );
   let dependentKeys = _dependentKeys.map(key => `${key}.[]`);
 
   let cp = new ComputedProperty(function() {
@@ -74,7 +83,7 @@ function multiArrayMacro(_dependentKeys, callback) {
   @public
 */
 export function sum(dependentKey) {
-  return reduceMacro(dependentKey, (sum, item) => sum + item, 0);
+  return reduceMacro(dependentKey, (sum, item) => sum + item, 0, 'sum');
 }
 
 /**
@@ -120,7 +129,7 @@ export function sum(dependentKey) {
   @public
 */
 export function max(dependentKey) {
-  return reduceMacro(dependentKey, (max, item) => Math.max(max, item), -Infinity);
+  return reduceMacro(dependentKey, (max, item) => Math.max(max, item), -Infinity, 'max');
 }
 
 /**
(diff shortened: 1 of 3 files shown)


bekzod:
fixes #15843

packages/ember-runtime/lib/computed/reduce_computed_macros.js
```diff
@@ -45,7 +45,9 @@ function arrayMacro(dependentKey, callback) {
     } else {
       return emberA();
     }
-  }, { dependentKeys: [ dependentKey ], readOnly: true });
+  }, { readOnly: true });
+
+  cp.property(dependentKey);
```
rwjblue: 👍 the fix is good, but I wanted to comment here to explain why this is different.

The `ComputedProperty` constructor does **not** run the provided dependent keys through `expandProperties` (perhaps it should, but that is another issue altogether).  So moving this out from the constructor options into `cp.property` forces it to run `expandProperties`.

rwjblue (changes requested):
This regression was introduced in https://github.com/emberjs/ember.js/pull/15550 which changed both the `reduceMacro`, `multiArrayMacro`,and `arrayMacro` ( the one fixed in this PR) invocations to directly pass `dependentKeys` in options (and therefore avoiding expand properties). 

Can you add a test for the same regression that uses `multiArrayMacro` (`Ember.computed.uniq` for example) and `reduceMacro` (`Ember.computed.sum` for example) using brace expansion, and update those implementations also?

packages/ember-runtime/lib/computed/reduce_computed_macros.js
bekzod: I think it is intentional that constructor argument  does not run through `expandProperties` to avoid extra work (so it expects expanded properties already) because I made `dependent keys ` passable through constructor :P actually before

bekzod:
I think rest of computed properties are not affected by regression, because computed properties that use `reduceMacro`  just expect array property name, I can add assertion for passing expandable properties to `computed` properties that use `reduceMacro` (for `sum`, `max`, `min`) if it is necessary

rwjblue:
> I can add assertion for passing expandable properties to computed properties that use reduceMacro/multiArrayMacro (for sum, max, min) if it is necessary

Seems good to me.  Can you add that assertion in a separate commit (as a `[BUGFIX beta]` instead of release)?

packages/ember-runtime/lib/computed/reduce_computed_macros.js
```diff
@@ -17,7 +17,12 @@ import compare from '../compare';
 import { isArray } from '../utils';
 import { A as emberA } from '../system/native_array';
 
-function reduceMacro(dependentKey, callback, initialValue) {
+function reduceMacro(dependentKey, callback, initialValue, name) {
+  assert(
+    `Dependent key passed to Ember.computed.${name}() shouldn't contain brace expading pattern.`,
```
rwjblue: typo `s/expading/expanding/`

packages/ember-runtime/lib/computed/reduce_computed_macros.js
```diff
@@ -51,7 +56,11 @@ function arrayMacro(dependentKey, callback) {
   return cp;
 }
 
-function multiArrayMacro(_dependentKeys, callback) {
+function multiArrayMacro(_dependentKeys, callback, name) {
+  assert(
+    `Dependents keys passed to Ember.computed.${name}() shouldn't contain brace expading pattern.`,
```
rwjblue: same as above

packages/ember-runtime/lib/computed/reduce_computed_macros.js
```diff
@@ -17,8 +17,12 @@ import compare from '../compare';
 import { isArray } from '../utils';
 import { A as emberA } from '../system/native_array';
 
+function reduceMacro(dependentKey, callback, initialValue, name) {
+  assert(
+    `Dependent key passed to Ember.computed.${name}() shouldn't contain brace expading pattern.`,
```
rwjblue: s/expading/expanding/

packages/ember-runtime/lib/computed/reduce_computed_macros.js
```diff
@@ -45,12 +49,18 @@ function arrayMacro(dependentKey, callback) {
...
+
+  cp.property(dependentKey); // this forces to expand properties GH #15855
 
   return cp;
 }
 
-function multiArrayMacro(_dependentKeys, callback) {
+function multiArrayMacro(_dependentKeys, callback, name) {
+  assert(
+    `Dependent keys passed to Ember.computed.${name}() shouldn't contain brace expading pattern.`,
```
rwjblue: s/exading/expanding/

rwjblue (changes requested):

urbany:
Hi, will this make it to 2.17?

rwjblue:
I believe so (and ultimately backported to 2.16 also)...

PR #15857, opened 2017-11-15
make consistent how `concatenatedProperties` merged

bekzod:
`CoreObject` would concat `instance` and `class` properties inconsistently.
When `class property` is `null or undefined` then `instance property` would be concatenated without `null or undefined` values, while other way around, when `instance property` was `null or undefined` and `class property` had some value it would concat with `null or undefined`

### DIFF
--- a/packages/ember-runtime/lib/system/core_object.js
+++ b/packages/ember-runtime/lib/system/core_object.js
@@ -180,11 +180,7 @@ function makeCtor(base) {
               let baseValue = self[keyName];
 
               if (hasConcatenatedProps && concatenatedProperties.indexOf(keyName) > -1) {
-                if (baseValue) {
-                  value = makeArray(baseValue).concat(value);
-                } else {
-                  value = makeArray(value);
-                }
+                value = makeArray(baseValue).concat(makeArray(value));
               }
 
               if (hasMergedProps && mergedProperties.indexOf(keyName) > -1) {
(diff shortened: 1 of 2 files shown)


bekzod:
this just aligns with how concatenated properties merged  in mixins https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/lib/mixin.ts#L149

packages/ember-runtime/lib/system/core_object.js
```diff
@@ -180,11 +180,7 @@ function makeCtor(base) {
               let baseValue = self[keyName];
 
               if (hasConcatenatedProps && concatenatedProperties.indexOf(keyName) > -1) {
-                if (baseValue) {
-                  value = makeArray(baseValue).concat(value);
-                } else {
-                  value = makeArray(value);
-                }
+                value = makeArray(baseValue).concat(makeArray(value));
```
rwjblue: The reason for the previous implementation is that we specifically do not want to create the extra array for every `Ember.Object.create()`.

Turbo87:
@rwjblue I'm assuming this can be closed?

packages/ember-runtime/lib/system/core_object.js
bekzod: but that fixes inconsistency shown in test

locks:
Closing due to inactivity. Thanks for the work!

PR #15858, opened 2017-11-15
Turn `loc` from a built-in helper into a regular helper

locks:
This is necessary for the RFC 236 `loc` deprecation, so the `@ember/string` package can override it once it is deprecated.

### DIFF
--- a/packages/ember-glimmer/lib/setup-registry.ts
+++ b/packages/ember-glimmer/lib/setup-registry.ts
@@ -16,6 +16,7 @@ import ComponentTemplate from './templates/component';
 import OutletTemplate from './templates/outlet';
 import RootTemplate from './templates/root';
 import OutletView from './views/outlet';
+import loc from './helpers/loc';
 
 interface Registry {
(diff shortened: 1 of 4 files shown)


packages/ember-glimmer/lib/setup-registry.ts
```diff
@@ -66,6 +68,10 @@ export function setupEngineRegistry(registry: Registry) {
 
   registry.optionsForType('helper', { instantiate: false });
 
+  registry.register('helper:loc', helper(function loc(params) {
```
rwjblue: Can you put the implementation in a separate file?

rwjblue (commented):
Small nitpick but otherwise 👍

packages/ember-glimmer/lib/setup-registry.ts
locks: Sure! My first instinct is to create a new package called `ember-template` and put it in there, sg?

rwjblue (approved):

PR #15828, opened 2017-11-06
Upgrade glimmer-vm to latest

smfoote:
There is much left to do here. This is just the first pass, where I have
done nothing but try to clear out the errors caused by the upgrade. Not
all of the errors are cleared yet, largely because several exports from
glimmer-vm have been removed and I don't know how to replace them (help
wanted). Below is a list of the needed changes and problems I have
encountered in the upgrade process.

### UPGRADE NOTES (starting at glimmer-vm 0.25.6)

* Simple is now in @glimmer/interfaces
* ComponentManager.getSelf cannot return null (returns RootReference?)
* ComponentManager.getTag returns ComponentStateBucket.tag (: Tag)
* Some code moved from ComponentManager.update to
* ComponentManager.didUpdate
* ComponentManager<ComponentStateBucket> -->
* ComponentManager<ComponentStateBucket, ComponentDefinition>
* ComponentManager.create signature changed
* ComponentStateBucket's defintion is now DefinitionState instead of
* ComponentDefinition
* ComponentManager.prepareArgs signature changed
* ComponentManager.getCapabilities is new
* CompiledDynamicTemplate is gone (no longer in runtime, anyway) see
* https://github.com/glimmerjs/glimmer-vm/commit/79f2b4521f5a7d48d04e3fd34e224ddf5bc4c241#diff-d561208b0ad153d9273c2188e16c7442
* Must implicitly include @glimmer/vm and @glimmer/opcode-compiler and
* @glimmer/encoder in ember-cli-build.js
* compileLayout -> scanLayout -> prepareLayout -> new WrappedBuilder
* isSafeString is no longer exported publicly. Is that intentional?
* normalizeTextValue is no longer exported publicly. Is that
* intentional?
* isComponentDefinition is gone, kind of replaced by
* `isCurriedComponentDefinition`. Neither is exported.
* getComponentDefintion is no longer part of Environment.
* referenceFromParts has been removed.
* f3444e9ae0c7751047bdcd856e61344bfb548e12
* readDOMAttr has been removed. 316805b9175e01698120b9566ec51c88d075026a

PR #15862, opened 2017-11-18
Adds type annotations for ember-glimmer macro implementations

tomdale:
This PR adds annotations for the various Ember features implemented as Glimmer VM "macros" (i.e. they tap into the low-level opcode builder API). Previously these were mostly typed as `any`s. I skipped typing the `component` helper implementation due to time constraints, but it should be straightforward to derive its types from the others I've done here.

### DIFF
--- a/packages/ember-glimmer/lib/syntax/input.ts
+++ b/packages/ember-glimmer/lib/syntax/input.ts
@@ -1,12 +1,15 @@
 /**
 @module ember
 */
+import * as WireFormat from '@glimmer/wire-format';
 import { assert } from 'ember-debug';
 import { wrapComponentClassAttribute } from '../utils/bindings';
 import { dynamicComponentMacro } from './dynamic-component';
 import { hashToArgs } from './utils';
+import { OpcodeBuilderDSL } from '@glimmer/runtime';
+import { unwrap } from '@glimmer/util';
 
-function buildSyntax(type: string, params: any[], hash: any, builder: any) {
+function buildSyntax(type: string, params: WireFormat.Core.Params, hash: WireFormat.Core.Hash, builder: OpcodeBuilderDSL) {
   let definition = builder.env.getComponentDefinition(type, builder.meta.templateMeta);
   builder.component.static(definition, [params, hashToArgs(hash), null, null]);
   return true;
@@ -148,7 +151,7 @@ function buildSyntax(type: string, params: any[], hash: any, builder: any) {
   @public
 */
 
-export function inputMacro(_name: string, params: any[], hash: any[], builder: any) {
+export function inputMacro(_name: string, params: WireFormat.Core.Params, hash: WireFormat.Core.Hash, builder: OpcodeBuilderDSL): boolean {
   let keys;
   let values;
   let typeIndex = -1;
@@ -164,7 +167,7 @@ export function inputMacro(_name: string, params: any[], hash: any[], builder: a
   if (!params) { params = []; }
 
   if (typeIndex > -1) {
-    let typeArg = values[typeIndex];
+    let typeArg = unwrap(values)[typeIndex];
     if (Array.isArray(typeArg)) {
       return dynamicComponentMacro(params, hash, null, null, builder);
     } else if (typeArg === 'checkbox') {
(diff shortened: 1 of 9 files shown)


packages/ember-glimmer/lib/syntax/input.ts
```diff
@@ -1,12 +1,15 @@
 /**
 @module ember
 */
+import * as WireFormat from '@glimmer/wire-format';
 import { assert } from 'ember-debug';
 import { wrapComponentClassAttribute } from '../utils/bindings';
 import { dynamicComponentMacro } from './dynamic-component';
 import { hashToArgs } from './utils';
+import { OpcodeBuilderDSL } from '@glimmer/runtime';
+import { unwrap } from '@glimmer/util';
```
rwjblue: I believe this is stripped in glimmer-vm, but I don’t believe that is true here.

packages/ember-glimmer/lib/syntax/render.ts
```diff
@@ -128,12 +132,17 @@ function makeComponentDefinition(vm: VM, args: Arguments) {
...
     params = [];
   }
-  let definitionArgs = [params.slice(0), hash, null, null];
-  let args = [params.slice(1), hashToArgs(hash), null, null];
-  builder.component.dynamic(definitionArgs, makeComponentDefinition, args);
+  let definitionArgs: ComponentArgs = [params.slice(0), hash, null, null];
+  let args: ComponentArgs = [params.slice(1), hashToArgs(hash), null, null];
+
+  // We have to coerce makeComponentDefinition into an `any` here because
+  // Glimmer VM is typing this too restrictively as a PathReference when it
```
rwjblue: Can you cross link to the issue in glimmer-vm that you will be opening to track removal of the coercion?

packages/ember-glimmer/lib/utils/bindings.ts
```diff
@@ -50,19 +51,24 @@ export function wrapComponentClassAttribute(hash: any[]) {
   let index = keys.indexOf('class');
 
   if (index !== -1) {
-    let [ type ] = values[index];
+    let [ type ] = values[index] as WireFormat.Expressions.TupleExpression | string;
 
     if (type === Ops.Get || type === Ops.MaybeLocal) {
-      let getExp = values[index];
+      let getExp = expectTupleExpression(values[index]);
```
rwjblue: Would it be better to use `@glimmer/util`‘s `expect`?

packages/ember-glimmer/lib/syntax/render.ts
tomdale: This API no longer exists in Glimmer VM, so I don't think I will be opening that issue.

packages/ember-glimmer/lib/utils/bindings.ts
tomdale: `expect` asserts truthiness, but I need to assert arrayiness.

packages/ember-glimmer/lib/syntax/input.ts
tomdale: I had that question too, @chadhietala probably knows that stuff the most intimately.

packages/ember-glimmer/lib/syntax/render.ts
rwjblue: @tomdale - So a future update of glimmer-vm will remove this comment (and what it is doing)?  Can we link to _anything_ that will help us understand this when we come back to this code in 6 months? I fear without any pointers to additional things to dig into, this will become yet another murder mystery for our future selves to solve. 🤔

packages/ember-glimmer/lib/utils/bindings.ts
rwjblue: Gotcha, thank you!

packages/ember-glimmer/lib/syntax/render.ts
tomdale: @rwjblue It is already removed in @smfoote's upgrade branch.

rwjblue: Awesome, thank you (sorry for the run around).

tomdale:
Superseded by #15828.

PR #15871, opened 2017-11-27
[BUGFIX lts] Refactor / fix error handling scenarios.

rwjblue:
This PR has a few very small changes to the error handling system, but has a large impact to developer ergonomics.

The primary changes in this PR are:

### Revert changes made in https://github.com/emberjs/ember.js/pull/14898

https://github.com/emberjs/ember.js/pull/14898 essentially made all error handling for things that are run-wrapped uncatchable, dramatically impacting development ergonomics.

The originally reported issue is a _very real_ problem that we need to guard against. To reproduce that issue, the following conditions must exist:

* The application must implement `Ember.onerror` in a way that does not rethrow errors.
* Throw an error during anything that uses `run.join`. The example scenario had a template like this:

```hbs
<button {{action 'throwsAnError'}}>Click me!</button>
```

To fix this error swallowing behavior, the commit being reverted made all errors hit within the run loop use `dispatchError`, which (during tests) has the default implementation of invoking `QUnit`'s `assert.ok(false)`. Unfortunately, this meant that it is now impossible to use a standard `try` / `catch` to catch errors thrown within anything "run-wrapped".

For example, these patterns were no longer possible after the commit in question:

```js
try {
  Ember.run(() => {
    throw new Error('This error should be catchable');
  });
} catch(e) {
  // this will never be hit during tests...
}
```

This ultimately breaks a large number of test suites that rely (rightfully so!) on being able to do things like:

```js
module('service:foo-bar', function(hooks) {
  setupTest(hooks);

  hooks.beforeEach(() => {
    this.owner.register('service:whatever', Ember.Service.extend({
      someMethod(argumentHere) {
        Ember.assert('Some random argument validation here', !argumentHere);
      }
    });
  });

  test('throws when argumentHere is missing', function(assert) {
    let subject = this.owner.lookup('service:foo-bar');

    assert.throws(() => {
      run(() =>
        subject.someMethod());
    }, /random argument validation/);
  });
});
```

The ergonomics of breaking standard JS `try` / `catch` is too much of a step backwards, so it is being reverted.

As mentioned earlier, the underlying problem reported in https://github.com/emberjs/ember.js/issues/14864 is very real and **must** be addressed also. Therefore, in order to prevent `Ember.onerror` from swallowing errors during tests `ember-qunit` (and `ember-mocha`) will implement checks to ensure that `Ember.onerror` (if present) properly rethrows errors during testing. The following is psuedo code of what that might look like:

```js
test('Ember.onerror is functioning properly', function(assert) {
  let thrown = new Error('onerror bubbling');
  if (!Ember.onerror) {
    assert.ok(true, 'no Ember.onerror is setup');
  } else {
    assert.throws(() => {
      Ember.onerror(thrown);
    }, /onerror bubbling/, 'Ember.onerror should rethrow when Ember.testing is true');
  }
});

test('window.onerror is functioning properly', function(assert) {
  assert.expect(1);

  let thrown = new Error('onerror bubbling');
  let originalQUnitOnError = QUnit.onError;
  QUnit.onError = function() {
    assert.ok(true, 'QUnit.onError is called');
  };

  try {
    let result = window.onerror(thrown.message);
  } finally {
    QUnit.onError = originalQUnitOnError;
  }
});
```

### Remove the requirement of `Ember.Test.Adapter`'s to implement the `exception` method.

In the majority of testing frameworks (e.g. Mocha, QUnit, Jasmine) a global error handler is attached to `window.onerror`. When an uncaught exception is thrown, this global handler is invoked and the test suite will properly fail.

Requiring that the provided test adapter provide an exception method is strictly worse (in all cases) than allowing the error to be bubbled upwards and caught/handled by either `Ember.onerror` (if present) or `window.onerror` (generally from the test framework).

Over time `Ember.Test.adapter.exception` should be deprecated in favor of relying on `Ember.onerror` and/or `window.onerror`.

### Simplify unhandled promise rejection bubbling.

When used in Ember, `RSVP` is configured to settle within Backburner's configured `actions` queue.

Prior to this change, any unhandled promise rejections would:

* `Ember.testing === true`
  * When `Ember.Test.adapter` was registered the `Test.adapter`'s `exception` method would be called, and the rejection would be logged to the console.
  * When `Ember.Test.adapter` was not registered, the `defaultDispatch` implementation would re-throw the error, and since RSVP settles in the run loop this means that the re-thrown error would be caught by the currently flushing Backburner queue's `invokeWithOnError` which sends any errors to `Ember.onerror` (if present). If `Ember.onerror` was not present, the exception would bubble up the "normal" unhandled exception system (and ultimately to `window.onerror`).
* `Ember.testing === false`
  * When `Ember.onerror` is present, it would be invoked with the rejection reason.
  * When `Ember.onerror` is not present, the rejection reason would be logged to the console.

After this change:

* When `Ember.Test.adapter` is present, its `exception` method is invoked with the rejection.
* Otherwise the rejection reason is rethrown.

The benefits of this are:

* It is now possible to debug rejected promises via "normal" JS exception debugging (e.g. break on uncaught exception).
* There are many fewer decision points, making it much easier to grok what is going on.

---

The majority of the changes are in new tests being added, I attempted to ensure that each of the following were under test:

* Testing
  * With `Ember.Test.adapter`
    * With `Ember.onerror`
      * Sync Backburner (`run` and `run.join`)
      * Async Backburner (`run.later`, `run.next`)
      * RSVP
    * Without `Ember.onerror`
      * Sync Backburner (`run` and `run.join`)
      * Async Backburner (`run.later`, `run.next`)
      * RSVP
  * Without `Ember.Test.adapter`
    * With `Ember.onerror`
      * Sync Backburner (`run` and `run.join`)
      * Async Backburner (`run.later`, `run.next`)
      * RSVP
    * Without `Ember.onerror`
      * Sync Backburner (`run` and `run.join`)
      * Async Backburner (`run.later`, `run.next`)
      * RSVP
* Development / Production
  * With `Ember.onerror`
    * Sync Backburner (`run` and `run.join`)
    * Async Backburner (`run.later`, `run.next`)
    * RSVP
  * Without `Ember.onerror`
    * Sync Backburner (`run` and `run.join`)
    * Async Backburner (`run.later`, `run.next`)
    * RSVP

Each commit in this PR attempts to "stand on its own" and explain the reasoning for the changes they contain. As such, it may be easier to review commit by commit instead of all at once.

### DIFF
--- a/packages/ember-metal/lib/error_handler.js
+++ b/packages/ember-metal/lib/error_handler.js
@@ -1,18 +1,6 @@
 import Logger from 'ember-console';
 import { isTesting } from 'ember-debug';
 
-// To maintain stacktrace consistency across browsers
-let getStack = error => {
-  let stack = error.stack;
-  let message = error.message;
-
-  if (stack && stack.indexOf(message) === -1) {
-    stack = `${message}\n${stack}`;
-  }
-
-  return stack;
-};
-
 let onerror;
 export const onErrorTarget = {
   get onerror() {
@@ -30,14 +18,6 @@ export function setOnerror(handler) {
 }
 
 let dispatchOverride;
-// dispatch error
-export function dispatchError(error) {
-  if (dispatchOverride) {
-    dispatchOverride(error);
-  } else {
-    defaultDispatch(error);
-  }
-}
 
 // allows testing adapter to override dispatch
 export function getDispatchOverride() {
@@ -45,15 +25,4 @@ export function getDispatchOverride() {
 }
 export function setDispatchOverride(handler) {
   dispatchOverride = handler;
-}
-
-function defaultDispatch(error) {
-  if (isTesting()) {
-    throw error;
-  }
-  if (onerror) {
-    onerror(error);
-  } else {
-    Logger.error(getStack(error));
-  }
 }
--- a/packages/ember-metal/lib/error_handler.js
+++ b/packages/ember-metal/lib/error_handler.js
@@ -16,7 +16,7 @@ let getStack = error => {
 let onerror;
 export const onErrorTarget = {
   get onerror() {
-    return dispatchOverride || onerror;
+    return onerror;
   }
 };
 
--- a/packages/ember-metal/lib/index.js
+++ b/packages/ember-metal/lib/index.js
@@ -19,7 +19,6 @@ export {
 export {
   getOnerror,
   setOnerror,
-  dispatchError,
   setDispatchOverride,
   getDispatchOverride
 } from './error_handler';
--- a/packages/ember-runtime/lib/ext/rsvp.js
+++ b/packages/ember-runtime/lib/ext/rsvp.js
@@ -1,7 +1,7 @@
 import * as RSVP from 'rsvp';
 import {
   run,
-  dispatchError
+  getDispatchOverride
 } from 'ember-metal';
 import { assert } from 'ember-debug';
 
@@ -21,7 +21,12 @@ RSVP.on('error', onerrorDefault);
 export function onerrorDefault(reason) {
   let error = errorFor(reason);
   if (error) {
-    dispatchError(error);
+    let overrideDispatch = getDispatchOverride();
+    if (overrideDispatch) {
+      overrideDispatch(error);
+    } else {
+      throw error;
+    }
   }
 }
 
--- a/packages/ember-runtime/tests/ext/rsvp_test.js
+++ b/packages/ember-runtime/tests/ext/rsvp_test.js
@@ -46,12 +46,10 @@ QUnit.test('Can reject with non-Error object', function(assert) {
   try {
     run(RSVP, 'reject', 'foo');
   } catch (e) {
-    ok(false, 'should not throw');
+    equal(e, 'foo', 'should throw with rejection message');
   } finally {
     setTesting(wasEmberTesting);
   }
-
-  ok(true);
 });
 
 QUnit.test('Can reject with no arguments', function(assert) {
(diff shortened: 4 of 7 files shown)


stefanpenner (approved):

packages/ember-metal/lib/error_handler.js
```diff
@@ -30,14 +30,6 @@ export function setOnerror(handler) {
 }
 
 let dispatchOverride;
-// dispatch error
-export function dispatchError(error) {
```
rwjblue: This was previously only used by `RSVP.on('error', ...)` (after reverting https://github.com/emberjs/ember.js/pull/14898), and `RSVP.on('error', ...)` in this PR now does the logic of `dispatchError` directly (which makes it much clearer that this is RSVP error handling specific).

fivetanley:
:pray:

rwjblue:
Rebased after landing https://github.com/emberjs/ember.js/pull/15874 to kick off another CI run using the latest Backburner version...

hjdivad (approved):

rwjblue:
Folks stumbling across this PR may be interested: @hjdivad and I have just published an updated version of [ember-test-friendly-error-handler](https://github.com/rwjblue/ember-test-friendly-catch-handler) addon which makes creating an `Ember.onerror` that properly rethrows in development but still allows testing of the "production mode".

rwjblue:
From the original PR description:

> As mentioned earlier, the underlying problem reported in #14864 is very real and must be addressed also. Therefore, in order to prevent Ember.onerror from swallowing errors during tests ember-qunit (and ember-mocha) will implement checks to ensure that Ember.onerror (if present) properly rethrows errors during testing.


Finally have this implemented: https://github.com/emberjs/ember-qunit/pull/304.

Issue #15864, opened 2017-11-20
promise chains and the bizarre consequence of registering a handler with Ember.onerror


kbullaughey:
It seems that when upgrading from 2.13 to 2.16 the effect of registering a global error handler using `Ember.onerror` has changed how promise chains are handled.

I have been using a pattern of handling errors in long promise chains that involves having step-specific reject handlers. Each of these handlers would be responsible for handling any errors in the previous step. Additionally, these handlers would continue rejecting all the way down the chain so that even if an error is handled, none of the rest of the chain is executed.

Unfortunately, this pattern seems not to be possible now when registering an error handler with `Ember.onerror`. The example in the Guides suggests that `Ember.onerror` can be used for logging errors. I would have expected that whether one registeres an error handler with `Ember.onerror` shouldn't affect which methods are called in the promise chain. However, in 2.16 this seems to be the case.

I illustrate this behavior in [an example repo](https://github.com/kbullaughey/ember-effect-of-onerror).

In the file `app/controllers/application.js` there is a boolean variable `registerGlobalHandler` that can be toggled to determine if a global error handler using `Ember.onerror` is registered.

When `registerGlobalHandler` is set to `false`, I get the expected behavior:

    problem in step 2: Error: step 1 error handler failed
    rsvp error: reraise

When `registerGlobalHandler` is set to `true`, the promise chain is continued, despite the error thrown in `step1Problem`, which is a reject handler passed to `then`.

In this case, the output looks like this:

    global error handler: Error: step 1 error handler failed
    step 3
    success

This behavior is quite problematic because it means that the chain will continue if there is an exception thrown in the `reject` handler. Also, it's bizarre that this behavior is contingent on whether a handler was registered with `Ember.onerror`.

kbullaughey:
I was able to reproduce the 2.13 behavior by starting with what's currently in the ember.js master branch (a 2.17 beta) and downgrading backburner to 0.4.0. I haven't yet worked out what changed in backburner to cause this problem.

rwjblue:
After reviewing all the wonderful research @kbullaughey did here (and in https://github.com/BackburnerJS/backburner.js/issues/274) along with a lengthy pairing session with @hjdivad, I believe that https://github.com/BackburnerJS/backburner.js/pull/277 fixes the issue and explains exactly what is happening.

PR #15873, opened 2017-11-28
[BUGFIX release] Update Backburner.js to 1.2.3.

rwjblue:
Includes fixes for `Ember.run.join` swallowing errors when an `Ember.onerror` exists.

Fixes https://github.com/emberjs/ember.js/issues/15864

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "babel-plugin-transform-es2015-template-literals": "^6.22.0",
     "babel-plugin-transform-proto-to-assign": "^6.23.0",
     "babel-template": "^6.24.1",
-    "backburner.js": "^1.2.2",
+    "backburner.js": "1.2.3",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -80,7 +80,7 @@
     "babel-plugin-transform-es2015-template-literals": "^6.22.0",
     "babel-plugin-transform-proto-to-assign": "^6.23.0",
     "babel-template": "^6.24.1",
-    "backburner.js": "^1.2.2",
+    "backburner.js": "1.2.3",
```
rwjblue: Pinned to exact version to avoid getting `backburner.js@1.3.0` (which contains other changes that should flow through a full canary / beta cycle) in beta.  

See https://github.com/emberjs/ember.js/pull/15874 for that bump (canary only).

PR #15881, opened 2017-11-29
Removed console polyfills/shims

t-sauer:
I think it is safe to add `console` of available globals in the ESLint config now that IE 9 support is removed, so I could also change that.

### DIFF
--- a/packages/ember-console/lib/index.js
+++ b/packages/ember-console/lib/index.js
@@ -1,37 +1,3 @@
-import { context } from 'ember-environment';
-
-function K() {}
-
-function consoleMethod(name) {
-  let consoleObj;
-  if (context.imports.console) {
-    consoleObj = context.imports.console;
-  } else if (typeof console !== 'undefined') { // eslint-disable-line no-undef
-    consoleObj = console; // eslint-disable-line no-undef
-  }
-
-  let method = typeof consoleObj === 'object' ? consoleObj[name] : null;
-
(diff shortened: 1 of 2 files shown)


packages/ember-console/lib/index.js
```diff
@@ -56,7 +22,7 @@ export default {
    @param {*} arguments
    @public
   */
-  log: consoleMethod('log') || K,
+  log: console.log, // eslint-disable-line no-console, no-undef
```
rwjblue: I'm _slightly_ torn here, I don't know if we can use `console.log` unbound from `console` for all supported platforms.  Do you know?

The safe bet here is essentially to do:

```js
log() { return console.log(...arguments); }
```

t-sauer: Tried `var log = console.log; log('foo');` in Safari, Firefox, Chrome and IE 11. Worked everywhere.  
I can change it to be safe though.

rwjblue: Checkout this MS Edge screenshot: 

MS Edge failure

I think this is for `console.debug`?

rwjblue: @t-sauer - No rush here, just curious if you had time to review?

t-sauer: I am on it right now. Seems like `.bind(console)` is indeed needed in some cases.

rwjblue (approved):
Thank you!

Issue #15624, opened 2017-09-02
[RFC 236] Ember.String Deprecation for @ember/string


locks:
# MIGRATED TO https://github.com/emberjs/rfc-tracking/issues/26

This is a meta-issue to track the implementation of https://github.com/emberjs/rfcs/blob/master/text/0236-deprecation-ember-string.md.

- ember.js
  - [ ] [deprecate Ember.String](https://github.com/emberjs/ember.js/pull/15739)
  - [x] [move htmlSafe and isHTMLSafe](https://github.com/emberjs/ember.js/pull/15626)
- ember-string -> `@ember/string`
  - [x] https://github.com/emberjs/ember-string/pull/1
- ember-string-prototype-extensions
  - [ ] https://github.com/locks/ember-string-prototype-extensions/pull/1
- ember-cli-babel
  - [x] [blacklist `@ember/string` if present](https://github.com/babel/ember-cli-babel/pull/176)
- [x] [babel-plugin-ember-modules-api-polyfill](https://github.com/ember-cli/babel-plugin-ember-modules-api-polyfill)
- ember-rfc176-data
  - [x] [Move `htmlSafe` and `isHTMLSafe` to `@ember/template`](https://github.com/ember-cli/ember-rfc176-data/pull/38)
- website
  - [ ] [Deprecation guide](https://github.com/emberjs/website/pull/2995)

Serabe:
@locks PTAL at https://github.com/ember-cli/ember-rfc176-data/pull/36 for the last part.

Serabe:
Another one https://github.com/ember-cli/ember-rfc176-data/pull/38

Serabe:
I've rebased everything except https://github.com/locks/ember-string-prototype-extensions/pull/1 and https://github.com/emberjs/website/pull/2995.

pixelhandler:
@Serabe @locks This has been inactive for some time, is this up to date? what remains todo?

Serabe:
AFAIK, everything. I closed my PR because I was rebasing it for over six months. Don't know if core is still interested on deprecating String prototype extensions.

locks:
I am back to being more active again, I'm aiming to tackle this before year end.

PR #15882, opened 2017-11-29
Remove Ember.String.fmt

locks:


rwjblue:
Seems good to me.  We need to ensure this functionality sticks around when you have `ember-2-legacy` addon installed, but its super easy to just move the implementation over there. 

Can you make an issue over on https://github.com/emberjs/ember-2-legacy to add support to add these back when the legacy addon is present?  I just don't want to lose track...

packages/ember-runtime/lib/system/string.js
```diff
@@ -49,35 +49,6 @@ const STRING_DECAMELIZE_REGEXP = (/([a-z\d])([A-Z])/g);
 
 const DECAMELIZE_CACHE = new Cache(1000, str => str.replace(STRING_DECAMELIZE_REGEXP, '$1_$2').toLowerCase());
 
-function _fmt(str, formats) {
```
sduquej: `_fmt` is used in `loc`:

https://github.com/emberjs/ember.js/blob/a7518b940908994ae0577973987fd4cb6a5f7bee/packages/ember-runtime/lib/system/string.js#L81-L88

and also referenced in the [documentation for it](https://emberjs.com/api/ember/2.17/classes/String/methods/loc?anchor=loc).

Think we'll need to update these too.

lorcan:
`Ember.String.loc` also depends on the same functionality of `Ember.String.fmt`. Is there a case for deprecating `Ember.String.loc`? Too late for 3.0.0, but worth flagging?

locks:
@lorcan thanks for the review. `loc` is already deprecated by [RFC 236](https://github.com/emberjs/rfcs/blob/master/text/0236-deprecation-ember-string.md), and you can follow the implementation of this deprecation in https://github.com/emberjs/ember.js/issues/15624.
As mentioned in [The Road to Ember 3.0](https://emberjs.com/blog/2017/10/03/the-road-to-ember-3-0.html), "No additional 3.0-targeted deprecations will be introduced."

If we manage to finish all the `Ember.String` work soon, the deprecation should land in the 3.x series, with an until of 4.0.0.

rwjblue (approved):

PR #15883, opened 2017-11-29
Remove this.resource from RouterDSL

locks:


rwjblue:
Since this is exposed as `Ember.RouterDSL` I believe we can definitely move the implementation directly into the legacy addon. Just need to make an issue on ember-2-legacy for this one, and cross link back here.

packages/ember-routing/tests/system/dsl_test.js
```diff
@@ -18,7 +18,6 @@ QUnit.module('Ember Router DSL', {
 });
 
 QUnit.test('should fail when using a reserved route name', function() {
-  expectDeprecation('this.resource() is deprecated. Use this.route(\'name\', { resetNamespace: true }, function () {}) instead.');
   let reservedNames = ['array', 'basic', 'object', 'application'];
 
   expect((reservedNames.length * 2) + 1);
```
rwjblue: Need to tweak this somewhat, the CI failure is due to it thinking it has 9 assertions and only actually getting 5

rwjblue:
@locks - Needs a rebase...

locks:
@rwjblue rebased

rwjblue (approved):

PR #15884, opened 2017-11-29
Remove currentWhen property from link-to component

locks:


rwjblue:
Since this is exposed as `Ember.LinkComponent`, I believe we can definitely move the implementation directly into the legacy addon (and use `.reopen`). Just need to make an issue on ember-2-legacy for this one, and cross link back here.

PR #15892, opened 2017-11-30
[CLEANUP] removed `Ember.required`

bekzod:


rwjblue:
I'm not 100% certain that this can be implemented in the addon. I believe we may need to go with the conditional flagging on this one.

@thoov - What do you think?

thoov:
Agreed. Its going to have to be flagged. @bekzod I would look here for some examples: https://github.com/emberjs/ember.js/pull/15900

rwjblue:
@bekzod - ping (when you have a chance)

bekzod:
@rwjblue so it should be runtime flag ? flags are not stripped in build time ?

rwjblue:
Confirm. It has to be a runtime flag because we cannot polyfill it from the ember-2-legacy addon.

bekzod:
does this look better ?

packages/ember-metal/tests/mixin/required_test.js
```diff
@@ -4,11 +4,14 @@ import {
...
   get
 } from '../..';
+import { ENV } from 'ember-environment';
 
 let PartialMixin, FinalMixin, obj;
-
+let originalEnvVal;
 QUnit.module('Module.required', {
   setup() {
+    originalEnvVal = ENV._ENABLE_PROPERTY_REQUIRED_SUPPORT
```
thoov: Missing a `;` here

packages/ember/lib/index.js
```diff
@@ -127,7 +127,6 @@ Ember.observersFor = metal.observersFor;
 Ember.removeObserver = metal.removeObserver;
 Ember._suspendObserver = metal._suspendObserver;
 Ember._suspendObservers = metal._suspendObservers;
-Ember.required = metal.required;
```
thoov: I believe we are going to have to guard this

packages/ember/tests/reexports_test.js
```diff
@@ -108,7 +108,6 @@ QUnit.module('ember reexports');
   ['removeObserver', 'ember-metal'],
   ['_suspendObserver', 'ember-metal'],
   ['_suspendObservers', 'ember-metal'],
-  ['required', 'ember-metal'],
```
thoov: I believe we are going to have to guard this

thoov (approved):

packages/ember/lib/index.js
bekzod: ok

packages/ember/tests/reexports_test.js
bekzod: ok

thoov:
LGTM @rwjblue can you take a look

rwjblue (approved):

krisselden:
In practice, this required doesn't do anything because of applyPartial and finishPartial doesn't handle it.  It only affects mixin.apply() which is rare.   For upgrading an addon could just make required() present and return null.

PR #14808, opened 2017-01-10
Remove shims for now

chancancode:
cc @rwjblue

PR #15898, opened 2017-11-30
Remove custom implementations of Map/MapWithDefault/OrderedSet

maxscott:
#13815 

`MapWithDefault`'s usage in ember-data is accounted for [here (emberjs/data/pull/5255)](https://github.com/emberjs/data/pull/5255)
`OrderedSet` should be installed as an addon by any project using it.

This keeps the vendor shim's `'ember-map'.default = Ember.Map`, which is now native `Map`. Unsure if it's better to remove the alias/shim altogether.


Turbo87:
> Unsure if it's better to remove the alias/shim altogether.

IMHO it should be removed completely

@rwjblue why are there shims in this repo at all? I thought only `ember-cli-shims` provided them?! 🤔

rwjblue:
> @rwjblue why are there shims in this repo at all? I thought only ember-cli-shims provided them?! 🤔

They are not used here. We noop'ed them just before initial npm release (by simply not `app.import`'ing them) in https://github.com/emberjs/ember.js/pull/14808, but never came back to clean up.  They should definitely be removed, but in a separate PR...

rwjblue:
I submitted https://github.com/emberjs/ember.js/pull/15909 to remove.

maxwerr:
@Turbo87 rebased. shims.js removed

maxwerr:
I figured today would be a good time to delete a bunch of private, seemingly unused code before someone else does! Lmk if this requires anything else I can do.

Turbo87 (approved):
I've rebased this PR. CI is green. v3.5.0 has been released a long time ago. I'd say this is ready to be merged.

/cc @rwjblue

rwjblue:
Thanks @maxscott and @Turbo87!

PR #15788, opened 2017-10-29
remove `routeArgs`

bekzod:
`routeArgs` makes things confusing it is better without it I think :)

PR #15902, opened 2017-11-30
[BUGFIX beta] Fix link-to throwing in integration tests

### LINKED ISSUE #15831: link-to in integration tests throw: Cannot read property 'applyIntent' of undefined
The following error occurs on canary (but works properly on prior releases) for [this test](https://github.com/rwjblue/___demo-integration-test-with-link-to-bug/blob/master/tests/integration/components/can-render-link-to-test.js):

```
not ok 47 Chrome 62.0 - component:o it can render a link-to
    ---
        actual: >
            false
        expected: >
            true
        stack: >
            Error: Assertion Failed: You attempted to define a `{{link-to "index"}}` but did not pass the parameters required for generating its dynamic segments. Cannot read property 'applyIntent' of undefined
                at new EmberError (http://localhost:7357/assets/vendor.js:13925:25)
                at assert (http://localhost:7357/assets/vendor.js:14168:15)
                at Class.computeLinkToComponentHref (http://localhost:7357/assets/vendor.js:17044:65)
                at ComputedPropertyPrototype.get (http://localhost:7357/assets/vendor.js:25789:28)
                at get (http://localhost:7357/assets/vendor.js:25131:20)
                at RootPropertyReference.compute (http://localhost:7357/assets/vendor.js:21709:40)
                at RootPropertyReference.value (http://localhost:7357/assets/vendor.js:21590:53)
                at ReferenceCache.initialize (http://localhost:7357/assets/vendor.js:924:52)
                at ReferenceCache.peek (http://localhost:7357/assets/vendor.js:898:29)
                at DynamicAttribute.flush (http://localhost:7357/assets/vendor.js:2776:37)
        message: >
            Assertion Failed: You attempted to define a `{{link-to "index"}}` but did not pass the parameters required for generating its dynamic segments. Cannot read property 'applyIntent' of undefined
```

The test is attempting to ensure that a component integration test is capable of rendering a `{{link-to}}` even when the router is not booted/available.

[Demo repo here](https://github.com/rwjblue/___demo-integration-test-with-link-to-bug/blob/master/tests/integration/components/can-render-link-to-test.js)

---

I suspect it was caused by one of the following PR's which have changed things in the routing and/or link-to area:

* https://github.com/emberjs/ember.js/pull/15788
* https://github.com/emberjs/ember.js/pull/15779
* https://github.com/emberjs/ember.js/pull/15762
* https://github.com/emberjs/ember.js/pull/15811

simonihmig:
Fixes #15831

Fixes a regression introduced by #15788 by returning early again from `routing.generateURL()` when `router._routerMicrolib` is not present. [This is the line](https://github.com/emberjs/ember.js/pull/15788/files#diff-f0eec040f8686d2b0ddec02e3eeab6e4L55) that got lost during the refactoring.

Was not able to write a regression test for this, as e.g. in [this test](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/tests/integration/components/link-to-test.js#L18-L24) `_routerMicrolib` is correctly setup, so was not able to write a failing test. If there is a way to setup the router in the same state as in integration tests (i.e. without `_routerMicrolib`), then I would be happy for any hint! 😉

Would be nice to get this into a minor beta release, as this is causing failing tests in ember-try scenarios for ember-beta now (was already failing for canary, but these were "allowed" failures. Now that canary has been promoted to beta, this is causing failed builds).

cc @rwjblue @bekzod

### DIFF
--- a/packages/ember-routing/lib/services/routing.js
+++ b/packages/ember-routing/lib/services/routing.js
@@ -48,13 +48,17 @@ export default Service.extend({
   },
 
   generateURL(routeName, models, queryParams) {
+    let router = get(this, 'router');
+    // return early when the router microlib is not present, which is the case for {{link-to}} in integration tests
+    if (!router._routerMicrolib) { return; }
+
     let visibleQueryParams = {};
     if (queryParams) {
       assign(visibleQueryParams, queryParams);
       this.normalizeQueryParams(routeName, models, visibleQueryParams);
     }
 
-    return get(this, 'router').generate(routeName, ...models, { queryParams: visibleQueryParams });
+    return router.generate(routeName, ...models, { queryParams: visibleQueryParams });
   },
 
   isActiveForRoute(contexts, queryParams, routeName, routerState, isCurrentWhenSpecified) {
--- a/packages/ember-glimmer/tests/integration/components/link-to-test.js
+++ b/packages/ember-glimmer/tests/integration/components/link-to-test.js
@@ -1,4 +1,4 @@
-import { moduleFor, ApplicationTest } from '../../utils/test-case';
+import { moduleFor, ApplicationTest, RenderingTest } from '../../utils/test-case';
 import { Controller } from 'ember-runtime';
 import { set } from 'ember-metal';
 import { LinkComponent } from '../../utils/helpers';
@@ -174,5 +174,19 @@ moduleFor('Link-to component with query-params', class extends ApplicationTest {
         content: 'Index'
       });
     });
+  }
+});
+
+moduleFor('Link-to component', class extends RenderingTest {
+  ['@test should be able to be inserted in DOM when the router is not present - block']() {
+    this.render(`{{#link-to 'index'}}Go to Index{{/link-to}}`);
+
+    this.assertText('Go to Index');
+  }
+
+  ['@test should be able to be inserted in DOM when the router is not present - inline']() {
+    this.render(`{{link-to 'Go to Index' 'index'}}`);
+
+    this.assertText('Go to Index');
   }
 });


packages/ember-routing/lib/services/routing.js
```diff
@@ -48,13 +48,16 @@ export default Service.extend({
   },
 
   generateURL(routeName, models, queryParams) {
+    let router = get(this, 'router');
+    if (!router._routerMicrolib) { return; }
```
rwjblue: Can you add a comment above this to explain why its needed?

rwjblue (commented):
To add a test I think we would need to have a test module that extends from rendering test (but not application test).

Something like what you linked, but slightly different:

```js
import { moduleFor, RenderingTest } from '../../utils/test-case';

moduleFor('Link-to component', class extends RenderingTest {
  ['@test should be able to be inserted in DOM when the router is not present - block']() {
    this.addTemplate('application', `{{#link-to 'index'}}Go to Index{{/link-to}}`);

    this.render(`{{#link-to 'index'}}Go to Index{{/link-to}}`);
    this.assertText('Go to Index');
  }

  ['@test should be able to be inserted in DOM when the router is not present - inline']() {
    this.addTemplate('application', `{{#link-to 'index'}}Go to Index{{/link-to}}`);

    this.render(`{{link-to 'Go to Index' 'index'}}`);
    this.assertText('Go to Index');
  }
});
```

You can add ^ to the same file, just below the current `moduleFor`...

packages/ember-routing/lib/services/routing.js
simonihmig: done

simonihmig:
> You can add ^ to the same file, just below the current moduleFor...

Thanks, that was very helpful! The `this.addTemplate()` calls were not needed and failing, but other than that this worked fine and indeed captured the error throwing the same exception as described in #15831!  🚀

rwjblue:
Merging here should auto publish the fix for `ember-canary`, but I’ll have to cherry-pick into beta branch to get a fixed ember-beta build out. Will try to do that as soon as I’m back to a computer (out at a play with the family for a bit).

simonihmig:
> Merging here should auto publish the fix for ember-canary, but I’ll have to cherry-pick into beta branch to get a fixed ember-beta build out.

FWIW this PR was targeting beta, but I see it has been cherry-picked to master already...

rwjblue:
Ya, I missed that while reviewing + merging on mobile (apparently GH's mobile view doesn't show the target branch?!!?!?).  @kategengler fixed it for us though.

simonihmig:
@rwjblue just for clarification/learning: isn't a `[BUGFIX beta]` PR supposed to target the beta branch? Or should I have targeted master instead?

PR #15905, opened 2017-12-01
Improve error message when calling `inject()`.

### LINKED ISSUE #15904: Confusing message when using wrong `inject`
After upgrading to the new modules syntax, I had to revert one small section of my app (complicated story; mostly irrelevant), and in one module I missed converting `inject()` back to `inject.service()`. The error message reported is… extremely confusing:

> "Assertion Failed: Injected properties must be created through helpers, see 'controller"service'"

I managed to figure it out based on the fact that I knew I'd messed with rewriting injections in those particular modules, but I can imagine that being *especially* confusing if you didn't. (Note that the codemod eliminates this problem by doing `import { inject as service } from '@ember/service'`, so the error would be different—and clearer—in that case.

This error message should be much clearer.

chriskrycho:
Fixes #15904

### DIFF
--- a/packages/ember-runtime/lib/inject.js
+++ b/packages/ember-runtime/lib/inject.js
@@ -10,7 +10,10 @@ import { assert } from 'ember-debug';
   @public
 */
 export default function inject() {
-  assert(`Injected properties must be created through helpers, see '${Object.keys(inject).join('"', '"')}'`);
+  const helpers = Object.keys(inject)
+    .map(k => `'inject.${k}'`)
+    .join(' or ');
+  assert(`Injected properties must be created through helpers, see '${helpers}'`);
 }
 
 // Dictionary of injection validations by type, added to by `createInjectionHelper`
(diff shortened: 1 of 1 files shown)


packages/ember-runtime/lib/inject.js
```diff
@@ -10,7 +10,10 @@ import { assert } from 'ember-debug';
   @public
 */
 export default function inject() {
-  assert(`Injected properties must be created through helpers, see '${Object.keys(inject).join('"', '"')}'`);
+  const helpers = Object.keys(inject)
```
bekzod: I think this should inlined inside `assert` else it won't be stripped

chriskrycho: I wondered about that, and can push an update to that effect tomorrow.

Serabe: That would solve the linting problem as well. Thank you!

chriskrycho: Updated!

rwjblue (approved):

rwjblue:
Thank you!

PR #15907, opened 2017-12-01
[BUGFIX] Ensure findBy works for key with dot.

sarathkumarbaskaran:
Ensure `Ember.findBy()` works for keys with dot. 

``` javascript
let arr = [
 { canada: [....] },
 { 'u.s.a.': [....] }
];
Ember.get(arr, 'u.s.a'); // returns `undefined`
```
Ember get() assumes this as a path, since the key has dot.

To solve this, we have two options, 
1. Revert the  `get(obj, key)` usage here.
2. Add additional validations to get method.

The second option is little complicated and it disturbs the default behavior of get method for path lookup.


sarathkumarbaskaran:
Closing this in favor of !15913

PR #15908, opened 2017-12-01
use `Map` for `Cache` default store

bekzod:


rwjblue:
Seems good to me.

@krisselden - Any objections / concerns?

mmun (approved):

PR #15903, opened 2017-12-01
[CLEANUP] Remove deprecate checks inside of Ember.deprecate

thoov:
Related: [ember-2-legacy](https://github.com/emberjs/ember-2-legacy/pull/11)

Apart of: #15876

PR #15911, opened 2017-12-01
[CLEANUP] Change resolver function deprecations in to assert

gowthamrm:


rwjblue:
Can you create a corresponding PR (or issue that links back here) at https://github.com/emberjs/ember-2-legacy?

rwjblue:
@thoov - Do you think we need to ENV flag guard this one?  How would we implement the legacy addon implementation otherwise?

packages/container/lib/registry.js
```diff
@@ -24,10 +24,6 @@ export default class Registry {
     this.fallback = options.fallback || null;
     this.resolver = options.resolver || null;
 
-    if (typeof this.resolver === 'function') {
```
thoov: I think we are going to have to put an ENV guard around this. [Here is an example of how](https://github.com/thoov/ember.js/blob/64d4708273d86137f8d2e8e28b3303316b3270e4/packages/ember-debug/lib/warn.js#L90) I just did a similar thing.

thoov:
@rwjblue Yes I think the simplest approach is to wrap it in an ENV. The only other ways would been to extend the registry past the point of it being worth it the work.

packages/container/lib/registry.js
thoov: Also here is an example of adding the config to [ember-2-legacy](https://github.com/emberjs/ember-2-legacy/pull/9)

gowthamrm: @thoov @rwjblue I have updated the PR with the ENV guard, Please have a look.

gowthamrm:
@rwjblue @thoov Can you check this out? Is there anything else need to be added?

packages/container/tests/registry_test.js
```diff
@@ -473,9 +474,11 @@ QUnit.test('`knownForType` is called on the resolver if present', function() {
   });
 });
 
-QUnit.test('A registry can be created with a deprecated `resolver` function instead of an object', function() {
+QUnit.test('A registry created with `resolver` function instead of an object throws deprecation', function() {
   expect(2);
 
+  ENV._ENABLE_RESOLVER_FUNCTION_SUPPORT = true;
```
thoov: We will want to be reset this value. You can look at this example for how I was reseting the ENV properties (https://github.com/emberjs/ember.js/pull/15903/files#diff-7c1b00291152c89da72c8940b7305c1a)

thoov:
@gowthamrm One final thing then it looks good for me. Thanks!

gowthamrm:
@thoov Thanks for your comments, I have updated the test now. 👍

packages/container/tests/registry_test.js
```diff
@@ -489,6 +503,20 @@ QUnit.test('A registry can be created with a deprecated `resolver` function inst
   equal(registry.resolve('foo:bar'), 'foo:bar-resolved', '`resolve` still calls the deprecated function');
 });
 
+QUnit.test('A registry created with `resolver` function instead of an object throws assertion', function() {
+  expect(1);
+
+  let registry;
+
+  throws(() => {
```
rwjblue: This should be swapped to use `expectAssertion` (so that it works properly in production builds since `Ember.assert` does not throw there).

rwjblue (commented):
Looking good, thank you for working on this!

gowthamrm:
@rwjblue Thank you, I have updated the test with the suggested changes, Please have a look.

rwjblue:
Thank you!

PR #15913, opened 2017-12-01
[BUGFIX] Ensure findBy works for key with dot notation.

sarathkumarbaskaran:
Ensure `Ember.findBy()` works for keys with dot. 

``` javascript
let arr = [
 { canada: [....] },
 { 'u.s.a.': [....] }
];
Ember.get(arr, 'u.s.a'); // returns `undefined`
```
Ember get() assumes this as a path, since the key has dot.

To solve this, we have two options, 
1. Revert the  `get(obj, key)` usage here.
2. Add additional validations to get method.

The second option is little complicated and it disturbs the default behavior of get method for path lookup.

### DIFF
--- a/packages/ember-metal/lib/property_get.js
+++ b/packages/ember-metal/lib/property_get.js
@@ -60,7 +60,7 @@ export function get(obj, keyName) {
 
   if (isDescriptor) {
     return value.get(obj, keyName);
-  } else if (isPath(keyName)) {
+  } else if (isPath(keyName) && !(keyName in obj)) {
     return _getPath(obj, keyName);
   } else if (value === undefined && 'object' === typeof obj && !(keyName in obj) &&
     typeof obj.unknownProperty === 'function') {
--- a/packages/ember-runtime/lib/mixins/enumerable.js
+++ b/packages/ember-runtime/lib/mixins/enumerable.js
@@ -47,10 +47,12 @@ function pushCtx(ctx) {
 
 function iter(key, value) {
   let valueProvided = arguments.length === 2;
-
-  return valueProvided ?
-    (item)=> value === get(item, key) :
-    (item)=> !!get(item, key);
+  let i = (item => {
+    let itemVal = item[key];
+    let cur = itemVal === undefined ? get(item, key) : itemVal;
+    return valueProvided ? value === cur : !!cur;
+  });
+  return i;
 }
 
 /**
--- a/packages/ember-runtime/tests/suites/enumerable/find.js
+++ b/packages/ember-runtime/tests/suites/enumerable/find.js
@@ -102,4 +102,33 @@ suite.test('should return first undefined property match', function() {
   equal(obj.findBy('bar', undefined), ary[1], 'findBy(\'bar\', undefined)');
 });
 
+suite.test('findBy keyName with dot and as path', function() {
+  let obj = {};
+  let arrWithPath = [
+    { id: 1, foo: { bar: true } },
+    { id: 2, foo: { bar: false } }
+  ];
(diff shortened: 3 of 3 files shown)


packages/ember-runtime/lib/mixins/enumerable.js
```diff
@@ -49,8 +49,8 @@ function iter(key, value) {
   let valueProvided = arguments.length === 2;
 
   return valueProvided ?
-    (item)=> value === get(item, key) :
-    (item)=> !!get(item, key);
+    (item)=> value === item[key] :
```
rwjblue: Hmm, this breaks `item[key]` being a computed property doesn't it?

sarathkumarbaskaran: @rwjblue 

> being a computed property doesn't it?

considering the array, a computed property will not affect the behavior. 

``` javascript
/* normal array */
arr.findBy('u.s.a.'); // returns undefined. Same applicable for filterBy, isEvery and isAny. 

/* `arr` a computed property */
this.get(arr).findBy('u.s.a.'); // returns undefined
```
correct me, if i misunderstood the use case.

> Ember.get(arr, 'u.s.a'); // returns `undefined`

This was a falsy example mentioned in the description.

rwjblue: Sorry, I’m saying that the property we are finding by is a computed property in the items. Using Ember.get today supports that (but is also the thing that allows nested path lookup), and I believe this change would break those use cases. 


For example:

```
let a = [
  { id: 1, foo: { bar: true } },
  { id: 2, foo: { bar: false } },
  { id: 3, foo: { bar: true } },
];

a.findBy(‘foo.bar’);
```

sarathkumarbaskaran: @rwjblue  Thanks for pointing out. You were right. 👍 

I have handled that use case. Please check.
Handled for both `enumerable` and  `get prop`.

packages/ember-metal/lib/property_get.js
```diff
@@ -61,7 +61,11 @@ export function get(obj, keyName) {
   if (isDescriptor) {
     return value.get(obj, keyName);
   } else if (isPath(keyName)) {
```
bekzod: I think `isPath(keyName) && !(keyName in obj)` would be enough here

bekzod:
IMHO adds a bit overhead to common case usage of `get` and `findBy`, unless `keys with dots` usage is really common among ember users :P I wouldn't add this feature

packages/ember-metal/lib/property_get.js
sarathkumarbaskaran: Good catch 👍

sarathkumarbaskaran:
@bekzod the concern is, we should not deviate from the native behavior of JS. 
I still remember, Ember is aligned with the native and/or logic of JS, later.

IMO, this does not seem to be an overhead. Actually, this handling reduces the overhead of all enumerables, except for computed property.

sarathkumarbaskaran:
@rwjblue could you spare some time here?

sarathkumarbaskaran:
@rwjblue please chime in, if you're free.

mmun:
@sarathkumarbaskaran I spoke with @rwjblue and we decided that we will not accept this feature as we feel the benefits do not outweigh the great increase in complexity to the internals, e.g. is `x.get('a.b.c.d')` equivalent to `x['a.b.c.d']`, `x['a.b.c'].d`, `x['a.b']['c.d']`, `x.a['b.c.d']`, etc?

Issue #15876, opened 2017-11-29
Ember 3.0 Related Cleanup


rwjblue:
- [x] Remove `phantomjs` usage from CI (tracking in https://github.com/emberjs/ember.js/pull/15877)
- [x] Remove IE9 and IE10 from CI runs. (tracking in https://github.com/emberjs/ember.js/pull/15879)
- [x] Leverage native WeakMaps :lock: @sivakumar-kailasam https://github.com/emberjs/ember.js/pull/15878
  - Remove `packages/ember-utils/lib/weak-map-utils.js`
  - Remove non-WeakMap fallback in `packages/ember-metal/lib/meta.js` [here](https://github.com/emberjs/ember.js/blob/fbc5d562487d6c10568cc9d3c04bcc507ce3c0d0/packages/ember-metal/lib/meta.js#L466-L496)
  - Remove non-WeakMap fallbacks in `packages/ember-metal/lib/transaction.js`
  - Remove non-WeakMap fallback in `packages/ember-glimmer/lib/utils/references.ts` [here](https://github.com/emberjs/ember.js/blob/fbc5d562487d6c10568cc9d3c04bcc507ce3c0d0/packages/ember-glimmer/lib/utils/references.ts#L53)
  - Remove `packages/ember-metal/lib/weak_map.js`
- [ ] Remove custom `Map` from `packages/ember-metal/lib/map.js` (being tracked in https://github.com/emberjs/ember.js/issues/13815) :lock: @maxwerr
- [x] Remove `console` polyfills / shims in `packages/ember-console/lib/index.js` https://github.com/emberjs/ember.js/pull/15881
- [x] `phantomjs` support / hacks https://github.com/emberjs/ember.js/pull/15880
  - [x] Remove `isPhantom` flag from `packages/ember-environment/lib/index.js`
  - [x] Remove workarounds for phantoms broken define property in `packages/ember-metal/lib/properties.js` ([here](https://github.com/emberjs/ember.js/blob/a92a7cd003f7202a397aef2bbe95397ffe483034/packages/ember-metal/lib/properties.js#L24-L39), [here](https://github.com/emberjs/ember.js/blob/a92a7cd003f7202a397aef2bbe95397ffe483034/packages/ember-metal/lib/properties.js#L225-L229), and [here](https://github.com/emberjs/ember.js/blob/a92a7cd003f7202a397aef2bbe95397ffe483034/packages/ember-metal/lib/properties.js#L183-L185))
[...]

PR #15914, opened 2017-12-01
[CLEANUP] Remove _actions functionality

thoov:
Along with related deprecations, asserts, and tests.

Related: [ember-2-legacy PR](https://github.com/emberjs/ember-2-legacy/pull/17)

Apart of: #15876

### DIFF
--- a/packages/ember-runtime/lib/mixins/action_handler.js
+++ b/packages/ember-runtime/lib/mixins/action_handler.js
@@ -210,9 +210,7 @@ const ActionHandler = Mixin.create({
       );
       target.send(...arguments);
     }
-  },
-
-  willMergeMixin() {}
+  }
 });
 
 export default ActionHandler;
--- a/packages/ember-runtime/lib/mixins/action_handler.js
+++ b/packages/ember-runtime/lib/mixins/action_handler.js
@@ -212,38 +212,7 @@ const ActionHandler = Mixin.create({
     }
   },
 
-  willMergeMixin(props) {
-    assert('Specifying `_actions` and `actions` in the same mixin is not supported.', !props.actions || !props._actions);
-
-    if (props._actions) {
-      deprecate(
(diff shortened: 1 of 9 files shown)


thoov:
Tests have been ported to the legacy addon

packages/ember-runtime/lib/mixins/action_handler.js
```diff
@@ -212,38 +212,7 @@ const ActionHandler = Mixin.create({
...
-        'Specifying actions in `_actions` is deprecated, please use `actions` instead.',
-        false,
-        { id: 'ember-runtime.action-handler-_actions', until: '3.0.0' }
-      );
-
-      props.actions = props._actions;
-      delete props._actions;
-    }
-  }
+  willMergeMixin() {}
```
rwjblue: I think we can remove the implementation completely?

rwjblue (commented):
Looking good! 

One inline tweak (we can remove the whole `willMergeMixin` function in the action handler mixin, and can you also add a comment above https://github.com/emberjs/ember.js/blob/d074b82205ad1c7b2e84e29cfc3d081f5cc2ae7c/packages/ember-metal/lib/mixin.js#L241

to make it clear that this can be removed after 3.4?

packages/ember-runtime/lib/mixins/action_handler.js
thoov: Will this have issues with https://github.com/emberjs/data/blob/master/addon/-private/system/model/model.js#L1889

rwjblue: No, removing this empty function shouldn’t affect that (we aren’t removing the functionality yet)...

thoov:
@rwjblue should be good now

rwjblue:
Awesome, thank you!!

PR #15915, opened 2017-12-01
[CLEANUP] Remove {{render}}

thoov:
Related: [ember-2-legacy PR](https://github.com/emberjs/ember-2-legacy/pull/21)

Apart of: #15876

### DIFF
--- a/packages/ember-glimmer/lib/syntax.ts
+++ b/packages/ember-glimmer/lib/syntax.ts
@@ -65,7 +65,11 @@ export function registerMacros(macro: any) {
 export function populateMacros(blocks: any, inlines: any) {
   inlines.add('outlet', outletMacro);
   inlines.add('component', inlineComponentMacro);
-  inlines.add('render', renderMacro);
+
+  if (ENV._ENABLE_RENDER_SUPPORT === true) {
+    inlines.add('render', renderMacro);
+  }
+
   inlines.add('mount', mountMacro);
   inlines.add('input', inputMacro);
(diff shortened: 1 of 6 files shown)


thoov:
@rwjblue I don't have a great way of getting rid of these other files other than putting a comment on them for future removal. Do you have any ideas?

Right now I'm just making sure that we don't include render is what we want just not sure if we can clean this up any more

packages/ember-glimmer/lib/syntax.ts
```diff
@@ -65,7 +65,11 @@ export function registerMacros(macro: any) {
 export function populateMacros(blocks: any, inlines: any) {
   inlines.add('outlet', outletMacro);
   inlines.add('component', inlineComponentMacro);
-  inlines.add('render', renderMacro);
+
+  if (ENV._ENABLE_RENDER_SUPPORT === true) {
```
rwjblue: I think this `ENV` is not imported (thats what one of the test failures indicated at least)..

rwjblue (commented):
The approach seems good to me, I don't think there is much more we can do to reduce payload size while still supporting what we need...

packages/ember-glimmer/lib/syntax.ts
rwjblue: @thoov - I think fixing this will fix CI.

thoov: @rwjblue Yep that fixed it. Should be good to go now

rwjblue (approved):

PR #15917, opened 2017-12-03
move `watchedEvents` to dev environment

bekzod:
since `watchedEvents` is unlikely to be used in production environment moved it to dev


bekzod:
maybe it is better not to export it in production mode :P

bekzod:
it is completely removed with #16162

mmun:
Hey @bekzod, I think the idea here is good and the idea definitely crossed my mind. These debugging abilities might come in useful some day (maybe? :P). But I want to do it in a way that has a completely zero footprint if possible. Anyways, just something to think about. And sorry for sweeping the rug out from under you.

bekzod:
Hi @mmun not at all, I think what you did was great 👍

PR #15922, opened 2017-12-04
[CLEANUP] Remove legacy controller proxy behavior

thoov:
Apart of: #15876

### DIFF
--- a/packages/ember-runtime/lib/mixins/-proxy.js
+++ b/packages/ember-runtime/lib/mixins/-proxy.js
@@ -105,11 +105,6 @@ export default Mixin.create({
   unknownProperty(key) {
     let content = get(this, 'content');
     if (content) {
-      deprecate(
-        `You attempted to access \`${key}\` from \`${this}\`, but object proxying is deprecated. Please use \`model.${key}\` instead.`,
-        !this.isController,
-        { id: 'ember-runtime.controller-proxy', until: '3.0.0' }
-      );
       return get(content, key);
     }
   },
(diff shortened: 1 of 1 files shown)


rwjblue (approved):
Awesome, thank you!

packages/ember-runtime/lib/mixins/-proxy.js
```diff
@@ -105,11 +105,6 @@ export default Mixin.create({
   unknownProperty(key) {
     let content = get(this, 'content');
     if (content) {
-      deprecate(
-        `You attempted to access \`${key}\` from \`${this}\`, but object proxying is deprecated. Please use \`model.${key}\` instead.`,
-        !this.isController,
-        { id: 'ember-runtime.controller-proxy', until: '3.0.0' }
-      );
       return get(content, key);
     }
```
Turbo87: @rwjblue shouldn't we remove the remaining code too instead of removing just the deprecation warning?

thoov: @Turbo87 What do you mean by remaining code?

Turbo87: the `return get(content, key);` 🤔

rwjblue: No, that code supports proxies generally. The deprecation was _only_ for when a Ember.Controller base class mixed in the proxy mixin (which is how Ember.ObjectController and Ember.ArrayController worked). Those two classes have been removed, and this mixin is private so there was just no need for the deprecation for when included in controllers...

Turbo87: I see, thanks :)

PR #15923, opened 2017-12-05
[CLEANUP] Remove didInitAttrs lifecycle method

thoov:
Apart of: #15876 

Related: [ember-2-legacy PR](https://github.com/emberjs/ember-2-legacy/pull/23/files)

### DIFF
--- a/packages/ember-views/lib/mixins/view_support.js
+++ b/packages/ember-views/lib/mixins/view_support.js
@@ -1,7 +1,7 @@
 import { guidFor, getOwner } from 'ember-utils';
 import { descriptor, Mixin } from 'ember-metal';
 import { assert, deprecate } from 'ember-debug';
-import { environment } from 'ember-environment';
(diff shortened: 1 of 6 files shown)


packages/ember-views/lib/mixins/view_support.js
```diff
@@ -70,7 +70,6 @@ export default Mixin.create({
   concatenatedProperties: ['attributeBindings'],
 
   [POST_INIT]() {
-    this.trigger('didInitAttrs');
```
rwjblue: I think this needs to be `ENV` wrapped, otherwise users in the addon will not ever have their `didInitAttrs` get invoked...

thoov: @rwjblue Oops forgot about that. You are right

thoov:
@rwjblue addressed your comment

rwjblue:
Awesome, thank you!

PR #15924, opened 2017-12-05
[BUGFIX beta] assert that `classNameBinding` items are strings

### LINKED ISSUE #15851: classNameBindings with ,, compiles but causes runtime problems
A double ,, in a classNameBinding will compile ok, but generates a runtime error afterwards.

e.g. 

 classNameBindings: ['isVertical:vertical:horizontal',,'isVisible:visible:hidden'],

produces:

curly.js:406 Uncaught TypeError: Cannot read property 'split' of undefined
    at curly.js:406
    at processComponentInitializationAssertions (curly.js:411)
    at CurlyComponentManager.create (curly.js:208)
    at runtime.js:1671
    at AppendOpcodes.evaluate (runtime.js:70)
    at VM.next (runtime.js:7246)
    at TemplateIterator.next (runtime.js:7317)
    at RootState.render (renderer.js:62)
    at exports.runInTransaction (ember-metal.js:652)
    at InteractiveRenderer._renderRoots (renderer.js:341)

using Ember 2.16.2.

GavinJoyce:
fixes https://github.com/emberjs/ember.js/issues/15851

(this was part of an Ember open source workshop in Intercom)

### DIFF
--- a/packages/ember-glimmer/lib/component-managers/curly.ts
+++ b/packages/ember-glimmer/lib/component-managers/curly.ts
@@ -412,7 +412,19 @@ export function validatePositionalParameters(named: NamedArguments, positional: 
 }
 
 export function processComponentInitializationAssertions(component: Component, props: any) {
-  assert(`classNameBindings must not have spaces in them: ${component.toString()}`, (() => {
+  assert(`classNameBindings must be strings: ${component}`, (() => {
+    let { classNameBindings } = component;
+    for (let i = 0; i < classNameBindings.length; i++) {
+      let binding = classNameBindings[i];
+
+      if (typeof binding !== 'string') {
+        return false;
+      }
+    }
+    return true;
+  })());
+
+  assert(`classNameBindings must not have spaces in them: ${component}`, (() => {
     let { classNameBindings } = component;
     for (let i = 0; i < classNameBindings.length; i++) {
       let binding = classNameBindings[i];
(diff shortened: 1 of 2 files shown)


packages/ember-glimmer/lib/component-managers/curly.ts
```diff
@@ -412,6 +412,18 @@ export function validatePositionalParameters(named: NamedArguments, positional:
 }
 
 export function processComponentInitializationAssertions(component: Component, props: any) {
+  assert(`classNameBindings must be strings: ${component.toString()}`, (() => {
```
rwjblue: Seems fine, but I would have assumed that the explicit `.toString()` shouldn't have been needed inside template strings.  Mind confirming?

packages/ember-glimmer/lib/component-managers/curly.ts
```diff
@@ -412,6 +412,18 @@ export function validatePositionalParameters(named: NamedArguments, positional:
...
+    let { classNameBindings } = component;
+    for (let i = 0; i < classNameBindings.length; i++) {
+      let binding = classNameBindings[i];
+
+      if (typeof(binding) !== 'string') {
```
rwjblue: I think we generally prefer `typeof binding !== 'string'`

rwjblue (commented):
Minor stylistic nit-picks, but overall looks good to me.

packages/ember-glimmer/lib/component-managers/curly.ts
GavinJoyce: Yep, that's the case. I've updated the other similar tests to remove `toString` and use template strings too

GavinJoyce:
@rwjblue thanks for the feedback, PR has been updated

rwjblue (approved):
You rock, thank you!!

PR #15916, opened 2017-12-02
[Cleanup] Remove orphaned outlets

lorcan:
Addresses https://github.com/emberjs/ember.js/issues/15876#issuecomment-348680323

PR #15900, opened 2017-11-30
[CLEANUP] Change warn deprecations to asserts

thoov:
Apart of #15876 

Related: [ember-2-legacy PR](https://github.com/emberjs/ember-2-legacy/pull/9)

PR #15925, opened 2017-12-05
[Cleanup] Remove orphaned outlets

lorcan:
Addresses https://github.com/emberjs/ember.js/pull/15916#issuecomment-349135327. Best reviewed [with whitespace off](https://github.com/emberjs/ember.js/pull/15925/files?w=1)

The ember-2-legacy plugin provides the environment variable `_ENABLE_ORPHANED_OUTLETS_SUPPORT ` with https://github.com/emberjs/ember-2-legacy/pull/24.

Later if/when we decide to remove the code associated with this deprecation we can use https://github.com/emberjs/ember.js/pull/15916 to get an idea of what's needed.

### DIFF
--- a/packages/ember-routing/lib/system/router.js
+++ b/packages/ember-routing/lib/system/router.js
@@ -1488,12 +1488,11 @@ function appendLiveRoute(liveRoutes, defaultParentState, renderOptions) {
     set(target.outlets, renderOptions.outlet, myState);
   } else {
     if (renderOptions.into) {
-      let orphanedOutletsDeprecation = `Rendering into a {{render}} helper that resolves to an {{outlet}} is deprecated.`;
       if (ENV._ENABLE_ORPHANED_OUTLETS_SUPPORT !== true) {
-        assert(orphanedOutletsDeprecation);
+        assert(`Rendering into a {{render}} helper that resolves to an {{outlet}} is no longer supported.`);
       } else {
         deprecate(
-          orphanedOutletsDeprecation,
+          `Rendering into a {{render}} helper that resolves to an {{outlet}} is deprecated.`,
           false,
           {
             id: 'ember-routing.top-level-render-helper',
--- a/packages/ember-routing/lib/system/router.js
+++ b/packages/ember-routing/lib/system/router.js
@@ -36,6 +36,7 @@ import {
 } from '../utils';
 import RouterState from './router_state';
 import { DEBUG } from 'ember-env-flags';
+import { ENV } from 'ember-environment';
 
 /**
 @module @ember/routing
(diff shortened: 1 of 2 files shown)


packages/ember-routing/lib/system/router.js
```diff
@@ -1492,23 +1493,28 @@ function appendLiveRoute(liveRoutes, defaultParentState, renderOptions) {
...
-        false,
-        {
-          id: 'ember-routing.top-level-render-helper',
-          until: '3.0.0',
-          url: 'https://emberjs.com/deprecations/v2.x/#toc_rendering-into-a-render-helper-that-resolves-to-an-outlet'
-        }
-      );
+      let orphanedOutletsDeprecation = `Rendering into a {{render}} helper that resolves to an {{outlet}} is deprecated.`;
+      if (ENV._ENABLE_ORPHANED_OUTLETS_SUPPORT !== true) {
+        assert(orphanedOutletsDeprecation);
```
lorcan: Is it reasonable to reuse the deprecation warning in the assertion?

thoov: We should change it from `is deprecated` to`is no longer supported` in the assert case. This does bring up an interesting question as the `{{render}}` helper is going away completely so I don't know if this case will ever be hit post 3.0.

packages/ember/tests/routing/basic_test.js
```diff
@@ -1931,8 +1932,24 @@ QUnit.test('Route should tear down multiple outlets', function() {
...
+      this.render({ into: 'nonexistent' });
+    }
+  });
+
+  expectAssertion(() => bootApplication(), 'Rendering into a {{render}} helper that resolves to an {{outlet}} is deprecated.');
+});
+
 
 QUnit.test('Route will assert if you try to explicitly render {into: ...} a missing template', function () {
+  ENV._ENABLE_ORPHANED_OUTLETS_SUPPORT = true;
```
thoov: Can we reset this value in the teardown like this (https://github.com/emberjs/ember.js/pull/15900/files#diff-7c1b00291152c89da72c8940b7305c1a)

thoov:
Looks good other than the 2 minor feedback items

packages/ember/tests/routing/basic_test.js
lorcan: Good call! Sorry I missed that 👍

lorcan:
This should be ready to go now :+1:

thoov:
@lorcan Looks good to me. @rwjblue r?

lorcan:
I think this is good to go. Is there anything I can do to improve it? cc @rwjblue

rwjblue:
Damn, I seem to have missed this one! I am very sorry!

@lorcan - Can you rebase and fix the conflict in `packages/ember/tests/routing/basic_test.js`?

packages/ember/tests/routing/basic_test.js
```diff
@@ -89,6 +89,7 @@ QUnit.module('Basic Routing', {
 
       originalLoggerError = Logger.error;
       originalRenderSupport = ENV._ENABLE_RENDER_SUPPORT;
+      originalEnabledOrphanedOutletSupport = ENV._ENABLE_ORPHANED_OUTLETS_SUPPORT;
 
       ENV._ENABLE_RENDER_SUPPORT = true;
```
lorcan: Now I'm wondering if we should set `ENV._ENABLE_ORPHANED_OUTLETS_SUPPORT` to true like we do for `ENV._ENABLE_RENDER_SUPPORT` here?

lorcan:
:+1: I've rebased that now, thanks @rwjblue

packages/ember/tests/routing/basic_test.js
thoov: No,

By setting it to `true` we hit the old deprecation flow for all test which is what we are moving away from (and we want all of the tests to go through the new assertion flow). Currently we have just the test below set to `true` so only that one test verifies that the "old" path still works.

PR #15928, opened 2017-12-06
refactor `ObserverSet` use Map

bekzod:


packages/ember-metal/lib/observer_set.js
```diff
@@ -1,66 +1,37 @@
-import { guidFor } from 'ember-utils';
-import { sendEvent } from './events';
 
-/*
```
rwjblue: This comment was useful (in general) for explaining the internal data structure used in `ObserverSet`.  Can you bring back a new comment that shows what things look like now?

.eslintrc.js
```diff
@@ -28,6 +28,7 @@ module.exports = {
     'setInterval': true,
     'clearInterval': true,
 
+    'Map': true,
```
rwjblue: 👍

rwjblue (changes requested):

bekzod:
already covered here #16187

PR #15931, opened 2017-12-06
utilize `Map` and `Set` in `property_events`

bekzod:


mmun:
The idea here seems good. Can you rebase?

packages/ember-metal/lib/property_events.js
```diff
@@ -115,15 +115,16 @@ function propertyDidChange(obj, keyName, _meta) {
...
 // called whenever a property is about to change to clear the cache of any dependent keys (and notify those properties of changes, etc...)
 function dependentKeysWillChange(obj, depKey, meta) {
   if (meta.isSourceDestroying() || !meta.hasDeps(depKey)) { return; }
   let seen = WILL_SEEN;
-  let top = !seen;
+  let top = seen === null;
 
   if (top) {
-    seen = WILL_SEEN = {};
+    seen = WILL_SEEN = new Map();
```
mmun: What do you think of having a single `Map` instance and .clear()ing it here?

packages/ember-metal/lib/property_events.js
```diff
@@ -115,40 +115,36 @@ function notifyPropertyChange(obj, keyName, _meta) {
...
 
-let DID_SEEN;
+let DID_SEEN = new Map();
 
 // called whenever a property has just changed to update dependent keys
 function dependentKeysDidChange(obj, depKey, meta) {
   if (meta.isSourceDestroying() || !meta.hasDeps(depKey)) { return; }
   let seen = DID_SEEN;
-  let top = !seen;
+  let top = seen.length === 0;
```
mmun: should be `size`?

packages/ember-metal/lib/property_events.js
```diff
@@ -115,40 +115,36 @@ function notifyPropertyChange(obj, keyName, _meta) {
...
   if (top) {
-    seen = DID_SEEN = {};
+    seen = DID_SEEN;
   }
 
   iterDeps(notifyPropertyChange, obj, depKey, seen, meta);
 
   if (top) {
-    DID_SEEN = null;
+    DID_SEEN.clear();
```
mmun: I'm not totally sure if it is correct to share the instance because `dependentKeysDidChange` can be called recursively. If you're not totally sure yourself we should probably change the logic back to how it was.

rwjblue (approved):

PR #15935, opened 2017-12-06
[BUGFIX lts] blueprints: fix framework detection to work with prerelease versions of ember-cli-mocha

simonihmig:
`semver.satisfies` does not cover prerelease versions like `*.*.*-beta` as required, so use `gte`

cc @Turbo87

### DIFF
--- a/blueprints/test-framework-detector.js
+++ b/blueprints/test-framework-detector.js
@@ -27,7 +27,7 @@ module.exports = function(blueprint) {
 
     } else if ('ember-cli-mocha' in dependencies) {
       let checker = new VersionChecker(this.project);
-      if (fs.existsSync(this.path + '/mocha-0.12-files') && checker.for('ember-cli-mocha', 'npm').satisfies('>=0.12.0')) {
+      if (fs.existsSync(this.path + '/mocha-0.12-files') && checker.for('ember-cli-mocha', 'npm').gte('0.12.0')) {
         type = 'mocha-0.12';
       } else {
         type = 'mocha';


Turbo87:
@rwjblue I think we can make this `[BUGFIX lts]` too, right?

simonihmig:
BTW, it just came to my mind that these last two changes (support `ember-mocha`, `ember-qunit` and beta versions) here are currently not covered by tests. I can maybe add some in a separate PR? 

But just wondered how. I don't think it makes sense to add them for each and every blueprint test case, making the number of permutations explode. But maybe just something like a smoke test, by adding those cases to one blueprint, e.g. `component-test`?

rwjblue:
> @rwjblue I think we can make this [BUGFIX lts] too, right?

Ya, I think so

rwjblue (approved):

simonihmig:
done! [`lts`]

rwjblue:
> BTW, it just came to my mind that these last two changes (support ember-mocha, ember-qunit and beta versions) here are currently not covered by tests. I can maybe add some in a separate PR?
>
> But just wondered how.

I would probably split things up and unit test the specific logic.

Something like:

```js
function detectType(dependencies, root) {
  if ('ember-qunit' in dependencies || 'ember-cli-qunit' in dependencies) {
    type = 'qunit';

  } else if ('ember-mocha' in dependencies) {
    type = 'mocha-0.12';

    var checker = new VersionChecker({ root });
    if (fs.existsSync(this.path + '/mocha-0.12-files') && checker.for('ember-cli-mocha', 'npm').gte('0.12.0')) {
      type = 'mocha-0.12';
    } else {
      type = 'mocha';
    }

  } else {
    this.ui.writeLine('Couldn\'t determine test style - using QUnit');
    type = 'qunit';
  }

  return type;
}

module.exports = function(blueprint) {
blueprint.supportsAddon = function() {
    return false;
  };

  blueprint.filesPath = function() {
    var dependencies = this.project.dependencies();
    var type = detectType(dependencies, this.project.root);

    return path.join(this.path, type + '-files');
  };

  return blueprint;
};

module.exports._detectType = detectType;
```


This way you can pass any number of different permutations along with something like `broccoli-test-helper` (like I test [ember-cli-version-checker itself](https://github.com/ember-cli/ember-cli-version-checker/blob/master/tests/index.js#L111-L120))...

rwjblue (approved):

rwjblue:
Sorry for the rebase issues @simonihmig, thanks for the quick turn around.

Issue #15933, opened 2017-12-06
QUnit Blueprints Quest Issue


alexander-alvarez:
__Background__
We need to implement changes in blueprints and docs realted to [RFC 232] and [RFC 268].
(Some background for those unfamiliar http://rwjblue.com/2017/10/23/ember-qunit-simplication/)

As of `ember-cli-qunit@^4.1.1` (which is in the default blueprint as of `ember-cli@2.17`), we can implement the "Simplified QUnit" style tests in all tests except for acceptance tests.
In a future release of `ember-qunit` (likely 3.2.x), we will be able to leverage the implementation of `setupApplicationTest` from [RFC 268] for use in acceptance-tests. Once these changes are implemented upstream, we must implement the blueprint updates.
Currently, we have no way to generate these test files but we can run the [`ember-qunit-codemod`] to migrate test files and continue to iterate on the final output in the blueprint specific pull requests.

The desired behavior when generating testing blueprints is to, generate "Simplified QUnit" style tests if supported by the host app. We have several tasks below to extend the blueprints to support the "Simplified QUnit" style blueprints alongside the legacy, and ember-mocha blueprints. We can likely add to the behavior of [`test-framework-detector.js`] to support this other extension. Once these changes are completed we also need to update the guides to be consistent with the current `ember-cli` codebase. We should work with @rwjblue and the learning team (cc @locks) to determine how we best teach this.

__Tasks__
- [x] (🔒 @Turbo87) Extend  [`test-framework-detector.js`] to support a directory called `qunit-rfc-232-files` which will be where the "Simplified QUnit" blueprints will live in each of the blueprint test folders (e.g `blueprints/component-test/qunit-rfc-232-files/`). The decision to use the `qunit-rfc-232-files` directory should be based on the host's `ember-cli-qunit`
- [x] Add `<type>-test` "simplified QUnit" blueprints
    - [x] acceptance-test -  https://github.com/emberjs/ember.js/issues/15999
    - [x] (🔒 @Turbo87) component-test
    - [x]  (🔒 @snewcomer) controller-test    
    - [x] (🔒 @snewcomer) helper-test
    - [x] (🔒 @snewcomer) initializer-test
    - [x] (🔒 @snewcomer) instance-initializer-test
    - [x] (🔒 @boyanyordanov) mixin-test
    - [x] (:lock: @rwwagner90) route-test
    - [x] (🔒 @boyanyordanov) service-test
    - [x] (🔒 @snewcomer) util-test
[...]

PR #15937, opened 2017-12-07
Helper test

snewcomer:
Ref #15933 

- [x] add blueprints
~~- [x] fixture helper tests (including mocha unit tests)~~ recommended to do in a separate PR
- [x] node-tests blueprint tests

### DIFF
--- a/node-tests/blueprints/helper-test-test.js
+++ b/node-tests/blueprints/helper-test-test.js
@@ -34,6 +34,26 @@ describe('Blueprint: helper-test', function() {
       });
     });
 
+    describe('with ember-cli-qunit@4.1.1', function() {
+      beforeEach(function() {
+        generateFakePackageManifest('ember-cli-qunit', '4.1.1');
+      });
+
+      it('helper-test foo/bar-baz', function() {
+        return emberGenerateDestroy(['helper-test', 'foo/bar-baz'], _file => {
+          expect(_file('tests/integration/helpers/foo/bar-baz-test.js'))
+            .to.equal(fixture('helper-test/rfc232.js'));
+        });
+      });
+
+      it('helper-test foo/bar-baz --unit', function() {
+        return emberGenerateDestroy(['helper-test', 'foo/bar-baz', '--unit'], _file => {
+          expect(_file('tests/unit/helpers/foo/bar-baz-test.js'))
+            .to.equal(fixture('helper-test/rfc232-unit.js'));
+        });
+      });
+    });
+
     describe('with ember-cli-mocha@0.11.0', function() {
       beforeEach(function() {
         modifyPackages([
--- a/node-tests/fixtures/helper-test/rfc232-unit.js
+++ b/node-tests/fixtures/helper-test/rfc232-unit.js
@@ -1,0 +1,13 @@
+import { fooBarBaz } from 'my-app/helpers/foo/bar-baz';
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+
+module('Unit | Helper | foo/bar baz', function(hooks) {
+  setupTest(hooks);
+
+  // Replace this with your real tests.
+  test('it works', function(assert) {
+    let result = fooBarBaz([42]);
+    assert.ok(result);
+  });
+});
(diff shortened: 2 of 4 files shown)


rwjblue:
Will need a rebase on top of the changes from https://github.com/emberjs/ember.js/pull/15938 and  https://github.com/emberjs/ember.js/pull/15941

node-tests/fixtures/helper-test/rfc232-unit.js
```diff
@@ -0,0 +1,12 @@
+import { fooBarBaz } from 'my-app/helpers/foo/bar-baz';
```
Turbo87: @rwjblue should we look this up on the container instead?

node-tests/fixtures/helper-test/rfc232.js
```diff
@@ -0,0 +1,16 @@
...
+
+module('Integration | Helper | foo/bar-baz', function(hooks) {
+  setupRenderingTest(hooks);
+
+  test('it renders', async function(assert) {
+    this.set('inputValue', '1234');
+
+    await render(hbs`{{foo/bar-baz inputValue}}`);
+
+    assert.equal(this.$().text().trim(), '1234');
```
Turbo87: please use `this.element.textContent.trim()` instead to avoid jQuery

node-tests/fixtures/helper-test/rfc232-unit.js
alexander-alvarez: https://github.com/rwjblue/ember-qunit-codemod/blob/master/__testfixtures__/ember-qunit-codemod/native-qunit-to-nested.output.js would suggest no
but I guess in the above scenarios we aren't really testing the helper itself, just the underlying function.

I have found it kind of hard to test class based helpers -- for example:
```js
this.render(hbs`{{my-helper data}}`);
```
may return objects that don't serialize well.


Would be great if we could do something like
```js
this.render(hbs`{{let actualResult (my-helper data)}}` );
assert.equal(this.actualResult, expectedResult);
```

Turbo87: If you're returning something other than a string or number you should probably not use a rendering test and use a "unit" test instead. In any case both tests should probably use the container. Rendering tests do that implicitly and unit tests explicitly.

see also https://github.com/rwjblue/ember-qunit-codemod/issues/17

node-tests/blueprints/helper-test-test.js
```diff
@@ -34,6 +34,26 @@ describe('Blueprint: helper-test', function() {
...
+          expect(_file('tests/integration/helpers/foo/bar-baz-test.js'))
+            .to.equal(fixture('helper-test/rfc232.js'));
+        });
+      });
+
+      it('helper-test foo/bar-baz --unit', function() {
+        return emberGenerateDestroy(['helper-test', 'foo/bar-baz', '--unit'], _file => {
+          expect(_file('tests/unit/helpers/foo/bar-baz-test.js'))
+            .to.equal(fixture('helper-test/rfc232-unit.js'));
+        });
```
snewcomer: I've found the default as `integration` to be confusing for helpers.  But that's probably just me though :man_shrugging:

node-tests/fixtures/helper-test/rfc232-unit.js
snewcomer: @Turbo87 should I change the unit test to use the container then using `factoryFor` then?

node-tests/blueprints/helper-test-test.js
rwjblue: Helpers are only logically used from a template, testing them outside of the template layer should be a fairly uncommon thing to do...

node-tests/fixtures/helper-test/rfc232-unit.js
rwjblue: Not right now, lets focus on 1:1 for now then we can follow up with a PR that adds usage of `owner.factoryFor('helper:foo').create()`.

rwjblue: I've discussed this at length with @krisselden and I believe he is opposed to the public blueprints instantiating helpers outside of the confines of the template layer.  I personally disagree, but I don't want to make a change and "sneak it in" without a more thorough discussion.  Does that make sense?

rwjblue (approved):

rwjblue:
Thank you @snewcomer!

PR #15943, opened 2017-12-07
blueprints/controller-test: Add RFC232 variants

snewcomer:
Progress on #15933

### DIFF
--- a/blueprints/controller-test/qunit-rfc-232-files/tests/unit/__path__/__test__.js
+++ b/blueprints/controller-test/qunit-rfc-232-files/tests/unit/__path__/__test__.js
@@ -6,7 +6,7 @@ module('<%= friendlyTestDescription %>', function(hooks) {
 
   // Replace this with your real tests.
   test('it exists', function(assert) {
-    let controller = this.owner.factoryFor('controller:<%= controllerPathName %>').create();
+    let controller = this.owner.lookup('controller:<%= controllerPathName %>');
     assert.ok(controller);
   });
 });
--- a/blueprints/controller-test/qunit-rfc-232-files/tests/unit/__path__/__test__.js
+++ b/blueprints/controller-test/qunit-rfc-232-files/tests/unit/__path__/__test__.js
@@ -1,0 +1,12 @@
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+
+module('<%= friendlyTestDescription %>', function(hooks) {
+  setupTest(hooks);
+
+  // Replace this with your real tests.
+  test('it exists', function(assert) {
+    let controller = this.owner.factoryFor('controller:<%= controllerPathName %>').create();
+    assert.ok(controller);
+  });
+});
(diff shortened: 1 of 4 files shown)


blueprints/controller-test/qunit-rfc-232-files/tests/unit/__path__/__test__.js
```diff
@@ -0,0 +1,12 @@
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+
+module('<%= friendlyTestDescription %>', function(hooks) {
+  setupTest(hooks);
```
snewcomer: Where do `needs` go?  Should we indicate how those can be added to the registry or is this not a thing anymore?

node-tests/fixtures/controller-test/rfc232-unit.js
```diff
@@ -0,0 +1,12 @@
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+
```
snewcomer: since only a unit test is generated, should this be named `rfc232-default` or is the current name fine?

blueprints/controller-test/qunit-rfc-232-files/tests/unit/__path__/__test__.js
alexander-alvarez: gloriously not a thing anymore.

to stub out lookups (if need be) we use `this.owner.register` -- example:
```js
this.owner.register('service:feature', Service.extend({
    myFeature: false
});
```
probably don't need to document that here though

node-tests/fixtures/controller-test/rfc232-unit.js
alexander-alvarez: as is seems fine to me, it's not lying -- it is a unit test.

alexander-alvarez (approved):
Thanks for hopping in to help @snewcomer

node-tests/fixtures/controller-test/rfc232-unit.js
Turbo87: > it is a unit test

is it though... :trollface: 

I'd just use `rfc232.js` in this case, but it doesn't matter much.

Turbo87 (approved):

blueprints/controller-test/qunit-rfc-232-files/tests/unit/__path__/__test__.js
```diff
@@ -0,0 +1,12 @@
...
+
+  // Replace this with your real tests.
+  test('it exists', function(assert) {
+    let controller = this.owner.factoryFor('controller:<%= controllerPathName %>').create();
```
rwjblue: Can you tweak this to use `this.owner.lookup(‘controller:foo’);`? Controllers are intended to be singletons, so we don’t have to `factoryFor(...).create()`..

rwjblue (changes requested):

rwjblue (approved):

rwjblue (approved):

PR #15945, opened 2017-12-07
blueprints/instance-initializers-test: Add RFC232 variants

snewcomer:
Progress on #15933

### DIFF
--- a/blueprints/instance-initializer-test/qunit-rfc-232-files/tests/unit/instance-initializers/__name__-test.js
+++ b/blueprints/instance-initializer-test/qunit-rfc-232-files/tests/unit/instance-initializers/__name__-test.js
@@ -1,7 +1,6 @@
 import Application from '@ember/application';
-import { run } from '@ember/runloop';
 
-import { initialize } from '<%= dasherizedModulePrefix %>/initializers/<%= dasherizedModuleName %>';
+import { initialize } from '<%= dasherizedModulePrefix %>/instance-initializers/<%= dasherizedModuleName %>';
 import { module, test } from 'qunit';
 import { setupTest } from 'ember-qunit';
 import destroyApp from '../../helpers/destroy-app';
@@ -10,20 +9,23 @@ module('<%= friendlyTestName %>', function(hooks) {
   setupTest(hooks);
 
   hooks.beforeEach(function() {
-    run(() => {
-      this.application = Application.create();
-      this.application.deferReadiness();
+    this.TestApplication = Application.extend();
+    this.TestApplication.instanceInitializer({
+      name: 'initializer under test',
+      initialize
     });
+    this.application = this.TestApplication.create({ autoboot: false });
+    this.instance = this.application.buildInstance();
   });
   hooks.afterEach(function() {
     destroyApp(this.application);
+    destroyApp(this.instance);
   });
 
   // Replace this with your real tests.
-  test('it works', function(assert) {
-    initialize(this.application);
+  test('it works', async function(assert) {
+    await this.instance.boot();
 
-    // you would normally confirm the results of the initializer here
     assert.ok(true);
   });
 });
--- a/blueprints/instance-initializer-test/qunit-rfc-232-files/tests/unit/instance-initializers/__name__-test.js
+++ b/blueprints/instance-initializer-test/qunit-rfc-232-files/tests/unit/instance-initializers/__name__-test.js
@@ -1,0 +1,29 @@
+import Application from '@ember/application';
+import { run } from '@ember/runloop';
+
+import { initialize } from '<%= dasherizedModulePrefix %>/initializers/<%= dasherizedModuleName %>';
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+import destroyApp from '../../helpers/destroy-app';
+
+module('<%= friendlyTestName %>', function(hooks) {
+  setupTest(hooks);
+
+  hooks.beforeEach(function() {
+    run(() => {
+      this.application = Application.create();
+      this.application.deferReadiness();
+    });
+  });
+  hooks.afterEach(function() {
+    destroyApp(this.application);
+  });
+
+  // Replace this with your real tests.
+  test('it works', function(assert) {
+    initialize(this.application);
+
+    // you would normally confirm the results of the initializer here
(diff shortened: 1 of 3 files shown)


blueprints/instance-initializer-test/qunit-rfc-232-files/tests/unit/instance-initializers/__name__-test.js
```diff
@@ -0,0 +1,29 @@
...
+      this.application.deferReadiness();
+    });
+  });
+  hooks.afterEach(function() {
+    destroyApp(this.application);
+  });
+
+  // Replace this with your real tests.
+  test('it works', function(assert) {
+    initialize(this.application);
```
rwjblue: Instance initializers expect to receive an Ember.ApplicationInstance instance, but this is an Ember.Application instance.

blueprints/instance-initializer-test/qunit-rfc-232-files/tests/unit/instance-initializers/__name__-test.js
```diff
@@ -0,0 +1,29 @@
...
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+import destroyApp from '../../helpers/destroy-app';
+
+module('<%= friendlyTestName %>', function(hooks) {
+  setupTest(hooks);
+
+  hooks.beforeEach(function() {
+    run(() => {
+      this.application = Application.create();
```
rwjblue: Can you swap this to `Application.create({ autoboot: false })` here?

blueprints/instance-initializer-test/qunit-rfc-232-files/tests/unit/instance-initializers/__name__-test.js
```diff
@@ -0,0 +1,29 @@
...
+      this.application.deferReadiness();
```
rwjblue: This shouldn’t be needed with my suggested changes

blueprints/instance-initializer-test/qunit-rfc-232-files/tests/unit/instance-initializers/__name__-test.js
```diff
@@ -0,0 +1,29 @@
...
+import { initialize } from '<%= dasherizedModulePrefix %>/initializers/<%= dasherizedModuleName %>';
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+import destroyApp from '../../helpers/destroy-app';
+
+module('<%= friendlyTestName %>', function(hooks) {
+  setupTest(hooks);
+
+  hooks.beforeEach(function() {
+    run(() => {
```
rwjblue: We shouldn’t need to `Ember.run` here, why is this required? Is there an error or something without it?

blueprints/instance-initializer-test/qunit-rfc-232-files/tests/unit/instance-initializers/__name__-test.js
rwjblue: To fix that, you would do something like:

```
module(‘asdf’, function(hooks) {

  hooks.beforeEach(function() {
    this.Application = Ember.Application.extend();
    this.application = this.Application.create({ autoboot: true });
    this.application.instanceInitializer({
      name: ‘initializer under test’,
      initialize
    });
  });

  hooks.afterEach(function() {
    runDestroy(this.instance);
    runDestroy(this.application);
  });

  test(‘it works’, async function(assert) {
    this.instance = this.application.buildInstance();
    await this.instance.boot();
    assert.ok(didTheInitializerWork);
  });
});
```

rwjblue (changes requested):
I’d like to restructure things a bit, I left an example in line would you mind giving that snippet a whirl and confirming that it works properly? I believe it should (barring any typos etc)

blueprints/instance-initializer-test/qunit-rfc-232-files/tests/unit/instance-initializers/__name__-test.js
snewcomer: @rwjblue Ah yeah overall my bad.  So with this test, `await this.instance.boot();` results in `Assertion Failed: You cannot use the same root element (BODY) multiple times in an Ember.Application`.  Failing on this line - https://github.com/emberjs/ember.js/blob/5f4ac602813d8fbd78f26fc955acf693e0b3d62a/packages/ember-views/lib/system/event_dispatcher.js#L179

blueprints/instance-initializer-test/qunit-rfc-232-files/tests/unit/instance-initializers/__name__-test.js
snewcomer: Yeah the error I get is the `You have turned on testing mode...` error.

blueprints/instance-initializer-test/qunit-rfc-232-files/tests/unit/instance-initializers/__name__-test.js
snewcomer: tested on a 2.17 application.

rwjblue: Would you mind pushing up your demo app so I can tinker with it to find the right combo?

rwjblue: Also, I made a really bad mistake in my comment above, it should be `{ autoboot: false }` (and I had `true` in the comment)

rwjblue: OK, just tested locally, and I think that it will work once updating to `{ autoboot: false }` (sorry about that).

* [example instance-initializer test](https://github.com/rwjblue/__emberjs-PR15945_demo/blob/master/tests/unit/instance-initializers/do-stuff-test.js)
* [diff of changes needed](https://github.com/rwjblue/__emberjs-PR15945_demo/commit/30a36e61691bf4c13ff0321726dc927b9aa93686)

rwjblue:
Awesome, thank you!

PR #15946, opened 2017-12-08
blueprints/initializer-test: Add RFC232 variants

snewcomer:
Progress on #15933

### DIFF
--- a/blueprints/initializer-test/qunit-rfc-232-files/tests/unit/initializers/__name__-test.js
+++ b/blueprints/initializer-test/qunit-rfc-232-files/tests/unit/initializers/__name__-test.js
@@ -1,5 +1,4 @@
 import Application from '@ember/application';
-import { run } from '@ember/runloop';
 
 import { initialize } from '<%= dasherizedModulePrefix %>/initializers/<%= dasherizedModuleName %>';
 import { module, test } from 'qunit';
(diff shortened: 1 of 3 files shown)


blueprints/initializer-test/qunit-rfc-232-files/tests/unit/initializers/__name__-test.js
```diff
@@ -0,0 +1,29 @@
...
+import { initialize } from '<%= dasherizedModulePrefix %>/initializers/<%= dasherizedModuleName %>';
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+import destroyApp from '../../helpers/destroy-app';
+
+module('<%= friendlyTestName %>', function(hooks) {
+  setupTest(hooks);
+
+  hooks.beforeEach(function() {
+    run(() => {
```
rwjblue: Similar to the conversation in the instance-initializer PR, lets tweak this to avoid the manual run loop usage and leverage only public APIs.

I believe the result here would be something like:

```js
module('Unit | Initializer | other stuff', function(hooks) {
  hooks.beforeEach(function() {
    this.TestApplication = Application.extend();
    this.TestApplication.initializer({
      name: 'other stuff',
      initialize
    });

    this.application = this.TestApplication.create({ autoboot: false });
  });

  hooks.afterEach(function() {
    destroyApp(this.application);
  });

  test('it works', async function(assert) {
    await this.application.boot();

    assert.equal(self.bar, true);
  });
});
```

See https://github.com/rwjblue/__emberjs-PR15945_demo/blob/master/tests/unit/initializers/other-stuff-test.js for a demo app...

rwjblue (changes requested):
Thanks for working on this! I'm sorry for the run around, but I think we are narrowing in on a really nice future for these types of tests.

rwjblue:
Sorry for the run around on this. Thank you for working through it!

PR #15947, opened 2017-12-08
blueprints/util-test: Add RFC232 variants

snewcomer:
Progress on #15933

### DIFF
--- a/blueprints/util-test/qunit-rfc-232-files/tests/unit/utils/__name__-test.js
+++ b/blueprints/util-test/qunit-rfc-232-files/tests/unit/utils/__name__-test.js
@@ -1,0 +1,13 @@
+import <%= camelizedModuleName %> from '<%= dasherizedModulePrefix %>/utils/<%= dasherizedModuleName %>';
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+
+module('<%= friendlyTestName %>', function(hooks) {
+  setupTest(hooks);
+
+  // Replace this with your real tests.
+  test('it works', function(assert) {
+    let result = <%= camelizedModuleName %>();
+    assert.ok(result);
+  });
+});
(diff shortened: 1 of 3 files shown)


alexander-alvarez (approved):

node-tests/fixtures/util-test/rfc232-unit.js
```diff
@@ -0,0 +1,13 @@
...
+
+module('Unit | Utility | foo bar', function(hooks) {
+  setupTest(hooks);
+
+  // Replace this with your real tests.
+  test('it works', function(assert) {
+    let result = fooBar();
+    assert.ok(result);
+  });
+});
```
Turbo87: the files should end with a `\n`

Turbo87 (changes requested):

blueprints/util-test/qunit-rfc-232-files/tests/unit/utils/__name__-test.js
```diff
@@ -0,0 +1,13 @@
+import <%= camelizedModuleName %> from '<%= dasherizedModulePrefix %>/utils/<%= dasherizedModuleName %>';
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+
+module('<%= friendlyTestName %>', function(hooks) {
+  setupTest(hooks);
```
rwjblue: Hmm. In general I think that we probably don’t need this? I would expect utils to invoke the util method and test its inputs and outputs. Let’s remove `setupTest` here for now?

rwjblue (changes requested):

blueprints/util-test/qunit-rfc-232-files/tests/unit/utils/__name__-test.js
Turbo87: agreed 👍

rwjblue:
Thank you!!

PR #15889, opened 2017-11-29
[CLEANUP] Remove Enumerable#contains

cibernox:
Part of https://github.com/emberjs/ember.js/pull/15876

PR #15948, opened 2017-12-08
[CLEANUP] remove ArrayMixin#contains

bekzod:
not sure if this should have been removed part of https://github.com/emberjs/ember.js/pull/15889


rwjblue (approved):

thoov:
@rwjblue Just confirming that this does't need a polyfill correct? Since we have one for Enumerable.contains and this inherits from Enumerable.

rwjblue:
@thoov - Right, that was my thought also.

PR #15950, opened 2017-12-08
blueprints/mixin-test: Added RFC-232 variant

boyanyordanov:
Progress for #15933 

A couple of things I am not sure about: 

- I wasn't sure whether the `setupTest` call is needed for mixing test (no setup was required before) and the codemod from https://github.com/rwjblue/ember-qunit-codemod didn't add it. I added it anyway, it looks logical after reading the RFC a couple of times. 

- and also there's an addon version of this blueprint. Should there also be an updated version for rfc232 for that as well ? 

P.S. Sorry If the questions are stupid. I am still learning how all this works under the hood.

### DIFF
--- a/node-tests/fixtures/mixin-test/rfc232.js
+++ b/node-tests/fixtures/mixin-test/rfc232.js
@@ -1,0 +1,12 @@
+import EmberObject from '@ember/object';
+import FooMixin from 'my-app/mixins/foo';
+import { module, test } from 'qunit';
+
+module('Unit | Mixin | foo', function() {
+  // Replace this with your real tests.
+  test('it works', function (assert) {
+    let FooObject = EmberObject.extend(FooMixin);
+    let subject = FooObject.create();
+    assert.ok(subject);
+  });
+});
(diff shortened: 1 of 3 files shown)


node-tests/fixtures/mixin-test/rfc232.js
```diff
@@ -0,0 +1,15 @@
+import EmberObject from '@ember/object';
+import FooMixin from 'my-app/mixins/foo';
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+
+module('Unit | Mixin | foo', function(hooks) {
+  setupTest(hooks);
```
Turbo87: I don't think we necessarily need the `setupTest()` call here since we're not using the container for anything 🤔

boyanyordanov: I will remove it.

boyanyordanov: One more question. Should I squash the commits after that, ince it's a small change reverting some of this ?

Turbo87: yes, I think that would be good, thanks! :)

Turbo87:
> also there's an addon version of this blueprint. Should there also be an updated version for rfc232 for that as well ?

I guess you are talking about the `addon.js` fixture file, right? I don't think we necessarily need an rfc232 version of that since the only thing that changes is the import path. but feel free to add it if you think that it might be useful :)

boyanyordanov:
I got confused a bit. Yeah, I was talking about the fixture. Looking at it now it doesn't seem that necessary to add another test just for the path replacement there.

Thanks a lot for the help!

boyanyordanov:
Done. Please let me know if additional changes are needed.

Turbo87 (approved):

rwjblue:
Awesome, thank you!

PR #15759, opened 2017-10-24
Typed ember glimmer

krisselden:
Fixes #15729

PR #15780, opened 2017-10-28
deprecate `disabledWhen` in `link-to`

bekzod:
just to align `link-to` with other components

PR #15952, opened 2017-12-08
[BUGFIX beta] fix regression of clicking link-to with disabled=true

simonihmig:
#15759 introduced a regression, allowing clicking on a `{{#link-to "somewhere" disabled=true}}` to still invoke the transition. This was because the default value of `_isDisabled` was moved to the init hook here: https://github.com/emberjs/ember.js/commit/8c68f5b26f50af456af214065d3b411006d420da#diff-49415576dd428811b5854523cc9b9adcR528, effectively overriding any previously set value. 

The added test covers this, and was failing before the change here.

### DIFF
--- a/packages/ember-glimmer/lib/components/link-to.ts
+++ b/packages/ember-glimmer/lib/components/link-to.ts
@@ -515,7 +515,6 @@ const LinkComponent = EmberComponent.extend({
   */
   init() {
     this._super(...arguments);
-    this._isDisabled = this._isDisabled || false;
 
     // Map desired event name to invoke function
     let eventName = get(this, 'eventName');
@@ -534,10 +533,14 @@ const LinkComponent = EmberComponent.extend({
   */
   disabled: computed({
     get(_key: string): boolean {
+      // always returns false for `get` because (due to the `set` just below)
+      // the cached return value from the set will prevent this getter from _ever_
+      // being called after a set has occured
       return false;
     },
+
     set(_key: string, value: any): boolean {
-      if (value !== undefined) { this.set('_isDisabled', value); }
+      this._isDisabled = value;
 
       return value ? get(this, 'disabledClass') : false;
     },
@@ -636,7 +639,7 @@ const LinkComponent = EmberComponent.extend({
 
     if (get(this, 'bubbles') === false) { event.stopPropagation(); }
 
-    if (get(this, '_isDisabled')) { return false; }
+    if (this._isDisabled) { return false; }
 
     if (get(this, 'loading')) {
       // tslint:disable-next-line:max-line-length
--- a/packages/ember-glimmer/lib/components/link-to.ts
+++ b/packages/ember-glimmer/lib/components/link-to.ts
@@ -515,7 +515,7 @@ const LinkComponent = EmberComponent.extend({
   */
   init() {
     this._super(...arguments);
-    this._isDisabled = false;
+    this._isDisabled = this._isDisabled || false;
 
     // Map desired event name to invoke function
     let eventName = get(this, 'eventName');
--- a/packages/ember-glimmer/lib/components/link-to.ts
+++ b/packages/ember-glimmer/lib/components/link-to.ts
@@ -72,25 +72,9 @@
   {{/link-to}}
   ```
 
-  any passed value to `disabled` will disable it except `undefined`.
-  to ensure that only `true` disable the `link-to` component you can
-  override the global behavior of `LinkComponent`.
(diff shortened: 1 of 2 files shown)


bekzod:
I think `disabled` is private in `link-to`, and you are supposed to used `disabledWhen` instead and https://github.com/emberjs/ember.js/pull/15780 supposed to fix it

simonihmig:
Hm, not sure... 
* even if it's private, it is supposed to work correctly, and all versions of Ember <= 2.17 do support the usage I fixed here, only beta and canary started to [fail](https://travis-ci.org/kaliber5/ember-bootstrap/builds/313485762)...
* the documentation regarding this is actually a bit misleading (cc @locks), given that there is the API docs of the `LinkComponent`, but also the `link-to` helper docs (*which are one and the same thing nowadays, right?*):
  * the [API docs](https://emberjs.com/api/ember/2.17/classes/LinkComponent?show=inherited%2Cprotected%2Cprivate) of the `LinkComponent` mark `disabled` as private, but do not mention `disabledWhen` at all!
  * the `link-to` [helper docs](https://emberjs.com/api/ember/2.17/classes/Ember.Templates.helpers/methods/link-to?anchor=link-to) mention both `disabled` (though private!?) and `disabledWhen`
* given that 👆, even if `disabled` is indeed private and mentioned by mistake in the docs, I would consider it an "intimate API" (that's how you call it, right?), i.e. should only be removed with a prior deprecation

So I would say either remove it all together (no-go IMHO), or fix the bug!?

rwjblue:
Agreed. We should fix this. Even if we decide later to deprecate.

simonihmig:
Anything missing here? /cc @rwjblue

rwjblue:
@simonihmig - Just updated the implementation here, mind taking another look?

rwjblue (approved):

packages/ember-glimmer/lib/components/link-to.ts
```diff
@@ -534,10 +518,14 @@ const LinkComponent = EmberComponent.extend({
   */
   disabled: computed({
     get(_key: string): boolean {
+      // always returns false for `get` because (due to the `set` just below)
+      // the cached return value from the set will prevent this getter from _ever_
+      // being called after a set has occured
       return false;
     },
+
     set(_key: string, value: any): boolean {
```
simonihmig: Unrelated to the changes here, but the typing seems wrong. Shouldn't it be `set(...): string|false`?

packages/ember-glimmer/lib/components/link-to.ts
```diff
@@ -72,25 +72,9 @@
   {{/link-to}}
   ```
 
-  any passed value to `disabled` will disable it except `undefined`.
-  to ensure that only `true` disable the `link-to` component you can
-  override the global behavior of `LinkComponent`.
+  any truthy value passed to `disabled` will disable it except `undefined`.
```
simonihmig: "except `undefined`" can be removed I think, as `undefined` is not truthy anyways?

packages/ember-glimmer/lib/components/link-to.ts
```diff
@@ -534,10 +518,14 @@ const LinkComponent = EmberComponent.extend({
...
     get(_key: string): boolean {
+      // always returns false for `get` because (due to the `set` just below)
+      // the cached return value from the set will prevent this getter from _ever_
+      // being called after a set has occured
       return false;
     },
+
     set(_key: string, value: any): boolean {
-      if (value !== undefined) { this.set('_isDisabled', value); }
+      this._isDisabled = value;
```
simonihmig: The changes here some right to me in general, but they do change the semantics a bit, don't they (the truthy vs. not undefined)? Could they break something for somebody?

simonihmig:
@rwjblue Uh, already merged? :) I added a few comments...

packages/ember-glimmer/lib/components/link-to.ts
rwjblue: @simonihmig - Yes, this should be typed as a `string | null` I believe.

packages/ember-glimmer/lib/components/link-to.ts
rwjblue: GAH! Thank you

packages/ember-glimmer/lib/components/link-to.ts
rwjblue: @simonihmig if you take a look at where `this._isDisabled` is actually checked (in `_invoke`), we have been doing a truthy check there for a while so changing this to always update it just makes it simpler...

PR #15953, opened 2017-12-08
fix typo

bekzod:


packages/ember-glimmer/lib/component-managers/curly.ts
```diff
@@ -96,7 +96,7 @@ function applyAttributeBindings(element: Simple.Element, attributeBindings: any,
 function tagName(vm: VM) {
   let dynamicScope: DynamicScope = vm.dynamicScope() as DynamicScope;
   // tslint:disable-next-line:no-shadowed-variable
-  let { tagName } = dynamicScope.view!;
+  let { tagName } = dynamicScope.view;
```
locks: I don't think this is a typo: https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#non-null-assertion-operator.
Can you confirm @krisselden?

locks (changes requested):

bekzod:
At least will be interesting to know what it does :)

packages/ember-glimmer/lib/component-managers/curly.ts
rwjblue: Yes, this is a typescript "thing". It essentially instructs the typescript compiler that the `dynamicScope.view` property could not possibly be undefined.

PR #15958, opened 2017-12-08
[docs release]Remove static denotation from computed prop methods

toddjordan:
This was a regression in api docs between 2.15 and 2.16

Fixes https://github.com/ember-learn/ember-api-docs/issues/405

Computed property functions were erroneously marked static, putting them in a whole different place in the UI.


rwjblue:
Thanks @toddjordan!

PR #15961, opened 2017-12-09
store dirtyable tag of helper in meta

bekzod:
just to bring consistency storing dirtyable tag in helper's meta

### DIFF
--- a/packages/ember-glimmer/lib/helper.ts
+++ b/packages/ember-glimmer/lib/helper.ts
@@ -5,9 +5,8 @@
 import { Dict, Opaque } from '@glimmer/util';
 import { DirtyableTag } from '@glimmer/reference';
 import { FrameworkObject } from 'ember-runtime';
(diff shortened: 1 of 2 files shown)


chancancode:
Those tags are not really the same thing. The `tagFor(helperInstance)` is invalidated whenever you do `helperInstance.set(...)`. If no one cares (i.e. if the object was never used as a dependency in templates) then the object would not have had a tag and the invalidation would have been a no-op.

The `[RECOMPUTE_TAG]` is to support the explicit invalidation API in class-based helpers (`this.recompute()`), similar to `this.rerender()` in curly components.

Perhaps because the helper instance aren't typically used in the templates directly, this will happen to work, but that would be more of a hack than brining consistency IMO.

bekzod:
thank you for clarifying, at least I know about the context now

chancancode:
👍

Issue #15939, opened 2017-12-07
Set-Only Computed property - behaviour mismatch


Exelord:
# Description
Hey, I noticed that in the case when we are creating `set-only` property and we want to access unexistent `get` we are getting an error `Cannot read property 'call' of undefined`. This is reasonable but how the user can know that is a set-only property instead of just assume that there is a bug. In case of `readOnly` we are notifying the user when he tries to access a setter that this property is read-only.

# Real use-case - problem
We want to override `set` function on our property. We are not forced to define a `get` (we don't know that this is required- not mentioned anywhere):
```js
sword: Ember.computed({
    set(key, value) {
      // some overrides here
      return value;
    },
})
```
and then we are just getting into the problem eg. during inspecting an object in ember-inspector or calling a `get` on it. 

We are just getting the error: `Cannot read property 'call' of undefined`

# Solution
Shouldn't we have sth similar to `readOnly` and inform the user that this property doesn't have a defined `getter` or it is a `writeOnly`?

In my opinion, in case of custom set-only properties, we should just return an `undefined` during the `getter` call and create analogically to `readOnly` the `writeOnly` which will rise an error in case of trying to get the `getter`.

What do think guys?

# Full Issue example
```js
let Lord = Ember.Object.extend({
  sword: Ember.computed({
    set(key, value) {
      return value;
    },
  })
});

let exe = Lord.create();

exe.get('sword'); // returns error ' Cannot read property 'call' of undefined'
```

Serabe:
Thank you!

bekzod:
@Serabe I think this https://github.com/emberjs/ember.js/pull/15962 was first :)

rwjblue:
Closed by https://github.com/emberjs/ember.js/pull/15962

PR #15962, opened 2017-12-09
Allow to define only setter in computed property

Exelord:
This PR will fix issue: #15939 

Before when we tried to access undefined getter we got into this problem:
screen shot 2017-12-09 at 11 32 24

### DIFF
--- a/packages/ember-metal/lib/computed.js
+++ b/packages/ember-metal/lib/computed.js
@@ -21,6 +21,8 @@ import {
 
 const DEEP_EACH_REGEX = /\.@each\.[^.]+\./;
 
+function noop() {}
+
 /**
   A computed property transforms an object literal with object's accessor function(s) into a property.
 
@@ -138,11 +140,12 @@ class ComputedProperty extends Descriptor {
       this._getter = config;
     } else {
       assert('computed expects a function or an object as last argument.', typeof config === 'object' && !Array.isArray(config));
-      assert('Config object passed to computed can only contain `get` or `set` keys.', Object.keys(config).every((key)=> key === 'get' || key === 'set'));
-      this._getter = config.get;
+      assert('Config object passed to computed can only contain `get` and `set` keys.', Object.keys(config).every((key)=> key === 'get' || key === 'set'));
+      assert('Computed properties must receive a getter or a setter, you passed none.', !!config.get || !!config.set);
+      this._getter = config.get || noop;
       this._setter = config.set;
     }
-    assert('Computed properties must receive a getter or a setter, you passed none.', !!this._getter || !!this._setter);
+
     this._suspended = undefined;
     this._meta = undefined;
     this._volatile = false;
--- a/packages/ember-metal/tests/computed_test.js
+++ b/packages/ember-metal/tests/computed_test.js
@@ -41,7 +41,7 @@ QUnit.test('computed properties defined with an object only allow `get` and `set
       set() {},
       other() {}
     });
-  }, 'Config object passed to computed can only contain `get` or `set` keys.');
+  }, 'Config object passed to computed can only contain `get` and `set` keys.');
 });
 
 if (EMBER_METAL_ES5_GETTERS) {
@@ -652,6 +652,20 @@ QUnit.test('setter can be omited', function(assert) {
   assert.ok(testObj.get('a') === '1');
   testObj.set('aInt', '123');
   assert.ok(testObj.get('aInt') === '123', 'cp has been updated too');
+});
+
+QUnit.test('getter can be omited', function(assert) {
+  let testObj = EmberObject.extend({
+    com: computed({
+      set(key, value) {
+        return value;
+      }
+    })
+  }).create();
+
+  assert.ok(testObj.get('com') === undefined);
+  testObj.set('com', '123');
+  assert.ok(testObj.get('com') === '123', 'cp has been updated');
 });
 
 QUnit.test('the return value of the setter gets cached', function(assert) {


packages/ember-metal/lib/computed.js
```diff
@@ -138,11 +138,12 @@ function ComputedProperty(config, opts) {
     this._getter = config;
   } else {
     assert('computed expects a function or an object as last argument.', typeof config === 'object' && !Array.isArray(config));
-    assert('Config object passed to computed can only contain `get` or `set` keys.', Object.keys(config).every((key)=> key === 'get' || key === 'set'));
-    this._getter = config.get;
+    assert('Config object passed to computed can only contain `get` and `set` keys.', Object.keys(config).every((key)=> key === 'get' || key === 'set'));
+    this._getter = config.get || function() {};
```
bekzod: I think it will be cleaner to put empty function outside of constructor, like `function K(){}` as it is done in other places, maybe it should also contain `warning` maybe that user hasn't passed getter this should be confirmed with core devs I guess

Exelord: Sure. I fix the problem in the source now. Could you take a look?
All remarks have been mentioned in the issue. Definitely worth to discuss.

packages/ember-metal/lib/computed.js
```diff
@@ -327,13 +328,17 @@ ComputedPropertyPrototype.didChange = function(obj, keyName) {
 
 ComputedPropertyPrototype.get = function(obj, keyName) {
   if (this._volatile) {
-    return this._getter.call(obj, keyName);
+    return this._getter ? this._getter.call(obj, keyName) : undefined;
```
bekzod: I don't think it is best solution, since `this._getter` won't be provided in super narrow case, and in common case it is provided, the best solution is to provide empty function as you did earlier

bekzod (changes requested):

packages/ember-metal/lib/computed.js
bekzod: `noop` function for `_getter` in constructor is best solution I think

Exelord: Well, it depends. If we use the `noop` function in the constructor we will allocate a memory even if we never call the `get`.

Right, now the solution is more complex but it fixes the source of the problem and saves resources. I noticed that when I look at `set` function definition. Now, they are implemented in a similar way.

Do you think this is a bad approach?

bekzod: > Well, it depends. If we use the noop function in the constructor we will allocate a memory even if we never call the get.

thats why I asked to keep `noop` function outside constructor ([example](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/renderer.ts#L184)), also not sure if it should contain warning inside `warn('getter missing .....')` this should be confirmed by core team I guess

> Right, now the solution is more complex but it fixes the source of the problem and saves resources. I noticed that when I looked at set function definition. Now, they are implemented in a similar way.

Not only it is complex, it is also adds extra overhead/processing to be done each time even `getter` is provided (which is 99% of time)

Thats why your first approach was best

bekzod: it is hotpath code therefore slightly more work will result in big overheads

Exelord: You suggested: `noop function for _getter in constructor is best solution I think` not, outside :p

Ok, will try that one. But isn't Ember.K already deprecated? Shouldn't we just use `() {}`?

bekzod: > You suggested: noop function for _getter in constructor is best solution I think not, outside :p
https://github.com/emberjs/ember.js/pull/15962#discussion_r155918993

> Ok, will try that one. But isn't Ember.K already deprecated? Shouldn't we just use () {}? 

global `Ember.K` yes deprecated, but using `noop` function  naming `K` is not :P, but calling it `noop` is more descriptive I think

bekzod (approved):

rwjblue:
@Exelord - Thanks for working on this!  I totally agree that we should address the issue you raised in https://github.com/emberjs/ember.js/issues/15939, but I don't think this is the right path forward. Instead of making a setter only CP silently return `undefined`, we should make it error (essentially emulating the behavior of `.readOnly`).

Something like:

```js
ComputedPropertyPrototype.get = function(obj, keyName) {
  assert(`Cannot call '.get' to retrieve ${keyName} on ${obj}, it has to getter.', typeof this._getter === 'function');
  // ...snip...
};
```

What do you think?

Exelord:
Thanks, @rwjblue :)
So, I see here 2 issues:

## 1. Overriding setter
### Problem
You are not allowed to override a setter and have a default behavior of getter without defining it.
```js
  currentUser: computed({
    get() {}, // need to be defined 😞 | right now, it will return the set cached value as expected

    set(key, value) {
      this.get('notifier').configure({ payload: { person: value } });
      return value;
    }
  }),
```

### Solution
I think the current state of the PR is perfectly fixing that issue. Right now you can just have a default getter behavior and override only a setter to extend his role.

## 2. There is no possibility to have a setter only property
This issue is mentioned by you. I agree that we should raise an error in case when we are trying to reach a getter analogically like the `readOnly` does with the setter, but for this purpose, I would like to create sth like `writeOnly` cp which will cover that case exclusively. Otherwise, we will kill the first possibility to just extend the setter responsibility. What do you think?

rwjblue:
> You are not allowed to override a setter and have a default behavior of getter without defining it.

Can you unpack this a bit? What does this mean? If you try to define a setter only CP it throws an error? When?

Exelord:
Basically, if you would have sth like:
```js
currentUser: computed({
    set(key, value) {
      this.get('notifier').configure({ payload: { person: value } });
      return value;
    }
  }),
```
and set the value `this.set('currentUser', { name: 'me' })` everything will work but then if you would try to get it by `this.get('currentUser')` it will raise an error (the missing get function which wasn't predicted by the implementation). That's what this PR fixes. (test's scenario)

So, right now you can't override a setter without defining an empty getter because you will fall into that issue.

rwjblue:
> So, right now you can't override a setter without defining an empty getter because you will fall into that issue when you will try to use the getter on it.

Yes, this makes sense to me. There is no getter to call, this isn’t something we should silently normalize it is an actual error in the users program. Also, If you override a CP you need to override both portions (if you don’t need to customize the getter, you can have its implementation be `return this._super()`).

Having a setter only CP (in general) seems “fine”, but if someone tries to `.get` it then it should fire a helpful assertion stating that the computed properly has no getter.

To reiterate: 

1) we should support `.set` for a setter only CP
2) attempting to `.get` a setter only CP should provide a useful error (the current error is horribly unhelpful)

Exelord:
Hmm... I'm not talking here about set-only properties. I'm talking about extending the setter. Analogically to getter I can omit the setter to just define custom getter but I can not omit the getter to define custom setter:
```js
currentUser: computed({
    get() {
      return 'test';
    }
}),
```
In that case, the cp is not get-only, it just has a custom getter. The same case should allow defining custom setter.

> Also, If you override a CP you need to override both portions

According to the docs and the code `set` and `get` are not the required keys, they are just allowed and can occur also in `or` relation.

> (if you don’t need to customize the getter, you can have its implementation returns this._super())

The solution here is just to define an empty function - it returns undefined which means that I will use a cache for it (the default behavior). The `super` shouldn't work here as there isn't exist any parent of `get`. get is not defined by default which causes the problem. In contrast, to set if the set is not defined there is fired a custom behaviour with caching.

bekzod:
I think best solution is to have `noop` function with warning inside, 
```javascript
function getNoop(val) {
  warn('You are missing getter, and you called get .....')
  return val
}

//then in constructor 
 this._getter = config.get || getNoop;
 this._setter = config.set

```

Exelord:
If we add the warning here we should also warn when there is only getter passed. So basically we should require defining them both.

bekzod:
> ...should also warn when there is only getter passed. 

passing only `setter` is unusual thats why warning is needed, passing only `getter` is not unusual most of the time only `getter` passed since it is computed property :)

Exelord:
Overriding setters can be useful when would you like to avoid using observers. (not always applicable).
Maybe the case is not usual but it against the documentation and assertions:
> Config object passed to computed can only contain `get` or `set` keys.
> Computed properties must receive a getter or a setter, you passed none.

From the logical point of view the current behaviour is arguing the reality.

rwjblue:
I don’t see a conflict with the documentation at all. If you only need a getter: All works fine. If you only need a setter: All works fine. The problem here is that you are only passing a setter but then somehow expect the getter to have some behavior too.

Exelord:
If you pass only a getter you can still use a working setter, but if you pass only a setter you can't use a working getter.

> Config object passed to computed can only contain get or set keys.

That sentence means that you can pass `get` or `set` (in real, you can pass `get` AND `set`) but indeed you can pass [`get` OR (`get` AND `set`)]
which is arguing with: 
> Computed properties must receive a getter or a setter, you passed none.

`get` || `set` !== `get`|| (`get` && `set`)

Serabe:
IMHO, set only properties should return `undefined` when they are `get`ted.

In vanilla JS:

```js
class MyClass {
  set hola(value) {
    console.log(value);
  }
}

let obj = new MyClass();
obj.hola = "hi!"; // outputs "hi!"
console.log(obj.hola); // outputs undefined
```

As we are trying to get closer, and closer to plain JS behaviour, having `obj.get('hola')` differ from `obj.hola` in JS classes seems wrong to me.

Exelord:
@rwjblue any more thoughts on that? Is it still not relevant? :/

mmun:
I agree with @Serabe.

mmun:
I'm persuaded by the argument that it is more likely for a developer to accidentally crash the Ember Inspector by clicking on such a property, or for something like QUnit to accidentally throw an error when its trying to serialize the assertion message, than it is for a developer to accidentally `get` a `set`-only property in their application code.

I'm happy to merge this as is after we squash all the commits down.

mmun:
@rwjblue what do you think of @bekzod's idea to have a development-only warning? Seems like a win-win.

packages/ember-metal/lib/computed.js
```diff
@@ -21,6 +21,8 @@ import {
 
 const DEEP_EACH_REGEX = /\.@each\.[^.]+\./;
 
+function noop() { /* noop */ }
```
bekzod: maybe /* noop */ comment not needed

rwjblue:
Seems like a good compromise....

Exelord:
```js
class MyClass {
  set hola(value) {
    console.log(value);
  }
}
```

The native getter doesn't fire any warning if it's not defined.

For me, it's natural that I want to override only a setter and keep a default getter behavior. 
Set only property doesn't mean it's write-only as get only property doesn't mean is read-only.

mmun:
On second thought, the warning would be almost as annoying in the inspector / JSON.stringify case. Happy to merge as is once the tests pass.

PR #15964, opened 2017-12-10
[BUGFIX beta] assert that attributes passed to component don't have duplicate keys

pizza:
If we invoke a component as such:
```
{{foo-bar 
  foo="pizza"
  foo="burger
  bar="pepperoni"
}}
```
I think it'd be useful to raise an exception saying that some of the attributes passed have duplicate keys (`foo` in this case)


bekzod:
I think https://github.com/emberjs/ember.js/pull/15653 is doing same thing

pizza:
@bekzod Cool, thanks for letting me know 😄

PR #15965, opened 2017-12-11
remove binding param in `changeProperties`

bekzod:
just throwing this for discussion, with availability of `arrow functions` and `Function.prototype.bind` I think `binding` param could be removed from `changeProperties` which will improve common case usage were `binding` is not supplied

### DIFF
--- a/packages/ember-metal/lib/property_events.js
+++ b/packages/ember-metal/lib/property_events.js
@@ -235,13 +235,12 @@ function endPropertyChanges() {
 
   @method changeProperties
   @param {Function} callback
-  @param [binding]
   @private
 */
-function changeProperties(callback, binding) {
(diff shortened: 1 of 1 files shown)


krisselden:
This is private but it is exported on Ember (I don't remember why).  We need to search usages in the wild, otherwise we may have to add a deprecation.

bekzod:
should I add depreciation or leave it as is ?

rwjblue:
Need to check for usages on emberobserver.com/code-search and determine if it is used at all.

bekzod:
I have checked it before there was no usage, https://emberobserver.com/code-search?codeQuery=changeProperties

rwjblue (approved):

PR #15968, opened 2017-12-12
Implement named args RFC

chancancode:
* [x] RFC https://github.com/emberjs/rfcs/pull/276
  * [ ] RFC merged (not a blocker for this PR)
* [x] Implemented `{{@foo}}`
  * [x] Reserved `{{@args}}`, `{{@arguments}}`
  * [x] Reserved everything that does not start with a lowercase letter (such as `@0`, `@1`, `@$`, ...)
* [x] Has tests
  * [x] Added a feature to the new test harness to run some tests only when a feature is enabled/disabled
* [x] ~~Added a "TBD" deprecation feature that allows the code paths to be flagged without rolling the deprecations out to users immediately. This is an internal, undocumented feature that is not intended to be used by libraries and apps (yet, should go through an RFC at least), and the API/implementation are subject to change.~~ (#15970)
* [x] ~~Flagged `{{attrs.*}}` for future deprecation~~ (#15970)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,5 +65,4 @@ env:
     - TEST_SUITE=blueprints
     - TEST_SUITE=travis-browsers DISABLE_JSCS=true DISABLE_JSHINT=true
     - TEST_SUITE=sauce
-    - TEST_SUITE=allow-backtracking-rerender ALLOW_BACKTRACKING=true
     - TEST_SUITE=each-package-tests BUILD_TYPE=alpha PUBLISH=true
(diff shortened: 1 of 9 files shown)


packages/ember-debug/lib/deprecate.js
```diff
@@ -159,50 +159,53 @@ if (DEBUG) {
...
-      deprecate(
-        missingOptionsUntilDeprecation,
-        options && options.until,
-        {
-          id: 'ember-debug.deprecate-until-missing',
-          until: '3.0.0',
-          url: 'https://emberjs.com/deprecations/v2.x/#toc_ember-debug-function-options'
-        }
-      );
+    if (until !== 'TBD' || ENV._ENABLE_TBD_DEPRECATIONS === true) {
```
rwjblue: Just flagging this as needing a bit more discussion / thought before landing.

rwjblue (commented):
Looks great!  Just need to figure out the official path forward on the TBD stuff...

chancancode:
I removed the TBD/deprecations from this PR and opened #15970 for those

rwjblue (approved):

PR #15970, opened 2017-12-12
Add internal "TBD" deprecations feature

chancancode:
* [x] Added a "TBD" deprecation feature that allows the code paths to be flagged without rolling the deprecations out to users immediately. This is an internal, undocumented feature that is not intended to be used by libraries and apps (yet, should go through an RFC at least), and the API/implementation are subject to change.
* [x] Flagged `{{attrs.*}}` for future deprecation


Turbo87:
closing due to inactivity

PR #15972, opened 2017-12-13
Revert "[BUGFIX release] fix issue with unchaining ChainNodes"

krisselden:
Reverts emberjs/ember.js#15849

This is a serious bug and this doesn't actually fix the issue, some refactoring to ember-metal has caused overriding watch to issue unbalanced watch/unwatch calls.  It needs to be bisected.

I would like us to pause refactorings to ember-metal until this is properly root caused.


krisselden:
closing in favor of #15982

PR #15974, opened 2017-12-13
Implement template-only components RFC

chancancode:
[RFC](https://github.com/emberjs/rfcs/pull/278)


rwjblue (approved):
Looks good to me!  There are some follow-up items that we need to tackle as the RFC progresses (e.g. runtime flag, can we rely on ember-source being consumed as an addon, etc), but those should likely be addressed in a separate PR.

packages/ember-glimmer/lib/component-managers/template-only.ts
```diff
@@ -0,0 +1,44 @@
...
+import AbstractManager from './abstract';
+
+class TemplateOnlyComponentLayoutCompiler {
+  static id = 'template-only';
+
+  constructor(public template: WrappedTemplateFactory) {
+  }
+
+  compile(builder: any) {
+    // TODO: use fromLayout
```
chancancode: I think @krisselden told me `fromLayout` is broken on this version of Glimmer, so I didn't bother.

PR #15978, opened 2017-12-14
[docs release] Document run as a function, so that it shows in @ember/runloop

toddjordan:
Addresses https://github.com/ember-learn/ember-api-docs/issues/406

makes run a function under @ember/runloop as described in https://github.com/ember-cli/ember-rfc176-data/blob/master/mappings.json

image


locks:
Thank yyoouu :D

PR #15979, opened 2017-12-14
fix variable declaration regression 

bekzod:
`itemsKeyIsAtThis` used before its declared


rwjblue:
As evidenced by the passing tests the way it was _does_ work. I tested locally (without transpilation):

```
function foo() {
  return bar;
}
let bar = "lol";
foo()
```

I only share this because I was also surprised that this was the behavior. I would have thought the exact thing you are fixing would cause an error (or not work).

Anyways, the change seems good to me, just thought I’d share...

bekzod:
super logical javascript

PR #15980, opened 2017-12-14
Fix get tests

chancancode:
cc @smfoote


chancancode:
Some of the types are still wrong in there – could also use generics, etc; but those parts I mostly just carried over from the current code, should still be a strict improvement.

rwjblue (approved):

PR #15891, opened 2017-11-30
[CLEANUP] remove `Freezable`

bekzod:

PR #15985, opened 2017-12-14
[CLEANUP] Removing unused imports

thoov:
Removing the unused imports that can be removed from the Freezable
PR #15891.


locks:
😗 👌

PR #15986, opened 2017-12-15
Clean up test globals

locks:
Part of https://github.com/emberjs/ember.js/issues/15988.

### DIFF
--- a/packages/ember-application/tests/system/engine_initializers_test.js
+++ b/packages/ember-application/tests/system/engine_initializers_test.js
@@ -1,12 +1,13 @@
 import { run } from 'ember-metal';
 import Engine from '../../system/engine';
+import {
+  moduleFor,
+  AbstractTestCase as TestCase
+} from 'internal-test-helpers';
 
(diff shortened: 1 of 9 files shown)


rwjblue (approved):

packages/ember-application/tests/system/engine_initializers_test.js
```diff
@@ -20,37 +20,37 @@ QUnit.module('Ember.Engine initializers', {
   }
 });
 
-QUnit.test('initializers require proper \'name\' and \'initialize\' properties', function() {
+QUnit.test('initializers require proper \'name\' and \'initialize\' properties', function(assert) {
   MyEngine = Engine.extend();
 
-  expectAssertion(() => {
+  assert.expectAssertion(() => {
```
rwjblue: This is causing CI failures, `assert.expectAssertion` isn't "a thing". This should be changed to `expectAssertion`.

rwjblue (changes requested):

PR #15989, opened 2017-12-15
Introduce `descriptorFor`

chancancode:
This abstracts away the fact that CP descriptors are stored on the prototypes currently, which should make it easier to move them into the meta.

Step 1 of #15984

### DIFF
--- a/packages/ember-metal/lib/chains.js
+++ b/packages/ember-metal/lib/chains.js
@@ -1,5 +1,5 @@
 import { get } from './property_get';
-import { meta as metaFor, peekMeta } from './meta';
+import { descriptorFor, meta as metaFor, peekMeta } from './meta';
 import { watchKey, unwatchKey } from './watch_key';
 import { cacheFor } from './computed';
 
@@ -13,8 +13,9 @@ function isObject(obj) {
   return typeof obj === 'object' && obj !== null;
 }
 
-function isVolatile(obj) {
-  return !(isObject(obj) && obj.isDescriptor && obj._volatile === false);
+function isVolatile(obj, keyName, meta) {
+  let desc = descriptorFor(obj, keyName, meta);
+  return !(desc !== undefined && desc._volatile === false);
 }
 
 class ChainWatchers {
@@ -321,7 +322,7 @@ function lazyGet(obj, key) {
   }
 
   // Use `get` if the return value is an EachProxy or an uncacheable value.
-  if (isVolatile(obj[key])) {
+  if (isVolatile(obj, key, meta)) {
     return get(obj, key);
   // Otherwise attempt to get the cached value of the computed property
   } else {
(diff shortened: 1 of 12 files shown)


packages/ember-metal/lib/chains.js
```diff
@@ -13,8 +13,9 @@ function isObject(obj) {
   return typeof obj === 'object' && obj !== null;
 }
 
-function isVolatile(obj) {
-  return !(isObject(obj) && obj.isDescriptor && obj._volatile === false);
+function isVolatile(obj, keyName, meta) {
+  let desc = descriptorFor(obj, keyName, meta);
+  return !(desc !== undefined && desc._volatile === false);
```
chancancode: This seems a bit strange to me. I would expect `desc  !== undefined && desc._volatile !== false` (as opposed to what amounts to `desc === undefined || desc._volatile !== false` now, which I ported straight from the existing code).

In other words, this is checking "either the property is NOT a descriptor, or if it is, its `_volatile` flag is anything other than `false` (which would include `undefined` etc). I think in practice that just funnels a lot more things through `get` today, which is perhaps not a big deal?

I would like someone to check the math here. /cc @krisselden

packages/ember-runtime/lib/system/core_object.js
```diff
@@ -125,23 +126,27 @@ function makeCtor() {
...
-            let isDescriptor = baseValue !== null && typeof baseValue === 'object' && baseValue.isDescriptor;
+            let possibleDesc = descriptorFor(this, keyName, m);
+            let isDescriptor = possibleDesc !== undefined;
 
-            if (hasConcatenatedProps && concatenatedProperties.indexOf(keyName) > -1) {
-              if (baseValue) {
-                value = makeArray(baseValue).concat(value);
-              } else {
-                value = makeArray(value);
+            if (!isDescriptor) {
```
chancancode: The previous code doesn't do this guard, but it seems strange to put a CP descriptor into a concatenated/merged property. How do we expect these features to interact? Should we assert here?

packages/ember-metal/lib/property_set.js
```diff
@@ -49,12 +50,16 @@ export function set(obj, keyName, value, tolerant) {
...
   }
 
+  let possibleDesc = descriptorFor(obj, keyName);
+
+  if (possibleDesc !== undefined) { /* computed property */
+    possibleDesc.set(obj, keyName, value);
+    return value;
+  }
+
   let currentValue = obj[keyName];
```
chancancode: This would also access `obj[keyName]` twice (same problem as `property_get`, see the comment there). However, there were no failing tests for this. Does it matter? Should I add one and change this to use `isDescriptor` instead?

rwjblue (commented):
This looks great to me! I think we should work on getting a benchmark run with these changes. So we can land...

chancancode:
Ran a benchmark against master (see attached) using emberaddons.com. There is no change in performance.

[results.pdf](https://github.com/emberjs/ember.js/files/1564484/results.pdf)

Issue #15988, opened 2017-12-15
Unify internal testing styles


thoov:
Currently there are two different styles that tests are written. [An older style](https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/tests/alias_test.js) which uses a QUnit global to define the module and tests via `QUnit.module()` and `QUnit.test()`. The old style also uses globals for the assertion such as `ok`, `equal`, etc. [A newer style](https://github.com/emberjs/ember.js/blob/d3d3627554e437603f6311f94f75a024d3218caf/packages/ember/tests/view_instrumentation_test.js) which does not use QUnit globals is partially rolled out in some of the packages. This issue tracks the work needed to move all tests over to the newer style and the related work surrounding it.

A breakdown of how to complete the conversion:

1) QUnit.module becomes moduleFor

```js
// Old Style:
QUnit.module('<MODULE_NAME>');

// New Style:
import { moduleFor, AbstractTestCase } from 'internal-test-helpers';

moduleFor('<MODULE_NAME>', class extends AbstractTestCase {
  //...
});
```

Note: There are a few test case classes already so check [here](https://github.com/emberjs/ember.js/tree/master/packages/internal-test-helpers/lib/test-cases) for other options.

2) QUnit.test becomes a property on the test case class

```js
// Old Style:
QUnit.test('<TEST_NAME>', function() {
  // test case here
});

// New Style:
moduleFor('<MODULE_NAME>', class extends TestCase {
  ['@test <TEST_NAME>'](assert) {
    // test case here
  }
});
```

3) Assertions should be references off of the callback assert argument

```js
// Bad:
QUnit.test('<TEST_NAME>', function() {
  equal(1, 1);
});

// Bad:
moduleFor('<MODULE_NAME>', class extends TestCase {
  ['@test <TEST_NAME>']() {
    this.assert.equal(1, 1);
  }
});

// Good:
moduleFor('<MODULE_NAME>', class extends TestCase {
  ['@test <TEST_NAME>'](assert) {
    assert.equal(1,1)
  }
});
```

Below are the packages that need to be converted:
- [x] container (🔒 @t-sauer) https://github.com/emberjs/ember.js/pull/15990
- [x] ember-application (🔒 @locks)
- [x] ember-debug (🔒 @locks) https://github.com/emberjs/ember.js/pull/15998
[...]

PR #15990, opened 2017-12-15
[Cleanup] Convert container tests to new style

t-sauer:
See #15988 

Two things I observed while converting:

- `expectAssertion` and `expectDeprecation` are not available on `this.assert`. Is this intended? Is there a newer replacement for that?
- There are a couple of `assert.expect(1)` which is not needed as far as I know, correct?

### DIFF
--- a/packages/container/tests/owner_test.js
+++ b/packages/container/tests/owner_test.js
@@ -1,16 +1,17 @@
 import { getOwner, setOwner, OWNER } from 'ember-utils';
+import { moduleFor, AbstractTestCase } from 'internal-test-helpers';
 
-QUnit.module('Owner', {});
+moduleFor('Owner', class extends AbstractTestCase {
+  ['@test An owner can be set with `setOwner` and retrieved with `getOwner`'](assert) {
+    let owner = {};
+    let obj = {};
 
-QUnit.test('An owner can be set with `setOwner` and retrieved with `getOwner`', function() {
-  let owner = {};
-  let obj = {};
+    assert.strictEqual(getOwner(obj), undefined, 'owner has not been set');
 
-  strictEqual(getOwner(obj), undefined, 'owner has not been set');
+    setOwner(obj, owner);
 
-  setOwner(obj, owner);
+    assert.strictEqual(getOwner(obj), owner, 'owner has been set');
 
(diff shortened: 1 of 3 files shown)


rwjblue:
> expectAssertion and expectDeprecation are not available on this.assert. Is this intended? Is there a newer replacement for that?

Yes, we should migrate to a custom `assert.throws` implementation (basically something that is aware of production and "does the right thing").  Definitely fodder for a future quest issue...

> There are a couple of assert.expect(1) which is not needed as far as I know, correct?

Correct, however when assertions are not in the top down flow of code I do think we should have an `assert.expect`.  Basically [require-expect](https://github.com/platinumazure/eslint-plugin-qunit/blob/master/docs/rules/require-expect.md) with the `except-simple` config option.

packages/container/tests/owner_test.js
```diff
@@ -1,16 +1,17 @@
 import { getOwner, setOwner, OWNER } from 'ember-utils';
+import { moduleFor, AbstractTestCase } from 'internal-test-helpers';
 
-QUnit.module('Owner', {});
+moduleFor('Owner', class extends AbstractTestCase {
+  ['@test An owner can be set with `setOwner` and retrieved with `getOwner`']() {
```
rwjblue: I'd generally prefer to add `assert` as the callback arg and use it below (instead of `this.assert`).  It is simpler, and "feels" more idiomatic to me.

@chancancode / @thoov / @locks - Thoughts?

rwjblue (commented):
Looks good to me, I only have one general comment around using `assert` argument instead of `this.assert`, but I'm happy to tackle that in a follow up...

packages/container/tests/owner_test.js
locks: I've been doing `this.assert` because of the code linked in the quest issue, but I prefer the argument version too.

thoov: I agree with having assert an argument over off of `this`. Definitely is more "standard" when looking at docs and existing testing approaches in the wild. I will update the issue to have more clear explanations about things we are looking for and better examples

t-sauer:
Updated from `this.assert` to the `assert` argument.

rwjblue (approved):

rwjblue:
Thank you!

PR #15992, opened 2017-12-16
[WIP] [FEATURE ember-metal-es5-getters] Implement feature

chancancode:
Moved to #16025

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -42,6 +42,11 @@ for a detailed explanation.
   Use Glimmer Components semantics for template-only components per
   [RFC](https://github.com/emberjs/rfcs/pull/278).
 
+* `ember-metal-es5-getters`
+
+  Define ES5 getters for computed properties, eliminating the need to access them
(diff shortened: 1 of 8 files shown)


btecu:
`package-lock.json` should not be included.

chancancode:
Seems like [this is a regression](https://github.com/emberjs/ember.js/files/1565411/results.pdf), I will investigate if the implementation can be improved.

chancancode:
The "Slightly different approach" commit (I might have broken IE11, but ignore that for now) seem to close the gap better getters on/off, but master vs either is still a (small) regression. So (perhaps unsurprisingly) moving the descriptor to meta does seems to have a small cost. But given the benefits, it's probably still worth it?

I'll see if I can get some results from ember-observer as well.

chancancode:
Since the alternative version was causing problems on IE 11, and it did not affect the feature-off case at all, I would like to leave that out for now and iterate on the feature-on performance separately (that is an important concern for enabling the feature, but not necessarily for merging the initial implementation).

rwjblue (approved):

PR #15993, opened 2017-12-16
removed `return this` in `lib/observer`

bekzod:
removed undocumented unused `return this`, it looks like something that can lead to problems :)

### DIFF
--- a/packages/ember-metal/lib/observer.js
+++ b/packages/ember-metal/lib/observer.js
@@ -23,8 +23,6 @@ import {
 export function addObserver(obj, _path, target, method) {
   addListener(obj, _path, target, method);
   watch(obj, _path);
-
-  return this;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


rwjblue:
I don't think we can arbitrarily remove the ability to chain, this is definitely a breaking change...

bekzod:
yes sure, should I wrap it `Proxy` object and deprecate it ? since chaining is not supported in other `Ember.` functions and since moving to modules import will chaining will work properly :P

mmun:
This is a top-level function so `this` will be the global or `undefined` depending on strict mode. I scanned the code base and we're not calling these functions with `apply` or `call`, putting them directly on a prototype, etc. I think its safe to remove these lines.

@bekzod Mind rebasing? May be a good idea to do a (careful!) audit of removable `return this;` in the whole code base.

mmun:
Ah, Rob pointed out that `Ember.addObserver(...).addObserver(...)` currently works. I missed this in my search. I can't imagine that anyone is actually relying on this behaviour though 🤔

mmun:
I think it is incredibly unlikely for someone to write that... Let's merge it and we'll revert if someone complains.

bekzod:
nice nice

PR #16007, opened 2017-12-18
fix deprecated option format

bekzod:


broccoli/packages.js
```diff
@@ -355,7 +355,7 @@ module.exports.rollupEmberMetal = function _rollupEmberMetal(tree, options) {
     rollup: {
       input: `index.js`,
       output: {
-        moduleId: 'ember-metal',
```
bekzod: showing this `options.moduleId is deprecated in favour of options.amd = { id: moduleId }` in build time

rwjblue (approved):

rwjblue:
I noticed this earlier as well, thanks for fixing!

PR #16009, opened 2017-12-19
remove `_beforeObserver` from `array_proxy`

bekzod:
@mmun does this look interesting ?


bekzod:
sorry I think I rushed a bit here :)

PR #16010, opened 2017-12-19
Cleanup ember-template-compiler's tests

karthiicksiva:
Related to #15988

### DIFF
--- a/packages/ember-template-compiler/tests/plugins/assert-input-helper-without-block-test.js
+++ b/packages/ember-template-compiler/tests/plugins/assert-input-helper-without-block-test.js
@@ -1,16 +1,16 @@
 import { compile } from '../../index';
+import { moduleFor, AbstractTestCase } from 'internal-test-helpers';
 
-QUnit.module('ember-template-compiler: assert-input-helper-without-block');
+moduleFor('ember-template-compiler: assert-input-helper-without-block', class extends AbstractTestCase {
+  ['@test Using {{#input}}{{/input}} is not valid']() {
 
-QUnit.test('Using {{#input}}{{/input}} is not valid', function() {
-  expect(1);
+    let expectedMessage =
+      `The {{input}} helper cannot be used in block form. ('baz/foo-bar' @ L1:C0) `;
 
-  let expectedMessage =
-    `The {{input}} helper cannot be used in block form. ('baz/foo-bar' @ L1:C0) `;
-
-  expectAssertion(() => {
-    compile('{{#input value="123"}}Completely invalid{{/input}}', {
-      moduleName: 'baz/foo-bar'
-    });
-  }, expectedMessage);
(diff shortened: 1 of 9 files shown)


packages/ember-template-compiler/tests/plugins/assert-input-helper-without-block-test.js
```diff
@@ -1,16 +1,17 @@
 import { compile } from '../../index';
+import { moduleFor, AbstractTestCase } from 'internal-test-helpers';
 
-QUnit.module('ember-template-compiler: assert-input-helper-without-block');
+moduleFor('ember-template-compiler: assert-input-helper-without-block', class extends AbstractTestCase {
+  ['@test Using {{#input}}{{/input}} is not valid'](assert) {
+    assert.expect(1);
```
thoov: Most of the time we can get rid of `assert.expect(1)` and I believe we can in these test cases. We will want to keep the `assert.expect(0)` however

karthiick:
@thoov I am not sure about how to move this? Any idea?
https://github.com/emberjs/ember.js/blob/master/packages/ember-template-compiler/tests/plugins/assert-reserved-named-arguments-test.js

//cc @rwjblue

thoov:
@karthiick There are a few approaches I can think of:

```js

if (EMBER_GLIMMER_NAMED_ARGUMENTS) {

moduleFor('ember-template-compiler: assert-reserved-named-arguments', class extends TestCase  {

	 ['@test @arguments is reserved'](assert) {
	      assert.expectAssertion(() => {
	        compile('{{@arguments}}', { moduleName: 'baz/foo-bar' });
	      }, `'@arguments' is reserved. ('baz/foo-bar' @ L1:C2) `);
	
	      assert.expectAssertion(() => {
	        compile(`{{#if @arguments}}Yup{{/if}}`, { moduleName: 'baz/foo-bar' });
	      }, `'@arguments' is reserved. ('baz/foo-bar' @ L1:C6) `);
	
	      assert.expectAssertion(() => {
	        compile(`{{input type=(if @arguments "bar" "baz")}}`, { moduleName: 'baz/foo-bar'});
	      }, `'@arguments' is reserved. ('baz/foo-bar' @ L1:C17) `);
    }

  /* 
	... add a test property explicitly for each (I like this as its easier to grep vs saving a few lines):
    '@arguments',
    '@args',
    '@Arguments', '@Args',
    '@A', '@FOO', '@Foo',
    '@.', '@_', '@-', '@$'
*/
});

} else {

moduleFor('ember-template-compiler: assert-reserved-named-arguments', class extends TestCase {
	/*
	 Add the else tests here
	*/
});
	
}

karthiick:
Made the changes. @thoov @rwjblue

rwjblue (approved):

rwjblue:
Thank you!

PR #16012, opened 2017-12-20
fixing regression in run-tests

ef4:
I broken this when I got rid of the unparseable Javascript that had been in this file. That version was apparently relying on implicit background promise resolution.

### DIFF
--- a/node-tests/fixtures/util-test/.eslintrc.js
+++ b/node-tests/fixtures/util-test/.eslintrc.js
@@ -1,0 +1,5 @@
+module.exports = {
+  rules: {
+    'no-unused-vars': 'off'
+  }
+}
--- a/node-tests/fixtures/util-test/rfc232.js
+++ b/node-tests/fixtures/util-test/rfc232.js
@@ -1,7 +1,7 @@
 import fooBar from 'my-app/utils/foo-bar';
 import { module, test } from 'qunit';
 
-module('Unit | Utility | foo bar', function() {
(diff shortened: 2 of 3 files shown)


node-tests/fixtures/util-test/rfc232.js
```diff
@@ -1,7 +1,7 @@
 import fooBar from 'my-app/utils/foo-bar';
 import { module, test } from 'qunit';
 
-module('Unit | Utility | foo bar', function() {
+module('Unit | Utility | foo bar', function(hooks) {
```
rwjblue: `hooks` is unused I believe

node-tests/fixtures/util-test/.eslintrc.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
```
rwjblue: We should fix this in a follow up.

rwjblue (approved):

node-tests/fixtures/util-test/rfc232.js
ef4: Right, which is why I removed it in my earlier PR.

This is a fixture that will be compared to blueprint output, and I was assuming it was intentional that the blueprint would have `hooks`. 

But yeah, possibly the blueprint should say: `function(/* hooks */) {` to make it clear that the argument is available if you want it but deliberately unused.

PR #16015, opened 2017-12-22
[CLEANUP] Convert ember-router tests to new style

thoov:
Apart of #15988 

During this I created a subtask on the Quest issue for us to move `expectAssertion` and `expectDeprecation` onto assert. This is just a subpart for the ember-router package so that these PRs arn't too big.


locks (approved):
✅

Issue #16014, opened 2017-12-21
concatenatedProperties mishandles {} and false


nathanhammond:
In a naive attempt to turn off `concatenatedProperties` I discovered this:

```javascript
var a = Ember.Component.extend({
  classNames: ['parent'],
  concatenatedProperties: {} // false also fails
});

console.log(a.proto().concatenatedProperties);
// => ["classNames", "classNameBindings", "attributeBindings", {}]
```

bekzod:
I think it should be disabled when `concatenatedProperties: null`,  can confirm such behavior exist for `mergedProperties` too, I can fix if we define right behavior first here :)

rwjblue:
I do not think we intend to support disabling `concatenatedProperties`. The provided example should trigger an assertion (basically if the value passed as `concatenatedProperties` is not an array, we should assert).

esbanarango:
@rwjblue @nathanhammond  Is this issue worth to be labeled with `Needs Team Discussion`? Or move this to [discuss](https://discuss.envoy.com/) and keep the conversation there?

I think we could close this as an issue.

pixelhandler:
Closing for now.

PR #16017, opened 2017-12-22
assert for non-array values in `concatenatedProperties` and `mergedProperties`

bekzod:
possible fix for #16014

### DIFF
--- a/packages/ember-metal/lib/mixin.ts
+++ b/packages/ember-metal/lib/mixin.ts
@@ -62,8 +62,10 @@ function concatenatedMixinProperties(
 ) {
   // reset before adding each new mixin to pickup concats from previous
   let concats = values[concatProp] || base[concatProp];
-  if (props[concatProp]) {
-    concats = concats ? a_concat.call(concats, props[concatProp]) : props[concatProp];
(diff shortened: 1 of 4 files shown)


rwjblue:
I mentioned in the other issue, I don't think this is the right path forward. I think we should assert when given anything that isn't an array...

bekzod:
ok

rwjblue:
Had a chance to pick this back up @bekzod?

bekzod:
looking at it now :)

bekzod:
@rwjblue just to note
```javascript
let mixinA = Mixin.create({
   concatenatedProperties: ['foo'],
   foo: ['bar']
})

let mixinB = Mixin.create({
   foo: 'boom'
})

mixin({}, mixinA, mixinB); // will not throw assertion 

let mixinC = Mixin.create({
   concatenatedProperties: 'boom'
})

mixin({}, mixinA, mixinC); // will throw assertion 

```

bekzod:
@rwjblue maybe we could also merge #15704 😉

bekzod:
closing, too much has changed and EmberObject probably will be gone soon

PR #16023, opened 2017-12-22
Make `ember-application` package tests pass without jQuery.

rwjblue:
This gets the `ember-application` package test suite passing when jQuery is not loaded.

To test a given package without jQuery, navigate to:

```
http://localhost:4200/tests/index.html?hidepassed&jquery=none&package=ember-application
```

Primary changes:

* Begin removing jQuery usage in application tests.
* Create shared qunit-fixture setup functionality.
* Remove jQuery usage in abstract test cases.
* Implement native DOM event dispatcher.
* Remove eager jQuery destructuring.

### DIFF
--- a/packages/ember-application/tests/system/application_test.js
+++ b/packages/ember-application/tests/system/application_test.js
@@ -212,7 +212,8 @@ moduleFor('Ember.Application, default resolver with autoboot', class extends Def
   }
 
   [`@test Minimal Application initialized with just an application template`]() {
-    jQuery('#qunit-fixture').html('<script type="text/x-handlebars">Hello World</script>');
+    this.setupFixture('<script type="text/x-handlebars">Hello World</script>');
+
     this.runTask(() => this.createApplication());
     this.assertInnerHTML('Hello World');
   }
--- a/packages/ember-application/tests/system/application_test.js
+++ b/packages/ember-application/tests/system/application_test.js
@@ -35,14 +35,13 @@ import {
 } from 'internal-test-helpers';
 
 moduleFor('Ember.Application, autobooting multiple apps', class extends ApplicationTestCase {
-  constructor() {
-    jQuery('#qunit-fixture').html(`
+  get fixture() {
+    return `
       <div id="one">
         <div id="one-child">HI</div>
       </div>
       <div id="two">HI</div>
-    `);
-    super();
+    `;
   }
 
   get applicationOptions() {
@@ -213,7 +212,6 @@ moduleFor('Ember.Application, default resolver with autoboot', class extends Def
 
   [`@test Minimal Application initialized with just an application template`]() {
     this.setupFixture('<script type="text/x-handlebars">Hello World</script>');
-
     this.runTask(() => this.createApplication());
     this.assertInnerHTML('Hello World');
   }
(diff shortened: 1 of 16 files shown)


packages/ember-application/tests/system/application_test.js
```diff
@@ -36,12 +36,13 @@ import {
 
 moduleFor('Ember.Application, autobooting multiple apps', class extends ApplicationTestCase {
   constructor() {
-    jQuery('#qunit-fixture').html(`
+    let fixture = document.querySelector('#qunit-fixture');
```
rwjblue: Update this to use `get fixture() {}`

packages/ember-application/tests/system/bootstrap-test.js
```diff
@@ -25,6 +23,6 @@ moduleFor('Ember.Application with default resolver and autoboot', class extends
 
   ['@test templates in script tags are extracted at application creation'](assert) {
     this.runTask(() => this.createApplication());
-    assert.equal(this.$('#app').text(), 'Hello World!');
+    assert.equal(document.querySelector('#app').textContent, 'Hello World!');
```
chancancode: This can be `getElementById`, but we can/should probably just make `#app` `this.element` so you can just use `this.assertText`

packages/ember/tests/helpers/link_to_test.js
```diff
@@ -655,7 +655,7 @@ moduleFor('The {{link-to}} helper - nested routes and link-to arguments', class
         {{#each model as |person|}}
           <li>
             {{#link-to 'item' person}}
```
chancancode: Why not `id=person.id` here too?

packages/internal-test-helpers/lib/test-cases/abstract.js
```diff
@@ -43,6 +48,11 @@ export default class AbstractTestCase {
     return run.next(callback);
   }
 
+  setupFixture(innerHTML) {
+    let fixture = document.querySelector('#qunit-fixture');
```
chancancode: `getElementById`?

packages/internal-test-helpers/lib/test-cases/abstract.js
```diff
@@ -88,11 +98,17 @@ export default class AbstractTestCase {
...
+      element = this.element.querySelector(selector);
+    } else {
+      element = selector;
+    }
+    return element.click();
   }
 
   textValue() {
-    return this.$().text();
+    return this.element ? this.element.textContent : '';
```
chancancode: does anything break without the guard?

chancancode (approved):
LGTM

packages/ember/tests/helpers/link_to_test.js
rwjblue: Good point, will update to that.

packages/internal-test-helpers/lib/test-cases/abstract.js
rwjblue: Yes, some of the `visit` tests with `shouldRender: false` throw without the guard. Specifically they are trying to confirm that nothing has rendered. 

I could remove the guard, and update those handful of tests to simply use `assert.equal(document.getElementById('qunit-fixture').children.length, 0)`.

chancancode: I see.. maybe put that in an `assertNothingRendered` method or something. Perhaps we can just assert `this.element = null`.. but I am not sure if that semantically works for all the tests.

rwjblue: It was only one test after the other visit test refactors that I did so I just put an inline assertion there...

PR #16039, opened 2018-01-01
Begin publishing npm tarballs to S3.

rwjblue:
This PR finally allows us to easily consume and test alpha, canary, beta, and release channels with `ember-source` as an `npm` dependency (previously to use these channel specific builds you **must** use them via `bower`).

For example, to use the canary channel from an app (or addon) after this lands you would have the following in your `package.json`:

```js
{
  // ...snip...
  devDependencies: {
    //...snip...
    "ember-source": "http://builds.emberjs.com/canary.tgz",
  }
}
```

The following assets are published for each channel (using `canary` for illustrative purposes, but the same paths are available for each):

```
http://builds.emberjs.com/canary.tgz
http://builds.emberjs.com/canary/daily/20180101.tgz
http://builds.emberjs.com/canary/shas/SOMESHAHERE.tgz
```

Once this lands, we can begin removing bower and globals build publishing infrastructure...

### DIFF
--- a/bin/publish_builds
+++ b/bin/publish_builds
@@ -5,9 +5,7 @@ echo -e "PULL_REQUEST: ${TRAVIS_PULL_REQUEST}\n"
 
 if [ "$TRAVIS_PULL_REQUEST" == "false" ] && [ "$PUBLISH" == "true" ]; then
 
-  if [ "$EMBER_ENV" != "production" ]; then
-    DISABLE_JSCS=true DISABLE_JSHINT=true ember build --environment=production
-  fi
+  ./bin/build-for-publishing.js
 
   yarn run docs # Generate documentation to be published.
   ./bin/publish_to_s3.js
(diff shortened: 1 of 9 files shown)


bin/publish_builds
```diff
@@ -5,11 +5,8 @@ echo -e "PULL_REQUEST: ${TRAVIS_PULL_REQUEST}\n"
 
 if [ "$TRAVIS_PULL_REQUEST" == "false" ] && [ "$PUBLISH" == "true" ]; then
 
-  if [ "$EMBER_ENV" != "production" ]; then
-    DISABLE_JSCS=true DISABLE_JSHINT=true ember build --environment=production
```
mixonic: `yarn build` seems to replace this, however `yarn build` does not `DISABLE_JSCS=true DISABLE_JSHINT=true`, if I read this correctly.

Is this something to be addressed?

rwjblue: Yes, but the fix is to remove all references to `DISABLE_JSCS` and `DISABLE_JSHINT` in other files...

Basically, they were environment variables that emberjs-build was aware of and made special, but since we migrated away from emberjs-build (sometime before Ember 2.10?) they no longer have meaning.

mixonic (approved):

rwjblue:
This has been backported to beta and release channels as well...

PR #16043, opened 2018-01-02
Module-unification component blueprint

GavinJoyce:
Part of https://github.com/ember-cli/ember-cli/issues/7530

This PR adds initial support for `ember g component foo-bar` from a module unification app.

TODO:

 * [x] [Add `project.isModuleUnification()`](https://github.com/ember-cli/ember-cli/pull/7541)
 * [x] Different app types:
    * [x] app
      * [x] component in root
      * [x] nested component
    * [x] addon
    * [x] addon --dummy
    * [x] in-repo addon
    * [x] add tests for different component test variations
 *  ~fix test failures since updating to the laters ember-cli~ (see https://github.com/emberjs/ember.js/pull/16043#issuecomment-366527400)

QUESTIONS:

 * [x] Do we have an example MU app which includes component tests? (might just work with `component-test.js`) (**update:** https://github.com/stonecircle/mu-first-post/pull/1)
 * [x] I assume that we don't need to support `--path` option? **update: correct, we don't need to support it**
 * [x] What's the timeline for https://github.com/emberjs/rfcs/pull/232? Should we support it in MU blueprints? **yes - support can land if future PRs though**
 * [x] Update to ember-cli `3.1.0-beta.1` **update: see https://github.com/emberjs/ember.js/pull/16043#discussion_r168765487**

USEFUL LINKS:

 * [MU App Golden Spike](https://github.com/iezer/mu-golden-spike)
 * [MU Addon example](https://github.com/stonecircle/mu-first-post)


rwjblue:
CI failure is due to eslint not being able to parse our templating, I believe we need to disable eslint from running on `blueprints/**/files/`...

package.json
```diff
@@ -94,7 +94,7 @@
     "chalk": "^2.3.0",
     "common-tags": "^1.4.0",
     "dag-map": "^2.0.2",
-    "ember-cli": "2.10.0",
+    "ember-cli": "github:ember-cli/ember-cli",
```
GavinJoyce: `project.isModuleUnification()` was just recently merged to master

blueprints/component/index.js
```diff
@@ -21,13 +25,45 @@ module.exports = {
...
+
+    return path.join(this.path, filesDirectory);
+  },
+
   fileMapTokens: function() {
+    let isModuleUnification = isModuleUnificationProject(this.project);
+
     return {
+      __root__(options) {
+        if (isModuleUnification) {
```
ro0gr: I think this check will break `--dummy` and `--in-repo-addon` cli flags. So if we have a module unification layout then
```
  ember g component some-component --dummy 
```
will generate component in the `src/` instead of `dummy/` dir.

ro0gr: Hm... I've just realized that dummy apps has to support module unification layout as well 🤔

GavinJoyce: Thanks, yes. I have a `Add support for addons` todo item above and I've [asked a question](https://embercommunity.slack.com/archives/C5JN29NTC/p1515352731000031) on #st-module-unification looking for an example MU addon so that I can add support for it too

GavinJoyce:
This ready for review. There is one small remaining item to point at the next beta release of ember-cli whenever it's released

package.json
```diff
@@ -100,7 +100,7 @@
     "chalk": "^2.3.0",
     "common-tags": "^1.7.2",
     "dag-map": "^2.0.2",
-    "ember-cli": "2.18.0",
+    "ember-cli": "github:ember-cli/ember-cli",
```
alexander-alvarez: do we want to keep this?

GavinJoyce: no, ideally we'll update this to the next ember cli beta

alexander-alvarez: sorry I read what was written but didn't understand properly.
the work looks good :smiley:

package.json
```diff
@@ -100,7 +100,7 @@
     "chalk": "^2.3.0",
     "common-tags": "^1.7.2",
     "dag-map": "^2.0.2",
-    "ember-cli": "2.18.0",
+    "ember-cli": "github:ember-cli/ember-cli#6a85734c52d129fc0824ec3fbf21bd01f9c9fa38",
```
GavinJoyce: I can't point to `` as intended as experiments are removed from betas and releases. Pointing to https://github.com/ember-cli/ember-cli/commit/6a85734c52d129fc0824ec3fbf21bd01f9c9fa38 instead

GavinJoyce:
Hm, there seems to be some failing tests around differences with `destroyApp` in unrelated tests

GavinJoyce:
There are a couple of things preventing me from currently upgrading to ember-cli `3.1.0-beta.1`:

 * It appears that [all experiments are disabled for beta and releases](https://github.com/ember-cli/ember-cli/blob/master/RELEASE.md#test). This means that the [`MODULE_UNIFICATION` experiment](https://github.com/ember-cli/ember-cli/blob/203164d6cc5be60918b39ce7e25530bdd22b739f/lib/experiments/index.js#L5) and `project.isModuleUnification` aren't available in ember-cli `3.1.0-beta.1` (we agreed that this is what I should ideally upgrade to in our st-module-unification call earlier this week)
 * There are a bunch of unrelated failing blueprint tests when updating to the latest ember-cli SHA, eg: https://github.com/emberjs/ember.js/pull/16253 

Some options:

 * revert https://github.com/ember-cli/ember-cli/pull/7586 to remove the experiment check from `project.isModuleUnification`, fix the unrelated test failures and wait for `3.1.0-beta.2`
 * upgrade ember-cli to https://github.com/ember-cli/ember-cli/commit/76175513c5e49caf5405b7304e83ddf3e799c68e (this is the current state of this PR and has no blockers to merging should we wish).

/cc @mixonic as you were keen for me to add the experiment check in `project.isModuleUnification`.

mixonic:
We discussed this at the MU strike team meeting this week. The SHA pinned in this PR is before several changes where made to Ember-CLI that seem to break the Ember blueprints. I'm worried that breakage could be impacting real apps (on Canary if not anything stable) when using generators.

However that breakage is not related to what @GavinJoyce has been digging into here.

This looks great and is wholly behind a feature flag. Let's continue to iterate in master. Thanks @GavinJoyce!

blueprints/component-test/index.js
```diff
@@ -6,6 +6,7 @@ const isPackageMissing = require('ember-cli-is-package-missing');
 const getPathOption = require('ember-cli-get-component-path-option');
 
 const useTestFrameworkDetector = require('../test-framework-detector');
+const { isModuleUnificationProject } = require('../module-unification');
```
ro0gr: This will break on node@4

GavinJoyce: Thanks @ro0gr, I'll update now

GavinJoyce: https://github.com/emberjs/ember.js/pull/16261

PR #16045, opened 2018-01-02
[BUGFIX beta] Fix double debug output

### LINKED ISSUE #16041: Ember.Logger.debug results in double output
This means that on a new Ember app using v3.0.0-beta.1 we see the following output for a debug build:

screenshot

The cause is the boolean logic here:

https://github.com/emberjs/ember.js/blob/bd7e62fdbd928dbc77a0810101caa152966a0ea7/packages/ember-console/lib/index.js#L92

I believe it should be refactored to:

```js
  debug() {
    if (console.debug) {
      return console.debug(...arguments); // eslint-disable-line no-console
    } else {
      return console.info(...arguments); // eslint-disable-line no-console
    }
  },
```

Serabe:
Fixes #16041

### DIFF
--- a/packages/ember-console/lib/index.js
+++ b/packages/ember-console/lib/index.js
@@ -89,7 +89,13 @@ export default {
    @public
   */
   debug() {
-    return console.debug && console.debug(...arguments) || console.info(...arguments); // eslint-disable-line no-console
+    /* eslint-disable no-console */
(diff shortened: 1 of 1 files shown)


rwjblue (approved):

rwjblue:
Thank you @Serabe!

Issue #15686, opened 2017-09-29
Microsoft Edge 15 on Saucelabs


locks:
Error:

```
[...]

PR #16047, opened 2018-01-03
Remove Saucelabs and use BrowserStack for cross-browser testing

kategengler:
- Use BrowserStack for cross-browser testing in place of SauceLabs. 
- Update the browsers that are tested on. Question: Are these what we should be testing against?
- Unfortunately remove the SauceLabs browser compatibility badge. BrowserStack unfortunately has no analog. Should Browser support be hardcoded in the README? 

This behaves slightly differently than SauceLabs: the test results are printed locally by testem, similar to how it is done for local browsers. An example of the output can be seen here: https://travis-ci.org/kategengler/ember.js/builds/324480730

**NEEDED**
- [x] BrowserStack Credentials for the Ember project. Update: Credentials added. Now just awaiting response from BrowserStack turning the account from free trial to open source account. 

** TODO **
- [x] Extract FailureOnlyReporter to lib and use as base for browser summarizing reporter
  
Fixes https://github.com/emberjs/ember.js/issues/15686.


rwjblue (commented):
👍

rwjblue:
@kategengler - I emailed the BrowserStack folks to get the ball rolling RE: credentials...

locks (approved):

rwjblue:
Thank you @kategengler!

Issue #15501, opened 2017-07-13
[QUEST] Internals should prefer generic array methods over prototype_extensions


Serabe:
This supersedes  #10899 that is stale.

- [x] addArrayObserver
- [x] removeArrayObserver
- [x] `objectAt`
- [x] `replace`
- [x] `remoteAt`
- [ ] `insertAt` https://github.com/emberjs/ember.js/pull/16048
- [ ] pushObject
- [ ] pushObjects
- [ ] re-org public api
- [ ] etc ...
- [ ] docs
- [ ] all the tests

The goal here is merely to decouple the internals from the the observable based prototype extensions.
This will hopefully give us the flexibility to make a future call re: prototype extensions in general. This also eases the pain of using ember with prototype extensions disabled.

All this is accomplished without breaking semver in anyway.

More todo:
- [x] Merge Enumerable + NativeArray
- [x] Merge MutableEnumerable + MutableArray
- [ ] Deprecate Enumerable as [bugfix release]
- [ ] Deprecate MutableEnumerable as [bugfix release]

alvincrespo:
@Serabe How can I help? This is my first time diving into Ember internals.

Serabe:
This might be closed as per https://github.com/emberjs/ember.js/pull/16048#issuecomment-367929653

Serabe:
Removing labels to avoid people working on this until there is some clarification on the goal.

Serabe:
Don't really know. source needs to be checked.

mmun:
@bekzod was working on this recently.

locks:
Thanks everyone for your help! Closing as there's no more work required here :)

PR #16048, opened 2018-01-03
Prefer generic `insertAt` over prototype_extensions

wongpeiyi:
https://github.com/emberjs/ember.js/issues/15501 `insertAt`

Duplicated one test for testing the prototype extension still works.

### DIFF
--- a/packages/ember-runtime/lib/mixins/mutable_array.js
+++ b/packages/ember-runtime/lib/mixins/mutable_array.js
@@ -32,6 +32,15 @@ export function removeAt(array, start, len) {
   return array;
 }
 
+export function insertAt(array, idx, object) {
+  if (idx > get(array, 'length')) {
+    throw new EmberError(OUT_OF_RANGE_EXCEPTION);
+  }
+
+  array.replace(idx, 0, [object]);
+  return array;
+}
+
 /**
   This mixin defines the API for modifying array-like objects. These methods
   can be applied only to a collection that keeps its items in an ordered set.
@@ -116,12 +125,7 @@ export default Mixin.create(EmberArray, MutableEnumerable, {
     @public
   */
   insertAt(idx, object) {
-    if (idx > get(this, 'length')) {
-      throw new EmberError(OUT_OF_RANGE_EXCEPTION);
-    }
-
-    this.replace(idx, 0, [object]);
-    return this;
+    return insertAt(this, idx, object);
   },
 
   /**
@@ -166,7 +170,7 @@ export default Mixin.create(EmberArray, MutableEnumerable, {
     @public
   */
   pushObject(obj) {
-    this.insertAt(get(this, 'length'), obj);
+    insertAt(this, get(this, 'length'), obj);
     return obj;
   },
 
@@ -261,7 +265,7 @@ export default Mixin.create(EmberArray, MutableEnumerable, {
     @public
   */
   unshiftObject(obj) {
-    this.insertAt(0, obj);
+    insertAt(this, 0, obj);
     return obj;
   },
 
--- a/packages/ember-runtime/lib/mixins/mutable_array.js
+++ b/packages/ember-runtime/lib/mixins/mutable_array.js
@@ -4,9 +4,10 @@
 
 import {
   get,
-  Mixin,
+  replace as emberReplace,
   beginPropertyChanges,
-  endPropertyChanges
+  endPropertyChanges,
+  Mixin
 } from 'ember-metal';
 import MutableEnumerable from './mutable_enumerable';
 import EmberArray, { objectAt } from './array';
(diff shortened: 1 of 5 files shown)


wongpeiyi:
@Serabe can you see if this is in line with what the epic requires? Don't think `insertAt` was actually extracted despite the checkbox being checked. 

Would like to take a stab at the remainder in this fashion if this looks ok

packages/ember-runtime/tests/suites/mutable_array/insertAt.js
```diff
@@ -1,18 +1,19 @@
 import { SuiteModuleBuilder } from '../suite';
 import { get } from 'ember-metal';
+import { insertAt } from '../../../mixins/mutable_array';
 
 const suite = SuiteModuleBuilder.create();
 
 suite.module('insertAt');
 
-suite.test('[].insertAt(0, X) => [X] + notify', function() {
+suite.test('insertAt([], 0, X) => [X] + notify', function() {
```
thoov: I think its beneficial to have some tests still use the array prototype until it is fully removed. I'm not sure if we have tests that specific check for this already elsewhere or if we should add those checks here

wongpeiyi: I followed the example of https://github.com/emberjs/ember.js/pull/13534/files#diff-d1dc85945ef368d3bcdd3e01234d3f8eR118 and duplicated the last test, keeping it to use the array prototype extension. See https://github.com/emberjs/ember.js/pull/16048/files/ac0a1455b73d59a5b1256f81096343a90f846ebd#diff-eaa813784954ae3195242cb1d1763b7dR171

I agree it's not very clear though. I can change the test's label but not sure if the test itself needs to be more specific.

thoov: I guess what I'm trying to get at is that we should maintain at least one test that simply checks that the prototypes are still extended. This could simply be something like:

```js
test('that prototypes are properly extended', function(assert) {
  assert.equal(typeof [].insertAt, 'function');
  // ... all other extensions
});
```

I don't know if we have something like this already but if not we should add it

wongpeiyi: Do you mean something like https://github.com/emberjs/ember.js/blob/master/packages/ember-runtime/tests/system/string/camelize_test.js#L6-L10 but the inverse?

Added a commit for that for both `removeAt` and `insertAt`

Serabe:
LGTM. Need approval from core yet.

locks:
I see with my little eye some git conflicts. Can you take a look @wongpeiyi?

packages/ember-runtime/tests/suites/mutable_array/insertAt.js
thoov: Yep that looks good. Thanks @wongpeiyi

wongpeiyi:
@locks thanks, rebased.

rwjblue (approved):

packages/ember-runtime/lib/mixins/mutable_array.js
```diff
@@ -19,6 +19,15 @@ import MutableEnumerable from './mutable_enumerable';
 import Enumerable from './enumerable';
 import { Error as EmberError } from 'ember-debug';
 
+function replace(array, idx, amt, objects) {
+  if (typeof array.replace === 'function') {
+    array.replace(idx, amt, objects);
+    return array;
+  }
+
+  return array.splice(idx, amt, ...objects);
```
rwjblue: Hmm, when the array has `.replace` we return the mutated array, but if it doesn't we return the deleted elements.

Can you tweak this so that its consistent?

packages/ember-runtime/lib/mixins/mutable_array.js
```diff
@@ -19,6 +19,15 @@ import MutableEnumerable from './mutable_enumerable';
 import Enumerable from './enumerable';
 import { Error as EmberError } from 'ember-debug';
 
+function replace(array, idx, amt, objects) {
```
rwjblue: I would have expected this to be much closer to [NativeArray#replace](https://github.com/emberjs/ember.js/blob/b3ad8b4999cb1043b1f76debf187187c139327cb/packages/ember-runtime/lib/system/native_array.js#L54-L71)...

rwjblue (changes requested):

wongpeiyi:
@rwjblue can I get your eyes on this again? Had to rebase, and tests failed because of master at that time

mmun:
@wongpeiyi FYI, I think we're going to take a slightly different approach. I will lay the foundation down soon and possibly start up a new quest issue.

rwjblue:
Sorry for the run around here @wongpeiyi ☹️

Issue #16049, opened 2018-01-03
Chai 4/Ember 3.0.0-beta.1 error: You attempted to access the `firstObject.inspect` property


Dhaulagiri:
Running tests with Chai4 and Ember 3 beta 1, I'm getting the below error 
 
> Error: Assertion Failed: You attempted to access the `firstObject.inspect` property (of ). Due to certain internal implementation details of Ember, the `firstObject` property previously contained an internal "descriptor" object (a private API), therefore `firstObject.inspect` would have been `undefined`. This internal implementation detail was never intended to be a public (or even intimate) API.

> This internal implementation detail has now changed and the (still private) "descriptor" object has been relocated to the object's "meta" (also a private API). Soon, accessing `firstObject` on this object will return the computed value (see RFC #281 for more details).

The error seems to stem from the way Chai 4's `.to.contain` logic works in a test like the below:

```
  let arr;
  beforeEach(function() {
    arr = [
      EmberObject.create({ id: 1 }),
      EmberObject.create({ id: 2 })
    ];
  });

  it('tests', function() {
      expect(arr).to.contain('foo');
  });
```
The Chai code that blows up is something like this:

`if (value && typeof value.inspect === 'function' &&`

I'm not sure how to reproduce this via twiddle, but you  clone [this repo](https://github.com/Dhaulagiri/test-app) and run the included test you can see the error.  Note that if I drop back to Chai 3 the problem goes away.

rwjblue:
The issue here is that chai@4 is attempting to provide a nice display message for the failing assertion with some custom display logic.  To accomplish that they are using a copy/fork of Node's `util.inspect()` ([see here](https://github.com/chaijs/chai/blob/19e2c18cffc0b21c007d5ac564207733742fa6aa/lib/chai/utils/inspect.js)).  When invoked, `util.inspect()` recursively inspects each property, and if an object has an `.inspect` function it will invoke it essentially as a way to customize the inspection output (ala `toString()`, but under specific situations).

In Node 7+ the recommended way to customize inspection changed from implementing an `.inspect` function on your object to using a custom symbol ([see here](https://github.com/nodejs/node/commit/59714cb7b3918c9e4fcd94013b778078d3710705)) which our existing descriptor trap already handles properly.

---

I believe the correct solution here is to add `inspect` to the list of exceptions in our descriptor trap here: 

[...]

PR #16050, opened 2018-01-03
[BUGFIX beta] add inspect and constructor to list of descriptor exceptions

Dhaulagiri:
In addition to `inspect` I ended up needing to add `constructor` as well to get things working (confirmed locally). Let me know if there is anything else I can do here.

Fixes https://github.com/emberjs/ember.js/issues/16049

### DIFF
--- a/packages/ember-metal/lib/properties.js
+++ b/packages/ember-metal/lib/properties.js
@@ -112,9 +112,19 @@ if (EMBER_METAL_ES5_GETTERS) {
         get(descriptor, property) {
           if (property === DESCRIPTOR) {
             return descriptor;
-          } else if (property === 'toString' || property == 'valueOf' ||
-            property == 'inspect' || property == 'constructor' ||
-            (Symbol && property === Symbol.toPrimitive)) {
+          } else if (
+            property === 'prototype' ||
+            property === 'constructor' ||
+            property === 'nodeType'
+          ) {
+            return undefined;
+          } else if (
+            property === 'toString' ||
+            property === 'valueOf' ||
+            property === 'inspect' ||
+            Symbol && property === Symbol.toPrimitive ||
+            Symbol && property === Symbol.toStringTag
+          ) {
             return () => '[COMPUTED PROPERTY]';
           }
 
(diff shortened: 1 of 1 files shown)


rwjblue:
@Dhaulagiri - I pushed some  additional commits here, can you review?

rwjblue:
Also, FWIW, with these changes your demo repo no longer fails due to the descriptor trap (though it does have a failing tests, but I think that is intentional).

rwjblue:
@Dhaulagiri - I added `nodeType` to the list of items that return `undefined` (jQuery is checking to see if it is an element there).

 Also, it seems like the first one is more app specific and might actually be a _correct_ assertion (e.g. something in the app is doing `this.isFree._dependentKeys`).

Dhaulagiri:
> Also, it seems like the first one is more app specific and might actually be a correct assertion

I'll take a closer look.  It may end up being another Chai thing though as this works fine our app but fails trying to do a deep equal comparison between two arrays with Chai.  I wonder if it's trying to iterate through the keys somehow to do the comparison?  It may take me a bit to find time to dig into that but the other stuff that is here will be a big help for us.

rwjblue:
> I'll take a closer look. It may end up being another Chai thing though as this works fine in our app but fails trying to do a deep equal comparison between two arrays with Chai.

Thanks for digging into it!  We can definitely iterate on what we have here....

> It may take me a bit to find time to dig into that but the other stuff that is here will be a big help for us.

No problem! I'm going to land this now then, so it makes it into the next beta (making it easier to test for others also)...

packages/ember-metal/lib/properties.js
```diff
@@ -112,7 +112,19 @@ if (EMBER_METAL_ES5_GETTERS) {
         get(descriptor, property) {
           if (property === DESCRIPTOR) {
             return descriptor;
-          } else if (property === 'toString' || property == 'valueOf' || (Symbol && property === Symbol.toPrimitive)) {
+          } else if (
+            property === 'prototype' ||
+            property === 'constructor' ||
+            property === 'nodeType'
```
chancancode: Are these not needed for the other branch (no native `Proxy`) as well?

rwjblue: Everything returns undefined by default in that branch, we only add assertions for the specific properties that we know about...

PR #16056, opened 2018-01-04
ember-glimmer w/o jQuery initial work

wycats:
This commit gets the vast majority of tests in ember-glimmer working
without jQuery.

It creates a small test facility called NodeQuery that looks enough like
jQuery for most of the existing tests to work, but does not attempt to
use the facility outside of tests.

Once all tests pass, we should consider a quest issue to clean this up
further, but it's nice to get so many tests passing with such a huge
diff.

As in the previous pull request, you can run the tests without jQuery for a package by navigating to:

```
http://localhost:4200/tests/index.html?hidepassed&jquery=none&package=ember-glimmer
```


rwjblue:
This is looking good.  I like the idea of the `NodeQuery` crutch, so that we can iterate on the cleanup efforts.

packages/internal-test-helpers/lib/test-cases/abstract.js
```diff
@@ -96,7 +98,19 @@ export default class AbstractTestCase {
   }
 
   $(sel) {
-    return sel ? jQuery(sel, this.element) : jQuery(this.element);
+    if (sel instanceof Element) {
```
rwjblue: Should probably test for `nodeType`, good for future refactor...

PR #16057, opened 2018-01-04
Refactor test helpers `visit` implementation to be consistent.

rwjblue:
Ensures that the `ApplicationInstance` instance is always gathered, regardless of if the `visit` itself fails, and also that it properly cleans up (even when the visit is rejected).

Requires #16054.

### DIFF
--- a/packages/ember-application/tests/system/visit_test.js
+++ b/packages/ember-application/tests/system/visit_test.js
@@ -142,7 +142,9 @@ moduleFor('Ember.Application - visit()', class extends ApplicationTestCase {
        * Visit on the application a second time. The application should remain
        * booted, but a new instance will be created.
        */
-      return this.visit('/');
+      return this.application.visit('/').then(instance => {
+        this.applicationInstance = instance;
+      });
     }).then(() => {
       assert.ok(appBooted === 1, 'the app should not be booted again');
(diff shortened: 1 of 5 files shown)


rwjblue:
Best way to review this is by just looking at the last commit: https://github.com/emberjs/ember.js/pull/16057/commits/94a54098f1673620cb9e31707438433106e0a760

rwjblue:
Rebased now that https://github.com/emberjs/ember.js/pull/16057 landed...

packages/ember-application/tests/system/visit_test.js
```diff
@@ -142,7 +142,9 @@ moduleFor('Ember.Application - visit()', class extends ApplicationTestCase {
        * Visit on the application a second time. The application should remain
        * booted, but a new instance will be created.
        */
-      return this.visit('/');
+      return this.application.visit('/').then(instance => {
+        this.applicationInstance = instance;
```
krisselden: isn't this already set?

rwjblue: No, just above we destroyed it and set it to null in this specific test.  This test is intentionally confirming that calling `this.application.visit()` will create _another_ application instance...

krisselden (approved):

PR #16060, opened 2018-01-05
[BUGFIX] Recompute link-to active prop when attrs.params changes

jasonmit:
Prior to 2.18.0, the link-to component would recompute `active`/`_active` when `attrs.params` changed.  This restores prior functionality where a deeply nested async relationship that resolves after initial render of a link-to and meets the criteria `this._isActive` then the active class is applied to the link element.  Currently, that functionality is broken and the active class is never applied when the async relationship settles.

I'll update tests before this is merged once it's clear that this wasn't an intentional breaking change.
  
<= 2.17.0 behavior https://github.com/emberjs/ember.js/blob/v2.17.0/packages/ember-glimmer/lib/components/link-to.js#L593

### DIFF
--- a/packages/ember-glimmer/lib/components/link-to.ts
+++ b/packages/ember-glimmer/lib/components/link-to.ts
@@ -573,7 +573,7 @@ const LinkComponent = EmberComponent.extend({
     return this.get('_active') ? get(this, 'activeClass') : false;
   }),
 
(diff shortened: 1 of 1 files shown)


rwjblue (commented):
Changes seem good to me, can you add a test case that fails prior to this (so we can ensure we don’t regress again)?

PR #16061, opened 2018-01-05
Work towards using QUnit 2 APIs.

rwjblue:
Migrates ember-runtime and ember-metal to using QUnit 2 APIs (avoiding global assertions, using beforeEach/afterEach instead of setup/teardown, etc).

It also updates the test harness for `moduleFor` and the older `ember-dev` system to honor both `setup`/`teardown` and `beforeEach`/`afterEach`.

### DIFF
--- a/packages/ember-metal/tests/alias_test.js
+++ b/packages/ember-metal/tests/alias_test.js
@@ -13,11 +13,11 @@ import {
 let obj, count;
 
 QUnit.module('ember-metal/alias', {
-  setup() {
+  beforeEach() {
     obj = { foo: { faz: 'FOO' } };
     count = 0;
   },
-  teardown() {
+  afterEach() {
     obj = null;
   }
 });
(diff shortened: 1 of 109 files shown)


rwjblue:
@thoov - This is somewhat related to the general purpose "test cleanup" quest issue, would you mind reviewing?

packages/ember-metal/tests/alias_test.js
```diff
@@ -108,5 +108,5 @@ function assertDirty(obj, callback, label) {
   let tag = tagFor(obj);
   let tagValue = tag.value();
   callback();
-  ok(!tag.validate(tagValue), label);
+  QUnit.config.current.assert.ok(!tag.validate(tagValue), label);
```
thoov: Whats your take on this vs passing assert into assertDirty?

thoov:
Other than the one comment I had it LGTM

packages/ember-metal/tests/alias_test.js
rwjblue: just my own laziness I think

thoov: I think there is just one occurrence of assertDirty so I say we hook it up and then merge this

PR #16071, opened 2018-01-05
More test cleanup for QUnit 2 compat.

rwjblue:
/cc @thoov


thoov:
Tagging #15988 so it shows up over there

PR #16072, opened 2018-01-05
[CLEANUP] Convert ember-metal/watching tests to new style

thoov:
Apart of #15988

### DIFF
--- a/packages/ember-metal/tests/watching/is_watching_test.js
+++ b/packages/ember-metal/tests/watching/is_watching_test.js
@@ -8,8 +8,7 @@ import {
   removeObserver,
   isWatching
 } from '../..';
-
-QUnit.module('isWatching');
+import { moduleFor, AbstractTestCase } from 'internal-test-helpers';
 
 function testObserver(assert, setup, teardown, key = 'key') {
   let obj = {};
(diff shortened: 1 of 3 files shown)


packages/internal-test-helpers/lib/module-for.js
```diff
@@ -62,6 +63,8 @@ export default function moduleFor(description, TestClass, ...mixins) {
   function generateTest(name) {
     if (name.indexOf('@test ') === 0) {
       QUnit.test(name.slice(5), assert => context[name](assert));
+    } else if (name.indexOf('@testBoth ') === 0) {
```
thoov: @rwjblue Im not sure about  `@testBoth` but Im not sure how we want to do this conceptually since we are trying to write "one" test that behind the scenes translates to 2.

I was thinking of something like:

```js
['@test foo'](assert) {
  const done = assert.async();

  testBoth('foo', (get, set, assert) {

  }, done);
}
```

But its going to create:

```js
QUnit.test('foo', () => {
  QUnit.test('foo 1', () => {})
  QUnit.test('foo 2', () => {})
});
```

And I dont think that is right either

rwjblue: In other places we have extracted the tests into a single class, then used getters for the "swappable" bits.

something like:

```js
class SomeStuffTest extends AccessorTests {
  ['@test does something or other'](assert) {
    let { get, set } = this.accessors;
    // normal test here;
  }
}

moduleFor('asdfasdf', class ThisGetAccessorTest extends SomeStuffTest {
  get accessors() {
    return { get: Ember.get, set: Ember.set };
  }
}

moduleFor('asdfasdf', class ThisGetAccessorTest extends SomeStuffTest {
  get accessors() {
    return { get(key) { return obj.get(key); }, set(key) { return obj.set(key); } };
  }
}
```

packages/internal-test-helpers/lib/module-for.js
```diff
@@ -41,9 +43,13 @@ export default function moduleFor(description, TestClass, ...mixins) {
   applyMixins(TestClass, mixins);
 
   let proto = TestClass.prototype;
-
   while (proto !== Object.prototype) {
-    Object.keys(proto).forEach(generateTest);
+    if(proto instanceof ObjectAccessorTestCase && !ENV.USES_ACCESSORS) {
```
thoov: @rwjblue I converted over to your example. This is probably the only "weirdness" and needed [to support this functionality](https://github.com/thoov/ember.js/blob/2908e947ad9a8053ef9a2e468a0e0923fbb888d4/packages/internal-test-helpers/lib/test-groups.js#L19-L25). Let me know your thoughts on this.

rwjblue: Do we ever set that ENV flag? I wonder why it exists? 🤔

thoov: I just grepped and can't find any usage of it nor can I find when this was added when looking at the git history. Now looking at this deeper seems like testBoth has never been "testing" both and instead just doing dummy asserts on the non ember metal side. Do you think we should remove this behavior or attempt to support it?

rwjblue: Kill it with 🔥

thoov: simpsons-fire

thoov:
@rwjblue I have updated this and it no longer uses the testboth method and I will refactor the other tests away from that in separate PRs.

thoov:
I have added a line item for testBoth removal

rwjblue (approved):

Issue #16058, opened 2018-01-05
[QUEST] Making jQuery Optional


wycats:
This year, we're going all-in on stripping down Ember to its bare essentials.

The first target in that effort is making jQuery optional, so applications that aren't using jQuery don't need to include the bulk. Successfully making jQuery optional cuts out around 30kb of minified/gzipped code, so it's a very juicy target (it's around 15% of the total size of an Ember app's initial vendor.js!)

@rwjblue has done a lot of work so far on this front, and as of 28da357496089a34e5a59638e5d60220bd2b9d7d, the `ember-application` package tests clean without jQuery. Now it's time to finish the job.

## Testing Ember

If this is your first time working on Ember, you should follow these steps:

1. Make sure you're on the latest version of Node or the latest LTS of Node
2. Update to the latest version of Yarn
3. Clone https://github.com/emberjs/ember.js.git
4. Run `yarn` to populate your `node_modules`
5. Run `yarn start`
6. Go to `http://localhost:4200/tests/index.html?hidepassed` in your browser.
7. Edit packages in your editor and refresh the browser. Rinse, repeat.

## Testing Ember Without jQuery

To test Ember without jQuery, you add in the `jquery=none` query parameter when running tests.

If you do that, you'll get a bunch of failures, so we're working on removing the dependency on jQuery one package at a time.

To test a single package without jQuery, run the tests by hitting this URL:

```
http://localhost:4200/tests/index.html?hidepassed&jquery=none&package=ember-glimmer
```

## Making Progress

In order to make progress on removing the jQuery dependency, you need to satisfy two criteria:

1. Fewer tests fail in `jquery=none` mode
2. No tests fail without `jquery=none`

Making progress in this way means that incremental work can always be merged into master, because there are no regressions when using Ember the way it's always been used.

## Current Status and Work Items

I (@wycats) got the `ember-glimmer` package down to 20 failing tests on the `no-jquery` branch of https://github.com/wycats/ember.js.

However, I violated Rule 2 under "making progress" and broke a bunch of tests when running jQuery.

So some things you could work on:

- [ ] Getting the tests passing again in normal mode in `ember-glimmer`
- [ ] Getting any remaining tests passing with `jquery=none` in `ember-glimmer`
- [ ] Identify which packages are already passing without jQuery on master
  - [ ] ember
  - [ ] ember-application
  - [ ] ember-console
  - [ ] ember-debug
  - [ ] ember-extension-support
  - [ ] ember-glimmer
[...]

PR #16075, opened 2018-01-06
remove jquery from ember-testing

snewcomer:
Edited: The path forward now is to warn the end user that if jQuery is disabled, they should use instead `@ember/test-helpers`. 

References  #16058

### DIFF
--- a/packages/ember-testing/lib/events.js
+++ b/packages/ember-testing/lib/events.js
@@ -1,5 +1,6 @@
-import { jQuery } from 'ember-views';
 import { run } from 'ember-metal';
+import { assign } from 'ember-utils';
+import isFormControl from './helpers/-is-form-control';
 
 const DEFAULT_EVENT_OPTIONS = { canBubble: true, cancelable: true };
 const KEYBOARD_EVENT_TYPES = ['keydown', 'keypress', 'keyup'];
@@ -7,19 +8,25 @@ const MOUSE_EVENT_TYPES = ['click', 'mousedown', 'mouseup', 'dblclick', 'mouseen
 
 export function focus(el) {
   if (!el) { return; }
-  let $el = jQuery(el);
-  if ($el.is(':input, [contenteditable=true]')) {
-    let type = $el.prop('type');
+  if (el.isContentEditable || isFormControl(el)) {
+    let type = el.getAttribute('type');
     if (type !== 'checkbox' && type !== 'radio' && type !== 'hidden') {
       run(null, function() {
-        // Firefox does not trigger the `focusin` event if the window
-        // does not have focus. If the document doesn't have focus just
-        // use trigger('focusin') instead.
+        let browserIsNotFocused = document.hasFocus && !document.hasFocus();
 
-        if (!document.hasFocus || document.hasFocus()) {
-          el.focus();
-        } else {
-          $el.trigger('focusin');
+        // makes `document.activeElement` be `element`. If the browser is focused, it also fires a focus event
+        el.focus();
+      
+        // Firefox does not trigger the `focusin` event if the window
+        // does not have focus. If the document does not have focus then
+        // fire `focusin` event as well.
+        if (browserIsNotFocused) {
+          // if the browser is not focused the previous `el.focus()` didn't fire an event, so we simulate it
+          fireEvent(el, 'focus', {
+            bubbles: false,
+          });
+      
+          fireEvent(el, 'focusin');
         }
       });
     }
@@ -43,7 +50,7 @@ export function fireEvent(element, type, options = {}) {
       clientX: x,
       clientY: y
     };
-    event = buildMouseEvent(type, jQuery.extend(simulatedCoordinates, options));
+    event = buildMouseEvent(type, assign(simulatedCoordinates, options));
   } else {
     event = buildBasicEvent(type, options);
   }
@@ -52,8 +59,16 @@ export function fireEvent(element, type, options = {}) {
 
 function buildBasicEvent(type, options = {}) {
   let event = document.createEvent('Events');
-  event.initEvent(type, true, true);
-  jQuery.extend(event, options);
+
+  // Event.bubbles is read only
+  let bubbles = options.bubbles !== undefined ? options.bubbles : true;
+  let cancelable = options.cancelable !== undefined ? options.cancelable : true;
+
+  delete options.bubbles;
+  delete options.cancelable;
+
+  event.initEvent(type, bubbles, cancelable);
+  assign(event, options);
   return event;
 }
 
@@ -61,7 +76,7 @@ function buildMouseEvent(type, options = {}) {
   let event;
   try {
     event = document.createEvent('MouseEvents');
-    let eventOpts = jQuery.extend({}, DEFAULT_EVENT_OPTIONS, options);
+    let eventOpts = assign({}, DEFAULT_EVENT_OPTIONS, options);
     event.initMouseEvent(
       type,
       eventOpts.canBubble,
@@ -88,7 +103,7 @@ function buildKeyboardEvent(type, options = {}) {
   let event;
   try {
     event = document.createEvent('KeyEvents');
-    let eventOpts = jQuery.extend({}, DEFAULT_EVENT_OPTIONS, options);
+    let eventOpts = assign({}, DEFAULT_EVENT_OPTIONS, options);
     event.initKeyEvent(
       type,
       eventOpts.canBubble,
--- a/packages/ember-testing/lib/helpers/fill_in.js
+++ b/packages/ember-testing/lib/helpers/fill_in.js
@@ -2,6 +2,7 @@
 @module ember
 */
 import { focus, fireEvent } from '../events';
+import isFormControl from './-is-form-control';
 
 /**
   Fills in an input element with some text.
@@ -32,7 +33,12 @@ export default function fillIn(app, selector, contextOrText, text) {
   el = $el[0];
   focus(el);
 
-  $el.eq(0).val(text);
+  if (isFormControl(el)) {
+    el.value = text;
+  } else {
+    el.innerHTML = text;
+  }
+
   fireEvent(el, 'input');
   fireEvent(el, 'change');
 
--- a/packages/ember-testing/lib/helpers/find.js
+++ b/packages/ember-testing/lib/helpers/find.js
@@ -2,6 +2,8 @@
 @module ember
 */
 import { get } from 'ember-metal';
+import { assert } from 'ember-debug';
+import { jQueryDisabled } from 'ember-views';
 
 /**
   Finds an element in the context of the app's container element. A simple alias
@@ -20,13 +22,16 @@ import { get } from 'ember-metal';
   ```
 
   @method find
-  @param {String} selector jQuery string selector for element lookup
+  @param {String} selector jQuery selector for element lookup
   @param {String} [context] (optional) jQuery selector that will limit the selector
                             argument to find only within the context's children
-  @return {Object} jQuery object representing the results of the query
+  @return {Object} DOM element representing the results of the query
   @public
 */
 export default function find(app, selector, context) {
+  if (jQueryDisabled) {
+    assert('If jQuery is disabled, please import and use helpers from @ember/test-helpers [https://github.com/emberjs/ember-test-helpers]. Note: `find` is not an available helper.');
+  }
   let $el;
   context = context || get(app, 'rootElement');
   $el = app.$(selector, context);
--- a/packages/ember-testing/lib/helpers/find_with_assert.js
+++ b/packages/ember-testing/lib/helpers/find_with_assert.js
@@ -22,7 +22,7 @@
   @param {String} [context] (optional) jQuery selector that will limit the
   selector argument to find only within the context's children
   @return {Object} jQuery object representing the results of the query
-  @throws {Error} throws error if jQuery object returned has a length of 0
+  @throws {Error} throws error if object returned has a length of 0
   @public
 */
 export default function findWithAssert(app, selector, context) {
--- a/packages/ember-testing/lib/support.js
+++ b/packages/ember-testing/lib/support.js
@@ -43,7 +43,7 @@ if (environment.hasDOM && !jQueryDisabled) {
         $.event.special.click = {
           // For checkbox, fire native event so checked state will be right
           trigger() {
-            if ($.nodeName(this, 'input') && this.type === 'checkbox' && this.click) {
+            if (this.nodeName === 'INPUT' && this.type === 'checkbox' && this.click) {
               this.click();
               return false;
             }
--- a/packages/ember-testing/tests/helpers_test.js
+++ b/packages/ember-testing/tests/helpers_test.js
@@ -9,10 +9,10 @@ import {
   RSVP
 } from 'ember-runtime';
 import { run } from 'ember-metal';
-import { jQuery } from 'ember-views';
 import {
   Component,
 } from 'ember-glimmer';
+import { jQueryDisabled } from 'ember-views';
 
 import Test from '../test';
 import setupForTesting from '../setup_for_testing';
@@ -20,7 +20,8 @@ import setupForTesting from '../setup_for_testing';
 import {
   pendingRequests,
   incrementPendingRequests,
-  clearPendingRequests
+  decrementPendingRequests,
+  clearPendingRequests,
 } from '../test/pending_requests';
 import {
   setAdapter,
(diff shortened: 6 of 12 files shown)


packages/ember-testing/lib/events.js
```diff
@@ -1,26 +1,17 @@
...
       run(null, function() {
-        // Firefox does not trigger the `focusin` event if the window
-        // does not have focus. If the document doesn't have focus just
-        // use trigger('focusin') instead.
-
-        if (!document.hasFocus || document.hasFocus()) {
-          el.focus();
-        } else {
-          $el.trigger('focusin');
-        }
```
snewcomer:
```
if (browserIsNotFocused) {
   // if the browser is not focused the previous `el.focus()` didn't fire an event, so we simulate it
   fireEvent(element, 'focus', {
     bubbles: false,
   });

   fireEvent(element, 'focusin');
}
```

Ah I prob need to add this line in there.

ro0gr:
> If we force the user to refactor

I believe we can't do this. At least I don't know any RFC with this discussed.
I think we have to preserve `jquery` support in the `ember-testing` package till the end of its days.
 
@rwjblue isn't `@ember/test-helpers` supposed to replace `ember-testing` package eventually?

snewcomer:
Another potential hairy spot - Listening for `ajaxSend`/`ajaxComplete` and triggering that event are easy with JS, but I believe one difficulty is getting all the event listeners [here](https://github.com/emberjs/ember.js/blob/master/packages/ember-testing/tests/helpers_test.js#L183) (across multiple browsers) with JS.  jQuery makes this nice and easy.  JS equivalent is `getEventListeners`, but I think that only works w/ Chrome Firefox. && Safari.

Could create our own data structure to hold these events instead.... :thinking:

packages/ember-testing/lib/events.js
```diff
@@ -1,26 +1,28 @@
...
-        // does not have focus. If the document doesn't have focus just
-        // use trigger('focusin') instead.
+        el.focus();
 
-        if (!document.hasFocus || document.hasFocus()) {
-          el.focus();
-        } else {
-          $el.trigger('focusin');
-        }
+        let browserIsNotFocused = document.hasFocus && !document.hasFocus();
```
snewcomer: Looks like there is a better, but similar sol'n [here](https://github.com/emberjs/ember.js/pull/16056/files#diff-f5610c59ca1ca80522629d553a3415c9R57)

snewcomer (commented):
Two items left include: 

1.  `find` - replace jquery with JS
2.  Bring back those tests with some data structure that holds / resets `ajaxSend` && `ajaxComplete`

packages/ember-testing/lib/helpers/find_with_assert.js
```diff
@@ -17,12 +17,12 @@
...
   @method findWithAssert
-  @param {String} selector jQuery selector string for finding an element within
+  @param {String} selector css selector string for finding an element within
   the DOM
-  @param {String} [context] (optional) jQuery selector that will limit the
+  @param {String} [context] (optional) css selector that will limit the
   selector argument to find only within the context's children
-  @return {Object} jQuery object representing the results of the query
-  @throws {Error} throws error if jQuery object returned has a length of 0
+  @return {Object} DOM element representing the results of the query
```
snewcomer: This has yet to be implemented.  `find` is still using jquery.  If we merge w/o the `find` change, I can revert these files.

packages/ember-testing/lib/events.js
```diff
@@ -1,25 +1,32 @@
...
+    let type = el.getAttribute('type');
     if (type !== 'checkbox' && type !== 'radio' && type !== 'hidden') {
       run(null, function() {
+        let browserIsNotFocused = document.hasFocus && !document.hasFocus();
+
         // Firefox does not trigger the `focusin` event if the window
         // does not have focus. If the document doesn't have focus just
         // use trigger('focusin') instead.
+        if (browserIsNotFocused) {
+          fireEvent(el, 'focusin');
```
rwjblue: FWIW, I totally realize that this isn't a mistake in your migrate to native from jQuery, and that it was preexisting.  During my testing of this functionality while migrating it to ember-test-helpers I don't think this is quite right, it should be something akin to this:

```js
  // makes `document.activeElement` be `element`. If the browser is focused, it also fires a focus event
  element.focus();

  // Firefox does not trigger the `focusin` event if the window
  // does not have focus. If the document does not have focus then
  // fire `focusin` event as well.
  if (browserIsNotFocused) {
    // if the browser is not focused the previous `el.focus()` didn't fire an event, so we simulate it
    fireEvent(element, 'focus', {
      bubbles: false,
    });

    fireEvent(element, 'focusin');
  }
```

Otherwise the event ordering is incorrect...
  

See [related implementation](https://github.com/emberjs/ember-test-helpers/blob/9836a34b782024e62e76321750f41bf2da44ee5d/addon-test-support/%40ember/test-helpers/dom/focus.js#L12-L27) from `@ember/test-helpers`.

packages/ember-testing/lib/events.js
```diff
@@ -88,7 +96,8 @@ function buildKeyboardEvent(type, options = {}) {
   let event;
   try {
     event = document.createEvent('KeyEvents');
-    let eventOpts = jQuery.extend({}, DEFAULT_EVENT_OPTIONS, options);
+    let eventOpts = assign({}, DEFAULT_EVENT_OPTIONS);
+    assign(eventOpts, options);
```
rwjblue: `Ember.assign` is a rough polyfill for `Object.assign` and does properly accept multiple arguments.  Can you update to:

```js
let eventOpts = assign({}, DEFAULT_EVENT_OPTIONS, options);
```

packages/ember-testing/lib/helpers/fill_in.js
```diff
@@ -32,7 +32,7 @@ export default function fillIn(app, selector, contextOrText, text) {
   el = $el[0];
   focus(el);
 
-  $el.eq(0).val(text);
+  el.value = text;
```
rwjblue: Can you confirm if `$el.eq(0).val(text)` supported content editable elements? I don't recall if it did, but if it did then this would be a regression (since when content editable set you need to set `innerHTML`).

See [related code](https://github.com/emberjs/ember-test-helpers/blob/master/addon-test-support/@ember/test-helpers/dom/fill-in.js#L39-L43) from `@ember/test-helper`.

packages/ember-testing/lib/support.js
```diff
@@ -19,13 +19,14 @@ const $ = jQuery;
...
 function testCheckboxClick(handler) {
   let input = document.createElement('input');
-  $(input)
-    .attr('type', 'checkbox')
-    .css({ position: 'absolute', left: '-1000px', top: '-1000px' })
-    .appendTo('body')
-    .on('click', handler)
-    .trigger('click')
-    .remove();
+  input.setAttribute('type', 'checkbox');
```
rwjblue: I think this should stay back in jQuery mode since its only ran behind a `!jQueryDisabled` guard

packages/ember-testing/tests/helpers_test.js
```diff
@@ -179,83 +185,83 @@ moduleFor('ember-testing: Helper setup', class extends HelpersTestCase {
...
-    );
-
-    this.application.injectTestHelpers();
-
-    assert.equal(
-      injected, 1,
-      'onInjectHelpers are called after injectTestHelpers'
-    );
-  }
+  // [`@test #setupForTesting attaches ajax listeners`](assert) {
```
rwjblue: Can you bring these tests back, but put a short circuit in when jQuery is disabled?

something like:

```js
['@test asdfasdf'](assert) {
  if (jQueryDisabled) {
    assert.expect(0);
    return;
  }
}
```

rwjblue (changes requested):

packages/ember-testing/lib/helpers/fill_in.js
snewcomer: Seems that both jquery and JS.  And yeah def need to set that innerHTML! :+1:  

```
el = document.createElement('div')
$(el).attr('contenteditable', 'true')
$(el).val('wat')
$(el).text() // or $(el).html() // ""
```

packages/ember-testing/lib/events.js
snewcomer: Getting these ie11 errors after refactoring to this. 

https://travis-ci.org/emberjs/ember.js/jobs/326476166

Looks like it is triggering the events multiple times and causing pain `click` triggering events in order.

packages/ember-testing/tests/helpers_test.js
snewcomer: So this one might be a bit difficult to get at the event listeners w/ `documentEvents = jQuery._data(document, 'events');` if jquery is enabled since it is now using `document.addEventListener` in setup-for-testing.  The only thing I was thinking was to track the events with a data structure of our own.  Thoughts?

packages/ember-testing/lib/helpers/find.js
```diff
@@ -29,6 +30,15 @@ import { get } from 'ember-metal';
 export default function find(app, selector, context) {
   let $el;
   context = context || get(app, 'rootElement');
-  $el = app.$(selector, context);
-  return $el;
+  if (jQueryDisabled) {
```
snewcomer: So to allow this to work in all cases, `find` needs to accept an element or selector...not just a selector.

packages/ember-testing/lib/helpers/find.js
```diff
@@ -29,6 +30,15 @@ import { get } from 'ember-metal';
...
+    let rootElement = document.querySelector(context);
+    $el = rootElement.querySelector(selector);
```
rwjblue: This should be returned, (we don't need to wrap in an array)

packages/ember-testing/lib/helpers/find.js
```diff
@@ -29,6 +30,15 @@ import { get } from 'ember-metal';
...
+    if ($el) {
+      return [$el];
+    }
+    return [];
+  } else {
+    $el = app.$(selector, context);
```
rwjblue: Can probably return this directly

packages/ember-testing/lib/helpers/find.js
rwjblue: Also, I think we should do:

```js
  let result;
  if (context instanceof Element) {
    result = context.querySelector(selector);
  } else if (typeof context === 'string') {
    result = document.querySelector(`${context} ${selector}`);
  } else {
    result = document.querySelector(selector);
  }
  return result;
```

packages/ember-testing/lib/helpers/click.js
```diff
@@ -10,14 +10,14 @@ import { focus, fireEvent } from '../events';
...
-  click('.some-jQuery-selector').then(function() {
+  click('.some-css-selector').then(function() {
     // assert something
   });
   ```
 
   @method click
-  @param {String} selector jQuery selector for finding element on the DOM
-  @param {Object} context A DOM Element, Document, or jQuery to use as context
+  @param {String} selector css selector for finding element on the DOM
```
simonihmig: Although the functionality has not changed and still uses jquery when it's available, this docs change could be misleading, suggesting there is a breaking change. Shouldn't this state something like `jQuery selector or css selector (if jQuery is not present) for finding element on the DOM`?

packages/ember-testing/lib/helpers/find.js
snewcomer: Fixed! :taco:   

So to support both jquery and JS, I think we need to return an array or array-like object from find.  Other helpers are expecting this (like findWithAssert).  Unless we wanted to go with only returning one element; however, perhaps users were using `find` to select multiple as show [here](https://github.com/emberjs/ember.js/blob/master/packages/ember-testing/tests/integration_test.js#L74).

packages/ember-testing/lib/helpers/click.js
simonihmig: Oh, seems you changed it already while I wrote this!? :)

snewcomer: That seems like a good option.  I have it now as `@param {String|Element} target the element or selector to click on`.  What do you think of this?  Should it be:

`the element or jQuery selector or css selector (if jQuery is not preset) to click on`

packages/ember-testing/tests/helpers_test.js
```diff
@@ -96,1168 +103,1186 @@ class HelpersApplicationTestCase extends HelpersTestCase {
...
-      'there are no ajaxComplete listers setup prior to calling injectTestHelpers'
-    );
+      assert.equal(
+        this.application._readinessDeferrals, 1,
+        'App is in deferred state after setupForTesting.'
+      );
+    }
 
-    setupForTesting();
+    // [`@test #setupForTesting attaches ajax listeners`](assert) {
```
snewcomer: If we want to bring back these tests, we would have to add back the `jquery.on` listeners in `setup_for_testing`.  I don't think there is a way to track all of the events listening in a cross browser way (`getEventListeners`; however, I think no IE).   We could add an explicit trackEventLIsteners method but I don't think we would want to do that either.
 
However, it seems to me that the below tests cover the functionality in `setup_for_testing` and these commented out tests could be deleted.

https://github.com/emberjs/ember.js/blob/master/packages/ember-testing/tests/helpers_test.js#L1074

snewcomer (commented):
I think almost done!  Lmk what you think about the conditionals I put around the tests.

packages/ember-testing/lib/helpers/click.js
simonihmig: Seems good to me

packages/ember-testing/lib/helpers/click.js
```diff
@@ -17,7 +17,7 @@ import { focus, fireEvent } from '../events';
 
   @method click
   @param {String} selector jQuery selector for finding element on the DOM
-  @param {Object} context A DOM Element, Document, or jQuery to use as context
+  @param {Object} context A DOM Element, Document, or css to use as context
```
snewcomer: need to revert this

packages/ember-testing/lib/helpers/find_with_assert.js
```diff
@@ -21,8 +21,8 @@
   the DOM
   @param {String} [context] (optional) jQuery selector that will limit the
   selector argument to find only within the context's children
-  @return {Object} jQuery object representing the results of the query
-  @throws {Error} throws error if jQuery object returned has a length of 0
+  @return {Object} DOM element representing the results of the query
```
snewcomer: need to revert this.

packages/ember-testing/tests/integration_test.js
```diff
@@ -52,39 +53,55 @@ moduleFor('ember-testing Integration tests of acceptance', class extends Autoboo
...
+          rows, 0,
+          'successfully stubbed an empty array of people'
+        );
+      });
+    } else {
+      this.runTask(() => this.application.advanceReadiness());
+      window.visit('/').then(() => {
+        expectAssertion(() => window.find('.name'), 
+        /If not using jQuery, please import and use methods from @ember\/test-helpers/);
+      });
```
snewcomer: This asserts that `assert` is thrown

packages/ember-testing/tests/helpers_test.js
snewcomer: @rwjblue lmk what you think about this remaining issue.

rwjblue: @snewcomer - Can you delete them?

rwjblue:
I left one inline comment, and other than that we need to squash down the commits into one and this should be good to go...

packages/ember-testing/lib/helpers/find.js
```diff
@@ -20,13 +22,16 @@ import { get } from 'ember-metal';
...
+  @param {String} selector jQuery selector for element lookup
   @param {String} [context] (optional) jQuery selector that will limit the selector
                             argument to find only within the context's children
-  @return {Object} jQuery object representing the results of the query
+  @return {Object} DOM element representing the results of the query
   @public
 */
 export default function find(app, selector, context) {
+  if (jQueryDisabled) {
+    assert('If not using jQuery, please import and use methods from @ember/test-helpers');
```
ro0gr: which helper exactly can be used? I think `@ember/test-helpers` doesn't provide any.

rwjblue: @ro0gr - Usages of `find` should be replaced by `this.element.querySelector` or `this.element.querySelectorAll`...

ro0gr: but the old good acceptance tests don't have `this.element`. Should this assertion have a reference to `setupTest` api?

rwjblue: FWIW - By the time this hits a release version, the guides will all have been updated to reference the new style (work progressing in https://github.com/emberjs/guides/pull/2145).

For now, lets update the assertion to tell the user that `find` is not available when jQuery is disabled without giving a more detailed recourse (we can tweak the message when we have better docs to cross link to)...

What do y'all think?

snewcomer:
@rwjblue It looks like the tests are passing but for a focus-in test in `ember-glimmer`.

rwjblue (approved):

Issue #16079, opened 2018-01-06
tryInvoke, canInvoke, compare, isEqual missing updated imports in docs


jenweber:
According to [RFC176](https://github.com/emberjs/rfcs/blob/master/text/0176-javascript-module-api.md) these should be part of the `@ember/utils` module, but API docs do not reflect this. I think they just got overlooked.

Documentation block [here](https://github.com/emberjs/ember.js/blob/241adea495be3363dfc4c09a749e73230fa80998/packages/ember-utils/lib/invoke.js)

Should follow the same pattern as [this example](https://github.com/emberjs/ember.js/blob/v2.18.0/packages/ember-metal/lib/is_blank.js#L5) with a class and fixed method

Update ought to target 2.18 LTS candidate and later

jenweber:
https://github.com/emberjs/ember.js/pull/16080

locks:
Addressed by PR mentioned above.

PR #16080, opened 2018-01-06
[BUGFIX] missing modules docs for tryInvoke, compare, isEqual #16079

jenweber:
See https://github.com/emberjs/ember.js/issues/16079. This should target at least 2.18 LTS candidate and 3.0-beta. 2.16 LTS would be good too. I have no idea how any of that is done so let me know if I need to open more PRs against other branches.

Needs a good look by @toddjordan 

Thank you to sukima for reporting!

### DIFF
--- a/packages/ember-runtime/lib/compare.js
+++ b/packages/ember-runtime/lib/compare.js
@@ -38,6 +38,10 @@ function spaceship(a, b) {
 }
 
 /**
+ @module @ember/utils
+*/
+
+/**
  Compares two javascript values and returns:
 
   - -1 if the first is smaller than the second,
@@ -72,7 +76,8 @@ function spaceship(a, b) {
   ```
 
  @method compare
- @for Ember
+ @for @ember/utils
+ @static
  @param {Object} v First value to compare
  @param {Object} w Second value to compare
  @return {Number} -1 if v < w, 0 if v = w and 1 if v > w.
--- a/packages/ember-runtime/lib/is-equal.js
+++ b/packages/ember-runtime/lib/is-equal.js
@@ -1,4 +1,7 @@
 /**
+ @module @ember/utils
+*/
+/**
   Compares two objects, returning true if they are equal.
 
   ```javascript
@@ -30,7 +33,8 @@
   ```
 
   @method isEqual
-  @for Ember
+  @for @ember/utils
+  @static
   @param {Object} a first object to compare
   @param {Object} b second object to compare
   @return {Boolean}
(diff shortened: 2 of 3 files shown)


jenweber:
The test suite passes locally.

CI fails at yarn install.
```error An unexpected error occurred: "https://registry.yarnpkg.com/pinkie-promise/-/pinkie-promise-2.0.1.tgz: Request failed \"404 Not Found\"".```

**Comment by jenweber** [comment]
Also missing `compare` which should belong to `@ember/utils`

[docs here](https://github.com/emberjs/ember.js/blob/5f4ac602813d8fbd78f26fc955acf693e0b3d62a/packages/ember-runtime/lib/mixins/comparable.js)

File: packages/ember-runtime/lib/is-equal.js [hunk 1]
```diff
@@ -30,7 +33,7 @@
   ```
 
   @method isEqual
```
toddjordan: Also needs an `@static` to show up as a function

packages/ember-runtime/lib/compare.js
```diff
@@ -72,7 +76,7 @@ function spaceship(a, b) {
   ```
 
  @method compare
```
toddjordan: needs an `@static`

packages/ember-utils/lib/invoke.js
```diff
@@ -12,7 +15,7 @@ import applyStr from './apply-str';
   ```
 
   @method canInvoke
```
toddjordan: `@static`

packages/ember-utils/lib/invoke.js
```diff
@@ -35,7 +38,7 @@ export function canInvoke(obj, methodName) {
   ```
 
   @method tryInvoke
```
toddjordan: `@static`

jenweber:
Thanks, I'll get those changes made

packages/ember-utils/lib/invoke.js
jenweber: resolved

packages/ember-utils/lib/invoke.js
jenweber: I removed canInvoke - turns out it's not a package/module which makes sense because it is private API

packages/ember-runtime/lib/compare.js
jenweber: resolved

packages/ember-runtime/lib/is-equal.js
jenweber: resolved

jenweber:
@toddjordan I added static to everything and double-checked that these methods are actually available in the utils module/package. `canInvoke` wasn't, so I reverted those changes.

toddjordan:
Tried it out.  Looks good!  Thanks @jenweber !!!

toddjordan (approved):

PR #16084, opened 2018-01-07
[BUGFIX release] Update `computed.sort` docs to avoid state leakage

rwjblue:


jenweber:
For reference, I checked through the Guides and Tutorials and I think we're good. No use of `computed.sort` in the first place.

PR #16085, opened 2018-01-07
[CLEANUP] upgrade to ember-cli 2.18.0

GavinJoyce:
I'm unsure if there is a reason why we're using ember-cli 2.10.0 or not. This PR upgrades it to 2.18.0 if there isn't

### DIFF
--- a/node-tests/blueprints/route-test.js
+++ b/node-tests/blueprints/route-test.js
@@ -11,6 +11,7 @@ const setupPodConfig = blueprintHelpers.setupPodConfig;
 const chai = require('ember-cli-blueprint-test-helpers/chai');
 const expect = chai.expect;
 const file = chai.file;
+const fs = require('fs-extra');
 
 describe('Blueprint: route', function() {
   setupTestHooks(this);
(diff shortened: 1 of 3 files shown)


GavinJoyce:
Seems to be failing due to a duplicate file:

screen shot 2018-01-07 at 19 57 55

GavinJoyce:
`ember new` now creates an `app/templates/application.hbs` file which is causing `ember generate route application` to hang on a `Overwrite app/templates/application.hbs` prompt

rwjblue:
We can probably avoid that by changing the name of the route we generate?  I don't think that `application` was a terribly intentional choice...

GavinJoyce:
I think it was intentional, it seems to be testing that an `application` route isn't added to `router.js`:

```js
it('route application', function() {
  return emberGenerate(['route', 'application']).then(() => {
    expect(file('app/router.js')).to.not.contain('this.route(\'application\')');
  });
});
```

GavinJoyce:
I'll just remove `app/templates/application.hbs` before running the command

rwjblue:
Good idea 👍

GavinJoyce:
CI test failure seems to be flakiness

rwjblue:
restarted browserstack

locks (approved):

PR #16087, opened 2018-01-08
[BUGFIX beta] Ensure `App.visit` resolves when rendering completed. 

### LINKED ISSUE #16059: Fix visit() promise resolution
The fix in this PR makes the issue worse, it is now less likely render is fully resolved when the visit promise is resolved.

https://github.com/emberjs/ember.js/pull/14591

This issue causes a lot of tests to force flush the run loop to ensure visit is done rendering.

To clarify, this issue isn't about error handling, it is about actually resolving the promise only when rendering is actually settled.

rwjblue:
Creates a new internal method (named `renderSettled`) that returns a promise which will resolve when rendering has settled. Settled in this context is defined as when all of the tags in use are "current" (e.g.  `renderers.every(r => r._isValid())`). When this is checked at the _end_ of  the run loop, this essentially guarantees that all rendering is completed.

Prior to this change, rendering was not guaranteed to have been completed before the promise returned from the `visit` API had resolved (yes, even though `afterRender` queue was being used). With these changes no additional delay is added, but we have still ensured that all rendering is completed...

Fixes https://github.com/emberjs/ember.js/issues/16059

### DIFF
--- a/packages/ember-application/lib/system/application-instance.js
+++ b/packages/ember-application/lib/system/application-instance.js
@@ -3,11 +3,11 @@
 */
 
 import { assign } from 'ember-utils';
-import { get, set, run, computed } from 'ember-metal';
-import { RSVP } from 'ember-runtime';
+import { get, set, computed } from 'ember-metal';
 import { environment } from 'ember-environment';
 import { jQuery } from 'ember-views';
 import EngineInstance from './engine-instance';
+import { renderSettled } from 'ember-glimmer';
 
 /**
   The `ApplicationInstance` encapsulates all of the stateful aspects of a
@@ -243,14 +243,8 @@ const ApplicationInstance = EngineInstance.extend({
         // No rendering is needed, and routing has completed, simply return.
         return this;
       } else {
-        return new RSVP.Promise((resolve) => {
-          // Resolve once rendering is completed. `router.handleURL` returns the transition (as a thennable)
-          // which resolves once the transition is completed, but the transition completion only queues up
-          // a scheduled revalidation (into the `render` queue) in the Renderer.
-          //
-          // This uses `run.schedule('afterRender', ....)` to resolve after that rendering has completed.
-          run.schedule('afterRender', null, resolve, this);
-        });
+        // Ensure that the visit promise resolves when all rendering has completed
+        return renderSettled().then(() => this);
       }
     };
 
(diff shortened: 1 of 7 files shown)


krisselden (approved):

packages/ember-application/lib/system/application-instance.js
```diff
@@ -243,14 +243,8 @@ const ApplicationInstance = EngineInstance.extend({
...
-        return new RSVP.Promise((resolve) => {
-          // Resolve once rendering is completed. `router.handleURL` returns the transition (as a thennable)
-          // which resolves once the transition is completed, but the transition completion only queues up
-          // a scheduled revalidation (into the `render` queue) in the Renderer.
-          //
-          // This uses `run.schedule('afterRender', ....)` to resolve after that rendering has completed.
-          run.schedule('afterRender', null, resolve, this);
-        });
+        // Ensure that the visit promise resolves when all rendering has completed
+        return renderSettled().then(() => this);
```
snewcomer: @rwjblue Sorry to bring up something old, but seeing some memory leaks in fastboot related to retaining the Container.  Is it necessary to return `this` in the callback?

What I'm worries about is never reaching a `settled` state or resolving, thus every request that comes in builds a new instance (w/ shoebox data) and the eventually swamps the server's heap.

snewcomer: On 3.5 right now

snewcomer: Screenshot 2019-04-12 at 13 58 39

some more info.  I'm wondering if we need to "settle" the deferred promise when `shouldRender` is false....So not this block but the above block.

rwjblue: I don't really follow. Can you file an issue with a reproduction?

snewcomer: @rwjblue - We found it.  Well @Serabe found it so I'll let him explain beyond this short answer! In an adapter, he found that `headers` as a computed was potentially leaking the container thus not allowing V8 to not gc anything.  The problem wasn't necessarily ajax/najax/node-fetch but the way we configured our adapter. 

```js
// adapter
headers: computed(function() {
```

to a getter fixed it.

```js
// adapter
get headers() {
```

We are on 3.5

PR #16092, opened 2018-01-08
[FEATURE ember-glimmer-named-arguments] update reserve names

chancancode:
Reserve some more names used in the named blocks RFC

I would like to avoid reserving `@main` if we could, but I am not sure if we have enough consensus on that (the proposal is to replace `@main` with `@block`).


rwjblue (approved):

chancancode:
Update: I discussed with @wycats @mmun and @machty and we agreed to _not_ reserve `@main`

PR #16093, opened 2018-01-09
[packages/ember/tests/routing/decoupled_basic_test.js] some work on migrating tests - part 1

NullVoxPopuli:


NullVoxPopuli:
is node 6 still supported? looks like it's taking a while to install on travis.

rwjblue:
Yep, it is a supported node LTS version.

packages/ember/tests/routing/decoupled_basic_test.js
```diff
@@ -507,4 +507,37 @@ class extends ApplicationTestCase {
...
   }
 
+
+  ['@test The Specials Page defaults to looking models up via `find`'](assert) {
+    this.router.map(function() {
+      this.route('home', { path: '/' });
+      this.route('special', { path: '/specials/:menu_item_id' });
+    });
+    //
+    // this.applicationInstance.MenuItem = EmberObject.extend();
```
NullVoxPopuli: @mixonic / @rwjblue, what's the proper way to convert this sort of thing? 

Currently, I get the error:
`Error: Assertion Failed: You used the dynamic segment menu_item_id in your route special, but undefined.MenuItem did not exist and you did not override your route's `model` hook.`

not sure what undefined.MenuItem is / where it's coming from.

mixonic: @NullVoxPopuli thanks for taking this work on.

I would expect the commented section here would be written as:

```js
let MenuItem = EmberObject.extend();
MenuItem.reopenClass({
  find(id) {
    return MenuItem.create({id});
  }
});
this.add('model:menu-item', MenuItem);
```

packages/ember/tests/routing/decoupled_basic_test.js
```diff
@@ -326,4 +327,217 @@ class extends ApplicationTestCase {
...
+
+    this.addTemplate('alternative_home', '<p>alternative home: {{myValue}}</p>');
+
+
+    return this.visit('/').then(() => {
+      let homeRoute = this.applicationInstance.lookup('route:home');
+      let myController = this.applicationInstance.lookup('controller:myController');
+      let text = this.$('p').text();
+
+      assert.deepEqual(homeRoute.controller, myController,
```
mixonic: This was `deepEqual` in the old test, but you almost certainly want `equal` to confirm they are the *same* instance. Please use `equal`.

I'm surprised this works with `myController` instead of `my-controller`. I expected the test mode resolver would fail with camelcase!

packages/ember/tests/routing/decoupled_basic_test.js
```diff
@@ -326,4 +327,217 @@ class extends ApplicationTestCase {
...
+    this.add('controller:myController', Controller.extend({
+      myValue: 'myController'
+    }));
+
+    return this.visit('/').then(() => {
+      let homeRoute = this.applicationInstance.lookup('route:home');
+      let myController = this.applicationInstance.lookup('controller:myController');
+      let text = this.$('p').text();
+
+      assert.deepEqual(homeRoute.controller, myController,
```
mixonic: Again, `equal`.

packages/ember/tests/routing/decoupled_basic_test.js
```diff
@@ -326,4 +327,217 @@ class extends ApplicationTestCase {
...
+  }
+
+  [`@test The Homepage with a 'setupController' hook modifying other controllers`](assert) {
+    this.router.map(function() {
+      this.route('home', { path: '/' });
+    });
+
+    this.add('route:home', Route.extend({
+      setupController(/* controller */) {
+        set(this.controllerFor('home'), 'hours', emberA([
```
mixonic: Please just use

```js
this.controllerFor('home').set('hours', // ...
```

The `Ember.set` thing is just weird.

mixonic: For that matter, the use of `emberA([` instead of simply a normal array is unnecessary.

packages/ember/tests/routing/decoupled_basic_test.js
```diff
@@ -326,4 +327,217 @@ class extends ApplicationTestCase {
...
+
+    return this.visit('/').then(() => {
+      let text = this.$('ul li:nth-child(3)').text();
+
+      assert.equal(text, 'Sunday: Noon to 6pm',
+        'The template was rendered with the hours context');
+    });
+  }
+
+  [`@test The Homepage with a computed context that does not get overridden`](assert) {
```
mixonic: please change the test to say `computed model that` instead of `context`, since `model` is the property with the CP.

packages/ember/tests/routing/decoupled_basic_test.js
```diff
@@ -326,4 +327,217 @@ class extends ApplicationTestCase {
...
+          'Monday through Friday: 9am to 5pm',
+          'Saturday: Noon to Midnight',
+          'Sunday: Noon to 6pm'
+        ]);
+      },
+
+      setupController(controller, model) {
+        assert.equal(this.controllerFor('home'), controller);
+
+        set(this.controllerFor('home'), 'hours', model);
```
mixonic: `this.controllerFor('home').set(`

packages/ember/tests/routing/decoupled_basic_test.js
```diff
@@ -326,4 +327,217 @@ class extends ApplicationTestCase {
...
+  }
+
+  [`@test The Homepage with a computed context that does not get overridden`](assert) {
+    this.router.map(function() {
+      this.route('home', { path: '/' });
+    });
+
+    this.add('controller:home', Controller.extend({
+      model: computed(function() {
+        return emberA([
```
mixonic: Don't need `emberA(`

packages/ember/tests/routing/decoupled_basic_test.js
```diff
@@ -326,4 +327,217 @@ class extends ApplicationTestCase {
...
+  }
+
+  [`@test The Homepage getting its controller context via model`](assert) {
+    this.router.map(function() {
+      this.route('home', { path: '/' });
+    });
+
+    this.add('route:home', Route.extend({
+      model() {
+        return emberA([
```
mixonic: Don't need `emberA`

packages/ember/tests/routing/decoupled_basic_test.js
```diff
@@ -326,4 +327,217 @@ class extends ApplicationTestCase {
...
+
+    this.add('route:special',  Route.extend({
+      model(params) {
+        return EmberObject.create({
+          menuItemId: params.menu_item_id
+        });
+      },
+
+      setupController(controller, model) {
+        set(controller, 'model', model);
```
mixonic: `controller.set`

packages/ember/tests/routing/decoupled_basic_test.js
```diff
@@ -326,4 +327,217 @@ class extends ApplicationTestCase {
...
+      }
+    }));
+
+    this.addTemplate('special', '<p>{{model.menuItemId}}</p>');
+
+    return this.visit('/').then(() => {
+      this.add('controller:special', Controller.extend());
+
+      return this.visit('/specials/1').then(() => {
```
mixonic: Chain promises, don't nest. For example:

```js
  this.add('controller:special', Controller.extend());
  return this.visit('/specials/1');
}).then(() => {
  let test = this.$('p').text();
  assert.equal(text, '1',
    'the model was used to render the template'
  );
});
```

packages/ember/tests/routing/decoupled_basic_test.js
```diff
@@ -326,4 +327,217 @@ class extends ApplicationTestCase {
...
+    //   find(id) {
+    //     return this.applicationInstance.MenuItem.create({
+    //       id: id
+    //     });
+    //   }
+    // });
+
+    this.add('route:main', Route.extend({
+      setupController(controller, model) {
+        set(controller, 'model', model);
```
mixonic: `controller.set`

mixonic: Actually you can just remove this route definition. This is the default implementation.

packages/ember/tests/routing/decoupled_basic_test.js
mixonic: Actually you can just remove `setupController`, this is the default implementation.

packages/ember/tests/routing/decoupled_basic_test.js
mixonic: Actually I think you can just remove `this.add('controller:special', Controller.extend());` here.

I don't see the relevance here of visiting `/` before the route that is being tested either. IMO you can drop the route `this.route('home', { path: '/' });` above then just make this:

```js
return this.visit('/specials/1').then(() => {
  let test = this.$('p').text();
  assert.equal(text, '1',
    'the model was used to render the template'
  );
});
```

mixonic (changes requested):
Thanks for working on this @NullVoxPopuli! Some feedback for you.

packages/ember/tests/routing/decoupled_basic_test.js
NullVoxPopuli: thanks! (and I agree)

packages/ember/tests/routing/decoupled_basic_test.js
NullVoxPopuli: done

packages/ember/tests/routing/decoupled_basic_test.js
NullVoxPopuli: still getting 
```
Error: Assertion Failed: You used the dynamic segment menu_item_id in your route special, but undefined.MenuItem did not exist and you did not override your route's `model` hook.
``` 

:-\

mixonic (approved):

mixonic:
@NullVoxPopuli Thank you for this work! Great stuff :-)

PR #16096, opened 2018-01-09
allow ember-glimmer tests to run w/o jquery

ro0gr:
part of #16058

there are two notable changes:

-  fixed bug with event bubbling in the non-jquery event-dispatcher + action handler.  This bug was detected by [this test](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/tests/integration/helpers/element-action-test.js#L445-L476) in no-jquery mode.
The problem was that we stopped event bubbling once we found the first action handler caused action handler [always returned](https://github.com/emberjs/ember.js/compare/master...ro0gr:ember-glimmer-no-jq?expand=1#diff-4a9f781ec9f748d565c7e148b1ef1f89L166) `false` despite the `bubbles` argument.
 - an assertion for `jquery` availability was added to the `ViewSupport.$()`. The assertion message is the following: "_You cannot access this.$() with `jQuery` disabled_". I'm glad to improve wording here.


ro0gr:
according to browserstack there is a remaining failing test on IE "triggers `focus-in` when focused". Looking into it

rwjblue:
We should address that test, it only fails sporadically.

rwjblue:
Thank you!

PR #16097, opened 2018-01-09
[CLEANUP beta] Remove `sync` runloop queue.

rwjblue:
`sync` queue was only used/present for bindings to synchronize, and Ember.Binding (along with related code) is now completely gone.


packages/ember-metal/lib/run_loop.js
```diff
@@ -16,11 +12,7 @@ function onEnd(current, next) {
   run.currentRunLoop = next;
 }
 
-const backburner = new Backburner(['sync', 'actions', 'destroy'], {
-  sync: {
```
krisselden: this functionality in backburner was only for the sync queue

krisselden (approved):

packages/ember-metal/lib/run_loop.js
stefanpenner: ya, we can remove it from BB and then cut a major

stefanpenner (approved):
LGTM :fire: :fire: :fire:

packages/ember-metal/lib/run_loop.js
rwjblue: Ya, we will have to wait for a short while (because the legacy addon still needs to enable this functionality through Ember 3.4), but totally agree that it will be awesome to get rid of 😆

PR #16098, opened 2018-01-09
Assert for forgetting .get() on an Ember proxy

chancancode:


packages/ember-metal/lib/property_get.js
```diff
@@ -13,6 +14,23 @@ const ALLOWABLE_TYPES = {
...
+export const PROXY_CONTENT = symbol('PROXY_CONTENT');
+
+export function getPossibleMandatoryProxyValue(obj, keyName) {
+  if (MANDATORY_GETTER && EMBER_METAL_ES5_GETTERS && HAS_NATIVE_PROXY) {
+    let content = obj[PROXY_CONTENT];
+    if (content === undefined) {
+      return obj[keyName];
+    } else {
+      /* global Reflect */
+      return Reflect.get(content, keyName, obj);
```
rwjblue: Do we know of platforms that have native `Proxy` but don't support `Reflect`?

chancancode: Doesn't look like it: http://kangax.github.io/compat-table/es6/#test-Reflect (the opposite could be true though)

rwjblue:
@chancancode - Where are we at with this?

chancancode:
Probably need a cleanup and polish pass. Cleaning up the code/message/test and stuff. @wycats said he may be able to finish it this weekend.

rwjblue (approved):

PR #16100, opened 2018-01-10
Support @only => QUnit.only

thoov:
Allows @only to be used for specifying a single test to be run via QUnit.only.


chancancode:
😀👍

Issue #15723, opened 2017-10-10
Update documentation to link to new URL API docs structure


locks:
So for each package in ember source, we'd like to convert api doc links to our new  url structure (example: https://emberjs.com/api/ember/release/classes/Application)

Additionally we'd like to convert references to old Ember namespaced classes to their simplified post RFC 176 names.  
(example: documentation might refer to Ember.Object, while we'd want it to be EmberObject now)
See the readme for https://github.com/ember-cli/ember-rfc176-data for the mappings.

here's a list of all subdirectories in the `packages/` directory.  There may or may not be things to convert in some of these.  if you verify these are not in need of conversion we can check them too.

Indicate in comments when you want to work on a directory and its yours. We will add your name to the item.  Also when you submit a PR, mention this issue "#15723" so that the link shows up here.

Thanks and Happy coding!

- [x] container :lock: @Parrryy               
- [x] ember-glimmer :lock: @Parrryy           
- [x] ember-views :lock: @Parrryy 
- [x] ember  :lock: @MarcoUmpierrez     
- [x] ember-metal :lock: @karthiick        
- [x] external-helpers :lock: @Parrryy 
- [x] ember-application  🔒 @Parrryy 
- [x] ember-routing  🔒 @ynotdraw      
- [x]  internal-test-helpers 🔒 @ynotdraw  
- [x] ember-console 🔒 @ynotdraw            
- [x] ember-runtime 🔒 @ynotdraw            
- [x] loader 🔒 @ynotdraw 
- [x] ember-debug   🔒 @williamhaley           
- [x] ember-template-compiler 🔒 @ynotdraw 
- [x] node-module 🔒 @ynotdraw 
- [x] ember-environment 🔒 @ynotdraw 
- [x] ember-testing 🔒 @ynotdraw 
- [x] ember-extension-support 🔒 @ynotdraw 
- [x] ember-utils 🔒 @ynotdraw 
  
  
**Update:**  Be sure to prepend your commits to `[doc beta]`, so we can get these changes into 3.0!
See https://github.com/emberjs/ember.js/blob/master/CONTRIBUTING.md#documentation

scissorsneedfoodtoo:
Hi there. I'd like to help out with this task, but I'm quite new to ember.js and contribution. Could you point me toward the documentation that needs to be updated and the new URL structure?

juventinoromero:
Hi! Is this issue taken? I can help!

toddjordan:
[...]

PR #16108, opened 2018-01-11
Update docs to new import syntax and use the new release URL

ynotdraw:
#15723 here are the four items I picked from the issue.  I went through the jsdocs and replaced old usages with the new import syntax as well.  Let me know if this is overkill/should not be included here.  Those changes are the large majority of what changed.

### DIFF
--- a/packages/ember-runtime/lib/ext/function.js
+++ b/packages/ember-runtime/lib/ext/function.js
@@ -72,7 +72,7 @@ if (ENV.EXTEND_PROTOTYPES.Function) {
     will instead clear the cache so that it is updated when the next `get`
     is called on the property.
 
-    See [Ember.ComputedProperty](/api/classes/Ember.ComputedProperty.html), [Ember.computed](/api/classes/Ember.computed.html).
+    See [ComputedProperty](/api/ember/release/classes/ComputedProperty), [@ember/object/computed](/api/ember/release/classes/@ember%2Fobject%2Fcomputed).
 
     @method property
     @for Function
--- a/packages/ember-runtime/lib/utils.js
+++ b/packages/ember-runtime/lib/utils.js
@@ -33,9 +33,12 @@ const { toString } = Object.prototype;
   not formally an array but appears to be array-like (i.e. implements `Ember.Array`)
 
   ```javascript
-  Ember.isArray();                                          // false
-  Ember.isArray([]);                                        // true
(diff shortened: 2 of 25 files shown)


packages/ember-runtime/lib/ext/function.js
```diff
@@ -72,7 +72,7 @@ if (ENV.EXTEND_PROTOTYPES.Function) {
     will instead clear the cache so that it is updated when the next `get`
     is called on the property.
 
-    See [Ember.ComputedProperty](/api/classes/Ember.ComputedProperty.html), [Ember.computed](/api/classes/Ember.computed.html).
+    See [ComputedProperty](/api/ember/release/classes/ComputedProperty), [Object.computed](/api/ember/release/classes/@ember%2Fobject%2Fcomputed).
```
toddjordan: I'd call it `[@ember/object/computed]` vs `[Object.computed]`

packages/ember-runtime/lib/utils.js
```diff
@@ -86,26 +89,30 @@ export function isArray(obj) {
...
+  typeOf(101);                    // 'number'
+  typeOf(new Number(101));        // 'number'
+  typeOf(true);                   // 'boolean'
+  typeOf(new Boolean(true));      // 'boolean'
+  typeOf(A);                      // 'function'
+  typeOf([1, 2, 90]);             // 'array'
+  typeOf(/abc/);                  // 'regexp'
+  typeOf(new Date());             // 'date'
+  typeOf(event.target.files);     // 'filelist'
+  Ember.typeOf(EmberObject.extend());  // 'class'
```
toddjordan: typeOf for these 2 lines too?

toddjordan:
This is awesome. Thanks!

packages/ember-runtime/lib/utils.js
ynotdraw: Doh! Good catch!

toddjordan:
Looks Great!  Thanks @ynotdraw !

toddjordan:
@ynotdraw do you mind squashing these commits into one?  Its easier to cherry pick into the beta branch that way...

ynotdraw:
Dangit - I messed up my squash as it pulled in other commits. Fixing again here in a few.

mikemunsie:
@toddjordan Just for my learning as well, couldn't we  just use the `merge + squash` feature from github? Or do you guys prefer manually squashing the commits and then setting up a PR?

PR #16113, opened 2018-01-11
Update docs for RFC 176 syntax and canonical URL

williamhaley:
#15723 Update `ember-debug` package to use rfc176 imports/naming conventions and update doc URLs to use `../release/..` for canonical path

### DIFF
--- a/packages/ember-debug/tests/main_test.js
+++ b/packages/ember-debug/tests/main_test.js
@@ -55,27 +55,27 @@ moduleFor('ember-debug', class extends TestCase {
     ENV._ENABLE_DEPRECATION_OPTIONS_SUPPORT = originalDeprecationOptions;
   }
 
-  ['@test Ember.deprecate does not throw if RAISE_ON_DEPRECATION is false'](assert) {
+  ['@test deprecate does not throw if RAISE_ON_DEPRECATION is false'](assert) {
     assert.expect(1);
 
     ENV.RAISE_ON_DEPRECATION = false;
 
     try {
       deprecate('Should not throw', false, { id: 'test', until: 'forever' });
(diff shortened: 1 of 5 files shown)


packages/ember-debug/tests/main_test.js
```diff
@@ -55,27 +55,27 @@ moduleFor('ember-debug', class extends TestCase {
     ENV._ENABLE_DEPRECATION_OPTIONS_SUPPORT = originalDeprecationOptions;
   }
 
-  ['@test Ember.deprecate does not throw if RAISE_ON_DEPRECATION is false'](assert) {
+  ['@test deprecate does not throw if RAISE_ON_DEPRECATION is false'](assert) {
```
williamhaley: Updating names in tests wasn't in the scope original issue, but I figured I could update them while I was in there. I'm happy to revert this though.

toddjordan: its what I was trying to say here, so yes, you are on the right track:

> Additionally we'd like to convert references to old Ember namespaced classes to their simplified post RFC 176 names.
> (example: documentation might refer to Ember.Object, while we'd want it to be EmberObject now)
> See the readme for https://github.com/ember-cli/ember-rfc176-data for the mappings.

toddjordan:
This looks great.  Thanks, @williamhaley !

toddjordan:
@williamhaley Could you reword the commit prepend [doc beta], so we can get this into 3.0?
See https://github.com/emberjs/ember.js/blob/master/CONTRIBUTING.md#documentation
(I'll also update the issue)

williamhaley:
@toddjordan commit tag updated 👍

PR #16120, opened 2018-01-13
[DOC release] Mark MutableEnumerable as private

mmun:
This model is mistakenly marked public. It should be marked private like its sibling `Enumerable`.


rwjblue:
Should be `[DOC release]`...

PR #16129, opened 2018-01-16
[BUGFIX beta] Migrate to using execa and pipe to stdout for publishing.

### LINKED ISSUE #16128: Build of v3.0.0-beta.3 tag published as `latest` dist tag
Build https://travis-ci.org/emberjs/ember.js/jobs/329226750

```
$ npm info ember-source

{ name: 'ember-source',
  description: 'A JavaScript framework for creating ambitious web applications',
  'dist-tags': { latest: '3.0.0-beta.3', beta: '3.0.0-beta.2', lts: '2.12.2' },
....
```

rwjblue:
The prior `execSync` seemed to work properly, however it obscured the output which makes it very hard to debug when there is an issue.

The migration to `execa` also fixes a bug (likely the cause of the recent invalid publishing of a beta as latest). Previously we were using `execSync('auto-dist-tag -w')` which would _only_ work if `auto-dist-tag` were on $PATH, but unfortunately we installed it into `node_modules/.bin`. The console output of this was lost (due to not properly piping to STDOUT), so we will need to confirm with another beta...

Also, add `set -e` to prevent publish on prior errors...

Fixes https://github.com/emberjs/ember.js/issues/16128

### DIFF
--- a/bin/build-for-publishing.js
+++ b/bin/build-for-publishing.js
@@ -1,10 +1,19 @@
 #!/usr/bin/env node
 'use strict';
+/* eslint-env node, es6 */
 
 const fs = require('fs');
 const path = require('path');
-const execSync = require('child_process').execSync;
+const execa = require('execa');
 const VERSION = require('../broccoli/version').VERSION;
+
+function exec(command, args) {
+  // eslint-disable-next-line
+  console.log(`\n\tRunning: \`${command} ${args.join(' ')}\``);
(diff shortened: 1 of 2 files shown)


locks (approved):
Looks good to me!

kategengler (approved):
👍 assuming node files run from bash that return promises exit non-zero when the promise rejects.

PR #16130, opened 2018-01-16
remove jquery from ember pkg

snewcomer:
Down to a few tests.

- [ ]  nested routes and link-to arguments: The {{link-to}} helper supports bubbles=false
- [ ]  nested routes and link-to arguments: The {{link-to}} helper supports bubbles=boundFalseyThing

Something is happening with the click handler

References #16058


packages/ember/tests/integration/multiple-app-test.js
```diff
@@ -4,17 +4,15 @@ import {
...
 import { resolve } from 'rsvp';
 
 moduleFor('View Integration', class extends ApplicationTestCase {
 
   constructor() {
-    jQuery('#qunit-fixture').html(`
-      <div id="one"></div>
-      <div id="two"></div>
-    `);
+    document.getElementById('qunit-fixture').innerHTML =
```
rwjblue: We have a specialized system in place for this (added in https://github.com/emberjs/ember.js/commit/ff9117a1d647c7bf098a44b881289d546b972fe7). You should be able to do:

```js
class extends ApplicationTestCase {
  get fixture() {
    return `<div id="one"></div><div id="two"></div>`;
  }
}
```

rwjblue (changes requested):

snewcomer:
@rwjblue Looking at the `link_to_test`, with jQuery off, do you happen to know if anything changes wrt `bubbles=false`?  Passes w/ jQuery, but fails w/o jQuery.  And the only change is to a find selector that passes.

rwjblue:
> do you happen to know if anything changes wrt bubbles=false?

No, I can't think of one off the top of my head...

snewcomer:
@rwjblue looks like this line is undefined so it jumps down to the parentNode and passes that to the eventListener.

https://github.com/emberjs/ember.js/blob/master/packages/ember-views/lib/system/event_dispatcher.js#L299

Starts here in the `view_support` mixin, eventually calling down into ember-metal.

https://github.com/emberjs/ember.js/blob/master/packages/ember-views/lib/mixins/view_support.js#L456

Failing to figure out the exact problem.  Seems at this point it may be test setup since I see other tests not using jquery and handling bubbling correctly.

ro0gr:
@snewcomer I've experienced a similar issue when removed `jquery` from the `ember-glimmer` package. The fix was to [return bubble value](https://github.com/emberjs/ember.js/pull/16096/commits/ad6bc61a32da05db600a36f9563bd15456c9fd9a#diff-4a9f781ec9f748d565c7e148b1ef1f89R168) from the action handler.

I've made a pr against your branch with a similar fix (https://github.com/snewcomer/ember.js/pull/1)
@rwjblue can you confirm if that is proper fix?

ro0gr:
> I've made a pr

This change broke 2 other tests:

 - EventDispatcher: events bubbling up can be prevented
 - View tree tests: getChildViews

Closed the pr. Sorry for the false positive 😫

ro0gr:
@snewcomer @rwjblue I believe this is a more correct approach https://github.com/snewcomer/ember.js/pull/2

rwjblue:
Hmm, @ro0gr I was thinking the issue was regarding [this return value](https://github.com/emberjs/ember.js/blob/d190957863dc95e2cdde88ca70d56992278c61e1/packages/ember-glimmer/lib/components/link-to.ts#L651) in `LinkComponent#_invoke` (roughly similar to the fix here https://github.com/emberjs/ember.js/pull/16096/commits/ad6bc61a32da05db600a36f9563bd15456c9fd9a).

I think that return from `_invoke` should become:

```js
let shouldBubble = get(this, 'bubbles') !== false;
if (!shouldBubble) { event.stopPropagation(); }
// ...snip...
return shouldBubble;
```

ro0gr:
@rwjblue unfortunately changing of `_invoke` to return a proper `bubbling` value doesn't fix 2 `link-to` tests from the description.

As far as I can see `_invoke` is [only used](https://github.com/emberjs/ember.js/blob/d190957863dc95e2cdde88ca70d56992278c61e1/packages/ember-glimmer/lib/components/link-to.ts#L506) as an event handler. I think it makes impossible to consume `_invoke`'s return value.

ro0gr:
So I'm out of ideas on how to fix `link-to`'s `bubbles` behavior. 

@rwjblue @snewcomer What do you think about skipping these 2 `link-to` tests in non-`jquery` mode for this last package?

snewcomer:
@rwjblue I looked at a few PR's @ro0gr put up and I still can't figure out a sol'n.  Any further ideas?

snewcomer:
closing for now.

PR #16132, opened 2018-01-16
Bring back `sync` queue with deprecation (until: 3.5.0).

rwjblue:
Removing the `sync` queue was done in https://github.com/emberjs/ember.js/pull/16097 and https://github.com/emberjs/ember.js/pull/16110. Unfortunately, addon authors have reported this as a regression in the 3.0.0 beta cycle (due to it not having been previously deprecated).

This reverts the removal, and adds a private API deprecation scheduled for removal in 3.5.0.

### DIFF
--- a/packages/ember-metal/lib/run_loop.js
+++ b/packages/ember-metal/lib/run_loop.js
@@ -2,6 +2,10 @@ import { assert, isTesting } from 'ember-debug';
 import {
   onErrorTarget
 } from './error_handler';
+import {
+  beginPropertyChanges,
+  endPropertyChanges
+} from './property_events';
 import Backburner from 'backburner';
 
 function onBegin(current) {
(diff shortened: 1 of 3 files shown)


vothaison:
Hi @rwjblue 

I don't see Ember.run.sync()  in (and since) v3.0.0; Where have it gone?

By the way, we have this code:
```
change: function (event) {
        this.set('data', 'SOME VALUE');
        Ember.run.sync();
        this.sendAction('on-change', event, this.item);
    },
```
The purpose is to sendAction after run.sync, to make sure that any properties depending on this 'data' have settled before we fire the change action (which may rely on 'data').

Can we change it to 
```
change: function (event) {
        this.set('data', this.parse(this.get('value')));
        Ember.run.next(this, function () {
            this.sendAction('on-change', event, this.item);
        });
    },
```
Everything is still "synced" before sendAction, right?

Thanks.

PR #16137, opened 2018-01-17
use es6 classes for `QueryParams` and `RouterState`

bekzod:


packages/ember-routing/lib/index.d.ts
```diff
@@ -192,11 +192,12 @@ export const Route: {
   teardownViews(): void;
   isRouteFactory: boolean;
 };
+
 export const QueryParams: {
```
locks: I think this needs to get turned into a `class`, TypeScript is complaining.

stefanpenner: ohoh, someone has angered TypeScript

locks: 😂 😂

PR #16144, opened 2018-01-18
[TESTS es-classes] Adds a failing test for didReceiveAttrs

pzuraq:
Adds a failing test for didReceiveAttrs in native classes. Class fields
will likely be the prefered way to initialize default values on class instances,
but because didReceiveAttrs is called in the base class constructor it is
initially called before class fields have had an opportunity to be setup.

cc @wycats


mmun:
Closed by #16185.

PR #16154, opened 2018-01-20
RFC 176 doc updates for ember-metal package

karthiicksiva:
#15723

### DIFF
--- a/packages/ember-metal/lib/map.js
+++ b/packages/ember-metal/lib/map.js
@@ -2,7 +2,7 @@ import { assert } from 'ember-debug';
 import { guidFor } from 'ember-utils';
 
 /**
-@module ember
+@module @ember/map
 */
 
 /*
@@ -217,7 +217,6 @@ class OrderedSet {
   we delete its entry in `keys` and `values`.
 
   @class Map
-  @namespace Ember
   @private
   @constructor
 */
@@ -262,7 +261,7 @@ class Map {
     @method set
     @param {*} key
     @param {*} value
-    @return {Ember.Map}
+    @return {Map}
     @private
   */
   set(key, value) {
@@ -363,7 +362,7 @@ class Map {
 
   /**
     @method copy
-    @return {Ember.Map}
+    @return {Map}
     @private
   */
   copy() {
@@ -373,8 +372,7 @@ class Map {
 
 /**
   @class MapWithDefault
-  @namespace Ember
-  @extends Ember.Map
+  @extends Map
   @private
   @constructor
   @param [options]
@@ -391,7 +389,7 @@ class MapWithDefault extends Map {
     @static
     @param [options]
       @param {*} [options.defaultValue]
-    @return {Ember.MapWithDefault|Ember.Map} If options are passed, returns
+    @return {MapWithDefault|Map} If options are passed, returns
       `MapWithDefault` otherwise returns `EmberMap`
     @private
   */
@@ -425,7 +423,7 @@ class MapWithDefault extends Map {
 
   /**
     @method copy
-    @return {Ember.MapWithDefault}
+    @return {MapWithDefault}
     @private
   */
   copy() {
--- a/packages/ember-metal/lib/run_loop.js
+++ b/packages/ember-metal/lib/run_loop.js
@@ -140,14 +140,14 @@ run.join = function() {
     initializeTinyMCE: on('didInsertElement', function() {
       tinymce.init({
         selector: '#' + this.$().prop('id'),
-        setup: Ember.run.bind(this, this.setupEditor)
+        setup: bind(this, this.setupEditor)
       });
     }),
 
     didInsertElement() {
       tinymce.init({
         selector: '#' + this.$().prop('id'),
-        setup: Ember.run.bind(this, this.setupEditor)
+        setup: bind(this, this.setupEditor)
       });
     }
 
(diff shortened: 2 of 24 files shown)


karthiick:
@locks I am not sure about how to update the reference to ember-runtime package link. 

```
  At the heart of Ember is Ember-Runtime, a set of core functions that provide
  cross-platform compatibility and object property observing.  Ember-Runtime is
  small and performance-focused so you can use it alongside other
  cross-platform libraries such as jQuery. For more details, see
  [Ember-Runtime](https://emberjs.com/api/modules/ember-runtime.html).

```

https://github.com/emberjs/ember.js/blob/v2.18.0/packages/ember-metal/lib/core.js#L16

After 2.15 version of the docs, it seems ember-runtime removed (https://emberjs.com/api/ember/2.15/modules/ember-runtime). How to proceed?

packages/ember-metal/lib/map.js
```diff
@@ -186,7 +186,7 @@ class OrderedSet {
 
   /**
     @method copy
-    @return {Ember.OrderedSet}
+    @return {OrderedSet}
     @private
   */
   copy() {
```
toddjordan: on line 220, we still have `@namespace` for Map.  We should remove this line so Map shows up as `Map` in the docs and not `Ember.Map`

packages/ember-metal/lib/map.js
```diff
@@ -63,7 +63,7 @@ class OrderedSet {
   /**
     @method create
     @static
-    @return {Ember.OrderedSet}
+    @return {OrderedSet}
```
toddjordan: According to https://github.com/ember-cli/ember-rfc176-data/blob/master/mappings.json its still `Ember.OrderedSet`

packages/ember-metal/lib/map.js
```diff
@@ -63,7 +63,7 @@ class OrderedSet {
   /**
```
toddjordan: We need to make the `@module` .   `@ember/map` and not `Ember`

packages/ember-metal/lib/run_loop.js
```diff
@@ -161,7 +161,7 @@ run.join = function() {
   });
   ```
 
-  In this example, we use Ember.run.bind to bind the setupEditor method to the
+  In this example, we use `@ember/runloop/bind` to bind the setupEditor method to the
```
toddjordan: I think you can just use `bind`

packages/ember-metal/lib/watching.js
```diff
@@ -21,7 +21,7 @@ import {
   invokes `Ember.propertyWillChange` and `Ember.propertyDidChange`. This is the
   primitive used by observers and dependent keys; usually you will never call
   this method directly but instead use higher level methods like
-  `Ember.addObserver()`
+  `@ember/object/observers/addObserver`
```
toddjordan: Just `addObserver` should do.

packages/ember-metal/tests/accessors/get_path_test.js
```diff
@@ -3,7 +3,7 @@ import { moduleFor, AbstractTestCase } from 'internal-test-helpers';
 
 let obj;
 
-moduleFor('Ember.get with path', class extends AbstractTestCase {
+moduleFor('@ember/object/get with path', class extends AbstractTestCase {
```
toddjordan: `get with path`

packages/ember-metal/tests/accessors/get_properties_test.js
```diff
@@ -1,7 +1,7 @@
 import { getProperties } from '../..';
 import { moduleFor, AbstractTestCase } from 'internal-test-helpers';
 
-moduleFor('Ember.getProperties', class extends AbstractTestCase {
+moduleFor('@ember/object/getProperties', class extends AbstractTestCase {
```
toddjordan: `getProperties`

packages/ember-metal/tests/is_blank_test.js
```diff
@@ -1,8 +1,8 @@
 import { isBlank } from '..';
 
-QUnit.module('Ember.isBlank');
+QUnit.module('@ember/utils/isBlank');
```
toddjordan: `isBlank`

packages/ember-metal/tests/is_empty_test.js
```diff
@@ -4,9 +4,9 @@ import {
   OrderedSet
 } from '..';
 
-QUnit.module('Ember.isEmpty');
+QUnit.module('@ember/utils/isEmpty');
 
-QUnit.test('Ember.isEmpty', function(assert) {
+QUnit.test('@ember/utils/isEmpty', function(assert) {
```
toddjordan: `isEmpty`

toddjordan:
Generally, I'd like for functions like `isEmpty` to be referred to by name alone or module separate from name, like "`isEmpty` from '@ember/utils'". Using `@ember/utils/isEmpty` mixes path with function and is unclear/inaccurate.

toddjordan:
@karthiick we should delete that block of documentation for `Ember`

karthiick:
@toddjordan I have made the changes. Let me know, if i missed something.

packages/ember-metal/lib/map.js
```diff
@@ -2,7 +2,7 @@ import { assert } from 'ember-debug';
 import { guidFor } from 'ember-utils';
 
 /**
-@module ember
+@module ember/map
```
toddjordan: Change to `@ember/map` and we are good

toddjordan:
@karthiick this is great.  I commented on just one more change and then I will merge.

karthiick:
@toddjordan Done!

toddjordan:
Looks great! Will merge when the CI passes.  Thanks @karthiick !!

PR #16158, opened 2018-01-21
Introduce namespaced component and service invocation

mixonic:
This commit adds support for invoking a component or injecting a service from a module unification namespace. It does so in a very targeted manner, and specifically *avoids* adding support for namespaces to lower-level APIs like lookup or factoryFor.

Ember itself doesn't perform any resolution steps for a lookup. Instead, it defers resolution to a resolver class provided to the application at runtime. For example this invocation:

```hbs
{{! app/templates/index.hbs }}
{{app-component}}
```

Causes a resolver invocation of:

```js
resolve(
  'template:components/app-component', // a "fullName" formatted type:name.
  'app/templates/index' // a referrer for this lookup
);
```

The resolver API has already been changed behind the module unification feature flag to pass the second argument of "referrer" when appropriate.

A component from another namespace is invoked by putting the namespace before a `::` delemiter. For example an invocation:

```hbs
{{! src/ui/components/app-component/template.hbs }}
{{my-addon::addon-component}}
```

In order to resolve this namespace the resolver must be passed `my-addon` is some form or another.

The current Ember container APIs are coupled heavily to the idea of a "fullName". This is a string formatted as `type:name`, and expected to resolve something. I'm hesitant to expand that syntax to something like `type:optional-namespace::name`, especially since we would like to replace the strings there with "specifiers" from the glimmer-di project as we adopt more of that project.

Additionally we want existing implementions of the resolver API to *not* be passed the namespace in the first argument. If they were, they may resolve certain lookups in ways that don't mesh our goals for the `::` syntax. For example a custom resolver in someone's app may be splitting on `:` an expecting a syntax in the string that matches `type:name`. It is ok to break those resolvers for namespaced lookups, and we would rather they fail in some way than resolve something that is wrong.

Instead, we've made an *additive* change to the resolver API. The "rawString", the string invoking the lookup, is passed in the third position:

```hbs
{{! src/ui/components/app-component/template.hbs }}
{{my-addon::addon-component}}
```

Resolves with:

```js
resolve(
  'template:components/',
  'src/ui/components/app-component/template',
  'my-addon::addon-component'
);
resolve(
  'component',
  'src/ui/components/app-component/template',
  'my-addon::addon-component'
);
```

Similarly service injections pass their "rawString":

```hbs
Ember.inject.service('my-addon::addon-service');
```

Resolves with:

```js
resolve(
  'service',
  undefined,
  'my-addon::addon-service'
);
```

The resolver is expected to parse the invocation string.

TODO:

* [ ] Complete a smoke test against ember-resolver master with the MU feature flag.
* [ ] Add a test for namespaced `main` lookups of services. `store: Ember.inject.service('ember-data')` and the addons `src/services/main.js` would be resolved.
* [ ] Rationalize what `moduleName` is on a template and what it should be.

### DIFF
--- a/packages/container/lib/container.js
+++ b/packages/container/lib/container.js
@@ -10,6 +10,7 @@ import {
   HAS_NATIVE_PROXY
 } from 'ember-utils';
 
+
 /**
  A container used to instantiate and cache objects.
 
@@ -223,8 +224,7 @@ function wrapManagerInDeprecationProxy(manager) {
 }
 
 function isSingleton(container, fullName, options) {
-  let res = container.registry.getOption(fullName, 'singleton', options) !== false;
-  return res;
+  return container.registry.getOption(fullName, 'singleton', options) !== false;
 }
 
 function isInstantiatable(container, fullName) {
--- a/packages/container/lib/container.js
+++ b/packages/container/lib/container.js
@@ -319,10 +319,10 @@ function buildInjections(container, injections) {
     let injection;
     for (let i = 0; i < injections.length; i++) {
       injection = injections[i];
-      let {name, rawString, type} = parseInjectionString(injection.fullName);
-      hash[injection.property] = lookupWithRawString(container, type, rawString || name);
+      let {name, namespace, type} = parseInjectionString(injection.fullName);
+      hash[injection.property] = lookupWithRawString(container, type, namespace || name);
       if (!isDynamic) {
-        isDynamic = !isSingleton(container, injection.fullName, {[RAW_STRING_OPTION_KEY]: rawString});
+        isDynamic = !isSingleton(container, injection.fullName, {[RAW_STRING_OPTION_KEY]: namespace});
       }
     }
   }
@@ -462,7 +462,7 @@ export function parseInjectionString(injectionString) {
     let fullName = type.indexOf(':') === -1 ? `${type}:${name}` : `${type}${name}`;
     return {
       fullName,
-      rawString: namespace,
+      namespace,
       type
     };
   }
@@ -485,7 +485,7 @@ export function factoryForWithRawString(container, type, rawString) {
     return container.factoryFor(`${type}:${rawString}`);
   } else {
     let [ namespace, name ] = rawString.split('::');
-    // type might already contain : eg. "template:components/"
+    // type might already contain ":" eg. "template:components/"
     let fullName = type.indexOf(':') === -1 ? `${type}:${name}` : `${type}${name}`;
     return container.factoryFor(fullName, {
       [RAW_STRING_OPTION_KEY]: namespace
--- a/packages/container/lib/container.js
+++ b/packages/container/lib/container.js
@@ -88,7 +88,7 @@ export default class Container {
    @return {any}
    */
   lookup(fullName, options) {
-    assert('fullName must be a proper full name', this.registry.isValidFullName(fullName, options));
+    assert('fullName must be a proper full name', this.registry.isValidFullName(fullName));
     return lookup(this, this.registry.normalize(fullName), options);
   }
 
@@ -148,7 +148,7 @@ export default class Container {
   factoryFor(fullName, options = {}) {
     let normalizedName = this.registry.normalize(fullName);
 
-    assert('fullName must be a proper full name', this.registry.isValidFullName(normalizedName, options));
+    assert('fullName must be a proper full name', this.registry.isValidFullName(normalizedName));
 
     if (options.source) {
       let expandedFullName = this.registry.expandLocalLookup(fullName, options);
--- a/packages/container/lib/container.js
+++ b/packages/container/lib/container.js
@@ -9,9 +9,6 @@ import {
   assign,
   HAS_NATIVE_PROXY
 } from 'ember-utils';
-import { symbol } from 'ember-utils';
-
-export const RAW_STRING_OPTION_KEY = symbol('RAW_STRING_OPTION_KEY');
 
 /**
  A container used to instantiate and cache objects.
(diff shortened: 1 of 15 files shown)


packages/container/lib/container.js
```diff
@@ -278,7 +281,7 @@ function isFactoryInstance(container, fullName, { instantiate, singleton }) {
 }
 
 function instantiateFactory(container, fullName, options) {
-  let factoryManager = EMBER_MODULE_UNIFICATION && options && options.source ? container.factoryFor(fullName, options) : container.factoryFor(fullName);
+  let factoryManager = EMBER_MODULE_UNIFICATION && options ? container.factoryFor(fullName, options) : container.factoryFor(fullName);
```
mixonic: I think we can just pass `container.factoryFor(fullName, options)` all the time.

packages/container/lib/container.js
```diff
@@ -440,3 +446,42 @@ class FactoryManager {
     return instance;
   }
 }
+
+export function parseInjectionString(injectionString) {
```
mixonic: document this

packages/container/lib/container.js
```diff
@@ -440,3 +446,42 @@ class FactoryManager {
...
+  } else {
+    return {
+      fullName: type,
+      rawString,
+      type
+    };
+  }
+}
+
+export function lookupWithRawString(container, type, rawString) {
```
mixonic: document this

packages/container/lib/container.js
```diff
@@ -440,3 +446,42 @@ class FactoryManager {
...
+  if (rawString.indexOf('::') === -1) {
+    return container.lookup(`${type}:${rawString}`);
+  } else {
+    return container.lookup(`${type}`, {
+      [RAW_STRING_OPTION_KEY]: rawString
+    });
+  }
+}
+
+export function factoryForWithRawString(container, type, rawString) {
```
mixonic: document this

packages/container/lib/index.js
```diff
@@ -9,4 +9,7 @@ export { default as Registry, privatize } from './registry';
 export {
   default as Container,
   FACTORY_FOR,
+  factoryForWithRawString,
+  lookupWithRawString,
+  RAW_STRING_OPTION_KEY as _RAW_STRING_OPTION_KEY
```
mixonic: Confirm these don't become public API

packages/container/lib/registry.js
```diff
@@ -1,12 +1,16 @@
...
+  parseInjectionString,
+  RAW_STRING_OPTION_KEY
+} from './container';
 import { DEBUG } from 'ember-env-flags';
 import { ENV } from 'ember-environment';
 
 const VALID_FULL_NAME_REGEXP = /^[^:]+:[^:]+$/;
-const missingResolverFunctionsDeprecation = 'Passing a `resolver` function into a Registry is deprecated. Please pass in a Resolver object with a `resolve` method.';
+
+let missingResolverFunctionsDeprecation = 'Passing a `resolver` function into a Registry is deprecated. Please pass in a Resolver object with a `resolve` method.';
```
mixonic: This line should remain whatever it was in master.

packages/container/lib/registry.js
```diff
@@ -546,7 +554,14 @@ export default class Registry {
     return assign({}, fallbackKnown, localKnown, resolverKnown);
   }
 
-  isValidFullName(fullName) {
+  isValidFullName(fullName, options) {
+    if(EMBER_MODULE_UNIFICATION) {
+      /* With a rawString, the fullName doesn't need to contain a : */
+      if (options && options[RAW_STRING_OPTION_KEY] && fullName) {
+        return true;
```
mixonic: This needs review with @rwjblue @dgeb I'm not quite sure what to do here. The gist is that previously we required `type:name` but this check happens at low levels and now `type` alone is permitted (the name ends up in the rawString). Perhaps we should move this check out of the deep parts of the container instead of removing it in this PR.

iezer:
@mixonic FYI I tested this with the latest `ember-resolver` and the [mu-golden-spike](https://github.com/iezer/mu-golden-spike) app and it works for both components and services. Here's a screenshot:

golden-spike 2018-01-30 15-03-00

packages/container/tests/container_test.js
```diff
@@ -724,15 +725,16 @@ if (EMBER_MODULE_UNIFICATION) {
...
+      let rawString = `${targetNamespace}::${name}`;
+      let specifier = `${type}:${name}`;
       let resolver = new ModuleBasedTestResolver();
       let registry = new Registry({resolver});
 
       resolver.add({
-        specifier: type,
-        rawString: rawString
+        specifier,
+        targetNamespace
```
mixonic: This could just be `namespace`, I don't think `target` is useful to discern it from anything else?

packages/container/lib/registry.js
iezer: @mixonic I believe type alone is no longer permitted so we can remove this.

packages/container/lib/container.js
iezer: Turns out some tests related to local lookup fail if we remove this.

iezer:
@mixonic and @dgeb We should clarify how we expect the "main" service of an addon to work. In my end-to-end test, If I have an addon called `ember-data` and I try to inject a service called `ember-data`, the glimmer resolver expects this to be in `/ember-data/src/services/service.js`, not `/ember-data/src/services/main.js`.

If I try to inject a service called `ember-data::secondary`, The resolver will find it in `/ember-data/src/services/secondary.js`.

mixonic:
Closing this in favor of https://github.com/emberjs/ember.js/pull/16319

PR #16164, opened 2018-01-22
[doc beta] fixing references to Ember in the subdirectory ember

MarcoUmpierrez:
In relation to [this issue #15723](https://github.com/emberjs/ember.js/issues/15723), I have fixed the references to Ember in the subdirectory `ember`.

There were some things I wasn't sure about; I list them here so anyone with more experience can guide me to replace them correctly:

1. [lib/index.js](https://github.com/emberjs/ember.js/blob/master/packages/ember/lib/index.js#L388): here appears `Ember.String.loc()` in a comment, but I wasn't sure if I should replace it with just "`loc()` helper" or "`loc()` helper in `@ember/string`".

2. [tests/error_handler_test.js](https://github.com/emberjs/ember.js/blob/master/packages/ember/tests/error_handler_test.js): 

- `Ember.onerror` has no replacement as far as I know. I researched it and found this [comment](https://github.com/ember-cli/ember-rfc176-data/issues/12#issuecomment-318718538) of rwjblue. The same happens with `Ember.testing`.
- I wasn't sure what to do with `Ember.Test`. Should I `import Test from 'test'`?

I think the rest should be ok, but I would appreciate if someone can check it to make sure that I didn't anything wrong 😄

### DIFF
--- a/packages/ember/lib/index.js
+++ b/packages/ember/lib/index.js
@@ -200,8 +200,10 @@ Object.defineProperty(Ember, 'LOG_BINDINGS', {
   and reporting code.
 
   ```javascript
+  import $ from 'jquery';
+
   Ember.onerror = function(error) {
-    Em.$.ajax('/report-error', 'POST', {
+    $.ajax('/report-error', 'POST', {
       stack: error.stack,
       otherInformation: 'whatever app state you want to provide'
     });
@@ -385,7 +387,7 @@ computed.collect = collect;
 
 /**
   Defines the hash of localized strings for the current language. Used by
-  the `Ember.String.loc()` helper. To localize, add string values to this
+  the `String.loc` helper. To localize, add string values to this
   hash.
 
   @property STRINGS
--- a/packages/ember/tests/error_handler_test.js
+++ b/packages/ember/tests/error_handler_test.js
@@ -1,6 +1,7 @@
 import Ember from 'ember';
 import { run } from 'ember-metal';
 import { DEBUG } from 'ember-env-flags';
+import RSVP from 'rsvp';
 
 const ONERROR = Ember.onerror;
 const ADAPTER = Ember.Test && Ember.Test.adapter;
@@ -61,7 +62,7 @@ QUnit.test('when Ember.onerror (which does not rethrow) is registered - sync run
 });
 
 if (DEBUG) {
-  QUnit.test('when Ember.Test.adapter is registered and error is thrown - sync run', function(assert) {
+  QUnit.test('when TestAdapter is registered and error is thrown - sync run', function(assert) {
     assert.expect(1);
 
     Ember.Test.adapter = {
@@ -73,7 +74,7 @@ if (DEBUG) {
     assert.throws(runThatThrowsSync, Error);
   });
 
-  QUnit.test('when both Ember.onerror (which rethrows) and Ember.Test.adapter are registered - sync run', function(assert) {
+  QUnit.test('when both Ember.onerror (which rethrows) and TestAdapter are registered - sync run', function(assert) {
     assert.expect(2);
 
     Ember.Test.adapter = {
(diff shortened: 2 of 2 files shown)


packages/ember/tests/routing/query_params_test/shared_state_test.js
```diff
@@ -2,10 +2,10 @@ import {
   Controller,
   Service
 } from 'ember-runtime';
-import Ember from 'ember';
 import { run } from 'ember-metal';
 import { jQuery } from 'ember-views';
 import { QueryParamTestCase, moduleFor } from 'internal-test-helpers';
+import { inject } from '@ember/service';
```
locks: Can you change this to `import { inject as service } from '@ember/service';`

packages/ember/tests/error_handler_test.js
```diff
@@ -1,9 +1,11 @@
 import Ember from 'ember';
 import { run } from 'ember-metal';
 import { DEBUG } from 'ember-env-flags';
+import TestAdapter from '@ember/test/adapter';
```
locks: This doesn't seem quite right, as CI is failing throughout all jobs.

locks (dismissed):
CI is complaining about `ember/tests/error_handler_test.js: "TestAdapter" is read-only`

packages/ember/tests/routing/query_params_test/shared_state_test.js
MarcoUmpierrez: @locks Done 😄

packages/ember/tests/error_handler_test.js
MarcoUmpierrez: @locks This one I wasn't completely sure because of this `Ember.Test.adapter = ADAPTER`. Maybe that's the reason why CI complains. Should I revert it?

packages/ember/tests/error_handler_test.js
```diff
@@ -20,7 +22,7 @@ QUnit.module('error_handler', {
     Ember.testing = TESTING;
     window.onerror = WINDOW_ONERROR;
     if (Ember.Test) {
-      Ember.Test.adapter = ADAPTER;
+      TestAdapter = ADAPTER;
```
toddjordan: I believe imported variables are treated as const and you can't assign to them like this.

packages/ember/tests/error_handler_test.js
toddjordan: Really updating live source code is not really the scope of #15723 and I'd be fine just updating inline docs here (and possibly assert/error messages).

packages/ember/lib/index.js
```diff
@@ -200,8 +200,10 @@ Object.defineProperty(Ember, 'LOG_BINDINGS', {
   and reporting code.
```
toddjordan: line 390 refers to `Ember.String.loc`.  Those types of refernces are what we are looking to update to `String.loc`

toddjordan:
As per your question, only mappings detailed here https://github.com/ember-cli/ember-rfc176-data/blob/master/mappings.json
should be converted

MarcoUmpierrez:
Thanks for your review @toddjordan! I've made the corrections according to your comments. I couldn't replace `Ember.inject.service`though because in Travis I got the following error for Edge, Safari and IE:

```
        Log: |
            { type: 'error',
              text: 'Could not find module @ember/service required by: 
              ember/tests/routing/query_params_test/shared_state_test at 
              http://127.0.0.1:7774/8786/dist/ember.debug.js, line 21\n' }
```

[Here](https://travis-ci.org/emberjs/ember.js/builds/333620699?utm_source=github_status&utm_medium=notification) is the link to the failed build in Travis.

toddjordan:
@MarcoUmpierrez Great!  I think this is good as is.  Like I mentioned I'm fine not changing any prod imports with this pr and sticking with the inline docs (and some messages).  

The last thing I'd ask if is if you could use git to squash the commits into one and prepend the commit comment with `[doc beta]`, so we know to make sure your updates get into 3.0 ❤️ 

Let me know if you need help with this.

MarcoUmpierrez:
@toddjordan done!

toddjordan:
Thanks @MarcoUmpierrez !!

PR #16166, opened 2018-01-22
[CLEANUP beta] Avoid before observers in ArrayProxy

mmun:
This PR removes one of two remaining internal usages of before observers (one in ArrayProxy and one in EachProxy).

In order to remain API compatible, we simulate `arrayWillChange` events by caching the values of the ArrayProxy upon accessing (e.g. when calling `objectAt` or getting the `length`). After `arrayWillChange` has been triggered we mark the cache as `dirty` and fire the `arrayDidChange` event. This will cause the cache to recompute on next access which will generally occur in response to the `arrayDidChange` event.

There are two further avenues of work:

1. It should be possible to use the indexes in the `_arrangedContentArrayDidChange` hook to optimize the syncing of the cache to only the slice that changed.
  - It's a bit of tricky code to get this right, but I should probably just bite the bullet so that things like `pushObject` don't become O(N).
2. It would be a good idea to make observation of `arrangedContent` lazy. You should only need to observe after you've accessed the value once, otherwise why would you care if it has changed?
  - When I tried implementing this it caused quite a few tests to fail in ember and ember-data. From a quick glance, most of these tests were synthetic unit tests and not indicative of how an array proxy would be used in an actual application. In any case, I prefer to make this timing change in a separate PR to make it easier to roll back if necessary.


packages/ember-runtime/lib/system/array_proxy.js
```diff
@@ -66,6 +65,17 @@ import { assert } from 'ember-debug';
   @public
 */
 export default EmberObject.extend(MutableArray, {
+  init() {
+    this._super(...arguments);
+    this._cache = null;
+    this._dirtyStart = 0;
+    this._arrangedContent = null;
+    this._addArrangedContentArrayObsever();
```
krisselden: we shouldn't need to do this until there is something to invalidate

krisselden (approved):

rwjblue (approved):

PR #16174, opened 2018-01-25
[BUGFIX lts] Enable _some_ recovery of errors thrown during render.

rwjblue:
If `env.commit()` is not called after it has started, and another `env.begin()` starts Glimmer throws an error:

```
Error: a glimmer transaction was begun, but one already exists. You may have a nested transaction
```

This commit works around the issue by ensuring that if a transaction had been started, it is cleaned up even if an error was thrown during render.

Fixes https://github.com/glimmerjs/glimmer-vm/issues/484


GavinJoyce:
I've confirmed that this restores the previous behaviour in earlier versions of Ember (for the cases in our app where we're causing exceptions)

PR #16175, opened 2018-01-25
[CLEANUP] Deprecate private propertyWillChange/propertyDidChange

mmun:
This PR removes all internal usage of private before observers. Consequently, we no longer internally need the private `propertyWillChange` and `propertyDidChange` functions. Anyone who was relying on these private functions should:

- Delete the call to `propertyWillChange` entirely.
- Replace the call to `propertyDidChange` with `notifyPropertyChange`.

Additionally, since we no longer need to differentiate between `:before` and `:change` I've removed that event suffix entirely.

---

Remaining items:

- [x] Write the deprecation guide
  - See https://github.com/emberjs/website/pull/3174


packages/ember-runtime/lib/mixins/observable.js
```diff
@@ -386,7 +386,7 @@ export default Mixin.create({
     @private
   */
   hasObserverFor(key) {
-    return hasListeners(this, `${key}:change`);
+    return hasListeners(this, `${key}`);
```
mmun: should just be `key`

rwjblue (approved):

cibernox (approved):

packages/ember-metal/lib/index.js
```diff
@@ -52,11 +52,11 @@ export { default as isEmpty } from './is_empty';
 export { default as isBlank } from './is_blank';
 export { default as isPresent } from './is_present';
 export { default as run } from './run_loop';
-export { default as ObserverSet } from './observer_set';
```
mmun: oops, not sure how this snuck in there

PR #16179, opened 2018-01-27
[BUGFIX beta] Fix a few bugs in the caching ArrayProxy implementation

mmun:
- Fix handling of negative indexes: `arrayProxy.replace(-1, 0, [obj]);`
- Change `arrangedContent` observer to use private `PROPERTY_DID_CHANGE`. See the test "should update if content changes while change events are deferred" for an example of why this is needed. This was also causing the ember-data tests to fail.
- Optimize `arrayProxy.get('length')` so that it no longer triggers the objects cache to revalidate.
- Changed the `length` computed property to be volatile since we are handling the cache ourselves.
- Added a comment explaining what the `this._objectsDirtyIndex` field is doing.
  - Switch to using -1 as a sentinel instead of `undefined`.
  
I've confirmed locally that this fixes the failing test in ember-data.

### DIFF
--- a/packages/ember-runtime/lib/system/array_proxy.js
+++ b/packages/ember-runtime/lib/system/array_proxy.js
@@ -5,8 +5,8 @@
 import {
   get,
   computed,
-  observer,
-  alias
+  alias,
+  PROPERTY_DID_CHANGE
 } from 'ember-metal';
 import {
   isArray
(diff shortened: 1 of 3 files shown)


rwjblue (approved):

rwjblue:
This will need to be pulled into beta (as the bug it fixes is also in beta)...

PR #16185, opened 2018-01-28
Fix didReceiveAttrs in ES Class

wycats:


packages/ember-glimmer/lib/component-managers/curly.ts
```diff
@@ -28,6 +28,7 @@ import {
   assert,
   deprecate,
 } from 'ember-debug';
+import { ENV } from 'ember-environment';
```
rwjblue: This is failing CI due to a linting error:

> ERROR: /home/travis/build/emberjs/ember.js/packages/ember-glimmer/lib/component-managers/curly.ts[32, 1]: Import sources within a group must be alphabetized.

I believe this line should be moved down below `import { DEBUG } from 'ember-env-flags';`.

rwjblue (dismissed):

krisselden:
What’s the status?

packages/ember-glimmer/lib/component-managers/curly.ts
mmun: Fixed.

PR #16187, opened 2018-01-28
Simplify ObserverSet implementation

mmun:
Now that before observers are removed, there's no need to carefully track listeners between `beginPropertyChanges` and `endPropertyChanges`. Now, we simply call all the observers that exist after `endPropertyChanges`.

@krisselden approved the idea in chat.

### DIFF
--- a/.eslintrc-browser.js
+++ b/.eslintrc-browser.js
@@ -29,6 +29,7 @@ module.exports = {
     'clearInterval': true,
     'console': true,
 
+    'Map': true,
     'Set': true,
     'Symbol': true,
     'WeakMap': true,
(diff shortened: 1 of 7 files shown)


rwjblue:
seems good to me...

bekzod:
this is also closes :P https://github.com/emberjs/ember.js/pull/15928/files don't mind it, I think this one more update now :)

mmun:
@bekzod The flush &rarr; forEach refactor looks good (so that observer set isn't coupled to sendEvent) if you want to rebase that part.

bekzod:
@mmun maybe I can just do follow up PR and close previous one  ?

mmun:
sure

PR #16188, opened 2018-01-28
decouple `sendEvent` from `ObserverSet`

bekzod:


bekzod:
cc @mmun

packages/ember-metal/lib/observer_set.js
```diff
@@ -28,28 +25,20 @@ export default class ObserverSet {
...
   }
 
-  flush() {
-    for (let i = 0; i < this.queue.length; i++) {
-      let { object, key } = this.queue[i];
-
-      if (object.isDestroying || object.isDestroyed) {
-        continue;
+  forEach(cb) {
+    this.added.forEach(function(keys, object) {
```
mmun: You need to loop over `this.queue` not `this.added` otherwise the observers may be triggered in the wrong order

mmun (changes requested):

packages/ember-metal/lib/observer_set.js
bekzod: ok got it

packages/ember-metal/lib/observer_set.js
```diff
@@ -28,28 +25,20 @@ export default class ObserverSet {
...
+      if (!(object.isDestroying || object.isDestroyed)) {
+        keys.forEach(function(key) {
+          cb(object, key, [object, key]);
```
mmun: Also I think it is better to just call with `cb(object, key)`. The last argument is specific to `sendEvent`.

packages/ember-metal/lib/observer_set.js
mmun: Arrow functions too ;)

packages/ember-metal/lib/observer_set.js
```diff
@@ -29,23 +27,22 @@ export default class ObserverSet {
     }
 
     if (!keys.has(key)) {
-      this.queue.push({ object, key });
+      this.queue.push(object, key);
```
mmun: I wrote it the way it was intentionally...

packages/ember-metal/lib/observer_set.js
```diff
@@ -29,23 +27,22 @@ export default class ObserverSet {
...
+    for (let i = 0; i < this.queue.length; i+=2) {
+      let object = this.queue[i];
 
       if (object.isDestroying || object.isDestroyed) {
         continue;
       }
 
-      sendEvent(object, key, [object, key]);
+      let key = this.queue[i + 1];
+      cb(object, key, [object, key]);
```
mmun: just `cb(object, key)`

packages/ember-metal/lib/observer_set.js
bekzod: ok 😉

mmun (approved):

mmun:
There's just some linting errors.

Issue #16198, opened 2018-01-30
yarn test fails with '/' in branch name.


Parrryy:
When running `yarn test` with a '/' in the branch name the test fails with the following message.

```
Test failed: ember-metal/core/main: Ember.VERSION is in alignment with SemVer v2.0.0
    Failed assertion: Ember.VERSION (3.1.0-fix/ember-glimmer-views-doc-update+ca894ce2)is valid SemVer v2.0.0, expected: true, but was: false
    at Object.<anonymous> (http://localhost:13141/ember-tests.js:42211:12)
    at runTest (http://localhost:13141/qunit/qunit.js:1478:30)
    at Test.run (http://localhost:13141/qunit/qunit.js:1464:6)
    at http://localhost:13141/qunit/qunit.js:1670:12
    at Object.advance (http://localhost:13141/qunit/qunit.js:1116:26)
    at begin (http://localhost:13141/qunit/qunit.js:3059:20)
```

It should convert illegal characters in the branch name to something legal.

Serabe:
Closed as the PR is merged.

PR #16200, opened 2018-01-31
[BUGFIX beta] prevent test error by converting illegal characters

Parrryy:
Addresses #16198. Converts all non-alphanumeric characters to '-' to prevent error when running test's.

### DIFF
--- a/broccoli/version.js
+++ b/broccoli/version.js
@@ -25,9 +25,11 @@ module.exports.VERSION = (() => {
   // * current "build type" or branch name (in CI this is generally not
   //   present, but it is very useful for local / testing builds)
   // * the sha for the commit
-  let packageVersion  = pkg.version;
+  let packageVersion = pkg.version;
   let sha = info.sha || '';
   let suffix = process.env.BUILD_TYPE || info.branch;
+  // * remove illegal non-alphanumeric characters from branch name.
+  suffix = suffix && suffix.replace(/[^a-zA-Z\d\s-]/, '-');
   let metadata = sha.slice(0, 8);
 
   return `${packageVersion}${suffix ? '-' + suffix : ''}+${metadata}`;


broccoli/version.js
```diff
@@ -25,8 +25,12 @@ module.exports.VERSION = (() => {
   // * current "build type" or branch name (in CI this is generally not
   //   present, but it is very useful for local / testing builds)
   // * the sha for the commit
-  let packageVersion  = pkg.version;
+  let packageVersion = pkg.version;
   let sha = info.sha || '';
+  // * remove illegal non-alphanumeric characters from branch name.
+  if(info.branch !== null && new RegExp(/[^a-zA-Z\d\s-]/).test(info.branch)) {
+    info.branch = info.branch.replace(/[^a-zA-Z\d\s-]/, '-');
```
rwjblue: Would you mind tweaking this a tad bit, I'd like to avoid mutating the `info` object here and avoid repeating the regexp.

I _think_ this might do the trick:

```js
// ...snip...
let sha = info.sha || '';
let suffix = process.env.BUILD_TYPE || info.branch;
suffix = suffix.replace(/[^a-zA-Z\d\s-]/, '-');
// ...snip...
```

@Parrryy - What do you think?

rwjblue (changes requested):

broccoli/version.js
Parrryy: Sounds good to me.

Parrryy: @rwjblue What about the case when the suffix is `process.env.BUILD_TYPE`? I'm not exactly sure what that is and if the `.replace` will effect it.

broccoli/version.js
```diff
@@ -25,9 +25,11 @@ module.exports.VERSION = (() => {
   // * current "build type" or branch name (in CI this is generally not
   //   present, but it is very useful for local / testing builds)
   // * the sha for the commit
-  let packageVersion  = pkg.version;
+  let packageVersion = pkg.version;
   let sha = info.sha || '';
   let suffix = process.env.BUILD_TYPE || info.branch;
+  // * remove illegal non-alphanumeric characters from branch name.
+  suffix = suffix.replace(/[^a-zA-Z\d\s-]/, '-');
```
Parrryy: @rwjblue The issue now is one I was having before where the `suffix` var is null. It passes tests locally, just on Travis it fails. Any suggestions?

rwjblue: Ah, I see. 

Something like:

`suffix = suffix && suffix.replace(...)`

Should work I believe...

Parrryy: @rwjblue Yup, all good now.

rwjblue (approved):

rwjblue:
Awesome, thank you!

PR #16202, opened 2018-01-31
Fix `let` RFC link in FEATURES.md

jacobq:
The `0000-block-let-template-helper.md` link should be updated to use the new name/number.


locks (approved):
Cheers!

PR #16206, opened 2018-02-01
Remove MutableArrayMixin, ArrayMixin, NativeArrayMixin cycle

chadhietala:
This removes a giant and annoying cycle between `MutableArrayMixin`, `ArrayMixin`, and `NativeArrayMixin`. To solve it we have simply collapsed all of these into a single file 😏. This also has allowed us to remove what we feel as the last `require` call. Once we have verified all of the `require` calls are gone we can think about creating a happy path of using Ember as loose modules.


mmun:
Looks good. I'm working on something that should make mixins/array.js a bit smaller.

packages/ember/lib/index.js
```diff
@@ -306,6 +307,7 @@ import {
   collect
 } from 'ember-runtime';
 
+Ember.A = A;
```
mmun: Why was this added?

mmun (changes requested):

packages/ember/lib/index.js
mmun: Nevermind, https://github.com/emberjs/ember.js/pull/16206/files#diff-7be349ee898133ade03bc4adca5582bcL154.

mmun (approved):

PR #16210, opened 2018-02-01
Show the right ContainerProxyMixin Docs

toddjordan:
Part of the fix for https://github.com/ember-learn/ember-api-docs/issues/423

ContainerProxy mixin had a bug in its documentation.  It had a `@class` annotation defined on one of its methods, making the Class public and documenting the method as a description to the class, when it is actually labelled private.

Before:
image

After: 
image


rwjblue:
Thank you!

PR #16215, opened 2018-02-05
Move each proxy into a weak map

mmun:
- Move each proxy into meta so it doesn't have to be a computed property.
- Deprecate `get`ing `@each`. I am treating this as an intimate private API.
- This required moving `objectAt` into ember-metal to avoid ember-metal depending on ember-runtime.
- Fix a test that should have been testing `array.objectAt(i)` instead of `objectAt(array, i)`.

Remaining items:

- [x] [Deprecation guide](https://github.com/emberjs/website/pull/3179)

### DIFF
--- a/packages/ember-runtime/lib/mixins/array.js
+++ b/packages/ember-runtime/lib/mixins/array.js
@@ -17,14 +17,14 @@ import {
   removeListener,
   sendEvent,
   hasListeners,
-  addObserver,
-  removeObserver,
-  meta,
   peekMeta,
+  eachProxyFor,
+  eachProxyArrayWillChange,
+  eachProxyArrayDidChange,
   beginPropertyChanges,
   endPropertyChanges
 } from 'ember-metal';
-import { assert } from 'ember-debug';
+import { assert, deprecate } from 'ember-debug';
 import Enumerable from './enumerable';
 import compare from '../compare';
 import { ENV } from 'ember-environment';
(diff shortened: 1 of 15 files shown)


packages/ember-runtime/lib/mixins/array.js
```diff
@@ -1237,126 +1233,20 @@ const ArrayMixin = Mixin.create(Enumerable, {
     @public
   */
   '@each': computed(function() {
-    // TODO use Symbol or add to meta
-    if (!this.__each) {
-      this.__each = new EachProxy(this);
-    }
+    deprecate(
+      "Getting the '@each' property is deprecated",
```
rwjblue: The message needs to include more details, or it’s going to be very hard to track down...

packages/ember-runtime/lib/mixins/array.js
```diff
@@ -1237,126 +1233,20 @@ const ArrayMixin = Mixin.create(Enumerable, {
...
+      {
+        id: 'ember-metal.getting-each',
+        until: '3.5.0',
+        url: 'https://emberjs.com/deprecations/v3.x#toc_getting-the-each-property'
+      }
+    );
 
-    return this.__each;
+    return meta(this).eachProxy;
   }).volatile().readOnly()
```
rwjblue: I can’t recall, why is this volatile (seems like it is stable based on the implementation)?

rwjblue (commented):
Looks good!

packages/ember-runtime/lib/mixins/array.js
mmun: I don't know exactly why it was volatile, but it certainly doesn't need to be now (in fact, we could remove this CP if we wanted).

mmun:
In chat, Rob suggested I move the each proxy into a weakmap rather than meta. I will do this.

mmun:
@rwjblue Updated.

PR #16217, opened 2018-02-05
Unify run loop setup.

rwjblue:
Declare run-loop queues in single centralized location.

Prior to this change the list of queue names was mutated _after_ the backburner instance was instantiated. This sort of private API manipulation is problematic because Backburner may move to an alternative internal structure.

This also makes it much much easier to grok the actual queues in the system and their ordering...


rwjblue:
@krisselden - This also makes `rsvpAfter` a "private" queue...

PR #16218, opened 2018-02-06
[BUGFIX beta] Prevent errors when using const `(get arr 1)`.

rwjblue:
Prior to this change, calling `{{get arr 1}}` would throw an error:

```
 pathReference.value(...).split is not a function
```

This commit fixes that, by only attempting to `.split` when the underlying value is not actually a string.

### DIFF
--- a/packages/ember-glimmer/lib/helpers/get.ts
+++ b/packages/ember-glimmer/lib/helpers/get.ts
@@ -79,8 +79,12 @@ class GetHelperReference extends CachedReference {
 
   static create(sourceReference: VersionedPathReference<Opaque>, pathReference: PathReference<string>) {
     if (isConst(pathReference)) {
-      let parts = pathReference.value().split('.');
-      return referenceFromParts(sourceReference, parts);
+      let value = pathReference.value();
+      if (typeof value === 'string' && value.indexOf('.') > -1) {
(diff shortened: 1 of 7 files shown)


rwjblue:
FYI - The `Ember.get` codepath is essentially untouched with one small exception: I added an early `return false` guard to the `isPath` utility function (to avoid splitting when the argument is not a string).

packages/ember-glimmer/lib/helpers/get.ts
```diff
@@ -79,8 +79,12 @@ class GetHelperReference extends CachedReference {
 
   static create(sourceReference: VersionedPathReference<Opaque>, pathReference: PathReference<string>) {
     if (isConst(pathReference)) {
-      let parts = pathReference.value().split('.');
-      return referenceFromParts(sourceReference, parts);
+      let value = pathReference.value();
+      if (typeof value === 'string' && value.indexOf('.') > -1) {
+        return referenceFromParts(sourceReference, value.split('.'));
```
rwjblue: FWIW, I am working on a follow up PR (which will only target master) to make the `referenceFromParts` utility function accept a single value (instead of an array), that will remove this duplication (here and below)...

rwjblue:
r+ from @krisselden in chat

PR #16220, opened 2018-02-06
[BUGFIX beta] Allow symbol usage throughout internals.

rwjblue:
Previously, this would result in an exception (both `Ember.get` and `Ember.set` expected a string for the key name argument).


mmun:
I'm wondering about notifyPropertyChange for this pr and the previous one. Should we guard it?

rwjblue:
@mmun - Add a few more tests in the general infrastructure (`addObserver`, `computed`, etc) that use both numbers and symbols, mind taking another look?

mmun (approved):

rwjblue:
Rebased...

locks:
Mega conflict. Should I attempt to rebase or would a fresh start be better?

locks:
Closing as I'm pretty sure this is no longer relevant!

PR #16221, opened 2018-02-07
cleanup get helper

bekzod:


rwjblue (approved):

packages/ember-glimmer/lib/helpers/get.ts
```diff
@@ -105,26 +113,13 @@ class GetHelperReference extends CachedReference {
...
-        innerTag.inner.update(innerReference.tag);
-      } else {
-        innerReference = NULL_REFERENCE;
-        innerTag.inner.update(CONSTANT_TAG);
-      }
-
+      innerReference = referenceFromPath(this.sourceReference, path);
+      innerTag.inner.update(innerReference.tag);
       this.innerReference = innerReference;
+      this.lastPath = path;
```
rwjblue: FWIW, I think this fixed another related bug (we previously _always_ created a new reference even if the path was not changed)...

PR #16222, opened 2018-02-08
Move computed property caches into a weak map

mmun:
Continuing our effort to make meta smaller...

FYI, the `cacheFor.{get,set,remove}` have never been used in a public add-on so I'm remove them. We weren't even using them internally :)

### DIFF
--- a/packages/ember-metal/lib/alias.js
+++ b/packages/ember-metal/lib/alias.js
@@ -6,7 +6,7 @@ import {
   Descriptor,
   defineProperty
 } from './properties';
-import { ComputedProperty } from './computed';
+import { ComputedProperty, getCacheFor } from './computed';
 import { meta as metaFor } from './meta';
 import {
   addDependentKeys,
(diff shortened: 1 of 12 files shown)


rwjblue (commented):
Looks great to me!  We should do a macro benchmark to sanity check things here...

rwjblue:
For those following along, we are trying to remove usage of `findOrCreateOwnMap` in `Meta`, and focus the meta class on only things that are inherited...

mmun:
Here are the results I got from benchmarking https://emberobserver.com/categories/components:

[results.pdf](https://github.com/emberjs/ember.js/files/1705743/results.pdf)

Here are meta counters for loading the index page:

|                    | master | experiment |
|--------------------|--------|------------|
| peekCalls          | 70096  | 62853      |
| peekParentCalls    | 0      | 0          |
| peekPrototypeWalks | 2347   | 2352       |
| setCalls           | 1072   | 1071       |
| deleteCalls        | 1      | 1          |
| metaCalls          | 22574  | 15353      |
| metaInstantiated   | 1072   | 1071       |

rwjblue:
Awesome, thank you!

PR #16223, opened 2018-02-08
Expose meta counters (privately).

rwjblue:
Various counters are being tracked around meta usage (how many instantiated, how many prototype walks, etc) but this data is not exposed (even privately) at the moment.

This PR updates `Ember.meta` to tack on `Ember.meta._counters` which exposes the counters object in debug builds only...


mmun:
Yes, please! I was assigning them to window.metaCounters in a conditional breakpoint 😆

PR #16227, opened 2018-02-08
Allow visit to specify Rehydration/Serialization or default ClientBuilder

rondale-sc:
Expose as a configurable option the `renderMode` option.  This option
would be to specify the clientBuilder to be used when Glimmer does it's
append run.

The serialization mode, used by SSR applications, is used to specify
additional markings necessary to get enough fidelity to accurately
rehydrate the DOM.  For example, it would provide additional comment
nodes with codes to ensure text nodes are separated when they are rather
than merged.

The rehydration mode is specifically designed to read DOM that is
produced by the serialization mode and accurately reproduce it.

A great description of how Rehydration works can be found in this
commit:

https://github.com/glimmerjs/glimmer-vm/commit/316805b9175e01698120b9566ec51c88d075026a

This PR allows the appropriate ElementBuilder interface to be used via
the `visit` API.

---

**Current plan:**

This is part of an arching plan to introduce Glimmer's
rehydration/serializtion modes to Ember proper.

There are 4 PR's that are all interwoven, of which, this is one.

- [x] Glimmer.js: glimmerjs/glimmer-vm#783 (comment)
- [x] Ember.js: emberjs/ember.js#16227
- [x] Fastboot: ember-fastboot/fastboot#185
- [ ] EmberCLI Fastboot: ember-fastboot/ember-cli-fastboot#580

In order of need to land they are:

Glimmer.js:
glimmerjs/glimmer-vm#783 (comment)

This resolves a rather intimate API problem. Glimmer-vm expects a very
specific comment node to exist to know whether or not the rehydration
element builder can do it's job properly. If it is not found on the
first node from rootElement it throws.

In fastboot however we are certain that there will already be existant
elements in the way that will happen before rendered content.

This PR just iterates through the nodes until it finds the expected
comment node. And only throws if it never finds one.

Ember.js:
emberjs/ember.js#16227

This PR modifies the visit API to allow a private _renderMode to be
set to either serialize or rehydrate. In SSR environments the
assumption (as it is here in this fastboot PR) is that we'll set the
visit API to serialize mode which ensures glimmer-vm's serialize element
builder is used to build the API.

The serialize element builder ensures that we have the necessary fidelty
to rehydrate correctly and is mandatory input for rehydration.

Fastboot:
ember-fastboot/fastboot#185

This allows enviroment variable to set _renderMode to be used in Visit
API. Fastboot must send content to browser made with the serialization
element builder to ensure rehydration can be sucessful.

EmberCLI Fastboot:
ember-fastboot/ember-cli-fastboot#580

Finally this does the fun part of disabling the current
clear-double-render instance-initializer

We first check to ensure we are in a non-fastboot environment. Then we
ensure that we can find the expected comment node from glimmer-vm's
serialize element builder. This ensures that this change will only
effect peoeple who use ember-cli-fastboot with the serialized output
from the currently experimental fastboot setup

Then we ensure ApplicationInstance#_bootSync specifies the rehydrate
_renderMode.

This is done in _bootSync this way because visit is currently not used
to boot ember applications. And we must instead set bootOptions this
way instead.

We also remove the markings for fastboot-body-start and
fastboot-body-end to ensure clear-double render instance-initializer
is never called.

### DIFF
--- a/packages/ember-application/lib/system/application-instance.js
+++ b/packages/ember-application/lib/system/application-instance.js
@@ -340,7 +340,7 @@ class BootOptions {
       @type boolean
       @default auto-detected
       @private
-      */
+    */
     this.isInteractive = environment.hasDOM; // This default is overridable below
 
 
--- a/packages/ember-application/lib/system/application-instance.js
+++ b/packages/ember-application/lib/system/application-instance.js
@@ -340,8 +340,17 @@ class BootOptions {
       @type boolean
       @default auto-detected
       @private
-    */
+      */
     this.isInteractive = environment.hasDOM; // This default is overridable below
+
+
+    /**
+      @property _renderMode
+      @type string
+      @default false
+      @private
+    */
+    this._renderMode = options._renderMode;
 
     /**
       Run in a full browser environment.
@@ -482,6 +491,7 @@ class BootOptions {
     // For compatibility with existing code
     env.hasDOM = this.isBrowser;
     env.isInteractive = this.isInteractive;
+    env._renderMode = this._renderMode;
     env.options = this;
     return env;
   }
--- a/packages/ember-application/tests/system/visit_test.js
+++ b/packages/ember-application/tests/system/visit_test.js
@@ -53,7 +53,7 @@ moduleFor('Application - visit()', class extends ApplicationTestCase {
     });
   }
 
-  [`@test renderMode: rehydrate`](assert) {
+  [`@test _renderMode: rehydrate`](assert) {
 
     let initialHTML = `<!--%+block:0%--><!--%+block:1%--><!--%+block:2%--><!--%+block:3%--><!--%+block:4%--><!--%+block:5%--><!--%+block:6%--><div class=\"foo\">Hi, Mom!</div><!--%-block:6%--><!--%-block:5%--><!--%-block:4%--><!--%-block:3%--><!--%-block:2%--><!--%-block:1%--><!--%-block:0%-->`;
 
@@ -64,7 +64,7 @@ moduleFor('Application - visit()', class extends ApplicationTestCase {
     let bootOptions = {
       isBrowser: false,
       rootElement,
-      renderMode: 'rehydrate'
+      _renderMode: 'rehydrate'
     };
 
     ENV._APPLICATION_TEMPLATE_WRAPPER = false;
@@ -75,14 +75,14 @@ moduleFor('Application - visit()', class extends ApplicationTestCase {
     });
   }
 
-  [`@test renderMode: serialize`](assert) {
+  [`@test _renderMode: serialize`](assert) {
     this.addTemplate('index', '<div class="foo">Hi, Mom!</div>');
     let rootElement = document.createElement('div');
 
     let bootOptions = {
       isBrowser: false,
       rootElement,
-      renderMode: 'serialize'
+      _renderMode: 'serialize'
     };
 
     ENV._APPLICATION_TEMPLATE_WRAPPER = false;
--- a/packages/ember-application/tests/system/visit_test.js
+++ b/packages/ember-application/tests/system/visit_test.js
@@ -53,30 +53,11 @@ moduleFor('Application - visit()', class extends ApplicationTestCase {
     });
   }
 
-  [`@test _renderMode: rehydrate`](assert) {
+  [`@test _renderMode: rehydration`](assert) {
+    assert.expect(2);
 
-    let initialHTML = `<!--%+b:0%--><!--%+b:1%--><!--%+b:2%--><!--%+b:3%--><!--%+b:4%--><!--%+b:5%--><!--%+b:6%--><div class=\"foo\">Hi, Mom!</div><!--%-b:6%--><!--%-b:5%--><!--%-b:4%--><!--%-b:3%--><!--%-b:2%--><!--%-b:1%--><!--%-b:0%-->`;
-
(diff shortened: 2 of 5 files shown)


josemarluedke:
Awesome to see Rehydration coming to Ember!

rondale-sc:
To assuage any fears of file size bloat this takes ember.min.js (gzipped) from 115.73kb to 115.89.  Which should definitely be address, but I shouldn't think be a blocker.

rwjblue:
> To assuage any fears of file size bloat this takes ember.min.js (gzipped) from 115.73kb to 115.89. Which should definitely be addressed, but I shouldn't think be a blocker.

Agreed.

rondale-sc:
Relevant to this ticket: https://github.com/glimmerjs/glimmer-vm/pull/780

rondale-sc:
After discussing this, it was decided that the configuration for renderMode should be a private var to avoid muddying the API surface.  Updated the PR accordingly

rwjblue:
looks like this needs a rebase

rondale-sc:
@rwjblue Rebased.  I'm still running tests locally, but should be gtg

rondale-sc:
Looks like I need to dig into this one.  Working on it now.

packages/ember-application/tests/system/visit_test.js
```diff
@@ -35,6 +37,67 @@ moduleFor('Application - visit()', class extends ApplicationTestCase {
...
+
+    let bootOptions = {
+      isBrowser: false,
+      rootElement,
+      _renderMode: 'serialize'
+    };
+
+    ENV._APPLICATION_TEMPLATE_WRAPPER = false;
+
+    return this.runTask(() => {
```
rwjblue: we should remove this runTask (otherwise this makes a nested run loop and we shouldn't need that)

packages/ember-application/tests/system/visit_test.js
```diff
@@ -35,6 +37,67 @@ moduleFor('Application - visit()', class extends ApplicationTestCase {
...
+      return this.visit('/', bootOptions)
+        .then((instance) => {
+          assert.equal(
+            instance.rootElement.firstChild.nodeValue,
+            '%+b:0%',
```
rwjblue: I don't really like the coupling to the internal / private mechanism that  Glimmer's serialize mode renderer uses internally (generally speaking I'd much prefer to test the _results_ rather than _how_ it works internally) but I realize this is somewhat of a chicken or the egg situation. 

Thoughts?

packages/ember-application/tests/system/visit_test.js
```diff
@@ -35,6 +37,67 @@ moduleFor('Application - visit()', class extends ApplicationTestCase {
...
+        this.applicationInstance.destroy();
+        this.applicationInstance = null;
+      });
+    }).then(() => {
+      bootOptions = {
+        isBrowser: false,
+        rootElement,
+        _renderMode: 'rehydrate'
+      };
+      this.application.visit('/', bootOptions).then(instance => {
```
rwjblue: We should be able to do `this.visit` again (since you are properly destroying/clearing it above).

rwjblue (commented):
Looking really nice! A few more things to cleanup... 

* After the initial serialize render we should grab the actual `div` element so that we can ensure it is "rehydrated"
* Add additional assertions that confirm the element is _exactly_ the initial element
* Add additional assertion that confirm the template isn't rendered into the root element twice...

packages/ember-application/tests/system/visit_test.js
rondale-sc: I totally agree that I'd prefer not to couple so tightly to the internal implementation.  Though to ensure we are actually doing some work this seems to be necessary.

I'd like in the future to expose the leading comment node in the serializer builder in glimmer as an export so we don't have to hardcode this value.  Perhaps that could be a next step?

packages/ember-application/tests/system/visit_test.js
```diff
@@ -35,6 +37,66 @@ moduleFor('Application - visit()', class extends ApplicationTestCase {
...
+          this.applicationInstance = null;
+        });
+      }).then(() => {
+        bootOptions = {
+          isBrowser: false,
+          rootElement,
+          _renderMode: 'rehydrate'
+        };
+
+        this.application.visit('/', bootOptions).then(instance => {
```
rondale-sc: @rwjblue turns out this is actually necessary.  When you remove the `this.application` and just call `visit` directly we get an error saying that you tried to use an object that is destroyed (sorry I don't have the exact wording of the error) . Either way nbd.

packages/ember-application/lib/system/application-instance.js
```diff
@@ -343,6 +343,15 @@ class BootOptions {
...
     this.isInteractive = environment.hasDOM; // This default is overridable below
 
+
+    /**
+      @property _renderMode
+      @type string
+      @default false
+      @private
+    */
+    this._renderMode = options._renderMode;
```
rondale-sc: This was recently changed and I think the commit history of this PR is now a little inaccurate to what actually happened.  I suspect we'll sqaush soon anyways, so I'm not worried about it, but if it is a concern I can flesh out the commit history.

packages/ember-application/tests/system/visit_test.js
rwjblue: Seems fine, but I still don't like it 😝

rwjblue (approved):

PR #16228, opened 2018-02-09
Consolidate ember-template-compiler global compat

chadhietala:
This moves the globals for ember-template-compiler into the main globals mode file.


rwjblue (commented):
Ya, this doesn’t really work. The ember-template-compiler.js uses the ember-template-compiler index as the main entry-point...

chadhietala:
Closing for now.

PR #16235, opened 2018-02-11
small optimization to `getPath` and `setPath`

bekzod:


packages/ember-metal/lib/property_set.js
```diff
@@ -109,18 +110,18 @@ if (MANDATORY_SETTER) {
...
   assert('Property set failed: You passed an empty path', keyName.trim().length > 0);
 
   let newPath = parts.join('.');
 
-  let newRoot = getPath(root, newPath);
+  let newRoot = _getPath(root, newPath);
 
   if (newRoot) {
-    return set(newRoot, keyName, value);
+    return _set(newRoot, keyName, value);
```
bekzod: using `_set` here to avoid `isPath` check

packages/ember-metal/lib/property_get.js
```diff
@@ -126,7 +127,7 @@ export function _getPath(root, path) {
       return undefined;
     }
 
-    obj = get(obj, parts[i]);
+    obj = _get(obj, parts[i]);
```
bekzod: using `_get` here to avoid `isPath` check

bekzod:
surprising benchmark results :) https://cl.ly/1Z3d0R3I3k0y/results.pdf

mmun:
This looks great! Can you rename _get/_set to _getKey/_setKey to align with watch/watchKey/watchPath. Similarly, for 💅, I think we can drop the underscores on _getKey/_setKey/_getPath/_setPath.

mmun:
@bekzod I just realized that it was _slower_ rather than _faster_. I'll have to take a closer look at what's going on before merging.

BTW you can just remove the _getPath export as well as the test for its presence.

rwjblue:
> BTW you can just remove the _getPath export as well as the test for its presence.

Good idea, @mmun...

bekzod:
with 80 samples same results https://cl.ly/2z2G0X0x0w1C/results.pdf will check other variations of code

mmun:
@rwjblue I haven't looked into it since. I'm also curious...

bekzod:
results look better if we drop `isPath` cache https://cl.ly/1q0G0r182I2h/results.pdf

bekzod:
needs another benchmark, prior benchmark was done on bad code (which failed tests)

krisselden (approved):

bekzod:
@krisseldon will be first thing that i do in the morning, last time I checked performed same as old code nothing huge

rwjblue:
@bekzod - When you take a look, you will need to rebase also...

bekzod:
after rebase got test failing for `Application, default resolver with autoboot / @test acts like a namespace`

locks:
Might be related to this? https://github.com/emberjs/ember.js/pull/16446/files

bekzod:
@locks it was failing before that change and after same thing, so not related I think

bekzod:
all green, issue was assertion in wrong place which was striped in prod so earlier bench is still valid

krisselden (changes requested):
Overall I would like to drop the isPath cache if it isn't buying us much but one thing I'm remembering is originally we had it like this, but it was faster to check if the key was a miss before checking isPath.  I'm for trying this again, but wondering if it would be better to add an isPath arg to get to override whether the indexOf check is done again.  Also wonder if caching splits would be better than caching isPath. Also if the isPath cache limit was too low.

It would be good to gather get() calls from some apps and make a microbenchmark for this.  The microbenchmark could still add some pollution at the start.

bekzod:
made separate PR #17422 that makes this redundant

PR #16238, opened 2018-02-11
Remove Array Suites

chadhietala:
This removes the suite infrastructure around testing all the array types in Ember. This is being removed because it relies on module state and has been a bit of a refactoring hazard according to @krisselden. This should be a 1 to 1 migration to the test infrastructure that was built for testing the new Glimmer VM. In addition to that I noticed several tests were not even being run and are now being ran.


packages/ember-runtime/tests/mutable-array/unshiftObjects-test.js
```diff
@@ -70,7 +70,7 @@ class UnshiftObjectsTests extends AbstractTestCase {
     this.assert.equal(observer.timesCalled('@each'), 0, 'should not have notified @each once');
     this.assert.equal(observer.timesCalled('length'), 1, 'should have notified length once');
 
-    this.assert.equal(observer.validate('firstObject'), false, 'should NOT have notified firstObject');
+    this.assert.equal(observer.validate('firstObject'), true, 'should NOT have notified firstObject');
```
chadhietala: @krisselden or @stefanpenner, these tests have not been running for at least 4 years from what I can tell. This is a failing test when it is set to `false`. I don't know if this test assertion is still correct.

mmun: This should notify (the message should be updated).

This is the logic we have used to notify firstObject for a long time:
https://github.com/emberjs/ember.js/blob/master/packages/ember-runtime/lib/mixins/array.js#L119

Basically, if the firstObject has been cached (it has in this test) and the 0<sup>th</sup> index has touched (it has because of the unshift) we notify firstObject. Regardless of whether the new value is the equal to the previous value.

chadhietala: +1. Yea I stepped through this code and was pretty sure the test was wrong, but wanted to make sure since they haven't been running in almost half a decade.

rwjblue:
great work here @chadhietala, thank you!

PR #15967, opened 2017-12-12
[FEATURE ember-module-unification] Support name-spaced lookups

iezer:
To support module unification, Ember needs to be able to pass lookup strings that have an an addon namespace, for example, `"component::my-addon:my-input"`

Related PRs:
 * https://github.com/ember-cli/ember-resolver/pull/215
 * https://github.com/glimmerjs/glimmer-resolver/pull/20

Prior work: https://github.com/emberjs/ember.js/pull/15368

@mixonic

PR #16240, opened 2018-02-13
[Module Unification] Introduce source to injections

mixonic:
Ember templates use a `source` argument to the container. This hints where to look for "local lookup", and also what namespace to look in for a partial specifier. For example a template in an addon's `src` tree can invoke components and helpers in that tree because it has a `source` argument passed to lookups.

This introduces the same functionality for service injections. A service can now accept a `source` argument and its lookup will apply to the namespace of that file.

Additionally:

* Use `expandLocalLookup` to for the resolution of local item instead of a new argument to `resolve` on resolvers.
* Get rid of `resolverCacheKey`, no longer needed.

### DIFF
--- a/packages/ember-metal/lib/injected_property.js
+++ b/packages/ember-metal/lib/injected_property.js
@@ -20,11 +20,12 @@ import { descriptorFor } from './meta';
   @private
 */
 export default class InjectedProperty extends ComputedProperty {
-  constructor(type, name) {
+  constructor(type, name, options) {
     super(injectedPropertyGet);
 
     this.type = type;
     this.name = name;
+    this.source = options ? options.source : undefined;
   }
 }
 
@@ -35,5 +36,6 @@ function injectedPropertyGet(keyName) {
   assert(`InjectedProperties should be defined with the inject computed property macros.`, desc && desc.type);
   assert(`Attempting to lookup an injected property on an object without a container, ensure that the object was instantiated via a container.`, owner);
 
-  return owner.lookup(`${desc.type}:${desc.name || keyName}`);
+  let specifier = `${desc.type}:${desc.name || keyName}`;
+  return owner.lookup(specifier, {source: desc.source});
 }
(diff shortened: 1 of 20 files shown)


mixonic:
I think this is ready to go in. All the work is behind a feature flag. The implementation is in line with the Module Unification Namespaces RFC https://github.com/emberjs/rfcs/pull/309

I'm going to look at rebasing https://github.com/emberjs/ember.js/pull/15967 on top of this.

packages/ember-metal/lib/injected_property.js
```diff
@@ -35,5 +36,10 @@ function injectedPropertyGet(keyName) {
   assert(`InjectedProperties should be defined with the inject computed property macros.`, desc && desc.type);
   assert(`Attempting to lookup an injected property on an object without a container, ensure that the object was instantiated via a container.`, owner);
 
-  return owner.lookup(`${desc.type}:${desc.name || keyName}`);
+  let specifier = `${desc.type}:${desc.name || keyName}`;
+  if (desc.source) {
```
rwjblue: I don’t fully grok why we need this conditional. Seems to be that calling lookup like `owner.lookup(specifier, { source: undefined })` (which is what would happen if `desc.source` is not present) would be fine. What am I missing?

packages/ember/tests/service_injection_test.js
```diff
@@ -43,3 +43,131 @@ if (EMBER_METAL_ES5_GETTERS) {
...
+     * container around singletons and and local lookup. The local lookup
+     * is cached and the global injection is then looked up incorrectly.
+     *
+     * The paractical rules of Ember's module unification config are such
+     * that services cannot be locally looked up, thus this case is really
+     * just a demonstration of what could go wrong if we permit arbitrary
+     * configuration (such as a singleton type that has local lookup).
+     */
+    ['@test Services can be injected with same name, one with source one without, and share an instance'](assert) {
+      let routeASource = 'service:src/ui/routes/route-a/controller';
```
rwjblue: Seems odd to have the source here be prefixed with `service:` (the actual type is `controller`)

rwjblue (approved):
Only had a few minor tweaks / questions, but looks good to me...

PR #16243, opened 2018-02-14
[DOC release] Fix issue with invalid YUIDoc parsing.

rwjblue:


Turbo87:
why does YUIDoc consider those? shouldn't they be ignored due to `/*` vs `/**`?

mixonic (approved):

rwjblue:
> why does YUIDoc consider those? shouldn't they be ignored due to /* vs /**?

Yep, definitely seems like a bug still. This is just a workaround (though it seems fine that these be non-block comments anyways).

PR #16244, opened 2018-02-14
[DOC release] fix documented module for expandProperties

kturney:


mmun:
Looks good. This should go into beta.

PR #16245, opened 2018-02-14
[BUGFIX lts] Ensure errors in deferred component hooks can be recovered.

rwjblue:
Previously, any errors thrown during `didInsertElement` would leave the running environment in an invalid state (`env.inTransaction` would be `true` but `this._transaction` would have been nullified).

This commit ensures that we _always_ reset `inTransaction` if `Environment.prototype.commit` is called. Thus avoiding an error RE: "calling commit on null"...


kyleshay (approved):
🎉

krisselden:
we should reserve this as last resort for internal errors

krisselden:
Safe invoke user hooks

PR #16250, opened 2018-02-16
[RFC 297] Deprecation of Ember.Logger

lupestro:
I've implemented the changes to the project. Since Ember itself no longer uses Ember.Logger, this PR could perhaps use one more test that deliberately calls the methods to verify that the console is reporting deprecation. Otherwise, this should be set to go.

### DIFF
--- a/packages/ember-console/lib/index.js
+++ b/packages/ember-console/lib/index.js
@@ -1,3 +1,10 @@
+import { deprecate } from 'ember-debug';
+
+// Deliver message that the function is deprecated
+
+const DEPRECATION_MESSAGE = 'Use of Ember.Logger is deprecated. Please use `console` for logging.';
+const DEPRECATION_ID = 'ember-console.deprecate-logger';
+const DEPRECATION_URL = 'https://emberjs.com/deprecations/v3.x#toc_use-console-rather-than-ember-logger';
 /**
    @module ember
 */
@@ -6,6 +13,8 @@
   Override this to provide more robust logging functionality.
 
   @class Logger
+  @deprecated Use 'console' instead
+
   @namespace Ember
   @public
 */
@@ -25,7 +34,10 @@ export default {
    @param {*} arguments
    @public
   */
-  log() { return console.log(...arguments);}, // eslint-disable-line no-console
+  log() { 
+    deprecate( DEPRECATION_MESSAGE, false, { id: DEPRECATION_ID, until: '4.0.0', url: DEPRECATION_URL } );
+    return console.log(...arguments); // eslint-disable-line no-console
+  },
 
   /**
    Prints the arguments to the console with a warning icon.
@@ -41,7 +53,10 @@ export default {
    @param {*} arguments
    @public
   */
-  warn() { return console.warn(...arguments);}, // eslint-disable-line no-console
+  warn() { 
+    deprecate( DEPRECATION_MESSAGE, false, { id: DEPRECATION_ID, until: '4.0.0', url: DEPRECATION_URL } );
+    return console.warn(...arguments); // eslint-disable-line no-console
+  },
 
   /**
    Prints the arguments to the console with an error icon, red text and a stack trace.
@@ -57,7 +72,10 @@ export default {
    @param {*} arguments
    @public
   */
-  error() { return console.error(...arguments);}, // eslint-disable-line no-console
+  error() { 
+    deprecate( DEPRECATION_MESSAGE, false, { id: DEPRECATION_ID, until: '4.0.0', url: DEPRECATION_URL } );
+    return console.error(...arguments); // eslint-disable-line no-console
+  },
 
   /**
    Logs the arguments to the console.
@@ -74,7 +92,10 @@ export default {
    @param {*} arguments
    @public
   */
-  info() { return console.info(...arguments);}, // eslint-disable-line no-console
+  info() { 
+    deprecate( DEPRECATION_MESSAGE, false, { id: DEPRECATION_ID, until: '4.0.0', url: DEPRECATION_URL } );
+    return console.info(...arguments); // eslint-disable-line no-console
+  },
 
   /**
    Logs the arguments to the console in blue text.
@@ -92,11 +113,11 @@ export default {
    @public
   */
   debug() {
+    deprecate( DEPRECATION_MESSAGE, false, { id: DEPRECATION_ID, until: '4.0.0', url: DEPRECATION_URL } );
     /* eslint-disable no-console */
     if (console.debug) {
       return console.debug(...arguments);
     }
-
     return console.info(...arguments);
     /* eslint-enable no-console */
   },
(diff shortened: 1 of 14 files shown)


lupestro:
Four tests failed because the change I made to the debug() method in ember-debug/lib/index.js didn't fare well in Edge and IE11.  I'll hop on Windows and nail it down in the morning.

lupestro:
This has been interesting. There are 20 tests that fail only in Edge and IE11- and only when you **don't** have the developer tools up. The point of failure lies in precisely two lines of code in the source. 

The first occurs with `(console.debug ? console.debug : console.log)(params)`, which was easy to work around by replacing the operator with an actual if/else, but the workaround only fixes one test case. The failure gave a pretty strong hint of at least part of what's going on - when Edge isn't in devtools, I think we're dealing with a bound member function of an object - but I need to investigate further. 

The second occurs doing `console.error.apply(this, errorArgs)` in one spot in the router, and is responsible for at least 18 and probably all 19 of the remaining cases. I can shim around that by wrapping a call to console.error in an un-exported tiny logger shim class for the router only to use. With a little deeper understanding, I can probably do better still, and have a great topic for a blog post from it, so I'm going to pursue that. However, I'll do the shim first, since I know that will work.

lupestro:
Someone please review packages\ember-routing\lib\system\router.js line 1193 - I debated between ease of understanding and absolute performance here - should it be yanked to the top of the module to be created once? How critical is that? I'll still chase something more elegant once I figure out how Edge is defining the console differently with devtools closed, but this approach turns clearly failing tests to consistently passing tests, at least on my Windows box.

packages/ember-console/lib/index.js
```diff
@@ -1,3 +1,14 @@
+import { deprecate } from 'ember-debug';
+
+// Deliver message that the function is deprecated
+let deprecateEmberLogger = function(){
+  deprecate(
+    'Use of Ember.Logger is deprecated. Please use `console` for logging.',
+    false,
+    { id: 'ember-console.deprecate-logger', until: '3.5.0', url: 'https://emberjs.com/deprecations/v3.x#toc_use-console-rather-than-ember-logger' }
```
locks: I think the `until` should be `4.0.0`.

locks (dismissed):

packages/ember-console/lib/index.js
```diff
@@ -1,3 +1,14 @@
+import { deprecate } from 'ember-debug';
+
+// Deliver message that the function is deprecated
+let deprecateEmberLogger = function(){
```
rwjblue: I definitely appreciate the desire to be a bit DRY here, but the current implementation has a larger footprint in production builds (where the deprecation is stripped). Would you mind inlining the deprecation in each of the methods instead (I am sorry that we can’t have nice things...)?

packages/ember-debug/lib/index.js
```diff
@@ -130,7 +136,9 @@ if (DEBUG) {
     @private
   */
   setDebugFunction('info', function info() {
-    Logger.info.apply(undefined, arguments);
+    /* eslint-disable no-console */
+    console.info.apply(console, arguments);
```
rwjblue: Can we do `console.info(...arguments)` here instead?

packages/ember-debug/lib/index.js
```diff
@@ -130,7 +136,9 @@ if (DEBUG) {
     @private
   */
   setDebugFunction('info', function info() {
-    Logger.info.apply(undefined, arguments);
+    /* eslint-disable no-console */
```
rwjblue: @For some of these disablings you may be able to make it a tad less noisy via `// eslint-disable-next-line no-console`

packages/ember-debug/lib/warn.js
```diff
@@ -51,10 +50,13 @@ if (DEBUG) {
   };
 
   registerHandler(function logWarning(message) {
-    Logger.warn(`WARNING: ${message}`);
-    if ('trace' in Logger) {
-      Logger.trace();
-    }
+    /* eslint-disable no-console */
+    console.warn(`WARNING: ${message}`); //eslint-disable-line no-console
+    // VERIFY! if ('trace' in Logger) {
```
rwjblue: Can probably kill this comment

packages/ember-debug/lib/warn.js
```diff
@@ -51,10 +50,13 @@ if (DEBUG) {
   };
 
   registerHandler(function logWarning(message) {
-    Logger.warn(`WARNING: ${message}`);
-    if ('trace' in Logger) {
-      Logger.trace();
-    }
+    /* eslint-disable no-console */
+    console.warn(`WARNING: ${message}`); //eslint-disable-line no-console
```
rwjblue: Seems like this comment isn’t needed (since the surrounding ones should do the trick)

rwjblue (changes requested):

packages/ember-console/lib/index.js
lupestro: See if you like the approach I just checked in - I've split the difference to ensure that repeatedly typing long strings can't result in unnecessary and hard-to-spot differences, while still calling deprecate clearly in each method.

packages/ember-routing/lib/system/router.js
```diff
@@ -106,7 +105,13 @@ const EmberRouter = EmberObject.extend(Evented, {
 
     if (DEBUG) {
       if (get(this, 'namespace.LOG_TRANSITIONS_INTERNAL')) {
-        routerMicrolib.log = Logger.debug;
+        /* eslint-disable no-console */
+        if (console.debug) {
```
rwjblue: lets remove this, and just use `console.log` version only

packages/ember-routing/lib/system/router.js
```diff
@@ -1181,7 +1189,7 @@ function logError(_error, initialMessage) {
     if (typeof error === 'string') { errorArgs.push(error); }
   }
 
-  Logger.error.apply(this, errorArgs);
+  console.error.apply(console, errorArgs); //eslint-disable-line no-console
```
rwjblue: can you update this to `console.error(...errorArgs)` instead?

packages/ember/tests/routing/basic_test.js
```diff
@@ -86,8 +85,8 @@ QUnit.module('Basic Routing', {
       setTemplate('home', compile('<h3 class="hours">Hours</h3>'));
       setTemplate('homepage', compile('<h3 class="megatroll">Megatroll</h3><p>{{model.home}}</p>'));
       setTemplate('camelot', compile('<section><h3 class="silly">Is a silly place</h3></section>'));
-
-      originalLoggerError = Logger.error;
+      
+      originalLoggerError = console.error; // eslint-disable-line no-console
```
rwjblue: Can you also update the variable name here (so it doesn't refer to logger)?

packages/ember-glimmer/lib/helpers/log.ts
```diff
@@ -24,8 +22,10 @@ import Logger from 'ember-console';
   @public
 */
 function log({ positional }: CapturedArguments) {
-  Logger.log.apply(null, positional.value());
-}
+    /* eslint-disable no-console */
+    console.log.apply(console, positional.value());
```
rwjblue: can you modify to `console.log(...positional.value())`?

rwjblue (changes requested):

lupestro:
Sorry for the big stack of stuff - shouldn't be much actual change - I sync'd up my fork/master and rebased the branch... no conflicts reported and everything tested out fine - this should have everything you requested...

rwjblue (approved):

rwjblue:
🎉

rwjblue:
Thanks @lupestro for all the hard work on this!

PR #16254, opened 2018-02-17
Re-enable missing tests

mixonic:
We were not running these tests because they lacked `@test` at the start of their method names.


mmun:
Looks like these tests are failing in node?

mmun:
Doh, was looking at the wrong tab :P Any idea why they're failing?

mixonic:
IE11 does not like it when you try to remove nodes from a parent they were already detached from. Moving the clear of `qunit-fixture` until after application teardown fixes exceptions in IE11.

rwjblue:
Thanks for tracking that down and landing this.

PR #15443, opened 2017-07-01
converted `ComputedProperty` to es6 class

bekzod:

PR #16260, opened 2018-02-19
Migrate ComputedProperty and InjectedProperty to native classes

mmun:
This is a cleanup of https://github.com/emberjs/ember.js/pull/15443. Details can be found in that thread.

Closes https://github.com/emberjs/ember.js/pull/15443.

### DIFF
--- a/packages/ember-metal/lib/injected_property.js
+++ b/packages/ember-metal/lib/injected_property.js
@@ -1,8 +1,6 @@
 import { getOwner } from 'ember-utils';
 import { assert } from 'ember-debug';
 import { ComputedProperty } from './computed';
-import { AliasedProperty } from './alias';
-import { Descriptor } from './properties';
 import { descriptorFor } from './meta';
 
 /**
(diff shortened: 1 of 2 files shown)


packages/ember-metal/lib/injected_property.js
```diff
@@ -21,12 +19,13 @@ import { descriptorFor } from './meta';
...
 */
-export default function InjectedProperty(type, name) {
-  this.type = type;
-  this.name = name;
+export default class InjectedProperty extends ComputedProperty {
+  constructor(type, name) {
+    super(injectedPropertyGet);
 
-  this._super$Constructor(injectedPropertyGet);
-  AliasedPropertyPrototype.oneWay.call(this);
```
mixonic: Was this doing something which is no longer needed?

mmun: That line was ensuring that you can clobber an injection if you set it. This is the default behaviour of getter-only, non-readonly computed properties, which we are now inheriting.

FWIW, there is a test specifically for this added in https://github.com/emberjs/ember.js/pull/9889

PR #16270, opened 2018-02-22
Es5 getters return raw descriptors

hjdivad:
With EMBER_METAL_ES5_GETTERS enabled, deprecate `Ember.get` implicitly calling `get` on a property that happens to be a descriptor, but which wasn't found via `descriptorFor`.

With EMBER_METAL_ES5_GETTERS disabled, and in prior versions of Ember,  descriptors on the instance are treated as special properties.  In fact, in earlier versions of Ember, all computed properties were stored this way.

paired with @rwjblue 

PR for the website deprecation: https://github.com/emberjs/website/pull/3209

### DIFF
--- a/packages/ember-metal/lib/property_get.js
+++ b/packages/ember-metal/lib/property_get.js
@@ -2,7 +2,7 @@
 @module @ember/object
 */
 
-import { assert } from 'ember-debug';
+import { assert, deprecate } from 'ember-debug';
 import { HAS_NATIVE_PROXY, symbol } from 'ember-utils';
 import { DESCRIPTOR_TRAP, EMBER_METAL_ES5_GETTERS, MANDATORY_GETTER } from 'ember/features';
 import { isPath } from './path_cache';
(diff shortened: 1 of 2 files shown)


mmun (approved):

packages/ember-metal/lib/property_get.js
```diff
@@ -96,6 +96,15 @@ export function get(obj, keyName) {
       if (DESCRIPTOR_TRAP && isDescriptorTrap(value)) {
         descriptor = value[DESCRIPTOR];
       } else if (isDescriptor(value)) {
+        deprecate(
+          `[DEPRECATED] computed property '${keyName}' was not set on object '${obj && obj.toString && obj.toString()}' via 'defineProperty'`,
+          !EMBER_METAL_ES5_GETTERS,
+          {
+            id: 'ember-meta.descriptor-on-object',
+            until: '3.5.0',
+            url: 'https://emberjs.com/deprecations/v3.x#toc_ember-computed-property-defineProperty'
```
mmun: For the link to work correctly, this should be the downcased header of the deprecation. E.g.

> Use notifyPropertyChange instead of propertyWillChange and propertyDidChange

becomes

> #toc_use-notifypropertychange-instead-of-propertywillchange-and-propertydidchange

hjdivad: @mmun thanks.  There will be an accompanying pr to emberjs/website although I need to work through exactly what the deprecation guide will be with @rwjblue.

PR #16271, opened 2018-02-23
[BUGFIX beta] Fix ChainNode unchaining

mmun:
A bug was introduced in #15850 where the last node in a chain was not being cleaned up correctly. This PR fixes this bug.

Relatedly, #15850 accidentally fixed a long standing bug with chains with single-character keys. This PR also includes a test for this second bug.

To assist with reviewing, here are a couple screenshots showing how the tests I've added fail on Ember 2.18 and on master before applying this fix.

Ember 2.18:

screen shot 2018-02-22 at 8 08 36 pm

master:

screen shot 2018-02-22 at 8 18 38 pm

### DIFF
--- a/packages/ember-metal/lib/chains.js
+++ b/packages/ember-metal/lib/chains.js
@@ -241,7 +241,7 @@ class ChainNode {
     let node = chains[key];
 
     // unchain rest of path first...
-    if (tails.length > 1) {
+    if (tails.length > 0) {
       node.unchain(tails.shift(), tails);
     }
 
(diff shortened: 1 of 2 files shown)


rwjblue (approved):

Turbo87:
thanks @mmun!

packages/ember-metal/lib/chains.js
```diff
@@ -241,7 +241,7 @@ class ChainNode {
     let node = chains[key];
 
     // unchain rest of path first...
-    if (tails.length > 1) {
+    if (tails.length > 0) {
```
Turbo87: I suspect this is somewhat of a hot path, right? should we use `!== 0` instead?

PR #16274, opened 2018-02-23
[BUGFIX beta] Ensure accessing a "proxy" itself does not error.

### LINKED ISSUE #16263: You attempted to access the `__DESCRIPTOR__` property (of <frontend@service:prototypes::ember581>)
Related to #16148

I have a service, injected in a route (extending the AuthenticatedRouteMixin). When I do `this.get('prototypes')`,so, in my use case I don't know if the error is coming from: 
 * The route accessing a service
 * The route itself
 * The service itself

Let me now if you need more information.
For now, I will try to revert 2.18 (ember 3.0.X seems to have problems with ie 11)

```
There is more information at https://github.com/emberjs/ember.js/issues/16148, and
you can help us improve this error message by telling us more about what happened in
this situation.
    at new EmberError (error.js:40)
    at assert (index.js:139)
    at Object.get (core_object.js:57)
    at isDescriptorTrap (ember-metal.js:3258)
    at get (ember-metal.js:3430)
    at Class.get (observable.js:38)
    at route.js:62
```
cc/ @wycats 
Maybe this is an issue in ember-simple-auth, so kindly cc/ to @marcoow

rwjblue:
Prior to this change, if `Ember.get` (or `this.get`) is used to access a property that is an Ember.Object with an `unknownProperty` that always returns a non-`undefined` value, the internals of `Ember.get` would check if that property is a "descriptor" and (due to the `unknownProperty` never returning `undefined`) the dev time assertion for accessing a proxy's content without using `Ember.get` would be triggered.

Fixes #16263

### DIFF
--- a/packages/ember-runtime/lib/system/core_object.js
+++ b/packages/ember-runtime/lib/system/core_object.js
@@ -98,6 +98,8 @@ function makeCtor() {
               property === 'willWatchProperty' ||
               property === 'didUnwatchProperty' ||
               property === 'didAddListener' ||
+              property === '__DESCRIPTOR__' ||
+              property === 'isDescriptor' ||
               property in target
             ) {
               return Reflect.get(target, property, receiver);
--- a/packages/ember-runtime/tests/system/core_object_test.js
+++ b/packages/ember-runtime/tests/system/core_object_test.js
@@ -1,3 +1,4 @@
+import { get } from 'ember-metal';
 import CoreObject from '../../system/core_object';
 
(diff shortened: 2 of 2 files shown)


wagenet:
To be clear, this should also resolve a similar situation where the proxy's content was destroyed and certain types of access to destroyed objects would raise assertions.

wagenet (approved):
This appears to resolve my issue.

sly7-7:
:heart: :party:

PR #16276, opened 2018-02-23
[WIP] [FEATURE] Add ability to use QUnit.todo

Serabe:
QUnit.todo works like QUnit.test, but in reverse. If no assertion fails,
the todo **does** fail.

While this can seem useless, it can boost contributions to Ember.js and
ease the triage work by knowing when an issue is actually fixed.


Serabe:
Work in progress. Needed to create a PR to see how this behaves in CI. After that, I'll update the CONTRIBUTING.

PR #16281, opened 2018-02-24
fix(link-to): add id to "inactive loading state" warning

lennyburdette:
Without an ID, this triggers an assertion with the message "When calling warn you must provide an options hash as the third parameter. options should include an id property."


rwjblue (approved):
Thank you!

PR #16286, opened 2018-02-25
[BUGFIX beta] Update to glimmer-vm@0.32.1.

rwjblue:
Fixes an issue with static numbers > 4000 being used in a template.


Turbo87:
Muchas gracias!

rwjblue:
This is already cherry-picked into the latest beta branch, so it should be included in tomorrows 3.1.0-beta.3....

PR #16289, opened 2018-02-26
Store guids in a WeakMap/Map

mmun:
This is a partial reimplementation of #15143. I didn't implement the part that changes guids to numbers.

Stores the guids in a WeakMap for objects and a Map for non-objects. Previously, weak maps were stored on the object.

I ran a benchmark with 100 samples and 8 times throttling on emberaddons.com. The results were insignificant:

I also added support for symbols.

[results.pdf](https://github.com/emberjs/ember.js/files/1756920/results.pdf)

### DIFF
--- a/packages/ember-utils/lib/guid.js
+++ b/packages/ember-utils/lib/guid.js
@@ -1,4 +1,6 @@
 import intern from './intern';
+import { isObject } from './spec';
+
 /**
  @module @ember/object
 */
@@ -35,9 +37,8 @@ export function uuid() {
 const GUID_PREFIX = 'ember';
 
 // Used for guid generation...
-const numberCache  = [];
-const stringCache  = {};
-
+const OBJECT_GUIDS = new WeakMap();
+const NON_OBJECT_GUIDS = new Map();
 /**
   A unique key used to assign guids and other private metadata to objects.
   If you inspect an object in your browser debugger you will often see these.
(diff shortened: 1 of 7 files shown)


packages/ember-utils/lib/guid.js
```diff
@@ -119,54 +98,33 @@ export function generateGuid(obj, prefix = GUID_PREFIX) {
...
+      guid = GUID_PREFIX + uuid();
+      OBJECT_GUIDS.set(value, guid);
+    }
+  } else {
+    guid = NON_OBJECT_GUIDS.get(value);
+
+    if (guid === undefined) {
+      if (typeof value === 'string') {
+        guid = 'st' + uuid();
+      } else if (typeof value === 'number') {
```
rwjblue: Not terribly significant, but I think I'd use a local variable here to avoid calling `typeof value` so many times in a row...

rwjblue (approved):
Looks nice to me, thanks for working on it!

We should update a few places to use `isObject`:

* https://github.com/emberjs/ember.js/blob/89fe61201a058c7a6edba705e96cccbd517df19c/packages/ember-metal/lib/property_get.js#L83
* https://github.com/emberjs/ember.js/blob/48c556444e0ac0cbbf232f7ae810b9d7d15f25f6/packages/ember-metal/lib/chains.js#L7

packages/ember-utils/lib/guid.js
mmun: Done.

PR #16293, opened 2018-02-26
[BUGFIX release] Ignore --pod for -addon blueprints: helper, initializer, and instance-initializer

jamescdavis:
In an addon, `helper`, `initializer`, and `instance-initializer` files are generated in the incorrect location under `app/` and attempt to import using an invalid module path when pod option is enabled.


rwjblue (approved):

rwjblue:
Thank you!

PR #16294, opened 2018-02-26
[BUGFIX beta] Fix input macro params handling

### LINKED ISSUE #16256: Input helper passed type does not work in Ember 3.1 beta
I was working on updating my apps and addons to 3.1 beta and noticed some strange behavior. I have components that wrap the `{{input}}` helper and only the first one was rendering an input. I created a reproduction here https://github.com/rwwagner90/inputs and I played around with it some. It appears the issue is passing `type` down as an argument. It chokes on:

```hbs
{{input type=type}}
```

I also tested this against https://s3.amazonaws.com/builds.emberjs.com/canary/shas/9300d1b7d3aa70987986b1997ecf02a542ea49be.tgz and there did not seem to be any difference.

Serabe:
Before this PR, having a component with an `{{input}}` with bound type
would make that input only to work the first time it was rendered. This
was because the first param would be shifter from params, and that would
modify the params of the rest of the instances.

This is breaking in current beta.

Fixes #16256

### DIFF
--- a/packages/ember-glimmer/tests/integration/helpers/input-test.js
+++ b/packages/ember-glimmer/tests/integration/helpers/input-test.js
@@ -446,6 +446,22 @@ moduleFor('Helpers test: {{input}} with dynamic type', class extends InputRender
 
     this.assertAttr('type', 'text');
   }
+
+  ['@test GH16256 input macro does not modify params in place']() {
+    this.registerComponent('my-input', {
+      template: `{{input type=inputType}}`
+    });
+
+    this.render(`{{my-input inputType=firstType}}{{my-input inputType=secondType}}`, {
+      firstType: 'password',
+      secondType: 'email'
+    });
+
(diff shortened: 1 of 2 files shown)


packages/ember-glimmer/tests/integration/helpers/input-test.js
```diff
@@ -446,6 +446,22 @@ moduleFor('Helpers test: {{input}} with dynamic type', class extends InputRender
 
     this.assertAttr('type', 'text');
   }
+
+  ['@test GH16256 input macro does not modify params in palce']() {
```
andyhot: typo for place

Serabe: Done. Thank you!

rwjblue (approved):

rwjblue:
Awesome sleuthing! Thank you!

Issue #16292, opened 2018-02-26
observers don't trigger in complex app


BryanCrotaz:
OK, this is really weird. I have a complex app with lots of addons involved where no observers are triggering. If I create a simple test route with a simple observer watching a plain string property it doesn't fire.  If I put that same code into a new application it works every time.

Using `ember-source 2.16.2, 2.16.4, 2.18.2`

This change (commenting the else) to `ember-metal/lib/property_events` fixes the symptoms.
```
function notifyObservers(obj, keyName, meta$$1) {
    if (meta$$1.isSourceDestroying()) {
      return;
    }

    var eventName = keyName + ':change';
    var listeners = void 0;
    if (deferred > 0) {
      listeners = observerSet.add(obj, keyName, eventName);
      accumulateListeners(obj, eventName, listeners, meta$$1);
    }// else {
      sendEvent(obj, eventName, [obj, keyName]);
    //}
  }
```

What could I have broken in my code that would cause this fix to work?

Or have I uncovered an edge case around Ember observers?

BryanCrotaz:
@rwjblue @cibernox - thoughts?

BryanCrotaz:
See PR #16295

BryanCrotaz:
@rwjblue @mmun as I suspected, the app is doing something to break the framework (e.g. deferring) - I would have been surprised if I'd found a bug in something this fundamental. However the app is huge with 20 or so addons so I can't reproduce this anywhere else.

What does the PR tell us about what the app has broken?

I'd like to find what pattern to search for in the source, and then to find a way to alert to a dev that they have broken the framework in this way, and add a warning to the appropriate docs.

BryanCrotaz:
Is deferring done with `beginPropertyChanges` and `endPropertyChanges`?

If so these appear to be global, so forgetting to call `endPropertyChanges` would seem to break all observers, application wide?

rwjblue:
My guess is an unbalanced call to `beginPropertyChanges`, you should add some logging around where the `deferred` variable is incremented and decremented. The thing to do is identify when it gets unbalanced and dig into what is doing things wrong...

mmun:
@BryanCrotaz that's right. you can put logging statements there (i.e. put a conditional breakpoint that contains `console.log('beginPropertyChanges', deferred)`

mmun:
You should also be able to put a debugger in `beginPropertyChange` and look up the callstack to see what is requesting the deferral.

[...]

PR #16295, opened 2018-02-26
Fix observers (#16292)

BryanCrotaz:
Seeing observers not firing in real world app (where they used to work). See issue #16292 

This PR fixes the issue - either it's a fix to an edge case, or it might point the way to what the app might have done to break the framework, in which case this advice might need to be added to the docs or guides.


Serabe:
These changes break tests.

BryanCrotaz:
@serabe I'm not clear what the `if` is doing here - what does it say about my app that removing the else makes my observers fire?

rwjblue:
You are making then fire even when deferred, something in your app must be deferring observers and never release the deferral

BryanCrotaz:
@rwjblue I wasn't aware of deferring - what does it mean in this context? And how do you defer an observer?

BryanCrotaz:
This is the observer that isn't firing:

```
// controller.js
import { later } from '@ember/runloop';
import { observer } from '@ember/object';
import { validator, buildValidations } from 'ember-cp-validations';
import { task, timeout } from 'ember-concurrency';
import { waitForProperty } from 'sc-pp-frontend/utils/wait-for';
const Validations = buildValidations({
  'data': validator('presence', true)
});

export default Controller.extend(Validations, {
  data: null,
  observerChanged: false,

  init() {
    this._super(...arguments);
    this.get('data');
  },

  dataChange: observer('data', function() {
    this.set('observerChanged', true);
    later(() => {
      this.set('observerChanged', false);
    }, 500);
  }),

  toggleTask: task(function*() {
    while(true) {
      yield timeout(1500);
      this.set('data', 'abc');
      yield timeout(1500);
      this.set('data', null);
    }
  }).on('init')
});


// template.hbs
TEST OBSERVER
<div>

  data: {{data}}
  <br/>
  Observer observed change: {{observerChanged}}
  <br/>
  {{input value=data }}

</div>
```

mmun:
@rwjblue's assessment is correct. The existing code is correct and something is wrong in your app / an addon. Can you push a demo app that contains the error?

mmun:
Please continue in https://github.com/emberjs/ember.js/issues/16292 as this PR is invalid (although it is indeed helpful for debugging the root cause :)).

PR #16300, opened 2018-02-27
Expose ability to detect serialization node.

rondale-sc:
This is dependent on this glimmer-vm PR to actually work https://github.com/glimmerjs/glimmer-vm/pull/788. Which means we'll need to wait until it is merged/bumped and Ember itself updates to use the release with it in it before we can land this.

The idea here is to insulate things that depend on the magic string we use to determine whether or not the DOM node we pass was produced by the serialization element builder.

This needs to be exposed on the global to ensure we can use it in the ember-cli-fastboot vendor file so we can determine how to boot the app when it is loaded in the browser.

To see where it'd be used in ember-cli-fastboot please see this PR with the in progress work:

https://github.com/ember-fastboot/ember-cli-fastboot/pull/580

### DIFF
--- a/packages/ember/lib/index.js
+++ b/packages/ember/lib/index.js
@@ -429,7 +429,9 @@ import {
   setTemplates,
   template,
   TextField,
-  TextArea
+  TextArea,
+  isSerializationFirstNode,
+  SERIALIZATION_FIRST_NODE_STRING
 } from 'ember-glimmer';
 
 Ember.Component = Component;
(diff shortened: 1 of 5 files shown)


packages/ember/lib/index.js
```diff
@@ -501,7 +503,9 @@ Ember.ViewUtils = {
   getViewClientRects: views.getViewClientRects,
   getViewBoundingClientRect: views.getViewBoundingClientRect,
   getRootViews: views.getRootViews,
-  getChildViews: views.getChildViews
+  getChildViews: views.getChildViews,
+  isSerializationFirstNode: isSerializationFirstNode,
+  SERIALIZATION_FIRST_NODE_STRING: SERIALIZATION_FIRST_NODE_STRING
```
rwjblue: I’d rather not export this string for now (we can always come back and do it later if needed)...

rondale-sc: Done. :+1:

rwjblue:
Just published glimmer-vm 0.32.2 with the function exported, mind updating?

rwjblue (approved):

PR #16302, opened 2018-02-28
Publicise change properties

BryanCrotaz:
See @mmun comment in https://github.com/emberjs/ember.js/issues/16292#issuecomment-369129781

> But you shouldn't use those functions because they are private. You should be using object.setProperties(...) or
> 
> changeProperties(() => {
>   doStuff();
> });
> which does the try-finally for you. (Hmm, unfortunately it looks like changeProperties is accidentally marked private.)


BryanCrotaz:
How do you get rid of all the other commits?
I've pulled upstream master into my fork master and then branched from there with a single commit - why does the PR show all the commits in that pull?

locks:
@BryanCrotaz according to the commit log, you did a commit on your master and then merged the upstream master into your master (by pulling), adding all of these commits. It might be easier to make a new PR and close this one?

locks:
Ah, I see the replacement!

PR #16305, opened 2018-02-28
cleanup `if statements` in `property_get`

bekzod:
gave little improvement in benchmarks https://cl.ly/0V1f1g19103B/results.pdf

### DIFF
--- a/packages/ember-metal/lib/property_get.js
+++ b/packages/ember-metal/lib/property_get.js
@@ -95,56 +95,51 @@ export function get(obj, keyName) {
   let isFunction = type === 'function';
   let isObjectLike = isObject || isFunction;
 
-  let descriptor;
+  if (!isObjectLike) {
+    return obj[keyName];
+  }
+
+  if (EMBER_METAL_TRACKED_PROPERTIES) {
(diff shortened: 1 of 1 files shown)


krisselden:
This is out of date.

packages/ember-metal/lib/property_get.js
```diff
@@ -95,56 +95,51 @@ export function get(obj, keyName) {
...
+  if (descriptor === undefined) {
+    if (DEBUG && HAS_NATIVE_PROXY) {
+      value = getPossibleMandatoryProxyValue(obj, keyName);
+    } else {
+      value = obj[keyName];
     }
 
-    if (descriptor !== undefined) {
-      return descriptor.get(obj, keyName);
+    if (isDescriptor(value)) {
```
krisselden: it is weird to do this call, when at this point descriptor could be defined and value === undefined

bekzod: maybe add test case what you mean or suggest change :P ? I just reorganized `if`s but the logic is same as was before

bekzod:
I think this is out of date

PR #16306, opened 2018-02-28
allow ember/helpers/link-to tests to run without jQuery

nlfurniss:
This removes jQuery usage in the `ember/helpers/link-to` tests as part of #16058

Status

- No tests fail w/o jQuery

- 2 tests are failing without jQuery related to bubbling; am investigating


ro0gr:
There is a similar pr which is also blocked by those bubbling tests https://github.com/emberjs/ember.js/pull/16130

Hope it will save you some time :)

nlfurniss:
Thanks, @ro0gr should have looked more carefully through the PR list. I guess I'll also bang my head against it for a little while :-)

PR #16307, opened 2018-03-01
[BUGFIX beta] Ensure proper .toString() of default components.

rwjblue:
In older Ember versions the default components were generally labelled as `Ember.Component` or `Ember.TextField` via the `NAME_KEY` property on the class itself. During the glimmer-vm upgrade which landed during the 3.1 canary cycle the custom `NAME_KEY` was lost for `Ember.TextField`.

This commit removes needless usage of `NAME_KEY` (not needed since our normal `Ember.Object` instance's `.toString` methods check for the factories own `toString` properly), and updates the actual paths to match the new JS modules API locations (thus reducing more "globals" shenanigans).

### DIFF
--- a/packages/ember-glimmer/lib/component.ts
+++ b/packages/ember-glimmer/lib/component.ts
@@ -9,7 +9,7 @@ import {
   PROPERTY_DID_CHANGE,
 } from 'ember-metal';
 import { TargetActionSupport } from 'ember-runtime';
-import { getOwner, NAME_KEY, symbol } from 'ember-utils';
+import { getOwner, symbol } from 'ember-utils';
 import {
   ActionSupport,
   ChildViewsSupport,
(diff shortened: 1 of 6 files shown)


mmun:
j/w where is this relied on?

mmun (approved):

rwjblue:
`CoreObject.prototype.toString` uses the factories `toString` for the instances own toString display.

https://github.com/emberjs/ember.js/blob/8ef8dc63cbc760e52692f17e75210f3e3515804e/packages/ember-runtime/lib/system/core_object.js#L571-L578

normally that means that we call `makeToString` on the resolver

https://github.com/emberjs/ember.js/blob/master/packages/container/lib/container.js#L384-L390

https://github.com/ember-cli/ember-resolver/blob/master/addon/resolvers/classic/index.js#L131-L133

but these specific classes are not in the resolver

rwjblue:
The reason I started digging into this (@rwwagner90 and I paired on this PR) was that the inspector had some tests that were trying to confirm that we had a valid string representation.

mmun:
:shipit:

Issue #16301, opened 2018-02-28
[QUEST] Glimmer Components in Ember


tomdale:
This quest issue tracks the implementation of Glimmer components in Ember.js. 

# The Plan

Glimmer.js components have the following features:

1. “Outer HTML” templates (no `tagName`, `attributeBindings`, etc.)
2. Arguments in templates are `@` prefixed, like `{{@firstName}}`
3. Arguments are set on the component as `this.args`
4. Component classes use JavaScript class syntax
5. Mutable component state is annotated with `@tracked` properties
6. Components are invoked via `<AngleBracket />` syntax
7. Attributes can be “splatted” via `…attributes`

In keeping with Ember’s spirit of incrementalism, we want to land this functionality piece by piece via an addon. This allows the community to start using features and providing feedback early in the process.

In fact, we’ve already started down the road to Glimmer components in Ember. The first two features have already started to land:

1. In template-only components, templates are “outer HTML” (assuming the `template-only-glimmer-components` [optional feature](https://github.com/emberjs/ember-optional-features) has been enabled.)
2. Arguments can be accessed in the template via the `@` prefix (e.g. `{{@firstName}}`).

This issue proposes finishing the process of bringing Glimmer components to Ember by allowing addons to provide alternate component implementations, then transforming the `@glimmer/component` package into an Ember addon that implements the Glimmer.js component API.

We’ll break that work into phases, each one unlocking benefits for existing Ember apps and addons.  Phase 0 is about adding the necessary primitives to Ember.js to support alternate component implementations. Phases 1, 2 and 3 are about incrementally enabling Glimmer.js component API.

While we go into depth on Phases 0 and 1, we will defer exploring the technical details of later phases until the first phases are closer to completion.

## Phase 0: Customizing Component Behavior

**TL;DR: Add a CustomComponentManager API to Ember.js to allow addons to implement custom component API.**

Currently, all components in an Ember app are assumed to be subclasses of `Ember.Component`.  In order to support alternate component APIs in Ember, we need some way to tell Ember when and how component behavior should change.

When we say “custom component behavior,” we specifically mean:

1. How component instances are created.
2. How component instances are destroyed.
3. How arguments are provided to the component instance.
4. How the component is notified of these lifecycle changes.

While Glimmer VM introduces the concept of a “component manager,” an object that makes these decisions, this API is very low level. It would be premature to adopt them directly as public API in Ember because they are difficult to write, easy to author in a way that breaks other components, and not yet stable.

[...]

PR #16308, opened 2018-03-01
Custom component manager

smfoote:
Add custom component manager and tests. See #16301 and ember-cli/ember-rfc176-data#43.

### DIFF
--- a/packages/ember-glimmer/lib/index.ts
+++ b/packages/ember-glimmer/lib/index.ts
@@ -303,4 +303,5 @@ export { UpdatableReference, INVOKE } from './utils/references';
 export { default as iterableFor } from './utils/iterable';
 export { default as DebugStack } from './utils/debug-stack';
 export { default as OutletView } from './views/outlet';
-export { default as CustomComponentManager, COMPONENT_MANAGER } from './component-managers/custom';
+export { default as CustomComponentManager } from './component-managers/custom';
+export { COMPONENT_MANAGER, componentManager } from './utils/custom-component-manager';
--- a/packages/ember-glimmer/lib/index.ts
+++ b/packages/ember-glimmer/lib/index.ts
@@ -303,3 +303,4 @@ export { UpdatableReference, INVOKE } from './utils/references';
 export { default as iterableFor } from './utils/iterable';
 export { default as DebugStack } from './utils/debug-stack';
 export { default as OutletView } from './views/outlet';
+export { default as CustomComponentManager, COMPONENT_MANAGER } from './component-managers/custom';
--- a/packages/ember-glimmer/lib/utils/custom-component-manager.ts
+++ b/packages/ember-glimmer/lib/utils/custom-component-manager.ts
@@ -10,6 +10,17 @@ import { GLIMMER_CUSTOM_COMPONENT_MANAGER } from 'ember/features';
 
 export const COMPONENT_MANAGER = symbol('COMPONENT_MANAGER');
 
+export function componentManager(obj: any, managerId: String) {
+  if ('reopenClass' in obj) {
+    return obj.reopenClass({
+      [COMPONENT_MANAGER]: managerId
+    });
+  }
+
+  obj[COMPONENT_MANAGER] = managerId;
+  return obj;
+}
+
 export default function getCustomComponentManager(owner: Owner, obj: {}): ComponentManager<ComponentStateBucket, DefinitionState> | undefined {
   if (!GLIMMER_CUSTOM_COMPONENT_MANAGER) { return; }
 
@@ -19,7 +30,7 @@ export default function getCustomComponentManager(owner: Owner, obj: {}): Compon
   if (!managerId) { return; }
 
   let manager = owner.lookup(`component-manager:${managerId}`) as ComponentManager<ComponentStateBucket, DefinitionState>;
-  assert(`Could not find custom component manager '${managerId}'`, !!manager);
+  assert(`Could not find custom component manager '${managerId}' for ${obj}`, !!manager);
 
   return manager;
 }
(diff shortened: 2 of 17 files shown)


packages/ember-glimmer/tests/integration/custom-component-manager-test.js
```diff
@@ -0,0 +1,252 @@
...
+
+if (GLIMMER_CUSTOM_COMPONENT_MANAGER) {
+  moduleFor('Components test: curly components with custom manager', class extends RenderingTest {
+    /*
+     * Helper to register a custom component manager. Provides a basic, default
+     * implementation of the custom component manager API, but can be overridden
+     * by passing custom hooks.
+     */
+    registerCustomComponentManager(overrides = {}) {
+      let options = Object.assign({
```
rwjblue: Needs to be updated to use assign from ember-utils (Object.assign isn’t available on all tested platforms).

packages/ember-glimmer/tests/utils/helpers.js
```diff
@@ -14,5 +14,6 @@ export {
   htmlSafe,
   SafeString,
   DOMChanges,
-  isHTMLSafe
+  isHTMLSafe,
+  COMPONENT_MANAGER
```
rwjblue: I don’t believe this is needed

packages/ember-views/lib/system/utils.js
```diff
@@ -88,18 +88,25 @@ export function getChildViews(view) {
...
+
+  childViews.push(getViewId(child));
 }
 
 export function collectChildViews(view, registry) {
   let ids = [];
   let views = [];
 
-  view[CHILD_VIEW_IDS].forEach(id => {
+  (CHILD_VIEW_IDS.get(view) || []).forEach(id => {
```
rwjblue: Seems like this should be split up on multiple lines, and in the case where `CHILD_VIEW_IDS.get(view)` is undefined we can return early (e.g. no need for the `forEach`)...

packages/ember/lib/index.js
```diff
@@ -439,6 +440,11 @@ Ember.TextField = TextField;
 Ember.TextArea = TextArea;
 Ember.LinkComponent = LinkComponent;
 
+Object.defineProperty(Ember, '_componentManager', {
```
rwjblue: Let’s call it `setComponentManager` here?

rwjblue (changes requested):

packages/ember-glimmer/lib/index.ts
```diff
@@ -300,7 +300,9 @@ export { default as AbstractComponentManager } from './component-managers/abstra
 // rather than the problem was solved
 // DebugStack should just test the assert message
 // it supports for example
-export { UpdatableReference } from './utils/references';
+export { UpdatableReference, RootReference } from './utils/references';
 export { default as iterableFor } from './utils/iterable';
 export { default as DebugStack } from './utils/debug-stack';
 export { default as OutletView } from './views/outlet';
+export { default as CustomComponentManager } from './component-managers/custom';
```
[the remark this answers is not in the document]
smfoote: Probably not.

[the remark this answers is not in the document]
smfoote: Actually, it's needed in the test file, which does not seem to be able to import from `lib`.

packages/ember/lib/index.js
smfoote: sans underscore?

packages/ember-views/lib/system/utils.js
```diff
@@ -88,27 +88,35 @@ export function getChildViews(view) {
...
 
 export function collectChildViews(view, registry) {
   let ids = [];
   let views = [];
+  let childViews = CHILD_VIEW_IDS.get(view);
 
-  view[CHILD_VIEW_IDS].forEach(id => {
-    let view = registry[id];
+  if (childViews) {
+    (CHILD_VIEW_IDS.get(view) || []).forEach(id => {
```
rwjblue: I don't think you need to query the weakmap again here, you can reliably do `childViews.forEach` directly (due to the guard

packages/ember/lib/index.js
```diff
@@ -439,6 +440,11 @@ Ember.TextField = TextField;
 Ember.TextArea = TextArea;
 Ember.LinkComponent = LinkComponent;
 
+Object.defineProperty(Ember, 'setComponentManager', {
```
rwjblue: darn, I'm sorry for the run around here (I know you implemented exactly what I said in the last comment), I think this still needs to be underscored

Can you change it to be `_setComponentManager`?

smfoote: Haha, I did `setComponentManager` first, then backed off. Shoulda gone with my gut.

packages/ember-views/lib/system/utils.js
smfoote: Shoot - you're totally right. I missed that.

packages/ember-glimmer/lib/utils/custom-component-manager.ts
```diff
@@ -0,0 +1,39 @@
...
+    return obj.reopenClass({
+      [COMPONENT_MANAGER]: managerId
+    });
+  }
+
+  return class extends obj {
+    static get [COMPONENT_MANAGER]() {
+      return managerId;
+    }
+  };
```
chancancode: Do we want to do this here instead of just mutating the class? The two branches do different things

smfoote: Are you saying this should be

```
obj[COMPONENT_MANAGER] = managerId;
```

instead of returning a new, extended class? That works for me. @rwjblue ?

rwjblue: Yeah, sounds good to me

rwjblue (approved):

PR #16309, opened 2018-03-01
Acceptance testing blueprint will use the app's setup-application-test blueprint.

ryanto:
See https://github.com/ember-cli/ember-cli/pull/7657 for the ember-cli part of this.

### DIFF
--- a/blueprints/acceptance-test/qunit-rfc-232-files/tests/acceptance/__name__-test.js
+++ b/blueprints/acceptance-test/qunit-rfc-232-files/tests/acceptance/__name__-test.js
@@ -1,6 +1,6 @@
 import { module, test } from 'qunit';
 import { visit, currentURL } from '@ember/test-helpers';
-import { setupApplicationTest } from 'ember-qunit';
+import setupApplicationTest from '<%= dasherizedPackageName %>/tests/helpers/setup-application-test';
 
 module('<%= friendlyTestName %>', function(hooks) {
   setupApplicationTest(hooks);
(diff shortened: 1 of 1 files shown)


blueprints/acceptance-test/qunit-rfc-232-files/tests/acceptance/__name__-test.js
```diff
@@ -1,6 +1,6 @@
 import { module, test } from 'qunit';
 import { visit, currentURL } from '@ember/test-helpers';
-import { setupApplicationTest } from 'ember-qunit';
+import setupApplicationTest from '<%= dasherizedPackageName %>/tests/helpers/setup-application-test';
```
bendemboski: Should use `<%= testFolderRoot %>` here, same as [here](https://github.com/emberjs/ember.js/blob/master/blueprints/acceptance-test/qunit-files/tests/acceptance/__name__-test.js#L2), no?

bendemboski (changes requested):

blueprints/acceptance-test/qunit-rfc-232-files/tests/acceptance/__name__-test.js
Turbo87: I'm wondering if we would need to detect if the file actually exists and if not fall back to the old behavior? 🤔

bendemboski: Oof, that would be a pain...I think that would involve `import require from 'require'` and using `require.has()` to import rather than ES6 imports, right?

bendemboski: Oh, or you mean detecting at blueprint execution time (rather than runtime)?

Turbo87: yes, that is what I meant. I roughly remember that we already did something like that somewhere for the `destroy-app.js` file 🤔

PR #16312, opened 2018-03-02
[BUGFIX beta] Rename private property Route#router to Router#_router

### LINKED ISSUE #16088: Can not use router service in route
Hi,

when I inject [RouterService](https://emberjs.com/api/ember/2.18/classes/RouterService) into some route, this property is replaced by `router:main`.

```javascript
import Route from '@ember/routing';
import { inject } from '@ember/service';

export default Route.extend({
  router: inject()
});
```

[API](https://emberjs.com/api/ember/2.18/classes/Route?show=inherited%2Cprotected%2Cprivate) does not expose `router` property (public or private), I can use named inject like `routing: inject('router')`, but it seems to me, that this should work out of box?

The intent is to use `ember-cli-meta-tags` and [urlFor()](https://emberjs.com/api/ember/2.18/classes/RouterService/methods/urlFor?anchor=urlFor) to generate URL for `rel` meta tags and the `router:main` does not have this method available.

Thanks

cibernox:
The fact that this private property was named `router` was preventing users
from injecting the router service into routes like this:

```js
export default Route.extend({
  router: service()
});
```

I consider that the above code is pretty reasonable, and it should be
the private property the one that should be renamed to a clearer `_router`
name that is less likely to collide with user-defined properties or methods.

If you deem this an _intimate_ API we could consider adding a deprecated
alias, but I don't have evidence to consider it such, as `router` is too
common of a term to search for in emberobserver.

Closes #16088

### DIFF
--- a/packages/ember-application/lib/system/engine.js
+++ b/packages/ember-application/lib/system/engine.js
@@ -502,7 +502,7 @@ function commonSetupRegistry(registry) {
   registry.injection('router', '_bucketCache', P`-bucket-cache:main`);
   registry.injection('route', '_bucketCache', P`-bucket-cache:main`);
 
-  registry.injection('route', 'router', 'router:main');
+  registry.injection('route', '_router', 'router:main');
 
   // Register the routing service...
   registry.register('service:-routing', RoutingService);
--- a/packages/ember-application/tests/system/application_test.js
+++ b/packages/ember-application/tests/system/application_test.js
@@ -135,7 +135,7 @@ moduleFor('Application', class extends ApplicationTestCase {
     verifyInjection(assert, application, 'router', '_bucketCache', P`-bucket-cache:main`);
     verifyInjection(assert, application, 'route', '_bucketCache', P`-bucket-cache:main`);
 
-    verifyInjection(assert, application, 'route', 'router', 'router:main');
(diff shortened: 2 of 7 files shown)


cibernox:
@Serabe ^

rwjblue (changes requested):
Yes, I think we need to deprecate `this.router` in `Ember.Route` instances (and add an entry on the website to describe how to handle the deprecation). 

When folks inject the router service, it would clobber this "deprecating getter" and things should work just fine...

cibernox:
Working on it

cibernox:
@rwjblue Done. I considered this a bugfix beta and I added the deprecation in 3.1, but let me know if you prefer it to stay in canary and go in 3.2.

rwjblue (approved):

rwjblue:
> let me know if you prefer it to stay in canary and go in 3.2

yeah, lets move to 3.2 on the deprecations app.  I don't want to land a last minute deprecation just before release that ends up souring the yummy goodness of 3.1...

cibernox:
@rwjblue Documented in the new deprecations app in https://github.com/ember-learn/deprecation-app/pull/106

Updated this PR so the URL points to the right anchor in the new app.

PR #16315, opened 2018-03-03
Remove function marked as deprecated until 3.0.0

cibernox:
IDK if this is here intentionally or it's just an oversight, but it should be safe to remove it now


locks:
I think you still need to remove it from the `Ember` namespace building code: https://travis-ci.org/emberjs/ember.js/jobs/348773814#L1108

rwjblue:
We need to confirm this isn’t needed for supporting the Ember-2-legacy addon

cibernox:
@rwjblue I revisited this and realized there was much more to it. Not it's fully removed.

rwjblue (changes requested):
Thanks for working on this!  You have essentially removed the `Ember.required()` concept here (which is great!).  Unfortunately, we have committed to maintaining support for this through Ember 3.4 when using the `ember-2-legacy` addon. 

tldr; This PR needs to be held from merging until master is representing 3.5.

cibernox:
It seems too entangled to port it to `ember-2-legacy`, so let's wait.

krisselden:
required though does nothing because everything does applyPartial, if the addon just made it not error but do nothing it would likely not affect anything in the wild. This feature has not really worked since before 1.0, it only does something for apply() and everything else uses extend which uses applyPartial

cibernox:
@krisselden you're saying thats if I add a dummy Ember.required in the addon we don't have to wait to remove it?

cibernox:
@rwjblue @krisselden Since this feature wasn't really working this should not have any impact in the wild other than the lost of "Ember.required", so I added a "placebo polyfill" that returns an empty function: https://github.com/emberjs/ember-2-legacy/pull/66

cibernox:
This feature has been "polyfilled" (kind-of) in the ember-2-legacy-addon, so this is ready for another review.

cibernox:
@rwjblue I this on time for the new release?

packages/ember/tests/reexports_test.js
```diff
@@ -256,6 +255,34 @@ let allExports = [
   ['ContainerDebugAdapter', 'ember-extension-support'],
 ];
 
-if (ENV._ENABLE_PROPERTY_REQUIRED_SUPPORT) {
-  allExports.push(['required', 'ember-metal']);
+allExports.forEach(reexport => {
```
rwjblue: I think you have a rebase issue here, if you notice this loop is now included above on line 10.

cibernox: Fixed

rwjblue (approved):

PR #16319, opened 2018-03-04
Namespaced invocation

mixonic:
Replaces https://github.com/emberjs/ember.js/pull/16158

This commit adds support for invoking a component or injecting a service from a module unification namespace. It is in alignment with [RFC #309](https://github.com/emberjs/rfcs/pull/309).

Use `::` to delimit namespaces from the name of an invocation or injection. For example:

```hbs
{{! src/ui/components/app-component/template.hbs }}
{{my-addon::addon-component}}
```

```js
// src/ui/components/app-component/component.js
import { inject } from '@ember/services';

  session: inject('ember-simple-auth::session')
});
```

### DIFF
--- a/packages/container/lib/container.js
+++ b/packages/container/lib/container.js
@@ -146,21 +146,15 @@ export default class Container {
     let normalizedName = this.registry.normalize(fullName);
 
     assert('fullName must be a proper full name', this.registry.isValidFullName(normalizedName));
+    assert(
+      'EMBER_MODULE_UNIFICATION must be enabled to pass a namespace option to factoryFor',
+      EMBER_MODULE_UNIFICATION || !options.namespace
+    );
 
-    if (options.source) {
-      let expandedFullName = this.registry.expandLocalLookup(fullName, options);
-      // if expandLocalLookup returns falsey, we do not support local lookup
-      if (!EMBER_MODULE_UNIFICATION) {
-        if (!expandedFullName) {
-          return;
-        }
-
-        normalizedName = expandedFullName;
-      } else if (expandedFullName) {
-        // with ember-module-unification, if expandLocalLookup returns something,
-        // pass it to the resolve without the source
-        normalizedName = expandedFullName;
-        options = {};
+    if (options.source || options.namespace) {
+      normalizedName = this.registry.expandLocalLookup(fullName, options);
+      if (!normalizedName) {
+        return;
       }
     }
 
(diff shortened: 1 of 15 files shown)


packages/container/lib/container.js
```diff
@@ -206,22 +200,16 @@ function isInstantiatable(container, fullName) {
 }
 
 function lookup(container, fullName, options = {}) {
-  let normalizedName = fullName;
-  if (options.source) {
-    let expandedFullName = container.registry.expandLocalLookup(fullName, options);
+  if (!EMBER_MODULE_UNIFICATION) {
```
rwjblue: Should be part of assertion condition

packages/ember-glimmer/lib/resolver.ts
```diff
@@ -208,15 +211,17 @@ export default class RuntimeResolver implements IRuntimeResolver<OwnedTemplateMe
...
+  private _lookupHelper(_name: string, meta: OwnedTemplateMeta): Option<Helper> {
+    const helper = this.builtInHelpers[_name];
     if (helper !== undefined) {
       return helper;
     }
 
     const { owner, moduleName } = meta;
 
-    const options: LookupOptions | undefined = makeOptions(moduleName);
+    const {name, namespace} = this._parseNameForNamespace(_name);
```
rwjblue: Can you guard this in the feature flag?

packages/ember-glimmer/lib/resolver.ts
```diff
@@ -257,8 +262,21 @@ export default class RuntimeResolver implements IRuntimeResolver<OwnedTemplateMe
...
+    if (namespaceDelimiterOffset !== -1) {
+      name = _name.slice(namespaceDelimiterOffset+2);
+      namespace = _name.slice(0, namespaceDelimiterOffset);
+    }
+
+    return {name, namespace};
+  }
+
+  private _lookupComponentDefinition(_name: string, meta: OwnedTemplateMeta): Option<ComponentDefinition> {
+    let {name, namespace} = this._parseNameForNamespace(_name);
```
rwjblue: Please guard this in the feature flag?

packages/ember-metal/lib/injected_property.js
```diff
@@ -24,8 +24,20 @@ export default class InjectedProperty extends ComputedProperty {
...
     this.type = type;
-    this.name = name;
     this.source = options ? options.source : undefined;
+
+    if (name) {
+      let namespaceDelimiterOffset = name.indexOf('::');
+      if (namespaceDelimiterOffset === -1) {
+        this.name = name;
+        this.namespace = undefined;
+      } else {
```
rwjblue: Let’s tweak this else case  to ensure it’s feature flagged

rwjblue (dismissed):
Looking really good! Only had a few inline comments

packages/ember-metal/lib/injected_property.js
mixonic: This is already guarded in: https://github.com/emberjs/ember.js/pull/16240/files#diff-11c397e816cb737423d40c0a91104f5fR40

`packages/ember-runtime/lib/inject.js` ensures that `options` is only passed if the feature flag is enabled 👍

rwjblue: Awesome, thank you

PR #16326, opened 2018-03-06
If syntax error message

### LINKED ISSUE #16322: Unhelpful syntax error message
I just updated to ember 3 and got the following error (everything else is working fine):

```
Uncaught Error: SYNTAX ERROR: #if requires a single argument
```

Considering the size of the codebase, I can't figure out where this error comes from (already looked at all `#if` so I suppose I missed one...) and didn't find out how or where it was generated in this repository so I can't send a PR either.

Could you add the full `#if` syntax please? It'd be really helpful...

AlexTraher:
This (hopefully) resolves [issue 16322](https://github.com/emberjs/ember.js/issues/16322) regarding an unclear error message when passing an incorrect number of arguments to an `if` helper.

### DIFF
--- a/packages/ember-template-compiler/lib/plugins/assert-if-helper-without-arguments.js
+++ b/packages/ember-template-compiler/lib/plugins/assert-if-helper-without-arguments.js
@@ -1,0 +1,47 @@
+import { assert } from 'ember-debug';
+import calculateLocationDisplay from '../system/calculate-location-display';
+
+export default function assertIfHelperWithoutArguments(env) {
+  let { moduleName } = env.meta;
+
+  return {
+    name: 'assert-if-helper-without-arguments',
+
+    visitor: {
+      BlockStatement(node) {
+        if (isInvalidBlockIf(node)) {
+          assert(`${blockAssertMessage(node.path.original)} ${calculateLocationDisplay(moduleName, node.loc)}`);
+        }
+      },
+
+      MustacheStatement(node) {
+        if (isInvalidInlineIf(node)) {
+          assert(`${inlineAssertMessage(node.path.original)} ${calculateLocationDisplay(moduleName, node.loc)}`);
+        }
+      },
+
+      SubExpression(node) {
+        if (isInvalidInlineIf(node)) {
+          assert(`${inlineAssertMessage(node.path.original)} ${calculateLocationDisplay(moduleName, node.loc)}`);
+        }
+      }
+    }
+  };
+}
+
+function blockAssertMessage(original) {
+  return `#${original} requires a single argument.`;
+}
+
+function inlineAssertMessage(original) {
+  return `The inline form of the '${original}' helper expects two or three arguments.`;
+}
+
+function isInvalidInlineIf(node) {
+  return node.path.original === 'if' && (!node.params || node.params.length < 2 || node.params.length > 3);
+}
+
+function isInvalidBlockIf(node) {
+  return node.path.original === 'if' && (!node.params || node.params.length !== 1);
+}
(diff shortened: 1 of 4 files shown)


packages/ember-template-compiler/tests/plugins/assert-if-helper-without-arguments-test.js
```diff
@@ -0,0 +1,34 @@
+import { compile } from '../../index';
+import { moduleFor, AbstractTestCase } from 'internal-test-helpers';
+
+moduleFor('ember-template-compiler: assert-if-helper-without-argument', class extends AbstractTestCase {
+  [`@test block if helper expects one argument`](assert) {
+    assert.throws(() => {
```
rwjblue: I believe this should be `expectAssertion`

packages/ember-template-compiler/tests/plugins/assert-if-helper-without-arguments-test.js
```diff
@@ -0,0 +1,34 @@
...
+       compile(`{{#if}}aVal{{/if}}`, {
+        moduleName: 'baz/foo-bar'
+      });
+    }, `#if requires a single argument. ('baz/foo-bar' @ L1:C0) `);
+
+    assert.throws(() => {
```
rwjblue: `expectAssertion`

packages/ember-template-compiler/tests/plugins/assert-if-helper-without-arguments-test.js
```diff
@@ -0,0 +1,34 @@
...
+
+    expectAssertion(() => {
+      compile(`{{#if}}aVal{{/if}}`, {
+        moduleName: 'baz/foo-bar'
+      });
+    }, `#if requires a single argument. ('baz/foo-bar' @ L1:C0) `);
+  }
+
+  [`@test inline if helper expects between one and three arguments`](assert) {
+    assert.throws(() => {
```
rwjblue: `expectAssertion`

rwjblue (changes requested):
Looking good, thanks for picking this up!

I left some inline comments (mostly just moving from `assert.throws` to `expectAssertion` which should fix an issue with the production build), but would also like to see some tests added confirming a few additional scenarios work:

* `{{some-thing foo=(if)}}` should error
* `{{some-thing foo=(if foo bar baz)}}` should not error
* `{{if foo bar baz}}` should not error

packages/ember-template-compiler/lib/plugins/assert-if-helper-without-arguments.js
```diff
@@ -0,0 +1,43 @@
...
+    }
+  };
+}
+
+function assertMessage(original) {
+  return `${original} requires a single argument.`;
+}
+
+function isInvalidInlineIf(node) {
+  return node.path.original === 'if' && (!node.params || node.params.length < 1);
```
rwjblue: Should we make the minimum length be 2?  I'm not sure how something like `{{if something}}` is useful (e.g. if there is no truthy or falsey argument, what is even happening?)...

What do you think @AlexTraher?

AlexTraher: @rwjblue Yeah agreed - I like the idea it fails loudly.

rwjblue:
Looks great! Can you squash down to a single commit?

rwjblue (approved):
Awesome job here, thank you very much!

AlexTraher:
Thanks for making the time to give feedback! :+1:

PR #16351, opened 2018-03-09
[BUGFIX beta] Bring RSVP.cast back from the dead

### LINKED ISSUE #16257: Uncaught TypeError: RSVP.cast is not a function
It seems that Ember v3.1.0-beta.1 silently updated RSVP to a version that no longer includes `RSVP.cast()`, but unfortunately there was no deprecation warning prior to this. `ember-intl` seems to use that function internally though which broke our tests :-/

/cc @jasonmit

Serabe:
Fixes #16257

### DIFF
--- a/package.json
+++ b/package.json
@@ -122,7 +122,7 @@
     "qunit": "^2.5.0",
     "route-recognizer": "^0.3.3",
     "router_js": "2.0.0-beta.3",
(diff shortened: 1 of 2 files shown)


rwjblue (approved):

rwjblue:
Thank you!

PR #16354, opened 2018-03-10
avoid creating `Map` in `dependentKeysDidChange`

bekzod:


packages/ember-metal/lib/property_events.js
```diff
@@ -116,22 +116,23 @@ function notifyPropertyChange(obj, keyName, _meta) {
...
-    seen = DID_SEEN = new Map();
+  if (IS_TOP_SEEN_MAP) {
+    IS_TOP_SEEN_MAP = false;
   }
 
   iterDeps(notifyPropertyChange, obj, depKey, seen, meta);
 
-  if (top) {
-    DID_SEEN = null;
+  if (IS_TOP_SEEN_MAP) {
```
krisselden: Won't this always be false? did you mean to set the local `top = IS_TOP_SEEN_MAP` and use top in the conditionals?

Do we have a test that covers this?

krisselden (changes requested):

packages/ember-metal/lib/property_events.js
bekzod: thanks for catch, fixed, not sure how to test this, though it is doing same thing as before

krisselden:
@rwjblue we definitely have lost test coverage if the seen map clearing doesn't matter

krisselden:
I didn't mean to close it, I hit the wrong button

bekzod:
yeah happens all the time to me ;P

krisselden (approved):

PR #16355, opened 2018-03-10
conditionally add bindings related methods to meta

bekzod:
`ENV._ENABLE_BINDING_SUPPORT` is not enabled there is no need to have such methods in `meta` ?

### DIFF
--- a/packages/ember-meta/lib/meta.ts
+++ b/packages/ember-meta/lib/meta.ts
@@ -390,78 +390,6 @@ export class Meta {
     }
   }
 
-  writeBindings(subkey: string, value: any) {
-    if (BINDING_SUPPORT) {
-      assert(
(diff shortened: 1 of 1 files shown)


rwjblue (commented):
Seems good,  but we should ensure the tests in Ember-2-legacy addon still pass. I recall there being a race condition where the addons code executes _after_ Ember is evaluated which would make this change break the addons ability to add binding Support...

krisselden:
@rwjblue addons should get a chance to add to the env, I think module unification if I recall is pulling config too early and caching it, we should ensure the env does get baked until all addons been initialized.

rwjblue:
@krisselden - I completely agree with what you said, but it has no relationship to this PR (or my prior comment).

bekzod:
@rwjblue this is updated with svelting flags

rwjblue (approved):
Awesome, thank you!

PR #16357, opened 2018-03-10
store mixins in set

bekzod:
avoids using `guidFor` for mixins


rwjblue (approved):
LGTM, also added @mmun for review...

rwjblue:
@mmun - when you have a minute...

mmun:
Nice.

PR #16359, opened 2018-03-10
cleanup `watch` methodes

bekzod:
removes some work from https://github.com/emberjs/ember.js/pull/14571 
since then I think root cause might have been fixed

### DIFF
--- a/packages/ember-metal/lib/watch_key.js
+++ b/packages/ember-metal/lib/watch_key.js
@@ -16,8 +16,6 @@ import {
 let handleMandatorySetter;
 
 export function watchKey(obj, keyName, _meta) {
-  if (typeof obj !== 'object' || obj === null) { return; }
-
   let meta = _meta === undefined ? metaFor(obj) : _meta;
   let count = meta.peekWatching(keyName) || 0;
   meta.writeWatching(keyName, count + 1);
@@ -79,9 +77,6 @@ if (MANDATORY_SETTER) {
 }
 
 export function unwatchKey(obj, keyName, _meta) {
-  if (typeof obj !== 'object' || obj === null) {
-    return;
-  }
   let meta = _meta === undefined ? peekMeta(obj) : _meta;
 
   // do nothing of this object has already been destroyed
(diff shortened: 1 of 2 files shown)


packages/ember-metal/lib/watch_key.js
```diff
@@ -16,8 +16,6 @@ import {
 let handleMandatorySetter;
 
 export function watchKey(obj, keyName, _meta) {
-  if (typeof obj !== 'object' || obj === null) { return; }
```
bekzod: since `meta` has assertion for non-object / non-function values https://github.com/emberjs/ember.js/blob/76dc4deb70cd1e5ae7d83374161925af50946a01/packages/ember-metal/lib/meta.js#L554 
so it is better to assert than be silent to uncover unnecessary calls of `watchKey `

packages/ember-metal/lib/watch_key.js
```diff
@@ -79,9 +77,6 @@ if (MANDATORY_SETTER) {
 }
 
 export function unwatchKey(obj, keyName, _meta) {
-  if (typeof obj !== 'object' || obj === null) {
```
bekzod: since `peekMeta` has assertion for non-object / non-function values, it is better to assert than be silent to uncover unnecessary calls of unwatchKey

mmun: Yes, I agree, though we should check that this doesn't cause any crashes in a few real world apps.

bekzod: will do

bekzod: checked in both emberobserver and emberaddons, both working properly and no regression in benchmarks https://cl.ly/441V333w3R3R/results.pdf

packages/ember-metal/lib/watch_key.js
krisselden: does the assert check for null objects?

bekzod: yes https://github.com/emberjs/ember.js/blob/76dc4deb70cd1e5ae7d83374161925af50946a01/packages/ember-metal/lib/meta.js#L554

krisselden (approved):

PR #16362, opened 2018-03-11
correct usage of `meta` property in `tagForProperty`

bekzod:


packages/ember-metal/lib/tags.js
```diff
@@ -15,12 +15,12 @@ function makeTag() {
 
 export function tagForProperty(object, propertyKey, _meta) {
   if (typeof object !== 'object' || object === null) { return CONSTANT_TAG; }
+  let meta = _meta === undefined ? metaFor(object) : _meta;
 
   if (isProxy(object)) {
     return tagFor(object, meta);
```
bekzod: `meta` does not exist yet :)

rwjblue:
FWIW, this was broken in 2a7eda06ca48a5ea982cff8d4d3373ba5f998c97.

@bekzod - Can you create a failing test?

bekzod:
things didn't break since `tagFor` `meta` argument is optional https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/lib/tags.js#L33
and `let meta;` got hoisted to top so it didn't cause any problems I think, 
I was thinking to make all optional `meta` arguments  mandatory so we don't check if meta passed each step like https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/lib/tags.js#L33

rwjblue (approved):

PR #16363, opened 2018-03-11
add weakset polyfill

bekzod:


packages/ember-metal/lib/weak_set.js
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  add(val) {
+    this._weakmap.set(val, true);
+    return this;
+  }
+
+  delete(val) {
+    this._weakmap.delete(val);
+    return this;
```
mmun: This should be `return this._weakmap.delete(val);`. It's a boolean that tells you whether an element was actually deleted or not.

mmun (changes requested):

packages/ember-metal/lib/weak_set.js
bekzod: ok thanks

mmun (approved):

bekzod:
needs restart I suppose

mmun:
@bekzod I think it's a legit error for IE11:

```
not ok 10 IE 11.0 - Global error: Unable to get property 'set' of undefined or null reference at http://127.0.0.1:7774/3242/dist/ember.debug.js, line 23688
```

packages/ember-metal/lib/weak_set.js
```diff
@@ -0,0 +1,22 @@
+const HAS_WEAK_SET = typeof WeakSet === 'function';
+
+class WeakSetPolyFill {
+  construcor() {
```
mmun: Typo. Should be `constructor` :)

bekzod: oops late night typing

PR #16375, opened 2018-03-14
Remove obsolete `test-helper` blueprint

Turbo87:
registerAsyncHelper() is no longer a thing with the new APIs and having this blueprint generate something outdated seems bad 🤔 

/cc @rwjblue


rwjblue:
Thank you!

PR #16376, opened 2018-03-14
[WIP] Export `RouterService` class

Turbo87:
This adds a public export of the `RouterService` class so that people are able to subclass it in their apps if needed.

I suppose this would also need to be added to the rfc176 repo. My suggestion would be `import RouterService from '@ember/routing/service';` or something like that.

/cc @rwjblue

### DIFF
--- a/packages/ember/lib/index.js
+++ b/packages/ember/lib/index.js
@@ -528,6 +528,7 @@ Ember.generateControllerFactory = routing.generateControllerFactory;
 Ember.generateController = routing.generateController;
 Ember.RouterDSL = routing.RouterDSL;
 Ember.Router = routing.Router;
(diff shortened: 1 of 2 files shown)


ef4:
This contradicts the RFC, which tries to avoid exposing any distinction between Router and RouterService in the public API. 

What is the use case you’re trying to satisfy?

Turbo87:
I implemented it based on some conversation on Slack a few days ago. I think the use case was adding computed properties to `service:router` to avoid having to use yet another router service for some derived properties.

ef4:
I don’t think that’s a good idea. Adding your own computer properties to framework-provided classes means we can’t add any future properties without breaking apps.

PR #16384, opened 2018-03-20
Turn ember-template-block-let-helper feature flag on by default

locks:


rwjblue (approved):

tomdale (approved):

tomdale:
We discussed this at today's core team meeting and there were no concerns about go-ing this feature. @locks will take point on updating API docs and guides to use `let`. @rwjblue also mentioned wanting to do a "polyfill" so addons can use this feature in older versions of Ember.

Turbo87:
@tomdale @locks  what are the rough rules on *when* to use the let helper vs. other solutions? seems to me like overuse of the feature could very well lead to more complicated templates, while I guess one of the goals was to simplify?

locks:
@Turbo87 which other solutions?

Turbo87:
> which other solutions?

that's what I'm asking

> @locks will take point on updating API docs and guides to use let.

sounded like *something* should be replaced by the `let` helper and I was wondering what that something would be

locks:
Ah, I believe Tom was thinking of `{{with}}`, but also documenting `{{let}}` in general in the Guides.

PR #16386, opened 2018-03-21
[FEAT] assert transitions on destroyed app instances

runspired:


rwjblue (commented):
Looks good! Can you add a test for the assertion also?

packages/ember-routing/tests/system/router_test.js
```diff
@@ -261,5 +261,20 @@ moduleFor('Ember Router', class extends AbstractTestCase {
...
-});
 
+  ['@test transitionTo should throw an error when called after owner is destroyed'](assert) {
+    let router = createRouter();
+
+    runDestroy(owner);
+
+    router.currentRouteName = 'route-a';
+
+    assert.throws(function() {
```
rwjblue: Since assertions are stripped in production builds (and we run tests against production builds) this will need to be tweaked. The main update should be to replace `assert.throws` with `expectAssertion` (which is a global).

packages/ember-routing/tests/system/router_test.js
```diff
@@ -261,5 +261,20 @@ moduleFor('Ember Router', class extends AbstractTestCase {
...
+      router.transitionTo('route-b');
+    }, "A transition was attempted from 'route-a' to 'route-b' but the application instance has already been destroyed.");
```
rwjblue: This will work fine once you migrate to `expectAssertion`, but I figured I’d point out that this definitely doesn’t do what you expect with `assert.throws`. Specifically, passing `assert.throws` two arguments (a function and a string) does not use the provided string as a matcher for the error message but instead uses the second arg as the label for the assertion. It’s a super super trolly api :sob:

rwjblue (dismissed):

packages/ember-routing/tests/system/router_test.js
runspired: sad face

rwjblue (commented):
```
 Failed assertion: Expected failing Ember.assert: 'A transition was attempted from 'route-a' to 'route-b' but the application instance has already been destroyed.', but got 'The route route-b was not found'., expected: A transition was attempted from 'route-a' to 'route-b' but the application instance has already been destroyed.
```

runspired:
@rwjblue odd. That test was running fine locally. I’ll take a peek.

stefanpenner (approved):

PR #16387, opened 2018-03-21
[FEAT] assert that actions cannot be sent from a destroyed/destroying object

runspired:


packages/ember-glimmer/tests/integration/components/target-action-test.js
```diff
@@ -259,6 +260,34 @@ moduleFor('Components test: sendAction', class extends RenderingTest {
...
+
+        toString() {
+          return 'component:rip-alley';
+        }
+      })
+    });
+
+    this.render('{{rip-alley}}');
+
+    runDestroy(component);
```
rwjblue: I’d vaguely prefer a more natural way to destroy the component, e.g. render it in an `if` and toggle the condition 

What do you think?

runspired: @rwjblue didn't think of that, but I agree sticking closer to real world is best. Will update to that. Do we have a helper or do I need to wrap a set in a run?

stefanpenner (approved):

runspired:
The failure is that browserstack timed out, all else passed. Can we reboot that?

rwjblue:
ya, I kicked that one..

rwjblue (approved):

rwjblue (approved):

Issue #16391, opened 2018-03-22
[Pls send halp!] Cleaning up the test output


chancancode:
♻️ ♻️ ♻️ **Please consider the environment before printing this Github issue.** ♻️ ♻️ ♻️

<img width="1194" alt="screenshot_3_22_18__1_46_pm" src="https://user-images.githubusercontent.com/55829/37774258-a5253a34-2dd7-11e8-807b-2c720507c877.png">

Currently, the test output is very noisy – even for green builds. This makes it very difficult find any _actual_ failures, especially on small screens and on phones.

Let's try to clean it up!

Here are the useless output I have found based on [this test run](https://travis-ci.org/emberjs/ember.js/jobs/356695733). Leave a comment about what you are looking into (🔒) so others won't duplicate the effort.

A lot of these probably have the same root cause (for example, because of how the default `Ember.onError` works) and probably require some refactoring in the test infrastructure.

* Build phase
  * [ ] `'container' is imported by tmp/rollup-cache_path-I6PjJI6s.tmp/run_loop.js, but could not be resolved – treating it as an external dependency`
  * [ ] `'ember-debug' is imported by tmp/rollup-cache_path-l0Z4sygU.tmp/index.js, but could not be resolved – treating it as an external dependency`
  * [x] `'assert' is imported from external module '@glimmer/util' but never used`
  * [x] `'assert' and 'SERIALIZATION_FIRST_NODE_STRING' are imported from external module '@glimmer/util' but never used` (🔒 @James-Byrne)
* Test phase
  * [x] `WARNING: Library "magic" is already registered with Ember.`
  * [ ] `console.trace`
  * [x] `DEPRECATION: Should not throw [deprecation id: test]` (🔒 @Draggha )
  * [x] `(node:6689) Warning: Possible EventEmitter memory leak detected. * listeners added. Use emitter.setMaxListeners() to increase limit` (🔒 @lukecoy )
  * [x] `(node:6689) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: *): Error: Navigation Timeout Exceeded: 900ms exceeded`
  * [x] `Error: Element .does-not-exist not found.` (🔒 @KamiKillertO )
  * [x] `Error: Catch me` (🔒 @KamiKillertO )
  * [x] `Error: the error` (🔒 @KamiKillertO )
  * [x] `Testing paused. Use `resumeTest()` to continue.`
  * [x] The library version banner on app boot  (🔒 @Draggha )
    ```
[...]

PR #16393, opened 2018-03-22

Draggha:
This resolves some Quest Tasks of Issue #16391:

```
Testing paused. Use resumeTest() to continue.
```
--> I switched the `console.info` with a call to `ember-debug.info` and then used the `setDebugFunction`-helper to set it to a NOOP and reset its behavior after each test accordingly.

```
WARNING: Library "magic" is already registered with Ember.
```
--> I overwrote the debug function "warn", so that the test case could register the same lib a second time without throwing an error.

### DIFF
--- a/packages/ember-metal/tests/libraries_test.js
+++ b/packages/ember-metal/tests/libraries_test.js
@@ -6,6 +6,7 @@ import { moduleFor, AbstractTestCase } from 'internal-test-helpers';
 
 let libs, registry;
 let originalWarn = getDebugFunction('warn');
+function noop() {}
 
 moduleFor('Libraries registry', class extends AbstractTestCase {
   beforeEach() {
@@ -33,6 +34,8 @@ moduleFor('Libraries registry', class extends AbstractTestCase {
   ['@test only the first registration of a library is stored'](assert) {
     assert.expect(3);
 
+    // overwrite warn to supress the double registration warning (see https://github.com/emberjs/ember.js/issues/16391)
+    setDebugFunction('warn', noop);
     libs.register('magic', 1.23);
     libs.register('magic', 2.23);
 
--- a/packages/ember-testing/lib/helpers/pause_test.js
+++ b/packages/ember-testing/lib/helpers/pause_test.js
@@ -2,7 +2,7 @@
 @module ember
 */
 import { RSVP } from 'ember-runtime';
-import { assert } from 'ember-debug';
+import { assert, info } from 'ember-debug';
 
 let resume;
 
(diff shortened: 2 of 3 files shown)


packages/ember-testing/lib/helpers/pause_test.js
```diff
@@ -57,9 +57,6 @@ export function resumeTest() {
  @public
 */
 export function pauseTest() {
-  // eslint-disable-next-line no-console
-  console.info('Testing paused. Use `resumeTest()` to continue.');
```
rwjblue: We don't want to remove this, it is intentionally logging to the console to let developers know how they can resume. Please revert this removal...

Instead of removing this, we should override `console.info` in the specific tests (and confirm it is called properly).

rwjblue (changes requested):
The fix for the library registration warning looks great, but the other one needs some tweaks before landing.

packages/ember-testing/lib/helpers/pause_test.js
Draggha: Thanks Robert! I figured that outright deleting this part could have been a bit too much.
Overwriting `console.info` should be possible, but what I fear is that the test code which relies on functions such as `pauseTest` will rely more and more on such workarounds. Every occurrence of console.info, console.warn and the like would need to be patched (now and in future tests).

chancancode: It might be a good idea to indirect these through a `@ember/debug` function (which makes it easier to swap out), what do you think @rwjblue?

packages/ember-metal/tests/libraries_test.js
```diff
@@ -33,6 +34,8 @@ moduleFor('Libraries registry', class extends AbstractTestCase {
   ['@test only the first registration of a library is stored'](assert) {
     assert.expect(3);
 
+    // overwrite warn to supress the double registration warning (see https://github.com/emberjs/ember.js/issues/16391)
+    setDebugFunction('warn', noop);
```
chancancode: @rwjblue do we need to set it back after the test, or is that automatically handled by the test harness?

rwjblue: It’s already resetting in the afterEach of this module

Draggha: Yup, that's why I didn't reset it afterwards. :)

Draggha:
Thanks to both of you for your patience and support. I updated the first task.

```
Testing paused. Use resumeTest() to continue.
```

I also updated the opening post to reflect the current state of this PR.

How would you like to proceed? Would you like to create a branch where this PR would land or will this land inm master? And would you like me open more PRs for other tasks from the issue or should I update this branch? I didn't find anything about your git workflow in the CONTRIBUTING guide. Maybe there is another document somewhere?

rwjblue (approved):
Thank you!!

PR #16396, opened 2018-03-23

KamiKillertO:
This resolves some Quest Tasks of Issue #16391:

- `Error: Catch me`
- `Error: the error.` 
- `Error: Element .does-not-exist not found.`

I've just remove a call to `console.error`, I don't know if he was here on purpose 🤷‍♂️.

### DIFF
--- a/packages/ember-testing/lib/test/adapter.js
+++ b/packages/ember-testing/lib/test/adapter.js
@@ -28,4 +28,6 @@ export function asyncEnd() {
 
 function adapterDispatch(error) {
   adapter.exception(error);
+
+  console.error(error.stack);// eslint-disable-line no-console
 }
--- a/packages/ember-testing/lib/test/adapter.js
+++ b/packages/ember-testing/lib/test/adapter.js
@@ -28,6 +28,4 @@ export function asyncEnd() {
 
 function adapterDispatch(error) {
   adapter.exception(error);
-  
-  console.error(error.stack);// eslint-disable-line no-console 
 }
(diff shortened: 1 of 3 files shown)


packages/ember-testing/lib/test/adapter.js
```diff
@@ -28,6 +28,4 @@ export function asyncEnd() {
 
 function adapterDispatch(error) {
   adapter.exception(error);
-  
-  console.error(error.stack);// eslint-disable-line no-console
```
rwjblue: Yep, we need to bring this back (and explicitly mock in tests where we expect an error).

Specifically, this console.error is the _only_ thing that notifies the console of failures (including the stack trace) when unexpected exceptions or rejections happen in tests...

rwjblue (changes requested):

packages/ember-testing/lib/test/adapter.js
KamiKillertO: Something like this?

```javascript
function adapterDispatch(error) {
   adapter.exception(error);
   
   adapter.outputError(error.stack); 
```

`outputError` is by default `console.error` and can be overridden with a noop like this.

```
Test.adapter = QUnitAdapter.create({
   outputError () {}
});
```

rwjblue: I was more thinking to just override `console.error` in the specific tests, then reset it in the `afterEach`

rwjblue (approved):

rwjblue:
Thank you!!

PR #16401, opened 2018-03-23

Draggha:
This PR solves the task: `the library version banner on app boot` from #16391

I have overwritten the debug function in certain test suites with a NOOP to supress the console output. But (!) I didn't do that for all tests, only for those that will be traversed in a full test run (and in that order). If a certain testsuite would be `skip`ped, the following test suite would still print the banner.
I would update the PR if this behavior was not intended. :)


rwjblue (approved):

rwjblue:
Thank you!

PR #16402, opened 2018-03-23

Draggha:
This PR solves the task: `DEPRECATION: Should not throw [deprecation id: test]` from #16391

I have overwritten the console.warn function with a NOOP on two tests to supress the console output and reverted that change after each test. I would have preferred to use ember-debug.warn in the deprecate function, but I'm a bit unsure if the es2015 modules would be compiled to other formats (cjs, umd, etc.) and maybe cause circular dependencies, since the deprecate function (which uses console.warn) is being imported into the index file which sets up the `setDebugFunction` and `getDebugFunction` helpers.


rwjblue (approved):

rwjblue:
Thank you!!

PR #16404, opened 2018-03-23
[BUGFIX beta] Add exception for `didRemoveListener` so evented proxy objects can

rondale-sc:
function

When `off` is called on an Evented proxyish object it checks to see if
didRemoveListener is a function (so that it can later call it).  This
check triggers the assertion for proxy-ish objects to use `get` rather
than installed es5-getter so we add a special case to allow it.

Prior to this change when you called `off` on the proxyish object it
would trigger an assertion telling you to call
`eventedProxyObj.get('didRemoveListener') rather than
`eventedProxyObj.didRemoveListener`.

A few notes about this test:

- the unknownProperty is how we let ember know this is a proxy
- the unknownProperty function must return a value other than undefined
- Evented Mixin must be mixed in

### DIFF
--- a/packages/ember-runtime/lib/system/core_object.js
+++ b/packages/ember-runtime/lib/system/core_object.js
@@ -92,6 +92,7 @@ function makeCtor() {
               property === 'willWatchProperty' ||
               property === 'didUnwatchProperty' ||
               property === 'didAddListener' ||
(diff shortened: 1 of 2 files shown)


rwjblue (approved):
Thanks @rondale-sc!

PR #16408, opened 2018-03-24
use `Map` for storing keys in `each_proxy`

bekzod:


packages/ember-metal/lib/each_proxy.js
```diff
@@ -44,23 +44,24 @@ class EachProxy {
...
+      });
     }
   }
 
   arrayDidChange(content, idx, removedCnt, addedCnt) {
     let keys = this._keys;
     let lim = addedCnt > 0 ? idx + addedCnt : -1;
     let meta = peekMeta(this);
-    for (let key in keys) {
+    keys.forEach((val, key)=> {
```
krisselden: I think changing to forEach will adversely affect the performance over Object.create(null) and key in.  This will have to be measured. Use emberaddons because it stresses the object model a bit more than emberobserver.

krisselden (changes requested):

packages/ember-metal/lib/each_proxy.js
bekzod: cool cool, will measure it np

bekzod: https://cl.ly/2x2n3915291C/results.pdf @krisselden all good I think

bekzod: use-map-for-keys - 0080c9197b027987a185642b60c6746b55a4fbda
master - d49f3edea0229d17c2ecd610d6d3b386a6a4310a

krisselden: seems ok, question, can you show some more of your benchmark work? what app did you test and did it use @each significantly?

locks:
@bekzod is this work still valid?

bekzod:
this is no longer relevant

PR #16409, opened 2018-03-24
Refactor each tests

chancancode:
* Resolves a TODO, so that tests using multiple `{{#each}}` helpers are testing against all the different array-like objects.
* Add tests for native Set.

### DIFF
--- a/packages/ember-glimmer/tests/integration/syntax/each-test.js
+++ b/packages/ember-glimmer/tests/integration/syntax/each-test.js
@@ -12,9 +12,10 @@ import {
   ArrayTestCases
 } from '../../utils/shared-conditional-tests';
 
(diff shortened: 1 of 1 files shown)


chancancode:
Tests using `Set` were failing on IE11 so I flagged them with `HAS_NATIVE_SYMBOL`. However, they _should_ work as IE11 has basic `Set` with `forEach` – need to investigate more some time.

chancancode:
This _was_ working, so we might have introduced a regression in the recent refactor: https://ember-twiddle.com/19a9147c236f00742aaa6898ad7de048

chancancode:
Figured it out – IE11 does not support passing arguments to the Set constructor

PR #16410, opened 2018-03-24
use `Map` as default store and avoid using `UNDEFINED`

bekzod:
using `Map` as default store will simplify things and avoids using `UNDEFINED` constant


bekzod:
needs restart

bekzod:
already done part of https://github.com/emberjs/ember.js/pull/16544

PR #16417, opened 2018-03-25
fix setting `undefined` issue

### LINKED ISSUE #14270: property_set.js won't modify object if value and currentValue is undefined
In newer ember versions, `Ember.set(object, 'field', undefined)` won't set the `field` key on an object, because the value is `undefined`.
It exits earlier [here](https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/lib/property_set.js#L62-L63) which causes no object modification.

I think this is a regression (introduced in `v2.7.0-beta.1`), because [previously](https://github.com/emberjs/ember.js/commit/9471527454b30b9a892d6d40f21735365b5a2897#diff-2c6910b295639eb6f0ea3f0f2c9e3520L72) it wouldn't enter that condition, if the value was `undefined`.

Using a 2.7.0 ember app:

On discourse:

A possible way to fix this would to reintroduce the `value !== undefined` check

bekzod:
fixes #14270


packages/ember-metal/tests/accessors/set_test.js
```diff
@@ -21,16 +21,14 @@ moduleFor('set', class extends AbstractTestCase {
       undefinedValue: undefined
     };
 
-    let newObj = {
-      undefinedValue: 'emberjs'
```
bekzod: this was hiding issue before :)

packages/ember-metal/lib/property_set.js
```diff
@@ -77,7 +77,7 @@ export function set(obj, keyName, value, tolerant) {
   } else if (currentValue === undefined && 'object' === typeof obj && !(keyName in obj) &&
     typeof obj.setUnknownProperty === 'function') { /* unknown property */
     obj.setUnknownProperty(keyName, value);
-  } else if (currentValue === value) { /* no change */
+  } else if (currentValue === value && keyName in obj) { /* no change */
```
bekzod: I feel like removing this line too, thoughts ?

rwjblue: Agree...

rwjblue (approved):

krisselden:
I the argument that this is a BUGFIX, not a regression, since it was regressed in 2.7 and stayed regressed until 3.1 shouldn't it be labelled a BUGFIX so we backport it when it is stable?

PR #16420, opened 2018-03-26
Refactor {{each-in}} tests

cibernox:
Replaces #15579

### DIFF
--- a/packages/ember-glimmer/tests/integration/syntax/each-in-test.js
+++ b/packages/ember-glimmer/tests/integration/syntax/each-in-test.js
@@ -1,36 +1,43 @@
-import { set, get } from 'ember-metal';
+import { get, set } from 'ember-metal';
+import { moduleFor, RenderingTest } from '../../utils/test-case';
+import { Object as EmberObject, ObjectProxy } from 'ember-runtime';
 import { strip } from '../../utils/abstract-test-case';
+import { HAS_NATIVE_SYMBOL } from 'ember-utils';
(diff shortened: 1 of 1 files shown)


packages/ember-glimmer/tests/integration/syntax/each-in-test.js
```diff
@@ -57,222 +62,231 @@ applyMixins(BasicEachInTest,
...
+
+    this.assertText('Smartphones: 8203JavaScript Frameworks: Infinity');
 
     this.assertStableRerender();
 
-    this.runTask(() => {
-      set(this.context, 'categories.Smartphones.reports.unitsSold', 100);
-      set(this.context, 'categories.Tweets', { reports: { unitsSold: 443115 } });
-    });
+    if (this.allowsSetProp) { // Not al backing data structures allow kvo tracking. Maps and Iterables don't
```
cibernox: @chancancode This is the idea I came up with to allow to run the same tests for every backing data structure but skip one specific part on data structures that do not allow KVO tracking: Maps and Iterables.

I considered (and I actually started with that approach) to define a `setProp` implementation that essentially replaced the Map/Iterable with a modified one, in an immutable fashino, but I had the impression that it was hiding the fact from the reader that the feature wasn't actually supported for some objects.

cibernox:
@chancancode Ready

cibernox:
This is finally green. Dat prettier refactor...

rwjblue:
Sorry @cibernox :(

PR #16423, opened 2018-03-26
Fix a few issues with ember-template-compiler

rwjblue:
* Import `@glimmer/compiler` directly in ember-template-compiler.
* Use explicit `/index` for importing from folder.


rwjblue:
Failures seem legit. Will need to dig in and figure out what’s going on...

PR #16431, opened 2018-03-28
Make run loop functions named exports.

rwjblue:


mmun:
Maybe I'm missing something, but don't we need to continuing setting `run.schedule = schedule` for back compat?

rwjblue:
Yep, definitely missed that one. See any others I missed on the global side?

rwjblue:
OK, just went through https://emberjs.com/api/ember/release/modules/@ember%2Frunloop and https://github.com/ember-cli/ember-rfc176-data/#emberrunloop to make sure all the globals that I had originally missed are included...

mmun (approved):

PR #16434, opened 2018-03-28
Cleanup / make working with TS in repo easier.

rwjblue:
* Remove "shifting" behavior to remove `packages/*/lib/` from the build path.
* Swap to using `.map` files (instead of inline sourcemaps) to prevent issues with broccoly-uglify-js
* Change target for `tsconfig.json` to `ES2017`
* Fix typings for `packages/ember-metal/lib/tracked.ts` (which still needs more work TBH)
* Move from `files` listing in`tsconfig.json` to "normal" `include` globbing (which will allow much
  easier path to landing TS in other packages).


packages/ember-metal/lib/tracked.ts
```diff
@@ -208,8 +199,8 @@ function descriptorForDataProperty(key, descriptor) {
       return this[shadowKey];
     },
 
-    set(newValue) {
-      tagFor(this).inner.dirty();
+    set(newValue: any) {
+      tagFor(this).inner!['dirty']();
```
rwjblue: This specifically "feels bad", but I'm not sure what the right thing to do here is.  The changes here squelch the type errors (which fail CI), but don't _really_ ensure that this works properly.

At the moment, this file is complete unused (the actual implementation lives in `./tracked.js`), but I'd love to followup with a more correct fix for this..

PR #16436, opened 2018-03-28
Refactor CoreObject to leverage native JS semantics.

rwjblue:
* Make `.extend()` _actually_ use `class` for proper inheritance
* Move the concepts of `NAME_KEY` to a weakmap
* "prettier" some files...
* prevent meta assertions from eagerly "to-string"ing an object (when the assertion is not going to run anyways)

### DIFF
--- a/packages/ember-metal/lib/namespace_search.js
+++ b/packages/ember-metal/lib/namespace_search.js
@@ -1,0 +1,200 @@
+import { getName, setName } from 'ember-utils';
+import { context } from 'ember-environment';
+
+// TODO, this only depends on context, otherwise it could be in utils
+// move into its own package
+// it is needed by Mixin for classToString
+// maybe move it into environment
+
+const hasOwnProperty = Object.prototype.hasOwnProperty;
+
+let searchDisabled = false;
+
+const flags = {
+  _set: 0,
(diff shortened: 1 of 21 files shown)


rwjblue:
@pzuraq - This is a slightly different approach than your PR, but has basically the same goal. When you are back in the saddle, can you rebase your tests on top of this branch and see how we do?

pzuraq:
👍this looks good based on a quick reading on my phone, will rebase the tests as soon as I’m back!

packages/ember-metal/lib/namespace_search.js
```diff
@@ -0,0 +1,200 @@
...
+      processNamespace(namespaces[i]);
+    }
+
+    unprocessedMixins = false;
+  }
+}
+
+export function classToString() {
+  let name = getName(this);
+  if (name !== void 0) {
```
bekzod: even if it was `undefined` babel would replace it to `void 0`, no ?

rwjblue: Yes, I believe so

rwjblue:
https://github.com/emberjs/ember.js/pull/16437 enables the testing infrastructure to ensure that we stay compliant...

PR #16439, opened 2018-03-29
Update eslint configuration.

rwjblue:
* Move to single eslint config file (leveraging overrides)
* Add eslint-plugin-node (to help ensure that we continue to be compatible with supported engines)
* Make configuration more strict (e.g. we previously allowed _all_ files to globally use qunit, mocha, etc)


.eslintrc.js
```diff
@@ -1 +1,129 @@
...
+    'eslint:recommended',
+  ],
+  plugins: [
+    "ember-internal"
+  ],
+
+  rules: {
+    'semi': 'error',
+    'no-unused-vars': 'error',
+    'no-useless-escape': 'off', // TODO: bring this back
```
rwjblue: FWIW, the eslint version upgrade (from 3.x to 4.x) brings some significant changes to the `no-useless-escape` rule which flags ~ 350 problems.

I disabled the rule for now, and will create another questlike issue to fix it up and remove this line.

packages/ember-metal/lib/map.js
```diff
@@ -271,7 +271,7 @@ class Map {
     let guid = guidFor(key);
 
     // ensure we don't store -0
-    let k = key === -0 ? 0 : key;
+    let k = key === -0 ? 0 : key; // eslint-disable-line no-compare-neg-zero
```
rwjblue: There are explicit tests around negative zero, which force me to disable the rule for this line. In the long run we will be deprecating and removing all of this code.

PR #16441, opened 2018-03-29
[CLEANUP] Converting ember-runtime tests to new format

thoov:
Apart of #15988 

Once this PR lands we can get rid of `testBoth` and `testWithDefault` as they are no longer used.


rwjblue (approved):

rwjblue:
Thank you!

PR #16445, opened 2018-03-30

allthesignals:
This PR solves the task: Error while processing route: * in #16391 

_*I'm submitting this mainly for feedback before I apply a fix to all the other tests (the noisy output appears in many places)*_

This changes the test the trigger an error from within the promise instance of the outer scope. It seems like throwing an error in a mocked `Route` hook will cause a rejected promise, but the error message throws in the console (since it's not being caught). If there is a better approach, let me know! Thanks.


locks:
Hi @allthesignals, there has been a lot of moving around of files.
Could you open a new PR if the issue is still present? Thank you for your time!

PR #16450, opened 2018-04-01
[CLEANUP] Removing testBoth and testWithDefault

thoov:
These functions have been refactored away from and are no longer in use.

Apart of #15988


locks (approved):

rwjblue:
👏

PR #16458, opened 2018-04-03
[CLEANUP] Moving off QUnit.config.current.assert

thoov:
Moving off of the QUnit global by refactoring away from QUnit.config.current.assert. There are still a couple more cases that need to be refactored away but those will be done in later PRs. 

This move as enabled by: https://github.com/emberjs/ember.js/pull/16441/files#diff-a1c01fcbc707fbc373ccbf31a27ed388

Apart of #15988


rwjblue (approved):

rwjblue:
Thank you!

PR #16465, opened 2018-04-04
Update ember-cli and fix failing tests

GavinJoyce:
In order to continue working on [MU Blueprints](https://github.com/ember-cli/ember-cli/issues/7530), we need to update the ember-cli version that we're using in ember.js.

The output of `ember new` has since been changed to emit new style tests. These changes result in a number of failing ember.js tests. For example, the `destroyApp` helper no longer exists on disk so the default generated test format output has changes for a bunch of `ember generate` blueprints.

I've updated the existing tests to expect the new default output and I've pinned `ember-cli` at [this SHA](https://github.com/ember-cli/ember-cli/commit/8f5a13a02d8ea5a04683565340611161c63bef89).

I've also [removed ember-cli config caching](https://github.com/ember-cli/ember-cli/pull/7739) which was also causing failing tests ember-source

### DIFF
--- a/blueprints/initializer-test/qunit-files/tests/unit/initializers/__name__-test.js
+++ b/blueprints/initializer-test/qunit-files/tests/unit/initializers/__name__-test.js
@@ -3,7 +3,7 @@ import { run } from '@ember/runloop';
 
 import { initialize } from '<%= dasherizedModulePrefix %>/initializers/<%= dasherizedModuleName %>';
 import { module, test } from 'qunit';
-<% if (destroyAppExists) { %>import destroyApp from '../../helpers/destroy-app';<% } else { %>import { run } from '@ember/runloop'; <% } %>
+<% if (destroyAppExists) { %>import destroyApp from '../../helpers/destroy-app';<% } %>
 
 module('<%= friendlyTestName %>', {
   beforeEach() {
--- a/blueprints/initializer-test/mocha-files/tests/unit/initializers/__name__-test.js
+++ b/blueprints/initializer-test/mocha-files/tests/unit/initializers/__name__-test.js
@@ -3,7 +3,7 @@ import { describe, it, beforeEach, afterEach } from 'mocha';
 import { run } from '@ember/runloop';
 import Application from '@ember/application';
 import { initialize } from '<%= dasherizedModulePrefix %>/initializers/<%= dasherizedModuleName %>';
-import destroyApp from '../../helpers/destroy-app';
+<% if (destroyAppExists) { %>import destroyApp from '../../helpers/destroy-app';<% } %>
 
 describe('<%= friendlyTestName %>', function() {
   let application;
@@ -16,7 +16,7 @@ describe('<%= friendlyTestName %>', function() {
   });
 
   afterEach(function() {
-    destroyApp(application);
+    <% if (destroyAppExists) { %>destroyApp(application);<% } else { %>run(application, 'destroy');<% } %>
   });
 
   // Replace this with your real tests.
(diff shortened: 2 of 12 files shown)


rwjblue:
> Are we happy to continue testing the (updated) default generator output as before and live with similar to the existing gaps in our coverage?

Yes.

node-tests/blueprints/component-test-test.js
```diff
@@ -49,7 +49,8 @@ describe('Blueprint: component-test', function() {
       });
     });
 
-    describe('with usePods=true', function() {
+    //TODO: GJ: figure out what changed here
+    describe.skip('with usePods=true', function() {
```
rwjblue: Maybe one of these:

https://github.com/ember-cli/ember-cli/pull/7164
https://github.com/ember-cli/ember-cli/pull/7491

GavinJoyce: thanks, a git bisect points to https://github.com/ember-cli/ember-cli/commit/5d5801a086ae5789d771a7e4ae158392e9fdf6b0

rwjblue: My guess is due to caching 🤔

GavinJoyce: Yes, removing the caching fixes the issue

https://github.com/ember-cli/ember-cli/blob/252e1ccd1a99222397c8c9b293eb5e5f376070e1/lib/utilities/get-config.js#L18

GavinJoyce: oh, there are some details on this in the previous PR: https://github.com/ember-cli/ember-cli/pull/7491#issuecomment-356756278

GavinJoyce: and here: https://github.com/ember-cli/ember-cli/pull/7491#discussion_r162199039

I'll sync up with @patocallaghan in the morning and we'll try to remedy this

rwjblue: I think in general that the caching is fine, we just need an escape hatch to bypass and reset...

tldr; cache invalidation is hard

patocallaghan: @GavinJoyce I wasn't aware that the tests had been moved elsewhere, I just thought they'd been deleted.

I had a look there and I remember it being quite tricky to make those particular tests work with `get-config`. The config had already been cached at the module level and the `replaceFile` logic in the test happened too late. It wasn't easy to "reach in". In other tests that tested this sort of thing we used multiple fixture directories so it worked fine.

We're not actually using this feature ourselves yet so if it unblocks your MU work I'm fine for the changes to be reverted for now.

@rwjblue any thoughts? another option is to remove the cache but I believe the instrumentation lib will call `getConfig` a lot which will result in lots of `Yam` objects being created.

GavinJoyce: Two options:

 * [remove config caching](https://github.com/ember-cli/ember-cli/pull/7739)
 * [revert "Allow instrumentation to be turned on using .ember-cl"](https://github.com/ember-cli/ember-cli/pull/7738)

GavinJoyce:
@rwjblue this is now 🍏 and ready for review

blueprints/initializer-test/qunit-files/tests/unit/initializers/__name__-test.js
```diff
@@ -3,7 +3,7 @@ import { run } from '@ember/runloop';
 
 import { initialize } from '<%= dasherizedModulePrefix %>/initializers/<%= dasherizedModuleName %>';
 import { module, test } from 'qunit';
-<% if (destroyAppExists) { %>import destroyApp from '../../helpers/destroy-app';<% } else { %>import { run } from '@ember/runloop'; <% } %>
```
GavinJoyce: `run` is already imported above

rwjblue (approved):

rwjblue:
Thank you @GavinJoyce!

PR #16466, opened 2018-04-04
[BUGFIX beta] revert jquery-less pendingRequest event handling

rondale-sc:
In general,Ember's testing system leverages `ajaxSend` and `ajaxComplete` events to know when `jQuery.ajax` requests are pending (and it should therefore "pause settledness" of the testing helpers like `click` / `andThen` / etc).

In the Ember 3.1 cycle Ember changed from using `jQuery.on` to `document.addEventListener` to subscribe to these events. Unfortunately, `jQuery.ajax` **does not** trigger listeners added in this way (we would have had to do `document.onajaxSend = function() { /* stuff here */ }`).  The result of this change is that "normal" acceptance testing helpers will _no longer_ be able to detect and wait for pending `jQuery.ajax`.

So, for example, you might expect that given the following

```js 
await click('#some-selector');

assert.equal( 
  $('#something-that-should-exist-after-an-ajax-call').text(), 
  'hi'
); 
```

the click helper would wait until `jQuery.ajax` calls were settled before the assertion, but prior to the changes in this commit the promise returned by `click` would **not** wait until the `jQuery.ajax` request was completed and therefore the assertion would fail.

---

This commit reverts the changes to use `document.addEventListener` back to using `jQuery.on` directly when `jQuery` is available.

Many bothans died to bring us this information. :sob: :cry: 😿 

Paired with @rwjblue & @iezer to resolve this issue.

### DIFF
--- a/packages/ember-testing/tests/helpers_test.js
+++ b/packages/ember-testing/tests/helpers_test.js
@@ -4,7 +4,7 @@ import { Route } from 'ember-routing';
 import { Controller, RSVP } from 'ember-runtime';
 import { later } from 'ember-metal';
 import { Component } from 'ember-glimmer';
-import { jQueryDisabled } from 'ember-views';
+import { jQueryDisabled, jQuery } from 'ember-views';
 
 import Test from '../test';
 import setupForTesting from '../setup_for_testing';
@@ -24,12 +24,6 @@ var noop = function() {};
 
 function registerHelper() {
   Test.registerHelper('LeakyMcLeakLeak', () => {});
-}
-
-function customEvent(name, xhr) {
-  let event = document.createEvent('CustomEvent');
-  event.initCustomEvent(name, true, true, { xhr });
-  document.dispatchEvent(event);
 }
 
 function assertHelpers(assert, application, helperContainer, expected) {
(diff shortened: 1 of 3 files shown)


packages/ember-testing/tests/helpers_test.js
```diff
@@ -1074,7 +1074,7 @@ if (!jQueryDisabled) {
         customEvent('ajaxComplete', xhr);
         assert.equal(
           pendingRequests(),
-          1,
```
rondale-sc: If you see all the customEvents and include the reset for `setupForTesting` we should end with 0 pendingRequests **not** 1.

rondale-sc: I am not sure why this test passed on a few of the travis scenarios.  I pushed this fix, we'll see where travis lands.

Dhaulagiri:
This sounds eerily similar to most of the failures we have been seeing in 3.1 🙏🏻 Will try this out and hopefully it fixes most of our issues!

rwjblue:
@Dhaulagiri - It is cherry-picked into the beta branch, should be able to update to:

```json
"ember-source": "https://s3.amazonaws.com/builds.emberjs.com/beta/shas/7169066a8fe5278ff9ff0f08a7d2ded09008464b.tgz",
```

In your package.json to test, please let us know!

Dhaulagiri:
🙏  this fixed all the unexplainable issues I was having, thank you! The remaining failures are things I've already opened issues for (but haven't gotten any reponse).

PR #16468, opened 2018-04-05

KamiKillertO:
This PR remove the error message `More context objects were passed than there are dynamic segments for route: stink-bomb`. (Quest #16391)

I've just override `console.error` with a noop function.


rwjblue:
I just confirmed that the modules teardown properly resets `console.error`, good job!

PR #16469, opened 2018-04-05
Begin building the Ember global dynamically.

rwjblue:
This is another (in a long line) of interim steps to make the Ember global auto-generated. Ultimately, we want to avoid encoding the data structure at runtime, and instead just emit the correct `Object.defineProperty` calls.

This interim step however, gives us a nice little win: we now lazily require the modules used for the `Ember` global. This means we finally stop eagerly evaluating the _entire_ `ember.prod.js` asset...

### DIFF
--- a/packages/ember-views/lib/system/jquery.js
+++ b/packages/ember-views/lib/system/jquery.js
@@ -1,5 +1,24 @@
 import { context, environment } from 'ember-environment';
 
+/**
+ @module jquery
+ @public
+ */
+
+/**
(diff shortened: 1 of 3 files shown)


rwjblue:
What do you think about this @krisselden / @stefanpenner? Worth trying to move forward here?

locks:
Is this still worth it? @rwjblue @pzuraq

locks:
Closing as no longer relevant.

PR #16470, opened 2018-04-06

lukecoy:
This PR removes the test phase warning message: `(node:6689) Warning: Possible EventEmitter memory leak detected. * listeners added. Use emitter.setMaxListeners() to increase limit` (See quest [#16391](https://github.com/emberjs/ember.js/issues/16391))

The test phase would [print the warning](https://travis-ci.org/emberjs/ember.js/jobs/356695733#L1183) after launching N headless chrome instances. After doing some reading on this [puppeteer issue](https://github.com/GoogleChrome/puppeteer/issues/594#issuecomment-325919885), it seems that the listeners spawn because we launch N chrome instances sequentially without closing them.

To mitigate this, one solution is to just create 1 chrome instance and open N new pages inside it, rather than creating several instances with 1 page each.

### DIFF
--- a/bin/run-tests.js
+++ b/bin/run-tests.js
@@ -14,6 +14,7 @@ var path = require('path');
 var finalhandler = require('finalhandler');
 var http = require('http');
 var serveStatic = require('serve-static');
+var puppeteer = require('puppeteer');
 
 // Serve up public/ftp folder.
 var serve = serveStatic('./dist/', { index: ['index.html', 'index.htm'] });
@@ -28,6 +29,9 @@ var PORT = 13141;
 // Listen.
 server.listen(PORT);
 
+// Cache the Chrome browser instance when launched for new pages.
+var browserPromise;
+
 function run(queryString) {
   return new RSVP.Promise(function(resolve, reject) {
     var url = 'http://localhost:' + PORT + '/tests/?' + queryString;
@@ -40,9 +44,11 @@ function runInBrowser(url, retries, resolve, reject) {
 
   console.log('Running Chrome headless: ' + url);
 
-  var puppeteer = require('puppeteer');
+  if (!browserPromise) {
+    browserPromise = puppeteer.launch();
+  }
 
-  puppeteer.launch().then(function(browser) {
+  browserPromise.then(function(browser) {
     browser.newPage().then(function(page) {
       /* globals window */
       var crashed;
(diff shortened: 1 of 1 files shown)


rwjblue (commented):
Seems great! 

Can you also ensure we call `page.close()` when we are done with a given test run? I think that we would need to add some `page.waitForFunction` ([docs here](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagewaitforfunctionpagefunction-options-args)) or something after the `page.goto`, then once QUnit.done fires we can call `page.close()`.

What do you think?

lukecoy:
Thanks @rwjblue :) Yep, will add `page.close()`.

In terms of the best approach, I'd think we can just add `page.close()` to the top of the existing [`page.exposeFunction('callPhantom', ...)`](https://github.com/emberjs/ember.js/pull/16470/files#diff-437a149f1b0f7d248f37db51346c8ff2R155) inline function, because that function gets called in the `QUnit.done` callback and allows us to close the page from node at the right time. I'm unsure that we need `page.waitForFunction` / how we'd use it.

In addition to that, to ensure that the `window.callPhantom` function is always called from the browser & to [avoid this if check](https://github.com/emberjs/ember.js/pull/16470/files#diff-437a149f1b0f7d248f37db51346c8ff2R145), I think we can utilize promise chaining so everything is called in the correct order, like this:
```
page.exposeFunction('callPhantom', ...)
  .then( return page.evaluateOnNewDocument(addLogging))
  .then( page.goto(url, { timeout: 900 }) )
```

Tests seem to work locally with these changes, any thoughts?

rwjblue:
> In terms of the best approach, I'd think we can just add page.close() to the top of the existing page.exposeFunction('callPhantom', ...) inline function, because that function gets called in the QUnit.done callback and allows us to close the page from node at the right time. I'm unsure that we need page.waitForFunction / how we'd use it.

Seems great!

> In addition to that, to ensure that the window.callPhantom function is always called from the browser & to avoid this if check, I think we can utilize promise chaining so everything is called in the correct order, like this (pseudo-code):

Perfect, thanks for working on this!

lukecoy:
Okay @rwjblue - should be good to go now. Thanks for the review, let me know if there's anything else!

Had to fix a bunch of linting issues after adding the promise chains, so the diff looks deceptively large now :)

bin/run-tests.js
```diff
@@ -136,52 +142,55 @@ function runInBrowser(url, retries, resolve, reject) {
...
+            });
           });
         });
       };
 
-      page.exposeFunction('callPhantom', function(message) {
-        if (message && message.name === 'QUnit.done') {
-          result = message.data;
-          var failed = !result || !result.total || result.failed;
+      page
```
rwjblue: This should be `return page` right? (I realize this was messed up before)

bin/run-tests.js
```diff
@@ -136,52 +142,55 @@ function runInBrowser(url, retries, resolve, reject) {
...
-
-      page.evaluateOnNewDocument(addLogging);
-
-      page.goto(url, { timeout: 900 });
+        })
+        .then(function() {
+          return page.evaluateOnNewDocument(addLogging);
+        })
+        .then(function() {
+          page.goto(url, { timeout: 900 });
```
rwjblue: `return page.goto(...)`

rwjblue (commented):
Awesome, looking much better thank you!

rwjblue (approved):

PR #16440, opened 2018-03-29
Add prettier to linting configuration.

rwjblue:
* Add prettier via eslint config
* Add `yarn lint:fix` command to easily "autofix" linting errors

It will be _much_ easier to review this PR commit by commit as the bulk of the changes are purely mechanical (`yarn eslint --fix .`). The first commit contains all of the prettier configuration / setup:

https://github.com/emberjs/ember.js/pull/16440/commits/a49b00743519603c03843137a2522473a6969281

PR #16473, opened 2018-04-07
[BUGFIX beta] Fix missing limb

brzpegasus:
The shrug emoticon went from having a broken arm (improperly escaped) to no arm during the [prettifying](https://github.com/emberjs/ember.js/pull/16440/files#diff-437a149f1b0f7d248f37db51346c8ff2L156) process.


rwjblue (approved):
Thank you!

Issue #16153, opened 2018-01-20
Class Documentation for Transition disappeared between Ember 2.16 and 2.17


toddjordan:
If you go to https://emberjs.com/api/ember/2.16/classes/Transition (select "show private" at the bottom of the nav menu.) You'll see the last time Transition was seen alive.

A little CSI forensic work is required to figure out what happened here. Could've been some refactoring that moved it way from a place where the yuidocs are being processed.

efx:
@toddjordan could you elaborate on 

> You'll see the last time Transition was seen alive.

I followed the steps but do not notice anything odd. 
<img width="860" alt="screen shot 2018-04-02 at 10 21 20" src="https://user-images.githubusercontent.com/472721/38199699-b02439f4-365f-11e8-93ee-c72d7f431454.png">

toddjordan:
@efx its because the steps above show where you can find it.

Try looking for it in 2.17, 2.18, or 3.0

efx:
ah, thanks; I see the issue now. Selecting 2.18 once you have navigated to that page brings up a 404.

efx:
The `Transition` docs now live here: https://github.com/tildeio/router.js/blob/master/lib/router/transition.js#L7-L21
Changes to https://github.com/tildeio/router.js's package.json probably are out of sync with ember's yuidoc.json (perhaps https://github.com/tildeio/router.js/blob/master/package.json#L26-L29 ?). I tried naively here:
https://github.com/efx/ember.js/commit/7a884338ada1ba3d7e853f417a6d61cafdb06227
Having some trouble linking this locally, so posting notes for public benefit.

rwjblue:
Fixed by #16474

Issue #12142, opened 2015-08-18
Documentation for route Transition object


fullsailor:
`Transition` is mentioned several times in the doucumentation for `Ember.Route`, but since it is defined in [tildeio/router.js](https://github.com/tildeio/router.js) it doesn't have documentation in Ember. 

There's brief documentation in [Route#beforeModel](http://emberjs.com/api/classes/Ember.Route.html#method_beforeModel), however its not everything. In particular I was looking for `Transition#send()` because an error mentioned it, but I didn't even know that was available.

/cc @locks

fullsailor:
What would be the correct way to document something without code in the Ember codebase? Just a file with the documentation comments?

rwjblue:
Transition comes from https://github.com/tildeio/router.js/blob/master/lib/router/transition.js and has some documentation there, but it needs cleanup to be properly parseable as YUIDoc.

Then we also need to update our `yuidoc.json` to include `bower_components/router.js/lib/` (similar to how we [include RSVP](https://github.com/emberjs/ember.js/blob/master/yuidoc.json#L19)).

mixonic:
@rwjblue is this considered complete?

iezer:
@rwjblue Looks like this was merged a while ago but the documentation is not on the website. If I build the docs locally off of `ember.js/master` I see the docs for `Transition` there. Maybe it requires a website release.

iezer:
CC @Gaurav0 ^^

Gaurav0:
@iezer Unless consensus is built to release earlier, I don't think it will appear before Ember 2.6 ish as it was only merged to canary. If you would like to see it sooner, please discuss on #documentation slack. cc/ @acorncom @locks

iezer:
That's great. Thanks.

Kerrick:
Unless I'm missing it, Transition is still not documented on emberjs.com -- and we're at 2.18.

fullsailor:
@Kerrick You're right, it looks like it was removed from the documentation in 2.17. The last time it was there was in 2.16. 

https://emberjs.com/api/ember/2.16/classes/Transition

toddjordan:
See #16153

Gaurav0:
@rwjblue Would copying the docs from router microlib repo to ember-source be an acceptable way to get this documented again?

rwjblue:
[...]

PR #16474, opened 2018-04-09
[DOC lts] Document Transition directly in source, remove yuidoc.json stuff

Gaurav0:
This is for #16153 

This takes an alternative approach to @efx 's [attempt](https://github.com/efx/ember.js/commit/7a884338ada1ba3d7e853f417a6d61cafdb06227) and simply copies the transition docs into ember-source, which haven't changed in a very long time. While a direct import might be preferred for synchronization purposes, this might be better so that transition docs don't simply get lost again.

If it is possible, I'd like to see this backported to all releases since 2.15.


toddjordan:
why is it that we can't just keep transition doc inline with its source?  Did we try changing lib to dist in yuidoc.json, or is there something keeping that from working?

Gaurav0:
I think we'd have to revert https://github.com/tildeio/router.js/blob/master/package.json#L26-L29 in order to do so. (The original source files are not included when you import router.js via npm.)

While we could do that, I'm afraid unless we have some automated testing for documentation around it, we will lose these docs in the future again, that's why I proposed simply copying them here: https://github.com/emberjs/ember.js/issues/12142#issuecomment-362302246

toddjordan:
A good way to check is from your ember.js project navigate to the path for transition in node_modules and see if the js file with documentation is there under dist.

Gaurav0:
@toddjordan @efx transition.js appears in that directory. In fact, running `yarn run docs` produces the Transition.html document on his branch, so I see no reason why he couldn't PR his branch as is.

rwjblue (approved):
I prefer this version, as future refactors in router.js microlib may add more files/directories that we might forget to add to exclude.  This seems better to me...

Gaurav0:
@rwjblue Why was this only merged to master? I wrote [DOC lts]

PR #16478, opened 2018-04-10
Null check for calculateToString method

lifeart:
Issue found in: https://travis-ci.org/emberjs/ember-inspector/jobs/364445291

https://github.com/emberjs/ember-inspector/pull/781

```javascript
Object.getPrototypeOf(null)
// Uncaught TypeError: Cannot convert undefined or null to object
//     at Function.getPrototypeOf (<anonymous>)
//     at <anonymous>:1:8
```
```
TypeError: Cannot convert undefined or null to object
                at Function.getPrototypeOf (<anonymous>)
                at calculateToString (http://localhost:7357/testing/assets/vendor.js:44972:29)
                at Mixin.classToString [as toString] (http://localhost:7357/testing/assets/vendor.js:44890:12)
                at mixins.forEach.mixin (http://localhost:7357/testing/ember_debug.js:3015:26)
                at Array.forEach (<anonymous>)
                at Class.mixinsForObject (http://localhost:7357/testing/ember_debug.js:3007:14)
                at Class.sendObject (http://localhost:7357/testing/ember_debug.js:2923:26)
                at Class.inspect (http://localhost:7357/testing/ember_debug.js:2347:35)
                at Object.<anonymous> (http://localhost:7357/testing/assets/tests.js:4637:22)
                at runTest (http://localhost:7357/testing/assets/test-support.js:4286:30)
```

How to reproduce?

```
function foo() {"use strict"; return this };
classToString(foo);
```


rwjblue (changes requested):
Seems good, can you create a regression test here that roughly simulates what the inspector is doing?

rwjblue:
Thank you for the added test! I'm still having trouble figuring out how this can actually happen though. Generally speaking, the following is being done:

```js
class Foo {}
Foo.prototype.toString = classToString;
```

How is it, in this circumstance, that you would have the `toString` function without a `this`?

rwjblue:
After a bit of digging, it _seems_ like the ember-inspector is essentially doing this (in the test that was failing from the CI log above):

```js
let obj = Ember.Object.create();
let mixins = Ember.Mixin.mixins(obj);
let mixinDetails = [];
mixins.forEach(mixin => {
  let name = mixin[Ember.NAME_KEY] || mixin.ownerConstructor;
  if (!name && typeof mixin.toString === 'function') {
    name = mixin.toString();
  }
  if (!name) {
    name = 'Unknown mixin';
  }
  mixinDetails.push({
    name: name.toString(), 
    properties: propertiesForMixin(mixin),
  });
});
```

Basically, what is going on is that the inspector is trying to call `.toString()` on each mixin and the current guard only checks for `Function.prototype` to break out of the loop which (since `this` is an instance at that point it would need to check for `Object.prototype` to break the loop).

packages/ember-metal/tests/namespace_search_test.js
```diff
@@ -0,0 +1,13 @@
+import { classToString } from '..';
+import { moduleFor, AbstractTestCase } from 'internal-test-helpers';
+
+moduleFor(
+  'NamespaceSearch',
+  class extends AbstractTestCase {
+    ['@test classToString: null as this inside class must not throw error'](assert) {
+      function F() {"use strict"; return this };
+      result = classToString(F.bind(null));
```
rwjblue: Can you update the regression test to do roughly:

```js
let mixin = Mixin.create();
mixin.toString(); // without the fix, this throws
```

packages/ember-metal/lib/namespace_search.js
```diff
@@ -184,6 +184,9 @@ function calculateToString(target) {
     }
     let superclass = target;
     do {
+      if (superclass === null) {
+        break;
+      }
       superclass = Object.getPrototypeOf(superclass);
       if (superclass === Function.prototype) {
```
rwjblue: Can you update this to be:

```js
if (superclass === Function.prototype || superclass === Object.prototype) {
  break;
}
```

packages/ember-metal/lib/namespace_search.js
```diff
@@ -184,6 +184,9 @@ function calculateToString(target) {
     }
     let superclass = target;
     do {
+      if (superclass === null) {
```
rwjblue: If we add `Object.prototype` check (mentioned in another comment) we can remove the explicit `null` check..

rwjblue (changes requested):

rwjblue:
Sorry for the run around here @lifeart, thank you for working on it!

lifeart:
@rwjblue `How is it, in this circumstance, that you would have the toString function without a this?`
https://github.com/emberjs/ember-inspector/pull/781/files#diff-7c2c7ea1dfc0ef384d1ec6e685711d15R480

rwjblue:
OK, looks great, would you mind squashing down to a single commit and prefixing the commit message with `[BUGFIX release]`?

Issue #16259, opened 2018-02-19
Assertion Failed: You attempted to access the `size` property


Turbo87:
on Ember 3.1.0-beta.1:

> Assertion Failed: You attempted to access the `size` property (of <DS.Errors:ember1800>).
Since Ember 3.1, this is usually fine as you no longer need to use `.get()`
to access computed properties. However, in this case, the object in question
is a special kind of Ember object (a proxy). Therefore, it is still necessary
to use `.get('size')` in this case.
>
> If you encountered this error because of third-party code that you don't control,
there is more information at https://github.com/emberjs/ember.js/issues/16148, and
you can help us improve this error message by telling us more about what happened in
this situation.

This seems to be caused by Ember Data using `isEmpty()` on a `DS.Errors` object. Since `isEmpty()` checks `obj.size` unconditionally it is causing the error above.

Turbo87:
Seems to be related to https://github.com/emberjs/ember.js/pull/9414

/cc @mixonic

rwjblue:
This was fixed in https://github.com/emberjs/data/pull/5340. The issue is that the DS.Errors unknownProperty returned null instead of undefined (and our proxy assertion only “allows” undefined).

Issue #16148, opened 2018-01-18
Further refine the error messages around allowing normal JavaScript property gets


wycats:
This issue is tracking problems that come up due to allowing normal JavaScript property lookup to work with computed properties.

## Context

We are in the process of landing a change that will allow you to use normal property lookups to get properties defined as computed properties in Ember.

In practice, what this means is that, in most cases (more later) you will be able to drop `.get('key')` and replace it with `.key`.

However, there are a handful of small historical issues that could make the transition a little more messy. If you landed on this issue, chances are you encountered a warning from Ember that you encountered one of these historical issues.

## Computed Properties used to be directly on the prototype

Until Ember 3.0, internal implementation details of Ember caused computed properties to be directly present on objects, so if you didn't use `.get()`, you would see the internal "Ember descriptor" representing the computed property (a private API).

This internal implementation detail was never intended to be a public (or even intimate) API.

This internal implementation detail has now changed and the (still private) "descriptor" object has been relocated to the object's "meta" (also a private API). This allows us to make normal property lookup work, and in general, very few people were relying on the presence of this internal descriptor.

If you encountered an error related to this problem, you are probably using an addon that relies on this now-defunct private implementation detail, and your best bet is to follow the instructions in the error message (and ask for help in #dev-ember).

## Ember Proxies are tricky

Secondly, since the new behavior relies on using `defineProperty`, it does **not** allow users of Ember proxies to stop using `.get()`.

Previously, since **all** Ember objects required you to use `.get()` to access properties, you could uniformly use `.get()` in your code and never need to worry about whether an object was an Ember proxy.

However, now that you can use normal property access for computed properties, Ember proxies are special. If an object is implemented using an Ember proxy, it should advertise itself as requiring `.get()` to access its proxied properties.

This is analogous to other libraries in the JavaScript ecosystem, such as `immutable.js`, that expose an API that is different from normal JS property access patterns.

Since proxies are relatively rare in Ember compared to property access on other Ember objects, emberjs/rfcs#281 (now merged) felt that the cost of losing consistent property access (everyone types `.get()` everywhere, so you don't need to know when an object is "special") was worth the benefit of normally not needing to use `.get()`.

During the transition, we are worried that people will accidentally migrate some `.get()`s on Ember proxies to regular property access, so we added an assertion to the code that fires if:

- you use a normal property get on an Ember proxy
- the property is not present on the object itself
- `unknownProperty` returns a value other than undefined

[...]

PR #16485, opened 2018-04-11
adjust assertion to allow for either undefined or null values

Dhaulagiri:
Based on conversation with @rwjblue in https://github.com/emberjs/ember.js/issues/16148 he suggested adjusting this assertion to allow `value` to be null.  I confirmed locally that this fixes the instances of https://github.com/emberjs/ember.js/issues/16259 we are seeing which we are unable to correct with the upstream Ember Data fix just yet.

If this looks good I can adjust the commit message as needed, but ideally this would be a [BUGFIX Release].


rwjblue:
Seems good to me, `[BUGFIX release]` seems like a good target...

Dhaulagiri:
@rwjblue i also targeted the release branch because master wouldn't let my app boot for reasons I didn't look into, is that a valid target as well?

rwjblue:
Sadly due to somewhat arcane release processes (that are completely my fault) we need to target this PR at master, then cherry-pick (or land separate PRs) into beta and release.

rwjblue:
Sorry for the run around @Dhaulagiri, thank you!

PR #16493, opened 2018-04-11
[BUGFIX release] Ensure proxies have access to `getOwner(this)`.

### LINKED ISSUE #16484: [3.1.0] getOwner returning false in ObjectProxy.init
`getOwner(this)` in `ObjectProxy.init` was returning the container in 3.0. Since 3.1 it returns false. 
 `this` looks quite different:

3.0:
screen shot 2018-04-11 at 16 33 57

3.1:
screen shot 2018-04-11 at 16 41 28

rwjblue:
Prior to this change, when in debug builds, any "proxy" (i.e. something with `unknownProperty`) would _not_ have access to an `owner` object via `getOwner(this)`. This is because of our association with the `FACTORY_FOR` weakmap is only setup for the actual `CoreObject` instance being created, but when `unknownProperty` exists we never expose the "real" `CoreObject` instance and instead expose a native JS `Proxy` (so that we can trap usage of `instance.foo` and properly error).

This fix ensures that the native proxy that is returned is also associated with the factory, and can properly access `getOwner(this)` / `._debugContainerKey` / etc.

Fixes #16484

### DIFF
--- a/packages/ember-runtime/lib/system/core_object.js
+++ b/packages/ember-runtime/lib/system/core_object.js
@@ -130,6 +130,8 @@ function makeCtor(base) {
               assert(messageFor(receiver, property), value === undefined);
             },
           });
+
+          FACTORY_FOR.set(self, FACTORY_FOR.get(this));
         }
 
         let m = meta(self);
--- a/packages/ember-runtime/tests/system/core_object_test.js
+++ b/packages/ember-runtime/tests/system/core_object_test.js
@@ -1,4 +1,4 @@
-import { getOwner } from 'ember-utils';
+import { getOwner, setOwner } from 'ember-utils';
 import { get } from 'ember-metal';
 import CoreObject from '../../lib/system/core_object';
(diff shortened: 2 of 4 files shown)


rwjblue:
I've pulled these changes into the beta and release branches, you should be able to use the latest release / beta builds once CI is completed and the changes should be included in the next release of each branch...

niranjan94:
Hi @rwjblue ! Thanks was the quick fix 😄 

> you should be able to use the latest release / beta builds once CI is completed

I'm unable to find the latest `3.1.x` release which has this fix (`3.1.1` - _the latest release_, does not). Am I missing something ? 🤔

Dhaulagiri:
@niranjan94 I had the same question yesterday, the answer can be found [here](https://github.com/emberjs/ember.js/pull/16494#issuecomment-380615805)

niranjan94:
Thanks a lot @Dhaulagiri ! That works.

PR #16494, opened 2018-04-11
[BUGFIX release] adjust assertion to allow for either undefined or null

Dhaulagiri:
https://github.com/emberjs/ember.js/pull/16485, but targeted against master

### DIFF
--- a/packages/ember-runtime/lib/system/core_object.js
+++ b/packages/ember-runtime/lib/system/core_object.js
@@ -127,7 +127,7 @@ function makeCtor(base) {
 
               let value = target.unknownProperty.call(receiver, property);
 
-              assert(messageFor(receiver, property), value === undefined);
+              assert(messageFor(receiver, property), value === undefined || value === null);
             },
           });
         }


rwjblue:
Thank you!

rwjblue:
I've pulled these changes into the beta and release branches, you should be able to use the latest release / beta builds once CI is completed and the changes should be included in the next release of each branch...

Dhaulagiri:
> you should be able to use the latest release / beta builds

@rwjblue is there a good place to find these or can we point package.json at `'ember/ember.js#release'`?

rwjblue:
Turns out that I didn't have a good way to explain this, so I just made it simpler (in https://github.com/ember-cli/ember-source-channel-url/pull/2).

This should be all you need to do to get the URL to put in your package.json:

```
npm install -g ember-source-channel-url
ember-source-channel-url release
```

That will print out the current URL for that channel, and you can put that directly in your `package.json` like:

```js
//...snip...
"ember-source": "<URL GOES HERE>",
//...snip...
```

Dhaulagiri:
Nice! I’ll try to see if there is a way to better document this in the ember repo.

PR #16507, opened 2018-04-12
Migrate getOwner/setOwner to @ember/application.

rwjblue:


ef4:
This appears to have caused `getOwner` to drop out of the API docs. I bisected that issue and ended up here.

rwjblue:
@ef4 - in which version?

rwjblue:
That should have been addressed by 
https://github.com/emberjs/ember.js/pull/16836

PR #16514, opened 2018-04-13
[BUGFIX beta] Bring back (with deprecation) Ember.EXTEND_PROTOTYPES.

### LINKED ISSUE #16511: [regression Canary] Ember.EXTEND_PROTOTYPES is undefined
In https://github.com/emberjs/ember.js/pull/16467 `Ember.EXTEND_PROTOTYPES` was removed from the framework. Common libraries like ember-inflector use this value ([source](https://github.com/emberjs/ember-inflector/blob/2316f89c59769b45f684308d4a3dce67bfdc0ac3/addon/lib/ext/string.js#L7)) and we need to have an intimate API deprecation cycle here.

For example:

```js
Object.defineProperty(Ember, 'EXTEND_PROTOTYPES', {
  enumerable: false,
  get() {
    deprecate('Accessing Ember.EXTEND_PROTOTYPES is deprecated, please migrate to Ember.ENV.EXTEND_PROTOTYPES');
    return ENV.EXTEND_PROTOTYPES;
  }
});
```

(where the value used to be set: [source](https://github.com/emberjs/ember.js/blob/9a81db740367499e1b35540b20751f7577a631d6/packages/ember/lib/index.js#L149))

/cc @rwjblue

rwjblue:
Fixes #16511

### DIFF
--- a/packages/ember/index.js
+++ b/packages/ember/index.js
@@ -110,6 +110,7 @@ import Resolver from '@ember/application/globals-resolver';
 import ApplicationInstance from '@ember/application/instance';
 import Engine from '@ember/engine';
 import EngineInstance from '@ember/engine/instance';
+import { deprecate } from 'ember-debug';
 
 // ****ember-environment****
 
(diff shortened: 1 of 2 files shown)


ursm:
@rwjblue 
This fix seems to be causing an error. Would you confirm it?

```
Uncaught Error: Assertion Failed: When calling `deprecate` you must provide an `options` hash as the third parameter.  `options` should include `id` and `until` properties.
```

rwjblue:
Yep. You are correct :sob:

PR #16516, opened 2018-04-13
[bugfix] Fix `proto` return value for native classes

pzuraq:
Native classes end up calling the nearest class constructor, and with
the recent refactor those class constructors return their own prototype
instead of the class instance's. This breaks certain functions that
currently rely on the `proto` function return the prototype of the
instance itself. We're seeing failures in `@ember-decorators/data`
due to this bug.

This reverts to returning the protoype on the instance directly. 

The other option would be to instead make `proto` a purely side-effecting
function that doesn't return anything, and change the instances where
it is called to use the prototype of the instance they are calling it on. Unsure
if there are any perf implications here.

cc @rwjblue @krisselden

### DIFF
--- a/packages/ember-runtime/lib/system/core_object.js
+++ b/packages/ember-runtime/lib/system/core_object.js
@@ -241,7 +241,7 @@ function makeCtor(base) {
     };
   }
 
-  Class.willReopen = () => {
+  Class.willReopen = function() {
     if (wasApplied) {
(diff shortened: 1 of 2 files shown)


rwjblue:
Thank you!

pzuraq:
Any chance this could be cherry-picked onto beta? Still having failures for data decorators on the beta branch now

PR #16520, opened 2018-04-14
[bugfix] Adds options checking ability to debug/deprecation test helpers

pzuraq:
With Ember v3 we now require that deprecations receive `id` and `until`
options. It's a hard error if they are not included, as happened
recently. This PR adds the ability to check for options when asserting
in debug helpers, so we can assert that all deprecations internally
receive these options. It also updates all outstanding cases where an
id/until value were not passed.

### DIFF
--- a/packages/ember-metal/tests/properties_test.js
+++ b/packages/ember-metal/tests/properties_test.js
@@ -81,7 +81,7 @@ moduleFor(
       assert.expect(3);
       let obj = { foo: 'bar' };
 
-      deprecateProperty(obj, 'baz', 'foo');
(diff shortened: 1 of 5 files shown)


pzuraq:
In general I matched the existing style of the MethodCallTracker class, but I'd be happy to refactor it to use modern class syntax and idioms. Wanted to make sure this was the correct direction before diving into that.

rwjblue:
Thank you very much for making sure the ember-dev assertion helpers flag this!

rwjblue (approved):

rwjblue:
Will need to cherry pick this into beta...

PR #16526, opened 2018-04-16
[BUGFIX beta] Ensure setting a `NAME_KEY` does not error.

### LINKED ISSUE #16479: [BUG][CANARY] Mixins throwing  `object is not extensible`
Both addon are throwing `TypeError: Cannot add property undefined, object is not extensible`
https://github.com/offirgolan/ember-cp-validations/blob/master/addon/validations/factory.js#L146
https://github.com/offirgolan/ember-parachute/blob/master/addon/query-params.js#L293

`ValidationsMixin[Ember.NAME_KEY] = 'Validations';`
Ember.NAME_KEY is `undefined`
ValidationsMixin seems to be immutable 

Addons using `Ember.NAME_KEY`
https://emberobserver.com/code-search?codeQuery=NAME_KEY&sort=score&sortAscending=false

rwjblue:
Fixes #16479

### DIFF
--- a/packages/ember-metal/lib/mixin.js
+++ b/packages/ember-metal/lib/mixin.js
@@ -2,7 +2,7 @@
 @module @ember/object
 */
 import { EMBER_METAL_ES5_GETTERS } from 'ember/features';
-import { assign, guidFor, ROOT, wrap, makeArray } from 'ember-utils';
+import { assign, guidFor, ROOT, wrap, makeArray, NAME_KEY } from 'ember-utils';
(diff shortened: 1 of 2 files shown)


blimmer:
I think I might be experiencing a similar issue with Google Recaptcha + Ember 3.

https://github.com/google/recaptcha/issues/285

In stepping through the debugger, the call stack seems to be when mixing in the `transform:date`

screen shot 2018-11-19 at 16 50 45

PR #16537, opened 2018-04-17
Migrate Ember.Service to @ember/service.

rwjblue:


rwjblue:
General r+ from @krisselden in chat on the `CoreObject` stuff specifically...

PR #16542, opened 2018-04-18
use `Object.defineProperties` for extending String and Function

bekzod:
just think it is much cleaner and readable this way, thoughts ?


krisselden (approved):

rwjblue:
Thank you!

PR #16543, opened 2018-04-18
collapse `if` statement in `defineProperty`

bekzod:


rwjblue (approved):
Nice catch, thank you!

PR #16544, opened 2018-04-18
Move Cache from ember-metal to ember-utils.

rwjblue:
Also remove usage of `UNDEFINED` symbol (just use normal Map behaviors).


packages/ember-utils/lib/cache.ts
```diff
@@ -0,0 +1,52 @@
...
+    private limit: number,
+    private func: (obj: T) => V,
+    private key?: (obj: T) => string | T,
+    private store?: any
+  ) {
+    this.store = store || new Map();
+  }
+
+  get(obj: T): V {
+    let key = this.key === undefined ? obj : this.key(obj);
```
krisselden: this path isn't used anymore

rwjblue: Great catch, updated!

PR #16548, opened 2018-04-19
Migrate ember-debug to @ember/debug and @ember/canary-features.

rwjblue:
* Move `Ember.assert`, `Ember.deprecate`, `Ember.warn`, `Ember.info`, `Ember.debug` to `@ember/debug`.
* Move `Ember.FEATURES`, `Ember.FEATURES.isEnabled` to `@ember/canary-features`
* Remove "build time generation" of feature flags module.
* Remove features information from `features.json` and inline it into `@ember/canary-features/index.ts`. The build time replacement still occurs, but the TS file is now the "source of truth".
* Remove `externs.d.ts`, it was only used for tracking feature flags which now live in a TS file anyways.
* Remove "warn on feature flags when running non-canary" concept. This can later be re-added, but should be build time (not runtime) as a normal addon warning.
* Remove `EMBER_GLIMMER_DETECT_BACKTRACKING_RERENDER` feature flag (replace it with a simple `if (DEBUG) {}` guard).
* Remove final remnants of "debug features" vs "production features". Now all features are the same for debug vs prod.
* Fix issue with `.gitignore` (previously disallowed any folders named `DEBUG`).


runspired:
@rwjblue you might be interested in this: https://emberjs.com/api/ember/3.1/functions/@ember%2Fapplication%2Fdeprecations/deprecate

Basically we've all been doing it wrong and `deprecate` is exported in two locations. The RFC was for the package in the docs, although I think the debug package makes way more sense.

rwjblue:
Yep, I knew of that, but since it makes _zero_ sense to me coming from `@ember/application/deprecations` **and** `@ember/debug` has special behaviors with ember-cli-babel (auto-stripping and whatnot) I kept these together. 

Likely will have to make a reexport from `@ember/application/deprecations` to satisify the docs and eventually deprecate that location...

PR #16550, opened 2018-04-19
[BUGFIX Beta] Decrement counter of pending requests in the next tick

cibernox:
Made to align with the implementation in ember-test-helpers: https://github.com/emberjs/ember-test-helpers/blob/7f9a3d082958ba993c6db863886d370d73fd98ef/addon-test-support/%40ember/test-helpers/settled.js#L53-L70


rwjblue:
`pendingRequests is maintained for ajaxSend and ajaxComplete events` is failing because it assumes things are sync, you will need to update those tests to deal with the new async...

cibernox:
@rwjblue Fixed!

rwjblue (approved):

Issue #16503, opened 2018-04-12
Uncaught TypeError: Cannot read property 'syscall' of null


nullrocket:
I have been banging my head against this every time I try to update from 3.0.0 to any >= 3.1 version beta or final.   
It happens after transitioning in and out of a particular route exactly 3 times.  Unfortunately I'm unable to create a reproduction except in this application and the stacktrace doesn't point to anything I can dig into.   Any idea where I can start looking?
  
``` javascript
AppendOpcodes.prototype.evaluate = function evaluate(vm$$1, opcode, type) {
            var operation = this.evaluateOpcode[type];
             if (operation.syscall) {  // here is where it errors.  type argument is 0
                operation.evaluate(vm$$1, opcode);
            } else {
                operation.evaluate(vm$$1.inner, opcode);
            }
        };

**Comment by Turbo87** [comment]
we are seeing similar issues in our Sentry logs after upgrading to Ember 3.1 🤔 

the stacktraces are not giving any hints unfortunately. this is one example from Safari:

```
TypeError: null is not an object (evaluating 'n.syscall')
  at evaluate(@glimmer/runtime.js:29:1)
  at evaluateSyscall(@glimmer/runtime.js:3310:1)
  at evaluateInner(@glimmer/runtime.js:3286:1)
  at evaluateOuter(@glimmer/runtime.js:3278:1)
  at next(@glimmer/runtime.js:4865:1)
  at execute(@glimmer/runtime.js:4850:1)
  at handleException(@glimmer/runtime.js:4363:1)
  at handleException(@glimmer/runtime.js:4532:1)
  at throw(@glimmer/runtime.js:4272:1)
  at evaluate(@glimmer/runtime.js:1105:1)
  at execute(@glimmer/runtime.js:4259:1)
  at methodName(@glimmer/runtime.js:4558:1)
  at runInTransaction(ember-metal.js:1152:1)
  at _renderRoots(ember-glimmer.js:4496:1)
  at _renderRootsTransaction(ember-glimmer.js:4529:1)
  at call(ember-glimmer.js:4572:1)
  at invoke(backburner.js:227:1)
  at flush(backburner.js:134:1)
  at flush(backburner.js:290:1)
  at end(backburner.js:423:1)
  at run(backburner.js:800:1)
```

and another from Chrome:

```
TypeError: Cannot read property 'syscall' of null
  at evaluate(@glimmer/runtime.js:29:1)
[...]

PR #16558, opened 2018-04-20
Ensure ComponentDefinitions do not leak heap space.

rwjblue:
Addresses https://github.com/emberjs/ember.js/issues/16503


krisselden:
I'm pretty cure we could just use a Map, and use the component or the template if there is no component as the key.

Issue #16477, opened 2018-04-10
[ember 3.1.0-beta.1] viewBox attribute on SVG component has incorrect capitalization 


mlwilkerson:
**Summary**

When rendering a component with `tagName` of `svg` and a bound `viewBox` attribute, the resulting DOM element has attribute `viewbox` (without the capital B).

**Impact:** the SVG will have the wrong size. (Observed in Chrome).

[Here is a reproduction of the issue.](https://github.com/mlwilkerson/ember-svg-viewbox-bug)
(I couldn't figure out how to JSBin or Ember-Twiddle it).

This problem does not exist in ember 3.0.0.

Similar past issue: https://github.com/emberjs/ember.js/issues/10303

st-h:
looks like this unfortunately persists with the 3.1.0 release

st-h:
Copied over from #16311 as it seems this applies to this issue only

> st-h:
I tried to create a test first, however my efforts to write something that matches the current implementations only showed that the current test setup happily accepts any diffs in case sensitive bindings (both in glimmer.vm and ember.js). Probably the test helpers should be refined... However, there are too many internals involved and I don't know the codebase well enough to make a substantial decision here.
So, I continued to find out what is actually going wrong and I believe to have found the offending [commit](https://github.com/emberjs/ember.js/commit/30bb21fa907f16a736599277911ecf41cc30a1ad) / [code](https://github.com/emberjs/ember.js/blame/master/packages/ember-glimmer/lib/utils/bindings.ts#L66). This essentially changed using the lower case name of the attribute when setting it within the [install method](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/utils/bindings.ts#L113)
Yet again, being very new to the ember/glimmer internals, I don't feel comfortable deciding what is the best way to fix this. I am happy to help getting this done with a little support though.

> rwjblue:
> Hmm. That change definitely does odd. Do Ember's own tests pass if you remove that .toLowerCase()? I'm unsure how this could be the cause of <button onClick=(action 'foo')></button)> not working because it seems like it would only be related to component's attribute bindings right?

> st-h:
@rwjblue maybe I should have posted that to the [svg related issue](https://github.com/emberjs/ember.js/issues/16477), as I am not totally sure if both issues really do have the same cause. However, I was specifically looking why an attributeBinding like `viewBox` would end up as `viewbox`.
[...]

Issue #16311, opened 2018-03-02
onEventName= vs. oneventname= regression in Ember 3.1-beta


mwpastore:
The following template code works in 3.0 (and as far as I know Ember 1.x and 2.x) but not in 3.1-beta:

```hbs
<button onClick={{action 'foo'}}>
```

Down-casing the attribute to `onclick` "solves" the problem. 

Developers coming from other frameworks or vanillajs probably expect this to work. And whether it should or shouldn't work, it will probably break some folks upgrading to Ember >3.0.

rwjblue:
Thanks for reporting! I'm not 100% sure if this will end up being considered a regression or not, I'll try to discuss in more detail with the team and decide...

tomdale:
Per discussion at the core team meeting, we want to treat this as a regression. Off the top of my head I'm not aware of any changes in the VM that should have caused this, so will need some investigation.

rwjblue:
@st-h - TBH, we mostly need someone to dig and help track down why we regressed. Perhaps a great first step is submitting a failing test case PR both here and in glimmer-vm...

st-h:
@rwjblue reporting back from my first extensive debugging adventure through the glimmer and ember internals :neckbeard: 
I tried to create a test first, however my efforts to write something that matches the current implementations only showed that the current test setup happily accepts any diffs in case sensitive bindings (both in glimmer.vm and ember.js). Probably the test helpers should be refined... However, there are too many internals involved and I don't know the codebase well enough to make a substantial decision here.
So, I continued to find out what is actually going wrong here and I believe to have found the offending [commit](https://github.com/emberjs/ember.js/commit/30bb21fa907f16a736599277911ecf41cc30a1ad) / [code](https://github.com/emberjs/ember.js/blame/master/packages/ember-glimmer/lib/utils/bindings.ts#L66). This essentially changed using the lower case name of the attribute when setting it within the [install method](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/utils/bindings.ts#L113)
Yet again, being very new to the ember/glimmer internals, I don't feel comfortable deciding what is the best way to fix this. I am happy to help getting this done with a little support though.

rwjblue:
Hmm. That change definitely does odd. Do Ember's own tests pass if you remove that `.toLowerCase()`?  I'm unsure how this could be the cause of `<button onClick=(action 'foo')></button)>` not working because it seems like it would only be related to component's attribute bindings right?

[...]

PR #16559, opened 2018-04-21
Fix property normalization

ef4:
This depends on https://github.com/glimmerjs/glimmer-vm/pull/801, once that is released we should edit this PR to also upgrade the glimmer packages (I made a small breaking change in the glimmer Environment API).

Fixes https://github.com/emberjs/ember.js/issues/16311 
Fixes https://github.com/emberjs/ember.js/issues/16477

The root cause was that glimmer was not doing correct property normalization, and Ember was trying -- incorrectly -- to compensate for that with a `toLowerCase`, which was enough to get our existing tests to pass but was not the correct semantics.


mehulkar:
👏 thank you!

Issue #16521, opened 2018-04-15
Special kind of Ember object (a proxy) error thrown in ember 3.1


toranb:
I recently upgraded an ember addon from ember 3.0 => 3.1 and found a model class that is created via [Proxy](https://github.com/toranb/ember-cli-simple-store/blob/master/addon/models/record-proxy.js) like so ... (note: RecordProxy is using `ObjectProxy from '@ember/object/proxy'`)

```js
return RecordProxy.create({
  _store: this,
  _type: type,
  _filter_value: actualId,
  _source: this._findAll(type),
  compute() {
    let filter_value = this.get("_filter_value");
    return this.get("_source").findBy(primaryKey, filter_value);
  }
});
```

And constructed it using a model object with a non computed property (ie: vanilla `function`)

```js
import { Model } from 'ember-cli-simple-store/model';

export default Model.extend({
  change_role() {
  }
});
```

Is now throwing an error like so ...

```You attempted to access the `change_role` property (of <(unknown mixin):ember244>).
Since Ember 3.1, this is usually fine as you no longer need to use `.get()`
to access computed properties. However, in this case, the object in question
is a special kind of Ember object (a proxy). Therefore, it is still necessary
to use `.get('change_role')` in this case.```

I'm curious if this is a regression or future deprecation just waiting to happen ;)

If I understand the error  ... as of ember 3.1 I would need to access this `change_role` function with a `Ember.get` ... but prior I could invoke it like a normal function (without `get`) but because it's a function (and not a computed/ or ES5 GETTER) I wouldn't anticipate having to use an explicit `Ember.get` to invoke it.

I had trouble creating this w/ ember twiddle but I was able to `ember new` using 3.1 to prove it's broken. Note: this app works fine if you were to flip the source dependency back to ember 3.0

https://github.com/toranb/simple-store-busted-ember-three-one/commit/427e8057588c1c971474084c48f7b767ff44eb44

Any helpful pointers on how to move forward here? Thank you in advance!

rwjblue:
The issue in this case with ember-cli-simple-store (which doesn't seem terribly simple to me 😉 ) is [here](https://github.com/toranb/ember-cli-simple-store/blob/493398261f9f11eace838b5a9959aeb9079a3fff/addon/models/record-proxy.js#L23). In Ember 3.1 doing that check will trigger the reported assertion because the `unknownProperty` of the proxy assertion will find `change_role` in the underlying model. 

[...]

PR #16560, opened 2018-04-21
[BUGFIX release] avoid strict assertion when object proxy calls thru for function

toranb:
Upgrading to ember v3.1 recently I bumped into this strict assertion around `ObjectProxy` which broke the consumers of that addon. More details (including a full example app) and discussion w/ Robert are included in the original issue I opened last week

https://github.com/emberjs/ember.js/issues/16521

### DIFF
--- a/packages/ember-runtime/lib/system/core_object.js
+++ b/packages/ember-runtime/lib/system/core_object.js
@@ -116,7 +116,9 @@ function makeCtor(base) {
 
               let value = target.unknownProperty.call(receiver, property);
 
-              assert(messageFor(receiver, property), value === undefined || value === null);
+              if (typeof value !== 'function') {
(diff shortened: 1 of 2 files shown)


rwjblue:
We discussed this exact thing at the core team meeting yesterday, and came to the same conclusion, and we are 👍 on moving forward with this tweak to the proxy assertion system. 

@toranb - Two small changes here and this is good to land (and can ultimately make its way to 3.1.x):

1. Looks like the code-quality job failed, should be able to fix via `yarn lint:fix`.
2. Can you prefix the commit (and PR title) with `[BUGFIX release]`?

toranb:
@rwjblue done and done my friend. Thanks for considering this patch for 3.1X :) I would greatly appreciate it!

rwjblue:
Thank you!

PR #16561, opened 2018-04-21
cleanup how component `id` attr is set

bekzod:
there is asymmetry how `id` attributes are set, this fixes it

### DIFF
--- a/packages/ember-glimmer/lib/component-managers/curly.ts
+++ b/packages/ember-glimmer/lib/component-managers/curly.ts
@@ -90,7 +90,8 @@ function applyAttributeBindings(
   }
 
   if (seen.indexOf('id') === -1) {
-    operations.setAttribute('id', PrimitiveReference.create(component.elementId), true, null);
+    let id = component.elementId ? component.elementId : guidFor(component);
+    operations.setAttribute('id', PrimitiveReference.create(id), false, null);
   }
 
   if (seen.indexOf('style') === -1) {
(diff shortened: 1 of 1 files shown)


packages/ember-glimmer/lib/component-managers/curly.ts
```diff
@@ -64,36 +64,26 @@ function isTemplateFactory(template: OwnedTemplate | TemplateFactory): template
...
-    if (seen.indexOf(attribute) === -1) {
-      seen.push(attribute);
-      AttributeBinding.install(element, component, parsed, operations);
-    }
-
-    i--;
-  }
-
-  if (seen.indexOf('id') === -1) {
-    operations.setAttribute('id', PrimitiveReference.create(component.elementId), true, null);
```
bekzod: `id` was set differently than when `attributeBindings` didn't exist https://github.com/emberjs/ember.js/pull/16561/files#diff-1d6263d5fa618e89a735e9df0fe95f81L327

packages/ember-glimmer/lib/component-managers/curly.ts
```diff
@@ -73,27 +73,26 @@ function applyAttributeBindings(
   component: Component,
   operations: ElementOperations
 ) {
-  let seen: string[] = [];
+  let seen: Set<String> = new Set();
```
krisselden: I doubt Set to be actually faster for less than < 10 attributes and most elements don't have that many.

We should definitely try to test this.

krisselden (changes requested):

packages/ember-glimmer/lib/component-managers/curly.ts
bekzod: such thought crossed my mind too, so reverting this changes :P

krisselden (changes requested):
Can you add a test and label as a bugfix? I'm not sure what you are fixing here.

bekzod:
not fixing here anything just avoiding calling `operations.setAttribute('id'.....` twice, initially thought that there is bug here because `operations.setAttribute('id'` was called twice @krisselden

rwjblue:
The changes here seem good, but I did want to point out that `ComponentOperations` (the thing called `operations` in `didCreateElement`) is specifically built to absorb the duplicate attribute setting in a "last one wins" strategy, and it avoids doing the DOM manipulation more than once.

PR #16562, opened 2018-04-21
correct conditionals in `component`

bekzod:


rwjblue (commented):
Does this fix any specific thing? Can we add a regression test for it?

packages/ember-glimmer/lib/component.ts
```diff
@@ -604,10 +604,8 @@ const Component = CoreView.extend(
       let args = this[ARGS];
       let reference = args && args[key];
 
-      if (reference) {
-        if (reference[UPDATE]) {
-          reference[UPDATE](get(this, key));
-        }
+      if (reference && UPDATE in reference) {
```
krisselden: lets not use then, use reference[UPDATE] !== undefined

krisselden: in general we should avoid falsy conversions

krisselden (changes requested):

krisselden:
@rwjblue he is just collapsing the if, there isn't change to behavior

packages/ember-glimmer/lib/component.ts
```diff
@@ -602,12 +602,10 @@ const Component = CoreView.extend(
...
 
       let args = this[ARGS];
-      let reference = args && args[key];
+      let reference = args !== undefined ? args[key] : undefined;
 
-      if (reference) {
-        if (reference[UPDATE]) {
-          reference[UPDATE](get(this, key));
-        }
+      if (reference !== undefined && UPDATE in reference) {
```
rwjblue: Can you update this to:

```js
if (reference !== undefined && reference[UPDATE] !== undefined) {
}
```

(no need to use `in` here AFAICT)

rwjblue (changes requested):

krisselden (approved):

rwjblue (approved):

PR #16566, opened 2018-04-22
use hash for `seen` in `Unique`

bekzod:


packages/ember-glimmer/lib/utils/iterable.ts
```diff
@@ -452,7 +449,7 @@ function KeyPath(keyPath: string): KeyFor {
 }
 
 function Unique(func: KeyFor): KeyFor {
-  let seen = new Set();
```
bekzod: seen was treated as empty object

rwjblue (approved):

rwjblue:
Thank you!

PR #16567, opened 2018-04-22
extract `uniqBy` as generic method and reuse internally

bekzod:
@mmun is this worth doing ? 😉

### DIFF
--- a/packages/ember-runtime/lib/utils.js
+++ b/packages/ember-runtime/lib/utils.js
@@ -1,8 +1,9 @@
 import { DEBUG } from '@glimmer/env';
-import { PROXY_CONTENT } from 'ember-metal';
+import { PROXY_CONTENT, get } from 'ember-metal';
 import { HAS_NATIVE_PROXY } from 'ember-utils';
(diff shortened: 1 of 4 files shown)


packages/ember-runtime/lib/utils.js
```diff
@@ -168,3 +169,21 @@ export function typeOf(item) {
 
   return ret;
 }
+
+export function uniqBy(array, key) {
+  assert(`first argument passed to \`uniqBy\` should be array`, isArray(array));
+
+  let ret = A();
+  let seen = new Set();
+  let getter = arguments.length === 1 ? item => item : get;
```
mmun: can you do something like
```js
const identityFunction = item => item;
export function uniqBy(array, key = identityFunction) {
```

bekzod: alright

mmun (approved):
This seems fine to do.

bekzod:
green

rwjblue:
Awesome, thank you!

PR #16570, opened 2018-04-22
reuse `indexOf`

bekzod:


packages/ember-runtime/lib/mixins/array.js
```diff
@@ -1094,10 +1078,11 @@ const ArrayMixin = Mixin.create(Enumerable, {
     }
 
     let ret = A();
+    // SameValueZero comparison (NaN !== NaN)
+    let predicate = value === value ? item => item === value : item => item !== item;
```
rwjblue: Why move this up to a function? Doesn’t seem to simplify or correct anything here...

rwjblue (commented):
Looks good overall, I commented on one specific section that might need to be reverted...

rwjblue (approved):

rwjblue:
Thanks!

PR #16571, opened 2018-04-22
[BUGFIX release] fix issue with class not recomputing

### LINKED ISSUE #16505: [3.1.0] "class" attribute not being recomputed
We have something like this in a template:

```hbs
{{some-component classNames="some-class fade" class=(if isLoggedIn "in")}}
```

This used to work fine until `3.1.0` - that is, when `isLoggedIn` is true, in addition to the classes in `classNames`, `in` is added. When `isLoggedIn` is false, then only the classes from `classNames` are used. However, in `3.1.0` this doesn't work any more. `in` is not added even when `isLoggedIn` is true. It works only if we add a non-empty string to the else clause - `(if isLoggedIn "in" "some-non-relevant-class")`.

bekzod:
fixes #16505

### DIFF
--- a/packages/ember-glimmer/lib/component-managers/curly.ts
+++ b/packages/ember-glimmer/lib/component-managers/curly.ts
@@ -330,11 +330,13 @@ export default class CurlyComponentManager
       IsVisibleBinding.install(element, component, operations);
     }
 
-    if (classRef && classRef.value()) {
+    if (classRef) {
+      let val = classRef.value();
       const ref =
-        classRef.value() === true
+        val === true || val === false
           ? new ColonClassNameBindingReference(classRef, dasherize(classRef['_propertyKey']), null)
           : classRef;
+
       operations.setAttribute('class', ref, false, null);
     }
 


packages/ember-glimmer/tests/integration/components/curly-components-test.js
```diff
@@ -404,9 +404,9 @@ moduleFor(
     ['@test class property on components can be dynamic']() {
       this.registerComponent('foo-bar', { template: 'hello' });
 
-      this.render('{{foo-bar class=(if fooBar "foo-bar")}}', {
-        fooBar: true,
-      });
+      this.render('{{foo-bar class=(if fooBar "foo-bar")}}');
+
+      this.runTask(() => set(this.context, 'fooBar', true));
```
bekzod: this would break test before

rwjblue:
Thank you for fixing! Can you grab the tests from this commit by @ef4 (and reset this particular test)?

https://github.com/emberjs/ember.js/commit/5f7e7294bf111d1cb82237d6c8deacffc4e4e364

rwjblue:
Also, needs to be `[BUGFIX release]` prefix I believe (since it is broken in current release branch)...

ef4:
I don't think this implementation is correct. It will not update is the value goes from boolean to non-boolean. 

Basically it's always very suspicious to look at `value()` when setting up bindings -- it locks you into whatever behavior goes with the value you had at that moment.

bekzod:
@ef4 cool, your solution looks more legit 👍, closing this one

PR #16573, opened 2018-04-22
access `length` directly

bekzod:
since support of native getters, we could access `array.length` directly  without `get`


mmun (approved):

rwjblue:
👏 thank you!

bekzod:
awesome 🎊

PR #16575, opened 2018-04-23
[DOC] Proper usage of _super in ArrayProxy init()

frank06:
The documentation for `ArrayProxy` does not clarify the importance of placing the call to `_super`**AFTER** setting the content, which is common when overriding the class.  Not doing this caused me so many hours of pain and frustration because the array "kind of" works, but sometimes behaves in a very weird way, that I believe it's imperative to have it documented.

### DIFF
--- a/packages/ember-runtime/lib/system/array_proxy.js
+++ b/packages/ember-runtime/lib/system/array_proxy.js
@@ -60,6 +60,22 @@ const ARRAY_OBSERVER_MAPPING = {
   ap.get('firstObject'); // . 'DOG'
   ```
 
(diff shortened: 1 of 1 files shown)


jenweber:
Is the ArrayProxy be something that should be overidden?

frank06:
Thanks Jen. I don't see why not. See this: https://stackoverflow.com/questions/42433275/ember-arrayproxy-with-init-not-working

MelSumner (approved):
This brings helpful clarity, thank you!

frank06:
Thank you @jenweber and @MelSumner

PR #16576, opened 2018-04-23
adds tests and small docs for function key support in `uniqBy`

bekzod:
support for function `key` in `uniqBy` is nice side-effect of previous refactoring

### DIFF
--- a/packages/ember-runtime/lib/utils.js
+++ b/packages/ember-runtime/lib/utils.js
@@ -177,10 +177,10 @@ export function uniqBy(array, key = identityFunction) {
 
   let ret = A();
   let seen = new Set();
-  let getter = typeof key === 'function' ? key : get;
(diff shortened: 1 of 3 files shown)


packages/ember-runtime/lib/utils.js
```diff
@@ -177,10 +177,10 @@ export function uniqBy(array, key = identityFunction) {
 
   let ret = A();
   let seen = new Set();
-  let getter = typeof key === 'function' ? key : get;
+  let getter = typeof key === 'function' ? key : item => get(item, key);
 
   array.forEach(item => {
-    let val = getter(item, key);
```
bekzod: also fixing issue here, when `key` is function it would be passed as argument to itself

rwjblue (approved):

rwjblue:
Thank you!

PR #16578, opened 2018-04-23
fix conditions in `lib/name`

bekzod:


packages/ember-metal/lib/namespace_search.js
```diff
@@ -169,6 +169,7 @@ function tryIsNamespace(lookup, prop) {
     );
   } catch (e) {
     // continue
+    return false;
```
bekzod: so `tryIsNamespace` returns only `boolean`

rwjblue: Hmm, seems like the return inside the `try` block returns an object, not boolean.  Am I misunderstanding something?

bekzod: my bad sorry

krisselden (approved):

rwjblue:
Thank you!

PR #16587, opened 2018-04-26
Fix issue with `Ember.trySet` on destroying objects

boris-petrov:
trySet is supposed to not blow up when used on destroyed/destroying objects. This commit makes sure it doesn't blow up one both - as the 'destroying' property on destroyed objects also has the value "true".

### DIFF
--- a/packages/ember-metal/lib/property_set.js
+++ b/packages/ember-metal/lib/property_set.js
@@ -49,9 +49,9 @@ export function set(obj, keyName, value, tolerant) {
     typeof keyName !== 'string' || keyName.lastIndexOf('this.', 0) !== 0
   );
 
-  if (obj.isDestroyed) {
+  if (obj.isDestroying) {
     assert(
-      `calling set on destroyed object: ${toString(obj)}.${keyName} = ${toString(value)}`,
+      `calling set on destroyed/destroying object: ${toString(obj)}.${keyName} = ${toString(value)}`,
       tolerant
     );
     return;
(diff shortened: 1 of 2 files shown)


krisselden:
Why do we want to change set to assert() on isDestroying? Right now, marking objects as isDestroying is done synchronous on destroy(), willDestroy() is async and will not fire change events on set to anything that is isDestroying.

What is the motivation for this change?

boris-petrov:
@krisselden - in our app, we used to check for `isDestroyed` everywhere. However, recently we noticed that we had some async code that was executed in a component which had `isDestroying` set to `true` but `isDestroyed` set to `false` and hence our checks were useless (I can try preparing a reproduction if needed). We then decided to change all checks to `isDestroying` as it is a "superset" of `isDestroyed`. Hence this pull-request. Calling `set` on a `isDestroying` component will break and `isDestroying` is always `true` when `isDestroyed` is `true` so I thought it's a win-win.

krisselden:
@boris-petrov the point of this flag is not a superset of isDestroyed, destruction is a mark/sweep operation, you were getting the async when the component was marked for destruction but not destroyed.  Currently, calling set on an object marked for destruction is allowed and will not fire change events if observers are also marked for destruction. This PR is a breaking change.

I would take a look at `ember-lifeline` or `ember-concurrency` for a better async pattern than checking `isDestroyed`.  It is better to make your async check a some sort of cancelation token and cancel it where it is appropriate like in component destroy (like the addons I mentioned do), rather than checking various state flags to derive whether your task is still valid. 

Another simple pattern is if your component only has one valid async task, just give each task launch an id and set the current id on the component and noop any async actions on `this` if `taskId !== this.currentTaskId`.

Issue #16586, opened 2018-04-26
Text in template helper 'action' doco makes no sense


shearichard:
I'm pretty sure that some of the text in the doco for the `action` template helper is not correct.

The underlying file is here https://github.com/emberjs/ember.js/blob/v3.0.0/packages/ember-glimmer/lib/helpers/action.ts and it can be seen rendered here https://www.emberjs.com/api/ember/3.0/classes/Ember.Templates.helpers/methods/action?anchor=action .

The text in question reads like this 

> Once that function is read (or if a function was passed), create a closure over that function and any arguments. 

I'm almost certain that the bit in the bracket, "or if a function was passed", is not complete as it makes no sense grammatically putting aside whether it makes any sense in the context of Ember.

Unfortunately I can't say what should be in those brackets otherwise I would do a PR.

**EDIT**: changed "make sense grammatically" to "makes no sense grammatically"

shearichard:
Having just re-read the paragraph in which that text appears I think I do understand what should be in those brackets but it would be helpful to get the views of others to see if I'm on the right track.

> In these contexts, the helper is called a "closure action" helper. Its behavior is simple: If passed a function name, read that function off the actions property of the current context. Once that function is read (or _immediately_ if a function was passed), create a closure over that function and any arguments. The resulting value of an action helper used this way is simply a function.

Would be interested in feedback on my suggestion.

Serabe:
@shearichard thank you! Please, open a PR so people can comment directly on the proposal and, if it is ok, approve it and merge it directly.

Thank you!

locks:
Fix is merged :)

PR #16589, opened 2018-04-28
Minor change to wording of doco for 'action' template helper

shearichard:
A proposed wording change to the 'action' template helper documentation.

This is with reference to #16586 .

I'm happy to be told otherwise by those more knowledgeable but I believe the current text just doesn't make sense , possibly due to an oversight when originally written.

I have suggested an additional word and minor change to punctuation which I believe will make the paragraph more useful .


rwjblue (approved):
Awesome, thank you!!

PR #16590, opened 2018-04-30
[DOC release] Resolve #15856 re interaction of `set` with `setUnknownProperty`

### LINKED ISSUE #15856: setUnknownProperty is not used as documented when return value is undefined
The "Unknown Properties" section of the [`set` documentation](https://emberjs.com/api/ember/2.15/classes/Ember.Object/methods?anchor=set) for 2.15 and at least a few earlier versions says:
> If `setUnknownProperty()` returns undefined, then `set()` will simply set the value on the object.

A [similar section for 2.16](https://emberjs.com/api/ember/2.16/classes/@ember%2Fobject/methods/set?anchor=set) says of `set`:
> Sets the value of a property on an object ... If the property is not defined but the object implements the setUnknownProperty method then that will be invoked as well.

However, when `set` invokes an object's `setUnknownProperty` method, nothing is done with the return value. In particular the property is not set on the object when the function returns undefined. This can be easily seen in [`property_set.js`](https://github.com/emberjs/ember.js/blob/v2.16.1/packages/ember-metal/lib/property_set.js#L59), where `setUnknownProperty` is called without assigning its return value to anything, and nothing else happens after the function call except returning from `set`. So rather than being invoked *as well*, `setUnknownProperty` is invoked *instead* of setting the property value on the object.

Is this an error in the docs or in my understanding of them, or is it a bug in the intended operation of the code? I'm happy to take a stab at a PR if it's the latter and there's interest.

mikeu:
Corrects the description of the usage of `setUnknownProperty` by `set` to better reflect the actual behaviour of the code, as discussed in #15856. Specifically, it no longer states that `set` will set the given property on the object when it calls `setUnknownProperty`, since this call is made _instead_ of the default behaviour, rather than _as well as_.

### DIFF
--- a/packages/ember-metal/lib/property_set.js
+++ b/packages/ember-metal/lib/property_set.js
@@ -13,9 +13,11 @@ import { isDescriptor, peekMeta, descriptorFor } from './meta';
 */
 /**
   Sets the value of a property on an object, respecting computed properties
-  and notifying observers and other listeners of the change. If the
-  property is not defined but the object implements the `setUnknownProperty`
-  method then that will be invoked as well.
+  and notifying observers and other listeners of the change.
+  If the specified property is not defined on the object and the object
+  implements the `setUnknownProperty` method, then instead of setting the
+  value of the property on the object, its `setUnknownProperty` handler
+  will be invoked with the two parameters `keyName` and `value`.
 
   ```javascript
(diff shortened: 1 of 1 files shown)


rwjblue (approved):

rwjblue:
Thank you!!

Issue #16427, opened 2018-03-27
[3.1.0-beta.5] notifyPropertyChange() fails for plain objects


kanongil:
This unit test will fail on the latest beta:

```js
test('notifies computed properties on plain objects', function (assert) {
  let obj = {
    a: 45,
    b: computed(function () {
      return ~~(this.a / 10);
    })
  };

  assert.equal(get(obj, 'b'), 4);

  obj.a = 10;
  Ember.notifyPropertyChange(obj, 'b');

  assert.equal(get(obj, 'b'), 1);      // fails - returned value is still 4
});
```

It works, as intended, if I do the same on an ember object.

mmun:
I assume you mean `computed('a', function(){ ... })`?

kanongil:
No, the code is as intended. Obviously, this is not what failed for me, but a reduced test case to expose the issue.

mmun:
This is working as intended then. Since `b` has no dependency on `a`, notifying `a` will not cause `b` to be invalidated.

kanongil:
Ah, so I did make an error. Yes `Ember.notifyPropertyChange(obj, 'b');` is the proper failing case.

Sorry about that. I have update the original post.

kanongil:
A bit of further investigation reveals that this usage is [scheduled for deprecation](https://www.emberjs.com/deprecations/v3.x/#toc_ember-meta-descriptor-on-object), that it also fails on canary, and that it works correctly if I use `Ember.defineProperty(obj, 'b', …)` to define the computed value.

mmun:
Yeah, but we still need to make it work for backwards compatibility. The fix should be relatively easy but I haven't had time to work on it.

PR #16602, opened 2018-05-02
fix issue `notifyPropertyChange` not working for plain obj

bekzod:
possible fix for #16427

### DIFF
--- a/packages/ember-metal/lib/meta.js
+++ b/packages/ember-metal/lib/meta.js
@@ -693,7 +693,17 @@ export function descriptorFor(obj, keyName, _meta) {
   let meta = _meta === undefined ? peekMeta(obj) : _meta;
 
   if (meta !== undefined) {
-    return meta.peekDescriptors(keyName);
+    let descriptor = meta.peekDescriptors(keyName);
+
+    if (descriptor === undefined && keyName in obj) {
+      // #GH16427 in case defined without `defineProperty`
+      let possibleDesc = obj[keyName];
+      if (isDescriptor(possibleDesc) && 'didChange' in possibleDesc) {
+        descriptor = possibleDesc;
+      }
+    }
+
+    return descriptor;
   }
 }
 
--- a/packages/ember-metal/lib/property_events.js
+++ b/packages/ember-metal/lib/property_events.js
@@ -86,8 +86,7 @@ function notifyPropertyChange(obj, keyName, _meta) {
 
   let possibleDesc = descriptorFor(obj, keyName, meta);
 
-  // shouldn't this mean that we're watching this key?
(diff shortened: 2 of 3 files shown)


rwjblue (commented):
IMHO, the fix needs to be in `descriptorFor` itself otherwise we will continue to find other APIs that don’t work correctly.

packages/ember-metal/lib/property_events.js
```diff
@@ -86,8 +86,7 @@ function notifyPropertyChange(obj, keyName, _meta) {
 
   let possibleDesc = descriptorFor(obj, keyName, meta);
 
-  // shouldn't this mean that we're watching this key?
-  if (possibleDesc !== undefined && possibleDesc.didChange) {
+  if (possibleDesc !== undefined && 'didChange' in possibleDesc) {
```
krisselden: this change seems unrelated and I'm curious why?

packages/ember-metal/lib/meta.js
```diff
@@ -693,7 +693,17 @@ export function descriptorFor(obj, keyName, _meta) {
   let meta = _meta === undefined ? peekMeta(obj) : _meta;
 
   if (meta !== undefined) {
-    return meta.peekDescriptors(keyName);
+    let descriptor = meta.peekDescriptors(keyName);
+
+    if (descriptor === undefined && keyName in obj) {
+      // #GH16427 in case defined without `defineProperty`
+      let possibleDesc = obj[keyName];
+      if (isDescriptor(possibleDesc) && 'didChange' in possibleDesc) {
```
krisselden: 'didChange' in possibleDesc doesn't belong in descriptorFor

Also we would need to move this deprecation and fallback code and make sure we don't have the fallback code elsewhere.
https://github.com/emberjs/ember.js/blob/470dddb9bbea32f0bf71219740b22c8dc8ecf68d/packages/ember-metal/lib/property_get.js#L117

krisselden (changes requested):

packages/ember-metal/lib/property_events.js
bekzod: just to avoid `boolean coercion`

krisselden: Last time I checked `in` was worse than boolean coercion in v8 (cc @bmeurer ) it is better to do possibleDesc.didChange !== undefined

bmeurer: I'd go with `possibleDesc.didChange !== undefined`. The `in` operator is not backed by ICs so it only pays off if the objects have vastly different shapes all the time.

krisselden: Thanks for the quick response

krisselden:
superseded by #16653

PR #16603, opened 2018-05-02
[BUGFIX beta] support mouseEnter/Leave events w/o jQuery

### LINKED ISSUE #16591: mouseEnter/Leave event delegation broken w/o jQuery
For components that have a `mouseEnter()` (and/or `mouseLeave()`) event handler, these are not called anymore when running without jQuery. 

I think this is because those events don't bubble, so event delegation based on the root element, which is what the  `EventDispatcher` is doing, does not work. Those events get only triggered for the element having the listener (the app's `rootElement` in this case). See https://developer.mozilla.org/en-US/docs/Web/Events/mouseenter

A possible workaround would be to use `mouseover` under the hood, to enable event delegation, and to make sure it behaves the same as the `mouseEnter` (i.e. doesn't trigger again for child elements). jQuery seems to do the same, that's why it works in jQuery mode: https://github.com/jquery/jquery/blob/899c56f6ada26821e8af12d9f35fa039100e838e/src/event.js#L666-L700

I could work on a PR, if we agree this ^ is the way forward!?

simonihmig:
As these events don't bubble, the `EventDispatcher`'s event delegation approach does not work here, when not using jQuery. jQuery has special handling of these events, by listening to `mouseover`/`mouseout` instead and dispatching fake `mouseenter`/`mouseleave` events. This adds similar handling to `EventDispatcher`'s native mode for these events.

Fixes #16591

### DIFF
--- a/packages/ember-glimmer/tests/integration/event-dispatcher-test.js
+++ b/packages/ember-glimmer/tests/integration/event-dispatcher-test.js
@@ -146,6 +146,56 @@ moduleFor(
 
       this.$('#is-done').trigger('click');
     }
+
+    ['@test delegated event listeners work for mouseEnter/Leave'](assert) {
+      let receivedEnterEvents = [];
+      let receivedLeaveEvents = [];
+
+      this.registerComponent('x-foo', {
+        ComponentClass: Component.extend({
+          mouseEnter(event) {
+            receivedEnterEvents.push(event);
+          },
+          mouseLeave(event) {
+            receivedLeaveEvents.push(event);
+          },
+        }),
+        template: `<div id="inner"></div>`,
+      });
+
+      this.render(`{{x-foo id="outer"}}`);
+
+      let parent = this.element;
+      let outer = this.$('#outer')[0];
+
+      // mouse moves over #outer
+      this.runTask(() => {
+        this.$('#outer').trigger('mouseenter', { canBubble: false, relatedTarget: parent });
+        this.$('#outer').trigger('mouseover', { relatedTarget: parent });
+      });
+      assert.equal(receivedEnterEvents.length, 1, 'mouseenter event was triggered');
+      assert.strictEqual(receivedEnterEvents[0].target, outer);
+
+      // mouse moves over #inner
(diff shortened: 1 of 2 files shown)


packages/ember-glimmer/tests/integration/event-dispatcher-test.js
```diff
@@ -143,6 +143,57 @@ moduleFor(
...
+          },
+        }),
+        template: `<div id="inner"></div>`,
+      });
+
+      this.render(`{{x-foo id="outer"}}`);
+
+      let parent = this.element;
+      let outer = this.$('#outer')[0];
+      let inner = this.$('#inner')[0];
```
btecu: `inner` is not used

simonihmig:
I rebased this today (had conflicts with my own jQuery.Event deprecation changes 😝). This *should* be green, but I always get some failing test runs due to yarn network timeouts. Already force pushed a few times to rerun the tests...

@rwjblue @krisselden would love a review! :)

rwjblue:
restarted those failed builds, will try to do a proper review today. Sorry for the delays...

simonihmig:
No problem, at least it's green now! :)

hakubo:
will this land in 3.4?

simonihmig:
@rwjblue ping ;)

hakubo:
@rwjblue can I help with this?

rwjblue (approved):

packages/ember-views/lib/system/event_dispatcher.js
```diff
@@ -290,27 +295,86 @@ export default EmberObject.extend({
...
+          let target = event.target;
+
+          do {
+            if (viewRegistry[target.id]) {
+              if (viewHandler(target, event) === false) {
+                event.preventDefault();
+                event.stopPropagation();
+                break;
+              }
+            } else if (target.hasAttribute('data-ember-action')) {
```
ilucin: `target` can be a text node which does not support hasAttribute method -> http://take.ms/vCrm8C

PR #16613, opened 2018-05-04
[BUGFIX release] Prevent errors in ember-engines + 3.1 + proxies.

### LINKED ISSUE #16610: "You attempted to access the `_onLookup`" error in Engine
I have created a [demo app](https://github.com/amk221/-ember-engine-get-issue/commit/353c0fd44401218d4f7271fe051a562390474f60) with a simple use of `unknownProperty`.

Using `{{myService.myUnknownProperty}}` works in the parent app, but not in the Engine.

Instead, I get the recently added [error](
https://github.com/emberjs/ember.js/blob/eca00feb5b4d83a5bed2bf7924b839f58ff6e7cc/packages/ember-runtime/lib/system/core_object.js#L81) that informs the user about the changes regarding [`.get`](https://www.emberjs.com/blog/2018/04/13/ember-3-1-released.html#toc_es5-getters-for-computed-properties-2-of-4)

(Sorry if this is in the wrong repo, the error is coming from ember.js repo not the Engine repo so thought I'd raise here)

rwjblue:
Setup:

* In Ember 3.1 any service that has an `unknownProperty` method will trigger an assertion if properties are accessed directly and the `unknownProperty` implementation does not return `undefined` (which is commonly true for things like "features" service which always returns true or false).

* Ember Engines registers any service instances from the host app into the engines container (with `instantiate: false` set). This is the mechanism by which `engineDependencies` works.

* When a service is looked up (regardless of `instantiate: false`) the container looks for and calls the `_onLookup` method so that the class can validate all of its injections.

Combining these two pieces means that any engines with an `engineDependencies` provided service that has an `unknownProperty` will throw an error when the service is looked up.

Fixes #16610

### DIFF
--- a/packages/ember-runtime/lib/system/core_object.js
+++ b/packages/ember-runtime/lib/system/core_object.js
@@ -109,6 +109,7 @@ function makeCtor(base) {
                 property === 'didAddListener' ||
                 property === 'didRemoveListener' ||
                 property === 'isDescriptor' ||
+                property === '_onLookup' ||
                 property in target
               ) {
                 return Reflect.get(target, property, receiver);
--- a/packages/ember-runtime/tests/system/core_object_test.js
+++ b/packages/ember-runtime/tests/system/core_object_test.js
@@ -1,7 +1,7 @@
 import { getOwner, setOwner } from 'ember-owner';
 import { get } from 'ember-metal';
 import CoreObject from '../../lib/system/core_object';
-import { moduleFor, AbstractTestCase } from 'internal-test-helpers';
+import { moduleFor, AbstractTestCase, buildOwner } from 'internal-test-helpers';
 
 moduleFor(
   'Ember.CoreObject',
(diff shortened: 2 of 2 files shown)


packages/ember-runtime/lib/system/core_object.js
```diff
@@ -109,6 +109,7 @@ function makeCtor(base) {
                 property === 'didAddListener' ||
                 property === 'didRemoveListener' ||
                 property === 'isDescriptor' ||
+                property === '_onLookup' ||
```
krisselden: I thought we were just going to exclude type 'function' in addition to undefined

rwjblue: We did that, but as you can see in this case the `unknownProperty` _always_ returns non-null non-undefined.

krisselden: doesn't it return a function?

rwjblue: Not on the instance which is what this test is doing. The test replicates how ember-engines passes down services from the host app into the engines own container. Roughly akin to:


    instance = hostAppOwner.lookup(name);
    engineOwner.register(name, instance, { instantiate: false })

`container.lookup` uses `factoryFor` first (so it can later call .create()) and that does the `_onLookup` probe on the result (an instance in this case).

toranb:
@rwjblue It doesn't appear this was ever merged into ember v3.1X (just ember v3.2+)

https://github.com/emberjs/ember.js/blob/coreobject-3-1/packages/ember-runtime/lib/system/core_object.js#L120

If you open the [changelog](https://github.com/emberjs/ember.js/blob/master/CHANGELOG.md) and search for this `[BUGFIX] Prevent errors in ember-engines + 3.1 + proxies.` you don't see it as part of anything v3.1

jamescdavis:
☝️

rwjblue:
:( sorry y’all. At this point (3.4 is latest) we probably won’t ship an update to 3.1 for this though...

PR #16632, opened 2018-05-13
[BUGFIX beta] computed.sort array should update if sort properties array is empty

miguelcobain:
Basically, when the sort properties array is empty, `computed.sort` was returning the original array, **but not** responding to any changes in that array.
The result was that any `computed.sort` never changed, even when the dependent array changed.

I hope the (previously failing) test is self explanatory.
Is there a chance we can backport this to previous ember versions?

Please let me know if anything else is needed from me. Thanks!


miguelcobain:
Failure is related to a Browserstack issue.

rwjblue (changes requested):

miguelcobain:
@rwjblue feedback was applied 👍

rwjblue (approved):

rwjblue:
Thank you!

PR #16643, opened 2018-05-16
[BUGFIX] Ensure that paths which start with // are properly persisted.

nathanhammond:
Given a URL like:
`http://www.example.com/<COUNTRY>/some/path`

There can exist unfortunate server code which can result in the (now-required-to-be-valid) URL:
`http://www.example.com//some/path`

After running through Ember's router we eventually end up at `HistoryLocation#pushState` which calculates the path: `'//some/path'`. So far so good.

This then delegates to `window.history.pushState(state, null, path)`. In the browser API `path` is allowed to be a relative URL. And now we have a problem:

> Uncaught DOMException: Failed to execute 'pushState' on 'History': A history state object with URL 'https://some/path' cannot be created in a document with origin 'https://www.example.com' and URL 'https://www.example.com//some/path'.

Given that all paths at this point are guaranteed to be fully-root-relative path names, we have an out: include the origin. I've elected to fix it one layer higher than `pushState` and `replaceState` to make the change just work for users such as:
https://github.com/dollarshaveclub/ember-router-scroll/blob/master/addon/locations/router-scroll.js
https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/addon/locations/none.js

I'm opening this for discussion before finishing the task. Should we handle this _before calling_  **-or-** _inside of_ `HistoryLocation#pushState` and `HistoryLocation#replaceState`?

### DIFF
--- a/packages/ember-routing/index.d.ts
+++ b/packages/ember-routing/index.d.ts
@@ -37,7 +37,7 @@ export const HistoryLocation: {
   replaceURL(path: string): void;
   getState(): any;
   pushState(path: string): void;
-  replaceState(paht: string): void;
+  replaceState(path: string): void;
   onUpdateURL(callback: ()=>any): void;
   formatURL(url: string): string;
   willDestroy(): void;
--- a/packages/ember-routing/lib/location/history_location.js
+++ b/packages/ember-routing/lib/location/history_location.js
@@ -82,6 +82,9 @@ export default EmberObject.extend({
 
     let state = this.getState();
     let path = this.formatURL(this.getURL());
+
+    path = `${window.location.origin}${path}}`;
+
     if (state && state.path === path) {
       // preserve existing state
       // used for webkit workaround, since there will be no initial popstate event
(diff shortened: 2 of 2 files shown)


rwjblue (commented):
I'm not sure this is guaranteed to be correct in all cases. It doesn't seem that `window.location.origin` includes all parts of the current URL (e.g. `username` or `password` portions).  [docs here](https://html.spec.whatwg.org/multipage/links.html#dom-hyperlink-origin)

I don't know that we intend to support this use case 🤔...

rwjblue:
It seems appropriate to me to have apps implement their own location API implementation to handle these application specific workarounds. What do you think?

nathanhammond:
As for workarounds, I've already got this:

```js
export default HistoryLocation.extend({
  history: {
    state: computed(() => {
      return window.history.state;
    }).volatile(),

    pushState(state, title, path) {
      window.history.pushState(state, title, `${window.location.origin}${path}`);
    },

    replaceState(state, title, path) {
      window.history.replaceState(state, title, `${window.location.origin}${path}`);
    }
  }
})
```

PR #16648, opened 2018-05-17
update ember-cli sha

GavinJoyce:
part of https://github.com/ember-cli/ember-cli/issues/7530

updating to https://github.com/ember-cli/ember-cli/commit/5bad602b6fa9c3baf9041394148fda8ef7f8bf6f

This will allow us to add test for blueprints run in MU in-repro addons

TODO:

 * [ ] fix failing tests


rwjblue:
Looks like the failing tests are due to not opting in to the MU feature (a change in Ember-cli made the experiments opt in via environment variables instead of enabled automatically).

PR #16653, opened 2018-05-17
[BUGFIX release] Fix descriptorFor to support fallback

### LINKED ISSUE #16427: [3.1.0-beta.5] notifyPropertyChange() fails for plain objects

krisselden:
Ensure all fallback cases are deprecated.

Fixes #16427 and add test for `notifyPropertyChange` not working for plain obj.

### DIFF
--- a/packages/ember-metal/lib/property_set.js
+++ b/packages/ember-metal/lib/property_set.js
@@ -1,12 +1,12 @@
 import { HAS_NATIVE_PROXY, toString } from 'ember-utils';
 import EmberError from '@ember/error';
 import { DEBUG } from '@glimmer/env';
-import { assert } from '@ember/debug';
+import { assert, deprecate } from '@ember/debug';
 import { getPossibleMandatoryProxyValue, _getPath as getPath } from './property_get';
 import { notifyPropertyChange } from './property_events';
 
 import { isPath } from './path_cache';
-import { isDescriptor, peekMeta, descriptorFor } from 'ember-meta';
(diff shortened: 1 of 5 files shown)


packages/ember-metal/lib/property_set.js
```diff
@@ -78,10 +78,39 @@ export function set(obj, keyName, value, tolerant) {
...
+      },
+    });
+
+    meta(obj).writeDescriptors(keyName, currentValue);
+
+    if (typeof currentValue.setup === 'function') {
+      currentValue.setup(obj, keyName);
+    }
+
+    return currentValue.get(obj, keyName);
```
rwjblue: Don’t we want to call `currentValue.set(obj, keyName, value)`? Why would we call `.get`?

rwjblue (commented):
We should add tests for the deprecated things. Adding a CP as a property on a POJO or something, then calling get, confirm the deprecation, confirm that we can updated a dep key and have the CP recompute, etc (same for set first).

rwjblue (approved):

PR #16657, opened 2018-05-18
Add eslint-plugin-import to validate imports.

rwjblue:
This ensures that all imports are resolvable and valid.

Future changes should enable more rules from eslint-plugin-import (such as ensuring that a given imported module is only specified once or preventing module cycles).

### DIFF
--- a/packages/ember/index.js
+++ b/packages/ember/index.js
@@ -106,6 +106,7 @@ import {
   TextArea,
   isSerializationFirstNode,
 } from 'ember-glimmer';
+// eslint-disable-next-line import/no-unresolved
 import VERSION from './version';
 import * as views from 'ember-views';
 import * as routing from 'ember-routing';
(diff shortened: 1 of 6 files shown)


packages/ember/index.js
```diff
@@ -106,6 +106,7 @@ import {
   TextArea,
   isSerializationFirstNode,
 } from 'ember-glimmer';
+// eslint-disable-next-line import/no-unresolved
 import VERSION from './version';
```
krisselden: We could always have a placeholder file for this and clobber it with the generated one

rwjblue: Yep, I am planning to add a file for the new "public modules api" for it instead though. I just wanted to land this first, as it massively helps to have lint failures on invalid module imports (as opposed to running the tests and hoping that there are failures...).

PR #16658, opened 2018-05-19
Convert ember-metal/transaction to typescript

csantero:
Also upgrades broccoli-typescript-compiler so as to pull in typescript 2.8, which I needed for the conditional types in `runInTransaction`.

### DIFF
--- a/packages/ember-metal/index.d.ts
+++ b/packages/ember-metal/index.d.ts
@@ -6,7 +6,17 @@ export const PROPERTY_DID_CHANGE: symbol;
 
 export function setHasViews(fn: () => boolean): null;
 
-export function runInTransaction(context: any, methodName: string): any;
+export type MethodKey<T> = { [K in keyof T]: T[K] extends (() => void) ? K : never }[keyof T];
+export type RunInTransactionFunc = <T extends object, K extends MethodKey<T>>(
+  context: T,
+  methodName: K
+) => boolean;
+export type DidRenderFunc = (object: any, key: string, reference: any) => void;
+export type AssertNotRenderedFunc = (obj: object, keyName: string) => void;
+
+export const runInTransaction: RunInTransactionFunc;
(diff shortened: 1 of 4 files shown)


rwjblue (commented):
Will need to bump eslint-typescript-parser to ^15.0.0 to stay compatible with typescript@2.8.

packages/ember-metal/index.d.ts
```diff
@@ -6,7 +6,17 @@ export const PROPERTY_DID_CHANGE: symbol;
...
-export function runInTransaction(context: any, methodName: string): any;
+export type MethodKey<T> = { [K in keyof T]: T[K] extends (() => void) ? K : never }[keyof T];
+export type RunInTransactionFunc = <T extends object, K extends MethodKey<T>>(
+  context: T,
+  methodName: K
+) => boolean;
+export type DidRenderFunc = (object: any, key: string, reference: any) => void;
+export type AssertNotRenderedFunc = (obj: object, keyName: string) => void;
+
+export const runInTransaction: RunInTransactionFunc;
```
rwjblue: I don't grok why we need this (along with the other lines above), I would have thought you could replace the preexisting `export function runInTransaction` line with `export { runInTransaction } from './lib/transaction';` and all of these types in this `index.d.ts` wouldn't need to be redefined. What am I missing?

csantero: @rwjblue You're right. I was over-thinking it. Fixed in latest commit.

csantero:
> Will need to bump eslint-typescript-parser to ^15.0.0 to stay compatible with typescript@2.8.

Updated in latest commit

csantero:
> Error: Navigation Timeout Exceeded: 30000ms exceeded

Travis seems to be timing out on one test suite.

rwjblue:
restarted

PR #16660, opened 2018-05-19
convert ember-template-compiler to typescript

csantero:
I wasn't sure what to do with the `ComponentNode` plugins. There doesn't seem to be any corresponding AST node type in `@glimmer/syntax`, so I just worked around it with `any`.

There are some TS errors originating from the `@glimmer/compiler` package. Not clear exactly what's going on there.


csantero:
I believe these TS errors are being caused due to ember.js using a higher level of strictness than `@glimmer/compiler`. Ember uses `--strict` while glimmer-vm does not. This PR would add the first TypeScript code that references `@glimmer/compiler`, and that triggers the TS compiler to look for errors in the node module. I'm not sure if there's any workaround for this beyond updating glimmer-vm to tighten its strictness settings (and fixing all the ensuing errors in that repo).

csantero:
After updating to glimmer 0.34.6, this is working now. Travis failure seems to be a transient browserstack error.

rwjblue:
Restarted

rwjblue:
Thank you @csantero !

PR #16661, opened 2018-05-20
[DOC release] Change name for Enumerable to comply with RFC 176

toddjordan:
Addresses https://github.com/ember-learn/ember-api-docs/issues/513


rwjblue:
Thank you @toddjordan!

PR #16662, opened 2018-05-20
[BUGFIX canary] fix conditional

bekzod:


rwjblue:
Can you check which versions this bug affects and update the commit message and PR title to include `[BUGFIX beta']` or `[BUGFIX release]` if either of those are affected?

bekzod:
introduced here 77e033e17076df0d096b44ce273436254fa138c2 so it affects only canary

rwjblue (approved):

PR #16669, opened 2018-05-21
Start enabling svelte builds

krisselden:


packages/@ember/deprecated-features/index.ts
```diff
@@ -0,0 +1 @@
+export const DESCRIPTOR_VALUES = '3.2.0';
```
rwjblue: Perhaps we could go with `DESCRIPTOR_FALLBACK_PROPERTY_LOOKUP` or `PROPERTY_BASED_DESCRIPTORS`?

rwjblue (commented):
Minor quibble RE: deprecation name, but I don’t feel very strongly. Happy to have this land as is...

rwjblue:
Released an updated version of babel-plugin-debug-macros and submitted https://github.com/emberjs/ember.js/pull/16673 to update here (so we can test things out).

PR #16670, opened 2018-05-22
[DOC release] Update class_names_support.js to include binding classes to false properties

qpowell:
This PR adds documentation for using `classNameBindings` to bind a class when a property is false. 

This is a feature that I use pretty frequently, and it's been a part of the guides since `1.10.0` from what I can tell (see: [Ember.js - Components: Customizing a Component's Element](https://guides.emberjs.com/v1.10.0/components/customizing-a-components-element/)), but it appears to be missing from the API docs.


locks (approved):
Thank you!

PR #16672, opened 2018-05-22
correct `isProxy` and reuse `isObject`

bekzod:


packages/ember-utils/lib/is_proxy.ts
```diff
@@ -4,9 +4,11 @@ import WeakSet from './weak_set';
 const PROXIES = new WeakSet();
 
 export function isProxy(object: any | undefined | null) {
-  return isObject(object) && PROXIES.has(object);
+  return PROXIES.has(object);
 }
 
 export function setProxy(object: object) {
-  return PROXIES.add(object);
```
bekzod: remove `return` since it was returning `PROXIES`

packages/ember-utils/lib/is_proxy.ts
```diff
@@ -4,9 +4,11 @@ import WeakSet from './weak_set';
 const PROXIES = new WeakSet();
 
 export function isProxy(object: any | undefined | null) {
-  return isObject(object) && PROXIES.has(object);
```
bekzod: `isObject(object)` should have been in `setProxy`

krisselden (approved):

PR #16675, opened 2018-05-23
cleanup chains

bekzod:


packages/ember-metal/lib/chains.ts
```diff
@@ -191,21 +190,13 @@ class ChainNode {
     this._paths = undefined;
 
     // It is false for the root of a chain (because we have no parent)
-    let isWatching = parent !== null;
-    let object;
-    if (isWatching) {
+    if ((this._isWatching = parent !== null)) {
```
locks: I'm not a fan of this pattern, it kinda seems like a mistake if you're reading with full attention.

packages/ember-metal/lib/chains.ts
```diff
@@ -266,16 +258,14 @@ class ChainNode {
...
-    if (this._chains === undefined) {
+    let chains = this._chains;
+    if (chains === undefined) {
       chains = this._chains = Object.create(null);
-    } else {
-      chains = this._chains;
     }
 
-    let node = chains[key];
+    let node = chains![key];
```
krisselden: the `!` shouldn't be needed because chains is already checked for undefined and assigned.

bekzod: linter is complaining if I do that

krisselden (changes requested):
the `!` chains shouldn't be needed

packages/ember-metal/lib/chains.ts
```diff
@@ -191,21 +190,14 @@ class ChainNode {
...
 
     // It is false for the root of a chain (because we have no parent)
-    let isWatching = parent !== null;
-    let object;
+    let isWatching = (this._isWatching = parent !== null);
     if (isWatching) {
       let parentValue = parent!.value();
       if (isObject(parentValue)) {
-        object = parentValue;
+        this._object = parentValue;
```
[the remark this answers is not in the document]
bekzod: should I remove `this._object = undefined` then ?

krisselden:
I believe this was superseded by #16680 if not, please rebase and reopen

bekzod:
@krisselden rebased please reopen :P cannot do it myself

packages/ember-metal/lib/chains.ts
```diff
@@ -256,16 +248,14 @@ class ChainNode {
...
-    if (this.chains === undefined) {
+    let chains = this.chains;
+    if (chains === undefined) {
       chains = this.chains = Object.create(null);
-    } else {
-      chains = this.chains;
     }
 
-    let node = chains[key];
+    let node = chains![key];
```
krisselden: I don't understand why the `!` is needed here.

krisselden: The `if (chains === undefined)` and assignment should remove undefined.

krisselden (changes requested):

bekzod:
I wouldn’t add it if linter was not complaining

packages/ember-metal/lib/chains.ts
bekzod: Linter is complaining if I remove it

bekzod:
any more comments on this ?

packages/ember-metal/lib/chains.ts
bekzod: probably bug in tslinter

bekzod:
@rwjblue I think this is good to go 😉

Issue #16517, opened 2018-04-13
[RFC 322] Ember.copy and Ember.Copyable Deprecation


lupestro:
This is a meta-issue to track the implementation of [https://github.com/emberjs/rfcs/pull/322]().

**ember-copy (addon)**
- [x] implement copy method and Copyable mix-in

**data**
- [x] remove use of copy() and Copyable from code and tests

**ember.js**
- [x]  adjust copy() to always copy arrays itself rather than delegating on Copyable
- [x]  remove use of copy() and Copyable from code and tests
- [x]  deprecate copy() and NativeArray.copy()
- [x]  deprecate Copyable() mixin use in object.extend() if feasible

**deprecation-app**
- [ ] document deprecation, Object.assign for shallow copies, and availability of add-in

**ember-cli-eslint**
- [x] ~~deprecation rules for copy() and Copyable~~ [ after discussion, decided not to do it ]

**ember-api**
- [ ] verify copy() and Copyable documentation reflect deprecation

I will link the requisite PRs here as they are opened.

rwjblue:
Wheeeee! Thank you @lupestro!

PR #16677, opened 2018-05-23
deprecate `NativeArray#copy`

bekzod:
as mentioned here: https://github.com/emberjs/ember.js/issues/16517

### DIFF
--- a/packages/ember-runtime/lib/copy.js
+++ b/packages/ember-runtime/lib/copy.js
@@ -1,6 +1,7 @@
 import { assert } from '@ember/debug';
 import EmberObject from './system/object';
 import Copyable from './mixins/copyable';
+
 /**
  @module @ember/object
 */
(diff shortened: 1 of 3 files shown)


packages/ember-runtime/lib/copy.js
```diff
@@ -19,12 +21,12 @@ function _copy(obj, deep, seen, copies) {
...
   assert(
     'Cannot clone an EmberObject that does not implement Copyable',
-    !(obj instanceof EmberObject) || Copyable.detect(obj)
+    !(obj instanceof EmberObject) || Copyable.detect(obj) || NativeArray.detect(obj)
   );
 
   // IMPORTANT: this specific test will detect a native array only. Any other
   // object will need to implement Copyable.
-  if (Array.isArray(obj)) {
+  if (Array.isArray(obj) || NativeArray.detect(obj)) {
```
krisselden: we only apply the NativeArray mixin to arrays, so Array.isArray should be enough.

krisselden (changes requested):

krisselden:
seems to be a crash happening in the IE 11 test in iexplore.exe trying again

krisselden (approved):

PR #16685, opened 2018-05-24
Wrap all existing deprecated features for svelting.

rwjblue:


packages/ember-meta/lib/meta.ts
```diff
@@ -390,67 +391,75 @@ export class Meta {
   }
 
   writeBindings(subkey: string, value: any) {
```
bekzod: maybe conditionally extend meta like done here: https://github.com/emberjs/ember.js/pull/16355

krisselden (approved):
maybe cleaner to do what @bekzod mentioned.

packages/ember-meta/lib/meta.ts
rwjblue: would you mind updating that PR to fix it up? I do agree that is nice

PR #16687, opened 2018-05-25
Avoid using jquery if disabled using flag

cibernox:


rwjblue (changes requested):
Also needs to add the flag to `packages/ember-environment/lib/env.ts` ([like this one](https://github.com/emberjs/ember.js/blob/236dd24f742bc38b4945b5bade9e50c4c519c51a/packages/ember-environment/lib/env.ts#L98))...

cibernox:
@rwjblue Done!

locks (approved):

packages/ember-views/lib/system/jquery.js
```diff
@@ -1,13 +1,14 @@
...
 
 let jQuery;
-export let jQueryDisabled = false;
+export let jQueryDisabled = ENV._JQUERY_INTEGRATION === false;
 
 if (hasDOM) {
   jQuery = context.imports.jQuery;
 
-  if (jQuery) {
+  if (!jQueryDisabled && jQuery) {
```
simonihmig: Not sure how this is all connected, but I see a few modules importing `jQuery` from this module, without checking for `jQueryDisabled`. Before this change, there was no case possible that `jQuery` was available while `jQueryDisabled` is true. This is now possible, with the flag being false but the jQuery global still available. So should we override `jQuery` to `undefined`, if `jQueryDisabled` is true?

cibernox: @simonihmig I think you might be right. Not if the user has explicitly disabled jQuery, `Ember.$` will be undefined, even if jQuery happens to be included in the page in some other way.

rwjblue: Ya, I agree. I think that we need to change all of our guards to be checking `jQueryDisabled` instead of the default export of this module.

tldr; it should absolutely be possible to tell Ember to treat jQuery as disabled _even if_ `window.$` is present...

cibernox: @rwjblue I already did in this PR

packages/ember-views/lib/system/event_dispatcher.js
```diff
@@ -7,11 +7,11 @@ import { assign } from '@ember/polyfills';
 import { assert } from '@ember/debug';
 import { get, set } from 'ember-metal';
 import { Object as EmberObject } from 'ember-runtime';
-import jQuery from './jquery';
+import jQuery, { jQueryDisabled } from './jquery';
 import ActionManager from './action_manager';
 import fallbackViewRegistry from '../compat/fallback-view-registry';
 
-const HAS_JQUERY = jQuery !== undefined;
+const HAS_JQUERY = jQueryDisabled === false;
```
rwjblue: I don't think we should check for `false` here (though I suppose it doesn't matter _too_ much).  In general we can trust that our `jQueryDisabled` export will always be boolean and will encapsulate the proper "default to true, unless false" logic.

packages/ember-views/lib/system/jquery.js
rwjblue: Doh! sorry :( I guess I thought we had more places that did the conditional...

packages/ember-views/lib/system/event_dispatcher.js
rwjblue: Hmm, any reason to not just use `jQueryDisabled` in this file instead of making the second constant (`HAS_JQUERY`)?

rwjblue (changes requested):
Can you also update:

https://github.com/emberjs/ember.js/blob/d97dcf7635718f988680d1d7acaf2be00ee81cf4/packages/@ember/application/tests/application_test.js#L372

https://github.com/emberjs/ember.js/blob/d97dcf7635718f988680d1d7acaf2be00ee81cf4/packages/ember/index.js#L564 

(only setup `Ember.$` if the jQuery is enabled)

rwjblue:
Darn, sorry. My last change suggestion made the reexport tests fail 

https://github.com/emberjs/ember.js/blob/4aed4567b8c6e4cf08e8dfa361406aa9def8bc5c/packages/ember/tests/reexports_test.js#L171

Should become:

```js
JQueryDisabled && ['$', 'ember-views', 'jQuery']
```

Then add a `.filter(Boolean)` to the end of that array. This will make it only check for `Ember.$` export when it should be there. 

Sorry for the run around...

rwjblue (approved):
Looks great to me, thank you!!

cibernox:
I had to rebase because conflicts with the deprecation of `originalEvent`, but it should be green now.

rwjblue (commented):
~~Looks like a logging test is failing?~~ kk

cibernox:
@simonihmig After rebasing over your deprecating a test is failing, but I think I've resolved conflicts correctly. Can you take a look to see if it's evident to your what is going on?

packages/@ember/application/tests/application_test.js
```diff
@@ -369,11 +369,11 @@ moduleFor(
       this.runTask(() => this.createApplication());
 
       assert.equal(messages[1], 'Ember  : ' + VERSION);
-      if (jQuery) {
-        assert.equal(messages[2], 'jQuery : ' + jQuery().jquery);
+      if (jQueryDisabled) {
+        assert.equal(messages[2], 'my-lib : ' + '2.0.0a');
         assert.equal(messages[3], 'my-lib : ' + '2.0.0a');
```
simonihmig: @cibernox this is related to the failing test, and it seems this line has to go into the else-block (which is when jQuery is enabled)! (as you have switched the order of the jQuery/no-jQuery blocks)

cibernox: Good catch, but it wasn't related with the failing test apparently.

simonihmig: I think it did, as the previous failed test was related to the env output, and that seems to be passing now. But a deprecation test is now failing: https://travis-ci.org/emberjs/ember.js/jobs/384219437#L1562 🤔

cibernox: That was the only failure I was seeing before. That was the one I don't see why it is failing because of my changes

simonihmig:
@cibernox Not sure what's going on. I would *guess* it's something about the changes around `jQueryDisabled`, that the tests expect the deprecation but it's not triggered. I am looking at `ENV` directly [here](https://github.com/emberjs/ember.js/blob/master/packages/ember-views/lib/system/jquery_event_deprecation.js#L23), so I could change that in some test. Don't if it's related. I would have to run the tests in the browser through the debugger, to really understand what's happening there. But not able to do so today, but I can have a look on Monday, if you don't beat me to it...

rwjblue:
I just pushed a commit that should fix the remaining issue which was basically that we didn't properly identify the difference between jQuery being explicitly enabled (which should not receive a deprecation) and when it was implicitly enabled (which _should_ receive a deprecation).

simonihmig:
@rwjblue thanks a lot! 👍

PR #16691, opened 2018-05-26
Deprecate Ember.Map, Ember.MapWithDefault and Ember.OrderedSet

### LINKED ISSUE #16678: Deprecate Ember.Map / Ember.MapWithDefault / Ember.OrderedSet
As agreed on in emberjs/rfcs#237, we need to deprecate creating the following objects:

* [`@ember/map`](https://github.com/emberjs/ember.js/blob/master/packages/%40ember/map/index.js) - Deprecation message should suggest to replace usage with native `Map` usage.
* [`@ember/map/with-default`](https://github.com/emberjs/ember.js/blob/master/packages/%40ember/map/with-default.js) - Deprecation should suggest to replace usage with native `Map` usage.
* [`@ember/ordered-set`](https://github.com/emberjs/ember.js/blob/master/packages/%40ember/map/lib/ordered-set.js) - Deprecation should mention adding `@ember/ordered-set` addon.
* Update the [conditional in @ember/ordered-set](https://github.com/emberjs/ember-ordered-set/blob/43ed8b5ce7b16e753503386ed9228b24b9d5cec3/addon/index.js#L6) to ensure it matches the version the removal lands in.

The deprecations should be `until: '3.5.0'`.

Artmann:
Fixes #16678

### DIFF
--- a/packages/@ember/map/index.js
+++ b/packages/@ember/map/index.js
@@ -1,4 +1,4 @@
-import { assert } from '@ember/debug';
+import { assert, deprecate } from '@ember/debug';
 import { guidFor } from 'ember-utils';
 import OrderedSet from './lib/ordered-set';
 import { copyMap } from './lib/utils';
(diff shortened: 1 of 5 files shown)


rwjblue (approved):
Awesome, thank you!

PR #16692, opened 2018-05-26
[RFC 322] Implemented copy/Copyable deprecation

lupestro:
At long last, I finally delivered an implementation of the RFC. Sorry for the delay. 
- Take a close look at the substitution of a shallow copy for a deep one on the three affected tests in decoupled_basic_test. If we have to do a deep copy for the test to be valid, I'll have to research what's needed.
- I explicitly made copy private - hopefully that's okay. Copyable had *always* been private.


rwjblue (approved):
Looks good to me! I’m going to land this and we can work on the svelting stuff in a follow up...

rwjblue:
Thank you!

PR #16693, opened 2018-05-27
Add URL to jQuery.Event deprecation

simonihmig:
See https://github.com/ember-learn/deprecation-app/pull/127

### DIFF
--- a/packages/ember-views/lib/system/jquery_event_deprecation.js
+++ b/packages/ember-views/lib/system/jquery_event_deprecation.js
@@ -33,6 +33,7 @@ export default function addJQueryEventDeprecation(jqEvent) {
             {
               id: 'ember-views.event-dispatcher.jquery-event',
               until: '4.0.0',
+              url: 'https://emberjs.com/deprecations/v3.x#toc_jquery-event',
             }
           );
           return target[name];


packages/ember-views/lib/system/jquery_event_deprecation.js
```diff
@@ -33,6 +33,7 @@ export default function addJQueryEventDeprecation(jqEvent) {
             {
               id: 'ember-views.event-dispatcher.jquery-event',
               until: '4.0.0',
+              url: 'https://emberjs.com/deprecations/v3.x#toc_jquery-event'
```
rwjblue: CI is failing on this line, needs a tailing comma I think.

You could also do `yarn lint:fix` I believe...

rwjblue (changes requested):

packages/ember-views/lib/system/jquery_event_deprecation.js
simonihmig: Oh well, didn't expect this little change to fail. Should probably setup prettier in my IDE I guess... Fixed now! 😬

rwjblue (approved):

PR #16694, opened 2018-05-29
remove redundant backtick in comment

xg-wang:


rwjblue (approved):

rwjblue:
Good catch, thank you!

PR #16696, opened 2018-05-29
[BUGFIX] allow same function in multiple observers and event listeners

### LINKED ISSUE #15246: Unexpected observer running on init.
```js
import Ember from 'ember';

function func1() {
  alert('func1');
}
function func2() {
  alert('func2');
}

  onInit1: Ember.on('init', func1),
  onObserve1: Ember.observer(func1),
  onObserve2: Ember.observer(func2)
});
```

`func1` runs twice but `func2` doesn’t run. Is that expected? I would expect `func1` to only run once.

Without `onObserve1`, `func1` runs once, but adding it to an observer causes it to run twice on init.

(observer keys left out because they didn't affect the test)

Also, if I change it to 

```js
  onInit1: Ember.on('init', function() { return func1.apply(this, arguments); }),
  onObserve1: Ember.observer(function() { return func1.apply(this, arguments); }),
```

it works as expected.

bekzod:
fixes https://github.com/emberjs/ember.js/issues/15246

### DIFF
--- a/packages/@ember/-internals/metal/lib/mixin.ts
+++ b/packages/@ember/-internals/metal/lib/mixin.ts
@@ -250,7 +250,7 @@ function mergeMixins(
   descs: { [key: string]: object },
   values: { [key: string]: object },
   base: { [key: string]: object },
-  keys: string[]
+  keys: Set<string>
 ): void {
   let currentMixin, props, key, concats, mergings;
 
@@ -285,7 +285,7 @@ function mergeMixins(
         if (!props.hasOwnProperty(key)) {
           continue;
         }
-        keys.push(key);
+        keys.add(key);
         addNormalizedProperty(base, key, props[key], meta, descs, values, concats, mergings);
       }
 
@@ -346,7 +346,7 @@ function updateObserversAndListeners(
     method: string
   ) => void
 ) {
-  if (paths) {
+  if (paths !== undefined) {
     for (let i = 0; i < paths.length; i++) {
       updateMethod(obj, paths[i], null, key);
     }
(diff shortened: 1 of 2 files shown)


packages/ember-metal/lib/mixin.ts
```diff
@@ -251,7 +251,7 @@ function mergeMixins(
   descs: { [key: string]: object },
   values: { [key: string]: object },
   base: { [key: string]: object },
-  keys: string[]
+  keys: Set<string>
```
bekzod: changing to `Set` avoids duplicate  keys which manifested itself with following code
```
     let MyMixin = Mixin.create({
        foo: observer('bar', function() {}),
      });

      let Mixin2 = Mixin.create({
        foo: observer('baz', function() {}),
      });

      let obj = mixin({}, MyMixin, Mixin2);
```

bekzod:
needs restart

packages/ember-metal/lib/mixin.ts
```diff
@@ -379,29 +382,28 @@ function applyMixin(obj: { [key: string]: any }, mixins: Mixin[], partial: boole
...
     }
 
     if (desc === undefined && value === undefined) {
-      continue;
+      return;
     }
 
-    if (descriptorFor(obj, key) !== undefined) {
-      replaceObserversAndListeners(obj, key, null, value);
+    if (descriptorFor(obj, key) === undefined) {
```
bekzod: flipped conditional

bekzod:
this is more like feature than bugfix I think

packages/ember-metal/lib/mixin.ts
```diff
@@ -348,14 +350,16 @@ function replaceObserversAndListeners(
...
+  next: Function | null,
+  seen: Set<Function>
 ): void {
   if (typeof prev === 'function') {
     updateObserversAndListeners(obj, key, getObservers(prev), removeObserver);
     updateObserversAndListeners(obj, key, getListeners(prev), removeListener);
   }
 
-  if (typeof next === 'function') {
+  if (typeof next === 'function' && !seen.has(next)) {
```
bekzod: with `seen` avoiding same `function` being passed through `updateObserversAndListeners` since same function might be used multiple times `observer` `event listeners`

rwjblue:
Ya, I’m not sure we actually want to make changes here. 🤔

Generally speaking we want folks to be removing observers, making them more ergonomic seems like it may make folks reach for them even more...

bekzod:
rebased

bekzod:
no longer valid

PR #16699, opened 2018-05-30
pass meta to `defineProperty`

bekzod:


packages/ember-runtime/lib/system/core_object.js
```diff
@@ -196,11 +196,7 @@ function makeCtor(base) {
             } else if (typeof self.setUnknownProperty === 'function' && !(keyName in self)) {
               self.setUnknownProperty(keyName, value);
             } else {
-              if (DEBUG) {
-                defineProperty(self, keyName, null, value); // setup mandatory setter
-              } else {
-                self[keyName] = value;
-              }
```
bekzod: `defineProperty` defaults to same logic if descriptor argument is `null` as above

rwjblue:
It definitely make sense to pass in the objects meta, but I’m not certain that we should change the production path to go through `defineProperty` here...

bekzod:
internally `defineProperty` would lead to same code path as descriptor argument is null, won't complicate things :P so will just pass meta as you said :)

bekzod:
internally `defineProperty` would lead to same code path as descriptor argument is null, won't complicate things :P so will just pass meta as you said :)

rwjblue (approved):

PR #16700, opened 2018-05-31
remove underscore and mark as `private` properties in meta

bekzod:
Is this good idea ?


rwjblue:
Marking them as private is definitely good, however I’d still prefer to keep them underscored. 

Being private only helps internal typescript consumers know not to touch these properties, but when Ember is consumed you would still be able to use them and now they would be without an underscore and seem “public”...

PR #16702, opened 2018-05-31
Cleanup meta

bekzod:


packages/ember-meta/lib/meta.ts
```diff
@@ -208,55 +223,32 @@ export class Meta {
...
           for (let innerKey in innerMap) {
             seen = seen === undefined ? new Set() : seen;
             if (!seen.has(innerKey)) {
               seen.add(innerKey);
-              calls = calls || [];
-              calls.push(innerKey, innerMap[innerKey]);
+              if (innerMap[innerKey] > 0) {
+                calls = calls || [];
+                calls.push(innerKey);
+              }
```
bekzod: pushing only dependencies which have count greater than zero, avoid checking it later downstream here: https://github.com/emberjs/ember.js/pull/16702/files#diff-8b1d0dc68f822472c1b8b7ae6148b656L168 and avoids calling callback function for deps count equal to zero

packages/ember-meta/lib/meta.ts
```diff
@@ -208,55 +223,32 @@ export class Meta {
...
-    let pointer: Meta | undefined = this;
-    while (pointer !== undefined) {
-      let deps = pointer._deps;
-      if (deps !== undefined && deps[subkey] !== undefined) {
-        return true;
-      }
-      pointer = pointer.parent;
-    }
-    return false;
+    let val = this._findInherited2('_deps', subkey);
```
bekzod: here I am reusing `_findInherited2`

packages/ember-meta/lib/meta.ts
```diff
@@ -148,7 +148,7 @@ export class Meta {
     return this[key] || (this[key] = new Set());
   }
 
-  _getInherited(key: string) {
+  _findInherited1(key: string): any | undefined {
```
bekzod: since `_getInherited` and  `_finInherited` are almost same, just `_finInherited` accepts 2 keys and goes 2 keys deep into meta object, so renamed `_getInherited` to `_finInherited1` which goes one level deep and accepts 1 key, and renamed other appropriately 
`_finInherited1` - accepts 1 key goes one level deep
`_finInherited2` - accepts 2 key goes two level deep
`_finInherited3` - accepts 3 key goes three level deep

bekzod:
added inline comments hope helps, goal is just reuse common parts and make things a little bit DRY

packages/ember-meta/lib/meta.ts
```diff
@@ -173,14 +173,29 @@ export class Meta {
...
+
   _hasInInheritedSet(key: string, value: any) {
     let pointer: Meta | undefined = this;
     while (pointer !== undefined) {
       let set = pointer[key];
-      if (set !== undefined) {
-        if (set.has(value)) {
-          return true;
-        }
+      if (set !== undefined && set.has(value)) {
```
bekzod: collapsing `if` statement

packages/ember-meta/lib/meta.ts
bekzod: this actually could be further refactored to have one `_finInherited` accepts list of keys that goes variable length deep, but I think that might cause performance issues but will check once this merged

bekzod:
any comments to this ?

rwjblue (commented):
I'm not sure about this one. The new names seem odd, and having names that don't logically relate to the intent of the methods makes it hard for me to follow the logic.

krisselden (approved):

bekzod:
rebased

packages/ember-meta/lib/meta.ts
locks: poking in case you still want to check for performance issues

PR #16707, opened 2018-06-01
[DOC beta] remove needless prohibition

ghost:
Originally introduced here:
https://github.com/emberjs/ember.js/pull/4403/files
Discussion question: https://discuss.emberjs.com/t/why-should-i-avoid-transitionto-and-full-urls/14870
This prohibition does not seem necessary so we should remove to avoid
teaching overhead.


rwjblue:
Seems good to me, but looks like it needs a rebase...

efx:
@rwjblue will do that now.

rwjblue (approved):

PR #16709, opened 2018-06-01
[BUGFIX beta] avoid ordered set deprecation

savvymas:
This is part of a fix for this bug: https://github.com/emberjs/ember-ordered-set/issues/19 by creating a private path that excludes the deprecation in the OrderedSet class.

### DIFF
--- a/packages/@ember/map/lib/ordered-set.js
+++ b/packages/@ember/map/lib/ordered-set.js
@@ -12,16 +12,17 @@ import { copyNull } from './utils';
   @constructor
   @private
 */
-export default class OrderedSet {
-  constructor() {
-    deprecate('Use of @ember/OrderedSet is deprecated. Please use native `Map` instead', false, {
-      id: 'ember-map-deprecation',
-      until: '3.5.0',
-    });
 
+/**
+ * This is exported so it can be used by the OrderedSet library.
+ * This is private do not use it.
+ @private
+ */
+
+export class __OrderedSet__ {
+  constructor() {
     this.clear();
   }
-
   /**
     @method create
     @static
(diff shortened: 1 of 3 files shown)


rwjblue:
Looking good! I think we will also need to export this on the global (in `packages/ember/index.js`) so that the addon can use `Ember.__OrderedSet__` if present...

packages/@ember/map/lib/ordered-set.js
```diff
@@ -166,3 +164,14 @@ export default class OrderedSet {
     return set;
   }
 }
+
+export default class OrderedSet extends __OrderedSet__ {
+  constructor() {
+    super();
+    deprecate('Use of @ember/OrderedSet is deprecated. Please use native `Map` instead', false, {
```
bekzod: maybe it should say, Please use native `Map` or  ember-ordered-set addon instead ?

bekzod: or "Please use native `Map` or `https://github.com/emberjs/ember-ordered-set` addon instead"

packages/@ember/map/lib/ordered-set.js
```diff
@@ -166,3 +164,14 @@ export default class OrderedSet {
...
+      id: 'ember-map-deprecation',
+      until: '3.5.0',
+    });
+    this.clear();
```
ef4: Bug here -- this needs to stay in the base implementation.

packages/@ember/map/lib/ordered-set.js
ef4: This is a good idea and can be a followup PR

PR #16710, opened 2018-06-02
add deprecation to `NAME_KEY`

bekzod:
since `NAME_KEY` is still widely used https://emberobserver.com/code-search?codeQuery=NAME_KEY

### DIFF
--- a/packages/ember/index.js
+++ b/packages/ember/index.js
@@ -199,7 +199,19 @@ Ember.canInvoke = utils.canInvoke;
 Ember.tryInvoke = utils.tryInvoke;
 Ember.wrap = utils.wrap;
 Ember.uuid = utils.uuid;
-Ember.NAME_KEY = utils.NAME_KEY;
+
+Object.defineProperty(Ember, 'NAME_KEY', {
+  enumerable: false,
+  get() {
+    deprecate('Using `Ember.NAME_KEY` is deprecated, override `.toString` instead', false, {
+      id: 'ember-name-key-usage',
+      until: '3.9.0',
+    });
+
+    return utils.NAME_KEY;
+  },
+});
+
 Ember._Cache = utils.Cache;
 
 // ****@ember/-internals/container****
--- a/packages/ember/tests/reexports_test.js
+++ b/packages/ember/tests/reexports_test.js
@@ -34,6 +34,12 @@ moduleFor(
       }, /EXTEND_PROTOTYPES is deprecated/);
     }
 
+    ['@test Ember.NAME_KEY is deprecated']() {
+      expectDeprecation(() => {
+        Ember.NAME_KEY;
+      }, 'Using `Ember.NAME_KEY` is deprecated, override `.toString` instead');
+    }
+
     '@test Ember.FEATURES is exported'(assert) {
       for (let feature in FEATURES) {
         assert.equal(
@@ -65,7 +71,6 @@ let allExports = [
   ['canInvoke', '@ember/-internals/utils'],
   ['tryInvoke', '@ember/-internals/utils'],
   ['wrap', '@ember/-internals/utils'],
-  ['NAME_KEY', '@ember/-internals/utils'],
 
   // @ember/-internals/container
   ['Registry', '@ember/-internals/container', 'Registry'],


packages/ember-runtime/lib/system/core_object.js
```diff
@@ -608,7 +609,7 @@ CoreObject.PrototypeMixin = Mixin.create({
     let hasToStringExtension = typeof this.toStringExtension === 'function';
     let extension = hasToStringExtension ? `:${this.toStringExtension()}` : '';
 
-    let ret = `<${getName(this) || FACTORY_FOR.get(this) || this.constructor.toString()}:${guidFor(
+    let ret = `<${ this[NAME_KEY] || getName(this) || FACTORY_FOR.get(this) || this.constructor.toString()}:${guidFor(
```
bekzod: reverted old behavior here, as some addon depend on it right now

bekzod:
@krisselden good idea ?

rwjblue:
IMHO, nearly all of these cases would be better implemented via a `.toString` on the class/mixin/etc (which would be backwards compatible). Also, most of these addons use it speculatively, and we ensured they didn't error by continuing to _have_ an `Ember.NAME_KEY` present.  

tldr; unless we are aware of actual breakage, I'd rather not bring back the lookup for `this[NAME_KEY]`...

bekzod:
Got it, should we at least keep deprecation ?

rwjblue:
Yes, that seems good to me, though it will need to tell folks to use `.toString` instead.

krisselden:
The usage of this.NAME_KEY in CoreObject instance method toString is likely never used, and the examples linked to are trying to add it to the constructor for classToString behavior.  I don't remember why we did the NAME_KEY in there, I think it may have been how the factory at one point set name by passing it to create().  I'd like to redo the string stuff around just constructor.name and factory name, remove the namespace stuff.

krisselden:
I see, you want to propose setName be used instead of this pattern, we use constructor to string so if they implement a toString method on the constructor, it will be used instead.

bekzod:
anyway public usage should be deprecated right ? if not then you can close it

packages/ember/index.js
```diff
@@ -201,7 +201,20 @@ Ember.canInvoke = utils.canInvoke;
...
 Ember.uuid = utils.uuid;
-Ember.NAME_KEY = utils.NAME_KEY;
+Ember.setName = utils.setName;
+
+Object.defineProperty(Ember, 'NAME_KEY', {
+  enumerable: false,
+  get() {
+    deprecate('Using `Ember.NAME_KEY` is deprecated, override `.toString` instead', false, {
+      id: 'ember-name-key-usage',
+      until: '4.0.0',
```
rwjblue: Lets say 3.5.0 here, I don't think `NAME_KEY` was ever public API...

rwjblue (changes requested):
One minor tweak, but seems good to me after that...

packages/ember/index.js
bekzod: done

rwjblue (approved):
Looks good, would you mind doing a deprecation-app entry for this?

packages/ember/tests/reexports_test.js
```diff
@@ -34,6 +34,12 @@ moduleFor(
       }, /EXTEND_PROTOTYPES is deprecated/);
     }
 
+    ['@test Ember.NAME_KEY is deprecated']() {
+      expectDeprecation(() => {
+        Ember.NAME_KEY;
+      }, 'Using `Ember.NAME_KEY` is deprecated, override `.toString` instead');
```
krisselden: should make it clear that it is static toString.   MyClass.toString = () => 'MyClass'

packages/ember/index.js
```diff
@@ -202,7 +202,20 @@ Ember.canInvoke = utils.canInvoke;
 Ember.tryInvoke = utils.tryInvoke;
 Ember.wrap = utils.wrap;
 Ember.uuid = utils.uuid;
-Ember.NAME_KEY = utils.NAME_KEY;
+Ember.setName = utils.setName;
```
krisselden: if we prefer toString on the constructor, why are we exporting this?

rwjblue: Derp. I missed this when I last reviewed. Good catch @krisselden.

krisselden: remove this and it is good to go

krisselden (dismissed):

Turbo87 (approved):
rebased, and adjusted the deprecation warning to stay until v3.9.0

/cc @rwjblue

PR #16715, opened 2018-06-04
Various angle bracket invocation fixes...

rwjblue:
* Bump glimmer-vm to 0.35.2
* Fix issue with implicit `this` on angle invocation of paths
* Move dasherization of angle invocations to compile time


rwjblue:
@chancancode mentioned in chat that we should probably move this to the compiler instead of doing it at lookup time, I agree and will work towards that.

Issue #14721, opened 2016-12-15
Aborting a transition clears the resolved model


btecu:
When you abort a transition in `afterModel` and go to an intermediate route, the model for the previous route does not exist, although it was resolved.

https://ember-twiddle.com/db66247a084b6028d31c29ef856d7301 (See the console log)

pixelhandler:
@locks is it possible that the model is not kept on the application route since setup controller was never reached?

btecu:
@pixelhandler No, that's not the case. You can test this easily but writing an empty `setupController()` (don't call `super` to set the `model` on the `controller`). You'll see that `modelFor` still returns.

btecu:
Any ideas on how to work around this?

pixelhandler:
@btecu do you still have this issue, or did you find a work around?

btecu:
@pixelhandler still have the issue. See #16716.

pixelhandler:
Noted that his has a PR for the test, #16716

btecu:
I talked to @rwjblue about this in person and to add a bit more context, in the application model I load some meta data that is only ever going to be loaded once for the lifetime of the application.

The issue appears when a user is loading say `/account` which is an authenticated route. In `beforeModel` (`routes/account.js`) I check whether the session is authenticated or not. If it's not, I abort the transition and put it in the session service to retry it later once the user authenticates, and finally I do a `intermediateTransitionTo('login')` to allow the user authenticate.

Somehow this causes the `application` route model to be **not** resolved.
The model is actually in the store because the workaround (which has it's own issues) is to do something like:
```
// https://github.com/emberjs/ember.js/issues/14721
// let application = this.modelFor('application');
let application = this.store.peekAll('application-model').get('firstObject');
```

pixelhandler:
@btecu glad you found a work around :)

PR #16716, opened 2018-06-04
Test for #14721

btecu:
Appears to be a `router.js` issue.

Perhaps the resolved model is not added to 
https://github.com/tildeio/router.js/blob/b2bbdec2f0ec16db1f95a9c03268bd2a286dea22/lib/router/transition.js#L28

Depending when / where the transition was aborted, it should still call https://github.com/tildeio/router.js/blob/17a4cd3b6ac8053f3eb20d4358e2bc37653774b6/lib/router/handler-info/resolved-handler-info.js#L10?


btecu:
@chadhietala Given recent `router` work, do you think you might be able to take a look at this?

kategengler:
Is this still relevant?

btecu:
> Is this still relevant?

Looks like, yes.

PR #16718, opened 2018-06-05
Cleanup array mixin

bekzod:


packages/ember-runtime/lib/mixins/array.js
```diff
@@ -936,17 +932,16 @@ const ArrayMixin = Mixin.create(Enumerable, {
     @method reduce
     @param {Function} callback The callback to execute
     @param {Object} initialValue Initial value for the reduce
-    @param {String} reducerProperty internal use only.
```
bekzod: `reducerProperty` seems not used anywhere and remains from SproutCore days I think

bekzod:
@mmun maybe you can review this ?

packages/ember-runtime/lib/mixins/array.js
```diff
@@ -964,17 +959,7 @@ const ArrayMixin = Mixin.create(Enumerable, {
...
-    this.forEach((x, idx) => {
-      let method = x && x[methodName];
-
-      if ('function' === typeof method) {
-        ret[idx] = args.length ? method.apply(x, args) : x[methodName]();
-      }
-    }, this);
-
-    return ret;
+    return this.map(item => tryInvoke(item, methodName, args));
```
rwjblue: This change breaks the API of `invoke`, which now no longer returns an `Ember.A` when prototype extensions are disabled.

bekzod: is wrapping it with `Ember.A` ok ? or revert will help ?

rwjblue: I'm going to revert just this part, and add some tests

rwjblue: https://github.com/emberjs/ember.js/pull/16785

PR #16724, opened 2018-06-07
using Set for storing application instances

IzzatN:
I think using Set data structure is more appropriate for storing application instances


rwjblue:
Thank you!

PR #16726, opened 2018-06-08
move `isDestroyed` check from `getPath` to `get`

bekzod:
make consistent `get` with `set` so it warns (maybe make it `assert` later) when used on destroyed object, I think it will help to discover hidden bugs

### DIFF
--- a/packages/@ember/-internals/metal/lib/property_get.ts
+++ b/packages/@ember/-internals/metal/lib/property_get.ts
@@ -2,9 +2,9 @@
 @module @ember/object
 */
 import { descriptorFor } from '@ember/-internals/meta';
-import { HAS_NATIVE_PROXY, symbol } from '@ember/-internals/utils';
+import { HAS_NATIVE_PROXY, symbol, toString } from '@ember/-internals/utils';
 import { EMBER_METAL_TRACKED_PROPERTIES } from '@ember/canary-features';
-import { assert } from '@ember/debug';
+import { assert, warn } from '@ember/debug';
 import { DEBUG } from '@glimmer/env';
 import { isPath } from './path_cache';
 import { tagForProperty } from './tags';
@@ -26,12 +26,14 @@ if (DEBUG && HAS_NATIVE_PROXY) {
   };
 }
 
-interface MaybeHasUnknownProperty {
-  unknownProperty?: (keyName: string) => any;
+let TOLERATE_DESTROYED: boolean;
+if (DEBUG) {
+  TOLERATE_DESTROYED = false;
 }
 
-interface MaybeHasIsDestroyed {
+interface ExtendedObject {
   isDestroyed?: boolean;
+  unknownProperty?: (keyName: string) => any;
 }
 
 // ..........................................................
@@ -90,6 +92,20 @@ export function get(obj: object, keyName: string): any {
     typeof keyName !== 'string' || keyName.lastIndexOf('this.', 0) !== 0
   );
 
+  if (DEBUG && !TOLERATE_DESTROYED) {
+    let objString: string;
+    TOLERATE_DESTROYED = true;
+    objString = toString(obj);
+    TOLERATE_DESTROYED = false;
+    warn(
+      `calling \`get\` on destroyed object: ${objString}.${keyName}`,
+      !(obj as ExtendedObject).isDestroyed ||
+        keyName === 'isDestroyed' ||
+        keyName === 'isDestroying',
+      { id: 'calling-get-on-destroyed' }
+    );
+  }
+
   let type = typeof obj;
 
   let isObject = type === 'object';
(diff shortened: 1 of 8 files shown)


packages/ember-metal/lib/is_empty.ts
```diff
@@ -1,4 +1,3 @@
-import { get } from './property_get';
```
rwjblue: This is a separate, unrelated change. Can you keep this PR focused specifically on changing `get` and `getPath` and send another PR for these changes?

packages/ember-metal/lib/property_get.ts
```diff
@@ -97,6 +94,16 @@ export function get(obj: object, keyName: string): any {
...
   assert('Cannot call `get` with an empty string', keyName !== '');
+  if (
+    (obj as ExtendedObject).isDestroyed &&
+    keyName !== 'isDestroyed' &&
+    keyName !== 'isDestroying'
+  ) {
+    warn(`calling \`get\` on destroyed object: ${toString(obj)}.${keyName}`, false, {
+      id: 'calling-get-on-destroyed',
+    });
+    return;
```
rwjblue: In general, I think a warning when accessing properties on destroyed objects is probably fine, but I do not think we should prevent accessing the underlying state (at least not right away).

Can you refactor this so that the conditionals are all part of the `warn` call (to ensure that it is all removed properly from production builds) and remove the early return?

packages/ember-metal/lib/property_get.ts
```diff
@@ -97,6 +94,16 @@ export function get(obj: object, keyName: string): any {
     typeof keyName !== 'string' || keyName.lastIndexOf('this.', 0) !== 0
   );
   assert('Cannot call `get` with an empty string', keyName !== '');
+  if (
+    (obj as ExtendedObject).isDestroyed &&
```
rwjblue: We should modify these `isDestroyed` checks to be using `meta.sourceIsDestroyed` instead...

packages/ember-views/lib/system/event_dispatcher.js
```diff
@@ -360,6 +360,10 @@ export default EmberObject.extend({
   },
 
   destroy() {
+    if (this.isDestroyed) {
```
rwjblue: Are we calling `destroy` multiple times?

rwjblue (changes requested):
I'm not certain here.  I

packages/ember-views/lib/system/event_dispatcher.js
bekzod: yep, was thing of adding warning to `destroy` for calling multiple times what do you think ?

packages/ember-metal/lib/property_get.ts
bekzod: that is how it was done before and same way done in `property_set`, maybe make it later with another PR just to keep things incremental so easier to track things if there is issue ?

packages/ember-metal/lib/property_get.ts
bekzod: so I should also remove `return` so it doesn't prevent accessing property of destroyed object ?

packages/ember-metal/lib/is_empty.ts
bekzod: it is a bit related :P , I did so to avoid warning for calling `isEmpty` on destroyed object in tests

packages/ember-metal/lib/property_get.ts
rwjblue: yes, and remove the conditional (moving those conditions to the `warn` itself)

packages/ember-metal/lib/property_get.ts
rwjblue: Yep, totally agreed! I was just trying to point out the issue generally, it definitely isn't related to this PR at all. I'm sorry I didn't mention it in this comment.

packages/ember-metal/lib/property_get.ts
bekzod: done, but `warn` seems to not lazily evaluate string part of warn

bekzod: added PR to `babel-plugin-debug-macros` https://github.com/chadhietala/babel-plugin-debug-macros/pull/72 to support lazy evaluation of string part of warn

packages/@ember/-internals/metal/lib/property_get.ts
```diff
@@ -90,6 +92,20 @@ export function get(obj: object, keyName: string): any {
     typeof keyName !== 'string' || keyName.lastIndexOf('this.', 0) !== 0
   );
 
+  if (DEBUG && !TOLERATE_DESTROYED) {
+    let objString: string;
+    TOLERATE_DESTROYED = true;
```
bekzod: added `TOLERATE_DESTROYED` flag, to avoid getting warning when `get` called within `toString`

bekzod:
@rwjblue minding giving another peek

bekzod:
oops test failing

PR #16728, opened 2018-06-08
make `length` property non configurable in ArrayProxy

bekzod:
this currently fails due to issue fixed in https://github.com/emberjs/ember.js/pull/16727


krisselden:
I prefer just using a getter here #16776

bekzod:
it is still overridable/configurable though :P, anyways closing I think it is more cleaner with getter

runspired:
If something like this moves forward we should keep an eye on ember-data tests and bug reports. We do a bit of manual length management for `Bad Reasons™`

PR #16729, opened 2018-06-08
[BUGFIX beta] Throw error if run.bind receives no method

### LINKED ISSUE #16652: Ember.run.bind should throw an error if the callback is undefined

Serabe:
This commit uses the same logic as backburner.join to know
which method is being bound.

Fixes #16652

### DIFF
--- a/packages/@ember/runloop/index.js
+++ b/packages/@ember/runloop/index.js
@@ -217,7 +217,30 @@ export function join() {
   @since 1.4.0
   @public
 */
-export const bind = (...curried) => (...args) => join(...curried.concat(args));
+export const bind = (...curried) => {
+  assert(
+    'could not find a suitable method to bind',
+    (function(methodOrTarget, methodOrArg) {
+      // Applies the same logic as backburner parseArgs for detecting if a method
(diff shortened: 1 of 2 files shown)


packages/@ember/runloop/index.js
```diff
@@ -217,7 +217,39 @@ export function join() {
   @since 1.4.0
   @public
 */
-export const bind = (...curried) => (...args) => join(...curried.concat(args));
+export const bind = (...curried) => {
+  if (!bindMethodPresent(...curried)) {
+    throw new Error('could not find a suitable method to bind');
```
rwjblue: Can we make this an assertion and inline the `bindMethodPresent` implementation in an IIFE as the predicate?

```js
assert('Could not find a suitable method to bind.', (function() {
  // guts of conditional below...
})());
```

Serabe:
No problem. I used an error because it was what was requested, but can
easily be done.

Serabe:
Done!

rwjblue (approved):

rwjblue:
Thank you!!

PR #16732, opened 2018-06-09
use `applyMixin` directly instead of using it through `Mixin`

bekzod:
I think it is cleaner and readable just to export `applyMixin`  and use it directly instead of accessing though `Mixin`

### DIFF
--- a/packages/ember-metal/lib/mixin.ts
+++ b/packages/ember-metal/lib/mixin.ts
@@ -361,7 +361,7 @@ function replaceObserversAndListeners(
   }
 }
 
-function applyMixin(obj: { [key: string]: any }, mixins: Mixin[], partial: boolean) {
+export function applyMixin(obj: { [key: string]: any }, mixins: Mixin[], partial: boolean) {
   let descs = {};
   let values = {};
   let meta = metaFor(obj);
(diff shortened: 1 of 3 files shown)


packages/ember-runtime/lib/system/core_object.js
```diff
@@ -31,7 +32,6 @@ import { assert } from '@ember/debug';
 import { DEBUG } from '@glimmer/env';
 import { ENV } from 'ember-environment';
 
-const applyMixin = Mixin._apply;
```
bekzod: probably code remains from days when dependencies where imported with global objects https://github.com/emberjs/ember.js/blob/97138247da55eee89be39fd98978c26b86b1aec4/packages/ember-runtime/lib/system/core_object.js

packages/ember-metal/lib/mixin.ts
```diff
@@ -548,14 +548,6 @@ export default class Mixin {
     }
   }
 
-  static _apply() {
-    return (applyMixin as any)(...arguments);
-  }
-
-  static applyPartial(obj: any, ...args: any[]) {
```
bekzod: removed static `applyPartial` since it is undocumented does not have usage in wild,  and `applyMixin` could be used instead

krisselden (approved):

bekzod:
ping tests pass

PR #16734, opened 2018-06-09
update guides link

xg-wang:
Global search replace then manually confirm
emberjs.com/guides => guides.emberjs.com/release


rwjblue:
Thank you!

PR #16735, opened 2018-06-10
[BUGFIX] observed properties not being marked as enum

### LINKED ISSUE #14594: Ember.set & Ember.setProperties fail to set properties "enumerable"
I'm currently using Ember `2.4.3` and wanted to upgrade, but the behaviour of `Ember.set` and `Ember.setProperties` in `2.7.0` onwards seems a bit buggy to me.

To reproduce the "bad" behaviour, do the following:
```javascript
const Test = Ember.Object.extend({
  myProp: null,
  didChangeMyProp: Ember.on('init', Ember.observer('myProp', function () {
    console.log('updated');
  })
});

const testInstance = Test.create();
Ember.setProperties(testInstance, {id: 3, myProp: {id: 1}});

// here, I would expect (and did get!) `['id', 'myProp']` in Ember v2.4.3
// from Ember v2.7.0 onwards, it prints `['id']`
console.log(Object.keys(testInstance));

// alternatively, if you just `set` the `myProp` property
// Ember.set(testInstance, 'myProp', {id: 1});
// the console will only print an empty array.

// however, if I don't initialize the `myProp` property in the `extend()` call, the Object.keys are set as usual
```

It seems, that somewhere - deep in the guts of `Ember.set` && `Ember.setProperties` the `meta` is declaring the property as "not enumerable"?!

I can't reproduce this "bug" at https://ember-twiddle.com as it seems to work over there, if I try these steps in the Chrome DevTools (interesstingly enough, it prints the Ember Version as `2.9.0` but if you call `Ember.VERSION` in the console, it prints `2.8.0` 😸 )

bekzod:
fixes #14594

### DIFF
--- a/packages/@ember/-internals/metal/lib/property_set.ts
+++ b/packages/@ember/-internals/metal/lib/property_set.ts
@@ -1,5 +1,5 @@
 import { descriptorFor, Meta, peekMeta } from '@ember/-internals/meta';
-import { HAS_NATIVE_PROXY, toString } from '@ember/-internals/utils';
+import { HAS_NATIVE_PROXY, lookupDescriptor, toString } from '@ember/-internals/utils';
 import { assert } from '@ember/debug';
 import EmberError from '@ember/error';
 import { DEBUG } from '@glimmer/env';
@@ -19,6 +19,7 @@ let setWithMandatorySetter: <T extends object, K extends Extract<keyof T, string
   keyName: K,
   value: T[K]
 ) => void;
+
 let makeEnumerable: (obj: object, keyName: string) => void;
 
 /**
@@ -125,9 +126,13 @@ if (DEBUG) {
   };
 
   makeEnumerable = (obj: object, key: string) => {
-    let desc = Object.getOwnPropertyDescriptor(obj, key);
+    let desc = lookupDescriptor(obj, key);
 
-    if (desc && desc.set && (desc.set as MandatorySetterFunction).isMandatorySetter) {
+    if (
+      desc !== null &&
+      desc.set !== undefined &&
+      (desc.set as MandatorySetterFunction).isMandatorySetter
+    ) {
       desc.enumerable = true;
       Object.defineProperty(obj, key, desc);
     }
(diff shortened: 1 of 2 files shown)


packages/ember-runtime/tests/system/core_object_test.js
```diff
@@ -93,6 +93,28 @@ moduleFor(
       getOwner(proxy);
     }
 
+    ['@test `Object.keys` works properly'](assert) {
```
rwjblue: Can you tweak this test title? I'd like to see it reference the original issue, and specifically what it is intended to test.

Something like this seems reasonable:

```
@test observed properties are enumerable when set GH#14594
```

Also, can you move this to two tests one in `packages/ember-metal/tests/accessors/set_test.js` and another in `packages/ember-metal/tests/set_properties_test.js`?

rwjblue (changes requested):
Fix looks good to me, only a few tweaks needed to the test then we can land...

packages/ember-runtime/tests/system/core_object_test.js
bekzod: sure

packages/ember-metal/tests/set_properties_test.js
```diff
@@ -1,4 +1,5 @@
-import { setProperties } from '..';
+import { CoreObject } from 'ember-runtime';
```
rwjblue: In general we shouldn't be importing from `ember-runtime` from within `ember-metal`.

The tests in ember-metal will need to be updated to use POJO's instead of `CoreObject`...

rwjblue (changes requested):

packages/ember-metal/tests/set_properties_test.js
bekzod: Issue does not happen with POJO objects it only happens with CoreObject, so moving back to CoreObject test

rwjblue (approved):

rwjblue:
Sorry it took so long here, but this looks good to me. Thanks again @bekzod!

PR #16739, opened 2018-06-11
make setup/teardown noop functions in descriptor

bekzod:

PR #16741, opened 2018-06-11
avoid double Object.defineProperty for native descriptors

bekzod:


packages/ember-metal/lib/properties.ts
```diff
@@ -188,16 +217,18 @@ export function defineProperty(
...
 
-    Object.defineProperty(obj, keyName, {
-      configurable: true,
-      enumerable,
-      get: DESCRIPTOR_GETTER_FUNCTION(keyName, value),
-    });
-
-    meta.writeDescriptors(keyName, value);
-
-    if (typeof desc.setup === 'function') {
```
bekzod: in setup `NativeDescriptor` would make another `Object.defineProperty` call

packages/ember-metal/lib/properties.ts
```diff
@@ -188,16 +217,18 @@ export function defineProperty(
...
-      desc.setup(obj, keyName);
+    if (desc instanceof NativeDescriptor) {
+      Object.defineProperty(obj, keyName, desc.desc);
+    } else {
+      Object.defineProperty(obj, keyName, {
+        configurable: true,
+        enumerable,
+        get: DESCRIPTOR_GETTER_FUNCTION(keyName, value),
+      });
+      meta.writeDescriptors(keyName, value);
```
rwjblue: The `meta.writeDescriptors` here was previously done in both cases and now its only done for non-native.

bekzod: Yes sorry, was going to ask if that is bad idea, else will investigate, tests don’t fail though

rwjblue: To be honest, I am unsure if it is important, but I would assume that we don't want to change the results of things like `descriptorFor` and whatnot.

rwjblue:
I just created a merge conflict here (by merging #16739 ), can you rebase / fix up the conflicts?

rwjblue (changes requested):
After chatting with @krisselden about this, I think we'd rather fix the double `Object.defineProperty` by moving _all_ of the descriptor define properties to be done in `Descriptor.prototype.setup`..

bekzod:
done

packages/ember-metal/lib/properties.ts
```diff
@@ -46,6 +50,36 @@ export abstract class Descriptor {
...
 
+/**
+  A wrapper for a native ES5 descriptor. In an ideal world, we wouldn't need
+  this at all, however, the way we currently flatten/merge our mixins require
+  a special value to denote a descriptor.
+
+  @class NativeDescriptor
+  @private
+*/
+export class NativeDescriptor extends Descriptor {
```
rwjblue: Given how this works now, I think we can move this back to its original location (in `descriptor.ts`), right?

rwjblue (changes requested):
Update looks nice, only one small change to make and this seems good to me...

packages/ember-metal/lib/properties.ts
bekzod: yep, just thought about avoiding extra import/exports :P

bekzod: done

rwjblue:
restarted ci

rwjblue (approved):

krisselden:
this is failing in Edge

image

krisselden:
@bekzod this looks good, are you investigating the CI failure?

bekzod:
briefly looked have no idea :P, will investigate more later when I have time :)

rwjblue:
restarted the browserstack job

krisselden:
I’d love to merge this, anyone able to investigate the IE failure?

bekzod:
there is quirk with Edge where `non-configurable property` do not throw when `delete`ed
Edge 17: https://cl.ly/9f5bfc86f5d7
IE11: https://cl.ly/a62f4c3a0fa7
fixed test by adding flag for Edge case 😝

bekzod:
needs restart, other than that it should work

bekzod:
ping @krisselden

krisselden (approved):

PR #16742, opened 2018-06-11
use `defineProperty` for property based descriptor depreciation 

bekzod:


rwjblue (commented):
Hmm, I think that @krisselden did this on purpose because we only need one of the paths in `Ember.defineProperty`...

rwjblue:
@krisselden - Can you confirm?

krisselden:
@rwjblue confirm.

This is so the deprecation is cleanly removed with the deprecated feature when using a svelte build.

bekzod:
cool

PR #16747, opened 2018-06-13
Update Custom Components feature to match final RFC.

rwjblue:
* Removes view hierarchy updating (not included in RFC)
* Updates custom manager hook method names to match final RFC
* Add `capabilities` system (still very basic)
* Disentangle curly component manager from custom component manager.
* Add optional capabilities (`destructor` / `asyncLifecycleCallbacks`)
* Add temporary global (private on the global, public via modules API)
  for `capabilities`
* Migrate `setComponentManager` to use `WeakMap`'s instead of mutation

(paired with @chancancode on some of the gnarly typescripting)

---

TODO:

- [x] Add more tests based on the examples from the RFC
- [x] Add tests using angle bracket invocation

Future TODO:

- [ ] Determine impact / APIs needed for proper ember-inspector support

### DIFF
--- a/packages/ember-glimmer/lib/component-managers/custom.ts
+++ b/packages/ember-glimmer/lib/component-managers/custom.ts
@@ -1,32 +1,115 @@
-import { ComponentCapabilities, Opaque, Option } from '@glimmer/interfaces';
+import { assert } from '@ember/debug';
+import {
+  ComponentCapabilities,
+  Dict,
+  Opaque,
+  Option,
+  ProgramSymbolTable,
+} from '@glimmer/interfaces';
 import { PathReference, Tag } from '@glimmer/reference';
-import { Arguments, Bounds, CapturedNamedArguments, PrimitiveReference } from '@glimmer/runtime';
+import {
+  Arguments,
+  CapturedArguments,
+  ComponentDefinition,
+  Invocation,
+  PrimitiveReference,
+  WithStaticLayout,
+} from '@glimmer/runtime';
 import { Destroyable } from '@glimmer/util';
-
-import { addChildView } from 'ember-views';
+import { Factory } from 'ember-owner';
+import { OwnedTemplateMeta } from 'ember-views';
 
 import Environment from '../environment';
-import { DynamicScope, Renderer } from '../renderer';
+import RuntimeResolver from '../resolver';
+import { OwnedTemplate } from '../template';
 import { RootReference } from '../utils/references';
 import AbstractComponentManager from './abstract';
-import DefinitionState from './definition-state';
 
-export interface CustomComponentManagerDelegate<T> {
-  version: 'string';
-  create(options: { ComponentClass: T; args: {} }): T;
-  getContext(instance: T): Opaque;
-  update(instance: T, args: {}): void;
-  destroy?(instance: T): void;
-  didCreate?(instance: T): void;
-  didUpdate?(instance: T): void;
-  getView?(instance: T): any;
+const CAPABILITIES = {
+  dynamicLayout: false,
+  dynamicTag: false,
+  prepareArgs: false,
+  createArgs: true,
+  attributeHook: false,
+  elementHook: false,
+  createCaller: false,
+  dynamicScope: true,
+  updateHook: true,
+  createInstance: true,
+};
+
+export interface OptionalCapabilities {
+  asyncLifecycleCallbacks?: boolean;
+  destructor?: boolean;
 }
 
-export interface ComponentArguments<T = {}> {
+export function capabilities(managerAPI: '3.4', options: OptionalCapabilities = {}): Capabilities {
+  assert('Invalid component manager compatibility specified', managerAPI === '3.4');
+
+  return {
+    asyncLifeCycleCallbacks: !!options.asyncLifecycleCallbacks,
+    destructor: !!options.destructor,
+  };
+}
+
+export interface DefinitionState<ComponentInstance> {
+  name: string;
+  ComponentClass: Factory<ComponentInstance>;
+  symbolTable: ProgramSymbolTable;
+  template?: any;
+}
+
+export interface Capabilities {
+  asyncLifeCycleCallbacks: boolean;
+  destructor: boolean;
+}
+
+export interface CustomComponentManagerArgs {
+  named: Dict<Opaque>;
   positional: Opaque[];
-  named: T;
 }
 
+export interface ManagerDelegate<ComponentInstance> {
+  capabilities: Capabilities;
+  createComponent(factory: Opaque, args: CustomComponentManagerArgs): ComponentInstance;
+  updateComponent(instance: ComponentInstance, args: CustomComponentManagerArgs): void;
+  getContext(instance: ComponentInstance): Opaque;
+}
+
+export function hasAsyncLifeCycleCallbacks<ComponentInstance>(
+  delegate: ManagerDelegate<ComponentInstance>
+): delegate is ManagerDelegateWithAsyncLifeCycleCallbacks<ComponentInstance> {
+  return delegate.capabilities.asyncLifeCycleCallbacks;
+}
+
+export interface ManagerDelegateWithAsyncLifeCycleCallbacks<ComponentInstance>
+  extends ManagerDelegate<ComponentInstance> {
+  didCreateComponent(instance: ComponentInstance): void;
+  didUpdateComponent(instance: ComponentInstance): void;
+}
+
+export function hasDestructors<ComponentInstance>(
+  delegate: ManagerDelegate<ComponentInstance>
+): delegate is ManagerDelegateWithDestructors<ComponentInstance> {
+  return delegate.capabilities.destructor;
+}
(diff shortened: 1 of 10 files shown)


packages/ember-glimmer/lib/component-managers/custom.ts
```diff
@@ -1,32 +1,119 @@
...
+): ComponentCapabilityMask {
+  assert('Invalid component manager compatibility specified', managerAPI === '3.4');
+
+  return {
+    asyncLifeCycleCallbacks: !!options.asyncLifecycleCallbacks,
+    destructor: !!options.destructor,
+  };
+}
+
+export interface DefinitionState<ComponentInstance = Opaque> {
```
rwjblue: from kris: lets remove the defaulted `Opaque`'s in these interfaces and swap to pass `Opaque` when we actually use them

packages/ember-glimmer/lib/resolver.ts
```diff
@@ -320,27 +322,41 @@ export default class RuntimeResolver implements IRuntimeResolver<OwnedTemplateMe
...
+        );
+
+        let definition = new CustomManagerDefinition(
+          name,
+          component,
+          manager,
+          layout || meta.owner.lookup<OwnedTemplate>(P`template:components/-default`)
+        );
+        this.componentDefinitionCache.set(key, definition);
+        return definition;
```
rwjblue: Need to call the `finalizer()` before this `return`

packages/ember-glimmer/lib/renderer.ts
```diff
@@ -32,7 +32,7 @@ export type IBuilder = (env: Environment, cursor: Cursor) => ElementBuilder;
 
 export class DynamicScope implements GlimmerDynamicScope {
   constructor(
-    public view: Component | {} | null,
```
rwjblue: Since we removed the dynamic scope manipulation in this PR, we can revert this change

packages/ember-glimmer/lib/resolver.ts
```diff
@@ -353,4 +369,16 @@ export default class RuntimeResolver implements IRuntimeResolver<OwnedTemplateMe
 
     return definition;
   }
+
+  _lookupComponentManager(owner: Owner, managerId: string): CustomComponentManager {
+    if (this.customManagerCache.has(managerId)) {
+      return this.customManagerCache.get(managerId)!;
+    }
+    let delegate = owner.lookup<ManagerDelegate>(`component-manager:${managerId}`);
+    let manager = new CustomComponentManager(delegate);
```
rwjblue: Lets also pass the `owner` to the constructor here. It will allow us to more easily participate in the renderer's view registry when we want to make it work for the Ember Inspector.

packages/ember-glimmer/lib/component-managers/custom.ts
```diff
@@ -1,32 +1,119 @@
...
+  name: string;
+  ComponentClass: Factory<ComponentInstance>;
+  symbolTable: ProgramSymbolTable;
+  template?: any;
+}
+
+export interface ComponentCapabilityMask {
```
chancancode: Probably just call this `Capabilities`. We probably want to mark the "content" of this private somehow, since from the user's/public API perspective, the function is returning an opaque (and explicitly unstable between versions) thing.

packages/ember-glimmer/lib/component-managers/custom.ts
```diff
@@ -1,32 +1,119 @@
...
+  asyncLifeCycleCallbacks: boolean;
+  destructor: boolean;
+}
+
+export interface CustomComponentManagerArgs {
```
chancancode: TODO move into glimmer

packages/ember-glimmer/lib/component-managers/custom.ts
```diff
@@ -1,32 +1,119 @@
...
+  extends ManagerDelegate<ComponentInstance> {
+  destroyComponent(instance: ComponentInstance): void;
+}
+
+export interface ComponentArguments {
+  positional: Opaque[];
+  named: Dict<Opaque>;
+}
+
+function valueForCapturedArgs(args: CapturedArguments): CustomComponentManagerArgs {
```
chancancode: TODO move into glimmer

packages/ember-glimmer/lib/component-managers/custom.ts
```diff
@@ -52,125 +139,129 @@ export interface ComponentArguments<T = {}> {
...
     super();
   }
 
   create(
     _env: Environment,
     definition: DefinitionState,
     args: Arguments,
-    dynamicScope: DynamicScope
-  ): CustomComponentState<T> {
+    _dynamicScope: DynamicScope
```
chancancode: if we don't need it we can just remove the arg right?

(where did we land w.r.t. view tracking again?)

packages/ember-glimmer/lib/component-managers/custom.ts
```diff
@@ -52,125 +139,129 @@ export interface ComponentArguments<T = {}> {
...
+export default class CustomComponentManager<ComponentInstance = Opaque>
+  extends AbstractComponentManager<CustomComponentState<ComponentInstance>, DefinitionState>
+  implements
+    WithStaticLayout<
+      CustomComponentState<ComponentInstance>,
+      DefinitionState<ComponentInstance>,
+      OwnedTemplateMeta,
+      RuntimeResolver
+    > {
+  constructor(private delegate: ManagerDelegate<ComponentInstance>) {
```
chancancode: I am not actually sure if we need this to be a stateful object. The "definition state" and "custom component state" already has the "delegate" in it so we don't have to store it again on `this`. Then this would just be stateless and can be stored in a constant. It's one fewer allocation per component, not sure if it matters much.

packages/ember-glimmer/lib/component-managers/custom.ts
```diff
@@ -52,125 +139,129 @@ export interface ComponentArguments<T = {}> {
...
+    let invocationArgs = valueForCapturedArgs(capturedArgs);
+    const component = delegate.createComponent(definition.ComponentClass, invocationArgs);
 
     return new CustomComponentState(delegate, component, capturedArgs);
   }
 
-  update({ component, args }: CustomComponentState<T>) {
-    this.delegate.update(component, args.value());
+  update({ component, args }: CustomComponentState<ComponentInstance>) {
+    this.delegate.updateComponent(component, valueForCapturedArgs(args));
```
chancancode: so, for example this would be

```ts
update({ delegate, component, args }: CustomComponentState<ComponentInstance>) {
  delegate.updateComponent(component, valueForCapturedArgs(args));
}
```

packages/ember-glimmer/lib/utils/custom-component-manager.ts
```diff
@@ -1,45 +1,27 @@
...
+const getPrototypeOf = Object.getPrototypeOf;
+const MANAGERS: WeakMap<any, string> = new WeakMap();
 
-export function componentManager(obj: any, managerId: String) {
-  if ('reopenClass' in obj) {
-    return obj.reopenClass({
-      [COMPONENT_MANAGER]: managerId,
-    });
-  }
+export function setComponentManager(managerId: string, obj: any) {
```
chancancode: The `set` language in the RFC supposedly give us the license to mutate the class. Would we want to do that instead of a manual prototype chain walk?

chancancode (approved):

packages/ember-glimmer/lib/utils/custom-component-manager.ts
rwjblue: The prototype walk is actually quite fast, and if we move this to setting on the object we break any objects that `Object.seal` or `Object.freeze` themselves (which is actually a good practice, and somewhat common).

Imagine this implementation for "template only components":

```js
owner.register('component-manager:pseudo-template-only', EmberObject.extend({
  capabilities: capabilities('3.4'),

  createComponent() {
    return null;
  },

  updateComponent() {},

  getContext() {
    return null;
  },
});

const BASE_COMPONENT = Object.freeze({});

setComponentManager('pseudo-template-only, BASE_COMPONENT);
```

If we were to mutate the object, we loose this...

packages/ember-glimmer/lib/component-managers/custom.ts
rwjblue: Confirm, will remove.

We landed on: `TODO: Make this work with the Ember Inspector` 😝

packages/ember-glimmer/lib/utils/custom-component-manager.ts
chancancode: ah yes I remember discussing this

packages/ember-glimmer/lib/component-managers/custom.ts
rwjblue: https://github.com/glimmerjs/glimmer-vm/pull/832 implements the interface in glimmer-vm. I'll follow up with a PR to remove the duplication once that lands.

PR #16750, opened 2018-06-15
[BUGFIX release] bring back isObject guard for ember-utils/is_proxy

ampatspell:
`WeakSet.has` throws `TypeError` if argument is not an object:
https://www.ecma-international.org/ecma-262/6.0/#sec-weakset.prototype.has

While recent Chrome versions just returns `false`.


rwjblue (approved):
Looks good, thank you!

rwjblue:
Updated PR title to target release, as it would be good to get this fixed in 3.2.

PR #16754, opened 2018-06-18
[BUGFIX release] Fix container destroy timing

wagenet:


rwjblue (changes requested):
Need to tweak some other parts of the container destroy lifecycle also here, I'll get those changes pushed up into this PR shortly...

packages/ember-runtime/lib/mixins/container_proxy.js
```diff
@@ -101,7 +101,7 @@ let containerProxyMixin = {
     this._super(...arguments);
 
     if (this.__container__) {
-      run(this.__container__, 'destroy');
+      schedule('destroy', this.__container__, 'destroy');
```
stefanpenner: I think more `join` like behavior?

stefanpenner (dismissed):

rwjblue:
Just pushed an update that should address the concerns that @krisselden / @stefanpenner brought up.

Should be good to land once green...

wagenet (commented):
Looks ok. Only point of confusion I had is that `finalizeDestroy` isn't automatically called. Is that as expected? Obviously, this is internal API, but I could see someone potentially calling only `destroy` and never calling `finalizeDestroy`.

rwjblue (approved):

rwjblue:
> Only point of confusion I had is that finalizeDestroy isn't automatically called. Is that as expected? Obviously, this is internal API, but I could see someone potentially calling only destroy and never calling finalizeDestroy.

@wagenet - The container itself is not actually exposed, any user-land code calling `.destroy()` would be on the "owner" which is exactly the thing that mixes in `ContainerProxyMixin` and gets the `destroy` implementation in this PR.

PR #16755, opened 2018-06-18
Ensure tests are run against Chrome 41

rondale-sc:
A recent error in a wrapper around WeakSet caused an issue that broke googlebot.  Chrome latest passed and was undisturbed.  This issue has since been resolved by:

https://github.com/emberjs/ember.js/commit/6346e827a67d0fab8ab02b312bcebca72e2c71d1#diff-ce0053d22123fcbc623fdbd0bd18c5f2

But Googlebot is considerably behind latest Chrome now.  And I'd wager that many people do intend for Googlebot to work.

I really just wanted to start a discussion to determine whether or not this is something that'd be valuable for us to include in our normal testing story.

### DIFF
--- a/testem.dist.js
+++ b/testem.dist.js
@@ -72,6 +72,27 @@ module.exports = {
       ],
       protocol: 'browser',
     },
+    /* Chrome 41 for Googlebot as outlined by:
+     * https://developers.google.com/search/docs/guides/rendering
+     */
+    BS_Chrome_Googlebot: {
+      exe: 'node_modules/.bin/browserstack-launch',
+      args: [
+        '--os',
+        'Windows',
+        '--osv',
+        '10',
+        '--b',
+        'chrome',
+        '--bv',
+        '41.0',
+        '-t',
+        '1200',
+        '--u',
+        '<url>',
+      ],
+      protocol: 'browser',
+    },
     BS_Firefox_Current: {
       exe: 'node_modules/.bin/browserstack-launch',
       args: [


rondale-sc:
Here is the most recent article I was able to find that positively (and officially) identifies Chrome 41 as the latest Googlebot:

https://developers.google.com/search/docs/guides/rendering

rwjblue:
I'm a bit uneasy hard coding this, how do we know when it is updated? Will this page be updated?

rwjblue:
But generally, 👍 on testing what actually matters to us...

testem.dist.js
```diff
@@ -72,6 +72,24 @@ module.exports = {
       ],
       protocol: 'browser',
     },
+    BS_Chrome_FortyOne: {
```
rwjblue: Can we call this `googlebot`, and link to that web page inline?

rondale-sc:
@rwjblue Yeah, I'm uneasy about it also.  But it seems the there is some opaqueness that we have to endure around googlebot.  Google does expose a tool "Fetch and Render" that let's you do some things, but seems you can't determine the exact version of Chromium they are using at any given time.

I read a few blogposts that mentioned that the `--headless` flag added to Chrome in v59 was for Googlebot, but there hasn't been anything I've unearthed that show that new version being in place.  I'm relatively certain that it is not on v59 because the linked issue broke a build for my project.

testem.dist.js
rondale-sc: Definitely.  Will make this change now.

mixonic:
I'm 👍 on this change. Ember's browser support policy in 3.x seems defined by negatives, platforms we do *not* support. We have not wrestled with a positive list of supported platforms because we realized around 3.x that items like Chrome 41 throw a wrench in the plan of "last 2 evergreen browsers + IE11".

We obviously support Ember sites being indexed by the Google search engine. If Chrome 41 is the best approximation of that, I think it should be part of our CI.

rwjblue:
I'm sold 😸

PR #16758, opened 2018-06-18
Test value parameter does not have to be a String

mfeckie:
The doc comments suggest that the test value parameter has to be a String, but this does not match the implementation.

Have also updated type definition here https://github.com/DefinitelyTyped/DefinitelyTyped/pull/26632

### DIFF
--- a/packages/ember-runtime/lib/mixins/array.js
+++ b/packages/ember-runtime/lib/mixins/array.js
@@ -727,7 +727,7 @@ const ArrayMixin = Mixin.create(Enumerable, {
 
     @method rejectBy
     @param {String} key the property to test
-    @param {String} [value] optional value to test against.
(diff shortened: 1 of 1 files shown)


packages/ember-runtime/lib/mixins/array.js
```diff
@@ -727,7 +727,7 @@ const ArrayMixin = Mixin.create(Enumerable, {
 
     @method rejectBy
     @param {String} key the property to test
-    @param {String} [value] optional value to test against.
+    @param {*} [value] optional value to test against.
```
rwjblue: Hmm, I had assumed that this was another property to check against, but it doesn't seem  like it.

rwjblue (approved):

rwjblue:
Thank you!

PR #16772, opened 2018-06-22
Remove unneeded things that create many mixins and merges.

krisselden:


krisselden:
The mixin polymorphism in ember-data was reaching into .PrototypeMixin relying on this including inherited mixins which was totally a mistake. /cc  @runspired

rwjblue:
Ya, we will need to clean that up to something like `meta(modelClass.proto()).hasMixin(thing)`.

runspired:
cc @igorT

PR #16774, opened 2018-06-23
[perf] Only apply PrototypeMixin if it exists

pzuraq:
Currently we are always applying the PrototypeMixin in `proto`, even if
it doesn't yet exist (as in the case of native classes). By checking to
see if it exists first, we should be able to avoid creating and applying
it unless it is necessary.


krisselden:
Half of what I was thinking, we also shouldn't create them until reopen, likely can just change the getter to a set on reopen and just check if undefined.

Currently on reopen results in multiple mixins.  Mixin.create() creates an empty mixin.

rwjblue:
discussed in slack, r+ from @krisselden on this

rwjblue:
Thanks @pzuraq!

PR #16776, opened 2018-06-25
Convert ObjectProxy and ArrayProxy

krisselden:


rwjblue (approved):

runspired:
EmberData currently relies on writing `length` in enumerables and ArrayProxies in the following locations:

RecordArray (extends ArrayProxy)
- https://github.com/emberjs/data/blob/master/addon/-private/system/record-arrays/record-array.js#L246

ManyArray (extends MutableArray)
- https://github.com/emberjs/data/blob/master/addon/-legacy-private/system/many-array.js#L67
- https://github.com/emberjs/data/blob/master/addon/-legacy-private/system/many-array.js#L163
- https://github.com/emberjs/data/blob/master/addon/-legacy-private/system/many-array.js#L180

There may be other places but these were the ones that came to mind quickly.

krisselden:
@runspired why?

krisselden:
@runspired also, why did you link to the ManyArray examples? as far as I know this doesn't affect them?  Do you expect in the proxy case it proxies the set back to the content? as far as I know that was not the old behavior, the old behavior for a CP without a setter for set is to redefine the property as a plain prop then set it, which seems extremely weird for a proxy.

krisselden:
you could just set content to an empty array instead of null, I didn't think null worked.  I'd prefer actually making null content act like an empty array than to do the setter clobbers with value prop

mmun (approved):

runspired:
@krisselden linked the `ManyArray` because I wasn't sure if `MutableArray` was part of the `ArrayProxy` family of things or not. I can implement the change to use an empty-array, and backport it to `2.18` so that it doesn't bite users mixing versions.

PR #16777, opened 2018-06-26
[WIP] [Backport 3.0] Merge pull request #16757 from emberjs/core-object-extend

krisselden:
Make CoreObject rely more on ES2015 class features and interop better.

(cherry picked from commit d424af0b23811b4e751eb5bd989928c6ac6197a1)

### DIFF
--- a/packages/ember-runtime/lib/system/core_object.js
+++ b/packages/ember-runtime/lib/system/core_object.js
@@ -7,7 +7,6 @@
 import {
   assign,
   guidFor,
-  generateGuid,
   makeArray,
   GUID_KEY_PROPERTY,
   symbol,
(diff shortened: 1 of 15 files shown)


packages/ember-runtime/lib/system/core_object.js
```diff
@@ -45,198 +42,135 @@ const reopen = Mixin.prototype.reopen;
...
-        for (let i = 0; i < props.length; i++) {
-          let properties = props[i];
+    // prepare prototype...
+    this.constructor.proto();
 
-          assert(
-            'EmberObject.create only accepts objects.',
-            typeof properties === 'object' || properties === undefined
-          );
+    this.__defineNonEnumerable(GUID_KEY_PROPERTY);
```
pzuraq: I don't think we need to do this _and_ have the getter below on line 511. Defining the GUID_KEY property directly on the instance does work with generateGuid, prevents it from going up the prototype chain, so maybe it's more performant?

rwjblue: It shouldn’t be needed in Ember 3+ since we _always_ use WeakMap’s for the guid (we only needed a property on the object in < 3 to support browsers without weakmaps)

pzuraq:
This is pretty stale, do we need these anymore? I think the current plan would be to backport/support LTS versions only, so 3.4, 2.18, 2.16

rwjblue:
Ya. Seems good.

PR #16784, opened 2018-06-28
[BUGFIX] Setting ArrayProxy#content in willDestroy resets length.

rwjblue:
Prior to this change, `ArrayProxy.prototype.length` was marked as dirty (and therefore recomputed) when `content` was set due to `arrangedContent` being an alias of `content`, and `ArrayProxy` implementing `PROPERTY_DID_CHANGE` to trap `arrangedContent` changes.

Unfortunately, `notifyPropertyChange` avoids notifying _some_ things when the object is destroying. This results in the preexisting `PROPERTY_DID_CHANGE` trap for `arrangedContent` is no longer called, and therefore `length` is never marked as dirty.

This fixes the condition, by implementing a `content` trap in `PROPERTY_DID_CHANGE` that marks length as dirty. The next time that `length` is accessed it will do the normal work to recalculate.

/cc @krisselden @runspired

### DIFF
--- a/packages/ember-runtime/lib/system/array_proxy.js
+++ b/packages/ember-runtime/lib/system/array_proxy.js
@@ -210,9 +210,7 @@ export default class ArrayProxy extends EmberObject {
     if (content) {
       replace(content, value, removedCount, added);
 
-      this._lengthDirty = true;
-      this._objectsDirtyIndex = 0;
-      this._objects = null;
+      this._invalidate();
     }
   }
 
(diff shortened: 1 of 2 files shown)


packages/ember-runtime/lib/system/array_proxy.js
```diff
@@ -230,6 +230,9 @@ export default class ArrayProxy extends EmberObject {
 
       this.arrayContentDidChange(0, oldLength, newLength);
       this._addArrangedContentArrayObsever();
+    } else if (key === 'content') {
```
krisselden: at this point we should extract into an invalidate method

rwjblue: Updated to use a shared `_invalidate` method.

PR #16785, opened 2018-06-28
[BUGFIX] Ensure `ArrayMixin#invoke` returns an Ember.A.

rwjblue:
This is a partial revert of bee179dd, moving back to a manual `forEach` so that we can control the return value a bit more. When prototype extensions are disabled, we cannot rely on `this.map` to return an extended array.

Fixes a regression detected by ember-data's test suite running against ember-source's canary channel.

### DIFF
--- a/packages/ember-runtime/lib/mixins/array.js
+++ b/packages/ember-runtime/lib/mixins/array.js
@@ -951,7 +951,11 @@ const ArrayMixin = Mixin.create(Enumerable, {
     @public
   */
   invoke(methodName, ...args) {
-    return this.map(item => tryInvoke(item, methodName, args));
+    let ret = A();
+
+    this.forEach(item => ret.push(tryInvoke(item, methodName, args)));
+
+    return ret;
   },
 
   /**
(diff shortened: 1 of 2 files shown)


runspired:
My only thought is that nothing outside of this one test in ember-data seems to have broken, and I haven't seen complaints from other folks (yet). The `this.map` behavior seems desirable. Do we maybe want to keep this fix in our pocket and just make ember-data not require the `A` for now?

rwjblue:
@runspired - The issue is that `this.map` for an `Ember.A()` doesn't always return an `Ember.A()` (when prototype extensions are disabled). Changing what `invoke` returns definitely seems like a breaking change to me.

Also, its unlikely to have affected many ember-data users yet because its only on canary, but I'm certain that folks are doing "special" things in their `buildURL` where they expect the snapshots to be an extended thing. Think:

```js
buildURL(type, ids, snapshots) {
  let first = snapshots.get('firstObject'); // :boom: without this change

  // ...snip...
}
```

PR #16788, opened 2018-06-29
[BUGFIX] remove checks on init since ember-cli resolver sets namespace after init.

krisselden:


kategengler (approved):

rwjblue:
This needs to be cherry picked into beta.

PR #16793, opened 2018-07-05
cleanup `_lookupPartial`

bekzod:


packages/ember-glimmer/lib/resolver.ts
```diff
@@ -262,21 +262,16 @@ export default class RuntimeResolver implements IRuntimeResolver<OwnedTemplateMe
...
-      return partial;
+      return new PartialDefinition(name, template);
     } else {
       throw new Error(`${name} is not a partial`);
     }
   }
 
   private _lookupModifier(name: string) {
-    let modifier = this.builtInModifiers[name];
-    if (modifier !== undefined) {
```
bekzod: there is already `undefined` check downstream in `handle` method, so this looks unnecessary

rwjblue (changes requested):

bekzod:
done

krisselden (approved):

PR #16794, opened 2018-07-05
[BUGFIX] Fix instance-initializer-test blueprint for RFC232

### LINKED ISSUE #16216: Ember 2.18 - Not possible to test instance-initializers
Hey,
as from 2.18 it is not possible to test instance initializers using current blueprint.
There are 2 errors:

1. Creating Application
```
Error: Assertion Failed: You cannot make a new Ember.Application using a root element that is an ancestor of an existing Ember.Application
```
2. Booting Application
When we built app instance `application.buildInstance()` it automatically boots app and fires the tested instance-initializer. So because of that, we are not able to prepare the test eg by registering own containers before running the initializer.

stopdropandrew:
instance-initializer tests generated from blueprints fail because the test is attempting to start the application but setupTest hook already has.

Here's an example app with the blueprint and fixed version of the tests: https://github.com/stopdropandrew/ember-fix-instance-initializer-test-blueprint

Fixes https://github.com/emberjs/ember.js/issues/16216


Exelord (approved):

rwjblue:
Thank you!

PR #16797, opened 2018-07-06
Drop autorun assertion

ef4:
This removes the autorun assertion in test mode. The original conditions that caused it to be added are no longer true, as we are on a version of backburner that uses microtask scheduling.

### DIFF
--- a/packages/@ember/runloop/index.js
+++ b/packages/@ember/runloop/index.js
@@ -1,4 +1,4 @@
-import { assert, deprecate, isTesting } from '@ember/debug';
+import { assert, deprecate } from '@ember/debug';
 import { onErrorTarget } from 'ember-error-handling';
 import { beginPropertyChanges, endPropertyChanges } from 'ember-metal';
 import Backburner from 'backburner';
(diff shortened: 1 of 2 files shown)


sukima:
This is the best!! I've been waiting for this for [almost a year](https://github.com/emberjs/rfcs/issues/175#issuecomment-326441449)!! 🎉 @ef4 you rock! You are an inspiration.

rwjblue:
I believe that there are other factors that we need to account for before we teach that using native async/await is ready for production app code usage (outside of the preexisting async route hooks). This change (removing autorun assertion) may be fine now that backburner itself is using microtask queue progression, but we should do some testing to be sure...

krisselden:
@rwjblue I think there is work to be done to improve native promises settling before render but I think it is ok to do this now.

krisselden (approved):

rwjblue:
OK, seems good to me. I do think that we still need to be careful RE: suggesting folks use native async await in app code, but removing the assertion as the logical next step seems fine to me.

runspired:
@krisselden @rwjblue I think that what we want is either to change backburner `onEnd` or glimmer's schedule from `onEnd` to flush via the `eventQueur`/`rAF` (whichever comes first) and to not schedule if already scheduled. This would be the least breaking change that would also ensure native promises and repeated run-loops don't troll us for render perf.

PR #16801, opened 2018-07-09
[Docs] Relax type requirement for RegistryProxy#register

### LINKED ISSUE #16796: [Docs] AppInstance#register w/ instantiate: false
When registering some value to the container with options `{ instantiate: false }`, a subsequent `Ember.getOwner(this).lookup()` will hand the value back to you directly. 

Demo: https://ember-twiddle.com/991fcc164bff88f2bc1edcff9370e90c?openFiles=initializers.foo.js%2C

This has worked for at least 3-4 years, and IIRC ember templates end up in the container by this mechanism. @stefanpenner informs me that this is also part of how the `config/environment.js` module works, in that `ENV` object is stored in the container as a POJO (w/ `instantate: false`)

[The API docs indicate](https://www.emberjs.com/api/ember/3.2/classes/ApplicationInstance/methods/register?anchor=register) that you must pass a `Function` into `register(containerKey, Function)`. Can we regard something like this to be an "intimate API", allowing consumers the flexibility to use the container in this way with confidence?

```ts
// Factory case (current documentation)
register(containerKey: string, value: Function, options?: { instantiate?: boolean, singleton?: boolean })

// Arbitrary value case (instantiate: false --> container doesn't care what the value is)
register(containerKey: string, value: any, options: { instantiate: false })
```

It may also be worth clarifying that the registration API wants an **Ember** factory (something that has a `static create()`)

mike-north:
Fixes #16796

### DIFF
--- a/packages/ember-runtime/lib/mixins/registry_proxy.js
+++ b/packages/ember-runtime/lib/mixins/registry_proxy.js
@@ -92,7 +92,7 @@ export default Mixin.create({
 
     @method register
     @param  fullName {String} type:name (e.g., 'model:user')
-    @param  factory {Function} (e.g., App.Person)
+    @param  factory {any} (e.g., App.Person)
     @param  options {Object} (optional) disable instantiation or singleton usage
     @public
    */


rwjblue (approved):

rwjblue:
Sorry for commit sniping, I just fell asleep at the switch and didn't help shepherd that PR in quickly then a simple one word change PR turned into rebase nightmare 😩

PR #16806, opened 2018-07-11
[Backport 2.16] CoreObject native class compatibility

pzuraq:
Backports the recent changes to CoreObject, along with changes to init
order proposed in RFC 337.


Turbo87:
@rwjblue @kategengler 2.16 is no longer supported, right? so I'm assuming this can be closed?

pzuraq:
I'd say this is still a stretch goal for the polyfill, but we probably won't have time in the near future, and this branch would definitely have to be worked on a fair amount. We can close for now and I can reopen when it's ready.

PR #16808, opened 2018-07-11
Rename `unknown` to `unusable`.

DanielRosenwasser:
TypeScript 3.0 will be introducing a new `unknown` type, (which is a "top" type, and likely does exactly what you want here). As an unfortunate result, this means that `unknown` is a reserved type name, so `unknown` will have to either be renamed, or removed entirely in upgrading to TypeScript 3.0.


rwjblue (approved):
Thank you!

PR #16814, opened 2018-07-12
Fix detached travis builds

krisselden:
Apparently travis always builds by checking out a commit so git info doesn't work for anything but the sha.


krisselden:
This is a quick fix, I need to think about this, travis is always detached, and I need unit tests, but I just wanted to PR something to fix master

kategengler (approved):

rwjblue (approved):

PR #16823, opened 2018-07-16
Update .travis.yml for beta

kategengler:


.travis.yml
```diff
@@ -91,17 +91,9 @@ jobs:
     - env: TEST_SUITE=travis-browsers
     - env:
```
rwjblue: This job should be removed

rwjblue (changes requested):

.travis.yml
kategengler: oic each-package-tests are above too

rwjblue (approved):

PR #16825, opened 2018-07-17
Host App to Engine Transition with Object type QP

### LINKED ISSUE #16655: Host App to Engine's Route Transition with QueryParam of type object is not working properly
In my host app repository, I have an in Repo lazy loading engine. 

When I proceed transition from my host app to engine's route with Query param of type object for the very first time, It shows the error like 

> SyntaxError: Unexpected token o in JSON at position 1

And if i do the same again, its working fine.

Actually, I found that router._queryParamsFor(handlerInfos) method in "router.js" returns the empty objects("map" and "qps") for the first time, host app tries to transtion to engine's route.

For the next time, it returns the value for the objects("map" and "qps"), this is happening becuase engine's instance is already created by the previous transition.

Also, if _queryParamsFor method returns empty object. Since, there is no type check for an object in _serializeQueryParam. It will return the string as "[object object]". .

```
_serializeQueryParam(value, type) {
   if (value === null || value === undefined) {
     return value;
   } else if (type === 'array') {
     return JSON.stringify(value);
   }

   return `${value}`;
 }
```

How can I proceed the transaction to an engine with object type QP?

kishoreyuvan:
fixes #16655

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/router.ts
+++ b/packages/@ember/-internals/routing/lib/system/router.ts
@@ -897,7 +897,7 @@ class EmberRouter extends EmberObject {
   _serializeQueryParam(value: unknown, type: string) {
     if (value === null || value === undefined) {
       return value;
-    } else if (type === 'array') {
+    } else if (type === 'array' || type === 'object') {
       return JSON.stringify(value);
     }
 


kishoreyuvan:
@rwjblue kindly check this.

kishoreyuvan:
@locks @rwjblue any update on this?

rwjblue (commented):
The change seems reasonable to me, but I think we'd need a previously failing test case in order to land this PR.

locks:
@kishoreyuvan are you able to add the test?

kategengler:
Closed as abandoned

PR #16835, opened 2018-07-20
Deprecate `Location.create`

bekzod:


packages/ember-routing/lib/location/api.js
```diff
@@ -1,118 +0,0 @@
...
-    }
-  }
-  ```
-
-  @class Location
-  @private
-*/
-export default {
-  /**
-   This is deprecated in favor of using the container to lookup the location
```
bekzod: this comments has been there since 2014, I think it is time now :P

packages/ember-routing/lib/system/router.js
```diff
@@ -524,25 +523,15 @@ const EmberRouter = EmberObject.extend(Evented, {
...
     let location = get(this, 'location');
-    let rootURL = get(this, 'rootURL');
     let owner = getOwner(this);
 
     if ('string' === typeof location && owner) {
       let resolvedLocation = owner.lookup(`location:${location}`);
-
-      if (resolvedLocation !== undefined) {
-        location = set(this, 'location', resolvedLocation);
-      } else {
```
rwjblue: Need to bring back this branch with a deprecation

packages/ember/index.js
```diff
@@ -582,7 +582,6 @@ Ember.ComponentLookup = views.ComponentLookup;
 Ember.EventDispatcher = views.EventDispatcher;
 
 // ****ember-routing****
-Ember.Location = routing.Location;
```
rwjblue: Needs a deprecation

rwjblue (changes requested):

packages/ember-routing/lib/location/api.js
rwjblue: AFAICT it was never actually deprecated, which means we can't remove it...

bekzod: thought since it is `private` and has `@deprecated` comment we could just drop it :P

rwjblue: Sure, but the deprecated comment doesn’t _actually_ mean that it is deprecated. The fact that it’s private mostly just means that we can remove after the next LTS...

packages/ember-routing/lib/location/api.js
```diff
@@ -87,6 +87,12 @@ export default {
     @private
   */
   create(options) {
+    deprecate(
+      `\`Location.create\` is deprecated, use the \`container\` to register and lookup the location implementation that you need`,
+      false,
+      { id: 'ember-routing.location-create', until: '3.5.0' }
```
rwjblue: Should be until 3.9.0 (master currently represents 3.5)

rwjblue (changes requested):
The until values for the deprecations here should be 3.9.0...

bekzod:
set to `3.9.0`

packages/ember-routing/lib/location/api.js
```diff
@@ -99,20 +105,43 @@ export default {
...
 
-    @private
-    @method getHash
-    @since 1.4.0
-  */
-  _getHash() {
-    return getHash(this.location);
+if (DEBUG && HAS_NATIVE_PROXY) {
+  /* globals Proxy Reflect */
+  Location._implementations = new Proxy(
```
rwjblue: Hmm, this strikes me as odd. Did we previously support `Location._implementations`?

packages/ember-routing/lib/location/hash_location.js
```diff
@@ -53,7 +52,9 @@ export default EmberObject.extend({
     @since 1.5.1
     @method getHash
   */
-  getHash: EmberLocation._getHash,
+  getHash() {
```
rwjblue: Why not `getHash,` here?

packages/ember-routing/lib/location/history_location.js
```diff
@@ -288,7 +287,9 @@ export default EmberObject.extend({
 
     @method getHash
   */
-  getHash: EmberLocation._getHash,
+  getHash() {
```
rwjblue: `getHash,` should work?

packages/ember-routing/lib/location/api.js
bekzod: I just put depreciation for `Location.implementations = { location: MyLocation }` case, when assigned in one go

packages/ember-routing/lib/location/hash_location.js
bekzod: `utils/getHash` expects location object, while this `class getHash` does not expect any arguments

packages/ember-routing/lib/location/history_location.js
bekzod: `utils/getHash` expects location object, while this `class getHash` does not expect any arguments

packages/ember-routing/lib/location/hash_location.js
rwjblue: Gotcha, thank you!

bekzod:
ping

sandstrom:
Just wanted to leave a comment and mention that this is also being worked on in this PR: https://github.com/emberjs/ember.js/pull/19510

PR #16838, opened 2018-07-21
Router(js) refactor (wip)

knownasilya:
Putting this here for visibility, this is still a WIP, please don't merge.

---

Notes from slack:

```no-highlight
Nathan Hammond
Couple of thoughts:
- I generally dislike things like `isSetup` because eventually everything has that check on it. I understand why it's there, but it'd be nice not to need it in the future.

Nathan Hammond
The volatile CPs are because you're aliasing properties that are outside of Ember's world. This is actually a use case for observers. Or possibly refactoring router.js to emit something we can watch for.
```

### DIFF
--- a/broccoli/packages.js
+++ b/broccoli/packages.js
@@ -25,7 +25,7 @@ module.exports.routerES = function _routerES() {
       external: ['route-recognizer', 'rsvp'],
       input: 'index.js',
       output: {
-        file: 'router.js',
+        file: 'router_js.js',
         format: 'es',
       },
     },
--- a/packages/@ember/-internals/routing/lib/services/router.js
+++ b/packages/@ember/-internals/routing/lib/services/router.js
@@ -219,9 +219,10 @@ const RouterService = Service.extend({
    */
   isActive(...args) {
     let { routeName, models, queryParams } = extractRouteArgs(args);
+    let router = this._router;
     let routerMicrolib = this._router._routerMicrolib;
 
-    if (!routerMicrolib.isActiveIntent(routeName, models, null)) {
+    if (!router._isActiveIntent(routeName, models, null)) {
       return false;
     }
     let hasQueryParams = Object.keys(queryParams).length > 0;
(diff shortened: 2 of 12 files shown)


knownasilya:
Maybe the CPs aren't update since they aren't volatile

knownasilya:
I'm thinking it has something to do with stubbing/mocking in the tests..

knownasilya:
@jelhan 🎆

packages/ember-routing/lib/services/router.js
```diff
@@ -197,9 +197,10 @@ const RouterService = Service.extend({
    */
   isActive(...args) {
     let { routeName, models, queryParams } = extractRouteArgs(args);
+    let router = this._router;
```
btecu: Could do `let { _isActiveIntent, state } = this._router;` and use those instead.

luxferresum:
How is this related to [that](https://github.com/emberjs/rfcs/pull/283#issuecomment-405727777)? Is there still help needed?

knownasilya:
Yes related. Feel free to submit against my fork. Next step would be to create/use public APIs from router.js

luxferresum:
how do you feel about moving to ts?

knownasilya:
Moving what to TS? This refactor or router.J's?

chadhietala:
Just giving you a heads up I started adventuring on this journey so that we can implement the rest of https://github.com/emberjs/rfcs/pull/95.

Also related:
- https://github.com/tildeio/router.js/pull/247
- https://github.com/emberjs/ember.js/pull/16888

Also branch when we realized we needed to push more things down into router.js

- https://github.com/emberjs/ember.js/commits/route-info

chadhietala:
It would be amazing if someone could use the type declarations from router.js and convert `ember-router` to TS.

knownasilya:
@chadhietala I've rebased against the changes in master, waiting on travis.

knownasilya:
What is you recommendation around using recognizer? Should that be imported explicitly or should the used methods be wrapped in ember router. For example: `router._routerMicrolib.recognizer.handlersFor(routeName);`

chadhietala:
Sorry was on holiday. So `RouteInfo` is a subset of `HandlerInfo` its going to become public API as described in the [RFC](https://github.com/emberjs/rfcs/blob/aea91e340df658600c0f20a3d9cce4fa5bc44981/text/0000-router-service.md#routeinfo-type).

locks:
hey @knownasilya, is this work still valid?

knownasilya:
I don't think so. Other people tackled refactors at the same time that it would take a while to unwind.

PR #16839, opened 2018-07-21
correctly pass debug name `computed.intersect`

bekzod:
was committed here: https://github.com/emberjs/ember.js/commit/aabc86b5e7506cf53d4df728051c3719b652b5ed#diff-be5c035bb6df695f80dbf5d940188c68R568 , wrongly passing name of function which used for assertion

### DIFF
--- a/packages/@ember/object/lib/computed/reduce_computed_macros.js
+++ b/packages/@ember/object/lib/computed/reduce_computed_macros.js
@@ -575,33 +575,37 @@ export let union = uniq;
   @public
 */
 export function intersect(...args) {
-  return multiArrayMacro(args, function(dependentKeys) {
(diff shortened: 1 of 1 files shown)


packages/@ember/object/lib/computed/reduce_computed_macros.js
```diff
@@ -575,33 +575,37 @@ export let union = uniq;
...
-        if (found === false) {
-          return false;
+          if (found === false) {
+            return false;
+          }
         }
-      }
 
-      return true;
-    }, 'intersect');
```
bekzod: here `intersect` was passed to `.filter`

rwjblue (approved):
Good catch!

PR #16840, opened 2018-07-21
cleanup `computed.collect`

bekzod:
suggesting simplified version of `computed.collect` which does not use `getProperties` and avoids iterating over hash object

### DIFF
--- a/packages/@ember/object/lib/computed/reduce_computed_macros.js
+++ b/packages/@ember/object/lib/computed/reduce_computed_macros.js
@@ -2,13 +2,7 @@
 @module @ember/object
 */
 import { assert } from '@ember/debug';
-import {
-  get,
-  ComputedProperty,
-  addObserver,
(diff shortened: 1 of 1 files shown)


packages/@ember/object/lib/computed/reduce_computed_macros.js
```diff
@@ -704,18 +704,12 @@ export function collect(...dependentKeys) {
...
-          if (properties[key] === undefined) {
-            res.push(null);
-          } else {
-            res.push(properties[key]);
-          }
-        }
-      }
-      return res;
+      let res = dependentKeys.map((key)=> {
+        let val = get(this, key);
```
bekzod: would be particularly useful if it actually concatenated array type values

Turbo87 (approved):
rebased, LGTM

rwjblue (approved):

PR #16841, opened 2018-07-23
fix: QUnit mixin-tests in external addons should import from dummy/mixins

mike-north:
Currently, mixin-tests generated within an external addon import their respective mixins **directly from the `/addon` namespace** like this;

```js
import FooMixin from 'my-addon/mixins/foo';
```

I think it should be like this, to align with all other tests that need to import their respective "object being tested" as a module. 

```js
import FooMixin from 'dummy/mixins/foo';
```
Even if my addon's mixin lives exclusively in the `/app` folder or the dummy app, importing from `/dummy` will cause the test to pass immediately after being generated.

##### Util Tests
https://github.com/emberjs/ember.js/blob/04208064f8ecf4d9a31a428090191b37859be08c/node-tests/fixtures/util-test/dummy.js#L1

##### Initializer Tests
https://github.com/emberjs/ember.js/blob/04208064f8ecf4d9a31a428090191b37859be08c/node-tests/fixtures/instance-initializer-test/dummy.js#L3

##### Helper Unit Tests
https://github.com/emberjs/ember.js/blob/04208064f8ecf4d9a31a428090191b37859be08c/blueprints/helper-test/qunit-rfc-232-files/tests/__testType__/helpers/__name__-test.js#L17

There may be others I haven't run into yet. The existing tests asserted the problematic result, and thus had to be altered.

### DIFF
--- a/blueprints/mixin-test/index.js
+++ b/blueprints/mixin-test/index.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const stringUtils = require('ember-cli-string-utils');
+
 const path = require('path');
 
 const useTestFrameworkDetector = require('../test-framework-detector');
@@ -35,8 +37,13 @@ module.exports = useTestFrameworkDetector({
   },
 
   locals: function(options) {
+    let projectName = options.inRepoAddon ? options.inRepoAddon : options.project.name();
+    let dasherizedModulePrefix = options.inRepoAddon
+      ? projectName
+      : stringUtils.dasherize(options.project.config().modulePrefix);
     return {
-      projectName: options.inRepoAddon ? options.inRepoAddon : options.project.name(),
+      dasherizedModulePrefix,
+      projectName,
       friendlyTestName: ['Unit', 'Mixin', options.entity.name].join(' | '),
     };
   },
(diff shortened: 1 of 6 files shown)


rwjblue:
Hmm, I think I either misunderstand what you are trying to do, or I disagree 😸.

We should _always_ be importing things that will be tested from the addon namespace. I definitely agree there should be consistency, but I think the change is the wrong direction...

Also, FWIW, the _only_ thing an addon should be including in its `app` folders are things that specifically rely on the resolver to be resolved. This would **not** include mixins or util functions (but would include things like components, helpers, and initializers/instance-initializers).

mike-north:
Consistency is easier to teach, and that's what I'm after here. If this *inconsistency* is a new finding, should I create an issue aiming to bring all auto-generated unit/component tests in line with

> We should always be importing things that will be tested from the addon namespace.

? Is this a blanket statement that applies to 
* Mixins
* Utils
* initializers
* helper unit tests

rwjblue:
> Is this a blanket statement that applies to
>
> * Mixins
> * Utils
> * initializers
> * helper unit tests

Yes, if code is coming from an addon and is being imported as normal ES modules it should be imported from the addon's namespace.

PR #16846, opened 2018-07-26
cleanup `view states`

bekzod:
* removed `cloneStates` looks redundant, initially might have been added to avoid tampering of global`states` object so now using `Object.freeze` instead.

* instead of mixing inheritance and `assign`ments just using `assign` to inherit properties from `default` state


krisselden (approved):

rwjblue:
closed/reopened to kick off a fresh CI job

Issue #11063, opened 2015-05-07
Bizard behaviour when transition with partially filled queryParams


novtor:
If we are on the route A with queryParams: {a:1, b:2}, the call from controller:
this.transitionToRoute('A', {'queryParams': {a:2}}
results in transitionning to the route A with queryParams as {a:2, b:2}
where b:2 seems to be taken from the previous transition.
Not sure if this is expected behaviour. I did not test, but I think it can be generalized as: transitioning to a route with partially filled queryParams sets unfilled queryParams with sticky values.

rwjblue:
Wow, I can't believe we have missed this issue completely.  I am terribly sorry!!!!

@novtor - There has been a pretty good amount of cleanup related to query params in the last few weeks, would you be able to whip up a ember-twiddle.com / emberjs.jsbin.com that reproduces this to see if it is fixed?

pixelhandler:
@novtor @rwjblue is this still an issue, perhaps we should close or create a new reproduction of this, what do you think?

pixelhandler:
Per our triage policy I'll close this out for now, feel free to reopen if the issue persists on the current release.

- https://github.com/emberjs/ember.js/blob/master/CONTRIBUTING.md#triaging-policy

PR #16848, opened 2018-07-27
[BUGFIX] normalize route and controller query params same way

bekzod:
This enables object style queryParams in `Controller`s so it is consistent with `Route`'s  qp definition 
```javascript 
Controller.extend({
  queryParams: {  foo: 'bar' }
});
```

Also enables defining `queryParams` with shorthand in `Route`, which was not possible before  
```javascript 
Route.extend({
  queryParams: {  foo: 'bar' }
});
```

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/route.js
+++ b/packages/@ember/-internals/routing/lib/system/route.js
@@ -24,7 +24,7 @@ import {
 import generateController from './generate_controller';
 import {
   stashParamNames,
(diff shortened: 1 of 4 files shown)


bekzod:
possible fix for https://github.com/emberjs/ember.js/issues/11063

locks:
Closing due to no longer being relevant. Thanks again for your work 💪

PR #16853, opened 2018-07-31
[BUGFIX beta] Allow ArrayProxy#pushObjects to accept ArrayProxy again

### LINKED ISSUE #16621: regression 3.0: ArrayProxy#pushObjects doesn't accept ArrayProxy
#16131 cleaned up a lot of `Enumerable` and `Array` mixin code.

However it also changed the assertion in `pushObjects` from [checking for Enumerable or Array](https://github.com/emberjs/ember.js/blob/d9ed94e5989ee641b723e9d1a01a3814053a6540/packages/ember-runtime/lib/mixins/mutable_array.js#L190) to [only checking Array](https://github.com/emberjs/ember.js/blob/db1717b34dc92a266ae54c687bc350b7dcb07de2/packages/ember-runtime/lib/mixins/mutable_array.js#L189)

This assertion now throws when calling `hasMany.pushObjects(otherHasManyOrRecordArray)` even though ember-data's `ManyArray` can accept these objects.

After going over this with @rwjblue I think the path forward is

- [ ] Drop the assertion in `pushObjects` entirely
- [ ] docs: Implementations have to assert their arguments in `replace` (which is the method implementations of `MutableArray` are expected to implement)

`NativeArray` already does assert in `replace` and `ArrayProxy` doesn't need to as it delegates to `content`.

nlfurniss:
This fixes #16621

@rwjblue and @hjdivad proposed the following solution:

- [x] Drop the assertion in pushObjects entirely
- [x] docs: Implementations have to assert their arguments in replace (which is the method implementations of MutableArray are expected to implement)

I'm don't know enough about Ember to know what `#2` means, but would be happy to implement it once I understand what needs to be added to the docs.

### DIFF
--- a/packages/ember-runtime/lib/mixins/array.js
+++ b/packages/ember-runtime/lib/mixins/array.js
@@ -1194,6 +1194,8 @@ const MutableArray = Mixin.create(ArrayMixin, MutableEnumerable, {
     You should replace amt objects started at idx with the objects in the
     passed array. You should also call `this.arrayContentDidChange()`
 
+    Note that this method is expected to validate the type(s) of objects that it expects.
+
     @method replace
     @param {Number} idx Starting index in the array to replace. If
       idx >= length, then append to the end of the array.
@@ -1318,9 +1320,6 @@ const MutableArray = Mixin.create(ArrayMixin, MutableEnumerable, {
     @public
   */
   pushObjects(objects) {
-    if (!Array.isArray(objects)) {
-      throw new TypeError('Must pass Enumerable to MutableArray#pushObjects');
-    }
     this.replace(this.length, 0, objects);
     return this;
   },
--- a/packages/ember-runtime/lib/system/array_proxy.js
+++ b/packages/ember-runtime/lib/system/array_proxy.js
@@ -133,6 +133,8 @@ export default class ArrayProxy extends EmberObject {
     return objectAt(get(this, 'arrangedContent'), idx);
   }
 
+  // See additional docs for `replace` from `MutableArray`:
+  // https://www.emberjs.com/api/ember/3.3/classes/MutableArray/methods/replace?anchor=replace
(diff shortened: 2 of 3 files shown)


nlfurniss:
Weird, wonder why that didn't fail when i ran the tests locally...

What should I do with [this failing test](https://github.com/emberjs/ember.js/blob/ab975417c47d49cd183b1e0470aea3c77c162682/packages/ember-runtime/tests/mutable-array/pushObjects-test.js#L5)? Remove it?

rwjblue:
Only the production build tests were failing because we are using `assert.throws`, however now we are no longer `throw new Error()` we rely on the `replace` implementation to do it and it does an `Ember.assert` (which is stripped in production builds).

tldr; `s/assert.throws/expectAssertion/` in that file

nlfurniss:
Added docs and travis ci passes 🎉

hjdivad:
> docs: Implementations have to assert their arguments in replace (which is the method implementations of MutableArray are expected to implement)

> I'm don't know enough about Ember to know what #2 means, but would be happy to implement it once I understand what needs to be added to the docs.

It means updating the docs in [MutableArray#replace](https://github.com/emberjs/ember.js/blob/b641131599e1a0057d8dcf8e3b715ba208712e26/packages/ember-runtime/lib/mixins/array.js#L1190-L1205) to make clear that part of the `MutableArray` contract is that any type checking done in insertion is done specifically in `replace`.

As a bonus [ArrayProxy#replace](https://github.com/emberjs/ember.js/blob/b641131599e1a0057d8dcf8e3b715ba208712e26/packages/ember-runtime/lib/system/array_proxy.js#L136) should probably refer to the `MutableArray#replace` docs.

nlfurniss:
Travis failed due to disconnect, but last 2 runs were clean

rwjblue (approved):

PR #16854, opened 2018-07-31
[BUGFIX beta] Don't thread FactoryManager through createComponent

chadhietala:
We should not be exposing the FactoryManager to the Custom Component Manager. Instead just pass the factory e.g. `.class` directly into the hook.


rwjblue (approved):
LGTM

rwjblue:
@jordpo - The change here was the correct one to make, perhaps you can open an issue to discuss your specific scenario...

jordpo:
Thanks @rwjblue . I created an issue [here](https://github.com/emberjs/ember.js/issues/17236)

PR #16533, opened 2018-04-17
Add example of router service usage in component

sandstrom:
closes #16518

Issue #16518, opened 2018-04-14
Improve docs for the Router Service


jenweber:
The Router Service docs need a more beginner-friendly description. Right now it says:

> The Router service is the public API that provides component/view layer access to the router.

I recommend that the "main" description be a PR on its own. [This](https://github.com/emberjs/ember.js/blob/v3.1.0/packages/ember-routing/lib/services/router.js#L15) is the file that needs to change. 

Ideally the most commonly used APIs should have code blocks demonstrating their use and some explanations of their purpose or how to use them. For example, see [EmberArray](https://emberjs.com/api/ember/3.1/classes/EmberArray). Additional PRs could include a code block demonstrating the use of each method or property.

Maintainers may feel free to edit this issue directly. @locks can you mark this help-wanted/new contributors if it looks good to you?

- [ ] https://github.com/emberjs/ember.js/issues/16596

## Tasks
- Methods
  - isActive (@rscotchmer) (@pauljeannot) https://github.com/emberjs/ember.js/pull/18471
    - [ ] explanation
    - [ ] code sample
  - recognize
    - [x] explanation
    - [ ] code sample
  - recognizeAndLoad
    - [x] explanation
    - [ ] code sample
  - replaceWith (@teamdraftbox)
    - [x] explanation
    - [ ] code sample
  - transitionTo
    - [x] explanation
    - [ ] code sample
  - urlFor
    - [x] explanation
    - [x] code sample
- Properties
  - currentRoute (@sdebarros)
    - [ ] explanation (what's a `RouteInfo`?)
    - [ ] code sample (needs usage sample)
  - currentRouteName
    - [x] explanation
    - [ ] code sample (needs usage sample)
  - currentURL
    - [x] explanation
    - [ ] code sample (needs usage sample)
  - location
    - [ ] explanation (should link to different Location types)
    - [ ] code sample
  - rootURL
    - [x] explanation
    - [ ] code sample (needs usage sample)
- Events
  - routeDidChange
    - [x] explanation
    - [x] code sample
  - routeWillChange
    - [x] explanation
    - [x] code sample

Parrryy:
I'd be happy to pick this up.

Serap08:
Is this issue still open? if yes, I can contribute.

amanriquez:
Looking for my first open source contribution, any help needed here?

hakilebara:
#16858

locks:
[...]

PR #16858, opened 2018-08-02
[Docs] Add a usage example to the router service

hakilebara:
adding to: https://emberjs.com/api/ember/release/classes/RouterService

A simple example of usage:

   ```javascript
   import { inject } from '@ember/service';

     router: inject(),

     actions: {
       next() {
         this.get('router').transitionTo('other.route');
       }
     }
   })
   ```

### DIFF
--- a/packages/ember-routing/lib/services/router.js
+++ b/packages/ember-routing/lib/services/router.js
@@ -3,8 +3,30 @@ import { readOnly } from '@ember/object/computed';
 import { shallowEqual, resemblesURL, extractRouteArgs } from '../utils';
 
 /**
-   The Router service is the public API that provides component/view layer
-   access to the router.
+   The Router service is the public API that provides access to the router.
+
+   The immediate benefit of the Router service is that you can inject it into components, 
+   giving them a friendly way to initiate transitions and ask questions about the current 
+   global router state.
+
+   In this example, the Router service is injected into a component to initiate a transaction 
+   to a dedicated route:
+   ```javascript
+   import Component from '@ember/component';
+   import { inject as service } from '@ember/service';
+
+   export default Component.extend({
+     router: service(),
+
+     actions: {
+       next() {
+         this.get('router').transitionTo('other.route');
+       }
+     }
+   });
+   ```
+
+   Like any service, it can also be injected into helpers, routes, etc.
 
    @public
    @class RouterService


hakilebara:
I didn't see that a previous PR has been opened on the same subject #16533

packages/ember-routing/lib/services/router.js
```diff
@@ -6,6 +6,23 @@ import { shallowEqual, resemblesURL, extractRouteArgs } from '../utils';
...
    access to the router.
 
+   A simple example of usage:
+
+   ```javascript
+   import Component from '@ember/component';
+   import { inject } from '@ember/service';
+
+   export default Component.extend({
+     router: inject(),
```
toddjordan: I actually like to use `import { inject as service } from '@ember/service';`
like in https://guides.emberjs.com/release/applications/dependency-injection/#toc_ad-hoc-injections

Then you can inject like `router: service(),`

packages/ember-routing/lib/services/router.js
```diff
@@ -6,6 +6,23 @@ import { shallowEqual, resemblesURL, extractRouteArgs } from '../utils';
...
+
+     actions: {
+       next() {
+         this.get('router').transitionTo('other.route');
```
toddjordan: also I think the community is moving towards using the get function, so 
```javascript
import { get } from '@ember/object';
...
get(this, 'router').transitionTo('other.route');
```

packages/ember-routing/lib/services/router.js
hakilebara: I updated the PR

packages/ember-routing/lib/services/router.js
hakilebara: I think we still use `this.get` in the docs https://emberjs.com/api/ember/3.3/functions/@ember%2Fservice/inject

packages/ember-routing/lib/services/router.js
```diff
@@ -6,6 +6,23 @@ import { shallowEqual, resemblesURL, extractRouteArgs } from '../utils';
    The Router service is the public API that provides component/view layer
    access to the router.
 
+   A simple example of usage:
```
locks: Can you be more descriptive and say what the example is showing?

toddjordan:
Ah, I see that previous PR #16533 , and that @locks @sivakumar-kailasam  and @rwjblue reviewed it.  Let me see where that PR is at...

packages/ember-routing/lib/services/router.js
hakilebara: I updated the PR to expand the intro and add an description to the example. 
I reused some text from the original RFC [RFC095](https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md).
@locks what do you think?

hakilebara:
adding ref to https://github.com/emberjs/ember.js/issues/16518

PR #16859, opened 2018-08-02
Create `willDestroyComponent` hook

smfoote:
Also, wrap all bounds-related calls in GLIMMER_COMPONENT_MANAGER_BOUNDS
conditional


packages/ember-glimmer/lib/component-managers/custom.ts
```diff
@@ -221,7 +223,9 @@ export default class CustomComponentManager<ComponentInstance>
   }
 
   didRenderLayout(state: CustomComponentState<ComponentInstance>, bounds: VMBounds) {
-    state.delegate.didRenderLayout(state.component, new Bounds(bounds));
+    if (GLIMMER_COMPONENT_MANAGER_BOUNDS) {
+      state.delegate.didRenderLayout(state.component, new Bounds(bounds));
```
chadhietala: One of the feedback from the RFC was that we should actually pass the `state` through instead of just the component. Can you update both hooks to pass it?

chadhietala (changes requested):

packages/ember-glimmer/lib/component-managers/custom.ts
chancancode: This file is just confusing. state.component _is_ the userland “bucket”.

smfoote: Done.

rwjblue (commented):
Changes look good, though the naming is changed again in the RFC (so we will need another round of updates).

PR #16860, opened 2018-08-03
[BUGFIX release] Clear chains in ProxyMixin when destroyed

pieter-v:
When a ObjectProxy is destroyed while the proxied object is still alive, then the ObjectProxy is leaked because the chains attached to the proxied object retains the ObjectProxy.
This is an issue for us, as ember-data uses proxies for their belongsTo relationships. (See https://github.com/emberjs/data/pull/5554 for more information)

A sample program that demonstrates this can be found at https://github.com/pieter-v/ember-proxy-leak

A memory snapshot before this fix:
before

A memory snapshot after this fix:
after

### DIFF
--- a/packages/@ember/-internals/runtime/lib/mixins/-proxy.js
+++ b/packages/@ember/-internals/runtime/lib/mixins/-proxy.js
@@ -61,9 +61,9 @@ export default Mixin.create({
     m.writableTag(() => combine([DirtyableTag.create(), UpdatableTag.create(CONSTANT_TAG)]));
   },
 
-  destroy() {
-    this._super(...arguments);
+  willDestroy() {
     this.set('content', null);
+    this._super(...arguments);
   },
 
   isTruthy: computed('content', function() {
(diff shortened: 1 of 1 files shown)


packages/ember-runtime/lib/mixins/-proxy.js
```diff
@@ -61,6 +61,11 @@ export default Mixin.create({
     m.writableTag(() => combine([DirtyableTag.create(), UpdatableTag.create(CONSTANT_TAG)]));
   },
 
+  destroy() {
+    this._super(...arguments);
+    this.set('content', null);
```
rwjblue: I'd prefer to set the content before supering. Do you see any issues with that?

pieter-v:
@rwjblue no problem, just left on a long weekend, will do it next Thursday

Op vr 3 aug. 2018 16:05 schreef Robert Jackson <notifications@github.com>:

rwjblue:
Awesome, enjoy!

krisselden (changes requested):
destroy should use ember set only willDestroy, destroy is only supposed to mark is destroying and schedule willDestroy which can set

krisselden (approved):

pieter-v:
rebased

Issue #16863, opened 2018-08-03
[Quest] Mocha Blueprints for RFCs 232/268


Dhaulagiri:
### Background
https://github.com/emberjs/ember.js/issues/15933 updated the blueprints for Qunit to reflect the changes made in RFC 232 and RFC 268. As of ember-mocha 0.14.0 we can implement these same blueprints for Mocha tests as well. 

The desired behavior when generating testing blueprints is to generate "Simplified Mocha" style tests if supported by the host app. We have several tasks below to extend the blueprints to support the "Simplified Mocha" style blueprints alongside the legacy, and ember-mocha blueprints.

### Tasks

 + [x]  https://github.com/emberjs/ember.js/pull/16865 (🔒 @dhaulagiri)  Extend [test-framework-detector.js](https://github.com/emberjs/ember.js/blob/master/blueprints/test-framework-detector.js) to support a directory called `mocha-rfc-232-files` which will be where the "Simplified Mocha" blueprints will live in each of the blueprint test folders (e.g `blueprints/component-test/mocha-rfc-232-files/`). The decision to use the `mocha-rfc-232-files` directory should be based on the host's `ember-mocha` version
 + [ ]  Add `<type>-test` "simplified Mocha" blueprints
   + [x] https://github.com/emberjs/ember.js/pull/16899 (🔒 @dhaulagiri) acceptance-test
   + [x]  https://github.com/emberjs/ember.js/pull/16865 (🔒 @dhaulagiri) component-test
   + [x] https://github.com/emberjs/ember.js/pull/16894 (🔒 @dhaulagiri) controller-test
   + [x]  https://github.com/emberjs/ember.js/pull/16896 (🔒 @dhaulagiri) helper-test
   + [ ]  initializer-test
   + [ ]  instance-initializer-test
   + [x]  https://github.com/emberjs/ember.js/pull/16914 (🔒 @dhaulagiri) mixin-test
   + [x]  https://github.com/emberjs/ember.js/pull/16897 (🔒 @dhaulagiri) route-test
   + [x]  https://github.com/emberjs/ember.js/pull/16913 (🔒 @dhaulagiri) service-test
   + [x] https://github.com/emberjs/ember.js/pull/16898 (🔒 @dhaulagiri) util-test

rwjblue:
[...]

PR #16865, opened 2018-08-05
Add RFC 232 variant detector for mocha & component blueprints

Dhaulagiri:
Addresses https://github.com/emberjs/ember.js/issues/16863 by adding logic to detect when rfc 232-style tests should be used with ember-mocha 0.14 and adds component rendering and unit test blueprints

### DIFF
--- a/blueprints/component-test/mocha-rfc-232-files/__root__/__testType__/__path__/__test__.js
+++ b/blueprints/component-test/mocha-rfc-232-files/__root__/__testType__/__path__/__test__.js
@@ -1,0 +1,38 @@
+<% if (testType === 'integration') { %>import { expect } from 'chai';
+import { describe, it } from 'mocha';
+import { setupRenderingTest } from 'ember-mocha';
+import { render } from '@ember/test-helpers';
+import hbs from 'htmlbars-inline-precompile';
+
+describe('<%= friendlyTestDescription %>', function() {
+  setupRenderingTest();
+
+  it('renders', async function() {
+    // Set any properties with this.set('myProperty', 'value');
+    // Handle any actions with this.set('myAction', function(val) { ... });
+
+    await render(hbs`{{<%= componentPathName %>}}`);
+
+    expect(this.element.textContent.trim()).to.equal('');
+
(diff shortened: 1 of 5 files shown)


Dhaulagiri:
@Turbo87 @rwjblue I'd like to get some momentum behind the remaining items in the quest issue, but I don't want to get too far before validating that the work in this PR is the right direction. Can you let me know any feedback??

simonihmig (approved):
LGTM! 👍

Dhaulagiri:
One thing I noticed is that it may be worth holding this until all the tests are converted. The way this is written, if a type of test hasn't been converted _no_ test file will be generated since the blueprints won't exist.

Dhaulagiri:
Actually, my comment above is not accurate. I was trying to generate a mixin test but that doesn't actually exist 🤷‍♂️ . I'll carry on doing these one at a time with the intention of getting them all done soon regardless.

Turbo87:
@Dhaulagiri looks good in general, but keep in mind that `ember-cli-mocha` should eventually also support this new style.

Dhaulagiri:
@Turbo87 that is good to know, I was under the impression that going forward one would simply use ember-mocha directly and not use ember-cil-mocha`.  Would it be worth accounting for that right now as part of this effort or should we tackle that separately?

I'll try to get a separate discussion going at some point, but it seems like we could also deprecate and remove the pre 0.12 blueprints since it has been quite a long time since those were the default.

Turbo87:
> I was under the impression that going forward one would simply use ember-mocha directly and not use ember-cil-mocha

eventually that should be possible, but right now `ember-cli-mocha` contains the blueprints that automatically adjust the `tests/test-helper.js` to use Mocha instead of QUnit, and some older versions of e.g. `ember-cli-eslint` are not aware of `ember-mocha` being a viable option. 

that said, the blueprints in `ember-cli-mocha` are not up-to-date with the changes in `ember-mocha` yet, which is precisely the reason why there is no current release yet.

rwjblue (approved):

PR #16869, opened 2018-08-07
Implement Bounds API

chadhietala:
This PR implements the Bounds API as described in [RFC#351](https://github.com/emberjs/rfcs/pull/351). /cc @smfoote


packages/ember-glimmer/lib/component-managers/custom.ts
```diff
@@ -71,9 +92,11 @@ export interface CustomComponentManagerArgs {
 
 export interface ManagerDelegate<ComponentInstance> {
   capabilities: Capabilities;
+  didCreateBounds(state: CustomComponentState<ComponentInstance>, bounds: Bounds): void;
```
krisselden: make Bounds an interface not a class.

krisselden (changes requested):

rwjblue:
Just to confirm, this should be closed (unmerged) right?

PR #16874, opened 2018-08-09
[BUGFIX] ES6 classes on/removeListener and observes/removeObserver interop

krisselden:
mixin has been applied.

### DIFF
--- a/packages/ember-meta/lib/meta.ts
+++ b/packages/ember-meta/lib/meta.ts
@@ -34,10 +34,24 @@ if (DEBUG) {
 
 export const UNDEFINED = symbol('undefined');
 
+function ALL() {}
+
 // FLAGS
-const SOURCE_DESTROYING = 1 << 1;
-const SOURCE_DESTROYED = 1 << 2;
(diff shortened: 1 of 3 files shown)


pzuraq (approved):

packages/ember-meta/lib/meta.ts
```diff
@@ -440,82 +452,65 @@ export class Meta {
...
+    target: object | null,
+    method: Function | string,
+    flags: ListenerFlags
+  ) {
+    let { _listeners: listeners } = this;
+    if (listeners === undefined) {
+      listeners = this._listeners = [];
     }
-    this._listenersFinalized = true;
+    listeners.push(eventName, target, method, flags);
```
stefanpenner: should this dedupe entries? Specifically last write win, so that an `add` followed by `remove` doesn't leak the add?

stefanpenner (approved):

pondthailamd (approved):

rwjblue:
Thank you @btecu, I had missed that...

PR #16877, opened 2018-08-11
[CLEANUP] Allow routes to be named "array" and "object"

### LINKED ISSUE #16872: Allow routes named “array”
@rwjblue commented that making “array” a reserved route name is an artifact of things that we don’t do anymore.

Robert, please confirm if my summary is correct and add some details. Feel free to edit my description directly. 

I’m wondering if this is a good issue for new contributors too.

camerondubas:
Hey all 👋 this is my first PR to Ember!

This resolves https://github.com/emberjs/ember.js/issues/16872 by removing "array" as a reserved route name.

### DIFF
--- a/packages/ember-routing/lib/system/dsl.js
+++ b/packages/ember-routing/lib/system/dsl.js
@@ -26,7 +26,7 @@ class DSL {
           return true;
         }
 
-        return ['basic', 'object', 'application'].indexOf(name) === -1;
+        return ['basic', 'application'].indexOf(name) === -1;
       })()
     );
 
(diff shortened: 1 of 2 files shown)


packages/ember-routing/lib/system/dsl.js
```diff
@@ -26,7 +26,7 @@ class DSL {
           return true;
         }
 
-        return ['array', 'basic', 'object', 'application'].indexOf(name) === -1;
+        return ['basic', 'object', 'application'].indexOf(name) === -1;
```
rwjblue: Can you also remove `object`? It was essentially the same boat as `array` and should die too...

camerondubas:
@rwjblue Updated

rwjblue:
Thank you @camerondubas!

PR #16879, opened 2018-08-13
[BUGFIX] isEmpty on nested objects

### LINKED ISSUE #16878: Calling isEmpty on instances of ObjectProxy wrapping ArrayProxy throws Assertion Failed: You attempted to access the `length` property
While working on one of our projects, I encountered an error which I managed to boil down to the following:

```js
Ember.isEmpty(Ember.ObjectProxy.create({
  content: Ember.ArrayProxy.create({
    content: Ember.A([])
  })
}))
```

Looks like a regression caused by utilisation of ES5 getters, it worked in Ember 3.0.0.

It breaks for example `ember-changeset` with `ember-data-model-fragments`.

janvotava:
So ArrayProxy inside ObjectProxy works with ES5 getters and there's no `Assertion Failed: You attempted to access the 'length' property` error.

Fixes #16878

### DIFF
--- a/packages/@ember/-internals/metal/lib/is_empty.ts
+++ b/packages/@ember/-internals/metal/lib/is_empty.ts
@@ -52,17 +52,15 @@ export default function isEmpty(obj: any): boolean {
     if (typeof size === 'number') {
       return !size;
     }
-  }
 
-  if (typeof obj.length === 'number' && objectType !== 'function') {
-    return !obj.length;
-  }
-
-  if (objectType === 'object') {
     let length = get(obj, 'length');
     if (typeof length === 'number') {
       return !length;
     }
+  }
+
+  if (typeof obj.length === 'number' && objectType !== 'function') {
+    return !obj.length;
   }
 
   return false;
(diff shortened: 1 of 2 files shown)


packages/ember-metal/lib/is_empty.ts
```diff
@@ -52,18 +52,16 @@ export default function isEmpty(obj: any): boolean {
...
-  }
 
-  if (objectType === 'object') {
     let length = get(obj, 'length');
     if (typeof length === 'number') {
       return !length;
     }
   }
 
+  if (typeof obj.length === 'number' && objectType !== 'function') {
```
rwjblue: Hmm, do we even need this check anymore?

sheebz45:
I didn’t do this

locks:
@janvotava do you have some time to pick this back up? 😅

wagenet:
@janvotava @locks I rebased this for you.

packages/ember-metal/lib/is_empty.ts
wagenet: This is what allows an empty string (`''`) to be marked as empty.

janvotava:
@locks I'm sorry. What exactly do you mean by that? 😅

locks:
All is good, @wagenet took care of it :)

PR #16882, opened 2018-08-14
component template path

greyhwndz:
fixed some typos regarding the location of component template path


rwjblue:
Hmm, this has a conflict (which seems surprising), can you double check?

Turbo87 (approved):

locks:
The paths are fixed in master.

PR #16892, opened 2018-08-17
Use bare `require` in conditional URL import

tomdale:
This commit changes the code that sets up URL parsing in Glimmer to use the bare, global `require` in Node.js environments, rather than importing from our `node-module` helper package.

The `node-module` package is a shim we emit in the AMD build that exposes the underlying Node `module` and `require` globals which would otherwise be shadowed by the AMD loader. However, this creates problems when compiling Ember to package formats where we want to ultimately rely on standard Node resolution semantics, since `node-module` is not a package that appears in `node_modules`.

Instead, we switch the code to use a bare `require` and rely on the existing `inject-node-globals` Babel transform to detect this usage of bare `require` and inject the appropriate AMD require, only in the AMD build.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/protocol-for-url.ts
+++ b/packages/@ember/-internals/glimmer/lib/protocol-for-url.ts
@@ -1,7 +1,6 @@
 /* globals module, URL */
 
 import { hasDOM } from '@ember/-internals/browser-environment';
-import { IS_NODE, require } from 'node-module';
 import Environment from './environment';
 
 let nodeURL: any;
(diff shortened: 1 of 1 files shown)


rwjblue (approved):
The changes here seem good to me.  I also confirmed we have test coverage of the `protocolForURL` invocation in node-land ([here](https://github.com/emberjs/ember.js/blob/cafbae485d2459b95fcca4301f7a63ad395701d3/tests/node/visit-test.js#L151)) and in normal browser environments ([here](https://github.com/emberjs/ember.js/blob/f73d8440d19cf86a10c61ddb89d45881acfcf974/packages/%40ember/-internals/glimmer/tests/integration/components/attribute-bindings-test.js#L892-L913)).

PR #16893, opened 2018-08-17
[BUGFIX beta] Enable @ember/object#get to be called with an empty string

stefanpenner:
the actual backport


rwjblue:
Closing now, 3.3 is unsupported and I believe that folks should just go to 3.4 (which is an LTS)...

PR #16898, opened 2018-08-18
add mocha rfc 232 util test

Dhaulagiri:
Addresses https://github.com/emberjs/ember.js/issues/16863

FWIW it doesn't look like we generate util tests for ember-cli-mocha 0.12 or greater


simonihmig:
I see the `util` item is already checked in the Quest issue #16863, although this hasn't been merged yet. Anything missing here?

rwjblue (approved):
Nope, just missed it :(

PR #16902, opened 2018-08-20
Tracked tests

smfoote:


packages/ember-metal/tests/tracked/validation_test.ts
```diff
@@ -49,17 +48,21 @@ if (EMBER_METAL_TRACKED_PROPERTIES) {
...
         class Tracked {
-          constructor(name) {
+          @tracked name: NameInterface;
+          constructor(name: NameInterface) {
             this.name = name;
           }
-        }
 
-        track(Tracked, ['name'], {
+          @tracked('name')
```
chadhietala: So my understanding is that these do not need the dependent key.

PR #16903, opened 2018-08-20
[Pre-work] Prepare @tracked decorator

chadhietala:
This is pre-work to expose `tracked` as a decorator. In this PR @smfoote moved the tests over from being a helper function that augmented the descriptors to an actual function that can be used as a decorator. This also updates all the tests to test `@tracked` in static form instead of property descriptor helpers.

/cc @wycats @tomdale


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
btecu: This file shouldn't be commited.

packages/ember-metal/lib/tracked.ts
```diff
@@ -96,12 +96,13 @@ class Tracker {
...
   @param dependencies Optional dependents to be tracked.
  */
-export function tracked(
-  _target: object,
-  key: string,
-  descriptor: PropertyDescriptor
-): PropertyDescriptor {
-  if ('value' in descriptor) {
+export function tracked(...dependencies: string[]): MethodDecorator;
+export function tracked(target: any, key: any): any;
```
tomdale: I believe these can be typed as `unknown` and `PropertyKey`, respectively.

packages/ember-metal/tests/tracked/get_test.ts
```diff
@@ -0,0 +1,140 @@
...
+        let obj = new Obj();
+
+        this.assert.equal(get(obj, ''), 'empty');
+      }
+
+      '@test should get a @tracked path'() {
+        class Key {
+          @tracked value = 'value';
+          constructor() {
+            this.value = 'value';
```
tomdale: Why do we need to set this in the constructor if we're already initializing the property above?

packages/ember-metal/tests/tracked/set_test.ts
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    return new Obj();
+  };
+
+  moduleFor(
+    '@tracked set',
+    class extends AbstractTestCase {
+      teardown() {
+        setHasViews(() => false);
```
tomdale: Why is this needed?

packages/ember-metal/tests/tracked/validation_test.ts
```diff
@@ -85,32 +89,35 @@ if (EMBER_METAL_TRACKED_PROPERTIES) {
...
+            self = this;
           }
         }
 
         defineProperty(
           EmberObject.prototype,
           'full',
           computed('name', function() {
-            let name = get(this, 'name');
+            let name = get(self, 'name');
```
tomdale: I think you're having to introduce the `self` local variable because TypeScript can't infer the type of `this` inside this computed property. If that's the case, a better way to do this would be to annotate the `this` type:

```ts
computed('name', function(this: EmberObject) { /*...*/ })
```

tomdale (changes requested):
Looks great, thank you for doing this. Just a few nitpicks, and I'd like to understand why in the tests we're initializing properties and then re-initializing them in the constructor to the same value. Unless I'm missing something, they're semantically identical.

packages/ember-metal/tests/tracked/get_test.ts
smfoote: Fixed

packages/ember-metal/lib/tracked.ts
smfoote: Fixed

packages/ember-metal/tests/tracked/set_test.ts
smfoote: Apparently it's not. Deleted.

packages/ember-metal/tests/tracked/validation_test.ts
smfoote: Cool! Done.

PR #16905, opened 2018-08-21
Move internal packages to @ember/-internals

tomdale:
Currently, Ember has several internal packages that don't have accompanying public module API (as documented in [RFC 176](https://github.com/emberjs/rfcs/blob/master/text/0176-javascript-module-api.md)), such as `ember-utils` or `ember-metal`.

In order to eventually distribute Ember as a series of standalone npm packages, these internals also need to be published so they can be shared across the multiple packages that depend on them.

Rather than pollute the `@ember` npm scope with many packages that are not intended to be consumed by app developers, this PR moves runtime internal packages to the single `@ember/-internals` package. Each package is accessible as a nested module inside this parent package. For example, `ember-utils` becomes `@ember/-internals/utils` and `ember-metal` becomes `@ember/-internals/metal`.

We are assuming that placing many logically independent packages into a single top-level package is acceptable because it is straightforward to treeshake with tools like webpack or Rollup. A single package for internals eases distribution and dependency management, in addition to surfacing less complexity to app developers.

As part of this PR, we also add `package.json` files to the emitted standalone ES module packages in `dist/es`. In the future, this will allow us to publish each package to npm in multiple module formats.


chadhietala:
Couple questions:

How does topology of `dist` change? How is this related or unrelated to the existing `ember-source`

tomdale:
@chadhietala To the best of my knowledge, the only real world impact this PR has is that the AMD module names will change. For example, `enifed('ember-metal', /*...*/)` becomes `enifed('@ember/-internals/metal', /*...*/)`.

If there are addons that are reaching into the AMD loader internals to find a specific module, those will likely break. Given how much of our internals are exposed on the `Ember` global, though, I'm hoping most folks monkeypatching internals is doing it that way.

Other than that, this change doesn't impact the `ember-source` package because we only publish a limited subset of the entire `dist` directory, via a `files` list in `package.json` (see e.g. https://unpkg.com/ember-source@3.3.2/dist/). This PR primarily targets the `es/` output directory which is not currently published to npm.

rwjblue (approved):

tomdale:
Reviewed this with @rwjblue. We are okay with merging, but have a few non-blocking changes we'd like to make in the future.

1. Get rid of deeply-nested `tests` directory (e.g. `packages/@ember/-internals/metal/tests`) in favor of a single, package-root-level `tests` directory (as before).
2. Build and publish the `ember-source` package from the `dist` directory as part of the build process, rather than a separate process for individual packages and one for the `ember-source` meta-package.
3. Move some of the internals code that is specific to existing public packages to those packages, using a scheme similar to Ember Data. E.g. move `@ember/-internals/runtime/lib/computed.js` to `@ember/object/-private/computed.js`. Some questions around circular dependencies here.
4. Probably want to distribute the ES module build as part of the `ember-source` package in the future (currently filtered by our `files` array in `package.json`) so that we can experiment with consuming them in Ember CLI.

PR #16907, opened 2018-08-23
Upgrade to TypeScript 3.0

tomdale:
This PR upgrades `broccoli-typescript-compiler` to 4.0.1, which also upgrades the version of TypeScript to 3.0.

There was only one small breaking change, which is that TypeScript 2.9 [changed `keyof` types to now include numbers and symbols in addition to strings](
https://blogs.msdn.microsoft.com/typescript/2018/05/31/announcing-typescript-2-9/#updated-keyof). I updated these call signatures to extract just keys of type `string` to maintain the previous behavior.


krisselden (approved):
looks good, I think we should move Opaque to unknown but that can be separate, also, anything that uses `any | undefined | null` should be `{} | undefined | null` or `object | undefined | null` as any is now really any,

PR #16912, opened 2018-08-25
Add tests to detect missing API documentation

jenweber:
### What it does
This PR includes a test of the highest-level yuidoc json. While a passing test does not guarantee that the docs will be rendered in the api app, it should help catch things that would definitely be missing.

### Why it's needed
Historically, ember.js minor releases have caused a regression in the API docs for our users of https://emberjs.com/api/ and a scramble to make a patch release.

### Changes made
- added a test to `bin/yuidoc-tests.js`
- added a command to `package.json` scripts,`test:docs` which builds the docs before running tests with qunit: `ember ember-cli-yuidoc qunit bin/yuidoc-tests.js`

### Help plz
- Is `bin` the right place for this test to go?
- What needs to happen to make the `test:docs` command run?

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -157,7 +157,10 @@ module.exports = {
       },
     },
     {
-      files: [ 'tests/node/**/*.js' ],
+      files: [
+        'tests/node/**/*.js',
+        'bin/yuidoc-tests.js'
+      ],
 
       env: {
         qunit: true
--- a/bin/yuidoc-tests.js
+++ b/bin/yuidoc-tests.js
@@ -10,7 +10,7 @@ const docData = JSON.parse(fs.readFileSync('./docs/data.json', 'utf8'));
 const expectedCounts = [
   {
     category: 'files',
-    count: 180,
+    count: 178,
   },
   {
     category: 'modules',
@@ -18,11 +18,11 @@ const expectedCounts = [
   },
   {
     category: 'classes',
-    count: 109,
+    count: 101,
   },
   {
     category: 'classitems',
-    count: 811,
+    count: 781,
   },
 ];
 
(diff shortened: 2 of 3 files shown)


kategengler:
I'm pretty sure `bin` is the right place for the test, but also not sure the correct place. I would have said `node-tests` but those seem to exclusively blueprint tests. 

To have the `test:docs` command run, I think you want to add a test suite case around here: https://github.com/emberjs/ember.js/blob/master/bin/run-tests.js#L213 And add a stage with that test suite here: https://github.com/emberjs/ember.js/blob/master/.travis.yml#L75 

It should probably be an early stage ala code quality and basic tests, so that CI runs fail fast if it is going to fail.

jenweber:
Thanks Katie! I'll try adding it into those files. I think maybe it should go before browserstack, but after the basic/code quality tests, since building the yuidocs takes a little while.

kategengler:
Sorry, looking closer, there are two stages: basic test and additional tests, and each `env` within them is parallelized (see https://travis-ci.org/emberjs/ember.js). How long does it take to build the yuidoc? If it is shorter than ~6 minutes (the time for the each-package-tests part of basic test stage), I think it should be in the basic test stage.

ef4:
We may want to combine idea from this PR and https://github.com/emberjs/ember.js/pull/16910.

This one is more thorough in that it looks at many categories, not just classitems. But the other one gives more specific feedback about which items disappeared or were added.

jenweber:
Lol, Ed, thank you for writing that! This is what I get for not looking at the PR queue. I'm in favor of closing my PR and adding the other checks into your test.

jenweber:
Closed in favor of #16910

PR #16919, opened 2018-08-29
[refactor] Adds initialization flag and isPrototypeMeta()

pzuraq:
Currently, `meta.isInitialized()` is used for two distinct purposes:

1. To check if the meta is for an _instance_ of `EmberObject` which is
currently being _initialized_.

2. To check if a meta is the meta for a class _prototype_.

In both cases, we want similar behavior - don't trigger event listeners,
mandatory setters, or `setUnknownProperty` in some cases. So, these two
functions were tied together historically out of convenience. This PR
separates them into two methods, `isInitialized` and `isPrototypeMeta`,
to clarify their purposes and allow us to distinguish the two states.

One quirk of `isPrototypeMeta` is that it still requires an object be
passed to it. This for the case where an object does not have it's own
meta, and is relying entirely on its parent's meta. In this case, the
parent meta is being used in the context of an instance, and should be
treated as if it were an instance meta for that particular usage.

### DIFF
--- a/packages/@ember/-internals/meta/lib/meta.ts
+++ b/packages/@ember/-internals/meta/lib/meta.ts
@@ -33,9 +33,13 @@ if (DEBUG) {
 export const UNDEFINED = symbol('undefined');
 
 // FLAGS
-const SOURCE_DESTROYING = 1 << 1;
-const SOURCE_DESTROYED = 1 << 2;
-const META_DESTROYED = 1 << 3;
+const enum MetaFlags {
+  NONE = 0,
+  SOURCE_DESTROYING = 1 << 0,
+  SOURCE_DESTROYED = 1 << 1,
+  META_DESTROYED = 1 << 2,
+  INITIALIZING = 1 << 3,
+}
 
 export class Meta {
   _descriptors: any | undefined;
@@ -46,7 +50,7 @@ export class Meta {
   _chains: any | undefined;
   _tag: Tag | undefined;
   _tags: any | undefined;
-  _flags: number;
+  _flags: MetaFlags;
   source: object;
   proto: object | undefined;
   _parent: Meta | undefined | null;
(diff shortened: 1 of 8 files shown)


yarn.lock
```diff
@@ -112,6 +112,10 @@
   version "0.0.38"
   resolved "https://registry.yarnpkg.com/@types/estree/-/estree-0.0.38.tgz#c1be40aa933723c608820a99a373a16d215a1ca2"
 
+"@types/node@^10.5.5":
```
pzuraq: Unsure why yarn decided to update these, I can force reset if it's a problem

krisselden: because someone pushed a change to package.json without checking the updated yarn.lock

rwjblue: Mind making an issue so we can ensure CI fails in that scenario (when package.json is out of sync with yarn.lock)?

rwjblue (commented):
Seems good to me...

yarn.lock
pzuraq: Done! #16920

rwjblue: That got fixed now, so we should drop the changes here to yarn.lock...

rwjblue (dismissed):

pzuraq:
Rebased, this PR should be good to go

packages/@ember/-internals/meta/lib/meta.ts
```diff
@@ -96,8 +97,20 @@ export class Meta {
...
-    return this.proto !== obj;
+  setInitializing() {
+    this._flags |= MetaFlags.INITIALIZING;
+  }
+
+  unsetInitializing() {
+    this._flags ^= MetaFlags.INITIALIZING;
+  }
+
+  isInitialized() {
```
krisselden: lets name the check here isInitializing then it doesn't have to `not` it, and all the places checking it don't have to `not` it back.

krisselden (dismissed):
Looks good, made one minor comment, after that it is good to go.

rwjblue (approved):

PR #16923, opened 2018-08-30
[BUGFIX beta] ES6 classes on/removeListener and observes/removeObserver interop v2

pzuraq:
This is a rework of #16874 which flattens and caches the state of event listeners more efficiently. Rather than rebuild the result of a `matchListeners` query each time, including deduping, we flatten the listeners down the hierarchy of metas the first time an event match is requested. This still defers the majority of the work early on (adding listeners is cheap) but also prevents us from having to do the work again later.

Should probably run ember-macro-benchmark against this to see if there is any notable perf difference before we merge. I'd also like to see if we can scope the revision counter down to a single hierarchy, so adding an observer on a prototype doesn't affect every class in the app.

cc @krisselden

### DIFF
--- a/packages/@ember/-internals/meta/lib/meta.ts
+++ b/packages/@ember/-internals/meta/lib/meta.ts
@@ -562,8 +562,20 @@ export class Meta {
         kind,
       } as Listener);
     } else {
-      // update own listener
-      listeners[i].kind = kind;
+      let listener = listeners[i];
+      // If the listener is our own function listener and we are trying to
+      // remove it, we want to splice it out entirely so we don't hold onto a
+      // reference.
+      if (
+        typeof method === 'function' &&
+        kind === ListenerKind.REMOVE &&
+        listener.kind !== ListenerKind.REMOVE
+      ) {
+        listeners.splice(i, 1);
+      } else {
+        // update own listener
+        listener.kind = kind;
+      }
     }
   }
 
--- a/packages/@ember/-internals/meta/lib/meta.ts
+++ b/packages/@ember/-internals/meta/lib/meta.ts
@@ -1,5 +1,5 @@
 import { lookupDescriptor, symbol, toString } from '@ember/-internals/utils';
-import { assert } from '@ember/debug';
+import { assert, deprecate } from '@ember/debug';
 import { DEBUG } from '@glimmer/env';
 import { Tag } from '@glimmer/reference';
 
@@ -509,6 +509,16 @@ export class Meta {
   }
 
   removeAllListeners(event: string) {
+    deprecate(
+      'The remove all functionality of removeListener and removeObserver has been deprecated. Remove each listener/observer individually instead.',
+      false,
+      {
+        id: 'events.remove-all-listeners',
+        until: '3.9.0',
+        url: 'https://emberjs.com/deprecations/v3.x#toc_events-remove-all-listeners',
+      }
+    );
+
     if (DEBUG) {
       counters!.removeAllListenersCalls++;
     }
@@ -553,8 +563,34 @@ export class Meta {
       i = -1;
     }
 
-    // if not found, push
+    // if not found, push. Note that we must always push if a listener is not
+    // found, even in the case of a function listener remove, because we may be
+    // attempting to add or remove listeners _before_ flattening has occured.
     if (i === -1) {
+      deprecate(
+        'Adding function listeners to prototypes has been deprecated. Convert the listener to a string listener, or add it to the instance instead.',
+        !(this.isPrototypeMeta(this.source) && typeof method === 'function'),
+        {
+          id: 'events.inherited-function-listeners',
+          until: '3.9.0',
+          url: 'https://emberjs.com/deprecations/v3.x#toc_events-inherited-function-listeners',
+        }
+      );
+
+      deprecate(
+        'You attempted to remove a function listener which did not exist on the instance, which means it was an inherited prototype listener, or you attempted to remove it before it was added. Prototype function listeners have been deprecated, and attempting to remove a non-existent function listener this will error in the future.',
+        !(
+          !this.isPrototypeMeta(this.source) &&
+          typeof method === 'function' &&
+          kind === ListenerKind.REMOVE
+        ),
+        {
+          id: 'events.inherited-function-listeners',
+          until: '3.9.0',
+          url: 'https://emberjs.com/deprecations/v3.x#toc_events-inherited-function-listeners',
+        }
+      );
+
       listeners.push({
         event,
         target,
@@ -567,9 +603,9 @@ export class Meta {
       // remove it, we want to splice it out entirely so we don't hold onto a
       // reference.
       if (
-        typeof method === 'function' &&
         kind === ListenerKind.REMOVE &&
-        listener.kind !== ListenerKind.REMOVE
+        listener.kind !== ListenerKind.REMOVE &&
+        typeof method === 'function'
       ) {
         listeners.splice(i, 1);
       } else {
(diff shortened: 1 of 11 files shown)


packages/@ember/-internals/meta/lib/meta.ts
```diff
@@ -441,82 +478,136 @@ export class Meta {
...
+        if (listeners === undefined) {
+          listeners = this._listeners = [] as Listener[];
+        }
+
+        for (let i = 0; i < parentListeners.length; i++) {
+          let listener = parentListeners[i];
+          let index = indexOfListener(listeners, listener.event, listener.target, listener.method);
+
+          if (index === -1) {
+            listeners.unshift(listener);
```
pzuraq: The current logic relies on the fact that we are updating objects that are shared, but this doesn't work in the `REMOVE_ALL` case. We'll need to splice out any inherited listeners in the event of an update to make sure we're updating everything correctly

packages/@ember/-internals/meta/lib/meta.ts
```diff
@@ -441,82 +480,178 @@ export class Meta {
...
+      } as Listener);
+    } else {
+      // update own listener
+      listeners[i].kind = kind;
+    }
+  }
+
+  /**
+    Flattening is based on a global revision counter. If the revision has
+    bumped it means that somewhere up the inheritance chain something has
```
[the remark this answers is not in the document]
pzuraq: True, should probably read "somewhere up a classes' inheritance chain"

pzuraq:
Benchmark against master, looks like a slight improvement 😄 

[results.pdf](https://github.com/emberjs/ember.js/files/2338578/results.pdf)
[results.json.txt](https://github.com/emberjs/ember.js/files/2338581/results.json.txt)

pzuraq:
Ran benchmarks against emberobserver.com as well and ended up with no significant results (double checked to confirm that substituting was happening correctly). It seems like emberaddons.com is doing something differently that causes the issue, going to add some counters to see if we can find any likely suspects.

pzuraq:
Quick update on research the past week into the regression. So far we've done a few different tests, including:

1. Running benchmarks against both **emberaddons.com** and **emberobserver.com**
2. Adding counters to various actions related to listeners, including adding, removing, and flattening/reopening listeners
3. Recording arguments passed to `addToListeners` and `removeFromListeners` to see what types of listeners are being added and removed, as well as recording what types of listeners are inherited
4. Attempting to rewrite the flattening logic to use different forms of array operations (concat, shift, unshift, slice)
5. Attempting to ensure references are not shared by _duplicating_ inherited listeners rather than inheriting directly by reference.

We weren't able to affect the regression in any significant way by rewriting the code as is (tests **4** and **5**). However, we gathered a few useful datapoints from the other tests:

* The regression occurs only on **emberaddons.com**, **emberobserver.com** is not affected.
* No reopen/reflattens were triggered on _either_ website.
* **emberaddons.com** has a significant number of listeners that are added via _inheritance_, whereas **emberobserver.com** does not. This appeared to be the only significant difference in counters between the two.
* No FunctionListeners were inherited in **emberaddons.com**. All inherited listeners were a single _string_ based listener, an observer on the EmberData Model class.

In discussing with @krisselden this morning we revisited the fact that the current in-master version of the code _does not_ always finalize listeners, and will instead walk the inheritance chain of metas upwards until the root (or until the first finalized meta is found). It may be that building the caches is more expensive than walking the inheritance chain in most cases.

Today I attempted to test this hypothesis by isolating the individual changes in this PR and running the benchmarks again. Here are the results and links to the individual commits

1. Converting listeners to objects **(No regression)**
    * [commit](https://github.com/pzuraq/ember.js/commit/86b432ba603ccfd310bd1703c7c5ce569236aff5)
    * [results.json](https://github.com/emberjs/ember.js/files/2368763/replace-listeners-with-objects-86b432ba6.json.txt)
    * [results.pdf](https://github.com/emberjs/ember.js/files/2368764/replace-listeners-with-objects-86b432ba6.pdf)

2. Always finalizing listeners before matching **(Regression)**
    * [commit](https://github.com/pzuraq/ember.js/commit/1d0dbab8bd585e493df05b385c9239a9ea5836e3)
    * [results.json](https://github.com/emberjs/ember.js/files/2368769/always-finalize-listeners-1d0dbab8bd.json.txt)
    * [results.pdf](https://github.com/emberjs/ember.js/files/2368770/always-finalize-listeners-1d0dbab8bd.pdf)

3. Replacing the finalizing logic with flattening logic, but only flattening in the case of a removal **(No regression)**
    * [commit](https://github.com/pzuraq/ember.js/commit/8c0d32ca5a296a415dabf3d2e855b46930ac8a1d)
    * [results.json](https://github.com/emberjs/ember.js/files/2368771/flatten-listeners-only-on-remove-8c0d32ca5.json.txt)
    * [results.pdf](https://github.com/emberjs/ember.js/files/2368772/flatten-listeners-only-on-remove-8c0d32ca5.pdf)

4. Always flattening listeners (current PR) **(Regression)**

pzuraq:
Further testing this PR on a large application showed no noticeable regression, which means that for two applications (including `emberobserver.com`) we saw no statistically significant change in performance.

packages/@ember/-internals/meta/lib/meta.ts
```diff
@@ -449,82 +493,211 @@ export class Meta {
...
+    // we put remove alls at start because rare and easy to check there
+    listeners.splice(inheritedEnd, 0, {
+      event,
+      target: null,
+      method: null,
+      kind: ListenerKind.REMOVE_ALL,
+    });
+  }
+
+  private pushListener(
```
rwjblue: Should we have a debug counter for this one?

pzuraq: `pushListener` exists just to deduplicate some code from `removeFromListeners` and `addToListeners`, and we have counters for both of those. It's also important to be able to tell what kind of listener object we are pushing (e.g. so we can tell if removing is much more rare than adding), so having unique counters per-wrapper function makes more sense here I think.

rwjblue (approved):

packages/@ember/-internals/metal/lib/events.ts
```diff
@@ -80,7 +80,13 @@ export function removeListener(
     target = null;
   }
 
-  metaFor(obj).removeFromListeners(eventName, target, method!);
+  let m = metaFor(obj);
+
+  if (!method) {
```
chadhietala: This likely should be

```
if (method === undefined) {
  //...
} else {
  //...
}
```

packages/@ember/-internals/meta/lib/meta.ts
```diff
@@ -449,82 +495,258 @@ export class Meta {
...
-        for (let index = 0; index < listeners.length; index += 4) {
-          if (listeners[index] === eventName) {
-            result = result || [];
-            pushUniqueListener(result, listeners, index);
-          }
+  matchingListeners(eventName: string): (string | boolean | object | null)[] | undefined | void {
+    let listeners = this.flattenedListeners();
+
+    if (DEBUG) {
+      counters.
```
btecu: Incomplete statement

pzuraq:
After a bit more work we managed to reduce the perf regression to a statistically insignificant amount. The changes were:

1. We used the parent's listeners when we the child doesn't have unique listeners, AND we cache them locally so we don't have to recurse upwards every time (the caching was pretty important it turned out).
2. Inlined the `setFlattened`/`isFlattened`/`shouldFlatten` methods. It actually made a minor but measurable impact, which is weird because they should have been inlined by the optimizing compiler.
3. We were creating an array for every call to a method, and making that lazy added a tiny boost.

We agreed that this PR should be merged for now, and that we should continue to investigate the perf characteristics as we have time since we _did_ think this was originally going to be an improvement, and there are some odd behaviors here.

The updates to the deprecation guide are [also ready for review](https://github.com/ember-learn/deprecation-app/pull/201).

[results.pdf](https://github.com/emberjs/ember.js/files/2494443/results.pdf)

rwjblue:
Updated PR to indicate `[BUGFIX beta]` since we want this in 3.6...

PR #16924, opened 2018-08-30
[CLEANUP beta]: Clean up router file

bvedad:
If statements flatten. Let used insted of var.


Turbo87 (approved):
@bvedad can you rebase this?

locks:
Thanks for the help. Alas, we do not want to flatten `DEBUG` conditions since it's a macro-ish thing in the build, and the second code change has since changed enough to make this change unviable.
Closing for now!

PR #16933, opened 2018-09-01
Failing test for concatenated classes on contextual glimmer components

cibernox:
We come from here: https://github.com/rwjblue/sparkles-component/issues/7

Seems that in contextual components we can't add default classes to html elements that also receive another class on splattributes.

I don't even know where to look, but if someone can point me in the right direction I can try.

The issue can also be seen in this repo I created to play with sparkles-components: https://github.com/cibernox/sparkles-ember-basic-dropdown


rwjblue (commented):
Thanks! We should add tests for contextual components with other attributes passed also since class has special behavior to merge, but we also should confirm that “normal” attributes work in the test suite.

cibernox:
@rwjblue I added another test for non-concatenable attributes like `title`.

There is however one scenario where I don't know what the expected behavior is:

```hbs
{{!-- ui-button.hbs --}}
<button title="foo" ...attributes></button>
```

```hbs
{{!-- ui-form.hbs --}}
{{yield (hash button=(component "ui-button" title="bar"))}}
```

```hbs
{{!-- application.hbs --}}
<UiForm as |f|>
  <f.button title="baz"></button>
</UiForm>
```

Will the title passes to the `component` helper be considered an attribute or an argument? I think the latter, so it shouldn't be merged used unless `@title` is used in the template. Is there a way to pass attributes to contextual components? Perhaps we should be able to invoke them with a different syntax like `(Component "ui-button" title="bar")` (_note the capital C)

rwjblue:
Correct, anything inside curlies (which the yielded contextual component is) is going to be passed as an argument (and not an attribute to be merged/splatted).

cibernox:
Any direction on where the problem may be? I free have time this week.

rwjblue:
Thank you @cibernox!

PR #16937, opened 2018-09-02
[Cleanup] Remove deprecated flags

btecu:


btecu:
@rwjblue Those flags are already unused.

rwjblue (approved):
Awesome, thanks for confirming! These flags seem like they are from the 1.x -> 2.x era, and the underlying functionality has been removed...

PR #16940, opened 2018-09-05
avoid calling `get` on empty cache

bekzod:


packages/@ember/-internals/glimmer/lib/resolver.ts
```diff
@@ -95,7 +95,7 @@ export default class RuntimeResolver implements IRuntimeResolver<OwnedTemplateMe
   } = BUILTIN_MODIFIERS;
 
   // supports directly imported late bound layouts on component.prototype.layout
-  private templateCache: Map<Owner, Map<TemplateFactory, OwnedTemplate>> = new Map();
```
bekzod: `WeakMap` is more appropriate for `templateCache` I think, as owner destroyed all cache will be destroyed along with it

packages/@ember/-internals/glimmer/lib/resolver.ts
```diff
@@ -188,11 +188,14 @@ export default class RuntimeResolver implements IRuntimeResolver<OwnedTemplateMe
...
   createTemplate(factory: TemplateFactory, owner: Owner): OwnedTemplate {
     let cache = this.templateCache.get(owner);
+    let template;
     if (cache === undefined) {
       cache = new Map();
       this.templateCache.set(owner, cache);
+    } else {
+      template = cache.get(factory);
     }
-    let template = cache.get(factory);
```
bekzod: this is called even if `cache` just created

packages/@ember/-internals/glimmer/lib/resolver.ts
rwjblue: Actually no, we absolutely want a `Map` here. This actually _resolved_ a memory leak in the 3.1 series.

We _never_ want to release templates during the lifetime of the application (which is what this resolver is scoped to).

rwjblue (changes requested):
Need to revert one part, but the changes in `createTemplate` seem good

locks (approved):

bekzod:
@locks can we restart this ?

bekzod:
good to merge :)

Turbo87 (approved):

PR #16957, opened 2018-09-10
use `assign` in `resolvedQueryParams`

bekzod:


packages/@ember/-internals/glimmer/lib/components/link-to.ts
```diff
@@ -323,6 +324,9 @@ import layout from '../templates/link-to';
   @see {Ember.Templates.helpers.link-to}
   @public
 **/
+
+const EMPTY_QUERY_PARAMS = Object.freeze({ values: {} });
```
rwjblue: Freeze doesn't freeze deeply, so I'd like to see:

```js
Const EMPTY_QUERY_PARAMS = Object.freeze({ values: Object.freeze({}) });
```

rwjblue (changes requested):

rwjblue (approved):

PR #16958, opened 2018-09-10
remove redundant `_getModels` in link-to

bekzod:
params already cloned here https://github.com/emberjs/ember.js/pull/16958/files#diff-69db29e32cf13d448d7c3db3008d1774L857 so it can be reused

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/components/link-to.ts
+++ b/packages/@ember/-internals/glimmer/lib/components/link-to.ts
@@ -827,19 +827,6 @@ const LinkComponent = EmberComponent.extend({
     return true;
   }),
 
-  _getModels(params: any[]) {
-    let modelCount = params.length - 1;
-    let models = new Array(modelCount);
-
-    for (let i = 0; i < modelCount; i++) {
-      let value = params[i + 1];
-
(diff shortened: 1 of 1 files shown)


Turbo87 (approved):

packages/@ember/-internals/glimmer/lib/components/link-to.ts
```diff
@@ -891,11 +878,8 @@ const LinkComponent = EmberComponent.extend({
     this.set('queryParams', queryParams);
 
     // 4. Any remaining indices (excepting `targetRouteName` at 0) are `models`.
-    if (params.length > 1) {
-      this.set('models', this._getModels(params));
-    } else {
-      this.set('models', []);
-    }
+    params.shift();
```
rwjblue: We should not mutate the params passed in to us

Turbo87: @rwjblue they are sliced at the very top of the function

rwjblue: Doh! thank you @Turbo87

PR #16978, opened 2018-09-14
[BUGFIX beta] properly teardown alias

bekzod:
attempt to balance consumed/unconsumed properties in `Alias`, 
in some cases `Alias` wouldn't teardown properly leaving unbalanced watch count in meta
1. alias created => `get` called on alias => alias destroyed (still watching on original prop)
2. alias created => observer added on alias => alias destroyed (still watching on original prop)


rwjblue (approved):
This seems good to me, thanks for working on it!

Can you prefix the commit and PR with `[BUGFIX beta]`?

rwjblue:
Thanks for updating the commit title, just waiting on @krisselden's review now...

krisselden (approved):

PR #16983, opened 2018-09-17
cleanup meta

bekzod:


packages/@ember/-internals/meta/lib/meta.ts
```diff
@@ -516,50 +516,46 @@ export class Meta {
...
+  Meta.prototype.peekValues = function(key: string) {
+    return this._findInherited2('_values', key);
   };
 
-  Meta.prototype.deleteFromValues = function(subkey: string) {
-    delete this._getOrCreateOwnMap('_values')[subkey];
+  Meta.prototype.deleteFromValues = function(key: string) {
+    if (this._values !== undefined) {
+      delete this._values[key];
+    }
```
bekzod: avoid creating `_values` map only to delete its prop

packages/@ember/-internals/meta/lib/meta.ts
```diff
@@ -516,50 +516,46 @@ export class Meta {
...
   };
 
-  Meta.prototype.readInheritedValue = function(key, subkey) {
-    let internalKey = `_${key}`;
-
+  Meta.prototype.readInheritedValue = function(key) {
```
bekzod: remove extra key which is always the same `_values`

rwjblue (commented):
Looks like we still have a `_bindings: any | undefined` on the class

krisselden:
Is this all the way done?

Issue #12107, opened 2015-08-14
Sticky query params broken with loading substate


tobymarsden:
Related to #11857:

If a route has a dynamic segment, a query parameter, a child route with a slow-loading model and a loading template, the query parameter is no longer sticky.
#12037 fixed the JS error that was occurring in #11857, but didn't fix the underlying problem.

As far as I can trace back, [`transition` inside `setup` on Route](https://github.com/emberjs/ember.js/blob/v1.13.8/packages/ember-routing/lib/system/route.js#L1117) is null, when it shouldn't be.

Here's a fiddle (Ember Canary) with no loading substate, which has the correct sticky query params: 
http://jsfiddle.net/tobymarsden/x0rngqL2/

And here's a fiddle with a loading substate added, where the query params are no longer sticky:
http://jsfiddle.net/tobymarsden/963qt5p9/

tobymarsden:
Fiddles moved to Ember-Twiddle and updated to Ember 2.1.0:

**Here sticky query params are working:**
https://ember-twiddle.com/91e7b67f87afc0836c4f

**A loading template is added, and now query params are not sticky:**
https://ember-twiddle.com/5b5a761bf8ea50365d80

workmanw:
It's possible that #10945 is also related. It's kind of similar in that a transition inside of an existing transition usually (always?) results in an `abort` call. Though because this is complicated, I can't say for sure.

thijsvdanker:
I'm having the same problem and created these twiddles:

**Both a dynamic segment and a loading template => bug**
https://ember-twiddle.com/63153573fe0150b655613b0ce5e83771?openFiles=templates.main-route.hbs%2C

**Only dynamic segment, no loading template => works**
https://ember-twiddle.com/c8a25c0d17cd05119360fc470eba6d3c?openFiles=templates.main-route.hbs%2C&route=%2Fparent%2Fsub-route

**Only loading template, no dynamic segment => works**
https://ember-twiddle.com/5571992920996228f84bbe06cf7cdb26?openFiles=templates.main-route.hbs%2C&route=%2Fparent%2Fsub-route

sunsetlincoln:
@thijsvdanker Should `//this.transitionTo({ queryParams: { two: 'two' }});` be uncommented in `main-route.js` to demonstrate the issue?

thijsvdanker:
[...]

Issue #14438, opened 2016-10-10
Bug: updating one query param causes the other QP to reset to its default value


lolmaus:
I found another bug with query params: updating one query param causes the other QP to reset to its default value!

The bug only happens in specific circumstances:
- Given controller (let's call it `parent`) must have a query param (say, `foo`).
- There must be some other query param (e. g. `bar`), either in the same controller or in a child controller.
- The `child` route's model hook must return a pending promise.
- There must be a `loading` template adjacent to the `child` route.
- The `parent` route must be nested under a `grandparent` route.
- The `grandparent` route must have a dynamic segment.

The resulting route structure is:
- `grandparent`
  - `parent`
    - `child` (this can be the implicit `index` route)
    - `loading`

Now, if you update `foo` to a non-default value, then updating `bar` will cause `foo` to reset to it's default value!

It's crazy, but removing any of the listed prerequisites prevents the bug from manifesting.

Here's a demo: https://ember-twiddle.com/319583e6a61a449259dc5ea1b6133818

I was able to narrow down the issue to this line:
https://github.com/emberjs/ember.js/blob/3141f835991dee848a4983eb064bfab54fd2e473/packages/ember-routing/lib/system/router.js#L830

``` js
          queryParams[qp.scopedPropertyName] = appCache.lookup(cacheKey, qp.prop, qp.defaultValue);
```

When all prerequisites are met and you update `bar`, then pausing execution at that line for the `foo` QP and running `appCache.lookup(cacheKey, qp.prop)` would return `undefined` and, subsequently, cause `foo` to reset to its default value. If you remove at least one prerequisite, then `appCache.lookup(cacheKey, qp.prop)` returns `foo`'s current value (expected).

Please help. I'm at a loss, I had to finish my work last week but I'm stuck with this. :sob: I've never even heard about `appCache` and I have no idea why it fails.

rwjblue:
Do you have a twiddle or repo that we can look at that demonstrates the issue?  Based on the special setup circumstances, it is likely that we won't be able to reproduce on our own and a demo would help push this forward...

lolmaus:
Yes, here's the demo: https://ember-twiddle.com/319583e6a61a449259dc5ea1b6133818

lolmaus:
It's crazy how specific the prerequisites are. But they are all legit, and the resulting behavior is totally wrong!

[...]

PR #16986, opened 2018-09-18
Failing test to demonstrate query params bug

rlivsey:
If there's:

1. a dynamic segment
2. a loading template
3. a route with a model hook that returns a promise
4. a query param defined

then the query param is not sticky on `link-to`s

Possibly related to https://github.com/emberjs/ember.js/issues/12107

Twiddle demonstrating the issue: https://ember-twiddle.com/97f2c3d7596c97d51846033ad23d75f2

### DIFF
--- a/packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
+++ b/packages/ember/tests/helpers/link_to_test/link_to_with_query_params_test.js
@@ -605,6 +605,49 @@ moduleFor(
         });
     }
 
+    ['@test link-to with dynamic segment and loading route regression test'](assert) {
+      this.router.map(function() {
+        this.route('foo', { path: ':foo' }, function() {
+          this.route('bar', function() {
+            this.route('baz');
+          });
+        });
+      });
+
+      this.addTemplate(
+        'foo.bar',
+        `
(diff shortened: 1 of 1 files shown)


lifeart:
Maybe it's related bug - https://github.com/emberjs/ember.js/issues/14438

rlivsey:
@lifeart ooh interesting, yes the conditions to trigger the bug do appear to be suspiciously similar

lolmaus:
Hi! I'm the author of #14438, which was from 2016.

I've just ran into the issue described here, and my setup is much simpler than in #14438. We have a single query param in the whole app, some dynamic segments and no loading routes.

The query param is defined on a parent controller `foo` with a default value. When I go from `foo.bar.index` to `foo.bar.baz` and back (I don't even touch the query param property!), the URL is updated with explicit query param value, and the `active` class on the corresponding link-to is lost.

rreckonerr:
I just checked if https://github.com/tildeio/router.js/pull/308 fixes your test case and it seems to work! Can someone else confirm it too?

rwjblue:
Included (and fixed) in https://github.com/emberjs/ember.js/pull/19249, thanks again @rlivsey!

PR #16987, opened 2018-09-18
Migrate to Babel 7

rwjblue:
* Updates to Babel 7
* Moves babel helpers into `@ember/-internal` (and removes custom build tooling around generating them)
* Update babel helpers to match current implementations
* Enforce sourcemap validation (via node-land tests)

TODO:

- [ ] Fix `Object.assign` -> `Ember.assign` transformation (currently causing IE11 to fail CI)
- [ ] Debug / fix issue with node tests


rwjblue:
Needs a rebase a fix for the node test failures...

btecu:
I did a rebase but there are tests failing (same as here). Looking at the output, it seems quite different compared to master in terms of size and even content. Left is master, right is this PR rebased to master:

screen shot 2018-10-18 at 23 50 50

Is this expected?

rwjblue:
ya, some differences are expected but we probably would want to dig in (the `ember-all.debug.js` one seems significantly different for example) to make sure things are actually ok.

PR #16988, opened 2018-09-18
Failing test for issue with controllerName + queryParams + transitionTo

### LINKED ISSUE #14560: Using `controllerName` with a parent controller that has a query-param causes assertion on Ember 2.10-beta.2.

cafreeman:
When using the `controllerName` property to share a controller across a parent and child route, `RouterService#transitionTo` fails when attempting to transition to the route that that is "sharing" via `controllerName`.

For example, `parent` has a controller that defines `queryParams` as `['foo']`, and `parent.child` uses `controllerName: 'parent'` to share the controller. Attempting to call `RouterService#transitionTo('parent.child')` will throw the following error:

```
Assertion Failed: You passed the `parent:foo` query parameter during a transition into parent.child, please update to foo
```

Based on the digging I've done so far, I can tell you that this assertion is happening in `router.js` [here](https://github.com/emberjs/ember.js/blob/master/packages/@ember/-internals/routing/lib/system/router.js#L978-L993), and appears to be failing due to `parentProp` being undefined.

There appears to be prior discussion/work on this issue, as seen in the following commits/issues:
- https://github.com/emberjs/ember.js/commit/1c8c6ecad42228e94deffb7672fc6de905fca007
- https://github.com/emberjs/ember.js/commit/3b9489cf6acb61c5ee22f8016f7486ef1e7133d4
- https://github.com/emberjs/ember.js/issues/15725

However, I haven't found anything related specifically to the `controllerName` issue, though the same assertion is being triggered in all cases.

Here is [a twiddle](https://ember-twiddle.com/6d2b17d81c92abd5299ea8d434124381) that demonstrates the issue.

Fixes https://github.com/emberjs/ember.js/issues/14560.

### DIFF
--- a/packages/ember/tests/routing/router_service_test/transitionTo_test.js
+++ b/packages/ember/tests/routing/router_service_test/transitionTo_test.js
@@ -260,34 +260,205 @@ moduleFor(
         .then(() => {
           assert.equal(this.routerService.get('currentURL'), '/child?sort=ASC');
         });
-    }
+      }
 
-    ['@test RouterService#transitionTo passing only queryParams works'](assert) {
-      assert.expect(2);
+      ['@test RouterService#transitionTo with basic query params does not remove query param defaults'](
+        assert
+      ) {
+        assert.expect(1);
 
-      this.add(
-        'controller:parent.child',
-        Controller.extend({
-          queryParams: ['sort'],
-        })
-      );
+        this.add(
+          'controller:parent.child',
+          Controller.extend({
+            queryParams: ['sort'],
+            sort: 'ASC',
+          })
+        );
 
-      let queryParams = this.buildQueryParams({ sort: 'DESC' });
+        let queryParams = this.buildQueryParams({ sort: 'ASC' });
 
-      return this.visit('/')
-        .then(() => {
-          return this.routerService.transitionTo('parent.child');
-        })
-        .then(() => {
-          assert.equal(this.routerService.get('currentURL'), '/child');
-        })
-        .then(() => {
-          return this.routerService.transitionTo(queryParams);
-        })
-        .then(() => {
-          assert.equal(this.routerService.get('currentURL'), '/child?sort=DESC');
-        });
-    }
+        return this.visit('/')
+          .then(() => {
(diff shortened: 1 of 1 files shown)


locks:
@cafreeman are you able to rebase this PR?

cafreeman:
@locks just rebased

btecu:
@cafreeman Is this no longer an issue?

PR #16990, opened 2018-09-19
move WeakSet polyfill to `@ember/polyfills`

bekzod:
it is more appropriate to have it under `@ember/polyfills` I think


packages/@ember/-internals/container/lib/container.ts
```diff
@@ -2,7 +2,7 @@ import { Factory, LookupOptions, Owner, OWNER, setOwner } from '@ember/-internal
 import { dictionary, HAS_NATIVE_PROXY } from '@ember/-internals/utils';
 import { EMBER_MODULE_UNIFICATION } from '@ember/canary-features';
 import { assert } from '@ember/debug';
-import { assign } from '@ember/polyfills';
+import { assign, WeakSet } from '@ember/polyfills';
```
rwjblue: Hmm, why is `WeakSet` added here (it wasn't imported previously)?

packages/@ember/polyfills/index.ts
```diff
@@ -5,4 +5,5 @@ let merge = MERGE ? deprecatedMerge : undefined;
 
 // Export `assignPolyfill` for testing
 export { default as assign, assign as assignPolyfill } from './lib/assign';
+export { default as WeakSet } from './lib/weak_set';
```
rwjblue: The `@ember/polyfills` package does _not_ intend to have this as public API, can you export it as an underscored property?

rwjblue (changes requested):

packages/@ember/-internals/container/lib/container.ts
bekzod: Isn't it used here https://github.com/emberjs/ember.js/pull/16990/files/04fbecca7d3f0a731f653bcc0023f3d3d8d782d9#diff-ed11ef2c87b370bc81065deecede9698L33

packages/@ember/polyfills/index.ts
bekzod: done

rwjblue (approved):

PR #16991, opened 2018-09-19
Failing test for angle bracket helper precedence

jrowlingson:
We ran into an issue when converting the following line in a test:

From this:

```await render(hbs`{{foo-map map=map}}`)```

To this:

```await render(hbs`<FooMap @map={{map}} />`)```

The issue, in our case, was that we were also using the `ember-composable-helpers` addon which includes a `map` helper. It appears that the `map` helper takes precedence over the `map` value set within the context. The solution is either to use `this.map` or rename `map`.

Should this be the other way around? Looking at the [RFC](https://github.com/emberjs/rfcs/blob/master/text/0311-angle-bracket-invocation.md#arguments) it is unclear to me what the expected behavior should be.

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/components/angle-bracket-invocation-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/components/angle-bracket-invocation-test.js
@@ -376,6 +376,21 @@ if (EMBER_GLIMMER_ANGLE_BRACKET_INVOCATION) {
         this.assertText('Hola');
       }
 
+      '@test arguments set in context should take precedence over helpers with same name'(assert) {
+        this.registerHelper('baz', () => { assert.ok(false, 'custom baz helper should not be called'); });
+
+        this.registerComponent('foo-bar', { template: '{{@foo}}' });
+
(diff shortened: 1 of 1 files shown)


rwjblue (commented):
Thank you for putting together a failing test for this! It makes understanding what is being reported much easier. 

Unfortunately, I don’t think this is actually a bug 😭. The invocation rules for mustaches are such that only the _first_ position will automatically invoke a helper, but subsequent path expressions (that are not surrounded in paren) are referring to paths on the template context. 

The best medicine here is to avoid the ambiguous case completely (generally by enabling the template lint rule that forbids it). That would change your invocation to `<FooBar @foo={{this.baz}} />` (and `this.baz` would be used in both curly and angle invocations).

jrowlingson:
Ah, this makes sense, thanks for the clarification.

> The best medicine here is to avoid the ambiguous case completely (generally by enabling the template lint rule that forbids it). 

For future readers - [`no-implicit-this`](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-implicit-this.md) is the rule to enable.

PR #16994, opened 2018-09-20
[refactor][tests] Removes new usage from tests

pzuraq:
Removes usage of `new EmberObject` from tests in preparation for
the changes of RFC 337


rwjblue (commented):
Most of these changes look good (where the test just happen to be using `new` but aren't about using `new`), but we need to ensure **some** tests remain that intentionally use `new` to confirm it continues to work...

pzuraq:
Yep, this only removes incidental usages, the actual tests for `new EmberObject` have been left alone: https://github.com/pzuraq/ember.js/blob/refactor/remove-new-usage-from-tests/packages/@ember/-internals/runtime/tests/system/core_object_test.js#L9-L34

PR #16998, opened 2018-09-22
Fix engine rendering tests setup

chadhietala:
This was missed in one of the migrations. Wasn't fatal but we should bring back the functionality.

_As the comment says:_

> This creates a handler function similar to what is in use by ember-engines internally. Specifically, it returns a promise when transitioning _into_ the first engine route, but returns the synchronously available handler _after_ the engine has been resolved.

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/application/engine-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/application/engine-test.js
@@ -19,12 +19,14 @@ moduleFor(
         // internally. Specifically, it returns a promise when transitioning _into_
         // the first engine route, but returns the synchronously available handler
         // _after_ the engine has been resolved.
-        _getHandlerFunction() {
-          let syncHandler = this._super(...arguments);
+        setupRouter() {
+          this._super(...arguments);
+          let syncHandler = this._routerMicrolib.getHandler;
+
           this._enginePromises = Object.create(null);
           this._resolvedEngines = Object.create(null);
 
-          return name => {
+          this._routerMicrolib.getHandler = name => {
             let engineInfo = this._engineInfoByRoute[name];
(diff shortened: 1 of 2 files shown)


packages/@ember/-internals/glimmer/tests/integration/application/engine-test.js
```diff
@@ -19,12 +19,14 @@ moduleFor(
         // internally. Specifically, it returns a promise when transitioning _into_
         // the first engine route, but returns the synchronously available handler
         // _after_ the engine has been resolved.
-        _getHandlerFunction() {
-          let syncHandler = this._super(...arguments);
+        setupRouter() {
+          this._super(...arguments);
+          let syncHandler = this._routerMicrolib.__proto__.getHandler;
```
rwjblue: I don’t fully grok why we have to pull this from `__proto__`, can you help me understand?

packages/@ember/-internals/glimmer/tests/integration/application/engine-test.js
```diff
@@ -19,12 +19,15 @@ moduleFor(
...
-          let syncHandler = this._super(...arguments);
+        setupRouter() {
+          this._super(...arguments);
+          let syncHandler = this._routerMicrolib.getHandler;
+
           this._enginePromises = Object.create(null);
           this._resolvedEngines = Object.create(null);
 
-          return name => {
+          debugger;
```
rwjblue: Extra debugger

rwjblue (approved):

rwjblue:
@chadhietala - Can you open an issue / pull request to fix ember-engines in the same way? As it stands, we have changed _how_ the monkey patching needs to work (as evidenced by the tests here) but not updated the addon itself...

PR #16999, opened 2018-09-23
[BUGFIX release] fix mouseEnter/Leave event delegation w/o jQuery

### LINKED ISSUE #16922: Mouse enter/leave event not firing with "jquery-integration": false
The enter/leave events don't always fire.

I think it has something to do with the component having a child component in it (a link-to, in this GIF):

Here is a demo repo
https://github.com/amk221/-ember-jquery-mouse-events#example

simonihmig:
The previous implementation did not work correctly when the element of
the component listening to mouseEnter/Leave was quickly moved over,
(i.e. no mouseover event triggered on this element) and the child node
is from another component, as demonstrated in the replication in #16922.

Added better test coverage, with better replication how mouseover/out events fire in reality. The last added tests was the one that catched the misbehaviour.

Related to #16603
Fixes #16922

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/event-dispatcher-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/event-dispatcher-test.js
@@ -167,34 +167,129 @@ moduleFor(
 
       let parent = this.element;
       let outer = this.$('#outer')[0];
+      let inner = this.$('#inner')[0];
 
       // mouse moves over #outer
       this.runTask(() => {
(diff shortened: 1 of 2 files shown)


rwjblue:
Since #16603 was landed in 3.4 (current release and LTS candidate), we need to update this to `[BUGFIX release]`.  Would you mind updating?

simonihmig:
@rwjblue sure. I updated the commit message. Targeting the master branch is still correct, right?

rwjblue:
Yep!

rwjblue (approved):

PR #17007, opened 2018-09-25
Router.js Generics & Route Infos Prep Work

chadhietala:
This PR pulls router.js's generics through Ember's router. It simultaneously introduces the concept of `RouteInfo`. `RouteInfo` is what we typically called `HandlerInfo`. While this was a private API we leaked it out in `did|willTransition` hooks. Because of that it effectively became intimate API. To compensate for this I have added deprecations around the following private apis

- `_routerMicrolib.currentHandlerInfos`
- `_routerMicrolib.getHandler`
- `transition.handlerInfos`
- `transition.state.handlerInfos`
- `handlerInfos[0].handler`

This is based on the usage from [EmberObserver](https://emberobserver.com/code-search?codeQuery=handlerInfos&regex=true)

There is going to be a quick follow that is going to expose `transition.to` and `transition.from` along with the `routeWillChange` and `routeDidChange` events as described by [Router Service RFC](https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md)

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/route.ts
+++ b/packages/@ember/-internals/routing/lib/system/route.ts
@@ -14,8 +14,7 @@ import { once } from '@ember/runloop';
 import { classify } from '@ember/string';
 import { DEBUG } from '@glimmer/env';
 import { TemplateFactory } from '@glimmer/opcode-compiler';
-import { Transition, TransitionState } from 'router_js';
-import HandlerInfo from 'router_js/dist/cjs/handler-info';
+import { InternalRouteInfo, Route as IRoute, Transition, TransitionState } from 'router_js';
 import {
   calculateCacheKey,
   normalizeControllerQueryParams,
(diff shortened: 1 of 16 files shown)


packages/@ember/-internals/routing/lib/system/router.ts
```diff
@@ -44,12 +44,91 @@ export interface QueryParam {
...
 const { slice } = Array.prototype;
 
+const DEPRECATION_MAP = new WeakMap();
+
+function deprecatedTransition(transition: Transition) {
+  if (transition.state) {
+    Object.defineProperty(transition.state, 'handlerInfos', {
+      get() {
+        deprecate(
+          'You attempted to use "transition.state.handlerInfos" which is a private API that will being removed.',
```
csantero: `will being` should say `will be`.

packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -331,7 +332,7 @@ class Route extends EmberObject {
     @method _reset
     @since 1.7.0
   */
-  _reset(isExiting: boolean, transition: Transition) {
+  reset(isExiting: boolean, transition: Transition) {
```
rwjblue: Seems like this is still private, why is the underscore removed? Also, shouldn't we be worried about userland naming collisions?

rwjblue (approved):
One somewhat minor nit-pick, but otherwise looks good to me...

PR #17010, opened 2018-09-26
playing with an idea

stefanpenner:
playing around with hacky improvements for -> https://github.com/tildeio/router.js/pull/87#discussion_r220652231

Basically, if a `this.transitionTo` fails synchronously we should know that the transition is actually not an aborted transition rather a real error that must be surfaced.

I suspect a much better solution is possible. 

cc @chadhietala  / @asakusuma


Turbo87:
closing due to inactivity, conflicts and failing tests. feel free to reopen when work on this continues.

PR #17013, opened 2018-09-27
Fix rendering of empty content with `{{{...}}}` in IE11

chancancode:
...or `{{...}}` with `htmlSafe('')`


chancancode:
Merged in 68ce8e3. The build was green, but there is a typo in the CHANGELOG and I didn't want to trigger another build.

PR #17033, opened 2018-10-03
Remove `hasOwnProperty` checks. Closes #17018.

### LINKED ISSUE #17018: Cannot pass hash helper to link-to (EmptyObject lacking hasOwnProperty)
Sometimes a route can resolve with multiple records like so:

```js
model() {
  const profile = ...;
  const user = ...;

  return RSVP.hash({
    profile,
    user,
  });
},
```

So then you'd think you could use `link-to` like so:

```hbs
{{link-to 'My Profile' 'profiles.show' (hash user=user profile=profile)}}
```

This doesn't work because the `hash` helper returns `EmptyObject` and `extractQueryParams` tries to call `.hasOwnProperty()` on it, resulting in an error.

There's a workaround to this which involves using an action with `transitionTo` but it feels like a poor experience.

This is an issue I've had with Ember for years. What's the correct way to pass multiple records to `link-to` if not this?

I'm using ember-cli 3.0.2.

---

Edit: I've started using the following helper in place of `hash` as a workaround:

```js
import { helper } from '@ember/component/helper';

export function plainObject(params, hash) {
  const object = {};
  // NOTE: `hash` is an `EmptyObject` so we convert to a POJO so that it can be
  // used with `link-to`.
  Object.keys(hash).forEach(k => object[k] = hash[k]);
  return object;
}

export default helper(plainObject);
```

btecu:
Fixes https://github.com/emberjs/ember.js/issues/17018

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/route.ts
+++ b/packages/@ember/-internals/routing/lib/system/route.ts
@@ -2204,10 +2204,6 @@ Route.reopen(ActionHandler, Evented, {
     let propertyNames = [];
 
     for (let propName in combinedQueryParameterConfiguration) {
-      if (!combinedQueryParameterConfiguration.hasOwnProperty(propName)) {
-        continue;
-      }
-
       // to support the dubious feature of using unknownProperty
       // on queryParams configuration
       if (propName === 'unknownProperty' || propName === '_super') {
--- a/packages/@ember/-internals/routing/lib/system/router.ts
+++ b/packages/@ember/-internals/routing/lib/system/router.ts
@@ -1665,9 +1665,6 @@ function forEachQueryParam(
   let qpCache = router._queryParamsFor(routeInfos);
 
   for (let key in queryParams) {
-    if (!queryParams.hasOwnProperty(key)) {
-      continue;
-    }
     let value = queryParams[key];
     let qp = qpCache.map[key];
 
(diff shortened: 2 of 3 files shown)


rwjblue:
@btecu - The changes seem good to me, but we need to try to come up with a failing test for the original issue. I did a quick test locally, by adding the following test to `packages/@ember/-internal/glimmer/tests/integration/components/link-to-test.js`, but the test passes without any changes to source... 🤔 	

```js
    ['@test [GH#17018] using `hash` helper for model']() {
      this.addTemplate('index', `{{link-to 'Post' 'post' (hash id="bar")}}`);
      this.addTemplate('post', 'Post: {{model.id}}');

      this.router.map(function() {
        this.route('post', { path: '/posts/:post_id' });
      });

      return this.visit('/')
        .then(() => {
          this.assertComponentElement(this.firstChild, {
            tagName: 'a',
            attrs: { href: '/posts/bar' },
            content: 'Post',
          });

          return this.click('a');
        })
        .then(() => {
          this.assertText('Post: bar');
        });
    }
```

pzuraq:
It appears that the original issue has been fixed. Let us know if you can provide a reproduction on the latest release/master!

btecu:
@pzuraq issue aside, this could have been merged as general cleanup.

PR #17034, opened 2018-10-03
[FEATURE Router Service] recognize and recognizeAndLoad

chadhietala:
This introduces 2 new methods to the router service.

`recognize`: Takes a string URL and returns a `RouteInfo` for the leafmost route represented by the URL. Returns null if the URL is not recognized. This method expects to receive the actual URL as seen by the browser including the app's `rootURL`.

`recognizeAndLoad`: Takes a string URL and returns a promise that resolves to a `RouteInfoWithAttributes` for the leafmost route represented by the URL. The promise rejects if the URL is not recognized or an unhandled exception is encountered. This method expects to receive the actual URL as seen by the browser including the app's `rootURL`.

This rounds out all APIs on the [router service](https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md#routerservice).

### DIFF
--- a/packages/@ember/-internals/routing/lib/services/router.ts
+++ b/packages/@ember/-internals/routing/lib/services/router.ts
@@ -1,5 +1,6 @@
 import { Evented } from '@ember/-internals/runtime';
 import { EMBER_ROUTING_ROUTER_SERVICE } from '@ember/canary-features';
+import { assert } from '@ember/debug';
 import { readOnly } from '@ember/object/computed';
 import Service from '@ember/service';
 import { Transition } from 'router_js';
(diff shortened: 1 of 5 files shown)


packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -266,6 +267,43 @@ if (EMBER_ROUTING_ROUTER_SERVICE) {
         this.trigger('routeDidChange', transition);
       });
     },
+
+    // Uncomment this when we go the feature
```
chadhietala: Will use `@category`

rwjblue (approved):

PR #17038, opened 2018-10-03
[FEATURE Router Service] Deprecate touching transition state

chadhietala:
This deprecates access to the `state`, `params` and `queryParams` fields on the `Transition` that were private but became defacto public API since there was no other way to access it. The router service will expose `RouteInfo` objects which is the public object that has this state on it.

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/route.ts
+++ b/packages/@ember/-internals/routing/lib/system/route.ts
@@ -15,7 +15,14 @@ import { once } from '@ember/runloop';
 import { classify } from '@ember/string';
 import { DEBUG } from '@glimmer/env';
 import { TemplateFactory } from '@glimmer/opcode-compiler';
-import { InternalRouteInfo, Route as IRoute, Transition, TransitionState } from 'router_js';
+import {
+  InternalRouteInfo,
+  PARAMS_SYMBOL,
+  Route as IRoute,
+  STATE_SYMBOL,
+  Transition,
+  TransitionState,
+} from 'router_js';
 import {
   calculateCacheKey,
   normalizeControllerQueryParams,
@@ -219,7 +226,7 @@ class Route extends EmberObject implements IRoute {
     }
 
     let transition = this._router._routerMicrolib.activeTransition;
-    let state = transition ? transition.state : this._router._routerMicrolib.state;
+    let state = transition ? transition[STATE_SYMBOL] : this._router._routerMicrolib.state;
 
     let fullName = route.fullRouteName;
     let params = assign({}, state!.params[fullName]);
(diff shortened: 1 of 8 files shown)


packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -219,7 +226,7 @@ class Route extends EmberObject implements IRoute {
     }
 
     let transition = this._router._routerMicrolib.activeTransition;
-    let state = transition ? transition.state : this._router._routerMicrolib.state;
+    let state = transition ? transition[STATE_SYMBOL] : this._router._routerMicrolib.state;
```
rwjblue: Need svelting here I think

packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -899,10 +906,10 @@ class Route extends EmberObject implements IRoute {
 
     if (transition) {
       // Update the model dep values used to calculate cache keys.
-      stashParamNames(this._router, transition.state!.routeInfos);
+      stashParamNames(this._router, transition[STATE_SYMBOL]!.routeInfos);
```
rwjblue: svelting?

packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -899,10 +906,10 @@ class Route extends EmberObject implements IRoute {
...
 
       let cache = this._bucketCache;
-      let params = transition.params;
+      let params = transition[PARAMS_SYMBOL];
```
rwjblue: Svelting?

packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -1152,7 +1159,7 @@ class Route extends EmberObject implements IRoute {
         if (transition.resolveIndex < 1) {
           return;
         }
-        return transition.state!.routeInfos[transition.resolveIndex - 1].context;
+        return transition[STATE_SYMBOL]!.routeInfos[transition.resolveIndex - 1].context;
```
rwjblue: Svelting? What do we return after the deprecation?

packages/@ember/-internals/routing/lib/system/router.ts
```diff
@@ -73,6 +76,56 @@ function defaultWillTransition(
   }
 }
 
+if (TRANSITION_STATE) {
```
rwjblue: svelte

packages/@ember/-internals/routing/lib/system/route.ts
chadhietala: I think you might be confused. `transition.state` is still needed internally but access to it from application code should be deprecated and we just tell people to use the `RouteInfo`. All of these symbols are just allowing the integration of ember-router and router.js to work. I could have simply renamed the method but I felt that using a faux-symbol may deter people further. In the future we should just make these real symbols.

packages/@ember/-internals/routing/lib/system/router.ts
chadhietala: This is a svelte import.

packages/@ember/-internals/routing/lib/system/route.ts
rwjblue: 👍 	thank you!

_some_ of the comments though are specifically about deprecations, and should still apply

packages/@ember/-internals/routing/lib/system/router.ts
rwjblue: DOH!

PR #17043, opened 2018-10-04
Make transition a public interface

chadhietala:


rwjblue (commented):
Seems good to me!

We should get one of the @emberjs/learning-team-managers to double check that this works well with our yuidoc post-processing.

auvipy (approved):

jenweber:
Just looking at the diff compared to 3.4, seems fine! Since it already shows up in 3.4 when you check the “private” filter, it should show up by default now. Thanks for this fix.

Issue #17016, opened 2018-09-28
[CI] Investiate testem.js error


chancancode:
Seeing these on latest master test runs:

```
[0927/165414.959275:INFO:CONSOLE(0)] "Refused to execute script from 'http://localhost:13141/testem.js' because its MIME type ('text/html') is not executable, and strict MIME type checking is enabled.", source: http://localhost:13141/tests/?package=@ember/application&dist=es (0)
```

We should investigate and fix the MIME type thing (which is probably a bug in one or more upstream projects). Since things are still running "fine" we probably don't even need the script in the first place and just need to figure out how to remove it.

KamiKillertO:
I'll have a look

PR #17044, opened 2018-10-05
[CI] Remove testem.js error

KamiKillertO:
This PR is related to this issue #17016.
As suggested I just remove the import of testem.js in the file `tests/index.html`and everything seems to works in local, all tests are green. With this PR we will see if everything still work as expected on travis

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -281,6 +281,12 @@ module.exports = function() {
     trees.push(emberTemplateCompilerBundle);
   }
 
+  let emberTestsEmptyTestem = new Funnel('tests', {
+    files: ['testem.js'],
+    destDir: '',
+    annotation: 'tests/testem.js',
+  });
+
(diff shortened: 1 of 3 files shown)


rwjblue:
FWIW - This likely makes `ember test` not work, can you double check?

KamiKillertO:
Ok my bad I was running the tests using `yarn test`, but with `ember test` the tests fail because testem is missing. 🤔 I'll explore other solutions ^^

rwjblue:
No worries! You can probably work around the error by adding an empty `testem.js` file to the build. That way, when using `ember test` The “real” `testem.js` would be found, but when running via `yarn test` (which is effectively just `ember serve` + some puppeteering) it would avoid the error because this new empty file would be found.

rwjblue (commented):
Awesome! Thank you!

PR #17054, opened 2018-10-08
[FEATURE array-helper] Implement array helper RFC #318

### LINKED ISSUE #17052: Implement + enable `array` Helper Feature
This is a tracking issue to enable the `array` helper as proposed in https://github.com/emberjs/rfcs/pull/318.

- [x] Implement
- [x] API Docs
- [x] Guides
- [x] Enable by default
- [ ] Author a polyfill (e.g. `ember-array-helper-polyfill`)?

josemarluedke:
This implements the array helper.
RFC: https://github.com/emberjs/rfcs/pull/318

Closes #17052.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/helpers/array.ts
+++ b/packages/@ember/-internals/glimmer/lib/helpers/array.ts
@@ -1,0 +1,40 @@
+import { PathReference } from '@glimmer/reference';
+import { Arguments, VM } from '@glimmer/runtime';
+import { Opaque } from '@glimmer/util';
+
+/**
+@module ember
+*/
+
+/**
+   Use the `{{array}}` helper to create an array to pass as an option to your
+   components.
+
+   ```handlebars
+   {{my-component people=(array
+     'Tom Dade'
+     'Yehuda Katz'
+     this.myOtherPerson)
+   }}
+   ```
+
+   Would result in an object such as:
+
+   ```js
+   ['Tom Date', 'Yehuda Katz', this.get('myOtherPerson')]
+   ```
+
+   Where the 3rd item in the array is bound to updates of the `myOtherPerson` property.
+
+   @method array
+   @for Ember.Templates.helpers
+   @param {Array} options
+   @return {Array} Array
+   @category array-helper
+   @since 3.7.0
+   @public
+ */
+
+export default function(_vm: VM, args: Arguments): PathReference<Opaque[]> {
+  return args.positional.capture();
+}
--- a/packages/@ember/-internals/glimmer/tests/integration/helpers/array-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/helpers/array-test.js
@@ -1,0 +1,287 @@
+import { RenderingTest, moduleFor } from '../../utils/test-case';
+import { strip } from '../../utils/abstract-test-case';
+import { Component } from '../../utils/helpers';
+import { set } from '@ember/-internals/metal';
+
+moduleFor(
+  'Helpers test: {{array}}',
+  class extends RenderingTest {
+    ['@test returns an array']() {
+      this.render(strip`
+        {{#with (array "Sergio") as |people|}}
+          {{#each people as |personName|}}
+            {{personName}}
+          {{/each}}
+        {{/with}}`);
+
+      this.assertText('Sergio');
+
+      this.assertStableRerender();
+    }
+
+    ['@test can have more than one value']() {
+      this.render(strip`
+        {{#with (array "Sergio" "Robert") as |people|}}
+          {{#each people as |personName|}}
+            {{personName}},
+          {{/each}}
+        {{/with}}`);
+
+      this.assertText('Sergio,Robert,');
+
+      this.assertStableRerender();
+    }
+
+    ['@test binds values when variables are used']() {
+      this.render(
+        strip`{{#with (array personOne) as |people|}}
+                {{#each people as |personName|}}
+                  {{personName}}
+                {{/each}}
+              {{/with}}`,
+        {
+          personOne: 'Tom',
+        }
+      );
+
+      this.assertText('Tom');
+
+      this.assertStableRerender();
+
+      this.runTask(() => set(this.context, 'personOne', 'Yehuda'));
+      this.assertText('Yehuda');
+
+      this.runTask(() => set(this.context, 'personOne', 'Tom'));
+      this.assertText('Tom');
+    }
+
+    ['@test binds multiple values when variables are used']() {
+      this.render(
+        strip`{{#with (array personOne personTwo) as |people|}}
+                {{#each people as |personName|}}
+                  {{personName}},
+                {{/each}}
+              {{/with}}`,
+        {
+          personOne: 'Tom',
+          personTwo: 'Yehuda',
+        }
+      );
+
+      this.assertText('Tom,Yehuda,');
+
+      this.assertStableRerender();
+
+      this.runTask(() => set(this.context, 'personOne', 'Sergio'));
+
+      this.assertText('Sergio,Yehuda,');
+
+      this.runTask(() => set(this.context, 'personTwo', 'Tom'));
+
+      this.assertText('Sergio,Tom,');
+
+      this.runTask(() => {
+        set(this.context, 'personOne', 'Tom');
+        set(this.context, 'personTwo', 'Yehuda');
+      });
+
+      this.assertText('Tom,Yehuda,');
+    }
+
+    ['@test array helpers can be nested']() {
+      this.render(
+        strip`{{#with (array (array personOne personTwo)) as |listOfPeople|}}
+                {{#each listOfPeople as |people|}}
+                  List:
+                  {{#each people as |personName|}}
+                    {{personName}},
+                  {{/each}}
+                {{/each}}
+              {{/with}}`,
+        {
+          personOne: 'Tom',
+          personTwo: 'Yehuda',
+        }
+      );
+
+      this.assertText('List:Tom,Yehuda,');
+
+      this.assertStableRerender();
+
+      this.runTask(() => set(this.context, 'personOne', 'Chad'));
+
+      this.assertText('List:Chad,Yehuda,');
+
+      this.runTask(() => set(this.context, 'personTwo', 'Balint'));
+
+      this.assertText('List:Chad,Balint,');
+
+      this.runTask(() => {
+        set(this.context, 'personOne', 'Tom');
+        set(this.context, 'personTwo', 'Yehuda');
+      });
+
+      this.assertText('List:Tom,Yehuda,');
+    }
+
+    ['@test should yield hash of an array of internal properties']() {
+      let fooBarInstance;
+      let FooBarComponent = Component.extend({
+        init() {
+          this._super();
+          fooBarInstance = this;
+          this.model = { personOne: 'Chad' };
+        },
+      });
+
+      this.registerComponent('foo-bar', {
+        ComponentClass: FooBarComponent,
+        template: `{{yield (hash people=(array model.personOne))}}`,
+      });
+
+      this.render(strip`
+        {{#foo-bar as |values|}}
+          {{#each values.people as |personName|}}
+            {{personName}}
+          {{/each}}
+        {{/foo-bar}}`);
+
+      this.assertText('Chad');
+
+      this.assertStableRerender();
+
+      this.runTask(() => set(fooBarInstance, 'model.personOne', 'Godfrey'));
+
+      this.assertText('Godfrey');
+
+      this.runTask(() => set(fooBarInstance, 'model', { personOne: 'Chad' }));
+
+      this.assertText('Chad');
+
+      this.runTask(() => set(fooBarInstance, 'model.personOne', 'Godfrey'));
+
+      this.assertText('Godfrey');
+    }
+
+    ['@test should yield hash of an array of internal and external properties']() {
+      let fooBarInstance;
+      let FooBarComponent = Component.extend({
+        init() {
+          this._super();
+          fooBarInstance = this;
+          this.model = { personOne: 'Chad' };
+        },
+      });
+
+      this.registerComponent('foo-bar', {
+        ComponentClass: FooBarComponent,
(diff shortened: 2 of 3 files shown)


packages/@ember/-internals/glimmer/lib/helpers/array.ts
```diff
@@ -0,0 +1,37 @@
...
+   ['Tom Date', 'Yehuda Katz', this.get('myOtherPerson')]
+   ```
+
+   Where the 3rd item in the array is bound to updates of the `myOtherPerson` property.
+
+   @method array
+   @for Ember.Templates.helpers
+   @param {Object} options
+   @return {Object} Array
+   @since 2.6.0
```
rwjblue: Seems wrong, possibly 3.7?

Also can you add `@category` + your feature flag to the docs, that lets us filter out API docs for unlanded features...

packages/@ember/-internals/glimmer/lib/helpers/array.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+   Use the `{{array}}` helper to create an array to pass as an option to your
+   components.
+
+   ```handlebars
+   {{my-component people=(array
+     'Tom Dade'
+     'Yehuda Katz'
+     myOtherPerson)
```
rwjblue: Lets use `this.myOtherPerson`

packages/@ember/-internals/glimmer/lib/helpers/array.ts
```diff
@@ -0,0 +1,37 @@
...
+   ```js
+   ['Tom Date', 'Yehuda Katz', this.get('myOtherPerson')]
+   ```
+
+   Where the 3rd item in the array is bound to updates of the `myOtherPerson` property.
+
+   @method array
+   @for Ember.Templates.helpers
+   @param {Object} options
+   @return {Object} Array
```
rwjblue: Should it be `@return {Array}`?

packages/@ember/-internals/glimmer/lib/helpers/array.ts
```diff
@@ -0,0 +1,37 @@
...
+
+   ```js
+   ['Tom Date', 'Yehuda Katz', this.get('myOtherPerson')]
+   ```
+
+   Where the 3rd item in the array is bound to updates of the `myOtherPerson` property.
+
+   @method array
+   @for Ember.Templates.helpers
+   @param {Object} options
```
rwjblue: I think we receive an array right?

packages/@ember/-internals/glimmer/tests/integration/helpers/array-test.js
```diff
@@ -0,0 +1,202 @@
...
+    ['@test returns an array']() {
+      this.render(strip`{{#with (array \"Sergio\") as |people|}}
+                          {{#each people as |personName|}}
+                            {{personName}}
+                          {{/each}}
+                        {{/with}}`);
+
+      this.assertText('Sergio');
+
+      this.runTask(() => this.rerender());
```
rwjblue: I think we can use `this.assertStableRerender()`?

packages/@ember/-internals/glimmer/tests/integration/helpers/array-test.js
```diff
@@ -0,0 +1,202 @@
...
+
+      this.assertText('Sergio');
+    }
+
+    ['@test can have more than one value']() {
+      this.render(strip`{{#with (array \"Sergio\" \"Robert\") as |people|}}
```
rwjblue: given the template string, I don't think we need to escape the quotes

packages/@ember/-internals/glimmer/tests/integration/helpers/array-test.js
```diff
@@ -0,0 +1,202 @@
+import { RenderingTest, moduleFor } from '../../utils/test-case';
+import { strip } from '../../utils/abstract-test-case';
+import { Component } from '../../utils/helpers';
+import { set } from '@ember/-internals/metal';
+
+moduleFor(
+  'Helpers test: {{array}}',
+  class extends RenderingTest {
+    ['@test returns an array']() {
+      this.render(strip`{{#with (array \"Sergio\") as |people|}}
```
rwjblue: Probably don't need to escape the quotes

packages/@ember/-internals/glimmer/tests/integration/helpers/array-test.js
```diff
@@ -0,0 +1,202 @@
...
+          personOne: 'Tom',
+          personTwo: 'Yehuda',
+        }
+      );
+
+      this.assertText('Tom,Yehuda,');
+
+      this.runTask(() => this.rerender());
+
+      this.assertText('Tom,Yehuda,');
```
rwjblue: Can probably replace the runTask and assertText with `this.assertStableRerender()`

packages/@ember/-internals/glimmer/tests/integration/helpers/array-test.js
```diff
@@ -0,0 +1,202 @@
...
+          personOne: 'Tom',
+          personTwo: 'Yehuda',
+        }
+      );
+
+      this.assertText('List:Tom,Yehuda,');
+
+      this.runTask(() => this.rerender());
+
+      this.assertText('List:Tom,Yehuda,');
```
rwjblue: Same as above, RE: stable re-render

rwjblue (commented):
Looking good! I left some pretty nit-picky comments inline (sorry about that), but also would like to see a few more tests:

* Confirming that passing an `array` result into a component invocation, and _that_ component using `{{#each` properly rerenders/updates as items change
* Checking that the output of `array` follows the "immutable" style by having a component that we control which is passed an `array` and we inspect it for `===` equality.

packages/@ember/-internals/glimmer/lib/helpers/array.ts
josemarluedke: I was wondering about that. Should the templates in the test also refer to `this` where there is a backing component? Also the `this.render` calls, are the second parameter considered arguments? I could change the tests to use `this.` and also `@`.

rwjblue (approved):

packages/@ember/-internals/glimmer/lib/helpers/array.ts
```diff
@@ -0,0 +1,38 @@
...
+   ```
+
+   Where the 3rd item in the array is bound to updates of the `myOtherPerson` property.
+
+   @method array
+   @for Ember.Templates.helpers
+   @param {Array} options
+   @return {Array} Array
+   @category array-helper
+   @since 2.7.0
```
btecu: This should be `3.6.0` at the earliest

packages/@ember/-internals/glimmer/lib/helpers/array.ts
```diff
@@ -0,0 +1,38 @@
...
+   @method array
+   @for Ember.Templates.helpers
+   @param {Array} options
+   @return {Array} Array
+   @category array-helper
+   @since 3.7.0
+   @public
+ */
+
+export default function(_vm: VM, args: Arguments) {
```
chancancode: Need return type

chancancode:
Should we debug freeze the array? The empty case is frozen like hashes iirc

packages/@ember/-internals/glimmer/lib/helpers/array.ts
josemarluedke: Would that be `any[]`?

josemarluedke: It seems seem to like that change alone.

screen shot 2018-10-10 at 11 00 56 am

The only way I found to make this work is by casting to any.

```ts
export default function(_vm: VM, args: Arguments): any[] {
  return args.positional.capture() as any;
}
```

rwjblue:
> Should we debug freeze the array? The empty case is frozen like hashes iirc

Yes, probably.

packages/@ember/-internals/glimmer/lib/helpers/array.ts
chancancode: I believe it should be `PathReference<Opaque>`, or `PathReference<Opaque[]>` if we want to be precise

chancancode:
Unfortunately debug freezing is not going to be super easy. We can merge this as-is and do a follow up PR for that. I also just realized that the current version of the glimmer code seems to always return a new array/object even for the empty case.

We can do something like:

```
class DebugFreezeReference<T> implements PathReference<T> {
  constructor(private inner: PathReference<T>) {
    this.tag = inner.tag;
  }

  value(): T {
    return Object.freeze(this.inner.value());
  }

  get(path: string): PathReference<Opaque> {
    return this.inner.get(path); // or should this be `new DebugFreezeReference(this.inner.get(path))`?
  }
}
```

packages/@ember/-internals/glimmer/lib/helpers/array.ts
chancancode: Can you also fix `hash.ts` to have a return type of `PathReference<Opaque>`? (or `PathReference<Dict<Opaque>>`)

josemarluedke: Pushed the change. I will open a new PR for the return type of hash helper to keep it separated.

packages/@ember/-internals/glimmer/tests/integration/helpers/array-test.js
```diff
@@ -0,0 +1,236 @@
...
+
+moduleFor(
+  'Helpers test: {{array}}',
+  class extends RenderingTest {
+    ['@test returns an array']() {
+      this.render(strip`{{#with (array "Sergio") as |people|}}
+                          {{#each people as |personName|}}
+                            {{personName}}
+                          {{/each}}
+                        {{/with}}`);
```
chancancode: does the linter complain if you do something like...

```
this.render(strip`
  {{#with (array "Sergio") as |people|}}
    {{#each people as |personName|}}
      {{personName}}
    {{/each}}
  {{/with}}`
);
```

packages/@ember/-internals/glimmer/tests/integration/helpers/array-test.js
```diff
@@ -0,0 +1,236 @@
...
+        }
+      );
+
+      this.assertText('Tom');
+
+      this.assertStableRerender();
+
+      this.runTask(() => set(this.context, 'personOne', 'Yehuda'));
+      this.assertText('Yehuda');
+    }
```
chancancode: Add another step to set it back to Tom (see #13127)

packages/@ember/-internals/glimmer/tests/integration/helpers/array-test.js
```diff
@@ -0,0 +1,236 @@
...
+      this.assertStableRerender();
+
+      this.runTask(() => set(this.context, 'personOne', 'Sergio'));
+
+      this.assertText('Sergio,Yehuda,');
+
+      this.runTask(() => set(this.context, 'personTwo', 'Tom'));
+
+      this.assertText('Sergio,Tom,');
+    }
```
chancancode: add another step to set it back to Tom/Yehuda (see #13127)

packages/@ember/-internals/glimmer/tests/integration/helpers/array-test.js
```diff
@@ -0,0 +1,236 @@
...
+
+      this.assertStableRerender();
+
+      this.runTask(() => set(this.context, 'personOne', 'Chad'));
+
+      this.assertText('List:Chad,Yehuda,');
+
+      this.runTask(() => set(this.context, 'personTwo', 'Balint'));
+
+      this.assertText('List:Chad,Balint,');
```
chancancode: add reset step

packages/@ember/-internals/glimmer/tests/integration/helpers/array-test.js
```diff
@@ -0,0 +1,236 @@
...
+
+      this.assertStableRerender();
+
+      this.runTask(() => set(fooBarInstance, 'model.personOne', 'Godfrey'));
+
+      this.assertText('Godfrey');
+
+      this.runTask(() => set(fooBarInstance, 'model', { personOne: 'Chad' }));
+
+      this.assertText('Chad');
```
chancancode: add reset step

packages/@ember/-internals/glimmer/tests/integration/helpers/array-test.js
```diff
@@ -0,0 +1,236 @@
...
+      this.render(strip`{{foo-bar people=(array "Tom" personTwo)}}`, { personTwo: 'Chad' });
+
+      this.assertText('Tom,Chad,');
+
+      this.assertStableRerender();
+
+      this.runTask(() => set(this.context, 'personTwo', 'Godfrey'));
+
+      this.assertText('Tom,Godfrey,');
+    }
```
chancancode: add reset step

chancancode:
Looking good. Probably should have a few more test testing "capturing" the array into JS values and make sure it is correct. Something like this:

```js
let captured;

this.registerHelper('capture', function([array]) {
  captured = array;
  return "captured";
});

this.render(`{{capture (array ...)}}`);

assert.deepEqual(captured, [...]);

this.runTask(this.set(...));

assert.deepEqual(captured, [...]);

...
```

Basically what you already did in the last test, but more coverage around the actual values, and that it updates correctly, etc. You can probably just merge that into your last test since it's already doing similar things.

josemarluedke:
I have addressed the last few comments from @chancancode. This should be ready for another review.

chancancode (approved):

chancancode:
@josemarluedke have you tried the indentation style in https://github.com/emberjs/ember.js/pull/17054#discussion_r224330568? If the linter prefer you write it like this, it's nbd

josemarluedke:
@chancancode Yes I did, I changed most of the code to that indentation style. I did not change all because some already [had a line break due](https://github.com/emberjs/ember.js/pull/17054/files#diff-938074d5f22da2cef01e6e0e38d85f36R182) to component args. But if you do think I should change that as well, I'm happy to do so.

rwjblue:
Thank you!

PR #17057, opened 2018-10-08
avoid calling `peekMeta` when meta does not exist

bekzod:
differentiate between not passing `meta` and  `meta` not existing.

many `meta` related functions allow passing `meta` as last argument and do check if `meta` passed as following:
```javascript
let meta = _meta === undefined ? peekMeta(obj) : _meta;
```
is such case `peekMeta()` is called even if last argument is passed as `undefined` since `meta` does not exist, to avoid it made `peekMeta` to return `null` instead of `undefined`


bekzod:
closing this for now, it is typescript lint nightmare

Issue #16989, opened 2018-09-19
Using input helper, how can we pass both the `value` and `event` to the action? 


elimau:
As per the ember latest (3.4), in the docs
https://guides.emberjs.com/release/templates/input-helpers/
it says if you set the value, the value will be sent to the action as the first param.
Otherwise, it will send the event as the first param.

However, I need both the `value` and the `event` to be passed to the action.
How can I get that to work with the latest ember?

The below code shows an example of a template code and component (js) code.

hbs:
`    {{input placeholder=placeholder class="type-in" autocomplete="off" value=value key-up="keyUp" key-down="keyDown" insert-newline="enter" bubbles=false }}
`

js:
`
    actions: {
        keyUp(text, event) { ... }
    }
`

The above code used to work in ember 2.9.0. Both the `value` and `event` used to get passed to the `keyUp` action. Now, I upgrade to ember 3.4, it only sends the `value`, the second param (event) is null.
Thanks for any help.

btecu:
I've also run into this in 3.4.

raphaelns-developer:
Same here. After updated to Ember 3.4 the `key-down` and `key-up` action stop to sending the `event`. I made a simple twiddle to show the error: https://ember-twiddle.com/198760441072c804776732de6b68954b?openFiles=controllers.application.js%2C

rwjblue:
the patch already landed, and will be included in the next 3.4.x version (hopefully releasing today/tomorrow)

PR #17115, opened 2018-10-11
[BUGFIX lts] Pass the event parameter to sendAction

ssutar:
This fixes the regression introduced in 3.4.x, where the `input` actions were not passed with event object as parameter. 

For example
```
{{input key-down="keyDown"}}
```
And
```
...
actions: {
  keyDown(value, event) {
    if (event.keyCode === SOME_KEY_CODE) {
      // DOES NOT WORK starting 3.4.x since `event` is undefined
    }
  }
}
```
For more details: https://github.com/emberjs/ember.js/issues/16989

Please note that passing actions like `key-down="keyDown"` is deprecated, the closure actions should be used instead like `key-down=action("keyDown")`

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/helpers/input-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/helpers/input-test.js
@@ -3,6 +3,7 @@ import { set } from '@ember/-internals/metal';
 import { Component } from '../../utils/helpers';
 import { RenderingTest, moduleFor } from '../../utils/test-case';
 import { runDestroy } from 'internal-test-helpers';
+import { jQuery } from '@ember/-internals/views';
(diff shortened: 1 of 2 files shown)


rwjblue:
I'm not sure I follow this. We generally haven't exposed the event before, right? Why change it now?

rwjblue:
Based on the comments in #16989 this is a regression (though it generally surprises me). We need to track down the source (and see what versions it affects).

ssutar:
From my investigation this is happening starting from `3.4.x`

rwjblue:
Chatted with @ssutar, he tracked down the regression to https://github.com/emberjs/ember.js/pull/16744. /cc @cibernox

PR #17126, opened 2018-10-17
Update to ember-cli@3.5.0.

rwjblue:
Brings loads of changes/fixes, but the reason I did it was to start using Broccoli 2...

### DIFF
--- a/node-tests/blueprints/acceptance-test-test.js
+++ b/node-tests/blueprints/acceptance-test-test.js
@@ -23,20 +23,24 @@ describe('Blueprint: acceptance-test', function() {
     it('acceptance-test foo', function() {
       return emberGenerateDestroy(['acceptance-test', 'foo'], _file => {
         expect(_file('tests/acceptance/foo-test.js')).to.equal(
-          fixture('acceptance-test/default.js')
+          fixture('acceptance-test/qunit-rfc268.js')
         );
       });
     });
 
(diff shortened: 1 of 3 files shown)


locks (approved):

rwjblue:
Yeah, its because the tests are setup to expect the wrong thing by default `ember new foo`. Its going to be a bit of a chore to fix 😩.

rwjblue:
I pushed one commit with fixes for `acceptance-test` file to show what I mean.

Turbo87:
oh... that makes sense... 😞

ppcano:
@simonihmig Will #17207 and #17198 finally unblock the `ember-cli@3.5` upgrade or there are something else that needs to be done.

I am working on the MU blueprints and I would like to test with new versions of `ember-cli` and `ember-cli-blueprint-test-helpers`.

simonihmig:
> Will #17207 and #17198 finally unblock the ember-cli@3.5 upgrade or there are something else that needs to be done.

@ppcano I guess there are a few more blueprints that need to be "fixed", which I haven't touched yet (e.g. mixins). I believe making each single test explicit about the test framework version used is the key here, as in this example: https://github.com/emberjs/ember.js/pull/17207/files#diff-b1525449225102a81123371a3a8c87f8R26

ppcano:
#17212 supersedes this PR

rwjblue:
Thank you @ppcano!

PR #17128, opened 2018-10-18
[BUGFIX lts] build: Turn off sourcemaps for TS-to-ES compilation

### LINKED ISSUE #16881: dist/ember.prod.js has multiple sourcemap declarations, breaking sourcemap build
## Summary

The current npm version of `ember-source`, 3.3.1 (**2018-09-07 update: issue remains in 3.4.0**), contains the file `dist/ember.prod.js`. This file contains multiple declarations of the form:

```
    //# sourceMappingURL=FILENAME.js.map
```

According to the [source map spec](https://sourcemaps.info/spec.html#h.lmz475t4mvbx), there should only be one line of this form, at the end of the file.

## Impact

This means any `vendor.js` file built with ember-cli will not have a valid  sourcemap file.

Please see https://github.com/emberjs/ember.js/issues/16844 for the error (the bug was erroneously closed as the issue was assumed to be with ember-cli, not ember-source).

## Cause

Including multiple `sourceMappingURL` declarations in `ember.prod.js` causes issues for fast-sourcemap-concat ([/lib/source-map.js#L91](https://github.com/ef4/fast-sourcemap-concat/blob/master/lib/source-map.js#L91)), since the `source-map-url` library only removes the first one.

This is valid behaviour since there should only be one `sourceMappingURL` declaration per file.

## Suggested fix

Remove the extra `sourceMappingURL` lines from `dist/ember.prod.js`.

## References and similar issues

https://github.com/broccolijs/broccoli-concat/issues/123
https://github.com/ef4/fast-sourcemap-concat/pull/54
https://github.com/ember-cli/ember-fetch/pull/120

Turbo87:
We seem to use Babel to compile ES to AMD modules and that now includes ES modules there were originally written in TypeScript. The TypeScript compiler is apparently outputting sourcemaps by default, but it looks like Babel can't handle those properly. Instead Babel keeps the sourcemap comments inside of the AMD modules, which causes the final concatenated file to have *multiple* sourcemap comments, which is invalid. Future versions of Babel might fix this but for now we should turn off the TypeScript sourcemaps, so that the final build output is valid again.

Resolves #16881 

/cc @rwjblue @buschtoens @gossi

### DIFF
--- a/tests/node/sourcemap-test.js
+++ b/tests/node/sourcemap-test.js
@@ -1,0 +1,27 @@
+var fs = require('fs');
+
+QUnit.module('sourcemap validation', function() {
+  var assets = ['ember.debug', 'ember.prod', 'ember.min'];
+
+  assets.forEach(asset => {
+    QUnit.test(`${asset} has only a single sourcemaps comment`, function(assert) {
+      var jsPath = `dist/${asset}.js`;
+      assert.ok(fs.existsSync(jsPath));
+
+      var contents = fs.readFileSync(jsPath, 'utf-8');
+      var num = count(contents, '//# sourceMappingURL=');
+      assert.equal(num, 1);
+    });
+  });
+});
+
+function count(source, find) {
+  var num = 0;
+
+  var i = -1;
+  while ((i = source.indexOf(find, i + 1)) !== -1) {
+    num += 1;
+  }
+
+  return num;
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,7 +83,9 @@ jobs:
       - TEST_SUITE=built-tests
       - EMBER_ENV=production
     - env: TEST_SUITE=old-jquery-and-extend-prototypes
-    - env: TEST_SUITE=node
+    - env:
+      - TEST_SUITE=node
+      - EMBER_ENV=production
       node_js: "6"
     - env: TEST_SUITE=blueprints
       node_js: "6"
--- a/broccoli/packages.js
+++ b/broccoli/packages.js
@@ -87,7 +87,11 @@ module.exports.getPackagesES = function getPackagesES() {
     include: ['**/*.ts'],
   });
 
-  let typescriptCompiled = typescript(debugTree(typescriptContents, `get-packages-es:ts:input`));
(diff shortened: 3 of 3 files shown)


rwjblue:
I mentioned this in discord chat, but would you mind bringing the test from https://github.com/emberjs/ember.js/pull/16987/commits/5e1f2f332efc5a8b69898c4c336e1752500ecf70 over? I really would like to do _something_ to ensure we don't regress this in the future...

Turbo87:
@rwjblue using `sourcemap-validator` did not help, as that only validates the *content* of the sourcemap, but not if it's correctly specified in the original JS code. I've added a test that explicitly checks that there is only ever exactly *one* sourcemap comment in the final assets.

rwjblue (commented):
Thank you @Turbo87

tests/node/sourcemap-test.js
```diff
@@ -0,0 +1,27 @@
+var fs = require('fs');
+
+QUnit.module('sourcemap validation', function() {
+  var assets = ['ember.debug', 'ember.prod', 'ember.min'];
+
+  assets.forEach(asset => {
+    QUnit.test(`${asset} has only a single sourcemaps comment`, function(assert) {
+      var jsPath = `dist/${asset}.js`;
+      assert.ok(fs.existsSync(jsPath));
```
rwjblue: The node tests are failing because this apparently runs in debug builds (not prod builds) so `.prod.js` and `.min.js` files are not present.

Can you tweak https://github.com/emberjs/ember.js/blob/8324f807e72f7facc258f8afa785eec6e993b532/.travis.yml#L86-L87 to:

```yml
- env:
  - TEST_SUITE=node
  - EMBER_ENV=production
  node_js: "6"
```

Turbo87:
@rwjblue looks like CI is not building the production assets. should I modify the tests to skip then?

rwjblue (changes requested):

rwjblue (approved):

buschtoens:
Thank you so much, @Turbo87 & @rwjblue ❤️

PR #17131, opened 2018-10-19
Remove deprecated 'targetObject'

btecu:


packages/@ember/-internals/glimmer/lib/component-managers/curly.ts
```diff
@@ -246,8 +246,8 @@ export default class CurlyComponentManager
     props[HAS_BLOCK] = hasBlock;
 
     // Save the current `this` context of the template as the component's
-    // `_targetObject`, so bubbled actions are routed to the right place.
-    props._targetObject = callerSelfRef.value();
+    // `_target`, so bubbled actions are routed to the right place.
+    props._target = callerSelfRef.value();
```
rwjblue: Should this just be `target`? If we use `target` it seems like we don't have to add an extra return in `target_action_support.js` below...

btecu: I believe some tests were failing when I initially tried that. Let me check again.

rwjblue (approved):

Issue #17121, opened 2018-10-16
local variables should always win over helper


chancancode:
We tried to fix this in #14520, but apparently we didn't catch all the cases.

Reproduction: https://ember-twiddle.com/dcce0cd429ceb1d454d62988937bfd15?openFiles=templates.application.hbs%2C

Options are:

1. just fix it
2. error for one cycle then fix
3. deprecate for one LTS cycle then fix

chancancode:
imo, these are also bugs but in the sub-expression positions, but that's probably more controversial: https://ember-twiddle.com/95790479cef1d910bdc12ffcfc446149?openFiles=templates.application.hbs%2C (cc @wycats @mmun)

chancancode:
Not sure if the approach is still good for the current glimmer codebase, but here was the last patch that fixed the `{{foo}}` cases: https://github.com/glimmerjs/glimmer-vm/pull/335

PR #17132, opened 2018-10-19
[BUGFIX beta] Assert when local variables shadow helper invocations

chancancode:
```hbs
{{#let this.foo as |foo|}}
  {{concat (foo)}}
            ~~~ shadowed helper invocation!
{{/let}}
```

Previously, this would have tried to resolve and invoke the helper `foo`, ignoring the presence of the `foo` local variable. This is inconsistent with our general lexical lookup rules. This is now an error (assertion).

This paves the way for allowing "contextual helpers" in the future, where the local variable `foo` contains an invocable helper value.

Partially addresses #17121, allowing the rest of the bugs to be fixed in Glimmer VM.


rwjblue (approved):

rwjblue:
Looks like we have some linting failures...

chancancode:
The linter wasn't happy with my Artisanal™ indentation. Not super happy with its suggestions but the linter is going to spend a lot more time looking at this file than me, so I could live with it :shrug:

Issue #16510, opened 2018-04-12
[regression] Closure component dot rule is unenforced


mixonic:
Closure component invocations follow a "dot rule". The can be invoked either with the `{{component}}` helper, or with a path containing a `.`.

For example these are **valid** invocations:

```hbs
{{#some-component as |wrappingHash|}}
  {{wrappingHash.someComponent}}
{{/some-component}}

{{this.somePassedInComponent}}

{{component somePassedInComponent}}
```

For another example these are **invalid** invocations (should not invoke the component, usually they show `[object Object]`):

```hbs
{{#some-component as |closureComponent|}}
  {{closureComponent}}
{{/some-component}}

{{somePassedInComponent}}
```

As of Ember 2.15 invocations of block param closure components without a `.` has been permitted. A reproduction: https://ember-twiddle.com/92561d3be8128785682b50666e05c25f?fileTreeShown=false&openFiles=templates.application.hbs%2C

/cc @iezer @rwjblue

chancancode:
dot rule was removed in #17135

PR #17135, opened 2018-10-19
[BUGFIX beta] Local variables should win over helpers

### LINKED ISSUE #17121: local variables should always win over helper

chancancode:
```hbs
{{#let this.foo as |foo|}}
  {{foo 1 2 3}}
    ~~~ local variable invocation!
{{/let}}
```

Previously, this would have tried to resolve and invoke the helper `foo`, ignoring the presence of the `foo` local variable. This is inconsistent with our general lexical lookup rules. This will now invoke `foo` local variable as a contextual component.

In other words, this removes the remaining "dot rule" exceptions for contextual components, as per [RFC #311](https://github.com/emberjs/rfcs/blob/master/text/0311-angle-bracket-invocation.md):

> We propose to relax that rule to match the proposed angle bracket invocation semantics (i.e. allowing local variables without a dot, as well as `@names`, but disallowing implicit `this` lookup).

This commit also fixed another related issue:

```hbs
{{#let (hash foo="foo-bar") as |h|}}
  {{h.foo 1 2 3}}
    ~~~~~ this is a string, not a component!
{{/let}}
```

Previously, this would have tried to resolve and invoke the `foo-bar` component. This is an unintended consequence (i.e. a bug) of the "dot rule" implementation. This will now raise a `TypeError` in development mode and result in undefined behavior in production builds (probably some other runtime error deep inside the Glimmer VM internals).

Fixes #17121, #16510.


rwjblue (approved):

packages/@ember/-internals/glimmer/tests/integration/components/contextual-components-test.js
```diff
@@ -1387,6 +1387,98 @@ moduleFor(
...
+    }
+
+    ['@test RFC#311 invoking named args (with arguments)']() {
+      this.registerComponent('x-outer', { template: '{{@inner 1 2 3}}' });
+
+      this.registerComponent('x-inner', {
+        ComponentClass: Component.extend().reopenClass({
+          positionalParams: 'params',
+        }),
+        template: 'inner:{{#each params as |param|}} {{param}}{{/each}}'
```
rwjblue: Missing a trailing comma (fails the linting config)

PR #17138, opened 2018-10-20
use `assign` to copy values of queryParam

bekzod:


packages/@ember/-internals/glimmer/lib/components/link-to.ts
```diff
@@ -732,16 +733,8 @@ const LinkComponent = EmberComponent.extend({
...
-    }
-
-    let values = queryParams.values;
-    for (let key in values) {
-      if (!values.hasOwnProperty(key)) {
-        continue;
-      }
-      resolvedQueryParams[key] = values[key];
+    if (queryParams) {
+      resolvedQueryParams = assign(resolvedQueryParams, queryParams.values);
```
rwjblue: I don’t think we need assignment here since we are mutating `resolvedQueryParams`

bekzod:
https://github.com/emberjs/ember.js/pull/16957 is better approach

PR #17143, opened 2018-10-22
[FEATURE] Modifier Manager

chadhietala:
This is intended to implement [RFC#373](https://github.com/emberjs/rfcs/pull/373)


rwjblue (commented):
Overall this looks good, can you add a test for the non-string fallback version of `setComponentManager`?

rwjblue (approved):

Issue #14174, opened 2016-08-31
Correct Query String Serialization and Deserialization


nathanhammond:
This bug supersedes #13816 and #14171 (and who knows how many others). The behavior for serializing objects to query strings inside of Ember is wrong for at least two reasons:
1. It is inconsistent with literally every other implementation I've reviewed (~20 at this point).
2. It does not match the behavior of `route-recognizer`, which we also bundle.

This means that our behavior does not work on hard refresh and also triggers strange serialization inconsistencies when run through a `string` or `object` path. I'm writing a full blog post about query strings which will be published this week which will detail a path forward.
- [Implementation in Ember.](https://github.com/emberjs/ember.js/blob/master/packages/ember-routing/lib/system/route.js#L421-L464)
- [`route-recognizer`'s `parseQueryString`.](https://github.com/tildeio/route-recognizer/blob/master/lib/route-recognizer.js#L455-L483)
- [`route-recognizer`'s `generateQueryString`.](https://github.com/tildeio/route-recognizer/blob/master/lib/route-recognizer.js#L423-L453)

I'm of the opinion that we should not attempt patches to this behavior without solving the underlying problem as we will simply introduce new and different failure modes. Until we have a complete solution (which I'll write as part of landing the new `route-recognizer`) we should ride out our existing known behavior and new reports should be closed as a duplicate of this bug.

pixelhandler:
@nathanhammond sounds like you have a plan forward for the solution as part of the route-recognizer work, nice. Will you put that write up an an RFC?

webark:
@nathanhammond do you have a link to that blog post?

nathanhammond:
Path forward: https://github.com/nathanhammond/query-string-spec

cwarny:
Any update on this? My use case is an array query param getting stringified. Any elegant workaround in the meantime?

sarupbanskota:
@cwarny Have you tried initializing the property corresponding to the query param as `A()`? 

e.g
``` js
// within controller
  tags: A(),
  queryParams: ['tags']
```

robclancy:
Is this ever going to be fixed?

robclancy:
I'm just using this mixin now...

```
import Mixin from '@ember/object/mixin';
import { typeOf } from '@ember/utils';

function basicArray(array) {
  if (typeOf(array) !== 'array') {
    return false;
  }

  return array.every((value) => {
    return ['string', 'number'].includes(typeOf(value));
  });
}

[...]

PR #17151, opened 2018-10-25
Fix deserialization of array query params 

tchan:
When Ember serializes an Array type param, it does so with JSON.stringify.
The deserialize converts back to an Array, but if you have a **slow connection**, the serialize method gets called again before deserialize which results in re-stringifying a string. 

Visual summary of what happens under the hood
screenshot

I've reproduced this behaviour with a twiddle here https://ember-twiddle.com/bc34a1cdf99ce9a013ace8bca3d25162?openFiles=controllers.application.js%2C

This is not ideal but for the above twiddle to work you need to use the [allow-control-allow-origin* chrome extension](https://chrome.google.com/webstore/detail/allow-control-allow-origi/nlfbmbojpeacfghkpbjhddihlkkiljbi?hl=en) as I'm making an external ajax request each time the query params is updated. Otherwise you'll get a console error ` No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'null' is therefore not allowed access.`. Not sure how to bypass this otherwise.

Steps: 
1. Simulate a slow connection by applying network throttling. In the gif I used a custom profile on Chrome with 10kb/s and 20,000ms of added latency.
2. Click a checkbox (arrays query param)
3. Start typing in the search bar.

Gif of what happens to the URL due to deserializing getting called before serialize which is apparent with a really slow internet connection
emberarrays

Gif of what happens with a standard connection (expected behaviour)
normalconnection

Related issues: 
https://github.com/emberjs/ember.js/issues/14174
https://github.com/emberjs/ember.js/issues/13591
https://github.com/emberjs/ember.js/pull/14171

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/router.ts
+++ b/packages/@ember/-internals/routing/lib/system/router.ts
@@ -895,6 +895,8 @@ class EmberRouter extends EmberObject {
   _serializeQueryParam(value: unknown, type: string) {
     if (value === null || value === undefined) {
       return value;
+    } else if (typeof value === 'string' && type === 'array') {
+      value = JSON.parse(value);
     } else if (type === 'array') {
       return JSON.stringify(value);
     }


acorncom:
@tchan think you could add a test here? This is likely to be a future area of churn so might help to have your case covered so it doesn’t come up again ...

tpetrone:
Greetings, any updates on this?
Tks

tpetrone:
Tks @tchan !

@arthurmde do you think we can tackle it?

PR #17169, opened 2018-10-31
[BUGFIX] Adds default implementations of Component lifecycle hooks

### LINKED ISSUE #17156: TypeError: with Native Class syntax on certain lifecycle hooks when calling `super`
Converting an app over to use native classes and noticed some component hooks throw errors when calling their superclass equivalent hook. `didReceiveAttrs` is an example of one hook which will error 

@rwjblue helped answer my question today on Discord
```
rwjblue Today at 1:52 PM
its a bug
basically the parent class doesn't have a didReceiveAttrs hook implemented
```

---------

```
export default class MyClass extends Component {
  didReceiveAttrs() {
	// This will throw a TypeError
    super.didReceiveAttrs(...arguments);

    // ... other stuff
  }
}
```

```
TypeError: (intermediate value).didReceiveAttrs is not a function
    at ModalGenericFooter.didReceiveAttrs (generic-footer.js:21)
    at ModalGenericFooter.trigger (core_view.js:62)
    at ModalGenericFooter.superWrapper [as trigger] (ember-utils.js:308)
    at CurlyComponentManager.create (ember-glimmer.js:3697)
    at Object.evaluate (runtime.js:1261)
    at AppendOpcodes.evaluate (runtime.js:46)
    at LowLevelVM.evaluateSyscall (runtime.js:2895)
    at LowLevelVM.evaluateInner (runtime.js:2867)
    at LowLevelVM.evaluateOuter (runtime.js:2859)
    at VM.next (runtime.js:4787)
    at VM.execute (runtime.js:4772)
    at TryOpcode.handleException (runtime.js:3752)
    at UpdatingVMFrame.handleException (runtime.js:3922)
    at UpdatingVM._throw [as throw] (runtime.js:3657)
    at Assert.evaluate (runtime.js:730)
    at UpdatingVM.execute (runtime.js:3644)
    at RenderResult.rerender (runtime.js:3949)
    at RootState._this30.render (ember-glimmer.js:4075)
    at TransactionRunner.runInTransaction (ember-metal.js:406)
    at InteractiveRenderer._renderRoots (ember-glimmer.js:4340)
    at InteractiveRenderer._renderRootsTransaction (ember-glimmer.js:4372)
    at InteractiveRenderer._revalidate (ember-glimmer.js:4412)
    at invokeWithOnError (backburner.js:283)
    at Queue.flush (backburner.js:186)
    at DeferredActionQueues.flush (backburner.js:353)
    at Backburner._end (backburner.js:808)
    at Backburner.end (backburner.js:558)
    at Backburner._run (backburner.js:847)
    at Backburner._join (backburner.js:829)
    at Backburner.join (backburner.js:612)
    at join (index.js:164)
    at has_element.js:18
    at flaggedInstrument (index.js:119)
    at Object.handleEvent (has_element.js:17)
    at Class.handleEvent (view_support.js:402)
    at HTMLButtonElement.<anonymous> (event_dispatcher.js:327)
    at HTMLBodyElement.dispatch (jquery.js:5183)
    at HTMLBodyElement.elemData.handle (jquery.js:4991)
```

pzuraq:
Adds default implementations for all of `Ember.Component`'s lifecycle hooks. Fixes #17156.

There are some notable concerns here:

1. Performance, default implementations means that we will be doing slightly more work in cases where users are calling `_super` in a hook which did not have a default implementation. This may be worth benchmarking, but I'm unsure if it would make that much of a difference.
2. The `didDestroyElement` hook was not previously documented, and may not have been considered public API. It is under test, and there are no internal usages of it besides test, so it may be a better idea to not publicly expose it and deprecate/remove it instead.

The alternative would be to remove lifecycle hook defaults entirely. As long as we're consistent in telling users "always call `super`" or "never call `super` unless its your own code" I think it would be fine. The current situation, where we sometimes-do-and-sometimes-don't, is the worst of both worlds. Removing lifecycle hooks entirely may require a deprecation.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/component.ts
+++ b/packages/@ember/-internals/glimmer/lib/component.ts
@@ -751,6 +751,7 @@ const Component = CoreView.extend(
      @public
      @since 1.13.0
      */
+    didReceiveAttrs() {},
 
     /**
      Called when the attributes passed into the component have been updated.
@@ -769,6 +770,7 @@ const Component = CoreView.extend(
      @public
      @since 1.13.0
      */
+    didRender() {},
 
     /**
      Called after a component has been rendered, both on initial render and
@@ -785,6 +787,7 @@ const Component = CoreView.extend(
      @public
      @since 1.13.0
      */
+    willRender() {},
 
     /**
      Called before a component has been rendered, both on initial render and
@@ -801,6 +804,7 @@ const Component = CoreView.extend(
      @public
      @since 1.13.0
      */
+    didUpdateAttrs() {},
 
     /**
      Called when the attributes passed into the component have been changed.
@@ -817,6 +821,7 @@ const Component = CoreView.extend(
      @public
      @since 1.13.0
      */
+    willUpdate() {},
 
     /**
      Called when the component is about to update and rerender itself.
@@ -833,6 +838,7 @@ const Component = CoreView.extend(
      @public
      @since 1.13.0
      */
+    didUpdate() {},
 
     /**
      Called when the component has updated and rerendered itself.
--- a/packages/@ember/-internals/glimmer/tests/integration/components/life-cycle-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/components/life-cycle-test.js
@@ -1571,6 +1571,24 @@ moduleFor(
         ]
       );
     }
+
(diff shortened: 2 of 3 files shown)


pzuraq:
cc @rwjblue

rwjblue (approved):

PR #17170, opened 2018-11-01
[FEATURE] allow to disable mandatory setters

wojciechb:
Hi all, :) 

This PR adds new ENV field, to allow us to disable mandatory setters (the ability to do so was removed in this PR: https://github.com/emberjs/ember.js/pull/16538/files and prevented us from updating to Ember 3.4 LTS). 

The reason for that is, that we have a large shared library of complex non-ember domain models, which needs to be integrated into Ember app.  Mainly due to performance reasons (and also complexity) of alternative solutions like proxying objects, we used to integrate them using notifyPropertyChange (which is part of public Ember API - https://www.emberjs.com/api/ember/3.5/classes/Observable/methods/notifyPropertyChange?anchor=notifyPropertyChange). Due to removal of that feature flag we cannot use notifyPropertyChange anymore.

 I guess this isn’t a common use case, but without that, we would be forced to rewrite loads of stuff.

### DIFF
--- a/packages/ember-environment/lib/env.ts
+++ b/packages/ember-environment/lib/env.ts
@@ -63,6 +63,19 @@ export const ENV = {
   */
   LOG_VERSION: true,
 
+  /**
+   The `MANDATORY_SETTERS` property, when true, tells Ember
+   that setting observed values needs to be done by Ember.set
+   Added to allow notifyPropertyChange still be used.
+
+   @property MANDATORY_SETTERS
+   @type Boolean
+   @default true
+   @for EmberENV
+   @public
+   */
+  MANDATORY_SETTERS: true,
+
   RAISE_ON_DEPRECATION: false,
 
   STRUCTURED_PROFILE: false,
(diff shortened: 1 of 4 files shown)


runspired:
@wojciechb why can't you use `notifyPropertyChange` anymore?

wojciechb:
> @wojciechb why can't you use `notifyPropertyChange` anymore?

Sorry, I wrote that in a little bit unclear way - I can use notifyPropertyChange, but I cannot set observed property directly without using Ember setters (and when using Ember setters notifyPropertyChange isn't needed). 
So it prevents me from integrating non-ember objects using notifyPropertyChange.

rwjblue (commented):
`MANDATORY_SETTER` is not intended to be optional in this way, I'm not certain we should move forward here. The only reason this worked previously was really just left over cruft from before we had modern tooling around DEBUG vs PROD builds.

If you feel strongly about your use case I think you would need to put together an RFC to propose this as a publicly available escape hatch...

rwjblue:
Closing for now. I think the right path forward here is via an RFC...

PR #17173, opened 2018-11-01
[docs] Adds documentation to CoreObject for native classes

pzuraq:
Adds documentation for native classes, prepping for making the officially public API.


packages/@ember/-internals/runtime/lib/system/core_object.js
```diff
@@ -135,6 +135,49 @@ function initialize(obj, properties) {
...
+  ```
+
+  Some notes about `class` usage:
+
+  * Instead of using `this._super()`, you should use standard `super` syntax in
+    native classes. See the [MDN docs on classes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes#Super_class_calls_with_super)
+    for more details.
+  * Avoid using the `constructor` in favor of `init`. Services are not available
+    during the `constructor` phase, and the nuances of the two can be confusing.
+  * Zebra-striping (using native classes, and switching back to the old Ember
```
rwjblue: While I know we have used this term a bit, I don't think its very widely understood outside of our immediate circles. I'd prefer to remove the "zebra striping" phrasing and just use the parenthetical directly...

Turbo87:
closing in favor of #17174

PR #17174, opened 2018-11-01
[docs] Adds documentation to CoreObject for native classes

pzuraq:
Adds documentation for native classes, prepping for making the officially public API.

### DIFF
--- a/packages/@ember/-internals/runtime/lib/system/core_object.js
+++ b/packages/@ember/-internals/runtime/lib/system/core_object.js
@@ -135,6 +135,63 @@ function initialize(obj, properties) {
 }
 
 /**
+  `CoreObject` is the base class for all Ember constructs. It establishes a
+  class system based on Ember's Mixin system, and provides the basis for the
+  Ember Object Model. `CoreObject` should generally not be used directly,
+  instead you should use `EmberObject`.
+
+  ## Usage
+
+  You can define a class by extending from `CoreObject` using the `extend`
+  method:
+
+  ```js
+  const Person = CoreObject.extend({
+    name: 'Tomster',
+  });
+  ```
+
+  For detailed usage, see the [Object Model](https://guides.emberjs.com/release/object-model/)
+  section of the guides.
+
+  ## Usage with Native Classes
+
+  Native JavaScript `class` syntax can be used to extend from any `CoreObject`
+  based class:
+
+  ```js
+  class Person extends CoreObject {
+    init() {
+      super.init(...arguments);
+      this.name = 'Tomster';
+    }
+  }
+  ```
+
+  Some notes about `class` usage:
+
+  * `new` syntax is not currently supported with classes that extend from
+    `EmberObject` or `CoreObject`. You must continue to use the `create` method
+    when making new instances of classes, even if they are defined using native
+    class syntax. If you want to use `new` syntax, consider creating classes
+    which do _not_ extend from `EmberObject` or `CoreObject`. Ember features,
+    such as computed properties and decorators, will still work with base-less
+    classes.
+  * Instead of using `this._super()`, you must use standard `super` syntax in
+    native classes. See the [MDN docs on classes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes#Super_class_calls_with_super)
+    for more details.
+  * Native classes support using [constructors](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes#Constructor)
+    to set up newly-created instances. Ember uses these to, among other things,
+    support features that need to retrieve other entities by name, like Service
+    injection and `getOwner`. To ensure your custom instance setup logic takes
+    place after this important work is done, avoid using the `constructor` in
+    favor of `init`.
+  * Properties passed to `create` will be available on the instance by the time
+    `init` runs, so any code that requires these values should work at that
+    time.
+  * Using native classes, and switching back to the old Ember Object model is
+    fully supported.
+
   @class CoreObject
   @public
 */


packages/@ember/-internals/runtime/lib/system/core_object.js
```diff
@@ -135,6 +135,49 @@ function initialize(obj, properties) {
...
+
+  ## Usage with Native Classes
+
+  Native Javascript `class` syntax can be used to extend from any `CoreObject`
+  based class:
+
+  ```js
+  class Person extends CoreObject {
+    init() {
+      super.init(...arguments);
```
mike-north: You may want to consider explicitly passing arguments to `super.init` in your example, so consumers know what `init()` receives.

packages/@ember/-internals/runtime/lib/system/core_object.js
```diff
@@ -135,6 +135,49 @@ function initialize(obj, properties) {
...
+      this.name = 'Tomster';
+    }
+  }
+  ```
+
+  Some notes about `class` usage:
+
+  * Instead of using `this._super()`, you should use standard `super` syntax in
```
mike-north:
```suggestion
  * Instead of using `this._super()`, you must use standard `super` syntax in
```
Is this advice universally-relevant enough that we can be more forceful?

packages/@ember/-internals/runtime/lib/system/core_object.js
```diff
@@ -135,6 +135,49 @@ function initialize(obj, properties) {
...
+    native classes. See the [MDN docs on classes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes#Super_class_calls_with_super)
+    for more details.
+  * Avoid using the `constructor` in favor of `init`. Services are not available
+    during the `constructor` phase, and the nuances of the two can be confusing.
```
mike-north: My guess is that `getOwner` is not available for use in the constructor body either. Maybe it would be better to say something more general like 

> Native classes support use constructors (MDN link) used to set up newly-created instances. Ember uses these to, among other things, support features that need to retrieve other entities by name, like Service injection and `getOwner`. To ensure your _custom_ instance setup logic takes place after this important work is done, avoid using the `constructor` in favor of `init`.

packages/@ember/-internals/runtime/lib/system/core_object.js
```diff
@@ -135,6 +135,49 @@ function initialize(obj, properties) {
...
+  * Using native classes, and switching back to the old Ember Object model is
+    fully supported.
```
mike-north: Is this supported/encouraged/discouraged?

```ts
export default class XFoo extends Ember.Component({

  // properties that I deliberately want to share between instances
  tagName: 'ul',

  fullName: computed('firstName', 'lastName', function() {
	return `${this.firstName} ${this.lastName}`;
  }),

  // immutable "defaults" for member data, deliberately shared across all instances
  firstName: '',  
  lastName: ''

}) {
 
  // standard ES6 class stuff

  logName() {
    console.log(this.fullName);
  }
}

```

packages/@ember/-internals/runtime/lib/system/core_object.js
pzuraq: yes, absolutely

packages/@ember/-internals/runtime/lib/system/core_object.js
pzuraq: Discouraged, since it will be very difficult to codemod half-way states like this. It's much easier for us to go from a clean-slate to full ES classes in the codemod.

packages/@ember/-internals/runtime/lib/system/core_object.js
pzuraq: I don't think we actually guarantee a shape to the arguments to `init`, which is why the recommended advice was to always do `this._super(...arguments);`

mike-north: Ok, what if I make an ember object subclass, and call `new Foo({ name: 'Mike' })`. Is this exactly the same as `Ember.Object.create`, in that "stuff is just there" by the time the init hook is fired?

pzuraq: `new Foo` is not supported, actually. We should probably call this out in the notes, but yes, if you do `Foo.create({ name: 'Mike' })` then the `name` property will be set by the time the `init` hook is fired.

rwjblue:
@pzuraq - I'd like to get this back on track, think you might have time to revive it soon?

rwjblue (approved):

packages/@ember/-internals/runtime/lib/system/core_object.js
pzuraq: I don't think this should be addressed in here, but in a linting rule. We shouldn't even suggest this is a possibility in our docs, ideally. We could cover it in guides though.

packages/@ember/-internals/runtime/lib/system/core_object.js
```diff
@@ -135,6 +135,62 @@ function initialize(obj, properties) {
...
+      super.init(...arguments);
+      this.name = 'Tomster';
+    }
+  }
+  ```
+
+  Some notes about `class` usage:
+
+  * `new` syntax is not currently supported with classes that extend from
+    `EmberObject`. You must continue to use the `create` method when making new
```
mike-north: The example above refers to `CoreObject`, and the above sentence would seem to suggest that I can use `new Foo()` as long as I extend from `CoreObject` instead of `EmberObject`. Is this accurate?

pzuraq: Good catch, that is not accurate. Will update to refer to `CoreObject` instead

Turbo87 (approved):

PR #17177, opened 2018-11-02
[RFC #338] Implement new template helpers

cibernox:
- [x] Implement `{{eq}}` helper
- [x] Implement `{{not-eq}}` helper
- [x] Implement `{{not}}` helper
- [x] Implement `{{and}}` helper
- [x] Implement `{{or}}` helper
- [x] Implement `{{gt}}` helper
- [x] Implement `{{gte}}` helper
- [x] Implement `{{lt}}` helper
- [x] Implement `{{lte}}` helper

**Open questions:**
- [ ] The `{{gt}}`, `{{gte}}`, `{{lt}}` and `{{lte}}` from `ember-truth-helpers` take a named argument `forceNumber=true` to transform input arguments to numbers. Do we want to replicate those too?


packages/@ember/-internals/glimmer/tests/integration/helpers/gt-test.js
```diff
@@ -0,0 +1,103 @@
...
+
+      this.assertText('yes');
+    }
+
+    ['@test of the first argument is null or undefined, it never pulls the second argument']() {
+      let didInvokeRight = false;
+      this.render(`{{gt left right}}`, {
+        left: undefined,
+        get right() {
+          debugger;
```
cibernox: @rwjblue Any guidance on how to define a getter or a CP to verify that its value is not eagerly pulled by the helper? Seems that right now the keys of this object are eagerly accessed by `Object.assign` by the setup of the test itself.

rwjblue: Instead of rendering gt directly render a component whose own template uses gt, then have that component have left and right CP’s that you can track when they are evaluated

cibernox:
@rwjblue Tests are failing because of missing documentation, however I've documented every helper, so I'm not sure what I'm missing.

packages/@ember/-internals/glimmer/lib/helpers/gt.ts
```diff
@@ -0,0 +1,38 @@
...
+  @since 2.7.0
+*/
+function gt({ positional: { references } }: CapturedArguments) {
+  let left = references[0].value();
+  if (left === undefined || left === null) {
+    return false;
+  }
+  let right = references[1].value();
+  if (right === undefined || right === null) {
+    return false;
```
makepanic: I think if right is `null`, gt can return `true` if one follows JS semantics:

packages/@ember/-internals/glimmer/lib/helpers/or.ts
```diff
@@ -0,0 +1,30 @@
...
+  {{! be validData either `isAdmin` or `isAuthor` are truthy}}
+  ```
+
+  @public
+  @method or
+  @for Ember.Templates.helpers
+  @since 2.7.0
+*/
+function or({ positional: { references } }: CapturedArguments) {
+  return references.reduce((acc: any, ref) => acc || ref.value(), false);
```
makepanic: This could return early too if references is an array.
Using `references.some` while checking for any reference value to be true would return early as or only cares for any true argument.

packages/@ember/-internals/glimmer/lib/helpers/and.ts
```diff
@@ -0,0 +1,30 @@
...
+  ```handlebars
+  {{and isAdmin isIdle validData}}
+
+  {{! be validData if `isAdmin` and `isIdle` are both truthy}}
+  ```
+
+  @public
+  @method and
+  @for Ember.Templates.helpers
+  @since 2.7.0
```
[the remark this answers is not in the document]
cibernox: Master is 2.7, am I wrong? 2.6 maybe?

packages/@ember/-internals/glimmer/lib/helpers/gt.ts
cibernox: TIL. Good catch.

packages/@ember/-internals/glimmer/lib/helpers/and.ts
```diff
@@ -0,0 +1,30 @@
...
+*/
+function and({ positional: { references } }: CapturedArguments) {
+  return references.reduce((acc: any, ref) => acc && ref.value(), true);
```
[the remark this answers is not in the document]
cibernox: I did some benchmark and a for loop was by far the fastest implementation.

packages/@ember/-internals/glimmer/lib/helpers/gt.ts
cibernox: Updated. `null` when used in numeric contexts is treated as 0.
If I just compare `left > right` when right can be null, typescript complains, so I explicitly handle that case. However if someone knows how to make typescript accept that, I think it would be dryer.
I added tests cases.

packages/@ember/-internals/glimmer/lib/helpers/or.ts
```diff
@@ -22,7 +22,14 @@ import { InternalHelperReference } from '../utils/references';
...
 function or({ positional: { references } }: CapturedArguments) {
-  return references.reduce((acc: any, ref) => acc || ref.value(), false);
+  let last: any = false;
+  for (let i = 0; i < references.length; i++) {
+    last = references[i].value();
+    if (last) {
+      return last;
+    }
+  }
+  return last;
```
makepanic: I don't know if V8 already does it, but one could try to remove the local variable to `return true` from within the loop or `return false` in the end

Turbo87:
```
for (let i = 0; i < references.length; i++) {
  if (references[i].value()) {
    return true;
  }
}
return false;
```

cibernox: I can't. And and or don't always return boolean, they return the last/first truthy value, whatever that is

Turbo87: I see, makes sense

rwjblue (changes requested):
Looking really good here, only major thing that I see is that it needs feature flagging...

cibernox:
I added a feature flag. It's enabled by default.

packages/@ember/-internals/glimmer/lib/helpers/and.ts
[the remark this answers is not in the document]
locks: do you mean 3.6?

packages/@ember/-internals/glimmer/lib/helpers/and.ts
```diff
@@ -0,0 +1,37 @@
...
+  ```handlebars
+  {{and isAdmin isIdle validData}}
+
+  {{! be validData if `isAdmin` and `isIdle` are both truthy}}
+  ```
+
+  @public
+  @method and
+  @for Ember.Templates.helpers
+  @since 2.7.0
```
locks:
```suggestion
  @since 3.7.0
```

packages/@ember/-internals/glimmer/lib/helpers/eq.ts
```diff
@@ -0,0 +1,30 @@
...
+  ```handlebars
+  {{eq type "button"}}
+
+  {{! be true if `type` is "button"}}
+  ```
+
+  @public
+  @method eq
+  @for Ember.Templates.helpers
+  @since 2.7.0
```
locks:
```suggestion
  @since 3.7.0
```

packages/@ember/-internals/glimmer/lib/helpers/gt.ts
```diff
@@ -0,0 +1,41 @@
...
+  ```handlebars
+  {{gt age 17}}
+
+  {{! be true if `age` is > 17}}
+  ```
+
+  @public
+  @method gt
+  @for Ember.Templates.helpers
+  @since 2.7.0
```
locks:
```suggestion
  @since 3.7.0
```

packages/@ember/-internals/glimmer/lib/helpers/gte.ts
```diff
@@ -0,0 +1,41 @@
...
+  ```handlebars
+  {{gt age 17}}
+
+  {{! be true if `age` is > 17}}
+  ```
+
+  @public
+  @method gte
+  @for Ember.Templates.helpers
+  @since 2.7.0
```
locks:
```suggestion
  @since 3.7.0
```

packages/@ember/-internals/glimmer/lib/helpers/not-eq.ts
```diff
@@ -0,0 +1,30 @@
...
+  ```handlebars
+  {{not-eq type "button"}}
+
+  {{! be true if `type` !== "button"}}
+  ```
+
+  @public
+  @method not-eq
+  @for Ember.Templates.helpers
+  @since 2.7.0
```
locks:
```suggestion
  @since 3.7.0
```

packages/@ember/-internals/glimmer/lib/helpers/not.ts
```diff
@@ -0,0 +1,30 @@
...
+  ```handlebars
+  {{not disabled}}
+
+  {{! be true if `disabled` is false}}
+  ```
+
+  @public
+  @method not
+  @for Ember.Templates.helpers
+  @since 2.7.0
```
locks:
```suggestion
  @since 3.7.0
```

packages/@ember/-internals/glimmer/lib/helpers/or.ts
```diff
@@ -0,0 +1,37 @@
...
+  ```handlebars
+  {{or isAdmin isAuthor}}
+
+  {{! be validData either `isAdmin` or `isAuthor` are truthy}}
+  ```
+
+  @public
+  @method or
+  @for Ember.Templates.helpers
+  @since 2.7.0
```
locks:
```suggestion
  @since 3.7.0
```

locks (changes requested):

packages/@ember/-internals/glimmer/lib/helpers/eq.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  Example:
+
+  ```handlebars
+  {{eq type "button"}}
+
+  {{! be true if `type` is "button"}}
+  ```
+
+  @public
```
rwjblue: The docs all need `@category ember-basic-template-helpers` (this _should_ allow the docs to avoid shipping to emberjs.com/api/ before we’ve enabled the feature flag)...

packages/@ember/canary-features/index.ts
```diff
@@ -18,6 +18,7 @@ export const DEFAULT_FEATURES = {
   EMBER_TEMPLATE_BLOCK_LET_HELPER: true,
   EMBER_METAL_TRACKED_PROPERTIES: null,
   EMBER_GLIMMER_ANGLE_BRACKET_INVOCATION: true,
+  EMBER_BASIC_TEMPLATE_HELPERS: true,
```
rwjblue: The initial value here should be `null`. This helps ensure the tests pass with and without the feature. After the RFC is merged we will begin reviewing to “enable by default” (checking api docs, guides, implementation, etc) and then will flip from null to true.

packages/@ember/-internals/glimmer/lib/helpers/and.ts
```diff
@@ -0,0 +1,37 @@
...
+*/
+function and({ positional: { references } }: CapturedArguments) {
+  let last: any = true;
+  for (let i = 0; i < references.length; i++) {
+    last = references[i].value();
+    if (!last) {
```
rwjblue: @chancancode - Should we consider `[]` as falsey here? In general, in template land I think we do right?

packages/@ember/-internals/glimmer/lib/helpers/gt.ts
```diff
@@ -0,0 +1,41 @@
...
+*/
+function gt({ positional: { references } }: CapturedArguments) {
+  let left = references[0].value();
+  if (left === undefined || left === null) {
+    return false;
+  }
+  let right = references[1].value();
+  if (right === undefined) {
```
rwjblue: Does this satisfy a specific edge case? I would have imagined that the following would have been enough:


```js
return left > right;
```

packages/@ember/-internals/glimmer/lib/helpers/gte.ts
```diff
@@ -0,0 +1,41 @@
...
+  let left = references[0].value();
+  if (left === undefined || left === null) {
+    return false;
+  }
+  let right = references[1].value();
+  if (right === undefined) {
+    return false;
+  }
+  if (right === null) {
+    return left >= 0;
```
rwjblue: This seems odd to me, why do we need to extra logic (vs just using the final `return left >= right`)?

packages/@ember/-internals/glimmer/lib/helpers/gt.ts
cibernox: I tried, but typescript doesn't like comparing things with null.

packages/@ember/canary-features/index.ts
locks:
```suggestion
  EMBER_BASIC_TEMPLATE_HELPERS: null,
```

locks (changes requested):

packages/@ember/-internals/glimmer/lib/helpers/gte.ts
cibernox: Same reason. Typescript doesn't ket me compare things with undefined or null. How can we silence that?

packages/@ember/-internals/glimmer/lib/helpers/and.ts
chancancode: yeah, I think we do. https://github.com/emberjs/ember.js/blob/1b7e613f4455c929010b59dd8d65c0b892ca3b08/packages/%40ember/-internals/glimmer/lib/utils/to-bool.ts#L4-L10

I bigger question is should we support proxies, I completely forgot about that. I think that needed to be discussed and clarified in the RFC 🤧

cibernox: That is surprising. I was expecting and/or helpers to behave just like && / II. Why this deviation from standard JS behaviour?

chancancode:
```hbs
{{#each @posts as |post|}}
  ...
{{else}}
  No posts!
{{/each}}
```

rwjblue: Well it really isn't about `each`, `{{#if someEmptyArray}}{{else}}{{/if}}` will render the falsey branch. I think it is very confusing to have `if` behave differently than `eq`/`not`.

chancancode: It works the same way for...

```hbs
{{#if @posts as |post|}}
  ...
{{else}}
  No posts!
{{/if}}
```

...and `with`, `let` etc. I think the bottom line is it would be pretty surprising if `{{else}}` doesn't work consistently in the control flow things in the "language".

So now what if you refactor from the first example into...

```hbs
{{#each (or @posts @reblogs) as |post|}}
  ...
{{else}}
  No posts!
{{/each}}
```

...what should this do?

chancancode: I think it is probably important for the "language" to have one boolean semantics, and for better or for worse empty arrays are falsey everywhere else.

cibernox:
Closing this as it would have to be reworked

locks:
For reference, the original RFC has been superseded. See https://github.com/emberjs/rfcs/pull/388#issuecomment-574307880 for full details.

PR #17179, opened 2018-11-05
Small typo in failure message

locks:


ef4 (approved):

rwjblue:
@ef4 - FWIW, I think the messages are backwards. The `no extraneous classitems` test says things about "accidental removal" and the `No missing classitems` one says stuff about adding new features.

PR #17180, opened 2018-11-05
[BUGFIX beta] Router Service State should be correct in events

chadhietala:


chadhietala:
Cherry picked onto the beta branch.

PR #17184, opened 2018-11-06
[BUGFIX] Ensures removeAllListeners doesn't break subsequent adds

### LINKED ISSUE #17183: Ember Inspector Promise test errors
We've recently been experiencing errors with ember beta and canary. They can bee seen here https://travis-ci.org/emberjs/ember-inspector/jobs/449687708

pzuraq:
This popped up as a bug in Ember Inspector because they use the REMOVE_ALL functionality during tests. REMOVE_ALL's are tricky, they don't neatly fall into the new model for listeners (which is why we are deprecating and removing them in 3.9) but for now, this should prevent that particular bug from occurring again.

Resolves #17183

### DIFF
--- a/packages/@ember/-internals/meta/lib/meta.ts
+++ b/packages/@ember/-internals/meta/lib/meta.ts
@@ -619,6 +619,10 @@ export class Meta {
       } else {
         // update own listener
         listener.kind = kind;
+
+        // TODO: Remove this when removing REMOVE_ALL, it won't be necessary
+        listener.target = target;
+        listener.method = method;
       }
     }
   }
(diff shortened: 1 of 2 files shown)


pzuraq:
cc @rwwagner90 @krisselden @rwjblue

RobbieTheWagner (approved):

rwjblue (commented):
Just confirming: this doesn't impact the performance characteristics?

pzuraq:
This should not, `REMOVE_ALL` is _very_ uncommon and when it's not a `REMOVE_ALL` those listeners will actually have the exact target and method, which is why the assignment won't be needed. The basic shape of the listeners will be the same in any case, though they may have slightly different types.

PR #17188, opened 2018-11-07
[BUGFIX beta][FEAT] Adds a second dist build which targets IE

pzuraq:
This PR adds a second dist to the build which targets IE. This allows
the primary bundles to drop ES5 compilation from babel, and allows us to
choose which bundles we want to use based on the consuming application's
targets.

In the long run, we want to actually enable consuming apps to build the
Ember source code themselves. This is a temporary in-between measure to
address a bug in native classes, in which users who do _not_ have IE as
a build target cannot `.extend` from native classes due to conflicts
in transpilation vs actual native class syntax.


ember-cli-build.js
```diff
@@ -110,14 +107,70 @@ module.exports = function() {
...
+    nodeTests(),
+
+    // test harness
+    testIndexHTML(),
+    jquery(),
+    qunit(),
+  ]);
+};
+
+function toTranspiledBundles(
```
pzuraq: There's a lot going into this function. Main thing I was trying to avoid is creating the exact same trees twice, given some of them can be fairly expensive (like rollup, I believe).

bin/run-tests.js
```diff
@@ -84,10 +84,15 @@ function generateEachPackageTests() {
     .forEach(generateTestsFor);
 }
 
-function generateBuiltTests() {
+function generateBuiltTests(ie) {
```
rwjblue: Instead of adding an arg, lets just push twice in the `testFunctions.push`'s below...

build-metadata.json
```diff
@@ -0,0 +1,6 @@
+{
```
rwjblue: this shouldn't be checked in (maybe we should add it to `.gitignore`?)

ember-cli-build.js
```diff
@@ -110,14 +107,70 @@ module.exports = function() {
...
+    files: ['testem.js'],
+    destDir: '',
+    annotation: 'tests/testem.js',
+  });
+
+  return new MergeTrees([
+    new Funnel(es, { destDir: 'es' }),
+    pkgAndTestESBundleDebug,
+    bundlesES6,
+    bundlesIE,
```
rwjblue: Should we maybe only do this for prod builds?

ember-cli-build.js
```diff
@@ -110,14 +107,70 @@ module.exports = function() {
     rollupPackage(packagesES, '@ember/-internals/container'),
   ]);
 
+  let bundlesES5 = toTranspiledBundles(
```
rwjblue: Thoughts on swapping things around a bit such that this is a straight pipeline (doing the es bundle then doing another babel pass on it to add the extra processing)?

rwjblue:
Oh, also can you update the commit message to start with `[BUGFIX beta]`?

pzuraq:
ALrighty, restructured to a better flow:

1. Gather all files
2. Transpile to ES5
3. Debug macros + strip and prep for production
4. Gather files into bundles
5. Transform AMD and concat files

This flow ended up making the most sense to me given that we _must_ do
the AMD transform last. It keeps most major operations at a per-file
level, and saves the file-divying and concatting for last.

rwjblue:
Hmm, I would not have expected the sizes to differ so much, would you mind attaching a .tar.gz of the `dist/` (before and after) so I can poke through it?

pzuraq:
Sure thing!

[dist-dual-build.tar.gz](https://github.com/emberjs/ember.js/files/2566388/dist-dual-build.tar.gz)
[dist-master.tar.gz](https://github.com/emberjs/ember.js/files/2566389/dist-master.tar.gz)

pzuraq:
I've been diffing the output bundles and while I haven't checked every single difference, a lot of them seem to come to a few categories:

1. Comments are not being removed from the final output, when they previously were. Comments are still removed from the final `min` build, so this isn't really a concern for that build and in fact they probably should not have been removed in the other builds.

2. Dead code elimination isn't quite working _in the `ember.prod.js` build_. Specifically, it appears to do a single pass and remove tautologies, but then does not remove functions which were only used in those tautological branches. This again does not affect the `min` build (verified by reading the actual function code).

3. Very small differences in the way transpilation/minification occurs. For instance, in one build an extra pair of enclosing parens may be emitted around an expression. In another, a variable which was inlined was no longer inlined. 

Surprisingly, removing the `minify-dead-code-elimination` plugin actually _reduces_ the final build size to below the original final build, so it seems like the plugin may have been interfering with the minification process more than helping.

```
# before
 - dist/ie/ember-template-compiler.js: 408.98 KB (81.99 KB gzipped)
 - dist/ie/ember-testing.js: 78.93 KB (17.95 KB gzipped)
 - dist/ie/ember-tests.js: 3.17 MB (420.78 KB gzipped)
 - dist/ie/ember-tests.prod.js: 1.07 MB (143.26 KB gzipped)
 - dist/ie/ember.debug.js: 1.91 MB (399.59 KB gzipped)
 - dist/ie/ember.min.js: 469.38 KB (120.36 KB gzipped)
 - dist/ie/ember.prod.js: 1.77 MB (371.62 KB gzipped)

# after
 - dist/ie/ember-template-compiler.js: 417.36 KB (83 KB gzipped)
 - dist/ie/ember-testing.js: 78.93 KB (17.95 KB gzipped)
 - dist/ie/ember-tests.js: 3.17 MB (420.78 KB gzipped)
 - dist/ie/ember-tests.prod.js: 1.11 MB (154.61 KB gzipped)
 - dist/ie/ember.debug.js: 1.91 MB (399.59 KB gzipped)
 - dist/ie/ember.min.js: 470.61 KB (120.07 KB gzipped)
 - dist/ie/ember.prod.js: 1.84 MB (386.31 KB gzipped)
```

PR #17189, opened 2018-11-08
remove redundant arg flipping in Evented#one

bekzod:
argument swapping is done in `addListener` downstream already

### DIFF
--- a/packages/@ember/-internals/runtime/lib/mixins/evented.js
+++ b/packages/@ember/-internals/runtime/lib/mixins/evented.js
@@ -94,11 +94,6 @@ export default Mixin.create({
     @public
   */
   one(name, target, method) {
-    if (!method) {
-      method = target;
-      target = null;
-    }
-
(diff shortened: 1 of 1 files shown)


packages/@ember/-internals/runtime/lib/mixins/evented.js
```diff
@@ -120,10 +115,12 @@ export default Mixin.create({
     @method trigger
     @param {String} name The name of the event
     @param {Object...} args Optional arguments to pass on
+    @return this
     @public
   */
   trigger(name, ...args) {
     sendEvent(this, name, args);
+    return this;
```
bekzod: also added `return this` since all other methods have it

rwjblue: Please revert this part, it’s unrelated to the main point of the PR and I don’t personally think we should encourage the chaining...

rwjblue (approved):

PR #17194, opened 2018-11-08
Refactor controller blueprint tests to use fixtures

simonihmig:
### What

Use fixtures for asserting the generated files' content for the `controller` and `controller-test` blueprints, instead of hard-coding it in the tests.

The `helper` tests already do this, and if this accepted I would like to follow up on this with all the other blueprint tests.

### Why?

I find it easier to read and maintain here. But the main motivation is updating and maintaining the [blueprints for ember-cli-typescript](https://github.com/typed-ember/ember-cli-typescript-blueprints). Those share the exact same logic (like support for pods, MU, different test frameworks etc.), only the generated files' content (and their extension) are different. By decoupling the blueprint logic from the blueprint's file contents in tests, this allows to easily update the blueprint's tests in `e-c-ts` in sync with the original ones, which currently is a *PITA*. For more context, see https://github.com/typed-ember/ember-cli-typescript-blueprints/issues/14

### DIFF
--- a/node-tests/blueprints/controller-test-test.js
+++ b/node-tests/blueprints/controller-test-test.js
@@ -21,19 +21,25 @@ describe('Blueprint: controller-test', function() {
       return emberNew();
     });
 
-    it('controller-test foo', function() {
-      return emberGenerateDestroy(['controller-test', 'foo'], _file => {
-        expect(_file('tests/unit/controllers/foo-test.js')).to.equal(
-          fixture('controller-test/default.js')
-        );
+    describe('with ember-cli-qunit@4.1.0', function() {
+      beforeEach(function() {
+        generateFakePackageManifest('ember-cli-qunit', '4.1.0');
       });
-    });
 
-    it('controller-test foo/bar', function() {
-      return emberGenerateDestroy(['controller-test', 'foo/bar'], _file => {
-        expect(_file('tests/unit/controllers/foo/bar-test.js')).to.equal(
-          fixture('controller-test/default-nested.js')
-        );
+      it('controller-test foo', function() {
+        return emberGenerateDestroy(['controller-test', 'foo'], _file => {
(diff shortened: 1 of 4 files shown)


node-tests/blueprints/controller-test-test.js
```diff
@@ -21,19 +21,25 @@ describe('Blueprint: controller-test', function() {
...
     });
 
-    it('controller-test foo', function() {
-      return emberGenerateDestroy(['controller-test', 'foo'], _file => {
-        expect(_file('tests/unit/controllers/foo-test.js')).to.equal(
-          fixture('controller-test/default.js')
-        );
+    describe('with ember-cli-qunit@4.1.0', function() {
+      beforeEach(function() {
+        generateFakePackageManifest('ember-cli-qunit', '4.1.0');
```
simonihmig: Note: this is not really needed here. But AFAICT having `ember-cli-qunit` with a higher version in an addons `dependencies` would cause `ember-cli-blueprint-test-helpers` to run the blueprint in tests with that version, causing a different output than the tests expect. So we make the expected `ember-cli-qunit` explicit here, so the installed version does not matter, and we can use the same test as-is for the TS blueprints.

rwjblue:
@simonihmig this is very helpful! FWIW, I _think_ updating all of the blueprints to use fixtures will help unblock https://github.com/emberjs/ember.js/pull/17126.

node-tests/blueprints/controller-test-test.js
rwjblue: This is exactly the scenario that I hit in https://github.com/emberjs/ember.js/pull/17126.

rwjblue (approved):

simonihmig:
Renamed the fixture files, and fixed some prettier issues. Build is green now...

rwjblue:
Thank you!

PR #17198, opened 2018-11-10
Refactor component blueprint tests to use fixtures

simonihmig:
This was a bit trickier than the previous refactorings, as with nested components and different paths the combinations exploded a bit, and having a separate fixture file for each of them would have been a bit too much. So...
* I renamed a few components / paths in tests, so there are not so many different combinations of `foo`'s, `x-foo`'s, `bar`'s and whatnot. But the semantics of each test hasn't been changed.
* I extended the `fixture` helper a bit to optionally support some string replacement (templating) functionality, to not have a dozen slightly different test fixtures (given different combinations of component names and paths)


ppcano:
#17212 may supersede this PR

PR #17207, opened 2018-11-13
Refactor util blueprint tests to use fixtures

simonihmig:


ppcano:
#17212 may supersede this PR

PR #17213, opened 2018-11-18
[BUGFIX beta] RouteInfoWithAttributes on transition

chadhietala:
It was pointed out that `transition.from` should always be a `RouteInfoWithAttributes` the active route path has been resolved. Also within the `routeDidChange` event `transition.to` has the same guarantees about being fully resolved so it to should be a `RouteInfoWithAttributes`.


rwjblue (approved):

chadhietala:
Cherry picked onto the beta branch

PR #17214, opened 2018-11-18
Needed for 3.6 - Clarify that currentRoute's value is private API

jenweber:
Following on discussion for the upcoming blog post, it seems that this documentation needs a little more detail. My understanding is that the property is public API, but the value is not.

Is this correct?

[Link to discussion and 3.6 post](https://github.com/emberjs/website/pull/3673#discussion_r234420576)


rwjblue:
Added @chadhietala 	to review as he would know for sure...

rwjblue:
Thanks for chiming in @chadhietala! 

@jenweber - OK to close this (unmerged), then?

jenweber:
Closed, thanks for confirming!

PR #17220, opened 2018-11-22
[BUGFIX beta] Fix cycle detection in Ember.copy

hjdivad:
Previously, key iteration in objects would recurse prior to updating `seen` and therefore miss the cycle check in the recursive call, leading to stack overflows.


hjdivad:
@rwjblue i presume `[BUGFIX beta]` is appropriate here?

rwjblue (approved):
Thank you!!

PR #17226, opened 2018-11-23
WIP: Decorator Implementation

NullVoxPopuli:
Not only is this a WIP, but the RFC still needs to be merged / accepted. 

Soon! :tm:

Hopefully this can be merged first: https://github.com/emberjs/ember.js/pull/17404
:)

Outstanding work:
 - mixin and subclassing (`_super`) support
 - more tests for the decorators
 - tests for the the other macros as well
 - cleanup
 - make the code simpler (rather than a sorta straight copy from ember-decorators)
 - adjust for latest decorator spec changes
 - benchmarking

### DIFF
--- a/broccoli/packages.js
+++ b/broccoli/packages.js
@@ -16,6 +16,7 @@ const WriteFile = require('broccoli-file-creator');
 const StringReplace = require('broccoli-string-replace');
 const GlimmerTemplatePrecompiler = require('./glimmer-template-compiler');
 const VERSION_PLACEHOLDER = /VERSION_STRING_PLACEHOLDER/g;
+const transfromBabelPlugins = require('./transforms/transform-babel-plugins');
 
 const debugTree = BroccoliDebug.buildDebugCallback('ember-source');
 
@@ -83,6 +84,13 @@ module.exports.getPackagesES = function getPackagesES() {
     exclude: ['**/*.ts'],
   });
 
+  // tsc / typescript handles decorators and class properties on its own
+  // so for non ts, transpile the proposal features (decorators, etc)
+  let transpiledProposals = debugTree(
+    transfromBabelPlugins(debugTree(nonTypeScriptContents, `get-packages-es:babel-plugins:input`)),
+    `get-packages-es:babel-plugins:output`
+  );
+
   let typescriptContents = new Funnel(debuggedCompiledTemplatesAndTypeScript, {
     include: ['**/*.ts'],
   });
@@ -95,7 +103,7 @@ module.exports.getPackagesES = function getPackagesES() {
 
   let debuggedCompiledTypescript = debugTree(typescriptCompiled, `get-packages-es:ts:output`);
 
-  let mergedFinalOutput = new MergeTrees([nonTypeScriptContents, debuggedCompiledTypescript], {
+  let mergedFinalOutput = new MergeTrees([transpiledProposals, debuggedCompiledTypescript], {
     overwrite: true,
   });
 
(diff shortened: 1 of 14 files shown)


package.json
```diff
@@ -77,6 +77,21 @@
     "resolve": "^1.6.0"
   },
   "devDependencies": {
+    "@babel/helper-module-imports": "^7.0.0",
```
NullVoxPopuli: yarn did this on its own.

Turbo87: yeah, I forgot to reorder the packages... 😞

broccoli/packages.js
```diff
@@ -83,6 +84,12 @@ module.exports.getPackagesES = function getPackagesES() {
     exclude: ['**/*.ts'],
   });
 
+  // tsc / typescript handles decorators and class properties on its own
+  // so for non ts, transpile the proposal features (decorators, etc)
+  let transpiledProposals = transfromBabelPlugins(
+    debugTree(nonTypeScriptContents, `get-packages-es:babel-plugins`)
```
Turbo87: I would suggest to put a `debugTree` inside **and** outside of it. `get-packages-es:babel:input` inside, and `get-packages-es:babel:output` outside

broccoli/transforms/transform-babel-plugins.js
```diff
@@ -0,0 +1,16 @@
...
+module.exports = function(tree) {
+  let options = {
+    sourceMaps: true,
+    plugins: [
+      ['@babel/plugin-proposal-decorators', { decoratorsBeforeExport: true, legacy: false }],
+      ['@babel/plugin-proposal-class-properties'],
+    ],
+    parserOpts: {
+      plugins: [['classProperties', {}], ['decorators', { decoratorsBeforeExport: true }]],
+    },
```
Turbo87: can you check if this works fine without the `parserOpts`? I vaguely remember that this was only needed for those babel steps that did not use the corresponding plugins

NullVoxPopuli: it does! most excellent!

broccoli/packages.js
NullVoxPopuli: done :)

jacobq:
Just curious, how does this related to [`ember-decorators`](https://github.com/ember-decorators/ember-decorators)? And which RFC is it related to? ([240 / ES Classes](https://github.com/emberjs/rfcs/pull/240)?)

packages/@ember/-internals/metal/lib/computed.ts
```diff
@@ -620,7 +626,117 @@ export default function computed(...args: (string | ComputedPropertyConfig)[]):
...
+      return target;
+    };
+  };
+}
+
+// computed(function() { ... })
+// computed('dependentKey', { get: function() { ... }})
+// computed('dependentKey', {})
+// computed('dependentKey', function() { ...})
+function isLegacyComputed(...args) {
```
pzuraq: I don't believe this implementation will work fundamentally. Consider the following test case:

```js
let firstNameAlias = computed('firstName', function() {
  return this.firstName;
});

let Class1 = EmberObject.extend({
  otherFirstName: firstNameAlias
});

class Class2 {
  @firstNameAlias otherFirstName;
}
```

As specified in the RFC, this must work (and it must work for computed macros in general to work, since this is fundamental to them).

NullVoxPopuli: I'll work on this case next, then :)

pzuraq:
Tests are invaluable @NullVoxPopuli, keep up the good work!

mike-north:
Does #17548 supersede this PR?

pzuraq:
Yes it does, closing since we merged the feature in a different PR

PR #17227, opened 2018-11-24
[BUGFIX beta] fix mouseEnter/Leave event delegation w/o jQuery for SVG & IE11

### LINKED ISSUE #17225: mouseenter: Object doesn't support property or method 'contains'
With `"jquery-integration": false` I'm getting `Object doesn't support property or method 'contains'` in IE11. It may have something to do with SVG.

screenshot 2018-11-22 at 12 22 49

https://github.com/emberjs/ember.js/blob/62643188404d1b8398cb133ac30cc0e866318a8d/packages/%40ember/-internals/views/lib/system/event_dispatcher.js#L338

<br>
Having to delay our jQuery-removal celebrations :)

simonihmig:
IE11 does not support `Node#contains()` on SVG elements, so in that case a custom fallback implementation is used.

Fixes #17225

### DIFF
--- a/packages/@ember/-internals/views/lib/system/event_dispatcher.js
+++ b/packages/@ember/-internals/views/lib/system/event_dispatcher.js
@@ -7,6 +7,7 @@ import jQuery, { jQueryDisabled } from './jquery';
 import ActionManager from './action_manager';
 import fallbackViewRegistry from '../compat/fallback-view-registry';
 import addJQueryEventDeprecation from './jquery_event_deprecation';
+import { contains } from './utils';
 
 /**
 @module ember
@@ -335,7 +336,7 @@ export default EmberObject.extend({
           while (
             target &&
             target.nodeType === 1 &&
-            (!related || (related !== target && !target.contains(related)))
+            (related === null || (related !== target && !contains(target, related)))
           ) {
             // mouseEnter/Leave don't bubble, so there is no logic to prevent it as with other events
             if (viewRegistry[target.id]) {
--- a/packages/@ember/-internals/glimmer/tests/integration/event-dispatcher-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/event-dispatcher-test.js
@@ -202,6 +202,62 @@ moduleFor(
       assert.strictEqual(receivedLeaveEvents[0].target, outer);
     }
 
+    ['@test delegated event listeners work for mouseEnter on SVG elements'](assert) {
+      let receivedEnterEvents = [];
+      let receivedLeaveEvents = [];
+
+      this.registerComponent('x-foo', {
+        ComponentClass: Component.extend({
+          tagName: 'svg',
+          mouseEnter(event) {
+            receivedEnterEvents.push(event);
+          },
+          mouseLeave(event) {
+            receivedLeaveEvents.push(event);
+          },
+        }),
+        template: `<g id="inner"></g>`,
+      });
+
+      this.render(`{{x-foo id="outer"}}`);
+
+      let parent = this.element;
+      let outer = this.$('#outer')[0];
+      let inner = this.$('#inner')[0];
+
+      // mouse moves over #outer
+      this.runTask(() => {
+        this.$(outer).trigger('mouseenter', { canBubble: false, relatedTarget: parent });
+        this.$(outer).trigger('mouseover', { relatedTarget: parent });
+        this.$(parent).trigger('mouseout', { relatedTarget: outer });
+      });
+      assert.equal(receivedEnterEvents.length, 1, 'mouseenter event was triggered');
+      assert.strictEqual(receivedEnterEvents[0].target, outer);
+
+      // mouse moves over #inner
+      this.runTask(() => {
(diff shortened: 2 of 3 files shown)


rwjblue:
Had one failure under the `/tests/?jquery=1.8.3` test run...

```
Test failed: EventDispatcher:  delegated event listeners work for mouseEnter on SVG elements", source:  (49)
[1124/013147.014267:INFO:CONSOLE(52)] "    Failed assertion: mouseenter event was triggered, expected: 1, but was: 0
    at Object.@test delegated event listeners work for mouseEnter on SVG elements (http://localhost:13141/ember-tests.js:19171:14)
    at Object.<anonymous> (http://localhost:13141/ember-tests.js:76563:37)
    at runTest (http://localhost:13141/qunit/qunit.js:1478:30)
    at Test.run (http://localhost:13141/qunit/qunit.js:1464:6)
    at http://localhost:13141/qunit/qunit.js:1670:12
    at advance (http://localhost:13141/qunit/qunit.js:1116:26)", source:  (52)
[1124/013147.014915:INFO:CONSOLE(52)] "    Failed assertion: Died on test #2     at generateTest (http://localhost:13141/ember-tests.js:76562:15)
    at Set.forEach (<anonymous>)
    at moduleFor (http://localhost:13141/ember-tests.js:76546:16)
    at http://localhost:13141/ember-tests.js:18961:27
    at internalRequire (http://localhost:13141/ember.debug.js:62:14)
    at Object.requireModule [as require] (http://localhost:13141/ember.debug.js:99:14)
    at http://localhost:13141/tests/?jquery=1.8.3&nolint=true:242:61: Cannot read property 'target' of undefined
TypeError: Cannot read property 'target' of undefined
    at Object.@test delegated event listeners work for mouseEnter on SVG elements (http://localhost:13141/ember-tests.js:19172:49)
    at Object.<anonymous> (http://localhost:13141/ember-tests.js:76563:37)
    at runTest (http://localhost:13141/qunit/qunit.js:1478:30)
    at Test.run (http://localhost:13141/qunit/qunit.js:1464:6)
    at http://localhost:13141/qunit/qunit.js:1670:12
    at advance (http://localhost:13141/qunit/qunit.js:1116:26)", source:  (52)
```

simonihmig:
Hm, ok, I'll check this. Although I did not touch the jQuery-based implementation, so the added test seems to have uncovered this!?

packages/@ember/-internals/views/lib/system/event_dispatcher.js
```diff
@@ -335,7 +336,9 @@ export default EmberObject.extend({
           while (
             target &&
             target.nodeType === 1 &&
-            (!related || (related !== target && !target.contains(related)))
+            (!related ||
+              (related !== target &&
+                !(target.contains ? target.contains(related) : contains(target, related))))
```
rwjblue: I'd _vaguely_ prefer to always use `contains(target, related)` and tweak the contains util to deal with the ternary, what do you think?

packages/@ember/-internals/views/lib/system/event_dispatcher.js
```diff
@@ -335,7 +336,9 @@ export default EmberObject.extend({
           while (
             target &&
             target.nodeType === 1 &&
-            (!related || (related !== target && !target.contains(related)))
+            (!related ||
```
rwjblue: I know this isn't new, but I'm curious why are we checking `!related` here? Can we tweak this to avoid boolean coercion? What do we expect `target.relatedTarget` to be? I was thinking it could only be either `undefined` or `null`, if thats correct we could (possibly in a different PR?) tweak to `related === undefined && related === null`...

packages/@ember/-internals/views/lib/system/event_dispatcher.js
simonihmig: Makes sense!

simonihmig: Done!

packages/@ember/-internals/views/lib/system/event_dispatcher.js
simonihmig: According to the [specs](https://www.w3.org/TR/uievents/#dom-mouseevent-relatedtarget) the  uninitialized value must be `null`, so I changes this to a strict null check.

simonihmig:
> Had one failure under the /tests/?jquery=1.8.3 test run...

Not sure why this is happening, but the root cause seems to be with jQuery. With `/tests/?jquery=1.10.0` the test is passing again!

How should we deal with that? Skip the test when using this older version of jquery? Or can we update the test suite's old jQuery version, like using the latest 1.x?

rwjblue:
> How should we deal with that?

Lets change our oldest tested jQuery to 1.10.0. Mind updating for that?

rwjblue:
Perfect, #17268 is landed. Ready for rebase / linting fix now.

simonihmig:
@rwjblue done, green! :)

rwjblue:
Thank you very much @simonihmig!

PR #17230, opened 2018-11-26
fixed rollupPackage for Windows

MelSumner:
Added path.isAbsolute to the rollupPackage function so files will be discovered properly no matter what OS (but in this case, especially Windows).


rwjblue:
Awesome, thanks for working through that!

PR #17233, opened 2018-11-28
[BUGFIX beta] Reverts EmberError to be a standard function

pzuraq:
This reverts a change from some years ago that made EmberError a native
class. The recent change to build non-transpiled classes is causing
failures in libraries that subclass EmberError (specifically Ember Data)
and with the messiness around how EmberError actually does the
extending, using native classes is not ideal.

Eventually, EmberError should be replaced entirely as it is not a good
idea to subclass Error at all. This will have to happen in the future,
and will likely require an RFC.


rwjblue (approved):

rwjblue:
@pzuraq can you mark this as BUGFIX beta?

PR #17238, opened 2018-11-29
[BUGFIX beta] Ensures Ember.A is a constructor

pzuraq:
The recent change to transpilation has made `new A()` break, because
A is an arrow function and arrow functions cannot be constructors. This
is a pretty common pattern in Ember apps and likely constitutes a
breaking change, so this PR fixes it by changing the arrow function to
a normal function.

### DIFF
--- a/packages/@ember/-internals/runtime/lib/mixins/array.js
+++ b/packages/@ember/-internals/runtime/lib/mixins/array.js
@@ -1641,12 +1641,36 @@ let A;
 
 if (ENV.EXTEND_PROTOTYPES.Array) {
   NativeArray.apply(Array.prototype);
-  A = arr => arr || [];
+
+  A = function(arr) {
+    deprecate(
+      '`new A()` has been deprecated, please update to calling A as a function: `A()`',
+      !(this instanceof A),
+      {
+        id: 'array.new-array-wrapper',
+        until: '3.9.0',
+        url: 'https://emberjs.com/deprecations/v3.x#toc_array-new-array-wrapper',
+      }
+    );
+
+    return arr || [];
+  };
 } else {
-  A = arr => {
+  A = function(arr) {
+    deprecate(
+      '`new A()` has been deprecated, please update to calling A as a function: `A()`',
+      !(this instanceof A),
+      {
+        id: 'array.new-array-wrapper',
+        until: '3.9.0',
+        url: 'https://emberjs.com/deprecations/v3.x#toc_array-new-array-wrapper',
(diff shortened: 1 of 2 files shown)


pzuraq:
cc @rwjblue

rwjblue (commented):
I don’t think we intended `new Ember.A()` (I’ve generally always used it without `new`), but it seems fine to fix if recent changes broke it.

Can you also add a test?

rwjblue:
Lets add a private API deprecation (`until: ‘3.9.0'`) for the new case actually, along with the test....

rwjblue:
FWIW, there is _a bit_ of usage of `new A(` / `new Ember.A` out in the wild, definitely justifies at least a private API deprecation like this:

https://emberobserver.com/code-search?codeQuery=new%20A(
https://emberobserver.com/code-search?codeQuery=new%20Ember.A(

packages/@ember/-internals/runtime/lib/mixins/array.js
```diff
@@ -1641,12 +1641,36 @@ let A;
...
-  A = arr => arr || [];
+
+  A = function(arr) {
+    deprecate(
+      '`new A()` has been deprecated, please update to calling A as a function: `A()`',
+      !(this instanceof A),
+      {
+        id: 'array.new-array-wrapper',
+        until: '3.9.0',
+        url: 'https://emberjs.com/deprecations/v3.x#toc_new-array-wrapper',
```
rwjblue: Seems to be wrong (the url doesn't link directly), when I find the right link it looks like:

https://deprecations-app-prod.herokuapp.com/deprecations/v3.x/#toc_array-new-array-wrapper

Can you confirm which is correct?

pzuraq: you are correct! Updated

packages/@ember/-internals/runtime/tests/system/native_array/a_test.js
```diff
@@ -12,5 +12,15 @@ moduleFor(
       assert.ok(EmberArray.detect(A()), 'returned an ember array');
       assert.ok(EmberArray.detect(A([1, 2])), 'returned an ember array');
     }
+
+    ['@test new Ember.A'](assert) {
+      expectDeprecation(() => {
+        assert.deepEqual(new A([1, 2]), [1, 2], 'array values were not be modified');
```
stefanpenner: ya, woah this was never an intended feature...

PR #17245, opened 2018-12-04
Cleanup CI configuration

Turbo87:
see commit messages, best reviewed commit-by-commit

/cc @rwjblue

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,10 +13,8 @@ before_install:
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
 
-  - sudo apt-key adv --fetch-keys http://dl.yarnpkg.com/debian/pubkey.gpg
-  - echo "deb http://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list
-  - sudo apt-get update -qq
-  - sudo apt-get install -y -qq yarn
+  - curl -o- -L https://yarnpkg.com/install.sh | bash
+  - export PATH=$HOME/.yarn/bin:$PATH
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -63,11 +34,34 @@ env:
...
 
+stages:
+  - basic test
+  - additional tests
+  - name: deploy
+    if: type IN (push)
+
+before_install:
+  - export DISPLAY=:99.0
+  - sh -e /etc/init.d/xvfb start
```
stefanpenner: Are these two lines still needed?

```js
  - export DISPLAY=:99.0
  - sh -e /etc/init.d/xvfb start
```

I don't believe we still require anything running in a virtual frame buffer. I suspect old browser testing needed this.

rwjblue: I suspect they are not needed any more

Turbo87: I was assuming that maybe Firefox still needed that, so I left it in there for now.

rwjblue (approved):

.travis.yml
Turbo87: looks like this is indeed still needed (see https://travis-ci.org/emberjs/ember.js/jobs/463512118)

I've dropped the commit that removes it again. I suggest merging this and then we can iterate forward.

stefanpenner: Oh, I didn't realize we had a smoke test for FF usage using xvfb. Future work could likely just use https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Headless_mode and avoid that. 

But that shouldn't hold up this PR, thanks for investigating.

stefanpenner (approved):

PR #17246, opened 2018-12-04
[BUGFIX beta] Adds early Android versions to legacy target list

pzuraq:
Currently, we only check for IE compatibility, but legacy Android browsers can also cause an issue. In general we need to solve this by respecting the browser targets of the client, but that will require more thorough changes to our build to actually use modules and drop bundling altogether.


lib/index.js
```diff
@@ -72,8 +72,11 @@ module.exports = {
     if (this.project.targets) {
       let browsers = browserslist(this.project.targets.browsers);
 
-      if (browsers.find(browser => browser.startsWith('ie'))) {
-        emberSourceDistPath = path.join(emberSourceDistPath, 'ie');
+      let hasIE = browsers.find(browser => browser.startsWith('ie'));
+      let hasLegacyAndroid = browsers.find(browser => browser.match(/android.*[1-4]\.[1-4]\.[1-4]/))
```
Turbo87: what about e.g. `android 4.0.1`?

pzuraq: yeah, discussed this with @rwjblue, he has a much better solution I'm going to incorporate.

rwjblue (approved):
Thank you, this seems much less brittle!

package.json
```diff
@@ -63,8 +63,8 @@
   "dependencies": {
     "broccoli-funnel": "^2.0.1",
     "broccoli-merge-trees": "^2.0.0",
-    "browserslist": "^4.3.4",
     "chalk": "^2.3.0",
+    "ember-cli-babel": "^7.1.3",
```
Turbo87: are we sure that this does not cause any unintended side-effects? 1. because `ember-cli-babel` might do something at app-buildtime and 2. because it's Babel 7, but we're still using Babel 6 to build the Ember.js assets

pzuraq: My understanding is that `ember-cli-babel` will only apply to addon source code, so `ember-source`s addon code, which doesn't exist. There's definitely the discrepancy with Babel 6, but since that is a _pre_publish step and fundamentally disconnected from `ember-cli-babel`, I figured it would make more sense to use the latest Babel instead. We can change this though if it seems problematic.

Turbo87: yeah, I just looked at https://unpkg.com/ember-source@3.5.1/lib/index.js, and this does indeed seem safe to do.

Turbo87 (approved):

rwjblue:
The browserstack failures seem related ([this CI job](
https://travis-ci.org/emberjs/ember.js/jobs/464120841)):

```
Running: ./node_modules/.bin/testem ci -f testem.dist.js --host 127.0.0.1 --port 7774
not ok 1 IE 11.0 - [undefined ms] - Global error: Expected ':' at http://127.0.0.1:7774/288551829212/dist/ember.debug.js, line 182
    ---
        Log: |
            { type: 'error',
              testContext: {},
              text:
               'Expected \':\' at http://127.0.0.1:7774/288551829212/dist/ember.debug.js, line 182\n' }
    ...
```

rwjblue (approved):

PR #17247, opened 2018-12-04
Simplify linting setup

Turbo87:
Instead of using a custom test runner for the lint checks we can use dedicated npm scripts and call them from `yarn lint` directly. Since there is no combined test case of tests+lint (anymore?) the linting integration in the test runner is not (or no longer) needed.

/cc @rwjblue


rwjblue (approved):

rwjblue:
👏

PR #17255, opened 2018-12-05
Simplify `node` tests execution

Turbo87:
see commit messages

/cc @rwjblue

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,10 +78,11 @@ jobs:
       - TEST_SUITE=built-tests
       - EMBER_ENV=production
     - env: TEST_SUITE=old-jquery-and-extend-prototypes
-    - env:
-      - TEST_SUITE=node
-      - EMBER_ENV=production
+    - name: Node.js Tests
       node_js: "6"
+      script:
+        - EMBER_ENV=production yarn ember build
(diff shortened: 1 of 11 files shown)


.travis.yml
```diff
@@ -78,10 +78,11 @@ jobs:
       - TEST_SUITE=built-tests
       - EMBER_ENV=production
     - env: TEST_SUITE=old-jquery-and-extend-prototypes
-    - env:
-      - TEST_SUITE=node
-      - EMBER_ENV=production
+    - name: Node.js Tests
       node_js: "6"
+      script:
+        - EMBER_ENV=production yarn ember build
```
rwjblue: I can’t recall specifically, but don’t all of the jobs already do a build?

Turbo87: yeah, but only when run through `yarn test` as the `pretest` script starts the build. IMHO being explicit here is easier to understand though compared to using the custom test runner.

Turbo87: we don't call `ember build` in `before_install` or `install` and `script` only does the prod build since we don't use `yarn test` there (which would cause the other `ember build`)

rwjblue (approved):
Awesome, thank you!

PR #17257, opened 2018-12-05
Update `qunit` to v2.8.0

Turbo87:
Extracted from #17255

... to figure out if that is what's causing CI to fail


Turbo87:
for future reference: https://github.com/qunitjs/qunit/pull/1252 is the PR that implemented it for QUnit and it was released in v2.5.1

rwjblue:
Whoops, sorry about that

rwjblue (approved):

PR #17259, opened 2018-12-05
Use ES6 code for custom testem reporters

Turbo87:
... and convert `testem.json` to JS

### DIFF
--- a/testem.travis-browsers.js
+++ b/testem.travis-browsers.js
@@ -1,4 +1,4 @@
-var FailureOnlyReporter = require('./lib/failure-only-testem-reporter');
+const FailureOnlyReporter = require('./lib/failure-only-testem-reporter');
 
 module.exports = {
   framework: 'qunit',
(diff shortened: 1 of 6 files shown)


testem.travis-browsers.js
```diff
@@ -1,4 +1,4 @@
-var FailureOnlyReporter = require('./lib/failure-only-testem-reporter');
+const FailureOnlyReporter = require('./lib/failure-only-testem-reporter');
```
rwjblue: The only difference in this file and the other one seems to be the list of browsers, can we instead use a single `testem.js` along with some environment variable inspection to determine the browsers?

rwjblue (approved):
Overall looks good to me, but had one question...

testem.travis-browsers.js
Turbo87: yeah, that might be a good idea for the future

PR #17262, opened 2018-12-06
[V8 Team Recommendation]  Fix JS Object as dictionary reuse

stefanpenner:
In recent discussion with the v8 team, they pointed out that our internal loader misuses `{}` as a dictionary, and recommended we use `Object.create(null)` instead.

The original loader already fixed this: https://github.com/ember-cli/loader.js/blob/b89c6932eed14abe9738993bfc21230ad1971e4f/lib/loader/loader.js#L31

We simply did not update this side of things.

### DIFF
--- a/packages/loader/lib/index.js
+++ b/packages/loader/lib/index.js
@@ -67,8 +67,8 @@ mainContext = this; // eslint-disable-line no-undef
   }
 
   if (typeof Ember.__loader === 'undefined') {
-    var registry = {};
-    var seen = {};
+    var registry = Object.create(null);
+    var seen = Object.create(null);
 
     enifed = function(name, deps, callback) {
       var value = {};


rwjblue (approved):
Seems good to me!

packages/loader/lib/index.js
```diff
@@ -67,8 +67,8 @@ mainContext = this; // eslint-disable-line no-undef
   }
 
   if (typeof Ember.__loader === 'undefined') {
-    var registry = {};
-    var seen = {};
+    var registry = Object.create(null);
+    var seen = Object.create(null);;
```
krisselden: has 2 semicolons

krisselden (dismissed):

rwjblue:
FYI - @krisselden cherry-picked this into the beta branch (since it wasn't labeled as such).

PR #17263, opened 2018-12-06
Simplify Browserstack test runner code

Turbo87:
/cc @kategengler @rwjblue 

bildschirmfoto 2018-12-06 um 01 26 48

### DIFF
--- a/bin/run-browserstack-tests.js
+++ b/bin/run-browserstack-tests.js
@@ -1,63 +1,34 @@
 /* eslint-disable no-console, node/no-unsupported-features */
 
-var RSVP = require('rsvp');
-var spawn = require('child_process').spawn;
+var execa = require('execa');
 
-function run(command, _args) {
-  var args = _args || [];
+function run(command, args = []) {
+  console.log('Running: ' + command + ' ' + args.join(' '));
 
-  return new RSVP.Promise(function(resolve, reject) {
-    console.log('Running: ' + command + ' ' + args.join(' '));
+  let p = execa(command, args);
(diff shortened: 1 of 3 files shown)


bin/run-browserstack-tests.js
```diff
@@ -1,71 +1,39 @@
-/* eslint-disable no-console */
+/* eslint-disable no-console, node/no-unsupported-features */
 
-var RSVP = require('rsvp');
-var spawn = require('child_process').spawn;
+var execa = require('execa');
```
rwjblue: Seems _generally_ odd to me to use `async` / `await` **and** `var` in the same file... 😹

bin/run-browserstack-tests.js
```diff
@@ -1,71 +1,39 @@
...
-  .finally(function() {
-    var promise = RSVP.resolve();
-    if (process.env.TRAVIS_JOB_NUMBER) {
-      promise = run('./node_modules/.bin/ember', ['browserstack:results']);
+(async function() {
+  await run('ember', ['browserstack:connect']);
+
+  try {
+    try {
+      // Calling testem directly here instead of `ember test` so that
```
rwjblue: Hmm, we could avoid the double build **and** still use `ember test` by way of `ember test --path=dist`, any objection?

Turbo87: I'm not sure what the advantage would be over just calling `testem` directly 🤔

bin/run-browserstack-tests.js
Turbo87: agreed, changed to `const`

PR #17266, opened 2018-12-06
Add failing test demonstrating an issue updating a component property from `didReceiveAttrs` in test mode

GavinJoyce:
This adds a failing test demonstrating an issue that I came across in our app when upgrading from 3.1 to 3.2 (the issue appears to be in 3.2 -> 3.5 and canary). The test added in this PR fails with a backtracking error:

screen shot 2018-12-06 at 10 24 30

I've also reproduced the issue in [this 3.5 ember app](https://github.com/GavinJoyce/ember-did-receive-test-bug). The error only occurs when testing the component, it works fine when running the app in development mode:

screen shot 2018-12-06 at 10 02 29

screen shot 2018-12-06 at 10 02 22

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/components/curly-components-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/components/curly-components-test.js
@@ -3279,6 +3279,24 @@ moduleFor(
       this.runTask(() => set(this.context, 'foo', 5));
     }
 
+    ['@test properties can be updated in `didReceiveAttrs`']() {
+      this.registerComponent('foo-bar', {
+        ComponentClass: Component.extend({
+          didReceiveAttrs() {
+            this._super();
+
+            this.set('name', 'Updated');
+          },
+        }),
+
+        template: 'name: {{name}}',
+      });
+
+      this.render(`{{foo-bar name=name}}`, { name: 'Provided' });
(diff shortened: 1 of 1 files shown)


rwjblue (commented):
Hmm. This does seem like backflow to me (setting inside the components didReceiveAttrs flows back upwards to the invocation / callsite)...

GavinJoyce:
Sorry, I'm mistaken about `{{foo-bar name=name}}` not raising a backtracking assertion in dev mode in 3.5, [it does](https://github.com/GavinJoyce/ember-did-receive-test-bug/pull/2):

3.5:

screen shot 2018-12-07 at 10 30 14

3.1:

screen shot 2018-12-07 at 10 30 51

GavinJoyce:
The behaviour seems to have changed here: https://github.com/emberjs/ember.js/pull/16185

rwjblue:
Sorry to have taken so long to reply here. Moving the invocation of `didReceiveAttrs` into the component manager (and out of the constructor flow) was specifically to help us ensure that auto-tracking things work properly (as well as remove more special behaviors around component constructors).

It is unfortunate that this bit y'all, but I think asserting in this context is correct behavior. Sorry again for the long delay in replying.

PR #17268, opened 2018-12-06
Update old jQuery versions for tests

simonihmig:
Updates the suite of tests for old jQuery versions to run with latest of
1.10.x, 1.12.x and 2.2.x.

Unlocks #17227 which is failing due to jQuery 1.8 presumably not handling
events on SVGs correctly.


rwjblue:
Thank you @simonihmig!

PR #17270, opened 2018-12-06
Update Broccoli dev dependencies

Turbo87:
see commit messages

I have manually confirmed that the ember build -prod output is the same as before

/cc @rwjblue


rwjblue:
Needs a rebase

rwjblue (approved):

PR #17275, opened 2018-12-06
Update `broccoli-uglify-sourcemap` to v2.2.0

Turbo87:
This is a somewhat significant change as we are now using `terser` under the hood, instead of `uglify-js/es`. On first glance, the slightly changed output looks alright, but this should probably be verified a little more thoroughly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,7 +117,7 @@
     "broccoli-source": "^1.1.0",
     "broccoli-string-replace": "^0.1.2",
     "broccoli-typescript-compiler": "^4.0.1",
-    "broccoli-uglify-sourcemap": "^2.0.2",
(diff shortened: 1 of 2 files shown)


rwjblue:
In general the `ember.min.js` file that we generate is actually not used (even in a consuming app during `ember build -prod`).

rwjblue (approved):

Turbo87:
okaaaay... then... why do we generate it? 😅

rwjblue:
Why not!?!?!!?

rwjblue:
yeah, but its CI time so.....

rwjblue:
(note: I do agree with you)

PR #17277, opened 2018-12-06
Remove unused `ember.min.js` build asset

Turbo87:
According to https://github.com/emberjs/ember.js/pull/17275#issuecomment-445059036 the `ember.min.js` asset is not being used anymore. This PR removes the asset from the build pipeline, since the uglify step takes a significant amount of time.

/cc @rwjblue


rwjblue (approved):

bin/run-tests.js
```diff
@@ -90,9 +90,6 @@ function generateBuiltTests() {
   testFunctions.push(function() {
     return run(common + '&nolint=true');
   });
-  testFunctions.push(function() {
-    return run(common + '&dist=min&prod=true');
```
rwjblue: Hmm, this reminds me, once upon a time we had legit issues when minified so we started running our test suite against the minified assset.

I'm not 100% sure that is still a valid concern/issue, but it warrants further thought...

rwjblue (changes requested):
Will add this for review with the rest of the team to confirm...

bin/run-tests.js
Turbo87: sounds like a valid concern, but as you said Ember.js is no longer controlling its own minification process, so I'm not sure it's helpful to run the tests against just that single version of `terser` while out in the wild other versions and possibly even other minifiers are being used 🤔

rwjblue: Yep, totally agree.

tomdale:
As discussed in the core team meeting today, we are in favor of removing this but want to make sure we have _some_ sort of end-to-end smoke test of Ember and Ember CLI working together, to catch minifier bugs but also the many other categories of regressions that can happen at integration points.

Until we get something like that in place, would it be possible to modify the emitted minified version to produce a `console.warn` that the minified version is deprecated and will stop being produced in the future? Even though I don't think anyone is actually using this file, if there are folks they'll at least get a bit of heads up.

Turbo87:
> Until we get something like that in place, would it be possible to modify the emitted minified version to produce a `console.warn` that the minified version is deprecated and will stop being produced in the future?

should be possible

I'll close this PR for now as we'll probably keep the minified version around for a few releases before we can remove it.

PR #17296, opened 2018-12-07
 Update `eslint` to v5.9.0 

Turbo87:
/cc @rwjblue


package.json
```diff
@@ -77,6 +77,19 @@
     "resolve": "^1.6.0"
   },
   "devDependencies": {
+    "@babel/helper-module-imports": "^7.0.0",
```
rwjblue: This is going to cause a conflict for you...

rwjblue (approved):

package.json
rwjblue: Huh, I guess not?!

Turbo87: as long as the change is identical in both branches it's not a conflict 😉

PR #17307, opened 2018-12-08
QUnit: Remove obsolete polyfill

Turbo87:
We use a QUnit version now that includes the PR that this polyfill was extracted from.


rwjblue (approved):
Awesome, thank you!

PR #17309, opened 2018-12-08
Remove obsolete/broken `livereload` option

Turbo87:
Ember CLI already injects a livereloader and `livereload.js` is not even available when run via `ember test --server`


rwjblue (approved):
Ember CLI doesn’t inject it, ember-cli-inject-livereload addon does (for ember s) and testem itself handles this (for ember test —server).

Regardless of that, I don’t actually think any of this code works at the moment anyways so it’s still good to remove...

rwjblue:
Looks like a conflict here, needs a rebase.

Turbo87:
@rwjblue fixed

PR #17310, opened 2018-12-08
tests: Remove unused and broken `printTestCounts()` functions

Turbo87:
This function is using jQuery even though it might not be available, and it is querying a selector which no longer seems to be used by QUnit. In addition to that it is not called from anywhere and since it's broken it doesn't seem to be used manually anymore either.

### DIFF
--- a/tests/index.html
+++ b/tests/index.html
@@ -39,18 +39,6 @@
       (function() {
         window.ENV = window.ENV || {};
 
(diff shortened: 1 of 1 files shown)


rwjblue:
At some point, I added console logging to emit the counts by default (via a testDone in tests/index.html).

PR #17321, opened 2018-12-09
[docs beta] refresh yarn lock so that it picks up rsvp 4.8.4

toddjordan:
Addresses https://github.com/ember-learn/ember-api-docs/issues/537

in yarn lock rsvp 4 is split into 2 entries, one taking rsvp 4.8.4 and one taking 4.8.2.  At yarn install time, yarn is picking 4.8.2 to install, and thus picking up older docs.
See: https://github.com/emberjs/ember.js/blob/beta/yarn.lock#L8451-L8459

For this PR I'm just running `yarn upgrade rsvp@^4.8.2`, which just consolidates the the 4.x releases under one entry.

This pr is point to beta.  its already fixed in release, probably from a refresh of the yarn lock.


rwjblue (approved):

rwjblue:
Restarted failed BrowserStack job...

rwjblue:
Hmm, failed again.  Restarting, but if it fails a third time we should try to dig in...

PR #17328, opened 2018-12-10
[BUGFIX release] Ensure that delayed transition retrys work

chadhietala:


package.json
```diff
@@ -146,7 +146,7 @@
     "puppeteer": "^1.3.0",
     "qunit": "^2.5.0",
     "route-recognizer": "^0.3.4",
-    "router_js": "^6.1.3",
+    "router_js": "^6.1.4",
```
rwjblue: Hmm, this is a bit odd, seems like we are getting 6.2.1 instead of 6.1.4. Would you mind pinning this minimum version to the current latest release?

chadhietala: 6.2.0 has the `buildRouteInfoMetadata` stuff in it.

rwjblue: But as you can see from the `yarn.lock` we are _already_ getting 6.2.1 in the `yarn.lock` so we might as well "just do it"...

rwjblue:
This will also need to be cherry picked into the beta branch to prevent a regression in 3.7.

PR #17330, opened 2018-12-10
[FEATURE] buildRouteInfoMetadata more test assertions

chadhietala:
This assertions are being added to ensure the timing semantics of `buildRouteInfoMetadata` as router.js's test suite currently doesn't allow us to correctly test this.


rwjblue (approved):
Seems good to me to have them here anyways, but we should hopefully get the router.js suite in better shape so we can have earlier warning...

PR #17340, opened 2018-12-11
Remove Hooks from qunit-rfc-232 util-test blueprint

hakubo:
It's not used
so a dev have to delete it by hand after generating a test


rwjblue (commented):
Seems good to me! I think you will also need to update the test fixtures we use (to fix the failing tests).

rwjblue:
Awesome, thank you!

PR #17344, opened 2018-12-11
internal-test-helpers: Move `NodeQuery` class out of the `test-cases` folder

Turbo87:
`NodeQuery` is a utility class, not an (abstract) test case.


rwjblue:
LGTM, pending CI run (travis has a delay kicking off jobs from webhooks ATM)

PR #17346, opened 2018-12-12
[BUGFIX release] Publish builds to s3 on release & beta channels even

### LINKED ISSUE #17297: release.json not updating
https://s3.amazonaws.com/builds.emberjs.com/release.json references 3.4.4 and is not updating. This is used by ember-try and `getChannelURL`

kategengler:
when there is a tag for the commit.

- Always publish as a tag when publishing locally (vs on Travis)

Fixes #17297

### DIFF
--- a/broccoli/build-info.js
+++ b/broccoli/build-info.js
@@ -20,7 +20,12 @@ function buildInfo(options) {
   let root = (options && options.root) || path.resolve(__dirname, '..');
   let packageVersion = (options && options.packageVersion) || readPackageVersion(root);
   let gitInfo = (options && options.gitInfo) || buildGitInfo(root);
-  let buildInfo = buildFromParts(packageVersion, gitInfo);
+  let buildInfo = buildFromParts(
+    packageVersion,
+    gitInfo,
+    process.env.TRAVIS,
+    process.env.TRAVIS_BRANCH
+  );
   if (!options) {
     cached = buildInfo;
   }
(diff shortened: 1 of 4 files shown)


config/s3ProjectConfig.js
```diff
@@ -63,12 +63,6 @@ function fileObject(baseName, extension, contentType, currentRevision, tag, date
     },
   };
 
-  if (tag) {
```
rwjblue: 👍 - This was never used (because we only publish to S3 for non-tagged builds).

broccoli/build-info.js
```diff
@@ -34,9 +34,9 @@ function buildInfo(options) {
 function buildGitInfo(root) {
   let info = gitRepoInfo(root);
   return {
-    sha: process.env.TRAVIS_COMMIT || info.sha,
-    branch: process.env.TRAVIS_BRANCH || info.branch,
-    tag: process.env.TRAVIS_TAG || info.tag,
+    sha: info.sha,
+    branch: process.env.TRAVIS_BRANCH || info.branch, // Travis builds are always detached
```
rwjblue: What is `info.branch` when detached? 

I kinda feel like we may want to put all the `TRAVIS` smarts in `buildFromParts` directly. What do you think?

rwjblue (commented):
Overall, it looks good to me (and should solve the issues we have been dealing with), only one inline question/comment...

broccoli/build-info.js
kategengler: `info.branch` when detached is null.

kategengler: @rwjblue Can you restate or clarify "I kinda feel like we may want to put all the TRAVIS smarts in buildFromParts directly. What do you think?"?

rwjblue: Basically, just that this could be `branch: info.branch` and down below where all the other travis detection logic.

kategengler: 👍

rwjblue (approved):

PR #17355, opened 2018-12-12
tests/glimmer: Remove unnecessary `internal-test-helpers` reexports

Turbo87:
These imports don't seem to serve any purpose other than making the imports harder to follow. By removing the reexports it was also possible to merge some imports that ultimately imported from the same module.

The PR also reorders a few imports in the order of: test framework, ember, other deps, relative imports

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/components/target-action-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/components/target-action-test.js
@@ -1,11 +1,13 @@
+import { moduleFor, RenderingTestCase, ApplicationTestCase } from 'internal-test-helpers';
+
 import { assign } from '@ember/polyfills';
-import { moduleFor, RenderingTestCase, ApplicationTestCase } from '../../utils/test-case';
-import { strip } from '../../utils/abstract-test-case';
+import { strip } from 'internal-test-helpers';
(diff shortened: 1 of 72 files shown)


packages/@ember/-internals/glimmer/tests/integration/components/target-action-test.js
```diff
@@ -1,12 +1,14 @@
+import { moduleFor, RenderingTestCase, ApplicationTestCase } from 'internal-test-helpers';
+
 import { assign } from '@ember/polyfills';
-import { moduleFor, RenderingTest, ApplicationTest } from '../../utils/test-case';
-import { strip } from '../../utils/abstract-test-case';
+import { strip } from 'internal-test-helpers';
```
rwjblue: Shouldn't this get collapsed?

rwjblue (approved):
FWIW, this abstraction existed when we needed to use the same set of tests to test against glimmer v1 and glimmer v2.

Definitely agree that we don't need it any longer...

packages/@ember/-internals/glimmer/tests/integration/components/target-action-test.js
Turbo87: yep, totally missed that. should be fixed now.

PR #17356, opened 2018-12-12
internal-test-helpers: Implement `get/set/unsetContext()` functions

Turbo87:
This is roughly similar to what `@ember/test-helpers` does to enable importable test helpers that work on the test context.

### DIFF
--- a/packages/internal-test-helpers/lib/module-for.js
+++ b/packages/internal-test-helpers/lib/module-for.js
@@ -33,7 +33,7 @@ export function setupTestClass(hooks, TestClass, ...mixins) {
       promises.push(instance.afterEach());
     }
 
-    return all(promises).then(() => {
+    return all(promises).finally(() => {
       unsetContext();
     });
   });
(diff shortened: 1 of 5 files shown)


packages/internal-test-helpers/lib/module-for.js
```diff
@@ -29,7 +33,9 @@ export function setupTestClass(hooks, TestClass, ...mixins) {
       promises.push(instance.afterEach());
     }
 
-    return all(promises);
+    return all(promises).then(() => {
```
rwjblue: Should probably be `finally` (we want to unset even if rejected)

rwjblue (changes requested):

packages/internal-test-helpers/lib/module-for.js
Turbo87: agreed, fixed

rwjblue (approved):

PR #17360, opened 2018-12-12
tests/glimmer: Simplify `TrustedContentTest` class

Turbo87:
`assertStableRerender()` is identical to the `super` implementation, and `assertInvariants()` is not used at all


Turbo87:
apparently `assertInvariants()` is still being used by the tests generate in one of the classes deeper down the inheritance tree.......

Issue #17190, opened 2018-11-08
v3.6.0-beta.3 JQuery.extend (recursively) is broken


pieter-v:
We use some jquery plugins (e.g.: bootstrap-daterangepicker), which uses the extend method from jquery. In v3.6.0-beta.3, this fails as jquery recursively iterates through the `[]` property of an array (we use the array prototype extensions of Ember)

**ember-source@3.5.1**
`$.extend(true, {}, {a:['a']})` gives
```
{a: Array(1)}
```

**ember-source@3.6.0-beta.3**
`$.extend(true, {}, {a:['a']})` gives
```
jquery.js:282 Uncaught RangeError: Maximum call stack size exceeded
    at Function.jQuery.extend.jQuery.fn.extend (jquery.js:282)
    at Function.jQuery.extend.jQuery.fn.extend (jquery.js:282)
    at Function.jQuery.extend.jQuery.fn.extend (jquery.js:282)
    at Function.jQuery.extend.jQuery.fn.extend (jquery.js:282)
    at Function.jQuery.extend.jQuery.fn.extend (jquery.js:282)
    at Function.jQuery.extend.jQuery.fn.extend (jquery.js:282)
    at Function.jQuery.extend.jQuery.fn.extend (jquery.js:282)
    at Function.jQuery.extend.jQuery.fn.extend (jquery.js:282)
    at Function.jQuery.extend.jQuery.fn.extend (jquery.js:282)
    at Function.jQuery.extend.jQuery.fn.extend (jquery.js:282)
```

rwjblue:
This is using the same version of jQuery between the two?

I’m not immediately aware of a change that would have caused this. We will probably need to bisect to track it down...

pieter-v:
To reproduce, I created a new project (so no extra dependencies)
```
ember new extend --yarn
```
This installs ember@3.5.1
I run the default application and in the browser console I executed the command
```
$.extend(true, {}, {a:['a']})
```
Then I upgraded ember
```
yarn upgrade ember-source@3.6.0-beta.3
```
And this gives the error

pieter-v:
JQuery uses `for ( name in options )` construct to iterate over the properties.

In ember 3.5.1
```
Object.getOwnPropertyDescriptor([].__proto__, '[]')
```
returns
```
{get: ƒ, set: undefined, enumerable: false, configurable: true}
```
and in ember 3.6.0-beta.3
the result is
```
[...]

PR #17374, opened 2018-12-14
[BUGFIX release] Fix cyclic references on Array.prototype

jacobq:
My first non-doc `ember.js` PR so please be kind :)

* Should detect the regression that causes #17190 
* Test currently fails (correctly) on master but passes on v3.5.1
* Not sure if the test is located in the right place or applied to the right object(s)
* Note: this only checks the first level (e.g. won't catch a nested reference)

### DIFF
--- a/packages/@ember/-internals/runtime/lib/mixins/array.js
+++ b/packages/@ember/-internals/runtime/lib/mixins/array.js
@@ -171,6 +171,18 @@ export function isArray(_obj) {
   return false;
 }
 
+/*
+  This allows us to define computed properties that are not enumerable.
+  The primary reason this is important is that when `NativeArray` is
+  applied to `Array.prototype` we need to ensure that we do not add _any_
+  new enumerable properties.
+*/
+function nonEnumerableComputed() {
+  let property = computed(...arguments);
+  property.enumerable = false;
+  return property;
+}
+
 // ..........................................................
 // ARRAY
 //
@@ -273,7 +285,7 @@ const ArrayMixin = Mixin.create(Enumerable, {
     @return this
     @public
   */
-  '[]': computed({
+  '[]': nonEnumerableComputed({
     get() {
       return this;
     },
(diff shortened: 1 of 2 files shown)


jacobq:
I am looking at this again as it seems to have failed to correctly identify the point in git commit history was introduced. (It appears to pass against 246435e7260f45d9e5cc35a4a8b4de4f011f3625 although [that seems to be where the regression happened](https://github.com/emberjs/ember.js/issues/17190#issuecomment-447501532))

jacobq:
@rwjblue Thanks for taking this on. Could you help me understand something here? How come when I checkout 246435e7260f45d9e5cc35a4a8b4de4f011f3625 and build it, running the script below in node shows the problem, but running the [test that import `NativeArray`](https://github.com/jacobq/ember.js/blob/3029e9df6ef110cdf4f68a379600e4b5308db86f/packages/%40ember/-internals/runtime/tests/array/apply-test.js#L7:L29) passes?

```js
global.EmberENV = {
  EXTEND_PROTOTYPES: false,
};

const Ember = require('./dist/ember.debug.js');

const isOK = (a) => {
  for (let p in a) {
    if (a[p] === a) {
      return false;
    }
  }
  return true;
}

const check = (klass) => {
  const a = new klass();
  const result = isOK(a);
  console.log(`${klass} ${result ? 'is ok' : 'is NOT OK -- cycle detected!'}`);
}

class ExtendedArray extends Array {}
Ember.NativeArray.apply(ExtendedArray.prototype);

[Array, ExtendedArray].forEach(check);
// outputs:
// function Array() { [native code] } is ok
// class ExtendedArray extends Array {} is NOT OK -- cycle detected!
```

jacobq:
Also, FWIW, there's an [`is-circual`](https://github.com/tjmehta/is-circular/blob/master/index.js) module that could be used instead of my home-made function.

rwjblue:
Just pushed a commit fixing the issue (details in the comments and commit message).

PR #17394, opened 2018-12-19
mixins/array: Remove deprecated code

Turbo87:
This code was flagged as deprecated and to-be-removed with Ember 3.5.0, which means we can get rid of it now! 🎉


Turbo87:
CI finally 🍏 ...

rwjblue (approved):

PR #17406, opened 2018-12-22
[BUGFIX] fix issue around how `undefined` stored in `_values` chain

### LINKED ISSUE #17022: DEFAULT_GETTER_FUNCTION returns wrong value if actual value is set to undefined
Hi!
While I was trying to upgrade my app by using ember-decorators, I encountered an issue with Ember's `DEFAULT_GETTER_FUNCTION`, `Meta#_findInherited2` to be precise. The issue is unrelated to the decorators themselves, but I noticed it because of the `observes` decorator, which is using `Ember.addObserver` instead of `Ember.observer` and changing my app code seemed like a trivial thing to do.

It looks like this issue is present in all recent Ember releases.

Here's an ember-twiddle which demonstrates the problem:
https://ember-twiddle.com/8c7201ea559a9591219ea58edce56d52?openFiles=tests.foo-test.js%2C

----

I don't know much about the internals of Ember's object system, but the core of this issue is that you can't set a regular property to `undefined` if it has an observer attached via the `Ember.observer` method. Observers added via `Ember.addObserver` are working fine, though.

The reason for this is the difference of the two getter methods, `DEFAULT_GETTER_FUNCTION` and `INHERITING_GETTER_FUNCTION`, and the `Meta` methods which they are using:
https://github.com/emberjs/ember.js/blob/v3.4.4/packages/ember-metal/lib/properties.ts#L72-L100

[`DEFAULT_GETTER_FUNCTION`](https://github.com/emberjs/ember.js/blob/v3.4.4/packages/ember-metal/lib/properties.ts#L72-L79), which seems to be the getter for observed regular properties if `Ember.observer` is being used, is calling [`Meta#peekValues`](https://github.com/emberjs/ember.js/blob/v3.4.4/packages/ember-meta/lib/meta.ts#L598-L600), which itself is calling [`Meta#_findInherited2`](https://github.com/emberjs/ember.js/blob/v3.4.4/packages/ember-meta/lib/meta.ts#L157-L197). As you can see in this method, the meta-object's value will only be returned if it's not `undefined`, otherwise the value of one of the parent meta-objects will be returned. This means that if you set the property to `undefined`, it'll return its parent value on the next lookup.

[`INHERITING_GETTER_FUNCTION`](https://github.com/emberjs/ember.js/blob/v3.4.4/packages/ember-metal/lib/properties.ts#L81-L100) on the other hand, which seems to be used as the getter for regular properties which are observed via `Ember.addObserver`, is calling [`Meta#readInheritedValue`](https://github.com/emberjs/ember.js/blob/v3.4.4/packages/ember-meta/lib/meta.ts#L606-L623), which is very similar to `Meta#_findInherited2`. However, here, it is also being checked whether a value exists on the meta-object via `subkey in map` and an `UNDEFINED` symbol will be returned if it doesn't exist. This means that `undefined` values are taken into account, which is the proper thing to do.

On a side note, wouldn't it make sense using `hasOwnProperty` here instead of checking for existing properties with `in`, even if it's just a plain object, as it is much faster, and also removing the first part of the if-statement's disjunction, as it is redundant?

Like I've already said, I'm unfamiliar with how Ember works under the hood, but why are there two similar methods being used for two different getter methods?

Thanks!

bekzod:
using `UNDEFINED` const when storing `undefined` values in `_values`

fixes #17022

### DIFF
--- a/packages/@ember/-internals/meta/lib/meta.ts
+++ b/packages/@ember/-internals/meta/lib/meta.ts
@@ -792,7 +792,7 @@ export interface Meta {
   writeValues(subkey: string, value: any): void;
   peekValues(key: string): any;
   deleteFromValues(key: string): any;
-  readInheritedValue(key: string, subkey: string): any;
+  readInheritedValue(key: string): any;
   writeValue(obj: object, key: string, value: any): any;
 }
 
@@ -808,34 +808,20 @@ if (DEBUG) {
     );
 
     let map = this._getOrCreateOwnMap('_values');
-    map[subkey] = value;
+    map[subkey] = value === undefined ? UNDEFINED : value;
   };
 
-  Meta.prototype.peekValues = function(subkey: string) {
-    return this._findInherited2('_values', subkey);
+  Meta.prototype.peekValues = function(key: string) {
+    let val = this._findInherited2('_values', key);
+    return val === UNDEFINED ? undefined : val;
   };
 
-  Meta.prototype.deleteFromValues = function(subkey: string) {
-    delete this._getOrCreateOwnMap('_values')[subkey];
+  Meta.prototype.deleteFromValues = function(key: string) {
+    delete this._getOrCreateOwnMap('_values')[key];
   };
 
-  Meta.prototype.readInheritedValue = function(key, subkey) {
-    let internalKey = `_${key}`;
-
-    let pointer: Meta | null = this;
-
-    while (pointer !== null) {
-      let map = pointer[internalKey];
-      if (map !== undefined) {
-        let value = map[subkey];
-        if (value !== undefined || subkey in map) {
-          return value;
-        }
-      }
-      pointer = pointer.parent;
-    }
-
-    return UNDEFINED;
+  Meta.prototype.readInheritedValue = function(key: string) {
+    return this._findInherited2('_values', key);
   };
 
   Meta.prototype.writeValue = function(obj: object, key: string, value: any) {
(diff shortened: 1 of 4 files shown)


packages/@ember/-internals/meta/lib/meta.ts
```diff
@@ -785,34 +785,27 @@ if (DEBUG) {
...
-          return value;
-        }
-      }
-      pointer = pointer.parent;
+  Meta.prototype.readInheritedValue = function(key: string) {
+    let val = this._findInherited2('_values', key);
+    if (val === UNDEFINED) {
+      val = undefined;
+    } else if (val === undefined) {
+      val = UNDEFINED;
```
rwjblue: I don't fully understand this branch, why would we return `UNDEFINED` here?

bekzod: I know it is confusing wanted to comment there but now I have better idea improving it, `UNDEFINED` has different meaning in `readInheritedValue` it means it reached the end of inheritance chain of `_values` and have found no value, in `this._findInherited2` `UNDEFINED` means there is `undefined` value found in `_values`, but now I think better way doing it would be removing all the translations above and changing usage of `UNDEFINED` downstream

bekzod:
rebase, maybe worth merging has reproduction and fix

rwjblue (approved):

rwjblue:
Thanks @bekzod !

PR #17411, opened 2018-12-26
[Feature] Update initializer blueprint for ember-mocha 0.14

simonihmig:
Related to #16863

PR #17412, opened 2018-12-26
[Feature] Update instance-initializer blueprint for ember-mocha 0.14

simonihmig:
Related to #16863

*Note: this blueprint and the one for initializers (#17411) follow what the existing QUnit RFC232 blueprints do. However I wondered why for those qunit-based blueprints, the initializer test calls `setupTest()`, while the instance-initializer does not. For now the two new PRs for ember-mocha 0.14 do the same, but if there is no good reason for this, I can update all of those...*

### DIFF
--- a/blueprints/instance-initializer-test/mocha-rfc-232-files/__root__/__testType__/__path__/__name__-test.js
+++ b/blueprints/instance-initializer-test/mocha-rfc-232-files/__root__/__testType__/__path__/__name__-test.js
@@ -1,0 +1,28 @@
+import { expect } from 'chai';
+import { describe, it, beforeEach, afterEach } from 'mocha';
+import Application from '@ember/application';
+import { initialize } from '<%= modulePrefix %>/instance-initializers/<%= dasherizedModuleName %>';
+<% if (destroyAppExists) { %>import destroyApp from '../../helpers/destroy-app';<% } else { %>import { run } from '@ember/runloop';<% } %>
+
+describe('<%= friendlyTestName %>', function() {
+  beforeEach(function() {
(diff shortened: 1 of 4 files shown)


blueprints/instance-initializer-test/mocha-rfc-232-files/__root__/__testType__/__path__/__name__-test.js
```diff
@@ -0,0 +1,28 @@
...
+    this.TestApplication.instanceInitializer({
+      name: 'initializer under test',
+      initialize
+    });
+    this.application = this.TestApplication.create({ autoboot: false });
+    this.instance = this.application.buildInstance();
+  });
+  afterEach(function() {
+    <% if (destroyAppExists) { %>destroyApp(this.application);<% } else { %>run(this.application, 'destroy');<% } %>
+    <% if (destroyAppExists) { %>destroyApp(this.instance);<% } else { %>run(this.instance, 'destroy');<% } %>
```
Turbo87: I would assume that the instance needs to be destroyed before the application.

@rwjblue thoughts?

rwjblue: Yes, confirm, we should destroy the `ApplicationInstance` instance before the `Application` instance. (naming is hard 😭 )

simonihmig: Updated this! 

Here's a separate PR for QUnit: https://github.com/emberjs/ember.js/pull/17712

PR #17422, opened 2019-01-03
get / set optimization 

bekzod:
* removed `isPath` cache
* added internal `getKey`, `setKey` and `setPath`

benchmark result: https://cl.ly/bcd1268bd7ec/results.pdf


bekzod:
needs another benchmark, since I fixed failing test

bekzod:
I think things diverged a lot better close this I think

PR #17426, opened 2019-01-04
[BUGFIX beta] fix 'strict mode does not allow function declarations'

stukalin:
We've discovered that on ios9 (at least) there is an error `strict mode does not allow function declarations` which prevents the application to work normally. We assume, although we're not quite sure, it was released with the v3.6.

Anyways, it's related to the fact that the function `cleanURL` was declared within an `if` block in the `router.ts`.

It turns out that the eslint rule which cares about such issues had been disabled. After re-enabling it I found out that there are only 2 issues (the one with the router and another one in the test). 

I guess it's worth making the rule enabled; I also fixed the 2 failures by assigning functions to variables.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -62,8 +62,6 @@ module.exports = {
         'no-unused-vars': 'off',
         'no-undef': 'off',
 
-        'no-inner-declarations': 'off',
-
         'import/export': 'off',
         'import/named': 'off',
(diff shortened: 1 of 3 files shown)


rwjblue (approved):

rwjblue:
Good sleuthing, and thanks for fixing it!

PR #17429, opened 2019-01-04
bin/publish_builds cleanup

mike-north:
Backtick command substitution `` `...` `` is legacy syntax with several issues.

- It has a series of undefined behaviors related to quoting in POSIX.
- It imposes a custom escaping mode with surprising results.
- It's exceptionally hard to nest.

source: https://github.com/koalaman/shellcheck/wiki/SC2006


rwjblue (approved):

rwjblue:
Seems good to me, thanks for the cleanup @mike-north!

PR #17451, opened 2019-01-08
Add failing test for query param scoping

wagenet:
The issue here is that for query param scoping, it is assumed that the only thing that matters are the URL params on the final route. However, it is entirely possible (as shown in the test) that this might not be sufficient to determine uniqueness. It's not clear to me how best to fix this since in some cases this assumption might be correct (e.g. the params are just the ids of Ember Data records) but in others it may not be (e.g. the final route's params aren't sufficient to determine the model).

It should also be noted that the guides say that the query params are tied to the model. This is definitely not correct as we see here. They are tied to the url params of the last route.

### DIFF
--- a/packages/ember/tests/routing/query_params_test.js
+++ b/packages/ember/tests/routing/query_params_test.js
@@ -870,6 +870,50 @@ moduleFor(
       assert.equal(indexController.get('omg'), 'lol');
     }
 
+    ['@test query params are based on actual model not on url params of lowest route only'](
+      assert
+    ) {
+      assert.expect(2);
+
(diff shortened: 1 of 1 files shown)


chadhietala:
Is this a regression as of 3.6?

wagenet:
@chadhietala I believe it's been present since at least 3.4.6 as we discovered it in our app which is stuck on 3.4 for unrelated reasons.

locks:

wagenet:
@locks I just rebased. Looks like CI still fails.

wagenet:
Rebased again just now.

PR #17452, opened 2019-01-08
Add a failing test for query params cache with resetNamespace

wagenet:
The basic issue here is that the cache key calculator uses the current route name to determine the prefix for the key which enables a correct value lookup. However, if you use `resetNamespace` the prefix won't match and it calculates the prefix incorrectly. In this example, it will try get the value from `parent.child.child_id` (IIRC) and instead of that getting the value of `child_id` from the `parent.child` route, it will try to get the value of `child.child_id` from the `parent` route.

### DIFF
--- a/packages/ember/tests/routing/query_params_test.js
+++ b/packages/ember/tests/routing/query_params_test.js
@@ -870,6 +870,46 @@ moduleFor(
       assert.equal(indexController.get('omg'), 'lol');
     }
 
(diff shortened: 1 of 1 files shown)


locks:

kategengler:
Is this still relevant?

wagenet:
I just rebased, lets see if it still fails!.

PR #17459, opened 2019-01-09
intermediateTransitionTo fails when router service is active

### LINKED ISSUE #17430: RouterService breaks Route#intermediateTransitionTo
I've encountered a bug while upgrading to Ember `3.6.1` and using the RouterService.

In the `RouterService`'s `init` hook, it sets up listeners on the `EmberRouter` for the `routeWillUpdate` and `routeDidUpdate` events. These are used as a pass-through for the same events on the RouterService itself. However, in Ember's debug build, they also freeze the transition's routeInfo objects (`from` and `to` properties). See here:
https://github.com/emberjs/ember.js/blob/v3.6.1/packages/%40ember/-internals/routing/lib/services/router.ts#L272-L285
https://github.com/emberjs/ember.js/blob/v3.6.1/packages/%40ember/-internals/routing/lib/services/router.ts#L13-L21

This can cause issues in certain scenarios where the transition parameter is undefined and accessing `from` or `to` on the `undefined` transition will throw errors.

In my case, the transition parameter was undefined after calling `transition.abort()` in the app's `SomeInitialRoute#beforeModel` hook and then calling `intermediateTransitionTo`. Consider this example:

```js
// this needs to be one of the app's initial routes, so for example the IndexRoute
export default IndexRoute extends Route {
  @service router;

  async beforeModel( transition ) {
    // access the RouterService, so it gets initialized
    this.get( "router" );

    // don't let the user access this route for now
    transition.abort();

    // show the loading screen instead
    this.intermediateTransitionTo( "loading" ); // <-- throws errors ⚠️⚠️⚠️

    // do some important stuff
    await new Promise( resolve => setTimeout( resolve, 1000 ) );
    // and eventually continue, eg. retry transition or go to a different route
    this.router.transitionTo( "foo" );
  }
}
```

I'm not entirely sure if this undefined transition parameter in the triggered events is intended. If it is, then the freezeRouteInfo method needs to be fixed. If it's not, then it must be an issue with RouterJS and its `activeTransition` property, which will stay undefined if the initial transition gets aborted, but I don't know anything about its implementation. However, this is where the undefined transition is coming from:
https://github.com/tildeio/router.js/blob/v6.2.0/lib/router/router.ts#L231-L238

pieter-v:
Fixes #17430

### DIFF
--- a/packages/@ember/-internals/routing/lib/services/router.ts
+++ b/packages/@ember/-internals/routing/lib/services/router.ts
@@ -10,13 +10,15 @@ import { extractRouteArgs, resemblesURL, shallowEqual } from '../utils';
 
 let freezeRouteInfo: Function;
 if (DEBUG) {
-  freezeRouteInfo = (transition: Transition) => {
-    if (transition.from !== null && !Object.isFrozen(transition.from)) {
-      Object.freeze(transition.from);
-    }
+  freezeRouteInfo = (transition: Transition | undefined) => {
+    if (transition) {
+      if (transition.from !== null && !Object.isFrozen(transition.from)) {
+        Object.freeze(transition.from);
+      }
 
-    if (transition.to !== null && !Object.isFrozen(transition.to)) {
-      Object.freeze(transition.to);
+      if (transition.to !== null && !Object.isFrozen(transition.to)) {
+        Object.freeze(transition.to);
+      }
     }
   };
 }


rwjblue (commented):
Can you also add a test?

bastimeyer:
This only prevents the errors to be thrown. The `routeWillChange` event will still be triggered with an undefined transition, and I don't think that's intended behavior.

chadhietala:
@pieter-v thank you for the PR, however as @bastimeyer mentioned the issue is actually in Router.js. I have https://github.com/tildeio/router.js/pull/277 open, which fixed the issue.

PR #17470, opened 2019-01-11
[FEAT] Implements the Computed Property Modifier deprecation RFCs

pzuraq:
- Deprecates `.property()`
- Deprecates `.volatile()`
- Deprecates `clobberSet` internally
- Keeps injected properties and the `store` property on routes
  overridable for testing purposes
- Rewrites lots of usages in tests. For some reason `volatile` was
  frequently used in tests when it was not needed.
- Adds `additionalDependentKeys` array to `map`, `filter` and `sort`
  array functions. `sort` was overlooked in RFC, but suffers from the
  same problem so it shouldn't need an additional RFC.
- Adds tests for `additionalDependentKeys`

### DIFF
--- a/packages/@ember/-internals/metal/lib/computed.ts
+++ b/packages/@ember/-internals/metal/lib/computed.ts
@@ -1,7 +1,7 @@
 import { meta as metaFor, peekMeta } from '@ember/-internals/meta';
 import { inspect } from '@ember/-internals/utils';
 import { EMBER_METAL_TRACKED_PROPERTIES } from '@ember/canary-features';
-import { assert, warn } from '@ember/debug';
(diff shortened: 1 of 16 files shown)


pzuraq:
Deprecation doc PR [here](https://github.com/ember-learn/deprecation-app/compare/add-computed-property-modifier-deprecations?expand=1)

rwjblue (approved):

rwjblue (approved):
Awesome job, thank you!

rwjblue (approved):
Deprecation guide PR landed in https://github.com/ember-learn/deprecation-app/pull/254, and I confirmed URL's for deprecations...

PR #17479, opened 2019-01-12
Restore fallback for `window.performance` in FastBoot

bobisjan:
This fallback was accidentally removed in #17456 (the issue appeared here 
https://github.com/ember-fastboot/ember-cli-fastboot/issues/596#issuecomment-453771176).

### DIFF
--- a/tests/node/instrumentation-test.js
+++ b/tests/node/instrumentation-test.js
@@ -1,0 +1,27 @@
+'use strict';
+
+var path = require('path');
+var distPath = path.join(__dirname, '../../dist');
+var emberPath = path.join(distPath, 'ember.debug');
+
+QUnit.module('instrumentation', function(hooks) {
+  hooks.afterEach(function() {
+    delete global.Ember;
+    delete require.cache[emberPath + '.js'];
+  });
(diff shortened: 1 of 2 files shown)


rwjblue:
Great catch, would you mind adding a test so we can avoid a regression here in the future? I’m thinking something in `tests/node` would work...

rwjblue:
Also, we need to check if #17456 landed in the beta branch, if it did we need this to be `[BUGFIX beta]`.

tests/node/instrumentation-test.js
```diff
@@ -0,0 +1,24 @@
...
+    var Ember = require(emberPath);
+
+    var result = Ember.instrument('render', {}, function() {
+      return 'hello';
+    });
+
+    assert.equal(result, 'hello', 'called block');
+
+    global.window = _originalWindow;
+    delete global.Ember;
```
rwjblue: Can you move the cleanup to hooks.afterEach? As written this cleanup would not happen if an earlier statement errors...

rwjblue:
Awesome, thank you!

PR #17483, opened 2019-01-15
Add (failing) tests for positional params with custom component managers

dfreeman:
@rwjblue has opened https://github.com/glimmerjs/glimmer-vm/pull/900 to fix the underlying issue with positional parameters for custom component managers, but suggested it would be helpful to have test coverage on the Ember side as well.

The tests here are pretty straightforward conversions of the two above them that test named args. I've confirmed that these pass when I apply the change from that PR to `opcode-builder` locally.


rwjblue:
Thank you for working on this @dfreeman! I pulled your commit into #17499 (along with the fix). The fix lands in:

* 3.4.8 (released today)
* 3.7.2 (released today)
* 3.8.0-beta.3 (should be released 2019-01-28)

PR #17487, opened 2019-01-16
Failing test and fix for #17243

### LINKED ISSUE #17243: `Ember.computed.oneWay` teardown is "resetting" the property in an inner component
We’ve found a real stumper of a regression in our app that git bisect says is caused by this commit (https://github.com/emberjs/ember.js/commit/65c20ab632dcd1ecd1256a14fe5ccce83ebc6953) in 3.5.0-beta.3.

We have two components. A property is passed in to the outer component, where we use `computed.oneWay` to duplicate the property, then pass that into the inner component, where it is used directly (not aliased or mutated). Our setup is similar to what is shown in [this Twiddle](https://ember-twiddle.com/46abef18e16555576668b771c16cd45d), though I wasn’t able to repro the bug on Twiddle. The property is computed once with some default values, then again on `didInsertElement` after we have done some measuring.

In production only, it appears that the property gets set to the default, then updated to the correct final value…then set back to the original (incorrect) value. This only happens in the inner component, even though we just pass the value in directly from the outer component and do not alias it or mutate it in the inner component. 

Adding a no-op observer on the `oneWay` property appears to “fix” the issue, which makes sense given the PR description for that commit (https://github.com/emberjs/ember.js/pull/16978), though I admittedly don’t understand that code well enough to understand the mechanism.

gitKrystan:
Previously #16978 tried to balance the amount of `addDependentKeys` and `removeDependentKeys`. However, it removes the dependent keys too eagerly, causing a different kind of imbalance, and in some cases, this causes the property tags to "miss" updates to aliased properties.

This fixes #17243 by only removing dependent keys in `teardown`.

Since #16978 coalesces the work to add dependent keys to happen at most once, it is now incorrect to eagerly remove them in `didUnwatch` because that could happen for a variety of reasons while there are still other interested parties (see attached test cases).

This limits the work of removing dependent keys to `teardown` only. The logic is that, while we want to defer setting up the "link" to the target property as lazily as possible, it is less important to "unlink" it eagerly once the work is done. It _may_ be possible to accomplish this, but the current amount of book-keeping is not sufficient to track the count properly. In our tests, we have created sequences like this:

1. setup (peekWatching = 0, so nothing happens)
2. get (consumed here)
3. willWatch (already consumed, no-op)
4. get (already consumed, no-op)
5. didUnwatch
6. ...

In this case, it would be incorrect to "unlink" at step 5. As of PR #16978, `CONSUMED` is essentially a boolean flag, so it is not sufficient to track the balance, and also, there is no counterpart to `get`, which makes eager "unlinking" impossible without much more book-keeping. It's unclear that it would be worthwhile to do that.

On the other hand, if we only "unlink" on teardown, this ensures that we are only "unlinking" at most once, and it is guaranteed that there are no longer any interested parties.

Fixes #17243

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/components/curly-components-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/components/curly-components-test.js
@@ -13,7 +13,7 @@ import {
 
 import { run } from '@ember/runloop';
 import { DEBUG } from '@glimmer/env';
-import { set, get, observer, on, computed } from '@ember/-internals/metal';
+import { alias, set, get, observer, on, computed } from '@ember/-internals/metal';
 import Service, { inject as injectService } from '@ember/service';
 import { Object as EmberObject, A as emberA } from '@ember/-internals/runtime';
 import { jQueryDisabled } from '@ember/-internals/views';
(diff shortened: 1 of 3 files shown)


chancancode:
cc @bekzod

rwjblue (approved):
Thank you for digging in here! I really appreciate the detailed explanation of the problem.

bekzod:
Looks good 👍

chancancode:
@rwjblue should this be back ported beta/release/lts?

rwjblue:
@chancancode yes, but we need to confirm the first version that had the issue (to make sure we don't go back _too_ far)

chancancode:
@gitKrystan do we know which version first broke this?

gitKrystan:
@chancancode First seen in 3.5.0-beta.3

jasonmit:
Thanks a ton for working on this.  This bug is also breaking our app currently.

ginomiglio:
We're also having issues with this after upgrading to 3.5. Any news on backporting?

chancancode:
@ginomiglio I believe this ended up landing in 3.8. Seeing that we are about to release 3.9 at this point, and that 3.8 is going to become a LTS release, I don't think we would consider backporting it further back. Are there any issues preventing you from upgrading to 3.8?

ginomiglio:
Ah, ok. It'll just force us to upgrade from 3.4 -> 3.8 directly, which will be...fun for our enormous monolithic Ember app 😅

chancancode:
@ginomiglio since both 3.4 and 3.8 are LTS releases, it is actually recommended that you go from the latest 3.4 release to the 3.8 release, since they should be better maintained than the point releases in between. If you ran into other issues, please do report them!

PR #17491, opened 2019-01-18
Fixed error message when super is not called in init()

### LINKED ISSUE #17490: Calling super inside the init raises an invalid error message
Hey there, thank you for all the effort you put into this framework.

I just upgraded a bunch of dependencies and realized that I now have to change my code from the `constructor` to `init` when inheriting a framework object. This was described in detail in the [rfc](https://github.com/emberjs/rfcs/blob/master/text/0337-native-class-constructor-update.md). I then got following error
```
Error: Assertion Failed: You must call `this._super(...arguments);` when overriding `init` on a framework object. Please update <component> to call `this._super(...arguments);` from `init`.
```

I was already calling `this._super(...arguments)` inside the init and found this [comment](https://github.com/ember-decorators/ember-decorators/issues/329#issuecomment-443769531) saying that one should actually call `super.init(...arguments)`.

CodingDive:
Fixes https://github.com/emberjs/ember.js/issues/17490

### DIFF
--- a/packages/@ember/-internals/runtime/lib/system/object.js
+++ b/packages/@ember/-internals/runtime/lib/system/object.js
@@ -80,7 +80,7 @@ if (DEBUG) {
 
     [ASSERT_INIT_WAS_CALLED]() {
       assert(
-        `You must call \`this._super(...arguments);\` when overriding \`init\` on a framework object. Please update ${this} to call \`this._super(...arguments);\` from \`init\`.`,
+        `You must call \`super.init(...arguments);\` or \`this._super(...arguments)\` when overriding \`init\` on a framework object. Please update ${this} to call \`super.init(...arguments);\` from \`init\` when using native classes or \`this._super(...arguments)\` when using \`EmberObject.extend()\`.`,
         this[INIT_WAS_CALLED]
       );
     }
--- a/packages/@ember/-internals/glimmer/tests/integration/components/curly-components-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/components/curly-components-test.js
@@ -2976,7 +2976,7 @@ moduleFor(
 
       expectAssertion(() => {
         this.render('{{foo-bar}}');
-      }, /You must call `this._super\(...arguments\);` when overriding `init` on a framework object. Please update .* to call `this._super\(...arguments\);` from `init`./);
+      }, /You must call `super.init\(...arguments\);` or `this._super\(...arguments\)` when overriding `init` on a framework object. Please update .*/);
     }
 
     ['@test should toggle visibility with isVisible'](assert) {
--- a/packages/@ember/-internals/glimmer/tests/integration/helpers/custom-helper-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/helpers/custom-helper-test.js
@@ -85,7 +85,7 @@ moduleFor(
 
       expectAssertion(() => {
         this.render('{{hello-world}}');
-      }, /You must call `this._super\(...arguments\);` when overriding `init` on a framework object. Please update .* to call `this._super\(...arguments\);` from `init`./);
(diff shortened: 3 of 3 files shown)


packages/@ember/-internals/runtime/lib/system/object.js
```diff
@@ -63,7 +63,7 @@ if (DEBUG) {
 
     [ASSERT_INIT_WAS_CALLED]() {
       assert(
-        `You must call \`this._super(...arguments);\` when overriding \`init\` on a framework object. Please update ${this} to call \`this._super(...arguments);\` from \`init\`.`,
+        `You must call \`super.init(...arguments);\` when overriding \`init\` on a framework object. Please update ${this} to call \`super.init(...arguments);\` from \`init\`.`,
```
rwjblue: `super.init` is only valid when extending via native classes, and does not work properly for classes created via `Ember.Object.extend()`

rwjblue:
> Does this._super(...arguments) also work in native classes?

No, when using native classes you do need to use `super.init` and when using `.extend()` you have to use `this._super`.

ReasonableDeveloper:
As @pixelhandler pointed out in the issue, if we knew [here](https://github.com/emberjs/ember.js/blob/7efa8356f2fd91fd6eabd262794fcfddca32aa03/packages/%40ember/-internals/runtime/lib/system/object.js#L64) whether the object is a native class or using `extend()`, I could be throwing the corresponding assertion error. Does anyone know how to achieve this?

Otherwise this PR should only get merged if `extend()` is completely deprecated and only native classes are supported.

pzuraq:
We could probably add language to this that says you should use `this._super()` in classic classes, and `super.init()` in native classes. I think a longer error message is fine if it's clearer

snewcomer:
I like @pzuraq's recommendation.

What happens if an addon provides a service as a native class and is injected into a mixin that does `.create()`?

pzuraq:
That shouldn't be an issue, this is only about construction of a class. Injections inject an instance, which handles its own construction, so it's not a cross-cutting concern.

snewcomer:
Thank you!  That makes sense.  This one has been tricking me for a bit.  I don't have an answer yet for issues like this:

https://github.com/DockYard/ember-in-viewport/issues/204
```You must call `this._super(...arguments);` when overriding `init` on a framework object. Please update <my-app@service:-observer-admin::ember1455> to call `this._super(...arguments);` from init```

`observer-admin` is a native class in a `-private` folder
https://github.com/DockYard/ember-in-viewport/blob/master/addon/-private/observer-admin.js#L13

That might stamp out an instance here:
https://github.com/DockYard/ember-in-viewport/blob/master/addon/services/in-viewport.js#L41

Any thoughts on why it is a `service` (`<my-app@service:-observer-admin::ember1455>`) and why something like an Ember Object that includes the mixin that injects this service might get this error?  (a while back I converted the `in-viewport` service to a native class b/c of this error as well.

pzuraq:
So that shouldn't be possible, those styles of names only get assigned to a class if they go through the resolver, which this class is not. It looks like, however, that `ObserverAdmin` _used_ to be a service: https://github.com/DockYard/ember-in-viewport/blob/v3.1.4/app/services/-observer-admin.js. The user also mentioned that this is breaking addons that use this addon. 

My guess is that they have an older version of the addon in their dependency tree, and possibly that it is clobbering the newer versions. You should tell them to install [ember-cli-dependency-lint](https://github.com/salsify/ember-cli-dependency-lint) to see if they can catch these kinds of bugs in the future.

rwjblue:
Updated the assertion messaging and rebased. I think this should be good to go once CI is happy...

rwjblue (approved):

rwjblue:
CI completed ([see here](https://travis-ci.org/github/emberjs/ember.js/builds/690030625), not sure why GH UI status isn't updating.

PR #17508, opened 2019-01-23
fix: docs should not use undocumented property

jelhan:
`url` property of EmberRouter is not documented. I'm not even sure if it was ever documented. Checked API docs for some version (2.18, 2.12, 2.4, 1.13) and didn't found it in there docs either.

Refactored the example given in `didTransition()` method therefore. Since user tracking is a good example, I kept it but refactored it to use public router service.

Also removed the usage of `.get()`, which is not needed anymore for recent versions of Ember.


rwjblue (approved):

rwjblue:
Thank you!

PR #17510, opened 2019-01-24
Ensure Ember.$.ajax works properly.

rwjblue:
The recent `Ember.$` deprecation accidentally introduced a regression when using `Ember.$` as _if_ it were the global `jQuery` object. The specific scenario that this cropped up in was an application calling `Ember.$.ajax`.

### DIFF
--- a/packages/ember/index.js
+++ b/packages/ember/index.js
@@ -561,19 +561,26 @@ Ember.VERSION = VERSION;
 
 // ****@ember/-internals/views****
 if (!views.jQueryDisabled) {
-  Ember.$ = function() {
-    deprecate(
-      "Using Ember.$() has been deprecated, use `import jQuery from 'jquery';` instead",
-      false,
-      {
-        id: 'ember-views.curly-components.jquery-element',
-        until: '4.0.0',
-        url: 'https://emberjs.com/deprecations/v3.x#toc_jquery-apis',
-      }
-    );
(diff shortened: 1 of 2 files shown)


packages/ember/tests/reexports_test.js
```diff
@@ -63,6 +63,12 @@ if (!jQueryDisabled) {
           assert.equal(body, document.body, 'Ember.$ exports working jQuery instance');
```
simonihmig: This test is failing now. I think because the deprecation is now already triggered when accessing `Ember.$`, and not when calling it. Moving this [line](https://github.com/emberjs/ember.js/pull/17510/files#diff-fdceed71491044ac9b9895ffdc26c52bR60) into the `expectDeprecation()` callback should fix this!

packages/ember/index.js
```diff
@@ -561,19 +561,26 @@ Ember.VERSION = VERSION;
...
+          until: '4.0.0',
+          url: 'https://emberjs.com/deprecations/v3.x#toc_jquery-apis',
+        }
+      );
+
+      return views.jQuery;
+    },
+
+    configurable: true,
+    enumerable: false,
```
simonihmig: Shouldn't this be true, to not change the previous enumerability behaviour?

simonihmig (commented):
Uh, yes, didn't think of this use case. Glad you catched it so early!

PR #17515, opened 2019-01-25
[DOC beta] Update (at)ember/string documentation for imported utils instead of String.prototype extensions

locks:


packages/@ember/string/index.ts
```diff
@@ -399,7 +411,7 @@ if (ENV.EXTEND_PROTOTYPES.String) {
       configurable: true,
       enumerable: false,
       writeable: true,
-      value: function() {
+      value: function () {
```
rwjblue: Spacing seems odd here, and is failing the build

locks: *shakes fist at VS Code*

rwjblue:
Thanks @locks!

Issue #17529, opened 2019-01-29
each-in breaks on keys that contain a period in 3.8.0-beta.1


jasonmit:
`{{each-in}}` will return `undefined` for the value whose key contains a period.

This bug does not exist in 3.7.2 and first appears in 3.8.0-beta.1 and is still present in 3.8.0-beta.3.

Fail test (drop in `packages/@ember/-internals/glimmer/tests/integration/syntax/each-in-test.js`)

```js
[`@only each-in supports keys with a period in them`]() {
  this.render(
    strip`
    <ul>
      {{#each-in categories as |_ item|}}
        <li>{{item.name}}</li>
      {{/each-in}}
    </ul>
  `,
    {
      categories: {
        // uncomment and run.  notice `items` is undefined
        'hello.world': { name: 'foo' },
        // uncomment and run.  notice it works as expected
        // hello world: { name: 'foo' },
      },
    }
  );

  // Empty
  this.assertHTML(strip`
    <ul>
      <li>foo</li>
    </ul>
  `);
}
```

locks:
@jasonmit could you send a PR with the failing test?

rwjblue:
We should definitely root cause this, but I'm not sure that this will be considered a bug in the end. Specifically, `get(obj, 'hello.world')` will look for `obj.hello.world` not `obj['hello.world']`...

chancancode:
Yeah pretty sure that’s the reason, each-in uses get so this is sort of a duplicate of the get with dot path ticket. Mayyyybe in this specific case we don’t need to use get in here after landing ES5 getters because this probably doesn’t work with proxies anyway?

chancancode:
Specifically https://github.com/emberjs/ember.js/blob/master/packages/%40ember/-internals/glimmer/lib/utils/iterable.ts#L128 

I agree I’m not sure if this is considered a bug and whether fixing this now will cause problems elsewhere or down the Ross

rwjblue:
Ya, and I actually think we do different things based on the heuristics on the object being iterated here:

https://github.com/emberjs/ember.js/blob/7df81ec411b2da73086f63bf6a26b113dfa5a1a2/packages/%40ember/-internals/glimmer/lib/utils/iterable.ts#L290-L298

Some of those iterators use `get` (`ObjectIterator.fromIndexable`) and some don't (`ObjectIterator.fromForEachable`, `MapLikeNativeIterator.from`)

chancancode:
[...]

PR #17166, opened 2018-10-31
get() / set() perf optimizations

kanongil:
I noticed that the ember-performance Chrome scores of `get()` and `set()` dropped dramatically when `ember-metal-es5-getters` was enabled, so I had a look at it.

I have managed to mostly bring it back for `get()`, and improve it for `set()` using the included patches. Both are roughly 2x the canary performance. Additionally, it seems to have improved the `run()` benchmark as well.

All the patches should work individually, in case you only think some of them are suitable.

**Chrome**
```
User Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/71.0.3578.20 Safari/537.36

.------------------------------------------------------------------------.
|                   Ember Performance Suite - Results                    |
|------------------------------------------------------------------------|
|         Name         |       Speed        | Error  | Samples |  Mean   |
|----------------------|--------------------|--------|---------|---------|
|  -- Ember 3.0.0 --   |                    |        |         |         |
| Ember.get            | 1,626,872.70 / sec | ∓1.08% | 64      | 0.00 ms |
| Ember.run            | 670,554.44 / sec   | ∓2.42% | 65      | 0.00 ms |
| Ember.set            | 606,311.51 / sec   | ∓1.12% | 64      | 0.00 ms |
|  -- Ember 3.4.5 --   |                    |        |         |         |
| Ember.get            | 770,484.17 / sec   | ∓1.65% | 61      | 0.00 ms |
| Ember.run            | 710,395.41 / sec   | ∓6.42% | 58      | 0.00 ms |
| Ember.set            | 408,123.26 / sec   | ∓0.47% | 65      | 0.00 ms |
|  -- Ember canary --  |                    |        |         |         |
| Ember.get            | 858,315.30 / sec   | ∓0.96% | 61      | 0.00 ms |
| Ember.run            | 739,482.59 / sec   | ∓0.34% | 66      | 0.00 ms |
| Ember.set            | 453,404.65 / sec   | ∓0.61% | 65      | 0.00 ms |
|  -- Ember dev --     |                    |        |         |         |
| Ember.get            | 1,487,276.86 / sec | ∓1.04% | 63      | 0.00 ms |
| Ember.run            | 818,811.25 / sec   | ∓0.53% | 66      | 0.00 ms |
| Ember.set            | 996,207.06 / sec   | ∓1.33% | 61      | 0.00 ms |
'------------------------------------------------------------------------'
```

[...]

PR #17535, opened 2019-01-30
chore: fail test for each-in not accessing props properly whose keys contain a period

jasonmit:
Related to #17529


bekzod:
probably caused by #17166 I think offending line https://github.com/emberjs/ember.js/blob/master/packages/@ember/-internals/metal/lib/property_get.ts#L99-L101

chancancode:
Seems like the problem is we used to check `obj[key]` before we even check if key is a path? To fix #17529, we can probably call getPath directly with a single-element array. But is this behavior change in #17166 really okay?

rwjblue:
We should add some tests if that is actually a thing we intend to support, but to my knowledge we have never "really" supported `get(obj, 'foo.bar.baz')` finding `obj['foo.bar.baz']`, have we?

jasonmit:
@rwjblue I haven't found anywhere documented detailing this getter path behavior either, so it may have been a bug this app was depending on for a few years.

If we're not supporting it, I'd still like to send over tests so the behavior stays consistent into the future.  In our case it was a bit trolly since the keys were not symbolic of a path they were ISO datetime strings.

rwjblue:
Ya, I totally agree that we should nail down the specific desired behavior with tests. If for nothing more than to ensure we don’t regress and change behavior _again_.

kategengler:
I'm going to go ahead and close this -- this was unspecified behavior and not an area we want to touch going forward.

PR #17539, opened 2019-01-31
removed duplicate line

IzzatN:
this line repeats on line 283

### DIFF
--- a/packages/ember/index.js
+++ b/packages/ember/index.js
@@ -203,8 +203,6 @@ Object.defineProperty(Ember, 'NAME_KEY', {
   },
 });
 
-Ember._Cache = utils.Cache;
-
 // ****@ember/-internals/container****
 Ember.Container = Container;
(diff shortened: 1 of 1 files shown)


packages/ember/index.js
```diff
@@ -203,8 +203,6 @@ Object.defineProperty(Ember, 'NAME_KEY', {
   },
 });
 
-Ember._Cache = utils.Cache;
```
stefanpenner: can you share where this is duplicated?

stefanpenner (changes requested):

packages/ember/index.js
IzzatN: sorry, on the line 283
https://github.com/emberjs/ember.js/blob/master/packages/ember/index.js#L283

locks (approved):

stefanpenner (approved):

locks:
Thank you!

PR #17552, opened 2019-02-02
[BUGFIX lts] support numbers in component names for Angle Brackets

### LINKED ISSUE #17549: Some components with numbers in the name don't work with Angle Brackets
As reported by @gmartell, if you have a number in a component name, it fails to be found if used with Angle Bracket invocation. It seems like a component `number1-problem` is ok as `<Number1-Problem />` but fails as `<Number1Problem />`.

[Minimal reproduction](https://ember-twiddle.com/eb642a7dac966923b64421cf9d4a333d)

jenweber:
Previously, a component named `c3-plot` would not work when invoked as `<C3Plot />`.
Now it will resolve correctly.

Closes #17549

Behavior:

```
'F3Bar' => 'f3-bar'
'Foo3Bar' => 'foo3-bar'
'Foo3barBaz' => 'foo3bar-baz'
'FooB3ar' => 'foo-b3ar'
```

Notably it does not support `Foo2bar => foo-2bar`, but it did not support it before this change either.

It's a bit edge-casey but I think given the prevalent use of C3 and D3 in JS, it makes sense to lean towards this patch's behavior.

### DIFF
--- a/packages/ember-template-compiler/lib/system/dasherize-component-name.ts
+++ b/packages/ember-template-compiler/lib/system/dasherize-component-name.ts
@@ -5,7 +5,7 @@ import { Cache } from '@ember/-internals/utils';
   `Ember.String.dasherize` would resolve it to `xfoo`..
 */
 const SIMPLE_DASHERIZE_REGEXP = /[A-Z]/g;
-const ALPHA = /[A-Za-z]/;
+const ALPHA = /[A-Za-z0-9]/;
 export default new Cache<string, string>(1000, key =>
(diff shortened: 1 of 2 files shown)


rwjblue (commented):
Changes look good, can you change the commit prefix to `[BUGFIX lts]`?

jenweber:
Done! Thanks for the tips on which files to look at to fix this.

rwjblue:
Hehe, well it _was_ my fault to begin with so....

rwjblue:
Thank you for fixing it!

PR #17553, opened 2019-02-02
[DOC release] Update RouterService docs to fix code examples for routeWillChange / routeDidChange

scottyd980:
The code examples were using routeWillUpdate / routeDidUpdate rather than the correct method name (routeWillChange / routeDidChange). I have updated them to match to reduce confusion.


rwjblue (approved):

rwjblue:
Thank you!!

PR #17557, opened 2019-02-04
Lazy router setup in non-application tests

### LINKED ISSUE #16831: RouterService.urlFor causes error in non-application tests.
### Steps to reproduce
1. Use `urlFor()` in a component and output the value in the template
1. Attempt to render the component in an integration/rendering test

```
TypeError: Cannot read property 'generate' of undefined
    at Class.generate (https://cdn.jsdelivr.net/npm/ember-source@3.2.2/dist/ember.debug.js:33524:59)
    at Class.urlFor (https://cdn.jsdelivr.net/npm/ember-source@3.2.2/dist/ember.debug.js:31371:48)
```

### Expected behavior
The test doesn't crash.

Since `autoboot=false` is in use it makes sense that there's no Router for RouterService to work with. However, I would expect it to return empty data rather than crash. Similar to how `{{link-to}}` behaves in the same scenario.

### Twiddle
If you click Run Tests you'll see the exception.
https://ember-twiddle.com/bc59e11d525fc3548519f66250173a32?openFiles=my-component.component.js%2C

xg-wang:
During non `setupApplicationTests` type tests, the Router being injected into `service:router` and `service:routing` does not start routing, in which it initializes its `_routerMicrolib`. Calling public API on `service:router` will throw in those tests.

This commit will trigger startRouting on router, in non application tests the router service should _just work™_.

Link tested on https://github.com/xg-wang/__router-test/commit/df287142eecce502ebab557386a22c5794edb878. It fails on [travis-ci](https://travis-ci.com/xg-wang/__router-test).

Fixes https://github.com/emberjs/ember.js/issues/16831

### DIFF
--- a/packages/@ember/-internals/routing/lib/services/router.ts
+++ b/packages/@ember/-internals/routing/lib/services/router.ts
@@ -57,7 +57,7 @@ export default class RouterService extends Service {
       this._router.startRouting();
     }
     return this._router;
-  };
+  }
 
   /**
      Transition the application into another route. The route may
@@ -154,12 +154,7 @@ export default class RouterService extends Service {
     let hasQueryParams = Object.keys(queryParams).length > 0;
 
     if (hasQueryParams) {
-      this.router._prepareQueryParams(
-        routeName,
-        models,
-        queryParams,
-        true /* fromRouterService */
-      );
+      this.router._prepareQueryParams(routeName, models, queryParams, true /* fromRouterService */);
       return shallowEqual(queryParams, routerMicrolib.state!.queryParams);
     }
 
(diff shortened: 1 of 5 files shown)


rwjblue:
I’m not sure this is the right direction. IMHO, tests that need the router (e.g. those testing generated href’s or things using the router service) should call `this.owner.setupRouter()` explicitly in the test.

xg-wang:
Can you explain the reason behind this? I think if using router service in a component is not discouraged the test experience should also be out-of-box.

rwjblue:
Hmm, @xg-wang I think I've made a mistake here, sorry for the double speak (between https://github.com/emberjs/ember.js/issues/16831#issuecomment-445598310 and my comments here). What do you think about reviving these changes?

nlfurniss:
Is it better to handle this here or to call `this.owner`.setupRouter()` in `setupRenderingTest`?

rwjblue:
> Is it better to handle this here or to call this.owner.setupRouter()insetupRenderingTest`?

IMHO, most rendering tests aren't going to need the router. So eagerly setting it up in all tests is a bit wasteful.

PR #17568, opened 2019-02-06
add ability to pass `log` helper as action

bekzod:
so following is possible: 

```javascript
  <button {{log 'rainbow'}}>Click!</button> // logs `rainbow` and `MouseEvent` on click
  <button onclick={{log 'rainbow'}}>Click!</button> // logs `rainbow` and `MouseEvent` on click
  <button onclick={{action (log 'rainbow') 'sun'}}>Click!</button>  // logs `rainbow`,  'sun' and `MouseEvent` on click
```

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/helpers/-log-action.ts
+++ b/packages/@ember/-internals/glimmer/lib/helpers/-log-action.ts
@@ -1,0 +1,19 @@
+import { Arguments, VM } from '@glimmer/runtime';
+import { ACTION, UnboundReference } from '../utils/references';
+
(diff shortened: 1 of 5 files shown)


rwjblue:
I think we would want to leverage [emberjs/rfcs#432](https://github.com/emberjs/rfcs/blob/master/text/0432-contextual-helpers.md) for this kind of thing instead.

If we went this route, we'd need an RFC to change the behavior. I'm going to close for now (in favor of either implementing emberjs/rfcs#432 or a new RFC).

PR #17570, opened 2019-02-06
[BUGFIX] passing ObjectProxy with a property size to `isEmpty` would throw assertion

### LINKED ISSUE #17032: `isEmpty` call on ProxyObject w/ `size` property throws '.get()' proxy assertion (3.1+)
Having an ObjectProxy (or any object using unknownProperty) with a property named 'size' of type String `isEmpty(myChangeset)` results in the `Assertion Failed: You attempted to access the 'size' property (of changeset:[object Object]). Since Ember 3.1 , this is usually fine as you no longer need to use '.get()'...` error.

This is caused at: https://github.com/emberjs/ember.js/blob/f73d8440d19cf86a10c61ddb89d45881acfcf974/packages/%40ember/-internals/metal/lib/is_empty.ts#L44 and only occurs in Ember 3.1+ (which is when the assertion was added) which also means that the assertion will trigger for any object with a "size" key that refers to a proxy.

Minimal reproduction: https://ember-twiddle.com/8e63470aa2e8ab8b07735334b169c96d?openFiles=controllers.application.js%2C

AFAIK `.size` is just used for Maps/Sets and the like? Might an extra check be necessary?

Somewhat similar to: https://github.com/emberjs/ember.js/issues/16259

cc @snewcomer

nickschot:
fixes #17032

### DIFF
--- a/packages/@ember/-internals/metal/lib/is_empty.ts
+++ b/packages/@ember/-internals/metal/lib/is_empty.ts
@@ -41,7 +41,7 @@ export default function isEmpty(obj: any): boolean {
     return none;
   }
 
-  if (typeof obj.size === 'number') {
+  if (typeof obj.unknownProperty !== 'function' && typeof obj.size === 'number') {
     return !obj.size;
   }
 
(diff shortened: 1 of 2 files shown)


locks:
Hi! Is this the same as https://github.com/emberjs/ember.js/pull/18882 ?

nickschot:
> Hi! Is this the same as #18882 ?

It seems similar. Not sure if `typeof obj.unknownProperty !== 'function'` results in the same as `typeof obj` (which is what happens in the other PR) though?

nickschot:
@locks rebased!

PR #17548, opened 2019-02-02
[FEAT] Implements Decorators

pzuraq:
This is a continuation of #17226, thanks a ton to @NullVoxPopuli for getting this started!

This PR refactors the Descriptor system in Ember metal, mostly replacing
it with a Decorator system instead. `defineProperty` has been updated to
apply decorators following the currently implemented spec. This means
that the value returned by `computed` goes through the exact same flow
when applied as a decorator using decorator syntax, and applied using
classic syntax.

I was considering pulling out the ComputedDecorator stuff into a separate
package, it's relatively isolated from the rest of `metal` at this point
so it would make some sense. Thoughts?

TODO:
- [x] Benchmark
- [x] Disable native application except behind a canary feature flag

PR #17572, opened 2019-02-06
[FEAT] Updates @tracked to match the Tracked Properties RFC

pzuraq:
Updates `@tracked` to be a classic decorator as well as a standard
native decorator, removes the ability to decorator native getters, and
adds more tests for various situations and updates the existing tests.

One thing to note is that users can still provide getters/setters to
classic classes via tracked:

```js
import EmberObject from '@ember/object';

const Person = EmberObject.extend({
  firstName: tracked({ value: 'Tom' }),
  lastName: tracked({ value: 'Dale' }),

  fullName: tracked({
    get() {
      return `${this.firstName} ${this.lastName}`;
    },
  }),
});
```

The reasoning behind this is there is no way to do this currently, and
it allows us to match behavior between native/classic exactly.
Alternatively we could expose the native descriptor decorator that is
internal only right now, but unfortunately `computed` will not be able
to fill this role since it _requires_ use of `Ember.set`, where native
getters/setters do not.

TODO:
- [x] Get #17548 merged
- [x] Put native invocation behind the decorators feature flag


rwjblue:
@pzuraq - Can you fix the package.json conflict here?

rwjblue (commented):
I think we may be missing some tests around recomputing during rendering.

@pzuraq - Would you mind adding some tests that use getters that then consume tracked properties in a template?

packages/@ember/-internals/metal/index.ts
```diff
@@ -64,3 +64,16 @@ export {
   isSearchDisabled as isNamespaceSearchDisabled,
   setSearchDisabled as setNamespaceSearchDisabled,
 } from './lib/namespace_search';
+
+import { DEBUG } from '@glimmer/env';
+import { setComputedDecorator } from './lib/decorator';
+import { tracked } from './lib/tracked';
+
+// We have to set this here because there is a cycle of dependencies in tracked
```
rwjblue: We should work to remove this cycle before landing

rwjblue (approved):

Issue #17569, opened 2019-02-06
Deprecate ApplicationController#currentPath and ApplicationController#currentRouteName


locks:
Tracks https://github.com/emberjs/rfcs/master/text/0421-deprecate-application-controller-props.md.

## Tasks
- [ ] Implement deprecation in source code https://github.com/emberjs/ember.js/pull/17573
- [ ] Write deprecation guide https://github.com/ember-learn/deprecation-app/pull/276

rwjblue:
We should migrate this to the corresponding emberjs/rfc-tracking issue now that they are created...

locks:
See https://github.com/emberjs/rfc-tracking/issues/5.

PR #17193, opened 2018-11-08
[Deprecation] Artisanally crafted deprecation for @rwjblue

chadhietala:
Implements https://github.com/emberjs/rfcs/pull/392/.

_100% Shade Grown Deprecations_

PR #17573, opened 2019-02-06
[DEPRECATE] - RFC 421 - deprecate app ctrl router props

mistahenry:
Implements [RFC #421](https://github.com/emberjs/rfcs/pull/421/files).
Part of https://github.com/emberjs/ember.js/issues/17569.

Please let me know if `notifyPropertyChange` was an okay way to solve this.

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/router.ts
+++ b/packages/@ember/-internals/routing/lib/system/router.ts
@@ -1,9 +1,14 @@
-import { computed, defineProperty, get, set } from '@ember/-internals/metal';
+import { computed, get, notifyPropertyChange, set } from '@ember/-internals/metal';
 import { getOwner, Owner } from '@ember/-internals/owner';
 import { A as emberA, Evented, Object as EmberObject, typeOf } from '@ember/-internals/runtime';
 import { EMBER_ROUTING_ROUTER_SERVICE } from '@ember/canary-features';
 import { assert, deprecate, info } from '@ember/debug';
-import { HANDLER_INFOS, ROUTER_EVENTS, TRANSITION_STATE } from '@ember/deprecated-features';
+import {
+  APP_CTRL_ROUTER_PROPS,
+  HANDLER_INFOS,
+  ROUTER_EVENTS,
+  TRANSITION_STATE,
+} from '@ember/deprecated-features';
 import EmberError from '@ember/error';
 import { assign } from '@ember/polyfills';
 import { cancel, once, run, scheduleOnce } from '@ember/runloop';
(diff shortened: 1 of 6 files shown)


locks:
Can you add the following URL to the deprecation, `https://emberjs.com/deprecations/v3.x#toc_application-controller-router-properties`?
It should become valid when https://github.com/ember-learn/deprecation-app/pull/276 gets merged.

mistahenry:
@locks Updated the deprecation with the url/ matching deprecation id

packages/@ember/-internals/routing/lib/system/router.ts
```diff
@@ -1637,16 +1637,44 @@ function updatePaths(router: EmberRouter) {
...
+    });
   }
-
-  set(appController, 'currentPath', path);
+  notifyPropertyChange(appController, 'currentPath');
 
   if (!('currentRouteName' in appController)) {
-    defineProperty(appController, 'currentRouteName');
+    Object.defineProperty(appController, 'currentRouteName', {
+      enumerable: false,
```
chadhietala: This shouldn't be needed.

packages/@ember/-internals/routing/lib/system/router.ts
```diff
@@ -1637,16 +1637,44 @@ function updatePaths(router: EmberRouter) {
   }
 
   if (!('currentPath' in appController)) {
-    defineProperty(appController, 'currentPath');
+    Object.defineProperty(appController, 'currentPath', {
+      enumerable: false,
```
chadhietala: This shouldn't be needed.

packages/@ember/-internals/routing/lib/system/router.ts
```diff
@@ -1637,16 +1637,44 @@ function updatePaths(router: EmberRouter) {
   }
 
   if (!('currentPath' in appController)) {
```
chadhietala: We need to wrap all the code that is set for deprecation in a deprecation flag. This allows you to build a version of Ember without any of the deprecated code. If you need an example of what you need to do I would look at. https://github.com/emberjs/ember.js/pull/17193

chadhietala (changes requested):
This looks great! Thanks for doing this, just a couple small things we need to address before merging.

packages/@ember/-internals/routing/lib/system/router.ts
mistahenry: I chose `export const APP_CTRL_ROUTER_PROPS = !!'3.9.0';` but that was only based on [the deprecation link](https://github.com/ember-learn/deprecation-app/pull/276/files) and my assumption that `3.9.0` refers to the version in which this is deprecated and not when the deprecation is removed.

chadhietala: It's actually when it needs to be removed so it would need to be `!!'4.0.0'`.

mistahenry: alright, I've updated it

chadhietala:
Thanks so much @mistahenry!

PR #17585, opened 2019-02-07
[BUGFIX canary] Un-deprecate Route#(will|did)Transition events

chadhietala:
After re-reading the RFC it appears that the `Route` level events were not supposed to be deprecated.

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/route.ts
+++ b/packages/@ember/-internals/routing/lib/system/route.ts
@@ -19,7 +19,7 @@ import {
   EMBER_ROUTING_BUILD_ROUTEINFO_METADATA,
   EMBER_ROUTING_ROUTER_SERVICE,
 } from '@ember/canary-features';
-import { assert, deprecate, info, isTesting } from '@ember/debug';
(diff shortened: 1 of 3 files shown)


kategengler:
I made a mistake here and thought that the hooks were deprecated (that's what I was trying to clarify in Discord). 

@ef4 said in Discord he does not have a strong opinion about the events on the Route being deprecated. 

I think what _is_ needed is an adjustment to the deprecation guide (I'll make a PR).

rwjblue (commented):
TBH, I think deprecating the events on both the route and router is good/fine...

jenweber:
Followup issue for changing the language in the deprecation guides: https://github.com/ember-learn/deprecation-app/issues/284

rwjblue:
> I think what _is_ needed is an adjustment to the deprecation guide (I'll make a PR).

We discussed this in todays Ember Core team meeting, and decided this is the right path forward. The implemented deprecations are good, the deprecation guide needs to be updated to be clearer.

@kategengler - Would you mind "owning" that update?

In the meantime, I'm going to close this PR...

Issue #17393, opened 2018-12-19
Computed property recomputes two times for a single set


boris-petrov:
[Here is a reproduction](https://github.com/boris-petrov/ember-recompute-bug).

`ember server`

Then open http://localhost:4200/ and type in the console `x.set('a', Ember.A([{ a: 2, b: 4 }]))` - you will see `computing y` twice in the console. If you remove any of the two dependencies of `y` (i.e. make it either `x.@each.a` or `x.@each.b`) - then it computes only once.

I think this is a bug.

jacobq:
I looked at your repro just for fun, and that behavior does surprise me, but at the same time it seems to be caused by returning the computed property from the observer. I don't understand why one would ever want to return a value from an observer. Could you explain this more?

I also tried tweaking your example a bit in https://github.com/boris-petrov/ember-recompute-bug/pull/1 in case it helps someone identify why this happens.

boris-petrov:
@jacobq - thanks for the answer. Actually the return from the observer is just a leftover as I translated this code from CoffeeScript. Getting the value is what is important, returning it does nothing. Our use case is of course much more involved, I don't want to even begin explaining it, but it boils down to essentially this code.

Thanks for the expansion of the example! We've workarounded (not sure if there is such a word) the issue for now but it would be nice if there is a fix at some point. :)

boris-petrov:
Yes, the observer does access the computed property - in a transitive way, but in the end it does. And yes, as you've noticed - if the observer is removed, then the bug goes away.

I'm not sure whether this needs any kind of "synchronization". I wouldn't think it does - it's simple getters and setters (and there is no asynchrony anywhere). And generally I'm against putting randomly stuff like `scheduleOnce` especially when I don't understand why it needs to be there.

jacobq:
Out of curiosity, does the recomputation of the computed property cause a problem (it should should be a pure function of the values it depends on so should produce the same result if re-evaluated) or are you just concerned about the performance?

boris-petrov:
It is a "pure" function, yes, however it causes a search to the BE to be repeated with the same query - we have our own computed properties that do searches and they do a new query when the computed property is recomputed. In this case it recomputes twice with the same value and causes two searches for the same thing. Not the end of the world, just a performance bug. :)

zeroe454:
`index.html > <!-- -->`

pixelhandler:
[...]

PR #17595, opened 2019-02-08
[FAILING TEST] computed property recomputes multiple times #17393

bekzod:
failing test for #17393


boris-petrov:
@bekzod - this test should now be passing on 3.25.1 (at least the original issue I reported has been fixed so I guess this test also will be) so perhaps you can rebase and then this can be merged.

PR #17597, opened 2019-02-09
classerize the blueprints

NullVoxPopuli:


rwjblue (commented):
We need to hold off a bit on this. Specifically, we should only be doing native class versions of the blueprints if the host app/addon has opted in to being “octane”.

I had worked through some of the details with @ppcano in discord, and I think they had started some work on it. C/D?

ppcano:
@NullVoxPopuli Closed by #17621

PR #17598, opened 2019-02-10
Refactor MU component blueprint

ppcano:
The component blueprint reuses the same `files` folder for Classic, Pod and MU layout. 

Now, the component blueprint works on the same way than the other blueprints. 

Small clean-up before starting the blueprint support for Native JS classes.


rwjblue (approved):

rwjblue:
👍 - thank you!

PR #17617, opened 2019-02-14
[FEAT][BETA] Tunnels the Owner through CoreObject#create

pzuraq:
This PR plucks the owner from `create` properties and passes it to the
root constructor, where it gets assigned. This allows classes to access
explicit dependency injections via `@service` or `@controller` while the
constructor is running, which means that we don't have to use `init` for
most classes. Routes, Controllers, and Services should generally be ok.
Classic components will not, however, nor will custom user defined
classes.


rwjblue:
I'm in favor...

rwjblue (approved):

Issue #17611, opened 2019-02-12
Error when using same template as partial and in outlet


viccatoff:
I have two different pages that renders same sub-template but in two different way:
1) one page render it using `partial` helper
2) second page render it using `outlet` (from sub-page)

Basically first page renders this block on page directly, and second page renders it inside a tab content.
When you navigate from first page to second - there is an error:

```
runtime.js:246 Uncaught TypeError: Cannot read property 'someVar' of null
    at Object.evaluate (runtime.js:246)
    at AppendOpcodes.evaluate (runtime.js:46)
    at LowLevelVM.evaluateSyscall (runtime.js:2889)
    at LowLevelVM.evaluateInner (runtime.js:2861)
    at LowLevelVM.evaluateOuter (runtime.js:2853)
    at VM.next (runtime.js:4781)
    at VM.execute (runtime.js:4766)
    at TryOpcode.handleException (runtime.js:3746)
    at UpdatingVMFrame.handleException (runtime.js:3916)
    at UpdatingVM._throw [as throw] (runtime.js:3651)
```

According to stack trace the error inside this method:

```
APPEND_OPCODES.add(96 /* ResolveMaybeLocal */, function (vm, _ref5) {
        var _name = _ref5.op1;

        var name = vm.constants.getString(_name);
        var locals = vm.scope().getPartialMap();
        var ref = locals[name];
        if (ref === undefined) {
            ref = vm.getSelf().get(name);
        }
        vm.stack.push(ref);
    });
```

`locals` variable is null.

There is no such error when you navigate from second page to first one.

I've created project in twiddle that reproduce this bug: [project](https://ember-twiddle.com/80041d0c35dbd0821df3d4f0adfe658a)

Not sure that it's a bug (maybe `locals` should be an empty object `{}` at least or add additional condition before `locals[name]`). I've found some work around for me that fixes this bug by using two additional sub-template - one for first page and one for second. And these two templates renders my general sub-template.

Alonski:
I am also running into this issue exactly as @viccatoff described :)
The partial is rendered initially and then when routing to the page with the same template as the partial the error is thrown.

Alonski:
Reproduction for my error that is related:
https://ember-twiddle.com/0571fff7b5464e2bac66cb8d874c19d5?openFiles=templates.application.hbs%2C

GavinJoyce:
[...]

PR #17633, opened 2019-02-18
failing test for GH#17611

GavinJoyce:
I paired on this with @Alonski. Here's a screencast of our pairing session: https://www.youtube.com/watch?v=R3nFU9WfNnw

This adds a failing test for https://github.com/emberjs/ember.js/issues/17611

screen shot 2019-02-18 at 11 41 40

While debugging, we made the following change in [`Op.ResolveMaybeLocal`](https://github.com/glimmerjs/glimmer-vm/blob/a15a7c86ae50a882f3f395a394cdaff6211ade30/packages/%40glimmer/runtime/lib/compiled/opcodes/expressions.ts#L64-L74) to make the test pass. Not knowing enough about glimmer-vm scope internals, neither of us are confident in this as an actual fix though:

screen shot 2019-02-18 at 11 46 39


locks:
Thanks for the work cooming up with the test! Partials have been deprecated in the mean time and from @chancancode's description in the linked issue, the "correct" fix might be a bit involved so I'm closing this PR.

PR #17634, opened 2019-02-18
[FEAT] Adds autotracked to the the Glimmer references system

pzuraq:
This PR adds autotracking around property references in Glimmer-Ember,
and adds high level acceptance tests that verify that it's working for
both components and helpers. This does _not_ test the behavior of
autotracking around modifiers, which is still ill-defined.

Note that while eventually simplifications may be possible, even
removing nested references altogether, currently it is not due to
two-way binding. Components that use two-way binding need to be able to
update a nested property reference, which means we need to store that
information somewhere accessible. It may still be possible to do this
only in cases where it is necessary (e.g. binding to a component's
arguments) but for the time being, this a smaller and iterative
approach.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/utils/references.ts
+++ b/packages/@ember/-internals/glimmer/lib/utils/references.ts
@@ -1,5 +1,15 @@
-import { didRender, get, set, tagFor, tagForProperty, watchKey } from '@ember/-internals/metal';
+import {
+  didRender,
+  get,
+  getCurrentTracker,
+  set,
+  setCurrentTracker,
+  tagFor,
+  tagForProperty,
+  watchKey,
+} from '@ember/-internals/metal';
 import { isProxy, symbol } from '@ember/-internals/utils';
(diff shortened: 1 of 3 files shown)


tomdale (approved):

packages/@ember/-internals/glimmer/lib/utils/references.ts
```diff
@@ -207,7 +220,25 @@ export class RootPropertyReference extends PropertyReference
...
+
+    if (EMBER_METAL_TRACKED_PROPERTIES) {
+      parent = getCurrentTracker();
+      tracker = setCurrentTracker();
+    }
+
+    let ret = get(_parentValue, _propertyKey);
+
+    if (EMBER_METAL_TRACKED_PROPERTIES) {
+      setCurrentTracker(parent!);
```
rwjblue: I’m surprised that you need the `!` here, if the function allows undefined we should just fix the signature?

rwjblue (commented):
Looks good! Can you add some tests around class based helpers too?

rwjblue:
We can address my comments in a follow up PR...

PR #17636, opened 2019-02-19
 [FEAT] Adds tracking logic to `ArrayMixin#objectAt`

pzuraq:
This PR adds tracking logic to `ArrayMixin#objectAt`. This will allow implementers of the EmberArray interface to automatically have KVO compliant changes be tracked. Note that this implementation will not do much for `emberA()` since it wraps native arrays, and does _not_ replace many methods that would have to be intercepted to get the same effect (`map`, `join`, `forEach`, etc). If we want to support `emberA()`, we can:

1. Use a native Proxy in modern browser to wrap all methods
2.  Manually wrap all methods in IE11/older browsers.

I decided to go with the more minimal approach here, since we could also do this as an addon (`trackedA()`). However, it _is_ important that `ArrayMixin` has some way to entangle it's `[]` tag with the autotrack stack for Ember Data at least.

Depends on #17634

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/components/tracked-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/components/tracked-test.js
@@ -1,5 +1,5 @@
 import { EMBER_METAL_TRACKED_PROPERTIES } from '@ember/canary-features';
-import { Object as EmberObject } from '@ember/-internals/runtime';
+import { A, Object as EmberObject } from '@ember/-internals/runtime';
 import { tracked, nativeDescDecorator as descriptor } from '@ember/-internals/metal';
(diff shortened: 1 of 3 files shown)


rwjblue:
Ready for a rebase...

pzuraq:
After talking with @krisselden I don't think this is actually necessary directly in Ember. Instead, implementers of the `ArrayMixin` and array interface should `get(arr, '[]')` in any operation that should cause the array to be tracked, e.g. `objectAt`:

```js
objectAt() {
  get(this, '[]'); // push tracked prop
  // ...
}
```

This should work because the other side of the contract is to always trigger a notification on that property. For `emberA()` and native arrays, we'll still need to figure out the path forward.

PR #17637, opened 2019-02-19
Cleanup router a tiny little bit

luxzeitlos:
This is just a few types, replacing `for` loops with `for of` loops and moving some stuff from `init` to the constructor or property initializers.

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/router.ts
+++ b/packages/@ember/-internals/routing/lib/system/router.ts
@@ -248,6 +248,11 @@ interface OutletState {
   outlets: NestedOutletState;
 }
 
+interface EngineInstance extends Owner {
+  boot(): void;
+  destroy(): void;
+}
+
 export interface QueryParam {
   prop: string;
   urlKey: string;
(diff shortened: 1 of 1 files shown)


packages/@ember/-internals/routing/lib/system/router.ts
```diff
@@ -467,8 +493,8 @@ class EmberRouter extends EmberObject {
       'application',
       { path: '/', resetNamespace: true, overrideNameAssertion: true },
       function(this: EmberRouterDSL) {
-        for (let i = 0; i < dslCallbacks.length; i++) {
-          dslCallbacks[i].call(this);
+        for (let dslCallback of dslCallbacks) {
```
rwjblue: We cannot use `for...of` loops due to global `Symbol` requirement (`Symbol` is not present on IE11).

rwjblue (changes requested):

packages/@ember/-internals/routing/lib/system/router.ts
luxferresum: gnrgs I hate this world. Sorry.

packages/@ember/-internals/routing/lib/system/router.ts
```diff
@@ -248,6 +248,11 @@ interface OutletState {
   outlets: NestedOutletState;
 }
 
+interface EngineInstance extends Owner {
```
rwjblue: technically I think its the other way around

rwjblue (approved):

rwjblue:
thank you!

PR #17655, opened 2019-02-22
[DOC release] Documented tagless component behavior

dbrusilovsky:
Detailed how components with an empty string tagName property behave.

@jenweber suggested that the API docs should cover the whole supported API surface, and while tagless components may be an anti-pattern, they are part of the API surface and their functionality is explicitly tested for.


jenweber (approved):
This looks good to me!

rwjblue (commented):
Awesome, thank you!

Issue #17678, opened 2019-02-27
Unable to build ember from master 


dbrusilovsky:
I'm trying to build ember using `yarn install` and `yarn build`, and am running into the following error:
```
Build Error (broccoli-persistent-filter:Babel) in @ember/-internals/browser-environment.js
​
Cannot read property 'getHash' of null
```
A paste of my process:
https://dpaste.de/Ne8h

Node version: v10.15.0
Yarn version: v1.13.0
Operating system: macos 10.14.3

The issue was replicated on another machine by @efx on discord.

dbrusilovsky:
Error log: https://dpaste.de/j3ZU

dbrusilovsky:
Offending commit appears to be e90aee2d0

mixonic:
@rwjblue @stefanpenner I ran into this as well tonight, seems related to the bump of hash-for-dep from 1.2.3 to 1.4.7. I opened a hacky quick fix for Ember in https://github.com/emberjs/ember.js/pull/17679.

rwjblue:
Thanks for reporting! Will dig in to see what’s going on...

stefanpenner:
I have an idea what the issues And with the reproduction it should be easy to tackle.

ppcano:
I don't know if this is related but when using my local `ember-source` repo with an Octane app, an exception is thrown:

```
Directory not found: /Users/ppcano/dev/ember/ember.js/dist/legacy
```

Steps:

```
cd ember.js
yarn link
yarn start
```

```
// other terminal
ember new oct-app -b ember-octane-app-blueprint --yarn
cd oct-app
yarn link ember-source
ember s -p 4201
```

pzuraq:
@ppcano You need to run `yarn start --environment="production"`, your app is targeting legacy browsers (IE11, etc) and looking for the legacy build, which is only built in production mode in the `ember-source` repo

stefanpenner:
@ppcano that doesn't seem related

stefanpenner:
@mixonic's revert addressed the issue and https://github.com/emberjs/ember.js/pull/17694 unreverts \w fix

PR #17679, opened 2019-02-28
Revert upgrade to hash-for-dep 1.4.7

mixonic:
Fixes the broken master build as reported in https://github.com/emberjs/ember.js/issues/17678

Something from the bump of hash-for-dep from 1.2.3 -> 1.4.7 seems to have introduced the build failure. By pinning these resolutions to conservative but valid versions builds of master are repaired.

Please track root cause analysis and other items on https://github.com/emberjs/ember.js/issues/17678. Keep this ticket limited to the contents of the PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
     "babel-plugin-module-resolver": "^3.2.0",
     "babel-template": "^6.26.0",
     "backburner.js": "^2.5.0",
-    "broccoli-babel-transpiler": "^7.2.0",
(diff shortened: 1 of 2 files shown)


rwjblue (approved):

stefanpenner:
I did a root cause, basically old hash-for-dep would not if it was given a funky input. Turns out that funky input not error'ing was depended on, so I'll restore that functionality to h-f-d and consider restoring the error, although a good error, in a major version.

PR #17682, opened 2019-02-28
[BUGFIX canary] Ensure @tracked setter triggers a revalidation.

rwjblue:
Prior to this change, setting a tracked property when there was on ambient runloop would never schedule a revalidation and would therefore never rerender.

This change ensures that _any_ set of a `@Tracked` property will enqueue a revalidation regardless if there is a current runloop or not.


tomdale:
Looks good to me. I'm a little surprised that the lower-level abstractions (maybe `dirty`?) aren't responsible for scheduling the runloop, but maybe there are performance concerns I'm not aware of. Either way, this is a good fix for a bad bug.

tomdale (approved):

pzuraq (approved):

PR #17666, opened 2019-02-25
WIP - Angle bracket blueprint

ppcano:
#17625

@sukima will continue the work.

Issue #17625, opened 2019-02-17
Angle Bracket invocation on blueprints


ppcano:
Related to https://github.com/emberjs/ember.js/issues/17524

Blueprints are a good mechanism to learn the framework API. If we are updating the docs to the Angle Bracket syntax; I think we should update also the code generated by the blueprints.

ppcano:
@MelSumner This is likely an issue that could be included on the [Octane Tracking issue](https://github.com/emberjs/ember.js/issues/17234)

rwjblue:
Yep, totally!

ppcano:
@rwjblue 

The component blueprint can create a nested component or receive the `path` option but as far as I know the angle bracket syntax does not support nested components.

I thought that the blueprint could fallback to the curly bracket syntax when angle bracket is not possible. Does this make sense?

sukima:
Has there been any progress on this or is it an Octane only thing. Can I start looking into this now? (I have no job at the moment) or should I move on to another issue?

sukima:
Regarding blueprints. If I understand the issue a blueprint can generate a component in any directory structure needed (including nested components). However I don't recall the blueprints generating invocations of such components. What I mean is that it can place the files in nested directories but then it is up to the user to invoke it. In this case the distinction to support both angle brackets vs curly brackets is some what a moot point.

It would seem to me the scope of the blueprint to recognize a nested component might be nothing more then to print out a warning message on the console that the user must use curly brackets to invoke the component.

rwjblue:
> However I don't recall the blueprints generating invocations of such components.

`ember generate component foo-bar` generates both a component and a component test, that test includes invocations of the component. I believe that is specifically what this issue is referring to.

PR #17685, opened 2019-03-01
Angle bracket blueprint

sukima:
This PR supersedes #17666 and attempts to address issue #17625

When generating the blueprints for a component the component-test blueprint will attempt to generate a test that uses Angle Bracket Syntax unless it sees that the component is nested in a path requiring the generated test to fallback and use Curly Bracket Syntax.


rwjblue (approved):

rwjblue:
Thank you!

Issue #17556, opened 2019-02-03
Generated helper does not work out of box under Module Unification feature


axyxnz:
When you run `ember generate helper hello-helper`, it generates two files (`hello-helper.js` and `hello-helper-test.js` under `src/ui/components` folder.

The content of `hello-heper.js` is following:
```
import { helper } from '@ember/component/helper';

export function helloHelper(params/*, hash*/) {
  return params;
}

export default helper(helloHelper);
```

Above generated code does not work. During runtime, Ember outputs following error in console:
`Uncaught Error: Compile Error: hello-helper is not a helper at opcode-compiler.js:294`

**To make it work, you need to change how helper exports.** If you modify `hello-helper.js` with following code, it works.
```
import { helper as buildHelper } from '@ember/component/helper';

export function helloHelper(params/*, hash*/) {
  return params;
}

export const helper = buildHelper(helloHelper);
```

Thanks @NullVoxPopuli for helping me out on Discord to solve this problem. I can update blueprint for helper and submit a PR to fix this if no one is addressing this issue under Module Unification meta issue.

ppcano:
Thanks for reporting the issue.

The helper blueprint creates the same content for both layouts (Classic and MU), it just places the files in different locations. As far as I know, this is correct and `helpers` should have the same API either for the Classic or MU layout. , or am I missing something?

@NullVoxPopuli could you provide more info about why this change was needed?

NullVoxPopuli:
I think it's just the convention going forward. 
I wouldn't be able to point to an RFC or anything, but as far as I've known, this has always been required of the octane layout.

ppcano:
Thanks, I found https://github.com/emberjs/ember.js/issues/16361 describing the public API. 

The guides should also be updated, I have added a comment at https://github.com/ember-learn/guides-source/issues/460#issuecomment-464419841

--------------------------------------

@axyxnz 

> I can update blueprint for helper and submit a PR to fix this if no one is addressing this issue under Module Unification meta issue.

A PR would be very much welcomed. Please, let me know if you are gonna work on it.

A possible implementation could be:

- define a folder `blueprints/helper/mu-files` with the new content.
- overwrite `filesPath` on `blueprint/helper/index.js`

```
   filesPath = function() {
    let rootPath = isModuleUnificationProject(this.project) ? 'mu-files' : 'files';
    return path.join(this.path, rootPath);
  };
```
- Tests run like

```
mocha node-test/blueprint/helper-test.js
```

[...]

PR #17689, opened 2019-03-01
Fix helper blueprint for module unification

sukima:
Prior to this change, the classic blueprint used a different construction for helper files. Since module unification this has changed.

This commit updates the blueprints to generate the new construction code for helpers when module unification is in use and still offers the classic construction code when not.

Fixes issue #17556


ppcano (approved):

ppcano:
I have tested it in a local octane app and works well. Thank you for working on it.

rwjblue:
Thank you!

PR #17690, opened 2019-03-01
[FEAT] Updates decorator implementation to stage 1

pzuraq:
This moves us back to stage 1 decorators for our internal decorator implementation. It also reverts to using a Symbol instead of a WeakMap for `@tracked` values. This is overall a better for perf, assuming we can assign the value of the symbol in the constructor via an initializer. Unfortunately stage 1 decorators don't allow this, but fortunately it should be very easy to add a babel transform specifically for `@tracked` that does, so we can get the ideal performance.

### DIFF
--- a/broccoli/transforms/transform-babel-plugins.js
+++ b/broccoli/transforms/transform-babel-plugins.js
@@ -4,8 +4,8 @@ module.exports = function(tree) {
   let options = {
     sourceMaps: true,
     plugins: [
-      ['@babel/plugin-proposal-decorators', { decoratorsBeforeExport: true, legacy: false }],
-      ['@babel/plugin-proposal-class-properties'],
+      ['@babel/plugin-proposal-decorators', { legacy: true }],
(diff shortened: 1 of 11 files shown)


rwjblue (approved):

GavinJoyce:
I'm seeing `Error: Assertion Failed: The options object passed to tracked() may only contain a 'value' or 'initializer' property, not both. Received: [kind,key,placement,descriptor,initializer]` for simple `@tracked count = 1;` in a recent app created with the `ember-octane-app-blueprint`.

The error occurs with the latest `ember-source` sha, not with a sha from yesterday

pzuraq:
The octane app blueprint needs to be updated to use the stage 1 decorator transforms. We can install `@ember-decorators/babel-transforms@v2` for now

GavinJoyce:
Thanks, here's a PR to downgrade `babel-transforms` in the octane blueprints:

https://github.com/ember-cli/ember-octane-blueprint/pull/58

PR #17691, opened 2019-03-02
[BUGFIX] Ensure tagForProperty works on class constructors

### LINKED ISSUE #17651: `EMBER_METAL_TRACKED_PROPERTIES` fails with Ember Data
@pzuraq 

Testing the `octane` blueprints with Ember Data; `update(propertyTag, tag)` triggers an exception.

screen shot 2019-02-22 at 4 50 18 am

Using a simple model like:

```
export default Model.extend({
  name: attr('string')
});
```

pzuraq:
`tagForProperty` was returning the constant tag for functions, which
meant it would return the constant tag for properties on class
constructors. If a computed property was defined on the class
constructor it would fail because of this.

Fixes #17651

### DIFF
--- a/packages/@ember/-internals/metal/lib/tags.ts
+++ b/packages/@ember/-internals/metal/lib/tags.ts
@@ -22,7 +22,8 @@ function makeTag(): TagWrapper<DirtyableTag> {
 }
 
 export function tagForProperty(object: any, propertyKey: string | symbol, _meta?: Meta): Tag {
-  if (typeof object !== 'object' || object === null) {
+  let objectType = typeof object;
+  if (objectType !== 'function' && (objectType !== 'object' || object === null)) {
     return CONSTANT_TAG;
   }
   let meta = _meta === undefined ? metaFor(object) : _meta;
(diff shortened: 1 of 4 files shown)


packages/@ember/-internals/metal/lib/tags.ts
```diff
@@ -22,7 +22,7 @@ function makeTag(): TagWrapper<DirtyableTag> {
 }
 
 export function tagForProperty(object: any, propertyKey: string | symbol, _meta?: Meta): Tag {
-  if (typeof object !== 'object' || object === null) {
+  if (typeof object !== 'function' && (typeof object !== 'object' || object === null)) {
```
rwjblue:
```suggestion
  let objectType = typeof object;
  if (objectType !== 'function' && (objectType !== 'object' || object === null)) {
```

packages/@ember/-internals/metal/tests/computed_test.js
```diff
@@ -82,6 +82,19 @@ moduleFor(
       assert.equal(count, 1, 'should have invoked computed property');
     }
 
+    ['@test computed property can be defined and accessed on a class constructor'](assert) {
```
rwjblue: Can we have a native class version of this test too?

```js
class Foo {}
Foo.bar = 123;
Ember.defineProperty(Foo, 'foo', computed(function() { return this.bar; }));
```

rwjblue (changes requested):

rwjblue (approved):

PR #17694, opened 2019-03-03
Fix hash for dep issue

stefanpenner:


stefanpenner:
@mixonic this reverts your revert https://github.com/emberjs/ember.js/pull/17679 but with the problem solved. (local prod builds work fine)

mixonic:
@stefanpenner the build failed here, and it seems like the cause was the `yarn.lock` re-roll bumping a TypeScript version. [Here is the failed `yarn run tsc --noEmit` run on Travis](https://travis-ci.org/emberjs/ember.js/jobs/500958610).

I force pushed a new commit the avoids a re-roll for `yarn.lock`. `hash-for-dep` 1.4.7 is still used in `yarn.lock` (in addition to 1.5.0 of course), but not in any way that seems to cause a CI failure. And TypeScript is not bumped.

I copied your original work to [this branch on my fork](https://github.com/mixonic/ember.js/tree/mixonic/fix-hash-for-dep-issue-original) in case you want to find it.

Seems like we should look at getting the re-rolled yarn lock green sometime soon.

mixonic (approved):

stefanpenner:
@mixonic thank you.

PR #17658, opened 2019-02-24
Add missing blueprints for MU `dummy` app.

ppcano:
Closes #17656

Fix the following commands for MU `dummy` apps:

- [x] `ember g helper hola --dummy`
- [x] `ember g mixin hola --dummy`
- [x] `ember g service hola --dummy`
- [x] `ember g util hola --dummy`

PR #17716, opened 2019-03-07
Fix blueprint helper test

ppcano:
It fixes the test when generating a helper in a MU `dummy` app.

It continues the work of #17658 that requires this change when #17689 was merged.


rwjblue:
Thank you @ppcano & @turbo87

PR #17719, opened 2019-03-08
Support for nested components in angle bracket invocation syntax

tomdale:
This is an implementation of the [Nested Invocations in Angle Bracket Syntax RFC](https://github.com/emberjs/rfcs/pull/457). It enables invoking components nested inside directories with angle bracket invocation syntax.

For example, if a component exists in `app/templates/components/foo-bar/baz-bang`, you can invoke it with curlies like this:

```hbs
{{foo-bar/baz-bang}}
```

This PR adds support for invoking that same component with angle brackets like this:

```hbs
<FooBar::BazBang />
```

The feature is added behind the `EMBER_GLIMMER_ANGLE_BRACKET_NESTED_LOOKUP` feature flag.


rwjblue:
Has conflicts with the PR just merged (which removes some of the other previously enabled features).

tomdale:
@rwjblue Sorry, will fix.

rwjblue:
Also, I'd strongly prefer coming to a conclusion in the RFC before implementing (unless we are also accepting a feature enabling the same feature with `/`).

rwjblue (approved):

PR #17731, opened 2019-03-13
[BUGFIX] Adds a warning to {{action}} for undecorated methods

pzuraq:
This PR adds a warning to the `{{action}}` modifier and the `{{action}}`
helper which tells the user they should add the `@action` decorator to
any method which is passed to it. We're only warning for the time being
since this _is_ a supported codepath, but has been actively discouraged
in the past.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/helpers/action.ts
+++ b/packages/@ember/-internals/glimmer/lib/helpers/action.ts
@@ -2,14 +2,15 @@
 @module ember
 */
 import { get } from '@ember/-internals/metal';
-import { assert } from '@ember/debug';
+import { EMBER_NATIVE_DECORATOR_SUPPORT } from '@ember/canary-features';
+import { assert, warn } from '@ember/debug';
 import { flaggedInstrument } from '@ember/instrumentation';
 import { join } from '@ember/runloop';
 import { DEBUG } from '@glimmer/env';
 import { isConst, VersionedPathReference } from '@glimmer/reference';
 import { Arguments, VM } from '@glimmer/runtime';
 import { Opaque } from '@glimmer/util';
-import { ACTION, INVOKE, UnboundReference } from '../utils/references';
+import { ACTION, ACTION_METHOD, INVOKE, UnboundReference } from '../utils/references';
 
 /**
   The `{{action}}` helper provides a way to pass triggers for behavior (usually
@@ -389,6 +390,16 @@ function makeClosureAction(
 
       assert(`An action named '${action}' was not found in ${target}`, fn);
     } else if (typeofAction === 'function') {
+      if (EMBER_NATIVE_DECORATOR_SUPPORT) {
+        warn(
+          `You passed a method, ${debugKey}, to the {{action}} helper which was not decorated with the '@action' decorator. All actions should be decorated with the '@action' decorator.`,
+          target[debugKey] !== action || action[ACTION_METHOD] === true,
+          {
+            id: 'action-without-decorator',
+          }
+        );
+      }
+
       self = context;
       fn = action;
     } else {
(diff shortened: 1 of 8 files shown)


packages/@ember/-internals/glimmer/lib/helpers/action.ts
```diff
@@ -389,6 +389,14 @@ function makeClosureAction(
 
       assert(`An action named '${action}' was not found in ${target}`, fn);
     } else if (typeofAction === 'function') {
+      warn(
```
rwjblue: Shouldnt this be flagged?

packages/@ember/-internals/glimmer/lib/modifiers/action.ts
```diff
@@ -212,13 +212,21 @@ export default class ActionModifierManager implements ModifierManager<ActionStat
...
-            '{{action}} helper which did not resolve to an action name (a ' +
+            '{{action}} modifier which did not resolve to an action name (a ' +
             'string). Perhaps you meant to use a quoted actionName? (e.g. ' +
             '{{action "' +
             actionLabel +
             '"}}).',
           typeof actionName === 'string' || typeof actionName === 'function'
         );
+
+        warn(
```
rwjblue: Needs feature flagging

rwjblue (changes requested):

chancancode:
Would this cause problems for passing actions into classic components?

{{foo-bar baz=(action “foo”)}}

Then inside Foo:

<button {{action this.bar}} />

Wouldn’t this trigger the warning?

pzuraq:
I believe you are correct, we should be able to check against the `ACTION` symbol on the method as well as `ACTION_METHOD`, and that should cover all cases I think. I'll add some tests that assert we do not warn if either version of the `action` helper is used.

chancancode:
Would it be better to check that it’s a method (on prototype only) since that’s what you are after? Sure I’ll miss `field: () => ...` but I’m not sure if there’s much to do about that, since it’s not really distinguishable for referencing a passed in thing in classic components

pzuraq:
Wouldn't that also miss plain, unbound methods on native classes? Is there a way we can check whether or not a method has been bound already?

chancancode:
The things that I care about:

1. Using the "actions hash" + `{{action "..."}}` (the string version) is equivalent to using `@action` in classic components, and should not trigger a warning.
2. Ideally, accessing a passed in thing with `{{action this.foo}}` (or even `{{action foo}}`? is that a thing?), which is how you access passed-in arguments, shouldn't trigger a warning. It should be the caller's job to do it.

I don't know how feasible the second thing is.

In the example I gave in the first comment, I think if `{{action "..."}}` "stamps" the result, it would actually work fine. (We should probably only do the stamping in dev mode?)

> Wouldn't that also miss plain, unbound methods on native classes?

I _think_ it would work? You are basically checking `Object.getPrototypeOf(this)[fnName] === fn`, and I think that would catch the case you want, though it will probably miss some other cases. I am open to other implementations, as long as it doesn't affect the two cases I mentioned above.

pzuraq:
Alright, in the latest iteration I first check if the value exists on the class instance (using the path, which is available in debug mode). That way, we warn IFF:

1. The value being passed to the modifier or helper is a function
2. The value being passed exists directly on the context it would be bound to, so it was passed like `{{action this.myAction}}`
3. The value has not been marked as an action method by the `@action` decorator.

This will also allow `ember-bind-helper` and any other helpers that return functions to work, since a yielded val

chancancode:
Do you mind adding tests for the cases you explicitly _did not_ want to trigger the warning, so future refactors and bug fixes won't break it?

pzuraq:
Yes, I added those tests a couple revisions ago: 
https://github.com/emberjs/ember.js/pull/17731/files#diff-75100a70de5647bea8404cc50e0f05b9R265

rwjblue (approved):

chancancode:
@pzuraq I mean the cases that we discussed above, such as:

```js
this.render(`{{foo-bar onClick=someFn}}`)
```

paired with

```hbs
<button {{action this.someFn}}>...</button>
```

should not warn

PR #17733, opened 2019-03-13
[BUGFIX beta] assert `<Input>` and `<Textarea>`

chancancode:
`input` and `textarea` are intended to be reserved component names, but the assertion was added to the wrong place. The end result is that only `(component "input")` and `(component "textarea")` were disallowed, but not when using the angle bracket syntax `<Input />` and `<Textarea />`.

This compliments the existing tests in https://github.com/emberjs/ember.js/blob/master/packages/@ember/-internals/glimmer/tests/integration/components/contextual-components-test.js#L1220-L1230


chancancode:
I would have suggested `[BUGFIX release]`, but we know from https://github.com/emberjs/rfcs/pull/459#issuecomment-470443892 that this would break at least one app, so `[BUGFIX beta]` seems more conservative, but I would like some advise on this.

(cc @gossi: either way, my suggestion would be to rename the component in your app asap)

rwjblue (approved):

gossi:
I'm not a blocker on this one. It's in my experimental octane/MU project. Assertion makes more sense to catch both invocations 👍

PR #17745, opened 2019-03-15
[FEATURE ember-glimmer-angle-bracket-nested-lookup] enable 👌

chancancode:
We discussed it in the core team meeting today and decided to "GO" the feature. However, due to EmberConf travels, we didn't have everyone in the meeting so I am waiting a day or so to see if there are any objections from other team members.


rwjblue (approved):
🤢

locks (approved):

jenweber (approved):
Looks good to me! This is exactly what I meant by showing both angle bracket and curly syntax.

PR #17749, opened 2019-03-16
A number of improvements to component docs

wycats:
- Removed legacy langauge about "view"s
- Removed legacy language about "context"s and generally
  avoid unnecessarily referring to the controller as an
  important concept involved in {{yield}}
- Added headers
- Clean up language around the dash rule (to be fully removed
  in a near-term commit)
- Clean up language about the `layout` property that was too
  general
- Remove reference to jQuery.Event
- Other assorted aesthetic cleanups

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/component.ts
+++ b/packages/@ember/-internals/glimmer/lib/component.ts
@@ -29,15 +29,14 @@ export const BOUNDS = symbol('BOUNDS');
 */
 
 /**
-  A `Component` is a view that is completely
-  isolated. Properties accessed in its templates go
-  to the view object and actions are targeted at
-  the view object. There is no access to the
-  surrounding context or outer controller; all
-  contextual information must be passed in.
+  A component is an isolated piece of UI, represented by a template and an
+  optional class. When a component has a class, its template's `this` value
+  is an instance of the component class.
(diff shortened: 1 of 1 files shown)


packages/@ember/-internals/glimmer/lib/component.ts
```diff
@@ -29,15 +29,14 @@ export const BOUNDS = symbol('BOUNDS');
...
-  isolated. Properties accessed in its templates go
-  to the view object and actions are targeted at
-  the view object. There is no access to the
-  surrounding context or outer controller; all
-  contextual information must be passed in.
-
-  The easiest way to create a `Component` is via
-  a template. If you name a template
+  A component is an isolated piece of UI, represented by a template and an
+  optional class. When a component has a class, its template's `this` value
```
jenweber: I find the "When a component..." sentence confusing, though accurate. It's ok as-is (shouldn't block a merge)

wycats: In general, I had trouble documenting "components" in `Ember.Component` and think we need a better general strategy going forward. This was the best thing I could come up with for now, but agree we should focus on how to document component APIs in the API docs ASAP.

PR #17767, opened 2019-03-18
Fixes 595 in ember-learn/ember-api-docs - missing has-block API documentation

habdelra:
This fixes issue: https://github.com/ember-learn/ember-api-docs/issues/595 

I think there are a bunch of helpers that are more foundational to glimmer, like `{{yield}}` and `{{with}}` where the documentation was just put in the `packages/@ember/-internals/glimmer/index.ts`, despite the real functionality living deeper below. This seems to fall into that bucket, which probably explains why the documentation is missing from the API docs. The clearest place where I could find where the `{{has-block}}` helper is handled is: 

https://github.com/emberjs/ember.js/blob/master/packages/%40ember/-internals/glimmer/lib/component-managers/curly.ts#L239-L242

But really, I think there are a lot of places that actually contribute to this functionality. Probably documenting `{{has-block}}` in the `packages/@ember/-internals/glimmer/index.ts` file is the most straightforward place to put it.

### DIFF
--- a/packages/@ember/-internals/glimmer/index.ts
+++ b/packages/@ember/-internals/glimmer/index.ts
@@ -184,7 +184,9 @@
 
   This component is invoked with a block:
   ```handlebars
-  {{#my-component}}Hi Jenn!{{/my-component}}
+  {{#my-component}}
+    Hi Jenn!
+  {{/my-component}}
   ```
 
   This component is invoked without a block:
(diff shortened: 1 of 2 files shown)


habdelra:
This also now addresses this issue: https://github.com/ember-learn/ember-api-docs/issues/594 where we are now adding documentation for the `{{has-block-params}}` helper.

rwjblue (commented):
Looking really good! Could you also add examples using angle bracket invocation?

```hbs
<MyComponent /> {{! without block }}
<MyComponent></MyComponent> {{! with block }}
```

Also, `has-block` accepts a single argument: the name of the block. The name (at the moment) is either "main" or "inverse" (though only curly components support inverse). I think that argument needs to be added to the docs for both `has-block` and `has-block-params`.

rwjblue (approved):
Thank you!!!

PR #17769, opened 2019-03-18
Adds test to confirm that GH#17018 is not an issue

shawnren:
There was a [report](https://github.com/emberjs/ember.js/issues/17018#event-1875388794) that devs could not pass a record to {{link-to}} using a hash helper, because:

> the hash helper returns EmptyObject and extractQueryParams tries to call .hasOwnProperty() on it, resulting in an error

However, no one has been able to reproduce the issue, and I've created a test that replicates the exact scenario in the original report and passes. This commit adds that test to verify that the issue reported no longer exists.


chancancode:
Thank you!

PR #17773, opened 2019-03-19
Revert "[BUGFIX] Adds a warning to {{action}} for undecorated methods"

chancancode:
Reverts emberjs/ember.js#17731

We definitely want to land this soon, but we have discovered that this ended up being a bit too broad. We intended for this to help guide users to add the `@action` decorator when trying Octane with native classes.

However, it ended up trigger in a lot of existing code (not using native classes, so the suggestion of adding decorators doesn't make sense), especially in addon (which is outside of the user's control).

We will try to land this again more narrowly soon, after fixing these issues (and adding tests around those cases).

### DIFF
--- a/packages/@ember/-internals/glimmer/index.ts
+++ b/packages/@ember/-internals/glimmer/index.ts
@@ -300,7 +300,7 @@ export { default as AbstractComponentManager } from './lib/component-managers/ab
 // rather than the problem was solved
 // DebugStack should just test the assert message
 // it supports for example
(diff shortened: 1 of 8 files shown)


pzuraq (approved):

chancancode:
Some known cases came from: https://github.com/miguelcobain/ember-paper/blob/dc125fb26161dc869434a09be5d1a69f9de41390/addon/templates/components/paper-autocomplete-trigger.hbs#L12

PR #17788, opened 2019-03-20
Fix native DOM events in Glimmer Angle Brackets

wagenet:


chancancode:
Thank you! Can you also add it to the angle bracket file?

wagenet:
@chancancode done. Looks like we need to have the Octane flags on too to trigger it.

.travis.yml
```diff
@@ -93,6 +93,10 @@ jobs:
       script:
         - yarn ember test -c testem.travis-browsers.js
 
+    - name: Octane
+      env:
+      - OVERRIDE_FEATURES=EMBER_METAL_TRACKED_PROPERTIES,EMBER_GLIMMER_ANGLE_BRACKET_BUILT_INS,EMBER_GLIMMER_ANGLE_BRACKET_NESTED_LOOKUP,EMBER_NATIVE_DECORATOR_SUPPORT
+
```
chancancode: This shouldn't be necessary: we already have a build that enables _all_ the optional features, it's just in the second "stage"

wagenet: I realized this later.

wagenet:
Disabling the tagless action test and adding `@input={{@input}}` to the template resolves this. Of course, disabling the check isn't a long term solution. Also, are there any other events that have this same issue?

packages/@ember/-internals/glimmer/lib/components/input.ts
```diff
@@ -148,6 +148,11 @@ if (EMBER_GLIMMER_ANGLE_BRACKET_BUILT_INS) {
   Input = Component.extend({
     tagName: '',
 
+    init() {
+      this[DISABLE_TAGLESS_EVENT_CHECK] = true;
```
chancancode: wrap this in `if (DEBUG) { ... }`

rwjblue (approved):

wagenet:
@chancancode looks like some of these events don't work in some of the browser tests. How should we handle this?

chancancode:
Maybe create a regular component, go through the loop and identify what is
supposed to work, and use that list for the test

wagenet:
@chancancode I think the issue is that different browsers support different events.

chancancode:
Yes. I mean do that inside your test. Step 1: register a normal component and render it, do the loop you have now and collect the list of supported events without any assertions. Step 2: loop through that smaller list and assert on the input component t

PR #17804, opened 2019-03-26
[WIP] deprecate view registry

chancancode:
This fixes an issue where event dispatching stops working if `id` is passed with splattributes, because `elementId` will diverge from the actual `id`, causing the element to not be found during event dispatching.

~~This patch does not fix the `elementId` divergent problem, but changes the framework code to not rely on `elementId` and use a weak map instead. I have ideas for fixing the `elementId` issue, and making it optional, but I'll probably submit it separately.~~ (It's fixed.)

It's probably best to review this commit-by-commit.

This also removes the view registry private API. The theory is that there are very few addons that uses it, and we can transition them to using the new `Ember.ViewUtils.getElementView` instead.

### DIFF
--- a/packages/@ember/-internals/views/lib/system/utils.js
+++ b/packages/@ember/-internals/views/lib/system/utils.js
@@ -1,6 +1,7 @@
 import { getOwner } from '@ember/-internals/owner';
 /* globals Element */
 import { guidFor } from '@ember/-internals/utils';
+import { assert } from '@glimmer/util';
 
 /**
 @module ember
@@ -26,20 +27,26 @@ export function constructStyleDeprecationMessage(affectedStyle) {
   );
 }
 
+const VIEW_REGISTRY = new Map();
+
+export function registerView(view) {
+  VIEW_REGISTRY.set(getViewId(view), view);
+}
+
+export function unregisterView(view) {
+  VIEW_REGISTRY.delete(getViewId(view));
+}
+
 /**
   @private
   @method getRootViews
   @param {Object} owner
 */
 export function getRootViews(owner) {
-  let registry = owner.lookup('-view-registry:main');
-
   let rootViews = [];
 
-  Object.keys(registry).forEach(id => {
-    let view = registry[id];
-
-    if (view.parentView === null) {
+  VIEW_REGISTRY.forEach(view => {
+    if (getOwner(view) === owner && view.parentView === null) {
       rootViews.push(view);
     }
   });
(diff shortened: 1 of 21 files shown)


packages/@ember/-internals/views/lib/system/utils.js
```diff
@@ -60,23 +60,30 @@ export function getViewId(view) {
...
 }
 
 export function setViewElement(view, element) {
-  return (view[VIEW_ELEMENT] = element);
+  VIEW_ELEMENT.set(view, element);
+  ELEMENT_VIEW.set(element, view);
+}
+
+// Not needed for GC, but for correctness
+export function clearViewElement(view) {
```
rwjblue: Shouldn't this also remove from `ELEMENT_VIEW`? Seems _slightly_ odd to have this mismatch vs setViewElement

rwjblue (commented):
Generally 👍, only had one small question

packages/@ember/-internals/views/lib/system/utils.js
```diff
@@ -60,23 +60,38 @@ export function getViewId(view) {
...
+export function setElementView(element, view) {
+  ELEMENT_VIEW.set(element, view);
 }
 
 export function setViewElement(view, element) {
-  return (view[VIEW_ELEMENT] = element);
+  VIEW_ELEMENT.set(view, element);
+}
+
+// These are not needed for GC, but for correctness
```
wycats: Can you write a little more here?

wycats (changes requested):

wycats (dismissed):

rwjblue (dismissed):
LGTM

mixonic:
Should `-view-registry` go through an intimate API deprecation re: [usages in the wild](https://www.emberobserver.com/code-search?codeQuery=-view-registry&sort=usages&sortAscending=false)?

rwjblue:
> Should -view-registry go through an intimate API deprecation re: usages in the wild?

I did review usages on EmberObserver.com and I do not think the limited usage merits intimate API deprecation. The usages in addons that are maintained (releases in the last year or so) are mostly official packages (`@ember/test-helpers`, `ember-resolver` in the MU flagged code, etc), the usages in maintained non-official addons seem to mostly be in custom testing utilities and actually do have a nice API to migrate to.

rwjblue:
@chancancode - Should we close, or revive?

chancancode:
huh yes, I thought we did the elementId fix already.. did we not?

chancancode:
Ah we did in fact land #17818 at least. We can probably just remove this now since the inspector switched to using the debug render tree. (Though I just added a test in #19382 that uses the view registry because I was lazy. 😉)

locks:
@chancancode is this still relevant?

kategengler:
Is this still relevant? It is labeled `wip`.

kategengler:
Closing as stale

PR #17816, opened 2019-03-28
[DOCS release] update paramsFor documentation

xiwcx:
It may be that I'm misunderstanding the meaning of "ancestor" here, but in my experience I've found that I can use `paramsFor` with both parent and child routes. This could benefit from some clarity and I am happy to work with someone on a more accurate description.

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/route.ts
+++ b/packages/@ember/-internals/routing/lib/system/route.ts
@@ -179,6 +179,10 @@ class Route extends EmberObject implements IRoute {
   /**
     Returns a hash containing the parameters of an ancestor route.
 
+    You may notice that `this.paramsFor` sometimes works when referring to a
+    child route, but this behavior should not be relied upon as only ancestor
+    routes are certain to be loaded in time.
+
     Example
 
     ```app/router.js


thomasmatecki (approved):
Looks good!

locks:
While it may technically, sometimes, work, accessing params for a route that is not a parent route is a very big potential trap due to timings and the like.
The parent route is the only one that you know for sure is properly initialized by the time you use `this.paramsFor`.

xiwcx:
Hi @locks, thanks much for the explanation. I added a version of your explanation (hope that's OK) in lieu of my change.

locks (approved):

rwjblue:
This PR needs to target the `master` branch (you can update that in the GH UI) instead of the `release` branch. I think updating the target branch and rebasing should drop out all of the extraneous changes that exist in this PR at the moment.

xiwcx:
@rwjblue thanks for the heads up. looks like the force push may have changed the underlying history? not 100% sure on how to fix this. let me know if it's preferable that i figure it out or just open a new PR that is branched correctly.

rwjblue:
OK, I think I fixed the commit history issue here. Mind double checking?

For posterity, I did:

```
git rebase -i origin/master
```

And deleted all but the last commit (the one you added).

locks:
Thanks @rwjblue and @xiwcx 🙏

xiwcx:
thanks in turn @rwjblue and @locks this has definitely been a positive first experience!

PR #17819, opened 2019-03-29
Add basic test infra for simulated @glimmer/component base class.

rwjblue:


chancancode:
Can't we just add @glimmer/component as a dev dependency and use the real thing?

rwjblue:
> Can't we just add @glimmer/component as a dev dependency and use the real thing?

Not easily. We don't build like a normal ember-cli addon, so to add it here would require us to also replicate a portion of ember-cli's build pipeline to emit it.

tldr; I definitely agree with your point, but I'd like to hold that until Ember is itself a better behaving addon (hopefully in the next couple of months).

pzuraq (approved):

PR #17824, opened 2019-03-30
Upgrade Glimmer and TypeScript

chancancode:
@glimmer/* 0.37.1 -> 0.38.1
typescript 3.0.1 -> 3.2.4

This also fixed a bug in the instrumentation code that swallowed all errors in code blocks that are instrumented 😱


packages/@ember/instrumentation/index.ts
```diff
@@ -123,83 +131,107 @@ const time = ((): (() => number) => {
...
+  binding: Binding
+): Result {
   try {
-    result = callback.call(binding);
+    return callback.call(binding);
   } catch (e) {
-    (payload as PayloadWithException).exception = e;
-    result = payload;
+    (payload as { exception: any }).exception = e;
+    throw e;
```
chancancode: o_0

PR #17825, opened 2019-03-30
Fix some types, remove `any`s from `references.ts`

chancancode:


packages/@ember/-internals/glimmer/lib/utils/references.ts
```diff
@@ -393,8 +371,8 @@ export class SimpleHelperReference extends CachedReference {
       let namedValue = named.value();
 
       if (DEBUG) {
-        maybeFreeze(positionalValue);
-        maybeFreeze(namedValue);
+        debugFreeze(positionalValue);
+        debugFreeze(namedValue);
       }
```
chancancode: Do we still need to wrap these in DEBUG?

PR #17833, opened 2019-04-01
[BUGFIX] Reverts the naming of setClassicDecorator externally

pzuraq:
The rename was intended to be internal only, the external API is
already in use so we should keep it as is. Name is unfortunate, but
it kind of is either way.


josemarluedke (approved):
🎉

buschtoens (approved):

buschtoens:
[ember-concurrency](https://github.com/machty/ember-concurrency/blob/f1649147285cdcc382289dc0cffe945f242b731c/addon/index.js#L22) is currently broken in Ember Octane because of this.

Thanks for fixing it so quickly! :)

buschtoens:
I currently do this to fix it:

```js
// @see https://github.com/emberjs/ember.js/pull/17833
if (!Ember._setComputedDecorator) {
  Ember._setComputedDecorator = Ember._setClassicDecorator;
}
```

```js
// ember-cli-build.js
app.import('vendor/fix-decorator.js');
```

igorT:
This broke Ember Data as well

rwjblue:
Just pushed an additional commit that deprecates instead.

PR #17834, opened 2019-04-02
 [BUGFIX beta] Prevents autotracking ArrayProxy creation

pzuraq:
This PR prevents an issue with immediate invalidation within the new
autotracking transaction. ArrayProxy currently reads from one of its
own properties, `hasArrayObservers`, and then immediately updates it
with `notifyPropertyChange`.

We avoid this by using a native descriptor instead of a computed, and
not using `get()` to access it. This way, nothing is tracked during
creation (even if it is mutated).

The original PR attempted to add array observers lazily, when the array was first accessed. However, tests in addons (such as Ember Data) were failing with this strategy. One issue was that adding observers to `length` wouldn't work, since there was no way to add the array observer when that occured. This strategy keeps the current semantics the same, and prevents the backtracking issue.

Also updated the PR to deprecate in `unknownProperty` if this occurs again, since it could happen in existing code today.

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/components/tracked-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/components/tracked-test.js
@@ -1,9 +1,10 @@
-import { Object as EmberObject, A } from '@ember/-internals/runtime';
+import { Object as EmberObject, A, ArrayProxy, PromiseProxyMixin } from '@ember/-internals/runtime';
 import {
   EMBER_CUSTOM_COMPONENT_ARG_PROXY,
   EMBER_METAL_TRACKED_PROPERTIES,
 } from '@ember/canary-features';
 import { computed, tracked, nativeDescDecorator as descriptor } from '@ember/-internals/metal';
+import { Promise } from 'rsvp';
 import { moduleFor, RenderingTestCase, strip, runTask } from 'internal-test-helpers';
 import GlimmerishComponent from '../../utils/glimmerish-component';
 import { Component } from '../../utils/helpers';
@@ -45,6 +46,31 @@ if (EMBER_METAL_TRACKED_PROPERTIES) {
 
         runTask(() => this.context.set('first', 'max'));
         this.assertText('max jackson | max jackson');
+      }
+
+      '@test creating an array proxy inside a tracking context does not trigger backtracking assertion'() {
+        let PromiseArray = ArrayProxy.extend(PromiseProxyMixin);
+
+        class LoaderComponent extends GlimmerishComponent {
+          get data() {
+            if (!this._data) {
+              this._data = PromiseArray.create({
+                promise: Promise.resolve([1, 2, 3]),
+              });
+            }
+
+            return this._data;
+          }
+        }
+
+        this.registerComponent('loader', {
+          ComponentClass: LoaderComponent,
+          template: '{{#each this.data as |item|}}{{item}}{{/each}}',
+        });
+
+        this.render('<Loader/>');
+
+        this.assertText('123');
       }
 
       '@test tracked properties that are uninitialized do not throw an error'() {
--- a/packages/@ember/-internals/runtime/lib/mixins/array.js
+++ b/packages/@ember/-internals/runtime/lib/mixins/array.js
@@ -19,6 +19,7 @@ import {
   removeArrayObserver,
   arrayContentWillChange,
   arrayContentDidChange,
+  nativeDescDecorator as descriptor,
 } from '@ember/-internals/metal';
 import { assert } from '@ember/debug';
 import Enumerable from './enumerable';
@@ -564,8 +565,12 @@ const ArrayMixin = Mixin.create(Enumerable, {
     @property {Boolean} hasArrayObservers
     @public
   */
-  hasArrayObservers: nonEnumerableComputed(function() {
-    return hasListeners(this, '@array:change') || hasListeners(this, '@array:before');
+  hasArrayObservers: descriptor({
+    configurable: true,
+    enumerable: false,
+    get() {
+      hasListeners(this, '@array:change') || hasListeners(this, '@array:before');
+    },
   }),
 
   /**
@@ -695,7 +700,7 @@ const ArrayMixin = Mixin.create(Enumerable, {
     foods.forEach((food) => food.eaten = true);
 
     let output = '';
-    foods.forEach((item, index, array) => 
+    foods.forEach((item, index, array) =>
       output += `${index + 1}/${array.length} ${item.name}\n`;
     );
     console.log(output);
@@ -725,7 +730,7 @@ const ArrayMixin = Mixin.create(Enumerable, {
 
   /**
     Alias for `mapBy`.
-    
+
     Returns the value of the named
     property on all items in the enumeration.
 
--- a/packages/@ember/-internals/runtime/lib/system/array_proxy.js
+++ b/packages/@ember/-internals/runtime/lib/system/array_proxy.js
@@ -111,11 +111,11 @@ export default class ArrayProxy extends EmberObject {
       this._arrangedContentRevision = value(this._arrangedContentTag);
     }
 
-    this._addArrangedContentArrayObsever();
+    this._addArrangedContentArrayObserver();
   }
 
   willDestroy() {
-    this._removeArrangedContentArrayObsever();
+    this._removeArrangedContentArrayObserver();
   }
 
   /**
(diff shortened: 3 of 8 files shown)


pzuraq:
cc @wagenet, this should fix the bug we found

packages/@ember/-internals/runtime/lib/system/array_proxy.js
```diff
@@ -11,10 +11,13 @@ import {
   removeArrayObserver,
   replace,
 } from '@ember/-internals/metal';
+import { symbol } from '@ember/-internals/utils';
 import EmberObject from './object';
 import { isArray, MutableArray } from '../mixins/array';
 import { assert } from '@ember/debug';
 
+const CONSUMED = symbol('ARRAY_PROXY_CONSUMED');
```
chadhietala: Is there a reason to use this over a `WeakMap`?

pzuraq: Not really, we could definitely update it to a `WeakMap` or `WeakSet`. I figured there _may_ be perf implications, but don't have a strong opinion one way or the other

rwjblue: Ya, I generally agree with @chadhietala here. I'd prefer to avoid adding arbitrary fields to the underlying objects. I doubt there is much difference between this implementation and a WeakMap/WeakSet based one, but I'm happy to test if you'd like to assuage concerns...

rwjblue (commented):
Are we certain that these two places are the only places that "consume" the array proxy? I haven't traced all the codepaths for the underlying mixins (`MutableArray`, `ArrayMixin`, etc)...

pzuraq:
So the general contract for any `EmberArray` class is that all accesses should go through `objectAt`, which makes it the primary API for accessing any item in the array. Then `length` is the only other API that you can use to view the state of the array, and I think the fact that these are the two values that the array proxy keeps in sync confirms this.

Other than that, the one thing that requires us to add observers is adding an observer to the array proxy itself, since it then must start forwarding events. This is why the PR changes all uses to calling `arr.addArrayObserver` if it exists as well.

We _do_ expose `Ember.addArrayObserver`, but I couldn't find a single usage of it on `EmberObserver`, and it's not documented. I think this PR effectively breaks array proxies if someone is using that API. We could make it a wrapper which calls `arr.addArrayObserver` if it exists, or we could drop it from being exported.

wagenet:
I'm pretty sure our app hits the problem in areas other than ArrayProxy. I'm working on getting more details. @wycats also seemed to think it was a larger issue.

pzuraq:
@wagenet yeah, it's definitely possible, if you have other cases we could definitely use them! I think lazily setting up the observers here is still a good change regardless, so our thinking was let's get this merged for now and keep finding more reproductions. If we find enough to suggest that this will be a major breaking issue, we'll have to rethink certain aspects of autotracking.

wagenet:
I have verified that this does not fix all of the issues in my app. I'm investigating further.

rwjblue:
> I'm pretty sure our app hits the problem in areas other than ArrayProxy. I'm working on getting more details. @wycats also seemed to think it was a larger issue.

Yep, totally agree that this doesn't fix everything, but I think it's a good change regardless. Specifically, making ArrayProxy lazier on its own is a "nice to have"...

rwjblue (approved):

packages/@ember/-internals/glimmer/tests/integration/components/tracked-test.js
```diff
@@ -44,6 +45,27 @@ if (EMBER_METAL_TRACKED_PROPERTIES) {
         this.assertText('max jackson | max jackson');
       }
 
+      '@test returning new promise array proxies does not cause perpetual rerenders'() {
```
hjdivad: I'd recommend renaming this.  This PR fixes `ArrayProxy` specifically, but the underlying problem is still there for promise array proxies (or any non-memoized getter that reads a `@tracked` value that causes it to be later mutated)

Maybe
```js
'@test returning ArrayProxy in un-memoized getter does not cause perpetual rerenders'
```

hjdivad (commented):
Also, i think @rwjblue mentioned this out of band as well, but we should fix `Ember.addArrayObserver` to delegate;  having `ArrayProxy` not consume on `addArrayObserver` seems like a mega-troll for users who run into it.  It may not be used in `EmberObserver` but I wouldn't want to bet on it not being used in apps.

pzuraq:
I was actually mistaken about `Ember.addArrayObserver` existing, I saw it in a test that I assumed was for globals, but it was for documentation.

packages/@ember/-internals/glimmer/tests/integration/components/tracked-test.js
rwjblue: @pzuraq - I think this is the only remaining issue to address before landing...

hjdivad (approved):
❤️ 👍

rwjblue:
@pzuraq - I think this change is still something we want to land, even though it isn't _required_ to fix the infinite rendering issues that originally caused us to look at this. Would you mind rebasing?

scottmessinger:
It looks like I'm running into this same error using `ObjectProxy.extend(PromiseProxyMixin)`. As soon as a return a ObjectPromiseProxy, things infinitely render. Any chance what I'm running into is related to this error?

pzuraq:
@scottmessinger is this with the tracked properties feature flag turned on?

scottmessinger:
@pzuraq I don't think so? How do I do that? I'm using the latest canary. Here's the features list I'm seeing: https://github.com/emberjs/ember.js/blob/master/FEATURES.md

scottmessinger:
@pzuraq Yes! I do have it enabled -- sorry, took me a while to google and find it. Here's what I have:

```
        EMBER_NATIVE_DECORATOR_SUPPORT: true,
        EMBER_METAL_TRACKED_PROPERTIES: true,
        EMBER_GLIMMER_ANGLE_BRACKET_NESTED_LOOKUP: true,
        EMBER_GLIMMER_ANGLE_BRACKET_BUILT_INS: true,
```

scottmessinger:
Also, the infinite rendering is happening if I return a promise from a getter and then pass it to the https://github.com/tildeio/ember-async-await-helper component. That library doesn't use ObjectProxy, so it's possible my issue is caused by something else.

pzuraq:
For sure, if you could upload a reproduction that would be super helpful. This is definitely the kind of issue we're keeping our eyes out for since we're trying to land tracked properties relatively soon!

scottmessinger:
Sure! What's the easiest way to do that these days? It doesn't look like ember twiddle is up to date. Would just a new ember app with octane be the best or is there anything faster?

pzuraq:
Yup, especially for canary bugs since ember-twiddle has never supported canary/canary-flags (maybe we can do some work on that in the near future..)

An app repo would be perfect.

scottmessinger:
@pzuraq Let me know if you need anything else or the failing test case above isn't helpful!

pzuraq:
Sorry, this is on my radar, I've just been super busy and haven't had a chance to dive in. I'll probably be able to get to this sometime next month

scottmessinger:
@pzuraq No problem! Let me know if there's anything else I can do to be helpful!!

krisselden (approved):

packages/@ember/-internals/runtime/lib/mixins/array.js
```diff
@@ -564,8 +565,10 @@ const ArrayMixin = Mixin.create(Enumerable, {
     @property {Boolean} hasArrayObservers
     @public
   */
-  hasArrayObservers: nonEnumerableComputed(function() {
-    return hasListeners(this, '@array:change') || hasListeners(this, '@array:before');
+  hasArrayObservers: descriptor({
```
rwjblue: Don’t we also need to add enumerable: false? Also, is caching important here?

rwjblue (commented):

packages/@ember/-internals/runtime/lib/mixins/array.js
pzuraq: Ah, yes, I meant to do that! 

For caching, I think it depends on how many times we expect folks to run `addArrayObserver` and `removeArrayObserver`, since nothing else really [uses it in the ecosystem](https://www.emberobserver.com/code-search?codeQuery=hasArrayObserver) or the codebase. It seems like that's [also a relatively uncommon thing](https://www.emberobserver.com/code-search?codeQuery=addArrayObserver), but we could also keep it as a CP if we're worried and use `untrack` to access it and prevent the issue.

rwjblue (commented):
Can you add a test to `packages/@ember/-internals/metal/tests/accessors/get_test.js` for the "mutate inside `unknownProperty`" deprecation?

packages/@ember/-internals/runtime/lib/system/array_proxy.js
```diff
@@ -111,11 +111,11 @@ export default class ArrayProxy extends EmberObject {
       this._arrangedContentRevision = value(this._arrangedContentTag);
     }
 
-    this._addArrangedContentArrayObsever();
```
rwjblue: LOL @ this typo, I stared at this line for ~ 3 minutes trying to figure out why it was in the git diff 😝

rwjblue (approved):

PR #17836, opened 2019-04-02
[CHORE] Node 6 is EOL this month

pzuraq:
Trying this out to see if it fixes the SIGSEGV bug


rwjblue (approved):

Turbo87:
this is violating our node support policy (see https://blog.emberjs.com/2016/09/07/ember-node-lts-support.html) and since it was released in the 3.10 beta is now failing all addon CI builds because they still use Node 6. IMHO we should revert this until the addon blueprint was updated to use Node 8 instead.

PR #17842, opened 2019-04-02
[RFC #435] Add feature flag and tests for forwarding modifiers with splattributes

cibernox:


rwjblue (approved):
LGTM, @chancancode mind doing a review pass also?

packages/ember-template-compiler/lib/plugins/assert-modifiers-not-in-components.ts
```diff
@@ -0,0 +1,54 @@
...
+            },
+
+            exit() {
+              scopes.pop();
+            },
+          },
+        },
+        enter(node: AST.ElementNode) {
+          if (node.modifiers.length > 0 && isComponentInvocation(node)) {
+            assert(`Modifiers cannot be applied to components. ${calculateLocationDisplay(
```
chancancode: nit: since we are about to roll this out/tech this to people, let’s avoid the “cannot be applied” wording, and go for something like “Passing modifiers to components require the xyz canary feature, which has not been stabilized yet. See RFC xyz for details.”

FEATURES.md
```diff
@@ -50,3 +50,10 @@ for a detailed explanation.
   with the angle bracket invocation sytnax.
 
   See [RFC #459](https://github.com/emberjs/rfcs/pull/459).
+
+* `ember-glimmer-pass-modifiers-with-splattributes`
```
chancancode: nit: pass -> forward to match the RFC title

cibernox:
@rwjblue before merging I take that we should have a non-alpha release of glimmer containing that feature, right?

chancancode (approved):
I wasn’t able to look closely at the tests on my phone but the rest seems good to me 👍

cibernox:
@chancancode updated

PR #17843, opened 2019-04-02
Remove intimate apis in the router

chadhietala:
Unsure if this should be cherry picked to the release or not.


rwjblue:
The CI failures seem related:

> Test failed: Deprecated HandlerInfos:  handlerInfos are deprecated and associated private apis", source:  (49)

pzuraq (approved):
Reminded me I need to do the same

PR #17845, opened 2019-04-03
[CLEANUP] Removes various deprecated APIs

pzuraq:
Also unsure if these should be picked onto the release, like #17843. We should really make deprecations hard fail as soon as they reach their `until` version, if they don't already.


rwjblue (approved):

igorT:
These definitely should not be picked onto the release at least for now, they break ED. I'm investigating whether the error is in our test suite or something deeper

igorT:
I think it might be only ED tests doing bad things, but I'll confirm once fixed

PR #17846, opened 2019-04-03
[BUGFIX beta] Fix template-only components clearing

chancancode:
The root cause is when the `createInstance` flag is disabled, Glimmer omits the `DidRenderLayout` opcode, which has the unintended consequence of [skipping the `popBlock` call](https://github.com/glimmerjs/glimmer-vm/blob/3b36bbe3ee650647ac530d239244b3108c3a9a4b/packages/%40glimmer/runtime/lib/compiled/opcodes/component.ts#L703), causing an imbalance in the stack. This means additional content unrelated to the component will be treated as part of the component's bounds, causing other errors later on. It also causes other issues such as blocks not getting finalized due to the unbalanced push and pops.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/component-managers/input.ts
+++ b/packages/@ember/-internals/glimmer/lib/component-managers/input.ts
@@ -18,7 +18,7 @@ const CAPABILITIES: ComponentCapabilities = {
   createCaller: true,
   dynamicScope: false,
   updateHook: true,
-  createInstance: false,
+  createInstance: true,
 };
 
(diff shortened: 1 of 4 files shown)


chancancode:
Not all the tests failed as intended because of #17847

rwjblue (approved):

rwjblue:
Can you also create an issue for the underlying problem over in glimmer-vm?

chancancode:
@rwjblue I'm going to get that fixed today

Issue #17840, opened 2019-04-02
event.stopPropagation() no longer stops propagation after removing jQuery


lolmaus:
Hi!

I've built and am maintaining the [ember-drag-sort](https://github.com/kaliber5/ember-drag-sort) addon which happens to be the only drag-n-drop addon in the Ember ecosystem with support for nested lists.

It's been working fine with jQuery. Demo: https://kaliber5.github.io/ember-drag-sort/

I've tried upgrading to Ember 3.8 without jQuery and discovered that `event.stopPropagation()` no longer stops propagation. As a result, events on child lists get overridden by events on parent lists, and reordering nested lists no longer works.

I've pushed the `gen-1` branch which is based on no-jQuery Ember 3.8. Here are some of problematic spots:

* https://github.com/kaliber5/ember-drag-sort/blob/gen-1/addon/components/drag-sort-list.js#L103
* https://github.com/kaliber5/ember-drag-sort/blob/gen-1/addon/components/drag-sort-item.js#L152

(Note that in the latter spot I tried adding `event.preventDefault()` and `return false` which didn't help and might have made things even worse, so it might make sense removing those when you try.)

This looks like a regression in Ember to me, but I can't confirm it myself. Please help!

lolmaus:
@chancancode Hi! What's going on there?

chancancode:
@lolmaus I'm not sure yet, but I ran into the same issue with one of the built-in components, we just didn't notice it because the those haven't been running on CI without jquery, due to some config issues

chancancode:
@simonihmig do you know if this is expected?

chancancode:
I think at least part of the problem is that, the even is not technically "propagating" since it already propagated all the way to ember's root element at this point, so `stopPropagation` only prevents it from propagating further up from that point on. It would be reasonable for us to emulate this in the custom propagation code, however I'm not sure that there is a good way to do it since there is no DOM api for querying the `stopPropagation` flag.

chancancode:
cancelBubble getter may work 🤔

simonihmig:
> do you know if this is expected?

Hm, so as mentioned to @lolmaus in private, returning `false` *should* have prevented bubbling, as this is covered here: https://github.com/emberjs/ember.js/blob/8174b88639c760b7bb53b5f3ea3d3a72b9e6ca1c/packages/%40ember/-internals/views/lib/system/event_dispatcher.js#L359. Not sure why this was not working, maybe some different issue.

[...]

PR #17847, opened 2019-04-03
Fix each-package-tests on CI

chancancode:
We have not been running the test for any of the -internals packages 😭


chancancode:
I had to "disable" the link component's bubbling test for now, it seems to be failing due to #17840

Issue #17703, opened 2019-03-04
Should @ember/instrumentation be public API?


xg-wang:
[@ember/instrumentation](https://github.com/emberjs/ember.js/blob/v3.8.0/packages/%40ember/instrumentation/index.ts#L36) has been marked private since @rwjblue 's [PR](https://github.com/emberjs/ember.js/pull/11362) in 2015, should the module be considered as public API?

runspired:
I think this is a question for RFC discussion

xg-wang:
Would love to write one. Should this issue moved to rfc repo?

xg-wang:
See comment https://github.com/emberjs/ember.js/pull/17848#issuecomment-480831971

PR #11362, opened 2015-06-07
[DOC beta] Enforce public/private declaration for API docs.

rwjblue:
This PR ensures that all YUIDoc comment blocks (those starting with `/**`) are properly marked as either `@private` or `@public`.

A number of times throughout the Ember 1.x cycle we have internally struggled with what things are or are not public. This PR "solves" that by making it extremely clear.

In addition to documenting each comment block, a new JSCS rule has been added to ensure **all** documentation blocks are labeled properly in the future.

---

Please note: This PR makes an initial attempt based on my best understanding. These `@public` and `@private` declarations can absolutely be reviewed and changed as needed.

PR #17848, opened 2019-04-03
[DOC Release] Mark @ember/instrumentation as public

xg-wang:
Based on description in https://github.com/emberjs/rfcs/blob/master/text/0176-javascript-module-api.md#organize-by-mental-model, `@ember/instrumentation` is deemed as public module. 
https://github.com/emberjs/ember.js/pull/11362 may have incorrectly marked it as private.
Closes https://github.com/emberjs/ember.js/issues/17703
cc/ @chadhietala

### DIFF
--- a/packages/@ember/instrumentation/index.ts
+++ b/packages/@ember/instrumentation/index.ts
@@ -33,7 +33,7 @@ interface MaybePerf {
 
 /**
 @module @ember/instrumentation
-@private
+@public
 */
 
 /**
@@ -84,7 +84,7 @@ interface MaybePerf {
 
   @class Instrumentation
   @static
-  @private
+  @public
 */
 export let subscribers: Subscriber<any>[] = [];
 let cache: { [key: string]: Listener<any>[] } = {};
@@ -129,7 +129,7 @@ function isCallback<Binding, Result>(
   @param {Object} payload
   @param {Function} callback Function that you're instrumenting.
   @param {Object} binding Context that instrument function is called with.
-  @private
+  @public
 */
 export function instrument<Result>(
   name: string,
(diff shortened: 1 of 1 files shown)


chancancode:
I do think the _module_ itself should be importable (if it's not already), but I'm not sure that we should mark the _methods_ as public. I was under the impression that they are in a similar state as the `ViewUtils` methods, but I could be wrong.

xg-wang:
@chancancode What would be the difference if we mark the module as public but not the methods?

chadhietala:
From the RFC:

> And finally, some packages that may be used by internals, extensions, or addons but are not used day-to-day by app developers:
>
> - @ember/instrumentation - Instrumentation hooks for measuring performance.
> - @ember/debug - Utility functions for debugging, and hooks used by debugger tools like Ember Inspector.

And the RFC enumerated what should be exported as public API:

https://github.com/emberjs/rfcs/blob/master/text/0176-javascript-module-api.md#emberinstrumentation

chancancode:
I think the context was that we agreed to give some reasonably popular intimate APIs an import path (so that they can be used at all), without necessarily committing to semver API stability. There are other private things on there too, e.g. `Map`, `guidFor`. We can discuss stabilizing them (or perhaps this particular one really is just a mistake), but I was just disagreeing on the fact that it was in the modules RFC automatically implies they are `@public`. I thought the original bug was that there weren't reachable (not importable) despite what the RFC said.

rwjblue:
We chatted about this for a bit at the most recent core team meeting. The rough consensus is that these APIs are not what we would prefer them to be, and instead of making them public now we should consider coming up with better alternatives. 

To be clear, this does **not** mean that the `@ember/instrumentation` APIs are at risk for "randomly being removed", they very clearly qualify as intimate APIs and therefore would require a deprecation RFC and the standard intimate API deprecation cycle (at least 1 LTS must include the deprecation).

rwjblue:
@xg-wang thank you for spurring the conversation here! I'm going to go ahead and close this PR now that we've discussed and come to a conclusion.

PR #17855, opened 2019-04-03
[BUGFIX beta] Expose computed getter functions

pzuraq:
A few addons, namely ember-macro-helpers, still rely access to
the private `_getter` function on CP descriptors. There's no
easy way around this, since ember-macro-helpers can be used to
combine _existing_ macros, so this fix exposes the _getter function
for the time being until alternative solutions can be made.


rwjblue (approved):
Can you add a TODO or something mentioning why we need it and that we should remove (likely with a private API deprecation)?

Other than that, seems fine to me...

rwjblue (approved):

PR #17858, opened 2019-04-04
[FEATURE] Implement injection parameter normalization RFC.

pzuraq:
Implements https://github.com/emberjs/rfcs/pull/451, and fixes the error message so it won't throw if properties aren't passed (`new EmberObject()`)


rwjblue:
Just updated to more closely match https://github.com/emberjs/rfcs/pull/451, should be ready for review.

packages/@ember/-internals/runtime/lib/system/core_object.js
```diff
@@ -40,13 +45,15 @@ const factoryMap = new WeakMap();
 
 const prototypeMixinMap = new WeakMap();
 
-let PASSED_FROM_CREATE;
-let initCalled; // only used in debug builds to enable the proxy trap
+const initCalled = DEBUG ? new WeakSet() : undefined; // only used in debug builds to enable the proxy trap
```
chadhietala: I believe you are going to fail IE11 builds due to the usage of `WeakSet`.

chadhietala (changes requested):
Overall this looks good to me. Just a small nit.

rwjblue (approved):

PR #17874, opened 2019-04-06
[BUGFIX beta] Fix no-jQuery EventDispatcher to prevent events bubbling up when event.stopPropagation() was called

### LINKED ISSUE #17840: event.stopPropagation() no longer stops propagation after removing jQuery

simonihmig:
Fixes #17840

### DIFF
--- a/packages/@ember/-internals/views/lib/system/event_dispatcher.js
+++ b/packages/@ember/-internals/views/lib/system/event_dispatcher.js
@@ -360,6 +360,8 @@ export default EmberObject.extend({
                 event.preventDefault();
                 event.stopPropagation();
                 break;
+              } else if (event.cancelBubble === true) {
+                break;
               }
             } else if (target.hasAttribute('data-ember-action')) {
               if (actionHandler(target, event) === false) {
--- a/packages/@ember/-internals/glimmer/tests/integration/components/link-to/routing-angle-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/components/link-to/routing-angle-test.js
@@ -8,7 +8,6 @@ import { alias } from '@ember/-internals/metal';
 import { subscribe, reset } from '@ember/instrumentation';
 import { Route, NoneLocation } from '@ember/-internals/routing';
 import { EMBER_IMPROVED_INSTRUMENTATION } from '@ember/canary-features';
-import { jQueryDisabled } from '@ember/-internals/views';
 
 if (EMBER_GLIMMER_ANGLE_BRACKET_BUILT_INS) {
   // IE includes the host name
@@ -882,12 +881,7 @@ if (EMBER_GLIMMER_ANGLE_BRACKET_BUILT_INS) {
           .then(() => {
             assert.equal(this.$('#contact').text(), 'Contact', 'precond - the link worked');
 
-            if (jQueryDisabled) {
-              // BUG: https://github.com/emberjs/ember.js/issues/17840
-              assert.equal(hidden, 1, 'The link bubbled anyway');
-            } else {
-              assert.strictEqual(hidden, 0, "The link didn't bubble");
-            }
+            assert.equal(hidden, 0, "The link didn't bubble");
           });
       }
 
@@ -938,13 +932,7 @@ if (EMBER_GLIMMER_ANGLE_BRACKET_BUILT_INS) {
           })
           .then(() => {
             assert.equal(this.$('#contact').text(), 'Contact', 'precond - the link worked');
-
-            if (jQueryDisabled) {
-              // BUG: https://github.com/emberjs/ember.js/issues/17840
-              assert.equal(hidden, 1, 'The link bubbled anyway');
-            } else {
-              assert.strictEqual(hidden, 0, "The link didn't bubble");
-            }
+            assert.equal(hidden, 0, "The link didn't bubble");
           });
       }
 
--- a/packages/@ember/-internals/glimmer/tests/integration/components/link-to/routing-curly-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/components/link-to/routing-curly-test.js
@@ -6,7 +6,6 @@ import { alias } from '@ember/-internals/metal';
 import { subscribe, reset } from '@ember/instrumentation';
 import { Route, NoneLocation } from '@ember/-internals/routing';
 import { EMBER_IMPROVED_INSTRUMENTATION } from '@ember/canary-features';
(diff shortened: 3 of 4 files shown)


chancancode (approved):
:clap:

chancancode:
Does return false work?

packages/@ember/-internals/views/lib/system/event_dispatcher.js
```diff
@@ -360,6 +360,8 @@ export default EmberObject.extend({
                 event.preventDefault();
                 event.stopPropagation();
                 break;
+              } else if (event.cancelBubble === true) {
+                break;
```
chancancode: This is probably not the correct nesting for this. I believe we want to check the flag after the event handler has run (currently the check runs when if there is no event handler). Further, we probably need the same treatment for the action branch too. So I think you probably want to do it before setting the target to the parent.

I feel like this should have failed the test tho?

chancancode (changes requested):

packages/@ember/-internals/views/lib/system/event_dispatcher.js
chancancode: Also, we should make sure to mirror the jquery/dom behavior that stop propagation doesn’t prevent handlers on the same element from firing

chancancode: On that note, we probably have another bug here. It seems like the code assumes you can either have a component element handler, or action handlers, but not both. I believe with the possibility of forwarding action modifiers through splattributes, this is no longer a correct assumption.

simonihmig: > I believe we want to check the flag after the event handler has run

In what scenario would that not be the case? `viewHandler(target, event)` will always be called before our `cancelBubble` check, right? 

> Further, we probably need the same treatment for the action branch too.

I tried to add a (failing) test for that, until I realized that the action handler will not receive a reference to the event, so cannot call `stopPropagation()`, or do I miss something?

> stop propagation doesn’t prevent handlers on the same element from firing

How would there be two different handlers (component event handlers) on the same element? You mean the case you mentioned in your following comment? 👇

> It seems like the code assumes you can either have a component element handler, or action handlers, but not both. I believe with the possibility of forwarding action modifiers through splattributes, this is no longer a correct assumption.

Hm, I see... 🤔Good catch! But we should tackle this in a separate PR I think?

simonihmig:
> Does return false work?

Yes, I think so! This was already  covered in a test: https://github.com/emberjs/ember.js/blob/561759e4d3cd2f978deb057e4e1f52d03098f40d/packages/%40ember/-internals/glimmer/tests/integration/event-dispatcher-test.js#L109

packages/@ember/-internals/views/lib/system/event_dispatcher.js
chancancode: > `viewHandler(target, event)` will always be called before our `cancelBubble` check, right?

You're right!


> action handler will not receive a reference to the event

Don't we pass it [here](https://github.com/emberjs/ember.js/blob/561759e4d3cd2f978deb057e4e1f52d03098f40d/packages/%40ember/-internals/views/lib/system/event_dispatcher.js#L292)?

> How would there be two different handlers (component event handlers) on the same element? You mean the case you mentioned in your following comment? 👇

Yep that's one case I could think of. I think you could also do `<div {{action "foo"}} {{action "bar"}} />`? Does that work? (Looking at the code, I'm not sure where this is handled.)

> But we should tackle this in a separate PR I think?

Seems fine to me!

simonihmig: > Don't we pass it here?

Yes, but that goes to https://github.com/emberjs/ember.js/blob/83263728b116fac6d4e8fcc798babf555a62ac2b/packages/%40ember/-internals/glimmer/lib/modifiers/action.ts#L125-L181, and there the event is not passed to the action handler. I believe this is how the action helper API works: the (element space(!), not closure) action helper passes only the custom args, not the event itself. That's why we have things like `bubbles=false`.

>  I think you could also do <div {{action "foo"}} {{action "bar"}} />? Does that work? (Looking at the code, I'm not sure where this is handled.)

We have that loop here: https://github.com/emberjs/ember.js/blob/561759e4d3cd2f978deb057e4e1f52d03098f40d/packages/%40ember/-internals/views/lib/system/event_dispatcher.js#L288

And a related test here, though for different events: https://github.com/emberjs/ember.js/blob/561759e4d3cd2f978deb057e4e1f52d03098f40d/packages/%40ember/-internals/glimmer/tests/integration/helpers/element-action-test.js#L1323-L1360

I tried to add a similar test with the same event type, but that fails, as the first handled action handler returns early here: https://github.com/emberjs/ember.js/blob/561759e4d3cd2f978deb057e4e1f52d03098f40d/packages/%40ember/-internals/views/lib/system/event_dispatcher.js#L292

So this seems to be not supported at this point. I guess it should be possible, by always looping over all action handlers, and return `false` if at least one returned `false`!?

simonihmig: Oh, wait a second. Actually that test with multiple `{{action}}` *does*  pass w/ jQuery, so it seems it should actually work, but doesn't in no-jQuery mode! 

I'll bring up another PR to fix this specific issue. As for the issue in this PR, I *think* this PR should be ok as it is?

simonihmig: Here we go: #17876

PR #17876, opened 2019-04-07
[BUGFIX beta] Fix no-jQuery EventDispatcher to allow handling multiple element-space actions

simonihmig:
Related to https://github.com/emberjs/ember.js/pull/17874#discussion_r272815873

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/helpers/element-action-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/helpers/element-action-test.js
@@ -712,6 +712,45 @@ moduleFor(
       this.assert.equal(originalHandlerWasCalled, true, 'the click handler was called');
     }
 
+    ['@test multiple actions with bubbles=false for same event are called but prevent bubbling']() {
+      let clickAction1WasCalled = false;
+      let clickAction2WasCalled = false;
+      let eventHandlerWasCalled = false;
+
+      let ExampleComponent = Component.extend({
+        actions: {
+          clicked1() {
+            clickAction1WasCalled = true;
+          },
+          clicked2() {
+            clickAction2WasCalled = true;
+          },
+        },
+      });
+
+      this.registerComponent('example-component', {
+        ComponentClass: ExampleComponent,
+        template: strip`
+        <a href="#"
+          {{action "clicked1" on="click" bubbles=false}}
+          {{action "clicked2" on="click" bubbles=false}}
+        >click me</a>`,
+        click() {
+          eventHandlerWasCalled = true;
+        },
+      });
+
+      this.render('{{example-component}}');
+
+      runTask(() => {
+        this.$('a').trigger('click');
+      });
+
+      this.assert.ok(clickAction1WasCalled, 'the first clicked action was called');
+      this.assert.ok(clickAction2WasCalled, 'the second clicked action was called');
+      this.assert.notOk(eventHandlerWasCalled, 'event did not bubble up');
+    }
+
     ['@test it should work properly in an #each block']() {
       let editHandlerWasCalled = false;
 
@@ -1357,6 +1396,40 @@ moduleFor(
       });
 
       this.assert.ok(doubleClickActionWasCalled, 'the doubleClicked action was called');
+    }
+
+    ['@test allows multiple actions for same event on a single element']() {
+      let clickAction1WasCalled = false;
+      let clickAction2WasCalled = false;
+
+      let ExampleComponent = Component.extend({
+        actions: {
+          clicked1() {
+            clickAction1WasCalled = true;
+          },
+          clicked2() {
+            clickAction2WasCalled = true;
+          },
+        },
+      });
+
+      this.registerComponent('example-component', {
+        ComponentClass: ExampleComponent,
+        template: strip`
+        <a href="#"
+          {{action "clicked1" on="click"}}
+          {{action "clicked2" on="click"}}
+        >click me</a>`,
+      });
+
+      this.render('{{example-component}}');
+
+      runTask(() => {
+        this.$('a').trigger('click');
+      });
+
+      this.assert.ok(clickAction1WasCalled, 'the first clicked action was called');
+      this.assert.ok(clickAction2WasCalled, 'the second clicked action was called');
     }
 
     ['@test it should respect preventDefault option if provided']() {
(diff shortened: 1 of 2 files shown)


chancancode:
Looking good!

https://github.com/emberjs/ember.js/blob/18d9087797ef833c160f9d85383159e1026b9d68/packages/%40ember/-internals/views/lib/system/event_dispatcher.js#L361-L371

It bothers mean that these two branches are quite diverged. You would really expect this to be something like:

```js
let rtn = false;

if (getElementView(target)) {
  rtn = viewHandler(target, event);
} else if (target.hasAttribute('data-ember-action')) {
  rtn = viewHandler(target, event);
}

if (rtn === false) {
  event.preventDefault();
  event.stopPropagation();
  break;
}
```

But it's not. I suspect this is probably a long standing, or maybe it was introduced during Glimmer 2 migration, but it seems wrong that returning `false` in component handlers vs action handlers mean different things. Specifically, in components it means `preventDefault` + `stopPropagation`, in the other it does something else. And it looks like you can't even `return false` in your own action handlers. Maybe we should try to track and fix that separately.

packages/@ember/-internals/views/lib/system/event_dispatcher.js
```diff
@@ -285,13 +285,16 @@ export default EmberObject.extend({
...
         }
 
+        let result = true;
         for (let index = 0; index < actions.length; index++) {
           let action = actions[index];
 
           if (action && action.eventName === eventName) {
-            return action.handler(event);
+            // return false if any of the action handlers returns false
+            result = result && action.handler(event);
```
chancancode: I think this is not _quite_ right: if the first handler returned false, that will prevent the remaining handlers (on the same elements) from running. That would be `event.stopImmediatePropagation()` semantics in jQuery, and I don't think that's what we want here.

chancancode:
By the way, is this right?

https://github.com/emberjs/ember.js/blob/18d9087797ef833c160f9d85383159e1026b9d68/packages/%40ember/-internals/views/lib/system/event_dispatcher.js#L342

I understand that the native `mouseEnter/Leave` events does not themselves bubble, however, since we are synthetically bubbling them, if you do return false or call `stopPropagation`, should we not stop the synthetic bubbling (i.e. breaking out of the loop)? Also the if you can `preventDefault` on the synthetic event (or return false in the component handler), are we supposed to call `preventDefault` on the underlying mapped event?

(The answer to those are whatever jQuery does, but I don't happen to know the answer to _that_. Anyway, it's not related to this patch so we can track/fix it separately if needed.)

packages/@ember/-internals/views/lib/system/event_dispatcher.js
simonihmig: Ah yes, excellent catch! I switched the order of those `&&` operands, and added another test to cover this!

simonihmig:
> It bothers mean that these two branches are quite diverged. You would really expect this to be something like:

Not sure about this, but @rwjblue might know? This was already present here: https://github.com/rwjblue/ember-native-dom-event-dispatcher/blob/master/addon/index.js#L124-L133, which the no-jquery code was based on AFAIK.

>  if you do return false or call stopPropagation, should we not stop the synthetic bubbling (i.e. breaking out of the loop)?

First, this loop is a bit different from the other (doing the synthetic bubbling), in that it tries only to mimic how mouseenter events are triggered, i.e. it dispatches separate events, and does not bubble up all the way to the root element. And by dispatching separate events, calling stopPropagation on one of them should not affect the others I think, as this would mean we would introduce different semantics for the mouseenter event, wouldn't it? (with normal event handlers stopPropagation has no effect for mouseenter, so why should it have here?)

Btw, this is related: https://github.com/emberjs/rfcs/issues/474. Have been thinking about this a while ago, but just today submitted this issue. Maybe you have some thoughts?

> The answer to those are whatever jQuery does, but I don't happen to know the answer to that.

The implementation of this mouseenter support was largely based on what jQuery does, so I think it *should* do more or less the same. But would have to check again to make sure...

chancancode:
All sounds reasonable to me! Thanks for working on this!

I'm thinking we would back port these to as far back as LTS, assuming these two bug exists those branches, what do you think? (also cc @rwjblue)

simonihmig:
> I'm thinking we would back port these to as far back as LTS, assuming these two bug exists those branches, what do you think?

Yes, sounds reasonable. Pretty sure those bugs must have existed for any other version as well!

rwjblue:
> I'm thinking we would back port these to as far back as LTS, assuming these two bug exists those branches, what do you think? (also cc @rwjblue)

agree

packages/@ember/-internals/glimmer/tests/integration/helpers/element-action-test.js
```diff
@@ -712,6 +712,45 @@ moduleFor(
...
+        >click me</a>`,
+        click() {
+          eventHandlerWasCalled = true;
+        },
+      });
+
+      this.render('{{example-component}}');
+
+      runTask(() => {
+        this.$('a').trigger('click');
```
chriskrycho: A question about this (as @nlfurniss and I are working on upgrading an app): this PR says it is fixing the no-jQuery `eventDispatcher`, but this test *appears* (and perhaps I'm simply misunderstanding this?) to be adding tests here and below for jQuery-based event trigger. Is this using jQuery *only* for event triggering, but *not* for dispatch? Thanks!

simonihmig: @chriskrycho You mean because of that `$()` call *looking* like jQuery? 

If so, it is actually *not*  jQuery here. Was also confused when I first saw this. But there is some test infrastructure that adds this without actually using jQuery. Probably helped migrating tests away from jQuery back then (my *guess*). See 
* https://github.com/simonihmig/ember.js/blob/52b6657d1984a2b524d980be7a290de416053bc0/packages/internal-test-helpers/lib/test-cases/abstract.js#L88 and 
* https://github.com/simonihmig/ember.js/blob/52b6657d1984a2b524d980be7a290de416053bc0/packages/internal-test-helpers/lib/node-query.js

chriskrycho: We thought that might be the case but wanted to double check. Thank you very much for clarifying!

PR #17891, opened 2019-04-09
[beta] Revert "[CHORE] Node 6 is EOL this month"

Turbo87:
Reverts emberjs/ember.js#17836

### DIFF
--- a/package.json
+++ b/package.json
@@ -165,7 +165,7 @@
     "typescript-eslint-parser": "^22.0.0"
   },
   "engines": {
-    "node": "8.* || >= 10.*"
+    "node": "6.* || 8.* || >= 10.*"
   },
   "ember-addon": {
     "after": "ember-cli-legacy-blueprints"


rwjblue (changes requested):
Please update to revert the `engines` change, but not the .travis.yml changes.

Turbo87:
hmm... why? seems bad to not test what we claim to support 🤔

rwjblue:
Feel free to fix the issue that https://github.com/emberjs/ember.js/pull/17836 was addressing in other ways then...

rwjblue:
The CI here is an example of exactly the failure on Node 6:

https://travis-ci.org/emberjs/ember.js/jobs/517663153

Turbo87:
> the issue that #17836 was addressing

"the SIGSEGV bug" is not a very verbose bug description 😉 

do we have any information on why this happens, when it started happening, etc?

pzuraq:
@Turbo87 unfortunately no, it started showing up and killing all of the builds 😕 I searched around after being blocked for some time, all I could find is that it indicated a segmentation fault, so something fairly low level. No results for `SIGSEGV` in connection to Babel.

Our best guess, and the reason we merged this PR, is that something low level upstream changed in a way that doesn't support Node 6 anymore, since it was being EOL'd. It could be something they knew would break, or something they didn't, we just aren't sure.

Our reasoning for this was that by the time Ember 3.10.0 releases, we will not be officially supporting Node v6 anymore, so we reasoned it was not problematic to drop support during beta and unblock our builds.

Turbo87:
@pzuraq the problem is that all the addons that follow our blueprint still use Node 6 on CI, but they *do* test the beta releases of Ember which now causes CI to fail for a lot of addons :-/

Turbo87:
@rwjblue @pzuraq done

rwjblue (approved):

PR #17910, opened 2019-04-12
[DEPRECATION] Add deprecation for Function.prototype extensions.

rwjblue:
Implements https://github.com/emberjs/rfcs/pull/0272.

TODO:

- [x] Deprecation guide entry https://github.com/ember-learn/deprecation-app/pull/330
- [x] Add url to deprecations

### DIFF
--- a/packages/@ember/-internals/runtime/lib/ext/function.js
+++ b/packages/@ember/-internals/runtime/lib/ext/function.js
@@ -4,8 +4,10 @@
 
 import { ENV } from '@ember/-internals/environment';
 import { on, computed, observer } from '@ember/-internals/metal';
+import { deprecate } from '@ember/debug';
+import { FUNCTION_PROTOTYPE_EXTENSIONS } from '@ember/deprecated-features';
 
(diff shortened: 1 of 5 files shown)


packages/@ember/-internals/runtime/lib/ext/function.js
```diff
@@ -77,6 +79,14 @@ if (ENV.EXTEND_PROTOTYPES.Function) {
       enumerable: false,
       writable: true,
       value: function() {
+        deprecate(
+          `Function prototype extensions have been deprecated, please migrate from function(){}.property('bar') to computed('bar', function() {}).`,
+          false,
+          {
+            id: 'ember-runtime.function.property',
+            until: '4.0.0',
```
btecu: Do we also need `since: '3.10.0'` (or `'3.11.0'`) here?

rwjblue: No, `deprecate`'s options do not include a `since` field. The intent is for the `@ember/deprecated-features` exports to do that, and ultimately be used to enable svelte infrastructure.

PR #17912, opened 2019-04-12
Ensure test harness provides a way to debug navigation away.

rwjblue:
When developing certain areas of Ember, a bug leads to navigation (e.g.  `LinkComponent` doesn't handle the click event properly and the browser attempts to actually navigate to whatever the href was with a full reload). This snippet attempts to make that situation somewhat easier to detect and resolve...


chancancode (approved):
OMG! yasss 😱

PR #17924, opened 2019-04-15
Remove babel-plugin-filter-imports dependency.

rwjblue:
It has been unused for some time...


rwjblue:
Whoops, this is used by stripForProd code.

Issue #16973, opened 2018-09-13
Router Service: Default Query Param Values are Serialized to the URL


iezer:
I have a route with a default query param specified in the controller. If I transition to this route using a `link-to`, a controller action, or a route action, the default query param is not serialized to the URL. However, if I transition by using the router service, by calling `router.transitionTo(...)`, the default value does appear in the URL.

Here's a [reproduction twiddle](https://ember-twiddle.com/33b0f6f502cea8c109bcdaf2e7cb7dc9?openFiles=templates.application.hbs%2C&route=%2Fqp-route%3Fpage%3D1).

Possibly related to #16594, although I don't get any errors in the console.

Exelord:
Confirmed!

iezer:
Turns out this is deliberate behavior described in the [Router Service RFC](https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md#query-parameter-semantics). It's implemented [here](https://github.com/emberjs/ember.js/blob/6070ad8e612254c4b026f8f9ca4f0639469684cd/packages/%40ember/-internals/routing/lib/system/router.js#L822-L824) as part of [this commit](https://github.com/emberjs/ember.js/commit/3b9489cf6acb61c5ee22f8016f7486ef1e7133d4) and there's even a [test](https://github.com/emberjs/ember.js/commit/3b9489cf6acb61c5ee22f8016f7486ef1e7133d4#diff-ea57fd4a82042201aae1f12d12af03a8R241).

This behavior is not documented anywhere other than the RFC so I'll open a PR with some documentation.

iezer:
@rwjblue the [Router Service RFC](https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md#query-parameter-semantics) describes adding a symbol `Ember.DEFAULT_VALUE` exposed as `import { DEFAULT_VALUE } from 'ember-routing';` for hiding default query params when using the Router service, but it seems like this was never implemented. Is it still part of the plan? If so I'd be happy to open work on it.

iezer:
Adding @miguelcobain and @chadhietala since you've worked on the RFC and Router service and may have some comments.

rwjblue:
Docs were fixed in 
https://github.com/emberjs/ember.js/pull/16977, considering this closed...

PR #17925, opened 2019-04-15
[FEATURE Router Service] DEFAULT_VALUE

meirish:
This is an attempt to add the ability to remove default query params from the URL when routing using the new Router service. This functionality is spec'd in the RFC for the Router Service in this section: https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md#query-parameter-semantics but hasn't been implemented.

Open questions: 
Are there other use cases / other tests that need to be added?
Is the finalizeQueryParams hook too late for setting the state everywhere it needs to be set?


jrjohnson:
@chadhietala any thoughts on this? I ran into #16973 and spent a while debugging before I realized this was a feature. I think getting this export in would make it much easier to document the behavior and save some time in the future.

meirish:
Hi friends! Any guidance / movement on this? Must not be a pain point for folks. Let me know if this approach seems sound or if there's any changes you'd want to see!

Thanks! ❤️

pzuraq:
I can't speak to why the feature wasn't implemented as spec'd, definitely seems a bit odd, but I do know that we're definitely looking to do a more thorough overhaul of query parameters in the near future. I don't think the specified behavior in the original RFC was far reaching enough, it still tied us to Routes/Controllers defining QPs, etc.

I've been working with @NullVoxPopuli on his [current queryParams RFC](https://github.com/NullVoxPopuli/rfcs/blob/router-computed-query-params/text/0000-router-service-computed-query-params.md), and in that world, new QPs that are not defined on the controller/route would clear by _default_ when transitioning, unless they were marked as sticky, in which case you would have to unset them yourself. I believe we were thinking that you would set the value to `null`/`undefined` to clear it.

Just giving some context here. Personally, I don't think it makes sense to have an additional import to unset values in the long run, I'd rather add a `router.unsetQueryParam` method to the router (mirroring the proposed `router.setQueryParam` method). I think this could be fine as a stop gap, but it also seems like it may cause churn to introduce an API that may be changed relatively soon.

jelhan:
> Personally, I don't think it makes sense to have an additional import to unset values in the long run, I'd rather add a `router.unsetQueryParam` method to the router (mirroring the proposed `router.setQueryParam` method).

This it not about unsetting query params that have been set before. It's about avoiding `RouterService.transitionTo()` of adding query params. Currently using `RouterService.transitionTo()` there is no way to avoid query params in the URL after transition - regardless if they were set before or not.

Assuming there is a route `foo`, which has a query param `bar`. Using `Route.transitionTo('foo')` or `Controller.transitionToRoute('foo')` would cause a URL `/foo` (unless a query param was set previously). Using `RouterService.transitionTo('foo')` you end up with a URL `/foo?bar=`. The URL is getting ugly as hell as soon as there are more query params...

`Ember.DEFAULT_VALUE` would provide a hacky but working solution to avoid this. IMO it's required to make `RouterService.transitionTo()` usable for route's that have query params.

pzuraq:
@jelhan that sounds more like a bug in the router service's `transitionTo` method, I believe based on the RFC and this discussion that the `DEFAULT_VALUE` symbol is specifically about removing QPs that _do_ have values already. I may be wrong though, if anyone has additional context on the original RFC to add that'd be very helpful.

jrjohnson:
> DEFAULT_VALUE symbol is specifically about removing QPs that do have values already.

I read this as `null` being a possible value which would be correctly removed when a transition happens so you don't end up with `/foo?bar=`.

meirish:
@pzuraq - @jelhan's characterization is correct, if you use `RouterService.transitionTo()` it _adds_ the query params to the URL using the default values if you haven't passed any. My understanding of the reasoning for DEFAULT_VALUE in the RFC is that the way the internal routing service checks for default values of query params is expensive and DEFAULT_VALUE would let you opt into that behavior with a quick and inexpensive check against a constant value.

pzuraq:
So, I can see that interpretation of the original RFC text. I actually think it's a bit underspecified here, because I believe the intent of the RFC was to handle a specific strange edge case of QP behavior, which is that non-nullish defaults still result in an empty QP, but doesn't cover what to do with nullish values in general.

To demonstrate, here's a [quick twiddle](https://ember-twiddle.com/ba3877c0f312a95652682803849eb62a?openFiles=controllers.application.js%2C&route=%2F%3Ffoo%3Dbaz).

We can see that there are 2 cases where the QP is empty:

1. When it's set to the default value, `'bar'`
2. When it's set to a nullish value (`undefined` or `null`).

I don't believe it was the intent of the RFC to remove the nullish behavior, because that doesn't seem to have anything to do with defaults - even though the QPs default is _not_ null/undefined, it will still be removed if we explicitly set the value to it.

jelhan:
> I don't believe it was the intent of the RFC to remove the nullish behavior, because that doesn't seem to have anything to do with defaults - even though the QPs default is _not_ null/undefined, it will still be removed if we explicitly set the value to it.

@pzuraq Am I getting you right, that we should treat this as a bug and open a separate issue for it?

meirish:
I didn’t realize the treatment of nullish values was different - can you think of a way forward on this? Should I edit the PR to treat nullish values the same as equaling DEFAULT _VALUE?

jelhan:
> There was an issue and it was closed because you could use DEFAULT_VALUE... which was never implemented: #16973

That one was not especially about the nullish behavior, was it?

pzuraq:
We discussed this at the core meeting today, and we're still missing some of the original context surrounding the inclusion of `DEFAULT_VALUE`, but it definitely was _not_ meant to be the only way to remove query params from the transition URL.

@meirish and @jelhan, can you provide more details about the use cases that you're trying to solve with this feature? A detailed example would help the most (e.g. I have a QP `foo` with a default value that is undefined, and I'm trying to `transitionTo({ foo: undefined })` and it shows the QP in the URL still)

jrjohnson:
@pzuraq my example:
I have [a route](https://github.com/ilios/frontend/blob/master/app/controllers/user.js#L19-L23) with several query params with a `boolean` `false` default value.  In order to use the `router` service I have to pass `Ember.DEFAULT_VALUE` for [all of them](https://github.com/ilios/frontend/blob/master/app/components/manage-users-summary.js%23L94-L98) otherwise the URL is filled with `isManagingBio=false&isManagingRoles=false` etc.

I have no objection to this getting handled without the `DEFAULT_VALUE` part, I find that to be brittle as changing a queryParam now requires tracking down all of the places it is transitioned to, but if there is no other way then at least this way should be easy to use.

NullVoxPopuli:
The URL is all strings though, booleans aren't a thing. You could set those to blank string to remove them?

jelhan:
> @meirish and @jelhan, can you provide more details about the use cases that you're trying to solve with this feature? A detailed example would help the most (e.g. I have a QP `foo` with a default value that is undefined, and I'm trying to `transitionTo({ foo: undefined })` and it shows the QP in the URL still)

I hope an Ember Twiddle says more than 1000 words: https://ember-twiddle.com/5322cb8ae449518a8f71459d23f9079a?openFiles=templates.foo.hbs%2C

I expect the link to not contain a query param regardless if I use the router service, the `transitionToRoute` method of a controller or the `<LinkTo>` component for the transition. I see a query param in the URL for router service only.

In all cases the transition target is a route having an empty string as default of the query param. I don't see the issue if the default is `null` or `undefined`. I see a similar issue if default is a non-empty string (e.g. `"bar"`) but my use case is an empty string.

lvegerano:
https://github.com/emberjs/ember.js/issues/19492 Related issue. Its mind boogling that this was an acceptable tradeoff.

meirish:
Closing as this has gone quite stale and I don't remember the specifics at this point. I do look forward to seeing any improvement on query params.

PR #17926, opened 2019-04-15
[DOC] Update docs to show `deprecate` function within `@ember/debug` package

### LINKED ISSUE #17185: [docs] API docs should suggest deprecate be imported from @ember/debug
from @rwjblue 
> ...Importing from @ember/debug has significant benefits (ember-cli-babel ensures that any deprecate calls are stripped from prod builds, this is not true of @ember/application/deprecate).

at https://github.com/mike-north/ember-qunit-snapshots/pull/18#discussion_r231283818

Currently, docs indicate that this may only be imported from `@ember/application/deprecations` and that [`@ember/debug` does not export a `deprecate` function](https://www.emberjs.com/api/ember/3.4/modules/@ember%2Fdebug/).

elwayman02:
The current docs show it within the [`@ember/application/deprecations`](https://api.emberjs.com/ember/3.9/functions/@ember%2Fapplication%2Fdeprecations/deprecate) package, when it physically lives within `@ember/debug`.

Fixes #17185

cc: @rwjblue

### DIFF
--- a/packages/@ember/debug/index.ts
+++ b/packages/@ember/debug/index.ts
@@ -211,7 +211,7 @@ if (DEBUG) {
   });
 
   /**
-   @module @ember/application
+   @module @ember/debug
    @public
   */
 
@@ -226,14 +226,14 @@ if (DEBUG) {
     incuring any performance penalty.
 
     ```javascript
-    import { deprecateFunc } from '@ember/application/deprecations';
+    import { deprecateFunc } from '@ember/debug';
 
     Ember.oldMethod = deprecateFunc('Please use the new, updated method', options, Ember.newMethod);
     ```
 
     @method deprecateFunc
     @static
-    @for @ember/application/deprecations
+    @for @ember/debug
     @param {String} message A description of the deprecation.
     @param {Object} [options] The options object for `deprecate`.
     @param {Function} func The new function called to replace its deprecated counterpart.
--- a/packages/@ember/debug/lib/deprecate.ts
+++ b/packages/@ember/debug/lib/deprecate.ts
@@ -24,7 +24,7 @@ export type DeprecateFunc = (message: string, test?: boolean, options?: Deprecat
 */
 /**
   Allows for runtime registration of handler functions that override the default deprecation behavior.
-  Deprecations are invoked by calls to [@ember/application/deprecations/deprecate](https://emberjs.com/api/ember/release/classes/@ember%2Fapplication%2Fdeprecations/methods/deprecate?anchor=deprecate).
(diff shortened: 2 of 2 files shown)


locks:
According to https://github.com/ember-cli/ember-rfc176-data, the current documentation is the correct one, so I'm closing this PR.
When we move to proper packages we should take care to handle these sort of edge cases!

rwjblue:
@locks - The documentation is wrong for `deprecate`, as is ember-rfc176-data.

rwjblue:
Specifically, importing `deprecate` from `@ember/application/deprecations` will result in the build tooling **not** properly stripping those deprecations.

While researching I found a number of (pretty prominent) addons using `import { deprecate } from '@ember/debug` ([see here](https://emberobserver.com/code-search?codeQuery=deprecate.%2B%40ember%2Fdebug&regex=true&sort=score&sortAscending=false)), but there are also quite a few more that `import { deprecate } from '@ember/application/deprecations'` ([see here](https://emberobserver.com/code-search?codeQuery=deprecate.%2B%40ember%2Fapplication%2Fdeprecations&regex=true&sort=score&sortAscending=false)).

🤔

locks:
Hm, should we come up with an action plan next framework meeting?

rwjblue:
> Hm, should we come up with an action plan next framework meeting?

Ya, sounds good. I think either way we need to make both imports work (and have the stripping behaviors), but seems like a great thing to discuss more broadly.

rwjblue:
We discussed this during today's Ember core team meeting:

1. We will update the build system to ensure importing `deprecate` from either `@ember/debug` or `@ember/application/deprecations` have the same production stripping behaviors.
2. Update ember-rfc176-data's mapping.json to allow both locations, and mark the `@ember/application/deprecations` path as `deprecated: true` (this will aid the eslint rule's --fix to allow quick and easy conversion).
3. Update the docs (this PR) to point folks to using `@ember/debug`

bekzod:
added deprecations https://github.com/ember-cli/ember-rfc176-data/pull/140

rwjblue:
@bekzod - We also need to make ember-cli-babel have the same stripping in prod behaviors for @ember/application/deprecations stuff

bekzod:
Ok will add then

rwjblue:
ember-cli-babel was updated in https://github.com/babel/ember-cli-babel/releases/tag/v7.8.0

rwjblue (approved):

rwjblue:
All the pieces are in place now, this should be good to go now.

rwjblue:
Thanks @elwayman02 / @xg-wang / @bekzod for pushing this forward!

Issue #17929, opened 2019-04-16
Tagless components / DOM method error message could be more informative


lupestro:
The message:
```
You can not define a function that handles DOM events in the <...> tagless component since it doesn't have any DOM element.
```
can easily leave the user wondering what to look for specifically in their code. With different ways of invoking things within handlebars and within the code and new options available daily :), it is hard to know precisely to what this does or doesn't apply.

If the message were to identify to the user what it detected in the component that violated the rule, it would save the user from having to remove things one-by-one and edit/build/run until it starts working -- or it finds the next one and gives the same message, at which the user has to remove everything and put things back in one at a time until it falls over. This can be tedious.

@rwjblue pointed out the following points of interest in the source for whoever takes it up:

https://github.com/emberjs/ember.js/blob/2854463ec3aa191da4e035954dafea1be347cf2f/packages/%40ember/-internals/glimmer/lib/component.ts#L731-L750

https://github.com/emberjs/ember.js/blob/d88da1b61d8dfc15138958682252c3bd1f2a28fb/packages/%40ember/-internals/glimmer/tests/integration/components/fragment-components-test.js#L55

rwjblue:
Fixed by https://github.com/emberjs/ember.js/pull/17930

PR #17930, opened 2019-04-16
[BUGFIX beta] Update assertion for events in tagless component to include method names

raido:
This is a take on to tackle issue: #17929 - to make assertion message a bit more explicit by including offending event handler function name in the message itself.

Additionally I wrapped it into DEBUG flag to avoid shipping this to production at all, not sure if this is a right thing to do. If not I can remove the wrap.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/component.ts
+++ b/packages/@ember/-internals/glimmer/lib/component.ts
@@ -12,6 +12,7 @@ import {
   ViewStateSupport,
 } from '@ember/-internals/views';
 import { assert } from '@ember/debug';
+import { DEBUG } from '@glimmer/env';
 import { DirtyableTag } from '@glimmer/reference';
 import { normalizeProperty, SVG_NAMESPACE } from '@glimmer/runtime';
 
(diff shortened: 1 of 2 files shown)


packages/@ember/-internals/glimmer/lib/component.ts
```diff
@@ -727,13 +728,10 @@ const Component = CoreView.extend(
...
       this[BOUNDS] = null;
 
-      // If in a tagless component, assert that no event handlers are defined
-      assert(
-        // tslint:disable-next-line:max-line-length
-        `You can not define a function that handles DOM events in the \`${this}\` tagless component since it doesn't have any DOM element.`,
-        this.tagName !== '' ||
-          !this.renderer._destinedForDOM ||
-          !(() => {
+      if (DEBUG) {
```
rwjblue:
```suggestion
      if (DEBUG && this.renderer._destinedForDOM) {
```

This _should_ fix the errors you are seeing in CI (around instantiating the event dispatcher in non-interactive / aka FastBoot mode).

packages/@ember/-internals/glimmer/lib/component.ts
```diff
@@ -727,13 +728,10 @@ const Component = CoreView.extend(
...
+        let eventName = '';
+        if (this.tagName === '') {
+          eventName = (() => {
```
rwjblue: Can we make this an array instead of a single eventName string? In theory it is possible to have more than one, and I'd love to list them all instead of playing "whack-a-mole"...

raido: It's an array now.

rwjblue (approved):
Looks great, thank you!

raido:
👍 

Closes: #17929

PR #17938, opened 2019-04-17
[BUGFIX beta] Expose mechanism to detect if a property is a computed.

rwjblue:
This adds a private/intimate method for use by the ember-inspector (and possibly other tools) to aid the migration to native decorator version of `computed` (since `computed(function() {}) instanceof ComputedProperty` is no longer usable).

From the docs:

> Allows checking if a given property on an object is a computed property.  For the most part, this doesn't matter (you would normally just access the property directly and use its value), but for some tooling specific scenarios (e.g. the ember-inspector) it is important to differentiate if a property is a computed property or a "normal" property.


RobbieTheWagner (approved):
LGTM! 👍

rwwagner90:
@rwjblue @pzuraq Do we have an ETA on when this might be merged?

PR #17940, opened 2019-04-17
[CLEANUP] Remove `sync` queue from @ember/runloop.

rwjblue:
Was deprecated `until: '3.5.0'`, should be safe to cleanup now.


pzuraq (approved):

GavinJoyce:
I can't seem to find any information on this, there's no entry in the [3.x deprecations list](https://deprecations.emberjs.com/v3.x/) for `ember-metal.run.sync` and I can't seem to find any mention in a previous RFC.

Are the details of why this was deprecated/removed captured anywhere?

rwjblue:
@GavinJoyce - The `sync` queue was used only for `Ember.Binding` flushing which was removed in 3.0, the functionality had to live in Ember through 3.4 due to the 3.x legacy compat addons. When it was originally removed some folks that were manually scheduling into `sync` reported it as a regression so we brought it back as a private API deprecation. This PR cleans up that long standing deprecation.

GavinJoyce:
👍 thanks

mydea:
Hmm, there was no deprecation warning for this. We've been using this, and now the app broke after updating to 3.11. We haven't seen any warning, so this was rather unexpected! Just fyi.

Is it generally safe to refactor this to use `actions` instead?

rwjblue:
@mydea - AFAICT there definitely was a deprecation cycle for this, in the diff of this PR the deprecations (which were labelled as `until: '3.5.0'`) being removed. Each entry point to add something into the `sync` queue should have issued a deprecation from around Ember 3.1 through 3.10 (including a number of LTS versions after which deprecated private API removal is allowed to be removed).

rwjblue:
> Is it generally safe to refactor this to use actions instead?

I think it depends on why you were using `sync`, if you expected deduplicated change notifications (the only "special" thing about `sync` queue) then no `actions` isn't a suitable replacement. I suspect that most folks that happened to use `sync` didn't actually care about that feature and they were just trying to use the "first queue", in which case `actions` is fine.

PR #17941, opened 2019-04-17
[FEATURE EMBER_GLIMMER_FN_HELPER] Initial implementation of fn helper.

rwjblue:
As proposed in https://emberjs.github.io/rfcs/0470-fn-helper.html.


packages/@ember/-internals/glimmer/lib/helpers/fn.ts
```diff
@@ -0,0 +1,18 @@
...
+import { ICapturedArguments } from '@glimmer/runtime/dist/types/lib/vm/arguments';
+import { assert } from '@ember/debug';
+
+function fnHelper({ positional }: ICapturedArguments) {
+  assert(`You must pass a function as the \`fn\` helpers first argument, you passed ${positional.at(0).value()}`, typeof positional.at(0).value() === 'function');
+
+  return () => {
+    let [fn, ...args] = positional.value();
+
+    return fn!['apply'](null, args);
```
pzuraq: So not preserving the calling context?

rwjblue: I honestly think thats a mistake and would end up causing us to bind functions to very bizarre `this` contexts. For example:

```js
  click() {
    this.something();
  }
});
```

Then invoked like:

```hbs
{{x-foo something=(fn this.blah)}}
```

When invoked, the `blah` function from the calling context would have `x-foo` as a `this` context. I think this scenario is massively more likely and is a huge footgun...

pzuraq: The main reasoning was to avoid diverging behavior when you do and don't use `fn`, so for instance:

```hbs
{{x-foo something=this.blah}}
{{x-foo something=(fn this.blah)}}
```

The first one will end up having the `this` context of the component, while the second will not. Since `fn` will only be used for binding arguments, and people will be passing around plain functions without arguments frequently, I think it makes sense to have the failure mode be the same for both. 

That said, I do think most of the time people shouldn't be using the `this` (if ever) so I won't argue strongly for it here.

rwjblue: > The first one will end up having the this context of the component, while the second will not.

Ya, I understand what you mean here. The issue is that _most_ folks are going to thing it should be bound to whatever the `this` is in that template, and not the `x-foo` instance.

> I do think most of the time people shouldn't be using the this (if ever) so I won't argue strongly for it here.

Maybe a nice compromise would be to assert on `this` usage inside the wrapped `fn` function in debug builds. If they used `@action` it wouldn't have the issue, but if they didn't they would at least get an error where they touch `this.ANYTHING` that told them that they didn't bind the function passed in to `fn`. 

What do you think?

pzuraq: Yeah, I think an assertion is fine. It gives us room to revisit this later on as well, if we regret it later 😛

rwjblue: OK, updated with the assertion path for now.

pzuraq (approved):

PR #17958, opened 2019-04-23
[CLEANUP] Make visit() fully async

pzuraq:
Makes the visit() test helper fully async. It now only resolves after
the runloop has fully emptied and everything has settled.

Also adds assert.throwsAsync and expectDeprecationAsync, which allow
us to wrap asynchronous blocks and capture errors and deprecations
within them. This makes this change much easier to deal with.


packages/internal-test-helpers/lib/ember-dev/setup-qunit.ts
```diff
@@ -29,4 +29,24 @@ export default function setupQUnit({ runningProdBuild }: { runningProdBuild: boo
       callback(hooks);
     });
   };
+
+  let originalThrows = QUnit.assert.throws;
+
+  (QUnit.assert as any).throwsAsync = async function(
```
rwjblue: Why do we need this? What does this do better than what `assert.rejects` does?

pzuraq: Discussed this offline, we aligned the naming and the API, but something about our version of `reject` is still working better for the time being.

rwjblue (commented):
Generally :+1:, but I think we need to do two more small tweaks to ensure this change doesn't troll us:

* Ensure that using the global `Promise` triggers an eslint failure
* Ensure that using `async` / `await` in non-test code triggers an eslint failure (I've used https://github.com/brendenpalmer/eslint-plugin-disable-features for this in the past)

rwjblue:
Chatted with @pzuraq, he is going to do a follow up PR to ensure we disable `async` / `await` in non-test code (hopefully today).

Issue #17949, opened 2019-04-18
destroyModifier is called in SSR mode


josemarluedke:
Using modifiers with FastBoot (SSR) should result in them not been executed at all in FastBoot, leaving it to be only executed in client environment. 

Currently: `installModifier` is not called in SSR, however, `destroyModifier` is.
Expected: `destroyModifier` to not be called in SSR.

Reproduction is here: https://github.com/josemarluedke/--ember-modiers-fastboot

rwjblue:
Thank you for reporting (and the reproduction)!

josemarluedke:
This issue is also present when using the on modifier from the feature flag under canary.

```
TypeError: element.removeEventListener is not a function
    at removeEventListener (/var/folders/67/6jb5flcd6d30lkgvj0_92c_m0000gn/T/broccoli-875498bYpOlTJAjU2/out-475-broccoli_merge_trees/assets/@ember/-internals/glimmer.js:8859:1)
    at OnModifierState.destroy (/var/folders/67/6jb5flcd6d30lkgvj0_92c_m0000gn/T/broccoli-875498bYpOlTJAjU2/out-475-broccoli_merge_trees/assets/@ember/-internals/glimmer.js:8832:1)
    at SimpleBlockTracker.destroy (/var/folders/67/6jb5flcd6d30lkgvj0_92c_m0000gn/T/broccoli-875498bYpOlTJAjU2/out-475-broccoli_merge_trees/assets/@glimmer/runtime.js:3606:1)
    at SimpleBlockTracker.destroy (/var/folders/67/6jb5flcd6d30lkgvj0_92c_m0000gn/T/broccoli-875498bYpOlTJAjU2/out-475-broccoli_merge_trees/assets/@glimmer/runtime.js:3606:1)
    at UpdatableBlockTracker.destroy (/var/folders/67/6jb5flcd6d30lkgvj0_92c_m0000gn/T/broccoli-875498bYpOlTJAjU2/out-475-broccoli_merge_trees/assets/@glimmer/runtime.js:3606:1)
    at SimpleBlockTracker.destroy (/var/folders/67/6jb5flcd6d30lkgvj0_92c_m0000gn/T/broccoli-875498bYpOlTJAjU2/out-475-broccoli_merge_trees/assets/@glimmer/runtime.js:3606:1)
[...]

PR #17960, opened 2019-04-23
[FEATURE] Implement `on` modifier.

rwjblue:
See https://emberjs.github.io/rfcs/0471-on-modifier.html for details.

Special thanks to @buschtoens's work on [ember-on-modifier](https://github.com/buschtoens/ember-on-modifier) which heavily inspired this implementation.


packages/@ember/-internals/glimmer/tests/integration/modifiers/on-test.js
```diff
@@ -0,0 +1,324 @@
+import { EMBER_GLIMMER_ON_MODIFIER } from '@ember/canary-features';
+import { moduleFor, RenderingTestCase, runTask } from 'internal-test-helpers';
+import { isChrome, isFirefox } from '@ember/-internals/browser-environment';
+import { privatize as P } from '@ember/-internals/container';
+
+const isIE11 = !window.ActiveXObject && 'ActiveXObject' in window;
```
pzuraq: lol

rwjblue: 😭 indeed

packages/@ember/-internals/glimmer/lib/modifiers/on.ts
```diff
@@ -0,0 +1,241 @@
...
+  }
+}
+
+function addEventListener(
+  element: Element,
+  eventName: string,
+  callback: EventListener,
+  options?: AddEventListenerOptions
+): void {
+  adds++;
```
pzuraq: Assuming these aren't in DEBUGs because you wanted to run these tests against IE?

pzuraq (approved):

packages/@ember/-internals/glimmer/lib/modifiers/on.ts
rwjblue: Yes, exactly. We can ultimately remove if we care, but it's very very cheap and almost certainly won't matter.

packages/@ember/-internals/glimmer/lib/modifiers/on.ts
```diff
@@ -0,0 +1,241 @@
...
+      args.positional.at(1) !== undefined && typeof args.positional.at(1).value() === 'function'
+    );
+    let userProvidedCallback = args.positional.at(1).value() as EventListener;
+    if (userProvidedCallback !== this.userProvidedCallback) {
+      this.userProvidedCallback = userProvidedCallback;
+      this.shouldUpdate = true;
+    }
+
+    if (this.shouldUpdate) {
+      let callback = (this.callback = function(this: Element, event) {
```
buschtoens: Do you think, it'd be a bit faster to only do this, if `DEBUG && passive || !SUPPORTS_EVENT_OPTIONS && once`?

So kinda like:

```ts
if (DEBUG && passive || !SUPPORTS_EVENT_OPTIONS && once) {
  let callback = (this.callback = function(this: Element, event) {
    if (DEBUG && passive) {
      event.preventDefault = () => {
        assert(
          `You marked this listener as 'passive', meaning that you must not call 'event.preventDefault()': \n\n${userProvidedCallback}`
        );
      };
    }

    if (!SUPPORTS_EVENT_OPTIONS && once) {
      removeEventListener(this, eventName, callback, options);
    }
    return userProvidedCallback.call(null, event);
  });
} else {
  this.callback = userProvidedCallback;
}
```

Or can we neglect this?

packages/@ember/-internals/glimmer/lib/modifiers/on.ts
```diff
@@ -0,0 +1,241 @@
...
+    // `{ once: true | false, passive: true | false, capture: true }
+    //
+    // `once` is handled via a custom callback that removes after first
+    // invocation so we only care about capture here as a boolean
+    element.removeEventListener(eventName, callback, true);
+  } else {
+    // used only in the following cases:
+    //
+    // * where there is no options
+    // * `{ once: true | false, passive: true, capture: false }
```
buschtoens:
```suggestion
    // * `{ once: true | false, passive: true | false, capture: false }
```

packages/@ember/-internals/glimmer/lib/modifiers/on.ts
```diff
@@ -0,0 +1,241 @@
...
+    // `{ once: true | false, passive: true | false, capture: true }
+    //
+    // `once` is handled via a custom callback that removes after first
+    // invocation so we only care about capture here as a boolean
+    element.addEventListener(eventName, callback, true);
+  } else {
+    // used only in the following cases:
+    //
+    // * where there is no options
+    // * `{ once: true | false, passive: true, capture: false }
```
buschtoens:
```suggestion
    // * `{ once: true | false, passive: true | false, capture: false }
```

packages/@ember/-internals/glimmer/lib/modifiers/on.ts
```diff
@@ -0,0 +1,241 @@
...
+  constructor(element: Element, args: CapturedArguments) {
+    this.element = element;
+    this.args = args;
+    this.tag = args.tag;
+  }
+
+  updateFromArgs() {
+    let { args } = this;
+
+    let { once, passive, capture }: AddEventListenerOptions = args.named.value();
```
buschtoens: In `ember-on-modifier`, we assert that only these three options are passed to guard against typos. Do you think we should add this here as well?

packages/@ember/-internals/glimmer/lib/modifiers/on.ts
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    // use prior state values for removal
+    removeEventListener(element, eventName, callback, options);
+
+    // read updated values from the state object
+    addEventListener(state.element, state.eventName, state.callback, state.options);
+
+    state.shouldUpdate = false;
+  }
```
buschtoens: If I'm reading this correctly, once the modifier is installed, it will only ever update to a new listener, but will never be uninstalled. I can't think of a scenario, where the modifier would need to be uninstalled from an element that _does not_ get removed from the DOM afterwards, but I think @chancancode made a case for this once on Discord.

packages/@ember/-internals/glimmer/tests/integration/modifiers/on-test.js
```diff
@@ -0,0 +1,324 @@
...
+
+      '@test SUPPORTS_EVENT_OPTIONS is correct (private API usage)'(assert) {
+        let manager = this.getOnManagerInstance();
+
+        let { SUPPORTS_EVENT_OPTIONS } = manager;
+
+        if (isChrome || isFirefox) {
+          assert.strictEqual(SUPPORTS_EVENT_OPTIONS, true, 'is true in chrome and firefox');
+        } else if (isIE11) {
+          assert.strictEqual(SUPPORTS_EVENT_OPTIONS, false, 'is true in chrome and firefox');
```
buschtoens:
```suggestion
          assert.strictEqual(SUPPORTS_EVENT_OPTIONS, false, 'is false in IE11');
```

Or was this intentional to keep the logout output the same?

packages/@ember/-internals/glimmer/tests/integration/modifiers/on-test.js
```diff
@@ -0,0 +1,324 @@
...
+            },
+            handleInnerClick() {
+              assert.step('inner clicked');
+            },
+          }
+        );
+
+        runTask(() => this.$('.inner').click());
+
+        assert.verifySteps(['inner clicked', 'outer clicked'], 'uses capture: false by default');
```
buschtoens: I did not know about the [`verifySteps`](https://api.qunitjs.com/assert/verifySteps) API. Nice! 👏

packages/@ember/-internals/glimmer/lib/modifiers/on.ts
```diff
@@ -0,0 +1,241 @@
...
+    let options: AddEventListenerOptions;
+    if (once || passive || capture) {
+      options = this.options = { once, passive, capture };
+    } else {
+      this.options = undefined;
+    }
+
+    assert(
+      'You must pass a valid DOM event name as the first argument to the `on` modifier',
+      args.positional.at(0) !== undefined && typeof args.positional.at(0).value() === 'string'
```
buschtoens: Similar to asserting that only `once`, `passive` and `capture` are passed as named arguments, we could also check that only 2 positional arguments are passed.

buschtoens (approved):
Looks awesome! Thank you for getting this into core so swiftly. ⚡️

packages/@ember/-internals/glimmer/lib/modifiers/on.ts
chancancode: 👉 https://twitter.com/chancancode/status/1109502949972074496

packages/@ember/-internals/glimmer/lib/modifiers/on.ts
rwjblue: I was thinking that we didn't care, but honestly its a small change and reduces some overhead so seems good to do. Will update.

packages/@ember/-internals/glimmer/lib/modifiers/on.ts
rwjblue: Hmm, the RFC specifically states the opposite (that passing other things is "ignored"):

From [here](https://emberjs.github.io/rfcs/0471-on-modifier.html#named-parameters):

> The following named parameters will be accepted as options to {{on}}:
> 
> * capture
> * once
> * passive
> 
> These will be passed forward to addEventListener as options in modern browsers, and will be polyfilled in older browsers. Other options will be ignored.
> 

I don't feel super strongly so I just followed the RFC on this.

@pzuraq - Do you think we should change to assert?

packages/@ember/-internals/glimmer/lib/modifiers/on.ts
rwjblue: @buschtoens - We definitely cleanup (see `getDestructor` just below, and the `destroy` method on the `OnModifierState` object. I will also add a test (using the assert count methodology) to ensure we don't accidentally regress this.

buschtoens: Ah, I see! So `getDestructor` returns a `Destroyable`, which is implemented through `destroy()` in `OnModifierState`. I was not familiar with that part of the Glimmer API yet. 👍

packages/@ember/-internals/glimmer/tests/integration/modifiers/on-test.js
rwjblue: Nope, was a copy/paste mistake

packages/@ember/-internals/glimmer/lib/modifiers/on.ts
rwjblue: Ya, this is one way the "internal" modifiers differ from the public API. The public API _effectively_ does the same thing (the `destroy` method on the state bucket invokes the `destroyModifier` hook on the custom manager).

rwjblue:
FYI - Updated to address review feedback:

* Extract shared utility for a `this` context that errors when used.
* Assert when more positional arguments are provided.
* Avoid wrapping callback needlessly (makes debugging a bit easier,
  since the event handler function is actually the user-land function in
  most cases.
* Incorporates a system to error if any `this` properties are
  accessed from within the callback in debug builds.
* Fixup comments for addEventListener / removeEventListener logic.
* Fix test assertion message.
* Add test ensuring the modifier is destroyed properly.

I think this should be good to go once CI is green, in case anyone wants another crack at a review first...

packages/@ember/-internals/glimmer/lib/helpers/fn.ts
```diff
@@ -1,43 +1,11 @@
...
-      has(_target: {}, property: string | symbol) {
-        assertOnProperty(property);
-
-        return false;
-      },
-    }
-  );
-}
-
+const context = buildUntouchableThis('`fn` helper');
```
buschtoens: Can't touch this

packages/@ember/-internals/glimmer/lib/modifiers/on.ts
```diff
@@ -0,0 +1,262 @@
...
+          }
+
+          if (!SUPPORTS_EVENT_OPTIONS && once) {
+            removeEventListener(this, eventName, callback, options);
+          }
+          return userProvidedCallback.call(untouchableContext, event);
+        });
+      } else if (DEBUG) {
+        // prevent the callback from being bound to the element
+        this.callback = userProvidedCallback.bind(untouchableContext);
```
buschtoens: Nice! 👍

buschtoens (approved):

packages/@ember/-internals/glimmer/lib/helpers/fn.ts
NullVoxPopuli: yaaaaas

CvX:
@rwjblue The current implementation of `on` modifier isn't FastBoot ready, is it? It throws "TypeError: element.removeEventListener is not a function" where `{{action}}` didn't.

rwjblue:
@CvX - That is being tracked over in https://github.com/emberjs/ember.js/issues/17949.

PR #17961, opened 2019-04-23
[FEATURE] Ensure `fn` can be nested.

### LINKED ISSUE #17959: `fn` helper does not partially apply arguments
I’ve noticed that the new `fn` helper does partially apply the arguments of combined `fn` calls. I’ve set up a sandbox to demonstrate (be sure to open your browser console): https://codesandbox.io/s/wnkp6m7mww (It’s worth noting that the `ember-on-modifier` modifier addon is not working with this sandbox (unsure why), so I'm falling back to an onclick handler). I expected the handler to recieve arguments of both `fn` calls, similar to how the `action` handle applys arguments. Instead the second argument is undefined.

The RFC [was not updated to explicitly cover this use case](https://github.com/emberjs/rfcs/pull/470#discussion_r273167122) and states that `fn` is essentially just essentially calling `this.passedMethod.call(…arguments);`, which wouldn’t necessarily result in the argument application behavior that the `action` helper provides. It's possible that this is the intended functionality, however, after some Discord discussion, it seems that this may be a bug that warrants some investigation.

---

Here's a sandbox with a more realistic example (again, be sure to open your console): https://codesandbox.io/s/2x9wyyqjxj I’ve intentionally left the `Stars` component very generic. That is, I’m not passing a movie ID in as a component argument; this component could be used for rating something other than movies. Using the `action` helper, the consumer can pass down an action with additional arguments (the movie id) and the `Stars` component can utilize that action and pass along its own arguments. The `fn` helper does not function in the same way, and the second argument in the `handleRatingUpdate` action handler results in `undefined`.

rwjblue:
Fixes https://github.com/emberjs/ember.js/issues/17959

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/helpers/fn.ts
+++ b/packages/@ember/-internals/glimmer/lib/helpers/fn.ts
@@ -4,6 +4,7 @@ import { DEBUG } from '@glimmer/env';
 import { Arguments, VM } from '@glimmer/runtime';
 import { ICapturedArguments } from '@glimmer/runtime/dist/types/lib/vm/arguments';
 import { InternalHelperReference } from '../utils/references';
+import { Opaque } from '@glimmer/util';
 
 let context: any = null;
 if (DEBUG && HAS_NATIVE_PROXY) {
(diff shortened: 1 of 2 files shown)


rwjblue:
Thank you @locks

PR #17965, opened 2019-04-24
[FEATURE] Enable fn helper by default.

rwjblue:


packages/@ember/-internals/glimmer/lib/helpers/fn.ts
```diff
@@ -6,6 +6,52 @@ import { InternalHelperReference } from '../utils/references';
...
+  receive the arguments from the template along with the arguments passed when
+  invoking the returned function. Importantly, the `fn` helper does **not**
+  attempt to ensure any specific `this` context is present within the callback
+  when it is executed. This means that if your callback function needs access to
+  a specific `this` context, you should ensure that it is bound prior to the `fn`
+  invocation.
+
+  For example, given the following component:
+
+  ```app/components/calculator.js
```
pzuraq: Does this work? I've only seen examples setting the language

pzuraq (approved):
Overall looks good, just wanted to check on the one documentation question, but think this is ok to merge after that.

packages/@ember/-internals/glimmer/lib/helpers/fn.ts
rwjblue: I think it does, and there are some other examples in the codebase already.

@mansona / @toddjordan / @jenweber - Can y'all confirm?

jenweber: I don't think this works but I will have a look.

The Guides uses the following syntax but I don't think it's the same for API docs:
```javascript {data-filename=some/file.js}

Comment (jenweber) [reply packages/@ember/-internals/glimmer/lib/helpers/fn.ts]: TIL, it does work as @rwjblue has written:

Screen Shot 2019-04-24 at 5 17 39 PM

Screen Shot 2019-04-24 at 5 17 51 PM

File: packages/@ember/-internals/glimmer/lib/helpers/fn.ts [hunk 2]
```diff
@@ -6,6 +6,52 @@ import { InternalHelperReference } from '../utils/references';
 import buildUntouchableThis from '../utils/untouchable-this';
 
 const context = buildUntouchableThis('`fn` helper');
+
+/**
+@module ember
+*/
+
+/**
+  The `fn` helper is used to wrap the provided callback function up with a
```
jenweber: I think this could use a more generic, new-learner explanation of what to use it for. Normally I'd say that API docs are fine to be more technical, but this one is the "new way to do actions" and will therefore be higher traffic for new learners. Here's an example:

_The `fn` helper lets you call a function in your template and pass it some arguments. For example, someone might use it to register a click handler:_
_(show example)_
_The way the `fn` helper works is, it wraps the provided callback callback function up with a..._

packages/@ember/-internals/glimmer/lib/helpers/fn.ts
```diff
@@ -6,6 +6,52 @@ import { InternalHelperReference } from '../utils/references';
...
+  set of arguments. When the returned function is invoked the callback will
+  receive the arguments from the template along with the arguments passed when
+  invoking the returned function. Importantly, the `fn` helper does **not**
```
jenweber: "When the returned function is invoked the callback will receive the arguments from the template along with the arguments passed when invoking the returned function."

Can you state this another way, or in more words? I am not sure what this sentence means. It sounds like there are two sets of arguments that it gets, but I only see one arg in the example.

jenweber (changes requested):

packages/@ember/-internals/glimmer/lib/helpers/fn.ts
rwjblue: @jenweber I tweaked the docs significantly, mind re-reviewing?

rwjblue:
@jenweber I tweaked the docs significantly, mind re-reviewing?

packages/@ember/-internals/glimmer/lib/helpers/fn.ts
```diff
@@ -6,6 +6,77 @@ import { InternalHelperReference } from '../utils/references';
...
+/**
+@module ember
+*/
+
+/**
+  The `fn` helper allows you to ensure a function that you are passing off
+  to another component, helper, or modifier has access to arguments that are
+  available in the template.
+
+  For example, if you have an `each` loop over a number of items and need to
```
jenweber:
```suggestion
  For example, if you have an `each` helper loop over a number of items, you may need to
```

packages/@ember/-internals/glimmer/lib/helpers/fn.ts
```diff
@@ -6,6 +6,77 @@ import { InternalHelperReference } from '../utils/references';
...
+  pass a function to a component invoked within the loop but that function needs
```
jenweber:
```suggestion
  pass a function to a component invoked within the loop. That function needs
```

packages/@ember/-internals/glimmer/lib/helpers/fn.ts
```diff
@@ -6,6 +6,77 @@ import { InternalHelperReference } from '../utils/references';
...
+  access to the item.
```
jenweber:
```suggestion
  access to the item. Here's how you could use the `fn` helper to pass both the function and its arguments together to the component:
```

packages/@ember/-internals/glimmer/lib/helpers/fn.ts
```diff
@@ -6,6 +6,77 @@ import { InternalHelperReference } from '../utils/references';
...
+  import Component from '@glimmer/component';
+  import { action } from '@ember/object';
+
+  export default class ItemsList extends Component {
+    @action
+    handleSelected(item) {
+      // ...snip...
+    }
+  }
+  ```
```
jenweber: I think it's helpful (but not necessary) to show the template first, and then the component js.

packages/@ember/-internals/glimmer/lib/helpers/fn.ts
```diff
@@ -6,6 +6,77 @@ import { InternalHelperReference } from '../utils/references';
...
+
+  ```app/templates/components/items-listing.hbs
+  {{#each @items as |item|}}
+    <DisplayItem @item=item @select={{fn this.handleSelected item}} />
+  {{/each}}
+  ```
+
+  In this case the `display-item` component will receive a normal function
+  that it can invoke and when it invokes the function the `handleSelected`
```
jenweber:
```suggestion
  that it can invoke. When it invokes the function, the `handleSelected`
```

packages/@ember/-internals/glimmer/lib/helpers/fn.ts
```diff
@@ -6,6 +6,77 @@ import { InternalHelperReference } from '../utils/references';
...
+  function will receive the `item` (from the `fn` invocation in the loop)
+  and any arguments passed when `display-item` invokes the function.
```
jenweber:
```suggestion
  and any arguments passed, thanks to the `fn` helper.
```

packages/@ember/-internals/glimmer/lib/helpers/fn.ts
```diff
@@ -6,6 +6,77 @@ import { InternalHelperReference } from '../utils/references';
...
+
+  ```app/templates/components/items-listing.hbs
+  {{#each @items as |item|}}
+    <DisplayItem @item=item @select={{fn this.handleSelected item}} />
+  {{/each}}
+  ```
+
+  In this case the `display-item` component will receive a normal function
+  that it can invoke and when it invokes the function the `handleSelected`
+  function will receive the `item` (from the `fn` invocation in the loop)
```
jenweber:
```suggestion
  function will receive the `item`
```

packages/@ember/-internals/glimmer/lib/helpers/fn.ts
```diff
@@ -6,6 +6,77 @@ import { InternalHelperReference } from '../utils/references';
...
+  Let's take look at what that means in a couple circumstances:
+
+  - When invoked as `this.args.select()` the `handleSelected` function will
+    receive the `item` from the loop as its first and only argument.
+  - When invoked as `this.args.selected('foo')` the `handleSelected` function
+    will receive the `item` from the loop as its first argument and the
+    string `'foo'` as its second argument.
+
+  In the example above, we used `@action` to ensure that `handleSelected` is
+  properly bound to the `items-list`, but lets explore what happens if we didn't:
```
jenweber:
```suggestion
  properly bound to the `items-list`, but let's explore what happens if we left out `@action`:
```

jenweber (commented):
Here are some small wording updates.

rwjblue:
@jenweber - Awesome, thank you! Updated with your suggestions...

rwjblue:
Chatted with @jenweber in #dev-ember-learning and she is 👍 with the last round of changes.

PR #17971, opened 2019-04-25
[BUGFIX release] Ensure query param only link-to's work in error states.

### LINKED ISSUE #17963: `<LinkTo>` without `@route` param but with `@query` breaks when navigating to error route
While working on https://github.com/ember-engines/ember-engines/pull/642 to make ember-engines compatible with Ember Octane, I noticed that https://github.com/emberjs/ember.js/pull/17772 appears to have broken an edge case for `<LinkTo>` without a `@route` argument, but with a `@query` argument.

I made a reproduction here: https://codesandbox.io/s/p32ozvvz0x

If no explicit `@route` is passed it will be set to (or remain) `UNDEFINED`:

https://github.com/emberjs/ember.js/blob/2cc35738f2938e6b9735d0d641214b0721d53a4b/packages/%40ember/-internals/glimmer/lib/components/link-to.ts#L895-L900

In this case `_route` returns the currently active route:

https://github.com/emberjs/ember.js/blob/2cc35738f2938e6b9735d0d641214b0721d53a4b/packages/%40ember/-internals/glimmer/lib/components/link-to.ts#L490-L493

This then breaks the `href` computed property in L791, if the router performs an intermediate transition to an `error` route:

https://github.com/emberjs/ember.js/blob/2cc35738f2938e6b9735d0d641214b0721d53a4b/packages/%40ember/-internals/glimmer/lib/components/link-to.ts#L749-L804

`routing.generateURL(route, models, query)` is called as:

```ts
routing.generateURL("blog.post.error", [], { lang: "Japanese" });
```

Which misses the `error` param:

```
Error: You must provide param `error` to `generate`.
```

This apparently used to work and while being a weird edge-case, I would still expect it to work.

2hu12:
Try to fix https://github.com/emberjs/ember.js/issues/17963, I'm not sure this is the right way and it's weird, but the old "qualifiedRouteName" property works in this way and "_route" is used similarly to "qualifiedRouteName".

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/components/link-to.ts
+++ b/packages/@ember/-internals/glimmer/lib/components/link-to.ts
@@ -487,7 +487,9 @@ if (EMBER_GLIMMER_ANGLE_BRACKET_BUILT_INS) {
     _currentRouterState: alias('_routing.currentState'),
     _targetRouterState: alias('_routing.targetState'),
 
-    _route: computed('route', '_currentRoute', function computeLinkToComponentRoute(this: any) {
+    _route: computed('route', '_currentRouterState', function computeLinkToComponentRoute(
+      this: any
+    ) {
       let { route } = this;
       return route === UNDEFINED ? this._currentRoute : route;
     }),
--- a/packages/@ember/-internals/glimmer/tests/integration/components/link-to/query-params-angle-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/components/link-to/query-params-angle-test.js
@@ -240,6 +240,77 @@ if (EMBER_GLIMMER_ANGLE_BRACKET_BUILT_INS) {
         });
       }
 
+      ['@test generates proper href for `LinkTo` with no @route after transitioning to an error route GH#17963'](
+        assert
+      ) {
+        this.router.map(function() {
+          this.route('bad');
+        });
+
+        this.add(
+          'controller:application',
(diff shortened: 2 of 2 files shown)


rwjblue (commented):
The fix seems reasonable (thank you!), but I think we need to add a previously failing test to ensure we don't regress again...

2hu12:
I try to add one test in "packages/@ember/-internals/glimmer/tests/integration/components/link-to/query-params-curly-test.js", but with no success. The test need an `{{#link-to}}` with no route parameter and the bug could be reproduced when transition to an error route which might have a rejected model, but when it encounters an error like `model() { RSVP.reject(new Error('some error')) }` in test, the whole test just terminated.

Do you have any idea on how to write this previously failing test? Thank you!

villander:
@rwjblue @dgeb  any idea about this? We are with ember-engines CI broken.

I'll be happy if you can add any tip to finish this, thanks

villander:
we are trying fix this on engines side as well - https://github.com/ember-engines/ember-engines/pull/647

villander:
@2hu12 unfortunately I don't was able to write tests for this  =/

rwjblue:
I just rebased and pushed a previously failing test (using the one @dgeb started over in #18061), once CI is green we should be good to go.

villander:
thanks @rwjblue <3

PR #17972, opened 2019-04-25
Prevent generation of .js.js or .hbs.hbs files with ember-cli

romgere:
https://github.com/ember-cli/ember-cli/issues/7699 Add/modify `blueprint.normalizeEntityName` to prevent creation of `.js.js`, `.js.hbs` or `.hbs.hbs` files when generate component, controller, helper, mixin, route, service, template and util.
Hope this is the right way to do it :thinking: ...

### DIFF
--- a/blueprints/component/index.js
+++ b/blueprints/component/index.js
@@ -72,7 +72,9 @@ module.exports = useEditionDetector({
   },
 
   normalizeEntityName: function(entityName) {
-    return normalizeEntityName(entityName);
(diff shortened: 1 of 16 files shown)


rwjblue (commented):
Seems ok to me, @Turbo87 what do you think?

Turbo87 (approved):

rwjblue:
Thank you @romgere!

PR #17973, opened 2019-04-25
[BUGFIX #17529] Refactor get to return object or path

eplata31:
I tracked issue [#17529](https://github.com/emberjs/ember.js/issues/17529) and as suggested on the thread, it worked on the get property.
I wrote a new test based on the one from jasonmit and got it to pass, also all other test passed
Screenshot from 2019-04-25 11-44-41

Hope this will help, but any comment, suggestion or advice is welcome with open arms.
I am kind of new in the open source world so I hope I follow all of your contributing guides well but if I didn't it will be awesome if you could point out what I missed. Thanks :D

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/syntax/each-in-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/syntax/each-in-test.js
@@ -297,6 +297,14 @@ class EachInTest extends AbstractEachInTest {
 
     this.assertText('Empty!');
   }
+
+  [`@test it supports keys with a period in them`]() {
+    this.makeHash({ 'key.period': 'val' });
+
+    this.render(`<ul>{{#each-in hash as |key value|}}<li>{{key}}: {{value}}</li>{{/each-in}}</ul>`);
+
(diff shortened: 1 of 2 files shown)


pzuraq:
I'm not sure we should incur extra overhead for this, I think supporting periods in keynames was actually a bug and wasn't really intended behavior.

jasonmit:
> I'm not sure we should incur extra overhead for this, I think supporting periods in keynames was actually a bug and wasn't really intended behavior.

In dev mode, might be useful then to throw with a helpful error.  People are running into this while upgrading and it's not immediately clear what broke outside of nothing rendering.

eplata31:
I agree with @jasonmit and also it will be good to close the related issues to avoid confusion and extra work.

devop911:
@pzuraq dot notation objects are fairly Common and i dont see a reason to let ember Break on a valid JSON

locks:
Sorry for the long turnaround! I think the ship has sailed on this 😅 We are trying to move off of EmberObject overall, too.

PR #17978, opened 2019-04-26
Make hasListeners public

richgt:
Allow for determining whether listeners exist on an object before calling removeListeners.
Method already exists, just making it public.


rwjblue:
Thanks for submitting! I've added the needs team response label, and will try to bring this up at the next ember core team meeting.

PR #17979, opened 2019-04-26
[BUGFIX] Don't get values for leaf dependencies

pzuraq:
We don't need to actually get the value of a leaf dependency, only its
tag, so we can stop early.

Also adds a test for native computed setters.

### DIFF
--- a/packages/@ember/-internals/metal/lib/chain-tags.ts
+++ b/packages/@ember/-internals/metal/lib/chain-tags.ts
@@ -81,6 +81,10 @@ export function getChainTagsForKey(obj: any, key: string) {
 
     chainTags.push(propertyTag);
 
+    if (segments.length === 0) {
+      break;
+    }
+
     descriptor = descriptorForProperty(current, segment);
 
     if (descriptor === undefined) {
@@ -103,7 +107,7 @@ export function getChainTagsForKey(obj: any, key: string) {
         } else {
           current = peekCacheFor(current).get(segment);
         }
-      } else if (segments.length > 0) {
+      } else {
         let placeholderTag = UpdatableTag.create(CONSTANT_TAG);
 
         metaFor(current)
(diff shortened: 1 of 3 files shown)


rwjblue:
One failure:

```
Test failed: computed - dependentkey:  circular keys should not blow up", source:  (49)
[0427/004834.417161:INFO:CONSOLE(52)] "    Failed assertion: should recache, expected: foo 3, but was: foo 1
    at Object.@test circular keys should not blow up (http://localhost:13141/ember-tests.js:41770:14)
    at Object.<anonymous> (http://localhost:13141/ember-tests.js:84067:37)
    at runTest (http://localhost:13141/qunit/qunit.js:3044:30)
    at Test.run (http://localhost:13141/qunit/qunit.js:3030:6)
    at http://localhost:13141/qunit/qunit.js:3257:12
    at processTaskQueue (http://localhost:13141/qunit/qunit.js:2623:24)", source:  (52)
```

pzuraq:
I started digging into this again because I thought it would be an easy fix, but after getting context back I realized what the issue is and it's probably not as easy as hoped. Going to record the issue for now, come back to it later.

Essentially, the issue is that when you have a cycle in Updatable tags, one tag will end up eagerly completing as the other is being calculated, and will be incorrect. This situation can really only occur with computed properties. The solution is, if an UpdatableTag encounters another tag that is _mid-update_, it returns the current value, but does _not_ save its `lastChecked` value so it isn't cached.

The only place to add this would either be within `UpdatableTag` itself, or in the combination `UpdatableDirtyableTag` that effectively exists right now. That class _also_ exists in more recent versions of Glimmer VM, along with a lot of other benefits, so it probably makes sense to update the VM first before we attempt to address this.

rwjblue:
FYI - @pzuraq submitted https://github.com/glimmerjs/glimmer-vm/pull/937 a while back to address this.

PR #18035, opened 2019-05-17
Move render options off the route instances

chadhietala:


locks (approved):

rwjblue (approved):

mehulkar:
Hey @chadhietala, we had been overriding the `enter` hook of routes to delay rendering and doing something like this:

```js
let foo;
enter() {
  this.connections = [];
  foo = () => {
     this.activate();
     this.trigger('activate');
  }
}

// and then at some later point call `foo()` to activate the route.
```

It doesn't look like [`ROUTE_CONNECTIONS` is exported](https://github.com/emberjs/ember.js/blob/master/packages/@ember/-internals/routing/index.ts) to the end user, but is there some private way to get access to it? Or any other suggestions you might have?

mehulkar:
Looks like [`enter` is private API](https://api.emberjs.com/ember/3.12/classes/Route/methods/enter?anchor=enter&show=private), so our bad for using it, but wondering if there's anything else we can do short of using our own `WeakMap` and overriding every hook that uses it.

rwjblue:
I don’t really understand the actual thing that you are trying to do with that code 🤔

mehulkar:
@rwjblue, it’s a bit confusing to me too, but the high level is that we’re wrapping outlet in a component that reads/understands outletState to control route transitions. Capturing the work in `enter` in a callback let’s us activate later.

rwjblue:
Ya, I’m just not sure that’s a thing that we’ve ever actually supported. 🤔

snewcomer:
Being able to override the `enter` and `exit` Route methods provides a declarative solution to animate in/out a route by controlling the timing of transitions.  One can effectively buffer `this.activate()` and `this.trigger('activate')`, allowing you to do something like a nice loading screen b/w route transitions with css animations as it `exits a route > enters a loading state > and enters new route`.  This fine grained control can happen in a custom outlet similar to this in [ember-a11y](https://github.com/ember-a11y/ember-a11y/blob/master/addon/components/focusing-inner.js#L17).  

As for a potential path forward, perhaps `routeWillChange` and/or `routeDidChange` will work since these get called with intermediate `loading` transitions.

snewcomer:
Actually I just realized we can "buffer" (client implementation) calling `super()` until the right moment and keep what was done in this PR.  Ignore me!

snewcomer:
I lied.  B/c we don't have access to `ROUTE_CONNECTIONS`, delaying `activate` in enter/exit to provide a declarative way of managing route transitions (say a fade in fade out) isn't possible.  We would need this as a property on the route to do so, which also seems it won't be supported.

snewcomer:
I'm going to back up and say nothing has to be done.  [`activate`](https://github.com/emberjs/ember.js/blob/v3.13.1/packages/%40ember/-internals/routing/lib/system/route.ts#L595) is a noop hook by default and whether this was needed in the past or not (or pertinent to a user's codebase), it is not needed now or code just needs to be shuffled around.  Ignore my previous messages!

PR #18042, opened 2019-05-23
[Docs] Fix API docs suggest import path of deprecate

### LINKED ISSUE #17185: [docs] API docs should suggest deprecate be imported from @ember/debug

xg-wang:
`import { deprecate } from '@ember/debug'` will strip code in production, this is not true for `@ember/applications/deprecations`

Fix https://github.com/emberjs/ember.js/issues/17185

### DIFF
--- a/packages/@ember/debug/index.ts
+++ b/packages/@ember/debug/index.ts
@@ -211,7 +211,7 @@ if (DEBUG) {
   });
 
   /**
-   @module @ember/application
+   @module @ember/debug
    @public
   */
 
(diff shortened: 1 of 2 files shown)


rwjblue:
This seems like a duplicate of https://github.com/emberjs/ember.js/pull/17926, would you mind reviewing and checking in with @elwayman02 to help push it forward?

xg-wang:
Ah sorry it's been a while I forgot that PR.. closing this

PR #18059, opened 2019-05-29
[FEATURE] Adds sync observers back to tracked properties

pzuraq:
This PR flushes synchronous observers in the tracked feature flag path
during each `notifyPropertyChange`, which gives these observers similar
timing semantics to what they currently have without the feature flag
enabled. These changes are meant to only affect observers, and while the
metadata required is stored with listeners, this PR doesn't add "async
listeners" as a concept to the system.

The most notable change is the addition of the `suspended` property to
active observers. This property is used by setters which are being set
(computed property setters, and in the future `dependentKeyCompat`
setters), and when observers are triggered to prevent re-entry.


packages/@ember/-internals/metal/lib/property_set.ts
```diff
@@ -12,7 +12,7 @@ import { DEBUG } from '@glimmer/env';
 import { descriptorForProperty } from './descriptor_map';
 import { isPath } from './path_cache';
 import { MandatorySetterFunction } from './properties';
-import { notifyPropertyChange } from './property_events';
+import { beginPropertyChanges, endPropertyChanges, notifyPropertyChange } from './property_events';
```
rwjblue:
```
$ tsc --noEmit
packages/@ember/-internals/metal/lib/property_set.ts(15,10): error TS6133: 'beginPropertyChanges' is declared but its value is never read.
packages/@ember/-internals/metal/lib/property_set.ts(15,32): error TS6133: 'endPropertyChanges' is declared but its value is never read.
```

packages/@ember/-internals/environment/lib/env.ts
```diff
@@ -112,6 +112,20 @@ export const ENV = {
   */
   _JQUERY_INTEGRATION: true,
 
+  /**
+    Whether the app defaults to using async observers.
+
+    This is not intended to be set directly, as the implementation may change in
+    the future. Use `@ember/optional-features` instead.
+
+    @property _JQUERY_INTEGRATION
```
josemarluedke:
```suggestion
    @property _DEFAULT_ASYNC_OBSERVERS
```

rwjblue (approved):

PR #18061, opened 2019-05-29
WIP: Failing test for #17963

dgeb:
Attempt to reproduce failure in #17963: `<LinkTo>` without `@route` param but with `@query` breaks when navigating to error route


dgeb:
This test is not currently failing, so it may not yet be accurately reproducing #17963.

/cc @buschtoens @rwjblue

rwjblue:
I fixed up the test (so that it was properly failing), and pushed it to @2hu12's branch for #17963. Once that passes CI, we should be good to go.

PR #18064, opened 2019-05-30
[BUGFIX lts] Fix 'hasAttribute is not a function' when jQuery is disabled

### LINKED ISSUE #16730: hasAttribute is not a function
We're getting a lot of errors from `packages/ember-views/lib/system/event_dispatcher.js` in our application.  It stems from a call to `target.hasAttribute('data-ember-action')` because hasAttribute is not defined on target. I haven't been able to identify what event is causing it, but it seems it is likely something where window is the target.

I'll submit a PR if you want to let me know how you want to guard against this. Seems like changing the line to `else if (target && target.hasAttribute && target.hasAttribute('data-ember-action'))` would work. I'm assuming there is a reason its a do-while loop and not while, otherwise converting to a while loop should also prevent it.

We're running Ember 3.1.2.  Also just fyi, we were previously receiving the same error on Ember 2.x when running ember-native-dom-event-dispatcher.  I previously reported the issue in that [repo.](https://github.com/rwjblue/ember-native-dom-event-dispatcher/issues/22)

[Source Code](https://github.com/emberjs/ember.js/blob/v3.2.0/packages/ember-views/lib/system/event_dispatcher.js#L17)
```js
      let handleEvent = (this._eventHandlers[event] = event => {
        let target = event.target;

        do {
          if (viewRegistry[target.id]) {
            if (viewHandler(target, event) === false) {
              event.preventDefault();
              event.stopPropagation();
              break;
            }
          } else if (target.hasAttribute('data-ember-action')) {
            if (actionHandler(target, event) === false) {
              break;
            }
          }

          target = target.parentNode;
        } while (target && target.nodeType === 1);
      });

      rootElement.addEventListener(event, handleEvent);
```

josemarluedke:
Fixes #16730.

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/event-dispatcher-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/event-dispatcher-test.js
@@ -555,6 +555,29 @@ if (jQueryDisabled) {
         assert.ok(receivedEvent, 'click event was triggered');
         assert.notOk(receivedEvent.originalEvent, 'event is not a jQuery.Event');
       }
+
+      ['@test native event on text node does not throw on hasAttribute [ISSUE #16730]'](assert) {
+        this.registerComponent('x-foo', {
+          ComponentClass: Component.extend({
+            actions: {
+              someAction() {},
+            },
+          }),
+          template: `<a id="inner" href="#" {{action 'someAction'}}>test</a>`,
+        });
+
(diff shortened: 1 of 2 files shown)


rwjblue:
Thank you!

rwjblue:
Backported to be included in 3.8.2 and 3.10.1 (should be published shortly).

PR #18066, opened 2019-05-30
[FEATURE EMBER_MODULE_UNIFICATION] Disable.

rwjblue:
It is extremely unlikely that the implementation of module unification that exists behind this feature flag will ever land, and due to the way we test optional features the presence of this feature flag makes it impossible to test other new features _without_ also testing module unification (meaning the features can't be tested independently).

This disables the module unification feature completely such that it is not possible to enable at runtime.


chancancode (approved):

locks (approved):

NullVoxPopuli:
😢

PR #18067, opened 2019-05-31
Fix links in documentation

mixonic:
Documentation links were pointing at `/api/`, when the Ember documentation does not live at https://api.emberjs.com under that path. Thus all these links have been broken.

For example on [`@ember/component`](https://api.emberjs.com/ember/release/modules/@ember%2Fcomponent) the link to "Ember.Templates.helpers" is dead.

And correct a few broken for other reasons.

---

This fixes the links in the *current release*, but these links *remain broken as that release becomes supplanted by a newer one*. This is because these links are to `/ember/release/`. So if you are on the 3.10 docs after 3.11 has been released, clicking these links would take you to the *incorrect* spot, 3.11.

And additionally *many, many old links are broken*. For example on [`ember-templates` for 1.13](https://api.emberjs.com/ember/1.13/modules/ember-templates) that same link to "Ember.Templates.helpers" is broken. I think we are unlikely to backport this work to every release of Ember, so it should probably be resolved in the API docs app.

/cc @mansona @emberjs/learning-team-managers @rwjblue

### DIFF
--- a/packages/@ember/-internals/glimmer/index.ts
+++ b/packages/@ember/-internals/glimmer/index.ts
@@ -63,14 +63,16 @@
   )}}></span>
   ```
 
-  Ember's built-in helpers are described under the [Ember.Templates.helpers](/api/ember/release/classes/Ember.Templates.helpers)
+  Ember's built-in helpers are described under the [Ember.Templates.helpers](/ember/release/classes/Ember.Templates.helpers)
   namespace. Documentation on creating custom helpers can be found under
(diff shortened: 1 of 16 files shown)


jenweber:
In light of discussion in the learning team chat, I suggest we merge this and then fix the rest in an API docs post-processing step that @sivakumar-kailasam is working on. Thanks for your work!

jenweber (approved):

rwjblue (approved):

PR #18070, opened 2019-06-01
[DOC release-3-10] Replace @ in decorators code samples with &#64;

jenweber:
This is a patch because YUIdoc can't parse the @ of decorators in code blocks, and it broke the API docs. See #18063 for examples of the broken state.

Really we need to fix this in YUIdocs but that's a long haul.

I'm not sure I got them all. I had a hard time composing regex that would work in the command line or VSCode. If anyone can run this kind of search, it would be nice: `(?<!'|backtick-goes-here|/|\* )@`. Translation - find an @ symbol that is not preceded by a single quote, backtick, slash, or `* ` aka a bullet point used by YUIdoc labels like `@method`. Most command line things couldn't deal with the ' and VSCode couldn't deal with the `\* ` in the negative lookbehind.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/components/textarea.ts
+++ b/packages/@ember/-internals/glimmer/lib/components/textarea.ts
@@ -35,7 +35,7 @@ import layout from '../templates/empty';
   import { tracked } from '@glimmer/tracking';
 
   export default class extends Component {
-    @tracked writtenWords = "Lots of text that IS bound";
+    &#64;tracked writtenWords = "Lots of text that IS bound";
   }
   ```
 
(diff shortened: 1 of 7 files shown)


jenweber:
cc @toddjordan @sivakumar-kailasam

sivakumar-kailasam:
Thanks for the PR Jen. Let's hold off on merging this until later today. I'm going to try fixing this at the jsonapi docs so that we don't have to do this.

sivakumar-kailasam:
Closing in favor of https://github.com/ember-learn/ember-jsonapi-docs/pull/64

PR #18071, opened 2019-06-03
[BUGFIX lts + beta] Ensure modifiers do not run in FastBoot modes.

### LINKED ISSUE #17949: destroyModifier is called in SSR mode

rwjblue:
Fixes https://github.com/emberjs/ember.js/issues/17949

Note: the fix is split into two different commits to make backporting easier


CvX:
I've tried it out in my production app and it does indeed work as intended! 🚀

PR #18074, opened 2019-06-05
Autotracking Bugfix: Make the model property on the controller tracked

NullVoxPopuli:
~TODO:~
 - ~write failing test~
 - ~add tracked when auto-tracking is enabled~

when the tracked flag is enabled, the controller's model property will be tracked by default.

### DIFF
--- a/packages/ember/tests/routing/route_controller_integration_test.js
+++ b/packages/ember/tests/routing/route_controller_integration_test.js
@@ -1,0 +1,56 @@
+import {
+  EMBER_NATIVE_DECORATOR_SUPPORT,
+  EMBER_METAL_TRACKED_PROPERTIES,
+} from '@ember/canary-features';
+import { Route } from '@ember/-internals/routing';
+import Controller from '@ember/controller';
+import { moduleFor, ApplicationTestCase } from 'internal-test-helpers';
+
+if (EMBER_METAL_TRACKED_PROPERTIES && EMBER_NATIVE_DECORATOR_SUPPORT) {
+  moduleFor(
+    'Route <-> Controller Integration',
+    class extends ApplicationTestCase {
+      ['@test properties that autotrack the model update when the model changes'](assert) {
+        assert.expect(2);
+
+        this.router.map(function() {
+          this.route('home', { path: '/home/:id' });
+        });
+
+        class HomeRoute extends Route {
+          async model({ id }) {
+            return { value: id };
+          }
+        }
+
+        class HomeController extends Controller {
+          get derivedProperty() {
+            return this.model.value || 'value is unset';
+          }
+        }
+
+        this.add('route:home', HomeRoute);
+        this.add('controller:home', HomeController);
+        this.addTemplate('home', '<h3 class="derivedProperty">{{this.derivedProperty}}</h3>');
+
+        return this.visit('/home/2')
+          .then(() => {
+            assert.equal(
+              document.querySelector('h3').innerText,
+              '2',
+              'the derived property matches the id'
+            );
+          })
+          .then(() => {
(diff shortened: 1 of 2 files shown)


rwjblue (approved):
Seems good to me. 

@pzuraq - Any objections?

pzuraq (approved):

packages/ember/tests/routing/route_controller_integration_test.js
```diff
@@ -0,0 +1,44 @@
...
+
+        this.add('route:home', HomeRoute);
+        this.add('controller:home', HomeController);
+        this.addTemplate('home', '<h3 class="derivedProperty">{{this.derivedProperty}}</h3>');
+
+        await this.visit('/home/2');
+        assert.equal(this.$('h3').text(), '2', 'the derived property matches the id');
+
+        await this.visit('/home/3');
+        assert.equal(this.$('h3').text(), '3', 'the derived property matches the id');
```
lifeart: do we need jquery selection?

NullVoxPopuli: probably not, but I saw other tests using it 🤷‍♂

lifeart: should it fail if jquery integration = false?

NullVoxPopuli: removed -- just using document.querySelector :-\

rwjblue: Ya, it’s an internal abstraction that we did during the “drip jQuery requirement” work

NullVoxPopuli:
C.I. error:
```
unknown Statement of type "ClassDeclaration"
```
:(

rwjblue:
I'm not sure exactly why this is an issue, but regenerator-runtime is throwing that error. Specifically, it doesn't like the inline class declarations within the test function. 

There are two "easy" choices:

* Rewrite to avoid using async / await -- This will almost certainly work, and should be _fairly_ simple.
* Rewrite `class Foo extends Controller {}` to `let Foo = class Foo extends Controller {}`-- I'm not 100% sure this will solve the issue, but it might.

NullVoxPopuli:
> Rewrite to avoid using async / await -- This will almost certainly work, and should be fairly simple.

is it worth configuring the tests to support async/await?

rwjblue:
They do support async / await in general, @pzuraq recently added that. The specific issue in this case is that (apparently) regenerator-runtime doesn't know what to do with a `ClassDeclaration`. The version in use, is older (0.11.1) so maybe it's fixed upstream already. I dunno.

rwjblue (approved):

rwjblue:
@NullVoxPopuli 	- Can you squash?

NullVoxPopuli:
squashed

NullVoxPopuli:
idk what's up with the git history here...

CHANGELOG.md
```diff
@@ -16,6 +16,10 @@
 - [#17940](https://github.com/emberjs/ember.js/pull/17940) [CLEANUP] Remove `sync` queue from @ember/runloop.
 - [#18026](https://github.com/emberjs/ember.js/pull/18026) Enabling featured discussed in 2019-05-03 core team meeting.
 
+### v3.10.1 (June 4, 2019)
```
NullVoxPopuli: where did this come from O.o

NullVoxPopuli:
@rwjblue fixed history :tm: :)

NullVoxPopuli:
looks like browserstack had a problem

NullVoxPopuli:
can someone click re-run?

Issue #18060, opened 2019-05-29
[CLEANUP] remove enabled feature flags on release


chancancode:
## PRs

- EMBER_GLIMMER_IN_ELEMENT https://github.com/emberjs/ember.js/pull/19196
- EMBER_DESTROYABLES https://github.com/emberjs/ember.js/pull/19203
- EMBER_CACHE_API https://github.com/emberjs/ember.js/pull/19204
- EMBER_ROUTING_MODEL_ARG https://github.com/emberjs/ember.js/pull/19206 
- EMBER_GLIMMER_INVOKE_HELPER https://github.com/emberjs/ember.js/pull/19207
- EMBER_GLIMMER_HELPER_MANAGER https://github.com/emberjs/ember.js/pull/19208
- EMBER_GLIMMER_SET_COMPONENT_TEMPLATE https://github.com/emberjs/ember.js/pull/19229

mul53:
@chancancode can i have a crack at this?

rajankumar549:
Hey, can i help to fix this issue. I want to contribute in this project. @chancancode

chancancode:
@mul53 @rajankumar549 you don't need permission, just go ahead and open PR. Preferably one feature flag at a time and maybe add a note here on what you are working on so others can pick something else.

neonick:
@chancancode seems it could be closed?

praaatik:
Is the issue still open? Seems to be merged

praaatik:
Hello @chancancode, I guess this issue could be closed now? Seems to be merged

kanavpruthi:
I am a beginner, and id like to work on a new issue here

chancancode:
@kanavpruthi the specific set of enabled features to remove has changed since this was opened, but there are now new ones that needed to be cleaned up https://github.com/emberjs/ember.js/blob/master/packages/@ember/canary-features/index.ts

Alpine-lines:
Should all of the features in /packages/@ember/canary-features/index.ts have their respective feature flags removed?

pzuraq:
@Alpine-lines any flag that has been set to `true`, and has been released in a minor version of Ember, should be removed. Currently, these ones: https://github.com/emberjs/ember.js/blob/master/packages/@ember/canary-features/index.ts#L19-L23

pzuraq:
The best way to approach this is with one PR per feature flag, to make it easier to review quickly

SOUMOJIT-CHAKRABORTY:
[...]

PR #18078, opened 2019-06-07
[FEATURE ember-glimmer-on-modifier] Remove feature flagging.

josemarluedke:
This helps with #18060.


rwjblue:
PR looks good, but we need to wait to land until 3.11.0 is released. We don't want to remove the feature flagging before the feature is released in a stable version, just in case we have to disable at the last minute (due to some regression or whatever).

josemarluedke:
@rwjblue Ahh, I thought {{on}} modifier was released in 3.10. My bad. We can hold it here for now.

rwjblue:
No worries! Thank you for doing the work!

PR #18091, opened 2019-06-12
[BUGFIX] Use WeakMap to store tracked property values.

### LINKED ISSUE #18075: [IE11] @tracked property initializer is not applied
In IE11 the property initializers for the `@tracked` decorator are not applied. The following test succeeds in Chrome, but fails in IE11.

[`tests/unit/tracked-test.js`](https://github.com/buschtoens/repro-tracked-ie-error/blob/master/tests/unit/tracked-test.js)

```js
import { setupTest } from 'ember-qunit';

module('Unit | @tracked', function(hooks) {
  setupTest(hooks);

  test('it applys the initializer', function(assert) {
    class Subject {
      @tracked
      property = 'foo';
    }

    const subject = new Subject();

    assert.strictEqual(subject.property, 'foo');
  });
});
```

Reproduction repo here: [`buschtoens/repro-tracked-ie-error`](https://github.com/buschtoens/repro-tracked-ie-error/blob/master/tests/unit/tracked-test.js)

rwjblue:
Using a `Symbol` causes issues on IE11 when using the core-js polyfill. Specifically, the polyfill adds every symbol to `Object.prototype` and therefore our `if (symbol in this) {` check was completely borked.

Migrating to a `WeakMap` exposed another issue: we were accidentally relying on the fact that the symbols would be inherited through the prototype chain (introduced by #18074). This meant that setting a tracked property on a prototype would actually affect the value seen by every instance of that prototype.

For example:

```js
class Thing {
  @Tracked baz;
}

Thing.prototype.baz = 'foo';

let thing = new Thing();

console.log(thing.baz)
```

Before the changes in this commit, the above `console.log` would emit `foo` and after these changes it would print `undefined`.

Fixes #18075

### DIFF
--- a/packages/@ember/-internals/metal/lib/tracked.ts
+++ b/packages/@ember/-internals/metal/lib/tracked.ts
@@ -1,4 +1,4 @@
-import { HAS_NATIVE_SYMBOL, isEmberArray, symbol as emberSymbol } from '@ember/-internals/utils';
+import { isEmberArray } from '@ember/-internals/utils';
 import { EMBER_NATIVE_DECORATOR_SUPPORT } from '@ember/canary-features';
 import { assert } from '@ember/debug';
 import { DEBUG } from '@glimmer/env';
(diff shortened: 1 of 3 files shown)


pzuraq (approved):
Looks good! I'm not sure what the test failure is exactly, the error message doesn't seem to say anything useful unfortunately 😕

PR #18096, opened 2019-06-13
Refactor template factory

chancancode:
Per [RFC #481](https://github.com/emberjs/rfcs/pull/481), we want to align the factory signature between templates and component managers. Since the factories for component managers are just a function that takes the owner, we decided to move template factories in the same direction here as well.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/setup-registry.ts
+++ b/packages/@ember/-internals/glimmer/lib/setup-registry.ts
@@ -55,7 +55,7 @@ export function setupApplicationRegistry(registry: Registry) {
   registry.injection('service:-dom-builder', 'bootOptions', '-environment:main');
   registry.injection('renderer', 'builder', 'service:-dom-builder');
 
-  registry.register(P`template:-root`, RootTemplate);
+  registry.register(P`template:-root`, RootTemplate as any);
   registry.injection('renderer', 'rootTemplate', P`template:-root`);
 
   registry.register('renderer:-dom', InteractiveRenderer);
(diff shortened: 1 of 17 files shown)


rwjblue:
Failed on linting:

```
$ tsc --noEmit
packages/@ember/-internals/glimmer/lib/component-managers/root.ts(27,22): error TS2554: Expected 1 arguments, but got 2.
```

Need to run `yarn lint:fix` I believe.

rwjblue:
I just rebased and fixed the linting failure.

@pzuraq / @krisselden - I'd love review here when y'all have a minute.

rwjblue:
CI is green now, still needs some review though...

Ping again @krisselden / @pzuraq 😸

packages/@ember/-internals/glimmer/lib/setup-registry.ts
```diff
@@ -80,22 +80,22 @@ export function setupApplicationRegistry(registry: Registry) {
 }
 
 export function setupEngineRegistry(registry: Registry) {
+  registry.optionsForType('template', { instantiate: false });
+
   registry.register('view:-outlet', OutletView);
-  registry.register('template:-outlet', OutletTemplate);
+  registry.register('template:-outlet', OutletTemplate as any);
```
krisselden: why does this need `as any`?

packages/@ember/-internals/glimmer/tests/unit/runtime-resolver-cache-test.js
```diff
@@ -43,94 +33,91 @@ moduleFor(
...
 
       this.assertText('One');
-      state = this.expectCacheChanges(
-        { componentDefinitionCount: 1 },
-        state,
+      this.expectCacheChanges(
+        {
+          componentDefinitionCount: 1,
+          // 1 from this.render, 1 from component-one
+          templateCacheMisses: 2,
```
krisselden: prior to this change, this path was only used for late bound templates that were cached because their lifetime wasn't in the DI system's container.

rwjblue: Yep, thats correct. Are you saying this is bad/wrong/etc?

buschtoens:
https://github.com/emberjs/ember.js/blob/099aec56cf72e8677be95d1fc35e33b10e222fc7/packages/%40ember/-internals/glimmer/lib/component-managers/outlet.ts#L72-L74

`definition.template` is a `Factory` now and this breaks. Where is the error? Should this actually be typed and used as a `Factory`, e.g.:

```ts
    if (DEBUG) {
      let owner = getOwner(this);
      let template = definition.template(owner);
      this._pushToDebugStack(`template:${template.referrer.moduleName}`, environment);
    }
```

Or should this class actually still be instantiated with an `OwnedTemplate`, like before?

buschtoens:
Example of a test suite failure: https://travis-ci.org/salsify/ember-css-modules/jobs/552819509

rwjblue:
`@ember/test-helpers` needs a corresponding change (with feature detection) to avoid the error. FWIW, this is why we did this change early in the canary cycle so we can shake out any issues prior to beta.

buschtoens:
@rwjblue Thanks for the pointer! I submitted a PR: https://github.com/emberjs/ember-test-helpers/pull/677

rwjblue:
https://github.com/emberjs/ember.js/pull/18168 should fix the specific issue that @buschtoens ran into (without necessarily coupling it to an `@ember/test-helpers` update).

PR #18110, opened 2019-06-18
[BUGFIX beta] join runloop in `Helper#recompute`

### LINKED ISSUE #14774: Ember 2.10.0 helper recompute bug
Calling recompute on a helper does not recompute in 2.10.0

Twiddle: https://ember-twiddle.com/3afc1bda1c824f4ed141b666e5521320?openFiles=helpers.my-helper.js%2C

When you change the dependency to 2.9 this code works, for 2.10 it does not work. It might be related to Glimmer update?

buschtoens:
Fixes https://github.com/emberjs/ember.js/issues/14774.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/helper.ts
+++ b/packages/@ember/-internals/glimmer/lib/helper.ts
@@ -6,6 +6,7 @@ import { Factory } from '@ember/-internals/owner';
 import { FrameworkObject, setFrameworkClass } from '@ember/-internals/runtime';
 import { symbol } from '@ember/-internals/utils';
 import { EMBER_FRAMEWORK_OBJECT_OWNER_ARGUMENT } from '@ember/canary-features';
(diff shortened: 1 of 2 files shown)


rwjblue (commented):
Looks great! Just wanted to confirm that this test failed prior to adding the `run.join`, right?

buschtoens:
@rwjblue Correct, it did! ☺️

rwjblue:
Thank you!

Windvis:
According to the [LTS info page](https://emberjs.com/releases/lts) LTS releases receive bugfixes for 30 weeks. Doesn't that mean that this fix should have been backported to the 3.8 LTS since the PR was merged in june? 😄

rwjblue:
Sure, if we had thought about it at the time, we definitely could have. At this point 3.8 is receiving security fixes only, so it will not get this fix.

PR #11854, opened 2015-07-21
[CLEANUP beta] Lots of deprecation removals.

rwjblue:
- [CLEANUP beta] Remove `length` from `OrderedSet` and `Map`.
- [CLEANUP beta] Remove `OrderedSet.prototype.length`.
- [CLEANUP beta] Remove `Ember.libraries.each`.
- [CLEANUP beta] Remove deprecated special `{{each}}` keys.
- [CLEANUP beta] Remove Ember.Location.registerImplementation.
- [CLEANUP beta] Remove `{{template}}` support.
- [CLEANUP beta] Remove Ember.Route#setupControllers deprecation.
- [CLEANUP beta] Remove Ember.Route#renderTemplates deprecation.

PR #18127, opened 2019-06-20
remove deprecated `Location.create`

bekzod:
Since removal of `Ember.Location.registerImplementation` in https://github.com/emberjs/ember.js/pull/11854 there is no public way registering location factory inside `Location` so makes `Location.create` redundant, if you still think it need deprecation warnings happy to add that instead


ef4:
Should we continue to export the interface? While the implementation of `Location` is not needed for anything, the interface is still what people should extend if implementing a custom `Location` in typescript, I think?

sandstrom:
I'm all for removing it. Just wanted to mention that I included this in the scope of my deprecate auto location RFC, so this PR will mark it as deprecated.

https://github.com/emberjs/ember.js/pull/19510

As mentioned in this PR though, it would probably make just as much sense to remove it outright.

Just wanted to comment here since my PR is related. I'm fine with both removing it right away as well as waiting for deprecation per my PR.

locks:
Let's center the discussion in https://github.com/emberjs/ember.js/pull/19510.

Issue #18076, opened 2019-06-06
[Regression] [v3.10] The `(query-params)` helper can only be used when invoking the `{{link-to}}` component.


Kerrick:
The PR #17779 introduced a backwards-incompatible change in v3.10, which should not have made it into the framework until v4.0 with deprecation warnings preceding its removal.

For many versions of Ember I've been able to do the following, but since this PR landed and v3.10 shipped, I now am presented with an error: <code>The `(query-params)` helper can only be used when invoking the `{{link-to}}` component.</code>

```hbs
{{!-- app/templates/index.hbs --}}
{{my-component optional-params=(if foo (query-params bar="baz"))}}
```

```hbs
{{!-- my-component.hbs --}}
{{#if optional-params}}
  {{#link-to "my.route" optional-params class="actionable"}}Schedule{{/link-to}}
{{else}}
  {{#link-to "my.route" class="passive"}}View{{/link-to}}
{{/if}}
```

This breaks our production application. This should be considered a breaking change. Its presence in v3.10.0 and v3.10.1 should be considered a regression, and I'd love to see a patch version v3.10.2.

rwjblue:
> The PR #17779 introduced a backwards-incompatible change in v3.10, which should not have made it into the framework until v4.0 with deprecation warnings preceding its removal.

I assure you, we understand how SemVer works.

rwjblue:
> ```hbs
> {{!-- my-component.hbs --}}
> {{#if optional-params}}
>   {{#link-to "my.route" class="actionable" query-params=optional-params}}Schedule{{/link-to}}
> {{else}}
>   {{#link-to "my.route" class="passive"}}View{{/link-to}}
> {{/if}}
> ```

@Kerrick - Can you double check that this is what you are actually using? I don't see any code that would support passing `query-params` as a hash argument to `{{link-to}}` (I checked in 3.8 and 3.9). Perhaps you mean:

```hbs
{{#link-to "my.route" class="actionable" optional-params}}Schedule{{/link-to}}
```

Kerrick:
> @Kerrick - Can you double check that this is what you are actually using? I don't see any code that would support passing query-params as a named argument to {{link-to}} (I checked in 3.8 and 3.9). > Perhaps you mean:

I was making a reduced example during a meeting and messed it up. Your correction is accurate.

> I assure you, we understand how SemVer works.

If what I said came across as harsh or reductive I apologize. I tend to over-explain things to a fault. I'm sorry about that.

[...]

Issue #18033, opened 2019-05-16
(query-params) helper assertion failing if wrapped in (if) helper


theroncross:
`{{#link-to "a" (if foo (query-params b="c") (query-params d="e"))}}Some link{{/link-to}}`

is failing the following assertion: https://github.com/emberjs/ember.js/blob/eee02fb72a47ff1ffbc9919fcf3fc4b74b937093/packages/%40ember/-internals/glimmer/lib/components/link-to.ts#L870

which was added here: https://github.com/emberjs/ember.js/commit/e6c040ffe60f72c124f9698ca1712cb812655826

erkie:
A workaround I found to work in this case is to use angle-brackets:

```hbs
<LinkTo @route="a" @query={{if foo (hash b="c") (hash d="e")}}>Some link</LinkTo>
```

bekzod:
related #18076

rwjblue:
Closing in favor of #18076, would you mind chiming in over there?

PR #18145, opened 2019-06-26
don't register `(query-params)` when angle-brackets enabled

bekzod:
when `EMBER_GLIMMER_ANGLE_BRACKET_BUILT_INS ` enabled there is no need for `(query-params)` since it is being transformed to `@query={{hash ....}}` when correctly used, if not correctly used it throw assertion

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/components/link-to.ts
+++ b/packages/@ember/-internals/glimmer/lib/components/link-to.ts
@@ -863,17 +863,6 @@ if (EMBER_GLIMMER_ANGLE_BRACKET_BUILT_INS) {
             this.query === UNDEFINED
           )
         );
-
-        if (DEBUG && this.query === UNDEFINED) {
-          let { _models: models } = this;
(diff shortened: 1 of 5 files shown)


chancancode:
I think it may still be needed for custom link-to subclasses

bekzod:
Yes, you are genius thanks

chancancode:
It probably could use a test since nothing failed 🤔🤔🤔

bekzod:
@chancancode if sub-classes of link-to exist they are already broken since when <AngleBrackets/> enabled they extend from angle bracket version of link-to which only supports passing queryparam as `@query=`, IMHO angle brackets flag should disabled for link-to and we need to deprecate (query-params ...) helper first to user `@query` or `query` first ?  because it is causing bunch of issues for users https://github.com/emberjs/ember.js/issues/18076  https://github.com/emberjs/ember.js/issues/18033

chancancode:
It is not that the component does not support passing query params using the helper + positional params, it's just that the `(query-params)` assertion was too aggressive.

PR #18148, opened 2019-06-26
[BUGFIX beta] fix(glimmer/utils/to-bool): `htmlSafe`

### LINKED ISSUE #17486: Empty `htmlSafe` string is treated as truthy
The `{{if}}` (and `{{unless}}`, `{{with}}`) template helpers treat an empty `htmlSafe` string as truthy, while an empty string is treated as falsy and even an empty array is treated as falsy.

```hbs
{{if ''             'true' 'false'}} => false
{{if (html-safe '') 'true' 'false'}} => true
{{if (array)        'true' 'false'}} => false
```

[Ember Twiddle](https://ember-twiddle.com/f178e6a099aa8669b620e3b3c897a53f?numColumns=2&openFiles=controllers.application.js%2Ctemplates.application.hbs)

Interestingly I do not understand why an empty array is treated as falsy, when looking at the code:

https://github.com/emberjs/ember.js/blob/1bea9afaac785551a17bf7767ecf5468da63f97b/packages/%40ember/-internals/glimmer/lib/helpers/if-unless.ts#L41-L47

buschtoens:
Fixes #17486.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/utils/to-bool.ts
+++ b/packages/@ember/-internals/glimmer/lib/utils/to-bool.ts
@@ -1,3 +1,4 @@
+import { isHTMLSafe } from '@ember/-internals/glimmer';
 import { get, tagForProperty } from '@ember/-internals/metal';
 import { isArray } from '@ember/-internals/runtime';
 import { isProxy } from '@ember/-internals/utils';
@@ -12,6 +13,8 @@ export default function toBool(predicate: unknown): boolean {
     consumeTag(tagForProperty(predicate as object, '[]'));
 
     return (predicate as { length: number }).length !== 0;
+  } else if (isHTMLSafe(predicate)) {
+    return Boolean(predicate.toString());
   } else {
     return Boolean(predicate);
   }
--- a/packages/@ember/-internals/glimmer/tests/utils/shared-conditional-tests.js
+++ b/packages/@ember/-internals/glimmer/tests/utils/shared-conditional-tests.js
@@ -3,6 +3,7 @@
 import { RenderingTestCase, applyMixins, runTask } from 'internal-test-helpers';
 
 import { assign } from '@ember/polyfills';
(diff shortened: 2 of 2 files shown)


rwjblue (approved):

buschtoens:
@rwjblue Can you try restarting the build? There was [a timeout](https://travis-ci.org/emberjs/ember.js/jobs/550912947#L1498-L1501).

```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

rwjblue:
I've restarted a few times, IE11 continues to fail in the same general way. 🤔

packages/@ember/-internals/glimmer/lib/utils/to-bool.ts
```diff
@@ -1,9 +1,12 @@
+import { isHTMLSafe } from '@ember/-internals/glimmer';
 import { isArray } from '@ember/-internals/runtime';
 import { Opaque } from '@glimmer/interfaces';
 
 export default function toBool(predicate: Opaque): boolean {
   if (isArray(predicate)) {
     return (predicate as { length: number }).length !== 0;
+  } else if (isHTMLSafe(predicate)) {
+    return Boolean(String(predicate));
```
rwjblue: Can we call the string? `predicate.toString()`?

buschtoens: So like this?

```suggestion
    return Boolean(predicate.toString());
```

rwjblue: ya, thats what I was thinking, but I'm not really sure if thats the issue or not (was just grasping at straws a bit)

locks:
Thanks for the help!

PR #18150, opened 2019-06-26
[BUGFIX lts] Remove all own listeners

### LINKED ISSUE #18140: Memory issues in 3.8
We're having memory retention issues upgrading from 3.4 to 3.8. When we use string based event listeners they get retained on the meta of the object that's being listened to. eg

```js
// "3.8.2"
var person = Ember.Object.extend(Ember.Evented, {
  greet() {
    this.trigger('greet');
  }
}).create();
var Foo = { hello() { console.log('hello') }  }
person.on('greet', Foo, 'hello')
person.greet()
// hello
Ember.meta(person)._listeners
 //  (4) ["greet", {…}, "hello", false]
person.off('greet', Foo, 'hello')
Ember.meta(person)._listeners
// [{event: "greet", target: {…}, method: "hello", kind: 2}] '_listeners' still has the event
```

After `person.off('greet', Foo, 'hello')` we would expect `Ember.meta(person)._listeners` to be an empty array - this is the behaviour we see in 3.4 and is also the behaviour we see when we use function listeners:

```js
person.on('greet', Foo, Foo.hello)
Ember.meta(person)._listeners
 //  (4) ["greet", {…}, "hello", false]
person.off('greet', Foo, Foo.hello)
Ember.meta(person)._listeners
// [] An empty array
```

It looks like the root cause is in a change that was added to retain these events, and instead give them a `kind` property of `ListenerFlags.REMOVE` (in https://github.com/emberjs/ember.js/pull/16874/files#diff-1c610ca34cd0e59afd9814e4e3b92806R471).

A later change was added to remove function listeners all together - presumably to fix issues with function listeners being retained (in https://github.com/emberjs/ember.js/pull/16923/commits/7164bd6c69f4c9b3ce33d129aac564dc1ec95920), but this was not done for string listeners.

This behaviour is causing large memory retention issues in our application and stopping us from upgrading. Should string listeners be cleaned up as well or should we change these all to function listeners to avoid the retention issues?

Have also tested this in 3.10 and see the same issue.

pzuraq:
We were previously only removing own listeners if they were functions,
this PR updates us to always remove all of them.

Fixes #18140

### DIFF
--- a/packages/@ember/-internals/meta/lib/meta.ts
+++ b/packages/@ember/-internals/meta/lib/meta.ts
@@ -594,14 +594,9 @@ export class Meta {
     } else {
       let listener = listeners[i];
 
-      // If the listener is our own function listener and we are trying to
-      // remove it, we want to splice it out entirely so we don't hold onto a
-      // reference.
-      if (
-        kind === ListenerKind.REMOVE &&
-        listener.kind !== ListenerKind.REMOVE &&
-        typeof method === 'function'
-      ) {
+      // If the listener is our own listener and we are trying to remove it, we
+      // want to splice it out entirely so we don't hold onto a reference.
+      if (kind === ListenerKind.REMOVE && listener.kind !== ListenerKind.REMOVE) {
         listeners.splice(i, 1);
       } else {
         assert(
--- a/packages/@ember/-internals/meta/tests/listeners_test.js
+++ b/packages/@ember/-internals/meta/tests/listeners_test.js
@@ -172,5 +172,33 @@ moduleFor(
         'one reopen call after mutating parents and flattening out of order'
       );
     }
+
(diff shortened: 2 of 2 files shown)


rwjblue (commented):
Changes seem generally good, but we really need to get a regression test in here too I think...

rwjblue (approved):

rwjblue:
Thank you!

BillyRayPreachersSon:
Will this fix be backported to Ember 3.8, so that LTS users get the benefit? This is really hampering us upgrading from 3.4 LTS, and the memory usage in our app goes through the roof.

Thanks!

pzuraq:
The fix is marked for LTS releases, so yes, it will.

PR #18158, opened 2019-06-27
[FEATURE set-component-template]

chancancode:
Implements the Ember side of [RFC #481](https://emberjs.github.io/rfcs/0481-component-templates-co-location.html).


rwjblue:
Fixed merge conflicts

packages/@ember/-internals/glimmer/lib/resolver.ts
```diff
@@ -63,6 +66,117 @@ function makeOptions(moduleName: string, namespace?: string): LookupOptions {
...
+  if (component === null && layout === null) {
+    return null;
+  } else {
+    return { component, layout } as LookupResult;
+  }
+}
+
+function lookupComponent(owner: Owner, name: string, options: LookupOptions): Option<LookupResult> {
+  if (options.source || options.namespace) {
+    if (EMBER_MODULE_UNIFICATION) {
```
pzuraq: Do we need these? I thought we'd removed them

pzuraq (approved):
Generally looks good! Just one minor question, everything else is 👍

packages/@ember/-internals/glimmer/lib/resolver.ts
rwjblue: @pzuraq - We disabled the flag (though when we did that @chancancode and I had already done the work here to support it) because it was impossible to test _either_ colocation or MU (we only test with all optionals or none) and some aspects of the feature conflicts with MU. 

tldr; we haven't removed the MU feature flagged code yet, and _just in case_ we need to salvage parts of it (who knows why) we wanted to leave things roughly in the same basic state before/after this PR

PR #18168, opened 2019-07-02
Fix issues with the recent template factory refactor and @ember/test-helpers v1.6.0

rwjblue:
This PR does two main things:

* Adds a smoke test that emulates what `@ember/test-helpers@1.6.0` does for `setupRenderingTest` type of tests
* Fixes issues within the `{{outlet}}` component manager when the result of `owner.lookup` is passed directly into the outlet state (which is what `@ember/test-helpers` has done for most of the 1.x series)

I do _personally_ think the fix/workaround is "fine" to leave in for quite some time, but it can be removed after `@ember/test-helpers@1.6.0` is "aged out" and no longer supported.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/syntax/outlet.ts
+++ b/packages/@ember/-internals/glimmer/lib/syntax/outlet.ts
@@ -12,6 +12,7 @@ import {
 import * as WireFormat from '@glimmer/wire-format';
 import { OutletComponentDefinition, OutletDefinitionState } from '../component-managers/outlet';
 import { DynamicScope } from '../renderer';
+import { isTemplateFactory } from '../template';
 import { OutletReference, OutletState } from '../utils/outlet';
 
 /**
@@ -124,6 +125,13 @@ function stateFor(
   if (render === undefined) return null;
   let template = render.template;
   if (template === undefined) return null;
+
+  // this guard can be removed once @ember/test-helpers@1.6.0 has "aged out"
+  // and is no longer considered supported
+  if (isTemplateFactory(template)) {
+    template = template(render.owner);
+  }
+
   return {
     ref,
     name: render.name,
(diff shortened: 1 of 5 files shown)


packages/@ember/-internals/glimmer/lib/syntax/outlet.ts
```diff
@@ -124,6 +125,13 @@ function stateFor(
   if (render === undefined) return null;
   let template = render.template;
   if (template === undefined) return null;
+
+  // this guard can be removed once @ember/test-helpers@1.6.0 has "aged out"
+  // and is no longer considered supported
```
buschtoens: So is the intention to still merge and release https://github.com/emberjs/ember-test-helpers/pull/677 or to rather make a new backwards-incompatible release of `@ember/test-helpers` that drops support for all Ember versions prior to https://github.com/emberjs/ember.js/pull/18096? 🤔

rwjblue: Ya, I still think we should still land https://github.com/emberjs/ember-test-helpers/pull/677.

packages/@ember/-internals/glimmer/lib/template.ts
```diff
@@ -8,6 +8,10 @@ import { SerializedTemplateWithLazyBlock } from '@glimmer/wire-format';
 export type StaticTemplate = SerializedTemplateWithLazyBlock<StaticTemplateMeta>;
 export type OwnedTemplate = Template<OwnedTemplateMeta>;
 
+export function isTemplateFactory(template: OwnedTemplate | Factory): template is Factory {
```
krisselden: Should we rename the interface below to TemplateFactory?

krisselden (approved):

krisselden:
We have a test that is hanging in the browserstack tests.

PR #18183, opened 2019-07-09
Correct property name in @union cp example

GCheung55:
The example usage of `@union` defines `ediblePlants` but uses `hamster.uniqueFruits` instead.

Change `ediblePlants` to `uniqueFruits`.


rwjblue (approved):
Thank you!

PR #18190, opened 2019-07-11
[FEATURE set-component-template] Add @ember/component/template-only

rwjblue:
As part of emberjs/rfcs#481 this adds the `@ember/component/template-only` module, and its associated implementation.

### DIFF
--- a/packages/@ember/component/template-only.ts
+++ b/packages/@ember/component/template-only.ts
@@ -1,0 +1,45 @@
+// This is only exported for types, don't use this class directly
+export class TemplateOnlyComponent {
+  constructor(public moduleName = '@ember/component/template-only') {}
+
+  toString(): string {
+    return this.moduleName;
+  }
+}
+
+/**
+  @module @ember/component/template-only
+  @public
+*/
+
+/**
+  This utility function is used to declare a given component has no backing class. When the rendering engine detects this it
+  is able to perform a number of optimizations. Templates that are associated with `templateOnly()` will be rendered _as is_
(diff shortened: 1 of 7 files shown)


rwjblue:
Updated to fix linting error.

packages/@ember/component/template-only.ts
```diff
@@ -0,0 +1,16 @@
+// This is only exported for types, don't use this class directly
+export class TemplateOnlyComponent {
+  constructor(public moduleName: string) {}
```
chancancode: `moduleName?`

packages/@ember/component/template-only.ts
```diff
@@ -0,0 +1,16 @@
...
+
+  toString(): string {
+    return '@ember/component/template-only';
```
chancancode: Should we include the module name here (if present?)

packages/@ember/component/template-only.ts
```diff
@@ -0,0 +1,16 @@
...
+  }
+}
+
+export default function templateOnlyComponent(moduleName: string): TemplateOnlyComponent {
```
chancancode: `moduleName?`

packages/@ember/component/template-only.ts
rwjblue: Defaulted to `@ember/component/template-only`, and updated the `toString` to `return this.moduleName`.

rwjblue:
@chancancode - Updated to use the passed in `moduleName` for the `toString`. Also did an initial pass at API docs for the new method.

PR #18207, opened 2019-07-16
 [CLEANUP lts] Change node-module into a standard package

pzuraq:
Right now, `node-module` is a precompiled package, but that makes it
harder to get it to properly integrate with dynamic builds. This changes
it to be a standard package that goes through the build process like
normal.

Depends on #18206


rwjblue (commented):
Ready for a rebase

broccoli/transforms/inject-node-globals.js
```diff
@@ -6,7 +6,9 @@ function injectNodeGlobals({ types: t }) {
     name: 'inject require',
     visitor: {
       Program: {
-        enter(path) {
+        enter(path, state) {
+          if (state.file.opts.filename.match(/node-module\/index.js/)) return;
```
rwjblue: `filename.endsWith('node-module/index.js')` might be easier to read:

```suggestion
          if (state.file.opts.filename.endsWith('node-module/index.js')) return;
```

rwjblue (approved):

broccoli/transforms/inject-node-globals.js
pzuraq: Can followup with this later, I'd like to get it merged for now

Issue #18084, opened 2019-06-09
Version 3.10.1 regression: use of const in strict mode, causing critical errors in e.g. Safari 9


samcic:
In `ember.min.js` version 3.10.1 there are various usages of the `const` keyword in `use strict` mode:

* `@ember/-internals/glimmer`, example: `const e=L(["template-compiler:main"]`
* `@ember/application/lib/application`, example: `const e=function(e,t){t...`
* `@ember/engine/index`, example: `const e=v(["-bucket-cache:main"])`
* `@ember/engine/instance`, example: `const e=u(["template-compiler:main"])`

These cause syntax errors in Safari 9 (e.g. on iPhone 6S) and the Ember app fails to start. The error is:

`SyntaxError: Use of const in strict mode`

Ember 3.9.1 (the previous version) didn't have this problem.

Repro:

`npm install -g ember-cli@3.10.1`
`ember new test-app`
`yarn install`
(edit `config/targets.js` to include `Safari >= 9`)
`ember build -prod`
(inspect `dist/vendor.js` and see the above-mentioned `const` occurrences)

Is this a bug, or am I missing something with respect to the build process?

rwjblue:
Hmm, this is indeed surprising. If you add IE11 to the `config/targets.js` do you still see the `const`s?  I'm guessing that our detection of whether or not we should use the "legacy" build or not is deciding not to use legacy when it should. 

My best _guess_ is that this as to do with https://github.com/emberjs/ember.js/pull/17859, as that is one of the only changes in the build pipeline (specifically WRT selecting legacy or modern builds).

samcic:
Yes, the `const`s are still there if IE11 is included in the targets.

samcic:
One thing I have noticed in the build process is the message:

> Browserslist: caniuse-lite is outdated. Please run next command `yarn upgrade caniuse-lite browserslist`

I'm guessing that this is a dependency of one of the ember-related packages and it's something that we're not responsible for updating ourselves. I wanted to mention it regardless though.

austinezell:
I am currently experiencing this issue as well. The only additional thing I notice that might be worth pointing out is that the application specific code is still being correctly transpiled to legacy; in the above repro steps, the `test-app.js` file generated successfully strips out all `let`/`const` or other non supported features based on what is in the `targets.js`. It is only the the `vendor.js` file appears to fail.

pzuraq:
[...]

PR #18208, opened 2019-07-17
 [BUGFIX lts] Compile Ember dynamically in consuming applications

pzuraq:
* **Phase 1, prepublish:** Run Typescript and Rollup on the main Ember
  packages, and strip out canary-features. Publish these packages, along
  with the dependencies for Ember, in the `dist/` folder.

* **Phase 2, in addon:** Run Babel transpilation using the consumer's
  `ember-cli-babel` and Babel configuration on the dist packages and
  dependencies, and bundle up the final `ember.js` file to serve to
  apps. This also includes `debug` flags and, in theory, svelting.

Two major changes that will occur because of this:

1. We will no longer be distributing `ember.prod.js`, `ember.debug.js`,
   `ember.min.js`, or `ember-testing.js`. These files existing may be
   something that people rely on, and the packages that _are_
   distributed aren't quite ready to build in a "normal" way, using
   webpack or another bundler.
2. We will no longer be _building_ `ember.prod.js`, `ember.min.js`, or
   `ember.debug.js` at all, only a single `ember.js` file. We no longer
   need separate builds, because the environment settings of the build
   will handle the differences for us.

Unfortunately, until we disentangle `ember-cli` from the implementation
details of `ember-source` we cannot convert Ember to _build_ like a
normal addon locally. Using the `EmberAddon` class blows up in many
small ways.


rwjblue:
Looks like the CI failures are related:

```
not ok 1 App Boot > App boots and routes to a URL
  ---
  message: "Died on test #1     at Object.<anonymous> (/home/travis/build/emberjs/ember.js/tests/node/app-boot-test.js:8:9): Could not find module `url` imported from `(require)`"
  severity: failed
  actual: null
  expected: undefined
  stack: Error: Could not find module `url` imported from `(require)`
    at missingModule (/home/travis/build/emberjs/ember.js/dist/ember.js:247:11)
    at findModule (/home/travis/build/emberjs/ember.js/dist/ember.js:258:7)
    at requireModule (/home/travis/build/emberjs/ember.js/dist/ember.js:24:15)
    at installProtocolForURL (/home/travis/build/emberjs/ember.js/dist/ember.js:6561:17)
    at new Environment$1 (/home/travis/build/emberjs/ember.js/dist/ember.js:6601:7)
    at Function.create (/home/travis/build/emberjs/ember.js/dist/ember.js:6613:14)
    at FactoryManager.create (/home/travis/build/emberjs/ember.js/dist/ember.js:1125:33)
    at instantiateFactory (/home/travis/build/emberjs/ember.js/dist/ember.js:929:63)
    at _lookup (/home/travis/build/emberjs/ember.js/dist/ember.js:861:12)
    at processInjections (/home/travis/build/emberjs/ember.js/dist/ember.js:969:26)
```

runspired:
Just a heads up, this PR may not be enough.

Currently in ember-source 3.11 when native extends is mixed with `.extend` travis-web fails against ember-data when:

- targets includes "last 2 edge"
- AND targets does NOT include IE11

It passes when:

- targets does not include edge
- AND targets does NOT include IE11

- targets includes "last 1 edge"
- AND targets does NOT include IE11

- targets does not include "edge" or includes "last 1 edge"/"last 2 edge"
- AND targets INCLUDES IE11

You can reproduce this by:

- cloning `ember-data`
- running `yarn install`
- running `node ./bin/packages-for-commit.js`
- running `yarn test-external:travis-web --skip-smoke-test`

This test will fail and you will find `travis-web` in the directory `../__external-test-cache/travis-web` in a state wherein you may alter the `targets.js` file and `CI` status as desired and run tests as normal in an ember app.

pzuraq:
Based on your description @runspired, this PR should fix those issues as well. I believe the problem is the same as #18084, where specific targets are causing us to ship the wrong build and causing the conflict. With this PR, we will be compiling Ember using the same targets as the consuming application, so this type of conflict won't be possible anymore.

BnitoBzh:
Hello everyone, when do you think this problem can be solved? We need our production applications to run on ios9, now for several weeks we are having trouble. The only solution we have now is to edit the build of applications after each deployment by replacing all `const` with `var`.

pzuraq:
Hey, sorry, have been working on perf tuning for tracked properties in 3.13 this last week. I'm wrapping that up, should be able to get to this sometime this week.

ember-cli-build.js
```diff
@@ -28,39 +20,50 @@ const {
...
-    dagES(),
-    routerES(),
-    routeRecognizerES(),
+const ENV = process.env.EMBER_ENV || 'development';
+const SHOULD_ROLLUP = process.env.SHOULD_ROLLUP !== 'false';
+const SHOULD_MINIFY = Boolean(process.env.SHOULD_MINIFY);
+
+/**
+ * There isn't a way for us to override targets through ember-cli-babel, and we
+ * don't want to introduce that functionality for reasons. This is a quick and
```
rwjblue: I'd like to address this over in ember-cli-babel (in a follow up), would you mind making an issue over there about the general use case?

BnitoBzh:
Hi ! is this pull request can be merged into ember 3.12.x or 3.11.x ?

pzuraq:
This is marked for backport to LTS, so it will be backported to 3.12 for sure

lib/index.js
```diff
@@ -66,48 +157,22 @@ module.exports = {
...
-        return flat.concat(jsAndMap);
-      }, [])
-      .filter(function(file) {
-        let fullPath = path.join(emberSourceDistPath, file);
-
-        return fs.existsSync(fullPath);
+    if (!isProduction && PRE_BUILT_TARGETS.every(target => targets.includes(target))) {
+      ember = new Funnel(tree, {
+        destDir: 'ember',
+        include: ['ember.debug.js', 'ember.debug.map', 'ember-testing.js', 'ember-testing.map'],
```
josemarluedke: Should `ember-template-compiler.js` be present here as well?

This change might have caused this: https://github.com/embroider-build/embroider/issues/311.

rwjblue: That seems plausible, mind making a PR so we can focus discussion over there?

josemarluedke: That's done at https://github.com/emberjs/ember.js/pull/18291.

Turbo87:
why is this PR marked as "BUGFIX lts"? the description is not mentioning any bugs that are being fixed by this and it seems like a pretty significant change for those that expect LTS to work the same way as before 🤔

snewcomer:
Has this been backported to 3.12?  Is this something I can help with.  This is the internal addon we have to fix this temporarily (we are on 3.12)

```js
'use strict';
const replace = require('broccoli-replace');

module.exports = {
    name: 'temporary-ember-fix',
    postprocessTree(type, tree) {
        if (type !== 'all') {
            return tree;
        }

        return replace(tree, {
            files: ['**/vendor*.js'],
            patterns: [{
                match: /const\s/g,
                replacement: 'var '
            }]
        });
    },

    isDevelopingAddon() {
        return true;
    }
};
```

pzuraq:
I don't believe it has yet, but it definitely should be. It's a large change, but it's been pretty stable so far, no more issues, and it shouldn't be too hard to pull it onto 3.12 (originally we were hoping to get it in then, but we decided against it since we didn't have enough time for testing).

If you can take a crack at it, that'd be very helpful, otherwise I'll try to find some time in the near future to get it going.

rwjblue:
Ya, the major thing to do is identify all the bugfix commits for these files that happened after this initially landed so we have a clearer idea of what to backport.

ssnielsen:
Very glad to see that this has/is being worked on :) We've recently upgraded to 3.12.0 and have had reports from our iOS 9 users about the issue mentioned in https://github.com/emberjs/ember.js/issues/18084

@rwjblue What do you estimate is the timeframe wrt. getting this fix into a future 3.12.x release?

rwjblue:
TBH, I don't know. We really need to figure out the full list of changes needed to backport, that is probably the main blocker. If someone has time to dig through the commit history for anything build related since this PR landed and make a PR targetting the `lts-3-12` branch that would be super helpful and make it massively more likely for it to happen "soon". After that actually doing the release is pretty simple once we get the changes there.

pzuraq:
FWIW I made https://github.com/pzuraq/ember-class-constructor-fix which _should_ fix most of the general issues in older builds. It's a more general fix too, which means it'll work prior to 3.12 as well.

#18084 is particularly strange though, since it _appears_ that the legacy build is not being used, despite very clear legacy targets. Will comment over there to try to get more details.

snewcomer:
This is phenomenal!  I love the approach.  Is this the fix or is backporting changes the fix?  Seems like the former?

pzuraq:
I think backporting would still be useful, if possible, but I don't think we need to spend extra cycles on it. This fix should work in any cases where we eagerly opt into the legacy build even if we shouldn't _actually_ do that, and as I was working through the fix I thought that was the only possible thing that could accidentally happen.

This is why #18084 is so strange to me, based on the code I don't think it should be possible. We basically only opt-into legacy IFF classes need to be compiled _OR_ another feature needs to be compiled, so 

1. If classes are compiled, then the legacy build is used, and user's classes are compiled, so nothing to fix.
2. If classes are _not_ compiled, and nothing else triggers the legacy build, then there's nothing to fix.
3. If classes are _not_ compiled, but something else is and triggers the legacy build, then user code is _not_ compiled, but framework code is, and the fix patches it.

I don't think it should be possible for user classes to be compiled, _and_ for the framework to be the legacy build, but maybe there's something I wasn't considering.

PR #18212, opened 2019-07-17
Remove Duplicate/Confusing Filter Tests

elwayman02:
The two tests removed duplicate functionality more extensively tested by other tests in this suite. Specifically:

`@test should filter based on object` - This tests the same behavior as `@test should filter on second argument if provided`, but against a smaller array. 
`@test should not return all objects on undefined second argument` - This tests the same behavior as `@test should correctly filter explicit undefined second argument`, but against an array where no objects have an explicit `undefined` value for that key. The second test correctly asserts that defined values should be filtered out *and* that explicit undefined values should be kept, making the first test unnecessary.


pzuraq (approved):

pzuraq:
Thanks for the PR!

PR #18213, opened 2019-07-18
Fixed dead links

HenryVonfire:


HenryVonfire:
It seems the issue comes from somewhere else (possibly from ember-jsonapi-docs as suggested by @jenweber ). Closing this PR then.

PR #18214, opened 2019-07-19
 Add deprecations for mouseEnter/Leave/Move Ember events

simonihmig:
As per [RFC486](https://github.com/emberjs/rfcs/blob/master/text/0486-deprecate-mouseenter.md)


packages/@ember/-internals/glimmer/lib/modifiers/action.ts
```diff
@@ -241,6 +241,20 @@ export default class ActionModifierManager implements ModifierManager<ActionStat
       dom,
       tag
     );
+
+    deprecate(
+      `Using the \`{{action}}\` modifier with \`${actionState.eventName}\` events has been deprecated.`,
+      actionState.eventName !== 'mouseEnter' &&
```
rwjblue: I'm a tad worried about:

```hbs
<div {{action 'foo' on=this.mouseMethod}}></div>
```

Do we have a test for that? Does it also emit the deprecation warning?

rwjblue (commented):
Looking really good! 

A few small things:

* Ensure there is a test for `<div {{action 'foo' on={{this.mouseMethod}}></div>`?
* Can we avoid adding these listeners when we do not detect the deprecated situations? Maybe I should modify my other PR to be limited to just these three events (so we can more easily land it)?

packages/@ember/-internals/glimmer/lib/modifiers/action.ts
simonihmig: There already is: https://github.com/emberjs/ember.js/pull/18214/files#diff-6f2e4a5d59aa17cdb7b0436b606195b1R1763

simonihmig:
> Can we avoid adding these listeners when we do not detect the deprecated situations? Maybe I should modify my other PR to be limited to just these three events (so we can more easily land it)?

Seems reasonable to me! (for others, we are talking about https://github.com/emberjs/ember.js/pull/17911)

Either way I think we can/should do it as a separate PR, as this is just concerned about the deprecations, and the other is about optimizations.

rwjblue:
Thank you!

PR #18217, opened 2019-07-22
[BUGFIX] Adds ability for computed props to depend on args 

pzuraq:
This PR adds the ability for computed properties and observers to watch
`args` on custom components. Currently this does not work because the
`args` object is not a standard object, but a proxy that forwards
property access to the `CapturedNamedArgs` instance for the component,
and consumes its tag for autotracking.

CPs currently ignore autotracking, and have to depend on `args` manually
via the `chainTags` system. In the first attempt to support this we used
the `UNKNOWN_PROPERTY_TAG` API, but that actually doesn't work for this
case because these _aren't_ unknown properties, they do "exist" on the
proxy and should be enumerable.

Rather than adding another general API to the `tagForProperty` function,
I opted to add some special logic to the `chainTags` crawler, similar to
what we do for `@each`. When the chain is looking up `args` as its
segment, it'll also check to see if the value is an args proxy, and if
so it'll use the (saved) CapturedNamedArgs directly. This check should
be low overhead since it's a string comparison, and only applies to CPs
and observers rather than _all_ code (compared to adding it to
`tagForProperty`, which would affect new code as well).

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/component-managers/custom.ts
+++ b/packages/@ember/-internals/glimmer/lib/component-managers/custom.ts
@@ -1,4 +1,4 @@
-import { consume } from '@ember/-internals/metal';
+import { ARGS_PROXY_TAGS, consume } from '@ember/-internals/metal';
 import { Factory } from '@ember/-internals/owner';
 import { HAS_NATIVE_PROXY } from '@ember/-internals/utils';
 import { OwnedTemplateMeta } from '@ember/-internals/views';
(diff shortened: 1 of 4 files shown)


packages/@ember/-internals/glimmer/lib/component-managers/custom.ts
```diff
@@ -176,6 +176,29 @@ export default class CustomComponentManager<ComponentInstance>
 
             return ref.value();
           },
+
+          has(_target, prop) {
+            assert('args can only be strings', typeof prop === 'string');
```
rwjblue: I think this is slightly incorrect wording, AFAICT the arg _key_ can only be a string (but the value can be anything), right?

packages/@ember/-internals/glimmer/tests/integration/components/tracked-test.js
```diff
@@ -485,6 +485,55 @@ if (EMBER_METAL_TRACKED_PROPERTIES) {
           assert.equal(outerRenderCount, 2, 'outer component updates based on context');
           assert.equal(innerRenderCount, 3, 'inner component updates based on outer component');
         }
+
+        '@test computed properties computed properties can depend on args'() {
```
rwjblue: The phrasing here seems odd, is it intentional?

packages/@ember/-internals/glimmer/tests/integration/components/tracked-test.js
```diff
@@ -485,6 +485,55 @@ if (EMBER_METAL_TRACKED_PROPERTIES) {
...
+          this.assertText('hello!');
+
+          runTask(() => this.context.set('text', 'hello world!'));
+          this.assertText('hello world!');
+
+          runTask(() => this.context.set('text', 'hello!'));
+          this.assertText('hello!');
+        }
+
+        '@test named args are enumerable'() {
```
rwjblue: Can you add another test for using `{{#each-in this.args as |arg value|}}`?

packages/@ember/-internals/glimmer/tests/integration/components/tracked-test.js
pzuraq: lol, definitely not, will fix 😛

packages/@ember/-internals/glimmer/tests/integration/components/tracked-test.js
pzuraq: Good thing we added that, turns out it is necessary to support symbols in the proxy. I decided to make the guard in the `get` trap a bit more generic to handle that.

rwjblue (approved):

PR #18222, opened 2019-07-30
[BUGFIX] Matches assertion behavior for CPs computing after destroy

pzuraq:
Previously, there was an assertion that threw whenever a CP was
calculated after destroy, _and_ updated its dependent keys (i.e.
updated chains). This meant that CPs without any dependencies would
still run after destroy, and everything would work. The new assertion
always threw if anyone accessed _any_ CP. This changes it to only throw
if the CP has dependencies, matching the old behavior.

### DIFF
--- a/packages/@ember/-internals/metal/lib/computed.ts
+++ b/packages/@ember/-internals/metal/lib/computed.ts
@@ -541,6 +541,8 @@ export class ComputedProperty extends ComputedDescriptor {
     let ret;
 
     if (EMBER_METAL_TRACKED_PROPERTIES) {
+      // For backwards compatibility, we only throw if the CP has any dependencies. CPs without dependencies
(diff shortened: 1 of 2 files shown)


packages/@ember/-internals/metal/lib/computed.ts
```diff
@@ -543,7 +543,7 @@ export class ComputedProperty extends ComputedDescriptor {
     if (EMBER_METAL_TRACKED_PROPERTIES) {
       assert(
         `Attempted to access the computed ${obj}.${keyName} on a destroyed object, which is not allowed`,
-        !metaFor(obj).isMetaDestroyed()
+        this._dependentKeys === undefined || !metaFor(obj).isMetaDestroyed()
```
igorT: Might want to leave a comment for why we do this check, but 🤷‍♂

igorT (approved):

PR #18226, opened 2019-08-02
[BUGFIX beta] Fix routing path with double slash

### LINKED ISSUE #14925: Ember 2.11 double forward slash causes application to break

Exelord:
Fix: https://github.com/emberjs/ember.js/issues/14925

When a user is accessing a path followed by `//` eg. `http://localhost:4200//admin/profile` the browser is raising the following error:
```
Failed to execute 'replaceState' on 'History': A history state object with URL 'http://admin/profile' cannot be created in a document with origin 'http://localhost:4200' and URL 'http://localhost:4200//admin/profile'.
```

This PR fixes by replacing any `//` in the URL path to `/`.


rwjblue (approved):
Seems good to me, thank you!

Issue #18228, opened 2019-08-06
Improve documentation of EmberArray


locks:
Going through http://api.emberjs.com/ember/release/classes/EmberArray, you will notice that some of the entries have minimal documentation and no code sample.
The goal is to review each entry, add code samples where missing, and optionally improve the prose.

Please comment in this thread if you pick one of the entries so it can be assigned to you!

## Methods
-  [x] addArrayObserver (@sdebarros) https://github.com/emberjs/ember.js/pull/18272
-  [x] any (@sukima) https://github.com/emberjs/ember.js/pull/18229
-  [ ] arrayContentDidChange
-  [ ] arrayContentWillChange
-  [ ] compact
-  [x] every (@sdebarros) https://github.com/emberjs/ember.js/pull/18236
-  [x] filter (@sdebarros) https://github.com/emberjs/ember.js/pull/18239
-  [x] filterBy (@sdebarros) https://github.com/emberjs/ember.js/pull/18239
-  [x] find (@brunoocasali) https://github.com/emberjs/ember.js/pull/18324
-  [x] findBy (@barryofguilder) https://github.com/emberjs/ember.js/pull/18233
-  [ ] forEach (@Frozenfire92) https://github.com/emberjs/ember.js/pull/18463
-  [x] getEach (@Frozenfire92) https://github.com/emberjs/ember.js/pull/18464
-  [x] includes (@Frozenfire92) https://github.com/emberjs/ember.js/pull/18465
-  [x] indexOf (@Frozenfire92) https://github.com/emberjs/ember.js/pull/18466
-  [x] invoke (@mukilane) https://github.com/emberjs/ember.js/pull/18283
-  [x] isAny (@mcfiredrill) https://github.com/emberjs/ember.js/pull/18505
-  [x] isEvery (@mcfiredrill) https://github.com/emberjs/ember.js/pull/18510
-  [x] lastIndexOf (@Charizard) https://github.com/emberjs/ember.js/pull/18511
-  [x] map (@imkathir) https://github.com/emberjs/ember.js/pull/18376
-  [x] mapBy (@imkathir) https://github.com/emberjs/ember.js/pull/18376
-  [x] objectAt
-  [x] objectsAt
-  [x] reduce
-  [x] reject (@mcfiredrill) https://github.com/emberjs/ember.js/pull/18503
-  [x] rejectBy (@beegan) https://github.com/emberjs/ember.js/pull/18447
-  [ ] removeArrayObserver 
[...]

PR #18229, opened 2019-08-06
[DOC] Update documentation for EmberArray.any

sukima:
Prior to this change, the wording was inaccurate. It was describing filtering not the short circuit behaviour of `any()`.

This change attempts to fix that discrepancy and expand on the example to include both context option and arrow function use.

Issue #18228

### DIFF
--- a/packages/@ember/-internals/runtime/lib/mixins/array.js
+++ b/packages/@ember/-internals/runtime/lib/mixins/array.js
@@ -860,11 +860,10 @@ const ArrayMixin = Mixin.create(Enumerable, {
   },
 
   /**
-    Returns `true` if the passed function returns true for any item in the
-    enumeration.
-
-    The callback method you provide should have the following signature (all
-    parameters are optional):
+    The any() method executes the callback function once for each element
+    present in the array until it finds the one where callback returns a truthy
+    value (i.e. `true`). If such an element is found, any() immediately returns
+    true. Otherwise, any() returns false.
 
     ```javascript
     function(item, index, array);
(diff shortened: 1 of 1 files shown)


packages/@ember/-internals/runtime/lib/mixins/array.js
```diff
@@ -874,18 +874,27 @@ const ArrayMixin = Mixin.create(Enumerable, {
     - `index` is the current index in the iteration.
     - `array` is the array object itself.
 
-    It must return a truthy value (i.e. `true`) to include an item in the
-    results. Any non-truthy return value will discard the item from the
-    results.
+    A return of a falsey value (i.e. `false`) will continue evaluating items in
```
locks: The current paragraph is describing the callback, so it does not make sense to talk about short-circuiting. I also do not think describing it this technically is useful for the average reader.
Maybe you can adjust the description of the callback and say that the result of `any` is true if the callback returns true, otherwise it's false?

packages/@ember/-internals/runtime/lib/mixins/array.js
```diff
@@ -874,18 +874,27 @@ const ArrayMixin = Mixin.create(Enumerable, {
...
+    the array until it reaches the end or encounters a truthy (i.e. `true`)
+    return value. If the evaluation reaches the end if the array the result
+    will be `false`. Conversely any evaluation that ends the loop (via
+    returning truthy) will result in `true`.
 
     Note that in addition to a callback, you can also pass an optional target
-    object that will be set as `this` on the context. This is a good way
-    to give your iterator function access to the current object.
+    object that will be set as `this` on the context. It can be a good way
+    to give your iterator function access to the current object in cases where
```
```suggestion
    to give your iterator function access to an object in cases where
```

packages/@ember/-internals/runtime/lib/mixins/array.js
```diff
@@ -874,18 +874,27 @@ const ArrayMixin = Mixin.create(Enumerable, {
...
+    an ES6 arrow function is not appropriate.
```
```suggestion
    an arrow function would not be appropriate.
```

packages/@ember/-internals/runtime/lib/mixins/array.js
```diff
@@ -874,18 +874,27 @@ const ArrayMixin = Mixin.create(Enumerable, {
...
 
     Usage Example:
 
     ```javascript
-    if (people.any(isManager)) {
+    let includesManager = people.any(function(person) {
```
locks: Maybe include the definition of `people` so that it's clearer what the code sample is achieving.

locks (changes requested):
Left some inline comments.

sukima:
I am not understanding how a comment change in this file broke Blueprint tests in CI!

sukima:
@locks I reworded everything. see if it is better. I did a little paraphrasing from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some but I think it is different enough to prevent a license issue.

locks (approved):

PR #18248, opened 2019-08-09
[BUGFIX beta] Ensures that observers are flushed after CPs are updated

simonihmig:
@pzuraq Feel free to pick that commit to test your #18223 PR with.


pzuraq:
Turned out it was a different issue that was fixed on the perf branch, I updated this PR with a fix for this particular issue, should be good to go!

Issue #18250, opened 2019-08-10
Quest - update API docs to show Octane and Native Classes


jenweber:
It's almost time to show Octane-style code samples in the API docs! Right now, our examples use Ember Objects, and we want them to use native classes and Component file co-location instead. Can you help? Here's how! If you have any questions, drop by the `#dev-ember-learning` channel on [Ember Discord](https://api.emberjs.com/ember/release). Keep reading for step by step instructions, a styleguide, and some examples of the kinds of changes we need help with.

## Step by step instructions

1. Comment on this issue to say which item you are volunteering for. The checklist of things we need help with is in the comment below. Only volunteer for items that you think you can open a PR for in a week or less.
2. Before you start working, double check to make sure it's not already been PR'd or claimed by someone else.
3. Fork this repo and check out the existing branch `octane-docs`. This branch has some of the work done already, and you will be adding to it.  If it's your first open source PR ever, check out [these instructions](https://medium.com/@jenweber/your-first-open-source-contribution-a-step-by-step-technical-guide-d3aca55cc5a6) to learn how to fork, branch, etc. or make changes via the GitHub website instead.
4. Locate the file that has the API doc. Our docs are written in YUIdoc and they live right next to the JavaScript for the API, here in this repository.
5. Make your changes! EmberObject is still supported, so we need to show both styles. [Here's an example](https://api.emberjs.com/ember/3.12/functions/@ember%2Fservice/inject). In general, you should give Octane examples first, and EmberObject second. In cases where multiple examples are needed, make all of them Octane, and then add just 1 for EmberObject.
If you don't know Octane syntax yet, don't worry. There's a [Cheatsheet](https://ember-learn.github.io/ember-octane-vs-classic-cheat-sheet/) and [guide to what is different](https://octane-guides-preview.emberjs.com/release/upgrading/editions/)
6. When you make a commit, it should start with `[DOC beta]`. For example `[DOC beta] update component example for Octane`.
7. Open a pull request, and target merging into the `octane-docs` branch. Please check the box that says "allow maintainers to make edits."
8. Celebrate and brag about it to all your friends

## Styleguide

All inline documentation is written using YUIDoc. Follow these rules when updating or writing new documentation. This is copied & pasted from [CONTRIBUTING.md](https://github.com/emberjs/ember.js/blob/master/CONTRIBUTING.md#pull-requests)

- All code blocks must be fenced
- All code blocks must have a language declared
- All code blocks must be valid code for syntax highlighting
[...]

PR #18252, opened 2019-08-11
[DOC beta] Octanify documentation

locks:
Part of https://github.com/emberjs/ember.js/issues/18250.

## Missing
- [ ] what to do with `mut` in `get` documentation
- [ ] bring back classic component example for `component` helper
- [ ] normalize `set` and `tracked` in `computed_macros` file
- [ ] review `each-in` for correctness/coesion (key is used before it's introduced)
- [ ] confirm `get` + `mut` example updates properly
- [ ] `action_handler` switch examples to `EmberComponent` so that we can still use the classic `.extend` API.
- [ ] investigate syntax highlighting mishaps - [hasBlock](https://api.emberjs.com/ember/3.12/classes/Ember.Templates.helpers/methods?anchor=hasBlock), [hasBlockParams](https://api.emberjs.com/ember/3.12/classes/Ember.Templates.helpers/methods?anchor=hasBlockParams)
- [ ] `observer` is `Helper` class example
- [ ] `RouterService#recognize` uses a strange ember component

## Missing code samples
- [ ] `RouterService#isActive`
- [ ] `RouterService#recognizeAndLoad`

## Open questions

- [x] classes, named or anonymous? (`class extends Route` vs `class Menu extends Route`) -> named

### Router Service
- [ ] constructor seems ok in code samples? (cc @pzuraq)
- [ ] `currentRoute` example is using `notEmpty` CP decorator. should we replace with native getter? (cc @pzuraq)

### unless
- Consider introducing `not` to the framework, and deprecating `unless` for `{{if (not)}}`

### Action Handler
- [ ] (?) `action_handler` remove mixins (?)

### Controller
- [ ] figure out what to do with `Controller#addObserver` - it's currently using an EmberComponent, which seems strange

### action helper
- [ ] figure out how we want to address classic EmberComponent in [action helper](https://api.emberjs.com/ember/3.12/classes/Ember.Templates.helpers/methods?anchor=action)

### partial

While the partial deprecation is not [currently implemented](https://github.com/emberjs/rfc-tracking/issues/38), it is possible that Octane will ship with it.
Regardless, we should definitely suggest GlimmerComponents instead.

### Helper (class)

- `@observer`?

### Route

- `on('activate', function(){` + native class?
- `willTransition`, `didTransition`, `loading`, `error` actions + native class?

### DIFF
--- a/packages/@ember/-internals/glimmer/index.ts
+++ b/packages/@ember/-internals/glimmer/index.ts
@@ -128,14 +128,14 @@
   template, an optional block passed to the component should render:
 
   ```app/templates/application.hbs
-  {{#labeled-textfield value=someProperty}}
+  <LabeledTextfield @value={{@model.name}}>
     First name:
-  {{/labeled-textfield}}
+  </LabeledTextfield>
   ```
 
-  ```app/templates/components/labeled-textfield.hbs
+  ```app/components/labeled-textfield.hbs
   <label>
-    {{yield}} {{input value=value}}
+    {{yield}} <Input @value={{value}} />
   </label>
   ```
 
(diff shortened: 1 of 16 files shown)


locks:
@rwjblue no reason, I was thinking that we could use this branch to review things before sending them off to master, but we can do that off of master and spare the indirection. I'm updating the instructions and my PR in that direction.

locks:
Closing this as I'm splitting it into multiple PRs.

PR #18266, opened 2019-08-14
[BUGFIX beta] Autotrack Modifiers and Helpers

pzuraq:
This PR adds autotracking for modifiers and class based helpers.
Modifiers can opt out of autotracking via a new capabilities flag on
modifier managers.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/modifiers/custom.ts
+++ b/packages/@ember/-internals/glimmer/lib/modifiers/custom.ts
@@ -1,6 +1,8 @@
+import { track, untrack } from '@ember/-internals/metal';
 import { Factory } from '@ember/-internals/owner';
+import { assert, deprecate } from '@ember/debug';
 import { Dict, Opaque, Simple } from '@glimmer/interfaces';
-import { CONSTANT_TAG, Tag } from '@glimmer/reference';
+import { combine, CONSTANT_TAG, createUpdatableTag, Tag, update } from '@glimmer/reference';
 import { Arguments, CapturedArguments, ModifierManager } from '@glimmer/runtime';
 
 export interface CustomModifierDefinitionState<ModifierInstance> {
@@ -9,11 +11,36 @@ export interface CustomModifierDefinitionState<ModifierInstance> {
   delegate: ModifierManagerDelegate<ModifierInstance>;
 }
 
-export interface Capabilities {}
+export interface OptionalCapabilities {
+  disableAutoTracking?: boolean;
+}
 
-// Currently there are no capabilities for modifiers
-export function capabilities(_managerAPI: string, _optionalFeatures?: {}): Capabilities {
-  return {};
+export interface Capabilities {
+  disableAutoTracking: boolean;
+}
+
+export function capabilities(
+  managerAPI: string,
+  optionalFeatures: OptionalCapabilities = {}
+): Capabilities {
+  if (managerAPI !== '3.13') {
+    managerAPI = '3.13';
+
+    deprecate(
+      'Modifier manager capabilities now require you to pass a valid version when being generated. Valid versions include: 3.13',
+      false,
+      {
+        until: '3.17.0',
+        id: 'implicit-modifier-manager-capabilities',
+      }
+    );
+  }
+
+  assert('Invalid modifier manager compatibility specified', managerAPI === '3.13');
+
+  return {
+    disableAutoTracking: Boolean(optionalFeatures.disableAutoTracking),
+  };
 }
 
 export class CustomModifierDefinition<ModifierInstance> {
@@ -39,6 +66,8 @@ export class CustomModifierDefinition<ModifierInstance> {
 }
 
 export class CustomModifierState<ModifierInstance> {
+  public tag = createUpdatableTag();
+
   constructor(
     public element: Simple.Element,
     public delegate: ModifierManagerDelegate<ModifierInstance>,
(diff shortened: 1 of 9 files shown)


packages/@ember/-internals/glimmer/lib/modifiers/custom.ts
```diff
@@ -9,11 +11,24 @@ export interface CustomModifierDefinitionState<ModifierInstance> {
...
-export interface Capabilities {}
+export interface OptionalCapabilities {
+  disableLifecycleTracking?: boolean;
+}
+
+export interface Capabilities {
+  disableLifecycleTracking: boolean;
+}
 
 // Currently there are no capabilities for modifiers
```
rwjblue: This comment should be removed...

packages/@ember/-internals/glimmer/lib/modifiers/custom.ts
```diff
@@ -9,11 +11,24 @@ export interface CustomModifierDefinitionState<ModifierInstance> {
...
-export function capabilities(_managerAPI: string, _optionalFeatures?: {}): Capabilities {
-  return {};
+export function capabilities(
+  managerAPI: '3.13',
+  optionalFeatures: OptionalCapabilities
+): Capabilities {
+  assert('Invalid modifier manager compatibility specified', managerAPI === '3.13');
```
rwjblue: This is a breaking change, we have to allow previously supported values for this

packages/@ember/-internals/glimmer/lib/modifiers/custom.ts
```diff
@@ -9,11 +11,24 @@ export interface CustomModifierDefinitionState<ModifierInstance> {
...
+
+export interface Capabilities {
+  disableLifecycleTracking: boolean;
+}
 
 // Currently there are no capabilities for modifiers
-export function capabilities(_managerAPI: string, _optionalFeatures?: {}): Capabilities {
-  return {};
+export function capabilities(
+  managerAPI: '3.13',
```
rwjblue: We need to allow whatever version introduced managers, and we should probably add a deprecation for custom managers not having their capabilities set (nearly all of the managers in existence that I can find do not set capabilities at all)

packages/@ember/-internals/glimmer/lib/modifiers/custom.ts
```diff
@@ -9,11 +11,24 @@ export interface CustomModifierDefinitionState<ModifierInstance> {
...
+  optionalFeatures: OptionalCapabilities
+): Capabilities {
+  assert('Invalid modifier manager compatibility specified', managerAPI === '3.13');
+
+  return {
+    disableLifecycleTracking: Boolean(optionalFeatures.disableLifecycleTracking),
```
rwjblue: This name does not match up at all with "auto tracking" to me, "disable lifecycle tracking" in the abstract seems like it would be more about handling teardown?

Some suggestions:

* `disableTracking`
* `disableTrackedPropertyTracking`
* `disableAutoTracking`

packages/@ember/-internals/glimmer/lib/modifiers/custom.ts
```diff
@@ -109,18 +126,36 @@ class InteractiveCustomModifierManager<ModifierInstance>
...
+  getTag({ args, tag }: CustomModifierState<ModifierInstance>): Tag {
+    return combine([tag, args.tag]);
   }
 
   install(state: CustomModifierState<ModifierInstance>) {
-    let { element, args, delegate, modifier } = state;
-    delegate.installModifier(modifier, element, args.value());
+    let { element, args, delegate, modifier, tag } = state;
+
+    let tracked = track(() => {
```
rwjblue: I would think this should be guarded as well?

```js
let { capabilities } = delegate;
let shouldAutoTrack = capabilities === undefined || capabilities.disableLifecycleTracking !== true;

let tracked;

if (shouldAutoTrack) {
  tracked = track(() => {
    delegate.installModifier(modifier, element, args.value());
  });
} else {
  delegate.installModifier(modifier, element, args.value());
}

if (tracked !== undefined) {
  update(tag, tracked);
}
```

packages/@ember/-internals/glimmer/lib/modifiers/custom.ts
```diff
@@ -109,18 +126,36 @@ class InteractiveCustomModifierManager<ModifierInstance>
...
+      update(tag, tracked);
+    }
   }
 
   update(state: CustomModifierState<ModifierInstance>) {
-    let { args, delegate, modifier } = state;
-    delegate.updateModifier(modifier, args.value());
+    let { args, delegate, modifier, tag } = state;
+
+    let tracked = track(() => {
```
rwjblue: Can we pick a better name for these `tracked` variables? I find it conceptually confusing to have a variable named `tracked` that is something different than `@tracked`. Maybe `combinedTrackingTag` or `consumedTrackedPropertiesTag`?

rwjblue (commented):
Looking good, I left some inline comments / suggestions.

I think we should also make non-class based helpers auto-track, can you add that here as well?

packages/@ember/-internals/glimmer/lib/modifiers/custom.ts
pzuraq: Yeah, I thought this was a bit weird. This is inconsistent with the component manager capabilities API, which asserts it _must_ be a particular version. Happy to make the change here, but maybe we should also make the change there as well?

packages/@ember/-internals/glimmer/lib/modifiers/custom.ts
pzuraq: Wrapping it in a `track` call prevents external `track`s from watching the values. My logic was that if you disabled tracking, you probably don't want _anything_ to be able to track these computes, so we should swallow any autotracking within and throw it away. We can use the `untrack` function for this though.

packages/@ember/-internals/glimmer/lib/modifiers/custom.ts
pzuraq: Yeahhhh, naming is hard lol, but yeah I'll go with the longer name

pzuraq:
> I think we should also make non-class based helpers auto-track, can you add that here as well?
We already had tests for those: https://github.com/emberjs/ember.js/blob/98b4aa9ae4b8dbe50e80201fe40be614bc509260/packages/%40ember/-internals/glimmer/tests/integration/helpers/tracked-test.js#L12-L214

packages/@ember/-internals/glimmer/lib/modifiers/custom.ts
rwjblue: @pzuraq - No the component manager one is correct, the issue here is that we:

1) don't require modifier managers to have specified capabilities (just a bug in initial implementation)
2) any that did specify capabilities could have passed any arbitrary thing for the first argument

We should fix this mistake by doing:

* Adding a deprecation when we encounter a manager without capabilities set
* Deprecate calling this method (`capabilities`) with a value other than 3.13 (for the new flag)

packages/@ember/-internals/glimmer/lib/modifiers/custom.ts
rwjblue: > you probably don't want anything to be able to track these computes, so we should swallow any autotracking within and throw it away

Hmm, I see. I think both are valid paths forward, definitely up to you here. Though if you don't mind, could you add a small inline comment about throwing away the tracker in the case where you have explicitly opted out?

rwjblue:
> We already had tests for those:

Hmm, maybe I am mis-reading those but they do not look like what I was thinking. I mean specifically a test where a non-class based helper accesses a tracked value within its computation function. All of the tests that you linked seemed to be asserting that the helper will be reevaluated if an input to it was derived from a tracked property.

```js
      '@test asdfasdfasdfasdfasdfasdfasdf'(assert) {
        let computeCount = 0;

		let currentUserService;
        this.registerService('current-user', Service.extend({
          name: tracked({ value: 'bob' }),

          init() {
            this._super(...arguments);
            currentUserService = this;
          }
        });

        this.registerComponent('person', {
            currentUser: inject('current-user')
          }),

          template: strip`
            {{hello-world this.currentUser}}
          `,
        });

        this.registerHelper('hello-world', ([service]) => {
          computeCount++;
          return service.name;
        });

        this.render('<Person/>');

        this.assertText('bob-value');

        assert.strictEqual(computeCount, 1, 'compute is called exactly 1 time');

        runTask(() => this.rerender());

        this.assertText('bob-value');

        assert.strictEqual(computeCount, 1, 'compute is called exactly 1 time');

        runTask(() => currentUserService.name = 'sal');

        this.assertText('sal-value');

        assert.strictEqual(computeCount, 2, 'compute is called exactly 2 times');
      }
```

pzuraq:
ah, gotcha, yeah that makes sense to add, will do 👍

pzuraq:
Alright, updated based on feedback, plus I added https://github.com/ember-cli/ember-rfc176-data/pull/166 to actually expose the modifier manager capabilities generator

rwjblue (approved):

Issue #18260, opened 2019-08-14
Router#url should not error when `location` is a string


rwjblue:
This function should check if `location` is a string, and avoid calling `.getURL()` on it if it is. `location` on `Ember.Router` **starts** as a string, and is only converted to an object when routing itself is initialized. In many tests routing is _never_ initialized, yet we unconditionally call `.getURL()` on the string (usually `'none'.getURL()`).

For context, this is mostly only an issue because when you `assert.strictEqual(someEmberObject, someOtherObject)` we happen to try to print out the diff of the **entire `owner` object** 😩. We should fix that separately, but in the meantime this method should not have the assumption that `location` is an object with `getURL` on it:

https://github.com/emberjs/ember.js/blob/5c723255c17326ceda374992f5c340965ea245e6/packages/%40ember/-internals/routing/lib/system/router.ts#L1793-L1795

I'd love it if someone were to tweak this to _something_ like:

```
  url: computed(function(this: Router<Route>) {
    let { location } = this;
    if (typeof location !== 'string') {
      return location.getURL();
    }
  }),
```

And then add a unit test that instantiates the router and calls `router.url` and confirms it does not throw an error.

ryanolsonx:
I'd love to pick up this one.

ryanolsonx:
Can we close this since the PR has been merged?

rwjblue:
Yep, thank you!!

PR #18267, opened 2019-08-14
[BUGFIX beta] Router#url should not error when `location` is a string

ryanolsonx:
See: #18260

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/router.ts
+++ b/packages/@ember/-internals/routing/lib/system/router.ts
@@ -1791,7 +1791,13 @@ EmberRouter.reopen(Evented, {
    @private
  */
   url: computed(function(this: Router<Route>) {
-    return get(this, 'location').getURL();
+    let location = get(this, 'location');
+
+    if (typeof location === 'string') {
+      return undefined;
+    }
+
+    return location.getURL();
   }),
 });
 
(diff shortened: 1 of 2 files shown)


packages/@ember/-internals/routing/lib/system/router.ts
```diff
@@ -1791,7 +1791,13 @@ EmberRouter.reopen(Evented, {
    @private
  */
   url: computed(function(this: Router<Route>) {
-    return get(this, 'location').getURL();
+    let location = get(this, 'location');
+
+    if (typeof location !== 'string') {
```
btecu: Could avoid the negation by swapping the returns

ryanolsonx: That's a good idea.  I'll do that.

rwjblue (approved):
Awesome, thank you!

rwjblue:
Hmm, I've restarted this CI job a couple of times and keep seeing the same failure.

rwjblue:
```
not ok 25 IE 11.0 - [35 ms] - {{on}} Modifier:  asserts if the provided callback accesses `this` without being bound prior to passing to on
    ---
        actual: >
            [object HTMLButtonElement]
        expected: >
            null
        stack: >
               at myFunc (http://127.0.0.1:7774/200750294740/dist/tests/ember-tests.js:41220:13)
        message: >
            this is null on browsers without native proxy support
        negative: >
            false
        browser log: |
    ...
```

rwjblue:
I can't figure out how that failure could be related to the changes here. 🤔

I think we need to get a test run in an actual IE11 box to see whats really going on (why does it pass on master, why does this seemingly unrelated change make it fail, etc).

ryanolsonx:
Yeah - that's pretty weird.  I rebased on master and hopefully things just pass this next time around.

rwjblue:
Yep, that fixed things up for us. Thank you again @ryanolsonx!

PR #18269, opened 2019-08-14
[BUGFIX]  Fix #18268 query param options not being read

### LINKED ISSUE #18268: [BUG] Nested query params don't work with refresh model
If you have
```
queryParams: {
  'nested.foo': {
    // By default, controller query param properties don't
    // cause a full transition when they are changed, but
    // rather only cause the URL to update. Setting
    // `refreshModel` to true will cause an "in-place"
    // transition to occur, whereby the model hooks for
    // this route (and any child routes) will re-fire, allowing
    // you to reload models (e.g., from the server) using the
    // updated query param values.
    refreshModel: true,

    // By default, the query param URL key is the same name as
    // the controller property name. Use `as` to specify a
    // different URL key.
    as: 'foobar'
  }
},
```

`_optionsForQueryParam` fails to work because it uses `get(this, `queryParams.${qp.urlKey}`)` when really the vaule is `queryParams[qp.urlKey]`. Same idea for props.

So with the current implementation of `_optionsForQueryParam` the nested query param doesn't work. And as a consequence when QPs change the Route can't determine if that QP is marked as refreshModel.

Example (with rudimentary fix):

https://ember-twiddle.com/0a1917cff72d5720d69e0382276e6722?openFiles=routes.application.js%2C

EWhite613:
Add fix for when queryParams are using a nested value and _optionsForQueryParam not getting the queryParam setting

Fixes #18268.


ef4 (approved):

ef4:
Thanks for this. Better late than never.

PR #18272, opened 2019-08-16
[DOCS] Add code example for EmberArray.addArrayObserver

sdebarros:
Part of https://github.com/emberjs/ember.js/issues/18228

### DIFF
--- a/packages/@ember/-internals/runtime/lib/mixins/array.js
+++ b/packages/@ember/-internals/runtime/lib/mixins/array.js
@@ -452,6 +452,29 @@ const ArrayMixin = Mixin.create(Enumerable, {
       `willChange` and `didChange` option.
     @return {EmberArray} receiver
     @public
+    @example
+        import Service from '@ember/service';
+
+        export default Service.extend({
+          data: Ember.A(),
+
+          init() {
+            this._super(...arguments);
+
+            this.data.addArrayObserver(this, {
+              willChange: 'dataWillChange',
+              didChange: 'dataDidChange'
+            });
+          },
+
+          dataWillChange(array, start, removeCount, addCount) {
+            console.log('array will change', array, start, removeCount, addCount);
+          },
+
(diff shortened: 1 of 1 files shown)


packages/@ember/-internals/runtime/lib/mixins/array.js
```diff
@@ -446,6 +446,32 @@ const ArrayMixin = Mixin.create(Enumerable, {
     which you can use to override the method names that will be invoked on the
     target.
 
+    Example:
```
locks: YUIDoc supports an `@example` annotation with a description https://yui.github.io/yuidoc/syntax/index.html. Could you update for that?

locks (changes requested):

packages/@ember/-internals/runtime/lib/mixins/array.js
sdebarros: I've moved the example to below an `@example` tag, let me know if the position is ok.

sivakumar-kailasam:
I tried this PR locally & YUIDoc doesn't seem to recognize the `@example` decorator in this file but the same works on  [packages/@ember/-internals/metal/lib/tracked.ts#L55](https://github.com/emberjs/ember.js/blob/master/packages/@ember/-internals/metal/lib/tracked.ts#L55). Maybe for now we can stick to adding it to the end of the description.

sdebarros:
@sivakumar-kailasam I ran `yarn run docs` and this is how I see the docs:
Screen Shot 2019-08-29 at 18 26 38

I couldn't find the docs for the file you mentioned. Do you mind letting me know how you are testing it?

The difference I see is that in that file `@example` doesn't have something nested, and the code is actually wrapped with backticks, so I think it's not being formatted by `@example` but the regular way.

sivakumar-kailasam (approved):
Tried it again & it works as expected now. I must've messed up something the last time I tried this. Thanks again for this PR 🙂

PR #18274, opened 2019-08-16
[BUGFIX beta] Allow CPs to depend on nested args

pzuraq:
CP chaining on nested args was not setup properly in the last couple of
merges. This fixes and adds a test case.

### DIFF
--- a/packages/@ember/-internals/metal/lib/chain-tags.ts
+++ b/packages/@ember/-internals/metal/lib/chain-tags.ts
@@ -49,6 +49,7 @@ export function getChainTagsForKey(obj: any, path: string) {
 
   let current: any = obj;
 
+  let pathLength = path.length;
   let segmentEnd = -1;
   // prevent closures
   let segment: string, descriptor: any;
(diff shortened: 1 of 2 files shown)


packages/@ember/-internals/metal/lib/chain-tags.ts
```diff
@@ -104,20 +105,28 @@ export function getChainTagsForKey(obj: any, path: string) {
...
       let namedArgs = ARGS_PROXY_TAGS.get(current.args);
       let ref = namedArgs.get(segment);
 
       chainTags.push(ref.tag);
 
-      if (segmentEnd !== -1) {
+      if (segmentEnd !== pathLength) {
         current = ref.value();
-        segment = path.substr(segmentEnd + 1);
+        // lastSegmentEnd = segmentEnd + 1;
```
rwjblue: Hmm, do we need these comments? Should we remove them?

pzuraq: ah, that was a mistake, definitely

rwjblue (approved):

PR #18281, opened 2019-08-17
[BUGFIX lts] Check length of targets

RobbieTheWagner:
We had a bug where if you had additional targets added, they would not be checked, we were just checking that all the prebuilt targets were included. This adds a check for length to ensure all the targets are there and no additional ones are.


pzuraq (approved):

pzuraq:
Thanks!!

rwjblue:
Good catch, thank you!

PR #18283, opened 2019-08-18
[DOC] Add examples for EmberArray#invoke

mukilane:
For #18228

PR #18292, opened 2019-08-21
[DOC] Updated EmberArray#invoke example

mukilane:
Follow up of #18283 
Updated to include example without any arguments.


locks (approved):
Thanks for the follow-up!

PR #18296, opened 2019-08-23
[BUGFIX lts] Ensure {{each-in}} can iterate over keys with periods

pzuraq:
This PR refactors the iterator in `each-in` to use standard object get
notation for getting values instead of `Ember.get`. This should work
because:

* All descriptors now install native getters
* Object iterators use `Object.keys` to get the list of keys to iterate
* `unknownProperty` is only ever triggered if a key is not `in` the
  object, which cannot be true if we just got the key using
  `Object.keys`. Thus `unknownProperty` will never be called for
  `each-in`.

This will allow users to use arbitrary keys for their objects, and
iterate over them with `each-in`. Keys will be tracked as if `get` were
used.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/utils/iterable.ts
+++ b/packages/@ember/-internals/glimmer/lib/utils/iterable.ts
@@ -1,6 +1,14 @@
-import { get, objectAt, tagFor, tagForProperty } from '@ember/-internals/metal';
+import {
+  consume,
+  get,
+  isTracking,
+  objectAt,
+  tagFor,
+  tagForProperty,
+} from '@ember/-internals/metal';
 import { _contentFor } from '@ember/-internals/runtime';
 import { guidFor, HAS_NATIVE_SYMBOL, isEmberArray, isProxy } from '@ember/-internals/utils';
+import { EMBER_METAL_TRACKED_PROPERTIES } from '@ember/canary-features';
 import { assert } from '@ember/debug';
 import {
   AbstractIterable,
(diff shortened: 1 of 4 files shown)


pzuraq:
Fixes #17529

packages/@ember/-internals/glimmer/lib/utils/iterable.ts
```diff
@@ -127,7 +135,22 @@ class ObjectIterator extends BoundedIterator {
...
       let values: Opaque[] = [];
       for (let i = 0; i < length; i++) {
-        values.push(get(obj, keys[i]));
+        let value: any;
+        let key = keys[i];
+
+        value = obj[key];
+
+        // Add the tag of the returned value if it is an array, since arrays
+        // should always cause updates if they are consumed and then changed
```
chancancode: Is that right? If you pass this to, e.g. `#each`, we handle this there already. If you just pass the value to, say, a component or a helper, I am not sure we need to invalidate those if the content of the array changes?

chancancode: I think the test you added should pass without consuming `[]`.

packages/@ember/-internals/glimmer/lib/utils/iterable.ts
```diff
@@ -127,7 +135,22 @@ class ObjectIterator extends BoundedIterator {
     } else {
       let values: Opaque[] = [];
       for (let i = 0; i < length; i++) {
-        values.push(get(obj, keys[i]));
+        let value: any;
+        let key = keys[i];
+
+        value = obj[key];
```
chancancode: This breaks when `obj` is a proxy, is that okay?

In general, if the solution requires avoiding `get` semantics, I think that is not okay. However, maybe in this case it is fine since we already did `Object.keys` above so it won't trigger `unknownProperty` anyway.

pzuraq: Correct, I started writing the code to do `unknownProperty` and realized that by definition, it could never run, which is why I omitted it.

packages/@ember/-internals/glimmer/lib/utils/iterable.ts
pzuraq: This is part of autotracking as a whole, it allows consumers to use arrays (both native and non-native) anywhere they consume a tracked value, and have mutations to that array be autotracked. In practice, this is like the native getters migration, it’ll mean users can access any array as if it were a normal, native array. They only need to use methods for updating the state of the array (we need native Proxy to allow them to update with native syntax).

Even when we do get to native Proxy though, this will allow us to omit a `get` handler in the Proxy, which should help us avoid perf issues with proxies

PR #18304, opened 2019-08-23
[BUGFIX lts] Check the EMBER_ENV environment variable after it is set

### LINKED ISSUE #18294: [beta] debug macros work incorrectly in 3.13.0-beta.3
Reproduction:

```sh
npx ember-cli new macro-test # will install 3.12
cd macro-test
$(npm bin)/ember build --environment production
grep "could not find a suitable method to bind" dist/assets/vendor-*.js
# note there's no output

npm install --save ember-source@3.13.0-beta.3
$(npm bin)/ember build --environment production
grep "could not find a suitable method to bind" dist/assets/vendor-*.js
# this will find the assert
```
[The assert in question](https://github.com/emberjs/ember.js/blob/v3.13.0-beta.3/packages/%40ember/runloop/index.js#L226) should be stripped from the build.

For whatever reason, leaving this assert in causes my app to fail to load in production, even though this test app works just fine.

cc @rwjblue

pzuraq:
The EMBER_ENV variable doesn't get set until addon
hooks actually start running, so we weren't correctly
determining the environment.

Fixes #18294

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -9,8 +9,6 @@ const buildDebugMacroPlugin = require('./build-debug-macro-plugin');
 const buildStripClassCallcheckPlugin = require('./build-strip-class-callcheck-plugin');
 const injectBabelHelpers = require('./transforms/inject-babel-helpers');
 
-const isProduction = process.env.EMBER_ENV === 'production';
-
 const PRE_BUILT_TARGETS = [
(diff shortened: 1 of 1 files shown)


pzuraq:
Fixes #18294

rwjblue (approved):

PR #18314, opened 2019-08-26
[BUGFIX beta] Use class inheritance for getters and setters

pzuraq:
Currently, `get` and `set` will bypass class inheritance and lookup
descriptors for CPs. If they exist, they call them directly. This
results in bugs where plain, undecorated getters and setters cannot
override CPs on parent classes.

This fix converts us to using class inheritance, looking up descriptors
recursively on the class to find the actual property that is meant to be
accessed for this instance. In the `get` case, this is simple removing
the bypass altogether. For the `set` case, this would result in
behavioral differences:

* An additional `get` triggered just before setting, since `set` gets
the previous value in order to know if it should notify.
* An additional property change notification for CPs, since they must
notify themselves (when set using a native setter, for instance).

Instead, this PR looks up the descriptor and checks to see if it's a
`CPSETTER` function, and triggers it directly if so.

As an alternative, we could put all CPSETTER functions directly into a `WeakSet` and check to see if the setter function exists in that set, rather than checking its `toString()`. I was concerned about perf in doing that, but it would be a bit safer.

### DIFF
--- a/packages/@ember/-internals/metal/lib/computed.ts
+++ b/packages/@ember/-internals/metal/lib/computed.ts
@@ -630,6 +630,8 @@ export class ComputedProperty extends ComputedDescriptor {
       } finally {
         endPropertyChanges();
       }
+
+      return ret;
     } else {
       return this.setWithSuspend(obj, keyName, value);
     }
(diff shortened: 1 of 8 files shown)


rwjblue (approved):

dfreeman:
Agreed with everything @rwjblue said about this being a weird situation that should be avoided, but we just chased down a bug in one of our apps came down to a superclass computed winning over a native getter in a subclass 😬 

Thanks for getting this squared away!

PR #18336, opened 2019-08-30
[DOC] Octanify router service documentation

locks:
Part of https://github.com/emberjs/ember.js/issues/18250.

### DIFF
--- a/packages/@ember/-internals/routing/lib/services/router.ts
+++ b/packages/@ember/-internals/routing/lib/services/router.ts
@@ -185,7 +185,7 @@ export default class RouterService extends Service {
       @service router;
       @service clipboard;
 
-      // Provided` in the template
+      // Provided in the template
       // { id: 'tomster', name: 'Tomster' }
       author: null,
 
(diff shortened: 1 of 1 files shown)


packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -177,14 +178,14 @@ export default class RouterService extends Service {
...
 
-    export default Component.extend({
-      router: service('router'),
-      clipboard: service('clipboard')
+    export default class CopyLinkComponent extends Component {
+      @service router;
+      @service clipboard;
 
-      // Provided in the template
+      // Provided` in the template
```
rwjblue:
```suggestion
      // Provided in the template
```

packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -195,19 +196,18 @@ export default class RouterService extends Service {
           // Clipboard now has /author/tomster/books
         }
       }
-    });
-    ```
```
rwjblue: Probably shouldn't be dropped (I think it will make the next prose section unhappy)

locks: Fixed, thanks!

rwjblue (approved):

PR #18338, opened 2019-09-01
[DOC] Octanify Route documentation

locks:
Part of #18336.

## TODO

- [x] confirm `templateName` in native classes
- [ ] `model`, async or not? any others?

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/route.ts
+++ b/packages/@ember/-internals/routing/lib/system/route.ts
@@ -1378,7 +1378,7 @@ class Route extends EmberObject implements IRoute {
     import Route from '@ember/routing/route';
 
     export default class PostCommentsRoute extends Route {
-      async model() {
+      model() {
         let post = this.modelFor('post');
 
         return post.comments;
@@ -2035,7 +2035,7 @@ function getEngineRouteName(engine: Owner, routeName: string) {
     import Route from '@ember/routing/route';
 
     export default class PostRoute extends Route {
-      async model({ post_id }) {
+      model({ post_id }) {
         // the server returns `{ id: 12 }`
         return fetch(`/posts/${post_id}`;
       }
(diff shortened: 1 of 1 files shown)


rwjblue:
> confirm templateName in native classes

Ya, `templateName` can be used in native classes. IMHO, its _generally_ something we should deprecate and remove, but it is still supported (and should work fine as a class field).

rwjblue:
> model, async or not? any others?

I would leave this choice to be made separately (it isn't really related to octane). We should probably make an issue for broader discussion (plausibly a larger core team discussion), and make a final conclusion here. My personal preference would be to make it async though.

packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -383,17 +383,17 @@ class Route extends EmberObject implements IRoute {
...
-        willTransition(transition) {
-          if (this.controller.get('userHasEnteredData')) {
-            this.controller.displayNavigationConfirm();
-            transition.abort();
-          }
+    import { action } from '@ember/object';
+
+    export default class ContactFormRoute extends Route {
+      @action
+      willTransition(transition) {
```
rwjblue: I'm slightly concerned about this one. Specifically, `Route` has both an action _and_ a method named `willTransition`. The `@action` decorator will result in both with be populated (with the same function). 

@pzuraq - Will this result in a loop?

locks:
@rwjblue I've left `async` out of the code samples pending a core team discussion.

packages/@ember/-internals/routing/lib/system/route.ts
pzuraq: I don't think it would result in a loop, it'll just be that the function gets called from two places instead of one, unless the `willTransition` function also calls the action

rwjblue:
kk, we should double check, but I don't think it does

packages/@ember/-internals/routing/lib/system/route.ts
pzuraq: I'm actually not seeing that the method exists, it appears to just be the action/event.

rwjblue (approved):

PR #18345, opened 2019-09-02
Implement RFC #523 (@model argument for route templates and {{mount}})

chancancode:
This patch implements emberjs/rfcs#523.

See the commit messages for more details.


packages/@ember/-internals/glimmer/lib/syntax/outlet.ts
```diff
@@ -85,17 +92,94 @@ export function outletMacro(
...
+          tag,
+          map,
+          names: ['model'],
+          references: [modelRef],
+          length: 1,
+          has(key: string): boolean {
+            return key === 'model';
+          },
+          get<T extends VersionedPathReference>(key: string): T {
+            assert('key !== model', key === 'model');
```
rwjblue: Am I correct in assuming that this asserstion is what would be thrown when you do `@foo` in a route's template? If so, it should definitely be made a bit more user friendly...

packages/@ember/-internals/glimmer/lib/syntax/outlet.ts
```diff
@@ -138,6 +222,7 @@ function stateFor(
     outlet: render.outlet,
     template,
     controller: render.controller,
+    model: render.model,
```
rwjblue: Just noting, that changing this structure will likely require changes in `@ember/test-helpers`:

* https://github.com/emberjs/ember-test-helpers/blob/fdcf31ab04b908e4561eeea102fc07a6f483d640/addon-test-support/%40ember/test-helpers/setup-rendering-context.ts#L115-L141
* https://github.com/emberjs/ember-test-helpers/blob/fdcf31ab04b908e4561eeea102fc07a6f483d640/addon-test-support/ember-test-helpers/legacy-0-6-x/test-module-for-component.js#L227-L239

packages/@ember/-internals/glimmer/lib/views/outlet.ts
```diff
@@ -57,6 +57,7 @@ export default class OutletView {
         outlet: TOP_LEVEL_OUTLET,
         name: TOP_LEVEL_NAME,
         controller: undefined,
+        model: undefined,
```
rwjblue: FWIW, I'd prefer to make this more generic in the internals. As mentioned above, `@ember/test-helpers` uses outlet state to support rendering tests, and it would be quite nice to allow them to use named arguments (instead of requiring `this.foo` from the test context). It seems that making this structure a tad more generic (but preventing routes from using it) would allow that.

packages/@ember/-internals/glimmer/tests/integration/application/engine-test.js
```diff
@@ -517,7 +518,12 @@ moduleFor(
             },
           })
         );
-        this.register('template:application_error', compile('Error! {{model.message}}'));
+        this.register(
+          'template:application_error',
+          compile(
+            EMBER_ROUTING_MODEL_ARG ? 'Error! {{@model.message}}' : 'Error! {{this.model.message}}'
```
rwjblue: This change seems fine in the abstract, but we need to ensure that at least _some_ tests remain confirming all three scenarios work properly:

* implicit fallback
* `this.model`
* `@model`

packages/@ember/-internals/glimmer/tests/integration/application/rendering-test.js
```diff
@@ -38,7 +38,7 @@ moduleFor(
       });
     }
 
-    ['@test it can access the model provided by the route']() {
+    ['@feature(!EMBER_ROUTING_MODEL_ARG) it can access the model provided by the route']() {
```
rwjblue: Why would this one only run when the feature is disabled? IT seems that this test should work either way to me (`this.model` is compatible with both before and after, and is not deprecated).

packages/@ember/-internals/glimmer/tests/integration/application/rendering-test.js
```diff
@@ -51,12 +51,70 @@ moduleFor(
...
+            return ['red', 'yellow', 'blue'];
+          },
+        })
+      );
+
+      this.addTemplate(
+        'application',
+        strip`
+        <ul>
+          {{#each @model as |item|}}
```
rwjblue: Do we expect using a named arg that is not provided to error or anything? Seems like we don't, right?

packages/@ember/-internals/glimmer/tests/integration/application/rendering-test.js
```diff
@@ -70,7 +128,7 @@ moduleFor(
       });
     }
 
-    ['@test it can render a nested route']() {
+    ['@feature(!EMBER_ROUTING_MODEL_ARG) it can render a nested route']() {
```
rwjblue: IMHO, this shouldn't be feature flagged. This test confirms `this.model` continues to work, and should stay around even after the feature flag is removed.

packages/@ember/-internals/glimmer/tests/integration/application/rendering-test.js
```diff
@@ -92,43 +150,84 @@ moduleFor(
...
+            <li>red</li>
+            <li>yellow</li>
+            <li>blue</li>
+          </ul>
+        `);
       });
     }
 
-    ['@test it can render into named outlets']() {
+    ['@feature(!EMBER_ROUTING_MODEL_ARG) it can render into named outlets']() {
```
rwjblue: Same as above, this test should pass both with and without the flag.

packages/@ember/-internals/glimmer/tests/integration/application/rendering-test.js
```diff
@@ -280,7 +379,7 @@ moduleFor(
         });
     }
 
-    ['@test it should produce a stable DOM when the model changes']() {
+    ['@feature(!EMBER_ROUTING_MODEL_ARG) it should produce a stable DOM when the model changes']() {
```
rwjblue: ditto

packages/@ember/-internals/glimmer/tests/integration/application/rendering-test.js
```diff
@@ -339,7 +466,7 @@ moduleFor(
         .then(() => this.assertText('b'));
     }
 
-    ['@test it should update correctly when the controller changes']() {
+    ['@feature(!EMBER_ROUTING_MODEL_ARG) it should update correctly when the controller changes']() {
```
rwjblue: (gonna stop commenting on these now, it seems the remainder of the file will need the same general tweaks/updates that I mentioned from above)

packages/@ember/-internals/glimmer/tests/integration/components/link-to/query-params-angle-test.js
```diff
@@ -235,7 +235,7 @@ moduleFor(
       });
     }
 
-    ['@test generates proper href for `LinkTo` with no @route after transitioning to an error route GH#17963'](
+    ['@feature(!EMBER_ROUTING_MODEL_ARG) generates proper href for `LinkTo` with no @route after transitioning to an error route GH#17963'](
```
rwjblue: This test should not be flagged.

packages/@ember/-internals/glimmer/tests/integration/components/link-to/routing-angle-test.js
```diff
@@ -916,7 +916,9 @@ moduleFor(
         });
     }
 
-    [`@test The <LinkTo /> component moves into the named route with context`](assert) {
+    [`@feature(!EMBER_ROUTING_MODEL_ARG) The <LinkTo /> component moves into the named route with context`](
```
rwjblue: Test should run both before and after the feature.

packages/@ember/-internals/glimmer/tests/integration/components/link-to/routing-angle-test.js
```diff
@@ -1591,7 +1676,7 @@ moduleFor(
         });
     }
 
-    ['@test [GH#17018] passing model to <LinkTo /> with `hash` helper works']() {
+    ['@feature(!EMBER_ROUTING_MODEL_ARG) [GH#17018] passing model to <LinkTo /> with `hash` helper works']() {
```
rwjblue: 👆

packages/@ember/-internals/glimmer/tests/integration/components/link-to/routing-curly-test.js
```diff
@@ -916,7 +916,9 @@ moduleFor(
         });
     }
 
-    [`@test The {{link-to}} component moves into the named route with context`](assert) {
+    [`@feature(!EMBER_ROUTING_MODEL_ARG) The {{link-to}} component moves into the named route with context`](
```
rwjblue: 😈

packages/@ember/-internals/glimmer/tests/integration/components/link-to/routing-curly-test.js
```diff
@@ -1539,7 +1624,7 @@ moduleFor(
         });
     }
 
-    [`@test The non-block form {{link-to}} component moves into the named route with context`](
+    [`@feature(!EMBER_ROUTING_MODEL_ARG) The non-block form {{link-to}} component moves into the named route with context`](
```
rwjblue: 🔪

packages/@ember/-internals/glimmer/tests/integration/components/link-to/routing-curly-test.js
```diff
@@ -1865,7 +2008,7 @@ moduleFor(
         });
     }
 
-    ['@test [GH#17018] passing model to link-to with `hash` helper works']() {
+    ['@feature(!EMBER_ROUTING_MODEL_ARG) [GH#17018] passing model to link-to with `hash` helper works']() {
```
rwjblue: 🙃

packages/@ember/-internals/metal/lib/transaction.ts
```diff
@@ -86,22 +86,16 @@ if (DEBUG) {
...
 
-          if (lastRef !== undefined) {
-            while (lastRef && lastRef.propertyKey) {
-              parts.unshift(lastRef.propertyKey);
-              lastRef = lastRef.parentReference;
-            }
-
-            label = parts.join('.');
+          if (lastRef && typeof lastRef.debug === 'function') {
+            label = `as \`${lastRef.debug()}\` in ${lastRenderedIn}`;
```
rwjblue: This is great, thank you for doing it!

packages/@ember/application/tests/visit_test.js
```diff
@@ -606,7 +606,7 @@ moduleFor(
       });
     }
 
-    [`@test Ember Islands-style setup`](assert) {
+    [`@feature(!EMBER_ROUTING_MODEL_ARG) Ember Islands-style setup`](assert) {
```
rwjblue: 🏓

packages/ember/tests/homepage_example_test.js
```diff
@@ -7,11 +7,11 @@ import { moduleFor, ApplicationTestCase } from 'internal-test-helpers';
 moduleFor(
   'The example renders correctly',
   class extends ApplicationTestCase {
-    ['@test Render index template into application outlet'](assert) {
+    ['@feature(!EMBER_ROUTING_MODEL_ARG) Render index template into application outlet'](assert) {
```
rwjblue: 🔀

packages/ember/tests/routing/decoupled_basic_test.js
```diff
@@ -260,15 +256,15 @@ moduleFor(
...
+      await this.visit('/');
 
-        assert.equal(text, 'YES I AM HOME', 'The homepage template was rendered');
-      });
+      assert.equal(this.$('p').text(), 'YES I AM HOME', 'The homepage template was rendered');
     }
 
-    [`@test Model passed via renderTemplate model is set as controller's model`](assert) {
-      this.addTemplate('bio', '<p>{{model.name}}</p>');
+    [`@feature(!EMBER_ROUTING_MODEL_ARG) Model passed via renderTemplate model is set as controller's model`](
```
rwjblue: 🤠

packages/ember/tests/routing/decoupled_basic_test.js
```diff
@@ -641,7 +670,35 @@ moduleFor(
       });
     }
 
-    [`@test The Specials Page getting its controller context by deserializing the params hash`](
+    [`@feature(!EMBER_ROUTING_MODEL_ARG) The Specials Page getting its controller context by deserializing the params hash`](
```
rwjblue: 🔕

packages/ember/tests/routing/router_service_test/transitionTo_test.js
```diff
@@ -157,14 +157,14 @@ moduleFor(
       });
     }
 
-    ['@test RouterService#transitionTo with dynamic segment'](assert) {
+    ['@feature(!EMBER_ROUTING_MODEL_ARG) RouterService#transitionTo with dynamic segment'](assert) {
```
rwjblue: 🌀

packages/ember/tests/routing/substates_test.js
```diff
@@ -381,10 +381,10 @@ moduleFor(
       return promise;
     }
 
-    ['@test Prioritized error substate entry works with preserved-namespaec nested routes'](
+    ['@feature(!EMBER_ROUTING_MODEL_ARG) Prioritized error substate entry works with preserved-namespaec nested routes'](
```
rwjblue:
```suggestion
    ['@feature(!EMBER_ROUTING_MODEL_ARG) Prioritized error substate entry works with preserved-namespace nested routes'](
```

rwjblue:
Most of my comments are around the idea of not testing `this.model` after the feature lands (which is not good). I _may_ have gotten carried away (🙃), but in general it seems largely fine if the goal is to:

* optimize for the non-feature tests to be completely deleted, and therefore doing more work upfront (by effectively duplicating all of the tests that use `{{model}}`)
* ensure there are _at least_ a few of each test type for `this.model` (to ensure we don't regress functionality)

packages/@ember/-internals/glimmer/lib/syntax/outlet.ts
chancancode: Fixed, it is actually supposed to return `UNDEFINED_REFERENCE`

packages/@ember/-internals/glimmer/lib/syntax/outlet.ts
chancancode: model is allowed to be `undefined`, so it shouldn't cause any errors if it's not present

packages/@ember/-internals/glimmer/lib/views/outlet.ts
chancancode: seems like a good future improvement 😄

(It is not _exactly_ trivial to translate this to be more generic either, due to the way things are implemented)

packages/@ember/-internals/glimmer/tests/integration/application/engine-test.js
chancancode: Addressed in the last commit

packages/@ember/-internals/glimmer/tests/integration/application/rendering-test.js
chancancode: correct, we do not, it would just be undefined

packages/@ember/-internals/glimmer/tests/integration/application/rendering-test.js
chancancode: I think we said we are fine testing the happy path (`@model` going forward) in these tests (there is a copy of the test below in the new style) and just add some targeted tests to avoid regression, which I did in the latest commit, if you still think this or any other tests specifically  need to be test both ways, feel free to comment again.

packages/@ember/-internals/glimmer/lib/views/outlet.ts
rwjblue: Ya, totally agree this isn't a blocker, but I would love to discuss this sometime with you (hopefully while the issue is still somewhat fresh in your mind).

packages/@ember/-internals/glimmer/tests/integration/application/rendering-test.js
rwjblue: Nope, this is great, thank you!

rwjblue (approved):

PR #18347, opened 2019-09-02
[DOC] Octanify helper API documentation

locks:
- [ ] The use of observer in the Helper class needs an explanation as to why it requires the classic syntax


rwjblue (approved):

rwjblue:
@locks we can track that remaining item in the meta issue

PR #18353, opened 2019-09-04
[BUGFIX beta] improve `fn` & `on` undefined callback message

patricklx:
include parent and property key in the message

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/helpers/fn-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/helpers/fn-test.js
@@ -127,7 +127,7 @@ moduleFor(
           arg1: 'foo',
           arg2: 'bar',
         });
-      }, /You must pass a function as the `fn` helpers first argument, you passed null/);
+      }, /You must pass a function as the `fn` helpers first argument, you passed this.myFunc to `fn` but it was null/);
     }
 
     '@test asserts if the provided function accesses `this` without being bound prior to passing to fn'(
--- a/packages/@ember/-internals/glimmer/lib/helpers/fn.ts
+++ b/packages/@ember/-internals/glimmer/lib/helpers/fn.ts
@@ -86,8 +86,9 @@ function fnHelper({ positional }: ICapturedArguments) {
   if (DEBUG && typeof callbackRef[INVOKE] !== 'function') {
     let callback = callbackRef.value();
 
+    const debug = (<any>callbackRef).debug && (<any>callbackRef).debug();
     assert(
-      `You must pass a function as the \`fn\` helpers first argument, you passed ${callback}`,
+      `You must pass a function as the \`fn\` helpers first argument, you passed ${debug} to \`fn\` but it was ${callback}`,
       typeof callback === 'function'
     );
   }
--- a/packages/@ember/-internals/glimmer/lib/modifiers/on.ts
+++ b/packages/@ember/-internals/glimmer/lib/modifiers/on.ts
@@ -101,11 +101,15 @@ export class OnModifierState {
       this.eventName = eventName;
       this.shouldUpdate = true;
     }
+    if (DEBUG) {
+      const debug = args.positional.at(1) && (<any>args.positional.at(1)).debug();
+      const value = args.positional.at(1) && args.positional.at(1).value();
+      assert(
+        `You must pass a function as the second argument to the \`on\` modifier, you passed ${debug} to \`on\` but it was ${value}`,
+        value !== undefined && typeof value === 'function'
+      );
+    }
 
-    assert(
-      'You must pass a function as the second argument to the `on` modifier',
-      args.positional.at(1) !== undefined && typeof args.positional.at(1).value() === 'function'
-    );
     let userProvidedCallback = args.positional.at(1).value() as EventListener;
     if (userProvidedCallback !== this.userProvidedCallback) {
       this.userProvidedCallback = userProvidedCallback;


rwjblue (commented):
Generally seems good, I think that CI is failing due to some testing the exact assertion message. Also, in a recent PR @chancancode added a `ref.debug` method that we _might_ be able to leverage here instead of manually digging into callbackRef.propertyKey.

patricklx:
Hi, I was still not quite happy with the `debug` function since it gave me not eough information on where the error is.
But I found there is a debug stack for the templates, So I add that now to the error message.
So now we would have a template stack + helper debug message.

packages/@ember/-internals/glimmer/tests/integration/helpers/fn-test.js
```diff
@@ -127,7 +127,7 @@ moduleFor(
           arg1: 'foo',
           arg2: 'bar',
         });
-      }, /You must pass a function as the `fn` helpers first argument, you passed null/);
+      }, /You must pass a function as the `fn` helpers first argument, you passed null from this.myFunc/);
```
pzuraq: I think a slight tweak to the wording here could be "you passed `this.myFunc` to `fn` but it was null", would be more clear what the user had passed to the helper

packages/@ember/-internals/glimmer/lib/renderer.ts
```diff
@@ -95,27 +95,39 @@ class RootState {
...
-      do {
-        iteratorResult = iterator.next();
-      } while (!iteratorResult.done);
-
-      let result = (this.result = iteratorResult.value);
-
-      // override .render function after initial render
-      this.render = () => result.rerender(options);
-    };
+    function handleError(fn: Function) {
```
pzuraq: Can you place this function into an `if (DEBUG)` statement? It looks like the debug stack is only included in development builds, and the `DEBUG` environment variable actually causes all of that code to be completely removed from production builds. This would prevent us from adding additional overhead in production, but allow us to keep the full benefits in dev.

pzuraq (changes requested):
Can you provide a description of what information this adds to the debug output? It seems to be serializing the full debug stack and adding it to any errors that occur, it'd be useful to see an example of what the output looks like to determine how helpful it would be

rwjblue:
@chancancode would mind taking a look at this one also? It seems related to recent work that you've done...

patricklx:
Hi, the description will look like this:
```
Uncaught Error: Assertion Failed: You must pass a function as the `fn` helpers first argument, you passed undefined
  in -> component: ui/routes/planets/details/posts-details
       ⤤ template: client/src/ui/routes/posts/details/template.hbs
       ⤤ template: packages/@ember/-internals/glimmer/lib/templates/outlet.hbs
       ⤤ template: client/src/ui/routes/application/template.hbs
       ⤤ template: packages/@ember/-internals/glimmer/lib/templates/outlet.hbs
    at assert (index.js:181)
    at fnHelper (index.js:7136)
    at InternalHelperReference.compute (index.js:855)
    at InternalHelperReference.value (index.js:400)
    at CapturedNamedArguments.value (runtime.js:5192)
    at processComponentArgs (index.js:4585)
    at CurlyComponentManager.create (index.js:4814)
    at Object.evaluate (runtime.js:1594)
    at AppendOpcodes.evaluate (runtime.js:94)
    at LowLevelVM.evaluateSyscall (runtime.js:3530)
```

unfortunately  it doesnt work for modifiers

packages/@ember/-internals/glimmer/lib/renderer.ts
```diff
@@ -95,7 +96,22 @@ class RootState {
       alwaysRevalidate: false,
     });
 
-    this.render = () => {
+    const handleError = (fn: () => void) => {
```
pzuraq: I think this can be made just a bit better:

```js
let handleError;

if (DEBUG) {
  handleError = fn => {
    return () => {
      try {
        fn();
      } catch (e) {
        e.message += '\n  in -> ' + env.debugStack.toString();
        env.debugStack._stack = [];
        throw e;
      }
    };
  }
}

let render = () => {
  let layout = template.asLayout();
  let handle = layout.compile();
  let iterator = renderMain(
  let result = (this.result = iteratorResult.value);

  // override .render function after initial render
  this.render = DEBUG ? 
    handleError(() => result.rerender(options)) : 
    () => result.rerender(options);
};

this.render = DEBUG ? handleError(render) : render;
```

This way we don't include the extra function call, and the `handleError` function will be fully removed from production builds.

rwjblue:
An update recently landed (in #18372) to add a new type of render debug support (this was needed for supporting custom component types in the ember-inspector), I'd be interested to see if you can take a look at that work and leverage it for some of the plumbing here.

packages/@ember/-internals/glimmer/lib/renderer.ts
```diff
@@ -95,7 +96,23 @@ class RootState {
...
 
-    this.render = () => {
+    let enhanceError;
+
+    if (DEBUG) {
+      enhanceError = (fn: () => void) => {
+        return () => {
+          try {
+            fn();
+          } catch (e) {
```
rwjblue: See my comments in https://github.com/emberjs/ember.js/pull/18388#pullrequestreview-290151477. I worry that adding a try/catch here will make local debugging _worse_ overall (when devtools is open).

packages/@ember/-internals/glimmer/lib/utils/debug-stack.ts
```diff
@@ -16,6 +16,13 @@ if (DEBUG) {
   DebugStack = class DebugStack {
     private _stack: TemplateElement[] = [];
 
+    toString() {
```
rwjblue: Lets add this to the new render stack, I suspect that the DebugStack will actually go away (in favor of using the same plumbing provided by the render stack).

rwjblue:
Looks great to me, would you mind squashing the commits down to one and prefixing it with `[BUGFIX beta]` (so we can make sure to get this into the 3.14 release)?

James1x0:
🙏Bitten by this often migrating to new syntaxes in octane. Any update @rwjblue ?

rwjblue (approved):

rwjblue:
Thanks again @patricklx!

deanpapastrat:
This is a godsend. Thank you for this 🙏

PR #18354, opened 2019-09-05
[DOC] Add example for typeOf(A())

amarkpark:
To distinguish typeOf(A) from typeOf(A()), since they return different values.

Calling `A` as constructor is deprecated.  `A()` is the recommended update.  It is a subtle but notable distinction.

### DIFF
--- a/packages/@ember/-internals/runtime/lib/type-of.js
+++ b/packages/@ember/-internals/runtime/lib/type-of.js
@@ -61,6 +61,7 @@ const { toString } = Object.prototype;
   typeOf(true);                   // 'boolean'
   typeOf(new Boolean(true));      // 'boolean'
   typeOf(A);                      // 'function'
+  typeOf(A());                    // 'array'
   typeOf([1, 2, 90]);             // 'array'
   typeOf(/abc/);                  // 'regexp'
   typeOf(new Date());             // 'date'


packages/@ember/-internals/runtime/lib/type-of.js
```diff
@@ -61,6 +61,7 @@ const { toString } = Object.prototype;
   typeOf(true);                   // 'boolean'
   typeOf(new Boolean(true));      // 'boolean'
   typeOf(A);                      // 'function'
```
rwjblue: tbh, I think this example is very confusing 😛 

Lets also change this one:

```suggestion
  typeOf(function() {});          // 'function'
```

amarkpark: All due respect, while that change may make perfect sense to you, to me (someone less experienced in the Ember universe, and thus FAR more likely to be consulting these docs for help or clarification in the first place) the original way I submitted it is far more helpful.

What you have here is a small sample of several data type primitives and several of the corresponding constructors.  In the case of testing these array constructors `typeOf(A)` as a contrast to `typeOf(A())` is the essence of my commit because THAT distinction was what threw me for a loop and was the crux of the problem with my code.

I consulted these docs after finding a couple of StackOverflow suggestions to use `if (typeOf(myArray) === function)` to distinguish between an Ember Array and a JS Array prototype when I was trying to determine if there would be Ember-specific actions available.  That confused me so I consulted the docs.  Long story short, I know that your update makes perfect sense to you, but the problem is that you already _know_ how all of these will act, and these docs are more for folks who need this specific clarification precisely because they _don't_ know how all of these will act.

jenweber: I think the addition of this doesn't hurt anything. Regular JavaScript typeofs are confusing overall, and so are Ember's by extension, unfortunately.

locks:
Thanks for the discussion everyone, merging!

PR #18358, opened 2019-09-05
[BUGFIX] Ensure QP definitions interop with tracked props

pzuraq:
This PR ensures QPs can be defined with native getters/setters or tracked properties. QP observers have been updated to always fire asynchronously, except in a few cases in the router's internals where we flush manually because we need to propagate changes immediately.

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/route.ts
+++ b/packages/@ember/-internals/routing/lib/system/route.ts
@@ -1,7 +1,10 @@
 import { OwnedTemplate, TemplateFactory } from '@ember/-internals/glimmer';
 import {
+  addObserver,
   computed,
   defineProperty,
+  descriptorForProperty,
+  flushAsyncObservers,
   get,
   getProperties,
   isEmpty,
@@ -17,9 +20,12 @@ import {
   setFrameworkClass,
   typeOf,
 } from '@ember/-internals/runtime';
+import { lookupDescriptor } from '@ember/-internals/utils';
+import { EMBER_METAL_TRACKED_PROPERTIES } from '@ember/canary-features';
 import Controller from '@ember/controller';
 import { assert, deprecate, info, isTesting } from '@ember/debug';
 import { ROUTER_EVENTS } from '@ember/deprecated-features';
+import { dependentKeyCompat } from '@ember/object/compat';
 import { assign } from '@ember/polyfills';
 import { once } from '@ember/runloop';
 import { classify } from '@ember/string';
@@ -940,6 +946,12 @@ class Route extends EmberObject implements IRoute {
 
     if (this._environment.options.shouldRender) {
       this.renderTemplate(controller, context);
+    }
+
+    // Setup can cause changes to QPs which need to be propogated immediately in
+    // some situations. Eventually, we should work on making these async somehow.
+    if (EMBER_METAL_TRACKED_PROPERTIES) {
+      flushAsyncObservers(false);
     }
   }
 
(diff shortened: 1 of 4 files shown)


packages/@ember/runloop/index.js
```diff
@@ -26,6 +27,7 @@ let flush;
 if (EMBER_METAL_TRACKED_PROPERTIES) {
   flush = function(queueName, next) {
     if (queueName === 'render' || queueName === _rsvpErrorQueue) {
+      flushSyncObservers();
```
rwjblue: Can you add a test specifically for this (e.g. unrelated to query params)?

rwjblue (dismissed):

packages/@ember/runloop/index.js
pzuraq: Discussed this with @wycats and @chancancode today actually, and it turns out we don't need to do this for this PR. I'm going to open up a followup with it and the tests.

rwjblue: Seems good. Though I _do_ think we need to do this specific thing (guarantee that the sync observers are flushed).

chancancode (approved):

packages/@ember/-internals/runtime/lib/mixins/observable.js
```diff
@@ -362,11 +362,12 @@ export default Mixin.create({
...
     @param {Object} target The target object to invoke
     @param {String|Function} method The method to invoke
+    @param {Boolean} async Whether the observer is async or not
     @return {Observable}
     @public
   */
-  addObserver(key, target, method) {
-    addObserver(this, key, target, method);
+  addObserver(key, target, method, async) {
+    addObserver(this, key, target, method, async);
```
rwjblue: We really need a test for this (unrelated to query params)

packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -2558,6 +2588,12 @@ Route.reopen(ActionHandler, Evented, {
         }
       }
 
+      // Some QPs have been updated, and those changes need to be propogated
+      // immediately. Eventually, we should work on making this async somehow.
+      if (EMBER_METAL_TRACKED_PROPERTIES && qpUpdated === true) {
+        flushAsyncObservers(false);
```
rwjblue: It is not clear to me at all why we need to sync flush here. Can you make an issue for us to track removing this (likely evaluating the specific failing test that caused it to be made sync)?

packages/@ember/-internals/runtime/lib/mixins/observable.js
```diff
@@ -379,11 +380,12 @@ export default Mixin.create({
...
     @param {Object} target The target object to invoke
     @param {String|Function} method The method to invoke
+    @param {Boolean} async Whether the observer is async or not
     @return {Observable}
     @public
   */
-  removeObserver(key, target, method) {
-    removeObserver(this, key, target, method);
+  removeObserver(key, target, method, async) {
+    removeObserver(this, key, target, method, async);
```
rwjblue: Needs a specific test, for users of observable (not related to query params).

packages/@ember/-internals/routing/lib/system/route.ts
pzuraq: For reference, the failing tests were:

```
<LinkTo /> component with query params (routing): the <LinkTo /> component applies activeClass when query params are not changed
{{link-to}} component with query params (routing): the {{link-to}} applies activeClass when query params are not changed
```

Issue #18275, opened 2019-08-16
Improvements to Octane component generators


chancancode:
### Add a Pro Tip™

When invoked with `--no-component-class` (which is the default in Octane), we should show the following Pro Tip™ at the end of the generator run:

```
❯❯❯ ember generate component foo
installing component
  create app/components/foo.hbs
installing component-test
  create tests/integration/components/foo-test.js

Tip: run `ember generate component-class foo` if you want to add a class
```

The exact message is TBD, but this is a perfectly good start.

When invoked with `--component-class=` set to any other values (whether using that flag or a 
related alias/shorthand), the message should not be shown.

### New alias, `--with-component-class`

Add a new alias, `--with-component-class`, as the "opposite" of `--no-component-class`.

This is roughly the same as running `ember generate component foo && ember generate component-class foo` back-to-back, other than that you should _not_ see the Pro Tip™ message. 

This flag serves as a way to say "I want a class" without having to think too much about "which one". Pre-Octane, it expands to `--component-class=@ember/component`. Post-Octane, it expands to `--component-class=@glimmer/component`.

This is also useful in guides and documentation, where you want to use the more verbose flag for readability, but not _as_ verbose as `--component-class=@glimmer/component`, which is explicitly drawing attention to the component class multiverse. It's not a bad thing to have choices, but most of the time you shouldn't have to care.

```
❯❯❯ ember generate component foo --with-component-class
installing component
  create app/components/foo.hbs
installing component-class
  create app/components/foo.js
installing component-test
  create tests/integration/components/foo-test.js
```

### New generator `ember generate component-class`

Split the class generation part into a separately invokable blueprint/generator, similar to how the `component-test` blueprint/generator can be invoked separately.

When the component generator is used with any of the `--component-class=...` other than `--no-component-class` (directly or through the aliases), the main component generator will delegate to this generator for the class generation, again similar to how `component-test` generator works.

Because `--with-component-class` just delegates to this under-the-hood, the have the same default class. Pre-Octane: `--component-class=@ember/component`. Post-Octane `--component-class=@glimmer/component`.

This generator should accept most of the same flags that is accepted by the main component generator. For example, and at minimum, it should be possible to call `ember generate component-class` with the `--component-class` and `--component-structure` flags, with the same aliases/shorthands supported by the main generator.

```
❯❯❯ ember generate component-class foo
installing component-class
  create app/components/foo.js
```

NullVoxPopuli:
dibs (will do after work / if I have a break)

PR #18359, opened 2019-09-06
WIP: create component-class blueprint

NullVoxPopuli:
Progress on https://github.com/emberjs/ember.js/issues/18275

### DIFF
--- a/blueprints/component/index.js
+++ b/blueprints/component/index.js
@@ -85,14 +85,14 @@ module.exports = {
     return this._super.install.apply(this, arguments);
   },
 
-  afterInstall(options) {
+  afterInstall({ componentClass, entity }) {
     this._super.afterInstall.apply(this, arguments);
 
-    // let ui = (this.ui = options.ui);
-    let { name } = options.entity;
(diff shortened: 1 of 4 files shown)


NullVoxPopuli:
As is, this could be feature-complete, but I was kinda hoping to be able to move component-class entirely into its own blueprint, so we don't need to keep the two blueprints in sync.

Not sure yet how to do that. 

I see that when you `ember g component`, you get two blueprints:
 - `component`
 - `component-test`

I couldn't figure out how `component-test` was getting thrown in there.

Last todos for this are tests

chancancode:
🙏

blueprints/component/index.js
```diff
@@ -37,6 +37,7 @@ module.exports = {
         { tc: '@ember/component/template-only' },
         { nc: '' },
         { 'no-component-class': '' },
+        { 'with-component-class': '@glimmer/component' },
```
chancancode: This should be `OCTANE ? '@glimmer/component' : '@ember/component'`

rwjblue:
Closed by https://github.com/emberjs/ember.js/pull/18396

PR #18363, opened 2019-09-06
[FEATURE ember-routing-model-arg] Enable by default.

rwjblue:
As discussed in todays team meeting.


pzuraq:
Do we need to update any API docs along with enabling this?

rwjblue:
@pzuraq - Yes, @chancancode is working up a docs specific PR for this now. Going to merge, so we can do a bit more testing but that PR should be up soon.

PR #18364, opened 2019-09-07
[DOC] Minor fixes for {{#each}} and {{#each-in}}

chancancode:
This is maybe part of #18363?

The `{{#each}}` docs is the only instance in the API docs where we reference a `model` property. It's not even clear if this is referring to the controller's property, but since it's just an arbitrary name for illustration purposes, I just changed it to match the rest of the examples given in the same block.

I also tried to make the `key` documentation a little more clear.

Since this is not directly coupled to #18363, this can be backported to beta.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/helpers/each-in.ts
+++ b/packages/@ember/-internals/glimmer/lib/helpers/each-in.ts
@@ -9,37 +9,44 @@ import { Opaque } from '@glimmer/util';
 /**
   The `{{#each}}` helper loops over elements in a collection. It is an extension
   of the base Handlebars `{{#each}}` helper.
+
   The default behavior of `{{#each}}` is to yield its inner block once for every
   item in an array passing the item as the first block parameter.
 
+  Assuming the `@developers` argument contains this array:
+
   ```javascript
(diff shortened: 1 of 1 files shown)


packages/@ember/-internals/glimmer/lib/helpers/each-in.ts
```diff
@@ -9,68 +9,92 @@ import { Opaque } from '@glimmer/util';
...
+  In this case, each time the `people` array is `map`-ed over, it will produce
+  an new array with completely different objects between renders. In these cases,
+  you can help Ember determine how these objects related to each other with the
+  `key` option:
+  
   ```handlebars
-  {{#each model key="id" as |item|}}
-  {{/each}}
+  <ul>
+    {{#each @developers key="name" as |person|}}
```
bradleypriest: I think you accidentally missed `index` here.
```suggestion
    {{#each @developers key="name" as |person index|}}
```

chancancode: Thank you!

chancancode: I ended up tweaking the examples a bit to be more consistent

rwjblue (approved):

PR #18382, opened 2019-09-13
Update computed sort sortDefinition type (in documentation)

ianknauer:
The rest of the documentation and assertions below indicate this should be an array (of strings) not a string.

### DIFF
--- a/packages/@ember/object/lib/computed/reduce_computed_macros.js
+++ b/packages/@ember/object/lib/computed/reduce_computed_macros.js
@@ -1428,7 +1428,7 @@ export function collect(...dependentKeys) {
   @param {String} itemsKey
   @param {Array} [additionalDependentKeys] optional array of additional
   dependent keys
-  @param {String or Function} sortDefinition a dependent key to an array of sort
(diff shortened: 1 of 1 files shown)


pzuraq:
So the actual type of `sort` is:

```ts
function sort(itemsKey: string, sortDefinition: string): ComputedDecorator;
function sort(itemsKey: string, sortDefinition: Function): ComputedDecorator;
function sort(itemsKey: string, additionalDependentKeys: string[], sortDefinition: Function): ComputedDecorator;

// full definition
function sort(itemsKey: string, sortDefinitionOrDependentKeys: string | string[] | Function, sortDefinition?: Function): ComputedDecorator;
```

Can you update the types to match this?

```
  @param {String} itemsKey
  @param {String|Function|Array} sortDefinitionOrDependentKeys The key of the sort definition (an array of sort properties), the sort function, or an array of additional dependent keys
  @param {Function?} sortDefinition The sort function (when used with additional dependent keys).
```

ianknauer:
@pzuraq i've updated it to match what you've got there.

pzuraq:
Awesome, thanks for doing this!

PR #18386, opened 2019-09-14
[DOC] Added example for EmberArray#firstObject

mukilane:
For #18228


mukilane:
Can we add [@readonly](https://jsdoc.app/tags-readonly.html) tag for `firstObject`?

locks:
@mukilane I'm not sure we have `@readonly` anywhere at the moment, so let's get this merged and see about it separately :)

locks (approved):
Thanks for the help!

PR #18372, opened 2019-09-10
Debug Render Tree (for Ember Inspector)

chancancode:
Introduce debug render tree

1. Before loading Ember, set `ENV._DEBUG_RENDER_TREE = true;`

   This controls whether to perform extra bookkeeping needed to make the `captureRenderTree` API work.

   This has to be set before the ember JavaScript code is evaluated.

   This is usually done by setting one of these...
   
   ```
   window.EmberENV = { _DEBUG_RENDER_TREE: true };
   ```

   ```
   window.ENV = { _DEBUG_RENDER_TREE: true };
   ```
   
   ...before the "vendor" `<script>` tag in `index.html`.

   Setting the flag after Ember is already loaded will not work correctly. It may appear to work somewhat, but fundamentally broken.
   
   This is not intended to be set directly. Ember Inspector will enable the flag on behalf of the user as needed.
   
   This flag is always on in development mode.
   
   The flag is off by default in production mode, due to the cost associated with the the bookkeeping work.

   The expected flow is that Ember Inspector will ask the user to refresh the page after enabling the feature. It could also offer a feature where the user add some domains to the "always on" list. In either case, Ember Inspector will inject the code on the page to set the flag if needed.

2. With the flag on, `Ember._captureRenderTree()` is available.

   It takes the *application instance* as an argument and returns an array of `CapturedRenderNode`:

   ```typescript
   interface CapturedRenderNode {
     type: 'outlet' | 'engine' | 'route-template' | 'component';
     name: string;
     args: {
       positional: unknown[];
       named: Dict<unknown>;
     };
     instance: unknown;
     bounds: Option<{
       parentElement: Simple.Element;
       firstNode: Simple.Node;
       lastNode: Simple.Node;
     }>;
     children: CapturedRenderNode[];
   }
   ```

PR #18387, opened 2019-09-14
[BUGFIX] Ensure `updateComponent` is fired consistently

chancancode:
Previously, if you have no args, or all static args, the custom component's (technically, the manager's) `updateComponent` hook will never fire. This is a VM bug that we need to fix, but in the mean time, this is a fine workaround that doesn't has much cost associated with it.


chancancode:
This is split from #18372

PR #18388, opened 2019-09-14
[BUGFIX beta] Don't mask errors in commit

chancancode:


packages/@ember/-internals/glimmer/lib/environment.ts
```diff
@@ -99,6 +99,12 @@ export default class Environment extends GlimmerEnvironment {
 
     try {
       super.commit();
+    } catch(error) {
+      if (completedWithoutError) {
+        throw error;
+      } else {
+        console.error(error);
```
pzuraq: So, if I'm reading this correctly, if `completedWithoutError` is true then `error` comes from the commit itself. What is the `error` if `completedWithoutError` is false? Is it the error of the outer `try/catch`?

chancancode: Correct. Basically we were already planning to throw a different error from where we are called.

pzuraq: Ahhh I see now, ok 👍

pzuraq (approved):

rwjblue (changes requested):
I discussed this with @chancancode a bit at the F2F. This change concerns me a bit, specifically that the debugging experience would suffer (since you will now not get an error inside your component's `didInsertElement` or other userland hook location because we try/catch here I think that chrome would stop here **deep** in the renderer internals).

This can be somewhat mitigated by only adding the `try`/`catch` when `completedWithoutError` is `false` in the Environment.prototype.commit.

locks:
@rwjblue @chancancode is this path being abandoned?

chancancode:
Pretty sure we did this already

PR #18399, opened 2019-09-17
[BUGFIX beta] Ensures that `@each` works with non-arrays and array-likes

chancancode:
`@each` should no-op with non arrays, since there is nothing to track,
and should work with array-likes (anything that implements the
`objectAt` and `length` properties.)

### DIFF
--- a/packages/@ember/-internals/metal/lib/chain-tags.ts
+++ b/packages/@ember/-internals/metal/lib/chain-tags.ts
@@ -1,7 +1,7 @@
 import { meta as metaFor, peekMeta } from '@ember/-internals/meta';
-import { isEmberArray } from '@ember/-internals/utils';
 import { assert, deprecate } from '@ember/debug';
 import { combine, createUpdatableTag, Tag, update, validate } from '@glimmer/reference';
+import { objectAt } from './array';
 import { getLastRevisionFor, peekCacheFor } from './computed_cache';
 import { descriptorForProperty } from './descriptor_map';
 import { tagForProperty } from './tags';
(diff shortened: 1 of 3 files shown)


chancancode:
@pzuraq the assertion further down about `typeof === 'object'` _may_ fall into the same category (but only if it works correctly on release)

pzuraq:
That assertion was a direct requirement of the `EachProxy`: https://github.com/emberjs/ember.js/blob/master/packages/@ember/-internals/metal/lib/each_proxy.ts#L128

pzuraq (approved):

chancancode:
@pzuraq Pushed a second test re: the other assertion. It seems to work on release but double check me. I think there are extra guards somewhere else that prevents it from going into the `EachProxy` code path?

chancancode:
Opened https://github.com/emberjs/ember.js/pull/18400 to confirm the tests work.

chancancode:
Nevermind! My test was invalid!

wycats:
lgtm

packages/@ember/-internals/metal/lib/chain-tags.ts
```diff
@@ -70,48 +70,71 @@ export function getChainTagsForKey(obj: any, path: string) {
...
+        // If the current object isn't an array, there's nothing else to do,
+        // we don't watch individual properties. Break out of the loop.
+        break;
+      } else if (arrLength === 0) {
+        // Fast path for empty arrays
+        chainTags.push(tagForProperty(current, '[]'));
+        break;
+      }
 
-      if (segment.indexOf('.') !== -1) {
```
chancancode: I believe this was dead code.

There are two cases on L93: `segmentEnd === -1` (`foo.@each.bar`) and `segmentEnd !== -1` (`foo.@each.bar.baz`, deprecated).

For the first case, `segment` is `path.slice(lastSegmentEnd, pathLength)` which would be `bar`. It is impossible for this to contain a `.` because we already checked previously.

For the second case, `segment` is `path.slice(lastSegmentEnd, segmentEnd)` which would also be `bar`. It is impossible for this to contain a `.`, because `segmentEnd` is the next occurrence of a `.` after the `@each.`, returned from `indexOf('.', lastSegmentEnd)`.

So in both cases, this conditional cannot ever by true so the branch is effectively dead but we are still doing part of the work (`segment.indexOf('.')`) for no reason.

PR #18410, opened 2019-09-18
Use ember-cli-htmlbars for inline precompilation if possible.

rwjblue:
ember-cli-htmlbars@4.0.0+ provides the ability to do inline precompilation out of the box (no need for a separate addon), this updates our test generator to import from the right place and avoid installing the second addon if we don't need it.


chancancode (approved):

blueprints/component-test/index.js
```diff
@@ -70,20 +75,32 @@ module.exports = useTestFrameworkDetector({
     return {
       path: getPathOption(options),
       testType: testType,
```
devotox: This can also be shortened

rwjblue: Great point, mind sending in a PR?

PR #18420, opened 2019-09-21
WIP: Provide Tracked  Debugging Tools

NullVoxPopuli:
This is super WIP.
atm, I'm trying to figure out how tracked works, and how I can get access to objects when things are tracked.

Ideally, what I'm trying to provide
 - history of "the tracked state"
 - provide reasons why a tracked property changed
   - what property changed on what object
 - provide references to objects that tracked properties are defined on so that ember inspector can tie in to all this
 
Eventually:
 - build a snapshot of the tree of dependencies and their values, and what objects the properties live on 

Latest Screenshot:
image

Tracking History at boot on emberclear:
image

Looks like Tracking History might not be feasible for real apps.
This took a while to print to the console:
image

It looks like the Memory in use still isn't that bad (I have no idea where this data would be on here, but it's all in low #s of MB)
image

After clicking around a bit, and sending myself some messages, it looks like I either have a memory leak, or tracking all this history _is_ actually problematic:
image

Yeah here is a memory snapshot after doing similar actions on the deployed site: 
image
I'm much more inclined to think I have less of a memory leak and more that I'm just tracking too much stuff. (I do have a memory leak -- it's just _much_ slower of a leak than what the initial memory snapshots show)

To mitigate this memory bloat, there is a start/stop method for optionally recording the tracked stacks.

### DIFF
--- a/packages/@ember/-internals/metal/index.ts
+++ b/packages/@ember/-internals/metal/index.ts
@@ -61,7 +61,7 @@ export { Mixin, aliasMethod, mixin, observer, applyMixin } from './lib/mixin';
 export { default as inject, DEBUG_INJECTION_FUNCTIONS } from './lib/injected_property';
 export { tagForProperty, tagFor, markObjectAsDirty, UNKNOWN_PROPERTY_TAG } from './lib/tags';
 export { default as runInTransaction, didRender, assertNotRendered } from './lib/transaction';
-export { consume, Tracker, tracked, track, untrack, isTracking } from './lib/tracked';
+export { consume, Tracker, tracked, track, untrack, isTracking } from './lib/tracked/index';
 
 export {
   NAMESPACES,
--- a/packages/@ember/-internals/metal/lib/tracked/debug-printing.ts
+++ b/packages/@ember/-internals/metal/lib/tracked/debug-printing.ts
@@ -54,12 +54,11 @@ function printDependents(
   let verboseOverride = getTrackingInfo().verbose;
 
   dependencies.forEach(dependency => {
-    let isChangedProperty = rootTag.tag.propertyName === dependency.propertyName;
-
-    let wasChanged = hasChanged(rootTag, dependency.dependencies);
+    let isChanged = rootTag.tag.lastChecked === dependency.revision;
+    let isChangedProperty = isChanged;
 
     if (!verboseOverride) {
-      if (!verbose && !wasChanged) {
+      if (!verbose && !isChangedProperty) {
         return;
       }
     }
@@ -81,4 +80,3 @@ function printDependents(
     printDependents(rootTag, dependency.dependencies, verbose, indent + 2);
   });
 }
-
--- a/packages/@ember/-internals/metal/lib/tracked/debug-printing.ts
+++ b/packages/@ember/-internals/metal/lib/tracked/debug-printing.ts
@@ -1,0 +1,84 @@
+import { TrackerSnapshot, TagSnapshot } from './debug-types';
+import { hasBeenSet, hasChanged, getTrackingInfo } from './debug-utils';
+
+export function prettyPrintTrackingInfo({ verbose = false, showInitial = true } = {}) {
+  let history = getTrackingInfo().history;
+  let revisions = Object.keys(history)
+    .map(revision => parseInt(revision, 10))
+    .sort((a, b) => a - b);
+
+  let i;
+  let currentRevision: number;
+  let currentBatch: TrackerSnapshot[];
+  let changedTag: TrackerSnapshot;
+
+  for (i = 0; i < revisions.length; i++) {
+    currentRevision = revisions[i];
+    currentBatch = history[currentRevision] || [];
+
+    if (!showInitial && `${currentRevision}` === '1') {
+      continue;
+    }
+
+    // eslint-disable-next-line no-console
+    console.log(`[Revision: ${currentRevision}]`, currentBatch);
+    changedTag = currentBatch[0];
+
+    currentBatch.forEach((tracker, idx: number) => {
+      let { objectName, propertyName, objectId } = tracker.tag;
+
+      let wasSet = hasBeenSet(tracker);
+
+      if (wasSet) {
+        // eslint-disable-next-line no-console
+        console.log(`  #${idx}: ${propertyName} on ${objectName} (#${objectId}) has been set!`);
+      } else {
+        // eslint-disable-next-line no-console
+        console.log(`  #${idx}: ${propertyName} on ${objectName} (#${objectId}) has changed!`);
+      }
+
+      printDependents(changedTag, tracker.dependencies, verbose);
+    });
+  }
+}
+
+function printDependents(
+  rootTag: TrackerSnapshot,
+  dependencies: TagSnapshot[],
+  verbose: boolean,
+  indent = 6
+) {
+  if (!dependencies || dependencies.length === 0) return;
+
+  let indentation = ' '.repeat(indent);
+  let verboseOverride = getTrackingInfo().verbose;
+
+  dependencies.forEach(dependency => {
+    let isChangedProperty = rootTag.tag.propertyName === dependency.propertyName;
+
+    let wasChanged = hasChanged(rootTag, dependency.dependencies);
+
+    if (!verboseOverride) {
+      if (!verbose && !wasChanged) {
+        return;
+      }
+    }
+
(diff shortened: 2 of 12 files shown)


packages/@ember/-internals/metal/index.ts
```diff
@@ -61,7 +61,7 @@ export { Mixin, aliasMethod, mixin, observer, applyMixin } from './lib/mixin';
 export { default as inject, DEBUG_INJECTION_FUNCTIONS } from './lib/injected_property';
 export { tagForProperty, tagFor, markObjectAsDirty, UNKNOWN_PROPERTY_TAG } from './lib/tags';
 export { default as runInTransaction, didRender, assertNotRendered } from './lib/transaction';
-export { consume, Tracker, tracked, track, untrack, isTracking } from './lib/tracked';
+export { consume, Tracker, tracked, track, untrack, isTracking } from './lib/tracked/index';
```
NullVoxPopuli: I did some re-organizing a bit, because I wanted all my debug stuff in a separate file. 
Putting my debug stuff in a separate file caused me to move `Tracker` to a new file.
Doing that caused me to move `tracked.ts` to `tracked/index.ts` (even though it wasn't really needed, I thought it'd be a nice organization win). Unfortunately, whatever bundling system were using wasn't making the assumption that `import {..} from './tracked'` should be the same as `import {..} from './tracked/index'` :-\

NullVoxPopuli:
What I have currently (not much):
image

packages/@ember/-internals/metal/lib/tracked/debugging.ts
```diff
@@ -0,0 +1,110 @@
+import { Tag, CONSTANT_TAG, UpdatableTag } from '@glimmer/reference';
+
+import { Tracker } from './tracker';
+
+type Option<T> = T | null;
+
+Ember.EMBER_DEBUG = {};
+
+let TAG_MAP = new Map<Tag | UpdatableTag, any>();
```
NullVoxPopuli: I was thinking about storing a map of tags to their host objects in here.
I later found out that none of the invocations to consume (or other places where `CURRENT_TRACKER.add` is called) contain a references to the object/instance. :(

packages/@ember/-internals/metal/lib/tracked/debugging.ts
```diff
@@ -0,0 +1,110 @@
...
+  tag: Tag | UpdatableTag;
+}
+
+Ember.EMBER_DEBUG.TRACKING = {
+  history: [],
+  stacks: new Set<Tracker>(),
+} as DebugTracking;
+
+function normalizeTrackerEntry(tag: any) {
+  // it looks like there are two types of data on a tracker's tag list.
```
NullVoxPopuli: this was confusing -- couldn't find any TS Types that described both of these shapes.
Is there a tool to search all types in the project / node_modules matching a certain shape?

packages/@ember/-internals/metal/lib/tracked/debugging.ts
```diff
@@ -0,0 +1,110 @@
...
+  // one type has the shape (which maybe isn't a tag?):
+  //   key: string
+  //   ref: object
+  //   tag: object
+  // the other is an actual Tag,
+  //   but with a private secret property _propertyName
+
+  let hostObject = TAG_MAP.get(tag);
```
NullVoxPopuli: this is always undefined -- I _can_ access the object on tags who have a `ref` -- which, I think are tags that are from component invocations.

packages/@ember/-internals/metal/lib/tracked/debugging.ts
```diff
@@ -0,0 +1,110 @@
...
+  });
+}
+
+Ember.EMBER_DEBUG.TRACKING.print = prettyPrintTrackingInfo;
+
+export function debugTracker(current: Tracker, parent: Option<Tracker>) {
+  // Convert the Tracker to an isolated moment in time
+  Ember.EMBER_DEBUG.TRACKING.history.push({
+    // hack around tags being a private field
+    tags: Array.from((current as any).tags.values()),
```
NullVoxPopuli: idk if this is fine, or if we want to make tags non-private 🤷‍♂

packages/@ember/-internals/metal/lib/tracked/tracker.ts
```diff
@@ -0,0 +1,33 @@
+import { combine, CONSTANT_TAG, Tag } from '@glimmer/reference';
```
NullVoxPopuli: this was just extracted from the main tracked.ts file

packages/@ember/-internals/metal/lib/tracked/debugging.ts
```diff
@@ -0,0 +1,110 @@
...
+  });
+
+  console.log('[ track event ] current:', current, 'parent:', parent);
+  // Ember.EMBER_DEBUG.TRACKING.print();
+}
+
+
+export function debugConsume(tracker: Tracker, tag: Tag | UpdatableTag, target: any) {
```
NullVoxPopuli: the tracker itself doesn't have much information on it.
It has a tags Set, and that's about it
image

packages/@ember/-internals/metal/lib/tracked/debugging.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  return result;
+}
+
+function prettyPrintTrackingInfo() {
+  let history = Ember.EMBER_DEBUG.TRACKING.history;
+
+  console.log(history, TAG_MAP);
+  history.forEach((tracker: TrackerSnapshot, i: number) => {
+    // tags seems to always be a touple
```
NullVoxPopuli: I don't know if this is always true.

In the app I'm debugging this with, this is my property dep tree:

```
count
|- doubled
|- <ApplyMath @num={{this.count}} />
   |- dividedByThree
      |- again
```

So, indirectly, the property `again` should have a stack of 3 high (but it doesn't?)?

count -> dividedByThree -> again

NullVoxPopuli:
Ooooooo

image

NullVoxPopuli:
Current output:
image

NullVoxPopuli:
Note to self: Computed Properties have a _less_ flat tree of dependencies: 
image

NullVoxPopuli:
And the history of computed properties that depend on tracked properties:
image

NullVoxPopuli:
Been working on cleaning up the output a bit (in emberclear): 
image
And adding a verbose option:
image
Not sure how useful that is though. 
non-changed tags aren't really useful. It'd be more useful if I could get the reference to the function or event that caused the change

NullVoxPopuli:
Even though this tool is pretty primitive so far, it's helped me improve my app, and pointed out some potential places of improvement in addons!

NullVoxPopuli:
There is now live watching of changes. Will be good for capturing long running loops, and side-effects / observers and such. 

I think this is in a good place to start integrating with ember inspector?

NullVoxPopuli:
Made a little app for easier-to-digest property tracking:
image
https://github.com/NullVoxPopuli/property-scenarios

packages/@ember/-internals/metal/lib/tracked/debug-printing.ts
```diff
@@ -0,0 +1,82 @@
...
+  for (i = 0; i < revisions.length; i++) {
+    currentRevision = revisions[i];
+    currentBatch = history[currentRevision] || [];
+
+    if (!showInitial && `${currentRevision}` === '1') {
+      continue;
+    }
+
+    // eslint-disable-next-line no-console
+    console.log(`[Revision: ${currentRevision}]`, currentBatch);
```
lifeart: https://developer.mozilla.org/en-US/docs/Web/API/Console/group

packages/@ember/-internals/metal/lib/tracked/debug-printing.ts
```diff
@@ -0,0 +1,82 @@
...
+        return;
+      }
+    }
+
+    if (!dependency.objectRef && !dependency.propertyName) {
+      // eslint-disable-next-line no-console
+      console.log(`${indentation} Intermediate Tracking Tag @ rev: ${dependency.revision}`);
+    } else {
+      // eslint-disable-next-line no-console
+      console.log(
```
lifeart: in theory, https://developer.mozilla.org/en-US/docs/Web/API/Console/table may help

NullVoxPopuli:
@pzuraq would any of this still be useful? 

I think when we've talked about this on discord, another approach is being taken.

gonna close this for cleanup purposes

PR #18429, opened 2019-09-24
[BUGFIX release] Fix incorrect error message for octane features.

### LINKED ISSUE #18430: feature:enable vs feature:enabled (message is misleading as the ...feature:enabled.... does not work
"enabled" should be "enable" in the following...(ember feature:enabled template-only-glimmer-components)
           
    * The template-only-glimmer-components optional feature should be enabled under Octane, run `ember feature:enabled template-only-glimmer-components` to enable it

This is the message that was shown while trying to bring up ember server.

rwjblue:
The correct command is `ember feature:enable` :sob: :weary:

Fixes: #18430

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -99,7 +99,7 @@ module.exports = {
         !optionalFeatures.isFeatureEnabled('template-only-glimmer-components')
       ) {
         message.push(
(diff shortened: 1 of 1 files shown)


chancancode (approved):
😆

chancancode:
back-ported to beta and release

PR #17948, opened 2019-04-18
[DEPRECATION] deprecate `Component#isVisible`

GavinJoyce:
part of https://github.com/emberjs/ember.js/issues/17918

TODO:

- [x] Update the deprecation message to include the component's `toString`
- [x] Add svelting for the new deprecation
- [x] Update documentation to mark `isVisible` as `@deprecated`
- [x] Deprecation guide PR: https://github.com/ember-learn/deprecation-app/pull/451

PR #18436, opened 2019-09-27
[DEPRECATION] Deprecate globals resolver

Gaurav0:
For https://github.com/emberjs/rfc-tracking/issues/20

- [x] Deprecate instantiating @ember/application/globals-resolver (in init of packages/@ember/application/globals-resolver.ts)
- [x] Deprecate using Ember.Resolver / Ember.DefaultResolver (in packages/ember/index.js)

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/application/engine-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/application/engine-test.js
@@ -1,4 +1,10 @@
-import { moduleFor, ApplicationTestCase, strip, runTaskNext } from 'internal-test-helpers';
+import {
+  moduleFor,
+  ApplicationTestCase,
+  ModuleBasedTestResolver,
+  strip,
+  runTaskNext,
+} from 'internal-test-helpers';
 
 import { Component } from '@ember/-internals/glimmer';
 import { Route } from '@ember/-internals/routing';
@@ -80,6 +86,8 @@ moduleFor(
       this.add(
         'engine:blog',
         Engine.extend({
+          Resolver: ModuleBasedTestResolver,
+
           init() {
             this._super(...arguments);
             this.register(
@@ -131,6 +139,8 @@ moduleFor(
       this.add(
         'engine:chat-engine',
         Engine.extend({
+          Resolver: ModuleBasedTestResolver,
+
           init() {
             this._super(...arguments);
             this.register('template:application', compile('Engine'));
@@ -167,6 +177,8 @@ moduleFor(
       this.add(
         'engine:blog',
         Engine.extend({
+          Resolver: ModuleBasedTestResolver,
+
           init() {
             this._super(...arguments);
             this.register('template:foo', compile('foo partial'));
@@ -202,6 +214,8 @@ moduleFor(
       this.add(
         'engine:chat-engine',
         Engine.extend({
+          Resolver: ModuleBasedTestResolver,
+
           init() {
             this._super(...arguments);
             this.register('template:foo', compile('foo partial'));
@@ -230,6 +244,8 @@ moduleFor(
       this.add(
         'engine:chat-engine',
         Engine.extend({
+          Resolver: ModuleBasedTestResolver,
+
           init() {
             this._super(...arguments);
             this.register('template:components/foo-bar', compile(`{{partial "troll"}}`));
@@ -289,6 +305,8 @@ moduleFor(
       this.add(
         'engine:blog',
         Engine.extend({
+          Resolver: ModuleBasedTestResolver,
+
           init() {
             this._super(...arguments);
 
@@ -344,6 +362,8 @@ moduleFor(
       this.add(
         'engine:blog',
         Engine.extend({
+          Resolver: ModuleBasedTestResolver,
+
           init() {
             this._super(...arguments);
             this.register(
@@ -486,6 +506,8 @@ moduleFor(
       this.add(
         'engine:blog',
         Engine.extend({
+          Resolver: ModuleBasedTestResolver,
+
           init() {
             this._super(...arguments);
             this.register('template:application', compile('Engine{{outlet}}'));
--- a/packages/@ember/application/globals-resolver.js
+++ b/packages/@ember/application/globals-resolver.js
@@ -4,7 +4,7 @@
 
 import { dictionary } from '@ember/-internals/utils';
 import { get, findNamespace } from '@ember/-internals/metal';
-import { assert, info } from '@ember/debug';
+import { assert, info, deprecate } from '@ember/debug';
 import { capitalize, classify, dasherize, decamelize } from '@ember/string';
 import { Object as EmberObject } from '@ember/-internals/runtime';
 import { getTemplate } from '@ember/-internals/glimmer';
@@ -77,6 +77,7 @@ import { DEBUG } from '@glimmer/env';
   @class GlobalsResolver
   @extends EmberObject
   @public
+  @deprecated
 */
 
 class DefaultResolver extends EmberObject {
@@ -93,9 +94,20 @@ class DefaultResolver extends EmberObject {
 
     @property namespace
     @public
+    @deprecated
   */
 
   init() {
+    deprecate(
+      'Using the globals resolver is deprecated. Use the ember-resolver package instead. See https://deprecations.emberjs.com/v3.x#toc_ember-deprecate-globals-resolver',
+      false,
+      {
+        until: '4.0.0',
+        id: 'globals-resolver',
+        url: 'https://deprecations.emberjs.com/v3.x#toc_ember-deprecate-globals-resolver',
+      }
+    );
+
     this._parseNameCache = dictionary(null);
   }
 
--- a/packages/@ember/application/globals-resolver.js
+++ b/packages/@ember/application/globals-resolver.js
@@ -9,6 +9,7 @@ import { capitalize, classify, dasherize, decamelize } from '@ember/string';
 import { Object as EmberObject } from '@ember/-internals/runtime';
 import { getTemplate } from '@ember/-internals/glimmer';
 import { DEBUG } from '@glimmer/env';
+import { GLOBALS_RESOLVER } from '@ember/deprecated-features';
 
 /**
   The DefaultResolver defines the default lookup rules to resolve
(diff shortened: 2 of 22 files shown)


runspired:
YEAAAAAASSSSSSSSSSSS cc @scalvert I was just complaining to you that this still existed the other day :D

packages/@ember/-internals/glimmer/tests/integration/application/engine-test.js
```diff
@@ -81,7 +81,9 @@ moduleFor(
         'engine:blog',
         Engine.extend({
           init() {
-            this._super(...arguments);
+            ignoreDeprecation(() => {
```
rwjblue: Hmm, its not clear why we hit the assertion here. AFAICT, this test isn't about using the deprecated resolver, so it should probably be updated to avoid the deprecation (vs ignoring it).

packages/@ember/-internals/glimmer/tests/integration/application/engine-test.js
```diff
@@ -168,7 +172,9 @@ moduleFor(
...
```
rwjblue: Same as above. Will only comment on these two instances, but I think most (possibly all?) of the `ignoreDeprecation` calls (in this test file at least) should be removed and we should figure out how to avoid the deprecation in the tests.

packages/@ember/application/globals-resolver.js
```diff
@@ -93,9 +94,19 @@ class DefaultResolver extends EmberObject {
...
 
   init() {
+    deprecate(
+      'Using the globals resolver is deprecated. Use the ember-resolver package instead. See https://deprecations.emberjs.com/v3.x#toc_ember-deprecate-globals-resolver',
+      false,
+      {
+        until: '4.0.0',
+        id: 'globals-resolver',
+      }
+    );
```
rwjblue: We need to add a `url` property here pointing to the deprecation guide URL for this.

packages/@ember/application/lib/application.js
```diff
@@ -844,7 +844,7 @@ const Application = Engine.extend({
       _loaded.application = undefined;
     }
 
-    if (this._applicationInstances.size) {
+    if (this._applicationInstances && this._applicationInstances.size) {
```
rwjblue: Hmm, why isn't `this._applicationInstances` defined (it is setup in `init`, which should be before anything else here has accessed it)?

packages/@ember/application/tests/application_instance_test.js
```diff
@@ -21,9 +20,16 @@ moduleFor(
...
       document.getElementById('qunit-fixture').innerHTML = `
-      <div id='one'><div id='one-child'>HI</div></div><div id='two'>HI</div>
-    `;
-      application = run(() => Application.create({ rootElement: '#one', router: null }));
+        <div id='one'><div id='one-child'>HI</div></div><div id='two'>HI</div>
+      `;
+
+      application = run(() => {
+        let app;
+        ignoreDeprecation(() => {
```
rwjblue: Is there anything about this test that is actually hitting the deprecation? Do we need to refactor the test harness to avoid using the globals resolver?

packages/@ember/application/tests/application_instance_test.js
```diff
@@ -165,7 +171,7 @@ moduleFor(
     }
 
     ['@test can build and boot a registered engine'](assert) {
-      assert.expect(11);
+      assert.expect(12);
```
rwjblue: Where does the extra assertion come from?

packages/@ember/application/tests/dependency_injection_test.js
```diff
@@ -13,7 +13,9 @@ moduleFor(
     constructor() {
       super();
 
-      application = run(EmberApplication, 'create');
+      expectDeprecation(() => {
```
rwjblue: This test is definitely using globals resolution mode, but it doesn't seem like it fundamentally needs to. We should probably make base class with a `setupRegistry` method then extend from that in two subclasses where one builds a global setup like the tests do currently and another uses the registry directly.

packages/ember/index.js
```diff
@@ -175,9 +176,29 @@ if (EMBER_EXTEND_PROTOTYPES) {
...
+Object.defineProperty(Ember, 'Resolver', {
+  get() {
+    deprecate(
+      'Using the globals resolver is deprecated. Use the ember-resolver package instead. See https://deprecations.emberjs.com/v3.x#toc_ember-deprecate-globals-resolver',
+      false,
+      {
+        id: 'ember.globals-resolver',
+        until: '4.0.0',
+      }
+    );
```
rwjblue: We'll need a `url` property here

packages/internal-test-helpers/lib/engine.js
```diff
@@ -0,0 +1,9 @@
+import Engine from '@ember/engine';
```
rwjblue: Ultimately, I think we shouldn't need this file. Just extending `Engine` shouldn't necessarily trigger a deprecation...

Gaurav0:
Thank you for reviewing and providing feedback, @rwjblue . I'm working on this.

packages/@ember/-internals/glimmer/tests/integration/application/engine-test.js
Gaurav0: So basically, if we extend Engine without supplying a Resolver other than DefaultResolver, we hit the deprecation.

packages/@ember/-internals/glimmer/tests/integration/application/engine-test.js
Gaurav0: I think I need to refactor all of the engines tests to use the ModuleBasedTestResolver instead of the DefaultResolver.

packages/@ember/application/globals-resolver.js
Gaurav0: Fixed.

packages/@ember/application/tests/application_instance_test.js
Gaurav0: We are not supplying a resolver, so it uses DefaultResolver by default.

packages/@ember/application/tests/application_instance_test.js
Gaurav0: I'm going to see if I can make it clear, but it requires some messing with QUnit.

packages/ember/index.js
Gaurav0: Done.

packages/internal-test-helpers/lib/engine.js
Gaurav0: Yes, it does, if you don't override the Resolver property to something other than the DefaultResolver.

packages/@ember/application/lib/application.js
Gaurav0: I am not sure why I needed this, it's been like 7 weeks. Anyway, reverted this change.

Gaurav0:
@rwjblue Updated. I hope this alleviates most of your concerns.

packages/@ember/-internals/glimmer/tests/integration/application/engine-test.js
rwjblue: Ya, that sounds right

packages/@ember/application/tests/dependency_injection/default_resolver_test.js
```diff
@@ -198,6 +198,7 @@ moduleFor(
     }
 
     [`@test no assertion for routes that extend from Route`](assert) {
+      assert.test.assertions = []; // clear assertions that occurred in beforeEach
```
rwjblue: Hmm, why do we need to clear the assertions?

packages/@ember/application/tests/dependency_injection_test.js
```diff
@@ -13,7 +13,12 @@ moduleFor(
     constructor() {
       super();
 
-      application = run(EmberApplication, 'create');
+      // Must use default resolver because test resolver does not normalize
```
rwjblue: Gotcha, so lets split the tests that rely on the globals resolver into their own moduleFor + class (e.g. `moduleFor('Application Dependency Injection - Globals Resolver [DEPRECATED]', class extends TestCase {})`).

packages/@ember/application/tests/dependency_injection/normalization_test.js
```diff
@@ -10,7 +10,12 @@ moduleFor(
     constructor() {
       super();
 
-      application = run(Application, 'create');
+      // Must use default resolver because test resolver does not normalize
```
rwjblue: Can you update the `moduleFor` name of this to indicate it is specifically for the globals resolver and that it is testing deprecated functionality?

packages/ember/index.js
```diff
@@ -175,9 +176,30 @@ if (EMBER_EXTEND_PROTOTYPES) {
...
+      {
+        id: 'ember.globals-resolver',
+        until: '4.0.0',
+        url: 'https://deprecations.emberjs.com/v3.x#toc_ember-deprecate-globals-resolver',
+      }
+    );
+
+    return Resolver;
+  },
+});
```
rwjblue: Since we issue a deprecation in `init`, wouldn't we end up with two deprecations for _roughly_ the same thing?

rwjblue (commented):
Thanks for working on this!

rwjblue:
I think we also need to add some svelting conditionals (so we can strip the deprecated code paths). The basic steps for this are:

* add another entry to `@ember/deprecated-features`
* import that flag in any file that implements the deprecated behavior (I think this is mostly the location that the globals resolver itself is defined)
* wrap the implementation code in an `if (GLOBALS_RESOLVER) {`

Some example PRs that have landed that might be helpful:

* https://github.com/emberjs/ember.js/pull/18491
* https://github.com/emberjs/ember.js/pull/17948
* https://github.com/emberjs/ember.js/pull/17910
* https://github.com/emberjs/ember.js/pull/17525

packages/@ember/application/tests/dependency_injection/default_resolver_test.js
Gaurav0: In beforeEach we are creaing an application using the default resolver.

packages/ember/index.js
Gaurav0: One would be for creating an application without specifying a resolver, the other for accessing through Ember.Resolver or Ember.DefaultResolver. I generally think most users will not likely run into both unless they specifically request Ember.DefaultResolver as their resolver during application.create.

rwjblue: Ya, agreed. Thank you!

packages/@ember/application/tests/dependency_injection_test.js
Gaurav0: There's only one test that relies on it, so I'm just going to move it into `default_resolver_test.js`.

packages/@ember/application/tests/dependency_injection/normalization_test.js
Gaurav0: Done!

packages/@ember/deprecated-features/index.ts
```diff
@@ -16,3 +16,4 @@ export const FUNCTION_PROTOTYPE_EXTENSIONS = !!'3.11.0-beta.1';
 export const MOUSE_ENTER_LEAVE_MOVE_EVENTS = !!'3.13.0-beta.1';
 export const EMBER_COMPONENT_IS_VISIBLE = !!'3.15.0-beta.1';
 export const PARTIALS = !!'3.15.0-beta.1';
+export const GLOBALS_RESOLVER = !!'3.16.0';
```
rwjblue:
```suggestion
export const GLOBALS_RESOLVER = !!'3.16.0-beta.1';
```

rwjblue (commented):
Looks great, only one minor nit-pick left.

rwjblue:
Thank you @Gaurav0!

Gaurav0:
Thank you @rwjblue !

PR #18450, opened 2019-10-02
[DOC beta] document `Transition.to` as nullable

jelhan:
Property `to` of `Transition` seems to be nullable. At least I got an error thrown in production that it was `null`. Please verify before merging that this is expected behavior. I've seen this after Ember Data has thrown an error caused by a 500 response.

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/transition.ts
+++ b/packages/@ember/-internals/routing/lib/system/transition.ts
@@ -139,7 +139,7 @@
  * where the router is transitioning to. It's important
  * to note that a `RouteInfo` is a linked list and this
  * property represents the leafmost route.
- * @property {?RouteInfo|RouteInfoWithAttributes} to
+ * @property {null|RouteInfo|RouteInfoWithAttributes} to
  * @public
  */
 
--- a/packages/@ember/-internals/routing/lib/system/transition.ts
+++ b/packages/@ember/-internals/routing/lib/system/transition.ts
@@ -139,7 +139,7 @@
  * where the router is transitioning to. It's important
  * to note that a `RouteInfo` is a linked list and this
  * property represents the leafmost route.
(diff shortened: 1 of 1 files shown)


packages/@ember/-internals/routing/lib/system/transition.ts
```diff
@@ -139,7 +139,7 @@
  * where the router is transitioning to. It's important
  * to note that a `RouteInfo` is a linked list and this
  * property represents the leafmost route.
- * @property {RouteInfo|RouteInfoWithAttributes} to
+ * @property {?RouteInfo|RouteInfoWithAttributes} to
```
rwjblue: Should we maybe say this instead:

```suggestion
 * @property {null|RouteInfo|RouteInfoWithAttributes} to
```

jelhan: I'm absolutely no expert in JSDoc but their documentation says one should use a question mark to indicate that it's nullable: https://jsdoc.app/tags-type But to be honest, I'm a little bit confused by the *non-nullable type*.

rwjblue: > JSDoc but their documentation says one should use a question mark to indicate that it's nullable

Technically, we are using yuidoc (which isn't _exactly_ the same as JSDoc though they are very similar). I'm totally willing to believe that this is the right way to document it, but I find it very confusing with the question mark (does it go on both of the types? only one?).

> But to be honest, I'm a little bit confused by the non-nullable type.

Ya, same.

jelhan: I wasn't able to find any mentioning of nullable type or `?` in yuidoc docs. Maybe that's one of the differences between yuidoc and JSDoc?

Accepted your suggestion.

locks (approved):
Thanks for the help!

PR #18451, opened 2019-10-02
[DOC beta] document `Transition.from` as nullable in JSDoc

jelhan:
The description already mentions that it could be `null`.

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/transition.ts
+++ b/packages/@ember/-internals/routing/lib/system/transition.ts
@@ -150,7 +150,7 @@
  * property represents the head node of the list.
  * In the case of an initial render, `from` will be set to
  * `null`.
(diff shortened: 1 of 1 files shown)


packages/@ember/-internals/routing/lib/system/transition.ts
```diff
@@ -150,7 +150,7 @@
  * property represents the head node of the list.
  * In the case of an initial render, `from` will be set to
  * `null`.
- * @property {RouteInfoWithAttributes} from
+ * @property {?RouteInfoWithAttributes} from
```
rwjblue: What do you think about this instead:

```suggestion
 * @property {null|RouteInfoWithAttributes} from
```

locks (approved):

PR #18457, opened 2019-10-02
[CLEANUP] remove polyfill for `history.state`

bekzod:
remove polyfill for `history.state`, it is [supported in IE 10 and others](https://developer.mozilla.org/en-US/docs/Web/API/History_API#Browser_compatibility),

### DIFF
--- a/packages/@ember/-internals/routing/lib/location/history_location.ts
+++ b/packages/@ember/-internals/routing/lib/location/history_location.ts
@@ -95,11 +95,7 @@ export default class HistoryLocation extends EmberObject implements EmberLocatio
     let history = this.history || window.history;
     set(this, 'history', history);
 
-    if (history && 'state' in history) {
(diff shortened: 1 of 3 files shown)


packages/@ember/-internals/routing/lib/location/history_location.ts
```diff
@@ -185,11 +180,7 @@ export default class HistoryLocation extends EmberObject implements EmberLocatio
     @return state {Object}
   */
   getState() {
```
rwjblue: Can we delete this method instead? Seems like it was only used as an abstraction over the polyfill?

bekzod: done

rwjblue (approved):

rwjblue:
Thank you!

PR #18461, opened 2019-10-03
Remove handling of baseURL in HistoryLocation

mehulkar:
The `baseURL` feature was [deprecated in Ember CLI 2.7](https://blog.emberjs.com/2016/04/28/baseurl).

### DIFF
--- a/packages/@ember/-internals/routing/lib/location/history_location.ts
+++ b/packages/@ember/-internals/routing/lib/location/history_location.ts
@@ -73,13 +73,6 @@ export default class HistoryLocation extends EmberObject implements EmberLocatio
   init() {
     this._super(...arguments);
 
-    let base = document.querySelector('base');
-    let baseURL: string | null = '';
-    if (base) {
-      baseURL = base.getAttribute('href');
-    }
-
-    set(this, 'baseURL', baseURL);
     set(this, 'location', this.location || window.location);
 
     this._popstateHandler = undefined;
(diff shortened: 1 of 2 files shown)


rwjblue:
ember-cli did indeed deprecate using `baseURL`, but this code supports running on a site that happens to have a `<base>` element on it and is not directly related to the deprecated ember-cli feature.

rwjblue:
If we do want to remove this, I'd guess that we'd need a deprecation cycle in Ember itself?

mehulkar:
That makes sense. Although can this feature really work in 3.x without Ember CLI support? 

Does this need an RFC?

rwjblue:
> Although can this feature really work in 3.x without Ember CLI support?

ember-cli@3.x stopped _emitting_ a `<base>` but what folks have in their own `index.html` files is completely unrealted to what ember-cli outputs. Using a `<base>` is a perfectly valid thing in normal hosting environments (though as we have found out, there are many issues/caveats).

mehulkar:
@rwjblue got it. Considering that [`<base>` is an html feature](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base), I'm assuming Ember should continue "supporting" it then, whatever that means?

Fine to close this if that's the case. I misunderstood how this tied into the CLI feature.

mehulkar:
Closing this down since supporting normal HTML is still desirable

PR #18463, opened 2019-10-04
[DOC] #18228 Add example to EmberArray.forEach documentation

Frozenfire92:
#18228

Adds an example for the EmberArray.forEachmethod

### DIFF
--- a/packages/@ember/-internals/runtime/lib/mixins/array.js
+++ b/packages/@ember/-internals/runtime/lib/mixins/array.js
@@ -658,6 +658,15 @@ const ArrayMixin = Mixin.create(Enumerable, {
     object that will be set as `this` on the context. This is a good way
     to give your iterator function access to the current object.
 
+    Example Usage:
+
+    ```javascript
+      let foods = ['apple', 'banana', carrot'];
+
+      foods.forEach(food => console.log(`You should eat a ${food}`));
+      foods.forEach((item, index, array) => console.log(`${index + 1}/${array.length}) ${item}`));
+    ```
+
     @method forEach
     @param {Function} callback The callback to execute
     @param {Object} [target] The target object to use
--- a/packages/@ember/-internals/runtime/lib/mixins/array.js
+++ b/packages/@ember/-internals/runtime/lib/mixins/array.js
@@ -664,15 +664,19 @@ const ArrayMixin = Mixin.create(Enumerable, {
     let foods = [
       { name: 'apple', eaten: false },
       { name: 'banana', eaten: false },
-      { name: carrot': eaten: false }
+      { name: 'carrot', eaten: false }
     ];
 
     foods.forEach((food) => food.eaten = true);
 
     let output = '';
     foods.forEach((item, index, array) => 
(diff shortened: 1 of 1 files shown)


packages/@ember/-internals/runtime/lib/mixins/array.js
```diff
@@ -658,6 +658,15 @@ const ArrayMixin = Mixin.create(Enumerable, {
     object that will be set as `this` on the context. This is a good way
     to give your iterator function access to the current object.
 
+    Example Usage:
+
+    ```javascript
+      let foods = ['apple', 'banana', carrot'];
+
+      foods.forEach(food => console.log(`You should eat a ${food}`));
```
locks: Hm, I think we should come up with an example that does some side effect in the `forEach` block, like setting a property in the items of the array.
What if each fruit was an object, and we had an `eaten` property that we toggle to true?

locks (changes requested):

packages/@ember/-internals/runtime/lib/mixins/array.js
```diff
@@ -658,6 +658,23 @@ const ArrayMixin = Mixin.create(Enumerable, {
...
+      { name: 'apple', eaten: false },
+      { name: 'banana', eaten: false },
+      { name: carrot': eaten: false }
+    ];
+
+    foods.forEach((food) => food.eaten = true);
+
+    let output = '';
+    foods.forEach((item, index, array) => 
+      output += `${index + 1}/${array.length}) ${item}\n`
```
```suggestion
      output += `${index + 1}/${array.length} ${item.name}\n`;
```

packages/@ember/-internals/runtime/lib/mixins/array.js
```diff
@@ -658,6 +658,23 @@ const ArrayMixin = Mixin.create(Enumerable, {
     object that will be set as `this` on the context. This is a good way
     to give your iterator function access to the current object.
 
+    Example Usage:
+
+    ```javascript
+    let foods = [
+      { name: 'apple', eaten: false },
+      { name: 'banana', eaten: false },
+      { name: carrot': eaten: false }
```
```suggestion
      { name: 'carrot', eaten: false }
```

packages/@ember/-internals/runtime/lib/mixins/array.js
```diff
@@ -658,6 +658,23 @@ const ArrayMixin = Mixin.create(Enumerable, {
...
+    ];
+
+    foods.forEach((food) => food.eaten = true);
+
+    let output = '';
+    foods.forEach((item, index, array) => 
+      output += `${index + 1}/${array.length}) ${item}\n`
+    );
```
```suggestion
    );
    console.log(output);
    // 1/3 apple
    // 2/3 banana
    // 3/3 carrot
```

locks (dismissed):

locks:
Left some suggestions :)

pzuraq (approved):
It looks like the API docs currently use native arrays for a lot of examples. I think this is not ideal, since `EmberArray` is really separate from native arrays (and we want to move more that direction), so in the future we should do a sweep and convert all examples to explicitly use `Ember.A()`, but for now I think this is fine.

PR #18465, opened 2019-10-04
[DOC] #18228 Update documentation for EmberArray.includes

Frozenfire92:
#18228

Update documentation for EmberArray.includes

### DIFF
--- a/packages/@ember/-internals/runtime/lib/mixins/array.js
+++ b/packages/@ember/-internals/runtime/lib/mixins/array.js
@@ -1234,7 +1234,7 @@ const ArrayMixin = Mixin.create(Enumerable, {
     the startAt value passed in as the index to search from. If less
     than or equal to `-1 * array.length` the entire array is searched.
 
-    This method is a Polyfill for ES 2016 Array.includes.
+    This method has the same behavior of JavaScript's [Array.includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes).
 
     ```javascript
     [1, 2, 3].includes(2);     // true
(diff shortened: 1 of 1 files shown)


packages/@ember/-internals/runtime/lib/mixins/array.js
```diff
@@ -1225,20 +1225,25 @@ const ArrayMixin = Mixin.create(Enumerable, {
...
 
     ```javascript
     [1, 2, 3].includes(2);     // true
     [1, 2, 3].includes(4);     // false
     [1, 2, 3].includes(3, 2);  // true
     [1, 2, 3].includes(3, 3);  // false
-    [1, 2, 3].includes(3, -1); // true
-    [1, 2, 3].includes(1, -1); // false
-    [1, 2, 3].includes(1, -4); // true
+    [1, 2, 3].includes(3, -1); // true, equivalent to includes(3, 2)
```
locks: I'm somewhat confused by the comments in here. I would suggest not including them, or having separate example blocks with an explanation included.

packages/@ember/-internals/runtime/lib/mixins/array.js
```diff
@@ -1225,20 +1225,25 @@ const ArrayMixin = Mixin.create(Enumerable, {
...
+    Used to determine if the array contains the passed object.
+    Returns `true` if found, `false` otherwise.
+
+    The optional `startAt` argument can be used to pass a starting
+    index to search from, effectively slicing the searchable portion 
+    of the array. If it's negative it will add the array length to
+    the startAt value passed in as the index to search from. If less
+    than or equal to `-1 * array.length` the entire array is searched.
+
     This method is a Polyfill for ES 2016 Array.includes.
```
```suggestion
    This method has the same behavior of JavaScript's [Array.includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes).
```

locks (changes requested):
Left some asks :)

RobbieTheWagner (approved):
This seems fine to me. Although, it doesn't really seem to have changed much from what we had before. @locks thoughts?

locks (approved):
Looks good!

locks:
Keep up the good work @Frozenfire92 😁

PR #18467, opened 2019-10-04
Add Comment for Describe

masitings:


config/browserlists.js
```diff
@@ -6,6 +7,7 @@ const allSupportedBrowsers = [
   'ie 11',
 ];
 
+// Browser that can adapt and running
```
rwjblue: This isn't quite right, this is the set of browsers we use for the prebuilt assets (though the application can configure whichever ones it wants).

config/browserlists.js
```diff
@@ -1,3 +1,4 @@
+// The list of supported browsers
```
rwjblue: I'd tweak this to say `The list of all supported browsers`

rwjblue (changes requested):
Thanks for the PR! I've left a few inline comments for tweaks.

Also wanted to mention, that its _generally_ not a good idea to send PR's from your forks `master` branch directly. Usually you'd want to make a branch in your fork and send the PR from there. No worries in this case, but figured I'd mention it in case you weren't familiar with the workflow. 😸

config/browserlists.js
Alonski: Do we really need this though?
A variable named `fullName` does not need a comment above it that says `The full name` right?

MelSumner:
This is not a valid PR. Closing issue.

PR #17622, opened 2019-02-15
[WIP] [DOCS]

CDNRae:
As per Issue #16518: added a more detailed explanation of the isActive method in Router Services, along with an example of how to use it.  Would love some feedback on how to improve what I've added!

PR #18471, opened 2019-10-06
[DOC] Update explanation of router service `isActive` method and add code example

pauljeannot:
Part of #16518

Took the [nice work done](https://github.com/emberjs/ember.js/pull/17622) by @rscotchmer, and updated it with new class syntax.

### DIFF
--- a/packages/@ember/-internals/routing/lib/services/router.ts
+++ b/packages/@ember/-internals/routing/lib/services/router.ts
@@ -244,12 +244,33 @@ export default class RouterService extends Service {
   }
 
   /**
-     Determines whether a route is active.
+     Returns `true` if `routeName/models/queryParams` is the active route, where `models` and `queryParams` are optional.
+     See [model](api/ember/release/classes/Route/methods/model?anchor=model) and
+     [queryParams](/api/ember/3.7/classes/Route/properties/queryParams?anchor=queryParams) for more information about these arguments.
+
+     In the following example, isActive will return `true` if the current route is `/posts`.
+     ```app/components/posts.js
+     import Component from '@glimmer/component';
+     import { inject as service } from '@ember/service';
+
+     export default class extends Component {
+       @service router;
+
+       displayComments() {
+         return this.get('router').isActive('posts');
+       }
+     });
+     ```
+
+     Usage example  with a model:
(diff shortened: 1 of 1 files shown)


packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -244,12 +244,33 @@ export default class RouterService extends Service {
...
+       @service router;
+
+       displayComments() {
+         return this.get('router').isActive('posts');
+       }
+     });
+     ```
+
+     Usage example  with a model:
+     ```javascript return this.get('router').isActive('posts', post.id)```
```
rwjblue: I think we need some newlines here (after "javascript" and before the ending backticks at least)

packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -244,12 +244,33 @@ export default class RouterService extends Service {
   }
 
   /**
-     Determines whether a route is active.
+     Returns `true` if `routeName/models/queryParams` is the active route, where `models` and `queryParams` are optional.
+     See [model](api/ember/release/classes/Route/methods/model?anchor=model) and
+     [queryParams](/api/ember/3.7/classes/Route/properties/queryParams?anchor=queryParams) for more information about these arguments.
+
+     In the following example, isActive will return `true` if the current route is `/posts`.
+     ```app/components/posts.js
```
locks:
```suggestion

     ```app/components/posts.js
```

packages/@ember/-internals/routing/lib/services/router.ts
locks:
```suggestion
     
     ```javascript
     return this.get('router').isActive('posts', post.id)
     ```
```

locks: We also try to make the code samples as complete as possible. Can you add some more context to it so it's easier to understand where it goes?

locks (changes requested):
Left some inline asks :)

pzuraq (commented):
This generally looks good, just need to fix the formatting around the examples (per @locks suggestions). More context for the examples would be good as well.

locks:
Thank you for your help!

PR #18476, opened 2019-10-09
[BUGFIX release] Ensure model can be observed by sync observers

### LINKED ISSUE #18427: [3.13] Regression when observing model changes in controller
If you can't understand the issue title, [here is a reproduction repo](https://github.com/boris-petrov/ember-controller-recompute-bug).

The last commit updates Ember to 3.13.1. Run it and open `http://localhost:4200/chat`. You'll see `recomputing model!` in the console. If you click the `root` link (which just changes the route) and then click the browser's back button, you'll see `recomputing model!` again. If you repeat that multiple times, you'll see the same message.

Checking out the previous commit which uses Ember 3.12 and repeating the same steps produces a different result - the observer is not triggered at all between route changes and the message is not displayed multiple times. I believe this is the correct behavior.

pzuraq:
This is a compatibility change, otherwise sync observers
will not be able to watch the model property of controllers
at all. Also fixes the naming of the final flag to Observable#addObserver
and Observable#removeObserver

Fixes #18427

### DIFF
--- a/packages/@ember/controller/lib/controller_mixin.js
+++ b/packages/@ember/controller/lib/controller_mixin.js
@@ -1,6 +1,8 @@
-import { Mixin, tracked } from '@ember/-internals/metal';
+import { Mixin, computed } from '@ember/-internals/metal';
 import { ActionHandler } from '@ember/-internals/runtime';
-import { EMBER_METAL_TRACKED_PROPERTIES } from '@ember/canary-features';
+import { symbol } from '@ember/-internals/utils';
+
+const MODEL = symbol('MODEL');
 
 /**
 @module ember
@@ -43,5 +45,13 @@ export default Mixin.create(ActionHandler, {
     @property model
     @public
   */
-  model: EMBER_METAL_TRACKED_PROPERTIES ? tracked() : null,
+  model: computed({
+    get() {
+      return this[MODEL];
+    },
+
+    set(key, value) {
+      return (this[MODEL] = value);
+    },
+  }),
 });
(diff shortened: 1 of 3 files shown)


packages/@ember/controller/tests/controller_test.js
```diff
@@ -3,7 +3,70 @@ import Service, { inject as injectService } from '@ember/service';
...
+      await this.visit('/');
+
+      this.assertText('1');
+
+      runTask(() => this.$('button').click());
+      this.assertText('2');
+    }
+
+    async '@test model can be observed with sync observers'(assert) {
+      let observerDidRun = false;
```
chancancode: nit: may as well make this a counter and assert that it fired exactly once

packages/@ember/controller/lib/controller_mixin.js
```diff
@@ -43,5 +46,13 @@ export default Mixin.create(ActionHandler, {
     @property model
     @public
   */
-  model: EMBER_METAL_TRACKED_PROPERTIES ? tracked() : null,
+  model: computed({
+    get() {
+      return this[MODEL];
```
chancancode: We may want to make this non-enumerable if there is an easy way to do it, but if it's too annoying it's not a big deal.

chancancode (approved):
Seems good to me

packages/@ember/controller/lib/controller_mixin.js
pzuraq: The only way to do it currently is something I think we'd rather deprecate sooner rather than later (it requires some not-so-nice prototype chain manipulation), so I'd rather not for the moment

PR #18477, opened 2019-10-09
[BUGFIX release] Allows @each to work with arrays that contain falsy …

### LINKED ISSUE #18474: [3.13] Regression when using @each on an array with undefined elements
[Here is a reproduction repo](https://github.com/boris-petrov/ember-each-bug)

The last commit downgrades Ember to 3.12 and the example works. The previous is on Ember 3.13 and does not.

Having a computed property that depends on an array with `undefined` in it by using `@each` is broken in 3.13.

P.S. The error is something like `When using @each to observe the array ``, the items in the array must be objects`.

pzuraq:
…values

This matches the original behavior as seen in the each_proxy.

Fixes #18474

### DIFF
--- a/packages/@ember/-internals/runtime/tests/system/object/computed_test.js
+++ b/packages/@ember/-internals/runtime/tests/system/object/computed_test.js
@@ -457,6 +457,28 @@ moduleFor(
       assert.deepEqual(n.normalized, []);
     }
 
+    ['@test @each works on array with falsy values'](assert) {
+      let obj = EmberObject.extend({
+        falsy: [null, undefined, false, '', 0, {}],
+        truthy: [true, 'foo', 123],
+
+        falsyComputed: computed('falsy.@each.foo', () => {
+          assert.ok(true, 'falsy computed');
+        }),
+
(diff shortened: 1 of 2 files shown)


RobbieTheWagner (approved):

packages/@ember/-internals/runtime/tests/system/object/computed_test.js
```diff
@@ -457,6 +457,28 @@ moduleFor(
...
+
+        truthyComputed: computed('truthy.@each.foo', () => {
+          assert.ok(true, 'truthy computed');
+        }),
+      }).create();
+
+      // should throw no errors
+      obj.falsyComputed;
+
+      assert.throws(() => {
```
rwjblue: Shouldn’t this be expectAssertion?

pzuraq: Ah, correct, wonder why prod tests didn't catch that

rwjblue:
Makes me worry that they are broken somehow :thinking:

We should investigate...

PR #18480, opened 2019-10-11
[Failing Test] {{on}} modifier removes valueless attributes

kturney:
Noticed today that using the {{on}} modifier on an element with the `required` attribute removed the attribute.

e.g.
template: `<input type='text' required {{on 'input' this.onInput}} >`
resulted in
DOM: `<input type='text' >`

However, adding a value to the attribute caused it to remain.

e.g.
template: `<input type='text' required='true' {{on 'input' this.onInput}} >`
resulted in DOM
DOM: `<input type='text' required >`

While adding a failing test for `required`, I also added checks for other boolean attributes.
`autofocus`, `required`, and `disabled` are all removed, while `readonly` is preserved.

I believe the problem is with the {{on}} modifier because
template: `<input type='text' required >`
results in
DOM: `<input type='text' required >`.

I looked through the {{on}} modifier source and nothing jumped out at me, so maybe it isn't directly an {{on}} problem though?

Side Note:
I made the `disabled` element a `button` to show that it is not only `input` that is affected.

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/modifiers/on-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/modifiers/on-test.js
@@ -374,13 +374,14 @@ moduleFor(
       this.assertCounts({ adds: 1, removes: 1 });
     }
 
-    '@test it respects boolean attributes'() {
+    '@test it respects valueless attributes'() {
       this.render(
         `
-        <input id='autofocus' autofocus {{on "input" this.callback}} />
-        <input id='readonly' readonly {{on "input" this.callback}} />
-        <input id='required' required {{on "input" this.callback}} />
-        <button id='disabled' disabled {{on "click" this.callback}}>Click Me</button> />
+        <input id='autofocus' autofocus {{on 'input' this.callback}} />
+        <input id='readonly' readonly {{on 'input' this.callback}} />
+        <input id='required' required {{on 'input' this.callback}} />
+        <button id='disabled' disabled {{on 'click' this.callback}}></button>
+        <video controls {{on 'canplay' this.callback once=true}}></video>
       `,
         {
           callback() {},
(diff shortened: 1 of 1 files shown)


pzuraq:
This appears to occur with any modifier except `{{action}}`, which leads me to believe this is a VM issue.

chancancode:
We can fix it with an AST transform until we finish the upgrade, which can also be backported to LTS

rwjblue:
VM upgrade is finished...

kturney:
seems the same as https://github.com/emberjs/ember.js/issues/18712

rwjblue:
Can you confirm that this is _not_ about boolean attributes, but is actually about valueless attributes (which are often boolean, 😸)?

kturney:
Would adding a video element with a controls attribute to the test confirm this? I didn't realize there was a distinction.

rwjblue:
@kturney ya, any attribute would do, even something like `<button data-foo {{on 'click' this.wahtever}}>Click Me!</button>`. If that works or not will point towards different types of fixes needed...

kturney:
@rwjblue I added a video element with a controls attribute to the test.

kategengler:
Is this still relevant?

NullVoxPopuli:
looks like this was fixed a while ago 

[repl](https://limber.glimdown.com/edit?c=JYWwDg9gTgLgBAbzhAdnAvnAZlCI4DkAAgKYgBGJUA9CBACbBbBUEDcAUBwDYnwDGAQ27dyg-gGs4AXjgAKAJQyAfInRcAPDDJhug7co5w4G4CjABXeMHrSCgqxCwR%2BFgM4E4DmE5fvECKiEZpYwnkIiYpLomNSGxqbmVnA2dlAkgvSo3ACenumZ2TkBQQQhVuHCouISMXBxRibl1rYE6QCOFsDp9Pkknd0k9CVoZUlhcBHV0bHxJuRWPmipBIxuguS8vXBrG1sjhPzcwJKVkTUxyhrUCzBLcxoAbjYkEJOoMLjcbgcEQuZ6PKTKpRKSofgkaSfCwkS7XZ70V6Ga7acB6AxwIA&format=gjs)

image

PR #18481, opened 2019-10-11
wrap validateType in DEBUG flag

bekzod:


packages/@ember/application/globals-resolver.js
```diff
@@ -7,10 +7,34 @@ import { get, findNamespace } from '@ember/-internals/metal';
 import { assert, info } from '@ember/debug';
 import { capitalize, classify, dasherize, decamelize } from '@ember/string';
 import { Object as EmberObject } from '@ember/-internals/runtime';
-import validateType from './lib/validate-type';
 import { getTemplate } from '@ember/-internals/glimmer';
 import { DEBUG } from '@glimmer/env';
 
+function validateType(resolvedType, parsedName) {
```
rwjblue: This seems to only be used in the one location below, can we remove this function and inline the implementation down there instead? (allows us to strip out the function as well)

packages/@ember/application/globals-resolver.js
```diff
@@ -144,10 +168,9 @@ class DefaultResolver extends EmberObject {
       if (parsedName.root && parsedName.root.LOG_RESOLVER) {
         this._logLookup(resolved, parsedName);
       }
-    }
-
-    if (resolved) {
-      validateType(resolved, parsedName);
+      if (resolved) {
+        validateType(resolved, parsedName);
+      }
```
rwjblue: Something like this:

```js
if (DEBUG && resolved) {
  let VALIDATED_TYPES = {
    route: ['isRouteFactory', 'Ember.Route'],
    component: ['isComponentFactory', 'Ember.Component'],
    view: ['isViewFactory', 'Ember.View'],
    service: ['isServiceFactory', 'Ember.Service'],
  };

  let validationAttributes = VALIDATED_TYPES[parsedName.type];

  if (validationAttributes) {
    let [factoryFlag, expectedType] = validationAttributes;

    assert(
      `Expected ${parsedName.fullName} to resolve to an ${expectedType} but ` +
        `instead it was ${resolvedType}.`,
      Boolean(resolvedType[factoryFlag])
    );
  } 
}
```

packages/@ember/application/globals-resolver.js
bekzod: yeah I was going to do that at first, but then wanted to confirm if that is fine

rwjblue (approved):

rwjblue:
Restarted browser stack job.

rwjblue:
Thank you!

PR #18482, opened 2019-10-11
[BUGFIX] allow firstObject/lastObject references in {{array}} helper

### LINKED ISSUE #18455: Error when using `{{items.firstObject}}` with `array` helper
The following template seems to throw a run time error in 3.14.0-beta.3:

```hbs
{{#let (array "one" "two" "three") as |items|}}
  first: {{items.firstObject}}
{{/let}}
```

The error is: `TypeError: Cannot read property 'tag' of undefined`

Screenshot 2019-10-02 at 17 08 18

bekzod:
fixes #18455

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/helpers/array.ts
+++ b/packages/@ember/-internals/glimmer/lib/helpers/array.ts
@@ -1,6 +1,7 @@
 import { PathReference } from '@glimmer/reference';
 import { Arguments, VM } from '@glimmer/runtime';
 import { Opaque } from '@glimmer/util';
+import { EmberCapturedArrayReference } from '../utils/references';
 
 /**
 @module ember
@@ -43,5 +44,5 @@ import { Opaque } from '@glimmer/util';
  */
 
 export default function(_vm: VM, args: Arguments): PathReference<Opaque[]> {
-  return args.positional.capture();
+  return new EmberCapturedArrayReference(args.positional.capture());
 }
--- a/packages/@ember/-internals/glimmer/lib/utils/references.ts
+++ b/packages/@ember/-internals/glimmer/lib/utils/references.ts
@@ -33,6 +33,7 @@ import {
 } from '@glimmer/reference';
 import {
   CapturedArguments,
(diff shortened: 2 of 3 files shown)


packages/@ember/-internals/glimmer/lib/helpers/array.ts
```diff
@@ -43,5 +44,5 @@ import { Opaque } from '@glimmer/util';
  */
 
 export default function(_vm: VM, args: Arguments): PathReference<Opaque[]> {
-  return args.positional.capture();
+  return new EmberCapturedArrayReference(args.positional.capture());
```
rwjblue: Hmm, do we need to make a whole new reference type for this? How does `{{this.foo.firstObject}}`/`{{this.foo.lastObject}}` work (unrelated to `{{array`) when `this.foo` is a "normal array"?

Why do the normal computed properties ([see here](https://github.com/emberjs/ember.js/blob/9a09374ef0f90d5239ebf2f538c902fb1fb04c12/packages/%40ember/-internals/runtime/lib/mixins/array.js#L323-L325)) for `firstObject`/`lastObject` that are added to `Array.prototype` not work automatically?

bekzod: I think the reason it does not work is: for array from component/controller it creates ObjectReference (RootReference), for array from `{{array}}` helper it is just array of references so `firstObject/lastObject` does not exist for it. So simplest way is just to create another _Reference_ to translate `firstObject/lastObject` to `[0]/[length -1]`,  if I have better solution I will make followup PR to improve it 

>  How does {{this.foo.firstObject}}/{{this.foo.lastObject}} work (unrelated to {{array) when this.foo is a "normal array"?
`{{this.foo.firstObject}}/{{this.foo.lastObject}}` should be good, since it uses `RootReference`
 there are tests covers for it too I think

locks:
Sorry for letting this linger. At this point I think it would be easier to remake the PR if a contributor is interested in this feature.

Issue #15635, opened 2017-09-05
[BUG 2.15] Number literals bigger than 9 digits break


cibernox:
This is a good one.

If you pass a numeric reference to a helper/component, it works with numbers up to 9 digits.
Number with 10 or 11 digits break glimmer with `TypeError: reference.value is not a function`
If the number has 12 digits it doesn't throw an exception, but the value received on the other side of the reference is `undefined`

I believe we have an access-out-of-bounds on the glimmer VM.

Quick repo to reproduce: https://github.com/cibernox/big-num-bug

Note that this only happens with number literals `{{my-helper 123123123123}}` not with bound values `{{my-helper num}}` no matter how big the value inside `num` is.

tomdale:
I think this is related to how the Glimmer compiler encodes primitive literals. JavaScript coerces numbers into 32-bit integers when applying bitwise operations, which we use to store type information (boolean, string, object, etc.) about the literals in the lower 3 bits.

In practice, this means that even though JavaScript allows you to represent values up to 2<sup>53</sup>, our combination of a) using bitwise operators and b) utilizing the lower 3 bits means that we can't represent values greater than 2<sup>29</sup>.

One way to deal with this would be to have a less-efficient primitive type (`BigNum`?) for >29-bit integers, although maybe it makes sense to have an external number pool just like we do with strings. Maybe @wycats or @chancancode can chime in since this is a system they came up with.

cibernox:
FWIW, it is rather unlikely that someone will hardcode number that big in real code, such numbers will usually come from somewhere as a binding, but it's not unusual to do it on integration tests.

rwjblue:
I made a twiddle reproduction for those interested in digging into it without cloning:

https://ember-twiddle.com/98647deda69ba3c58a58b1c6decb0add?openFiles=templates.application.hbs%2C

wycats:
I think the solution to this problem is to use a constant for large numbers, just as we do for floats.

We could also use more operand slots to get a bigger number, but I'm tempted to say that the optimization at work here can be limited to relatively small (29 bit) integers.

pixelhandler:
@cibernox so you think we need to document the current upper limit for a number at 2<sup>29</sup> ? At least until there is support for a bigger integer? cc\ @locks

cibernox:
I'd ask first some glimmer-vm expert for an estimation of how big this change will be, and if they think it can be fixed within a few weeks, I wouldn't document it.

[...]

PR #18484, opened 2019-10-12
[Bug Fix: 15635]  Fix Displaying of number literals bigger than 9 digits

saravanak:
Fixes  https://github.com/emberjs/ember.js/issues/15635. 

What: 
As suggested here https://github.com/emberjs/ember.js/issues/15635#issuecomment-487258966, 
we transform the AST for big number literals into (-parse-int "1546416000000").

I've used `String(number).length > 9` to identify cases where we need to do the transform.

Added tests for the AST Transform, and the new internal `-parse-int` helper.

@chancancode Can you have a look at this?

I am not sure if the target branch is right though...

Need some pointers on 

 - Updating the guides (This is internal, so imho, we don't need to update the documentation)


packages/ember-template-compiler/lib/plugins/transform-big-number-into-parse-int-helper.ts
```diff
@@ -0,0 +1,51 @@
...
+  let { builders: b } = env.syntax;
+
+  return {
+    name: 'transform-big-number-into-parse-int-helper',
+
+    visitor: {
+
+      NumberLiteral(number: AST.NumberLiteral): AST.Node {
+        //See https://github.com/emberjs/ember.js/issues/15635 for the choice for [9].
+        if(String(number.original).length > 9) {
```
Mooninaut: Needs to use `length > 8` The bug can occur with values >= 536870912 and < 1000000000, for example 678881280.

packages/@ember/-internals/glimmer/lib/helpers/-parse-int.ts
```diff
@@ -0,0 +1,11 @@
+import { Arguments, VM } from '@glimmer/runtime';
+import { InternalHelperReference } from '../utils/references';
+
+function parseIntHelper({ positional }: any) {
```
chancancode:
```suggestion
function parseIntHelper({ positional }: CapturedArguments) {
```

You'll have to import `CapturedArguments` from `@glimmer/interfaces`

packages/@ember/-internals/glimmer/lib/helpers/-parse-int.ts
```diff
@@ -0,0 +1,11 @@
...
+  let value = positional.at(0).value();
+  return parseInt(value);
```
chancancode:
```suggestion
  return parseInt(value, 10);
```

packages/@ember/-internals/glimmer/tests/integration/helpers/-parse-int-test.js
```diff
@@ -0,0 +1,14 @@
...
+  class extends RenderingTestCase {
+    ['@test parseInt helper']() {
+      this.render(`{{-parse-int "123456789"}}`, {
+        someTruth: true,
+      });
+
+      this.assertTextNode(this.firstChild, '123456789');
+    }
+  }
+);
```
chancancode: Instead of having a unit test for the helper, I would prefer that we have an acceptance/integration test for being able to render/pass big numbers.

packages/ember-template-compiler/lib/plugins/transform-big-number-into-parse-int-helper.ts
chancancode: I think we should just parse the number with `parseInt` and test the actual value, instead of the length of the string

packages/ember-template-compiler/tests/plugins/transform-big-number-to-parse-int-variation-test.js
```diff
@@ -0,0 +1,12 @@
+import TransformTestCase from '../utils/transform-test-case';
+import { moduleFor } from 'internal-test-helpers';
+
+moduleFor(
+  'ember-template-compiler: transforming big numbers into parseInt form',
+  class extends TransformTestCase {
+    ['@test it transforms an big number to the parseInt form']() {
+      this.assertTransformed(`{{foo-helper 1234567890}}`, `{{foo-helper (-parse-int "1234567890")}}`);
+      this.assertTransformed(`{{foo-helper 123456789}}`, `{{foo-helper 123456789}}`);
```
chancancode: Can you add a few more test for things like `{{1234567890}}`, `{{some-helper (another-helper 1234567890)}}`?

chancancode (changes requested):
Overall looks good to me, left some comments for things that need to be fixed before merging

packages/@ember/-internals/glimmer/tests/integration/helpers/-parse-int-test.js
chancancode: https://github.com/emberjs/ember.js/blob/master/packages/@ember/-internals/glimmer/tests/integration/content-test.js would probably be a good place for it

packages/ember-template-compiler/lib/plugins/transform-big-number-into-parse-int-helper.ts
chancancode: We need to check for floats as well, maybe just use `parseFloat` here, check that it is an integer (`parseFloat(...) % 1 === 0`?) and is out of bounds, both the upper and lower (negative) bounds.

lifeart: or `new Number('1').valueOf()`

saravanak: > Overall looks good to me, left some comments for things that need to be fixed before merging

@chancancode @lifeart @Mooninaut, 

Thanks for the comments...
 
Given that it was an old bug, I should have checked the replication before attempting the PR; my bad. 

Here are the results though:

From the [twiddle](https://ember-twiddle.com/98647deda69ba3c58a58b1c6decb0add?openFiles=templates.application.hbs%2C) shared by @rwjblue in the bug description, I had 

```mustache
{{go-boom-for-big-numbers 1234567893}}
{{go-boom-for-big-numbers 12345678933}}
<div/>
{{go-boom-for-big-numbers 1234567893232323.2323}}
{{go-boom-for-big-numbers 1234567893323232323.2323}}
<div/>
{{go-boom-for-big-numbers -1234567893232323.2323}}
{{go-boom-for-big-numbers -1234567893323232323.2323}}
<div/>
{{go-boom-for-big-numbers -1234567893}}
{{go-boom-for-big-numbers -12345678933}}
```

The following are the results: 

**Ember 3.0.0**
(No display, console error) : `Uncaught TypeError: reference$$1.value is not a function`

**Ember 3.1.4 and newer**
```
160826069 12345678933
1234567893232323.2 1234567893323232300
-1234567893232323.2 -1234567893323232300
-160826069 -12345678933
```

There is also an error in the 3.1.4 results: `1234567893` is displayed as `160826069`. I think this should be treated as a glimmer-vm bug and tracked separately. 

Looks like this is fixed 'upstream' in newer versions of glimmer-vm. I was not able to locate the actual diff that fixed this in the glimmer-vm code. 

This makes PR obsolete, as the relevant fixes are done on the glimmer-vm code.   

I'd like a go-ahead and on continuation of this PR if its still needed..

saravanak:
> Overall looks good to me, left some comments for things that need to be fixed before merging

@chancancode @lifeart @Mooninaut, 

Thanks for the comments...
 
Given that it was an old bug, I should have checked the replication before attempting the PR; my bad. 

Here are the results though:

From the [twiddle](https://ember-twiddle.com/98647deda69ba3c58a58b1c6decb0add?openFiles=templates.application.hbs%2C) shared by @rwjblue in the bug description, I had 

```mustache
{{go-boom-for-big-numbers 1234567893}}
{{go-boom-for-big-numbers 12345678933}}
<div/>
{{go-boom-for-big-numbers 1234567893232323.2323}}
{{go-boom-for-big-numbers 1234567893323232323.2323}}
<div/>
{{go-boom-for-big-numbers -1234567893232323.2323}}
{{go-boom-for-big-numbers -1234567893323232323.2323}}
<div/>
{{go-boom-for-big-numbers -1234567893}}
{{go-boom-for-big-numbers -12345678933}}
```

The following are the results: 

**Ember 3.0.0**
(No display, console error) : `Uncaught TypeError: reference$$1.value is not a function`

**Ember 3.1.4 and newer**
```
160826069 12345678933
1234567893232323.2 1234567893323232300
-1234567893232323.2 -1234567893323232300
-160826069 -12345678933
```

There is also an error in the 3.1.4 results: `1234567893` is displayed as `160826069`. I think this should be treated as a glimmer-vm bug and tracked separately. 

Looks like this is fixed 'upstream' in newer versions of glimmer-vm. I was not able to locate the actual diff that fixed this in the glimmer-vm code. 

This makes PR obsolete, as the relevant fixes are done on the glimmer-vm code.   

I'd like a go-ahead and on continuation of this PR if its still needed..

chancancode:
Rather than a test for the transform, maybe we should just have a set of runtime tests like this:

```js
let i = Number.MAX_SAFE_INTEGER;

while (i >= 1) {
  test(`it can handle ${i}`, function() {
    this.registerHelper('is-equal-integer', function([actual, expected]) {
      return typeof actual === 'number' && actual === expected;
    });

    this.render(`[{{${i}}}][{{this.i}}][{{is-equal-integer ${i} this.i}}]`, { i });

    this.assertText(`[${i}][${i}][true]`);
  });

  i = Math.round(i / 2);
};

let j = Number.MIN_SAFE_INTEGER;

while (j <= -1) {
  test(`it can handle ${j}`, function() {
    this.registerHelper('is-equal-integer', function([actual, expected]) {
      return typeof actual === 'number' && actual === expected;
    });

    this.render(`[{{${j}}}][{{this.j}}][{{is-equal-integer ${j} this.j}}]`, { j });

    this.assertText(`[${j}][${j}][true]`);
  });

  j = Math.round(j / 2);
};
```

PR #18499, opened 2019-10-21
Use all defined BrowserStack launchers in CI

kategengler:
I noticed that https://github.com/emberjs/ember.js/pull/16755 added a Googlebot-like launcher, but did not add it to the list of launchers that would run. 

This change makes it so all launchers defined are run. It removes the Googlebot-like, since Google has moved to an evergreen Chrome since that time. It also removes other unused launcher definitions.


pzuraq (approved):

pzuraq:
This looks good, and Firefox/Chrome are covered by the other tests so no need to run BrowserStack against them, correct?

kategengler:
Correct

pzuraq:
Awesome, thanks for doing this!

PR #18500, opened 2019-10-21
[BUGFIX release] Remove requirement for jquery in Octane

pzuraq:
Based on community feedback and early testing, the core team
has decided that disabling jquery-integration does not need to be a
requirement for enabling Octane mode, so we can remove it from
the error messages.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -79,9 +79,22 @@ module.exports = {
         );
       }
 
-      if (optionalFeaturesMissing || optionalFeatures.isFeatureEnabled('jquery-integration')) {
+      if (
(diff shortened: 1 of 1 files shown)


lib/index.js
```diff
@@ -79,12 +79,6 @@ module.exports = {
         );
       }
 
-      if (optionalFeaturesMissing || optionalFeatures.isFeatureEnabled('jquery-integration')) {
```
rwjblue: Instead of removing this, we should ensure that the flag is set (not `undefined`).

rwjblue (approved):

PR #18503, opened 2019-10-23
[DOC] add code example for EmberArray reject

mcfiredrill:
part of #18228

### DIFF
--- a/packages/@ember/-internals/runtime/lib/mixins/array.js
+++ b/packages/@ember/-internals/runtime/lib/mixins/array.js
@@ -857,7 +857,7 @@ const ArrayMixin = Mixin.create(Enumerable, {
 
     ```javascript
     const food = [
-      { food: 'apple', isFruit: true},
+      { food: 'apple', isFruit: true },
       { food: 'bread', isFruit: false },
       { food: 'banana', isFruit: true }
     ];
--- a/packages/@ember/-internals/runtime/lib/mixins/array.js
+++ b/packages/@ember/-internals/runtime/lib/mixins/array.js
@@ -863,7 +863,7 @@ const ArrayMixin = Mixin.create(Enumerable, {
     ];
     const nonFruits = food.reject(function(thing) {
       return thing.isFruit;
-    }); // [{food: 'beans', isFruit: false}]
+    }); // [{food: 'bread', isFruit: false}]
     ```
 
     @method reject
--- a/packages/@ember/-internals/runtime/lib/mixins/array.js
+++ b/packages/@ember/-internals/runtime/lib/mixins/array.js
@@ -856,7 +856,11 @@ const ArrayMixin = Mixin.create(Enumerable, {
     Example Usage:
 
     ```javascript
-    const food = Ember.A().addObjects([{food: 'apple', isFruit: true}, {food: 'bread', isFruit: false}, food: 'banana', isFruit: true}]);
+    const food = [
+      { food: 'apple', isFruit: true},
+      { food: 'bread', isFruit: false },
+      { food: 'banana', isFruit: true }
+    ];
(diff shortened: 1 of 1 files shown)


packages/@ember/-internals/runtime/lib/mixins/array.js
```diff
@@ -278,7 +278,7 @@ const ArrayMixin = Mixin.create(Enumerable, {
 
     ```javascript
     let peopleToMoon = ['Armstrong', 'Aldrin'];
-    
+
```
mcfiredrill: Sorry for the noise, my editor removed the whitespace in this file.

packages/@ember/-internals/runtime/lib/mixins/array.js
```diff
@@ -853,6 +853,15 @@ const ArrayMixin = Mixin.create(Enumerable, {
     object that will be set as "this" on the context. This is a good way
     to give your iterator function access to the current object.
 
+    Example Usage:
+
+    ```javascript
+    const food = Ember.A().addObjects([{food: 'apple', isFruit: true}, {food: 'bread', isFruit: false}, food: 'banana', isFruit: true}]);
```
mcfiredrill: Is this line too long?

locks: I would format it so that there is one object per line.
More importantly, if you want to create an EmberArray with items, you can do `Ember.A([1,2,3])`, you don't need to create the array and add objects. The other methods assume that prototype extensions are enabled, so in this example you can define a regular array and  it will work in applications:

```suggestion
    const food = [
      { food: 'apple', isFruit: true},
      { food: 'bread', isFruit: false },
      { food: 'banana', isFruit: true }
    ];
```

packages/@ember/-internals/runtime/lib/mixins/array.js
locks: It's helpful, because those whitespaces shouldn't be there in the first place!

locks (dismissed):
Left an inline ask

packages/@ember/-internals/runtime/lib/mixins/array.js
mcfiredrill: Awesome, thank you!

packages/@ember/-internals/runtime/lib/mixins/array.js
```diff
@@ -853,6 +853,19 @@ const ArrayMixin = Mixin.create(Enumerable, {
...
+
+    ```javascript
+    const food = [
+      { food: 'apple', isFruit: true},
+      { food: 'bread', isFruit: false },
+      { food: 'banana', isFruit: true }
+    ];
+    const nonFruits = food.reject(function(thing) {
+      return thing.isFruit;
+    }); // [{food: 'beans', isFruit: false}]
```
rwwagner90: This should be `bread` not `beans` right?

RobbieTheWagner (changes requested):

packages/@ember/-internals/runtime/lib/mixins/array.js
mcfiredrill: I suppose to be consistent with my other pull requests, yes. 🍞

mcfiredrill: Ah sorry, I see my mistake now. I will fix it.

mcfiredrill: Fixed here https://github.com/emberjs/ember.js/pull/18503/commits/ed2fe07eac1f94efd5e849233ceeeba95776c9b3

RobbieTheWagner (approved):

PR #18504, opened 2019-10-23
[DOCS] Update EmberArray invoke docs to use native class

Charizard:
The documentation for EmberArray#invoke was already added by @mukilane as part of #18283 but this PR just changes it to use native class.


locks (approved):
Thanks for the help ;)

Issue #18459, opened 2019-10-03
[Deprecation Tests] Usage of window.ENV in tests/index.html


Alonski:
A recent PR introduced a deprecation for using window.ENV.
Tests Index.html is using this:
https://github.com/emberjs/ember.js/blob/master/tests/index.html#L41

Caused by: #18441

rwjblue:
@astronomersiva - Think you might have a minute to shoot off a PR fixing?

rwjblue:
Closed by #18521

PR #18521, opened 2019-10-25
[CLEANUP] ISSUE-18459: chore: updating ENV to EmberENV and fixing tests

Mikek2252:
#18459 Removing deprecation and updating associated tests.


rwjblue:
Thank you!

PR #18523, opened 2019-10-27
[DOC] New component syntax for getOwner example

esbanarango:
This PR updates the usage example for [getOwner](http://api.emberjs.com/ember/3.13/functions/@ember%2Fapplication/getOwner).

### DIFF
--- a/packages/@ember/-internals/owner/index.ts
+++ b/packages/@ember/-internals/owner/index.ts
@@ -61,19 +61,20 @@ export const OWNER = symbol('OWNER');
 
   // Usage:
   //
-  //   {{play-audio audioType=model.audioType audioFile=model.file}}
+  //   <PlayAudio @audioType={{model.audioType}} @audioFile={{model.file}}/>
   //
-  export default Component.extend({
-    audioService: computed('audioType', function() {
+  export default class extends Component {
+    @computed('audioType')
+    get audioService() {
       let owner = getOwner(this);
       return owner.lookup(`service:${this.get('audioType')}`);
-    }),
+    }
 
     click() {
       let player = this.get('audioService');
       player.play(this.get('audioFile'));
     }
-  });
+  }
   ```
 
   @method getOwner
--- a/packages/@ember/-internals/owner/index.ts
+++ b/packages/@ember/-internals/owner/index.ts
@@ -52,7 +52,7 @@ export const OWNER = symbol('OWNER');
   into the owner.
 
   For example, this component dynamically looks up a service based on the
-  `audioType` passed as an attribute:
+  `audioType` passed as an argument:
 
   ```app/components/play-audio.js
   import Component from '@glimmer/component';
@@ -60,7 +60,7 @@ export const OWNER = symbol('OWNER');
 
   // Usage:
   //
-  //   <PlayAudio @audioType={{model.audioType}} @audioFile={{model.file}}/>
+  //   <PlayAudio @audioType={{@model.audioType}} @audioFile={{@model.file}}/>
   //
   export default class extends Component {
     get audioService() {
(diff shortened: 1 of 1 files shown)


pzuraq (approved):

packages/@ember/-internals/owner/index.ts
```diff
@@ -61,19 +61,20 @@ export const OWNER = symbol('OWNER');
...
-  //   {{play-audio audioType=model.audioType audioFile=model.file}}
+  //   <PlayAudio @audioType={{model.audioType}} @audioFile={{model.file}}/>
   //
-  export default Component.extend({
-    audioService: computed('audioType', function() {
+  export default class extends Component {
+    @computed('audioType')
+    get audioService() {
       let owner = getOwner(this);
       return owner.lookup(`service:${this.get('audioType')}`);
```
pzuraq: Can we update this to be a Glimmer component as well? E.g. `this.args.audioType`

pzuraq (changes requested):

esbanarango:
@pzuraq Thank you for the reviewed. I just addressed the comments 🙏.

packages/@ember/-internals/owner/index.ts
```diff
@@ -55,25 +55,26 @@ export const OWNER = symbol('OWNER');
...
 
   // Usage:
   //
-  //   {{play-audio audioType=model.audioType audioFile=model.file}}
+  //   <PlayAudio @audioType={{model.audioType}} @audioFile={{model.file}}/>
   //
-  export default Component.extend({
-    audioService: computed('audioType', function() {
+  export default class extends Component {
+    @computed('audioType')
```
pzuraq: Ah, sorry, one more change: Remove the `@computed`, it shouldn't be necessary with Glimmer components (and it would have to point to `args.audioType` anyways)

esbanarango: oh didn't know that. So cool! 👌

packages/@ember/-internals/owner/index.ts
```diff
@@ -55,25 +55,24 @@ export const OWNER = symbol('OWNER');
...
   ```app/components/play-audio.js
-  import Component from '@ember/component';
-  import { computed } from '@ember/object';
+  import Component from '@glimmer/component';
   import { getOwner } from '@ember/application';
 
   // Usage:
   //
-  //   {{play-audio audioType=model.audioType audioFile=model.file}}
+  //   <PlayAudio @audioType={{model.audioType}} @audioFile={{model.file}}/>
```
locks:
```suggestion
  //   <PlayAudio @audioType={{@model.audioType}} @audioFile={{@model.file}}/>
```

packages/@ember/-internals/owner/index.ts
```diff
@@ -55,25 +55,24 @@ export const OWNER = symbol('OWNER');
...
-    audioService: computed('audioType', function() {
+  export default class extends Component {
+    get audioService() {
       let owner = getOwner(this);
-      return owner.lookup(`service:${this.get('audioType')}`);
-    }),
+      return owner.lookup(`service:${this.args.audioType}`);
+    }
 
     click() {
```
locks: I think the name of this method should be changed so it doesn't give the impression that glimmer components also have DOM event handlers. What do you think of something like `onPlay`?

packages/@ember/-internals/owner/index.ts
```diff
@@ -55,25 +55,24 @@ export const OWNER = symbol('OWNER');
   `audioType` passed as an attribute:
```
locks:
```suggestion
  `audioType` passed as an argument:
```

esbanarango:
Comments addressed @locks @pzuraq.

pzuraq (approved):

pzuraq:
Awesome, thanks for doing this (and for the patience with the back and forth!)

PR #18553, opened 2019-11-15
Ensure errors during rendering withing Application.visit reject visit.

rwjblue:
Paired with @krisselden (and kinda @pzuraq :wink:)

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/renderer.ts
+++ b/packages/@ember/-internals/glimmer/lib/renderer.ts
@@ -1,5 +1,5 @@
 import { ENV } from '@ember/-internals/environment';
-import { runInTransaction } from '@ember/-internals/metal';
+import { notifyPropertyChange, runInTransaction, tagForProperty } from '@ember/-internals/metal';
 import { getViewElement, getViewId } from '@ember/-internals/views';
 import { assert } from '@ember/debug';
 import { backburner, getCurrentRunLoop } from '@ember/runloop';
@@ -220,6 +220,15 @@ function resolveRenderPromise() {
   }
 }
 
+function rejectRenderPromise(error: unknown) {
+  if (renderSettledDeferred !== null) {
+    let reject = renderSettledDeferred.reject;
+    renderSettledDeferred = null;
+
+    backburner.join(null, reject, error);
+  }
+}
+
 let loops = 0;
 function loopEnd() {
   for (let i = 0; i < renderers.length; i++) {
@@ -274,6 +283,8 @@ export abstract class Renderer {
     this._isRenderingRoots = false;
     this._removedRoots = [];
     this._builder = builder;
+
+    tagForProperty(this, '_roots');
   }
 
   // renderer HOOKS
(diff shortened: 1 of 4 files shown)


packages/@ember/-internals/glimmer/lib/renderer.ts
```diff
@@ -510,7 +525,16 @@ export abstract class Renderer {
     if (this._isValid()) {
       return;
     }
-    this._renderRootsTransaction();
+
+    if (renderSettledDeferred !== null) {
```
rwjblue: We only use a `try`/`catch` when we need to manually reject the `renderSettled` promise. This allows us to ensure that the normal developer ergonomics are still good, while supporting the requirement of rejecting the `ApplicationInstance.prototype.visit` promise. 

Specifically, if we always wrapped in `try`/`catch` when you have an error thrown from a userland hook (e.g. component constructor, didInsertElement, etc) the "pause on uncaught exceptions" feature of DevTools will pause deep in the renderer internals. Without the `try`/`catch` the breakpoint you hit will actually be in userland code at the location of the error.

packages/@ember/application/instance.js
```diff
@@ -246,22 +247,14 @@ const ApplicationInstance = EngineInstance.extend({
...
-        return renderSettled().then(() => this);
-      }
-    };
+    let shouldRender = bootOptions.options.shouldRender;
 
     let handleTransitionReject = error => {
       if (error.error) {
         throw error.error;
       } else if (error.name === 'TransitionAborted' && router._routerMicrolib.activeTransition) {
         return router._routerMicrolib.activeTransition.then(
```
krisselden: catch

packages/@ember/-internals/glimmer/lib/renderer.ts
```diff
@@ -1,5 +1,5 @@
 import { ENV } from '@ember/-internals/environment';
-import { runInTransaction } from '@ember/-internals/metal';
+import { runInTransaction, notifyPropertyChange, tagForProperty } from '@ember/-internals/metal';
```
rwjblue: Linting failure:

```
ERROR: /home/travis/build/emberjs/ember.js/packages/@ember/-internals/glimmer/lib/renderer.ts:2:10 - Named imports must be alphabetized.
```

packages/@ember/application/instance.js
```diff
@@ -276,10 +266,12 @@ const ApplicationInstance = EngineInstance.extend({
...
 
-    // getURL returns the set url with the rootURL stripped off
-    return router
-      .handleURL(location.getURL())
-      .then(handleTransitionResolve, handleTransitionReject);
+    return Promise.all([
+      shouldRender && renderSettled(),
+
+      // getURL returns the set url with the rootURL stripped off
+      router.handleURL(location.getURL()).then(null, handleTransitionReject),
```
krisselden: .catch(handleTransitionReject)  instead

packages/@ember/-internals/glimmer/lib/renderer.ts
dnalagatla: Thank you @rwjblue

kategengler:
Is this still relevant?

kategengler:
Closing since this is unlikely to be picked up in this form

PR #18557, opened 2019-11-16
[BUGFIX release] Ensure production tests actually run

pzuraq:
Production tests were not actually running in some of the test suites, because we had a `pretest` command that would build the development version immediately after we built the production version.


rwjblue:
@pzuraq - I see you prefixed with `BUGFIX lts`, but its not clear if this actually is broken back to 3.12 can you confirm?

PR #18570, opened 2019-11-23
[BREAKING BUGFIX release] Fix backtracking rerender assertion

pzuraq:
The old backtracking rerender assertion required us to call assertNotRendered
on every property set. This was normally done in notifyPropertyChange, but
tracked properties bypassed that and went directly to markObjectAsDirty, which
meant they didn't ever check themselves when updating.


rwjblue:
Just to confirm, this is targeting release only because the newer assertion infrastructure replaces assertNotRendered, right? 

Can you PR this test against master (helps confirm the new assertion also catches this)?

pzuraq:
Yes, and sure thing!

rwjblue:
I’m a little concerned with doing a patch release that adds additional assertions, but it may be worth it (because users really need to avoid these sorts of invalid mutation scenarios).

Thoughts on the pros/cons there?

Also, can you update the commit message and PR title to be `[BREAKING BUGFIX release]` (since this is certainly both a breaking change **and** a bug fix

pzuraq:
### Pros

* Users who update to the latest patch release of 3.14, now and in the future, will be more prepared for the changes to the backtracking assertion in 3.15.
* It could catch some subtle bugs that users don't realize they are experiencing today.

### Cons

* Users of 3.13 will still be affected, so this doesn't protect all current/future users.
* Users could experience sudden breakage on a patch release, which would not be a great experience.
* Could cause more feelings of churn overall - if a user has to make changes once for 3.14, and then again for 3.15, it may feel worse than having to deal with them all at once.

### Alternatives

* Leave the state of the code as is, and let 3.15 rollout with the updated transaction. It's much more thorough, and will definitely prevent this misusage (as the test shows). Also batches all potential changes together, so it may feel like a lot, but it'll all be at once.
* Also backport to 3.13. It's against policy, but we could make an exception so that any future users are covered, and can't accidentally fall into these invalid uses.

Issue #16349, opened 2018-03-09
RouterService forces model to refresh when queryParams unchanged


ro0gr:
I have a query param defined in the controller and set it up to `refreshModel` in the route.

Then I inject the `router` service to the component and trigger a transition just with a sub-route name(no passing query params).

**expected**: since I haven't passed [`Ember.DEFAULT_VALUE`](https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md#query-parameter-semantics) query params should be preserved during the transition.

**actual**:  query param treated as changed which causes the parent route to reload.

Reproduction: https://ember-twiddle.com/efda3c9e57835c987404a8f5968c7579?openFiles=routes.application.js%2C&route=%2Fin

ro0gr:
@ef4 am I right by considering this behavior as a bug?

ro0gr:
The other interesting thing which appears in the twiddle is that default `queryParams` are reflected in the URL only after the second `router.transitionTo()` invocation.

[Router RFC](https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md#query-parameter-semantics) sheds some light on the default query params semantics. But it's unclear how the  `transtitionTo` should behave when no query params are passed in.

ro0gr:
This issue is very similar to https://github.com/emberjs/ember.js/issues/15801. It may have the same root cause. However this issue is reproducible without passing any query params to the `router.transtionTo()`

Exelord:
I've got the similar issue right now. And I found a problem. If your QP default is a `null` it's always visible as a change for the router which triggers the model reloading. Try to change it to `undefined` or other, then it should not reload the model.

ro0gr:
@Exelord actually in the twiddle I have a non-empty string as a default QP value.

rwjblue:
I believe that this is addressed now, @chadhietala may be able to confirm.

ro0gr:
I've just verified it against ember-source@3.6(cause its release notes say: `Final stage of the router service RFC (see emberjs/rfcs#95`) and the issue seems to be persited.

Here is a [repo with reproduction](https://github.com/ro0gr/qparams-repro) for ember-source@3.6 if helpful.

arun-jerry:
is there a work around for this? other than reloading the window.

aaxelb:
[...]

PR #18579, opened 2019-11-25
Add failing tests for RouterService forcing unnecessary refresh (#16349)

aaxelb:
a bug exists (#16349):
1. add a query param with default value and `refreshModel: true`
2. using the router service, transition between child routes
3. observe the parent model hooks re-running every time

This PR just adds failing tests for the router service's `transitionTo` and `replaceWith`.


nickschot:
> Is this still relevant?

I'd say yes, this likely relates to [#router](https://github.com/emberjs/ember.js/issues/19497) and the others mentioned in that thread.

PR #18580, opened 2019-11-26
remove peekmeta in property_set

bekzod:
before meta was used multiple paces in `set` so made sense to have it to avoid calling peekMeta internally


rwjblue (approved):
Seems good to me, thanks!

PR #18581, opened 2019-11-26
Fix `bin/build-for-publishing.js` when `auto-dist-tag` is not globally.

rwjblue:
`execa@1` defaulted `preferLocal` to `true`, but 2.x changed the default. This updates the script to specify the option regardless.


kategengler:
Closing as publishing is working and this is old :)

PR #18584, opened 2019-11-27
[BUGFIX lts] Backport dynamic Ember compilation from 3.13

timiyay:
This PR backports the recent changes to the Ember compilation pipeline in Ember 3.13, which dynamically compiles the Ember source based on a dependent app's browser targets.

The list of PRs to backport was supplied by @pzuraq.


timiyay:
The last commit timed out in CI.

Closing and reopening this PR, as I was told that's one way to trigger a rebuild of CI.

timiyay:
@pzuraq backport commits are done.

I squashed each PR into a single commit. Other than that, it was bog-standard cherry picks all the way.

I'm going to continue some smoke testing against our apps, and also figure out whether the CI failures are genuine, or just Browserstack timeouts.

NullVoxPopuli:
What are the advantages of these features? I haven't heard of them before now

pzuraq:
This isn't a feature, so much as a massive bugfix. Before this change, we were precompiling, bundling, and shipping Ember. Due to native class interop issues, we had to try to match the target browsers of the consumer. We tried by shipping 2 builds simultaneously, one for modern browsers, one for legacy browsers, but it resulted in a large number of bugs.

This change compiles Ember dynamically, like any other Ember addon, using `ember-cli-babel`. That way, it compiles with the same targets as the end user.

timiyay:
FWIW @kanongil's recent PR https://github.com/emberjs/ember.js/pull/18576 fixes the most urgent and impactful regression - insta-crashing caused by leftover `const` in legacy builds. That PR will fix our company's immediate problems.

These dynamic build features are definitely an improvement but, if landing this backport proves challenging, we'd be perfectly happy use https://github.com/emberjs/ember.js/pull/18576, and wait until our next LTS bump to consume these dynamic builds.

pzuraq:
There are two reasons why I want to land this:

1. That is probably not the last bug that's waiting to surface. We've had this happen a few times, with a different set of browsers triggering a regression. 

2. Libraries like Ember Data can't upgrade to use native classes until a few LTS releases have the ability to avoid bugs caused by the legacy builds.

You got in 90% of the way @timiyay, I haven't had a chance to review just yet but it looks good, and I'll figure out what's going on with the IE11 tests 😄

kanongil:
This backport seem outside the scope of the kind of "critical bug fixes" I would expect in an LTS. As such, it should not be merged.

pzuraq:
This is very much in the scope of LTS bugfixes, as it solves a number of longstanding, critical issues.

kanongil:
If you merge a "bugfix" of this scope, which could introduce entirely new bugs, I hope you are prepared to extend bugfix support for the LTS release into the "security fixes only" phase.

pzuraq:
The code introduced in the PR has been used on master for some time and has been tested pretty thoroughly because of this. At this point, we’ve caught most of the issues with the initial implementation. This is why we didn’t merge this PR before 3.12 was released, because we did expect there to be issues, but I’m pretty confident we’ve caught them at this point.

timiyay:
Closing due to inactivity, though I'll keep the fork and branch around in case we need to revive it.

The urgent bug was fixed in https://github.com/emberjs/ember.js/pull/18576, so the backport is less of a priority now methinks.

PR #18594, opened 2019-12-06
Updating the examples for ObjectProxy documentation

millerm:
The examples used in the  ProxyObject documentation could be improved. I have adjusted the examples to use `const` when declaring variables. I have also changed the names of the variables from `object` to `exampleObject` and `proxy` to `exampleProxy`. My rationale for this is that `object` could be confused with the `Object` data type.


rwjblue:
Would you mind rebasing this against `master` and updating the PR target branch?

PR #18597, opened 2019-12-06
remove _meta arg from tagForProperty and tagFor

bekzod:
`_meta` args are not being used anymore, make sense to remove it to avoid extra checks


pzuraq:
I'm currently in the middle of the VM upgrade and rewrite these methods a decent amount, so ideally would like to hold off to get that merged. This is a good fix though, and thank you for submitting it!

bekzod:
made redundant by glimmer update :P so closing

PR #18599, opened 2019-12-07
use `bb.ensureInstance` instead of `bb.join` in renderer.ts

bekzod:


rwjblue (approved):

rwjblue:
I haven't dug in, but it looks like the test failures here are related to your changes?

packages/@ember/-internals/glimmer/lib/renderer.ts
```diff
@@ -87,7 +86,6 @@ class RootState {
     this.env = env;
     this.root = root;
     this.result = undefined;
-    this.shouldReflush = false;
```
bekzod: leftover from the past, so removed it

packages/@ember/-internals/glimmer/lib/renderer.ts
```diff
@@ -68,7 +68,6 @@ class RootState {
   public env: Environment;
   public root: Component | OutletView;
   public result: RenderResult | undefined;
-  public shouldReflush: boolean;
```
bekzod: leftover from the past, so removed it

packages/@ember/-internals/glimmer/tests/integration/syntax/each-test.js
```diff
@@ -555,7 +555,7 @@ class EachTest extends AbstractEachTest {
 
     runTask(() => set(this.objectAt(0), 'value', 3));
 
-    this.assertText('PrevNextPrev2NextPrev3Next');
```
bekzod: surprisingly this didn't assert before

bekzod:
after checking test failures, I had to add `renderSettled` few places, probably because there is no extra run loop spawn there is not enough time to settle for those case, haven't had time to investigate too deep on that

locks:
@bekzod you have some conflicts!

Issue #18641, opened 2019-12-28
Async methods are treated as falsy in template conditionals


DuBistKomisch:
```
export default class Whatever extends Component {
  doSomething() {
    ...
  }

  async doSomethingAsync() {
    ...
  }
}
```

```
{{if this.doSomething 'true' 'false'}}
{{if this.doSomethingAsync 'true' 'false'}}
```

This renders `true false`, but should render `true true`. Happens in latest Firefox and Chromium on Linux. I'm not using the babel polyfill AFAIK, and I'm on latest stable Ember (3.15.0) but I've seen this on previous versions without realising what was happening.

This is particularly problematic when passing optional `@action`s into other components. Workaround is to wrap the async method in a sync method and use that instead.

pzuraq:
This does seem like a bug, I would have expected an `async` method to still be truthy. We should look into this asap.

rwjblue:
Fixed by #18651

PR #18651, opened 2020-01-03
[BUGFIX beta] Async functions being treated as falsey in template conditionals

mjanjic01:
Fixes https://github.com/emberjs/ember.js/issues/18641

The [typeOf TYPE_MAP](https://github.com/emberjs/ember.js/blob/56d60eac53/packages/%40ember/-internals/runtime/lib/type-of.js#L6) was missing an `[object AsyncFunction]` entry, which results in the typeOf helper returning `object` as the type for a given async function.

This will propagate the error to the [isArray method conditional](https://github.com/emberjs/ember.js/blob/56d60eac53/packages/@ember/-internals/runtime/lib/mixins/array.js#L168) and the [toBool util](https://github.com/emberjs/ember.js/blob/56d60eac537933e20ef66e735299b81dba9a1788/packages/@ember/-internals/glimmer/lib/utils/to-bool.ts#L4) ultimately used by the [inlineIf helper](https://github.com/emberjs/ember.js/blob/v3.4.8/packages/ember-glimmer/lib/helpers/if-unless.ts#L132) resulting in the bug.

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/utils/shared-conditional-tests.js
+++ b/packages/@ember/-internals/glimmer/tests/utils/shared-conditional-tests.js
@@ -402,6 +402,7 @@ const IfUnlessWithTestCases = [
     ObjectProxy.create({ content: true }),
     Object,
     function() {},
+    async function() {},
     new String('hello'),
     new String(''),
     new Boolean(true),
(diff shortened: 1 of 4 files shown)


rwjblue (approved):

rwjblue:
Thank you!!

PR #18668, opened 2020-01-09
[BUGFIX release] Consumes tags after fully getting values

pzuraq:
Currently get() consumes some tags before getting the values,
and getting the values itself can cause tag changes and notifications.
In some cases, this is invalid, because the value that is being notified
hasn't actually been consumed yet (it is in the process of being
consumed. More generally, consumption should always occur immediately
_after_ the relevant calculation - in this case, the getter.

### DIFF
--- a/packages/@ember/-internals/metal/tests/tracked/validation_test.js
+++ b/packages/@ember/-internals/metal/tests/tracked/validation_test.js
@@ -402,5 +402,25 @@ moduleFor(
         });
       }, /You attempted to update `foo` on `EmberObject`, but it had already been used previously in the same computation/);
     }
+
+    ['@test get() does not entangle in the autotracking stack until after retrieving the value'](
+      assert
+    ) {
+      assert.expect(0);
+
+      class EmberObject {
+        get foo() {
+          notifyPropertyChange(this, 'foo');
+
+          return 123;
+        }
+      }
+
+      let obj = new EmberObject();
+
+      track(() => {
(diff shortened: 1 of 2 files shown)


lifeart:
looks like https://github.com/machty/ember-concurrency/issues/337 may be related

makepanic:
Thanks, this seems to solve https://github.com/emberjs/data/issues/6943 .

Is there a plan on when this will be merged?

packages/@ember/-internals/metal/tests/tracked/validation_test.js
```diff
@@ -402,5 +402,25 @@ moduleFor(
         });
       }, /You attempted to update `foo` on `EmberObject`, but it had already been used previously in the same computation/);
     }
+
+    ['@test get() does not entangle in the autotracking stack until after retrieving the value'](
```
rwjblue: Can you explain this test? How did it fail before? How does it confirm what it says it is confirming?

pzuraq: Before, `get` would consume the tag for the property immediately, then actually get the _value_ of the property. If the property signaled a change via `notifyPropertyChange` in this time, it would throw the backtracking assertion. This is the fundamental problem that led to the bug in Ember Data.

After, we consume all tags for the first time _after_ getting resolved values, whatever they end up being, which means it is safe to signal property changes while calculating the value.

PR #18682, opened 2020-01-14
Add docs about when tracked is not needed

RobbieTheWagner:
@pzuraq @jenweber this is just a start, but I think we should clearly articulate when tracked is needed and when it is not. Please let me know your thoughts and what additional cases we should add.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/glimmer-tracking-docs.ts
+++ b/packages/@ember/-internals/glimmer/lib/glimmer-tracking-docs.ts
@@ -97,6 +97,13 @@
     }
   }
   ```
+  
+  ### When `@tracked` is not needed
+  
+  * The property is already a `computed`
+  * ember-data attrs (which are computeds behind the scenes)
+  * Using Ember `get` / `set` will auto track
+  
 
   @method tracked
   @static


packages/@ember/-internals/glimmer/lib/glimmer-tracking-docs.ts
```diff
@@ -97,6 +97,13 @@
     }
   }
   ```
+  
+  ### When `@tracked` is not needed
+  
+  * The property is already a `computed`
+  * ember-data attrs (which are computeds behind the scenes)
+  * Using Ember `get` / `set` will auto track
```
btecu: What about passed in properties (`@something` / `this.args.something`)?

rwwagner90: Are those auto tracked? I don't actually know.

pzuraq:
I don’t think this level of knowledge belongs in the API docs. We should more clearly explain the general case here, which is that if something desugars down to accessing a tracked property, `@tracked` is not needed.

The mental model for Ember Data attributes should be: `@attr` adds a native getter that does some data related things, one of which is getting/setting a tracked property/storage cell. This is more or less what it is, effectively.

The specific use cases here should be covered in:

* Ember Data api docs for `@attr`, `@belongsTo`, `@hasMany`
* Ember Data guides
* Octane upgrade guides
* Octane cheatsheet

rwwagner90:
@pzuraq I think we could benefit from a central location that lists where tracked is needed and where it is not. As a seasoned Ember dev, I am unlikely to go look at the docs on `@attr` for example, since I've been using Ember data's `attr` forever.

packages/@ember/-internals/glimmer/lib/glimmer-tracking-docs.ts
pzuraq: Those are autotracked, yes.

pzuraq:
@rwwagner90 the issue is it's not going to be possible to list all of those locations, because we don't actually limit them in the framework. For instance, you could create your own data library, or a library like `ember-apollo` that wraps another library, and your own API that autotracks correctly without your users having to use `@tracked` anywhere. In the end, what is autotracked and what isn't needs to be documented by the libraries themselves.

Personally I think a non-exhaustive list that _appears_ to be exhaustive would be worse, and that's what I'd like to avoid.

rwwagner90:
> Personally I think a non-exhaustive list that appears to be exhaustive would be worse, and that's what I'd like to avoid.

That is totally fair, but perhaps we could not say the list is exhaustive, but just tips like that `attr` is really computed and tracked by default

pzuraq:
I'd be open to that, but I'd also like to explain the general principle in the same place and refer to these as examples of that general principle.

rwwagner90:
> I'd be open to that, but I'd also like to explain the general principle in the same place and refer to these as examples of that general principle.

That sounds perfect to me!

packages/@ember/-internals/glimmer/lib/glimmer-tracking-docs.ts
rjschie: In [here](https://guides.emberjs.com/release/upgrading/current-edition/glimmer-components/#toc_arguments) it describes:

>`args` and its values are automatically tracked, so there is no need to annotate them, the `fullName` getter will invalidate properly when they change and the component will rerender (if `fullName` is used in the template).

rwjblue:
https://guides.emberjs.com/release/in-depth-topics/autotracking-in-depth/ seems like the right place for the kinds of in-depth explanation that @pzuraq is referring to (and already contains at least _some_ of this info), so I'm going to close this in favor of tweaking that guide...

PR #18691, opened 2020-01-19
[BUGFIX release] Updated blueprints for component and helper tests to output the correct hbs import statement

### LINKED ISSUE #18626: Ember 3.15 - Helper test always imports hbs from htmlbars-inline-precompile
## Description

The blueprint for a helper test hardcoded the `hbs` import statement.

```javascript
```

https://github.com/emberjs/ember.js/blob/master/blueprints/helper-test/qunit-rfc-232-files/__root__/__testType__/__collection__/__name__-test.js#L4

For Ember 3.15, which uses `ember-cli-htmlbars ^4.2.0`, I believe the blueprint would need to generate the line:

```javascript
import { hbs } from 'ember-cli-htmlbars';
```

https://github.com/ember-cli/ember-cli-htmlbars#tagged-template-usage--migrating-from-htmlbars-inline-precompile

## Proposed Solution

The blueprint for a component test shows how we can check the version of `ember-cli-htmlbars` in order to dynamically add the import statement:

https://github.com/emberjs/ember.js/blob/master/blueprints/component-test/index.js#L89-L97

https://github.com/emberjs/ember.js/blob/master/blueprints/component-test/qunit-rfc-232-files/__root__/__testType__/__path__/__test__.js#L4

Please check both QUnit and Mocha.

ijlee2:
## Description

This PR closes #18626.

## Tests

I wanted to write tests per contributing guide, but couldn't figure out how to make `this.project.addons` include `ember-cli-htmlbars` in a test. Currently, there isn't a test that checks the component test when the version of `ember-cli-htmlbars` is at least `4.0.0-alpha.1`, so I didn't have an example that I could follow.

It seemed like `ember-cli-blueprint-test-helpers` currently doesn't support updating the `addons` attribute (https://github.com/ember-cli/ember-cli-blueprint-test-helpers/blob/master/lib/modify-packages.js#L20-L26). If I missed something and setting `addons` is possible, please let me know and I'd be happy to add tests.

I did manually test locally using `yarn link ember-source` on a new Ember 3.15 project. If the version of `ember-cli-htmlbars` is at least `4.0.0-alpha.1`, the generators will add the named import statement. If the version is less, or if `ember-cli-htmlbars` doesn't exist, the generators will add the default import statement.

### DIFF
--- a/blueprints/component-test/index.js
+++ b/blueprints/component-test/index.js
@@ -62,9 +62,9 @@ module.exports = useTestFrameworkDetector({
       componentPathName = [options.path, dasherizedModuleName].filter(Boolean).join('/');
     }
 
-    let hbsImportStatement = this._useSeparateInlinePrecompileAddon()
-      ? "import hbs from 'htmlbars-inline-precompile';"
-      : "import { hbs } from 'ember-cli-htmlbars';";
+    let hbsImportStatement = this._useNamedHbsImport()
+      ? "import { hbs } from 'ember-cli-htmlbars';"
+      : "import hbs from 'htmlbars-inline-precompile';";
 
     let templateInvocation = invocationFor(options);
     let componentName = templateInvocation;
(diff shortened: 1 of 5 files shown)


blueprints/component-test/index.js
```diff
@@ -86,21 +86,21 @@ module.exports = useTestFrameworkDetector({
...
   },
 
-  _useSeparateInlinePrecompileAddon() {
+  _useNamedHbsImport() {
     let htmlbarsAddon = this.project.addons.find(a => a.name === 'ember-cli-htmlbars');
 
-    if (htmlbarsAddon === undefined) {
+    if (htmlbarsAddon && semver.gte(htmlbarsAddon.pkg.version, '4.0.0-alpha.1')) {
       return true;
-    } else if (semver.gte(htmlbarsAddon.pkg.version, '4.0.0-alpha.1')) {
```
ijlee2: This line implies that, if the version is `ember-cli-htmlbars` is 3.x or less, we will still use the named import (because we are returning `undefined`). Could you verify if this was intended or a bug?

rwjblue: 3.x should use `import hbs from 'htmlbars-inline-precompile'`, which AFAICT is still what is done after this change.

rwjblue (approved):

PR #18697, opened 2020-01-24
[BUGFIX beta] Adds a canary feature flag for named blocks

pzuraq:
Named blocks do not appear to be working just yet, but this ensures that
whenever the work does land for them they won't be enabled without us
explicitly deciding to enable them.


packages/ember-template-compiler/lib/plugins/assert-against-named-blocks.ts
```diff
@@ -0,0 +1,36 @@
...
+        }
+      },
+    },
+  };
+}
+
+function assertMessage(tagName: string, moduleName: string, node: AST.ElementNode) {
+  let sourceInformation = calculateLocationDisplay(moduleName, node.loc);
+
+  return `Named blocks are not currently available, attempted to use the {{${tagName}}} named block. ${sourceInformation}`;
```
chancancode: This is *passing* the named block right? And shouldn't the this be `<${tagName}>`?

We should also block `{{yield to=...}}` where `...` is not either `"default"` or `"inverse"` (the literals)

rwjblue (approved):

PR #18699, opened 2020-01-25
[CLEANUP] Remove unused canary flags

pzuraq:
Removes the existing unused canary feature flags. Each commit in this PR removes one flag, so they can be reviewed individually.

The largest changes come last, in removing `EMBER_MODULE_UNIFICATION`. I fully removed the `source` and `namespace` options from the injection APIs, along with `expandLocalLookup` from the registry/resolver APIs. As far as I can tell, these APIs were never actually documented or released, and were used exclusively by a few experimental addons that were trying to jump the gun on MU. Removing them really helps to clean things up, so I think it would be ideal.


locks:
Rebased.

PR #18707, opened 2020-01-28
Merge pull request #18707 from emberjs/bugfix/fix-object-proxy-tags

### LINKED ISSUE #18704: [3.13+] ObjectProxy not rerendering
Original report:

We're experienceing an almost identical issue when bumping an app from 3.12 to 3.13.

We have a utility function which returns a CP which returns an ObjectProxy which proxies the object passed to the utility (e.g. myUtil('model')). We've been using this primarily to store updates to form values locally.
```js
let myUtil = function (contentKey) {
  return computed(
    contentKey,
    function () {
      return MyProxyObject.create({ content: this.get(contentKey) });
    }
  );
}
```

We have a unit test (this is very abbreviated): 
```js
let foo = { bar: false };
let myController = Controller.extend({
  wrappedFoo: myUtil('foo'),
  computedBar: computed(
    'wrappedFoo.bar',
    function () { return this.get('wrappedFoo.bar'); }
  ),
  booledBar: bool('wrappedFoo.bar')
}).create({ foo });

assert.notOk(myController.get('aliasedBar'));
myController.set('wrappedFoo.bar', true);
assert.ok(myController.get('computedBar')); // was passing, now failing
assert.ok(myController.get('booledBar')); // was passing, now failing
```

cc @theroncross, [Original context](https://github.com/emberjs/ember.js/issues/18689#issuecomment-577294423)

I'm currently unable to reproduce this issue, with the following test case added to Ember directly:

```js
    '@test it should work when returned from a computed property'(assert) {
      function myUtil(contentKey) {
        return computed(contentKey, function() {
          return ObjectProxy.create({ content: this.get(contentKey) });
        });
      }

      let foo = { bar: false };
      let myController = EmberObject.extend({
        wrappedFoo: myUtil('foo'),
        computedBar: computed('wrappedFoo.bar', function() {
          return this.get('wrappedFoo.bar');
        }),
        booledBar: bool('wrappedFoo.bar'),
      }).create({ foo });

      assert.notOk(myController.get('aliasedBar'));
      myController.set('wrappedFoo.bar', true);
      assert.ok(myController.get('computedBar'));
      assert.ok(myController.get('booledBar'));
    }
```

pzuraq:
Currently, the Proxy mixin uses the private `UNKNOWN_PROPERTY_TAG` API
to provide the tags for the content property that the mixin is proxying
to. However, previously the chains system would do this _and_ still
entangle changes to the local property. This allowed users to manually
notify in subclasses of Proxy.

This PR adds the tag for the property back to the returned tag, so it
works the same as before. It also refactors `setupMandatorySetter`,
since we now have to do that work in two places (to avoid re-entry).

Fixes #18704

### DIFF
--- a/packages/@ember/-internals/runtime/lib/mixins/-proxy.js
+++ b/packages/@ember/-internals/runtime/lib/mixins/-proxy.js
@@ -13,8 +13,9 @@ import {
   CUSTOM_TAG_FOR,
   getChainTagsForKey,
 } from '@ember/-internals/metal';
-import { setProxy } from '@ember/-internals/utils';
+import { setProxy, setupMandatorySetter } from '@ember/-internals/utils';
 import { assert } from '@ember/debug';
+import { DEBUG } from '@glimmer/env';
 import { combine, update, tagFor } from '@glimmer/validator';
 
 export function contentFor(proxy) {
@@ -58,10 +59,21 @@ export default Mixin.create({
   }),
 
   [CUSTOM_TAG_FOR](key) {
+    let tag = tagFor(this, key);
+
+    if (DEBUG) {
+      // TODO: Replace this with something more first class for tracking tags in DEBUG
+      tag._propertyKey = key;
+    }
+
     if (key in this) {
-      return tagFor(this, key);
+      if (DEBUG) {
+        setupMandatorySetter(tag, this, key);
+      }
+
+      return tag;
     } else {
-      return combine(getChainTagsForKey(this, `content.${key}`));
+      return combine([tag, ...getChainTagsForKey(this, `content.${key}`)]);
     }
   },
 
--- a/packages/@ember/-internals/metal/lib/tags.ts
+++ b/packages/@ember/-internals/metal/lib/tags.ts
@@ -7,7 +7,6 @@ import {
   toString,
 } from '@ember/-internals/utils';
 import { assert, deprecate } from '@ember/debug';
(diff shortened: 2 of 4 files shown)


packages/@ember/-internals/runtime/lib/mixins/-proxy.js
```diff
@@ -58,7 +59,16 @@ export default Mixin.create({
   }),
 
   [UNKNOWN_PROPERTY_TAG](key) {
-    return combine(getChainTagsForKey(this, `content.${key}`));
+    let tag = tagFor(this, key);
+
+    if (DEBUG) {
+      setupMandatorySetter(tag, this, key);
```
rwjblue: Doesn't this mean that we automatically define each of these keys on the proxy instance? E.g. `Object.keys` output would be changed.

That seems a bit wrong, no?

rwjblue (changes requested):

packages/@ember/-internals/runtime/lib/mixins/-proxy.js
pzuraq: It does, my mistake there. We should likely have a test for this.

rwjblue (approved):

theroncross:
Any chance this will make it back to 3.13?

pzuraq:
This will be backported to 3.16, since it is an LTS. We do not support 3.13 anymore.

PR #18709, opened 2020-01-29
[BUGFIX beta] Fix `this` in `@tracked` initializer

### LINKED ISSUE #18690: @tracked initializer does not have access to `this`
This started failing on canary since the Glimmer VM upgrade. The problem appears to be this line:

https://github.com/glimmerjs/glimmer-vm/blob/master/packages/@glimmer/validator/lib/tracking.ts#L118

This should probably be `initializer.call(self)`, or maybe the whoever made the initializer should have made it an arrow function closure. Either way, the result is that if you have something like `@tracked foo = this.bar`, it doesn't work anymore.

chancancode:
Fixes #18690


packages/@ember/-internals/metal/tests/tracked/validation_test.js
```diff
@@ -44,6 +44,43 @@ moduleFor(
...
+      let snapshot = value(tag);
+
+      assert.equal(obj.first, 'first: second', 'The value initializes correctly');
+      assert.equal(validate(tag, snapshot), true);
+
+      snapshot = value(tag);
+      assert.equal(validate(tag, snapshot), true);
+
+      // TODO: https://github.com/glimmerjs/glimmer-vm/pull/1018
+      // obj.second = '2nd';
```
chancancode: https://github.com/glimmerjs/glimmer-vm/pull/1018

pzuraq (approved):
Looks good, thanks for getting this done!

PR #18586, opened 2019-11-27
add trailing ' to setEach documentation

betocantu93:
Found typo in setEach documentation, missing the trailing '

PR #18718, opened 2020-01-31
add trailing ' to setEach

betocantu93:
Remake of #18586 

Typo, add trailing '


locks (approved):
Thanks for spotting and correcting the error!

PR #18721, opened 2020-02-01
[BUGFIX release] Backport autotracking bugfixes

pzuraq:


packages/@ember/-internals/utils/lib/mandatory-setter.ts
```diff
@@ -1,8 +1,12 @@
 import { assert } from '@ember/debug';
+import { _WeakSet as WeakSet } from '@ember/polyfills';
 import { DEBUG } from '@glimmer/env';
+import { Tag } from '@glimmer/validator';
```
rwjblue: I don't think `@glimmer/validator` exists in this branch (this fails the linting)

rwjblue (changes requested):

packages/@ember/-internals/utils/lib/mandatory-setter.ts
pzuraq: Ah, yeah, strange that tests are passing (I ran them locally)

rwjblue: It’s because it’s only used for types, not runtime

rwjblue: Rebased and fixed the import, verified linting passes...

rwjblue (approved):

Issue #18696, opened 2020-01-24
[3.17.0-beta] Destroying -in-element: `Failed to execute 'removeChild' on 'Node'`


simonihmig:
`Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.`

This error is thrown since `3.17.0-beta` (IIRC also in canary versions before), apparently when destroying an `-in-element`. I did not dig into this, but all tests of `ember-bootstrap` that involve `-in-element` (modals, dropdowns, tooltips, popovers) are [failing with that error](https://travis-ci.org/kaliber5/ember-bootstrap/jobs/636799791#L960). 

To confirm I just triggered a rebuild of `ember-popper` - which also uses `-in-element` - and the same [failing tests there](https://travis-ci.org/kybishop/ember-popper/jobs/592763135).

*Btw, both addons use `ember-maybe-in-element`, but I believe this should not matter as it is a build-time AST transform.*

Likely related to #18621

/cc @pzuraq @cibernox

pzuraq:
There have been changes to the `{{-in-element}}` helper in preparation for making it a public API, so it could be that the usage of it needs to change

simonihmig:
@pzuraq I have created a reproduction here: https://github.com/simonihmig/in-element-reproduction

This fails even when taking the new semantics (`insertBefore`) into account, so I don't think this particular error is related to wrong usage!

To reproduce it seems these preconditions need to come together:
* using a rendering test. I was not able to reproduce in normal app code, neither was I able to create a failing test in the Ember codebase. Maybe it is somehow related to the way the test infrastructure does the teardown of DOM? 🤔
* the destination element is part of the test element (`#ember-testing`). It did work when rendering to a detached DOM element.
* and - and this is the most interesting part - the template needs to generate some *empty text nodes*. The error thrown seems to be about removing those text nodes (after the test finished!).

The latter point is easy to see in the failing test [here](https://github.com/simonihmig/in-element-reproduction/blob/master/tests/integration/components/in-element-test.js#L19-L29), as the (passing) first test is identical except for the white space! 

*Note: both tests pass for Ember 3.16.0 (when removing the unsupported `insertBefore=null`)*

chancancode:
I ran into this issue (Skylight tests currently failing on beta) with a similar setup.

I think the problem is that we are using `-in-element` to render into an element that was torn down before the `-in-element` block itself got torn down.

I think it boils down to something like this:

```hbs
{{!-- application.hbs --}}
[...]

PR #18727, opened 2020-02-04
[BUGFIX beta] Avoid breaking {{-in-element}} usage

chancancode:
The intimate `{{#-in-element}}` API previously defaults to appending to the remote element without clearing. With the VM upgrade, the default has changed to clearing the remote element as specified by the RFC. Since we are not ready to promote the public API yet, we should preserve the existing behavior and let addon authors migrate to the new public API directly when it is ready.

### DIFF
--- a/packages/ember-template-compiler/lib/plugins/transform-in-element.ts
+++ b/packages/ember-template-compiler/lib/plugins/transform-in-element.ts
@@ -66,18 +66,18 @@ export default function transformInElement(env: ASTPluginEnvironment): ASTPlugin
 
           // replicate special hash arguments added here:
           // https://github.com/glimmerjs/glimmer-vm/blob/ba9b37d44b85fa1385eeeea71910ff5798198c8e/packages/%40glimmer/syntax/lib/parser/handlebars-node-visitors.ts#L340-L363
-          let hasInsertBefore = false;
+          let needsInsertBefore = true;
           let hash = node.hash;
           hash.pairs.forEach(pair => {
             if (pair.key === 'insertBefore') {
               assert(
-                `Can only pass a null literal to insertBefore in -in-element, received: ${JSON.stringify(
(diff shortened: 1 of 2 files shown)


packages/ember-template-compiler/lib/plugins/transform-in-element.ts
```diff
@@ -66,28 +66,31 @@ export default function transformInElement(env: ASTPluginEnvironment): ASTPlugin
...
           });
 
           let guid = b.literal('StringLiteral', `%cursor:${cursorCount++}%`);
           let guidPair = b.pair('guid', guid);
           hash.pairs.unshift(guidPair);
 
-          if (!hasInsertBefore) {
-            let undefinedLiteral = b.literal('UndefinedLiteral', undefined);
-            let nextSibling = b.pair('insertBefore', undefinedLiteral);
+          debugger;
```
rwjblue: 🔪

chancancode: oops

rwjblue:
Just wanted to clarify, what is the relationship to https://github.com/emberjs/ember.js/issues/18696?

rwjblue (approved):

chancancode:
@rwjblue unrelated; I have a different fix for that on the VM side. `ember-cli-head` uses `-in-element` on the `<head>` element, which is outside the app's render result and thus does not trigger #18696. However, it now clears the `<head>` element whenever it's rendered. Among other breakages, it causes the qunit harness to be immediately destyled 🙈

simonihmig:
This makes sense, thanks @chancancode for fixing that!

FWIW, I was not sure if it was intentional to land the semantic changes here even when not making this public yet, so I tried to fix my polyfill for that change (https://github.com/kaliber5/ember-in-element-polyfill/pull/37), but this seems unnecessary now. Which is good!

On that matter: what is the reason for not making `in-element` public yet?

PR #18729, opened 2020-02-07
[BUGFIX lts] Revert container deprecation

chancancode:
We agreed that it was important to fix the error cases in #18717, but while the work may be unnecessary, there isn't necessarily any "correctness" issues with looking up things during destruction. Using a deprecation to warn about potentially "unnecessary" work is a bit heavy-handed, and at minimum requires some more continued discussions, so reverting that part of the PR for now (while keeping the errors intact).


pzuraq (approved):

packages/@ember/-internals/container/tests/container_test.js
```diff
@@ -736,75 +736,6 @@ moduleFor(
...
-      let instance = container.lookup('service:foo');
-      assert.ok(instance, 'precond lookup successful');
-
-      runTask(() => {
-        container.destroy();
-        container.finalizeDestroy();
-      });
-    }
-
-    '@test instantiating via container.lookup during destruction enqueues destruction'(assert) {
```
rwjblue: This test should not have been removed. It is testing that things looked up during destruction are properly destroyed (aka the “correctness” fix mentioned in the PR description).

PR #18730, opened 2020-02-08
Workaround for #15635 (integer encoding issue)

chancancode:
Based on https://github.com/emberjs/ember.js/pull/18484#issuecomment-566276195

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/resolver.ts
+++ b/packages/@ember/-internals/glimmer/lib/resolver.ts
@@ -23,6 +23,7 @@ import InternalComponentManager, {
 import { TemplateOnlyComponentDefinition } from './component-managers/template-only';
 import { isHelperFactory, isSimpleHelper } from './helper';
 import { default as componentAssertionHelper } from './helpers/-assert-implicit-component-helper-argument';
+import { default as parseIntHelper } from './helpers/-i';
 import { default as inputTypeHelper } from './helpers/-input-type';
 import { default as normalizeClassHelper } from './helpers/-normalize-class';
 import { default as trackArray } from './helpers/-track-array';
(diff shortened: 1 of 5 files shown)


packages/ember-template-compiler/lib/plugins/safe-integers-bugfix.ts
```diff
@@ -0,0 +1,86 @@
...
+}
+
+function requiresWorkaround(node: AST.MustacheStatement): node is MustacheStatement & { path: NumberLiteral };
+function requiresWorkaround(node: AST.NumberLiteral): boolean;
+function requiresWorkaround(node: AST.MustacheStatement | AST.NumberLiteral): boolean {
+  if (node.type === 'MustacheStatement' && node.path.type === 'NumberLiteral') {
+    return requiresWorkaround(node.path);
+  } else if (node.type === 'NumberLiteral') {
+    // TODO: narrow this down?
+    return isInteger(node);
```
chancancode: We can likely narrow this down, pending investigation by @krisselden

packages/ember-template-compiler/lib/plugins/safe-integers-bugfix.ts
```diff
@@ -0,0 +1,86 @@
...
+  } else {
+    return false;
+  }
+}
+
+function isInteger(node: AST.NumberLiteral): boolean {
+  return Number.isInteger(node.value);
```
btecu: `isInteger` is not supported by `IE` although I'm guessing this code is only going to execute in `node.js`.

chancancode: That’s a good point since we support running the template compiler in the browser

rwjblue (approved):

rwjblue (approved):

packages/@ember/-internals/glimmer/lib/resolver.ts
```diff
@@ -250,6 +251,7 @@ const BUILTINS_HELPERS: IBuiltInHelpers = {
   unless: inlineUnless,
   '-hash': hash,
   '-each-in': eachIn,
+  '-i': parseInt,
```
krisselden: not a fan of shadowing globals

krisselden (approved):

Issue #18665, opened 2020-01-07
Router currentURL Not Set in 3.15


jherdman:
In Ember 3.12 the following code would report the current URL of the application:

```
class MyRouter extends Router {
  @service router;

  activate() {
    super.activate(...arguments);
    console.log('here you are:', this.router.currentURL);
  }
}
```

In Ember 3.15 this will always report `null`. This is expected to work per [RouterService#currentURL](http://api.emberjs.com/ember/3.15/classes/RouterService/properties/currentURL?anchor=currentURL). Cursory investigation shows that both `currentRouteName` nor `currentRoute` are set either.

Curiously the `RouterService` reports the correct URL when used within a Component.

Sample application: https://github.com/jherdman/where-art-thou. Note that the console will always log a `null` URL, whereas the component correctly reports the URL.

chriskrycho:
I'd guess, if you test this, that it works at other points in a route's lifecycle – e.g. `willTransition` or `didTransition`, or actions triggered from a controller via `this.send`; have you checked that? This is definitely breakage, but it's also (unfortunately) underspecified in [the RFC](https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md), and therefore there may not be coverage on it in the implementation.

Aside: I can't find any indication that `Router#activate` is public API? Do you mean `Route#activate` instead?

```js
class MyRoute extends Route {
  @service router;

  activate() {
    super.activate(...arguments);
    console.log('here you are:', this.router.currentURL);
  }
}
```

jherdman:
Hi Chris! Thank you for the reply.

For completeness I've added a `didTransition` action too, but it sadly shows the same bug. As far as I can tell, this does not work at all on routes.

> Aside: I can't find any indication that EmberRouter#activate is public API? Do you mean Route#activate instead?

I think there was a bit of misunderstanding, I am indeed testing Route#activate. I've amended the snippet in my issue. The linked app has a fully functioning demo of the broken behaviour.

chriskrycho:
That's pretty weird indeed. 😬 I won't have a chance to poke at this till next week at earliest, but that's apt to hit our app as well as we upgrade, so at least investigating further is reasonably high priority.

locks:
https://github.com/emberjs/ember.js/issues/18585 seems related.

jherdman:
I'm noticing this is affecting test helpers in 3.16 too.

xg-wang:
[...]

PR #18738, opened 2020-02-12
[DOC release] Add note about router service event and property update…

xg-wang:
… timing

Also adding a bunch of hook/event assertions in tests to clearly state the supported timing order.

Related https://github.com/emberjs/ember.js/issues/18665

### DIFF
--- a/packages/@ember/-internals/routing/lib/services/router.ts
+++ b/packages/@ember/-internals/routing/lib/services/router.ts
@@ -439,6 +439,16 @@ export default class RouterService extends Service {
     }
     ```
 
+    `routeDidChange` will be called after any `Route`'s
+    [didTransition](/ember/release/classes/Route/events/didTransition?anchor=didTransition)
+    action has been fired.
+    The updates of properties
+    [currentURL](/ember/release/classes/RouterService/properties/currentURL?anchor=currentURL),
+    [currentRouteName](/ember/release/classes/RouterService/properties/currentURL?anchor=currentRouteName)
+    and
+    [currentRoute](/ember/release/classes/RouterService/properties/currentURL?anchor=currentRoute)
+    are completed at the time `routeDidChange` is called.
+
     @event routeDidChange
     @param {Transition} transition
     @public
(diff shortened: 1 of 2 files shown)


xg-wang:
I'm not entirely sure if it should be documented as public for exact timing order. In fact I would like to deprecate `Route`'s `didTransition` action, then make it public saying `currentURL` etc updates after model hooks and before the final `routeDidChange` event during mid-transition.

packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -439,6 +439,10 @@ export default class RouterService extends Service {
     }
     ```
 
+    The exact timing to fire `routeDidChange` is after the Route's
+    [didTransition](/ember/release/classes/Route/events/didTransition?anchor=didTransition)
+    action fired.
```
rwjblue: I think that it is good to explicitly state this, but I'd prefer to tweak the verbiage here a bit. Maybe something:

```
`routeDidChange` will be called after any `Route`'s [didTransition](/ember/release/classes/Route/events/didTransition?anchor=didTransition) action has been fired.
```

packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -469,6 +473,9 @@ RouterService.reopen(Evented, {
      * `blog.index` when you visit `/blog`
      * `blog.post` when you visit `/blog/some-post-id`
 
+     This property is updated after Route's [didTransition](/ember/release/classes/Route/events/didTransition?anchor=didTransition)
+     action and before `routeDidChange` being fired.
```
rwjblue: FWIW, I would actually prefer to change these to actually be based on `routeDidChange`.

rwjblue (commented):
Overall, I think this is great! Thanks for working on it!

rwjblue:
> in fact I would like to deprecate Route's didTransition action

Not sure I agree here. While I do much prefer to encourage usage of `routeDidChange` over action usage, it is actually pretty annoying to write the code for an individual route to subscribe to `routeDidChange` and determine if it was the target of the transition or if it was above or below the "pivot point" ([this code](https://github.com/nickiaconis/ember-prefetch/blob/master/addon/-private/diff-route-info.js) is the kind of thing I am referring to here).  Until we have a good answer here, I don't feel good about deprecating the actions...

xg-wang:
Forgot to mention when I said I would like to deprecate `didTransition`, a similar `routeDidChange` which bubbles similar to the action event needs be added.

But I think you are right at a better solution may be to have something to  determine if it was the target of the transition or if it was above or below the "pivot point".

rwjblue (approved):
Thank you!

PR #18741, opened 2020-02-12
[BUGFIX release] Don't setup mandatory setters on array indexes

### LINKED ISSUE #18739: [3.16] Bug with `set` and array
The following code:

```javascript
const x = A([1]);
x.addObserver('0', function() {});
x.removeAt(0);
x.pushObject(2);
console.log(x[0]);
x.set(0, 3);
console.log(x[0]);
```

Prints two times `2`. The same is true when using `Ember.set` instead of the `set` method on the array.

The funny thing is that removing either the `removeAt` line or the `addObserver` line fixes the problem.

Any ideas for a workaround would be welcome as this is a problem for us.

P.S. Calling `x.replace(0, 1, [3]);` instead of `x.set(0, 3);` seems to work fine. However, this doesn't work in my case because I have a custom implementation of EmberArray and I've overwritten `replace` to behave differently (that's why I'm calling `Ember.set` internally and noticed that bug). How can I call the "original" `replace`? I tried `this._super` and a few other things as well but to no avail.

pzuraq:
It turns out that removing items from an array (e.g. via `splice`) will
also remove their descriptors, if any. Could be a browser bug, but I'm
not sure we intended to add setters directly to arrays in the first
place, and if we can't definitely manage them we probably shouldn't.

Fixes #18739

### DIFF
--- a/packages/@ember/-internals/utils/lib/mandatory-setter.ts
+++ b/packages/@ember/-internals/utils/lib/mandatory-setter.ts
@@ -14,6 +14,19 @@ export let setWithMandatorySetter:
 
 type PropertyDescriptorWithMeta = PropertyDescriptor & { hadOwnProperty?: boolean };
 
+function isElementKey(key: string | number | symbol) {
+  return typeof key === 'number' ? isPositiveInt(key) : isStringInt(key as string);
+}
+
+function isStringInt(str: string) {
+  let num = parseInt(str, 10);
+  return isPositiveInt(num) && str === String(num);
+}
+
+function isPositiveInt(num: number) {
+  return num >= 0 && num % 1 === 0;
+}
+
 if (DEBUG) {
   let SEEN_TAGS = new WeakSet();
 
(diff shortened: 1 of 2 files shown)


packages/@ember/-internals/utils/lib/mandatory-setter.ts
```diff
@@ -34,6 +34,12 @@ if (DEBUG) {
 
     SEEN_TAGS!.add(tag);
 
+    if (Array.isArray(obj) && !isNaN(Number(keyName))) {
```
rwjblue: After discussion with @krisselden, he suggested we change to using something more like:

```js
function isElementKey(key) {
    return typeof key === 'number' ? isPositiveInt(key) : isStringInt(key);
}

function isStringInt(str) {
    let num = parseInt(str, 10);
    return isPositiveInt(num) && str === '' + num;
}

function isPositiveInt(num) {
    return num >= 0 && num % 1 === 0;
}

// this would replace the conditional
if (isElementKey(keyName)) {
  return;
}
```

rwjblue: @pzuraq ping 😺

rwjblue (approved):

rwjblue (approved):

PR #18743, opened 2020-02-14
[BUGFIX beta] Upgrades Glimmer VM to 0.47.9

### LINKED ISSUE #18693: Bug using hash and array helpers together
There seems to be a bug that was introduced in ember-source@3.17.0-beta.1 🤔but interestingly it doesn't seem to have been happening on ember-canary before the beta was released

Essentially it looks like using the `{{array}}` and `{{hash}}` helpers together doesn't work with the error: 

```
TypeError: Cannot convert object to primitive value
```

I have produced a replication and run tests with ember try here: https://github.com/mansona/bug-hash-array-reproduction/pull/1 and the build is here https://travis-ci.com/mansona/bug-hash-array-reproduction/builds/145761580

You will see that it fails on ember-beta and ember-canary.

Here is the example of the template that fails: 

```
<table>
  <tbody>
    {{#each
      (array
        (hash value='' name='all')
        (hash value='x' name="horizontal")
        (hash value='y' name="vertical")
        (hash value='t' name="top")
        (hash value='r' name="right")
        (hash value='b' name="bottom")
        (hash value='l' name="left")
      ) as |item|
    }}

      <tr>
        <td>{{item.value}}</td>
        <td>{{item.name}}</td>
      </tr>
    {{/each}}
  </tbody>
</table>
```
The table is not required, I just put it in there to make it clear what an example app might be 👍

### LINKED ISSUE #18732: hasBlock always returns true for curly components in beta and canary
Here's simplified code with the problem.

```hbs
{{#card-form card=model as |form|}}
  {{!-- hasBlock will be `false` in 3.14 and `true` in beta and canary --}}
  {{form.submit-button}}

  {{!-- hasBlock will be true in any version --}}
  {{#form.submit-button}}
    Block text
  {{/form.submit-button}}
{{/card-form}}
```

`card-form.hbs`:

```hbs
{{yield
  (hash submit-button=(component "card-submit-button"))
}}
```

pzuraq:
Fixes #18732 
Fixes #18693


krisselden:
Is there tests for the fixed issues?

rwjblue (commented):
Update seems good, but I’d love to get some tests for those specific regressions.

PR #18752, opened 2020-02-20
Update documentation for redirect() hook

mehulkar:
I keep reading and re-reading this documentation and find it very confusing. This is my attempt at making it easier to understand.

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/route.ts
+++ b/packages/@ember/-internals/routing/lib/system/route.ts
@@ -1055,7 +1055,7 @@ class Route extends EmberObject implements IRoute {
     called almost at the same time, but they have an important
     distinction when calling `this.transitionTo` to a child route
     of the current route. From `afterModel`, this new transition
-    invalidates the current transition, causing `beforeModel`, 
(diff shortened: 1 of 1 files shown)


packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -1053,17 +1053,14 @@ class Route extends EmberObject implements IRoute {
```
rwjblue: The verbiage just above here is pretty confusing too:

> If you call `this.transitionTo` from inside of this hook, this route will not be entered in favor of the other hook.

Maybe you could update that also? I'm thinking something like:

```
Calling `this.transitionTo` from inside of the `redirect` hook will cause the current transition (into the route that has implemented `redirect`) to be aborted.
```

rwjblue (approved):
Going to land this, but I'd love another PR tweaking that section I called out...

packages/@ember/-internals/routing/lib/system/route.ts
mehulkar: Done in #18766

PR #18754, opened 2020-02-21
[BUGFIX beta] Add test for #18693

GCheung55:
[BUGFIX beta] Add test for #18693


pzuraq (approved):

rwjblue:
Updated to target master, should fail, then we can retarget to the glimmer-vm upgrade branch.

rwjblue:
Tests fail for correct reason against master:

> Cannot convert object to primitive value

rwjblue:
Thank you @GCheung55!

PR #18759, opened 2020-02-21
[DOCS release] improve types and docs for `transitionTo` methods

chriskrycho:
- Both `Route.transitionTo` and `RouterService.transitionTo` support query-params-only transitions, where the caller invokes the method like `transitionTo({ queryParams: { ... } })`. Document this invocation.
- A number of the type declarations in the router implementation were incorrect, overly loose, or both. Fix a number of them, and clearly mark safe casts with `// SAFETY: ...` and unsafe casts with `// UNSAFE: ...` comments explaining why the cast is or is not valid. Hopefully this will make it easier to (a) further improve the safety of this code later and (b) narrow down the cause of bugs since unsafe behavior is explicitly flagged.

### DIFF
--- a/packages/@ember/-internals/routing/lib/services/router.ts
+++ b/packages/@ember/-internals/routing/lib/services/router.ts
@@ -4,7 +4,7 @@ import { readOnly } from '@ember/object/computed';
 import Service from '@ember/service';
 import { DEBUG } from '@glimmer/env';
 import { Transition } from 'router_js';
-import EmberRouter from '../system/router';
+import EmberRouter, { QueryParam } from '../system/router';
 import { extractRouteArgs, resemblesURL, shallowEqual } from '../utils';
 
 let freezeRouteInfo: Function;
@@ -119,9 +119,11 @@ export default class RouterService extends Service {
        attempted transition
      @public
    */
-  transitionTo(...args: string[]) {
+  transitionTo(...args: (string | object)[]) {
     if (resemblesURL(args[0])) {
-      return this._router._doURLTransition('transitionTo', args[0]);
+      // NOTE: this `args[0] as string` cast is safe and TS correctly infers it
+      // in 3.6+, so it can be removed when TS is upgraded.
+      return this._router._doURLTransition('transitionTo', args[0] as string);
     }
 
     let { routeName, models, queryParams } = extractRouteArgs(args);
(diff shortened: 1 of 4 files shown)


packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -110,18 +110,21 @@ export default class RouterService extends Service {
...
      @param {Object} [options] optional hash with a queryParams property
-       containing a mapping of query parameters
+       containing a mapping of query parameters. May be supplied as the only
+      parameter to trigger a query-parameter-only transition.
      @return {Transition} the transition object associated with this
        attempted transition
      @public
    */
-  transitionTo(...args: string[]) {
+  transitionTo(...args: (string | object)[]) {
```
rwjblue: FWIW, I'd love to see us to the overloads for these methods

packages/@ember/-internals/routing/lib/system/router.ts
```diff
@@ -517,7 +517,7 @@ class EmberRouter extends EmberObject {
     @public
   */
   transitionTo(...args: unknown[]) {
-    if (resemblesURL(args[0] as string)) {
+    if (resemblesURL(args[0])) {
```
rwjblue: Mind making an issue for a random aside that this made me think about?

If you call `.transitionTo` with a thing that "resembles a URL" but also pass more than one string arg, we silently swallow the options 😭

packages/@ember/-internals/routing/lib/utils.ts
```diff
@@ -8,20 +8,28 @@ import EmberRouter, { PrivateRouteInfo, QueryParam } from './system/router';
...
 
-export function extractRouteArgs(args: any[]) {
+export function extractRouteArgs(args: unknown[]) {
   args = args.slice();
   let possibleQueryParams = args[args.length - 1];
 
   let queryParams;
-  if (possibleQueryParams && possibleQueryParams.hasOwnProperty('queryParams')) {
-    queryParams = args.pop().queryParams;
+  if (possibleQueryParams && (possibleQueryParams as object).hasOwnProperty('queryParams')) {
```
rwjblue: We shouldn't use this object's `.hasOwnProperty` 😢 (this breaks when you pass in the result of `(hash` for example, because there is no prototype)

packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -110,18 +110,21 @@ export default class RouterService extends Service {
...
      @method transitionTo
-     @param {String} routeNameOrUrl the name of the route or a URL
-     @param {...Object} models the model(s) or identifier(s) to be used while
+     @param {String} [routeNameOrUrl] the name of the route or a URL
+     @param {...Object} [models] the model(s) or identifier(s) to be used while
        transitioning to the route.
      @param {Object} [options] optional hash with a queryParams property
-       containing a mapping of query parameters
+       containing a mapping of query parameters. May be supplied as the only
+      parameter to trigger a query-parameter-only transition.
```
rwjblue: Can you make an inline code example for this case in the docs above?

packages/@ember/-internals/routing/lib/system/router.ts
chriskrycho: I figured something like that might happen given the way this works. 😬 Opening the issue now!

chriskrycho: [issue!](https://github.com/emberjs/ember.js/issues/18760)

packages/@ember/-internals/routing/lib/utils.ts
chriskrycho: Yeah, I assumed this was a bug for exactly that reason, but opted very carefully not to make any functional changes in this PR. I'm very happy to submit a follow-on for that, but wanted to keep this one to runtime-safe docs-and-types-only changes.

packages/@ember/-internals/routing/lib/services/router.ts
chriskrycho: Yes, great callout! Will push in just a minute!

chriskrycho: 9808da9

packages/@ember/-internals/routing/lib/utils.ts
rwjblue: Sounds good, mostly just reading the PR made me cringe a bit 🗡

rwjblue (approved):

rwjblue:
Thanks @chriskrycho!

chriskrycho:
👍 😁

Issue #18211, opened 2019-07-17
[Regression] 3.11 ignores checked attribute on input


jrjohnson:
```hbs
<input type="checkbox" checked {{action "someAction"}}>
```
Shows the checkbox checked in `3.10.2`.

Starting with `3.11.0-beta.1` it is **not** checked.

These slight format variations:
```hbs
<input type="checkbox" checked click={{action "someAction"}}>
<input type="checkbox" checked onclick={{action "someAction"}}>
<input type="checkbox" checked={{true}} {{action "someAction"}}>
```
all work and show up as checked.

There don't seem to be errors thrown.
I may just be binding this action in an old/unsupported way.

rwjblue:
Closed by https://github.com/emberjs/ember.js/pull/18743 (confirmed by https://github.com/emberjs/ember.js/pull/18762).

PR #18762, opened 2020-02-22
[BUGFIX beta] Adding test for #18211

GCheung55:
Tests to verify #18211 .


rwjblue:
Thank you @GCheung55!

PR #18767, opened 2020-02-26
[BUGFIX release] Fix observer leaks

krisselden:


packages/@ember/-internals/metal/lib/destroy.ts
```diff
@@ -0,0 +1,19 @@
+import { Meta, peekMeta } from '@ember/-internals/meta/lib/meta';
+import { schedule } from '@ember/runloop';
+import { destroyObservers } from './observer';
+
+export function destroy(obj: object): boolean {
```
rwjblue: We should update `Ember.destroy` to use this. After reading, I was confused by the return value. Maybe an inline comment that explains what the boolean is for (`false` for "already done" and `true` for "just did the destroy").

packages/@ember/-internals/metal/tests/accessors/get_test.js
```diff
@@ -180,7 +181,7 @@ moduleFor(
     // BUGS
     //
 
-    ['@test (regression) watched properties on unmodified inherited objects should still return their original value'](
+    async '@test (regression) watched properties on unmodified inherited objects should still return their original value'(
```
rwjblue: Why change this to async (with no await)?

packages/@ember/-internals/metal/tests/computed_test.js
```diff
@@ -962,15 +1027,26 @@ moduleFor(
...
 moduleFor(
   'computed - default setter',
   class extends AbstractTestCase {
+    afterEach() {
+      if (obj !== undefined) {
+        destroy(obj);
+        obj = undefined;
+      }
+      return runLoopSettled();
+    }
```
rwjblue: Might have been easier to make a custom class in this file:

```js
let obj;
class ObjCleaningTestCase extends AbstractTestCase {
  afterEach() {
    if (obj !== undefined) {
      destroy(obj);
      obj = undefined;
      return runLoopSettled();
    }
  }
}
```

And updated the `moduleFor`'s use it instead of `AbstractTestCase`

rwjblue (approved):

packages/@ember/-internals/metal/lib/destroy.ts
krisselden: added documentation @returns

tests/docs/expected.js
```diff
@@ -164,7 +163,7 @@ module.exports = {
     'defer',
     'deferReadiness',
     'defineProperty',
-    'deleteMeta',
+    'destroy',
```
rwjblue: Destroy is already in the list, you should be able to delete this

rwjblue (approved):

ginomiglio:
Could someone explain the scope of the bug this PR fixes? (And do we know what version introduced it?)

PR #18770, opened 2020-02-27
[BUGFIX release] Make ArrayProxy Lazy 

### LINKED ISSUE #18750: [3.17.0 beta regression] Array Proxy not updating
I've found what appears to be a regression with `ArrayProxy` that was introduced in this commit https://github.com/emberjs/ember.js/pull/18703 in **3.17.0-beta.3**. **3.17.0-beta.2** is working as expected.

These changes were also backported in https://github.com/emberjs/ember.js/pull/18721 to 3.16.2 which means the regression is also on 3.16.2, but works on 3.16.1.

I've got a reproducing test in my repo here. While the test is a bit convoluted this is a bare bones reproduction of some code in our app where we are seeing the issue.

https://github.com/patocallaghan/array-proxy-regression/blob/master/tests/unit/array-proxy-test.js

It seems that modifying the ArrayProxy internally is not picking up all the changes.

/cc @pzuraq

### LINKED ISSUE #18763: [3.16.2] Regression with ember-power-select
I'm using ember-power-select (latest v3) to show a select element with options provided by ember-data.

With ember 3.16.1, this used to work to populate the dropdown. After upgrading to 3.16.2 or 3.16.3, this no longer works. No error or warning, just a dropdown with only the current element shown. There is an API call being performed, but the result seems to be ignored.

Upgrading between patch versions shouldn't cause such regressions I suppose?

As there's no console output, I'm not sure what else to report. Please advice.

`select-customer.ts`:
```ts
export default class SelectCustomer extends Component {
    @inject store!: DS.Store;
    options: any
    init() {
        super.init()
        this.options = this.store.findAll('customer')
    }
}
```

`select-customer.hbs`:
```hbs
<PowerSelect @options={{this.options}} @searchEnabled={{true}} @search={{this.search}}
    @selected={{this.selected}} @onChange={{this.onChange}} as |customer|>
    {{customer.name}}
</PowerSelect>
```

```
00:15:14.521 DEBUG: -------------------------------
00:15:14.550 DEBUG: Ember      : 3.16.2 index.js:194
00:15:14.551 DEBUG: Ember Data : 3.16.0 index.js:194
00:15:14.551 DEBUG: jQuery     : 2.1.1 index.js:194
00:15:14.551 DEBUG: -------------------------------
```

### LINKED ISSUE #18773: [Possible 3.15 regression] Model Fragment's `fragmentArray` throws backtracking rerender assertion from 3.15.x
Our app relies heavily on the addon [ember-data-model-fragments](https://github.com/lytics/ember-data-model-fragments), an addon which integrates with Ember Data to allow you to treat blobs of JSON as models. I'm currently trying to upgrade our app from **ember-source** 3.14.x to 3.15.x but I've found a blocker which is preventing our upgrade due to a change in ember-source 3.15. (I've also tested with v3.17.0-beta.6 and it's still the same)

This issue is a similar shape to the issue ["Possible 3.15/Octane regression with ember-table addon"](https://github.com/emberjs/ember.js/issues/18613), backtracking rerender assertions have started triggered due to the ember-source changes. I guess what I’m trying to figure out:

1. Is this is a valid triggering of the backtracking rerender assertion? Or ...
2. Is this an edge-case bug and the assertion shouldn't be triggered?

If this is a case where ember-data-model-fragments is doing bad stuff, I’d like to try and understand how to fix to unblock myself 😃

From [ember-source 3.15.0-beta.4](https://github.com/emberjs/ember.js/releases/tag/v3.15.0-beta.4) it appears that tracked models which have properties using `fragmentArray` will throw a backtracking rerender assertion causing things to blow up.

There are two suspect commits in ember-source where this was introduced. 

> emberjs/ember.js#17834 [BUGFIX] Prevents autotracking ArrayProxy creation
> emberjs/ember.js#18554 [BREAKING BUGFIX] Adds autotracking transaction

I think it's emberjs/ember.js#17834 which introduced it (cc-ing @pzuraq who made this change).

I've created an extremely simplified test case in https://github.com/lytics/ember-data-model-fragments/pull/358 and it throws the following assertion:

```
Error: Assertion Failed: You attempted to update `[]` on `Array`, but it had already been used previously in the same computation.  Attempting to update a value after using it in a computation can cause logical errors, infinite revalidation bugs, and performance issues, and is not supported.

`[]` was first used:

- While rendering:
  
  application
    index
      my-component
        this.record

Stack trace for the first usage: 
    at get (http://localhost:7357/assets/vendor.js:14851:32)
    at Class._addArrangedContentArrayObserver (http://localhost:7357/assets/vendor.js:28971:44)
    at Class.init (http://localhost:7357/assets/vendor.js:28831:12)
    at Class.init (http://localhost:7357/assets/vendor.js:87960:12)
    at Class.superWrapper [as init] (http://localhost:7357/assets/vendor.js:30845:22)
    at initialize (http://localhost:7357/assets/vendor.js:29137:9)
    at Function.create (http://localhost:7357/assets/vendor.js:29720:9)
    at createFragmentArray (http://localhost:7357/assets/vendor.js:88297:32)

Stack trace for the update:
    at dirtyTagFor (http://localhost:7357/assets/vendor.js:55641:11)
    at markObjectAsDirty (http://localhost:7357/assets/vendor.js:14006:32)
    at notifyPropertyChange (http://localhost:7357/assets/vendor.js:14043:5)
    at arrayContentDidChange (http://localhost:7357/assets/vendor.js:14144:7)
    at replaceInNativeArray (http://localhost:7357/assets/vendor.js:14210:5)
    at Array.replace (http://localhost:7357/assets/vendor.js:27255:39)
    at Class.setupData (http://localhost:7357/assets/vendor.js:88000:15)
```

The problem appears to happen in the following lines in ember-data-model-fragments, specifically 298 (where the array is created) and 300 (where the array is updated) https://github.com/lytics/ember-data-model-fragments/blob/c00e2b1a40b296e0c0c3c21ca0f463d12888cd27/addon/attributes.js#L298-L300

The array is first set up in `createArray` at 

https://github.com/lytics/ember-data-model-fragments/blob/c00e2b1a40b296e0c0c3c21ca0f463d12888cd27/addon/attributes.js#L195-L208

the array is modified a second time in 

https://github.com/lytics/ember-data-model-fragments/blob/f1c7060fb7560aaea26760e0270234823b9e20db/addon/array/stateful.js#L79 

In my [test reproduction](https://github.com/lytics/ember-data-model-fragments/pull/358) I am specifically marking the property as `@tracked`

```js
@tracked selectedRecord = this.store.createRecord('my-model', { fragments: [] });
```

pzuraq:
This PR refactors ArrayProxy to only begin observing the proxied array's
changes the first time the proxy is actually used. This fixes a number
of bugs that have popped up recently, and also ensures that ArrayProxy
creation does not accidentally entangle with the autotracking stack. It
also simplifies the code overall, and conceptually is more in line with
the way that autotracking works.

In addition, it adds a CUSTOM_TAG_FOR to the proxy, to forward the `[]`
and `length` tags from its content directly rather than attempting to
managed them itself. This keeps the system from encountering issues
related to tag updates, and from needing to re-notify when changes are
propagated.

As part of fixing this, I also discovered that the `hasArrayObservers`
API has been broken since ~3.11. I fixed it and added a test.

Fixes #18763
Fixes #18750 
Fixes #18773

### DIFF
--- a/packages/@ember/-internals/runtime/lib/system/array_proxy.js
+++ b/packages/@ember/-internals/runtime/lib/system/array_proxy.js
@@ -11,14 +11,13 @@ import {
   removeArrayObserver,
   replace,
   getChainTagsForKey,
-  tagForProperty,
+  CUSTOM_TAG_FOR,
   arrayContentDidChange,
-  arrayContentWillChange,
 } from '@ember/-internals/metal';
 import EmberObject from './object';
 import { isArray, MutableArray } from '../mixins/array';
 import { assert } from '@ember/debug';
-import { combine, update, validate, value } from '@glimmer/validator';
+import { combine, validate, value, tagFor } from '@glimmer/validator';
 
 const ARRAY_OBSERVER_MAPPING = {
   willChange: '_arrangedContentArrayWillChange',
@@ -106,18 +105,24 @@ export default class ArrayProxy extends EmberObject {
     this._length = 0;
 
     this._arrangedContent = null;
-
     this._arrangedContentIsUpdating = false;
-    this._arrangedContentTag = combine(getChainTagsForKey(this, 'arrangedContent'));
-    this._arrangedContentRevision = value(this._arrangedContentTag);
+    this._arrangedContentTag = null;
+    this._arrangedContentRevision = null;
+  }
 
-    this._addArrangedContentArrayObserver();
+  [PROPERTY_DID_CHANGE]() {
+    this._revalidate();
+  }
 
-    update(tagForProperty(this, '[]'), combine(getChainTagsForKey(this, 'arrangedContent.[]')));
-    update(
-      tagForProperty(this, 'length'),
-      combine(getChainTagsForKey(this, 'arrangedContent.length'))
-    );
+  [CUSTOM_TAG_FOR](key) {
+    if (key === '[]' || key === 'length') {
+      // revalidate eagerly if we're being tracked, since we no longer will
+      // be able to later on due to backtracking re-render assertion
+      this._revalidate();
+      return combine(getChainTagsForKey(this, `arrangedContent.${key}`));
+    }
+
+    return tagFor(this, key);
   }
 
   willDestroy() {
@@ -236,10 +241,6 @@ export default class ArrayProxy extends EmberObject {
     }
   }
 
-  [PROPERTY_DID_CHANGE]() {
-    this._revalidate();
-  }
-
   _updateArrangedContentArray() {
     let oldLength = this._objects === null ? 0 : this._objects.length;
     let arrangedContent = get(this, 'arrangedContent');
@@ -301,13 +302,21 @@ export default class ArrayProxy extends EmberObject {
   }
 
   _revalidate() {
+    if (this._arrangedContentIsUpdating === true) return;
+
     if (
-      !this._arrangedContentIsUpdating &&
+      this._arrangedContentTag === null ||
       !validate(this._arrangedContentTag, this._arrangedContentRevision)
     ) {
-      this._arrangedContentIsUpdating = true;
-      this._updateArrangedContentArray();
-      this._arrangedContentIsUpdating = false;
+      if (this._arrangedContentTag === null) {
+        // This is the first time the proxy has been setup, only add the observer
+        // don't trigger any events
+        this._addArrangedContentArrayObserver();
+      } else {
+        this._arrangedContentIsUpdating = true;
+        this._updateArrangedContentArray();
+        this._arrangedContentIsUpdating = false;
+      }
 
       this._arrangedContentTag = combine(getChainTagsForKey(this, 'arrangedContent'));
       this._arrangedContentRevision = value(this._arrangedContentTag);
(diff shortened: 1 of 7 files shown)


packages/@ember/-internals/runtime/lib/system/array_proxy.js
```diff
@@ -124,6 +124,16 @@ export default class ArrayProxy extends EmberObject {
     this._removeArrangedContentArrayObserver();
   }
 
+  [CUSTOM_TAG_FOR](key) {
+    if (key === '[]') {
```
krisselden: is 'length' not needed?

pzuraq: This is the tag that entangled by `get` and `#each`, so I don't think so. If `length` is entangled, that means user code has definitely entangled it by accessing it directly, so it'll already have updated (since it's a real property).

pzuraq: Ended up adding it since we need to entangle the tag anyways.

rwjblue (approved):

PR #18782, opened 2020-02-28
Ensure the test suite runs with Octane's set of optional features.

rwjblue:
Prior to these change we would only run the test suite in "classic" mode,
and did not run the tests with the combination of octane optional features
setup.

This adds a new test run for "classic", and makes the default test run use
"octane"'s set of optional features.

### DIFF
--- a/tests/index.html
+++ b/tests/index.html
@@ -40,6 +40,15 @@
       (function() {
         window.EmberENV = window.EmberENV || {};
 
+        // Determine edition, and apply correct EmberENV settings
+        var edition = QUnit.urlParams.edition || 'octane';
+
+        if (edition === 'octane') {
+          EmberENV._TEMPLATE_ONLY_GLIMMER_COMPONENTS = true;
+          EmberENV._APPLICATION_TEMPLATE_WRAPPER = false;
+          EmberENV._DEFAULT_ASYNC_OBSERVERS = true;
+        }
+
         // Test for "hooks in EmberENV.EMBER_LOAD_HOOKS['hookName'] get executed"
         EmberENV.EMBER_LOAD_HOOKS = EmberENV.EMBER_LOAD_HOOKS || {};
         EmberENV.EMBER_LOAD_HOOKS.__before_ember_test_hook__ = EmberENV.EMBER_LOAD_HOOKS.__before_ember_test_hook__ || [];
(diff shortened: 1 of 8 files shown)


tests/index.html
```diff
@@ -40,6 +40,15 @@
       (function() {
         window.EmberENV = window.EmberENV || {};
 
+        // Determine edition, and apply correct EmberENV settings
+        var edition = QUnit.urlParams.edition || 'octane';
+
+        if (edition === 'octane') {
+          EmberENV._TEMPLATE_ONLY_GLIMMER_COMPONENTS = true;
+          //EmberENV._APPLICATION_TEMPLATE_WRAPPER = false;
```
rwjblue: This is commented out because it causes a test failure in this backtracking rerender test:

https://github.com/emberjs/ember.js/blob/cfb74eb8b83b70773085b0542e8e6a08c7eb176d/packages/@ember/-internals/glimmer/tests/integration/application/rendering-test.js#L976-L1025

Is failing during cleanup of test harness. It presents in the same way as https://github.com/emberjs/ember-test-helpers/issues/768.

rwjblue:
Seeing sporadic failures on CI for these two tests:

```
Test failed: changeProperties:  observers added/removed during changeProperties should do the right thing.", source:  (49)
[0228/021013.969380:INFO:CONSOLE(52)] "    Failed assertion: addObserver called before the first change sees only 1, expected: 1, but was: 0
Test failed: changeProperties:  calling changeProperties while executing deferred observers works correctly", source:  (49)
[0228/021014.060138:INFO:CONSOLE(52)] "    Failed assertion: expected: 1, but was: 0
```

tests/index.html
```diff
@@ -40,6 +40,15 @@
       (function() {
         window.EmberENV = window.EmberENV || {};
 
+        // Determine edition, and apply correct EmberENV settings
+        var edition = QUnit.urlParams.edition || 'octane';
+
+        if (edition === 'octane') {
+          EmberENV._TEMPLATE_ONLY_GLIMMER_COMPONENTS = true;
```
krisselden: is there a source of truth for this? not a big deal if not just curious

krisselden (approved):

tests/index.html
rwjblue: https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/config/optional-features.json

PR #18787, opened 2020-02-28
[DOC release] `fn` helper has no `value` param

aaxelb:
the `mut` helper docstring has an example using the `fn` helper with
`value` param, but `fn` has no `value` param.

looks like a small oversight updating the docs away from `action`.


locks:
Well spotted! Thank you for the contribution.

locks:
Keep'em coming :)

PR #18792, opened 2020-03-04
[DOCS] Update route.ts

hakilebara:
Update `activate()` and `deactivate()` code snippets to use native ES6 classes.


MelSumner (approved):
Thank you for updating this, @hakilebara 👍

locks (changes requested):
Thank you for the contribution, but note that this is the documentation for the _event_, the method is located [below](https://github.com/emberjs/ember.js/blob/8ef7978feb839e5f05bef93753e1a83bdef44102/packages/%40ember/-internals/routing/lib/system/route.ts#L589-L597).
Since the event needs the use of `on` (and I don't think we expose it as a decorator?), this code sample should continue using `EmberObject` syntax. Improving the documentation for the method is a good idea tho!

locks:
Closing this since the release branch has been reset for 3.17.

PR #18807, opened 2020-03-11
[BUGFIX LTS] do not throw on stable `elementId`

### LINKED ISSUE #18147: 3.11 regression when using "elementId" in templates
After upgrading my code to 3.11, I receive an error: `"Error: Changing a view's elementId after creation is not allowed"`.

Since I don't change the `elementId` anywhere in my code, and the new value is identical to the recorded value, I have explored further and found it to be a new bug.

I can trigger the issue with the following:

#### templates/application.hbs
```hbs
{{my-test elementId="my-id" handler=(action test) value=myValue}}
```

#### controllers/application.js

```js
export default Controller.extend({
    myValue: 123,

    test() {
        console.log('not used...');
    },

    init() {
        this._super(...arguments);

        next(() => {
            this.set('myValue', 321);
        });
    }
});
```

#### templates/components/my-test.hbs
```hbs
Empty...
```

You can find failing copy here: https://github.com/kanongil/ember-fail

Note that I also tested it with angle-bracket components, and here it doesn't fail until I make the value `{{mut}}`.

It seems that the bug is triggered by a combination of using elementId, an action handler, and a mutable value that is set after it has been rendered.

chriskrycho:
When refactoring from observers to setters during the 3.11 release (in 405d4238), an apparently-transparent change caused a regression around invocations of components including `elementId=...`. *Any* rerender of the component which included `elementId` assignment now triggered the assertion, rather than *only* changes which actually updated the value of `elementId`, because all invocations strigger a `set` on the property.

The simplest reproduction of this bug (given a component `foo-bar`):

```hbs
{{foo-bar
  elementId='stable'
  changingValue=this.fromBackingClass
}}
```

Changing the value `fromBackingClass` on the backing class *always* triggers the assertion, even though it is actually impossible for it to change the `elementId` value, because the assertion in the setter throws regardless of what the value is. (The observer-based did not throw in the same conditions because it would not retrigger when the observed key on the view did not change.)

The fix is to check equality between the passed `elementId` value and the previously set value, and only throw if they differ.

Fixes #18147

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/components/curly-components-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/components/curly-components-test.js
@@ -109,6 +109,39 @@ moduleFor(
       }
     }
 
+    ['@test elementId is stable when other values change']() {
+      let changingArg = 'arbitrary value';
+      let parentInstance;
+      this.registerComponent('foo-bar', {
+        ComponentClass: Component.extend({
+          init() {
+            this._super(...arguments);
+            parentInstance = this;
+          },
+          changingArg: changingArg,
+        }),
+        template: '{{quux-baz elementId="stable-id" changingArg=this.changingArg}}',
+      });
+
+      this.registerComponent('quux-baz', {
+        ComponentClass: Component.extend({}),
+        template: '{{changingArg}}',
+      });
+
+      this.render('{{foo-bar}}');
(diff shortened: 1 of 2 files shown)


chriskrycho:
@rwjblue @pzuraq I would definitely appreciate your eyes on this change!

pzuraq (approved):

pzuraq:
Looks good, thanks for getting this done!

PR #18814, opened 2020-03-13
[WIP] Instrument glimmer components

achambers:
WIP

### Initial render
image

### Rerender
image

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/component-managers/custom.ts
+++ b/packages/@ember/-internals/glimmer/lib/component-managers/custom.ts
@@ -1,6 +1,6 @@
 import { CUSTOM_TAG_FOR } from '@ember/-internals/metal';
 import { Factory } from '@ember/-internals/owner';
-import { HAS_NATIVE_PROXY } from '@ember/-internals/utils';
+import { guidFor, HAS_NATIVE_PROXY } from '@ember/-internals/utils';
 import { EMBER_CUSTOM_COMPONENT_ARG_PROXY } from '@ember/canary-features';
 import { assert } from '@ember/debug';
 import { DEBUG } from '@glimmer/env';
@@ -18,6 +18,7 @@ import {
 import { ComponentRootReference, PathReference } from '@glimmer/reference';
 import { unwrapTemplate } from '@glimmer/util';
 import { consume, createTag, isConst, Tag } from '@glimmer/validator';
+import { _instrumentStart } from '@ember/instrumentation';
 
 import { ENV } from '@ember/-internals/environment';
 import { EmberVMEnvironment } from '../environment';
@@ -142,6 +143,10 @@ export interface ManagerDelegateWithDestructors<ComponentInstance>
 export interface ComponentArguments {
   positional: unknown[];
   named: Dict<unknown>;
+}
+
+export function initialRenderInstrumentDetails(component: any): any {
+  return { object: `<${component.constructor.name}:${guidFor(component)}:gc>`, initialRender: true };
 }
 
 /**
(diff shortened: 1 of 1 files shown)


packages/@ember/-internals/glimmer/lib/component-managers/custom.ts
```diff
@@ -144,6 +145,14 @@ export interface ComponentArguments {
   named: Dict<unknown>;
 }
 
+export function initialRenderInstrumentDetails(component: any): any {
+  return { object: `<${component.constructor.name}:${guidFor(component)}:gc>`, initialRender: true };
```
achambers: The existing implementation in the `curly` component manager uses `component.instrumentDetails` to generate the `object` string, which in turn relies on the `toString` function of `EmberObject`. I wasn't really sure the best way forward here so just used this for now.

pzuraq:
@achambers thanks for getting this started!

I think that if we were to start instrumenting components in general, we would probably want to only do so in DEBUG builds, or if the Ember Inspector were installed and open. It's a lot of extra work to do in general for every user.

Beyond that, we've been discussing moving the DebugRenderTree upstream into the Glimmer VM, which would make more sense because it could be more generalized. I had a conversation about this functionality with @krisselden the other day, and he also thought it would make most sense if we required instrumenting to be an opt-in feature, so you manually mark a component before it shows up in the timeline.

This is all just some context, that refactor would be pretty large anyways so it could definitely pull this work along with it if/when it happens 😄 just wanted to let you know what the current thinking is

achambers:
> @achambers thanks for getting this started!
> 
> I think that if we were to start instrumenting components in general, we would probably want to only do so in DEBUG builds, or if the Ember Inspector were installed and open. It's a lot of extra work to do in general for every user.
> 
> Beyond that, we've been discussing moving the DebugRenderTree upstream into the Glimmer VM, which would make more sense because it could be more generalized. I had a conversation about this functionality with @krisselden the other day, and he also thought it would make most sense if we required instrumenting to be an opt-in feature, so you manually mark a component before it shows up in the timeline.
> 
> This is all just some context, that refactor would be pretty large anyways so it could definitely pull this work along with it if/when it happens 😄 just wanted to let you know what the current thinking is

Thanks for taking the time to reply @pzuraq. Makes sense. Also, some extra context that @krisselden provided in Discord also made some very valid points.

So, just for context, we've been doing some serious performance profiling of a page on our Ember app and the [ember-perf-timeline](https://github.com/ember-best-practices/ember-perf-timeline) addon has been super useful. However, we've needed to convert all of our Glimmer components to Ember components just so that we could profile them. Which is what spurred on the conversation with @rwjblue that lead to me spiking this out.

Would I be correct in summarizing your reply by saying that there are a couple of places that the instrumentation _could_ take place but the correct and most desirable place for it to happen is actually inside the Glimmer VM which is a considerable piece of work. And, reading between the lines, this isn't likely to happen any time soon?

Is there anything we can do to help out with this or help move it forward or is it a deeper issue than that needs solid core team focus and involvement?

And out of interest, what sense of priority does being able to profile Glimmer components have in relation to other things going on?

kategengler:
Is this still relevant?

achambers:
> Is this still relevant?

By the looks of @pzuraq 's repy, this PR is not, but the intent is, as far as I understand. I could be wrong but I'm not aware that instrumentation of Glimmer components was implemented. Not having this meant that Glimmer components don't appear in the rendering tab of the Ember Inspector and can't be measured on the Chrome performance tab.

If this has, indeed, been implemented then this is no longer relevant and my understanding is outdated :)

PR #18815, opened 2020-03-13
Optimize the cache

pieter-v:
Our application has a large data set. One use-case includes destroying a large amount of data.
Analysing this scenario, I noticed that 20 seconds was spent in `Cache.get`: 10 seconds in `store.get` and 10 seconds in `store.has`.
By assuming that the return value never will be `undefined` (I verified this in the code), then the `store.has` is not needed.
If it is required that the return value of the function should also include `undefined`. Then this algorithm could be modified to optimise the hits over the misses. By changing the test `value !== undefined` too `value !== undefined || this.store.has(key)`

### DIFF
--- a/packages/@ember/-internals/utils/lib/cache.ts
+++ b/packages/@ember/-internals/utils/lib/cache.ts
@@ -1,9 +1,4 @@
-// The type `V` is just anything but `undefined`, so the `Cache.get` can be optimised
-// since a call to `Map.has` is not needed anymore
-export default class Cache<
-  T,
-  V extends object | number | string | boolean | bigint | symbol | null
-> {
+export default class Cache<T, V> {
   public size = 0;
   public misses = 0;
   public hits = 0;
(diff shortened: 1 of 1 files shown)


packages/@ember/-internals/utils/lib/cache.ts
```diff
@@ -1,21 +1,26 @@
-export default class Cache<T, V> {
+export default class Cache<
+  T,
+  V extends object | number | string | boolean | bigint | symbol | null
```
pzuraq: Can you add a comment explaining why this is essentially any type except `undefined`? This is a common technique/optimization when using Maps so it should be apparent, but a comment would help prevent anyone from accidentally removing this optimization in the future.

pzuraq (commented):
It looks like there's a test failing that explicitly tests the `undefined` behavior. I'm not sure it's ever actually used, though, and like you pointed out we could do a different optimization if it is.

I'd personally be ok with this optimization, though I think that ideally we would perform a benchmark to see what the difference is. We've been working on getting benchmarks like setup and automated, but we're not quite there yet.

rwjblue (commented):
In general, this seems like a good change to make. We just need to make sure that `undefined` as a value is still cacheable (probably by using a sentinel value and comparing it directly, but we'd need to test to see if that kills the test improvement).

kategengler:
Is this still relevant?

kategengler:
Closing as abandoned

PR #18821, opened 2020-03-17
[BUGFIX release] Fix {{each-in}} for array-like iterables

pzuraq:
Prior to the VM upgrade, the `{{each-in}}` iterator would treat
array-like native iterables as if they were an array, with values being
the value and the position in the order of the iteration being the
index. Map-like native iterables, e.g. ones whose values were arrays of
key/value pairs, would by contrast _extract_ the keys and values, and
use the key as the index and the value as the value instead of the
tuple.

This was lost in translation, and this PR adds it back, along with a
test.

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/syntax/each-in-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/syntax/each-in-test.js
@@ -618,7 +618,7 @@ moduleFor(
 );
 
 moduleFor(
-  'Syntax test: {{#each-in}} with ES6 Maps',
+  'Syntax test: {{#each-in}} with ES6 Maps and Sets',
   class extends EachInTest {
     createHash(pojo) {
       let map = new Map();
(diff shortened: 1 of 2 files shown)


packages/@ember/-internals/glimmer/tests/integration/syntax/each-in-test.js
```diff
@@ -672,6 +672,44 @@ moduleFor(
       </ul>
     `);
     }
+
+    [`@test it supports native ES6 Sets`]() {
```
rwjblue: Hmm, why use `{{#each-in` with a `Set`? I would have expected to use `{{#each` 🤔

pzuraq: Agreed, but this behavior did exist beforehand. We could throw an error instead potentially, not sure if it was considered public API.

rwjblue: Throwing seems better to me, also I assume this is just something you noticed in passing (since there are no linked issues)?

pzuraq: Yeah. Ok, sounds good 👍

rwjblue:
@pzuraq - Is this still valid?

PR #18824, opened 2020-03-20
Remove deprecated `window.ENV`

btecu:


rwjblue (approved):

rwjblue:
Changes seem good, but need to get CI passing before we can land...

locks:
Rerunning the browserstack job seems to have done the trick.

PR #18825, opened 2020-03-20
Remove deprecated `implicit-modifier-manager-capabilities`

btecu:


rwjblue (approved):

rwjblue:
Changes seem good, but need to get CI passing before we can land...

PR #18828, opened 2020-03-20
prepend 'TODO: ' to 'Replace this with your real tests' comments

jaydgruber:
### What
- adds `TODO: ` to the generated `Replace this with your real tests` comments in test files
- adds the same to corresponding node-tests fixtures

### Why
- for those in a rush / without strong reviews, you end up with plenty of checked in test files that only have canary tests:
```javascript
  test('it exists', function (assert) {
    let service = this.owner.lookup('service:company');
    assert.ok(service);
  });
```
- many devs have tooling that highlights TODO or FIXME or similar comments
- this change will nudge developers to start writing a meaningful test earlier in the process
- inspired by adding a lint rule [here](https://github.com/ember-cli/eslint-plugin-ember/issues/717)

As far as I can tell, YUIDOC does not have a standard format for TODO comments

JSDOC has `@todo`

I found a good list of conventions while looking for a highlighter plugin for SublimeText [here](https://forum.sublimetext.com/t/todo-syntax-highlighting/27604/11)

### DIFF
--- a/blueprints/controller-test/mocha-0.12-files/__root__/__testType__/__path__/__test__.js
+++ b/blueprints/controller-test/mocha-0.12-files/__root__/__testType__/__path__/__test__.js
@@ -8,7 +8,7 @@ describe('<%= friendlyTestDescription %>', function() {
     // needs: ['controller:foo']
   });
 
-  // Replace this with your real tests.
+  // TODO: Replace this with your real tests.
   it('exists', function() {
     let controller = this.subject();
     expect(controller).to.be.ok;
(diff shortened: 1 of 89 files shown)


bmish (approved):

jaydgruber:
Looks like I missed a test against a fixture somewhere

jaydgruber:
aha. I missed the entire node-tests directory. that oughta do it

rwjblue (approved):

rwjblue:
Thank you!

PR #18831, opened 2020-03-21
[BUGFIX lts] Fix transpilation issues (e.g. `_createSuper` is not a function) when used with Babel 7.9.0+.

### LINKED ISSUE #18827: (0 , _emberBabel.createSuper) is not a function. Recent babel 7.9.0 seems to break all apps
Looks like even all apps are broken that pickup `@babel` `7.9.0` stuff which was recently published.

If you do `ember new tester-if-broken`. Then remove the `config/target.js` file you'll get that error after you visit the page after a `ember s`

Not sure if this is the right repo (let me know if I should move it)

But definitely seems related to the recent @babel 7.9.0 (pushed 3 hours ago).  If have the 7.8.3 in my app, works fine. But as soon as I upgrade something that has babel it will pick up the new 7.9.0 and it will break.

rwjblue:
Due to the way Ember itself is compiled, we have custom versions of the `@babel/helpers` that we inline into the IIFE that contains Ember. This is obviously pretty unideal (largely because it forces us to be misaligned with Babel and therefore brittle to reasonable changes that they make).

These helpers have been pretty stable for quite a while, but recently Babel 7.9.0 introduced a couple of new helpers:

* `createSuper` allows subclassing from native classes, even when partially transpiled).
* `createForOfIteratorHelperLoose`

Versions of `ember-source` prior to 3.13 fully transpiled with the Babel version in `emberjs/ember.js`s _own_ `yarn.lock` at publish time. The build process was significantly changed in `ember-source@3.13.0` so that `ember-source` would behave just like a normal addon. This was a **huge** improvement!! Unfortunately, those changes also made us vulnerable to changes made to Babel (new helpers added).

We **must** resolve that issue (likely by removing these custom external helpers in favor of using `includeExternalHelpers` from `ember-cli-babel`), but in the meantime this commit adds the required helpers which should fix the reported issue in supported Ember versions.

Fixes https://github.com/emberjs/ember.js/issues/18827


rwjblue:
Confirmed failure:

```
Uncaught TypeError: (0 , _emberBabel.createSuper) is not a function
TypeError: (0 , _emberBabel.createSuper) is not a function
    at http://localhost:13141/ember-template-compiler.js:6962:46
    at http://localhost:13141/ember-template-compiler.js:7303:4
    at internalRequire (http://localhost:13141/tests/ember.js:64:14)
    at internalRequire (http://localhost:13141/tests/ember.js:60:22)
    at internalRequire (http://localhost:13141/tests/ember.js:60:22)
    at internalRequire (http://localhost:13141/tests/ember.js:60:22)
    at internalRequire (http://localhost:13141/tests/ember.js:60:22)
    at internalRequire (http://localhost:13141/tests/ember.js:60:22)
    at require (http://localhost:13141/tests/ember.js:97:14)
    at http://localhost:13141/ember-template-compiler.js:17947:92", source:  (52)
```

rwjblue:
FYI - These changes will need to be backported for 3.16 (affected LTS), 3.17 (current release), and 3.18 (current beta).

locks (approved):

pzuraq (approved):

Turbo87:
to avoid such issues in the future would it make sense for use to switch to `~` constraints for Babel?

rwjblue:
> to avoid such issues in the future would it make sense for use to switch to ~ constraints for Babel?


@Turbo87 - It would not have made a difference in this case. The drift is from ember-cli-babel itself, I suppose we could pin it with `~` instead of `^` (on `@babel/preset-env`), but that also seems pretty bad. The real issue is what I described above: we shouldn't "custom override" the babel helpers like this.

rwjblue:
FYI - I've just queued up the CI jobs that will publish 3.17.1 and 3.16.4 (the next beta should be published tomorrow).

To follow along:

* 3.17.1 will be published by https://travis-ci.org/github/emberjs/ember.js/builds/665660999
* 3.16.4 will be published by https://travis-ci.org/github/emberjs/ember.js/builds/665668537

If either of these are red, it means the release didn't publish so please ping me and let me know!

rwjblue:
v3.17.1 is published with this fix (should resolve things for folks generating new apps), but I made a mistake when pulling back these changes for 3.16 (just forgot to include the commits when cherry-picking other fixes 🤦‍♂) and v3.16.4 does **not** include this fix.

I have kicked off a publishing of v3.16.5 (definitely with these changes) and it is working its way through CI now:

https://travis-ci.org/github/emberjs/ember.js/builds/665843150

jherdman:
@rwjblue by any chance has this been upstreamed to beta yet?

rwjblue:
@jherdman - It is in `beta` branch, but a beta hasn't been tagged / released yet. Should happen today.

jherdman:
Much, love! Thank you so much for all that you do.

poplevente:
@rwjblue do you know when will be published to npm?

rwjblue:
@poplevente - v3.18.0-beta.3 was published yesterday (roughly 17 hours ago).

poplevente:
@rwjblue i don't see that on npmjs

rwjblue:
@poplevente - It is there 😸 https://www.npmjs.com/package/ember-source/v/3.18.0-beta.3

poplevente:
@rwjblue i mean update ember-cli
on our build machine we install ember-cli as a global and that fails now with the error from this ticket

poplevente:
@rwjblue our problem is that we have a project which was running with ember-source@3.13
worked well before a few days ago. now we receive the _createSuper error. if we go to 3.17.1 it works, but we receive errors and warning with deprecation messages.
is there a version we can go back which was similar to the old 3.13? i've tried 3.12.3, and it works, but will it have any impact? or we better do an updage to 3.17.1 and do the deprecation? (which is time and money :( )

rwjblue:
3.13 is unsupported (contains loads of bugs fixed in more recent versions), 3.12 works well and is a supported LTS that is unaffected by the issue this PR fixes (though it does not contain all of the octane features, e.g. tracked).

As far as I see it, you have a few options:

* upgrade to supported Ember versions (either 3.16 which is an LTS or 3.17 which is the current release)
* downgrade to 3.12 (supported for security fixes for a bit longer)
* fix your applications lock file such that you are not picking up Babel 7.9

poplevente:
@rwjblue awesome thanks!

packages/external-helpers/lib/external-helpers.js
```diff
@@ -117,3 +122,74 @@ export function objectDestructuringEmpty(obj) {
...
+    arr2[i] = arr[i];
+  }
+
+  return arr2;
+}
+
+/*
+  Does not differ from default implementation.
+*/
+function unsupportedIterableToArray(o, minLen) {
```
DLTKPranjalGupta, line 167: @rwjblue any reason why these 2 functions are not exported?

I am upgrading ember version for few of my old repos and CI fails with `TypeError: (0 , _emberBabel.arrayLikeToArray) is not a function`.
We added a patch to export these functions, and it resolved the issue.

kategengler: These were not exported because they were never intended to be public. 

This entire file is gone on main since https://github.com/emberjs/ember.js/pull/20873 and unused even longer. 

What version of Ember did you upgrade to? I suggest you check https://github.com/ember-cli/ember-new-output to make sure your package.json has updated all the required packages.

PR #18835, opened 2020-03-24
[BUGFIX] Make ArrayProxy Lazy

pzuraq:
This PR refactors ArrayProxy to only begin observing the proxied array's
changes the first time the proxy is actually used. This fixes a number
of bugs that have popped up recently, and also ensures that ArrayProxy
creation does not accidentally entangle with the autotracking stack. It
also simplifies the code overall, and conceptually is more in line with
the way that autotracking works.

In addition, it adds a CUSTOM_TAG_FOR to the proxy, to forward the `[]`
and `length` tags from its content directly rather than attempting to
managed them itself. This keeps the system from encountering issues
related to tag updates, and from needing to re-notify when changes are
propagated.

As part of fixing this, I also discovered that the `hasArrayObservers`
API has been broken since ~3.11. I fixed it and added a test.

Original PR: https://github.com/emberjs/ember.js/pull/18770

### DIFF
--- a/packages/@ember/-internals/runtime/lib/system/array_proxy.js
+++ b/packages/@ember/-internals/runtime/lib/system/array_proxy.js
@@ -11,14 +11,14 @@ import {
   removeArrayObserver,
   replace,
   getChainTagsForKey,
-  tagForProperty,
+  CUSTOM_TAG_FOR,
   arrayContentDidChange,
-  arrayContentWillChange,
+  createTagForProperty,
 } from '@ember/-internals/metal';
 import EmberObject from './object';
 import { isArray, MutableArray } from '../mixins/array';
 import { assert } from '@ember/debug';
-import { combine, update, validate, value } from '@glimmer/reference';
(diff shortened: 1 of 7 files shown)


rwjblue:
Can you cross link the original PR?

packages/@ember/-internals/runtime/lib/system/array_proxy.js
```diff
@@ -11,14 +11,14 @@ import {
   removeArrayObserver,
   replace,
   getChainTagsForKey,
-  tagForProperty,
+  CUSTOM_TAG_FOR,
```
rwjblue: When did this API land? I just want to make sure it existed in 3.16...

rwjblue (approved):

packages/@ember/-internals/runtime/lib/system/array_proxy.js
pzuraq: it landed with one of the more recent bugfixes, and I believe was backported.

PR #18837, opened 2020-03-24
[BUGFIX] Fix `willDestroy` on class helpers

### LINKED ISSUE #18830: Class-based helpers' willDestroy() hook called multiple times in tests
In Ember 3.17, class-based helpers' `willDestroy()` hooks are being called twice from acceptance tests.

See [this twiddle](https://ember-twiddle.com/212dc178f1da9dc03ad270fbec3ea60f?openFiles=helpers.my-helper%5C.js%2C) for a repro as well as stack traces to the two calls.

This does not happen if you change the Ember.js version to 3.16, so it's new in 3.17. This is causing problems in my helpers that do things like manage the lifetime of `@ember/object/events` listeners (it double-removes and the second removal throws the listener-not-registered assertion).

pzuraq:
Class based helpers were having `willDestroy` be called twice, once by
the Glimmer VM's built-in destroyables implementation, and once
scheduled by `destroy` (which is the correct destruction). This is
because we were registering the class itself directly as the
destroyable, and Glimmer VM detected the `willDestroy` function on it
and called it eagerly.

This fix creates an intermediate destructor object with a `destroy()`
hook, but long term we need to finish the refactors to destroyables in
the VM and make it so these types of collisions can't happen again.

Added a basic lifecycle test for helpers to prevent this specific issue
from reoccuring.

Fixes #18830

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/helper.ts
+++ b/packages/@ember/-internals/glimmer/lib/helper.ts
@@ -32,14 +32,14 @@ export interface SimpleHelper<T = unknown> {
 
 export function isHelperFactory(
   helper: any | undefined | null
-): helper is SimpleHelperFactory | ClassHelperFactory {
+): helper is Factory<SimpleHelper | HelperInstance, HelperFactory<SimpleHelper | HelperInstance>> {
   return (
     typeof helper === 'object' && helper !== null && helper.class && helper.class.isHelperFactory
   );
 }
 
-export function isSimpleHelper(helper: SimpleHelper | HelperInstance): helper is SimpleHelper {
-  return (helper as any).destroy === undefined;
+export function isClassHelper(helper: SimpleHelper | HelperInstance): helper is HelperInstance {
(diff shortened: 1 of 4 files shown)


rwjblue:
Can you update this to `[BUGFIX release]`?

yarn.lock
```diff
@@ -2,7 +2,7 @@
 # yarn lockfile v1
 
 
-"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.5.5", "@babel/code-frame@^7.8.3":
+"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.8.3":
```
rwjblue: Can you reset this files changes? Seems unrelated to the other changes here...

pzuraq: FWIW I think this just happened when I ran `yarn` locally`, unsure. But yes, will reset

rwjblue (approved):

PR #18852, opened 2020-03-31
Add a simple (failing) test for named block usage

dfreeman:
Based on a conversation in Discord with @rwjblue and @locks yesterday, it sounds like the level of effort needed to get named blocks working in Ember at this point is a known-unknown.

Just to get the ball rolling, I pulled together a simple test gated on the `EMBER_NAMED_BLOCKS` feature. What I found was that basic usage appears to work for template-only components and the `GlimmerishComponent` test backing class, but not with Ember components.

Unfortunately my knowledge of the VM is minimal enough that I don't have much of a clue _why_ that's the case, but hopefully it's at least a useful data point for someone who knows more than me 😄


rwjblue:
Should pass once https://github.com/emberjs/ember.js/pull/18869 is landed and this is rebased.

rwjblue:
OK, ready for a rebase.

locks:
Rebased!

PR #18853, opened 2020-03-31
Add a failing test case for 3.16 regression relating to array mutation and `didReceiveAttrs`

brendenpalmer:
We're updating our application from 3.12 -> 3.16 and discovered this bug where `didReceiveAttrs` is called from changes that originated within the component itself.

After debugging this issue with @pzuraq, he suggested opening a PR with a failing test case. We _may_ have a workaround for this regression in our application, but we figured it would be good to at least document this regression somewhere.

Note that I confirmed that this test case passes on `lts-3-12`.

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/components/curly-components-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/components/curly-components-test.js
@@ -1496,6 +1496,42 @@ moduleFor(
       this.assertText('In layout - someProp: wycats');
     }
 
+    ['@test mutating an array from within component should not trigger `didReceiveAttrs`'](assert) {
+      let didReceiveAttrsCount = 0;
+
+      this.registerComponent('non-block', {
+        ComponentClass: Component.extend({
+          didReceiveAttrs() {
+            this._super(...arguments);
+            didReceiveAttrsCount++;
+          },
+
+          actions: {
+            click() {
+              this.model = this.model.replace(0, 1, ['foo']);
+            },
+          },
+        }),
+        template: `
+          <button {{action "click"}}>foobar</button>
+          <br>
+          <ul>
+            {{#each @model as |item|}}
+              <li>{{item}}</li>
+            {{/each}}
+          </ul>
+        `,
+      });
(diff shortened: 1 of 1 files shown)


rwjblue (commented):
This doesn't seem like a regression to me 🤔. I definitely see what you mean (it does seem odd to call `didReceiveAttrs` again), but the public API for `didReceiveAttrs` has always been that we guarantee that it will be called when anything changes but **not** exactly the number of times called (this is roughly the same thing as with observers).

Can you add the same test with an object, I think we should generally match the behavior of that here. 

<details>
<summary>Thinking something like this.</summary>

```js
    ['@test mutating an array from within component should not trigger `didReceiveAttrs`'](assert) {
      let didReceiveAttrsCount = 0;

      this.registerComponent('non-block', {
          didReceiveAttrs() {
            this._super(...arguments);
            didReceiveAttrsCount++;
          },

          actions: {
            click() {
              this.model = this.set('model.foo', 'baz');
            },
          },
        }),
        template: `
          <button {{action "click"}}>Click Me!</button>
          <br>
          {{@model.foo}}
        `,
      });

      this.render('{{non-block model=model}}', {
        model: { foo: "bar" },
      });

      runTask(() => this.$('button').click());

      assert.equal(didReceiveAttrsCount, 1, '`didReceiveAttrs` only called once');
    }
```
</details>

brendenpalmer:
I can add another test case for an object, but I believe for an object `didReceiveAttrs` is only called once when the change is from within the component. I might be mis-interpreting the docs here, but they do say that `didReceiveAttrs` shouldn't be called for re-renders that are initiated internally.

chriskrycho:
@rwjblue reference [here](https://guides.emberjs.com/v3.12.0/components/the-component-lifecycle/#toc_formatting-component-attributes-with-didreceiveattrs).

kategengler:
This bug has been around so long that I feel we are unlikely to ever fix it, especially as `didReceiveAttrs` is not part of Octane-style Ember, let alone Polaris, so for that reason I am going to close it.

PR #18854, opened 2020-04-01
Pass value through to `PROPERTY_DID_CHANGE` to avoid calling `get` when setting values for computed props

brendenpalmer:
As part of trying to upgrade our application from 3.12 -> 3.16, I discovered a regression where setting a value for a computed property, then immediately retrieving that value would return `undefined`. In our test suite and application this causes subtle bugs and unexpected behavior. With help from @pzuraq, we were able to trace this down to `PROPERTY_DID_CHANGE` firing at the wrong time, specifically for computed properties.

As part of this I also removed an unused `setWithSuspend` method.

Note that I confirmed that this test case passes on `lts-3-12`.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/component.ts
+++ b/packages/@ember/-internals/glimmer/lib/component.ts
@@ -778,7 +778,7 @@ const Component = CoreView.extend(
       this._super();
     },
 
-    [PROPERTY_DID_CHANGE](key: string) {
+    [PROPERTY_DID_CHANGE](key: string, value?: unknown) {
       if (this[IS_DISPATCHING_ATTRS]) {
         return;
       }
@@ -787,7 +787,7 @@ const Component = CoreView.extend(
       let reference = args !== undefined ? args[key] : undefined;
 
       if (reference !== undefined && reference[UPDATE] !== undefined) {
-        reference[UPDATE](get(this, key));
+        reference[UPDATE](arguments.length === 2 ? value : get(this, key));
       }
     },
 
--- a/packages/@ember/-internals/metal/lib/property_events.ts
+++ b/packages/@ember/-internals/metal/lib/property_events.ts
@@ -24,12 +24,18 @@ let deferred = 0;
   @for @ember/object
   @param {Object} obj The object with the property that will change
   @param {String} keyName The property key (or path) that will change.
-  @param {Meta} meta The objects meta.
+  @param {Meta} [_meta] The objects meta.
+  @param {unknown} [value] The new value to set for the property
   @return {void}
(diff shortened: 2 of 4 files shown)


rwjblue:
FYI - Linting is failing:

```
$ yarn lint:tsc && yarn lint:tslint && yarn lint:eslint && yarn lint:docs
$ tsc --noEmit
packages/@ember/-internals/metal/lib/computed.ts(261,11): error TS6133: '_suspended' is declared but its value is never read.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

packages/@ember/-internals/metal/lib/property_events.ts
```diff
@@ -24,12 +24,18 @@ let deferred = 0;
...
   @return {void}
   @since 3.1.0
   @public
 */
-function notifyPropertyChange(obj: object, keyName: string, _meta?: Meta | null): void {
+function notifyPropertyChange(
+  obj: object,
+  keyName: string,
+  _meta?: Meta | null,
+  _shouldFirePropertyDidChange = true
```
pzuraq, line 37: Discussed this with @rwjblue this morning, and we realized a better solution here would be to pass the value optionally as the last argument to `notifyPropertyChange` here. This way, `PROPERTY_DID_CHANGE` can receive the value for this use case, and otherwise lookup the value if its undefined, and we don't have to call `PROPERTY_DID_CHANGE` again later on and copy all that logic.

brendenpalmer: Thanks for the review! I just updated the PR to pass the `value` through to `notifyPropertyChange`, then to `PROPERTY_DID_CHANGE` to avoid calling `get`.

packages/@ember/-internals/glimmer/lib/component.ts
```diff
@@ -787,7 +787,7 @@ const Component = CoreView.extend(
       let reference = args !== undefined ? args[key] : undefined;
 
       if (reference !== undefined && reference[UPDATE] !== undefined) {
-        reference[UPDATE](get(this, key));
+        reference[UPDATE](value === undefined ? get(this, key) : value);
```
pzuraq, line 790:
```suggestion
        reference[UPDATE](arguments.length === 2 ? value : get(this, key));
```

pzuraq (approved):

pzuraq:
Thanks a ton @brendenpalmer! If you could cherry-pick this onto master as well that'd be super helpful!

PR #18857, opened 2020-04-01
Pass value through to `PROPERTY_DID_CHANGE` to avoid calling `get` when setting values for computed props

brendenpalmer:
This is just a cherry-pick of the fix from this PR onto `master`: https://github.com/emberjs/ember.js/pull/18854


pzuraq (approved):

pzuraq:
Thanks again!

PR #18859, opened 2020-04-03
[BUGFIX] Fix transitionTo with scoped aliased queryParam.

ttill:
When the router service scopes query params passed to `transitionTo`, this would raise an assertion when a query param has an alias.

Side note: There is some relevant discussion in #18244 on the complexity of `_hydrateUnsuppliedQueryParams`.


auvipy (approved):

rwjblue (approved):
Thank you @ttill! Sorry for the delay here.

PR #18861, opened 2020-04-03
[BUGFIX release] Update glimmer-vm to 0.50.1 to improve dev time performance.

### LINKED ISSUE #18798: Chrome slowness
I've created a [simple app](https://github.com/amk221/select-box-test) that renders 500 glimmer components in a loop. Mousing over one sets it as active.

Here is a quick video demoing it: https://www.youtube.com/watch?v=ySuUKGfpoGw

There are 3 parts to the video:

* development environment with the dev tools open.
* development environment
* production environment

As you can see, with the dev tools open, it's barely usable (and this makes day to day development very frustrating!)

I understand that there are assertions stripped in production builds, which is why it is faster. But, I do not expect to see such a _large_ difference with the dev tools open.

### LINKED ISSUE #18834: Rendering perfomance issue with each and tables
This is extended from a conversation on Discord. 

I discovered a significant rendering performance issue in my app.  After refactoring a table I was seeing a new significant delay after a click when rendering some table rows.

@pzuraq Asked if I could try with a production build. The performance was significantly better than the development build. He thought perhaps it was another instance of a known issue with "Error slowness"

I have created a simple reproduction of the error here -> https://github.com/dgavey/ember-table-rendering-performance . There is a checkbox that either renders the columns of the table using an each for the column display, or skips the each and just renders every column. 

If the checkbox is on, the initial rendering of the table is quite a bit slower than it is when it's off.  This was evaluated in a development build.

In a production build, the render time between options seems to be much closer. 

In this reproduction the render difference does not seem to be any where near as dramatic as it is in my app, but so far this is the closest I could come to finding where the performance difference was while keeping the reproduction simple.

rwjblue:
Removes capturing stack traces for all tracked items, as just calling
`new Error()` ends up taking quite a bit of time.

Fixes #18834
Fixes #18798


bendemboski:
[here](https://github.com/rwjblue/ember.js/pull/2)'s a PR fixing the tests to help shepherd this along 😃

rwjblue:
Thanks @bendemboski!

locks (approved):

PR #18867, opened 2020-04-10
[FEATURE] Add support for public `in-element`

simonihmig:
As per [RFC287](https://emberjs.github.io/rfcs/0287-promote-in-element-to-public-api.html).
* ~This does not yet add API docs, will follow in a separate PR~
* The removal of the hereby deprecated `{{-in-element}}` is slated for 4.0. The RFC suggested the next version after the next LTS, but given that the deprecation will be added at 3.19 at the earliest, I would think 4.0 seems appropriate (depending on how many more 3.x releases are expected)


locks:
Thank you for the help! I don't think this is wrapped in a feature flag, so I would like to see API documentation be merged with the implementation so we make sure any release with the feature also contains the documentation. I can help with it!

simonihmig:
@locks added a commit with the API docs!

acorncom (approved):

simonihmig:
Travis build itself is [green](https://travis-ci.org/github/emberjs/ember.js/builds/673565708), but still shown here as pending. 🤔

pzuraq (approved):
This looks good to me! Thanks for getting this done 😄

pzuraq:
Ah, one thing actually, could you add a canary feature flag that guards this? Shouldn't be too hard, it should probably just re-enable the transform that errors for `{{in-element}}`

simonihmig:
> Ah, one thing actually, could you add a canary feature flag that guards this? 

@pzuraq done. Hope I got the usage of feature flags right, please re-review!

packages/@ember/canary-features/index.ts
```diff
@@ -20,6 +20,7 @@ export const DEFAULT_FEATURES = {
   EMBER_CUSTOM_COMPONENT_ARG_PROXY: true,
   EMBER_GLIMMER_SET_COMPONENT_TEMPLATE: true,
   EMBER_ROUTING_MODEL_ARG: true,
+  EMBER_GLIMMER_IN_ELEMENT: null,
```
pzuraq, line 23: I think it's _probably_ fine to enable by default in this PR too, since it's not a major refactor. The feature flag is mostly to give us a way to turn it off in case we do run into major issues during `beta`. Will confirm though

simonihmig: Ah right. Now that you are mentioning this, I remember that I wanted to ask about this 😝

simonihmig:
Travis seems to undergo maintenance, the build is not showing up here so far, but it *is*  running [here](https://travis-ci.org/github/emberjs/ember.js/builds/673863089)

pzuraq (approved):

packages/@ember/canary-features/index.ts
pzuraq: Sorry, seems I was mistaken. Assuming travis passes I think this is good to go!

pzuraq:
Thanks again @simonihmig!

PR #18869, opened 2020-04-13
[BUGFIX release] Update to glimmer-vm 0.50.2.

rwjblue:
* Fixes an issue with nested tracking frames getting out of balance.
* Fixes an issue with the upcoming named block feature with classic components.


pzuraq:
I think this should be merged back to LTS, not entirely sure but I think 0.50 is on the LTS release

rwjblue:
Added the same tests in https://github.com/emberjs/ember.js/pull/18872. Tests pass locally, and should land just to add confidence there.

PR #18871, opened 2020-04-13
[BUGFIX beta] Make assertions from `on` and `fn` more actionable.

rwjblue:
Prior to these changes, the assertions for `fn` and `on` were very inactionable. This leverages the debug render tree information to print the template contexts (route hierarchy, component invocation heirarchy, etc) along with the path name used for the callback argument.

This is essentially a rewrite of the work started by @patricklx in #18353 (since that PR had to be reverted due to changes in glimmer-vm internals since it was proposed).

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/helpers/fn.ts
+++ b/packages/@ember/-internals/glimmer/lib/helpers/fn.ts
@@ -1,6 +1,6 @@
 import { assert } from '@ember/debug';
 import { DEBUG } from '@glimmer/env';
-import { CapturedArguments, VM, VMArguments } from '@glimmer/interfaces';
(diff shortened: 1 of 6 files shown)


patricklx:
nice to finally see this :)
but i wonder if there isn't a more generic approach to apply logRenderStackForPath to all errors generated during template render/update and include all passed params in the error as well. for all modifiers/helpers etc...

pzuraq (approved):

rwjblue:
> I wonder if there isn't a more generic approach to apply logRenderStackForPath to all errors generated during template render/update and include all passed params in the error as well. for all modifiers/helpers etc...

Ya, I totally agree. I'll think on it, not sure exactly where that fits.

PR #18882, opened 2020-04-18
[BUGFIX] isEmpty on nested objects

### LINKED ISSUE #16878: Calling isEmpty on instances of ObjectProxy wrapping ArrayProxy throws Assertion Failed: You attempted to access the `length` property

locks:
Because of a re-structuring, it was too much work rebasing or even cherry-picking https://github.com/emberjs/ember.js/pull/16879 (over 300 git conflicts ;_;) so I reimplemented the fix, sorry @janvotava.

Fixes https://github.com/emberjs/ember.js/pull/16879.
Fixes https://github.com/emberjs/ember.js/issues/16878.

### DIFF
--- a/packages/@ember/-internals/metal/lib/is_empty.ts
+++ b/packages/@ember/-internals/metal/lib/is_empty.ts
@@ -54,15 +54,15 @@ export default function isEmpty(obj: any): boolean {
     }
   }
 
-  if (typeof obj.length === 'number' && objectType !== 'function') {
-    return !obj.length;
-  }
-
   if (objectType === 'object') {
     let length = get(obj, 'length');
     if (typeof length === 'number') {
       return !length;
     }
+  }
+
+  if (typeof obj.length === 'number' && objectType !== 'function') {
+    return !obj.length;
   }
 
   return false;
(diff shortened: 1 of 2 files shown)


packages/@ember/-internals/metal/lib/is_empty.ts
```diff
@@ -53,17 +53,17 @@ export default function isEmpty(obj: any): boolean {
...
-
+  
   if (objectType === 'object') {
     let length = get(obj, 'length');
     if (typeof length === 'number') {
       return !length;
     }
   }
+  
+  if (typeof obj.length === 'number' && objectType !== 'function') {
```
locks, line 64: @rwjblue you had a comment about this check, but removing it made tests fail so I didn't touch it.

rwjblue: I just don't grok how this can ever be hit. The check **just above** here is checking all `"object"`s that have a numeric `length` property.

lifeart: @rwjblue it has different `objectType` (object vs function)

lifeart: looks like it's true for string case `typeof 'foo'.length === 'number' && typeof 'foo' !== 'function'`

PR #18904, opened 2020-04-21
[RFC#236] Deprecate Ember.String/(at)ember/string

locks:
## Tasks

The following tasks should have their own deprecation ids, IMO.
At first I was trying to put it all in the same deprecation guide, but it soon turned out to be too confusing.

- [ ] Deprecate `String.prototype` extensions


NullVoxPopuli:
Continuation: https://github.com/emberjs/ember.js/pull/19084

PR #18909, opened 2020-04-21
Revert dependabot version bumps that didn't go through CI

locks:
This reverts commit 6438e9c1ee465b3368789ecd07db7c52fcc6fce7.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "ember-cli-normalize-entity-name": "^1.0.0",
     "ember-cli-path-utils": "^1.0.0",
     "ember-cli-string-utils": "^1.1.0",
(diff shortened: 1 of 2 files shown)


kategengler (approved):

kategengler:
Will dependabot reopen these updates after we do this?

locks:
I _think_ so?

rwjblue:
Yes, I was thinking that it would reopen them also. Basically, we don't know which bump caused the fundamental issue, so having dependabot reopen them makes work that out much easier (IMHO).

locks:
I will trigger it from the dasboard if it doesn't soon!

PR #18919, opened 2020-04-23
[BUGFIX beta] Add error for modifier manager without capabilities.

rwjblue:
Now that the deprecation is removed (in 3.19.0-beta.1) folks that were previously getting the deprecation would get a very bizarre error:

```
Cannot read property 'disableAutoTracking' of undefined
```

This adds a much more helpful assertion.

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/custom-modifier-manager-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/custom-modifier-manager-test.js
@@ -36,6 +36,24 @@ class CustomModifierManager {
 moduleFor(
   'Basic Custom Modifier Manager',
   class extends ModifierManagerTest {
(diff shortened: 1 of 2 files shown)


pzuraq (approved):

packages/@ember/-internals/glimmer/tests/integration/custom-modifier-manager-test.js
```diff
@@ -36,6 +36,24 @@ class CustomModifierManager {
 moduleFor(
   'Basic Custom Modifier Manager',
   class extends ModifierManagerTest {
+    '@test throws a useful error when missing capabilities'(assert) {
```
pzuraq, line 39:
```suggestion
    '@test throws a useful error when missing capabilities'() {
```

PR #18928, opened 2020-04-24
Update deprecation for `{{-in-element}}` to be until: 3.25.0.

rwjblue:
We can always delay removal until the community has fully migrated, but having the deprecation be through two LTS (master will be 3.20, then another LTS in 3.24) is more than enough for a private API.

/cc @simonihmig @chancancode


chancancode (approved):

simonihmig (approved):
I wasn't sure we are really going to see so many minor releases in the 3.x cycle, but if that's the case this seems reasonable!

rwjblue:
@simonihmig yeah, honestly I don't think we have any clue how many we will have. If 4.0 ends up happening first that seems totally fine to me too...

pzuraq (approved):

Issue #18460, opened 2019-10-03
[Docs] CONTRIBUTING.md Reporting A Bug outdated


Alonski:
Here: https://github.com/emberjs/ember.js/blob/master/CONTRIBUTING.md#reporting-a-bug
Number 3: States that a Twiddle or JSBin should be provided to showcase the bug.

This should be updated as those aren't necessarily the best ways to showcase a bug.

Possible solutions:
- Remove JSBin and add CodeSandbox.
- Remove Twiddle and JSBin and suggest users create a demo repository on GitHub.

pzuraq:
I agree that these seem like the best things folks can do at the moment, since twiddle doesn’t work with the most recent versions of Ember

rwjblue:
I'd personally prefer to suggest "making a demo repo on GitHub". I wish CodeSandbox was a good enough suggestion, but I just don't find it stable enough to be able to use it to quickly dig into a reported issue.

mehulkar:
I’ve been adding `ember-possible-bug` to my repro repos! Might be useful to recommend? https://github.com/topics/ember-possible-bug

alexeykostevich:
You've raised a very good issue, @Alonski !

Currently, [Contributing](https://github.com/emberjs/ember.js/blob/master/CONTRIBUTING.md) docs mention 2 tools:
- [ember-twiddle](http://ember-twiddle.com), which supports multiple versions of Ember.js and Ember CLI.
- [JS Bin](http://emberjs.jsbin.com/?html,js,output), which uses pure Ember.js 1.x.

To cover the whole range of possible issues, it would be ideal if a reproduction environment:
- supported Ember CLI projects to allow to reproduce issues in the framework and build tools
- worked online and didn't require a multi-step setup
- supported external dependencies to reproduce integration issues and issues in community's libraries
- provided a familiar development experience to let people with different backgrounds (e.g. Angular) to start quickly

Unfortunately, it seems there is no ideal solution right now. The following tools could be a good solutions but need some investment from the community:
- [StackBlitz](https://stackblitz.com/) — an online IDE with JS/TS, Angular and React support. There is an issue (https://github.com/stackblitz/core/issues/647) for Ember.js support but there is no progress on this so far.
- [CodeSandbox](https://codesandbox.io/) — an online IDE with a wide range of supported frameworks (including Ember.js) but it looks a bit unstable at this point.

[...]

Issue #18906, opened 2020-04-21
Create an Issue Template for the project


alexeykostevich:
According to the project's [Community Profile](https://github.com/emberjs/ember.js/community), the repository setup can be improved to match the recommended community standards. 

<img width="756" alt="Screen Shot 2020-04-21 at 11 51 57 pm" src="https://user-images.githubusercontent.com/5205189/79873966-235d5200-842b-11ea-8ec6-45ab8ad4d6ca.png">

One of the valuable improvements can be creating an [Issue Template](https://help.github.com/en/github/building-a-strong-community/using-templates-to-encourage-useful-issues-and-pull-requests) to encourage useful issues, simplify reporting and subsequent investigation.

Here are some examples from other repositories:
- [microsoft/TypeScript](https://github.com/microsoft/TypeScript/issues/new/choose)
- [angular/angular](https://github.com/angular/angular/issues/new/choose)
- [aurelia/aurelia](https://github.com/aurelia/aurelia/issues/new/choose)
- [telerik/kendo-angular](https://github.com/telerik/kendo-angular/issues/new/choose)

The template can be potentially used across the whole organization (e.g. for Ember Data and Ember CLI, etc.).

alexeykostevich:
Here is a draft version of a bug template based on the examples above and Ember's [CONTRIBUTING - Reporting a Bug](https://help.github.com/en/github/building-a-strong-community/using-templates-to-encourage-useful-issues-and-pull-requests) doc.

```
<!---
Thank you for filing your contribution! Before you create a new issue, please search for similar issues. It's possible somebody has encountered this bug already.
-->

## 🐛 Issue Description


## 🤔 Expected Behavior


## 😕 Current Behavior


## 🔬 Minimal Reproduction
<!--
Please, provide minimal reproduction of the issue using one of the following:
- Ember Twiddle: http://ember-twiddle.com/
- JSBin: http://emberjs.jsbin.com/
- Create a GitHub repository: https://guides.emberjs.com/release/getting-started/quick-start/
-->


## 🌍 Environment

- Ember version: -
- OS: -
- Browser: -
- NPM/Node/Yarn: -
```

rwjblue:
Seems like a great suggestion! Mind sending a PR?

alexeykostevich:
Thank you, @rwjblue ! I'll create a PR today/tomorrow.

alexeykostevich:
@rwjblue , I've created a PR. Please, let me know if you have any concerns.

[...]

PR #18930, opened 2020-04-26
Remove JSBin from the list of recommended issue reproduction tools.

alexeykostevich:
Removes [JSBin](http://emberjs.jsbin.com) from the list of recommended issue reproduction tools since Ember does not support the `<script>` tag anymore.

Addresses https://github.com/emberjs/ember.js/issues/18460.


alexeykostevich:
@locks , here is the change we've agreed to make while discussing https://github.com/emberjs/ember.js/issues/18906.

locks (approved):
Thank you!

Issue #18796, opened 2020-03-05
Asset size regression in v3.17.0


Turbo87:
We noticed in one of our apps that the Ember v3.17.0 release significantly increase our asset sizes:

## Before

 - dist/assets/app.js: 613.18 KB (78.46 KB gzipped)
 - dist/assets/vendor.js: 2.34 MB (558.13 KB gzipped)

## After

 - dist/assets/app.js: 690.43 KB (89.22 KB gzipped)
 - dist/assets/vendor.js: 2.42 MB (572.39 KB gzipped)

/cc @pzuraq @rwjblue

makepanic:
Can confirm in a fresh new ember project:

__2.16.0__

* vendor-79a1173ef882e83086e2dfed38dbbf24.js: 692.58 KB (176.09 KB gzipped)
* size-regression-ef4e4e07dcbd678f6b97ceb963991481.js: 9.82 KB (2.15 KB gzipped)

__2.17.0__

* vendor-75d1bae5b6f40054c0b623620914b48a.js: 715.12 KB (183.64 KB gzipped)
* size-regression-bca0d5e922df97037f2fa0bbe7f1fb56.js: 9.94 KB (2.2 KB gzipped)

Did some vendor diffs and it seems to be caused by some glimmer-vm updates, e.g. https://github.com/glimmerjs/glimmer-vm/commit/4c56c216465410f5bd6f4f53fb7d4508f8048f09#diff-64c3cabffb164d9a2c4bf2d427094e19

boris-petrov:
Same here.

**2.16.3**
- vendor: 886KB (213KB brotli)
- app: 1.01MB (126KB brotli)

**2.17.0**
- vendor: 907KB (219KB brotli)
- app: 1.15MB (137KB brotli)

Also, I noticed a ~5% performance degradation in our integration tests going from 2.16.3 to 2.17.0.

rwjblue:
@boris-petrov - Mind checking that for us? I think you should be able to remove the minifier from your apps `package.json`...

krisselden:
Looks like this is coming from changes in the glimmer wireformat that is bloating the template sizes a bit. In particular doing a diff on the app is https://github.com/glimmerjs/glimmer-vm/blob/11cc83b2adf220bef298a1a6298c9e4e750c9fe1/packages/%40glimmer/interfaces/lib/compile/wire-format.d.ts#L76

boris-petrov:
[...]

PR #18941, opened 2020-05-01
[BUGFIX release] Update to glimmer-vm packages to 0.52.0

### LINKED ISSUE #18855: [3.17.2] WillDestroy is called at a wrong time for components inside {{each}}
If a component is rendered in {{each}}  block for items added to the list after initial rendering then
for those components `willDestroy()` is called only after `{{each}}` block is removed.

This issue is very similar to #18797 and it may be a result of the fix for that issue.

### Why is this a problem
Our application uses Ember Leaflet to display a map with markers. Markers are displayed using `{{each}}` block with Ember Leaflet `<Marker>` components inside. The component relies on life-cycle hooks to actually add and remove markers using leaflet.js library.

Not calling `willDestroy` results in markers staying on a map after they were removed from a list.

### Reproduction Scenario
I have created an Ember Twiddle to illustrate the problem:
https://ember-twiddle.com/db073f2618c1f191f75ce249f89cb639

It creates a list with 2 elements and have buttons to add and remove elements. It removes whole `{{each}}` block if the list is empty. It also logs component construction and `willDestroy()` calls.

- Start with  items 1 and 2 in the list
- Add item 3
- Remove item 1:  willDestroy(1) is called
- Remove item 3:  **PROBLEM:** willDestroy(3) is not called.
- Remove item 2: willDestroy(list) is called followed by willDestroy(1) and ** willDestroy(3) (delayed)**
- Add item 4: the list get rendered again
- Add item 5:
- Remove item 5:  **PROBLEM:** willDestroy(5) is not called.
- Remove item 4: willDestroy(list) is called followed by willDestroy(4) and ** willDestroy(5 (delayed))**

If the Twiddle is configured for Ember 3.17.0, then those problematic `willDestroy()` are not called at all. Issue #18797

If the Twiddle is configured for Ember 3.16.2, then those problematic `willDestroy()` are called as expected right after element removal from the list

### LINKED ISSUE #18864: [3.17] [REGRESSION] re-rendering an `{{#each}}` with multiple null values only renders one iteration
When updating an `{{#each}}`-rendered array to be `[ null, null, null ]`, only one iteration of the array is rendered, as if it just contained `[ null ]`. This rendering test passes in 3.16 and fails in 3.17:

```javascript
  test('rerendering array of nulls works', async function(assert) {
    this.set('items', [ false, false, false ]);
    await render(hbs`
      {{#each this.items as |item|}}
        <p>item: {{item}}</p>
      {{/each}}
    `);
    assert.dom('p').exists({ count: 3 });

    this.set('items', [ null, null, null ]);
    await settled();
    assert.dom('p').exists({ count: 3 });
  });
```

[here](https://github.com/bendemboski/rerender-null-array) is an app that reproduces the issue and demonstrates that it's a regression since 3.16.

rwjblue:
* Reduces template size growth to near 3.16 levels (still ~ 3% higher than 3.16)
* Ensures destroyable items added to an `{{#each` after initial render are invoked properly
* Fixes an issue with key collisions in `{{#each` 

Addressing https://github.com/emberjs/ember.js/issues/18796
Fixes https://github.com/emberjs/ember.js/issues/18864
Fixes https://github.com/emberjs/ember.js/issues/18855


pzuraq (approved):

rwjblue:
@krisselden - Thank you for working on the template size regression!

PR #18960, opened 2020-05-09
[BUGFIX lts] More assertions for Application lifecycle methods

chancancode:
While debugging an issue in the ember-inspector test harness, we eventually we were dealing with very subtle hanging and errors that were ultimately caused by trying to boot an already destroyed `Application`.

The issue was very difficult to debug partly due to some states (like `_bootPromise`) was reset in `willDestroy`, which is not necessary, but was enough to cause other lifecycle methods (like `boot`) to happily restart the process, but just hangs forever later on.

This removes the state reset form `willDestroy` and just adds a lot more assertions in general, hopefully making these kind of situations fail louder and earlier.

### DIFF
--- a/packages/@ember/application/lib/application.js
+++ b/packages/@ember/application/lib/application.js
@@ -387,6 +387,16 @@ const Application = Engine.extend({
     @return {ApplicationInstance} the application instance
   */
   buildInstance(options = {}) {
+    assert(
+      'You cannot build new instances of this application since it has already been destroyed',
+      !this.isDestroyed
+    );
+
+    assert(
+      'You cannot build new instances of this application since it is being destroyed',
+      !this.isDestroying
+    );
+
     options.base = this;
     options.application = this;
     return ApplicationInstance.create(options);
@@ -516,7 +526,7 @@ const Application = Engine.extend({
     @method domReady
   */
   domReady() {
-    if (this.isDestroyed) {
+    if (this.isDestroying || this.isDestroyed) {
       return;
     }
 
@@ -560,10 +570,19 @@ const Application = Engine.extend({
       'You must call deferReadiness on an instance of Application',
       this instanceof Application
     );
+
+    assert('You cannot defer readiness since application has already destroyed', !this.isDestroyed);
+
     assert(
-      'You cannot defer readiness since the `ready()` hook has already been called.',
+      'You cannot defer readiness since the application is being destroyed',
+      !this.isDestroying
+    );
+
+    assert(
+      'You cannot defer readiness since the `ready()` hook has already been called',
       this._readinessDeferrals > 0
     );
+
     this._readinessDeferrals++;
   },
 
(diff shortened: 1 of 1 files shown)


chancancode:
...seems like this broke many things 😅

packages/@ember/application/lib/application.js
```diff
@@ -387,6 +387,16 @@ const Application = Engine.extend({
...
   */
   buildInstance(options = {}) {
+    assert(
+      'You cannot build new instances of this application since it is being destroyed',
+      this.isDestroying
+    );
+
+    assert(
+      'You cannot build new instances of this application since it has already been destroyed',
+      this.isDestroyed
```
rwjblue, line 397: This assertion needs to be moved up above the other one. If we don't, it can never be triggered (since we will never have `this.isDestroying` be false and `this.isDestroyed` be true).

Also, shouldn't the assertion be `!this.isDestroyed` here?

packages/@ember/application/lib/application.js
```diff
@@ -387,6 +387,16 @@ const Application = Engine.extend({
     @return {ApplicationInstance} the application instance
   */
   buildInstance(options = {}) {
+    assert(
+      'You cannot build new instances of this application since it is being destroyed',
+      this.isDestroying
```
rwjblue, line 392: We want to assert that we are not destroying when this method is called , should this be `!this.isDestroying`?

packages/@ember/application/lib/application.js
```diff
@@ -516,7 +526,7 @@ const Application = Engine.extend({
     @method domReady
   */
   domReady() {
-    if (this.isDestroyed) {
+    if (this.isDestroying || this.isDestroyed) {
```
rwjblue, line 529: FWIW, once we go `this.isDestroying = true` we never change it again. I think checking both makes the code easier to read (and avoids everyone having to internalize that particular detail) but I don't think it is actually going to change anything about the semantics.

packages/@ember/application/lib/application.js
```diff
@@ -560,10 +570,19 @@ const Application = Engine.extend({
       'You must call deferReadiness on an instance of Application',
       this instanceof Application
     );
+
+    assert(
+      'You cannot defer readiness since the application is being destroyed',
+      this.isDestroying
```
rwjblue, line 576: !this.isDestroying, right?

rwjblue (changes requested):

packages/@ember/application/lib/application.js
chancancode: 🤦

packages/@ember/application/lib/application.js
chancancode: 🤦

packages/@ember/application/lib/application.js
chancancode: 🤦

rwjblue (approved):

PR #18979, opened 2020-05-18
Reduce dependabot frequency

sandstrom:
Running dependabot too often creates a lot of noise in the commit logs, making it nearly useless to track actual progress.

Running it less often would reduce this problem (security updates still running immediately).

image

*I proposed something similar for Ember Data: https://github.com/emberjs/data/pull/7195*

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,0 +1,13 @@
+# Create Pull-Requests for NPM updates automatically
+# https://dependabot.com/docs/config-file/
+
+version: 1
+update_configs:
+  # Keep package.json (& lockfiles) up to date.
+  #
+  # Security updates will be created immediately,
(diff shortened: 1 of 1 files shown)


rwjblue:
Thanks for the kick here! I’ve also been fairly annoyed with the GH notification noise (IMHO GitHub needs to support filtering not notifications differently from human ones in their new notifications UI).

I’ve modified the frequency to monthly in the dependabot ui for most of the emberjs, ember-cli, ember-codemods, ember-polyfills, and rwjblue org repos. 

I did it in the UI mostly to avoid having to maintain the same config in so many disparate places (and because it was fairly easy to update all of the locations).

sandstrom:
awesome, thanks @rwjblue

PR #18984, opened 2020-05-19
[FEATURE] Adds Cache API

pzuraq:
Adds the cache API specified in RFC 615. Since this API is essentially
just exposing the API that was implemented in Glimmer VM, and tested there,
I don't think tests are necessary to add in Ember itself.


packages/@ember/-internals/metal/lib/cache.ts
```diff
@@ -0,0 +1,128 @@
...
+  getValue(counter); // 2
+  ```
+
+  @method createCache
+  @static
+  @for @glimmer/tracking/primitives/cache
+  @public
+*/
+export function createCache<T>(fn: () => T) {
+  return glimmerCreateCache(fn);
```
rwjblue, line 56: Can you remind me why we can just export this directly? I know we talked about it, but I can’t recall if we had resolved

rwjblue (commented):
Can you add `@category` with the feature flag name to the JSDocs (so the pending features get stripped).

packages/@ember/-internals/metal/lib/cache.ts
pzuraq: ah, I remember we discussed this actually and it would be fine, any additional arguments are not document so it's fine, I'll update

rwjblue (approved):

rwjblue (approved):

PR #18993, opened 2020-05-28
[FEATURE deprecate-get-with-default] Deprecate getWithDefault

chrisrng:
Based on RFC 0554

Also added:
https://github.com/ember-learn/deprecation-app/pull/600

### DIFF
--- a/packages/@ember/-internals/metal/lib/property_get.ts
+++ b/packages/@ember/-internals/metal/lib/property_get.ts
@@ -2,7 +2,7 @@
 @module @ember/object
 */
 import { HAS_NATIVE_PROXY, isEmberArray, isProxy, symbol } from '@ember/-internals/utils';
-import { assert } from '@ember/debug';
+import { assert, deprecate } from '@ember/debug';
 import { DEBUG } from '@glimmer/env';
 import {
   consumeTag,
@@ -181,12 +181,22 @@ export function _getPath<T extends object>(root: T, path: string | string[]): an
   @param {Object} defaultValue The value to return if the property value is undefined
   @return {Object} The property value or the defaultValue.
   @public
+  @deprecated
 */
 export function getWithDefault<T extends object, K extends Extract<keyof T, string>>(
   root: T,
   key: K,
   defaultValue: T[K]
 ): T[K] {
+  deprecate(
+    'Using getWithDefault has been deprecated. Instead, consider using Ember get and explicitly checking for undefined.',
+    false,
+    {
+      id: 'ember-metal.get-with-default',
+      until: '4.0.0',
+    }
+  );
+
   let value = get(root, key);
 
   if (value === undefined) {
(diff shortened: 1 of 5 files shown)


elwayman02:
Thank you! It was confusing seeing the docs not show this as deprecated.

packages/@ember/-internals/metal/lib/property_get.ts
```diff
@@ -181,12 +181,22 @@ export function _getPath<T extends object>(root: T, path: string | string[]): an
...
   key: K,
   defaultValue: T[K]
 ): T[K] {
+  deprecate(
+    'Using getWithDefault has been deprecated. Instead, consider using Ember get and explicitly checking for undefined.',
+    false,
+    {
+      id: 'ember-metal.get-with-default',
+      until: '4.0.0',
+    }
```
rwjblue, line 197: We need to add a `url` property that links to the deprecation guide URL

rwjblue (changes requested):
I think we also need to deprecate `this.getWithDefault` (I suspect this deprecation does fire in that case, but the docs at least need to be updated):

https://github.com/emberjs/ember.js/blob/fc9afa58b7b476ef5ebe86079a08bcd7c1e62008/packages/%40ember/-internals/runtime/lib/mixins/observable.js#L415-L423

packages/@ember/-internals/metal/lib/property_get.ts
chrisrng: Still waiting on this https://github.com/ember-learn/deprecation-app/pull/600 but put the URL that it will generate

chrisrng:
> I think we also need to deprecate `this.getWithDefault` (I suspect this deprecation does fire in that case, but the docs at least need to be updated):
> 
> https://github.com/emberjs/ember.js/blob/fc9afa58b7b476ef5ebe86079a08bcd7c1e62008/packages/%40ember/-internals/runtime/lib/mixins/observable.js#L415-L423

Ah yeah added the doc here!

packages/@ember/-internals/metal/lib/property_get.ts
```diff
@@ -194,6 +194,7 @@ export function getWithDefault<T extends object, K extends Extract<keyof T, stri
     {
       id: 'ember-metal.get-with-default',
       until: '4.0.0',
+      url: 'https://deprecations.emberjs.com/v3.x#toc_ember-metal-get-with-default',
```
rwjblue, line 197: The link doesn't work just yet, do we expect it to be live already?

rwjblue: Oh I see, I just merged https://github.com/ember-learn/deprecation-app/pull/600. Once that is published and we can validate the URL is working as expected this should good to go...

rwjblue (approved):

packages/@ember/-internals/metal/lib/property_get.ts
chrisrng: Works now! :)

PR #18997, opened 2020-05-29
[BUGFIX LTS] invoke methods correctly in TextSupport sendAction

### LINKED ISSUE #18994: Components using text-support mixin do not work native class actions correctly
A simple reproduction:

```js
import Component from '@glimmer/component'; // could be @ember/component also
import { action } from '@ember/object';

export default class MyComponent extends Component {
  @action onFocus() {
    console.log('this never gets hit');
  }
}
```

```hbs
<Textarea @focus-in={{this.onFocus}} />
```

It *does* work if you instead invoke—

- with `{{fn}}`:

    ```hbs
    <Textarea @focus-in={{fn this.onFocus}} />
    ```

- with `{{action}}`:

    ```hbs
    <Textarea @focus-in={{action this.onFocus}} />
    ```

- with `{{action}}` and a string name:

    ```hbs
    <Textarea @focus-in={{action "onFocus"}} />
    ```

The problem is here:

https://github.com/emberjs/ember.js/blob/8cf29959f76eee822192cb820347113d050b204e/packages/%40ember/-internals/views/lib/mixins/text_support.js#L309-L327

The key is that we look up the value from `attrs` first, and `attrs` always ends up with arguments wrapped in `MutableCell` (to support two-way binding), so the check whether `typeof` the action is `"function"` always fails (it's `"object"` instead). There is special handling for `{{action}}` and `{{fn}}` so they *don't* have this, which is why using them works.

We need to make sure we *explicitly* handle the case where the item is a `MutableCell`, and having done so we'll be able to correctly "fallback" for *normal* method-style actions.

cc. @rwjblue. I'll try to have a PR up for this tomorrow morning, with a test for the failure case and a fix.

chriskrycho:
The introduction of the `attrs` API in Ember 1.13 included wrapping items passed to components with `MutableCell`, to support two-way binding. Although two-way binding is gone from much of Ember, the text input components (`Input` and `Textarea`) continue to support it, via the `TextSupport` mixins. The `sendAction` function used by the mixin previously assumed that the only options were for an action to be a string or a function -- not a function wrapped in a `MutableCell`.

The result was that this code, which would be expected to work, did not: it would simply never be invoked.

```hbs
<Textarea @focus-in={{this.didFocusIn}} />
```

Accordingly, add logic to `sendAction` in `text_support.js` to unwrap a mutable cell if it is set, and otherwise to carry on with the logic as it was previously.

Resolves #18994

### DIFF
--- a/packages/@ember/-internals/views/lib/mixins/text_support.js
+++ b/packages/@ember/-internals/views/lib/mixins/text_support.js
@@ -6,6 +6,7 @@ import { get, set, Mixin } from '@ember/-internals/metal';
 import { TargetActionSupport } from '@ember/-internals/runtime';
 import { deprecate } from '@ember/debug';
 import { SEND_ACTION } from '@ember/deprecated-features';
+import { MUTABLE_CELL } from '@ember/-internals/views';
 
 const KEY_EVENTS = {
   13: 'insertNewline',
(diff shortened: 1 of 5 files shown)


packages/@ember/-internals/views/lib/mixins/text_support.js
```diff
@@ -306,11 +307,19 @@ export default Mixin.create(TargetActionSupport, {
 // sendAction semantics for TextField are different from
 // the component semantics so this method normalizes them.
 function sendAction(eventName, view, event) {
-  let actionName = get(view, `attrs.${eventName}`) || get(view, eventName);
+  let action = get(view, `attrs.${eventName}`);
```
chriskrycho, line 310: The use of `actionName` for this was *super* confusing when reading the code, since it was only actually the *name* of an action *some* of the time. I opted to change it (back) to `action` to make it clear that it's *always* an action—sometimes as a string-named action, sometimes as a function.

rwjblue (approved):

rwjblue:
Thank you!

PR #18998, opened 2020-05-31
[CLEANUP]: remove redundant `-input-type` helper

bekzod:
transform that used `-input-type` [helper](https://github.com/emberjs/ember.js/blob/305731ca307eeed8abbd09dd7ab6771031dbefee/packages/ember-template-compiler/lib/plugins/transform-input-type-syntax.ts#L73) is long gone, and logic choosing `-checkbox vs -text-field` has moved to [input template](https://github.com/emberjs/ember.js/blob/master/packages/@ember/-internals/glimmer/lib/templates/input.hbs)


chancancode:
Thank you!!

PR #19001, opened 2020-06-01
[BUGFIX LTS] invoke methods correctly in TextSupport sendAction

chriskrycho:
The introduction of the `attrs` API in Ember 3.13 included wrapping items passed to components with `MutableCell`, to support two-way binding. Although two-way binding is gone from much of Ember, the text input components (`Input` and `Textarea`) continue to support it, via the `TextSupport` mixins. The `sendAction` function used by the mixin previously assumed that the only options were for an action to be a string or a function -- not a function wrapped in a `MutableCell`.

The result was that this code, which would be expected to work, did not: it would simply never be invoked.

```hbs
<Textarea @focus-in={{this.didFocusIn}} />
```

Accordingly, add logic to `sendAction` in `text_support.js` to unwrap a mutable cell if it is set, and otherwise to carry on with the logic as it was previously.

Backports the fix for #18994 (f16d1747) to 3.16 LTS.

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/components/input-angle-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/components/input-angle-test.js
@@ -1,5 +1,6 @@
 import { RenderingTestCase, moduleFor, runDestroy, runTask } from 'internal-test-helpers';
 
+import { action } from '@ember/object';
 import { assign } from '@ember/polyfills';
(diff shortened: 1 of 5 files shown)


chriskrycho:
@rwjblue as promised!

rwjblue (approved):

rwjblue:
Thank you @chriskrycho!

PR #19009, opened 2020-06-10
[Chore]: Add integration test for accessing a property during init

snewcomer:
Screen Shot 2020-06-20 at 7 37 39 AM

close https://github.com/emberjs/ember.js/issues/19003

Fixed by #19023

Issue #19003, opened 2020-06-03
belongsTo relationship causes "used previously in the same computation" error


wongpeiyi:
First posted in https://github.com/emberjs/data/issues/7196 but was referred here

When an API payload is received with an association not already set on a Ember Data model, the following error is seen:

```
Error: Assertion Failed: You attempted to update `content` on `<(unknown):ember173>`, 
but it had already been used previously in the same computation.  Attempting to update 
a value after using it in a computation can cause logical errors, infinite 
revalidation bugs, and performance issues, and is not supported.

`content` was first used:

- While rendering:
  
  application
    foo
      this.dataLength

Stack trace for the update:
    at dirtyTagFor (validator.js:514)
    at markObjectAsDirty (index.js:637)
    at notifyPropertyChange (index.js:675)
    at set (index.js:1622)
    at Proxy.set (observable.js:176)
    at InternalModel._updatePromiseProxyFor (-private.js:4379)
    at InternalModel.getBelongsTo (-private.js:4272)
```

This happens when the model is being consumed in a component template (via a getter) that depends on `post.user`

### Reproduction

https://github.com/wongpeiyi/data-computation

### Description

- When creating and saving a Post, no User association is set yet
- The API (mirage) sets a User and responds with a payload
- Ember Data updates the Post model
- Because the model is being consumed by the component getter that depends on `filterBy('user.id')`, the error is thrown

The error is not thrown when:
- Not filtering (or not consuming the user association)
- Using `{{@data.length}}` in the template instead of the component getter
- Using a getter on the controller without a component

This seems like a use case that should "just work"

### Versions

```cli
yarn list v1.19.2
warning Filtering by arguments is deprecated. Please use the pattern option instead.
└─ ember-source@3.18.1
✨  Done in 0.72s.
yarn list v1.19.2
warning Filtering by arguments is deprecated. Please use the pattern option instead.
└─ ember-cli@3.18.0
✨  Done in 0.65s.
yarn list v1.19.2
├─ @ember-data/adapter@3.18.0
├─ @ember-data/canary-features@3.18.0
├─ @ember-data/debug@3.18.0
├─ @ember-data/model@3.18.0
├─ @ember-data/private-build-infra@3.18.0
├─ @ember-data/record-data@3.18.0
├─ @ember-data/rfc395-data@0.0.4
├─ @ember-data/serializer@3.18.0
├─ @ember-data/store@3.18.0
[...]

PR #19023, opened 2020-06-20
[BUGFIX lts] dont access properties during init unless required

runspired:
Previously we were greedily accessing properties during init in all cases if a value for the property was passed into create. In the case of native getters/setters this could lead to rendering invalidations if initialization was triggered during render since the value would immediately change after access.

This possibly supplants #19009 and may fix if so #19003 but more investigation should be done to confirm that.

### DIFF
--- a/packages/@ember/-internals/runtime/lib/system/core_object.js
+++ b/packages/@ember/-internals/runtime/lib/system/core_object.js
@@ -102,9 +102,8 @@ function initialize(obj, properties) {
       let isDescriptor = possibleDesc !== undefined;
 
       if (!isDescriptor) {
-        let baseValue = obj[keyName];
-
         if (hasConcatenatedProps && concatenatedProperties.indexOf(keyName) > -1) {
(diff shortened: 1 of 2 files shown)


pzuraq (approved):

runspired:
After further investigation this issue began in the 3.15 release and was not present prior, so backporting to 3.16 will allow EmberData to use this pattern back to 3.12 :)

rwjblue:
Thanks for tracking this down!

rwjblue:
@runspired / @snewcomer - This is included in ember-source@3.20.0-beta.3, mind testing to confirm that this fixes that issue in Ember Data?

runspired:
@rwjblue I tested this locally before submitting, but I'll run a quick test to double ensure :)

rwjblue:
Ah, gotcha, thank you! I mostly just want a little bit of sanity check before releasing the patch release of 3.19 (and then more sanity check of that patch release before backporting to 3.16).

Issue #19015, opened 2020-06-15
[Documentation] LinkTo Query Param support is undocumented


elwayman02:
### 📙 Describe the Issue
The API docs for `LinkTo` make no mention of query params whatsoever, despite them being a supported API and even being briefly mentioned in the [routing guides](https://guides.emberjs.com/release/routing/query-params/).

### 🔬 Minimal Reproduction
[LinkTo API docs](https://api.emberjs.com/ember/release/classes/Ember.Templates.components/methods/input?anchor=LinkTo)
[LinkTo code/documentation](https://github.com/emberjs/ember.js/blob/v3.18.1/packages/%40ember/-internals/glimmer/lib/components/link-to.ts#L14)

### 😕 Actual Behavior
Query params are undocumented

### 🤔 Expected Behavior
Clear documentation of how `@query` works as an input to `LinkTo`

### ➕ Additional Context
N/A

chrc:
Can this issue be closed?

rahulk94:
+1 to closing this issue

kategengler:
This has been done.

PR #19024, opened 2020-06-21
[DOC] Added example for LinkTo query param property

chrc:
Add documentation of how @query works as an input to LinkTo.

Issue #19015


locks (approved):
Thanks!

PR #19025, opened 2020-06-22
[BUGFIX beta] Updates Glimmer VM to 0.54.0

pzuraq:
Updates the VM to the latest version, incorporating the changes for destroyables. This PR does not add any new public APIs to Ember, the plan is to do that in a followup.


packages/@ember/-internals/glimmer/lib/component-managers/mount.ts
```diff
@@ -142,20 +146,8 @@ class MountManager extends AbstractManager<EngineState, EngineDefinitionState>
...
-          environment.extra.debugRenderTree.willDestroy(controller);
-          environment.extra.debugRenderTree.willDestroy(bucket);
-          engine.destroy();
-        },
-      };
-    } else {
-      return engine;
-    }
+  getDestroyable(bucket: EngineState): Option<Destroyable> {
+    return bucket.engine;
```
krisselden, line 150: I assume the controller is a child of the engine and is registered to it

pzuraq: Yes, based on the code prior to this, I believe that must have been true. This is effectively the equivalent to the code it's replacing.

packages/@ember/-internals/glimmer/lib/component-managers/curly.ts
```diff
@@ -331,6 +331,10 @@ export default class CurlyComponentManager
         instance: component,
         template: state.template,
       });
+
+      registerDestructor(bucket, () => {
+        environment.extra.debugRenderTree.willDestroy(bucket);
+      });
```
rwjblue, line 337: Shouldn't this be wrapped (the removed code was)? 

```suggestion
      if (ENV._DEBUG_RENDER_TREE) {
        registerDestructor(bucket, () => {
          environment.extra.debugRenderTree.willDestroy(bucket);
        });
      }
```

packages/@ember/-internals/glimmer/lib/component-managers/custom.ts
```diff
@@ -285,6 +285,10 @@ export default class CustomComponentManager<ComponentInstance>
         instance: component,
         template: definition.template,
       });
+
+      registerDestructor(bucket, () => {
+        env.extra.debugRenderTree.willDestroy(bucket);
+      });
```
rwjblue, line 291: Maybe I'm missing something here, we used to only do the work to track / remove stuff when `ENV._DEBUG_RENDER_DEBUG` was set. Has that changed?

packages/@ember/-internals/glimmer/lib/component-managers/curly.ts
pzuraq: It is, if you expand the diff a bit, I combined it with the pre-existing `ENV._DEBUG_RENDER_TREE` setup code that existed in the `create` hook.

packages/@ember/-internals/glimmer/lib/component-managers/custom.ts
pzuraq: Same as above, there was always some setup code in the `create` hook so we can combine it with that

rwjblue (approved):

PR #19028, opened 2020-06-25
[BUGFIX lts] Ensure setter CP's with dependent keys on curly components can be two way bound

richgt:


packages/@ember/-internals/glimmer/tests/integration/components/curly-components-test.js
```diff
@@ -2904,6 +2904,84 @@ moduleFor(
       this.assertText('initial value');
     }
 
+    ['@test GH#18147 - a two way binding flows upstream to a parent component through a CP']() {
```
boris-petrov, line 2907: Typo in the issue ID - should be 18417 instead of 18147.

richgt: Fixed

packages/@ember/-internals/glimmer/lib/component-managers/curly.ts
```diff
@@ -301,6 +304,35 @@ export default class CurlyComponentManager
       }
     }
 
+    // ensure two way binding works when the component has defined a computed
+    // property with both a setter and dependent keys, in that scenario without
+    // the sync observer added below the caller's value will never be updated
+    //
+    // See GH#18147 / GH#19028 for details.
```
boris-petrov, line 311: Same typo here. :smile: @richgt

richgt: Also fixed.

rwjblue:
Added a few reviewers...

richgt:
@rwjblue - we have some scenarios where this appears to not completely fix the issue. We're working on further small-scale reproduction, as we can't quite extract it from our application code, but we do have several other instances of computed properties still not successfully updating up to their two-way bound parents. I'll let you know if we're able to successfully provide smaller-scale reproductions.

richgt:
> @rwjblue - we have some scenarios where this appears to not completely fix the issue. We're working on further small-scale reproduction, as we can't quite extract it from our application code, but we do have several other instances of computed properties still not successfully updating up to their two-way bound parents. I'll let you know if we're able to successfully provide smaller-scale reproductions.

Gut instinct right now tells me it has something to do with two-way binding to properties inside an object (`twb-value=object.value`). Again, still haven't come away with a better reproduction, but we're continuing to work on it. It's pretty consistent across our codebase. Will update when I have more.

pzuraq:
does the observer have to be a sync observer? Can it be async?

rwjblue:
> does the observer have to be a sync observer? Can it be async?

Ya probably, I'll test.

rwjblue:
After discussing with @pzuraq (offline) I think I'm also going to see if it is possible to do from the CP side (instead of the curly component manager side). Specifically, the thought process is that we instantiate **many more** `@ember/component` instances than (we expect) folks use setter CP's with dependent keys. The current implementation introduces a cost for **every component instantiation** (we have to loop over all the named arguments, check if that ran into a setter CP that used dep keys, and then add the observer).

packages/@ember/-internals/metal/lib/computed.ts
```diff
@@ -675,6 +676,25 @@ export class ComputedProperty extends ComputedDescriptor {
     let ret;
 
     try {
+      if (obj instanceof Ember.Component) {
```
rwjblue, line 679: I think we should avoid the `instanceof` check here. We don't want `@ember/-internals/metal` to have imports from `@ember/-internals/glimmer`, and we can't leverage the `Ember` global (this is partially why the Node tests are failing ATM).

I think we should update to use this flag that we set (and AFAICT is only set for Ember.Component):

https://github.com/emberjs/ember.js/blob/730c638eb52791709cd004ad4324b51ce2b43353/packages/%40ember/-internals/glimmer/lib/component.ts#L718

Which would change things to:

```suggestion
      if (obj.isComponent === true) {
```

packages/@ember/-internals/metal/lib/computed.ts
```diff
@@ -675,6 +676,25 @@ export class ComputedProperty extends ComputedDescriptor {
...
+        // ensure two way binding works when the component has defined a computed
+        // property with both a setter and dependent keys, in that scenario without
+        // the sync observer added below the caller's value will never be updated
+        //
+        // See GH#18417 / GH#19028 for details.
+        let descriptor: ComputedDescriptor;
+        descriptor = descriptorForProperty(obj, keyName);
+        let meta = metaFor(obj);
```
rwjblue, line 687: Since `descriptorForProperty` accepts an optional meta (and we are just going to look one up immediately after calling) we should do the meta lookup first and pass it to `descriptorForProperty`:

```suggestion
        let meta = metaFor(obj);
        let descriptor = descriptorForProperty(obj, keyName, meta);
```

packages/@ember/-internals/metal/lib/computed.ts
```diff
@@ -675,6 +676,25 @@ export class ComputedProperty extends ComputedDescriptor {
...
+        if (
+          meta.isInitializing() &&
+          descriptor !== undefined &&
+          descriptor._dependentKeys !== undefined &&
+          descriptor._dependentKeys.length > 0
+        ) {
```
rwjblue, line 693: I'm curious though, why are we doing `descriptorForProperty` at all. The class that this is in **is** the descriptor, see here for where that is setup:

https://github.com/emberjs/ember.js/blob/8b64cc52f828941667dc7ac267c46a2752bccf1f/packages/@ember/-internals/metal/lib/decorator.ts#L62-L69

So, I would expect that we can literally use `this` instead.

packages/@ember/-internals/metal/lib/computed.ts
```diff
@@ -675,6 +676,25 @@ export class ComputedProperty extends ComputedDescriptor {
...
+          addObserver(obj, keyName, obj[PROPERTY_DID_CHANGE].bind(obj, keyName), undefined, true);
```
rwjblue, line 694: I'd generally prefer to use an arrow function here (we don't have the same issue we had over in the component manager because this isn't in a loop) instead of `.bind`.

rwjblue: We also have a bit of a logic issue here: we will add an observer for **every** set that happens.

rwjblue (changes requested):

packages/@ember/-internals/metal/lib/computed.ts
richgt: I thought that was the purpose of `meta.isInitializing`... we'll only do it on the initialization...

richgt (commented):
LGTM

rwjblue (approved):

rwjblue:
Chatted a bit with @krisselden about the updated fix here, we need to do a (hopefully) quick performance test to confirm this does not introduce a regression for existing apps. I believe that @pzuraq mentioned that he would run the comparison.

pzuraq (approved):

pzuraq:
Benchmarked, and this doesn't appear to have a significant impact. We're good to merge! Thanks a ton @richgt!

[artifact-49.pdf](https://github.com/emberjs/ember.js/files/5025364/artifact-49.pdf)

PR #19029, opened 2020-06-27
[BUG]: DEBUG mode does not set property on tracked, previous accessed obj

snewcomer:
In DEBUG mode, an update to a tracked `obj` previous set on `MANDATORY_SETTERS` WeakMap is not applied.  This fixes the issue in the PR below but breaks everything else :(

- [ ] test

ref https://github.com/poteto/ember-changeset/pull/516


snewcomer:
Fixed by https://github.com/emberjs/ember.js/pull/18961

PR #19034, opened 2020-07-01
[BUGFIX beta] Disable named blocks

chancancode:
See https://github.com/ember-polyfills/ember-named-blocks-polyfill/pull/17

Please continue to use the polyfill in the meantime.


chancancode:
I guess this should be merged into beta (with the flag set to false) and another one into master with the flag set to null. Can’t do it since I’m on mobile, but don’t want to forget.

kategengler (approved):

PR #19045, opened 2020-07-17
Replace Travis CI with a Github Actions workflow to run builds

kategengler:
This recreates our current CI from Travis on GitHub Actions. 

Currently Github Actions does not support any yaml reuse, so it is a bit verbose, but supposedly they will support in the future. 

Had to be a bit creative to re-create the cron builds for beta, release, etc. because Actions only currently supports scheduled builds on the default branch. 

For this to work, we need to set up a number of secrets in the repo:

- NPM_AUTH_TOKEN, Token for publishing to npm
- S3_SECRET_ACCESS_KEY, For the bucket where we deploy channels
- S3_ACCESS_KEY_ID
- PERSONAL_TOKEN, Needs to be a Github token with `repo` access to this repo, due to a limitation in actions that does not let a push from one workflow with the default token to trigger another workflow run

Created with a local branch to hopefully see the actions run, supersedes #19013


kategengler:
Build is failing due to network.

kategengler:
Example of the cron build: https://github.com/kategengler/ember.js/actions/runs/172361296 
(Beta on my fork is set to error to show how it looks)

kategengler:
A build that would have deployed to s3 https://github.com/kategengler/ember.js/runs/884295095?check_suite_focus=true

It failed due to authentication (I don't have credentials in my fork for the emberjs bucket), but when you expand the job you can see what it was going to upload and to what path.

kategengler:
A build that would deploy to npm, again failing due to no auth:
https://github.com/kategengler/ember.js/runs/884295543?check_suite_focus=true

CONTRIBUTING.md
```diff
@@ -250,19 +250,19 @@ And in case we didn't emphasize it enough: we love tests!
 
 NOTE: Partially copied from https://raw.github.com/thoughtbot/factory_girl_rails/master/CONTRIBUTING.md
 
-# Travis CI Tests
+# CI (Github Actions) Tests
 
-We use [Travis CI](https://travis-ci.org/emberjs/ember.js/pull_requests) to test each PR before it is merged.
+We use [GitHub Actions](https://github.com/ember.js/ember.js/actions?query=workflow%3ABuild+event%3Apull_request) to test each PR before it is merged.
```
ijlee2, line 255: I noticed that the current URL is missing the period and redirects to 404.
```suggestion
We use [GitHub Actions](https://github.com/emberjs/ember.js/actions?query=workflow%3ABuild+event%3Apull_request) to test each PR before it is merged.
```

kategengler: Good catch, thanks!

.github/workflows/cron.yml
```diff
@@ -0,0 +1,56 @@
...
+        uses: jitterbit/await-check-suites@v1
+        with:
+          ref: ${{ env.SHA }}
+          appSlugFilter: github-actions
+          onlyFirstCheckSuite: true
+          timeoutSeconds: 1800
+
+      - name: Cleanup
+        # Always cleanup the cron-* branches
+        if: ${{ success() }} || ${{ failure() }}
```
chancancode, line 55:
```suggestion
        if: always()
```

I think you don't need `${{ ... }}` for `if` because it's always an expression, and AFAICT `always()` does what we want

chancancode (approved):
I think there are a couple of refactors we should make, such as using volta as @rwjblue requested, and also potentially using the [deployments](https://developer.github.com/v3/repos/deployments/) API to simplify the nightly builds, but assuming this works, I think we can start building on GH and iterate from there.

.github/workflows/cron.yml
kategengler: All the examples have `${{ ... }}` 🤔 https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#job-status-check-functions

But you're right that `always()` is more appropriate because we do want this to clean up even when cancelled.

.github/workflows/build.yml
```diff
@@ -0,0 +1,259 @@
+name: Build
```
rwjblue, line 1: I'd like to rename this to something else, probably Test?

rwjblue (approved):

PR #19050, opened 2020-07-21
Bump router_js to v7.0.0

xg-wang:
router_js v7 contains typescript 3.5 -> 3.9 bump with some type refinement.

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/router.ts
+++ b/packages/@ember/-internals/routing/lib/system/router.ts
@@ -1438,7 +1438,11 @@ function calculatePostTransitionState(
 
     // If the routeInfo is not resolved, we serialize the context into params
     if (!routeInfo.isResolved) {
-      params[routeInfo.name] = routeInfo.serialize(routeInfo.context);
+      params[routeInfo.name] = routeInfo.serialize(routeInfo.context as
+        | {
+            [key: string]: unknown;
+          }
+        | undefined);
     } else {
       params[routeInfo.name] = routeInfo.params;
     }
--- a/package.json
+++ b/package.json
@@ -137,7 +137,7 @@
     "rollup-plugin-commonjs": "^9.3.4",
     "rollup-plugin-node-resolve": "^4.2.4",
     "route-recognizer": "^0.3.4",
-    "router_js": "^6.2.6",
+    "router_js": "^7.0.0",
     "rsvp": "^4.8.5",
     "serve-static": "^1.14.1",
     "simple-dom": "^1.4.0",
--- a/packages/@ember/-internals/routing/lib/system/route.ts
+++ b/packages/@ember/-internals/routing/lib/system/route.ts
@@ -98,7 +98,14 @@ class Route extends EmberObject implements IRoute {
   _internalName!: string;
   _names: unknown;
 
-  serialize!: (model: {}, params: string[]) => object | undefined;
+  serialize!: (
+    model: {},
+    params: string[]
+  ) =>
+    | {
+        [key: string]: unknown;
+      }
+    | undefined;
 
   _router!: EmberRouter;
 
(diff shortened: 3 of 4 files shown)


packages/@ember/-internals/routing/lib/system/router.ts
```diff
@@ -1438,7 +1438,11 @@ function calculatePostTransitionState(
 
     // If the routeInfo is not resolved, we serialize the context into params
     if (!routeInfo.isResolved) {
-      params[routeInfo.name] = routeInfo.serialize(routeInfo.context);
+      params[routeInfo.name] = routeInfo.serialize(routeInfo.context as
```
xg-wang, line 1441: @rwjblue seems `serialize` accepting `null` is okay, should we change the types in router_js or just cast here?

rwjblue: What does `null` mean in that context?

Either way, if it is a supported value in the router_js code, we ought to type it that way...

xg-wang: It traces back to `Route.model` can return `null` https://github.com/tildeio/router.js/blob/v7.0.0/lib/router/route-info.ts#L26, the model can be assigned to `routeInfo.context` https://github.com/tildeio/router.js/blob/v7.0.0/lib/router/route-info.ts#L472

rwjblue: Sure, but `TransitionState.prototype.params` is just this:

```ts
interface IParams {
  [key: string]: unknown;
}
```

https://github.com/tildeio/router.js/blob/cfe549c68aca9f14e834658e12b2633c435198db/lib/router/transition-state.ts#L7-L10

Why wouldn't `unknown allow `null` there? I assume we must be narrowing incorrectly here in Ember (not in `router_js`).

xg-wang: `routeInfo.serialize` return type is fine here, the problem is `routeInfo.context` passed to `_context` cannot be `null`

```ts
  // https://github.com/tildeio/router.js/blob/v7.0.0/lib/router/route-info.ts#L221
  serialize(_context?: Dict<unknown> /*<- this is problematic*/): Dict<unknown> | undefined /*<- this is fine*/ {
    return this.params || {};
  }
```

The actual code in the base class `InternalRouteInfo` and inherit  `UnresolvedRouteInfoByObject` can deal with falsy `_context`
https://github.com/tildeio/router.js/blob/v7.0.0/lib/router/route-info.ts#L515

rwjblue: I guess I'm a bit surprised that this has anything to do with `model` (https://github.com/tildeio/router.js/blob/cfe549c68aca9f14e834658e12b2633c435198db/lib/router/route-info.ts#L23-L26), since the `context` argument here is effectively the params (not the model result).

Regardless, you are totally correct that `route-info` itself is not internally consistent.

https://github.com/tildeio/router.js/blob/cfe549c68aca9f14e834658e12b2633c435198db/lib/router/route-info.ts#L205

And 

https://github.com/tildeio/router.js/blob/cfe549c68aca9f14e834658e12b2633c435198db/lib/router/route-info.ts#L221

Should match...

rwjblue:
Going to land this for now, we can fix the mismatch in a follow up.

packages/@ember/-internals/routing/lib/system/router.ts
xg-wang: Ugh it does not seem straight forward to make them match, I ran into promise / no-promise type error.
My guess from reading the code is that the router_js routeInfo `serialize` method only accepts resolved model or identifier (param). Its context can be these or unresolved model promise, but not inside `calculatePostTransitionState`.

Likely router_js needs a larger type refactoring around different RouteInfos but I don't have a clear idea now.

rwjblue: Ya, I was thinking that you want the resolved values from https://github.com/tildeio/router.js/blob/cfe549c68aca9f14e834658e12b2633c435198db/lib/router/route-info.ts#L205

So basically:

```ts
Dict<unknown> | null | undefined
```

PR #19055, opened 2020-07-26
[CLEANUP] remove dependence on `component.elementId` in `ViewRegistry`

bekzod:
Using`Set` for `ViewRegistry` to simplify things, and remove dependence on `component.elementId` and `component guid`.
As a side-effect it enables registering components with duplicate `elementId`s (unsure if it is good or bad thing), at least for template-only components you can use duplicate ids currently.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/renderer.ts
+++ b/packages/@ember/-internals/glimmer/lib/renderer.ts
@@ -246,13 +246,9 @@ function loopEnd() {
 backburner.on('begin', loopBegin);
 backburner.on('end', loopEnd);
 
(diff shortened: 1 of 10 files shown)


rwjblue:
> at least for template-only components you can use duplicate ids currently

True, but these don't end up in the view registry as far as I know...

rwjblue (commented):
Ultimately we want to remove the view registry all together, we just need to ensure that we can provide access to the expected APIs (other than view registry) when we do.

bekzod:
Thanks for explaining, i think it is better close it then

PR #19069, opened 2020-08-01
[CLEANUP] remove some internal usage `view-registry`

bekzod:
* removed `_viewRegistry` from `renderer`
* removed usage of `view-registry` from `getChildViews`
* removed `this.owner.lookup('-view-registry:main');` internally and using `VIEW_REGISTRY` directly

Should we add deprecation for `this.owner.lookup('-view-registry:main');` ?


packages/@ember/application/instance.js
```diff
@@ -35,6 +35,7 @@ import { renderSettled } from '@ember/-internals/glimmer';
 */
 
 const ApplicationInstance = EngineInstance.extend({
+  VIEW_REGISTRY,
```
bekzod, line 38: using it here only to make the test pass, is there a better way ?

rwjblue:
Eeck, sorry @bekzod mind rebasing again?

@chancancode - This seem good to you?

locks:
This would need to much reworking 😅.

PR #19073, opened 2020-08-03
Update Ember.Component API docs to use native classes

mehulkar:
Some notable issues with this, which might be why this was never done..

- `tagName = 'lkasjd'` works, but is technically supposed to come from a decorator. See https://github.com/emberjs/rfcs/issues/589
- I don't know if attributeBindings, etc work out of box with native classes
- Didn't finish updating the whole document, thought I'd make sure this is actually something that should happen before completing.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/component.ts
+++ b/packages/@ember/-internals/glimmer/lib/component.ts
@@ -126,9 +126,11 @@ export const BOUNDS = symbol('BOUNDS');
 
   ```app/components/person-profile.js
   import Component from '@ember/component';
-
-  export default Component.extend({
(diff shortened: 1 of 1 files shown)


mehulkar:
@locks pointed me to a upgrade doc saying that native classes for classic components are NOT supported, so leaving this as is. https://guides.emberjs.com/release/upgrading/current-edition/native-classes/#toc_native-classes-for-classic-component

cc @pzuraq as he said it might be worth revisiting. i tend to agree that there are too many papercuts here to officially recommend, and the time/energy it will take to iron out those papercuts would be better spent adopting glimmer components!

PR #19075, opened 2020-08-04
integration/components/life-cycle: fix removeComponent

tashvaqa:
`this.componentRegistry` stores component ids not instances.


rwjblue (approved):
Nice find, thanks!

PR #19077, opened 2020-08-05
[BUGFIX release] Simplify get and computeds

pzuraq:
This refactor simplifies get and computeds in the following ways:

- Removes several brand checks that were unnecessary from `get`:
  - `isEmberArray`: Ember Arrays that are not native arrays will
    automatically be tracked through proper usage.
  - `isProxy`: This was only necessary for conditionals, and updates in
    the VM allow us to autotrack that instead.
- Uses `tagFor` instead of `tagForProperty` in many places including
  `get`. This allows us to avoid the brand check for
  `CUSTOM_PROPERTY_TAG`, which is really only necessary for chain tags.
- Updates everywhere that looks up tags multiple times on the same
  object to use a shared `tagMeta` so we don't lookup that map multiple
  times.
- Moves computed revision and value caches onto Meta, so we're looking
  up fewer maps.
- Creates a new AutoComputed descriptor for computed properties that
  use autotracking, so we can simplify the get logic for standard CPs.


packages/@ember/-internals/metal/lib/computed_cache.ts
```diff
@@ -1,59 +1,69 @@
...
+//   @method cacheFor
+//   @static
+//   @for @ember/object/internals
+//   @param {Object} obj the object whose property you want to check
+//   @param {String} key the name of the property whose cached value you want
+//     to return
+//   @return {Object} the cached value
+//   @public
+// */
+// export function getCachedValueFor(obj: object, key: string): any {
```
krisselden, line 39: did you mean to comment out or delete?

pzuraq: meant to delete, thanks for catching that

pzuraq:
Benchmark results against current master of Ember and Glimmer VM: 
[artifact-53.pdf](https://github.com/emberjs/ember.js/files/5029485/artifact-53.pdf)

packages/@ember/-internals/runtime/lib/mixins/-proxy.js
```diff
@@ -73,7 +74,17 @@ export default Mixin.create({
 
       return tag;
     } else {
-      return combine([tag, ...getChainTagsForKey(this, `content.${key}`, addMandatorySetter)]);
+      let tags = [tag];
+
+      tags.push(tagFor(this, 'content', meta));
```
krisselden, line 79: inline this into the array initialization

packages/@ember/-internals/runtime/lib/system/array_proxy.js
```diff
@@ -119,7 +120,14 @@ export default class ArrayProxy extends EmberObject {
       // revalidate eagerly if we're being tracked, since we no longer will
       // be able to later on due to backtracking re-render assertion
       this._revalidate();
-      return combine(getChainTagsForKey(this, `arrangedContent.${key}`, addMandatorySetter));
+      let tags = [this._arrangedContentTag];
+      let arrangedContent = get(this, 'arrangedContent');
+
+      if (isObject(arrangedContent)) {
+        tags.push(tagForProperty(arrangedContent, key, addMandatorySetter));
+      }
```
krisselden, line 128: in the else here there is no point in it being an array and calling combine, why not move the combine inside here

krisselden (approved):

PR #19080, opened 2020-08-05
Lazily setup the router in non-application tests

xg-wang:
During non setupApplicationTests type tests, the Router being injected
into service:router and service:routing does not start routing, in which
it initializes its _routerMicrolib. Public API on service:router will
throw in those tests.
This commit will trigger startRouting on router, in non application
tests the router service should just work.

---
Original PR: https://github.com/emberjs/ember.js/pull/17557
I cannot reopen bc force pushed the branch...

This PR is created with rebase and few changes:
1. using `setupRouter`(the semantics are a bit different, `startRouting` will do an initial transition which we don't necessarily want)
2. rename `non_application_test.js` to `non_application_test_test.js` (because testing usages in tests 😸)

### DIFF
--- a/packages/@ember/-internals/routing/lib/services/router.ts
+++ b/packages/@ember/-internals/routing/lib/services/router.ts
@@ -71,8 +71,8 @@ export default class RouterService extends Service {
     if (router !== undefined) {
       return router;
     }
-    const owner = getOwner(this) as any;
-    router = owner.lookup('router:main');
+    const owner = getOwner(this) as Owner;
+    router = owner.lookup('router:main') as EmberRouter;
     router.setupRouter();
     return (this[ROUTER] = router);
   }
--- a/packages/@ember/-internals/routing/lib/services/router.ts
+++ b/packages/@ember/-internals/routing/lib/services/router.ts
@@ -1,4 +1,4 @@
-import { getOwner } from '@ember/-internals/owner';
+import { getOwner, Owner } from '@ember/-internals/owner';
 import { Evented } from '@ember/-internals/runtime';
 import { symbol } from '@ember/-internals/utils';
 import { assert } from '@ember/debug';
@@ -77,10 +77,10 @@ export default class RouterService extends Service {
     return (this[ROUTER] = router);
   }
 
-  init() {
-    super.init(...arguments);
+  constructor(owner: Owner) {
+    super(owner);
 
-    const router = getOwner(this).lookup('router:main') as EmberRouter;
+    const router = owner.lookup('router:main') as EmberRouter;
 
     router.on('routeWillChange', (transition: Transition) => {
       if (DEBUG) {
--- a/packages/@ember/-internals/routing/lib/services/router.ts
+++ b/packages/@ember/-internals/routing/lib/services/router.ts
@@ -11,7 +11,7 @@ import { Transition } from 'router_js';
 import EmberRouter, { QueryParam } from '../system/router';
 import { extractRouteArgs, resemblesURL, shallowEqual } from '../utils';
 
-const ROUTER = symbol('ROUTER');
+const ROUTER = symbol('ROUTER') as string;
 
 let freezeRouteInfo: Function;
 if (DEBUG) {
@@ -80,14 +80,16 @@ export default class RouterService extends Service {
   init() {
     super.init(...arguments);
 
-    this._router.on('routeWillChange', (transition: Transition) => {
+    const router = getOwner(this).lookup('router:main') as EmberRouter;
+
+    router.on('routeWillChange', (transition: Transition) => {
       if (DEBUG) {
         freezeRouteInfo(transition);
       }
       this.trigger('routeWillChange', transition);
     });
 
-    this._router.on('routeDidChange', (transition: Transition) => {
+    router.on('routeDidChange', (transition: Transition) => {
       if (DEBUG) {
         freezeRouteInfo(transition);
       }
--- a/packages/@ember/-internals/routing/lib/services/router.ts
+++ b/packages/@ember/-internals/routing/lib/services/router.ts
@@ -1,4 +1,6 @@
+import { getOwner } from '@ember/-internals/owner';
 import { Evented } from '@ember/-internals/runtime';
+import { symbol } from '@ember/-internals/utils';
 import { assert } from '@ember/debug';
 import { readOnly } from '@ember/object/computed';
 import { assign } from '@ember/polyfills';
@@ -8,6 +10,8 @@ import { consumeTag, tagFor } from '@glimmer/validator';
 import { Transition } from 'router_js';
 import EmberRouter, { QueryParam } from '../system/router';
 import { extractRouteArgs, resemblesURL, shallowEqual } from '../utils';
+
+const ROUTER = symbol('ROUTER');
 
 let freezeRouteInfo: Function;
 if (DEBUG) {
@@ -62,7 +66,16 @@ function cleanURL(url: string, rootURL: string) {
    @class RouterService
  */
 export default class RouterService extends Service {
-  _router!: EmberRouter;
+  get _router(): EmberRouter {
+    let router = this[ROUTER];
+    if (router !== undefined) {
+      return router;
+    }
+    const owner = getOwner(this) as any;
+    router = owner.lookup('router:main');
+    router.setupRouter();
+    return (this[ROUTER] = router);
+  }
 
   init() {
     super.init(...arguments);
--- a/packages/@ember/-internals/routing/lib/system/router.ts
+++ b/packages/@ember/-internals/routing/lib/system/router.ts
@@ -127,6 +127,7 @@ class EmberRouter extends EmberObject {
   location!: string | IEmberLocation;
   rootURL!: string;
   _routerMicrolib!: Router<Route>;
+  _didSetupRouter = false;
 
   currentURL: string | null = null;
   currentRouteName: string | null = null;
@@ -397,9 +398,8 @@ class EmberRouter extends EmberObject {
     @private
   */
   startRouting() {
-    let initialURL = get(this, 'initialURL');
-
     if (this.setupRouter()) {
+      let initialURL = get(this, 'initialURL');
       if (initialURL === undefined) {
         initialURL = get(this, 'location').getURL();
       }
@@ -411,6 +411,10 @@ class EmberRouter extends EmberObject {
   }
 
   setupRouter() {
+    if (this._didSetupRouter) {
+      return false;
+    }
+    this._didSetupRouter = true;
     this._setupLocation();
 
     let location = get(this, 'location');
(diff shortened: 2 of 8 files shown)


packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -62,6 +62,12 @@ function cleanURL(url: string, rootURL: string) {
  */
 export default class RouterService extends Service {
   _router!: EmberRouter;
+  get router() {
+    if (!this._router._didSetupRouter) {
```
rwjblue, line 66: Since `setupRouter` already does this, do we need to check `this._router._didSetupRouter` here (feels vaguely bad to be looking at private props of a private prop...)?

packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -62,6 +62,12 @@ function cleanURL(url: string, rootURL: string) {
  */
 export default class RouterService extends Service {
   _router!: EmberRouter;
+  get router() {
```
rwjblue, line 65: I worry about directly exposing the private router here. If folks do `this.router.router` now they have access to the `router:main` thing, which doesn't seem like exactly what we want 🤔.

I do like forcing all of our callers to go through the getter though 🤔

xg-wang: Maybe `get privateRouter()`? Folks do have access to `_router` anyways, it's mostly about naming it to callout the private access.

rwjblue: >  Folks do have access to _router anyways, it's mostly about naming it to callout the private access.

Ya, exactly. But IMHO, at that point why do we need to keep two properties? I'm wondering if we can remove the auto-injection of `_router` (as `router:main`) from here:

https://github.com/emberjs/ember.js/blob/3ef84958940e5aa7e551eb0404b3966c8c9728b2/packages/%40ember/application/lib/application.js#L1170

Instead, we could do this:

```js
const ROUTER = symbol('ROUTER'); // uses the internal symbol utility method

// ...snip...

get _router() {
  const router = this[ROUTER];
  if (router !== undefined) {
    return router;
  }

  let owner = getOwner(this);
  owner.setupRouter();

  return this[ROUTER] = owner.router;
}
```

xg-wang: Change a bit to do `owner.lookup('router:main')`, because in `RenderingTestCase` it appears there's no `owner.router`.
`router:main` is also injected into `service:-routing`, but I'm skipping those since I'm not sure if it's safe to change engine.

xg-wang:
@rwjblue (friendly ping) can you take another look?

packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -61,7 +65,16 @@ function cleanURL(url: string, rootURL: string) {
...
+  get _router(): EmberRouter {
+    let router = this[ROUTER];
+    if (router !== undefined) {
+      return router;
+    }
+    const owner = getOwner(this) as any;
+    router = owner.lookup('router:main');
+    router.setupRouter();
+    return (this[ROUTER] = router);
+  }
```
rwjblue, line 77: Maybe we could refactor to:

```suggestion
  get _router(): EmberRouter {
    let router = this[ROUTER];
    if (router !== undefined) {
      return router;
    }
    
    const owner = getOwner(this) as any;
    owner.setupRouter();
    
    return this[ROUTER] = owner.router;
  }
  
  constructor(owner) {
    super(owner);
    
    let { router } = owner;
    
    router.on('routeWillChange', () => { /* snip */ });
    router.on('routedidChange', () => { /* snip */ });
  }
```

This would require moving that `didSetupRouter` flag back to the `owner` though...

What do you think?

packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -61,7 +65,16 @@ function cleanURL(url: string, rootURL: string) {
...
 export default class RouterService extends Service {
-  _router!: EmberRouter;
+  get _router(): EmberRouter {
+    let router = this[ROUTER];
+    if (router !== undefined) {
+      return router;
+    }
+    const owner = getOwner(this) as any;
+    router = owner.lookup('router:main');
+    router.setupRouter();
```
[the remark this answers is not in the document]
xg-wang: We can do another lookup in `init` call to get the `_router`

packages/@ember/-internals/routing/lib/services/router.ts
xg-wang: Do we need `EMBER_FRAMEWORK_OBJECT_OWNER_ARGUMENT` if moving the event listeners from `init` to `constructor`?

rwjblue: I'm not sure I follow what you mean.

packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -8,6 +10,8 @@ import { Transition } from 'router_js';
 import EmberRouter, { QueryParam } from '../system/router';
 import { extractRouteArgs, resemblesURL, shallowEqual } from '../utils';
 
+const ROUTER = symbol('ROUTER') as string;
```
rwjblue, line 13: Why do you have to cast to string here? Most of the other usages in the codebase don't seem to cast...

xg-wang: It's weird I'm getting lint error if not doing so
```
packages/@ember/-internals/routing/lib/services/router.ts:69:23 - error TS2538: Type 'symbol' cannot be used as an index type.

69     let router = this[ROUTER];
```

packages/@ember/-internals/routing/lib/services/router.ts
xg-wang: Ah sorry please ignore this comment \_(:з」∠)\_ I must have mistaken https://github.com/emberjs/rfcs/pull/451 with something else

packages/@ember/-internals/routing/lib/services/router.ts
xg-wang: As to explain why it's only erroring here, I'm seeing these patterns in the codebase
image

rwjblue: Ugh. Maybe we should make `symbol` be typed this way to begin with (to avoid each callsite from having this pain). Either way, seems fine to do separately.

packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -61,19 +65,30 @@ function cleanURL(url: string, rootURL: string) {
...
+    router.setupRouter();
+    return (this[ROUTER] = router);
+  }
+
+  constructor(owner: Owner) {
+    super(owner);
 
-  init() {
-    super.init(...arguments);
+    const router = owner.lookup('router:main') as EmberRouter;
```
rwjblue, line 82: Just to confirm, we have to `.lookup` here because `owner.router` may not be setup?

packages/ember/tests/routing/router_service_test/non_application_test_test.js
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    get routerService() {
+      return this.owner.lookup('service:router');
+    }
+
+    afterEach() {
+      super.afterEach();
+    }
+
```
rwjblue, line 45: This doesn't `return super.afterEach()` so if the super classes `afterEach` returned a promise we wouldn't properly wait for the async, but ultimately since we don't have any customizations here, we can probably delete this (`super.afterEach()` would be called naturally).

```suggestion
```

packages/@ember/-internals/routing/lib/system/router.ts
```diff
@@ -472,7 +476,9 @@ class EmberRouter extends EmberObject {
       this._toplevelView = OutletView.create();
       this._toplevelView.setOutletState(liveRoutes as GlimmerOutletState);
       let instance: any = owner.lookup('-application-instance:main');
-      instance.didCreateRootView(this._toplevelView);
+      if (instance) {
+        instance.didCreateRootView(this._toplevelView);
+      }
```
rwjblue, line 481: On a cursory review, this seems _somewhat_ unrelated to the other changes. What circumstances would have forced `_setOutlets` to be called without an `-application-instance:main` (IOW why is this change needed)?

packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -61,19 +65,30 @@ function cleanURL(url: string, rootURL: string) {
...
  */
 export default class RouterService extends Service {
-  _router!: EmberRouter;
+  get _router(): EmberRouter {
+    let router = this[ROUTER];
+    if (router !== undefined) {
+      return router;
+    }
+    const owner = getOwner(this) as any;
+    router = owner.lookup('router:main');
```
rwjblue, line 74: Can we cast these to the correct values?

```suggestion
    const owner = getOwner(this) as Owner;
    router = owner.lookup('router:main') as EmberRouter;
```

Not sure if I was missing something here...

rwjblue:
Thanks for pushing this forward @xg-wang!

packages/@ember/-internals/routing/lib/system/router.ts
xg-wang: hmmm `RenderingTestCase` does not have `-application-instance:main` registered. I guess it's just how ember internal test setup, not having any real impact on user's test. Should we use another base test case class?

packages/@ember/-internals/routing/lib/services/router.ts
xg-wang: Made the change because in the added `RenderingTestCase` there's no `owner.router`, and `router` is also not on `Owner` type.

packages/@ember/-internals/routing/lib/system/router.ts
rwjblue: No, I think its fine but I'd like to make an explicit test case so we don't accidentally make rendering test case have one and remove the guard. Specifically, I think tests that use a custom resolver (e.g. ember-engines's `engineResolverFor`) will have this exact issue.

xg-wang: Created a RouterNonApplicationTestCase

xg-wang:
@rwjblue the tests are passing (they failed because timeout issue before) can you take a look?

rwjblue (approved):

rwjblue:
Thanks @xg-wang!

lifeart:
looks like documentation, mentioned `setupRouter` should be updated too https://guides.emberjs.com/release/tutorial/part-2/route-params/

xg-wang:
@lifeart Good catch! I'm taking a look at how to update that page

PR #19081, opened 2020-08-05
[BUGFIX release] Adds isProduction flag for the template compiler

### LINKED ISSUE #18598: Increased asset size from `-assert-implicit-component-helper-argument`
Unconditionally including the `-assert-implicit-component-helper-argument` (added in #17135) in builds can lead to significant increases in code size, especially in large apps with lots of components. Also related: #17251.

Given that Ember can now do different *template* builds for prod and dev, it would be great if we could pull this assertion back out of prod builds!

pzuraq:
This PR adds an `isProduction` flag option to the template compiler that can be used to change the behavior of the compiler for production builds. We generally do this already for the JS side of Ember, but not for templates, and this closes that gap.

Notes:

Exposes an `env.isProduction` (or for legacy style AST plugins `this.options.isProduction`) boolean that can be used from within the plugin itself to toggle debug vs prod only behaviors.

Fixes https://github.com/emberjs/ember.js/issues/18598

### DIFF
--- a/packages/ember-template-compiler/lib/plugins/assert-against-named-blocks.ts
+++ b/packages/ember-template-compiler/lib/plugins/assert-against-named-blocks.ts
@@ -1,7 +1,7 @@
-import { StaticTemplateMeta } from '@ember/-internals/views';
 import { assert } from '@ember/debug';
-import { AST, ASTPlugin, ASTPluginEnvironment } from '@glimmer/syntax';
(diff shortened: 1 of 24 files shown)


rwjblue:
@pzuraq - I pushed an updated implementation, this should be ready for another look...

pzuraq (commented):
Changes look perfect!

PR #19082, opened 2020-08-07
[BUGFIX release] Simplify mixin application

pzuraq:
Simplifies mixin application in a number of ways:

- Ensures that we only access `meta` once for a given object when
  applying mixins
- Ensures that we only `peekDescriptors` once for a given descriptor
- Minimizes the number of if/else branches and defaulting in general
- Breaks apart `defineProperty` so that we can do less work per
  definition when mixing in mixins, since we know more about what
  possible operations will occur.
- Removes extra brand checks from `defineProperty` (Array.isArray) so
  we don't penalize every defineProperty for doing that.
- Only revalidate observers once per mixin application.
- Only brand check `didDefineProperty` once per mixin application.
- Replace `for..in` loops with `Object.keys` since we only care about
  own properties.
- Only loop once in `extractAccessors`.
- Combine observer and listener meta into a single object so we only
  need to do one lookup. In most cases, this is undefined in modern
  applications, so no extra memory cost.
- Simplify CP descriptor property lookups (remove duplicate functions).

Benchmark results:

[artifact-60.pdf](https://github.com/emberjs/ember.js/files/5039417/artifact-60.pdf)


packages/@ember/-internals/metal/lib/descriptor_map.ts
```diff
@@ -1,10 +1,7 @@
 import { Meta, peekMeta } from '@ember/-internals/meta';
 import { assert } from '@ember/debug';
 
-const DECORATOR_DESCRIPTOR_MAP: WeakMap<
-  import('./decorator').Decorator,
-  import('./decorator').ComputedDescriptor | boolean
-> = new WeakMap();
+const DECORATOR_DESCRIPTOR_MAP: WeakMap<Function, unknown | boolean> = new WeakMap();
```
krisselden, line 4: `unknown | boolean` is the same as `unknown`

PR #19086, opened 2020-08-07
Pass transition object to activate/deactivate hooks and events

### LINKED ISSUE #19022: [Bug] activate hook doesn't receive transition as an argument
The `activate` hook is called by the private `enter` method:

https://github.com/emberjs/ember.js/blob/v3.20.0-beta.2/packages/@ember/-internals/routing/lib/system/route.ts#L380-L384

which is called by the underlying router.js library

https://github.com/tildeio/router.js/blob/v6.2.5/lib/router/router.ts#L520

Passing along the `transition` would allow the activate hook to know if the _leaf_ route was activated (i.e. if `this.fullRouteName === transition.targetName`).

I'm not sure if this is an oversight or not, but it did not seem significant enough to file a feature request / RFC.

mehulkar:
Closes #19022

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/route.ts
+++ b/packages/@ember/-internals/routing/lib/system/route.ts
@@ -360,9 +360,9 @@ class Route extends EmberObject implements IRoute {
 
     @method exit
   */
-  exit() {
-    this.deactivate();
-    this.trigger('deactivate');
+  exit(transition?: Transition) {
+    this.deactivate(transition);
+    this.trigger('deactivate', transition);
     this.teardownViews();
   }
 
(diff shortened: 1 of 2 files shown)


packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -599,7 +599,7 @@ class Route extends EmberObject implements IRoute {
     @since 1.0.0
     @public
   */
-  deactivate() {}
+  deactivate(_transition?: Transition) {}
```
rwjblue, line 602: Can you add:

```
@param {Transition} transition
```

To the API docs just above (for both this and activate).

mehulkar: good call, added

rwjblue (approved):

rwjblue:
Thank you!

PR #19090, opened 2020-08-11
 [DOC] Fix hint about event handler argument order

maxhq:


rwjblue:
Seems great, thank you!

Issue #18416, opened 2019-09-20
router.replaceWith calls pushState on the history object with refreshModel QP


mehulkar:
https://github.com/mehulkar/ember-example-replace-push-state

I think this may have the same root cause as #15801, but it manifests differently. 

## Problem

If a queryParam is configured to `refreshModel: true`, calls to `router.replaceWith` end up calling `pushState` on history instead of `replaceState`. 

I'm not totally sure how this would manifest as a "bug" in a browser for an end user, but in my case, I have a custom `Location` using a custom `history` object, and calling `pushState` when we really meant to replace causes it to get an extra entry history entry.

## Debugging

It seems that Ember (or underlying `router.js` lib thinks that because of the queryParams refreshModel config, this replaceWith transition is invalid, and it cancels it and generates its own. 

Because router.replaceWith is implemented like this:

```
replaceWith() {
  this.transitionTo(...arguments).method('replace');
}
```

the `transitionTo()` promise gets replaced, and all the async stuff that causes the URL to update happens on the *new* promise, which does not yet have the `urlMethod` set to `replace`. 

One possible solution here is that instead of using `transitionTo()`, a `Transition` object should be constructed with `urlMethod` set as part of the constructor, so that it does the right thing throughout its lifecycle, but that seems like a really big change.

mehulkar:
#5566 may also be related

mehulkar:
As recommended in https://github.com/emberjs/ember.js/issues/5566#issuecomment-64989276, I'm able to solve the issue by using the private `queryParamDidChange` method instead

mehulkar:
Converting to a native class in the route seems to break this workaround, so I might be back at square 1.

alexlafroscia:
I noticed something similar in my app (it could the exact same issue but it's a little hard to tell)

I put together a Twiddle that shows the problem we are facing

https://ember-twiddle.com/39fe48997dd6e62f6381e89d3c6aa74c?openFiles=routes.first.index%5C.js%2C

Essentially, the flow of the logic is

1. Click on a link to a route
2. A route within that path does a `replaceWith` to set some default query params
3. Further down the routing "stack", another `replaceWith` is performed to redirect to different route within the routing hierarchy of the parent that added the query param

I would expect that the browser's history receives a single `pushState` with the _final_ URL, at the "child" route, with the query param.

[...]

PR #19092, opened 2020-08-12
test: add failing test around `replaceWith` and `refreshModel`

alexlafroscia:
I've been trying to see if I can bet to the bottom of what is going on in #18416. I figured I would start with a failing test and work backward from there, but I couldn't get very far before hitting a different failure altogether.

I don't intend to land the test in this partial state, but hope that this can be a starting place for actually fixing this bug.

I want _want_ the test to verify is that, when calling `replaceWith` in `afterModel`, the router does not unintentionally push a new state into the history stack, which is what actually happens. However, before I can even get to that kind of assertion in the test, I run into this error:

Screen Shot 2020-08-12 at 5 11 23 PM

(Screenshot from an [ember twiddle I created](https://ember-twiddle.com/4073a593a39b6f60fefec94c7f8011b2) to make sure that this error was actually happening in a "real" Ember app and not just the test suite. The test suite will fail with the same error.)

What appears to be happening is that the `Transition` object is created without any `routeInfos`

https://github.com/tildeio/router.js/blob/d885da22340da98dcadb45427766efade54ce832/lib/router/router.ts#L123-L130

but then those `routeInfos` are passed around to a number of functions that expect them to exist as a side-effect of calling `finalizeQueryParamChange`.

I am going to assume that this is a fix that will need to at least _start_ in `router.js` and then continue through to here, but figured I would at least post this PR as some small step toward getting to an actual fix.

### DIFF
--- a/packages/ember/tests/routing/query_params_test.js
+++ b/packages/ember/tests/routing/query_params_test.js
@@ -628,6 +628,39 @@ moduleFor(
       assert.equal(indexModelCount, 2, 'index model hook ran again due to refreshModel');
     }
 
+    async ['@test refreshModel does not force `replaceWith` to add a history entry'](assert) {
+      // Default to `omg` being empty
+      this.setSingleQPController('index', 'omg', null);
+
+      this.add(
+        'route:index',
+        Route.extend({
+          queryParams: {
+            omg: {
+              refreshModel: true,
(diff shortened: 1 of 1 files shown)


kategengler:
Is this still relevant?

PR #19094, opened 2020-08-14
Resolves #19004: RouterService#isActive() now consumes currentURL and…

### LINKED ISSUE #19004: Is RouterService#isActive(someRoute) supposed to entagle with tracking?
Just noticed that if I have:

```ts
@service router;

get isTabActive() {
  return this.router.isActive(this.args.route); // route is static for this scenario
}
```
isTabActive never updates when the URL changes / a router transition occurs

NullVoxPopuli:
… currentRouteName

### DIFF
--- a/packages/@ember/-internals/routing/lib/services/router.ts
+++ b/packages/@ember/-internals/routing/lib/services/router.ts
@@ -4,8 +4,8 @@ import { readOnly } from '@ember/object/computed';
 import { assign } from '@ember/polyfills';
 import Service from '@ember/service';
 import { DEBUG } from '@glimmer/env';
-import { Transition } from 'router_js';
 import { consumeTag, tagFor } from '@glimmer/validator';
+import { Transition } from 'router_js';
 import EmberRouter, { QueryParam } from '../system/router';
 import { extractRouteArgs, resemblesURL, shallowEqual } from '../utils';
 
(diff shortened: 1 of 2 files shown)


NullVoxPopuli:
> ERROR: /home/runner/work/ember.js/ember.js/packages/@ember/-internals/routing/lib/services/router.ts:8:1 - Import sources within a group must be alphabetized.

I don't understand:

image

NullVoxPopuli:
oh! it's because it's tslint, not eslint. I don't have a plugin for tslint

packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -309,6 +310,19 @@ export default class RouterService extends Service {
...
+    // in
+    // get isBarActive() {
+    //   return isActive('foo.bar');
+    // }
+    //
+    // you'd expect isBarActive to be dirtied when the route changes.
+    //
+    // https://github.com/emberjs/ember.js/issues/19004
+    consumeTag(tagFor(this, 'currentURL'));
+    consumeTag(tagFor(this, 'currentRouteName'));
```
rwjblue, line 324: These are both aliased computed properties, I'd prefer to entangle the actual root state itself (not the CP's here).

Thoughts @pzuraq?

pzuraq: I think that's true, but I'm unsure if it's practical given the state of the router right now, unless you mean entangling `_router.currentURL`. I think entangling the `currentURL` state makes sense in the meantime, until we refactor the router. I'm not sure why we would have to entangle `currentRouteName` as well though, we should try to our dependencies here I think.

rwjblue: > unless you mean entangling _router.currentURL

Ya, I mostly mean: don't use the alias CP, entangle the root state from `router:main` directly.

NullVoxPopuli: done & pushed

PR #19098, opened 2020-08-17
Add a cache step to the github workflow for node_modules based on the yarn.lock

kategengler:


kpfefferle:
Is caching `node_modules` measurably faster than a fresh `yarn install`? When we previously tried GitHub Actions caching for our Ember projects at work, the cache restore step consistently was *less* performant than building the dependencies from scratch. I'd certainly be happy to find out if performance has improved since then.

kategengler:
@kpfefferle That had been my experience before as well, which is why there isn't already a cache step in the build. I happened to try `actions/cache` again yesterday on a private repo and it dropped the install step from 2 minutes to a combined 20s (cache restore, install, and save). 

We should be able to confirm any saved time before merging this PR by seeing the later steps use the cache from earlier steps (it could also be seen by re-running the PR's jobs and seeing a cache hit that way).

kpfefferle:
@kategengler Thanks for the update - it sounds like it might be time for me to give it a try again too! 🙂

kategengler:
You can see the results on the Blueprint Tests step and later -- trades a 1min 40s install step for 20 seconds or so cache restore and then install.

rwjblue:
> You can see the results on the Blueprint Tests step and later -- trades a 1min 40s install step for 20 seconds or so cache restore and then install.

Interesting. FWIW, I had exactly the same experience as both you and @kpfefferle (that using cache was actually _slower_ or roughly no savings). 🤔

rwjblue (approved):

PR #19104, opened 2020-08-23
Quicker load using ejsr method

ghost:
Please add the code as with estestend to prevent bug and extentibility


lifeart:
Looks like scam

PR #19106, opened 2020-08-25
[BUGFIX lts] Ensure `destroy` methods on `CoreObject` are invoked.

rwjblue:
When an instance of `CoreObject` is destroyed with `@ember/destroyable` (or the polyfill when using Ember 3.20+), the instances own `destroy` method is never called. In _general_ "this is fine" because the majority of consumers are using `willDestroy` for their custom object teardown needs. Unfortunately, this isn't _quite_ good enough and any consumers that are directly relying on extending from `CoreObject` and having a custom `destroy` method called will no longer function.

The prime example of this is the `Owner` instance itself. Currently, running `destroy(owner)` **does not** actually destroy any instantiated objects in the container. This is because `ContainerProxyMixin` implements `destroy` in order to call `this.__container__.destroy()` which is where the actual cleanup happens. That whole system should be refactored to leveraging `@ember/destroyable` system directly, but in the meantime this case (and any others where folks have implemented a destroy method in Mixins or other subclasses of Ember.CoreObject) should **absolutely** be fixed.

This fix ensures that any `destroy(someInstance)` calls on a `CoreObject` directly invoke the `destroy` method on the instance in the same manner as Ember < 3.20 (`.destroy()` is called eagerly, before `willDestroy`).

### DIFF
--- a/packages/@ember/-internals/runtime/lib/system/core_object.js
+++ b/packages/@ember/-internals/runtime/lib/system/core_object.js
@@ -39,6 +39,14 @@ const prototypeMixinMap = new WeakMap();
 
 const initCalled = DEBUG ? new WeakSet() : undefined; // only used in debug builds to enable the proxy trap
 
+const destroyCalled = new Set();
+
+function ensureDestroyCalled(instance) {
+  if (!destroyCalled.has(instance)) {
+    instance.destroy();
+  }
+}
+
 function initialize(obj, properties) {
   let m = meta(obj);
 
@@ -256,6 +264,7 @@ class CoreObject {
       });
     }
 
+    registerDestructor(self, ensureDestroyCalled, true);
     registerDestructor(self, () => self.willDestroy());
 
     // disable chains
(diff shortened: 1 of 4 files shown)


packages/@ember/-internals/runtime/lib/system/core_object.js
```diff
@@ -256,6 +256,7 @@ class CoreObject {
       });
     }
 
+    registerDestructor(self, () => self.destroy(), true);
```
rwjblue, line 259: Note that this is specifically an eager destructor to match the previous semantics. The `.destroy` method would always have gotten called **before** the downstream objects are marked as `isDestroying` and **before** the instances own `.willDestroy` is called.

rwjblue:
> That whole system should be refactored to leveraging @ember/destroyable system directly

FWIW, I am working on another PR (that won't be `[BUGFIX lts]`) to do this.

packages/@ember/-internals/runtime/lib/system/core_object.js
```diff
@@ -256,6 +266,7 @@ class CoreObject {
       });
     }
 
+    registerDestructor(self, ensureDestroyCalled, true);
```
rwjblue, line 269: Note that this is specifically an eager destructor to match the previous semantics. The `.destroy` method would always have gotten called **before** the downstream objects are marked as `isDestroying` and **before** the instances own `.willDestroy` is called.

krisselden (approved):

PR #19107, opened 2020-08-25
[REFACTOR] Monomorphic Reference

pzuraq:
Integrates https://github.com/glimmerjs/glimmer-vm/pull/1142


packages/@ember/-internals/glimmer/index.ts
```diff
@@ -400,7 +400,6 @@ export { default as AbstractComponentManager } from './lib/component-managers/ab
 // TODO just test these through public API
 // a lot of these are testing how a problem was solved
 // rather than the problem was solved
-export { INVOKE } from './lib/helpers/mut';
```
rwjblue, line 403: This is used as intimate API by a few addons, most notably `ember-concurrency`.

https://github.com/machty/ember-concurrency/blob/067fb79524b69ba4118765bb709ef93aec4800bc/addon/utils.js#L75-L90

We need a plan for how to deal with those downstream issues.

rwjblue (changes requested):

packages/@ember/-internals/glimmer/lib/component-managers/curly.ts
```diff
@@ -381,32 +392,32 @@ export default class CurlyComponentManager
...
-      operations.setAttribute('id', PrimitiveReference.create(id), false, null);
+      operations.setAttribute('id', createPrimitiveRef(id), false, null);
       if (EMBER_COMPONENT_IS_VISIBLE) {
         installIsVisibleBinding!(rootRef, operations);
       }
     }
 
     if (classRef) {
-      const ref = new SimpleClassNameBindingReference(classRef, classRef['propertyKey']);
+      const ref = createSimpleClassNameBindingRef(classRef);
```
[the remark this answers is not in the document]
pzuraq: that was not used for debug output, it was used for the binding behavior we discussed, captured in this test: https://github.com/emberjs/ember.js/pull/19107/files#diff-878fb3529c8ad9603bd82036fca2293dL462

packages/@ember/-internals/glimmer/lib/helpers/if-unless.ts
```diff
@@ -3,40 +3,10 @@
 */
 
 import { assert } from '@ember/debug';
-import { CapturedArguments, VMArguments } from '@glimmer/interfaces';
-import { HelperRootReference } from '@glimmer/reference';
+import { VMArguments } from '@glimmer/interfaces';
+import { createComputeRef, valueForRef } from '@glimmer/reference';
 import toBool from '../utils/to-bool';
```
rwjblue, line 8: semi unrelated question, why do we import `toBool` from here? I would expect for us to _always_ use `import { toBool } from '@glimmer/global-context'`

packages/@ember/-internals/glimmer/lib/helpers/each-in.ts
```diff
@@ -173,17 +173,5 @@ class EachInReference implements PathReference {
...
-  }
-}
-
-export class EachInWrapper {
-  constructor(public inner: unknown) {}
-}
-
-export default function(args: VMArguments) {
-  return new EachInReference(args.positional.at(0));
+  });
```
rwjblue, line 176: Can we give this a label?

packages/@ember/-internals/glimmer/lib/helpers/fn.ts
```diff
@@ -80,38 +79,41 @@ const context = buildUntouchableThis('`fn` helper');
...
-      return (fn as Function).call(context, ...args, ...invocationArgs);
-    }
-  };
-}
-
 export default function(args: VMArguments) {
-  return new HelperRootReference(fn, args.capture());
+  let positional = args.positional.capture();
+
+  return createComputeRef(
```
rwjblue, line 85: Can we add a debugLabel to this one?

packages/@ember/-internals/glimmer/lib/helpers/mut.ts
```diff
@@ -1,10 +1,10 @@
 /**
 @module ember
 */
-import { symbol } from '@ember/-internals/utils';
 import { assert } from '@ember/debug';
 import { VMArguments } from '@glimmer/interfaces';
-import { PathReference, UPDATE_REFERENCED_VALUE } from '@glimmer/reference';
+import { createInvokableRef, isUpdatableRef } from '@glimmer/reference';
+import { _WeakSet } from '@glimmer/util';
```
rwjblue, line 7: Hmm, this seems unused? How is linting happy?

packages/@ember/-internals/glimmer/lib/helpers/mut.ts
```diff
@@ -79,61 +79,11 @@ import { PathReference, UPDATE_REFERENCED_VALUE } from '@glimmer/reference';
...
-  //
-  // 1. (mut "not a path") – passing a literal, result from a helper
-  //    invocation, etc
-  //
-  // 2. (mut receivedValue) – passing a value received from the caller
-  //    that was originally derived from a literal, result from a helper
-  //    invocation, etc
-  //
-  // This message is alright for the first case, but could be quite
-  // confusing for the second case.
```
rwjblue, line 135: I'd like to keep this comment

packages/@ember/-internals/glimmer/lib/syntax/outlet.ts
```diff
@@ -66,87 +71,48 @@ export function outletHelper(args: VMArguments, vm: VM) {
...
-    if (state === undefined) {
-      return undefined;
-    }
+      if (state !== null) {
+        let named = dict<Reference>();
+        named.model = childRefFromParts(outletRef, ['render', 'model']);
 
-    let { render } = state;
+        if (DEBUG) {
+          named.model.debugLabel = '@model';
```
rwjblue, line 100: Should this use the `createAliasReference` (or wahtever it is called)?

packages/@ember/-internals/glimmer/tests/integration/application/debug-render-tree-test.ts
```diff
@@ -129,53 +129,53 @@ if (ENV._DEBUG_RENDER_TREE) {
...
+        //         name: 'foo.inner',
+        //         args: { positional: [], named: { model: 'zomg' } },
+        //         instance: this.controllerFor('foo.inner'),
+        //         template: 'my-app/templates/foo/inner.hbs',
+        //         bounds: this.nodeBounds(this.element.lastChild),
+        //         children: [],
+        //       }),
+        //     ],
+        //   }),
+        // ]);
```
rwjblue, line 178: 😱 Plz make 😄

packages/@ember/-internals/glimmer/tests/integration/components/curly-components-test.js
```diff
@@ -459,44 +459,6 @@ moduleFor(
       });
     }
 
-    ['@test should apply classes of the dasherized property name when bound property specified is true']() {
```
rwjblue, line 462: We chatted a bit about this in DM's, would you mind summarizing the result of our conversation (RE: why this is fine to remove)?

packages/@ember/-internals/metal/lib/property_get.ts
```diff
@@ -125,15 +126,25 @@ export function _getProp(obj: object, keyName: string) {
       } else {
         value = (obj as MaybeHasUnknownProperty).unknownProperty!(keyName);
       }
-    }
 
-    if (isTracking()) {
-      consumeTag(tagFor(obj, keyName));
+      if (isTracking()) {
+        consumeTag(tagForProperty(obj, keyName));
```
rwjblue, line 131: Moving to `tagForProperty` here seems a little bit unrelated to a glimmer-vm update, can you explain the motivation (and point out the previously failing test case that this resolves)?

rwjblue (changes requested):

packages/@ember/-internals/glimmer/lib/helpers/array.ts
```diff
@@ -42,6 +42,8 @@ import { ReifyPositionalReference } from '@glimmer/runtime';
    @public
  */
 
-export default function(args: VMArguments): PathReference {
-  return new ReifyPositionalReference(args.positional.capture());
+export default function(args: VMArguments): Reference<unknown[]> {
+  let captured = args.positional.capture();
+
+  return createComputeRef(() => reifyPositional(captured), null, 'array');
```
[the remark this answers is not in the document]
pzuraq: This is because `action` returns an UnboundReference, which is unusual. Most helpers return a `ComputeRef`, which adds the `result of a...` portion in Glimmer VM. I did it this way so that we wouldn't be repeating `result of a...` for every single helper, which would add byte size and maintenance cost.

Agreed that we should come up with a better system in the future though, I do think the way that helpers are defined should be more consistent.

packages/@ember/-internals/glimmer/lib/helpers/each-in.ts
pzuraq: This helper will never be directly referenced by a user, since it is added via a transform. The only way to reference a helper directly, FWIW, is to do:

```hbs
{{#each (helper) as |helperResult|}}
  {{helperResult.foo}}
{{/each}}
```

So my rule of thumb is, only add labels to helpers that users can actually create themselves and do this with.

packages/@ember/-internals/glimmer/lib/helpers/fn.ts
pzuraq: This one does have a debug label.

packages/@ember/-internals/glimmer/lib/helpers/fn.ts
```diff
@@ -80,38 +79,41 @@ const context = buildUntouchableThis('`fn` helper');
...
+
+      return (...invocationArgs: unknown[]) => {
+        let [fn, ...args] = reifyPositional(positional);
+
+        if (isInvokableRef(callbackRef)) {
+          let value = args[0] || invocationArgs[0];
+          return updateRef(callbackRef, value);
+        } else {
+          return (fn as Function).call(context, ...args, ...invocationArgs);
+        }
```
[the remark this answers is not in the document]
pzuraq: That assertion asserts the callback is a function if it's not invokable, it doesn't throw always. The else would be triggered for every reference that was not wrapped with `{{mut}}`

packages/@ember/-internals/glimmer/lib/helpers/get.ts
```diff
@@ -90,55 +95,39 @@ import { referenceFromParts } from '../utils/references';
...
 export default function(args: VMArguments) {
-  let sourceReference = args.positional.at(0);
-  let pathReference = args.positional.at(1);
+  let sourceRef = args.positional.at(0);
+  let pathRef = args.positional.at(1);
 
-  let path = pathReference.value();
-
-  if (pathReference.isConst()) {
+  if (isConstRef(pathRef)) {
```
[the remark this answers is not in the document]
pzuraq: So, not at this point, because we have no guarantee that `pathRef` and `sourceRef` have been executed at least once. If they are primitives (e.g. `{{get "foo" "length"}}`) then they will be, but that's a fairly rare case. The `pathRef` being a primitive is much more common though, which is why we check it.

Basically, primitives are guaranteed to be constant at first, and Compute refs need to be calculated at least once. We don't want to pull on a ref until it's actually going to be used, so there are cases where it doesn't make sense to check.

packages/@ember/-internals/metal/lib/property_get.ts
pzuraq: Ah, I committed this by mistake. My bad.

packages/@ember/-internals/glimmer/tests/integration/components/curly-components-test.js
pzuraq: Sure thing. Basically, the exact behavior here seems to be an "accidental feature". We traced it back to when it was first implemented, a long time ago, and it popped up after a number of refactors to the ways that bindings worked. This functionality was never documented, and we haven't found any users of it in existing applications, so we believe it is safe to remove.

packages/@ember/-internals/glimmer/lib/helpers/action.ts
```diff
@@ -412,3 +438,9 @@ function makeClosureAction(
     });
   };
 }
+
+// TODO: HAX this is cause the above code assumes that ref will be an object with
+// methods, which is not true after the recent refactors
+function invokeRef(this: Reference, value: unknown) {
+  updateRef(this, value);
+}
```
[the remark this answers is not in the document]
pzuraq: I think this actually is the best way to solve this issue, without rewriting a lot of the code. The code above:

1. Finds an action function, usually on the `actions` hash
2. Calls it with the target as the correct `this` context

Previously, `UPDATE_REFERENCED_VALUE` was a method on the reference itself, so this made a bit more sense. Now, it isn't, and so we need to create a function that can have `this` bound to it when called. This allows us to use the same codepath to call `updateRef` on the reference.

It's a bit hacky, but given this is legacy code we don't really want to rewrite, I think it's probably fine for now. Will update the comment to be a bit more explanatory.

packages/@ember/-internals/glimmer/lib/helpers/if-unless.ts
pzuraq: 👍 converted to using that instead

packages/@ember/-internals/glimmer/lib/helpers/mut.ts
pzuraq: Unsure, removed the import though. Could be because it has an underscore in front of it?

packages/@ember/-internals/glimmer/lib/helpers/mut.ts
pzuraq: 👍

packages/@ember/-internals/glimmer/lib/syntax/outlet.ts
pzuraq: Yup, that works!

packages/@ember/-internals/glimmer/lib/helpers/each-in.ts
rwjblue: Hmm, I don't understand. Isn't this talking about `{{#each-in` which is public API?

packages/@ember/-internals/glimmer/lib/helpers/mut.ts
rwjblue: Ya, possibly due to the underscoring.

rwjblue (approved):
Looks like linting is angry, but I'm otherwise 👍.

PR #19111, opened 2020-08-27
[BUGFIX lts] Fixes ArrayProxy length reactivity

### LINKED ISSUE #19105: [Bug] Ember 3.20.4 Does not Update When ED related records are removed
### 🐞 Describe the Bug
When records are removed from an ember data relationship the change is not being reflected in a template.

### 🔬 Minimal Reproduction
Reproduced in https://github.com/jrjohnson/reproduction-template-model-relationships
Steps: 
1) `npm start`
2) Visit http://localhost:4200/tests?devmode
3) Open Console
4) Clicking Add POSTs a new record and re-computes the count and then `{{log}}` output.
5) Clicking Clear PATCHs and removes the records, but does not update the count of the `{{log}}`.

### 😕 Actual Behavior
Removing all records from an Ember Data relationship array does not update the template.

### 🤔 Expected Behavior
Changing this array should always change the template.

### 🌍 Environment

- Ember: 3.20.4
- Node.js/npm: v12.18.2 / 6.14.6
- OS: OSX
- Browser: FF/Safari/Chrome

pzuraq:
Recent refactors for performance left a gap in the reactivity model,
specifically for the `length` property of ArrayProxy. We now only use
the `CUSTOM_TAG_FOR` API when getting chain tags, which is much better
for most `get()` usages since we don't have to do an extra brand check
that is not very commonly used. However, the `length` property of
ArrayProxy does not have a normal property tag, and it is a native
getter which does not entangle the value naturally when accessed via
autotracking.

This PR sets up the `length` tag eagerly whenever the content of the
ArrayProxy changes, and consumes it when the `length` property is
accessed. It also does the same thing for the `arrTag`, though that is
only used for chain tags. Normally, when the array tag is accessed, it
tracks things properly through the `objectAt` API.

Fixes #19105

### DIFF
--- a/packages/@ember/-internals/runtime/lib/system/array_proxy.js
+++ b/packages/@ember/-internals/runtime/lib/system/array_proxy.js
@@ -18,7 +18,7 @@ import { isObject } from '@ember/-internals/utils';
 import EmberObject from './object';
 import { isArray, MutableArray } from '../mixins/array';
 import { assert } from '@ember/debug';
-import { combine, validateTag, valueForTag, tagFor } from '@glimmer/validator';
+import { combine, consumeTag, validateTag, valueForTag, tagFor } from '@glimmer/validator';
 
 const ARRAY_OBSERVER_MAPPING = {
   willChange: '_arrangedContentArrayWillChange',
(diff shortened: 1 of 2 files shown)


rwjblue (approved):

wuarmin:
@rwjblue is that fix included in Ember 3.20 LTS?

Issue #18912, opened 2020-04-23
runInDebug() not stripped in production


mydea:
I am using ember-source 3.18.0, and just noticed that `runInDebug` is not stripped from production builds.

This is my code:

```js
import Route from '@ember/routing/route';
import { runInDebug } from '@ember/debug';

export default class ApplicationRoute extends Route {
  beforeModel() {
    runInDebug(() => {
      // do something
    });
    // ...
  }
}
```

I noticed on our CI that the build size increased whenever I added something inside of the `runInDebug` method. I then ran the build locally (`ember b --prod`) and checked the generated app.js file, and indeed the whole method with all the content was still there.

Note: It is not actually _run_ it production, but the code is still in the bundle.

mydea:
Hmm, I tried to make a minimal reproduction, but there it seemed to work... I will try to find out why it is not working in our actual app. Maybe anybody has any ideas there?

mydea:
I made a minimal reproduction here:

https://github.com/mydea/ember-issue-reproductions/tree/run-in-debug

Steps to reproduce:

* Run `ember b --prod`
* Go to `dist/assets/ember-issue-reproductions.js`
* Search for `runInDebug`, see that the whole function body is still in there

I also tried it all the way down to 3.8, with no change.

pzuraq:
Hmm, this function was added in 1.5 and isn't the modern way of doing this. We currently recommend that you use the `DEBUG` flag, which is more flexible:

```js
import Route from '@ember/routing/route';
import { DEBUG } from '@glimmer/env';

export default class ApplicationRoute extends Route {
  beforeModel() {
    if (DEBUG) {
      // do something
    }
    // ...
  }
}
```

This is more inline with the wider ecosystem (which uses `NODE_ENV` in general). The way it works is it gets turned to `true`/`false` depending on environment, and then Terser/Uglify remove it because they recognize it as dead code.

We do document the `runInDebug` function though and say that it is stripped in production, the docs may need to be updated. I'll look into this.

pzuraq:
For reference, the debug stripping/tooling that is definitely supported is documented in ember-cli-babel's README: https://github.com/babel/ember-cli-babel#debug-tooling

mydea:
Yeah, I was actually looking around in the docs searching for exactly this, then stumbled over `runInDebug()` and though that this would work as well.

[...]

PR #19126, opened 2020-09-03
Update documentation for `runInDebug`

mydea:
The documentation was incorrect, see: 
https://github.com/emberjs/ember.js/issues/18912

It is not really a "recommended" API anymore, but the documentation is simply incorrect as of now.


rwjblue (approved):
Thank you!

PR #19133, opened 2020-09-08
New options for deprecate() function

mehulkar:
This implements the `for` and `since` options for the `deprecate()` function as detailed in github.com/emberjs/rfcs/pull/649 Not sure if it's the right place to start, but seemed like the lowest hanging fruit. 

cc @rwjblue @pzuraq

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/component.ts
+++ b/packages/@ember/-internals/glimmer/lib/component.ts
@@ -752,6 +752,10 @@ const Component = CoreView.extend(
           id: 'ember-views.event-dispatcher.mouseenter-leave-move',
           until: '4.0.0',
           url: 'https://emberjs.com/deprecations/v3.x#toc_component-mouseenter-leave-move',
+          for: 'ember-source',
+          since: {
+            enabled: '3.13.0-beta.1',
+          },
         }
       );
       deprecate(
@@ -761,6 +765,10 @@ const Component = CoreView.extend(
           id: 'ember-views.event-dispatcher.mouseenter-leave-move',
           until: '4.0.0',
           url: 'https://emberjs.com/deprecations/v3.x#toc_component-mouseenter-leave-move',
+          for: 'ember-source',
+          since: {
+            enabled: '3.13.0-beta.1',
+          },
         }
       );
       deprecate(
@@ -770,6 +778,10 @@ const Component = CoreView.extend(
           id: 'ember-views.event-dispatcher.mouseenter-leave-move',
           until: '4.0.0',
           url: 'https://emberjs.com/deprecations/v3.x#toc_component-mouseenter-leave-move',
+          for: 'ember-source',
+          since: {
+            enabled: '3.13.0-beta.1',
+          },
         }
       );
     },
--- a/packages/@ember/-internals/glimmer/lib/environment.ts
+++ b/packages/@ember/-internals/glimmer/lib/environment.ts
@@ -64,6 +64,10 @@ if (DEBUG) {
       deprecate(message, false, {
         id: 'autotracking.mutation-after-consumption',
         until: '4.0.0',
+        for: 'ember-source',
+        since: {
+          enabled: '3.21.0',
+        },
       });
     },
 
--- a/packages/@ember/-internals/glimmer/lib/helpers/action.ts
+++ b/packages/@ember/-internals/glimmer/lib/helpers/action.ts
@@ -415,6 +415,10 @@ function makeClosureAction(
       {
         until: '3.25.0',
         id: 'actions.custom-invoke-invokable',
+        for: 'ember-source',
+        since: {
+          enabled: '3.23.0-beta.1',
+        },
       }
     );
 
--- a/packages/@ember/-internals/glimmer/lib/modifiers/action.ts
+++ b/packages/@ember/-internals/glimmer/lib/modifiers/action.ts
@@ -156,6 +156,10 @@ export class ActionState {
           {
             until: '3.25.0',
             id: 'actions.custom-invoke-invokable',
+            for: 'ember-source',
+            since: {
+              enabled: '3.23.0-beta.1',
+            },
           }
         );
 
@@ -226,6 +230,10 @@ export default class ActionModifierManager implements ModifierManager<ActionStat
         id: 'ember-views.event-dispatcher.mouseenter-leave-move',
         until: '4.0.0',
         url: 'https://emberjs.com/deprecations/v3.x#toc_action-mouseenter-leave-move',
+        for: 'ember-source',
+        since: {
+          enabled: '3.13.0-beta.1',
+        },
       }
     );
 
--- a/packages/@ember/-internals/metal/lib/chain-tags.ts
+++ b/packages/@ember/-internals/metal/lib/chain-tags.ts
@@ -109,6 +109,10 @@ function getChainTags(
         {
           until: '3.17.0',
           id: 'ember-metal.computed-deep-each',
+          for: 'ember-source',
+          since: {
+            enabled: '3.13.0-beta.3',
+          },
         }
       );
 
--- a/packages/@ember/debug/lib/deprecate.ts
+++ b/packages/@ember/debug/lib/deprecate.ts
@@ -10,13 +10,18 @@ declare global {
   };
 }
 
+export type DeprecationStages = 'available' | 'enabled';
+
 export interface DeprecationOptions {
   id: string;
   until: string;
   url?: string;
+  for: string;
+  since: Partial<Record<DeprecationStages, string>>;
 }
 
 export type DeprecateFunc = (message: string, test?: boolean, options?: DeprecationOptions) => void;
+export type MissingOptionDeprecateFunc = (id: string) => string;
 
 /**
  @module @ember/debug
@@ -64,7 +69,11 @@ let registerHandler: (handler: HandlerCallback) => void = () => {};
 let missingOptionsDeprecation: string;
 let missingOptionsIdDeprecation: string;
 let missingOptionsUntilDeprecation: string;
+let missingOptionsForDeprecation: MissingOptionDeprecateFunc = () => '';
+let missingOptionsSinceDeprecation: MissingOptionDeprecateFunc = () => '';
 let deprecate: DeprecateFunc = () => {};
+let FOR_MISSING_DEPRECATIONS = new Set();
+let SINCE_MISSING_DEPRECATIONS = new Set();
 
 if (DEBUG) {
   registerHandler = function registerHandler(handler: HandlerCallback) {
@@ -154,6 +163,13 @@ if (DEBUG) {
     '`options` should include `id` and `until` properties.';
   missingOptionsIdDeprecation = 'When calling `deprecate` you must provide `id` in options.';
   missingOptionsUntilDeprecation = 'When calling `deprecate` you must provide `until` in options.';
+
+  missingOptionsForDeprecation = (id: string) => {
+    return `When calling \`deprecate\` you must provide \`for\` in options. Missing options.for in "${id}" deprecation`;
+  };
+  missingOptionsSinceDeprecation = (id: string) => {
+    return `When calling \`deprecate\` you must provide \`since\` in options. Missing options.since in "${id}" deprecation`;
+  };
   /**
    @module @ember/debug
    @public
@@ -176,8 +192,10 @@ if (DEBUG) {
       "view.helper.select".
     @param {string} options.until The version of Ember when this deprecation
       warning will be removed.
+    @param {String} options.for A namespace for the deprecation, usually the package name
+    @param {Object} options.since Describes when the deprecation became available and enabled.
     @param {String} [options.url] An optional url to the transition guide on the
-      emberjs.com website.
+          emberjs.com website.
     @static
     @public
     @since 1.0.0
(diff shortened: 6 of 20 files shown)


packages/@ember/debug/lib/deprecate.ts
```diff
@@ -14,6 +14,7 @@ export interface DeprecationOptions {
   id: string;
   until: string;
   url?: string;
+  for: string;
```
rwjblue, line 17: I think we also need `since`

mehulkar, line 17: add since here

packages/@ember/debug/lib/deprecate.ts
```diff
@@ -187,6 +192,28 @@ if (DEBUG) {
...
+    if (!options.for) {
+      deprecate(missingOptionsForDeprecation, Boolean(options!.for), {
+        id: 'ember-source.deprecation-without-for',
+        until: '4.0.0',
+        for: 'ember-source',
+        since: {
+          available: '3.22.0',
+        },
+      });
+    }
```
rwjblue, line 204: Based on our convo today, let's wrap these in another guard to ensure we only issue a deprecation once for missing `for` for a specific `id`. I'm thinking something like:

```js
    if (!options.for && !FOR_MISSING_DEPRECATIONS.has(options.id)) {
      FOR_MISSING_DEPRECATION.add(options.id);
      
      deprecate(missingOptionsForDeprecation, Boolean(options!.for), {
        id: 'ember-source.deprecation-without-for',
        until: '4.0.0',
        for: 'ember-source',
        since: {
          available: '3.22.0',
        },
      });
    }
```

This avoids issuing the `for` missing deprecation for **every** invocation (which could be a massive number).

rwjblue: Also, we should ensure that we include information on the original `options.id` in the deprecation message. At the moment, I think we are using a static `missingOptionsForDeprecation` message, but I'd like to see the original deprecation `id` in the message.

packages/@ember/debug/lib/deprecate.ts
```diff
@@ -187,6 +192,28 @@ if (DEBUG) {
...
+    if (!options.since) {
+      deprecate(missingOptionsSinceDeprecation, Boolean(options!.since), {
+        id: 'ember-source.deprecation-without-since',
+        until: '4.0.0',
+        for: 'ember-source',
+        since: {
+          available: '3.22.0',
+        },
+      });
+    }
```
rwjblue, line 215: Same here, RE: the two above points:

* use a debug only `Set` to avoid issuing the same deprecation many many times
* include information about the original deprecation id that is triggering the "knock on" deprecation

rwjblue (changes requested):
Thanks for working on this @mehulkar!

We should:

* Add some deprecation flagging around the new deprecations (see [`8b7d402` (#18436)](https://github.com/emberjs/ember.js/pull/18436/commits/8b7d402426eb40b3dfc2caea64547adbd1b236e6) as an example)
* Update all existing callsites of `deprecate` to pass the new options (Ember itself should not trigger this deprecation)
* Address the inline comments

packages/@ember/debug/lib/deprecate.ts
mehulkar: ✅

mehulkar:
> Add some deprecation flagging around the new deprecations (see 8b7d402 (#18436) as an example)

@rwjblue bit confused about what to do here. Is the "deprecated feature" here the ability to call `deprecate()` without `for`/`since`? In that case, what exactly would be svelte'd? Just the checks for it? Or would the type definitions all have to be guarded inside the flag? What about the call sites?

packages/@ember/-internals/glimmer/lib/component.ts
```diff
@@ -752,6 +752,10 @@ const Component = CoreView.extend(
           id: 'ember-views.event-dispatcher.mouseenter-leave-move',
           until: '4.0.0',
           url: 'https://emberjs.com/deprecations/v3.x#toc_component-mouseenter-leave-move',
+          for: 'ember-source',
+          since: {
+            enabled: '3.21.0',
```
rwjblue, line 757: I think the "real" enabled versions are in here:

https://github.com/emberjs/ember.js/blob/master/packages/%40ember/deprecated-features/index.ts#L6-L19

Can you update each of these to match those versions?

packages/@ember/-internals/glimmer/lib/environment.ts
```diff
@@ -64,6 +64,10 @@ if (DEBUG) {
       deprecate(message, false, {
         id: 'autotracking.mutation-after-consumption',
         until: '4.0.0',
+        for: 'ember-source',
+        since: {
+          enabled: '3.21.0',
```
mehulkar, line 69: https://github.com/emberjs/ember.js/commit/2bf857e3af0d87c08bc28c6a09ccfa902f15b615

packages/@ember/-internals/glimmer/lib/helpers/action.ts
```diff
@@ -415,6 +415,10 @@ function makeClosureAction(
       {
         until: '3.25.0',
         id: 'actions.custom-invoke-invokable',
+        for: 'ember-source',
+        since: {
+          enabled: '3.21.0',
```
mehulkar, line 420: not sure if this is different from using `{{action}}` has a helper in templates?

packages/@ember/-internals/glimmer/lib/modifiers/action.ts
```diff
@@ -156,6 +156,10 @@ export class ActionState {
           {
             until: '3.25.0',
             id: 'actions.custom-invoke-invokable',
+            for: 'ember-source',
+            since: {
+              enabled: '3.21.0',
```
mehulkar, line 161: not sure what this should be

packages/@ember/-internals/glimmer/lib/helpers/action.ts
rwjblue: Ya, it is different, but it was never exposed as public API. This is a private API deprecation specifically added so that we don't break a private API usages within ember-concurrency.

packages/@ember/-internals/glimmer/lib/modifiers/action.ts
rwjblue: 3.23.0-beta.1

packages/@ember/-internals/glimmer/lib/helpers/action.ts
rwjblue: Should be 3.23.0-beta.1

packages/@ember/-internals/metal/lib/chain-tags.ts
```diff
@@ -109,6 +109,10 @@ function getChainTags(
         {
           until: '3.17.0',
           id: 'ember-metal.computed-deep-each',
+          for: 'ember-source',
+          since: {
+            enabled: '3.21.0',
```
rwjblue, line 114: 3.13.0-beta.3

packages/@ember/debug/lib/deprecate.ts
```diff
@@ -187,6 +203,32 @@ if (DEBUG) {
...
 
+    if (!options!.for && !FOR_MISSING_DEPRECATIONS.has(options!.id)) {
+      FOR_MISSING_DEPRECATIONS.add(options!.id);
+
+      deprecate(missingOptionsForDeprecation(options!.id), Boolean(options!.for), {
+        id: 'ember-source.deprecation-without-for',
+        until: '4.0.0',
+        for: 'ember-source',
+        since: {
+          available: '3.22.0',
```
rwjblue, line 214:
```suggestion
          available: '3.24.0',
```

packages/@ember/debug/lib/deprecate.ts
```diff
@@ -187,6 +203,32 @@ if (DEBUG) {
...
+
+    if (!options!.since && !SINCE_MISSING_DEPRECATIONS.has(options!.id)) {
+      SINCE_MISSING_DEPRECATIONS.add(options!.id);
+
+      deprecate(missingOptionsSinceDeprecation(options!.id), Boolean(options!.since), {
+        id: 'ember-source.deprecation-without-since',
+        until: '4.0.0',
+        for: 'ember-source',
+        since: {
+          available: '3.22.0',
```
rwjblue, line 227:
```suggestion
          available: '3.24.0',
```

packages/@ember/debug/lib/deprecate.ts
```diff
@@ -14,9 +14,12 @@ export interface DeprecationOptions {
   id: string;
   until: string;
   url?: string;
+  for: string;
+  since: object;
```
rwjblue, line 18: I think we want this to be:

```ts
export type DeprecationStages = 'available' | 'enabled';
export interface DeprecateOptions {
  // ...snip...
  since: Record<DeprecationStages, string>;
}
```

packages/@ember/debug/lib/deprecate.ts
```diff
@@ -170,13 +184,15 @@ if (DEBUG) {
     @param {String} message A description of the deprecation.
     @param {Boolean} test A boolean. If falsy, the deprecation will be displayed.
     @param {Object} options
-    @param {String} options.id A unique id for this deprecation. The id can be
+    @param {String} [options.id] A unique id for this deprecation. The id can be
```
rwjblue, line 187: I think the brackets here infer that the value is optional, see http://yui.github.io/yuidoc/syntax/index.html#param.

```suggestion
    @param {String} options.id A unique id for this deprecation. The id can be
```

packages/@ember/debug/lib/deprecate.ts
```diff
@@ -170,13 +184,15 @@ if (DEBUG) {
...
       used by Ember debugging tools to change the behavior (raise, log or silence)
       for that specific deprecation. The id should be namespaced by dots, e.g.
       "view.helper.select".
-    @param {string} options.until The version of Ember when this deprecation
+    @param {string} [options.until] The version of Ember when this deprecation
```
rwjblue, line 191: I think the brackets here infer that the value is optional, see http://yui.github.io/yuidoc/syntax/index.html#param.

```suggestion
    @param {string} options.until The version of Ember when this deprecation
```

packages/@ember/debug/lib/deprecate.ts
```diff
@@ -170,13 +184,15 @@ if (DEBUG) {
...
       warning will be removed.
     @param {String} [options.url] An optional url to the transition guide on the
+    @param {String} [options.for] A namespace for the deprecation, usually the package name
+    @param {Object} [options.since] Describes when the deprecation became available and enabled.
       emberjs.com website.
```
rwjblue, line 196: The trailing "emberjs.com website" bit went with `options.url`, and `for`/`since` are not optional.

```suggestion
    @param {String} options.for A namespace for the deprecation, usually the package name
    @param {Object} options.since Describes when the deprecation became available and enabled.
    @param {String} [options.url] An optional url to the transition guide on the
          emberjs.com website.
```

packages/@ember/debug/tests/main_test.js
```diff
@@ -273,6 +328,77 @@ moduleFor(
...
 
+    ['@test deprecate without options.for triggers an assertion'](assert) {
+      assert.expect(1);
+
+      assert.throws(
+        () => deprecate('message1', false, { id: 'test', until: 'forever' }),
+        new RegExp(missingOptionsForDeprecation('test')),
+        'proper assertion is triggered when options.for is missing'
+      );
+    }
```
rwjblue, line 339: Unlike the other options, these don't throw (they issue deprecations).

```suggestion
    ['@test deprecate without options.for triggers a deprecation'](assert) {
      assert.expect(1);

      expectDeprecation(
        () => deprecate('message1', false, { id: 'test', until: 'forever' }),
        new RegExp(missingOptionsForDeprecation('test')),
        'deprecation is triggered when options.for is missing'
      );
    }
```

packages/@ember/debug/tests/main_test.js
```diff
@@ -273,6 +328,77 @@ moduleFor(
...
+        () =>
+          deprecate('foo', false, {
+            id: 'test',
+            until: 'forever',
+            for: 'me',
+          }),
+        new RegExp(missingOptionsSinceDeprecation('test')),
+        'proper assertion is triggered when options.since is missing'
+      );
+    }
```
rwjblue, line 378: Issues a deprecation, not an assertion.

```suggestion
    ['@test deprecate without options.since triggers a deprecation'](assert) {
      assert.expect(1);

      expectDeprecation(
        () =>
          deprecate('foo', false, {
            id: 'test',
            until: 'forever',
            for: 'me',
          }),
        new RegExp(missingOptionsSinceDeprecation('test')),
        'deprecation is triggered when options.since is missing'
      );
    }
```

packages/@ember/debug/lib/deprecate.ts
mehulkar: ugh, so sloppy, sorry.

packages/@ember/debug/tests/main_test.js
mehulkar: hmm, 🤔 , I assumed it was becuase of `ENV.RAISE_ON_DEPRECATION = true` being set in the `beforeEach`?

Do you want to change the actual code after changing this test case also?

mehulkar: changing this test does make it keep passing, but then it seems like we should also change `ENV.RAISE_ON_DEPRECATION = false` in these tests, which leave some ground uncovered when the flag is true.

mehulkar: chatted offline. using `expectDeprecation` is a better pattern, because it handles both the dev and prod builds (in prod, deprecations don't run, so the test would do the right thing there). in this case, it doesn't matter because this test never runs against the prod build (b/c it's a debug specific test), but it's better to follow the same pattern as the rest of the test suite.

packages/@ember/debug/lib/deprecate.ts
```diff
@@ -10,13 +10,18 @@ declare global {
...
 }
 
+export type DeprecationStages = 'available' | 'enabled';
+
 export interface DeprecationOptions {
   id: string;
   until: string;
   url?: string;
+  for: string;
+  since: Partial<Record<DeprecationStages, string>>;
```
mehulkar, line 20: Thanks for the assist @dfreeman!

packages/@ember/debug/tests/main_test.js
mehulkar: @rwjblue I had to revert this. `expectDeprecation` stubs out the `deprecate()` function so it never calls it the second time :/ Not sure what the best path forward is.

rwjblue: Ahh. Gotcha. I think this is probably fine then. It seems that we can't satisfy the consistency constraint without a bunch of hoop jumping, and this package's tests aren't even ran in prod builds (where `expectDeprecation` actually is required).

rwjblue (approved):
Thanks for working on this @mehulkar!

rwjblue:
@mehulkar - Mind squashing this down? Once that is done, we can land...

mehulkar:
Squashed and pushed!

rwjblue:
Thanks again!

PR #19135, opened 2020-09-09
Remove module unification usages & MODULE_UNIFICAITION flag

lifeart:
Since Module Unification support is deprecated, we could remove few lines of code, related to MU blueprints, and MU flag itself, this PR lands such changes.


packages/@ember/-internals/glimmer/lib/resolver.ts
```diff
@@ -422,7 +366,7 @@ export default class RuntimeResolverImpl implements RuntimeResolver<OwnedTemplat
       return null;
     }
 
-    return (args, vm) => {
+    return (args: { capture: () => any }, vm: { associateDestroyable: (arg0: {}) => void }) => {
```
rwjblue, line 369: I think these types can be imported from `@glimmer/runtime` (instead of having to type them manually in the signature), I'm curious why this needed to be changed (did you have warnings or something?)...

packages/@ember/-internals/container/lib/container.ts
```diff
@@ -403,7 +399,7 @@ function instantiateFactory(
   // SomeClass { singleton: true, instantiate: true } | { singleton: true } | { instantiate: true } | {}
   // By default majority of objects fall into this case
   if (isSingletonInstance(container, fullName, options)) {
-    let instance = (container.cache[normalizedName] = factoryManager.create());
+    let instance = (container.cache[normalizedName] = factoryManager.create() as CacheMember);
```
rwjblue, line 402: Why do we need to cast here?

packages/@ember/-internals/container/lib/container.ts
```diff
@@ -223,8 +222,8 @@ export default class Container {
 
     assert('fullName must be a proper full name', this.registry.isValidFullName(normalizedName));
     assert(
-      'EMBER_MODULE_UNIFICATION must be enabled to pass a namespace option to factoryFor',
-      EMBER_MODULE_UNIFICATION || !options.namespace
+      'namespace must be enabled to pass a namespace option to factoryFor',
```
rwjblue, line 225: I think this is kinda wrong. The message here doesn't read quite right. I think the idea here is that we can remove the entire `options.namespace` system here...

rwjblue (changes requested):

packages/@ember/-internals/container/lib/container.ts
lifeart: I have typescript error In vscode about assign unknown property

packages/@ember/-internals/glimmer/lib/resolver.ts
lifeart: after vscode restart I did not get such errors, removing autocased definitions

packages/@ember/-internals/container/lib/container.ts
lifeart: image

packages/@ember/-internals/container/lib/container.ts
lifeart: asserts removed

rwjblue (approved):

rwjblue:
Thanks @lifeart!

Issue #18981, opened 2020-05-18
Ember 3.18 - Cycles in tags are not allowed [query params]


lifeart:
![image](https://user-images.githubusercontent.com/1360552/82253446-aa81e500-9959-11ea-8611-6825f7c0df7b.png)

reproduction: https://ember-twiddle.com/4106c69bb441077e93a8fcbba5fb3620?openFiles=templates.application%5C.hbs%2C

 click on link 1, click on 2, and check console

gabrielgrant:
arg, think i've been running into this the past couple days thinking I was doing something silly. Any suggestions on a workaround?

chancancode:
Great reproduction @lifeart. I narrowed down the reproduction a little and left some comments on what I found, both in the template and in the controller code. This seems to specifically impact array-typed QPs backed by tracked properties, with links that uses the default QP value.

https://ember-twiddle.com/1390a8f5eefe3b9ce755944bf989afcc?openFiles=templates.application%5C.hbs%2C

I think this clearly seems like a bug, but if you are desperate for a work around, since the QP only affect array-typed QPs, I suppose you can do something like this to manually convert it into a string-typed QP to avoid the issue:

```js
// Don't copy this code as-is, without testing, I just typed-out-loud on the issue thread!

class MyController extends Controller {
  queryParams = [{ _projects: 'projects' }];

  @tracked _projects = '';

  get projects() {
    if (this._projects === '') {
      return [];
    } else {
      return JSON.parse(`[${this._projects}]`);
    }
  }

  set projects(array) {
    if (Array.isArray(array) && array.length > 0) {
      this._projects = JSON.stringify(array).slice(1, -1);
    } else {
      this._projects = '';
    }
  }
}
```

arunyegappan:
I face the same issue. Any updates on this?

s-chiriac:
Is there any chance this will get looked at soon? I'd love to contribute, but I'm not familiar enough with the codebase to find a solution. It's preventing my team from upgrading to 3.20

Samsinite:
Looks like a fix is in 3.21, just appears to need back-ported to a few LTS releases.

rwjblue:
Ya, the fix introduced regressions that we are still working on fixing (https://github.com/emberjs/ember.js/pull/19178 fixed some, but introduced https://github.com/emberjs/ember.js/issues/19179 which ATM is still unfixed).

anlumo:
[...]

PR #19138, opened 2020-09-14
[BUGFIX lts] Fix tag cycles in query parameters

pzuraq:
Currently, we wrap query parameters with `@dependentKeyCompat` manually
to ensure they work with the legacy router setup, which watches QPs
using observers. When a QP value is labeled as `@tracked`, we don't need
to replace it with `@dependentKeyCompat`, but since it's just a plain
native getter (from the decorator) we do anyways. This results in a tag
cycle being created, which can result in a negative performance impact,
as every render pass will be invalidated by the cycle and require a
subsequent full revalidation.

This bug would have been caught by our ALLOW_CYCLES logic, but that
logic was faulty - it was allowing too many cycles, anything that was
accessed via `get`. It should only have allowed cycles from computeds
themselves.

This PR fixes the bug with ALLOW_CYCLES, and the subsequent bug with
`@tracked` by only wrapping with `@dependentKeyCompat` if the value is
not `@tracked`. It also adds an assertion to prevent users from using
`@dependentKeyCompat` with `@computed` or `@tracked`.

Verified that the test failed after fixing ALLOW_CYCLES, and passed
again after the fix.

### DIFF
--- a/packages/@ember/object/compat.ts
+++ b/packages/@ember/object/compat.ts
@@ -2,14 +2,20 @@ import { Meta } from '@ember/-internals/meta';
 import {
   Decorator,
   DecoratorPropertyDescriptor,
+  descriptorForProperty,
   isElementDescriptor,
   setClassicDecorator,
 } from '@ember/-internals/metal';
 import { assert } from '@ember/debug';
 import { consumeTag, tagFor, track, UpdatableTag, updateTag } from '@glimmer/validator';
 
-let wrapGetterSetter = function (_target: object, key: string, desc: PropertyDescriptor) {
+let wrapGetterSetter = function (target: object, key: string, desc: PropertyDescriptor) {
   let { get: originalGet } = desc;
+
+  assert(
+    'You attempted to use @dependentKeyCompat on a property that already has been decorated with either @computed or @tracked. @dependentKeyCompat is only necessary for native getters that are not decorated with @computed.',
+    descriptorForProperty(target, key) === undefined
+  );
 
   if (originalGet !== undefined) {
     desc.get = function () {
--- a/packages/@ember/-internals/metal/lib/alias.ts
+++ b/packages/@ember/-internals/metal/lib/alias.ts
@@ -12,7 +12,7 @@ import {
   validateTag,
   valueForTag,
 } from '@glimmer/validator';
-import { finishLazyChains, getChainTagsForKey } from './chain-tags';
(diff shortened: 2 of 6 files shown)


rwjblue:
@pzuraq - Can you review the open issues RE: QP + cycles, and cross link the ones this fixes?

packages/@ember/object/compat.ts
```diff
@@ -2,15 +2,21 @@ import { Meta } from '@ember/-internals/meta';
...
 } from '@ember/-internals/metal';
 import { assert } from '@ember/debug';
 import { consumeTag, tagFor, track, UpdatableTag, updateTag } from '@glimmer/validator';
 
-let wrapGetterSetter = function(_target: object, key: string, desc: PropertyDescriptor) {
+let wrapGetterSetter = function(target: object, key: string, desc: PropertyDescriptor) {
   let { get: originalGet } = desc;
 
+  assert(
+    'You attempted to use @computed on a property that already has been decorated with either @computed or @tracked. @dependentKeyCompat is only necessary for native getters that are not decorated with @computed.',
```
chriskrycho, line 16: @pzuraq do you mean "You attempted to use @dependentKeyCompat..." here?

chancancode:
🙌 https://github.com/emberjs/ember.js/issues/18981

chancancode:
We should probably add a QP test for this?

rwjblue:
> We should probably add a QP test for this?

Yes, absolutely. That was why I was hoping we could cross link the related issues, and create test cases for them...

chancancode:
I think I have a small repro in twiddle, but maybe not small enough for a test case

pzuraq:
Note that there was already a test case, which was passing mistakenly due to the ALLOW_CYCLES logic being faulty. Fixing that logic caused the test to fail, and then the rest of the PR fixed it. Noted this in the original PR description.

@rwjblue did link me to a twiddle with an additional failure that somehow was caused by using the `{{log}}` helper, I added that as a test case as well.

packages/@ember/object/compat.ts
rwjblue: Good catch, fixed!

rwjblue (approved):
Rebased on top of all the prettier/eslint changes that just landed on master, and fixed up the error message issue that @chriskrycho pointed out.

Should be good to land once CI is happy...

Samsinite:
Does this fix the performance issue that was present when query params were used alongside tracked properties?

PR #19142, opened 2020-09-19
[BUGFIX beta] reimplement $.ready()

### LINKED ISSUE #19134: App booted before DOM ready without jQuery
https://github.com/emberjs/ember.js/blob/b049c37e66587718cef38cde63b59b0789505d43/packages/%40ember/application/lib/application.js#L485-L491

Without jQuery, this code no longer works and boots the app immediately (as soon as the bundle is loaded, but possibly before DOM read).

Workaround for now using an initializer:

```js
export function initialize(app) {
  if (document.readyState === 'loading') {
    app.deferReadiness();

    let callback = () => {
      if (document.readyState !== 'loading') {
        app.advanceReadiness();
        document.removeEventListener('readystatechange', callback);
      }
    };

    document.addEventListener('readystatechange', callback);
  }
}

export default {
  initialize
};
```

We should reimplement `$.ready()` without relying on jQuery.

rreckonerr:
## TODOs
- [x] refactor waitForDOMReady away from jQuery
- [x] figure out if `!this.$ === !hasDOM`
- [x] fix failing tests in `packages/@ember/application/tests/readiness_test.js`
  - [x] find a way to restore mocked readyState or a better way to mock it
  - [x] ~improve coverage for advanceReadiness/deferReadiness cases~

## NOTES
- jquery `.ready()` [implementation](https://github.com/jquery/jquery/blob/df6858df2ed3fc5c424591a5e09b900eb4ce0417/src/core/ready.js#L58-L78): 
  ```js
  function completed() {
    document.removeEventListener( "DOMContentLoaded", completed );
    window.removeEventListener( "load", completed );
    jQuery.ready();
  }


  // Catch cases where $(document).ready() is called
  // after the browser event has already occurred.
  if ( document.readyState !== "loading" ) {


    // Handle it asynchronously to allow scripts the opportunity to delay ready
    window.setTimeout( jQuery.ready );


  } else {


    // Use the handy event callback
    document.addEventListener( "DOMContentLoaded", completed );


    // A fallback to window.onload, that will always work
    window.addEventListener( "load", completed );
  }
  ```

### DIFF
--- a/packages/@ember/application/lib/application.js
+++ b/packages/@ember/application/lib/application.js
@@ -7,7 +7,7 @@ import { ENV } from '@ember/-internals/environment';
 import { hasDOM } from '@ember/-internals/browser-environment';
 import { assert, isTesting } from '@ember/debug';
 import { DEBUG } from '@glimmer/env';
-import { bind, join, once, run, schedule } from '@ember/runloop';
+import { join, once, run, schedule } from '@ember/runloop';
 import {
   libraries,
   processAllNamespaces,
@@ -204,6 +204,15 @@ const Application = Engine.extend({
     @public
   */
   rootElement: 'body',
+
+  /**
+
+    @property _document
+    @type Document | null
+    @default 'window.document'
+    @private
+  */
+  _document: hasDOM ? window.document : null,
 
   /**
     The `Ember.EventDispatcher` responsible for delegating events to this
@@ -483,10 +492,15 @@ const Application = Engine.extend({
     @method waitForDOMReady
   */
   waitForDOMReady() {
-    if (!this.$ || this.$.isReady) {
+    if (this._document === null || this._document.readyState !== 'loading') {
       schedule('actions', this, 'domReady');
     } else {
-      this.$().ready(bind(this, 'domReady'));
+      let callback = () => {
+        this._document.removeEventListener('DOMContentLoaded', callback);
+        run(this, 'domReady');
+      };
+
+      this._document.addEventListener('DOMContentLoaded', callback);
     }
   },
 
--- a/packages/@ember/application/tests/readiness_test.js
+++ b/packages/@ember/application/tests/readiness_test.js
@@ -2,52 +2,39 @@ import { moduleFor, ModuleBasedTestResolver, ApplicationTestCase } from 'interna
 import { run } from '@ember/runloop';
 import EmberApplication from '..';
 
-let jQuery, application, Application;
-let readyWasCalled, domReady, readyCallbacks;
+let application, Application, _document, callbacks;
+let readyWasCalled = 0;
 
-// We are using a small mock of jQuery because jQuery is third-party code with
-// very well-defined semantics, and we want to confirm that a jQuery stub run
-// in a more minimal server environment that implements this behavior will be
-// sufficient for Ember's requirements.
+const dispatchEvent = (eventName) => {
+  callbacks[eventName].forEach((callback) => callback());
+};
+
+const removeEventListener = (eventName, callbackToRemove) => {
+  callbacks[eventName] = callbacks[eventName].filter((callback) => callback !== callbackToRemove);
+};
+
+const addEventListener = (eventName, callback) => {
+  callbacks[eventName] ? callbacks[eventName].push(callback) : (callbacks[eventName] = [callback]);
+};
+
 moduleFor(
   'Application readiness',
   class extends ApplicationTestCase {
     constructor() {
       super();
 
-      readyWasCalled = 0;
-      readyCallbacks = [];
-
-      let jQueryInstance = {
-        ready(callback) {
-          readyCallbacks.push(callback);
-          if (jQuery.isReady) {
-            domReady();
-          }
-        },
-      };
-
-      jQuery = function () {
-        return jQueryInstance;
-      };
-      jQuery.isReady = false;
-
-      let domReadyCalled = 0;
-      domReady = function () {
-        if (domReadyCalled !== 0) {
-          return;
-        }
-        domReadyCalled++;
-        for (let i = 0; i < readyCallbacks.length; i++) {
-          readyCallbacks[i]();
-        }
+      callbacks = [];
+      _document = {
+        removeEventListener,
+        addEventListener,
       };
 
       Application = EmberApplication.extend({
-        $: jQuery,
         Resolver: ModuleBasedTestResolver,
+        _document,
 
         ready() {
+          this._super();
           readyWasCalled++;
         },
       });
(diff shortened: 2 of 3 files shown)


rreckonerr:
Fixes #19142

chancancode:
Thanks for looking into this. I talked to @rwjblue about this and he wanted to do some benchmarks to confirm this restoring the old behavior won't cause any regressions. I think he was planning to benchmark it against my workaround, but if this works then they can probably test this directly.

I also realized a problem with the approach here. I think we are not supposed to rely on `document` existing due to fastboot, but that presents a chicken-and-egg problem because I'm not sure how we would get the non-global `document` object to use here, since it's usually passed/injected via the `boot()` method as one of the options.

Maybe it's not a big deal since fastboot disables autoboot and wouldn't hit this code path anyway. @rwjblue/@krisselden do you have any thoughts? Is this just fine or should we at least guard accessing the `document` global and the event handling methods?

packages/@ember/application/lib/application.js
```diff
@@ -483,13 +483,22 @@ const Application = Engine.extend({
     @method waitForDOMReady
   */
   waitForDOMReady() {
-    if (!this.$ || this.$.isReady) {
+    if (!hasDOM || this.documentReadyState() !== 'loading') {
```
rreckonerr, line 486: @chancancode Shouldn't `!hasDom` guard us against accessing `document` in the fastboot environment in addition to `if (this.autoboot) { this.waitForDOMReady() }`?

```js
// packages/@ember/-internals/browser-environment/lib/has-dom.ts

export default typeof self === 'object' &&
  self !== null &&
  (self as any).Object === Object &&
  typeof Window !== 'undefined' &&
  self.constructor === Window &&
  typeof document === 'object' &&
  document !== null &&
  self.document === document &&
  typeof location === 'object' &&
  location !== null &&
  self.location === location &&
  typeof history === 'object' &&
  history !== null &&
  self.history === history &&
  typeof navigator === 'object' &&
  navigator !== null &&
  self.navigator === navigator &&
  typeof navigator.userAgent === 'string';
```

chancancode: @rreckonerr good point! so it's already guarded 👍

packages/@ember/application/lib/application.js
```diff
@@ -483,13 +483,22 @@ const Application = Engine.extend({
...
       schedule('actions', this, 'domReady');
     } else {
-      this.$().ready(bind(this, 'domReady'));
+      let callback = function() {
```
chancancode, line 489: Use an arrow function here, then you don't need `bind(...)`. Otherwise, `bind(...)` will return a different function than the variable in `callback`, and your removal code won't find it.

packages/@ember/application/lib/application.js
chancancode, line 486: IMO we can just inline `document.readyState` here. As you pointed out, `hasDOM` already checked that `document` exists so you don't need to check it again.

rreckonerr: I had to add the `documentReadyState()` method to be able to test it. Couldn't find another way to mock a read-only `readyState` property 🥺

packages/@ember/application/lib/application.js
rreckonerr: Whoops! Already fixed it

packages/@ember/application/lib/application.js
rreckonerr: In fact, I didn't find a way to restore `readyState` after `Object.defineProperty(document, "readyState", ...)`, so it was polluting the global state.

chancancode: Yeah, that's a good point, but in that case we have bigger problems, e.g. we are leaking event handlers on the actual global document object in tests. I'm not sure what's the best approach here, but if we are going to mock, we should just mock the whole document object (the parts that we need), including `addEventListener` and `removeEventListener`. I'll probably pass it in to the constructor/init of `Application` and default it to the global document object.

packages/@ember/application/tests/readiness_test.js
```diff
@@ -110,26 +105,7 @@ moduleFor(
...
-      domReady();
-
-      assert.equal(readyWasCalled, 0, "ready wasn't called yet");
-
-      run(() => {
-        application.advanceReadiness();
-      });
-
-      assert.equal(readyWasCalled, 1, 'ready was called now all readiness deferrals are advanced');
+      assert.equal(callbacks['DOMContentLoaded'].length, 0);
```
rreckonerr, line 108: @chancancode This test case is failing because of event listener added in `waitForDOMReady()`.
Should we remove it in `advanceReadiness()` method to not pollute the global state?

packages/@ember/application/lib/application.js
```diff
@@ -205,6 +205,15 @@ const Application = Engine.extend({
...
   rootElement: 'body',
 
+  /**
+
+    @property rootDocument
+    @type Document
+    @default 'window.document'
+    @public
+  */
+  rootDocument: window.document,
```
chancancode, line 215: I think this has to be `hasDOM ? document : null` or something like that

packages/@ember/application/tests/readiness_test.js
chancancode: Isn't the callback supposed to remove it once it is called?

packages/@ember/application/lib/application.js
```diff
@@ -205,6 +205,15 @@ const Application = Engine.extend({
   */
   rootElement: 'body',
 
+  /**
+
+    @property rootDocument
+    @type Document
+    @default 'window.document'
+    @public
```
chancancode, line 213: @private

chancancode: Wouldn't hurt to make call this field `_document` also to make it clear that it's private

packages/@ember/application/lib/application.js
```diff
@@ -205,6 +205,15 @@ const Application = Engine.extend({
   */
   rootElement: 'body',
 
+  /**
+
+    @property rootDocument
+    @type Document
```
chancancode, line 211: `Document | null`

chancancode: (see below)

packages/@ember/application/lib/application.js
```diff
@@ -483,10 +492,15 @@ const Application = Engine.extend({
     @method waitForDOMReady
   */
   waitForDOMReady() {
-    if (!this.$ || this.$.isReady) {
+    if (!hasDOM || this.rootDocument.readyState !== 'loading') {
```
chancancode, line 495: If the defaulting above already checked `hasDOM`, we can probably just check for `this._document === null` here

packages/@ember/application/tests/readiness_test.js
```diff
@@ -2,52 +2,39 @@ import { moduleFor, ModuleBasedTestResolver, ApplicationTestCase } from 'interna
...
-let readyWasCalled, domReady, readyCallbacks;
+let application, Application, rootDocument, callbacks;
+let readyWasCalled = 0;
+
+const dispatchEvent = eventName => {
+  callbacks[eventName].forEach(callback => callback());
+};
+
+const removeEventListener = (eventName, callback) => {
+  callbacks[eventName] = callbacks[eventName].filter(({ name }) => name !== callback.name);
```
chancancode, line 13: probably should just be `.filter(item => item !== callback)`

packages/@ember/application/tests/readiness_test.js
rreckonerr: Indeed! Had to trigger 'DOMContentLoaded' in the test, since readyState should become `complete` eventually

rreckonerr:
By the way, is `$.ready()` the last thing that stops us from passing foreign document object for iframe support?

rwjblue (approved):

Issue #15611, opened 2017-08-30
queryParams array attribute becomes a string if other attribute changes before model is resolved


mpugach:
See [this twiddle](https://ember-twiddle.com/4e746489c6021ed3ca6b20b7fecc47af) please.

There are two queryParams:
- `categories` array
- `searchQuery` string

The model resolves the current time with a 1-second timeout.

Steps:
1. Open the console and see the current value of `categories`.
2. Add at least one category by hitting the "Add category" button (do not enter anything into the text input on this stage).
3. Observe `categories` is still an array in the console.
4. Enter any text into the input field.
5. If the `searchQuery` is modified before the transition completes, the `categories` becomes a string. From this point, it becomes even worse with every modification because it escapes escape symbols on every iteration.

MikiDi:
I can confirm above issue (with Ember 3.17, Firefox 79.0).  

Steps to reproduce:
- define 2 query params as described above
- Have some change to the array content trigger a model reload (make sure it's not in the default state, so it appears in the URL)
- In the `beforeModel`-hook, make a change to the `searchQuery` string ("the other query param")

From this point on, you should see the array query param getting escaped recursively

rreckonerr:
Can you please check if #19236 fixes the issue?

mpugach:
@rreckonerr, thank you for your contribution!

Have you tried modifying [twidle.json](https://ember-twiddle.com/4e746489c6021ed3ca6b20b7fecc47af)? Because constructing the dummy application locally in the nearest future is too much for me.

rwjblue:
Ya still seems to be an issue with 3.22.1. https://ember-twiddle.com/2d80d71daa9a9a99676aef471cb55b90?openFiles=templates.application.hbs%2C

rwjblue:
This seems like the same as what @mfeckie is working on over in https://github.com/emberjs/ember.js/pull/19145.

sly7-7:
@rwjblue I've reproduced your twiddle in https://github.com/sly7-7/ember-repro-15611 (ember 3.25.1), and it looks like this is fixed.
So I guess this could be closed, and #19145 maybe not necessary

rwjblue:
Thanks @sly7-7!

PR #19145, opened 2020-09-23
Bugfix/query param serialization

mfeckie:
When Ember serializes a query parameter of type `Array`, it does so with JSON.stringify.

Deserialize converts the `string` back to an `Array`.

If there are slow / high latency network connections, it is possible for the `serialize` method to get called in rapid succesion, prior to `deserialize` being called.  This results in `JSON.stringify` being called multiple times on the value, which results in to being converted from an `Array` to a `string`

```js
array = ['red', 'blue'];
// Array [ "red", "blue" ]
once = JSON.stringify(array);
// "[\"red\",\"blue\"]"
twice = JSON.stringify(once);
// "\"[\\\"red\\\",\\\"blue\\\"]\""
```

I've put together a reproduction over on [ember-twiddle](https://ember-twiddle.com/9a887d0edf25027edade31f759e2e9be). **note** if you try too hard, it may crash the browser because the strings grow exponentially.

QueryParams

I've added failing tests for this scenario and then implemented a fix.

I have added a `while` loop to recursively deal with scenarios where the QP's may have been repeatedly encoded.  I don't _think_ this could result in an infinite loop, rather it would result in a error breaking the loop.

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/router.ts
+++ b/packages/@ember/-internals/routing/lib/system/router.ts
@@ -735,7 +735,7 @@ class EmberRouter extends EmberObject {
   _serializeQueryParam(value: unknown, type: string) {
     if (value === null || value === undefined) {
       return value;
-    } else if (type === 'array' && value instanceof Array) {
+    } else if (type === 'array' && Array.isArray(value)) {
       return JSON.stringify(value);
     } else if (type === 'array' && typeof value === 'string') {
       while (typeof value === 'string') {
--- a/packages/@ember/-internals/routing/lib/system/router.ts
+++ b/packages/@ember/-internals/routing/lib/system/router.ts
@@ -735,7 +735,12 @@ class EmberRouter extends EmberObject {
   _serializeQueryParam(value: unknown, type: string) {
     if (value === null || value === undefined) {
       return value;
-    } else if (type === 'array') {
+    } else if (type === 'array' && value instanceof Array) {
+      return JSON.stringify(value);
+    } else if (type === 'array' && typeof value === 'string') {
+      while (typeof value === 'string') {
+        value = JSON.parse(value);
+      }
       return JSON.stringify(value);
     }
 
--- a/packages/ember/tests/routing/query_params_test.js
+++ b/packages/ember/tests/routing/query_params_test.js
@@ -1229,7 +1229,7 @@ moduleFor(
     }
 
     async ['@test (de)serialization: arrays when corrupted'](assert) {
-      assert.expect(7);
+      assert.expect(6);
 
       this.add(
         'controller:index',
--- a/packages/ember/tests/routing/query_params_test.js
+++ b/packages/ember/tests/routing/query_params_test.js
@@ -1228,6 +1228,29 @@ moduleFor(
       this.assertCurrentPath('/?foo=%5B%5D', 'longform supported');
     }
 
+    async ['@test (de)serialization: arrays when corrupted'](assert) {
+      assert.expect(7);
+
+      this.add(
+        'controller:index',
+        Controller.extend({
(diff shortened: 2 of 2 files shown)


packages/@ember/-internals/routing/lib/system/router.ts
```diff
@@ -735,7 +735,12 @@ class EmberRouter extends EmberObject {
   _serializeQueryParam(value: unknown, type: string) {
     if (value === null || value === undefined) {
       return value;
-    } else if (type === 'array') {
+    } else if (type === 'array' && value instanceof Array) {
```
rwjblue, line 738: I think we should use `Array.isArray` here (not `instanceof Array`)

rwjblue (commented):
🤔 Can we prevent the double encoding from happening to begin with?

packages/@ember/-internals/routing/lib/system/router.ts
mfeckie: Done

mfeckie:
I'd love to hear how to prevent the double encoding 😄

rwjblue:
Well, in general, `_serializeQueryParam` is called from `Route.prototype.serializeQueryParam` which is only called in two places; once with the initial / default value (unlikely to be related IMHO), and in `Route.prototype.finalizeQueryParamChange` here:

https://github.com/emberjs/ember.js/blob/15217a3ab36a35c8b26361c3155a30efe4a0f8ae/packages/@ember/-internals/routing/lib/system/route.ts#L2555-L2557

Assuming that is the case, it would infer that the controller's local property representing the QP is actually changing from an array to a string (for some period of time). Where is _that_ happening?

rwjblue:
@mfeckie - Any chance to look into that?

mfeckie:
@rwjblue Sure, will do some digging :)

mfeckie:
After many hours of digging today, I _think_ it's because of this

https://github.com/emberjs/ember.js/blob/15217a3ab36a35c8b26361c3155a30efe4a0f8ae/packages/%40ember/-internals/routing/lib/system/router.ts#L1621-L1638

`value` on line 1633 I think is always a `string` because it's from the URL

https://github.com/emberjs/ember.js/blob/15217a3ab36a35c8b26361c3155a30efe4a0f8ae/packages/%40ember/-internals/routing/lib/system/router.ts#L1633-L1636

This isn't a 'problem' when a full cycle completes, I think, because `serialize` and `deserialize` both complete.

When additional changes happen (e.g. user adds more characters to search) serialize ends up being called multiple times and deserialize doesn't get called the same number of times.

At this point, I can say that the proposed fix does work, but I can't say that this investigation left me feeling like I truly know what's going on.

rreckonerr:
@mfeckie Did you check the latest master? There's a PR merged that should cover this issue https://github.com/emberjs/ember.js/pull/19236

mfeckie:
No, I didn't.  I opened this PR back in September and haven't been keeping track of changes

rwjblue:
We should still try to land these tests, even if the other PR fixed the underlying issue.

rreckonerr:
@rwjblue Why should we support query parameters serialized multiple times?

sly7-7:
@mfeckie @rreckonerr From what I see, the `base` serialization issue is fixed. But if we specifically want the 'multiple encoding' thing to work, then I think we need this 
```
while (typeof value === 'string') {
        value = JSON.parse(value);
      }
```
loop. That beeing said, since the underlying issue is fixed, this case should not happen. Or maybe I miss something.

cc/ @rwjblue

sly7-7:
@rwjblue Perhaps this one is not necessary., see https://github.com/emberjs/ember.js/issues/15611#issuecomment-791922164

rwjblue:
Thanks @mfeckie and @sly7-7!

Issue #16832, opened 2018-07-19
Component helper not updating when component name changes


nlfurniss:
I’m upgrading my app from Ember 2.18.2 to 3.1.2 and I’m running into an issue with the `component` helper and yielding components.

Here is a [Twiddle demonstrating the issue](https://ember-twiddle.com/d65fa55ddb07dade0afffcb4256ee48b). When the Ember dependency is set to 2.18.2 the text changes. When you set Ember to 3.0.0 the text changes. When Ember is set to 3.1.3 the text does not update.

For anyone else running into this issue, Tom Dale provided a workaround in [this Twiddle](https://ember-twiddle.com/0fb1fe1f6428d554a077efe08ee017c4?openFiles=templates.index.hbs%2C).

Rob Jackson:
```
Tom and I dug into it a bit and came up with the same work around (using the component helper). The issue comes down to this AST transform completely failing when there are no positional or named arguments to an invocation:

https://github.com/emberjs/ember.js/blob/v3.4.0-beta.1/packages/ember-template-compiler/lib/plugins/transform-dot-component-invocation.ts#L88
```

pixelhandler:
@nlfurniss I see you have a workaround now, thanks for sharing the solution and pointing out the issue with "no positional or named arguments".

Feel free to close this when ready

nlfurniss:
@pixelhandler why did you close this? There is a workaround, but it's a regression in Ember that should be fixed

rwjblue:
Whoops!

pixelhandler:
@nlfurniss just checking back, is this resolved in a current release of Ember ?

nlfurniss:
No it has not

chancancode:
(I don't think the bug/fix is really about the AST transform, those are just "opportunistic" – when there are syntactic positions that are known to be component invocations we hard coded them to be such, but this should work)

chancancode:
This is a simplified test case: https://ember-twiddle.com/5f7ed4b003e14ef40e47f8de1aea5be6

chancancode:
If anyone has time, would be good to turn that simplified test case into a failing test, both here and in glimmer-vm.

lifeart:
@rwjblue looks like issue covered in https://github.com/emberjs/ember.js/blob/master/packages/%40ember/-internals/glimmer/tests/integration/components/dynamic-components-test.js#L32

rwjblue:
[...]

PR #19146, opened 2020-09-23
test case for #16832

lifeart:
ref: https://github.com/emberjs/ember.js/issues/16832#issuecomment-697502463

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/components/dynamic-components-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/components/dynamic-components-test.js
@@ -9,6 +9,80 @@ import { backtrackingMessageFor } from '../../utils/backtracking-rerender';
 moduleFor(
   'Components test: dynamic components',
   class extends RenderingTestCase {
+    ['@test Component helper updating when component name changes #16832 [curly]']() {
+      this.registerComponent('foo-bar', {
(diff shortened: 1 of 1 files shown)


kategengler:
Is this still relevant?

johanrd:
@lifeart I tried to reproduce now, and I am happy to announce that at least one of the two tests passes now

**TLDR: this is seemingly no longer a problem if you use angle bracket invocation <MyComponent instead of {{curly}}:**        

  - {{comp}} (curly) — **broken**, doesn't re-render when name changes (see [test failure](https://github.com/johanrd/ember.js/actions/runs/22807979720/job/66160117193?pr=3#step:5:18))
  - <MyComponent /> (angle bracket) — **works fine** (see [succeeding test suite](https://github.com/johanrd/ember.js/pull/4))
                                                                                                                                                                                                                     
So the `(component)` helper itself seems to track reactivity correctly. It's only curly invocation of the yielded result that fails to pick up the update.                                                                   
                                                                                                                                                                                                                     
…Since curly component invocation is the legacy/discouraged syntax and angle brackets are the modern way, this is arguably low-priority. The fix for any user hitting this is just: use `<Comp />` instead of  `{{comp}}`. Might be worth closing #16832 and #19146 with that note rather than fixing deep Glimmer VM curly invocation reactivity.?

PR #19147, opened 2020-09-23
[DOC] add sparse arrays iteration details #14725

lifeart:
ref: https://github.com/emberjs/ember.js/issues/14725#issuecomment-696986762

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/helpers/each-in.ts
+++ b/packages/@ember/-internals/glimmer/lib/helpers/each-in.ts
@@ -15,6 +15,9 @@ import { consumeTag } from '@glimmer/validator';
   The default behavior of `{{#each}}` is to yield its inner block once for every
   item in an array passing the item as the first block parameter.
 
+  For performance reasons we are using a standard `for` loop for `{{each}}` iteration,
+  what's why sparse arrays is not supported with `{{each}}`, but supported with `{{each-in}}`.
+
   Assuming the `@developers` argument contains this array:
 
   ```javascript
@@ -120,7 +123,7 @@ import { consumeTag } from '@glimmer/validator';
  */
 
 /**
-  The `{{each-in}}` helper loops over properties on an object.
+  The `{{each-in}}` helper loops over properties on an object or values on sparse arrays.
 
   For example, given this component definition:
 


jenweber:
@krisselden is this good to merge? Thank you!

packages/@ember/-internals/glimmer/lib/helpers/each-in.ts
```diff
@@ -15,6 +15,9 @@ import { consumeTag } from '@glimmer/validator';
   The default behavior of `{{#each}}` is to yield its inner block once for every
   item in an array passing the item as the first block parameter.
 
+  For performance reasons we are using a standard `for` loop for `{{each}}` iteration,
+  what's why sparse arrays is not supported with `{{each}}`, but supported with `{{each-in}}`.
```
rwjblue, line 19: FWIW, I don't know that "for performance reasons" is a good thing to say here.

Also, have we confirmed this is still true? There have been some quasi-recent (since 2016 which is when https://github.com/emberjs/ember.js/issues/14725#issuecomment-269389814 was) changes that will attempt to leverage native iterables if possible. In that case, I _think_ sparse arrays are handled more as expected? We should test either way to confirm.

---

If we do need to mention that sparse arrays are unsupported, we should just say that (and not compare to `each-in` or mention performance).

packages/@ember/-internals/glimmer/lib/helpers/each-in.ts
```diff
@@ -120,7 +123,7 @@ import { consumeTag } from '@glimmer/validator';
  */
 
 /**
-  The `{{each-in}}` helper loops over properties on an object.
+  The `{{each-in}}` helper loops over properties on an object or values on sparse arrays.
```
rwjblue, line 126: Conceptually, it seems pretty odd to suggest folks use `{{each-in` for arrays...

rwjblue (changes requested):

packages/@ember/-internals/glimmer/lib/helpers/each-in.ts
lifeart: @pzuraq could you clarify it or suggest better wording?

locks:
Hey friends, what's the status here?

locks:
Poke, is this still true?

kategengler:
Closing as obsolete

PR #19154, opened 2020-09-26
Upgrade prettier

BarryThePenguin:
✨ `prettier@2` ✨ 

Hiding white-space may help 👀 https://github.com/emberjs/ember.js/pull/19154/files?diff=split&w=1

PR #19156, opened 2020-09-26
Upgrade to eslint@7

BarryThePenguin:
✨ `eslint@7` ✨

Looks like there are a few new lint errors, specifically around tests with duplicate names?

@chancancode pointed out that only one of each test may be running?

Includes #19154 too


rwjblue:
@BarryThePenguin - Ready for a rebase (now that #19154 is landed), can you organize basically the same way as that PR? One commit for config changes, one commit for any manual fixes you have to do, and a final commit for `eslint --fix`?

rwjblue:
Hmm, actually. I'm going to take a crack at just merging locally (seems like you've already got it setup that way).

rwjblue:
Thank you @BarryThePenguin! If you still have time/resources, tackling other dependency updates would be amazing...

BarryThePenguin:
No problem ☺️ fixing the test names was the big motivator for the eslint upgrade, so I'm glad that's fixed. I wasn't sure where to start with them 😂

I've got another branch locally with some `@typescript/eslint` changes. It may be a little bit more involved as it requires a few decisions.. 🙈

PR #19160, opened 2020-09-28
[FEAT] Implements Helper Managers

pzuraq:
Implements the helper manager feature specified in RFC 625. Highlights:

1. Adds `getDebugName` to the interface for helper managers. This is an optional hook that is used for better logging purposes, and matches other internal APIs we've added recently.

2. `hasScheduledEffect` has not yet been implemented, and attempting to use it will cause an assertion to be thrown.

Helper managers are not exposed with this PR, and the version passed to `helperCapabilities` is optimistic, but can be changed when we do expose them (along with an appropriate feature flag).


packages/@ember/-internals/glimmer/lib/resolver.ts
```diff
@@ -358,32 +365,28 @@ export default class RuntimeResolverImpl implements RuntimeResolver<OwnedTemplat
...
-      return createHelperRef(helper, args.capture());
-    };
+    if (manager === undefined) {
+      return null;
+    }
+
+    if (isInternalHelperManager(manager)) {
+      return customHelper(manager, factory);
+    } else if (EMBER_GLIMMER_HELPER_MANAGER) {
+      return customHelper(manager, factory);
```
pzuraq, line 386: This should be passing `factory.class` instead of `factory`, to match the other public facing APIs for managers. The manager receives the definition it was applied to, not the factory we wrap it with.

packages/@ember/-internals/glimmer/lib/helper.ts
```diff
@@ -30,16 +39,10 @@ export interface SimpleHelper<T = unknown> {
...
 }
 
-export function isHelperFactory(
-  helper: any | undefined | null
-): helper is Factory<SimpleHelper | HelperInstance, HelperFactory<SimpleHelper | HelperInstance>> {
-  return (
-    typeof helper === 'object' && helper !== null && helper.class && helper.class.isHelperFactory
-  );
-}
+const INTERNAL_MANAGERS = new WeakSet();
```
pzuraq, line 42: I'm not sure it matters much, but I originally didn't use a WeakSet because the only type of manager it was important to pass the `factory` for is `ClassicHelperManager`. That manager is stateless, and a `===` comparison for that manager is going to be faster even that checking if the value exists in a `WeakSet`.

packages/@ember/-internals/glimmer/lib/resolver.ts
pzuraq: Seems that our other public facing APIs do not agree here either. @rwjblue and I discussed, it makes more sense for these APIs to receive the definition itself and not the factory, especially taking into account the future with imports. We'll update modifier managers to receive the definition via a new capability.

packages/@ember/-internals/glimmer/lib/helper.ts
rwjblue: I think we still need it actually, for the feature flagging. We only want to use the `customHelper` method for our own internal helper managers when the feature flag is disabled.

packages/@ember/-internals/glimmer/lib/resolver.ts
rwjblue: I changed it to match what we do for modifiers, but unfortunately you are right: modifier managers and component managers have different APIs in this respect.

I'll update this PR to pass the `owner.factoryFor(...).class` value in here, and also open another PR to tweak the new capability for modifier managers.

rwjblue: Updated...

pzuraq (commented):
LGTM

rwjblue (approved):

PR #19163, opened 2020-09-28
[BUGFIX beta] Use args proxy for modifier managers.

### LINKED ISSUE #19162: Modifier Managers should not re-consume their args on destroy
According to the [current modifier manager spec](https://github.com/emberjs/rfcs/blob/master/text/0373-Element-Modifier-Managers.md#destroymodifier): 

> destroyModifier will be called when the modifier is no longer needed. This is intended for performing object-model level cleanup.

then, in the following code example:
```ts
 destroyModifier(instance, args) {
    if (instance.willDestroyDOM !== undefined) {
      instance.willDestroyDOM();
    }
  }
```
args are passed to `destroyModifier`.

I think this is a bug because:
 - it means that args are all re-consumed on _destruction_ of a render / component tree
 - `if` guards in templates no longer can guarantee the safety of the shape of data
 - in order for modifiers' destruction to not trigger exceptions wherever args' derived data is derived, you need to add additional guards all over your code, or do `foo?.bar?.bax` everywhere (this generates a lot of boilerplate)
  
------------------------------

**I propose we _remove_ `args` from the `destroyModifier` hook so that** 
  - none of the above issues exist anymore
  - teardown is slightly faster due to not re-consuming args

----------------------

For example, today, the following isn't safe:

```hbs
{{#if this.data.foo}}
  <SomeComponent @data={{this.data}} />
{{/if}}
```
```ts
// some-component
export default setComponentTemplate(hbs`
  <span {{some-modifier this.nestedData}}>breaks</span>
`, class extends Component {
  get nestedData() {
    // you should be able to assume foo exists, because of the condition in the previous template
    // but today, you can't assume that :(
    return this.args.data.foo.dataOnFoo;
  }
});
```

Minimal Reproduction / Description of the Problem: https://ember-twiddle.com/5a3fa797b26e8807869340792219a5ee?openFiles=components.demo%5C.hbs%2Ccomponents.demo%5C.hbs

Relevant Discord Threads:
 - [one](https://discordapp.com/channels/480462759797063690/485447409296736276/722483725157400576)
 - [two](https://discordapp.com/channels/480462759797063690/608346628163633192/743649497518702613)

rwjblue:
This PR does two main things:

1. Generalizes the args proxy for use in components, helpers, and modifiers. This does change the semantics slightly of `positional` arguments for component managers, notably it would make it so using computed properties with them would require a bit more work from custom component managers. The new semantics are more inline with the future of Ember, but if there many custom managers that use positional args we may want to reconsider this change.
2. Introduce the `3.22` manager API version to ensure that custom modifier managers only consume the arguments that are actually used by the manager. In order for specific modifiers to take advantage of these changes, they need to be updated to use `modifierCapabilities('3.22')` (instead of `3.13`).

Fixes #19162

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/component-managers/custom.ts
+++ b/packages/@ember/-internals/glimmer/lib/component-managers/custom.ts
@@ -1,13 +1,9 @@
 import { ENV } from '@ember/-internals/environment';
-import { CUSTOM_TAG_FOR } from '@ember/-internals/metal';
 import { Factory } from '@ember/-internals/owner';
-import { HAS_NATIVE_PROXY } from '@ember/-internals/utils';
 import { assert } from '@ember/debug';
-import { DEBUG } from '@glimmer/env';
 import {
+  Arguments,
   Bounds,
-  CapturedArguments,
-  CapturedNamedArguments,
   ComponentCapabilities,
   ComponentDefinition,
   Destroyable,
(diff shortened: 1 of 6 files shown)


rwjblue:
Need to coordinate with @pzuraq to avoid stepping on his toes over in https://github.com/emberjs/ember.js/pull/19160 (which does similar things).

rwjblue:
I've pulled in the changes from https://github.com/emberjs/ember.js/pull/19160 so we should be able to easily rebase that on top of these changes once this lands.

pzuraq (approved):
Looks good, thanks for doing this!

PR #19164, opened 2020-09-28
[BUGFIX lts] Entangles custom EmberArray implementations when accessed

pzuraq:
Restores the previous logic that existed for entangling custom EmberArray implementations when they are accessed, along with a test for the functionality.


rwjblue (approved):

stefanpenner:
@pzuraq / @rwjblue curious if there is an ETA for when the corresponding LTS will be released?

PR #19171, opened 2020-09-30
[FEAT] Implements invokeHelper

pzuraq:
Reopening #19165, thanks for getting it started @NullVoxPopuli!

Implements `invokeHelper` as described in [RFC 626](https://github.com/emberjs/rfcs/blob/master/text/0626-invoke-helper.md). Everything is behind a feature flag, ready to flip!


packages/@ember/-internals/glimmer/lib/helpers/custom.ts
```diff
@@ -62,12 +73,91 @@ function hasDestroyable(manager: HelperManager): manager is HelperManagerWithDes
   return manager.capabilities.hasDestroyable;
 }
 
+let ARGS_CACHES = DEBUG ? new WeakMap<SimpleArgsProxy, Cache<Partial<Arguments>>>() : undefined;
+
+function getArgs(proxy: SimpleArgsProxy): Partial<Arguments> {
+  return getValue(DEBUG ? ARGS_CACHES!.get(proxy)! : proxy.argsCache!)!;
+}
+
+class SimpleArgsProxy {
```
NullVoxPopuli, line 82: Does something like this already exist for modifiers?

packages/@ember/-internals/glimmer/lib/utils/managers.ts
```diff
@@ -57,20 +66,26 @@ function getManager<D extends ManagerDelegate>(
...
+      managers = new WeakMap();
+      OWNER_MANAGER_INSTANCES.set(owner, managers);
+    }
   }
 
   let instance = managers.get(factory);
 
   if (instance === undefined) {
-    instance = factory(owner);
+    instance = factory(owner!);
```
NullVoxPopuli, line 88: How do you know that owner exists here?

NullVoxPopuli (commented):
Eeeeeeeee :tada:

packages/@ember/-internals/glimmer/lib/helpers/custom.ts
pzuraq: It's a more complicated version which proxies each individual argument. Using that would require users to pass in each arg as a function, which would be a bit prohibitive and is not what was spec'd. So, this one is a simpler proxy that basically calls the users compute function and then returns the proper value whenever `positional` or `named` is accessed.

packages/@ember/-internals/glimmer/lib/utils/managers.ts
pzuraq: It may actually be `undefined`, in the case of helper managers, but for modifier/component managers it will always exist at this point (since you are not currently allowed to pass an `undefined` owner to `getModifierManager` or `getComponentManager`). The types here are a bit messy, but I think it _should_ be fine.

packages/@ember/-internals/glimmer/tests/integration/helpers/invoke-helper-test.js
```diff
@@ -0,0 +1,272 @@
...
+          }
+        }
+
+        let helper = invokeHelper(context, TestHelper);
+
+        runTask(() => destroy(context));
+
+        assert.ok(isDestroyed(context), 'context destroyed');
+        assert.ok(isDestroyed(helper), 'helper cache destroyed');
+      }
```
rwjblue, line 228: Can you modify this test to associate a destroyable with the `context` and use `assert.step` to verify the ordering / relationship between destroyables associated with `context` and `helper`?

packages/@ember/-internals/glimmer/tests/integration/helpers/invoke-helper-test.js
```diff
@@ -0,0 +1,272 @@
...
+          }
+        }
+
+        let helper = invokeHelper(context, TestHelper);
+
+        runTask(() => destroy(helper));
+
+        assert.notOk(isDestroyed(context), 'context NOT destroyed');
+        assert.ok(isDestroyed(helper), 'helper cache destroyed');
+      }
```
rwjblue, line 247: Same as above, can you ensure the ordering relationship of the helper's own destroyables and the contexts?

packages/@ember/-internals/glimmer/tests/integration/helpers/invoke-helper-test.js
```diff
@@ -0,0 +1,272 @@
...
+      ['@test asserts if no context object is passed']() {
+        expectAssertion(() => {
+          invokeHelper(undefined, class extends Helper {});
+        }, /Expected a context object to be passed as the first parameter to invokeHelper, got undefined/);
+      }
+
+      ['@test asserts if no manager exists for the helper definition']() {
+        expectAssertion(() => {
+          invokeHelper({}, class {});
+        }, /Expected a helper definition to be passed as the second parameter to invokeHelper, but no helper manager was found. Did you use setHelperManager to associate a helper manager?/);
```
rwjblue, line 268: As mentioned above, this message should be augmented to include _some_ information RE: the helper that was passed.

packages/@ember/-internals/glimmer/lib/helpers/custom.ts
```diff
@@ -62,12 +73,91 @@ function hasDestroyable(manager: HelperManager): manager is HelperManagerWithDes
...
+
 export default function customHelper(
   manager: HelperManager<unknown>,
   definition: HelperDefinition
 ): GlimmerHelper {
-  return (args, vm) => {
-    const bucket = manager.createHelper(definition, argsProxyFor(args.capture(), 'helper'));
+  return (vmArgs, vm) => {
+    const args = argsProxyFor(vmArgs.capture(), 'helper');
+    const bucket = manager.createHelper(definition, args);
```
rwjblue, line 160: Thank you for cleaning this up, much easier to read this way

packages/@ember/-internals/glimmer/lib/helpers/custom.ts
```diff
@@ -62,12 +73,91 @@ function hasDestroyable(manager: HelperManager): manager is HelperManagerWithDes
...
+  const manager = getHelperManager(owner, definition)!;
+
+  // TODO: figure out why assert isn't using the TS assert thing
+  assert(
+    `Expected a helper definition to be passed as the second parameter to invokeHelper, but no helper manager was found. Did you use setHelperManager to associate a helper manager?`,
+    manager
+  );
+
+  let args = new SimpleArgsProxy(context, computeArgs);
+  let helper = manager.createHelper(definition, args);
```
rwjblue, line 127: Lets call this `bucket` (matches what we call it in the actual `customHelper` implementation below, and helps us avoid any assumptions around what it is...)

rwjblue (changes requested):
Looking really good! I left a few inline comments/questions, but I think we are also missing some test cases:

* Usage of `invokeHelper` with custom helper managers (will need to be guarded by both feature flags, but should still be tested)
* Ensuring that associating destructors to the the helper context works properly (for both `Ember.Helper`/`Ember.Helper.helper` helpers and custom helpers)
* Usage of non-class based helpers
* Confirming the caching of the `computeArgs`, specifically how many times the third argument is expected to be invoked in various contexts
* Confirming that a helper that accesses none of the args (positional or named), and nothing else is "constant"
* Confirming that a helper that accesses only args that themselves are constant (e.g. the `computeArgs` function is something that uses no tags), is constant

packages/@ember/-internals/glimmer/lib/helpers/custom.ts
```diff
@@ -62,12 +73,91 @@ function hasDestroyable(manager: HelperManager): manager is HelperManagerWithDes
   return manager.capabilities.hasDestroyable;
 }
 
+let ARGS_CACHES = DEBUG ? new WeakMap<SimpleArgsProxy, Cache<Partial<Arguments>>>() : undefined;
```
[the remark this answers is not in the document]
pzuraq: correct, exactly

packages/@ember/-internals/glimmer/lib/helpers/custom.ts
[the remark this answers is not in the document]
pzuraq: Getters on instances of objects are really bad for performance, they should always be on a prototype. Cc @krisselden

packages/@ember/-internals/glimmer/lib/helper.ts
```diff
@@ -145,9 +151,20 @@ class ClassicHelperManager implements HelperManager<ClassicHelperStateBucket> {
...
+    if (definition.class === undefined) {
+      ownerInjection = {};
+      setOwner(ownerInjection, this.owner!);
+    }
+
     return {
-      instance: definition.create(),
+      instance: definition.create(ownerInjection),
       args,
     };
```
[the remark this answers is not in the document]
pzuraq: we can't call `new definition` because it extends from `EmberObject`, that would throw an error. We could create a single `ownerInjection` object for the entire manager though, so there are fewer of them

packages/@ember/-internals/glimmer/lib/helpers/custom.ts
```diff
@@ -62,12 +73,91 @@ function hasDestroyable(manager: HelperManager): manager is HelperManagerWithDes
...
+  );
+
+  const owner = getOwner(context);
+  const manager = getHelperManager(owner, definition)!;
+
+  // TODO: figure out why assert isn't using the TS assert thing
+  assert(
+    `Expected a helper definition to be passed as the second parameter to invokeHelper, but no helper manager was found. Did you use setHelperManager to associate a helper manager?`,
+    manager
+  );
```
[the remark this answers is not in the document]
pzuraq: It does have a test, but will add info about the definition 👍

packages/@ember/-internals/glimmer/lib/helpers/custom.ts
```diff
@@ -62,12 +73,91 @@ function hasDestroyable(manager: HelperManager): manager is HelperManagerWithDes
...
+    });
+  } else {
+    throw new Error('TODO: unreachable, to be implemented with hasScheduledEffect');
+  }
+
+  if (hasDestroyable(manager)) {
+    let destroyable = manager.getDestroyable(helper);
+
+    associateDestroyableChild(context, cache);
+    associateDestroyableChild(cache, destroyable);
```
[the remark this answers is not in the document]
pzuraq: The RFC explicitly specifies that you can call `destroy(cache)` to eagerly destroy the helper without destroying the parent context. This gives the user more fine-grained control over when the helper is destroyed, if they want to manage its lifecycle more directly (e.g. dynamically creating/destroying helpers).

`invokeHelper` always returns a `cache`, the other branch has not been implemented yet, but will return an empty cache when it is. This is just a reminder to implement that code.

packages/@ember/-internals/glimmer/lib/helpers/custom.ts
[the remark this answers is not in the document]
rwjblue: > Getters on instances of objects are really bad for performance

Say more here? This seems overly vague, this advice is normally used when we are talking about `Object.defineProperty` not when defining POJO's with getters directly on them.

packages/@ember/-internals/glimmer/lib/helper.ts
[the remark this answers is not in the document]
rwjblue: @pzuraq - Can you link me to where we error for that? A quick test locally suggests that there is no error or deprecation when calling `new Ember.Object()`...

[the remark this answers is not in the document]
rwjblue: I assume you mean this one? 

https://github.com/emberjs/ember.js/blob/v3.16.10/packages/@ember/-internals/runtime/lib/system/core_object.js#L288-L299

Which you removed in https://github.com/emberjs/ember.js/pull/19088...

packages/@ember/-internals/glimmer/lib/helpers/custom.ts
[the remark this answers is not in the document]
rwjblue: Bah, that is dumb. Everything is horrible. 🔥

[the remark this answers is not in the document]
rwjblue: @pzuraq - Still need to ensure that this object is frozen in debug (could probably put it in the class constructor).

packages/@ember/-internals/glimmer/lib/helpers/custom.ts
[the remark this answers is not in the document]
rwjblue: OK, that part makes sense, but I still think that the `cache` probably still needs to be associated with the context regardless of `hasDestroyable` on the manager. The only thing this branch should be doing is:

```js
let destroyable = manager.getDestroyable(bucket); // reminder to rename this to bucket ;)
associateDestroyableChild(cache, destroyable);
```

packages/@ember/-internals/glimmer/lib/helpers/custom.ts
```diff
@@ -63,12 +74,94 @@ function hasDestroyable(manager: HelperManager): manager is HelperManagerWithDes
...
+let ARGS_CACHES = DEBUG ? new WeakMap<SimpleArgsProxy, Cache<Partial<Arguments>>>() : undefined;
+
+function getArgs(proxy: SimpleArgsProxy): Partial<Arguments> {
+  return getValue(DEBUG ? ARGS_CACHES!.get(proxy)! : proxy.argsCache!)!;
+}
+
+class SimpleArgsProxy {
+  argsCache?: Cache<Partial<Arguments>>;
+
+  constructor(
```
rwjblue, line 86: Needs to be frozen, current helpers don't see this object but those using a custom manager could

packages/@ember/-internals/glimmer/tests/integration/helpers/invoke-helper-test.js
```diff
@@ -0,0 +1,582 @@
...
+        assert.equal(instance.value, 5, 'helper works');
+        assert.equal(count, 1, 'helper only called once');
+
+        helper.number = 5;
+
+        assert.equal(instance.value, 6, 'helper works');
+        assert.equal(count, 2, 'helper called a second time');
+      }
+
+      '@test helper that with constant args is costant'(assert) {
```
rwjblue, line 310: Typo: costant -> constant

packages/@ember/-internals/glimmer/tests/integration/helpers/invoke-helper-test.js
```diff
@@ -0,0 +1,618 @@
...
+      }
+
+      '@test asserts if no manager exists for the helper definition'() {
+        expectAssertion(() => {
+          invokeHelper({}, class {});
+        }, /Expected a helper definition to be passed as the second parameter to invokeHelper, but no helper manager was found. The definition value that was passed was `\(unknown function\)`. Did you use setHelperManager to associate a helper manager with this value?/);
+      }
+    }
+  );
+
```
rwjblue, line 440: I think this needs to be:

```suggestion
  if (
    EMBER_GLIMMER_HELPER_MANAGER &&
    EMBER_GLIMMER_INVOKE_HELPER
  ) {
```

Specifically, if we enable one before the other (which almost always does happen) these tests may fail. As far as I can tell **both** features are required for these tests...

pzuraq: probably hard to see but it was nested in the other `if`, I can see how that's confusing though. Updated to check both 👍

rwjblue: Ah, gotcha. The indentation was hard to spot on mobile for sure.

rwjblue (approved):

rwjblue:
🎂👌💫

PR #19172, opened 2020-09-30
[BUGFIX lts] Ensures that tracked properties initialize property

pzuraq:
Fixes tracked property initialization for classic classes.

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/components/tracked-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/components/tracked-test.js
@@ -14,6 +14,44 @@ import { Component } from '../../utils/helpers';
 moduleFor(
   'Component Tracked Properties',
   class extends RenderingTestCase {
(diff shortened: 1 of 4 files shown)


packages/@ember/-internals/runtime/lib/system/core_object.js
```diff
@@ -111,7 +111,7 @@ function initialize(obj, properties) {
       }
 
       if (isDescriptor) {
-        possibleDesc.set(obj, keyName, value);
+        obj[keyName] = value;
```
rwjblue, line 114: Chatted with @pzuraq about this, he is going to bring back the prior code (to avoid looking up the descriptor _again_).

rwjblue (changes requested):

rwjblue (approved):

ef4:
This caused an app failure in the wild, see https://github.com/emberjs/ember.js/issues/19179

PR #19173, opened 2020-09-30
Enforce usage of `capabilities` generation.

rwjblue:
Prior to this change it was possible to avoid using a given type's `capabilities` builder function (intentionally or on accident) by doing something like:

```js
class MyManager {
  capabilities = {
    // magical properties from Ember's internals
  }
}
```

The API's that we _intended_ folks to be using is something like:

```js
import { capabilties as modifierCapabilities } from '@ember/modifier';

class MyManager {
  capabilities = modifierCapabilities('3.22');
}
```

This commit ensures that Ember's own internal structures can not be "spoofed" (avoiding our constraints, or creating a frankenstein manager).

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/utils/managers.ts
+++ b/packages/@ember/-internals/glimmer/lib/utils/managers.ts
@@ -1,8 +1,10 @@
 import { Owner } from '@ember/-internals/owner';
-import { deprecate } from '@ember/debug';
+import { assert, deprecate } from '@ember/debug';
 import { COMPONENT_MANAGER_STRING_LOOKUP } from '@ember/deprecated-features';
+import { DEBUG } from '@glimmer/env';
+import { _WeakSet } from '@glimmer/util';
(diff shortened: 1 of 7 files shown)


packages/@ember/-internals/glimmer/lib/utils/managers.ts
```diff
@@ -17,6 +18,8 @@ const COMPONENT_MANAGERS = new WeakMap<
   ManagerFactory<ComponentManagerDelegate<unknown> | InternalComponentManager>
 >();
 
+const FROM_CAPABILITIES = DEBUG ? new WeakSet() : undefined;
```
pzuraq, line 21: Doesn't look like the `WeakSet` polyfill is being used here, may be why browserstack is failing

pzuraq (approved):
Looks good, besides the IE11 issue!

packages/@ember/-internals/glimmer/lib/utils/managers.ts
rwjblue: Nice, thanks for catching that!

PR #19175, opened 2020-10-01
Replace babel-eslint with @typescript-eslint/parser

BarryThePenguin:
`@typescript-eslint/parser` parses both JavaScript _and_ TypeScript

`babel-eslint` uses Babel, and a [Babel configuration file](https://babeljs.io/docs/en/configuration), to parse JavaScript.

`ember/ember.js` doesn't appear to use any language features that require `babel-eslint` :eyes:, so using just `@typescript-eslint/parser` should be enough

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -47,9 +47,9 @@ module.exports = {
 
       rules: {
         '@typescript-eslint/ban-ts-comment': 'warn',
-        '@typescript-eslint/ban-types': 'warn',
-        '@typescript-eslint/no-empty-function': 'warn',
-        '@typescript-eslint/no-this-alias': 'warn',
+        '@typescript-eslint/ban-types': 'off',
+        '@typescript-eslint/no-empty-function': 'off',
+        '@typescript-eslint/no-this-alias': 'off',
         '@typescript-eslint/no-var-requires': 'warn',
 
         // TODO: Enable and fix these rules
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,10 +2,11 @@ const path = require('path');
 
 module.exports = {
   root: true,
-  parser: 'babel-eslint',
+  parser: '@typescript-eslint/parser',
   extends: [
     'eslint:recommended',
     'plugin:import/errors',
+    'plugin:import/typescript',
     'plugin:qunit/recommended',
     'plugin:prettier/recommended',
   ],
@@ -23,13 +24,10 @@ module.exports = {
   },
 
   settings: {
-    'import/core-modules': ['require', 'backburner', 'router', 'ember/version', 'node-module'],
-    'import/parsers': {
-      '@typescript-eslint/parser': ['.ts'],
-    },
+    'import/core-modules': ['require', 'backburner', 'router', '@glimmer/interfaces'],
     'import/resolver': {
       node: {
-        extensions: ['.js', '.ts'],
+        extensions: ['.js', '.ts', '.d.ts'],
         paths: [path.resolve('./packages/')],
       },
     },
@@ -37,14 +35,9 @@ module.exports = {
 
   overrides: [
     {
-      files: ['**/*.ts'],
+      files: ['*.ts'],
 
-      parser: '@typescript-eslint/parser',
-      extends: [
-        'plugin:@typescript-eslint/eslint-recommended',
-        'plugin:import/typescript',
-        'prettier/@typescript-eslint',
-      ],
+      extends: ['plugin:@typescript-eslint/recommended', 'prettier/@typescript-eslint'],
 
       parserOptions: {
         sourceType: 'module',
@@ -53,9 +46,11 @@ module.exports = {
       },
 
       rules: {
-        // the TypeScript compiler already takes care of this and
-        // leaving it enabled results in false positives for interface imports
-        'no-unused-vars': 'off',
+        '@typescript-eslint/ban-ts-comment': 'warn',
+        '@typescript-eslint/ban-types': 'warn',
+        '@typescript-eslint/no-empty-function': 'warn',
+        '@typescript-eslint/no-this-alias': 'warn',
+        '@typescript-eslint/no-var-requires': 'warn',
 
         // TODO: Enable and fix these rules
         // Typescript provides better types with these rules enabled
@@ -64,13 +59,10 @@ module.exports = {
         'prefer-rest-params': 'off',
 
         'import/export': 'off',
-        // TODO: Remove with typescript 3.8
-        // Use `import type` instead
-        'import/named': 'off',
-        'import/no-unresolved': 'off',
       },
     },
     {
+      // TODO: files: ['packages/**/*.[jt]s'],
       files: ['packages/**/*.js'],
 
       parserOptions: {
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -57,8 +57,6 @@ module.exports = {
         'prefer-spread': 'off',
         'prefer-const': 'off',
         'prefer-rest-params': 'off',
-
-        'import/export': 'off',
       },
     },
     {
--- a/package.json
+++ b/package.json
@@ -91,8 +91,6 @@
     "@typescript-eslint/parser": "^4.3.0",
     "auto-dist-tag": "^1.0.0",
     "aws-sdk": "^2.686.0",
-    "babel-eslint": "^10.1.0",
-    "babel-plugin-module-resolver": "^4.0.0",
     "babel-template": "^6.26.0",
     "backburner.js": "^2.6.0",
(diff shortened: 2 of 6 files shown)


.eslintrc.js
```diff
@@ -2,10 +2,11 @@ const path = require('path');
 
 module.exports = {
   root: true,
-  parser: 'babel-eslint',
+  parser: '@typescript-eslint/parser',
   extends: [
     'eslint:recommended',
     'plugin:import/errors',
+    'plugin:import/typescript',
```
BarryThePenguin, line 9: Allows `eslint-plugin-import` to [parse TypeScript files](https://github.com/benmosher/eslint-plugin-import#typescript)

.eslintrc.js
```diff
@@ -23,28 +24,20 @@ module.exports = {
   },
 
   settings: {
-    'import/core-modules': ['require', 'backburner', 'router', 'ember/version', 'node-module'],
-    'import/parsers': {
-      '@typescript-eslint/parser': ['.ts'],
-    },
+    'import/core-modules': ['require', 'backburner', 'router', '@glimmer/interfaces'],
```
BarryThePenguin, line 27: `@glimmer/interfaces` [defines modules](https://unpkg.com/@glimmer/interfaces@0.62.1/package.json) for [`main`](https://unpkg.com/@glimmer/interfaces@0.62.1/dist/commonjs/es2017/index.js), [`jsnext:main`](https://unpkg.com/@glimmer/interfaces@0.62.1/dist/modules/es5/index.js), [`module`](https://unpkg.com/@glimmer/interfaces@0.62.1/dist/modules/es2017/index.js), and [`typings`](https://unpkg.com/@glimmer/interfaces@0.62.1/dist/types/index.d.ts), but only `typings` is published to npm

Adding `@glimmer/interfaces` to `import/core-modules` still the correct types to be resolved, but `eslint-plugin-import` will ignore it..

Also note I was able to remove `ember/version`, and `node-module`

.eslintrc.js
```diff
@@ -23,28 +24,20 @@ module.exports = {
   },
 
   settings: {
-    'import/core-modules': ['require', 'backburner', 'router', 'ember/version', 'node-module'],
-    'import/parsers': {
-      '@typescript-eslint/parser': ['.ts'],
-    },
```
BarryThePenguin, line 29: This is handled by [`plugin:import/typescript`](https://github.com/benmosher/eslint-plugin-import/blob/c51b6a97de1e0edad77a247d5da4c36db7c8d039/config/typescript.js#L12-L14)

.eslintrc.js
```diff
@@ -53,24 +46,21 @@ module.exports = {
...
         // Typescript provides better types with these rules enabled
         'prefer-spread': 'off',
         'prefer-const': 'off',
         'prefer-rest-params': 'off',
-
-        'import/export': 'off',
-        // TODO: Remove with typescript 3.8
-        // Use `import type` instead
-        'import/named': 'off',
-        'import/no-unresolved': 'off',
```
BarryThePenguin, line 70: Was able to remove these rules, and manually fix the `import/export` issues

.eslintrc.js
```diff
@@ -53,24 +46,21 @@ module.exports = {
...
       },
     },
     {
+      // TODO: files: ['packages/**/*.[jt]s'],
```
BarryThePenguin, line 63: Enabling this results in ~80 errors from the `require-yuidoc-access` and `no-const-outside-module-scope` rules. I'm not sure how important that is though 😅

package.json
```diff
@@ -91,8 +91,6 @@
     "@typescript-eslint/parser": "^4.3.0",
     "auto-dist-tag": "^1.0.0",
     "aws-sdk": "^2.686.0",
-    "babel-eslint": "^10.1.0",
-    "babel-plugin-module-resolver": "^4.0.0",
```
BarryThePenguin, line 95: 👋

packages/@ember/-internals/metal/lib/injected_property.ts
```diff
@@ -87,3 +80,5 @@ export default function inject(
     return decorator;
   }
 }
+
+export default inject;
```
BarryThePenguin, line 84: This is a workaround for the "multiple default exports" issue from `import/export`

This change is in a seperate commit, so is easy to revert..

.eslintrc.js
```diff
@@ -53,24 +46,21 @@ module.exports = {
       },
 
       rules: {
-        // the TypeScript compiler already takes care of this and
-        // leaving it enabled results in false positives for interface imports
-        'no-unused-vars': 'off',
+        '@typescript-eslint/ban-ts-comment': 'warn',
```
BarryThePenguin, line 49: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/ban-ts-comment.md

.eslintrc.js
```diff
@@ -53,24 +46,21 @@ module.exports = {
...
+        '@typescript-eslint/ban-types': 'warn',
```
BarryThePenguin, line 50: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/ban-types.md

I noticed a lot of usages of `any`, `object`, `Function`. This rule has a few options, but unsure how helpful they'll be here..

.eslintrc.js
```diff
@@ -53,24 +46,21 @@ module.exports = {
...
+        '@typescript-eslint/no-empty-function': 'warn',
```
BarryThePenguin, line 51: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-empty-function.md

I don't really understand the purpose of this rule? But it exists 😅 Can probably turn it off..

```suggestion
        '@typescript-eslint/no-empty-function': 'off',
```

.eslintrc.js
```diff
@@ -53,24 +46,21 @@ module.exports = {
...
+        '@typescript-eslint/no-this-alias': 'warn',
```
BarryThePenguin, line 52: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-this-alias.md

Used a quite a few times in tests. Again, unsure how helpful it would be? Could just turn off..

```suggestion
        '@typescript-eslint/no-this-alias': 'off',
```

.eslintrc.js
```diff
@@ -53,24 +46,21 @@ module.exports = {
...
+        '@typescript-eslint/no-var-requires': 'warn',
```
BarryThePenguin, line 53: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-var-requires.md

Looks like most usages of this rule are already disabled for `tslint`. May need corresponding eslint equivalents

.eslintrc.js
rwjblue: Let's do this in a follow up...

.eslintrc.js
pzuraq: We use `object` to mean "valid WeakMap key" in places, which I do think is a somewhat valid usage. Maybe we can make a type alias specifically for that, or start using record or some other type of key.

I'd like to try to slowly eliminate `any` and `Function` usage.

.eslintrc.js
rwjblue: The `require-yuidoc-access` rule is much more important that no-const-outside-module-scope. In typescript land, you are sometimes **forced** to use `const` outside of module scope to indicate to `tsc` that the variable can never change (other wise it often warns you that "variable could be undefined" if any method calls happen after declaration and before a reference).

I'm happy to include a fix for the `require-yuidocs-access` in this PR, or to just have a follow up PR that fixes it. Totally up to you.

package.json
rwjblue: Ya, this is good to remove. IIRC, we leverage `ember-cli-babel` to do our transpilation now and it handles this resolution for us.

rwjblue (approved):
Thanks for continuing to push this forward @BarryThePenguin!

.eslintrc.js
BarryThePenguin: I might do that in a followup PR.. I might make some time to add more test cases to https://github.com/emberjs/eslint-plugin-ember-internal first

BarryThePenguin:
Thanks for reviewing :wave:

I'll put together some followup PRs when I have the opportunity

.eslintrc.js
BarryThePenguin: The issues I've run into are `require-yuidocs-access` recognising the different styles of block comments used, specifically..

```js
/**
 * @param {string} foo
 * @returns {string}
 */

// vs

/**
  @param {string} foo
  @returns {string}
 */
```

PR #19184, opened 2020-10-05
[FEATURE modernized-built-in-components] Initial implementation

chancancode:
Adds the feature flag for [RFC #671](https://github.com/emberjs/rfcs/pull/671) and start implementing some of the proposed deprecations. The deprecations are gated behind the feature flag so they won't be visible until the RFC is merged and the feature is enabled.

### DIFF
--- a/packages/@ember/canary-features/index.ts
+++ b/packages/@ember/canary-features/index.ts
@@ -23,6 +23,7 @@ export const DEFAULT_FEATURES = {
   EMBER_DESTROYABLES: true,
   EMBER_GLIMMER_HELPER_MANAGER: true,
   EMBER_GLIMMER_INVOKE_HELPER: true,
+  EMBER_MODERNIZED_BUILT_IN_COMPONENTS: null,
 };
 
 /**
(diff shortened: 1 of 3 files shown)


packages/@ember/canary-features/index.ts
```diff
@@ -52,10 +53,10 @@ export const FEATURES = assign(DEFAULT_FEATURES, ENV.FEATURES);
   @public
 */
 export function isEnabled(feature: string): boolean {
-  let featureValue = FEATURES[feature];
+  let value = FEATURES[feature];
 
-  if (featureValue === true || featureValue === false) {
-    return featureValue;
+  if (value === true || value === false) {
+    return value;
```
chancancode, line 59: I think I did this because at one point the linter complained that `featureValue` is already defined in the outer scope (the function on L67). Not sure why it was complaining then and why it stopped complaining now, but it seems good to address it anyway since it can indeed be a bit confusing.

rwjblue (approved):

PR #19188, opened 2020-10-07
[CLEAN beta] remove EMBER_GLIMMER_SET_COMPONENT_TEMPLATE feature flag

Maxgit3:
Removing one feature flag


pzuraq:
Hey @Maxgit3, thank you for the PR!

It looks like there are still some references to this feature flag in the codebase. This one appears to have been split up in a confusing way, so no worries there. Here's an example of one: https://github.com/emberjs/ember.js/blob/351e8486a67a63e04dc5954b10a5e8b992c67f30/blueprints/component/index.js#L18

We should make sure that all references to the flag itself or anything with the same name are removed, along with any codepaths that are no longer used.

Maxgit3:
understood.  I removed the flag at first but that led to other conflicts in the code base.  would you also like the test removed that uses the feature?

chancancode:
The feature is enabled and mainlined, so any tests that were previously flagged should be unflagged and kept around as a regular test.

blueprints/component-class/index.js
```diff
@@ -14,7 +14,7 @@ const OCTANE = has('octane');
 // TODO: this should be reading from the @ember/canary-features module
 // need to refactor broccoli/features.js to be able to work more similarly
 // to https://github.com/emberjs/data/pull/6231
-const EMBER_GLIMMER_SET_COMPONENT_TEMPLATE = true;
+const EMBER_GLIMMER_SET_COMPONENT_TEMPLATE = false;
```
chancancode, line 17: The flag has been _enabled_ so we want to remove this variable and treat it as if it is always true in places that uses/references it.

packages/@ember/canary-features/index.ts
```diff
@@ -16,7 +16,7 @@ export const DEFAULT_FEATURES = {
   EMBER_LIBRARIES_ISREGISTERED: null,
   EMBER_IMPROVED_INSTRUMENTATION: null,
   EMBER_NAMED_BLOCKS: null,
-  EMBER_GLIMMER_SET_COMPONENT_TEMPLATE: true,
+  EMBER_GLIMMER_SET_COMPONENT_TEMPLATE: false,
```
chancancode, line 19: Same for here, we actually want to remove this line, and replace anything that uses it to have truthy semantics. You can start by replacing any references to it by with literally just the literal `true`, but eventually you'd want to simplify the code.

So, for instance... 

```js
import { EMBER_GLIMMER_SET_COMPONENT_TEMPLATE } from '@ember/canary-features';

// ...

let someThing;

if (EMBER_GLIMMER_SET_COMPONENT_TEMPLATE) {
  someThing = () => {
    doThisWhenFeatureIsEnabled();
  };
} else {
  someThing = () => {
    doThisWhenFeatureIsDisabled();
  };
}
```

becomes 

```js
// don't need the import anymore

// ...

let someThing;

if (true /* treat the flag as always enabled */) {
  someThing = () => {
    doThisWhenFeatureIsEnabled();
  };
} else {
  someThing = () => {
    doThisWhenFeatureIsDisabled();
  };
}
```

then further simplifying it...

```js
// don't need the import anymore

// ...


function someThing() {
  doThisWhenFeatureIsEnabled();
}

// the "feature disabled branch can be removed 
```

blueprints/component-class/index.js
```diff
@@ -14,7 +14,6 @@ const OCTANE = has('octane');
 // TODO: this should be reading from the @ember/canary-features module
 // need to refactor broccoli/features.js to be able to work more similarly
 // to https://github.com/emberjs/data/pull/6231
```
nlfurniss, line 16: Assume this should be removed as well?

blueprints/component/index.js
```diff
@@ -15,7 +15,6 @@ const OCTANE = has('octane');
 // TODO: this should be reading from the @ember/canary-features module
 // need to refactor broccoli/features.js to be able to work more similarly
 // to https://github.com/emberjs/data/pull/6231
```
nlfurniss, line 17: Assume this should be removed as well?

blueprints/component-class/index.js
chancancode: Yeah, so as `this.EMBER_GLIMMER_SET_COMPONENT_TEMPLATE` below should be removed entirely as well

blueprints/component/index.js
Maxgit3: sorry for the confusion.  I am new to open source

packages/@ember/canary-features/index.ts
Maxgit3: sorry for the confusion.  I am new to open source.

Maxgit3:
did you also want this.EMBER_GLIMMER_SET_COMPONENT_TEMPLATE removed from blueprints/component  and blueprints/component-class.   I also see the name in an async function  => async '@feature(EMBER_GLIMMER_SET_COMPONENT_TEMPLATE) templateOnlyComponent()'   in debug-render-tree-test.ts

kategengler:
Closing as abandoned

PR #19191, opened 2020-10-09
replace this.get('router') with this.router in RoutingService docs

rohitpaulk:
👋 

Wasn't sure what "channel" this applies to, so I just used `[DOC]` instead of `[DOC channel]`.


rwjblue (approved):
Awesome, thank you!

PR #19193, opened 2020-10-13
[BUGFIX release] Ensure user lifecycle hooks are untracked

### LINKED ISSUE #19192: Backtracking assertion "more aggressive" in 3.22
### 🐞 Describe the Bug
The autotracking.mutation-after-consumption deprecation throws an error which breaks apps instead of a warning

### 🔬 Minimal Reproduction
https://ember-twiddle.com/c677ed60d39938c97ea7438d177c0ef3?openFiles=controllers.application%5C.js%2C

The console has to be opened to see the error

### 😕 Actual Behavior
Deprecation throws an error 

### 🤔 Expected Behavior
Deprecation is a warning

### 🌍 Environment

- Ember: 3.22.0
- Node.js/npm: -
- OS: -
- Browser: -

pzuraq:
When we refactored the VM to use autotracking internally, some lifecycle hooks were tracked that previously weren't on classic components. This PR ensures those hooks are no longer tracked, maintaining their previous behavior.

Fixes #19192

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/components/curly-components-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/components/curly-components-test.js
@@ -12,7 +12,7 @@ import {
 
 import { run } from '@ember/runloop';
 import { DEBUG } from '@glimmer/env';
-import { alias, set, get, observer, on, computed } from '@ember/-internals/metal';
+import { alias, set, get, observer, on, computed, tracked } from '@ember/-internals/metal';
 import Service, { inject as injectService } from '@ember/service';
 import { Object as EmberObject, A as emberA } from '@ember/-internals/runtime';
 import { jQueryDisabled } from '@ember/-internals/views';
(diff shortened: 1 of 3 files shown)


rwjblue (commented):
Should these be untracked though? Seems like they should at least emit a deprecation, no?

rwjblue (approved):

rwjblue:
FWIW, we chatted RE: my concern above and decided that we did not want to introduce deprecations _at the moment_ for this. Specifically: resolving these deprecations would **largely** just be a waste of work (since the real path forward is migration to `@glimmer/component` which doesn't have these issues to begin with).

packages/@ember/-internals/glimmer/tests/integration/components/curly-components-test.js
```diff
@@ -3732,6 +3732,75 @@ moduleFor(
...
+            this.foo;
+            this.foo = 123;
+          }
+
+          didDestroyElement() {
+            this.foo;
+            this.foo = 123;
+          }
+        },
+        template: '{{this.baz}}',
```
xg-wang, line 3791: Dumb question when reading the diff, should it be `{{this.foo}}`?

PR #19197, opened 2020-10-13
[BUGFIX lts] Restores the shadowed property set behavior

### LINKED ISSUE #19179: [Bug] Computed property smashing regression on 3.21.3
### 🐞 Describe the Bug

In an app using ember-cli-typescript, a controller can declare the type of its `model` like this:

```js
export default class Application extends Controller {
  model: Whatever;
}
```

This works fine in 3.21.2. But using ember-source 3.21.3, it causes `this.model` to be undefined.

### 🔬 Minimal Reproduction

Run this app and see it crash: https://github.com/ef4/--bug-repro

Switch the app to use ember-source 3.21.2 and see it not crash.

Or comment out the type declaration for `model` in `app/controllers/application.ts` and see it not crash.

### 😕 Actual Behavior

The `model` computed property appears to be getting smashed by typescript's compiler output (which emits something like `Object.defineProperty(this, 'model', void 0)` in the constructor).

### 🤔 Expected Behavior

Should behave the same as in ember 3.21.2.

pzuraq:
Restores the previous behavior that would happen when setting a shadowed
property.

Fixes #19179 on release/lts

### DIFF
--- a/packages/@ember/-internals/metal/lib/decorator.ts
+++ b/packages/@ember/-internals/metal/lib/decorator.ts
@@ -1,5 +1,6 @@
 import { Meta, meta as metaFor, peekMeta } from '@ember/-internals/meta';
 import { assert } from '@ember/debug';
+import { _WeakSet } from '@ember/polyfills';
 
 export type DecoratorPropertyDescriptor = PropertyDescriptor & { initializer?: any } | undefined;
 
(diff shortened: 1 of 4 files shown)


rwjblue:
@pzuraq - Can you update the description here with cross links to the relevant threads (the PR that made this change initially, the issue that was reported after that landed, etc). As it stands, this will be hard to connect the dots in a few weeks/months.

rwjblue (approved):
Looks good to me! Please don't forget to update the description with links to the related PRs and issues...

PR #19198, opened 2020-10-13
[BUGFIX] Asserts when users attempt to set a shadowed value with Ember.set

pzuraq:
Prevents users from setting a value if it's been shadowed, which can cause inconsistencies in behavior.

### DIFF
--- a/packages/@ember/-internals/metal/lib/property_set.ts
+++ b/packages/@ember/-internals/metal/lib/property_set.ts
@@ -1,8 +1,13 @@
-import { HAS_NATIVE_PROXY, setWithMandatorySetter, toString } from '@ember/-internals/utils';
+import {
+  HAS_NATIVE_PROXY,
+  lookupDescriptor,
+  setWithMandatorySetter,
+  toString,
+} from '@ember/-internals/utils';
 import { assert } from '@ember/debug';
 import EmberError from '@ember/error';
 import { DEBUG } from '@glimmer/env';
-import { descriptorForProperty } from './decorator';
+import { CPSETTERS, descriptorForProperty } from './decorator';
 import { isPath } from './path_cache';
 import { notifyPropertyChange } from './property_events';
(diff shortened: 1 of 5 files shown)


packages/@ember/-internals/metal/lib/property_set.ts
```diff
@@ -70,6 +75,18 @@ export function set(obj: object, keyName: string, value: any, tolerant?: boolean
...
   if (descriptor !== undefined) {
+    if (DEBUG) {
+      let instanceDesc = lookupDescriptor(obj, keyName);
+      let name = instanceDesc?.set?.toString();
+
+      assert(
+        `Attempted to set \`${toString(
+          obj
+        )}.${keyName}\` using Ember.set(), but the property was a computed or tracked property that was shadowed by another property declaration. This can happen if you defined a tracked or computed property on a parent class, and then redefined it on a subclass.`,
+        name && (name.indexOf('CPSETTER_FUNCTION') !== -1 || name.indexOf('TRACKED_SETTER') !== -1)
```
rwjblue, line 86: Instead of indexing the name can we keep the weakset system (for debug only)?

rwjblue (commented):
Do you think asserting in `Ember.set` is enough? Wouldn't folks also assume a local assignment works (no set) and not hit this error?

packages/@ember/-internals/metal/lib/property_set.ts
pzuraq: we can, it's just more of a pain to maintain

pzuraq: also FWIW what we do for `CPGETTER_FUNCTION`: https://github.com/emberjs/ember.js/blob/bugfix/assert-when-setting-shadowed-property/packages/@ember/-internals/metal/lib/decorator.ts#L103

pzuraq:
@rwjblue the issue with local assignment is there's no way for us to intercept that, it's actually impossible to do unfortunately.

PR #19200, opened 2020-10-15
[BUGFIX lts] Prevents infinite rerenders when errors occur during render

pzuraq:
This adds a preventative measure against infinite rerenders that happen
due to an unrecoverable error during rendering. This type of error is
really annoying from a DX perspective and can really throw developers
for a loop. The most common cases appear to be caused by the Ember
Inspector, which polls after a render has occured whether on not it was
succesful, and in doing so also dirties state, causing a rerender.

Screen Shot 2020-10-14 at 6 49 47 PM

An example of this is seen in this [example reproduction](https://github.com/xg-wang/getter-error)
with the following steps:

1. Load the app
2. Click the To View link to enter the page with the error
3. Open up the Ember Inspector to activate it
4. Reload the page

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/renderer.ts
+++ b/packages/@ember/-internals/glimmer/lib/renderer.ts
@@ -3,6 +3,7 @@ import { getOwner, Owner } from '@ember/-internals/owner';
 import { getViewElement, getViewId, OwnedTemplateMeta } from '@ember/-internals/views';
 import { assert } from '@ember/debug';
 import { backburner, getCurrentRunLoop } from '@ember/runloop';
+import { DEBUG } from '@glimmer/env';
 import {
   Bounds,
   Cursor,
@@ -77,6 +78,34 @@ export class DynamicScope implements GlimmerDynamicScope {
   }
 }
 
+// This wrapper logic prevents us from rerendering in case of a hard failure
+// during render. This prevents infinite revalidation type loops from occuring,
+// and ensures that errors are not swallowed by subsequent follow on failures.
+function errorLoopTransaction(fn: () => void) {
+  if (DEBUG) {
+    return () => {
+      let didError = true;
+
+      try {
+        fn();
+        didError = false;
+      } finally {
+        if (didError) {
+          // Noop the function so that we won't keep calling it and causing
+          // infinite looping failures;
+          fn = () => {
+            console.warn(
+              'Attempted to rerender, but the Ember application has had an unrecoverable error occur during render. You should reload the application after fixing the cause of the error.'
+            );
+          };
+        }
+      }
+    };
+  } else {
+    return fn;
+  }
+}
+
 class RootState {
   public id: string;
   public result: RenderResult | undefined;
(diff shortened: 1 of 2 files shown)


packages/@ember/-internals/glimmer/lib/renderer.ts
```diff
@@ -77,6 +78,30 @@ export class DynamicScope implements GlimmerDynamicScope {
...
+      let didError = true;
+
+      try {
+        fn();
+        didError = false;
+      } finally {
+        if (didError) {
+          // Noop the function so that we won't keep calling it and causing
+          // infinite looping failures;
+          fn = () => {};
```
rwjblue, line 96: We should do some sort of logging to indicate that the app is now in a frozen state (rendering wise).

packages/@ember/-internals/glimmer/tests/integration/components/error-handling-test.js
```diff
@@ -1,166 +1,170 @@
...
 import { moduleFor, RenderingTestCase, runTask } from 'internal-test-helpers';
 
 import { set } from '@ember/-internals/metal';
 
 import { Component } from '../../utils/helpers';
 
-moduleFor(
-  'Errors thrown during render',
-  class extends RenderingTestCase {
-    ['@test it can recover resets the transaction when an error is thrown during initial render'](
```
rwjblue, line 10: Since this PR is not attempting to change the production experience, this test (error occurs, and still rerenders) should exist in the else (eg still tested in prod).

rwjblue (commented):
Maybe we could detect multiple "tries" before completely nooping? There are cases where an error will occur but a rerender **does** succeed. I could imagine keeping track of how many times we'd tried and errored and then bail after some arbitrary number (~ 3?)...

pzuraq:
@rwjblue the issue to me is that after a single try, we could be in a very inconsistent state. Like, we may have errored after having appended some DOM, but not the full DOM that was meant to render. What does it even mean for the render to complete at that point? Does it even make sense? Is it helpful?

Happy to discuss further though before making the decision, this was just my first attempt at solving the problem.

lifeart:
I saw same behavior for second described case in https://github.com/glimmerjs/glimmer-vm/issues/1173

rwjblue (approved):

bobisjan:
Should this prevention be also applied for production builds?

We experienced infinite rerenders in FastBoot by error in component yesterday. The bug in component was easy to fix, but infinite rerenders had negative impact to server, hence it would be desired to prevent them if possible.

rwjblue:
@bobisjan - Possibly, ya, but the issue is that 1) FastBoot environments are not expected to **ever** rerender and 2) we want whatever shred of a chance there is for resuming to work in production, which means we would have to only do this in prod for SSR environments.

bobisjan:
@rwjblue,

> 1) FastBoot environments are not expected to ever rerender

Does this mean that there may be other _unexpected_ source of the infinite rerenders in FastBoot?

rwjblue:
I think most likely we should prevent rerenders **always** from FastBoot, not just upon error.

PR #19210, opened 2020-10-16
Enable dependabot for github actions

kategengler:
This will enable dependabot for github actions. 

I am not sure how the addition of a dependabot.yml will interact with existing configuration that I assume must be in a UI somewhere? Should I add configuration for npm?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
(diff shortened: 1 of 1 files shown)


rwjblue:
I personally prefer to use the UI (https://app.dependabot.com/accounts/emberjs) for managing the settings. It is much easier to manage all of our repos at once that way...

kategengler:
Totally fine using the UI; was going based off of github docs that only indicated the config file as an option.

PR #19211, opened 2020-10-16
[RFC#236] Deprecate Ember.String.loc and {{loc}}

locks:
Part of [RFC#236 - `Ember.String` deprecation](https://github.com/emberjs/rfc-tracking/issues/26).

## Tasks
- [x] deprecate `Ember.String.loc`
  - [x] make sure deprecation URL is published and matches
- [x] deprecate `{{loc}}`

### DIFF
--- a/packages/@ember/string/index.ts
+++ b/packages/@ember/string/index.ts
@@ -110,8 +110,23 @@ function _fmt(str: string, formats: any[]) {
   @param {Array} formats Optional array of parameters to interpolate into string.
   @return {String} formatted string
   @public
+  @deprecated
 */
 export function loc(str: string, formats: any[]): string {
+  deprecate(
+    'loc is deprecated, please use a dedicated localization solution like ember-intl. More alternatives listed at https://emberobserver.com/categories/internationalization.',
+    false,
+    {
+      id: 'ember-string.loc',
+      until: '4.0.0',
+      for: 'ember-source',
+      url: 'https://deprecations.emberjs.com/v3.x#toc_ember-string-loc',
+      since: {
+        available: '3.24',
+      },
+    }
+  );
+
   if (!Array.isArray(formats) || arguments.length > 2) {
     formats = Array.prototype.slice.call(arguments, 1);
   }
@@ -324,6 +339,7 @@ if (ENV.EXTEND_PROTOTYPES.String) {
       @for @ember/string
       @static
       @private
+      @deprecated
     */
     loc: {
       configurable: true,
--- a/packages/@ember/-internals/glimmer/lib/helpers/loc.ts
+++ b/packages/@ember/-internals/glimmer/lib/helpers/loc.ts
@@ -36,6 +36,7 @@ import { helper } from '../helper';
   @param {String} str The string to format.
   @see {String#loc}
   @public
+  @deprecated
 */
 export default helper(function (params) {
   return loc.apply(null, params as any /* let the other side handle errors */);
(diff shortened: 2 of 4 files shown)


locks:
I think I'm missing something obvious in the tests but alas, I cannot figure it out at the moment. If anyone can lend a hand, please do 🙏

locks:
It seems I was not using the `expectDeprecation` wrapper correctly. The test suite should hopefully now pass.

packages/@ember/string/index.ts
```diff
@@ -111,6 +112,16 @@ function _fmt(str: string, formats: any[]) {
   @public
 */
 export function loc(str: string, formats: any[]): string {
+  deprecate('loc is deprecated.', false, {
```
rwjblue, line 115: This is pretty short and sweet. Do you think we need to say something a bit longer here?

locks: I'm finishing the deprecation guide and then I'll update the ember.js side with the final link and text!

packages/@ember/string/index.ts
```diff
@@ -111,6 +112,16 @@ function _fmt(str: string, formats: any[]) {
...
+    id: 'ember-string.loc',
+    until: '4.0.0',
+    for: '@ember/string',
+    url: 'https://deprecations.emberjs.com/v3.x#toc_ember-string-loc',
+    since: {
+      available: '3.25',
+    },
```
mehulkar, line 122: 🎉 ❤️

locks:
@rwjblue description updated. Also pinged other learning team members.

packages/@ember/string/index.ts
```diff
@@ -111,6 +112,20 @@ function _fmt(str: string, formats: any[]) {
   @public
 */
 export function loc(str: string, formats: any[]): string {
+  deprecate(
+    'loc is deprecated, please use a dedication localization solution like ember-intl. More alternatives listed at https://emberobserver.com/categories/internationalization.',
```
jaredgalanis, line 116:
```suggestion
    'loc is deprecated, please use a dedicated localization solution like ember-intl. More alternatives listed at https://emberobserver.com/categories/internationalization.',
```

jaredgalanis (approved):
Looks good 👍, one small thing.

packages/@ember/string/index.ts
```diff
@@ -111,6 +112,20 @@ function _fmt(str: string, formats: any[]) {
   @public
 */
 export function loc(str: string, formats: any[]): string {
+  deprecate(
+    'loc is deprecated, please use a dedicated localization solution like ember-intl. More alternatives listed at https://emberobserver.com/categories/internationalization.',
+    false,
+    {
+      id: 'ember-string.loc',
+      until: '4.0.0',
+      for: '@ember/string',
```
locks, line 121: Should this be `ember-source`? Do I need to update the library version of `@ember/string`? cc @rwjblue

rwjblue: Ya, should be `ember-source`. It represents the package users consume that issues the deprecation (which is ember-source).

rwjblue (changes requested):
Looks good, just needs that one change of `for` in the deprecation.

packages/@ember/string/index.ts
locks: Updated.

rwjblue (approved):

rwjblue:
@locks - Can you squash down the commits?

locks:
Rebased and squashed!

Issue #15281, opened 2017-05-25
Add assertion if `_super` not called in `willDestroy` for "FrameworkObject".


rwjblue:
This will likely have to start out as a deprecation (with `until: '2.17.0'`) and then be swapped to an assertion.

For the same reason that `init` should always have `_super` called, we should always ensure `_super` is called in `willDestroy` for `FrameworkObject`'s (`Component`, `Route`, `Controller`, `Service`).

Code for `init` not supering is [here](https://github.com/emberjs/ember.js/blob/e9e2f2b69f625716539b854f4a8934fa945faec8/packages/ember-runtime/lib/system/object.js#L71-L83).

stefanpenner:
@rwjblue 👍

rwjblue:
I had a conversation with @runspired about this, and he suggested that perhaps `destroy` is the better "required super" hook as it seems more paired with `init`...

Thoughts?

stefanpenner:
They both have purposes.

1. destroy is the "mark" phase, you best not make any changes that cause change events (but cancelling timers and stuff is phase game.
2. willDestroy is the sweep phase, you can do w/e here although it will fire "later".

rwjblue:
Ya, that makes sense (and is a good explanation of the reasoning behind the two).  I'm just wondering which we should choose for "required supering".  `destroy`, `willDestroy`, or both?

stefanpenner:
I feel we need to for both, but i could be convinced I suppose.

Turbo87:
@rwjblue I've looked into this, but it seems like there is no way to reuse the implementation that does the same thing for `init`, because while there is an `init` event that we can listen to there are no `destroy` or `willDestroy` events.

pixelhandler:
@Turbo87 @rwjblue @stefanpenner is this still an issue, perhaps we should close, what do you think?

rwjblue:
I still think this is worthwhile to do.

kategengler:
As pointed out by @johanrd in https://github.com/emberjs/ember.js/pull/21130#issuecomment-4016849697 we have solved this with lint rules.

PR #19212, opened 2020-10-17
[CLEANUP beta] Assert `_super` called in `willDestroy` for "FrameworkObject"

NoneOfMaster:
Contributes to #15281

- send `willDestroy` event from "CoreObject"
- assert `_super` is called in `willDestroy` for "FrameworkObject"

If accepted, this would close the issue except that in the comments it was decided that an assertion should be added for `destroy` as well (perhaps that could be a separate issue).

### DIFF
--- a/packages/@ember/-internals/runtime/lib/system/core_object.js
+++ b/packages/@ember/-internals/runtime/lib/system/core_object.js
@@ -369,7 +369,10 @@ class CoreObject {
     }
 
     registerDestructor(self, ensureDestroyCalled, true);
-    registerDestructor(self, () => self.willDestroy());
+    registerDestructor(self, () => {
+      self.willDestroy();
+      sendEvent(self, 'willDestroy');
+    });
 
     // disable chains
     let m = meta(self);
--- a/packages/@ember/-internals/runtime/lib/system/object.js
+++ b/packages/@ember/-internals/runtime/lib/system/object.js
@@ -7,7 +7,7 @@ import { symbol } from '@ember/-internals/utils';
 import { addListener } from '@ember/-internals/metal';
 import CoreObject from './core_object';
 import Observable from '../mixins/observable';
-import { assert } from '@ember/debug';
+import { assert, deprecate } from '@ember/debug';
 import { DEBUG } from '@glimmer/env';
 
 /**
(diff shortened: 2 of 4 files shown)


packages/@ember/-internals/runtime/lib/system/core_object.js
```diff
@@ -266,7 +266,10 @@ class CoreObject {
     }
 
     registerDestructor(self, ensureDestroyCalled, true);
-    registerDestructor(self, () => self.willDestroy());
+    registerDestructor(self, () => {
+      self.willDestroy();
+      sendEvent(self, 'willDestroy');
```
rwjblue, line 271: I'm curious, why is this added?

packages/@ember/-internals/runtime/lib/system/object.js
```diff
@@ -44,21 +44,36 @@ Observable.apply(FrameworkObject.prototype);
...
         this[INIT_WAS_CALLED]
       );
     }
+
+    [ASSERT_WILL_DESTROY_WAS_CALLED]() {
+      assert(
+        `You must call \`super.willDestroy(...arguments);\` or \`this._super(...arguments)\` when overriding \`willDestroy\` on a framework object. Please update ${this} to call \`super.willDestroy(...arguments);\` from \`willDestroy\` when using native classes or \`this._super(...arguments)\` when using \`EmberObject.extend()\`.`,
+        this[WILL_DESTROY_WAS_CALLED]
+      );
+    }
```
rwjblue, line 74: I worry that making this an assertion will be a breaking change, and I think we need to issue a deprecation for at least one LTS cycle first.

packages/@ember/-internals/runtime/lib/system/core_object.js
NoneOfMaster: This corresponds to the listener added to the FrameworkObject and is meant to follow the implementation of the preexisting handling for `init`: https://github.com/emberjs/ember.js/pull/19212/files#diff-9b45b66d303cab2e652ff010d5825e5ca583e9c658f03d1a638acf4305fab295R77-R78

packages/@ember/-internals/runtime/lib/system/object.js
NoneOfMaster: I'm happy to look into any part of this which is actionable on my part if you let me know where to start.

NoneOfMaster: Rebased and took a crack at a deprecation here instead.

kategengler:
Is this still relevant?

kategengler:
Closing this as both stale and not the direction we are looking for; this has been superceded by https://github.com/emberjs/ember.js/pull/21130#issuecomment-4016849697

PR #19218, opened 2020-10-20
[FEATURE modernized-built-in-components] First-pass implementation

chancancode:
First-pass implementation for converting the `<Input>` component into an internal (not classic) component compatibly. This passes all the existing tests.

Some follow-up work before this is ready for prime time:

- Implement any remaining classic component features not covered by the existing tests.

- Detect any unimplementable classic component features and deopt into the previous implementation.

- Write the second deprecation RFC and get the proposal merged.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/components/input.ts
+++ b/packages/@ember/-internals/glimmer/lib/components/input.ts
@@ -1,11 +1,168 @@
 /**
 @module @ember/component
 */
-import { assert } from '@ember/debug';
+import { hasDOM } from '@ember/-internals/browser-environment';
+import { tracked } from '@ember/-internals/metal';
+import { guidFor } from '@ember/-internals/utils';
+import { jQuery, jQueryDisabled } from '@ember/-internals/views';
+import { EMBER_MODERNIZED_BUILT_IN_COMPONENTS } from '@ember/canary-features';
+import { assert, deprecate, warn } from '@ember/debug';
+import {
+  JQUERY_INTEGRATION,
+  MOUSE_ENTER_LEAVE_MOVE_EVENTS,
+  SEND_ACTION,
+} from '@ember/deprecated-features';
+import { action } from '@ember/object';
 import { setComponentTemplate, setInternalComponentManager } from '@glimmer/manager';
+import { isConstRef, isUpdatableRef, Reference, updateRef, valueForRef } from '@glimmer/reference';
+import { untrack } from '@glimmer/validator';
 import InternalManager from '../component-managers/internal';
 import InputTemplate from '../templates/input';
 import InternalComponent from './internal';
+
+let isValidInputType: (type: string) => boolean;
+
+if (hasDOM && EMBER_MODERNIZED_BUILT_IN_COMPONENTS) {
+  const INPUT_TYPES: Record<string, boolean | undefined> = Object.create(null);
+  const INPUT_ELEMENT = document.createElement('input');
+
+  INPUT_TYPES[''] = false;
+  INPUT_TYPES['text'] = true;
+  INPUT_TYPES['checkbox'] = true;
+
+  isValidInputType = (type: string) => {
+    let isValid = INPUT_TYPES[type];
+
+    if (isValid === undefined) {
+      try {
+        INPUT_ELEMENT.type = type;
+        isValid = INPUT_ELEMENT.type === type;
+      } catch (e) {
+        isValid = false;
+      } finally {
+        INPUT_ELEMENT.type = 'text';
+      }
+
+      INPUT_TYPES[type] = isValid;
+    }
+
+    return isValid;
+  };
+} else {
+  isValidInputType = (type: string) => type !== '';
+}
+
+function NOOP() {}
+
+const UNINITIALIZED: unknown = Object.freeze({});
+
+/**
+ * This interface paves over the differences between these three cases:
+ *
+ * 1. `<Input />` or `<Input @value="some string" />`
+ * 2. `<Input @value={{this.value}} />`
+ * 3. `<Input @value={{to-upper-case this.value}} />`
+ *
+ * For the first set of cases (any const reference in general), the semantics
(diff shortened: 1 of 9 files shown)


chancancode:
This passes all the existing tests and is ready for review – at least it will be a useful starting point for some conversation. IMO the remaining work can be done in follow-up PRs. I'm also not sure whether we want to do a very rigious audit on what `Ember.Component` features or just do the main ones we know about and let people report bugs of any untested features. There are things like `@tagName` which _technically_ works but I don't know that anyone would do that intentionally on the `<Input>` component (that is a more valid/expected thing to do to the `<LinkTo>` component, on the other hand).

packages/@ember/-internals/glimmer/lib/components/input.ts
```diff
@@ -112,9 +269,510 @@ import InternalComponent from './internal';
...
+        !Object.getOwnPropertyDescriptor(Input.prototype, `_${argument}`)
+      );
+
+      Object.defineProperty(Input.prototype, `_${argument}`, {
+        get(this: Input): unknown {
+          deprecate(
+            `Passing the \`@${argument}\` argument to <Input> is deprecated. ` +
+              `Instead, please pass the attribute directly, i.e. \`<Input ${attribute}={{...}} />\` ` +
+              `instead of \`<Input @${argument}={{...}} />\` or \`{{input ${argument}=...}}\`.`,
+            true /* TODO !(argument in this.args) */,
```
chancancode, line 456: Actually enabling all these deprecations require fixing _lots_ of tests, so leave them disabled for now.

packages/@ember/-internals/glimmer/lib/components/input.ts
```diff
@@ -112,9 +269,510 @@ import InternalComponent from './internal';
   @public
 */
 export default class Input extends InternalComponent {
+  moderinzed = Boolean(EMBER_MODERNIZED_BUILT_IN_COMPONENTS);
```
rwjblue, line 272: typo

packages/@ember/-internals/glimmer/lib/components/input.ts
```diff
@@ -112,9 +269,510 @@ import InternalComponent from './internal';
...
+      let descriptor = Object.getOwnPropertyDescriptor(Input.prototype, argument);
+
+      if (descriptor) {
+        const superGetter = descriptor.get;
+        assert(`Expecting ${argument} to be a getter on Input`, typeof superGetter === 'function');
+
+        Object.defineProperty(Input.prototype, argument, {
+          ...descriptor,
+          get(this: Input): unknown {
+            if (attribute === 'class' && argument in this.args) {
```
rwjblue, line 478: Can you add a comment here about why `class` is hard coded?

packages/@ember/-internals/glimmer/lib/templates/input.hbs
```diff
@@ -1,7 +1,91 @@
-{{~#let (component '-checkbox') (component '-text-field') as |Checkbox TextField|~}}
-  {{~#if this.isCheckbox~}}
-    <Checkbox @target={{this.caller}} @__ARGS__={{this.args}} ...attributes />
-  {{~else~}}
-    <TextField @target={{this.caller}} @__ARGS__={{this.args}} ...attributes />
-  {{~/if~}}
-{{~/let~}}
+{{~#if this.moderinzed ~}}
```
rwjblue, line 1: typo

packages/@ember/-internals/glimmer/lib/components/input.ts
```diff
@@ -112,9 +269,510 @@ import InternalComponent from './internal';
...
+            let value = this['valueFor'].call(this, event);
+
+            if (methodName in this.args) {
+              deprecate(
+                `Passing the \`@${methodName}\` argument to <Input> is deprecated. ` +
+                  `This would have overwritten the internal \`${methodName}\` method on ` +
+                  `the <Input> component and prevented it from functioning properly. ` +
+                  `Instead, please use the {{on}} modifier, i.e. \`<Input {{on "${eventName}" ...}} />\` ` +
+                  `instead of \`<Input @${methodName}={{...}} />\` or \`{{input ${methodName}=...}}\`.`,
+                true /* TODO !descriptor */,
```
rwjblue, line 582: Can you open an issue with some guidance and then link back to it from here? Basically, we need _some_ path to enabling this deprecation...

packages/@ember/-internals/glimmer/lib/templates/input.hbs
```diff
@@ -1,7 +1,91 @@
...
+    {{on "dragstart" this._dragStart}}
+    {{on "drag" this._drag}}
+    {{on "dragenter" this._dragEnter}}
+    {{on "dragleave" this._dragLeave}}
+    {{on "dragover" this._dragOver}}
+    {{on "drop" this._drop}}
+    {{on "dragend" this._dragEnd}}
+    {{on "mouseenter" this._mouseEnter}}
+    {{on "mouseleave" this._mouseLeave}}
+    {{on "mousemove" this._mouseMove}}
```
rwjblue, line 81: We should either read the event dispatcher for its values here, or just defer to it

chancancode: I actually wrote the code to address this and submitted #19269. However, as explained in that PR, I don't want to introduce a new global name for a one-off thing we are making here. Given that we expect the contextual modifier RFC to land pretty soon and this is still behind a flag, I would prefer to punt on addressing this for now.

packages/@ember/-internals/glimmer/lib/components/input.ts
chancancode: https://github.com/emberjs/ember.js/issues/19270

pzuraq (approved):

packages/@ember/-internals/glimmer/lib/templates/input.hbs
chancancode: #19382

PR #19223, opened 2020-10-23
[BUGFIX] `<LinkTo>` should link within the engine when used inside one

chancancode:
See https://github.com/ember-engines/ember-engines/issues/692#issuecomment-714490847. This should allow ember-engines to remove the `<LinkTo>` monkey patch (at least that's the intention).

Not sure what versions to target (if any) for the backport. This could arguably be a breaking bugfix for people using engines but not using the ember-engines addon? I don't know if that combination exists in the real world though.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/components/link-to.ts
+++ b/packages/@ember/-internals/glimmer/lib/components/link-to.ts
@@ -3,8 +3,11 @@
 */
 
 import { alias, computed } from '@ember/-internals/metal';
+import { getOwner } from '@ember/-internals/owner';
+import RouterState from '@ember/-internals/routing/lib/system/router_state';
 import { isSimpleClick } from '@ember/-internals/views';
 import { assert, warn } from '@ember/debug';
+import { EngineInstance, getEngineParent } from '@ember/engine';
 import { flaggedInstrument } from '@ember/instrumentation';
 import { inject as injectService } from '@ember/service';
 import { DEBUG } from '@glimmer/env';
@@ -487,6 +490,13 @@ const LinkComponent = EmberComponent.extend({
   init() {
     this._super(...arguments);
 
+    assert(
+      'You attempted to use the <LinkTo> component within a routeless engine, this is not supported. ' +
+        'If you are using the ember-engines addon, use the <LinkToExternal> component instead. ' +
+        'See https://ember-engines.com/docs/links for more info.',
+      !this._isEngine || this._engineMountPoint !== undefined
+    );
+
     // Map desired event name to invoke function
     let { eventName } = this;
     this.on(eventName, this, this._invoke);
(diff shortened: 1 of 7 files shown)


packages/@ember/-internals/glimmer/tests/integration/application/engine-test.js
```diff
@@ -875,7 +875,7 @@ moduleFor(
 
     ["@test query params don't have stickiness by default between model"](assert) {
       assert.expect(1);
-      let tmpl = '{{#link-to "blog.category" 1337}}Category 1337{{/link-to}}';
+      let tmpl = '{{#link-to "category" 1337}}Category 1337{{/link-to}}';
```
chancancode, line 878: So this is one way you can tell this is a breaking change. Do people really use engines without the engines addon though?

villander: I think it makes ember-engines more magic than expected, I like the way how ember-engines are explicit on route path

cc: @dgeb @rwjblue

packages/@ember/-internals/glimmer/lib/components/link-to.ts
```diff
@@ -487,6 +490,13 @@ const LinkComponent = EmberComponent.extend({
   init() {
     this._super(...arguments);
 
+    assert(
```
villander, line 493: great stuff!

packages/@ember/-internals/glimmer/tests/integration/application/engine-test.js
rwjblue: @chancancode I don't think we actually support engines without the ember-engines addon. All of our documentation and guidance on usage of engines has been in context of using the addon (roughly akin to how we now are allowed to assume folks use ember-source through ember-cli).

rwjblue (approved):

rwjblue:
The issues with CI were already fixed, restarting to see if that works. It _might_ require a rebase though.

packages/@ember/-internals/glimmer/tests/integration/application/engine-test.js
villander: @rwjblue and how about don't use `engine.route-name` anymore?

chancancode:
Rebased

chriskrycho:
Heads up: I haven't had a chance to dig in yet and it's extremely unlikely I will have a chance to do so until well after the start of the year, but I have strong reason to believe this broke our app *very hard* on 3.24. Specifically, when running 3.24.0-beta.2, effectively every single test in the suite either completely fails to run or dies on instantiation of link-to's, with errors like:

> Assertion failed: You attempted to generate a link for the "company-creation.create" route, but did not pass the models required for generating its dynamic segments. There is no route named career-services.company-creation.create

When I say it hits nearly all our tests: out of ~30,000 tests, only ~6,900 run, and of those somewhere around 6,000–6,100 fail with this error. 😬

rwjblue:
@chriskrycho - ya, it requires changes in ember-engines to accomodate (otherwise we double prefix)

rwjblue:
@chriskrycho - Can you create an issue on ember-engines repo and cross link here (so we can track fixing)?

villander:
yes, we need to make changes to [LinkComponent](https://github.com/ember-engines/ember-engines/blob/master/packages/ember-engines/addon/components/link-to-component.js). Also it requires changes to [Ember Engines Guides. ](https://ember-engines.com/)

chriskrycho:
I filed a bug there; note also that we should make a point to call out the need to upgrade engines when upgrading to 3.24 in the blog post.

Issue #19074, opened 2020-08-04
Assertion/error message when typo'ing '::' as ':' in angle-bracket invocation is unhelpful/confusing


pgengler:
If, while trying to angle-bracket-invoke a nested component (e.g., `components/namespace/my-component.js`) you accidentally typo `::` as `:`, the error message that results is unhelpful for diagnosing where the problem is, and not particularly enlightening on _what_ the problem is.

Here's an example: https://ember-twiddle.com/14626a818ec6081cea2d32040a0597c1?openFiles=templates.application%5C.hbs

The error that generates (visible in the console) is `Uncaught Error: Assertion Failed: fullName must be a proper full name`; it doesn't include any file or line number information for which template had the issue (or where in the template it was), and it doesn't give any hints about what the problem might be.

Nelias:
This one has been fixed here: #19228

kategengler:
Sounds like it was fixed but also it isn't relevant in Template Tag

PR #19228, opened 2020-10-25
Fix a misleading error message in an angle bracket nested component

Nelias:
This one fixes the issue #19074 

When there is a wrong or non-existing name in path, the other function takes care of it, so I focused mostly on the :: separator. The regexp of fullName is also based on the ":" signs. The error message for non-existing component is already there: 'Cannot find component namespace/my-componen' and my implemented message fires after this one does not anymore (when the names are proper).


packages/@ember/-internals/container/lib/registry.ts
```diff
@@ -629,6 +629,9 @@ export default class Registry implements IRegistry {
     return assign({}, fallbackKnown, localKnown, resolverKnown);
   }
 
+  fullNameErrorMessage =
+    'Angle bracket nested component requires the :: separator between the names and the directory names must be in PascalCase.';
```
lifeart, line 633: wondering, could we add correct notation for it into message? or correct example (`<Foo-Bar/Baz />` should be `<FooBar::baz />`

Nelias: I thought about it but usually an `Error` message does not have examples and just points to the source of a problem, unlike `Warning`.

Nelias: @rwjblue @mehulkar @jenweber what do you think?

kategengler:
Is this still relevant?

NullVoxPopuli:
it's a good DX / clarity improvement

kategengler:
Closing because in modern ember components are imported (in gjs/gts)

Issue #19231, opened 2020-10-26
[Bug] Native Proxy's set is not triggered on a model attribute when calling Ember.set in Ember 3.21.3


andreyfel:
This is a regression which started failing since Ember 3.21.3

I have a native js Proxy which I'm using with ED Models. (Something similar to ember-changeset)
I'm wrapping a model with a proxy and intercepting get and set hooks.
Then I use Ember.set on a proxy. It used to end up in Proxy handler's set hook, but since Ember 3.21.3 it is no longer working.

Here is a unit test which is passing with ember-source@3.21.2 and failing with ember-source@3.21.3:

```
import { module, test } from 'qunit';
import { set } from '@ember/object';
import Model, { attr } from '@ember-data/model';
import { setupTest } from 'ember-qunit';

module('Unit | Utility | proxy', function(hooks) {
  setupTest(hooks);

  function createMyProxy(model) {
    return new Proxy(model, {
      set() {
        console.log('Proxy setter swallowed set!');
        return true;
      }
    });
  }

  class MyModel extends Model {
    @attr foo;
  }

  hooks.beforeEach(function() {
    this.owner.register('model:my-model', MyModel);
  });

  test('it works', function(assert) {
    const model = this.owner.lookup('service:store').createRecord('my-model', { foo: 'bar' });
    const proxy = createMyProxy(model);
    set(proxy, 'foo', 'baz');
    assert.equal(model.foo, 'bar', 'Set was swallowed by the Proxy and the underlying model is not changed');
  });
});
```

rwjblue:
This was fixed by #19232

PR #19232, opened 2020-10-26
[BUGFIX beta] Restores the shadowed property set behavior

pzuraq:
Restores the previous behavior that would happen when setting a shadowed
property.

Based on #19231 it appears that our previous fix isn't enough, and the implications
of the change were wider spread than we initially believed. So, this restores the 
previous behavior. The plan is to add a lint rule to prevent unintentional shadowing
from causing issues, e.g. `model: Foo;` should instead be `declare model: Foo;`.

### DIFF
--- a/packages/@ember/-internals/metal/lib/decorator.ts
+++ b/packages/@ember/-internals/metal/lib/decorator.ts
@@ -71,12 +71,10 @@ export abstract class ComputedDescriptor {
   abstract set(obj: object, keyName: string, value: any | null | undefined): any | null | undefined;
 }
 
-export let CPGETTERS: WeakSet<() => unknown>;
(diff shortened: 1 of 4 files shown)


rwjblue (approved):

andreyfel:
@rwjblue @pzuraq Hey guys! What is the plan here? Will it be merged? Will it be backported to 3.21.x?

pzuraq:
@andreyfel it will not be backported to 3.21, per our support policy 3.21 is no longer supported. It is fixed on 3.22, and will be fixed on 3.23+ via this PR.

rwjblue:
Right. We still need to release. It's been on my plate for a few days but I've been super swamped at work. Will try to get it done this weekend.

andreyfel:
Sure, no problem! Thanks @rwjblue!

PR #19234, opened 2020-10-27
[RFC#236] Deprecate String prototype extensions

locks:
Part of https://github.com/emberjs/rfc-tracking/issues/26.
PR twin of https://github.com/ember-learn/deprecation-app/pull/696.

This PR deprecate using the `String.prototype` extension version of `@ember/string` methods.
Calling `"Tomster and Zoey".underscore()` will throw a deprecation error.

I did not apply the wrapper method to `loc` as that is completely deprecated and has its own id.

### DIFF
--- a/packages/@ember/string/index.ts
+++ b/packages/@ember/string/index.ts
@@ -6,6 +6,7 @@ export { getStrings as _getStrings, setStrings as _setStrings } from './lib/stri
 
 import { ENV } from '@ember/-internals/environment';
 import { Cache } from '@ember/-internals/utils';
+import { deprecate } from '@ember/debug';
 import { getString } from './lib/string_registry';
 
 const STRING_DASHERIZE_REGEXP = /[ _]/g;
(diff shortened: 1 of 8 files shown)


packages/@ember/string/index.ts
```diff
@@ -279,6 +280,26 @@ export function capitalize(str: string): string {
...
 if (ENV.EXTEND_PROTOTYPES.String) {
+  let deprecateEmberStringPrototypeExtension = function (
+    name: string,
+    fn: (utility: string, ...options: any) => string | string[],
+    message = `String prototype extensions are deprecated. Please import ${name} from '@ember/string' instead.`
+  ) {
+    return function (this: string) {
+      deprecate(message, false, {
+        id: 'ember-string.prototype-extensions',
+        for: '@ember/string',
```
rwjblue, line 291: Should be 
```suggestion
        for: 'ember-source',
```

rwjblue (changes requested):
Will need to wrap the prototype extension tests in expectDeprecation

rwjblue (approved):

PR #19236, opened 2020-10-28
[BUGFIX release] Prevent multiple qp serialization on activeTransition

### LINKED ISSUE #14174: Correct Query String Serialization and Deserialization

rreckonerr:
We're using serialized query parameters for active transition instead of deserialized ones. This results in multiple serialization of qp's.

Closes #14174

### DIFF
--- a/packages/ember/tests/routing/query_params_test.js
+++ b/packages/ember/tests/routing/query_params_test.js
@@ -137,6 +137,55 @@ moduleFor(
       });
     }
 
+    ['@test Calling transitionTo does not serialize query params already serialized on the activeTransition'](
+      assert
+    ) {
+      assert.expect(3);
+
+      this.router.map(function () {
+        this.route('parent', function () {
+          this.route('child');
+          this.route('sibling');
+        });
+      });
+
+      this.add(
+        'route:parent.child',
+        Route.extend({
+          afterModel() {
+            this.transitionTo('parent.sibling');
+          },
+        })
+      );
+
+      this.add(
+        'controller:parent',
+        Controller.extend({
+          queryParams: ['array', 'string'],
+          array: [],
+          string: '',
+        })
+      );
+
+      // `/parent/child?array=["one",2]&string=hello`
+      return this.visit('/parent/child?array=%5B%22one%22%2C2%5D&string=hello').then(() => {
+        this.assertCurrentPath(
+          '/parent/sibling?array=%5B%22one%22%2C2%5D&string=hello',
+          'redirected to the sibling route, instead of child route'
+        );
+        assert.equal(
+          this.getController('parent').get('string'),
+          'hello',
+          'controller has value from the active transition'
+        );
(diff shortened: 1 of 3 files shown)


packages/ember/tests/routing/query_params_test.js
```diff
@@ -137,6 +137,55 @@ moduleFor(
...
+      this.add(
+        'controller:parent',
+        Controller.extend({
+          queryParams: ['array', 'string'],
+          array: [],
+          string: '',
+        })
+      );
+
+      // `/parent/child?array=["one",2]&string=hello`
```
rreckonerr, line 170: @chancancode @rwjblue 

This test case fails with not really helpful error:
```json
Promise rejected during " Calling transitionTo does not serialize multiple times query params already on the activeTransition": Assertion Failed: Cannot call `meta` on string
```

Which comes from `emberA` trying to convert a string to array.
```ts
emberA(JSON.parse(value as string)`
```
https://github.com/emberjs/ember.js/blob/20da5b22c9afc287dd4fef206ce7a9b3de0933d6/packages/%40ember/-internals/routing/lib/system/router.ts#L786-L797

But this error is not thrown in an ember application.

Wrapping `emberA(JSON.parse(value as string)` in a try-catch block results in failing test assertion. 

Should we do something to improve error reporting in `_deserializeQueryParam`? Like, check if `JSON.parse(value)` actually returns an array?

rwjblue: > Should we do something to improve error reporting in _deserializeQueryParam? Like, check if JSON.parse(value) actually returns an array?

Yes, I think this seems like the best path.

rwjblue (approved):

PR #19237, opened 2020-10-28
Document RFC0816 in `HistoryLocation` API docs

sclaxton:
UUIDs on history location states is a longstanding feature that has been maintained for years and is backed up by the RFC process in [RFC0816](https://github.com/emberjs/rfcs/blob/master/text/0186-track-unique-history-location-state.md). I propose we document this feature in the official docs alongside the API docs for the `HistoryLocation` class (where this feature is implemented).


rwjblue (approved):
Thank you!

rwjblue:
merged manually in 3727d0c

PR #19238, opened 2020-10-29
[CLEANUP] Removes the `expandLocalLookup` API

pzuraq:
`expandLocalLookup` was an experimental API that was tied
closely to Module Unification. The API was shipped, but was never fully
released, as it was still marked as private. It was primarily used for
experimentation in general. Currently, 3 addons use it, all of which are
experimental addons that try to establish MU like file structures.

This PR removes the API to simplify resolution as a whole, so we can
continue the work toward strict mode.


rwjblue (approved):
I do think this is somewhat breaking conceptually, but _in practice_ there are no actual consumers of this API (because we never shipped MU which was where it was going to be used).

PR #19247, opened 2020-11-02
Improvements to changelog generation script

pgengler:
This PR refactors & improves the changelog generation script, hitting the Github API to grab a PR's title when it's referenced in a merge commit but the PR title isn't part of the commit message (as is happening sometimes, causing `undefined` to be output instead of a useful message).

Since this script now hits the Github API more, I added a facility for specifying an API access token via the `GITHUB_TOKEN` environment variable. This can be useful when the number of requests exceeds the amount allowed for anonymous API access.

### DIFF
--- a/bin/changelog.js
+++ b/bin/changelog.js
@@ -19,23 +19,97 @@ const GitHubApi = require('github');
 const execSync = require('child_process').execSync;
 
 const github = new GitHubApi({ version: '3.0.0' });
+if (process.env.GITHUB_TOKEN) {
+  github.authenticate({
+    type: 'token',
+    token: process.env.GITHUB_TOKEN,
+  });
+}
 const compareCommits = RSVP.denodeify(github.repos.compareCommits);
+const getPullRequest = RSVP.denodeify(github.pullRequests.get);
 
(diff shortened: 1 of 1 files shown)


kategengler:
On the `beta` branch, running `HEAD=beta PRIOR_VERSION=v3.23.0-beta.3 ./bin/changelog.js | uniq | pbcopy` still results in a changelog with two PRs with blank titles.

bin/changelog.js
```diff
@@ -19,82 +19,61 @@ const GitHubApi = require('github');
...
 
 const github = new GitHubApi({ version: '3.0.0' });
+if (process.env.GITHUB_TOKEN) {
+  github.authenticate({
+    type: 'token',
+    token: process.env.GITHUB_TOKEN,
+  });
+}
 const compareCommits = RSVP.denodeify(github.repos.compareCommits);
+const getPullRequest = RSVP.denodeify(github.pullRequests.get);
```
rwjblue, line 29: FWIW, we can migrate to using async/await here too (seems fine as a separate PR) and use `util.promisify` from the standard lib

rwjblue (approved):

PR #19248, opened 2020-11-04
[DOC] add metadata property for RouteInfo

jamescdavis:
The `metadata` property is [present](https://github.com/tildeio/router.js/blob/d885da22340da98dcadb45427766efade54ce832/lib/router/route-info.ts#L53) on `RouteInfo`, not just `RouteInfoWithAttributes`.

Also described in the RFC: https://emberjs.github.io/rfcs/0398-RouteInfo-Metadata.html#routeinfometadata

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/route-info.ts
+++ b/packages/@ember/-internals/routing/lib/system/route-info.ts
@@ -167,6 +167,13 @@
 */
 
 /**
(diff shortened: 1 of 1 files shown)


rwjblue (approved):

rwjblue:
Thank you!

Issue #11857, opened 2015-07-21
Dynamic Segments / Query Params / Loading State Produces error with `cacheValuePrefix`


spencer516:
If a route has a Dynamic Segment, a query parameter, a Child Route with a slow-loading model and a loading template, an error occurs when a non-default query parameter is provided.

Please see http://emberjs.jsbin.com/xahiwi/edit?html,js,console,output 

Clicking on the "CLICK ME!" button produces the error `TypeError: null is not a valid argument for 'in' (evaluating 'cacheValuePrefix in values')` 

Error occurs here: https://github.com/emberjs/ember.js/blob/v1.13.5/packages/ember-routing/lib/ext/controller.js#L155 

This error does not occur in `1.13.2`, but occurs in `1.13.3` and after. 

Likely culprit is: https://github.com/emberjs/ember.js/commit/78c0edb103fb49a3a42b9d0df5cb5eedaa4b790b

rwjblue:
@jmurphyau - Do you happen to have a little time to look into this one?

jmurphyau:
Yep - will do

mike-north:
Just hit this snag as well. Any way I can help expedite a fix?

rwjblue:
@mike-north - I'd start along these lines:

1) failing test case
2) Understand what is failing and why
3) try fixing

Any or all of the above items would be super helpful.

jmurphyau:
I seen that the issue is with `cacheValuePrefix in values` and my first instinct was to check if `values` was an object to fix this problem but I wanted to investigate if `values` being null was valid behaviour or not..

I didn't want to just chuck that check in and in doing so potentially hiding the actual problem..

Haven't had a chance yet but should get to this over the next few days

jmurphyau:
So I don't think `values` should actually be a null value - but wow the scenario/code path is so complex that I've just decided to solve the issue at hand rather then trying to troubleshoot/debug why it's null. See PR #12037

tobymarsden:
The result of `value` being null is that query params are no longer sticky.

As far as I can trace back, [`transition` inside `setup` on Route](https://github.com/emberjs/ember.js/blob/v1.13.8/packages/ember-routing/lib/system/route.js#L1117) is null, when it shouldn't be.

PR #19249, opened 2020-11-05
[BUGFIX release] QP's bug for intermediate transitions

rreckonerr:
### Information
This is to make sure that https://github.com/tildeio/router.js/pull/308 covers bug that resulted in missing query params in certain scenarios. Since there's a bunch of issues related to this bug we should write and combine failing tests together in this PR and land them after https://github.com/tildeio/router.js/pull/308 is released.

### Related issues
- https://github.com/emberjs/ember.js/issues/14438
- https://github.com/emberjs/ember.js/pull/16986
- https://github.com/emberjs/ember.js/issues/12107
- https://github.com/emberjs/ember.js/issues/11857

### TODOs
- [x] Tests for link-to component (rebased #16986)
- [x] Tests for for https://github.com/emberjs/ember.js/issues/14438
- [x] Upgrade router.js

### DIFF
--- a/package.json
+++ b/package.json
@@ -137,7 +137,7 @@
     "rollup-plugin-commonjs": "^9.3.4",
     "rollup-plugin-node-resolve": "^4.2.4",
     "route-recognizer": "^0.3.4",
-    "router_js": "^7.1.0",
+    "router_js": "^7.1.1",
     "rsvp": "^4.8.5",
     "serve-static": "^1.14.1",
     "simple-dom": "^1.4.0",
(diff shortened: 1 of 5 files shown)


rwjblue:
Great work, thanks for working through this!

rreckonerr:
My pleasure :)

sly7-7:
@rwjblue The tests are not enough here, and unfortunately, there is still a bug somewhere, but I think it's hard to reproduce in tests. I've investigated a bit, and found that during an intermediate transition (at least when I have an explicit loading template), although the url is still with the correct qp values, the controller properties of these qps are still reset to their default values, and thus update the (parent) template. When the 'loading' transition is over, then the qps are back to their expected 'updated' values, and the templates update accordingly.

As a workaround, I've removed the loading template, and used a loading action to manually manage a loading flag in the parent controller, and conditionnaly display either the `{{outlet}}` or a loading component.

PR #19253, opened 2020-11-06
[RFC #481] change return of getComponentTemplate from null to undefined to align with RFC

jamescdavis:
From [RFC #481](https://github.com/emberjs/rfcs/blob/master/text/0481-component-templates-co-location.md#low-level-primitives):

> The `getComponentTemplate` function takes a component class and returns the template associated with the given component class, if any, or one of its superclasses, if any, or **`undefined`** if no template association was found.

/cc @chancancode


chancancode (approved):
@rwjblue do you think anything else may care?

pzuraq:
I this was an intentional divergence from the RFC in order to align with the the `setManager` APIs, and it would be breaking to change this.

PR #19254, opened 2020-11-07
[RFC 673] Deprecate tryInvoke

locks:
Implements [RFC #673 "Deprecate tryInvoke"](https://emberjs.github.io/rfcs/0673-deprecate-tryinvoke.html).
[Deprecation guide PR](https://github.com/ember-learn/deprecation-app/pull/705).

There seemed to be some internal usage of `tryInvoke`, so I exported an underscored tryInvoke for internal usage, and call it from the public tryInvoke.
I retrieved the deprecation URL from the preview, which should only change if the id changes.

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/components/life-cycle-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/components/life-cycle-test.js
@@ -11,7 +11,6 @@ import { schedule } from '@ember/runloop';
 import { set, setProperties } from '@ember/-internals/metal';
 import { A as emberA } from '@ember/-internals/runtime';
 import { getViewId, getViewElement, jQueryDisabled } from '@ember/-internals/views';
-import { tryInvoke } from '@ember/-internals/utils';
 
 import { Component } from '../../utils/helpers';
 
@@ -1645,9 +1644,9 @@ if (!jQueryDisabled) {
         expectDeprecation(() => {
           let comp = owner.lookup('component:foo-bar');
           runAppend(comp);
-          runTask(() => tryInvoke(component, 'destroy'));
+          runTask(() => component.destroy?.());
         }, 'Using this.$() in a component has been deprecated, consider using this.element');
-        runTask(() => tryInvoke(component, 'destroy'));
+        runTask(() => component.destroy?.());
       }
     }
   );
--- a/packages/@ember/-internals/utils/lib/invoke.ts
+++ b/packages/@ember/-internals/utils/lib/invoke.ts
@@ -1,3 +1,5 @@
+import { deprecate } from '@ember/debug';
+
 /**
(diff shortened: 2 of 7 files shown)


packages/@ember/-internals/glimmer/tests/integration/components/life-cycle-test.js
```diff
@@ -1645,9 +1645,9 @@ if (!jQueryDisabled) {
         expectDeprecation(() => {
           let comp = owner.lookup('component:foo-bar');
           runAppend(comp);
-          runTask(() => tryInvoke(component, 'destroy'));
+          runTask(() => _tryInvoke(component, 'destroy'));
         }, 'Using this.$() in a component has been deprecated, consider using this.element');
-        runTask(() => tryInvoke(component, 'destroy'));
+        runTask(() => _tryInvoke(component, 'destroy'));
```
rwjblue, line 1650: We should do the thing we tell our users to do here 😝

packages/@ember/-internals/utils/lib/invoke.ts
```diff
@@ -46,11 +48,34 @@ export function canInvoke(obj: any | null | undefined, methodName: string): obj
...
+    }
+  );
+
+  return _tryInvoke(obj, methodName, args);
+}
+
+export function _tryInvoke(
+  obj: any | undefined | null,
+  methodName: string,
+  args: Array<any | undefined | null>
```
rwjblue, line 78: Let's not split this, we should update internal usages in the same way we expect our users to update.

rwjblue (changes requested):

packages/@ember/-internals/utils/lib/invoke.ts
locks: On it. I was aiming at smallest change possible but it makes sense.

rwjblue: Ya, I'm worried if even _we_ can't follow our own guidance 😩 😝

locks:
@rwjblue code corrected!

rwjblue (commented):
Looks good, can you also update the `ember-cli-babel` version listed in `package.json` to be _at least_ the current version? (that will help ensure we are using a new enough `@babel/preset-env` that knows to transpile this properly for consuming apps)

locks:
@rwjblue good catch! I should've thought of that as I _just_ had this problem in an app this week 😂 FYI for anyone running into this, you need `@babel/preset-env` v7.8.0 or higher.
I made the ember-cli-babel update a separate commit, do you want me to manually squash it?

rwjblue:
> I made the ember-cli-babel update a separate commit, do you want me to manually squash it?

Nah. Having separate commits is great, when they each do something logical.

rwjblue (approved):

PR #19255, opened 2020-11-08
[RFC 674] Deprecate transition methods of controller and route

locks:
Implements [RFC #674 "Deprecate transition methods of controller and route"](https://emberjs.github.io/rfcs/0674-deprecate-transition-methods-of-controller-and-route.html).
Deprecation guide [online](https://deprecations.emberjs.com/v3.x/#toc_routing-transition-methods).

I kept the original behavior in the tests and wrapped them in `expectDeprecation` because I'm assuming there might be timing differences and whatnot, the purpose of the test is to exercise the specific API? Please let me know if I should adjust something.

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/application/rendering-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/application/rendering-test.js
@@ -696,7 +696,9 @@ moduleFor(
         'route:index',
         Route.extend({
           activate() {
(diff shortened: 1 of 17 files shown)


locks:
Ready to go!

locks:
@rwjblue ping x)

rwjblue (approved):

PR #19259, opened 2020-11-10
Improve internal router type

xg-wang:
Fixing few ts warnings
- `serialize` signature is based on actual [router.js type](https://github.com/tildeio/router.js/pull/316) and https://github.com/emberjs/rfcs/pull/120
- `overrideNameAssertion` and `path` appear to be mis-duplicated from `RouteOptions`


rwjblue (approved):
Nice, thank you!

PR #19261, opened 2020-11-12
[FEATURE] Upstreams the DebugRenderTree to Glimmer VM

pzuraq:
Updates to Glimmer VM 0.64.0 ([release notes](https://github.com/glimmerjs/glimmer-vm/releases/tag/v0.64.0))


packages/@ember/-internals/glimmer/tests/unit/runtime-resolver-cache-test.js
```diff
@@ -38,6 +39,8 @@ moduleFor(
           helperDefinitionCount: 1,
           // from this.render
           templateCacheMisses: 1,
+          // from debugRenderTree
+          templateCacheHits: ENV._DEBUG_RENDER_TREE ? 1 : 0,
```
pzuraq, line 43: These values changed due to the new DebugRenderTree integration looking up templates an additional time

packages/@ember/-internals/glimmer/tests/integration/application/debug-render-tree-test.ts
```diff
@@ -885,7 +885,7 @@ if (ENV._DEBUG_RENDER_TREE) {
 
         this.addComponent('hello-world', {
           ComponentClass: null,
-          template: 'Hello World',
+          template: '{{@name}}',
```
rwjblue, line 888: Why do all of these change? It isn't clear why it should have to change from the static string in order to pass the test...

packages/@ember/-internals/glimmer/index.ts
```diff
@@ -405,4 +405,3 @@ export { capabilities } from './lib/component-managers/custom';
 export { capabilities as modifierCapabilities } from './lib/modifiers/custom';
 export { helperCapabilities, HelperManager, invokeHelper } from './lib/helpers/custom';
 export { isSerializationFirstNode } from './lib/utils/serialization-first-node-helpers';
-export { CapturedRenderNode } from './lib/utils/debug-render-tree';
```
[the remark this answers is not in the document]
rwjblue: Noticed that it comes from `@glimmer/interface`

packages/@ember/-internals/glimmer/tests/integration/application/debug-render-tree-test.ts
pzuraq: So, this is caused by us setting `createInstance` === `false` in TemplateOnly components. The issue is that args are no longer pushed onto the stack unless they are _used_ within the component. So unless we use the arg with `{{@name}}`, it will not appear in the debug render tree.

We could possibly change this behavior in DEBUG only, and just have a different experience in production. I was aiming to unify production and DEBUG behavior here, which is why I made this choice.

rwjblue: We chatted a bit about this in DMs. Let's make sure we only modify the tests that are explicitly for template only components (to add the args), that should let us verify that the args are _not_ stripped in debug builds for components with a backing instance.

packages/@ember/-internals/glimmer/lib/component-managers/custom.ts
```diff
@@ -245,28 +222,12 @@ export default class CustomComponentManager<ComponentInstance>
...
-      bucket.env.extra.debugRenderTree.didRender(bucket, bounds);
-    }
-  }
-
-  didUpdateLayout(bucket: CustomComponentState<ComponentInstance>, bounds: Bounds) {
-    if (ENV._DEBUG_RENDER_TREE) {
-      bucket.env.extra.debugRenderTree.didRender(bucket, bounds);
-    }
+  getCapabilities(): ComponentCapabilities {
+    return CAPABILITIES;
```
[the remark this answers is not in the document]
pzuraq: `updateHook` is required to be able to support older custom component capabilities, unfortunately. We still support `3.4` in the component manager version, which allows users to manually opt into the `updateHook` capability.

We could potentially switch out component managers based on the capabilities that the high level component manager has, I've been holding off on doing that until we are sure it'll be worth the extra complexity

rwjblue (approved):

PR #19264, opened 2020-11-12
[DOC] Update array.filterBy example to include all properties

KrishnaRPatel:
The API Docs for `EmberArray.filterBy` mistakenly implies other properties are stripped off the returned objects. The other methods don't imply this (like `findBy` which shows in the example that the full object is returned). The docs could be confusing if a dev takes this example at face value.


locks:
Well spotted @KrishnaRPatel, thanks for sending the fix along!

PR #19267, opened 2020-11-12
[FEAT] Upstream manager infrastructure

pzuraq:
Release: https://github.com/glimmerjs/glimmer-vm/releases/tag/v0.65.0


packages/@ember/-internals/glimmer/lib/component-managers/internal.ts
```diff
@@ -41,37 +41,31 @@ export interface InternalComponentState {
 }
 
 export class InternalComponentDefinition
```
rwjblue, line 43: Calling these interfaces "internal" now seems odd (since glimmer-vm also has a set of things it calls "internal"). Not sure what the best name here would be, but something like `EmberInternal*` seems better than the current confusion (at least _my_ confusion 😜 ).

packages/@ember/-internals/glimmer/lib/helper.ts
```diff
@@ -197,7 +197,7 @@ class ClassicHelperManager implements HelperManager<ClassicHelperStateBucket> {
   }
 }
 
-setHelperManager((owner) => new ClassicHelperManager(owner), Helper);
+setHelperManager((owner: Owner | undefined) => new ClassicHelperManager(owner!), Helper);
```
rwjblue, line 200: Seems **very** suspect to update the signature here to allow undefined **and** to use `owner!`.

packages/@ember/-internals/glimmer/lib/helpers/custom.ts
```diff
@@ -430,6 +415,8 @@ export function invokeHelper(
     manager
   );
 
+  assert('Invoke helper does not support internal helpers yet', !isInternalHelper(manager));
```
rwjblue, line 418: This seems like a new addition? Should we consider backporting this?

packages/@ember/-internals/glimmer/lib/modifiers/action.ts
```diff
@@ -8,12 +8,13 @@ import {
...
   DynamicScope,
-  ModifierManager,
+  GlimmerTreeChanges,
+  InternalModifierManager,
   VMArguments,
 } from '@glimmer/interfaces';
 import { isInvokableRef, updateRef, valueForRef } from '@glimmer/reference';
 import { registerDestructor } from '@glimmer/runtime';
-import { createUpdatableTag } from '@glimmer/validator';
+import { createUpdatableTag, UpdatableTag } from '@glimmer/validator';
```
rwjblue, line 17: Not sure how you feel, but I'd love it if type only imports were done separately....

```suggestion
import { createUpdatableTag } from '@glimmer/validator';
import type { UpdatableTag } from '@glimmer/validator';
```

packages/@ember/-internals/glimmer/lib/resolver.ts
```diff
@@ -361,6 +370,11 @@ export default class RuntimeResolverImpl implements RuntimeResolver<Owner> {
         manager === SIMPLE_CLASSIC_HELPER_MANAGER
     );
 
+    assert(
+      'internal helpers are not supported via `getHelperManager` yet',
```
rwjblue, line 374: should include more information here, for example it should include the debug render tree stack and the name of the helper they asked for

packages/@ember/-internals/glimmer/lib/resolver.ts
```diff
@@ -382,7 +396,14 @@ export default class RuntimeResolverImpl implements RuntimeResolver<Owner> {
       if (modifier !== undefined) {
         let manager = getModifierManager(owner, modifier.class!);
 
-        return new CustomModifierDefinition(name, modifier, manager!, this.isInteractive);
+        assert('Expected a modifier manager to exist', manager);
```
rwjblue, line 399: Needs better error messaging (render debug stack, _any_ information about the actual thing that is `modifier.class` here, etc)

packages/@ember/-internals/glimmer/lib/resolver.ts
```diff
@@ -382,7 +396,14 @@ export default class RuntimeResolverImpl implements RuntimeResolver<Owner> {
...
+
+        assert(
+          'Internal modifier managers are not supported via `getModifierManager` yet',
```
rwjblue, line 402: More info plz (name that is being used, render debug stack, etc)

packages/@ember/-internals/glimmer/tests/integration/custom-component-manager-test.js
```diff
@@ -524,6 +525,11 @@ moduleFor(
     }
 
     '@test capabilities helper function must be used to generate capabilities'(assert) {
+      if (!DEBUG) {
+        assert.expect(0);
+        return;
+      }
```
rwjblue, line 531: Do we have a "pragma" for this (e.g. `@prod-test` or some such)?

packages/@ember/debug/index.ts
```diff
@@ -21,9 +21,7 @@ export type DebugFunctionType =
   | 'runInDebug'
   | 'deprecateFunc';
 
-export type AssertFunc =
-  | ((desc: string) => never)
```
rwjblue, line 25: >   | ((desc: string) => never)

What was this overload even for? Seems weird...

rwjblue (changes requested):

packages/@ember/-internals/glimmer/lib/component-managers/internal.ts
pzuraq: This is something that I plan on addressing in the next set of PRs, which will upstream most of the component manager logic. Part of that will be unifying internal and public component managers so they can both be used with `setComponentManager`, and in general use the same infrastructure. This way for instance, curly components will be able to be used in strict mode, since their internal manager will be associated with `setComponentTemplate`. 

As part of that work, this manager will likely be renamed `InputComponentManager` or something like that. I could rename it as part of this PR I suppose, it's just a bit weird since the resolver in Ember still has a notion of "internal", and I didn't want to get rid of that all at once either.

packages/@ember/-internals/glimmer/lib/helpers/custom.ts
pzuraq: This is a new addition, it is to allow us to associate internal helpers directly with a function via `setHelperManager`. This way we can make all of the built-in helpers work in strict mode, when you import them.

Currently, those internal helpers get resolved above, and we don't even associate any internal helpers directly with `setHelperManager`, so this assertion should never be triggered. I added it to satisfy the type constraints.

packages/@ember/-internals/glimmer/lib/modifiers/action.ts
pzuraq: I mainly add imports via autocomplete (Command + . in VSCode), so if we can set it up so that it does `type` imports for types, I'm happy with that

packages/@ember/-internals/glimmer/lib/resolver.ts
pzuraq: The debug render tree stack will be included in general, because we throw an error. There's no reason to manually add the stack, it'll always be logged regardless. Will add the name.

packages/@ember/debug/index.ts
pzuraq: I have no idea, but it was messing with the `asserts` from TS so I removed, and everything seems to work 🤷

packages/@ember/-internals/glimmer/tests/integration/custom-component-manager-test.js
pzuraq: I don't think so, but we should probably add one

packages/@ember/-internals/glimmer/lib/modifiers/action.ts
pzuraq: We also get a linting error for this, so I'm going to leave it as is for now

rwjblue (approved):

PR #19269, opened 2020-11-14
[INTERNAL] Introduce `InternalModifier`

chancancode:
This commit introduces an `InternalModifier` API. Basically like the `InternalComponent` API introduced a while back but for modifiers.

It will probably makes sense to port things like `{{on}}` to use this eventually, but right now it's unused and untested.

I originally wrote the code to address https://github.com/emberjs/ember.js/pull/19218#discussion_r520138592. However, it occurred to me that in order to use an internal modifier we currently would have to register it with some kind of global name that would leak into the same global namespace that users can invoke.

It's not a nonstarter – we have done things like that before. However, soon we will be able to take advantage of [RFC #432](https://github.com/emberjs/rfcs/blob/73685c28378118bebb5e359b80e00b839a99f622/text/0432-contextual-helpers.md) and avoid the issue. Given the expected timeline of the work, I decided to punt on that for the time being so the code remain unused for now.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/modifiers/internal.ts
+++ b/packages/@ember/-internals/glimmer/lib/modifiers/internal.ts
@@ -1,0 +1,107 @@
+import { Owner, setOwner } from '@ember/-internals/owner';
+import { guidFor } from '@ember/-internals/utils';
+import { assert } from '@ember/debug';
+import {
+  CapturedArguments,
+  Destroyable,
+  InternalModifierManager as ModifierManager,
+  VMArguments,
+} from '@glimmer/interfaces';
+import { valueForRef } from '@glimmer/reference';
+import { registerDestructor, setModifierManager } from '@glimmer/runtime';
+import { SimpleElement } from '@simple-dom/interface';
+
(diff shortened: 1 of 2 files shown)


packages/@ember/-internals/glimmer/lib/modifiers/internal.ts
```diff
@@ -0,0 +1,107 @@
...
+class InternalModifierManager
+  implements ModifierManager<InternalModifierState, typeof InternalModifier> {
+  constructor(private readonly owner: Owner) {}
+
+  create(
+    element: SimpleElement,
+    factory: typeof InternalModifier,
+    args: VMArguments
+  ): InternalModifierState {
+    assert('element must be an HTMLElement', element instanceof HTMLElement);
```
chancancode, line 72: It was a bit strange that this was a _SimpleElement_ and requires a cast, since I thought modifiers only run in the browser. Also wasn't sure if this is the canonical way to cast it these days.

pzuraq (approved):

rwjblue (approved):

PR #19271, opened 2020-11-14
Implement in-app transitions for native anchor elements so that LinkTo isn't needed

NullVoxPopuli:
Adds support for using solely `<a>` tags for in-app navigation.

Ref: https://github.com/emberjs/rfcs/pull/391

Tested using: https://github.com/NullVoxPopuli/ember-tutorial-link-component/pull/1

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/event-dispatcher-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/event-dispatcher-test.js
@@ -1,4 +1,5 @@
-import { RenderingTestCase, moduleFor, runTask } from 'internal-test-helpers';
+import { ApplicationTestCase, RenderingTestCase, moduleFor, runTask } from 'internal-test-helpers';
+import Controller from '@ember/controller';
 
 import { Component } from '../utils/helpers';
 import { getCurrentRunLoop, run } from '@ember/runloop';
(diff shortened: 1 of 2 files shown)


packages/@ember/-internals/views/lib/system/event_dispatcher.js
```diff
@@ -382,6 +382,20 @@ export default EmberObject.extend({
               if (actionHandler(target, event) === false) {
                 break;
               }
+            } else if (event.type === 'click' && target.tagName.toLowerCase() === 'a') {
+              event.preventDefault();
+              event.stopPropagation();
+
+              let isSameOrigin = window.location.origin === target.origin;
+
+              if (isSameOrigin) {
```
lifeart, line 391: this may not work if 2+ ember apps running in same page (aka "microfrontends"), should we check is route registered in app or not before transition?

NullVoxPopuli: how does this work with link-to?

NullVoxPopuli:
Questions:
 - should we prevent a click if currently loading, like: https://github.com/emberjs/ember.js/blob/8482098b10d1068ffe90fadd7343cf24d00f88d2/packages/%40ember/-internals/glimmer/lib/components/link-to.ts#L717
 - should be do nothing if currently on the same route? https://github.com/emberjs/ember.js/blob/8482098b10d1068ffe90fadd7343cf24d00f88d2/packages/%40ember/-internals/glimmer/lib/components/link-to.ts#L730
 - should this also used flaggedInstrument: https://github.com/emberjs/ember.js/blob/8482098b10d1068ffe90fadd7343cf24d00f88d2/packages/%40ember/-internals/glimmer/lib/components/link-to.ts#L741

packages/@ember/-internals/views/lib/system/event_dispatcher.js
NullVoxPopuli: maybe I should do a `recognize` before `transitionTo`

lifeart: I think link-to don't work with it, but `a` will invoke default browser behavior for same origin,  `/app/1/` ->  `/app/2`

NullVoxPopuli: updated. 

_however_. 

I'm not sure how to test this additional behavior 🤔 maybe I can use sinon, or some other stubbing on the window? idk. I'll need to look around at the other link tests

rwjblue (commented):
@bobisjan - Mind taking a look? We were mostly fleshing this out over in https://github.com/rwjblue/ember-router-helpers/pull/267 (essentially as a polyfill).

rwjblue:
I think the same issue that was mentioned in the last comment over there is an issue here too. https://emberjs.github.io/rfcs/0391-router-helpers.html#route-globs-and-route-blacklisting

packages/@ember/-internals/glimmer/tests/integration/event-dispatcher-test.js
```diff
@@ -616,6 +617,179 @@ if (jQueryDisabled) {
...
+          `
+          <nav>
+            <a class="home" href="/">Home</a>
+            <a class="about" href="/about">About</a>
+            <a class="blog" href="/blog">Blog</a>
+            <a class="post" href="/blog/1">Post</a>
+            <a class="deep-nest" href="/blog/1/some-param">Deeply Nested</a>
+            <a class="QPs" href="/blog/1/some-param?applicationQP=1&q=2&foo=3">with QPs</a>
+            <a class="QPHash" href="/blog/1/some-param?applicationQP=1&q=2&foo=3#target">with QPs and hash</a>
+            <a class="hash" href="/blog/1/some-param#target">with hash</a>
```
lifeart, line 647: should we add `bad` cases?

`<a href="javascript:void(0)" id="loginlink">login</a>`
`<a heef="#id-on-page">to some block</a>`
`<a href="#" onclick="return false;">hello</a>`

GCheung55:
I think this is a good addition and could replace [ember-href-to](https://github.com/intercom/ember-href-to) addon. However, this change does not allow the exclusion of links which has been very useful through ember-href-to where I am working on a project where a page is being migrated, allowing development with the same URL as the old page is still being supported.

E.g.
Production has `/foo/bar` served by a server.
Development app is replacing `/foo/bar`, which should be handled by the app.
On production, links to `/foo/bar` from within the Ember app is not handled as an in-app transition, allowing the old page to still be navigatable until the page is enabled/migrated.

NullVoxPopuli:
@GCheung55 if the link isn't present in your app, it'll fallback to normal anchor behavior.

Is there something else I'm missing?

mehulkar:
I thought I saw somewhere that the goal was to get rid of EventDispatcher?

GCheung55:
> @GCheung55 if the link isn't present in your app, it'll fallback to normal anchor behavior.
> 
> Is there something else I'm missing?

What I tried to describe is that the link DOES exist in the app but still needs to maintain normal anchor behavior as a path to a seamless migration from a non-Ember page to an Ember page.

I'll try to describe in terms of BDD:

GIVEN that a path /foo exists as a non-Ember page and an Ember page
AND a web server controls routing traffic to non-Ember pages and Ember pages
WHEN navigating to /foo, in a non-Ember page or Ember page
THEN a full page load occurs and loads the non-Ember page.

GIVEN that the web server is updated to direct /foo traffic to Ember pages
WHEN navigating to /foo, in a non-Ember page or Ember page
THEN a full page load occurs and loads the Ember page.

^ This is existing behavior that I can control with ember-href-to addon. Once the web server has been updated, the links to /foo in the Ember app are updated to be in-app navigation. The PR doesn't support this scenario described above.

kategengler:
Is this still relevant?

NullVoxPopuli:
it's a feature that's desperately needed, else hacks get implemented:
https://github.com/CrowdStrike/ember-url-hash-polyfill/blob/main/addon/index.ts

But maybe a new Polaris router would do this for us somehow?

kategengler:
You have an open issue to amend an RFC, so I wasn't sure if this particular implementation was still relevant.

NullVoxPopuli:
> so I wasn't sure if this particular implementation was still relevant.

I am not sure, but also, I implemented this in userland here:

https://ember-primitives.pages.dev/4-routing/proper-links

Hopefully the new router can have this built in.

PR #19293, opened 2020-12-03
Update the Glimmer VM to 0.68.0

pzuraq:
Updates the VM to the latest version (0.68.0), which includes a number of major
changes:

- Refactors the resolver to not cache values, as this is unnecessary
- Refactors the resolver to return helper/modifier definitions directly
- Refactors the resolver to return ResolvedComponentDefinitions
- Updates various component definitions
- Removes custom helper, modifier, and component managers
- Updates built-in helpers to be based on objects with associated
  helpers/helper managers.
- **Potentially breaking**: Modifier managers do not actually pass the real factory, 
  even with legacy 3.13 capabilities. The pass a look-alike instead, which is generally 
  similar, but does not include type injections. As such, if anyone was relying on type
  injections there could be breakage. This is pretty unlikely, which is why we are
  making this change, but in the event that it was something folks were using we
  should fix this.

[Notes for the VM release](https://github.com/glimmerjs/glimmer-vm/releases/tag/v0.68.0).


rwjblue (approved):
Changes look good to me overall, will need to land the last GlimmerVM PR and then update.

PR #19298, opened 2020-12-06
[Bug]: route serialize does not extract param off proxy

### LINKED ISSUE #7380: Using a belongsTo relationship in a LinkTo doesn't work
### Description

I noticed this while working on a [recent change in empress-blog](https://github.com/empress/empress-blog/pull/96) to try and workaround an issue where we [can't rely on the sort order of hasMany relationships](https://discuss.emberjs.com/t/manual-pagination-in-ember-data-relationships/18087)

Essentially this issue is quite simple to explain, and I have a feeling that it is one of those issues that quite a few people might be facing but because the workaround is so simple it likely gets overlooked. In my case it would end up being a breaking change to empress-blog and something that bleeds into the empress-blog-template ecosystem 🙈  so I would prefer to avoid the "workaround" of saying "just pass the id into the link". 

The situation is that I have a post object that has a relationship to `primaryTag` and I want to link to the "tag page" of the primaryTag (showing all other content with this same tag) and I try to do something like this: 

```
<LinkTo @route="tag" @model={{@model.tag}}>{{@model.tag.title}}</LinkTo>
```

but when you do that you get the following in the console: 

```
Uncaught Error: Assertion Failed: You attempted to generate a link for the "tag" route, but did not pass the models required for generating its dynamic segments. You must provide param `id` to `generate`.
```

### Reproduction

I created a github repo with a simple reproduction, and a PR that "introduces the bug" to the main branch 😂 

https://github.com/mansona/ember-data-bug-reproduction/pull/1

It is being deployed to netlify so you can see it doesn't work if you check the preview deploy 👍 

### Versions

```cli
➜  ember-data-relationship-link git:(main) npm list ember-source && npm list ember-cli && npm list ember-data
ember-data-relationship-link@0.0.0 /Users/mansona/temp/ember-data-relationship-link
└── ember-source@3.18.1

ember-data-relationship-link@0.0.0 /Users/mansona/temp/ember-data-relationship-link
└── ember-cli@3.18.0

ember-data-relationship-link@0.0.0 /Users/mansona/temp/ember-data-relationship-link
└── ember-data@3.18.0
```

Note: I have tested this (using ember-try) on lots of other versions of ember and ember-data, this test app just seems a little behind. Let me know if you would like me to test this in a more recent ember-data 👍

snewcomer:
close https://github.com/emberjs/data/issues/7380

`name in model` does not work with our content in a Proxy.

This does not work where `tag` is a ED relationship on `@model`.

```
<LinkTo @route="tag" @model={{@model.tag}}>{{@model.tag.title}}</LinkTo>
```

```
Uncaught Error: Assertion Failed: You attempted to generate a link for the "tag" route, but did not pass the models required for generating its dynamic segments. You must provide param `id` to `generate`.
```

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/route.ts
+++ b/packages/@ember/-internals/routing/lib/system/route.ts
@@ -18,7 +18,7 @@ import {
   Object as EmberObject,
   typeOf,
 } from '@ember/-internals/runtime';
-import { lookupDescriptor } from '@ember/-internals/utils';
+import { isProxy, lookupDescriptor } from '@ember/-internals/utils';
 import Controller from '@ember/controller';
 import { assert, deprecate, info, isTesting } from '@ember/debug';
 import { ROUTER_EVENTS } from '@ember/deprecated-features';
@@ -63,6 +63,8 @@ export function defaultSerialize(
       object[name] = get(model, name);
     } else if (/_id$/.test(name)) {
       object[name] = get(model, 'id');
+    } else if (isProxy(model)) {
+      object[name] = get(model, name);
     }
   } else {
     object = getProperties(model, params);
--- a/packages/@ember/-internals/routing/tests/system/route_test.js
+++ b/packages/@ember/-internals/routing/tests/system/route_test.js
@@ -4,6 +4,7 @@ import Service, { inject as injectService } from '@ember/service';
 import { Object as EmberObject } from '@ember/-internals/runtime';
 import EmberRoute from '../../lib/system/route';
 import { defineProperty } from '../../../metal';
+import ObjectProxy from '@ember/-internals/runtime/lib/system/object_proxy';
 
 let route, routeOne, routeTwo, lookupHash;
 
@@ -278,6 +279,12 @@ moduleFor(
       let model = { post_id: 3 };
 
       assert.deepEqual(route.serialize(model, ['post_id']), { post_id: 3 }, 'serialized correctly');
+    }
+
+    ['@test returns model.id if model is a Proxy'](assert) {
+      let model = ObjectProxy.create({ content: { id: 3 } });
+
+      assert.deepEqual(route.serialize(model, ['id']), { id: 3 }, 'serialized Proxy correctly');
     }
 
     ['@test returns undefined if model is not set'](assert) {


packages/@ember/-internals/routing/tests/system/route_test.js
```diff
@@ -280,6 +281,19 @@ moduleFor(
...
     }
 
+    ['@test returns model.id if model is a Proxy'](assert) {
+      // `key` in target should fail
+      // let model = new Proxy({ id: 3 }, {
+      //   get(target, prop) { target[prop] },
+      //   has() {
+      //     return false;
+      //   }
+      // });
```
snewcomer, line 291: I put this example here in case we wanted a more general solution vs the ObjectProxy (which more closely resembles Ember-Data relationship)

runspired (approved):

packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -62,6 +62,8 @@ export function defaultSerialize(
       object[name] = get(model, name);
     } else if (/_id$/.test(name)) {
       object[name] = get(model, 'id');
+    } else {
+      object[name] = get(model, name);
```
rwjblue, line 66: Conceptually this is pretty odd (e.g. if we were going to fall back to use `get(model, name)` then we shouldn't check `name in model` at the first branch above)

I think we could try to o something special like:

```js
} else if (isProxy(model)) {
  object[name] = get(model, name);
}
```

runspired: @rwjblue as long as the helper util doesn't check for a symbol-ish thing that EmberData can't easily make use of (e.g. it checks for the well known proxy methods) then I don't see an issue with doing it this way.

packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -63,6 +63,8 @@ export function defaultSerialize(
       object[name] = get(model, name);
     } else if (/_id$/.test(name)) {
       object[name] = get(model, 'id');
+    } else if (isProxy(model)) {
+      object[name] = get(model, name);
```
snewcomer, line 67: I could also put this in the first if expression with `&&`.   Up to you if you feel having this final else if is better.  Or perhaps it doesn't matter 😄 .

packages/@ember/-internals/routing/lib/system/route.ts
rwjblue: @runspired ya, its basically anything that uses the proxy mixin

https://github.com/emberjs/ember.js/blob/23ee5d1a8aa63678f680392e2d5f3d035873e75c/packages/%40ember/-internals/runtime/lib/mixins/-proxy.js#L76

rwjblue (approved):

PR #19302, opened 2020-12-10
[FEATURE strict-mode] Update VM for Strict Mode

pzuraq:
Updates the Glimmer VM to the latest version, which includes strict
mode. Strict mode is currently guarded behind a canary flag, but for the
most part only involves changes to the VM. The biggest changes are to
the template compiler's `compile` function, which has to change in order
to still be functional as scope values must be provided in some way, and
the resolver, which needs to provide keyword built-ins now in strict
templates.

Release notes for the VM: https://github.com/glimmerjs/glimmer-vm/releases/tag/v0.69.0


packages/ember-template-compiler/lib/system/bootstrap.ts
```diff
@@ -2,12 +2,13 @@
 @module ember
 */
 
+import { TemplateFactory } from '@glimmer/interfaces';
```
rwjblue, line 5:
```suggestion
import type { TemplateFactory } from '@glimmer/interfaces';
```

packages/ember-template-compiler/lib/system/compile-options.ts
```diff
@@ -1,14 +1,16 @@
+import { EMBER_STRICT_MODE } from '@ember/canary-features';
 import { assign } from '@ember/polyfills';
+import { PrecompileOptions } from '@glimmer/compiler';
```
rwjblue, line 3:
```suggestion
import type { PrecompileOptions } from '@glimmer/compiler';
```

packages/ember-template-compiler/lib/system/compile-options.ts
```diff
@@ -1,14 +1,16 @@
...
 import { AST, ASTPlugin, ASTPluginEnvironment, Syntax } from '@glimmer/syntax';
 import PLUGINS from '../plugins/index';
-import { CompileOptions, EmberPrecompileOptions, PluginFunc } from '../types';
+import { EmberPrecompileOptions, PluginFunc } from '../types';
```
rwjblue, line 6:
```suggestion
import type { EmberPrecompileOptions, PluginFunc } from '../types';
```

packages/ember-template-compiler/lib/system/precompile.ts
```diff
@@ -2,8 +2,8 @@
 @module ember
 */
 
-import { precompile as glimmerPrecompile, PrecompileOptions } from '@glimmer/compiler';
-import { CompileOptions } from '../types';
+import { precompile as glimmerPrecompile } from '@glimmer/compiler';
+import { EmberPrecompileOptions } from '../types';
```
rwjblue, line 6:
```suggestion
import type { EmberPrecompileOptions } from '../types';
```

packages/ember-template-compiler/lib/system/compile.ts
```diff
@@ -33,9 +36,17 @@ export default function compile(templateString: string, options: CompileOptions
...
+  let [block, usedLocals] = precompileJSON(templateSource, compileOptions(options));
+  let reifiedScopeValues = usedLocals.map((key) => scopeValues[key]);
+
+  let templateBlock: SerializedTemplateWithLazyBlock = {
+    id: String(templateId++),
+    block: JSON.stringify(block),
+    moduleName: options.moduleName ?? options.meta?.moduleName ?? '(unknown template module)',
+    scope: reifiedScopeValues.length > 0 ? () => reifiedScopeValues : null,
+    isStrictMode: options.strictMode ?? false,
+  };
```
rwjblue, line 49: I wonder if we should do this in https://github.com/emberjs/ember.js/blob/8dae0cab0a18e62117347d619b69fb2f3a4ac4f2/packages/%40ember/-internals/glimmer/tests/utils/helpers.js#L1 (since it seems this is mostly used by our own tests)

Instead of changing the external facing public API here. 🤔

packages/@ember/-internals/glimmer/tests/integration/application/debug-render-tree-test.ts
```diff
@@ -27,6 +27,10 @@ interface CapturedBounds {
   lastNode: SimpleNode;
 }
 
+function compileTemplate(templateSource: string, options: Partial<EmberPrecompileOptions>) {
+  return compile(templateSource, options) as any;
+}
```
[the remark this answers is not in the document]
pzuraq: So, `compile` currently returns, and has always returned, a TemplateFactory, not a normal Factory with a `create` method on it. But for some reason, it was typed as `any`, and so it somehow was allowed to work in these tests just assigning it directly. I just didn't want to change the tests to figure out what it should actually do, but if we know we can update it.

rwjblue (approved):

packages/@ember/-internals/glimmer/lib/components/input.ts
```diff
@@ -117,6 +118,15 @@ export default class Input extends InternalComponent {
   }
 }
 
-setInternalComponentManager(InternalManager.for('input'), Input);
+// Use an opaque handle so implementation details are
+export const InputComponent = {
+  // Factory interface
+  create(): never {
+    throw assert('Use constructor instead of create');
```
rwjblue, line 125: Throw assert seems odd? (Non blocking as it was already like this)

rwjblue (approved):

PR #19306, opened 2020-12-14
[FEATURE strict-mode] Enable classic component usage in strict mode

pzuraq:
Previously classic components were not possible to use in strict mode,
since there was no way to pass the owner to them as they were not
resolved with the factory system. This refactor fixes this by using the
new owner API, which passes the owner into the `create` hook of
component managers. In addition, makes the `renderer` property of views
a standard injection that is lazy, so we are no longer injecting it via
the implicit injection API.

Release notes for Glimmer VM: https://github.com/glimmerjs/glimmer-vm/releases/tag/v0.70.0


packages/@ember/-internals/glimmer/lib/setup-registry.ts
```diff
@@ -39,9 +39,7 @@ export function setupApplicationRegistry(registry: Registry): void {
   registry.register(P`template:-root`, RootTemplate as any);
   registry.injection('renderer', 'rootTemplate', P`template:-root`);
 
-  registry.register('renderer:-dom', InteractiveRenderer);
-  registry.register('renderer:-inert', InertRenderer);
-
+  registry.register('renderer:-main', Renderer);
```
rwjblue, line 42:
```suggestion
  registry.register('renderer:-dom', Renderer);
```

https://emberobserver.com/code-search?codeQuery=renderer%3A-dom

rwjblue (approved):

PR #19308, opened 2020-12-15
[DOC release] Use correct class declaration

mrloop:
Use `extends` when extending `Route` class.


locks (approved):
Very well spotted! Thanks for the help.

PR #19309, opened 2020-12-15
[FEATURE dynamic-helper-and-modifier] Add dynamic helper usage

pzuraq:
Adds the ability to use dynamic helper based on either helper definitions directly, or currying helpers with the `(helper)` keyword. Also adds a canary flag for this feature.


packages/ember-template-compiler/lib/plugins/transform-component-invocation.ts
```diff
@@ -172,21 +172,13 @@ function isInlineInvocation(
 }
 
 function isIllegalName(node: AST.PathExpression, hasLocal: (k: string) => boolean): boolean {
-  return isThisPath(node) || isDotPath(node) || isNamedArg(node) || isLocalVariable(node, hasLocal);
```
rwjblue, line 175: Just to confirm, removing `isThisPath` here because you are now allowed to have this paths in the VM, right?

rwjblue (commented):
Overall seems good!

We need address the gap in helper invocation (even without the `(helper` helper, you could be exposed to issues with the AST transform that rewrites `{{foo.bar 'bat'}}` to `{{component foo.bar 'bat'}}`).

pzuraq:
We were able to fix the issue with component transforms upstream and avoid it altogether, we can remove the transform entirely now 😄

rwjblue (approved):

PR #19317, opened 2020-12-18
Updates the VM in order to upstream built-ins

pzuraq:
Updates to to the latest VM version, and upstreams various built-in
keywords and modifiers. The majority of tests have been kept, but tests
which tested assertions have been removed as equivalent tests exist in
Glimmer VM now, and they introduce extra overhead without increasing
test coverage.

Release notes for the VM: https://github.com/glimmerjs/glimmer-vm/releases/tag/v0.73.0


rwjblue (approved):

Alonski:
Wanted to say that this changed causes an "issue" with the docs where they no longer point to the actual source code for various things, for example built in helpers.
Docs for fn helper: https://api.emberjs.com/ember/4.4/classes/Ember.Templates.helpers/methods/fn?anchor=fn
Clicking to the source (Defined in) takes me to: 
https://github.com/emberjs/ember.js/blob/v4.4.2/packages/%40ember/-internals/glimmer/lib/helpers/fn.ts#L5
Which isn't correct. That is "just" the types.
The source is now here:
https://github.com/glimmerjs/glimmer-vm/blob/master/packages/@glimmer/runtime/lib/helpers/fn.ts

PR #19322, opened 2020-12-22
Fix description of issue around toggleProperty

### LINKED ISSUE #19323: [Documentation] Note about toggleProperty in documentation of Controller queryParam is misleading
### 📙 Describe the Issue
Reference to `toggleProperty` in Controller queryParam documentation is confusing, and misdescribes the actual issue.

### 🔬 Minimal Reproduction
The phrase here: "By default, Ember coerces query parameter values using toggleProperty. This behavior may lead to unexpected results."
https://api.emberjs.com/ember/3.23/classes/Controller/properties/queryParams?anchor=queryParams

### 😕 Actual Behavior
This implies that Ember does something to queryParams with `toggleProperty` by default. After going through commit history I found the original issue that led to this documentation: https://github.com/emberjs/ember.js/issues/13686

From my reading of this issue, the problem is not that Ember does _anything_ with `toggleProperty` by default, but rather that Ember (reasonably) parses query parameters as strings by default. The issue is that if that a query parameter is toggled with `toggleProperty` (intentionally, by the end-user), _then_ there may be unexpected behavior because the property was interpreted as a string. This is totally different from what the documentation implies.

Recent documentation has made the default-string behavior clear, but still retains this strange note about "coercing using toggleProperty", which does not describe anything that is happening.

### 🤔 Expected Behavior
I have only briefly read through the related issues, but I've created #19322 that reflects my understanding of the gotcha, and I think is much improved over the original. If possible, this should be backported to previous versions, as it is relevant to as far back as the less-clear documentation was introduced.

cincodenada:
The original description was unclear, but implied a totally different problem was at hand. This fixes #19323

### DIFF
--- a/packages/@ember/-internals/routing/lib/ext/controller.ts
+++ b/packages/@ember/-internals/routing/lib/ext/controller.ts
@@ -15,7 +15,7 @@ ControllerMixin.reopen({
     the values of these query parameters to the variables
     `this.category` and `this.page`.
 
-    By default, query parameters are parsed and treated as strings. This
+    By default, query parameters are parsed as strings. This
     may cause unexpected behavior if a query parameter is used with `toggleProperty`,
     because the initial value set for `param=false` will be the string `"false"`, which is truthy.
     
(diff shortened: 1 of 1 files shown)


rwjblue (approved):

rwjblue:
Thank you!

PR #19330, opened 2021-01-07
[DOC lts] Remove mention on `transition-to` helper

chancancode:
The `transition-to` helper does not exists.

I suppose we could mention how to do this with the `{{on}}` modifier along with a custom action, but that would still be inappropriate for accessibility reasons in most cases. We should probably just deprecate this functionality, but in the meantime, we can probably just leave it at this.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/components/link-to.ts
+++ b/packages/@ember/-internals/glimmer/lib/components/link-to.ts
@@ -231,8 +231,7 @@ import layout from '../templates/link-to';
   </li>
   ```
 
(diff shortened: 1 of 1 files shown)


boris-petrov:
Well, there are [ember-router-helpers](https://github.com/rwjblue/ember-router-helpers)... maybe mention that instead?

rwjblue:
We tend to not mention addons in the API docs (and `transition-to` is a helper that will be provided out of the box soon).

rwjblue (approved):

PR #19336, opened 2021-01-12
[BUGFIX lts] Ensure Component Lookup Is Well Formed

chadhietala:
This PR introduces an assertion during the component name refinement to
assert that the lookup syntax is well formed. If a developer
accidentally types ":" instead of "::" you will end up with a runtime
erorr that looks like:

```
Uncaught Error: Assertion Failed: fullName must be a proper full name
    at assert (vendor.js:52729)
    at Container.lookup (vendor.js:16581)
    at Class.lookup (vendor.js:43950)
    at layoutFor (vendor.js:28680)
    at lookupComponentPair (vendor.js:28697)
    at lookupComponent (vendor.js:28711)
    at RuntimeResolver._lookupComponentDefinition (vendor.js:28992)
    at RuntimeResolver.lookupComponentHandle (vendor.js:28837)
    at CompileTimeLookup.lookupComponentDefinition (vendor.js:25065)
    at LazyCompiler.resolveLayoutForTag (vendor.js:60366)
```

This message is not actionable for developers. This PR will causes a
compile time error to occur with the following information:

```
Malformed component lookup in "test.js". Got <Foo:Bar \/> but you must use \"::\" to indicate a lookup
```

### DIFF
--- a/packages/ember-template-compiler/lib/system/compile-options.ts
+++ b/packages/ember-template-compiler/lib/system/compile-options.ts
@@ -1,4 +1,5 @@
 import { EMBER_STRICT_MODE } from '@ember/canary-features';
+import { assert } from '@ember/debug';
 import { assign } from '@ember/polyfills';
 import { PrecompileOptions } from '@glimmer/compiler';
 import { AST, ASTPlugin, ASTPluginEnvironment, Syntax } from '@glimmer/syntax';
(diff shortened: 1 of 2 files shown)


rwjblue (approved):

packages/ember-template-compiler/lib/system/compile-options.ts
```diff
@@ -16,6 +21,11 @@ export default function compileOptions(
     _options,
     {
       customizeComponentName(tagname: string): string {
+        assert(
+          `Malformed component lookup in "${_options.moduleName}". Got <${tagname} /> but you must use "::" to indicate a lookup`,
```
chancancode, line 25: I think the docs referred to this as "nested components". Should we say something like:

```suggestion
          `You tried to invoke a component named <${tagname} /> in "${_options.moduleName}", but that is not a valid name for a component. Did you mean to use the "::" syntax for nested components?`,
```

PR #15626, opened 2017-09-02
[DOC RELEASE] Move htmlSafe and isHTMLSafe to Ember.Template in documentation

locks:
For https://github.com/emberjs/rfcs/pull/236.

PR #19339, opened 2021-01-17
[BUGFIX beta] Deprecate importing htmlSafe and isHTMLSafe from @ember/string

jamescdavis:
As per [RFC #236](https://github.com/emberjs/rfcs/blob/master/text/0236-deprecation-ember-string.md), this deprecates importing `htmlSafe` and `isHTMLSafe` from `@ember/string`.

The [tracking issue for RFC #236](https://github.com/emberjs/rfc-tracking/issues/26) says that these will be deprecated via linting, but the [linked issue](https://github.com/emberjs/ember.js/pull/15626) is only a docs change that doesn't make it clear that importing from `@ember/string` is deprecated. There has been much confusion around whether this is actually deprecated (especially with regard to the `@ember/string` type definitions).

I'm hoping this can be backported to 3.24 since the the string prototype extensions were deprecated in that version.

### DIFF
--- a/packages/ember/index.js
+++ b/packages/ember/index.js
@@ -651,8 +651,36 @@ if (ENV.EXTEND_PROTOTYPES.String) {
     return htmlSafe(this);
   };
 }
-Ember.String.htmlSafe = htmlSafe;
-Ember.String.isHTMLSafe = isHTMLSafe;
+const deprecateImportFromString = function (
+  name,
+  message = `Importing ${name} from '@ember/string' is deprecated. Please import ${name} from '@ember/template' instead.`
+) {
+  deprecate(message, false, {
+    id: 'ember-string.htmlsafe-ishtmlsafe',
+    for: 'ember-source',
+    since: {
+      enabled: '3.25',
+    },
+    until: '4.0.0',
+    url: 'https://deprecations.emberjs.com/v3.x/#toc_ember-string-htmlsafe-ishtmlsafe',
+  });
+};
+Object.defineProperty(Ember.String, 'htmlSafe', {
+  enumerable: true,
+  configurable: true,
+  get() {
+    deprecateImportFromString('htmlSafe');
+    return htmlSafe;
+  },
+});
+Object.defineProperty(Ember.String, 'isHTMLSafe', {
+  enumerable: true,
+  configurable: true,
+  get() {
+    deprecateImportFromString('isHTMLSafe');
+    return isHTMLSafe;
+  },
+});
 
 /**
   Global hash of shared templates. This will automatically be populated
--- a/packages/ember/tests/reexports_test.js
+++ b/packages/ember/tests/reexports_test.js
@@ -68,8 +68,28 @@ moduleFor(
       });
     }
 
-    ['@test Ember.String.isHTMLSafe exports correctly'](assert) {
-      confirmExport(Ember, assert, 'String.isHTMLSafe', '@ember/-internals/glimmer', 'isHTMLSafe');
+    ['@test Ember.String.htmlSafe exports correctly (but deprecated)'](assert) {
+      let glimmer = require('@ember/-internals/glimmer');
+      expectDeprecation(() => {
+        assert.equal(
+          Ember.String.htmlSafe,
+          glimmer.htmlSafe,
+          'Ember.String.htmlSafe is exported correctly'
+        );
+      }, /Importing htmlSafe from '@ember\/string' is deprecated/);
+      assert.notEqual(glimmer.htmlSafe, undefined, 'Ember.String.htmlSafe is not `undefined`');
+    }
+
+    ['@test Ember.String.isHTMLSafe exports correctly (but deprecated)'](assert) {
+      let glimmer = require('@ember/-internals/glimmer');
+      expectDeprecation(() => {
+        assert.equal(
+          Ember.String.isHTMLSafe,
+          glimmer.isHTMLSafe,
+          'Ember.String.isHTMLSafe is exported correctly'
+        );
+      }, /Importing isHTMLSafe from '@ember\/string' is deprecated/);
+      assert.notEqual(glimmer.isHTMLSafe, undefined, 'Ember.String.isHTMLSafe is not `undefined`');
     }
 
     ['@test Ember.EXTEND_PROTOTYPES is present (but deprecated)'](assert) {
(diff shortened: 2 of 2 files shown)


rwjblue:
> I'm hoping this can be backported to 3.24 since the the string prototype extensions were deprecated in that version.

We typically don't backport deprecations into `release` 🤔

packages/ember/index.js
```diff
@@ -651,8 +651,27 @@ if (ENV.EXTEND_PROTOTYPES.String) {
...
+      },
+      until: '4.0.0',
+      url: 'https://deprecations.emberjs.com/v3.x/#toc_ember-string-htmlsafe-ishtmlsafe',
+    });
+
+    return fn(...arguments);
+  };
+};
+Ember.String.htmlSafe = deprecateImportFromString('htmlSafe', htmlSafe);
+Ember.String.isHTMLSafe = deprecateImportFromString('isHTMLSafe', isHTMLSafe);
```
rwjblue, line 674: Instead of returning a function that does the deprecation, can you make these a getter that issues the deprecation upon access (those getters would then return the original values)? There are other examples like that in this file.

rwjblue: That would look something like:

```js
Object.defineProperty(Ember.String, 'htmlSafe', {
  enumerable: true,
  configurable: true,
  get() {
    // do the deprecation
    // then return the normal value
  }
});
```

rwjblue (changes requested):

locks:
> We typically don't backport deprecations into release 🤔
3.24 is also an LTS candidate, if that impacts any decision.

jamescdavis:
@rwjblue I had based this on the approach used in #19234, but I like the getter approach better. Will update.

As far as backporting to 3.24, my thinking is this usage has *already* been deprecated for some time (as of the [docs change](https://github.com/emberjs/ember.js/pull/15626)) there was just no deprecation warning ever added or mention of it on http://deprecations.emberjs.com (which has caused confusion, to the point where someone has even [removed the usage from the type defs](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/50545)). It's only come to a head recently because of ember-data's new dependency on the `@ember/string` package and it's types overriding DT's, thus causing this usage to no longer typecheck (which I've [fixed](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/50653)) making people believe the usage was *removed*. As @locks mentioned, being an LTS candidate, IMHO we should get a definitive answer as to the state of this usage in that release.

jamescdavis:
As an another example of the confusion, here's a discussion from over a year ago: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38571#discussion_r328344686

packages/ember/index.js
jamescdavis: Should `htmlSafe` and `isHTMLSafe` get separate deprecation ids? I just made up `ember-string.htmlsafe-ishtmlsafe`.  Is there a pattern I should be following here?

jamescdavis: I also just invented the URL https://deprecations.emberjs.com/v3.x/#toc_ember-string-htmlsafe-ishtmlsafe
I'm not sure if there's a pattern I should follow there as well (I was looking at #19234).

rwjblue:
I'm all for providing a definitive answer, ensuring the docs have been updated and the deprecation guide has landed does that sufficiently.

Introducing deprecations in a patch release is just very bad form. For example, a decent number of apps and addon use "throw on deprecation" and releasing a new deprecation warning will start failing their CI. We have a train model for a reason, landing this now and pulling it into the current beta channel should be perfectly fjne (especially RE: sending a clear signal).

packages/ember/index.js
rwjblue: > Should `htmlSafe` and `isHTMLSafe` get separate deprecation ids? I just made up `ember-string.htmlsafe-ishtmlsafe`.  Is there a pattern I should be following here?

Dealers choice (I think either is fine)😊

rwjblue: > I also just invented the URL https://deprecations.emberjs.com/v3.x/#toc_ember-string-htmlsafe-ishtmlsafe
> 
> I'm not sure if there's a pattern I should follow there as well (I was looking at #19234).

We usually figure out the urls by making the PR and grabbing them from the running app. Once the deprecation guide PR is merged, we can ensure that the URL is correct, then we can land this.

jamescdavis:
@rwjblue that makes perfect sense! Salsify is one of those apps! 😉 
I'll update it to be since 3.25.

packages/ember/index.js
jamescdavis: Deprecation app PR: https://github.com/ember-learn/deprecation-app/pull/758

packages/ember/index.js
```diff
@@ -651,8 +651,36 @@ if (ENV.EXTEND_PROTOTYPES.String) {
...
+const deprecateImportFromString = function (
+  name,
+  message = `Importing ${name} from '@ember/string' is deprecated. Please import ${name} from '@ember/template' instead.`
+) {
+  deprecate(message, false, {
+    id: 'ember-string.htmlsafe-ishtmlsafe',
+    for: 'ember-source',
+    since: {
+      available: '3.25',
+    },
```
rwjblue, line 663: Can you set `enabled` instead of `available`?

```suggestion
    since: {
      enabled: '3.25',
    },
```


Until the deprecation stages RFC is fully rolled out, we are just using `enabled` for deprecations (it means that folks get the deprecation right away and don't have to opt in).

packages/ember/index.js
```diff
@@ -651,8 +651,36 @@ if (ENV.EXTEND_PROTOTYPES.String) {
...
+    until: '4.0.0',
+    url: 'https://deprecations.emberjs.com/v3.x/#toc_ember-string-htmlsafe-ishtmlsafe',
```
rwjblue, line 665: Now that https://github.com/ember-learn/deprecation-app/pull/758 is merged, can you double check the deployed URL to make sure this is right?

jamescdavis: Looks good! Things seem to jump around after page load so you don't end up exactly in the right spot, but the anchor  but the link *does* match what's in the TOC. This jumping issue seems to affect *all* the anchors.

locks: I think the jumping around is the code blocks rendering. I'll make an issue!

rwjblue (approved):

rwjblue:
Thank you @jamescdavis!

PR #19342, opened 2021-01-18
[BUGFIX release] Fix misleading LinkTo error message

### LINKED ISSUE #18640: Misleading error on 3.15 migration
During my Octane migration adventures I stumbled upon this error:

```
index.js:172 Uncaught Error: Assertion Failed: You attempted to generate a link 
for the "cashbook" route, but did not pass the models required for generating 
its dynamic segments. Assertion Failed: The @action decorator must be 
applied to methods when used in native classes
```

Screenshot 2019-12-27 at 22 31 47

I got confused on the first part of "generating link", but the main problem is in the "@action decorator". Once I solved this, the error on "cashbook route" went away.

I believe there are two ways to improve the DX:
 1) It looks like that whatever yields the "cashbook route" error (I assume a `LinkTo` component?) once it catches _some_ exception it wrongly assumes that the `models` were not provided.
 2) If the second error complaining about `@action` would give me a hint about the file that was incorrect, I would find the source of the problem much quicker.

-----

Foot note for whoever might suffer from the same issue: I happened to have `action` and `task` (from ember-concurrency) defined in one controller with *the same names*. After running `npx ember-native-class-codemod http://localhost:4200 ./app` codemod, their names collided in the class-level namespace.

ef4:
This fixes #18640. 

An easy way to demonstrate the original problem is to make a route that throws from `serialize`:

```js
export default class FooRoute extends Route {
  serialize() {
    throw new Error('boom');
  }
}
```

Before this change, you would see:

Screen Shot 2021-01-18 at 1 45 14 PM

Notice that both the message and the stack trace are misleading.

After this change, you would see:

Screen Shot 2021-01-18 at 1 45 31 PM

In which the message is not so misleading anymore, and the stack trace points to the real site of the problem.

I also confirmed that this still gives reasonable results for the case of failing to pass enough dynamic parameters. Before:

Screen Shot 2021-01-18 at 1 44 30 PM

And after:

Screen Shot 2021-01-18 at 1 44 49 PM

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/components/link-to.ts
+++ b/packages/@ember/-internals/glimmer/lib/components/link-to.ts
@@ -811,11 +811,8 @@ const LinkComponent = EmberComponent.extend({
           return routing.generateURL(route, models, query);
         } catch (e) {
           // tslint:disable-next-line:max-line-length
-          assert(
-            `You attempted to generate a link for the "${this.route}" route, but did not ` +
-              `pass the models required for generating its dynamic segments. ` +
-              e.message
-          );
+          e.message = `While generating link to route "${this.route}": ${e.message}`;
+          throw e;
         }
       } else {
         return routing.generateURL(route, models, query);


rwjblue (approved):

rwjblue:
Nice! Thank you!

PR #19345, opened 2021-01-20
[RFC 698] Deprecate `<LinkTo>` positional arguments

chancancode:
Implement [RFC 698](https://github.com/emberjs/rfcs/blob/master/text/0698-deprecate-link-to-positional-arguments.md)

The tests still need to be updated/fixed. I'd like to hand this off if someone else has the bandwidth to do that in the next few days so I can implement another deprecation RFC, but if no one is able to pick it up I'll come back to this after.

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/application/engine-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/application/engine-test.js
@@ -889,7 +889,9 @@ moduleFor(
 
     ["@test query params don't have stickiness by default between model"](assert) {
       assert.expect(1);
-      let tmpl = '{{#link-to "category" 1337}}Category 1337{{/link-to}}';
+
+      let tmpl = '<LinkTo @route="category" @model={{1337}}>Category 1337</LinkTo>';
+
       this.setupAppAndRoutableEngine();
       this.additionalEngineRegistrations(function () {
         this.register('template:category', compile(tmpl));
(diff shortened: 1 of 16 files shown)


packages/ember-template-compiler/lib/plugins/transform-link-to.ts
```diff
@@ -144,16 +150,52 @@ function transformPositionalLinkToIntoNamedArguments(
...
+
+    if (!hasBlock) {
+      message += " and pass a block for the link's content.";
+    }
+
+    message += '.';
+
+    if (node.loc?.source) {
+      message += ` (at ${node.loc.source} L${node.loc.start.line}:${node.loc.start.column})`;
+    }
```
chancancode, line 189: Use `calculateLocationDisplay` instead

packages/ember-template-compiler/lib/plugins/transform-link-to.ts
```diff
@@ -144,16 +150,52 @@ function transformPositionalLinkToIntoNamedArguments(
...
+
+    deprecate(message, false, {
+      id: 'ember-glimmer.link-to.positional-arguments',
+      until: '4.0.0',
+      for: 'ember-source',
+      since: {
+        enabled: '3.25.0-beta.1',
```
chancancode, line 196: Probably should be 3.26.0-beta.1

packages/@ember/-internals/glimmer/tests/integration/application/engine-test.js
```diff
@@ -928,7 +930,7 @@ moduleFor(
     ) {
       assert.expect(2);
       let tmpl =
-        '{{#link-to "author" 1337 class="author-1337"}}Author 1337{{/link-to}}{{#link-to "author" 1 class="author-1"}}Author 1{{/link-to}}';
+        '<LinkTo @route="author" @model=1337 class="author-1337">Author 1337</LinkTo><LinkTo @route="author" @model=1 class="author-1">Author 1</LinkTo>';
```
chancancode, line 933: Not that it super matters but this probably should've been translated as ={{1337}}

chancancode: Same for a couple other similar occurrences

chancancode (commented):
When it makes sense, do you mind adding [DEPRECATED] to the deprecated tests to make it easier to grep and delete them? Otherwise lgtm

PR #19346, opened 2021-01-20
[RFC 445] Deprecate `{{#with}}`

chancancode:
Also transforms `{{#with}}` into the equivalent combination of `{{#let}}` and/or `{{#if}}` to allow removing the legacy upstream code on the Glimmer VM side.

See [RFC 445](https://github.com/emberjs/rfcs/blob/master/text/0445-deprecate-with.md).

### DIFF
--- a/packages/ember-template-compiler/lib/plugins/deprecate-with.ts
+++ b/packages/ember-template-compiler/lib/plugins/deprecate-with.ts
@@ -1,0 +1,189 @@
+import { assert, deprecate } from '@ember/debug';
+import { AST, ASTPlugin } from '@glimmer/syntax';
+import calculateLocationDisplay from '../system/calculate-location-display';
+import { EmberASTPluginEnvironment } from '../types';
+import { isPath } from './utils';
+
+/**
+ @module ember
+*/
+
+/**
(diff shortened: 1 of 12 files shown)


packages/ember-template-compiler/lib/plugins/deprecate-with.ts
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        deprecate(`\`{{#with}}\` is deprecated. ${recommendation} ${sourceInformation}`, false, {
+          id: 'ember-glimmer.with-syntax',
+          until: '4.0.0',
+          for: 'ember-source',
+          since: {
+            enabled: '3.26.0-beta.1',
+          },
+        });
```
chancancode, line 115: We'd need to add a url to the deprecation guide entry. This can probably be done post-merge and I'd like to hand that (writing the deprecation guide) off to someone else so I can focus on implementing other deprecation RFCs.

rwjblue (approved):

packages/ember-template-compiler/lib/plugins/deprecate-with.ts
chancancode: @locks

PR #19348, opened 2021-01-21
[DOC] Correctly mark public link-to component named arguments

bertdeblock:
Based on [this RFC summary](https://github.com/emberjs/rfcs/blob/modernizing-built-in-components/text/0707-modernize-built-in-components-2.md#summary) and [this part of the API docs](https://api.emberjs.com/ember/release/classes/Ember.Templates.components/methods/LinkTo?anchor=LinkTo).

cc @chancancode.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/components/link-to.ts
+++ b/packages/@ember/-internals/glimmer/lib/components/link-to.ts
@@ -199,19 +199,6 @@ import layout from '../templates/link-to';
   When transitioning into the linked route, the `model` hook will
   be triggered with parameters including this passed identifier.
 
-  ### Allowing Default Action
-
-  By default the `<LinkTo>` component prevents the default browser action by calling
-  `preventDefault()` to avoid reloading the browser page.
-
-  If you need to trigger a full browser reload pass `@preventDefault={{false}}`:
-
-  ```handlebars
-  <LinkTo @route='photoGallery' @model={{this.aPhotoId}} @preventDefault={{false}}>
-    {{this.aPhotoId.title}}
-  </LinkTo>
-  ```
-
   ### Supplying a `tagName`
 
   By default `<LinkTo>` renders an `<a>` element. This can be overridden for a single use of
@@ -380,7 +367,7 @@ const LinkComponent = EmberComponent.extend({
     @property loadingClass
     @type String
     @default loading
-    @private
+    @public
   **/
   loadingClass: 'loading',
 
@@ -391,7 +378,7 @@ const LinkComponent = EmberComponent.extend({
     @property disabledClass
     @type String
     @default disabled
-    @private
+    @public
   **/
   disabledClass: 'disabled',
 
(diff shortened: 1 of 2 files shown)


packages/@ember/-internals/glimmer/lib/components/link-to.ts
```diff
@@ -406,6 +406,18 @@ const LinkComponent = EmberComponent.extend({
   **/
   replace: false,
 
+  /**
+    Determines whether the `LinkComponent` will prevent the default
```
bertdeblock, line 410: Based on the description of the `Allowing Default Action` section in [the LinkTo API docs](https://api.emberjs.com/ember/release/classes/Ember.Templates.components/methods/LinkTo?anchor=LinkTo).

chancancode: Now that I think about it, I wonder if this should just be deprecated and removed as part of RFC 707.

This seems only useful in the context of changing the tagName to something other than `<a>` (which is deprecated in RFC 707).

When the tagName is `<a>`, the default action of the browser is to navigate away, so there is no sense in not preventing it.

Thoughts?

Looping in @rwjblue as well.

bertdeblock: I'm not sure I completely understand what you mean. I've never used `@preventDefault` myself, but I can imagine it being useful for some use cases to open a route in a new tab.

bertdeblock: Wow sorry, scratch that. I was confused. Yeah I can't immediately imagine a scenario where a refresh is desired.

rwjblue: @chancancode - ya seems correct to me

rwjblue (commented):
@chancancode - So... Merge now and update the RFC RE: deprecating `preventDefault`?

chancancode:
@rwjblue @bertdeblock How about we move the docs currently in the "Allowing Default Action" section into the documentation for the property itself, and mark the property as `@private`? That way the documentation can still technically be found if someone needs it but we avoid marking it as `@public` just to have it deprecated soon after.

Reading the docs again – I suppose there is case to be made for keeping this even if `@tagName` is deprecated. In theory, you may want a link to act like an "ember link" most of the time, except when certain condition is true, _then_ you want it to actually trigger a full page reload. Something like this:

```hbs
<LinkTo @route="dashboard" @preventDefault={{this.requiresRefresh}}>
  Return to dashboard
</LinkTo>
```

Where `this.requiresRefresh` is conditionally true, perhaps when the user is a new sign up that just finished the on boarding process, something like that.

If someone wants to make the case for it, I think it wouldn't cost much to keep the functionality around. Personally, I think it's not really needed, and there are probably better ways of accomplishing that without using `<LinkTo>`, so I am still leaning towards deprecating it, but I could go either way on that.

chancancode:
I opened a PR on the RFC repo to amend this if we want to go forward: https://github.com/emberjs/rfcs/pull/710

bertdeblock:
This sounds good to me. I don't have any incentive to keep this functionality around, but of course maybe other people do. Let me know when / if I should update this PR to move the docs 👍.

chancancode:
I think you can go ahead, if we ended up wanting to make it public it's easy enough to change the tag.

bertdeblock:
Moved the docs (hopefully I understood it correctly). Let me know if anything.

chancancode (approved):
@rwjblue seems good to you?

bertdeblock:
Gentle ping. Anything left I should take care of?

chancancode (approved):
Sorry for the delay!

PR #19349, opened 2021-01-21
[DOC] Call super in Glimmer component willDestroy hook example

bertdeblock:
It seems this is also automatically added when running `eslint . --fix`. Adding it to the docs makes sense I think.

cc @ijlee2.


ijlee2:
@bertdeblock Thanks, the changes look all right to me. Let's let someone on the Ember.js team have a look and provide feedback in more detail.

rwjblue (approved):

rwjblue:
Thanks!

PR #19351, opened 2021-01-21
[BUGFIX beta] Ensure locals do not clobber components of the same name

### LINKED ISSUE #19334: [Bug beta] angle bracket invocation incorrectly shadowed by local variable with different case
```hbs
<div class="rentals">
  <ul class="results">
    {{#each @model as |rental|}}
      <li><Rental @rental={{rental}} /></li>
    {{/each}}
  </ul>
</div>
```

This template gives the following error:

```
Expected a dynamic component definition, but received an object or function that did not have a component manager associated with it. The dynamic invocation was `<(result of a `unknown` helper)>` or `{{(result of a `unknown` helper)}}`, and the incorrect definition is the value at the path `(result of a `unknown` helper)`, which was: Object
```

The issue is that `<Rental ...>` got incorrectly compiled into a symbol lookup with the same symbol ID as `|rental|`. In other words, the compiler treated the template as if it was this:

```hbs
<div class="rentals">
  <ul class="results">
    {{#each @model as |rental|}}
      <li><rental @rental={{rental}} /></li>
    {{/each}}
  </ul>
</div>
```

Naming the local variable something else that doesn't "conflict" with a component name inside the block fixes the issue. For example, this works:

```hbs
<div class="rentals">
  <ul class="results">
    {{#each @model as |property|}}
      <li><Rental @rental={{property}} /></li>
    {{/each}}
  </ul>
</div>
```

rwjblue:
Updates glimmer-vm to include https://github.com/glimmerjs/glimmer-vm/pull/1253.

Fixes https://github.com/emberjs/ember.js/issues/19334


chancancode:
Do we need a test?

I think basically:

1. register a component called `foo` (e.g. template: "foo component")
2. have a block param called `|foo|` (e.g. a string literal "foo block param")
3. Invoke `<Foo />` inside the block (should render "foo component" instead of "foo block param")

rwjblue:
Ya, good call. Added the test now.

chancancode (approved):

PR #19356, opened 2021-01-25
make `_super` property non enumerable

### LINKED ISSUE #19289: [Bug] Property "_super" is enumerable on Array, creating conflicts with external libs (ex : Quantcast Choice RGPD)
### 🐞 Describe the Bug
When extending javascript's native Array object, ember.js add a property _super to Array, and this property is enumerable, making it appear as a user property when using `for..in` on an array.

This behavior may conduct to unexpected results and so create conflicts with externals libraries.

### 🔬 Minimal Reproduction

Iterating over an array of objects with `for..in` and setting a new property to the current object will raise an error.

- Ember Twiddle: https://ember-twiddle.com/0a1aeb1142d7a07981f9203fb8836984

### 😕 Actual Behavior
This particular code will create an exception : Uncaught TypeError: can't define property "newproperty": Function is not extensible.

This is because the `for..in` loop will pass `_super` as an enumerable entry, and so we will try creating a new property name `newproperty` on the entry associated with `_super`, which is the function ROOT(), which is not extensible, and so an exception is raised.

### 🤔 Expected Behavior
`_super` should not be iterable as this is creating a strong conflict with normal JS's arrays behavior, and risk to introduce conflict with externals libs.

### 🌍 Environment

- Ember: 3.18.1 (and probably many other, originally discovered using Discourse forum which use ember.js in version 
- Node.js/npm: -
- OS: Linux (maybe others)
- Browser: Firefox, Chromium, probably others

### ➕ Additional Context
As an example of a conflict with externals libraries, i've discovered this bug because of an issue using the Quantcast's Choice.js solution for RPGD consent on a Discourse forum.

You can actually find the issue i've opened on the Discourse forum : https://meta.discourse.org/t/conflict-with-quantcast-rgpd-solution-choice-and-ember-js/170896/6

This issue offer a more detailed description of the bug, and also **a quickfix i've found to make `_super` not enumerable anymore and so fix this bug even on old versions.**

PS : I would really need someone to confirm if this is a bug or a strange but expected behavior ?

lifeart:
attemp to fix https://github.com/emberjs/ember.js/issues/19289

### DIFF
--- a/packages/@ember/-internals/metal/lib/mixin.ts
+++ b/packages/@ember/-internals/metal/lib/mixin.ts
@@ -222,7 +222,7 @@ function applyMergedProperties(
   if (hasFunction) {
     Object.defineProperty(newBase, '_super', {
       enumerable: false,
-      writable: false,
+      writable: true,
       value: ROOT,
     });
   }
@@ -427,7 +427,7 @@ export function applyMixin(obj: { [key: string]: any }, mixins: Mixin[], _hideKe
 
   Object.defineProperty(obj, '_super', {
     enumerable: false,
-    writable: false,
+    writable: true,
     value: ROOT,
   });
 
--- a/packages/@ember/-internals/metal/lib/mixin.ts
+++ b/packages/@ember/-internals/metal/lib/mixin.ts
@@ -221,6 +221,7 @@ function applyMergedProperties(
 
   if (hasFunction) {
     Object.defineProperty(newBase, '_super', {
+      configurable: true,
       enumerable: false,
       writable: true,
       value: ROOT,
@@ -426,6 +427,7 @@ export function applyMixin(obj: { [key: string]: any }, mixins: Mixin[], _hideKe
   let keysWithSuper: string[] = [];
 
   Object.defineProperty(obj, '_super', {
+    configurable: true,
     enumerable: false,
     writable: true,
     value: ROOT,
--- a/packages/@ember/-internals/metal/lib/mixin.ts
+++ b/packages/@ember/-internals/metal/lib/mixin.ts
@@ -220,7 +220,12 @@ function applyMergedProperties(
   }
 
   if (hasFunction) {
-    newBase._super = ROOT;
+    Object.defineProperty(newBase, '_super', {
+      configurable: false,
+      enumerable: false,
+      writable: false,
+      value: ROOT
+    });
   }
 
   return newBase;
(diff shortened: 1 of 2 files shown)


packages/@ember/-internals/metal/lib/mixin.ts
```diff
@@ -220,7 +220,11 @@ function applyMergedProperties(
   }
 
   if (hasFunction) {
-    newBase._super = ROOT;
+    Object.defineProperty(newBase, '_super', {
+      enumerable: false,
+      writable: true,
+      value: ROOT,
+    });
```
rwjblue, line 227: Should probably also be `configurable: true`

packages/@ember/-internals/metal/lib/mixin.ts
```diff
@@ -421,7 +425,11 @@ export function applyMixin(obj: { [key: string]: any }, mixins: Mixin[], _hideKe
   let keys: string[] = [];
   let keysWithSuper: string[] = [];
 
-  (obj as any)._super = ROOT;
+  Object.defineProperty(obj, '_super', {
+    enumerable: false,
+    writable: true,
+    value: ROOT,
```
rwjblue, line 431: Can you add `configurable: true` as well?

rwjblue (changes requested):

packages/@ember/-internals/metal/lib/mixin.ts
lifeart: good catch!

miguelcobain:
Is there anything stopping this fix?

mp-jojo:
Would be really nice to have this merged ASAP, we are trying to use Quantcast Choice, and it won't work until this fix is applied. I manually added this code change in our development environment, and it works. I am however unsure how to use my own fork when building to production, would be much easier to have this merged :)

lifeart:
@locks, @rwjblue cold we land it? or some additional changes needed?

locks:
@rwjblue could you take another look?

wycats:
In general, I'm pretty worried that changes to Ember.Object at this point (especially something like this) would fix one interaction with external code and break many others. This is especially concerning because breaking existing patterns turns working code into broken code.

What are your thoughts on that problem?

KroneckerDeIta:
I've noticed this issue with Ember.js 3.28.6 and Leaflet 1.9.3. Had to put in a temporary fix as mentioned on Discourse forum linked on associated bug.

velrest:
Would it be an option to add a flag (like we have with the prototype extension) for this so we don't have to break the existing api but can still play around with it?

jackhair:
Running in to this issue using SpeedCurve's RUM script (lux.js) on our Ember site

lbeougher:
We are having the same issue trying to run Quantcast CMP on Discourse

kategengler:
As @wycats said, we don't want to change Ember.Object at this point. Ember projects in general should be moving to native classes. 

Since this has had no movement in two years, I'm going to close.

PR #19357, opened 2021-01-25
Deprecate window global

NullVoxPopuli:


packages/ember/index.js
```diff
@@ -171,7 +171,7 @@ import {
 const Ember = (typeof context.imports.Ember === 'object' && context.imports.Ember) || {};
 
 Ember.isNamespace = true;
-Ember.toString = function () {
+Ember.toString = function() {
```
rwjblue, line 174: you'll want to roll back these unrelated prettier changes (also seems like maybe your editor was applying a different prettier version than our linting uses)

packages/ember/index.js
```diff
@@ -457,7 +457,7 @@ Ember.isEqual = isEqual;
 Ember.inject = function inject() {
   assert(
     `Injected properties must be created through helpers, see '${Object.keys(inject)
-      .map((k) => `'inject.${k}'`)
+      .map(k => `'inject.${k}'`)
```
rwjblue, line 460: 🥒

packages/ember/index.js
```diff
@@ -647,11 +647,11 @@ Ember.HTMLBars = {
 };
 
 if (ENV.EXTEND_PROTOTYPES.String) {
-  String.prototype.htmlSafe = function () {
+  String.prototype.htmlSafe = function() {
```
rwjblue, line 650: 🍉

packages/ember/index.js
```diff
@@ -647,11 +647,11 @@ Ember.HTMLBars = {
...
     return htmlSafe(this);
   };
 }
-const deprecateImportFromString = function (
+const deprecateImportFromString = function(
```
rwjblue, line 654: 🍅

packages/ember/index.js
```diff
@@ -791,6 +791,36 @@ if (IS_NODE) {
...
+        writable: false,
+        get() {
+          return Ember;
+        },
+      });
+
+      return Ember;
+    },
+  });
+}
```
rwjblue, line 822: I think this needs to be just above where `context.exports.Ember = context.exports.Em = Ember;` is.

`context.exports` _is_ `window` IIRC

rwjblue (changes requested):
I think this will also need a couple of tests

packages/ember/index.js
NullVoxPopuli: for sure, idk what happened here (very WIP PR) :p

packages/ember/index.js
NullVoxPopuli: oh! this is good to know. thanks!

pzuraq:
So had a quick convo with @rwjblue about this, and we realized we're going to need to make some additional changes. Currently, `ember-cli-babel` still transpiles `import * from '@ember/*'` to use the Ember global, and it does _not_ add `import Ember from 'ember';` at the top. In addition, we don't actually _define_ an `ember` module - that is also transpiled via `ember-cli-babel`.

So, in this package we need to add the Ember package so you can actually import it. In the babel plugin that `ember-cli-babel` adds, we need to update it to use `import Ember from 'ember';` to the top. It should also only do that IF the Ember version is high enough, so probably 3.26.0-beta.1 assuming we get this PR in before then.

NullVoxPopuli:
Next Steps?:
- [ ] update ember-cli-babel
- [ ] add deprecation guide
- [ ] merge this PR?

rwjblue (changes requested):
In order to land we need to:

* Update https://github.com/ember-cli/babel-plugin-ember-modules-api-polyfill 
  * Add a new configuration option (`useEmberModule` maybe?) to opt in to emitting the `import Ember from 'ember';` bit or doing the (currently default) "just use `window.Ember`" thing
* Update ember-cli-babel
  * to use that new version of babel-plugin-ember-modules-api-polyfill
  * to pass `true` or `false` to that new option based on the running version of `ember-source`
* Update this PR to actually emit a `ember` module (current Ember apps actually have no `ember` module at runtime, we have to fix that)

btecu:
Aren't there some important parts that have not been exposed via modules yet? How would that work with this?
For example `Ember.onError`, is used for error trackers such as Bugsnag and Sentry.

pzuraq:
@btecu as noted in the RFC, you will still be able to `import Ember from 'ember';`. This is just about accessing the object without importing it.

NullVoxPopuli:
babel plugin: https://github.com/ember-cli/babel-plugin-ember-modules-api-polyfill/pull/175

pzuraq:
@NullVoxPopuli can you rebase this PR? Now that #19390 has landed, everything should be much easier to deprecate since the global is now setup lazily (you can do it in `@ember/-internals/bootstrap`). Also we need to add a test that verifies the deprecation runs when we use the global value

rwjblue:
https://github.com/ember-cli/babel-plugin-ember-modules-api-polyfill/releases/tag/v3.3.0 includes new `useEmberModule` option. Next up: update ember-cli-babel.

simonihmig:
This has been superseded by https://github.com/emberjs/ember.js/pull/19457, so can be closed, right?

Issue #17791, opened 2019-03-22
Injecting router service into EmberRouter infinitely recurses


locks:
Code sample:

```
// app/router.js
const Router = EmberRouter.extend({
  location: config.locationType,
  rootURL: config.rootURL,
  doRouting: service('router'),

  init() {
    this._super(...arguments);

    this.doRouting.on('routeDidChange', () => {
    })
  },
```

Reproduction:

efx:
Per our discussion on discord, Injecting the router works but you must invoke `this._super(...arguments);`
https://github.com/efx/emberjs-17791/commit/8d8dd5afa1b9841742d1127a95689ab98c085632#diff-f3a289058604b2b069d07bb8e2cda60cR8

locks:
@efx that code snippet is not injecting the router service, I believe that is a different situation!

ursqontis:
I can also confirm this behaviour, on Ember 3.8

efx:
@ursqontis for the record, you can listen to the router events from within the router. You should just need to do `this.on('routeDidChange')`. This snippet has the working example: https://github.com/efx/emberjs-17791/commit/8d8dd5afa1b9841742d1127a95689ab98c085632#diff-f3a289058604b2b069d07bb8e2cda60cR8

chriskrycho:
If this is intended to be the public API for doing this in `EmberRouter`, can we document it? cc. @pzuraq @rwjblue @chadhietala? If it's *not* public API, I'm happy to write a small RFC to *make* it public API.

rwjblue:
I think it should generally be considered a bug if service injections fail. The issue in this case is that the router service itself gets router:main injected into it.

What is happening here is:

* an injection so that service:router receives router:main when it is instantiated
* router:main is looked up
* router:main looks up service:router on instantiation (this is done in debug builds for all service injections, so that we can provide a good error when a referenced service isn’t available in the system. As opposed to only erroring _if_ you happen to attempt to use the service)
* in order to instantiate service:router the container attempts to create router:main (to satisfy that injection)
* loop 

The fix here is to make the router service _not_ eagerly lookup the router:main...

chriskrycho:
@rwjblue, how hard a lift is that? I’m going to have a little time in the next few weeks when I could probably hit this if the path is clear!

rwjblue:
[...]

PR #19358, opened 2021-01-26
[Deprecation]: implicit injections

snewcomer:
Implements the implicit injections RFC

https://github.com/emberjs/rfcs/pull/680
maybe ref https://github.com/emberjs/ember.js/issues/17791

### DIFF
--- a/packages/@ember/-internals/extension-support/lib/data_adapter.js
+++ b/packages/@ember/-internals/extension-support/lib/data_adapter.js
@@ -166,6 +166,9 @@ class TypeWatcher {
 export default EmberObject.extend({
   init() {
     this._super(...arguments);
+
+    this.containerDebugAdapter = getOwner(this).lookup('container-debug-adapter:main');
+
     this.releaseMethods = emberA();
     this.recordsWatchers = new Map();
     this.typeWatchers = new Map();
@@ -181,7 +184,6 @@ export default EmberObject.extend({
     @since 1.5.0
     @public
   **/
-  containerDebugAdapter: undefined,
 
   /**
     The number of attributes to send
--- a/packages/@ember/-internals/extension-support/lib/data_adapter.js
+++ b/packages/@ember/-internals/extension-support/lib/data_adapter.js
@@ -234,6 +234,7 @@ export default EmberObject.extend({
     @private
     @property flushWatchers
     @since 3.26.0
+
    */
 
   /**
--- a/packages/@ember/-internals/extension-support/tests/data_adapter_test.js
+++ b/packages/@ember/-internals/extension-support/tests/data_adapter_test.js
@@ -15,13 +15,17 @@ const DataAdapter = EmberDataAdapter.extend({
   },
   init() {
     this._super(...arguments);
-    this.set('containerDebugAdapter', {
-      canCatalogEntriesByType() {
-        return true;
-      },
-      catalogEntriesByType() {
-        return emberA(['post']);
-      },
+    Object.defineProperty(this, 'containerDebugAdapter', {
+      get() {
+        return {
+          canCatalogEntriesByType() {
+            return true;
+          },
+          catalogEntriesByType() {
+            return emberA(['post']);
+          },
+        }
+      }
     });
   },
 });
--- a/packages/@ember/-internals/extension-support/tests/data_adapter_test.js
+++ b/packages/@ember/-internals/extension-support/tests/data_adapter_test.js
@@ -15,17 +15,13 @@ const DataAdapter = EmberDataAdapter.extend({
   },
   init() {
     this._super(...arguments);
-    Object.defineProperty(this, 'containerDebugAdapter', {
-      get() {
-        return {
-          canCatalogEntriesByType() {
-            return true;
-          },
-          catalogEntriesByType() {
-            return emberA(['post']);
-          },
-        }
-      }
+    this.set('containerDebugAdapter', {
+      canCatalogEntriesByType() {
+        return true;
+      },
+      catalogEntriesByType() {
+        return emberA(['post']);
+      },
     });
   },
 });
--- a/packages/@ember/-internals/routing/lib/ext/controller.ts
+++ b/packages/@ember/-internals/routing/lib/ext/controller.ts
@@ -1,4 +1,5 @@
 import { get } from '@ember/-internals/metal';
+import { getOwner } from '@ember/-internals/owner';
 import ControllerMixin from '@ember/controller/lib/controller_mixin';
 import { deprecateTransitionMethods, prefixRouteNameArg } from '../utils';
 
@@ -8,6 +9,15 @@ import { deprecateTransitionMethods, prefixRouteNameArg } from '../utils';
 
 ControllerMixin.reopen({
   concatenatedProperties: ['queryParams'],
+
+  init() {
+    this._super(...arguments);
+    let owner = getOwner(this);
+    if (owner) {
+      this.namespace = owner.lookup('application:main');
+      this.target = owner.lookup('router:main');
+    }
+  }
 
   /**
     Defines which query parameters the controller accepts.
--- a/packages/@ember/-internals/routing/lib/ext/controller.ts
+++ b/packages/@ember/-internals/routing/lib/ext/controller.ts
@@ -17,7 +17,7 @@ ControllerMixin.reopen({
       this.namespace = owner.lookup('application:main');
       this.target = owner.lookup('router:main');
     }
-  }
+  },
 
   /**
     Defines which query parameters the controller accepts.
--- a/packages/@ember/-internals/routing/lib/system/route.ts
+++ b/packages/@ember/-internals/routing/lib/system/route.ts
@@ -114,7 +114,10 @@ class Route extends EmberObject implements IRoute {
       let router = owner.lookup<EmberRouter>('router:main');
       let bucketCache = owner.lookup<BucketCache>(P`-bucket-cache:main`);
 
-      assert('Expected route injections to be defined', router && bucketCache);
+      assert(
+        'ROUTER BUG: Expected route injections to be defined on the route. This is an internal bug, please open an issue on Github if you see this message!',
+        router && bucketCache
+      );
 
       this._router = router;
       this._bucketCache = bucketCache;
--- a/packages/@ember/-internals/routing/lib/system/route.ts
+++ b/packages/@ember/-internals/routing/lib/system/route.ts
@@ -102,17 +102,22 @@ class Route extends EmberObject implements IRoute {
   controller!: Controller;
   currentModel: unknown;
 
-  _bucketCache: BucketCache | undefined;
+  _bucketCache!: BucketCache;
   _internalName!: string;
   _names: unknown;
-  _router: EmberRouter | undefined;
+  _router!: EmberRouter;
 
   constructor(owner: Owner) {
     super(...arguments);
 
     if (owner) {
-      this._router = owner.lookup('router:main');
-      this._bucketCache = owner.lookup(P`-bucket-cache:main`);
+      let router = owner.lookup<EmberRouter>('router:main');
+      let bucketCache = owner.lookup<BucketCache>(P`-bucket-cache:main`);
+
+      assert('Expected route injections to be defined', router && bucketCache);
+
+      this._router = router;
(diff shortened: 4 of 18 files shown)


packages/@ember/-internals/routing/lib/ext/controller.ts
```diff
@@ -9,6 +10,15 @@ import { deprecateTransitionMethods, prefixRouteNameArg } from '../utils';
 ControllerMixin.reopen({
   concatenatedProperties: ['queryParams'],
 
+  init() {
+    this._super(...arguments);
+    let owner = getOwner(this);
+    if (owner) {
+      this.namespace = owner.lookup('application:main');
+      this.target = owner.lookup('router:main');
+    }
```
snewcomer, line 19: Do we want to ensure every test has the right owner setup so we don't have to guard?

packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -98,8 +99,22 @@ class Route extends EmberObject implements IRoute {
   controller!: Controller;
   currentModel: unknown;
 
+  _bucketCache: BucketCache | undefined;
   _internalName!: string;
   _names: unknown;
+  _router: EmberRouter | undefined;
```
snewcomer, line 105: How can we force this to be just `_router: EmberRouter;`?

packages/@ember/-internals/routing/lib/ext/controller.ts
snewcomer: e.g. see `controller_test.js` where I did this

packages/@ember/-internals/utils/lib/inspect.ts
```diff
@@ -49,6 +49,9 @@ function inspectValue(value: any | null | undefined, depth: number, seen?: WeakS
       if (value.toString === objectToString || value.toString === undefined) {
         break;
       }
+      if (value.constructor && value.constructor.name !== 'Class') {
+        return value.constructor.name;
+      }
```
snewcomer, line 54: Thoughts on this?  For non anonymous classes, I found this gives us a better debug msg.  Lmk if you are ok with this and I will fix the tests.

snewcomer (commented):
Some comments for @pzuraq and @rwjblue

packages/@ember/-internals/extension-support/tests/data_adapter_test.js
```diff
@@ -15,13 +15,17 @@ const DataAdapter = EmberDataAdapter.extend({
...
   init() {
     this._super(...arguments);
-    this.set('containerDebugAdapter', {
-      canCatalogEntriesByType() {
-        return true;
-      },
-      catalogEntriesByType() {
-        return emberA(['post']);
-      },
+    Object.defineProperty(this, 'containerDebugAdapter', {
```
rwjblue, line 18: Why does this have to be `Object.defineProperty` now?

packages/@ember/-internals/extension-support/lib/data_adapter.js
```diff
@@ -53,18 +53,18 @@ export default EmberObject.extend({
...
-    @default undefined
-    @since 1.5.0
-    @public
-  **/
-  containerDebugAdapter: undefined,
+      @property containerDebugAdapter
+      @default undefined
+      @since 1.5.0
+      @public
+    **/
```
rwjblue, line 65: These docs should be moved to be after `init`

packages/@ember/-internals/extension-support/lib/data_adapter.js
```diff
@@ -53,18 +53,18 @@ export default EmberObject.extend({
...
+    this.containerDebugAdapter = getOwner(this).lookup('container-debug-adapter:main');
```
rwjblue, line 66: Should we make htis lazy (e.g. with a getter or `computed`)?

packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -98,8 +99,22 @@ class Route extends EmberObject implements IRoute {
...
+
+  constructor() {
+    super(...arguments);
+
+    let owner = getOwner(this);
```
rwjblue, line 110: I think `owner` is passed as the first argument to `constructor`, so you can use that instead of calling `getOwner(this)`.

packages/@ember/-internals/runtime/lib/system/core_object.js
```diff
@@ -73,8 +90,8 @@ function initialize(obj, properties) {
     let keyNames = Object.keys(properties);
 
     for (let i = 0; i < keyNames.length; i++) {
-      let keyName = keyNames[i];
-      let value = properties[keyName];
+      var keyName = keyNames[i];
+      var value = properties[keyName];
```
rwjblue, line 94: 🤔 no?

packages/@ember/-internals/runtime/lib/system/core_object.js
```diff
@@ -112,12 +129,48 @@ function initialize(obj, properties) {
...
+            if (desc.value !== value) {
+              // implicit injection does not match value descriptor set on object
+              implicitInjectionDeprecation(keyName, `You have defined '${keyName}' for ${inspect(obj)} as a value which does not match the implicit injection for '${keyName}'.  Please migrate to '@service ${keyName}'.`);
+            }
+          } else {
+            // wrapper getter for original adding one time deprecation
+            // TODO add setter
+            Object.defineProperty(obj, keyName, {
+              configurable: true,
+              enumerable: false,
```
rwjblue, line 164: I think we expect it to be enumerable don't we?

packages/@ember/-internals/runtime/lib/system/core_object.js
```diff
@@ -1120,4 +1173,21 @@ if (!HAS_NATIVE_SYMBOL) {
...
+      id: 'ember-object.implicit-injection',
+      until: '4.0.0',
+      url: 'https://',
+      for: 'ember-source',
+      since: {
+        enabled: '3.24.0',
+      },
+    }
+  );
+}
```
rwjblue, line 1191: I would generally prefer to just call `deprecate` multiple times, and not use this wrapper function 🤔

packages/@ember/-internals/utils/lib/inspect.ts
rwjblue: `value.constructor` is not guaranteed to be a function https://github.com/glimmerjs/glimmer-vm/issues/1210

packages/@ember/application/lib/application.js
```diff
@@ -1167,14 +1167,12 @@ function commonSetupRegistry(registry) {
...
 
-  registry.injection('router:main', 'namespace', 'application:main');
-
   registry.register('location:auto', AutoLocation);
   registry.register('location:hash', HashLocation);
   registry.register('location:history', HistoryLocation);
   registry.register('location:none', NoneLocation);
 
-  registry.register(P`-bucket-cache:main`, {
+  registry.register(`-bucket-cache:main`, {
```
rwjblue, line 1175: Why remove the `P`?

rwjblue (changes requested):

packages/@ember/-internals/runtime/lib/system/core_object.js
pzuraq: Ah the reason for this is to avoid the "would cause a new closure" error. @snewcomer I think you could fix this by hoisting the `let` outside of the loop itself, that way it's clearer where the `let` exists

packages/@ember/-internals/extension-support/lib/data_adapter.js
snewcomer: That seems like a possible route.  Went the value route to match the injection as closely as possible.  Happy to go either route

packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -98,8 +100,26 @@ class Route extends EmberObject implements IRoute {
...
+  _router!: EmberRouter;
+
+  constructor(owner: Owner) {
+    super(...arguments);
+
+    if (owner) {
+      let router = owner.lookup<EmberRouter>('router:main');
+      let bucketCache = owner.lookup<BucketCache>(P`-bucket-cache:main`);
+
+      assert('Expected route injections to be defined', router && bucketCache);
```
rwjblue, line 115: This assertion isn't really helpful messaging wise, given this message I would have no clue what it is trying to say...

packages/@ember/-internals/routing/lib/system/router.ts
```diff
@@ -1093,6 +1100,9 @@ class EmberRouter extends EmberObject {
           }
         } else {
           let cacheKey = calculateCacheKey(qp.route.fullRouteName, qp.parts, state.params);
+
+          assert('expected appCache to be defined', appCache);
```
rwjblue, line 1104: Needs better message

packages/@ember/-internals/routing/lib/system/router.ts
```diff
@@ -1420,7 +1430,9 @@ export function triggerEvent(
       } else {
         // Should only hit here if a non-bubbling error action is triggered on a route.
         if (name === 'error') {
-          handler!._router._markErrorAsHandled(args[0] as Error);
+          assert('expected handler and handler router to exist', handler && handler._router);
```
rwjblue, line 1433: tweak / cleanup this message (users would have no clue what it means / is for)

packages/@ember/-internals/runtime/lib/system/core_object.js
```diff
@@ -73,6 +90,7 @@ function initialize(obj, properties) {
     let keyNames = Object.keys(properties);
 
     for (let i = 0; i < keyNames.length; i++) {
+      // Hot path so 'var' to avoid closure, which may cause perf issues
```
rwjblue, line 93: these aren't `var` anymore, this comment should be removed

packages/@ember/-internals/runtime/lib/system/core_object.js
```diff
@@ -64,6 +66,21 @@ function initialize(obj, properties) {
...
 
+    let injectedProperties;
+    if (DEBUG) {
+      // these are all the implicit injectinos
+      injectedProperties = [];
+
+      let factory = getFactoryFor(obj);
+      if (factory) {
+        for (let injection in factory.injections) {
+          if (factory.injections[injection] === properties[injection]) {
```
rwjblue, line 77: What is this checking? What is `properties` here?

pzuraq: `properties` is `Object.assign({}, injections, createProps)` basically. So it's possible that the implicit injection has been overridden by an explicit value (an argument in a classic component for instance)

packages/@ember/-internals/runtime/tests/system/object/create_test.js
```diff
@@ -197,7 +211,9 @@ moduleFor(
       assert.equal(result.foo.bar, 'foo');
     }
 
-    ['@test does raise deprecation if descriptor is a getter and not equal to the implicit deprecation'](assert) {
+    ['@test does not raise deprecation if descriptor is a getter and not equal to the implicit deprecation'](
```
snewcomer, line 214: What this change intended?  Based on the code changes, it seems so?

packages/@ember/-internals/routing/lib/system/route.ts
pzuraq: I agree, I wasn't really meaning these assertions to be for users. They were mainly for types, and for us for internal development purposes. Since these were previously `!` assertions, my assumption was that these couldn't be issues.

I updated the assertions to be more explicit that these are internal bugs.

packages/@ember/-internals/runtime/tests/system/object/create_test.js
pzuraq: Yes, after thinking about it some more my conclusion was that we can't really assert about getters, specifically because when a user defines a getter, they also must define a setter. Otherwise, the implicit injection will do nothing (or error, if JS strict mode is enabled). With that in mind, its possible that the user could define a setter that swallows the implicit injection, handling the deprecated behavior essentially.

Happy to discuss further if that doesn't make sense

rwjblue (approved):

rwjblue (approved):

PR #19359, opened 2021-01-26
[DEPRECATION] Deprecates old browser support policy

pzuraq:
Adds the deprecation specified in https://github.com/emberjs/rfcs/blob/master/text/0685-new-browser-support-policy.md

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -246,6 +246,12 @@ module.exports = {
     let ember;
     let targets = (this.project && this.project.targets && this.project.targets.browsers) || [];
 
+    if (targets.includes('ie 11')) {
+      this.ui.writeWarnLine(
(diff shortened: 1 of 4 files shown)


rwjblue (approved):
Updated to fix the deprecation predicate (a falsey value issues a deprecation, so it was previously deprecating on all _but_ IE11).

rwjblue:
Will need to be updated to avoid issuing the deprecation on IE11 (or we need to make the "fail on deprecation" setup ignore this specific id).

Also, seems like we have an issue in Node / SSR. The error that is being thrown seems wrong though (I suspect an earlier deprecation causes the assets to be loaded incorrectly, but haven't verified).

sandstrom:
Good bye IE11. You won't be missed 😄 

f3faa371b84b0703c1a53bb0395a3509

Issue #19360, opened 2021-01-27
[Bug] 3.25.0-beta.3: curly component names are normalized differently


dfreeman:
### 🐞 Describe the Bug
In `ember-source@3.25.0-beta.2` and prior, a component invoked in block form like `{{#pkg::component-name}}` would be looked up via the resolver as `pkg::component-name`.

Starting with `beta.3` with the upgrade to `glimmer-vm@0.73.2`, that name is normalized via `customizeComponentName` to become `pkg/component-name`. For a customized resolver where `::` is a meaningful token, such components no longer resolve correctly.

### 🔬 Minimal Reproduction

The following is resolved as `component:pkg::component-name` in beta.2, but `component:pkg/component-name` in beta.3:

```hbs
{{#pkg::component-name}}
  hello!
{{/pkg::component-name}}
```

Note that an inline invocation like `{{pkg::component-name}}` does not go through `customizeComponentName` and so still resolves as before.

### 😕 Actual Behavior
Block curly component names have `::` normalized to `/`.

### 🤔 Expected Behavior
Block curly component names aren't subjected to the same normalization rules as angle-bracket components, as in prior releases.

### 🌍 Environment

- Ember: 3.25.0-beta.3

### ➕ Additional Context

Additional discussion of the details of this issue is at https://github.com/glimmerjs/glimmer-vm/pull/1255

rwjblue:
Fixed by https://github.com/emberjs/ember.js/pull/19363

PR #19363, opened 2021-01-30
[BUGFIX beta] Update VM, fix component name preprocessing

pzuraq:


packages/@ember/-internals/glimmer/lib/environment.ts
```diff
@@ -50,29 +51,36 @@ setGlobalContext({
...
-      deprecate(message, false, {
-        id: 'autotracking.mutation-after-consumption',
-        until: '4.0.0',
-        for: 'ember-source',
-        since: {
-          enabled: '3.21.0',
-        },
-      });
-    },
+      let override = VM_ASSERTION_OVERRIDES.find((o) => o.id === id);
```
rwjblue, line 59: Can't use `.find` on IE11 (this is failing browser stack)

rwjblue: I think we should probably change the overrides to be a map / pojo where the key is the `id`.  What do you think?

packages/@ember/-internals/glimmer/lib/environment.ts
```diff
@@ -50,29 +51,36 @@ setGlobalContext({
...
+
+      assert(override?.message ?? msg, test);
+    }
+  },
+
+  deprecate(msg: string, test: unknown, options: { id: string }) {
+    if (DEBUG) {
+      let { id } = options;
+
+      let override = VM_DEPRECATION_OVERRIDES.find((o) => o.id === id);
```
rwjblue, line 69: Same here RE: `.find`

rwjblue (changes requested):

jamescdavis:
Fixes #19360

rwjblue (approved):

PR #19365, opened 2021-01-30
[BUGFIX release] Remove non-existing re-export from helper-addon blueprint

simonihmig:
The blueprint for an addon's `app/helpers/foo.js` contains a re-export of a named export of `addon/helpers/foo.js` which does not exist, as it [only has a default export](https://github.com/emberjs/ember.js/blob/master/blueprints/helper/files/__root__/__collection__/__name__.js).

FWIW, Embroider is constantly complaining about this (rightly).


rwjblue:
Thank you @simonihmig!

PR #19370, opened 2021-02-02
[BUGFIX beta] Update glimmer-vm to prevent errors for older inline precompilation

### LINKED ISSUE #19331: [Bug] Helper crashes renderer with current beta (regression)
### 🐞 Describe the Bug

CI suddenly fails for https://github.com/kanongil/ember-stable-hash on beta and canary.

### 🔬 Minimal Reproduction

Failing [helper](https://github.com/kanongil/ember-stable-hash/blob/master/addon/helpers/stable-hash.js):
```js
import { setProperties } from '@ember/object';
import Helper from '@ember/component/helper';

export default class StableHash extends Helper {
  constructor(...args) {
    super(...args);

    this.obj = {};
  }

  compute(positional, named) {
    setProperties(this.obj, named);
    return this.obj;
  }
}
```

Fails the tests in [stable-hash-test.js](https://github.com/kanongil/ember-stable-hash/blob/master/tests/integration/helpers/stable-hash-test.js). Eg. this:
```js
test('returns a hash with the right key-value', async function(assert) {
  await render(hbs`{{#with (stable-hash name=\"Sergio\") as |person|}}{{person.name}}{{/with}}`);

  assert.dom().hasText('Sergio');
});
```

### 😕 Actual Behavior

```
not ok 1 Chrome 62.0 - [138 ms] - helper:stable-hash: returns a hash with the right key-value
    ---
        actual: >
            [object Object]
        stack: >
            TypeError: _a.call is not a function
                at meta (http://localhost:7357/assets/vendor.js:43790:86)
                at compilable (http://localhost:7357/assets/vendor.js:45534:51)
                at TemplateImpl.asLayout (http://localhost:7357/assets/vendor.js:46242:28)
                at new OutletComponentDefinition (http://localhost:7357/assets/vendor.js:7444:150)
                at http://localhost:7357/assets/vendor.js:9062:13
                at http://localhost:7357/assets/vendor.js:47061:37
                at track (http://localhost:7357/assets/vendor.js:57034:7)
                at valueForRef (http://localhost:7357/assets/vendor.js:47060:44)
                at Assert.evaluate (http://localhost:7357/assets/vendor.js:50032:48)
                at UpdatingVM._execute (http://localhost:7357/assets/vendor.js:52711:16)
        message: >
            _a.call is not a function
        negative: >
            false
        browser log: |
            ERROR:

            Error occurred:

            - While rendering:
              -top-level
                application
```
### 🤔 Expected Behavior

Pass the test, like the current release.

### 🌍 Environment

- Ember: current beta & canary
- Node.js/npm: v10.23.1 / 6.14.10
- OS: trusty

rwjblue:
This updates glimmer-vm to avoid emitting the `scope` property on precompiled template blocks when not using strict mode (when using strict mode `scope` will either be `null` or a function).

This is specifically to work around a bug in older versions of babel-plugin-htmlbars-inline-precompile (versions from v2.0.0 through v4.0.1) where they would (incorrectly) attempt to build an `ObjectExpression` when a `null` value is found in the template precompilation result (D'oh!).

Note: those versions of babel-plugin-htmlbars-inline-precompile should still be migrated away from (they are quite old, and will definitely not work for newer features like strict mode).

Fixes https://github.com/emberjs/ember.js/issues/19331


rwjblue:
Will need to be rebased on https://github.com/emberjs/ember.js/pull/19371 (will fail until that lands).

rwjblue:
Rebased to incorporate changes from #19371

PR #19371, opened 2021-02-02
Deprecate implicit `this` property lookup fallback

rwjblue:
Lands the deprecation proposed in [emberjs/rfcs#308](https://github.com/emberjs/rfcs/blob/master/text/0308-deprecate-property-lookup-fallback.md):

The [no-implicit-this](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-implicit-this.md) template linting rule has been enabled for Ember applications by default since around Ember 3.16 (included in the `octane` preset).

This PR will take that a step further and issue a deprecation for templates that are still using `{{foo}}` to mean `{{this.foo}}`.

Note: It does provide for an easy way to temporarily disable the deprecation messages (e.g. to reduce console output to hone in on other issues). This is intended for sporadic usage while debugging **not** as a way to mitigate the deprecation (the fallback behavior will be removed in 4.0.0).


packages/@ember/-internals/glimmer/lib/environment.ts
```diff
@@ -100,6 +106,15 @@ const VM_DEPRECATION_OVERRIDES: (DeprecationOptions & { message?: string })[] =
       enabled: '3.21.0',
     },
   },
+  {
+    id: 'this-property-fallback',
+    disabled: ENV._DISABLE_PROPERTY_FALLBACK_DEPRECATION,
+    until: '4.0.0',
+    for: 'ember-source',
```
pzuraq, line 113: The `url` for this deprecation will be https://deprecations.emberjs.com/v3.x#toc_this-property-fallback, based on https://github.com/ember-learn/deprecation-app/pull/773

pzuraq (commented):
Overall looks good, didn't realize that class bindings could have `this` in them. The only thing is to add the URL for the deprecation guide.

packages/@ember/-internals/glimmer/lib/environment.ts
rwjblue: Thanks! updated

PR #19374, opened 2021-02-02
Add hasBlock and hasBlockParams deprecation messages

patocallaghan:
Adds a deprecation for `hasBlock` and `hasBlockParams`. [RFC](https://github.com/emberjs/rfcs/blob/master/text/0689-deprecate-has-block.md)

**key:** `has-block-and-has-block-params` (taken from https://github.com/ember-learn/deprecation-app/pull/778)

```
`hasBlock` is deprecated. Use `has-block` instead.
```

```
`hasBlockParams` is deprecated. Use `has-block-params` instead.
```

/cc @pzuraq


pzuraq (approved):

pzuraq:
Thanks for doing this, really appreciate it!

PR #19324, opened 2020-12-25
Use ES6 property syntax in location example

sandstrom:
I hope this is alright.

PR #19376, opened 2021-02-03
ES6 property syntax in Location example

sandstrom:
Update API Location example to use ES6 syntax for the `implementation` property.

Continuation of https://github.com/emberjs/ember.js/pull/19324#issuecomment-767801005


rwjblue (approved):
Thank you!

PR #19379, opened 2021-02-04
[CLEANUP] Refactor DataAdapter to not use observers or array observers

pzuraq:
The DataAdapter class is exposed specifically for the Ember Inspector to
support Ember Data. Currently, the adapter relies on both array
observers and standard observers. This PR refactors the adapter to use
autotracking-based methods for tracking changes instead, with minimal
API churn.

The main change is to introduce a regular poll that checks for changes
after each backburner runloop. There are two types of polling done:

1. `TypeWatcher`, which checks to see if a record array has changed at
   all (e.g. contents have changed)
2. `RecordsWatcher`, which watches the array as a whole _and_ the
   individual records within it. Whenever any changes occur, it
   reiterates the records array, and looks for new records, updated
   records, and removed records, and notifies of any changes.

This removes the need for both array observers _and_ standard observers.
Standard observers were previously used to detect changes on the model
instances, but now we autotrack _serializing_ them, and that
autotracking is used to detect if they've been updated.

\## Breaking Changes

- The `recordsRemoved` callback from `watchRecords` has changed. It now
  receives an array of removed records, _not_ an index and count. On the
  consuming side in Ember Inspector, we'll have to update the code to
  remove the items passed via id explicitly.
- The `observeRecord` method has been removed. Because serializing the
  object _is_ observing it (via autotracking) there is no need to
  manually observe it. Users who are still implementing this method
  (e.g. Ember Data) can still keep it, and it shouldn't cause issues, it
  just will no longer be called.


packages/@ember/-internals/extension-support/lib/data_adapter.js
```diff
@@ -1,8 +1,119 @@
...
-import { scheduleOnce } from '@ember/runloop';
-import { get, objectAt, addArrayObserver, removeArrayObserver } from '@ember/-internals/metal';
+import { backburner } from '@ember/runloop';
+import { get } from '@ember/-internals/metal';
 import { dasherize } from '@ember/string';
 import { Namespace, Object as EmberObject, A as emberA } from '@ember/-internals/runtime';
+import { consumeTag, createCache, getValue, tagFor, untrack } from '@glimmer/validator';
+
+function iterate(arr, fn) {
+  if (Symbol.iterator in arr) {
```
rwjblue, line 9: Can't use `Symbol` here  (it isn't defined on IE11)

packages/@ember/-internals/extension-support/lib/data_adapter.js
```diff
@@ -1,8 +1,119 @@
...
+
+class RecordsWatcher {
+  recordCaches = new Map();
+
+  constructor(records, recordsAdded, recordsUpdated, recordsRemoved, wrapRecord, release) {
+    let { recordCaches } = this;
+
+    this.release = release;
+
+    let added, updated, removed;
```
rwjblue, line 26: Why are these declared outside of the cache? It doesn't seem like they are used anywhere else, wouldn't it make more sense to declare them inside?

packages/@ember/-internals/extension-support/lib/data_adapter.js
```diff
@@ -92,6 +205,24 @@ export default EmberObject.extend({
...
   acceptsModelName: true,
 
+  /**
+     Map from records arrays to RecordsWatcher instances
+
+     @private
+     @property recordsWatchers
+     @since 3.26.0
+   */
+  recordsWatchers: null,
```
rwjblue, line 215: We don't need to initialize this to `null` on the prototype just to clobber that in the constructor

```suggestion
```

packages/@ember/-internals/extension-support/lib/data_adapter.js
```diff
@@ -92,6 +205,24 @@ export default EmberObject.extend({
...
+
+  /**
+    Map from records arrays to TypeWatcher instances
+
+    @private
+    @property typeWatchers
+    @since 3.26.0
+   */
+  typeWatchers: null,
```
rwjblue, line 224:
```suggestion
```

packages/@ember/-internals/extension-support/lib/data_adapter.js
```diff
@@ -100,7 +231,7 @@ export default EmberObject.extend({
     @property releaseMethods
     @since 1.3.0
   */
-  releaseMethods: emberA(),
+  releaseMethods: null,
```
rwjblue, line 234:
```suggestion
```

packages/@ember/-internals/extension-support/lib/data_adapter.js
```diff
@@ -53,6 +163,9 @@ export default EmberObject.extend({
   init() {
     this._super(...arguments);
     this.releaseMethods = emberA();
+    this.recordsWatchers = new Map();
+    this.typeWatchers = new Map();
+    this.flushWatchers = this.flushWatchers.bind(this);
```
rwjblue, line 168: Why bind like this?

packages/@ember/-internals/extension-support/lib/data_adapter.js
```diff
@@ -284,28 +409,33 @@ export default EmberObject.extend({
...
-        // (which is all we care about as far as model types are concerned).
-        if (removedCount > 0 || addedCount > 0) {
-          scheduleOnce('actions', this, onChange);
+    let { typeWatchers } = this;
+
+    let typeWatcher = typeWatchers.get(records);
+
+    if (!typeWatcher) {
+      if (typeWatchers.size === 0) {
+        backburner.on('end', this.flushWatchers);
```
rwjblue, line 422: We should only add `on('end', this.flushWatchers)` once (seems like we might add it twice).

rwjblue (commented):
I left a few inline comments, but we also need to change things around to avoid adding more than one `backburner.on('end', ...)` callback as well as ensuring that we clean it up upon destruction. As it stands now, we leak in a few ways (registering the same callback many times, never removing but adding another callback per-container, etc).

packages/@ember/-internals/extension-support/lib/data_adapter.js
pzuraq: I actually thought so too! But they are actually captured by the _inner_ caches, the ones created for each item. So, they have to be a shared reference.

packages/@ember/-internals/extension-support/lib/data_adapter.js
pzuraq: Hmm, this code should only run in Chrome/Firefox shouldn't it, since it's only for inspector? I can guard it regardless, just in case.

packages/@ember/-internals/extension-support/lib/data_adapter.js
pzuraq: Well, you can't do `this.flushWatchers = () => this.flushWatchers();`, that'll cause infinite recursion. Open to suggestions

packages/@ember/-internals/extension-support/lib/data_adapter.js
pzuraq: ah, you're correct. I'll extract this out to a shared function

packages/@ember/-internals/extension-support/lib/data_adapter.js
rwjblue: Inspector supports (or at least did) running outside of an extension from a bookmarklet. You could inspect in other browsers this way (I've personally used it for Safari at least).

packages/@ember/-internals/extension-support/lib/data_adapter.js
rwjblue: Hmm, I don't see how 🤔

packages/@ember/-internals/extension-support/lib/data_adapter.js
rwjblue: When I commented I hadn't noticed the callback usage. I think I've usually seen `this._boundFlushWatchers = () => this.flushWatchers();` (or use `.bind`). 

I didn't don't care about this though, so if you don't think it should change feel free to resolve.

packages/@ember/-internals/extension-support/lib/data_adapter.js
pzuraq: Specifically here: https://github.com/emberjs/ember.js/pull/19379/files#diff-f98ed9e6e2838cfeb4d67e382b98069cc174b21c1fec5b390faa83cdc9f7a43aR49

The cache for the item captures the reference to `updated`, and then it gets cached. Later on, it pushes into the same `updated` array the next time we load the cache. If we create a new `updated` array, then it's holding a reference to an older `updated`.

I can try to think about how to make this more intuitive, agree it's not super clear.

pzuraq: Here's an alternative implementation, do you think this is clearer?

```js
class RecordsWatcher {
  recordCaches = new Map();

  added = [];
  updated = [];
  removed = [];

  getCacheForItem(record) {
    let recordCache = this.recordCaches.get(record);

    if (!recordCache) {
      let hasBeenAdded = false;

      recordCache = createCache(() => {
        if (!hasBeenAdded) {
          this.added.push(this.wrapRecord(record));
          hasBeenAdded = true;
        } else {
          this.updated.push(this.wrapRecord(record));
        }
      });

      this.recordCaches.set(record, recordCache);
    }
  }

  constructor(records, recordsAdded, recordsUpdated, recordsRemoved, wrapRecord, release) {
    this.release = release;
    this.wrapRecord = wrapRecord;

    this.recordArrayCache = createCache(() => {
      let seen = new Set();

      // Track `[]` for legacy support
      consumeTag(tagFor(records, '[]'));

      iterate(records, (record) => {
        getValue(this.getCacheForItem(record));
        seen.add(record);
      });

      // Untrack this operation because these records are being removed, they
      // should not be polled again in the future
      untrack(() => {
        this.recordCaches.forEach((cache, record) => {
          if (!seen.has(record)) {
            this.removed.push(wrapRecord(record));
            this.recordCaches.delete(record);
          }
        });
      });

      if (this.added.length > 0) {
        recordsAdded(this.added);
        this.added = [];
      }

      if (this.updated.length > 0) {
        recordsUpdated(this.updated);
        this.updated = [];
      }

      if (this.removed.length > 0) {
        recordsRemoved(this.removed);
        this.removed = [];
      }
    });
  }

  revalidate() {
    getValue(this.recordArrayCache);
  }
}
```

tests/docs/expected.js
```diff
@@ -115,9 +115,9 @@ module.exports = {
     'cacheFor',
     'camelize',
     'canCatalogEntriesByType',
-    'canInvoke',
     'cancel',
```
pzuraq, line 118: This file was partially sorted, but not fully. I decided to sort it to make it consistent.

packages/@ember/-internals/extension-support/lib/data_adapter.js
```diff
@@ -1,8 +1,122 @@
...
-import { scheduleOnce } from '@ember/runloop';
-import { get, objectAt, addArrayObserver, removeArrayObserver } from '@ember/-internals/metal';
+import { backburner } from '@ember/runloop';
+import { get } from '@ember/-internals/metal';
 import { dasherize } from '@ember/string';
 import { Namespace, Object as EmberObject, A as emberA } from '@ember/-internals/runtime';
+import { consumeTag, createCache, getValue, tagFor, untrack } from '@glimmer/validator';
+
+function iterate(arr, fn) {
+  if (typeof Symbol !== 'undefined' && Symbol.iterator in arr) {
```
rwjblue, line 9: FWIW, I'm pretty sure we have a util for something like `HAS_NATIVE_SYMBOL`

rwjblue (changes requested):
Still doesn't seem to `backburner.off('end', this.flushWatchers)` upon destruction (which means it leaks across multiple owners / tests).

rwjblue (approved):

snewcomer:
https://github.com/emberjs/data/pull/7435/files

We have a minor test issue as a result of this refactor.  Any ideas on how we can wait for the flush to happen?  I see you are taking advantage of `runLoopSettled` in the tests here...

pzuraq:
Correct, you should be able to use the `settled()` helper from `@ember/test-helpers`, which is essentially the same. I believe I tested this out locally before we merged this change

snewcomer:
Oh jeez.  I was putting the `settled()` after the async action but before `set`.  Now it works with it in the right place.  Thanks!

Issue #19270, opened 2020-11-14
[FEATURE modernized-built-in-components] Tracking Issue


chancancode:
The remaining tasks, in no particular order (most of these can be parallelized)

* [x] Merge [RFC #671](https://github.com/emberjs/rfcs/pull/671)
* [x] Merge https://github.com/emberjs/ember.js/pull/19218
* [x] Enable the deprecations introduced in https://github.com/emberjs/ember.js/pull/19218
* [x] Address https://github.com/emberjs/ember.js/pull/19218#discussion_r5201385
* [x] Identify, implement and deprecate other classic component features
* [x] Implement the "deopt"
* [x] Submit the "part 2" deprecation RFC
* [x] Migrate `<Textarea>` to be based on the new implementation
* [x] Refactor `<LinkTo>`
* [x] Enable the feature flag
* [ ] Deprecation guide entries
* [ ] Profit

Some guidance on each of these...

## Merge https://github.com/emberjs/ember.js/pull/19218 (🔒 @rwjblue)

Should be good to go, as far as I can tell.

## Enable the deprecations introduced in https://github.com/emberjs/ember.js/pull/19218

The PR introduced a few new deprecations, but they are disabled even when the feature flag is enabled, by hard coding the deprecation condition to `true`. In each case though, I left the actual condition in a JS comment.

The reason that these deprecations were not enabled (behind the feature flag) as part of the original PR was that those features are used in the internal test suite, so enabling them would require a modest amount of work fixing up the existing tests.

The deprecations are:

* [ ] Passing the `@id` argument to <Input> is deprecated. Instead, please pass the attribute directly, i.e. `<Input id={{...}} />` instead of `<Input @id={{...}} />` or `{{input id=...}}`. (`id` is just an example here, this applies to other attributes as well.)
* [ ] Passing the `@change` argument to <Input> is deprecated. This would have overwritten the internal `change` method on the <Input> component and prevented it from functioning properly. Instead, please use the {{on}} modifier, i.e. `<Input {{on "change" ...}} />` instead of `<Input @change={{...}} />` or `{{input change=...}}`. (`change` is just an example here, this applies to other event methods as well.)
* [ ] Passing the `@click` argument to <Input> is deprecated. Instead, please use the {{on}} modifier, i.e. `<Input {{on "click" ...}} />` instead of `<Input @click={{...}} />` or `{{input click=...}}`. (`click` is just an example here, this applies to other event as well.)
[...]

PR #19380, opened 2021-02-04
[FEATURE modernized-built-in-components] Implement input attrs deprecations

eramod:
Tracking issue: https://github.com/emberjs/ember.js/issues/19270

More specifically, this PR enables the <Input> attribute deprecations introduced in https://github.com/emberjs/ember.js/pull/19218, fixes broken tests, and updates all test {{input}} invocations to use <Input> instead.

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/components/input-angle-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/components/input-angle-test.js
@@ -1,4 +1,11 @@
-import { RenderingTestCase, moduleFor, runDestroy, runTask } from 'internal-test-helpers';
+import {
+  RenderingTestCase,
+  maybeExpectDeprecation,
+  moduleFor,
+  runDestroy,
+  runTask,
+} from 'internal-test-helpers';
+import { EMBER_MODERNIZED_BUILT_IN_COMPONENTS } from '@ember/canary-features';
 
 import { action } from '@ember/object';
 import { assign } from '@ember/polyfills';
@@ -305,28 +312,32 @@ moduleFor(
       // this.assertAttr('tabindex', '30'); //NOTE: failing in IE (TEST_SUITE=sauce)
     }
 
-    ['@test dynamic attributes (named argument)']() {
-      this.render(
-        `
-      <Input @type="text" @value={{this.value}}
-        @disabled={{this.disabled}}
-        @placeholder={{this.placeholder}}
-        @name={{this.name}}
-        @maxlength={{this.maxlength}}
-        @minlength={{this.minlength}}
-        @size={{this.size}}
-        @tabindex={{this.tabindex}}
-      />`,
-        {
-          value: 'Original value',
-          disabled: false,
-          placeholder: 'Original placeholder',
-          name: 'original-name',
-          maxlength: 10,
-          minlength: 5,
-          size: 20,
-          tabindex: 30,
-        }
+    ['@test [DEPRECATED] dynamic attributes (named argument)']() {
+      maybeExpectDeprecation(
+        EMBER_MODERNIZED_BUILT_IN_COMPONENTS,
+        () => {
+          this.render(
+            `<Input @type="text" @value={{this.value}}
+              @disabled={{this.disabled}}
+              @placeholder={{this.placeholder}}
+              @name={{this.name}}
+              @maxlength={{this.maxlength}}
+              @minlength={{this.minlength}}
+              @size={{this.size}}
+              @tabindex={{this.tabindex}}/>`,
+            {
+              value: 'Original value',
+              disabled: false,
+              placeholder: 'Original placeholder',
+              name: 'original-name',
+              maxlength: 10,
+              minlength: 5,
+              size: 20,
+              tabindex: 30,
+            }
+          );
+        },
+        /Passing the `@(disabled|placeholder|name|maxlength|minlength|size|tabindex)` argument to <Input> is deprecated\./
       );
 
       this.assertNotDisabled();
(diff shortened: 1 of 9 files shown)


packages/@ember/-internals/glimmer/tests/integration/components/input-angle-test.js
```diff
@@ -305,28 +312,33 @@ moduleFor(
...
+              placeholder: 'Original placeholder',
+              name: 'original-name',
+              maxlength: 10,
+              minlength: 5,
+              size: 20,
+              tabindex: 30,
+            }
+          );
+        },
+        /Passing the `@.+` argument to <Input> is deprecated\./
```
chancancode, line 341:
```suggestion
        /Passing the `@(disabled|placeholder|name|maxlength|minlength|size|tabindex)` argument to <Input> is deprecated\./
```

Let's scope this regex more narrowly. I realized we would want to make sure the deprecation is _not_ triggered for `@type` and `@value`.

chancancode: (same thing for the other `@.+` elsewhere)

packages/@ember/-internals/glimmer/tests/integration/components/input-angle-test.js
```diff
@@ -1181,16 +1240,21 @@ moduleFor(
...
-      runTask(() => {
-        set(this.context, 'disabled', true);
-        set(this.context, 'value', 'Updated value');
-        set(this.context, 'placeholder', 'Updated placeholder');
-        set(this.context, 'name', 'updated-name');
-        set(this.context, 'maxlength', 11);
-        set(this.context, 'size', 21);
-        set(this.context, 'tabindex', 31);
-      });
+      maybeExpectDeprecation(
```
chancancode, line 1243:
```suggestion

      maybeExpectDeprecation(
```

chancancode: nit: add back the blank line in between for grouping

packages/@ember/-internals/glimmer/tests/integration/components/input-angle-test.js
```diff
@@ -1244,7 +1314,17 @@ moduleFor(
...
+            EMBER_MODERNIZED_BUILT_IN_COMPONENTS,
+            () => {
+              this.render(`<Input ${attrs.replace('%x', value)} />`);
+            },
+            /Passing the `@.+` argument to <Input> is deprecated\./
+          );
+        } else {
+          this.render(`<Input ${attrs.replace('%x', value)} />`);
+        }
       }
```
chancancode, line 1329: On second thought, we can do something slightly better for this case:

```ts
class InputAttributesTest extends InputRenderingTest {
  renderInput(value = 25) {
    this.render(`<Input ${attrs.replace('%x', value)} />`);
  }

  ['@test value over default max but below set max is kept']() {
    // ...
  }

  // ...
}

// These are the permutations of the set:
// ['type="range"', 'min="-5" max="50"', value="%x"']

[
  '@type="range" min="-5" max="50" @value="%x"',
  '@type="range" @value="%x" min="-5" max="50"',
  'min="-5" max="50" @type="range" @value="%x"',
  'min="-5" max="50" @value="%x" @type="range"',
  '@value="%x" min="-5" max="50" @type="range"',
  '@value="%x" @type="range" min="-5" max="50"',
].forEach((attrs) => {
  moduleFor(
    `[GH#15675] Components test: <Input ${attrs} />`,
    class extends InputAttributesTest {}
  );
});

[
  '@type="range" @min="-5" @max="50" @value="%x"',
  '@type="range" @value="%x" @min="-5" @max="50"',
  '@min="-5" @max="50" @type="range" @value="%x"',
  '@min="-5" @max="50" @value="%x" @type="range"',
  '@value="%x" @min="-5" @max="50" @type="range"',
  '@value="%x" @type="range" @min="-5" @max="50"',
].forEach((attrs) => {
  moduleFor(
    `[GH#15675] [DEPRECATED] Components test: <Input ${attrs} />`,
    class extends InputAttributesTest {
      renderInput(value = 25) {
        maybeExpectDeprecation(
          EMBER_MODERNIZED_BUILT_IN_COMPONENTS,
          () => super.renderInput(value),
          /Passing the `@(min|max)` argument to <Input> is deprecated\./
        );
      }
    }
  );
});
```

Basically we will define the test class once. Just having the class defined doesn't do anything by itself (it doesn't add the tests to the test suite), it's only when we pass the class to the `moduleFor()` function that actually loops through the methods and find the ones starting with `@test` and pass then to `QUnit.test(...)`.

This allows us to split the existing array into two halves – the deprecated ones and the not deprecated ones. It allows us to give the modules different names (so we can mark it as `[DEPRECATED]`), and avoids the extra duplication in `renderValue` as a bonus.

packages/@ember/-internals/glimmer/tests/integration/components/input-curly-test.js
```diff
@@ -266,18 +285,24 @@ moduleFor(
...
+          {{input type="text"
+            disabled=true
+            value="Original value"
+            placeholder="Original placeholder"
+            name="original-name"
+            maxlength=10
+            minlength=5
+            size=20
+            tabindex=30
+          }}`);
```
chancancode, line 302: nit: this indentation style looks pretty good to me, let's apply this to the other test cases you touched since the auto-formatter made some of them look pretty strange.

packages/ember-testing/tests/helpers_test.js
```diff
@@ -846,11 +841,9 @@ if (!jQueryDisabled) {
 
         this.addTemplate(
           'index',
-          `
-        <input type="text" id="first"
```
chancancode, line 850: This one was using the `<input>` HTML element before, so we shouldn't change it. I didn't actually check if it was _intentional_ or _important_ to the test that it uses `<input>` instead of `<Input />`, but we shouldn't _need_ to change that as part of this PR so better to leave it as-is. If there is a good reason to update those tests separately from landing this deprecation, we could always do that as a separate PR to make it clear.

packages/ember-testing/tests/helpers_test.js
```diff
@@ -880,8 +873,8 @@ if (!jQueryDisabled) {
         this.addTemplate(
           'index',
           `
-        <input type="text" id="first" class="in-test">
-        <input type="text" id="second" class="in-test">
```
chancancode, line 884: same for these two – leave as-is

packages/internal-test-helpers/index.js
```diff
@@ -7,6 +7,7 @@ export { default as moduleFor, setupTestClass } from './lib/module-for';
 export { default as strip } from './lib/strip';
 export { default as applyMixins } from './lib/apply-mixins';
 export { default as getTextOf } from './lib/get-text-of';
+export { default as maybeExpectDeprecation } from './lib/maybe-expect-deprecation';
```
chancancode, line 10: nice!

packages/internal-test-helpers/lib/maybe-expect-deprecation.ts
```diff
@@ -0,0 +1,16 @@
+import { ExpectDeprecationFunc } from './ember-dev/deprecation';
```
chancancode, line 1: Have you tried just doing `import './ember-dev/deprecation';`? I _think_ that should cause the compiler to "see" the `declare global { ... }` block in that file, which should make `expectDeprecation` available here.

packages/internal-test-helpers/lib/maybe-expect-deprecation.ts
```diff
@@ -0,0 +1,16 @@
...
+import { Message } from './ember-dev/utils';
+
+declare let expectDeprecation: ExpectDeprecationFunc;
+
+export default function maybeExpectDeprecation(
+  featureFlag: boolean | null,
+  callback: () => void,
+  message: Message
+): void | undefined | Message {
```
chancancode, line 10:
```suggestion
): void {
```

This should be returning `void` only, since none of the branches actually returned anything.

chancancode:
Thanks for working on this! Left a few comments, but otherwise looks great to me!

PR #19381, opened 2021-02-05
[DEPRECATION] Add Array Observers deprecation 

pzuraq:
Adds the array observers deprecation specified in RFC 692: https://github.com/emberjs/rfcs/blob/master/text/0692-deprecate-array-observers.md

Depends on #19379


rwjblue (approved):
Generally LGTM, needs a rebase now that #19379 is landed.

PR #19384, opened 2021-02-07
Use qunit-dom in helper and component test blueprints

bertdeblock:


NullVoxPopuli:
how do we know if these tests pass right away? I know there is a smidgen difference between hasText and containsText, where hasText requires exact matches (trimmed, obvs) -- idk if that matters for this tho

NullVoxPopuli (approved):

bertdeblock:
@NullVoxPopuli Thanks for checking / approving. AFAICT, `hasText` is the right qunit-dom equivalent for the current `assert.equal` approach, which checks for an exact match as well (trimming unnecessary whitespace included). Regarding the fact if they pass or not, the correct output is tested, but I'm not sure if these files actually run as well during the test suite.

rwjblue (approved):

rwjblue:
Thank you!

PR #19385, opened 2021-02-08
Perform early caching on singleton object instances

Ravenstine:
This PR is intended to address this issue:
https://github.com/emberjs/ember.js/issues/17791

To summarize, injecting the `service:router` into `router:main` causes an infinite recursion error because `service:router` attempts to lookup `router:main` before `router:main` has had a chance to be cached as a singleton.  Caching of singletons doesn't happen until after initialization has completed.

Screenshot 2021-02-07 at 20 44 33

Although this seems to be a peculiar use of the router service, this can be a problem when accessing other services that use `service:router` inside of the router before it has fully initialized, as would be the case if one was using [ember-router-scroll](https://github.com/DockYard/ember-router-scroll) and tried accessing `service:router-scroll`.

I'm not married at all to the approach taken here, and since this is my first contribution I'd be glad to take guidance.  I landed on this approach because it seems like it would prevent most infinite loops that could occur by a circular reference between a singleton through a service.

Basically, I chose to allow CoreObject to preemptively add the object instance to the container cache before it has been initialized, which is why further calls to `owner.lookup('router:main')` at that point won't try to spawn another router instance even when `init()` on the router has yet to have been called.

My reasoning is that it seems preferable that a service access a singleton that hasn't been 100% setup yet as opposed to running into an infinite loop.  For instance, even if something like the router hasn't fully initialized, it can still be useful to have other parts of an app be able to add event listeners to it.  There are ways to get around this issue, of course, but I can't imagine a philosophy around why such hypothetical workarounds must be taken in this case.

One possible drawback might be that an object instance that encounters an error when `init()` is called would still remain in `container.cache`, which wouldn't happen without this change.  Of course it would be very easy to perform a "rollback", but I left that out because there didn't actually seem to be a requirement for keeping instances that error on `init()` out of the cache.  A try-catch block could definitely be added if that's what we need.

Like I said, guidance on this would be welcome if we are open to something akin to what this PR is currently doing.  I tried to come up with a way to signal that an instance should be cached that wasn't intrusive to existing APIs.

One thing I also chose to do was to promote `_didSetupRouter` to being a public property; this is because I chose to access it inside `service:router` so that it doesn't attempt to perform `router.setupRouter()` before `router.startRouting()` has a chance to preform on `router:main`(which prevents that method from performing).  After searching the codebase and the issues in this repo, I couldn't come up with a reason why it needed to be private.

### DIFF
--- a/packages/@ember/-internals/container/lib/container.ts
+++ b/packages/@ember/-internals/container/lib/container.ts
@@ -76,6 +76,7 @@ export default class Container {
   readonly registry: Registry & DebugRegistry;
   cache: { [key: string]: CacheMember };
   factoryManagerCache!: { [key: string]: FactoryManager<any, any> };
+  singletonInstanceNames: Set<string>;
   readonly validationCache!: { [key: string]: boolean };
   isDestroyed: boolean;
   isDestroying: boolean;
@@ -85,6 +86,7 @@ export default class Container {
     this.owner = options.owner || null;
     this.cache = dictionary(options.cache || null);
     this.factoryManagerCache = dictionary(options.factoryManagerCache || null);
+    this.singletonInstanceNames = new Set();
     this.isDestroyed = false;
     this.isDestroying = false;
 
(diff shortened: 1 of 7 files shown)


Ravenstine:
Closing this in favor of #19405 which has now been merged.

PR #19386, opened 2021-02-08
Deprecate passing input actions as strings

eramod:
Tracking issue: #19270

Fixes input tests for passing actions to components as strings deprecation

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/components/input-curly-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/components/input-curly-test.js
@@ -439,11 +439,11 @@ moduleFor(
             },
           },
         });
-      }, 'Passing actions to components as strings (like `{{input key-press="foo"}}`) is deprecated. Please use closure actions instead (`{{input key-press=(action "foo")}}`). (\'-top-level\' @ L1:C0) ');
+      }, /Passing actions to components as strings \(like `({{input key-press="foo"}}|<Input @key-press="foo" \/>)`\) is deprecated\./);
 
       expectDeprecation(() => {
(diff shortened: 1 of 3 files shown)


packages/@ember/-internals/glimmer/tests/integration/components/input-curly-test.js
```diff
@@ -439,11 +439,11 @@ moduleFor(
             },
           },
         });
-      }, 'Passing actions to components as strings (like `{{input key-press="foo"}}`) is deprecated. Please use closure actions instead (`{{input key-press=(action "foo")}}`). (\'-top-level\' @ L1:C0) ');
+      }, /(Passing actions to components as strings \(like `{{input key-press="foo"}}`\) is deprecated\.|Passing actions to components as strings \(like `<Input @key-press="foo" \/>`\) is deprecated\.)/);
```
chancancode, line 442: To make this more readable let's restrict the `|` to the part that's actually different:
```suggestion
      }, /Passing actions to components as strings \(like `({{input key-press="foo"}}|<Input @key-press="foo" \/>)`\) is deprecated\./);
```

chancancode: (same for the rest)

chancancode (approved):
LGTM other than refactoring the regex a bit

packages/@ember/-internals/glimmer/tests/integration/components/input-curly-test.js
eramod: That definitely looks better. I updated all of them to that format.

Issue #19364, opened 2021-01-30
[Bug] `LinkTo` with incomplete model failing in rendering tests


Turbo87:
### 🐞 Describe the Bug

We have a few components that use `LinkTo` components but are only used on subroutes with dynamic segments and have not explicitly added `@model` (e.g. only used within `/organizations/:org_id`). This works fine in production and up until Ember 3.24 this also worked fine in application and rendering tests.

With 3.24 Ember now automatically starts the routing layer whenever something on the `router` service is accessed, and now the `LinkTo` component is complaining that we did not pass the necessary models when any previous thing in the template for example accesses `router.currentRouteName`.

### 🔬 Minimal Reproduction
Describe steps to reproduce. If possible, please, share a link with a minimal reproduction.

```js
// router.js

Router.map(function() {
  this.route('parent', { path: 'parent/:parent_id' }, function () {
    this.route('child');
  });
});
```

```js
test('broken', async function(assert) {
  // access a property on the router
  this.owner.lookup('service:router').currentRouteName;

  // render a LinkTo with incomplete model
  await render(hbs`<LinkTo @route="parent.child">Link</LinkTo>`);
});
```

### 😕 Actual Behavior

> Assertion Failed: You attempted to generate a link for the "parent.child" route, but did not pass the models required for generating its dynamic segments. Cannot read property 'shouldSupercede' of undefined

### 🤔 Expected Behavior

> The `LinkTo` should render without an `href` like it did prior to Ember 3.24, or the `router.currentRouteName` access should not have the side-effect of automatically starting the routing layer, which is causing these issues.

### 🌍 Environment

- Ember: 3.24

### ➕ Additional Context

somewhat related: 
- https://github.com/emberjs/ember.js/pull/19080
- https://github.com/buschtoens/ember-link/pull/518
- https://github.com/buschtoens/ember-link/pull/507

/cc @rwjblue @xg-wang

xg-wang:
Sorry for the delay, I won't have much time until the weekend to look into the fix

rwjblue:
Chatted about this with @xg-wang earlier, I think what we should do is to change this guard:

https://github.com/emberjs/ember.js/blob/e25c0f258603fbfd57a47c14c00f69b93805ec46/packages/%40ember/-internals/routing/lib/services/routing.ts#L57-L60

To be something like:

```ts
    let router = this.router;
    // return early when the router microlib is not present, which is the case for {{link-to}} in integration tests
[...]

PR #19326, opened 2020-12-29
Lazily setup the router in non-application tests for <LinkTo> component

xg-wang:
## Summary
PR #19080 fixed the issue where accessing public router service throwing
error, however, the PR did not support LinkTo component which uses the
internal -routing service. This PR adds similar lazy setup to
-routing.router.

## Todo
- [x] edit https://github.com/ember-learn/super-rentals-tutorial/edit/master/src/markdown/tutorial/part-2/09-route-params.md to remove `this.owner.setupRouter();` for next release

PR #19387, opened 2021-02-09
[BUGFIX lts] LinkTo with incomplete model failing in rendering tests

xg-wang:
LinkTo needs route context to allow omitting model from current active
route. Without the guard, tests where LinkTo rendered in tests without
routing transition started will break. See https://github.com/emberjs/ember.js/issues/19364

Add failing test before the fix:
While generating link to route "dynamicWithChild.child": can't access
property "shouldSupercede", newHandlerInfo is undefined

### DIFF
--- a/packages/@ember/-internals/routing/lib/services/routing.ts
+++ b/packages/@ember/-internals/routing/lib/services/routing.ts
@@ -54,8 +54,10 @@ export default class RoutingService extends Service {
 
   generateURL(routeName: string, models: {}[], queryParams: {}) {
     let router = this.router;
-    // return early when the router microlib is not present, which is the case for {{link-to}} in integration tests
-    if (!router._routerMicrolib) {
+    // return early when the router microlib is not present, which is the case for <LinkTo/> in integration tests
+    // also return early when transition has not started, when rendering in tests without visit(),
+    // we cannot infer the route context which <LinkTo/> needs be aware of
+    if (!router._routerMicrolib || !router._initialTransitionStarted) {
       return;
     }
 
--- a/packages/@ember/-internals/routing/lib/services/routing.ts
+++ b/packages/@ember/-internals/routing/lib/services/routing.ts
@@ -54,10 +54,9 @@ export default class RoutingService extends Service {
 
   generateURL(routeName: string, models: {}[], queryParams: {}) {
     let router = this.router;
-    // return early when the router microlib is not present, which is the case for <LinkTo/> in integration tests
-    // also return early when transition has not started, when rendering in tests without visit(),
+    // Return early when transition has not started, when rendering in tests without visit(),
     // we cannot infer the route context which <LinkTo/> needs be aware of
-    if (!router._routerMicrolib || !router._initialTransitionStarted) {
+    if (!router._initialTransitionStarted) {
       return;
     }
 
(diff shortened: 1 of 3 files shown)


xg-wang:
This PR and https://github.com/emberjs/ember.js/pull/19080 should go into 3.24 (LTS)

packages/@ember/-internals/routing/lib/services/routing.ts
```diff
@@ -54,8 +54,10 @@ export default class RoutingService extends Service {
 
   generateURL(routeName: string, models: {}[], queryParams: {}) {
     let router = this.router;
-    // return early when the router microlib is not present, which is the case for {{link-to}} in integration tests
-    if (!router._routerMicrolib) {
+    // return early when the router microlib is not present, which is the case for <LinkTo/> in integration tests
```
rwjblue, line 57: This part of the comment should be removed, it is no longer correct

rwjblue (changes requested):
Looks good, can you prefix the commit message with `[BUGFIX lts]`? That is what let's us know to cherry pick things back into the current LTS (3.24).

packages/@ember/-internals/routing/lib/services/routing.ts
xg-wang: Good catch! Removed

xg-wang:
@rwjblue do I also need to cherry-pick https://github.com/emberjs/ember.js/pull/19080 and title it [BUGFIX lts]?

rwjblue (approved):

rwjblue:
> @rwjblue do I also need to cherry-pick #19326 and title it [BUGFIX lts] and [BUGFIX release]?

No, I'll take care of that.

rwjblue:
Done, 3.25.1 is released and 3.24.2 is mid-release.

rogeraraujo90:
Hello everyone, regarding this fix, The LinkTo component will not render `href` anymore during Integration tests?

My Integration test did an assert in the link in the Ember 3.20 that started fail in Ember 3.24.

lougreenwood:
@rogeraraujo90 Do you still have a problem here? Seems that I just noticed that `<LinkTo @route="bla">` where `bla` has a dynamic segment doesn't generate an href in integration tests.

rogeraraujo90:
@lougreenwood try add `this.owner.setupRouter();` in the beforeEach hook!

lougreenwood:
Thanks, did that solve it for you?

I'm also already using `setupRouter` (and as I understand it, it should be necessary in 3.28 anymore) - seems to be a bug where if there is no `@model` arg on the `linkTo` and the `linkTo` points to a route with dynamic segments, the `href` isn't added to the `linkTo`.

I've reproduced by both linking to routes without dynamic segments (`href` is added) and also adding arbitrary `@model="1"` to the template being tested, which also works. 

The links work fine in acceptance tests and production.

I'm planning on creating a minimal repto tomorrow and finding the release where this gets broken.

rogeraraujo90:
> Thanks, @rogeraraujo90, did that solve it for you?
> 
> I'm also already using `setupRouter` (and as I understand it, it should be necessary in 3.28 anymore) - seems to be a bug where if there is no `@model` arg on the `linkTo` and the `linkTo` points to a route with dynamic segments, the `href` isn't added to the `linkTo`.
> 
> I've reproduced by both linking to routes without dynamic segments (`href` is added) and also adding arbitrary `@model="1"` to the template being tested, which also works.
> 
> The links work fine in acceptance tests and production.
> 
> I'm planning on creating a minimal repto tomorrow and finding the release where this gets broken.

Good, go ahead!

I'm in Ember 3.24 yet, and using `setupRouter` the `href` is being generated correctly in the integration tests. To be more specific. my project is running `3.24.5`

PR #19388, opened 2021-02-09
[Deprecation] `renderTemplate` and `disconnectOutlet` methods

patocallaghan:
Deprecates `renderTemplate` ([RFC 418](https://emberjs.github.io/rfcs/0418-deprecate-route-render-methods.html)) and `disconnectOutlet` ([RFC 491](https://emberjs.github.io/rfcs/0491-deprecate-disconnect-outlet.html))

Taken from the [Deprecations list](https://www.notion.so/Deprecations-bbe769a13fb54bc8beafff175de4323a) 

### `renderTemplate`
* [x] Deprecation guide added to deprecations app
* [x] Deprecation message `The usage of 'renderTemplate' is deprecated. Please see the following deprecation guide to migrate.`
* [x] Key: `route-render-methods`
* [x] Call to `deprecate`
* [x] Fix tests

### `disconnectOutlet`
* [x] Deprecation guide added to deprecations app
* [x] Deprecation message: `"The usage of 'disconnectOutlet' is deprecated. Please see the following deprecation guide to migrate.`
* [x] Key: `disconnect-outlet`
* [x] Call to `deprecate`
* [x] Fix tests

/cc @locks

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/route.ts
+++ b/packages/@ember/-internals/routing/lib/system/route.ts
@@ -1705,6 +1705,15 @@ class Route extends EmberObject implements IRoute {
     @public
   */
   disconnectOutlet(options: string | { outlet: string; parentView?: string }) {
+    deprecate('The usage of `disconnectOutlet` is deprecated.', false, {
+      id: 'router-render-methods.disconnect-outlet',
+      until: '4.0.0',
+      url: 'TODO',
+      for: 'ember-source',
+      since: {
+        enabled: '3.26.0',
+      },
+    });
     let outletName;
     let parentView;
     if (options) {
--- a/packages/@ember/-internals/glimmer/tests/integration/application/debug-render-tree-test.ts
+++ b/packages/@ember/-internals/glimmer/tests/integration/application/debug-render-tree-test.ts
@@ -4,6 +4,7 @@ import {
   moduleFor,
   strip,
 } from 'internal-test-helpers';
+// import { ExpectDeprecationFunc } from '../../../../../../internal-test-helpers/lib/ember-dev/deprecation';
 
 import { ENV } from '@ember/-internals/environment';
 import { Component, setComponentManager } from '@ember/-internals/glimmer';
@@ -180,6 +181,10 @@ if (ENV._DEBUG_RENDER_TREE) {
       }
 
       async '@test named outlets'() {
+        // TODO - fix this type error
+        // @ts-ignore-next-line
+        expectDeprecation(() => {}, 'The usage of `disconnectOutlet` is deprecated.');
+
         this.addTemplate(
           'application',
           strip`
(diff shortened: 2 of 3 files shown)


chancancode:
There was another draft PR #19187, seems good for you to take over but you may find it helpful to reference that.

patocallaghan:
Thanks @chancancode! I'll do that 🙇‍♂️

packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -1705,6 +1705,15 @@ class Route extends EmberObject implements IRoute {
     @public
   */
   disconnectOutlet(options: string | { outlet: string; parentView?: string }) {
+    deprecate('The usage of `disconnectOutlet` is deprecated.', false, {
+      id: 'router-render-methods.disconnect-outlet',
+      until: '4.0.0',
+      url: 'TODO',
```
rwjblue, line 1711: Should be auto-deployed shortly, so you can grab the URL

packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -1705,6 +1705,15 @@ class Route extends EmberObject implements IRoute {
...
+      for: 'ember-source',
+      since: {
+        enabled: '3.26.0',
```
rwjblue, line 1714:
```suggestion
        enabled: '3.27.0',
```

rwjblue (commented):
Might also need to add the `render` deprecation here too?

packages/@ember/-internals/routing/lib/system/route.ts
patocallaghan: @rwjblue so isn't the current beta cycle 3.26 when we'd expect this to land?

pzuraq: So, beta has already branched away from master. This happens 6 weeks before the beta is released. During those 6 weeks, when we land features on master, they are not by _default_ included in beta - they are targeting the release _after_ beta (e.g. `canary`). 

We can, and sometimes do, patch them into the beta branch (e.g. for bugfixes). For features and deprecations, we approach it on a case-by-case basis, but we tend to avoid it unless there is a good reason and it's not a high risk change. In this case, I think it's reasonable to not backport, so we would be targeting master, which is currently `3.27.0`.

patocallaghan: I've decided to break the `disconnectOutlet` deprecation out into https://github.com/emberjs/ember.js/pull/19407. The `render` and `renderTemplate` deprecations are taking longer than I expected (lots of tests to fix).

rwjblue:
@patocallaghan - Where are we at with this one?

patocallaghan:
@rwjblue this can be closed now. Thanks!

PR #19390, opened 2021-02-09
Refactor the internal Ember loader to use the standard Ember CLI loader

pzuraq:
This PR refactors the internal Ember loader so that it uses the standard
Ember CLI loader instead for modules. This means that modules will be
defined in the main namespace, and importable from there, instead of in
a hidden namespace that only Ember can use.

Notes:

- Code is still loaded and built via `treeForVendor` and included as a
  vendor file. This needs to be the case for the time being for
  bootstrapping.
- Loader code is still included for Node support. If `define` and
  `require` are not already defined, then a backup shim is used instead.
- Modules are now exposed from Ember, but `ember-cli-babel` still
  transpiles them to global references. This unblocks us from being able
  to make all modules work normally, however.
- `require` shim module is no longer defined, we reference `define` and
  `require` as globals instead (which is more accurate). In the future
  we should update this to use Embroider's conventions.
- `__loader` is still exposed on the Ember object, referencing the same
  values as before.


lib/index.js
```diff
@@ -237,18 +237,11 @@ module.exports = {
...
       }),
 
       concatBundle(emberTestingFiles, {
         outputFile: 'ember-testing.js',
-        footer: `
-          var testing = require('ember-testing');
-          Ember.Test = testing.Test;
-          Ember.Test.Adapter = testing.Adapter;
-          Ember.Test.QUnitAdapter = testing.QUnitAdapter;
-          Ember.setupForTesting = testing.setupForTesting;
```
rwjblue, line 250: Where do these go?

packages/ember-template-compiler/lib/system/compile.ts
```diff
@@ -1,7 +1,6 @@
 /**
 @module ember
 */
-import require, { has } from 'require';
```
rwjblue, line 4: Hmm, why wouldn't we want this `require` to be imported? Without this import we are effectively using the global `require` (ala cjs)

packages/internal-test-helpers/lib/confirm-export.js
```diff
@@ -1,5 +1,3 @@
-import require from 'require';
```
rwjblue, line 1: if we are using the standard ember-cli loader then this shouldn't be removed

tests/index.html
```diff
@@ -10,10 +10,14 @@
       }
     </style>
     <script src="./polyfill.js"></script>
+    <script src="./loader.js"></script>
     <script src="./qunit/qunit.js"></script>
     <script src="/testem.js"></script>
 
     <script type="text/javascript">
+      window.define = define;
+      window.require = require;
```
rwjblue, line 19: What is even happening here?

tests/index.html
```diff
@@ -90,7 +94,7 @@
     </script>
 
     <script type="text/javascript">
-      var setTesting = Ember.__loader.require('@ember/debug')['setTesting'];
```
rwjblue, line 93: Why doesn't this work?

packages/loader/lib/index.js
```diff
@@ -56,57 +76,31 @@ mainContext = this; // eslint-disable-line no-undef
...
-    require = function (name) {
-      return internalRequire(name, null);
-    };
+    if (!callback) {
+      value.deps = [];
+      value.callback = deps;
+    } else {
+      value.deps = deps;
+      value.callback = callback;
+    }
```
rwjblue, line 93: Do we actually need a branch here? Doesn't our babel setup always emit 3 args?

tests/index.html
pzuraq: I started this branch building on the deprecation work, so this is a change that will need to happen to avoid the deprecation.

lib/index.js
pzuraq: Updated to make them lazy getters on the global Ember object, but we still need to `require('ember-testing');` because it reopens some classes eagerly and tests fail without it.

tests/index.html
pzuraq: Good catch, this was from an earlier issue I was trying to solve, it turned out to be something else but I left this here by mistake.

packages/loader/lib/index.js
pzuraq: Seems like we don't removing

packages/ember/index.js
```diff
@@ -782,30 +772,99 @@ runLoadHooks('Ember.Application', Application);
...
+    escapeExpression,
+  },
+};
+
+let EmberHTMLBars = {
+  template,
+};
+
+function defineEmberTemplateCompilerLazyLoad(key) {
+  Object.defineProperty(Ember, key, {
```
rwjblue, line 787: Should this set up individual getters for each property instead?

packages/loader/lib/index.js
```diff
@@ -1,12 +1,32 @@
...
+      : null;
+
+  if (globalObj === null) {
+    throw new Error('unable to locate global object');
+  }
+
+  if (typeof globalObj.define === 'function' && typeof globalObj.require === 'function') {
+    define = globalObj.define;
+    require = globalObj.require;
+
```
rwjblue, line 24: I think we might need to accommodate the no conflict APIs for `loader.js` here

https://github.com/ember-cli/loader.js#no-conflict

Since `loader` doesn't expose `loader.require`/`loader.define` or a way to access what the "new" aliases are we can't do this just yet.

tests/index.html
```diff
@@ -10,6 +10,7 @@
       }
     </style>
     <script src="./polyfill.js"></script>
+    <script src="./loader.js"></script>
```
rwjblue, line 13: Why do we need this? I thought we included the loader inside the assets themselves, if we do then I wouldn't expect to need to load loader again.

tests/index.html
```diff
@@ -180,11 +181,11 @@
       skipPackages = (skipPackages && skipPackages.split(',')) || [];
       var skipPackageRegexp   = new RegExp('^('+skipPackages.join('|')+')/');
 
-      for (var moduleName in Ember.__loader.registry) {
+      for (var moduleName in require._eak_seen) {
```
rwjblue, line 184: Can we use `require.entries` instead of `_eak_seen`?

packages/ember/index.js
```diff
@@ -782,30 +772,99 @@ runLoadHooks('Ember.Application', Application);
...
+defineEmberTestingLazyLoad('setupForTesting');
+
 runLoadHooks('Ember');
 
-export default Ember;
+Ember.__loader = {
+  require,
+  // eslint-disable-next-line no-undef
+  define,
+  registry: require._eak_seen,
```
rwjblue, line 864: Can we use this instead:

```suggestion
  registry: require.entries,
```

packages/loader/lib/index.js
```diff
@@ -56,57 +76,21 @@ mainContext = this; // eslint-disable-line no-undef
...
-    Ember.__loader = {
-      define: define,
-      require: require,
-      registry: registry,
-    };
-  } else {
-    define = Ember.__loader.define;
-    require = Ember.__loader.require;
-  }
+  require._eak_seen = registry;
```
rwjblue, line 95:
```suggestion
  require.entries = registry;
```

https://github.com/ember-cli/loader.js/blob/ccd773bae5153b070f861a44d556d79446db8ac5/lib/loader/loader.js#L331

tests/index.html
pzuraq: This makes it generally work more similarly to how these libraries would work with the real Ember CLI loader, because otherwise they would each have their own isolated `define` and `require`

rwjblue (approved):

PR #19391, opened 2021-02-10
[FEATURE modernized-built-in-components] Enable input event handlers

eramod:
Tracking issue: #19270

More specifically, this PR enables the input event handlers deprecations introduced in #19218 and fixes broken tests

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/components/input-angle-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/components/input-angle-test.js
@@ -151,7 +151,13 @@ class InputRenderingTest extends RenderingTestCase {
       <TestComponent ${argsFor('custom')} />
     `;
 
-    this.render(template, { actions });
+    maybeExpectDeprecation(
+      EMBER_MODERNIZED_BUILT_IN_COMPONENTS,
+      () => {
+        this.render(template, { actions });
+      },
+      /Passing the `@(touchStart|touchMove|touchEnd|touchCancel|keyDown|keyUp|keyPress|mouseDown|mouseUp|contextMenu|click|doubleClick|focusIn|focusOut|submit|input|change|dragStart|drag|dragEnter|dragLeave|dragOver|drop|dragEnd|mouseEnter|mouseLeave|mouseMove|focus-in|focus-out|key-press|key-up|key-down)` argument to <Input> is deprecated\./
+    );
 
     this.assert.ok(this.$('input').length === 2);
 
@@ -621,7 +627,7 @@ moduleFor(
             },
           },
         });
-      }, /Passing actions to components as strings \(like `({{input key-press="foo"}}|<Input @key-press="foo" \/>)`\) is deprecated\./);
+      }, /(Passing actions to components as strings \(like `({{input key-press="foo"}}|<Input @key-press="foo" \/>)`\) is deprecated\.|Passing the `@key-press` argument to <Input> is deprecated\.)/);
 
       expectDeprecation(() => {
         this.triggerEvent('keypress', { key: 'A' });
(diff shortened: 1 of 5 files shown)


packages/@ember/-internals/glimmer/tests/integration/components/input-angle-test.js
```diff
@@ -621,30 +627,34 @@ moduleFor(
             },
           },
         });
-      }, /Passing actions to components as strings \(like `({{input key-press="foo"}}|<Input @key-press="foo" \/>)`\) is deprecated\./);
+      }, /(Passing actions to components as strings \(like `({{input key-press="foo"}}|<Input @key-press="foo" \/>)`\) is deprecated\.|Passing the `@key-press` argument to <Input> is deprecated\.)/);
```
chancancode, line 630: It's probably better to do this in this case:

```js
maybeExpectDeprecation(
  EMBER_MODERNIZED_BUILT_IN_COMPONENTS,
  () => {
    expectDeprecation(
      () => this.render(...),
      /Passing actions to components as strings \(like `({{input key-press="foo"}}|<Input @key-press="foo" \/>)`\) is deprecated\./
    );
  },
  /Passing the `@key-press` argument to <Input> is deprecated\./
);
```

packages/@ember/-internals/glimmer/tests/integration/components/input-angle-test.js
```diff
@@ -777,7 +793,7 @@ moduleFor(
             },
           },
         });
-      }, /Passing actions to components as strings \(like `({{input key-down="foo"}}|<Input @key-down="foo" \/>)`\) is deprecated\./);
```
chancancode, line 780: Same for this, seems better to split it

packages/@ember/-internals/glimmer/tests/integration/components/input-angle-test.js
```diff
@@ -677,13 +687,19 @@ moduleFor(
     ['@test triggers `focus-in` when focused'](assert) {
```
chancancode, line 687: Add [DEPRECATED]

packages/@ember/-internals/glimmer/tests/integration/components/input-angle-test.js
```diff
@@ -787,20 +803,24 @@ moduleFor(
     ['@test sends an action with `<Input @key-down={{action "foo"}} />` when a key is pressed'](
```
chancancode, line 803: Add [DEPRECATED]

packages/@ember/-internals/glimmer/tests/integration/components/input-angle-test.js
```diff
@@ -787,20 +803,24 @@ moduleFor(
...
       assert
     ) {
-      assert.expect(2);
```
chancancode, line 790: If you are removing the `assert.expect(2);`, better to use this pattern:

```js
let trigged = 0;

// ...
actions: {
  foo() {
    trigged++;
    // ...
  }
}

// ...

this.triggerEvent(...);

assert.equal(triggered, 1, "They action was trigged exactly once");
```

Otherwise, if the action was never trigged, the test would not fail as written

packages/@ember/-internals/glimmer/tests/integration/components/input-angle-test.js
```diff
@@ -823,7 +843,7 @@ moduleFor(
             },
           },
         });
-      }, /Passing actions to components as strings \(like `({{input key-up="foo"}}|<Input @key-up="foo" \/>)`\) is deprecated\./);
+      }, /(Passing actions to components as strings \(like `({{input key-up="foo"}}|<Input @key-up="foo" \/>)`\) is deprecated\.|Passing the `@key-up` argument to <Input> is deprecated\.)/);
```
chancancode, line 846: Let's split it

packages/@ember/-internals/glimmer/tests/integration/components/input-angle-test.js
```diff
@@ -833,20 +853,25 @@ moduleFor(
     ['@test sends an action with `<Input @key-up={{action "foo"}} />` when a key is pressed'](
```
chancancode, line 853: [DEPRECATED]

packages/@ember/-internals/glimmer/tests/integration/components/input-angle-test.js
```diff
@@ -862,11 +887,17 @@ moduleFor(
     }
 
     ['@test triggers a method with `<Input @key-up={{this.didTrigger}} />`'](assert) {
```
chancancode, line 889: [DEPRECATED]

packages/@ember/-internals/glimmer/tests/integration/components/input-curly-test.js
```diff
@@ -447,30 +453,35 @@ moduleFor(
             },
           },
         });
-      }, /Passing actions to components as strings \(like `({{input key-press="foo"}}|<Input @key-press="foo" \/>)`\) is deprecated\./);
+      }, /(Passing actions to components as strings \(like `({{input key-press="foo"}}|<Input @key-press="foo" \/>)`\) is deprecated\.|Passing the `@key-press` argument to <Input> is deprecated\.)/);
```
chancancode, line 456: The same comments from the angle file applies to this file as well

chancancode (commented):
Seems like a couple of tests in in the <Textarea /> tests need to be fixed as well

packages/@ember/-internals/glimmer/tests/integration/components/input-angle-test.js
eramod: I tried this and it seems to result in getting the deprecation at the wrong time: 

Screen Shot 2021-02-10 at 11 22 20 AM

Full test for reference: 

```['@test [DEPRECATED] sends an action with `<Input @key-press="foo" />` is pressed'](assert) {
      assert.expect(4);
      maybeExpectDeprecation(
        EMBER_MODERNIZED_BUILT_IN_COMPONENTS,
        () => {
          expectDeprecation(
            () => {
              this.render(`<Input @value={{this.value}} @key-press='foo' />`, {
                value: 'initial',
                actions: {
                  foo(value, event) {
                    assert.ok(true, 'action was triggered');
                    if (jQueryDisabled) {
                      assert.notOk(event.originalEvent, 'event is not a jQuery.Event');
                    } else {
                      assert.ok(event instanceof jQuery.Event, 'jQuery event was passed');
                    }
                  },
                },
              });
            },
            /Passing actions to components as strings \(like `({{input key-press="foo"}}|<Input @key-press="foo" \/>)`\) is deprecated\./
          );
        },
        /Passing the `@key-press` argument to <Input> is deprecated\./
      );```

chancancode: Oh hm, I guess expectDeprecation doesn’t support being nested, that’s quite unfortunate but maybe what you has is the best we could do without refactoring that infrastructure. Sorry about that. @rwjblue can you confirm?

PR #19395, opened 2021-02-10
[LTS fix] only return empty href when LinkTo href generation throws error

xg-wang:
## Summary
In 3.24.0-3.24.2 we made changes to let LinkTo href generation returns
empty when initial transition is not started. This is not true and
causing regression on the super-rental tutorial https://github.com/ember-learn/super-rentals-tutorial/pull/176.

In previous tutorials, the LinkTo component will return empty href when
`this.owner.setupRouter()` is not called. It generates valid href if
`setupRouter()` is called and complete model are passed for dynamic
segments.

In our previous changes, we removed the requirement to call
`setupRouter`, and always return empty href when initial transition is
not started. This is not expected when passing complete dynamic segments
without initial transition.

## Test done
I linked the fix and `super-rentals-tutorials` pass test

### DIFF
--- a/packages/@ember/-internals/routing/lib/services/routing.ts
+++ b/packages/@ember/-internals/routing/lib/services/routing.ts
@@ -52,25 +52,28 @@ export default class RoutingService extends Service {
     this.router._prepareQueryParams(routeName, models, queryParams);
   }
 
-  generateURL(routeName: string, models: {}[], queryParams: {}) {
-    let router = this.router;
-    try {
-      let visibleQueryParams = {};
-      if (queryParams) {
-        assign(visibleQueryParams, queryParams);
-        this.normalizeQueryParams(routeName, models, visibleQueryParams as QueryParam);
-      }
+  _generateURL(routeName: string, models: {}[], queryParams: {}) {
+    let visibleQueryParams = {};
+    if (queryParams) {
+      assign(visibleQueryParams, queryParams);
+      this.normalizeQueryParams(routeName, models, visibleQueryParams as QueryParam);
+    }
 
-      return router.generate(routeName, ...models, {
-        queryParams: visibleQueryParams,
-      });
-    } catch (e) {
+    return this.router.generate(routeName, ...models, {
+      queryParams: visibleQueryParams,
+    });
+  }
+
+  generateURL(routeName: string, models: {}[], queryParams: {}) {
+    if (this.router._initialTransitionStarted) {
+      return this._generateURL(routeName, models, queryParams);
+    } else {
       // Swallow error when transition has not started.
       // When rendering in tests without visit(), we cannot infer the route context which <LinkTo/> needs be aware of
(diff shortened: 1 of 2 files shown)


packages/@ember/-internals/routing/lib/services/routing.ts
```diff
@@ -54,21 +54,25 @@ export default class RoutingService extends Service {
...
+        queryParams: visibleQueryParams,
+      });
+    } catch (e) {
+      // Swallow error when transition has not started.
+      // When rendering in tests without visit(), we cannot infer the route context which <LinkTo/> needs be aware of
+      if (!router._initialTransitionStarted) {
+        return;
+      } else {
+        throw e;
+      }
```
rwjblue, line 74: Let's avoid the try/catch when we have started a transition, if we don't do that we break the ability to use the "break on uncaught exception" in the devtools.

I think a plausible solution here is something like:

```js
_generateURL(routeName: string, models: {}[], queryParams: {}) {

}

generateURL(routeName: string, models: {}[], queryParams: {}) {
  let router = this.router;

  if (router._initialTransitionStarted) {
    try {
      return this._generateURL(routeName, models, queryParams);
    } catch (error) {
      return;
    }
  } else {
    return this._generateURL(routeName, models, queryParams);
  }
}
```

rwjblue (changes requested):

rwjblue (approved):

rwjblue:
Very bizarre failure on IE11:

```
not ok 8 IE 11.0 - [98 ms] - Components test: <Input @type='checkbox' />:  sends an action with `<Input EVENT={{action "foo"}} />` for native DOM events
    ---
        actual: >
            touchstart,touchmove,touchend,touchcancel,keydown,keyup,keypress,mousedown,mouseup,contextmenu,change,click,dblclick,submit,input,change,dragstart,drag,dragenter,dragleave,dragover,drop,dragend,focusin
        expected: >
            touchstart,touchmove,touchend,touchcancel,keydown,keyup,keypress,mousedown,mouseup,contextmenu,change,click,dblclick,submit,input,change,dragstart,drag,dragenter,dragleave,dragover,drop,dragend
        stack: >
```

I'm going to restart CI as I don't _think_ any code in `<LinkTo />` should affect `<Input />` tests..

chancancode:
@xg-wang @rwjblue Is this what's causing the tutorial build to fail? https://github.com/ember-learn/super-rentals-tutorial/runs/1891443801?check_suite_focus=true#step:11:2684

If so, do you mind triggering a build on the tutorial after this went out to release/beta (Monday probably?) to confirm that it fixed the issue? You can go find the latest cron job that failed (e.g. https://github.com/ember-learn/super-rentals-tutorial/actions/runs/562659205) and click "Re-run jobs". If it won't let you, you can open a PR with an empty or trivial commit to force a build on the PR itself and close it after it ran successfully.

/cc @kategengler as well

Issue #19393, opened 2021-02-10
[Bug] Incorrect "import htmlSafe from '@ember/string'" deprecation warning when importing from '@ember/template' 


urbany:
Hi everyone, after upgrating to latest stable ember-source@3.25.1 I'm getting:

> DEPRECATION: Importing htmlSafe from '@ember/string' is deprecated. Please import htmlSafe from '@ember/template' instead. [deprecation id: ember-string.htmlsafe-ishtmlsafe]

the thing is my helper is importing correctly:
`import { htmlSafe } from '@ember/template';
`
but when i check the compiled version on the browser it is compiling to:
`Ember.String.htmlSafe`

which is triggering the deprecation warning.

@locks suggested on Discord it might be related to https://github.com/ember-cli/ember-rfc176-data needing an update

locks:
We seem to have missed some necessary work for the correct implementation of the deprecation. I will be using this comment to track the necessary work.

## Tracking

- [ ] Nerf deprecation in a patch
- [ ] export non-deprecated `htmlSafe` and `isHTMLSafe` from a private namespace in the `Ember` global, e.g. `Ember._Template`
- [ ] figure out how to have per-Ember version mappings in ember-rfc176-data
- [ ] incorporate change in [babel-plugin-ember-modules-api-polyfill](https://github.com/ember-cli/babel-plugin-ember-modules-api-polyfill) and [ember-cli-babel](https://github.com/babel/ember-cli-babel)
- [ ] re-introduce deprecations for `htmlSafe` and `isHTMLSafe`
  - [ ] make sure that `String.prototype.htmlSafe` is deprecated

jamescdavis:
As the person who did the [original deprecation PR](https://github.com/emberjs/ember.js/pull/19339), I apologize for the inconvenience!

nadnoslen:
Not that big of a deal @jamescdavis 😉

DLiblik:
> Not that big of a deal @jamescdavis 😉

Tell that to our CI server test logger LOL... just found this out today when our log files went super-nova recently

All good!

PR #19396, opened 2021-02-11
[BUGFIX release] Revert deprecation of htmlSafe and isHTMLSafe

locks:
Starts addressing https://github.com/emberjs/ember.js/issues/19393.

### DIFF
--- a/packages/ember/index.js
+++ b/packages/ember/index.js
@@ -598,7 +598,7 @@ if (EMBER_MODERNIZED_BUILT_IN_COMPONENTS) {
           `Using Ember.${name} or importing from '${path}' has been deprecated, install the ` +
             `\`ember-legacy-built-in-components\` addon and use \`import { ${name} } from ` +
             `'ember-legacy-built-in-components';\` instead`,
-          true,
+          false,
           {
             id: 'ember.legacy-built-in-components',
             until: '4.0.0',
--- a/packages/ember/index.js
+++ b/packages/ember/index.js
@@ -598,7 +598,7 @@ if (EMBER_MODERNIZED_BUILT_IN_COMPONENTS) {
           `Using Ember.${name} or importing from '${path}' has been deprecated, install the ` +
             `\`ember-legacy-built-in-components\` addon and use \`import { ${name} } from ` +
             `'ember-legacy-built-in-components';\` instead`,
-          false,
+          true,
           {
             id: 'ember.legacy-built-in-components',
             until: '4.0.0',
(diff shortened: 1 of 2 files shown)


packages/ember/index.js
```diff
@@ -652,42 +659,6 @@ if (EMBER_GLIMMER_INVOKE_HELPER) {
...
-  },
-});
-Object.defineProperty(Ember.String, 'isHTMLSafe', {
-  enumerable: true,
-  configurable: true,
-  get() {
-    deprecateImportFromString('isHTMLSafe');
-    return isHTMLSafe;
-  },
-});
```
rwjblue, line 689: Hmm, don't we still need these properties? Were they already there?

locks: I moved them up with the rest of of the `Ember.String` methods: https://github.com/emberjs/ember.js/pull/19396/files#diff-2f99d92a23bae60fccaee9fcfeda8b77ca73723657662cfadfa4d82751045e14R453.
Since we're not calling the deprecation method, I thought they could all be assigned together.

rwjblue: D'oh! Sorry about that, I totally missed it in the diff.

> Since we're not calling the deprecation method, I thought they could all be assigned together.

I think that maybe it would be better to keep the infra around for the deprecation, and change the `deprecateImportFromString` to pass `true` to deprecate (with a nice comment explaining that it needs to be changed back to `false` when we have addressed that issue).

packages/ember/index.js
```diff
@@ -652,42 +659,6 @@ if (EMBER_GLIMMER_INVOKE_HELPER) {
...
-if (ENV.EXTEND_PROTOTYPES.String) {
-  String.prototype.htmlSafe = function () {
-    return htmlSafe(this);
-  };
-}
-const deprecateImportFromString = function (
-  name,
-  message = `Importing ${name} from '@ember/string' is deprecated. Please import ${name} from '@ember/template' instead.`
-) {
-  deprecate(message, false, {
```
rwjblue, line 664: I was thinking you could leave all the code, but change this to `true` with a nice comment...

That would make it really easy to bring it back

locks: Totally forgot about that flag 🤦

locks:
I tried making the tests into `todos` but it didn't work 😭

rwjblue (approved):

jordpo:
@rwjblue can we get a patch for 3.25 or should we just wait for 3.26?

chancancode:
Yes we plan on back porting this

Issue #19353, opened 2021-01-23
[Bug] Optional chaining not transpiled for node/FastBoot


simonihmig:
### 🐞 Describe the Bug

The Ember code base contains instances of optional chaining, in this case it' about [this line](https://github.com/emberjs/ember.js/blob/c12e918edbcdc3fc3f42e8677742ec4bcecec00c/packages/%40ember/-internals/runtime/lib/mixins/array.js#L1350). Although targets have been properly defined (`node: 'current'`), when running a fresh new app with FastBoot in node 10, this does not get transpiled as expected, and FastBoot throws.

### 🔬 Minimal Reproduction

Here is a reproduction: https://github.com/simonihmig/ember-transpilation-reproduction. This is basically a fresh new 3.24 app, with `ember-cli-fastboot` installed and `targets.js` updated for node.

Make sure you are using node 10 (pinned for `volta`), then run `yarn start` to see the FastBoot error

### 😕 Actual Behavior

Optional chaining should get transpiled for node 10, which does not support it. But it does not, and node 10 throws this when running the app in FastBoot:

```
/var/folders/q3/yzggf_l965zbzps_bk8xr37c0000gn/T/broccoli-12464N2d65poMnE2Z/out-170-append_ember_auto_import_analyzer/assets/vendor.js:26317
      this.forEach(item => ret.push(item[methodName]?.(...args)));
                                                     ^

SyntaxError: Unexpected token .
    at new Script (vm.js:83:7)
    at VMSandbox.eval (/Users/simonihmig/Projects/test-addon/node_modules/fastboot/src/vm-sandbox.js:13:22)
    at /Users/simonihmig/Projects/test-addon/node_modules/fastboot/src/ember-app.js:190:15
    at Array.forEach (<anonymous>)
    at EmberApp.loadAppFiles (/Users/simonihmig/Projects/test-addon/node_modules/fastboot/src/ember-app.js:187:21)
    at EmberApp.retrieveSandboxedApp (/Users/simonihmig/Projects/test-addon/node_modules/fastboot/src/ember-app.js:235:10)
    at new EmberApp (/Users/simonihmig/Projects/test-addon/node_modules/fastboot/src/ember-app.js:61:21)
    at FastBoot._buildEmberApp (/Users/simonihmig/Projects/test-addon/node_modules/fastboot/src/index.js:114:17)
    at new FastBoot (/Users/simonihmig/Projects/test-addon/node_modules/fastboot/src/index.js:52:10)
[...]

PR #19397, opened 2021-02-11
[BUGFIX lts] Force building Ember bundles when `targets.node` is defined

mansona:
This is my first attempt at fixing the issue described in https://github.com/emberjs/ember.js/issues/19353

This does not follow @pzuraq's recommendation in https://github.com/emberjs/ember.js/issues/19353#issuecomment-766415799 because it would essentially mean that the prebuilt dev bundle would have Node v10 as the lowest common denominator for all modern browser features, which means that we will not be able to make use of native optional chaining until Node v10 is EOL in April (and going forward this is likely going to cause the dev bundles to be a little bit behind constantly)

Instead this is opting out of using the pre-built bundle when you have `node: "current"` or really anything listing `node` in your targets file. This will obviously make the build slower for anyone using fastboot but really we need to make a decision here because the current out-of-the box fastboot experience is very broken since optional chaining was added to the ember.js codebase (e.g. https://github.com/emberjs/ember.js/blob/v3.24.0/packages/%40ember/-internals/runtime/lib/mixins/array.js#L1350 ) 

Let me know if you have any questions 👍

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -270,6 +270,7 @@ module.exports = {
 
     let ember;
     let targets = (this.project && this.project.targets && this.project.targets.browsers) || [];
+    let targetNode = (this.project && this.project.targets && this.project.targets.node) || false;
 
     if (targets.includes('ie 11')) {
       this.ui.writeWarnLine(
@@ -282,7 +283,9 @@ module.exports = {
     if (
       !isProduction &&
       PRE_BUILT_TARGETS.every((target) => targets.includes(target)) &&
-      targets.length === PRE_BUILT_TARGETS.length
+      targets.length === PRE_BUILT_TARGETS.length &&
+      // if node is defined in targets we can't reliably use the prebuilt bundles
+      !targetNode
     ) {
       ember = new Funnel(tree, {
         destDir: 'ember',


lib/index.js
```diff
@@ -282,7 +283,9 @@ module.exports = {
     if (
       !isProduction &&
       PRE_BUILT_TARGETS.every((target) => targets.includes(target)) &&
-      targets.length === PRE_BUILT_TARGETS.length
+      targets.length === PRE_BUILT_TARGETS.length &&
+      // if node is defined in targets we can't reliably use the prebuilt bundles
+      !targetNode
```
rwjblue, line 288: I think this is only true if we are _also_ using `ember-cli-fastboot` or `fastboot`, we can further narrow this down.

mansona: so... I think you are probably right, in that it wouldn't make much sense in having `node: "current"` or any sort of node target in your targets file if you are not using ember-cli-fastboot or fastboot, but I think the code as written is "more correct".

Essentially we are pre-building the ember files for certain targets, and if our **actual** targets are outside that set then we need to revert to providing the tree rather than the pre-built bundle. It just so happens that we don't target node in our pre-built bundles and this is an instant opt out because of that.

Another think we could explore is to just hash the full targets (not just the browsers) in some way and compare that to the hash of the targets that the pre-built files were built with. This would have the same effect since it will likely never have node in the full targets file.

Thoughts?

mansona: So we discussed this in the Fastboot meeting and have agreed that it is probably fine to go ahead with. I have confirmed that the concern @rwjblue had about the potential performance impact of this change is not actually an issue. I mentioned in the meeting that it seemed the same speed to build, but that was without any data so I did a quick comparison: 

Linked to the version from this PR: 

```
➜  ember-website git:(master) time ember build

Missing symlinked npm packages:
Package: ember-source
  * Specified: ~3.24.0
  * Symlinked: 3.27.0

Environment: development
cleaning up...
Built project successfully. Stored in "dist/".
ember build  24.98s user 7.98s system 112% cpu 29.295 total
```

Using the latest release of ember (at the time): 

```
time ember build
Environment: development
cleaning up...
Built project successfully. Stored in "dist/".
ember build  25.77s user 7.59s system 122% cpu 27.223 total
```

Those build times are within the margin of error 👍

pzuraq:
> Because it would essentially mean that the prebuilt dev bundle would have Node v10 as the lowest common denominator for all modern browser features, which means that we will not be able to make use of native optional chaining until Node v10 is EOL in April (and going forward this is likely going to cause the dev bundles to be a little bit behind constantly)

Just curious, what's the motivation here exactly? Is it easier to debug native optional chaining/other non-transpiled features? Is it better for performance?

PR #19398, opened 2021-02-11
[BUGFIX] Ensure entries are properly defined on loader

pzuraq:
Ensures the `registry` field on `Ember.__loader` is setup properly, and
that the shim loader has `entries` defined on it correctly.

### DIFF
--- a/packages/ember/index.js
+++ b/packages/ember/index.js
@@ -790,7 +790,7 @@ function defineEmberTemplateCompilerLazyLoad(key) {
     configurable: true,
     enumerable: true,
     get() {
-      if (require.has('ember-template-compiler')) {
+      if (has('ember-template-compiler')) {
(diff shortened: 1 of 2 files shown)


mixonic (approved):

packages/ember/index.js
```diff
@@ -863,7 +863,8 @@ Ember.__loader = {
   require,
   // eslint-disable-next-line no-undef
   define,
-  registry: require.entries,
+  // eslint-disable-next-line no-undef
+  registry: typeof requirejs !== 'undefined' ? requirejs.entries : require.entries,
```
mixonic, line 867: I was confused as to why `requirejs` would ever be `undefined`, but @rwjblue reminded me that this is the case in Node (fastboot, or even just during template compilation).

PR #19402, opened 2021-02-12
Change the mount API to accept a function for defining engine routes

jsturgis:
This change updates the mount API method to accept a function similar
to how the route API currently works so that engine routes can be defined
in the host application's router.js file.

Some Ember applications utilize in-repo ember-engines strictly for their
lazy loading and code organization features and not for their isolation features.

In this context it makes more sense to define engine routes in the host application's
router.js file.

[RFC](https://github.com/emberjs/rfcs/pull/722)


kategengler:

PR #19405, opened 2021-02-12
[BUGFIX release] Avoid instantiation errors when `app/router.js` injects the router service.

Ravenstine:
Related to my previous PR: #19385

And issue: #17791

@rwjblue Since we talked about this with @snewcomer recently, maybe you can weigh in on this approach?

The intention is to prevent an infinite recursion when `service:router` is invoked before `router:main` has a chance to be initialized and cached.  This PR does it by making `service:router` more "passive" so that the responsibility of forwarding `routeWillChange` and `routeDidChange` events is handled by `router:main`, as opposed to looking up(and possibly instantiating) `router:main` and listening to events on it from `service:router`.  The active behavior of `service:router` instantiating `router:main` seems unexpected, and doesn't appear to be required in any tests.

Specifically, the `service:router` has been changed to only use `router:main` if it's already available and `router:main` now forwards those two route events to `service:router` if that service has been instantiated.  That way there's no possibility of getting caught in a loop if `service:router` happens to be used during initialization of `router:main`, which can happen unexpectedly when using a service from an add-on that references `service:router`.

Definitely open to any kind of feedback.


packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -34,6 +19,17 @@ function cleanURL(url: string, rootURL: string) {
...
+  try {
+    return owner.lookup('router:main', { instantiate: false }) || null;
+  } catch (error) {
+    if (error.message !== 'Could not create factory') {
+      throw error;
+    }
+    return null;
+  }
+}
+
```
Ravenstine, line 32: @rwjblue For some reason including `{ instantiate:false }` when looking up the router results in a `Could not create factory` error, so for now I wrapped this in a try-catch because I'm not sure of another more proper way of looking up something without attempting to create it.  Is there an alternative or do you think that could be a bug?

packages/ember/tests/routing/router_service_test/non_application_test_test.js
```diff
@@ -47,19 +47,36 @@ moduleFor(
...
+      assert.equal(this.routerService.get('rootURL'), null);
+      assert.equal(this.routerService.get('currentRoute'), null);
+    }
+
+    ['@test RouterService properties of router can be accessed with default when router is present'](
+      assert
+    ) {
+      assert.expect(5);
+      let router = this.owner.lookup('router:main');
+      router.setupRouter();
```
Ravenstine, line 60: Some of these tests here are written with the idea that a router has been setup, so I appended this to a few of them.

packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -72,32 +68,10 @@ export default class RouterService extends Service {
       return router;
     }
     const owner = getOwner(this) as Owner;
-    router = owner.lookup('router:main') as EmberRouter;
-    router.setupRouter();
+    router = getRouter(owner) as EmberRouter | null;
```
rwjblue, line 71: I think the original code here is fine, right? As long as we remove the DI injection (which I think was already done?) and remove the `constructor` here completely, I think we should be good to go.

rwjblue (commented):
Since we only need to forward the events when someone calls `routerService.on('routeWillChange')`, why don't we implement `on` and setup the wiring there?

packages/@ember/-internals/routing/lib/services/router.ts
Ravenstine: > Since we only need to forward the events when someone calls routerService.on('routeWillChange'), why don't we implement on and setup the wiring there?

That was actually one solution I initially looked at.  My thought is that, since these events are ultimately emanating from `router:main`, it's easier to reason about to have it be responsible for broadcasting events to where they need to go, rather than have `service:router` be responsible for looking up the router and forwarding events from that to itself and counting on `router:main` being in existence.  Having `service:router` lookup `router:main` when `on` is called, setup listening to events from `router:main`, seems a bit more complicated than just calling `trigger` on `service:router` at the point where we're already emitting these events.  Since `service:router` is really like a safe proxy around `router:main`, I think it makes sense to keep the service as passive as possible, including any hooking up of events.

I think it's not necessarily bad to override `on` in `service:router` to get this job done; philosophically, I believe it's a touch clearer not to do it that way.  A possible drawback to the approach in this PR currently is that `service:router` is always looked up when `route{Will|Did}Change` gets called, but `lookup` _seems_ fairly lightweight.

Ravenstine: > Since we only need to forward the events when someone calls routerService.on('routeWillChange'), why don't we implement on and setup the wiring there?

That was actually one solution I initially looked at.  My thought is that, since these events are ultimately emanating from `router:main`, it's easier to reason about to have it be responsible for broadcasting events to where they need to go, rather than have `service:router` be responsible for looking up the router and forwarding events from that to itself and counting on `router:main` being in existence.  Having `service:router` lookup `router:main` when `on` is called, setup listening to events from `router:main`, seems a bit more complicated than just calling `trigger` on `service:router` at the point where we're already emitting these events.  Since `service:router` is really like a safe proxy around `router:main`, I think it makes sense to keep the service as passive as possible, including any hooking up of events.

I don't think it's bad to override `on` in `service:router` to get this job done; philosophically, I believe it's a touch clearer not to do it that way.  The one advantage to it I see is that setting up event listening would only happen once, whereas looking up `service:router` and calling `trigger` on it would happen every time `route{Will|Did}Change` happens, although it _seems_ like `lookup` is pretty lightweight.

snewcomer: Oh and touching on the `setupRouter`, I though we may have discussed that this might be unnecessary/duplicated.  I could be remembering wrong as well.  Happy to bring that back if you would rather keep it here.   

https://github.com/emberjs/ember.js/blob/8c85b2b52f304d7ca3987f97d9d2fbf18bd40d37/packages/%40ember/-internals/routing/lib/services/routing.ts#L33

rwjblue (commented):
Let's go the route of https://github.com/emberjs/ember.js/pull/19454, and look up the router service during init. That way we can avoid relying on `{ instantiate: false }`. It does add some small amount of overhead (automatic creation of `service:router`), but in practice I don't think that will matter much and the resulting simplicity of the implementation will be easier to maintain.

What do you think @Ravenstine?

Ravenstine:
@rwjblue That makes sense.  Doing `{ instantiate: false }` doesn't seem to really save us much.  @snewcomer and I will work on this tomorrow.

rwjblue:
kk, awesome thanks for taking a look @Ravenstine!

packages/ember/tests/routing/router_service_test/currenturl_lifecycle_test.js
```diff
@@ -241,17 +245,17 @@ moduleFor(
...
             ],
             [
-              'parent.index - parent.index',
+              'parent.index - parent.loading',
               'parent.loading routeWillChange: parent.index - parent.loading',
               '/',
             ],
             [
-              'parent.index - parent.index',
+              'parent.index - parent.loading',
```
Ravenstine, line 258: These tests changed during the development of this PR, so this just gets it to pass with the work we've done here.  Not sure if this it totally correct?  Since we've changed how the router service gets invoked, I think this might be okay.

packages/ember/tests/routing/router_service_test/currenturl_lifecycle_test.js
```diff
@@ -241,17 +245,17 @@ moduleFor(
             ['parent.index - parent.index', 'parent.child model', '/'],
             ['parent.index - parent.index', 'parent.loading activate', '/'],
             [
-              'parent.index - parent.index',
+              'parent.index - parent.loading',
```
sly7-7, line 248: I'm concerned by this change, because this was the goal of https://github.com/emberjs/ember.js/pull/19450/ to make routerService.currentRouteName in sync with routerService.currentRoute.name

I think this check https://github.com/emberjs/ember.js/pull/19405/files#diff-02300d4f8ddd433fa4571addf26231e52255efd632d0b9972b02086584bf15ccR309 should be put at the end of the routeWillChange handler in the router

packages/ember/tests/routing/router_service_test/currenturl_lifecycle_test.js
sly7-7: Hello, thank you for updating this PR.
I was checking this change while you were actually writing you comment. See just below :)

packages/ember/tests/routing/router_service_test/currenturl_lifecycle_test.js
Ravenstine: @sly7-7 Ah, yes.  I wasn't exactly sure whether it was the test that needed fixing, but I understand it better now.  Changing the order of when the event is triggered on the router service seems to pass the tests. (with this test reverted to its original form)

packages/@ember/-internals/routing/lib/services/router.ts
sly7-7: @Ravenstine @snewcomer Hi think keeping the call to router.setupRouter() in the service is ok here. I even think the whole setupRouter() thing should be removed, and the logic should be put in the router constructor.

sly7-7:
Thanks a lot to both of you @snewcomer @Ravenstine. This looks good to me. I just don't know what should be the choice for the setupRouter() call. I have my opinion, but I'm so new to the router area that I probably miss something.

rwjblue (approved):
Awesome, this is great work. Thank you @Ravenstine @snewcomer @sly7-7!

packages/@ember/-internals/routing/lib/services/router.ts
rwjblue: @sly7-7 we can probably remove the lazy initialization code from the router service ya, but I think `router.setupRouter()` itself has to stick around...

Issue #19404, opened 2021-02-12
[Bug] Inline if eagerly evaluates both results before test clause returns


rtablada:
### 🐞 Describe the Bug
In 3.17-3.21 (importantly this includes LTS 3.20) the inline `if` helper eagerly evaluates both return actions regardless of the value of the test clause. This appears to only happen when one of the result actions uses a helper and that helper is passed in a component argument.

### 🔬 Minimal Reproduction

1. Create a new Ember app
2. Create a component `my-component` that uses and inline `if` and helpers for the return values (the test condition does not seem to matter)
3. Invoke the component with an arg used by the if statement results

[Github Repo](https://github.com/rtablada/ember-if-question)
[Code Sandbox](https://codesandbox.io/s/affectionate-waterfall-ko8qu?file=/app/components/navigation.hbs)

### 😕 Actual Behavior
When the inline `if` statement uses a helper with a component argument as a param as one of the possible results, both results are evaluated.

### 🤔 Expected Behavior
The expected behavior (and behavior in 3.16 or >3.22) would be that only the subtree/action for the correct value would be evaluated, and that this would only be evaluated after the logic test was assessed.

### 🌍 Environment

- Ember: - 3.17-3.21
- Node.js/npm: - Node 14
- OS: - OSX
- Browser: - Chrome

rtablada:
I should clarify, for most use cases this likely won't matter since helpers are often stateless. However in the case where a helper could throw an error this would stop rendering and cause a potential app crash.

Example:

```hbs
{{if @param (my-angry-helper @param) (some-safe-helper @value)
```

If `my-angry-helper` throws an error when `@param` is falsey (or more likely undefined) then this error in the VM causes rendering errors.

locks:
Addressed by making user-created helpers lazily computed in the linked PR.

PR #19406, opened 2021-02-12
[BUGFIX lts] Makes user helper computation lazy instead of eager

pzuraq:
From 3.17-3.21, the behavior of certain helpers was slightly different.
Essentially, if we detected that the arguments to the helper were
constant, we would eagerly run the helper to determine if it was
constant. If it was constant, we would then not have to update the
helper.

This was a bit mistaken, because as we realized later on when we fully
converted the VM to use autotracking, we cannot eagerly detect if
arguments are constant in all cases. The solution was to have a system
which would convert cached values to be constant if no autotracked
values were detected during the first computation, whenever that
occured.

Unfortunately, in the affected versions, helpers sometimes run eagerly
when previously they would not. This PR fixes that for user helpers,
which could contain bugs if they ran eagerly when they previously
ran lazily.

Fixes https://github.com/emberjs/ember.js/issues/19404

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/utils/references.ts
+++ b/packages/@ember/-internals/glimmer/lib/utils/references.ts
@@ -2,18 +2,38 @@ import { getDebugName, isObject } from '@ember/-internals/utils';
 import { debugFreeze } from '@ember/debug';
 import { DEBUG } from '@glimmer/env';
 import { CapturedArguments, Environment } from '@glimmer/interfaces';
-import { HelperRootReference, RootReference, VersionedPathReference } from '@glimmer/reference';
+import { InternalHelperFunction, RootReference, VersionedPathReference } from '@glimmer/reference';
 import { PrimitiveReference } from '@glimmer/runtime';
-import { consumeTag, deprecateMutationsInAutotrackingTransaction } from '@glimmer/validator';
+import {
+  combine,
+  consumeTag,
+  createUpdatableTag,
+  deprecateMutationsInAutotrackingTransaction,
+  Revision,
+  Tag,
+  track,
+  UpdatableTag,
+  updateTag,
+  validateTag,
+  valueForTag,
+} from '@glimmer/validator';
 import { HelperInstance, isClassHelper, RECOMPUTE_TAG, SimpleHelper } from '../helper';
 
-export class EmberHelperRootReference<T = unknown> extends HelperRootReference<T> {
+export class EmberHelperRootReference<T = unknown> extends RootReference<T> {
+  private computeRevision: Revision | null = null;
+  private computeValue?: T;
+  private computeTag: Tag | null = null;
+  private valueTag?: UpdatableTag;
+  private fn: InternalHelperFunction<T>;
+
   constructor(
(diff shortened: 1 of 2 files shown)


rtablada:
This would be great to get merged and released to LTS 3.20. Our large app is blocked on this because of errors thrown by eagerly run helpers especially in conjunction with inline `if`

rwjblue:
@pzuraq - Can you add a test that previously failed? Then we can pull that test forward to master to confirm the general use case continues to pass even without these changes...

rwjblue (approved):

rtablada:
@rwjblue is there a word on when this will get a 3.20 LTS patch release? Our team is currently having to use a forked & compiled version.

rwjblue:
:( I forgot about this when doing the other releases this weekend, I'll try to do it tomorrow but please feel free to ping me again if I forget...

rwjblue:
I've just tagged and kicked off publishing of https://github.com/emberjs/ember.js/releases/tag/v3.20.7, it should be published to npm once https://travis-ci.org/github/emberjs/ember.js/builds/773433834 CI job is completed.

ilucin:
@pzuraq @rwjblue I'm trying to upgrade from 3.23 to 3.24 and this same issue is popping up. Is it possible that the regression that occurred in 3.17-21 also happened again in 3.24?

The reproduction is as simple as [this](https://monosnap.com/file/nKRjBjkeSIgvBqDt7IVxsGTiGt6EHL) (`ember new` + one change in `application.hbs`).

Same thing is happening on:
- 3.25.4
- 3.26.1
- 3.27.5

This regression (eager helper invocation) is tricky because it introduces subtle bugs and it's hard to refactor. Can you push this fix to 3.24 and later versions?

pzuraq:
The entire reference system was refactored during that time period, so while the symptoms may be the same, the root cause will be different. The new system is fundamentally lazy, and we have tests to ensure laziness that have not failed including this one, so it's very surprising to me that you're still encountering this issue even in the latest version of Ember. I'll take a look.

pzuraq:
Ah, ok, so this one does make sense because `fn` is special cased. It does not eagerly compute anything in production, but in development builds it does in order to check if users actually passed in a function. This is because we were finding many, many bugs caused by users not actually passing a function in at all, and just assuming one would be there.

So to be clear, this will not happen with _any other helper_. It is only `fn`, only to prevent those types of bugs. But clearly, it's an issue here because it prevents this usage of `fn` which seems completely legitimate 😩 

It makes me wonder at this point if we should make `(fn null)` and `(fn undefined)` just return the value, rather than throwing an error. I'll raise this with the core team and figure out a path forward.

ilucin:
@pzuraq thanks for quick response! Great, this means I can safely update my app, just need to figure what to do in development (and tests also).

My thoughts on this kind of behaviour:
- it's useful (dooh, obviously) :)
- it should be configurable with build flags that can be turned off in development
- the error log should tell me that it is a "development only" error
- discrepancies between development and production environments are undesireable
- shouldn't interrupt legitimate use cases, as you've already mentioned

Maybe if you can detect if there's and `{{if @var}}` expression that is guarding against null call to `fn`?

PR #19407, opened 2021-02-12
Add deprecation for `Route#disconnectOutlet`

patocallaghan:
Adds deprecation for `Route#disconnectOutlet`. See [RFC 491](https://emberjs.github.io/rfcs/0491-deprecate-disconnect-outlet.html) for more info.

I was initially going to ship the `disconnectOutlet`, `render` and `renderTemplate` deprecations together in https://github.com/emberjs/ember.js/pull/19388 but the amount of tests which break and require a `expectDeprecation` is a bit overwhelming and taking longer than expected 😅 I've decided to break this one out instead.

```js
id: 'route-disconnect-outlet',
until: '4.0.0',
url: 'https://deprecations.emberjs.com/v3.x/#toc_route-disconnect-outlet',
for: 'ember-source',
since: {
  enabled: '3.27.0',
},
```

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/application/debug-render-tree-test.ts
+++ b/packages/@ember/-internals/glimmer/tests/integration/application/debug-render-tree-test.ts
@@ -1,11 +1,10 @@
 import {
   ApplicationTestCase,
+  expectDeprecation,
   ModuleBasedTestResolver,
   moduleFor,
   strip,
 } from 'internal-test-helpers';
-import { ExpectDeprecationFunc } from '../../../../../../internal-test-helpers/lib/ember-dev/deprecation';
-declare let expectDeprecation: ExpectDeprecationFunc;
 
 import { ENV } from '@ember/-internals/environment';
 import { Component, setComponentManager } from '@ember/-internals/glimmer';
--- a/packages/@ember/-internals/glimmer/tests/integration/application/debug-render-tree-test.ts
+++ b/packages/@ember/-internals/glimmer/tests/integration/application/debug-render-tree-test.ts
@@ -4,6 +4,8 @@ import {
   moduleFor,
   strip,
 } from 'internal-test-helpers';
+import { ExpectDeprecationFunc } from '../../../../../../internal-test-helpers/lib/ember-dev/deprecation';
+declare let expectDeprecation: ExpectDeprecationFunc;
 
 import { ENV } from '@ember/-internals/environment';
 import { Component, setComponentManager } from '@ember/-internals/glimmer';
@@ -225,7 +227,10 @@ if (ENV._DEBUG_RENDER_TREE) {
               if (showHeader) {
                 this.render('header', { outlet: 'header' });
               } else {
-                this.disconnectOutlet('header');
+                expectDeprecation(
+                  () => this.disconnectOutlet('header'),
+                  'The usage of `disconnectOutlet` is deprecated.'
+                );
               }
             }
           }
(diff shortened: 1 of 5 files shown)


packages/@ember/-internals/glimmer/tests/integration/application/debug-render-tree-test.ts
```diff
@@ -4,6 +4,8 @@ import {
   moduleFor,
   strip,
 } from 'internal-test-helpers';
+import { ExpectDeprecationFunc } from '../../../../../../internal-test-helpers/lib/ember-dev/deprecation';
```
pzuraq, line 7: This should likely be added to the above imports from `internal-test-helpers`, and the `ExpectDeprecationFunc` type/interface should be exported from that package

packages/@ember/-internals/glimmer/tests/integration/application/debug-render-tree-test.ts
```diff
@@ -4,6 +4,8 @@ import {
...
+declare let expectDeprecation: ExpectDeprecationFunc;
```
pzuraq, line 8: we probably should declare this globally somewhere, rather than just in the same file. It can probably be done in the same file that `ExpectDeprecationFunc` is defined

pzuraq (changes requested):
Overall looks good, I think we should handle the types for `expectDeprecation` a bit differently, but other than that this is great!

packages/@ember/-internals/glimmer/tests/integration/application/debug-render-tree-test.ts
patocallaghan: Thanks for the feedback 👍

I'm still trying to get my head around TypeScript. I see there's already a global declaration for `expectDeprecation` and some functions injected into `window`
 https://github.com/emberjs/ember.js/blob/0fa78e0dd84a1c8aa572c2db9bf12664602248bf/packages/internal-test-helpers/lib/ember-dev/deprecation.ts#L15-L22

https://github.com/emberjs/ember.js/blob/0fa78e0dd84a1c8aa572c2db9bf12664602248bf/packages/internal-test-helpers/lib/ember-dev/deprecation.ts#L80

Should I be exporting the `expectDeprecation` function instead of having it created it inline, or is there some other magic I can rely on?

patocallaghan: @pzuraq @rwjblue I'm struggling a bit to figure out how to get the `declare let expectDeprecation: ExpectDeprecationFunc;` typing available from `internal-test-helpers`. My TypeScript foo isn't great.

There was an attempt in [`a7f16ce` (#19407)](https://github.com/emberjs/ember.js/pull/19407/commits/a7f16ce2995376248cbbc155ca69cb928f7a4008) but unfortunately the value is `undefined` which makes as I amn't assigning it to anything. Also it's not straight-forward to export the existing `expectDeprecation` as it's a method on the `DeprecationAssert` and accesses instance methods on the class.

Should the `declare let` be added to the `internal-helpers/index.d.ts` file instead?

pzuraq: Discussed offline, we can export the `expectDeprecation` function in the same file where it is defined as a `let`:

```ts
export let expectDeprecation: ExpectDeprecationFunc = () => {
  throw new Error('err');
};
```

It can then be assigned when we inject the deprecation functions

patocallaghan: @pzuraq This should be ready to go now

rwjblue (approved):

patocallaghan:
@rwjblue I've resolved the conflict 🙇‍♂️

rwjblue:
Thank you @patocallaghan!

PR #19410, opened 2021-02-15
Update test blueprints to support ember-mocha2

yads:
ember-mocha2 is a maintained fork of ember-mocha, until we can merge the fork into the mainline of ember-mocha it would be good to support it.
https://www.npmjs.com/package/ember-mocha2


bertdeblock:
Can be closed since support for `ember-mocha` was dropped.

PR #19412, opened 2021-02-16
[BUGFIX release] Updates Glimmer VM to 0.76.0

### LINKED ISSUE #19341: [Bug] Value of the input is not updated if the it was previously updated with the same string (regression starting 3.22)
### 🐞 Describe the Bug
Let's say you have a native input with a value bound: <input value={{this.myValue}}>.
1) this.myValue is a tracked property. It gets updated to `foo`, the input gets `foo` as a value. 
2) Then the user interacts something different into the input let's say `bar`. 
3) Something triggers update of the `myValue` to `foo`.
4) I expect the input value to be updated to `foo`, but it remains `bar`.

### 🔬 Minimal Reproduction
Here is an integration test which reproduces the issue:
It passes for 3.21.3 and fails in 3.22.0. It fails in 3.24.1 as well.
```
import { setupRenderingTest } from 'ember-qunit';
import { hbs } from 'ember-cli-htmlbars';
import { blur, fillIn, render } from '@ember/test-helpers';
import { action } from '@ember/object';

module('Integration | Component | input value', function(hooks) {
  setupRenderingTest(hooks);

  class MyModel {
    @tracked value = '';

    @action
    onBlur() {
      this.value = 'foo';
    }
  }

  async function testChange(assert, input) {
    await fillIn('.my-input', input);
    await blur('.my-input');

    assert.dom('.my-input').hasValue(`foo`);
  }

  test('it renders', async function(assert) {
    this.model = new MyModel();

    await render(hbs`
      <input
        {{on "blur" this.model.onBlur}}
        class="my-input"
        value={{this.model.value}}
      />
    `);

    await testChange(assert, 'bar');
    await testChange(assert, 'baz');
  });
});
```

### ➕ Additional Context
The use case for that is an input which is used as a parser, user inputs something hits TAB, the parser interprets the input and updates the string. If the results of 2 consecutive parses are the same then the user ends up with the string he filled in, not the parsed one.

### LINKED ISSUE #19400: [Bug] Unable to get length of string in v3.25+
### 🐞 Describe the Bug
Using `{{get this.someString "length"}}` on a string worked up through `v3.24.2`, but fails in `v3.25.0+`.

### 🔬 Minimal Reproduction
In any template (I did this in `application.hbs` of a fresh ember app).

```hbs
{{get this.someString "length"}}
{{log (get this.someString "length")}}
```

### 😕 Actual Behavior
Nothing displays in template and console reports `undefined`.

### 🤔 Expected Behavior
Should output the string length in the template and then in the console.

### 🌍 Environment

- Ember: 3.24.2
- Node.js/npm: v14.15.4
- OS: -OSX
- Browser: Firefox/Chrome/Safari

### LINKED ISSUE #19409: [Bug] `if` helper returns `null` instead of `undefined`
### 🐞 Describe the Bug

The `else` path of an inline `if` helper used to return `undefined`, but starting with v3.25 it now returns `null`, which can be problematic in components that check for the existence of an argument via `this.args.foo !== undefined`.

### 🔬 Minimal Reproduction

```hbs
<SomeComponent @foo={{if this.bar "baz"}} />
```

### 😕 Actual Behavior

When `bar` is truthy then `foo` will be `"bar"`, if not, then it will be `null`.

### 🤔 Expected Behavior

When `bar` is truthy then `foo` will be `"bar"`, if not, then it will be `undefined`.

### 🌍 Environment

- Ember: 3.25.1

pzuraq:
Release notes: https://github.com/glimmerjs/glimmer-vm/releases/tag/v0.76.0

Fixes #19409
Fixes https://github.com/emberjs/ember.js/issues/19400
Fixes https://github.com/emberjs/ember.js/issues/19341


rwjblue (approved):

andreyfel:
Is there a plan to backport this fix to 3.24?

pzuraq:
3.24 was on a much earlier version of the VM, so I had to make a special release branch for it and merge the fix for #19341 to it: https://github.com/glimmerjs/glimmer-vm/pull/1272

The plan is to release that, then merge the fix just for #19341. The other bugfixes are due to the more recent VM refactors, so are not applicable.

andreyfel:
Sounds great! Looking forward to seeing it landed on 3.24!

ryanflowers:
@rwjblue thank you for the fix here, was this ever back ported to 3.24? Ive looked through the release log and I dont see 3.24 having G VM 0.76. I am trying to move to 3.24 however see this issue. 3.25 bring other issues for example named block errors  :( 

https://github.com/emberjs/ember.js/pull/19318

PR #19417, opened 2021-02-18
Fix undefined route info error

dmytro-krekota:
It is for fixing an error which I faced:
_TypeError: Cannot read property 'name' of undefined_

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/router.ts
+++ b/packages/@ember/-internals/routing/lib/system/router.ts
@@ -947,7 +947,7 @@ class EmberRouter extends EmberObject {
    */
   _queryParamsFor(routeInfos: PrivateRouteInfo[]) {
     let routeInfoLength = routeInfos.length;
-    let leafRouteName = routeInfos[routeInfoLength - 1].name;
+    let leafRouteName = routeInfos[routeInfoLength - 1]?.name;
     let cached = this._qpCache[leafRouteName];
     if (cached !== undefined) {
       return cached;


rwjblue (changes requested):
I'm vaguely worried about adding this sort of guard because it often covers up other underlying problems. Do you know why the value was `undefined`? Can you add a test that previously failed and now passes?

sly7-7:
@rwjblue  I'll look into it. There are some open issues mentionning this this Type Error, and currently there is an other spot in the router where we do this kind of check: https://github.com/emberjs/ember.js/blob/56c3520b0e52e5bc7f357b5865c34a23eefd98e7/packages/%40ember/-internals/routing/lib/system/router.ts#L1496
Thinking a bit about it, I can't see a legit case where the RouteInfos can be empty (maybe I misunderstand this thing, but for that means there is no (active ?) route at all, which would be very surprising. (except maybe during the initial transition ?)
I don't know if I'm clear in this description, but in all case, naively I thought after setupRouter(), there was at least one RouteInfos, the one concerning the ApplicationRoute.

dmytro-krekota:
@rwjblue I do not know why the value was undefined, will try to add a test.

@sly7-7 should be an initial transition, but I can not reproduce it now in my app.

sly7-7:
@dmytro-krekota there has been (hopefully) some improvements in the router area, landed in 3.24.3  and 3.25.3. So maybe this is it ?

dmytro-krekota:
@sly7-7 no, I use the 3.12.0 version. Maybe updating can help to fix it.

sly7-7:
> @sly7-7 should be an initial transition, but I can not reproduce it now in my app.

I'm confused here, you can't reproduce it even in your app on ember 3.12.0 version ?
When I reviewed some other issues ending up to this error (`TypeError: Cannot read property 'name' of undefined`), I think they were either playing with loading state, redirect during inititial transition, and query params (maybe with the refreshModel flag set to true)

dmytro-krekota:
@sly7-7 yes.
No "refreshModel" in my code.
"Redirect during the initial transition" - can be, I have some manual route transitions.

sly7-7:
Thank you @dmytro-krekota :)

patrickberkeley:
I'm seeing the originally reported/fixed issue on this PR after upgrading ember from 3.25.0 to 3.26.1 and using the Router service to call `replaceWith` instead of calling `replaceWith` on the route itself. Here's more of the call stack that leads to routeInfos being empty:

```

TypeError: Cannot read property 'name' of undefined     at Class._queryParamsFor (http://localhost:4204/assets/vendor.js:35463:59)     at ApplicationRoute.finalizeQueryParamChange (http://localhost:4204/assets/vendor.js:34405:29)     at Class.triggerEvent (http://localhost:4204/assets/vendor.js:35922:27)     at PrivateRouter.triggerEvent (http://localhost:4204/assets/vendor.js:34763:43)     at PrivateRouter.finalizeQueryParamChange (http://localhost:4204/assets/vendor.js:78116:12)     at PrivateRouter.queryParamsTransition (http://localhost:4204/assets/vendor.js:77611:37)     at PrivateRouter.getTransitionByIntent (http://localhost:4204/assets/vendor.js:77683:37)     at PrivateRouter.transitionByIntent (http://localhost:4204/assets/vendor.js:77632:21)     at PrivateRouter.doTransition (http://localhost:4204/assets/vendor.js:77768:19)     at PrivateRouter.transitionTo (http://localhost:4204/assets/vendor.js:78240:19)
```

When I debug that stack, the new `Transition` the router creates does just have empty routeInfos. In fact everything on it is empty. This `Transition` is what is subsequently used to set private `__STATE__` which is looked up and passed to `_queryParamsFor`.

Note, I only see this issue in tests.

patrickberkeley:
I spoke too soon in my previous comment. After upgrading, I'm not occasionally seeing this error in my app too, not just in tests. In the app, the source of the error is empty routeInfos on the Transition in the Route model hook:

https://github.com/emberjs/ember.js/blob/ac201b8e3e52226dbc74daf7be476985f9109a61/packages/@ember/-internals/routing/lib/system/route.ts

PR #19418, opened 2021-02-18
It's 2021

simon04:


Windvis:
I don't think that's how copyright notices work. After some quick googling I found [different](https://opensource.stackexchange.com/questions/5778/why-do-licenses-such-as-the-mit-license-specify-a-single-year) [sources](https://www.copyrightlaws.com/copyright-symbol-notice-year/) which state that the year in the copyright notice should be the year of first release. So it shouldn't be updated every year..

lifeart (approved):

lifeart:
Lol

PR #19422, opened 2021-02-24
Fix glimmer-component-docs.ts typo in Markdown

lolmaus:
Was also affecting a heading anchor link. Kudos to @kaermorchen.


rwjblue:
Thank you!

PR #19426, opened 2021-02-24
[BUGFIX release] Expose `@glimmer/syntax` from template compiler privately for use in Embroider.

### LINKED ISSUE #18095: Improve template-compiler API for use in embroider
I'm opening this to discuss a solution to an Ember & Embroider integration issue. I don't think it rises to the level of an RFC because this is an extremely low-level API more akin to the boundary between Ember and Glimmer than to user-facing public API.

Embroider [currently does hacks](https://github.com/embroider-build/embroider/blob/da9b3923a2307c9cb498d98cc00c65fb312a3f26/packages/core/src/template-compiler.ts#L43-L103) to get access to the features inside `ember-template-compiler.js` that it needs.

There are two main issues:

1. The template compiler uses shared module state, so any time I want to reconfigure it I need to evade the require cache and eval everything all over again. This makes embroider's tests slower, and it also impacts real app builds because each classic addon with custom AST transforms gets its own separately-configured template compiler instance in Embroider's stage1.

2. Embroider needs to be able to run all configured AST transforms but then print back to HBS. This lets us quarantine custom AST transforms to stage1 only, and have a clean standard template compiler in stage3. We currently do this by getting access to:
  - `import { preprocess } from '@glimmer/syntax'` to parse `(string) => AST`
  - `import { defaultOptions, registerPlugin } from 'ember-template-compiler/lib/system/compile-options'` to ensure we're calling `preprocess` with exactly the right arguments to achieve full compatibility with more traditional ember builds.
  - `import { print } from '@glimmer/syntax'` to print `(AST) => string`

If I could instead have my dream API, it would fix problem 1 by exposing a constructor for making new compiler instances, each with its own independent configuration. And it would fix problem 2 by giving the instances an interface like:

```ts
interface TemplateCompiler {
  parse(moduleName: string, templateContents: string): AST;
  print(AST): string;
  runTransforms(moduleName: string, AST): AST;
}
```

While we're at it, we might consider shipping ember-source with unbundled modules for the template compiler, in addition to the bundled one. That would be a nice way to add new interface without altering the existing interface. And we could include the d.ts files.

rwjblue:
This enables Embroider to remove some gnarly hacks (or at least only resort to them when operating on older Ember).

Closes https://github.com/emberjs/ember.js/issues/18095

### DIFF
--- a/packages/ember-template-compiler/index.ts
+++ b/packages/ember-template-compiler/index.ts
@@ -1,5 +1,6 @@
 import { ENV } from '@ember/-internals/environment';
 import { FEATURES } from '@ember/canary-features';
+import * as _GlimmerSyntax from '@glimmer/syntax';
 import VERSION from 'ember/version';
 import require from 'require';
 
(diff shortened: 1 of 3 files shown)


ef4:
Thanks! Haven't had a chance to try this yet. I think we have been broken under canary for a little while due to template-compiler distribution changes, this gives me a stable target to switch to.

rwjblue:
>  I think we have been broken under canary for a little while due to template-compiler distribution changes

Yes, that is precisely why I did this 😉. Seemed better than just fixing the monkey patch...

@ef4 - The test added here is **basically** exactly what we want to do.

PR #19429, opened 2021-02-25
[DEPRECATE] `registerPlugin` / `unregisterPlugin` and legacy class based AST plugins

rwjblue:
The usage of `registerPlugin`/`unregisterPlugin` **requires** global state mutation which leaks AST plugins between addon's and applications.  Tools like Embroider and ember-cli-htmlbars have avoided this by ensuring that we purge the Node module caching system between each attempt at using the template compiler. This results in _massive_ memory growth (the entire `ember-template-compiler.js` source string is included in memory for each addon in the build, as well as any JIT artificats).

This is **very _bad_**.

This PR deprecates using `registerPlugin`/`unregisterPlugin` along with the legacy AST transform API. All of these APIs are private APIs (as is anything to do with the template compilation AST), but they are used fairly often so a proper deprecation cycle is useful.

Related:

* https://github.com/emberjs/ember.js/issues/18095
* https://github.com/ember-cli/ember-cli-htmlbars/pull/130 (is required to avoid the deprecation)
* [Addons (64) using AST plugins](https://emberobserver.com/code-search?codeQuery=htmlbars-ast-plugin)

### DIFF
--- a/packages/ember-template-compiler/lib/system/compile-options.ts
+++ b/packages/ember-template-compiler/lib/system/compile-options.ts
@@ -1,5 +1,5 @@
 import { EMBER_STRICT_MODE } from '@ember/canary-features';
-import { assert } from '@ember/debug';
+import { assert, deprecate } from '@ember/debug';
 import { assign } from '@ember/polyfills';
 import { PrecompileOptions } from '@glimmer/compiler';
 import { AST, ASTPlugin, ASTPluginEnvironment, Syntax } from '@glimmer/syntax';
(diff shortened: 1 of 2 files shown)


pzuraq (approved):

rwjblue:
https://github.com/ember-cli/ember-cli-htmlbars/pull/661 should resolve the deprecation for ember-cli-htmlbars.

rwjblue:
https://github.com/ember-cli/ember-cli-htmlbars/releases/tag/v5.6.0 avoids the deprecation added here

PR #19436, opened 2021-02-28
[BUGFIX beta] Support observer keys with colons

### LINKED ISSUE #19343: [Bug] Observers don't work if the key contains colon ':' characters
### 🐞 Describe the Bug
Observers internally use colon in the `eventName` and assume there is only one colon when a `split(':')` operation is performed in `activateObserver`. By registering an observer with a key name that contains colons, you can break Ember's assumption.

### 🔬 Minimal Reproduction
```js
emberObj.addObserver('key:with:colons', () => {
  console.log('I will never fire');
});

emberObj.set('key:with:colons', 'foo');
```

### 😕 Actual Behavior
The callback method never fires

### 🤔 Expected Behavior
The callback method fires

### 🌍 Environment

- Ember: - 3.16.10
- Node.js/npm: - node 10.23.0, npm 6.13.7
- OS: - mac mojave 10.14.6
- Browser: - chrome 87.0.4280.141

### ➕ Additional Context
The issue appears to be in the `activateObserver` function

```js
  function activateObserver(target, eventName, sync = false) {
    var activeObservers = getOrCreateActiveObserversFor(target, sync);

    if (activeObservers.has(eventName)) {
      activeObservers.get(eventName).count++;
    } else {
      // the following line assumes there are no colons in the user-specified portion of the eventName
      var [path] = eventName.split(':');
      var tag = (0, _reference.combine)(getChainTagsForKey(target, path));
      activeObservers.set(eventName, {
        count: 1,
        path,
        tag,
        lastRevision: (0, _reference.value)(tag),
        suspended: false
      });
    }
  }
```

One gross fix would be something like
```js
// split the string on colon, drop the final value, re-join on colon
var path = eventName.split(':').slice(0, -1).join(':')
```

But someone with a deeper understanding of the observer code may know of something more elegant

bertdeblock:
Fixes #19343.

### DIFF
--- a/packages/@ember/-internals/metal/lib/observer.ts
+++ b/packages/@ember/-internals/metal/lib/observer.ts
@@ -96,7 +96,7 @@ export function activateObserver(target: object, eventName: string, sync = false
   if (activeObservers.has(eventName)) {
     activeObservers.get(eventName)!.count++;
   } else {
-    let [path] = eventName.split(':');
+    let path = eventName.substring(0, eventName.lastIndexOf(':'));
     let tag = getChainTagsForKey(target, path, tagMetaFor(target), peekMeta(target));
 
     activeObservers.set(eventName, {
--- a/packages/@ember/-internals/metal/tests/observer_test.js
+++ b/packages/@ember/-internals/metal/tests/observer_test.js
@@ -87,6 +87,21 @@ moduleFor(
       assert.equal(count, 1, 'should have invoked observer');
     }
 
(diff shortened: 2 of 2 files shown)


rwjblue (approved):
Generally seems good to me, thanks for working on it!

I've pinged @pzuraq and @krisselden for review also...

pzuraq (approved):

rwjblue:
Thanks!

PR #19442, opened 2021-03-01
Add deprecation for `Route#render` method

patocallaghan:
Adds deprecation for `Route#render`. See [RFC 418](https://emberjs.github.io/rfcs/0418-deprecate-route-render-methods.html) for more info. This deprecation has the same id/url as `Route#renderTemplate` as they were part of the same RFC.

```js
id: 'route-render-template'
until: '4.0.0',
url: 'https://deprecations.emberjs.com/v3.x/#toc_route-render-template'
for: 'ember-source',
since: {
  enabled: '3.27.0',
},
```


packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -1485,6 +1485,29 @@ class Route extends EmberObject implements IRoute {
...
+                    referenced by name. Defaults to the parent template
+    @param {String} [options.outlet] the outlet inside `options.into` to render into.
+                    Defaults to 'main'
+    @param {String|Object} [options.controller] the controller to use for this template,
+                    referenced by name or as a controller instance. Defaults to the Route's paired controller
+    @param {Object} [options.model] the model object to set on `options.controller`.
+                    Defaults to the return value of the Route's model hook
+    @private
+   */
+  _render(name?: string, options?: PartialRenderOptions) {
```
patocallaghan, line 1505: Refactors `render` to be `_render`, and call that from within `renderTemplate` and `render` instead.

rwjblue: Let's use the `symbol` utility method from `@ember/-internals/utils` to make something more truly private.

packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -1652,9 +1675,16 @@ class Route extends EmberObject implements IRoute {
     @public
   */
   render(name?: string, options?: PartialRenderOptions) {
-    let renderOptions = buildRenderOptions(this, name, options);
-    ROUTE_CONNECTIONS.get(this).push(renderOptions);
-    once(this._router, '_setOutlets');
+    deprecate('Usage of `render` is deprecated.', false, {
```
rwjblue, line 1678: Lets include some more information in the deprecation message. I'm thinking `this.routeName` at least...

rwjblue (changes requested):
Implementation generally looks good, only had a couple small suggestions.

patocallaghan:
@rwjblue I think I understand the `symbol` mechanism. Are you suggesting we do something like the following instead?

```js
const RENDER = symbol('render');

// internal functionality now declared outside the class
function _render(name, options) {
  let renderOptions = buildRenderOptions(this, name, options);
  ROUTE_CONNECTIONS.get(this).push(renderOptions);
  once(this._router, '_setOutlets');
}

class Route extends EmberObject implements IRoute {
  constructor() {
    // ... snip ...
    // Make sure to bind to `this`
    this[RENDER] = _render.bind(this);
  }
  // ... snip ...
  renderTemplate(name, options) {
    this[RENDER]();
  }
}
```

rwjblue:
Something more like:

```ts
const RENDER = symbol('render');

class Route extends EmberObject implements IRoute {
  [RENDER]() {
    // the implementation here
  }

  render() {
    // issue a deprecation
    return this[RENDER](...arguments);
  }
}
```

patocallaghan:
@rwjblue I've incorporated your feedback so I think it should be good to go now

rwjblue (approved):

rwjblue:
Thank you!

PR #19448, opened 2021-03-04
Ensure query params are preserved through an intermediate loading state transition

### LINKED ISSUE #19307: [Bug] Query param sticky value lost
### 🐞 Describe the Bug
Query parameters lose their sticky value and revert to the default defined within the controller of the parent route when transitioning to a child route. Both the parent and child route have slow model hooks and loading substates. The last working version that does not have this bug is ember-source `2.22.0`, all versions since then up to `2.24.0-beta.2` contain this bug.

#19249 was aimed at resolving this issue but it seems the use case presented in the reproduction might have slipped by.

### 🔬 Minimal Reproduction
- Checkout the [reproduction repository](https://github.com/patsy-issa/ember-qp-bug-reproduction)
- `yarn install` & `yarn start`
- Navigate to `/media`
- click on the movies link
- select an item from the movie list

### 😕 Actual Behavior

During the transition the list will revert back to the albums for a short duration before transitioning into the correct route. This is due to the `filter` query parameter's value being set to its [default value](https://github.com/patsy-issa/ember-qp-bug-reproduction/blob/1b1f23e5068059e20feb91c60bab0cdb87614b5f/app/controllers/media.js#L6) when the loading route is activating.

ember-qp-bug

### 🤔 Expected Behavior
The `filter` query parameter's value does not change when the loading route is active.

### 🌍 Environment

- Ember: `^2.22.1`
- Node.js/npm: 14.1.0/6.14.4
- OS: macOS Catalina 10.15.6
- Browser: Brave Version 1.18.70

### ➕ Related issues

- #14438
- #16986
- #12107
- #11857

sly7-7:
@rwjblue @rreckonerr I think I've reproduced a failing test that highlights #19307.

Let me know if it's correct (and not a 'false' failing). Then I don't know how to fix, either it's in ember or in the router microlib.

Fixes #19307

### DIFF
--- a/package.json
+++ b/package.json
@@ -141,7 +141,7 @@
     "rollup-plugin-commonjs": "^9.3.4",
     "rollup-plugin-node-resolve": "^4.2.4",
     "route-recognizer": "^0.3.4",
-    "router_js": "^7.1.1",
+    "router_js": "^7.2.0",
     "rsvp": "^4.8.5",
     "serve-static": "^1.14.1",
     "simple-dom": "^1.4.0",
--- a/packages/ember/tests/routing/substates_test.js
+++ b/packages/ember/tests/routing/substates_test.js
@@ -1,5 +1,7 @@
 import { RSVP } from '@ember/-internals/runtime';
 import { Route } from '@ember/-internals/routing';
+import Controller from '@ember/controller';
+
 import { moduleFor, ApplicationTestCase, runTask } from 'internal-test-helpers';
 
 let counter;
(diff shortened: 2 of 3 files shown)


sly7-7:
I think I've found a solution, I mean, at least something that makes this test pass, and not breaking the others.
I'm trying to reproduce this test in the router lib, with no luck at the moment.

sly7-7:
The issue is now (hopefully) fixed. This pr can be merged after a new router.js release, and bumped here.

rwjblue:
Updated to https://github.com/tildeio/router.js/releases/tag/v7.2.0 (which includes https://github.com/tildeio/router.js/pull/319).

PR #19450, opened 2021-03-07
Update current route in willChangeRoute during an intermediate transition

### LINKED ISSUE #19449: [Bug] Router Service doesn't update currentRoute on loading states
### 🐞 Describe the Bug
The Router Service's `currentRoute` property doesn't update on loading states, even though `currentRouteName` does.

### 🔬 Minimal Reproduction
https://ember-twiddle.com/5dbab877e11a84a2ede3267d27f58c40

An example is provided in the Ember Twiddle. Note that the Twiddle is on Ember 3.18.1, but this also happens in the app I work on at Ember 3.24.1.

### 😕 Actual Behavior
When a route has a loading state and there are computeds/observers that rely on `router.currentRoute`, the `currentRoute` never shows loading state routes.

### 🤔 Expected Behavior
There should be loading routes in `currentRoute`.

### 🌍 Environment
- Ember: - 3.18.1, 3.24.1
- Node.js/npm: -
- OS: -
- Browser: -

### ➕ Additional Context
I found the same bug reported in the past at https://github.com/emberjs/ember.js/issues/17162, and it's closed now as seemingly fixed, but the bug is still present.

sly7-7:
Updated to follow @rwjblue comment.
Fixes #19449

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/router.ts
+++ b/packages/@ember/-internals/routing/lib/system/router.ts
@@ -286,6 +286,12 @@ class EmberRouter extends EmberObject {
 
       routeWillChange(transition: Transition) {
         router.trigger('routeWillChange', transition);
+        // in case of intermediate transition we update the current route
+        // to make router.currentRoute.name consistent with router.currentRouteName
+        // see https://github.com/emberjs/ember.js/issues/19449
+        if (transition.isIntermediate) {
+          router.set('currentRoute', transition.to);
+        }
       }
 
       routeDidChange(transition: Transition) {
--- a/package.json
+++ b/package.json
@@ -141,7 +141,7 @@
     "rollup-plugin-commonjs": "^9.3.4",
     "rollup-plugin-node-resolve": "^4.2.4",
     "route-recognizer": "^0.3.4",
-    "router_js": "^7.2.0",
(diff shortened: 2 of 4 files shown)


packages/@ember/-internals/routing/lib/system/router.ts
```diff
@@ -548,7 +548,7 @@ class EmberRouter extends EmberObject {
 
   intermediateTransitionTo(name: string, ...args: any[]) {
     this._routerMicrolib.intermediateTransitionTo(name, ...args);
-
+
```
rwjblue, line 551: some odd whitespace here

rwjblue (commented):
We chatted a little bit about this in discord this afternoon, I think the thing we decided was:

* Update `router_js` to add an `isIntermediateTransition` to the `InternalTransition` base class
* Once landed + released, update EmberRouter's `routeWillChange` to detect that value and set `this.currentRoute` to that transitions `.to` value
* Add some inline comments explaining why 😛

sly7-7:
Will need https://github.com/tildeio/router.js/pull/321 and a then router bump

rwjblue (approved):
Nice work! Thanks for picking this up!

PR #19452, opened 2021-03-08
Prevent changelog script from erroring when the commitInfo object has

kategengler:
no author value


rwjblue (approved):

sly7-7:
@kategengler @rwjblue I'm sorry for this. I think this was caused be by, having a different git config at work using username/email not known by github (we work on a private gitlab instance). I will update my local config when commiting on ember related repos.

kategengler:
@sly7-7 No need to be sorry at all! You're definitely not the only committer without the author property and it is a small guard 😄

PR #19453, opened 2021-03-08
Don't refresh a parent route's model when the query param doesn't change

misterbyrne:
Test case for https://github.com/emberjs/ember.js/issues/17494
Plus a potential (if a bit messy) fix.

Routes that have a query param with a default value and `refreshModel: true`, will be refreshed each time a transition happens - but only if that transition is invoke from the router service.

This works as expected if you invoke the transition using the internal router. This is because a some of the internal router's related behaviour is explicitly bimodal - there's a boolean parameter named `_fromRouterService` that controls this.

The root of the problem seems to be in the `_pruneDefaultQueryParamValues` method. If a transition is invoked from the router service, this currently strips default query params. This means that when later compared to the current state, the query params appear unequal, resulting in a refresh.

I've added a fix which prevents the default param values from being stripped for this specific invocation - with yet another boolean param (named `_pruneDefaultQueryParamValues `). Not ideal.

### DIFF
--- a/packages/ember/tests/routing/router_service_test/transitionTo_test.js
+++ b/packages/ember/tests/routing/router_service_test/transitionTo_test.js
@@ -363,6 +363,64 @@ moduleFor(
       });
     }
 
+    async ['@test RouterService#transitionTo should not refresh parent models if the parent query params stay the same'](
+      assert
+    ) {
+      let parentModelHookCallCount = 0;
+      this.add(
+        'route:parent',
+        Route.extend({
+          queryParams: {
+            parentParam: {
+              refreshModel: true,
+            },
+          },
+          model() {
+            parentModelHookCallCount++;
+            return 'model';
+          },
+        })
+      );
+      this.add(
+        'route:parent.child',
+        Route.extend({
+          queryParams: {
+            childParam: {
+              refreshModel: true,
+            },
+          },
+        })
+      );
+      this.add(
+        'controller:parent',
+        Controller.extend({
+          parentParam: 'parentParamDefaultValue',
+        })
+      );
+      this.add(
+        'controller:parent.child',
+        Controller.extend({
+          childParam: 'childParamDefaultValue',
+        })
+      );
+
+      await this.visit('/?parentParam=parentParamDefaultValue');
+      assert.equal(parentModelHookCallCount, 1);
+
+      try {
+        await this.routerService.transitionTo('parent.child', {
+          queryParams: {
+            childParam: 'nonDefaultValue',
+          },
+        });
+      } catch (e) {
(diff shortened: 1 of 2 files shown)


packages/ember/tests/routing/router_service_test/transitionTo_test.js
```diff
@@ -364,6 +364,62 @@ moduleFor(
...
+      );
+      this.add(
+        'controller:parent.child',
+        Controller.extend({
+          childParam: 'childParamDefaultValue'
+        })
+      );
+
+      await this.visit('/?parentParam=parentParamDefaultValue');
+      assert.equal(parentModelHookCallCount, 1);
```
sly7-7, line 409: This looks weird to me. If I'm not missing something, the app should stay in the index route at this time, so I don't think the model hook of the `parent` route should have been called.

sly7-7:
@misterbyrne Thank you :). I'm trying to figure out the router works these days, and I discovered https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md#query-parameter-semantics which I think is what you describe. So I would say the behavior is intented. That beeing said I find that inconsistent behavior may be error prone.
Plus, in the current documentation: https://guides.emberjs.com/release/routing/query-params/#toc_default-values-and-deserialization this is still referring to the default qp value are not in the url.
You have to go to https://api.emberjs.com/ember/3.25/classes/RouterService/methods/transitionTo?anchor=transitionTo to see the description of the default param value behavior.

cc/ @rwjblue

packages/ember/tests/routing/router_service_test/transitionTo_test.js
```diff
@@ -364,6 +364,64 @@ moduleFor(
...
+      this.add(
+        'controller:parent.child',
+        Controller.extend({
+          childParam: 'childParamDefaultValue',
+        })
+      );
+
+      await this.visit('/?parentParam=parentParamDefaultValue');
+      assert.equal(parentModelHookCallCount, 1);
+
```
sly7-7, line 410: I'm what is the this.routerService.currentURL at this point.

lhmzhou (approved):

packages/ember/tests/routing/router_service_test/transitionTo_test.js
misterbyrne: It's `/?parentParam=parentParamDefaultValue`. I assume that this is because the when transitioning to a support URL, the query params are left alone. But when transitioning using a route name and `queryParams` hash, default values are not included in the resulting URL.

kategengler:
As we plan to rework the router for Polaris, we're hesitant to touch existing router code (every bug we fix is one someone else relies upon).

PR #19454, opened 2021-03-09
inject the router service inside ember-router

sly7-7:
Related #19405 (maybe replace?)

I don't know if it's the right way, and feel maybe weird.
 
cc/ @Ravenstine  @rwjblue

### DIFF
--- a/packages/@ember/-internals/routing/lib/services/router.ts
+++ b/packages/@ember/-internals/routing/lib/services/router.ts
@@ -5,26 +5,11 @@ import { assert } from '@ember/debug';
 import { readOnly } from '@ember/object/computed';
 import { assign } from '@ember/polyfills';
 import Service from '@ember/service';
-import { DEBUG } from '@glimmer/env';
 import { consumeTag, tagFor } from '@glimmer/validator';
-import { Transition } from 'router_js';
 import EmberRouter, { QueryParam } from '../system/router';
 import { extractRouteArgs, resemblesURL, shallowEqual } from '../utils';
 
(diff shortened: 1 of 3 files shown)


rwjblue:
Conceptually this seems good to me also, does it pass tests?

What do you think @ravenstine @snewcomer?

sly7-7:
Unless I missed something, they pass locally. Thanks to @Ravenstine for writing it :) in it's PR.
I realize some work has been done between his PR and the master, I just had to wire the router service into the ember router

snewcomer:
@sly7-7 This seems like a close direct copy with some modifications.  Why don't we continue with comments and iterate on @Ravenstine's PR?

rwjblue:
Either is fine, @sly7-7 was using the `Co-authored-by` thing to attribute the work from @ravenstine, but I'm not sure why it doesn't show that in the commit above. 🤔

rwjblue:
OK, so looks like this strategy definitely works.

@ravenstine - Think you can incorporate this into https://github.com/emberjs/ember.js/pull/19405?

rwjblue:
Thanks for testing this out @sly7-7! I'm going to go ahead and close this one, in favor of landing these tweaks over in #19405.

snewcomer:
That certainly is my fault for making that assumption and I sincerely apologize!  

So the only difference is Ben's throws an error if the router svc isn't available whereas yours fails gracefully.  We weren't sure which approach is preferred but throwing the error caught a few tests that needed to `setupRouter` explicitly.

https://github.com/emberjs/ember.js/pull/19405/files#r575540397

PR #19456, opened 2021-03-10
create and (trying to) fix async loading of Routes 

sly7-7:
This is a follow up of https://github.com/emberjs/ember.js/pull/19266

### DIFF
--- a/packages/ember/tests/routing/model_loading_test.js
+++ b/packages/ember/tests/routing/model_loading_test.js
@@ -2,11 +2,15 @@
 import { Route } from '@ember/-internals/routing';
 import Controller from '@ember/controller';
 import { Object as EmberObject, A as emberA } from '@ember/-internals/runtime';
-import { moduleFor, ApplicationTestCase, getTextOf, runLoopSettled } from 'internal-test-helpers';
+import {
+  moduleFor,
+  ApplicationTestCase,
+  getTextOf,
+  runLoopSettled,
+  lazyLoadingRouterOptions,
+} from 'internal-test-helpers';
 import { run } from '@ember/runloop';
 import { computed, set } from '@ember/-internals/metal';
-import { isDestroying } from '@glimmer/destroyable';
-import RSVP from 'rsvp';
 
 let originalConsoleError;
 
@@ -958,49 +962,7 @@ moduleFor(
   'Route - model loading (simulated within lazy engine)',
   class extends LoadingTests {
     get routerOptions() {
-      return {
-        location: 'none',
-        setupRouter() {
-          this._super(...arguments);
-          let getRoute = this._routerMicrolib.getRoute;
-          this._enginePromises = Object.create(null);
-          this._resolvedEngines = Object.create(null);
-
-          let routes = new Map();
-          let routePromises = new Map();
-          this._routerMicrolib.getRoute = (name) => {
-            if (routes.has(name)) {
-              return routes.get(name);
-            }
-
-            if (routePromises.has(name)) {
-              return routePromises.get(name);
-            }
-
-            // if (name.indexOf('loading')) {
-            //   let route = getRoute(name);
-            //   routes.set(name, route);
-            //   return route;
-            // }
-
-            let promise = new RSVP.Promise((resolve) => {
-              setTimeout(() => {
-                if (isDestroying(this)) {
-                  return;
-                }
-
-                let route = getRoute(name);
-
-                routes.set(name, route);
-                resolve(route);
-              }, 10);
-            });
-            routePromises.set(name, promise);
-
-            return promise;
-          };
-        },
-      };
+      return lazyLoadingRouterOptions;
     }
   }
 );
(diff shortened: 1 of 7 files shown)


packages/ember/tests/routing/model_loading_test.js
```diff
@@ -968,6 +977,12 @@ moduleFor(
               return routePromises.get(name);
             }
 
+            // if (name.indexOf('loading')) {
```
sly7-7, line 980: @rwjblue basically, if you uncomment these lines, the tests are green again.

rwjblue:
Chatted a bit with @sly7-7 about this in discord this morning, the conclusion we came to was that Ember should never attempt to enter into a loading or error route that itself is not loaded. 

I believe that @sly7-7 is going to update this code:

https://github.com/emberjs/ember.js/blob/b945e5c963fc75950186bd2c4213614ea253e8b0/packages/%40ember/-internals/routing/lib/system/router.ts#L1411-L1416

Check that `getRoute` returns a non-promise before returning true.

_Something_ like:

```js
function routeHasBeenDefined(owner: Owner, router: any, localName: string, fullName: string) {
  let routerHasRoute = router.hasRoute(fullName);
    
  if (routeHasRoute && !isPromise(router.getRoute(fullName))) {
    let ownerHasRoute =
      owner.hasRegistration(`template:${localName}`) || owner.hasRegistration(`route:${localName}`);

    return ownerHasRoute;
  }

  return false;
}
```

packages/ember/tests/routing/model_loading_test.js
```diff
@@ -815,22 +816,26 @@ class LoadingTests extends ApplicationTestCase {
...
-      run(() => router.send('showPost', { id: '2' }));
-      run(() => router.send('editPost'));
-      assert.equal(editCount, 2, 'set up the edit route twice without failure');
-      assert.deepEqual(editedPostIds, ['1', '2'], 'modelFor posts.post returns the right context');
-    });
+    await this.visit('/posts/1');
+    assert.ok(true, '/posts/1 has been handled');
+    let router = this.applicationInstance.lookup('router:main');
+    run(() => router.send('editPost'));
+    await runLoopSettled();
```
sly7-7, line 831: @rwjblue I have doubt about that artificial synchronisation

packages/ember/tests/routing/model_loading_test.js
```diff
@@ -941,6 +946,14 @@ class LoadingTests extends ApplicationTestCase {
         assert.equal(childcount, 2);
       });
   }
+
+  ['@test What about loading states'](assert) {
```
sly7-7, line 950: Just a reminder to me, to remove this test

packages/internal-test-helpers/lib/lazy-router-options.ts
```diff
@@ -0,0 +1,42 @@
+import { Router } from '@ember/-internals/routing';
+import { isDestroying } from '@glimmer/destroyable';
+import RSVP from 'rsvp';
+
+export default class LazyRouter extends Router {
```
sly7-7, line 5: Not sure if the name fits well.

sly7-7:
@rwjblue Hi, so I've played with that PR, and I'm blocked, mainly because I miss knowledge on the testing infrastructure, and also when it comes to engines.
It seems to me that I need some waiters helpers (in order to check the loading states), but I couldn't find a way to make them work.
Also, I noted this test: https://github.com/emberjs/ember.js/blob/master/packages/@ember/-internals/glimmer/tests/integration/application/engine-test.js#L26 use quite but not the same routerOptions() as in the 'helper' I created. I was wondering it their purposes are the same or not.
Let me know when we can eventually schedule a chat to talk about that, or if you see what I mean here, you could also help me with tiny snippets here. As you prefer :)

kategengler:
Is this still relevant?

sly7-7:
@kategengler I'm sorry I didn't notice your question until now 😞  I don't know if this is still relevant, as far as I remember, an error could happen when a route is loaded by an engine.
There was also a PR in the router https://github.com/tildeio/router.js/pull/322 and quickly looking at the code this one may be still relevant (but lack of tests that I failed to write).
I don't know if the router in ember has been touched sinced 3 years now, but if nothing has changed, then maybe this PR also could be relevant. I've never encountered this use-case in prod app, so I couldn't tell if the bug mentioned could really occur.

bwbuchanan:
I'm still seeing this `stashParamNames` error in an app I'm developing that makes use of loading states. Have been working around it for a while by applying the patch in #19266

NullVoxPopuli:
given that we have `@embroider/router`, which does async-loading of routes, do we still need this?

sly7-7:
@NullVoxPopuli I'm sorry not seeing this earlier. I really think this could be closed

bwbuchanan:
I haven't needed to patch around this bug since migrating away from `@embroider/compat`. I expect that it still affects legacy apps using route splitting and loading states, however.

PR #19458, opened 2021-03-12
Refactor router: remove owner presence checks

sly7-7:
@rwjblue This follows the discussion we had about this. 
That way, I think it removes a useless small mental overhead.
Also as it was a concern only in tests, I think it's better.

Feedback welcome :)

To go a little further, I was wondering if the setupRouter() thing could be removed too ?
Maybe it's by design, but usually it feels wrong to me when there is such a 'two steps construction' before we can actually interact with an instance.

### DIFF
--- a/packages/@ember/-internals/routing/tests/system/router_test.js
+++ b/packages/@ember/-internals/routing/tests/system/router_test.js
@@ -1,4 +1,3 @@
-import { setOwner } from '@ember/-internals/owner';
 import HashLocation from '../../lib/location/hash_location';
 import HistoryLocation from '../../lib/location/history_location';
 import AutoLocation from '../../lib/location/auto_location';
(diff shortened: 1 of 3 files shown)


sly7-7:
🤔  The failing browserstack tests are not related to this, are they ?

rwjblue:
> I was wondering if the setupRouter() thing could be removed too ?

Ya, this is intentional. There are a number of scenarios where we only want to do partial setup. For example, in rendering tests (for a normal ember-cli project) we might want to render `<a href=` content but **not** perform initial routing (since you aren't on any specific route in that scenario).

packages/@ember/-internals/routing/tests/system/router_test.js
```diff
@@ -42,12 +40,6 @@ moduleFor(
       owner = null;
     }
 
-    ['@test can create a router without an owner'](assert) {
-      createRouter(undefined, { disableSetup: true, skipOwner: true });
-
-      assert.ok(true, 'no errors were thrown when creating without a container');
-    }
```
rwjblue, line 49: This test is very odd, :+1: on removing

rwjblue (approved):
Thanks for cleaning this up! I restarted CI (the failure in IE11 was unrelated).

rwjblue:
@sly7-7 - I've made a merge conflict by landing #19405 first, would you mind resolving?

rwjblue (approved):

rwjblue:
Thanks!

Issue #19447, opened 2021-03-04
[Bug] destroyModifier triggers ember-data:method-calls-on-destroyed-store


wongpeiyi:
### 🐞 Describe the Bug
Upon acceptance test completion, modifier destroy on app teardown results in `ember-data:method-calls-on-destroyed-store` warning if the modifier args is a RecordArray

This happens from 3.22 onwards. I believe this happens because teardown results in the modifier calling `registerDestructor(state, () => delegate.destroyModifier(instance, state.args));` but the args then accesses the store when `store.isDestroying`

Not sure if its more an ember data issue and should be posted there instead. 

Perhaps related to https://github.com/emberjs/ember.js/issues/19162 ? 

### 🔬 Minimal Reproduction

Might be totally off in my understanding of what's happening, so I came up with a minimal reproduction here: https://github.com/wongpeiyi/modifier-destroy 

### 😕 Actual Behavior

Acceptance test results in deprecation warnings

### 🤔 Expected Behavior

Warnings shouldn't be shown. Modifier shouldn't recompute args on destroy?

### 🌍 Environment

- Ember: 3.22 - 3.25

boris-petrov:
This is perhaps fixed by https://github.com/glimmerjs/glimmer-vm/pull/1276?

wongpeiyi:
That looks promising, thank you

kategengler:
Going to assume this has been fixed by that linked PR

PR #19469, opened 2021-03-19
[BUGFIX release] Prevent eager argument consumption on modifier destruction

### LINKED ISSUE #19131: [Bug] Ember Data error after upgrading Ember to 3.21.1
### 🐞 Describe the Bug
This is an odd one - I wasn't sure if to raise here or in ED. (I tried discord too)

After upgrading Ember from 3.19 to 3.21.1, the demo app test suite fails with an Ember Data error. 

```
You looked up the 'child' relationship on a 'parent' with id 1 but some of the associated records
were not loaded. Either make sure they are all loaded together with the parent record, or specify
that the relationship is async (`belongsTo({ async: true })`)
```

### 🔬 Minimal Reproduction

I have whittled down the cause to the following reproduction app

[https://github.com/amk221/-ember-inverse-rel-error](https://github.com/amk221/-ember-inverse-rel-error/commit/6bcd6961b6d07d03556b5a8559d36cd585c1338f#diff-539124910b2d9910320f2b3c5516639c1a1e417847d0ae2769c1f839e13b713dR5)

### 😕 Actual Behavior
The app bombs out with the above error 

### 🤔 Expected Behavior
The error should not happen, because all the required data is present. And the error does not happen < 3.21.1.

### 🌍 Environment

- Ember: 3.19/3.21.1
- Node.js/npm: -
- OS: -
- Browser: -

### ➕ Additional Context
It's something to do with the `inverse` relationship.

pzuraq:
Upgrades the VM to include the fix for eager argument consumption on
modifier destruction.

Fixes #19131 on canary


rwjblue:
@pzuraq - Is this just for canary, or should we backport to release?

pzuraq:
We can backport this to release, for 3.24 and 3.20 LTS it'll have to be a different bugfix altogether because the VM was different back then. Luckily this code was in Ember at that time, so should be pretty easy to fix.

rwjblue (approved):

boris-petrov:
Perhaps also fixes https://github.com/emberjs/ember.js/issues/19447?

PR #19470, opened 2021-03-19
[Bug]: set symbol on factory objects only in Prod for sealed objects

snewcomer:
ref https://github.com/emberjs/data/pull/7435

https://github.com/emberjs/ember.js/blob/725cdf29c5d367458a71a313686970db703ca5fa/packages/%40ember/-internals/metal/lib/mixin.ts#L589

Screen Shot 2021-03-19 at 12 36 27 PM

### DIFF
--- a/packages/@ember/-internals/metal/lib/mixin.ts
+++ b/packages/@ember/-internals/metal/lib/mixin.ts
@@ -579,9 +579,6 @@ export default class Mixin {
     if (DEBUG) {
       // Eagerly add INIT_FACTORY to avoid issues in DEBUG as a result of Object.seal(mixin)
       this[INIT_FACTORY] = null;
-    }
-
-    if (DEBUG) {
       /*
         In debug builds, we seal mixins to help avoid performance pitfalls.
 
--- a/packages/@ember/-internals/metal/lib/mixin.ts
+++ b/packages/@ember/-internals/metal/lib/mixin.ts
@@ -1,6 +1,7 @@
 /**
 @module @ember/object
 */
+import { INIT_FACTORY } from '@ember/-internals/container';
 import { ENV } from '@ember/-internals/environment';
 import { Meta, meta as metaFor, peekMeta } from '@ember/-internals/meta';
 import {
@@ -574,6 +575,9 @@ export default class Mixin {
     this.mixins = buildMixinsArray(mixins);
     this.ownerConstructor = undefined;
     this._without = undefined;
+
+    // This exists for objects that `Object.seal` or `Object.freeze` in DEBUG only
+    Object.defineProperty(this, INIT_FACTORY, { value: null });
 
     if (DEBUG) {
       /*
--- a/packages/@ember/-internals/metal/lib/mixin.ts
+++ b/packages/@ember/-internals/metal/lib/mixin.ts
@@ -576,8 +576,10 @@ export default class Mixin {
     this.ownerConstructor = undefined;
     this._without = undefined;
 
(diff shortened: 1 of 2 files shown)


packages/@ember/-internals/metal/lib/mixin.ts
```diff
@@ -575,6 +576,9 @@ export default class Mixin {
     this.ownerConstructor = undefined;
     this._without = undefined;
 
+    // This exists for objects that `Object.seal` or `Object.freeze` in DEBUG only
```
snewcomer, line 579: Perhaps I could add a better comment here.  `seal` is the one that prevents adding.  `freeze` also prevents mutating, so if that was the case for the mixin, this would still fail in ember-data.

Perhaps

```
// We eagerly add INIT_FACTORY to avoid issues in DEBUG as a result of Object.seal(mixin)
```

packages/@ember/-internals/metal/lib/mixin.ts
```diff
@@ -575,6 +576,9 @@ export default class Mixin {
     this.ownerConstructor = undefined;
     this._without = undefined;
 
+    // Eagerly add INIT_FACTORY to avoid issues in DEBUG as a result of Object.seal(mixin)
+    Object.defineProperty(this, INIT_FACTORY, { value: null, writable: true });
```
rwjblue, line 580: If this is only needed in `DEBUG` builds, can you wrap this in a guard (or move it just below into the existing `if (DEBUG) {` check just below)?

rwjblue: Also, why do we need `Object.defineProperty` here? Wouldn't this do the trick:

```js
this[INIT_FACTORY] = null;
```

packages/@ember/-internals/metal/lib/mixin.ts
```diff
@@ -575,6 +576,9 @@ export default class Mixin {
     this.ownerConstructor = undefined;
     this._without = undefined;
 
+    // Eagerly add INIT_FACTORY to avoid issues in DEBUG as a result of Object.seal(mixin)
+    this[INIT_FACTORY] = null;
```
rwjblue, line 580: I'm still curious RE: debug guard here. Have you looked at that angle yet? It _seems_ like we should probably only need this when in debug so moving this down (into the if debug guard) would work just the same...

snewcomer: Oh yes, you are correct.  Sorry about that.  I was confusing myself since the ultimate assignment is a production path.  But the `Object.seal` I believe only happens in DEBUG.  Fixed!

pzuraq (approved):

packages/@ember/-internals/metal/lib/mixin.ts
```diff
@@ -575,6 +576,11 @@ export default class Mixin {
     this.ownerConstructor = undefined;
     this._without = undefined;
 
+    if (DEBUG) {
+      // Eagerly add INIT_FACTORY to avoid issues in DEBUG as a result of Object.seal(mixin)
+      this[INIT_FACTORY] = null;
+    }
+
     if (DEBUG) {
```
rwjblue, line 578: Sorry @snewcomer for the run around here, would you mind merging these two `if (DEBUG) {` statements? They are back to back, and it looks a bit odd...

snewcomer: I clearly don't have my thinking hat on today.  Fixed!

rwjblue (approved):

rwjblue:
Thank you!

PR #19471, opened 2021-03-19
[FEATURE routing-router-service-refresh] Add a refresh method to the router service

Windvis:
This adds the refresh method to the RouterService as described in RFC 631.

RFC PR: https://github.com/emberjs/rfcs/pull/631


packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -393,6 +394,32 @@ export default class RouterService extends Service {
...
+   */
+  refresh(routeName?: string) {
+    if (!routeName) {
+      return this._router._routerMicrolib.refresh();
+    }
+
+    assert(`The route "${routeName}" was not found`, this._router.hasRoute(routeName));
+    assert(`The route "${routeName}" is currently not active`, this.isActive(routeName));
+
+    let pivotRoute = this._router._routerMicrolib.getRoute(routeName) as Route;
```
Windvis, line 418: I'm not sure if this is the best way to get a reference to the route instance.

Windvis: I also added `as Route` since TS used the following type:
```ts
abstract getRoute(name: string): T | Promise<T>;
```

And that isn't compatible with the `refresh` method of `_routerMicrolib` since it expects a `Route`.

packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -393,6 +394,32 @@ export default class RouterService extends Service {
     return this._router._routerMicrolib.recognizeAndLoad(internalURL);
   }
 
+  /**
+   * Refreshes all currently active routes, doing a full transition.
```
Windvis, line 398: I just copied the documentation text from the RFC. Should I add example code similar to the other methods?

Windvis:
[The RFC comments mentioned](https://github.com/emberjs/rfcs/pull/631#issuecomment-668311276) that the Route#refresh method should be reimplemented to use the RouterService#refresh method. Should that be done? And if so, should it be done as part of this PR?

packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -393,6 +394,32 @@ export default class RouterService extends Service {
...
+   * @param {String} [routeName] the route to refresh (along with all child routes)
+   * @return Transition
+   * @public
+   */
+  refresh(routeName?: string) {
+    if (!routeName) {
+      return this._router._routerMicrolib.refresh();
+    }
+
+    assert(`The route "${routeName}" was not found`, this._router.hasRoute(routeName));
```
Windvis, line 415: I'm not sure if these are clear enough.I used the same assert as in the [router](https://github.com/emberjs/ember.js/blob/d414901668464b86fa96e2c724a8ce907d2b064d/packages/%40ember/-internals/routing/lib/system/router.ts#L878) for the first one.

packages/ember/tests/routing/router_service_test/refresh_test.js
```diff
@@ -0,0 +1,99 @@
...
+      await this.routerService.refresh();
+      assert.equal(parentCounter, 5);
+      assert.equal(childCounter, 3);
+      assert.equal(sisterCounter, 2);
+    }
+
+    async ['@test RouterService#refresh verifies that the provided route exists'](assert) {
+      await this.visit('/');
+
+      assert.throws(() => {
```
Windvis, line 84: Is this the best way to test if an assert is thrown?

Windvis: It seems to fail CI, so I'm really starting to think this isn't the way :smile:. Any pointers?

Windvis: I think I found what I was looking for: https://github.com/emberjs/ember.js/blob/22bfcfdac0aeefcf333fb2d6697772934201b43b/packages/internal-test-helpers/lib/ember-dev/assertion.ts#L9

packages/@ember/-internals/routing/lib/services/router.ts
knownasilya: could you do `getRoute<Route>(routeName)`?

Windvis: @knownasilya That triggers a type error :smile:. `Expected 0 type arguments, but got 1.`

The issue is that the [router.js getRoute method](https://github.com/tildeio/router.js/blob/4f5d411e9ba44efea003bc3159484b4060ebfc2a/lib/router/router.ts#L58) has the following type definition:
`abstract getRoute(name: string): T | Promise<T>;`

While the refresh method [expects a Route type](https://github.com/tildeio/router.js/blob/4f5d411e9ba44efea003bc3159484b4060ebfc2a/lib/router/router.ts#L882).

So without the `as Route` the following type error is shown:
`Argument of type 'Route | Promise<Route>' is not assignable to parameter of type 'Route | undefined'.`

When I look at [the actual getRoute implementation](https://github.com/emberjs/ember.js/blob/22bfcfdac0aeefcf333fb2d6697772934201b43b/packages/%40ember/-internals/routing/lib/system/router.ts#L190) the return type doesn't include the Promise<T> though. So I'm not sure what is up with that, but I have limited TS experience :smile:.

packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -393,6 +394,32 @@ export default class RouterService extends Service {
...
+   * Returns a promise that will be resolved once the refresh is complete.
+   * All resetController, beforeModel, model, afterModel, redirect, and setupController
+   * hooks will be called again. You will get new data from the model hook.
+   *
+   * @method refresh
+   * @param {String} [routeName] the route to refresh (along with all child routes)
+   * @return Transition
+   * @public
+   */
+  refresh(routeName?: string) {
```
Windvis, line 410: [The RFC named this `pivotRouteName`.](https://emberjs.github.io/rfcs/0631-refresh-method-for-router-service.html#detailed-design) which I found a little confusing since it's a new concept that isn't used anywhere else. It's named like that in router.js as well, but I'm not sure if it should be exposed like that to the end user.

Using `routeName` is just a personal preference and I'm not sure how much room for deviation there is when implementing an RFC, so let me know if I should change it back :smile:.

Windvis, line 410: It seems I need to add a feature flag as well.

Windvis: Renamed it back to pivotRouteName anyway, since that is how it was accepted.

rwjblue:
Mind rebasing (there is a conflict now in packages/@ember/canary-features/index.ts)?

packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -474,6 +476,39 @@ export default class RouterService extends Service {
...
+        return this._router._routerMicrolib.refresh();
+      }
+
+      assert(`The route "${pivotRouteName}" was not found`, this._router.hasRoute(pivotRouteName));
+      assert(
+        `The route "${pivotRouteName}" is currently not active`,
+        this.isActive(pivotRouteName)
+      );
+
+      let pivotRoute = this._router._routerMicrolib.getRoute(pivotRouteName) as Route;
```
rwjblue, line 505: Why do we have to `.getRoute` here?

packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -474,6 +476,39 @@ export default class RouterService extends Service {
...
+     * it will refresh only that route and its descendents.
+     * Returns a promise that will be resolved once the refresh is complete.
+     * All resetController, beforeModel, model, afterModel, redirect, and setupController
+     * hooks will be called again. You will get new data from the model hook.
+     *
+     * @method refresh
+     * @param {String} [pivotRouteName] the route to refresh (along with all child routes)
+     * @return Transition
+     * @public
+     */
```
rwjblue, line 493: Can you add `@category` here?

packages/ember/tests/routing/router_service_test/refresh_test.js
```diff
@@ -0,0 +1,100 @@
...
+    'Router Service - refresh',
+    class extends RouterTestCase {
+      async ['@test RouterService#refresh can be used to re-run the model hooks of active routes'](
+        assert
+      ) {
+        let parentCounter = 0;
+        this.add(
+          'route:parent',
+          Route.extend({
+            model() {
```
rwjblue, line 16: Let's prefer native class syntax here

packages/@ember/-internals/routing/lib/services/router.ts
rwjblue: Also, FWIW, `.getRoute` can return a promise (e.g. `ember-engines` or Embroider with lazy route bundles).

rwjblue (changes requested):
Thanks for working on this!!

packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -474,6 +476,39 @@ export default class RouterService extends Service {
...
+    /**
+     * Refreshes all currently active routes, doing a full transition.
+     * If a route name is provided and refers to a currently active route,
+     * it will refresh only that route and its descendents.
+     * Returns a promise that will be resolved once the refresh is complete.
+     * All resetController, beforeModel, model, afterModel, redirect, and setupController
+     * hooks will be called again. You will get new data from the model hook.
+     *
+     * @method refresh
+     * @param {String} [pivotRouteName] the route to refresh (along with all child routes)
```
rwjblue, line 490: I think you are right, using `routeName` here seems better. I personally think that internally referring to it as a "pivot route" is good (like the internal local variable name) to infer the  "the specified route and its descendents" thing without having to say it everywhere.

packages/@ember/-internals/routing/lib/services/router.ts
Windvis: I'm not sure what that means exactly. Do you mean like [this](https://github.com/emberjs/ember.js/blob/b97e2019e86b3c203554b0b5d16107add90e8e34/packages/%40ember/component/template-only.ts#L28)?

It looks like the feature flag key, so `@category EMBER_ROUTING_ROUTER_REFRESH` in this case?

packages/@ember/-internals/routing/lib/services/router.ts
Windvis: So, does that mean I change it back to `routeName`? :smile:

packages/@ember/-internals/routing/lib/services/router.ts
Windvis: > Why do we have to `.getRoute` here?

This comes back to [this earlier discussion](https://github.com/emberjs/ember.js/pull/19471#discussion_r597990897) where I questioned that as well. I did it like this because I had no idea how else to do it with the methods that are available on the router / microlib. And it seemed like it had logic related to engine routes as well. What's the proper way to do it exactly? Should I just look up the route in the container? Do I need to do anything specific for engine routes?
> Also, FWIW, .getRoute can return a promise (e.g. ember-engines or Embroider with lazy route bundles).

Yea, TS did indeed warn me about that :smile:. But the [`getRoute` implementation](https://github.com/emberjs/ember.js/blob/22bfcfdac0aeefcf333fb2d6697772934201b43b/packages/%40ember/-internals/routing/lib/system/router.ts#L190) in the Router had a `Route` return type, so I assumed it was ok.

rwjblue: Ya, but ember-engines (and Embroider) override that method so that it sometimes returns a promise.

packages/@ember/-internals/routing/lib/services/router.ts
rwjblue: Yes, exactly!

packages/@ember/-internals/routing/lib/services/router.ts
rwjblue: Haha, sorry yes. I'd like to document the param as `routeName` here.

packages/@ember/-internals/routing/lib/services/router.ts
Windvis: Thanks for the info, that clears it up!

> What's the proper way to do it exactly? Should I just look up the route in the container? Do I need to do anything specific for engine routes?

I just retrieve the route instance through the container now. Let me know if I should do it differently.

rwjblue (approved):
Sorry for the delay here @Windvis, this looks great!

PR #19472, opened 2021-03-22
[BUGFIX LTS] Prevent transformation of block params called `attrs`

Windvis:
This prevents the `transform-attrs-into-args` transform plugin from transforming block params that are named "attrs" and are yielded by components.

The following snippet would render nothing before this change even though a value was yielded from the component template:
```hbs
<Component as |attrs|>{{attrs}}</Component>
```


Windvis:
Naming things is not my strongest skill and English isn't my native tongue, so do let me know if I need to change / rewrite some sentences somewhere!

rwjblue (approved):
Awesome, thank you!!!

PR #19473, opened 2021-03-23
[BUGFIX release] Exposes getTemplateLocals utility from Glimmer VM

pzuraq:
Exposes the `getTemplateLocals` utility from Glimmer VM so that it can
be used for strict mode template compilation.


packages/ember-template-compiler/index.ts
```diff
@@ -37,7 +37,7 @@ export {
 export { RESOLUTION_MODE_TRANSFORMS, STRICT_MODE_TRANSFORMS } from './lib/plugins/index';
 export { EmberPrecompileOptions } from './lib/types';
 
-export { preprocess as _preprocess, print as _print } from '@glimmer/syntax';
+export { preprocess as _preprocess, print as _print, getTemplateLocals } from '@glimmer/syntax';
```
rwjblue, line 40: All of glimmer syntax is exported (just below), let's just rely on that export for now.

rwjblue (changes requested):

rwjblue (approved):

Issue #19459, opened 2021-03-15
[Bug] Assertion in `link-component` is incorrectly running when using `link-to-external`


hmajoros:
### 🐞 Describe the Bug
[This change](https://github.com/emberjs/ember.js/commit/56af48bb413f89766e0b355ca94c029e62695667#diff-4ef2173974821b22319e3588d4b1fbba945c84e9997aad54e722787802fbdaa9R493) was added to provide better boundaries for linking inside/between engines with the `<LinkTo>` component. However, `link-to-external` extends `link-component`, and thus this assertion runs on the `init` of any `LinkToExternal` usage. When inside of a routeless engine, there is currently no way to link out, as both `<LinkTo>` and `<LinkToExternal>` throw this assertion.

### 🔬 Minimal Reproduction
Using `ember-engines`, create a routeless engine & use `<LinkToExternal>` inside this routeless engine

### 😕 Actual Behavior
Assertion fails when using link-to-external inside of a routeless engine

### 🤔 Expected Behavior
routeless engines should allow the usage of link-to-externala

### 🌍 Environment

- Ember: 3.24.3
- Node.js/npm: 14.15.1
- OS: Big Sur
- Browser: All

hmajoros:
@chancancode I think the best way to fix this is to move this assertion check into a method inside `link-component`, and then have `link-to-external` overwrite this method with a noop. Thoughts?

rwjblue:
@hmajoros - Ya, that seems ok to me (I really do want to get rid of `<LinkToExternal />` all together, that is just a massively different problem 😝 ). Happy to land + backport a fix for this.

hmajoros:
ember.js PR: https://github.com/emberjs/ember.js/pull/19477
ember-engines PR: https://github.com/ember-engines/ember-engines/pull/761

hmajoros:
Closing issue as both ember & ember-engines fixes have now landed & been backported to 3.24.

PR #19477, opened 2021-03-25
[BUGFIX LTS] move LinkTo assertion into a method so LinkToExternal can override

hmajoros:
Part one of a fix for #19459. Once this lands I'll push a change in ember-engines to override this method to just be a noop.

First affected ember version is 3.24

Existing tests continue to pass.


rwjblue (changes requested):
Awesome, thanks for picking this up!!

What was the first affected Ember version? If it was 3.24, can you update the commit message and PR title to be prefixed with `[BUGFIX lts]`?

rwjblue (approved):

rwjblue:
Thank you!

Issue #19299, opened 2020-12-07
[Documentation] Route loading event code snippet seems misleading


nickeidler:
### 📙 Describe the Issue
In the API documentation for the Route `loading` event, there is a code snippet that shows use of the loading event to populate state on a controller. However, while working with Ember I've found that the controller is not always defined during execution of the loading action on the route. Thus, the code snippet seems a little misleading to me, since it implies that the controller will always be instantiated (otherwise this snippet would occasionally throw an exception). Should the snippet be amended to guard against this case?

### 🔬 Minimal Reproduction
See the code snippet referenced in https://api.emberjs.com/ember/3.22/classes/Route/events/loading?anchor=loading

### ➕ Additional Context
I ran into this issue while implementing similar functionality. The only difference is in my case we mutate the corresponding `controller` to the current `route`. e.g. I am in route `Foo`'s `loading` action mutating a property on controller `Foo`. I found via aggregate error logging that sometimes the controller `Foo` would be `undefined` in this hook. This surprised me, since I was writing code similar to what is sanctioned in the docs.

nickeidler:
@chriskrycho FYI

PR #19485, opened 2021-03-29
[DOC] Update loading action code example

nickeidler:
Added a guard against `undefined` controller in the `loading()` action documentation. Also updated syntax from using `set()` to directly setting values. See issue #19299 for context.

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/route.ts
+++ b/packages/@ember/-internals/routing/lib/system/route.ts
@@ -500,7 +500,7 @@ class Route extends EmberObject implements IRoute {
     export default class ApplicationRoute extends Route {
       @action
       loading(transition, route) {
-        const controller = this.controllerFor('foo');
+        let controller = this.controllerFor('foo');
 
         // The controller may not be instantiated when initially loading
         if (controller) {
--- a/packages/@ember/-internals/routing/lib/system/route.ts
+++ b/packages/@ember/-internals/routing/lib/system/route.ts
@@ -502,6 +502,7 @@ class Route extends EmberObject implements IRoute {
       loading(transition, route) {
         const controller = this.controllerFor('foo');
 
+        // The controller may not be instantiated when initially loading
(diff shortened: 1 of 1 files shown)


packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -500,12 +500,15 @@ class Route extends EmberObject implements IRoute {
...
       @action
       loading(transition, route) {
-        let controller = this.controllerFor('foo');
-        controller.set('currentlyLoading', true);
+        const controller = this.controllerFor('foo');
 
-        transition.finally(function() {
-          controller.set('currentlyLoading', false);
-        });
+        if (controller) {
```
acorncom, line 505: @nickeidler makes sense that controllers aren't always instantiated now that I stop to think about it, but it doesn't seem immediately apparent. Might be worth a code comment above this line explaining that briefly. Something like:

```suggestion
        // controller may not be instantiated when initially loading
        if (controller) {
```

nickeidler: Good point-- added a comment.

acorncom (approved):
LGTM. @locks any additional thoughts?

packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -500,12 +500,16 @@ class Route extends EmberObject implements IRoute {
     export default class ApplicationRoute extends Route {
       @action
       loading(transition, route) {
-        let controller = this.controllerFor('foo');
-        controller.set('currentlyLoading', true);
+        const controller = this.controllerFor('foo');
```
acorncom, line 503: @nickeidler I'd actually expect that we'll want this to go back to a `let`, Ember tends to default to a liberal let approach (https://madhatted.com/2016/1/25/let-it-be)

locks:
Let's get it merged. Thanks!

PR #19487, opened 2021-03-30
[BUGFIX beta] Allow passing a string literal to {{helper}} and {{modifier}}

chancancode:
As per [RFC #432](https://github.com/emberjs/rfcs/blob/master/text/0432-contextual-helpers.md#invoking-contextual-modifiers)

However, dynamic string values are currently disallowed per earlier framework team weekly meeting discussion to not make it harder for Embroider to heuristically analyze the dynamic usages.

Eventually we will want to do the same for components as well, and this AST transform would work there too.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/helpers/-disallow-dynamic-resolution.ts
+++ b/packages/@ember/-internals/glimmer/lib/helpers/-disallow-dynamic-resolution.ts
@@ -1,0 +1,57 @@
+/**
+@module ember
+*/
+import { assert } from '@ember/debug';
+import { CapturedArguments } from '@glimmer/interfaces';
+import { createComputeRef, valueForRef } from '@glimmer/reference';
+import { internalHelper } from './internal-helper';
+
+export default internalHelper(({ positional, named }: CapturedArguments) => {
+  assert(
+    `[BUG] wrong number of positional arguments, expecting 1, got ${positional.length}`,
+    positional.length === 1
+  );
+
+  let nameOrValueRef = positional[0];
+
+  assert(`[BUG] expecting \`type\` named argument`, 'type' in named);
+  assert(`[BUG] expecting \`loc\` named argument`, 'loc' in named);
+  assert(`[BUG] expecting \`original\` named argument`, 'original' in named);
+
+  let typeRef = named.type;
+  let locRef = named.loc;
+  let originalRef = named.original;
+
+  // Bug: why do these fail?
+  // assert('[BUG] expecting a string literal for the `type` argument', isConstRef(typeRef));
+  // assert('[BUG] expecting a string literal for the `loc` argument', isConstRef(locRef));
+  // assert('[BUG] expecting a string literal for the `original` argument', isConstRef(originalRef));
+
+  const type = valueForRef(typeRef);
+  const loc = valueForRef(locRef);
+  const original = valueForRef(originalRef);
+
+  assert('[BUG] expecting a string literal for the `type` argument', typeof type === 'string');
+  assert('[BUG] expecting a string literal for the `loc` argument', typeof loc === 'string');
+  assert(
+    '[BUG] expecting a string literal for the `original` argument',
+    typeof original === 'string'
+  );
+
+  return createComputeRef(() => {
+    let nameOrValue = valueForRef(nameOrValueRef);
+
+    assert(
+      `Passing a dynamic string to the \`(${type})\` keyword is disallowed. ` +
+        `(You specified \`(${type} ${original})\` and \`${original}\` evaluated into "${nameOrValue}".) ` +
+        `This ensures we can statically analyze the template and determine which ${type}s are used. ` +
+        `If the ${type} name is always the same, use a string literal instead, i.e. \`(${type} "${nameOrValue}")\`. ` +
+        `Otherwise, import the ${type}s into JavaScript and pass them to the ${type} keyword. ` +
+        'See https://github.com/emberjs/rfcs/blob/master/text/0496-handlebars-strict-mode.md#4-no-dynamic-resolution for details. ' +
+        loc,
+      typeof nameOrValue !== 'string'
+    );
+
+    return nameOrValue;
+  });
+});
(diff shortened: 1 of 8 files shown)


packages/@ember/-internals/glimmer/lib/helpers/-disallow-dynamic-resolution.ts
```diff
@@ -0,0 +1,57 @@
...
+  assert(`[BUG] expecting \`original\` named argument`, 'original' in named);
+
+  let typeRef = named.type;
+  let locRef = named.loc;
+  let originalRef = named.original;
+
+  // Bug: why do these fail?
+  // assert('[BUG] expecting a string literal for the `type` argument', isConstRef(typeRef));
+  // assert('[BUG] expecting a string literal for the `loc` argument', isConstRef(locRef));
+  // assert('[BUG] expecting a string literal for the `original` argument', isConstRef(originalRef));
```
chancancode, line 28: @pzuraq I don't think this has to block merging but it seems odd that we lost the const reference propagation here:

```hbs
{{helper (-disallow-dynamic-resolution this.helperName type="helper" loc=" (some-file.hbs:1:2) " original="this.helperName") "helper args"}}
                                                            ~~~~~~~~     ~~~~~~~~~~~~~~~~~~~~~~~          ~~~~~~~~~~~~~~~~~
```

It seemed strange that `isConstRef` returned false on this. The logic in `isConstRef` checks for `ref.tag === CONST_TAG` but `ref.tag` is `null` here. Seems like a bug?

packages/@ember/-internals/glimmer/lib/helpers/-resolve.ts
```diff
@@ -0,0 +1,38 @@
...
+@module ember
+*/
+import { Owner } from '@ember/-internals/owner';
+import { assert, runInDebug } from '@ember/debug';
+import { CapturedArguments } from '@glimmer/interfaces';
+import { createConstRef, isConstRef, valueForRef } from '@glimmer/reference';
+import { internalHelper } from './internal-helper';
+
+export default internalHelper(({ positional }: CapturedArguments, owner: Owner | undefined) => {
+  // why is this allowed to be undefined in the first place?
```
chancancode, line 11: @pzuraq This seemed strange. I suspect this is leftover from when you thought you didn't have to propagate owner in strict mode and left the `| undefined` in the signature, even though in practice it shouldn't ever be `undefined`?

packages/@ember/-internals/glimmer/tests/integration/custom-modifier-manager-test.js
```diff
@@ -581,6 +581,41 @@ moduleFor(
...
 
+    '@feature(EMBER_DYNAMIC_HELPERS_AND_MODIFIERS) Can resolve a modifier'() {
+      this.registerModifier(
+        'replace',
+        defineSimpleModifier((element, [text]) => (element.innerHTML = text ?? 'Hello, world!'))
+      );
+
+      // BUG: this should work according to the RFC
+      // this.render(
+      //   '[<div {{modifier "replace"}}>Nope</div>][<div {{modifier (modifier "replace") "wow"}}>Nope</div>]'
```
chancancode, line 592: cc @pzuraq for upstream bug that we should fix before release (currently this gives an error message saying `modifier` keyword cannot be used as a modifier)

chancancode: See https://github.com/emberjs/rfcs/blob/master/text/0432-contextual-helpers.md#invoking-contextual-modifiers

packages/@ember/-internals/glimmer/tests/integration/custom-modifier-manager-test.js
```diff
@@ -581,6 +581,41 @@ moduleFor(
...
+      this.registerModifier(
+        'replace',
+        defineSimpleModifier((element) => (element.innerHTML = 'Hello, world!'))
+      );
+
+      expectAssertion(
+        () =>
+          this.render(
+            // BUG: this should work according to the RFC
+            // '<div {{modifier this.name}}>Nope</div>',
```
chancancode, line 611: Same as above

packages/@ember/-internals/glimmer/lib/resolver.ts
```diff
@@ -181,12 +183,27 @@ const BUILTIN_KEYWORD_HELPERS = {
...
+  // Bug: this may be a quirk of our test setup?
+  // In prod builds, this is a no-op helper and is unused in practice. We shouldn't need
+  // to add it at all, but the current test build doesn't produce a "prod compiler", so
+  // we ended up running the debug-build for the template compliler in prod tests. Once
+  // that is fixed, this can be removed. For now, this allows the test to work and does
+  // not really harm anything, since it's just a no-op pass-through helper and the bytes
+  // has to be included anyway. In the future, perhaps we can avoid the latter by using
+  // `import(...)`?
+  BUILTIN_KEYWORD_HELPERS['-disallow-dynamic-resolution'] = disallowDynamicResolution;
+}
```
chancancode, line 205: cc @rwjblue though it probably doesn't have to block merging

packages/@ember/-internals/glimmer/lib/helpers/-resolve.ts
```diff
@@ -0,0 +1,38 @@
...
+  let fullNameRef = positional[0];
+
+  assert('[BUG] expecting a string literal as argument', isConstRef(fullNameRef));
+
+  const fullName = valueForRef(fullNameRef);
+
+  assert('[BUG] expecting a string literal as argument', typeof fullName === 'string');
+  assert('[BUG] expecting a valid full name', fullName.split(':').length === 2);
+
+  runInDebug(() => {
```
rwjblue, line 28: Seems odd to use runInDebug here (we don't do the same stripping for runInDebug as we do for the other style; ultimately we should deprecate runInDebug), can you swap to `if (DEBUG) {` instead?

packages/@ember/-internals/glimmer/lib/resolver.ts
rwjblue: Ya we should fix, I agree. But it doesn't _really_ matter for landing.

rwjblue (approved):

PR #19488, opened 2021-03-30
[DOC beta] Added page-title.ts to helpers and  expected docs

Angel-gc:
I'm not sure if the short sentence is enough of a description for the page-title addon, but I tried to keep it concise.


Angel-gc:
trailing whitespace error

PR #19491, opened 2021-04-06
fix: lookup/register and singleton flag behavior

runspired:
I noticed this evening that `getOwner(this).lookup('thing:foo', { singleton: false })` was no longer working as documented here https://api.emberjs.com/ember/3.25/classes/ApplicationInstance/methods/lookup?anchor=lookup

As best as I can tell, this API has been broken for at least 5 years. Given the long length of time this has gone unnoticed, we may simply desire to remove this functionality entirely and not restore it. Users have an alternative using `factoryFor`.


rwjblue (commented):
This definitely seems like you are right (and this PR fixes it). It seems odd that no one noticed the issue though.

PR #19499, opened 2021-04-20
[BUGFIX beta] [DEPRECATION] Deprecate `@foo={{helper}}`

chancancode:
Per [RFC 496](https://github.com/emberjs/rfcs/blob/master/text/0496-handlebars-strict-mode.md#3-no-implicit-invocation-of-argument-less-helpers), this commit deprecates the implicit invocation of argument-less helpers and requires explicit parenthesis if:

1. Non-strict mode, AND
2. `helper` is a free variable (not `this.helper`, `@helper` or
   a local variable), AND
3. No arguments are provided to the helper, AND
4. It's in a component invocation's named argument position (not
   `<div id={{helper}}>` or `<Foo bar={{helper}}>`), AND
5. Not parenthesized (not `@foo={{(helper)}}`), AND
6. Not interpolated (not `@foo="{{helper}}"`).

The cases in real apps where all of the above are true should be quite rare, as most helpers take at least one argument.

This is probably not the most efficient/minimal changes to the compiler infrastructure, but I chose to optimize for making the change easier to remove/rollback (i.e. adding new nodes instead of changing existing ones) as we won't be needing this for very long.

This also bumps `@glimmer/*` to latest which also brings in the bugfix commit https://github.com/glimmerjs/glimmer-vm/pull/1293.


packages/@ember/-internals/glimmer/tests/integration/helpers/custom-helper-test.js
```diff
@@ -909,4 +923,95 @@ if (DEBUG) {
...
+        );
+
+        this.assertText('[false][Hello, world!]');
+        this.assertStableRerender();
+      }
+
+      // TODO: this one really should work, and there is a passing test in glimmer-vm,
+      // but somehow it doesn't work here. This is almost certainly a VM bug as something
+      // is trying to call `block.compile()` but `block` is the reference for `this.foo`.
+      // So the execution stack is probably off-by-one or something.
```
chancancode, line 990: cc @pzuraq, this seems mildly concerning, though unlikely to be introduced here and so don't seem like it would need to block merging. The test is ported from https://github.com/glimmerjs/glimmer-vm/blob/a19ce398de64077a52ebfae9f4878cec49a089c8/packages/%40glimmer/integration-tests/test/syntax/argument-less-helper-paren-less-invoke-test.ts#L66-L77, which does pass there.

chancancode: Actually – when I revert to `@glimmer/*` to 0.77.5, this particular test does pass, so there does appear to be a regression on 0.78.0. Do you think it is possible that https://github.com/glimmerjs/glimmer-vm/pull/1293 introduced  a stack bug here? I'm not sure why the same test on the vm side is not catching this, but perhaps it's a strict mode vs loose more thing?

chancancode:
Hmm. Not sure what’s up with the failures. Did I do something wrong when I publish the packages? Cc @rwjblue

PR #19500, opened 2021-04-20
Change Typings for `setupController`

stephencweiss:
As `_transition` is not used, it should not be required. I'd prefer to take it out completely, but in that case, I'm not sure if there are going to be other ramifications. Happy to investigate that if there's appetite.

Also, happy to be told I'm totally off base here!

Cheers!

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/route.ts
+++ b/packages/@ember/-internals/routing/lib/system/route.ts
@@ -1339,7 +1339,7 @@ class Route extends EmberObject implements IRoute {
     @since 1.0.0
     @public
   */
-  setupController(controller: Controller, context: {}, _transition: Transition) {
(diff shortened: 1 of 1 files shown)


stephencweiss:
A little additional context. I was recently working on a project where I upgraded the types for routing from `3.0.9` to `3.16.13` and started getting errors. 
Screen_Shot_2021-04-20_at_4 00 18_PM

The fix is easy enough, but looking into it, it seems like it's totally unnecessary... Also the [docs](https://api.emberjs.com/ember/release/classes/Route/methods/setupController?anchor=setupController) don't say anything about `transition` (likely because it's unused) -- so, rather than making a breaking change by taking `_transition` _out_ of the API, making it optional seems like a happy medium.

stephencweiss:
Closing in favor of #19502

PR #19502, opened 2021-04-21
make transition optional on setupController [19501]

### LINKED ISSUE #19501: [Documentation] Transition is passed into the setup controller but unused
### 📙 Describe the Issue

In the [Setup Controller](https://github.com/emberjs/ember.js/blob/master/packages/@ember/-internals/routing/lib/system/route.ts#L1335-L1347), we require `transition`, however it's never used. 

This is confusing, particularly as the [documentation](https://api.emberjs.com/ember/release/classes/Route/methods/setupController?anchor=setupController) notes that `transition` is part of the API, but doesn't include it in the example.

### 🔬 Minimal Reproduction
Please, share a link to the page with the documentation and share steps to reproduce.

```js
import Route from '@ember/routing/route';

export default class Auth extends Route {
  setupController(controller, model) {
    super.setupController(controller, model);
  }
}
```

### 😕 Actual Behavior
Screen_Shot_2021-04-20_at_4 00 18_PM

### 🤔 Expected Behavior
Not passing `transition` should not result in a build time error.

stephencweiss:
Closes #19501.

19501 notes that transition parameter was marked as required
on the API for the setupController (and would fail for Typescript
projects if not present).

This change makes transition optional as it is unused by
setupController and should therefore not break applications which
do not pass it in.

Removing the parameter would potentially break backwards
compatibility. As a compromise, it's marked as optional only.

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/route.ts
+++ b/packages/@ember/-internals/routing/lib/system/route.ts
@@ -1335,11 +1335,11 @@ class Route extends EmberObject implements IRoute {
     @method setupController
     @param {Controller} controller instance
     @param {Object} model
-    @param {Object} transition
+    @param {Transition} [transition]
     @since 1.0.0
     @public
   */
-  setupController(controller: Controller, context: {}, _transition: Transition) {
+  setupController(controller: Controller, context: {}, _transition?: Transition) {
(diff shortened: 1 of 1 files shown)


mixonic:
@stephencweiss Thank you!

PR #19504, opened 2021-04-23
Perf: Bugfix: Don't read before write for anything with a descriptor

runspired:
Mostly opening this to see what tests fail. Currently when using `set` or `EmberObject.set` ember will first read a prop before writing to it unless that property happens to have a `COMPUTED_SETTER` which is only true for things that utilized `tracked` or `computed`. This causes performance issues and bugs when read before write is not expected.

### DIFF
--- a/packages/@ember/-internals/metal/lib/property_set.ts
+++ b/packages/@ember/-internals/metal/lib/property_set.ts
@@ -74,7 +74,7 @@ export function set<T = unknown>(obj: object, keyName: string, value: T, toleran
 export function _setProp(obj: object, keyName: string, value: any) {
   let descriptor = lookupDescriptor(obj, keyName);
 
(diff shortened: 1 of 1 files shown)


rwjblue (commented):
Hmm, is this still something to move forward @runspired?

runspired:
It's something I feel we should change but it doesn't seem we have a clear path to changing it. Wouldn't might the thoughts of @pzuraq

kategengler:

runspired:
its relevant until starbeam, though I'm not sure what the path forward is. @NullVoxPopuli may have a better idea he's been poking around those parts

NullVoxPopuli:
I don't know what the impact of this PR is, as I have no clue about the older get/set patterns anymore. (I jumped on tracked when 3.13 was canary).

Can you rebase?

PR #19514, opened 2021-04-28
Remove unnecessary reopens

wagenet:


packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -2752,16 +2740,7 @@ if (ROUTER_EVENTS) {
     },
   };
 
-  Route.reopen(ROUTER_EVENT_DEPRECATIONS, {
-    _paramsFor(routeName: string, params: {}) {
```
wagenet, line 2756: Since this was in the `ROUTER_EVENTS` conditional it actually would have broken things in the future.

wagenet:
Looks like you actually need to use reopen if you want people to be able to use reopen themselves at a later point.

PR #19517, opened 2021-04-28
Improve internal types

wagenet:
Improve types for @ember/object

Includes:
  - Object, { action } from '@ember/object'
  - CoreObject from '@ember/object/core'
  - Observable from '@ember/object/observable'
  - { dependentKeyCompat } from '@ember/object/compat'

Note that types for `extend`, `reopen`, and `reopenClass` are not
exported. Also, `get` and `set` have naive types. All of these will
be deprecated in future releases.

Many types were copied from DT. We should make sure to give appropriate credit, but I suspect we may already be doing that.

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/route.ts
+++ b/packages/@ember/-internals/routing/lib/system/route.ts
@@ -96,19 +96,20 @@ export function hasDefaultSerialize(route: Route): boolean {
   @public
 */
 
-class Route extends EmberObject implements IRoute {
+class Route extends EmberObject.extend(ActionHandler, Evented) implements IRoute, Evented {
   static isRouteFactory = true;
 
-  routeName!: string;
-  fullRouteName!: string;
   context: {} = {};
-  controller!: Controller;
   currentModel: unknown;
 
   _bucketCache!: BucketCache;
   _internalName!: string;
-  _names: unknown;
+
+  private _names: unknown;
+
   _router!: EmberRouter;
+  _topLevelViewTemplate!: any;
+  _environment!: any;
 
   constructor(owner: Owner) {
     super(...arguments);
@@ -129,6 +130,13 @@ class Route extends EmberObject implements IRoute {
     }
   }
 
+  // Implement Evented
+  on!: (name: string, method: ((...args: any[]) => void) | string) => this;
+  one!: (name: string, method: string | ((...args: any[]) => void)) => this;
+  trigger!: (name: string, ...args: any[]) => any;
+  off!: (name: string, method: string | ((...args: any[]) => void)) => this;
+  has!: (name: string) => boolean;
+
   serialize!: (
     model: {},
     params: string[]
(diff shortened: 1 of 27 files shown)


packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -2752,16 +2803,7 @@ if (ROUTER_EVENTS) {
     },
   };
 
-  Route.reopen(ROUTER_EVENT_DEPRECATIONS, {
-    _paramsFor(routeName: string, params: {}) {
```
wagenet, line 2756: I'm pretty sure it was a bug for this to be in a conditional reopen. We always depend on `_paramsFor`.

packages/@ember/-internals/runtime/types/observable.d.ts
```diff
@@ -0,0 +1,81 @@
+import { ObserverMethod } from './-private/types';
+
+/**
+ * This mixin provides properties and property observing functionality, core features of the Ember object model.
+ */
```
chriskrycho, line 5: A discussion point (cc. @pzuraq @dfreeman @jamescdavis): throughout, this uses the `keyof this` to make keys support both auto-complete and type-checking against property keys. However, it *drops* the feature from DT where it returns the type. I *think*, if we drop support for building types mixins, we can actually enable `strictFunctionTypes` and keep the return type correct?

packages/@ember/object/type-tests/compat.test.ts
```diff
@@ -0,0 +1,4 @@
+import { dependentKeyCompat } from '@ember/object/compat';
+import { expectTypeOf } from 'expect-type';
+
+expectTypeOf(dependentKeyCompat).toMatchTypeOf<Function>();
```
chriskrycho, line 4:
```suggestion
expectTypeOf(dependentKeyCompat).toMatchTypeOf<PropertyDecorator>();
```

chriskrycho (commented):
Thanks for tackling this! One note: *if possible* I think it'd be extremely helpful to separate the functional changes from the type-only changes. In particular, I'm extremely leery of any shuffling in the router here (once bitten twice shy, and the router has bitten about four months off of my life 😂).

packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -127,6 +130,13 @@ class Route extends EmberObject implements IRoute {
     }
   }
 
+  // Implement Evented
+  on!: (name: string, method: ((...args: any[]) => void) | string) => this;
+  one!: (name: string, method: string | ((...args: any[]) => void)) => this;
+  trigger!: (name: string, ...args: any[]) => any;
```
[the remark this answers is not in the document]
wagenet: I copied this from evented.d.ts :/

chriskrycho (approved):
Thanks for these!

.eslintrc.js
```diff
@@ -35,7 +35,7 @@ module.exports = {
 
   overrides: [
     {
-      files: ['*.ts'],
+      files: ['*.ts', '*.d.ts'],
```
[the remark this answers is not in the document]
wagenet: Yes, it should. I think this must have been from an attempt to fix something I fixed in another way.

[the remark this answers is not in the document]
wagenet: Fixed.

packages/@ember/object/index.d.ts
```diff
@@ -1 +1,17 @@
...
+//   defineProperty,
+//   get,
+//   getProperties,
+//   getWithDefault,
+//   notifyPropertyChange,
+//   observer,
+//   set,
+//   setProperties,
+//   trySet,
+// } from '@ember/-internals/metal';
```
rwjblue, line 15: Why is this here? In general, I think we'd prefer to avoid landing a commented out section like this (basically we should bring it back, delete it, or add a comment explaining why its commented).

rwjblue (approved):

packages/@ember/object/index.d.ts
wagenet: It's stuff that we want to export but isn't ready yet. I'll remove it and add it back properly in another PR.

stefanpenner:
@wagenet these files appear to unexpectedly end up in the build output, which seems unintentional.

stefanpenner:
^^ fix landed, all should be resolved.

PR #19527, opened 2021-05-03
[CLEANUP] Remove Globals Resolver

btecu:


rwjblue:

```

git fetch origin
git rebase origin/v4-cleanup

# fix any conflicts

# push the rebase
git push -f
```

</details>

PR #19529, opened 2021-05-03
[CLEANUP] Remove deprecated mouse events

btecu:


packages/@ember/-internals/views/lib/system/event_dispatcher.js
```diff
@@ -311,7 +302,7 @@ export default EmberObject.extend({
       // Special handling of events that don't bubble (event delegation does not work).
       // Mimics the way this is handled in jQuery,
       // see https://github.com/jquery/jquery/blob/899c56f6ada26821e8af12d9f35fa039100e838e/src/event.js#L666-L700
-      if (MOUSE_ENTER_LEAVE_MOVE_EVENTS && EVENT_MAP[event] !== undefined) {
+      if (EVENT_MAP[event] !== undefined) {
```
simonihmig, line 305: You can delete the whole if-block here, as that does the special handling of these non-bubbling events, and only leave the else-block instead. The (now removed) `MOUSE_ENTER_LEAVE_MOVE_EVENTS` feature flag can effectively be seen as being always `false`, so in that way this if-block is basically dead code. Also the `EVENT_MAP` itself can be removed.

simonihmig (commented):
Merging #19227 has caused some conflicts here, sorry about that! :wink:

One test is failing in CI, I haven't really been able to identify it by looking through the logs though...

packages/@ember/-internals/views/lib/system/event_dispatcher.js
btecu: Thanks @simonihmig, I've removed the entire `if` block.

rwjblue:

```

git fetch origin
git rebase origin/v4-cleanup

# fix any conflicts

# push the rebase
git push -f
```

</details>

mixonic:
@btecu this PR was auto-closed when v4-cleanup was merged to master. Can you re-open for us to review/merge against master? This is excellent work and I don't want to lose it.

PR #19530, opened 2021-05-05
[DOC release] fix passing params to named blocks examples

chrisgame:
I tried following the docs when first trying to pass parameters to named blocks.
Currently the docs state `{{yield to="title" @person.name}}` as being valid but this gave me an error.
I think the `to=` needs to be the last argument so the examples should be `{{yield @person.name  to="title"}}`


locks (approved):
I believe you are correct! Thank you for taking notice and helping us correct the problem :)

PR #19536, opened 2021-05-06
fix computed import deprecation import path

knownasilya:
This is the current message:

```
deprecate.js:136 DEPRECATION: Using `computed.oneWay` has been deprecated. Instead, import the value directly from @ember/object/computed:

  import { oneWay } from '@ember/runloop'; 
```

Notice how it should not be `runloop`


rwjblue (approved):
Eeeck! Thank you for fixing this @knownasilya!

PR #19540, opened 2021-05-11
[BUGFIX release] Ensure ember-testing is loaded lazily

### LINKED ISSUE #19538: [Bug] Could not find module `ember-testing` imported from `@ember/test/index` on ember 3.27.0
### 🐞 Describe the Bug
Updating ember app to v3.27.0 causing an error: Could not find module `ember-testing` imported from `@ember/test/index`

### 🔬 Minimal Reproduction
Create an app with latest `ember-cli`. Update `ember-source` to v3.27.0
Build the app with `npm run build`. Go to dist folder. Serve the folder. In my case, I'm using `python -m "http.server"`
You'll see the error message as: Uncaught Error: Could not find module `ember-testing` imported from `@ember/test/index`

I'm seeing the same error with my production app.

Here is the sample [repo](https://github.com/sinankeskin/ember327).

### 😕 Actual Behavior
App successfully built but when I run and reach to main route I'm seeing this error and app doesn't work.

### 🤔 Expected Behavior
App should be working.

### 🌍 Environment

- Ember: - 3.26.1
- Node.js/npm: - v14.16.1/7.6.3
- OS: - Windows 10
- Browser: - Microsoft Edge  90.0.818.51 x64

pzuraq:
The recently modules API update means we are now loading real modules,
not polyfills based on the global. This means that the modules
themselves are _eagerly required_, rather than being references to a
value on the global. For example, previously, this:

```js
import { registerWaiter } from '@ember/test';

if (someCondition) {
  registerWaiter(() => {});
}
```

Would become this:

```js
if (someCondition) {
  Ember.Test.registerWaiter(() => {});
}
```

In either example, `registerWaiter` may or may not be called based on
the state of `someCondition`. However, in the second case, if
`Ember.Test` is not defined at all, it's completely ok as long as
`someCondition` is `false`. It's never called, so we never get an error
telling us `Ember.Test` is undefined.

Without the transform, the module is eagerly required, along with all of
its dependencies. If no one included `ember-testing`, then that means
it will throw an error immediately.

This PR makes the `@ember/test` module load `ember-testing` lazily, and
if it's not available (e.g. in a production environment) it replaces the
values with a function that throws a helpful error.

Fixes #19538, https://github.com/babel/ember-cli-babel/issues/406

### DIFF
--- a/packages/@ember/test/index.js
+++ b/packages/@ember/test/index.js
@@ -1,11 +1,27 @@
-import { Test } from 'ember-testing';
+import require, { has } from 'require';
 
-const {
-  registerAsyncHelper,
-  registerHelper,
-  registerWaiter,
-  unregisterHelper,
-  unregisterWaiter,
-} = Test;
+export let registerAsyncHelper;
+export let registerHelper;
+export let registerWaiter;
+export let unregisterHelper;
+export let unregisterWaiter;
 
-export { registerAsyncHelper, registerHelper, registerWaiter, unregisterHelper, unregisterWaiter };
+if (has('ember-testing')) {
+  let { Test } = require('ember-testing');
+
+  registerAsyncHelper = Test.registerAsyncHelper;
+  registerHelper = Test.registerHelper;
+  registerWaiter = Test.registerWaiter;
+  unregisterHelper = Test.unregisterHelper;
+  unregisterWaiter = Test.unregisterWaiter;
+} else {
+  let testingNotAvailableMessage = () => {
+    throw new Error('Attempted to use test utilities, but `ember-testing` was not included');
+  };
+
+  registerAsyncHelper = testingNotAvailableMessage;
+  registerHelper = testingNotAvailableMessage;
+  registerWaiter = testingNotAvailableMessage;
+  unregisterHelper = testingNotAvailableMessage;
+  unregisterWaiter = testingNotAvailableMessage;
+}


packages/@ember/test/index.js
```diff
@@ -1,11 +1,27 @@
...
+  let { Test } = require('ember-testing');
+
+  registerAsyncHelper = Test.registerAsyncHelper;
+  registerHelper = Test.registerHelper;
+  registerWaiter = Test.registerWaiter;
+  unregisterHelper = Test.unregisterHelper;
+  unregisterWaiter = Test.unregisterWaiter;
+} else {
+  let testingNotAvailableMessage = () => {
+    throw new Error('Attempted to use test utilities, but `ember-testing` was not included');
```
runspired, line 19: probably this should include a note that the user may want to wrap their usage in `if (DEBUG) {}` or in `if (Ember.testing) {}`, otherwise lgtm.

runspired (approved):

chriskrycho:
Starting some digging: I'm seeing what appears to be a related error on 3.28.0-beta.6. It's not the *same* issue, but appears similar: it appeared for us on 3.27.0 and continued through 3.27.3 and up into the 3.28 beta series. I will likely open a new issue and linking it to this one tomorrow as I start chasing it down!

chriskrycho:
Starting some digging: I'm seeing what appears to be a related error on 3.28.0-beta.6. It's not the *same* issue, but appears similar: it appeared for us on 3.27.0 and continued through 3.27.3 and up into the 3.28 beta series. I will likely open a new issue and linking it to this one tomorrow as I start chasing it down!

PR #19543, opened 2021-05-12
[BUGFIX lts] Backport {{hash}} laziness

pzuraq:
Backports the fixes from https://github.com/emberjs/ember.js/pull/19511

The internal `_query` property of LinkTo had to be made into an autotracking
dependency using dependentKeyCompat in order to handle the fact
that hash never updates now. This was not an issue on master since
LinkTo has been refactored there.


rwjblue:
@pzuraq - Where are we at with this?

pzuraq:
The PR needs to be updated to add the additional fixes that landed recently. I was planning on updating it once we landed those in release and were pretty sure it wouldn't need any more changes.

chriskrycho:
Status check here—IIRC, @pzuraq @rwjblue we ended up deciding this *shouldn’t* be back ported, because of the various issues that showed up? And in any case, the period in which we *would* have back ported it to 3.24 is loooong past. I think we should go ahead and close this.

kategengler:
At this point, we're clearly not backporting.

PR #19547, opened 2021-05-14
[FEATURE cached] Adds cached decorator

pzuraq:
Adds the cache decorator specified in RFC 566: https://github.com/emberjs/rfcs/blob/master/text/0566-memo-decorator.md

Also moves the `@tracked` API docs to be in a more relevant place, and removes the old private API docs which duplicated that information.


rwjblue (changes requested):
This looks good overall, just needs feature flagging then should be good to go!

rwjblue:
@NullVoxPopuli - Can you pick this up (basically making a new PR to copy over the docs that @pzuraq wrote)?

NullVoxPopuli:
can do!

PR #19551, opened 2021-05-24
[CLEANUP] Remove deprecated jQuery integration

simonihmig:
This removes the `Ember.$()` and `this.$()` jQuery integration, deprecated by RFC386.

There is still more work to do, but I wanted to keep the scope of PRs small:
* remove EventDispatcher jQuery code - hopefully we can get #19227 merged first to not again cause major merge conflicts. @rwjblue 😬
* further internal cleanup

Note: as a breaking change, this should only get merged when it's clear that current `master` represents the "4.0 train" - not 100% sure if that decision has been taken already!?


rwjblue:
will likely need a rebase?

simonihmig:
> will likely need a rebase?

I had to fix a linting error, but other than that this should be ok!

simonihmig:
Done

mixonic (approved):

mixonic:
I'm not clear on where we sit with the `v4-cleanup` stuff now. `master` is currently 4.0, so I would land this if it was pointed there.

mixonic:
@simonihmig this PR was auto-closed when v4-cleanup was merged to master. Can you re-open for us to review/merge against master? This is excellent work and I don't want to lose it.

Issue #17228, opened 2018-11-24
Event.target not set for mouseEnter/Leave w/o jQuery in IE11


simonihmig:
While working on #17227, and having somehow figured out how to run the (ES5-transpiled, jQuery-less) tests in the browser (used `/tests/index.html?ie=true&dist=prod&skipPackage=container%2Cember-testing%2C%40ember%2Fdebug&moduleId=4a92ac35&jquery=none`), I found that a couple of tests are failing in IE11 due to `Event.target` being `null` when using the special handling of mouseEnter/Leave when jQuery is not used. 

It's only those [assertions for `Event.target`](https://github.com/emberjs/ember.js/blob/master/packages/%40ember/-internals/glimmer/tests/integration/event-dispatcher-test.js#L179), the actual events get fired correctly. See here:
<img width="990" alt="bildschirmfoto 2018-11-24 um 02 09 38" src="https://user-images.githubusercontent.com/1325249/48963375-59c24380-ef91-11e8-8f40-6f01cdfe3630.png">

This brings up two questions:

1. why weren't these failing in CI? I *guess* the browserstack tests do not run the no-jquery tests, but not sure. @rwjblue can you confirm? 

2. how to fix this? The event target is set here for the fake event used in the special handling of mouseEnter/Leave (introduced in #16999): https://github.com/emberjs/ember.js/blob/master/packages/%40ember/-internals/views/lib/system/event_dispatcher.js#L326
It seems for IE11 this is a readonly, non-configurable property, so neither directly setting it nor using `defineProperty()` (as it is now) works for IE. And of course we cannot use a native Proxy. 🤔

rwjblue:
> why weren't these failing in CI? I guess the browserstack tests do not run the no-jquery tests, but not sure. @rwjblue can you confirm?

Yes, exactly. We only run IE11 with jQuery.

pixelhandler:
@rwjblue would this be a fair statement, "Ember w/o jQuery is only supported on evergreen browsers, and Safari, so IE11 is must use jQuery" ?

simonihmig:
FWIW, for one of our apps, we dropped jQuery long *before* we dropped IE11 support. And AFAIK the other tests run fine w/o jQuery (would need to be confirmed), only this rather edge-casey thing makes some problems...

rwjblue:
I believe this is resolved, @simonihmig can you confirm and close if you agree?

simonihmig:
[...]

PR #19552, opened 2021-05-24
[CLEANUP] Remove jQuery integration in EventDispatcher

simonihmig:
Second part after #19551 to remove deprecations of RFC386.


simonihmig:
Ok, so there are 4 failings tests for IE11. These are all related to the also deprecated moveEnter/Leave events. They fail due to the inability to override `Event.target` in IE11, which has been discussed in https://github.com/emberjs/ember.js/issues/17228. As was also confirmed in that issue, tests in CI were only passing for IE11 because they *only* ran in jQuery integration mode. Now with that removed, they start to fail in CI. 

Again, we cannot fix this behavior in IE11, and as these events are deprecated anyway, it seems unreasonable to spend more time on fixing/disabling those tests here. The removal of support for these events if already underway in #19529 (/cc @btecu), so I would suggest we try to land this first, and then rebase this PR which should then make CI happy.

pzuraq:
We probably should have a separate PR removing IE11 from CI. Once we start landing breaking changes on canary, we won't need to support IE11 anymore (as it's deprecated in 4.0), so no need to test against it.

simonihmig:
> Once we start landing breaking changes on canary

I wondered about this, so *when* actually is this going to happen? The package.json in `master` is saying this is already 4.0.0-alpha. But seems this is not set in stone? Is there still a Core team decision pending?

pzuraq:
@simonihmig yeah I'm trying to figure this out now. I think what could possibly happen is, we start landing the cleanup/fixes for 4.0 on canary and move forward. If we do end up needing to have another 1-2 releases after 3.28, we could re-release the 3.28 LTS and continue supporting it on a different branch. We've done this before when a beta was particularly unstable and we had to push it back, so I think we could do something similar here if we needed to, but I want to get consensus from the rest of the team about that.

rwjblue:

```

git fetch origin
git rebase origin/v4-cleanup

# fix any conflicts

# push the rebase
git push -f
```

</details>

simonihmig:
Done

rwjblue:
OK, I think we need another rebase to drop IE11 from the browser stack setup and then this should be good, right?

simonihmig:
@rwjblue yep, did that, green now!

mixonic:
@simonihmig this PR was auto-closed when v4-cleanup was merged to master. Can you re-open for us to review/merge against master? This is excellent work and I don't want to lose it.

PR #19555, opened 2021-05-25
[BUGFIX release] Improve class based tranform deprecation message

chancancode:
This was quite difficult to track down since it doesn't tell you _which_ plugin is causing the issue. It only reports the name of the template which is not actually relevant.

### DIFF
--- a/packages/ember-template-compiler/lib/system/compile-options.ts
+++ b/packages/ember-template-compiler/lib/system/compile-options.ts
@@ -82,11 +82,16 @@ interface LegacyPlugin {
 
 export type LegacyPluginClass = new (env: ASTPluginEnvironment) => LegacyPlugin;
 
-function wrapLegacyPluginIfNeeded(_plugin: PluginFunc | LegacyPluginClass): PluginFunc {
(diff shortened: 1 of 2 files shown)


packages/ember-template-compiler/lib/system/compile-options.ts
```diff
@@ -102,28 +107,29 @@ function wrapLegacyPluginIfNeeded(_plugin: PluginFunc | LegacyPluginClass): Plug
       let pluginInstantiated = false;
 
       return {
-        name: _plugin.constructor && _plugin.constructor.name,
+        name: plugin.name,
```
chancancode, line 110: Pretty sure this was a big previously: if `plugin` is a class, then `plugin.constructor` would have been `Function`, so this would always be the string `"Function"`.

chancancode: (Hopefully the name wasn't used to deduplicate plugins!)

wagenet (approved):

rwjblue (approved):

PR #19557, opened 2021-05-25
[BUGFIX release] Refine Ember Global deprecation message

chancancode:
Example message:

Example message:

```
Usage of the Ember Global is deprecated. You should import the Ember module or the specific API instead.

See https://deprecations.emberjs.com/v3.x/#toc_ember-global for details.

Usages of the Ember Global may be caused by an outdated ember-cli-babel dependency. The following steps may help:

* Upgrade your `devDependencies` on `ember-cli-babel` to `^7.26.6`.
* Upgrade the following addons to the latest version:
  * active-model-adapter
  * ember-animated
  * ember-async-await-helper
  * ember-attacher
  * ember-cli-showdown
  * ember-freestyle
  * ember-md5

### Important ###

In order to avoid repeatedly showing the same deprecation messages, no further deprecation messages will be shown for usages of the Ember Global until ember-cli-babel is upgraded to v7.26.6 or above.

To see all instances of this deprecation message at runtime, set the `EMBER_GLOBAL_DEPRECATIONS` environment variable to "all", e.g. `EMBER_GLOBAL_DEPRECATIONS=all ember test`.

### Details ###

Prior to v7.26.6, ember-cli-babel sometimes transpiled imports into the equivalent Ember Global API, potentially triggering this deprecation message even when you did not directly reference the Ember Global.

The following outdated versions are found in your project:

* ember-cli-babel@6.11.0, currently used by:
  * ember-md5@6.11.0
    * Depends on ember-cli-babel@6.11.0

* ember-cli-babel@6.18.0, currently used by:
  * active-model-adapter@6.18.0
    * Depends on ember-cli-babel@^6.8.2
  * ember-angle-bracket-invocation-polyfill@6.18.0
    * Depends on ember-cli-babel@^6.17.0
    * Added by ember-animated@0.11.0
  * ember-async-await-helper@6.18.0
    * Depends on ember-cli-babel@^6.16.0
  * ember-cli-deploy-bugsnag@6.18.0 (Dormant)
    * Depends on ember-cli-babel@^6.16.0
  * ember-cli-showdown@6.18.0
    * Depends on ember-cli-babel@^6.8.1
  * ember-inflector@6.18.0
    * Depends on ember-cli-babel@^6.0.0
    * Added by active-model-adapter@2.2.0
  * ember-maybe-import-regenerator@6.18.0 (Dormant)
    * Depends on ember-cli-babel@^6.0.0-beta.4
    * Added by ember-animated@0.11.0
  * ember-named-arguments-polyfill@6.18.0 (Dormant)
    * Depends on ember-cli-babel@^6.6.0
    * Added by ember-animated@0.11.0
  * ember-raf-scheduler@6.18.0
    * Depends on ember-cli-babel@^6.6.0
    * Added by ember-attacher@1.2.3 > ember-popper@0.11.3
  * ember-runtime-enumerable-includes-polyfill@6.18.0 (Dormant)
    * Depends on ember-cli-babel@^6.9.0
    * Added by ember-freestyle@0.12.11

* ember-cli-babel@7.26.5, currently used by:
  * @embroider/macros@7.26.5 (Compatible)
    * Depends on ember-cli-babel@^7.23.0
    * Added by ember-cli-mirage@2.2.0
    * Added by ember-exam@6.1.0
  * direwolf (your app)
    * Depends on ember-cli-babel@7.26.5
  * ember-attacher@7.26.5 (Compatible)
    * Depends on ember-cli-babel@^7.23.0
  * ember-fetch@7.26.5 (Compatible)
    * Depends on ember-cli-babel@^7.23.0
  * ember-freestyle@7.26.5 (Compatible)
    * Depends on ember-cli-babel@^7.23.0
  * ember-inflector@7.26.5 (Compatible)
    * Depends on ember-cli-babel@^7.23.0
    * Added by ember-data@3.26.0
    * Added by ember-cli-mirage@2.2.0
  * ember-source@7.26.5 (Dormant)
    * Depends on ember-cli-babel@^7.23.0
  * qunit-dom@7.26.5 (Dormant)
    * Depends on ember-cli-babel@^7.23.0

Note: Addons marked as "Dormant" does not appear to have any JavaScript files. Therefore, even if they are using an old version ember-cli-babel, they are unlikely to be the cuplrit of this deprecation and can likely be ignored.

Note: Addons marked as "Compatible" are already compatible with ember-cli-babel@7.26.6. Try upgrading your `devDependencies` on `ember-cli-babel` to `^7.26.6`.

```

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -10,6 +10,7 @@ const buildStripClassCallcheckPlugin = require('./build-strip-class-callcheck-pl
 const injectBabelHelpers = require('./transforms/inject-babel-helpers').injectBabelHelpers;
 const debugTree = require('broccoli-debug').buildDebugCallback('ember-source:addon');
 const vmBabelPlugins = require('@glimmer/vm-babel-plugins');
+const semver = require('semver');
 
 const PRE_BUILT_TARGETS = [
   'last 1 Chrome versions',
@@ -40,6 +41,17 @@ add(
   path.join(__dirname, '..', 'dist', 'ember-template-compiler.js')
 );
 
+function* walkAddonTree(project, pathToAddon = []) {
+  for (let addon of project.addons) {
+    yield [addon, pathToAddon];
+    yield* walkAddonTree(addon, [...pathToAddon, `${addon.name}@${addon.pkg.version}`]);
+  }
+}
+
+function requirementFor(pkg, deps = {}) {
+  return deps[pkg];
+}
+
 module.exports = {
   init() {
     this._super.init && this._super.init.apply(this, arguments);
@@ -64,10 +76,13 @@ module.exports = {
   name: 'ember-source',
   paths,
   absolutePaths,
+  _bootstrapEmber: "require('@ember/-internals/bootstrap').default();",
   _jqueryIntegrationEnabled: true,
 
   included() {
     this._super.included.apply(this, arguments);
+
+    this._issueGlobalsDeprecation();
 
     const { has } = require('@ember/edition-utils');
 
@@ -244,7 +259,7 @@ module.exports = {
     return new MergeTrees([
       concatBundle(emberFiles, {
         outputFile: 'ember.js',
-        footer: "require('@ember/-internals/bootstrap');",
+        footer: this._bootstrapEmber,
       }),
 
       concatBundle(emberTestingFiles, {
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -309,4 +324,195 @@ module.exports = {
...
+          'no further deprecation messages will be shown for usages of the Ember Global ' +
+          'until ember-cli-babel is upgraded to v7.26.6 or above.\n\n' +
+          'To see all instances of this deprecation message at runtime, ' +
+          'set the `EMBER_GLOBAL_DEPRECATIONS` environment variable to "all", ' +
+          'e.g. `EMBER_GLOBAL_DEPRECATIONS=all ember test`.\n';
+      }
+
+      message +=
+        '\n### Details ###\n\n' +
+        'Prior to v7.26.6, ember-cli-babel sometimes transpile imports into the equivilant Ember Global API, ' +
```
runspired, line 449:
```suggestion
        'Prior to v7.26.6, ember-cli-babel sometimes transpiles imports into the equivilant Ember Global API, ' +
```

lib/index.js
```diff
@@ -309,4 +324,236 @@ module.exports = {
 
     return debugTree(new MergeTrees([ember, templateCompiler, jquery]), 'vendor:final');
   },
+
+  _issueGlobalsDeprecation() {
+    if (process.env.EMBER_ENV === 'production') {
+      return;
```
rwjblue, line 330: Don't we still need to set the bootstrap string?

chancancode: There is a default (same as the current one) on the prototype (?)

lib/index.js
```diff
@@ -64,11 +76,14 @@ module.exports = {
   name: 'ember-source',
   paths,
   absolutePaths,
+  _bootstrapEmber: "require('@ember/-internals/bootstrap').default();",
```
chancancode, line 79: @rwjblue

rwjblue: Ah, gotcha sorry

stefanpenner:
This may have broken embroider builds: https://github.com/embroider-build/embroider/issues/833

stefanpenner:
The contract between Embroider & ideal Add-ons is that they are statically consumable. This change, takes a step backwards from that approach. As it requires the code which invokes the "bootstrapping" code to be generated dynamically at build time. Each such dynamic injection of code must either be avoided, or replicated both here and in embroider which seems like an anti pattern.

lib/index.js
```diff
@@ -309,4 +324,236 @@ module.exports = {
...
+        return undefined;
+      }
+    })(this.project.root);
+
+    let groupedByTopLevelAddon = Object.create(null);
+    let groupedByVersion = Object.create(null);
+    let projectInfo;
+
+    for (let [addon, pathToAddon] of walkAddonTree(this.project)) {
+      let version = addon.pkg.version;
```
stefanpenner, line 347: We may  (speculation, and untested) want to skip the current iteration if the `pathToAddon` has already appeared. Some large applications, especially those using engines can have a very LARGE number of add-ons that must be considered (10s or 100s of thousands), and it is often the case that these extra traversals can cause CLI boot-time regressions.

But if the `pathToAddon` has already been considered it is safe to use the previous result, and skip the upcoming work.

lib/index.js
```diff
@@ -309,4 +324,236 @@ module.exports = {
 
     return debugTree(new MergeTrees([ember, templateCompiler, jquery]), 'vendor:final');
   },
+
+  _issueGlobalsDeprecation() {
```
stefanpenner, line 328: This function looks quite complex, but I don't see any tests ensuring it is both accurate and remains accurate overtime. Are these present elsewhere or is this an oversight?

stefanpenner:
Reading this further given that https://github.com/emberjs/ember.js/blob/master/lib/index.js#L305-L317 exists, do these deprecations run for folks lucky enough to get a prebuilt version of ember? Or do we simply create the warning for them, but never present it?

stefanpenner:
Speaking with rob, it would appear the following might be the best course of action:

1) restore this bootstrap code to be an IFFE (good default)
2) if the above detection detects the problematic case (improper duplicate versions of ember-cli babel) overlay a new IFFE file with the appropriate error message and merge it in.

This is viable because embroider versions aren't effected by this issue, nor are peoples applications which aren't detected to be invalid, also folks with prebuilt versions of ember don't benefit from the newly built bootstrap file.

PR #19558, opened 2021-05-25
[CLEANUP] Remove IE11 support

simonihmig:
Remove IE11 deprecation warnings and tests, as per RFC 685.

Removing the transpilation for IE11 unveiled some mismatch of other transpilation settings vs. the testing environments:
* Safari tests were failing, as babel settings were "last 2 Safari versions" (which according to latest caniuse data is 14.1 + 14), but BrowserStack was configured for Safari 12.1. Changed this to 14.
* node tests were as well failing with some syntax errors. Added a `node: 'current'` setting to targets, but only when running node tests, to keep browser tests untouched.

### DIFF
--- a/config/browserlists.js
+++ b/config/browserlists.js
@@ -1,7 +1,7 @@
 const allSupportedBrowsers = [
   'last 2 Chrome versions',
   'last 2 Firefox versions',
-  'last 2 Safari versions',
+  'Safari 12',
   'last 2 Edge versions',
 ];
 
--- a/config/browserlists.js
+++ b/config/browserlists.js
@@ -3,7 +3,6 @@ const allSupportedBrowsers = [
   'last 2 Firefox versions',
   'last 2 Safari versions',
   'last 2 Edge versions',
-  'ie 11',
 ];
 
 const modernBrowsers = [
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -183,6 +183,7 @@ jobs:
         env:
           DISABLE_SOURCE_MAPS: true
           BROCCOLI_ENV: production
+          SHOULD_TRANSPILE_FOR_NODE: true
         run: yarn ember build -prod
       - name: test
         run: yarn test:node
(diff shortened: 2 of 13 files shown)


simonihmig:
Hm, the CI UI seems to be somehow broken here, indicating that there is a pending "Browserstack Tests (Safari, Edge, IE11)" job, which actually does not exist (has been renamed, and the renamed one exists and is green). When you click on the "Checks" tab, all jobs are listed as expected and green. 🤔

simonihmig:
Btw, there are probably one or two more iterations possible here, to remove remaining IE11 cruft, like workarounds for missing APIs. But I wanted to keep the scope here to the most essential changes.

config/browserlists.js
```diff
@@ -3,7 +3,6 @@ const allSupportedBrowsers = [
   'last 2 Firefox versions',
   'last 2 Safari versions',
   'last 2 Edge versions',
-  'ie 11',
 ];
```
simonihmig, line 6: Not sure if we actually need to keep these two separate lists, as they are mostly the same now (with Edge basically being Chrome)?

SergeAstapov:
@simonihmig per [PRF #685](https://github.com/emberjs/rfcs/blob/master/text/0685-new-browser-support-policy.md#non-evergreen-browsers) (and [Ember.js Browser Support Policy](https://emberjs.com/browser-support/) ultimately) Safari 12 *is* supported by Ember.js.

I wonder if IC still should be configured for Safari 12?

simonihmig:
> However, per PRF #685 (and Ember.js Browser Support Policy ultimately) Safari 12 is supported by Ember.js.

Oh, right, good catch! Thanks!

I did not intend to implement that RFC fully here, more to focus on getting rid of IE for unblocking further refactorings and other semi-related PRs (#19552). But at least we shouldn't cause regressions in test coverage here.

So I reverted the BrowserStack config change, and added `Safari 12` to targets, which now also makes the Safari tests pass again!

*Uh, blueprint tests are failing somehow, but that seems unrelated*

rwjblue:

```

git fetch origin
git rebase origin/v4-cleanup

# fix any conflicts

# push the rebase
git push -f
```

</details>

simonihmig:
Done

rwjblue (approved):

PR #19560, opened 2021-05-26
[BUGFIX release] Improve computed.* and run.* deprecation message

chancancode:
Show details about outdated ember-cli-babel and only show the first instance by default.

Example message:

```
Using `computed.reads` has been deprecated. Instead, import the value directly from @ember/object/computed:

import { reads } from '@ember/object/computed';

These usages may be caused by an outdated ember-cli-babel dependency. The following steps may help:

* Upgrade the following addons to the latest version:
  * active-model-adapter
  * ember-animated
  * ember-async-await-helper
  * ember-attacher
  * ember-cli-showdown
  * ember-md5

### Important ###

In order to avoid repeatedly showing the same deprecation messages, no further deprecation messages will be shown for theses deprecated usages until ember-cli-babel is upgraded to v7.26.6 or above.

To see all instances of this deprecation message, set the `EMBER_RUNLOOP_AND_COMPUTED_DOT_ACCESS_DEPRECATIONS` environment variable to "all", e.g. `EMBER_RUNLOOP_AND_COMPUTED_DOT_ACCESS_DEPRECATIONS=all ember test`.

### Details ###

Prior to v7.26.6, ember-cli-babel sometimes transpiled imports into the equivalent Ember Global API, potentially triggering this deprecation message indirectly, even when you did not observe these deprecated usages in your code.

The following outdated versions are found in your project:

* ember-cli-babel@6.11.0, currently used by:
  * ember-md5@6.11.0
    * Depends on ember-cli-babel@6.11.0

* ember-cli-babel@6.18.0, currently used by:
  * active-model-adapter@6.18.0
    * Depends on ember-cli-babel@^6.8.2
  * ember-angle-bracket-invocation-polyfill@6.18.0
    * Depends on ember-cli-babel@^6.17.0
    * Added by ember-animated@0.11.0
  * ember-async-await-helper@6.18.0
    * Depends on ember-cli-babel@^6.16.0
  * ember-cli-deploy-bugsnag@6.18.0 (Dormant)
    * Depends on ember-cli-babel@^6.16.0
  * ember-cli-showdown@6.18.0
    * Depends on ember-cli-babel@^6.8.1
  * ember-inflector@6.18.0
    * Depends on ember-cli-babel@^6.0.0
    * Added by active-model-adapter@2.2.0
  * ember-maybe-import-regenerator@6.18.0 (Dormant)
    * Depends on ember-cli-babel@^6.0.0-beta.4
    * Added by ember-animated@0.11.0
  * ember-named-arguments-polyfill@6.18.0 (Dormant)
    * Depends on ember-cli-babel@^6.6.0
    * Added by ember-animated@0.11.0
  * ember-raf-scheduler@6.18.0
    * Depends on ember-cli-babel@^6.6.0
    * Added by ember-attacher@1.2.3 > ember-popper@0.11.3
  * ember-runtime-enumerable-includes-polyfill@6.18.0 (Dormant)
    * Depends on ember-cli-babel@^6.9.0
    * Added by ember-freestyle@0.12.11

Note: Addons marked as "Dormant" does not appear to have any JavaScript files. Therefore, even if they are using an old version ember-cli-babel, they are unlikely to be the cuplrit of this deprecation and can likely be ignored.
```

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -415,101 +415,86 @@ module.exports = {
       }
     }
 
-    let message =
-      'Usage of the Ember Global is deprecated. ' +
-      'You should import the Ember module or the specific API instead.\n\n' +
-      'See https://deprecations.emberjs.com/v3.x/#toc_ember-global for details.\n\n' +
-      'Usages of the Ember Global may be caused by an outdated ember-cli-babel dependency. ' +
-      'The following steps may help:\n\n';
+    let suggestions = 'The following steps may help:\n\n';
 
-    let hasActionableSteps = false;
+    let hasActionableSuggestions = false;
 
     if (projectInfo) {
-      message += '* Upgrade your `devDependencies` on `ember-cli-babel` to `^7.26.6`.\n';
-      hasActionableSteps = true;
+      suggestions += '* Upgrade your `devDependencies` on `ember-cli-babel` to `^7.26.6`.\n';
+      hasActionableSuggestions = true;
     } else if (compatibleTopLevelAddons.length > 0) {
       // Only show the compatible addons if the project itself is up-to-date, because updating the
       // project's own dependency on ember-cli-babel to latest may also get these addons to use it
       // as well. Otherwise, there is an unnecessary copy in the tree and it needs to be deduped.
       if (isYarnProject === true) {
-        message +=
+        suggestions +=
           '* Run `npx yarn-deduplicate --packages ember-cli-babel` followed by `yarn install`.\n';
       } else if (isYarnProject === false) {
-        message += '* Run `npm dedupe`.\n';
+        suggestions += '* Run `npm dedupe`.\n';
       } else {
-        message +=
+        suggestions +=
           '* If using yarn, run `npx yarn-deduplicate --packages ember-cli-babel` followed by `yarn install`.\n' +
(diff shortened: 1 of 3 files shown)


rwjblue (approved):

chancancode:
@rwjblue the first CI run failed on my PR on the IE 11 build, I thought the reason was that the bootstrap ("footer") code is just concatenated and not transpiled, so I started converting my code to avoid arrows and stuff

but now the build is green and I am a bit confused. I am done with the conversion anyway, should I push?

I made some minor cosmetic fixes to the message, so I would have to figure out what I need to reapply if we are keeping it as-is

chancancode:
I pushed the extra commit into a different branch in case you think it is necessary: https://github.com/emberjs/ember.js/commit/72c3fd71609b4b07502def8fe103b98d69912b7d

The concat step (where the footer is inserted) happens after `transpileTree`, which is why I thought this would be needed, however the tests now passes without it so 🤷🏼‍♂️

rwjblue:
@chancancode - Ya, I think we should transpile. Though it is fairly unlikely that folks are using IE11 for development, it should be possible.

chancancode:
I’m still confused about why the test wasn’t failing, but yeah seems good to me, opened #19564

PR #19566, opened 2021-05-27
Finish future public types for @ember/object

wagenet:
* [x] import EmberObject from '@ember/object'; | Ember.Object **(without extend, reopen, and reopenClass; with limited get/set)**
* [x] import { action } from '@ember/object'; | Ember._action
* [DEPRECATED] ~import { aliasMethod } from '@ember/object'; | Ember.aliasMethod~ 
* [x] import { computed } from '@ember/object'; | Ember.computed
* [x] import { defineProperty } from '@ember/object'; | Ember.defineProperty
* [x] import { get } from '@ember/object'; | Ember.get **(always returns unknown)**
* [x] import { getProperties } from '@ember/object'; | Ember.getProperties
* [DEPRECATED] ~import { getWithDefault } from '@ember/object'; | Ember.getWithDefault~
* [UNDOCUMENTED] ~import { notifyPropertyChange } from '@ember/object'; | Ember.notifyPropertyChange~
* [x] import { observer } from '@ember/object'; | Ember.observer
* [x] import { set } from '@ember/object'; | Ember.set **(allows anything to be set!)**
* [x] import { setProperties } from '@ember/object'; | Ember.setProperties **(allows anything to be set!)**
* [x] import { trySet } from '@ember/object'; | Ember.trySet
* [x] import { dependentKeyCompat } from '@ember/object/compat'; | Ember._dependentKeyCompat
* [ ] import ComputedProperty from '@ember/object/computed'; | Ember.ComputedProperty **(Unclear that we want or need to export this.)**
* [x] import { alias } from '@ember/object/computed'; | Ember.computed.alias
* [x] import { and } from '@ember/object/computed'; | Ember.computed.and **(decorator only)**
* [x] import { bool } from '@ember/object/computed'; | Ember.computed.bool **(decorator only)**
* [x] import { collect } from '@ember/object/computed'; | Ember.computed.collect **(decorator only)**
* [x] import { deprecatingAlias } from '@ember/object/computed'; | Ember.computed.deprecatingAlias **(decorator only)**
* [x] import { empty } from '@ember/object/computed'; | Ember.computed.empty **(decorator only)**
* [x] import { equal } from '@ember/object/computed'; | Ember.computed.equal **(decorator only)**
* [x] import { expandProperties } from '@ember/object/computed'; | Ember.expandProperties **(decorator only)**
* [x] import { filter } from '@ember/object/computed'; | Ember.computed.filter **(decorator only)**
* [x] import { filterBy } from '@ember/object/computed'; | Ember.computed.filterBy **(decorator only)**
* [x] import { filterProperty } from '@ember/object/computed'; | Ember.computed.filterProperty **(decorator only)**
* [x] import { gt } from '@ember/object/computed'; | Ember.computed.gt **(decorator only)**
* [x] import { gte } from '@ember/object/computed'; | Ember.computed.gte **(decorator only)**
* [x] import { intersect } from '@ember/object/computed'; | Ember.computed.intersect **(decorator only)**
* [x] import { lt } from '@ember/object/computed'; | Ember.computed.lt **(decorator only)**
* [x] import { lte } from '@ember/object/computed'; | Ember.computed.lte **(decorator only)**
* [x] import { map } from '@ember/object/computed'; | Ember.computed.map **(decorator only)**
* [x] import { mapBy } from '@ember/object/computed'; | Ember.computed.mapBy **(decorator only)**
* [x] import { mapProperty } from '@ember/object/computed'; | Ember.computed.mapProperty **(decorator only)**
* [x] import { match } from '@ember/object/computed'; | Ember.computed.match **(decorator only)**
* [x] import { max } from '@ember/object/computed'; | Ember.computed.max **(decorator only)**
* [x] import { min } from '@ember/object/computed'; | Ember.computed.min **(decorator only)**
* [x] import { none } from '@ember/object/computed'; | Ember.computed.none **(decorator only)**
* [x] import { not } from '@ember/object/computed'; | Ember.computed.not **(decorator only)**
* [x] import { notEmpty } from '@ember/object/computed'; | Ember.computed.notEmpty **(decorator only)**
* [x] import { oneWay } from '@ember/object/computed'; | Ember.computed.oneWay **(decorator only)**
* [x] import { or } from '@ember/object/computed'; | Ember.computed.or **(decorator only)**
* [x] import { readOnly } from '@ember/object/computed'; | Ember.computed.readOnly **(decorator only)**
* [x] import { reads } from '@ember/object/computed'; | Ember.computed.reads **(decorator only)**
* [x] import { setDiff } from '@ember/object/computed'; | Ember.computed.setDiff **(decorator only)**
* [x] import { sort } from '@ember/object/computed'; | Ember.computed.sort **(decorator only)**
* [x] import { sum } from '@ember/object/computed'; | Ember.computed.sum **(decorator only)**
* [x] import { union } from '@ember/object/computed'; | Ember.computed.union **(decorator only)**
* [x] import { uniq } from '@ember/object/computed'; | Ember.computed.uniq **(decorator only)**
* [x] import { uniqBy } from '@ember/object/computed'; | Ember.computed.uniqBy **(decorator only)**
* [x] import CoreObject from '@ember/object/core'; | Ember.CoreObject  **(without extend, reopen, and reopenClass)**
* [x] import Evented from '@ember/object/evented'; | Ember.Evented **(NEEDS TO BE REVISITED)**
* [x] import { on } from '@ember/object/evented'; | Ember.on
* [x] import { addListener } from '@ember/object/events'; | Ember.addListener **(NEEDS TO BE REVISITED)**
* [x] import { removeListener } from '@ember/object/events'; | Ember.removeListener **(NEEDS TO BE REVISITED)**
* [x] import { sendEvent } from '@ember/object/events'; | Ember.sendEvent **(NEEDS TO BE REVISITED)**
* [x] import { cacheFor } from '@ember/object/internals'; | Ember.cacheFor
* [DEPRECATED] ~import { copy } from '@ember/object/internals'; | Ember.copy~
* [x] import { guidFor } from '@ember/object/internals'; | Ember.guidFor
* [x] import Mixin from '@ember/object/mixin'; | Ember.Mixin
* [x] import Observable from '@ember/object/observable'; | Ember.Observable **(without Mixin)**
* [x] import { addObserver } from '@ember/object/observers'; | Ember.addObserver
* [x] import { removeObserver } from '@ember/object/observers'; | Ember.removeObserver
* [ ] import PromiseProxyMixin from '@ember/object/promise-proxy-mixin'; | Ember.PromiseProxyMixin **(unsure what to do with this)**
* [ ] import ObjectProxy from '@ember/object/proxy'; | Ember.ObjectProxy **(unsure what to do with this)**


packages/@ember/-internals/routing/lib/ext/controller.ts
```diff
@@ -169,7 +169,7 @@ ControllerMixin.reopen({
     deprecateTransitionMethods('controller', 'transitionToRoute');
 
     // target may be either another controller or a router
-    let target = get(this, 'target');
+    let target = get(this, 'target') as any;
```
chriskrycho, line 172: I was going to ask about typing this and then saw that it's a mixin and :sob:

Can we possibly do a double cast, `as any as (Route | Controller)`?

packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -1102,13 +1106,13 @@ class Route extends EmberObject.extend(ActionHandler, Evented) implements IRoute
     // Assign the route's controller so that it can more easily be
     // referenced in action handlers. Side effects. Side effects everywhere.
     if (!this.controller) {
-      let qp = get(this, '_qp');
-      let propNames = qp !== undefined ? get(qp, 'propertyNames') : [];
+      let qp = get(this, '_qp') as any;
```
chriskrycho, line 1109: Any chance we can *not* `as any` this one?

packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -2090,7 +2094,7 @@ class Route extends EmberObject.extend(ActionHandler, Evented) implements IRoute
     let controllerName = this.controllerName || this.routeName;
     let owner = getOwner(this);
     let controller = owner.lookup<Controller>(`controller:${controllerName}`);
-    let queryParameterConfiguraton = get(this, 'queryParams');
+    let queryParameterConfiguraton = get(this, 'queryParams') as this['queryParams'];
```
chriskrycho, line 2097: Huh. Is this a knock-on effect from having `get` not be well-typed? If so, maybe we right types which understand how to do key lookup for it for internal purposes, at least?

packages/@ember/-internals/runtime/lib/system/core_object.d.ts
```diff
@@ -1,15 +1,39 @@
-import { EmberClassConstructor, EmberInstanceArguments, Objectify } from './-private/types';
+/**
```
chriskrycho, line 1: This is also in the bucket of things I *think* we probably want to *not* add to core? 🤔 Though I do note you have it all as `@internal` and maybe it's useful to have internally but not publish. The trick will be that we probably also need to publish it in that secondary package, but having both may be a necessary evil for the transition period. cc. @jamescdavis @dfreeman for thoughts.

packages/@ember/object/type-tests/compat/dependent-key-compat.test.ts
```diff
@@ -0,0 +1,17 @@
+import { dependentKeyCompat } from '@ember/object/compat';
+import { expectTypeOf } from 'expect-type';
+
+expectTypeOf(dependentKeyCompat).toMatchTypeOf<PropertyDecorator>();
```
chriskrycho, line 4: Huh: why `toMatchTypeOf` rather than `toEqualTypeOf` here?

packages/@ember/object/type-tests/compat/dependent-key-compat.test.ts
```diff
@@ -0,0 +1,17 @@
...
+
+// Example (without irrelevant details) from https://api.emberjs.com/ember/3.18/functions/@ember%2Fobject%2Fcompat/dependentKeyCompat
+class Person {
+  firstName!: string;
+  lastName!: string;
```
chriskrycho, line 9:
```suggestion
  declare firstName: string;
  declare lastName: string;
```

packages/@ember/object/type-tests/computed.test.ts
```diff
@@ -0,0 +1,28 @@
...
+  get fullName() {
+    return `${this.firstName} ${this.lastName}`;
+  }
+
+  @(computed('firstName', 'lastName').readOnly())
+  get readonlyFullName() {
+    return `${this.firstName} ${this.lastName}`;
+  }
+
+  // NOTE: This should probably not work, but types do currently allow it.
```
chriskrycho, line 17: Iiiiiinteresting. I don't know that there's a way we could do otherwise, either. 🤔

packages/@ember/object/type-tests/computed.test.ts
```diff
@@ -0,0 +1,28 @@
+import { computed } from '@ember/object';
```
chriskrycho, line 1: Could probably add an `expectTypeOf` here to check for the return type of the various forms:

```ts
expectTypeOf(computed('foo')).toEqualTypeOf<PropertyDecorator>();
```

etc.—and the same for the CP macro type tests below.

packages/@ember/object/type-tests/evented/index.test.ts
```diff
@@ -0,0 +1,7 @@
+import { expectTypeOf } from 'expect-type';
+
+import Mixin from '@ember/object/mixin';
+import Evented from '@ember/object/evented';
+
+// A very naive test that at least makes sure we can import this
```
chriskrycho, line 6: This is good; any chance we can also pull over some of the type tests [from DT](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/ember__object/test/event.ts) here? Me, I want Evented burned to the ground, but if we’re shipping types for it here, we should cover the ground. 😩

packages/@ember/object/type-tests/mixin/index.test.ts
```diff
@@ -0,0 +1,9 @@
+import { expectTypeOf } from 'expect-type';
+
+import Mixin from '@ember/object/mixin';
+
+let newMixin = Mixin.create({
+  foo: 'bar',
+});
+
+expectTypeOf(newMixin).toMatchTypeOf<Mixin>();
```
chriskrycho, line 9: This is definitely an interesting case we'll need to think about in terms of the Classic support package when we do that. Supporting `.extend()` is one thing, but technically supporting `Mixin.create()` goes with it… which seems 😬 🤔 (This is totally fine and good for *this* PR, of course.)

packages/@ember/object/type-tests/set.test.ts
```diff
@@ -0,0 +1,14 @@
...
+
+import { expectTypeOf } from 'expect-type';
+
+let foo = { baz: 1 };
+
+expectTypeOf(set(foo, 'baz', 2)).toEqualTypeOf<number>();
+
+// FIXME: This is definitely a potential footgun. You're allowed to
+// change the type without any warning.
+expectTypeOf(set(foo, 'baz', 'aaaahhh')).toEqualTypeOf<string>();
```
chriskrycho, line 11: Yeah, this is why we originally required the `T[K]` type here. My current thinking is that maybe we *do* support that for standalone `get` and `set`, *not* out of the box but via the extension types? But with an eye to deprecating both post-Polaris. Hmm.

packages/@ember/object/type-tests/try-set.test.ts
```diff
@@ -0,0 +1,16 @@
+import { trySet } from '@ember/object';
+import { expectTypeOf } from 'expect-type';
+
+let obj = { name: 'Zoey' };
+
+let result = trySet(obj, 'contacts.twitter', '@emberjs');
+expectTypeOf(result).toEqualTypeOf<string>();
```
chriskrycho, line 7: Per my note above about this API:

```suggestion
expectTypeOf(result).toEqualTypeOf<string | undefined>();
```

chriskrycho (changes requested):
This is the overall right direction, I think; I had a few questions and comments throughout but I think *most* of it can land as is, and we can continue iterating. Thank you for doing this and sorry for the six month delay in getting back to it!

package.json
```diff
@@ -149,7 +149,10 @@
     "testem": "^3.1.0",
     "testem-failure-only-reporter": "^0.0.1",
     "tslint": "^5.20.1",
-    "typescript": "^4.2.4"
+    "typescript": "^4.3.2"
+  },
+  "resolutions": {
+    "typescript": "^4.3.2"
```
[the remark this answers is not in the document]
wagenet: I don't remember. It's gone now.

packages/@ember/-internals/meta/lib/meta.ts
```diff
@@ -86,22 +86,36 @@ type Listener = StringListener | FunctionListener;
 let currentListenerVersion = 1;
 
 export class Meta {
+  /** @internal */
```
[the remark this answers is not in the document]
wagenet: I believe this was for hoped use with api-extractor. Unfortunately, we're still blocked on the multiple Entry Points issue.

packages/@ember/-internals/metal/lib/computed.ts
```diff
@@ -38,14 +39,36 @@ import {
 } from './property_events';
 import { set } from './property_set';
 
-export type ComputedPropertyGetter = (keyName: string) => any;
-export type ComputedPropertySetter = (keyName: string, value: any, cachedValue?: any) => any;
+export type ComputedPropertyGetterFunction<T> = (this: any, key: string) => T;
```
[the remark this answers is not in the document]
wagenet: I think it was for internal type safety. I'll look into using unknown instead.

packages/@ember/-internals/metal/lib/computed.ts
```diff
@@ -1028,38 +1061,52 @@ class ComputedDecoratorImpl extends Function {
   @return {ComputedDecorator} property decorator instance
   @public
 */
-export function computed(target: object, key: string, desc: PropertyDescriptor): PropertyDescriptor;
-export function computed(...args: (string | ComputedPropertyConfig)[]): ComputedDecorator;
+// @computed without parens or computed with descriptor args
```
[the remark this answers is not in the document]
wagenet: The fact that these live in @ember/-internals makes me think I didn't intend for them to reach the public.

packages/@ember/-internals/meta/lib/meta.ts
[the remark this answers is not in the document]
wagenet: In general, everything in -internals is private except for the case where we intentionally re-export it in @ember/object. It turns out that Meta is one of the cases where we do re-export.

packages/@ember/-internals/metal/lib/computed.ts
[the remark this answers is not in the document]
wagenet: It doesn't look like these are re-exported in @ember/object so this should be considered entirely private.

packages/@ember/-internals/metal/lib/decorator.ts
```diff
@@ -13,9 +13,13 @@ export type Decorator = (
...
-export function isElementDescriptor(
-  args: any[]
-): args is [object, string, DecoratorPropertyDescriptor] {
+export type ElementDescriptor = [
+  target: object,
+  propertyName: string,
+  descriptor: DecoratorPropertyDescriptor
+];
+
+export function isElementDescriptor(args: any[]): args is ElementDescriptor {
```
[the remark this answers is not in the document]
wagenet: Looks like it's easily done.

packages/@ember/-internals/metal/lib/get_properties.ts
```diff
@@ -31,7 +31,12 @@ import { get } from './property_get';
   @return {Object}
   @public
 */
-export default function getProperties(obj: object, keys?: string[]): object {
+function getProperties<L extends string[]>(obj: object, list: L): { [Key in L[number]]: unknown };
+function getProperties<L extends string[]>(
+  obj: object,
+  ...list: L
+): { [Key in L[number]]: unknown };
+function getProperties(obj: object, keys?: string[]): object {
```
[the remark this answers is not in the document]
wagenet: Also easily done!

packages/@ember/-internals/metal/lib/libraries.ts
```diff
@@ -84,7 +84,7 @@ if (EMBER_LIBRARIES_ISREGISTERED) {
 if (DEBUG) {
   Libraries.prototype.logVersions = function (): void {
     let libs = this._registry;
-    let nameLengths = libs.map((item) => get(item, 'name.length'));
+    let nameLengths = libs.map((item) => get(item, 'name.length')) as number[];
```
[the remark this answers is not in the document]
wagenet: Switched to an assert.

packages/@ember/-internals/metal/lib/mixin.ts
```diff
@@ -71,12 +72,12 @@ function concatenatedMixinProperties(
   return concats;
 }
 
-function giveDecoratorSuper(
+function giveDecoratorSuper<Get, Set>(
```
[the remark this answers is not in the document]
wagenet: Done.

packages/@ember/-internals/metal/lib/properties.ts
```diff
@@ -59,7 +59,7 @@ import { revalidateObservers } from './observer';
 export function defineProperty(
   obj: object,
   keyName: string,
-  desc?: Decorator | undefined | null,
+  desc?: Decorator | PropertyDescriptor | undefined | null,
```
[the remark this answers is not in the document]
wagenet: I don't remember. If I remove it some tests fail.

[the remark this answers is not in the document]
wagenet: It allows this test to pass https://github.com/wagenet/ember.js/blob/types3/packages/@ember/object/type-tests/define-property.test.ts#L6

packages/@ember/-internals/routing/lib/ext/controller.ts
wagenet: I'll switch to an assert.

packages/@ember/-internals/routing/lib/system/route.ts
wagenet: Yup.

packages/@ember/-internals/runtime/lib/system/core_object.d.ts
wagenet: This is in the internals package still and I do think we'll want it here. Whether we want to re-export it in @ember/object/core is another question.

packages/@ember/object/type-tests/compat/dependent-key-compat.test.ts
wagenet: Because it can also return a Decorator.

packages/@ember/object/type-tests/computed.test.ts
wagenet: This is partially completed.

packages/@ember/object/type-tests/evented/index.test.ts
wagenet: I thought I had, but we can bring over more if I missed some.

packages/@ember/object/type-tests/set.test.ts
wagenet: Sounds like we should discuss this some more.

wagenet:
Looks like I have some test failures to resolve. Otherwise I feel pretty good about this.

chriskrycho:
Excellent. Let me know when those are green, and also perhaps take a look at the relevant changes [here](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/58060) to see if we're misaligned anywhere! (I'll try to do the same when I come back to this next week.)

packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -115,11 +116,18 @@ export default class RouterService extends Service {
...
+    routeNameOrURL: string,
+    ...modelsAndOptions: [...any[], Record<string, unknown>]
+  ): Transition;
+  transitionTo(routeNameOrURL: string, ...models: any[]): Transition;
+  transitionTo(...modelsAndOptions: [...any[], Record<string, unknown>]): Transition;
+  transitionTo(...models: any[]): Transition;
+  transitionTo(...args: (string | any | Record<string, unknown>)[]): Transition {
     if (resemblesURL(args[0])) {
       // NOTE: this `args[0] as string` cast is safe and TS correctly infers it
       // in 3.6+, so it can be removed when TS is upgraded.
```
chriskrycho, line 129: Here and on lines 340-342 below, the comments are now out of date. Maybe a fast follow so we can actually land this thing, though.

packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -334,7 +357,8 @@ class Route extends EmberObject.extend(ActionHandler, Evented) implements IRoute
       names = (routeInfo && routeInfo['_names']) || [];
     }
 
-    let qps = get(this, '_qp.qps');
+    // SAFETY: Since `_qp` is protected we can't infer the type
```
chriskrycho, line 360: 💙 for this comment and all those like it.

packages/@ember/-internals/routing/lib/services/router.ts
wagenet: I'll correct this but wait to push.

wagenet: Nevermind, I've corrected and pushed.

packages/@ember/object/type-tests/computed/expand-properties.test.ts
```diff
@@ -0,0 +1,13 @@
+import { expandProperties } from '@ember/object/computed';
```
chriskrycho, line 1: I learned about this last week working on the DT update. It's... such a bananas thing. It makes sense that we have it, but I won't be sad to see it gone someday.

chriskrycho (approved):
👏🏼 thank you *so* much for this. Huge work, and will make all the rest of our work on this over the months ahead waaaaay easier.

PR #19567, opened 2021-05-27
[BUGFIX lts] Ensure legacy LinkTo query params work with `hash`

pzuraq:
Fixes legacy LinkTo's query params that rely on `hash`. Now that `hash` never changes/updates directly, we need to use autotracking since we read all of the keys on the hash. The alternative would be to add observers to each key.


pzuraq:
We discussed this direction today and decided it is not an acceptable fix, as other computeds dependent on hashes in a similar way in the ecosystem could regress as well. We'll be working on an alternative fix.

PR #19583, opened 2021-06-01
[BUGFIX lts] Ensure hash objects correctly entangle as dependencies

pzuraq:
The previous bugfixes to `{{hash}}` caused a change to the semantics of
computed properties that depend on a hash. Specifically, because
`{{hash}}` objects are now proxies, they are _constant_, never updating
again after they are initially created. This is fine if you depend on
an individual key in a hash, but breaks if you depend directly on the
hash itself:

```js
computed('hash.foo', function() {}) // this works

computed('hash', function() {}) // this will no longer rerun
```

This is used occasionally when you wish to depend on the dynamic keys
of a dictionary, like so:

```js
computed('hash', function() {
  let values = [];

  for (let key in this.hash) {
    values.push(hash[key]);
  }

  return values;
})
```

Notably, this is not a problem with autotracking, because autotracking
will entangle the usage of these keys dynamically. So this is only a
problem with legacy systems such as `computed` and `observer` which
cannot dynamically add dependencies based on the function's runtime.

To fix this, we need to determine if a dependency is a hash when a
computed or an observer depends upon it, and then entangle all of its
keys if it is. Unfortunately, we cannot do this by just checking what
the value is, because we do not load the value of leaf dependencies
(the last segment of a dependency) as a perf optimization.

This tries to instead create a new type of "ComputedProperty" that gets
installed whenever a value is set to be a HashProxy. This gives us a
way to load the value of a property _only_ when it is a hash, or when it
was set to a hash at least once.

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/components/link-to/query-params-angle-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/components/link-to/query-params-angle-test.js
@@ -350,6 +350,30 @@ moduleFor(
       assert.equal(theLink.attr('href'), '/?foo=ASL');
     }
 
+    async ['@test supplied QP properties can be bound in legacy components'](assert) {
+      expectDeprecation(/Passing the `@tagName` argument to/);
+
+      this.addTemplate(
+        'index',
+        `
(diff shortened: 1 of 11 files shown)


pzuraq:
Depends on https://github.com/glimmerjs/glimmer-vm/pull/1319

PR #19587, opened 2021-06-03
[DOC] Add missing `Helper` class methods

### LINKED ISSUE #19585: [Documentation] missing methods for the `Helper` class
### 📙 Describe the Issue
It seems the [api docs for the `Helper` class](https://api.emberjs.com/ember/3.26/classes/Helper) only list the `compute` and `recompute` methods. I'm fairly certain that `willDestroy` is considered public API so at least that one is missing. 

I'm guessing the proper `@extends` annotation is missing from the YUIDoc comment?

### 🔬 Minimal Reproduction
https://api.emberjs.com/ember/3.26/classes/Helper

### 😕 Actual Behavior
Not all public methods are listed.

### 🤔 Expected Behavior
All methods are listed

### ➕ Additional Context
N/A

Windvis:
This adds the public `CoreObject` methods to the API docs of the `Helper` class which looks like this:

Screenshot from 2021-06-03 18-30-04

I assumed these are public API since the other classes that extend `CoreObject` (or its derivatives) also have them listed. If this isn't the case I think that `willDestroy` should still be documented since that is used in at least 1 already merged RFC.

Closes #19585

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/helper.ts
+++ b/packages/@ember/-internals/glimmer/lib/helper.ts
@@ -78,6 +78,7 @@ export interface SimpleHelper<T = unknown> {
   Additionally, class helpers can call `recompute` to force a new computation.
 
   @class Helper
+  @extends CoreObject
   @public
   @since 1.13.0
 */


rwjblue:
Thanks!

PR #19594, opened 2021-06-07
[BUGFIX release] Revert lazy hash changes

chancancode:
A series of BUGFIX release landed on the 3.27 release channel in an attempt to restore the previous lazy behavior. This is still desirable, but unfortunately the fixes caused other regressions. We have a new approach in mind that should be more compatible, but at this point, we have gone through a lot of changes with not a lot of time to test the approaches in the real world.

So the decision is to revert all of the related fixes on 3.27. This would restore the 3.26 buggy behavior, which was already broken since early 3.2x releases. We will land the new fix on 3.28 instead and let it go through a full beta cycle in order to access any real world impact properly.


pzuraq (approved):

rwjblue (approved):

snewcomer:
There is one strange issue I am seeing with these changes reverted.  An input loses focus and the active element becomes the body when updating query params.  Verified bug on 3.27.4 but not 3.27.3.  I could put together an example repro.

Issue #19546, opened 2021-05-14
[Bug] LinkTo won't transition if a parent element stopPropagation


jbescoyez:
### 🐞 Describe the Bug

When a `<LinkTo>` is inserted inside an element that `stopPropagation` of `click`'s, it won't transition but rather fallback on a default `<a>` behavior.

### 🔬 Minimal Reproduction

* Put a `<LinkTo>` inside an element that `stopPropagation` on `click`.

https://ember-twiddle.com/bac43dc0c91d12e7a9afb1d4adfcaabd

### 😕 Actual Behavior

* No transition
* Default `<a>` behavior

### 🤔 Expected Behavior

* An Ember transition to the `route`.

### 🌍 Environment

- Ember: 3.26.1
- Node.js/npm: 12.19.0 / 6.14.8
- OS: MacOS 10.15.5
- Browser: Chrome 90.0.4430.212

chancancode:
This is one of those hairy things that should be fixed in 3.27 / RFC #671. This is true for all classic components – because they use event delegation and listens on the `rootElement` (`body` by default), if the event is canceled before it makes it to the top then it will not trigger on the component. This is something to generally pay attention to when mixing classic components and modifiers. In the long run it will be less of an issue as components are migrated, and in the case of `<LinkTo>` specifically, 3.27 moves away from the classic implementation.

kpfefferle:
@chancancode Did you mean to link to https://github.com/emberjs/rfcs/pull/671 instead of https://github.com/emberjs/ember.js/issues/671?

We are now experiencing an issue in our Production builds after upgrading to 3.27 where `LinkTo`s will (seemingly) randomly decide to act as standard `<a>` tags instead of as Ember route transitions. The work related to the RFC seems to be a likely culprit 😬

chancancode:
@kpfefferle hmm, that is strange: I would think in 3.27 you would be *less* likely to run into issues like that, because on the modernized version the handler is attached directly on the <a> element using {{on}}, whereas before it is using event delegation on the <body>, so if the event get canceled before it reaches the top things would not work as expected.

One thing you can try is to call `LinkComponent.reopen()`. This forces ember to fallback to the old implementation (you will get a deprecation), so you can use that to test if the issues you are expecting is related to the change or not.

sinankeskin:
[...]

PR #19597, opened 2021-06-10
[BIGFIX release] Fix `<LinkTo>` with nested children

chancancode:
During bubbling, `event.target` may point to a child element whereas `event.currentTarget` always points to the element where the handler was attached, which is what we want here.

Reported in a comment on #19546, though this may be a distinct issue from the original report as it was reported as a default-cancelling parent element interfering with the nested `<LinkTo>`, and this is the other way around.

cc @lele-melis @kpfefferle


rwjblue (approved):

chancancode:
Release in 3.27.5

PR #19599, opened 2021-06-11
Fix Changelog formatting

Turbo87:
The different heading levels are causing issues for services that are trying to read and filter the changelog

/cc @rwjblue


rwjblue (approved):
Thank you!

PR #19608, opened 2021-06-22
[BUGFIX lts] Ensures chain tags do not call unknownProperty unless necessary

### LINKED ISSUE #1146: Nightly Run Failure: 960921280
Nightly run failed on: 22 Jun 2021
https://github.com/hjdivad/ember-m3/actions/runs/960921280

pzuraq:
Fixes https://github.com/hjdivad/ember-m3/issues/1146

### DIFF
--- a/packages/@ember/-internals/metal/lib/chain-tags.ts
+++ b/packages/@ember/-internals/metal/lib/chain-tags.ts
@@ -174,6 +174,27 @@ function getChainTags(
 
     chainTags.push(propertyTag);
 
+    // If we're at the end of the path, processing the last segment, and it's
+    // not an alias, we should _not_ get the last value, since we already have
+    // its tag. There's no reason to access it and do more work.
+    if (segmentEnd === pathLength) {
+      // If the key was an alias, we should always get the next value in order to
+      // bootstrap the alias. This is because aliases, unlike other CPs, should
+      // always be in sync with the aliased value.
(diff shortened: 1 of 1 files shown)


runspired:
I'm investigating if this also fixes the beta/canary failures in ember-data

runspired:
Applied this change locally and unfortunately it does not fix the test failures in ember-data. I'm not convinced we have the same root-cause as ember-m3 vs just a similar cause, though I could believe it because the test failures are very similarly around over-notification.

In our case we have an observer (not array observer) watching either `prop.@each` or `prop.[]` where prop is an async hasMany relationship. In one of the tests it is easy to covert the relationship to sync and so I was able to observe a failure still occurring with just the MutableArray implementation involved. It is worth noting that ember-data's tests *have not* defaulted observers to async and also *have not* manually passed options to these observers to set them as async/sync, which should mean that they are running in sync mode.

I'll happily set aside time to debug a root cause with you,

pzuraq:
This fix ended up not being the path we want to go down, we're going to do a different fix instead.

rwjblue:
@pzuraq What is the alternate path? As far as I can tell, both ember-data and ember-m3 are still borked ATM...

pzuraq:
@rwjblue the full fix landed in Glimmer VM and needs to be updated in Ember still: https://github.com/glimmerjs/glimmer-vm/pull/1323

The short term solution was to revert all of the lazy hash changes in general, since they were broken, and then update to the latest version of Glimmer VM. I thought we had reverted these on canary and beta, did that not happen?

Issue #19617, opened 2021-07-02
Deprecation Cleanup for 4.0


kategengler:
**PLEASE FLAG REMOVAL/CLEANUP PRs WITH `[CLEANUP beta]` IN THE PR TITLE.**

The following are the deprecations in ember.js that are `until` 4.0 or earlier. Each item includes the id of the deprecation in backticks. Searching for this id in the source will find the invocations of that deprecation.

## Remaining

- [x] https://github.com/emberjs/ember.js/pull/19846 - Optional feature: `application-template-wrapper` https://github.com/emberjs/ember.js/blob/5a36048abfda243a7715cc6304d69ecc7cc9bddb/lib/index.js#L131
- [x] https://github.com/emberjs/ember.js/pull/19846 - Optional feature: Unspecified or specified `classic` edition https://github.com/emberjs/ember.js/blob/5a36048abfda243a7715cc6304d69ecc7cc9bddb/lib/index.js#L123
- [x] https://github.com/emberjs/ember.js/pull/19838 - Change `bower` usage to a `throw` instead of warn at https://github.com/emberjs/ember.js/blob/ce98e570d21f74348fb64ebd9fd4624a3fba7d92/lib/index.js#L47
- [x] https://github.com/emberjs/ember.js/pull/19836 - Optional feature: `template-only-glimmer-components` https://github.com/emberjs/ember.js/blob/5a36048abfda243a7715cc6304d69ecc7cc9bddb/lib/index.js#L140
- [x] https://github.com/emberjs/ember.js/pull/19833 Array Observers `array-observers` [Deprecation Guide](https://deprecations.emberjs.com/v3.x#toc_array-observers) [Deprecation Invocation Site](https://github.com/emberjs/ember.js/blob/71f9cd0c2212cbce988a6056bac08bf2c6c9db6c/packages/@ember/-internals/metal/lib/array.ts#L99) + 1 other in same file
- [x] Ember.Component.reopen `ember.component.reopen`[Deprecation Guide]() [Deprecation Invocation Site](https://github.com/emberjs/ember.js/blob/9609b8f30e0cdbc2235063f571af619f4362f73b/packages/@ember/-internals/glimmer/lib/component.ts#L1156) see also https://github.com/emberjs/ember.js/pull/19744#issuecomment-954956205
[...]

PR #19624, opened 2021-07-13
Remove deprecated disabledWhen

nlfurniss:
Part of #19617

`TEST_SUITE=all yarn test` passes locally

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/components/link-to.ts
+++ b/packages/@ember/-internals/glimmer/lib/components/link-to.ts
@@ -614,7 +614,6 @@ if (EMBER_MODERNIZED_BUILT_IN_COMPONENTS) {
     });
   }
 
-  // @disabledWhen
   {
     let superIsSupportedArgument = prototype['isSupportedArgument'];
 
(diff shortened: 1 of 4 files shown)


packages/@ember/-internals/glimmer/lib/components/link-to.ts
```diff
@@ -614,32 +614,13 @@ if (EMBER_MODERNIZED_BUILT_IN_COMPONENTS) {
...
-              }
-            );
-
-            return true;
-          }
-        }
-
         return superIsSupportedArgument.call(this, name);
       },
     });
```
mixonic, line 626: I think *all* of this can go, as what remains after your patch is only the call of the super method?

packages/@ember/-internals/glimmer/lib/components/link-to.ts
```diff
@@ -657,22 +638,6 @@ if (EMBER_MODERNIZED_BUILT_IN_COMPONENTS) {
...
-              until: '4.0.0',
-            }
-          );
-
-          return Boolean(this.named('disabledWhen'));
-        }
-
         return superGetter.call(this);
       },
     });
```
mixonic, line 643: Same here?

packages/@ember/-internals/glimmer/tests/integration/components/link-to/routing-angle-test.js
```diff
@@ -410,7 +331,7 @@ moduleFor(
       assert.equal(
         this.$('h3.about').length,
         1,
-        'Transitioning did occur when disabledWhen became false'
+        'Transitioning did occur when disabled became false'
```
mixonic, line 334: :+1:

mixonic (changes requested):

mixonic:

packages/@ember/-internals/glimmer/lib/components/link-to.ts
nlfurniss: Not familiar with this code: would `isDisabled` still work if I remove it?

mixonic: I think so. And even a few lines above this can get dropped I believe.

You could poke around for a test of `isDisabled` to confirm the removal doesn't regress anything.

mixonic:
Replaced by https://github.com/emberjs/ember.js/pull/19669

PR #19627, opened 2021-07-14
Remove for and since deprecations from deprecate method

nlfurniss:
Part of #19617

```
Deprecate without for ember-source-deprecation-without-for Deprecation Invocation Site
Deprecate without since ember-source-deprecation-without-since Deprecation Invocation Site
```

`TEST_SUITE=all yarn test` passes locally

### DIFF
--- a/packages/@ember/debug/lib/deprecate.ts
+++ b/packages/@ember/debug/lib/deprecate.ts
@@ -72,8 +72,6 @@ let missingOptionsUntilDeprecation: string;
 let missingOptionsForDeprecation: MissingOptionDeprecateFunc = () => '';
 let missingOptionsSinceDeprecation: MissingOptionDeprecateFunc = () => '';
 let deprecate: DeprecateFunc = () => {};
-let FOR_MISSING_DEPRECATIONS = new Set();
-let SINCE_MISSING_DEPRECATIONS = new Set();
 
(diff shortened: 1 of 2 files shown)


packages/@ember/debug/lib/deprecate.ts
```diff
@@ -204,32 +202,8 @@ if (DEBUG) {
...
-        id: 'ember-source.deprecation-without-since',
-        until: '4.0.0',
-        for: 'ember-source',
-        since: {
-          enabled: '3.24.0',
-        },
-      });
-    }
+    assert(missingOptionsForDeprecation(options!.id), Boolean(options!.for));
+    assert(missingOptionsSinceDeprecation(options!.id), Boolean(options!.since));
```
mixonic, line 206: Seems inconsistent that these asserts bother to include the deprecation id and the earlier asserts do not.

mixonic (approved):

packages/@ember/debug/lib/deprecate.ts
nlfurniss: Yeah I wasn't sure why, but seems like good so I left it

mixonic:

mixonic:
Replaced by https://github.com/emberjs/ember.js/pull/19643

PR #19633, opened 2021-07-15

nlfurniss:


mixonic:
Replaced by https://github.com/emberjs/ember.js/pull/19649

PR #19634, opened 2021-07-15

nlfurniss:

### DIFF
--- a/packages/@ember/polyfills/tests/assign_test.js
+++ b/packages/@ember/polyfills/tests/assign_test.js
@@ -49,7 +49,9 @@ moduleFor(
   'Ember.assign (polyfill)',
   class extends AssignTests {
     assign() {
-      return assignPolyfill(...arguments);
+      return expectDeprecation(() => {
+        assignPolyfill(...arguments);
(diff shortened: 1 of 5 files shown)


packages/@ember/polyfills/tests/assign_test.js
```diff
@@ -49,7 +49,9 @@ moduleFor(
   'Ember.assign (polyfill)',
   class extends AssignTests {
     assign() {
-      return assignPolyfill(...arguments);
+      return expectDeprecation(() => {
```
mixonic, line 52:
```suggestion
      expectDeprecation(() => {
```

If we can safely discard the return value of `assignPolyfill`'s result, then we simply don't need a return value.

mixonic (changes requested):
Minor ask, thanks for this patch!

mixonic:
Also, I believe we need someone to commit to writing the deprecation guide in the next few weeks (before the beta cut) in order to land this. It seems like we need 4.x section in the guides app https://github.com/ember-learn/deprecation-app/issues/898

mixonic:

PR #19643, opened 2021-07-18
Remove deprecate without for and since

nlfurniss:


packages/@ember/debug/lib/deprecate.ts
```diff
@@ -204,32 +202,8 @@ if (DEBUG) {
...
-        id: 'ember-source.deprecation-without-since',
-        until: '4.0.0',
-        for: 'ember-source',
-        since: {
-          enabled: '3.24.0',
-        },
-      });
-    }
+    assert(missingOptionsForDeprecation(options!.id), Boolean(options!.for));
+    assert(missingOptionsSinceDeprecation(options!.id), Boolean(options!.since));
```
chancancode, line 206: We should probably take the opportunity to refactor the logic here.

For example, I can't think of a reason why we should treat `until` any differently than `for` and `since`. `id` may be the only "special" one since we can't include the `id` for the deprecation in the error message if it's missing. But other than that, the rest can probably share the logic in the same helper function, e.g. `missingOptionsDeprecation(options.id)`.

Also, I _think_ `assert` is now typed as a TypeScript assertion function. So if you make the first check `assert(..., options !== null && typeof options === 'object');`, I think you won't have to write `options!` for the lines below. Likewise, if you have `assert(..., typeof options.id === 'string');` then for the rest of the function TypeScript should know that `options.id` is a present and is a string.

chancancode: Since you will be making this change anyway, can you rebase and squash your commits?

nlfurniss: Refactored to have most assertions use one function.

I tried the above, but the linter wasn't happy. I'm not experience w/TS so I'm going to leave it as-is unless someone objects.

PR was rebase w/master so no more merge commits.

PR #19649, opened 2021-07-18

nlfurniss:


nlfurniss:
Rebased and pushed. Not sure why lint failed on a file I didn't edit 🤨

packages/@ember/deprecated-features/index.ts
```diff
@@ -13,3 +13,4 @@ export const FUNCTION_PROTOTYPE_EXTENSIONS = !!'3.11.0-beta.1';
 export const MOUSE_ENTER_LEAVE_MOVE_EVENTS = !!'3.13.0-beta.1';
 export const PARTIALS = !!'3.15.0-beta.1';
 export const GLOBALS_RESOLVER = !!'3.16.0-beta.1';
+export const ASSIGN = !!'4.0.0';
```
rwjblue, line 16:
```suggestion
export const ASSIGN = !!'4.0.0-beta.1';
```

This will be the first version with the deprecation

rwjblue (commented):
Only one small tweak here, but I'd love another PR that actually removes the implementation in `packages/@ember/polyfills/lib/assign.ts`. Now that we only support browsers that include `Object.assign` we can basically do:

```js
export default Object.assign;
```

(also removing the `babel-plugin-transform-object-assign` usage)

packages/@ember/polyfills/lib/assign.ts
```diff
@@ -28,6 +30,20 @@ export function assign(target: object, ...sources: any[]): any;
   @static
 */
 export function assign(target: object) {
+  deprecate(
+    'Use of `assign` has been deprecated. Please use `Object.assign` or the spread operator instead.',
+    false,
+    {
+      id: 'ember-polyfills.deprecate-assign',
+      until: '5.0.0',
+      url: 'https://deprecations.emberjs.com/v4.x/#toc_ember-polyfills-deprecate-assign',
```
rwjblue, line 39: Need to make a PR to the deprecation app, and land it so that we can merge this (once we confirm the URL works)

nlfurniss: Way ahead of you 😉 

https://github.com/ember-learn/deprecation-app/pull/899

nlfurniss:
> Only one small tweak here, but I'd love another PR that actually removes the implementation in `packages/@ember/polyfills/lib/assign.ts`. Now that we only support browsers that include `Object.assign` we can basically do:
> 
> ```js
> export default Object.assign;
> ```
> 
> (also removing the `babel-plugin-transform-object-assign` usage)

Add that here, since it seems like it should be in one PR, but happy to move if you'd like.

and the babel transform was removed in #19563

mixonic (approved):

mixonic:
Looks good to me. I rebased to resolve a conflict and force pushed.

PR #12685, opened 2015-12-05
[FEATURE ember-application-engines] Initial extraction of engines from applications

dgeb:
This initial PR for engines follows the strategy proposed in the [Engines RFC](https://github.com/emberjs/rfcs/pull/10).

It introduces a base class, `Engine`, which is wholly extracted from (and then extended by) `Application`. Similarly, `EngineInstance` has been wholly extracted from (and then extended by) `ApplicationInstance`.

The `Engine` base class is now responsible for maintaining a registry and initializers.

The `EngineInstance` base class is now responsible for maintaining an instance-level registry and container.

`Ember.Engine` is only exposed when the new `ember-application-engines` feature flag is enabled.

The APIs and behavior of `Application` and `ApplicationInstance` remain identical, as do tests and documentation.

Some tests of extracted behavior were duplicated for engines, but the original tests for applications were kept in place. Any test duplication should be revisited before the feature flag is enabled.

The primary purpose of this PR is to allow for further experimentation with engines in an addon. It will also allow behavior to be folded into ember core gradually as experimentation proceeds.

PR #19652, opened 2021-07-18
[CLEANUP] Remove Globals Resolver

btecu:


mixonic:
@btecu there was a failure in the Node tests here, and this PR will require a rebase as well.

btecu:
@mixonic I might need some help with the Node tests failure if you have some time.

mixonic:
@btecu the `setup-app` boilerplate, and many of the tests in `tests/node/` are using deprecate API. Normally this is not permitted in the Ember test suite, and a test will fail if a deprecation is present without expectation. However these tests run in their own realm. Without that enforcement, it seems like they have fallen far off the best-practice path.

My commit adds a resolver to the app, thus avoiding the default resolver (which after your removal is just a hard failure).

btecu:
@mixonic thanks for the explanation and your help!

packages/@ember/engine/index.js
```diff
@@ -417,15 +416,6 @@ Engine.reopenClass({
...
   },
 
-  /**
-    Set this to provide an alternate class to `DefaultResolver`
-
-    @deprecated Use 'Resolver' instead
-    @property resolver
-    @public
-  */
-  resolver: null,
```
mixonic, line 427: This is weird. I see it was added in https://github.com/btecu/ember.js/commit/7b0a7ecb5b9e863bded0094d8b7e5c1861e1d52f#diff-a7583c2d819fdabaff78778f93e5112adc888bc3026338522862f08ddfadb4abR364 (or https://github.com/emberjs/ember.js/pull/12685/files#diff-a7583c2d819fdabaff78778f93e5112adc888bc3026338522862f08ddfadb4abR377 in the PR) by @dgeb but I don't see any usage of it. What am I missing? It was added with a `@deprecated` flag in jsdoc, but I don't see what is firing the deprecation.

mixonic (commented):
Left an item I'd appreciate another set of eyes on. I think this is all good.

packages/@ember/engine/index.js
dgeb: > but I don't see what is firing the deprecation.

I wonder if the code firing the deprecation was removed prior to 3.0 without removing the deprecated prop? Regardless, it doesn't seem that the `resolver` class prop is referenced, so it seems safe to remove this now.

dgeb (approved):
Looks good to me. Thanks @btecu and @mixonic for seeing this through.

PR #19653, opened 2021-07-18
Remove run and computed dot access

mixonic:

Removing this code there is a bit of stuff adjacent to it regarding Ember global access. I haven't touched the Ember global stuff here, but this should probably land first and we can rip more code out during the Ember global removal.


packages/internal-test-helpers/lib/confirm-export.js
```diff
@@ -25,10 +25,6 @@ export default function confirmExport(Ember, assert, path, moduleId, exportName,
       desc = null;
     }
 
-    if (isDeprecated) {
-      expectDeprecation();
```
mixonic, line 29: `expectDeprection()` called without a callback sets up an expectation that a deprecation occurs anywhere during the test run. As `confirmExport` is called many times in the re-export test, this means if any *one* re-export* was expecting a deprecation satisfaction of that expectation would mean any later expectation was satisfied regardless of a new occurrence.

TLDR, https://github.com/emberjs/ember.js/pull/19653/files#diff-a9c227f21c32919d338aab29999b3ceaf7ca894a7bd3d7183ea81eda7578bdafL307 was a false positive. There was never a deprecation being issues for `expandProperties`, but the test passed because the computed macro tests *did* have deprecations.

Since `confirmExport` is not being used in a manner compatible with `isDeprecated` being an argument, and since there are no uses of it with that argument being `true` after this PR lands, I have removed the functionality from `confirmExport`. We can re-introduce it correctly when needed.

packages/ember/tests/reexports_test.js
```diff
@@ -25,9 +25,7 @@ moduleFor(
           assert,
           path,
           moduleId,
-          exportName,
-          isDeprecated,
-          `Ember.${path} exports correctly`
```
mixonic, line 30: This 7th argument (a string description) is not used. Here it is removed.

Note: also see https://github.com/emberjs/ember.js/pull/19653/files#r671884855 regarding the removal of `isDeprecated` on the line above.

rwjblue (approved):

PR #19654, opened 2021-07-18
Remove @ember/string methods from native prototype

mixonic:
Deprecated API removal per https://github.com/emberjs/ember.js/issues/19617

~This patch does *not* removed the setting of `String.prototype.htmlSafe`, as installed at https://github.com/emberjs/ember.js/blob/a884ee5db7a69899f2689e2da2324bb85114c805/packages/ember/index.js#L548-L552. I don't believe that was ever deprecated. Maybe we just missed it?~

~As such, this also does not completely remove `ENV.EXTEND_PROTOTYPES.String`~

_edit: The above caveat was discussed and you can read the comments here for the details._

### DIFF
--- a/packages/@ember/-internals/environment/lib/env.ts
+++ b/packages/@ember/-internals/environment/lib/env.ts
@@ -16,7 +16,7 @@ export const ENV = {
   ENABLE_OPTIONAL_FEATURES: false,
 
   /**
-    Determines whether Ember should add to `Array` and `String`
+    Determines whether Ember should add to `Array`
     native object prototypes, a few extra methods in order to provide a more
     friendly API.
 
@@ -36,7 +36,6 @@ export const ENV = {
   */
   EXTEND_PROTOTYPES: {
     Array: true,
-    String: true,
   },
 
   /**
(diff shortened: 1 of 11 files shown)


rwjblue (changes requested):
Also needs to remove these things:

https://github.com/emberjs/ember.js/blob/ff1603207e2e255be69968654908f45cfbc18eee/packages/%40ember/-internals/environment/lib/env.ts#L40

https://github.com/emberjs/ember.js/blob/ff1603207e2e255be69968654908f45cfbc18eee/packages/%40ember/-internals/environment/lib/env.ts#L221

https://github.com/emberjs/ember.js/blob/ff1603207e2e255be69968654908f45cfbc18eee/packages/%40ember/-internals/environment/lib/env.ts#L225

rwjblue:
> As such, this also does not completely remove ENV.EXTEND_PROTOTYPES.String

Hmph. This is unfortunate. 🤔

locks:
> Hmph. This is unfortunate. 🤔

I've added to this week's agenda so we have a think.

mixonic:
@rwjblue I can't do those removals until we are sure we can remove `htmlSafe`, right? This PR could land as-is, and we could do the `htmlSafe` removal and complete removal of `EXTEND_PROTOTYPES.String` in another PR.

Or we can block this until we figure out the path for `htmlSafe`. I'd rather decouple and just land this, but your review can make the call.

mixonic:
At the July 23rd Ember Framework Core meeting we agreed that `String.prototype.htmlSafe` was intended to be deprecated. We don't expect it is in very high use across the Ember community, and we're comfortable expanding the string prototype deprecation to include it in a point release for 3.27 and in 3.28.0.

This PR should include `String.prototype.htmlSafe` in its removals, but before this is merged we should first write and land the PR to add it to the deprecations.

packages/@ember/-internals/environment/lib/env.ts
```diff
@@ -36,8 +36,6 @@ export const ENV = {
   */
   EXTEND_PROTOTYPES: {
     Array: true,
-    Function: true,
```
mixonic, line 39: There are two commits in this PR, and the first makes this unrelated change to remove `Function` from this list and docs. The code deletion for function methods on prototype was done at https://github.com/emberjs/ember.js/pull/19663 but this removal was missed.

kategengler (approved):

PR #19657, opened 2021-07-18
Remove hasBlock and hasBlockParams

mixonic:
In favor of `(has-block)` and `(has-block-params)`.

I've updated a number of tests to use the new interface instead of removing those tests. If this system is well tested in Glimmer, perhaps that is unnecessary, but it seems safe.


rwjblue (approved):

rwjblue:
@mixonic we missed some stuff here https://github.com/emberjs/ember.js/blob/2f6a72e8686f34d96a9008b29e5b0cb7561b8f98/packages/ember-template-compiler/lib/plugins/transform-has-block-syntax.ts

mixonic:
@rwjblue I see that file as removed in this PR (https://github.com/emberjs/ember.js/pull/19657/files#diff-ade9f2c13818cdb5fdb622474e656ca3b2fc69d140c3728505d9f0c5593ec439), and it seems to be deleted on `master` (https://github.com/emberjs/ember.js/tree/master/packages/ember-template-compiler/lib/plugins).

Was there something else that was missed for deletion?

PR #19659, opened 2021-07-19
Remove renderTemplate, disconnectOutlet, render

mixonic:
Removing these features also effectively removes support for named outlets, however I did not update the docs or code to prevent a name from being passed to `{{outlet 'foo'}}` as it was not explicitly deprecated.


rwjblue (approved):

mixonic:
The comment here around `{{outlet 'foo'}}` being dropped as API was discussed on July 23rd with Ember Framework Core. We agreed that `{{outlet 'foo'}}` should become an assertion in dev builds. In a production build, `'foo'` is an argument passed where `{{outlet}}` would expect none, and so it would be silently ignored. Asserting in dev will prevent users who expect an argument here from stumbling into an unexpected behavior.

PR #19660, opened 2021-07-19
Remove attrs/attrs-arg-access

nlfurniss:
Part of #19617

[Deprecation Guide](https://deprecations.emberjs.com/v3.x/#toc_attrs-arg-access)

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/components/curly-components-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/components/curly-components-test.js
@@ -1257,12 +1257,16 @@ moduleFor(
     }
 
     ['@test non-block with properties on attrs']() {
-      expectDeprecation(
-        "Using {{attrs}} to reference named arguments has been deprecated. {{attrs.someProp}} should be updated to {{@someProp}}. ('my-app/templates/components/non-block.hbs' @ L1:C24) "
-      );
+      expectAssertion(() => {
+        this.registerComponent('non-block', {
+          template: 'In layout - someProp: {{attrs.someProp}}',
+        });
+      }, "Using {{attrs}} to reference named arguments is not supported. {{attrs.someProp}} should be updated to {{@someProp}}. ('my-app/templates/components/non-block.hbs' @ L1:C24) ");
+    }
 
+    ['@test non-block with properties on this.attrs']() {
       this.registerComponent('non-block', {
-        template: 'In layout - someProp: {{attrs.someProp}}',
+        template: 'In layout - someProp: {{this.attrs.someProp}}',
       });
 
       this.render('{{non-block someProp=this.prop}}', {
@@ -1469,30 +1473,11 @@ moduleFor(
       );
     }
 
-    ['@test this.attrs.foo === attrs.foo === @foo === foo']() {
-      expectDeprecation(
-        "Using {{attrs}} to reference named arguments has been deprecated. {{attrs.value}} should be updated to {{@value}}. ('my-app/templates/components/foo-bar.hbs' @ L1:C8) "
-      );
-
-      expectDeprecation(
-        "Using {{attrs}} to reference named arguments has been deprecated. {{attrs.value}} should be updated to {{@value}}. ('my-app/templates/components/foo-bar.hbs' @ L1:C31) "
-      );
-
-      expectDeprecation(
-        "Using {{attrs}} to reference named arguments has been deprecated. {{attrs.items}} should be updated to {{@items}}. ('my-app/templates/components/foo-bar.hbs' @ L1:C82) "
-      );
-
-      expectDeprecation(
-        "Using {{attrs}} to reference named arguments has been deprecated. {{attrs.items}} should be updated to {{@items}}. ('my-app/templates/components/foo-bar.hbs' @ L1:C135) "
-      );
-
+    ['@test this.attrs.foo === @foo === foo']() {
       this.registerComponent('foo-bar', {
         template: strip`
-        Args: {{this.attrs.value}} | {{attrs.value}} | {{@value}} | {{this.value}}
+        Args: {{this.attrs.value}} | {{@value}} | {{this.value}}
         {{#each this.attrs.items as |item|}}
-          {{item}}
-        {{/each}}
-        {{#each attrs.items as |item|}}
           {{item}}
         {{/each}}
         {{#each @items as |item|}}
@@ -1518,11 +1503,11 @@ moduleFor(
         this.context.set('model.items', [1]);
       });
 
-      this.assertText(strip`Args: lul | lul | lul | lul1111`);
+      this.assertText(strip`Args: lul | lul | lul111`);
 
       runTask(() => this.context.set('model', { value: 'wat', items: [1, 2, 3] }));
 
-      this.assertText('Args: wat | wat | wat | wat123123123123');
+      this.assertText('Args: wat | wat | wat123123123');
     }
 
     ['@test non-block with properties on self']() {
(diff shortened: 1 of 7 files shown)


packages/@ember/-internals/glimmer/tests/integration/components/curly-components-test.js
```diff
@@ -1514,62 +1486,6 @@ moduleFor(
       );
     }
 
-    ['@test this.attrs.foo === attrs.foo === @foo === foo']() {
```
mixonic, line 1517: Perhaps this test should become `['@test @foo === this.foo']() {`? I'm not sure that is covered elsewhere, and it seems worth retaining that assertion.

mixonic (changes requested):

packages/@ember/-internals/glimmer/tests/integration/components/curly-components-test.js
nlfurniss: added back and modified test

rwjblue: This should also still include `this.attrs.foo === this.foo === @foo`, the only thing removed is the "magical" rewrite of `{{attrs.foo}}` -> `{{@foo}}`.

rwjblue (changes requested):

packages/@ember/-internals/glimmer/tests/integration/components/curly-components-test.js
nlfurniss: based on the attrs transform test, it seems like `this.attrs` is also deprecated:

```js
expectDeprecation(() => {
        this.assertTransformed(`{{this.attrs.foo}}`, `{{@foo}}`);
      }, /Using {{attrs}} to reference named arguments has been deprecated. {{attrs.foo}} should be updated to {{@foo}}./);
```
is this not the case?

mixonic: I think at that point the question becomes "Was `this.attrs` in `EmberComponent` deprecated? Can it be removed?`

If the `EmberComponent` system is setting a JS property `attrs` on the instance still, then `{{this.attrs.foo}}` remains a reasonable thing to use *even if it was deprecated*.

However I agree that the main goal here was the complete elimination of `this.attrs` in all places. I'll add this to our agenda for Friday with Ember framework core in case we don't have clarity by then.

mixonic: This was clarified with Ember FW Core on July 23rd.

Despite the fact that `this.attrs` was deprecated, we should retain support for it in 4.0. We *should* add an ember-template-lint rule against using `this.attrs` and `this.args` in templates, but as they are each properties on the backing class for a template there is no reason the syntax/usage itself is invalid. The framework should not do anything special.

So, yes, here we can retain `this.attrs.foo === this.foo === @foo` without deprecation.

nlfurniss: k made those changes

packages/@ember/-internals/glimmer/tests/integration/helpers/closure-action-test.js
```diff
@@ -250,14 +250,10 @@ moduleFor(
...
-    ['@test [#12718] a nice error is shown when a bound action function is undefined and it is passed as attrs.foo']() {
-      expectDeprecation(
-        "Using {{attrs}} to reference named arguments has been deprecated. {{attrs.external-action}} should be updated to {{@external-action}}. ('my-app/templates/components/inner-component.hbs' @ L1:C43) "
-      );
-
+    ['@test [#12718] a nice error is shown when a bound action function is undefined and it is passed as this.attrs.foo']() {
       this.registerComponent('inner-component', {
         template:
-          '<button id="inner-button" {{action (action attrs.external-action)}}>Click me</button>',
+          '<button id="inner-button" {{action (action this.attrs.external-action)}}>Click me</button>',
```
mixonic, line 256:
```suggestion
          '<button id="inner-button" {{action (action @external-action)}}>Click me</button>',
```

packages/@ember/-internals/glimmer/tests/integration/helpers/closure-action-test.js
```diff
@@ -250,14 +250,10 @@ moduleFor(
       }, /An action could not be made for `.*` in .*\. Please confirm that you are using either a quoted action name \(i\.e\. `\(action '.*'\)`\) or a function available in .*\./);
     }
 
-    ['@test [#12718] a nice error is shown when a bound action function is undefined and it is passed as attrs.foo']() {
-      expectDeprecation(
-        "Using {{attrs}} to reference named arguments has been deprecated. {{attrs.external-action}} should be updated to {{@external-action}}. ('my-app/templates/components/inner-component.hbs' @ L1:C43) "
-      );
-
+    ['@test [#12718] a nice error is shown when a bound action function is undefined and it is passed as this.attrs.foo']() {
```
mixonic, line 253:
```suggestion
    ['@test [#12718] a nice error is shown when a bound action function is undefined and it is passed as @foo']() {
```

packages/ember-template-compiler/lib/plugins/transform-attrs-into-args.ts
```diff
@@ -64,24 +64,17 @@ export default function transformAttrsIntoArgs(env: EmberASTPluginEnvironment):
...
-              until: '4.0.0',
-              for: 'ember-source',
-              since: {
-                enabled: '3.26.0',
-              },
-            }
-          );
+          if (node.this === false) {
+            assert(
+              `Using {{attrs}} to reference named arguments has been deprecated. {{attrs.${
```
mixonic, line 69: *If* we added an assertion:

```suggestion
              `Using {{attrs}} to reference named arguments is unsupported. {{attrs.${
```

And this test should be renamed to `assert-against-attrs` in the style of https://github.com/emberjs/ember.js/pull/19691/files, it is no longer a transform. The transforming behaviors here should be dropped.

**However** this removal does not need to add an assert against `{{attrs}}`. We can just drop the transform.

The suggestion in my last review was that a template lint could be added to ember-template-lint against `{{this.attrs}}`. Once we remove the deprecation in this PR, `{{this.attrs}}` will Just Work in Ember components like any other property access. We should lint against it, but we cannot *break* it. That would make it special, and our goal is to have fewer special things.

packages/@ember/-internals/glimmer/tests/integration/application/engine-test.js
```diff
@@ -273,7 +273,7 @@ moduleFor(
         `The use of \`{{partial}}\` is deprecated, please refactor the "troll" partial to a component`
       );
       expectDeprecation(
-        'Using {{attrs}} to reference named arguments has been deprecated. {{attrs.wat}} should be updated to {{@wat}}. (L1:C2) '
+        'Using {{this.attrs}} to reference named arguments has been deprecated. {{this.attrs.wat}} should be updated to {{@wat}}. (L1:C2) '
```
mixonic, line 276: Something is off target- we should not still have a deprecation. `{{this.attrs}}` should work in an Ember component context because it behaves like any other property access, and `attrs` are set on Ember components.

The goal of this removal is to: a) drop any special transform of `{{attrs}}` b) removed the deprecation warning. That likely has the side effect that `{{this.attrs}}` is no longer deprecated, and that is ok.

mixonic (changes requested):

packages/@ember/-internals/glimmer/tests/integration/helpers/closure-action-test.js
nlfurniss: would we not want both? or is that overkill?

packages/ember-template-compiler/lib/plugins/transform-attrs-into-args.ts
nlfurniss: It doesn't seem (based on [a comment](https://github.com/emberjs/ember.js/blob/master/packages/ember-template-compiler/lib/plugins/transform-attrs-into-args.ts#L24)) that the transform is not needed. Also, if I comment out the transform, tests that use `this.attrs.foo` fail with

```
Expected: "In layout - someProp: something here"
Result: "In layout - someProp: [object Object]"
```

packages/@ember/-internals/glimmer/tests/integration/application/engine-test.js
nlfurniss: Yup, this was a mistake on my part. Odd that it didn't fail even though there was no deprecation.

Removed

packages/@ember/-internals/glimmer/tests/integration/helpers/closure-action-test.js
mixonic: `this.attrs` should have nothing "special" about it in 4.0. It didn't in 3.x either, except for the deprecation warning.

The question here is: Is the test asserting when a the actions is passed via an internal property of the component class (`this.any.thing.at.all` including `this.attrs.foo`) or read from external arguments (`@foo`)?

In my reading the test appears more targeted toward an external argument. The variable for the action is named `external-action` for example.

mixonic:
Thanks for all the work on this @nlfurniss

PR #19661, opened 2021-07-19
Remove EMBER_EXTEND_PROTOTYPES

nlfurniss:
Part of #19660

[Deprecation Guide](https://github.com/emberjs/ember.js/issues/19617)

Tests pass locally


mixonic:
I *think* I would like to see https://github.com/emberjs/ember.js/pull/19654 land first as I think it is a subset of this removal? I haven't looked closely at them.

See also the issue around `htmlSafe` mentioned on that PR.

nlfurniss:
> I _think_ I would like to see #19654 land first as I think it is a subset of this removal? I haven't looked closely at them.
> 
> See also the issue around `htmlSafe` mentioned on that PR.

As I understand it (and this may be wildly incorrect), this PR removes setting/reading the extension prototype setting on the `Ember` obj, while setting/reading it from EmberEnv is allowed:

`'Accessing Ember.EXTEND_PROTOTYPES is deprecated, please migrate to Ember.ENV.EXTEND_PROTOTYPES'`

PR #19662, opened 2021-07-19
Remove LinkTo's tagName

nlfurniss:
Part of #19617

[Deprecation Guide](https://github.com/emberjs/ember.js/issues/19617)

Tests pass locally

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/components/link-to.ts
+++ b/packages/@ember/-internals/glimmer/lib/components/link-to.ts
@@ -418,40 +418,6 @@ if (EMBER_MODERNIZED_BUILT_IN_COMPONENTS) {
     });
   }
 
-  // @tagName
-  {
-    let superOnUnsupportedArgument = prototype['onUnsupportedArgument'];
-
-    Object.defineProperty(prototype, 'onUnsupportedArgument', {
-      configurable: true,
-      enumerable: false,
-      value: function onUnsupportedArgument(this: LinkTo, name: string): void {
-        if (name === 'tagName') {
-          let tagName = this.named('tagName');
-
-          deprecate(
-            `Passing the \`@tagName\` argument to <LinkTo> is deprecated. Using a <${tagName}> ` +
-              'element for navigation is not recommended as it creates issues with assistive ' +
-              'technologies. Remove this argument to use the default <a> element. In the rare ' +
-              'cases that calls for using a different element, refactor to use the router ' +
-              'service inside a custom event handler instead.',
-            false,
-            {
-              id: 'ember.link-to.tag-name',
-              for: 'ember-source',
-              since: {},
-              until: '4.0.0',
-            }
-          );
-
-          this.modernized = false;
-        } else {
-          superOnUnsupportedArgument.call(this, name);
-        }
-      },
-    });
-  }
-
   // @bubbles & @preventDefault
   {
     let superIsSupportedArgument = prototype['isSupportedArgument'];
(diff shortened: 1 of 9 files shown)


packages/@ember/-internals/glimmer/lib/components/link-to.ts
```diff
@@ -418,40 +418,6 @@ if (EMBER_MODERNIZED_BUILT_IN_COMPONENTS) {
...
-            false,
-            {
-              id: 'ember.link-to.tag-name',
-              for: 'ember-source',
-              since: {},
-              until: '4.0.0',
-            }
-          );
-
-          this.modernized = false;
```
mixonic, line 447: It is asserted that passing `<LinkTo @tagName="div">` still renders an `<a>` after this removal?

nlfurniss: [This bit of code](https://github.com/emberjs/ember.js/blob/master/packages/@ember/-internals/glimmer/lib/components/internal.ts#L256) causes any test I write to fail due to `<unsupportedProp> is deprecated` so I think it's pretty well aserted

nlfurniss: Good call. I added an `assert` to the `validateArguments` method and some tests that assert for this error, but not sure if that's the right place to throw versus where [`@href` throws](https://github.com/emberjs/ember.js/blob/master/packages/@ember/-internals/glimmer/lib/components/link-to.ts#L404-L419)

rwjblue (approved):

rwjblue:
Needs a rebase (one of the other PR's must have caused a conflict).

packages/@ember/-internals/glimmer/lib/components/link-to.ts
```diff
@@ -80,6 +80,11 @@ class LinkTo extends InternalComponent implements DeprecatingInternalComponent {
       !('model' in this.args.named && 'models' in this.args.named)
     );
 
+    assert(
+      'Passing the `@tagName` argument to <LinkTo> is not supported.',
+      !('tagName' in this.args.named)
+    );
```
mixonic, line 86: Two concerns:

With this implemented as an assert, what does that mean for production? Would a `@tagName` be applied in production, even if it the assert stopped you in development? Seems bad if so. Maybe it should be a `throw`.

But, further, IMO we don't need a check at all. I don't think we're planning to add asserts for every old component argument which was removed in 4.0. They were removed, which means passing one of those arguments (like `@tagName`) could just mean it is ignored. I can ask around on this second point, but the first concern about production behavior *not* applying the tag name needs to be confirmed.

mixonic:
FYI @nlfurniss I rebased and force pushed this branch.

mixonic (changes requested):
Per https://github.com/emberjs/ember.js/pull/19669#issuecomment-884286415 we should hold off on this until the built-in components are extracted. @chancancode has kicked off that work I believe.

packages/@ember/-internals/glimmer/lib/components/link-to.ts
nlfurniss: Yup, makes sense re: `assert`. As to the second point, classic components will keep `tagName`, so it seems like something special will need to be done for `linkTo`.

mixonic:
Supplanted by https://github.com/emberjs/ember.js/pull/19806. Thank you @nlfurniss !

PR #19666, opened 2021-07-19
[CLEANUP] Remove jQuery integration in EventDispatcher

simonihmig:
Reopened #19552


packages/@ember/-internals/glimmer/tests/integration/components/input-angle-test.js
```diff
@@ -557,18 +557,13 @@ moduleFor(
...
         this.render(`<Input @enter="foo" />`, {
           actions: {
-            foo(value, event) {
+            foo() {
               assert.ok(true, 'action was triggered');
-              if (jQueryDisabled) {
-                assert.notOk(event.originalEvent, 'event is not a jQuery.Event');
-              } else {
-                assert.ok(event instanceof jQuery.Event, 'jQuery event was passed');
-              }
```
mixonic, line 571: This test will be deleted in later clean up so this seems fine.

packages/@ember/-internals/glimmer/tests/integration/components/input-angle-test.js
```diff
@@ -582,17 +577,12 @@ moduleFor(
...
       this.render(`<Input @enter={{action 'foo'}} />`, {
         actions: {
-          foo(value, event) {
+          foo() {
             assert.ok(true, 'action was triggered');
-            if (jQueryDisabled) {
-              assert.notOk(event.originalEvent, 'event is not a jQuery.Event');
-            } else {
-              assert.ok(event instanceof jQuery.Event, 'jQuery event was passed');
-            }
```
mixonic, line 595: Can we assert the presence of `event` here and `event instanceof Event` or something?

I believe these two comments apply across the rest of this file and some other files. In the test which isn't for deprecated behavior, please assert the `event` argument is passed.

mixonic (changes requested):

packages/@ember/-internals/glimmer/tests/integration/components/input-angle-test.js
simonihmig: Not only does this totally make sense, but trying to add that assertion uncovered an actual bug, as the event passed *was* a jQuery event. 😱

The internal components seem do use some additional jQuery related magic, that I was not aware of: https://github.com/emberjs/ember.js/blob/3ce13cea235cde8a87d89473533c453523412764/packages/%40ember/-internals/glimmer/lib/components/internal.ts#L556-L576

So this will need some more cleanup...

rwjblue:
FYI - I moved this into draft so I can keep track of where review is needed vs where more work is in progress, please move back out of draft when its ready to review again.

simonihmig:
Pushed an update:
* rebased
* added back assertions for (native) events (addressing https://github.com/emberjs/ember.js/pull/19666#discussion_r672718928)
* removed `jQueryEventShim`

Note: there are still a few jQuery-related leftovers, like the deprecated feature flag, which are not directly related to the narrower scope of this PR, and which I plan to address in a follow-up PR when this is merged.

mixonic (approved):

PR #19575, opened 2021-05-28
[CLEANUP] Cleanup IE11 leftovers

simonihmig:
Further cleanup after dropping IE11 support: (best reviewed commit after commit)
* Use WeakSet for lazy events
* Remove Object.entries/.values polyfills
* Remove HAS_NATIVE_SYMBOL and HAS_NATIVE_PROXY flags (as all supported browser support these)

PR #19667, opened 2021-07-19
[CLEANUP] Cleanup IE11 leftovers

simonihmig:
Reopened #19575

### DIFF
--- a/packages/@ember/-internals/routing/lib/services/router.ts
+++ b/packages/@ember/-internals/routing/lib/services/router.ts
@@ -10,7 +10,7 @@ import Route from '../system/route';
 import EmberRouter, { QueryParam } from '../system/router';
 import { extractRouteArgs, resemblesURL, shallowEqual } from '../utils';
 
-const ROUTER = symbol('ROUTER') as string;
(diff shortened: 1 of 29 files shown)


packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -10,7 +10,7 @@ import Route from '../system/route';
 import EmberRouter, { QueryParam } from '../system/router';
 import { extractRouteArgs, resemblesURL, shallowEqual } from '../utils';
 
-const ROUTER = symbol('ROUTER') as string;
+const ROUTER = (symbol('ROUTER') as unknown) as string;
```
simonihmig, line 13: Following up on https://github.com/emberjs/ember.js/pull/19575#discussion_r671691460:

I get this TS error in my branch without it:

image

This comes from the change in https://github.com/emberjs/ember.js/pull/19667/files#diff-8519bc32ded2c3feefab86f01e1308398bacc99f2c10d4b28d372c0853d33203. The type of the exported `symbol` changed there, as the return type of the function is now always a native `symbol` (was `string | symbol` before).

mixonic:
Thank you @simonihmig !

PR #19669, opened 2021-07-21
Remove deprecated disabledWhen

nlfurniss:
Resurrected this from the previous PR tracking v4-cleanup.

Part of #19617

All `link-to` tests, including ones setting `disabled` pass locally.


rwjblue (approved):

chancancode:
Eh, I was just about to comment we probably want to wait on all of these features related to link-to and do them together. The idea is we should be able to just remove `-link-to` (and likewise the old implementation for other built-in components) altogether and be done with it, since the new implementation didn't include any of these deprecated features as you have discovered. The issue is, we are supposed to move the old implementations into https://github.com/emberjs/ember-legacy-built-in-components. Right now, the add-on simply reaches into the internals and re-export those old implementations (the one you just modified) so this will break the addon. I don't remember if this particular feature is supposed to keep working on the legacy addon, but we probably don't want to touch them at all until someone has time to finish moving the implementation into the addon. I'd suggest that we revert this and hold off on removing the deprecations for legacy built-in components for now.

PR #19670, opened 2021-07-22
Remove .volatile()

nlfurniss:
Part of #19617

[Deprecation guide](https://deprecations.emberjs.com/v3.x#toc_computed-property-volatile)

Tests pass locally

### DIFF
--- a/packages/@ember/-internals/runtime/tests/legacy_1x/mixins/observable/observable_test.js
+++ b/packages/@ember/-internals/runtime/tests/legacy_1x/mixins/observable/observable_test.js
@@ -298,120 +298,83 @@ moduleFor(
     beforeEach() {
       lookup = context.lookup = {};
 
-      expectDeprecation(() => {
-        object = ObservableObject.extend({
-          computed: computed({
-            get() {
-              this.computedCalls.push('getter-called');
-              return 'computed';
-            },
-            set(key, value) {
-              this.computedCalls.push(value);
-            },
-          }).volatile(),
+      object = ObservableObject.extend({
+        computed: computed({
+          get() {
+            this.computedCalls.push('getter-called');
+            return 'computed';
+          },
+          set(key, value) {
+            this.computedCalls.push(value);
+          },
+        }),
 
-          computedCached: computed({
-            get() {
-              this.computedCachedCalls.push('getter-called');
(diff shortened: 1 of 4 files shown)


packages/@ember/-internals/runtime/tests/legacy_1x/mixins/observable/observable_test.js
```diff
@@ -399,13 +397,13 @@ moduleFor(
...
-      // verify each call count.  cached should only be called once
+      // verify each call count. cached should only be called once
       w('computedCalls dependentFrontCalls dependentCalls').forEach((key) => {
-        assert.equal(object[key].length, 2, `non-cached property ${key} should be called 2x`);
+        assert.equal(object[key].length, 1, `non-cached property ${key} should be called 1x`);
       });
 
       w('computedCachedCalls dependentCachedCalls').forEach((key) => {
-        assert.equal(object[key].length, 1, `non-cached property ${key} should be called 1x`);
+        assert.equal(object[key].length, 1, `cached property ${key} should be called 1x`);
```
mixonic, line 406: This test can be dramatically simplified. The volatile CPs, after the removal of the volatility, are simply plain CPs with cached semantics. There is not need to test those same semantics twice.

Similarly I think `dependentFrontCalls` and `dependentCalls` were at one time differentiated by the argument order to `computed(`. Now there is only one order, so there is no need to assert the same thing twice.

Just take a look at the CPs after your patch. I'm pretty sure they now repeat themselves a few times, just with different inherited-from-old-code variable names.

mixonic (changes requested):

packages/@ember/-internals/runtime/tests/legacy_1x/mixins/observable/observable_test.js
nlfurniss: Sounds good. I left it as is because I wasn’t sure what the difference between some of the test CPs were so erred on the side of caution

nlfurniss: Ok, done (and I think I removed the correct ones 🤞)

mixonic (approved):
Awesome! Thank you!

PR #19671, opened 2021-07-22
Remove .property()

nlfurniss:
Part of #19662

[Deprecation guide](https://deprecations.emberjs.com/v3.x#toc_computed-property-property)

Tests pass locally

### DIFF
--- a/packages/@ember/-internals/metal/lib/computed.ts
+++ b/packages/@ember/-internals/metal/lib/computed.ts
@@ -734,81 +734,6 @@ class ComputedDecoratorImpl extends Function {
   }
 
   /**
-    Sets the dependent keys on this computed property. Pass any number of
-    arguments containing key paths that this computed property depends on.
-
-    Example:
-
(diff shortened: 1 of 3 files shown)


packages/@ember/-internals/metal/lib/computed.ts
```diff
@@ -733,81 +733,6 @@ class ComputedDecoratorImpl extends Function {
...
-        id: 'computed-property.property',
-        until: '4.0.0',
-        url: 'https://deprecations.emberjs.com/v3.x#toc_computed-property-property',
-        for: 'ember-source',
-        since: {
-          enabled: '3.9.0-beta.1',
-        },
-      }
-    );
-    (descriptorForDecorator(this) as ComputedProperty)._property(...keys);
```
mixonic, line 807: Not a blocker, but `_property` could be refactored to be more strictly internal now

mixonic (approved):

PR #19673, opened 2021-07-22
Remove computed deep each

nlfurniss:
Part of #19662
`ember-metal.computed-deep-each`
Tests pass locally

The code in `getChainTags` was deleted because it seems that that condition won't be reached since setting a deep dependent key will throw an error before it can be `get`ed.

### DIFF
--- a/packages/@ember/-internals/metal/lib/chain-tags.ts
+++ b/packages/@ember/-internals/metal/lib/chain-tags.ts
@@ -1,6 +1,6 @@
 import { Meta, meta as metaFor, peekMeta } from '@ember/-internals/meta';
 import { isObject } from '@ember/-internals/utils';
-import { assert, deprecate } from '@ember/debug';
+import { assert } from '@ember/debug';
 import { _WeakSet } from '@glimmer/util';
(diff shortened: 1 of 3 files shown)


mixonic (commented):
@pzuraq This could use your eyes. I'm not sure which site should be changed to raise the error, and if the error should be a hard `throw` or simply a dev-only `assert`. As this code is performance sensitive, dev-only seems fine by my own intuition, but I'd like more feedback.

pzuraq:
This should be a dev-only assert, definitely. I believe the location of the new error is correct though, it should catch both cases and properly prevent them (since the `each` case is a special case of the nested chaining case)

nlfurniss:
> This should be a dev-only assert, definitely. I believe the location of the new error is correct though, it should catch both cases and properly prevent them (since the `each` case is a special case of the nested chaining case)

Sounds good, updated

mixonic (approved):

mixonic:
This looks great to me, I pushed a rebase of the branch to address a conflict and will merge.

PR #19674, opened 2021-07-22
Remove ability to override computed property

nlfurniss:
Part of #19617. [Deprecation guide](https://deprecations.emberjs.com/v3.x#toc_computed-property-override)

Most tests seem fine, though not sure if [this one ](https://github.com/emberjs/ember.js/compare/master...nlfurniss%3Aremove-computed-property-override?body=&expand=1&title=Remove+ability+to+override+computed+property#diff-ab4aedbf885726372cc81f2ca06640b26c5a895d25354717454e972131a3caa4R966)still has value..

### DIFF
--- a/packages/@ember/-internals/metal/lib/computed.ts
+++ b/packages/@ember/-internals/metal/lib/computed.ts
@@ -29,14 +29,12 @@ import {
 } from './decorator';
 import expandProperties from './expand_properties';
 import { addObserver, setObserverSuspended } from './observer';
-import { defineProperty } from './properties';
 import {
   beginPropertyChanges,
   endPropertyChanges,
   notifyPropertyChange,
   PROPERTY_DID_CHANGE,
 } from './property_events';
-import { set } from './property_set';
 
 export type ComputedPropertyGetter = (keyName: string) => any;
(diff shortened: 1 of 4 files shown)


packages/@ember/-internals/metal/lib/computed.ts
```diff
@@ -433,7 +431,7 @@ export class ComputedProperty extends ComputedDescriptor {
     }
 
     if (!this._setter) {
-      return this.clobberSet(obj, keyName, value);
+      assert(`Cannot override the computed property \`${keyName}\` on ${toString(obj)}.`);
     }
```
mixonic, line 435:
```js
assert(`Cannot override the computed property \`${keyName}\` on ${toString(obj)}.`, this._setter !== undefined);
```

Is likely preferred, as it would cleanly strip the `this._setter` access out of production builds.

mixonic (changes requested):

packages/@ember/-internals/metal/lib/computed.ts
nlfurniss: yeah, dunno what I was thinking… 😅

PR #19675, opened 2021-07-22
[CLEANUP beta] Remove jQuery usage from ember-testing

simonihmig:
As advised by feedback from core team, this removes all jQuery related testing helpers from ember-testing, along with related utilities (event stuff) and tests:
* `find`
* `findWithAssert`
* `triggerEvent`
* `keyEvent`
* `click`
* `fillIn`

All other helpers remain untouched for future official deprecation.

---

~This is meant as a prerequisite to further jQuery cleanup, like removing the internal `jQueryDisabled` flag.~

~Basically I am assuming that in 4.0 `jQueryDisabled` is implicitly always true (similar to assuming `JQUERY_INTEGRATION` is always false). So here I am basically doing manual dead code elimination. However that means:~
* ~ember-testing's `find()` is always throwing an assertion now, and as such probably most other helpers that internally use `find()`?~
* ~some test suites have been completely removed, as all of it was wrapped in `if (!jQueryDisabled) { ... }` (so under that above assumption that is `if (false) {}` now)~

~Note sure how this whole package was meant to migrate to 4.0, so please provide advice. AFAIK the legacy testing API's (before `@ember/test-helpers`) weren't formally deprecated, were they? But nowadays without jQuery, they seem not even usable anymore?~

~/cc @rwjblue~


mixonic:
Hm, @simonihmig I believe the testing APIs should be removed if they are now always erring. They are replaced by https://github.com/emberjs/ember-test-helpers.

If that is acceptable for API that error, I think another question is what happens to all the non-erroring helpers offered in the same manner. Those might need deprecation targeting 5.0 removal. I'll add this to our framework discussion Friday.

mixonic:
@simonihmig We discussed this with framework core. The group agreed that any testing APIs that were reliant on jQuery should simply be removed in 4.0. If there are testing APIs that do not reference jQuery, they should remain. Separately we could work on deprecating those, but it would target 5.0 at this point.

simonihmig:
> The group agreed that any testing APIs that were reliant on jQuery should simply be removed in 4.0. If there are testing APIs that do not reference jQuery, they should remain. 

@mixonic Did that now! See the PR's updated description. Ready for review then!

snewcomer (approved):
This is wonderful!!

mixonic (approved):
I pushed an amended PR which drops changes to the `yarn.lock` file that I believe were unrelated.

simonihmig:
> I pushed an amended PR which drops changes to the yarn.lock file that I believe were unrelated.

Oh, didn't notice, thanks! I believe that happened to me more than once, idk why, maybe different versions of `yarn`? 
I could create a PR to add the `volta` config to `package.json`, so at least volta users would use the same deterministic versions, does that make sense?

PR #19676, opened 2021-07-22
[CLEANUP beta] Remove internal jQuery usage

simonihmig:
Removes leftover internal jQuery references, and `JQUERY_INTEGRATION` and `jQueryDisabled` flags.

~Should be passing once #19666 and #19675 (or some other PR that removes usage of `jQueryDisabled` from `ember-testing`) have landed.~


simonihmig:
@mixonic rebased and ready for review

packages/@ember/application/lib/application.js
```diff
@@ -1170,14 +1161,4 @@ function commonSetupRegistry(registry) {
   registry.register('service:router', RouterService);
 }
 
-function registerLibraries() {
-  if (!librariesRegistered) {
-    librariesRegistered = true;
-
-    if (JQUERY_INTEGRATION && hasDOM && !jQueryDisabled) {
-      libraries.registerCoreLibrary('jQuery', jQuery().jquery);
```
mixonic, line 1178: Slightly unrelated, but is `libraries` really used for anything after this removal? I don't think it is.

mixonic:
@kategengler note, I belatedly updated the title here to include `[CLEANUP beta]`

kategengler:
👍 As long as it is a recent merged and title is changed before the next beta, I'll catch any updated tags. I look through both commits & PRs due to inconsistency.

packages/@ember/application/lib/application.js
simonihmig: Hm, you mean the whole `libraries` thing, or just the import here? The import is used here for logging the Ember and possibly other registered addon versions to the console. And for that `libraries.register()` is still used in addons (I am using it [here](https://github.com/kaliber5/ember-bootstrap/blob/master/vendor/ember-bootstrap/register-version.template) for example, but this reminds me I actually have to refactor this to not use the `Ember` global!) and [ember-data itself](https://github.com/emberjs/data/blob/master/packages/-ember-data/addon/-private/core.js#L12). So I think this has to stay!?

PR #19677, opened 2021-07-22
[CLEANUP beta] Remove jQuery from build

simonihmig:
Removes jQuery from dependencies, the build process and test runs.

Should be passing once the other open jQuery related PRs have landed (#19666, #19675, #19676)

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -3,7 +3,6 @@
 const MergeTrees = require('broccoli-merge-trees');
 const Funnel = require('broccoli-funnel');
 const path = require('path');
-const resolve = require('resolve');
 const concatBundle = require('./concat-bundle');
 const buildDebugMacroPlugin = require('./build-debug-macro-plugin');
 const buildStripClassCallcheckPlugin = require('./build-strip-class-callcheck-plugin');
@@ -33,7 +32,6 @@ function add(paths, name, path) {
 add(paths, 'prod', 'vendor/ember/ember.js');
 add(paths, 'debug', 'vendor/ember/ember.js');
 add(paths, 'testing', 'vendor/ember/ember-testing.js');
-add(paths, 'jquery', 'vendor/ember/jquery/jquery.js');
 
 add(
   absolutePaths,
@@ -97,25 +95,6 @@ module.exports = {
 
       if (
         optionalFeaturesMissing ||
-        typeof optionalFeatures.isFeatureExplicitlySet !== 'function'
-      ) {
-        message.push(
-          '* Unable to detect if jquery-integration is explicitly set to a value, please update `@ember/optional-features` to the latest version'
-        );
-      }
-
-      if (
-        optionalFeaturesMissing ||
-        (typeof optionalFeatures.isFeatureExplicitlySet === 'function' &&
-          !optionalFeatures.isFeatureExplicitlySet('jquery-integration'))
-      ) {
-        message.push(
-          `* The jquery-integration optional feature should be explicitly set to a value under Octane, run \`ember feature:disable jquery-integration\` to disable it, or \`ember feature:enable jquery-integration\` to explicitly enable it`
-        );
-      }
-
-      if (
-        optionalFeaturesMissing ||
         optionalFeatures.isFeatureEnabled('application-template-wrapper')
       ) {
         message.push(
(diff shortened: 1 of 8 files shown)


nlfurniss:
@simonihmig this diff should fix the eslint issues and after rebasing the tests should pass as well

```diff
diff --git a/ember-cli-build.js b/ember-cli-build.js
index cdf3daf58..5c92e7a61 100644
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -14,7 +14,6 @@ Error.stackTraceLimit = Infinity;
 
 const {
   routerES,
-  jquery,
   loader,
   qunit,
   handlebarsES,
diff --git a/lib/index.js b/lib/index.js
index 19bf17a76..778268f56 100644
--- a/lib/index.js
+++ b/lib/index.js
@@ -3,7 +3,6 @@
 const MergeTrees = require('broccoli-merge-trees');
 const Funnel = require('broccoli-funnel');
 const path = require('path');
-const resolve = require('resolve');
 const concatBundle = require('./concat-bundle');
 const buildDebugMacroPlugin = require('./build-debug-macro-plugin');
 const buildStripClassCallcheckPlugin = require('./build-strip-class-callcheck-plugin');
```

simonihmig:
> this diff should fix the eslint issues and after rebasing the tests should pass as well

Yeah, thanks @nlfurniss! Ready for Review! @mixonic

lib/index.js
```diff
@@ -165,11 +144,12 @@ module.exports = {
...
 
     this._jqueryIntegrationEnabled =
-      optionalFeaturesMissing || optionalFeatures.isFeatureEnabled('jquery-integration');
+      !optionalFeaturesMissing && optionalFeatures.isFeatureEnabled('jquery-integration');
 
     if (this._jqueryIntegrationEnabled) {
-      this.ui.writeWarnLine(
-        'Setting the `jquery-integration` optional feature flag to `true`, or not providing a setting at all, has been deprecated. You must add the `@ember/optional-features` addon and set this feature to `false`. This warning will become an error in Ember 4.0.0.\n\nFor more information, see the deprecation guide: https://deprecations.emberjs.com/v3.x/#toc_optional-feature-jquery-integration'
+      const SilentError = require('silent-error');
+      throw new SilentError(
```
mixonic, line 151: This is the right thing to do? Or should we assert if anyone *sets* this feature enabled/disabled at all? As in, they should just remove it from their configuration.

The approach of require `false` is certainly more gentle, but a new 4.0 user would probably look curiously at a requirement to set a feature to disabled which... could not possibly be enabled.

mixonic (commented):
Close! I kicked off a rebuild to get around browserstack issues. This is very exciting.

lib/index.js
simonihmig: Hm, right, I actually meant to only throw when this is explicitly set to true. I (wrongly) assumed this is already false by default, but [it is not](https://github.com/emberjs/ember-optional-features/blob/master/features/jquery-integration.js#L7). So I will need to revisit this, to make it match [RFC705 Phase 2](https://emberjs.github.io/rfcs/0705-deprecate-jquery-optional-feature.html#phase-2). So I would basically use [`isFeatureExplicitlySet`](https://github.com/emberjs/ember-optional-features/blob/master/index.js#L71) to check if is explicitly set to true, or do noting (optional feature has no effect) otherwise, right?

simonihmig:
> I kicked off a rebuild to get around browserstack issues

Seems FF and Browserstack tests are still failing, so something *real*? 🤔 Wouldn't know why though. [`libEGL missing`](https://github.com/emberjs/ember.js/pull/19677/checks?check_run_id=3391265210#step:7:123), does that ring a bell for you/someone?

lib/index.js
simonihmig: I have addressed this now!

tests/index.html
```diff
@@ -134,7 +126,7 @@
           testsTotal = testsPassed = testsFailed = 0;
 
           if (QUnit.urlParams.hideskipped) {
-            $('#qunit-tests').addClass('hideskipped');
+            document.getElementById('qunit-tests').classList.add('hideskipped');
```
simonihmig, line 129: This was the reason for failing FF and Browserstack tests! They used `hideskipped` in `testem.ci-browsers.js`, which was still jQuery based here. Still don't understand why this gave us these obscure error messages...

simonihmig:
This took too long to make it green, sorry about that. But it finally passes all tests! :tada:

I was able to reproduce the failing Firefox CI job failure and fix it, see the inline comments... so @mixonic this is ready now for final review!

lib/index.js
```diff
@@ -164,12 +142,15 @@ module.exports = {
...
-        'Setting the `jquery-integration` optional feature flag to `true`, or not providing a setting at all, has been deprecated. You must add the `@ember/optional-features` addon and set this feature to `false`. This warning will become an error in Ember 4.0.0.\n\nFor more information, see the deprecation guide: https://deprecations.emberjs.com/v3.x/#toc_optional-feature-jquery-integration'
+    if (
+      !optionalFeaturesMissing &&
+      optionalFeatures.isFeatureEnabled('jquery-integration') &&
+      typeof optionalFeatures.isFeatureExplicitlySet === 'function' &&
+      optionalFeatures.isFeatureExplicitlySet('jquery-integration')
+    ) {
+      const SilentError = require('silent-error');
+      throw new SilentError(
+        'Setting the `jquery-integration` optional feature flag to `true` was deprecated in Ember 3.x and removed in Ember 4.0.0. You must add the `@ember/optional-features` addon and set this feature to `false`.\n\nFor more information, see the deprecation guide: https://deprecations.emberjs.com/v3.x/#toc_optional-feature-jquery-integration'
```
rwjblue, line 153: I think this prose needs work now that the logic has changed. We don't want folks installing the optional features addon at all, we want them to remove the configuration from their `optional-features.json` file...

tests/index.html
rwjblue: Good catch!!!

rwjblue (commented):
Thanks @simonihmig! Only one minor comment left...

mixonic (approved):

PR #19678, opened 2021-07-23
Remove window.Ember global

mixonic:

### DIFF
--- a/tests/node/fastboot-sandbox-test.js
+++ b/tests/node/fastboot-sandbox-test.js
@@ -43,7 +43,7 @@ function buildSandboxContext(precompile) {
     URL,
 
     // Convince jQuery not to assume it's in a browser
-    module: { exports: {} },
+    module: { exports: {}, require() {} },
   };
 
   // Set the global as `window`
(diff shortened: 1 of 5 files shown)


tests/node/fastboot-sandbox-test.js
```diff
@@ -43,7 +43,7 @@ function buildSandboxContext(precompile) {
     URL,
 
     // Convince jQuery not to assume it's in a browser
-    module: { exports: {} },
+    module: { exports: {}, require() {} },
```
mixonic, line 46: This is maybe the most suspicious change in here. As these tests are simply executing scripts in a context, they don't directly use any module system (CommonJS or ES). Adding `require() {}` to the exports fools Ember into thinking this is a CommonJS env (which is effectively what this line was doing anyway) and thus Ember sets `.exports` to be `Ember`. This hack only works for one this-which-exports-via-CommonJS in a context.

I'm not familiar enough with how this code runs in Fastboot to know if this is a meaningful test. cc @rwjblue

mixonic: @rwjblue suggested this should land, and that if Fastboot as a follow up we can address it during the beta.

kategengler (approved):

PR #19680, opened 2021-07-25
[CLEANUP beta] Deprecate owner.inject

snewcomer:
RFC: https://github.com/emberjs/rfcs/pull/680/

See https://github.com/emberjs/rfcs/blob/sn/owner-inject-deprecation/text/0680-implicit-injection-deprecation.md#1-deprecate-implicit-injection-on-target-object for the reasons why we first added the initial deprecation in `3.26.0` and are now adding another deprecation but removing the functionality of `app.inject`.

- [x] deprecate ability to inject properties on Ember framework objects
- [x] Remove infrastructure supporting `injection`
- [x] remove deprecations ending in 4.0.0 from #19358 
- [x] remove blueprint inject function
- [x] remove all implicit injections in codebase

ref https://github.com/emberjs/ember.js/issues/19617


mixonic:
@snewcomer can the cleanup/removal of the old API land in its own PR with `[CLEANUP beta]`? I worry this has entangled the removals with the addition of a deprecation, but they should not be coupled.

snewcomer:
@mixonic So this timeline took me a bit to grep but is laid out in this [RFC](https://github.com/emberjs/rfcs/blob/sn/owner-inject-deprecation/text/0680-implicit-injection-deprecation.md#detailed-design) .

1. >= 3.26 deprecated the use of implicit injections.  See https://deprecations.emberjs.com/v3.x#toc_implicit-injections.  So by now, everybody who is using implicit injections has been warned and hopefully made adjustments (including users of ember-data who relied on the implicit `store` deprecation).
2. 4.x deprecates `owner.inject`.  In addition, we will remove the supporting infrastructure to manage injections.  `inject` is now effectively now a noop. 
3. 5.x will remove the injection function and supporting code.

Sry if the above was rehashing something you already know.  But the deprecation + removal of the supporting infrastructure for `inject` go hand in hand.

packages/@ember/-internals/glimmer/tests/integration/outlet-test.js
```diff
@@ -9,8 +9,9 @@ moduleFor(
       super(...arguments);
 
       let CoreOutlet = this.owner.factoryFor('view:-outlet');
-
-      this.component = CoreOutlet.create();
+      let outletTemplateFactory = this.owner.lookup('template:-outlet');
+      let environment = this.owner.lookup('-environment:main');
+      this.component = CoreOutlet.create({ environment, template: outletTemplateFactory });
```
snewcomer, line 14: ☝️ this is important to note.  This was refactored from the implicit injection.

packages/@ember/-internals/container/lib/registry.ts
```diff
@@ -468,116 +447,32 @@ export default class Registry implements IRegistry {
...
    @param {String} injectionName
+   @deprecated
    */
-  injection(fullName: string, property: string, injectionName: string) {
-    assert(
-      `Invalid injectionName, expected: 'type:name' got: ${injectionName}`,
-      this.isValidFullName(injectionName)
+  injection(fullName: string, property: string) {
+    deprecate(
+      `owner.inject has been deprecated and will be removed in v5.0.0. Please refactor to explicitly instantiate ${property} on ${fullName}.`,
```
pzuraq, line 464: We should add a note that this method no longer does anything, and is safe to remove. Something like:

```suggestion
      `As of Ember 4.0.0, owner.inject no longer injects values into resolved instances, and calling the method has been deprecated. Since this method no longer does anything, it is fully safe to remove this injection. As an alternative to this API, you can refactor to explicitly instantiate ${property} on ${fullName}, or look it up directly using the \`getOwner\` API.`,
```

packages/@ember/-internals/glimmer/lib/renderer.ts
```diff
@@ -290,14 +291,15 @@ export class Renderer {
...
-    let { document, env, rootTemplate, _viewRegistry, builder } = props;
+  static create(props: { _viewRegistry: any }): Renderer {
+    let { _viewRegistry } = props;
+    let document = getOwner(props).lookup('service:-document') as SimpleDocument;
+    let env = getOwner(props).lookup('-environment:main') as {
+      isInteractive: boolean;
+      hasDOM: boolean;
+    };
+    let rootTemplate = getOwner(props).lookup(P`template:-root`) as TemplateFactory;
+    let builder = getOwner(props).lookup('service:-dom-builder') as IBuilder;
```
pzuraq, line 302:
```suggestion
    let owner = getOwner(props);
    let rootTemplate = owner.lookup(P`template:-root`) as TemplateFactory;
    let builder = owner.lookup('service:-dom-builder') as IBuilder;
```

Better to do the lookup once rather than multiple times

pzuraq (approved):
So I believe that `_lazyInjections` can also be removed as part of this work. As far as I can tell, it's basically creating and setting up lazy versions of implicit injections, only in DEBUG, and I can't see any usage in the ecosystem at all: https://emberobserver.com/code-search?codeQuery=_lazyInjections

It's explicitly marked as a private API as well, so should be fine to remove. Overall this PR looks amazing and once those changes are made, I think we're good to merge!

tests/docs/expected.js
```diff
@@ -32,7 +32,6 @@ module.exports = {
     '_helpers',
     '_hydrateUnsuppliedQueryParams',
     '_initializersRan',
-    '_injections',
     '_internalReset',
     '_invoke',
     '_lazyInjections',
```
snewcomer, line 37: I checked out `lazyInjections` and it looks like it is used to validate service injections and not implicit.

https://github.com/emberjs/ember.js/blob/b818f5408836867bdf725583126c994080eb0bf1/packages/%40ember/-internals/glimmer/tests/integration/components/curly-components-test.js#L2898

mixonic:
@snewcomer I don't understand the statement:

> 4.x deprecates owner.inject. In addition, we will remove the supporting infrastructure to manage injections. inject is now effectively now a noop.

Deprecated functionality is not removed, it is merely warned against in development builds.

Do you mean "if a user were to use an explicit injection as suggested by the deprecation message, then any `owner.inject` based type-based injection for the same property could be safely considered a no-op and removed" or some other phrasing?

packages/@ember/-internals/container/lib/registry.ts
```diff
@@ -468,116 +447,32 @@ export default class Registry implements IRegistry {
...
    @param {String} injectionName
+   @deprecated
    */
-  injection(fullName: string, property: string, injectionName: string) {
-    assert(
-      `Invalid injectionName, expected: 'type:name' got: ${injectionName}`,
-      this.isValidFullName(injectionName)
+  injection(fullName: string, property: string) {
+    deprecate(
+      `As of Ember 4.0.0, owner.inject no longer injects values into resolved instances, and calling the method has been deprecated. Since this method no longer does anything, it is fully safe to remove this injection. As an alternative to this API, you can refactor to explicitly instantiate ${property} on ${fullName}, or look it up directly using the \`getOwner\` API.`,
```
mixonic, line 464:
```suggestion
      `As of Ember 4.0.0, owner.inject no longer injects values into resolved instances, and calling the method has been deprecated. Since this method no longer does anything, it is fully safe to remove this injection. As an alternative to this API, you can refactor to explicitly inject \`${property}\` on \`${fullName}\`, or look it up directly using the \`getOwner\` API.`,
```

?

snewcomer: That is a good suggestion.  Some cases are `inject`; some cases are `instantiate`.  But I think the second half of the sentence covers `instantiate`.

mixonic: Annoyingly I think it may make the suite fail :-p

Great work here :muscle:

packages/@ember/-internals/container/lib/registry.ts
```diff
@@ -468,116 +447,32 @@ export default class Registry implements IRegistry {
...
-      `Invalid injectionName, expected: 'type:name' got: ${injectionName}`,
-      this.isValidFullName(injectionName)
+  injection(fullName: string, property: string) {
+    deprecate(
+      `As of Ember 4.0.0, owner.inject no longer injects values into resolved instances, and calling the method has been deprecated. Since this method no longer does anything, it is fully safe to remove this injection. As an alternative to this API, you can refactor to explicitly inject \`${property}\` on \`${fullName}\`, or look it up directly using the \`getOwner\` API.`,
+      false,
+      {
+        id: 'remove-owner-inject',
+        until: '5.0.0',
+        url: 'https://deprecations.emberjs.com/v3.x#toc_implicit-injections',
```
mixonic, line 469: This should get a 4.x deprecation entry. The 3.x one is much more information than is probably needed, explains how to move away from APIs which the user already shouldn't be using, and is just confusing as it is in the 3.x series.

Screenshot from 2021-08-19 21-36-13

mixonic (changes requested):

packages/@ember/-internals/container/lib/registry.ts
snewcomer: https://github.com/ember-learn/deprecation-app/pull/932

mixonic: If you can update the link here to point to the correct 4.x deprecation page we can land this PR independent of https://github.com/ember-learn/deprecation-app/pull/932, IMO

snewcomer: ok fixed!

mixonic (approved):
Awesome stuff here @snewcomer !

mixonic:
Force pushed a single commit for a clean history, letting CI run.

packages/@ember/-internals/glimmer/tests/integration/outlet-test.js
stefanpenner: I believe this  needs to be ported to `@ember/test-helpers` https://github.com/emberjs/ember-test-helpers/blob/9219247259ef6b4d8a339ca71a1374e2ffe69603/addon-test-support/%40ember/test-helpers/setup-rendering-context.ts#L266

stefanpenner:
I believe this currently incompatible with `@ember/test-helpers`... I'm going to try and [add the capability](https://github.com/emberjs/ember.js/pull/19680#discussion_r694190928) to `@ember/test-helpers` now, but it will require folks upgrade.

packages/@ember/-internals/glimmer/tests/integration/outlet-test.js
stefanpenner: working on that now.

snewcomer:
@stefanpenner I would be happy to add this.  I love adding this type of regression testing!

packages/@ember/-internals/glimmer/tests/integration/outlet-test.js
stefanpenner: https://github.com/emberjs/ember-test-helpers/pull/1110

stefanpenner:
@snewcomer awesome. At the very-least, I think ember could get lots of milage out of smoke testing `@ember/test-helpers`.

stefanpenner:
proposed fix for `@ember/test-helpers` https://github.com/emberjs/ember-test-helpers/pull/1110

stefanpenner:
@snewcomer also, if you have cycles a small inline comment in ember.js, specifically around https://github.com/emberjs/ember.js/pull/19680#discussion_r690022310 to inform authors that changes need to be made in parallel with `@ember/test-helpers` would be dope.

Ultimately, having ember provide a "public api" for test-helpers to tap into here could prove to be the best long-term solution. But may not be the most pragmatic choice right now.

packages/@ember/-internals/glimmer/tests/integration/outlet-test.js
stefanpenner: landed and released as @ember/test-helpers@2.4.2

stefanpenner:
Fix for my above issue landed as @ember/test-helpers@2.4.2

snewcomer:
@stefanpenner This PR to smoke test ember-test-helpers, ember-string and ember-legacy-built-in-components is passing now!  Would appreciate a review if you have some time.

https://github.com/emberjs/ember.js/pull/19728

PR #19681, opened 2021-07-26
[BUGFIX release] Restore previous hash behavior

pzuraq:
Restores the previous hash behavior where `hash` will lazily update in
templates, but eagerly evaluate and create a POJO in JS. Also preserves
the deprecation on setting on the hash.


rwjblue:
Thanks for working on this! Can you cross link the related issues / PR's that this essentially reverts / tweaks (for easier archeology in the future)?

PR #19682, opened 2021-07-29
[DOC] Clarify invokeHelper example

LevelbossMike:
The `invokeHelper`-docs don't explain that one has to create a getter that accesses the helper's value via `getValue` from `@glimmer/tracking/primitives/cache`.

I only realised this was by design after looking at the tests here - https://github.com/emberjs/ember.js/blob/21bd70c773dcc4bfe4883d7943e8a68d203b5bad/packages/%40ember/-internals/glimmer/tests/integration/helpers/invoke-helper-test.js#L34

The current docs state that it was possible to access an invoked helper's value by accessing it via `.value` which does not work in current Ember.

### DIFF
--- a/packages/@ember/helper/index.ts
+++ b/packages/@ember/helper/index.ts
@@ -252,14 +252,18 @@
   The `invokeHelper` function can be used to create a helper instance in
   JavaScript.
 
+  To access a helper's value you have to use `getValue` from
+  `@glimmer/tracking/primitives/cache`.
+
   ```js
   // app/components/data-loader.js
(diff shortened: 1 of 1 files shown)


LevelbossMike:
It's unclear to me why the tests are failing. Can I retrigger a build somehow? Pushing to my fork didn't seem to retrigger the build.

mixonic (approved):
@pzuraq gave me a gut check that this is indeed the correct API. Thank you @LevelbossMike !

PR #19683, opened 2021-07-29
[router.ts] Convert this._super to super

elwayman02:
The router class has already been converted to a native JS class, so it should be using `super.willDestroy` in its `willDestroy` hook instead of `this._super`.

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/router.ts
+++ b/packages/@ember/-internals/routing/lib/system/router.ts
@@ -802,7 +802,7 @@ class EmberRouter extends EmberObject.extend(Evented) implements Evented {
       this._toplevelView = null;
     }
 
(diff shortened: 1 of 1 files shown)


packages/@ember/-internals/routing/lib/system/router.ts
```diff
@@ -802,7 +802,7 @@ class EmberRouter extends EmberObject.extend(Evented) implements Evented {
       this._toplevelView = null;
     }
 
-    this._super(...arguments);
+    super.willDestroy();
```
elwayman02, line 805: Per errors from tslint, we need to drop `...arguments` here because the underlying base method for `willDestroy` is already known to accept 0 arguments.

rwjblue (approved):

PR #19690, opened 2021-08-01
[BUGFIX release] Deprecate htmlSafe via prototype

mixonic:
In Ember 3.24 various string methods added to the `String.prototype` were deprecated for removal in Ember 4.0. `htmlSafe` (the version available via string prototype) was supposed to be included in those deprecations, however dues to its implementation being different it was missed. This omission can be understood as a bug.

This patch deprecates `String.prototype.htmlSafe` targeting Ember 4.0. This will allow the removal of *all* string prototype extensions in 4.0 as intended by the original deprecation.

See also: https://github.com/emberjs/ember.js/pull/19654#issuecomment-890386906

### DIFF
--- a/packages/ember/index.js
+++ b/packages/ember/index.js
@@ -493,6 +493,19 @@ Ember._captureRenderTree = captureRenderTree;
 
 if (ENV.EXTEND_PROTOTYPES.String) {
   String.prototype.htmlSafe = function () {
+    deprecate(
+      `String prototype extensions are deprecated. Please import htmlSafe from '@ember/template' instead.`,
+      false,
+      {
+        id: 'ember-string.prototype-extensions',
+        for: 'ember-source',
+        since: {
+          enabled: '3.27.6',
+        },
+        until: '4.0.0',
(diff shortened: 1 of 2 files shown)


packages/ember/index.js
```diff
@@ -493,6 +493,19 @@ Ember._captureRenderTree = captureRenderTree;
...
 if (ENV.EXTEND_PROTOTYPES.String) {
   String.prototype.htmlSafe = function () {
+    deprecate(
+      `String prototype extensions are deprecated. Please import htmlSafe from '@ember/template' instead.`,
+      false,
+      {
+        id: 'ember-string.prototype-extensions',
+        for: 'ember-source',
+        since: {
+          enabled: '3.27.6',
```
mixonic, line 503: Should be the next 3.27 release.

kategengler (approved):

nlfurniss:
@mixonic should this be added to the deprecation-app? I can do that if so

mixonic:
@nlfurniss yeah I think so, it can join the other string deprecation entry. It doesn't need to be a new entry entirely.

PR #19691, opened 2021-08-01
Add build assertion against `{{outlet named}}`

mixonic:
Named outlets are effectively removed in Ember 4.0, as the render hook and `renderTemplate` method which are used to configure them have been removed. Add an assertion to catch anyone trying to use this API going forward.

See also: https://github.com/emberjs/ember.js/pull/19659#issuecomment-890384173


packages/ember-template-compiler/tests/plugins/assert-against-named-outlets-test.js
```diff
@@ -0,0 +1,26 @@
+import { compile } from '../../index';
+import { moduleFor, AbstractTestCase } from 'internal-test-helpers';
+
+moduleFor(
+  'ember-template-compiler: assert-against-named-outlets',
```
mixonic, line 5: I'm not sure where to look in CI output to assert this was run on CI.

rwjblue: https://github.com/emberjs/ember.js/pull/19691/checks?check_run_id=3214769876#step:8:7625

kategengler (approved):

PR #19692, opened 2021-08-02
Fix assign polyfill

mydea:
In https://github.com/emberjs/ember.js/pull/19649 `assign` was re-written to just pass through to `Object.assign`. However, the implementation is not ideal - it assigns all the arguments, which includes the `target` itself. This can lead to problems, e.g. in ember-test-helpers when you have an implementation like this: `assign(event, options)`, it will end up as `Object.assign(event, event, options)`, and complain that we try to set read-only attributes on `event`.

See: https://github.com/emberjs/ember-test-helpers/issues/1077

### DIFF
--- a/packages/@ember/polyfills/lib/assign.ts
+++ b/packages/@ember/polyfills/lib/assign.ts
@@ -29,7 +29,7 @@ export function assign(target: object, ...sources: object[]): object;
   @public
   @static
 */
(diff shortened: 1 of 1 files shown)


rwjblue (approved):
Thank you!!!

stefanpenner:
:shipit:

PR #19695, opened 2021-08-04
Remove {{partial}}

mixonic:


nlfurniss:
As a note, prob should add an item to the checklist to remove `partial` code from glimmer-vm

kategengler (approved):

PR #19697, opened 2021-08-05
[BUGFIX lts] ensure `deserializeQueryParam` is called for lazy routes

brendenpalmer:
Currently it's possible for `deserializeQueryParam` to not be called on lazy routes, which can break deserialization of query params. As soon as `getFullQueryParams` is called it caches query params state regardless of whether all routes have resolved; this PR updates to only cache state if all routes have resolved, however we still process query params in both of these cases. I've also added a test-case here to prevent this from regressing in the future.

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/route.ts
+++ b/packages/@ember/-internals/routing/lib/system/route.ts
@@ -2082,11 +2082,21 @@ export function getFullQueryParams(router: EmberRouter, state: TransitionState<R
     return state['fullQueryParams'];
   }
 
-  state['fullQueryParams'] = {};
-  Object.assign(state['fullQueryParams'], state.queryParams);
+  let fullQueryParamsState = {};
+  let haveAllRouteInfosResolved = state.routeInfos.every((routeInfo) => routeInfo.route);
 
-  router._deserializeQueryParams(state.routeInfos, state['fullQueryParams'] as QueryParam);
-  return state['fullQueryParams'];
+  Object.assign(fullQueryParamsState, state.queryParams);
+
+  router._deserializeQueryParams(state.routeInfos, fullQueryParamsState as QueryParam);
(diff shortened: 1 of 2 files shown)


packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -2082,11 +2082,21 @@ export function getFullQueryParams(router: EmberRouter, state: TransitionState<R
     return state['fullQueryParams'];
   }
 
-  state['fullQueryParams'] = {};
-  Object.assign(state['fullQueryParams'], state.queryParams);
+  let fullQueryParamsState = {};
+  let haveAllRouteInfosResolved = state.routeInfos.every((routeInfo) => routeInfo.route);
```
rwjblue, line 2086: The master branch no longer supports IE11 (which is why `.every` works in CI here), but if we are going to backport to 3.24 (current LTS) then we'll need to migrate to be IE11 compatible.

rwjblue (changes requested):
Thanks for working on this @brendenpalmer!

packages/@ember/-internals/routing/lib/system/route.ts
brendenpalmer: Looking at [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every#browser_compatibility), I think `Array.prototype.every` was initially part of ES5 and rolled out in IE9. For some reason, I also thought that `.every` was much newer, but it seems like this _should_ work in IE11.

rwjblue: Apparently, I'm bonkers. `.every` **was** in IE9...

rwjblue (approved):

PR #19699, opened 2021-08-05
ensure `deserializeQueryParam` is called for lazy routes

brendenpalmer:
opening a PR to the `lts-3-24` branch, based on original PR here: https://github.com/emberjs/ember.js/pull/19697


kategengler:
As @mixonic mentioned, the commit tag of [BUGFIX lts] in the original PR is enough to get the fix into our next LTS point release. 

Thanks for the fix!

PR #19704, opened 2021-08-18
[CLEANUP beta] Remove template-compiler.registerPlugin

nlfurniss:
Part of #19617
[Deprecation site](https://github.com/emberjs/ember.js/blob/e28a201b9ccfb16f93089cea9207fca9b552d98c/packages/ember-template-compiler/lib/system/compile-options.ts#L93)
[Deprecation PR](https://github.com/emberjs/ember.js/pull/19429)

### DIFF
--- a/packages/ember-template-compiler/lib/system/compile-options.ts
+++ b/packages/ember-template-compiler/lib/system/compile-options.ts
@@ -1,6 +1,5 @@
 import { EMBER_STRICT_MODE } from '@ember/canary-features';
-import { assert, deprecate } from '@ember/debug';
-import { AST, ASTPlugin, ASTPluginEnvironment, Syntax } from '@glimmer/syntax';
+import { assert } from '@ember/debug';
 import { RESOLUTION_MODE_TRANSFORMS, STRICT_MODE_TRANSFORMS } from '../plugins/index';
 import { EmberPrecompileOptions, PluginFunc } from '../types';
 import COMPONENT_NAME_SIMPLE_DASHERIZE_CACHE from './dasherize-component-name';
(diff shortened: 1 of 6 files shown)


packages/ember-template-compiler/tests/basic-usage-test.js
```diff
@@ -65,10 +52,6 @@ moduleFor(
     }
 
     '@test registerPlugin based transforms can be avoided'(assert) {
```
mixonic, line 54: I think this whole test is not needed? If there is a test of the basic functionality of plugin elsewhere here, then this test is probably no longer adding coverage. The name of the test here and comments in the body suggest that it was narrowly concerned with the interactions of the two APIs.

mixonic (dismissed):

packages/ember-template-compiler/tests/basic-usage-test.js
nlfurniss: Good call. Also found another thing to cleanup as well.

packages/ember-template-compiler/lib/system/compile-options.ts
```diff
@@ -80,110 +79,3 @@ interface LegacyPlugin {
 }
 
 export type LegacyPluginClass = new (env: ASTPluginEnvironment) => LegacyPlugin;
```
mixonic, line 81: I think this can be removed now. There is a referencing in a typing file I'm not sure exactly how to unravel (seems to be entangled with a `__raw` property?) but do you think it can also be removed?

mixonic (dismissed):
Very close, awesome work

packages/ember-template-compiler/lib/system/compile-options.ts
nlfurniss: I think it can be, but because I’m a TS noob I wasn’t sure how 😓

mixonic:
Thank you @nlfurniss !

PR #19707, opened 2021-08-20
[CLEANUP beta] Remove Application Controller Router Properties

btecu:


packages/@ember/-internals/glimmer/tests/integration/components/link-to/routing-angle-test.js
```diff
@@ -1708,36 +1708,6 @@ moduleFor(
...
-        this.route('about');
-      });
-
-      await this.visit('/');
-
-      await this.click('#about-link');
-
-      expectDeprecation(() => {
-        let currentRouteName = this.applicationInstance
-          .lookup('controller:application')
```
mixonic, line 1731: I don't think this test was specifically testing deprecated behavior, I think the deprecated API was only used during test of a legitimate behavior we want to retain.

I believe the right approach here is to refactor the test to use the router service instead of `currentRouteName` via the application controller, remove the `expectDeprecation(` code, and retain the test.

Can you dig into it?

packages/@ember/-internals/glimmer/tests/integration/components/link-to/routing-curly-test.js
```diff
@@ -1822,44 +1822,6 @@ moduleFor(
...
-        () => this.visit('/'),
-        /Passing the `@target` argument to <LinkTo> is deprecated\./,
-        EMBER_MODERNIZED_BUILT_IN_COMPONENTS
-      );
-
-      await this.click('#about-link > a');
-
-      expectDeprecation(() => {
-        let currentRouteName = this.applicationInstance
-          .lookup('controller:application')
```
mixonic, line 1853: Again, I this test should be retained with the other `expectDeprecation` left in place, but the code for the specific deprecated API removed here refactored.

packages/@ember/application/tests/reset_test.js
```diff
@@ -129,9 +128,6 @@ moduleFor(
           let location = initialRouter.get('location');
 
           assert.equal(location.getURL(), '/one');
-          expectDeprecation(() => {
-            assert.equal(get(initialApplicationController, 'currentPath'), 'one');
```
mixonic, line 133: Similar to above, can this assertion be changed to assert the router service with `currentRoutePath`? https://api.emberjs.com/ember/3.27/classes/RouterService/properties/currentRouteName?anchor=currentRouteName

This is an integration test, and I wouldn't want to lose the coverage that Ember's routing system in general is up to date after a visit call.

packages/@ember/application/tests/reset_test.js
```diff
@@ -160,9 +156,6 @@ moduleFor(
           );
 
           assert.equal(location.getURL(), '/one');
-          expectDeprecation(() => {
-            assert.equal(get(applicationController, 'currentPath'), 'one');
```
mixonic, line 164: Same here as elsewhere in this file.

packages/ember/tests/routing/decoupled_basic_test.js
```diff
@@ -79,14 +79,6 @@ moduleFor(
       });
     }
 
-    get currentPath() {
-      let currentPath;
-      expectDeprecation(() => {
-        currentPath = this.applicationInstance.lookup('controller:application').get('currentPath');
```
mixonic, line 85: ...and again here, use the router service.

(I'm surprised these were not all refactored when the deprecation was added.)

mixonic (changes requested):
@btecu I think in the testing suite in general (both in places I flagged and even in places I didn't flag) the right path here is to refactor the assertions to use the router service public API instead of the application controller API. Removing the assertions (or tests) entirely is too much of a change, and degrades out test coverage of how Ember reflect routing state.

Can you give it another look through that lens?

btecu:
@mixonic updated.

mixonic (approved):
@btecu this is amazing!!

PR #19708, opened 2021-08-21
[CLEANUP beta] Remove class-binding-and-class-name-bindings-in-templates

nlfurniss:
Part of #19617

[Deprecation guide](https://deprecations.emberjs.com/v3.x/#toc_class-binding-and-class-name-bindings-in-templates)

Passing the `class` argument is still supported per the deprecation guide, so only removed `classNameBinding` and `classBinding`


packages/ember-template-compiler/lib/plugins/transform-old-class-binding-syntax.ts
```diff
@@ -1,61 +1,33 @@
...
 import { AST, ASTPlugin } from '@glimmer/syntax';
-import calculateLocationDisplay from '../system/calculate-location-display';
 import { Builders, EmberASTPluginEnvironment } from '../types';
 
 export default function transformOldClassBindingSyntax(env: EmberASTPluginEnvironment): ASTPlugin {
   let b = env.syntax.builders;
-  let moduleName = env.meta?.moduleName;
 
   return {
     name: 'transform-old-class-binding-syntax',
```
mixonic, line 8: I feel like more of this file can perhaps be ripped out. A lot of the logic seems to be around merging the three items of `class`, `classBinding` and `classNameBindings` however as only one of those things remains the logic of merging them shouldn't be needed?

I think in theory this whole file should be removed? Or if it is needed for `class`, what is the crisp explanation of what it is doing special for that argument?

@rwjblue you have a comment `TODO: add helpful deprecation...` further down in this file which I'm not sure how to apply now that we are doing the removal.

mixonic (changes requested):

packages/ember-template-compiler/lib/plugins/transform-old-class-binding-syntax.ts
nlfurniss: Removed the whole transform and tests pass locally

mixonic (approved):

Issue #19494, opened 2021-04-11
[Bug] 3.26: Router is not setup in a non-application test


grayt0r:
### 🐞 Describe the Bug
Ember 3.24 included a change that meant the router service automatically worked in non-application tests (i.e. no need for `this.owner.setupRouter()`) - this worked great in 3.25, but seems to have regressed in 3.26.

### 🔬 Minimal Reproduction
Run the tests in this repo: https://github.com/grayt0r/ember-router-setup-in-tests.

Relevant code changes are [here](https://github.com/grayt0r/ember-router-setup-in-tests/compare/a98cb485063a12c17f4ed8ff42aed50fe8dc3f0e...52ed881a1324696a11991d7b0c7455d725652d02).

### 😕 Actual Behavior
The router is not setup and the test fails with an error (the exact error depends on which router method the component calls):

```
Cannot read property 'generate' of undefined
```

### 🤔 Expected Behavior
The router is setup, the link is correctly rendered and the test passes.

### 🌍 Environment

- Ember: - v3.26.1

### ➕ Additional Context
The test passes if I downgrade to 3.25.4 or if I add `this.owner.setupRouter()` to the test setup.

I think it might be related to this change: https://github.com/emberjs/ember.js/pull/19405.

/cc @Ravenstine @rwjblue

rwjblue:
Chatted with @Ravenstine about this today, they are going to work on a failing test reproduction first so we can figure out the right way to fix...

chriskrycho:
Hit this as a blocker for upgrading to 3.28. @Ravenstine did you ID the root cause and/or a path forward? Happy to collaborate on landing that!

grayt0r:
@chriskrycho sorry for the delay, but that fix appears to solve our issues, thank you very much! 🙌

chriskrycho:
Excellent!

chriskrycho:
Noticed this was still open despite the issue being resolved, just closed it!

PR #19733, opened 2021-08-26
[BUGFIX release] fix router test regression in urlFor and recognize

chriskrycho:
As part of the improvements made between 3.24 and 3.28, the router microlib is now lazily loaded. When these changes were made, there were a couple cases where it *should* be possible to access router state in a non-application test (integration etc.) but it currently is not because the router is not necessarily set up. Since `setupRouter` is idempotent, call it in those functions so that if it is *not* set up, it gets set up, and otherwise it will continue working as expected.

Besides the change to the tests here, I have verified that this restores the previous behavior in our app.

Related to #19494. (Not marking as "fixes" because I'm not sure this covers 100% of that!)

### DIFF
--- a/packages/@ember/-internals/routing/lib/services/router.ts
+++ b/packages/@ember/-internals/routing/lib/services/router.ts
@@ -241,6 +241,7 @@ export default class RouterService extends Service {
      @public
    */
   urlFor(routeName: string, ...args: any[]) {
+    this._router.setupRouter();
     return this._router.generate(routeName, ...args);
   }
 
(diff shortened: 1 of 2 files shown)


rwjblue (approved):

NullVoxPopuli:
should this apply to `isActive` as well?

sly7-7:
@chriskrycho I remember working a bit on #19405 (In fact, I were clumpsy, because I made a similar PR on my own side). The setupRouter() thing, before, was here, https://github.com/emberjs/ember.js/pull/19405/files#diff-bdcd334bfbf30baf672240b941910d9462e59b46fb9091402ee107f214c40045L76, and I think it would be a good move to put that setup in this again. So that accessing the router the first time, ensure it is set up. In my Pr https://github.com/emberjs/ember.js/pull/19454/ I did'nt remove the setupRouter thing, and all tests were green

chriskrycho:
@sly7-7 I only  just saw this, sorry—I don't have any particular attachment to the way I implemented this, so if you've got a better one, landing it seems great!

PR #19709, opened 2021-08-22
[CLEANUP beta] Remove deprecated `Router Events`

btecu:

PR #19749, opened 2021-09-09
[CLEANUP beta] Remove `deprecate-router-events` support code

xg-wang:
Deprecation https://deprecations.emberjs.com/v3.x/#toc_deprecate-router-events

This PR removes
- Overriding`willTransition` and `didTransition` hooks on EmberRouter
- `willTransition` and `didTransition` events on `Route`

Based on https://github.com/emberjs/rfcs/blob/af64915b5ecde010fce09309a47ee6d2447588d0/text/0095-router-service.md#transition-object I think `willTransition` and `didTransition` on `Route` are not deprecated. 
They are triggered from https://github.com/tildeio/router.js/blob/4f5d411e9ba44efea003bc3159484b4060ebfc2a/lib/router/router.ts#L811


snewcomer:
note ref - https://github.com/emberjs/ember.js/pull/19709

Either one is fine by me!

xg-wang:
It looks like https://github.com/emberjs/ember.js/pull/19709 removed the deprecation warnings but still triggers those events if app has the route event. But it also appears I missed removing used test code part, I picked those to this PR.

xg-wang:
@rwjblue I checked the router service RFC and I think actions on `Route` are not deprecated

packages/@ember/-internals/routing/lib/system/router.ts
```diff
@@ -380,50 +371,22 @@ class EmberRouter extends EmberObject.extend(Evented) implements Evented {
...
-                },
-              }
-            );
-          }
-        }
+        assert(
+          'You attempted to override the "didTransition" method which has been deprecated. Please inject the router service and listen to the "routeDidChange" event.',
+          router.didTransition === defaultDidTransition
+        );
         router.didTransition(infos);
```
snewcomer, line 380: I'm wondering if we can take this a step further and make defaultDidTransition && defaultWillTransition the defacto method that is called when didTransition/willTransition is called on the PrivateRouter. I'm wondering if we can take everything in, say defaultDidTransition and just throw it in here and then get rid of the `assert`.

rwjblue (approved):

PR #19755, opened 2021-09-19
remove willTransition and didTransition docs and deprecation code

knownasilya:


packages/@ember/-internals/routing/lib/system/router.ts
```diff
@@ -381,23 +379,6 @@ class EmberRouter extends EmberObject.extend(Evented) implements Evented {
...
-                url: 'https://deprecations.emberjs.com/v3.x#toc_deprecate-router-events',
-                for: 'ember-source',
-                since: {
-                  enabled: '3.11.0',
-                },
-              }
-            );
-          }
-        }
         router.didTransition(infos);
```
knownasilya, line 382: Is this the correct way to deal with this? Or do these methods go away entirely?

packages/@ember/-internals/routing/lib/system/router.ts
```diff
@@ -406,23 +387,6 @@ class EmberRouter extends EmberObject.extend(Evented) implements Evented {
...
-                url: 'https://deprecations.emberjs.com/v3.x#toc_deprecate-router-events',
-                for: 'ember-source',
-                since: {
-                  enabled: '3.11.0',
-                },
-              }
-            );
-          }
-        }
         router.willTransition(oldInfos, newInfos, transition);
```
knownasilya, line 390: Is this the correct way to deal with this? Or do these methods go away entirely?

packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -561,76 +561,6 @@ class Route extends EmberObject.extend(ActionHandler, Evented) implements IRoute
...
-        return true; // Bubble the didTransition event
-      }
-    }
-    ```
-
-    @event didTransition
-    @since 1.2.0
-    @public
-  */
-
```
knownasilya, line 633: The deprecation guide is not clear, and the code at the bottom of this file, if the hooks on the Route are deprecated as well.

packages/@ember/-internals/routing/lib/system/router.ts
snewcomer: @knownasilya My initial feeling is we need to make `defaultDidTransition` && `defaultWillTransition` the defacto method that is called when didTransition/willTransition is called on the PrivateRouter.  I'm wondering if we can take everything in, say `defaultDidTransition` and just throw it in here.  What do you think?

chriskrycho:
Is this still relevant, or was it superseded by someone else's work during the push for Ember 4.0?

knownasilya:
I need to fix conflicts and mark as ready.

PR #19760, opened 2021-09-22
[chore]: require ember-auto-import 2.0.0

### LINKED ISSUE #19752: Require ember-auto-import `>= 2` or higher with ember-source@4.0.0
In order to allow ember-source to migrate to be a V2 addon, we **must** require that `ember-auto-import@2` is present in the consuming application. 

Concretely, we should add code to the addon entry point to check `this.project.addons` includes `ember-auto-import@2` installed. That code would go somewhere around here:

https://github.com/emberjs/ember.js/blob/e53d6bb4bbb5230829eb589fdba2388de3cd28cf/lib/index.js#L45

snewcomer:
close #19752


ef4:
Thanks for working on this. But there's an easier way to do this -- ember-auto-import already implements this check.

I submitted #19761 to solve it that way.

PR #19762, opened 2021-09-27
[CLEANUP beta] Update GlimmerVM to 0.81

rwjblue:
High level breaking changes:

* removes deprecation of mutations during helper compute
* removes deprecation of mutations during unknownProperty
* `@glimmer/integration-tests`, `@glimmer/manager`, `@glimmer/validator`
  * [#1330](https://github.com/glimmerjs/glimmer-vm/pull/1330) Remove deprecated support for mutation after consumption during certain manager hooks ([@snewcomer](https://github.com/snewcomer))
* `@glimmer/manager`
  * [#1328](https://github.com/glimmerjs/glimmer-vm/pull/1328) Remove deprecated Component Manager version 3.4 ([@nlfurniss](https://github.com/nlfurniss))
* `@glimmer/integration-tests`, `@glimmer/manager`
  * [#1329](https://github.com/glimmerjs/glimmer-vm/pull/1329) Remove deprecated Modifier Manager version 3.13 ([@nlfurniss](https://github.com/nlfurniss))

See full details here:

https://github.com/glimmerjs/glimmer-vm/releases/tag/v0.81.0

Closes #19735
Closes #19731

PR #19774, opened 2021-10-06
bump to 0.82 glimmer-vm

snewcomer:
https://github.com/glimmerjs/glimmer-vm/blob/master/CHANGELOG.md#boom-breaking-change


nlfurniss:
doesnt look like all the changes from #19762 are present, and I'm guessing they'd need to be 🤷

rwjblue:
@nlfurniss - This PR targets that one (thats why some of those changes don't show up in the diff)

rwjblue:
Merged into #19762 now, we can continue to merge PRs into that branch.

Issue #19609, opened 2021-06-25
[Tests] Improve test suite coverage for Routing


jenweber:
The goal of this issue is to write some new tests or modify existing tests to improve coverage. There have been a series of routing bugs that may be related, and improved test coverage may have been able to prevent them. Multiple people can help work on this issue - just comment with what you are working on, and be sure to read through other people's comments.

To do:
- read through some of the issues below
- Write a test to help cover one of the problem areas

Past bug reports/issues:
- https://github.com/emberjs/ember.js/issues/19497
- https://github.com/emberjs/ember.js/issues/18577
- https://github.com/emberjs/ember.js/issues/17494
- https://github.com/emberjs/ember.js/issues/17118
- https://github.com/emberjs/ember.js/issues/16349
- https://github.com/emberjs/ember.js/issues/14875
- https://github.com/emberjs/ember.js/issues/14721
- https://github.com/emberjs/ember.js/issues/12169
- https://github.com/emberjs/ember.js/issues/11563

laulujan:
Hello, I would like to give it a try on one of the tests needed, I'll check on the reports and comment later on the one I'll be working on!!!

jenweber:
Thanks @laulujan! Let us know if you have any questions. I haven’t worked in this area of the codebase myself, but I can find people who have!

iigorcunha:
Hello @jenweber, I saw the test but how can I find the coverage? is there any place to see a coverage report? I would like to work on it.

Aditya-eddy:
@jenweber  can you please tell what is meant by write a test. I am a newbie

jenweber:
Hi @Aditya-eddy, welcome and thanks for your willingness to help out! I think the best way to learn about writing tests is to do the official Ember tutorial: https://guides.emberjs.com/release/tutorial/part-1/ Throughout the sections, you will write many different kinds of tests, and then Github Issues like this one will make more sense.

jenweber:
Also for anyone else looking into helping with this ticket - unfortunately I don't have much additional context or guidance to share to share. Part of the task at hand is to figure out what the tests are like for this area of the codebase and what can be improved. I haven't worked on routing at all myself. I just know that this is a problem area based on user reports and some comments from other core team members. Good luck!

arjot-rai:
[...]

PR #19775, opened 2021-10-07
[DOC] Mark `replaceWith` and `replaceRoute` as deprecated

bertdeblock:
Related to https://github.com/emberjs/rfcs/pull/674.


btecu:
Before merging this change, it would be good to see some of the issues linked here addressed https://github.com/emberjs/ember.js/issues/19609.

Currently, on a route `this.transitionTo` and `this.router.transitionTo` might not work the same, so that should be addressed before this deprecation.

PR #13518, opened 2016-05-17
Allow engine instances to spawn other engine instances.

dgeb:
Continues the move upstream of functionality from [ember-engines](https://github.com/dgeb/ember-engines) into ember core:
- Introduces a new `ENGINE_PARENT` symbol, as well as `getEngineParent` and `setEngineParent` methods, for tracking the parents of engine instances.
- Introduces a new `buildChildEngineInstance` method on `EngineInstance`.

PR #19786, opened 2021-10-12
Promote `*engineParent` methods for public API

villander:
this API was introduced before by https://github.com/emberjs/ember.js/pull/13518 to enable developers and ember internal tracking the parents of engine instances. Sometimes we want to use **getEngineParent** as an immediate parent (which usually will be the host app, but maybe another engine) to get the desirable parent to do something since getting the **rootApplication** or access the parent context across **engines**.

A great example, using `glimmer-apollo` I can for example share the Apollo Client across engines (nested or not )

On `ember-source: 3.28.1` I already can access these methods using `import { getEngineParent, setEngineParent } from '@ember/engine';` now we just need to make it public in the docs

cc: @rwjblue @dgeb

### DIFF
--- a/packages/@ember/engine/lib/engine-parent.js
+++ b/packages/@ember/engine/lib/engine-parent.js
@@ -13,7 +13,7 @@ const ENGINE_PARENT = symbol('ENGINE_PARENT');
   @return {EngineInstance} The parent engine instance.
   @for @ember/engine
   @static
-  @private
+  @public
 */
 export function getEngineParent(engine) {
   return engine[ENGINE_PARENT];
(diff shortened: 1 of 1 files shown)


rwjblue:
I don't fully understand the motivating use cases here. Can you explain in more detail why this API is required to support this?

villander:
@rwjblue @runspired can we merge it?

kategengler:
Is this still relevant?

villander:
@kategengler yes! It’s still relevant IMO.

cc @runspired @SergeAstapov

kategengler:
Moving to a draft until someone here wants to own it cc @void-mAlex

void-mAlex:
I do not believe this is desirable, it breaks isolations concerns of the engine
if you want to share the same apollo client share a service that hold access to the same instance
each engine at every level decides what they share to their children, the child doesn't get a choice about the parents

kategengler:
Closing as @void-mAlex is our current Engines owner

PR #19793, opened 2021-10-18
Determine next alpha version based on latest alpha dist-tag and value

kategengler:
for version in package.json


rwjblue (approved):
One minor question, but otherwise LGTM

PR #19795, opened 2021-10-20
Switch from QUnit `assert.equal()` to `assert.strictEqual()`

bmish:
[no-assert-equal](https://github.com/platinumazure/eslint-plugin-qunit/blob/master/docs/rules/no-assert-equal.md) is now a recommended rule in [eslint-plugin-qunit v7](https://github.com/platinumazure/eslint-plugin-qunit/blob/master/CHANGELOG.md#700). In JavaScript, we should generally use strict equality checks like [assert.strictEqual](https://api.qunitjs.com/assert/strictEqual/).

This PR updates all references in the codebase including internal tests and blueprints used for code generation.

Related to how we are upgrading ember-cli to use eslint-plugin-qunit v7:
* https://github.com/ember-cli/ember-cli/pull/9671
* https://github.com/ember-cli/ember-cli/pull/9667
* https://github.com/emberjs/data/pull/7726

### DIFF
--- a/node-tests/fixtures/component-test/unit.js
+++ b/node-tests/fixtures/component-test/unit.js
@@ -7,7 +7,7 @@ moduleForComponent('x-foo', 'Unit | Component | x-foo', {
 });
 
 test('it renders', function(assert) {
-  
+
   // Creates the component instance
   /*let component =*/ this.subject();
   // Renders the component to the page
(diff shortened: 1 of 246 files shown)


bmish:
Looks like there are 104 test failures that will mostly require changing between `null` and `undefined`. This will be tedious, manual work to do so I'm going to close this for now. Updating the blueprints was already done in #19825.

PR #19804, opened 2021-10-25
Make getProperties() allocate less & more readable

izelnakri:


jelhan:
API docs explicitly mention invocation with multiple arguments as well as with an array as second argument as supported: https://api.emberjs.com/ember/release/functions/@ember%2Fobject/getProperties

Dropping support for one of them would be a breaking change in my opinion. I guess it would require a RFC and must be deprecated before removing.

I agree that it is not optimal to support two different invocation syntax. I would support a RFC settling for one. It should harmonize all APIs and not only touch one in my opinion.

I'm not a core team member. Not sure if this has been discussed before.

izelnakri:
@jelhan I agree with you.

rwjblue:
>  I would support a RFC settling for one.

FWIW, it isn't obvious that we should actually have _any_ in the long run. Specifically, I'd much prefer folks destructure instead...

chriskrycho:
Thanks for the idea here! Given the way it works, and @rwjblue's comment, I'm going to do as you suggested here, @izelnakri, and close it. Appreciate the thought and attempt, though!

PR #19808, opened 2021-10-30
[CLEANUP beta] remove the `--test-type` option from the helper blueprint

### LINKED ISSUE #18245: `helper-test --test-type unit` generates broken code in Ember 3.11
It looks like https://github.com/emberjs/ember.js/pull/15609 was merged without also adjusting the [`helper-test` blueprint](https://github.com/emberjs/ember.js/blob/fb60c41d1aab65be3a249b6c9ea9627c80d8e9c7/blueprints/helper-test/qunit-files/__root__/__testType__/__collection__/__name__-test.js#L16). Since the helper module no longer exports a named function the generated code is invalid by default.

I'm not sure how this should be solved. Unit tests can still be useful for helpers that return values which are not intended for rendering.

Windvis:
This removes the `--test-type` option from the helper blueprint. The unit test blueprint [has been broken for a while now](https://github.com/emberjs/ember.js/issues/18245), and it was suggested that this option should be removed since rendering tests are the recommended testing type for helpers.

Closes #18245


node-tests/fixtures/helper-test/mocha.js
```diff
@@ -1,5 +1,4 @@
 import { expect } from 'chai';
-
```
Windvis, line 2: I've created a second commit which removes the extra newlines from the fixtures. I can squash this into the first one if we want that change.

rwjblue (approved):
This is great, thank you!!

PR #19809, opened 2021-10-30
[bug] Generate url with nested query params

snewcomer:
ref https://github.com/emberjs/ember.js/issues/19791

Blocked by https://github.com/tildeio/route-recognizer/pull/175

### DIFF
--- a/packages/ember/tests/routing/router_service_test/urlFor_test.js
+++ b/packages/ember/tests/routing/router_service_test/urlFor_test.js
@@ -113,6 +113,26 @@ moduleFor(
       });
     }
 
+    ['@test RouterService#urlFor returns URL for simple route with nested object as query params'](assert) {
+      assert.expect(1);
+
+      let queryParams = this.buildQueryParams({
+        filter: {
+          user: {
+            name: {
+              $contains: 'foo',
+            },
+          },
+        },
+      });
+
+      return this.visit('/').then(() => {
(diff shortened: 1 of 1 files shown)


kategengler:

packages/ember/tests/routing/router_service_test/urlFor_test.js
```diff
@@ -113,6 +113,26 @@ moduleFor(
...
+              $contains: 'foo',
+            },
+          },
+        },
+      });
+
+      return this.visit('/').then(() => {
+        let expectedURL = this.routerService.urlFor('parent.child', queryParams);
+
+        assert.equal('/child?filter=[user][name][$contains]=foo', expectedURL);
```
sukima, line 132: question: This looks off. are we sure this is encoded correctly?

Per [RFC3986](https://www.rfc-editor.org/rfc/rfc3986#section-2.2):

> If data for a URI component would conflict with a reserved character's purpose as a delimiter, then the conflicting data must be percent-encoded before the URI is formed.
 
And states that `=` is indeed a **reserved character** as a sub-delim. I'm a little confused how to parse this. Is `filter` the key to the key/value pair or is `[user][name][$contains]` or `filter=[user][name][$contains]` the key?

suggestion: per [this SO answer](https://stackoverflow.com/a/78046581/227176) a convention might be: `?filter[user][name[]$contains]=foo` instead.

PR #19826, opened 2021-11-05
[FEATURE] Enable `@cached` decorator from `@glimmer/tracking`

rwjblue:
This was discussed and approved to be enabled by default (for 4.1) in a core team meeting a while back, but I forgot to actually submit the change. 🤦


NullVoxPopuli (approved):
Woo!

PR #19827, opened 2021-11-05
[DOC] Fix tracked decorator usage in code examples

### LINKED ISSUE #19803: tracked decorator documentation out of date
https://github.com/emberjs/ember.js/blob/67d465f7603685b20a4dc27833164ceab7003559/packages/%40ember/-internals/metal/lib/tracked.ts#L46-L73

The in-code documentation seems to be out of date, specifically the @example Dependents (unsure, maybe my reading is wrong).

The 'tracked' decorator doesn't seem to handle dependencies in this way, and injecting this example into a project results in an error: 

> Error: Assertion Failed: tracked() may only receive an options object containing 'value' or 'initializer', received eatenApples

SergeAstapov:
Fixes #19803

### DIFF
--- a/packages/@ember/-internals/metal/lib/tracked.ts
+++ b/packages/@ember/-internals/metal/lib/tracked.ts
@@ -63,7 +63,6 @@ import { SELF_TAG } from './tags';
     @tracked
     eatenApples = 0
 
-    @tracked('eatenApples')
     get remainingApples() {
       return totalApples - this.eatenApples;
     }


jelhan:
At least this param documentation seems to be wrong as well:

```
@param dependencies Optional dependents to be tracked.
```

But I fear the text may also cause more confusion than it resolves.

>   In the case that you have a computed property that depends other
>  properties, you want to track both so that when one of the
>  dependents change, a rerender is scheduled.
>  In the following example we have two properties,
>  `eatenApples`, and `remainingApples`.

I think most confusing is the statement "you want to track both", the root state and the getter. Instead it should state that getters are updated automatically - and that manual decoration of them is not needed - in my opinion.

SergeAstapov:
@jelhan got point! this may be confusing to someone new-ish to Ember. Seem like remnants of computed properties description. I'll give it a short to rewrite

rwjblue:
@SergeAstapov I converted to draft while you work on that, just mark it ready when you are ready for a review! 😸

SergeAstapov:
figured out:
the root of confusion is that code examples in [`packages/@ember/-internals/metal/lib/tracked.ts`](https://github.com/emberjs/ember.js/blob/master/packages/%40ember/-internals/metal/lib/tracked.ts) outdated (as reported in #19803)

`packages/@ember/-internals/metal/lib/tracked.ts` is actually *`@private`* and does not show up in the API and docs. Only way to discover it - via searching through source code.

`packages/@glimmer/tracking/index.ts` is the module marked as `@public` and what actually shows up in https://api.emberjs.com/ember/3.28/functions/@glimmer%2Ftracking/tracked

I'm not sure there is anything we can fix here beyond just copy-paste from `packages/@glimmer/tracking/index.ts`.

@rwjblue is it just a temp thing `packages/@ember/-internals/metal/lib/tracked.ts` or there are some legit backward compatibility use cases why we do not use real import (or just re-export) from Glimmer? (from what I can tell, there seem to be either computed or classic classes interop in Ember version)

snewcomer (approved):

rwjblue:
> @rwjblue is it just a temp thing packages/@ember/-internals/metal/lib/tracked.ts or there are some legit backward compatibility use cases why we do not use real import (or just re-export) from Glimmer?

Not sure what you mean here by "Glimmer", we do re-export but maybe you mean from @glimmer/tracking?

rwjblue:
https://github.com/emberjs/ember.js/blob/b3a6241ca20a014300cbc69e44aff212fe89bc43/packages/%40glimmer/tracking/index.ts#L1

https://github.com/emberjs/ember.js/blob/b3a6241ca20a014300cbc69e44aff212fe89bc43/packages/%40ember/-internals/metal/index.ts#L56

But maybe you mean the extra bits in here? 

https://github.com/emberjs/ember.js/blob/b3a6241ca20a014300cbc69e44aff212fe89bc43/packages/%40ember/-internals/metal/lib/tracked.ts#L85

PR #19806, opened 2021-10-27
[CLEANUP beta] Drop export of built-ins, remove legacy components

mixonic:
Part of https://github.com/emberjs/ember.js/issues/19617

* Remove the export of `@ember/component/checkbox`
* Remove the export of `@ember/component/text-field`
* Remove the export of `@ember/component/text-area`
* Remove the export of `@ember/component/link-component`
* Remove the assignment of `Ember.Checkbox`
* Remove the assignment of `Ember.TextField`
* Remove the assignment of `Ember.TextArea`
* Remove the assignment of `Ember.LinkComponent`
* Remove the assignment of `Ember.TextSupport`
* Remove the assignment of `Ember.TargetActionSupport`
* ~Remove the assignment of `Ember._Input` (what is this?)~ I brought `_Input` back as it is under test. I'm not familiar with its usage.
* Hard-code `EMBER_MODERNIZED_BUILT_IN_COMPONENTS` to `true` path in most places. The remaining are a few spots where there are deprecations targeting *5.0*.
* Hard-code the built-in components (`<Textarea>`, `<Input>`, `<LinkTo>`) for the modernized path
* Remove the `query-params` helper (it is only needed by the legacy `LinkComponent` and was deprecated).
* Review https://github.com/emberjs/ember.js/pull/19669 and ensure all `disabledWhen` is removed.
* Review https://github.com/emberjs/ember.js/pull/19662 and ensure all `@tagName` support is removed
* Review all `ember.built-in-components.*` deprecations and ensure their functionality is removed.
* Remove code supporting `ember-glimmer.link-to.positional-arguments`, basically https://github.com/emberjs/ember.js/blob/master/packages/ember-template-compiler/lib/plugins/transform-link-to.ts

PR #19664, opened 2021-07-19
Remove Reopening Built In Components

nlfurniss:
Part of #19617

[Deprecation Guide](https://deprecations.emberjs.com/v3.x#toc_ember-built-in-components-reopen)

PR #19831, opened 2021-11-09
[CLEANUP beta] Remove Ember.Component.reopen

SergeAstapov:
Part of #19617


SergeAstapov:
There is related #19664 for the same removal but in built-in components

SergeAstapov:
@mixonic looks like it's done as part of #19806 - would you like to have a small targeted PR or it does not make sense?

mixonic:
@SergeAstapov we would like it to be one big PR, I think. This is because the removed components and much of their capability (like `reopen`) should all be contained within https://github.com/emberjs/ember-legacy-built-in-components/pull/1. `Component.reopen` itself should be removed and I think will *not* be in https://github.com/emberjs/ember-legacy-built-in-components/pull/1, but regardless I think the codepaths are likely tangled enough we should do it in a single go.

I'll review this PR to make sure I'm pulling these same things out in https://github.com/emberjs/ember.js/pull/19806. 

Thank you for this work though, it will be a helpful reference and good gut check.

SergeAstapov:
@mixonic thank you for explanation! Feel free to close this PR once it's not needed (I wasn't sure was doing all things right way anyways, at least CI was happy)

PR #19746, opened 2021-09-07
[CLEANUP beta] remove addArrayObserver

snewcomer:
Part of https://github.com/emberjs/ember.js/issues/19617

Deprecation: https://deprecations.emberjs.com/v3.x/#toc_array-observers
RFC: https://github.com/emberjs/rfcs/pull/692

WIP

PR #19833, opened 2021-11-12
[CLEANUP beta] Remove deprecated array observers

kategengler:

Supplants https://github.com/emberjs/ember.js/pull/19746

* Remove `someArray.addArrayObserver` https://github.com/emberjs/ember.js/pull/19833/files#diff-8116921433c6e1664250d5735e4f9daeb6ea9e00940783e0c4db7adf88aa8772L532
* Remove `someArray.removeArrayObserver` https://github.com/emberjs/ember.js/pull/19833/files#diff-8116921433c6e1664250d5735e4f9daeb6ea9e00940783e0c4db7adf88aa8772L548
* Remove `someArray.hasArrayObservers` https://github.com/emberjs/ember.js/pull/19833/files#diff-8116921433c6e1664250d5735e4f9daeb6ea9e00940783e0c4db7adf88aa8772L571
* Remove `someArray.arrayContentWillChange` https://github.com/emberjs/ember.js/pull/19833/files#diff-8116921433c6e1664250d5735e4f9daeb6ea9e00940783e0c4db7adf88aa8772L642
* Remove `someArray.arrayContentDidChange` https://github.com/emberjs/ember.js/pull/19833/files#diff-8116921433c6e1664250d5735e4f9daeb6ea9e00940783e0c4db7adf88aa8772L670 (also at https://github.com/emberjs/ember.js/pull/19833/files#diff-79a0015ca0e6a6c3661d9119f4caf6b26b111af6bb0e5bb784558631c599a81cL357)
* On the internal version of `addArrayObserver`, remove the optionality of `willChange` and `didChange` arguments to name the hooks. In framework code (and now in test) these are always provided.
* Retain basically all the test coverage of array observers since they remain used by `ArrayProxy` internally (and through that interface, in Ember Data).

Co-authored-by: Matthew Beale <matt.beale@madhatted.com>


packages/@ember/-internals/metal/lib/array.ts
```diff
@@ -58,82 +56,49 @@ export function replaceInNativeArray<T>(
...
   operation(obj, '@array:change', target, didChange);
 
-  if (hasObservers === notify) {
-    notifyPropertyChange(obj, 'hasArrayObservers');
-  }
+  /*
+   * Array proxies have a `_revalidate` method which must be called to set
+   * up their internal array observation systems.
+   */
+  obj._revalidate?.();
```
mixonic, line 85: In this patch `hasArrayObserver` has been removed. That property was not directly deprecated in 3.x, however we've removed a number of APIs in this patch which were related to `someArray.addArrayObserver` (APIs which were only meaningful once you had an observer installed).

Because it is removed, it no longer needs to be notified of change. However the `_revalidate` hook on `ArrayProxy` instances was being called via `PROPERTY_DID_CHANGE` symbol notification. Without trying to rewrite that system we have two options: Either call `_revalidate()` directly, or keep the property `hasArrayObserver` around just to trigger the `PROPERTY_DID_CHANGE`. Here I've chosen to call `_revalidate()` directly.

pzuraq (approved):

PR #19834, opened 2021-11-12
[FEATURE] Enable `RouterService#refresh` method

Windvis:
This probably needs to be discussed in a core meeting first but it would be nice if this feature made it into the 4.1 release :innocent: .


locks:
@Windvis we just discussed and decided to move ahead :) Thanks for the help 💪

PR #19838, opened 2021-11-12
[CLEANUP beta] Remove check to see if `ember` is defined as a Bower dependency

bertdeblock:
Part of #19617.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -44,13 +44,6 @@ module.exports = {
   init() {
     this._super.init && this._super.init.apply(this, arguments);
 
-    if ('ember' in this.project.bowerDependencies()) {
-      // TODO: move this to a throw soon.
-      this.ui.writeWarnLine(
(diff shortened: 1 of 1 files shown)


lib/index.js
```diff
@@ -45,9 +45,8 @@ module.exports = {
     this._super.init && this._super.init.apply(this, arguments);
 
     if ('ember' in this.project.bowerDependencies()) {
-      // TODO: move this to a throw soon.
-      this.ui.writeWarnLine(
-        'Ember.js is now provided by node_module `ember-source`, please remove it from bower'
+      throw new SilentError(
+        'Ember.js is now provided by the `ember-source` node module, please remove `ember` from your `bower.json` file.'
       );
     }
```
mixonic, line 51: Can we remove the whole `if ('ember' in this.project.bowerDependencies()) {` block? cc @rwjblue ?

bertdeblock: Sure, let me know if I should remove.

Issue #19545, opened 2021-05-13
Ember 4.0 Tracking Issue


MelSumner:
# Summary
The purpose of this issue is to provide a way to track the status of the work necessary to ship v.4.0 of Ember.js 

The list of issues should be updated to include links to individual issues as they are created. The individual issues should contain more details for each item on this list. 

As a reminder: Major versions of Ember do not ship new features- instead, major versions remove previously deprecated features. To read more about Ember's philosophy regarding _editions_, please see https://emberjs.com/editions/. 

# Issues

- [x] **blocking** Remove deprecated features (See #19617 for open tasks) (includes IE11 removals)
- [x] [Tracking for Remove jQuery by Default](https://github.com/emberjs/rfc-tracking/issues/3) - Done from the Ember 4.0 perspective.
- [x] :lock: `Ember.assign` deprecation implementation https://github.com/emberjs/ember.js/pull/19649
- [x] **blocking** Deprecation guide for `Ember.assign`
- [x] :lock: Ember `auto` location deprecation implementation https://github.com/emberjs/ember.js/pull/19510 - we got this in targeting 4.1
- [x] _non-blocking for 4.0_ :lock: Deprecation guide for `auto` https://github.com/ember-learn/deprecation-app/pull/882
- [ ] _non-blocking:_ [Remove classic examples from API docs](https://github.com/emberjs/ember.js/issues/19702)
- [ ] **blocking** [Make sure we aren't showing deprecated methods in the guides](https://github.com/ember-learn/guides-source/issues/1711) - mostly done except willTransition/didTransition
- [x] **blocking** [Update the optional features guide](https://github.com/ember-learn/guides-source/issues/1712) - has a PR
- [x] **blocking** [Update the Octane upgrade guide](https://github.com/ember-learn/guides-source/issues/1713) - has a PR
- [x] Land https://github.com/emberjs/ember.js/pull/19825 otherwise newly generated files will error due to the newer version of `eslint-plugin-qunit` in the 4.0 new project blueprint 
- [x] [New browser-support policy](https://github.com/emberjs/rfcs/blob/master/text/0685-new-browser-support-policy.md): One to drop IE11 from Ember Data CI.
- [x] **blocking** [New browser-support policy](https://github.com/emberjs/rfcs/blob/master/text/0685-new-browser-support-policy.md): Need a canonical place to publish supported browsers, and we need to have the list for 4.0 in hand. And write it into the blog post.
[...]

PR #19840, opened 2021-11-13
Remove assertion if model has no `find` method in implicit record loading

snewcomer:
https://github.com/emberjs/ember.js/issues/19545
https://github.com/emberjs/data/pull/7744#issuecomment-966832106

There is no `find` method on model class instances.  Ideally this change to early return was made when ember-data removed the `find` method from model instances - see [here](https://github.com/emberjs/data/blob/v0.13/packages/ember-data/lib/system/model/model.js#L510) for old definition.


snewcomer:
I guess other pkgs could have a a `find` method as well.  Closing for now.

PR #19843, opened 2021-11-14
[CLEANUP beta] Inelegant throw for arg-less-paren-less helper

mixonic:

Convert the deprecation for `argument-less-helper-paren-less-invocation` into a `throw`. A more elegant version of this fix would convert the deprecation upstream in Glimmer VM into an assert, or would update the syntax to match a more desired functionality. I've taken a quick stab at that, but wasn't able to come up with something I was happy with quickly.

I suggest we land this for 4.0, and we can continue to improve the message in the `throw`, or remove the `throw`, or add an `assert` instead at the Glimmer VM level, in followup patches.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/environment.ts
+++ b/packages/@ember/-internals/glimmer/lib/environment.ts
@@ -66,6 +66,15 @@ setGlobalContext({
     if (DEBUG) {
       let { id } = options;
 
+      if (id === 'argument-less-helper-paren-less-invocation') {
+        throw new Error(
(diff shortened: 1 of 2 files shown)


rwjblue (approved):

mixonic:
I'm taking @rwjblue's approval and merging. I have opened an issue upstream in Glimmer-VM for the cleanup to either improve this error message or change functionality. See https://github.com/glimmerjs/glimmer-vm/issues/1363

@chancancode I would still value your opinion, and I'm happy to do some more followup if you think it is critical for 4.0.

PR #19849, opened 2021-11-16
Fix typo in error messages

danwenzel:
Cannot is one word


rwjblue:
@danwenzel - Would you mind rebasing? Looks like the CI jobs never actually ran here....

rwjblue:
Oh, also looks like a few tests will need to be updated as well. The new error was introduced in https://github.com/emberjs/ember.js/commit/298a086c06085756caa1ac4c6f340f77061534e2 and there are some test assertions that are trying to validate `Can not`...

danwenzel:
sorry folks, my email notifications must have been messed up, and I missed these comments.

I happened to notice this typo is fixed now, and had forgotten all about this.  Thanks for bringing it across the line!

PR #19850, opened 2021-11-19
Slight fix in the code documentation

pbernery:
The comment was probably referring an old version of the sample code.


locks (approved):
Thanks!

PR #19868, opened 2021-12-01
[BUGFIX release-3-28] Fix #19867

### LINKED ISSUE #19867: [Bug] Injecting Ember Data Store to Avoid Implicit Injection is Broken
### 🐞 Describe the Bug

In advance of implicit injections of the Ember Data's Store being removed we've been refactoring to explicit injections of the store when we need it. As of Ember v3.28.7 we're seeing errors to this effect:

```
this.store.findAll is not a function
```

Note that this is not unique to `findAll` — any method on Store will fail.

### 🔬 Minimal Reproduction

1. Clone https://github.com/jherdman/ember-data-store-sadness
2. `npm install && npm start`
3. Visit `http://localhost:4200`
4. Open your dev console, see error

### 😕 Actual Behavior

Page does not load, an error is raised as described above.

### 🤔 Expected Behavior

1. No deprecation warning is raised
2. Store works as expected

### 🌍 Environment

- Ember: - 3.28.7
- Node.js/npm: - v16.13.0
- OS: - macOS 12.0.1
- Browser: - Firefox 94.0.2

### ➕ Additional Context

Bug does not occur in Ember v3.28.6

mixonic:
Fixes https://github.com/emberjs/ember.js/issues/19867. Confirmed fix on the reproductions at https://github.com/jherdman/ember-data-store-sadness/pull/1 and https://github.com/empress/empress-blog/pull/148.

A build of this branch can be downloaded at https://gist.github.com/mixonic/fe502b7f84eba571b9bde80e78f1b720. You can try out the branch in your application with that build.

The way most Ember apps are setup in late 3.28 (an implicit injection of `store` but also an explicit injection) was not under test. This patch adds a test, and fixes a bug when applications use that pattern.

The prior implementation of the `store` injection deprecation on routes (added in #19854) used a wrapper `DeprecatedStoreInjection` in some cases which was unwrapped by the store getter. A deprecation implemented on `CoreObject` for when a explicit and implicit deprecation mis-match conflicted with that logic.

Here, refactor away from the `DeprecatedStoreInjection` wrapper to instead use a WeakSet to track injected instances. This does mean the deprecation is only going to fire in newer browsers (not in IE11), but I think that is acceptable.

### DIFF
--- a/packages/@ember/-internals/container/index.ts
+++ b/packages/@ember/-internals/container/index.ts
@@ -11,5 +11,5 @@ export {
   getFactoryFor,
   setFactoryFor,
   INIT_FACTORY,
-  DeprecatedStoreInjection,
+  deprecatedStoreInjections,
 } from './lib/container';
--- a/packages/@ember/-internals/container/lib/container.ts
+++ b/packages/@ember/-internals/container/lib/container.ts
@@ -49,12 +49,8 @@ if (DEBUG) {
   }
 }
 
-export class DeprecatedStoreInjection {
-  store: unknown;
-  constructor(store: unknown) {
-    this.store = store;
-  }
-}
+export const deprecatedStoreInjections =
+  DEBUG && window.WeakSet ? new window.WeakSet<object>() : undefined;
 
 export interface ContainerOptions {
   owner?: Owner;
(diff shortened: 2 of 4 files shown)


SergeAstapov:
@mixonic confirmed via https://github.com/ember-learn/ember-cli-addon-docs/pull/1030/checks and works like a charm on my local

ef4 (approved):

PR #15667, opened 2017-09-14
[BUGFIX beta] allow numeric `0` key to work with `get` helper

bekzod:
fixes #15666

PR #15366, opened 2017-06-16
[BUGFIX beta] Allow numeric keys for the `get` helper.

rwjblue:
Fixes #13296
Closes #13311 

/cc @duggiefresh

Note: This commit is straight out of a rebase of #13311. Normally I would just push to the branch, but that PR predates github's feature that allows maintainers to do this.

PR #19871, opened 2021-12-06
[DOC] Document get helper array element access with second arg of type number

jamescdavis:
I've updated the docs for the `get` helper to indicate that the 2nd argument can be a number. This has been supported since Ember 2.14 and was added in #15366 (along with a test to make sure it continues to work). This test [still exists in Ember 4](https://github.com/emberjs/ember.js/blob/v4.0.1/packages/%40ember/-internals/glimmer/tests/integration/helpers/get-test.js#L62-L85).

I've also updated the docs to indicate that the `get` helper can be used for array element access via index with a second dynamic arg that has a numeric value. #15667 fixed and added a test specifically for array element access via numeric index, which is [still present](https://github.com/emberjs/ember.js/blob/v4.0.1/packages/%40ember/-internals/glimmer/tests/integration/helpers/get-test.js#L132-L146), and #16218 fixed this for a _const_ number second arg and added a test specifically for array index access via a const number, which is [also still present](https://github.com/emberjs/ember.js/blob/v4.0.1/packages/%40ember/-internals/glimmer/tests/integration/helpers/get-test.js#L87-L105).

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/helpers/get.ts
+++ b/packages/@ember/-internals/glimmer/lib/helpers/get.ts
@@ -3,8 +3,9 @@
 */
 
 /**
-  Dynamically look up a property on an object. The second argument to `{{get}}`
-  should have a string value, although it can be bound.
+  Dynamically look up a property on an object or an element in an array.
+  The second argument to `{{get}}` should have a string or number value,
+  although it can be bound.
 
   For example, these two usages are equivalent:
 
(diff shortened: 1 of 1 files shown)


jamescdavis:
Guides [now reflect this usage](https://github.com/ember-learn/guides-source/pull/1750): https://guides.emberjs.com/release/components/helper-functions/#toc_the-get-helper

jamescdavis:
@rwjblue, tagging you for visibility 🙂

rwjblue (approved):

Issue #19518, opened 2021-04-30
[Bug Canary] class helper invocation fails in dev mode


patricklx:
### 🐞 Describe the Bug
the invocation of the helper fails here https://github.com/emberjs/ember.js/blob/master/packages/%40ember/-internals/glimmer/lib/helper.ts#L194

before its also checked if .class actually exists or not. I guess it should also happen in getDebugName
https://github.com/emberjs/ember.js/blob/master/packages/%40ember/-internals/glimmer/lib/helper.ts#L164

pzuraq:
Can you provide a reproduction or some steps for producing this error? As is, I'm not sure how to diagnose/fix this issue.

patricklx:
okay, right, i just noticed this happens when using helpers coming from local variables.
https://ember-twiddle.com/e03f28ed71a45e0e9daf7da76ff295fd?numColumns=2&openFiles=templates.application%5C.hbs%2Chelpers.my-helper%5C.js
I was unable to get it to build in ember-twiddle...
https://gitpod.io#snapshot/a23f617d-5ef3-4057-aa4a-24692b15fc5f

NullVoxPopuli:
Is this still an issue? looks like a bugfix PR was merged

PR #19878, opened 2021-12-18
[Bugfix release] allow class-based helpers to work in strict-mode

### LINKED ISSUE #19877: [Bug] class-based helpers throw weird error in strict mode (getDebugName)
### 🐞 Describe the Bug
Error wasn't expected, but I think phrasing of the error could be improved?

image

### 🔬 Minimal Reproduction

1. define class based helper in my-app/helpers
2. using ember-template-imports (or strict mode some other way), define this template:
  ```js
  import MyHelper from 'my-app/helpers/my-helper';
  
  <template>
    {{#let (MyHelper 1 2) as |result|}}
      {{result}}
    {{/let}}
  </template>
  ```

### 😕 Actual Behavior
image

actual helper is:
```js
import { getOwner } from '@ember/application';
import Helper from '@ember/component/helper';

export default class GetService extends Helper {
  compute([name]: [string]) {
    return getOwner(this).lookup(`service:${name}`);
  }
}
```

### 🤔 Expected Behavior
primarily, I expected this to work:
```js
import service from 'limber/helpers/service';

<template>
  {{#let (service 'editor') as |context|}}
```
but a more helpful error would have been preferable, given the situation?

### 🌍 Environment

- Ember: - 4.0.1
- Node.js/npm: - 16.13.0 / 8.1.3 / yarn 1.22.17
- OS: - Ubuntu 21.10
- Browser: - FireFox

NullVoxPopuli:
Resolves https://github.com/emberjs/ember.js/issues/19877

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/helper.ts
+++ b/packages/@ember/-internals/glimmer/lib/helper.ts
@@ -179,7 +179,7 @@ class ClassicHelperManager implements HelperManager<ClassicHelperStateBucket> {
   }
 
   getDebugName(definition: ClassHelperFactory) {
-    return getDebugName!(definition.class!['prototype']);
+    return getDebugName!((definition.class || definition)!['prototype']);
   }
 }
 
--- a/packages/@ember/-internals/glimmer/tests/integration/custom-helper-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/custom-helper-test.js
@@ -1,0 +1,78 @@
+import { RenderingTestCase, moduleFor, strip } from 'internal-test-helpers';
+
+import { precompileJSON } from '@glimmer/compiler';
+import { getTemplateLocals } from '@glimmer/syntax';
+import { createTemplateFactory } from '@ember/template-factory';
+
+import { Helper } from '../../index';
+import { setComponentTemplate } from '@glimmer/manager';
+import { templateOnlyComponent } from '@glimmer/runtime';
+
+// eslint-disable-next-line ember-internal/require-yuidoc-access
+/**
+ * The template-compiler does not support strict mode at this time.
+ * precompile from ember-template-compiler returns a string and
+ * not a template-factory, so it doesn't help with strict-mode testing.
+ *
+ * We also can't import from `@ember/template-compiler` because it
+ * doesn't exist to this kind of test, otherwise we'd be able to use
+ * precompileTemplate, which would be perfect :D
+ *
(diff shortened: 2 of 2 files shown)


smoke-tests/ember-test-app/tests/integration/strict-mode-test.js
```diff
@@ -0,0 +1,28 @@
...
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import { precompileTemplate } from '@ember/template-compilation';
+
+import service from 'ember-test-app/helpers/service';
+
+module('tests/integration/components/gjs', function (hooks) {
+  setupRenderingTest(hooks);
+
+  test('it works with ember helpers', async function (assert) {
```
NullVoxPopuli, line 11: I added a whole app for just this tset because I couldn't find any other place where I'd have access to all the thing that this test requires.

If such a place exists, I'm happy to move this test there.
I think having a full app though to make sure things are roughly still wired up ok is good

packages/@ember/-internals/glimmer/lib/helper.ts
```diff
@@ -179,7 +179,7 @@ class ClassicHelperManager implements HelperManager<ClassicHelperStateBucket> {
   }
 
   getDebugName(definition: ClassHelperFactory) {
-    return getDebugName!(definition.class!['prototype']);
+    return getDebugName!((definition.class || definition)!['prototype']);
```
NullVoxPopuli, line 182: when a class-based helper is used in strict mode, there is no class property

patricklx:
might this be related? https://github.com/emberjs/ember.js/issues/19518

NullVoxPopuli:
@patricklx I think it might be!! yeah!

mixonic:
This fix is probably ok, but I would like it tested in the existing test harness and not an entirely new suite. Adding a whole new kind of smoke test should a) be decoupled into its own PR and b) does not preclude ensuring regressions have coverage added in the internal non-smoke test suite.

That is the main blocker I see here. Is this a bugfix-release? Bugfix-lts?

NullVoxPopuli:
> Is this a bugfix-release? Bugfix-lts?

It's a bug on 4.0+, afaict (I haven't checked earlier) -- so, bugfix-release?

>  but I would like it tested in the existing test harness

where would a test like this go?

> a) be decoupled into its own PR

can do!

NullVoxPopuli:
I'm having issues with the test. any help is much appreciated <3

kategengler:
Merging on approvals from #19884

PR #19882, opened 2021-12-24
Initial implementation of the unique-id helper

wycats:
The implementation works, and still needs:

- [x] Docs
- [x] Feature Flag


locks:
@wycats @MelSumner what do we need to land this?

Windvis:
I created a PR (which targets this branch) that adds the feature flag.

Turbo87:
Bildschirmfoto 2022-06-14 um 12 18 09

is it expected that IDs produced by this helper don't work with `querySelectorAll()`? 🤔

PR #19894, opened 2022-01-11
Remove support for non-Octane blueprints

cafreeman:
The current blueprints test suite covers cases for both octane and non-octane blueprints, even though non-octane blueprints are no longer supported. As a result, [this PR to upgrade ember-cli to 4.1.0](https://github.com/emberjs/ember.js/pull/19893) is blocked because all of the non-octane blueprint tests fail (since the CLI no longer generates non-Octane files).

This PR unblocks the ember-CLI upgrades by removing all of the non-Octane blueprint tests. In addition, it also removes all of the pre-Octane blueprints and fixtures since they are no longer supported and can't be generated in the current version of Ember CLI.


locks (approved):

blueprints/component-addon/index.js
```diff
@@ -4,9 +4,8 @@ const path = require('path');
 const stringUtil = require('ember-cli-string-utils');
 const getPathOption = require('ember-cli-get-component-path-option');
 const normalizeEntityName = require('ember-cli-normalize-entity-name');
-const useEditionDetector = require('../edition-detector');
```
rwjblue, line 7: I think this file (`blueprints/edition-detector.js`) still needs to be deleted, right?

node-tests/blueprints/component-class-test.js
```diff
@@ -52,213 +52,134 @@ describe('Blueprint: component-class', function () {
...
+        expect(_file('app/components/foo.js')).to.equal(glimmerComponentContents);
       });
     });
 
-    it('component-class x-foo --path foo', function () {
-      return emberGenerateDestroy(['component-class', 'x-foo', '--path', 'foo'], (_file) => {
-        expect(_file('app/components/x-foo.js')).to.equal(fixture('component/component-dash.js'));
-      });
+    // classic default
+    it('component-class foo --component-structure=classic --component-class=@ember/component', function () {
```
rwjblue, line 60: Can you double check what the RFC says RE: "--component-structure=classic"? I think we probably don't need that flag value anymore...

node-tests/blueprints/controller-test.js
```diff
@@ -118,32 +86,19 @@ describe('Blueprint: controller', function () {
 
       it('controller foo.js --pod podModulePrefix', function () {
         return emberGenerateDestroy(['controller', 'foo.js', '--pod'], (_file) => {
-          expect(_file('app/pods/foo.js/controller.js')).to.not.exist;
-          expect(_file('tests/unit/pods/foo.js/controller-test.js')).to.not.exist;
```
rwjblue, line 122: Why remove these?

node-tests/blueprints/route-test.js
```diff
@@ -1058,27 +559,5 @@ describe('Blueprint: route', function () {
         );
       });
     });
-
-    describe('ember-page-title is installed', function () {
```
rwjblue, line 1062: Why are these `ember-page-title` related tests deleted (are they brought back in another file somewhere)?

node-tests/blueprints/route-test.js
```diff
@@ -480,516 +474,17 @@ describe('Blueprint: route', function () {
...
-            fixture('route-test/default.js')
-          );
-
-          expect(file('app/router.js')).to.contain("this.route('foo')");
-          expect(file('app/router.js')).to.not.contain("this.route('foo.js')");
-        }).then(() => {
-          expect(file('app/router.js')).to.not.contain("this.route('foo')");
-        });
-      });
-    });
```
rwjblue, line 759: We still need these pods tests

rwjblue (changes requested):

node-tests/blueprints/route-test.js
cafreeman: Since `ember-page-title` is included by default now, I deleted these tests and instead added page title-related assertions to *all* of the route tests since all routes generate templates with page title anyway.

blueprints/component-addon/index.js
cafreeman: Good catch

node-tests/blueprints/controller-test.js
cafreeman: 🤔 looking at this diff, it seems like there were some copy/paste shenanigans going on with the pod tests. gonna take another pass on this file

node-tests/blueprints/route-test.js
bertdeblock: Shouldn't we still have tests that verify that the `page-title` helper is not present when `ember-page-title` is not installed? I could be wrong, but it seems all tests now assume the helper is always present?

node-tests/blueprints/route-test.js
cafreeman: they still exist, they're just higher up in the file https://github.com/cafreeman/ember.js/blob/cfreeman/remove-non-octane-blueprint-tests/node-tests/blueprints/route-test.js#L240-L282

node-tests/blueprints/route-test.js
cafreeman: Hmm, I'm not sure. Since `ember-page-title` is installed by default, and these are blueprint tests, the blueprints in a new Ember app *will* always have page-title present unless people go and remove it after the fact. I guess we could still include a test or two for the case where it isn't installed just to make sure that it doesn't break if users remove it.

bertdeblock: I think having at least one test that still verifies the absence of the helper when `ember-page-title` is not installed is a good idea. These blueprints will also be used by people who upgrade their app and might have deleted `ember-page-title` in the past because e.g. they use `ember-cli-head` or a different approach. Just a suggestion though! Someone from the core team should decide.

rwjblue: Ya, agree with @bertdeblock here. If the project doesn't have ember-page-title we need to not add helper invocations in the blueprint.

node-tests/blueprints/component-class-test.js
cafreeman: For anyone referencing this later, after discussing with @rwjblue we decided this would be fine in a separate PR.

rwjblue (approved):

PR #19895, opened 2022-01-12
Improve types for DeprecationOptions

chriskrycho:
The previous design allowed invalid values for `DeprecationOptions`:

- `{}` (that is, an empty object)
- `{ enabled: 'v4.1.0' }` (that is, without `available`)

Both of these violate [the text of the RFC](https://emberjs.github.io/rfcs/0649-deprecation-staging.html#deprecate)! The new design *requires* that at least `available` be present, and that `enabled` cannot be present *without* `available`:

> If a value for a stage if present, the value of all previous stages must also be present.

This PR also removes the exported key names for the stages, which are implicit in the interface-based API and which are currently unused outside this module.

---

(Noticed this while working on the Ember v4 types on DefinitelyTyped, and made this change there as well.)

### DIFF
--- a/packages/@ember/-internals/container/lib/registry.ts
+++ b/packages/@ember/-internals/container/lib/registry.ts
@@ -159,7 +159,7 @@ export default class Registry implements IRegistry {
    @param {Object} options
    @return {Container} created container
    */
-  container(options?: ContainerOptions) {
+  container(options?: ContainerOptions): Container {
     return new Container(this, options);
   }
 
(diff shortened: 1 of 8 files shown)


chriskrycho:
Ha! This is failing because a bunch of places within Ember are invoking it using the illegal invocations! I WIN!

rwjblue (approved):

chriskrycho:
Aaaaand we're ✅!

rwjblue (approved):

PR #19900, opened 2022-01-18
Remove TSLint

### LINKED ISSUE #19899: Tooling: migrate from TSLint to ESLint for TypeScript internals
TSLint has been deprecated for a couple years. It’s well past time to move over to the `@typescript-eslint`-powered tooling.

Tasks that need doing here (and which most anyone in the community should be able to do):

0. Add the `@typescript-eslint/parser`, `@typescript-eslint/plugin`, etc. dependencies as documented on [the TypeScript ESLint website](https://typescript-eslint.io).
1. Create a mapping from [the tslint config](https://github.com/emberjs/ember.js/blob/master/tslint.json) to a modified version of the `@typescript-eslint/recommended` config, which should probably be our baseline.
2. Add the correct configuration to the existing [repo-wide ESLint config](https://github.com/emberjs/ember.js/blob/master/.eslintrc.js), using `@typescript-eslint/parser` for `.ts` files only so that there’s no impact on `.js` files from switching parsers.
3. Create a PR landing the above!

Given the rules in use, this should be fairly straightforward; the majority of the work is in step (1), and in particular making sure that we set things to the corresponding *level* of warning.

Feel free to reach out to Typed Ember folks (@dfreeman, @jamescdavis, and me) [on Discord](https://discord.com/channels/480462759797063690/484421406659182603) as well as current maintainers of Ember.

wagenet:
Fixes https://github.com/emberjs/ember.js/issues/19899


package.json
```diff
@@ -148,8 +146,7 @@
     "simple-dom": "^1.4.0",
     "testem": "^3.6.0",
     "testem-failure-only-reporter": "^0.0.1",
-    "tslint": "^6.1.3",
```
chriskrycho, line 151: :fire:

packages/@ember/-internals/utils/lib/mandatory-setter.ts
```diff
@@ -105,7 +105,7 @@ if (DEBUG) {
     if (setters !== undefined && setters[keyName] !== undefined) {
       Object.defineProperty(obj, keyName, setters[keyName]);
 
-      setters[keyName] = undefined;
+      delete setters[keyName];
```
chriskrycho, line 108: Note: I *think* we probably don’t mind because this is in debug-only code, and it’s totally reasonable from a satisfying-the-compiler POV, but this is apt to throw the runtime into deopt mode for the `setters` object, because `delete` totally messes up the shape of the object and the VM (AFAIU) has to basically go into "treat it like a dictionary" mode instead of treating it like a class with a known shape. Most likely this is *already* happening in this spot given all the shenanigans involved with this debug feedback, but worth noting.

chriskrycho (approved):
Love it. Thank you!

PR #19902, opened 2022-01-19
Add esplint to prevent new eslint warnings

wagenet:
It has been suggested that https://github.com/lint-todo/eslint-formatter-todo would be a better choice than esplint. However, ember-template-lint also uses this library. That means that both can't be run in parallel since they both try to write to the same location. A good improvement to lint-todo would be to allow for a configurable output directory. Since this hasn't yet happened, I propose merging this then switching over when lint-todo is updated.


locks:
@wagenet do you still intend to pursue this work?

wagenet:
@locks I think the idea is still good, but would have to re-evaluate the approach at this point.

kategengler:
Is this still relevant?

wagenet:
I think work like this is worthwhile if we want to burn down our one-off disabled lints.

NullVoxPopuli:
@wagenet any chance you want to do some of the eslint-surpressions work that you've been doing internally at work on this PR? <3

kategengler:
Moving to draft since it is not ready. I'm also wondering if we'd use the eslint bulk suppressions instead now.

wagenet:
Bulk suppressions is probably the way to go.

PR #19903, opened 2022-01-19
Resolve some eslint warnings

wagenet:
These warnings aren't actually all currently enabled, but would be added with #19902.


NullVoxPopuli (approved):
noice

locks (approved):

PR #19905, opened 2022-01-19
Refactoring

wagenet:
The goal here is to avoid TS warnings and risky casts.

### DIFF
--- a/packages/@ember/object/compat.ts
+++ b/packages/@ember/object/compat.ts
@@ -6,6 +6,7 @@ import {
   isElementDescriptor,
   setClassicDecorator,
 } from '@ember/-internals/metal';
+import type { ElementDescriptor } from '@ember/-internals/metal/lib/decorator';
 import { assert } from '@ember/debug';
 import { consumeTag, tagFor, track, UpdatableTag, updateTag } from '@glimmer/validator';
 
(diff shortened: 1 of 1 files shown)


packages/@ember/object/compat.ts
```diff
@@ -148,25 +156,33 @@ export function dependentKeyCompat(
...
-    'The @dependentKeyCompat decorator must be applied to getters/setters when used in native classes',
-    (desc !== null && typeof desc!.get === 'function') || typeof desc!.set === 'function'
-  );
-
-  return wrapGetterSetter(target, key!, desc!);
 }
 
 setClassicDecorator(dependentKeyCompat as Decorator);
+
+function isPropertyDescriptor(value: unknown): value is PropertyDescriptor {
```
NullVoxPopuli, line 173: should this have its own tests?, or are the existing tests sufficient?

wagenet: It's not public, so I don't think it needs its own tests.

chancancode (approved):
Not sure how much it matters but the new code won't get debug-stripped when you write it like this

packages/@ember/object/compat.ts
```diff
@@ -6,6 +6,7 @@ import {
   isElementDescriptor,
   setClassicDecorator,
 } from '@ember/-internals/metal';
+import { ElementDescriptor } from '@ember/-internals/metal/lib/decorator';
```
chriskrycho, line 9:
```suggestion
import type { ElementDescriptor } from '@ember/-internals/metal/lib/decorator';
```

packages/@ember/object/compat.ts
```diff
@@ -121,23 +122,30 @@ let wrapGetterSetter = function (target: object, key: string, desc: PropertyDesc
...
+    assert(
+      'The @dependentKeyCompat decorator must be applied to getters/setters when used in native classes',
+      desc != null && (typeof desc.get === 'function' || typeof desc.set === 'function')
+    );
+
+    return wrapGetterSetter(target, key, desc);
+  } else {
+    const desc = args[0];
+
+    assert('expected valid PropertyDescriptor', isPropertyDescriptor(desc));
```
chriskrycho, line 143: Per @chancancode's comment, it might be worth inlining this. It's not in a hot path, but enough of these *do* add up bytes-wise.

PR #19906, opened 2022-01-19
Force broccoli-typescript-compiler to use newer TypeScript

wagenet:
Resolves the warnings visible here https://github.com/emberjs/ember.js/actions/runs/1715562324.

See also https://github.com/tildeio/broccoli-typescript-compiler/pull/110

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,7 +101,7 @@
     "broccoli-rollup": "^2.1.1",
     "broccoli-source": "^3.0.1",
     "broccoli-string-replace": "^0.1.2",
-    "broccoli-typescript-compiler": "^7.0.0",
+    "broccoli-typescript-compiler": "tildeio/broccoli-typescript-compiler#typescript-peer",
     "broccoli-uglify-sourcemap": "^4.0.0",
     "common-tags": "^1.8.2",
     "core-js": "^2.6.5",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -148,6 +148,9 @@
     "testem-failure-only-reporter": "^0.0.1",
     "typescript": "^4.5.4"
   },
+  "resolutions": {
+    "typescript": "^4.5.4"
+  },
```
chriskrycho, line 153: Question about this: why not update `broccoli-typescript-compiler` instead?

wagenet: We're on the latest version. That said, it would make sense to send a PR to the library.

wagenet: Maybe https://github.com/tildeio/broccoli-typescript-compiler/pull/110 ?

chriskrycho: Sorry, yeah, I meant “make the change upstream” but that was totally unclear in my phrasing.

krisselden:
why not update it in broccoli-typescript-compiler?

wagenet:
@krisselden see https://github.com/tildeio/broccoli-typescript-compiler/pull/110

wagenet:
See #19938.

PR #19907, opened 2022-01-20
Future public types for @ember/application

wagenet:


packages/@ember/application/instance.d.ts
```diff
@@ -0,0 +1,37 @@
...
+    location?: string;
+    isInteractive?: boolean;
+  });
+
+  toEnvironment(): Record<string, unknown>;
+}
+
+export default class ApplicationInstance implements EngineInstance {
+  lookup<T>(fullName: string, options?: LookupOptions): T | undefined;
+  factoryFor<T, C>(fullName: string, options?: LookupOptions): Factory<T, C> | undefined;
```
chriskrycho, line 25: This has the same issue. Basically, the rule of thumb is: if the generic is only used once, it's wrong in some way—because it's not actually constraining anything. It is the type system equivalent of an unused parameter to a function.

```suggestion
  factoryFor(fullName: string, options?: LookupOptions): Factory<unknown, unknown> | undefined;
```

packages/@ember/application/instance.d.ts
```diff
@@ -0,0 +1,37 @@
...
+  register<T, C>(fullName: string, factory: Factory<T, C>, options?: object): void;
```
chriskrycho, line 26: To elaborate slightly: here, there *is* a constraint here, and TypeScript will infer it correctly in usage, so it's valid. That's the key difference with this vs. the previous ones.

packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -93,6 +93,12 @@ export function hasDefaultSerialize(route: Route): boolean {
   return route.serialize === defaultSerialize;
 }
 
+function getOwnerChecked(route: Route): Owner {
+  let owner = getOwner(route);
+  assert('Route is unexpectedly missing an owner', owner);
+  return owner;
+}
```
chriskrycho, line 100: One concern: this seems good, but it looks like we'll end up shipping a bunch of extra copies of it to prod this way?

packages/@ember/-internals/owner/index.ts
```diff
@@ -30,12 +30,15 @@ import EngineInstance from '@ember/engine/instance';
 export interface Owner {
   lookup<T>(fullName: string, options?: LookupOptions): T | undefined;
   factoryFor<T, C>(fullName: string, options?: LookupOptions): Factory<T, C> | undefined;
```
chriskrycho, line 32: Ah, is this where the types are coming from? Ugh. We should think about how to iterate on it, because as it is it lets users of the interface simply lie to the compiler ("this is totally a `Something`, yep, just trust me") and my experience is that those are *often* mistaken: not least because they do not and cannot catch typos. In this case, the types are at least *slightly* safer in that they consistently return `| undefined`, which is the most common kind of error produced by these kinds of types—but this can and will still produce errors down the line. 😩 If we need to ship `@ember/application` supporting this for now and then make a follow-on cross-cutting change to approach it another way, that seems reasonable.

We should also think about the best way to approach this: the use of a `Registry`-style API would solve it, but that has its own tradeoffs and downsides.

packages/@ember/-internals/glimmer/lib/setup-registry.ts
```diff
@@ -17,7 +18,9 @@ export function setupApplicationRegistry(registry: Registry): void {
   // association won't leak
   registry.register('service:-dom-builder', {
     create(props) {
-      let env = getOwner(props).lookup('-environment:main') as { _renderMode: string };
+      let owner = getOwner(props);
+      assert('DomBuilderService is unexpectedly missing an owner', owner);
+      let env = owner.lookup('-environment:main') as { _renderMode: string };
```
chriskrycho, line 23: Do you mind adding a `// SAFETY` comment here explaining why we trust this to be safe?

chriskrycho (changes requested):
👍🏼 overall, couple of small tweaks.

packages/@ember/application/instance.d.ts
```diff
@@ -0,0 +1,37 @@
...
+    rootElement?: string | Element;
+    location?: string;
+    isInteractive?: boolean;
+  });
+
+  toEnvironment(): Record<string, unknown>;
+}
+
+export default class ApplicationInstance implements EngineInstance {
+  lookup<T>(fullName: string, options?: LookupOptions): T | undefined;
```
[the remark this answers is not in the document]
wagenet: Yeah, that makes sense. I just copied from https://github.com/emberjs/ember.js/blob/master/packages/@ember/-internals/owner/index.ts#L31.

packages/@ember/application/instance.d.ts
wagenet: There's a constraint in `Factory`.

packages/@ember/-internals/glimmer/lib/setup-registry.ts
wagenet: I'm actually not sure why offhand. I didn't write this, just got reformatted.

packages/@ember/-internals/routing/lib/system/route.ts
wagenet: Yeah, I could just inline the assert, it was annoying, heh.

packages/@ember/application/instance.d.ts
chriskrycho: Because it is in the return position, it will not be inferred (it cannot be!) and therefore it acts unconstrained.

wagenet: We’ll I removed it anyway.

packages/@ember/-internals/glimmer/lib/setup-registry.ts
chriskrycho: Ahhh, I see. Fair!

packages/@ember/-internals/routing/lib/system/route.ts
wagenet: Inlined.

chriskrycho (approved):
👏🏼

PR #19909, opened 2022-01-20
Future public types for @ember/array

wagenet:
Includes #19910 

We should probably consider whether we actually want to make these all fully public. Many of these usages are probably not recommended anymore. That said, if we choose not to make this public, we should still make sure we use the internal improvements from this PR.


wagenet:
Something very odd is happening here where end up with empty exports for `utils`.

packages/@ember/-internals/glimmer/lib/helper.ts
```diff
@@ -76,11 +79,26 @@ export interface SimpleHelper<T = unknown> {
...
+  /**
+    Override this function when writing a class-based helper.
+
+    @method compute
+    @param {Array} params The positional arguments to the helper
+    @param {Object} hash The named arguments to the helper
+    @public
+    @since 1.13.0
+  */
+  compute(params: unknown[], hash: object): unknown;
```
chriskrycho, line 92: This is totally reasonable and good for now; we'll probably want to revisit it once we have consensus around the Glimmer Component type changes, as we'll likely want to align roughly with that for a Glint-compatible signature here, but I don't want to try to do anything but this for *now*.

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
```diff
@@ -0,0 +1,103 @@
+type Value<T, K extends string> = K extends keyof T ? T[K] : unknown;
```
chriskrycho, line 1: A thing I'll check on tomorrow (when hopefully finishing the v4 DT push!)—we should make sure this (and similar on the `@ember/application` PR) aligns with the updated types there. For reference, PR is [here][DT]. The types shouldn't change between now and landing it; it's all just config stuff to get the major version landed.

[DT]: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/58060

chriskrycho (commented):
Didn't get to finish this, did get a start. Looks totally reasonable so far, will have a few small tweaks at most I expect!

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export type NatifiedArray<T> = T[] & NativeArray<T>;
+
+interface EmberArray<T> {
+  length: number;
+  objectAt(idx: number): T | undefined;
+}
+declare class EmberArray<T> {
```
chriskrycho, line 9: Why the separate interface merged with the class?

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
```diff
@@ -0,0 +1,103 @@
...
+  objectsAt(indexes: number[]): Array<T | undefined>;
```
chriskrycho, line 10: Ah, this is something our current types on DT miss!

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
```diff
@@ -0,0 +1,103 @@
...
+    callback: (this: Target, item: T, index: number, arr: this) => boolean,
+    target?: Target
+  ): NatifiedArray<T>;
+  reject<Target>(
+    callback: (this: Target, item: T, index: number, arr: this) => boolean,
+    target?: Target
+  ): NatifiedArray<T>;
+  filterBy(key: string): NatifiedArray<T>;
+  rejectBy(key: string): NatifiedArray<T>;
+  find<Target = null>(
```
chriskrycho, line 37: We should make this match the types from `Array`, so `Target` should default to `void` rather than `null`.

```suggestion
  find<Target = void>(
```

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
```diff
@@ -0,0 +1,103 @@
...
+    callback: (this: Target, item: T, index: number, arr: this) => boolean,
+    target?: Target
+  ): T | undefined;
+  findBy<K extends string>(key: K, value?: Value<T, K>): T | undefined;
+  every<Target = null>(
```
chriskrycho, line 42:
```suggestion
  every<Target = void>(
```

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
```diff
@@ -0,0 +1,103 @@
...
+    callback: (this: Target, item: T, index: number, arr: this) => boolean,
+    target?: Target
+  ): boolean;
+  isEvery<K extends string>(key: K, value?: Value<T, K>): boolean;
+  any<Target = null>(
```
chriskrycho, line 47:
```suggestion
  any<Target = void>(
```

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
```diff
@@ -0,0 +1,103 @@
...
+  includes(object: T, startAt?: number): boolean;
+  sortBy(key: string): T[];
+  uniq(): NatifiedArray<T>;
+  uniqBy(key: string): NatifiedArray<T>;
+  without(value: T): NatifiedArray<T>;
+}
+
+export default EmberArray;
+
+interface MutableArray<T> extends EmberArray<T> {
```
chriskrycho, line 71: Notably, I believe this is backwards of the relationship in Ember itself (per the docs, and per the current types). We should flip it to match.

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
```diff
@@ -0,0 +1,103 @@
...
+  setObjects(object: T[]): this;
+  removeObject(object: T): this;
+  removeObjects(objects: T[]): this;
+  addObject(obj: T): this;
+  addObjects(objects: T[]): this;
+}
+
+export { MutableArray };
+
+export type NativeArray<T> = Array<T> & MutableArray<T>;
```
chriskrycho, line 95: We can write this as an interface instead:

```ts
interface NativeArray<T> extends Array<T>, MutableArray<T> {}
```

In general, this is preferred over intersections (and generally produces better diagnostics in type errors, etc.).

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export function A<T>(
+  arr?: T
+): T extends Array<infer TT> ? TT[] & NativeArray<TT> : unknown[] & NativeArray<unknown>;
```
chriskrycho, line 99: Two notes:

- the `TT` is a bit unusual in my experience; maybe `A` instead?
- Given how `NativeArray` is defined, I think the intersections are redundant

```suggestion
): T extends Array<infer A> ? NativeArray<A> : NativeArray<unknown>;
```

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
```diff
@@ -0,0 +1,103 @@
...
+  without(value: T): NatifiedArray<T>;
+}
+
+export default EmberArray;
+
+interface MutableArray<T> extends EmberArray<T> {
+  replace(idx: number, amt: number, objects: T[]): void;
+}
+
+declare class MutableArray<T> extends EmberArray<T> {
```
chriskrycho, line 75: Same question as above about the separate interface vs. class body.

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
```diff
@@ -0,0 +1,103 @@
+type Value<T, K extends string> = K extends keyof T ? T[K] : unknown;
+
+export type NatifiedArray<T> = T[] & NativeArray<T>;
+
```
chriskrycho, line 4: This is the same as `NativeArray` itself.

```suggestion
```

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
```diff
@@ -0,0 +1,103 @@
...
+  forEach<Target>(
+    callback: (this: Target, item: T, index: number, arr: this) => void,
+    target?: Target
+  ): this;
+  getEach<K extends string>(key: K): NatifiedArray<Value<T, K>>;
+  setEach<K extends string>(key: K, value: Value<T, K>): this;
+  map<R, Target>(
+    callback: (this: Target, item: T, index: number, arr: this) => R,
+    target?: Target
+  ): NatifiedArray<R>;
```
chriskrycho, line 25: Unless we have a good reason to, we should make this match the `map` from JS `Array`—either manually, or by simply doing `Array<T>['map']` (which is what we do on DefinitelyTyped):

```suggestion
  map: Array<T>['map'];
```

While I am sympathetic to the interest of being more robust than TS' own types are, in our experience on Typed Ember that has actually caused more difficulty (e.g. because it makes assignability more complicated, etc.).

---

That being said: if we know that this returns `NativeArray` rather than `EmberArray`, this makes a fair bit of sense. In that case, it's conventional to use `U` for mapping functions (T → U).

```suggestion
  map<U, Target>(
    callback: (this: Target, item: T, index: number, arr: this) => R,
    target?: Target
  ): NativeArray<U>;
```

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
```diff
@@ -0,0 +1,103 @@
...
+
+interface EmberArray<T> {
+  length: number;
+  objectAt(idx: number): T | undefined;
+}
+declare class EmberArray<T> {
+  objectsAt(indexes: number[]): Array<T | undefined>;
+  get firstObject(): T | undefined;
+  get lastObject(): T | undefined;
+  slice(beginIndex?: number, endIndex?: number): NatifiedArray<T>;
```
chriskrycho, line 13:
```suggestion
  slice(beginIndex?: number, endIndex?: number): NativeArray<T>;
```

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
```diff
@@ -0,0 +1,103 @@
...
+  indexOf(object: T, startAt?: number): number;
+  lastIndexOf(object: T, startAt?: number): number;
+  forEach<Target>(
+    callback: (this: Target, item: T, index: number, arr: this) => void,
+    target?: Target
+  ): this;
+  getEach<K extends string>(key: K): NatifiedArray<Value<T, K>>;
```
chriskrycho, line 20:
```suggestion
  getEach<K extends string>(key: K): NativeArray<Value<T, K>>;
```

packages/@ember/array/type-tests/proxy.ts
```diff
@@ -0,0 +1,20 @@
+import EmberArray, { A, NatifiedArray } from '@ember/array';
```
chriskrycho, line 1:
```suggestion
import EmberArray, { A, NativeArray } from '@ember/array';
```

packages/@ember/array/type-tests/proxy.ts
```diff
@@ -0,0 +1,20 @@
...
+let content = A([foo]);
+
+// We can't infer the correct type through `create`;
+let proxy = ArrayProxy.create({ content }) as ArrayProxy<Foo, NatifiedArray<Foo>>;
+
+expectTypeOf(proxy).toMatchTypeOf<EmberArray<Foo>>();
+expectTypeOf(proxy).toMatchTypeOf<MutableArray<Foo>>();
+
+expectTypeOf(proxy.content).toEqualTypeOf<NatifiedArray<Foo>>();
+expectTypeOf(proxy.arrangedContent).toEqualTypeOf<NatifiedArray<Foo>>();
```
chriskrycho, line 20:
```suggestion
expectTypeOf(proxy.content).toEqualTypeOf<NativeArray<Foo>>();
expectTypeOf(proxy.arrangedContent).toEqualTypeOf<NativeArray<Foo>>();
```

packages/@ember/array/type-tests/proxy.ts
```diff
@@ -0,0 +1,20 @@
...
+import { expectTypeOf } from 'expect-type';
+
+class Foo {}
+
+let foo = new Foo();
+
+let content = A([foo]);
+
+// We can't infer the correct type through `create`;
+let proxy = ArrayProxy.create({ content }) as ArrayProxy<Foo, NatifiedArray<Foo>>;
```
chriskrycho, line 14:
```suggestion
let proxy = ArrayProxy.create({ content }) as ArrayProxy<Foo, NativeArray<Foo>>;
```

packages/@ember/array/type-tests/index.ts
```diff
@@ -0,0 +1,214 @@
+import EmberArray, { A, NatifiedArray, isArray, makeArray } from '@ember/array';
```
chriskrycho, line 1:
```suggestion
import EmberArray, { A, NativeArray, isArray, makeArray } from '@ember/array';
```

packages/@ember/array/type-tests/index.ts
```diff
@@ -0,0 +1,214 @@
...
+
+class Target {
+  foo = 2;
+}
+
+let target = new Target();
+
+expectTypeOf(arr).toMatchTypeOf<EmberArray<Foo>>();
+expectTypeOf(arr).toMatchTypeOf<MutableArray<Foo>>();
+expectTypeOf(arr).toEqualTypeOf<NatifiedArray<Foo>>();
```
chriskrycho, line 28:
```suggestion
expectTypeOf(arr).toEqualTypeOf<NativeArray<Foo>>();
```

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
```diff
@@ -0,0 +1,103 @@
...
+  setEach<K extends string>(key: K, value: Value<T, K>): this;
+  map<R, Target>(
+    callback: (this: Target, item: T, index: number, arr: this) => R,
+    target?: Target
+  ): NatifiedArray<R>;
+  mapBy<K extends string>(key: K): NatifiedArray<Value<T, K>>;
+  filter<Target>(
+    callback: (this: Target, item: T, index: number, arr: this) => boolean,
+    target?: Target
+  ): NatifiedArray<T>;
```
chriskrycho, line 30:
```suggestion
  ): NativeArray<T>;
```

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
```diff
@@ -0,0 +1,103 @@
...
+    callback: (this: Target, item: T, index: number, arr: this) => void,
+    target?: Target
+  ): this;
+  getEach<K extends string>(key: K): NatifiedArray<Value<T, K>>;
+  setEach<K extends string>(key: K, value: Value<T, K>): this;
+  map<R, Target>(
+    callback: (this: Target, item: T, index: number, arr: this) => R,
+    target?: Target
+  ): NatifiedArray<R>;
+  mapBy<K extends string>(key: K): NatifiedArray<Value<T, K>>;
```
chriskrycho, line 26:
```suggestion
  mapBy<K extends string>(key: K): NativeArray<Value<T, K>>;
```

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
```diff
@@ -0,0 +1,103 @@
...
+  filter<Target>(
+    callback: (this: Target, item: T, index: number, arr: this) => boolean,
+    target?: Target
+  ): NatifiedArray<T>;
+  reject<Target>(
+    callback: (this: Target, item: T, index: number, arr: this) => boolean,
+    target?: Target
+  ): NatifiedArray<T>;
+  filterBy(key: string): NatifiedArray<T>;
+  rejectBy(key: string): NatifiedArray<T>;
```
chriskrycho, line 36:
```suggestion
  ): NativeArray<T>;
  filterBy(key: string): NativeArray<T>;
  rejectBy(key: string): NativeArray<T>;
```

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
```diff
@@ -0,0 +1,103 @@
...
+  ): boolean;
+  isAny<K extends string>(key: K, value?: Value<T, K>): boolean;
+  reduce<V>(
+    callback: (summation: V, current: T, index: number, arr: this) => V,
+    initialValue?: V
+  ): V;
+  invoke<K extends string>(
+    methodName: K,
+    ...args: Value<T, K> extends (...args: any[]) => any ? Parameters<Value<T, K>> : unknown[]
+  ): NatifiedArray<Value<T, K> extends (...args: any[]) => any ? ReturnType<Value<T, K>> : unknown>;
```
chriskrycho, line 59:
```suggestion
  ): NativeArray<Value<T, K> extends (...args: any[]) => any ? ReturnType<Value<T, K>> : unknown>;
```

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
```diff
@@ -0,0 +1,103 @@
...
+  toArray(): T[];
+  compact(): NatifiedArray<Exclude<T, null>>;
```
chriskrycho, line 61:
```suggestion
  compact(): NativeArray<Exclude<T, null>>;
```

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
```diff
@@ -0,0 +1,103 @@
...
+  includes(object: T, startAt?: number): boolean;
+  sortBy(key: string): T[];
+  uniq(): NatifiedArray<T>;
+  uniqBy(key: string): NatifiedArray<T>;
+  without(value: T): NatifiedArray<T>;
```
chriskrycho, line 66:
```suggestion
  uniqBy(key: string): NativeArray<T>;
  without(value: T): NativeArray<T>;
```

packages/@ember/array/index.ts
```diff
@@ -0,0 +1,2 @@
+export { Array as default, NatifiedArray, isArray, A } from '@ember/-internals/runtime';
```
chriskrycho, line 1:
```suggestion
export { Array as default, NativeArray, isArray, A } from '@ember/-internals/runtime';
```

packages/@ember/-internals/utils/lib/make-array.ts
```diff
@@ -29,8 +29,7 @@ const { isArray } = Array;
  @return {Array}
  @private
  */
-function makeArray<T>(obj: T[]): T[];
-function makeArray<T>(obj: T): T[];
+function makeArray<T, TT>(obj: T): T extends TT[] ? T : T extends null | undefined ? [] : [T];
```
chriskrycho, line 32: Two notes:

- We don't actually need the `TT` param here at all, we can just 
- the last bit should be an array of `T` rather than a one-tuple of T, I think? Just by intent, I would be surprised to get back a tuple.

```suggestion
function makeArray<T>(obj: T): T extends Array<unknown> ? T : T extends null | undefined ? [] : T[];
```

packages/@ember/array/type-tests/index.ts
```diff
@@ -0,0 +1,214 @@
...
+  expectTypeOf(index).toEqualTypeOf<number>();
+  expectTypeOf(arr).toEqualTypeOf(arr);
+});
+arr.forEach(function (_item) {
+  // @ts-expect-error the EmberArray version isn't recognized
+  expectTypeOf(this).toEqualTypeOf(target);
+}, target);
+
+expectTypeOf(arr.getEach('bar')).toEqualTypeOf<NatifiedArray<number>>();
+expectTypeOf(arr.getEach('missing')).toEqualTypeOf<NatifiedArray<unknown>>();
```
chriskrycho, line 63:
```suggestion
expectTypeOf(arr.getEach('bar')).toEqualTypeOf<NativeArray<number>>();
expectTypeOf(arr.getEach('missing')).toEqualTypeOf<NativeArray<unknown>>();
```

packages/@ember/array/type-tests/index.ts
```diff
@@ -0,0 +1,214 @@
...
+expectTypeOf(arr.setEach('missing', 'anything')).toEqualTypeOf(arr);
+
+let mapped = arr.map((item, index, arr) => {
+  expectTypeOf(item).toEqualTypeOf<Foo>();
+  expectTypeOf(index).toEqualTypeOf<number>();
+  expectTypeOf(arr).toEqualTypeOf(arr);
+  return 1;
+});
+// NOTE: This should be a NatifiedArray, but the native version takes precedence
+expectTypeOf(mapped).toEqualTypeOf<number[]>();
```
chriskrycho, line 79: I think we might be able to do `toMatchTypeOf`?

```suggestion
expectTypeOf(mapped).toMatchTypeOf<NativeArray<number>>();
```

packages/@ember/array/type-tests/index.ts
```diff
@@ -0,0 +1,214 @@
...
+
+arr.map(function (_item) {
+  // @ts-expect-error the EmberArray version isn't recognized
+  expectTypeOf(this).toEqualTypeOf(target);
+}, target);
+
+expectTypeOf(arr.mapBy('bar')).toEqualTypeOf<NatifiedArray<number>>();
+expectTypeOf(arr.mapBy('missing')).toEqualTypeOf<NatifiedArray<unknown>>();
```
chriskrycho, line 87:
```suggestion
expectTypeOf(arr.mapBy('bar')).toEqualTypeOf<NativeArray<number>>();
expectTypeOf(arr.mapBy('missing')).toEqualTypeOf<NativeArray<unknown>>();
```

packages/@ember/array/type-tests/index.ts
```diff
@@ -0,0 +1,214 @@
...
+
+let filtered = arr.filter((item, index, arr) => {
+  expectTypeOf(item).toEqualTypeOf<Foo>();
+  expectTypeOf(index).toEqualTypeOf<number>();
+  expectTypeOf(arr).toEqualTypeOf(arr);
+  return true;
+});
+// NOTE: This should be a NatifiedArray, but the native version takes precedence
+expectTypeOf(filtered).toEqualTypeOf<Foo[]>();
```
chriskrycho, line 96:
```suggestion
expectTypeOf(filtered).toMatchTypeOf<NativeArray<Foo>>();
```

packages/@ember/array/type-tests/index.ts
```diff
@@ -0,0 +1,214 @@
...
+  expectTypeOf(this).toEqualTypeOf(target);
+}, target);
+
+let rejected = arr.reject((item, index, arr) => {
+  expectTypeOf(item).toEqualTypeOf<Foo>();
+  expectTypeOf(index).toEqualTypeOf<number>();
+  expectTypeOf(arr).toEqualTypeOf(arr);
+  return true;
+});
+expectTypeOf(rejected).toEqualTypeOf<NatifiedArray<Foo>>();
```
chriskrycho, line 108:
```suggestion
expectTypeOf(rejected).toEqualTypeOf<NativeArray<Foo>>();
```

packages/@ember/array/type-tests/index.ts
```diff
@@ -0,0 +1,214 @@
...
+arr.reject(function (_item) {
+  expectTypeOf(this).toEqualTypeOf(target);
+  return true;
+}, target);
+
+expectTypeOf(arr.filterBy('bar')).toEqualTypeOf<NatifiedArray<Foo>>();
+expectTypeOf(arr.filterBy('missing')).toEqualTypeOf<NatifiedArray<Foo>>();
+
+expectTypeOf(arr.rejectBy('bar')).toEqualTypeOf<NatifiedArray<Foo>>();
+expectTypeOf(arr.rejectBy('missing')).toEqualTypeOf<NatifiedArray<Foo>>();
```
chriskrycho, line 118:
```suggestion
expectTypeOf(arr.filterBy('bar')).toEqualTypeOf<NativeArray<Foo>>();
expectTypeOf(arr.filterBy('missing')).toEqualTypeOf<NativeArray<Foo>>();

expectTypeOf(arr.rejectBy('bar')).toEqualTypeOf<NativeArray<Foo>>();
expectTypeOf(arr.rejectBy('missing')).toEqualTypeOf<NativeArray<Foo>>();
```

packages/@ember/array/type-tests/index.ts
```diff
@@ -0,0 +1,214 @@
...
+  expectTypeOf(index).toEqualTypeOf<number>();
+  expectTypeOf(arr).toEqualTypeOf(arr);
+  return 1;
+}, 1);
+expectTypeOf(reduced).toEqualTypeOf<number>();
+// Native behavior takes precedence here
+expectTypeOf(arr.reduce((summation, _item) => summation)).toEqualTypeOf<Foo>();
+
+expectTypeOf(arr.invoke('hi')).toEqualTypeOf<NatifiedArray<string>>();
+expectTypeOf(arr.invoke('withArgs', 1, 'two')).toEqualTypeOf<NatifiedArray<string>>();
```
chriskrycho, line 175:
```suggestion
expectTypeOf(arr.invoke('hi')).toEqualTypeOf<NativeArray<string>>();
expectTypeOf(arr.invoke('withArgs', 1, 'two')).toEqualTypeOf<NativeArray<string>>();
```

packages/@ember/array/type-tests/index.ts
```diff
@@ -0,0 +1,214 @@
...
+// @ts-expect-error Doesn't allow calling with invalid args
+arr.invoke('withArgs', 'invalid');
+expectTypeOf(arr.invoke('missing')).toEqualTypeOf<NatifiedArray<unknown>>();
```
chriskrycho, line 178:
```suggestion
expectTypeOf(arr.invoke('missing')).toEqualTypeOf<NativeArray<unknown>>();
```

packages/@ember/array/type-tests/index.ts
```diff
@@ -0,0 +1,214 @@
...
+// Currently, args passed to unrecognized methods are ignored
+arr.invoke('missing', 'invalid');
+
+expectTypeOf(arr.toArray()).toEqualTypeOf<Foo[]>();
+
+expectTypeOf(arr.compact()).toEqualTypeOf<NatifiedArray<Foo>>();
+expectTypeOf(A([foo, null]).compact()).toEqualTypeOf<NatifiedArray<Foo>>();
```
chriskrycho, line 185:
```suggestion
expectTypeOf(arr.compact()).toEqualTypeOf<NativeArray<Foo>>();
expectTypeOf(A([foo, null]).compact()).toEqualTypeOf<NativeArray<Foo>>();
```

packages/@ember/array/type-tests/index.ts
```diff
@@ -0,0 +1,214 @@
...
+// Doesn't enforce keys
+expectTypeOf(arr.sortBy('missing')).toEqualTypeOf<Foo[]>();
+
+expectTypeOf(arr.uniq()).toEqualTypeOf<NatifiedArray<Foo>>();
+
+expectTypeOf(arr.uniqBy('bar')).toEqualTypeOf<NatifiedArray<Foo>>();
+// Doesn't enforce keys
+expectTypeOf(arr.uniqBy('missing')).toEqualTypeOf<NatifiedArray<Foo>>();
+
+expectTypeOf(arr.without(foo)).toEqualTypeOf<NatifiedArray<Foo>>();
```
chriskrycho, line 202:
```suggestion
expectTypeOf(arr.uniq()).toEqualTypeOf<NativeArray<Foo>>();

expectTypeOf(arr.uniqBy('bar')).toEqualTypeOf<NativeArray<Foo>>();
// Doesn't enforce keys
expectTypeOf(arr.uniqBy('missing')).toEqualTypeOf<NativeArray<Foo>>();

expectTypeOf(arr.without(foo)).toEqualTypeOf<NativeArray<Foo>>();
```

packages/@ember/array/type-tests/index.ts
```diff
@@ -0,0 +1,214 @@
...
+// @ts-expect-error invalid type
+arr.without(1);
+
+expectTypeOf(isArray(arr)).toEqualTypeOf<boolean>();
+
+expectTypeOf(makeArray(arr)).toEqualTypeOf<NatifiedArray<Foo>>();
```
chriskrycho, line 208:
```suggestion
expectTypeOf(makeArray(arr)).toEqualTypeOf<NativeArray<Foo>>();
```

packages/@ember/array/type-tests/index.ts
```diff
@@ -0,0 +1,214 @@
...
+
+expectTypeOf(arr.compact()).toEqualTypeOf<NatifiedArray<Foo>>();
+expectTypeOf(A([foo, null]).compact()).toEqualTypeOf<NatifiedArray<Foo>>();
+
+expectTypeOf(arr.includes(foo)).toEqualTypeOf<boolean>();
+// @ts-expect-error Invalid type
+arr.includes(1);
+
+// For some reason this doesn't return a NatifiedArray
+expectTypeOf(arr.sortBy('bar')).toEqualTypeOf<Foo[]>();
```
chriskrycho, line 192: As with the others, does it work with `toMatchTypeOf`?

```suggestion
expectTypeOf(arr.sortBy('bar')).toMatchTypeOf<NativeArray<Foo>>();
```

chriskrycho (changes requested):

chriskrycho:
Meant to say on the review: thank you for doing this! 💙

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
wagenet: It's probably wrong. The `interface` was supposed to be stuff you have to declare. The way Mixins are handled always confuses me!

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
wagenet: This is correct, but the situation is confusing.

EmberArray is a Mixin for immutable arrays.
MutableArray is a Mixin that extends EmberArray and adds mutability.
NativeArray is the native Array with MutableArray mixed into it.

If there's a better way to handle all of this, let me know.

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
wagenet: Yes, this may not be necessary if we get NativeArray defined correctly.

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
wagenet: It isn't the same right now, but if we get it defined correctly, I agree that it should be.

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
wagenet: Yes, it does return a NativeArray, that said, it seems like the true native Array types might win out in TS anyway.

packages/@ember/-internals/utils/lib/make-array.ts
wagenet: We did need it, but the correct NativeArray types may make it unnecessary.

packages/@ember/array/type-tests/index.ts
wagenet: Possibly. I'll review more.

packages/@ember/array/type-tests/index.ts
wagenet: Maybe. I'll review this as well.

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
wagenet: I figured out why the native types were winning and resolved that.

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
chriskrycho: No, you have the overall right shape of it. I thought the relationship was inverted (because in the types we have on DefinitelyTyped, it is!). Thanks for the clarification. 👍🏼

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
```diff
@@ -0,0 +1,101 @@
...
+  addObjects(objects: T[]): this;
+}
+
+declare const MutableArray: Mixin;
+export { MutableArray };
+
+// NOTE: We have to Omit some definitions from Array because MutableArray defines them differently.
+export interface NativeArray<T>
+  extends Omit<Array<T>, 'every' | 'filter' | 'find' | 'forEach' | 'map' | 'reduce' | 'slice'>,
+    MutableArray<T> {}
```
chriskrycho, line 95: «sobbing in TypeScript»

chriskrycho (approved):

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
chriskrycho: Opened PR as DefinitelyTyped/DefinitelyTyped#58424 in the meantime.

packages/@ember/-internals/utils/lib/make-array.ts
chriskrycho: Ah, I see what the issue was. Glad the fix for that type resolved it. Thank you!

chriskrycho (approved):

PR #19910, opened 2022-01-20
Avoid losing some internal type information

wagenet:


packages/@ember/-internals/runtime/lib/type-of.js
```diff
@@ -77,7 +77,7 @@ const { toString } = Object.prototype;
 
   @method typeOf
   @for @ember/utils
-  @param {Object} item the item to check
+  @param item the item to check
```
NullVoxPopuli, line 80: what's the type if the type annotation is removed? `any`?

packages/@ember/-internals/glimmer/lib/helper.ts
```diff
@@ -183,6 +187,10 @@ class ClassicHelperManager implements HelperManager<ClassicHelperStateBucket> {
   }
 }
 
+function isFactoryManager<T, C>(obj: unknown): obj is FactoryManager<T, C> {
+  return obj != null && 'class' in (obj as FactoryManager<T, C>);
```
NullVoxPopuli, line 191: is a `.class` property the _only_ requirement for calling `create` on `obj`?

NullVoxPopuli (approved):

packages/@ember/-internals/runtime/lib/type-of.js
wagenet: I believe so.

wagenet: It's either `any` or `unknown`.

packages/@ember/-internals/glimmer/lib/helper.ts
wagenet: https://github.com/emberjs/ember.js/blob/master/packages/@ember/-internals/container/lib/container.ts#L250-L253

NullVoxPopuli: for extra certainty, should this function also check that create exists and is a function?

PR #19914, opened 2022-01-22
[Bugfix release] make setOwner check its shape, so that we can be cer…

NullVoxPopuli:
…tain people arent mis-using it

The motivation from this comes from a couple places (in TypeScript):
 - in practical usage, `getOwner` _needs_ casting in order to be used (annoying for consumers of `getOwner`)
 - typed-ember will only document "what is correct" (which makes sense), so in order for `getOwner` to have a useful return type, we need to ensure an owner with _incorrect_ shape cannot be set in the first place.

### DIFF
--- a/packages/@ember/-internals/owner/index.ts
+++ b/packages/@ember/-internals/owner/index.ts
@@ -1,6 +1,6 @@
-import {getOwner as glimmerGetOwner, setOwner as glimmerSetOwner} from '@glimmer/owner';
-import {assert} from '@ember/debug';
-import {DEBUG} from '@glimmer/env';
+import { getOwner as glimmerGetOwner, setOwner as glimmerSetOwner } from '@glimmer/owner';
+import { assert } from '@ember/debug';
+import { DEBUG } from '@glimmer/env';
 
 /**
 @module @ember/application
@@ -20,7 +20,7 @@ export interface Factory<T, C extends FactoryClass | object = FactoryClass> {
   name?: string;
   fullName?: string;
   normalizedName?: string;
-  create(props?: {[prop: string]: any}): T;
+  create(props?: { [prop: string]: any }): T;
 }
 
 export interface EngineInstanceOptions {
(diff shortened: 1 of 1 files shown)


packages/@ember/-internals/owner/index.ts
```diff
@@ -99,5 +101,16 @@ export function getOwner(object: any): Owner {
   @public
 */
 export function setOwner(object: any, owner: Owner): void {
+  if (DEBUG) {
+    let publicApiType = owner as any;
+
+    assert(`Expected owner to be truthy`, publicApiType);
```
NullVoxPopuli, line 107: this doesn't include the entirety of the `Owner`, but what I left out is optional anyway (`?:`)

packages/@ember/-internals/owner/index.ts
```diff
@@ -99,5 +101,27 @@ export function getOwner(object: any): Owner {
...
+    );
+    assert(
+      'Passed owner must have a function, `register`',
+      typeof publicApiType.register === 'function'
+    );
+    assert(
+      'Passed owner must have a function, `hasRegistration`',
+      typeof publicApiType.hasRegistration === 'function'
+    );
+  }
```
NullVoxPopuli, line 124: `buildChildEngineInstance` is omitted because it _seemed_ like private api _even though_ the type declares it as required.
idk if this is the right call, but tests were failing with that check, so that leads me to believe that either tests are incorrectly building an owner, or `buildChildEngineInstance` needs to be marked as optional

NullVoxPopuli: looks like there are tests passing `undefined` or `null` to `setOwner` 🤔

wagenet (approved):
This looks good to me. I assume all the new asserts are for things that are already implicitly required.

ef4:
There are dozens of places where people who aren't using typescript might pass the wrong type into one of Ember's public APIs. Why is this one important enough to spend effort on checking runtime shape?

`getOwner` already returns type `Owner`. What casting are you referring to?

ef4 (changes requested):
I don't think this change is well-motivated. Let's discuss my question above before merging.

NullVoxPopuli:
> getOwner already returns type Owner. What casting are you referring to?

This comes from a discussion with the typed-ember folks: [Discord Link](https://discord.com/channels/480462759797063690/484421406659182603/934460087861395526)

Specifically here: [`getOwner` from DT](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/ember__application/index.d.ts#L122)

ef4:
`@glimmer/owner` doesn't care about the type of the owner, but that's why Ember wraps it in Ember-specific functions that *do* care:

https://github.com/emberjs/ember.js/blob/6b77e9b940a23698201a9d16d1fe420507f048b0/packages/%40ember/-internals/owner/index.ts#L85

https://github.com/emberjs/ember.js/blob/6b77e9b940a23698201a9d16d1fe420507f048b0/packages/%40ember/-internals/owner/index.ts#L101

The typings should reflect this. They should contain a type for Owner and they should use it for setOwner and getOwner.

That doesn't require a change in Ember.

NullVoxPopuli:
cc @chriskrycho :shrug:

chriskrycho:
@ef4 cool, this is good to know. The docs don't cover it *at all*: there's no public notion of what an `Owner` is from the Ember API POV. Typed Ember’s rule has been that we don’t write types for things which aren’t expressly public API somewhere, because we don’t want to get disconnected from Ember as the source of truth. I’m happy to work with folks to get that documented and then we’ll update types accordingly.

To make this less of a pain for users, I will probably open a PR for DefinitelyTyped defining the minimal amount of API which is apparent from the docs and/or internal types, while we work out what should be documented.

chriskrycho:
I opened #19916 to hammer down what we can/should publish as public API.

ef4:
Thanks. 

Regardless of how #19916 goes, this PR is not the path forward. Either setOwner supports *anything*, in which case these assertions are wrong, or setOnly only supports `Owner`, in which case the typings can just say that and these assertions aren't necessary.

PR #19923, opened 2022-01-26
Future public @ember/engine types

wagenet:


packages/@ember/-internals/container/lib/container.ts
```diff
@@ -143,7 +143,7 @@ export default class Container {
    @param {String} [options.source] The fullname of the request source (used for local lookup)
    @return {any}
    */
-  lookup(fullName: string, options: LookupOptions): any {
+  lookup<T>(fullName: string, options?: TypeOptions): T | undefined {
```
chriskrycho, line 146: This should actually return `unknown`, since it's impossible to infer from the arguments what the return type `T` would actually be.

```suggestion
  lookup(fullName: string, options?: TypeOptions): unknown {
```

I know this matches the *other* implementations, but we should take care to move *toward* rather than *away from* safety with this work where our hands are not already tied in some way.

I expect you introduced this because there is client code which is relying on being able to freely cast because of the previous `any` return type?

packages/@ember/-internals/container/lib/registry.ts
```diff
@@ -79,12 +79,12 @@ export default class Registry implements IRegistry {
...
-  readonly _normalizeCache: { [key: string]: string };
-  readonly _options: { [key: string]: TypeOptions };
-  readonly _resolveCache: { [key: string]: object };
-  readonly _typeOptions: { [key: string]: TypeOptions };
+  readonly registrations: { [key: string]: object | undefined };
+  _localLookupCache: { [key: string]: object | undefined };
+  readonly _normalizeCache: { [key: string]: string | undefined };
+  readonly _options: { [key: string]: TypeOptions | undefined };
+  readonly _resolveCache: { [key: string]: object | undefined };
+  readonly _typeOptions: { [key: string]: TypeOptions | undefined };
```
chriskrycho, line 87: 1. Can we just use `Record<string, TypeOptions>` etc. here?
2. It looks like this as well as the `in` check below are both because we don't yet have `noUncheckedIndexAccess` enabled, so we're sort of pushing that safety in manually?

packages/@ember/-internals/container/lib/registry.ts
```diff
@@ -417,7 +417,7 @@ export default class Registry implements IRegistry {
 
   getOptions(fullName: string): TypeOptions | undefined {
     let normalizedName = this.normalize(fullName);
-    let options = this._options[normalizedName];
+    let options = normalizedName in this._options ? this._options[normalizedName] : undefined;
```
chriskrycho, line 420: Hmm. This shouldn't be complaining about the access. 🤔 (See [playground][play].) Since it's a string index, and given the type you wrote, it *should* correctly resolve it in either the `Record` or the index type form. What error was TS giving you here?

[play]: https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgCoE8AOEDymzAD2IAzsgN4BQyymhYcYhA-AFzIlhSgDmA3JQC+lSqEixEKHGAAW0CtWQB3GYwCqmNsgDkICI0LaBwyggA2cEmQBihQgpqZuAN0YoA+oXxFS7AEoQCIRQACYAPJzcIDwANGhYuN7EZAA+yACuICEQMKAQIQB8yAC8yDgARgBWgWAAdAhQ+pAAFCDpZmYAlAKOLm7InrLQ7OTIANoA1hDo7JG8ALrs0nJQyGmZ2bl6IciCJWVVNfWNbq3tXQKKPBBgeATJzTDnAHJwALYQs1y8newY2HcfKkMlkcnkdlQaDQgqQwMgvPdSPtZMASLVPElSGMnh1Xh95j0oY0wOkoCB4ZiSISTDRrrchlBHi93p8ON9or8ygy1iDNuCHESbqTySi0YMVtjmfjqUJKEA

packages/@ember/-internals/glimmer/lib/resolver.ts
```diff
@@ -56,16 +57,12 @@ function instrumentationPayload(name: string) {
   return { object: `component:${name}` };
 }
 
-function componentFor(
-  name: string,
-  owner: Owner,
-  options?: LookupOptions
-): Option<Factory<{}, {}>> {
+function componentFor(name: string, owner: Owner): Option<Factory<{}, {}>> {
```
chriskrycho, line 60: This looks like an unrelated-to-types change – is this a dead code path or something?

packages/@ember/-internals/owner/index.ts
```diff
@@ -27,11 +22,12 @@ export interface EngineInstanceOptions {
 }
 
 import EngineInstance from '@ember/engine/instance';
+import { TypeOptions } from '../container/lib/registry';
 export interface Owner {
-  lookup<T>(fullName: string, options?: LookupOptions): T | undefined;
-  factoryFor<T, C>(fullName: string, options?: LookupOptions): Factory<T, C> | undefined;
-  register<T, C>(fullName: string, factory: Factory<T, C>, options?: LookupOptions): void;
-  hasRegistration(name: string, options?: LookupOptions): boolean;
+  lookup<T>(fullName: string, options?: TypeOptions): T | undefined;
```
chriskrycho, line 27: As before, I'm fine if we leave this *for now*, but we should start working to eliminate it. 😩

packages/@ember/-internals/runtime/lib/mixins/container_proxy.d.ts
```diff
@@ -0,0 +1,11 @@
...
+import Mixin from '../../types/mixin';
+
+interface ContainerProxy {
+  ownerInjection: Container['ownerInjection'];
+  lookup: Container['lookup'];
+  factoryFor: Container['factoryFor'];
+}
+declare const ContainerProxy: Mixin;
+
+export { ContainerProxy as default };
```
chriskrycho, line 11: I believe the normal `export default` form should work; does it not for some reason?

```suggestion
export default ContainerProxy;
```

packages/@ember/-internals/runtime/lib/mixins/registry_proxy.d.ts
```diff
@@ -0,0 +1,17 @@
...
+  hasRegistration: Registry['has'];
+  registeredOption: Registry['getOption'];
+  registerOptions: Registry['options'];
+  registeredOptions: Registry['getOptions'];
+  registerOptionsForType: Registry['optionsForType'];
+  registeredOptionsForType: Registry['getOptionsForType'];
+}
+declare const RegistryProxyMixin: Mixin;
+
+export { RegistryProxyMixin as default };
```
chriskrycho, line 17:
```suggestion
export default RegistryProxyMixin;
```

chriskrycho (changes requested):
Will finish this review tomorrow, but WIP feedback!

packages/@ember/-internals/container/lib/container.ts
wagenet: This makes sense. I’ll give it a second look tomorrow.

packages/@ember/-internals/container/lib/registry.ts
wagenet: We do have noUncheckedIndexAccess now. I’ll revisit this.

packages/@ember/-internals/container/lib/registry.ts
wagenet: Not sure what was happening before but it seems ok now.

packages/@ember/-internals/glimmer/lib/resolver.ts
wagenet: The type was incorrect. `factoryFor` doesn't take options.

packages/@ember/-internals/owner/index.ts
wagenet: I'll fix it now.

packages/@ember/-internals/glimmer/lib/resolver.ts
chriskrycho: Glad we caught it by doing this, then. Thanks! 😃

packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -278,8 +278,8 @@ class Route<T = unknown>
     super(owner);
 
     if (owner) {
-      let router = owner.lookup<EmberRouter>('router:main');
-      let bucketCache = owner.lookup<BucketCache>(P`-bucket-cache:main`);
+      let router = owner.lookup('router:main') as EmberRouter;
```
chriskrycho, line 281: **~:** Same comment throughout this file of course. It's an improvement regardless.

All of these plus a conversation with @dfreeman earlier have me thinking: we *should* be able to provide an internal registry extension *that isn't published* which makes these all work. That's good follow-on which I'm happy to advise about. (We can make that work with the microsyntax using template literal types.)

packages/@ember/-internals/routing/lib/system/route.ts
```diff
@@ -1874,7 +1874,7 @@ class Route<T = unknown>
     let controllerName = this.controllerName || this.routeName;
     let owner = getOwner(this);
     assert('Route is unexpectedly missing an owner', owner);
-    let controller = owner.lookup<Controller>(`controller:${controllerName}`);
+    let controller = owner.lookup(`controller:${controllerName}`) as Controller;
```
chriskrycho, line 1877: **~:** One thing this flags up, just looking at this whole section, is how important dev assertions actually are here. We have, AFAICT, *zero* guarantee from the types here that the `controllerName` is actually a controller. We have *some* guarantees from tests of the runtime, but there's nothing which the compiler can check.

packages/@ember/-internals/routing/lib/system/router.ts
```diff
@@ -312,11 +312,11 @@ class EmberRouter<R extends Route = Route> extends EmberObject.extend(Evented) i
     this._resetQueuedQueryParameterChanges();
     this.namespace = owner.lookup('application:main');
 
-    let bucketCache: BucketCache | undefined = owner.lookup(P`-bucket-cache:main`);
+    let bucketCache = owner.lookup(P`-bucket-cache:main`) as BucketCache | undefined;
```
chriskrycho, line 315: **~:** as throughout.

packages/@ember/-internals/views/lib/system/utils.ts
```diff
@@ -44,7 +44,7 @@ interface View {
   @param {Object} owner
 */
 export function getRootViews(owner: Owner): View[] {
-  let registry = owner.lookup<Dict<View>>('-view-registry:main')!;
+  let registry = owner.lookup('-view-registry:main') as Dict<View>;
```
chriskrycho, line 47: **~:** as throughout, for reference.

packages/@ember/application/type-tests/instance.test.ts
```diff
@@ -19,15 +23,14 @@ expectTypeOf(
   instance.register('service:store-singleton', Store, { singleton: true, instantiate: true })
 ).toEqualTypeOf<void>();
 
-expectTypeOf(instance.lookup<Store>('service:store')).toEqualTypeOf<Store | undefined>();
+expectTypeOf(instance.lookup('service:store')).toEqualTypeOf<unknown>();
```
chriskrycho, line 26: If we add that registry, these will (happily!) flag it up for us as a case where it no longer *equals* that type. :tada:

packages/@ember/debug/lib/capture-render-tree.ts
```diff
@@ -20,7 +20,7 @@ import { expect } from '@glimmer/util';
   @since 3.14.0
 */
 export default function captureRenderTree(app: Owner): CapturedRenderNode[] {
-  let renderer = expect(app.lookup<Renderer>('renderer:-dom'), `BUG: owner is missing renderer`);
+  let renderer = expect(app.lookup('renderer:-dom') as Renderer, `BUG: owner is missing renderer`);
```
chriskrycho, line 23: **~:** could use a better assertion to check that it's not just present but is the right/expected thing.

packages/@ember/engine/lib/engine-parent.ts
```diff
@@ -15,7 +16,7 @@ const ENGINE_PARENT = symbol('ENGINE_PARENT');
   @static
   @private
 */
-export function getEngineParent(engine) {
+export function getEngineParent(engine: EngineInstance): EngineInstance | undefined {
```
chriskrycho, line 19: 💙

chriskrycho (approved):
Guide:

| Mark | Meaning |
| - | - | 
| **?** | informational, non-blocking |
| **~** | not great, non-blocking |
| **X** | blocking |

Spoilers: no **X**, so :shipit: .

packages/@ember/-internals/glimmer/lib/component-managers/curly.ts
```diff
@@ -133,7 +133,7 @@ export default class CurlyComponentManager
 
     if (layout === undefined) {
       if (layoutName !== undefined) {
-        let _factory = owner.lookup<TemplateFactory>(`template:${layoutName}`);
+        let _factory = owner.lookup(`template:${layoutName}`) as TemplateFactory;
```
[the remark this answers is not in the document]
wagenet: I don't know that it's safe, but it's how it was before so it probably is. I figure having `as` is safer (as in we know we're doing something potentially risky) than the way it was before where it wasn't clear that there was risk!

packages/@ember/-internals/glimmer/lib/resolver.ts
```diff
@@ -176,10 +173,9 @@ export default class ResolverImpl implements RuntimeResolver<Owner>, CompileTime
       return helper;
     }
 
-    const factory = owner.factoryFor<
-      SimpleHelper | HelperInstance,
-      HelperFactory<SimpleHelper | HelperInstance>
-    >(`helper:${name}`);
+    const factory = owner.factoryFor(`helper:${name}`) as
+      | Factory<SimpleHelper, HelperFactory<SimpleHelper>>
+      | Factory<HelperInstance, HelperFactory<HelperInstance>>;
```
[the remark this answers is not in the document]
wagenet: Yup, unsure this is safe, but it seems like it must be.

packages/@ember/-internals/glimmer/tests/integration/application/debug-render-tree-test.ts
```diff
@@ -670,7 +670,7 @@ if (ENV._DEBUG_RENDER_TREE) {
         ]);
 
         runTask(() => {
-          let controller = instance!.lookup<Controller>('controller:application')!;
+          let controller = (instance!.lookup('controller:application') as Controller)!;
```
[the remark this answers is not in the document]
wagenet: At least this is a test.

packages/@ember/-internals/routing/lib/system/generate_controller.ts
```diff
@@ -16,7 +16,7 @@ import { DEBUG } from '@glimmer/env';
 */
 
 export function generateControllerFactory(owner: Owner, controllerName: string): Factory<{}> {
-  let Factory = owner.factoryFor<any, any>('controller:basic')!.class;
+  let Factory = (owner.factoryFor('controller:basic') as Factory<any, any>).class;
```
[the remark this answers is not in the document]
wagenet: This needs more love.

PR #19944, opened 2022-02-04
Remove code related to named outlets

### LINKED ISSUE #19942: `{{outlet}}` implementation still contains code for supporting named outlets
For example:
https://github.com/emberjs/ember.js/blob/v4.1.0/packages/%40ember/-internals/glimmer/lib/syntax/outlet.ts#L52-L58

Noticed this when creating #19941.

Related RFC:
https://emberjs.github.io/rfcs/0418-deprecate-route-render-methods.html

bertdeblock:
I _think_ this can go as named outlets are not supported anymore? Not sure if I got everything, feedback appreciated!

Related RFC:
https://emberjs.github.io/rfcs/0418-deprecate-route-render-methods.html

Closes #19942.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/syntax/outlet.ts
+++ b/packages/@ember/-internals/glimmer/lib/syntax/outlet.ts
@@ -6,7 +6,6 @@ import {
   childRefFromParts,
   createComputeRef,
   createDebugAliasRef,
-  createPrimitiveRef,
   Reference,
   valueForRef,
 } from '@glimmer/reference';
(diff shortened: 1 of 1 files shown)


wagenet (approved):
Looks correct to me, but we may still want someone who actually worked on named outlets to confirm.

packages/@ember/-internals/glimmer/lib/syntax/outlet.ts
```diff
@@ -38,27 +38,18 @@ import { OutletState } from '../utils/outlet';
...
-
-    if (args.positional.length === 0) {
-      nameRef = createPrimitiveRef('main');
-    } else {
-      let maybeNameRef = args.positional[0];
-      assert('Expected at least one positional arg', maybeNameRef);
-      nameRef = maybeNameRef;
-    }
 
+    let nameRef: Reference<string> = createPrimitiveRef('main');
```
ef4, line 52: I think this should go too. Below, `outlets[valueForRef(nameRef)]` becomes `outlets.main`

bertdeblock: Removed. Thanks!

ef4 (approved):
This looks good as it is and could be merged, but I also suggested here further work to find and eliminate more named-outlet related code.

locks:
I discussed with @bertdeblock and I'll be merging this and he'll open an issue with your suggestion for future work.
Thanks everyone for sseeing this thru 💪

PR #19945, opened 2022-02-06
Pre-format blueprints to match Prettier rules.

muziejus:
Ember uses Prettier, but it has the blueprints and blueprint fixtures ignored by ESLint.

As such, the templates for the blueprints as well as the fixtures against which the blueprints are tested break Prettier formatting rules, most notably forbidding `function()` in favor of `function ()`.

There are two steps here: 
- forcing Prettier to format the fixtures (which are regular JavaScript and, hence, can formatted with no trouble) 
- adjusting the blueprint files, as well as occasionally the actual blueprint and sometimes the corresponding test to match the fixtures.

When I had to adjust the blueprint proper and/or the corresponding test, it was usually because either was not checking against a files template or a fixture. This takes the form of adjusting `{${EOL}}` and taking out the linebreak.

Somehow, running `ember g route foo` on my computer yields:

```js
import Route from '@ember/routing/route';

export default class FooRoute extends Route {}
```

I suppose Ember/Ember-cli reformats the files as it is making them, because the fixture against which this command [is tested](https://github.com/emberjs/ember.js/blob/master/node-tests/fixtures/route/route.js) in Ember is:

```js
import Route from '@ember/routing/route';

export default class FooRoute extends Route {
}
```

With this PR, the `route` blueprint generates something with `Route {}` and the fixture against which the blueprint is tested is also expecting `Route {}`.

### DIFF
--- a/blueprints/acceptance-test/mocha-files/tests/acceptance/__name__-test.js
+++ b/blueprints/acceptance-test/mocha-files/tests/acceptance/__name__-test.js
@@ -3,18 +3,18 @@ import { expect } from 'chai';
 import startApp from '<%= dasherizedPackageName %>/tests/helpers/start-app';
 <% if (destroyAppExists) { %>import destroyApp from '<%= dasherizedPackageName %>/tests/helpers/destroy-app';<% } else { %>import { run } from '@ember/runloop';<% } %>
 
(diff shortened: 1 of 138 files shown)


locks:
Thank you for making the blueprints tidier :)

PR #19947, opened 2022-02-07
More type tests

wagenet:


packages/@ember/template/type-tests/index.test.ts
```diff
@@ -0,0 +1,13 @@
+import { SafeString } from '@ember/-internals/glimmer';
+import { htmlSafe, isHTMLSafe } from '@ember/template';
+import { expectTypeOf } from 'expect-type';
+
+let safe = htmlSafe('<div>someString</div>');
+expectTypeOf(safe).toEqualTypeOf<SafeString>();
+
+expectTypeOf(isHTMLSafe(safe)).toEqualTypeOf<boolean>();
```
chriskrycho, line 8: Question this raised for me: does `isHTMLSafe` narrow? If so, we should capture that in a test. If *not*, should we do a follow-on to *make* it narrow? I’m trying to think through whether/where that would be useful, but I can’t see that it would *hurt* anything, and I can imagine places where you might require a `SafeString` as an argument or some such. 🤔

packages/@ember/destroyable/type-tests/assert-destroyables-destroyed.test.ts
```diff
@@ -0,0 +1,6 @@
+import { assertDestroyablesDestroyed } from '@ember/destroyable';
+import { expectTypeOf } from 'expect-type';
+
+if (assertDestroyablesDestroyed) {
```
chriskrycho, line 4: Why the conditional? Is it a thing which may not exist at the level of the *import*? 🧐

packages/@ember/destroyable/type-tests/enable-destroy-tracking.test.ts
```diff
@@ -0,0 +1,6 @@
+import { enableDestroyableTracking } from '@ember/destroyable';
+import { expectTypeOf } from 'expect-type';
+
+if (enableDestroyableTracking) {
```
chriskrycho, line 4: Same question as above.

packages/@ember/template/type-tests/index.test.ts
wagenet: Looks like it is https://github.com/emberjs/ember.js/blob/8455848de50242657ae156848f05553a692afce1/packages/%40ember/-internals/glimmer/lib/utils/string.ts#L106

packages/@ember/destroyable/type-tests/enable-destroy-tracking.test.ts
wagenet: Same.

packages/@ember/debug/tests/main_test.js
```diff
@@ -59,7 +59,7 @@ moduleFor(
           id: 'test',
           until: 'forever',
           for: 'me',
-          since: { enabled: '1.0.0' },
+          since: { available: '1.0.0', enabled: '1.0.0' },
```
chriskrycho, line 62: 💙

packages/@ember/debug/type-tests/deprecate.test.ts
```diff
@@ -0,0 +1,63 @@
...
+  since: { available: '1.0.0', enabled: '1.0.0' },
+});
+
+// @ts-expect-error requires since
+deprecate('foo', true, {
+  id: 'foo',
+  until: '5.0.0',
+  for: 'my-module',
+});
+
```
chriskrycho, line 61: Could also use one additional one here:

```ts
// @ts-expect-error requires 'available' if 'enabled' present
deprecate('foo', true, {
  id: 'foo',
  until: '5.0.0',
  since: { enabled: '1.0.0' },
});
```

ember-cli-build.js
```diff
@@ -232,6 +232,8 @@ function templateCompilerBundle(emberPackages, transpileTree) {
           '@ember/-internals/*/tests/**' /* internal packages */,
           '*/*/tests/**' /* scoped packages */,
           '*/tests/**' /* packages */,
+          '*/*/type-tests/**' /* scoped packages */,
+          '*/type-tests/**' /* packages */,
```
chriskrycho, line 236: 💙 thank you for this.

packages/@ember/-internals/glimmer/lib/helpers/action.ts
```diff
@@ -389,34 +389,32 @@ function makeDynamicClosureAction(
...
 
   assert(
     `Action passed is null or undefined in (action) from ${target}.`,
     action !== undefined && action !== null
   );
 
-  let typeofAction = typeof action;
-
-  if (typeofAction === 'string') {
+  if (typeof action === 'string') {
```
chriskrycho, line 410: Interesting. I assume it didn't type-check without this change? From TS 4.5 forward, I would expect it *would*. 🤔

packages/@ember/runloop/index.ts
```diff
@@ -99,8 +128,20 @@ export const _backburner = new Backburner(_queues, {
   @return {Object} return value from invoking the passed function.
   @public
 */
-export function run() {
-  return _backburner.run(...arguments);
+export function run<F extends (...args: any[]) => any>(method: F, ...args: Parameters<F>): void;
```
chriskrycho, line 131: Given the repeated use of `(...args: any[]) => any` I'd suggest extracting a type alias:

```ts
type AnyFn = (...args: any[]) => any;
```

Then you can use that throughout, and it'll be both briefer and clearer (and be a single place to tell ESLint not to yell at us about `any`).

packages/@ember/runloop/index.ts
```diff
@@ -213,7 +268,31 @@ export function join() {
   @since 1.4.0
   @public
 */
-export const bind = (...curried) => {
+export function bind<
```
chriskrycho, line 271: One downside to trying to *actually* type `bind`, and why I ultimately chose not to on DT, is that it makes our `bind` stricter than TS' own. I'm fine with merging this for now, but we should plan to revisit it before publishing types; in general my inclination is to *match* what TS does (sometimes even at the cost of strictness) for clean interoperability with the rest of the ecosystem.

packages/@ember/runloop/index.ts
```diff
@@ -99,8 +128,20 @@ export const _backburner = new Backburner(_queues, {
   @return {Object} return value from invoking the passed function.
   @public
 */
-export function run() {
-  return _backburner.run(...arguments);
+export function run<F extends (...args: any[]) => any>(method: F, ...args: Parameters<F>): void;
+export function run<T, F extends (this: T, ...args: any[]) => any>(
```
chriskrycho, line 132: We should check how this works compared to what I implemented on DT for Ember v4 types. The version on DT has some limitations, so this may be preferable. (Or you may have stolen it from there! I don’t remember exactly what I wrote. :joy:)

packages/@ember/-internals/runtime/lib/compare.ts
```diff
@@ -32,9 +35,10 @@ const TYPE_ORDER = {
 //                                |       `._    `.    \
 //                                `._________`-.   `.   `.___
 //                                              SSt  `------'`
-function spaceship(a, b) {
+function spaceship(a: number, b: number): Compare {
   let diff = a - b;
-  return (diff > 0) - (diff < 0);
+  // SAFETY: It will be one of the known values
+  return (Number(diff > 0) - Number(diff < 0)) as Compare;
```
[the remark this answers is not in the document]
wagenet: Sure, it's weird, lol.

packages/@ember/-internals/runtime/lib/compare.ts
```diff
@@ -135,15 +144,24 @@ export default function compare(v, w) {
       return spaceship(vLen, wLen);
     }
     case 'instance':
-      if (Comparable.detect(v)) {
+      if (isComparable(v) && v.compare) {
```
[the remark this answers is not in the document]
wagenet: Might be a refactoring remnant.

packages/@ember/-internals/glimmer/lib/helpers/action.ts
wagenet: Yeah, it did not work :/

packages/@ember/runloop/index.ts
wagenet: Yeah, makes sense.

packages/@ember/runloop/index.ts
wagenet: I did my own research 😝

packages/@ember/runloop/index.ts
wagenet: We could also add a catch-all type to allow more?

wagenet: I'm also open to loosening it up if people find enough issues in practice.

packages/@ember/-internals/runtime/lib/compare.ts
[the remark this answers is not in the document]
wagenet: For some reason `Comparable` allows for `compare` to be null.

packages/@ember/runloop/index.ts
wagenet: I think we should do this in its own PR.

chriskrycho: Agreed re: doing it in its own PR.

packages/@ember/runloop/index.ts
chriskrycho: Yeah, I go back and forth on `bind` and friends. In general I think we should probably also just get rid of *most* of the stuff in `@ember/runloop`, including this, but that’s a different discussion entirely. 😅

packages/@ember/-internals/glimmer/lib/helpers/action.ts
chriskrycho: Ahhhh, I see why. It isn't discriminating a union or anything quite like that, so the compiler loses track of it. 👍🏼

chriskrycho (approved):

PR #19948, opened 2022-02-07
Future @ember/component types

wagenet:


packages/@ember/-internals/glimmer/lib/component.ts
```diff
@@ -648,402 +649,431 @@ let lazyEventsProcessed = new WeakMap<EventDispatcher, WeakSet<object>>();
...
+    @property layoutName
+    @type String
+    @default undefined
+    @private
+  */
+  layoutName?: string;
+
+  attributeBindings?: Array<string>;
+}
+class Component
```
chriskrycho, line 683: There is a (*small*) runtime risk here, if I remember correctly: there are some hazards around zebra-striping with classic classes. 🦓 CC @wycats @pzuraq @rwjblue @chancancode – I don’t remember what the issue was, but I do remember that we recommended against “zebra-striping” within chains, in the “it should work, 99% of the time, if you dot all your i’s and cross all your t’s but it’s easy to just not zebra-stripe” bucket. Does anyone remember the details, and if so is it an ongoing concern such that we *shouldn’t* do this conversion?

packages/@ember/-internals/glimmer/lib/helper.ts
```diff
@@ -256,7 +265,11 @@ setHelperManager(() => SIMPLE_CLASSIC_HELPER_MANAGER, Wrapper.prototype);
   @public
   @since 1.13.0
 */
-export function helper(helperFn: HelperFunction): HelperFactory<SimpleHelper> {
+export function helper<
+  T = unknown,
+  P extends unknown[] = unknown[],
+  N extends Dict<unknown> = Dict<unknown>
```
chriskrycho, line 271: I *think* we want these not to have defaults—they should "just work" by way of the definition.

```suggestion
  T,
  P extends unknown[],
  N extends Dict<unknown>
```

(This is how the types on DefinitelyTyped work, and I came to believe that the default types for `ember-modifier` were the wrong move as well. That’s why `modifier` itself [does *not* set defaults](https://github.com/ember-modifier/ember-modifier/blob/709949c6ce33f652c96b6a70dcea0b543ee98967/addon/-private/functional/modifier.ts#L26-L32); and [the `FunctionalModifier` return type](https://github.com/ember-modifier/ember-modifier/blob/709949c6ce33f652c96b6a70dcea0b543ee98967/addon/-private/functional/modifier.ts#L5-L9) does only for backwards compatibility.)

packages/@ember/-internals/views/lib/views/core_view.d.ts
```diff
@@ -0,0 +1,12 @@
...
+import EmberObject from '@ember/object';
+import Evented from '@ember/object/evented';
+
+interface CoreView extends Evented, ActionHandler {}
+declare class CoreView extends EmberObject {
+  parentView: CoreView | null;
+  renderer: Renderer;
+}
+
+export { CoreView as default };
```
chriskrycho, line 12: Why not `export default CoreView;`?

packages/@ember/helper/type-tests/invoke-helper.test.ts
```diff
@@ -14,6 +15,9 @@ class PlusOne extends Helper {
 }
 
 export default class PlusOneComponent extends Component {
+  // Glint would help with this
+  declare args: { number: number };
```
chriskrycho, line 19: This is for EmberComponent, right? Maybe just make some nonsense here that isn't `args` to avoid future confusion?

chriskrycho (changes requested):
👍🏼 overall, couple small tweaks and then let's ship it.

packages/@ember/-internals/glimmer/lib/component.ts
chriskrycho: To elaborate: the problem isn't internal, but for external consumers, since [we currently still document](https://api.emberjs.com/ember/4.1/classes/Component) `Component.extend({ ... })` as the preferred way to define a Classic component.

wagenet: The issue is that some properties passed in extend won't override properties on the base class. I'll investigate these cases.

packages/@ember/helper/type-tests/invoke-helper.test.ts
wagenet: Ah yeah, that's right. I always get them mixed up 😬

packages/@ember/-internals/runtime/lib/mixins/action_handler.d.ts
```diff
@@ -1,3 +1,6 @@
 import { Mixin } from '@ember/-internals/metal';
 
-export default class ActionHandler extends Mixin {}
+export default class ActionHandler extends Mixin {
+  actions?: Record<string, (...args: unknown[]) => void>;
```
chriskrycho, line 4: 1. Does it *have* to return `void` (i.e. will anything else be thrown away anyway)? I think the answer is no, it can return anything. The type-checker will more or less shrug at it I think, but it should probably be `unknown`.

2. I don't think the arguments will type check in practice either. 🤔 [playground][p] For types like this, `unknown[]` is usually wrong for arguments, because it means subclasses can never have *narrower* types in the function parameters than than, because it breaks assignability/Liskov substitution/variance rules (whichever way is easiest to think about it for you; they all boil down to the same issue here).

Net, the type should *probably* be:

```suggestion
  actions?: Record<string, (...args: any[]) => unknown>;
```

(Technically, `...args: never[]` is the "right" answer here, because all functions are valid super types of that, since `never` is the "bottom" type… but that confuses the heck out of people. And reasonably so.)

[p]: https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEkoGdnwIJgC4EsD2AdgBJQHAQgzwDeAUPPFNvgcgPwBc8ASuHjMAA8yLDBwEA5gBp4ACgB0i2BORcArgQDWBPAHcCAbQC6ASngBeAHzwAbnhzBLAbloBfWrSSp4AMTx54EAAPLBAyNExcQhIyCio6BiYo1gsaegZ4CTxZAigAWxAuETFJGSgJQvgCNTyAI0ozBIyGOCw1GAJ4AAMAEmpcgtd4HDQ+8pAhgE8QWDQ8CGAul2b3Bnd3IA

packages/@ember/-internals/runtime/lib/system/core_object.d.ts
```diff
@@ -11,18 +11,39 @@ import { Owner } from '@ember/-internals/owner';
  * Implementation is carefully chosen for the reasons described in
  * https://github.com/typed-ember/ember-typings/pull/29
  */
-type EmberClassConstructor<T> = new (owner: Owner) => T;
+export type EmberClassConstructor<T> = new (owner: Owner) => T;
 
-type MergeArray<Arr extends any[]> = Arr extends [infer T, ...infer Rest]
+export type MergeArray<Arr extends any[]> = Arr extends [infer T, ...infer Rest]
```
chriskrycho, line 16: Should these also get `@internal`?

packages/@ember/-internals/views/lib/views/core_view.ts
```diff
@@ -42,7 +47,7 @@ const CoreView = FrameworkObject.extend(Evented, ActionHandler, {
   */
   parentView: null,
 
-  instrumentDetails(hash) {
+  instrumentDetails(hash: any) {
```
chriskrycho, line 50: Fine to do a follow-on pass here, but would `Record<string, unknown>` work as a first pass?

packages/@ember/helper/type-tests/set-helper-manager.test.ts
```diff
@@ -41,7 +42,7 @@ export const SIMPLE_CLASSIC_HELPER_MANAGER = new SimpleClassicHelperManager();
 
 expectTypeOf(
   setHelperManager(() => SIMPLE_CLASSIC_HELPER_MANAGER, Wrapper.prototype)
-).toEqualTypeOf<Wrapper>();
+).toEqualTypeOf<Wrapper<any, any, any>>();
```
chriskrycho, line 45: Can this just specify the types?

```suggestion
).toEqualTypeOf<Wrapper<unknown, unknown[], Dict<unknown>>>();
```

chriskrycho (changes requested):
This should be the last bucket of fixes for this round—thank you for driving this forward.

packages/@ember/-internals/runtime/lib/system/core_object.d.ts
wagenet: Seems like it.

packages/@ember/helper/type-tests/set-helper-manager.test.ts
wagenet: That doesn't work unfortunately. I don't know exactly why.

chriskrycho: Huh. I can pull it and take a look after merging.

chriskrycho (approved):

packages/@ember/helper/type-tests/set-helper-manager.test.ts
chriskrycho: If you're curious: it's because referencing `Wrapper.prototype` ends up causing TS to resolve the relevant type of `Wrapper.prototype` when passed to `setHelperManager` as `Wrapper<any, any, any>`. You can explicitly cast it *there*, but that obviously doesn't help. (I *thought* that microsoft/TypeScript#47607 might be able make it work, but nope: that doesn't support this particular pattern.)

PR #19950, opened 2022-02-09
Update error message instructions

ef4:
Since Ember 3.25 it's unnecessary to pass a helper or modifier class to the corresponding template keyword before you can use them. They are first-class values that can be used directly.

This updates the error message to stop suggesting the now-unnecessary use of the helper and modifier keywords.

Before:

> Assertion Failed: Passing a dynamic string to the (helper) keyword is disallowed. (You specified (helper s) and s evaluated into "my-helper".) This ensures we can statically analyze the template and determine which helpers are used. If the helper name is always the same, use a string literal instead, i.e. (helper "my-helper"). Otherwise, import the helpers into JavaScript **and pass them to the helper keyword**. See https://github.com/emberjs/rfcs/blob/master/text/0496-handlebars-strict-mode.md#4-no-dynamic-resolution for details. ('sample2/templates/application.hbs' @ L3:C9)

After:

> Assertion Failed: Passing a dynamic string to the (helper) keyword is disallowed. (You specified (helper s) and s evaluated into "my-helper".) This ensures we can statically analyze the template and determine which helpers are used. If the helper name is always the same, use a string literal instead, i.e. (helper "my-helper"). Otherwise, import the helpers into JavaScript **and pass them directly**. See https://github.com/emberjs/rfcs/blob/master/text/0496-handlebars-strict-mode.md#4-no-dynamic-resolution for details. ('sample2/templates/application.hbs' @ L3:C9)


lifeart (approved):

locks:
:shipit: Thanks!

PR #19958, opened 2022-02-10
Fix types for canInvoke

wagenet:


packages/@ember/-internals/utils/lib/invoke.ts
```diff
@@ -16,7 +16,7 @@
   @return {Boolean}
   @private
 */
-export function canInvoke(obj: any | null | undefined, methodName: string): obj is Function {
+export function canInvoke(obj: any | null | undefined, methodName: string): boolean {
```
chriskrycho, line 19: Heh, `any | null | undefined` is just `any`. Maybe also `unknown` here?

```ts
export function canInvoke(obj: unknown, methodName: string): boolean {
  return (
    typeof obj === 'object' &&
    obj !== null &&
    (obj as Record<string, unknown>)[methodName] === 'function'
  );
}
```

chriskrycho (changes requested):
lolwut to the original code here :joy:

packages/@ember/-internals/utils/lib/invoke.ts
wagenet: Works for me.

PR #19962, opened 2022-02-10
Convert built-in blueprints to TypeScript

cafreeman:
This PR is the ember.js counterpart to the [work done in ember-cli to add support for typescript blueprints](https://github.com/ember-cli/ember-cli/pull/9768), both of which constitute the implementation of https://github.com/emberjs/rfcs/pull/776.

This PR converts many (though not all) of the built-in ember blueprints. Notably, it excludes the mixin blueprint as well as the older test blueprints for both mocha and qunit. The only test blueprints that were converted were the post-testing overhaul blueprints (i.e. the ones that are now the defaults).

In order to prevent any compatibility issues that could arise from someone upgrading `ember-source` while still on an older version of Ember CLI, this PR also includes a [polyfill for the typescript blueprint functionality](https://github.com/ember-polyfills/ember-cli-typescript-blueprint-polyfill).


package.json
```diff
@@ -64,6 +66,7 @@
     "ember-cli-normalize-entity-name": "^1.0.0",
     "ember-cli-path-utils": "^1.0.0",
     "ember-cli-string-utils": "^1.1.0",
+    "ember-cli-typescript-blueprint-polyfill": "^0.1.0",
```
cafreeman, line 69: Will update this to `^1.0.0` once we know there aren't any other changes needed.

rwjblue:
This looks great!!! I think the only thing left is to feature flag this so that we can easily toggle it on/off without reverting. Unfortunately, we don't have infrastructure setup to be able to read the `@ember/canary-features` module in node-land so it's a tad more annoying/difficult that more standard runtime feature flagging.

What I think we should do is:

* Create a shared utility at `blueprints/-can-emit-typescript.js`, that looks something like:

```js
// TODO: this should be reading from the @ember/canary-features module
// need to refactor broccoli/features.js to be able to work more similarly
// to https://github.com/emberjs/data/pull/6231
const EMBER_TYPESCRIPT_BLUEPRINTS = false;

module.exports = function() {
  return 'EMBER_TYPESCRIPT_BLUEPRINTS' in process.env 
    ? process.env.EMBER_TYPESCRIPT_BLUEPRINTS
    : EMBER_TYPESCRIPT_BLUEPRINTS;
};
```

* Then update the polyfill to accept a second parameter (either options or a boolean) that allows us to **force** downlevel everyone to `.js`
* Then update the blueprints here to test with that environment variable on and off (I think the output is actually nearly identical, with the only difference being the extension?)

rwjblue:
Chatted with @cafreeman offline, and I think he's come up with a better (and simpler) way to do the feature flagging.

rwjblue:
Eeeck. Looks like we have some conflicts now, my apologies RE: the delays landing this.

@cafreeman - Mind rebasing / fixing things up?

Issue #19916, opened 2022-01-24
[Documentation] No documented public API for `Owner`


chriskrycho:
As briefly discussed in #19914, there is not (and to my knowledge has never been) a definition of the public API of the `Owner` as expected by `getOwner` and `setOwner`. The API docs specify only `Object` as the type for the owner object returned by `getOwner` or expected as the second param to `setOwner`:

- [`getOwner`](https://api.emberjs.com/ember/4.1/functions/@ember%2Fapplication/getOwner)
- [`setOwner`](https://api.emberjs.com/ember/4.1/functions/@ember%2Fapplication/setOwner)

Additionally, while the `getOwner` docs include a short snippet showing the use of `lookup`, there is no further documentation anywhere in Ember's API docs (as far as I can tell) of what exactly an `Owner` is.

On #19914, @ef4 suggested that the types within Ember represent the intended public API:

https://github.com/emberjs/ember.js/blob/6b77e9b940a23698201a9d16d1fe420507f048b0/packages/%40ember/-internals/owner/index.ts#L30-L39

If that is true, we should:

1. Document that as a public type, so that we can 
2. Have something authoritative to point to when we open a PR to DefinitelyTyped addressing this.

---

Some context:

1. Past (verbal) discussions with various folks have never been conclusive here, so I’m opening this to get us as quickly as possible to an agreed-upon answer that we can document and publish—this is causing folks some pain as they upgrade to the v4 types. Earlier versions of the types returned `any` here, because they predated better tools like `unknown`, and in the absence of a publicly-documented API, `unknown` (or perhaps the equally useless to end users `object`) are the only options on offer.

2. Typed Ember has always maintained a policy that we do not publish types on DefinitelyTyped which do not correspond to documented public API for Ember. We are absolutely not in the business of *defining* the public API for Ember! Rather, we are in the business of *accurately representing it in the types*.

ef4:
IMO it hardly matters whether people originally intended it to be public. It's definitely relied on too heavily for us to ever break it without a lengthy migration plan.

There are 53 places in Ember itself that will likely blow up if anybody tries to `setOwner` something that's not an `Owner`, which you can see if you change `getOwner` to return `unknown`:

```
yarn run v1.22.17
$ tsc --noEmit
packages/@ember/-internals/glimmer/lib/component-managers/curly.ts(134,24): error TS2571: Object is of type 'unknown'.
[...]

PR #19967, opened 2022-02-11
Refactor owner

wagenet:


packages/@ember/-internals/owner/index.ts
```diff
@@ -16,25 +19,26 @@ export interface Factory<T, C extends FactoryClass | object = FactoryClass> {
...
 }
 
-export interface EngineInstanceOptions {
-  mountPoint: string;
-  routable: boolean;
-}
-
-import EngineInstance from '@ember/engine/instance';
-import { TypeOptions } from '../container/lib/registry';
+// A combination of the public methods on ContainerProxyMixin and RegistryProxyMixin
```
chriskrycho, line 22: As a next step (doesn't need to be in this PR, but is related to #19916), we should add docs for all of these bits of public API.

That in turn will let us expose a copy of these types in DefinitelyTyped, which will unblock folks between now and when we publish these.

chriskrycho (approved):
:shipit: 💙

PR #19969, opened 2022-02-11
Improve BootOptions types and add more type-tests

wagenet:


packages/@ember/application/instance.d.ts
```diff
@@ -1,22 +1,31 @@
...
+  isBrowser?: boolean;
+  shouldRender?: boolean;
+  document?: Document | null;
+  rootElement?: string | Element | null;
+  location?: string | null;
+  // Private?
+  isInteractive?: boolean;
+}
+
+export class CBootOptions {
```
chriskrycho, line 13: How (if at all) is this used?

chriskrycho (approved):

PR #19971, opened 2022-02-14
[BUGFIX release] Don't serialize default QPs on RouterService

### LINKED ISSUE #19493: [Bug] Router service transitionTo adds unspecified query params (with their default values)
### 🐞 Describe the Bug
When using the `transitionTo` and `replaceWith` methods on the `RouterService`, query params that are specified on the controller for the new route, and which have default values with are non-null, are added to the resulting URL, even though they are not specified in the `transitionTo` or `replaceWith` call.

### 🔬 Minimal Reproduction
A small reproduction (with a failing test) is here: https://ember-twiddle.com/a796bad00bded99fa45bfbaab159b485?openFiles=tests.acceptance.transition-to-query-params-test%5C.js

### 😕 Actual Behavior
Using `RouterService#transitionTo` adds query parameters that have non-null default values on the controller but were not specified in the `transitionTo` call.

### 🤔 Expected Behavior
Using `RouterService#transitionTo` should not add unspecified query parameters to the URL (it should match the (now-deprecated) `Controller#transitionToRoute` and `Route#transitionTo` methods.)

### 🌍 Environment
- Ember: - from at least 3.11.1 (earliest version the ember-twiddle example runs with) through at least 3.26.0

### ➕ Additional Context
I know there have been a few issues about `RouterService#transitionTo` not pruning query parameters that have default values (e.g., #19492), but this is different in that it is **added** query parameters, _with_ their default values. That makes sense given the [Router Service RFC](https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md)'s note that "default values will not be stripped from generated URLs."

I think the behavior described here is a bug it's *adding* query params that weren't specified, and it's behavior which contravenes the RFC's rationale for not pruning default-valued query params.
The RFC notes that
> Determining their default values is expensive, because it involves instantiating the corresponding controller, even in cases where we will never visit its route.

However, in this case, in order to 1) discover the query params, and 2) discover their default values in order to add them to the URL, the controller had to be instantiated anyway.

wagenet:
Fixes #19493.

### DIFF
--- a/packages/@ember/-internals/routing/lib/services/router.ts
+++ b/packages/@ember/-internals/routing/lib/services/router.ts
@@ -133,7 +133,6 @@ class RouterService<R extends Route> extends Service.extend(Evented) {
     let { routeName, models, queryParams } = extractRouteArgs(args);
 
     let transition = this._router._doTransition(routeName, models, queryParams, true);
-    transition['_keepDefaultQueryParamValues'] = true;
 
     return transition;
   }
@@ -324,21 +323,31 @@ class RouterService<R extends Route> extends Service.extend(Evented) {
     let hasQueryParams = Object.keys(queryParams).length > 0;
 
     if (hasQueryParams) {
+      // UNSAFE: casting `routeName as string` here encodes the existing
+      // assumption but may be wrong: `extractRouteArgs` correctly returns it
+      // as `string | undefined`. There may be bugs if `_prepareQueryParams`
+      // does not correctly account for `undefined` values for `routeName`.
+      //  Spoilers: under the hood this currently uses router.js APIs which
+      // *do not* account for this being `undefined`.
+      let targetRouteName = routeName as string;
+
       queryParams = Object.assign({}, queryParams);
       this._router._prepareQueryParams(
-        // UNSAFE: casting `routeName as string` here encodes the existing
-        // assumption but may be wrong: `extractRouteArgs` correctly returns it
-        // as `string | undefined`. There may be bugs if `_prepareQueryParams`
-        // does not correctly account for `undefined` values for `routeName`.
-        //  Spoilers: under the hood this currently uses router.js APIs which
-        // *do not* account for this being `undefined`.
-        routeName as string,
+        targetRouteName,
         models,
         queryParams,
         true /* fromRouterService */
       );
 
-      return shallowEqual(queryParams, routerMicrolib.state!.queryParams);
+      let currentQueryParams = Object.assign({}, routerMicrolib.state!.queryParams);
+      this._router._prepareQueryParams(
+        targetRouteName,
+        models,
+        currentQueryParams,
+        true /* fromRouterService */
+      );
+
+      return shallowEqual(queryParams, currentQueryParams);
     }
 
     return true;
--- a/packages/ember/tests/routing/router_service_test/replaceWith_test.js
+++ b/packages/ember/tests/routing/router_service_test/replaceWith_test.js
@@ -107,7 +107,7 @@ moduleFor(
         });
     }
 
-    ['@test RouterService#replaceWith with basic query params does not remove query param defaults'](
+    ['@test RouterService#replaceWith with basic query params removes query param defaults'](
       assert
     ) {
       assert.expect(1);
@@ -133,7 +133,7 @@ moduleFor(
           return this.routerService.replaceWith('parent.child', queryParams);
         })
         .then(() => {
-          assert.deepEqual(this.state, ['/', '/child?sort=ASC']);
+          assert.deepEqual(this.state, ['/', '/child']);
         });
     }
   }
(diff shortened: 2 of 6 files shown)


packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -324,21 +323,31 @@ class RouterService<R extends Route> extends Service.extend(Evented) {
...
       queryParams = Object.assign({}, queryParams);
       this._router._prepareQueryParams(
-        // UNSAFE: casting `routeName as string` here encodes the existing
-        // assumption but may be wrong: `extractRouteArgs` correctly returns it
-        // as `string | undefined`. There may be bugs if `_prepareQueryParams`
-        // does not correctly account for `undefined` values for `routeName`.
-        //  Spoilers: under the hood this currently uses router.js APIs which
-        // *do not* account for this being `undefined`.
-        routeName as string,
+        targetRouteName,
```
NullVoxPopuli, line 336: what happens when targetRouteName here is undefined?

should we remove the `as` and instead narrow the type with an assert?

wagenet: I don’t know. It was like this before and I just moved things around. I do want to review all `as` casts someone soon.

locks (approved):

snewcomer:
Thanks for doing this!!

rwjblue:
I think backporting this to beta is probably fine, but I'd hold off on backporting to `release`. This is very likely to accidentally break things for folks not even realizing they were relying on the prior behavior. I definitely think the change is correct, it's just more risky than a patch release IMO.

boris-petrov:
I updated Ember from 4.2.0 to 4.3.0 and a test started failing. I have a component that does:

```ts
this.router.replaceWith(...this.model.routeComponents, { queryParams: this.model.queryParameters })
```

Where `this.model.routeComponents` is `['some-route', 'some parameter']` and `this.model.queryParameters` is `undefined`. This used to work fine before. After the update:

```
More context objects were passed than there are dynamic segments for the route: some-route
```

It's because of the "empty" `queryParams` parameter. If I remove it, it works. Is that what this PR fixes or is this a regression?

boris-petrov:
@wagenet, @rwjblue, @snewcomer, @locks, @NullVoxPopuli - sorry for pinging you all but any ideas about my comment above?

NullVoxPopuli:
@boris-petrov 
sorry for taking so long to reply, can you provide the definitions of `routeComponents` and `querpParameters`?

NullVoxPopuli:
> which definitely is a bug I believe.

I'd agree with that -- `{}` should behave the same as not passing it.

want to submit a PR?

boris-petrov:
> sorry for taking so long to reply, can you provide the definitions of `routeComponents` and `querpParameters`?

No worries. Well, `routeComponents` is just a list of strings and `queryParameters` is a hash (but in this case `undefined`).

> want to submit a PR?

It's probably going to be difficult for me, I'm completely unfamiliar with the code. But I did add a [PR](https://github.com/emberjs/ember.js/pull/20042) with tests.

wagenet:
@boris-petrov Thanks for the PR! I've added my comments there.

packages/ember/tests/routing/router_service_test/replaceWith_test.js
```diff
@@ -107,7 +107,7 @@ moduleFor(
         });
     }
 
-    ['@test RouterService#replaceWith with basic query params does not remove query param defaults'](
+    ['@test RouterService#replaceWith with basic query params removes query param defaults'](
```
chriskrycho, line 110: I believe this is incorrect, and while it fixed the issue identified in #19493, this assertion was correct before, per [the RFC](https://rfcs.emberjs.com/id/0095-router-service/#query-parameter-semantics):

> - default values will not be stripped from generated URLs. For example, `urlFor('my-route', { sortBy: 'title' })` will always include `?sortBy=title`, whether or not `title` is the default value of `sortBy`.
> 
> - to explicitly unset a query parameter, you can pass the symbol `Ember.DEFAULT_VALUE` as its value. For example, `transitionTo('my-route', { sortBy: Ember.DEFAULT_VALUE })` will result in a URL that does not contain any `?sortBy=`.

For #19493, if we want to address the issue raised there, I think we need a *separate* test here which captures the case where the user invokes the service method(s) *without* default params, to confirm that we don’t needlessly instantiate and *add* the defaults. @pgengler correctly notes on that issue that the behavior seen there seems contrary to the rest of the intent of the RFC around default query params *but* (a) the relevant specific behavior is not specified by the RFC, (b) the *actual* behavior has been present for a long time, and (c) given that the `transitionTo()` in question is *to the target route*, I think we would end up instantiating the controller *anyway*?

We should have a quick chat about this with Framework folks to nail down what the path forward is. I suspect we need a revert here as a starting point, which I expect will fix #20051, and then to make a decision about the desired path for #19493.

chriskrycho: Update on the above: see comment [here](https://github.com/emberjs/ember.js/issues/20051#issuecomment-1275061241); this looks unrelated.

PR #19973, opened 2022-02-15
[FEATURE default helper manager] glimmer vm upgrade + default helper manager / plain functions as helpers

NullVoxPopuli:


Windvis:
I created a PR (https://github.com/glimmerjs/glimmer-vm/pull/1384) to add the needed feature flag on the Glimmer side: https://github.com/glimmerjs/glimmer-vm/pull/1384

When I linked it into my local ember-source repo I did notice some tests were failing on the Ember side, so it probably requires more changes than simply bumping the version and enabling the feature. I might be wrong though, it was getting late :sweat_smile:.

chancancode:
I released the flag @Windvis added as 0.84.1 so we can proceed with this (w.r.t. the remaining steps in https://github.com/glimmerjs/glimmer-vm/issues/1383). Note that in Ember, we should default the feature to be off unless specifically turned on by the ember feature flag (the ember feature flag's value should in turns be used to set the glimmer flag @Windvis added)

chancancode:
Specifically:

1. Add flag to https://github.com/emberjs/ember.js/blob/master/packages/%40ember/canary-features/index.ts#L14-L18
2. In https://github.com/emberjs/ember.js/blob/f89dd90290e993ce8bc9ea5f7f103a82687a2be9/packages/%40ember/-internals/glimmer/lib/environment.ts#L21-L88, set the value based on the flag
3. Add/update the relevant tests taking into account the value of the flag

ef4:
This was superseded by #20052.

PR #19977, opened 2022-02-16
Add a feature flag for the unique-id helper

Windvis:
This adds the required feature flag to get #19882 rolling again.

I've seen other feature flags use template template transforms to throw warnings but I didn't do that here since that would show the error when people use the polyfill: https://github.com/ctjhoa/ember-unique-id-helper-polyfill. So I kept it simple. Let me know if I should use a different strategy.


locks:
Thanks for the help! I'll merge as is and we can track it in the original PR.

PR #19980, opened 2022-02-18
[DOC] Update `@since` tag for `unique-id` helper

bertdeblock:


MelSumner (approved):
We were just too excited to ship it. Thanks for the PR! (ETA: it should be 4.4.0 but made that edit)

packages/@ember/-internals/glimmer/lib/helpers/unique-id.ts
```diff
@@ -21,7 +21,7 @@
 
   @method unique-id
   @for Ember.Templates.helpers
-  @since 4.1.0
+  @since 4.3.0
```
MelSumner, line 24:
```suggestion
  @since 4.4.0
```

PR #19983, opened 2022-02-23
Internal test helpers TS

wagenet:
This work also exposed some issues in the existing types.

Review notes: The first commit just renames the files. The second commit actually makes the changes. You'll be able to see the changes easier if you look at just the second commit.


packages/internal-test-helpers/lib/node-query.ts
```diff
@@ -26,23 +37,29 @@ export default class NodeQuery {
...
+  find(selector: string | Element) {
     assertSingle(this);
 
     return this[0].querySelector(selector);
   }
 
-  findAll(selector) {
-    let nodes = [];
+  findAll(selector: string) {
+    let nodes: HTMLElement[] = [];
```
chriskrycho, line 47: I think this should be `Element` instead? And also should have a return type.

```suggestion
  findAll(selector: string): NodeQuery {
    let nodes: Element[] = [];
```

packages/internal-test-helpers/lib/node-query.ts
```diff
@@ -26,23 +37,29 @@ export default class NodeQuery {
...
 
     this.nodes.forEach((node) => {
-      nodes.push(...node.querySelectorAll(selector));
+      nodes.push(...(node.querySelectorAll(selector) as NodeListOf<HTMLElement>));
```
chriskrycho, line 50: If so, then this can also drop the cast.

```suggestion
      nodes.push(...node.querySelectorAll(selector));
```

packages/internal-test-helpers/lib/run.ts
```diff
@@ -13,7 +13,7 @@ export function runDestroy(toDestroy: any): void {
   }
 }
 
-export function runTask(callback: (...args: any[]) => any): void {
+export function runTask<F extends () => any>(callback: F): ReturnType<F> {
```
chriskrycho, line 16: Was dropping the params intentional? If not:

```suggestion
export function runTask<F extends (...args: any[]) => any>(callback: F): ReturnType<F> {
```

packages/internal-test-helpers/lib/test-cases/abstract-application.ts
```diff
@@ -1,10 +1,19 @@
...
+import Application from '@ember/application';
+import ApplicationInstance, { BootOptions } from '@ember/application/instance';
+import Router from '@ember/routing/router';
 
-export default class AbstractApplicationTestCase extends AbstractTestCase {
-  _ensureInstance(bootOptions) {
+export default abstract class AbstractApplicationTestCase extends AbstractTestCase {
+  _applicationInstancePromise?: Promise<ApplicationInstance>;
+
+  declare application: Application;
```
chriskrycho, line 12: Can you add a note explaining where this comes from? `declare` in a context like this makes me twitchy. 🤪

packages/internal-test-helpers/lib/test-cases/abstract-application.ts
```diff
@@ -30,17 +39,24 @@ export default class AbstractApplicationTestCase extends AbstractTestCase {
...
+
+    let element;
+    if (ENV._APPLICATION_TEMPLATE_WRAPPER) {
+      element = document.querySelector('#qunit-fixture > div.ember-view');
     } else {
-      return (this._element = document.querySelector('#qunit-fixture'));
+      element = document.querySelector('#qunit-fixture');
     }
+
+    return (this._element = element);
```
chriskrycho, line 56: Absolutely not a blocker, since this code was already here, but I find this *way* clearer:

```ts
if (!this._element) {
  this._element = ENV._APPLICATION_TEMPLATE_WRAPPER
    ? document.querySelector('#qunit-fixture > div.ember-view')
    : document.querySelector('#qunit-fixture');
}

return this._element;
```

packages/internal-test-helpers/lib/test-cases/application.ts
```diff
@@ -3,15 +3,26 @@ import Application from '@ember/application';
...
     this.application = runTask(this.createApplication.bind(this, applicationOptions));
 
-    this.resolver = this.application.__registry__.resolver;
+    // TODO: Review this cast
+    let resolver = this.application.__registry__.resolver;
+    emberAssert('expected a resolver', resolver instanceof Resolver);
+    this.resolver = resolver;
 
     if (this.resolver) {
       this.resolver.add('router:main', Router.extend(this.routerOptions));
```
chriskrycho, line 28: Given the changes above, I think this is now a meaningless check, since it wouldn't pass the `emberAssert` if there were no resolver?

chriskrycho (changes requested):
Overall 👍🏼, just a couple small tweaks!

packages/@ember/-internals/glimmer/lib/helpers/action.ts
```diff
@@ -310,7 +310,7 @@ export default internalHelper(
     } else {
       fn = makeDynamicClosureAction(
         valueForRef(context) as object,
-        target,
+        target!,
```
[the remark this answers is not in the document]
wagenet: Yeah, there are already a lot too :/

packages/internal-test-helpers/lib/equal-inner-html.ts
```diff
@@ -6,11 +6,11 @@ let ieSVGInnerHTML = (() => {
   let div = document.createElement('div');
   let node = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
   div.appendChild(node);
-  let clone = div.cloneNode(true);
+  let clone = div.cloneNode(true) as HTMLDivElement;
```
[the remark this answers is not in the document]
wagenet: Yeah, it fails to determine the type of element.

packages/internal-test-helpers/lib/run.ts
wagenet: It isn't ever called with args so didn't seem terrible to simplify.

packages/internal-test-helpers/lib/test-cases/abstract-application.ts
wagenet: Probably should be abstract.

chriskrycho (approved):

packages/internal-test-helpers/lib/run.ts
chriskrycho: Seems good!

packages/internal-test-helpers/lib/equal-inner-html.ts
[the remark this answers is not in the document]
chriskrycho: Yeah, it should be `: this` as the return type on `clone`. I might file an issue on the DOM lib.

PR #19986, opened 2022-02-24
Convert Object internals to TS

wagenet:
There is some risk here since native classes handle mixins and property overwriting differently than `create/extend` does. To mitigate the risks, I've taken special care to carefully replace uses of `_super` to call the mixin where appropriate. I also made sure not to define class values for any properties that are overridable via `create/extend`. Potentially people are overwriting some properties that we no longer allow, but those properties were not marked as part of the public API.

I recommend reviewing with whitespace turned off.


packages/@ember/controller/lib/controller_mixin.d.ts
```diff
@@ -14,8 +14,6 @@ interface ControllerMixin<T> {
 
   // From routing/lib/ext/controller
 
-  concatenatedProperties: string[];
```
wagenet, line 17: Again, this is from `EmberObject`.

packages/@ember/-internals/runtime/lib/system/core_object.ts
```diff
@@ -3,7 +3,7 @@
 */
 
 import { getFactoryFor, setFactoryFor } from '@ember/-internals/container';
-import { getOwner } from '@ember/-internals/owner';
+import { getOwner, Owner } from '@ember/-internals/owner';
```
chriskrycho, line 6:
```suggestion
import { getOwner, type Owner } from '@ember/-internals/owner';
```

packages/@ember/-internals/runtime/lib/system/core_object.ts
```diff
@@ -131,7 +167,7 @@ function initialize(obj, properties) {
     }
   }
 
-  sendEvent(obj, 'init', undefined, undefined, undefined, m);
+  sendEvent(obj, 'init', undefined, undefined, m);
```
chriskrycho, line 170: Was this just a straight-up bug happening with this `sendEvent()` before? 😬

packages/@ember/-internals/runtime/lib/system/core_object.ts
```diff
@@ -666,9 +707,15 @@ class CoreObject {
     @param {Object} [arguments]* Object containing values to use within the new class
     @public
   */
-  static extend() {
+  // Args should be `Array<Mixin | Record<string, unknown>>` but this causes
+  // types to be overly strict.
+  static extend<Statics, Instance>(
```
chriskrycho, line 712: Just leaving a note for us that we're going to want to think about how to handle these when we start publishing types: I think (?!?) and wrote in the RFC that we *don't* want to publish these, because they mostly give people just enough rope to hang themselves with in classic classes.

packages/@ember/-internals/runtime/lib/system/core_object.ts
```diff
@@ -713,7 +760,12 @@ class CoreObject {
     @param [arguments]*
     @public
   */
-  static create(props, extra) {
+  static create<Class extends typeof CoreObject, Args extends Array<Record<string, any>>>(
```
chriskrycho, line 763: And the same here.

packages/@ember/-internals/runtime/lib/system/core_object.ts
```diff
@@ -970,22 +1033,21 @@ class CoreObject {
...
-CoreObject.isClass = true;
-CoreObject.isMethod = false;
+  static isClass = true;
+  static isMethod = false;
 
-function flattenProps(...props) {
-  let { concatenatedProperties, mergedProperties } = this;
-  let hasConcatenatedProps =
-    concatenatedProperties !== undefined && concatenatedProperties.length > 0;
-  let hasMergedProps = mergedProperties !== undefined && mergedProperties.length > 0;
```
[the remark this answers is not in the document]
chriskrycho: Huh. Nice bit of cleanup.

packages/@ember/-internals/runtime/lib/system/object.ts
```diff
@@ -20,33 +20,27 @@ import { DEBUG } from '@glimmer/env';
...
+export default EmberObject;
 
-export let FrameworkObject;
-
-FrameworkObject = class FrameworkObject extends CoreObject {
-  get _debugContainerKey() {
-    let factory = getFactoryFor(this);
-    return factory !== undefined && factory.fullName;
-  }
-};
```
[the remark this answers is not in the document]
chriskrycho: Seems like it should be. Maybe do it in a separate PR just to be safe, though? This stuff around factories is... gnarly.

packages/@ember/-internals/views/lib/views/core_view.ts
```diff
@@ -20,21 +19,17 @@ import states from './states';
...
+interface CoreView extends FrameworkObject, Evented, ActionHandler, View {}
+class CoreView extends FrameworkObject.extend(Evented, ActionHandler) {
+  isView = true;
 
-  init() {
-    this._super(...arguments);
-    this._state = 'preRender';
-    this._currentState = this._states.preRender;
-  },
+  _states = states;
```
chriskrycho, line 27: One important note: this has meaningfully different semantics than what it replaced: it's now one reference per instance to the states object, rather than a single reference on the prototype. While in this case it should be a *small* impact, it's not zero, especially for `CoreView`. We should see if we can come up with a better way of doing this to avoid that. :thinking:

packages/@ember/-internals/views/lib/views/core_view.ts
```diff
@@ -45,14 +40,32 @@ const CoreView = (FrameworkObject.extend(Evented, ActionHandler, {
     @default null
     @private
   */
-  parentView: null,
+  declare parentView: View | null;
+
+  init(properties: object | undefined) {
```
chriskrycho, line 45: This whole pattern seems like it's in the "this looks like it should work but it makes me *super* nervous" bucket. Mind at least leaving a detailed comment in the places you've done it explaining the method so that when we come along to work on it (hopefully including just removing a lot of it!) 🔜 it'll be easier to follow?

chriskrycho (changes requested):

packages/@ember/-internals/glimmer/lib/component.ts
```diff
@@ -649,415 +648,445 @@ let lazyEventsProcessed = new WeakMap<EventDispatcher, WeakSet<object>>();
...
-          // In FastBoot we have no EventDispatcher. Set to null to not try again to look it up.
-          this.__dispatcher = null;
+    }
+
+    if (DEBUG && eventDispatcher && this.renderer._isInteractive && this.tagName === '') {
+      let eventNames = [];
+      let events = eventDispatcher.finalEventNameMapping;
+
+      for (let key in events) {
+        let methodName = events[key]!;
```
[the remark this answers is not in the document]
wagenet: I’ll give it a try.

packages/@ember/-internals/glimmer/lib/component.ts
```diff
@@ -649,415 +648,445 @@ let lazyEventsProcessed = new WeakMap<EventDispatcher, WeakSet<object>>();
...
-            }
-          });
+  init(properties: object | undefined) {
+    super.init(properties);
 
-          lazyEventsProcessedForComponentClass.add(proto);
-        }
-      }
+    // Handle methods from ViewMixin
+    this._superRerender = this.rerender;
```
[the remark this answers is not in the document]
wagenet: I’ll give it a try

packages/@ember/-internals/meta/lib/meta.ts
```diff
@@ -354,7 +355,7 @@ export class Meta {
   addToListeners(
     eventName: string,
     target: object | null,
-    method: Function | string,
+    method: Function | string | Symbol,
```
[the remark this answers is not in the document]
wagenet: This method is actually called with a Symbol in at least one place.

packages/@ember/-internals/runtime/lib/system/core_object.ts
wagenet: I believe so

packages/@ember/-internals/runtime/lib/system/core_object.ts
wagenet: Yeah, I think that’s probably correct

packages/@ember/-internals/views/lib/views/core_view.ts
wagenet: Good catch. I’ll review

packages/@ember/-internals/views/lib/views/core_view.ts
```diff
@@ -21,10 +21,15 @@ import states from './states';
 */
 
 interface CoreView extends FrameworkObject, Evented, ActionHandler, View {}
-class CoreView extends FrameworkObject.extend(Evented, ActionHandler) {
+class CoreView extends FrameworkObject.extend(Evented, ActionHandler, {
+  // Continue to declare `_states` here so that we have a single reference on the prototype
+  // instead of one on each instance.
+  _states: states,
+}) {
```
chriskrycho, line 28: :rofl-sob:

chriskrycho (approved):
This looks good. Should be good to merge once we rebase on the factory handling extraction.

PR #19987, opened 2022-02-25
Convert Application and Engine to TS

wagenet:
Will require careful review since it converts `extend` based classes to native classes.


packages/@ember/-internals/container/lib/container.ts
```diff
@@ -410,8 +409,8 @@ function destroyDestroyables(container: Container): void {
     let value = cache[key];
     assert('has cached value', value);
 
-    if (value.destroy) {
-      value.destroy();
+    if ((value as any).destroy) {
```
wagenet, line 412: Type guard?

packages/@ember/-internals/container/lib/container.ts
```diff
@@ -373,13 +372,13 @@ function instantiateFactory(
...
   if (isSingletonInstance(container, fullName, options)) {
-    let instance = (container.cache[normalizedName] = factoryManager.create() as CacheMember);
+    let instance = (container.cache[normalizedName] = factoryManager.create());
 
     // if this lookup happened _during_ destruction (emits a deprecation, but
     // is still possible) ensure that it gets destroyed
     if (container.isDestroying) {
-      if (typeof instance.destroy === 'function') {
-        instance.destroy();
+      if (typeof (instance as any).destroy === 'function') {
```
wagenet, line 380: Could make a type guard function for this.

packages/@ember/-internals/container/lib/container.ts
```diff
@@ -429,8 +428,8 @@ function resetMember(container: Container, fullName: string) {
   if (member) {
     delete container.cache[fullName];
 
-    if (member.destroy) {
-      member.destroy();
+    if ((member as any).destroy) {
```
wagenet, line 431: Type guard?

packages/@ember/-internals/container/lib/registry.ts
```diff
@@ -31,9 +28,9 @@ export interface IRegistry {
   getOptions(fullName: string): TypeOptions | undefined;
   getOptionsForType(type: string): TypeOptions | undefined;
   knownForType(type: string): KnownForTypeResult;
-  makeToString<T, C>(factory: Factory<T, C>, fullName: string): string;
+  makeToString(factory: Factory<object>, fullName: string): string;
```
wagenet, line 31: @chriskrycho I know this will make you happy.

packages/@ember/-internals/container/lib/registry.ts
```diff
@@ -217,8 +213,6 @@ export default class Registry implements IRegistry {
 
     let normalizedName = this.normalize(fullName);
 
-    this._localLookupCache = Object.create(null);
```
wagenet, line 220: This appears to be entirely unused.

packages/@ember/-internals/container/lib/registry.ts
```diff
@@ -550,25 +542,28 @@ if (DEBUG) {
...
+function resolve(
+  registry: Registry,
+  _normalizedName: string
+): Factory<object> | object | undefined {
   let normalizedName = _normalizedName;
 
   let cached = registry._resolveCache[normalizedName];
   if (cached !== undefined) {
-    return cached as Factory<T, C>;
+    return cached as Factory<object>;
```
wagenet: We can probably get rid of the `as` here.

packages/@ember/-internals/glimmer/lib/component-managers/outlet.ts
```diff
@@ -103,7 +103,10 @@ class OutletComponentManager
         let mountPoint = currentOwner.mountPoint;
 
         state.engine = currentOwner;
-        state.engineBucket = { mountPoint };
+
+        if (mountPoint) {
+          state.engineBucket = { mountPoint };
+        }
```
wagenet, line 109: The code that uses this seems to expect that `engineBucket` is only set if it has a `mountPoint`:

https://github.com/emberjs/ember.js/blob/649114f70f2c9288338845b50d7f18e519fc7c7f/packages/%40ember/-internals/glimmer/lib/component-managers/outlet.ts#L135-L144

packages/@ember/-internals/glimmer/tests/integration/helpers/helper-manager-test.js
```diff
@@ -43,6 +43,10 @@ class TestHelper {
   }
 
   willDestroy() {}
+
+  // HelperInstance
+  compute() {}
+  destroy() {}
```
wagenet, line 49: These methods always should have been set.

packages/@ember/-internals/owner/index.ts
```diff
@@ -11,36 +13,21 @@ export interface FactoryClass {
...
-  registerOptions(fullName: string, options: TypeOptions): void;
-  registeredOptions(fullName: string): TypeOptions | undefined;
-  registerOptionsForType(type: string, options: TypeOptions): void;
-  registeredOptionsForType(type: string): TypeOptions | undefined;
+export function isFactory(obj: unknown): obj is Factory<object> {
+  return obj != null && typeof (obj as Factory<object>).create === 'function';
 }
 
+// A combination of the public methods on ContainerProxyMixin and RegistryProxyMixin
+export interface Owner extends IRegistry, IContainer {}
```
wagenet, line 29: Reduce duplication!

packages/@ember/-internals/runtime/lib/system/namespace.ts
```diff
@@ -34,17 +36,26 @@ import EmberObject from './object';
   @public
 */
 export default class Namespace extends EmberObject {
-  init() {
+  static NAMESPACES = NAMESPACES;
+  static NAMESPACES_BY_ID = NAMESPACES_BY_ID;
+  static processAll = processAllNamespaces;
+  static byName = findNamespace;
+
+  constructor(owner?: Owner) {
```
wagenet, line 44: Switch back to `init`.

packages/@ember/application/instance.ts
```diff
@@ -52,7 +69,7 @@ const ApplicationInstance = EngineInstance.extend({
     @private
     @property {Object} customEvents
   */
-  customEvents: null,
+  customEvents: Record<string, string | null> | null = null;
```
wagenet, line 72: Needs a `declare` and proper default handling in `init`.

packages/@ember/application/instance.ts
```diff
@@ -271,36 +299,32 @@ const ApplicationInstance = EngineInstance.extend({
     return router
       .handleURL(location.getURL())
       .then(handleTransitionResolve, handleTransitionReject);
-  },
+  }
 
   willDestroy() {
```
wagenet, line 304: Do we expect people will/should override this?

packages/@ember/application/lib/application.ts
```diff
@@ -860,21 +924,21 @@ const Application = Engine.extend({
   */
   ready() {
     return this;
-  },
+  }
 
   // This method must be moved to the application instance object
   willDestroy() {
```
wagenet, line 930: Should/will people override this?

packages/@ember/application/type-tests/index.test.ts
```diff
@@ -13,21 +13,22 @@ class App extends Application {
...
 // You shouldn't really be setting this, so let's at least check that we can read it
-expectTypeOf(app.eventDispatcher).toEqualTypeOf<EventDispatcher>();
+expectTypeOf(app.eventDispatcher).toEqualTypeOf<EventDispatcher | null>();
 
 expectTypeOf(app.buildInstance()).toEqualTypeOf<ApplicationInstance>();
 expectTypeOf(app.deferReadiness()).toEqualTypeOf<void>();
 expectTypeOf(app.advanceReadiness()).toEqualTypeOf<void>();
-expectTypeOf(app.boot()).toEqualTypeOf<Promise<Application>>();
-expectTypeOf(app.ready()).toEqualTypeOf<void>();
+expectTypeOf(app.boot()).toEqualTypeOf<Promise<ApplicationInstance>>();
```
wagenet: I think this is wrong.

wagenet: Apparently this matcher will match any `Promise`. That seems unfortunate.

packages/@ember/engine/index.ts
```diff
@@ -2,18 +2,20 @@ export { getEngineParent, setEngineParent } from './lib/engine-parent';
...
+import EngineInstance, { EngineInstanceOptions } from '@ember/engine/instance';
 import { RoutingService } from '@ember/-internals/routing';
 import { ContainerDebugAdapter } from '@ember/-internals/extension-support';
 import { ComponentLookup } from '@ember/-internals/views';
 import { setupEngineRegistry } from '@ember/-internals/glimmer';
+import RegistryProxyMixin from '@ember/-internals/runtime/lib/mixins/registry_proxy';
+import { ResolverClass } from '@ember/-internals/container/lib/registry';
 
-function props(obj) {
+function props(obj: any) {
```
chriskrycho, line 18: 1. If we need to keep this, [`object` should be safe][1].

2. As far as I can tell, `for...in` has *exactly* the same semantics as `Object.keys()`:

    > `Object.keys()` returns an array whose elements are strings corresponding to the enumerable properties found directly upon `object`. The ordering of the properties is the same as that given by looping over the properties of the object manually.

    —[MDN `Object.keys()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys)

    > The for...in statement iterates over all [enumerable properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Enumerability_and_ownership_of_properties) of an object that are keyed by strings (ignoring ones keyed by [Symbols](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol)), including inherited enumerable properties.

    —[MDN `for...in`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in)

[1]: https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABABwE52QZwBRwEYBWAXIvgQKbQCUiA3gFCKIA25UK6y5qs5mJmKKhhgA5gG0AuogC8iKQG56jRMDipE2VuwDW5AJ6IRpQjQZMmaDN16YAdMhCYAFtj36qSpgF9lTVGwgqEhWXDwwfEq+QA

Net, I think we can just drop this in favor of `Object.keys()`.

packages/@ember/application/instance.ts
```diff
@@ -61,10 +65,10 @@ const ApplicationInstance = EngineInstance.extend({
     @private
     @property {String|DOMElement} rootElement
   */
-  rootElement: null,
+  rootElement: string | Element | null = null;
```
chriskrycho, line 68: This is the per-instance vs. per-prototype thing again. Here, though, I think it *should* not matter: it's a single string or element (or `null`!) and I *think* there will be a per-instance version of the actual string or element in any case—there basically has to be.

packages/@ember/application/instance.ts
```diff
@@ -120,15 +124,23 @@ const ApplicationInstance = EngineInstance.extend({
...
-    return this.lookup('router:main');
-  }).readOnly(),
+  get router() {
+    if (this._router) {
+      return this._router;
+    }
+
+    let router = this.lookup('router:main');
+    assert('expected an instance of Router', router instanceof Router);
+    return (this._router = router);
```
chriskrycho, line 142: In general @rwjblue and I both try to invert this pattern:

```suggestion
    if (!this._router) {
      let router = this.lookup('router:main');
      assert('expected an instance of Router', router instanceof Router);
      this._router = router;
    }

    return this._router;
```

Obviously just a taste thing, but it makes it so the condition is the place where the special case work happens, and the non-conditional flow is identical every time.

packages/@ember/application/instance.ts
```diff
@@ -120,15 +124,23 @@ const ApplicationInstance = EngineInstance.extend({
...
+  }
 
-  setupRegistry(options) {
-    this.constructor.setupRegistry(this.__registry__, options);
-  },
+  setupRegistry(options: BootOptions) {
+    (this.constructor as typeof ApplicationInstance).setupRegistry(this.__registry__, options);
+  }
+
+  _router?: Router;
```
chriskrycho, line 133: Given how this is used, it *shouldn't* matter over-much, but it's worth note that this results in a `defineProperty` emit when targeting older ES versions, and the corresponding semantics on later ES versions.

packages/@ember/application/instance.ts
chriskrycho: @rwjblue, thoughts?

packages/@ember/application/instance.ts
```diff
@@ -324,51 +335,124 @@ ApplicationInstance.reopenClass({
...
-      Interactive mode: whether we need to set up event delegation and invoke
-      lifecycle callbacks on Components.
-
-      @property isInteractive
-      @type boolean
-      @default auto-detected
-      @private
-    */
-    this.isInteractive = Boolean(environment.hasDOM); // This default is overridable below
+class CBootOptions {
```
chriskrycho, line 338: Why the name? To differentiate from the exported `BootOptions` interface? Sometimes for things like this I'll do `_BootOptions` or similar instead to indicate that it's distinct and internal-only.

chriskrycho (changes requested):
26/42 down! Will finish reviewing tomorrow morning.

packages/@ember/-internals/runtime/lib/mixins/container_proxy.d.ts
```diff
@@ -1,15 +1,17 @@
-import { Container } from '@ember/-internals/container';
+import Container, { FactoryManager } from '@ember/-internals/container/lib/container';
 import { TypeOptions } from '@ember/-internals/container/lib/registry';
 import { Mixin } from '@ember/-internals/metal';
 import { Factory } from '@ember/-internals/owner';
 
-interface ContainerProxyMixin {
+export interface IContainer {
```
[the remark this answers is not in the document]
wagenet: IIRC it was to disambiguate.

packages/@ember/-internals/runtime/lib/mixins/registry_proxy.d.ts
```diff
@@ -3,22 +3,33 @@ import { TypeOptions } from '@ember/-internals/container/lib/registry';
 import { Mixin } from '@ember/-internals/metal';
 import { Factory } from '@ember/-internals/owner';
 
-interface RegistryProxyMixin {
-  /** @internal */
-  __registry__: Registry;
-  resolveRegistration<T, C>(fullName: string): Factory<T, C> | undefined;
-  register(fullName: string, factory: Factory<unknown>, options?: TypeOptions): void;
+export interface IRegistry {
```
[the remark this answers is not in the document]
wagenet: IIRC it was to disambiguate.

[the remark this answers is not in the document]
wagenet: IIRC it was to disambiguate.

packages/@ember/-internals/runtime/lib/system/core_object.ts
```diff
@@ -28,7 +28,7 @@ import { _WeakSet as WeakSet } from '@glimmer/util';
 import { destroy, isDestroying, isDestroyed, registerDestructor } from '@glimmer/destroyable';
 import { OWNER } from '@glimmer/owner';
 
-type EmberClassConstructor<T> = new (owner: Owner) => T;
+type EmberClassConstructor<T> = new (owner?: Owner) => T;
```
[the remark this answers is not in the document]
wagenet: I think maybe sometimes.

packages/@ember/-internals/routing/lib/system/router.ts
```diff
@@ -302,9 +302,11 @@ class EmberRouter<R extends Route = Route> extends EmberObject.extend(Evented) i
     return path.join('.');
   }
 
-  constructor(owner: Owner) {
+  constructor(owner?: Owner) {
     super(owner);
 
+    assert('BUG: Missing owner', owner);
```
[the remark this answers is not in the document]
wagenet: Unfortunately, trying to make it required here when it is not required on in `EmberClassConstructor` confuses TS. Per https://github.com/emberjs/ember.js/blob/bf44cc0bee640d794bad8ffb62ee48fa28c9ced8/packages/%40ember/-internals/runtime/lib/system/core_object.ts#L774 it is not always required.

packages/@ember/-internals/runtime/lib/mixins/container_proxy.d.ts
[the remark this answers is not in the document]
wagenet: `IContainer` and `IRegistry` are used as part of the `Owner` interface. I don't want to merge with the mixins since `ContainerProxyMixin` since that has the `__container__` property and `RegistryProxyMixin` has the `__registry__` property, neither of which belong on `Owner`.

packages/@ember/engine/index.ts
wagenet: Making this change causes tests to fail... I haven't figured out why so I just left it as-is.

chriskrycho (approved):
:shipit:

PR #20000, opened 2022-03-01
export `renderSettled` via `@ember/renderer` package

cafreeman:
Serves as the first step in the implementation of https://github.com/cafreeman/rfcs/blob/dont-use-this-for-template-values-in-tests/text/0785-remove-set-get-in-tests.md

This PR simply exposes an existing utility (which is already fully tested) in a public package so that it can be consumed in `ember-test-helpers`.

### DIFF
--- a/packages/@ember/renderer/index.ts
+++ b/packages/@ember/renderer/index.ts
@@ -1,0 +1,21 @@
+/**
+  @module @ember/renderer
+  @public
+*/
+
+/**
+ * @class Renderer
+ * @public
+ */
+
+/**
+  Returns a promise which will resolve when rendering has settled. Settled in
+  this context is defined as when all of the tags in use are "current" (e.g.
+  `renderers.every(r => r._isValid())`). When this is checked at the _end_ of
+  the run loop, this essentially guarantees that all rendering is completed.
+
+  @method renderSettled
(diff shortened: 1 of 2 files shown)


jacobq:
No fair! You got github item number 20000, you lucky :laughing: (I got #20001 though)

cafreeman:
@jacobq my lucky day :)

chriskrycho (approved):
In case anyone else is wondering, I had to refresh myself on the RFC. We are indeed introducing a new top-level module here:

> In order to implement `rerender`, we would also expose the work done by @rwjblue on [renderSettled](https://github.com/emberjs/ember.js/blob/703b9ca2653a6b479a762b30dca1a33eaa13d8ab/packages/%40ember/-internals/glimmer/lib/renderer.ts#L219-L240) as public API in a new `@ember/renderer` module so that it could then be imported and used by `@ember/test-helpers` without issuue [sic].

chriskrycho:
Whoops! This could not have been merged yet, as we haven’t merged the RFC and it still needs the feature flag to go with it. We’ll get that later but sorted, probably early next week.

kategengler:
@chriskrycho Branch point should be next Monday (21st), so the flag is needed by then

chriskrycho:
cc @cafreeman

PR #20010, opened 2022-03-07
Lock down TS version

wagenet:
We should do this since there may be breaking changes in "minor" TS releases.


chriskrycho (approved):
Strongly agreed, good catch. Fast follow: bump to 4.6?

PR #20011, opened 2022-03-08
Remove suppressImplicitAnyIndexErrors

wagenet:


packages/@ember/-internals/routing/lib/system/router.ts
```diff
@@ -1735,7 +1745,7 @@ export function triggerEvent<R extends Route>(
     }
   }
 
-  let defaultHandler = defaultActionHandlers[name];
+  let defaultHandler = (defaultActionHandlers as any)[name];
```
chriskrycho, line 1748: Can we instead requiring that `name` be `keyof typeof defaultActionHandlers` or otherwise tighten this down to avoid the `as any`? Not the end of the world if not; we can do future passes looking for `as any`s, but figured I'd check.

packages/@ember/-internals/routing/lib/utils.ts
```diff
@@ -268,21 +264,22 @@ export function prefixRouteNameArg<T extends NamedRouteArgs<Route> | UnnamedRout
...
-export function shallowEqual(a: {}, b: {}): boolean {
-  let k;
+export function shallowEqual<A extends object, B extends object>(a: A, b: B): boolean {
+  let kA: keyof A;
   let aCount = 0;
   let bCount = 0;
-  for (k in a) {
-    if (Object.prototype.hasOwnProperty.call(a, k)) {
-      if (a[k] !== b[k]) {
+  for (kA in a) {
```
chriskrycho, line 271: As above, I'm confused why `for...in` isn't inferring correctly here. 🤔

packages/@ember/-internals/runtime/lib/system/object.ts
```diff
@@ -35,10 +35,12 @@ interface FrameworkObject extends EmberObject {}
 let FrameworkObject = class FrameworkObject extends EmberObject {};
 
 if (DEBUG) {
-  let INIT_WAS_CALLED = symbol('INIT_WAS_CALLED');
+  const INIT_WAS_CALLED = Symbol.for('INIT_WAS_CALLED');
```
chriskrycho, line 38:
```suggestion
  const INIT_WAS_CALLED = Symbol('INIT_WAS_CALLED');
```

packages/@ember/-internals/views/lib/compat/attrs.js
```diff
@@ -1,3 +1 @@
-import { symbol } from '@ember/-internals/utils';
-
-export let MUTABLE_CELL = symbol('MUTABLE_CELL');
+export const MUTABLE_CELL = Symbol.for('MUTABLE_CELL');
```
chriskrycho, line 1:
```suggestion
export const MUTABLE_CELL = Symbol('MUTABLE_CELL');
```

packages/@ember/debug/lib/deprecate.ts
```diff
@@ -127,7 +127,7 @@ if (DEBUG) {
 
       if (error instanceof Error) {
         if (error.stack) {
-          if (error['arguments']) {
+          if ((error as any)['arguments']) {
```
chriskrycho, line 130: We can use `in` here safely, esp. since we don't actually need to *use* `error.arguments` ([playgrounnd][1]):

```suggestion
          if ('arguments' in  error) {
```

[1]: https://www.typescriptlang.org/play?ts=4.6.2#code/CYUwxgNghgTiAEEQBd4gFzwK4DsDWOA9gO44BQAlgGbwAUCFOAzslDmCITQKIwyEwAlPADeZePGp0QAOhZQweYWIkSptAOSwA5lgC2IHMiYbJONMvGqJViQF8rDu0A

packages/@ember/engine/lib/engine-parent.ts
```diff
@@ -1,10 +1,9 @@
 /**
 @module @ember/engine
 */
-import { symbol } from '@ember/-internals/utils';
 import EngineInstance from '../instance';
 
-const ENGINE_PARENT = symbol('ENGINE_PARENT');
+export const ENGINE_PARENT: unique symbol = Symbol.for('ENGINE_PARENT');
```
chriskrycho, line 6:
```suggestion
export const ENGINE_PARENT = Symbol('ENGINE_PARENT');
```

packages/internal-test-helpers/lib/ember-dev/run-loop.ts
```diff
@@ -1,14 +1,14 @@
...
+  hooks.afterEach(function (assert: QUnit['assert']) {
     if (_getCurrentRunLoop() || _hasScheduledTimers()) {
       let done = assert.async();
       // use a setTimeout to allow the current run loop to flush via autorun
       setTimeout(() => {
         // increment expected assertion count for the assertions just below
-        if (assert['test'].expected !== null) {
-          assert['test'].expected += 2;
+        if ((assert as any)['test'].expected !== null) {
+          (assert as any)['test'].expected += 2;
```
chriskrycho, line 11: Weird, is `['test']` missing from the types for QUnit proper?

packages/internal-test-helpers/lib/factory.ts
```diff
@@ -1,7 +1,8 @@
-function setProperties(object: object, properties: object) {
+function setProperties<T extends object>(object: T, properties: Partial<T>) {
   for (let key in properties) {
     if (Object.prototype.hasOwnProperty.call(properties, key)) {
-      object[key] = properties[key];
+      let val = properties[key];
+      object[key] = val!;
```
chriskrycho, line 5:
```suggestion
      // SAFETY: we know that the property exists on `properties` because we
      // just checked as much with `hasOwnProperty`.
      object[key] = properties[key]!;
```

packages/internal-test-helpers/lib/get-all-property-names.ts
```diff
@@ -1,9 +1,10 @@
-export default function getAllPropertyNames(Klass: Function) {
+export default function getAllPropertyNames<T>(Klass: { prototype: T }): Set<keyof T & string> {
```
chriskrycho, line 1: What is the `& string` doing here? 🤔 Is it just for the sake of excluding non-`string` keys, e.g. symbols? If so, mind leaving a comment to that effect?

packages/internal-test-helpers/lib/get-all-property-names.ts
```diff
@@ -1,9 +1,10 @@
...
   let proto = Klass.prototype;
-  let properties: Set<string> = new Set();
+  let properties: Set<keyof T & string> = new Set();
 
   while (proto !== Object.prototype) {
-    let names = Object.getOwnPropertyNames(proto);
+    // SAFETY: This should be safe.
```
chriskrycho, line 6: …but why? :joy: Again, the explanation that the intersection with string gets us string-only properties of `T` is probably enough.

chriskrycho (changes requested):
Heroic effort! I have some feedback, but this is 98% 🚀

packages/@ember/-internals/glimmer/lib/resolver.ts
```diff
@@ -118,8 +118,18 @@ const BUILTIN_KEYWORD_HELPERS = {
   '-in-el-null': inElementNullCheckHelper,
 };
 
+const BUILTIN_HELPERS = {
```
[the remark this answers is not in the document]
wagenet: Yeah, I was trying to get more detailed types, but seems ok to punt on that for now.

packages/@ember/-internals/metal/lib/mixin.ts
```diff
@@ -539,6 +539,8 @@ export default class Mixin {
   /** @internal */
   _without: any[] | undefined;
 
+  [INIT_FACTORY]: null;
```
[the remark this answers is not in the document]
wagenet: Yeah. I think the optional null is correct here :/

packages/@ember/-internals/glimmer/lib/utils/process-args.ts
```diff
@@ -36,11 +35,14 @@ export function processComponentArgs(namedArgs: CapturedNamedArguments) {
...
 }
 
-const REF = symbol('REF');
+const REF = Symbol.for('REF');
 
 class MutableCell {
   public value: any;
-  constructor(ref: any, value: any) {
+  [MUTABLE_CELL]: boolean;
+  [REF]: Reference<unknown>;
```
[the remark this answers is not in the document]
wagenet: I'm not sure what it's supposed to be. I'll leave it for now.

packages/@ember/-internals/metal/lib/namespace_search.ts
```diff
@@ -109,11 +113,12 @@ function _processNamespace(paths: string[], root: Namespace, seen: Set<Namespace
   setName(root, id);
 
   // Loop over all of the keys in the namespace, looking for classes
-  for (let key in root) {
+  let key: keyof N;
+  for (key in root) {
```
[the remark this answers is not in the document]
wagenet: I don't know. I tried changing it back and it works. 🤷

packages/@ember/-internals/metal/lib/properties.ts
```diff
@@ -125,7 +125,7 @@ export function defineValue(
     if (DEBUG) {
       setWithMandatorySetter!(obj, keyName, value);
     } else {
-      obj[keyName] = value;
+      (obj as any)[keyName] = value;
```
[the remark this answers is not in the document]
wagenet: I tried this but it ends up being pretty far reaching. Probably something to try in another PR.

packages/@ember/-internals/metal/lib/property_get.ts
```diff
@@ -14,9 +14,9 @@ export let getPossibleMandatoryProxyValue: (obj: object, keyName: string) => any
 
 if (DEBUG) {
   getPossibleMandatoryProxyValue = function getPossibleMandatoryProxyValue(obj, keyName): any {
-    let content = obj[PROXY_CONTENT];
+    let content = (obj as any)[PROXY_CONTENT];
```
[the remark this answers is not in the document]
wagenet: Yeah, changes end up far reaching here as well.

packages/@ember/-internals/metal/lib/property_set.ts
```diff
@@ -70,15 +70,15 @@ export function _setProp(obj: object, keyName: string, value: any) {
   let descriptor = lookupDescriptor(obj, keyName);
 
   if (descriptor !== null && COMPUTED_SETTERS.has(descriptor.set!)) {
-    obj[keyName] = value;
+    (obj as any)[keyName] = value;
```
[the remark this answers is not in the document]
wagenet: Also far-reaching.

packages/@ember/-internals/routing/lib/location/api.ts
```diff
@@ -129,5 +126,5 @@ export default {
     return implementationClass.create(...arguments);
   },
 
-  implementations: {},
+  implementations: {} as Record<string, { create: (...args: any[]) => EmberLocation }>,
```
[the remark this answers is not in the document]
wagenet: See line 107 which fetches from `implementations` and then line 126 which calls `create` on that object expecting an `EmberLocation`.

packages/@ember/debug/lib/deprecate.ts
wagenet: Interestingly this breaks the next line that does `error.stack`.

packages/internal-test-helpers/lib/ember-dev/run-loop.ts
wagenet: Correct.

packages/internal-test-helpers/lib/get-all-property-names.ts
wagenet: Yes, there is code downstream that expects a string.

packages/@ember/-internals/glimmer/lib/resolver.ts
```diff
@@ -166,10 +166,12 @@ export default class ResolverImpl implements RuntimeResolver<Owner>, CompileTime
   lookupHelper(name: string, owner: Owner): Option<HelperDefinitionState> {
     assert(
       `You attempted to overwrite the built-in helper "${name}" which is not allowed. Please rename the helper.`,
-      !(BUILTIN_HELPERS[name] && owner.hasRegistration(`helper:${name}`))
+      !(
+        (BUILTIN_HELPERS as Record<string, object>)[name] && owner.hasRegistration(`helper:${name}`)
```
chriskrycho, line 170: If I'm reading the code right above, this is now how `BUILTIN_HELPERS` is defined type-wise, so the cast should be unnecessary? (And the same throughout below, of course.)

packages/@ember/-internals/glimmer/lib/resolver.ts
```diff
@@ -209,11 +211,11 @@ export default class ResolverImpl implements RuntimeResolver<Owner>, CompileTime
   }
 
   lookupBuiltInHelper(name: string): HelperDefinitionState | null {
-    return BUILTIN_KEYWORD_HELPERS[name] ?? null;
+    return (BUILTIN_KEYWORD_HELPERS as Record<string, object>)[name] ?? null;
   }
 
   lookupModifier(name: string, owner: Owner): Option<ModifierDefinitionState> {
-    let builtin = BUILTIN_MODIFIERS[name];
+    let builtin = (BUILTIN_MODIFIERS as Record<string, object>)[name];
```
chriskrycho, line 218: Can we (or do we) supply the same type definition such that it works correctly without the cast for *this one* here and below?

packages/@ember/-internals/routing/lib/system/router.ts
```diff
@@ -47,6 +47,15 @@ import Router, {
 import type { Timer } from 'backburner';
 import { EngineRouteInfo } from './engines';
 import EngineInstance from '@ember/engine/instance';
+import { QueryParams } from 'route-recognizer';
+
+type AnyFn = (...args: any[]) => any;
+
+type MethodNamesOf<T> = {
+  [K in keyof T]: T[K] extends AnyFn ? K : never;
+}[keyof T];
```
chriskrycho, line 56: Not urgent to do before merging, but we should definitely extract these kinds of common type utilities into a shared location to avoid duplicating them/drifting implementations.

packages/@ember/-internals/routing/lib/location/api.ts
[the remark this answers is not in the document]
chriskrycho: Ah, I did misread it. Thanks for correcting!

chriskrycho (approved):
I saw one more unnecessary cast and so pushed a fix for it; let's do this!

PR #20013, opened 2022-03-08
Upgrade glimmer

wagenet:


Windvis:
I think this is a duplicate of https://github.com/emberjs/ember.js/pull/19973 which also adds some tests.

wagenet:
@Windvis I believe you are correct. Thanks for the heads up!

PR #20016, opened 2022-03-10
Update docs to match native class syntax

ef4:
The example code snippet here assumes it will be used in a classic-style class body, but all the rest of the docs are updated for native class syntax.


locks (approved):

kategengler:
Does this also need to be done in master and beta?

locks:
@kategengler I would say so, yes

PR #20017, opened 2022-03-12
[BUGFIX beta] bind all methods on the router service for use with fn …

NullVoxPopuli:
…and default helper manager

**This PR is rebased off of https://github.com/emberjs/ember.js/pull/20018 so review that first**

This resolves an issue where this code:
```hbs
{{#if (this.router.isActive 'some.route')}}
```
Would lose the `this` binding.
[Discovered here](https://github.com/NullVoxPopuli/limber/pull/452/files#diff-968eba8222b55852309d2cd2c5814657ed4cdf53dfc78f9f40851a6e83fa47e0L21)

This is only an issue since the:
https://github.com/emberjs/rfcs/pull/756
(and https://github.com/NullVoxPopuli/ember-functions-as-helper-polyfill/)

[Explanation here](https://github.com/NullVoxPopuli/ember-functions-as-helper-polyfill/issues/103#issuecomment-1031496450)

_but_, I'm still upset that bound methods on classes aren't default 😅 
anywho, I used `@action` here, idk if that's the right call since action does a bunch of other stuff, too.

I also noticed that the refresh method was still marked as canary.
refresh was declared as released in 4.1: https://blog.emberjs.com/ember-4-1-released/
so I removed the feature flag.

### DIFF
--- a/packages/@ember/-internals/routing/lib/services/router.ts
+++ b/packages/@ember/-internals/routing/lib/services/router.ts
@@ -2,6 +2,7 @@ import { getOwner } from '@ember/-internals/owner';
 import { Evented } from '@ember/-internals/runtime';
 import { symbol } from '@ember/-internals/utils';
 import { assert } from '@ember/debug';
+import { action } from '@ember/object';
 import { readOnly } from '@ember/object/computed';
 import Service from '@ember/service';
 import { consumeTag, tagFor } from '@glimmer/validator';
@@ -122,6 +123,7 @@ class RouterService<R extends Route> extends Service.extend(Evented) {
        attempted transition
      @public
    */
+  @action
   transitionTo(...args: RouteArgs<R>): Transition {
     if (resemblesURL(args[0])) {
       // NOTE: this `args[0] as string` cast is safe and TS correctly infers it
@@ -171,6 +173,7 @@ class RouterService<R extends Route> extends Service.extend(Evented) {
        attempted transition
      @public
    */
+  @action
   replaceWith(...args: RouteArgs<R>): Transition {
     return this.transitionTo(...args).method('replace');
   }
@@ -243,6 +246,7 @@ class RouterService<R extends Route> extends Service.extend(Evented) {
      @return {String} the string representing the generated URL
      @public
    */
+  @action
   urlFor(routeName: string, ...args: ModelFor<R>[] | [...ModelFor<R>[], RouteOptions]) {
     this._router.setupRouter();
     return this._router.generate(routeName, ...args);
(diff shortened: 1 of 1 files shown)


NullVoxPopuli:
updated

chriskrycho:
@NullVoxPopuli mind rebasing now that #20018 is merged? That will help one of us review!

NullVoxPopuli:
@chriskrycho done! :tada:

chriskrycho:
@NullVoxPopuli we talked about this at a Framework team meeting a few weeks ago but responding fell off of folks' radars because of the focus on EmberConf. Net, we want to solve this *not* by binding these but by making direct method invocation work correctly—in which case, the invocation you have here will “just work”, and importantly folks will be able to use *any* methods directly. An example I’ve used a few times lately that falls out of that is using methods like [the relatively new `Array.prototype.at`](http://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at):

```ts
interface Signature {
  Args: {
    names: string[];
    selectedIdx: number;
  };
}

<template type:signature={Signature}>
  <p>{{(names.at selectedIdx)}}</p>
</template>
```

(N.B. that specific syntax doesn't yet work; but something *like* it should soon.)

chriskrycho:
I mentioned this in Discord but copying it here for folks following the thread: the idea is to align it with JS. If you have the array, it works, otherwise you need to bind it—no different than in JS.

chriskrycho:
I put this on the agenda for today’s spec group meeting (spec group == breakout session from the Framework Team meeting, where we deep dive on technical details of things, which other relevant folks are invited to as makes sense!) so we can hopefully make some progress on it.

chriskrycho:
The second chunk of [the notes from today's spec meeting notes](https://www.notion.so/emberteam/9a55c49297c34764839a9fecd8f3470e) covers the discussion on this!

kategengler:
Is this still relevant?

NullVoxPopuli:
Yup, here's a repro: https://limber.glimdown.com/edit?c=JIZwBAFgLlAOIC4D0SDmBLKECuAjAdAMYD2AtkgKam4UBOAViJdXfo0rNgDZdIBMABgEBGAOxgQUdDzC0KXCgDcAhgDsoAfgBQWgAb7UjMF3SKKYWHMXoKAdy3pSsYrShgAwmWeqK6sADNaMjAAcgABVBNSUjokEidiH3UQgG4HBNcwAG8wKFplQgBrCgATMABfAKDSUIiomNokPILC9FVUVPTnTJyQOmtCc0rA4PCqGka%2B2gGKTsdutxzEiqrRsPHY0mIS9H8bWk6tCgAPBbASin9lbjdCLmUQcAAJeS5iAHUXLjKTqF8S8CeBJJRZaMBgMJTGayYjYP60NLgsFgAA8fyc9z%2BAD5keCslksOgQPggnDWESAIKEKRmUKlTAhcrlZEoppUWCYig45n6XRaIA&format=glimdown

NullVoxPopuli:
I think we're going to solve this in the VM though, and have functions auto-bound to their host object.

e.g.:

`{{this.router.isActive "foo"}}`
is roughly the same as

```js
let isActive = this.router.isActive.bind(this.router);

isActive('foo');
```

PR #20019, opened 2022-03-12
[CLEANUP beta] get ci green again

NullVoxPopuli:
had a bit of a linssue

### DIFF
--- a/broccoli/packages.js
+++ b/broccoli/packages.js
@@ -268,7 +268,8 @@ module.exports.glimmerES = function glimmerES(environment) {
   if (environment === 'development') {
     let hasGlimmerDebug = true;
     try {
-      require.resolve('@glimmer/debug'); // eslint-disable-line node/no-missing-require
+      // eslint-disable-next-line
+      require.resolve('@glimmer/debug');
     } catch (e) {
       hasGlimmerDebug = false;
     }


broccoli/packages.js
```diff
@@ -268,7 +268,8 @@ module.exports.glimmerES = function glimmerES(environment) {
   if (environment === 'development') {
     let hasGlimmerDebug = true;
     try {
-      require.resolve('@glimmer/debug'); // eslint-disable-line node/no-missing-require
+      // eslint-disable-next-line node/no-extraneous-require
```
NullVoxPopuli, line 271: I guess a lint config changed somewhere and this wasn't updated 🤷

NullVoxPopuli: it also seems that local and CI have different sets of rules... 😩

broccoli/packages.js
```diff
@@ -268,7 +268,8 @@ module.exports.glimmerES = function glimmerES(environment) {
   if (environment === 'development') {
     let hasGlimmerDebug = true;
     try {
-      require.resolve('@glimmer/debug'); // eslint-disable-line node/no-missing-require
+      // eslint-disable-next-line
```
NullVoxPopuli, line 271: there appears to be a discrepancy between the lint rules used locally and what's on C.I.
so, I thought it best to just disable all of eslint for this next line, and we can figure out how to fix the environment later.

More important to unblock PRs for now

chriskrycho: It’s actually more important to figure out what’s actually happening: otherwise things could diverge further over time and that just creates more work over time!

NullVoxPopuli: disagree, all PRs are blocked because of how CI doesn't run tests unless linting passes.

chriskrycho: That is the correct and desirable outcome. If you want those unblocked, the way to do it is to go after the actual issue rather than trying to paper over it. Otherwise we’ll just be here doing the same thing again in a few weeks! 😅

NullVoxPopuli: well if you know how, be my guest. I have such limited time since all my ember.js/emberjs time right not is outside work hours, I think this is an acceptable tradeoff, and given how momentum is so slow in general, I think this is still the best approach _i can offer_, so ... if everyone wants stale PRs 🤷 this is all I got

kategengler:
Afaict, CI is green on [master](https://github.com/emberjs/ember.js/actions?query=branch%3Amaster) and [beta](https://github.com/emberjs/ember.js/actions?query=branch%3Abeta), so I'm curious what is going on here

chriskrycho:
@NullVoxPopuli can you try rebasing? Otherwise, this may be something specific to your PRs, because #20020 is also ✅ with no issues.

NullVoxPopuli:
Locally, using Volta, I:
 - reset to latest emberjs/ember.js (reset --hard)
 - yarn
 - yarn lint:eslint --quiet
 - no errors :tada: 
 - switch to this branch
 - rebase
 - yarn
 - yarn lint:eslint --quiet
 - no errors :tada: 
 - try reverting commit
 - yarn lint:eslint --quiet
 - no errors :tada: 
 
 ¯\_( ツ )_/¯

PR #20020, opened 2022-03-13
[DOCS] Add explicit module names for the `RouteInfo` and `Transition` classes

Windvis:
This came up in a [Discord discussion](https://discord.com/channels/480462759797063690/480777444203429888/945819493479886858). 

[`RouteInfo`](https://api.emberjs.com/ember/release/classes/RouteInfo) is currently listed under the ember module instead of `@ember/routing`. It seems that YUIDoc uses the last defined module name if a class doesn't define one itself. It also seems to parse files alphabetically. RouteInfo doesn't define a module name and comes after the `generate_controller` file which uses `ember` as the module name. Adding the explicit module name fixes the problem:

image

Transition does appear in the right package, but I added the module anyway to prevent it from changing in the future if files are moved / refactored.

### DIFF
--- a/packages/@ember/-internals/routing/lib/system/route-info.ts
+++ b/packages/@ember/-internals/routing/lib/system/route-info.ts
@@ -1,6 +1,10 @@
 export { RouteInfo, RouteInfoWithAttributes } from 'router_js';
 
 /**
+ @module @ember/routing
+*/
+
+/**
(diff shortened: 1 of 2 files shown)


chriskrycho (approved):

chriskrycho:
Thank you!

PR #20024, opened 2022-03-16
[BUGFIX beta] Correctly associate props with factory and owner in FactoryManager

### LINKED ISSUE #20023: Incorrect association of props with owner and factory in FactoryManager.create?
This code is either wrong or quite confusing:

https://github.com/emberjs/ember.js/blob/d8f82663599cdced3f51a5b6b0e73308b640fc8c/packages/%40ember/-internals/container/lib/container.ts#L507-L513

It appears to be assuming that the `setOwner` and `setFactoryFor` will continue to work for `props` after it is rebound on line 512—but they do not, because it is a completely new object. If I'm reading this method correctly, the original, empty POJO will simply be GC'd, since it's unused after the end of the method.

This has been this way for quite some time, but the cleanups for Ember v4 made it much more apparent.

I *believe* the correct update here is to change the entire chunk to simply be:

```ts
let props = options ? { ...options } : {};
setOwner(props, container.owner!);
setFactoryFor(props, this);
```

That would maintain the object identity and therefore make owner and factory lookups for this `props` object continue to work correctly.

chriskrycho:
Previously, we stomped the `props` binding with an `Object.assign()`, which meant that the original empty props object would get GC'd after the end of the method and the item passed into the class created at the end of the `FactoryManager.create` call would be a *different* object, which does *not* have the factory or owner associations.

Fixes #20023

Also fixes a bunch of unnecessary type casts.

### DIFF
--- a/packages/@ember/-internals/container/lib/container.ts
+++ b/packages/@ember/-internals/container/lib/container.ts
@@ -504,13 +504,9 @@ export class FactoryManager<T, C extends FactoryClass | object = FactoryClass> {
       );
     }
 
-    let props = {};
+    let props = options ? { ...options } : {};
     setOwner(props, container.owner!);
     setFactoryFor(props, this);
-
-    if (options !== undefined) {
-      props = Object.assign({}, props, options);
-    }
 
     if (DEBUG) {
       let lazyInjections;


rwjblue (changes requested):
The container changes seem good! Can you revert the `routing.ts` / `router.ts` changes (and move those type only things to another PR)?

chriskrycho:
Yes! Thanks for flagging those up; did not mean to include them! Updating momentarily.

chriskrycho:
I've just checked and I do not think we can backport this one—not only b/c it's not a clean back port, but because it's doesn't work when I do the equivalent refactor there! It also doesn't affect the underlying issue we found in #20025, so it seems fine to me to just land it a fix on master and beta.

PR #20033, opened 2022-03-18
Fix RouterService#currentRouteName documentation

lcoq:
The documentation was probably copy-pasted from RouterService#currentURL


bertdeblock:
@lcoq I think the idea of the docs here is to display what `currentRouteName` would be for a specific _URL_, so I'm not sure if these changes are correct.

lcoq:
indeed, sorry

PR #20040, opened 2022-03-31
[BUGFIX LTS] setupRouter on property access (regression since >3.25)

NullVoxPopuli:
This still needs to be tested -- but an issue with currentRoute being falsey in rendering tests popped up in a 3.25 -> 3.28 upgrade.

cc @gnclmorais

### DIFF
--- a/packages/@ember/-internals/routing/lib/services/router.ts
+++ b/packages/@ember/-internals/routing/lib/services/router.ts
@@ -2,7 +2,6 @@ import { getOwner, Owner } from '@ember/-internals/owner';
 import { Evented } from '@ember/-internals/runtime';
 import { symbol } from '@ember/-internals/utils';
 import { assert } from '@ember/debug';
-import { readOnly } from '@ember/object/computed';
 import { assign } from '@ember/polyfills';
 import Service from '@ember/service';
 import { consumeTag, tagFor } from '@glimmer/validator';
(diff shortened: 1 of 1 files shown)


packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -511,7 +510,10 @@ RouterService.reopen(Evented, {
      @type String
      @public
    */
-  currentRouteName: readOnly('_router.currentRouteName'),
```
NullVoxPopuli, line 514: why were these still using readOnly?

chriskrycho: Likely because no one had updated them.

However, we should test to confirm that this has the same semantics: I *believe* people depending on these intermediate getters with classic computed properties (`@computed` or any of the CP macros) will not get the expected reactivity behavior unless these are decorated with `@dependentKeyCompat`.

NullVoxPopuli: ah, so would it be safe to say that _until_ `@computed` is deprecated _and_ removed, all getters in the framework must use `@dependentKeyCompat`?

chriskrycho: Minimally, that is true for anything which should be 'reactive'.

chriskrycho:
@NullVoxPopuli is this still relevant? I believe there were other fixes that landed on 3.28 that addressed related issues.

NullVoxPopuli:
Probably, i'll rebase and double check

NullVoxPopuli:
I haven't come across a need for this in a long while, so I'm going to close for now

PR #20041, opened 2022-03-31
enhancement: do not throw Proxy `Ember.get` error from CoreObject

snewcomer:
It would be nice to get rid of this message, especially when returning async data to a component tree without resolving at the Route.

Eps since Proxy is supported in 4.0 (dropped IE11)

### DIFF
--- a/packages/@ember/-internals/runtime/lib/system/core_object.ts
+++ b/packages/@ember/-internals/runtime/lib/system/core_object.ts
@@ -244,21 +244,7 @@ class CoreObject {
     (this.constructor as typeof CoreObject).proto();
 
     let self;
-    if (DEBUG && hasUnknownProperty(this)) {
-      let messageFor = (obj: unknown, property: unknown) => {
-        return (
-          `You attempted to access the \`${String(property)}\` property (of ${obj}).\n` +
-          `Since Ember 3.1, this is usually fine as you no longer need to use \`.get()\`\n` +
-          `to access computed properties. However, in this case, the object in question\n` +
-          `is a special kind of Ember object (a proxy). Therefore, it is still necessary\n` +
-          `to use \`.get('${String(property)}')\` in this case.\n\n` +
-          `If you encountered this error because of third-party code that you don't control,\n` +
-          `there is more information at https://github.com/emberjs/ember.js/issues/16148, and\n` +
-          `you can help us improve this error message by telling us more about what happened in\n` +
-          `this situation.`
-        );
-      };
-
+    if (hasUnknownProperty(this)) {
       /* globals Proxy Reflect */
       self = new Proxy(this, {
         get(target: CoreObject & HasUnknownProperty, property, receiver) {
@@ -282,12 +268,6 @@ class CoreObject {
             property in target
           ) {
             return Reflect.get(target, property, receiver);
-          }
-
-          let value = target.unknownProperty.call(receiver, property);
-
-          if (typeof value !== 'function') {
-            assert(messageFor(receiver, property), value === undefined || value === null);
           }
         },
       });
--- a/packages/@ember/-internals/runtime/lib/system/core_object.ts
+++ b/packages/@ember/-internals/runtime/lib/system/core_object.ts
@@ -258,11 +258,6 @@ class CoreObject {
             property === 'toJSON' ||
             property === 'toString' ||
             property === 'toStringExtension' ||
-            property === 'didDefineProperty' ||
-            property === 'willWatchProperty' ||
-            property === 'didUnwatchProperty' ||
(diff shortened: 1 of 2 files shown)


packages/@ember/-internals/runtime/lib/system/core_object.ts
```diff
@@ -245,20 +245,6 @@ class CoreObject {
...
-          `to use \`.get('${String(property)}')\` in this case.\n\n` +
-          `If you encountered this error because of third-party code that you don't control,\n` +
-          `there is more information at https://github.com/emberjs/ember.js/issues/16148, and\n` +
-          `you can help us improve this error message by telling us more about what happened in\n` +
-          `this situation.`
-        );
-      };
-
       /* globals Proxy Reflect */
       self = new Proxy(this, {
```
snewcomer, line 249: Should we always return a proxy in Prod as well?

ef4:
I don't think this is correct. The error that this assertion is trying to prevent still exists. Running this branch, I tested:

```js
import ObjectProxy from '@ember/object/proxy';

let p = ObjectProxy.extend({
  unknownProperty(prop) {
    return `you said ${prop}`
  }
}).create();

console.log(p.get('thing1')); // logs "you said thing1"
console.log(p.thing2); // logs undefined
```

I think it would be *possible* to use native Proxy to make `unknownProperty` work without `get`, but that's not what this PR is doing. It's turning an assertion into a silent error.

Given that a browser-native alternative to ObjectProxy exists, the bar for reimplementing ObjectProxy is high -- a successful PR would need to convincingly demonstrate that it's not going to break any weird edge cases, since the whole reason ObjectProxy exists is for backward compatibility.

Users who don't want to use `get` should switch to native Javascript Proxy.

ef4:
I don't think this is correct. The error that this assertion is trying to prevent still exists. Running this branch, I tested:

```js
import ObjectProxy from '@ember/object/proxy';

let p = ObjectProxy.extend({
  unknownProperty(prop) {
    return `you said ${prop}`
  }
}).create();

console.log(p.get('thing1')); // logs "you said thing1"
console.log(p.thing2); // logs undefined
```

I think it would be *possible* to use native Proxy to make `unknownProperty` work without `get`, but that's not what this PR is doing. It's turning an assertion into a silent error.

Given that a browser-native alternative to ObjectProxy exists, the bar for reimplementing ObjectProxy is high -- a successful PR would need to convincingly demonstrate that it's not going to break any weird edge cases, since the whole reason ObjectProxy exists is for backward compatibility.

Users who don't want to use `get` should switch to native Javascript Proxy.

snewcomer:
Thanks for the response.  That was the gist of a few questions I have posed in various channels. 

> @rwjblue Would you happen to have a minute to comment on the above PR.  Without this error, an ObjectProxy would return undefined for obj.foo, which I assume is still not desired.  Do you see a path forward to getting rid of this message now that we support Proxies?

We might have to work on this in ember-data land.  A common scenario is trying to optimistically render (not block at model hook) a page with not-yet-resolved ember-data relationships.  So, as you stated, we might not be able to solve it here (ember-m3 solved with with some clever Proxy behavior I heard).

snewcomer:
closing for now...

chancancode:
👍🏼 we should probably remove [this reference](https://github.com/snewcomer/ember.js/blob/32dc083d4db6d93b837c366f5f157bb32e8bc6cf/packages/%40ember/-internals/runtime/lib/system/core_object.ts#L256) to the issue though, since the issue was closed 🙃 

In any case, I agree the more actionable/higher payoff thing to do is probably refactor Ember Data away from using _this_ proxy implementation. We should basically deprecate this as we really don't want any more new code coming to rely on this, but we also don't want to spew a bunch of deprecations to Ember Data users.

Coming to think of it, I suppose a middle ground is to start emitting the this deprecation for 5.0, but export a private handshake for Ember Data to disable the message for the ED use cases as long as the ED team feels confident that it can be resolved by 5.0?

chancancode:
That being said, I do think there is a path for removing the need for this now that we have native proxies in the browser.

We [require](https://github.com/snewcomer/ember.js/blob/32dc083d4db6d93b837c366f5f157bb32e8bc6cf/packages/%40ember/-internals/runtime/lib/system/core_object.ts#L210-L216) the use of `.create()` to instantiate `EmberObject`s, so it is definitely possible to move the current logic around so that `EmberProxy.create()` returns a native `Proxy` object that encapsulate the `unknownProperty` logic (or just direct the Proxy's `get` trap to `Ember.get` on the wrapped object).

It doesn't have to be either this or Ember Data, we could do both. But if you have to pick one I do think the effort is better spent refactoring things from depending on Ember proxies.

PR #20042, opened 2022-04-01
Add a couple of tests to test passing empty `queryParams` to `replaceWith`

boris-petrov:
Which should fail. Check [this](https://github.com/emberjs/ember.js/pull/19971#issuecomment-1074771144).

### DIFF
--- a/packages/ember/tests/routing/router_service_test/replaceWith_test.js
+++ b/packages/ember/tests/routing/router_service_test/replaceWith_test.js
@@ -136,5 +136,29 @@ moduleFor(
           assert.deepEqual(this.state, ['/', '/child']);
         });
     }
+
+    ['@test RouterService#replaceWith with empty query params'](assert) {
+      assert.expect(1);
+
+      return this.visit('/')
+        .then(() => {
+          return this.routerService.replaceWith('parent.child', { queryParams: {} });
+        })
+        .then(() => {
+          assert.deepEqual(this.state, ['/child']);
+        });
+    }
+
+    ['@test RouterService#replaceWith with `undefined` in query params'](assert) {
+      assert.expect(1);
+
+      return this.visit('/')
+        .then(() => {
+          return this.routerService.replaceWith('parent.child', { queryParams: undefined });
+        })
+        .then(() => {
+          assert.deepEqual(this.state, ['/child']);
+        });
+    }
   }
 );


packages/ember/tests/routing/router_service_test/replaceWith_test.js
```diff
@@ -136,5 +136,29 @@ moduleFor(
...
+          assert.deepEqual(this.state, ['/child']);
+        });
+    }
+
+    ['@test RouterService#replaceWith with `undefined` in query params'](assert) {
+      assert.expect(1);
+
+      return this.visit('/')
+        .then(() => {
+          return this.routerService.replaceWith('parent.child', { queryParams: undefined });
```
wagenet, line 157: In this case, this is something we could support but it's also not 100% clear to me that we want to. Can you explain more to me about the use case?

wagenet (changes requested):
I've added some questions.

packages/ember/tests/routing/router_service_test/replaceWith_test.js
```diff
@@ -136,5 +136,29 @@ moduleFor(
           assert.deepEqual(this.state, ['/', '/child']);
         });
     }
+
+    ['@test RouterService#replaceWith with empty query params'](assert) {
+      assert.expect(1);
+
+      return this.visit('/')
+        .then(() => {
+          return this.routerService.replaceWith('parent.child', {});
```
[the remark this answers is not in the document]
boris-petrov: Sorry, you're right, fixed!

packages/ember/tests/routing/router_service_test/replaceWith_test.js
boris-petrov: Well, you can see what I had in my code [here](https://github.com/emberjs/ember.js/pull/19971#issuecomment-1074771144). The idea is that users of that component could set in `model` their own `routeComponents` and (optionally) `queryParameters` and the component would transition to that. Users might *not* add `queryParameters` which would mean that this call:

```ts
this.router.replaceWith(...this.model.routeComponents, { queryParams: this.model.queryParameters })
```

Would become:

```ts
this.router.replaceWith(...this.model.routeComponents, { queryParams: undefined })
```

Which used to work on Ember < 4.3.0 but broke on 4.3.0. Not saying it's good, just saying it used to work. And is useful. :smile:

wagenet: Fair enough! I'll look into it more next week.

wagenet: So I played around with this some and it looks like router.js would also need an update to handle this case correctly. In your case, a simple solution is probably `this.model.queryParameters || {}`. I'm open to more discussion however.

boris-petrov: hm, but as I said, this used to work in the previous Ember version without changes to `router.js`. But I don't know the current code, of course. Also, note that `this.model.queryParameters || {}` also doesn't work - check the first test of the two - it tests exactly that (and fails).

wagenet: Ah, empty query params should work at least. I'll see about fixing that.

wagenet: Also, just because something worked before doesn't mean we guarantee that it will keep working, only if it is something that was clearly documented or widely used to the point where it's basically official. Honestly, this whole API is somewhat awkward so it's not actually clear to me what's most correct. I'm happy to hear another opinion on this.

boris-petrov:
I've involuntarily closed this. I've opened it again as it's still relevant I believe.

PR #20052, opened 2022-04-09
[FEATURE default helper manager] Add the default helper manager

Windvis:
This updates the Glimmer VM to version 0.84.2 which includes the default helper manager and a feature flag to conditionally enable that feature in Ember.

This PR supersedes #19973.


Windvis:
Draft since there are more changes needed in the VM when using functions that are stored on class properties or passed as arguments.

chancancode (approved):
@Windvis do you have anything else you think needs to be done or are you happy with everything? at a glance it seems good to me

wycats:
@chancancode any reason not to merge this?

Windvis:
@chancancode I think it's good to go. :+1: The RFC mentions some documentation updates but that's part of the guides so nothing needs to be added here I think.

Windvis:
@chancancode I think it's good to go. :+1: The RFC mentions some documentation updates but that's part of the guides so nothing needs to be added here I think.

PR #20059, opened 2022-04-15
use Math.sign for spaceship function implementation

### LINKED ISSUE #19988: Replace internal "spaceship" function with Math.sign
This:

https://github.com/emberjs/ember.js/blob/1cc378e3a865774e87b238953cf81e1a41fd244a/packages/%40ember/-internals/runtime/lib/compare.ts#L38-L42

Can be replaced with [the `Math.sign` function](http://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/sign), which is supported on all modern (i.e. non-IE) browser.

locks:
Closes https://github.com/emberjs/ember.js/issues/19988.

### DIFF
--- a/packages/@ember/-internals/runtime/lib/compare.ts
+++ b/packages/@ember/-internals/runtime/lib/compare.ts
@@ -36,6 +36,9 @@ type Compare = -1 | 0 | 1;
 //                                `._________`-.   `.   `.___
 //                                              SSt  `------'`
 function spaceship(a: number, b: number): Compare {
+  // SAFETY: `Math.sign` always returns `-1` for negative, `0` for zero, and `1`
+  // for positive numbers. (The extra precision is useful for the way we use
+  // this in the context of `compare`.)
   return Math.sign(a - b) as Compare;
 }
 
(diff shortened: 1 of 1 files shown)


packages/@ember/-internals/runtime/lib/compare.ts
```diff
@@ -36,9 +36,7 @@ type Compare = -1 | 0 | 1;
 //                                `._________`-.   `.   `.___
 //                                              SSt  `------'`
 function spaceship(a: number, b: number): Compare {
-  let diff = a - b;
-  // SAFETY: Number casts true into 1 and false into 0. Therefore, this must end up as one of the Compare values.
```
chriskrycho, line 40: The comment needs updating but needs to remain because this OS do an unsafe cast.

chriskrycho (changes requested):
Thanks! Needs one tweak!

packages/@ember/-internals/runtime/lib/compare.ts
chriskrycho: Resolved via a5027fefd

chriskrycho (approved):

PR #20064, opened 2022-04-25
Convert view internals to TS

wagenet:


packages/@ember/-internals/glimmer/lib/component.ts
```diff
@@ -857,7 +857,7 @@ class Component
       if (!lazyEventsProcessedForComponentClass.has(proto)) {
         let lazyEvents = eventDispatcher.lazyEvents;
 
-        lazyEvents.forEach((mappedEventName: string, event: string) => {
+        lazyEvents.forEach((mappedEventName: string | null, event: string) => {
```
wagenet, line 860: Given the check on the next line we definitely expect a possible null here.

packages/@ember/-internals/views/lib/mixins/action_support.ts
```diff
@@ -5,8 +5,16 @@ import { inspect } from '@ember/-internals/utils';
...
 
-const mixinObj = {
-  send(actionName, ...args) {
+/**
+ @class ActionSupport
+ @namespace Ember
+ @private
+*/
+interface ActionSupport {
+  send(actionName: string, ...args: unknown[]): void;
```
wagenet, line 14: @chriskrycho's favorite!

packages/@ember/-internals/views/lib/system/event_dispatcher.ts
```diff
@@ -239,50 +249,58 @@ export default EmberObject.extend({
...
+        if (actionState) {
+          actions = [actionState];
+        }
       }
 
       // We have to check for actions here since in some cases, jQuery will trigger
       // an event on `removeChild` (i.e. focusout) after we've already torn down the
       // action handlers for the view.
       if (!actions) {
+        // FIXME: This branch is never called in tests. Improve tests or remove
```
wagenet, line 303: Not sure if there's anything more we should do about this right now.

packages/@ember/-internals/glimmer/lib/component.ts
chriskrycho: Out of curiosity, does `lazyEvents` itself not have the right type? Seems like we want it to be there instead?

packages/@ember/-internals/views/lib/mixins/action_support.ts
chriskrycho: ***BURN IT TO THE GROUND. SALT THE GROUND WHERE IT STOOD. GLASS THE GROUND FROM ORBIT.***

packages/@ember/-internals/views/lib/mixins/view_support.ts
```diff
@@ -3,12 +3,34 @@ import { descriptorForProperty, Mixin, nativeDescDecorator } from '@ember/-inter
 import { assert } from '@ember/debug';
 import { hasDOM } from '@ember/-internals/browser-environment';
 import { matches } from '../system/utils';
+import type Component from '@ember/component';
+import { View } from '@ember/-internals/glimmer/lib/renderer';
+import { SimpleElement } from '@simple-dom/interface';
 
-function K() {
+function K(this: unknown) {
```
chriskrycho, line 10: BURN IT. GAHHHHH.

packages/@ember/-internals/views/lib/system/event_dispatcher.ts
chriskrycho: Huh. Is it even an issue anymore given that we don't have jQuery *integration* anymore?

chriskrycho (approved):

packages/@ember/-internals/glimmer/lib/component.ts
wagenet: Types are correct. Looks like this ends up being redundant.

packages/@ember/-internals/views/lib/system/event_dispatcher.ts
wagenet: I suspect we just don't need it. But am also a bit hesitant to remove it without fully understanding.

PR #20065, opened 2022-04-26
Convert more internals to TS

wagenet:


packages/@ember/-internals/metal/lib/array.ts
```diff
@@ -17,11 +20,17 @@ export function objectAt<T>(array: T[] | EmberArray<T>, index: number): T | unde
 }
 
 export function replace<T>(
-  array: NativeArray<T> | MutableArray<T>,
+  array: T[] | MutableArray<T>,
```
wagenet, line 23: Since we're checking `Array.isArray` below we should obviously be allowing any array.

packages/@ember/-internals/metal/lib/computed.ts
```diff
@@ -903,8 +903,8 @@ export function computed(
 }
 
 export function autoComputed(
-  ...config: [ComputedPropertyObj]
-): ComputedDecorator | DecoratorPropertyDescriptor {
+  ...config: [ComputedPropertyObj | ComputedPropertyGetterFunction]
+): ComputedDecorator {
```
wagenet, line 907: These types are so weird, but this seems to be what we want.

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
```diff
@@ -26,27 +26,27 @@ interface EmberArray<T> extends Enumerable {
   ): NativeArray<U>;
   mapBy<K extends string>(key: K): NativeArray<Value<T, K>>;
   filter<Target>(
-    callback: (this: Target, item: T, index: number, arr: this) => boolean,
+    callback: (this: Target, item: T, index: number, arr: this) => unknown,
```
wagenet, line 29: I believe this is correct. We don't want to force people to return a specific type, allowing people to pass values that will get evaluated as truthy/falsy.

packages/@ember/-internals/runtime/lib/mixins/array.d.ts
```diff
@@ -103,5 +103,8 @@ export function A<A>(arr: A): A extends Array<infer V> ? NativeArray<V> : Native
 export function A<T>(): NativeArray<T>;
 
 export function removeAt<T>(array: T[] | EmberArray<T>, start: number, len: number): EmberArray<T>;
-export function uniqBy<T>(array: T[], keyOrFunc: string | ((item: T) => unknown)): T[];
-export function isArray(obj: unknown): boolean;
+export function uniqBy<T>(
+  array: T[] | EmberArray<T>,
+  keyOrFunc: string | ((item: T) => unknown)
+): T[] | EmberArray<T>;
+export function isArray(obj: unknown): obj is ArrayLike<unknown> | EmberArray<unknown>;
```
wagenet, line 110: Looking at the underlying code I believe this is pretty similar to what actually happens. TBH we probably want to stop using this method sooner rather than later.

packages/@ember/-internals/runtime/lib/system/array_proxy.ts
```diff
@@ -30,7 +30,7 @@ import {
 } from '@glimmer/validator';
 import { PropertyDidChange } from '@ember/-internals/metal/lib/property_events';
 
-function isMutable<T>(obj: EmberArray<T>): obj is MutableArray<T> {
+function isMutable<T>(obj: T[] | EmberArray<T>): obj is T[] | MutableArray<T> {
```
wagenet, line 33: Judging by the following line this clearly should allow regular arrays.

packages/@ember/-internals/runtime/lib/system/array_proxy.ts
```diff
@@ -226,7 +226,7 @@ class ArrayProxy<T, C extends EmberArray<T> = EmberArray<T>>
     let content = get(this, 'content');
     assert('[BUG] Called objectAtContent without content', content);
     assert('Mutating a non-mutable array is not allowed', isMutable(content));
-    content.replace(idx, amt, objects);
+    replace(content, idx, amt, objects);
```
wagenet, line 229: `content.replace` would only work on EmberArray, but the utility method `replace` will work on either.

packages/@ember/object/lib/computed/computed_macros.ts
```diff
@@ -47,13 +48,13 @@ function generateComputedWithPredicate(name, predicate) {
       let lastIdx = dependentKeys.length - 1;
 
       for (let i = 0; i < lastIdx; i++) {
-        let value = get(this, dependentKeys[i]);
+        let value = get(this, dependentKeys[i]!);
```
wagenet, line 51: Should add a SAFETY comment here.

packages/@ember/object/lib/computed/computed_macros.ts
```diff
@@ -47,13 +48,13 @@ function generateComputedWithPredicate(name, predicate) {
...
         if (!predicate(value)) {
           return value;
         }
       }
 
-      return get(this, dependentKeys[lastIdx]);
+      return get(this, dependentKeys[lastIdx]!);
```
wagenet, line 57: Should add a SAFETY comment here.

packages/@ember/object/computed.ts
```diff
@@ -1,5 +1,13 @@
 export { ComputedProperty as default, expandProperties, alias } from '@ember/-internals/metal';
 
+// import { DeprecationOptions } from '@ember/debug';
+
+// export { alias, expandProperties } from '@ember/-internals/metal';
+
+// export function or(dependentKey: string, ...additionalDependentKeys: string[]): PropertyDecorator;
+
+// export function reads(dependentKey: string): PropertyDecorator;
+
```
chriskrycho, line 10: Looks like a mistake that these are left in but commented out?

packages/@ember/object/lib/computed/computed_macros.ts
chriskrycho:
```suggestion
        // SAFETY: `lastIdx` is derived from the length of `dependentKeys`
        let value = get(this, dependentKeys[i]!);
```

packages/@ember/object/lib/computed/computed_macros.ts
chriskrycho:
```suggestion
      // SAFETY: `lastIdx` is derived from the length of `dependentKeys`
      return get(this, dependentKeys[lastIdx]!);
```

packages/ember-testing/lib/adapters/adapter.js
```diff
@@ -1,9 +1,5 @@
 import { Object as EmberObject } from '@ember/-internals/runtime';
 
-function K() {
-  return this;
-}
-
```
chriskrycho, line 6: BLESSINGS ON YOUR HEAD FOR A THOUSAND YEARS

packages/@ember/-internals/glimmer/lib/renderer.ts
```diff
@@ -239,7 +235,7 @@ export function renderSettled() {
     // a chance to resolve (because its resolved in backburner's "end" event)
     if (!_getCurrentRunLoop()) {
       // ensure a runloop has been kicked off
-      _backburner.schedule('actions', null, K);
```
chriskrycho, line 242: *Might* still make sense to have an extracted `const NO_OP = () => {}` here so we don't allocate a new one for every one of these, even if the name `K` is accursed forever.

chriskrycho (approved):
LGTM overall – let's get those `SAFETY` comments in and then merge it. Thank you again!

packages/@ember/object/computed.ts
wagenet: Yeah, they should go.

packages/@ember/-internals/glimmer/lib/renderer.ts
wagenet: Do you think I should do this before merge? It seems unlikely that anyone else will go back and change it in the future if we don't do it now.

chriskrycho: Gut check, yes, for exactly that reason.

packages/@ember/-internals/glimmer/lib/renderer.ts
```diff
@@ -266,7 +264,7 @@ function loopEnd() {
         throw new Error('infinite rendering invalidation detected');
       }
       loops++;
-      return _backburner.join(null, K);
+      return _backburner.join(null, () => {});
```
chriskrycho, line 267:
```suggestion
      return _backburner.join(null, NO_OP);
```

chriskrycho (approved):

PR #20087, opened 2022-05-10
feat: import type

snewcomer:
Typescript already [elides](https://www.typescriptlang.org/play/?ssl=1&ssc=1&pln=3&pc=1#code/JYWwDg9gTgLgBAbzgeQHYBsCeBVAzgUwBMBBXAFUzHzgF84AzKCEOAIlYG4AoUSWROHiKkAagEN0AV2p1GzNpy5dC+AMboxUaqoipc8MQC4UGHARLlK+bjr3wARnAC8g86InTuQA) imports that are only used for type annotations. This PR proposes to make the import type explicit and not allow import X if only used in the type position.

auto fixed with eslint.

This PR checks to see if we can support `import type` syntax. I'm looking at this because I often find it hard to determine what imports (esp from third party packages) are for types or runtime code. 

## Docs
https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/consistent-type-imports.md

### TODO
sort type imports and regular imports?


packages/@ember/-internals/container/lib/registry.ts
```diff
@@ -1,9 +1,10 @@
-import { Factory } from '@ember/-internals/owner';
+import type { Factory } from '@ember/-internals/owner';
 import { dictionary, intern } from '@ember/-internals/utils';
 import { assert, deprecate } from '@ember/debug';
-import { set } from '@ember/object';
+import type { set } from '@ember/object';
 import { DEBUG } from '@glimmer/env';
-import Container, { ContainerOptions, LazyInjection } from './container';
+import type { ContainerOptions, LazyInjection } from './container';
+import Container from './container';
```
chriskrycho, line 7: This is annoying: the rule doesn't seem to understand that these can be intermixed freely now. 🤔 

([playground](https://www.typescriptlang.org/play?#code/PTAEAEBcGcFoEsDmA7A9gJwKalrUArAV2klAAtMtRJVQBbAQwGttILQAjTAG1QHdQfDEwBQ8OgAcMpAORQAnhMzRgmOl3QB9Tag75MAY0gyA3CLGTpoAKLrKAeT2HIAGlABvaouwATTADMGQm5SBmgbewAVb1AAX1B-dFQ6UDk1DWBdfSNTCyl0Uk9MAA8lI2ile384hKSUmRKyyFhIb1yRRucKzCqAHlsNR2zIAD4ACgBKADoaawBHQgZubr7rKO9xibMgA))

packages/@ember/-internals/extension-support/lib/container_debug_adapter.ts
```diff
@@ -1,13 +1,9 @@
...
-  A as emberA,
-  typeOf,
-  Namespace,
-  Object as EmberObject,
-  NativeArray,
-} from '@ember/-internals/runtime';
-import { getOwner, Owner } from '@ember/-internals/owner';
-import { Resolver } from '@ember/-internals/container/lib/registry';
+import type { NativeArray } from '@ember/-internals/runtime';
+import { A as emberA, typeOf, Namespace, Object as EmberObject } from '@ember/-internals/runtime';
```
chriskrycho, line 3: Same issue here. This should actually be:

```suggestion
import {
  A as emberA,
  typeOf,
  Namespace,
  Object as EmberObject,
  type NativeArray,
} from '@ember/-internals/runtime';
```

🤔 I think I’m probably fine with taking it even so, but it’s quite annoying.

Do you mind testing that the lint rule minimally doesn't *reject* this form, even if it doesn't auto-fix *to* it?

chriskrycho (changes requested):
This looks good overall – have one thing I’d like to check before merging! Let me know what the answer is on that, and we’ll proceed if it’s 👍🏼

packages/@ember/-internals/extension-support/lib/container_debug_adapter.ts
snewcomer: I does reject it and auto fix doesn't fix it.

Screen Shot 2022-05-10 at 8 13 38 AM
.

chriskrycho (approved):

PR #20089, opened 2022-05-12
Convert a few more files to TS

wagenet:


packages/@ember/object/index.ts
```diff
@@ -130,18 +131,43 @@ export {
...
+interface HasActions {
+  actions: Record<string | symbol, unknown>;
+}
+
+function setupAction(
+  target: object,
+  key: string | symbol,
+  actionFn: Function
+): TypedPropertyDescriptor<unknown> {
+  // SAFETY: Accessing an unknown property on a non-null value isn't unsafe.
```
chriskrycho, line 157: I think we don't need the comment now that it has the guard!

packages/@ember/object/index.ts
```diff
@@ -130,18 +131,43 @@ export {
...
+  if (hasProto(target)) {
     target.constructor.proto();
   }
 
+  let targetWithActions = target as HasActions;
```
chriskrycho, line 162: By contrast, should there be a guard here? Even just `assert()`-ing that `'actions' in target && typeof (target as HasActions).actions === 'object'` or something?

packages/@ember/object/index.ts
```diff
@@ -164,13 +190,27 @@ function setupAction(target, key, actionFn) {
...
-    let decorator = function (target, key, desc, meta, isClassicDecorator) {
+export function action(
+  target: ElementDescriptor[0],
+  key: ElementDescriptor[1],
+  desc: ElementDescriptor[2]
+): PropertyDescriptor;
+export function action(desc: PropertyDescriptor): ExtendedMethodDecorator;
+export function action(
+  ...args: ElementDescriptor | [PropertyDescriptor]
+): PropertyDescriptor | ExtendedMethodDecorator {
```
chriskrycho, line 201: Oh wow. I… want this thing deprecated Soon™.

packages/@ember/object/index.ts
```diff
@@ -189,14 +229,18 @@ export function action(target, key, desc) {
...
+  let [target, key, desc] = args;
+
+  actionFn = desc?.value;
 
   assert(
     'The @action decorator must be applied to methods when used in native classes',
     typeof actionFn === 'function'
   );
 
+  // SAFETY: TS types are weird with decorators. This should work.
```
chriskrycho, line 241: "should work" 😆 😂 🤣 

👍🏼 though

packages/@ember/template-compilation/index.ts
```diff
@@ -2,7 +2,7 @@ import { DEBUG } from '@glimmer/env';
 
 export { compile as compileTemplate } from 'ember-template-compiler';
 
-export let precompileTemplate;
+export let precompileTemplate: () => never | undefined;
```
chriskrycho, line 5: `never | undefined` is just `undefined`, since `never` is a subtype of *all* types. In this position it's the inverse of writing `any | undefined`, which would unify to `any` since `any` is the *super*-type of all types.

```suggestion
export let precompileTemplate: () => undefined;
```

I *do* wish TS had a way to annotate that something would throw under certain conditions, but alas: no.

chriskrycho (changes requested):

packages/@ember/object/index.ts
wagenet: Correct!

packages/@ember/object/index.ts
wagenet: Well not exactly, since we immediately set actions if it is not set, but I admit that this is kind of awkward.

packages/@ember/template-compilation/index.ts
wagenet: This is wrong, it should be: `export let precompileTemplate: (() => never) | undefined`

chriskrycho: Ahhh, 👍🏼

chriskrycho: Well… actually… in that case you're saying that *no* actual invocation of `precompileTemplate` *ever* returns?

chriskrycho: Ah, so *locally* that is true, but it seems like in the *other* places this is used that'd be a bit weird? I don't actually understand what the intent of this is.

PR #20091, opened 2022-05-12
Fix doc code block languages

wagenet:


locks (changes requested):
This breaks our API documentation rendering. What's the motivation for the change?

wagenet:
@locks oops, I didn't realize that. This is definitely a non-standard way of annotating code blocks so I was trying to bring us more in line with the standard way of doing things in preparation for generating more documentation with TS. However, I think we'll have to keep two parallel versions in the near term, so I guess I'll just change this in the TS specific blocks.

PR #20092, opened 2022-05-17
Internals Refactor

wagenet:
This consolidates a lot of stuff from @ember/-internals into its relevant packages. Not only is this better organizationally, but it puts us in a better place for correctly building type exports as it reduces the number of potential import locations for the same types.


packages/@ember/-internals/runtime/lib/system/object_proxy.ts
```diff
@@ -1,4 +1,4 @@
-import FrameworkObject from './object';
+import { FrameworkObject } from '@ember/object';
```
wagenet, line 1: This is actually a code change since previous this was really `EmberObject` not `FrameworkObject. However, I believe it is what was intended.

packages/@ember/-internals/runtime/tests/system/object_proxy_test.js
```diff
@@ -103,6 +103,7 @@ moduleFor(
       if (DEBUG) {
         let proxy = ObjectProxy.extend({
           init() {
+            this._super();
```
wagenet, line 106: Because ObjectProxy now correctly extends FrameworkObject, there's a check to ensure `this._super` is called.

packages/ember/index.js
```diff
@@ -71,12 +65,15 @@ import EmberError from '@ember/error';
...
 import { getOwner, setOwner } from '@ember/-internals/owner';
+import EmberArray, { A, NativeArray, isArray } from '@ember/array';
+import MutableArray from '@ember/array/mutable';
 import Application, { onLoad, runLoadHooks } from '@ember/application';
 import ApplicationInstance from '@ember/application/instance';
 import Namespace from '@ember/application/namespace';
 import Engine from '@ember/engine';
 import EngineInstance from '@ember/engine/instance';
 import Enumerable from '@ember/enumerable';
+import MutableEnumerable from '@ember/enumerable/mutable';
```
wagenet, line 76: This was moved to @ember/enumerable to help avoid issues caused by circular imports.

packages/@ember/-internals/runtime/index.ts
```diff
@@ -2,20 +2,12 @@ export { default as RegistryProxyMixin } from './lib/mixins/registry_proxy';
...
-  default as Array,
-  NativeArray,
-  A,
-  MutableArray,
-  removeAt,
-  uniqBy,
-  isArray,
-} from './lib/mixins/array';
 export { default as Comparable } from './lib/mixins/comparable';
+export { NativeArray } from '@ember/array';
```
wagenet, line 6: We re-export here since there are tests that want this to exist in `@ember/-internals`. These should be revisited in the future.

packages/@ember/-internals/runtime/index.ts
```diff
@@ -2,20 +2,12 @@ export { default as RegistryProxyMixin } from './lib/mixins/registry_proxy';
...
 export { default as ArrayProxy } from './lib/system/array_proxy';
 export { default as ActionHandler } from './lib/mixins/action_handler';
 export { default as _ProxyMixin, contentFor as _contentFor } from './lib/mixins/-proxy';
-export { default as MutableEnumerable } from './lib/mixins/mutable_enumerable';
+export { default as MutableEnumerable } from '@ember/enumerable/mutable';
```
wagenet, line 10: We re-export here since there are tests that want this to exist in `@ember/-internals`. These should be revisited in the future.

sandstrom:
Out of curiosity, is there anything holding this PR back?

wagenet:
@sandstrom, just waiting on proper review!

sandstrom:
@wagenet Alright, great! 😄

packages/@ember/object/tests/reopen_test.js
```diff
@@ -1,12 +1,12 @@
-import { get } from '@ember/-internals/metal';
-import EmberObject from '../../../lib/system/object';
+import { get } from '@ember/object';
+import CoreObject from '@ember/object/core';
```
chriskrycho, line 2: Hmm, are you *sure* about this change? Seems like we should probably prefer to keep it the same for *this* change, and then if we think this is correct, update in a follow-on PR?

packages/@ember/routing/auto-location.ts
```diff
@@ -1 +1,375 @@
...
+  getHash,
+  getPath,
+  getQuery,
+  replacePath,
+  supportsHashChange,
+  supportsHistory,
+} from './lib/location-utils';
+
+/**
+@module @ember/routing
```
chriskrycho, line 18:
```suggestion
@module @ember/routing/auto-location
```

packages/@ember/routing/hash-location.ts
```diff
@@ -1 +1,176 @@
-export { HashLocation as default } from '@ember/-internals/routing';
+import EmberObject, { set } from '@ember/object';
+import { bind } from '@ember/runloop';
+import type { ILocation as EmberLocation, UpdateCallback } from '@ember/routing/location';
+import { getHash } from './lib/location-utils';
+
+/**
+@module @ember/routing
```
chriskrycho, line 7:
```suggestion
@module @ember/routing/hash-location
```

packages/@ember/routing/history-location.ts
```diff
@@ -1 +1,281 @@
-export { HistoryLocation as default } from '@ember/-internals/routing';
+import EmberObject, { set } from '@ember/object';
+import type { ILocation as EmberLocation, UpdateCallback } from '@ember/routing/location';
+import { getHash } from './lib/location-utils';
+
+/**
+@module @ember/routing
```
chriskrycho, line 6:
```suggestion
@module @ember/routing/history-location
```

packages/@ember/routing/history-location.ts
```diff
@@ -1 +1,281 @@
...
+*/
+
+let popstateFired = false;
+
+function _uuid() {
```
chriskrycho, line 11: Unrelated to this PR: we should seriously consider using [`Crypto.randomUUID()`](http://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) for this… in a few years. Safari only just added support in 15.4. :weary:

packages/@ember/routing/location.ts
```diff
@@ -1 +1,130 @@
...
+  onUpdateURL(callback: UpdateCallback): void;
+  formatURL(url: string): string;
+  detect?(): void;
+  initState?(): void;
+  destroy(): void;
+}
+
+export type UpdateCallback = (url: string) => void;
+/**
+@module @ember/routing
```
chriskrycho, line 18:
```suggestion
@module @ember/routing/location
```

packages/@ember/routing/none-location.ts
```diff
@@ -1 +1,135 @@
-export { NoneLocation as default } from '@ember/-internals/routing';
+import EmberObject, { set } from '@ember/object';
+import { assert } from '@ember/debug';
+import type { ILocation as EmberLocation, UpdateCallback } from '@ember/routing/location';
+
+/**
+@module @ember/routing
```
chriskrycho, line 6:
```suggestion
@module @ember/routing/none-location
```

packages/@ember/routing/route.ts
```diff
@@ -1 +1,2567 @@
...
+};
+
+type MaybeParameters<T> = T extends AnyFn ? Parameters<T> : unknown[];
+type MaybeReturnType<T> = T extends AnyFn ? ReturnType<T> : unknown;
+
+export const ROUTE_CONNECTIONS = new WeakMap();
+const RENDER = Symbol('render');
+
+/**
+@module @ember/routing
```
chriskrycho, line 73:
```suggestion
@module @ember/routing/route
```

packages/@ember/routing/route.ts
```diff
@@ -1 +1,2567 @@
...
+      this._environment = owner.lookup('-environment:main');
+    }
+  }
+
+  // Implement Evented
+  declare on: (name: string, method: ((...args: any[]) => void) | string) => this;
+  declare one: (name: string, method: string | ((...args: any[]) => void)) => this;
+  declare trigger: (name: string, ...args: any[]) => any;
+  declare off: (name: string, method: string | ((...args: any[]) => void)) => this;
+  declare has: (name: string) => boolean;
```
chriskrycho, line 300: For another PR, but I saw it and wanted to call it out here—since this is using `declare`, maybe we should tackle it via an `interface` merge instead?

packages/@ember/routing/router-service.ts
```diff
@@ -1,5 +1,744 @@
-export {
-  RouterService as default,
-  RouteInfo,
-  RouteInfoWithAttributes,
-} from '@ember/-internals/routing';
+import { getOwner } from '@ember/-internals/owner';
```
chriskrycho, line 1: Should this have a `@module` annotation for docs?

packages/@ember/routing/router.ts
```diff
@@ -1 +1,1933 @@
...
+  TransitionState,
+} from 'router_js';
+import Router, { logAbort, STATE_SYMBOL } from 'router_js';
+import type { Timer } from 'backburner';
+import EngineInstance from '@ember/engine/instance';
+import type { QueryParams } from 'route-recognizer';
+import type { AnyFn, MethodNamesOf, OmitFirst } from '@ember/-internals/utils/types';
+
+/**
+@module @ember/routing
```
chriskrycho, line 51:
```suggestion
@module @ember/routing/router
```

packages/@ember/routing/router.ts
```diff
@@ -1 +1,1933 @@
...
+import { RouterState } from '@ember/routing/internals';
+import {
+  calculateCacheKey,
+  extractRouteArgs,
+  getActiveTargetName,
+  resemblesURL,
+} from './lib/utils';
+import type { RouteArgs, RouteOptions } from './lib/utils';
+import EmberLocation from '@ember/routing/location';
+import type { ILocation as IEmberLocation } from '@ember/routing/location';
```
chriskrycho, line 17: Another case where the imports should be able to be merged on a recent-enough version of TS. And I checked just now: it's actually 4.5, and supports this—

```suggestion
import EmberLocation, type { ILocation as IEmberLocation } from '@ember/routing/location';
```

(Distinctly not a blocker of course.)

packages/@ember/utils/lib/type-of.ts
```diff
@@ -98,7 +98,7 @@ const { toString } = Object.prototype;
   @public
   @static
 */
-export function typeOf(item: unknown): TypeName {
+function typeOf(item: unknown): TypeName {
```
chriskrycho, line 101: Question: any reason to prefer the *separate* `export default` vs. just doing this?

```suggestion
export default function typeOf(item: unknown): TypeName {
```

packages/@ember/utils/lib/type-of.ts
```diff
@@ -123,3 +123,5 @@ export function typeOf(item: unknown): TypeName {
 
   return ret;
 }
+
+export default typeOf;
```
chriskrycho, line 127: And then if not, we can drop this:

```suggestion
```

packages/ember/index.js
```diff
@@ -10,8 +10,7 @@ import { FEATURES, isEnabled } from '@ember/canary-features';
 import * as EmberDebug from '@ember/debug';
```
chriskrycho, line 10: This in particular shows the improvement. 👏🏼 Also I hope to destroy this entire package sometime *soon-ish*. 😂

chriskrycho (changes requested):
I noted a bunch of places where our `@module` declarations in the docs seem to have been simply wrong, and AFAICT this is one part of why the actual generated module docs are substantially less useful than might have been hoped. Between your reorganization and adding those in, I think it might make a nice win for the docs while we're still on YUIDoc (though a colleague is working on doing the work to transition us to TSDoc-powered docs, almost certainly via TypeDoc). That said… it’s possible these are *intentionally* only showing the top-level module, which while deeply unhelpful may be important for some other reason at the moment?

cc. @jenweber @locks: do we *want* to have only (for example) `@ember/routing` rather than `@ember/routing/route` for the `@module` declarations?

---

A related thought: we should consider whether we might want to introduce a `public`/`private` directory, so that you end up with this for input/output structure:

```
.
├── dist
│   ├── -private
│   │   ├── internal-only.d.ts
│   │   └── internal-only.js
│   └── public
│       ├── index.d.ts
│       ├── index.js
│       ├── other.d.ts
│       └── other.js
└── src
    ├── -private
    │   └── internal-only.ts
    └── public
        ├── index.ts
        └── other.ts
```

Once [this Typedoc issue](https://github.com/TypeStrong/typedoc/issues/1937) is resolved, we can theoretically start leveraging package `exports` maps to make it so consumers *cannot* use the `-private` directories (if we so desire), with a very straightforward per-package config:

```json
{
  "exports": {
    ".": "./dist/public/index.js",
    "./package.json": "./package.json",
    "./*": "./dist/public/*.js"
  }
}
```

packages/@ember/-internals/glimmer/tests/integration/components/link-to/routing-angle-test.js
```diff
@@ -5,9 +5,11 @@ import {
   runTask,
 } from 'internal-test-helpers';
 import Controller, { inject as injectController } from '@ember/controller';
-import { A as emberA, RSVP } from '@ember/-internals/runtime';
+import { A as emberA } from '@ember/array';
+import { RSVP } from '@ember/-internals/runtime';
```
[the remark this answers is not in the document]
wagenet: We've got extra stuff in @ember/-internals/runtime/lib/ext/rsvp.ts. Might still be fine to import from RSVP directly here though.

packages/@ember/-internals/runtime/lib/mixins/action_handler.ts
```diff
@@ -2,7 +2,8 @@
 @module ember
 */
 
-import { Mixin, get } from '@ember/-internals/metal';
+import Mixin from '@ember/object/mixin';
+import { get } from '@ember/-internals/metal';
```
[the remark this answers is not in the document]
wagenet: IIRC circularity was the issue, but I'd have to check again.

packages/@ember/-internals/runtime/lib/mixins/target_action_support.ts
```diff
@@ -3,7 +3,8 @@
 */
 
 import { context } from '@ember/-internals/environment';
-import { get, Mixin, computed } from '@ember/-internals/metal';
+import { get, computed } from '@ember/-internals/metal';
```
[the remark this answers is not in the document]
wagenet: Yeah, I'm pretty sure it's not great if `@ember/-internals` is importing from packages that import from it.

packages/@ember/-internals/runtime/tests/array/forEach-test.js
```diff
@@ -1,5 +1,5 @@
 import { guidFor } from '@ember/-internals/utils';
-import { get } from '@ember/-internals/metal';
+import { get } from '@ember/object';
```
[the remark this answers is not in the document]
wagenet: This is a test, so it's probably handled differently.

packages/@ember/application/index.ts
```diff
@@ -1,3 +1,1212 @@
+/**
+@module @ember/application
+*/
+
 export { getOwner, setOwner } from '@ember/-internals/owner';
-export { onLoad, runLoadHooks, _loaded } from './lib/lazy_load';
-export { default } from './lib/application';
+import { dictionary } from '@ember/-internals/utils';
```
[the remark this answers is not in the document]
wagenet: 💯

packages/@ember/application/type-tests/instance.test.ts
```diff
@@ -1,8 +1,8 @@
 import type { Factory, Owner } from '@ember/-internals/owner';
+import type EngineInstance from '@ember/engine/instance';
+import type { BootOptions } from '@ember/engine/instance';
```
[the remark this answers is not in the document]
wagenet: Looks like we're on 4.6. Maybe we should upgrade.

packages/@ember/routing/router.ts
wagenet: Hmm... I'm getting syntax errors and we're on 4.6.

chriskrycho: Yeah, for posterity, the correct syntax is:

```ts
import EmberLocation, { type ILocation as IEmberLocation } from '@ember/routing/location';
```

chriskrycho (approved):
One more tweak, not a blocker!

packages/@ember/application/type-tests/instance.test.ts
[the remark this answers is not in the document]
wagenet: We could do:

```ts
import type { default as EngineInstance,  BootOptions } from '@ember/engine/instance';
```

But I kind of feel like it is better to just have two lines.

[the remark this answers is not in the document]
chriskrycho: Agreed.

chriskrycho (approved):
Let's do it! …maybe starting by fixing CI? 😂

packages/@ember/object/tests/reopen_test.js
wagenet: A few lines down, the tests claim they're for CoreObject, which I think is why I changed it.

packages/@ember/routing/history-location.ts
runspired: EmberData uses randomUUID with an optional polyfill (last 2 safari versions have randomUUID)

The bigger issue is that it is not available to insecure contexts (http)

PR #20094, opened 2022-05-19
[DOC] Document that `history` is the default `locationType` value

bertdeblock:
This was changed to `history` a while ago:
https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/config/environment.js#L8.


MelSumner (approved):
👍

PR #20113, opened 2022-06-11
Application class: convert docs to ES6 syntax

tniezurawski:
I think we could update the docs on the `Application` class to ES6 syntax. Although, I believe the docs are not wrong and you still can `.extend()` but having ES6 syntax would make it more modern and be in line with the current blueprint: https://github.com/ember-cli/ember-cli/blob/v4.4.0/blueprints/app/files/app/app.js

### DIFF
--- a/packages/@ember/application/lib/application.ts
+++ b/packages/@ember/application/lib/application.ts
@@ -43,11 +43,11 @@ import type { SimpleDocument, SimpleElement } from '@simple-dom/interface';
   initialized.
 
   ```app/app.js
-  const App = Application.extend({
+  export default class App extends Application {
     ready() {
       // your code here
     }
-  })
+  }
   ```
 
   Because `Application` ultimately inherits from `Ember.Namespace`, any classes
@@ -84,12 +84,12 @@ import type { SimpleDocument, SimpleElement } from '@simple-dom/interface';
   ```app/app.js
   import Application from '@ember/application';
 
-  let App = Application.extend({
-    customEvents: {
+  export default class App extends Application {
+    customEvents = {
       // add support for the paste event
       paste: 'paste'
     }
-  });
+  }
   ```
 
   To prevent Ember from setting up a listener for a default event,
@@ -99,13 +99,13 @@ import type { SimpleDocument, SimpleElement } from '@simple-dom/interface';
   ```app/app.js
   import Application from '@ember/application';
 
-  let App = Application.extend({
-    customEvents: {
+  export default class App extends Application {
+    customEvents = {
       // prevent listeners for mouseenter/mouseleave events
       mouseenter: null,
       mouseleave: null
     }
-  });
+  }
   ```
 
   By default, the application sets up these event listeners on the document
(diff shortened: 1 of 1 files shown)


tniezurawski:
Could someone help me in understanding why the `CI / Browserstack Tests (Safari, Edge)` test is failing?

chancancode:
I think that test suite is (unfortunately) know to be a bit flaky so it's probably not something your PR caused. Can you rebase your changes against the master branch? You previously targeted the release branch, but that branch get recreated every time a new release comes out (which just happened today). In general it's better to target the master branch for that reason. I tried changing the base in the UI, but that doesn't seem to be sufficient in this case.

tniezurawski:
@chancancode Any chance to get this merged?

chancancode (approved):

chancancode:
Thank you!

> Is this something that should be addressed as well? I see the link is pointing to release and js file:

Ah, thank you for pointing that out. That's a trickier question and I'll defer to @emberjs/learning-team-managers. There probably isn't an easy answer. The link is technically pointing you to the correct place, and in a sense that's the only correct place to point at, in that the content you saw rendered on that page is based on the source code of the release branch (the same page could have been moved/rewritten/deleted on the master branch, to be deployed when the next version is released, for example). It's just that as far as the development workflow of this repository, we typically develop against the master branch and cherry-pick changes down to beta/release as needed/appropriate.

packages/@ember/application/lib/application.ts
```diff
@@ -43,11 +43,11 @@ import type { SimpleDocument, SimpleElement } from '@simple-dom/interface';
...
-  const App = Application.extend({
+  export default class App extends Application {
     ready() {
       // your code here
     }
-  })
+  }
   ```
 
   Because `Application` ultimately inherits from `Ember.Namespace`, any classes
```
chancancode, line 53: @tniezurawski if you are interested in doing more edits for this class, I feel like most of this `Ember.Namespace` and "`Application` as a container that holds the other classes in your application" stuff should be deleted.

It was referring to to [globals resolver](https://github.com/emberjs/rfcs/pull/331) which is deprecated/removed at this point. I thought we had deprecated `Ember.Namespace` as well but apparently we didn't. In any case, it doesn't really do anything useful at this point so we don't need to mention it here.

chancancode: The "Event Delegation" stuff probably should go as well though it requires slightly more thoughts on what to do with the content (straight up delete it, or moving it somewhere else, etc).

Most idiomatic Ember apps uses `{{on}}` which doesn't go through event delegation at all and is equivalent to `addEventListener` directly on the element. The event delegation stuff is still used by classic components (`@ember/components`) so it's not completely gone, per se, but definitely doesn't need to be mentioned so front and center here, and setting it largely has no effect at all for a lot of modern apps.

Maybe we can keep it as a `@property` level documentation, or maybe it's more appropriate to talk about this in details in the docs for `@ember/component`.

cc @emberjs/learning-team-managers

PR #20115, opened 2022-06-11
Remove internal files in `blueprints-js` folder

bertdeblock:
I _think_ this was forgotten over in https://github.com/emberjs/ember.js/pull/19962.
These files are already located [here](https://github.com/emberjs/ember.js/blob/master/blueprints/-addon-import.js) and [here](https://github.com/emberjs/ember.js/blob/master/blueprints/test-framework-detector.js).


kategengler (approved):

kategengler:
Nice catch!

PR #20119, opened 2022-06-13
[DOC] Document `helper` and `modifier` keywords

### LINKED ISSUE #19925: [Documentation] Documentation issue report
Document `helper` and `modifier` keywords.

SergeAstapov:
Closes #19925

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/helpers/helper.ts
+++ b/packages/@ember/-internals/glimmer/lib/helpers/helper.ts
@@ -1,0 +1,41 @@
+/**
+ @module ember
+ */
+
+/**
+ Use the `{{helper}}` helper to create contextual helper so
+ that it can be passed around as first-class values in templates.
+
+ ```handlebars
+ {{#let (helper "join-words" "foo" "bar" separator=" ") as |foo-bar|}}
+
+   {{!-- this is equivalent to invoking `{{join-words "foo" "bar" separator=" "}}` --}}
+   {{foo-bar}}
+
+   {{!-- this will pass the helper itself into the component, instead of invoking it now --}}
+   <MyComponent @helper={{helper foo-bar "baz"}} />
+
+   {{!-- this will yield the helper itself ("contextual helper"), instead of invoking it now --}}
+   {{yield foo-bar}}
+ {{/let}}
+ ```
+
+ ### Arguments
+
+ The `{{helper}}` helper works similarly to the [`{{component}}`](./component?anchor=component) and
+ [`{{modifier}}`](./modifier?anchor=modifier) helper:
+
+ * When passed a string (e.g. `(helper "foo")`) as the first argument,
+   it will produce an opaque, internal "helper definition" object
+   that can be passed around and invoked elsewhere.
+
+ * Any additional positional and/or named arguments (a.k.a. params and hash)
+   will be stored ("curried") inside the definition object, such that, when invoked,
+   these arguments will be passed along to the referenced helper.
+
+
+ @method helper
+ @for Ember.Templates.helpers
+ @public
+ @since 3.27.0
+ */
--- a/packages/@ember/-internals/glimmer/lib/helpers/modifier.ts
+++ b/packages/@ember/-internals/glimmer/lib/helpers/modifier.ts
@@ -1,0 +1,40 @@
+/**
+ @module ember
+ */
+
+/**
+ Use the `{{modifier}}` helper to create contextual modifier so
+ that it can be passed around as first-class values in templates.
+
+ ```handlebars
+ {{#let (modifier "click-outside" click=this.submit) as |on-click-outside|}}
+
+   {{!-- this is equivalent to `<MyComponent {{click-outside click=this.submit}} />` --}}
+   <MyComponent {{on-click-outside}} />
+
+   {{!-- this will pass the modifier itself into the component, instead of invoking it now --}}
+   <MyComponent @modifier={{modifier on-click-outside "extra" "args"}} />
+
+   {{!-- this will yield the modifier itself ("contextual modifier"), instead of invoking it now --}}
+   {{yield on-click-outside}}
+ {{/let}}
+ ```
+
+ ### Arguments
+
+ The `{{modifier}}` helper works similarly to the [`{{component}}`](./component?anchor=component) and
+ [`{{helper}}`](./helper?anchor=helper) helper:
+
+ * When passed a string (e.g. `(modifier "foo")`) as the first argument,
+   it will produce an opaque, internal "modifier definition" object
+   that can be passed around and invoked elsewhere.
+
+ * Any additional positional and/or named arguments (a.k.a. params and hash)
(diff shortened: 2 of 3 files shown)


packages/@ember/-internals/glimmer/lib/helpers/modifier.ts
```diff
@@ -0,0 +1,121 @@
...
+   <div class={{on-click mouseenter=(action "highlight") mouseleave=(action "unhighlight")}} />
+
+   {{!-- runtime error: not a component --}}
+   <on-click />
+ {{/let}}
+ ```
+
+ ### Arguments
+
+ The `{{modifier}}` helper works similarly to the `Component` and `Helper` helper:
```
chancancode, line 107: same comment as above

chancancode (approved):
Broadly 👍🏼 from me.

I think we can quibble on whether the examples copied from the RFCs are actually useful for explaining what the feature is. In the RFC they are more geared towards exhausting most of the possibilities, in a spec-by-examples kind of way. In documentation that may be a bit much and the point kind of gets lost. In other words, we may want to trim it down so that we only have one or two succinct/common use cases that gets the main point across.

That being said, any docs are certainly better than no docs. So if @SergeAstapov wants to do another pass now I think that's great, but otherwise we can also just merge it and let someone else iterate on this another time.

packages/@ember/-internals/glimmer/lib/helpers/helper.ts
```diff
@@ -0,0 +1,129 @@
...
+   <foo-bar />
+
+   {{!-- runtime error: not a modifier --}}
+   <div {{foo-bar}}>
+ {{/let}}
+ ```
+
+ ### Arguments
+
+ The `{{helper}}` helper works similarly to the `Component` and `Modifier` helper:
```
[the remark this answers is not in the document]
SergeAstapov: updated with below snippet which I think should do the trick:

```md
[`{{component}}`](./component?anchor=component)
```

packages/@ember/-internals/glimmer/lib/helpers/modifier.ts
```diff
@@ -0,0 +1,121 @@
+/**
+ @module ember
+ */
+
+/**
+ Use the `{{modifier}}` helper to create contextual modifier so
+ that it can be passed around as first-class values in templates.
+
+ ```handlebars
+ {{#let (modifier "on" click=this.submit) as |on-click|}}
```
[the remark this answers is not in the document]
SergeAstapov: @chancancode got point! I've replaced `on` with `click-outside` as an example

SergeAstapov:
> we may want to trim it down so that we only have one or two succinct/common use cases that gets the main point across.

I actually did that originally but then decided to provide exhaustive list examples.

trimmed down to 3 examples for both `helper` and `modifier` helpers

packages/@ember/-internals/glimmer/lib/helpers/helper.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+/**
+ Use the `{{helper}}` helper to create contextual helper so
+ that it can be passed around as first-class values in templates.
+
+ ```handlebars
+ {{#let (helper "join-words" "foo" "bar" separator=" ") as |foo-bar|}}
+
+   {{foo-bar}}
```
chancancode, line 12:
```suggestion
   {{!-- this is equivalent to invoking `{{join-words "foo" "bar" separator=" "}}` --}}
   {{foo-bar}}
```

packages/@ember/-internals/glimmer/lib/helpers/helper.ts
```diff
@@ -0,0 +1,42 @@
...
+
+   {{!-- these will pass the helper itself into the component, instead of invoking it now --}}
+
+   <MyComponent @helper={{helper foo-bar "baz"}} />
+
+   {{!-- these will yield the helper itself ("contextual helper"), instead of invoking it now --}}
+
+   {{yield foo-bar}}
```
chancancode, line 20:
```suggestion
   {{!-- this will pass the helper itself into the component, instead of invoking it now --}}
   <MyComponent @helper={{helper foo-bar "baz"}} />

   {{!-- this will yield the helper itself ("contextual helper"), instead of invoking it now --}}
   {{yield foo-bar}}
```

packages/@ember/-internals/glimmer/lib/helpers/modifier.ts
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+/**
+ Use the `{{modifier}}` helper to create contextual modifier so
+ that it can be passed around as first-class values in templates.
+
+ ```handlebars
+ {{#let (modifier "click-outside" click=this.submit) as |on-click-outside|}}
+
+   <MyComponent {{on-click-outside}} />
```
chancancode, line 12:
```suggestion
   {{!-- this is equivalent to `<MyComponent {{click-outside click=this.submit}} />` --}}
   <MyComponent {{on-click-outside}} />
```

packages/@ember/-internals/glimmer/lib/helpers/modifier.ts
```diff
@@ -0,0 +1,41 @@
...
+
+   {{!-- these will pass the modifier itself into the component, instead of invoking it now --}}
+
+   <MyComponent @modifier={{modifier on-click-outside "extra" "args"}} />
+
+   {{!-- these will yield the modifier itself ("contextual modifier"), instead of invoking it now --}}
+
+   {{yield on-click-outside}}
```
chancancode, line 20:
```suggestion
   {{!-- this will pass the modifier itself into the component, instead of invoking it now --}}
   <MyComponent @modifier={{modifier on-click-outside "extra" "args"}} />

   {{!-- this will yield the modifier itself ("contextual modifier"), instead of invoking it now --}}
   {{yield on-click-outside}}
```

packages/@ember/-internals/glimmer/lib/helpers/helper.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ @module ember
+ */
+
+/**
+ Use the `{{helper}}` helper to create contextual helper so
+ that it can be passed around as first-class values in templates.
+
+ ```handlebars
+ {{#let (helper "join-words" "foo" "bar" separator=" ") as |foo-bar|}}
```
chancancode, line 10: One more thing (same for the other snippet): `foo-bar` is it actually legal to have a dash in the local variable (block param) name? I probably should know this but I don't 🙃

chancancode: According to AST explorer it seems to be fine so that's good enough for me

chancancode (approved):

yads:
A use case for this is to conditionally add a click handler to an element e.g.

```
<div role="{{if this.doButton 'button'}}" {{(if this.doButton (modifier "on" "click" this.handleClick))}}>
  Click Me
</div>
```

PR #20125, opened 2022-06-19
[CLEANUP beta] Replace deprecated substr() method with substring() method

kirtan-desai:
Replaced all implementations of deprecated String.prototype.substr() method with String.prototype.substring() method.

### DIFF
--- a/packages/@ember/-internals/routing/lib/services/router.ts
+++ b/packages/@ember/-internals/routing/lib/services/router.ts
@@ -18,7 +18,7 @@ function cleanURL(url: string, rootURL: string) {
     return url;
   }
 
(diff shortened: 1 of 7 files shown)


packages/@ember/-internals/routing/lib/services/router.ts
```diff
@@ -18,7 +18,7 @@ function cleanURL(url: string, rootURL: string) {
     return url;
   }
 
-  return url.substr(rootURL.length, url.length);
+  return url.substring(rootURL.length);
```
chancancode, line 21: This seems incorrect

chancancode (approved):

PR #20126, opened 2022-06-20
Replacing Firefox detection using InstallTrigger resolves #20098

### LINKED ISSUE #20098: [Bug] Firefox 100 `InstallTrigger` deprecation warning
### 🐞 Describe the Bug
Starting from Firefox 100 a deprecation warning is shown in the browser console when an Ember app is opened. The `InstallTrigger` global is deprecated and Ember seems to still use it for [browser detection](https://github.com/emberjs/ember.js/blob/55ffe6326f11efcaeb278cdf7e0f86543daa9f04/packages/%40ember/-internals/browser-environment/index.ts#L15).

More information about the deprecation can be found in the [bugzilla issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1754441).

It doesn't seem like they give a clear way forward for the UA Agent detection use-case. I also haven't taken a detailed look at the code yet but on it seems that these browser constants are used for working around older browser behaviors which might no longer be relevant with the new browser support policy?

### 🔬 Minimal Reproduction

1. Open any [Ember application](https://emberjs.com/) in Firefox 100
2. Open the dev tools console pane and notice the deprecation warning

### 🌍 Environment

- Ember: -
- Node.js/npm: -
- OS: -
- Browser: Firefox 100

### ➕ Additional Context
https://bugzilla.mozilla.org/show_bug.cgi?id=1754441

kirtan-desai:
Deprecated InstallTrigger was used to detect Firefox. Replaced it with regex based UA detection using suggestion from here https://bugzilla.mozilla.org/show_bug.cgi?id=1754441#c30

### DIFF
--- a/packages/@ember/-internals/browser-environment/index.ts
+++ b/packages/@ember/-internals/browser-environment/index.ts
@@ -1,6 +1,5 @@
 import hasDom from './lib/has-dom';
 
-declare const InstallTrigger: unknown;
 declare const chrome: unknown;
 declare const opera: unknown;
 declare const MSInputMethodContext: unknown;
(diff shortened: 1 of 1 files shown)


chancancode:
Thank you!!

PR #20127, opened 2022-06-20
[bug] TypeWatcher invoking itself during onChange

### LINKED ISSUE #1875: Ember Inspector causing "Assertion Failed: Illegal set of identifier"
**Describe the bug**

After experiencing random cases of the `Assertion Failed: Illegal set of identifier` error I looked for any information and ended up on [this StackOverflow question](https://stackoverflow.com/questions/69262963/emberjs-data-assert-in-setrecorddatafor-illegal-set-of-identifier) and more specifically [this comment by HappyDude](https://stackoverflow.com/questions/69262963/emberjs-data-assert-in-setrecorddatafor-illegal-set-of-identifier#comment124776510_69262963) stating that this error can sometimes be caused by Ember Inspector.

I disabled Ember Inspector and it fixed the issue.

One element I can add is that the model causing the issue appeared twice in the inspector Data tab.

**To Reproduce**

I'm very sad and sorry not to be able to give any useful reproduction steps as it happens randomly.

**Environment**

Ember: 3.28.8
Ember Data: 3.28.6
Ember Inspector: 4.5.11
Firefox: 95.0.2 (64-bit) on a Mac

---

Taking the opportunity to thank all the team working on this otherwise wonderful tool that Ember Inspector is ❤️

patricklx:
When using ember inspector and loading data through api it can cause a tracking assertion for [] and also ["Assertion Failed: Illegal set of identifier"](https://github.com/emberjs/ember-inspector/issues/1875#)

this happens when the type is first found and has new data. the type watcher will call onChange immediately, which calls this.wrapModelType  then this.getRecords and then ember data does an internal sync again which causes this issue.

similar issue here: https://github.com/emberjs/data/issues/8006

also fixes https://github.com/emberjs/ember-inspector/issues/1875

### DIFF
--- a/packages/@ember/debug/data-adapter.ts
+++ b/packages/@ember/debug/data-adapter.ts
@@ -1,6 +1,6 @@
 import type { Owner } from '@ember/-internals/owner';
 import { getOwner } from '@ember/-internals/owner';
-import { _backburner } from '@ember/runloop';
+import { _backburner, next } from '@ember/runloop';
 import { get } from '@ember/object';
 import { dasherize } from '@ember/string';
 import Namespace from '@ember/application/namespace';
@@ -141,7 +141,7 @@ class TypeWatcher {
       consumeTag(tagFor(records, '[]'));
 
       if (hasBeenAccessed === true) {
-        onChange();
+        next(onChange);
       } else {
         hasBeenAccessed = true;
       }


patricklx:
@snewcomer or @wagenet? You were the last 2 editing this file. Any chance to have a look at this?

packages/@ember/debug/data-adapter.ts
```diff
@@ -141,7 +141,7 @@ class TypeWatcher {
       consumeTag(tagFor(records, '[]'));
 
       if (hasBeenAccessed === true) {
-        onChange();
+        next(() => onChange());
```
bertdeblock, line 144: Nitpick, but I think you can also do `next(onChange);` and avoid creating a new function?

patricklx: sure, done

patricklx:
would also be nice to backport to older releases

wagenet (dismissed):
I think this makes sense. Looks low-risk overall.

wagenet (approved):
After consultation with @runspired I'm approving this.

PR #20129, opened 2022-06-23
[BUGFIX LTS] Fix mixin merging with getters

chriskrycho:
Our handling of getters and setters works correctly for cases where we have only classic classes, or only native classes, or native classes extending classic classes. In some cases, however, it fails when we have "zebra-striping" of class types, with a classic class extending a native class which extends a (stack of) classic class(es). We correctly handle native getters in mixin and POJOs passed to `.extend()`, but—quite reasonably—do *not* handle getters on native classes when calling `.extend(SomeNativeClass)`. The result is that those getters are not decorated the way they are in mixins, so mixin merging ends up invoking the getter rather than defining a decorator which returns it.

Fixes #18860

### DIFF
--- a/packages/@ember/-internals/runtime/tests/mixins/accessor_test.js
+++ b/packages/@ember/-internals/runtime/tests/mixins/accessor_test.js
@@ -1,0 +1,35 @@
+import EmberObject from '@ember/object';
+import { moduleFor, RenderingTestCase } from 'internal-test-helpers';
+
+moduleFor(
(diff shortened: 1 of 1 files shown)


kategengler:

chriskrycho:
@kategengler nope, but happily I think @wycats and co. fixed it separately earlier this year. (For posterity: it was left as a draft because I did not have a chance to get anything but the tests implemented when I first threw this up!)

PR #20138, opened 2022-07-06
[DOCS] add examples for yield

nicolechung:
## What changed
Added in more examples for `yield`

https://api.emberjs.com/ember/4.4/classes/Ember.Templates.helpers/methods/yield?anchor=yield

- [x] add yield with hash helper
- [x] add yield with hash and component helper
- [x] add yield for multiple block parameters
- [x] add yield with hash and component helper with attributes and args

## Rationale
Coming from React, yield is a bit of a weird concept. Having more examples of how to use `yield` with `hash` and `component` helpers will help people making components who are coming from another framework and are trying to make more composable components with a small API footprint (more "HTML" like components)

Partly inspired by: https://www.youtube.com/watch?v=uTFxv0kReYk&list=PLRtznLaEWGuf87iG3LZarzyMb24JCX9Pv

- wanted to show that you could use @arguments and attributes
- wanted to show that you could move components around

### DIFF
--- a/packages/@ember/-internals/glimmer/index.ts
+++ b/packages/@ember/-internals/glimmer/index.ts
@@ -142,6 +142,157 @@
     First name: <input type="text" />
   </label>
   ```
+
+  `yield` can also be used with the `hash` helper:
+
+  ```app/templates/application.hbs
+  <DateRanges @value={{@model.date}} |range|>
+    Start date: {{range.start}}
+    End date: {{range.end}}
+  </DateRanges>
+  ```
+
+  ```app/components/date-ranges.hbs
+  <div>
+    {{yield (hash start=@value.start end=@value.end)}}
+  </div>
+  ```
+
+  Result:
+
+  ```html
+  <div>
+    Start Date: July 1st
+    End Date: July 30th
+  </div>
+  ```
+
+  Multiple values can be yielded as block params:
+
+  ```app/templates/application.hbs
+  <Banner @value={{@model}} |title subtitle body|>
+    <h1>{{title}}</h1>
+    <h2>{{subtitle}}</h2>
+    {{body}}
+  </Banner>
+  ```
+
+  ```app/components/Banner.hbs
+  <div>
+    {{yield "Hello title" "hello subtitle" "body text"}}
+  </div>
+  ```
+
+  Result:
+
+  ```html
+  <div>
+    <h1>Hello title</h1>
+    <h2>hello subtitle</h2>
+    body text
+  </div>
+  ```
+
+  However, it is preferred to use the hash helper, as this can prevent breaking changes to your component and also simplify the api for the component.
+
+  Multiple components can be yielded with the `hash` and `component` helper:
+
+  ```app/templates/application.hbs
+  <Banner @value={{@model}} |banner|>
+    <banner.Title>Banner title</banner.Title>
+    <banner.Subtitle>Banner subtitle</banner.Subtitle>
+    <banner.Body>A load of body text</banner.Body>
+  </Banner>
+  ```
+
+  ```app/components/banner.ts
+  import Title from './title';
+  import Subtitle from './subtitle';
+  import Body from './body';
+
+  export default class Banner extends Component {
+    Title = Title;
+    Subtitle = Subtitle;
+    Body = Body;
+  }
+  ```
+
+  ```app/components/banner.hbs
+  <div>
+    {{yield (hash
+      Title=this.Title
+      Subtitle=this.Subtitle
+      Body=(component this.Body defaultArg="some value")
+    )}}
+  </div>
+  ```
+
+  Result:
+
+  ```html
+  <div>
+    <h1>Banner title</h1>
+    <h2>Banner subtitle</h2>
+    A load of body text
+  </div>
+  ```
+
+  A benefit of using this pattern is that the user of the component can change the order the components are displayed.
+
+  ```app/templates/application.hbs
+  <Banner @value={{@model}} |banner|>
+    <banner.Subtitle>Banner subtitle</banner.Subtitle>
+    <banner.Title>Banner title</banner.Title>
+    <banner.Body>A load of body text</banner.Body>
+  </Banner>
+  ```
+
+  Result:
+
+  ```html
+  <div>
+    <h2>Banner subtitle</h2>
+    <h1>Banner title</h1>
+    A load of body text
+  </div>
+  ```
+
+  Another benefit to using `yield` with the `hash` and `component` helper
+  is you can pass attributes and arguments to these components:
+
+  ```app/templates/application.hbs
+  <Banner @value={{@model}} |banner|>
+    <banner.Subtitle class="mb-1">Banner subtitle</banner.Subtitle>
+    <banner.Title @variant="loud">Banner title</banner.Title>
+    <banner.Body>A load of body text</banner.Body>
+  </Banner>
+  ```
+
+  ```app/components/banner/subtitle.hbs
+  {{!-- note the use of ..attributes --}}
+  <div ...attributes>
+    {{yield}}
+  </div>
+  ```
+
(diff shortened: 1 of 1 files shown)


packages/@ember/-internals/glimmer/index.ts
```diff
@@ -142,6 +142,145 @@
...
+    <banner.Title>Banner title</banner.Title>
+    <banner.Subtitle>Banner subtitle</banner.Subtitle>
+    <banner.Body>A load of body text</banner.Body>
+  </Banner>
+  ```
+
+  ```app/components/Banner.hbs
+  <div>
+    {{yield (hash
+       Title=(component 'path/to/title')
```
NullVoxPopuli, line 211: these paths can be a little wonkey -- do we instead want to promote using component values?

```js
import Title from ''./title";
import Subtitle from ''./subtitle";
import Body from ''./body";

export default class Banner extends Component {
  Title = Title;
  Subtitle = Subtitle;
  Body = Body;
}
```
```hbs
<div>
  {{yield (hash
    Title=this.Title
    Subtitle=this.Subtitle
    Body=(component this.Body defaultArg="some value")
  )}}
</div>
```

jenweber (approved):
Thank you! It is awesome to have some more examples of what you can do with yield. This fills a gap in our docs overall.

To review this, I looked over the diff and consider whether this belonged here or in the Guides. I believe it belongs here because more involved examples (than usual) are required in order to understand how to use this feature. I appreciate the briefness of the examples as well, given that these are the API docs and not a tutorial.

Next steps are that someone from Framework core should review and merge. Maybe @locks ?

packages/@ember/-internals/glimmer/index.ts
```diff
@@ -142,6 +142,157 @@
...
+  <div>
+    {{yield (hash start=@value.start end=@value.end)}}
+  </div>
+  ```
+
+  Result:
+
+  ```html
+  <div>
+    Start Date: July 1st
```
bertdeblock, line 165:
```suggestion
    Start date: July 1st
```

Nitpick, but the initial template uses a lowercase `d`. Same for `End date`.

packages/@ember/-internals/glimmer/index.ts
```diff
@@ -142,6 +142,157 @@
...
+
+  ```app/templates/application.hbs
+  <Banner @value={{@model}} as |title subtitle body|>
+    <h1>{{title}}</h1>
+    <h2>{{subtitle}}</h2>
+    {{body}}
+  </Banner>
+  ```
+
+  ```app/components/Banner.hbs
```
bertdeblock, line 180:
```suggestion
  ```app/components/banner.hbs
```

packages/@ember/-internals/glimmer/index.ts
```diff
@@ -142,6 +142,157 @@
...
+
+  ```app/templates/application.hbs
+  <Banner @value={{@model}} as |banner|>
+    <banner.Title>Banner title</banner.Title>
+    <banner.Subtitle>Banner subtitle</banner.Subtitle>
+    <banner.Body>A load of body text</banner.Body>
+  </Banner>
+  ```
+
+  ```app/components/banner.ts
```
bertdeblock, line 208:
```suggestion
  ```app/components/banner.js
```

AFAICT, the guides have only been using `.js`?

packages/@ember/-internals/glimmer/index.ts
```diff
@@ -142,6 +142,157 @@
...
+  <Banner @value={{@model}} as |banner|>
+    <banner.Subtitle class="mb-1">Banner subtitle</banner.Subtitle>
+    <banner.Title @variant="loud">Banner title</banner.Title>
+    <banner.Body>A load of body text</banner.Body>
+  </Banner>
+  ```
+
+  ```app/components/banner/subtitle.hbs
+  {{!-- note the use of ..attributes --}}
+  <div ...attributes>
```
bertdeblock, line 273:
```suggestion
  <h2 ...attributes>
```

As this component has been rendering an `h2` tag in the previous examples.

packages/@ember/-internals/glimmer/index.ts
```diff
@@ -142,6 +142,157 @@
...
+
+  ```app/templates/application.hbs
+  <Banner @value={{@model}} as |banner|>
+    <banner.Subtitle class="mb-1">Banner subtitle</banner.Subtitle>
+    <banner.Title @variant="loud">Banner title</banner.Title>
+    <banner.Body>A load of body text</banner.Body>
+  </Banner>
+  ```
+
+  ```app/components/banner/subtitle.hbs
```
bertdeblock, line 271: The file path doesn't match the import statement in the previous example.

The banner component is located at `app/components/banner.js`.

The import statement for the subtitle component is `import Subtitle from './subtitle';`, which means the subtitle component is actually located at `app/components/subtitle.hbs`.

We could also update the import statements to:
```js
import Title from './banner/title';
import Subtitle from './banner/subtitle';
import Body from './banner/body';
```

packages/@ember/-internals/glimmer/index.ts
```diff
@@ -142,6 +142,157 @@
...
+  {{!-- note the use of ..attributes --}}
+  <div ...attributes>
+    {{yield}}
+  </div>
+  ```
+
+  ```app/components/banner/title.hbs
```
bertdeblock, line 278: Same remark about the file path.

packages/@ember/-internals/glimmer/index.ts
```diff
@@ -142,6 +142,157 @@
...
+  {{#if @variant "loud"}}
```
bertdeblock, line 279:
```suggestion
  {{#if (eq @variant "loud")}}
```

The current example throws a syntax error.

packages/@ember/-internals/glimmer/index.ts
nicolechung: I changed it to the latter to indicate that these components are to be used inside of "banner"

packages/@ember/-internals/glimmer/index.ts
nicolechung: Thanks for catching this!

jelhan: Truth helpers haven't been shipped yet, have they? Using a helper, which is not available by default, might be confusing.

NullVoxPopuli: Let's get them shipped 🙃 

Cc @chancancode 🥳

nicolechung:
Question: I can't seem to `re-request` review or resolve comments? - but if anyone wants to (re) take a look, that would be appreciated!

nicolechung:
Question: Should (I also) do a comparison with `let` versus `yield` here?

let with hash + component helper: have to render the components in the same file
yield with hash + component helper: allows the consumer of the component render how they want with args and attributes

packages/@ember/-internals/glimmer/index.ts
chriskrycho: I *don’t* think we should do that. Among other things, that creates a per-instance binding, which isn’t a lot but also isn’t necessary or great for performance; and in this case it introduces a backing class for *only* that. Using the actual paths is better until we switch to teaching `<template>` and can just show yielding the values normally.

bertdeblock (approved):

jenweber:
Hi Nicole, thanks again for your work on this!

I recommend that the framework team does a final review & merge this, and any additional improvements can be a followup PR.

I discussed the issue of Guides vs API docs as the home for this with locks. After that conversation, I think eventually it would be good to show examples like this in a cookbook or “advanced” section of the guides. If anyone wants to give that a try, in addition to this PR moving forward, that’s welcome! We have room for more commentary there.

mixonic (approved):

mixonic:
Thank you! This seems reasonable to me.

PR #20146, opened 2022-07-27
[DOC] Enhancing the cached decorator documentation

SolPier:
Just a bit of cleanup (I'm the previous author of this part of the documentation):
- some back-ticks were still needed to isolate the code in the plain text sections.
- I reformulated a few sentences as some words/expressions were redundant or needed clarification.
- changed the `newValue` to `nextValue` since the website syntax highlighting was misinterpreting it as a keyword (cf https://api.emberjs.com/ember/4.5/functions/@glimmer%2Ftracking/cached).

I'm happy to be able to set it right, I was a bit surprised (and ashamed) to see the mistakes when the text went online on the website :sweat_smile: .

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/glimmer-tracking-docs.ts
+++ b/packages/@ember/-internals/glimmer/lib/glimmer-tracking-docs.ts
@@ -233,6 +233,11 @@
   the subsequent cache invalidations of the `@cached` properties who were
   using this `trackedProp`.
 
+  As a reminder, do not use this piece of code inside a tracked getter,
+  as the dependency chain could lead to an infinite loop. Mutating an adjacent
+  property from a getter is not a good practice anyway, even with a caching
+  mechanism reducing reruns.
+
   The cost of these edge-guards adds up to the tradoff calculation of using
   a caching strategy, hence requiring a very sensitive and moderate approach
   regarding performance.
--- a/packages/@ember/-internals/glimmer/lib/glimmer-tracking-docs.ts
+++ b/packages/@ember/-internals/glimmer/lib/glimmer-tracking-docs.ts
@@ -161,7 +161,7 @@
   are invalidated. This is useful when a getter is expensive and
   used very often.
 
-  For instance, in this GuestList class, we have the sortedGuests
+  For instance, in this `GuestList` class, we have the `sortedGuests`
   getter that sorts the guests alphabetically:
 
   ```javascript
@@ -176,11 +176,11 @@
     }
   ```
 
-  Every time sortedGuests is accessed, a new array will be created and sorted,
+  Every time `sortedGuests` is accessed, a new array will be created and sorted,
   because JavaScript getters do not cache by default. When the guest list
   is small, like the one in the example, this is not a problem. However, if
   the guest list were to grow very large, it would mean that we would be doing
-  a large amount of work each time we accessed sortedGetters. With @cached,
+  a large amount of work each time we accessed `sortedGuests`. With `@cached`,
   we can cache the value instead:
 
   ```javascript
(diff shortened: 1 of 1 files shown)


packages/@ember/-internals/glimmer/lib/glimmer-tracking-docs.ts
```diff
@@ -196,43 +196,46 @@
...
+  Now the `sortedGuests` getter will be cached based on autotracking.
   It will only rerun and create a new sorted array when the guests tracked
   property is updated.
 
+
   ### Tradeoffs
 
   Overuse is discouraged.
 
   In general, you should avoid using `@cached` unless you have confirmed that
```
NullVoxPopuli, line 208: _while we're here_, should we mention object reference integrity?

the current text talks about primitive values -- but how about object-values?

each time a getter is accessed and an object is returned, those are different objects.

:thinking:

packages/@ember/-internals/glimmer/lib/glimmer-tracking-docs.ts
```diff
@@ -196,43 +196,46 @@
...
-  value did not change.
-  For example updating an integer value from `5` to an other `5`.
+  since tracked properties will always invalidate.
+  For example updating an integer value from `5` to an other `5` will trigger
+  a rerun of the cached properties building from this integer.
 
   Avoiding a cache invalidation in this case is not something that can
   be achieved on the `@cached` decorator itself, but rather when updating
-  the underlying values, by applying a diff checking mecanism:
+  the underlying tracked values, by applying some diff checking mecanisms:
```
NullVoxPopuli, line 224:
```suggestion
  the underlying tracked values, by applying some diff checking mechanisms:
```

packages/@ember/-internals/glimmer/lib/glimmer-tracking-docs.ts
```diff
@@ -196,43 +196,46 @@
...
 
   ```javascript
-  if (newValue !== this.trackedProp) {
-    this.trackedProp = newValue;
+  if (nextValue !== this.trackedProp) {
+    this.trackedProp = nextValue;
```
NullVoxPopuli, line 228: a caveat may need to be mentioned here -- this can cause the infinite re-validation assertion message if this code exists in a getter (you can't modify a "tracked prop used during render" during render)

packages/@ember/-internals/glimmer/lib/glimmer-tracking-docs.ts
SolPier: True, but I can't think about a use case that we could help by documenting a solution / explanation.
Do you have a example at hand ?
All I'm coming up with is related to 'tracking' only, with no specifics things about the 'caching' parts of it.

packages/@ember/-internals/glimmer/lib/glimmer-tracking-docs.ts
SolPier: Done !
It seemed like a weird move to set properties from a getter but considering it would be from a cached one its not impossible to build some spaghetti code/logic on top of it.
The end result would scream to be put into a resource, so I see where you got the idea for the caveat from 😉.
Once resource/use becomes available, I will be amending this file to reference the new solution for the more complex needs.

packages/@ember/-internals/glimmer/lib/glimmer-tracking-docs.ts
```diff
@@ -196,43 +196,51 @@
...
+
+  As a reminder, do not use this piece of code inside a tracked getter,
+  as the dependency chain could lead to an infinite loop. Mutating an adjacent
+  property from a getter is not a good practice anyway, even with a caching
+  mechanism reducing reruns.
 
-  The cost of these edge-guards adds up to the tradoff calculation of using
-  a caching strategy, hence requiring a very sensitive approach regarding
-  performance.
+  The cost of these edge-guards adds up to the trad-off calculation of using
```
geneukum, line 241: I think that we might want trade-off or tradeoff here instead of trad-off.

mixonic (approved):
Wonderful, thank you @SolPier

PR #20151, opened 2022-08-03
feat: add isExternal argument to simplify ember-engines

BobrImperator:
- Adds isExternal argument to LinkTo component 
  - based on the flag it disables assertions that are causing Routeless engines to fail, ember-engines internally overrides them.
  - based on the flag, it changes the behavior of `namespaceRoute` method, so it uses the provided route directly without changing it further - in that case ember-engines is responsible for passing correctly resolved routes.


SergeAstapov:
@BobrImperator FYI this contradicts the long term and more stable solution proposed in https://github.com/emberjs/rfcs/pull/819

kategengler:
Closing since it contradicts the RFC

PR #20158, opened 2022-08-18
[DOC] `renderSettled` docs were referring to `rerender`

cvx:
I believe this was incorrect?

Alternatively that whole sentence could go.


rwjblue (approved):
Thank you!

PR #20164, opened 2022-08-25
[internal] improve `Cache` types and implementation

chriskrycho:
I happened to look at this while working on publishing preview types, and noticed that the types here were using a non-null assertion, and switched it to an explicit cast with a `// SAFETY: ...` comment.

Next, use the optional parameter initialization shorthand to make the constructor more clear (a constructor which defines properties *and* has a body is extremely confusing to contributors, including me!) and avoid doing something in the body which the language can do for us, and eliminate an `any` there by defining the type correctly as a `Map`.

Finally, switch one use of `any` to `unknown`, in one of the usage sites for `Cache`.

### DIFF
--- a/packages/@ember/-internals/utils/lib/cache.ts
+++ b/packages/@ember/-internals/utils/lib/cache.ts
@@ -3,15 +3,17 @@ export default class Cache<T, V> {
   public misses = 0;
   public hits = 0;
 
-  constructor(private limit: number, private func: (obj: T) => V, private store?: any) {
-    this.store = store || new Map();
-  }
+  constructor(
+    private limit: number,
+    private func: (obj: T) => V,
+    private store: Map<T, V> = new Map()
+  ) {}
 
   get(key: T): V {
     if (this.store.has(key)) {
       this.hits++;
-
-      return this.store.get(key);
+      // SAFETY: we know the value is present because `.has(key)` was `true`.
+      return this.store.get(key) as V;
     } else {
       this.misses++;
       return this.set(key, this.func(key));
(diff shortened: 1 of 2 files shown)


packages/@ember/-internals/utils/lib/cache.ts
```diff
@@ -3,15 +3,17 @@ export default class Cache<T, V> {
   public misses = 0;
   public hits = 0;
 
-  constructor(private limit: number, private func: (obj: T) => V, private store?: any) {
-    this.store = store || new Map();
-  }
+  constructor(
+    private limit: number,
+    private func: (obj: T) => V,
+    private store: Map<T, V> = new Map()
```
chriskrycho, line 9: It appears that this initialization form (vs. `|| = new Map()`) is not passing the tests – I will check on why tomorrow!

packages/@ember/-internals/utils/lib/cache.ts
```diff
@@ -3,15 +3,17 @@ export default class Cache<T, V> {
...
+  ) {}
 
   get(key: T): V {
-    if (this.store.has(key)) {
+    let value = this.store.get(key);
+    if (value) {
```
wagenet, line 14: This assumes we never cache a falsy value. I assume you've already verified that.

chriskrycho: Ah, that's interesting. 🤔 I'll double check—and that may actually be the cause of the failure, and the reason we were "double"-checking before.

chriskrycho: That was indeed the thing. Resolved by reverting that behavior. «sigh»

wagenet (approved):

packages/@ember/-internals/utils/lib/cache.ts
chriskrycho: It was actually the possibility of nullish values in the map.

Issue #20165, opened 2022-08-26
Add public import for `uniqueId` helper


chriskrycho:
This landed as available via string resolution at `unique-id` but did not come with a public import, meaning it doesn't currently work in `<template>`/strict mode. Per discussion at the Framework team meeting a few weeks ago, it should be exported from `@ember/helper`, the same as the other public helpers which are not built-ins (`hash`, `array`, `concat`, `get`, and `fn` etc.).

It needs two things to be public:

- [ ] #20171
- [ ] A note in the blog post for that release (which should be 4.8).

Additionally, we'll want to update ecosystem support for it:

- [x] add it to [the list of allowed built-ins](https://github.com/ember-template-lint/ember-template-lint/blob/master/lib/rules/no-curly-component-invocation.js#L14) for the `no-curly-component-invocation` rule in `ember-template-lint`
    - ember-template-lint/ember-template-lint#2592
    - ember-template-lint/ember-template-lint#2593

geneukum:
I should be able to take a look at this within a day or two. Should we also add it to the [list of 'allowed' built-ins](https://github.com/ember-template-lint/ember-template-lint/blob/master/lib/rules/no-curly-component-invocation.js#L14) for the no-curly-component-invocation ember-template-lint rule?

chriskrycho:
@geneukum good call. Let me add that to the list of tasks – and also, a colleague mentioned to me late Friday that he’s also interested in helping; let's coordinate efforts in [#dev-ember-js on Discord](https://discord.com/channels/480462759797063690/485447409296736276)!

geneukum:
I've opened up the issue: https://github.com/ember-template-lint/ember-template-lint/issues/2592 and PR: https://github.com/ember-template-lint/ember-template-lint/pull/2593 to hopefully get the ember-template-lint part of this handled.

TechieQian:
That colleague is me! That discord channel link doesn't seem to work.

oscareng:
does this issue still need fixing?

PR #20171, opened 2022-08-29
Create public import for `uniqueId` helper

TechieQian:
Adds public import for `uniq-id` 

https://github.com/emberjs/ember.js/issues/20165


packages/@ember/helper/index.ts
```diff
@@ -314,3 +314,4 @@
 
 export { setHelperManager, helperCapabilities as capabilities } from '@glimmer/manager';
 export { invokeHelper, hash, array, concat, get, fn } from '@glimmer/runtime';
+export { default as uniqId} from "../-internals/glimmer/lib/helpers/unique-id";
```
bertdeblock, line 317:
```suggestion
export { default as uniqueId } from '../-internals/glimmer/lib/helpers/unique-id';
```

bertdeblock (commented):
I think we also need to add an entry [here](https://github.com/emberjs/ember.js/blob/master/packages/ember/tests/reexports_test.js#L163-L171)?

packages/@ember/helper/index.ts
TechieQian: Acknowledged! Also adding a test and putting this in draft for now!

packages/@ember/-internals/glimmer/lib/helpers/unique-id.ts
```diff
@@ -41,7 +41,7 @@ export default internalHelper((): Reference<string> => {
 // implementation details into an intimate API. It also ensures that the UUID
 // always starts with a letter, to avoid creating invalid IDs with a numeric
 // digit at the start.
-function uniqueId() {
+export function uniqueId() {
```
TechieQian, line 44: I am unsure about the defaulted export of this file, which returns a type `HelperDefinitionState`, so I also exported the actual function here.

packages/@ember/helper/index.ts
```diff
@@ -314,3 +314,4 @@
 
 export { setHelperManager, helperCapabilities as capabilities } from '@glimmer/manager';
 export { invokeHelper, hash, array, concat, get, fn } from '@glimmer/runtime';
+export { uniqueId } from '@ember/-internals/glimmer';
```
TechieQian, line 317: It seems like module resolution doesn't let me re-export this from `@ember/-internals/glimmer/lib/helpers.../` directly. So first I had to export it from the `-internals/glimmer` index first.

packages/@ember/-internals/glimmer/tests/integration/helpers/unique-id-test.js
```diff
@@ -1,8 +1,34 @@
...
+  runTask,
+} from 'internal-test-helpers';
 import { setProperties } from '@ember/object';
+import { uniqueId } from '@ember/helper';
 import { EMBER_UNIQUE_ID_HELPER } from '@ember/canary-features';
 
 if (EMBER_UNIQUE_ID_HELPER) {
+  moduleFor(
+    'Helpers test: {{unique-id}} JS',
+    class JSRenderingTest extends AbstractTestCase {
```
TechieQian, line 15: I am not sure if the extends AbstractTestCase is required here so semi blindly following patterns here.

packages/@ember/-internals/glimmer/lib/helpers/unique-id.ts
chriskrycho: I *believe* this is correct, but we actually don't have *literally any* prior art here as far as I can tell—this is the first time we've really needed to think about helpers-which-are-functions-and-should-be-usable-in-JS! Accordingly, I'm going to ask @wycats (who implemented) or @rwjblue or @chancancode (who have more context than I do) to confirm. I *believe* we're doing the `internalHelper()` dance here with the default export to make sure that this works as a helper when invoked as `unique-id`. However, that export will, as far as I can tell, *not* be usable in JS directly, but *will* be usable in a template.

But I think we *do* want `uniqueId` to be usable in JS contexts exactly like in HBS contexts, and therefore I think what you've done here is correct.

packages/@ember/-internals/glimmer/tests/integration/helpers/unique-id-test.js
chriskrycho: This seems 👍🏼

packages/ember/index.js
```diff
@@ -427,6 +428,7 @@ Ember._concat = concat;
 Ember._get = get;
 Ember._on = on;
 Ember._fn = fn;
+Ember._uniqueId = uniqueId;
```
chriskrycho, line 431: (For us as maintainers, not for you as a contributor!)

I will put this on the agenda for tomorrow's Framework team meeting; as a matter of policy I think we should *avoid* adding things to the `Ember` namespace.

packages/@ember/-internals/glimmer/lib/helpers/unique-id.ts
chriskrycho: All right, we chatted about this today and we on the Framework Core Team side need to do some follow-up work (to which you will be welcome to contribute if you like!) for simple function-based helpers like this one and how they relate to the ability to use [plain old functions as helpers](https://blog.emberjs.com/plain-old-functions-as-helpers) in Ember. For now, however, we are going to make this work *only* as a traditional helper, rather than as a plain JS function.

Accordingly, we can drop this import and make a couple other tweaks, and this should be good to go!

```suggestion
function uniqueId() {
```

packages/@ember/-internals/glimmer/index.ts
```diff
@@ -324,3 +324,4 @@ export {
   setComponentManager,
 } from './lib/utils/managers';
 export { isSerializationFirstNode } from './lib/utils/serialization-first-node-helpers';
+export { uniqueId } from './lib/helpers/unique-id';
```
chriskrycho, line 327: Here, you'll need to update to re-export the default:

```suggestion
export { default as uniqueId } from './lib/helpers/unique-id';
```

packages/@ember/-internals/glimmer/tests/integration/helpers/unique-id-test.js
```diff
@@ -1,8 +1,34 @@
-import { RenderingTestCase, strip, moduleFor, runTask } from 'internal-test-helpers';
+import {
+  AbstractTestCase,
+  RenderingTestCase,
+  strip,
+  moduleFor,
+  runTask,
+} from 'internal-test-helpers';
 import { setProperties } from '@ember/object';
+import { uniqueId } from '@ember/helper';
```
chriskrycho, line 9: Here, you'll need to make a couple changes:

1. Use [the `invokeHelper()` utility](https://api.emberjs.com/ember/4.6/functions/@ember%2Fhelper/invokeHelper) to test this behavior.
2. `invokeHelper()` relies on using `getValue` from `@glimmer/tracking/primitives/cache`, and for that you'll need to import from the corresponding internal location.

Net, the updates for imports should look *roughly* like this:

```suggestion
import { uniqueId, invokeHelper } from '@ember/helper';
import { getValue } from '@glimmer/validator';
```

packages/@ember/-internals/glimmer/lib/helpers/unique-id.ts
chriskrycho: All right, we chatted about this today and we on the Framework Core Team side need to do some follow-up work (to which you will be welcome to contribute if you like!) for simple function-based helpers like this one and how they relate to the ability to use [plain old functions as helpers](https://blog.emberjs.com/plain-old-functions-as-helpers) in Ember. For now, however, we are going to make this work *only* as a traditional helper, rather than as a plain JS function.

Accordingly, we can drop this import and make a couple other tweaks, and this should be good to go!

```suggestion
function uniqueId() {
```

packages/@ember/-internals/glimmer/tests/integration/helpers/unique-id-test.js
chriskrycho, line 9: Here, you'll need to make a couple changes:

1. Use [the `invokeHelper()` utility](https://api.emberjs.com/ember/4.6/functions/@ember%2Fhelper/invokeHelper) to test this behavior.
2. `invokeHelper()` relies on using `getValue` from `@glimmer/tracking/primitives/cache`, and for that you'll need to import from the corresponding internal location.

Net, the updates for imports should look *roughly* like this:

```suggestion
import { uniqueId, invokeHelper } from '@ember/helper';
import { getValue } from '@glimmer/validator';
```

packages/@ember/-internals/glimmer/tests/integration/helpers/unique-id-test.js
```diff
@@ -1,8 +1,34 @@
...
+  moduleFor(
+    'Helpers test: {{unique-id}} JS',
+    class JSRenderingTest extends AbstractTestCase {
+      constructor(assert) {
+        super(assert);
+        this.assert = assert;
+      }
+      ['@test it can be invoked as a JS function']() {
+        let first = uniqueId();
+        let second = uniqueId();
```
chriskrycho, line 22: Here, then, you should be able to use those new imports like this:

```suggestion
        let first = getValue(invokeHelper({}, uniqueId()));
        let second = getValue(invokeHelper({}, uniqueId()));
```

The rest of the test *should* “just work” once you’ve added that in. This gets at the bit we’ll need to do follow-up work around—this obviously isn’t a great experience in JS! However, we don’t need to tackle that here, and it needs some broader design work!

packages/ember/index.js
chriskrycho: Confirmed in today’s discussion: we will *not* add this to the `Ember` namespace, so we can remove this addition entirely!

```suggestion
```

packages/ember/index.js
```diff
@@ -3,6 +3,7 @@ import require, { has } from 'require';
 import { getENV, getLookup, setLookup } from '@ember/-internals/environment';
 import * as utils from '@ember/-internals/utils';
 import { Registry, Container } from '@ember/-internals/container';
+import { uniqueId } from '@ember/-internals/glimmer';
```
chriskrycho, line 6:
```suggestion
```

chriskrycho (changes requested):
Thanks again for working on this! We talked about it today at the Framework Core Team meeting, and noted that there were a couple gaps between the RFC which introduced `unique-id` and the actual implementation:

- the introduction of plain old functions as helpers via [RFC #0756: Default Helper Manager](https://rfcs.emberjs.com/id/0756-helper-default-manager)
- the lack of an import in support of template strict mode and first-class component templates
- the lack of a clear definition of 

These are *not* your responsibility to address as part of this change, however, so we’ve come up with a path that provides the public import we need for `<template>` support, as outlined in the rest of this review.

If you’re interested, we need to do some work separately to solve those issues, by writing a small RFC which defines how we should approach helpers like this going forward, not least so they *can* be usable as normal JS functions without the extra bits you'll see below. Let me know if you’re interested in that: @wycats is available to help get that written!

Again, though, the only things we need to do to land *this* are to address the comments in the review. Thanks for doing this!

packages/@ember/-internals/glimmer/lib/helpers/unique-id.ts
TechieQian: makes sense to me! pls loop me in on the follow-up work!

packages/@ember/-internals/glimmer/tests/integration/helpers/unique-id-test.js
TechieQian: Hrm, assuming we want just `uniqueId` since its not fn. So trying 
`getValue(invokeHelper({}, uniqueId)` I get the following very intentional sounding error.

```
Died on test #1: Found a helper manager, but it was an internal built-in helper manager. `invokeHelper` does not support internal helpers yet.
```

I dug into `@glimmer/runtime` a bit, and it seems like we just determine this to be an internal built-in helper manager based on the fact that it takes in a type function.

Here's the minimal repro of the situation:

```        
   let myRef = internalHelper(() => createConstRef('foo', 'unique-id'));
    try {
      invokeHelper({}, myRef);
    } catch (ex) {
      console.error('see error:', ex);
    }
```

I am missing some context of what an internal helper is vs external(?). Had this just been a `Reference`, I see there are helper functions like `valueForRef(createConst('foo', 'unique-id')) // => foo`

chriskrycho:
Sorry, forgot about this in the midst of a bunch of other things—just kicked back off CI for it and will take another review pass later today!

TechieQian:
> Sorry, forgot about this in the midst of a bunch of other things—just kicked back off CI for it and will take another review pass later today!

Hey @chriskrycho , let me know how we should proceed here. As it stands, I dont think we could just use `invokeHelper` per my comment on the suggestion.

chriskrycho:
Yeah, I noted that and will bring it up with folks next week! Thanks for checking in. Just had end-of-quarter extra stuff these past couple weeks.

wycats:
I met with @TechieQian and we discussed the bulk of the open issues here.

I took some (very incomplete and possibly cryptic) notes during the meeting: https://gist.github.com/wycats/6800716eb6744f63d9b9396670d5c625

The next steps are to meet about this in a spec meeting. @TechieQian is available to attend whenever the discussion is on the agenda. It's a pretty meaty topic, so we'll want to allocate plenty of time to discuss it.

wycats:
One observation that we made: since "functions as helpers" is now the law of the land, it basically works to export the `uniqueId` JS function and have people use it in `<template>`.

However, we'd need to be sure that we guarantee that when a function with no autotracking dependencies is used as a helper, it's never spuriously re-evaluated. I believe that this is the case today, but we'd need to be really sure.

bertdeblock (approved):
Looks good AFAICT!

chriskrycho:
1. This needs a rebase.
2. Last time we were here, CI was failing, and there were open questions about how to get it the rest of the way across the line. It was not clear to me from @wycats' comment above whether it's actually unblocked here or not.

NullVoxPopuli:
> I believe that this is the case today, but we'd need to be really sure.

I have never observed behavior to the contrary -- functions with no consumed tracked data never re-evaluate (unless the parent context is torn down and re-rendered -- which would be expected, ofc)

Can we move forward with this PR? <3

NullVoxPopuli:
I pushed up a rebase + some minor tweaks in #20171 -- thanks for getting this started @TechieQian !!! you're the best :tada:

PR #20175, opened 2022-08-30
Convert packages/ember to TS

wagenet:


packages/@ember/-internals/glimmer/lib/helper.ts
```diff
@@ -266,30 +298,6 @@ export const SIMPLE_CLASSIC_HELPER_MANAGER = new SimpleClassicHelperManager();
...
-  @for @ember/component/helper
-  @public
-  @since 1.13.0
-*/
-export function helper<T, P extends unknown[], N extends Dict<unknown>>(
-  helperFn: HelperFunction<T, P, N>
-): HelperFactory<SimpleHelper<T, P, N>> {
-  return new Wrapper(helperFn);
-}
+export { helper };
```
chriskrycho, line 301: Why the switch to doing this rather than just using `export function helper` as it was originally?

chriskrycho:
When I look to repro locally, I don't even get as far as the failures which show up in CI; it fails on `@ember/debug`. In `fish`:

```fish
$ env TEST_SUITE=package PACKAGE="@ember/debug" yarn test
```

packages/@ember/-internals/glimmer/lib/helper.ts
wagenet: No good reason, I'll switch it back.

Issue #20172, opened 2022-08-29
[QUEST] Reformatting Ember API docs for Typescript


ttbach:
# Summary
Currently, there is an ongoing initiative to add Typescript support to Ember API documentation. More specifically, that means adding the ability for API docs to be generated from typescript instead of jsdocs. In order to do that, we will need to modify the API code, such that it will be well formed to pass through a documentation generator for TS projects ([typedoc](https://github.com/TypeStrong/typedoc)). That means for modules that import Mixins, we need to move the jsdocs comments close to the functions that are defined inside the interface. For more details, see the Example section below.

# Issues

File | Package | Count | Owner | Status
-- | -- | -- | -- | --
index.ts | @ember/array | 3 |  @slijack-in | https://github.com/emberjs/ember.js/pull/20189
index.ts | @ember/controller | 1 |  @jalexakos | https://github.com/emberjs/ember.js/pull/20202
index.ts | @ember/enumerable | 1 |  @jalexakos | DONE
mutable.ts | @ember/enumerable | 1 |  @52052100 | DONE
evented.ts | @ember/object | 1 |  @pablonajera | https://github.com/emberjs/ember.js/pull/20189
mixin.ts | @ember/object | 4 |  @TechieQian | DONE
observable | @ember/object | 1 |  @52052100 | https://github.com/emberjs/ember.js/pull/20191
promise-proxy-mixin.ts | @ember/object | 1 | @TechieQian | https://github.com/emberjs/ember.js/pull/20192

# Example
Let's take a look at [objectsAt in array.ts](https://github.com/emberjs/ember.js/blob/master/packages/%40ember/array/index.ts#L369). Currently, the jsdocs are above the function definition of `objectsAt`

```ts
const EmberArray = Mixin.create(Enumerable, {
  ...
  /**
    This returns the objects at the specified indexes, using `objectAt`.
    ```javascript
    let arr = ['a', 'b', 'c', 'd'];
    arr.objectsAt([0, 1, 2]);  // ['a', 'b', 'c']
    arr.objectsAt([2, 3, 4]);  // ['c', 'd', undefined]
    ```
    @method objectsAt
    @param {Array} indexes An array of indexes of items to return.
    @return {Array}
    @public
   */
  objectsAt(indexes: number[]) {
    return indexes.map((idx) => objectAt(this, idx));
  },
  ...
});
```

[...]

PR #20189, opened 2022-09-08
[DOC] moving comment docs to interface methods – `Evented`

pablonajera:
In response to [QUEST#20172](https://github.com/emberjs/ember.js/issues/20172).

Moving docs for evented.ts where automation failed, as assigned. Ignoring @ comments, such as @param & @returns.

### DIFF
--- a/packages/@ember/object/evented.ts
+++ b/packages/@ember/object/evented.ts
@@ -61,6 +61,13 @@ interface Evented {
     be set as the "this" for the callback. This is a good way to give your
     function access to the object triggering the event. When the target
     parameter is used the callback method becomes the third argument.
+
+    @method on
+    @param {String} name The name of the event
+    @param {Object} [target] The "this" binding for the callback
+    @param {Function|String} method A function or the name of a function to be called on `target`
+    @return this
+    @public
   */
   on<Target>(
     name: string,
@@ -76,6 +83,13 @@ interface Evented {
     This function takes an optional 2nd argument that will become the "this"
     value for the callback. When the target parameter is used the callback method
     becomes the third argument.
+
+    @method one
+    @param {String} name The name of the event
+    @param {Object} [target] The "this" binding for the callback
+    @param {Function|String} method A function or the name of a function to be called on `target`
+    @return this
+    @public
   */
   one<Target>(
     name: string,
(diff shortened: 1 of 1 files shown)


ttbach (approved):
LGTM. Nice job!

ttbach (changes requested):
Due to comments raised by @chriskrycho [here](https://github.com/emberjs/ember.js/pull/20191#issuecomment-1244018007), let's remove the other jsdocs entry inside `Mixin.create()`

pablonajera:
@ttbach sounds good, all done!

ttbach (approved):

chriskrycho (changes requested):
~~Hey, thanks for doing this work. However, there are now doc comments on both the fields in the `interface` and in body of the `Mixin.create()` call. All items marked `@public` should be documented on the interface and *only* there. You should simply *move* the comments for those methods from the `Mixin.create()` body to the `interface`, replacing any comments which already exist on the `interface`. Thanks!~~ (was looking at the wrong branch!)

chriskrycho (approved):
~~This is good to go, thanks!~~ (Previous review here was mistaken, I was looking at the wrong branch!) [But it still has some fixes needed; see *next* review. 🤪]

packages/@ember/object/evented.ts
```diff
@@ -123,19 +76,13 @@ const Evented = Mixin.create({
     This function takes an optional 2nd argument that will become the "this"
     value for the callback. When the target parameter is used the callback method
     becomes the third argument.
-
-    @method one
-    @param {String} name The name of the event
-    @param {Object} [target] The "this" binding for the callback
-    @param {Function|String} method A function or the name of a function to be called on `target`
-    @return this
-    @public
```
chriskrycho, line 132: Ah, linting caught something important for us. All of these annotations need to be copied over to the new form of the docs! They are used by the doc generator!

chriskrycho (changes requested):
Aaaaaand one more review putting it back in "needs changes": we need to preserve all the comment properties like `@method`, `@param`, `@return`, `@public`, etc.

packages/@ember/object/evented.ts
pablonajera: Haha no problem, will do!

pablonajera: Done! Please re-review when convenient

chriskrycho (approved):

chriskrycho (approved):

PR #20191, opened 2022-09-09
[DOC] Reformatting Ember API docs for Typescript - `@ember/object/observable`

52052100:
In response to https://github.com/emberjs/ember.js/issues/20172.

Move comments and jsdocs from Mixin.create({...}) to interface for observable.ts.

### DIFF
--- a/packages/@ember/object/observable.ts
+++ b/packages/@ember/object/observable.ts
@@ -91,29 +91,113 @@ export type ObserverMethod<Target, Sender> =
 */
 interface Observable {
   /**
-   * Retrieves the value of a property from the object.
-   */
+    Retrieves the value of a property from the object.
+
+    This method is usually similar to using `object[keyName]` or `object.keyName`,
+    however it supports both computed properties and the unknownProperty
+    handler.
+
+    Because `get` unifies the syntax for accessing all these kinds
+    of properties, it can make many refactorings easier, such as replacing a
+    simple property with a computed property, or vice versa.
+
+    ### Computed Properties
+
+    Computed properties are methods defined with the `property` modifier
+    declared at the end, such as:
+
+    ```javascript
+    import { computed } from '@ember/object';
+
+    fullName: computed('firstName', 'lastName', function() {
+      return this.get('firstName') + ' ' + this.get('lastName');
+    })
+    ```
+
+    When you call `get` on a computed property, the function will be
+    called and the return value will be returned instead of the function
+    itself.
+
+    ### Unknown Properties
+
+    Likewise, if you try to call `get` on a property whose value is
+    `undefined`, the `unknownProperty()` method will be called on the object.
+    If this method returns any value other than `undefined`, it will be returned
+    instead. This allows you to implement "virtual" properties that are
+    not defined upfront.
+  */
   get(key: string): unknown;
 
   /**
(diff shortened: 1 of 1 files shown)


ttbach (approved):

chriskrycho (changes requested):
Thanks so much for doing this! Note that we don't want to just *duplicate* the entries; we want to *move* them. So the existing doc comments on the body of `Mixin.create()` on lines 340 and following need to be removed as well.

ttbach:
> Thanks so much for doing this! Note that we don't want to just _duplicate_ the entries; we want to _move_ them. So the existing doc comments on the body of `Mixin.create()` on lines 340 and following need to be removed as well.

@chriskrycho if we remove existing doc comments, doesn't that remove the comments in prod right now? I wasn't sure, which is why I told @52052100 to do that. Feel free to correct me if I'm wrong.

chriskrycho:
> if we remove existing doc comments, doesn't that remove the comments in prod right now?

Nope, for two reasons:

1. The prod stuff is built and served statically; it’s not re-crawling this every time (that would be :scream:).
2. The comments still exist; they’re just moving to a new location in the file, which will generate the same output when run through YUIDoc.

If we have both, I don’t even know what YUIDoc will do—it’s quite possible and even likely it will end up with *duplicates*, which we definitely don’t want.

52052100:
> > if we remove existing doc comments, doesn't that remove the comments in prod right now?
> 
> Nope, for two reasons:
> 
> 1. The prod stuff is built and served statically; it’s not re-crawling this every time (that would be 😱).
> 2. The comments still exist; they’re just moving to a new location in the file, which will generate the same output when run through YUIDoc.
> 
> If we have both, I don’t even know what YUIDoc will do—it’s quite possible and even likely it will end up with _duplicates_, which we definitely don’t want.

Thank you for your explanation, will remove the comments inside Mixin.create()

ttbach (approved):

packages/@ember/object/observable.ts
```diff
@@ -435,15 +280,137 @@ const Observable = Mixin.create({
...
+    @method endPropertyChanges
+    @return {Observable}
+    @private
+  */
+  endPropertyChanges() {
+    endPropertyChanges();
+    return this;
+  },
+
+  /**
```
chriskrycho, line 396: This comment needs to be moved to the interface.

chriskrycho (changes requested):
One missed – let me know when you’ve moved that one and we’ll be ready to merge it. Thanks!

packages/@ember/object/observable.ts
52052100: I've moved the missing comment, thanks!

chriskrycho (approved):

chriskrycho:
Thank you! I’ve triggered CI and set it to merge once it's ✅!

chriskrycho (changes requested):
Ah, this won't pass because linting is correctly identifying a gap in the original instructions. You'll need to copy over all the annotations (`@public`, `@method`, etc.)—those are critical parts of how the doc generator tool works. Please incorporate those and then I'll re-review and h*ten* it should be ready to merge.

chriskrycho (approved):
Thanks for making the update!

PR #20192, opened 2022-09-11
[DOC] Reformatting Ember API docs for Typescript - proxy-mixin-interface

TechieQian:
In response to https://github.com/emberjs/ember.js/issues/20172.

### DIFF
--- a/packages/@ember/object/promise-proxy-mixin.ts
+++ b/packages/@ember/object/promise-proxy-mixin.ts
@@ -111,24 +111,44 @@ interface PromiseProxyMixin<T> {
   /**
     If the proxied promise is rejected this will contain the reason
     provided.
+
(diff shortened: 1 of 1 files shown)


ttbach (approved):

chriskrycho (changes requested):
Thanks for doing this! Looks good overall, but the doc comments on the `interface` need to include *all* of the original doc comments from the `Mixin` definition. Please copy over the things like `@property`, `@default`, and `@public`, or the doc generator won't work correctly. Also, please remove the corresponding doc comments from the mixin entirely. Thanks!

TechieQian:
Sounds good! done

chriskrycho (approved):
Thanks for making the changes here!

PR #20199, opened 2022-09-19
[DOC] Format API docs for Typescript - @ember/array

shibulijack:
**Issue**: https://github.com/emberjs/ember.js/issues/20172.

**Details**:
- Moved comments from Mixin.create({...}) to interface.
- Added missing interface property for `[]` in `array/index`.
- Added interface properties for `array/proxy`

### DIFF
--- a/packages/@ember/array/index.ts
+++ b/packages/@ember/array/index.ts
@@ -255,6 +255,9 @@ interface EmberArray<T> extends Enumerable {
 
     Your array must support the `length` property. Your replace methods should
     set this property whenever it changes.
+
+    @property {Number} length
+    @public
   */
   length: number;
   /**
@@ -275,6 +278,11 @@ interface EmberArray<T> extends Enumerable {
     arr.objectAt(4);   // undefined
     arr.objectAt(5);   // undefined
     ```
+
+    @method objectAt
+    @param {Number} idx The index of the item to return.
+    @return {*} item at index or undefined
+    @public
   */
   objectAt(idx: number): T | undefined;
   /**
@@ -286,6 +294,11 @@ interface EmberArray<T> extends Enumerable {
     arr.objectsAt([0, 1, 2]);  // ['a', 'b', 'c']
     arr.objectsAt([2, 3, 4]);  // ['c', 'd', undefined]
     ```
+
+    @method objectsAt
+    @param {Array} indexes An array of indexes of items to return.
+    @return {Array}
+    @public
    */
   objectsAt(indexes: number[]): Array<T | undefined>;
   /**
@@ -301,6 +314,10 @@ interface EmberArray<T> extends Enumerable {
     peopleToMoon.set('[]', ['Collins']); // ['Collins']
     peopleToMoon.get('[]'); // ['Collins']
     ```
+
+    @property []
+    @return this
+    @public
   */
   '[]': this;
   /**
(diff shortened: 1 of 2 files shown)


ttbach (approved):
lgtm

packages/@ember/array/index.ts
```diff
@@ -250,86 +250,13 @@ function mapBy<T>(this: EmberArray<T>, key: string) {
...
-  },
-
   /**
     __Required.__ You must implement this method to apply this mixin.
 
     Your array must support the `length` property. Your replace methods should
     set this property whenever it changes.
-
-    @property {Number} length
-    @public
```
chriskrycho, line 330: We need to preserve these annotations in the switchover! They're used by the doc generation tool in knowing (a) what to emit (`@public`) and (b) how to render it in the docs (the type, in particular).

packages/@ember/array/index.ts
```diff
@@ -1552,18 +1328,8 @@ const MutableArray = Mixin.create(EmberArray, MutableEnumerable, {
     colors.insertAt(2, 'yellow');  // ['red', 'green', 'yellow', 'blue']
     colors.insertAt(5, 'orange');  // Error: Index out of range
     ```
-
-    @method insertAt
-    @param {Number} idx index of insert the object at.
-    @param {Object} object object to insert
-    @return {EmberArray} receiver
-    @public
```
chriskrycho, line 1560: Same for the annotations for this mixin!

packages/@ember/array/proxy.ts
```diff
@@ -169,30 +196,10 @@ class ArrayProxy<T, C extends EmberArray<T> | T[] = T[]>
     this._removeArrangedContentArrayObserver();
   }
 
-  /**
-    The content array. Must be an object that implements `Array` and/or
-    `MutableArray.`
-
-    @property content
-    @type EmberArray
-    @public
```
chriskrycho, line 178: Same for these.

chriskrycho (changes requested):
Overall looks good; there's a miss here on needing to copy over the various doc annotations as well, but otherwise 👍🏼 – thanks. Let me know when you've updated for the annotations, and I'll re-review and we'll merge!

packages/@ember/array/index.ts
shibulijack: Thanks for clarifying, Chris. This PR initially had those annotations earlier but I removed it later after referring to similar PRs. I'll add them back now.

chriskrycho: Yeah, not your fault at all—thank you!

chriskrycho (approved):
Thanks for making the changes!

packages/@ember/array/index.ts
```diff
@@ -1224,29 +1021,17 @@ const EmberArray = Mixin.create(Enumerable, {
       }); // false (true && false && false)
     ```
 
-    @method reduce
-    @param {Function} callback The callback to execute
-    @param {Object} initialValue Initial value for the reduce
-    @return {Object} The reduced value.
+    @method isAny
```
chriskrycho, line 1024: Whoops, ended up with the wrong doc comment annotations copied over here—you have the `isAny` annotations on `reduce`!

chriskrycho (changes requested):
One miss!

packages/@ember/array/index.ts
shibulijack: Oops! 😅 
Fixed it.

chriskrycho (approved):

PR #20202, opened 2022-09-21
[DOC] Reformat API docs for Typescript - @ember/controller

jalexakos:
## Summary
Addressing Issue: [20172](https://github.com/emberjs/ember.js/issues/20172)

## Details
- Moved documentation for public APIs from `Mixin.create()` to the `ControllerMixin` interface
- Removed all unnecessary JSDocs tags (comments with `@` preceding them)
- Left private methods and properties documentation untouched

### DIFF
--- a/packages/@ember/controller/index.ts
+++ b/packages/@ember/controller/index.ts
@@ -32,20 +32,6 @@ interface ControllerMixin<T> extends ActionHandler {
   _qpDelegate: unknown | null;
 
   isController: true;
-  target: unknown | null;
-  model: T;
-
-  queryParams: Array<ControllerQueryParam>;
-
-  transitionToRoute(...args: RouteArgs<Route>): Transition;
-
-  replaceRoute(...args: RouteArgs<Route>): Transition;
-}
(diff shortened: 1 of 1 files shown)


packages/@ember/controller/index.ts
```diff
@@ -58,40 +44,16 @@ const ControllerMixin = Mixin.create(ActionHandler, {
     as part of the application's initialization process. In most cases the
     `target` property will automatically be set to the logical consumer of
     actions for the controller.
-
-    @property target
-    @default null
```
chriskrycho, line 63: This was just a miss in the original writeup, so not your fault, but we need to keep these! They're part of how the doc generator knows what to output.

chriskrycho (changes requested):
Thanks for doing this! One tweak needed – let me know when you've made it and I'll re-review and merge.

packages/@ember/controller/index.ts
jalexakos: Whoops! Re-added 😅

chriskrycho (approved):
Thanks for making the updates here! 🚀

PR #20203, opened 2022-09-21
[FEATURE] Update to Typescript 4.8

chriskrycho:
We were previously using 4.6. We'll aim to keep this up to date going forward.

Here, update the dependency and then address two safety issues it caught on either 4.7 or 4.8 but did not back on 4.6.

In the *next* PR I’ll also be adding CI support for testing a range, but *only* targeting 4.8 at the moment with it because we don’t *want* to couple ourselves to a wider range needlessly. See that PR for further details!

### DIFF
--- a/package.json
+++ b/package.json
@@ -157,7 +157,7 @@
     "simple-dom": "^1.4.0",
     "testem": "^3.8.0",
     "testem-failure-only-reporter": "^1.0.0",
-    "typescript": "~4.6.4"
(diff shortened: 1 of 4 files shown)


wagenet (approved):

chriskrycho:
Just rebased this on top of `master` since I *think* #20205 will have fixed the linting issues. 🤞🏼

PR #20218, opened 2022-10-03
[BUGFIX beta] Remove `Function` and `String` from `EXTEND_PROTOTYPES` type

bertdeblock:


types/preview/ember/index.d.ts
```diff
@@ -240,8 +240,6 @@ declare module 'ember' {
     namespace ENV {
       const EXTEND_PROTOTYPES: {
```
bertdeblock, line 241:
```suggestion
      const EXTEND_PROTOTYPES: boolean | {
```

Should this be this ^ as well? Since `EXTEND_PROTOTYPES: false` also works.

chriskrycho: If it *only* allows `false`, then:

```suggestion
      const EXTEND_PROTOTYPES: false | {
```

Otherwise, go for your suggestion!

bertdeblock: `true` also works indeed.

bertdeblock: Updated!

chriskrycho:
Looks like blueprints CI failed: maybe a timeout (not at my machine right now)?

chriskrycho (approved):

PR #20086, opened 2022-05-06
Improve types for some array internals

wagenet:
It's hard to tell in this diff, but some of the EmberArray types changed a little bit to be more accurate.

PR #20219, opened 2022-10-04
[BUGFIX] Restore mixin based EmberArray detection

jrjohnson:
This is needed for ember-data to tap into and masquerade as an EmberArray to avoid being clobbered by Ember.A().

Refs emberjs/data#8202

This functionality was dropped in ~#20092~ #20086

I couldn't figure out where a test for this would go, but opening a PR as a starting place for discussion and happy to update as needed.

### DIFF
--- a/packages/@ember/-internals/utils/lib/ember-array.ts
+++ b/packages/@ember/-internals/utils/lib/ember-array.ts
@@ -1,4 +1,4 @@
-import type EmberArray from '@ember/array';
+import EmberArray from '@ember/array';
 import { _WeakSet } from '@glimmer/util';
 
 const EMBER_ARRAYS = new _WeakSet();
@@ -8,5 +8,9 @@ export function setEmberArray(obj: object) {
 }
 
 export function isEmberArray(obj: unknown): obj is EmberArray<unknown> {
-  return EMBER_ARRAYS.has(obj as object);
+  if (EMBER_ARRAYS.has(obj as object)) {
+    return true;
+  }
+
+  return EmberArray.detect(obj);
 }
--- a/packages/@ember/-internals/utils/lib/ember-array.ts
+++ b/packages/@ember/-internals/utils/lib/ember-array.ts
@@ -1,4 +1,4 @@
-import EmberArray from '@ember/array';
+import type EmberArray from '@ember/array';
 import { _WeakSet } from '@glimmer/util';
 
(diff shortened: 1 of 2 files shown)


runspired:
I'd be willing to add an ember-version dependent workaround if we don't want to restore the detect API, but there needs to be a mechanism for preventing what A can do to an object for EmberData to be able to help folks pave a path away from A.

wagenet:
@jrjohnson darn, my attempt to shift blame failed 😄

That said, I believe this is reasonable, I'll see if I can get someone else to second it.

wagenet (approved):

chriskrycho (approved):
Seems reasonable to me, too. DIAF mixins *and* `EmberArray`… but until then yeah, let's do this to avoid breaking folks.

jrjohnson:
I'm not sure how to fix the test failures here, I *think* it maybe has to do with typescript and using the EmberArray import for the type and in code, but I don't know enough about typescript to fix this. I've tried various syntax and tried to figure out the difference between import type and import as well as looking around for other examples of something like this in the ember source, but I think I don't have enough of an idea on how typescript works to figure this out.

wagenet:
Annoyingly, I couldn't reproduce them locally either. I've been pretty busy as well 😕

jrjohnson:
I'm surprised this doesn't fail for you locally @wagenet, it certainly does for me. Maybe we can connect on discord and pair on debugging?

I think maybe this is a circular dependency issue. `packages/@ember/-internals/utils/lib/ember-array.ts` has `import EmberArray from '@ember/array';` and `packages/@ember/array/index.ts` has `import { isEmberArray, setEmberArray } from '@ember/-internals/utils';` and it's not immediately clear to me how to disconnect these.

wagenet:
@jrjohnson pairing would make sense. Are you on the Ember Discord? Maybe we can connect there.

jrjohnson:
@wagenet / @runspired I'm not sure if this is the right way to go, but wanted to put some strawman code up for discussion. Calls `preventEmberA` when it is present to prevent and object from being Ember.A()ed.

wagenet:
That's definitely one option. I think also, we want to change the behavior of Ember.A to not mutate anymore. The docs say that it doesn't (which we know is not true). I've talked with others on the core team about this. I'll try to look into it more this week.

PR #20224, opened 2022-10-14
Integrate types for the `warn` method

r-52:
This PR consist of two commits (if they are okay, I can them squash them)

1. Mention the `id` property in the options object of the `warn` method. The `deprecate` method mentions the sub-options properties and I tried to align the doc's for the warn method.
2. integrate the Typescript types for the `warn` method. Right now, the `test` parameter can be used as a `test`-condition or as a shortcut for the `options` object to always display the warning. The `options` object itself is required for the warning.

What do you think?

### DIFF
--- a/packages/@ember/debug/lib/warn.ts
+++ b/packages/@ember/debug/lib/warn.ts
@@ -93,10 +93,11 @@ if (DEBUG) {
     @for @ember/debug
     @static
     @param {String} message A warning to display.
-    @param {Boolean} test An optional boolean. If falsy, the warning
-      will be displayed.
-    @param {Object} options An object that can be used to pass a unique
-      `id` for this warning.  The `id` can be used by Ember debugging tools
+    @param {Boolean|Object} test An optional boolean. If falsy, the warning
+      will be displayed. If `test` is an object, the `test` parameter can
+      be used as the `options` parameter and the warning is displayed.
+    @param {Object} options
+    @param {String} options.id The `id` can be used by Ember debugging tools
       to change the behavior (raise, log, or silence) for that specific warning.
       The `id` should be namespaced by dots, e.g. "ember-debug.feature-flag-with-features-stripped"
     @public
--- a/packages/@ember/debug/lib/warn.ts
+++ b/packages/@ember/debug/lib/warn.ts
@@ -112,8 +112,10 @@ if (DEBUG) {
     assert(missingOptionsDeprecation, Boolean(options));
     assert(missingOptionsIdDeprecation, Boolean(options && options.id));
 
-    // SAFETY: we checked this by way of the `arguments` check above.
-    invoke('warn', message, test as boolean, options);
+    // SAFETY: we have explicitly assigned `false` if the user invoked the
+    // arity-2 version of the overload, so we know `test` is always either
+    // `undefined` or a `boolean` for type-safe callers.
+    invoke('warn', message, test as boolean | undefined, options);
   };
 }
 


packages/@ember/debug/lib/warn.ts
```diff
@@ -9,7 +9,11 @@ export interface WarnOptions {
 }
 
 export type RegisterHandlerFunc = (handler: HandlerCallback<WarnOptions>) => void;
-export type WarnFunc = (message: string, test?: boolean, options?: WarnOptions) => void;
+export type WarnFunc = (
+  message: string,
+  test: boolean | WarnOptions,
+  options?: WarnOptions
+) => void;
```
chriskrycho, line 16: While this will mostly work, the right way to do this looks a little bit different, because we want to avoid it being legal for someone to pass the options *twice*, which this signature would allow. Instead, we want to use an [overload](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads). You can see how that would work, using an interface type instead, [here][p]. Also, the actual implementation *requires* that the options object be passed (it will assert if it's not).

I believe this should do the trick:

```suggestion
export interface WarnFunc {
  (message: string, options: WarnOptions): void;
  (
    message: string,
    test: boolean,
    options: WarnOptions
  ): void;
}
```

[p]: https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgOpyiA8gBzMAexAGdkBvAKGWWABMAuZYsKUAcwG4KBfCi0SLEQp0mAGIBXEAnJVkACgC2EYsThsIjZqxBsANMgJ5CJRqOzGixAJSMAbgTpdqSlWo1aW7A5GaMARgQEADYQcCAGRvhWZhgW0SS2yA5OPHy0EAjBGCihYMgA7nGx4lIIXBQA9JXIAAJgxAC0EAAeOJlgzVBQBFAURZjyAOQEABbIkOMgBCpD1lzVdQ3NbR1dPX0DIMNjExBTM8RDPlASEPP9cTvjk8jTsydnBmQ0DMhDoxDBwQRDyNwXLbXPYHB7kV6MD5fH5-AELGr1JqtdoITrQDaXQYjG77O6HY7guiQz7fX7-Z4Q94kmH-C4UIA

packages/@ember/debug/lib/warn.ts
```diff
@@ -106,7 +111,7 @@ if (DEBUG) {
     assert(missingOptionsDeprecation, Boolean(options));
     assert(missingOptionsIdDeprecation, Boolean(options && options.id));
 
-    invoke('warn', message, test, options);
+    invoke('warn', message, test as boolean, options);
```
chriskrycho, line 114: Let's add a `SAFETY` comment here explaining why the cast is safe.

```suggestion
    // SAFETY: we have explicitly assigned `false` if the user invoked the
    // arity-2 version of the overload, so we know `test` is always either
    // `undefined` or a `boolean` for type-safe callers.
    invoke('warn', message, test as boolean | undefined, options);
```

chriskrycho (changes requested):
Thanks for opening this! One small tweak needed on the TS, and in fact this helps us catch an error in what the types here allowed!

kategengler:
@emberjs/typescript-core is this relevant?

gitKrystan:
> @emberjs/typescript-core is this relevant?

It should be OK to merge with @chriskrycho's suggestions.

gitKrystan (approved):

wagenet (approved):

wagenet:
I rebased and it turns out we'd already integrated most of the changes in other PRs. That said, there are some documentation improvements here that we could merge in.

kategengler:
Enabled auto-merge!

PR #20226, opened 2022-10-16
[CLEANUP] use `console.debug` directly

r-52:
Right now `console.info` is called without a fallback to `console.log`. On the other hand `console.log` had a fallback to `console.log` - it looks like (based on caniuse) that all major browsers support the different `console.*` methods.
This commit follows the same pattern like `console.info`.


bertdeblock (approved):

chriskrycho (approved):
Thank you!

PR #20231, opened 2022-10-20
Add to the preview-types an overload of `computed` with `get` and `set`

boris-petrov:
Which I believe is missing.

cc @chriskrycho, @wagenet

### DIFF
--- a/types/preview/@ember/object/index.d.ts
+++ b/types/preview/@ember/object/index.d.ts
@@ -27,6 +27,10 @@ declare module '@ember/object' {
     ...args: [...dependentKeys: string[], callback: AnyFunction]
   ): BasicComputedProperty;
 
+  export function computed(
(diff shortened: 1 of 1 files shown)


chriskrycho (commented):
I'll need to double check my conversion notes, but I believe these were removed intentionally. Per RFC 0800 we are intentionally providing only minimal support for the Ember Classic APIs, and while this form technically works I don't know (a) whether it is documented or (b) whether there is any reason to use it over switching to a proper `get`/`set` on the class itself. I will follow up later today.

chriskrycho:
Poked at this again and chatted with @dfreeman and we actually are going to do the *opposite* here. We will not support any form of passing-a-function-to-`@computed`. Instead, users should use the decorated getter form. Thanks for flagging this up for us!

PR #20234, opened 2022-10-26
[DRAFT/WIP] Publish stable types

chriskrycho:
This isn't quite ready for prime time yet, but it does prove out the basic viability of @chadhietala's suggestion that we may not need to fix all the circularities before we can publish. We in fact can just publish as is: none of the circularities are *hard* blockers, at least for *this* part of the effort (though they will be for getting docs published with these as the source of truth).

The things we will need to do to be able to execute the rest of the way on this, given our goals:

- [x] Fix a couple cases where we are using private names in public types, specifically around the `OWNER` from `@glimmer/owner`.
- [ ] Put the hacky `generate-tsconfigs.mjs` script somewhere besides the root, expand its capabilities to include wrapping generated modules in a `declare module` statement and running Prettier on the result, and rename it accordingly.
- [x] Land @wagenet's in-progress PR (#20175) to convert the `ember` package to TS so we can publish types for it, which will unblock some of the other Ember packages as well.

An open question: do we want to try to properly exclude the private and internal parts of Ember's APIs? Publishing them may actually be preferable and even to some degree necessary for (officially-maintained) "collaborator" libraries which use private APIs.


chriskrycho:
The remaining emit error is blocked on #20175.

There are at least two other issues flagged up by CI here:

- A build error:

    ```
    You specified a `"srcDir": @ember/-internals/browser-environment` which does not exist and did not specify `"allowEmpty": true`.
    ```

    I don't believe that's related to this PR (it's unclear how it could be!), but will investigate further early next week.

- A types error for `@types/node` on TS 4.9, fixed via #20235.

chriskrycho:
Superseded by the actual implementation at #20275, which used the ideas from this but basically none of its implementation!

PR #20239, opened 2022-10-29
[DOC] add `in-element` example  for `insertBefore` usage

### LINKED ISSUE #19911: [Documentation] in-element template helper does not show how to use insertBefore
### 📙 Describe the Issue
The documentation available at https://api.emberjs.com/ember/3.21/classes/Ember.Templates.helpers/methods/in-element?anchor=in-element mentions that there is an optionall named argument called `insertBefore`, but it does not feature a code sample that uses it.

### 🤔 Expected Behavior
Something like the following code example:

```handlebars
{{#in-element this.destinationElement insertBefore=null}}
  <div>Some content</div>
{{/in-element}}
```

Along with a short description saying that this will append the content instead of replacing it.

ghost:
The documentation available for the `in-element` helper mentions that there is an optional named argument, insertBefore, but it does not feature a code sample that uses it. Let's add a code sample which does use it to help developers using the API docs.

Fixes #19911.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/syntax/in-element.ts
+++ b/packages/@ember/-internals/glimmer/lib/syntax/in-element.ts
@@ -31,12 +31,13 @@
  append the block content to the end of any existing content. Any other value than
  `null` is currently not supported.
 
- For example:
- ```handlebars
- {{#in-element this.destinationElement insertBefore=null}}
-  <div>Some content</div>
- {{/in-element}}
- ```
+     For example:
+
+     ```handlebars
+     {{#in-element this.destinationElement insertBefore=null}}
+       <div>Some content</div>
+     {{/in-element}}
+     ```
 
  @method in-element
  @for Ember.Templates.helpers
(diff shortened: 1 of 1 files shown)


packages/@ember/-internals/glimmer/lib/syntax/in-element.ts
```diff
@@ -27,8 +27,16 @@
...
+ `destinationElement`. Passing `null` to `insertBefore` changes the behaviour to
+ append the block content to the end of any existing content. Any other value than
+ `null` is currently not supported.
+
+ For example:
+ ```handlebars
+ {{#in-element this.destinationElement insertBefore=null}}
+  <div>Some content</div>
+ {{/in-element}}
+ ```
```
chriskrycho, line 39: Markdown formatting:

```suggestion
     For example:

     ```handlebars
     {{#in-element this.destinationElement insertBefore=null}}
       <div>Some content</div>
     {{/in-element}}
     ```
```

This will make it part of the actual bullet point describing the argument, and will also make sure the code block is properly parsed as distinct from the preceding comment regardless of the Markdown parser used.

chriskrycho (approved):

chriskrycho:
Thanks! I applied the tweak I suggested, and it will merge when it's green on CI!

geneukum:
Awesome, thanks so much @chriskrycho!

PR #20240, opened 2022-10-29
[DOC] clarify `filter` computed callback method behavior

### LINKED ISSUE #19559: [Documentation] filter computed property
The filter function should probably say whether returning true or false filters or doesn't. It only mentions a callback with no reference to how the callback filters

### ➕ Additional Context
Originally reported at https://github.com/ember-learn/ember-api-docs/issues/619.

ghost:
Right now, the filter function's API docs mention a callback, but do not clarify how the callback function filters down the array. Let's explicitly specify that truthy values will keep the array element, and that anything else will not. This should make the usage a little bit more clear.

Fixes #19559.

### DIFF
--- a/packages/@ember/object/lib/computed/reduce_computed_macros.ts
+++ b/packages/@ember/object/lib/computed/reduce_computed_macros.ts
@@ -429,8 +429,7 @@ export function mapBy(dependentKey: string, propertyKey: string) {
 }
 
 /**
-  Filters the array down to the elements that pass the test implemented by the
-  callback.
+  Filters the array by the callback, like the `Array.prototype.filter` method.
 
   The callback method you provide should have the following signature:
   - `item` is the current item in the iteration.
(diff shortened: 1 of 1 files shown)


packages/@ember/object/lib/computed/reduce_computed_macros.ts
```diff
@@ -440,6 +441,9 @@ export function mapBy(dependentKey: string, propertyKey: string) {
   function filterCallback(item, index, array);
   ```
 
+  In the callback method, return a value that coerces to true to keep the
+  element, or to false otherwise.
```
chriskrycho, line 445:
```suggestion
  In the callback, return a truthy value that coerces to true to keep the
  element, or a falsy to reject it.
```

packages/@ember/object/lib/computed/reduce_computed_macros.ts
```diff
@@ -429,7 +429,8 @@ export function mapBy(dependentKey: string, propertyKey: string) {
 }
 
 /**
-  Filters the array by the callback.
+  Filters the array down to the elements that pass the test implemented by the
+  callback.
```
chriskrycho, line 433: If we are tweaking this, I think the best way to do so is probably to refer to `Array.prototype.filter`.

```suggestion
  Filters the array by the callback, like the `Array.prototype.filter` method.
```

chriskrycho (commented):
Thanks! Suggested (and will apply) a couple tweaks here.

chriskrycho (approved):

PR #20245, opened 2022-10-31
Ember.A should not modify underlying array

wagenet:
Resolves #20219

### DIFF
--- a/packages/@ember/array/index.ts
+++ b/packages/@ember/array/index.ts
@@ -2,7 +2,7 @@
 @module @ember/array
 */
 import { DEBUG } from '@glimmer/env';
-import { PROXY_CONTENT } from '@ember/-internals/metal';
+import { ComputedProperty, descriptorForDecorator, PROXY_CONTENT } from '@ember/-internals/metal';
 import {
   objectAt,
   replaceInNativeArray,
--- a/packages/@ember/array/index.ts
+++ b/packages/@ember/array/index.ts
@@ -1199,6 +1199,25 @@ interface EmberArray<T> extends Enumerable {
   */
   without(value: T): NativeArray<T>;
 }
+
+const emberArrayBrackets = nonEnumerableComputed({
+  get() {
+    return this;
+  },
+  set(_key, value) {
+    this.replace(0, this.length, value);
+    return this;
+  },
+});
+
+const emberArrayFirstObject = nonEnumerableComputed(function () {
+  return objectAt(this, 0);
+}).readOnly();
+
+const emberArrayLastObject = nonEnumerableComputed(function () {
+  return objectAt(this, this.length - 1);
+}).readOnly();
+
 const EmberArray = Mixin.create(Enumerable, {
   init() {
     this._super(...arguments);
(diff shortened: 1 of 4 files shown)


packages/@ember/array/index.ts
```diff
@@ -1193,6 +1193,25 @@ interface EmberArray<T> extends Enumerable {
...
+  },
+});
+
+const emberArrayFirstObject = nonEnumerableComputed(function () {
+  return objectAt(this, 0);
+}).readOnly();
+
+const emberArrayLastObject = nonEnumerableComputed(function () {
+  return objectAt(this, this.length - 1);
+}).readOnly();
```
wagenet, line 1213: I don't love how we're assigning them to variables here, but I can't find any other good way to get a reference to their CPs.

packages/@ember/array/index.ts
```diff
@@ -1935,6 +1942,127 @@ NativeArray.keys().forEach((methodName) => {
...
+    return new EmberAProxy(arr);
+  }
+
+  constructor(arr: T[] = []) {
+    let self = this;
+
+    let proxy = new Proxy(arr, {
+      get(target, prop /*, _receiver */) {
+        // These nonEnumerableComputed properties need special handling
+        if (prop === '[]') {
```
wagenet, line 1980: Ideally we'd have something more generic, but we aren't going to be adding stuff to EmberArray going forwards so this should be fine.

chriskrycho:
@wagenet not that "Extend Prototypes" is failing on CI.

chriskrycho (commented):
This seems reasonable to me, but I'd still really like it if someone who's more familiar with this could give it some 👀 as well. @rwjblue do you have 10 minutes (tops) to look at it?

wagenet:
I think I'd like to put this behind a flag. I think (hope) it won't cause issues for people, but who knows what people are actually doing and whether they're relying on the mutating behavior.

jrjohnson:
I agree that it should go behind a flag. Because many addons call `A()` on basically everything there are probably lots of components in production today that pass something through an addon helper such as `{{map-by "id" this.posts}}` and then rely on some prototype pollution elsewhere in a getter or action without realizing that the template code had modified the original `this.posts`. Being able to opt into this change seems like a good way to find out where the breaks are outside of the full deprecation in https://github.com/emberjs/rfcs/pull/864

locks:
@wagenet I'm working on pushing `tracked-built-ins` to recommended and saw this PR. Is this something we still want to push forward?

wagenet:
@locks This was partially motivated by issues Ember Data was having. @runspired what's the status of that stuff?

runspired:
still needed

kategengler:
Is this still relevant?

runspired:
@kategengler I believe so

kategengler:
@wagenet should we merge then?

wagenet:
@kategengler I would like to and it is flagged so it should be safe. That said, I haven't really gotten a proper review from anyone still.

kategengler:
> @kategengler I would like to and it is flagged so it should be safe. That said, I haven't really gotten a proper review from anyone still.

I don't have a clue here or I would offer you one :) I'm just triaging PRs.

NullVoxPopuli:
can this be rebasod?

packages/@ember/array/index.ts
```diff
@@ -2124,8 +2253,18 @@ if (ENV.EXTEND_PROTOTYPES.Array) {
       // SAFETY: If it's a true native array and it is also an EmberArray then it should be an Ember NativeArray
       return arr as unknown as NativeArray<T>;
     } else {
-      // SAFETY: This will return an NativeArray but TS can't infer that.
-      return NativeArray.apply(arr ?? []) as NativeArray<T>;
+      if (EMBER_A_NON_MODIFYING) {
+        // Remove this in Ember 5.0.
```
NullVoxPopuli, line 2257: 5.0? :sweat_smile:

wagenet: Heh... should be 6.0 now, I guess.

kategengler:
Since we're planning to deprecate Ember Array and have not yet fixed this, should this be closed?

kategengler:
Closing since we plan to deprecate Ember Array

PR #20256, opened 2022-11-07
[BUGFIX release] Correct types for Ember Arrays

### LINKED ISSUE #20230: [Preview Types Bug] `map` on an `EmberArray` returns `EmberArray` but shouldn't
### 🐞 Describe the Bug
This applies for the preview-types from Ember 4.8.0.

### 🔬 Minimal Reproduction
```ts
const x: NativeArray<string> = A();
const y = x.map(z => z);
```

### 😕 Actual Behavior
`y` is typed as `NativeArray`.

### 🤔 Expected Behavior
`y` is typed as `Array`.

Note that this is true at runtime only for modern browsers. For old browsers, the current types are correct... see [here](https://github.com/emberjs/ember.js/blob/v4.8.0/packages/@ember/array/index.ts#L1941). In modern browsers, `map` comes from the browser. It returns a native JS array. Otherwise an Ember version of [map](https://github.com/emberjs/ember.js/blob/v4.8.0/packages/@ember/array/index.ts#L721) is used which actually returns an `EmberArray`. I believe that should be fixed and the types should be changed to return native JS arrays.

Same applies for `mapBy` (which uses `map`), `filter`, `filterBy` and perhaps others.

### 🌍 Environment

- Ember: - 4.8.0

cc @chriskrycho, @wagenet

wagenet:
This PR does a couple of things to improve array types:
* More fully unifies internal array types with external preview types.
* Corrects NativeArray types to not override methods from the true native JS array.

Fixes #20230

### DIFF
--- a/packages/@ember/-internals/utils/types.ts
+++ b/packages/@ember/-internals/utils/types.ts
@@ -1,10 +1,18 @@
 export type AnyFn = (...args: any[]) => any;
 
-export type MethodNamesOf<T> = {
-  [K in keyof T]: T[K] extends AnyFn ? K : never;
-}[keyof T];
+// The formatting here is designed to help make this type actually be
+// comprehensible to mortals, including the mortals who came up with it.
+// prettier-ignore
+export type MethodsOf<T> = {
+  // This `keyof` check is the thing which gives us *only* these keys, and no
+  // `foo: never` appears in the final type.
+  [K in keyof T as T[K] extends AnyFn ? K : never]:
+    // While this makes sure the resolved type only has `AnyFn` in it, so that
+    // the resulting type is known to be only function types.
+    T[K] extends AnyFn ? T[K] : never;
+};
 
-export type MethodsOf<O> = Pick<O, MethodNamesOf<O>>;
+export type MethodNamesOf<T> = keyof MethodsOf<T>;
 
 export type MethodParams<T, M extends MethodNamesOf<T>> = Parameters<MethodsOf<T>[M]>;
 
--- a/packages/@ember/-internals/utils/types.ts
+++ b/packages/@ember/-internals/utils/types.ts
@@ -4,4 +4,10 @@ export type MethodNamesOf<T> = {
   [K in keyof T]: T[K] extends AnyFn ? K : never;
 }[keyof T];
 
+export type MethodsOf<O> = Pick<O, MethodNamesOf<O>>;
+
+export type MethodParams<T, M extends MethodNamesOf<T>> = Parameters<MethodsOf<T>[M]>;
+
+export type MethodReturns<T, M extends MethodNamesOf<T>> = ReturnType<MethodsOf<T>[M]>;
+
 export type OmitFirst<F> = F extends [any, ...infer R] ? R : [];
--- a/packages/@ember/array/index.ts
+++ b/packages/@ember/array/index.ts
@@ -20,12 +20,12 @@ import MutableEnumerable from '@ember/enumerable/mutable';
 import { compare, typeOf } from '@ember/utils';
 import { ENV } from '@ember/-internals/environment';
 import Observable from '@ember/object/observable';
-import type { AnyFn } from '@ember/-internals/utils/types';
+import type { MethodNamesOf, MethodParams, MethodReturns } from '@ember/-internals/utils/types';
 import type { ComputedPropertyCallback } from '@ember/-internals/metal';
 
 export { makeArray } from '@ember/-internals/utils';
 
-type Value<T, K extends string> = K extends keyof T ? T[K] : unknown;
+export type EmberArrayLike<T> = EmberArray<T> | NativeArray<T>;
 
 const EMPTY_ARRAY = Object.freeze([] as const);
 
@@ -319,7 +319,8 @@ interface EmberArray<T> extends Enumerable {
     @return this
     @public
   */
-  '[]': this;
+  get '[]'(): this;
+  set '[]'(newValue: T[] | EmberArray<T>);
   /**
     The first object in the array, or `undefined` if the array is empty.
 
@@ -515,7 +516,7 @@ interface EmberArray<T> extends Enumerable {
     @return {Array} The mapped array.
     @public
   */
-  getEach<K extends string>(key: K): NativeArray<Value<T, K>>;
+  getEach<K extends keyof T>(key: K): NativeArray<T[K]>;
   /**
     Sets the value on the named property for each member. This is more
     ergonomic than using other methods defined on this helper. If the object
@@ -535,7 +536,7 @@ interface EmberArray<T> extends Enumerable {
     @return {Object} receiver
     @public
   */
-  setEach<K extends string>(key: K, value: Value<T, K>): this;
+  setEach<K extends keyof T>(key: K, value: T[K]): this;
   /**
     Maps all of the items in the enumeration to another value, returning
     a new array. This method corresponds to `map()` defined in JavaScript 1.6.
@@ -593,7 +594,8 @@ interface EmberArray<T> extends Enumerable {
     @return {Array} The mapped array.
     @public
   */
-  mapBy<K extends string>(key: K): NativeArray<Value<T, K>>;
+  mapBy<K extends keyof T>(key: K): NativeArray<T[K]>;
+  mapBy(key: string): NativeArray<unknown>;
   /**
     Returns a new array with all of the items in the enumeration that the provided
     callback function returns true for. This method corresponds to [Array.prototype.filter()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter).
@@ -746,10 +748,6 @@ interface EmberArray<T> extends Enumerable {
     @public
   */
   rejectBy(key: string, value?: unknown): NativeArray<T>;
-  find<S extends T, Target = void>(
-    predicate: (this: void, value: T, index: number, obj: T[]) => value is S,
-    thisArg?: Target
-  ): S | undefined;
   /**
     Returns the first item in the array for which the callback returns true.
     This method is similar to the `find()` method defined in ECMAScript 2015.
@@ -792,6 +790,10 @@ interface EmberArray<T> extends Enumerable {
     @return {Object} Found item or `undefined`.
     @public
   */
+  find<S extends T, Target = void>(
+    predicate: (this: void, value: T, index: number, obj: T[]) => value is S,
+    thisArg?: Target
+  ): S | undefined;
   find<Target = void>(
     callback: (this: Target, item: T, index: number, arr: this) => unknown,
     target?: Target
@@ -824,7 +826,8 @@ interface EmberArray<T> extends Enumerable {
     @return {Object} found item or `undefined`
     @public
   */
-  findBy<K extends string>(key: K, value?: Value<T, K>): T | undefined;
+  findBy<K extends keyof T>(key: K, value?: T[K]): T | undefined;
+  findBy(key: string, value?: unknown): T | undefined;
   /**
     Returns `true` if the passed function returns true for every item in the
     enumeration. This corresponds with the `Array.prototype.every()` method defined in ES5.
@@ -904,7 +907,8 @@ interface EmberArray<T> extends Enumerable {
     @since 1.3.0
     @public
   */
-  isEvery<K extends string>(key: K, value?: Value<T, K>): boolean;
+  isEvery<K extends keyof T>(key: K, value?: T[K]): boolean;
+  isEvery(key: string, value?: unknown): boolean;
   /**
     The any() method executes the callback function once for each element
     present in the array until it finds the one where callback returns a truthy
@@ -972,7 +976,8 @@ interface EmberArray<T> extends Enumerable {
     @since 1.3.0
     @public
   */
-  isAny<K extends string>(key: K, value?: Value<T, K>): boolean;
+  isAny<K extends keyof T>(key: K, value?: T[K]): boolean;
+  isAny(key: string, value?: unknown): boolean;
   /**
     This will combine the values of the array into a single value. It
     is a useful way to collect a summary value from an array. This
@@ -1061,10 +1066,10 @@ interface EmberArray<T> extends Enumerable {
     @return {Array} return values from calling invoke.
     @public
   */
-  invoke<K extends string>(
-    methodName: K,
-    ...args: Value<T, K> extends AnyFn ? Parameters<Value<T, K>> : unknown[]
-  ): NativeArray<Value<T, K> extends AnyFn ? ReturnType<Value<T, K>> : unknown>;
+  invoke<M extends MethodNamesOf<T>>(
+    methodName: M,
+    ...args: MethodParams<T, M>
+  ): NativeArray<MethodReturns<T, M>>;
   /**
     Simply converts the object into a genuine array. The order is not
     guaranteed. Corresponds to the method implemented by Prototype.
@@ -1086,7 +1091,7 @@ interface EmberArray<T> extends Enumerable {
     @return {Array} the array without null and undefined elements.
     @public
   */
-  compact(): NativeArray<Exclude<T, null>>;
+  compact(): NativeArray<NonNullable<T>>;
   /**
     Used to determine if the array contains the passed object.
     Returns `true` if found, `false` otherwise.
@@ -1176,6 +1181,7 @@ interface EmberArray<T> extends Enumerable {
     @public
   */
   uniqBy(key: string): NativeArray<T>;
+  uniqBy(callback: (value: T) => unknown): NativeArray<T>;
   /**
     Returns a new array that excludes the passed value. The default
     implementation returns an array regardless of the receiver type.
@@ -1567,7 +1573,7 @@ interface MutableArray<T> extends EmberArray<T>, MutableEnumerable {
     @return object same object passed as a param
     @public
   */
-  pushObject(obj: T): this;
+  pushObject(obj: T): T;
   /**
     Add the objects in the passed array to the end of the array. Defers
     notifying observers of the change until all objects are added.
@@ -1599,7 +1605,7 @@ interface MutableArray<T> extends EmberArray<T>, MutableEnumerable {
     @return object
     @public
   */
-  popObject(): T | undefined;
+  popObject(): T | null | undefined;
   /**
     Shift an object from start of array or nil if none are left. Works just
     like `shift()` but it is KVO-compliant.
@@ -1632,7 +1638,7 @@ interface MutableArray<T> extends EmberArray<T>, MutableEnumerable {
     @return object same object passed as a param
     @public
   */
-  unshiftObject(object: T): this;
+  unshiftObject(object: T): T;
   /**
     Adds the named objects to the beginning of the array. Defers notifying
     observers until all objects have been added.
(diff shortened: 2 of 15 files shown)


chriskrycho:
Top-level request (I'll review further later): can you add an explanation to the PR so we can refer to it later to understand *why* these specific changes are correct?

packages/@ember/-internals/utils/types.ts
```diff
@@ -1,7 +1,15 @@
 export type AnyFn = (...args: any[]) => any;
 
+export type MethodsOf<O> = {
+  [K in keyof O]: O[K] extends AnyFn ? O[K] : never;
+};
```
chriskrycho, line 5: This doesn't actually work, unfortunately. You need something like the final `keyof` as used in `MethodNamesOf` below to make it properly strip out the bits which are actually `never`. It turns out the easiest way to get `MethodsOf` is to reuse `MethodNamesOf` in conjunction with `Pick`:

```ts
type MethodsOf<O> = Pick<O, MethodNamesOf<O>>;
```

[playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAggdiAYnKBeKAKAdDghgJwHMBnALilwQG0BdASjQD4KEBuAKHYgA8wB7fMCihIUALIRgACz4ATYgHkAZgB4FzdAG92UKFQDSUAJYoA1hBB8lUBTXIKDNKD2AQ482AmRQA-DcdQ5HAQAG4Q+BwAvhzsItAAQpKu+BLScoqq6mhQAApGAMamagA04pIysgByuAC2EBlqjIwxPPyCwuDQqRXVdQ0AKhpQ2rp6hiZQ5pbW-XZQ-QEubh7wSCh+hkGh4RxQkVRTVvM0MQD0p1AAVgCuxELEfFCW11D5lFDEENAyAO7GQj8BKZiFAfkZpMIpNAeLUwAAbepQABGEDhfB+7FkEHycII0BxuGIILEuBAKJUg2GOlGIVwcOuEB85H6u1GNVwYBUAFVGBglHByBhgMyGKhmFy6OQSWSINzmuxIpw4lAABrZbpyXr1ZQqaXk65wUxwdFwJocZUATXV5XSOr1soNRpNZtinSgAC1solgMkNfI7aT9YbjT9Tc0gA)

types/preview/@ember/array/mutable.d.ts
```diff
@@ -15,7 +15,7 @@ declare module '@ember/array/mutable' {
     /**
      * __Required.__ You must implement this method to apply this mixin.
      */
-    replace(idx: number, amt: number, objects: T[]): this;
+    replace(idx: number, amt: number, objects: T[]): void;
```
wagenet, line 18: In some cases we do return `this`, but the docs state that we do not return.

packages/@ember/array/index.ts
```diff
@@ -515,7 +519,8 @@ interface EmberArray<T> extends Enumerable {
     @return {Array} The mapped array.
     @public
   */
-  getEach<K extends string>(key: K): NativeArray<Value<T, K>>;
+  getEach<K extends keyof T>(key: K): NativeArray<T[K]>;
+  getEach<K extends CompoundPropertyKey<T>>(key: K): NativeArray<unknown>;
```
chriskrycho, line 523: Instead of introducing additional type shenanigans here, since it ultimately just becomes `NativeArray<unknown>`, we can just use `string`. (This is how we do it with the overloads in Ember and Ember Data).

```suggestion
  getEach(key: string): NativeArray<unknown>;
```

packages/@ember/array/index.ts
```diff
@@ -20,12 +20,15 @@ import MutableEnumerable from '@ember/enumerable/mutable';
...
+import type { MethodNamesOf, MethodParams, MethodReturns } from '@ember/-internals/utils/types';
 import type { ComputedPropertyCallback } from '@ember/-internals/metal';
 
 export { makeArray } from '@ember/-internals/utils';
 
-type Value<T, K extends string> = K extends keyof T ? T[K] : unknown;
+export type EmberArrayLike<T> = EmberArray<T> | NativeArray<T>;
+
+// We don't currently attempt to fully type-check compound keys
+type CompoundPropertyKey<T> = `${keyof T & string}.${string}`;
```
chriskrycho, line 31: Then we don't need this!

```suggestion
```

packages/@ember/array/index.ts
```diff
@@ -746,10 +752,6 @@ interface EmberArray<T> extends Enumerable {
     @public
   */
   rejectBy(key: string, value?: unknown): NativeArray<T>;
-  find<S extends T, Target = void>(
-    predicate: (this: void, value: T, index: number, obj: T[]) => value is S,
-    thisArg?: Target
-  ): S | undefined;
```
wagenet, line 752: Just moved this.

packages/@ember/array/index.ts
```diff
@@ -1061,10 +1070,10 @@ interface EmberArray<T> extends Enumerable {
...
     @public
   */
-  invoke<K extends string>(
-    methodName: K,
-    ...args: Value<T, K> extends AnyFn ? Parameters<Value<T, K>> : unknown[]
-  ): NativeArray<Value<T, K> extends AnyFn ? ReturnType<Value<T, K>> : unknown>;
+  invoke<M extends MethodNamesOf<T>>(
+    methodName: M,
+    ...args: MethodParams<T, M>
+  ): NativeArray<MethodReturns<T, M>>;
```
wagenet, line 1076: The public preview types were better here.

packages/@ember/array/index.ts
```diff
@@ -1086,7 +1095,7 @@ interface EmberArray<T> extends Enumerable {
     @return {Array} the array without null and undefined elements.
     @public
   */
-  compact(): NativeArray<Exclude<T, null>>;
+  compact(): NativeArray<NonNullable<T>>;
```
wagenet, line 1098: The public preview types were better here.

packages/@ember/array/index.ts
```diff
@@ -1176,6 +1185,7 @@ interface EmberArray<T> extends Enumerable {
     @public
   */
   uniqBy(key: string): NativeArray<T>;
+  uniqBy(callback: (value: T) => unknown): NativeArray<T>;
```
wagenet, line 1188: This was missing before.

packages/@ember/array/index.ts
```diff
@@ -1632,7 +1642,7 @@ interface MutableArray<T> extends EmberArray<T>, MutableEnumerable {
     @return object same object passed as a param
     @public
   */
-  unshiftObject(object: T): this;
+  unshiftObject(object: T): T;
```
wagenet, line 1645: This was incorrect before.

packages/@ember/array/index.ts
```diff
@@ -1892,6 +1902,167 @@ const MutableArray = Mixin.create(EmberArray, MutableEnumerable, {
...
+  | 'setObjects'
+  | 'removeObject'
+  | 'removeObjects'
+  | 'addObject'
+  | 'addObjects'
+  | 'setEach';
+
+// This is the same as MutableArray, but removes the actual native methods that exist on Array.prototype.
+interface MutableArrayWithoutNative<T>
+  extends Omit<MutableArray<T>, IGNORED_MUTABLE_ARRAY_METHODS | RETURN_SELF_ARRAY_METHODS> {
```
wagenet, line 1946: NativeArray is MutableArray with all actual native methods removed. Unfortunately, because of how Omit works, all methods that return `this` will end up not returning the correct `this`, so we also have to redefine those methods here.

packages/@ember/array/type-tests/index.ts
```diff
@@ -37,9 +40,9 @@ expectTypeOf(arr.firstObject).toEqualTypeOf<Foo | undefined>();
 
 expectTypeOf(arr.lastObject).toEqualTypeOf<Foo | undefined>();
 
-expectTypeOf(arr.slice()).toEqualTypeOf<NativeArray<Foo>>();
-expectTypeOf(arr.slice(1)).toEqualTypeOf<NativeArray<Foo>>();
-expectTypeOf(arr.slice(1, 2)).toEqualTypeOf<NativeArray<Foo>>();
+expectTypeOf(arr.slice()).toEqualTypeOf<Foo[]>();
+expectTypeOf(arr.slice(1)).toEqualTypeOf<Foo[]>();
+expectTypeOf(arr.slice(1, 2)).toEqualTypeOf<Foo[]>();
```
wagenet, line 45: We're actually hitting the native slice.

packages/@ember/array/type-tests/index.ts
```diff
@@ -49,7 +52,7 @@ expectTypeOf(arr.lastIndexOf(new Foo())).toEqualTypeOf<number>();
 // @ts-expect-error checks param type
 arr.lastIndexOf('invalid');
 
-expectTypeOf(arr.forEach((item: Foo) => String(item))).toEqualTypeOf(arr);
+expectTypeOf(arr.forEach((item: Foo) => String(item))).toEqualTypeOf<void>();
```
wagenet, line 55: Native forEach returns void.

packages/@ember/array/type-tests/index.ts
```diff
@@ -58,43 +61,44 @@ arr.forEach((item, index, arr) => {
   expectTypeOf(arr).toEqualTypeOf(arr);
 });
 arr.forEach(function (_item) {
-  expectTypeOf(this).toEqualTypeOf(target);
```
wagenet, line 61: Native methods behave differently here.

packages/@ember/array/type-tests/index.ts
```diff
@@ -58,43 +61,44 @@ arr.forEach((item, index, arr) => {
...
+  // No-op
 }, target);
 
 expectTypeOf(arr.getEach('bar')).toEqualTypeOf<NativeArray<number>>();
-expectTypeOf(arr.getEach('missing')).toEqualTypeOf<NativeArray<unknown>>();
+// @ts-expect-error property is unknown
+arr.getEach('missing');
```
wagenet, line 69: We updated types to not allow this case.

packages/@ember/array/type-tests/index.ts
```diff
@@ -58,43 +61,44 @@ arr.forEach((item, index, arr) => {
...
 
 expectTypeOf(arr.setEach('bar', 2)).toEqualTypeOf(arr);
 // @ts-expect-error string is not assignable to bar
 arr.setEach('bar', 'string');
 
-// Can set unknown property
-expectTypeOf(arr.setEach('missing', 'anything')).toEqualTypeOf(arr);
+// @ts-expect-error property is unknown
+arr.setEach('missing', 'anything');
```
wagenet, line 76: We updated types to not allow this.

packages/@ember/array/type-tests/index.ts
```diff
@@ -166,16 +172,14 @@ let reduced = arr.reduce((summation, item, index, arr) => {
   return 1;
 }, 1);
 expectTypeOf(reduced).toEqualTypeOf<number>();
-// NOTE: This doesn't match native behavior and is a bit weird
-expectTypeOf(arr.reduce((summation, _item) => summation)).toEqualTypeOf<unknown>();
+expectTypeOf(arr.reduce((summation, _item) => summation)).toEqualTypeOf<Foo>();
```
wagenet, line 175: This does match native types now.

packages/@ember/array/type-tests/mutable.ts
```diff
@@ -1,63 +1,67 @@
-import type MutableArray from '@ember/array/mutable';
-import { A } from '@ember/array';
+import MutableArray from '@ember/array/mutable';
 
 import { expectTypeOf } from 'expect-type';
 
-class Foo {}
+class Foo {
+  constructor(public name: string) {}
```
wagenet, line 6: This needed a property for the checks below to fully work.

packages/@ember/array/index.ts
```diff
@@ -593,7 +598,8 @@ interface EmberArray<T> extends Enumerable {
     @return {Array} The mapped array.
     @public
   */
-  mapBy<K extends string>(key: K): NativeArray<Value<T, K>>;
+  mapBy<K extends keyof T>(key: K): NativeArray<T[K]>;
+  mapBy<K extends CompoundPropertyKey<T>>(key: K): NativeArray<unknown>;
```
chriskrycho, line 602:
```suggestion
  mapBy(key: string): NativeArray<unknown>;
```

packages/@ember/array/index.ts
```diff
@@ -824,7 +830,8 @@ interface EmberArray<T> extends Enumerable {
     @return {Object} found item or `undefined`
     @public
   */
-  findBy<K extends string>(key: K, value?: Value<T, K>): T | undefined;
+  findBy<K extends keyof T>(key: K, value?: T[K]): T | undefined;
+  findBy<K extends CompoundPropertyKey<T>>(key: K, value?: unknown): T | undefined;
```
chriskrycho, line 834:
```suggestion
  findBy(key: string, value?: unknown): T | undefined;
```

packages/@ember/array/index.ts
```diff
@@ -904,7 +911,8 @@ interface EmberArray<T> extends Enumerable {
     @since 1.3.0
     @public
   */
-  isEvery<K extends string>(key: K, value?: Value<T, K>): boolean;
+  isEvery<K extends keyof T>(key: K, value?: T[K]): boolean;
+  isEvery<K extends CompoundPropertyKey<T>>(key: K, value?: unknown): boolean;
```
chriskrycho, line 915:
```suggestion
  isEvery(key: string, value?: unknown): boolean;
```

packages/@ember/array/index.ts
```diff
@@ -972,7 +980,8 @@ interface EmberArray<T> extends Enumerable {
     @since 1.3.0
     @public
   */
-  isAny<K extends string>(key: K, value?: Value<T, K>): boolean;
+  isAny<K extends keyof T>(key: K, value?: T[K]): boolean;
+  isAny<K extends CompoundPropertyKey<T>>(key: K, value?: unknown): boolean;
```
chriskrycho, line 984:
```suggestion
  isAny(key: string, value?: unknown): boolean;
```

packages/@ember/debug/container-debug-adapter.ts
```diff
@@ -88,9 +88,9 @@ export default class ContainerDebugAdapter extends EmberObject {
     @return {Array} An array of strings.
     @public
   */
-  catalogEntriesByType(type: string): NativeArray<string> {
-    let namespaces = emberA(Namespace.NAMESPACES);
-    let types = emberA<string>();
+  catalogEntriesByType(type: string): string[] {
+    let namespaces = Namespace.NAMESPACES;
+    let types: string[] = [];
```
wagenet, line 93: This is just documented as returning an array.

packages/@ember/debug/data-adapter.ts
```diff
@@ -544,22 +544,22 @@ export default class DataAdapter<T> extends EmberObject {
...
+  getModelTypes(): Array<{ klass: unknown; name: string }> {
     let containerDebugAdapter = this.containerDebugAdapter;
 
     let stringTypes = containerDebugAdapter.canCatalogEntriesByType('model')
       ? containerDebugAdapter.catalogEntriesByType('model')
       : this._getObjectsOnNamespaces();
 
     // New adapters return strings instead of classes.
-    let klassTypes = emberA(stringTypes).map((name) => {
+    let klassTypes = stringTypes.map((name) => {
```
wagenet, line 555: These uses of `emberA` were unnecessary since `map` and `filter` are native methods.

packages/@ember/debug/data-adapter.ts
```diff
@@ -544,22 +544,22 @@ export default class DataAdapter<T> extends EmberObject {
     @method getModelTypes
     @return {Array} Array of model types.
   */
-  getModelTypes(): NativeArray<{ klass: unknown; name: string }> {
+  getModelTypes(): Array<{ klass: unknown; name: string }> {
```
wagenet, line 547: This is just documented as returning an array.

type-tests/preview/@ember/array-test/array.ts
```diff
@@ -26,13 +27,11 @@ expectTypeOf(people.objectAt(0)).toEqualTypeOf<Person | undefined>();
 expectTypeOf(people.objectsAt([1, 2, 3])).toEqualTypeOf<Array<Person | undefined>>();
 
 expectTypeOf(people.filterBy('isHappy')).toMatchTypeOf<Person[]>();
-expectTypeOf(people.filterBy('isHappy')).toMatchTypeOf<MutableArray<Person>>();
+expectTypeOf(people.filterBy('isHappy')).toMatchTypeOf<NativeArray<Person>>();
```
wagenet, line 30: NativeArray !== MutableArray

type-tests/preview/ember/ember-module-tests.ts
```diff
@@ -153,7 +153,7 @@ expectTypeOf(Ember.Application.create()).toEqualTypeOf<Ember.Application>();
 expectTypeOf(new Ember.ApplicationInstance()).toEqualTypeOf<Ember.ApplicationInstance>();
 expectTypeOf(Ember.ApplicationInstance.create()).toEqualTypeOf<Ember.ApplicationInstance>();
 // Ember.Array
-const a1: Ember.Array<string> = Ember.A([]);
+const a1: Ember.NativeArray<string> = Ember.A([]);
```
wagenet, line 156: NativeArray !== EmberArray

types/preview/@ember/array/-private/native-array.d.ts
```diff
@@ -1,22 +1,180 @@
...
   // Get an alias to the global Array type to use in inner scope below.
   type Array<T> = T[];
 
+  type AnyArray<T> = EmberArray<T> | Array<T>;
+
+  /**
+   * The final definition of NativeArray removes all native methods. This is the list of removed methods
+   * when run in Chrome 106.
+   */
+  type IGNORED_MUTABLE_ARRAY_METHODS =
```
wagenet, line 16: The below should be the same as internal.

types/preview/@ember/array/index.d.ts
```diff
@@ -4,6 +4,9 @@ declare module '@ember/array' {
   import Enumerable from '@ember/array/-private/enumerable';
   import NativeArray from '@ember/array/-private/native-array';
 
+  // We don't currently attempt to fully type-check compound keys
+  type CompoundPropertyKey<T> = `${keyof T & string}.${string}`;
```
wagenet, line 8: Bring these more in line with internal types.

types/preview/@ember/array/mutable.d.ts
```diff
@@ -45,12 +45,12 @@ declare module '@ember/array/mutable' {
      * Pop object from array or nil if none are left. Works just like `pop()` but
      * it is KVO-compliant.
      */
-    popObject(): T;
+    popObject(): T | null | undefined;
```
wagenet, line 48: These types were incorrect before.

packages/@ember/array/index.ts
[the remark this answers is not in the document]
chriskrycho: Ah, so this catches the case where we have `get(anArray, 'lol-nope')` and makes it not type check. In general, we haven’t tried to do that: people *do* (and from a public API can :upside_down:) stash unknown stuff on instances with `set` and then later `get` it, which should be "safe" (if not recommended).

packages/@ember/array/index.ts
wagenet: Resolved

packages/@ember/array/index.ts
wagenet: Resolved.

packages/@ember/array/index.ts
```diff
@@ -792,6 +790,10 @@ interface EmberArray<T> extends Enumerable {
     @return {Object} Found item or `undefined`.
     @public
   */
+  find<S extends T, Target = void>(
```
chriskrycho, line 793: Note: conventionally/idiomatically it's `U` in a case like this (since T → U in the alphabet). Not a thing I care about changing, to be sure, just worth noting!

packages/@ember/array/index.ts
```diff
@@ -1892,6 +1898,167 @@ const MutableArray = Mixin.create(EmberArray, MutableEnumerable, {
 /**
 @module ember
 */
+
+type AnyArray<T> = EmberArray<T> | Array<T>;
+
+/**
+ * The final definition of NativeArray removes all native methods. This is the list of removed methods
+ * when run in Chrome 106.
+ */
```
chriskrycho, line 1907: Note for posterity/future readers looking at this PR historically—we definitely need this, but we need even more to just get the heck rid of it, since this could in principle change dynamically over time.

packages/@ember/array/type-tests/mutable.ts
```diff
@@ -1,63 +1,67 @@
...
+let originalArr = [foo];
+// This is not really the ideal way to set things up.
+MutableArray.apply(originalArr);
+let arr = originalArr as unknown as MutableArray<Foo>;
 
 expectTypeOf(arr).toMatchTypeOf<MutableArray<Foo>>();
 
 expectTypeOf(arr.replace(1, 1, [foo])).toEqualTypeOf<void>();
-// TODO: Why doesn't this fail?
+// @ts-expect-error invalid item
```
chriskrycho, line 19: It's nice that all of these fail now the way we want them to. 👍🏼

packages/@ember/debug/container-debug-adapter.ts
```diff
@@ -88,9 +86,9 @@ export default class ContainerDebugAdapter extends EmberObject {
...
```
chriskrycho, line 91: Hmm. Documentation is very apt to be incorrect in this space, though; folks who wrote the docs likely were not thinking about the differences (because the differences are subtle!). Let's avoid changing the behavior here—we can always do so in a later PR, but keeping types changes separate from behavior changes will help us avoid snafus.

packages/@ember/debug/data-adapter.ts
chriskrycho: On the one hand, this is *definitely* a change I'm excited to be able to make; on the other, my note here is the same as above. Let's do it in a separate PR so it's easy to trace back to if for some reason it ends up being an issue for folks.

packages/@ember/debug/data-adapter.ts
```diff
@@ -571,8 +571,8 @@ export default class DataAdapter<T> extends EmberObject {
     @return {Array} Array of model type strings.
   */
   _getObjectsOnNamespaces() {
-    let namespaces = emberA(Namespace.NAMESPACES);
-    let types = emberA<string>();
+    let namespaces = Namespace.NAMESPACES;
+    let types: string[] = [];
```
chriskrycho, line 575: Same here, presumably.

chriskrycho (changes requested):
Couple of tweaks needed, but this is a really nice step forward. 💙

packages/@ember/array/index.ts
wagenet: I think I just copied this from somewhere else :)

packages/@ember/array/type-tests/mutable.ts
wagenet: I think we just needed a better test item :)

chriskrycho (approved):

PR #20260, opened 2022-11-09
Refactor overrides.js

azhar-beg:
- Replaced `requirementFor` function to arrow function
- Replaced `ef-else` blog to `ternary`, static method `for`.
- Replaced for loop with array method reduce, `printList`.
- Fixed spellings.

### DIFF
--- a/lib/overrides.js
+++ b/lib/overrides.js
@@ -13,11 +13,9 @@ function* walkAddonTree(project, pathToAddon = []) {
   }
 }
 
-function requirementFor(pkg, deps = {}) {
-  return deps[pkg];
-}
+const requirementFor = (pkg, deps = {}) => deps[pkg];
 
-const KNWON_DORMANT_ADDONS = Object.freeze([
+const KNOWN_DORMANT_ADDONS = Object.freeze([
   'ember-angle-bracket-invocation-polyfill',
   'ember-fn-helper-polyfill',
   'ember-in-element-polyfill',
@@ -26,13 +24,11 @@ const KNWON_DORMANT_ADDONS = Object.freeze([
 ]);
 module.exports = class Overrides {
   static for(project, env = process.env) {
-    if (env.EMBER_ENV === 'production') {
-      return new Overrides([]);
-    } else {
-      return new Overrides(Overrides.addonsInfoFor(project), {
-        showAllEmberGlobalDeprecations: env.EMBER_GLOBAL_DEPRECATIONS === 'all',
-      });
-    }
+    return env.EMBER_ENV === 'production'
+      ? new Overrides([])
+      : new Overrides(Overrides.addonsInfoFor(project), {
+          showAllEmberGlobalDeprecations: env.EMBER_GLOBAL_DEPRECATIONS === 'all',
+        });
   }
 
   static *addonsInfoFor(project) {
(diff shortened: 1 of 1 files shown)


azhar-beg:
Refactored overrides.js

lib/overrides.js
```diff
@@ -88,18 +84,13 @@ module.exports = class Overrides {
...
+    return list.reduce((listString, listItem) => {
+      if (Array.isArray(listItem)) {
+        listString += `${indent}* ${listItem[0]}\n`;
+        listString += Overrides.printList(listItem[1], indent + '  ');
       }
-    }
-
-    return output;
+      listString += `${indent}* ${listItem}\n`;
+    }, '');
```
chriskrycho, line 93: While I personally enjoy a good `reduce`, I think we're better leaving this code as is—if it's not specifically fixing a bug, we're just changing good working code for aesthetic reasons.

lib/overrides.js
```diff
@@ -26,13 +24,11 @@ const KNWON_DORMANT_ADDONS = Object.freeze([
...
-    } else {
-      return new Overrides(Overrides.addonsInfoFor(project), {
-        showAllEmberGlobalDeprecations: env.EMBER_GLOBAL_DEPRECATIONS === 'all',
-      });
-    }
+    return env.EMBER_ENV === 'production'
+      ? new Overrides([])
+      : new Overrides(Overrides.addonsInfoFor(project), {
+          showAllEmberGlobalDeprecations: env.EMBER_GLOBAL_DEPRECATIONS === 'all',
+        });
```
chriskrycho, line 31: Similarly, while I'm a fan of ternaries and expression-oriented programming, this isn't actually an *improvement* in a strict sense. It's just *different*. Better to leave it unchanged!

lib/overrides.js
```diff
@@ -13,11 +13,9 @@ function* walkAddonTree(project, pathToAddon = []) {
   }
 }
 
-function requirementFor(pkg, deps = {}) {
-  return deps[pkg];
-}
+const requirementFor = (pkg, deps = {}) => deps[pkg];
```
chriskrycho, line 16: The same goes for this—the arrow function is cleaner-looking, but it does actually have different semantics. (It can't be referred to *earlier* in the file, now, since it isn't hoisted!)

lib/overrides.js
```diff
@@ -13,11 +13,9 @@ function* walkAddonTree(project, pathToAddon = []) {
...
 
-const KNWON_DORMANT_ADDONS = Object.freeze([
+const KNOWN_DORMANT_ADDONS = Object.freeze([
```
chriskrycho, line 18: Big fan of the spelling fixes, though!

chriskrycho (changes requested):
Thanks for the desire to help! Let's avoid the behavioral changes but keep the spelling changes. Thanks!

azhar-beg:
Thanks for the feedback. I will update the changes.

lib/overrides.js
praveeno: Thanks dude for be so cool about these changes, and being helpful by helping people to understand what's more important and what's not, we need more people like you in open source,

chriskrycho (approved):

chriskrycho:
Thanks!

PR #20268, opened 2022-11-16
[CLEANUP] Make `intern` utility type safe

chriskrycho:
Experimenting with removing/improving some *other* types while implementing RFC 0821. revealed that this function needed to preserve its input type when that type was narrower than just `string`, so introduce a type parameter to keep that information around. Then, switch to using `Object.create(null)` so it can be used in the `Record` type safely.

Pulling this out into its own commit makes it easier to land, and it also allows us to guarantee that we are making changes *only* to the types and module layout for the PR implementing RFC 0821.

### DIFF
--- a/packages/@ember/-internals/utils/lib/intern.ts
+++ b/packages/@ember/-internals/utils/lib/intern.ts
@@ -37,8 +37,8 @@
   @private
   @return {String} interned version of the provided string
 */
-export default function intern(str: string): string {
-  let obj: Record<string, number> = {};
+export default function intern<S extends string>(str: S): S {
+  let obj: Record<S, number> = Object.create(null);
   obj[str] = 1;
   for (let key in obj) {
     if (key === str) {


packages/@ember/-internals/utils/lib/intern.ts
```diff
@@ -37,8 +37,8 @@
   @private
   @return {String} interned version of the provided string
 */
-export default function intern(str: string): string {
-  let obj: Record<string, number> = {};
+export default function intern<S extends string>(str: S): S {
+  let obj: Record<S, number> = Object.create(null);
```
jamescdavis, line 41: To keep this equivalent to `{}`:
```suggestion
  let obj: Record<S, number> = Object.create(Object.prototype);
```
`Object.create(null)` will have _no_ prototype, but `{}` has `Object`'s prototype:

Screen Shot 2022-11-15 at 7 39 21 PM

jamescdavis: Or just cast? ¯\\\_(ツ)\_/¯
```suggestion
  let obj = {} as Record<S, number>;
```

chriskrycho: Yeah, no prototype is actually fine here. I ran this it @rwjblue and the behavioral change here is totally fine; I just wanted it in a separate PR since it is a behavioral and not purely types change.

jamescdavis (approved):

PR #20271, opened 2022-11-17
[FEATURE] Add new `@ember/owner` package (RFC 0821)

chriskrycho:
Introduce the `@ember/owner` package, as specified in [RFC 0821: API for Type-Only Imports](https://rfcs.emberjs.com/id/0821-public-types/).

This involves a number of significant refactors, as well as a fair bit of cleanup along the way. For details; see the commits and their messages, which are fairly extensive and which I took quite a bit of time to get into a usable form for posterity. At a high level, the approach here is as follows:

**First,** the `@ember/-internals/owner` package defines both `Owner`, for the public API, and an `InternalOwner` type, which includes all of the types of `Owner` and more besides. To avoid circularity, `@ember/-internals/owner` also defines the core types for `ContainerProxy` and `RegistryProxy`, which together define Ember's actual notion of an owner in practice. (The "owner" instances Ember users interact with today are `EngineInstance` or its subclass `ApplicationInstance`; `EngineInstance` is created by applying `ContainerProxyMixin` and `RegistryProxyMixin`.)

The resulting type hierarchy looks like this (and, let us be clear: this is is a significant *improvement* because there are no module graph cycles now!):

```mermaid
classDiagram
  BasicRegistry <|-- Owner
  BasicRegistry <|-- RegistryProxy

  BasicContainer <|-- Owner
  BasicContainer <|-- ContainerProxy

  RegistryProxy <|-- InternalOwner
  ContainerProxy <|-- InternalOwner

  RegistryProxy <|.. RegistryProxyMixin : implements
  ContainerProxy <|.. ContainerProxyMixin : implements

  Owner <|.. EngineInstance : implements
  InternalOwner <|.. EngineInstance : implements
  RegistryProxyMixin <|-- EngineInstance : applies
  ContainerProxyMixin <|-- EngineInstance : applies

  EngineInstance <|-- ApplicationInstance

  <<Interface>> BasicRegistry
  BasicRegistry : register()

  <<Interface>> BasicContainer
  BasicContainer : lookup()
  BasicContainer : factoryFor()

  <<Interface>> Owner

  <<Interface>> RegistryProxy
  RegistryProxy : resolveRegistration()
  RegistryProxy : unregister()
  RegistryProxy : hasRegistration()
  RegistryProxy : registeredOption()
  RegistryProxy : registeredOptions()
  RegistryProxy : registerOptionsForType()
  RegistryProxy : registeredOptionsForType()

  <<Interface>> ContainerProxy
  ContainerProxy : ownerInjection()

  <<Interface>> InternalOwner
```

For comparison, the API we *want* to have in the future, and indeed the *public* API we have committed to for `Owner`—and thus a significant motivation for this refactor:

```mermaid
classDiagram
  class Owner {
    register()
    lookup()
    factoryFor()
  }
```

Critically, with these changes, the types in `@ember/-internals/container` and `@ember/-internals/runtime` depend on those in `@ember/-internals/owner` *but not vice versa*. Additionally, many other `@ember/*` packages can now depend solely on `@ember/owner`, not needing to refer to `@ember/-internals` at all to refer to the `Owner` APIs.

**Second**, along the same lines, `@ember/-internals/owner` also becomes the home of both the public definitions of types like `Factory` and `FactoryManager` *and* the home of internal versions which extend them, and which Ember's internals rely on. The relations here are not so tangled, so I am not supplying a diagram to go with them; the point is that in the same way there are `Factory<T>` and `InternalFactory<T, C> extends Factory<T>` and so on.

**Finally**, with the internals appropriately refactored, the `@ember/owner` package re-exports `Owner` and the supporting/associated types, per the RFC.

With these changes in place, we are actually (at last!) in position to do two big things:

1. Publish types from Ember's source: this was the last known blocker other than some build scripting. (Watch for a future PR which does just that!)

2. Begin eliminating all the extraneous parts of the various owner APIs, via our normal deprecation policy: since `EngineInstance` already implements a superset of the `Owner` contract (and indeed explicitly `extends Owner`), we can progressively eliminate the duplicate APIs.


packages/@ember/-internals/owner/index.ts
```diff
@@ -222,37 +305,162 @@ export function setOwner(object: object, owner: Owner): void {
   glimmerSetOwner(object, owner);
 }
 
-export interface ContainerMixin extends Owner {
-  ownerInjection(): void;
+// Defines the type for the ContainerProxyMixin. When we rationalize our Owner
+// *not* to work via mixins, we will be able to delete this entirely: this
+// overload for `lookup()` and all of `ownerInjection()` will go away.
```
dfreeman, line 318: I don't see an overload for `lookup()` here—did that change after this comment was written?

packages/@ember/-internals/owner/index.ts
```diff
@@ -208,12 +218,18 @@ export interface Factory<T extends object> {
   // able to provide a *narrower* interface than "exactly the public fields on
   // the class" while still falling back to the "exactly the public fields on
   // the class" for the general case. :sigh:
+  //
+  // We stills upply both signatures because even though the second one means
```
dfreeman:
```suggestion
  // We still supply both signatures because even though the second one means
```

type-tests/preview/@ember/owner-tests.ts
```diff
@@ -35,19 +41,19 @@ aFactory.create({
...
-// @ts-expect-error
 aFactory.create({ hasProps: true, unrelatedNonsense: 'yep yep yep' });
 
 // But this should be legal.
 const goodPojo = { hasProps: true, unrelatedNonsense: 'also true' };
 aFactory.create(goodPojo);
 
-// while this should be rejected for *type error* reasons, not EPC
+// This should also be rejected, though for *type error* reasons, not EPC; alas,
+// it cannot, for the same reason.
```
dfreeman, line 55: ~Wasn't there a change in an earlier commit from something like `create(props: object)` to `create(props: Partial<T>)`? I would expect that to reject `hasProps: 'huzzah'` here.~

Nevermind, I see that got undone in 7cc146db09c5969349eefe6221c1a5ae603be644

dfreeman (commented):
This looks like a solid change to me—what little feedback I have is mostly just about comment text, I think. Thank you for the thorough descriptions and commit breakdown!

packages/@ember/owner/index.ts
```diff
@@ -31,7 +31,57 @@ interface GetOwner {
   (object: object): Owner | undefined;
 }
 
-const getOwner = internalGetOwner as GetOwner;
+// SAFETY: this is *only* safe for public API, because we are promising more
+// than we can actually *totally* guarantee we uphold otherwise. Specifically,
+// users *can* do things (including abusing the `ReliablyHasOwner` type or
```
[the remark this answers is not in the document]
chriskrycho: Nope, thank you – I thought I had deleted all of these!

packages/@ember/-internals/owner/index.ts
chriskrycho: It did, and I thought I had deleted it; thanks for the catch here as well!

packages/@ember/-internals/owner/index.ts
chriskrycho: EWWWWW, what did I write?!?

packages/@ember/owner/index.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  - `Factory`, Ember's primary interface for something which can create class
+    instances registered with the DI system.
+
+  - `FactoryManager`, an interface for inspecting a `Factory`'s class.
+
+  - `Resolver`, an interface defining the contract for the object responsible
+    for mapping string names to the corresponding classes. For example, when
+    you write `@service('session')`, a resolver is responsible to map that back
+    to the `Session` service class in your codebase. Normally
```
chriskrycho, line 20:
```suggestion
    to the `Session` service class in your codebase. Normally, this is handled
    for you automatically with `ember-resolver`, which is the main implementor
    of this interface.
```

packages/@ember/owner/index.ts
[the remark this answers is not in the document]
chriskrycho: Ah, it turns out I *had* deleted this; you caught this from an earlier commit. 😅

packages/@ember/-internals/owner/index.ts
chriskrycho, line 318:
```suggestion
// *not* to work via mixins, we will be able to delete this entirely, in favor
// of just using the Owner class itself.
```

packages/@ember/-internals/owner/index.ts
chriskrycho: Ah, but I deleted this separately because I couldn't make it work (as you found on a later commit).

packages/@ember/-internals/glimmer/lib/resolver.ts
```diff
@@ -55,24 +54,31 @@ function instrumentationPayload(name: string) {
   return { object: `component:${name}` };
 }
 
-function componentFor(name: string, owner: Owner): Option<Factory<object> | object> {
-  let fullName = `component:${name}`;
+function componentFor(
+  name: string,
+  owner: InternalOwner
+): Option<InternalFactory<object> | object> {
+  let fullName = `component:${name}` as const;
```
wagenet, line 61: What's the point of `as const` here?

packages/@ember/-internals/glimmer/lib/views/outlet.ts
```diff
@@ -39,7 +42,11 @@ export default class OutletView {
     Object.assign(this, injections);
   }
 
-  static create(options: any): OutletView {
+  static create(options: {
+    environment: BootEnvironment;
+    application: InternalOwner;
+    template: TemplateFactory;
+  }): OutletView {
```
wagenet, line 49: Goodbye `any`!

packages/@ember/application/index.ts
```diff
@@ -30,6 +30,16 @@ import RouterService from '@ember/routing/router-service';
 import type { EngineInstanceOptions } from '@ember/engine/instance';
 import type { SimpleDocument, SimpleElement } from '@simple-dom/interface';
 
+/**
+ * @deprecated Use `import { getOwner } from '@ember/owner';` instead.
```
wagenet, line 34: Should we runtime deprecate this function as well?

wagenet (commented):
Partial review.

packages/@ember/application/index.ts
chriskrycho: Yes, but I wanted to land that separately from this—both because it's safer and keeps the types/behavior separately, and also (perhaps more importantly) because we usually try not to turn on the deprecation in the same release as we put out the replacement for it, because otherwise people get hammered with deprecations they haven't had a chance to do fix, including from addons.

packages/@ember/-internals/glimmer/lib/resolver.ts
chriskrycho: Throughout, they all tell TS to notice the actual structure of the string, which in turn makes any manually-constructed string like this valid to use as a `FullName` as long as it has the right shape.

wagenet: Ah, got it!

chriskrycho:
All right, going to go ahead and merge this, and if/as folks catch issues we'll fix bugs!

packages/@ember/routing/router.ts
```diff
@@ -645,15 +651,38 @@ class EmberRouter<R extends Route = Route> extends EmberObject.extend(Evented) i
...
+      // TODO(SAFETY): At least one test runs without this set correctly. At a
+      // later time, update the test to configure this correctly. The test ID:
+      // `Router Service - non application test:  RouterService#transitionTo with basic route`
+      let instance = owner.lookup('-application-instance:main') as ApplicationInstance;
+      // let instance = owner.lookup('-application-instance:main') as ApplicationInstance | undefined;
+      // assert('[BUG] unexpectedly missing `-application-instance:main`', instance !== undefined);
+
       if (instance) {
-        instance.didCreateRootView(this._toplevelView);
+        // SAFETY: LOL. This is calling a deprecated API with a type that we
```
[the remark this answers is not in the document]
chriskrycho: Pretty much!

packages/@ember/engine/instance.ts
```diff
@@ -45,14 +47,22 @@ export interface EngineInstanceOptions {
...
   /**
    @private
    @method setupRegistry
    @param {Registry} registry
    @param {BootOptions} options
    */
+  // This is effectively an "abstract" method: it defines the contract a
+  // subclass (e.g. `ApplicationInstance`) must follow to implement this
+  // behavior, but an `EngineInstance` has no behavior of its own here.
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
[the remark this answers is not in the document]
chriskrycho: It does, and we probably should for things like this at least.

type-tests/preview/@ember/owner-tests.ts
```diff
@@ -35,19 +41,19 @@ aFactory.create({
   hasProps: false,
 });
 
-// These should be rejected by way of EPC
-// @ts-expect-error
+// NOTE: it would be nice if these could be rejected by way of EPC, but alas: it
```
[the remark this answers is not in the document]
chriskrycho: “Excess Property Checking”: the feature where if you pass an object *literal* as an argument with extra properties than the parameter type specifies, TS will error since that is basically guaranteed to e incorrect.

PR #20272, opened 2022-11-18
[DOC] Add documentation for RFC 0821 items

chriskrycho:
- Introduce or relocate documentation for all the `@ember/owner` types.
- Make sure all documented owner and routing types correctly express their type hierarchies in the docs so linking and inheritance flags both work correctly.
- Move the documentation for `RouteInfo` and `Transition` to their new modules. Additionally, eliminate the now-empty `lib` module for the `Transition` type since it *only* existed to provide documentation.
- Update coverage expectation to include the new items.

A note: this, extremely annoyingly, uses `@class` annotations for a great many types which are *not classes*. Our current API documentation infrastructure has no idea what types are (reasonably enough: it was built for a pure JS world!), so things like `interface` or especially `type FullName = ...` have… hacks… to make them show up correctly in the docs. However, they *do* work correctly in the docs; I ran against the API Docs app for most of the dev cycle.


dfreeman (commented):
Retroactive 👍

PR #20274, opened 2022-11-21
[BUGFIX release] Add some missing types to preview types

wagenet:
- Adds some new type tests for preview types
- Removes some non-public API from the preview types
- Adds some missing APIs to preview types


type-tests/preview/@ember/component-test/set-component-template.ts
```diff
@@ -0,0 +1,8 @@
+import { setComponentTemplate } from '@ember/component';
+import type { TemplateFactory } from 'htmlbars-inline-precompile';
+import { expectTypeOf } from 'expect-type';
+
+// Good enough for testing
+let factory = {} as TemplateFactory;
+
+expectTypeOf(setComponentTemplate(factory, {})).toEqualTypeOf<object>();
```
chriskrycho, line 8: Two alternative approaches that work well for spots like this:

1. Use `declare`:

    ```suggestion
    declare let factory: TemplateFactory
    
    expectTypeOf(setComponentTemplate(factory, {})).toEqualTypeOf<object>();
    ```

2. Test the parameters and return type explicitly instead:

    ```suggestion
    expectTypeOf(setComponentTemplate).parameters.toEqualTypeOf<[TemplateFactory, object]>();
    expectTypeOf(setComponentTemplate).returns.toEqualTypeOf<object>();
    ```

Both of these avoid needing the cast, which is not a *particularly* big deal here but is nice in the "modeling what we want to see" sense.

chriskrycho (approved):

PR #20275, opened 2022-11-22
Introduce infra to publish stable types

chriskrycho:
Provide a script which runs the compiler against a new tsconfig for generating types, wraps all the generated modules in `declare module` statements, and then creates an `index.d.ts` which uses side-effect style imports to expose them all, just the same as the preview types but generated from source.

> [!NOTE]
> Critically, this infrastructure does not expose *any* stable types in and of itself. Instead, it introduces a list of all the types still in preview mode which acts as a filter, and currently *all* modules are in that filter. Stabilizing the types for a given module will mean removing modules from that list and removing the corresponding preview types definitions.

There are… some hacks involved in this, as was foretold in the PR introducing preview types (#20180). Because Ember does not ship the real packages people import from in a Node-resolveable way (and [import maps][im] are not yet stable and therefore not useful for TS in particular), the way this works is anything but preferable. Medium-term, we need to figure out how to generate module rollups which *don’t* require us to manually do this wrapping, but this will do the trick for now.

[im]: https://github.com/WICG/import-maps

In support of this:

- Restructure how we handle the root `tsconfig.json` file by introducing a base compiler-options file which it extends, which in turn we can further extend off of for future needs.
- Fix a type error that was preventing type generation.


types/preview/index.d.ts
```diff
@@ -1,11 +1,21 @@
 /**
-  @module
-  Makes Ember's types for the packages it publishes available by importing this
-  path from `ember-source`:
+  Provides stable type definitions for Ember.js. It is maintained by hand, and
```
wagenet, line 2: I'm a bit confused by the wording here.

types/publish.mjs
```diff
@@ -0,0 +1,443 @@
...
+    contents = fs.readFileSync(modulePath, { encoding: 'utf-8' });
+  } catch (e) {
+    console.error(`Error reading ${modulePath}: ${e}`);
+    return;
+  }
+
+  let moduleNameForDeclaration = moduleName.replace('/index.d.ts', '');
+
+  let string = new MagicString(contents);
+  string.indent('  ').prepend(`declare module '${moduleNameForDeclaration}' {\n`).append('}\n');
```
wagenet, line 433: Am I correct that you're just using MagicString for these indent methods? It doesn't look like you're writing a source map.

wagenet (commented):
This seems promising, but I haven't tested it yet.

types/publish.mjs
chriskrycho: Ah, right. We can (if we want) switch to a more manual approach and the `indent-string` package, which is a bit smaller and lighter weight. I ended up here after having started there because I was trying to keep around the existing source maps that TS was generating, but I cannot see a way to do that, unfortunately.

packages/@ember/object/-internals.ts
```diff
@@ -7,9 +7,23 @@ import { symbol } from '@ember/-internals/utils';
 import { DEBUG } from '@glimmer/env';
 import EmberObject from '.';
 
+// Here we have runtime shenanigans to add debug-only errors to the class in dev
+// builds. Those runtime shenanigans produce the need for type-level shenanigans
+// to match: if we just assign without an explicit type annotation on the `let`
+// binding below for `FrameworkObject`, TS gets stuck because this creates
+// `FrameworkObject` with a class expression (rather than the usual class
+// declaration form). That in turn means TS needs to be able to fully name the
+// type produced by the clsas expression, which includes the `OWNER` symbol from
```
dfreeman, line 16:
```suggestion
// type produced by the class expression, which includes the `OWNER` symbol from
```

types/publish.mjs
```diff
@@ -0,0 +1,445 @@
...
+ */
+function wrapInDeclareModule(moduleName) {
+  let modulePath = path.join(TYPES_DIR, moduleName);
+
+  /** @type {string} */
+  let contents;
+  try {
+    contents = fs.readFileSync(modulePath, { encoding: 'utf-8' });
+  } catch (e) {
+    console.error(`Error reading ${modulePath}: ${e}`);
```
dfreeman, line 428: Do we care about exiting with a nonzero code if something like this happens?

chriskrycho: Ah, good point. Yes, we should; I'll push up a change to that effect.

dfreeman (approved):

PR #20276, opened 2022-11-22
[DOC] Add missing closing bracket in `RouterService` docs

bertdeblock:
Also added a few minor improvements along the way if that's okay.


packages/@ember/routing/router-service.ts
```diff
@@ -454,12 +454,12 @@ class RouterService<R extends Route> extends Service.extend(Evented) {
     import Route from '@ember/routing';
     import { service } from '@ember/service';
 
-    export default class extends Route {
+    export default class ContactFormRoute extends Route {
```
bertdeblock, line 457: So it matches the output of `ember g route contact-form`.

chriskrycho (approved):
💙

Issue #20185, opened 2022-09-06
add type tests infrastructure for supported TS versions


chriskrycho:
Part of implementing [RFC 0800: TypeScript Adoption Plan](https://rfcs.emberjs.com/id/0800-ts-adoption-plan) is: supported versions of TypeScript. This issue exists to tracking adding it to CI. We also need to do ember-learn/ember-website#955 once the CI infra is updated!

PR #20288, opened 2022-11-29
[FEATURE] Stable types for `@ember/owner`

chriskrycho:
This PR introduces the first stable types published from Ember's source code! 🎉 🚀 

Here's how it works:

1. Using the infrastructure introduced in #20185 (with some bug fixes), we include references to the types from `@ember/owner` and its dependency `@ember/-internals/owner` in the `types/stable/index.d.ts` file we generate as part of the build (3ca9834).

2. Accordingly, we now treat generating these types as part of the build operation: `yarn build` both runs a normal Ember build to generate the JS (`build:js`) and runs `types/publish.mjs` to emit these type definitions.

3. Since those types now exist, we also *remove* the corresponding types from `types/preview`, so that end users who are using our recommended imports—

    ```ts
    import 'ember-source/types';
    import 'ember-source/types/preview';
    ```

    —will not get type conflicts from *our own* types (also 3ca9834).

4. We collapse the two sets of type tests together (74c2895). This was just a miss before: I didn't realize that we were going to *necessarily* want to be testing them as a coherent whole, but of course, that's how our consumers will be experiencing them, so that's how we need to test them. The combination of (1) and (3) made this *really* obvious. 

As part of this PR, I introduced an absolute nightmare/monstrosity in abusing regex find and replace for the `types/publish.mjs` script to get rid of `declare` usages in the emitted module (see 86f6e52). In a future PR—indeed, the *next* PR, before landing any more such changes—I will switch that over to using `recast` to remove those in a more reasonable way, and then running `prettier` against the results so that the emitted type definition files look reasonable.


dfreeman (commented):
High level this all makes sense to me! Were all the `types/preview/**/tsconfig.json` files unneeded to begin with or did moving the tests around somehow impact that?

chriskrycho:
> Were all the `types/preview/**/tsconfig.json` files unneeded to begin with or did moving the tests around somehow impact that?

I believe they were unneeded in the first place, they were just leftover cruft from the DefinitelyTyped migration: DefinitelyTyped *does* need them given how its structure works, but we don't because we are doing the `declare module` shenanigans.

PR #20296, opened 2022-12-01
[BUGFIX LTS] Controller queryParams are readonly

wagenet:
* [x] Update DefinitelyTyped

### DIFF
--- a/packages/@ember/controller/index.ts
+++ b/packages/@ember/controller/index.ts
@@ -86,7 +86,7 @@ interface ControllerMixin<T> extends ActionHandler {
     @property queryParams
     @public
   */
-  queryParams: Array<ControllerQueryParam>;
+  queryParams: Readonly<Array<ControllerQueryParam>>;
 
   /**
     Transition the application into another route. The route may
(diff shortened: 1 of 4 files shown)


chriskrycho (approved):

chriskrycho:
Blergh, I realized after approving this when looking at the DT PR that this doesn't actually help us: this change doesn't actually make `queryParams` be a `ReadonlyArray`, since subclasses can make it a regular array—and in fact that's what we're doing explicitly in the tests here. 😩 So we should revert this.

[playground](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEkoGdnwKIA8oFsAOECA3gLABQ88AjnsgFzwBKIUwA9gHYQCeAgjDCjcAPMgAuMAJYcA5gD5yAX3LlQSOPEJj4IbPkIMsuAiADc5XccIA6Gsmt4ArsgAWACgBELkBAhsPAJQqZEio8ABibGw6mGIgHMBoRvrE5JR28AC88ADaXj5+HgC65mTKZKrg0Bpa8ABmUQyRbKUNbLa0Ds7u+b7+AaZAA)

wagenet:
All I wanted was for it to allow Readonly object to be passed in for queryParams, e.g. `queryParams = Object.freeze({ ... })`

PR #20298, opened 2022-12-06
Stop using the @ember/error package internally and improve types

wagenet:
This package just re-exports the native Error class. We should remove it entirely, but in the meantime, stop using it internally.


chriskrycho (approved):
👍🏼 ship it.

wagenet:
See also https://github.com/DefinitelyTyped/DefinitelyTyped/pull/63511

PR #20299, opened 2022-12-06
Build types when checking

wagenet:
This makes it easy to just run `yarn type-check` while developing.

Should we also remove the separate build step from the Linting CI?


chriskrycho (approved):
Updating the CI config seems reasonable to me 👍🏼

Issue #20303, opened 2022-12-06
Notes on Disentangling Ember's Internal Packages


chriskrycho:
In order to publish types without *hackzors*, to have properly tree-shake-able internals, and generally to be able to move faster on delivering features in Ember, we need to eliminate circular dependencies between the packages which make up Ember. This will allow us to introduce local `tsconfig.json` and `package.json` files which in turn we can use to actually *enforce* that API contract in the future.

This GitHub Issue represents public working notes on things I identify while working on TypeScript, and which others should feel empowered to pick up and work on. This is thankless, mostly invisible, wholly internal work, but if the framework is going to be able to actually move forward on *lots* of our goals, this *needs* to get done.

---

The major challenge is `@ember/-internals`, which has a *bunch* of stuff in it, much of which circularly depends on things in other packages.

Examples:

- [ ] `LinkTo` is in `@ember/-internals/glimmer/lib/components`, but depends on `@ember/engine` and `@ember/routing`. Notionally, `LinkTo` should be a part of `@ember/routing`, *not* a thing built into our Glimmer integration (!).

- [ ] All of the built-in components (`LinkTo`, `Textarea`, `Input`[^others]) are registered into the registry for an engine in `@ember/-internals/glimmer/lib/setup-registry`. That should presumably run the other way.

    The open question there is: *how*? This is 

- [ ] A lot of `@ember/-internals` code uses `@ember/debug`, which itself depends on `@ember/-internals` for:
    - `browser-environment`
    - `utils/types`: can (and should!) be extracted to their own package
    - private `owner` types: can just use the public API for `@ember/owner`
    - `environment` types

    The `browser-environment` and `environment` dependencies don't have a clear direction.

- ***TODO: more will go here over time!***

---

If you pick up one of these, leave a comment below indicating it and if you open a PR, reference this issue (but not with a "closing" keyword).

[^others]: Similarly to `LinkTo`, `Input` and `Textarea` should presumably have their own home (`@ember/component`, maybe? Or even upstreamed into `@glimmer/component`? That bit needs design/RFC work): wherever it is they’re imported from, or an associated `internals` directory they get re-exported from. These two aren't actually *blockers* for anything else, though, AFAICT.

kategengler:
Going to close this as we've extensively changed the package publishing since then and this issue isn't useful for tracking any further work.

PR #20309, opened 2022-12-06
Move is_proxy to @ember/-internals/runtime

wagenet:
Chipping away at #20303.


chriskrycho (dismissed):
Seems good. Noting for ease of history trawling: this is related to #20303.

chriskrycho:
Test failure, @wagenet!

packages/@ember/-internals/metal/lib/property_get.ts
```diff
@@ -172,11 +171,3 @@ _getProp({ unknownProperty() {} }, 1 as any);
 
 get({}, 'foo');
 get({}, 'foo.bar');
-
-let fakeProxy = {} as ProxyMixin<unknown>;
-setProxy(fakeProxy);
```
wagenet, line 177: I'm not sure we want to merge this, but it's necessary to avoid circularity that causes build issues.

chriskrycho: @pzuraq can you add some color on why this was added, and how important it is for perf for these?

chriskrycho: @wagenet what's the circularity issue it causes?

packages/@ember/-internals/runtime/index.ts
```diff
@@ -1,10 +1,11 @@
 export { default as RegistryProxyMixin } from './lib/mixins/registry_proxy';
 export { default as ContainerProxyMixin } from './lib/mixins/container_proxy';
 export { default as Comparable } from './lib/mixins/comparable';
-export { NativeArray } from '@ember/array';
```
chriskrycho, line 4: Unrelated change?

packages/@ember/-internals/metal/lib/property_get.ts
pzuraq: when using `get` with a proxy it goes through a different code path, so this was attempting to warm that up. Unsure if it was really that impactful, and it will be less impactful if proxy usage has dropped (which was the goal at the time)

chriskrycho (approved):
Other than the open questions, the changes seem reasonable to me. I asked pzuraq if he could comment on the perf issue; we might also be able to find some other spot to do that warmup if it's important!

packages/@ember/-internals/runtime/index.ts
wagenet: Split out to https://github.com/emberjs/ember.js/pull/20343.

wagenet:
I don't understand the type check failure here.

kategengler:
Is this still relevant?

wagenet:
Yes, I think this is still worth doing, but it's also not urgent and I haven't had time to do more here.

kategengler:
Moving to draft as it is not ready

PR #20310, opened 2022-12-06
Replace most `InternalOwner` uses with `Owner`

chriskrycho:
We only *need* the public API in most places, since most things which depend on the internal/private API are actually using `EngineInstance` or `ApplicationInstance`, and there aren't many of them! There are a couple remaining spots we have `InternalOwner`:

- When defining the type of `EngineInstance` itself, we use the type to guarantee we implement what we *say* we implement.

- We use `hasRegistration` or `resolveRegistration` in a couple places; both can be replaced with `Owner#factoryFor`, but I am keeping those separate so that this can be a clean types-only change.

Part of the work for #20303.

### DIFF
--- a/packages/@ember/routing/router.ts
+++ b/packages/@ember/routing/router.ts
@@ -1717,6 +1717,7 @@ function findRouteStateName(route: Route, state: string) {
   return routeHasBeenDefined(owner, router, stateName, stateNameFull) ? stateNameFull : '';
 }
 
+// TODO: rewrite `InternalOwner` to `Owner` by switching to `factoryFor`.
 /**
   Determines whether or not a route has been defined by checking that the route
   is in the Router's map and the owner has a registration for that route.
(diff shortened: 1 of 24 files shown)


packages/@ember/routing/router.ts
```diff
@@ -1717,6 +1717,7 @@ function findRouteStateName(route: Route, state: string) {
   return routeHasBeenDefined(owner, router, stateName, stateNameFull) ? stateNameFull : '';
 }
 
+// TODO: rewrite `InternalOwner` to `Owner` by switching to `factoryFor`.
```
wagenet, line 1720: When do you expect this to be addressed?

wagenet (approved):

packages/@ember/routing/router.ts
chriskrycho: In about ten minutes. :D Just keeping the types-only PR "clean".

PR #20311, opened 2022-12-06
Move @ember/-internals/utils/types to @ember/-internals/utility-types

wagenet:
Part of #20303.


chriskrycho (approved):
Seems good. I'll probably follow up by extracting it yet further to a dedicated type utils package, but this is good.

PR #20314, opened 2022-12-06
Add @ember/-debug-basic for internal use

wagenet:
`captureRenderTree` requires @ember/-internals/glimmer which has a lot of dependencies. The rest of the root @ember/debug import is pretty isolated, so making @ember/debug/-basic for internal use reduces a lot of entanglement.

For #20303.


packages/@ember/-internals/glimmer/tests/integration/application/actions-test.js
```diff
@@ -1,7 +1,7 @@
 import { moduleFor, ApplicationTestCase, RenderingTestCase, runTask } from 'internal-test-helpers';
 
 import Controller from '@ember/controller';
-import { getDebugFunction, setDebugFunction } from '@ember/debug';
+import { getDebugFunction, setDebugFunction } from '@ember/debug/basic';
```
chriskrycho, line 4: Got an actual typo error which the linter caught here.

wagenet:
I'm not sure what's going on with the tests here.

packages/@ember/debug/index.ts
```diff
@@ -1,341 +1,12 @@
...
-
-          debug(`For more advanced debugging, install the Ember Inspector from ${downloadURL}`);
-        }
-      },
-      false
-    );
-  }
-}
-
+// Unfortunately export * doesn't appear to work.
```
chriskrycho, line 3: For posterity, can you clarify at least in the PR here in what manner `export * from` doesn't work? Otherwise one of us is going to be wondering in a year. 😂

packages/ember/index.ts
```diff
@@ -12,7 +12,8 @@ import { meta } from '@ember/-internals/meta';
 import * as metal from '@ember/-internals/metal';
 import { FEATURES, isEnabled } from '@ember/canary-features';
 import * as EmberDebug from '@ember/debug';
-import { assert, captureRenderTree, deprecate } from '@ember/debug';
+import { assert, deprecate } from '@ember/debug';
+import { captureRenderTree } from '@ember/debug';
```
chriskrycho, line 16: Seems like these could be joined back together?

chriskrycho (changes requested):
I think separating this out is good, but the package naming is wrong. (I wasn't following fully when we discussed it in Discord, sorry.) The problem is that it moves us *away* from npm-ready package names, because `@ember/debug/-basic` is *not* a legal package name. Instead, we probably want something like `@ember/-debug-basic` or `@ember/-basic-debug`, which per your comment to me is probably a good foundation for `@ember/debug` itself at some point. In the very near term, that's the part we care about most for being able to actually roll up our types without doing the shenanigans we are in #20316.

(For lookers-on, what Peter and I had discussed is: We can deprecate any other bits, move out the parts which are built on top of both `debug` and `-internals` into their own package, and then remove the `-debug-basic` once it's no longer needed.)

packages/@ember/debug/index.ts
wagenet: Looks like it does work after all. Not sure what was going on.

chriskrycho:
The prod build tests [are legitimately failing](https://github.com/emberjs/ember.js/actions/runs/3959758878/jobs/6783117811#step:6:14) because of changes from this PR, @wagenet – it looks like it's not even hitting all the expected assertions in most of the tests it's failing on.

kategengler:
Is this still relevant?

wagenet:
I think so, but I haven't had the time to do anything and it's not urgent.

kategengler:
Converting to a draft as it is not ready.

PR #20316, opened 2022-12-08
Reimplement publish-types script with recast

chriskrycho:
This is far more robust against all the kinds of things our modules do:

- It supports having `declare module` and similar in the input source.
- It does not do dumb string replacement to remove `declare`, which can easily remove things from comments or the like.
- It correctly handles relative imports (`./` and chains of `../`).

For this to work correctly, we needed to have more recent versions of the Babel parser, so this freshens the lock file for all `@babel/*` dependencies. It also adds `recast` and drops `magic-string`.

This has been tested by enabling more than 50% of the unpublished modules; the resulting output is correct from the point of view of this script but exposes a number of fixes we will need to make for our types as we enable more and more of those.


packages/@ember/-internals/owner/index.ts
```diff
@@ -82,8 +82,8 @@ export interface DIRegistry extends Record<string, Record<string, unknown>> {}
 type ResolveFactoryManager<
   Type extends string,
   Name extends string
-> = DIRegistry[Type][Name] extends object
-  ? FactoryManager<DIRegistry[Type][Name]>
+> = DIRegistry[Type][Name] extends infer RegistryEntry extends object
+  ? FactoryManager<RegistryEntry>
```
chriskrycho, line 86: Unclear *why*, but `tsc` started grumping about assignability without this.

types/publish.mjs
```diff
@@ -471,5 +493,175 @@ function wrapInDeclareModule(moduleName) {
...
 }
 
+/**
+  Rewrite a given module declaration:
+
+  - Tranform the main body of the module into a new top-level `declare module`
+    statement.
+      - Remove all `declare` modifiers from items in the module itself.
+      - Update all `import` specifiers to be absolute in terms of the package
+        name (which means )
```
wagenet, line 503: `which means`... what?

wagenet (approved):

types/publish.mjs
chriskrycho: Dang it. 😂 I'll fix it.

chriskrycho:
```suggestion
        name, which means handling both `./` and `../` correctly.
```

PR #20327, opened 2022-12-21
[BUGFIX LTS] Fix the types for the mutation-methods of `NativeArray`

boris-petrov:
It's really annoying that I can't do:

```ts
const array = [1, 2, 3] as readonly number[];
const other = A().setObjects(array); // same with `pushObjects`, `unshiftObjects`, `removeObjects` and `addObjects`
```

I believe this PR fixes that.

cc @chriskrycho

### DIFF
--- a/types/preview/@ember/array/-private/native-array.d.ts
+++ b/types/preview/@ember/array/-private/native-array.d.ts
@@ -7,7 +7,7 @@ declare module '@ember/array/-private/native-array' {
   // Get an alias to the global Array type to use in inner scope below.
   type Array<T> = T[];
 
-  type AnyArray<T> = EmberArray<T> | Array<T>;
+  type AnyArray<T> = EmberArray<T> | Array<T> | ReadonlyArray<T>;
 
   /**
    * The final definition of NativeArray removes all native methods. This is the list of removed methods


types/preview/@ember/array/-private/native-array.d.ts
```diff
@@ -4,10 +4,11 @@ declare module '@ember/array/-private/native-array' {
   import MutableArray from '@ember/array/mutable';
   import Mixin from '@ember/object/mixin';
 
-  // Get an alias to the global Array type to use in inner scope below.
+  // Get aliases to the global Array types to use in inner scope below.
   type Array<T> = T[];
+  type ReadonlyArray<T> = readonly T[];
```
chriskrycho, line 9: This actually just duplicates a built-in type of the same name, and we don't *need* to otherwise reference it past this.

```suggestion
```

types/preview/@ember/array/-private/native-array.d.ts
```diff
@@ -4,10 +4,11 @@ declare module '@ember/array/-private/native-array' {
...
 
-  type AnyArray<T> = EmberArray<T> | Array<T>;
+  type AnyArray<T> = EmberArray<T> | ReadonlyArray<T>;
```
chriskrycho, line 11: So this will "just work", but we also want to keep around the ability to work explicitly with regular arrays:

```suggestion
type AnyArray<T> = EmberArray<T> | Array<T> | ReadonlyArray<T>;
```

types/preview/@ember/array/-private/native-array.d.ts
```diff
@@ -4,10 +4,11 @@ declare module '@ember/array/-private/native-array' {
   import MutableArray from '@ember/array/mutable';
   import Mixin from '@ember/object/mixin';
 
-  // Get an alias to the global Array type to use in inner scope below.
+  // Get aliases to the global Array types to use in inner scope below.
```
chriskrycho, line 7: And we can revert this.

```suggestion
  // Get an alias to the global Array type to use in inner scope below.
```

chriskrycho (changes requested):
Thanks!

types/preview/@ember/array/-private/native-array.d.ts
boris-petrov: That's true... but the same is true for the `Array` type above. It's also built-in. I wasn't sure why it was added but decided to be consistent. I'm fine both ways. I'll remove the `ReadonlyArray` if you prefer?

types/preview/@ember/array/-private/native-array.d.ts
boris-petrov: Doesn't `ReadonlyArray<T>` include all `Array<T>`?

boris-petrov: That is, an `Array<T>` can always be cast to/assigned to a `ReadonlyArray<T>` I believe?

types/preview/@ember/array/-private/native-array.d.ts
chriskrycho: The `Array` type has historically been *shadowed* locally in a way that `ReadonlyArray` hasn't, and so we needed a way to refer to it distinctly. (Totally fair not to have that context.) Here, I suspect it could go away as well, but we should tackle that separately!

types/preview/@ember/array/-private/native-array.d.ts
chriskrycho: *Cast to*, yes, but *assigned to*, no: it depends on what position it's in in terms of variance whether it's legal to substitute one for the other. Explicitly specifying that we accept both any place we have `AnyArray<T>` works around that.

boris-petrov: OK, I did what you suggest. But if you have the time, can you show me an example (in the TypeScript playground for example) of when an `Array<T>` can't be passed as a `ReadonlyArray<T>`?

chriskrycho (approved):

types/preview/@ember/array/-private/native-array.d.ts
chriskrycho: Yep, any callback function parameter will get you. [example](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXxAA8oBbABwhAElUAeAFQD4AKMAIwC55nYYoBPLgCUQUYHgj8AgjD78GjAJTwAvI3gy5CxVwBuOLMADcAKFCRYCFOmx4CxcpQDyyDAtaduvAV00DtquoiYhLSsv5MOvD6hqYmJtaYuPjiyGyUAKIkzKgAzlyoyCRsIDAA2gC6UYXFpZXwAN4m8C3wcBjIMPh5AHQkUGQ5gfD4AFTwAEyKJgC+8USkFNSozKnpIFmKpguOIC4YqzhpmSRb8UA)

PR #20328, opened 2022-12-21
Normalize to always import from `'backburner.js'`

chriskrycho:
We had a mix of different import locations for this, both the ES imports via `backburner.js` and the Ember CLI-munged version available at `backburner`. Switch to only using the Node-resolveable version, so TS can resolve it correctly; and enforce it by removing `backburner` from the list of `core` modules in the ESLint config.


chriskrycho:
This is blocked on figuring out what part of the build pipeline makes `backburner.js` *only* importable as `backburner`.

chriskrycho:
Fixed along the way to #20449, and I just forgot to close it!

PR #20333, opened 2022-12-28
[DOC] Improve didTransition in routing

olenderhub:
Issue:

1. This part of code: `this.controller.get('errors.base').clear();` breaks https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-controller-access-in-routes.md rule.
2. Using `didTransition` this way is deprecated in favor of https://deprecations.emberjs.com/v3.x/#toc_deprecate-router-events

### DIFF
--- a/packages/@ember/routing/route.ts
+++ b/packages/@ember/routing/route.ts
@@ -145,7 +145,7 @@ interface Route<T = unknown> extends IRoute<T> {
     export default class LoginRoute extends Route {
       @action
       didTransition() {
-        this.controller.get('errors.base').clear();
+        // your code there
         return true; // Bubble the didTransition event
       }
     }


packages/@ember/routing/route.ts
```diff
@@ -145,12 +145,14 @@ interface Route<T = unknown> extends IRoute<T> {
...
       didTransition() {
-        this.controller.get('errors.base').clear();
+        // your code there
         return true; // Bubble the didTransition event
       }
     }
     ```
 
+    Example above is deprecated in favor of https://deprecations.emberjs.com/v3.x/#toc_deprecate-router-events
+
```
chriskrycho, line 155: Actually, the `Route.didTransition` and `Route.willTransition` actions were *not* removed: these are per-route lifecycle hooks, while the deprecation was for `Router.didTransition` and `Router.willTransition` (note: `Router`, not `RouterService`). This is easy to get confused by: I myself had to go dig back through all the details! And honestly we probably *should* remove these in favor of "please just use the router service", but I expect we will do so via general router cleanup. Accordingly, let's refer this change:

```suggestion
```

chriskrycho (changes requested):
Thanks for this, and good fix overall! I have one correctly we need to make, but I'll happily land the other!

olenderhub:
@chriskrycho thank you for the clarification. I misunderstood this. I removed what you mentioned in the comment and pushed by force code related to removing the old approach.

chriskrycho (approved):
No problem! Like I said, I had to go back and check it myself, so I sympathize. 😂

PR #20344, opened 2023-01-04
Deprecate @ember/string when used from ember-source; point users to add the `@ember/string` addon

kategengler:
- Duplicate a portion of @ember/string in @ember/-internals/string for internal use by ember-source without triggering deprecations

### DIFF
--- a/packages/ember/index.ts
+++ b/packages/ember/index.ts
@@ -633,39 +633,40 @@ Object.defineProperty(Ember, 'TEMPLATES', {
   enumerable: false,
 });
 
-const deprecateImportFromString = function (
-  name: string,
-  message = `Importing ${name} from '@ember/string' is deprecated. Please import ${name} from '@ember/template' instead.`
-) {
-  deprecate(message, false, {
-    id: 'ember-string.htmlsafe-ishtmlsafe',
-    for: 'ember-source',
-    since: {
-      available: '4.10',
-      enabled: '4.10',
-    },
-    until: '5.0.0',
-    url: 'https://deprecations.emberjs.com/v3.x/#toc_ember-string-htmlsafe-ishtmlsafe',
-  });
-};
-// NOTE: these are expressly *not* in the public API, because they were
-// deprecated and removed. TODO: remove them after we land the TS conversion,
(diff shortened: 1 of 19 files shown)


wagenet (approved):
Looks good. For completeness we possibly should add @ember/-internals/string to the list in publish.mjs, though we probably don't actually need it since its is never re-exported.

packages/ember/index.ts
```diff
@@ -633,39 +633,40 @@ Object.defineProperty(Ember, 'TEMPLATES', {
...
-Object.defineProperty(Ember.String, 'htmlSafe', {
-  enumerable: true,
-  configurable: true,
-  get() {
-    deprecateImportFromString('htmlSafe');
-    return htmlSafe;
-  },
-});
-Object.defineProperty(Ember.String, 'isHTMLSafe', {
+function deprecateStringUseOnEmberModule() {
```
kategengler, line 636: The prior deprecation here would show `Importing htmlSafe from '@ember/string' is deprecated. Please import htmlSafe from '@ember/template`, but would only show when someone was doing 

```
import Ember from 'ember';
Ember.String.htmlSafe('foo');
```

Now, they will see 'Using `Ember.String` is deprecated. Please import methods directly from `@ember/string`' If they do that with `htmlSafe` or `isHTMLSafe` they will then get a deprecation about importing from `@ember/template`. It's a little annoying but I think it is the cleanest way forward and most projects will have updated the imports already.

packages/ember/index.ts
```diff
@@ -633,39 +633,40 @@ Object.defineProperty(Ember, 'TEMPLATES', {
...
+  );
+}
+
+Object.defineProperty(Ember, 'String', {
   enumerable: true,
   configurable: true,
   get() {
-    deprecateImportFromString('isHTMLSafe');
-    return isHTMLSafe;
+    deprecateStringUseOnEmberModule();
```
kategengler, line 657: This deprecation must be here, and not in the definitions of the methods on `@ember/string` because it only applies when methods are referenced off the Ember module.

wagenet (approved):

PR #20352, opened 2023-01-13
Enable `EMBER_TYPESCRIPT_BLUEPRINTS` flag

### LINKED ISSUE #20177: Enable `EMBER_TYPESCRIPT_BLUEPRINTS` feature
Currently, this feature is [turned off](https://github.com/emberjs/ember.js/blob/master/blueprints/-maybe-polyfill-typescript-blueprints.js#L2) and users need to explicitly run `EMBER_TYPESCRIPT_BLUEPRINTS=true ember g ...` to generate TS files. This can be confusing when people also set `"isTypeScriptProject": true` in their `.ember-cli` file.
This issue serves as a reminder to enable this feature or to determine what's needed to enable it.

/cc @chriskrycho

bertdeblock:
Closes https://github.com/emberjs/ember.js/issues/20177.


chriskrycho (commented):
Seems good here – let me know when you're ready to move it out of Draft!

bertdeblock:
@chriskrycho Done! Was mostly waiting on a "good to go", wasn't sure if we could already land this.

chriskrycho:
Yeah, let me do a quick pass on the actual blueprints to make sure they're good, but/and we should have folks kick the tires a bunch during the 4.12 beta cycle—that’s what the beta cycle is for!

chriskrycho:
We need to add the service registry to the service blueprints, but I believe we're otherwise good to go; I'm putting up a PR to do that now.

chriskrycho:
Chatting with @cafreeman, he reminded me that we also need the blueprints to support generating signatures for components!

PR #20356, opened 2023-01-19
[FEATURE] generate `Signature` in component blueprints

chriskrycho:
Since [RFC 0748][0748], Glimmer components support a `Signature` type which specifies any arguments they accept, any blocks they yield, and the element types to which `...attributes` are applied. Incorporate the signature when generating TS blueprints for Glimmer components, including template-only components. Do *not* generate them for Ember (classic) components, since there are severe limitations in using classic components that way, and since we only support the Octane programming model in our TypeScript support per [RFC 0800][0800].

Unblocks #20352 and thus #20177!

[0748]: https://rfcs.emberjs.com/id/https://rfcs.emberjs.com/id/0748-glimmer-component-signature/
[0800]: https://rfcs.emberjs.com/id/https://rfcs.emberjs.com/id/0800-ts-adoption-plan/

### DIFF
--- a/blueprints/-maybe-polyfill-typescript-blueprints.js
+++ b/blueprints/-maybe-polyfill-typescript-blueprints.js
@@ -8,10 +8,12 @@ function canEmitTypeScript() {
 }
 
 module.exports = function (context) {
-  if (canEmitTypeScript()) {
+  let canUseTypeScript = canEmitTypeScript();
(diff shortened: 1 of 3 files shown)


wagenet (approved):

chriskrycho:
Odd, I wonder why that didn't fail locally? I'll dig in and get that sorted.

PR #20358, opened 2023-01-25
[DRAFT] Redesign API for trusted (unsafe) strings

chriskrycho:
This is a draft PR to accompany a forthcoming proposal to improve this API along lines folks have talked about for years; I drafted the change to make it easier to discuss concretely with the accompanying RFC. This should not be taken as indicating we *are* doing this, only that I am *proposing* we do it and making it easy if we decide to do so!

---

- Introduce an opaque public type for the result of marking a string as trusted, `TrustedString`. Make all the mechanical details of the class private, accessible *only* within the internal module. Define the type as an opaque type that other users can neither instantiate nor subclass, which implements the `SafeString` contract from Glimmer, i.e. it provides only a `toHTML(): string` method.

- Deprecate `htmlSafe` and `isHTMLSafe` and introduce new names which indicate what they actually do/are: `unsafelyTrustString` and `isTrustedString`.

### DIFF
--- a/packages/@ember/-internals/glimmer/index.ts
+++ b/packages/@ember/-internals/glimmer/index.ts
@@ -451,7 +451,13 @@ export { default as LinkTo } from './lib/components/link-to';
 export { default as Textarea } from './lib/components/textarea';
 export { default as Component } from './lib/component';
 export { default as Helper, helper } from './lib/helper';
-export { SafeString, htmlSafe, isHTMLSafe } from './lib/utils/string';
+export {
+  type TrustedString,
+  unsafelyTrustString,
+  isTrustedString,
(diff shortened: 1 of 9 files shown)


wagenet:
See also https://github.com/emberjs/rfcs/pull/443

kategengler:

chriskrycho:
Still worth doing, and this approach is probably roughly right, but obviously I am not going to land it. I will close this, but/and feel free to point people here as useful prior art, and folks should feel *very* free to `git cherry-pick` this and run with it themselves if anyone wants to!

chriskrycho:
…also it needs an RFC. Which is why we did not land it.

PR #20361, opened 2023-01-26
[BUGFIX BETA] Support a `router.ts` file when generating routes

### LINKED ISSUE #20357: [Bug] Error generating Typescript route blueprints
### 🐞 Describe the Bug
When I generate a new typescript ember application, I cannot generate a route using typescript files and does not update the `router.ts` file either

### 🔬 Minimal Reproduction
- Generate a new project `ember new my-profile --lang en --yarn true --typescript true`
- cd into the directory
- ember generate route location 

### 😕 Actual Behavior
image

### 🤔 Expected Behavior
A location.ts file is created with a template file and the router.ts file is updated

### 🌍 Environment

- Ember: - 4.10.0
- Node.js/npm: - 16.18.1
- OS: - Mac OS
- Browser: - Google Chrome

### ➕ Additional Context
I also tried `EMBER_TYPESCRIPT_BLUEPRINTS=true ember generate route location` and this generates the file with `ts` extension but I get an error about not being able to find `router.js` 

Also when generating project I saw this warning

image

bertdeblock:
Closes #20357.

### DIFF
--- a/blueprints/route/index.js
+++ b/blueprints/route/index.js
@@ -5,6 +5,7 @@ const path = require('path');
 const chalk = require('chalk');
 const stringUtil = require('ember-cli-string-utils');
 const EmberRouterGenerator = require('ember-router-generator');
+const SilentError = require('silent-error');
 
 const maybePolyfillTypeScriptBlueprints = require('../-maybe-polyfill-typescript-blueprints');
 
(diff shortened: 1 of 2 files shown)


blueprints/route/index.js
```diff
@@ -148,21 +149,30 @@ function updateRouter(action, options) {
...
   if (options.dummy && options.project.isEmberCLIAddon()) {
-    routerPathParts = routerPathParts.concat(['tests', 'dummy', root, 'router.js']);
+    routerPathParts.push('tests', 'dummy', 'app');
   } else {
-    routerPathParts = routerPathParts.concat([root, 'router.js']);
+    routerPathParts.push('app');
   }
 
-  return routerPathParts;
+  for (const routerFile of ['router.js', 'router.ts']) {
```
bertdeblock, line 161: I guess we can also check for the `isTypeScriptProject` option, but that doesn't necessarily mean there's a `router.ts` file as well.

blueprints/route/index.js
```diff
@@ -148,21 +149,30 @@ function updateRouter(action, options) {
...
+    let routerPath = path.join(...routerPathParts, routerFile);
+
+    if (fs.existsSync(routerPath)) {
+      return routerPath;
+    }
+  }
+
+  throw new SilentError(
+    'Could not find router file. Please make sure your project has a `router.js` or `router.ts` file.'
```
bertdeblock, line 170: Seemed like a nicer DX compared to the default Node error.

blueprints/route/index.js
chriskrycho: I think you're right, and I think we should definitely do the right thing if we find `router.ts` even in a non-`isTypeScritpProject` setup. One note, though: what should happen if there are *both* of these present? Presumably we should error on that condition, because it is not clear which one we should update. So maybe:

```js
let jsRouterPath = path.join(...routerPathParts, 'router.js');
let tsRouterPath = path.join(...routerPathParts, 'router.ts');

let jsRouterPathExists = fs.existsSync(jsRouterPath);
let tsRouterPathExists = fs.existsSync(tsRouterPath);

if (jsRouterPathExists && tsRouterPathExists) {
  throw new SilentError('Found both a `router.js` and a `router.ts` file. Please make sure your project only has one or the other.')
}

if (jsRouterPathExists) return jsRouterPath;
if (tsRouterPathExists) return tsRouterPath;

// existing error
```

node-tests/blueprints/route-test.js
```diff
@@ -301,6 +301,24 @@ describe('Blueprint: route', function () {
         });
       });
     });
+
```
chriskrycho, line 304: And then a test for that new error case?

chriskrycho (changes requested):
This is great. One small tweak and let's do it!

Process-wise: since we enabled the flag, this looks like a `BUGFIX BETA` I think?

chriskrycho (approved):

PR #20373, opened 2023-02-14
[BUGFIX BETA] Make the *type* for `SafeString` public

chriskrycho:
This type has been effectively "intimate" for many years, and fits in the same bucket as e.g. `Transition`: it is not user-constructible, but will be constructed by the framework and users need to be able to name it. For example, `ember-intl` needs to be able to see that a string has been marked as trusted to do the right thing to emit it.

Internally, clean up a few long-standing TS issues (`any` etc.), make `SafeString` explicitly implement the contract from Glimmer so that if that contract changes, we will know at the definition site, and make the implementation details of how `SafeString` handles the string it wraps `private`. (This does not use a `#`-private field because private class fields have some non-trivial overhead in transpiled contexts, and `SafeString` can appear in fairly hot rendering paths.)

---

I will back port this to the preview types as well once we merge this.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/utils/string.ts
+++ b/packages/@ember/-internals/glimmer/lib/utils/string.ts
@@ -2,15 +2,17 @@
 @module @ember/template
 */
 
-export class SafeString {
-  public string: string;
+import type { SafeString as GlimmerSafeString } from '@glimmer/runtime';
+
+export class SafeString implements GlimmerSafeString {
+  private __string: string;
 
   constructor(string: string) {
-    this.string = string;
+    this.__string = string;
   }
 
   toString(): string {
-    return `${this.string}`;
+    return `${this.__string}`;
   }
 
   toHTML(): string {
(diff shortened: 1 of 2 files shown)


wagenet (approved):

packages/@ember/-internals/glimmer/lib/utils/string.ts
```diff
@@ -2,15 +2,17 @@
 @module @ember/template
 */
 
-export class SafeString {
-  public string: string;
+import type { SafeString as GlimmerSafeString } from '@glimmer/runtime';
+
+export class SafeString implements GlimmerSafeString {
```
chriskrycho, line 7: I realized after landing this: we'll need docs here to make it show up appropriately. I'll follow up to that effect tomorrow.

jrjohnson:
I'm not sure where this falls on the 'is a bug / is a mistake in our code' spectrum, but we were accessing the results of `htmlSafe(str).string` in tests and now need to change these to `htmlSafe(str).toString()` after this change. It's fairly straightforward to fix, but I'm surprised the `string` property which was maybe sort of public got killed in a minor version update.

chriskrycho:
The class itself was not public, so relying on any part of it was subject to breakage at any time. The field was only `public` for *internal use*. When making the class public, I chose to hide that part of the implementation so that if we evolve this in the future we can do so without breaking the contract. In the future, don’t rely on private API, and you won’t get broken when it changes!

jrjohnson:
I hear that, and honestly this is no big deal, but I wasn't relying on anything like a type when we wrote this code 8 years ago 😁, what we were using was the result of `htmlSafe()` and I'm guessing it was a POJO or possible some object that we could discover has a `string` property on it in a debugger. It was an easy fix, I just wanted to add a comment here for the next sluther and hopefully make it a bit more SEO for someone else looking.

PR #20382, opened 2023-02-21
[CLEANUP] Remove deprecated injection

wagenet:


chriskrycho (changes requested):
We'll also need to remove `inject` from `RegistryProxyMixin`, [here](https://github.com/emberjs/ember.js/blob/0e4c831b4cf9e0acc5f0568cb866a6edf8505f69/packages/@ember/-internals/runtime/lib/mixins/registry_proxy.ts#L92).

chriskrycho (approved):

PR #20385, opened 2023-02-21
[BUGFIX LTS] Improve error for owner methods called after destroy

chriskrycho:
The previous error handling gives you a stack trace and a notice, but does not tell you what was being asked for with the lookup. This makes the traces much less helpful, especially for minified/prod builds.


rwjblue (approved):

runspired:
I took this a step further in a patch lately and it got me thinking it would be nice to be able to hook this particular error and the "called set" error (basically the `your app is torn down` errors) for decorating with additional context. Specifically I patched in the ability to grab the owner's guid and then map that to the owner guid from a qunit test context to ensure confidence in which test had been borked.

PR #20388, opened 2023-03-01
[BUGFIX LTS] Don't run getters while applying mixins

wycats:
This change ensures that getters are never evaluated while applying
mixins.

It relies on the fact that all getters (including undecorated ones) get
converted into classic decorators when the mixin is originally created.

### DIFF
--- a/packages/@ember/object/mixin.ts
+++ b/packages/@ember/object/mixin.ts
@@ -14,7 +14,6 @@ import {
   type ComputedPropertyObj,
   type ComputedPropertySetter,
   type ComputedDescriptor,
-  isAccessor,
   isClassicDecorator,
 } from '@ember/-internals/metal';
 import {
--- a/packages/@ember/object/mixin.ts
+++ b/packages/@ember/object/mixin.ts
@@ -8,12 +8,14 @@ import { guidFor, observerListenerMetaFor, ROOT, wrap } from '@ember/-internals/
 import { assert } from '@ember/debug';
 import { DEBUG } from '@glimmer/env';
 import { _WeakSet } from '@glimmer/util';
-import type {
-  ComputedDecorator,
-  ComputedPropertyGetter,
-  ComputedPropertyObj,
-  ComputedPropertySetter,
-  ComputedDescriptor,
+import {
+  type ComputedDecorator,
+  type ComputedPropertyGetter,
+  type ComputedPropertyObj,
+  type ComputedPropertySetter,
+  type ComputedDescriptor,
+  isAccessor,
+  isClassicDecorator,
 } from '@ember/-internals/metal';
 import {
   ComputedProperty,
@@ -235,7 +237,7 @@ function mergeMixins(
   keys: string[],
   keysWithSuper: string[]
 ): void {
-  let currentMixin;
+  let currentMixin: MixinLike | undefined;
 
   for (let i = 0; i < mixins.length; i++) {
     currentMixin = mixins[i];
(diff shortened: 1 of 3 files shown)


chriskrycho:
Looks like this is aimed at the issue ID'd in #19916?

gitKrystan:
@chriskrycho It's related to https://github.com/emberjs/ember.js/issues/18860 and https://github.com/emberjs/ember.js/pull/20129

chriskrycho:
Yep, and #20129 was a failing test for #19916. 👍🏼

wycats:
@chriskrycho Do you mean to refer to "No documented public API for `Owner`"?

chriskrycho:
Nope, brain fog. I meant #18860, which Krystan linked. 😮‍💨

tsconfig.json
```diff
@@ -4,6 +4,6 @@
     "outDir": "dist",
     "noEmit": true
   },
-  "include": ["packages/**/*.ts"],
+  "include": ["packages/**/*.ts", "packages/**/*.js"],
```
chriskrycho, line 7: Hmm, why this change? We're intentionally *not* including `.js` files here, and also don't have `checkJs: true`, so it shouldn't particularly affect the behavior of the compiler. 🤔

packages/@ember/object/mixin.ts
```diff
@@ -8,12 +8,14 @@ import { guidFor, observerListenerMetaFor, ROOT, wrap } from '@ember/-internals/
...
-  ComputedPropertyObj,
-  ComputedPropertySetter,
-  ComputedDescriptor,
+import {
+  type ComputedDecorator,
+  type ComputedPropertyGetter,
+  type ComputedPropertyObj,
+  type ComputedPropertySetter,
+  type ComputedDescriptor,
+  isAccessor,
```
chriskrycho, line 17: You imported this but didn't use it *and* it doesn't appear to be used below, so I think we can simply drop this line from the `import` statement?

```suggestion
```

chriskrycho (changes requested):
Couple small things and then we should be able to land and back port this. Thanks for getting it fixed!

tsconfig.json
gitKrystan: @wycats Mentioned to me that this changed fixed some issues with auto-import in VSCode. If we decide to keep it, it should probably go in a separate PR.

packages/@ember/object/mixin.ts
gitKrystan: Interestingly, the [files tab](https://github.com/emberjs/ember.js/pull/20388/files) of this PR is showing linter (tsc) errors related to this, but they didn't fail CI. Is that configuration intentional?

tsconfig.json
wycats: Yeah, totally agree. I meant to separate it out but I think we forgot.

wycats: TL;DR if you don't include the JS files in the tsconfig then the language server doesn't do any cross-module resolving in the JS files, which:

- breaks auto-import
- causes types from other modules to not resolve, which creates a lot of inferred `any`s for no reason (and breaking intellisense)
- causes some types to resolve to the published Ember types (which are present in node_modules), which is pretty confusing.

I meant to check whether jsconfig would be good enough to resolve these issues. Since you're the TS expert here, would you prefer to add the `js` files to `tsconfig` (without `checkJS`), or try to separate them into another config file.

This could either mean `jsconfig` (if it works) or creating a second `tsconfig` and link to it from the main one with a project reference.

I'm open to anything that addresses the workflow issues, and I'm totally fine with separating it into a separate PR.

gitKrystan:
@chriskrycho I think this is ready

tsconfig.json
chriskrycho: Ah, so this specifically ends up affecting tests, which are largely unconverted? Makes sense. I'm totally fine with `js` files in the config; I just wanted to understand where they had come up since all our actual implementation is now in TS—but "tests" makes sense. We can land that right after this. Thanks for explaining!

chriskrycho (approved):

chriskrycho:
I am planning to do an LTS release (so: back to 4.8) by the end of this week and will make sure this is included!

PR #20398, opened 2023-03-04
[BUGFIX LTS] Fix runloop types on TS 5.0+

Andarist:
related to https://github.com/microsoft/TypeScript/issues/53080

---

**Edit by @chriskrycho, 2023/03/04:** TS 5.0 catches a new assignability error here, and our (very) complicated set of types for `bind` and friends got flagged by that. The terminal case for `PartialParams` needs to be a fixed empty tuple rather than `never`, so that the type system can see that it will be resolved as 'no arguments' rather than 'nothing' in that scenario.

> **Note**: this fixes the one bug on our side. There is also a bug fix in the latest TS nightly (`typescript@next`) build, which combined with this PR should get us unblocked against TS nightlies again. We're in good shape for TS 5.0 and the upcoming 5.1 as a result!

> **Note**: I have marked this as `BUGFIX LTS` because it affects our *build*. End users will not be affected by these changes until we have types publishing from source for this (presumably in 5.1 or 5.2), but any fixes we *do* make to 4.8 or 4.12 will see this in their CI runs, so we need to back-port it.

### DIFF
--- a/packages/@ember/runloop/index.ts
+++ b/packages/@ember/runloop/index.ts
@@ -14,7 +14,7 @@ type PartialParams<P extends any[]> = P extends [infer First, ...infer Rest]
   : // This is necessary to handle optional tuple values
   Required<P> extends [infer First, ...infer Rest]
   ? [] | [First | undefined] | [First | undefined, ...PartialParams<Partial<Rest>>]
-  : never;
+  : [];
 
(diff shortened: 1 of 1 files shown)


chriskrycho (approved):
Oh, totally didn't see this before opening #20400, which I can now close as a duplicate – thank you!

PR #20400, opened 2023-03-04
[BUGFIX LTS] Fix runloop types on TS 5.0+

chriskrycho:
TS 5.0 catches a new set of assignability errors here, and our (very) complicated set of types for `bind` and friends got flagged by that. The terminal case for `PartialParams` needs to be a fixed empty tuple rather than `never`, so that the type system can see that it will be resolved as 'no arguments' rather than 'nothing' in that scenario.

> **Note**: this fixes the one bug on our side. There is also a bug fix in the latest TS nightly (`typescript@next`) build, which combined with this PR should get us unblocked against TS nightlies again. We're in good shape for TS 5.0 and the upcoming 5.1 as a result!

> **Note**: I have marked this as `BUGFIX LTS` because it affects our *build*. End users will not be affected by these changes until we have types publishing from source for this (presumably in 5.1 or 5.2), but any fixes we *do* make to 4.8 or 4.12 will see this in their CI runs, so we need to back-port it.


chriskrycho:
Duplicate of #20398. 🎉

PR #20402, opened 2023-03-06
[CLEANUP] Drop support for `ember-cli-mocha` and `ember-mocha` when generating test blueprints

bertdeblock:
`ember-cli-mocha` was deprecated a long time ago and `ember-mocha` simply doesn't work with Ember v4 and above.
Instead of implementing deprecations that no one will ever see, we decided to drop support for `ember-cli-mocha` and `ember-mocha` in Ember v5.


kategengler:
Thank you!

PR #20405, opened 2023-03-09
[BUGFIX LTS 4.4] Don't run getters while applying mixins

gitKrystan:
Backport of https://github.com/emberjs/ember.js/pull/20388

This change ensures that getters are never evaluated while applying mixins.

It relies on the fact that all getters (including undecorated ones) get converted into classic decorators when the mixin is originally created.

### DIFF
--- a/package.json
+++ b/package.json
@@ -155,7 +155,7 @@
     "typescript": "~4.6.0"
   },
   "engines": {
-    "node": ">= 12.*"
+    "node": ">= 14.*"
   },
   "ember-addon": {
     "after": "ember-cli-legacy-blueprints"
--- a/.github/workflows/alpha-releases.yml
+++ b/.github/workflows/alpha-releases.yml
@@ -2,7 +2,7 @@ name: Alpha Releases
 
 on:
   schedule:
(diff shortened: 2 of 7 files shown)


wycats (commented):
Looks great to me! I backported the node version change, because outdated node is what was making the smoke test fail.

package.json
```diff
@@ -155,7 +155,7 @@
     "typescript": "~4.6.0"
   },
   "engines": {
-    "node": ">= 12.*"
+    "node": ">= 14.*"
```
kategengler, line 158: I don't think we can bump this in a point release since it would be a breaking change that would be very unexpected. Bumping the version for a particular CI job should be fine though I am curious why it failed: theoretically the smoke test should work on 4.4 without changes.

wycats: Isn't the rule about releases that each release supports whatever node version is still under maintenance?

kategengler: We discussed on discord, but writing here as well: I think a node version bump in a patch release would be very surprising to users. The policy states only when we drop support on the primary branch of ember-cli https://github.com/ember-cli/ember-cli/blob/master/docs/node-support.md

PR #20407, opened 2023-03-14
[DOC release] 'empty function' typo in isEmpty util description

prashanthsasidharan:


NullVoxPopuli (approved):
LGTM! thank you!! :tada:

phndiaye (approved):

types/preview/@ember/utils/index.d.ts
```diff
@@ -12,8 +12,8 @@ declare module '@ember/utils' {
   export function isBlank(obj?: unknown): boolean;
 
   /**
-   * Verifies that a value is `null` or an empty string, empty array,
-   * or empty function.
+   * Verifies that a value is null or undefined, an empty string,
```
chriskrycho, line 15: We should go ahead and keep the backticks here—they affect how this gets rendered! It's the difference between these two:

> Verifies that a value is null or undefined, an empty string, or an empty array.



> Verifies that a value is `null` or `undefined`, an empty string, or an empty array.

chriskrycho:
```suggestion
   * Verifies that a value is `null` or `undefined`, an empty string,
```

chriskrycho (approved):

PR #20410, opened 2023-03-16
[BUGFIX types] Add missing type declarations in the preview types for @ember/template-compilation

NullVoxPopuli:
Adds types for `@ember/template-compilation`

### DIFF
--- a/type-tests/@ember/template-compilation.ts
+++ b/type-tests/@ember/template-compilation.ts
@@ -1,0 +1,31 @@
+import { precompileTemplate } from '@ember/template-compilation';
+import { setComponentTemplate } from '@ember/component';
+import templateOnly from '@ember/component/template-only';
+
+import { expectTypeOf } from 'expect-type';
+
+// Valid usages
+precompileTemplate(`Hello World`, { moduleName: 'foo' });
+precompileTemplate(`Hello World`, { moduleName: 'foo', strictMode: false });
+precompileTemplate(`Hello World`, { strictMode: false });
+precompileTemplate(`Hello World`, { strictMode: true, scope: () => ({}) });
+precompileTemplate(`Hello World`, { strictMode: true, scope: () => ({ setComponentTemplate }) });
+precompileTemplate(`Hello World`, { strictMode: true, moduleName: 'hello', scope: () => ({}) });
+
+// Integration, since this is the primary use case for precompileTemplate
+expectTypeOf(setComponentTemplate(precompileTemplate(`Hello World`), templateOnly())).toBeObject();
+
+// @ts-expect-error scope is required when strictMode is true
+precompileTemplate(`Hello World`, { strictMode: true });
+
+// @ts-expect-error scope must be a function
+precompileTemplate(`Hello World`, { strictMode: true, scope: {} });
+
+// @ts-expect-error scope must return an object
+precompileTemplate(`Hello World`, { strictMode: true, scope: () => {} });
+
+// @ts-expect-error scope must return an object and arrays are not the kind of object we want
+precompileTemplate(`Hello World`, { strictMode: true, scope: () => [] });
+
+// @ts-expect-error scope has no purpose when strictMode is false
+precompileTemplate(`Hello World`, { strictMode: false, scope: () => ({}) });
--- a/types/preview/@ember/template-compilation/index.d.ts
+++ b/types/preview/@ember/template-compilation/index.d.ts
@@ -1,0 +1,28 @@
+/**
+ * References:
+ *  - https://github.com/emberjs/rfcs/pull/496
+ *  - https://github.com/emberjs/rfcs/pull/779
+ */
+declare module '@ember/template-compilation' {
+  import type { TemplateFactory } from '@ember/component/-private/glimmer-interfaces';
+
+  interface MaybeModuleName {
+    moduleName?: string;
+  }
+
+  interface LooseMode extends MaybeModuleName {
+    strictMode?: false;
+  }
+
+  interface StrictMode extends MaybeModuleName {
+    strictMode: true;
+    scope: () => Record<string, unknown>;
+  }
+
+  type PrecompileTemplateOptions = LooseMode | StrictMode;
+
+  export function precompileTemplate(
+    template: string,
+    options?: PrecompileTemplateOptions
+  ): TemplateFactory;
+}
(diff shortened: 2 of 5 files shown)


type-tests/@ember/template-compilation.ts
```diff
@@ -0,0 +1,28 @@
...
+// Valid usages
+precompileTemplate(`Hello World`, { moduleName: 'foo' });
+precompileTemplate(`Hello World`, { strictMode: true });
+precompileTemplate(`Hello World`, { strictMode: false });
+precompileTemplate(`Hello World`, { strictMode: true, scope: () => ({}) });
+precompileTemplate(`Hello World`, { strictMode: true, scope: () => ({ setComponentTemplate }) });
+precompileTemplate(`Hello World`, { strictMode: true, moduleName: 'hello', scope: () => ({}) });
+
+// Integration, since this is the primary use case for precompileTemplate
+expectTypeOf(setComponentTemplate(precompileTemplate(`Hello World`), templateOnly())).toBeObject();
```
NullVoxPopuli, line 16: is there any value in keeping expectTypeOf here since setComponentTemplate is already tested?

types/preview/@ember/template-compilation/index.d.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * References:
+ *  - https://github.com/emberjs/rfcs/pull/496
+ *  - https://github.com/emberjs/rfcs/pull/779
+ */
+declare module '@ember/template-compilation' {
+  interface TemplateFactory {
+    __htmlbars_inline_precompile_template_factory: any;
+  }
```
jamescdavis, line 9: We also have this branded interface defined here: https://github.com/emberjs/ember.js/blob/0c5518ea7b8eb6503e0d3d8a47153e0d619185be/types/preview/%40ember/component/index.d.ts#L11-L13

and here: https://github.com/emberjs/ember.js/blob/0c5518ea7b8eb6503e0d3d8a47153e0d619185be/types/preview/ember/index.d.ts#L374-L379

Seems like we should seek to have only one definition of this?

I notice we dropped the export when brining in [`htmlbars-inline-precompile` types from DT](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d38588fa1cd9fcab721c3ce78a24234b72d14ebc/types/htmlbars-inline-precompile/index.d.ts#L34-L36) (was never exported from [@ember/component](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d38588fa1cd9fcab721c3ce78a24234b72d14ebc/types/ember__component/index.d.ts#L19-L21)). I don't know that it necessarily makes sense to export from here, but this would make the third definition in these types.

If we _do_ export, I like this suggestion: 
> In the future, we will coordinate a change which makes this export only the type side of a class with a private property, such that it *cannot* be used.

from: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d38588fa1cd9fcab721c3ce78a24234b72d14ebc/types/htmlbars-inline-precompile/index.d.ts#L31-L33

@chriskrycho, @dfreeman thoughts?

jamescdavis: Maybe we should put it in https://github.com/emberjs/ember.js/blob/0c5518ea7b8eb6503e0d3d8a47153e0d619185be/types/preview/%40ember/component/-private/glimmer-interfaces.d.ts#L1

https://github.com/glimmerjs/glimmer-vm/blob/82ea580f8f4dcabfab98abc353b3a01891934f67/packages/%40glimmer/interfaces/lib/template.d.ts#L57

chriskrycho: Centralizing to a single definition seems good; I like that suggestion, @jamescdavis!

NullVoxPopuli: It this PR or separate?

I'd prefer separate so PRs can remain focused

chriskrycho: Separate seems good.

chriskrycho: Ah, no, actually. 😂 Let's not *introduce another copy* here. Pull it from `ember.js/types/preview/@ember/component/-private/glimmer-interfaces.d.ts`, as @jamescdavis suggested. That isn't in conflict with PRs remaining focused—quite the contrary!

types/preview/@ember/template-compilation/index.d.ts
```diff
@@ -0,0 +1,26 @@
...
+  type PrecompileTemplateOptions =
+    | {
+        strictMode?: false;
+        moduleName?: string;
+      }
+    | {
+        strictMode: true;
+        moduleName?: string;
+        scope: () => Record<string, unknown>;
+      };
```
chriskrycho, line 20: I have a mild preference for this layout, which is clearer about the shared code and what exactly the options are:

```suggestion
interface MaybeModuleName {
  moduleName?: string;
}

interface LooseMode extends MaybeModuleName {
  strictMode?: false;
}

interface StrictMode extends MaybeModuleName {
  strictMode: true;
  scope: () => Record<string, unknown>;
}

type PrecompileTemplateOptions = LooseMode | StrictMode;
```

chriskrycho (changes requested):
A couple suggested changes!

types/preview/@ember/template-compilation/index.d.ts
NullVoxPopuli: done and pushed!

types/preview/@ember/template-compilation/index.d.ts
NullVoxPopuli: applied!

type-tests/@ember/template-compilation.ts
chriskrycho: Seems fine to keep it here as well.

chriskrycho (approved):
Thank you!

boris-petrov:
This is missing the `compileTemplate` function. Is this on purpose?

NullVoxPopuli:
Is it public api for `@ember/template-compilation`?

It's from a different package: https://github.com/emberjs/ember.js/blob/main/packages/%40ember/template-compilation/index.ts#L4
Which doesn't have docs on api.emberjs.com

chriskrycho:
It is indeed public API, and in fact is recommended by a *different* public API:

https://github.com/emberjs/ember.js/blob/5de3020d0e070efd7a800c38bc750616f7f6eab2/packages/%40ember/template-compilation/index.ts#L35

We can probably do a backport for the 4.12 LTS types as well stable; it is not relevant for 5.1 and after, where the types are present by way of compiling from course.

PR #20411, opened 2023-03-16
Set volta config so folks with volta don't use the wrong version of yarn

NullVoxPopuli:
With the current version of volta, when I installed yarn globally, I got a goofy version:
```
❯ yarn --version
4.0.0-rc.39
```
having a non-v1 version of yarn as default has been allowing me to notice which projects don't have a volta config :sweat_smile:


chriskrycho (approved):
👍🏼

PR #20417, opened 2023-03-22
Fix types for `Location` after deprecation removal

chriskrycho:
1. Make the exported interface the default export in both the internal definitions and the exported types.

2. Move the internal type tests to the general type tests package so we guarantee that our published types match our internals.

3. Update the preview types import list, removing the `auto-location` import and including the `location` import so it is visible in the exported types.

4. Add more test coverage to types to confirm that the DI registry has all of them.

5. Test both internal and published types in CI, so that we don't have misses of the sort which led to previous commits landing with types in a broken state.


wagenet (commented):
Changes look reasonable. I did not check to see if they are actually correct.

PR #20429, opened 2023-04-03
[BUGFIX stable] TS lint-level error in `mixins/-proxy.ts`

chriskrycho:
Explicitly do a consistent return from a number of spots which TS 5.1 nightlies are catching, and which previous versions did not. In these cases, we explicitly return `undefined`, which is equivalent to the previous behavior.

Fixes a lint-style error with TS nightly.


dfreeman:
Looks like there are two more instances flagged in `@ember/debug` and `@ember/object`

dfreeman (approved):

PR #20435, opened 2023-04-08
Custom locations are a supported public API

ef4:
See docs at:

https://github.com/emberjs/ember.js/blob/60d2e0cddb353aea0d6e36a72fda971010d92355/packages/%40ember/routing/location.ts#L31-L33

If you create `app/locations/whatever.js` you can set Router.location to `"whatever"` to resolve it.

In addition to using the resolver in this way with a custom string, you can avoid stringly resolution by passing a Location instance directly.

### DIFF
--- a/types/preview/@ember/routing/router.d.ts
+++ b/types/preview/@ember/routing/router.d.ts
@@ -3,6 +3,7 @@ declare module '@ember/routing/router' {
   import Evented from '@ember/object/evented';
   import RouterDSL from '@ember/routing/-private/router-dsl';
   import Transition from '@ember/routing/transition';
+  import Location from '@ember/routing/location';
 
   /**
    * The `Ember.Router` class manages the application state and URLs. Refer to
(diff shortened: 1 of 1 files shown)


types/preview/@ember/routing/router.d.ts
```diff
@@ -21,7 +22,7 @@ declare module '@ember/routing/router' {
      *
      * @note the `'auto'` location is [deprecated](https://deprecations.emberjs.com/v4.x/#toc_deprecate-auto-location).
      */
-    location: 'history' | 'hash' | 'none' | 'auto';
+    location: 'history' | 'hash' | 'none' | 'auto' | Location | string;
```
chriskrycho, line 25: The string literals and `| string` are redundant, so let's do this:

```suggestion
    location: Location | string;
```

Also, we should make sure to update the types on the implementation as well.

chriskrycho (changes requested):
Ah, cool. Wasn't clear from the docs, so this is handy; we need to update the implementation as well.

kategengler:
Closing; this is very out of date and from a quick look I believe the public types support this

PR #20436, opened 2023-04-12
Update `master` to `main` where used

kategengler:
Some CI is currently broken without this change.


kategengler:
Merging over failing typescript build because that is failing on main.

PR #20438, opened 2023-04-12
[BREAKING BUGFIX beta] Drop support for Node 14

rwjblue:
Node 14 is EOL'ed as of 2023-04-30, which will happen **before** 5.0.0 stable will be out, so IMO we should probably just land it now as part of 5.0.


kategengler (approved):

kategengler:
The type job failure is unrelated.

Issue #19891, opened 2022-01-07
[Bug] LinkTo inside inline SVG reloads application


vihai:
### 🐞 Describe the Bug
A LinkTo component inside an inline SVG correctly produces an <a> element, however when clicked the application is reloaded instead of transitioning. Seems related to #19546

### 🔬 Minimal Reproduction
```
<svg version="1.1" viewBox="0 0 200 100" xml:space="preserve" xmlns="http://www.w3.org/2000/svg">
  <LinkTo @route="my.route">
    <rect x="0" y="0" width="200" height="100"></rect>
  </LinkTo>
</svg>
```

### 🌍 Environment

- Ember: 4.1
- Node.js/npm: 16.13.1
- OS: Ubuntu 21.10
- Browser: Firefox 95 / Chrome 97

johanrd:
Still reproducible in ember 6.11

@MelSumner [comment on #20443](https://github.com/emberjs/ember.js/pull/20443#issuecomment-1522193378) 
> Question: Is it valid HTML to put a link inside of an SVG? Typically it should be the other way around. Additionally, an SVG is not eligible to receive interactions (but a link is).

Yes, it seems valid. The SVG spec defines its own https://developer.mozilla.org/en-US/docs/Web/SVG/Element/a which is a container element that can wrap shapes, text, and other SVG content to make them clickable. This is standard, well-supported SVG:
```html
<svg viewBox="0 0 200 100" xmlns="http://www.w3.org/2000/svg">
  <a href="/some-page">
    <rect x="0" y="0" width="200" height="100" />
  </a>
</svg>
```

### Diagnosis
The problem seems to be:
https://github.com/emberjs/ember.js/blob/02c62b0d7cbec49271949fcf608e66f1ccb5b3a6/packages/%40ember/-internals/glimmer/lib/components/link-to.ts#L371

When `<LinkTo>` renders `<a>` inside an `<svg>`, the browser creates an SVGAElement (not an HTMLAnchorElement), because elements inside `<svg>` are in the SVG namespace. The assertion fails, and preventDefault is never called, so the browser does a full navigation instead of an Ember transition.

kategengler:
Fixed in #21107

PR #20443, opened 2023-04-18
[BUGFIX beta] preventDefault in SVG LinkTo

Recour:
When wrapping `LinkTo` in an SVG element, the click `event.currentTarget` is not an `instanceof HTMLAnchorElement` but of `SVGAElement`. `preventDefault` is only triggered when `isSelf`, but it needs to be triggered when wrapped inside an SVG as well, otherwise clicking the element causes a reload.

Fix for https://github.com/emberjs/ember.js/issues/19891


MelSumner:
Question: Is it valid HTML to put a link inside of an SVG?

PR #20445, opened 2023-04-25
[CLEANUP beta] Update browser support per RFC #685

kategengler:
https://rfcs.emberjs.com/id/0685-new-browser-support-policy

May be worth considering an update to the policy to drop Safari versions with < 0.25% share.


chriskrycho (approved):
👍🏼 This looks good. I'd likewise be in favor of dropping older Safari with < 0.25% usage (I personally would be fine with a 1% cutoff but different folks may have different tolerance levels).

chriskrycho:
To expand on that: `>0.25% and not dead` would still have us supporting mobile Safari 12, though, at 0.38%. This is almost certainly due to a "cliff" in device support (Apple's overall OS release support is long, tends to have those kinds of cliffs); mobile Safari 12.2–12.5 actually has higher usage than 14.0–14.4 (though lower than 14.5–14.8), so changing the policy to that wouldn't change what we target.

kategengler:
The RFC is hardcoded at Safari 12, so if we do eventually want to drop, we need to consider/pay attention to it.

chriskrycho:
Yep, I think if we revisit we should do it in terms of a % point, while preserving the notion of non-evergreen browsers not dropping support. The 12.x iOS Safari versions are now less than a third of the usage they were when we adopted the current policy, so I think it would be perfectly reasonable to bump the target—notably, even just targeting 0.5% would bump all the way up to 15.6. (I don't have a strong opinion that we *should* do that, but I do think we *could* do that!)

wycats:
I agree that we should consider a usage-based cutoff for Safari (the one "non-evergreen" exception in our current RFC).

Might we also want to consider a cutoff based on official Apple support?

Catalina came out in 2019 and is [no longer officially maintained](https://endoflife.date/macos), but as far as I can tell, we're still supporting [the version of Safari that came with Mojave](https://github.com/emberjs/ember.js/blob/main/testem.browserstack.js#L10), which hasn't received an update since 2019.

iOS 12 (which came with Safari 12), on the other hand, is [still limping along](https://endoflife.date/ios) with security updates, but has been out of active support since 2019.

All of these details conspire to keep Safari 12 above 0.25% but consistently below 0.5%.

> notably, even just targeting 0.5% would bump all the way up to 15.6.

In the case of Safari, would we want to say that the number is the cumulative number of users using <= the target version?

While each of 12 and 14, on their own, aren't above 0.5%, more than 0.5% of mobile users use a version of Safari "no newer than Safari 14." In the special case of non-evergreen browsers, isn't that the intuitively correct metric?

PR #20447, opened 2023-04-27
[BUGFIX LTS 3.28] Don't run getters while applying mixins

wycats:
Backports #20388

This change ensures that getters are never evaluated while applying
mixins.

It relies on the fact that all getters (including undecorated ones) get
converted into classic decorators when the mixin is originally created.

### DIFF
--- a/package.json
+++ b/package.json
@@ -136,7 +136,7 @@
     "mkdirp": "^1.0.4",
     "mocha": "^7.1.2",
     "prettier": "^2.1.2",
-    "puppeteer": "^3.0.4",
+    "puppeteer": "^13.5.1",
     "qunit": "^2.14.0",
     "rollup-plugin-commonjs": "^9.3.4",
     "rollup-plugin-node-resolve": "^4.2.4",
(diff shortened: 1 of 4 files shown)


wycats:
@kategengler I backported a bunch of the CI changes to get the tests passing but I'm not sure if I did it correctly. Can you give them a quick once over?

package.json
```diff
@@ -150,7 +150,7 @@
     "tslint": "^5.20.1"
   },
   "engines": {
-    "node": "10.* || >= 12.*"
+    "node": ">= 14.*"
```
kategengler, line 153: We definitely cannot update engines in 3.28 point release

kategengler:
@wycats What about CI was failing? We typically cherry-pick just commits back to the older versions and CI usually keeps working.

wycats:
@kategengler I'll take another look at it. I think it might be tricky to avoid upgrades without somehow pinning other infrastructure like Embroider stuff, which we also might not want to do.

kategengler (approved):

PR #20455, opened 2023-05-01
Fix s3 publishing on Node 16/npm 8

kategengler:
Pull in s3 publishing from ember-publisher and update. 
Convert the publish-to-s3 script to esm.

Change the s3 publishing config to no longer strip the +sha info from the version. This is the actual source of the publishing failure and it seems down to a change in the name of the tarball in npm 6 vs npm 8.


jenweber (approved):
Thank you Katie for this fix! It seems like maybe the best way to test this is to give it a try. Let me know if there’s anything else you want in terms of a review or testing.

PR #20461, opened 2023-05-16
[BUGFIX LTS] Fix cyclic module crash

ef4:
`@ember/engine/index` and `@ember/engine/instance` have circular dependencies on each other.

This would be OK, except that `instance.ts` is eagerly copying a class off `index.ts`, at a point in time where it might not actually be available.

Whether this bug hits you depends on which side of the pair gets evaluated first. I only happened to hit it when testing an app with engines and fastboot on Ember 5. The exception looks like:

```
TypeError: Right-hand side of 'instanceof' is not an object
```

The fix is to not eagerly pull the class off the module.

### DIFF
--- a/packages/@ember/engine/instance.ts
+++ b/packages/@ember/engine/instance.ts
@@ -17,8 +17,6 @@ import type Application from '@ember/application';
 import type { BootEnvironment } from '@ember/-internals/glimmer';
 import type { SimpleElement } from '@simple-dom/interface';
 
-const CEngine = Engine;
-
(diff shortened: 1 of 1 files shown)


MelSumner (approved):
I like the improved clarity of the naming here as well. 👍

chriskrycho (approved):
Looks good. Looks like this should probably be `BUGFIX LTS`?

ef4:
Rewrote the commit and the PR to say "BUGFIX LTS".

PR #20480, opened 2023-06-28
[BUGFIX stable] Fix type dependency issues

### LINKED ISSUE #20479: [Bug] Types in 5.1 refer to missing dependencies
### 🐞 Describe the Bug

The new stable types in ember-source 5.1 import from `@simple-dom/interface` and `@glimmer/syntax`, neither of which is a declared dependency that's guaranteed to be present in an app.

Both of them are interior dependencies of ember's devDependencies.

### 🔬 Minimal Reproduction

You can see this failing in embroider CI here:

https://github.com/embroider-build/embroider/actions/runs/5391837867/jobs/9789371760

chriskrycho:
Add `@simple-dom/interface` and `@glimmer/syntax` to `dependencies` so they are present in the graph for consumers who will not be getting them transitively from the `@glimmer/*` deps which are pre-built into Ember. (Medium-term, it would be preferable for our type emit to roll up these kinds of dependencies the same way Ember's build does *or* for Ember's build to stop rolling them up; whichever we choose, the goal will be for them to *match*.)
Fixes #20479.


NullVoxPopuli (approved):
tyty

PR #20482, opened 2023-06-28
[BUGFIX stable] Allow `store` to be injected on `Route`

chriskrycho:
In the public types, TypeScript (correctly) complains when users try to inject the store with `@service store` on a `Route` subclass, because the item is an accessor on `Route` itself, and the service injection "happens" to work only because the `set store(value)` smashes itself as well as the corresponding `get store()` using `defineProperty`.

Work around this by renaming the getter in question to `_store` and updating the one usage to check for a `store` property before falling back to the `_store` property. Happily, this also dovetails with the plan to deprecate the automatic `findModel` behavior this exists to support: once that is removed, this entire code path can be cleaned up trivially.

### DIFF
--- a/packages/@ember/routing/route.ts
+++ b/packages/@ember/routing/route.ts
@@ -1234,7 +1234,13 @@ class Route<Model = unknown> extends EmberObject.extend(ActionHandler, Evented) 
     @private
   */
   findModel(...args: any[]) {
-    return (get(this, 'store') as any).find(...args);
+    // SAFETY: it's all absurd lies; there is no explicit contract for `store`
+    // and we allow people to register *anything* here and we call it: GLHF! The
+    // fallback path here means we correctly handle the case where there is no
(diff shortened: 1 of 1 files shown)


tests/docs/expected.js
```diff
@@ -484,7 +484,7 @@ module.exports = {
     'sort',
     'sortBy',
     'startRouting',
-    'store',
+    '_store',
```
gitKrystan, line 487: ELI5 this file?

chriskrycho: Every name Ember exports gets checked here as a suuuuper basic "smoke test" to make sure we don't accidentally stop publishing something.

chriskrycho: Also, there is apparently some *other* usage of `store` so I have reverted this change. 😂

wagenet (approved):

PR #20490, opened 2023-06-29
[BUGFIX stable] Fix model arg to a Route `serialize`

### LINKED ISSUE #20487: [Bug] `undefined` is required in a route's `serialize` method in the stable types
### 🐞 Describe the Bug
Check the repro.

```
app/routes/foo.ts:8:19 - error TS2416: Property 'serialize' in type 'FooRoute<T>' is not assignable to the same property in base type 'Route<T>'.
  Type '(object: Child) => any' is not assignable to type '(model: T | undefined, params: string[]) => { [key: string]: unknown; } | undefined'.
    Types of parameters 'object' and 'model' are incompatible.
      Type 'T | undefined' is not assignable to type 'Child'.
        Type 'undefined' is not assignable to type 'Model'.

8   public override serialize(object: Child): any {
                    ~~~~~~~~~
```

### 🔬 Minimal Reproduction
[Repo](https://github.com/boris-petrov/ember-type-bugs).

Run `npm install` and then `./node_modules/.bin/tsc`.

### 😕 Actual Behavior
Type error on `serialize`.

### 🤔 Expected Behavior
No error. Removing the last commit (which upgrades Ember) and rerunning `npm install` and then `tsc` will remove the errors.

### 🌍 Environment

- Ember: 5.1.0
- Node.js/npm: 20.3.1/9.7.2
- OS: Linux
- Browser: N/A

cc @chriskrycho

chriskrycho:
Previously, the internal type here always explicitly added `| undefined` to the type of the `model` param. This was likely reasonable when the conversion first happened, since the type here was simply a minimal `IModel | undefined`. Since the param is now driven by the type param for the `Route`, though, and defaults to `unknown`, we can just use the `Model` type param directly, and the implementation handles it correctly since the was actually already, albeit invisibly, just `unknown`, since `unknown | undefined` === `unknown`.

Fixes #20487.

### DIFF
--- a/packages/@ember/routing/route.ts
+++ b/packages/@ember/routing/route.ts
@@ -335,7 +335,7 @@ class Route<Model = unknown> extends EmberObject.extend(ActionHandler, Evented) 
     @since 1.0.0
     @public
   */
-  serialize(model: Model | undefined, params: string[]): { [key: string]: unknown } | undefined {
+  serialize(model: Model, params: string[]): { [key: string]: unknown } | undefined {
     if (params.length < 1 || !model) {
       return;
     }
--- a/type-tests/@ember/routing-test/route.ts
+++ b/type-tests/@ember/routing-test/route.ts
@@ -35,6 +35,27 @@ class ModelTest extends Route {
   }
 }
 
+class Model {
+  baseClass = true;
+}
+
+class Child extends Model {
+  baseClass = false;
+  extraStuff = true;
+}
+
+class FooRoute<T extends Model> extends Route<T> {
+  serialize(object: T): any {
+    return null;
+  }
+}
+
+class FooChildRoute extends FooRoute<Child> {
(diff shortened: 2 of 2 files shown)


packages/@ember/routing/route.ts
```diff
@@ -335,7 +335,7 @@ class Route<Model = unknown> extends EmberObject.extend(ActionHandler, Evented)
     @since 1.0.0
     @public
   */
-  serialize(model: Model | undefined, params: string[]): { [key: string]: unknown } | undefined {
+  serialize(model: Model, params: string[]): { [key: string]: unknown } | undefined {
     if (params.length < 1 || !model) {
```
boris-petrov, line 339: Perhaps this `!model` check is not needed any more in this case?

gitKrystan (approved):

PR #20507, opened 2023-07-22
Adding stage3-compatible decorator implementations

ef4:
(This should stay draft and not merge until we have some consensus on the ecosystem-spanning parts of the migration.)

This is an exploration to see how easy it is to make the ember-provided decorators work under both legacy decorators and stage3 decorators.

(I don't think it's hard!)

The first commit here implements `@tracked`. I tested it in an application with the configuration:

```js
 const app = new EmberApp(defaults, {
    'ember-cli-babel': {
      disableDecoratorTransforms: true,
    },
    babel: {
      plugins: [
        [
          '@babel/plugin-proposal-decorators',
          {
            version: '2023-05',
          },
        ],
      ],
    },
  });
```

With those settings, `@tracked foo` becomes a runtime error and will tell you to use `@tracked accessor foo` instead, and when you do it all works correctly.

### DIFF
--- a/packages/@ember/-internals/metal/lib/decorator-util.ts
+++ b/packages/@ember/-internals/metal/lib/decorator-util.ts
@@ -78,9 +78,21 @@ export type ClassAutoAccessorDecorator = (
   init?: (initialValue: unknown) => unknown;
 } | void;
 
+export type Decorator =
+  | ClassMethodDecorator
+  | ClassGetterDecorator
+  | ClassSetterDecorator
+  | ClassFieldDecorator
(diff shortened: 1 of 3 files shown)


kategengler:

ef4:
Still planning to pick this back up. Now that the build refactor has landed it's a bit easier to test.

ef4:
Also: https://github.com/beerinho/ember-glob-import-demo

ef4:
Superseded by https://github.com/emberjs/ember.js/pull/20868

Issue #20362, opened 2023-01-28
Test TypeScript blueprints using TypeScript enabled Ember apps


bertdeblock:
Even though we had enabled the `EMBER_TYPESCRIPT_BLUEPRINTS` flag when running blueprint tests, and have now enabled that flag by default, we still don't actually verify the generation of TypeScript files. That is because all the Ember apps during tests are generated _without_ the `--typescript` flag. To test the TypeScript blueprints correctly, we must use TypeScript enabled Ember apps.

~~NOTE: We don't necessarily have to create these apps using the `--typescript` flag, as we would run into https://github.com/ember-cli/ember-cli/issues/10045 I think (would make running those tests a lot slower). Making sure that the `isTypeScriptProject` option in the `.ember-cli` file is set to `true` should be sufficient as well I think.~~

PR #20511, opened 2023-07-27
[FEATURE] Add template tag component blueprints via --strict flag as per RFC #779

IgnaceMaes:
Implements step one of the blueprint work as per [RFC #779](https://rfcs.emberjs.com/id/0779-first-class-component-templates/#blueprints)

Allows e.g.:

```sh
ember generate component foo-bar --strict
```

Prior syntax is still enabled by default, or by passing the loose flag:

```sh
ember generate component foo-bar --loose
```


NullVoxPopuli:
very excited for this!

blueprints/component/index.js
```diff
@@ -51,6 +51,12 @@ module.exports = {
       default: OCTANE ? 'flat' : 'classic',
       aliases: OCTANE ? [{ fs: 'flat' }, { ns: 'nested' }, { cs: 'classic' }] : [{ cs: 'classic' }],
     },
+    {
+      name: 'authoring-format',
+      type: ['loose', 'strict'],
+      default: 'loose',
+      aliases: [{ loose: 'loose' }, { strict: 'strict' }],
```
IgnaceMaes, line 58: Would it be accepted to add a `template-tag` + `tt` alias that map to `strict`? As most folks don't end up knowing the loose/strict mode terminology but rather the user facing name template tag.

node-tests/blueprints/component-test.js
```diff
@@ -324,6 +329,21 @@ describe('Blueprint: component', function () {
       );
     });
 
+    it('component foo --strict', function () {
+      return emberGenerateDestroy(['component', 'foo', '--strict'], (_file) => {
+        expect(_file('app/components/foo.gjs')).to.equal(templateTagContents);
```
IgnaceMaes, line 334: This is currently failing for 

```
yarn test:blueprints:ts
```

If someone could point me in the direction where TypeScript test files are converted to js that would be helpful. I suspect it's not handling `.gts` yet.

node-tests/blueprints/component-test.js
```diff
@@ -324,6 +329,21 @@ describe('Blueprint: component', function () {
...
+
+        expect(_file('tests/integration/components/foo-test.gjs')).to.equal(
```
IgnaceMaes, line 336: The test file isn't being created when generating a component in this test:

```
  288 passing (2m)
  1 failing

  1) Blueprint: component
       in app - octane
         component foo --strict:
     AssertionError: expected "tests/integration/components/foo-test.gjs" to exist

parent path "tests/integration/components" does not exist.
parent path "tests/integration" exists and contains:
- .gitkeep
```

When trying out the local version using `yarn link` it does however generate the test file correctly. Anything I'm missing here?

node-tests/blueprints/component-test.js
bertdeblock: https://github.com/ember-cli/ember-cli/blob/master/lib/models/blueprint.js#L476 maybe?

IgnaceMaes: The logic being in a different repository would explain a lot 😅 

Thanks for sharing, I'll fiddle around with that and see if I can get it working.

blueprints/component/index.js
IgnaceMaes: Added these as I feel it makes most sense. Feel free to reopen on disagreement.

kategengler:
@IgnaceMaes What is the status here?

IgnaceMaes:
@kategengler This should be good to go, functionality wise. But I could use some guidance on how to move this ahead.

This MR depended on https://github.com/ember-cli/ember-cli/pull/10332 which got merged and released in `ember-cli@5.5`. (yey 🎉) But should this functionality be merged, considering it now requires a peerDependency on `ember-cli@^5.5.0`? Should it be backported in `ember-cli` to older LTS's? Or can it just be merged?

node-tests/blueprints/component-test.js
IgnaceMaes: Fixed upstream in ember-cli: https://github.com/ember-cli/ember-cli/pull/10332

kategengler:
Does it work without the `--strict` flag for earlier ember-cli?

blueprints-js/component-test/qunit-rfc-232-files/__root__/__testType__/__path__/__test__.gjs
```diff
@@ -0,0 +1,36 @@
+<% if (testType === 'integration') { %>import { module, test } from 'qunit';
+import { setupRenderingTest } from '<%= modulePrefix %>/tests/helpers';
+import { render } from '@ember/test-helpers';
+import <%= componentName %> from '<%= modulePrefix %>/components/<%= componentPathName %>';
+
+module('<%= friendlyTestDescription %>', function (hooks) {
+  setupRenderingTest(hooks);
+
+  test('it renders', async function (assert) {
+    // Set any properties with this.set('myProperty', 'value');
```
IgnaceMaes, line 10: These comments should be updated. Using `this` in template tag tests is not recommended as the context within the component would refer to the backing class instead.

IgnaceMaes: Updated to reflect test waiter syntax in https://rfcs.emberjs.com/id/0785-remove-set-get-in-tests/

node-tests/blueprints/component-test-test.js
```diff
@@ -38,6 +38,14 @@ describe('Blueprint: component-test', function () {
           );
         });
       });
+
+      it('component-test x-foo --strict', function () {
+        return emberGenerateDestroy(['component-test', 'x-foo', '--strict'], (_file) => {
+          expect(_file('tests/integration/components/x-foo-test.gjs')).to.equal(
+            fixture('component-test/rfc232.gjs')
```
IgnaceMaes, line 45: This fails when running the TypeScript blueprint tests:

```
  1) Blueprint: component-test
       in app
         with default setup
           component-test x-foo --strict:
     AssertionError: expected "tests/integration/components/x-foo-test.gjs" to exist

parent path "tests/integration/components" exists and contains:
- x-foo-test.gts
      at /.../ember.js/node-tests/blueprints/component-test-test.js:44:75
      at /.../ember.js/node_modules/.pnpm/ember-cli-blueprint-test-helpers@0.19.2/node_modules/ember-cli-blueprint-test-helpers/lib/ember-generate-destroy.js:34:17
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

I'm not sure what the next step is here. The other tests don't check on TypeScript-specific files. Is the expect assumed to automatically work for TS too, but it doesn't here?

Welcoming any input 😄

IgnaceMaes:
> Does it work without the `--strict` flag for earlier ember-cli?

It does! So I suppose this is fine then.

node-tests/blueprints/component-test-test.js
IgnaceMaes: Validating the path dynamically would work:

```js
const filePath = fs.existsSync('tests/integration/components/x-foo-test.gts')
  ? 'tests/integration/components/x-foo-test.gts'
  : 'tests/integration/components/x-foo-test.gjs';
```

But that seems like logic which shouldn't be in the test.

kategengler: You're getting TypeScript output when the TypeScript blueprints should be converted to JS by the blueprint system (unless it is a typescript project). 

I believe this https://github.com/ember-polyfills/ember-cli-typescript-blueprint-polyfill/blob/main/src/utils.js#L13-L15 needs to include gts, but there may be other changes needed for the blueprint system to convert gts to gjs.

Windvis: I can try to find some time to push this over the finish line as well :+1:

IgnaceMaes: That would be much appreciated. Realistically, I won't have time to look into this in the near future :/

Windvis: Ok, that was confusing.. :smile: 

I [updated the polyfill](https://github.com/ember-polyfills/ember-cli-typescript-blueprint-polyfill/pull/5) and used the branch here, but that still didn't work because the polyfill [disables itself](https://github.com/ember-polyfills/ember-cli-typescript-blueprint-polyfill/blob/8bede4c079219d3f7a68fedabed08296c89460a0/src/typescript-blueprint-polyfill.js#L11-L13) if it thinks it's no longer needed. It seems we are depending on [ember-cli 4.10](https://github.com/emberjs/ember.js/blob/0f9b3fcebb9d813f438a946f4b7d253e11311672/package.json#L130) which _does_ support .ts -> .js in the blueprints, but not .gts -> .gjs (which was released in 5.5 I think). Once I bump the version to 5.12 the test no longer seems to fail.

What was extra confusing is that there are no tests that verify the .ts version of the blueprints: https://github.com/emberjs/ember.js/issues/20362

blueprints/component/files/__root__/__templatepath__/__templatename__.gts
```diff
@@ -0,0 +1,3 @@
+<template>
+  {{yield}}
+</template>
```
Windvis, line 3: I think we can TO types already: https://typed-ember.gitbook.io/glint/environments/ember/template-only-components

Windvis:
In Octane the component-class defaults to `--no-component-class`. I'm wondering if that's the most useful default for .gts components though, especially since switching from template only .gts files to class based ones isn't as easy due to the types.

Since this is a new flag, and the RFC doesn't specify it, would it make sense to generate a class based .gts file by default, and a TO version if the `--component-class=@ember/component/template-only` is provided?

In any case, even if we stick with the "TO" by default, we should probably still generate a class based version if a class is provided (and throw an error if it's not glimmer)?

NullVoxPopuli:
> isn't as easy due to the types.

I think we can do:
```glimmer-ts
<template>
 ...
</template> satisfies TOC<...>
```

NullVoxPopuli:
Linting for gjs/gts has landed in ember-cli

PR #20521, opened 2023-08-09
[FEATURE] Add TypeScript support for looking up controllers in DI registry

sergey-panov:
This PR enables controller type discovery through `Owner::lookup` method:

```ts
declare module '@ember/controller' {
  interface Registry {
    foo: FooController;
  }
}

const fooController = this.owner.lookup('controller:foo'); // TS resolves fooController's type as FooController.
const barController = this.owner.lookup('controller:bar'); // If not specified via Registry interface, will resolve type as Controller | undefined
```

Complementary PR to `@types/ember__controller`: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66067

### DIFF
--- a/packages/@ember/controller/index.ts
+++ b/packages/@ember/controller/index.ts
@@ -387,7 +387,7 @@ export { Controller as default, ControllerMixin };
   }
   ```
 
-  Then `@service` can check that the service is registered correctly, and APIs
+  Then `@inject` can check that the service is registered correctly, and APIs
   like `owner.lookup('controller:example')` can return `ExampleController`.
 */
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
--- a/packages/@ember/controller/index.ts
+++ b/packages/@ember/controller/index.ts
@@ -366,3 +366,29 @@ export function inject(
 }
 
 export { Controller as default, ControllerMixin };
+
+/**
+  A type registry for Ember `Controller`s. Meant to be declaration-merged so string
+  lookups resolve to the correct type.
+
+  Blueprints should include such a declaration merge for TypeScript:
+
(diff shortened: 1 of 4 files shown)


chriskrycho:
We'll discuss, and I expect we'll merge this, since we have the corresponding APIs covered elsewhere. I want to say, however: I didn't implement it for a reason, which is that folks should absolutely move away from using controller lookup like this, for a host of reasons, most of all that controller lifecycle is weird and that 99% of things people do with controllers should be done with routes, services, and components instead. (The only real exception is query param management!) Like I said, though: since the API is there, I expect we will likely support it.

While we are discussing, folks should be aware that you can implement this yourself in any app or addon *without* needing official support. (This is the other part of the reason I did not prioritize implementing it when adding general DI registry support for `lookup`.)

packages/@ember/controller/index.ts
```diff
@@ -366,3 +366,29 @@ export function inject(
...
+  }
+
+  declare module '@ember/controller' {
+    export interface Registry {
+      example: ExampleController;
+    }
+  }
+  ```
+
+  Then `@service` can check that the service is registered correctly, and APIs
```
chriskrycho, line 390:
```suggestion
  Then `@inject` can check that the service is registered correctly, and APIs
```

chriskrycho (changes requested):
I'm making one tweak to the docs and then we can land this; note that I'm also tweaking what it lands as—a feature, not a bug fix. The previous types were *correct* here—they would never have caused a runtime error—so this is best flagged as a new capability.

Do please consider migrating away from controller injections to using service injections or passing arguments instead, as that is much more aligned with where the framework is headed!

PR #20522, opened 2023-08-13
Use maintained fork of `eslint-plugin-node`

voxpelli:
Since it can be hard for users of `eslint-plugin-node` to find out that there's a maintained fork from the official [ESLint Community project](https://github.com/eslint-community) I'm doing a drive where I'm making PR:s to some projects to raise awareness.

See: https://github.com/eslint-community/eslint-plugin-n/issues/100

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -151,8 +151,8 @@ module.exports = {
         es6: true,
       },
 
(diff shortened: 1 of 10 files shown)


NullVoxPopuli:
thank you!!! this is helpful <3

PR #20376, opened 2023-02-17
[RFC-774] Deprecate implicit record loading in Ember Route

wagenet:
[RFC-774](https://github.com/emberjs/rfcs/pull/774)

Deprecation App: https://github.com/ember-learn/deprecation-app/pull/1334

PR #20524, opened 2023-08-17
[BUGFIX beta] Add url to `deprecate-implicit-route-model` deprecation

bertdeblock:
Addresses https://github.com/emberjs/ember.js/pull/20376#discussion_r1286038077.

### DIFF
--- a/packages/@ember/routing/route.ts
+++ b/packages/@ember/routing/route.ts
@@ -1251,8 +1251,9 @@ class Route<Model = unknown> extends EmberObject.extend(ActionHandler, Evented) 
       {
         id: 'deprecate-implicit-route-model',
         for: 'ember-source',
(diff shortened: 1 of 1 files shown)


packages/@ember/routing/route.ts
```diff
@@ -1251,8 +1251,9 @@ class Route<Model = unknown> extends EmberObject.extend(ActionHandler, Evented)
       {
         id: 'deprecate-implicit-route-model',
         for: 'ember-source',
-        since: { available: '5.3.0' },
+        since: { available: '5.3.0', enabled: '5.3.0' },
```
bertdeblock, line 1254: Added `enabled` as well, since it will be enabled right away when released I think.

kategengler:
thank you!!

PR #20505, opened 2023-07-19
[BUGFIX release] Remove use of `this.element` in `component-test` and `helper-test` blueprints

bertdeblock:
See https://github.com/emberjs/ember-qunit/blob/master/docs/migration.md#migrating-to-native-typescript-support-in-v610.

PR #20527, opened 2023-08-19
Do not use `this.element` in `renderSettled` code example

bertdeblock:
Similar to https://github.com/emberjs/ember.js/pull/20505.


NullVoxPopuli (approved):

kategengler:
Thank you! 

Merging over the failed browserstack test because we have a flaky one that is now reliably failing and this is just docs.

PR #20543, opened 2023-09-08
Add JSDoc for @cached for the built-in types

NullVoxPopuli:
Context: https://github.com/emberjs/rfcs/pull/914#issuecomment-1712072164

The docs [we have for `@cached`](https://github.com/emberjs/ember.js/blob/main/packages/%40glimmer/tracking/index.ts#L151) already or just for the docs.
This PR adds JSDoc for `@cached` that'll show up in folks' editors.

Demo here: https://github.com/NullVoxPopuli/cached-status-2023-09-08/blob/main/app/app.ts#L14
(provided local linking is done)

image

### DIFF
--- a/packages/@ember/-internals/metal/lib/cached.ts
+++ b/packages/@ember/-internals/metal/lib/cached.ts
@@ -7,35 +7,96 @@ import { createCache, getValue } from '@glimmer/validator';
 /**
  * @decorator
  *
- * Memoizes the result of a getter based on autotracking.
- *
- * The `@cached` decorator can be used on native getters to memoize their return
- * values based on the tracked state they consume while being calculated.
- *
- * By default a getter is always re-computed every time it is accessed. On
- * average this is faster than caching every getter result by default.
- *
- * However, there are absolutely cases where getters are expensive, and their
- * values are used repeatedly, so memoization would be very helpful.
- * Strategic, opt-in memoization is a useful tool that helps developers
- * optimize their apps when relevant, without adding extra overhead unless
- * necessary.
- *
- * @example
- *
- * ```ts
- * import { tracked, cached } from '@glimmer/tracking';
- *
- * class Person {
- *   @tracked firstName = 'Jen';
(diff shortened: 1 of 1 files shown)


packages/@ember/-internals/metal/lib/cached.ts
```diff
@@ -4,7 +4,100 @@
...
+  ```
+
+  Here equal values won't update the property, therefore not triggering
+  the subsequent cache invalidations of the `@cached` properties who were
+  using this `trackedProp`.
+
+  As a reminder, do not use this piece of code inside a tracked getter,
+  as the dependency chain could lead to an infinite loop. Mutating an adjacent
+  property from a getter is not a good practice anyway, even with a caching
+  mechanism reducing reruns.
```
mixonic, line 90: This paragraph confuses me a bit. The example above is probably implied to be inside a `@cached` getter based on context, but this paragraph when warns about using it with `@tracked`.

Furthermore, it isn't clear why the code snippet above implies an "adjacent" property. Maybe because the context of how a simple diffing example should be used is incomplete?

packages/@ember/-internals/metal/lib/cached.ts
```diff
@@ -4,7 +4,100 @@
...
+
+  The cost of these edge-guards adds up to the trade-off calculation of using
+  a caching strategy, hence requiring a very sensitive and moderate approach
+  regarding performance.
+
```
mixonic, line 95:
```suggestion
```

:scissors: perhaps? I think this was already make clear in the paragraph above (which I think warns against it three times in different ways). This sentence is a bit awkward in its wording, and I don't think it adds anything that will be missed.

packages/@ember/-internals/metal/lib/cached.ts
NullVoxPopuli: updated!

mixonic (approved):
Awesome. Thanks @NullVoxPopuli

PR #20547, opened 2023-09-27
On internal-strict-modules, get linting passing

NullVoxPopuli:
Minimally gets linting green on Ci.

Like the other PR,
C.I. is not expected to pass on `internal-strict-modules`.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -41,6 +41,7 @@ module.exports = {
   overrides: [
     {
       files: ['*.ts'],
+      excludedFiles: ['packages/ember/version.d.ts', 'vite.config.ts'],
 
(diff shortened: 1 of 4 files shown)


.eslintrc.js
```diff
@@ -41,6 +41,7 @@ module.exports = {
   overrides: [
     {
       files: ['*.ts'],
+      excludedFiles: ['packages/ember/version.d.ts', 'vite.config.ts'],
```
NullVoxPopuli, line 44: type-aware lints are included in here, but these file have no associated tsconfig.json, and thus cannot use type-aware lints.

kategengler:
Is this still relevant?

NullVoxPopuli:
nope!

PR #20557, opened 2023-10-14
Upgrade to ember-cli-babel 8.2.0

NullVoxPopuli:
ember-cli-babel 8 includes nice things for supporting `content-tag`.

Ideally, however, once ember-source is a v2 addon, we can just remove ember-cli-babel.

The breaking change from ember-cli-babel 7 to 8: https://github.com/emberjs/ember-cli-babel/releases/tag/v8.0.0

is that the consuming project must have declared `@babel/core`, which is already in the blueprint: https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/package.json#L29
(but was not added until ember-cli 5.3).

So, does this mean that ember-source can't upgrade ember-cli-babel?
and _instead_ should convert to v2 addon? (I'm thinking it does).

However, [this needs finished](https://github.com/emberjs/ember.js/compare/main...internal-strict-modules), but I don't know [what's going on anymore](https://discord.com/channels/480462759797063690/1156567765780537415/1159822495944298516).

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,6 @@
     "build": "npm-run-all build:*",
     "docs": "ember ember-cli-yuidoc",
     "start": "ember serve",
-    "link:glimmer-vm": "node bin/link-glimmer-vm-packages.mjs",
     "lint": "npm-run-all --continue-on-error --aggregate-output --parallel \"lint:!(fix)\"",
     "lint:docs": "qunit tests/docs/coverage-test.js",
     "lint:eslint": "eslint --report-unused-disable-directives --cache .",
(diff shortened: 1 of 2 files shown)


anandfresh (approved):

NullVoxPopuli:
oh probably, I should rebase

kategengler:
and convert from draft, if you want it merged :)

package.json
NullVoxPopuli, line 1: looks like lints don't enforce sort order here -- leaving tooling to mess things up.....

NullVoxPopuli:
Upgrade also happened over here: https://github.com/emberjs/ember.js/pull/20602/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L90

chancancode:
Yeah, I am not sure what is the answer to your questions (basically: would it have been a breaking change if we required a new peer dependency?). I did write in my commit message that, if we couldn’t upgrade easily, I only needed the static block plugin so I could also configure that directly

chancancode:
But if we are going to do that, feel free to land it here and I’ll rebase when it’s available. I don’t necessarily think that other branch is intended for direct merging anyway

PR #20559, opened 2023-10-17
Switch to pnpm

ef4:
This is the first step in a refactor toward using strict ES modules throughout our own internals.

pnpm ensures that all our packages are only using their declared dependencies.

I updated the existing broccoli-based tooling here to not rely on sloppy transitive dep leakage, even though I want to land an upcoming PR that would replace that machinery entirely. I think it's easier to land this in stages.


.github/actions/setup/action.yml
```diff
@@ -0,0 +1,48 @@
...
+      with:
+        version: 8
+        run_install: false
+
+    - name: Get pnpm store directory
+      shell: bash
+      run: |
+        echo "STORE_PATH=$(pnpm store path --silent)" >> $GITHUB_ENV
+
+    - uses: actions/cache@v3
```
kategengler, line 27: `actions/setup-node` supports a `pnpm` as a key for their `cache` option. If you are aware of this and this is still needed, 👍 but just wanted to check

.github/workflows/alpha-releases.yml
```diff
@@ -10,35 +10,24 @@ jobs:
...
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/setup-node@v3
-        with:
-          node-version: 16.x
-          cache: yarn
-          registry-url: 'https://registry.npmjs.org'
-      - name: install dependencies
-        run: yarn install --frozen-lockfile --non-interactive
+      - uses: ./.github/actions/setup
```
kategengler, line 30: The registry-url needs to be passed to setup-node here for pushing to npm to work

.github/workflows/ci.yml
```diff
@@ -290,13 +229,7 @@ jobs:
...
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/setup-node@v3
-        with:
-          node-version: 16.x
-          registry-url: 'https://registry.npmjs.org'
-          cache: yarn
-      - name: install dependencies
-        run: yarn install --frozen-lockfile --non-interactive
+      - uses: ./.github/actions/setup
```
kategengler, line 232: Again the registry-url is needed for deploying

kategengler (approved):

PR #20561, opened 2023-10-24
Upgrade VM `@glimmer/*` packages from 0.84.3 to 0.85.13

NullVoxPopuli:
This is post-the-major-refactor-that-yehuda-did.

Huge thanks to @wycats and @chancancode for helping me through this upgrade process :tada: 

## GlimmerVM changes

- some cruft / techdebt removed (a few polyfills, etc)
- no more enums (which have a runtime cost)
- better handling of locals detection:
  - https://github.com/glimmerjs/glimmer-vm/pull/1421
- Various breaking API changes (things got re-organized a bit) 
  
## Changes

- `@glimmer/interfaces` no longer provides the `Option` type. This has been internalized to `@ember/-internals/utility-types` (renamed to Nullable to match, tho)
- `@glimmer/util` no longer provides a `WeakSet` polyfill, so ember now uses native `WeakSet` -- which has been supported by everyone for nearly all of time: https://caniuse.com/?search=WeakSet
- ComponentDefinition's capabilities property is now of type CapabilityMask, instead of InternalComponentCapability
- @glimmer/validator's setTrackingTransactionEnv is now only accessible via the 'debug' export
- Fix some Reference<unknown> situations, some types to Reference<more specific>, and some pushing casting down to the last possible place
- CurriedType is now in @glimmer/vm, because it's a value, not just a type/interface
- `programCompilationContext` now takes a third argument
- Remove https://babeljs.io/docs/babel-plugin-transform-block-scoping 
  - this is included in `@babel/preset-env` (so its used if needed, as determined by targets, which our targets don't determine that we need it)

<details><summary>blockers that are no longer</summary>

- [x] bring back `@glimmer/env`'s DEBUG: https://github.com/glimmerjs/glimmer-vm/pull/1472

- [x] `@glimmer/*` needs to move to conditional exports (instead of pushing import.meta.env on consumers)

</details>

<details><summary>not good ideas (or not good right now), now abandoned</summary>

- [ ] `@glimmer/compiler` depends on `@glimmer/local-debug-flags`, PR: https://github.com/glimmerjs/glimmer-vm/pull/1461
  ```
  node_modules/.pnpm/@glimmer+compiler@0.85.3/node_modules/@glimmer/compiler/lib/compiler.ts:7:34
  error TS2307: Cannot find module '@glimmer/local-debug-flags' or its corresponding type declarations.
  
  7 import { LOCAL_SHOULD_LOG } from '@glimmer/local-debug-flags';
  ```
  this is a devDependency that should be compiled away during publish.

- [ ] ember-source needs to support import.meta.env, PR: https://github.com/emberjs/ember-cli-babel/pull/506
- [ ] consuming projects need to support import.meta.env (we want to pass this through), PR: https://github.com/emberjs/ember-cli-babel/pull/506

</details>


testem.browserstack.js
```diff
@@ -58,7 +58,7 @@ const BrowserStackLaunchers = {
 };
 
 module.exports = {
-  test_page: 'dist/tests/index.html?hidepassed&hideskipped&timeout=60000',
+  test_page: 'dist/tests/index.html?hidepassed&hideskipped&timeout=60000&disable_local_debug',
```
NullVoxPopuli, line 61: All these testem changes are using a new flag form glimmer-vm which revealed a real (existing) bug that we need to fix.

Some info here: https://github.com/emberjs/ember.js/issues/20573

chancancode: Is the "local debug" code still shipped with ember-source or do they get stripped out somewhere?

packages/@ember/-internals/glimmer/tests/integration/application/debug-render-tree-test.ts
```diff
@@ -99,7 +99,7 @@ if (ENV._DEBUG_RENDER_TREE) {
           this.outlet({
             type: 'route-template',
             name: 'index',
-            args: { positional: [], named: { model: undefined } },
+            args: { positional: [], named: { model: undefined }, errors: {} },
```
chancancode, line 102: We should probably change how this works: https://github.com/glimmerjs/glimmer-vm/commit/626b0e7caec9e6bff60752be71996e783eaeb55b#r132314108

packages/@ember/-internals/glimmer/lib/syntax/outlet.ts
```diff
@@ -110,7 +110,12 @@ export const outletHelper = internalHelper(
   }
 );
 
-function stateFor(ref: Reference, outlet: OutletState | undefined): OutletDefinitionState | null {
+function stateFor(
+  // SAFETY: default type here is Reference<unknown>, which isn't assignable to the
+  //         type in OutletDefinitionState which is Reference<Outlet | undefiend>
+  ref: Reference<any>,
```
chancancode, line 116: This is the wrong fix, consider the input to this function:

```ts
let outletRef = createComputeRef(() => {
  let state = valueForRef(scope.get('outletState') as Reference<OutletState | undefined>);
  return state?.outlets?.main;
});
```

`outletRef` did and should still have type `Reference<OutletState | undefined>`, which appears to be the case since the second argument `outlet` is derived from `valueForRef(ref)`. So the proper fix is to type this argument as `Reference<OutletState | undefined>` and, if that didn't work, trace what broke.

testem.browserstack.js
NullVoxPopuli: It's stripped out with `DEBUG` from `@glimmer/env`

lib/index.js
```diff
@@ -8,7 +8,7 @@ const buildDebugMacroPlugin = require('./build-debug-macro-plugin');
 const buildStripClassCallcheckPlugin = require('./build-strip-class-callcheck-plugin');
 const injectBabelHelpers = require('./transforms/inject-babel-helpers').injectBabelHelpers;
 const debugTree = require('broccoli-debug').buildDebugCallback('ember-source:addon');
-const vmBabelPlugins = require('@glimmer/vm-babel-plugins');
+const { default: vmBabelPlugins } = require('@glimmer/vm-babel-plugins');
```
chancancode, line 11:
```suggestion
const vmBabelPlugins = require('@glimmer/vm-babel-plugins').default;
```

Curious how come this was the only one that needs changing?

NullVoxPopuli: everything else uses actual ESM.
The whole build sytsem got swapped out in https://github.com/glimmerjs/glimmer-vm/pull/1427 or https://github.com/glimmerjs/glimmer-vm/pull/1426

vm-babel-plugins is compiled to have:
```js
'use strict';

Object.defineProperty(exports, '__esModule', { value: true });

var node_module = require('node:module');

const require$1 = node_module.createRequire((typeof document === 'undefined' ? require('u' + 'rl').pathToFileURL(__filename).href : (document.currentScript && document.currentScript.src || new URL('index.cjs', document.baseURI).href)));
function defaultTo(value, defaultVal) {
  return value === undefined ? defaultVal : value;
}


// ...
 
exports.default = generateVmPlugins;
//# sourceMappingURL=index.cjs.map

```

https://www.npmjs.com/package/@glimmer/vm-babel-plugins?activeTab=code

packages/@ember/-internals/glimmer/lib/helpers/action.ts
```diff
@@ -298,16 +298,23 @@ export default internalHelper((args: CapturedArguments): Reference<Function> =>
...
       valueForRef(context) as object,
-      target!,
-      action,
+      // SAFETY: glimmer-vm should expose narrowing utilities for references
+      //         as is, `target` is still `Reference<unknown>`.
+      //         however, we never even tried to narrow `target`, so this is potentially risky code.
+      target! as Reference<MaybeActionHandler>,
+      // SAFETY: glimmer-vm should expose narrowing utilities for references
+      //         as is, `action` is still `Reference<unknown>`
+      action as Reference<string | AnyFn>,
```
chancancode, line 307: I think this is just exposing a real bug, we really have no business making these unfounded claims here. We should keep them as `Reference<unknown>` and push it onto the consumption side to actually narrow the type with debug assertions when they have actual values.

packages/@ember/-internals/glimmer/lib/helpers/unique-id.ts
```diff
@@ -30,7 +30,12 @@ import { createConstRef } from '@glimmer/reference';
 import { internalHelper } from './internal-helper';
 
 export default internalHelper((): Reference<string> => {
-  return createConstRef(uniqueId(), 'unique-id');
+  // SAFETY: glimmer-vm should change the signature of createUnboundRef to use a generic
+  //         so that the type param to `Reference<?>` can infer from the first argument.
+  //
+  //         However, since ember-source@4.5, we can use functions explicitly,
+  //         so we should investigate if we even need this "internalHelper" wrapper.
```
chancancode, line 37: It's more the other way around – making const references is the optimization, using the user-land plain function helpers facilities would be creating extra wrappers, tracking frames, etc

packages/@ember/-internals/glimmer/lib/component-managers/curly.ts
```diff
@@ -6,6 +6,7 @@ import {
 } from '@ember/-internals/owner';
 import { enumerableSymbol, guidFor } from '@ember/-internals/utils';
 import { addChildView, setElementView, setViewElement } from '@ember/-internals/views';
+import type { Option } from '@ember/-internals/utility-types';
```
chancancode, line 9: The type was just renamed `Nullable` upstream it seems? We should use that rather than duplicating it here

packages/@ember/-internals/glimmer/lib/syntax/outlet.ts
NullVoxPopuli: idk why this was so complicated initially -- I guess I fixed it later? idk. resolved

testem.browserstack.js
NullVoxPopuli: _HOWEVER_, DEBUG is passed on to consumers of ember-source, so I think glimmer-vm needs to compile this away

lib/index.js
NullVoxPopuli: Resolving, as this is a correct change

packages/@ember/-internals/glimmer/lib/helpers/unique-id.ts
NullVoxPopuli: Resolved, thanks for explaining!

packages/@ember/-internals/glimmer/lib/helpers/action.ts
NullVoxPopuli: well, isn't `isInvokableRef(action)` supposed to narrow to `string | AnyFn`?

NullVoxPopuli: Getting rid of the casts here causes:
image

which... feels worse?

In any case, if this is a real bug, I think it should be handled in a separate PR as it's been around for a long while, and only exposed through glimmer-vm's new stricter types.

packages/@ember/-internals/glimmer/lib/component-managers/curly.ts
NullVoxPopuli: Nullable is not exported.

This seems like a too-simple type to let another project keep maintenance of.

When we combine ember-source and glimmer-vm, we can unify these concepts in a `*/utility-types` package or something.

packages/@ember/-internals/glimmer/lib/helpers/action.ts
chancancode: In general, we don't know what is _in_ a reference without dereferencing it, just like there is no correct way to implement `isFunctionOfString(func: () => unknown): func is () => string` without actually calling it, unless we are sure that we were the one that made the reference in the first place and specifically know what is in it, which doesn't seem to be the case here.

What I am saying is that, since we don't know what is in it, we should keep it as `Reference<unknown>` on this line, and 100% the intention is specifically to _cause_ that TypeError you found. This identifies the place where we are currently making an incorrect claim.

You then change those to `Reference<unknown>` as well, which pushes the type error further down, to the line where we currently, incorrectly, does `let something: TButProbablyInferred = valueForRef(ref);`. Now here is the real issue, but this is where the place we can actually do something about it, since, we now have the actual dereferenced value:

```ts
let something = valueForRef(ref); // now something is correctly `unknown`
assert("something must be a string!", typeof something === "string");
// now something is `string`
```

packages/@ember/-internals/glimmer/lib/component-managers/curly.ts
chancancode: It probably should be exported 🤷🏼‍♂️ it is simple, yes, but if it is (and probably it is) used in the signatures of public functions, etc, then it should be exported publicly too. But anyway, for now it is fine to have a copy here, but we should probably at least update to match its upstream name to avoid confusion.

packages/@ember/-internals/glimmer/lib/helpers/action.ts
NullVoxPopuli: Fixed in 510d28c01

packages/@ember/-internals/glimmer/lib/component-managers/curly.ts
NullVoxPopuli: done!

packages/@ember/-internals/glimmer/lib/helpers/action.ts
```diff
@@ -347,29 +351,37 @@ function makeArgsProcessor(valuePathRef: Reference | false, actionArgsRef: Refer
...
   context: object,
   targetRef: Reference<MaybeActionHandler>,
-  actionRef: Reference<string | AnyFn>,
+  actionRef: Reference<unknown>,
   processArgs: (args: unknown[]) => unknown[],
   debugKey: string
 ) {
+  let action = valueForRef(actionRef);
+
+  isClosureAction(action);
```
chancancode, line 371: Do the assert inline: `assert` is special and gets stripped in prod build. Extracting it like this will leave behind an empty function. In this case you don't gain anything by extracting it, but if you ever want/need to do it for abstraction/reuse reasons then you have to do some careful `if (DEBUG)` dance

NullVoxPopuli: hm -- this fights TS -- because we cant have type assertions inline
See: https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgQwM6oKazgXzgMyghDgHIABDEAIywHoATDagVwHNSBuAKG4wA9I2GAE8wGOAEEAdiIBi0uAF44ACgB0m5FDaoAXClkBtALoBKZQD5DIntwA2GePwMtpAa2kQA7tLtpMWFVSEFQOABo4UXEIfDh+ZSUVUnw3AGMYYAhpUjMDAKwYVHi4YGKZeT9efAgIVX4zO1TpDKzFGrrtNgMKhQsEXhwgA

but also, wouldn't empty functions get optimized away by a minifier, same as true/false constants?

> careful if (DEBUG) dance

this would negate the type guard :sweat_smile:

chancancode: Ah yes I see the disconnect. Generally, this works, and was what I meant:

https://www.typescriptlang.org/play?ssl=20&ssc=2&pln=11&pc=1#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXygGdCQYMAKAWxGKgHMQAueQjGLVOgGnjD2CzY8zNAGtUOAO6oAlMyIkyhXv0G5UAbgBQW0JFgIOGUoihgEAJRCJSIdCAA8AFQB88AN5b48AG5QIyEzwTloAvjp60HBIaJjqvv6BAGI4MFaIzjwW8CAAHsaowMrptvbOLi7kcIjMFnLB2rrgUQgo6EL4lFCiIADCEDiEyHAAgnF45GYdzKzsnPAAPvBJsR31PjhYwI1t453dIAAiAJ6oUJRYYP2DwyBjHZN76bXWpeYOYhLSLuubwB5eFSoVgEPbwAC8CQCIBSaWsjw66RkjW8ClIFEB3gARFN4pRkCCAEYIKAsNgcOjwVIEGLtdRYriY+AYY4ABxAOEQoI6EPBkKxswpWMWSxZ7M53PifP5uw6WMByJ03i6PWuQ1GewR6kVoSAA

The issue is that `typeof x === 'function'` only narrows to `Function` and here it is claiming that we expect it to be an `AnyFn`, which is `(...args: any[]) => any`. This distinction is kind of an (IMO) in-the-weeds TypeScript quirk, and I think in this case, in this particular file, we don't really lose any TypeSafety by swapping out `AnyFn` with `Function`, arguably it's "semantically more strict/correct" since we have no idea what kind of function is in there.

In fact, if you trace it all the way to how this is used, everything bottoms out (or tops out, I guess?) on L294/295 where it actually gets assigned as a `Function` anyway.

So:


https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXygGdCQYMAKMPYLbPALnjQGtUcB3VASkdRADdSAbgBQoSLAQp0dfERJlyAWxDEoAcxCNCGGFlTqANPCqoasxizaceBYqQyET1WrlSix4aHHj6MpRCgwBAAlEERSEHQQAB4AFQA+eABvEXh4figIZC14OJEAXxFPCR9pTDcMrJyAMRwYMMR44xD4EAAPfzMnRsjo+ISE8jhERhDbOI8Aein4AGUsJQAHCCxELBBgEu8pNAq8X1R-GFQsgAkQCCXSAfJEVEZyWHUx8L7gmKsOVASueABeJK9OD9RK2fg4LDADwdJb1DDwUCBZAQBF+UinCAXK6kchPGAveDAqIfL6cX6vCIgj41PayJKAlJpeAQEAI+6MWkyNwedIzeAAOiFxWZ9wB8CUUGYIAAIgBPU5KLBgADCEBwhGQcAAgvtUPj1FwPMz+UKBcV0vIHOQAESS5j6dQEeAjd5SGA4JRIfBsBGEAAW30O8Aw-qwTlAShwNqNhVjxXKsglUtlCqgStV6s1Or1Tz1jUpbs+qFY3wp8C5eqZ6VMOgIVf+VWyIDqDXCedkjXjlvsimZ6RtQSTSmQdYARggoPAdHoDPB6s7E24bYZ+yG5dccIh60n-nv4DaZ46bfAAD6n9eb7dDyp7xs2pd4G3M7sutla-D2kBqjVakC62QOzcWMihER9PxTH9s3-XMbwYaddEdM8KzpYDOVQg5UnSNZ4HIDANxALcd1vfdD0QgwY2rdJ4FNYVqLsBQKDI2cnTg1AnF9adA04YNQ3DREQCjGNRHSAo2ggEgqPSOAMA-Yi8BE+AigKIA

chancancode: And, for reference, the annoying debug dance looks something like this:

https://www.typescriptlang.org/play?noUncheckedIndexedAccess=true&noUnusedLocals=true&noUnusedParameters=true&useUnknownInCatchVariables=true&exactOptionalPropertyTypes=true&noFallthroughCasesInSwitch=true&noImplicitOverride=true&noPropertyAccessFromIndexSignature=true#code/CYUwxgNghgTiAEYD2A7AzgF3gEQKICEBVAcQC54AjJJCEKFAbgChRJYEMBPABwQEEUnAGIp4AXngAKAHSzYAczTl6nANoBdAJTiAfPBXNW0OPABmAVxRgMAS1T60aEDAySAtiEdR5Icphg2KPIANIiowDa2qOSWANYoSADuKJrKjs4YaGEoEVGMTAW0WDZoAsIo5JJQ1nYV8HEJydpietV58CXwZSLMTDamUnhExNoA3kzwHaWCIuJSbbUxKPFJKco19p3doi3wXLxIAwv2YqfwAEQWVnnnzAC+BVcbom5QsSAAwhBIaOZwfM8qs8lismvBxpMAPSQ+AAST2AAskOZ5AisJxkQByCAQeAoEAgYD6eAAQngCOcCEkJW2JKBeU02go4Cg5ic8AAKgBlRD0TFYJwcBFQLAAAxpMxQoqm8FApkCkRAEE4DjQNnk+OAE3g0PgQiQMHgaCQHngcCgxtEkVlSE8KH5YTc3GggXJlI6ogwFPgzqgnHkMGROVC3EDFCgFGV8AAVmysBFgPasMKAG5ChD+GxgI0gDC2IJZAKojDa3VlRJ+0L9eAY8w2vEEolQUmhRIIUGIkUytBQUwIZlgVnsi1OFzwVBRmCWLLVoYkGUYKcgbUjjKSc7HF5xygIZtPG5V6blem1Rn3JhAA

chancancode: > but also, wouldn't empty functions get optimized away by a minifier, same as true/false constants?

I'm not sure! Does it work in terser?

chancancode: But:

1. It doesn't matter in this case anyway if we switch to `Function`
2. I think even if it does, we probably don't do it (and instead do those annoying `if (DEBUG)` dance) because we don't want to rely on the function happen to be empty; it's too easy for someone to accidentally see the function lying around and add more lines during a refactor, and it's not stripped anymore, where as the other more annoying pattern makes it clear that you shouldn't do it

NullVoxPopuli: > I'm not sure! Does it work in terser?

Yeah, Starbeam relies heavily on optimizing away empty functions, constructors, static blocks, etc

NullVoxPopuli: Changing to `Function` all the way through affects the public API of `@ember/runloop`, so I think I'll cast on line 424.

View here: 10db7f68a

testem.browserstack.js
NullVoxPopuli: done!

Hopefully we can return the opt-in capability once we add import.meta.env everywhere, hopefully

packages/@ember/-internals/glimmer/tests/integration/application/debug-render-tree-test.ts
NullVoxPopuli: as an update, we reverted the change in glimmer-vm to later be re-implemented post-error-recovery PR

packages/@ember/-internals/glimmer/lib/helpers/action.ts
```diff
@@ -417,7 +417,7 @@ function makeClosureAction(
   return (...args: any[]) => {
     let payload = { target: self, args, label: '@glimmer/closure-action' };
     return flaggedInstrument('interaction.ember-action', payload, () => {
-      return join(self, fn, ...processArgs(args));
+      return join(self, fn as AnyFn, ...processArgs(args));
```
chancancode, line 420: Yeah I think it's valid to push the cast down to here 👍🏼

packages/@ember/-internals/glimmer/lib/helpers/action.ts
```diff
@@ -299,8 +298,13 @@ export default internalHelper((args: CapturedArguments): Reference<Function> =>
...
       valueForRef(context) as object,
-      target!,
-      action,
+      // SAFETY: glimmer-vm should expose narrowing utilities for references
+      //         as is, `target` is still `Reference<unknown>`.
+      //         however, we never even tried to narrow `target`, so this is potentially risky code.
+      target! as Reference<MaybeActionHandler>,
+      // SAFETY: glimmer-vm should expose narrowing utilities for references
+      //         as is, `action` is still `Reference<unknown>`
+      action as Reference<unknown>,
```
chancancode, line 307: did some changes got lost with a rebase or something?

packages/@ember/-internals/glimmer/lib/helpers/action.ts
NullVoxPopuli: <3

packages/@ember/-internals/glimmer/lib/helpers/action.ts
NullVoxPopuli: yeah.... :(

packages/@ember/-internals/glimmer/lib/helpers/action.ts
```diff
@@ -349,55 +353,51 @@ function makeArgsProcessor(valuePathRef: Reference | false, actionArgsRef: Refer
...
 
   assert(
     `Action passed is null or undefined in (action) from ${target}.`,
     action !== undefined && action !== null
   );
 
   if (typeof action === 'string') {
-    self = target;
-    fn = (target.actions && target.actions[action as string])!;
+    self = target as object;
```
chancancode, line 399:
```suggestion
    assert('target must be an object', target !== null && typeof target === 'object');
    self = target;
```

packages/@ember/-internals/glimmer/lib/helpers/action.ts
```diff
@@ -349,55 +353,51 @@ function makeArgsProcessor(valuePathRef: Reference | false, actionArgsRef: Refer
...
+    fn = (target as { actions: Record<string, Function> })?.actions?.[action as string] as Function;
```
chancancode, line 400:
```suggestion
    const value = (target as { actions?: Record<string, unknown> }).actions?.[action];
    assert(`An action named '${action}' was not found in ${target}`, Boolean(unchecked));
    assert(`An action named '${action}' was found in ${target}, but is not a function`, typeof unchecked === 'function');
    fn = value;
```

`typeof action` should already be string yeah? The guard on 398 should have narrowed it. Some unnecessary question mark/unchecked casts here.

chancancode (approved):
Other than fixing that final unchecked type cast I think we are good, can you apply those suggestions (verbatim or spirit of it) and squash when merging?

packages/@ember/-internals/glimmer/lib/helpers/action.ts
```diff
@@ -349,55 +353,58 @@ function makeArgsProcessor(valuePathRef: Reference | false, actionArgsRef: Refer
...
-    fn = (target.actions && target.actions[action as string])!;
+    let value = (target as { actions?: Record<string, unknown> }).actions?.[action];
+    assert(`An action named '${action}' was not found in ${target}`, Boolean(value));
+    assert(
+      `An action named '${action}' was found in ${target}, but is not a function`,
+      typeof value === 'function'
+    );
+    fn = value;
 
     assert(`An action named '${action}' was not found in ${target}`, Boolean(fn));
```
chancancode, line 409:
```suggestion
```

PR #20567, opened 2023-11-04
chore: remove AWS SDK for JavaScript v2

trivikr:
The only place where AWS SDK for JavaScript is used is in script to publish to S3, and it moved to v3 in https://github.com/emberjs/ember.js/pull/20455

Looks like the PR missed removing AWS SDK for JavaScript v3 which is no longer used.


kategengler:
Thank you!

PR #20571, opened 2023-11-09
set isTracked on getter

patricklx:
Instead of the descriptor. It's just ignored by defineProperty.

This would make it useful for the inspector to detect tracked property without hacks


chancancode:
Can you expose this as

chancancode:
Can you expose these APIs with functions instead? maybe `_hasOwnTrackedProperty` and `_hasTrackedProperty` with the latter walking the prototype chain. Put it in some internal module and make it as `@internal` until we can RFC it. Probably better to keep track of this with WeakMap of Sets instead of mutating the function.

patricklx:
I just saw that i can do meta(obj). peekDescriptor(prop) instanceof TrackedDescriptor

PR #20577, opened 2023-11-18
[BUGFIX canary] Restore/deprecate outlet `TemplateFactory` support

### LINKED ISSUE #20576: @ember/test-helpers rendering test broke on canary after #20570
Reported by @ijlee2 https://github.com/ijlee2/embroider-css-modules/actions/runs/6891050277

Background:

* `TemplateFactory` is `(owner) => Template`
* The `template` field on `OutletState` was always meant to be a `Template`
* `@ember/test-helpers` used the internal API incorrectly and passes a `TemplateFactory` in place of a `Template`
* #20570 removed these lines https://github.com/emberjs/ember.js/commit/f5572ac9c63e2d42a0b59d1fca25ad98eb170bb2#diff-45a89ad591d3fa8ed8cb55e9a963de378f51709d32e00053137ea4905a297c6fL123-L127
* The comments indicated that the incorrect usage was fixed in https://github.com/emberjs/ember-test-helpers/commit/11b4b93df18d4c372a85c56c55c1f03fa095b489
* Either that comment was inaccurate or the situation regressed, as they are now a few more spots that have this problem in the test helper rendering code that has this problem
* I am inclined to say we fix it on their side, but maybe an assertion here would be helpful

chancancode:
Fixes #20576

The test is a bit misleading in that, this is not actually restored to support the very old 1.6.0 version of ember-test-helpers. The issue is that while that particular spot of the invalid usage was fixed in 1.6.1, other similarly invalid patterns have popped up in the code since then. However, the old tests was probably a close enough approximation of the general problem.

This is not to avoid fixing the actual issue, but it's just that it was very difficult for us to support for a short time, and there may be other code that copied the same invalid pattern elsewhere.

We could probably backport this to all the way up to LTS if we want with minimal effort, since the code I brought back is basically the same as what was previously there, just with the deprecation.

See also https://github.com/emberjs/ember-test-helpers/pull/1445

### DIFF
--- a/packages/ember/tests/ember-test-helpers-test.js
+++ b/packages/ember/tests/ember-test-helpers-test.js
@@ -138,4 +138,124 @@ module('@ember/test-helpers emulation test', function () {
       });
     });
   });
+
+  module('v1.6.0', function () {
+    let EMPTY_TEMPLATE = compile('');
+
+    function settled() {
+      return new Promise(function (resolve) {
+        let watcher = setInterval(() => {
+          if (_getCurrentRunLoop() || _hasScheduledTimers()) {
+            return;
+          }
+
+          // Stop polling
+          clearInterval(watcher);
+
+          // Synchronously resolve the promise
+          run(null, resolve);
+        }, 10);
+      });
+    }
+
+    async function setupContext(context) {
+      // condensed version of https://github.com/emberjs/ember-test-helpers/blob/v1.6.0/addon-test-support/%40ember/test-helpers/build-owner.ts#L38
+      // without support for "custom resolver"
+      await context.application.boot();
+
+      context.owner = await context.application.buildInstance().boot();
+    }
+
+    function setupRenderingContext(context) {
+      let { owner } = context;
+      let OutletView = owner.factoryFor('view:-outlet');
+      let environment = owner.lookup('-environment:main');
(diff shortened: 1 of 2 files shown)


packages/ember/tests/ember-test-helpers-test.js
```diff
@@ -138,4 +138,124 @@ module('@ember/test-helpers emulation test', function () {
...
+
+    let templateId = 0;
+    function render(template, context) {
+      let { owner } = context;
+      let toplevelView = owner.lookup('-top-level-view:main');
+      templateId += 1;
+      let templateFullName = `template:-undertest-${templateId}`;
+      owner.register(templateFullName, template);
+
+      let outletState = {
```
NullVoxPopuli, line 201: is every property in here needed?

NullVoxPopuli (approved):

packages/ember/tests/ember-test-helpers-test.js
chancancode: Yeah, pretty much, the actual thing is typed in TypeScript in the runtime code (it's the interface you commented on last time)

chancancode: But also, the point was to replicate the same code in test helpers, which I gather at one point this was probably close to verbatim, and it's still not that far off today

PR #20579, opened 2023-11-27
Mark `cached` as a `MethodDecorator` instead of `PropertyDecorator`

boris-petrov:
I believe that's more correct as `cached` uses/needs `descriptor` which is only available in `MethodDecorator`.

cc @dfreeman, @chriskrycho, @jamescdavis, @wagenet

Sorry for pinging so many people but I'm not sure who I should. :smile: Is there like a TypeScript team that I can mention?

### DIFF
--- a/packages/@ember/-internals/metal/lib/cached.ts
+++ b/packages/@ember/-internals/metal/lib/cached.ts
@@ -93,7 +93,7 @@ import { createCache, getValue } from '@glimmer/validator';
   @for @glimmer/tracking
   @public
  */
-export const cached: PropertyDecorator = (...args: any[]) => {
+export const cached: MethodDecorator = (...args: any[]) => {
   const [target, key, descriptor] = args;
 
   // Error on `@cached()`, `@cached(...args)`, and `@cached propName = value;`


kategengler:
@emberjs/typescript-core ^

gitKrystan:
I've run into this also. Pinging the rest of the TS team to see if there's a reason it was typed as `PropertyDecorator` in the first place.

chriskrycho:
As you can see here at the implementation, it is… not well-typed, and I think the use of `PropertyDecorator` was basically just someone’s guess at the right thing at some point, which has stuck around. 😂 On digging back through [the relevant TS docs][docs], I think this should actually be an accessor decorator (see [here specifically][accessor]), and as far as I can tell the only corresponding type TS supplies is indeed `MethodDecorator`, so this seems like the right move. 👍🏼 

[docs]: https://www.typescriptlang.org/docs/handbook/decorators.html
[accessor]: https://www.typescriptlang.org/docs/handbook/decorators.html#accessor-decorators

kategengler:
Thank you!

PR #20580, opened 2023-11-28
provide debug utilities for inspector via event communication

patricklx:
this would allow inspector to communicate with emberjs without relying on any globals like the define, require functions and without the barrel file.
it would be required for inspector to provide more functionality when using vite or also when building with embroider webpack and staticEmberSources=true

would also like to https://github.com/glimmerjs/glimmer-vm/pull/1489
as part of this

https://github.com/emberjs/ember-inspector/pull/2554

### DIFF
--- a/packages/@ember/-internals/metal/lib/cached.ts
+++ b/packages/@ember/-internals/metal/lib/cached.ts
@@ -113,7 +113,10 @@ export const cached: MethodDecorator = (...args: any[]) => {
     throwCachedGetterOnlyError(key);
   }
 
-  CacheMap.set(target, [...(CacheMap.get(target) || []), key]);
+  if (!CacheMap.has(target)) {
+    CacheMap.set(target, new WeakSet());
+  }
+  CacheMap.get(target).add(key);
 
   const caches = new WeakMap();
   const getter = descriptor.get;
@@ -150,5 +153,5 @@ function throwCachedInvalidArgsError(args: unknown[] = []): never {
 }
 
 export function isCachedProperty(object: object, prop: string) {
-  return (CacheMap.get(object) || []).includes(prop);
+  return CacheMap.get(object)?.has(prop);
 }
--- a/packages/@ember/-internals/metal/lib/cached.ts
+++ b/packages/@ember/-internals/metal/lib/cached.ts
@@ -4,6 +4,8 @@
 import { DEBUG } from '@glimmer/env';
 import { createCache, getValue } from '@glimmer/validator';
 
+const CacheMap = new WeakMap();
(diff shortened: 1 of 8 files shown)


kategengler:
maybe @chancancode ?

NullVoxPopuli:
We now need this to proceed with https://github.com/emberjs/rfcs/pull/1015

packages/@ember/-internals/metal/lib/cached.ts
```diff
@@ -111,6 +113,8 @@ export const cached: MethodDecorator = (...args: any[]) => {
     throwCachedGetterOnlyError(key);
   }
 
+  CacheMap.set(target, [...(CacheMap.get(target) || []), key]);
```
NullVoxPopuli, line 116: should this be a WeakMap of WeakSets? rather than a WeakMap of Arrays?

packages/@ember/debug/ember-inspector.ts
NullVoxPopuli, line 1: I think this implementation is good for now, but we'll need to change things later, because I think we want ember-source to push panels into the inspector somehow.

needs some planning there, but it'd be great if ember-source's part of the inspector could be tested within ember-source, and than the inspector slurps those up (automatic updates depending on ember-source version!)

patricklx: I think pushing panels is not needed. more important would be to push a ember_debug directly from ember.js instead of having inspectors ember_debug try to support multiple versions.

packages/@ember/-internals/metal/lib/cached.ts
patricklx: changed to set instead of array. weakset does not work, since we are saving pure values (string)

packages/@ember/debug/ember-inspector.ts
NullVoxPopuli: well the thinking here is that because ember.js is has more people working on it, it could be updated more easily than the inspector -- and then the inspector also doesn't need to maintain compatibility with everything going forward -- as it would be provided by ember-source 

we should probs move to talking about this on the discord though -- lots of details to work through

ef4:
Are these things not already accessible via their existing private import paths?

patricklx:
Not with ember static source enabled. Then it's not accessible.

patricklx:
Biggest issue us the boot load event.

patricklx:
E.g. https://github.com/emberjs/ember-inspector/pull/2051

As an example

kategengler:
@mansona Is this still relevant after the inspector changes?

PR #20582, opened 2023-11-30
Reducing internal usage of AMD loader

ef4:
(This PR is re-opened from https://github.com/emberjs/ember.js/pull/20581. I merged that one prematurely and immediately force-pushed to revert it.)

There are several places where ember-source depends unnecessarily on details of the AMD loader:

 optionally loading the ember-testing implementation
 optionally loading the runtime template compiler
This PR replaces them with more direct forms of coordination via module-scoped state.

My overarching goal here is to eliminate all usage of require inside the actual ES module packages so they are true ES modules.

In contrast, the traditional ember.js, ember-testing.js, and ember-template-compiler.js bundles will still use require as their implementation to maintain backward compatibility. Where we need to patch over the weirdness of this (from the perspective of actual compliant ES modules), we do so entirely within the build that produces the bundles.

This is a step toward the proposed strict-es-modules optional feature. This implementation cleanup is happening before advancing that RFC because this cleanup is good either way and does not make any public API changes.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -55,7 +55,7 @@ module.exports = {
         '@typescript-eslint/ban-types': 'off',
         '@typescript-eslint/no-empty-function': 'off',
         '@typescript-eslint/no-this-alias': 'off',
-        '@typescript-eslint/no-var-requires': 'warn',
+        '@typescript-eslint/no-var-requires': 'error',
         '@typescript-eslint/consistent-type-imports': 'error',
 
         // TODO: Enable and fix these rules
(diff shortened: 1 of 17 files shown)


ef4:
@kategengler do you have an opinion about how to tag this commit? It's internal refactor.

kategengler:
If it is an internal refactor, would you even want it mentioned in the changelog?

kategengler:
Yeah we wouldn't typically backport an internal refactor

ef4:
This is ready to go.

ef4:
(This already got code reviewed in #20581)

PR #20587, opened 2023-12-01
Reform internal use of template compiler

ef4:
Previously, ember-source had a weird custom way of compiling templates that appear within its own source. This PR makes it work the same way we do everywhere else, using babel-plugin-ember-template-compilation.

This required solving a boostrapping problem: the babel plugin needs ember-template-compiler.js, our build produces
ember-template-compiler.js, but our build wants to use the babel plugin. I solved it by making the template compiler directly evaluatable in node, with no build at all.

The *full* legacy ember-template-compiler API has a lot of silly things shoved into it (like a reexport of all of the `Ember` metapackage), and that was inextricably circular, so I introduced `ember-template-compiler/minimal` for this internal use case.

### DIFF
--- a/broccoli/packages.js
+++ b/broccoli/packages.js
@@ -14,7 +14,6 @@ const { VERSION } = require('./version');
 const PackageJSONWriter = require('./package-json-writer');
 const WriteFile = require('broccoli-file-creator');
 const StringReplace = require('broccoli-string-replace');
-const GlimmerTemplatePrecompiler = require('./glimmer-template-compiler');
 const VERSION_PLACEHOLDER = /VERSION_STRING_PLACEHOLDER/g;
 const canaryFeatures = require('./canary-features');
 
@@ -54,32 +53,21 @@ module.exports.qunit = function _qunit() {
 
 module.exports.getPackagesES = function getPackagesES() {
   let input = new Funnel(`packages`, {
-    exclude: ['loader/**', 'external-helpers/**'],
+    exclude: ['loader/**', 'external-helpers/**', '**/node_modules'],
     destDir: `packages`,
   });
 
   let debuggedInput = debugTree(input, `get-packages-es:input`);
 
-  let compiledTemplatesAndTypescript = new GlimmerTemplatePrecompiler(debuggedInput, {
-    persist: true,
-    glimmer: require('@glimmer/compiler'),
-    annotation: `get-packages-es templates -> es`,
-  });
-
-  let debuggedCompiledTemplatesAndTypeScript = debugTree(
-    compiledTemplatesAndTypescript,
-    `get-packages-es:templates-output`
-  );
-
   let nonTypeScriptContents = debugTree(
-    new Funnel(debuggedCompiledTemplatesAndTypeScript, {
+    new Funnel(debuggedInput, {
       srcDir: 'packages',
       exclude: ['**/*.ts'],
     }),
     'get-packages-es:js:output'
   );
 
-  let typescriptContents = new Funnel(debuggedCompiledTemplatesAndTypeScript, {
+  let typescriptContents = new Funnel(debuggedInput, {
     include: ['**/*.ts'],
   });
 
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -65,16 +65,21 @@ module.exports = function ({ project }) {
   let emberBundles = withTargets(project, emberSource.buildEmberBundles.bind(emberSource));
 
   let packages = debugTree(
-    new MergeTrees([
-      // dynamically generated packages
-      emberVersionES(),
+    new MergeTrees(
+      [
+        // packages/** (after typescript compilation)
+        getPackagesES(),
 
-      // packages/** (after typescript compilation)
-      getPackagesES(),
+        emberVersionES(),
 
-      // externalized helpers
-      babelHelpers(),
-    ]),
+        // externalized helpers
+        babelHelpers(),
+      ],
+      {
+        // we're replacing the ember/verion file with the actual version number
+        overwrite: true,
+      }
+    ),
     'packages:initial'
   );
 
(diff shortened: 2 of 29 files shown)


broccoli/packages.js
```diff
@@ -54,32 +53,21 @@ module.exports.qunit = function _qunit() {
 
 module.exports.getPackagesES = function getPackagesES() {
   let input = new Funnel(`packages`, {
-    exclude: ['loader/**', 'external-helpers/**'],
+    exclude: ['loader/**', 'external-helpers/**', '**/node_modules'],
```
ef4, line 56: This PR makes all the packages real pnpm workspaces, which was necessary to make the ember-template-compiler package actually evaluatable in node. Previously not all packages were. This extra filtering is necessary now that pnpm is creating node_modules directories inside the packages.

ember-cli-build.js
```diff
@@ -65,16 +65,21 @@ module.exports = function ({ project }) {
...
 
-      // externalized helpers
-      babelHelpers(),
-    ]),
+        // externalized helpers
+        babelHelpers(),
+      ],
+      {
+        // we're replacing the ember/verion file with the actual version number
+        overwrite: true,
```
ef4, line 80: The template compiler imports the version. Previously that would die because the version file is dynamically created. Now we have a placeholder file on disk that gets replaced in the build, so that our internal usage of template compiler doesn't die.

lib/index.js
```diff
@@ -145,6 +145,17 @@ module.exports = {
...
 
+    let compilerPath;
+    if (isEmberSource) {
+      // Here we are using the template compiler by directly evaluating it in
+      // node, because we *are* the build that produces
+      // ember-template-compiler.js.
+      compilerPath = path.resolve(__dirname, '../broccoli/glimmer-template-compiler');
+    } else {
+      // When we're building an app, we use the already built template compiler.
+      compilerPath = path.resolve(__dirname, '../dist/ember-template-compiler.js');
```
ef4, line 156: This keeps the exact old behavior when building apps, and only changes the behavior when ember is building itself.

package.json
```diff
@@ -111,6 +111,8 @@
     "@embroider/shared-internals": "^2.5.0",
     "@rollup/plugin-babel": "^6.0.3",
     "@simple-dom/document": "^1.4.0",
+    "@swc/core": "^1.3.100",
+    "@swc-node/register": "^1.6.8",
```
ef4, line 115: This is how we can directly evaluate the template compiler without a build.

packages/@ember/-internals/glimmer/lib/templates/empty.ts
```diff
@@ -0,0 +1,4 @@
+import { precompileTemplate } from '@ember/template-compilation';
+export default precompileTemplate('', {
+  moduleName: 'packages/@ember/-internals/glimmer/lib/templates/empty.hbs',
+});
```
ef4, line 4: I was able to drop all special support for hbs files, and several hacks around their types, by making them just be the equivalent javascript.

packages/ember/version.ts
```diff
@@ -0,0 +1,2 @@
+// this file gets replaced with the real value during the build
+export default 'VERSION_GOES_HERE';
```
ef4, line 2: By having a real file on disk, we get to eliminate hacks around the types and we avoid crashing if you try to directly evaluate in node, without a build.

pnpm-workspace.yaml
```diff
@@ -0,0 +1,4 @@
+packages:
+ - 'packages/*'
+ - 'packages/@ember/*'
+ - 'packages/@glimmer/*'
```
ef4, line 4: When I converted this repo to pnpm, it looks like pnpm was auto-detecting all our packages. But it turns out it was not detecting all of them. This makes it explicit.

smoke-tests/ember-test-app/pnpm-workspace.yaml
```diff
@@ -0,0 +1,2 @@
+# this tells pnpm we are a separate project with a separate install, not part of
+# the monorepo's workspaces.
```
ef4, line 2: Adding pnpm-workspace.yaml at the root required also adding this, in order to keep the old behavior where this subdir gets an entirely independent `pnpm install`.

Arguably it would be better to just make this subdir another workspace, but that turns out to be non-trivial given NPM dependency hell, so I didn't try to attempt that change in this PR.

types/publish.mjs
```diff
@@ -98,20 +98,6 @@ ${MODULES_PLACEHOLDER}
...
-// `*-ext.d.ts` modules) these still need the module wrapper added.
-const HAND_COPIED_BUT_NEEDS_POSTPROCESSING = [
-  'ember/version.d.ts',
-  '@ember/-internals/glimmer/lib/templates/empty.d.ts',
-  '@ember/-internals/glimmer/lib/templates/input.d.ts',
-  '@ember/-internals/glimmer/lib/templates/link-to.d.ts',
-  '@ember/-internals/glimmer/lib/templates/outlet.d.ts',
-  '@ember/-internals/glimmer/lib/templates/root.d.ts',
-  '@ember/-internals/glimmer/lib/templates/textarea.d.ts',
-];
```
ef4, line 113: All of this was helping typescript deal with fake files that don't exist. But now all of them are real files.

packages/ember/version.d.ts
```diff
@@ -1,2 +0,0 @@
-declare const VERSION: string;
```
ef4, line 1: This type was actually a lie. The generated version module only contains a default export. So this PR is also a bugfix to the types.

ef4:
Oh wow, GitHub automatically changed the base branch after merging the other one, nice.

NullVoxPopuli (approved):
Super excited for this! thanks for the self review, that was super helpful!

PR #20591, opened 2023-12-11
docs: Update textarea.ts with link to Class (same way as Input)

johanrd:


NullVoxPopuli (approved):

chancancode:
I wonder if it's more appropriate to use relative links here `../../../...` so that it stays pinned to the same version, but as you said this is just replicating the existing pattern, so that can be looked into separately.

PR #20593, opened 2023-12-11
Enable tsconfig isolatedModules

ef4:
This enables typescript's [isolatedModules](https://www.typescriptlang.org/tsconfig#isolatedModules) flag. It forces our source to be clearer about what is a type vs a value.

This is a step toward a reformed build for ember-source. isolatedModules means we can use babel to handle the typescript instead of needing a global tsc step. This PR doesn't make that switch, that will come as a followup.


packages/@ember/-internals/glimmer/lib/component.ts
```diff
@@ -1000,7 +1000,7 @@ class Component<S = unknown>
     );
 
     let element = _element;
-    let isSVG = element.namespaceURI === Namespace.SVG;
+    let isSVG = element.namespaceURI === 'http://www.w3.org/2000/svg';
```
ef4, line 1003: This code was relying on an implementation detail of typescript enums to turn a type into a runtime value. That is forbidden under isolatedModules because it doesn't work.

Thankfully the value here is literally hard-coded into web standards and will absolutely never change, so I see no downside to copying it into our source here.

packages/@ember/-internals/glimmer/lib/glimmer-component-docs.ts
```diff
@@ -383,3 +383,5 @@
   @module @glimmer/component
   @public
 */
+
+export {};
```
ef4, line 387: This is an empty file that exists to produce documentation. Without any imports or exports typescript worries that it's a non-module script, and that would be an error under isolatedModules.

NullVoxPopuli (approved):
Looks straight-forward, thanks!!

gitKrystan (approved):

PR #20595, opened 2023-12-12
[BUGFIX LTS] An addon re-export should always be a JavaScript file

bertdeblock:
We should never generate `.ts` files in an addon's `app` folder.


kategengler:
Thank you, should this be backported?

kategengler (approved):

PR #20596, opened 2023-12-12
[BUGFIX internal] fix pnpm invocation for nightly TS run

chriskrycho:
pnpm long had a warning for installing to the root of a workspace without this explicit flag, and it has now become an error. In the current shape of the workspace, installing in the root is appropriate, particularly for this job. Accordingly, just add the flag. (This matches the configuration for [regular CI jobs](https://github.com/emberjs/ember.js/blob/b4e52451770e95eec8986578052baffb4f18a72d/.github/workflows/ci.yml#L52).

### DIFF
--- a/.github/workflows/night-ts.yml
+++ b/.github/workflows/night-ts.yml
@@ -7,7 +7,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/setup
-      - run: pnpm add --save-dev typescript@next
+      - run: pnpm add --save-dev typescript@next --workspace-root
       - run: pnpm type-check
   notify:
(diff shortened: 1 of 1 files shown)


NullVoxPopuli (approved):
aka `-w` if folks never type out the full thing :sweat_smile: 

(me)

kategengler:
Thank you!

chriskrycho:
@NullVoxPopuli yep! For scripts in cases like this, I always use the long form because it makes it much easier for anyone else coming along and reading it to understand. In general,  I treat the short form as "useful for typing at the CLI" and the long form as "useful for making a script comprehensible to myself or anyone else".

PR #20597, opened 2023-12-12
[FEATURE] Add explicit support TS 5.1, 5.2, and 5.3

chriskrycho:
Ember has had implicit support for these since their release, and has tested against all of them successfully via the nightly job. Include them explicitly in the CI range so Ember does not *regress* against any of them. (They should also, separately, be added to a public list of supported versions, in line with [the semver-ts spec][spec].)

[spec]: https://www.semver-ts.org/2-conformance.html

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,7 @@ jobs:
     needs: ['types']
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


NullVoxPopuli (approved):

kategengler:
Thank you!

PR #20603, opened 2023-12-12
[BUGFIX beta] internal templates should be strictMode

ef4:
As of https://github.com/emberjs/ember.js/pull/20587, the modules published under `dist/packages` act more like normal addon code,  in that their templates are published as calls to `precompileTemplate` rather than wire format.

But that makes this the first time that Embroider is seeing these templates, and because they are in loose mode and use the dynamic component helper they are un-analyzable.

This PR:
 - switches them to strict mode
 - fixes the declared types for `precompileTemplate`, which were a lie before. `scope` is allowed in non-strict mode. `scope` is optional even in strict mode
 - fixes a spelling error in a recently-introduced internal method, because I noticed it and it annoyed me.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/templates/empty.ts
+++ b/packages/@ember/-internals/glimmer/lib/templates/empty.ts
@@ -1,4 +1,5 @@
 import { precompileTemplate } from '@ember/template-compilation';
 export default precompileTemplate('', {
   moduleName: 'packages/@ember/-internals/glimmer/lib/templates/empty.hbs',
+  strictMode: true,
 });
--- a/packages/@ember/-internals/glimmer/lib/templates/outlet.ts
+++ b/packages/@ember/-internals/glimmer/lib/templates/outlet.ts
@@ -1,4 +1,10 @@
 import { precompileTemplate } from '@ember/template-compilation';
-export default precompileTemplate(`{{component (-outlet)}}`, {
+import { outletHelper } from '../syntax/outlet';
+
+export default precompileTemplate(`{{component (outletHelper)}}`, {
   moduleName: 'packages/@ember/-internals/glimmer/lib/templates/outlet.hbs',
+  strictMode: true,
+  scope() {
+    return { outletHelper };
+  },
 });
--- a/packages/@ember/-internals/glimmer/lib/templates/root.ts
+++ b/packages/@ember/-internals/glimmer/lib/templates/root.ts
@@ -1,4 +1,5 @@
 import { precompileTemplate } from '@ember/template-compilation';
 export default precompileTemplate(`{{component this}}`, {
   moduleName: 'packages/@ember/-internals/glimmer/lib/templates/root.hbs',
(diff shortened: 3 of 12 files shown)


packages/@ember/-internals/glimmer/lib/templates/empty.ts
```diff
@@ -1,4 +1,5 @@
 import { precompileTemplate } from '@ember/template-compilation';
 export default precompileTemplate('', {
   moduleName: 'packages/@ember/-internals/glimmer/lib/templates/empty.hbs',
+  strictMode: true,
```
ef4, line 4: This one was not a problem, I just updated all of them to strict mode because why not?

packages/@ember/-internals/glimmer/lib/templates/outlet.ts
```diff
@@ -1,4 +1,10 @@
 import { precompileTemplate } from '@ember/template-compilation';
-export default precompileTemplate(`{{component (-outlet)}}`, {
+import { outletHelper } from '../syntax/outlet';
+
+export default precompileTemplate(`{{component (outletHelper)}}`, {
```
ef4, line 4: This was one of the problematic cases. In loose mode, this could resolve a string into a component, which would be unsafe to optimize. In strict mode that kind of resolution is forbidden, so embroider can tell this is safe.

packages/@ember/test/index.ts
```diff
@@ -17,7 +17,7 @@ registerWaiter = testingNotAvailableMessage;
 unregisterHelper = testingNotAvailableMessage;
 unregisterWaiter = testingNotAvailableMessage;
 
-export function registerTestImplementaiton(impl: typeof EmberTesting) {
+export function registerTestImplementation(impl: typeof EmberTesting) {
```
ef4, line 20: This is newly-added private API appearing for the first time in this beta. Seems a good time to fix my typo.

ef4:
This PR will be necessary but probably not sufficient to [get the ember-cli embroider scenario passing](https://github.com/ember-cli/ember-cli/pull/10414).

I also see an issue around test setup evaluation order that I'm investigating now.

packages/@ember/-internals/glimmer/lib/templates/root.ts
```diff
@@ -1,4 +1,5 @@
 import { precompileTemplate } from '@ember/template-compilation';
 export default precompileTemplate(`{{component this}}`, {
   moduleName: 'packages/@ember/-internals/glimmer/lib/templates/root.hbs',
+  strictMode: true,
```
ef4, line 4: Notice that this benefits from being strictMode and *does not* need anything from scope. Hence the type declaration change in this PR.

ef4:
The other issue is https://github.com/embroider-build/embroider/pull/1717

NullVoxPopuli (approved):

kategengler (approved):

PR #20606, opened 2023-12-17
Update compiler-options.json

### LINKED ISSUE #20604: Turn on `noImplicitOverride`
Today we discussed and agreed it would be a good idea to turn on the TypeScript `noImplicitOverride` feature. We often use overrides for integration with glimmer-vm etc, so turning that on for our own code would help catch issues during upgrades.

Manas-maker:
Fixes #20604

### DIFF
--- a/tsconfig/compiler-options.json
+++ b/tsconfig/compiler-options.json
@@ -22,6 +22,7 @@
     "esModuleInterop": false,
     "allowSyntheticDefaultImports": false,
     "isolatedModules": true,
+    "noImplicitOverride": true,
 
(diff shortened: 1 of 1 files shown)


anandfresh (approved):

kategengler:
@NullVoxPopuli Do we actually want this change?

NullVoxPopuli:
not as far as I can tell

NullVoxPopuli:
like, more strictness is better, sure, but we have a good chunk of things to work on right now that are more important

PR #20609, opened 2023-12-20
Bump all glimmer-vm dependencies to 0.87.1

NullVoxPopuli:
Contains some perf improvements as well as a memory leak fix for a feature we introduced in 3.25, described here: https://guides.emberjs.com/release/in-depth-topics/rendering-values/

Here is the changelog: https://github.com/glimmerjs/glimmer-vm/releases/tag/v0.86.0


packages/@ember/-internals/glimmer/tests/integration/modifiers/on-test.js
```diff
@@ -32,18 +31,6 @@ moduleFor(
       );
     }
 
-    '@test SUPPORTS_EVENT_OPTIONS is correct (private API usage)'(assert) {
```
NullVoxPopuli, line 35: We don't need this test anymore.
https://github.com/glimmerjs/glimmer-vm/pull/1535

From the implementation notes:
```
/*
  Internet Explorer 11 does not support `once` and also does not support
  passing `eventOptions`. In some situations it then throws a weird script
  error, like:
  ```
  Could not complete the operation due to error 80020101
  ```
  This flag determines, whether `{ once: true }` and thus also event options in
  general are supported.
*/
```

ef4:
This is now passing after I removed ts@next. The PR will still show that as an expected check and I don't think you can fix that without reopening the PR, so I'm just going to ignore that check.

packages/@ember/-internals/glimmer/lib/utils/bindings.ts
```diff
@@ -58,7 +58,9 @@ export function installAttributeBinding(
   let [prop, attribute, isSimple] = parsed;
 
   if (attribute === 'id') {
-    let elementId = get(component, prop);
+    // SAFETY: `get` could not infer the type of `prop` and just gave us `unknown`.
+    //         we may want to throw an error in the future if the value isn't string or null/undefined.
+    let elementId = get(component, prop) as string | null;
```
chriskrycho, line 63: N.b. the right way to do this is to use a dev-time `assert()` that is stripped in prod builds instead!

PR #20617, opened 2024-01-02
Removing ts@next from CI

ef4:
We're seeing potentially-buggy behavior from the currently unreleased typescript. To unblock other PRs I'm taking it out of the CI matrix for the present.


NullVoxPopuli (approved):

ef4:
For some reason `next` was called out explicitly in required status checks in the GitHub settings. I removed that, but it doesn't update this PR.

PR #20620, opened 2024-01-03
[CLEANUP canary] Remove final eager consumption of AMD loader

ef4:
This is the last place where Ember treats the AMD loader as mandatory. After this change I have Ember's whole test suite passing with no AMD loader!

### DIFF
--- a/packages/ember/index.ts
+++ b/packages/ember/index.ts
@@ -506,8 +506,10 @@ namespace Ember {
     get define() {
       return (globalThis as any).define;
     },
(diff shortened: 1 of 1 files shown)


NullVoxPopuli (approved):

ef4:
(I thought I had already removed the "ts@next expected" rule from the branch protections but apparently I didn't hit save last time. This PR is passing all other tests and that one will stay yellow forever.)

ef4:
Oh, nevermind, github updated so we're green.

PR #20623, opened 2024-01-15
Replace master with main in links and text

ellnix:
Saw that clicking on "see CONTRIBUTING" made github tell me that master was renamed to main so I thought I would update the links to point to `main`.

This is my first time looking at this project so I apologize in advance if I changed something erroneously.


kategengler:
Thank you!

PR #20624, opened 2024-01-16
Enable noImplicitOverride and add override keyword to properties/functions that are overridden.

### LINKED ISSUE #20604: Turn on `noImplicitOverride`

shama:
Fixes GH-20604

This enables `noImplicitOverride` in TypeScript and adds the `override` keyword to all properties and functions that have been overridden.


kategengler:
This isn't currently desired (see #20606)

PR #20629, opened 2024-01-24
[bugfix] allow `modifier` helper to resolve built-ins like `on`

### LINKED ISSUE #19869: [Bug] The `on` modifier does not work with the `modifier` helper
### 🐞 Describe the Bug

This works:

```hbs
<div {{(if this.condition (modifier 'my-modifier' this.bar))}}>
```

### 🔬 Minimal Reproduction

This doesn't:

```hbs
<div {{(if this.condition (modifier 'on' 'scroll' this.bar))}}>
```

Can't create a Twiddle because Twiddle is severely outdated, sorry.

### 😕 Actual Behavior

> Uncaught Error: Assertion Failed: Attempted to invoke `(-resolve "modifier:on")`, but on was not a valid modifier name.

### 🤔 Expected Behavior

Should apply the `on` modifier to the element.

### 🌍 Environment

- Ember: 3.28.4
- Node.js/npm: 14.16.0
- OS: Build: Ubuntu, Browser: Windows 10
- Browser: Chrome, Firefox

CC @simonihmig

fivetanley:
partially fixes #19869

Due to [GlimmerVM's assertion][1], this still does not fix the issue in strict mode. We also can't work around the strict mode limitation with an AST transform since Glimmer's transforms run first.

[1]: https://github.com/glimmerjs/glimmer-vm/blob/2ddbbc4a9b97db4f326c4d92021f089c464ab093/packages/%40glimmer/compiler/lib/passes/1-normalization/keywords/utils/curry.ts#L53

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/resolver.ts
+++ b/packages/@ember/-internals/glimmer/lib/resolver.ts
@@ -151,7 +151,7 @@ const BUILTIN_KEYWORD_MODIFIERS: Record<string, ModifierDefinitionState> = {
   action: actionModifier,
 };
 
-const BUILTIN_MODIFIERS: Record<string, object> = {
+export const BUILTIN_MODIFIERS: Record<string, object> = {
   ...BUILTIN_KEYWORD_MODIFIERS,
   on,
 };
--- a/packages/@ember/-internals/glimmer/lib/setup-registry.ts
+++ b/packages/@ember/-internals/glimmer/lib/setup-registry.ts
@@ -12,6 +12,7 @@ import { Renderer } from './renderer';
 import OutletTemplate from './templates/outlet';
 import RootTemplate from './templates/root';
 import OutletView from './views/outlet';
+import { BUILTIN_MODIFIERS } from './resolver';
 
 export function setupApplicationRegistry(registry: Registry): void {
   // because we are using injections we can't use instantiate false
@@ -56,5 +57,9 @@ export function setupEngineRegistry(registry: Registry): void {
 
   if (!ENV._TEMPLATE_ONLY_GLIMMER_COMPONENTS) {
     registry.register(P`component:-default`, Component);
+  }
+
+  for (let [name, modifier] of Object.entries(BUILTIN_MODIFIERS)) {
+    registry.register(`modifier:${name}`, modifier, { instantiate: false });
   }
 }
--- a/packages/@ember/-internals/glimmer/tests/integration/custom-modifier-manager-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/custom-modifier-manager-test.js
@@ -476,6 +476,28 @@ moduleFor(
       this.assertStableRerender();
     }
 
+    '@test can resolve built-in modifiers'(assert) {
+      let wasCalled = false;
+      let id = 'wow-what-an-original-id';
+      this.render(
+        `<div id='${id}' {{(if this.isModifying (modifier "on" "click" this.callAction))}} />`,
+        {
+          callAction: () => {
+            wasCalled = true;
+          },
+        }
+      );
+
+      let element = document.querySelector(`#${id}`);
+      element.click();
+
+      assert.false(wasCalled, 'modifier should not be set up');
+
+      runTask(() => set(this.context, 'isModifying', true));
+      element.click();
+      assert.true(wasCalled, 'on modifier can be used');
+    }
+
     '@test Cannot dynamically resolve a modifier'(assert) {
       this.registerModifier(
         'replace',


NullVoxPopuli:
Why does ember have its own record of this stuff? I would have thought this behavior (and fix) would be in glimmer-vm, since `modifier` is a keyword.

fivetanley:
@NullVoxPopuli I can't speak for the original authors', but the implementation currently rewrites

```handlebars
{{modifier "foo"}}
````

to

```handlebars
{{modifier (-resolve "modifier:foo")}}
```

via [this AST transform](https://github.com/emberjs/ember.js/blob/main/packages/ember-template-compiler/lib/plugins/transform-resolutions.ts).

The `-resolve` helper basically does a [`owner.hasRegistration` check](https://github.com/emberjs/ember.js/blob/be51c93b2b79f5fca8d36db1fcae36a049aeeb25/packages/%40ember/-internals/glimmer/lib/helpers/-resolve.ts#L38) and doesn't use the [resolver's logic at all](https://github.com/emberjs/ember.js/blob/be51c93b2b79f5fca8d36db1fcae36a049aeeb25/packages/%40ember/-internals/glimmer/lib/resolver.ts#L219).

I'm not sure why Ember has knowledge of these things vs these checks/etc being done in Glimmer itself.

NullVoxPopuli:
that's goofy. 

I wonder if we can instead do all the resolution at build time instead of runtime.

I feel like I have either a very faint memory, or I made one up, of @ef4 mentioning something about this in the last week or two.

Like, automatically converting things to strict-mode so there's no runtime cost to any of this :thinking:

ef4:
The `modifier` helper has always been forbidden from accepting strings. It must not accept strings.

The same is true for the `helper` helper. Only `component` accepts them, and that only in non-strict mode, because it makes static analysis impossible.

ef4:
But also: do people not realize it's a very bad idea to be using this feature at all in new code in 2024? It moves you further away from adopting template tag and further away from adopting embroider. I promise I will work hard to break it in Ember 6.0.

If you need to yield a contextual modifier, Import your modifier in javascript and use the value directly. That works all the way back to at least Ember 3.28. That's what's aligned with template tag and embroider both.

fivetanley:
@ef4 should this be the case for built-ins like `on` ? It feels cumbersome to have to find the import for this (I can't seem to find it in the documentation, had to source dive for it). I noticed @NullVoxPopuli opened several RFCs related to this, one of them making it so `on` doesn't need an import https://github.com/emberjs/rfcs/pull/997

fivetanley:
Closing this for now since I found a workaround to get around my problem. This could potentially be revisited in Glimmer itself if the RFCs in my previous comment are accepted. Here's a link to the workaround https://github.com/emberjs/ember.js/issues/19869#issuecomment-1909118910

ef4:
Adding more built-in keywords would probably be fine but that's very different from resolving strings to modifiers.

chancancode:
That is not correct, this is a long standing bug (linked issue in the original post). All of these accept strings in non-strict mode templates as specified in the RFC; we specifically agreed to make them all work consistently in normal templates and made the static resolution requirement strictly a strict-mode thing.

packages/@ember/-internals/glimmer/lib/resolver.ts
```diff
@@ -151,7 +151,7 @@ const BUILTIN_KEYWORD_MODIFIERS: Record<string, ModifierDefinitionState> = {
   action: actionModifier,
 };
 
-const BUILTIN_MODIFIERS: Record<string, object> = {
+export const BUILTIN_MODIFIERS: Record<string, object> = {
   ...BUILTIN_KEYWORD_MODIFIERS,
   on,
 };
```
chancancode, line 157: This is used in `lookupModifier` on L219. The problem is, it seems like, `(modifier "...")` didn't get implemented in glimmer-vm, or at least when we landed this feature in ember, it didn't, so ember added an AST transform hack to change `(modifier "...")` into `(modifier (-resolve "modifier:..."))`. If that was implemented in glimmer-vm properly like `(component "...")` is, then it should/would have gone through the codepath in L219 and everything should work. That would be the most proper fix.

I was a bit hesitant on fixing it by adding them to the resolver, but I suppose that would be [the most consistent](https://github.com/emberjs/ember.js/blob/be51c93b2b79f5fca8d36db1fcae36a049aeeb25/packages/%40ember/-internals/glimmer/lib/setup-registry.ts#L52-L55). So I suppose why not.

However, I do think we need to fix this properly upstream; the problem with the current AST-transform hack in Ember is it will, as far as I can tell, still do the same AST transform in strict mode templates, which is probably punching a pretty big hole into the entire Embroider/static template universe.

NullVoxPopuli:
speaking of keywords, on is (on main) a keyword now

kategengler:
Moving to draft

NullVoxPopuli:
a few things have changed since this PR was submitted:
- on is a keyword now
- strict mode is default

[things have been fixed](https://limber.glimdown.com/edit?c=JYWwDg9gTgLgBAYQuCA7Apq%2BAzKy4DkAAgOYA2oI6UA9AMbKQZYEDcAUKJLHAN5wwoAQzoBrdABM4AXzi58xcpWo1BI0cFQk27dugAe3eBPTYhAVzLw6ZIQGc7cABLoyZCAHVoZKQZiYJRyQUZnhedjg4IjUxSTgGcyw4AF44AAYOCLhNOih0KiTUgAoAShSAPgEAC2A7ADoEpIBqVIBGTMiAHn9wW39yrK6wcoBNCHM4KqEAN3R4ilipGCq5gCNzGBg0Pl5l2obxrGlZGFB0es6aYd1IyN5eAGIydHgikAgJYGxgajhtghswDEBGq%2BxyeQKMDK9jgAB9VhAIGBYcdBrdOutNtsYABPMDoZIAIkxW1QhJ2CKRx3KCAWokuJLQA1uOxozxgqK6qnyYD66AG0nYQA&format=gjs)

```gjs

export default class HelloWorld extends Component {
  @tracked count = 0;

  increment = () => this.count += 1;

  <template>
    <p>You have clicked the button {{this.count}} times.</p>

    {{#let (modifier on 'click' this.increment) as |boop|}}
      <button type="button" {{boop}}>Click</button>
    {{/let}}
  </template>
}
```

PR #20645, opened 2024-02-07
[BUGFIX beta] Make sure the `component-class` blueprint generates a signature in a TS project

### LINKED ISSUE #20598: [Bug] `ember g component-class foo` does not generate a component signature in a TypeScript project
### 🐞 Describe the Bug

`ember g component-class foo` does not generate a component signature in a TypeScript project.

### 🔬 Minimal Reproduction

Run `ember g component-class foo` in a TS project.

### 😕 Actual Behavior

```js

export default class FooComponent extends Component {}
```

### 🤔 Expected Behavior

```ts

interface FooSignature {
  // The arguments accepted by the component
  Args: {};
  // Any blocks yielded by the component
  Blocks: {
    default: []
  };
  // The element to which `...attributes` is applied in the component template
  Element: null;
}

export default class FooComponent extends Component<FooSignature> {}
```

### 🌍 Environment

- Ember: 5.4
- Node.js/npm: /
- OS: /
- Browser: /

### ➕ Additional Context

/

bertdeblock:
Closes https://github.com/emberjs/ember.js/issues/20598.

### DIFF
--- a/blueprints/-utils.js
+++ b/blueprints/-utils.js
@@ -1,9 +1,33 @@
 const { dasherize } = require('ember-cli-string-utils');
+const { EOL } = require('os');
+
+function generateComponentSignature(componentName) {
+  let args = `  // The arguments accepted by the component${EOL}  Args: {};`;
+
+  let blocks =
+    `  // Any blocks yielded by the component${EOL}` +
+    `  Blocks: {${EOL}` +
+    `    default: []${EOL}` +
+    `  };`;
+
+  let element =
+    `  // The element to which \`...attributes\` is applied in the component template${EOL}` +
+    `  Element: null;`;
+
+  return (
+    `export interface ${componentName}Signature {${EOL}` +
+    `${args}${EOL}` +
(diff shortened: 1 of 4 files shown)


blueprints/-utils.js
```diff
@@ -1,9 +1,33 @@
...
+    `  Blocks: {${EOL}` +
+    `    default: []${EOL}` +
+    `  };`;
+
+  let element =
+    `  // The element to which \`...attributes\` is applied in the component template${EOL}` +
+    `  Element: null;`;
+
+  return (
+    `interface ${componentName}Signature {${EOL}` +
```
dfreeman, line 18: Generally the guidance we've given with Glint is to always export your signature, since it's useful when you start talking about places where other components are yielding or or accepting this one as an arg.

bertdeblock: Thanks @dfreeman. Updated accordingly.

kategengler:
Thank you!

PR #20656, opened 2024-03-06
[BUGFIX] Update router.js

wagenet:
This addresses a real bug that AuditBoard was hitting. However, nobody else has reported this, so it's not clear if we need to back port this.

Bug Fix PR: https://github.com/tildeio/router.js/pull/336

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,7 +99,7 @@
     "ember-router-generator": "^2.0.0",
     "inflection": "^2.0.1",
     "route-recognizer": "^0.3.4",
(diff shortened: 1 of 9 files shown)


NullVoxPopuli (approved):

kategengler:
@wagenet Can you describe what the bug looked like to a user of ember.js? (to inform the changelog)

wagenet:
What would happen is that in certain cases, the `transition.to` would lose the `queryParams` that the previous state actually did have.

PR #17499, opened 2019-01-22
[BUGFIX beta] Update to glimmer-vm 0.37.1.

rwjblue:
Fixes a few issues:

* Usage of positional arguments with custom components.
* Forwarding attributes via `...attributes` to a dynamic
  component.
* Prevent errors when rendering many template blocks
  (`Error: Operand over 16-bits. Got 65536`).

Full changes here:

https://github.com/glimmerjs/glimmer-vm/compare/v0.37.0...v0.37.1

---

Closes #17146
Closes #17160
Closes #17483

PR #20658, opened 2024-03-09
VM Upgrade: T.S. '89

NullVoxPopuli:
image

Changelog: https://github.com/glimmerjs/glimmer-vm/releases/tag/v0.89.0

I had some issues with `this.attrs`, so I just remove the problematic tests, since `attrs` in general was supposed to be removed in v4, see: https://github.com/emberjs/ember.js/pull/20671 for that effort.


chancancode:
Rebase after https://github.com/emberjs/ember.js/pull/20672 lands

packages/@ember/-internals/glimmer/lib/modifiers/action.ts
```diff
@@ -200,6 +200,10 @@ class ActionModifierManager implements InternalModifierManager<ActionState, obje
     return new ActionState(element, owner, actionId, actionArgs, named, positional);
   }
 
+  getDebugInstance(state: ActionState): unknown {
+    return state;
+  }
```
chancancode, line 205: I think we talked about this and we want to return `null`? We don't really want to leak the internal state bucket, and it's not really useful for it to show up on the inspector

packages/@ember/-internals/glimmer/lib/modifiers/internal.ts
```diff
@@ -78,6 +78,10 @@ export class InternalModifierManager
     return null;
   }
 
+  getDebugInstance(state: InternalModifierState): unknown {
+    return state;
+  }
```
chancancode, line 83: I think just declare the class `abstract` and leave this out. We have to decide case-by-case what we want to expose here. And so far this is only used by `{{action}}` anyway

packages/@ember/-internals/glimmer/tests/integration/components/curly-components-test.js
```diff
@@ -2492,7 +2492,7 @@ moduleFor(
           },
           value: null,
         }),
-        template: '<div id="inner-value">{{wrapper.content}}</div>',
+        template: '<div id="inner-value">{{this.wrapper.content}}</div>',
```
chancancode, line 2495: Something is fishy/scary that these continued to work, seemingly without assertions/deprecations either. Hopefully the new VM code is erroring consistently now, but this means we may have implicit this code in the wild that was accidentally working

packages/@ember/-internals/glimmer/tests/integration/helpers/hash-test.js
```diff
@@ -217,92 +217,5 @@ moduleFor(
...
-
-      runTask(() => {
-        expectDeprecation(() => {
-          set(fooBarInstance.hash, 'firstName', 'Godfrey');
-          set(fooBarInstance.hash, 'lastName', 'Chan');
-        }, /You set the '.*' property on a {{hash}} object/);
-      });
-
-      this.assertText('Godfrey Chan');
-    }
```
chancancode, line 306: This went from deprecation to nothing (instead of an assertion). I don't think it's illegal per-se, I think we are technically allowed to do it, but that's when we tend to get ourselves into trouble – see the `{{#with}}` thing recently. Next time at least consider whether we can keep an assertion around for at least 1-2 LTS.

packages/ember-template-compiler/lib/plugins/assert-against-attrs.ts
```diff
@@ -81,9 +81,8 @@ function isAttrs(node: AST.PathExpression, symbols: string[]) {
   }
 
   if (name === 'attrs') {
-    if (node.this === true) {
-      node.parts.shift();
-      node.original = node.original.slice(5);
+    if (node.head.type === 'ThisHead') {
+      node.tail.shift();
```
chancancode, line 85: Rebase away

packages/ember-template-compiler/lib/plugins/transform-resolutions.ts
```diff
@@ -149,7 +149,9 @@ export default function transformResolutions(env: EmberASTPluginEnvironment): AS
 }
 
 function isLocalVariable(node: AST.PathExpression, hasLocal: (k: string) => boolean): boolean {
-  return !node.this && node.parts.length === 1 && hasLocal(node.parts[0]!);
+  return (
+    !(node.head.type === 'ThisHead') && node.tail.length === 1 && hasLocal(node.head.original!)
```
chancancode, line 153: Why the ! on `.original`? It should always exist?

packages/ember-template-compiler/lib/plugins/transform-wrap-mount-and-outlet.ts
```diff
@@ -43,8 +43,10 @@ export default function transformWrapMountAndOutlet(env: EmberASTPluginEnvironme
     name: 'transform-wrap-mount-and-outlet',
 
     visitor: {
-      Program: node,
+      Template: node,
       ElementNode: node,
+      BlockStatement: node,
```
chancancode, line 48: this is wrong (see below)

packages/ember-template-compiler/lib/plugins/utils.ts
```diff
@@ -12,25 +12,41 @@ export function isStringLiteral(node: AST.Expression): node is AST.StringLiteral
...
+  return node;
+}
+
 export function trackLocals() {
   let locals = new Map();
 
   let node = {
-    enter(node: AST.Program | AST.Block | AST.ElementNode) {
-      for (let param of node.blockParams) {
+    enter(node: AST.Template | AST.Block | AST.ElementNode | AST.BlockStatement) {
```
chancancode, line 31: You don't want `AST.BlockStatement` here. `BlockStatement` is the `{{#foo}}...{{/foo}}`, the block params is on the default/inverse blocks (which should get visited by the `Block` visitor). 

There _is_ a `params` on `BlockStatement` but it means a completely different thing, it is the positional argument for the invocation, which has a type of `AST.Expression[]` hence you question above.

On the other hand, `node.params` on `AST.Template | AST.Block | AST.ElementNode` should all be `AST.VarHead[]`.

packages/ember-template-compiler/lib/plugins/utils.ts
```diff
@@ -12,25 +12,41 @@ export function isStringLiteral(node: AST.Expression): node is AST.StringLiteral
...
+      let params = 'params' in node ? node.params : node.blockParams;
```
chancancode, line 32: All the nodes you checked here should have `params`. Why this check? Unless you intend for this `enter` hook to be more generic, in which case you need to type `node: AST.Node` and do a different check here, because there are other things with a `params` property that is unrelated.

packages/ember-template-compiler/lib/plugins/utils.ts
```diff
@@ -12,25 +12,41 @@ export function isStringLiteral(node: AST.Expression): node is AST.StringLiteral
...
+      for (let param of params) {
+        let name = getLocalName(param);
```
chancancode, line 34: Once you fix the types/guards `getLocalName` shouldn't need to exist

packages/ember-template-compiler/tests/plugins/assert-splattribute-expression-test.js
```diff
@@ -1,48 +0,0 @@
-import { moduleFor, AbstractTestCase } from 'internal-test-helpers';
```
chancancode, line 1: I think we should keep this around and just change to use `assert.throws`. Generally we want to keep the coverage even if it duplicates what is already tested in the VM. The rule of thumb is if some day we decided to replace `glimmer-vm` with something else, we shouldn't loose coverage on anything that ember cares about.

types/publish.mjs
```diff
@@ -109,8 +109,7 @@ async function main() {
 
   // The majority of those items should be excluded entirely, but in some cases
   // we still need to post-process them.
-  let excludes = remappedLocationExcludes
-    .concat(sideEffectExcludes);
+  let excludes = remappedLocationExcludes.concat(sideEffectExcludes);
```
chancancode, line 112: This seems unrelated, it doesn't hurt but I'd revert it just to keep the diff minimal. I did notice we don't really have prettier set up here, which we can add in a different PR if we want.

packages/ember-template-compiler/tests/plugins/transform-component-invocation-test.js
chancancode, line 1: I don't know what this test file is trying to test anymore 😐

This probably corresponds to some kind of transform we did here to make some of these syntax work before they were implemented in the VM (this file probably predates the VM even existing), but we don't do that anymore, and the test doesn't actually do anything other than, checking these syntax compiles without error. 

Surely I hope that is already covered indirectly by any tests that actually verifies/exercises the feature. This is not that different from smoke testing `<p>hi</p>` and `{{this.foo}}` compiles without error, which, I suppose isn't inherently useless, but if we are going to do that we would want to add a more cases.

Not really your problem but I wonder if we should just delete this whole file.

packages/@ember/-internals/glimmer/lib/modifiers/internal.ts
NullVoxPopuli: TS says its required, -- I'll return null by default tho

packages/@ember/-internals/glimmer/lib/modifiers/action.ts
NullVoxPopuli: right! updated

packages/ember-template-compiler/lib/plugins/transform-resolutions.ts
NullVoxPopuli: yes, but nothing complained about it. It's been removed now.

packages/ember-template-compiler/lib/plugins/utils.ts
NullVoxPopuli: `params does not exist on type Template` - we can play with removing it I guess, but I feel like it was needed for something and I don't remember -- I'll see if I rediscover that tomorrow

packages/ember-template-compiler/lib/plugins/utils.ts
NullVoxPopuli: Still needs to exist because `param` can be of type string (from `blockParams`), which exist on `Template`

types/publish.mjs
NullVoxPopuli: yeah, it's annoying to have an editor that enables prettier when a `.prettierrc.js` is detected, but some files haven't been formatted with it. :(

NullVoxPopuli: updated

packages/ember-template-compiler/lib/plugins/transform-wrap-mount-and-outlet.ts
NullVoxPopuli: thanks! updated!

packages/ember-template-compiler/lib/plugins/assert-against-attrs.ts
NullVoxPopuli: other PR isn't merged yet :sweat_smile:

packages/ember-template-compiler/lib/plugins/utils.ts
chancancode: oh yeah, that’s right, cause we don’t have locs for the js scope locals 👍

packages/ember-template-compiler/lib/plugins/assert-against-attrs.ts
chancancode: Merged now!

NullVoxPopuli: updated!

packages/ember-template-compiler/tests/plugins/assert-splattribute-expression-test.js
NullVoxPopuli: `@glimmer/syntax` would still be used tho, even if we swap out the VM

NullVoxPopuli: (I'm intending on leaving this file deleted, partly because `@glimmer/syntax` is not glimmer-vm)

packages/@ember/-internals/glimmer/tests/integration/components/curly-components-test.js
kategengler: I think `expectAssertion` below would have swallowed any other assertions, even if it matched on `backtrackingMessageFor`

packages/ember-template-compiler/tests/plugins/assert-splattribute-expression-test.js
chancancode: Not necessarily, all those packages were invented out of thin air at the same time, it’s still an external (or internal, I guess?) dependency and ember needs to test its public api

NullVoxPopuli: It has it's own tests to tho. And we're planning on moving all that in to this repo

chancancode: Yes, that's what we have _always_ done. I didn't invent it just to slow down this particular PR. Often we have almost exactly the same test twice. See, for example: https://github.com/emberjs/ember.js/pull/17499

The tests are already there and it should be easy enough to change them to match (it's ok to not match the error messages very precisely, since the point is just to check that there are _some_ errors. 

Could we revisit how we do thing in general? Yes? But as you said you don't also want the creep the scope of the PR so unless this is a particularly important hill for you we can discuss that another time/day.

If it did turn out to be very challenging to keep the tests working we can of course revisit if it's worth the time. But as a matter of default, this is the status quo (and I don't think it's a pointless status quo/just useless formalities), and is what we should do.

NullVoxPopuli: The intent of this file has been restored, albeit, rewritten due to the nature of how the error was surfaced.

I removed the line checking, because, by the logic given, it's not relevant, and we want to make sure that syntax restrictions don't regress, due to not trusting `@glimmer/syntax` :sweat_smile: (which I don't agree with, but I can't care anymore :shrug: <3 )

packages/ember-template-compiler/tests/plugins/assert-splattribute-expression-test.js
```diff
@@ -1,48 +1,27 @@
...
-      expectAssertion(() => {
-        compile('<div>{{...attributes}}</div>', {
-          moduleName: 'foo-bar',
-        });
-      }, this.expectedMessage(`'foo-bar' @ L1:C7`));
+    '@test {{...attributes}} is not valid path expression'(assert) {
+      assert.throws(() => compile('<div>{{...attributes}}</div>'), message);
     }
 
-    '@test {{...attributes}} is not valid path expression'() {
```
NullVoxPopuli, line 25: I deleted this test entirely because it was the exact same as the one above

packages/@ember/-internals/glimmer/lib/modifiers/internal.ts
chancancode: Yeah, that's why I said *declare the class `abstract`*... when you _concretely_ implement/extend an abstract class, you are required to implement all the abstract members (which is the point of abstract classes). 

_However,_ you can also abstractly extend a class (`abstract class Foo extends AnotherAbstractClass`), in which case, you are allowed to leave out any abstract members, essentially deferring/propagating the requirement to your children

packages/@ember/-internals/glimmer/lib/modifiers/internal.ts
```diff
@@ -78,6 +78,10 @@ export class InternalModifierManager
     return null;
   }
 
+  getDebugInstance(): unknown {
+    return null;
+  }
+
```
chancancode, line 84:
```suggestion
```

chancancode: Was hoping I could use this to commit the change directly from the UI, saving a roundtrip, but it wouldn't let me also add a diff for the `abstract class ...` because that line was unchanged 😐

NullVoxPopuli: this is a required method

```
Class 'InternalModifierManager' incorrectly implements interface 'InternalModifierManager<InternalModifierState, typeof InternalModifier>'.
     Property 'getDebugInstance' is missing in type 'InternalModifierManager' but required in type 'InternalModifierManager<InternalModifierState, typeof InternalModifier>'. [2420]
```

packages/@ember/-internals/glimmer/lib/modifiers/internal.ts
```diff
@@ -78,9 +78,7 @@ export class InternalModifierManager
     return null;
   }
 
-  getDebugInstance(): unknown {
-    return null;
-  }
+  abstract getDebugInstance(state: InternalModifierState): unknown;
```
NullVoxPopuli, line 81: ah, this is what you meant

packages/@ember/-internals/glimmer/lib/modifiers/internal.ts
NullVoxPopuli: nevermind: https://github.com/emberjs/ember.js/pull/20658/commits/a13c6275687d119a8de0dd9cb0f9755df2b8ba4d#r1544906209

packages/@ember/-internals/glimmer/lib/modifiers/internal.ts
chancancode: Yeah, it’s a little bit of an edge case because we are implementing an interface. If we are extending an abstract super class you wouldn’t have to explicitly propagate it like this.

packages/ember-template-compiler/lib/plugins/index.ts
```diff
@@ -36,8 +34,6 @@ export const STRICT_MODE_TRANSFORMS = Object.freeze(
     TransformActionSyntax,
     TransformEachInIntoEach,
     TransformInElement,
-    AssertSplattributeExpressions,
-    TransformEachTrackArray,
```
chancancode, line 40: This is revealing a real issue: https://github.com/emberjs/ember.js/pull/20673

We have to add it back, and deal it somehow on the glimmer-vm side. My initial thought is that glimmer-vm can assume everything that starts with a `-` is a keyword that ember will provide at runtime and not error on those (since it couldn't have been a JavaScript identifier), but maybe that will mask too many real errors when people are migrating existing templates (not uncommon to have helpers starting with dash in apps).

Maybe we'll have to add a `KeywordHead` subnode, or just make it something you are very unlikely to type, like `__GLIMMER_KEYWORD__` or something. Not sure.

But it is a real problem. If we don't deal with it, we would have broken `{{outlet}}`, `{{mount}}`, {{#each-in}}`, `{{#in-element}}` (perhaps among other things) for strict mode templates

chancancode (approved):

PR #20660, opened 2024-03-13
Implement RFC#995, Deprecate component template resolution

NullVoxPopuli:
RFC: https://github.com/emberjs/rfcs/pull/995
Supersedes: https://github.com/ember-cli/ember-resolver/pull/971


gitKrystan (approved):

packages/@ember/-internals/glimmer/lib/component-managers/curly.ts
```diff
@@ -133,6 +134,12 @@ export default class CurlyComponentManager
     let factory: TemplateFactory;
 
     if (layout === undefined) {
+      let theSetTemplate = getComponentTemplate(component);
```
NullVoxPopuli, line 137: a chance to early exit in classic land without resolving a layout

packages/@ember/-internals/glimmer/tests/integration/application/debug-render-tree-test.ts
```diff
@@ -811,7 +830,7 @@ if (ENV._DEBUG_RENDER_TREE) {
             name: 'hello-world',
             args: { positional: [], named: { name: 'first' } },
             instance: null,
-            template: 'my-app/templates/components/hello-world.hbs',
```
NullVoxPopuli, line 814: this field is based of `moduleName`, which we've talked about getting rid of. 

It still works when a template is resolved, but when using `setComponentTemplate`, there is no module name to associate -- could come from anywhere

ef4:
@mixonic volunteered to help review this

packages/@ember/-internals/glimmer/lib/resolver.ts
```diff
@@ -67,6 +68,11 @@ function layoutFor(
 ): Nullable<Template> {
   let templateFullName = `template:components/${name}` as const;
 
+  deprecateUntil(
```
ef4, line 71: It looks like this will fire a deprecation if somebody has a JS with no HBS at all?

I think we would only want to fire the deprecation if `layoutFor` successfully finds a result.

packages/@ember/-internals/glimmer/lib/resolver.ts
```diff
@@ -99,6 +105,12 @@ function lookupComponentPair(
     }
   }
 
+  if (isRemoved(DEPRECATIONS.DEPRECATE_COMPONENT_TEMPLATE_RESOLVING.options)) {
```
kategengler, line 108: You should do `DEPRECATIONS.DEPRECATE_COMPONENT_TEMPLATE_RESOLVING.isRemoved`

packages/@ember/-internals/glimmer/tests/integration/components/append-test.js
```diff
@@ -37,7 +38,9 @@ class AbstractAppendTest extends RenderingTestCase {
     this.ids.push(component.elementId);
   }
 
-  ['@test lifecycle hooks during component append'](assert) {
+  [`${testUnless(
+    DEPRECATIONS.DEPRECATE_COMPONENT_TEMPLATE_RESOLVING.isRemoved
+  )} lifecycle hooks during component append`](assert) {
```
ef4, line 43: This seems like test coverage that we would still want to keep after removing the deprecation?

packages/@ember/-internals/glimmer/tests/integration/components/append-test.js
```diff
@@ -283,7 +286,9 @@ class AbstractAppendTest extends RenderingTestCase {
     );
   }
 
-  ['@test appending, updating and destroying a single component'](assert) {
+  [`${testUnless(
+    DEPRECATIONS.DEPRECATE_COMPONENT_TEMPLATE_RESOLVING.isRemoved
+  )} appending, updating and destroying a single component`](assert) {
```
ef4, line 291: Similarly, I think we want to keep this test after deprecation removal?

packages/@ember/-internals/glimmer/tests/integration/components/append-test.js
```diff
@@ -389,7 +394,9 @@ class AbstractAppendTest extends RenderingTestCase {
     assert.equal(renderer._roots.length, 0, 'released the root component');
   }
 
-  ['@test appending, updating and destroying multiple components'](assert) {
+  [`${testUnless(
+    DEPRECATIONS.DEPRECATE_COMPONENT_TEMPLATE_RESOLVING.isRemoved
+  )} appending, updating and destroying multiple components`](assert) {
```
ef4, line 399: ditto

packages/@ember/-internals/glimmer/lib/component-managers/curly.ts
ef4: This is a little surprising in the context of a deprecation PR. Do we have a bug here? Like, are there cases where an `@ember/component` doesn't support template colocation without this change?

If there's a bugfix we might want to land it separately from adding the deprecation. And make sure it has its own test coverage. (Not sure if this PR includes such coverage, haven't found it yet.)

NullVoxPopuli: >  doesn't support template colocation without this change?

it does -- and I confirmed by adding a specific test about `getComponentTemplate` and removing the code here -- however -- this code is about looking up the template for the debug tree -- which is no longer a problem -- so I'm going to just remove this code as it looks like all the tests pass.

There isn't a bug to fix, imo.

packages/@ember/-internals/glimmer/lib/resolver.ts
NullVoxPopuli: you got it!

packages/@ember/-internals/glimmer/tests/integration/components/append-test.js
NullVoxPopuli: I can't figure this out -- test is too tied to specific VM behaves, it seems

packages/@ember/-internals/glimmer/tests/integration/components/append-test.js
NullVoxPopuli: for all of the append tests: https://github.com/emberjs/ember.js/pull/20660#discussion_r1571117787

packages/@ember/-internals/glimmer/lib/resolver.ts
NullVoxPopuli: fixed! thanks!

packages/@ember/-internals/glimmer/tests/integration/components/append-test.js
NullVoxPopuli: I did figure something out by not using any of the same test apis, and requiring slightly different assertions due to using the whole rendering system instead of manually creating factories, but I think the intent of the test is maintained in the new way

packages/@ember/-internals/glimmer/tests/integration/components/append-test.js
NullVoxPopuli: the newly added test covers these use cases (destroy and update)

packages/@ember/-internals/glimmer/tests/integration/components/append-test.js
NullVoxPopuli: the newly added test covers these use cases

ef4 (approved):

PR #20666, opened 2024-03-22
[FEATURE] Assert on deprecations in ember-source that have an until older or equal to the current ember-source version

kategengler:
- [ ] String.prototype.replaceAll apparently only since Safari 13ish
- [ ] What is going on with `@ember/version` module in Node tests and in Smoke tests?
- [ ] Throwing on the `until` will only happen in development due to deprecation stripping
- [ ] What to do about `expectDeprecation`? Other than the above issues this works fine without modifying any tests or test helpers because Ember's own tests do not exercise `deprecate` but completely stub it. The modifications here to test files and helpers are an attempt to allow the test suite to run both with deprecations all throwing and as they are default. These changes keep getting grosser. To the point I think either we need to rework `expectDeprecation` and/or add an `internalDeprecate` that is used internally by Ember that will not be stripped in production builds.

### DIFF
--- a/packages/internal-test-helpers/lib/ember-dev/deprecation.ts
+++ b/packages/internal-test-helpers/lib/ember-dev/deprecation.ts
@@ -177,6 +177,8 @@ class DeprecationAssert extends DebugAssert {
         tracker.expectCall(message, ['id', 'until']);
       });
     } else {
+      let { assert } = QUnit.config.current;
+      assert.ok(true, 'Balance assertion counts');
       this.expectNoDeprecation(func);
     }
   }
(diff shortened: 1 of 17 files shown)


packages/internal-test-helpers/lib/ember-dev/deprecation.ts
```diff
@@ -177,6 +177,8 @@ class DeprecationAssert extends DebugAssert {
         tracker.expectCall(message, ['id', 'until']);
       });
     } else {
+      let { assert } = QUnit.config.current;
+      assert.ok(true, 'Balance assertion counts');
```
NullVoxPopuli, line 181: <3

kategengler:
`expectDeprecation` should fail when deprecation is throwing. 

Tests and code need to be guarded as if we are removing the code.

PR #20668, opened 2024-03-26
Run all tests with all deprecations enabled, and with them at the default state for the version

kategengler:
- Create a listing of deprecations that should be used as the source of truth for deprecations issued by Ember itself
- Convention to pass the test/isEnabled to deprecate/expectDeprecation so that our testing mechanism will work

### DIFF
--- a/packages/@ember/-internals/deprecations/index.ts
+++ b/packages/@ember/-internals/deprecations/index.ts
@@ -12,6 +12,50 @@ function deprecation(options: DeprecationOptions) {
     isEnabled: isEnabled(options),
   };
 }
+
+/*
+  To add a deprecation, you must add a new entry to the `DEPRECATIONS` object.
+  The entry should be an object with the following properties:
+
+  * `id` (required): A string that uniquely identifies the deprecation. This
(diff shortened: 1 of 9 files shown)


NullVoxPopuli (approved):

packages/@ember/-internals/deprecations/index.ts
```diff
@@ -0,0 +1,23 @@
...
+}
+
+function deprecation(options: DeprecationOptions) {
+  return {
+    options,
+    test: !isEnabled(options),
+    isEnabled: isEnabled(options),
+  };
+}
+export const DEPRECATIONS = {
```
NullVoxPopuli, line 15: this might be tricky when it comes to me generating the list of things to deprecate for 'ember' -- maybe I'll codemod it.

kategengler: Can you elaborate?

kategengler: It should be the same list you needed to generate previously, but now will all be referenced from one location. The message can still vary by location used (and is expected to do so).

PR #20669, opened 2024-03-26
[FEATURE] Make deprecations throw when the `until` for `ember-source` has passed

kategengler:
- Add a new `deprecateUntil` that checks whether a deprecation has been removed and throws if it has. This explicitly uses `throw` and not assert so that it will remain in production builds. This function should be used internally in ember-source in place of `deprecate`.
- Add utilities `testIf` and `testUnless` to guard tests based on whether a deprecation has been removed. Uses this for the existing deprecation on @ember/routing
- Add a query param to tests `overridedeprecationversion` that takes a version to be the current version for removing deprecation purposes
- Add a CI build to use the above to test with deprecations broken

For https://rfcs.emberjs.com/id/0830-evolving-embers-major-version-process

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,6 +102,29 @@ jobs:
           ALL_DEPRECATIONS_ENABLED: true
         run: pnpm test
 
+  deprecations-broken-test:
+    name: Debug and Prebuilt (All Tests by Package + Canary Features) with Deprecations as Errors
+    runs-on: ubuntu-latest
+    needs: [ basic-test, lint, types ]
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/setup
+      - name: build
+        env:
+          DISABLE_SOURCE_MAPS: true
+          BROCCOLI_ENV: production
+        run: pnpm ember build
+      - name: Upload build
+        uses: actions/upload-artifact@v3
+        with:
+          name: dist
+          path: dist
+      - name: test
+        env:
+          TEST_SUITE: each-package
+          OVERRIDE_DEPRECATION_VERSION: '15.0.0' # Throws on all deprecations with an until before or equal to this version
+        run: pnpm test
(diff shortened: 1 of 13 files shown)


ef4 (approved):

chancancode:
@kategengler @ef4 I have a question: does this still get stripped correctly in prod build? looking at just the code here, it's not obvious it would – but maybe the code that handles that is out-of-tree somewhere else?

kategengler:
@chancancode No, it does not ...  the `deprecate` within `deprecateUntil` should be stripped but the `throw` should remain because we want to actually break the APIs at the `until`

shama:
After upgrading to `ember-source@5.9.0` I now see hundreds of deprecation messages in the terminal when running `ember build/serve`. Is this expected? Is there a way to silence or figure out where these deprecations are originating from?

```
DEPRECATION: The parts property on path nodes is deprecated, use head and tail instead
DEPRECATION: The 'Program' visitor node is deprecated. Use 'Template' or 'Block' instead (node was 'Block') 
DEPRECATION: The parts property on path nodes is deprecated, use head and tail instead
DEPRECATION: The parts property on path nodes is deprecated, use head and tail instead
DEPRECATION: The this property on path nodes is deprecated, use head.type instead
DEPRECATION: The parts property on path nodes is deprecated, use head and tail instead
```

kategengler:
@shama This is not related to this PR -- this PR makes deprecations throw when the until is passed, so when ember-source 6.0 is released, deprecations with an until of 6.0 will throw immediately so that we can take our time garbage collecting the old code. 

The deprecations you are referencing are not from ember-source (they print on the command line). These deprecations came from `@embroider/macros` and were handed in this PR https://github.com/embroider-build/embroider/pull/1967 and released in v1.16.3 of that package.

shama:
@kategengler My bad, thank you for pointing me to that embroider issue!

PR #20673, opened 2024-03-29
Add some additional coverage for strict mode/internal keywords

chancancode:
Ember currently uses AST plugins to implement certain features that glimmer-vm does not natively provide, such as `{{#each-in}}`, `{{outlet}}`, `{{mount}}` and some features in `{{#in-element}}`. These rewrites the AST and insert private keywords e.g. `{{#each (-each-in)}}`. These tests ensures we have _some_ basic coverage for those features in strict mode.

Ultimately, our test coverage for strict mode is quite inadequate. This is particularly important as we expect more apps to start adopting the feature. Ideally we would run our entire/most of our test suite against both strict and resolution modes, and these things would be implicitly covered elsewhere, but until then, these coverage are essential.

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/components/strict-mode-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/components/strict-mode-test.js
@@ -258,6 +258,48 @@ moduleFor(
       this.render('<Foo/>');
       this.click('button');
     }
+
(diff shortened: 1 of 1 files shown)


NullVoxPopuli (approved):

chancancode:
See also https://github.com/emberjs/ember.js/pull/20658/files#r1544930585

PR #20675, opened 2024-04-02
Build reform

ef4:
1. This makes Ember's test suite build with Vite, with no AMD loader present. Which ensures that Ember is strictly following the ES module spec.

2. It also makes Ember's prepublication build use Rollup instead of a custom broccoli pipeline. Again, because we want the ES module graph driving the build. This step is Rollup instead of Vite because there's enough backward-compatibility weirdness that Vite's more opinionated packaging up of Rollup features was not helpful.

TODOs:
 - [x] address dependency on decorator-transforms (which is breaking for browser support unless we're careful to only use it in prebuild)
 - [x] do size comparison on prod bundles (to ensure we properly remove `ember-testing`, etc)
 - [x] do a size comparison on an app's required total node_modules
 - [x] ~expand the smoke-tests so they exercise the non-prebuilt legacy bundle code path~ (this is not a thing now, we've returned to shipping ember.prod.js for the legacy bundles so we can avoid shipping our entire legacy bundle build into apps)
 - [x] expand the smoke-tests so they exercise the embroider strictEmberSource code path
 - [x] before merging do a final comparison of the defines in the bundles, before and after the pr
 - [x] replace the PREBUILT variant (which is currently being ignored). We could reimplement prebuilt, or we could expand the end-to-end app tests since they would more naturally cover that case.
 - [x] remove vestigial broccoli build code and audit for unused dependencies
 - [x] update to stable releases of babel-plugin-debug-macros and decorator-transforms
 - [x] restore compatibility with old ember-cli-htmlbars versions


ef4:
Notes to self on next steps:

 - the prebuilt ember.debug.js and ember-testing.js and ember-template-compiler.js are all traditionally built with hard-coded DEBUG=true and assertions not stripped. We can keep that behavior easily enough.
 - in the ESM output, we should implement the DEBUG and assertion stripping in terms of embroider/macros in preparation for being a real v2 addon. Plus, that will work correctly after the build has been reduced by rollup at prepublication time, since embroider/macros is wholly external, whereas `@ember/debug` is something rollup is itself building and trying to optimize away.
 - treeForVendor will need to manually run the embroider macros. We can't move ember's code into treeForAddon until we have a strict-es-modules flag to flip (too much stuff expects them to exist early in vendor.js).

ef4:
I think "All Tests" jobs are taking forever because filtering down suites by URL isn't working, so we end up running the whole suite many times.

packages/@ember/-internals/glimmer/tests/integration/components/curly-components-test.js
```diff
@@ -2537,7 +2537,7 @@ moduleFor(
         template: '<div id="inner-value">{{this.wrapper.content}}</div>',
       });
 
-      let expectedBacktrackingMessage = backtrackingMessageFor('content', 'Wrapper', {
+      let expectedBacktrackingMessage = backtrackingMessageFor('content', '_?Wrapper', {
```
ef4, line 2540: This is interpreted as a regex, and when testing production builds under rollup we see the class name `_Wrapper`.

tests/node/template-compiler-test.js
```diff
@@ -28,33 +28,5 @@ QUnit.module('ember-template-compiler.js', function () {
         'compile function is present'
       );
     });
-
-    QUnit.test('can access _Ember.ENV (private API used by ember-cli-htmlbars)', function (assert) {
```
ef4, line 32: This private API stopped being used in ember-cli-htmlbars 6.1.

ef4:
Confirmed in a real app's production build that this reduces vendor.js by 48kB. Most of that is probably from using decorator-transforms internally, so that we're not forced to transpile all other class features.

ef4:
I audited the differences caused by this PR in the AMD loader contents of a real app in both dev and prod.

The only change is that `ember-babel` disappears, which is fine.

index.html
```diff
@@ -0,0 +1,77 @@
...
+
+      // Handle testing feature flags
+      if (QUnit.urlParams.ENABLE_OPTIONAL_FEATURES) {
+        EmberENV.ENABLE_OPTIONAL_FEATURES = true;
+      }
+
+      EmberENV['RAISE_ON_DEPRECATION'] = true;
+
+      if (QUnit.urlParams.ALL_DEPRECATIONS_ENABLED) {
+          EmberENV['_ALL_DEPRECATIONS_ENABLED'] = QUnit.urlParams.ALL_DEPRECATIONS_ENABLED;
```
kategengler, line 38:
```suggestion
          EmberENV['_ALL_DEPRECATIONS_ENABLED'] = !!QUnit.urlParams.ALL_DEPRECATIONS_ENABLED;
```
nearly got through again

kategengler (approved):

index.html
ef4: Oof, thanks. This was missed because the other PR fixing it isn't merged yet. If it was I would have seen the conflict.

kategengler: ah yeah, if it got caught as a move -- I need to follow up on that

Issue #19861, opened 2021-12-01
[Bug] Unable to prevent default <LinkTo> behavior without blocking all event handlers


zzzzBov:
### 🐞 Describe the Bug

When using the `<LinkTo>` component, it's occasionally helpful to override the default link behavior, such as by opening a modal instead of navigating to a page of the same content.

Ideally this would be handled via a click handler using `event.preventDefault()`, but unfortunately [LinkTo doesn't check `event.defaultPrevented` before performing its default click behavior][1].

This leads to some complications because calling `stopPropagation`/`stopImmediatePropagation` prevent _all_ event handlers from executing, which can be problematic if other handlers&mdash;such as analytics tracking&mdash;are also bound to the element or any of its ancestors.

### 🔬 Minimal Reproduction

https://ember-twiddle.com/c5143970c9a673adff023dc106faedcc is an example that shows two links that have click handlers.

### 😕 Actual Behavior

The first uses `preventDefault` which is taken to the `/example` route, but still allows the parent event handler to execute.
The second uses `stopPropagation` which does not navigate to `/example`, but also doesn't allow the parent event handler to execute.

### 🤔 Expected Behavior

In the case of the first link, I'd expect clicking on the link to call both handlers and _not_ navigate to the `/example` route.

### 🌍 Environment

- Ember: -
- Node.js/npm: -
- OS: -
- Browser: -

### ➕ Additional Context

[1]: https://github.com/emberjs/ember.js/blob/c37df4dadf49920cc70a6a1811dd9fb7a955a6e4/packages/%40ember/-internals/glimmer/lib/components/link-to.ts#L136-L183

MelSumner:
Should Ember support this? Opening a modal should be done with a button, not a link.

zzzzBov:
Navigating to a document is a link behavior. Not all modals are stand-alone documents, but in my very specific case it is, and I need to perform some non-default functionality for visitors that use the link directly.

What I don't want to have to do is build a custom component to render a link to a route when all that's needed is an extra check in LinkTo:

```
if (!isSimpleClick(event) || event.defaultPrevented) {
  return;
}
```

Someone middle- or control-clicking on the link ought to be taken to the correct document, without the special handling (I'm aware my example code doesn't include the `isSimpleClick` check necessary for that behavior).

NullVoxPopuli:
Instead of using a click handler to open the Modal, what if you you use the router service and derived data to open the modal?

like:
```js
@service router;

[...]

PR #20680, opened 2024-04-15
[BUGFIX beta] honour preventDefault() in LinkTo

deanylev:
There was some discussion around this in https://github.com/emberjs/ember.js/issues/19861 but to me the current behaviour definitely seems like more of a bug than a feature. 

If `<LinkTo>` is a replacement for `<a>` then all behaviour should be carried over, I don't think the component should be opinionated about conventional/appropriate uses for the tag.


kategengler:
Closing for the above reasons

PR #20681, opened 2024-04-19
Update ember-cli-babel

ef4:
ember-cli-babel 8 uses a peerDep on @babel/core, which is why this also adds @babel/core explicitly to dependencies.

This is motivated by my work in the build-reform branch, which wants to rely on the static block support added in ember-cli-babel 8.2.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "unlink:all": "node bin/unlink-all.mjs"
   },
   "dependencies": {
+    "@babel/core": "^7.24.4",
     "@babel/helper-module-imports": "^7.16.7",
     "@ember/edition-utils": "^1.2.0",
     "@glimmer/compiler": "0.92.0",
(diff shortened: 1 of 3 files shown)


NullVoxPopuli (approved):

package.json
```diff
@@ -56,6 +56,7 @@
     "unlink:all": "node bin/unlink-all.mjs"
   },
   "dependencies": {
+    "@babel/core": "^7.24.4",
```
NullVoxPopuli, line 59: Do we not need to forward the peer to consumers?

ef4: That's a choice. We can either have it in dependencies or in peerDependencies. In dependencies is more conservative and non-breaking.

ef4: (The path forward to getting rid of all this is when it's a v2 addon it won't try to bring any babel shenanigans at all.)

PR #20685, opened 2024-04-22
Remove support for non-glimmer template-only components

ef4:
The template-only-glimmer-components optional feature became mandatory at 4.0. This removes the implementation for the unreachable branch.


NullVoxPopuli (approved):

boris-petrov:
I don't understand. I updated Ember to 5.10, removed `template-only-glimmer-components` from `config/optional-features.json`, and I got:

```
The template-only-glimmer-components optional feature should be enabled under Octane, run `ember feature:enable template-only-glimmer-components` to enable it
```

Same for the `application-template-wrapper` setting. Am I supposed to leave these in the `optional-features.json` file even though they are not used?

ef4:
Yeah it looks like that assertion needs to get cleaned up. You can leave the setting in place for now to silence it. If you want to make a PR taking it out it's here: https://github.com/emberjs/ember.js/blob/0da8387a91d6c3ca0c81238e3a3775ce459cb298/lib/index.js#L95

PR #20687, opened 2024-04-23
rationalize test matrix

ef4:
This surfaces all variants of our test suite in one github actions matrix.

Previously, each job you'd see in ci.yml actually invoked some combination of other jobs that were governed by run-tests.js. Now all the possible test suite variants are handled in one flat list and we can choose the list of combinations we want to test, and see each of their results independently in github actions output.

I did not completely reproduce the old behavior. We *could*, but I think some of what it was testing was more by accident than intentional. The "prebuilt" variant for example is worth testing, but not under every possible other flag. It gets built with the exact same pipeline we use for the not-prebuilt stuff.

This change also eliminates the per-package filtering that was being managed by run-tests.js. For interactive use, we have perfectly adequate filtering via qunit's UI. For CI runs, the splitting into packages only to re-concatenate them in series was actively counter-productive.

### DIFF
--- a/bin/run-tests.js
+++ b/bin/run-tests.js
@@ -1,13 +1,43 @@
 /* eslint-disable no-console */
 'use strict';
 
+/* 
+  Test Variants
+
+  These are all accepted as environment variables when running `ember test` or
+  as query params when directly invoking the test suite in the browser. 
+*/
+const variants = [
+  // When true, even deprecations that are not yet at the "enabled" version will
+  // be enabled, so we can ensure that they and their tests will continue to
+  // function correctly when we hit the enabled version.
+  'ALL_DEPRECATIONS_ENABLED',
+
+  // This overrides the current version of ember for purposes of seeing how
+  // deprecations behave. We use it in CI to prove that after a deprecation has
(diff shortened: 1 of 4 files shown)


tests/index.html
```diff
@@ -36,27 +36,27 @@
 
 
         // Handle extending prototypes
-        EmberENV['EXTEND_PROTOTYPES'] = !!QUnit.urlParams.extendprototypes;
+        EmberENV['EXTEND_PROTOTYPES'] = !!QUnit.urlParams.EXTEND_PROTOTYPES;
```
ef4, line 39: The spelling differences here are because I saw no reason to manage a translation between environment variables and query params. They can all just look the same.

bin/run-tests.js
```diff
@@ -52,94 +82,13 @@ function runInBrowser(url, attempts) {
   return getBrowserRunner().run(url, attempts);
 }
 
-let testFunctions = [];
```
NullVoxPopuli, line 55: it's nice to see all this go!

NullVoxPopuli (approved):

PR #20693, opened 2024-05-06
Fix running tests with all deprecations enabled.

kategengler:
The refactor to put the QUnit parameter directly into the EmberENV instead of using process.env to set the QUnit parameter resulted in the property being a string instead of a boolean

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/application/debug-render-tree-test.ts
+++ b/packages/@ember/-internals/glimmer/tests/integration/application/debug-render-tree-test.ts
@@ -788,13 +788,7 @@ if (ENV._DEBUG_RENDER_TREE) {
         ]);
       }
 
-      async [`${testUnless(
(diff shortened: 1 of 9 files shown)


kategengler:
@ef4 Can you think of a good way to guard this with a test? It is the kind of thing that can break without noticing, easily.

ef4:
I should have hand-checked this case. Now that you've fixed it, I think it falls into the area of test infrastructure that it's more economical to trust and hand-check-when-you-touch-it than to continuously test.

But if we want to test it, I think it would require making a test that leaves detectable traces in the log output (like `console.log("I checked that all deprecations are enabled")`, adapting the test runner job to notice that output and make it available to subsequent jobs, and having a final job that looks at the output of all the matrix runs to assert that at least one produced the expected signature.

kategengler:
Yeah that's the only style I could come up with too -- I agree that we just need to keep an eye on it.

PR #20695, opened 2024-05-09
Fix tests when deprecations are enabled

NullVoxPopuli:
:crossed_fingers:

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/application/debug-render-tree-test.ts
+++ b/packages/@ember/-internals/glimmer/tests/integration/application/debug-render-tree-test.ts
@@ -788,13 +788,7 @@ if (ENV._DEBUG_RENDER_TREE) {
         ]);
       }
 
-      async [`${testUnless(
-        DEPRECATIONS.DEPRECATE_COMPONENT_TEMPLATE_RESOLVING.isRemoved
-      )} template-only components`]() {
-        expectDeprecation(
-          /resolved templates/,
-          DEPRECATIONS.DEPRECATE_COMPONENT_TEMPLATE_RESOLVING.isEnabled
-        );
+      async [`@test template-only components`]() {
         this.addTemplate(
           'application',
           strip`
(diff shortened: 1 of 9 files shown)


packages/@ember/-internals/glimmer/tests/integration/application/debug-render-tree-test.ts
```diff
@@ -788,13 +785,7 @@ if (ENV._DEBUG_RENDER_TREE) {
...
       }
 
-      async [`${testUnless(
-        DEPRECATIONS.DEPRECATE_COMPONENT_TEMPLATE_RESOLVING.isRemoved
-      )} template-only components`]() {
-        expectDeprecation(
-          /resolved templates/,
-          DEPRECATIONS.DEPRECATE_COMPONENT_TEMPLATE_RESOLVING.isEnabled
-        );
+      async [`@test template-only components`]() {
```
kategengler, line 788: This is weird, no? What made you put that there in the first place?

packages/@ember/-internals/glimmer/tests/integration/components/component-template-test.js
```diff
@@ -31,8 +25,6 @@ moduleFor(
     [`${testUnless(
       DEPRECATIONS.DEPRECATE_COMPONENT_TEMPLATE_RESOLVING.isRemoved
     )} it takes precedence over resolver`]() {
```
kategengler, line 27: If this test doesn't trigger the deprecation, why would it be removed when the deprecation is removed?

NullVoxPopuli: because it's a scenario we don't need to care about.

Currently, we're testing if `setComponentTemplate` is used and no resolution happens.

But resolution can _never_ happen once the deprecated is removed, so the whole test becomes invalid.

packages/@ember/-internals/glimmer/tests/integration/application/debug-render-tree-test.ts
NullVoxPopuli: Probably a couple things:
- I originally was not updating these tests to use setComponenttemplate
- I changed (per Ed's suggestion) the timing of when the deprecation is thrown to only occur if there is a _found_ resolved template, rather than anytime `layoutFor` is called (this ended up simplifying the implementation in resolver.ts)

kategengler: So should this test remain after the deprecation is removed then?

PR #20700, opened 2024-05-22
Update deprecation links for `template-action` & `component-template-resolving`

mkszepp:
The deprecation pages are now online, so we can update the urls in deprecation.

- https://deprecations.emberjs.com/id/deprecate-template-action
- https://deprecations.emberjs.com/id/deprecate-component-template-resolving

### DIFF
--- a/packages/@ember/-internals/deprecations/index.ts
+++ b/packages/@ember/-internals/deprecations/index.ts
@@ -111,7 +111,7 @@ export const DEPRECATIONS = {
   }),
   DEPRECATE_TEMPLATE_ACTION: deprecation({
     id: 'template-action',
-    url: 'https://deprecations.emberjs.com/v5.x/#toc_deprecate-template-action',
+    url: 'https://deprecations.emberjs.com/id/deprecate-template-action',
     until: '6.0.0',
     for: 'ember-source',
     since: {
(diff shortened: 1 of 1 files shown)


packages/@ember/-internals/deprecations/index.ts
```diff
@@ -111,7 +111,7 @@ export const DEPRECATIONS = {
   }),
   DEPRECATE_TEMPLATE_ACTION: deprecation({
     id: 'template-action',
-    url: 'https://deprecations.emberjs.com/id/template-action',
```
NullVoxPopuli, line 114: These id urls were supposed to work. Hmm

Ah, the url should be: https://deprecations.emberjs.com/id/deprecate-template-action/

packages/@ember/-internals/deprecations/index.ts
```diff
@@ -121,7 +121,7 @@ export const DEPRECATIONS = {
   }),
   DEPRECATE_COMPONENT_TEMPLATE_RESOLVING: deprecation({
     id: 'component-template-resolving',
-    url: 'https://deprecations.emberjs.com/id/component-template-resolving',
+    url: 'https://deprecations.emberjs.com/v5.x/#toc_deprecate-component-template-resolving',
```
NullVoxPopuli, line 124: Should be: https://deprecations.emberjs.com/v5.x/#toc_deprecate-component-template-resolving

NullVoxPopuli:
Thanks for submitting a pr to fix the links!

There are id links for each deprecation, which is important for not dealing with large pages, and losing scroll, etc.

I've left a comment with the actual urls for each - if you could update those, that'd be fantastic!

packages/@ember/-internals/deprecations/index.ts
mkszepp: ah cool... havn't known that... now updated

NullVoxPopuli (approved):

kategengler:
I actually think this was a mistake when we merged into the deprecation-app -- the files added there should have been named without `deprecate` in the filename to match the ids of the deprecations. Am I correct @mansona ?

PR #20702, opened 2024-06-03
Deprecate array prototype extensions

simonihmig:
As per [RFC848](https://rfcs.emberjs.com/id/0848-deprecate-array-prototype-extensions).

This was a bit trickier than I actually expected, here is what I did and failed to do:
* The RFC says the deprecation should trigger when `EmberENV.EXTEND_PROTOTYPES` is not false. But given that there is some data munging happening [here](https://github.com/emberjs/ember.js/blob/f2940aaf714e9b17b143b6426a91d4fb65c5f70f/packages/%40ember/-internals/environment/lib/env.ts#L205-L211) that is turning `EXTEND_PROTOTYPES: false` to become `EXTEND_PROTOTYPES: { Array: false }`, which I was not eager to change, the deprecation is skipped for both of these. I think that is ok?
* I tried to add the deprecation to the [env.ts](https://github.com/emberjs/ember.js/blob/f2940aaf714e9b17b143b6426a91d4fb65c5f70f/packages/%40ember/-internals/environment/lib/env.ts#L205) file where the `EXTEND_PROTOTYPES` config is evaluated (1st commit)
* this didn't work, as it was causing a circular module dependency between that module and `deprecations/index.ts`, so I moved the deprecation call to the latter. Not ideal, but I had no better idea. (2nd commit)
* I have no test that is asserting for the deprecation to be issued, because with that implementation the deprecation is triggered when the module is loaded, so way before any test could call `expectDeprecation`. Any idea how to add a test here?
* Given that I was not able to silence the deprecation by a `expectDeprecation`, I added `RAISE_ON_DEPRECATION=false` to the test matrix scenarios that have `EXTEND_PROTOTYPES` to make them be able to actually run. (3rd commit)

### DIFF
--- a/packages/@ember/-internals/deprecations/index.ts
+++ b/packages/@ember/-internals/deprecations/index.ts
@@ -129,6 +129,16 @@ export const DEPRECATIONS = {
       enabled: '5.10.0',
     },
   }),
+  DEPRECATE_ARRAY_PROTOTYPE_EXTENSIONS: deprecation({
+    id: 'deprecate-array-prototype-extensions',
+    url: 'https://deprecations.emberjs.com/id/deprecate-deprecate-array-prototype-extensions',
+    until: '6.0.0',
+    for: 'ember-source',
+    since: {
+      available: '5.10.0',
+      enabled: '5.10.0',
+    },
+  }),
 };
 
 export function deprecateUntil(message: string, deprecation: DeprecationObject) {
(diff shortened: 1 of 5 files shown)


packages/@ember/-internals/deprecations/index.ts
```diff
@@ -129,6 +129,16 @@ export const DEPRECATIONS = {
       enabled: '5.10.0',
     },
   }),
+  DEPRECATE_ARRAY_PROTOTYPE_EXTENSIONS: deprecation({
+    id: 'deprecate-array-prototype-extensions',
+    url: 'https://deprecations.emberjs.com/id/deprecate-deprecate-array-prototype-extensions',
```
simonihmig, line 134: This points to https://github.com/ember-learn/deprecation-app/pull/1192, which we also need to get merged.

ef4:
Thanks @simonihmig. 

We should backport this beta.

PR #20704, opened 2024-06-05
Remove `Component` suffix from blueprints

bertdeblock:
This is part of the `First-Class Component Templates` RFC:
https://github.com/emberjs/rfcs/blob/master/text/0779-first-class-component-templates.md#blueprints

> We should also update the name of the class generated for a component class...


NullVoxPopuli (approved):

IgnaceMaes (approved):

kategengler:
Thank you!

PR #20712, opened 2024-07-04
feat: support custom transitioning classes for LinkTo

Techn1x:
It has long been an issue that async model hooks with loading substates causes ember applications to feel sluggish for transitions. The LinkTo doesn't change until the full transition is complete

See this issue from 2014 https://github.com/emberjs/ember.js/issues/4908

For my applications, for many years now, I've avoided them wherever possible, opting to use ember-concurrency tasks or reactiveweb tracked functions instead. Instant transitions & allows for much greater control over loading state etc.

I came up against a situation where I needed to use an async model hook for a kinda slow request, and wasn't happy about how slow it made the page feel, particularly due to the LinkTo. I ended up delving into LinkTo source code and found out about `ember-transitioning-in` and `ember-transitioning-out` - if only I had known years ago!

I think it's important that more users know about these classes, and that that's why their pages feel slow when they use async model hooks. By making `transitioningInClass` and `transitioningOutClass` customizable, this should help with visibility a lot, ideally accompanying an API reference update as well.

---

Alternatively, as mentioned in that issue, should the LinkTo active class be updated to be eager? This is probably a better solution, but I understand that might have wider impacts. The change proposed in this PR shouldn't cause any breaking changes.

---

The PR I linked has a good gif demonstrating the "slow" LinkTo update, but here's a new demonstration video that shows it as well. Note how long it takes the "Reports" side nav item to update to active/blue after I click it

https://github.com/emberjs/ember.js/assets/1049837/ed8b02e6-4249-4968-b355-1c49323641dd

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/components/link-to.ts
+++ b/packages/@ember/-internals/glimmer/lib/components/link-to.ts
@@ -511,14 +511,18 @@ class _LinkTo extends InternalComponent {
       isMissing(className) || typeof className === 'string' || typeof className === 'boolean'
     );
 
+    let defaultClassNames = {
+      active: 'active',
+      loading: 'loading',
+      disabled: 'disabled',
+      transitioningIn: 'ember-transitioning-in',
+      transitioningOut: 'ember-transitioning-out',
+    } as const
+
     if (className === true || isMissing(className)) {
-      return ` ${state}`;
+      return ` ${defaultClassNames[state]}`;
     } else if (className) {
       return ` ${className}`;
-    } else if (className === 'transitioningIn') {
-      return ' ember-transitioning-in';
-    } else if (className === 'transitioningOut') {
-      return ' ember-transitioning-out';
     } else {
       return '';
     }
--- a/packages/@ember/-internals/glimmer/lib/components/link-to.ts
+++ b/packages/@ember/-internals/glimmer/lib/components/link-to.ts
@@ -312,10 +312,10 @@ class _LinkTo extends InternalComponent {
       classes += this.classFor('active');
 
       if (this.willBeActive === false) {
-        classes += ' ember-transitioning-out';
+        classes += this.classFor('transitioningOut');
       }
     } else if (this.willBeActive) {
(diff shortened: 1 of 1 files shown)


Techn1x:
Knowing about the transitioning classes is the way to solve the issue, here's a tailwind example of how you can make it look eager to update
```ts
<LinkTo class="bg-white [&.active]:bg-blue-325 [&.ember-transitioning-in]:bg-blue-325 [&.active][&.ember-transitioning-out]:bg-white"
```

This PR just tries to make the transitioning classes visible & public

kategengler:
Since this changes public API, it would require an [RFC](https://github.com/emberjs/rfcs/). 

However, I believe alterations to `LinkTo` are unlikely -- the future is likely to be plain anchors as discussed in this older RFC https://rfcs.emberjs.com/id/0391-router-helpers 

This RFC hasn't yet been implemented but there are a number of addons to accomplish the same (ex https://github.com/rwjblue/ember-router-helpers)

Techn1x:
Thanks, no worries, I expected as much. I created this PR to raise awareness both for users and for maintainers about this long-standing issue. I won't be going through the RFC process for this.

I appreciate the mention of the route helpers addon, I might use it but from the outside its not really clear its for an RFC and I'm reluctant to use it if the behaviour will change. I'm also reluctant due to the sheer amount of issues I've had with transitions and queryParams and active state in general even while using officially sanctioned code. This is not a dig, I know fixes & changes are coming, just my reasoning.

Based on the ember guides, users are expected to have slow-feeling applications. If the ember guides mention this & how to handle it I think that would be very positive. I am unsure what that "official" solution/guide to this would be, I don't think there is one without using these currently private classes.

NullVoxPopuli:
@Techn1x do you think it'd be possible to implement your behavior here:
https://ember-primitives.pages.dev/4-routing/proper-links.md

I bet this could be done for _all links_ without using LinkTo, (native `<a>`, etc) via document.querySelectorAll(a[href*="transition.intent"]), etc

PR #20713, opened 2024-07-05
fix: add missing location property names

jaredgalanis:
- these missing property names mean that when the api docs are built the releasing member of the learning team needs to go in and manually fill in the property names in the json that is outputted.
- For example: https://github.com/ember-learn/ember-api-docs-data/pull/29/commits/13542f25575e107972f7ce0eca4207d5acce018d


kategengler:
Thank you!

PR #20720, opened 2024-07-12
[BUGFIX release] Fix deprecation guide url

ef4:
This deprecation content is not yet landed but it's about to land, and the url here is wrong.


simonihmig:
Oh no 🙈 ... thank you!

PR #20722, opened 2024-07-13
Remove checks for no longer existing optional features

boris-petrov:
Fixes the issue described in [this comment](https://github.com/emberjs/ember.js/pull/20685#issuecomment-2215326919).

cc @ef4

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -78,24 +78,6 @@ module.exports = {
         );
       }
 
-      if (
-        optionalFeaturesMissing ||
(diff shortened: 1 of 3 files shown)


NullVoxPopuli (approved):

boris-petrov:
I see also both are used in `smoke-tests/app-template/config/optional-features.json` - unsure if they have to still be there. Also, in `packages/@ember/component/template-only.ts` there is a comment about `template-only-glimmer-components` - perhaps it also must be changed.

anandfresh (approved):

Issue #20709, opened 2024-06-20
circular import between @ember/object/index and @ember/object/observable


patricklx:
### 🐞 Describe the Bug
there is a circular import between 
`'@ember/object/index';`
and
`'@ember/object/observable';`
and 
`@ember/array/index`.

those files are actually also importing from `@ember/-internals/metal`, so they could use that directly to access `get, set`

https://github.com/emberjs/ember.js/blob/main/packages/%40ember/object/observable.ts#L14
https://github.com/emberjs/ember.js/blob/main/packages/%40ember/object/index.ts#L13

this causes rollup to sometimes build it wrong and have runtime errors later. (testing with vite+dep optimizer)

kategengler:
cc @ef4

ef4:
Fixed in #20734

PR #20723, opened 2024-07-17
check circular imports

patricklx:
to fix issues like #20709

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -86,6 +86,17 @@ module.exports = {
       },
     },
     {
+      files: ['packages/**/*.[jt]s'],
+      rules: {
+        'import/no-cycle': [
+          'error',
+          {
+            ignoreExternal: true,
+          },
+        ],
+      },
+    },
+    {
       // TODO: files: ['packages/**/*.[jt]s'],
       files: ['packages/**/*.js'],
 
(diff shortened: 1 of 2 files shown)


boris-petrov:
My personal experience with `npx cycle-import-check` is that it also gives a warning for cycles in `type` imports (which I don't believe is a problem). The `eslint-plugin-import` that I mentioned in Discord works correctly in that case.

patricklx:
well, this isnt't working correctly anyway because it only considers relative imports. But we have a lot of import like `@ember/object`.
i didnt get eslint-plugin-import to work, always gets stack out of memory.
and not sure if it will have the same issue with those imports

boris-petrov:
I also have the following settings for the ESLint plugin in `.eslintrc.js` (v8):

```js
  settings: {
    'import/core-modules': ['moment'],
    'import/parsers': {
      '@typescript-eslint/parser': ['.ts'],
    },
    'import/resolver': {
      typescript: {
        alwaysTryTypes: true,
      },
    },
  }
```

Not sure if that helps in any way.

patricklx:
ok, i think i now found the correct setup

anandfresh (approved):

ef4:
These ideas got incorporated into #20734

patricklx:
> These ideas got incorporated into #20734

okay, but this could have the advantage that it shows the import cycles during linting and in the github action flow

Issue #20724, opened 2024-07-23
5.10: test waiters broken?


Turbo87:
### 🐞 Describe the Bug

After upgrading the crates.io codebase from 5.9 to 5.10 several tests started failing for no apparent reason. It looks like the test waiters are not working as intended and are causing the tests to be flaky. When I run the test suite locally the subset of failing tests is different from CI, which indicates that race conditions might be responsible.

### 🔬 Minimal Reproduction

see https://github.com/rust-lang/crates.io/pull/9035

### 😕 Actual Behavior

Previously succeeding tests are failing.

### 🤔 Expected Behavior

Previously succeeding tests should keep passing, if the test waiter system is used correctly.

### 🌍 Environment

- Ember: 5.10.1
- Ember-CLI: 5.10.0
- Node.js/npm: 20.15.1
- OS: Linux & MacOS
- Browser: Chrome & Firefox

NullVoxPopuli:
This matches what i've been seeing in open source repos, too

jelhan:
I noticed the same in one app. The `click` test helper resolves before a network request triggered by that click has settled. The network request is done with `fetch`. The project uses `ember-fetch@8.1.2`. `fetch` is imported from `fetch` package in the file doing the network request.

The tests are working as expected with `ember-source@5.9.0`. They are starting to fail if upgrading `ember-source` to `5.10.1`. I don't see any entry in the changelog for `5.10.0` and `5.10.1`, which may be related.

NullVoxPopuli:
Looks like it may have been the legacy waiters that broke: 
![image](https://github.com/user-attachments/assets/c0e050fb-cb61-4f4b-b62e-2a98fd5f3e9b)
 (this is from the `@ember/test-helpers` test suite)
 
 @jelhan do you have a minimal repro of the click resolving too early?

jelhan:
> @jelhan do you have a minimal repro of the click resolving too early?

Sadly not. I expect reproducing is not that easy. Only 2 out of many tests in the app are affected.

Aren't the failing tests in `@ember/ember-test-helpers` enough to reproduce and debug?

ef4:
#20726 should be a complete fix for this.

kategengler:
This should be fixed in v5.10.2 and in v5.11.0-beta.2

jelhan:
> This should be fixed in v5.10.2 and in v5.11.0-beta.2

I can confirm that v5.10.2 is working as expected in the app which was affected by the bug before.

PR #20726, opened 2024-07-30
[BUGFIX release] Remove unintentional duplication from ember-testing bundle

ef4:
Working on a fix for https://github.com/emberjs/ember.js/issues/20724

The ember-testing bundle is inlining copies of other ember packages that cause state to duplicate, which is most noticeable in breaking legacy test waiters.

- [x] Next step here is to sort out the AMD/ESM compat for the external references between the bundles.

### DIFF
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -18,13 +18,18 @@ const testDependencies = ['qunit', 'vite'];
 
 export default [
   esmConfig(),
-  legacyBundleConfig('./broccoli/amd-compat-entrypoints/ember.debug.js', 'ember.debug.js', true),
-  legacyBundleConfig('./broccoli/amd-compat-entrypoints/ember.debug.js', 'ember.prod.js', false),
-  legacyBundleConfig(
-    './broccoli/amd-compat-entrypoints/ember-testing.js',
-    'ember-testing.js',
-    true
-  ),
+  legacyBundleConfig('./broccoli/amd-compat-entrypoints/ember.debug.js', 'ember.debug.js', {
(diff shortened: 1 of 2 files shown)


NullVoxPopuli (approved):

rollup.config.mjs
```diff
@@ -419,19 +432,29 @@ function templateCompilerConfig() {
...
-      get FEATURES() { return require('@ember/canary-features').FEATURES },
-      get VERSION() { return require('ember/version').default },
+      __esModule: true,
+      default: {
+        get ENV() { return require('@ember/-internals/environment').ENV },
+        get FEATURES() { return require('@ember/canary-features').FEATURES },
+        get VERSION() { return require('ember/version').default },
+      },
+    }`,
+    '@ember/-internals/glimmer': `{
```
NullVoxPopuli, line 447: are these stubs that don't matter / aren't used in the actual output?

ef4:
@kategengler this is ready and should be "bugfix release".

rollup.config.mjs
ef4: They're used in the ember-template-compiler.js bundle. Which has always done wonky things so that it will share the main ember implementations of things if you load it into the browser, but bring its own copies of *some* of the things and stub the others if you load it by itself in node.

ef4: The node tests accurately reflect the things that older versions of ember-cli-htmlbars actually expect the ember-template-compiler.js bundle to do.

kategengler (approved):

PR #20727, opened 2024-07-30
Fix TS-Nightly

NullVoxPopuli:
There is not enough context for this file 🙈 

unless there is some hidden build step?
in any case, we're well passed v4 now, so there is no point in keeping that logic around.


kategengler:
lint failed

anandfresh (approved):

NullVoxPopuli:
Superseded by https://github.com/emberjs/ember.js/pull/20732

PR #20734, opened 2024-08-23
Removed circular dependencies

ef4:


patricklx:
You might also want https://github.com/emberjs/ember.js/pull/20723 for this:)

ef4:
Yea, good idea, we should prevent new cycles from creeping in. 

There is a spot in the rollup config where we could just fail the whole build if rollup sees a cycle.

ef4:
Did a review with @kategengler and I have a few more minor changes and then this is good to go.

ef4:
I investigated the one proposed minor change, which was whether we should better hide these newly created modules from outside consumers, but that would not be consistent with how these packages are already published. Only a few (like `@ember/-internals`) get rollup up before publication.

PR #20743, opened 2024-09-07
[BUGFIX release] Backporting circularity fixes

ef4:
This is backporting https://github.com/emberjs/ember.js/pull/20734/files to current release (5.x).


ef4:
@kategengler can we put this in 5.11.x patch?

My motivation here is that the circularity bugs get in the way of testing apps under vite, and it's nice to not have to deal with their 6.0 upgrades simultaneously.

kategengler:
I think we should backport to beta first. I can do it via the typical cherry-pick unless you think there are a lot of conflicts?

ef4:
Probably no conflicts.

PR #20746, opened 2024-09-12
[BUGFIX beta] Upgrade glimmer-vm for some perf improvements

NullVoxPopuli:
This should be backported to 5.12

------

The only thing in this release is the removal of meant-for-vm-only-debugging code.

Via: https://github.com/glimmerjs/glimmer-vm/pull/1606

image

https://ember-performance-testing-prod.pages.dev/report?benchmarks=%5B%22Render%20complex%20html%20(%40glimmer%2Fcomponent)%22%5D&emberVersions=%5B%22ember-5-10-local-glimmer%22%2C%22ember-5-10%22%2C%22ember-5-5%22%2C%22ember-5-4%22%2C%22ember-canary-local-glimmer%22%2C%22ember-canary%22%5D


rollup.config.mjs
```diff
@@ -195,6 +195,7 @@ export function exposedDependencies() {
       '@glimmer/owner',
       '@glimmer/opcode-compiler',
       '@glimmer/runtime',
+      '@glimmer/validator',
```
ef4, line 198: Since ember-source does directly use `@glimmer/validator`, this seems correct. But why wasn't it needed before now? 

I based this list on the classical behavior, which did not have `@glimmer/validator` on the list of top-level glimmer packages to include. It  got included in the build output because it's a dependency of some of these others. Did the other packages drop their dependency on it?

NullVoxPopuli: I also found this sus.

> Did the other packages drop their dependency on it?

that is my leading hypothesis, yea.

The whole package graph between the vm-provided packages changed quite a bit in the ember 5.5 to 5.6 bump, and then it changed again with this change as `@glimmer/debug` is now _much_ slimmer (due to removing a bunch of VM-specific debug code), it makes sense that the import/module graph would change -- the imports in debug are different --- but I don't know why that affected glimmer/validator as debug imports from the vm and _previously_ also the utils -- but _neither_ vm/index.js nor util/index.js import from validator.

I need a good tool to diff folders with regex filtering or something.

tangentially -- while this change to the VM is an improvement to prod builds, the work on speeding up the VM isn't done -- but I do think for _some_ perf-aware teams, it'll unblock the 5.5 -> 5.12 upgrade, while I / others keep poking around looking for perf opportunities (I already know of two more to look at, but I'm not sure how to make the tooling do the build time cleanup for me yet).

ef4 (approved):

PR #20751, opened 2024-09-13
Bring @glimmer/component into ember repo

ef4:
The glimmerjs/glimmer.js monorepo is littered with "standalone Glimmer.js" functionality that we don't use and that is not being maintained. But it also contains two packages that are critical to ember: `@glimmer/component` and `@glimmer/tracking`.

I want to bring both of those packages into here instead, while jettisoning their extra complexity that exists only to make them portable to the "standalone Glimmer.js" use case.

This PR brings over @glimmer/component, and simplifies it by:

 - dropping support for Ember < 3.13.
 - converting to a v2 addon

It also adds end-to-end test coverage that exercises @glimmer/component and @glimmer/tracking in both classic and embroider apps.

After we decide to land this, a followup will be to add release automation for `@glimmer/component`.

- [x] unify the API docs (there were already standalone files containing only docs that should now move alongside the implementation)
- [x] adjust ember-source's peerdep so this is non-breaking.


packages/@ember/-internals/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.0.0",
   "description": "Internal APIs shared by Ember packages. Warning: this package does not follow SemVer and is subject to break at any time!",
   "type": "module",
+  "private": true,
```
NullVoxPopuli, line 6: I verified that all the other package.jsons were also private.
Surprised this wasn't caught before -- but also not surprised --  because we're not using release-plan (or anything to enforce package.json consistency)

NullVoxPopuli: I think for now, we could very well set up release plan for just `@glimmer/component` if we wanted.

packages/@glimmer/component/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "node": ">= 18"
+  },
+  "typesVersions": {
+    "*": {
+      "*": [
+        "dist/*"
+      ]
+    }
+  },
+  "ember-addon": {
```
NullVoxPopuli, line 36: since glimmer/component doesn't depend on any embery things, does it need to be an addon at all?

packages/@glimmer/component/src/-private/component.ts
```diff
@@ -0,0 +1,286 @@
...
+export type ExpandSignature<T> = T extends any ? _ExpandSignature<T> : never;
+
+/**
+ * @internal we use this type for convenience internally; inference means users
+ *   should not normally need to name it
+ */
+export type Args<S> = ExpandSignature<S>['Args']['Named'];
+
+/**
+ * The `Component` class defines an encapsulated UI element that is rendered to
```
NullVoxPopuli, line 104: do these docs show up on the api docs?

packages/@glimmer/component/src/-private/component.ts
```diff
@@ -0,0 +1,286 @@
+import { DEBUG } from '@glimmer/env';
+
+const DESTROYING = new WeakMap<GlimmerComponent<object>, boolean>();
```
NullVoxPopuli, line 3: I know this is copied from the the old repo -- but I feel like we could use the destroyable primitives instead (which i'm more than happy to PR after this (#20751) lands)

```js
export {
  assertDestroyablesDestroyed,
  associateDestroyableChild,
  destroy,
  enableDestroyableTracking,
  isDestroying,
  isDestroyed,
} from '@glimmer/destroyable';

import { registerDestructor } from '@glimmer/destroyable';
```

https://github.com/glimmerjs/glimmer-vm/blob/main/packages/%40glimmer/destroyable/index.ts#L204

packages/@glimmer/component/src/-private/ember-component-manager.ts
```diff
@@ -0,0 +1,49 @@
...
+  capabilities = CAPABILITIES;
+
+  destroyComponent(component: GlimmerComponent): void {
+    if (component.isDestroying) {
+      return;
+    }
+
+    setDestroying(component);
+
+    schedule('actions', component, component.willDestroy);
```
NullVoxPopuli, line 40: where is this type of destruction scheduled in specific queues? does all destruction work like this?
I don't know that I've seen this outside of `@glimmer/component`

`destroy` in `@glimmer/destroyable` does have scheduling behavior as well :thinking:  https://github.com/glimmerjs/glimmer-vm/blob/main/packages/%40glimmer/destroyable/index.ts#L153 -- but it's different.

packages/@glimmer/component/package.json
ef4: I thought that too, but it imports from `@ember/destroyable`, `@ember/component`, and `@ember/runloop`. Those only work in an addon.

packages/@glimmer/component/src/-private/component.ts
ef4: Yeah that makes sense as a followup refactor.

packages/@glimmer/component/src/-private/component.ts
ef4: Good question. 

I see there are actually comment-only files that already existed to hold the `@glimmer/component` docs in this repo. Those can be merged with the real files now.

kategengler:
Even if we add release-plan just for @glimmer/component, we need to consider ember.js publishing. Right now, ember.js publishing runs on every tag. Changelogs are generated with bin/changelog.js and depend upon all PRs being just for that package (and follows references in commits via `-x` for cherry-picks to find originating PRs). Tags in the repo are just for ember.js (and named with just the version). There are a handful of things out there expecting to parse these. 

I think it can be done but there's a lot to update.

ef4:
I think the smallest next action to move this ahead would be to decide on a tagging strategy that would allowed tagged releases for `@glimmer/component` that don't interfere with tagged releases for ember-source.

I created https://github.com/emberjs/ember.js/pull/20753 as one possibility.

ef4:
When we bump ember-source in the app blueprint to the version containing this change, it would also be a good time to bump `@glimmer/component` to `^2.0.0`. Prior to that point, 2.0.0 causes peer dep warnings.

SergeAstapov:
minor - looks we missed to bring `readme.md` file so it does not get it's way to npm.

I guess we could just copy-paste the existing one?

e.g. 
https://www.npmjs.com/package/@glimmer/component/v/2.0.0?activeTab=readme
https://www.npmjs.com/package/@glimmer/component/v/1.1.2?activeTab=readme

PR #20753, opened 2024-09-20
Update tagging strategy

ef4:
This is preparation to make it possible to have more than one published package living in this monorepo.

It changes the tagging strategy so that releases of ember-source should be tagged like `v6.0.0-ember-source` instead of `v6.0.0`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,7 +126,7 @@ jobs:
         run: pnpm vite build --mode=development
 
       - name: Set BrowserStack Local Identifier
-        if: startsWith(github.ref, 'refs/tags/v')
+        if: startsWith(github.ref, 'refs/tags/v') && endsWith(github.ref, '-ember-source')
         run: |
           BROWSERSTACK_LOCAL_IDENTIFIER="$GITHUB_RUN_ID-$GITHUB_RUN_ATTEMPT"
           echo "BROWSERSTACK_LOCAL_IDENTIFIER=$BROWSERSTACK_LOCAL_IDENTIFIER" >> $GITHUB_ENV
@@ -210,7 +210,7 @@ jobs:
         blueprint-test,
         browser-test,
       ]
-    if: startsWith(github.ref, 'refs/tags/v')
+    if: startsWith(github.ref, 'refs/tags/v') && endsWith(github.ref, '-ember-source')
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,148 +1,241 @@
 # Ember.js Releases
 
-## Concepts
+See [Ember.js Releases](https://emberjs.com/releases/) for an overview of how Ember.js releases work.
 
-* 6-week release cycle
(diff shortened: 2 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -126,7 +126,7 @@ jobs:
         run: pnpm vite build --mode=development
 
       - name: Set BrowserStack Local Identifier
-        if: startsWith(github.ref, 'refs/tags/v')
+        if: startsWith(github.ref, 'refs/tags/v') && endsWith(github.ref, '-ember-source')
```
kategengler, line 129: Apparently you still cannot comment on unchanged lines on github:

Do we want to run this workflow at all for changes and or tags of @glimmer/component?

ef4: Yes. The CI workflow is good test coverage for everything together, including @glimmer/component. 

With #20751 we'll actually get better integration test coverage of @glimmer/component than we've ever had.

kategengler:
I PR'd with a more comprehensive update to RELEASE.md https://github.com/emberjs/ember.js/pull/20754

The only other concern I have is that PRs for @glimmer/component will show up in the changelog via script for `ember.js` and need to be edited out, but I don't expect it'll be high-volume, so that should be fine for now.

kategengler (approved):

RELEASE.md
```diff
@@ -1,148 +1,241 @@
...
-1. `git push origin beta`
+See [Ember.js Releases](https://emberjs.com/releases/) for an overview of how Ember.js releases work.
+
+Ember.js consists of many packages. A release of `Ember.js` is considered complete upon the release blog post on [the Ember blog](https://blog.emberjs.com/tag/release).
+
+This document is intended for maintainers of Ember.js. It describes the process for creating releases.
+
+## ember-source
+
+`ember-source` follows the Ember.js [release train/cycle](https://blog.emberjs.com/new-ember-release-process/). There is a stable release roughly every 6 weeks (every 4 versions excluding `x.0`) and a major release after every `x.12` release. New deprecations targeting the next major are not permitted to land after the `x.10` release. All public API changes should have an RFC and all features should behind feature flags. Features should not be enabled by default until the RFC has had a successful the [Ready for Release](https://github.com/emberjs/rfcs#ready-for-release) Stage.
```
kategengler, line 11:
```suggestion
`ember-source` follows the Ember.js [release train/cycle](https://blog.emberjs.com/new-ember-release-process/). There is a stable release roughly every 6 weeks (every 4 versions excluding `x.0` is an LTS candidate) and a major release after every `x.12` release. New deprecations targeting the next major are not permitted to land after the `x.10` release. All public API changes should have an RFC and all features should behind feature flags. Features should not be enabled by default until the RFC has had a successful the [Ready for Release](https://github.com/emberjs/rfcs#ready-for-release) Stage.
```

.github/workflows/ci.yml
kategengler: Looking at this again, I think this is wrong -- the BrowserStack Local Identifier needs to be set on any tag, not just ember-source tags, if we're still running the BrowserStack jobs

PR #20761, opened 2024-09-30
Prepare to implement tracked storage, RFC#669

NullVoxPopuli:
Motivation here is that I want to stop folks from using https://github.com/ember-polyfills/ember-tracked-storage-polyfill (and swap its implementation to use the real implementation (here, hopefully)

----------------

This doesn't make sense to continue work on until 
- https://github.com/emberjs/ember.js/pull/20751

lands -- because I don't really want to continue the fake-glimmer packages we have in `@ember/-internals/metal`

Additionally, this RFC, https://rfcs.emberjs.com/id/0669-tracked-storage-primitive#re-implementing-tracked-with-storage, kinda just likes like Starbeam concepts, so it should be super easy:
```js
export function trackedStorage(...) {
  return Cell(...);
}

export function getValue(cell) {
  return cell.current;
}

export function setValue(cell, value) {
  return cell.set(value);
}
```
ez

### DIFF
--- a/packages/@glimmer/tracking/primitives/storage.ts
+++ b/packages/@glimmer/tracking/primitives/storage.ts
@@ -1,0 +1,32 @@
+/**
+ * https://rfcs.emberjs.com/id/0669-tracked-storage-primitive
+ *
+ * See, in particular:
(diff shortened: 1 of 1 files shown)


kategengler:
Can we close this?

NullVoxPopuli:
yup

PR #20768, opened 2024-10-02
Allow GJS/GTS route templates

ef4:
In the process of starting to write guides content for template tag, @kategengler and I decided it's too clunky to force new learners to mix HBS and GJS.

In practice, experienced ember devs who are adopting template tag are often using [ember-route-template](https://github.com/discourse/ember-route-template) so they can author routes in template tag. We considered just proposing ember-route-template for the default blueprint, but then realized that in that case it's actually just much better to build the same feature into ember.

AFAIK, this PR is the full implementation and it's ready to go, pending an RFC.

This does not preclude also doing a more comprehensive Route Manager API that would cleanup more about routes. But this seems like an important smallest step we can make today to improve the experience.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -26,6 +26,12 @@ module.exports = {
 
     'disable-features/disable-async-await': 'error',
     'disable-features/disable-generator-functions': 'error',
+    'import/no-unresolved': [
+      'error',
+      {
+        ignore: ['@ember/template-compiler'],
+      },
+    ],
   },
 
   settings: {
--- a/packages/@ember/-internals/glimmer/lib/component-managers/outlet.ts
+++ b/packages/@ember/-internals/glimmer/lib/component-managers/outlet.ts
@@ -46,6 +46,7 @@ export interface OutletDefinitionState {
   template: Template;
   controller: unknown;
   model: unknown;
+  component: object | undefined;
 }
 
 const CAPABILITIES: InternalComponentCapabilities = {
--- a/packages/@ember/-internals/glimmer/lib/syntax/outlet.ts
+++ b/packages/@ember/-internals/glimmer/lib/syntax/outlet.ts
@@ -89,6 +89,9 @@ export const outletHelper = internalHelper(
             return model;
           });
 
+          let componentRef = childRefFromParts(outletRef, ['render', 'component']);
+          named['component'] = componentRef;
+
           if (DEBUG) {
             named['model'] = createDebugAliasRef!('@model', named['model']);
           }
@@ -160,6 +163,7 @@ function stateFor(
     name: render.name,
     template,
     controller: render.controller,
+    component: render.component,
     model: render.model,
   };
 }
--- a/packages/@ember/-internals/glimmer/tests/integration/application/rendering-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/application/rendering-test.js
@@ -8,8 +8,6 @@ import { tracked } from '@ember/-internals/metal';
 import { set } from '@ember/object';
 import { backtrackingMessageFor } from '../../utils/debug-stack';
 import { runTask } from '../../../../../../internal-test-helpers/lib/run';
-
-// @ts-ignore build-time feature
 import { template } from '@ember/template-compiler';
 
 moduleFor(
--- a/packages/@ember/-internals/glimmer/tests/integration/application/rendering-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/application/rendering-test.js
@@ -9,6 +9,9 @@ import { set } from '@ember/object';
 import { backtrackingMessageFor } from '../../utils/debug-stack';
 import { runTask } from '../../../../../../internal-test-helpers/lib/run';
 
+// @ts-ignore build-time feature
+import { template } from '@ember/template-compiler';
+
 moduleFor(
   'Application test: rendering',
   class extends ApplicationTestCase {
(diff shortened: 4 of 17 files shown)


NullVoxPopuli:
I am a fan, and this reduces the divergences in [the TS support PR for the v2 app blueprint](https://github.com/embroider-build/app-blueprint/pull/80/files#diff-0967fa9032cebb2176cbcbb70b334310fa56c0bf3ae20aa84933219a29d88ead)

ef4:
cc @chancancode: this would essentially copy your ember-route-template implementation into ember.

chancancode:
Yea, I think it definitely should be built-in and the wrapper shouldn't be necessary. Mostly just have to decide whether we are happy enough with `@model` and `@controller` for now, and if those somehow locks us into an undesirable path. Couldn't think of any obvious problems off the top of my head, generally +1.

packages/@ember/routing/route.ts
```diff
@@ -1825,6 +1826,15 @@ export function getRenderState(route: Route): RenderState | undefined {
...
 }
 
+import { precompileTemplate } from '@ember/template-compilation';
+
+function RouteTemplate(Component: object): TemplateFactory {
+  return precompileTemplate(`<Component @model={{@model}} @controller={{this}} />`, {
+    strictMode: true,
+    scope: () => ({ Component }),
+  });
+}
```
chancancode, line 1836: This is.. fine.. I guess. I don't have time to dig deeper but it feels unnecessary. Somewhere not that much deeper (inside this repository I am pretty sure), we are interfacing with the low-level APIs in glimmer-vm that lets us render arbitrary template with arbitrary `this` arg and named arguments, so this extra indirection sticks out to me

chancancode: This may also not be fine, actually. It at least _looks_ like this could potentially leak and be slow. Making and returning a new template each time this is likely problematic, as each unique template needs to be "runtime compiled" which is, IIRC, not-GC-able, at least not until the owner is destroyed.

chancancode: Basically – `@model` is already a magic argument ember adds at this integration point. So we must already be very close to the place where we can inject any number of these extra arguments. We should just be able to add `@controller` the same way.

If we _must_ have a wrapper template, I think we should be able to do a single static template and pass `@component` as an argument, rather than building dynamically building a new one each time.

.eslintrc.js
```diff
@@ -26,6 +26,12 @@ module.exports = {
 
     'disable-features/disable-async-await': 'error',
     'disable-features/disable-generator-functions': 'error',
+    'import/no-unresolved': [
+      'error',
+      {
+        ignore: ['@ember/template-compiler'],
```
ef4, line 32: This is a build-time feature that we added in RFC 931 but haven't used internally in ember-source yet. The whole implementation of it lives in babel-plugin-ember-template-compilation.

packages/@ember/-internals/glimmer/lib/component-managers/outlet.ts
```diff
@@ -46,6 +46,7 @@ export interface OutletDefinitionState {
   template: Template;
   controller: unknown;
   model: unknown;
+  component: object | undefined;
```
ef4, line 49: The choice of type here reflects that a component is anything with a component manager, and having a component manager requires being a weakmap key.

packages/@ember/-internals/glimmer/lib/syntax/outlet.ts
```diff
@@ -89,6 +89,9 @@ export const outletHelper = internalHelper(
             return model;
           });
 
+          let componentRef = childRefFromParts(outletRef, ['render', 'component']);
+          named['component'] = componentRef;
```
ef4, line 93: Unlike `model` above, this argument doesn't change for the lifetime of this render state. Its lifetime is the same as `outletRef.render.template`, rather than `outletRef.render.model`.

packages/@ember/-internals/glimmer/tests/integration/application/rendering-test.js
```diff
@@ -553,5 +554,88 @@ moduleFor(
...
+        'I am the component'
+      );
+    }
+
+    async ['@test can switch between component-defined routes']() {
+      this.router.map(function () {
+        this.route('first');
+        this.route('second');
+      });
+      this.add('template:first', template('First'));
```
ef4, line 617: This `template` is the JS version of our `<template></template>`, so the value here is a component.

packages/@ember/routing/route.ts
```diff
@@ -1825,6 +1826,15 @@ export function getRenderState(route: Route): RenderState | undefined {
   return route[RENDER_STATE];
 }
 
+import { precompileTemplate } from '@ember/template-compilation';
+
+const RoutableComponent = precompileTemplate(
```
ef4, line 1831: I had to name it this for the LOLs.

packages/@ember/routing/route.ts
ef4: I implemented this suggestion so now the template is stable and the component is parameterized.

ef4:
RFC:  https://github.com/emberjs/rfcs/pull/1046

packages/@ember/-internals/glimmer/lib/syntax/outlet.ts
chancancode: If it doesn't change then something like `createConstRef(state.component, '@component')` should do it. It'll result in less bounds tracking for the DOM etc

packages/@ember/-internals/glimmer/lib/syntax/outlet.ts
```diff
@@ -89,6 +89,9 @@ export const outletHelper = internalHelper(
...
+
           if (DEBUG) {
             named['model'] = createDebugAliasRef!('@model', named['model']);
           }
```
chancancode, line 97: Also not really sure what's up with this, does this just predate `createComputeRef` taking an optional label argument? (In any case the positioning of the new code split these two related chunks)

kategengler:
This was replaced by https://github.com/emberjs/ember.js/pull/20800

PR #20769, opened 2024-10-03
Start the Runtime Template Compiler from RFC 931

NullVoxPopuli:
I keep running in to needs for the new compiler, so here we go:

https://github.com/emberjs/rfcs/pull/931/files?diff=unified&w=0

(in particular, trying to use vite with [ember-source forced as a v2 addon](https://github.com/NullVoxPopuli/limber/pull/1820/files#diff-670e31b1eceb4987f89c5f78d0989e31999296c29dbd3f7b57a49f69d0122a2eR24) )

Todo

- [ ] Tests
- [ ] Make `scope` support passing the instance (probably in the VM)


ef4:
Superseded in https://github.com/emberjs/ember.js/pull/20776

PR #20771, opened 2024-10-05
Add some TS component blueprint tests

Windvis:
While looking into #20511 I noticed there weren't any tests for the `--typescript` version of the blueprints.

This adds a some ts tests for the component blueprint. We can add tests for other blueprint in follow up PRs.

I included a pnpm patch for the `ember-cli/ember-cli-blueprint-test-helpers` package since it doesn't accept any extra cli arguments for `ember new` in its current state: https://github.com/ember-cli/ember-cli-blueprint-test-helpers/issues/391

This also fixes a bug where `ember g component foo.ts` would generate `.ts.hbs` files since .ts wasn't normalized yet.

An extra bonus is that CI now runs a bit faster since the blueprint tests run in parallel.

Partially solves: https://github.com/emberjs/ember.js/issues/20362


pnpm-lock.yaml
```diff
@@ -8,13 +8,18 @@ overrides:
...
+    path: patches/ember-cli-blueprint-test-helpers@0.19.2.patch
+
 importers:
 
   .:
     dependencies:
       '@babel/core':
         specifier: ^7.24.4
-        version: 7.24.4
+        version: 7.24.4(supports-color@8.1.1)
```
Windvis, line 22: There are a lot of lock changes, but I used the pnpm version that was listed in the package.json (8.10). I guess it's possible someone else used a different version in the past?

package.json
```diff
@@ -51,8 +51,9 @@
     "lint:eslint:fix": "npm-run-all \"lint:eslint --fix\"",
     "lint:fix": "npm-run-all lint:*:fix",
     "test": "node bin/run-tests.js",
-    "test:blueprints:js": "EMBER_TYPESCRIPT_BLUEPRINTS=false pnpm test:blueprints:ts",
-    "test:blueprints:ts": "mocha node-tests/blueprints/**/*-test.js",
+    "test:blueprints:js": "EMBER_TYPESCRIPT_BLUEPRINTS=false pnpm test:blueprints:transpiled-js",
+    "test:blueprints:transpiled-js": "mocha node-tests/blueprints/**/*-test.js",
+    "test:blueprints:ts": "mocha node-tests/blueprints-ts/**/*-test.js",
```
Windvis, line 56: Splitting the .ts tests into a separate script seemed the easiest, but `:ts` was already used so I renamed the other one. Not sure about the naming though.

package.json
```diff
@@ -51,9 +51,10 @@
     "lint:eslint:fix": "npm-run-all \"lint:eslint --fix\"",
     "lint:fix": "npm-run-all lint:*:fix",
     "test": "node bin/run-tests.js",
-    "test:blueprints:js": "EMBER_TYPESCRIPT_BLUEPRINTS=false pnpm test:blueprints:ts",
-    "test:blueprints:ts": "mocha node-tests/blueprints/**/*-test.js",
-    "test:blueprints": "pnpm test:blueprints:js && pnpm test:blueprints:ts",
+    "test:blueprints:js": "EMBER_TYPESCRIPT_BLUEPRINTS=false pnpm test:blueprints:transpiled-js",
+    "test:blueprints:transpiled-js": "mocha node-tests/blueprints/**/*-test.js",
+    "test:blueprints:ts": "mocha node-tests/blueprints-ts/**/*-test.js",
+    "test:blueprints": "npm-run-all test:blueprints:*",
```
Windvis, line 57: This script is used in CI and it seems to be the slowest job. We could probably split this up into separate jobs so the different blueprint tests are run in parallel which should speed things up (especially if we add more ts blueprint tests).

node-tests/blueprints-ts/component-test.js
```diff
@@ -0,0 +1,341 @@
...
+            replace: {
+              component: 'foo',
+              componentInvocation: 'Foo',
+            },
+          })
+        );
+      });
+    });
+
+    // classic default
```
Windvis, line 82: The classic tests need to be removed: #20773

package.json
```diff
@@ -51,9 +51,10 @@
     "lint:eslint:fix": "npm-run-all \"lint:eslint --fix\"",
     "lint:fix": "npm-run-all lint:*:fix",
     "test": "node bin/run-tests.js",
-    "test:blueprints:js": "EMBER_TYPESCRIPT_BLUEPRINTS=false pnpm test:blueprints:ts",
-    "test:blueprints:ts": "mocha node-tests/blueprints/**/*-test.js",
-    "test:blueprints": "pnpm test:blueprints:js && pnpm test:blueprints:ts",
+    "test:blueprints:js": "EMBER_TYPESCRIPT_BLUEPRINTS=false pnpm test:blueprints:js-from-ts",
+    "test:blueprints:js-from-ts": "mocha node-tests/blueprints/**/*-test.js",
+    "test:blueprints:ts": "mocha node-tests/blueprints-ts/**/*-test.js",
+    "test:blueprints": "npm-run-all test:blueprints:*",
```
Windvis, line 57: We can remove this one now since we no longer use it in CI, and I don't expect anyone to run all of them like this locally either.

blueprints/component/index.js
```diff
@@ -231,7 +231,7 @@ module.exports = {
 
   normalizeEntityName(entityName) {
     return normalizeEntityName(
-      entityName.replace(/\.js$/, '') //Prevent generation of ".js.js" files
+      entityName.replace(/\.(js|ts)$/, '') //Prevent generation of ".js.js" and ".ts.ts" files
```
NullVoxPopuli, line 234: Why does js.js happen?

NullVoxPopuli (approved):
Do you have a pr for the test-helpers as well?

blueprints/component/index.js
Windvis: When people add an extension when generating a component: `ember g component-class foo.js`. I just noticed when copying one of the tests and adjusting it for .ts.

Other ts blueprints will have the same issue.

Windvis:
> Do you have a pr for the test-helpers as well?

No, what test-helpers do you mean?

Windvis:
@NullVoxPopuli Is creating a PR in the [helpers repo](https://github.com/ember-cli/ember-cli-blueprint-test-helpers) and getting that merged / released a blocker for this one, or can we merge with the patch? That repo doesn't seem to be maintained and it has no Github actions setup either. I don't really have the time to get that repo in a good state so PRs can be merged I'm afraid.

Other than that I think I should split off the .ts.ts fix into a separate one (and fix it for all the .ts blueprints at the same time maybe?), but then this should be good to go.

Windvis:
Going to close this since it's outdated and we can just use `--typescript` where needed.

PR #20774, opened 2024-10-10
Start implementing the runtime compiler

wycats:
There's still a lot to do and clean up, but the bones of an implementation are here.


wycats:
Superseded by #20776 20776

PR #20775, opened 2024-10-10
add inspector debug

patricklx:
This needs 
- [ ] https://github.com/glimmerjs/glimmer-vm/pull/1489
- [ ] https://github.com/glimmerjs/glimmer-vm/pull/1575
- [ ] a way to correctly detect glimmer component or setup inspector debug info on it. Because it raises errors on some properties. Maybe this is not required anymore, since v2 removes the deprecation warning? https://github.com/emberjs/ember-inspector/wiki/customize-debug-info-of-components
- [ ] https://github.com/emberjs/ember-inspector/pull/2599

inspector needs:
 * name of mixins to display in object inspector
 * access to rsvp
 * application instance
   * __container__.cache
   * lookup()
   * runloop, backburner
 * the object inspector 
   * identify properties and its type: getter, service, tracked computed
   * track which properties changed
   * get tracked deps
 * access data-adapter:main


package.json
```diff
@@ -235,6 +237,46 @@
       "@ember/controller/index.js": "ember-source/@ember/controller/index.js",
       "@ember/debug/container-debug-adapter.js": "ember-source/@ember/debug/container-debug-adapter.js",
       "@ember/debug/data-adapter.js": "ember-source/@ember/debug/data-adapter.js",
+      "@ember/debug/ember-inspector-support/adapters/basic.js": "ember-source/@ember/debug/ember-inspector-support/adapters/basic.js",
```
NullVoxPopuli, line 240: do all of these need to be separate entries here? or can we bundle it all under ember-inspector-support.js?

patricklx: I'm not sure if it was me or some automated task that did this. Probably not necessary to have all here. Should never be directly imported by client code

ef4:
The two major points of feedback we discussed about this are:

 - the bulk of ember_debug really is inspector-owned code that we don't want to maintain here. We need to identify the stable API that is actually needs from Ember and expose that from within ember-source, and let inspector continue to worry about loading its own code and passing the ember stable API to that code.
 - no new code should be going into the AMD compat entry points. New modules are necessarily available to embroider builds, and they can also be available to classic builds by not filtering them out of the `renamed-modules` [here](https://github.com/emberjs/ember.js/blob/d4fd1fd8d55537cf07fd0a5043353c47cd82b2c7/lib/index.js#L37).

patricklx:
as a first step i reduces the files that are accessing ember

NullVoxPopuli:
We have a new approach we're going for (in RFC)

Thanks for exploring this, and making us care more about it!

PR #20776, opened 2024-10-10
Inplement runtime portion of RFC #931

wycats:


NullVoxPopuli:
Safari 17.6 passes locally for me. 
I see that browserstack is using Safari 12. 
We don't actually support that.

package.json
```diff
@@ -292,6 +292,30 @@
       "@ember/runloop/index.js": "ember-source/@ember/runloop/index.js",
       "@ember/service/index.js": "ember-source/@ember/service/index.js",
       "@ember/template-compilation/index.js": "ember-source/@ember/template-compilation/index.js",
+      "@ember/template-compiler/-internal-primitives.js": "ember-source/@ember/template-compiler/-internal-primitives.js",
```
NullVoxPopuli, line 295: just tested and this new stuff is not available outside of embroider 👍

NullVoxPopuli:
I tried this out in an embroider app here:
https://github.com/NullVoxPopuli/ember-source-pr-20776-testing

and it didn't work (seems like nothing is rendered?).


maybe whatever is going on with CI will be resolved when this reproduction is resolved?

NullVoxPopuli:
Looks like there are two copies of `@glimmer/manager` now, which breaks setComponentTemplate

ef4:
Your repro link is broken and this is not enough information to say what's going wrong. There are known module-duplication bugs that effect ember-source that are fixed only in the very latest `main` of embroider.

NullVoxPopuli:
> Your repro link is broken and this is not enough information to say what's going wrong.

I don't understand, the link works.
Also, all templates are just.. _not rendered_ (which I said 🤔 ).

Unless you mean something more specific? 

> There are known module-duplication bugs that effect ember-source that are fixed only in the very latest main of embroider.

in any case, I suppose for this feature we don't _want_ to fix for embroider3? which means we have to disable/skip the failing test?

NullVoxPopuli:
I confirm things work as expected in the vite blueprint:

image

NullVoxPopuli (approved):
All green!

ef4 (approved):

PR #20781, opened 2024-10-22
Initial implementation of renderComponent

wycats:
There's more in this commit than there should be, largely because this
commit bundles a change to the plugins that fixes lexical scope bugs.

I intend to separate those changes out into their own PR before
attempting to merge this one. This PR also needs a flag for the new API.

Finally this commit adds some new testing infrastructure to generalize
the base render tests so it can be used with templates that are not
registered into a container. This is useful more generally, and could
be used in other places in the test suite in the future.

### DIFF
--- a/packages/@ember/-internals/glimmer/lib/renderer.ts
+++ b/packages/@ember/-internals/glimmer/lib/renderer.ts
@@ -6,23 +6,24 @@ import { guidFor } from '@ember/-internals/utils';
 import { getViewElement, getViewId } from '@ember/-internals/views';
 import { assert } from '@ember/debug';
 import { _backburner, _getCurrentRunLoop } from '@ember/runloop';
-import { destroy } from '@glimmer/destroyable';
+import { associateDestroyableChild, destroy, isDestroyed } from '@glimmer/destroyable';
 import { DEBUG } from '@glimmer/env';
 import type {
   Bounds,
   CompileTimeCompilationContext,
+  CompileTimeResolver,
   Cursor,
   DebugRenderTree,
-  DynamicScope as GlimmerDynamicScope,
   ElementBuilder,
   Environment,
+  DynamicScope as GlimmerDynamicScope,
   RenderResult,
   RuntimeContext,
+  RuntimeResolver,
   Template,
   TemplateFactory,
 } from '@glimmer/interfaces';
 
-import { CurriedType } from '@glimmer/vm';
 import type { Nullable } from '@ember/-internals/utility-types';
 import { programCompilationContext } from '@glimmer/opcode-compiler';
 import { artifacts, RuntimeOpImpl } from '@glimmer/program';
@@ -35,22 +36,27 @@ import {
   DOMChanges,
   DOMTreeConstruction,
   inTransaction,
+  renderComponent as glimmerRenderComponent,
   renderMain,
   runtimeContext,
 } from '@glimmer/runtime';
 import { unwrapTemplate } from '@glimmer/util';
 import { CURRENT_TAG, validateTag, valueForTag } from '@glimmer/validator';
+import { CurriedType } from '@glimmer/vm';
 import type { SimpleDocument, SimpleElement, SimpleNode } from '@simple-dom/interface';
 import RSVP from 'rsvp';
-import type Component from './component';
+import { hasDOM } from '../../browser-environment';
+import type ClassicComponent from './component';
 import { BOUNDS } from './component-managers/curly';
 import { createRootOutlet } from './component-managers/outlet';
 import { RootComponentDefinition } from './component-managers/root';
 import { NodeDOMTreeConstruction } from './dom';
 import { EmberEnvironmentDelegate } from './environment';
+import { StrictResolver } from './renderer/strict-resolver';
 import ResolverImpl from './resolver';
 import type { OutletState } from './utils/outlet';
 import OutletView from './views/outlet';
+import { registerDestructor } from '@ember/destroyable';
 
 export type IBuilder = (env: Environment, cursor: Cursor) => ElementBuilder;
 
@@ -120,17 +126,71 @@ function errorLoopTransaction(fn: () => void) {
   }
 }
 
-class RootState {
+type RootState = ClassicRootState | ComponentRootState;
+
+class ComponentRootState {
+  readonly type = 'component';
+
+  #result: RenderResult | undefined;
+  #render: () => void;
+
+  constructor(
+    state: RendererState,
+    definition: object,
+    options: { into: Cursor; args?: Record<string, unknown> }
+  ) {
+    this.#render = errorLoopTransaction(() => {
+      let iterator = glimmerRenderComponent(
+        state.runtime,
+        state.builder(state.env, options.into),
+        state.compilation,
+        state.owner,
+        definition,
+        options?.args
+      );
+
+      let result = (this.#result = iterator.sync());
+
+      associateDestroyableChild(this, this.#result);
+
+      // override .render function after initial render
+      this.#render = errorLoopTransaction(() => result.rerender({ alwaysRevalidate: false }));
+    });
+  }
+
+  isFor(_component: ClassicComponent): boolean {
+    return false;
+  }
+
+  render(): void {
+    this.#render();
+  }
+
+  destroy(): void {
+    destroy(this);
+  }
+
+  get destroyed(): boolean {
+    return isDestroyed(this);
+  }
+
+  get result(): RenderResult | undefined {
+    return this.#result;
+  }
+}
+
+class ClassicRootState {
+  readonly type = 'classic';
   public id: string;
   public result: RenderResult | undefined;
   public destroyed: boolean;
   public render: () => void;
 
   constructor(
-    public root: Component | OutletView,
+    public root: ClassicComponent | OutletView,
     public runtime: RuntimeContext,
     context: CompileTimeCompilationContext,
-    owner: InternalOwner,
+    owner: object,
     template: Template,
     self: Reference<unknown>,
     parentElement: SimpleElement,
@@ -200,18 +260,18 @@ class RootState {
   }
 }
 
-const renderers: Renderer[] = [];
+const renderers: BaseRenderer[] = [];
 
 export function _resetRenderers() {
   renderers.length = 0;
 }
 
-function register(renderer: Renderer): void {
+function register(renderer: BaseRenderer): void {
   assert('Cannot register the same renderer twice', renderers.indexOf(renderer) === -1);
   renderers.push(renderer);
 }
 
-function deregister(renderer: Renderer): void {
+function deregister(renderer: BaseRenderer): void {
   let index = renderers.indexOf(renderer);
   assert('Cannot deregister unknown unregistered renderer', index !== -1);
   renderers.splice(index, 1);
@@ -219,7 +279,7 @@ function deregister(renderer: Renderer): void {
 
 function loopBegin(): void {
   for (let renderer of renderers) {
-    renderer._scheduleRevalidate();
+    renderer.rerender();
   }
 }
 
@@ -259,7 +319,7 @@ function resolveRenderPromise() {
 let loops = 0;
 function loopEnd() {
   for (let renderer of renderers) {
-    if (!renderer._isValid()) {
+    if (!renderer.isValid()) {
       if (loops > ENV._RERENDER_LOOP_LIMIT) {
         loops = 0;
         // TODO: do something better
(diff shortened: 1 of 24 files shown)


packages/@ember/-internals/glimmer/lib/renderer.ts
```diff
@@ -281,69 +341,265 @@ interface ViewRegistry {
...
+    isInteractive: boolean;
+    hasDOM: boolean;
+    document: SimpleDocument;
+  };
+}
 
-  private _owner: InternalOwner;
-  private _context: CompileTimeCompilationContext;
-  private _runtime: RuntimeContext;
+export class RendererState {
```
NullVoxPopuli, line 359: how different is all this from how Ember renders an app? 🤔

NullVoxPopuli: ah I see: https://github.com/emberjs/ember.js/pull/20781/files#diff-cdf95f0c6c10226a54fbac271478aaa1defd908f0f998fec0f484e52d2544184R603 and https://github.com/emberjs/ember.js/pull/20781/files#diff-cdf95f0c6c10226a54fbac271478aaa1defd908f0f998fec0f484e52d2544184L492 <- original code here, and a lot is added

packages/@ember/-internals/glimmer/lib/renderer.ts
```diff
@@ -378,108 +724,112 @@ export class Renderer {
...
+      return getViewElement(component);
     } else {
       throw new Error(
         'Accessing `this.element` is not allowed in non-interactive environments (such as FastBoot).'
       );
     }
   }
 
-  getBounds(view: View): {
+  getBounds(component: View): {
```
NullVoxPopuli, line 827: what's a "view"? 😅 🙈

packages/@ember/-internals/glimmer/lib/renderer/strict-resolver.ts
```diff
@@ -0,0 +1,40 @@
...
+  RuntimeResolver as VMRuntimeResolver,
+} from '@glimmer/interfaces';
+import { BUILTIN_HELPERS, BUILTIN_KEYWORD_HELPERS } from '../resolver';
+
+///////////
+
+/**
+ * Resolution for non built ins is now handled by the vm as we are using strict mode
+ */
+export class StrictResolver implements VMRuntimeResolver, VMCompileTimeResolver {
```
NullVoxPopuli, line 15: why do we need a resolver at all? can we not have one? 😅 

like, why do we still need lookupHelper?

packages/@ember/-internals/glimmer/lib/resolver.ts
```diff
@@ -120,7 +120,7 @@ function lookupComponentPair(
   }
 }
 
-const BUILTIN_KEYWORD_HELPERS: Record<string, object> = {
+export const BUILTIN_KEYWORD_HELPERS: Record<string, object> = {
```
NullVoxPopuli, line 123: are these _all_ things ember adds that glimmer doesn't have?

some of these `-`-prexifed ones are weird.
what's resolve? in-el-null? 🙈 are they used?

packages/@ember/-internals/glimmer/tests/integration/components/render-component-test.ts
```diff
@@ -0,0 +1,284 @@
...
+  renderComponent(
+    component: object,
+    options: { expect: string } | { classic: ClassicComponentShape }
+  ) {
+    let { owner } = this;
+
+    run(() => {
+      this.component = renderComponent(component, {
+        owner,
+        env: { document: document, isInteractive: true, hasDOM: true },
```
NullVoxPopuli, line 48: what happens when isInteractive is false?

packages/@ember/-internals/glimmer/tests/integration/components/render-component-test.ts
```diff
@@ -0,0 +1,284 @@
...
+    '@test Can use a custom modifier in scope'() {
+      let foo = defineSimpleModifier((element) => (element.innerHTML = 'Hello, world!'));
+      let Root = defComponent('<div {{foo}}></div>', { scope: { foo } });
+
+      this.renderComponent(Root, { expect: '<div>Hello, world!</div>' });
+    }
+
+    '@test Can shadow keywords'() {
+      let ifComponent = defineComponent({}, 'Hello, world!');
+      let Bar = defComponent('{{#if}}{{/if}}', { scope: { if: ifComponent } });
```
NullVoxPopuli, line 92: I understand that this is how strict mode is supposed to work, but I hope no one does this in a real project.

Could allow us to make some good memes tho, like how Cish folks redefine true to be false or whatever

packages/@ember/-internals/glimmer/tests/integration/components/render-component-test.ts
```diff
@@ -0,0 +1,284 @@
...
+    '@test Can use array'() {
+      let Root = defComponent('{{#each (array "Hello, world!") as |value|}}{{value}}{{/each}}', {
+        scope: { array },
+      });
+
+      this.renderComponent(Root, { expect: 'Hello, world!' });
+    }
+
+    '@test Can use concat'() {
+      let Root = defComponent('{{(concat "Hello" ", " "world!")}}', { scope: { concat } });
```
NullVoxPopuli, line 214: wrapping `()` shouldn't be needed

packages/@ember/-internals/glimmer/tests/integration/components/render-component-test.ts
```diff
@@ -0,0 +1,284 @@
...
+
+      this.renderComponent(Root, { expect: 'Hello, world!' });
+    }
+
+    '@test Can use get'() {
+      let Root = defComponent(
+        '{{#let (hash value="Hello, world!") as |hash|}}{{(get hash "value")}}{{/let}}',
```
NullVoxPopuli, line 221: same here with get, `{{get hash "value"}}` should be equiv

NullVoxPopuli: unless I am missing what is being tested here 🙃

NullVoxPopuli: (or if this just isn't an important distinction)

packages/@ember/-internals/glimmer/tests/integration/components/render-component-test.ts
```diff
@@ -0,0 +1,284 @@
...
+
+      this.renderComponent(Root, { expect: '<button>Click</button>' });
+
+      clickElement('button');
+
+      assert.verifySteps(['handleClick']);
+    }
+
+    // Ember currently uses AST plugins to implement certain features that
+    // glimmer-vm does not natively provide, such as {{#each-in}}, {{outlet}}
```
NullVoxPopuli, line 246: I don't know if I've gotten an answer on this -- but shouldn't the VM have each-in? how important is it where built ins live?
if we put more builtins in ember (say: on, hash, fn, array, element), are we saying that Glimmer shouldn't be used without ember?

NullVoxPopuli (approved):

packages/@ember/-internals/glimmer/lib/renderer.ts
```diff
@@ -281,69 +341,265 @@ interface ViewRegistry {
...
+  } else {
+    return { element: into as SimpleElement, nextSibling: null };
+  }
+}
+
+/**
+ * This function returns `undefined` if there was an error rendering the
+ * component.
+ *
+ * @fixme restructure this to return a result containing the error rather than
```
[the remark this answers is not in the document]
wycats: I intended it as a pre-merge note to decide whether it should be fixed or accepted as tolerable. It should still be addressed one way or another before merging.

packages/@ember/-internals/glimmer/lib/renderer.ts
```diff
@@ -281,69 +341,265 @@ interface ViewRegistry {
...
+ * undefined.
+ */
+export function renderComponent(
```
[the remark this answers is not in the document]
wycats: I think that's sensible. Maybe it should just be an option like it is in Glimmer: do you want the client-side element builder or the rehydrating one. That's the only difference in behavior between hydrating and normal rendering.

packages/@ember/-internals/glimmer/lib/renderer.ts
wycats: TL;DR almost every use of the word "view" means "classic component". I considered renaming all `view` to `classicComponent` but it was expanding the scope of this PR a lot. That said, it's actually really clarifying when you see an internal API that literally only takes classic components. It means it must not be all that fundamental to the rendering process since empirically we actually do render a lot of Glimmer components.

That said, it also illustrates a way in which route templates are special: they rely on mounting via APIs on the renderer, and (before this PR) there was literally no way to mount a root component or an outlet component in Ember that was implemented as a Glimmer component.

packages/@ember/-internals/glimmer/lib/renderer/strict-resolver.ts
wycats: Originally, I thought: I want Glimmer to support the "no resolver" case, and then we wouldn't need one. We'd still want it for now to avoid blocking this PR on that Glimmer change.

As I was implementing this feature, though, I realized that keywords like `array` and `hash` and builtins used by AST transformations are *not implemented* as *either* Glimmer keywords *or* AST transformations. Instead, many of them are implemented as custom helpers looked up through the resolver. This is *definitely wrong*, but it's another thing I don't want to try to fix at the same time as the rest of this PR.

packages/@ember/-internals/glimmer/lib/resolver.ts
wycats: Many of these are produced by AST transformations. I sort of understand how that came about, but it's not even the best way to accomplish the goal. And now that AST transformations have the ability to create imports in their containing files, we should absolutely update this code to do that rather than rely on these rube goldberg shenanigans.

packages/@ember/-internals/glimmer/tests/integration/components/render-component-test.ts
wycats: When `isInteractive` is false, modifiers don't run. When `hasDOM` is _true_, we default to getting `document` (and I think other browser-only globals) from `globalThis`. When it's _false_, you're required to explicitly provide them.

I think `isInteractive` makes sense (and exists in Glimmer), but `hasDOM` feels like a hack that escaped the lab and should be replaced with being explicit in the internals and using the default browser behavior at the edge.

packages/@ember/-internals/glimmer/tests/integration/components/render-component-test.ts
wycats: It's also impossible to represent using `<template>`, since `<template>` populates the scope with names in the JavaScript lexical scope, and `if` is a keyword. I don't think we ever specified this, but I assume that the content tag syntax is only expected to work in modules, or you'd be able to define a `yield` variable and shadow the `{{yield}}` keyword :yikes:.

packages/@ember/-internals/glimmer/tests/integration/components/render-component-test.ts
wycats: This test is explicitly testing the `get` keyword, so I had to come up with an example that uses `get`. ;)

packages/@ember/-internals/glimmer/tests/integration/components/render-component-test.ts
wycats: I think the VM should indeed have `each-in`. I also think that implementing these things as AST plugins is causing a lot more trouble than we seem to have realized. For now, I decided to document and clean up the behavior, but I agree strongly with the implication of your various comments that we should move a lot of the nuts-and-bolts implementations of low-level things into the VM.

packages/@ember/-internals/glimmer/lib/renderer.ts
wycats: For the record, I think that a lot of this could be factored to share more code. But when I tried to do more than I did here, I kept running into subtle assumptions I was missing in the original code, so I decided to share the obvious stuff and allow more sharing to happen incrementally once people have time to take a careful inventory of the expected behavior of the original code.

NullVoxPopuli:
RFC: https://github.com/emberjs/rfcs/pull/1099

hoping to get this landed soon :tm:

NullVoxPopuli:
Implementation finished here: https://github.com/emberjs/ember.js/pull/20962

This can be closed. Thanks for your initial pass on this, @wycats !

PR #20782, opened 2024-10-22
Fixes Ember keyword shadowing

wycats:
This commit consistently fixes the interaction between keywords and lexical variables (JS or Handlebars) to match the specified behavior: keywords are **always** superseded by in-scope lexical variables.

Ember keywords are implemented as AST transformations, and there were two sources of bugs that made implementation of these keywords inconsistent with the specified behavior:

1. In some cases, AST transforms applied to situations where the keyword in question was shadowed by a Handlebars block param. This means that the variable would be silently overridden by the AST transform.
2. In all cases, AST transforms applied when the keyword was shadowed by a lexical variable (i.e. in-scope JS variable) of the same name. This isn't supposed to happen, as lexical JS variables are supposed to have the same rules as variables bound via JS block params.

This commit has tests for many of these cases. Some additional tests are forthcoming, mostly in cases where the transform in question wasn't tested at all before this change.

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/helpers/array-test.js
+++ b/packages/@ember/-internals/glimmer/tests/integration/helpers/array-test.js
@@ -1,4 +1,10 @@
-import { RenderingTestCase, moduleFor, strip, runTask } from 'internal-test-helpers';
+import {
+  RenderingTestCase,
+  defineComponent,
+  moduleFor,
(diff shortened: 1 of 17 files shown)


NullVoxPopuli (approved):
lgtm :shipit:

ef4 (approved):

PR #20785, opened 2024-10-29
releasing @glimmer/component 2.0.0

ef4:
I tested beta.22
 - in typescript
 - in a classic app
 - in a stable embroider (webpack) app
 - in a main embroider (vite) app

 So this is ready to go.


ef4:
Published and tagged.

PR #20792, opened 2024-11-12
[BUGFIX release] importing ember-source/types no longer works (it should)

### LINKED ISSUE #20789: In 6.1-alpha, importing `ember-source/types` doesn't work, but the types: [] method does
idk if this is just how TS is changing, but in one PR: https://github.com/NullVoxPopuli/limber/pull/1864

I noticed that all of the `import 'ember-source/types'` lines no longer define the types of the ember-source modules.

Now, I know the blueprint, and other ecosystems have been preferring types (in tsconfig.json#compilerOptions#types):
```
"types": ["./node_modules/ember-source/types/stable"]
```

it's just a bit lengthy as this doesn't respect package.json#exports.

It is suspicious that changing the ember version caused the difference in behavior.

NullVoxPopuli:
Resolves: https://github.com/emberjs/ember.js/issues/20789

Which allows folks to use:
```ts
import 'ember-source/types';
```
in some ts file.

or

```ts
types: ['ember-source/types']
```
in tsconfig.json

Tested locally:
image
image

And one additional error state (which is reality before this PR)
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,9 @@
   ],
   "exports": {
     "./*": "./dist/packages/*",
+    "./types": {
+      "types": "./types/stable/index.d.ts"
+    },
     "./dist/ember-template-compiler.js": "./dist/ember-template-compiler.js",
     "./package.json": "./package.json"
   },


kategengler:
You tagged this `release` -- does it affect 6.0 or only 6.1?

NullVoxPopuli:
It does affect 6.0, yes

ef4 (approved):

PR #20796, opened 2024-11-15
Fix changelog for `@glimmer/component` v2 (supported version)

### LINKED ISSUE #20790: `@glimmer/component` v2 is not working for ember < 4.10
### 🐞 Describe the Bug
While whitelisting `@glimmmer/component` v2 in ember-basic-dropdown we have discoverd that isn't working in ember <= 4.8

The reason is that there was used  `@ember/owner` inside `@glimmer/components`: https://github.com/emberjs/ember.js/blob/main/packages/%40glimmer/component/src/index.ts#L5

Error:
`Error: Could not find module `@ember/owner` imported from ember-basic-dropdown/components/basic-dropdown`

CI error:
https://github.com/cibernox/ember-basic-dropdown/actions/runs/11777892189/job/32803253327?pr=941#step:6:202

The glimmer changelogs are telling ember < 3.13 is dropped see https://github.com/emberjs/ember.js/blob/main/packages/%40glimmer/component/CHANGELOG.md

mkszepp:
Like discussed in #20790 we should fix the changelog for `@glimmer/component` v2.

The package supports only `ember-source` >= 4.10, because it use `@ember/owner` which was introduced in [v4.10](https://github.com/emberjs/ember.js/blob/main/CHANGELOG.md#v4100-january-11-2023) of ember

~In additional there was added the `peerDependency`, so that consumer apps are getting dependency warning for incorrect version use~

close #20790

### DIFF
--- a/packages/@glimmer/component/CHANGELOG.md
+++ b/packages/@glimmer/component/CHANGELOG.md
@@ -3,13 +3,13 @@
 ## 2.0.0 2024-10-29
 
  - BREAKING: converted to V2 addon
- - BREAKING: dropped support for ember < 3.13
+ - BREAKING: dropped support for ember < 4.10
 
 
 ## 2.0.0-beta.22 2024-10-29
 
  - BREAKING: converted to V2 addon
- - BREAKING: dropped support for ember < 3.13
+ - BREAKING: dropped support for ember < 4.10
 
 
 ## Older Releases
--- a/packages/@glimmer/component/package.json
+++ b/packages/@glimmer/component/package.json
@@ -23,9 +23,6 @@
   "devDependencies": {
     "typescript": "5.1"
   },
-  "peerDependencies": {
-    "ember-source": ">= 4.10.0"
-  },
   "engines": {
     "node": ">= 18"
   },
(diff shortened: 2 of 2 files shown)


ef4:
The documentation updates are good. But I don't think we should add the peer dependency.

It's the "correct" thing to do, but given how buggy and misunderstood peer dependencies are I think it's just going to cause extra pain. 

For example:

1. If we add a peerDep from @glimmer/component to ember-source.
1. And AddonOne has a peerDependency on @glimmer/component (and this is a realistic thing that addons already do)
2. And AddonTwo depends on AddonOne.
3. And AddonTwo doesn't have a peerDep on ember-source
4. Then some package managers will install an entire second copy of ember-source, probably on a different version, to satisfy the peer dep. The symptom of this failure will range from "bizarre typescript errors" to runtime crashes due to missing component managers to silently bloated production builds.

Sometimes you really do need peer deps, and need to teach people how to deal with the fallout. But this particular peer dep isn't strictly necessary because there is nothing in @glimmer/component that is trying to import from a path starting with `ember-source`.

mkszepp:
@ef4 thanks for explaining, PR has now only changelog changes

PR #20800, opened 2024-11-22
[FEATURE] Allow components in routes

ef4:
Second attempt at #20768

* Refactor to make component the primary path, normalizes templates into a custom component with `{{this}}` set to controller
* Keep the core `{{outlet}}` responsibilities separate from what goes into the outlet
* Ensures proper debug render tree output – `route-template` node only appears when a template (the custom component) is used
* Ensures this works with test-helpers & others that
* It doesn't differentiate between components and templates once normalized so `@controller` and `@model` is passed to both

TODO:

* Feature flag
* Test and integrate this in @ember/test-helpers + smoke test
* Inlined some TODO comments to investigate/simplify things

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -26,6 +26,12 @@ module.exports = {
 
     'disable-features/disable-async-await': 'error',
     'disable-features/disable-generator-functions': 'error',
+    'import/no-unresolved': [
+      'error',
+      {
+        ignore: ['@ember/template-compiler'],
+      },
+    ],
   },
 
   settings: {
--- a/packages/@ember/-internals/glimmer/lib/component-managers/outlet.ts
+++ b/packages/@ember/-internals/glimmer/lib/component-managers/outlet.ts
@@ -1,25 +1,23 @@
 import type { InternalOwner } from '@ember/-internals/owner';
+import type { Nullable } from '@ember/-internals/utility-types';
 import { assert } from '@ember/debug';
 import EngineInstance from '@ember/engine/instance';
 import { _instrumentStart } from '@ember/instrumentation';
+import { precompileTemplate } from '@ember/template-compilation';
 import type {
-  CapturedArguments,
   CompilableProgram,
   ComponentDefinition,
-  CapabilityMask,
   CustomRenderNode,
   Destroyable,
   Environment,
   InternalComponentCapabilities,
-  Template,
   VMArguments,
   WithCreateInstance,
   WithCustomDebugRenderTree,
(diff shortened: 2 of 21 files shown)


ef4:
(This was reworked by @chancancode, I just opened the PR.)

ef4:
@chancancode what is the detail on the todo item "integrate this in @ember/test-helpers + smoke test"?

The PR does contain an end-to-end smoke test already, and I don't know what would need to change in `@ember/test-helpers` related to this.

chancancode:
The render test helpers (and perhaps other consumers) reach through private
APIs to directly set outlet state on the “root outlet view” without going
through routing.

TODO Now:

Part of the improvement here is to ensure that code path also works with
components by moving the normalization code out of routing into rendering.
Should have a test to ensure it works.

Can be later:

Said test helpers were updated to support components and does that using
the backwards wrapping into a template trip. With this they should (with
the appropriate version guards, but they already have infra for that) be
able to stop doing that and just pass a component in renderState.template.
I think that would be a good improvement and have few moving parts and
artifacts (avoid additional tracking frames in error messages etc).

 If we update the implementation in test helpers, in this repo there is a
specific test files that is supposed to mimic what they are doing to ensure
we don’t accidentally break test helpers. So that should be done together.
Of course could also just move that into this monorepo etc.

chancancode:
Not _quite_, I think we do have tests that tries to roughly cover that already, and indeed I don't think anything I changed would have broken that: https://github.com/emberjs/ember.js/blob/main/packages/ember/tests/ember-test-helpers-test.js

What I am saying was that, one key change I made here over the original PR was that I moved the normalization logic outside of the routing layer (when we resolve the route templates) and pushed it into the rendering layer (where we _render_ the outlet).

One reason that is an improvement is because ("things like") ember-test-helpers may [bypasses the routing code and set the outlet state directly](https://github.com/emberjs/ember-test-helpers/blob/master/addon/src/setup-rendering-context.ts#L158-L189). So the refactor I did ensures that would also work.

To that end I think it's good to have _some_ test exercising that code path, confirming setting the outlet state directly with a component actually does work. One way to do that is add another module in the [ember-test-helpers smoke test](https://github.com/emberjs/ember.js/blob/main/packages/ember/tests/ember-test-helpers-test.js) and speculatively change to what we think ember-test-helpers should do moving forward. An artificial unit test somewhere else may also work, but I don't think it'd be that much simpler.

ef4:
Got it, thanks.

Issue #20802, opened 2024-11-23
Improve Readability of Static for() Method Implementation


sson68x:
During a code review of the `static for()` method in [overrides.js](https://github.com/emberjs/ember.js/blob/main/lib/overrides.js), I observed that while the method effectively manages add-on overrides, the readability could be improved. Specifically, the method contains nested loops and conditions that make the code harder to follow and maintain, especially for new contributors. Simplifying the structure and reducing nesting could improve code clarity and facilitate easier contribution from developers with varying experience levels.

I recommend refactoring the method to reduce complexity, improve readability, and make it more approachable for developers who may be unfamiliar with the current implementation.

NullVoxPopuli:
Are you ai... From a different project?

What overrides?
What static for?

kategengler:
I do see the method and the file, however I think it is fine. I do think that this entire file could probably be removed -- cc @ef4 to check that assumption?

PR #20808, opened 2024-12-06
Modify class Overrides

sson68x:
Minor update to the class Overrides snippet for readability


bertdeblock:
It's awesome that you want to contribute, but this probably isn't a very useful change to make.
You could also consider looking at https://help-wanted.emberjs.com and find something you might want to pick up if you still want to contribute.

kategengler:
We discussed this in the issue you opened https://github.com/emberjs/ember.js/issues/20802 -- this code is not worth touching.

PR #20809, opened 2024-12-07
Clean up `EMBER_TYPESCRIPT_BLUEPRINTS` feature flag

bertdeblock:
Was [enabled by default](https://github.com/emberjs/ember.js/pull/20352) almost two years ago.
Seems like we can clean this up, and remove all `blueprints-js` files from this repo?


blueprints/-addon-import.js
```diff
@@ -4,18 +4,9 @@ const stringUtil = require('ember-cli-string-utils');
 const path = require('path');
 const inflector = require('inflection');
 
-const maybePolyfillTypeScriptBlueprints = require('./-maybe-polyfill-typescript-blueprints');
-
 module.exports = {
   description: 'Generates an import wrapper.',
 
-  shouldTransformTypeScript: true,
```
bertdeblock, line 12: `*-addon` blueprints (app re-exports) always need to generate JS files, so no need to try and transform TS -> JS for these.

PR #20810, opened 2024-12-08
Define `packageManager` in package file

bertdeblock:


package.json
```diff
@@ -397,6 +397,7 @@
       ]
     }
   },
+  "packageManager": "pnpm@8.10.0",
```
bertdeblock, line 400: This helps Corepack users use the correct pnpm version locally.

.github/actions/setup/action.yml
```diff
@@ -11,7 +11,6 @@ runs:
     - uses: pnpm/action-setup@v4
       name: Install pnpm
       with:
-        version: 8
```
bertdeblock, line 14: The action also looks at the `packageManager` field. One less location to update the desired pnpm version.

NullVoxPopuli (approved):

PR #20813, opened 2024-12-15
Enable `deprecate-import-from-ember` deprecation

NullVoxPopuli:
What are folks thoughts on enabling this? and then the deprecation messages would force us to work on improving the inspector experience?

I have some ideas for how to bring back compatibility.

Our setup would be something like this:

```js
const EMBER_INSPECTOR_SUPPORT = Symbol.for('__EMBER_INSPECTOR__');
globalThis[EMBER_INSPECTOR_SUPPORT] ||= {};
globalThis[EMBER_INSPECTOR_SUPPORT].init = async function loadEmberInspectorUtilities() {
  // store things on the secret object 2 lines above
  globalThis[EMBER_INSPECTOR_SUPPORT].tracking = await import(...);
  /* ... etc ... */.container = await import(...);

  // ensuring to set up an event listener pair for communicating via postMessage with the inspector
}
```

This should probably _wrap_ all of this code: https://github.com/emberjs/ember.js/pull/20775/files#diff-b2c569a1648b65b7314971b3bd201a82fda450a5cccc8c3d415b864166c98e27

And then on the app side, we only need to load the module, since everything the inspector needs is await-imported
```js
import '@ember/inspector-support';
```

And then on the inspector side, we access the globalThis
```js
// loading the inspector
await globalThis[EMBER_INSPECTOR_SUPPORT].init();

// we can now start building the UI / Panels in the ember-inspector project.
```

@PatrickJS , @ef4 thoughts?

### DIFF
--- a/packages/@ember/-internals/deprecations/index.ts
+++ b/packages/@ember/-internals/deprecations/index.ts
@@ -95,7 +95,7 @@ export const DEPRECATIONS = {
     return deprecation({
       id: `deprecate-import-${dasherize(importName).toLowerCase()}-from-ember`,
       for: 'ember-source',
-      since: { available: '5.10.0' },
+      since: { available: '5.10.0', enabled: '6.3.0' },
       until: '7.0.0',
       url: `https://deprecations.emberjs.com/id/import-${dasherize(
         importName
--- a/packages/@ember/-internals/deprecations/index.ts
+++ b/packages/@ember/-internals/deprecations/index.ts
@@ -95,7 +95,7 @@ export const DEPRECATIONS = {
     return deprecation({
       id: `deprecate-import-${dasherize(importName).toLowerCase()}-from-ember`,
       for: 'ember-source',
-      since: { available: '5.10.0', enabled: '6.3.0' },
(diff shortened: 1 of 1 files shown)


kategengler:
Could you bring this comment to the RFC? That's really where enabled gets decided -- the 'Ready for Release' stage

patricklx:
There is already a wip pr for compatibility #20775 .
The minimal setup would be to just manually define the AMD modules inspector needs and trigger the 'ember' event so inspector nows when ember was loaded

NullVoxPopuli:
My preference would be to ditch AMD, even for the inspector :see_no_evil:

patricklx:
Then we can just work on #20775 :)
But well, removing the barrel is not about removing AMD. I think the only thing left that the inspector needs from the barrel import is actually the 'ember' event, at the very end of the barrel file. But there might be some more especially at the boot up code of inspector

NullVoxPopuli:
> Then we can just work on https://github.com/emberjs/ember.js/pull/20775 :)

I'll review soon (Tomorrow?)

kategengler (approved):

PR #20825, opened 2025-01-13
Remove the `helper` wrapper from the helper generator

Windvis:
Plain function helpers have been supported since ember-source 4.5 so we no longer need the wrapping `helper` call.


NullVoxPopuli (approved):

kategengler:
Thank you!

PR #20826, opened 2025-01-16
Add unit tests for the ComputedProperty class

Oleari19:
Description:
This pull request introduces a new file, computed-property-test.js, which adds unit tests for the get method of the ComputedProperty class. These tests validate different scenarios and edge cases to ensure proper behavior of the get method when interacting with computed properties in Ember.js.

Test Cases:
        CT1: Validates that get returns undefined when the revision is not set, and validateTag returns false.
        CT2: Ensures get returns the expected computed value when the revision is set, and validateTag returns false.
        CT3: Confirms get returns the computed value when the revision is not set, but validateTag returns true.
        CT4: Validates the behavior when dependent keys are present, and validateTag returns false.
        CT5: Tests get when the computed property returns an array value.

Simulated Conditions:
        Used helper functions such as metaFor, tagFor, tagMetaFor, and validateTag to simulate metadata and tag-related behaviors.

Assertions:
        Each test uses assert methods like equal and deepEqual to confirm that the returned values match the expected outcomes.

Purpose:

These tests enhance the coverage and reliability of the ComputedProperty class, ensuring its behavior aligns with expected results under various conditions.


kategengler:
Computeds are extensively unit tested already https://github.com/emberjs/ember.js/blob/main/packages/%40ember/-internals/metal/tests/computed_test.js

Computed properties, while not yet deprecated, are a part of the classic Ember Object model and are replaced by [@tracked](https://guides.emberjs.com/release/in-depth-topics/autotracking-in-depth/).

kategengler:
Closing since there was no response. Computeds are well tested already.

PR #20835, opened 2025-01-23
[FEATURE] Add a `--route-authoring-format` option to the route generator

Windvis:
Now that https://github.com/emberjs/ember.js/pull/20800 is merged, it would also be nice if we could generate new routes with a .gjs / .gts template already.

```
ember g route foo --route-authoring-format=strict
```
or the shorthand:
```
ember g route foo --strict
```
or as the project's default by adding `routeAuthoringFormat: 'strict'` in .ember-cli

### Current status
Ready to review!

### DIFF
--- a/blueprints/route/files/__root__/__templatepath__/__templatename__.gts
+++ b/blueprints/route/files/__root__/__templatepath__/__templatename__.gts
@@ -1,0 +1,14 @@
+import type { TOC } from '@ember/component/template-only';
+<% if (addTitle) {%>import { pageTitle } from 'ember-page-title';<%}%>
+
+interface <%= routeName %>Signature {
+  Args: {
+    model: unknown;
+    controller: unknown;
+  };
+}
+
+<template>
+  <% if (addTitle) {%>{{pageTitle "<%= routeName %>"}}
+  <%}%>{{outlet}}
+</template> satisfies TOC<<%= routeName %>Signature>;
--- a/blueprints/route/index.js
+++ b/blueprints/route/index.js
@@ -29,6 +29,12 @@ module.exports = {
       name: 'reset-namespace',
       type: Boolean,
     },
+    {
+      name: 'route-authoring-format',
+      type: ['loose', 'strict'],
+      default: 'loose',
+      aliases: [{ loose: 'loose' }, { strict: 'strict' }],
+    },
   ],
 
   init() {
@@ -78,6 +84,23 @@ module.exports = {
         return 'app';
       },
     };
+  },
+
+  files() {
+    let files = this._super.files.apply(this, arguments);
+
+    if (this.options.routeAuthoringFormat === 'strict') {
+      const strictFilesToRemove =
+        this.options.isTypeScriptProject || this.options.typescript ? '.gjs' : '.gts';
+      files = files.filter(
+        (file) =>
+          !(file.endsWith('.js') || file.endsWith('.hbs') || file.endsWith(strictFilesToRemove))
+      );
+    } else {
+      files = files.filter((file) => !(file.endsWith('.gjs') || file.endsWith('.gts')));
+    }
+
+    return files;
   },
 
   locals: function (options) {
--- a/node-tests/fixtures/route/strict-route-template.gts
+++ b/node-tests/fixtures/route/strict-route-template.gts
@@ -1,0 +1,14 @@
+import type { TOC } from '@ember/component/template-only';
+import { pageTitle } from 'ember-page-title';
+
+interface FooSignature {
+  Args: {
+    model: unknown;
+    controller: unknown;
+  };
+}
+
+<template>
+  {{pageTitle "Foo"}}
+  {{outlet}}
+</template> satisfies TOC<FooSignature>;
--- a/blueprints/route/files/__root__/__templatepath__/__templatename__.gjs
+++ b/blueprints/route/files/__root__/__templatepath__/__templatename__.gjs
@@ -1,0 +1,6 @@
+<% if (addTitle) {%>import { pageTitle } from 'ember-page-title';
+
+<%}%><template>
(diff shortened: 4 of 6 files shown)


package.json
```diff
@@ -128,7 +128,7 @@
     "babel-plugin-ember-template-compilation": "^2.1.1",
     "dag-map": "^2.0.2",
     "decorator-transforms": "2.0.0",
-    "ember-cli": "^4.10.0",
+    "ember-cli": "^5.12.0",
```
Windvis, line 131: This bump is needed because the .gts -> .gjs conversion is only available since ember-cli 5.5. I took 5.12 since it's the current LTS, but I can switch to a different version if needed.

package.json
```diff
@@ -396,7 +396,7 @@
   },
   "packageManager": "pnpm@8.10.0",
   "volta": {
-    "node": "16.20.0",
+    "node": "18.20.6",
```
Windvis, line 399: Node 16 support was already dropped (https://github.com/emberjs/ember.js/pull/20716) so this simply updates the volta config to match that. 

Without this the content-tag package doesn't work correctly (Webassembly related errors).

blueprints/route/files/__root__/__templatepath__/__templatename__.gts
Windvis, line 1: The .gts -> .gjs conversion doesn't seem to be working as expected and it fails the blueprint test. I'm not sure what's up with that.

blueprints/route/index.js
```diff
@@ -80,6 +85,16 @@ module.exports = {
...
 
+  files() {
+    let files = this._super.files.apply(this, arguments);
+
+    return files.filter((file) => {
+      const extensionToRemove = this.options.strictTemplate ? '.hbs' : '.gts';
+
+      return !file.endsWith(extensionToRemove);
+    });
+  },
```
Windvis, line 96: This isn't doing the correct thing yet and breaks all the tests :smile:

blueprints/route/index.js
```diff
@@ -29,6 +29,11 @@ module.exports = {
       name: 'reset-namespace',
       type: Boolean,
     },
+    {
+      name: 'strict-template',
```
Windvis, line 33: Not sure if this is the best name for the option. Suggestions?

package.json
Windvis: It seems this bump also causes other test failures related to pods, need to look into this further but the failures are unrelated to the new code here.

blueprints/route/files/__root__/__templatepath__/__templatename__.gts
Windvis: Resolved in babel-remove-types v1.0.1.

package.json
Windvis: I worked around it by using the `--pod` argument where needed instead. Most pod tests already did that, only a specific set of tests where using the `usePods` option.

blueprints/route/index.js
Windvis: Another option could be to name it something like `--strict-route-template` but also add a `--strict` alias. That way people can add the more verbose version to their ember-cli file in (the unlikely) case they don't want to use the strict option for the components version.

NullVoxPopuli:
I've added this to the agenda for us to discuss a name on tuesday. thanks for working on this!!!

blueprints/route/index.js
Windvis: A third option is to match the setup from the [component PR](https://github.com/emberjs/ember.js/pull/20862/files#diff-b3654ccb6dc2075bf987b07346f57bfd89f7874ca631cf76040a1afe49ee0a9eR53).

And add a `route-template-authoring-format` option with `--loose` and `--strict` shorthands.

Windvis: We went with `route-authoring-format` since that's the most consistent with the component version.

blueprints/route/files/__root__/__templatepath__/__templatename__.gts
```diff
@@ -0,0 +1,6 @@
+<% if (addTitle) {%>import { pageTitle } from 'ember-page-title';
+
+<%}%><template>
+  <% if (addTitle) {%>{{pageTitle "<%= routeName %>"}}
+  <%}%>{{outlet}}
+</template>
```
Windvis, line 6: To check, do we need a newline here?

blueprints/route/files/__root__/__templatepath__/__templatename__.gts
```diff
@@ -0,0 +1,6 @@
+<% if (addTitle) {%>import { pageTitle } from 'ember-page-title';
+
```
Windvis, line 2: We should add types and a .gts fixture to match.

kategengler (approved):

kategengler:
I'm 👍 but this will need rebasing after #20862 lands

NullVoxPopuli:
there are no conflicts -- why rebase? 🤔

kategengler:
@NullVoxPopuli 
CleanShot 2025-04-29 at 11 33 59@2x

Windvis:
Yea, there are indeed conflicts, and there are also some other things I adjust (.gts tests, more types, ...). I need to find some time to look into it again.

node-tests/fixtures/route/strict-route-template.gts
```diff
@@ -0,0 +1,14 @@
+import type { TOC } from '@ember/component/template-only';
+import { pageTitle } from 'ember-page-title';
+
+interface FooSignature {
+  Args: {
+    model: unknown;
+    controller: unknown;
```
Windvis, line 7: I matched what the [template-tag codemod outputs](https://github.com/embroider-build/embroider/blob/4574bb78d25a3d2bffc602b1484e089b0f5c4af1/packages/template-tag-codemod/src/index.ts#L121).

kategengler:
Thank you!

PR #20801, opened 2024-11-22
[BUGFIX LTS] Upgrade glimmer vm dependencies

NullVoxPopuli:
Companion PR: https://github.com/glimmerjs/glimmer-vm/pull/1693

---

Update: too many changes to backport to 5.12, probably

---

NOTE:

This should be backported to 5.12

---

Most of what's in these VM changes is prep for bigger bugfixes and improvements (especially to debug tooling / inspector stuff).

The timing of this PR is early, as we don't want a _big_ VM upgrade all at once.
Every change has gone through performance testing, and found either no regression, or slight improvement.

Changelog: https://github.com/glimmerjs/glimmer-vm/pull/1626

Updates:
- Code supporting partials has been removed
- Better debug symbols in strict-mode

PR #20842, opened 2025-02-07
Upgrade glimmer-vm

NullVoxPopuli:
Replaces: https://github.com/emberjs/ember.js/pull/20801

--------------------

> [!WARNING]  
> This requires we drop TS < 5, because we need `verbatimModuleSyntax`. TS v5 came out in March of 2023, so folks have had plenty of time to move to a newer version (and this is even beyond the window of our Major release cadence, and I believe dropping this is fine, even though, we now _require_ consumers use `verbatimModuleSyntax`.

Most of what's in these VM changes is prep for bigger bugfixes and improvements (especially to debug tooling / inspector stuff).

The timing of this PR is early, as we don't want a _big_ VM upgrade all at once.
Every change has gone through performance testing, and found either no regression, or slight improvement.

Changelog: 
- https://github.com/glimmerjs/glimmer-vm/releases
- https://github.com/glimmerjs/glimmer-vm/pull/1626

Updates:
- Code supporting partials has been removed
- Better debug symbols in strict-mode 

-------------

notes:

long ago, I made https://github.com/NullVoxPopuli/tsc-lite because `tsc` does not the _wrong thing_ by hiding your most important errors at the top of the buffer (and sometimes you have enough errors that it surpasses your max-buffer, so you can't scroll to the most important error). tsc-lite shows only the first few chunks of the tsc output, so you can focus on what to fix first, in the proper order.

This has come in handy here, as we broke the types in glimmer-vm (they need to be rolled up).


NullVoxPopuli:
Progress
image

bin/run-tests.js
```diff
@@ -81,7 +81,13 @@ run()
     process.exit(0); // eslint-disable-line n/no-process-exit
   })
   .catch(function (err) {
-    console.error(chalk.red(err.toString()));
+    console.error(chalk.red(`Error!`));
+
+    if ('passed' in err) {
```
NullVoxPopuli, line 86: This change will print out:
```
{ passed: 40140, failed: 4, total: 40144, runtime: 47482 }
```
where as before it would print out `[object Object]`

ef4 (approved):

kategengler:
Is our "rolling window" for TS support documented anywhere better than the 'Stability' heading here? https://blog.emberjs.com/stable-typescript-types-in-ember-5-1/

cc @emberjs/typescript-core

NullVoxPopuli:
Perhaps: 
- https://www.semver-ts.org/#for-package-authors

NullVoxPopuli:
It might be this:

> Packages using the “rolling window” policy should normally support all TypeScript versions released during the current ‘LTS’ of other core packages/runtimes/etc. they support, and drop support for them only when they drop support for that ‘LTS’, to minimize the number of major versions in the ecosystem.

So if we have 2+ LTS' with 5.0, we're good with that being our minimum, I think?
With the same logic, if we have 2+ LTS' with 5.7, we're good with that being our minimum? :see_no_evil:

kategengler:
I understand what the rolling support window concept _is_ -- I am looking for somewhere we have documented which literal versions are supported by which versions of Ember that could be linked to in the release blog post. How can users easily tell which versions of Ember support which TS versions?

NullVoxPopuli:
I haven't seen that information exist anywhere.

Do we have similar information for node support? should it live in the same place?

kategengler:
Node support is here https://github.com/ember-cli/ember-cli/blob/master/docs/node-support.md
Which should _also_ probably be linked from the website, at very least.

gitKrystan:
> afaict the only place it is somewhat documented is in the CI config

Indeed this is the case.

PR #20850, opened 2025-02-15
Fix source maps in published packages

chancancode:
I noticed this problem from this warning in my build:

```
Warning: ignoring input sourcemap for ../rewritten-packages/@embroider/synthesized-vendor/vendor/ember/ember-testing.js because
ENOENT: no such file or directory, open '.../node_modules/.embroider/rewritten-packages/@embroider/synthesized-vendor/vendor/ember/ember-template-compiler.js.map'
```

Indeed, it seems like as of 5.10.0-beta.1, we stopped publishing source maps:

https://unpkg.com/browse/ember-source@5.10.0-beta.1/dist/

...despite having every intention to do so still.

Last good version:

https://unpkg.com/browse/ember-source@5.10.0-alpha.4/dist/

It took me way to long to spot this, but the problem is in the extension; in the broccoli build, they had a `$name.map` filename, whereas in the new build (#20675) it's `$name.js.map`.

Adjusted the entries in package.json to match; also removed the types publish script that was accidentally included in the tarball.

* * *

As for my original quest to fix the warning, it turns out this is not sufficient. The file is placed there by @embroider/compat these days:

https://github.com/embroider-build/embroider/pull/2164

It'll have to be updated to also emit the source map file (if it is present, since this has been broken for a while).

But this is a prerequisite to fixing that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,16 +25,16 @@
     "dist/packages",
     "dist/dependencies",
     "dist/ember-template-compiler.js",
-    "dist/ember-template-compiler.map",
+    "dist/ember-template-compiler.js.map",
     "dist/ember.debug.js",
-    "dist/ember.debug.map",
+    "dist/ember.debug.js.map",
     "dist/ember.prod.js",
-    "dist/ember.prod.map",
+    "dist/ember.prod.js.map",
     "dist/ember-testing.js",
-    "dist/ember-testing.map",
+    "dist/ember-testing.js.map",
     "docs/data.json",
     "lib",
-    "types"
+    "types/stable"
   ],
   "repository": {
     "type": "git",
--- a/lib/index.js
+++ b/lib/index.js
@@ -130,14 +130,19 @@ module.exports = {
 
     let templateCompiler = new Funnel(tree, {
       destDir: 'ember',
-      include: ['ember-template-compiler.js', 'ember-template-compiler.map'],
+      include: ['ember-template-compiler.js', 'ember-template-compiler.js.map'],
     });
 
     let which = isProduction ? 'prod' : 'debug';
 
     let ember = new Funnel(tree, {
       destDir: 'ember',
-      include: [`ember.${which}.js`, `ember.${which}.map`, 'ember-testing.js', 'ember-testing.map'],
+      include: [
+        `ember.${which}.js`,
+        `ember.${which}.js.map`,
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -25,16 +25,16 @@
     "dist/packages",
     "dist/dependencies",
```
chancancode, line 26: So this one is also broken. Last good version is:

https://unpkg.com/browse/ember-source@5.10.0-alpha.4/dist/dependencies/

If it's been broken for so long and nobody complained, I'm not sure if it needs to be brought back. There are a lot of code in the build scripts to try and support this, but clearly something isn't working. I don't care about those and did not investigate why.

If we want to remove it, it would take some effort to read through the build code to find and clean up the defunct code for this, so I'm leaving that out of scope for this PR for someone else to decide/pick up.

chancancode:
I'm also not sure how important _these_ source maps are and if it warrants a LTS backport, so someone else would have to decide that. I think sources maps, in general, _are_ important, but I don't know what, if anything, actually uses these `dist` files and if this would cause problems for those consumers.

package.json
```diff
@@ -25,16 +25,16 @@
...
+    "dist/ember.debug.js.map",
     "dist/ember.prod.js",
-    "dist/ember.prod.map",
+    "dist/ember.prod.js.map",
     "dist/ember-testing.js",
-    "dist/ember-testing.map",
+    "dist/ember-testing.js.map",
     "docs/data.json",
     "lib",
-    "types"
```
NullVoxPopuli, line 37: Why'd this change?

chancancode: https://unpkg.com/browse/ember-source@6.2.0/types/

NullVoxPopuli: Fair! Thank you!

ef4 (approved):

ef4:
Thanks, this PR looks good. 

> but I don't know what, if anything, actually uses these dist files and if this would cause problems for those consumers.

This ember bundles in dist are used by:

 - classic builds
 - embroider builds with
   - ember-source < 6.2 and
   - @embroider/core < 4.0.0-alpha
   - embroider configuration `staticEmberSource: false`

With the exception of ember-template-compiler.js which is still used by pretty much everybody.

kategengler:
So, should this be backported?

PR #20851, opened 2025-02-17
Fix nightly TS

NullVoxPopuli:
Looks like it wasn't building `@glimmer/component`, so this should do it

Tested locally via:
```bash
pnpm install
pnpm build
pnpm add --save-dev typescript@next --workspace-root
pnpm type-check
```


kategengler:
Thank you!

PR #20857, opened 2025-02-25
Upgrade to pnpm 10

ef4:
This is working toward harmonization with the glimmer-vm monorepo.


CvX:
Please note that Dependabot [might not be compatible](https://github.com/dependabot/dependabot-core/issues/11246) with v10 yet (v9 upgrade before it was a [breaking](https://github.com/dependabot/dependabot-core/issues/1736#issuecomment-2061127509) one)

kategengler:
We don't make much use of dependabot in this repo, so that's probably okay.

PR #20862, opened 2025-03-01
[FEATURE] Strict component blueprint support

Windvis:
This implements the blueprint work of [RFC #779](https://rfcs.emberjs.com/id/0779-first-class-component-templates/#blueprints).

I squashed all the work from #20511 into a single commit to make rebases easier.

Closes #20511

### Current status:
Should be good to review. I'll clean up the git history once I know if the current approach is fine and I no longer need the option to revert individual commits.


Windvis:
Something seems to go wrong when we remove the types from the new .gts test file.

```
  1) Blueprint: component-test
       in app
         with default setup
           component-test x-foo --strict:
     SyntaxError: unknown: Unexpected token, expected "," (12:46)

  10 |     // and update using state.myProperty = 1; await rerender();
  11 |     // Handle any actions with function myAction(val) { ... };/* ___NEWLINE___ */
> 12 |     await render(template = __TEMPLATE_TAG_0__;);/* ___NEWLINE___ */
     |                                               ^
  13 |     assert.dom().hasText('');/* ___NEWLINE___ */
  14 |     // Template block usage:
  15 |     await render(template = __TEMPLATE_TAG_1__;);/* ___NEWLINE___ */
  ```

I think this might be a bug in ember-cli's .gts [detyping logic](https://github.com/ember-cli/ember-cli/blob/8f8d02b56ffe201a98e33c8247b0569395f3d82d/lib/models/blueprint.js#L539-L566), but I'm not familiar enough with ember-template-imports to know the way forward here.

**Comment by Windvis** [comment]
Cross posting the open question from https://github.com/emberjs/ember.js/pull/20511#issuecomment-2393872939:

In Octane the component-class defaults to --no-component-class. I'm wondering if that's the most useful default for .gts components though, especially since switching from template only .gts files to class based ones isn't as easy due to the types.

Since this is a new flag, and the RFC doesn't specify it, would it make sense to generate a class based .gts file by default, and a TO version if the --component-class=@ember/component/template-only is provided?

File: package.json [hunk 1]
```diff
@@ -127,7 +127,7 @@
     "babel-plugin-ember-template-compilation": "^2.1.1",
     "dag-map": "^2.0.2",
     "decorator-transforms": "2.0.0",
-    "ember-cli": "^4.10.0",
+    "ember-cli": "^6.1.0",
```
Windvis, line 130: This actually installed v6.2.2 due to the `^`so I should just list that version instead or use `~` instead.

ef4:
Naming proposal from the tooling team meeting discussion: can we use the edition setting in package.json to control both component and route blueprint generation?

```js
 "ember": {
    "edition": "polaris" // causes all components and routes to generate with template tag
  },
```

Also, if separate settings for each feature are also desirable, the value could be "polaris" vs "octane"

```
// .ember-cli
componentAuthoringFormat = "polaris"
routeAuthoringFormat = "octane"
```

Windvis:
> Naming proposal from the tooling team meeting discussion: can we use the edition setting in package.json to control both component and route blueprint generation?

@ef4 I think we can use that edition value to flip the defaults, but I wouldn't use it as the only option. The [RFC explicitly mentioned](https://rfcs.emberjs.com/id/0779-first-class-component-templates/#transition-path) a `--strict` and `--loose` flag for the component  generator.

Checking the edition value feels a bit out of scope for this PR though, since it would require other code changes (or the merge of #20863). The wider ecosystem might also not be ready yet for a value other than `octane` so I don't think we can recommend people from updating that edition value in their package.json files until we actually are closer to shipping it?

My preference would be to stick with the `--component-authoring-format="strict"`  with `--strict` and `--loose` aliases, as currently implemented, get it merged, and follow up with a PR that adds support for setting `edition: "polaris"` where needed.

My question about the argument name wasn't for this PR, but the route one :smile:, but using `route-authoring-format` makes sense so I can go for that, and make it match the setup here.

node-tests/fixtures/component-test/rfc232.gjs
Windvis, line 1: Probably should add a .gts version of this as well.

node-tests/blueprints/component-test.js
```diff
@@ -270,6 +271,52 @@ describe('Blueprint: component', function () {
...
+        (_file) => {
+          expect(_file('app/components/foo.gts')).to.equal(
+            fixture('component/glimmer-component.gts')
+          );
+        }
+      );
+    });
   });
 
   describe('in addon', function () {
```
Windvis, line 322: Probably should add some addon tests as well.

node-tests/blueprints/component-test.js
```diff
@@ -270,6 +271,52 @@ describe('Blueprint: component', function () {
         }
       );
     });
+
+    it('component foo --strict', function () {
+      return emberGenerateDestroy(['component', 'foo', '--strict'], (_file) => {
+        expect(_file('app/components/foo.gjs')).to.equal(
+          fixture('component/template-only-component.gjs')
```
Windvis, line 278: This one still fails because detyping the .gts version has a different output. We should probably use a .gjs without detyping for this scenario.

Windvis:
The smoke test scenarios seem to fail because of the included patch for ember-cli:

```
ERR_PNPM_PATCH_NOT_APPLIED  The following patches were not applied: ember-cli@6.2.2

Either remove them from "patchedDependencies" or update them to match packages in your dependencies.
Error: Process completed with exit code 1.
```

Not sure why though, since that patch config is not in the package.json of the smoke test app. Not sure what the fix is here. I think we _need_ the patch, unless the fix in ember-cli is backported and released, and we can update to that one.

node-tests/fixtures/component/template-only-component.gts
```diff
@@ -0,0 +1,16 @@
...
+  Args: {};
+  // Any blocks yielded by the component
+  Blocks: {
+    default: []
+  };
+  // The element to which `...attributes` is applied in the component template
+  Element: null;
+}
+
+<template>
```
Windvis, line 14: This needs to be `export default <template>` it seems, otherwise it doesn't seem to work. Not sure if that's a bug or intended.

ef4: I think that is a typescript or glint issue. Template tag by itself is supposed to imply export default, and that definitely works when no TS is involved.

NullVoxPopuli: I think it's a babel-plugin-ember-template-compilation issue, actually.

Here is a failing test: https://github.com/emberjs/babel-plugin-ember-template-compilation/pull/70

NullVoxPopuli: from discord: https://discord.com/channels/480462759797063690/485447409296736276/1337849624635113593

NullVoxPopuli: maybe another failing test? https://github.com/embroider-build/content-tag/pull/96

idk, this one seems like it's barfing on TS Syntax for some reason?

johanrd: Yes, I also see this when migrating from ember-route-template. The workaround to add `export default` before `<template>` seems to work, but it was unexpected per the [examples in the RFC](https://rfcs.emberjs.com/id/1046-template-tag-in-routes/) that were linked to in the release notes of 6.3.0.

Windvis: Resolved by content-tag v3.1.2 so we can safely output it like this :+1:

blueprints/component/index.js
```diff
@@ -135,6 +161,21 @@ module.exports = {
         }
       });
     }
+    if (this.options.componentAuthoringFormat === 'strict') {
+      const strictFilesToRemove =
+        this.options.isTypeScriptProject || this.options.typescript ? '.gjs' : '.gts';
```
Windvis, line 166: I'm not 100% sure if the `isTypeScriptProject` check is needed. It's possible that when that is true, `options.typescript` is also true, but I would have to double check.

Windvis:
I decided to work around all the different .gts detyping issues by simply adding .gjs blueprint files and selecting the correct one based on the host project. This allows us to drop all the workarounds and also means we are no longer blocked by the ember-cli backport PR.

I think this is the best way forward, and we can always refactor later if needed.

node-tests/fixtures/component-test/addon.gts
```diff
@@ -0,0 +1,28 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'dummy/tests/helpers';
+import { render } from '@ember/test-helpers';
+import Foo from 'dummy/components/foo';
```
Windvis, line 4: I'm not sure about this one, I think we may want to import from the addon namespace in this scenario, right? So from `my-addon/components/foo` in the tests?

ef4: Yes, I think it would be better to use the addon's namespace.

package.json
```diff
@@ -127,7 +127,7 @@
     "babel-plugin-ember-template-compilation": "^2.1.1",
     "dag-map": "^2.0.2",
     "decorator-transforms": "2.0.0",
-    "ember-cli": "^4.10.0",
+    "ember-cli": "^6.3.0-beta.3",
```
kategengler, line 130: Should this now go to 6.3.0?

kategengler:
I'm happy to merge after updating the cli version to non-beta.

package.json
kategengler: I'm going to make this change

PR #20863, opened 2025-03-03
[CLEANUP] Cleanup "octane edition" code

Windvis:
Octane has been the default for multiple major versions, so we can remove and simplify the code.


lib/index.js
```diff
@@ -82,44 +81,6 @@ module.exports = {
...
-
-    if (
-      !optionalFeaturesMissing &&
-      optionalFeatures.isFeatureEnabled('jquery-integration') &&
-      typeof optionalFeatures.isFeatureExplicitlySet === 'function' &&
-      optionalFeatures.isFeatureExplicitlySet('jquery-integration')
-    ) {
-      throw new SilentError(
-        'Setting the `jquery-integration` optional feature flag to `true` was deprecated in Ember 3.x and removed in Ember 4.0.0. You must add the `@ember/optional-features` addon and set this feature to `false`.\n\nFor more information, see the deprecation guide: https://deprecations.emberjs.com/v3.x/#toc_optional-feature-jquery-integration'
-      );
```
Windvis, line 121: Not sure about this change. I _think_ we can simply remove these errors, but we could also keep them around (without the octane check).

NullVoxPopuli (approved):
lgtm

Issue #20876, opened 2025-03-18
Breaking type changes in 6.4


ef4:
Between ember 6.3 and 6.4-beta, a bunch of `@glimmer/*` packages made breaking changes to how they publish their types. Those types are incorporated as part of Ember's public API.

An example is `@glimmer/manager` 0.92.4 -> 0.94.8 which dropped the `types` key in package.json in favor of `exports.*.types`. This forces apps to change their `moduleResolution` setting, which they cannot always do because other dependencies may not yet work under that setting.

This is going to cause real apps to be blocked from upgrading. I don't think it's reasonable to force a global config change in a minor.

I think we should either 

1. patch the glimmer packages to bring back type compatibility, 
2. or work on rolling their types into ember-source and dropping them from ember-source's `dependencies` instead of incorporating them by reference.

Option 2 is consistent with our vision of integrating the glimmer-vm repo into this repo and making all the internal glimmer packages an entirely internal concern. That gives us the maximal freedom to refactor them. It's also consistent with the fact that their **implementation** is already rolled into ember-source. It's Not Good that their implementation comes from ember-source but their types come from the real packages.

kategengler:
Is this resolved now?

ef4:
It's fixed in https://github.com/emberjs/ember.js/pull/20877

PR #20877, opened 2025-03-18
[BUGFIX beta] Re-add support for moduleResolution=node10

NullVoxPopuli:
VM PR: https://github.com/glimmerjs/glimmer-vm/pull/1741

Text from that PR Description:

This feels crazy to add support for the **node10** moduleResolution setting, but as pointed out in:
https://github.com/emberjs/ember.js/issues/20876

Some projects aren't up to date enough to use more modern moduleResolution settings -- and every TS dep needs to be compat with both in order to allow a consumer to migrate.

From the [tsconfig docs](https://www.typescriptlang.org/tsconfig/#moduleResolution)

> Default: Node10 if [module](https://www.typescriptlang.org/tsconfig/#module) is CommonJS; Node16 if [module](https://www.typescriptlang.org/tsconfig/#module) is Node16 or  Node18; NodeNext if [module](https://www.typescriptlang.org/tsconfig/#module) is NodeNext; Bundler if [module](https://www.typescriptlang.org/tsconfig/#module) is Preserve; Classic otherwise.

_No one should be using Node10 if they're writing ESM_.

`moduleResolution: node` was renamed to `moduleResolution: node10` with the release of TypeScript 5.0, which was released in March of 2023, which at the time, ember LTS was 4.8 with 4.12 being released on May 2023.

So it is within the TS Rolling support policy to be reasonable to have dropped support for `moduleResolution: node10`, and earlier TS versions.
(assuming config changes't aren't covered under [semver-ts](https://www.semver-ts.org/) -- I opened [this issue on semver-ts](https://github.com/semver-ts/semver-ts/issues/31))

However, supporting node10 is easy, as this PR is one line change in each published package.json


ef4:
@kategengler can we get a new beta that includes this?

kategengler:
@ef4 Is it urgent or is the regularly schedule beta on Monday okay?

ef4:
Regularly scheduled is fine, just wanted to make sure it gets into some beta. Thanks.

Issue #20816, opened 2024-12-23
Clean up deprecations that were until 6.0


kategengler:
Clean up and remove code gated by deprecations that were [`until: 6.0`](https://github.com/emberjs/ember.js/blob/main/packages/%40ember/-internals/deprecations/index.ts#L105-L141).

 - [ ] DEPRECATE_IMPLICIT_ROUTE_MODEL
 - [ ] DEPRECATE_TEMPLATE_ACTION
 - [ ] DEPRECATE_COMPONENT_TEMPLATE_RESOLVING
 - [ ] DEPRECATE_ARRAY_PROTOTYPE_EXTENSIONS -- this one is atypical and requires more digging to figure out what needs to be removed and what tests need to be updated

For each deprecation, you'll find calls to `deprecateUntil` with the deprecation, for example: 

```
deprecateUntil(`Usage of the \`{{action}}\` modifier is deprecated. Migrate to native functions and function invocation.`,
DEPRECATIONS.DEPRECATE_TEMPLATE_ACTION);
```

These calls and everything in that code path should be removed. As of 6.0, these calls with `until: 6.0` behave as throws so none of the code in that path is in use.

There are also calls to `testUnless` with the deprecations, like so:
```
`${testUnless(DEPRECATIONS.DEPRECATE_TEMPLATE_ACTION.isRemoved
```

These tests can be removed.

Finally, the deprecations themselves should be removed from `deprecations/index.ts`.

kategengler:
This is complete!

PR #20881, opened 2025-03-22
[CLEANUP] Remove DEPRECATE_COMPONENT_TEMPLATE_RESOLVING

tcjr:
Related to #20816.

The removal of this deprecation means that the resolver's internal function `layoutFor` would always return `null`.   It's only called in one place, so I removed the function and updated `lookupComponentPair` to use `null` in place of the return from `layoutFor`.


packages/@ember/-internals/glimmer/tests/integration/components/append-test.js
```diff
@@ -286,355 +278,6 @@ class AbstractAppendTest extends RenderingTestCase {
     );
   }
 
-  [`${testUnless(
```
kategengler, line 289: @NullVoxPopuli can you confirm that these tests were not erroneously marked with the deprecation?  At a glance it seems like they were coincidentally using the deprecated behavior but I'm not sure.

packages/@ember/-internals/glimmer/tests/integration/components/append-test.js
```diff
@@ -286,355 +278,6 @@ class AbstractAppendTest extends RenderingTestCase {
...
-
-      oldRegisterComponent.call(this, name, options);
-    };
-
-    this.registerComponent('x-parent', {
-      ComponentClass: Component.extend({
-        layoutName: 'components/x-parent',
-      }),
-
-      resolveableTemplate:
```
NullVoxPopuli, line 380: if we change this to `template`, and then unwrap the expectDeprecation thing, we should be good to go! 

(keeping the tests, but also resolving the deprecation)

tcjr: I restored the tests (unchanged) in this file, removing the `testUnless` deprecation wrapper.  

I'm not sure how to fix the failing tests; I'm still working on that.  Just changing `resolveableTemplate` to `template` was insufficient (different resolution errors).

NullVoxPopuli:
it might not be rebased -- like main could have some other deprecation removed that is still present in this branch?

kategengler:
@NullVoxPopuli ah I see -- wonder if we should gate the size comment on the PR being up-to-date?

NullVoxPopuli:
Since Ci/Tests run with an ephemeral "what would be if the PR were merged" branch, is there a way we can get that ref and use that?

kategengler:
> Since Ci/Tests run with an ephemeral "what would be if the PR were merged" branch, is there a way we can get that ref and use that?

hm, github.ref should already be that on pull_request...

kategengler:
@tcjr I think something has gone wrong here -- it is showing files as new that already exist 🤔

tcjr:
> @tcjr I think something has gone wrong here -- it is showing files as new that already exist 🤔
Fouled up this branch. Gonna make a clean PR.

PR #20882, opened 2025-03-22
[CLEANUP] Remove DEPRECATE_TEMPLATE_ACTION

tcjr:
Related to #20816.

The cleanup for this deprecation is pretty substantial, as it completely deletes two longstanding pieces of code from the resolver: the action modifier and the action helper.  One consequence of this is that `action` is no longer a built-in keyword and can presumably be used now like any other helper and modifier.


kategengler (approved):

tcjr:
@NullVoxPopuli @kategengler  I restored the non-modifer parts of the modifier source file.  All lints and tests and checks pass locally.  

I am not sure how much of that _very_ old event dispatcher code is in use anymore.  Maybe "send" or some other old event mechanism still hits that code path.

kategengler:
While we deprecated and remove the action modifier and helper we didn't deprecate or remove the `actions` hash or the TargetActionSupport stuff https://github.com/emberjs/rfcs/blob/deprecate-target-action-support/text/0000-deprecate-target-action-support.md has some more info.

tcjr:
In a post-`{{action}}` modifier world, should we just remove this test?

```js
  QUnit.test('Should not attempt to render element modifiers GH#14220', function (assert) {
    assert.expect(1);

    this.template('application', "<div {{action 'foo'}}></div>");

    return this.renderToHTML('/').then(function (html) {
      assert.htmlMatches(html, '<body><div></div></body>');
    });
  });
```

From [app-boot-test.js](https://github.com/emberjs/ember.js/blob/main/tests/node/app-boot-test.js#L138-L146)

NullVoxPopuli:
> In a post-{{action}} modifier world, should we just remove this test?

We could replace action with `on`

tcjr:
> We could replace action with `on`

I believe this a FastBoot test that is verifying that a modifier that adds something to the DOM doesn't do it in the FastBoot context.  For `{{action}}`, it adds a `data-` attribute.  The `on` modifier doesn't modify the DOM (afaik), so _I think_ we would need to create one for this test that changes the HTML.

packages/@ember/-internals/glimmer/lib/resolver.ts
```diff
@@ -158,9 +155,9 @@ if (DEBUG) {
   BUILTIN_HELPERS['-disallow-dynamic-resolution'] = disallowDynamicResolution;
 }
 
-const BUILTIN_KEYWORD_MODIFIERS: Record<string, ModifierDefinitionState> = {
-  action: actionModifier,
-};
+// With the implementation of RFC #1006, the `action` modifer was removed. It was the
```
kategengler, line 158:
```suggestion
// With the implementation of [RFC #1006](https://rfcs.emberjs.com/id/1006-deprecate-action-template-helper), the `action` modifer was removed. It was the
```

small edit to add rfc url for later and to kick ci

packages/@ember/-internals/glimmer/lib/resolver.ts
```diff
@@ -158,9 +155,9 @@ if (DEBUG) {
   BUILTIN_HELPERS['-disallow-dynamic-resolution'] = disallowDynamicResolution;
 }
 
-const BUILTIN_KEYWORD_MODIFIERS: Record<string, ModifierDefinitionState> = {
-  action: actionModifier,
-};
+// With the implementation of [RFC #1006](https://rfcs.emberjs.com/id/1006-deprecate-action-template-helper), the `action` modifer was removed. It was the
```
kategengler, line 158:
```suggestion
// With the implementation of RFC #1006 (https://rfcs.emberjs.com/id/1006-deprecate-action-template-helper), the `action` modifer was removed. It was the
```
another to bump ci

packages/@ember/-internals/glimmer/lib/resolver.ts
```diff
@@ -158,9 +155,9 @@ if (DEBUG) {
   BUILTIN_HELPERS['-disallow-dynamic-resolution'] = disallowDynamicResolution;
 }
 
-const BUILTIN_KEYWORD_MODIFIERS: Record<string, ModifierDefinitionState> = {
-  action: actionModifier,
-};
+// With the implementation of RFC #1006 (https://rfcs.emberjs.com/id/1006-deprecate-action-template-helper), the `action` modifer was removed. It was the
```
kategengler, line 158:
```suggestion
// With the implementation of [RFC #1006](https://rfcs.emberjs.com/id/1006-deprecate-action-template-helper), the `action` modifer was removed. It was the
```
and again

packages/@ember/-internals/glimmer/lib/resolver.ts
kategengler, line 158:
```suggestion
// With the implementation of RFC #1006(https://rfcs.emberjs.com/id/1006-deprecate-action-template-helper), the `action` modifer was removed. It was the
```
again

NullVoxPopuli:
> In a post-{{action}} modifier world, should we just remove this test?

maybe we just delete the test? it doesn't seem to be doing much. Then we can land this PR! 🎉

kategengler:
I've enabled auto-merge, thanks!

PR #20883, opened 2025-03-24
Port the sizes workflow from glimmer-vm

NullVoxPopuli:
Example from glimmer-vm: https://github.com/glimmerjs/glimmer-vm/pull/1690#issuecomment-2610621905

We should probably ship a prod build of ember (even if we don't have ember-cli / vite consumers use it), because its a good indicator of the impact of terser settings.

This PR does not do anything with prod builds as ember-source currently does not have one and relies on the app to correctly configure that stuff.

The sizes action is a good indicator of growth / regression / shrinkage for all the modules, and allows us to hopefully prevent situations like what happened during the 5.6 VM upgrade.

### DIFF
--- a/.github/workflows/on-main.yml
+++ b/.github/workflows/on-main.yml
@@ -1,4 +1,4 @@
-name: "Build on #main"
+name: "on #main"
 
 on:
   push:
--- a/.github/workflows/receive-pr.yml
+++ b/.github/workflows/receive-pr.yml
@@ -1,4 +1,9 @@
 # https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/
+#
+# Do a build
+# Measure assets sizes
+# Upload artifact
+# Consumed by size.yml for comparison
 name: Receive PR
 
 # read-only repo token
@@ -14,14 +19,24 @@ jobs:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup
       - run: pnpm build
+      - run: sudo snap install dust
 
       - name: Save PR number
         run: |
           mkdir -p ./pr
           echo ${{ github.event.number }} > ./pr/NR
 
-      - name: Save dist output
-        run: mv dist pr/dist
+      - name: "Get sizes for development outputs"
+        id: dev
+        run: |
+          cd dist
+          dust --ignore_hidden \
+              --reverse --apparent-size \
+              --filter ".+.js$" \
+              --no-percent-bars \
+              --only-file \
+              --full-paths > pr/out.txt
+
 
       - uses: actions/upload-artifact@v4
         with:
--- a/.github/workflows/receive-pr.yml
+++ b/.github/workflows/receive-pr.yml
@@ -1,0 +1,29 @@
+# https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/
+name: Receive PR
+
+# read-only repo token
+# no access to secrets
+on:
+  pull_request:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup
+      - run: pnpm build
+
+      - name: Save PR number
+        run: |
+          mkdir -p ./pr
+          echo ${{ github.event.number }} > ./pr/NR
+
+      - name: Save dist output
+        run: mv dist pr/dist
+
+      - uses: actions/upload-artifact@v4
+        with:
+          name: pr
+          path: pr/
--- a/.github/workflows/size-pr.yml
+++ b/.github/workflows/size-pr.yml
@@ -4,7 +4,7 @@
 # Measure assets sizes
 # Upload artifact
 # Consumed by size.yml for comparison
-name: Receive PR
+name: "Size: PR"
 
 # read-only repo token
 # no access to secrets
@@ -42,3 +42,4 @@ jobs:
         with:
           name: pr
           path: pr/
+          overwrite: true
--- a/.github/workflows/size-pr.yml
+++ b/.github/workflows/size-pr.yml
@@ -35,7 +35,7 @@ jobs:
               --filter ".+.js$" \
               --no-percent-bars \
               --only-file \
-              --full-paths > /out.txt
+              --full-paths > out.txt
           cp out.txt ../pr/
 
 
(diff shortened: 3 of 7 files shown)


.github/workflows/size.yml
```diff
@@ -0,0 +1,127 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      BRANCH:
+        description: Branch to checkout
+        required: false
+        default: 'main'
+        type: string
+  # Prevents forks / PRs from getting a GITHUB_TOKEN with write access
+  pull_request_target:
```
kategengler, line 12: I do not think this is a correct use of `pull_request_target`. My understanding is that `pull_request_target` actually gives PRs a token with read/write permissions and should NOT run (`pnpm build` would count) code from the pull requests themselves.

https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target 

says

> For workflows that are triggered by the pull_request_target event, the GITHUB_TOKEN is granted read/write repository permission unless the permissions key is specified and the workflow can access secrets, even when it is triggered from a fork. Although the workflow runs in the context of the base of the pull request, you should make sure that you do not check out, build, or run untrusted code from the pull request with this event. Additionally, any caches share the same scope as the base branch. To help prevent cache poisoning, you should not save the cache if there is a possibility that the cache contents were altered. For more information, see [Keeping your GitHub Actions and workflows secure: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests) on the GitHub Security Lab website.

kategengler (changes requested):

.github/workflows/size.yml
kategengler: https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/

.github/workflows/size.yml
```diff
@@ -0,0 +1,104 @@
+name: Size
+
+# read-write repo token
+# access to secrets
+#
+# https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/
+on:
+  workflow_run:
```
kategengler, line 8: I would make this workflow run on `pull_request` and save `out.txt` from the PR to artifacts. We could then add another workflow to run the sizes on `main` and keep that in an artifact as well to save some CI time.

The other workflow, `comment-sizes-pr` or similar, would use both outputs to comment on the PR.

NullVoxPopuli: done (tho I think you have the workflows backwards :upside_down_face: )

.github/workflows/on-main.yml
```diff
@@ -0,0 +1,37 @@
+name: "on #main"
```
kategengler, line 1: Can we name this something like "Sizes - main"

.github/workflows/receive-pr.yml
```diff
@@ -0,0 +1,44 @@
+# https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/
+#
+# Do a build
+# Measure assets sizes
+# Upload artifact
+# Consumed by size.yml for comparison
+name: Receive PR
```
kategengler, line 7: Can we name this something like "Sizes - pr"

.github/workflows/receive-pr.yml
```diff
@@ -0,0 +1,44 @@
...
+              --reverse --apparent-size \
+              --filter ".+.js$" \
+              --no-percent-bars \
+              --only-file \
+              --full-paths > pr/out.txt
+
+
+      - uses: actions/upload-artifact@v4
+        with:
+          name: pr
```
kategengler, line 43: This will fail the second time it runs -- name must be unique or have `overwrite: true` https://github.com/actions/upload-artifact#inputs

.github/workflows/size.yml
```diff
@@ -0,0 +1,95 @@
+name: Size
```
kategengler, line 1: Can we name this something like `Sizes - comment PR`

.github/workflows/size.yml
```diff
@@ -0,0 +1,95 @@
...
+    types:
+      - completed
+
+jobs:
+  compare_sizes:
+    name: 'Compare Sizes and Comment'
+    runs-on: 'ubuntu-latest'
+
+    steps:
+      - uses: actions/checkout@v4
```
kategengler, line 19: Does this still need to checkout and install?

.github/workflows/size.yml
```diff
@@ -0,0 +1,95 @@
...
+      #  | Dev  |   x1    |  y1  |
+      #  | Prod |   x2    |  y2  |
+      #
+      #  NOTE: we we don't have a prod build for this library
+      #        because we currently expect non-compiler usage
+      #        (so consumers should have terser or similar properly configured for DCE)
+      #
+      #########################
+      - uses: mshick/add-pr-comment@v2
+        with:
```
kategengler, line 72: Does this need some sort of configuration of which PR to comment on?

kategengler (changes requested):

.github/workflows/on-main.yml
NullVoxPopuli: Yeah, I was thinking everything that needs to run on main could go in here, like eventually #main-docs-publishing, benchmark results, etc

.github/workflows/size.yml
NullVoxPopuli: oh probably now that it doesn't run on pull_request, ya

.github/workflows/size-pr.yml
```diff
@@ -0,0 +1,45 @@
...
+              --reverse --apparent-size \
+              --filter ".+.js$" \
+              --no-percent-bars \
+              --only-file \
+              --full-paths > pr/out.txt
+
+
+      - uses: actions/upload-artifact@v4
+        with:
+          name: pr
```
kategengler, line 43: I don't think this works -- I think the name is global for all runs and so this will overwrite the artifact for all PRs

kategengler: I think you want to concat the run id

.github/workflows/size.yml
NullVoxPopuli: Nope!

.github/workflows/size-pr.yml
NullVoxPopuli: done, thanks!

kategengler (approved):

PR #20891, opened 2025-04-01
[CLEANUP] Remove DEPRECATE_COMPONENT_TEMPLATE_RESOLVING deprecation

tcjr:
Related to #20816.

The removal of this deprecation means that the resolver's internal function `layoutFor` would always return `null`.   It's only called in one place, so I removed the function and updated `lookupComponentPair` to use `null` in place of the return from `layoutFor`.

While doing the cleanup, we determined that some of the tests were erroneously marked with the deprecation, so for those tests -- in the `tests/integration/components/append-test.js` file -- the `testUnless` wrapper was removed.  One test was modified to build the test component using `template` instead of `resolvableTemplate`.

### DIFF
--- a/packages/@ember/-internals/deprecations/index.ts
+++ b/packages/@ember/-internals/deprecations/index.ts
@@ -102,16 +102,6 @@ export const DEPRECATIONS = {
       ).toLowerCase()}-from-ember`,
     });
   },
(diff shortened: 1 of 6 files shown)


NullVoxPopuli:
awww yeeee, look at that 0.2KB reduction! nice!

kategengler (approved):

PR #20908, opened 2025-05-06
Working toward glimmer-vm repo integration

ef4:
This is a work-in-progress on harmonizing the glimmer-vm repo with this repo so they can eventually merge.


ef4:
This now has merged the history and unified the test suites. 

 - [ ] ~test logging is missing on CI~ (not specific to this branch)
 - [x] there are four remaining failures (that we can't see in detail) on the production mode tests
 - [ ] lints are skipped, we want to move the glimmer-vm packages into the real packages dir first in order to unify their type-build tsconfigs.

kategengler (approved):

sandstrom:
Have you discussed merging more Ember repositories under the `emberjs` org? 

Repos under orgs other than `emberjs`:

- github.com/ember-cli
- github.com/ember-learn
- github.com/glimmerjs (now moved)
- github.com/ember-fastboot (dead?)
- github.com/ember-engines (dead?)
- github.com/ember-template-lint
- github.com/ember-codemods
- github.com/ember-a11y
- github.com/ember-modifier
- github.com/embroider-build
- github.com/ember-tooling
- github.com/adopted-ember-addons
- github.com/universal-ember

I would suggest something like:

- `emberjs` org; Main org, which would include all official stuff (Ember CLI, Embroider, IDE tooling, etc).
- `ember-community` org; An Ember core team controlled community org, where the org is managed by Ember, but each repo isn't. This is clearly stated. For example ember-a11y, ember-codemods and perhaps experiments like ember-engines and ember-fastboot (if they are still relevant) could live there. Things that aren't a good fit for adopted-ember-addons (I would also rename it to `ember-adopted-addons` for symmetry with other repo names).
- `ember-adopted-addons` org; already a pretty good org scope, no change (but I would rename from `adopted-ember-addons` to `ember-adopted-addons`).

kategengler:
Hi @sandstrom. I appreciate your thoughts on this. We merged glimmer-vm in and later, router_js, for technical reasons more than any organizational reasons. glimmer-vm being part of the ember repo makes sense because it is (for the most part) only used by ember.js. When it was separate we suffered an integration step whenever we made changes in glimmer-vm. 

router_js had a similar problem -- as we look towards making a Route Manager, we need to more easily be able to iterate between the two and the library deliniation was arbitrary.

We have a comprehensive list of orgs and repos but don't feel there needs to be a logical grouping in each github org. I am working on a tool to surface all the repos for discoverability purposes. 

A primary reason we keep many orgs is github actions usage limits and concurrency. 

I'd also like to correct: `ember-engines` and `fastboot` are official parts of Ember and not experiments. Engines was [RFC'd](https://rfcs.emberjs.com/id/0010-engines/). However, while supported by many people from Ember core, `adopted-ember-addons` is not owned by the ember project.

sandstrom:
Alright, I understand your point of view. Thanks for taking time answering 😀 

Happy Easter! 🐣🐇

PR #20909, opened 2025-05-09
Remove microtask in runtime compiler

NullVoxPopuli:
This came out of
- https://github.com/emberjs/ember.js/pull/20907

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/components/runtime-template-compiler-implicit-test.ts
+++ b/packages/@ember/-internals/glimmer/tests/integration/components/runtime-template-compiler-implicit-test.ts
@@ -1,6 +1,5 @@
-import { tracked } from '@ember/-internals/metal';
 import { template } from '@ember/template-compiler/runtime';
-import { RenderingTestCase, defineSimpleModifier, moduleFor, runTask } from 'internal-test-helpers';
+import { RenderingTestCase, defineSimpleModifier, moduleFor } from 'internal-test-helpers';
 import GlimmerishComponent from '../../utils/glimmerish-component';
 import { on } from '@ember/modifier/on';
 import { fn } from '@ember/helper';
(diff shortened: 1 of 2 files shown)


packages/@ember/template-compiler/lib/template.ts
```diff
@@ -242,12 +242,10 @@ export function template(
   const normalizedOptions = compileOptions(options);
   const component = normalizedOptions.component ?? templateOnly();
 
-  queueMicrotask(() => {
```
NullVoxPopuli, line 245: TODO: for @NullVoxPopuli 
- https://github.com/emberjs/rfcs/pull/931
- does the RFC say anything about this?


Potential use case:
- forward references (like functions)

ef4:
- the global CI failure for Basic Tests is probably unrelated
 - but when you make it run locally with `vite build` (like CI does), rollup is killing parts of the test that are only used by the `eval`.
 - when that happens, we observe a secondary bug behavior, which is that instead of erroring on trying to invoke an undefined component, the template renders and emits a custom element named `<state.component>`. This violates the expectation that all paths containing dots are supposed to be component invocations.

NullVoxPopuli:
> but when you make it run locally with vite build (like CI does), rollup is killing parts of the test that are only used by the eval.

fixed.



Hoping we can merge soon, cause I really need 
- https://github.com/emberjs/ember.js/issues/20914 resolved <3 :partying_face:

kategengler:
Do we want to open another issue for this?

> when that happens, we observe a secondary bug behavior, which is that instead of erroring on trying to invoke an undefined component, the template renders and emits a custom element named <state.component>. This violates the expectation that all paths containing dots are supposed to be component invocations.

packages/@ember/-internals/glimmer/tests/integration/components/runtime-template-compiler-implicit-test.ts
```diff
@@ -7,6 +7,29 @@ import { fn } from '@ember/helper';
...
+    async '@test can immediately render a runtime-compiled template'() {
+      class State {
+        get component() {
+          return template(`hello there`);
+        }
+      }
+
+      let state = new State();
+
+      hide(state);
```
kategengler, line 19: @ef4 Now we know `hide` isn't just for eslint

kategengler (approved):

packages/@ember/-internals/glimmer/tests/integration/components/runtime-template-compiler-implicit-test.ts
NullVoxPopuli: I think somewhere it was document (on hide maybe?) that this is for tricking JIT stuff

PR #20915, opened 2025-06-06
Change assert to warn for unneeded restriction around creating sub-applications

NullVoxPopuli:
Alternative to: https://github.com/emberjs/ember.js/pull/20910

PR description from there:

A path to unblocking:
- https://github.com/NullVoxPopuli/limber/pull/1925
  - This Limber PR has a patch of the change in this EmberJS PR  

While working new repl infra for https://limber.glimdown.com, I want to take a more "islands" approach, like https://astro.build -- it enables a ton of power -- however, there is an assertion that prevents using nested applications -- which kinda feels silly since we allow it via "engines".

Like with how we use modifiers and 3rd party libraries that manage DOM, whenever we use an element that has some external control of its descendents, we must make the same assumptions and allowances --- generally this is safe, as once we let 3rd party scripts manage their subtree, ember doesn't mess around in there (because when someone would instruct ember to mess around in _someone else's_ subtree, that's when you run in to problems -- but this is easy enough to avoid).

In my use case, I have:
- host ember app (could be any other app though, even react)
  - run through repl stuff
    - be given an element to put somewhere
      - render that element
        - in my case for https://limber.glimdown.com, the element is rendered via a component.
          so this component is mounting a whole new app, and nothing is is going to mess with the subtree.   
          This works via using the new `template()` API for the runtime compiler:
          ```js
          const element = await compiler.compile('gjs', code);
      
          component = template(`{{element}}`, { scope: () => ({ element }) });
          ```


packages/@ember/application/tests/application_test.js
```diff
@@ -74,14 +74,15 @@ moduleFor(
...
-          this.createSecondApplication({
-            rootElement: '#one-child',
-          })
-        );
-      });
+    [`@test you cannot make a new application that is a descendant of an existing application`](
+      assert
+    ) {
+      runTask(() =>
+        this.createSecondApplication({
```
kategengler: There is an `expectWarning`

NullVoxPopuli: legit

kategengler:
@ef4 Are you okay with this as a way to unblock @NullVoxPopuli without needing to fully deprecate and remove EventDispatcher (yet)?

kategengler (approved):

ef4:
No, I don't think this is a good idea. Either we support this case or we don't. If we do, then we should just drop the assertion entirely as in #20910. If we don't, then the assertion should stay. 

Until then @NullVoxPopuli can opt into "I know this is unsupported" mode by continuing to use pnpm patch. 

Also it makes no sense to only change this assertion without also changing the assertions around it. If "descendant of existing ember app" is OK then so is "ancestor of existing ember app", for example.

NullVoxPopuli:
makes sense -- I'd prefer to remove the assertions altogether as in https://github.com/emberjs/ember.js/pull/20910/files

but I don't yet know where this falls down. I know this is all in event_dispatcher code, but what uses the event_dispatcher?

How would someone know if they've created a broken situation?

kategengler:
Just from searching, it looks like @ember/component uses EventDispatcher for `on` as well as the browser events it can handle. There's setup and public properties on @ember/application that customize the event dispatcher (root element, custom events, the instance itself?)

kategengler:
Moving to draft since there isn't consensus here

PR #20918, opened 2025-06-09
[Next] Kill ArrayProxy, ObjectProxy, and PromiseProxy

wagenet:
For this to land in main we will need to have first deprecated `@ember/array/proxy`, `@ember/object/promise-proxy-mixin`, and `@ember/object/proxy`.


packages/@ember/-internals/runtime/tests/core/is_empty_test.js
```diff
@@ -1,27 +1,9 @@
...
-      let arrayProxy = ArrayProxy.create({ content: emberA([]) });
-      let objectProxy = ObjectProxy.create({ content: arrayProxy });
-
-      assert.equal(
-        true,
-        isEmpty(objectProxy),
-        'for an ArrayProxy inside ObjectProxy that has empty content'
-      );
-    }
+    // TODO: Add real tests here
```
wagenet, line 7: We should add some tests for this function if we want to keep it.

packages/@ember/array/index.ts
```diff
@@ -1853,6 +1846,10 @@ const MutableArray = Mixin.create(EmberArray, MutableEnumerable, {
     endPropertyChanges();
     return this;
   },
+
+  destroy() {
+    destroyObservers(this);
+  },
```
wagenet, line 1852: Somehow the previous ArrayProxy mixin still caused this to be present even though it wasn't mixed in here. This is slightly concerning.

NullVoxPopuli (approved):

packages/@ember/-internals/runtime/tests/core/is_empty_test.js
NullVoxPopuli: the utils are already deprecated -- so I don't know what this would test!

NullVoxPopuli: https://github.com/emberjs/rfcs/pull/1056

packages/@ember/array/index.ts
NullVoxPopuli: Once observers are removed tho 🤔

PR #20919, opened 2025-06-09
[Next] Kill Action Mixins

wagenet:
To merge this to main we would have to deprecate `send` and the actions hash.


packages/ember/tests/routing/query_params_test.js
```diff
@@ -740,46 +740,47 @@ moduleFor(
...
-      assert.equal(getTextOf(document.getElementById('test-value')), '2');
-      this.assertCurrentPath('/?foo=2');
-
-      document.getElementById('test-button').click();
-      await runLoopSettled();
-
-      assert.equal(getTextOf(document.getElementById('test-value')), '3');
-      this.assertCurrentPath('/?foo=3');
-    }
+    // TODO: Do we want to test this? Or are we ok with letting it go?
```
wagenet: We should review this.

NullVoxPopuli: why was it disabled?

we can add controllers via:
```js
this.add('controller:application', class extends Controller {
  @service router;
  foo = 1;
  increment = () => {
    this.incrementProperty('foo');
    this.router.refresh();
  }
});

```

PR #20926, opened 2025-06-10
Remove all view mixins

wagenet:
These are all internal so this should be able to merge without any deprecation work.


NullVoxPopuli (approved):
Seems fine -- tho I see a lot more stuff to delete!

PR #20936, opened 2025-06-14
[Next] Remove the barrel file (already RFC'd)

NullVoxPopuli:
This will happen on main during 7.0 / 7.1

The deprecation has already released in ember-source 6.5


NullVoxPopuli:
It's removed now

PR #20939, opened 2025-06-20
Add trustedHTML via alias/re-export of htmlSafe

NullVoxPopuli:
htmlsafe is a misleading phrase, so here we are aliasing to what we really mean: that we trust the HTML 

we should do an RFC to deprecate htmlSafe

### DIFF
--- a/packages/@ember/-internals/glimmer/index.ts
+++ b/packages/@ember/-internals/glimmer/index.ts
@@ -456,7 +456,7 @@ export {
   type FunctionBasedHelper,
   type FunctionBasedHelperInstance,
 } from './lib/helper';
-export { SafeString, trustedHTML, isTrustedHTML, htmlSafe, isHTMLSafe } from './lib/utils/string';
+export { SafeString, trustHTML, isTrustedHTML, htmlSafe, isHTMLSafe } from './lib/utils/string';
 export { Renderer, _resetRenderers, renderSettled } from './lib/renderer';
 export {
   getTemplate,
--- a/packages/@ember/-internals/glimmer/index.ts
+++ b/packages/@ember/-internals/glimmer/index.ts
@@ -456,7 +456,14 @@ export {
   type FunctionBasedHelper,
   type FunctionBasedHelperInstance,
 } from './lib/helper';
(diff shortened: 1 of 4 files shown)


packages/@ember/-internals/glimmer/index.ts
```diff
@@ -456,7 +456,7 @@ export {
   type FunctionBasedHelper,
   type FunctionBasedHelperInstance,
 } from './lib/helper';
-export { SafeString, htmlSafe, isHTMLSafe } from './lib/utils/string';
+export { SafeString, trustedHTML, isTrustedHTML, htmlSafe, isHTMLSafe } from './lib/utils/string';
```
wagenet: Not to bike shed on this, since I'm happy to see this land, even as-is, but what about `trustHTML` so the method is a verb?

wagenet: Also, I'd really love to see `SafeString` become `TrustedHTML` as well, but I know that's more of a lift.

NullVoxPopuli: am fan, verb fits more with what we're doing, and I'll reject all noun-proposals going forward for this haha

NullVoxPopuli: > Also, I'd really love to see SafeString become TrustedHTML as well, but I know that's more of a lift.

ya, I could alias this, too

kategengler:
I'm 👍 but types are failing

packages/@ember/-internals/glimmer/lib/utils/string.ts
```diff
@@ -5,39 +5,39 @@
...
   rendered in HTML, Ember will not perform any escaping.**
 
   Note:
 
   1. This does not *make* the string safe; it means that some code in your
-     application has *marked* it as safe using the `htmlSafe()` function.
+     application has *marked* it as trusted using the `trustHTML()` function.
 
-  2. The only public API for getting a `SafeString` is calling `htmlSafe()`. It
+  2. The only public API for getting a `TrutsedHTML` is calling `trustHTML()`. It
```
bertdeblock, line 16: Small typo => `TrutsedHTML` => `TrustedHTML`

PR #20958, opened 2025-08-01
fix: Make the inject deprecation message more understandable

MichalBryxi:
Maybe it's just me, but looking at: `Importing `inject` from `@ember/service` is deprecated. Please import `service` instead` has few issues:

- Which file / line triggered this issue? (This would be an ideal information to have, but I guess it's not really possible?)
- What does `importing X from Y` even mean? Are we talking about ES2015 keyword? Is that some Ember internal nomenclature? Something tied to the build system itself (like import in babel config?), etc...
- What does `please import service instead` mean? Can I get an example?

I personally **love** unambiguous, easy to follow instructions that has examples of the most likely cases and are copy&pasteable.

Hence this PR.

### DIFF
--- a/packages/@ember/service/index.ts
+++ b/packages/@ember/service/index.ts
@@ -27,7 +27,7 @@ export function inject(
   ...args: [] | [name: string] | ElementDescriptor
 ): PropertyDecorator | DecoratorPropertyDescriptor | void {
   deprecateUntil(
-    'Importing `inject` from `@ember/service` is deprecated. Please import `service` instead.',
+    "Importing `inject` from `@ember/service` is deprecated. Please import `service` instead. This likely means that you want to find lines: `import { inject as service } from '@ember/service'` and replace them with: `import { service } from '@ember/service';`",
     DEPRECATIONS.DEPRECATE_IMPORT_INJECT
   );
 


bertdeblock:
The deprecation should link to the guide I think:
https://deprecations.emberjs.com/id/importing-inject-from-ember-service

NullVoxPopuli:
The stack / log trace will take you to the line in your project where the offending code is.

I don't think it makes sense to change deprecation messages to have _likely_ hints (when the stack has the definitive location)


As @bertdeblock pointed out -- the deprecation includes the link: https://deprecations.emberjs.com/id/importing-inject-from-ember-service/

which has the details you seek <3 


I'm going to close this PR, but I really do appreciate your efforts to improve the messaging! <3 

Unclear errors _are_ bugs, but I believe all the information is currently available to you.
Though, since it hasn't been obvious, I'd like to dive in with you as to what you saw, and how we can improve what was presented to you. 
Do you have any screenshots?

MichalBryxi:
Screenshot 2025-10-26 at 00 31 43

Not sure anymore which project prompted this PR. But I stand corrected as now the respective user space code seems to be mentioned in the stack trace correctly. Which is an-ok:

> @http://localhost:4200/app/components/station/index.gts:27:6

PR #19257, opened 2020-11-09
Add test ensuring `@model` is never updated to an undefined value

rwjblue:

Issue #18987, opened 2020-05-20
@model value timing issue


HeroicEric:
We run into a difference in behavior between `@model` and `this.model` that I'm not sure how to best describe but here is a twiddle that demonstrates the issue:

https://ember-twiddle.com/29dd9ad4820bdf987c92eb7acf1d7697?openFiles=routes.index%5C.js%2Ctemplates.components.person%5C.hbs&route=%2Fsome-other-route

It seems that the value of `@model` changes to `undefined` before all of the components in the template have been destroyed. This is different than the behavior seen when using `this.model`, where the value remains until after all of the components in the template have been destroyed.

For example, if we return a value from the model hook:

```js
// app/routes/index.js
export default class IndexRoute extends Route {
  model() {
    return {
      firstName: 'Little',
      lastName: 'Sebastian'
    };
  }
}
```

And pass that value to a component in the route's template:

```hbs
{{! app/templates/index.hbs }}
<Person @person={{@model}} />
```

And try to access the value in the component's `willDestroy` lifecycle hook:

```js
// app/components/person.js
export default class extends Component {
  willDestroy() {
    console.log({ person: this.args.person });
  }
}
```

Transitioning from the `index` route to another route causes the value of `@model` to become `undefined` by the time the `willDestroy` hook for the component runs.

This issue goes away of we instead pass the value to the component using `this.model`:

```hbs
{{! app/templates/index.hbs }}
<Person @person={{this.model}} />
```

NullVoxPopuli:
I tried making a failing test case in `packages/@ember/-internals/glimmer/tests/integration/application/rendering-test.js`, but ran in to an issue where this bug isn't present with `@ember/component`.

So, this may be an issue with `@glimmer/component`s component-manager / maybe how `@model` and `this.model` are interpreted differently? unsure now.

NullVoxPopuli:
for posterity, the passing test (using `@ember/component`):
```ts
    async ['@feature(EMBER_ROUTING_MODEL_ARG) interior mutations on the model: @model exists as long as this.model (in willDestroy of component)'](
      assert
    ) {
      assert.expect(4);

      this.router.map(function() {
        this.route('foo');
      });

      this.add(
        'route:index',
        Route.extend({
          model() {
            return { firstName: 'Carol', lastName: 'Danvers' };
          },
        })
      );

      this.addComponent('person', {
        ComponentClass: class PersonComponent extends Component {
          tagName = '';
          willDestroy() {
[...]

PR #20959, opened 2025-08-01
Add a test that verifies `@model` is stable between route transitions

Windvis:
This adds a failing test for the issue mentioned in https://github.com/emberjs/ember.js/issues/18987 based on the [StackBlitz reproduction](https://stackblitz.com/edit/ember-cli-editor-output-exrpkxkg?file=app%2Ftemplates%2Fapplication.hbs) and supersedes https://github.com/emberjs/ember.js/pull/19257.

I needed to do some changes to the Vite and CI setup to make the `@glimmer/component` imports working in the test files. Not sure if what I did is the best solution but it works for the reproduction at least.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,6 +73,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup
+      - name: rollup
+        run: pnpm build
       - name: build
         run: pnpm vite build --mode=development
       - name: test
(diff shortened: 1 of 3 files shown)


NullVoxPopuli:
do you happen to know if this ever worked?

Windvis:
> do you happen to know if this ever worked?

I don't think so, no. As far as I know it has been an issue since `@model` was introduced, which is why the [template-lint rule](https://github.com/ember-template-lint/ember-template-lint/blob/main/docs/rule/no-model-argument-in-route-templates.md) exists.

It's just a tricky to reproduce issue so it _seemed_ that it was "solved" at some point in time.

PR #20962, opened 2025-08-18
RFC#1099: renderComponent

NullVoxPopuli:
Implements RFC#1099 - `renderComponent()`

Copies over the work from 
- https://github.com/emberjs/ember.js/pull/20781
  But this PR was created before some VM updates, so the changes from the upgrade PR here
  https://github.com/emberjs/ember.js/pull/20842/files
  need to be re-applied to the work done in #20781 

See RFC for differences between the original PR and this one:
- https://github.com/emberjs/rfcs/pull/1099

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/components/render-component-test.ts
+++ b/packages/@ember/-internals/glimmer/tests/integration/components/render-component-test.ts
@@ -20,10 +20,13 @@ import { run } from '@ember/runloop';
 import { associateDestroyableChild, registerDestructor } from '@glimmer/destroyable';
 import { renderComponent, type RenderResult } from '../../../lib/renderer';
 import { trackedObject } from '@ember/reactive';
+import { cached, tracked } from '@glimmer/tracking';
+import Service, { service } from '@ember/service';
+import type Owner from '@ember/owner';
 
 class RenderComponentTestCase extends AbstractStrictTestCase {
   component: (RenderResult & { rerender: () => void }) | undefined;
-  owner: object;
+  owner: Owner;
 
   constructor(assert: QUnit['assert']) {
     super(assert);
@@ -34,6 +37,14 @@ class RenderComponentTestCase extends AbstractStrictTestCase {
 
   get element() {
     return document.querySelector('#qunit-fixture')!;
+  }
+
+  assertChange({ change, expect }: { change: () => void; expect: string }) {
+    run(() => change());
+
+    assertHTML(expect);
+
+    this.assertStableRerender();
   }
 
   renderComponent(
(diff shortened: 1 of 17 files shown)


packages/@ember/-internals/glimmer/tests/integration/components/render-component-test.ts
```diff
@@ -0,0 +1,591 @@
+import {
+  AbstractStrictTestCase,
+  assertClassicComponentElement,
+  assertHTML,
+  buildOwner,
+  clickElement,
+  defComponent,
+  defineComponent,
```
mansona, line 8: what's the difference between defComponent and defineComponent? 🙈

NullVoxPopuli: defineComponent is traditional loosemode and defComponent is strict only

locks: why not `defineStrictComponent`?

NullVoxPopuli: idk -- this code was copied from a prior implementation attempt

It's ultimately not important right now, but I'm happy to rename this util in a separate PR :tada:

NullVoxPopuli: I should clarify, yesterday I wasn't in a mood to push updates unless there were more comments to make pushing (and all of CI) worthwhile.

Apologies -- defineStrictComponent is just obectively better, and I'll rename. thank you for the suggestion! <3

NullVoxPopuli: actually, I was wrong:
```
/**
 * The signature of this test helper is closer to the signature of the
 * `template()` function in `@ember/template-compiler`. It can express the same
 * functionality as {@linkcode defineComponent}, but most tests still use
 * `defineComponent`. Migrating tests from {@linkcode defineComponent} is
 * straightforward, and there's no *semantic* reason not to do so.
 */
```

it seems I should punt on changes here and instead make `template` work in tests (and get rid of this helper)

ef4: Rather than `template()`, I think it would be fine to go straight to `<template></template>` in the tests.

The test suite runs in Vite, if it doesn't already have template tag support enabled adding it should be trivial.

packages/@ember/-internals/glimmer/tests/integration/components/render-component-test.ts
```diff
@@ -0,0 +1,591 @@
...
+
+      this.renderComponent(Root, { args, expect: '2' });
+
+      this.assertChange({
+        change: () => args.foo++,
+        expect: '3',
+      });
+    }
+
+    '@skip when args are a custom tracked class, the rendered component response appropriately'() {
```
ef4, line 203: I'm curious why this doesn't just naturally work?

ef4 (approved):
I commented above on refactoring goal for tests and also asked a question about a skipped test, but neither of those is a blocker for this PR.

packages/@ember/-internals/glimmer/tests/integration/components/render-component-test.ts
NullVoxPopuli: yea, I consider this not working as a bug -- i suspect it is because of how ArgsProxy relies on `Object.keys` and `@tracked` properties are not enumerable.

Fixing this may require changes to how we handle args / ArgsProxy

ef4: Oof, yeah, non-enumerability of tracked always sneaks up on me. You don't notice it until you notice it.

PR #20968, opened 2025-09-04
Implement RFC #334 - Deprecate Ember Utils

wagenet:
https://github.com/emberjs/rfcs/pull/1056

I think we just need to settle on the deprecation id and exact versions. We may also want to add `@deprecated` to the JSDoc.


wagenet:
And the deprecation guide: https://github.com/ember-learn/deprecation-app/pull/1417

PR #20970, opened 2025-09-04
Deprecate Evented

wagenet:
We still need to figure out correct deprecation ids.

https://github.com/ember-learn/deprecation-app/pull/1404


wagenet:
I just realized that the approach of two mixins may not work with `detect`. We might be able to do what we want with a Proxy.

_EDIT:_ A Proxy won't easily work either because we look for whether it's in a `Set` and the Proxy-wrapped version is not _identical_ to the original. As annoying as this is, the best approach is probably just to have a way to temporarily disable the deprecation when used in Ember internals.

_EDIT 2:_ I've reworked this and it should work as expected now.

packages/@ember/-internals/runtime/lib/mixins/evented.ts
ef4, line 1: Notes from RFC discussion: this file might have been unintentionally included.

packages/@ember/-internals/deprecations/index.ts
```diff
@@ -122,6 +122,13 @@ export const DEPRECATIONS = {
     until: '7.0.0',
     url: 'https://deprecations.emberjs.com/id/using-amd-bundles',
   }),
+  DEPRECATE_EVENTED: deprecation({
+    id: 'ember-evented',
+    for: 'ember-source',
+    since: { available: '6.12.0' },
```
kategengler, line 128:
```suggestion
    since: { available: '7.1.0' },
```

packages/ember/barrel.ts
```diff
@@ -520,17 +520,26 @@ namespace Ember {
...
-
-   `window.addEventListener('error', eventHandler);`
-
-   For promise rejections, you'll want to use the unhandledrejection event.
-   `window.addEventListener('unhandledrejection', eventHandler);`
-
-   If you really need the original behavior:
-
-   `import { getOnerror, setOnerror } from '@ember/-internals/error-handling';`
+    A function may be assigned to `Ember.onerror` to be called when Ember
```
kategengler, line 523: This reverts some docs changes

.gitignore
```diff
@@ -54,3 +54,4 @@ npm-debug.log
 # couple of the files. Once it is, we can switch this over to just ignoring
 # `types/stable` entirely.
 types/stable
+.puppeteer-cache/
```
kategengler, line 57: This was removed intentionally -- we no longer use puppeteer

kategengler:
Do you think you could split this into a few PRs?
- Replacing the API on router
- Replacing internal usages
- Deprecating evented

PR #20974, opened 2025-09-08
[Bugfix Beta] Fix issue with renderer's cleanup since the introduction of renderComponent

NullVoxPopuli:
From: https://github.com/bertdeblock/failing-ember-beta-test

Very surprised our existing tests didn't catch this

It turns out we haven't been doing much owner-linked destruction testing.

If folks use clearRender, things work great.
The issue is that if the owner is destroyed, we want everything used with the owner to also be destroyed -- and somehow this linkage got missed during the renderComponent work.

Implementation PR, for reference: https://github.com/emberjs/ember.js/pull/20962

### DIFF
--- a/smoke-tests/scenarios/basic-test.ts
+++ b/smoke-tests/scenarios/basic-test.ts
@@ -110,7 +110,7 @@ function basicTest(scenarios: Scenarios, appName: string) {
               import { module, test } from 'qunit';
               import { clearRender, render } from '@ember/test-helpers';
               import { setupRenderingTest } from 'ember-qunit';
-              import { registerDestructor } from '@ember/destroyable';
+              import { destroy, registerDestructor } from '@ember/destroyable';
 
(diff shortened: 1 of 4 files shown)


smoke-tests/scenarios/basic-test.ts
```diff
@@ -106,6 +106,37 @@ function basicTest(scenarios: Scenarios, appName: string) {
             `,
           },
           integration: {
+            'destruction-test.gjs': `
```
NullVoxPopuli, line 109: this test is passing in 6.7

smoke-tests/scenarios/basic-test.ts
NullVoxPopuli, line 1: I couldn't find any place `@glimmer/component` was tested, so I just added some tests here

ef4 (approved):

PR #20982, opened 2025-09-22
Add thank you to our contributors

NullVoxPopuli:
Since the main branch is protected, this new workflow will create a PR [like here](https://github.com/universal-ember/ember-primitives/pull/535), that contains content like these examples:

https://discord.com/channels/480462759797063690/491905849405472769/1419786480611491981

### DIFF
--- a/.github/workflows/contributors.yml
+++ b/.github/workflows/contributors.yml
@@ -1,0 +1,17 @@
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  contrib-readme-job:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+
+    steps:
+      - name: Contribute List
+        uses: akhilmhdh/contributors-readme-action@v2.3.11
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/contributors.yml
+++ b/.github/workflows/contributors.yml
@@ -7,7 +7,6 @@ jobs:
   contrib-readme-job:
     runs-on: ubuntu-latest
     permissions:
-      contents: write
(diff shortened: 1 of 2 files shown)


kategengler:
Is this valuable? Won't it be a long list that gets skimmed over?

NullVoxPopuli:
I think it could make contributors feel proud of their accomplishments, and something they can show off to their friends. Hopefully increasing likelihood of contribution so new folks are more willing to do their first contribution

.github/workflows/contributors.yml
```diff
@@ -0,0 +1,17 @@
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  contrib-readme-job:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
```
kategengler: I'm not comfortable with it having `contents: write`. PRs seem fine. I wonder if it were just missing the permissions would it default to PR?

NullVoxPopuli: Only one way to find out (i removed it)

kategengler: I read the code of the action -- it won't. It would be really easy to alter the action to have a flag, though.

NullVoxPopuli: https://github.com/akhilmhdh/contributors-readme-action/issues/87

johanrd:
> Is this valuable? Won't it be a long list that gets skimmed over?

Yeah, I may agree with this: The repo page already has a 'contributors' aside on the right, which may be enough?
Screenshot 2025-10-06 at 14 16 39

NullVoxPopuli:
Fair enough

kategengler:
I don't mind adding it but do not want to give it `contents: write`

PR #20988, opened 2025-09-28
[BUGFIX beta] Drop ember-cli-htmlbars-inline-precompile from component-test blueprint

mansona:
Closes https://github.com/emberjs/ember.js/pull/20983

I asked Florian to target the oldest branch that it applies to but we don't do it that way in ember-source 🙈 sorry for the confusion

@kategengler this needs to be backported to all active versions, at least to beta so I can fix the tutorial 👍

### DIFF
--- a/blueprints/component-test/index.js
+++ b/blueprints/component-test/index.js
@@ -2,7 +2,6 @@
 
 const path = require('path');
 const stringUtil = require('ember-cli-string-utils');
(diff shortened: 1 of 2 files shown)


NullVoxPopuli (approved):

kategengler:
Sorry I forgot about this!

I do the beta tomorrow unless you need it sooner?

PR #20993, opened 2025-10-20
test: add edge case tests for number comparison in compare utility

virajsabhaya23:
missing edge cases for comparison between diff sets of number


packages/@ember/utils/tests/compare_test.js
```diff
@@ -84,5 +84,15 @@ moduleFor(
       assert.equal(compare('b', zero), 0, 'Second item comparable - returns  0 (negated)');
       assert.equal(compare('c', one), -1, 'Second item comparable - returns  1 (negated)');
     }
+    
+    ['@test comparing numbers with edge cases'](assert) {
```
NullVoxPopuli, line 88: I think the above tests cover these 🤔

kategengler: Yes, they are all covered above.

PR #20994, opened 2025-10-22
[BUGFIX release] loosemode interop not working with renderComponent

NullVoxPopuli:
We don't need a `StrictResolver`, because there is no resolving in strict mode.

The resolver that is constructed for the strict renderer is what is used for the whole of that rendering tree.

When a strict-mode component encounters something not in the scope bag, it throws a compile-time syntax error -- so resolution never occurs.

Big thanks to @vlascik for reporting and providing a repro


packages/@ember/-internals/glimmer/tests/integration/components/render-component-test.ts
```diff
@@ -597,6 +597,32 @@ moduleFor(
...
+      this.renderComponent(Root, { expect: 'Hi: THERE' });
+
+      run(() => destroy(this));
+
+      assertHTML('');
+    }
+
+    '@test strict-mode components cannot lookup things in the registry'(assert: Assert) {
+      this.owner.register('helper:a-helper', (str: string) => str.toUpperCase());
+      let Root = defComponent('{{a-helper "hi"}}', {});
```
NullVoxPopuli: isStrictMode is false for some reason -- we could key off this and use the StrictResolver logic

Image

vlascik:
there seems to be also something else going on - in the original repro, the corrected code now passes the first call to renderComponent, but the second one, even if it's the same render function call, explodes again with `TypeError: Cannot read properties of null (reading 'syscall')`

NullVoxPopuli:
yeah, that is indeed a separate issue -- worthy of a separate PR and repro -- and one that I haven't been able to repro myself consistently.

Lets chat in discord about how to re-create that error

PR #20995, opened 2025-10-30
[BUGFIX] renderComponent error: 'attempted to close a tracking frame, but one was not open'

NullVoxPopuli:
This happens when:
- renderComponent is called twice, sequentially
- the renderedComponent calls another component
- and the rendered component calls a helper
- the owner is passed to renderComponent

these 4 things are required to reproduce the error
- not passing the owner results in a passing test
- not invoking a child component results in a passing test
- not invoking a helper results in a passing test

### DIFF
--- a/packages/@ember/-internals/glimmer/tests/integration/components/render-component-test.ts
+++ b/packages/@ember/-internals/glimmer/tests/integration/components/render-component-test.ts
@@ -656,7 +656,9 @@ moduleFor(
       );
 
       this.renderComponent(Root, {
-        expect: [`<div id="a">hi:a</div><br>`, `<div id="b">hi:b</div>`, ``, ``].join('\n'),
+        expect: [`<div id="a">a:Hi: THERE</div><br>`, `<div id="b">b:Hi: THERE</div>`, ``, ``].join(
+          '\n'
+        ),
       });
 
       run(() => destroy(this));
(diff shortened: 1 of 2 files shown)


packages/@ember/-internals/glimmer/tests/integration/components/render-component-test.ts
```diff
@@ -604,6 +604,39 @@ moduleFor(
...
+        });
+      }
+      let A = defComponent('a:<Child />', { scope: { Child } });
+      let B = defComponent('b:<Child />', { scope: { Child } });
+      let Root = defComponent(
+        [
+          `<div id="a"></div><br>`,
+          `<div id="b"></div>`,
+          `{{render A 'a' owner}}`,
+          `{{render B 'b' owner}}`,
```
void-mAlex, line 625:
```suggestion
          `{{render B 'b' owner}}`,
          `{{render B 'b'}}`,
          `{{render A 'a'}}`,
```
what happens when you mix owner and non owner calls?

kategengler (approved):

packages/@ember/-internals/glimmer/tests/integration/components/render-component-test.ts
NullVoxPopuli: a good question -- I'll explore in another PR -- because it may reveal an existing problem about trying to render different things into the same element

Issue #21005, opened 2025-11-16
EmberArray.reduce does not mirror native reduce when initialValue is omitted


weillercarvalho:
### 🐞 Describe the Bug

-    EmberArray.prototype.reduce (packages/@ember/array/index.ts:1383) assumes an initialValue always exists. When you omit it, the accumulator starts as undefined and the callback runs against every item, so results diverge from native JS behavior and documented examples.
  - When the enumerable is empty and no initialValue is provided, the helper silently returns undefined instead of throwing TypeError('Reduce of empty array with no initial value') like the native API.

### 🔬 Minimal Reproduction

```
import { A } from '@ember/array';

  let numbers = A([1, 2, 3]);
  numbers.reduce((sum, value) => sum + value);
  // => NaN (should be 6)

  let empty = A([]);
  empty.reduce(() => 0);
  // => undefined (should throw TypeError)
```

### 😕 Actual Behavior
  - The first callback invocation receives sum === undefined, so any arithmetic or string concatenation yields incorrect results.
  - Empty enumerables without an initialValue never throw, masking bugs and diverging from the spec/MDN docs referenced in our own API docs.

### 🤔 Expected Behavior
  - When initialValue is omitted, the first element should seed the accumulator and iteration should begin at index 1, matching Array.prototype.reduce.
  - Calling reduce on an empty enumerable without an initialValue should synchronously throw TypeError('Reduce of empty array with no initial value').

### 🌍 Environment
 
- Ember: 6.10.0-alpha.1 (repo HEAD)
- Ember-CLI: 6.3.1
- Node.js/npm: 22.19.0 / 10.9.3
- OS: Ubuntu 22.04.5 LTS (WSL2)
- Browser: Chrome headless (Puppeteer 24.3.0 bundle)

kategengler:
We have an accepted RFC to deprecate all of `@ember/array`, so we will not be making any changes to EmberArray.

`EmberArray` has not generally been needed for quite some time. Please read the RFC and the [draft deprecation guides](https://github.com/ember-learn/deprecation-app/pull/1406/files) for how to move away.

PR #21006, opened 2025-11-16
[BUGFIX beta] Align EmberArray.reduce with native semantics, related with issue #21005.

weillercarvalho:
## Summary

  - Fixes EmberArray.prototype.reduce so it behaves like native Array.prototype.reduce when initialValue is omitted.
  - Adds regression coverage for both the implicit‐initial scenario and the empty-array TypeError.

  ## Customer Impact / Use Case

  Apps relying on EmberArray.reduce without an explicit initialValue currently get undefined as the first accumulator and never see
  the expected TypeError on empty collections, yielding incorrect results and masking bugs.

  ## Implementation Details

  - Detect whether initialValue was provided (arguments.length > 1), seed the accumulator accordingly, and reuse objectAt to keep proxy support intact.
  - Throw TypeError('Reduce of empty array with no initial value') when the array is empty and no seed is passed.
  - Extend packages/@ember/-internals/runtime/tests/array/reduce-test.js with two new tests covering these code paths across every helper used in runArrayTests.
  
  ## Tests

All unit tests passed after modification.
Captura de tela 2025-11-16 073436

### DIFF
--- a/packages/@ember/-internals/runtime/tests/array/reduce-test.js
+++ b/packages/@ember/-internals/runtime/tests/array/reduce-test.js
@@ -8,6 +8,12 @@ class ReduceTests extends AbstractTestCase {
     this.assert.equal(res, 6);
   }
 
+  '@test uses the first item as the initial accumulator when not provided'() {
+    let obj = this.newObject([1, 2, 3]);
+    let res = obj.reduce((previousValue, item) => previousValue + item);
+    this.assert.equal(res, 6);
+  }
+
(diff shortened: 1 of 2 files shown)


johanrd:
this can probably be closed. Tests seem to be succeeding, at least after the very similar fix and tests in ~~https://github.com/emberjs/ember.js/pull/21006~~ edit: https://github.com/emberjs/ember.js/pull/21168

NullVoxPopuli:
tyty

PR #21014, opened 2025-11-27
Rename default export `GlimmerComponent` to `Component`

evoactivity:
Because the export is named `GlimmerComponent` typescript is unable to offer it as a suggestion when using `Component`. By convention we have people write `import Component from '@glimmer/component`, as such they will never be suggested the correct import.

image

You have to write `import Gli...` to get the correct suggestion.

image

Should have no effect on consumers as this is a default export and as such can be named anything by the consumer.

One risk has been raised by @kategengler in that this might cause problems with the docs and a redirect might need to be added.

### DIFF
--- a/packages/@ember/-internals/glimmer/index.ts
+++ b/packages/@ember/-internals/glimmer/index.ts
@@ -450,7 +450,7 @@ export { default as RootTemplate } from './lib/templates/root';
 export { default as Input } from './lib/components/input';
 export { default as LinkTo } from './lib/components/link-to';
 export { default as Textarea } from './lib/components/textarea';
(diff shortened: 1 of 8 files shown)


evoactivity:
An alternative would be to _only_ rename GlimmerComponent, then both would be recommended by typescript.

evoactivity:
ember's Component export is used in a lot more places than my ide was aware of.

I think just renaming the GlimmerComponent should be enough.

NullVoxPopuli:
> ember's Component export is used in a lot more places than my ide was aware of.

probs better as a separate pr anyway

NullVoxPopuli (approved):

PR #21017, opened 2025-12-02
Fix NaN handling in compare() function

Bitshifter-9:
- Add explicit NaN checks in spaceship helper to ensure it always returns -1, 0, or 1
- Treat NaN as less than all other numbers and equal to itself for ordering
- Add comprehensive test coverage for NaN edge cases
- Fixes potential runtime errors from unexpected NaN return values
#21016

### DIFF
--- a/packages/@ember/utils/lib/compare.ts
+++ b/packages/@ember/utils/lib/compare.ts
@@ -40,6 +40,20 @@ type Compare = -1 | 0 | 1;
 //                                `._________`-.   `.   `.___
 //                                              SSt  `------'`
 function spaceship(a: number, b: number): Compare {
+  // Handle NaN cases explicitly to ensure we always return -1, 0, or 1
+  const aIsNaN = Number.isNaN(a);
+  const bIsNaN = Number.isNaN(b);
+
+  if (aIsNaN && bIsNaN) {
+    return 0; // NaN === NaN for ordering purposes
+  }
+  if (aIsNaN) {
+    return -1; // NaN < any number
+  }
+  if (bIsNaN) {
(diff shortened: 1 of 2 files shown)


NullVoxPopuli:
Tyty for contributing!

I have a question tho,
When actually using compare in a sort function, is the behavior any different?

If it is, could you you add a test like that? Thanks!

Bitshifter-9:
@NullVoxPopuli 
I've verified the behavior when using 
compare
 with Array.prototype.sort(). It works as expected: since we treat NaN as smaller than all other numbers, NaN values are sorted to the beginning of the array.

I've added a new test case to explicitly cover this sorting scenario:

    ['@test sort with compare'](assert) {
      let arr = [NaN, 10, -5, 0, NaN, 5];
      arr.sort(compare);

      assert.ok(Number.isNaN(arr[0]), 'index 0 is NaN');
      assert.ok(Number.isNaN(arr[1]), 'index 1 is NaN');
      assert.equal(arr[2], -5, 'index 2 is -5');
      assert.equal(arr[3], 0, 'index 3 is 0');
      assert.equal(arr[4], 5, 'index 4 is 5');
      assert.equal(arr[5], 10, 'index 5 is 10');
    }

NullVoxPopuli:
> with Array.prototype.sort(). It works as expected:

Even before this change?

ef4:
`compare()` is already deprecated, so we are unlikely to decide to change it.

Also, this directly contradicts Javascript semantics:

```js
aassert.equal(compare(NaN, NaN), 0, 'NaN should equal NaN for ordering');
```

Because `NaN == NaN` evaluates to false.

PR #21024, opened 2026-01-06
[Bugfix LTS] Add @ember/reactive into the AMD bundles

NullVoxPopuli:
Solves
- https://github.com/tracked-tools/tracked-built-ins/issues/445

(hopefully)


ef4:
I think if you add a unit test in `smoke-tests/scenarios/basic-test.ts` it will reproduce the failure and show the fix working.

PR #21046, opened 2026-01-06
Add benchmark infra with starting basic bench

NullVoxPopuli:
I looked at https://github.com/TracerBench/tracerbench-compare-action
and opened https://github.com/TracerBench/tracerbench-compare-action/issues/93
I don't feel tracerbench-compare-action _yet_ supports how I think we should run our benchmark.

It works. Also deletes the old benchmark.

Running `pnpm bench` generates you a report like this:
image

and each measurement has details we can look at
image


ef4:
I tried `pnpm bench` but I don't think that's an up-to-date instruction?

NullVoxPopuli:
Note: I've only been able to run the bench on linux, mac and I do not get along at all :(

PR #21048, opened 2026-01-07
Implement `(element 'xyz')` 

NullVoxPopuli:
- [x]  copy tests from ember-element-helper
- [ ] implement the element keyword in strict mode (safest, existing using is shadowed)

Need to take a quick detour to add gjs/gts testing in this repo


NullVoxPopuli:
Implemented in another pr

